PR #579, opened 2023-08-17
refactor handlers for simplicity

krancour:
There's a lot of unnecessary complexity in the handler pattern we've been using.

Other than vanity (handlers in their own package because of some vague sense that, hierarchically, they are deserving of their own folder), there is no real reason for handlers to be in their own separate package. No one besides the API server uses these. They are, unquestionably _part of_ the API server -- which does very little _besides_ delegate to these handlers.

The choice to keep the handlers in a separate package also forced us into several layers of indirection that were hard to follow and it also forced our hand on moving server config to a separate package to break a dependency cycle that otherwise would not have existed. (imho, it's more straightforward to users of a package if config types live in the same package as the things they configure, so the fact that we were forced into this created another smell.)

This PR moves handler functions into the `api` package and refactors them to use the `server` type as a receiver, which eliminates a lot of indirection that was needed to accommodate passing around things like kubeclients when handlers were in a separate package.

Overall this should be cleaner, easier to follow, and easier to replicate when new handlers are added.

Note to reviewers: When files are moved _and_ modified, Git has a harder time recognizing it as a move/rename, so I broke this PR into two commits: one to move files without any changes and another to actually perform the changes. This means it will be easiest to review this PR by looking directly at commit aeaf2c9919b404b167d37cc1789278c2959eedf6


alexmt (commented):
I agree with the change.  In fact me and @gdsoumya implemented the same refactoring in Akuity Platform API. 

@devholic can you please take a look at PR as well? I want to make sure we are on the same page and not quietly fighting for different API designs.

jessesuen (approved):
I agree with Alex/Kent. Let's go ahead and merge this since it's going to be hard to maintain this PR

PR #583, opened 2023-08-21
add optional field to image subscriptions to link it back to a source CODE (not config) repo 

krancour:
... as discussed this morning.

### DIFF
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -118,6 +118,14 @@ type ImageSubscription struct {
 	//+kubebuilder:validation:MinLength=1
 	//+kubebuilder:validation:Pattern=`^(([\w\d\.-]+)(:[\d]+)?/)?[a-z0-9-]+(/[a-z0-9-]+)*$`
 	RepoURL string `json:"repoURL"`
+	// GitRepoURL optionally specifies the URL of a Git repository that contains
+	// the source code for the image repository referenced by the RepoURL field.
+	// When this is specified, Kargo MAY be able to infer and link to the exact
+	// revision of that source code that was used to build the image.
+	//
+	//+kubebuilder:validation:Optional
+	//+kubebuilder:validation:Pattern=`^(((https?://)|([\w-]+@))([\w\d\.]+)(:[\d]+)?/(.*))?$`
+	GitRepoURL string `json:"gitRepoURL,omitempty"`
 	// UpdateStrategy specifies the rules for how to identify the newest version
 	// of the image specified by the RepoURL field. This field is optional. When
 	// left unspecified, the field is implicitly treated as if its value were
@@ -561,6 +569,10 @@ func (e *StageStateStack) Push(states ...StageState) {
 type Image struct {
 	// RepoURL describes the repository in which the image can be found.
 	RepoURL string `json:"repoURL,omitempty"`
+	// GitRepoURL specifies the URL of a Git repository that contains the source
+	// code for the image repository referenced by the RepoURL field if Kargo was
+	// able to infer it.
+	GitRepoURL string `json:"gitRepoURL,omitempty"`
 	// Tag identifies a specific version of the image in the repository specified
 	// by RepoURL.
 	Tag string `json:"tag,omitempty"`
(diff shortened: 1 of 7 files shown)


api/v1alpha1/stage_types.go
```diff
@@ -118,6 +118,14 @@ type ImageSubscription struct {
...
 	//+kubebuilder:validation:Pattern=`^(([\w\d\.-]+)(:[\d]+)?/)?[a-z0-9-]+(/[a-z0-9-]+)*$`
 	RepoURL string `json:"repoURL"`
+	// GitRepoURL optionally specifies the URL of a Git repository that contains
+	// the source code for the image repository referenced by the RepoURL field.
+	// When this is specified, Kargo MAY be able to infer and link to the exact
+	// revision of that source code that was used to build the image.
+	//
+	//+kubebuilder:validation:Optional
+	//+kubebuilder:validation:Pattern=`^(((https?://)|([\w-]+@))([\w\d\.]+)(:[\d]+)?/(.*))?$`
+	GitRepoURL string `json:"gitRepoURL"`
```
jessesuen, line 128: 1.  I feel this should have `,omitempty`. 
2. I'd like to avoid confusion between `repoURL` and `gitRepoURL`. `git` is not actually the differentiator here (e.g. RepoURL might also be git) but rather, it's the URL of the app source code. So I propose instead:
* `appRepoURL`
* `appSourceRepoURL`

krancour: Let me double-check on `omitempty`. There are some peculiar/counter-intuitive rules about how that impacts the CRD generation.

The second point is a little off, however. "git" _is_ the differentiator here. The struct that this is enclosed in is `ImageSubscription` and `RepoURL` is the URL of an _image_.

I think `GitRepoURL` accurately denotes that this is the source repo that corresponds to that image.

krancour: Ah... yes. Missing `omitempty` here was an oversight. Thanks for catching it.

What I was thinking of was the reverse -- _having_ omitempty forced a required field to be optional in the CRD, no matter that the kubebuilder annotations say.

Fixing.

krancour: Fixed in d061a27a28a6230d3f12e755a651bc66ee0e8219

jessesuen: > RepoURL is the URL of an image.

Oh, I see. In that case gitRepoURL could work. 

> The struct that this is enclosed in is `ImageSubscription`

If this is only in the `ImageSubscription`, will we be able to show linkage from either (1) a Stage that subscribes to another Stage or (2) a Stage that subscribes to a git location?

jessesuen (approved):

api/v1alpha1/stage_types.go
krancour: Hmmmm... not easily. Thanks for pointing that out. Let me think on this. I've got an idea that could work and it would probably lessen the effort on the front end pretty significantly. Going to convert this to a draft for now.

krancour: @jesse

> If this is only in the ImageSubscription, will we be able to show linkage from either (1) a Stage that subscribes to another Stage or (2) a Stage that subscribes to a git location?

I've updated the PR to account for this. It's in the state now, too.

internal/controller/stages/images.go
```diff
@@ -58,11 +61,29 @@ func (r *reconciler) getLatestImages(
...
+		if strings.HasPrefix(gitRepoURL, baseUrl) {
+			return fn(gitRepoURL, tag)
+		}
+	}
+	return ""
+}
+
+func getGithubImageSourceURL(gitRepoURL, tag string) string {
+	return fmt.Sprintf("%s/tree/%s", git.NormalizeGitURL(gitRepoURL), tag)
+}
```
jessesuen, line 89: I think this reconstruction can happen in UI? At least that's where Argo CD is doing it

krancour: I think that's one way to do it, but this way, CLI and `kubectl` users still get the advantage of it.

jessesuen (approved):

PR #584, opened 2023-08-21
bcrypt admin password

krancour:
... as discussed this morning.

### DIFF
--- a/charts/kargo/templates/api/deployment.yaml
+++ b/charts/kargo/templates/api/deployment.yaml
@@ -41,7 +41,7 @@ spec:
 {{- if .Values.api.adminAccount.enabled }}
             - name: ADMIN_ACCOUNT_ENABLED
               value: "true"
-            - name: ADMIN_ACCOUNT_PASSWORD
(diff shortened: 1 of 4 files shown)


charts/kargo/templates/api/deployment.yaml
```diff
@@ -41,7 +41,7 @@ spec:
 {{- if .Values.api.adminAccount.enabled }}
             - name: ADMIN_ACCOUNT_ENABLED
               value: "true"
-            - name: ADMIN_ACCOUNT_PASSWORD
+            - name: HASHED_ADMIN_ACCOUNT_PASSWORD
```
jessesuen, line 44:
```suggestion
            - name: ADMIN_ACCOUNT_PASSWORD_HASH
```

Consistent prefixes for related configs is preferable

krancour: Agreed

krancour: Fixed in cd2d7101c22a13e71453bd6bf9bf318f0de289d3

jessesuen (approved):

PR #586, opened 2023-08-22
cli: account for jwts that don't expire

krancour:
I've just discovered that service account tokens can be JWTs that don't expire, so we need to account for that when checking if a token the CLI is holding in its configuration is expired or not.

### DIFF
--- a/internal/cli/client/token_refresher.go
+++ b/internal/cli/client/token_refresher.go
@@ -50,19 +50,24 @@ func (t *tokenRefresher) refreshToken(
 	insecureTLS bool,
 ) (config.CLIConfig, error) {
 	jwtParser := jwt.NewParser(jwt.WithoutClaimsValidation())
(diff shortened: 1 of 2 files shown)


internal/cli/client/token_refresher.go
```diff
@@ -50,19 +50,24 @@ func (t *tokenRefresher) refreshToken(
 	insecureTLS bool,
 ) (config.CLIConfig, error) {
 	jwtParser := jwt.NewParser(jwt.WithoutClaimsValidation())
-	var claims jwt.RegisteredClaims
+	var untrustedClaims jwt.RegisteredClaims
```
krancour, line 53: Changing the name of this symbol is technically unrelated to this change, but while I was already touching this file, I thought it worthwhile to make it clear that these claims aren't trusted (since the signature isn't verified).

jessesuen (approved):

Issue #482, opened 2023-07-21
some kind of sso support is needed


krancour:

krancour:
This has been adequately addressed by the above mentioned PRs.

PR #588, opened 2023-08-23
authorizing k8s client

krancour:
Part of #482

This PR solves a few different problems. It adds (and uses) a custom k8s client with the following benefits:

1. No matter how you authenticate (admin token, identity token, or bearer token from local kubeconfig), some kind of access review takes place to verify that you're permitted to perform the requested operation, _then_ a single underlying client with a cache is used to actually perform that operation. We get all the benefits of the cache _and_ the benefits of RBAC enforced as if we used an ad-hoc client built form the user's own creds.

1. The above means it is finally safe to enable client authn -- which was temporarily disabled up until now.

1. The custom client combines the familiar controller-runtime `client.Client` interface with a `Watch` function that is absent from that interface. It prevents us from having to use _both_ a controller-runtime `client.Client` and a client-go `dynamic.Interface`.

51e48c57d7a745cec0fe34b2ee67a2ff8062daa3 is the crux of this PR -- and even half of that change is just new test cases. The other commits just adapt to _using_ this new client.

Note: There is still no mapping from SSO user identity to a Kubernetes SA, so logging in with SSO basically gives you the same powers as admin for now.


internal/api/handler/watch_stages_v1alpha1.go
```diff
@@ -24,12 +24,9 @@ type WatchStageV1Alpha1Func func(
...
 
-func WatchStageV1Alpha1(
-	kubeCli client.Client,
-	dynamicCli dynamic.Interface,
-) WatchStageV1Alpha1Func {
-	validateProject := newProjectValidator(kubeCli)
-	stageCli := dynamicCli.Resource(kargov1alpha1.GroupVersion.WithResource("stages"))
+func WatchStageV1Alpha1(client kubernetes.Client) WatchStageV1Alpha1Func {
+	validateProject := newProjectValidator(client)
+	// stageCli := dynamicCli.Resource(kargov1alpha1.GroupVersion.WithResource("stages"))
```
gdsoumya, line 29: probably unwanted comment

gdsoumya (approved):
LGTM, small nitpicks

internal/api/handler/watch_stages_v1alpha1.go
krancour: Yup. Nice catch.

krancour: Fixed!

PR #589, opened 2023-08-23
controller fix: status updates have no effect

krancour:
This reverts #567 since all status updates seem to have no effect since then.

cc @alexmt... I'd obviously love to have the changes from #567 back if you know what may be wrong with it.


alexmt (approved):
Sure, will work on fix and send PR again

PR #591, opened 2023-08-24
Add Kubernetes resource manifest parser

devholic:
This commit adds parser to parse Kubernetes manifest file (yaml/json) to let CLI or API to extract Kubernetes objects from it.

### DIFF
--- a/api/v1alpha1/promotion_types.go
+++ b/api/v1alpha1/promotion_types.go
@@ -36,6 +36,10 @@ type Promotion struct {
 	Status PromotionStatus `json:"status,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


krancour:
I think I might prefer to see this functionality added somewhere besides the `yaml` package.

For one thing, this handles JSON as well as YAML. If that isn't reason enough, the stuff that's already in this package is much, _much_ lower-level and generic than what you're adding here.

This probably deserves its own package -- `manifests` maybe.

Then use of the package also looks very tidy -- `manifestParser := manifest.NewParser()` 😄

krancour (approved):
LGTM!

PR #592, opened 2023-08-24
Add handlers for create/update/delete resources with manifest

devholic:
#### Notes

- Part of https://github.com/akuity/kargo/issues/262
- #591 Should merged before this PR

---

This PR adds handlers for create/update/delete resources with manifest. By adding these handlers, we can allow users to create/update/delete resources without explicit permission on Kubernetes cluster.

### DIFF
--- a/charts/kargo/templates/api/cluster-role.yaml
+++ b/charts/kargo/templates/api/cluster-role.yaml
@@ -16,6 +16,8 @@ rules:
       - get
       - list
       - watch
+      - patch
+      - update
       - delete
   - apiGroups:
       - kargo.akuity.io


internal/api/handler/create_resource_v1alpha1.go
```diff
@@ -0,0 +1,75 @@
...
+		parsed, err := parseKubernetesManifest(req.Msg.GetManifest())
+		if err != nil {
+			return nil, connect.NewError(connect.CodeInvalidArgument, errors.Wrap(err, "parse manifest"))
+		}
+
+		var res []*svcv1alpha1.CreateResourceResult
+		for _, obj := range parsed {
+			if req.Msg.GetNamespace() != "" {
+				obj.SetNamespace(req.Msg.GetNamespace())
+			}
```
krancour, line 36: It seems here that `req.Msg.GetNamespace()` takes precedence over any value already set in the object. This behavior is different from that of `kubectl` and I think people might find that surprising.

I had to experiment to verify what `kubectl` actually does in this scenario.

`kubectl` surfaces an error for any resource that directly specifies a namespace that differs from the value set via the `--namespace` flag.

It looks like this:

```
the namespace from the provided object "x" does not match the namespace "bar". You must pass '--namespace=foo' to perform this operation.
```

imho, they're using the right approach -- when there's any ambiguity, don't make assumptions.

This applies to the other handlers as well.

internal/api/handler/create_resource_v1alpha1.go
```diff
@@ -0,0 +1,75 @@
...
+			if err := validateProject(ctx, obj.GetNamespace()); err != nil {
```
krancour, line 37: It seems conceivable to me -- or even likely -- that the manifests you receive could actually _contain_ the namespace/project resource. i.e. "Here's a chunk of YAML that is the project and everything in it."

`kubectl` (and `helm`) both have rules about the order in which resources are created.

I think it would be both onerous and unnecessary to replicate those rules in their entirety because I think there are only certain resources that we should support here. For instance -- no one should be using this to create `Service`, `Deployment`, or `Ingress` resources. (We're not trying to be a replacement for `kubectl`.)

So the order might be something like this:

1. Namespaces/projects
1. Secrets, Policies, Roles, RoleBindings
1. Stages
1. Promotions

internal/yaml/yaml.go
```diff
@@ -3,14 +3,49 @@ package yaml
...
 	"gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	kubeyaml "k8s.io/apimachinery/pkg/util/yaml"
 )
 
+type KubernetesManifestParserFunc func(data []byte) ([]*unstructured.Unstructured, error)
+
+func NewKubernetesManifestParser(scheme *runtime.Scheme) KubernetesManifestParserFunc {
```
krancour, line 21: I think I might prefer to see this somewhere besides the `yaml` package.

For one thing, this handles JSON as well as YAML. If that isn't reason enough, the stuff that's already in this package is much, _much_ lower-level and generic than what you're adding here.

This probably deserves it's own package -- `manifests` maybe.

Then use of the package also looks very tidy -- `manifestParser := manifest.NewParser()` 😄

internal/api/handler/create_resource_v1alpha1.go
krancour: And then that order probably needs to be reversed for deletes.

internal/api/handler/create_resource_v1alpha1.go
krancour: > I think we can remove support for overriding project (namespace) for now to make things simpler, WDYT?

I think that's sensible. 👍

internal/controller/stages/stages_test.go
```diff
@@ -722,29 +722,31 @@ func TestSync(t *testing.T) {
...
+				getAvailableStatesFromUpstreamStagesFn: tc.getAvailableStatesFromUpstreamStagesFn,
+			}
 			newStatus, err := reconciler.syncStage(context.Background(), testStage)
-			testCase.assertions(
-				testCase.initialStatus,
+			tc.assertions(
+				tc.initialStatus,
 				newStatus,
-				testCase.kargoClient,
+				tc.kargoClient,
```
krancour, line 749: I'm unclear about the role these changes play in this PR.

krancour (approved):
LGTM after merge conflicts are resolved.

PR #593, opened 2023-08-24
Add CLI commands for create/apply/delete resources with manifest

devholic:
Resolves https://github.com/akuity/kargo/issues/262

This PR adds CLI commands for create/apply/delete resources with manifest.

### DIFF
--- a/internal/controller/applications/applications.go
+++ b/internal/controller/applications/applications.go
@@ -7,10 +7,14 @@ import (
 	argocd "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 	"github.com/akuity/kargo/internal/controller"
@@ -37,6 +41,7 @@ func SetupReconcilerWithManager(
 	}
 	return ctrl.NewControllerManagedBy(argoMgr).
 		For(&argocd.Application{}).
+		WithEventFilter(AppHealthSyncStatusChangePredicate{}).
 		WithOptions(controller.CommonOptions()).
 		Complete(newReconciler(kargoMgr.GetClient()))
 }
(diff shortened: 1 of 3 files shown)


internal/cli/create/create.go
```diff
@@ -1,62 +1,126 @@
...
+			var createErrs []error
+			for _, filename := range flag.Filenames {
+				path := filename
+				if !filepath.IsAbs(path) {
+					path = filepath.Join(currentDir, path)
+				}
+				rawManifest, err := os.ReadFile(path)
+				if err != nil {
+					return pkgerrors.Wrap(err, fmt.Sprintf("read %q", filename))
+				}
```
krancour, line 78: fwiw, `kubectl` also allows `--file` to point at a directory. That doesn't seem to be possible here and I think that difference could be surprising.

internal/cli/create/create.go
```diff
@@ -1,62 +1,126 @@
...
+			kargoSvcCli, err := client.GetClientFromConfig(ctx, opt)
 			if err != nil {
-				return errors.Wrap(err, "new printer")
+				return pkgerrors.Wrap(err, "get client from config")
 			}
-			return printer.PrintObj(res, opt.IOStreams.Out)
+
+			var successRes []*kargosvcapi.CreateResourceResult_CreatedResourceManifest
+			var createErrs []error
+			for _, filename := range flag.Filenames {
```
krancour, line 70: I think looping through files and sending them one at a time might be problematic...

See [this comment](https://github.com/akuity/kargo/pull/592#discussion_r1304943410) on #592.

I think concern about proper order to create resources forces you to concatenate all the manifest files together and submit them as one.

internal/cli/option/flag.go
```diff
@@ -4,6 +4,12 @@ import "github.com/spf13/pflag"
 
 type FlagFn func(*pflag.FlagSet)
 
+func Filenames(v *[]string) FlagFn {
+	return func(fs *pflag.FlagSet) {
+		fs.StringSliceVarP(v, "filename", "f", nil, "Filename to use to create the resource")
```
krancour, line 9: Since, I believe, directories are supported now, and _I think_ URLs are also since you're re-using parts of `kubectl` for this, this might be more complete wording:

> Filename, directory, or URL to files to use to create the resource

☝️ Cribbed from `kubectl` as well

krancour: Also, I  notice this flag is used by all three of create, update, and delete. "Filename to use to create the resource" seems like the wrong message for update and delete sub-commands.

internal/cli/update/update.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+type Flags struct {
+	Filenames []string
+}
+
+func NewCommand(opt *option.Option) *cobra.Command {
+	var flag Flags
+	cmd := &cobra.Command{
+		Use:   "update [--project=project] -f (FILENAME)",
```
krancour, line 28: Wondering if the sub-command should perhaps be `apply` for improved consistency with kubectl -- since users are presumably already very familiar with that.

krancour: @devholic is apply different from update?

krancour (approved):

PR #594, opened 2023-08-24
fix ISS claim on admin tokens

krancour:
Up until now we've been using a hard coded value `kargo` in the `ISS` (issuer) claim of ID tokens issued for the admin account.

This is not technically a correct thing to do as it bears insufficient information about the provenance of the token and that matters when the token is presented back to the API server by the client. We use the `ISS` claim to decide which logic to apply to properly verify the token. (i.e. Verify it as an admin token, a token from an OIDC identity provider, or a SA token issued by Kubernetes.)

Making this change will prevent a Kargo API server from attempting to verify admin tokens clearly issued by _other_ Kargo API servers. (fwiw, that verification would fail anyway as long as both servers used different signing keys, but we probably should not take it for granted that 100% of all operators will use adequately unique signing keys. _Someone_ will use `key` or `password`.)

This PR updates the API server to use the configured protocol and hostname as the value of the `ISS` claim.

### DIFF
--- a/internal/api/config/config.go
+++ b/internal/api/config/config.go
@@ -15,7 +15,7 @@ import (
 type AdminConfig struct {
 	// HashedPassword is a bcrypt hash of the password for the admin account.
 	HashedPassword string `envconfig:"ADMIN_ACCOUNT_PASSWORD_HASH" required:"true"`
-	// TokenIssuer the value to be used in the ISS claim of ID tokens issued for
+	// TokenIssuer is the value to be used in the ISS claim of ID tokens issued for
 	// for the admin account.
 	TokenIssuer string `envconfig:"ADMIN_ACCOUNT_TOKEN_ISSUER" required:"true"`
 	// TokenAudience is the value to be used in the AUD claim of ID tokens issued
--- a/internal/api/config/config.go
+++ b/internal/api/config/config.go
@@ -15,8 +15,11 @@ import (
 type AdminConfig struct {
 	// HashedPassword is a bcrypt hash of the password for the admin account.
 	HashedPassword string `envconfig:"ADMIN_ACCOUNT_PASSWORD_HASH" required:"true"`
+	// TokenIssuer the value to be used in the ISS claim of ID tokens issued for
+	// for the admin account.
+	TokenIssuer string `envconfig:"ADMIN_ACCOUNT_TOKEN_ISSUER" required:"true"`
 	// TokenSigningKey is the key used to sign ID tokens for the admin account.
-	TokenSigningKey []byte `envconfig:"TOKEN_SIGNING_KEY" required:"true"`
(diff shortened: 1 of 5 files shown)


internal/api/handler/admin_login_v1alpha1.go
```diff
@@ -46,7 +46,7 @@ func AdminLoginV1Alpha1(cfg *config.AdminConfig) AdminLoginV1Alpha1Func {
 			jwt.SigningMethodHS256,
 			jwt.RegisteredClaims{
 				IssuedAt:  jwt.NewNumericDate(now),
-				Issuer:    "kargo",
+				Issuer:    cfg.TokenIssuer,
```
jessesuen, line 49: Not related to the PR, but I noticed we issue a 1-hour hard-wired expiration time.

While you are here, maybe consider making it configurable with an `ADMIN_ACCOUNT_TOKEN_EXPIRY` env var?

jessesuen (approved):
FWIW, having a hard-wired, well-known issuer like `kubernetes/serviceaccount` does come in handy in a migration case, or you are just changing the DNS name of your server.

But since this is specifically for human username/password admin login and not a service account token, I think this is fine.

If it were for other, long-lived service account token used by robots (and not humans), I think using a well-known issuer like `kargo` is acceptable because of the protection/assumption that there would be different signing keys between the instances. And the fact changing the hostname of kargo wouldn't suddenly result in invalidating all issued robot tokens.

internal/api/handler/admin_login_v1alpha1.go
krancour: I'll submit a follow-up.

internal/api/config/config.go
```diff
@@ -15,11 +15,14 @@ import (
 type AdminConfig struct {
 	// HashedPassword is a bcrypt hash of the password for the admin account.
 	HashedPassword string `envconfig:"ADMIN_ACCOUNT_PASSWORD_HASH" required:"true"`
+	// TokenIssuer the value to be used in the ISS claim of ID tokens issued for
```
krancour, line 18:
```suggestion
	// TokenIssuer is the value to be used in the ISS claim of ID tokens issued for
```

internal/api/config/config.go
```diff
@@ -15,11 +15,14 @@ import (
 type AdminConfig struct {
 	// HashedPassword is a bcrypt hash of the password for the admin account.
 	HashedPassword string `envconfig:"ADMIN_ACCOUNT_PASSWORD_HASH" required:"true"`
+	// TokenIssuer is the value to be used in the ISS claim of ID tokens issued for
+	// for the admin account.
```
krancour, line 19:
```suggestion
	// the admin account.
```

PR #598, opened 2023-08-25
fix(controller): use patch to make stage/promotion status changes to avoid conflict errors

jessesuen:
Reverts https://github.com/akuity/kargo/pull/589, which reverted  https://github.com/akuity/kargo/pull/567.

But additionally adds `patch` privileges to `kargo-controller` 

```yaml
- apiGroups:
  - kargo.akuity.io
  resources:
  - stages/status
  - promotions/status
  verbs:
  - update
  - patch  # <<< added
```

### DIFF
--- a/api/v1alpha1/promotion_types.go
+++ b/api/v1alpha1/promotion_types.go
@@ -36,6 +36,10 @@ type Promotion struct {
 	Status PromotionStatus `json:"status,omitempty"`
 }
 
+func (p *Promotion) GetStatus() *PromotionStatus {
+	return &p.Status
+}
+
 // PromotionSpec describes the desired transition of a specific Stage into a
 // specific StageState.
 type PromotionSpec struct {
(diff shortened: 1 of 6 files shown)


krancour:
@jessesuen 

> But additionally adds patch privileges to kargo-controller

Is that what the problem was? That never showed up in the logs.

jessesuen:
Yes, that was the problem. It showed up for me like:

```
time="2023-08-25T03:55:15Z" level=error msg="error updating Stage status: stages.kargo.akuity.io \"test\" is forbidden: User \"system:serviceaccount:kargo:kargo-controller\" cannot patch resource \"stages/status\" in API group \"kargo.akuity.io\" in the namespace \"kargo-guestbook\"" namespace=kargo-guestbook stage=test
```

I'm running an image from this PR in my environment and the status is getting updated properly.

alexmt (approved):
I was just about to start working on it. Thank you @jessesuen !

PR #601, opened 2023-08-28
chart: simplify secret --> env var pattern

### LINKED ISSUE #597: Use `envFrom.secretRef` instead of `env[].valueFrom.secretKeyRef`
I noticed we are repeating a pattern that I regret in Argo CD. That is, we have individual env vars, with `secretKeyRef` to specific keys in the same `kargo-api` secret:

```yaml
            env:
            - name: ADMIN_ACCOUNT_PASSWORD_HASH
              valueFrom:
                secretKeyRef:
                  name: kargo-api
                  key: adminPassword
            - name: ADMIN_ACCOUNT_TOKEN_SIGNING_KEY
              valueFrom:
                secretKeyRef:
                  name: kargo-api
                  key: tokenSigningKey
```

We used the same technique in Argo CD, and I really wish we hadn't because:
1. The key name in the secret, was different than the env var (we would use dot notation like `admin.enabled`). The same is happening here with `tokenSigningKey` and `ADMIN_ACCOUNT_TOKEN_SIGNING_KEY`.
2. It resulted in a ton more YAML than necessary in the Deployment.

I think we should do this instead:

```yaml
          envFrom:
          - configMapRef:
              name: kargo-api
          - secretRef:
              name: kargo-api
```

And the Secret/ConfigMap itself would have env vars as the keys. e.g.:

```yaml
data:
  ADMIN_ACCOUNT_PASSWORD_HASH: xxxxx
  ADMIN_ACCOUNT_TOKEN_SIGNING_KEY: yyyyy
```

The benefits I see are:
* There is only one well-known documented "key" to adjust. Not both `tokenSigningKey` and `ADMIN_ACCOUNT_TOKEN_SIGNING_KEY` and making sure `tokenSigningKey` is properly declared and referenced from the Deployment.
* It reduces the amount of YAML necessary in the Deployment.

krancour:
Fixes #597

### DIFF
--- a/charts/kargo/templates/api/deployment.yaml
+++ b/charts/kargo/templates/api/deployment.yaml
@@ -21,9 +21,8 @@ spec:
         {{- include "kargo.selectorLabels" . | nindent 8 }}
         {{- include "kargo.api.labels" . | nindent 8 }}
       annotations:
-        {{- if .Values.api.adminAccount.enabled }}
-        secret-checksum/kargo-api: {{ include (print $.Template.BasePath "/api/secret.yaml") . | sha256sum }}
-        {{- end }}
+        configmap/checksum: {{ include (print $.Template.BasePath "/api/configmap.yaml") . | sha256sum }}
+        secret/checksum: {{ include (print $.Template.BasePath "/api/secret.yaml") . | sha256sum }}
     spec:
       serviceAccount: kargo-api
       containers:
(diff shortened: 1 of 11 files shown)


charts/kargo/templates/api/deployment.yaml
```diff
@@ -45,16 +45,6 @@ spec:
...
-                secretKeyRef:
-                  name: kargo-api
-                  key: adminPassword
-            - name: ADMIN_ACCOUNT_TOKEN_SIGNING_KEY
-              valueFrom:
-                secretKeyRef:
-                  name: kargo-api
-                  key: tokenSigningKey
             - name: ADMIN_ACCOUNT_TOKEN_TTL
               value: {{ .Values.api.adminAccount.tokenTTL }}
```
jessesuen, line 49: So I know that I only mentioned Secret in the original issue, but I realize now I also meant to extend this to our non-sentive settings. WDYT about doing the same thing with envFrom but having our config centralized in a `kargo-api` ConfigMap?

charts/kargo/templates/api/deployment.yaml
```diff
@@ -78,6 +68,11 @@ spec:
             - name: OIDC_CLIENT_ID
               value: {{ .Values.api.oidc.clientID }}
 {{- end }}
+{{- end }}
+{{- if .Values.api.adminAccount.enabled }}
+          envFrom:
+            - secretRef:
+                name: kargo-api
 {{- end }}
```
jessesuen, line 76: I think we should unconditionally have the kargo-api secret but have our if condition `{{- if .Values.api.adminAccount.enabled }}` in the Secret instead.

charts/kargo/templates/api/deployment.yaml
krancour: Ah... I didn't misunderstand it as only applying to secrets... but I definitely did misunderstand the idea that _all_ the logic for what env vars are defined / not defined / and how they're defined would all move to the secrets/configmaps.

That makes a lot of sense.

Will fix.

krancour: Ok... I've made the change more encompassing. We don't use `env:` anywhere now. We only use `envFrom:` and reference applicable secret and/or configmap.

jessesuen (approved):
This looks much cleaner!

PR #605, opened 2023-08-29
feat: add ability to refresh a stage. add status.observedGeneration

jessesuen:
1. This introduces a new endpoint to "refresh" a stage, which basically means setting `metadata.annotations["kargo.akuity.io/refresh"]` to the current RFC3339 timestamp in UTC. The controller will clear this annotation (if it is set) during reconciliation. The presence of this annotation will give the UI the ability to visually indicate if a Stage is currently being reconciled by the controller (e.g. show a spinner). I also updated the applications controller to use this annotation.

2. This also introduces `status.observedGeneration`, which is updated to `metadata.generation` during every reconciliation. observedGeneration is a standard practice in Kubernetes controllers to allow API clients to know if the status in a controller can be trusted (e.g. did the controller reconcile the current spec of this Stage?). API clients who depend on the accuracy of `.status` are expected to always compare if `metadata.generation == status.observedGeneration` before trusting the contents of .status. 

Note that the Kubernetes API server will only bump `metadata.generation` when there is a modification to `spec`, but will **_not_** bump generation if only `metadata.annotations` alone is modified.


gdsoumya (approved):
LGTM

krancour:
@jessesuen I merged #579 this morning. There were a lot of conflicts to resolve there, and there ended up being a lot to resolve here as well...

The fastest way to fix it turned out to be starting fresh from main and cherry-picking the one commit that mattered from this branch.

Everything should be set now, but letting you know in case you want to double-check anything.

jessesuen:
Thanks @krancour . Rebased the change ontop of main.

krancour (approved):

PR #606, opened 2023-08-29
Add Get/List promotion APIs

devholic:
This PR adds Get/List Promotion APIs to show users Promotion details

### DIFF
--- a/cmd/controlplane/api.go
+++ b/cmd/controlplane/api.go
@@ -40,7 +40,7 @@ func newAPICommand() *cobra.Command {
 			if err != nil {
 				return pkgerrors.Wrap(err, "error loading REST config")
 			}
-			kubeCli, err := kubernetes.NewClient(ctx, restCfg, kubernetes.ClientOptions{
+			kubeClient, err := kubernetes.NewClient(ctx, restCfg, kubernetes.ClientOptions{
 				NewInternalClient: newClientForAPI,
 			})
 			if err != nil {
@@ -60,7 +60,7 @@ func newAPICommand() *cobra.Command {
 				}).Info("SSO via OpenID Connect is enabled")
 			}
 
-			srv, err := api.NewServer(cfg, kubeCli)
+			srv, err := api.NewServer(cfg, kubeClient)
 			if err != nil {
 				return pkgerrors.Wrap(err, "error creating API server")
 			}


cmd/controlplane/api.go
```diff
@@ -34,7 +40,9 @@ func newAPICommand() *cobra.Command {
 			if err != nil {
 				return pkgerrors.Wrap(err, "error loading REST config")
 			}
-			client, err := kubernetes.NewClient(ctx, restCfg, kubernetes.ClientOptions{})
+			kubeCli, err := kubernetes.NewClient(ctx, restCfg, kubernetes.ClientOptions{
```
krancour, line 43: Nit: Ending clients with "cli" can be confusing because CLI means something.

internal/controller/stages/stages_test.go
```diff
@@ -722,29 +722,31 @@ func TestSync(t *testing.T) {
...
+				getAvailableStatesFromUpstreamStagesFn: tc.getAvailableStatesFromUpstreamStagesFn,
+			}
 			newStatus, err := reconciler.syncStage(context.Background(), testStage)
-			testCase.assertions(
-				testCase.initialStatus,
+			tc.assertions(
+				tc.initialStatus,
 				newStatus,
-				testCase.kargoClient,
+				tc.kargoClient,
```
krancour, line 749: I'm not clear on the role of these changes in this PR.

internal/kubeclient/indexer.go
```diff
@@ -54,22 +63,40 @@ func indexStagesByArgoCDApplications(shardName string) client.IndexerFunc {
...
-	case api.PromotionPhaseComplete, api.PromotionPhaseFailed:
-		return nil
+// IndexOutstandingPromotionsByStage creates index for Promotions in non-terminal states by Stage
+func IndexOutstandingPromotionsByStage(ctx context.Context, mgr ctrl.Manager) error {
+	return IndexPromotionsByStage(ctx, mgr, filterNonOutstandingPromotionPhases)
+}
+
+func filterNonOutstandingPromotionPhases(promo *api.Promotion) bool {
+	return !slices.Contains(NonOutstandingPromotionPhases, promo.GetStatus().Phase)
+}
```
krancour, line 83: Would it possibly make sense to just build two separate indices?

krancour: Please disregard my comment. I (too quickly) read the code incorrectly.

PR #609, opened 2023-08-29
feat(ui): Add refresh Stage button

rpelczar:
image


jessesuen:
Can we get this to disable the refresh button and show a spinner as long as `kargo.akuity.io/refresh` is present on the stage?

However, I would only do this if we are also watching the Stage so that the spinner could automatically stop spinning once the `kargo.akuity.io/refresh` annotation disappears.

rpelczar:
> Can we get this to disable the refresh button and show a spinner as long as `kargo.akuity.io/refresh` is present on the stage?
> 
> However, I would only do this if we are also watching the Stage so that the spinner could automatically stop spinning once the `kargo.akuity.io/refresh` annotation disappears.

@jessesuen done

jessesuen (approved):

jessesuen:
Thank you!

PR #610, opened 2023-08-30
feat: support wildcards in kargo.akuity.io/authorized-stage annotation

jessesuen:
I found it a bit tedious to exactly match the authorized stage by the exact name. I would like to make the UX easier so that users can allow all stages in a project using a glob wildcard. e.g.:

```yaml
metadata:
  annotations:
    kargo.akuity.io/authorized-stage: kargo-demo:*
```

Or even:
```yaml
metadata:
  annotations:
    kargo.akuity.io/authorized-stage: '*:*'
```

### DIFF
--- a/internal/controller/promotions/argocd.go
+++ b/internal/controller/promotions/argocd.go
@@ -148,11 +148,21 @@ func (r *reconciler) authorizeArgoCDAppUpdate(
 	}
 	allowedNamespaceGlob, err := glob.Compile(tokens[0])
 	if err != nil {
-		return permErr
+		return errors.Errorf(
+			"Argo CD Application %q in namespace %q has invalid glob expression: %q",
+			app.Name,
+			app.Namespace,
+			tokens[0],
+		)
 	}
 	allowedNameGlob, err := glob.Compile(tokens[1])
 	if err != nil {
-		return permErr
+		return errors.Errorf(
+			"Argo CD Application %q in namespace %q has invalid glob expression: %q",
+			app.Name,
+			app.Namespace,
+			tokens[1],
+		)
 	}
 	if !allowedNamespaceGlob.Match(stageMeta.Namespace) || !allowedNameGlob.Match(stageMeta.Name) {
 		return permErr
--- a/internal/controller/promotions/argocd.go
+++ b/internal/controller/promotions/argocd.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	argocd "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
+	"github.com/gobwas/glob"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


krancour:
> I found it a bit tedious to exactly match the authorized stage by the exact name.

I agree it's tedious, but I have seen that tedium as a feature rather than a flaw...

> I would like to make the UX easier so that users can allow all stages in a project using a glob wildcard.

Logically, only _one_ Stage can be responsible for any updates with a given Application.

Requiring this to be set exactly right can probably _prevent_ some mistakes that the wildcards would enable.

jessesuen:
> I have seen it as a feature (sanity check) rather than a flaw.

I disagree that it's a feature to disallow flexibility in how to configure RBAC. That's like saying kube RBAC should not allow wildcards in Roles.

```yaml
- apiGroups:
  - apps
  resources:
  - get
  verbs:
  - '*'
```

We're not precluding anyone from using explicit references. This is just adding the ability when explicit names get in the way. In my case, I'm using an AppSet to generate my apps:

```
kind: ApplicationSet
metadata:
  name: guestbook
spec:
  generators:
  - git:
      repoURL: https://github.com/jessesuen/guestbook-deploy.git
      revision: HEAD
      directories:
      - path: env/*
  template:
    metadata:
      name: guestbook-{{path.basename}}
      annotations:
        kargo.akuity.io/authorized-stage: XXXXXX # this is hard to coordinate 
    spec:
      destination:
        namespace: guestbook-{{path.basename}}
        server: https://kubernetes.default.svc
      project: guestbook
      source:
        path: '{{path}}'
        repoURL: https://github.com/jessesuen/guestbook-deploy.git
        targetRevision: HEAD
```

And so I don't have a convenient way to make sure the annotation matches the stage unless I rely on a naming convention in how I lay out my files in the repo.  So the lack of this feature makes it annoying to use.

krancour:
> if it does not compromise security

To be very clear, my objection wasn't based on security concerns, but rather on it opening the door to undefined behavior. What happens when more than one Stage tries to manage a single Application? (Mayhem.)

jessesuen:
> What happens when more than one Stage tries to manage a single Application? (Mayhem.)

Actually, why shouldn't that be supported in the future? IMO, there might be multiple stages in a multi-stage production deployment for the same Argo CD application. For example:

The same production argo cd app, might be broken down to multiple stages to deploy
* `prod-stage-1`: Set feature flag for 10% users
* `prod-stage-2`: Update feature flag for 50% user 
* `prod-stage-3`: Full rollout of feature flag

We would have to think about what it means to "promote" something like I just described, but in all cases, I want the same Argo CD application to be refreshed and synced.

But all of this is beside the point. I still have a problem of: I was unable to use generated Apps from an AppSet set because the annotation wasn't possible to coordinate.

In order to get around this limitation, I had to change my behavior and name all of my directories exactly the same as the stages. Our users shouldn't make the same compromise. Having a requirement to reference a single stage is necessarily limiting for a reason I think will not hold true in the future (multi-stage, single app syncs).

internal/controller/promotions/argocd.go
```diff
@@ -145,8 +146,15 @@ func (r *reconciler) authorizeArgoCDAppUpdate(
 			app.Namespace,
 		)
 	}
-	allowedNamespace, allowedName := tokens[0], tokens[1]
-	if stageMeta.Namespace != allowedNamespace || stageMeta.Name != allowedName {
+	allowedNamespaceGlob, err := glob.Compile(tokens[0])
+	if err != nil {
+		return permErr
```
krancour, line 151: Nit: While it's safe to say mutation of the App isn't authorized in this case, the underlying error here is a failure to compile the glob expression and it's being completely suppressed.

imho, we should either wrap and return the glob compilation error, or if we still want to return `permErr`, then at least log the compilation error.

Same for the name glob.

jessesuen: I agree

krancour:
> Actually, why shouldn't that be supported in the future? IMO, there might be multiple stages in a multi-stage production deployment for the same Argo CD application. For example...

Definitely a good example. 👍 

Just making sure we understand that multiple Stages managing one App is _currently_ problematic.

> But all of this is beside the point. I still have a problem of: I was unable to use generated Apps from an AppSet set because the annotation wasn't possible to coordinate.

Understood and agreed.

jessesuen:
>  should either wrap and return the glob compilation error

I surface the glob error now. PTAL.

krancour (approved):
LGTM

PR #612, opened 2023-08-30
feat: add PromoteSubscribers API/CLI

jessesuen:
1. Add PromoteSubscribers API/CLI, which accepts a Stage and one of its historical freights. It then identifies all subscribers of that Stage and creates Promotion objects for each of them.

CLI usage:
```
$ kargo stage promote-subscribers kargo-guestbook staging --freight=c353927ca7af42b38c0cdcfa393b2c552740e547
Promotion Created: "prod-use1.01h92dysvxkx7nmhptkhgtsd28.c353927"
Promotion Created: "prod-usw1.01h92dyswv62mr2jx8rrsn4z40.c353927"
Promotion Created: "prod-euw1.01h92dysxe2y02mqzhxk9ehrq6.c353927"
```

2. Promotions objects are now consistently named with the following convention (previously, promotions names were different between auto-promote vs. API promote)
```
<stagename>.<ulid>.<7-character-freight-id>
```

A [ULID](https://github.com/ulid/spec) is basically a UUID, but uses less characters, and lexographically sorted so that when listing, the most recent will appear in the last. 

Example:
```
$ k get promos  -w
NAME                                           STAGE       STATE                                      PHASE       AGE
dev.01h92e6wcfws9x3pjeg0dsva0d.f08b2e7         dev         f08b2e72c9b2b7b263da6d55f9536e49b5ce972c               8m43s
prod-euw1.01h92dysxe2y02mqzhxk9ehrq6.c353927   prod-euw1   c353927ca7af42b38c0cdcfa393b2c552740e547   Completed   13m
prod-use1.01h92dysvxkx7nmhptkhgtsd28.c353927   prod-use1   c353927ca7af42b38c0cdcfa393b2c552740e547   Completed   13m
prod-usw1.01h92dyswv62mr2jx8rrsn4z40.c353927   prod-usw1   c353927ca7af42b38c0cdcfa393b2c552740e547   Completed   13m
```

3. Eliminated some repeated copy-and-paste code in API handlers with a helper


jessesuen:
I am missing unit tests for this, but please start on the review while I work on that.

internal/api/promote_subscribers_v1alpha1.go
```diff
@@ -0,0 +1,90 @@
...
+	if err := s.validateProject(ctx, req.Msg.GetProject()); err != nil {
+		return nil, err
+	}
+	stage, err := getStage(ctx, s.client, req.Msg.GetProject(), req.Msg.GetStage())
+	if err != nil {
+		return nil, err
+	}
+	if err = validateFreightExists(req.Msg.GetFreight(), stage.Status.History); err != nil {
+		return nil, err
+	}
```
krancour, line 35: For a _regular_ promotion, the target state must be among the _available_ states. Only healthy states from upstream flow downstream, so for a _regular_ promotion, the target state is always a healthy one.

For this case where we're pushing state instead of pulling, I wonder if there should be some check to avoid promotion of a state that's known to be no good.

internal/api/promote_subscribers_v1alpha1.go
```diff
@@ -0,0 +1,90 @@
...
+
+	subscribers, err := s.findStageSubscribers(ctx, stage)
+	if err != nil {
+		return nil, err
+	}
+	if len(subscribers) == 0 {
+		return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("Stage %q has no subscribers", req.Msg.GetStage()))
+	}
```
krancour, line 43: I have no strong opinion here, but I at last wondered if this should just be an implicit no-op instead of an error. I'm ok with either, but thought it was worth bringing up.

internal/api/promote_subscribers_v1alpha1.go
```diff
@@ -0,0 +1,90 @@
...
+
+	var promoteErrs []error
+	var createdPromos []*v1alpha1.Promotion
+	for _, subscriber := range subscribers {
+		if err := validateFreightExists(req.Msg.GetFreight(), subscriber.Status.AvailableStates); err != nil {
+			// TODO(JS): currently we create promotions to all of this Stage's subscribers, ignoring
+			// whether or not the freight *also* appears in the availableStates of the subscriber.
+			// Normally, it should always be the case that if it's in our history, it should also
+			// appear in our subscriber's availableStates. For now, just log a warning if we are
+			// promoting something that for some reason, has not yet appeared there.
```
krancour, line 55: Ah. This partially answers my question [here](https://github.com/akuity/kargo/pull/612/files#r1313457336)

internal/api/promote_subscribers_v1alpha1.go
```diff
@@ -0,0 +1,90 @@
...
+			continue
+		}
+		for _, upstream := range s.Spec.Subscriptions.UpstreamStages {
+			if upstream.Name != stage.Name {
+				continue
+			}
+			subscribers = append(subscribers, s)
+		}
+	}
+	return subscribers, nil
```
krancour, line 89: Would it be worth building an index for this instead of iterating over all Stages?

We don't have to do that now, but maybe we could at least leave a TODO comment about that?

internal/api/refresh_stage_v1alpha1_test.go
```diff
@@ -108,13 +108,13 @@ func TestRefreshStage(t *testing.T) {
...
-			// Make sure we set timestamp is close to now
-			// Assume it doesn't take 3 seconds to run this unit test.
 			stage := res.Msg.GetStage()
 			annotation := stage.Metadata.Annotations[kubev1alpha1.AnnotationKeyRefresh]
 			refreshTime, err := time.Parse(time.RFC3339, annotation)
 			require.NoError(t, err)
-			require.LessOrEqual(t, time.Now().UTC().Sub(refreshTime), 3*time.Second)
+			// Make sure we set timestamp is close to now
+			// Assume it doesn't take 3 seconds to run this unit test.
+			require.WithinDuration(t, time.Now(), refreshTime, 3*time.Second)
```
krancour, line 117: Somehow I didn't know about `require.WithinDuration()`. That is super useful!

internal/kargo/kargo.go
```diff
@@ -0,0 +1,56 @@
...
+	// We just want a unique ID that can be sorted lexicographically
+	promoName := strings.ToLower(fmt.Sprintf("%s.%s.%s", shortStageName, ulid.Make(), shortHash))
+
+	ownerRef := metav1.NewControllerRef(&stage, kubev1alpha1.GroupVersion.WithKind("Stage"))
+
+	promotion := kubev1alpha1.Promotion{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            promoName,
+			Namespace:       stage.Namespace,
+			OwnerReferences: []metav1.OwnerReference{*ownerRef},
```
krancour, line 43: This is a nice addition.

krancour (approved):
Had a few questions and comments, but there is nothing blocking my LGTM...

LGTM

internal/api/promote_subscribers_v1alpha1.go
jessesuen: It's a good idea. I will add a TODO

internal/api/refresh_stage_v1alpha1_test.go
jessesuen: Me too! Just learned about it myself and wanted to make my other test more readable.

internal/api/promote_subscribers_v1alpha1.go
jessesuen: It's a good call out. I actually didn't have this check before, but I added it last night minute because I felt it was more than likely a user error if they were promoting something to subscribers and there was none. We might need to introduce a flag to allow this, if the use case does pops up (I can't think of one).

internal/api/promote_subscribers_v1alpha1.go
jessesuen: I didn't actually realize there might be freight in freight history that was not considered healthy. I think the suggestion makes sense.

internal/api/promote_subscribers_v1alpha1.go
krancour: Not that it stops us from figuring out the right behavior here, but maybe the UI doesn't need to show the promote to subscribers button if there are no subscribers?

krancour (approved):

internal/api/promote_subscribers_v1alpha1.go
jessesuen: I agree UI could do a lot to make this mostly a non-issue. I'm happy to remove this if you want. I also don't have a strong opinion

internal/api/promote_subscribers_v1alpha1.go
jessesuen: Added health check.

PR #615, opened 2023-08-30
Merge scattered direct/indirect dependency groups

devholic:
To identify module changes clearly, this PR merges scattered direct/indirect dependency groups.


aborilov (approved):

jessesuen:
How did this get reordered? Manually? If so, is that sustainable? Usually I add to `go.mod` just using `go get` and don't pay attention to how it got modified.

Issue #614, opened 2023-08-30
Kargo doesn't support simultaneous Dex CLI and Web based SSO login well


jessesuen:
I tried configuring Kargo to handle web and CLI-based SSO login with Dex, but did not have a good time.

Here was my original attempt with dex (take note of `redirectURIs`):

```yaml
connectors:
- config:
    clientID: $CONNECTOR_GITHUB_CLIENT_ID
    clientSecret: $CONNECTOR_GITHUB_CLIENT_SECRET
    orgs:
    - name: akuity
    redirectURI: https://dex.demo.akuity.io/callback
    teamNameField: slug
    useLoginAsID: true
  id: github
  name: GitHub
  type: github
issuer: https://dex.demo.akuity.io
oauth2:
  skipApprovalScreen: true
staticClients:
- id: kargo
  name: Kargo
  public: true
  redirectURIs:
  - https://kargo.demo.akuity.io/login
  - http://localhost
  - http://localhost/auth/callback
storage:
  config:
    inCluster: true
  type: kubernetes
web:
  allowedOrigins:
  - https://kargo.demo.akuity.io
```

The above will work for web login, but fail CLI login with the following error:

```
Bad Request
Unregistered redirect_uri ("http://localhost:57289/auth/callback").
```

The problem is that the port is included as part of the URI, and so the port must be included in dex's config.yaml and exactly match one of the redirectURIs **including the port number**. What makes matters worse, is that `kargo login` will use a random port by default. And so the only way this could work is if the dex admin chose a hard-wired port to register to dex (e.g. 1234). And then that admin must instruct all of their kargo users to run the `kargo login` command with the `--port 1234` flag that they chose to register to dex.

We can do better.

Looking at Dex's `validateRedirectURI` logic: https://github.com/dexidp/dex/blob/70d7a2c7c1bb2646b1a540e49616cbc39622fb83/server/oauth2.go#L632-L636

We see that Dex does allow for randomized ports for CLI login. The catch is that
1. The static client's `redirectURIs` list must be empty
2. The static client must be public (ours already is).
3. The URI being redirected to must be `http://localhost/<subpath>`

So, in other words, this would work:

```yaml
  staticClients:
  - id: kargo
    name: Kargo
    redirectURIs:
    - https://kargo.demo.akuity.io/login
    public: true
  - id: kargo-cli
    name: Kargo CLI
    public: true
```

[...]

PR #617, opened 2023-08-30
feat: support CLI specific OIDC client (separate from web)

jessesuen:
Resolves https://github.com/akuity/kargo/issues/614

1. Introduces a new/optional `OIDC_CLI_CLIENT_ID` env var configuration, which is what the CLI will use
2. The `OIDC_CLI_CLIENT_ID` will be published in Kargo's public endpoint, so that the CLI will know to use it when performing the OAuth2 exchange
3. Finally, when verifying tokens, Kargo will iterate both the web OIDC_CLIENT_ID verifier as well as the CLI OIDC_CLI_CLIENT_ID.


krancour (dismissed):
See [comment](https://github.com/akuity/kargo/issues/614#issuecomment-1699145274) on the related issue.

jessesuen:
> See https://github.com/akuity/kargo/issues/614#issuecomment-1699145274 on the related issue.

Please see response. The suggestion to use a single single static clients with public mode doesn't work.

internal/api/option/auth.go
```diff
@@ -44,44 +44,79 @@ type authInterceptor struct {
 		ctx context.Context,
 		rawToken string,
 	) (string, []string, bool)
-	oidcTokenVerifyFn   func(context.Context, string) (*oidc.IDToken, error)
+	oidcTokenVerifyFn   goOIDCIDTokenVerifyFn
 	oidcExtractGroupsFn func(*oidc.IDToken) ([]string, error)
 }
 
+// goOIDCIDTokenVerifyFn is a github.com/coreos/go-oidc/v3/oidc/IDTokenVerifier.Verify() function
+type goOIDCIDTokenVerifyFn func(ctx context.Context, rawIDToken string) (*oidc.IDToken, error)
```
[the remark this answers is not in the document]
jessesuen: I wanted to make sure it was understood that it was a function definition from the upstream `github.com/coreos/go-oidc`, so I intentionally called it `goOIDC`. I'd like to keep it for clarity.

krancour:
@jessesuen I see now why this is needed.

This mostly LGTM.

I think the thing that may be missing from this PR is it doesn't automatically give the same treatment to the included Dex when that is enabled.

jessesuen:
> I think the thing that may be missing from this PR is it doesn't automatically give the same treatment to the included Dex when that is enabled.

Good catch. I can add it.

charts/kargo/values.yaml
```diff
@@ -150,6 +150,12 @@ api:
     ## you by your identity provider.
     clientID: <client ID>
 
+    ## The client ID for the OIDC client.
+    ## 
+    # Some OIDC providers (such as Dex) require a separate clientID for different
+    # callback URLs such as the web app vs. CLI http://localhost login.
```
krancour, line 156: As with the regular `clientID` field above, could we add a note here that this field is ignored if the built-in Dex is enabled -- in which case it will be automatically configured?

krancour: Non-blocking.

krancour (approved):
One non-blocking nit. LGTM

Issue #623, opened 2023-08-30
Stage/Freight Health is not always applicable


jessesuen:
If there's no way/reason to assess health (e.g. no associated Argo CD App), we should simply omit health entirely instead of populating the state with `Unknown`. Another use case when I'm using a Stage [for control-flow purposes](https://github.com/akuity/kargo/issues/622), health doesn't make sense either. Since `Unknown` is generally considered an "bad" state. It would be better not to show anything at all when it is not applicable.

Currently we **_require_** an Argo CD health status in a state, in order for that state to be made available to subscribers. Which means Kargo has a hard requirement with Argo CD.

I think there are two things needed:
1. controller should ~not populate Unknown~ populate `N/A` if there is nothing to assess (e.g. no Argo CD app). The state should appear in the history and as an availableState to subscribers
2. UI ~can't assume health is always there~ should not display any health if it is `N/A`

krancour:
Note to self: NA is a kind of pseudo-healthy.

krancour:
I will address this in the course of completing work on #205.

jessesuen:
@krancour is this fixed by https://github.com/akuity/kargo/pull/685?

krancour:
@jessesuen if you're happy with nil Health as opposed to the N/A state that was originally proposed, then yes.

jessesuen:
Yep I think nil is actually my preference especially since health is now more of a reflection of Argo CD app health..

PR #626, opened 2023-08-31
feat(ui): Add Healthy status icons to DAG view

rpelczar:
image


jessesuen:
Not sure if you saw https://github.com/akuity/kargo/issues/623, but Stage health will not always be applicable, and will be missing entirely in some stages in the future. Will you need to change anything to accommodate?

rpelczar:
@jessesuen I saw this but wanted to implement this case when the API is ready. (I'm not sure if the response will change)

So, if currentState/health is empty, we shouldn't show anything on the node component, right?

jessesuen:
> So, if currentState/health is empty, we shouldn't show anything on the node component, right?

Correct we shouldn't show anything. Based on this morning's conversation, the current thinking is health will be "not applicable" instead of empty. And then the UI can choose not to show anything.

jessesuen (approved):

Issue #625, opened 2023-08-30
Get Freight / Timeline API


jessesuen:
In order to power the timeline (aka freight-line) view, we need GetFreight API endpoint that will iterate all stages, iterate all frieght in those stages, and return an ordered list of freight to be presented in UI.

PR #629, opened 2023-09-01
feat: Add QueryFreight API

jessesuen:
Resolves https://github.com/akuity/kargo/issues/625

This PR introduces the QueryFreight API that is needed to power the timeline (aka freight-line) view. The term Query is used instead of List, because this API supports `group_by` and `order_by` and returns a nested structure instead of a simple list.

Here is an example query:
```json
{"project":"kargo-guestbook","group_by":"container_repo"}
```

And the return value:
```json
{
    "groups": {
        "ghcr.io/jessesuen/guestbook": {
            "freights": [
                {
                    "id": "f08b2e72c9b2b7b263da6d55f9536e49b5ce972c",
                    "first_seen": "2023-08-25T06:04:19Z",
                    "provenance": "",
                    "commits": [],
                    "images": [
                        {
                            "repo_url": "ghcr.io/jessesuen/guestbook",
                            "tag": "v0.0.4"
                        }
                    ],
                    "charts": []
                },
                {
                    "id": "c353927ca7af42b38c0cdcfa393b2c552740e547",
                    "first_seen": "2023-08-28T15:18:57Z",
                    "provenance": "",
                    "commits": [],
                    "images": [
                        {
                            "repo_url": "ghcr.io/jessesuen/guestbook",
                            "tag": "v0.0.5"
                        }
                    ],
                    "charts": []
                },
                {
                    "id": "6075e1ab632d6ed194b6ecda069998c46702b6ea",
                    "first_seen": "2023-08-28T15:23:57Z",
                    "provenance": "",
                    "commits": [],
                    "images": [
                        {
                            "repo_url": "ghcr.io/jessesuen/guestbook",
                            "tag": "v0.0.6"
                        }
                    ],
                    "charts": []
                }
            ]
        }
    }
}
```

NOTE: protobuf does not provide a way to support something like:
```proto
message QueryFreightsResponse {
  map<string, repeated github.com.akuity.kargo.pkg.api.v1alpha1.StageState> groups = 1;
}
```

And so the return value is more deeply nested than I would have liked.


jessesuen:
Another example:

Request:
```json
{
    "project": "kargo-guestbook",
    "stage": "prod-euw1"
}
```

Result:
```json
{
    "groups": {
        "": {
            "freights": [
                {
                    "id": "c353927ca7af42b38c0cdcfa393b2c552740e547",
                    "first_seen": "2023-08-28T15:18:57Z",
                    "provenance": "",
                    "commits": [],
                    "images": [
                        {
                            "repo_url": "ghcr.io/jessesuen/guestbook",
                            "tag": "v0.0.5"
                        }
                    ],
                    "charts": []
                }
            ]
        }
    }
}
```

internal/api/query_freights_v1alpha1.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, err
+	}
+
+	var stages []kubev1alpha1.Stage
+	if req.Msg.GetStage() != "" {
+		stage, err := getStage(ctx, s.client, req.Msg.GetProject(), req.Msg.GetStage())
+		if err != nil {
+			return nil, err
+		}
+		stages = append(stages, *stage)
```
krancour, line 49: Super small nit: You don't need to append here. You can just `stages = []kubev1alpha1.Stage{*stage}`.

internal/api/query_freights_v1alpha1.go
```diff
@@ -0,0 +1,144 @@
...
+	} else {
+		var list kubev1alpha1.StageList
+		if err := s.client.List(ctx, &list, client.InNamespace(req.Msg.GetProject())); err != nil {
+			return nil, connect.NewError(connect.CodeInternal, err)
+		}
+		stages = list.Items
+	}
+
+	seen := make(map[string]kubev1alpha1.StageState)
```
krancour, line 58: Nit: I know some people seem to be averse to the pattern, but you _could_ use set semantics here `seen := make(map[string]struct{})`

It's fewer allocations on the heap.

krancour (approved):
No feedback more than a nit.

LGTM after #612 is in.

jessesuen:
The PR is lacking testing. I'll be adding that before merge.

internal/api/query_freights_v1alpha1.go
jessesuen: Oh yes, my normal preference is actually:
```golang
make(map[string]bool)
```

So that the code is a bit more readable:
```golang
if seen[f.ID] {

}
```

which I prefer over:
```
if _, ok := seen[f.ID]; ok {

}
```

The reason it's a struct, is because at one point I thought I needed to do something with the actual object in the helper. But looking at the code, I ended up not doing anything with it. So I will change it to `make(map[string]bool)`

PR #630, opened 2023-09-01
control flow stages

### LINKED ISSUE #622: Control-Flow Stages
In my view, stages will not always correspond to something like an environment that needs to be updated, an Argo CD app, etc.... Consider the following deployment DAG:

```
                    /---> [usw1]---\   /---> [usw2]
[test] -> [staging]-----> [use1]--->[X]----> [use2]
                    \---> [euw1]---/   \---> [euw2]
```

The `[X]` Stage is what I'll call (for lack of a better word) a control-flow stage, which is basically just a stage that I want to use as an interaction point in the UI/API.

My use case is the following:
* The [X] stage is subscribed to three stages
* Once a state makes it past (any or all) of usw1, use1, euw1, I'd like for a way to manually simultaneously promote to usw2, use2, euw2 (using the [PromoteSubscribers feature](https://github.com/akuity/kargo/pull/612))

The way I think this could work is: we could make it possible for the promotion mechanism of this to be simply passthrough of what it is subscribed to, then immediately make that state it available in the history to subscribers. As a future improvement, we could extend this with additional user defined rules, such as requiring that the state appeared in all upstream stages history.

But the main point is not how it works, but to call out the need for a Stage as less as a thing to affect an environment but as a control flow to the overall pipeline.

krancour:
Fixes #622

Built on top of #628

Merge after #628 and only review commits starting with df73a4ed4e7e7c96435cabb9d0023021308c7603

This works by removing the constraint that all Stages must have one or more promotion mechanisms defined.

It makes this possible:

Screenshot 2023-09-01 at 1 30 26 PM

### DIFF
--- a/internal/controller/promotion/composite.go
+++ b/internal/controller/promotion/composite.go
@@ -43,6 +43,10 @@ func (c *compositeMechanism) Promote(
 	stage *api.Stage,
 	newState api.StageState,
 ) (api.StageState, error) {
+	if stage.Spec.PromotionMechanisms == nil {
+		return newState, nil
+	}
+
 	newState = *newState.DeepCopy()
 
 	logger := logging.LoggerFromContext(ctx)
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -68,10 +68,12 @@ type StageSpec struct {
 	//+kubebuilder:validation:Required
 	Subscriptions *Subscriptions `json:"subscriptions"`
 	// PromotionMechanisms describes how to incorporate newly observed materials
-	// into the Stage. This is a required field.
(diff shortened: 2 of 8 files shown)


internal/controller/promotion/composite.go
```diff
@@ -43,6 +43,10 @@ func (c *compositeMechanism) Promote(
 	stage *api.Stage,
 	newState api.StageState,
 ) (api.StageState, error) {
+	if stage.Spec.PromotionMechanisms == nil {
+		return newState, nil
+	}
+
```
jessesuen, line 49: Nice. I was hoping it would be this easy.

jessesuen (approved):

PR #636, opened 2023-09-05
s/state/freight/ in the CLI, API server, and controller

### LINKED ISSUE #624: Rename state to freight
A while back, we talked about how the term state was quite generic and overloaded word. Keeping with the shipping/cargo/transport theme, we agreed that "Freight "was a good substitute.

> * [freight](https://www.merriam-webster.com/dictionary/freight)
> 2 a : goods to be shipped : [CARGO](https://www.merriam-webster.com/dictionary/cargo)
> The freight arrived by steamboat.

I think we're at the point where we need to make this switch. It'll just be more and more painful the further off we put this.

krancour:
Fixes #624 

This PR changes "state" to "freight" in the CLI, API server, controller, and docs (where applicable; there are plenty of legitimate uses of the word "state" remaining).

**Note: This PR breaks the UI because codegen updates TS API bindings, but UI src hasn't been touched.**

@rbreeze do you maybe want to add on to this branch to fix the UI?

Other than the broken UI, there are no functional changes in this PR and manual end-to-end testing still checks out.


jessesuen:
@rbreeze agreed to prioritize this and work off your branch

jessesuen (approved):

PR #649, opened 2023-09-06
Manage all kubernetes runtime scheme in `kubeclient` package

devholic:
This commit allows all components to use same base runtime scheme.

### DIFF
--- a/internal/kubeclient/scheme.go
+++ b/internal/kubeclient/scheme.go
@@ -1,0 +1,61 @@
+package kubeclient
+
+import (
+	argocd "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
+	"github.com/pkg/errors"
+	authzv1 "k8s.io/api/authorization/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	kubescheme "k8s.io/client-go/kubernetes/scheme"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+)
+
+var kargoSchemes = runtime.SchemeBuilder{
+	corev1.AddToScheme,
+	kargoapi.AddToScheme,
+}
+
+func newScheme(extraSchemes ...func(*runtime.Scheme) error) (*runtime.Scheme, error) {
+	scheme := runtime.NewScheme()
+	for _, addToScheme := range extraSchemes {
+		if err := addToScheme(scheme); err != nil {
+			return nil, errors.Wrap(err, "add api to scheme")
+		}
+	}
+	return scheme, nil
+}
+func NewAPIScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
+
+func NewAppManagerScheme() (*runtime.Scheme, error) {
+	return newScheme(
+		corev1.AddToScheme,
+		argocd.AddToScheme,
+	)
+}
+
+func NewCLIScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
+
+func NewKargoScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
+
+func NewKargoManagerScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
+
+func NewKubernetesScheme() (*runtime.Scheme, error) {
+	return newScheme(kubescheme.AddToScheme, kargoapi.AddToScheme)
+}
+
+func NewGarbageCollectorScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
+
+func NewWebhooksServerScheme() (*runtime.Scheme, error) {
+	return newScheme(append(kargoSchemes, authzv1.AddToScheme)...)
+}
--- a/cmd/cli/root.go
+++ b/cmd/cli/root.go
@@ -20,6 +20,7 @@ import (
 	"github.com/akuity/kargo/internal/cli/login"
 	"github.com/akuity/kargo/internal/cli/option"
 	"github.com/akuity/kargo/internal/cli/stage"
+	"github.com/akuity/kargo/internal/kubeclient"
 )
 
 // rootState holds state used internally by the root command.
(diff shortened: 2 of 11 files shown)


krancour:
DRYing things up is nice, but not all components actually use the same scheme. For instance, the API server needing access to authv1 is a specialized concern. Most tests do not need such a broad scheme.

I'm not opposed to cleaning this up in some way, but I'm not sure that one-size-fits-all is the right answer.

krancour:
I still have concerns about this. It's a red flag to me when I see a bunch of unrelated concerns being aggregated in one package. We don't want to end up with a "junk drawer." `kubeclient` looks to be trending in that direction and I'd like to avoid making it worse.

Explaining why I consider these various schemes _unrelated_: I think the ideal is that each component should specify the scheme it uses, and for the sake of DRYness, do so in just one place. In most cases, that DRYness can be achieved with a package-private helper function. The scheme it uses isn't anyone else's business.

Keeping all of this in one centralized package poses its own problems as well. What happens when you need to add another API to the scheme used by one component and in doing so unnecessarily at that API to another component's scheme? After a while this is messier than if we hadn't done it.

In the few cases where a package-private helper function isn't DRY enough because the same scheme is used by multiple packages, I understand what the utility of moving these things into a separate, centralized package was, but in all of those cases, I have also noticed that functionality that probably should have been in one package to begin with was unnecessarily spread across two. In those cases, a small bit of refactoring would make the scheme used by only one package and would restore the viability of the package-private helper approach.

I'll add more specific comments in-line with the changes to describe how I think we can do this more cleanly.

internal/kubeclient/scheme.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, addToScheme := range extraSchemes {
+		if err := addToScheme(scheme); err != nil {
+			return nil, errors.Wrap(err, "add api to scheme")
+		}
+	}
+	return scheme, nil
+}
+func NewAPIScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
```
krancour, line 30: This function is used by two packages:

* `internal/api`
* `internal/kubeclient/manifest`

In `internal/kubeclient/manifest` it is used only to test the parser, which is only used in `internal/api` -- meaning the parser probably should live in `internal/api` as well.

If the parser moves to `internal/api`, it's viable to use a package-private helper function to DRY up references to this particular scheme.

internal/kubeclient/scheme.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewAppManagerScheme() (*runtime.Scheme, error) {
+	return newScheme(
+		corev1.AddToScheme,
+		argocd.AddToScheme,
+	)
+}
```
krancour, line 37: This function is only called from cmd/controlplane -- and even at that only in one place. This could be package-private if we wanted. Or we could just not bother. It doesn't actually DRY anything up.

internal/kubeclient/scheme.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewCLIScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
+
+func NewKargoScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
```
krancour, line 45: This function is used three times, but only in one package. It would be just as DRY to use a package-private helper function.

internal/kubeclient/scheme.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewKargoManagerScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
```
krancour, line 49: This function is only called from cmd/controlplane -- and even at that only in one place. This could be package-private if we wanted. Or we could just not bother. It doesn't actually DRY anything up.

internal/kubeclient/scheme.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewKubernetesScheme() (*runtime.Scheme, error) {
+	return newScheme(kubescheme.AddToScheme, kargoapi.AddToScheme)
+}
```
krancour, line 53: This function is only called from `internal/api/kubernetes` -- and even at that only in one place. This could be package-private if we wanted. Or we could just not bother. It doesn't actually DRY anything up.

In addition to this, the name of the function is misleading because it returns a scheme that includes the Kargo API as well as the core k8s API.

internal/kubeclient/scheme.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewGarbageCollectorScheme() (*runtime.Scheme, error) {
+	return newScheme(kargoSchemes...)
+}
```
krancour, line 57: This function is used by two packages:

* `cmd/controlplane`
* `internal/garbage`

A tiny amount of refactoring would make this used in only by `internal/garbage`.

Specifically, the garbage collector's "constructor" could supply the client (and the scheme it uses) itself instead of relying on `cmd/controlplane` to inject it.

After that, this could be DRYed up using a package-private helper function.

I think this also points to a smell that's recurring for most components... we probably don't need `cmd/controlplane` _ever_ injecting clients or schemes into internal components that could reasonably build those themselves, which might be tidier.

internal/kubeclient/scheme.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewWebhooksServerScheme() (*runtime.Scheme, error) {
+	return newScheme(append(kargoSchemes, authzv1.AddToScheme)...)
+}
```
krancour, line 61: This function is only called from `cmd/controlplane` -- and even at that only in one place. This could be package-private if we wanted. Or we could just not bother. It doesn't actually DRY anything up.

internal/kubeclient/scheme.go
krancour: This is also another spot where maybe the contructor should take over this responsibility. See https://github.com/akuity/kargo/pull/649/files#r1318950057

internal/kubeclient/scheme.go
krancour: This is also another spot where maybe the contructor should take over this responsibility. See https://github.com/akuity/kargo/pull/649/files#r1318950057

krancour:
Ok... played with this a bit. _This was quick and dirty_, but I was thinking of something along these lines:

https://github.com/akuity/kargo/compare/main...krancour:kargo:krancour/dryer-schemes?expand=1

This gets us down to every major component building their scheme just once.

You can search for instances of `runtime.NewScheme` to see that this is much DRYer it is now.

To do this, I had to do preliminary refactoring to eliminate the existing smells I talked about that were, quite frankly, my own fault. This way we're fixing the existing problems instead of just putting a band-aid on them.

Let me know what you think of this approach.

krancour:
@devholic what I did was quick... it probably needs more scrutiny because it's definitely a larger change.

krancour:
Also I appreciate the effort and attention you brought to the schema problem. It was very clearly not sustainable the way we were doing it before. That's my bad.

krancour:
I'm making an issue for this and we'll revisit it after v0.1.0.

PR #650, opened 2023-09-06
update PromotionPhases

### LINKED ISSUE #600: Rework Promotion phase terminology
I noticed a promotion is in one of these phases:

```
	PromotionPhasePending    PromotionPhase = "Pending"
	PromotionPhaseInProgress PromotionPhase = "Promoting"
	PromotionPhaseComplete   PromotionPhase = "Completed"
	PromotionPhaseFailed     PromotionPhase = "Failed"
```

Some things I would like to change:
* `Completed` is what I consider an aggregate phase of any non-running phase. I also consider the term ambiguous (since something that is Failed is also complete). 
* The term `Promoting` is a bit specific and would prefer a more generic `Running` phase to leave room for when Promotions might be doing something but not specifically promoting. It's easier colloquially as well ("get all running promotions" vs. "get all promoting promotions").

I would like to re-use the state machine of Argo CD (and keep the `Pending` phase that Kargo needs but Argo CD doesn't):

```
	OperationRunning     OperationPhase = "Running"
	OperationFailed      OperationPhase = "Failed"
	OperationError       OperationPhase = "Error"
	OperationSucceeded   OperationPhase = "Succeeded"
```

(the only one I excluded is `Terminating` because I don't feel that concept is necessary yet).

The term `Completed` would mean Failed/Error/Succeeded, but it is not part of the enum.

If you are wondering what might be the difference between `Failed` vs. `Error`, `Failed` would indicate an obvious user-related problem (e.g. user rejected a pull request related to the promotion). Whereas `Error` would indicate something abnormal (connection / networking issues).

krancour:
Fixes #600 

This goes a little bit beyond #600. While auditing where and how the PromotionPhases I was modifying are used, I discovered:

* A lot of references to "outstanding" or "non-outstanding" Promotions. I felt like this was a bit ambiguous. For instance, it wouldn't be crazy for someone to assume that a Promotion that is Running is _not_ outstanding (but it is).

  I updated the code to use less ambiguous terminology -- terminal or non-terminal.
  
  This also presented the opportunity to add a convenient `phase.IsTerminal()` function that makes all other code that needs to make such a determination less brittle when we add new phases.

* Places where predicates used in indexing Promotions by non-terminal (formerly "non-outstanding") used the word "filter" which can also be confusing. The purpose of a filter is to _exclude_ things, while the purpose of a predicate it to _select_ things. So I made updates to these as well to improve the clarity of the code.

I couldn't spot any UI changes that were required to complement the phase name changes, but @rbreeze or @rpelczar can correct me if I missed something.

### DIFF
--- a/api/v1alpha1/promotion_types.go
+++ b/api/v1alpha1/promotion_types.go
@@ -30,6 +30,11 @@ const (
 	PromotionPhaseErrored PromotionPhase = "Errored"
 )
 
+// IsTerminal returns true if the PromotionPhase is a terminal one.
+func (p *PromotionPhase) IsTerminal() bool {
+	return *p == PromotionPhaseSucceeded || *p == PromotionPhaseErrored
+}
+
 //+kubebuilder:resource:shortName={promo,promos}
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
--- a/api/v1alpha1/promotion_types.go
+++ b/api/v1alpha1/promotion_types.go
@@ -7,10 +7,27 @@ import (
 type PromotionPhase string
 
 const (
-	PromotionPhasePending    PromotionPhase = "Pending"
-	PromotionPhaseInProgress PromotionPhase = "Promoting"
-	PromotionPhaseComplete   PromotionPhase = "Completed"
-	PromotionPhaseFailed     PromotionPhase = "Failed"
+	// PromotionPhasePending denotes a Promotion that has not been executed yet.
+	// i.e. It is currently waiting in a queue. Queues are stage-specific and
+	// prioritized by Promotion creation time.
+	PromotionPhasePending PromotionPhase = "Pending"
+	// PromotionPhaseRunning denotes a Promotion that is actively being executed.
+	//
+	// TODO: "Active" is the operative word here. We are leaving room for the
+	// possibility in the near future that an in-progress Promotion might be
+	// paused/suspended pending some user action.
+	PromotionPhaseRunning PromotionPhase = "Running"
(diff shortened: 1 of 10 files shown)


api/v1alpha1/promotion_types.go
```diff
@@ -7,12 +7,34 @@ import (
...
+	//
+	// TODO: "For technical reasons" is the operative phrase here. We are leaving
+	// room for the possibility in the near future that a Promotion might fail
+	// as a result of some user action.
+	PromotionPhaseErrored PromotionPhase = "Errored"
 )
 
+// IsTerminal returns true if the PromotionPhase is a terminal one.
+func (p *PromotionPhase) IsTerminal() bool {
+	return *p == PromotionPhaseSucceeded || *p == PromotionPhaseErrored
```
jessesuen, line 35: I  originally suggested using "complete" terminology, but I can see the argument that complete might be confused with success. So I'm okay with this.

jessesuen (approved):

jessesuen:
@rpelczar - the promotion list UI that you are currently working on will need to accommodate this change.

api/v1alpha1/promotion_types.go
krancour: Oh... I knew you wanted to change what _was_ Completed to Succeeded...

I think I got my wires crossed a bit and thought we were getting rid of "Completed" entirely. Re-reading #600 now, I see you meant for "completed" to becomes what I now call "terminal."

Sorry for the confusion.

But as you noted, "terminal" may avoid further confusion, so in the words of the late, great Bob Ross: "We don't make mistakes. Just happy little accidents." :smile:

PR #653, opened 2023-09-06
feat: support supplying bcrypt hash of admin password

jessesuen:
We should allow users to supply the bcrypt hash of the password instead of the plaintext password in the values.yaml. Since the hash is safe to store in plaintext/git but a password is not.

Additionally, the `bcrypt` function is not hermetic/stable/gitops-friendly and so if user uses password, every time user applies the helm chart, they will get a different hash. But supplying a hash **_is_** hermetic and wouldn't change between applies.


krancour:
Good call on this.

krancour (approved):

Issue #656, opened 2023-09-06
Promotion: invalid memory address or nil pointer dereference caused controller crash


jessesuen:
Tried to use bookkeeper promotion, not sure if the spec is correct, but it resulted in a controller panic.

```yaml
apiVersion: kargo.akuity.io/v1alpha1
kind: Stage
metadata:
  creationTimestamp: "2023-08-25T06:04:19Z"
  generation: 7
  name: dev
  namespace: kargo-guestbook
  resourceVersion: "127234591"
  uid: 32626a1f-c00f-468a-8e76-d1a0d83e7920
spec:
  promotionMechanisms:
    argoCDAppUpdates:
    - appName: guestbook-dev
      appNamespace: argocd
    gitRepoUpdates:
    - bookkeeper: {}
      readBranch: ""
      repoURL: https://github.com/jessesuen/guestbook-deploy.git
      writeBranch: env/dev
  subscriptions:
    repos:
      git:
      - branch: main
        repoURL: https://github.com/jessesuen/guestbook-deploy.git
      images:
      - gitRepoURL: https://github.com/jessesuen/guestbook.git
        repoURL: ghcr.io/jessesuen/guestbook
        updateStrategy: SemVer
```

Promotion:
```yaml
apiVersion: kargo.akuity.io/v1alpha1
kind: Promotion
metadata:
  creationTimestamp: "2023-09-06T20:35:31Z"
  generation: 1
  name: dev.01h9p3a3qwkfajkay4j0zfgq9n.6075e1a
  namespace: kargo-guestbook
  ownerReferences:
  - apiVersion: kargo.akuity.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Stage
    name: dev
    uid: 32626a1f-c00f-468a-8e76-d1a0d83e7920
  resourceVersion: "127228619"
  uid: e74751e1-4360-4930-88d8-ec3c9c632d29
spec:
  freight: 6075e1ab632d6ed194b6ecda069998c46702b6ea
  stage: dev
status:
  phase: Pending
```

Logs
```
time="2023-09-06T20:37:34Z" level=info msg="Starting Kargo Controller" commit=50e24a3071582e43de758b90818f0adde99a4634 version=devel+50e24a3.dirty
time="2023-09-06T20:37:45Z" level=info msg="kustomize build /tmp/4076788782/repo/env/dev" dir= execID=4f5dd
[...]

PR #657, opened 2023-09-06
fix: don't crash controller on reconcile panics

jessesuen:
Partially resolves https://github.com/akuity/kargo/issues/656

Depends on https://github.com/akuity/kargo/pull/650

If a promotion function panics, we will fail the promotion rather than crash the controller, and the promotion will have a proper message and phase:

```yaml
kind: Promotion
metadata:
  creationTimestamp: "2023-09-06T21:11:10Z"
  generation: 1
  name: dev.01h9p5bczmwqj1z5hhf6880p8f.e9b3b60
  namespace: kargo-guestbook
  ownerReferences:
  - apiVersion: kargo.akuity.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Stage
    name: dev
    uid: 32626a1f-c00f-468a-8e76-d1a0d83e7920
  resourceVersion: "127330362"
  uid: 5482cac2-16ed-4f5f-ad55-2d3a4e5c8db7
spec:
  freight: e9b3b603e17ea2e291535b605a58c2e62f1949be
  stage: dev
status:
  error: 'runtime error: invalid memory address or nil pointer dereference'
  phase: Failed
```

### DIFF
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -2,6 +2,7 @@ package promotions
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
@@ -326,16 +327,26 @@ func (r *reconciler) serializedSync(
 
 				phase := kargoapi.PromotionPhaseSucceeded
 				phaseError := ""
-				if err = r.promoteFn(
-					promoCtx,
-					promo.Spec.Stage,
-					promo.Namespace,
-					promo.Spec.Freight,
-				); err != nil {
-					phase = kargoapi.PromotionPhaseErrored
-					phaseError = err.Error()
-					logger.Errorf("error executing Promotion: %s", err)
-				}
+
+				func() {
+					defer func() {
+						if err := recover(); err != nil {
+							logger.Errorf("Promotion panic: %v", err)
+							phase = kargoapi.PromotionPhaseErrored
+							phaseError = fmt.Sprintf("%v", err)
+						}
+					}()
+					if err = r.promoteFn(
+						promoCtx,
+						promo.Spec.Stage,
+						promo.Namespace,
+						promo.Spec.Freight,
+					); err != nil {
+						phase = kargoapi.PromotionPhaseErrored
+						phaseError = err.Error()
+						logger.Errorf("error executing Promotion: %s", err)
+					}
+				}()
 
 				if err = kubeclient.PatchStatus(ctx, r.kargoClient, promo, func(status *kargoapi.PromotionStatus) {
 					status.Phase = phase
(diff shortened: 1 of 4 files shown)


jessesuen:
I should rebase this ontop of https://github.com/akuity/kargo/pull/650.

krancour:
Do we want this change long term? If we do it here, idk why we wouldn't be doing it all over the place, and obviously we won't because this isn't customary in Go.

So I'm suggesting that maybe we leave a comment here that notes this can be removed when Bookkeeper is fixed.

Otherwise LGTM

jessesuen:
> Do we want this change long term? If we do it here, idk why we wouldn't be doing it all over the place, and obviously we won't because this isn't customary in Go.

We do. One single bad resource should not cause system-wide unavailability. The risk of not doing this means complete downtime / denial of service.  In fact this is a common enough scenario that api-machinery controller runtime [has a helper for this](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/runtime#HandleCrash) (but I don't want to use it here since we need to fail the promotion rather than requeue and retry reconciliation later).

krancour:
Fair point... is it then worthwhile to apply something similar to the other two reconcilers? (Stages and Applications)

jessesuen:
> Fair point... is it then worthwhile to apply something similar to the other two reconcilers? (Stages and Applications)

Yea we should. I wasn't sure if those were impacted if they were built using kubebuilder (I know the promo reconciler was custom built). I'll go check those other reconcilers and improve those if they crash the controller.

krancour:
> I wasn't sure if those were impacted if they were built using kubebuilder

Yea... kubebuilder feels overly-opinionated in regard to overall project structure, so we just use controller runtime directly.

jessesuen:
OK, I improved it so all reconcilers will recover from panic. Looks like controller-runtime has made this even easier than it used to be, by introducing a `RecoverPanic: true` flag:

```
	return controller.Options{
		RecoverPanic: true,
	}
```

krancour:
> Looks like controller-runtime has made this even easier than it used to be, by introducing a RecoverPanic: true flag

Beautiful!

internal/controller/promotions/promotions.go
```diff
@@ -326,16 +328,26 @@ func (r *reconciler) serializedSync(
...
+						promoCtx,
+						promo.Spec.Stage,
+						promo.Namespace,
+						promo.Spec.Freight,
+					); err != nil {
+						phase = kargoapi.PromotionPhaseErrored
+						phaseError = err.Error()
+						logger.Errorf("error executing Promotion: %s", err)
+					}
+				}()
```
krancour, line 350: Are these changes still required in light of the recover panic option?

jessesuen: Yes I intentionally left this in, because this has the added benefit of patching the promotion so that promo enters a Error state. Without this, the Promos that cause the panic, will continue to reconcile in a loop until the crash is dealt with. Unlike Stages/Applications (which do have to reconcile forever), at least with Promos we can stop the bleeding.

krancour: That makes sense!

krancour (approved):

PR #658, opened 2023-09-06
Improved health states

### LINKED ISSUE #623: Stage/Freight Health is not always applicable

krancour:
Fixes #623

WIP because corresponding UI changes are missing.

I'm not sure exactly what we want to do with NotApplicable in the UI. @rbreeze ?

This also probably needs a rebase after #650 is merged.


krancour:
Closing this as #676 probably renders it moot.

@jessesuen do let me know if you think HealthState NotApplicable will still be a thing, but I don't think so.

PR #660, opened 2023-09-06
feat: argocd app namespace is optional and not defaulted

jessesuen:
The Stage's argocd application namespace (typically `argocd`) currently:

1. is required that it be specified
2. and if left unspecified, we will default the value to that of the Stage namespace

The issue I have with this is:
1. application namespace is generally not a concern for end users (in the default and most common operation mode of Argo CD).
2. the correct namespace will be more commonly `argocd` instead of the Stage's namespace
3. an admin should be able to change this centrally, without touching everyone's stages
4. the defaulting behavior done by our webhook, will cause problems with GitOps (because this is a list field)

This PR makes it such that:
1. we don't default the namespace in the webhook
2. we will choose to use ARGOCD_NAMESPACE if unspecified
3. and if ARGOCD_NAMESPACE is unspecified, we will use `argocd`

### DIFF
--- a/internal/webhook/stage/webhook.go
+++ b/internal/webhook/stage/webhook.go
@@ -41,24 +41,9 @@ func SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-func (w *webhook) Default(_ context.Context, obj runtime.Object) error {
-	stage := obj.(*kargoapi.Stage) // nolint: forcetypeassert
+func (w *webhook) Default(_ context.Context, _ runtime.Object) error {
 	// Note that defaults are applied BEFORE validation, so we do not have the
 	// luxury of assuming certain required fields must be non-nil.
-	if stage.Spec != nil {
-
-		if stage.Spec.PromotionMechanisms != nil {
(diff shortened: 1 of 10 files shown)


krancour:
Agree. In fact I thought this was done already. I see it was a TODO. That must be what I'm actually remembering.

internal/webhook/stage/webhook.go
```diff
@@ -42,23 +42,9 @@ func SetupWebhookWithManager(mgr ctrl.Manager) error {
 }
 
 func (w *webhook) Default(_ context.Context, obj runtime.Object) error {
-	stage := obj.(*kargoapi.Stage) // nolint: forcetypeassert
+	_ = obj.(*kargoapi.Stage) // nolint: forcetypeassert
```
krancour, line 45: What is the function of this line?

Also glad to see the comment about nil checks remains even though this function is basically a no-op now. It will help avoid problems in the future if we have to touch this function again. 👍

jessesuen: Not sure why I left it, tbh. I'll remove it.

krancour (approved):

Issue #644, opened 2023-09-05
UI: more linkage


jessesuen:
Kargo UI needs more linkage:
- [ ] to the Argo CD application
- [ ] to the application git repository tag (if applicable)
- [ ] to the git repository (and possibly commits)
- [ ] to the container registry (if docker hub, ghcr.io, quay.io, etc...)
- [ ] to the helm repository

jessesuen:
For 1 (the argo cd application), it requires us to expose a Kargo setting to know the base URL(s) of Argo CD instances.

PR #661, opened 2023-09-07
feat: add GetConfig API

jessesuen:
Partially resolves https://github.com/akuity/kargo/issues/644

This introduces a GetConfig API endpoint, which the UI will need for things like deep linking to Argo CD applications.

The first item we return in the config, is Argo CD instances and their corresponding URLs. e.g.:

```json
{
   "argocd_shards":{
      "":{
         "url":"https://argocd.com",
         "namespace":"argocd"
      }
   }
}
```

Using the above response the UI can reconstruct a deep link to the application. The normal, non-sharded behavior, will use the empty string as the shardName.


jessesuen:
This API is intended to be authenticated, and so that's why it's separate from GetPublicConfig

krancour (approved):

alexmt (approved):

Issue #672, opened 2023-09-07
Kargo evaluated a Freight to be Unhealthy when app was still Progressing


jessesuen:
I promoted a git commit via Kargo. Kargo initiated the Argo CD app sync operation. The sync completed, but the app took some time to fully roll out and was progressing for some time. But Kargo assessed the Freight as `Unhealthy` during stage Reconciliation because `Argo CD Application "guestbook-dev" in namespace "argocd" has health state "Progressing"`:

<img width="230" alt="image" src="https://github.com/akuity/kargo/assets/12677113/79914383-b2aa-42c3-9c97-c8ec812aa88d">

```yaml
  history:
  - commits:
    - branch: main
      healthCheckCommit: 10d7fd861e94dd57a63c95b05dc21ad885b0703e
      id: 5c7d29985412024a39dff4b0286bf9eb3f2a0450
      repoURL: https://github.com/jessesuen/guestbook-deploy.git
    firstSeen: "2023-09-07T22:48:07Z"
    health:
      issues:
      - Argo CD Application "guestbook-dev" in namespace "argocd" has health state
        "Progressing"
      status: Unhealthy
    id: 4e1f65309798525de29191ea734ea4f980f07b43
    images:
    - gitRepoURL: https://github.com/jessesuen/guestbook/tree/v0.0.6
      repoURL: ghcr.io/jessesuen/guestbook
      tag: v0.0.6
```

Issue #670, opened 2023-09-07
Argo CD Application Sync status should not impact Freight health


jessesuen:
Many Argo CD users run in a mode where apps are in a perpetual state of drift, and it is unreasonable to expect that Sync status will be Synced for the freight to be healthy. In other words, even after a sync operation is completed, a user's app might still be OutOfSync. Today, the app associated with a Stage must be Synced in order for the Freight to be healthy.

Freight health should really only rely on app health. If we need to know when a sync operation completed, we can look at whether or not the `operation` field of an application is present or not (as opposed to `status.sync.status`).

PR #673, opened 2023-09-08
fix: introduce progressing health state for progressing/syncing argocd applications

alexmt:
Closes https://github.com/akuity/kargo/issues/672

https://github.com/akuity/kargo/assets/426437/5bca2f77-09ab-4a5a-a0f0-842fb952f6b4


krancour:
Two things...

1. There is other pending health work in #658, so maybe we want to consider combining those? The only problem with #658 is it currently lacks UI changes for the new states it introduces. i.e. They needs the same UI treatment that Progressing gets in this PR.

2. "Progressing" is a little confusing to me in this context.

    What does it mean, practically speaking, for a Stage's current Freight to be in this state that is _different_ from being unhealthy? Health here is a matter of suitability for further promotion. If I ask "Is the Stage healthy?" and an App is progressing, the simple answer is "no it's not." (And the reason field already does say why.)

    I start to feel like Kargo's a leaky abstraction if we let this level of detail that's an Argo CD=specific concern bubble up to a higher layer in the stack where that level of detail seems less consequential.

jessesuen:
> What does it mean, practically speaking, for a Stage's current Freight to be in this state that is different from being unhealthy?
> Health here is a matter of suitability for further promotion

I agree that there is definitely confusion. I believe the problem is that we have been equating promotability with argo cd app health. But at the same time, we've been continuously updating the currentFreight health with Argo CD app health changes. And we have been presenting these changes in UI as living health.

It seems to me that we need to separate the concept of Argo CD health from Freight being "promotable." This is what I propose:

* Health status should be a living assessment of the underlying Argo CD app(s) associated with the Stage. Health can flap, and would have all of the Argo CD health statuses (Healhty, Degraded, Progressing, Unknown). Health is not applicable on Stages that have no Argo CD app association.

* We introduce a new binary status of a Freight: Qualified vs. Unqualified. A qualified freight has gone through the necessary checks to be promotable to the next stage. If a Stage has an Argo CD app(s), it  simply means the apps at one point in time became healthy with that freight. Once a Freight has been qualified, it will never become unqualified.

krancour:
> We introduce a new binary status of a Freight: Qualified vs. Unqualified. A qualified freight has gone through the necessary checks to be promotable to the next stage. If a Stage has an Argo CD app(s), it simply means the apps at one point in time became healthy with that freight. Once a Freight has been qualified, it will never become unqualified.

I like that. That makes sense. And it avoids what would have been an eventual problem of "Hey... this isn't healthy anymore but it was still available to the next Stage. What gives?"

Just to make sure we're on the same page... Freight has to be re-qualified at each stage, right? i.e. Qualifying in test _doesn't_ mean its smooth sailing all the way to prod, right? A freight is qualified for prod after it first becomes healthy in the stage _prior_ to prod?

jessesuen:
> Freight has to be re-qualified at each stage, right?

Correct.

krancour:
Ok... The separation of "live health" from qualification definitely clears up my confusion.

I will make a separate issue for that.

I'll rebase #658 on this PR after it's merged, so no need to worry about other pending health-related changes. And I think I should also be able to look at the UI changes in this PR to see how to take #658 the rest of the way.

internal/controller/stages/argocd.go
```diff
@@ -88,15 +89,49 @@ func (r *reconciler) checkHealth(
...
+			Issues: []string{fmt.Sprintf("Argo CD Application %q in namespace %q is progressing",
+				app.Name, app.Namespace)},
+		}
+	default:
+		return kargoapi.Health{Status: kargoapi.HealthStateHealthy}
+	}
+}
+
+func stageHealthForAppSync(app *argocd.Application, revision string) kargoapi.Health {
+	if app.Status.Sync.Status != argocd.SyncStatusCodeSynced ||
```
krancour, line 120: I think if we change this line a little, we can fix #670 and close #675 without merging.

krancour: Although, [this](https://github.com/akuity/kargo/pull/675#issuecomment-1711021158) would also have to be addressed.

krancour: Nevermind. Will just rebase #670 on this.

krancour (approved):
Lots of health-related changes pending, but this is a step in the right direction.

jessesuen (approved):

PR #675, opened 2023-09-08
evaluate whether an App is synced only on basis of revision

### LINKED ISSUE #670: Argo CD Application Sync status should not impact Freight health

krancour:
Fixes #670 

If applicable, this still checks whether an App is "synced," but it's based solely on revision and not on sync state.

### DIFF
--- a/internal/controller/stages/argocd.go
+++ b/internal/controller/stages/argocd.go
@@ -124,8 +124,7 @@ func stageHealthForAppHealth(app *argocd.Application) kargoapi.Health {
 }
 
 func stageHealthForAppSync(app *argocd.Application, revision string) kargoapi.Health {
-	if app.Status.Sync.Status != argocd.SyncStatusCodeSynced ||
-		(revision != "" && app.Status.Sync.Revision != revision) {
+	if revision != "" && app.Status.Sync.Revision != revision {
 
 		if app.Operation != nil && app.Operation.Sync != nil {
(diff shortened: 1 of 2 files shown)


jessesuen:
Since I merged the predicate PR, can you update the predicate such that we refresh the stage when `status.sync.revision` changes?

https://github.com/akuity/kargo/blob/main/internal/controller/applications/applications.go#L176-L180

krancour:
@jessesuen maybe...

This PR probably gets closed depending on what @alexmt does [here](https://github.com/akuity/kargo/pull/673#discussion_r1319312296).

krancour:
> This PR probably gets closed depending on what @alexmt does https://github.com/akuity/kargo/pull/673#discussion_r1319312296.

Sorry... just realized this needs to change irrespective of what happens with this PR or #675.

I will open a separate PR to fix that.

jessesuen (approved):

PR #677, opened 2023-09-08
Add `CreateOrUpdate` API

devholic:
CLI's `apply` command not works if resource doesn't exists. This PR fixes this behavior by changing CLI to call `CreateOrUpdate` API instead of `Update` API to create resource if not exists.


internal/cli/apply/apply.go
```diff
@@ -58,42 +58,62 @@ kargo apply -f stage.yaml
...
+				if printer == nil {
+					name := types.NamespacedName{
+						Namespace: obj.GetNamespace(),
+						Name:      obj.GetName(),
+					}.String()
+					fmt.Fprintf(opt.IOStreams.Out, "%s Created: %q\n", obj.GetKind(), name)
+					continue
+				}
+				_ = printer.PrintObj(&obj, opt.IOStreams.Out)
+			}
```
krancour, line 98: Is there any reason other than getting resource name and namespace that the response contains full YAML representations of the affected resources? It seems unnecessarily heavyweight to include all of that in the response and unmarshal it just to get these two pieces of info which could just be string fields in a more succinct response.

I am probably missing some other reason that we are doing it this way?

krancour: Ok. I hadn't anticipated that as something that was supported. Just looked at `kubectl` and you can indeed specify `--output` on an `apply`. So I guess this is in keeping with our attempt to not surprise anyone who's already well-accustomed to `kubectl`.

Thanks for the explanation.

SGTM

krancour (approved):

PR #685, opened 2023-09-08
separate stage health and freight qualification

### LINKED ISSUE #570: Replace Stage health information with structures checks status
UI needs to visualize the current stage state. Right now stage CRD/API provides “health” that includes aggregated status (healthy/unhealthy/unknown) and the list of “issues”.

I think ‘health’ is confusing since there are a lot of checks not related to health. Simplest example: stage which is “unhealthy” because the related Argo CD app is not synced. This is how it looks like in UI:

image

UI is showing a broken heart, which does not make sense because UI has no idea what health means. 

```yaml
  currentState:
    firstSeen: "2023-08-15T21:41:50Z"
    health:
      issues:
      - Argo CD Application "kargo-demo-test" in namespace "argocd" is not synced
        to revision ""
      status: Unhealthy
    id: 4b94f4bf6adcca4765b9a2c554647b649646784b
```

# Proposal

Replace high-level health information with detailed/structured status of performed checks. Example to illustrate the idea (structure needs to be discussed):

```yaml
  currentState:
    firstSeen: "2023-08-15T21:41:50Z"
    checks:
      argocd:
      - appName: kargo-demo-test
         sync:
            valid: false
            reason: Argo CD Application "kargo-demo-test" in namespace "argocd" is not synced to revision "abc"
    id: 4b94f4bf6adcca4765b9a2c554647b649646784b
```

The predefined structure allows UI to render meaningful information about stage state. E.g. if UI knows status has information about the application sync state it would use appropriate icon and URL that redirects user to Argo CD application.

### LINKED ISSUE #676: Separate current health of stage+freight from freight's promotability to next stage(s)
In #673, @jessesuen wrote:

> we have been equating promotability with argo cd app health. But at the same time, we've been continuously mirroring the currentFreight health with Argo CD app health changes. And we have been presenting these changes in UI as living health.
>
> It seems to me that we need to separate the concept of Argo CD health from Freight being "promotable." This is what I propose:
>
>    Health status should be a living assessment of the underlying Argo CD app(s) associated with the Stage. Health can flap, and would have all of the Argo CD health statuses (Healhty, Degraded, Progressing, Unknown). Health is not applicable on Stages that have no Argo CD app association. Health is also not applicable to Freight. It is simply a mirror of Argo CD app health.
>
>    We introduce a new binary status of a Freight: Qualified vs. Unqualified. A qualified freight has gone through the necessary checks to be promotable to the next stage. If a Stage has an Argo CD app(s), it simply means the apps at one point in time became healthy with that freight. If there is no associated Argo CD app, then Freight is immediately qualified for downstream stages. Once a Freight has been qualified in a Stage, it will never become unqualified.

I would like to work on this issue if that's ok.

krancour:
Fixes #676

Does _not_ fix #570 -- that will be addressed separately.

Note to reviewer(s): It will be easiest to look at this commit by commit. All codegen is isolated to one commit that you can ignore.

I have also validated this e2e.


internal/controller/stages/stages.go
```diff
@@ -328,7 +328,10 @@ func (r *reconciler) syncStage(
 				Status: kargoapi.HealthStateHealthy,
 			}
 		}
-		status.CurrentFreight.Health = &health
+		status.Health = &health
+		if health.Status == kargoapi.HealthStateHealthy {
+			status.CurrentFreight.Qualified = true
+		}
```
jessesuen, line 334: I think we shouldn't set health on Stages that don't have Argo CD App association. Not all stages in the UI should have hearts, such as the ones which are used for control-flow

jessesuen (approved):
I have one request which is not to give health status to stages that don't have argo cd apps behind it. Other than that LGTM

internal/controller/stages/stages.go
krancour: Totally agree. The PR with the new `NotApplicable` state would have addressed that, but I ended up closing it.

Especially since the structure of Health is going to change completely when I work on #570, I will get this for now to where Health is simply nil if not applicable.

jessesuen (approved):

PR #687, opened 2023-09-08
feat: add status.currentPromotion to Stage

jessesuen:
A Stage's `status.currentPromotion` is now managed by Kargo and contains a reference to the promotion name + freight information. 

1. Set at the beginning of the promo reconciler `promoteFn()`.
2. It is set to nil at the end of `promoteFn()`
3. For redundancy, it is also set to nil by the stage reconciler, once it has gotten past the `hasNonTerminalPromos` check.

With this information, UI can show when a Stage is in the middle of a Promotion (without needing to look at the promo list).


internal/controller/stages/stages.go
```diff
@@ -313,6 +313,7 @@ func (r *reconciler) syncStage(
 	}
 
 	status.ObservedGeneration = stage.Generation
+	status.CurrentPromotion = nil
```
krancour, line 316: I know you mentioned it in the PR's description, but I'm not clear about the need for this line.

krancour: Once I understand, LGTM

krancour: Nevermind. The operative phrase was "once it has gotten path the no non-terminal promos check."

LGTM

krancour (approved):

internal/controller/stages/stages.go
jessesuen: It's for resilience. We can't rely solely on the promo reconciler for clearing this because promos can be deleted out of band and never get a chance to clear the promotion.

Originally, I only had it in the stage reconciler, but the stage didnt' get reconciled until several minutes after the promo finished. So I need it there so UI is reactive.

krancour: > but the stage didnt' get reconciled until several minutes after the promo finished

Ya... slow refresh is a vexing problem when there aren't associated Argo CD Apps. ☹️

Issue #474, opened 2023-07-18
UI: Migrate Stages DAG view to a library which supports custom element nodes


rbreeze:
The Antd graphing library used in #472 does not support using a custom DOM element as a node (since it uses `canvas`), so any customization must be done through images. This isn't conducive to extensions nor future features and enhancements, so we should find a different library to render the DAG which supports custom elements.
```[tasklist]
### Tasks
- [ ] https://github.com/akuity/kargo/issues/474
```

rpelczar:
In my opinion, until we don't know what we need, we don't have to plan to migrate to another library. API for extending the node element is available [here](https://g6.antv.antgroup.com/en/api/shape-properties), and I think it is more than we need for now. We can add any text or button, I am not sure about svg icons only.

rbreeze:
I agree that we don't have to start migrating yet. That being said, with the current approach the only way to add new information and features to the nodes in the Graph view is through SVG. This will be a deterrent to adding new features, and introduce an additional maintenance burden and mental overhead to working on the Graph view. It will also complicate the process of building extensions to this view, which we want to avoid to help encourage community users to build extensions.

PR #688, opened 2023-09-08
feat: render stages DAG using React & dagre library

alexmt:
Closes https://github.com/akuity/kargo/issues/474

https://github.com/akuity/kargo/assets/426437/3b06131c-d799-4330-abba-cd8603fc6b6a


jessesuen (approved):
looks great! I think the diagonal lines also look better

PR #700, opened 2023-09-11
check stage health even when there is no current freight

### LINKED ISSUE #692: Stage health should be populated even if there is not current freight
We've agreed to separate stage health and freight promotability. So even if stage don't run any  freight it still should have health

krancour:
Fixes #692

I decided to tackle this separately from #570 after I saw how simple this was.

The test refactoring here is mostly mere cleanup for test cases that were impacted by the change anyway.


alexmt (approved):
LGTM

PR #702, opened 2023-09-11
updated Argo CD and cert-manager in quick start

kencochrane:
updated Argo CD and cert-manager in the quick start and makefile


morey-tech (approved):

krancour (approved):
Local dev env comes up clean with these changes.

Thanks @kencochrane!

Issue #699, opened 2023-09-11
Promos fail when Stage's availableFreight is stale


jessesuen:
When using PromoteSubscribers feature, Promos will get created, but they might immediately fail because availableFreight of that Stage might not have been updated yet.

<img width="1056" alt="image" src="https://github.com/akuity/kargo/assets/12677113/4556b5ec-7aaf-4110-a720-03c87b9411bf">

PR #711, opened 2023-09-12
feat: use event handlers to re-reconcile related objects

jessesuen:
Partially resolves https://github.com/akuity/kargo/issues/699

This PR does a few things:

1. The stage reconciler watches promos that enter into a terminal state. Then it enqueues the owning Stage
2. The stage reconciler watches all stages whose history has changed. Then it enqueues downstream Stages
3. Fixes a bug in the logger used in the AppHealthSyncStatusChangePredicate

Using the above technique, we can make Stage reconciliation reactive enough that https://github.com/akuity/kargo/issues/699 becomes less noticeable, but still possible.


jessesuen:
I think it may be possible to completely remove the App reconciler, if we use the same technique that I did here where the Stage watches Promotion objects (as a separate effort).

krancour:
I understand parts 2 and 3.

I am confused by part 1 and why the stages reconciler would also be watching promos now...

Wouldn't it be simpler to have the promos reconciler enqueue the owning stage for reconciliation when it completes a promotion?

jessesuen:
@krancour I don't think the Kubebuilder approach will work:

1. It doesn't appear there is a way where I can only enqueue the Stage only if the Promo goes from non-terminal to terminal. e.g. there is no Update() mechanism to compare two objects to see if it changed state in the way that I care about.
2. What we have today, only to builds an efficient index on listing non-terminal promos by stage, to make the reconcile slightly faster, but it doesn't have a mechanism to requeue Stages.
3. The Kubebuilder technique relies on a cruder `Owns` helper to reconcile the Cronjob, but it does so on _**any**_ update, but we only care when a promotion is finished. If you see the docs, it is only doing the first half of what I want:

```
// Owns defines types of Objects being *generated* by the ControllerManagedBy, and configures the ControllerManagedBy to respond to
// create / delete / update events by *reconciling the owner object*.  This is the equivalent of calling
// Watches(&source.Kind{Type: <ForType-forInput>}, &handler.EnqueueRequestForOwner{OwnerType: apiType, IsController: true}).
func (blder *Builder) Owns(object client.Object, opts ...OwnsOption) *Builder {
``` 

So I think the approach I am taking is valid. Here is the example I followed:

https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/builtins/main.go#L67-L72

krancour:
> I don't think the Kubebuilder approach will work

Yep. I see now that `.Owns()` doesn't really offer a way to limit the Promo changes that you want to trigger a Stage reconciliation.

What you're doing here makes sense and agrees with the docs. 👍 

LGTM

krancour (approved):

Issue #708, opened 2023-09-12
Return native kube API types instead of a proto conversion


jessesuen:
In Kargo API, we convert from kubernetes types, to a different type which is returned in our gRPC/Connect server. And so we have conversion code like the following:

```golang
func ToPromotionProto(p kargoapi.Promotion) *v1alpha1.Promotion {
	metadata := p.ObjectMeta.DeepCopy()
	metadata.SetManagedFields(nil)

	return &v1alpha1.Promotion{
		ApiVersion: p.APIVersion,
		Kind:       p.Kind,
		Metadata:   typesmetav1.ToObjectMetaProto(*metadata),
		Spec: &v1alpha1.PromotionSpec{
			Stage:   p.Spec.Stage,
			Freight: p.Spec.Freight,
		},
		Status: &v1alpha1.PromotionStatus{
			Phase: string(p.Status.Phase),
			Error: p.Status.Error,
		},
	}
}
```

This is undesirable for many reasons:
* it must be manually updated
* it is error prone (e.g. we forgot to add apiVersion and kind at one point)
* omitempty semantics are inconsistent from native kubernetes type, and so our API ends up returning fields set to be empty strings as opposed to omitted strings.
* We face limitations such as: https://github.com/akuity/kargo/issues/707

Instead of converting and maintaining two different identical types, we should generate proto files from [go-to-protobuf](https://github.com/kubernetes/code-generator/tree/master/cmd/go-to-protobuf) and return the native type from our API.

jessesuen:
Kube has moved off of gogoproto. They now maintain:

https://github.com/kubernetes/gengo

which is used in:

https://github.com/kubernetes/code-generator/tree/master/cmd/go-to-protobuf

jessesuen:
Ah. I mistakenly thought Kube moved off of gogoproto, but I see they [have not](https://github.com/kubernetes/code-generator/blob/71364b5549ac6cce68b84a760fdd05174f6073da/go.mod#L8).

Hmmm, I agree we should not re-introduce gogoproto. Unfortunately, I'm at a loss on how to deal with the inconsistency that 
 is currently caused by redefining our types in our API. For example the omitempty inconsistency and the [enum problem](https://github.com/akuity/kargo/issues/707):

```yaml
apiVersion: kargo.akuity.io/v1alpha1
kind: Stage
metadata:
  name: staging
  generateName: ""     ####  native type would not return this because omitempty works properly
```

jessesuen:
[...]

Issue #707, opened 2023-09-12
bug: "enum" values don't look right in the ui


krancour:
I'm not sure if this is a UI problem or more of a problem with our protobuf definitions.

Here's a "live manifest" from a Stage:

![Screenshot 2023-09-11 at 8 09 26 PM](https://github.com/akuity/kargo/assets/1821014/9280f174-b119-439d-a926-d5e885a01bd3)

The `status` field _should_ read `Healthy`.

Comparing to the CLI, `Healthy` is what you would see, _but_ the CLI is doing type conversions from the protobuf types back to the same v1alpha1 types used by the controller. I suppose the UI is not doing that.

Do we need to add that to the UI, or are we just not using enums properly in the protobuf definitions?

cc @jessesuen @devholic 

Note: Disregard the other values that are all caps. They're junk related to a different PR I'm working on.

__Edit:__ I wonder if there's any real reason that can't be a string field over the wire. It would make type conversions easier and I feel it would be no less safe, because only controllers ever set status values anyway.

jessesuen:
I think the way we are getting live objects is flawed. We do this conversion of kubernetes types to gRPC types which has been proven error prone (e.g. forgetting to include `apiVersion` and `kind` in our proto types.

IMO, I wish we did not do this conversion in our API and would return the native kinds, **_as-is_**.

krancour:
@jessesuen and I spoke and it looks like https://github.com/kubernetes/gengo might be a replacement for gogoproto that could allow us to work with native types over the wire.

jessesuen:
Actually k8s hasn't moved off gogoproto. But we should do a stop-gap measure to return a string instead of a enum as you suggested. @devholic can you take this one?

krancour:
This is adequately addressed by #714 and there's a separate issue open (#708) for the rest of it. Closing.

PR #714, opened 2023-09-12
Set `EmitUnpopulated` to false; Use string for proto enums

devholic:
- https://github.com/akuity/kargo/issues/707
- https://github.com/akuity/kargo/issues/708


krancour:
Agree with this change... just test-driving it...

krancour (approved):
LGTM!

PR #729, opened 2023-09-12
docs: enable ui in quickstart

### LINKED ISSUE #717: Add section for starting the UI and using the admin user.
Defaults to requiring a port-forward, consider using options to expose on node port.

morey-tech:
The bare minimum to enable UI during the quick start. 
- Used new port `30084` for API nodePort to avoid changing example apps.
- I did not include steps to use the UI, other than to access it, to avoid the guide getting stale due to rapid changes to the UI.

Closes #717.


krancour (approved):

docs/docs/20-quickstart.mdx
```diff
@@ -137,9 +139,17 @@ helm install kargo \
...
+  --set api.service.type=NodePort \
+  --set api.service.nodePort=30084 \
+  --set 'api.adminAccount.password=admin' \
   --wait
 ```
 
+The Kargo dashboard will be accessible at
+[localhost:8084](http://localhost:8084).
+
+The Admin password is `admin`.
```
krancour, line 151: I don't care too much one way or another, but dummy SSO should work right out of the box, which may be even easier than this.

Either way, I'm good with it.

morey-tech: I prefer the consistency of both UIs having the admin user and password.

But for my curiosity, how would I enable the dummy SSO?

kencochrane (approved):

docs/docs/20-quickstart.mdx
krancour: I misspoke... it is not enabled by default, so please disregard. 🤦‍♂️

PR #736, opened 2023-09-13
feat: promote stages from freightline

rbreeze:
Screenshot 2023-09-12 at 17 35 57
Screenshot 2023-09-12 at 17 36 08


jessesuen:
Issue 1: I can't tell the content of what is about to be promoted

https://github.com/akuity/kargo/assets/12677113/f3387d2f-b312-458f-95f2-776f8205da77

I would say in the confirmation dialog, we should expand this much wider with full details, including the image, git repo, link, etc... similar to how we show that when just clicking on various freight.

jessesuen:
Issue 2: Now that we have this UI, let's remove the availableFreight tab completely in stage view.

image

jessesuen:
This is unrelated to the Promote button being added in this PR, but it has to do with Freightline in general so putting feedback here:

Issue 3: the default view of freightline should show more information:

image


In the above screenshot, each box just shows freight ID (not that useful) and the same two icons for all items in the line (also not that useful).I propose we expand this by default and get rid of the pinning feature. 

image

Clicking into a freight could still expand to show more metadata (git commit message, last seen, container repository).

jessesuen:
Issue 4: We lost the downstream subscribers affected by the promotion.

I liked the level of detail that we previously provided when we clicked the PromoteSubscribers button, where we listed the Stages that would be affected by the promote. If we implement my suggestion in Issue 1, we should have enough space to present the same message we used to.

rbreeze:
@jessesuen I've updated to address your comments, with some tweaks:

1. Instead of a popup, I simply expand the Freight box to include both the contents and the confirmation
2. Instead of writing which stages will be promoted, I highlight them in the DAG

CleanShot 2023-09-12 at 23 29 08@2x
CleanShot 2023-09-12 at 23 28 57@2x
CleanShot 2023-09-12 at 23 29 46@2x
CleanShot 2023-09-12 at 23 30 03@2x

jessesuen (approved):
I think that should be good enough for now. LGTM!

PR #742, opened 2023-09-13
feat: qualify all availableFreight in control-flow stages

jessesuen:
In my project, I have a control-flow Stage used to promote simultaneously to all downstream Stages. However, currently it requires I create promotion jobs to get freight into that history. It doesn't make sense for me to kick off a bunch of promotion jobs for control-flow stages that are essentially no-ops for getting freight into history. I simply want all availableFreight to be immediately become promotable to downstream.

This PR does the following:
1. A stage with no promotion mechanisms copies all availableFreight into history, and marks them all as qualified.
2. If, for some reason, a promotion job was run against a control-flow stage (which is technically something we should try and prevent in API), it will act as a no-op instead of muck around with the Stage's history.


alexmt (approved):

rbreeze (approved):
LGTM

PR #748, opened 2023-09-13
feat(charts): add generated readme to kargo

### LINKED ISSUE #718: Add section for the Helm chart values.
This section should be generated from the Helm chart values to avoid manual updates on each change.

morey-tech:
Closes #718. Refactor the Helm values file comments to support the [bitnami-labs/readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm) tooling. I have manually generated the readme to confirm it works. The next steps will be to automate the generation on releases and update a section in the docs with the content.
- Comments have been converted from multi-line to single-line to support the `@param` syntax.
- I've done some minor refactoring to the contents of the parameter descriptions, mainly to take advantage of markdown syntax for the readme.
- Some values have been uncommented instead of defaulting in the templates (e.g. the `api.replicas` was commented out but is now set to `1` the default`) so that they appear in the value column of the readme.

The only difference in the generated manifests using the default values compared to before these changes are to the admin account hash, which is due to the default value being empty quotes instead of placeholder text.
```diff
% diff helm-before.yaml helm-after.yaml                                                                                                                             
73,74c73,74
<   ADMIN_ACCOUNT_PASSWORD_HASH: $2a$10$5HiAsllXy3QjNAkjYChJSOPPD.z70o3eqL2CZ9/R0C8hIYzN.rw/.
<   ADMIN_ACCOUNT_TOKEN_SIGNING_KEY: "<signing key>"
---
>   ADMIN_ACCOUNT_PASSWORD_HASH: $2a$10$d1zy3B4xt65fjlaFwcRmeevGPp6ASl1X4p1lBrHjV32yXWC43vyvK
>   ADMIN_ACCOUNT_TOKEN_SIGNING_KEY:
732c732
<         secret/checksum: b425f7bd873326434d6e96c0ab9831ce25cd2e8c1f07a7c2355b4532ecf23cea
---
>         secret/checksum: 97bbedb624d101c522d2990c4f83cc2595d0ddbf1c6c63dfddbd0afbf8e65dc9
```

The readme can be generated using `% bash hack/helm-docs/helm-docs.sh`, assuming the user has already done `npm install -g @bitnami/readme-generator-for-helm`. This shouldn't be of much consequence as the process will be automated in the near future.


kencochrane:
It would be cool if we added a github action that checked if the helm chart changed at all in a PR, and if so it would run this command to regenerate the readme file, and also copy the readme over to the website docs to keep them up to date. This way it is fully automated and we don't have to do anything.

morey-tech:
> It would be cool if we added a github action that checked if the helm chart changed at all in a PR, and if so it would run this command to regenerate the readme file, and also copy the readme over to the website docs to keep them up to date. This way it is fully automated and we don't have to do anything.

That's my plan for the next steps. But I wanted to break down the change into smaller pieces. First, get the readme and script in. Then I'll work on the action to automate it.

kencochrane:
The changes look fine to me; the only concern is that it makes it harder to read the values file. Should we add a comment at the top of the values file to have them reference the README file for easier-to-read documentation, or do you think that is so obvious it doesn't need to be said?

kencochrane (approved):

morey-tech:
It depends on where the user lands. If they go looking for the chart (like I normally do for something like Argo CD) and they land in [the chart folder](https://github.com/akuity/kargo/tree/morey-tech-718-helm-docs/charts/kargo) they'll likely stop there and use the readme. If they go to inspect the value (`helm inspect values`) it won't be so obvious. To account for that case, I'll add a comment to the values directing to the readme.

Makefile
```diff
@@ -121,6 +121,8 @@ codegen:
 		paths=./...
 	pnpm --dir=ui install --dev
 	pnpm --dir=ui run generate:schema
+	npm install -g @bitnami/readme-generator-for-helm
```
morey-tech, line 124: I put the install for the readme-generator tool in the target instead of in `ui/package.json` since it's not a part of the UI.

krancour:
I promise I'll get around to looking at the actual values.yaml changes on Monday. The README that is generated from this is very nice though. 🚀

krancour:
@morey-tech are you able to deal with the merge conflicts?

krancour:
I resolved merge conflicts, but I bet I've gotta run code/doc gen now...

PR #749, opened 2023-09-13
feat(ui): more linkage

rpelczar:
https://github.com/akuity/kargo/assets/7262558/786cc2dd-c347-41e5-95bb-a5779c651d7d

### DIFF
--- a/ui/src/features/project/project-details/nodes/repo-node.tsx
+++ b/ui/src/features/project/project-details/nodes/repo-node.tsx
@@ -2,6 +2,8 @@ import { faDocker, faGit } from '@fortawesome/free-brands-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Tooltip } from 'antd';
 
+import { urlWithProtocol } from '@ui/utils/url';
+
 import { NodeType, NodesRepoType } from '../types';
 
 import * as styles from './repo-node.module.less';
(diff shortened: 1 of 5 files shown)


ui/src/features/project/project-details/nodes/repo-node.tsx
```diff
@@ -45,10 +47,10 @@ const RepoNodeBody = ({ label, value }: { label: string; value: string }) => (
   <>
     <div className={styles.label}>{label}</div>
     <Tooltip title={value}>
-      <div className={styles.value}>
+      <a href={urlWithProtocol(value)} className={styles.value} target='_blanl' rel='noreferrer'>
```
kencochrane, line 50: Should that be `target='_blank'`  instead of `target='blanl'` ?

rpelczar: good catch!

ui/src/features/stage/stage-actions.tsx
```diff
@@ -60,8 +63,51 @@ export const StageActions = ({ stage }: { stage: Stage }) => {
...
+  const argoCDAppsLinks = React.useMemo(() => {
+    const shards = Object.values(config?.argocdShards || {});
+
+    return shards.flatMap((shard) => {
+      return stage.spec?.promotionMechanisms?.argocdAppUpdates.map((argoCD) => ({
+        label: argoCD.appName,
+        url: `${shard.url}/applications/${shard.namespace}/${argoCD.appName}`
+      }));
+    });
+  }, [config, stage]);
```
jessesuen, line 76: When mapping ArgoCD apps to a shard, we can identify which shard based on the following:
```yaml
kind: Stage
metadata:
  annotations:
    kargo.akuity.io/shard: foo
```

In the normal case, the `kargo.akuity.io/shard` key might be missing entirely. Otherwise the value of the key should be one of the map keys in the GetConfig Argo CD URLs API

jessesuen (approved):

PR #773, opened 2023-09-15
various small fixes on resource types

### LINKED ISSUE #762: branch on git subs can be implicitly HEAD when not specified
Tripped on this trying to onboard a new kargo project

```
.writeBranch: Invalid value: "": spec.promotionMechanisms.gitRepoUpdates[0].writeBranch in body should be at least 1 chars long
```

writeBranch can be implicitly HEAD of the remote, if user doesn't supply it. This would cut down on our boilerplate and make examples more universally compatible.

krancour:


jessesuen:
> ReadBranch and WriteBranch should never be the same, so they cannot both be optional or the result will too often be that the both reference the same branch, which is an error condition. 

Don't we have a check for this?

> If only one can be optional, it should be ReadBranch

Agree.

The hiccup I encountered is when I wasn't even using git subscriptions (just image subscription). And so I did want to write to default branch of my git repo. e.g.:

```
kind: Stage
metadata:
  name: 05pct
  namespace: kargo-canary
spec:
  subscriptions:
    repos:
      images:
      - repoURL: ghcr.io/jessesuen/rollouts-demo
        updateStrategy: Latest
  promotionMechanisms:
    gitRepoUpdates:
    - repoURL: https://github.com/jessesuen/kargo-canary.git
      writeBranch: main
      kustomize:
        images:
        - image: ghcr.io/jessesuen/rollouts-demo
          path: base/05pct
    argoCDAppUpdates:
    - appName: kargo-canary
```

My question is: if we already have a check to ensure readBranch != writeBranch (although I'm not sure if we do), why not rely solely on that check, and relax the validation so that both fields can be optional to help in the case where git subscriptions are not even used?

krancour:
> Don't we have a check for this?

> why not rely solely on that check, and relax the validation so that both fields can be optional to help in the case where git subscriptions are not even used?

We do have that check, but but it happens during promotion. i.e. A little late in the game.

I'm trying to remember why it happens there instead of right up front in the validating webhook. I know I had a reason, but _if_ I could successfully move that check into the webhook, I think I could get it to where you want.

That said, I do still think that in the average case there is a benefit to making users mindful of where they are writing to. I think the cases where this is moot due to no git subscription are probably outliers since that's not real gitops.

But not a hill I'd die on. 😄

krancour:
I think this PR still contains valid improvements even if it _doesn't_ close #762.

jessesuen (approved):

jessesuen:
writebranch being optional is a nice to have and don't want to spend too much time on this, so I'll drop it for now. 
Implicit readBranch is definitely desired over implicit writeBranch 

> I'm trying to remember why it happens there instead of right up front in the validating webhook. I know I had a reason

My guess is that it's impossible to detect across many stages where the zero stage is reading and the later stage is writing. But when your dealing with the actual git repo during promotion, it's easy to check.

krancour:
> My guess is that it's impossible to detect across many stages where the zero stage is reading and the later stage is writing. But when your dealing with the actual git repo during promotion, it's easy to check.

Yes... that was it.

I'll still see what I can do, but won't spend a lot of time on it.

PR #783, opened 2023-09-18
keep all available freight qualified

krancour:
Whether a Stage's available Freight are marked as qualified or not is not supposed to matter. (And doesn't.) Qualification in an immediately upstream Stage is a _pre-req_ for making it _into_ that collection to begin with.

It can, however, spark confusion if available Freight are marked as unqualified. Is it available or isn't it?

This PR ensures that available Freight _always_ will appear qualified. They are.

The flip-side of this is that Freight is now explicitly _disqualified_ when it is promoted to be a Stage's _current_ Freight.

### DIFF
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -401,12 +401,11 @@ func (r *reconciler) promote(
 		return nil
 	}
 
-	var targetFreightIndex int
 	var targetFreight *kargoapi.Freight
-	for i, availableFreight := range stage.Status.AvailableFreight {
+	for _, availableFreight := range stage.Status.AvailableFreight {
 		if availableFreight.ID == freightID {
-			targetFreightIndex = i
 			targetFreight = availableFreight.DeepCopy()
+			targetFreight.Qualified = false
 			break
 		}
 	}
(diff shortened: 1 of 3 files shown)


internal/controller/promotions/promotions.go
```diff
@@ -444,7 +443,6 @@ func (r *reconciler) promote(
 		// control-flow stages in the first place)
 		if stage.Spec.PromotionMechanisms != nil {
 			status.CurrentFreight = &nextFreight
-			status.AvailableFreight[targetFreightIndex] = nextFreight
```
krancour, line 447: I can't see that there is any longer a need for this.

internal/controller/stages/stages.go
```diff
@@ -577,6 +577,7 @@ func (r *reconciler) getLatestFreightFromRepos(
 		Commits:   latestCommits,
 		Images:    latestImages,
 		Charts:    latestCharts,
+		Qualified: true,
```
krancour, line 580: This auto-qualifies any Freight that comes directly from repos.

rbreeze (approved):
LGTM

internal/controller/promotions/promotions.go
```diff
@@ -401,12 +401,11 @@ func (r *reconciler) promote(
...
 	}
 
-	var targetFreightIndex int
 	var targetFreight *kargoapi.Freight
-	for i, availableFreight := range stage.Status.AvailableFreight {
+	for _, availableFreight := range stage.Status.AvailableFreight {
 		if availableFreight.ID == freightID {
-			targetFreightIndex = i
 			targetFreight = availableFreight.DeepCopy()
+			targetFreight.Qualified = false
```
krancour, line 408: This disqualifies current Freight. The next Stage reconciliation will qualify it if the Stage is healthy.

PR #786, opened 2023-09-19
cli: add special table printer for stages

krancour:
Before:

```shell
kargo get stages --project kargo-demo            
NAME   AGE
uat    121m
test   122m
```

After:

```shell
kargo get stages --project kargo-demo
NAME   CURRENT FREIGHT                            HEALTH    AGE
test   b73f9d1afaca87254b64e64e5439557e86dcba79   Healthy   123m
uat    b73f9d1afaca87254b64e64e5439557e86dcba79   Healthy   121m
```

We need this if we're to have a CLI-centric quickstart. See #721

Long-term, idk about using this package for printing tables. It was nothing but pain.

### DIFF
--- a/internal/cli/get/get.go
+++ b/internal/cli/get/get.go
@@ -2,15 +2,18 @@ package get
 
 import (
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/slices"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/duration"
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/utils/pointer"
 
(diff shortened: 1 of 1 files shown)


internal/cli/get/get.go
```diff
@@ -104,6 +107,49 @@ kargo get stages my-project -o json
...
+		ColumnDefinitions: []metav1.TableColumnDefinition{
+			{Name: "Name", Type: "string"},
+			{Name: "Current Freight", Type: "string"},
+			{Name: "Health", Type: "string"},
+			{Name: "Age", Type: "string"},
+		},
+		Rows: make([]metav1.TableRow, len(items)),
+	}
+	for i, item := range items {
+		// This func is only ever passed Stages
```
rbreeze, line 130: Is this enforced somehow? If not, should we duplicate this comment at the start of the function?

rbreeze (approved):
LGTM, just a small comment

internal/cli/get/get.go
krancour: Technically, nothing enforces it, but the function name should hopefully be a dead giveaway.

PR #790, opened 2023-09-19
Add `get promotions` command to CLI

### LINKED ISSUE #789: cli doesn't support `get promotions`
This turns out to be a major impediment to the CLI-centric quickstart called for in #721.

API endpoints already exist for get/list Promotions, at least.

We would need these columns in the tabular representation of the response:

```shell
NAME                                               STAGE   FREIGHT                                    PHASE       AGE
test-to-d9a3e3e54b11b3e4a763e7cb8b1098089b567686   test    d9a3e3e54b11b3e4a763e7cb8b1098089b567686   Succeeded   55s
```

devholic:
Resolves #789


internal/cli/get/projects.go
```diff
@@ -0,0 +1,70 @@
...
+	"golang.org/x/exp/slices"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	typesv1alpha1 "github.com/akuity/kargo/internal/api/types/v1alpha1"
+	"github.com/akuity/kargo/internal/cli/client"
+	"github.com/akuity/kargo/internal/cli/option"
+	v1alpha1 "github.com/akuity/kargo/pkg/api/service/v1alpha1"
+)
+
+func newProjectsCommand(opt *option.Option) *cobra.Command {
```
krancour, line 18: This confused me at first. The name appears as if it could be responsible for the `kargo new project` command. If we named it `newGetProjectsCommand` it might avoid that confusion.

Same applies for the other new subcommands that are introduced.

internal/cli/get/promotions.go
```diff
@@ -0,0 +1,127 @@
...
+	Stage option.Optional[string]
+}
+
+func newPromotionsCommand(opt *option.Option) *cobra.Command {
+	flag := PromotionsFlags{
+		Stage: option.OptionalString(),
+	}
+	cmd := &cobra.Command{
+		Use:     "promotions --project=project [--stage=stage] [NAME...]",
+		Aliases: []string{"promotion"},
```
krancour, line 32: May I suggest `proto` and `protos` as additional aliases?

internal/cli/get/promotions.go
```diff
@@ -0,0 +1,127 @@
...
+					}
+				}
+			}
+			if err := printObjects(opt, res); err != nil {
+				return err
+			}
+			return resErr
+		},
+	}
+	option.OptionalProject(opt.Project)(cmd.Flags())
```
krancour, line 96: Should --project be _required_ instead of optional?

internal/cli/get/stages.go
```diff
@@ -0,0 +1,117 @@
...
+	return &metav1.Table{
+		ColumnDefinitions: []metav1.TableColumnDefinition{
+			{Name: "Name", Type: "string"},
+			{Name: "Current Freight", Type: "string"},
+			{Name: "Health", Type: "string"},
+			{Name: "Age", Type: "string"},
+		},
+		Rows: rows,
+	}
+}
```
krancour, line 117: This is nicer than how I had it.

Thanks for cleaning it up!

krancour:
Minor, non-blocking feedback. Pre-emptively approving.

krancour (approved):

PR #803, opened 2023-09-20
fix(docs): UI is now https & new error with expired ghcr.io auth

morey-tech:
In the move to v0.1.0, the UI is now https default so the link is updated. Also, a new error is shown when the user has expired ghcr.io auth for `docker`, so the tip is updated.


morey-tech:
Superseded by #808.

PR #804, opened 2023-09-20
fix(chart): tokenSigningKey null default results in outOfSync in Argo CD

morey-tech:
With the `api.adminAccount.tokenSigningKey` unset by default, in combination with [the template using `quote`](https://github.com/akuity/kargo/blob/c61e826b9e0a34369346c6999e3a4e55d48d2d3f/charts/kargo/templates/api/secret.yaml#L14), results in an Argo CD `Application` deploying the `kargo` chart to always show as out of sync.

Xnapper-2023-09-20-09 44 23

Two solutions to this are:
1. The user sets a static value for the `tokenSigningKey`
2. The chart sets the default value of the `tokenSigningKey` to `""`.

I've tested `2` by setting a parameter override for `api.adminAccount.tokenSigningKey` to `""` in the `Application`, which no longer shows out of sync after each sync.

<details><summary>This `Application` can be used to reproduce the issue.</summary>

```yaml
kind: Application
metadata:
  name: 'kargo'
spec:
  destination:
    name: 'in-cluster'
    namespace: 'kargo'
  source:
    chart: 'kargo'
    repoURL: 'ghcr.io/akuity/kargo-charts'
    targetRevision: 0.1.0
    helm:
      parameters:
      - name: "api.service.type"
        value: NodePort
      - name: "api.service.nodePort"
        value: '30084'
      - name: "api.adminAccount.password"
        value: admin
      # Uncomment to test the proposed solution.
      # - name: "api.adminAccount.tokenSigningKey"
      #   value: '""'  project: 'default'
  syncPolicy:
    syncOptions:
    - CreateNamespace=true
 ```
</details>

### CODE REVIEW

File: charts/kargo/values.yaml [hunk 1]
```diff
@@ -93,7 +93,7 @@ api:
     ## @param api.adminAccount.password A password for the admin account. Ignored if `passwordHash` is set. It is suggested that you generate this using a password manager or a command like: `openssl rand -base64 29 \| tr -d "=+/" \| cut -c1-25`.
     password: admin
     ## @param api.adminAccount.tokenSigningKey Key used to sign ID tokens (JWTs) for the admin account. It is suggested that you generate this using a password manager or a command like: `openssl rand -base64 29 \| tr -d "=+/" \| cut`
-    tokenSigningKey:
+    tokenSigningKey: ""
```
krancour, line 96: What do you think about using a default value like `change me`?

I don't think the API server can actually handle this being blank.

morey-tech: `change me` would likely have the same effect, so I'm on board. But I have tested with `tokenSigningKey: ''`, and the API server works for me.

krancour: Ooooh.

I just remembered that this exists: http://masterminds.github.io/sprig/flow_control.html

For things where we don't want to provide a default and we _insist_ on the user providing their own value, this could be helpful.

krancour:
Closed in favor of #829

PR #818, opened 2023-09-21
feat: ability to display image history in UI

rbreeze:
Stores on/off (per project) in localStorage

CleanShot 2023-09-21 at 11 55 50@2x


ui/src/utils/prefs.ts
rpelczar, line 1: Can we name this file use-local-storage so we know it is a hook?

ui/src/features/project/project-details/images.tsx
```diff
@@ -1,41 +1,138 @@
...
         }
-        stages.add(stage.metadata?.name as string);
+        stages[stage.metadata?.name as string] = {
+          opacity: 1,
+          backgroundColor: colors[stage.metadata?.uid as string]
+        };
       });
     });
-    return [images, getStageColors([...stages])];
+    return [images];
```
rpelczar, line 121: This array is unnecessary here.

rpelczar (approved):
I added two minor comments.

ui/src/features/project/project-details/images.tsx
rbreeze: Thanks for catching that 🙂

PR #824, opened 2023-09-21
rename image update strategies and remove image updater-related workarounds

### LINKED ISSUE #817: rename image update strategies to avoid confusion
Image Updater has already been down this road...

https://github.com/argoproj-labs/argocd-image-updater/pull/456 changes the names of some of the image update strategies used by Image Updater, for the better, but this change has not made it into the most recent release of Image Updater (v0.12.2 as of this writing).

That change was made expressly to mitigate confusion that resulted from the older names. "Does `latest` mean it just tracks the `latest` tag?" (No. It doesn't.)

We should start using the new names (to avoid the same confusion) even though this change hasn't made it into an Image Updater release yet.

We can just map the newer strategy names to the older ones before invoking Image Updater programmatically.

cc @christianh814 because I think the confusion over how each image update strategy actually works may have caused some difficulty for you recently.

krancour:
Fixes #817

cc @christianh814 I think this may clear up some confusion you encountered.


internal/images/images.go
```diff
@@ -29,13 +22,32 @@ func init() {
...
+	// TODO: Be on the lookout for an Image Updater release >= 0.13.0.
+	//
+	// Image Updater is changing the name of two update strategies to avoid
+	// confusion caused by their older names.
+	//
+	// This change can be seen here:
+	//   https://github.com/argoproj-labs/argocd-image-updater/pull/456
+	//
+	// This change has not yet made its way into an Image Updater release, but we
+	// want to support these new strategy names NOW to avoid the same confusion.
```
jessesuen, line 41: Not going to gate the PR, but we could import image updater from tip if we want the changes now.

Longer term, we need a way to untangle ourselves from argo-cd, gitops-engine, image-updater, even if it means copying and pasting the libraries that we want. But that's a whole other effort.

jessesuen (approved):

internal/images/images.go
krancour: > Not going to gate the PR, but we could import image updater from tip if we want the changes now.

That's not a bad idea, especially since there's [another change](https://github.com/argoproj-labs/argocd-image-updater/pull/559) that's been merged into image updater's main branch that we currently have [a workaround for](https://github.com/akuity/kargo/blob/main/internal/images/semver.go#L5-L16) while waiting for it to be released.

krancour: Using the latest from their `master` branch now, which enabled us to drop two workarounds.

> Longer term, we need a way to untangle ourselves from argo-cd, gitops-engine, image-updater, even if it means copying and pasting the libraries that we want. But that's a whole other effort.

💯

jessesuen (approved):

PR #828, opened 2023-09-22
make mutating webhook for promos set owner ref

### LINKED ISSUE #815: Promotion's owning Stage should be set by webhook
If the controller or the CLI creates a Promotion, it sets an owner reference to the applicable Stage.

This is an important detail because it plays a role in enqueing the owning Stage for reconciliation when a Promotion reaches a terminal phase.

But the owner references _does not_ get set if you create the Promotion with `kubectl`. I don't believe this is a _major_ problem, but for sure it delays Stage reconciliation.

We should set the owner reference in a webhook to ensure it's _always_ set.

This issue came to light via #811 and #813, so thanks to @tal-hason for helping to discover this.

krancour:
Fixes #815


jessesuen (approved):
Great improvement!

PR #829, opened 2023-09-22
chart: require admin password and token signing key to be set

krancour:
Alternative to #804 insofar as it addresses the same problem, but uses a different approach.

This also improves the security footing of the chart by _requiring_ users to select their own strong admin account password and token signing key.

This is better than setting a default, because too many users will take the default values, never realizing they've done anything wrong.

cc @morey-tech


charts/kargo/values.yaml
```diff
@@ -88,11 +88,11 @@ api:
   adminAccount:
     ## @param api.adminAccount.enabled Whether to enable the admin account.
     enabled: true
-    ## @param api.adminAccount.passwordHash Bcrypt password hash for the admin account. If specified, will ignore `api.adminAccount.password`.
+    ## @param api.adminAccount.passwordHash Bcrypt password hash for the admin account. If specified, will ignore `password`. A value MUST be provided for either this field or `password`.
```
morey-tech, line 91:
```suggestion
    ## @param api.adminAccount.passwordHash Bcrypt password hash for the admin account. If specified, will ignore `password`. A value **must** be provided for either this field or `password`.
```

charts/kargo/values.yaml
```diff
@@ -88,11 +88,11 @@ api:
...
     passwordHash:
-    ## @param api.adminAccount.password A password for the admin account. Ignored if `passwordHash` is set. It is suggested that you generate this using a password manager or a command like: `openssl rand -base64 29 \| tr -d "=+/" \| cut -c1-25`.
-    password: admin
-    ## @param api.adminAccount.tokenSigningKey Key used to sign ID tokens (JWTs) for the admin account. It is suggested that you generate this using a password manager or a command like: `openssl rand -base64 29 \| tr -d "=+/" \| cut`
+    ## @param api.adminAccount.password A password for the admin account. Ignored if `passwordHash` is set. It is suggested that you generate this using a password manager or a command like: `openssl rand -base64 29 \| tr -d "=+/" \| cut -c1-25`. A value MUST be provided for either this field or `passwordHash`.
```
morey-tech, line 93:
```suggestion
    ## @param api.adminAccount.password A password for the admin account. Ignored if `passwordHash` is set. It is suggested that you generate this using a password manager or a command like: `openssl rand -base64 29 \| tr -d "=+/" \| cut -c1-25`. A value **must** be provided for either this field or `passwordHash`.
```

charts/kargo/values.yaml
```diff
@@ -88,11 +88,11 @@ api:
...
+    password:
+    ## @param api.adminAccount.tokenSigningKey Key used to sign ID tokens (JWTs) for the admin account. It is suggested that you generate this using a password manager or a command like: `openssl rand -base64 29 \| tr -d "=+/" \| cut`. A value MUST be provided for this field.
```
morey-tech, line 95:
```suggestion
    ## @param api.adminAccount.tokenSigningKey Key used to sign ID tokens (JWTs) for the admin account. It is suggested that you generate this using a password manager or a command like: `openssl rand -base64 29 \| tr -d "=+/" \| cut`. A value **must** be provided for this field.
```

docs/docs/30-how-to-guides/10-installing-kargo.md
```diff
@@ -36,16 +36,28 @@ that Kargo is intended to be installed into the same cluster as the Argo CD
...
   oci://ghcr.io/akuity/kargo-charts/kargo \
   --namespace kargo \
   --create-namespace \
+  --set api.adminAccount.password=admin \
+  --set api.adminAccount.tokenSigningKey=iwishtowashmyirishwristwatch \
   --wait
 ```
 
+:::caution
+If deploying to an internet-connected cluster, be certain to do one of the
```
morey-tech, line 53: `internet-connected` is ambiguous here. Maybe `internet-facing`?
```suggestion
If deploying to an internet-facing cluster, be certain to do one of the
```

charts/kargo/values.yaml
morey-tech, line 95:
```
% openssl rand -base64 29 | tr -d "=+/" | cut 
usage: cut -b list [-n] [file ...]
       cut -c list [file ...]
       cut -f list [-s] [-w | -d delim] [file ...]
```
What is the intention of `| cut` on the suggested command?

morey-tech (approved):
Love it!

charts/kargo/values.yaml
krancour: I don't have super strong feelings about this, but all caps MUST is a conventional way of calling out absolute requirements in technical docs -- most often seen if RFCs.

morey-tech: I switched them to markdown bold to keep it consistent since I had done the same in the switch to the generated readme PR. So we need to choose our (the project's) preference

krancour: I didn't realize it was already the other way in other places. I'm fine with this suggestion then.

PR #845, opened 2023-09-25
chore: Lengthen admin acct token TTL

rbreeze:


krancour:
This was discussed in Slack the other day. Personally, I favor leaving this TTL small because we actually want to discourage over-leveraging the admin account. It's tantamount to working as `root` in Linux.

What we discussed the other day, and I am sorry I didn't make an issue for it, was leaving this default as-is, but overriding it in the quickstart scripts and in the Tiltfile.

That way it becomes less of a hassle for people who are either actively developing Kargo and people who are kicking the tires, but people who install to prod without reading the docs will still have a short TTL on the admin token to minimize the window for that token to be abused.

krancour (changes requested):
See comment.

krancour:
Thinking about this more... 24h isn't so bad. LGTM

We _might_ consider suggesting in docs that this be turned down for production installs.

krancour (approved):

krancour:
@rbreeze codegen needs to be run on this. `values.yaml` file changes update the chart's `README.md`.

PR #873, opened 2023-09-28
fix app syncing to account for syncpolicy/options

### LINKED ISSUE #874: Kargo doesn't instruct Argo CD to create namespace when option is set on the Application
I have a `Stage` that has auto promotion enabled in it's `PromotionPolicy`. So when I first deploy the Stage it tries to sync the Argo CD Application

```yaml
kind: Application
metadata:
  name: dev-app
  namespace: argocd
spec:
  destination:
    name: in-cluster
    namespace: dev-app
  project: default
  source:
    path: app
    repoURL: https://github.com/christianh814/kargo-demo-deploy
    targetRevision: env/dev
  syncPolicy:
    syncOptions:
      - CreateNamespace=true
```

However it FAILS

image

The error from Argo CD is "namespace not found"

image

When I use this app with "Vanilla" Argo CD, it works as expected.

krancour:
Fixes #874

Create namespace option (among others) wasn't being honored when Kargo programmatically syncs an App.

### DIFF
--- a/internal/controller/promotion/argocd.go
+++ b/internal/controller/promotion/argocd.go
@@ -159,9 +159,27 @@ func (a *argoCDMechanism) doSingleUpdate(
 	app.ObjectMeta.Annotations[argocd.AnnotationKeyRefresh] =
 		string(argocd.RefreshTypeHard)
 	app.Operation = &argocd.Operation{
+		InitiatedBy: argocd.OperationInitiator{
+			Username:  "kargo-controller",
+			Automated: true,
+		},
+		Info: []*argocd.Info{
+			{
+				Name:  "Reason",
+				Value: "Promotion triggered a sync of this Application resource.",
+			},
+		},
 		Sync: &argocd.SyncOperation{
 			Revisions: []string{},
 		},
+	}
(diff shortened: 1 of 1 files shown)


jessesuen (approved):

internal/controller/promotion/argocd.go
```diff
@@ -159,10 +159,28 @@ func (a *argoCDMechanism) doSingleUpdate(
...
 		},
 	}
+	if app.Spec.SyncPolicy != nil {
+		if app.Spec.SyncPolicy.Retry != nil {
+			app.Operation.Retry = *app.Spec.SyncPolicy.Retry
+		}
+		if app.Spec.SyncPolicy.SyncOptions != nil {
+			app.Operation.Sync.SyncOptions = app.Spec.SyncPolicy.SyncOptions
+		}
+	}
```
jessesuen, line 183: This looks consistent with autoSync() implementation of argo-cd controller, which is what we want.

krancour: Thanks @jessesuen!

PR #890, opened 2023-09-29
Update governance.yml to include kind/feature-request

kencochrane:
Updated governance.yml to include kind/feature-request

### DIFF
--- a/.github/governance.yml
+++ b/.github/governance.yml
@@ -11,6 +11,7 @@ issue:
         - enhancement
         - question
         - refactor
+        - feature-request
       multiple: true
       needs: true
 


krancour:
Thanks! Is it possible to have this tag applied automatically when someone selects the `Enhancement proposal` option when opening an issue?

Currently, it looks like that option adds the `enhancement` tag.

I suppose maybe `Enhancement proposal` could also be changed to match this label. i.e. Becomes `Feature request`.

krancour:
Nevermind... we can address those in a follow-up:

https://github.com/akuity/kargo/issues/893

LGTM

krancour (approved):

PR #908, opened 2023-10-02
ci(stale): add scheduled workflow

morey-tech:
On weekdays, this workflow will run once in the morning*. If an issue of `priority/low` or `priority/wontfix` has been inactive for 90 days, it will be marked as stale (by adding the `stale` label), and a comment will be left. If it remains inactive for another 30 days, it will be closed (but can be re-opened if needed, i.e. not locked).

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,32 @@
+name: 'Close Stale PRs'
+on:
+  schedule:
+    - cron: '32 11 * * 1-5'  # 11:32 UTC / 07:32 EDT / 06:32 EST from Mon-Fri.
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30  # Don't wait 6 hours to cancel workflow if it hangs.
+    steps:
+      - uses: actions/stale@v5
+        with:
+          stale-pr-label: 'stale'
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had any recent activity.
+            This issue will be closed if no further activity occurs but can be reopened if it becomes relevant again.
+          stale-pr-label: 'stale'
+          stale-pr-message: |
+            This PR has been automatically marked as stale because it has not had any recent activity.
+
+            - If this PR is not ready for review or implementation please [mark it as a draft](https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/).
+            - If this PR is ready for implementation but is intentionally on hold, please comment with a clear indication of why and add the `on hold` label to avoid it being marked as stale.
+
+            This PR will be closed if no further activity occurs but can be reopened if it becomes relevant again.
+          days-before-stale: 30
+          days-before-close: 7
+          exempt-draft-pr: true
+          exempt-pr-labels: 'on hold'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           stale-issue-label: 'stale'
           stale-issue-message: |
-            This issue has been automatically marked as stale because it has not had any recent activity. It will be closed if no further activity occurs but can be reopened if it becomes relevant again.
+            This issue has been automatically marked as stale because it had no activity for 90 days. It will be closed if no activity occurs in the next 30 days but can be reopened if it becomes relevant again.
           days-before-stale: -1  # Don't stale PRs.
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+          stale-pr-label: 'stale'
+          stale-pr-message: |
+            This PR has been automatically marked as stale because it has not had any recent activity.
+
+            - If this PR is not ready for review or implementation please [mark it as a draft](https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/).
+            - If this PR is ready for implementation but is intentionally on hold, please comment with a clear indication of why and add the `on hold` label to avoid it being marked as stale.
+
+            This PR will be closed if no further activity occurs but can be reopened if it becomes relevant again.
+          days-before-stale: 30
+          days-before-close: 7
```
morey-tech, line 30: I went a bit aggressive on the close/stale timings. I wonder for issues if we should exempt ones that have an `accepted` (or some other verb) so that they won't be considered stale

krancour: That's a really good idea.

I can think of a number of existing labels that would be a good indicator that we are committed to working on something, but if we want to start with the easiest thing and tweak it as necessary, what if we exempt anything labeled `priority/normal|high|urgent`.

If we do a decent job triaging issues, all of those probably reflect that at least one maintainer thinks the issue needs to be addressed.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30  # Don't wait 6 hours to cancel workflow if it hangs.
+    steps:
+      - uses: actions/stale@v5
+        with:
+          stale-pr-label: 'stale'
```
krancour, line 17:
```suggestion
          stale-issue-label: 'stale'
```

morey-tech: Good catch, fixed in https://github.com/akuity/kargo/pull/908/commits/7505c22e5eca20ba86829b3f85fb5608a94fb6e2

.github/workflows/stale.yml
morey-tech: That's an elegant solution; love it. Added in https://github.com/akuity/kargo/pull/908/commits/7505c22e5eca20ba86829b3f85fb5608a94fb6e2

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+
+            - If this PR is not ready for review or implementation please [mark it as a draft](https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/).
+            - If this PR is ready for implementation but is intentionally on hold, please comment with a clear indication of why and add the `on hold` label to avoid it being marked as stale.
+
+            This PR will be closed if no further activity occurs but can be reopened if it becomes relevant again.
+          days-before-stale: 30
+          days-before-close: 7
+          exempt-issue-labels: 'priority/low,priority/normal,priority/high,priority/urgent'
+          exempt-draft-pr: true
+          exempt-pr-labels: 'on hold'
```
morey-tech, line 33: `on hold` or `on-hold`? Or push people to use drafts instead of a label for PRs that are not ready?

krancour: Draft is probably good enough for now. We'll learn what's not working and tweak it later.

morey-tech:
Updated to only target issues, after 90 days.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30  # Don't wait 6 hours to cancel workflow if it hangs.
+    steps:
+      - uses: actions/stale@v5
+        with:
+          stale-issue-label: 'stale'
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had any recent activity. It will be closed if no further activity occurs but can be reopened if it becomes relevant again.
```
krancour, line 18: Nit: Should this mention the 90/30?

LGTM either wait.

krancour (approved):

.github/workflows/stale.yml
morey-tech:
```suggestion
            This issue has been automatically marked as stale because it had no activity for 90 days. It will be closed if no activity occurs in the next 7 days but can be reopened if it becomes relevant again.
```

PR #911, opened 2023-10-02
add Freight and Warehouse resource types

### LINKED ISSUE #631: Freight CRD
While working on the [QueryFreight API](https://github.com/akuity/kargo/pull/629), it made me realize that freight is just metadata that we are constantly carrying over from Stage to Stage. And so the implementation of QueryFreight, is a bit awkward. It iterates all Stages first to collect all freight in each Stage. Then it does a bunch of deduplication code to prevent duplicates from returning. Which btw, is not fool-proof: there's nothing to stop someone from editing their stage's freight contents and messing up the result. Principally, we consider freight as something immutable. 

For this reason, I started to wonder if we should start treating freights as a first-class resource in Kargo with its own CRD.

Here are some of the benefits I see:
1. Easier to query. The QueryFreight API would be improved to just deal with the CRs.
2. Freight would become more immutable. We would treat the Freight CR as something that should only ever be created and deleted (for GC), but never updated or patched. We could use kubernetes RBAC to ensure this is the case.
3. Lifecycle of Freight would be decoupled from lifecycle of stages. In other words, we could recreate a Stage without losing Freight.
4. We would reduce the amount of duplicate metadata in the Stage status, since details of the freight (e.g. image tag) would be kept in the CR. 
5. We could keep more freight history. Currently, we limit to 10 freight per Stage. But by putting it into a different CR, we could keep an indefinite amount of freight.
6. When it is a kubernetes object, we could be more creative on how to take advantage of that property. For example, we could:
    * use ownership references to indicate the freight is still in use and prevent garbage collection.
    * label and index freights with the way we intend to query against them
    * have other things besides Stages, deal with and reference Freight

krancour:
Fixes #631

Regrettably a large PR, so here's a guide to getting through it. Individual commits are narrowly scoped.

# Commits

## introduce freight resource type

* Adds top-level `Freight` resource type
* Moves several subordinate types from `stage_types.go` to `freight_types.go`, largely unchanged
* Adds and tests some `Freight`-related helpers that are used to DRY up other parts of the code base
* Adds a new `SimpleFreight` type for places (like a `StageStatus`) where we might want a representation of a `Freight` without _using_ the top-level `Freight` resource type

## introduce warehouse resource type

* Adds top-level `Warehouse` resource type (can be renamed later if we don't like "Warehouse")
* Moves several subscription-related subordinate types from `stage_types.go` to `warehouse_types.go`, largely unchanged
* Adds and tests some `Warehouse`-related helpers that are used to DRY up other parts of the code base
* Makes "zero `Stages`" subscribe to a `Warehouse` instead of expressing repository subscriptions directly

## add more helpers and tests for existing resource types

* Adds and tests helpers for existing resource types to help DRY up other parts of the code base

## corresponding changes to protobuf definitions

* These are minimal, but are certain to break the UI

## run codegen

* Enough said
* Ignore this whole commit

## introduce freight webhooks

* Defaults `Freight` name to match its ID (which is deterministically derived from its contents)
* Ensures it contains at least one commit, image, or chart

## introduce warehouse webhooks

* Ports subscription validation over logic from `Stage` webhooks, largely unchanged

## DRY up and improve tests for existing webhooks

* The new webhooks were making the older ones look bad; cleaned them up a so their implementations are all pretty consistent

## add useful new indexes

* Adds some new indices to the `kubeclient` package; used by new and updated reconcilers

Note: The `kubeclient` package is where we've _been_ putting these, but it's a little bit of a "junk drawer" and I plan to clean it up later.

## introduce warehouse reconciler

* Polls repos, creates new `Freight`
* Most functionality was ported over unchanged from the `Stage` reconciler

## update stage reconciler

* Almost completely re-written
* Finds new `Freight` by looking for... `Freight`; does not poll repos directly anymore
* Tested much, much more thoroughly than before

## update promotion reconciler

* Mostly incidental changes related to `SimpleFreight` being used in many placed where `Freight` was previously used (before it became a top-level resource type)

## corresponding chart updates

* Update webhook registrations and component permissions

## corresponding Tiltfile changes

* categorize new CRDs to keep the dashboard tidy

## corresponding api server changes -- NEED JESSE'S HELP

* This is the most "suss" part of the whole PR
* I couldn't quite get all of this working, in part because I didn't fully understand the role of some of the "query freight" endpoints. Thought this might be better left to @jessesuen 
* Had to disable a handful of tests

# Known Issues

* Breaks API server
* Breaks the UI
* "Zero `Stages`" subscribed directly to a `Warehouse` do not re-reconcile immediately after the `Warehouse` finds new `Freight`; I have ideas for how to fix this in a follow-up

# Other Follow-Up Items

* Fix API server
* Fix UI
* Add `Freight` and `Warehouse`-related API endpoints for 
* Add `Freight` and `Warehouse`-related CLI commands
* Make `Warehouse`s "refreshable"; i.e. "poll now"


api/v1alpha1/warehouse_types.go
```diff
@@ -0,0 +1,171 @@
...
+	Branch string `json:"branch,omitempty"`
+}
+
+// ImageSubscription defines a subscription to an image repository.
+type ImageSubscription struct {
+	// RepoURL specifies the URL of the image repository to subscribe to. The
+	// value in this field MUST NOT include an image tag. This field is required.
+	//
+	//+kubebuilder:validation:MinLength=1
+	//+kubebuilder:validation:Pattern=`^(\w+([\.-]\w+)*(:[\d]+)?/)?(\w+([\.-]\w+)*)(/\w+([\.-]\w+)*)*$`
```
moensch, line 77: We need to make we don't regress https://github.com/akuity/kargo/pull/915

krancour: Double-check me, but I think I copied this from #915 this morning.

moensch: :embarrassed: https://github.com/akuity/kargo/pull/931

krancour: No worries. I'm afk for a few days. Will look soon.

api/v1alpha1/stage_types.go
```diff
@@ -81,157 +67,46 @@ const (
...
-// Subscriptions describes a Stage's sources of material.
+// Subscriptions describes a Stage's sources of Freight.
 type Subscriptions struct {
-	// Repos describes various sorts of repositories a Stage uses as sources of
-	// material. This field is mutually exclusive with the UpstreamStages field.
-	Repos *RepoSubscriptions `json:"repos,omitempty"`
-	// UpstreamStages identifies other Stages as potential sources of material
+	// Warehouse is a subscription to a Warehouse. This field is mutually
+	// exclusive with the UpstreamStages field.
+	Warehouse string `json:"warehouse,omitempty"`
```
jessesuen, line 104: To future proof this, we may need to make this more like:

```
  warehouseRef:
    name: xxxx
```

api/v1alpha1/warehouse_types.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// WarehouseSpec describes sources of versioned artifacts to be included in
+// Freight produced by this Warehouse.
+type WarehouseSpec struct {
+	// Subscriptions describes sources of artifacts to be included in Freight
+	// produced by this Warehouse.
+	//
+	//+kubebuilder:validation:Required
+	Subscriptions *RepoSubscriptions `json:"subscriptions"`
```
jessesuen, line 41: Have we considered this alternative:

```golang
  Subscriptions []Subscription
```

with "oneof" like semantic for subscription: 

```golang
type Subscription struct {
	// Git describes subscriptions to Git repositories.
	Git *GitSubscription `json:"git,omitempty"`
	// Images describes subscriptions to container image repositories.
	Images *ImageSubscription `json:"images,omitempty"`
	// Charts describes subscriptions to Helm charts.
	Charts *ChartSubscription `json:"charts,omitempty"`
}
```

api/v1alpha1/stage_types.go
krancour: Just making a note here that when we spoke on Zoom, we agreed to think about this more and tackle it in a follow-up if we deem it necessary.

api/v1alpha1/warehouse_types.go
krancour: We just spoke on Zoom, but for posterity, it is the way it is because "oneof" semantics are currently hard to enforce without doing that validation oneself in a webhook. I'm willing to add that validation in order to get slightly cleaner-looking YAML, but we agreed to do it in a follow-up.

I will open a separate issue for this so we don't lose track of it.

api/v1alpha1/stage_types.go
krancour: I will open a separate issue for this so we don't lose track of it.

krancour: #955

api/v1alpha1/warehouse_types.go
krancour: #954

jessesuen (approved):
I reviewed mainly the spec (which I provided some comments on and agree changes can be done in a follow up). Don't want to block this any longer, so LGTM!

Great work!

krancour:
@jessesuen I am still fixing the API endpoints.

PR #915, opened 2023-10-03
Relax image repo and git repo regexes

### LINKED ISSUE #846: image subscription repoURL does not allow `.` in the image name
The current regular expression to validate `repoURL` only allows a dot (`.`) in the host part, but not the image name.

For example, this here should be a valid image name: `docker.io/someuser/some.repo/other-image`

Since a `.` can be used in any part of an image name, the regex should be updated:

https://github.com/akuity/kargo/blob/f3cc86511854ab1755f24ae705ea19b08c9750fa/api/v1alpha1/stage_types.go#L154-L159

See: https://docs.docker.com/engine/reference/commandline/tag/#description

> The path consists consists of slash-separated components. Each component may contain lowercase letters, digits and separators. A separator is defined as a period, one or two underscores, or one or more hyphens.

### LINKED ISSUE #870: relax regex for git repo urls
From #867

The regex for git repo URLs needs to be relaxed.

```
one or more objects failed to apply, reason: Stage.kargo.akuity.io "develop" is invalid: [spec.promotionMechanisms.gitRepoUpdates[0].repoURL: Invalid value: "https://gitea-http.gitea.svc/gotham/gotham-cd": spec.promotionMechanisms.gitRepoUpdates[0].repoURL in body should match '^((https?://)|([\w-]+@))([\w\d.]+)(:[\d]+)?/(.)$', spec.subscriptions.repos.git[0].repoURL: Invalid value: "https://gitea-http.gitea.svc/gotham/gotham-cd.git": spec.subscriptions.repos.git[0].repoURL in body should match '^((https?://)|([\w-]+@))([\w\d.]+)(:[\d]+)?/(.)$'].
```

moensch:
Fixes #846
Fixes #870

## Note
I do have some code for a unit test ready-ish to validate this regex, but it does use `go/ast` and `go/parser` and is a bit .. involved. I pushed Copilot to its limits :)

Holler if you'd like me to add it.

### before
image

### after
https://regex101.com/r/mjk9Ar/1
image

### before
image

### after
https://regex101.com/r/cnnffH/1
image

### DIFF
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -155,7 +155,7 @@ type ImageSubscription struct {
 	// value in this field MUST NOT include an image tag. This field is required.
 	//
 	//+kubebuilder:validation:MinLength=1
-	//+kubebuilder:validation:Pattern=`^(([\w\d\.-]+)(:[\d]+)?/)?[a-z0-9-]+(/[a-z0-9-]+)*$`
+	//+kubebuilder:validation:Pattern=`^(\w+([\.-]\w+)*(:[\d]+)?/)?(\w+([\.-]\w+)*)(/\w+([\.-]\w+)*)*$`
 	RepoURL string `json:"repoURL"`
 	// GitRepoURL optionally specifies the URL of a Git repository that contains
 	// the source code for the image repository referenced by the RepoURL field.
@@ -163,7 +163,7 @@ type ImageSubscription struct {
 	// revision of that source code that was used to build the image.
 	//
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:validation:Pattern=`^(((https?://)|([\w-]+@))([\w\d\.]+)(:[\d]+)?/(.*))?$`
+	//+kubebuilder:validation:Pattern=`^https://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
 	GitRepoURL string `json:"gitRepoURL,omitempty"`
 	// UpdateStrategy specifies the rules for how to identify the newest version
 	// of the image specified by the RepoURL field. This field is optional. When
@@ -263,7 +263,7 @@ type GitRepoUpdate struct {
 	// RepoURL is the URL of the repository to update. This is a required field.
 	//
 	//+kubebuilder:validation:MinLength=1
-	//+kubebuilder:validation:Pattern=`^((https?://)|([\w-]+@))([\w\d\.]+)(:[\d]+)?/(.*)$`
+	//+kubebuilder:validation:Pattern=`^https://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
 	RepoURL string `json:"repoURL"`
 	// ReadBranch specifies a particular branch of the repository from which to
 	// locate contents that will be written to the branch specified by the
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,8 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "sunday"
     commit-message:
       prefix: "chore(deps):"
 
@@ -14,6 +15,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+      day: "sunday"
     commit-message:
       prefix: "chore(deps):"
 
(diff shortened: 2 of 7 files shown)


api/v1alpha1/stage_types.go
```diff
@@ -155,15 +155,15 @@ type ImageSubscription struct {
...
+	//+kubebuilder:validation:Pattern=`^(([\w\d\.-]+)(:[\d]+)?/)?[a-z0-9-_\.]+(/[a-z0-9-_\.]+)*$`
 	RepoURL string `json:"repoURL"`
 	// GitRepoURL optionally specifies the URL of a Git repository that contains
 	// the source code for the image repository referenced by the RepoURL field.
 	// When this is specified, Kargo MAY be able to infer and link to the exact
 	// revision of that source code that was used to build the image.
 	//
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:validation:Pattern=`^(((https?://)|([\w-]+@))([\w\d\.]+)(:[\d]+)?/(.*))?$`
+	//+kubebuilder:validation:Pattern=`^((https?://)|([\w.-@]+@))([\w.-]+)(:[\d]+)?(/(.*))?$`
```
krancour, line 166: I think your new expression is a clear improvement over the old one, but I think there are some other issues that should be corrected while we're already touching this.

tbh, I probably copied and pasted this regex from somewhere else in the very early days of Kargo.

The system doesn't actually support http or ssh yet, so while parts of this need to be relaxed, the far left side needs to be tightened up until such time that we might add http or ssh support.

Here's what I propose: ^https://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$

Broken down:

| Expression Part | Explanation |
|-----------------|-------------|
| ^ | |
| https:// | HTTPS only |
| (\w+([\.-]\w+)*@)? | Optional token@ before domain name; must start and end with a word character |
| \w+([\.-]\w+)* | Domain name; must start and end with a word character |
| (:[\d]+)? | Optional port number |
| (/.*)? | Anything goes after the first / |
| $ | |

Screenshot 2023-10-04 at 11 27 02 AM

What do you think about this? It does retain your improvements.

Also applicable on line 266.

api/v1alpha1/stage_types.go
```diff
@@ -155,15 +155,15 @@ type ImageSubscription struct {
 	// value in this field MUST NOT include an image tag. This field is required.
 	//
 	//+kubebuilder:validation:MinLength=1
-	//+kubebuilder:validation:Pattern=`^(([\w\d\.-]+)(:[\d]+)?/)?[a-z0-9-]+(/[a-z0-9-]+)*$`
+	//+kubebuilder:validation:Pattern=`^(([\w\d\.-]+)(:[\d]+)?/)?[a-z0-9-_\.]+(/[a-z0-9-_\.]+)*$`
```
krancour, line 158: Yours is a clear improvement, but I think we can tighten this up further.

I propose: `^(\w+([\.-]\w+)*(:[\d]+)?/)?(\w+([\.-]\w+)*)(/\w+([\.-]\w+)*)*$`

Broken down:

| Expression Part | Explanation |
|-----------------|-------------|
| `^` | |
| `(\w+([\.-]\w+)*(:[\d]+)?/)?` | Optional host with optional port number  |
| `(\w+([\.-]\w+)*)` | Required token starts and ends with a word character; this could be image name, or could be "owner" if followed by more stuff |
| `(/\w+([\.-]\w+)*)*` | Any number of additional tokens separated by forward slash; each token must begin and end with a word character |
| `$` | |

Screenshot 2023-10-04 at 11 54 58 AM

What do you think about this? Although tighter overall, it does relax the things you were looking for, I think?

krancour:
@moensch this is _stellar_ and extemely thorough. ❤️  Changes to regular expressions are sometimes very hard to review. I love how easy you made it to see the before and after results.

I did make some suggestions because if we're going to touch these particular expressions anyway, I think there are other outstanding things that need to be addressed.

I'd like to know what you think. I especially want to know if you feel the proposed alternatives do indeed retain the "relaxation" you were out to get, which I think can be summarized as allowing both `\` and `.` as separators in more places.

api/v1alpha1/stage_types.go
moensch: I tried to get fancy squashing inside my branch so your comments got disassociated - but I accepted all your suggestions.

api/v1alpha1/stage_types.go
moensch: I tried to get fancy squashing inside my branch so your comments got disassociated - but I accepted all your suggestions.

krancour (approved):

PR #916, opened 2023-10-03
chore: Remove outdated node install script

moensch:
The node installation script used in `Dockerfile.dev` is no longer supported and introduced a 60s wait every time the dev image was built.

Upgrading to using new instructions per https://github.com/nodesource/distributions/blob/69a45587cd87bd8c700e40bb8a8160e0c28d71d8/README.md#new-update-%EF%B8%8F


krancour:
Thanks @moensch! I will take a closer look at this as soon as I've got a chance.

krancour (approved):
LGTM

Thanks @moensch!

PR #922, opened 2023-10-04
Update dependabot to run weekly

kencochrane:
Update Dependabot to run weekly on Sunday

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,8 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


rbreeze (approved):
LGTM

terrytangyuan (approved):

krancour:
_Separate from this_, we could think about whether we want dependabot suggesting major updates at all. Those are almost always going to break us and I think they're things we should opt-into instead of having them thrust upon us.

krancour (approved):

alexmt (approved):
thank you!

PR #924, opened 2023-10-05
feat(docs): Add Discord link to header

### LINKED ISSUE #923: Add link to discord server in docs
Please add our discord server link in header of our docsite

image

pabrahamsson:
resolves #923 

Not terribly familiar with how to generate these docs but this request seemed trivial enough.
I wasn't able to find instructions in the repo for how to build/generate the docs, so any pointers are very welcome.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -79,6 +79,11 @@ const config = {
             label: 'GitHub',
             position: 'right',
           },
+          {
+            href: 'https://discord.gg/dHJBZw6ewT',
+            label: 'Discord Community',
+            position: 'right',
+          },
         ],
       },
       footer: {


docs/docusaurus.config.js
```diff
@@ -79,6 +79,11 @@ const config = {
             label: 'GitHub',
             position: 'right',
           },
+          {
+            href: 'https://discord.gg/dHJBZw6ewT',
+            label: 'Discord',
```
christianh814, line 84: Label should be "Discord Community"

docs/docusaurus.config.js
```diff
@@ -79,6 +79,11 @@ const config = {
...
+            position: 'right',
```
christianh814, line 85: Not sure if it matters...I was thinking to position it "left" - but I'm okay with it on the right (not a hill I will die on)

christianh814 (changes requested):
See my comments. I think the only change I'd like is the label.

docs/docusaurus.config.js
krancour: The GitHub link is on the right, so I think this is exactly where it belongs. Next to it.

krancour (approved):

pabrahamsson:
> See my comments. I think the only change I'd like is the label.

Thanks @christianh814, I've udpated the PR based on your comment.

krancour:
Thank you @pabrahamsson!

PR #930, opened 2023-10-06
fix: better default color picking + don't change colors on new stages

rbreeze:
- Moves stages color map to React context
- Stores color map, per project, in browser Local Storage
- **If new stages are added, colors of previously existing stages do not change by default**
- Better color choices for stages by default (not necessary to add annotations to get pretty "rainbow" effect anymore")
- If new annotations are found, they override the existing map without changing it
- Option to "re-map" stage colors with a button click. This is good for two use cases:
      a. You add new stage(s) and it messes up your colors, and you don't mind changing the colors of existing stages
      b. You add new color annotation(s) to existing stages and want to re-map the colors of other stages to fit. Annotations are considered when re-mapping (if a stage is annotated with a color, no other stage will be assigned that color)


Marvin9 (commented):
Will there be unused data (in local storage) of projects as they get deleted?

rpelczar (approved):

rbreeze:
@Marvin9 as Rafal said, since there's no delete project option in the UI, I think it is okay to omit handling of the deletion case.

@rpelczar and I discussed, and ideally we will implement a better solution long term which provides consistent color mapping across all browsers; this case can be handled with this future improvement

PR #948, opened 2023-10-12
improve ci and release processes

krancour:
These same changes were applied to Kargo Render earlier today in the course of trying to work out permission issues that were partly caused by that repo's recent rename (from bookkeeper).

This allows us to delete some of our Action secrets in favor of a per-execution username/secret that GH Actions manages _for_ us. The updated versions of Actions we use are in order to take full advantage of that.


jessesuen (approved):

bikramnehra (approved):
LGTM!

PR #962, opened 2023-10-13
chore: use teams instead of individuals in CODEOWNERS

### LINKED ISSUE #959: our CODEOWNERS file has issues
I've kept wondering why global owners' reviews aren't counted as binding for doc changes...

Now I realize this is because the rule for `/docs/` _overrides_ the global rule rather than just _adding_ reviewers for that one dir.

Easy to fix.

alexmt:
Swapping individuals with teams in CODEOWNERS to avoid spamming everyone with emails on each PR.

Also closes #959

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,6 +1,6 @@
 # This file is described here: https://help.github.com/en/articles/about-code-owners
 
-* @alexmt @devholic @gdsoumya @jessesuen @kencochrane @krancour @rbreeze
+* @akuity/maintainers
 
 # Doc owners
-/docs/ @dhpup @morey-tech
+/docs/ @akuity/maintainers @akuity/doc-maintainers


CODEOWNERS
```diff
@@ -1,6 +1,6 @@
 # This file is described here: https://help.github.com/en/articles/about-code-owners
 
-* @alexmt @devholic @gdsoumya @jessesuen @kencochrane @krancour @rbreeze
+* @akuity/maintainers
 
 # Doc owners
-/docs/ @dhpup @morey-tech
+/docs/ @akuity/doc-maintainers
```
krancour, line 6:
```suggestion
/docs/ @akuity/doc-maintainers @akuity/doc-maintainers
```

If you do this, you will also end up closing #959

alexmt: make sense. Accepted

krancour (approved):

PR #971, opened 2023-10-17
fix fig step in release.yaml

krancour:
This _should_ fix the step that opens a PR with fig completion info upon release.

The old config was copied from https://github.com/withfig/push-to-fig-autocomplete-action#usage and appears to be incorrect.

New config comes from the fig doc site:

https://fig.io/docs/guides/integrating/push-to-autocomplete-action


gdsoumya (approved):
LGTM

PR #985, opened 2023-10-18
docs: link to stage rationale blog post

morey-tech:


docs/docs/15-concepts.md
```diff
@@ -10,7 +10,7 @@ This section covers important Kargo concepts.
 To eliminate confusion, Kargo avoids the term "environment" altogether in favor
 of something more precise: _stage_. The important feature of a _stage_ is that its
 name ("test" or "prod," for instance) denotes an application instance's
-_purpose_ and not its _location_.
+_purpose_ and not its _location_. [This blog post](https://akuity.io/blog/kargo-stage-not-environment/) discusses the rationale behind this.
```
krancour, line 13: The added sentence LGTM, but while we're touching this paragraph anyway, can we fix what I think is an obvious problem?

We jump right into "To eliminate confusion..." without any context at all explaining that something might have been confusing to begin with...

I think a cursory mention that the term "environment" can mean different things to different people would be a good lead-in.

christianh814: +1 to Kent, we can actually take the example from the blog too about this.

krancour (approved):

Issue #956, opened 2023-10-13
break go mod dependency on argo cd and gitops engine


krancour:
These dependencies are holding us back from upgrading more important transitive dependencies -- like Kubernetes.

This meta-issue tracks several other issues related to this:

- [ ] #951
- [ ] TBD...

PR #989, opened 2023-10-19
use a minimal subset of argo cd resource types

krancour:
This is part of #956

cc @jessesuen -- you said you've done this before, so your 👀 would be especially appreciated.


jessesuen (approved):
Yep, this was precisely what I was expecting.

PR #1001, opened 2023-10-20
add get freight command to cli

### LINKED ISSUE #991: add CLI commands to get/list Freight
This is a pre-req for #988 since a number of steps in the revised Quickstart will need to take advantage of these two commands.

We can use the QueryFreight API for these two commands, so this shouldn't be blocked by any API endpoints that aren't implemented yet.

We can deal with delete separately in a follow-up.

We're not ready to talk about create yet since, for now, Freight is created only by Warehouses.

krancour:
Fixes #991

This also:

* Makes the query freight API return actual Freight resources instead of `SimpleFreight` (which is a simplified representation of a Freight suitable for use in Stage status; this is how I should have done this to begin with back in #911).
* Fixes a bug in the query freight API where the list always came back empty when no group by clause is specified.


rbreeze (approved):
LGTM!

Issue #990, opened 2023-10-19
copy git url normalization logic from argo cd


krancour:
Part of #956

krancour:
Closed by #1002

PR #1002, opened 2023-10-21
feat: copy argocd func

georgettica:
to be less dependant on argocd, this was done

related to #990


krancour:
@georgettica thanks for doing this! I might end up moving the package later, but this is perfect for right now. Thanks again!

krancour (approved):

krancour:
@georgettica sorry for the delay. All set.

PR #1015, opened 2023-10-23
invoke kargo render via cli

### LINKED ISSUE #951: Perform branch rendering by invoking kargo-render as a CLI instead of library
# Proposed Feature

We should invoke `kargo-render` as a CLI as opposed to using it as a library. 

# Motivation

We want to eliminate our go.mod dependency to Argo CD / gitops engine, which is holding us back from upgrading kubernetes libraries. This is because `kargo-render` imports argo-cd/gitops-engine to leverage its manifest rendering.

# Suggested Implementation

Fork/exec `kargo-render` when it is necessary to render manifests to a branch.

krancour:
Fixes #951

Part of #956


rbreeze (approved):
LGTM

PR #1017, opened 2023-10-23
feat: unstable nightly builds

34fathombelow:
This PR will build container images and helm-charts daily from the main branch ( M-Fri).   The format for the unstable builds will be `v0.1.2-unstable-20231023`.

Unstable CLI binaries will come next with the documentation on how to use unstable builds for testing.  Once this PR is merged and a build is kicked off at the daily scheduled time of 6P.M. PST you can use the following to get started.

```yaml
helm install kargo \
oci://ghcr.io/akuity/kargo-charts-unstable/kargo --version 0.1.2-unstable-20231023 \
--namespace kargo \
--create-namespace \
--set image.repository=ghcr.io/akuity/kargo-unstable \
--set api.service.type=NodePort \
--set api.adminAccount.password=admin \
--set api.adminAccount.tokenSigningKey=iwishtowashmyirishwristwatch \
```


.github/workflows/release.yaml
```diff
@@ -36,9 +57,11 @@ jobs:
       id: meta
       uses: docker/metadata-action@v4
       with:
-        images: ghcr.io/akuity/kargo
+        images: ${{ steps.repo.outputs.repo }}
         flavor: latest=false
-        tags: type=semver,pattern={{raw}}
+        tags: |
+          type=semver,pattern={{raw}}
```
krancour, line 63: Should this line be applicable only when the event is a release?

.github/workflows/release.yaml
```diff
@@ -4,6 +4,8 @@ on:
   release:
     types: 
     - created
+  schedule:
+    - cron: '0 1 * * 1-5' # 6 P.M. PST
```
krancour, line 8:
```suggestion
    - cron: '0 13 * * 1-5' # 6 P.M. PST
```

I just want to take it for a spin... we can set it back to the original time after.

krancour:
@34fathombelow thank you for this! This all looks sensible to me. Couple Qs posed inline.

More general... I know you'd asked about the possibility of a separate workflow. I think I might prefer that only because it seems the behavior for the official release vs nightly are more divergent than we maybe anticipated initially. What's your take? If you prefer it as is, I'll go along with it.

.github/workflows/release.yaml
34fathombelow: It is not required, a scheduled [event](https://github.com/docker/metadata-action#priority-attribute) has a higher priority attribute which will cause this tag to be skipped.  I have tested the entire workflow in a private repo for testing.  I have confirmed this is working as expected.   We could be explicit and add a check if you want.

34fathombelow:
> @34fathombelow thank you for this! This all looks sensible to me. Couple Qs posed inline.
> 
> More general... I know you'd asked about the possibility of a separate workflow. I think I might prefer that only because it seems the behavior for the official release vs nightly are more divergent than we maybe anticipated initially. What's your take? If you prefer it as is, I'll go along with it.

I personally do not have a preference. I choose to include it in the release workflow to reduce the maintainability of a separate workflow.

krancour:
Gonna merge this and let'er rip at 13:00. That will validate it all and if we want to refactor a bit later, we can.

.github/workflows/release.yaml
```diff
@@ -4,6 +4,8 @@ on:
   release:
     types: 
     - created
+  schedule:
+    - cron: '0 13 * * 1-5' # 6 P.M. PST
```
krancour, line 8:
```suggestion
    - cron: '0 17 * * 1-5' # 5 P.M. UTC
```

I'm a bug dummy who forgot that these times are UTC. This should go off at the top of the coming hour.

PR #1023, opened 2023-10-24
fix nightly build cron expression

krancour:
This was modified earlier for testing purposes. Putting it back to the way we actually wanted it.

cc @34fathombelow


34fathombelow (approved):
LGTM, thank you!

PR #1024, opened 2023-10-25
feat: CLI unstable nightly builds

34fathombelow:
PR introduces nightly unstable builds for the CLI binaries.  Downloads will be similar to the Akuity CLI.  

```bash
curl -sSL -o kargo "https://dl.akuity.io/kargo-cli/$(curl -sL https://dl.akuity.io/kargo-cli/unstable.txt)/darwin/$(uname -m)/kargo"
````
Once implemented I'm more than happy to create a separate workflow for nightly builds if desired.  The two GitHub Secrets in this PR have already been added to the repo.  Documentation will follow after testing.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -200,6 +200,46 @@ jobs:
       run: |
         echo "hash-${{ matrix.os }}-${{ matrix.arch }}=$(sha256sum bin/kargo* | awk -F 'bin/' '{print $1 $2}'| base64 -w0)" >> "$GITHUB_OUTPUT"
 
+  publish-unstable-cli:
+    needs: publish-image
+    if: github.event_name != 'release'
+    runs-on: ubuntu-latest
+    container:
+      image: golang:1.21.3-bookworm
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+    - uses: actions/cache@v3
+      with:
+        path: /go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Build CLI
+      env:
+        GOFLAGS: -buildvcs=false
+        VERSION: ${{ needs.publish-image.outputs.unstable-version }}
+        GIT_COMMIT: ${{ github.sha }}
+        GIT_TREE_STATE: clean
+      run: make nightly-cli
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: .github/workflows/release.yaml [hunk 1]
```diff
@@ -200,6 +200,46 @@ jobs:
...
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Build CLI
+      env:
+        GOFLAGS: -buildvcs=false
+        VERSION: ${{ needs.publish-image.outputs.unstable-version }}
+        GIT_COMMIT: ${{ github.sha }}
+        GIT_TREE_STATE: clean
+      run: make nightly-cli
```
krancour, line 224: This is the only part I'm slightly little fuzzy on...

I'm wondering why we wouldn't use a matrix here, set env vars from that matrix, then use the existing `build-cli` target (e.g. same approach as the `publish-cli` job above, but just setting `VERSION` differently and pushing the binaries to a different location.

My sense is there is probably a sound reason for deviating from that and I'm not aware of it...

34fathombelow: That is a fair question.  Each time a binary is built we have to create a invalidation for `unstable.txt` and update it.  Rather than creating a invalidation 6 times during each scheduled build we will only be doing it once.   I wanted to make this less error-prone.

Compiling all 6 binaries at once should only take ~12mins.

krancour: > Each time a binary is built we have to create a invalidation for unstable.txt...

Like I said... I was guessing there was a reason I was overlooking.

Thanks for humoring me!

krancour: Ok... one idea...

As far as I gather, the purpose of the`unstable.txt` is just to name the latest unstable version?

I agree we only need to do that once instead of 6 times. Could it happen in a separate job that is downstream from `publish-unstable-cli` and that way `publish-unstable-cli` can be a little more similar to `publish-cli`?

34fathombelow: Correct the `unstable.txt` will allow a user to run `curl -sSL -o kargo "https://dl.akuity.io/kargo-cli/$(curl -sL https://dl.akuity.io/kargo-cli/unstable.txt)/darwin/$(uname -m)/kargo"` to pull the latest unstable CLI binary.

We could create a separate job to invalidate `unstable.txt` but it would not be DRY.  We also need to change the output to `-o bin/kargo-cli/${VERSION}/${GOOS}/${GOARCH}/${BIN_NAME}`

krancour (approved):

PR #1029, opened 2023-10-25
fix(ui): Update UI for Freight/Warehouse changes

rbreeze:


jessesuen:
I built and installed this, but it seems like i can't get UI to promote the freight even though it shows connected

https://github.com/akuity/kargo/assets/12677113/5ba1728b-6073-4d44-83c4-9d31c315c4c1

krancour (approved):
LGTM 🚀

PR #1031, opened 2023-10-26
add regex option for kargo.akuity.io/authorized-projects annotation. …

lrotim:
…Removes overhead if kargo projects follow specific name pattern

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -2,6 +2,7 @@ package credentials
 
 import (
 	"context"
+	"regexp"
 	"strings"
 
 	"github.com/argoproj/argo-cd/v2/applicationset/utils"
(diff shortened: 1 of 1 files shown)


krancour:
@lrotim thanks for this!

Before approving, I want to understand a little more about the usage pattern this enables.

One bit of background -- we don't anticipate the credential borrowing feature to be especially popular since Argo CD only ever needs read permissions and Kargo needs r/w. The principle of least privilege suggests they should each get their own credentials with the minimum permissions they each need. I mention this only because it's something we may eventually entertain removing.

Back to this specific feature...

This would allow credentials used by Argo CD to be borrowed by not one, but _multiple_ Kargo projects. I'm trying to understand what cases might require multiple Kargo projects to access the same repository.

On the other hand, it's obvious that this could be useful for credential _templates_ (e.g. creds for all respos starting `https://github.com/my-org`).

So I'm curious to understand what you intend to do and wonder if limiting this to credential templates might be the right thing to do.

lrotim:
@krancour thx for the reply.

I was looking in a way how to have credentials defined in one place and have them used by multiple kargo projects. It is somewhat more practical to maintain - especially if I can use the single creds template for multiple projects/repos. 

I kind of missed the "kargo secrets" as "borrowing argoCD creds" was in the example in the quickstart - sorry for missing that. 

I actually opened a new PR [here](https://github.com/akuity/kargo/pull/1041), and would like to hear your thoughts if we can move that one forward - as I don't like the idea of sharing creds with ArgoCD in the first place.

krancour:
Closing in favor of #1041 since I believe it gets us closer to #1048.

PR #1038, opened 2023-10-27
feat(promo): support long-lived promotions and improve reactiveness

jessesuen:
This change refactors the promo controller to support long lived promotions as well as make it more reactive to completions of other promos on the stage.

Changes:
1. PromoteFn is called during normal Reconcile() instead of background go routine
2. Reconcile() is a no-op for promos that are already terminal, or is not their turn to run
3. If no promotions for a stage is currently running, Reconcile() will only allow the highest priority promo to reconcile (based on creationTimestamp and name). All others will be a no-op
4. Reconcile() allows any Promotion that is already Running to continue to run. This set us up for long-lived promotions (to be implemented in future PR).
5. PriorityQueue has a Peek() function
6. PriorityQueue ignores nil objects if pushed
7. PriorityQueue is a no-op if the pushed object already exists in the queue. This allows Push() to idempotent because Reconcile() calls it repeatedly

The following video shows 20 promotions getting created via a bash command, and then the promo reconciler working through them one by one, from earliest to latest:

https://github.com/akuity/kargo/assets/12677113/fbe53b04-4146-40db-b9b6-3ddc2ebf019c

Ignore the fact that they error, that is expected.

### DIFF
--- a/internal/controller/promotions/promoqueues.go
+++ b/internal/controller/promotions/promoqueues.go
@@ -86,10 +86,10 @@ func (pqs *promoQueues) initializeQueues(ctx context.Context, promos kargoapi.Pr
 	}
 }
 
-// tryActivate tries to mark the given Pending promotion as the active one so it can reconcile.
+// tryBegin tries to mark the given Pending promotion as the active one so it can reconcile.
 // Returns true if the promo is already active or became active as a result of this call.
 // Returns false if it should not reconcile (another promo is active, or next in line).
-func (pqs *promoQueues) tryActivate(ctx context.Context, promo *kargoapi.Promotion) bool {
+func (pqs *promoQueues) tryBegin(ctx context.Context, promo *kargoapi.Promotion) bool {
 	if promo == nil || promo.Spec == nil {
 		return false
 	}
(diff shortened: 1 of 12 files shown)


krancour:
Super excited for this! Reviewing is my top priority on Monday!

internal/controller/promotions/promotions.go
```diff
@@ -90,29 +116,12 @@ func newReconciler(
...
 // move the current state of the cluster closer to the desired state.
 func (r *reconciler) Reconcile(
 	ctx context.Context,
 	req ctrl.Request,
 ) (ctrl.Result, error) {
-	// We count all of Reconcile() as a critical section of code to ensure we
-	// don't start reconciling a second Promotion before lazy initialization
-	// completes upon reconciliation of the FIRST promotion.
-	r.promoQueuesByStageMu.Lock()
-	defer r.promoQueuesByStageMu.Unlock()
```
krancour, line 114: TIL that multiple calls to `sync.Once` don't return until the one that's actually doing something returns. i.e. Built-in mutex! Nice!

internal/controller/promotions/promotions.go
```diff
@@ -158,205 +168,77 @@ func (r *reconciler) Reconcile(
...
-		*status = newStatus
+	logger = logger.WithFields(log.Fields{
+		"stage":   promo.Spec.Stage,
+		"freight": promo.Spec.Freight,
 	})
-	if updateErr != nil {
-		logger.Errorf("error updating Promotion status: %s", updateErr)
+	logger.Debug("executing Promotion")
+
+	// Update promo status as Running to give visibility in UI
```
krancour, line 196: Nit: This isn't only for UI purposes. It's also what allows this active promotion to continue being reconciled, so I would suggest removing the bit that make this look UI-specific:

```suggestion
	// Update promo status as Running
```

internal/controller/promotions/promotions.go
```diff
@@ -158,205 +168,77 @@ func (r *reconciler) Reconcile(
...
-			if err := kubeclient.PatchStatus(ctx, r.kargoClient, &promo, func(status *kargoapi.PromotionStatus) {
-				status.Phase = kargoapi.PromotionPhasePending
-			}); err != nil {
-				return errors.Wrapf(
-					err,
-					"error updating status of Promotion %q in namespace %q",
-					promo.Name,
-					promo.Namespace,
-				)
+	func() {
```
krancour, line 210: What's the reason for the anonymous function declaration here? As far as I can see, what's inside this block functions the same with or without it.

krancour: Nevermind. I get it now.

krancour: A comment explaining it might help, but otherwise 👍

internal/controller/promotions/promoqueues.go
```diff
@@ -0,0 +1,156 @@
...
+				"depth":     pq.Depth(),
+			}).Debug("Stage-specific Promotion queue initialized")
+		}
+	}
+}
+
+// tryActivate tries to mark the given Pending promotion as the active one so it can reconcile.
+// Returns true if the promo is already active or became active as a result of this call.
+// Returns false if it should not reconcile (another promo is active, or next in line).
+func (pqs *promoQueues) tryActivate(ctx context.Context, promo *kargoapi.Promotion) bool {
```
krancour, line 92: Nit: I think "begin" and "conclude" would be more easily understandable verbs on a promotion than activate/deactivate.

Although I suggest renaming those functions, I wouldn't go as far as to suggest renaming variables containing the word "active," as I think the meaning is clear enough in those cases. i.e. Rename functions for clarity, but leave the implementations as they are.

Also, not a hill I'd die on, so do what you will with this suggestion.

internal/controller/promotions/promoqueues.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	if activePromoName := pqs.activePromoByStage[stageKey]; activePromoName != "" {
+		// There is already an active promo. It's either this promo or someone else.
+		return activePromoName == promo.Name
+	}
+
+	// If we get here, the Stage does not have any Promotions Running against it.
+	// Now check if it this promo is the one that should run next.
+	first := pq.Peek()
```
krancour, line 125: I really like this approach and wish I'd thought of it from the start.

krancour:
@jessesuen with only a few non-blocking nits, this all looks good to me.

Want to make sure that I properly understand that the extent of the "long-running" promotion support here is that running promotions can be re-reconciled, whereas they were not before. That doesn't do much for as at the moment, but opens some doors. Have I got that right?

krancour (approved):

jessesuen:
> Want to make sure that I properly understand that the extent of the "long-running" promotion support here is that running promotions can be re-reconciled, whereas they were not before. That doesn't do much for us at the moment, but opens some doors. Have I got that right?

You have it exactly right. All of our promo mechanisms happen synchronously so it doesn't really help, yet.

The next PR may need to add concepts like Resume() to the promo interface, so that a promo that has already started, might do a different piece of work (e.g. check on the status of pull request) for the 2nd reconcile vs. the first reconcile (create the pull request).

internal/controller/promotions/promoqueues.go
jessesuen: I don't mind renaming. I chose activate because it's becoming active, but am fine with begin/conclude as well. Will make the change

internal/controller/promotions/promotions.go
jessesuen: I actually didn't know about `sync.Once` until this refactor and so I think I was the one doing the learning 😉 .

internal/controller/promotions/promoqueues.go
jessesuen: Renamed to `tryBegin` and `conclude`.

PR #1041, opened 2023-10-27
enable defining global credentials in configurable namespace(s)

lrotim:
When setting up credentials for multiple kargo projects it would be practical to be able to put it in kargos namespace and have it available for multiple kargo projects.

Much like is the case with how argoCD enables storing creds in "his" namespace


charts/kargo/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: kargo
 description: Kargo
 type: application
-version: 0.1.0
-appVersion: v0.1.0
+version: 0.1.1
+appVersion: v0.1.1
```
krancour, line 6: These should never be touched... the release process sets them correctly.

krancour:
@lrotim thanks for this!

I think a few of us maintainers will have to have a conversation about the security implications of this before making any decision.

krancour:
Took some time to consider this over the weekend... adding my own thoughts to the discussion here:

1. I understand the need, but...

1. This would weaken the tenancy model, so...

1. It should be opt-in at install-time, just like Argo CD credential borrowing, but...

1. I don't like the idea of it being in Kargo's own namespace. I think most operating systems and Kubernetes all provide precedent for a separation between system-space and user-space. From my point of view, Kargo's own namespace is system-space and shouldn't be touched by users. Argo CD does _not_ play by these rules, but imho, that was a mistake that we shouldn't blindly emulate, so...

1. I would propose that in addition to being opt-in, that operators should designate a specific namespace to be used expressly for this purpose (shared credentials) at install-time.

@lrotim would this meet your needs?

@jessesuen wdyt?

jessesuen:
The global credentials use case is valid, and I agree with the need to allow credentials that can be reused across projects. However, as Ken alluded to, this has a lot of security implications so I think the PR won't be able to go in as is.

> It should be opt-in at install-time, just like Argo CD credential borrowing, but...

The crux of the issue is: global credentials needs more controls. It won't be a simple true/false configuration but rather something like:

- https://github.com/my-org/* can be accessed by `*` kargo projects (this is probably the common case)

But it could also be:


> I don't like the idea of it being in Kargo's own namespace. I think most operating systems and Kubernetes all provide precedent for a separation between system-space and user-space. 

I don't see any issue with keeping the credentials in the kargo namespace, so long as the only user who can service/access these credentials kargo admin. Since this PR does not affect self-service

That said, we don't have to decide for the user. The implementation can/should be such that we don't explicitly require that they *must* reside in `kargo` namespace, but that we can reference the credentials explicitly. I suspect most people will choose to k


> Argo CD does not play by these rules, but imho, that was a mistake that we shouldn't blindly emulate, so...

Argo CD has a softer tenancy that required us to re-implement tenancy controls in API/controller. We've mostly avoided that by starting with Namespaces as projects, and how credentials are stored per project.

jessesuen:
I filed https://github.com/akuity/kargo/issues/1048 for credential sharing / global credentials. As it stands, the approach of looking to `kargo` and then project namespace is too simplistic for what we want, and I think it will be better to spend more time to discuss a more proper/flexibile design.

lrotim:
@krancour @jessesuen thx for the feedback. 

I understand this should take some consideration. I drafted a PR to be more concrete on the first steps.

I also considered pushing an update to move this from being explicitly kargo namespace, but figured I'd like to hear your feedback first and keep the changes to the minimum. 

I updated the PR to cover such behaviour now (not using kargo ns explicitly but allowing to define any number of namespaces for such purpose)

If you think that would be good enough "for now" with a TODO for wider considerations that would be great, but I understand if you  want to consider wider usecases and "wait" for a cleaner solution. 

Would be nice to have this as I'm trying to get my company to use kargo as early adopters. I've been tackling similar issues in the company and would prefer to use this as opposed to building "in-house". 

I've tried setting it up now, and current creds restrictions are a bit of a nuisance for the configuration I've setup, but we'll make due. I'll follow what you decide on this one

lrotim:
Also, I agree - there should be some "access control" to specify which projects might access them. I wanted to keep the changes to the minimum and see if you would consider an option to add this for "admins" to share the creds, and design it in more detail later down the roadmap

jessesuen:
@lrotim totally understand this current limitation is a big inconvenience. In https://github.com/akuity/kargo/issues/1048#issuecomment-1786080485 I likened the problem to the propagation of imagePullSecrets to different namespaces. I also mentioned that a workaround might be to use the same technique you are probably already using for imagePullSecrets, but for kargo credentials. Is that something that might work for you in the interim?

lrotim:
@jessesuen Yes, for short/midterm we already have a workaround that will do, but wanted to try and contribute here. 
I'm happy to contribute/update this PR meet the spec defined in your linked issue, please let me know if this is something you would prefer to keep on official maintainers - if that is the case, we can close this up and just leave the link to the issue

internal/credentials/credentials.go
```diff
@@ -67,14 +67,19 @@ type kubernetesDatabase struct {
 // carries out the important task of indexing Credentials stored in Kubernetes
 // Secrets by repository type + URL.
 func NewKubernetesDatabase(
+	sharedCredentialsNamespaces []string,
 	argoCDNamespace string,
```
krancour, line 71: Since there are multiple bits of configuration to include now, I would have a preference for starting to pass this config as an instance of a new type -- let's say `KubernetesDatabaseConfig`.

Something like this:

```golang
type KubernetesDBConfig struct {
	ArgoCDCredentialBorrowingEnabled bool     `envconfig:"ARGOCD_ENABLE_CREDENTIAL_BORROWING" default:"false"`
	ArgoCDNamespace                  string   `envconfig:"ARGOCD_NAMESPACE" default:"argocd"`
	SharedCredentialsNamespaces      []string `envconfig:"SHARED_CREDENTIALS_NAMESPACES" default:""`
}

func KubernetesDBConfigFromEnv() KubernetesDBConfig {
	cfg := KubernetesDBConfig{}
	return cfg
}
```

krancour: Also note that envconfig can parse comma-separated strings into a `[]string` for you. :smile:

internal/credentials/credentials.go
```diff
@@ -119,6 +124,39 @@ func (k *kubernetesDatabase) Get(
...
+					sharedCredsNamespace,
+					labels.Set(map[string]string{
+						kargoSecretTypeLabel: common.LabelValueSecretTypeRepoCreds,
+					}).AsSelector(),
+					credType,
+					repoURL,
+					true, // repoURL is a prefix
+				); err != nil {
+					return creds, false, err
+				}
```
krancour, line 157: Shouldn't we break out of the loop after finding credentials?

internal/credentials/credentials.go
```diff
@@ -119,6 +124,39 @@ func (k *kubernetesDatabase) Get(
 		); err != nil {
 			return creds, false, err
 		}
+
+		if secret == nil {
```
krancour, line 128: I think this whole block should be outdented one level from where it is. i.e. Close the previous `if secret == nil` block before this block.

krancour: Mainy for readability.

charts/kargo/templates/controller/configmap.yaml
```diff
@@ -15,6 +15,9 @@ data:
   {{- if .Values.kubeconfigSecrets.kargo }}
   KUBECONFIG: /etc/kargo/kubeconfigs/kubeconfig.yaml
   {{- end }}
+  {{- if .Values.controller.sharedCredentials.enabled }}
+  KARGO_SHARED_CREDENTIALS_NAMESPACES: {{ .Values.controller.sharedCredentials.namespaces }}
```
krancour, line 19: For better or worse, we don't prefix env vars with `KARGO_`, so let's change this to `SHARED_CREDENTIALS_NAMESPACES`, please.

cmd/controlplane/controller.go
```diff
@@ -142,6 +143,7 @@ func newControllerCommand() *cobra.Command {
 				argoClientForCreds = appMgr.GetClient()
 			}
 			credentialsDB := credentials.NewKubernetesDatabase(
+				strings.Split(os.GetEnv("KARGO_SHARED_CREDENTIALS_NAMESPACES", ""), ","),
 				os.GetEnv("ARGOCD_NAMESPACE", "argocd"),
```
krancour, line 147: See [this comment](https://github.com/akuity/kargo/pull/1041/files#r1394755140)

This would become something like:

```golang
credentialsDB := credentials.NewKubernetesDatabase(
	credentials.KubernetesDatabaseConfigFromEnv(),
	kargoMgr.GetClient(),
	argoClientForCreds,
)
```

krancour:
Ok... I agree with where this conversation went...

Let the "Kargo admin" pick namespace(s) for this purpose. If they pick `kargo`, fine -- that's their prerogative.

I think this PR moves the needle most of the way toward what #1048 is after and I think we should move ahead with it.

@lrotim you did a great job making this configurable.

I have a few nitpicky comments about implementation, but nothing major. Once they're set, let's merge this.

jessesuen:
OK I took a close look at this, and I agree that this PR gets us as close as we need for the foreseeable future.

Introducing a `SHARED_CREDENTIALS_NAMESPACES` env var that a kargo admin can set at the system level, will solve the common desire to have global credentials defined *once*. And I don't think we need to solve the shared credential problem between only two projects (and not globally).

I realize now that the `SHARED_CREDENTIALS_NAMESPACES`  env var is effectively the same as my proposal to have search paths as described in https://github.com/akuity/kargo/issues/1048:

> A kargo admin can choose to configure "search paths" of namespaces for which to find credentials (similar to shell PATH env var) at the global level. When set at a global level, the search paths will apply to all projects. Global search paths are naturally opt-in by virtue of setting the system-wide search path to something.

So I'm good with the approach. I think we just need to fix some of the PR comments and I think this can be gtg!

charts/kargo/templates/controller/configmap.yaml
jessesuen: I wonder if we should differentiate between what I'm calling "global" credentials and a different/future feature to share credentials between projects, not at the global scope. 

So I'm proposing we name this `GLOBAL_CREDENTIALS_NAMESPACES` since that's what this PR will enable.

krancour:
@lrotim if you're time-constrained, we're happy to take this PR the rest of the way and give you the credit. You did most of the heavy lifting, after all. Just let us know.

lrotim:
@krancour Thanks for the detailed feedback. I've had a working weekend, so I didn't have time to get back to this.

I wanted to touch upon this on the weekend and update the PR, taking comments into consideration due Monday. 

If you want to move this faster than that, I don't mind if you wanna wrap it up, will check on it on Saturday.

krancour:
@lrotim I was only touching base as a courtesy in case you had a ton of other stuff on your plate. There's no rush.

lrotim:
@krancour @jessesuen when you have the time, please check out if you find that changes are inlined with the comments

krancour:
@lrotim I'm out sick today, but will have my 👀 on this first thing tomorrow. I can't say enough how much we appreciate this.

charts/kargo/values.yaml
```diff
@@ -177,6 +177,13 @@ controller:
   ## @param controller.enabled Whether the controller is enabled.
   enabled: true
 
+  ## All settings relating to shared credentials (used across multiple kargo projects)
+  globalCredentials:
+    ## @param controller.globalCredentials.enable Whether to enable the controller to look for credentials stored in specific namespaces used to share credentials.
+    enabled: false
+    ## @param controller.namespaces List of namespaces to look for shared credentials as CSV (comma separated values).
+    namespaces: ""
```
krancour, line 185: @lrotim, I think it would feel more natural for this to be a list of strings instead of a single, comma-delimited string.

cmd/controlplane/controller.go
```diff
@@ -133,16 +132,15 @@ func newControllerCommand() *cobra.Command {
 				}
 			}
 
-			var argoClientForCreds client.Client
+			credentialsDbOpts := make([]credentials.KubernetesDatabaseOption, 0)
 			if types.MustParseBool(
 				os.GetEnv("ARGOCD_ENABLE_CREDENTIAL_BORROWING", "false"),
 			) {
-				argoClientForCreds = appMgr.GetClient()
+				credentialsDbOpts = append(credentialsDbOpts, credentials.WithArgoClient(appMgr.GetClient()))
```
krancour, line 139: I don't think the client should be treated as an option, because it's 100% required. Could we have two args: client and opts?

krancour: Sorry... I mis-spoke here.

The "Kargo client" is 100% required. This one is optional, depending on whether credential borrowing is enabled.

So... imho, treat the "Kargo client" as required and leave the "Argo client" in the realm of opts.

lrotim: In practice, I've seen a lot of "opts" usage for both required and "optional args".  

But makes sense if you wanna make this explicit

krancour (approved):

Issue #977, opened 2023-10-17
add get/list/refresh Warehouses and create Freight endpoints


krancour:
Follow up to #911

These are needed to get the UI in good working order again.

iirc, @rbreeze you wanted this issue?

rbreeze:
Yep, happy to take it!

krancour:
@rbreeze note the edit to the title... create Freight not needed at this time.

PR #1055, opened 2023-11-02
feat: add CRUD/Refresh `Warehouse` endpoints

devholic:
Resolves https://github.com/akuity/kargo/issues/977

### DIFF
--- a/internal/api/create_warehouse_v1alpha1.go
+++ b/internal/api/create_warehouse_v1alpha1.go
@@ -25,12 +25,6 @@ func (s *server) CreateWarehouse(
 			return nil, connect.NewError(connect.CodeInvalidArgument, errors.Wrap(err, "invalid yaml"))
 		}
 	case req.Msg.GetTyped() != nil:
-		if err := validateProjectAndWarehouseName(
-			req.Msg.GetTyped().GetProject(),
-			req.Msg.GetTyped().GetName(),
-		); err != nil {
-			return nil, err
-		}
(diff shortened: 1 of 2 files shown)


jessesuen:
Could you add a Refresh endpoint and controller implementation similar to how stages can be refreshed? This will be necessary to detect new images/git commits, etc...

See:
https://github.com/akuity/kargo/blob/958747dcf0b537cdbea14ee0ac8fb6ac8dda8dd2/api/v1alpha1/stage_helpers.go#L37-L57

internal/api/create_warehouse_v1alpha1.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+func (s *server) CreateWarehouse(
+	ctx context.Context,
+	req *connect.Request[svcv1alpha1.CreateWarehouseRequest],
+) (*connect.Response[svcv1alpha1.CreateWarehouseResponse], error) {
+	var warehouse kargoapi.Warehouse
+	switch {
+	case req.Msg.GetYaml() != "":
+		if err := yaml.Unmarshal([]byte(req.Msg.GetYaml()), &warehouse); err != nil {
```
gdsoumya, line 24: Minor nitpick, do you think we should also validateProjectAndWarehouseName here to keep it consistent with the other case? The yaml could be missing the name and project too right?

gdsoumya (changes requested):

gdsoumya (approved):
LGTM

PR #1058, opened 2023-11-04
image updater replacement

### LINKED ISSUE #958: develop our own image repo subscription logic...
... or copy from Image Updater.

Image Updater had dependencies on Argo CD and GitOps Engine that we are trying to eliminate from Kargo.

This is part of #956

### LINKED ISSUE #1149: Warehouse image updateStrategy: NewestBuild not pulling the Tags curretly
# Checklist

* [v] I've searched the issue queue to verify this is not a duplicate bug report.
* [v] I've included steps to reproduce the bug.
* [v] I've pasted the output of `kargo version`.
* [v] I've pasted logs, if applicable.

# Description

When creating a warehouse with an image subscription with the updateStrategy = NewestBuild

the image tags that are arriving from the registry are not updated.

# Screenshots

the image list from the registry:
image

The image that the warehouse received:
image

# Steps to Reproduce

Create a warehouse with a subscription:

```YAML
kind: Warehouse
metadata:
  name: gotham
spec:
  subscriptions:
      - image:
          repoURL: quay.io/gotham/batman
          updateStrategy: NewestBuild
      - image:
          repoURL: quay.io/gotham/joker
          updateStrategy: NewestBuild
      - image:      
          repoURL: quay.io/gotham/robin
          updateStrategy: NewestBuild
      - git:
          repoURL: https://github.com/gotham-demo/gotham-cd.git
          branch: main
```

# Version

```
V0.2.1
```

# Logs

```
Controller log:
[kargo-controller-798f754644-f4nn4-controller.log](https://github.com/akuity/kargo/files/13412135/kargo-controller-798f754644-f4nn4-controller.log)

```

krancour:
Fixes #958

Fixes #1149

Part of #956

This large PR is organized in a series of narrowly-scoped commits to make the review easier.

1. create a replacement for image updater

    For a variety of reasons, I didn't want to just copy image updater code to our repo. I've developed a leaner, meaner, alternative that is heavily inspired by Image Updater, but one I hope will be less burden to maintain in the long run. This commit is the bulk of this PR, but it's all new code and isolated to just one package. Unit test coverage is high. Integration tests are included also, but disabled by default because they talk to Docker Hub and GHCR and are likely to fail due to rate limiting unless authenticated.

2. make linter cover ignored tests

    With new integration tests hidden behind a build flag, the linter required some adjusting to still cover those tests.

3. remove references to image updater

4. resource type updates

    There are breaking changes in here. I realized how `ImageUpdateStrategy` came directly from Image Updater, but doesn't make sense in the context in which we use it. Warehouses subscribe to repositories. Those subscriptions _select_ tags, but they play no role in updating anything. That is the purview of a Stage's promotion mechanisms. Renaming to `ImageTagSelectionStrategy` seemed to make more sense and this seemed like the right time to do it.

    The "alphabetical" strategy has also changed to the "lexical" strategy. I think the majority of the time, someone using this strategy is trying to select the latest _timestamped_ image. Comparisons of two similar tags with different trailing timestamps hinges on comparison of numeric characters, so "lexical" just seemed to be a more accurate descriptor of what the strategy does.

5. proto definition update

    Similar updates to previous item.

6. run codegen

7. corresponding controller updates

    Use the new `image.TagSelector` interface.

8. corresponding api server changes

    Just changes to type conversion functions.

9. corresponding webhook changes

    Use new `image` package (instead of Image Updater) to validate platform constraints.

10. ~~corresponding ui changes~~

    ~~Minimal. Just one field name.~~
    
    None required after recent rebase.

11. go mod tidy

    Loses dependencies on Image Updater.


krancour:
Due to its size, I'm amending this with some notes to help reviewer(s):

## Tag Selection Strategies

Renamed from "image update strategies," which makes less sense in our context.

As before, there are four strategies for finding new tags. There implementations are equivalent to what is in Image Updater, but are quite a bit more straightforward now.

1. `Lexical` (formerly `Alphabetical`)

    1. List all tags; return nil if none are found
    1. Remove ones that don't match an optional regex OR are explicitly ignored
    1. Return nil if no tags remain 
    1. Sort remaining tags by name, in lexically descending order 
    1. Select the 0 element
    1. Retrieve corresponding manifest
    1. Return nil if the selected tag doesn't meet optional platform constraints
    1. Return tag name + digest

1. `NewestBuild`

    1. List all tags; return nil if none are found
    1. Remove ones that don't match an optional regex OR are explicitly ignored
    1. Return nil if no tags remain 
    1. Retrieve corresponding manifests for all remaining tags in order to learn creation date (this is very chatty with the registry -- see notes about caching)
    1. Sort remaining tags by creation date, newest first
    1. Select the 0 element
    1. Return nil if the selected tag doesn't meet optional platform constraints
    1. Return tag name + digest

1. `SemVer`

    1. List all tags
    1. Remove ones that:
        1. Don't match an optional regex
        1. Are explicitly ignored
        1. Do not look like a semantic version
        1. Aren't matched by an optional version constraint
    1. Return nil if no tags remain 
    1. Sort them by semantic version, in descending order. Ties when sorting (e.g. 1.0 == 1.0.0) are broken lexically.
    1. Select the 0 element
    1. Retrieve corresponding manifest
    1. Return nil if the selected tag doesn't meet optional platform constraints
    1. Return tag name + digest

1. `Digest`

    Useful if you insist on subscribing to a mutable tag like "latest." (Although I don't endorse this.)
    
    See related issue #1078 
    
    1. List all tags
    1. Select the one specific tag of interest
    1. Retrieve corresponding manifest
    1. Return nil if the selected tag doesn't meet optional platform constraints
    1. Return tag name + digest

## Other Notes

* Custom transport (`http.RoundTripper`) is used to enforce rate limits client-side on a per-registry basis.

* Cache is used to avoid unnecessary calls to the registry. Unlike Image Updater, our cache is keyed _only_ off of image digests. i.e. We're only ever caching anything where the value is guaranteed never to differ in the registry over time. imho, it is a mistake that Image Updater did not implement their cache this way.

* Cache has a hard-coded TTL of 30 minutes on every entry. It would be unusual for more than this period of time to elapse between execution of the tag selection logic for any one warehouse, so cache hits will be common and misses will be infrequent. Over time as each Warehouse selects newer tags/digests, queries for older manifests will naturally cease and entries will be evicted due to TTL elapsing.

jessesuen (approved):

krancour:
Rebased and forced pushed to resolve go mod dependency conflicts.

No other changes since approval.

PR #1085, opened 2023-11-09
Upgrade the docs to Docusaurus version 3.0.0

kencochrane:
This upgrades the docs to version 3.0.0. I'm not sure if it is fully working correctly. The "Allow cookies" banner doesn't seem to be working, so we will need to fix that before we merge. @krancour maybe you can take a look and see if you know what is causing the issue.


kencochrane:
Some changes are required to Netlify; I'm trying to figure out how to make this. @krancour, is that something you can help point me in the right direction?

krancour (approved):

PR #1088, opened 2023-11-11
feat: add refresh warehouse CLI and controller functionality

jessesuen:
1. Controller will clear `kargo.akuity.io/refresh` annotation during reconciliation
2. Added WatchWarehouses API endpoint
3. Added `refresh warehouse` and `refresh stage` CLI commands

```
$ kargo refresh warehouse -p kargo-guestbook guestbook
warehouse 'kargo-guestbook/guestbook' refreshed
```

4. Added `option.ExactArgs()` and `option.MinimumNArgs()` helper to print help when the exact number of CLI args is not supplied. e.g.:

```
$ kargo login
Usage:
  kargo login server-address [flags]

Examples:
kargo login https://kargo.example.com --sso

Flags:
  -a, --admin             Log in as the Kargo admin user; mutually exclusive with --kubeconfig and --sso
  -h, --help              help for login
  -k, --kubeconfig        Log in using a token obtained from the local Kubernetes configuration's current context; mutually exclusive with --admin and --sso
  -P, --password string   Specify the password for non-interactive admin user login; only used with --admin
  -p, --port int          Port to use for the callback URL; 0 selects any available, unprivileged port; only used with --sso
  -s, --sso               Log in using OpenID Connect and the server's configured identity provider; mutually exclusive with --admin and --kubeconfig

Global Flags:
      --insecure-skip-tls-verify   Skip TLS certificate verification
      --local-server               Use local server

Error: accepts 1 arg(s), received 0
```


krancour:
@jessesuen before I go too deep reviewing details, I have a slight concern about CLI and UX. This PR doesn't introduce the issue, but it does compound it...

For the most part, @devholic structured the CLI commands as `kargo <verb> <type> ...` and I believe this was specifically to emulate `kubectl` to capitalize on user familiarity with that CLI.

This PR adds a command of the type `kargo <type> <verb> ...`. We have a few others like that, notably `kargo stage <verb> ...`.

I'm concerned about the two divergent approaches complicating the UX.

Without correcting the _existing_ issues (which could be done in a follow-up) what do you think about structuring the new command as `kargo refresh warehouse ...`?

I think `refresh` is a verb that applies to things besides warehouses (even if we haven't implemented the commands yet). `kargo refresh stage` is one that comes to mind.

wdyt?

krancour (approved):
Excellent!

PR #1091, opened 2023-11-11
make warehouse do shallow, single-branch clones

### LINKED ISSUE #1090: idea: make warehouse controller faster by doing shallow clone
Right now, the warehouse controller clones repos on every reconcile just to get the last commit ID _and message_. This is wildly inefficient.

We could use `git ls-remote` instead to get the ID, but we'd lose the message...

If we have to clone, we can still make this faster without losing functionality by doing a shallow clone (depth = 1) of a single branch (HEAD unless otherwise specified).

krancour:
Fixes #1090


rbreeze (approved):
LGTM

internal/controller/git/git.go
```diff
@@ -95,6 +95,19 @@ type repo struct {
...
+type CloneOptions struct {
+	// Branch is the name of the branch to clone. If not specified, the default
+	// branch will be cloned.
+	Branch string
+	// SingleBranch indicates whether the clone should be a single-branch clone.
+	SingleBranch bool
+	// Shallow indicates whether the clone should be with a depth of 1. This is
+	// useful for speeding up the cloning process when all we care about is the
+	// latest commit from a single branch.
+	Shallow bool
```
jessesuen, line 108: Since I only see CloneOptions used one place, is it true that this PR makes it such that all clones performed by kargo are shallow?

jessesuen (approved):
LGTM

Issue #1076, opened 2023-11-07
add api server endpoint for renaming/re-labeling freight


krancour:
Follow-up to #1072

Note that this isn't simply an "update Freight" endpoint because Freight are otherwise immutable.

krancour:
#1162 addressed this.

PR #1097, opened 2023-11-12
add alias label to freight

### LINKED ISSUE #1072: add mutable name or label attribute to freight
I've thought for some time that it would be convenient for users to be able to apply human-friendly names to pieces of freight that they deem important. This would be more convenient than always referring to individual pieces of freight by their id (which is a sha1 digest).

Talking with @rbreeze yesterday, we also were thinking it would be convenient if pieces of freight had more human-friendly names/labels _by default_ as well. We like the Docker strategy of generating cute names for running containers, like "frozen-wookie." There are many libraries available for generating such names.

So this issue calls for the addition of a new attribute and for that attribute to default to a generated, but human-friendly name that can be changed/overridden later.

API endpoints, CLI commands, and UI features for actually executing the re-naming/re-labeling can come in a follow-up.

krancour:
Fixes #1072


jessesuen (commented):
Questions:
* is it possible for a Freight to have 2+ alias's similar to how a git commit can have multiple tags?
* do we intend for aliases to have the property that no two freight (in the same project) can be labeled with the same alias value?

krancour:
> is it possible for a Freight to have 2+ alias's similar to how a git commit can have multiple tags?

I hadn't quite been looking at it in the same terms as git tags, but that is an interesting idea.

At present, this is just a (single) friendlier supplement to a Freight's name/ID because SHAs are awkward for humans. I think @rbreeze plans to incorporate this into the UI somehow.

#1076 will add support for _changing_ the alias. (So you can label important freight as you wish.) I hadn't given consideration to supporting multiples.

I chose to implement this as a label because it makes finding Freight by its alias easy even if you're using `kubectl`.

I'm not sure how we could make that particular aspect of this work if we allowed multiples.

Happy to hear more about what you think about this.

> do we intend for aliases to have the property that no two freight (in the same project) can be labeled with the same alias value?

Yes. Already the controller is confirming no other Freight is using a newly generated alias before it labels a new piece of Freight with it.

#1076 would include a similar safeguard in an admission/validating webhook.

rbreeze:
@jessesuen, @krancour and I discussed this morning. Personally I think supporting just one alias (for now, at least) is good enough. I agree with @krancour that being able to leverage `kubectl` to filter by label is useful, and from a UI perspective, multiple aliases would have cascading effects WRT interface changes.

Are there specific use cases where you envision multiple aliases being useful?

PR #1103, opened 2023-11-14
fix infinite promos on error

### LINKED ISSUE #1101: bug: infinite failed promotions
If a promotion fails for any reason, a new promotion just like it is immediately created -- and will likely fail for the same reason.

This can lead to an infinite loop creating failed promotions.

To re-create, a convenient way to force a promotion failure is configure your stage to work with a Kustomize overlay that doesn't exist.

krancour:
Fixes #1101


jessesuen (approved):
Great!

PR #1115, opened 2023-11-15
chore: add awscli for nightly build

34fathombelow:
🤦 I did not realize that the CLI images were built in a container.  The container by default does not include the awscli.

Fixes: https://github.com/akuity/kargo/actions/workflows/release.yaml?query=event%3Aschedule


krancour (approved):

krancour:
Thanks @34fathombelow!

Issue #1123, opened 2023-11-16
flaky test: TestPriorityQueue


krancour:
This test fails periodically:

```
=== RUN   TestPriorityQueue
    queues_test.go:42: 
        	Error Trace:	/__w/kargo/kargo/internal/controller/runtime/queues_test.go:42
        	Error:      	Not equal: 
        	            	expected: 49
        	            	actual  : 48
        	Test:       	TestPriorityQueue
--- FAIL: TestPriorityQueue (0.00s)
```

krancour:
This was fixed by #1129

PR #1128, opened 2023-11-16
ignore me: hunting down the root case of a flakey test

krancour:
Ignore this PR entirely.

Related to #1123 

1. I cannot replicate the issue locally

1. I cannot even see that the assertion that's allegedly failing is even tested in the indicated location.

This PR is so that I can run and re-run CI in an attempt to nail down the issue.


krancour:
Found it. The branch I was troubleshooting on was out of date and I hadn't realized.

PR #1138, opened 2023-11-17
fix(ui): Node view styles improvements

rpelczar:
Improvements:
- Too long warehouse name truncated
- Fixed styles in the warehouse node view (the border-bottom is not visible)
- Added truncating styles to the repo link instead of characters limitation (chars have different widths, so sometimes the link has two lines)
- Aligned health status in the stage node view

Before:
image

After:
image


rbreeze (approved):
LGTM, thanks!

PR #1140, opened 2023-11-17
docs: clarify need to fork all branches

crenshaw-dev:
If you fork only `main`, the three apps pointing to other branches will fail to sync.


crenshaw-dev:
lol I need to read further down the docs

Issue #864, opened 2023-09-28
Kargo `get` should have `--all-projects`


christianh814:
When running `kargo get` it would be nice to be able to see everything by passing `--all-projects` (maybe `-P` to match Kubernetes' version of `--all-namespaces` with the shorthand of `-A`)

krancour:
Totally.

Note to whoever works on this, there may be special considerations for this operation with respect to whatever we end up doing for SSO user --> cluster SA mappings.

krancour:
> Note to whoever works on this, there may be special considerations for this operation with respect to whatever we end up doing for SSO user --> cluster SA mappings.

This can be ignored for the moment, which makes this a relatively small change -- good for hacktoberfest.

maksimstankevic:
Hi! I submitted https://github.com/akuity/kargo/pull/1143 for this.

maksimstankevic:
I'm closed #1143 as it looks like now #1098  is standing in the way and it's not clear "--all-projects" option for get commands will live.

krancour:
> closed https://github.com/akuity/kargo/pull/1143 as it looks like now https://github.com/akuity/kargo/issues/1098 is standing in the way

#1148 actually

Closing as explained [here](https://github.com/akuity/kargo/pull/1143#issuecomment-1821009018).

Issue #1102, opened 2023-11-14
SSO OIDC mapping to ServiceAccount for Auth-N


jessesuen:
# Proposed Feature

Today we have incomplete SSO support and anyone who logs in is a kargo admin. Kargo needs a way to scope the privileges of what the current user can do.

# Motivation

This is table stakes.

# Suggested Implementation

Discussed with @krancour , @devholic , @gdsoumya. We will use a similar technique that used in [argo-workflows](https://argoproj.github.io/argo-workflows/argo-server-sso/#sso-rbac).

1. The `kargo-api` Role will be given the following additional privileges:
* list/get/watch ServiceAccounts in all namespaces
* the ability to perform `SubjectAccessReviews` or `LocalSubjectAccessReview` (not sure if we can use Local)

2. A kargo project owner, who has permissions to his/her own namespace, will be able to annotate ServiceAccounts with a mapping to OIDC users and groups. Something like:

```yaml
kind: ServiceAccount
metadata:
  annotations:
    kargo.akuity.io/rbac.groups: devops,prod-admins
    kargo.akuity.io/rbac.sub: jesse@akuity.io,kent@akuity.io
  name: devops
```

3. When authorizing a request, the `kargo-api` server will:
* Iterate all ServiceAccounts in the project namespace in question
* For each ServiceAccount in project namespace
    * Check if ServiceAccount has an OIDC group/user associated with it
        * If yes perform a SubjectAccessReview to see if the ServiceAccount has permission to perform the request
            * If yes, allow the request to occur (using `kargo-api`'s permission)

4. If the project level ServiceAccounts had no ability to perform the request, kargo-api should fall back to **_globally_** configured ServiceAccounts that a kargo administrator defined in the `kargo` namespace. This will allow a kargo admin to configure some global defaults (e.g. permit everyone who is able to login, to have a `viewer` Role)

NOTE: unlike argo-workflows, the above approach has the benefit of *aggregating* the permissions of multiple ServiceAccounts, due to the fact that we iterate multiple ServiceAccounts and perform multiple `SubjectAccessReview`s.

# Open Issues

We agreed to have a simple CSV mapping:

```yaml
  annotations:
    kargo.akuity.io/rbac.groups: devops,prod-admins
    kargo.akuity.io/rbac.sub: jesse@akuity.io,kent@akuity.io
```

But for a global configuration, we would need the ability to support something like wildcard globs:

```yaml
  annotations:
    kargo.akuity.io/rbac.sub: *
```

Issue #1098, opened 2023-11-12
improve network efficiency of cli commands


krancour:
I noticed today that CLI commands of the form `kargo get <type> <name>` are performing a list operation and then selecting the desired items from the results.

This is going to be wildly inefficient for large numbers of Freight and/or Promotions, so I think this is something we can improve.

I'd like to see us make selections on the server side and return only what was requested.

API endpoints may need to be refactored a bit to accommodate this.

PR #1143, opened 2023-11-18
Issue/864

maksimstankevic:
Hey! I'd like to contribute to kargo and this is my first PR, meant for https://github.com/akuity/kargo/issues/864


gdsoumya (changes requested):
The current approach is done on the client side, which is acceptable but imo as more efficient solution we should implement this on the server.

krancour:
@gdsoumya, fyi, this seems to be a problem in general... I opened #1098 over this.

krancour:
This has the potential to clash with #1102/#1148, which I noted [here](https://github.com/akuity/kargo/issues/864#issuecomment-1739184919) was a possibility.

That clash is resolvable through the right permissions (permission to list Kargo resources across all namespaces), but I am trying to wrap my head around how likely/frequently that will be given out.

Ultimately, I am curious if we're doing this because there's a demonstrable need for it or just because `kubectl` has `-A`, so we feel the need for something equivalent. Put another way, how often do we suppose someone will _want_ to see Stages, Freight, or Promotions across projects? What sort of workflow does that support when most everything one does in Kargo is scoped to the project-level?

@christianh814 let's chat offline about this?

@maksimstankevic I'm also interested in hearing your perspective. Is this PR purely altruistic (thank you!) or do you have a personal need for this feature? I think hearing the perspective of an actual user will help to make a more informed decision about this.

maksimstankevic:
I'm closing this PR as it looks like now  https://github.com/akuity/kargo/issues/1098 is standing in the way and it's not clear "--all-projects" option for get commands will live.

krancour:
The clash is actually with #1148

Just to put some closure on this for anyone following...

#864 was opened more from a parity-with-kubectl perspective and not out of a demonstrated need for the feature. If it weren't for the clash with #1148, that would have been fine, _but_ reconciling these two things isn't worth the effort until there's a demonstrated need for #864.

If that need ever arises, this will be re-opened.

Thanks for the effort that went into this @maksimstankevic.

PR #1148, opened 2023-11-20
feat: allow mapping SSO context to ServiceAccount

devholic:
https://github.com/akuity/kargo/issues/1102

This commit maps SSO context (user information) to ServiceAccount to authorize user to access Kubernetes API with the matching ServiceAccount's permission.

@jessesuen For now, this PR doesn't support glob pattern as annotation values. IMHO, it would be better to make users to configure Auth-related settings explicitly - since it will not allow unintended subjects by default. WDYT?

### DIFF
--- a/internal/api/option/auth.go
+++ b/internal/api/option/auth.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	libClient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 	"github.com/akuity/kargo/internal/api/config"
 	"github.com/akuity/kargo/internal/api/user"
 	"github.com/akuity/kargo/internal/kubeclient"
@@ -271,6 +272,20 @@ func (a *authInterceptor) listServiceAccounts(
 			kubeclient.ServiceAccountsByGroupIndexField: group,
 		})
 	}
+	kargoNamespaces := make(map[string]struct{})
+	if a.cfg.KargoNamespace != "" {
+		// Kargo namespace is allowed by default
+		kargoNamespaces[a.cfg.KargoNamespace] = struct{}{}
+	}
+	nsList := &corev1.NamespaceList{}
+	if err := a.internalClient.List(ctx, nsList, libClient.MatchingLabels{
+		kargoapi.LabelProjectKey: kargoapi.LabelTrueValue,
+	}); err != nil {
+		return nil, errors.Wrap(err, "list namespaces")
+	}
+	for _, ns := range nsList.Items {
+		kargoNamespaces[ns.GetName()] = struct{}{}
+	}
 	accounts := make(map[string]map[types.NamespacedName]struct{})
 	for _, q := range queries {
 		list := &corev1.ServiceAccountList{}
@@ -281,6 +296,9 @@ func (a *authInterceptor) listServiceAccounts(
 			key := types.NamespacedName{
 				Namespace: sa.GetNamespace(),
 				Name:      sa.GetName(),
+			}
+			if _, ok := kargoNamespaces[key.Namespace]; !ok {
+				continue
 			}
 			if _, ok := accounts[key.Namespace]; !ok {
 				accounts[key.Namespace] = make(map[types.NamespacedName]struct{})


krancour:
👀

api/v1alpha1/annotations.go
```diff
@@ -0,0 +1,7 @@
+package v1alpha1
+
+const (
+	AnnotationKeyRefresh      = "kargo.akuity.io/refresh"
+	AnnotationKeyRBACGroups   = "kargo.akuity.io/rbac.groups"
+	AnnotationKeyRBACSubjects = "kargo.akuity.io/rbac.sub"
+)
```
krancour, line 7: I really like the placement of these here. There are probably other labels and annotations we could move to here in a separate PR.

charts/kargo/templates/api/cluster-role.yaml
```diff
@@ -19,6 +19,28 @@ rules:
...
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - users
+      - groups
+      - serviceaccounts
+    verbs:
+      - impersonate
```
krancour, line 37: This is a lot of privilege that I know @jessesuen and I both wanted to avoid.

I'll review the rest of the PR and then circle back to here... I think there are ways we can avoid this.

charts/kargo/templates/api/cluster-role.yaml
```diff
@@ -19,6 +19,28 @@ rules:
...
+  - apiGroups:
+      - authorization.k8s.io
+    resources:
+      - selfsubjectaccessreviews
+    verbs:
+      - create
```
krancour, line 43: This permission, afaik, doesn't ever have to be explicitly granted. All users are able to perform selfsubjectaccess reviews.

charts/kargo/templates/api/cluster-role.yaml
krancour: Ok... I see what's going on.

I see you are impersonating service accounts to perform a selfsubjectaccess review.

It would be safer to _not_ grant impersonation privileges and instead grant prividleges to conduct a subjectaccessreview (on others; _not_ "self").

internal/api/kubernetes/client.go
```diff
@@ -529,74 +533,140 @@ func getAuthorizedClient(
...
-		// We should replace this with a SubjectAccessReview on the SA to determine
-		// if the desired operation is allowed. If it is, we can return the
-		// internalClient.
-		return internalClient, nil
+		serviceAccounts := make(map[string]corev1.ServiceAccount)
+		{
+			list := &corev1.ServiceAccountList{}
+			if err := internalClient.List(ctx, list, libClient.MatchingFields{
+				kubeclient.ServiceAccountsBySubjectIndexField: userInfo.Username,
+			}); err != nil {
```
krancour, line 551: I believe we can constrain this operation by namespace.

If we don't there's a possibility that the user has mappings to a lot of different ServiceAccounts in a lot of different projects, and then we'll end up doing a lot of SubjectAccessReviews.

Also, per #1102, we're only supposed to take into account ServiceAccounts in the project namespace and _then_ fall back on looking for Services accounts in Kargo's own namespace only if no suitable ones are found at the project level.

internal/api/kubernetes/client.go
```diff
@@ -529,74 +533,140 @@ func getAuthorizedClient(
...
+		}
+		for _, g := range userInfo.Groups {
+			group := strings.TrimSpace(g)
+			if group == "" {
+				continue
+			}
+			list := &corev1.ServiceAccountList{}
+			if err := internalClient.List(ctx, list, libClient.MatchingFields{
+				kubeclient.ServiceAccountsByGroupIndexField: group,
+			}); err != nil {
```
krancour, line 568: Same as [here](https://github.com/akuity/kargo/pull/1148/files#r1399606016).

krancour:
@devholic, overall this looks very close to what we need. I left just a few comments on how to do this with fewer permissions and fewer access reviews.

internal/api/kubernetes/client.go
krancour: > For example, if the ServiceAccount has a ClusterRole, do we have to deny accessing to the other namespace? 

Not explicitly, no. I don't want to interfere with normal RBAC in any way, of course, but simply defining rules for _how we resolve_ user --> SA mappings does not interfere as far as I can see.

As to what those rules ought to be...

Global mappings in the kargo namespace notwithstanding, it makes sense if project admins have _exclusive_ control over what SA each user is mapped to when interacting with their project. In that case, mappings that other project admins made in other namespaces would be irrelevant.

internal/api/option/auth.go
```diff
@@ -263,6 +263,15 @@ func (a *authInterceptor) authenticate(
...
 	}
+	if rawToken == "test-hoon" {
+		return user.ContextWithInfo(
+			ctx,
+			user.Info{
+				Username: "hoon@akuity.io",
+				Groups:   []string{"devops"},
+			},
+		), nil
+	}
```
krancour, line 274: I don't think you meant to leave this here.

krancour:
@devholic looking good. I have two changes I want to propose:

1. All of this new logic for selecting SAs that the user is mapped to, followed by SARs until one of those SAs turns out to be suitable is all going to be repeated for _every_ call the API server makes to the k8s client. Since some API server endpoints make _many_ calls to the k8s client, this can end up being pretty inefficient.

    There is no getting around doing SARs just-in-time, unfortunately, but we _can_ find all the SAs that the user is mapped to just once per request to the API server by doing it as part of the authentication process. We can include the full set of SAs the user is mapped to as a new field in the `user.Info` struct that gets bound to the request context. (We don't need the full SA structs -- just their namespace and name.)

    Besides not having to work out what the mappings are multiple times per request to the API, this approach would also give us the opportunity to preempt the entire request and return a 403 if the user isn't mapped _any_ SAs at all.

1. I wasn't imagining that we fall back on a completely _static_ set up SAs in the `kargo` namespace when no mappings yield SAs for the user in project namespace(s). I was imagining the mappings in the `kargo` namespace would _still_ be dynamic.

    i.e. I'd rather see `rbac.kargo.akuity.io/fallback-service-account` go away and have the SAs in that NS just play by the same rules as the projects using the `rbac.kargo.akuity.io/sub` and `rbac.kargo.akuity.io/groups` annotations.

    Why it matters:

    With the static set of SAs annotated with `rbac.kargo.akuity.io/fallback-service-account`, there would be no way for a Kargo admin to designate a broad set of permissions _only for specific users or groups_. But by playing by the same rules as in the other namespaces, this possibility exists.

cmd/controlplane/api.go
```diff
@@ -36,18 +36,22 @@ func newAPICommand() *cobra.Command {
...
 			}
 			kubeClient, err := kubernetes.NewClient(ctx, restCfg, kubernetes.ClientOptions{
+				KargoNamespace:    cfg.KargoNamespace,
 				NewInternalClient: newClientForAPI,
 			})
 			if err != nil {
-				return errors.Wrap(err, "error creating Kubernetes client")
+				return errors.Wrap(err, "create Kubernetes client")
+			}
+			internalClient, err := newClientForAPI(ctx, restCfg, kubeClient.Scheme())
```
krancour, line 51: I think this can be improved. The way it is, we'll end up with _two_ distinct internal clients, and each of them will have their own in-memory cache. It won't be good for our memory consumption.

Can you possibly do something like this (pseudocode) so we have only one internal client?

```
internalClient, err := newClientForAPI(ctx, restCfg, kubeClient.Scheme())
// Handle error
kubeClient, err := kubernetes.NewClient(ctx, restCfg, kubernetes.ClientOptions{
  KargoNamespace:    cfg.KargoNamespace,
  NewInternalClient: func(...) ... {
    return internalClient
  }
})
```

internal/api/kubernetes/client.go
```diff
@@ -89,6 +95,8 @@ type Client interface {
 		namespace string,
 		opts metav1.ListOptions,
 	) (watch.Interface, error)
+
+	getInternalClient() libClient.Client
```
krancour, line 99: I can't figure out why this is added to the interface. I must be missing something?

internal/api/kubernetes/client.go
```diff
@@ -510,93 +522,174 @@ func gvrAndKeyFromObj(
...
+				accounts, ok := userInfo.ServiceAccounts[ns]
+				if !ok {
+					continue
+				}
+				for _, name := range accounts {
+					serviceAccounts = append(serviceAccounts, types.NamespacedName{
+						Namespace: ns,
+						Name:      name,
+					})
+				}
```
krancour, line 579: Can this be simplified to one line?

```
serviceAccounts = append(serviceAccounts, accounts...)
```

internal/api/kubernetes/client.go
```diff
@@ -510,93 +522,174 @@ func gvrAndKeyFromObj(
...
+				for ns, accounts := range userInfo.ServiceAccounts {
+					if ns == key.Namespace || ns == kargoNamespace {
+						continue
+					}
+					for _, name := range accounts {
+						serviceAccounts = append(serviceAccounts, types.NamespacedName{
+							Namespace: ns,
+							Name:      name,
+						})
+					}
```
krancour, line 592: Same question as [here](https://github.com/akuity/kargo/pull/1148/files#r1403556969)

internal/api/kubernetes/client.go
```diff
@@ -510,93 +522,174 @@ func gvrAndKeyFromObj(
...
+					// return the internal client so that most operations are performed by a
+					// single client with a built-in cache.
+					return internalClient, nil
+				}
+				if !apierrors.IsForbidden(err) {
+					return nil, errors.Wrap(err, "get service account specific client")
+				}
+			}
+			return nil, newForbiddenError(ra)
+		}
```
krancour, line 613: Instead of only selecting relevant SAs in the loop above and then doing SARs this additional loop, we could probably have the first loop do the SARs directly.

That would spend less time building a where many SAs may never get checked since the loop short circuits as soon one appropriate SA is found.

internal/api/kubernetes/client.go
```diff
@@ -510,93 +522,174 @@ func gvrAndKeyFromObj(
...
+			}
+			for _, sa := range serviceAccounts {
+				_, err := getUserClient(
+					ctx,
+					internalClient.Scheme(),
+					ra,
+					withServiceAccount(sa.Namespace, sa.Name),
+				)
+				if err == nil {
+					// Note: At present, we never return the user-specific client. We always
```
krancour, line 603: Let's remove the "at present" part, because this isn't going to change.

internal/api/kubernetes/client.go
```diff
@@ -510,93 +522,174 @@ func gvrAndKeyFromObj(
...
+		_, err := getUserClient(
+			ctx,
+			internalClient.Scheme(),
+			ra,
+			withBearerToken(userInfo.BearerToken),
+		)
+		if err != nil {
+			return nil, errors.Wrap(err, "get user-specific client")
+		}
+		// Note: At present, we never return the user-specific client. We always
```
krancour, line 630: Again, we can probably remove the "at present" part.

internal/api/option/auth.go
```diff
@@ -243,6 +255,43 @@ func (a *authInterceptor) WrapStreamingHandler(
...
+		{
+			kubeclient.ServiceAccountsBySubjectIndexField: username,
+		},
+	}
+	for _, group := range groups {
+		queries = append(queries, libClient.MatchingFields{
+			kubeclient.ServiceAccountsByGroupIndexField: group,
+		})
+	}
+	accounts := make(map[string]map[string]struct{})
```
krancour, line 273: Could we save ourselves from more looping later for every authz by building this collection differently here at authn time?

```
accounts := make(map[string]map[types.NamespacedName]struct{})
```

internal/api/kubernetes/client.go
krancour: Ah... I now see why this is the way it is...

See this [comment](https://github.com/akuity/kargo/pull/1148/files#r1403580078)

internal/api/kubernetes/client.go
```diff
@@ -510,93 +522,174 @@ func gvrAndKeyFromObj(
...
+// If so, the user-specific client is returned.
+//
+//nolint:unparam
 func getUserClient(
 	ctx context.Context,
 	scheme *runtime.Scheme,
-	bearerToken string,
-	resourceAttrs authv1.ResourceAttributes,
+	ra authv1.ResourceAttributes,
+	opts ...userClientOption,
```
krancour, line 672: I know I wrote this originally...

I must have thought at one time that we might potentially _use_ the user-specific client, but we don't and it seems clear that we never will.

We should probably change the name of the function and return types to reflect that this is just carrying out some access review. There's no need to return the client.

internal/api/kubernetes/client.go
```diff
@@ -510,93 +522,174 @@ func gvrAndKeyFromObj(
...
+					continue
+				}
+				for _, name := range accounts {
+					serviceAccounts = append(serviceAccounts, types.NamespacedName{
+						Namespace: ns,
+						Name:      name,
+					})
+				}
+			}
+			// If the operation is cluster-scoped, all ServiceAccounts are candidates.
```
krancour, line 581: This I am not sure about...

I don't think the API server deals with _any_ cluster-scoped resources, so I'd rather just leave this out for now and consider putting it back later if we figure out it's needed.

internal/api/kubernetes/client.go
```diff
@@ -510,93 +522,174 @@ func gvrAndKeyFromObj(
...
+			if key.Namespace == "" {
+				for ns, accounts := range userInfo.ServiceAccounts {
+					if ns == key.Namespace || ns == kargoNamespace {
```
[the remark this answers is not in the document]
krancour: Actually... nevermind about that... see [this comment](https://github.com/akuity/kargo/pull/1148/files#r1403589278)

krancour:
@devholic this is looking awesome. I made some more suggestions, but I think they're mostly small things -- mainly aimed at fewer/tighter loops.

krancour:
> For now, this PR doesn't support glob pattern as annotation values. IMHO, it would be better to make users to configure Auth-related settings explicitly - since it will not allow unintended subjects by default. WDYT?

💯 agree with @devholic.

What stops someone from coming along with a new/unanticipated group and thereby acquiring permissions there were not meant to have?

I think if we _really_ want that, we need to put a lot of deliberate thought into how to prevent such a scenario.

I think it's right to leave it out for now.

internal/api/kubernetes/client.go
krancour: I forgot about namespace. You're right. Leave this, but let's continue putting more thought into it.

internal/api/option/auth.go
```diff
@@ -243,6 +256,46 @@ func (a *authInterceptor) WrapStreamingHandler(
...
+			}
+			accounts[key.Namespace][key] = struct{}{}
+		}
+	}
+	res := make(map[string][]types.NamespacedName)
+	for ns, names := range accounts {
+		for name := range names {
+			res[ns] = append(res[ns], name)
+		}
+	}
```
krancour, line 296: How much do we really need this loop?

Can we just return the `map[string]map[types.NamespacedName]struct{}` instead of returning `map[string][]types.NamespacedName`?

I think that would be ok, because I don't think order really matters when we use this information for authz decisions later on.

wdyt?

internal/kubeclient/indexer.go
```diff
@@ -26,6 +28,9 @@ const (
 	StagesByArgoCDApplicationsIndexField = "applications"
 	StagesByUpstreamStagesIndexField     = "upstreamStages"
 	StagesByWarehouseIndexField          = "warehouse"
+
+	ServiceAccountsByGroupIndexField   = "group"
+	ServiceAccountsBySubjectIndexField = "subject"
```
krancour, line 33: Tiniest nit: Should these be plural "groups" and "subjects" because I think both of these indices can end up being mulit-valued.

internal/kubeclient/indexer.go
```diff
@@ -253,3 +258,53 @@ func indexStagesByWarehouse(obj client.Object) []string {
...
+	}
+	groups := strings.Split(rawGroups, ",")
+	refinedGroups := make([]string, 0, len(groups))
+	for _, g := range groups {
+		if group := strings.TrimSpace(g); group != "" {
+			refinedGroups = append(refinedGroups, group)
+		}
+	}
+	return refinedGroups
+}
```
krancour, line 285: Had a thought about all the new indexing functions...

Should we be validating that the SA's namespace is indeed a Kargo project?

We either have to do that when building the index _or_ after using the index.

I think it's cheaper to do it when building the index, because this function will fire only when a SA is modified and I think that should happen a lot less frequently than when we _use_ the indices during the authn process.

internal/kubeclient/indexer.go
krancour: Non-blocking.

krancour:
Looking fantastic. I think there are two small things to give some consideration to, and one very tiny nit that you can choose to deal with or not.

krancour (approved):
🔥 LGTM 🔥 

Thanks for all the hard for on this @devholic!

Issue #1075, opened 2023-11-07
add ui feature for manually qualifying a piece of freight for a given stage


krancour:
Follow up to #1073

krancour:
@rbreeze assigned to you just so someone is on point. Feel free to re-assign. Also lmk if you thin we can't do this in v0.3.0.

PR #1151, opened 2023-11-20
rename Freight's "qualified" field to "verifiedIn"; add "approvedFor" field

### LINKED ISSUE #1033: rename Freight's "qualified" field to "verifiedIn"; add "approvedFor" field
"Qualified" can be a misleading term. It makes sense or it doesn't, depending how you look at it...

Pieces of Freight become qualified _in_ a Stage when they first become healthy in that Stage.

A piece of Freight becomes _available_ to a Stage when it's been _qualified in_ any of that Stage's upstream Stages...

But another way to say "available to Stage x" would be "_qualified for_ Stage x."

_Qualified in_ vs _qualified for_ -- two distinctly different semantic meanings.

Kargo uses the former, but anyone could be forgiven for thinking it is the latter.

Is this asking for trouble? Is there a less ambiguous way to articulate the semantics we're using?

krancour:
Fixes #1033

I have tested this e2e and confirm that "verifiedIn" works identically to how "qualified" worked previously.

"approvedFor" is harder to test because we haven't implemented #1075 yet and `kubectl edit/apply` do not allow mutating status subresources.

@rbreeze and I discussed and we're going to merge this to unblock #1075 and then work out any kinks then.


rbreeze (approved):
LGTM!

PR #1152, opened 2023-11-21
fix(ui): Checking token expiration before requests

### LINKED ISSUE #877: UI not refreshing to login screen after session token is expiered
when the UI is left open for a while the session token expires and the UI still shows the last view without updates.

if the token expires the UI should return to the login screen without the user needing to refresh twice( 1st time to get the token expire error, 2nd to return to the login screen)

also, add a session token timeout value in the helm chart(API server)

rpelczar:
Fixes #877

In this PR, I added a handler to check the token expiration before requests.

### DIFF
--- a/ui/src/config/transport.ts
+++ b/ui/src/config/transport.ts
@@ -6,15 +6,39 @@ import { paths } from './paths';
 
 export const authTokenKey = 'auth_token';
 
-const errorHandler: Interceptor = (next) => async (req) => {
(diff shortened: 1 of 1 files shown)


krancour:
@rpelczar this is for #877, right?

krancour:
Works beautifully @rpelczar!

krancour (approved):

PR #1160, opened 2023-11-21
on new freight, enqueue subscribed stages

### LINKED ISSUE #1157: add subscribed warehouses to the work queue after new freight is created
We aren't doing this today, so it can take a while for Stages subscibed to a Warehouse to notice new Freight.

krancour:
Fixes #1157

This PR indexes Stages by Warehouse. Whenever new Freight is created, a new event handler finds Stages subscribed to the Warehouse that created the Freight and enqueues them for reconciliation.

This makes Stages that subscribe directly to Warehouses far more responsive.

There is a tiny bit of renaming of existing event handlers in this PR. I've added two in the past two days and wanted to adopt a different naming convention that seems it can be applied consistently as the number of handlers likely grows.

cc @tal-hason

### DIFF
--- a/internal/controller/stages/stages.go
+++ b/internal/controller/stages/stages.go
@@ -193,6 +193,11 @@ func SetupReconcilerWithManager(
 		return errors.Wrap(err, "index Stages by upstream Stages")
 	}
 
+	// Index Stages by Warehouse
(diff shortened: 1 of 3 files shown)


rbreeze (approved):
Awesome LGTM!

PR #1171, opened 2023-11-24
feat(ui): Refresh token exchange

### LINKED ISSUE #877: UI not refreshing to login screen after session token is expiered

rpelczar:
Fixes #877 

In this PR, I implemented a refresh token exchange.

How it works: when the token is expired before any API call, the user is redirected to the token renewal page, where the loader is visible for a second, and after the token exchange process, the user is redirected to the previous page.

The perfect solution is a request in the background, but unfortunately, we can't use React hooks inside the interceptor for now (library limitations), which is needed for our auth flow.

### DIFF
--- a/ui/src/features/auth/context/auth-context-provider.tsx
+++ b/ui/src/features/auth/context/auth-context-provider.tsx
@@ -19,6 +19,8 @@ export const AuthContextProvider = ({ children }: PropsWithChildren) => {
 
   const logout = React.useCallback(() => {
     localStorage.removeItem(authTokenKey);
+    localStorage.removeItem(refreshTokenKey);
+
     setToken(null);
   }, []);
 
--- a/ui/src/features/auth/context/auth-context-provider.tsx
+++ b/ui/src/features/auth/context/auth-context-provider.tsx
@@ -1,15 +1,20 @@
 import React, { PropsWithChildren } from 'react';
 
-import { authTokenKey } from '@ui/config/transport';
+import { authTokenKey, refreshTokenKey } from '@ui/config/auth';
 
 import { AuthContext } from './auth-context';
 
 export const AuthContextProvider = ({ children }: PropsWithChildren) => {
   const [token, setToken] = React.useState(localStorage.getItem(authTokenKey));
 
-  const login = React.useCallback((t: string) => {
-    localStorage.setItem(authTokenKey, t);
-    setToken(t);
+  const login = React.useCallback((token: string, refreshToken?: string) => {
+    localStorage.setItem(authTokenKey, token);
+
+    if (refreshToken) {
+      localStorage.setItem(refreshTokenKey, refreshToken);
+    }
+
+    setToken(token);
   }, []);
 
   const logout = React.useCallback(() => {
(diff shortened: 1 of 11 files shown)


ui/src/features/auth/context/auth-context-provider.tsx
```diff
@@ -1,15 +1,20 @@
...
 
-  const login = React.useCallback((t: string) => {
-    localStorage.setItem(authTokenKey, t);
-    setToken(t);
+  const login = React.useCallback((token: string, refreshToken?: string) => {
+    localStorage.setItem(authTokenKey, token);
+
+    if (refreshToken) {
+      localStorage.setItem(refreshTokenKey, refreshToken);
+    }
```
krancour, line 15: This is never getting set for me, but I haven't quite figured out why.

krancour: And my dex is definitely supporting it:

```
{
  "issuer": "https://localhost:30081/dex",
  "authorization_endpoint": "https://localhost:30081/dex/auth",
  "token_endpoint": "https://localhost:30081/dex/token",
  "jwks_uri": "https://localhost:30081/dex/keys",
  "userinfo_endpoint": "https://localhost:30081/dex/userinfo",
  "device_authorization_endpoint": "https://localhost:30081/dex/device/code",
  "grant_types_supported": [
    "authorization_code",
    "refresh_token",
    "urn:ietf:params:oauth:grant-type:device_code"
  ],
  "response_types_supported": [
    "code"
  ],
  "subject_types_supported": [
    "public"
  ],
  "id_token_signing_alg_values_supported": [
    "RS256"
  ],
  "code_challenge_methods_supported": [
    "S256",
    "plain"
  ],
  "scopes_supported": [
    "openid",
    "email",
    "groups",
    "profile",
    "offline_access"
  ],
  "token_endpoint_auth_methods_supported": [
    "client_secret_basic",
    "client_secret_post"
  ],
  "claims_supported": [
    "iss",
    "sub",
    "aud",
    "iat",
    "exp",
    "email",
    "email_verified",
    "locale",
    "name",
    "preferred_username",
    "at_hash"
  ]
}
```

krancour: idk if it's a factor, but I'm using a self-signed cert. In the CLI, I ignore refresh tokens if the user picked `--insecure-skip-tls-verify`. This is because I want to force them to re-authenticate in order to re-evaluate that decision.

idk if this is a factor here as well and possibly the source of my difficulties?

rpelczar: @krancour Can you check the response from the server? Is there a refreshToken?

krancour: @rpelczar I made a short video.

Note to an nefarious evildoers: The fragments of tokens you see in the video are for an IDP running locally and are completely useless. 😁 

https://github.com/akuity/kargo/assets/1821014/3268b270-70d4-456c-afa1-5a6a66f60493

rpelczar: @krancour I found a bug, and fixed it.

krancour (approved):

Issue #1071, opened 2023-11-07
api server needs endpoint for deleting freight


krancour:
Users may often wish to delete a piece of freight that they consider invalid or failed so that it does not clutter their freightline.

krancour:
The endpoint should return an appropriate error if there is an attempt to delete Freight that is currently in use by any Stage. In fact, we should probably incorporate that into an admission/validating webhook so that kubectl users get the same protection.

PR #1176, opened 2023-11-28
feat: check freight is in-use before deleting

devholic:
Part of https://github.com/akuity/kargo/issues/1071

This commit adds a validating webhook that checks if the freight is in-use before deleting it.


gdsoumya (approved):

krancour:
@devholic 🔥

krancour (approved):

PR #1177, opened 2023-11-29
feat: add delete freight endpoint to api server

devholic:
Resolves https://github.com/akuity/kargo/issues/1071

### DIFF
--- a/internal/api/delete_freight_v1alpha1.go
+++ b/internal/api/delete_freight_v1alpha1.go
@@ -6,7 +6,7 @@ import (
 	"connectrpc.com/connect"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


internal/api/delete_freight_v1alpha1.go
```diff
@@ -0,0 +1,46 @@
...
+				connect.CodeNotFound,
+				errors.Errorf("freight %q not found", req.Msg.GetName()),
+			)
+		}
+		return nil, errors.Wrap(err, "get freight")
+	}
+	if err := s.client.Delete(ctx, &freight); err != nil {
+		return nil, errors.Wrap(err, "delete freight")
+	}
+	return connect.NewResponse(&svcv1alpha1.DeleteFreightResponse{}), nil
```
krancour, line 45: @devholic you may have to double-check me on this, but I believe the delete operation will return a not found error if the resource doesn't exist. i.e. I think we can skip the get entirely...

Like I said... double-check... I'm about 80% sure

krancour (approved):
LGTM

Issue #1150, opened 2023-11-20
Kargo cli configuration for Projects


christianh814:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

```shell
$ kargo version
Client Version: v0.2.1
```

# Proposed Feature

A lot of commands require to pass `--project`. It would be nice to be able to set projects "globally" in the CLI. So something like...

```
kargo config set-project <my project name>
```

# Motivation

By having this feature, it will allow for less typing

# Suggested Implementation

Probably have  a `~/.config/kargo/config.yaml` file with a `project:` field

krancour:
💯

krancour:
@devholic do you have bandwidth for this one?

krancour:
> Yep, will handle this tomorrow

@devholic sorry... I jumped the gun. We don't _need_ this for v0.3.0. So this is "if time permits."

PR #1179, opened 2023-11-29
feat: allow users to set default `project` in CLI

devholic:
Resolves https://github.com/akuity/kargo/issues/1150

- Introduces `config set` and `config unset` to deal with CLI configuration
- Moves CLI command packages to under the `internal/cli/cmd` to avoid import cycle


krancour (approved):
🔥

PR #1181, opened 2023-11-29
fix(ui): clarify why freight is available for promotion

### LINKED ISSUE #1158: make it more obvious WHY freight is available to a stage
See #1153 

Some message or other visual means of making the user more aware of what Freight is considered "available" to each Stage might clear up confusion for anyone who has been misled into thinking the two promote buttons on each Stage follow strict push/pull semantics.

For example (lifted from #1153):

> It could be something as simple as a thin banner that appears above the freightline during promotion, saying something like, "These Freight have all qualified in the Stages immediately upstream from `x`." I'm no UI expert though. I defer to @rbreeze for better ideas of how me might visually clear up this confusion.

rbreeze:
Fixes #1158

Screenshot 2023-11-29 at 10 37 10
Screenshot 2023-11-29 at 10 37 29

### DIFF
--- a/ui/src/features/project/project-details/nodes/stage-node.tsx
+++ b/ui/src/features/project/project-details/nodes/stage-node.tsx
@@ -117,7 +117,7 @@ const Nodule = (props: {
         left: props.begin ? -noduleHeight / 2 : 'auto',
         right: props.begin ? 'auto' : -noduleHeight / 2
       }}
-      className={`z-10 flex items-center justify-center hover:text-white border border-sky-300 border-solid hover:bg-blue-400 absolute rounded-lg ${
+      className={`select-none z-10 flex items-center justify-center hover:text-white border border-sky-300 border-solid hover:bg-blue-400 absolute rounded-lg ${
         props.selected ? 'text-white bg-blue-400' : 'bg-white text-blue-500'
       }`}
     >
--- a/ui/src/features/freightline/freightline.tsx
+++ b/ui/src/features/freightline/freightline.tsx
@@ -123,11 +123,11 @@ export const Freightline = (props: {
           ) : (
             <>
               <FontAwesomeIcon icon={faTriangleExclamation} className='mr-2' />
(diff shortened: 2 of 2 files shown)


ui/src/features/project/project-details/nodes/stage-node.tsx
```diff
@@ -117,7 +117,7 @@ const Nodule = (props: {
         left: props.begin ? -noduleHeight / 2 : 'auto',
         right: props.begin ? 'auto' : -noduleHeight / 2
       }}
-      className={`z-10 flex items-center justify-center hover:text-white border border-sky-300 border-solid hover:bg-blue-400 absolute rounded-lg ${
+      className={`select-none z-10 flex items-center justify-center hover:text-white border border-sky-300 border-solid hover:bg-blue-400 absolute rounded-lg ${
```
rbreeze, line 120: Prevents the button text from being selected when clicking repeatedly

krancour:
@rbreeze I like the presentation. I only want to suggest some rewording.

We recently stopped using "qualified" because the meaning was too ambiguous. "Qualified in" and "qualified for" would both be reasonable interpretations of "qualified" and have very different meanings. You probably recall that convo...

We subsequently shifted to using "verified in" and "approved for," so the following would be more accurate:

> Available freight includes all that have been verified in any immediately upstream stage OR approved for this stage.

For promote to subscribers, the changes needed are less extensive:

> Available freight includes all that have been verified in this stage.

But are these too wordy? 🤔 

I welcome further tweaks.

rbreeze:
@krancour Thanks for the reminder, I forgot about the wording change 😄

Both of those sound good to me. There's a lot of real estate to show the message, and given that most people will either read it once or not at all, I don't think it's too wordy. 

In fact, that gives me an idea for a future improvement to this: the option to dismiss the banner once you've read it. And perhaps a question mark button to bring it back. I'll open an issue to track

krancour (approved):
🔥 

This will clear the misconception that promotions are merely "pulls" or "pushes."

PR #1189, opened 2023-11-30
feat: Endpoint to manually approve freight for a stage

### LINKED ISSUE #1073: add api server endpoint for manually qualifying a piece of freight
Depends on #1033 

This would enable hotfixes _without_ cluttering stage DAGs with subscriptions from prod (on the far right) to a warehouse (on the far left).

rbreeze:
Closes #1073


internal/api/approve_freight_v1alpha1.go
```diff
@@ -0,0 +1,65 @@
...
+	approvedStage, ok := st.(kargoapi.ApprovedStage)
+	if !ok {
+		return nil, connect.NewError(connect.CodeInternal,
+			fmt.Errorf("could not cast stage %q to approved stage", key.String()))
+	}
+	freight.Status.ApprovedFor[req.Msg.GetStage()] = approvedStage
+
+	if err := s.client.Update(ctx, &freight); err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
```
krancour, line 62: Two things here...

1. We can get by without the type assertion (casting). `ApprovedFor` is a map of strings to empty structs (which is Go's weird way of accomplishing set semantics). All we care about is carrying out the approval if the stage's name isn't already a key in that map. We can return early if it's already there.

2. The `s.client.Updated(...)` unfortunately creates a race condition. If two users try to manually approve the same freight for different stages are roughly the same time, the second update clobbers the first. I was guilty of doing this, too until until recently.

    It's gotta be a _patch_ operation instead. (And fwiw, the patch only evades the race condition because it's a map and not a list. 😰)

This bit of code for handling a new verification is almost line-for-line the same thing that's needed for a manual approval and should dodge both problems. Just field names change, obviously:

https://github.com/akuity/kargo/blob/main/internal/controller/stages/stages.go#L688-L705

internal/api/delete_stage_v1alpha1.go
```diff
@@ -35,6 +37,32 @@ func (s *server) DeleteStage(
...
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
+
+	for i := range approvedFreightList.Items {
+		freight := approvedFreightList.Items[i]
+		delete(freight.Status.ApprovedFor, stage.Name)
+		if err := s.client.Status().Update(ctx, &freight); err != nil {
+			return nil, connect.NewError(connect.CodeInternal, err)
+		}
```
krancour, line 63: Same thing as [here](https://github.com/akuity/kargo/pull/1189/files#r1410074709) re: needing to use a patch, unfortunately.

krancour (approved):

Issue #1186, opened 2023-11-29
Reconcile two "global" config approaches


krancour:
#1041 added support for specifying namespaces that contain globally accessible credentials (opt-in, of course).

Similarly, #1148 includes falling back on the `kargo` namespace of sso user --> service account mappings.

imho, we should follow up on #1041 to give it the same flexibility as #1148 -- i.e. let the admin who installs Kargo decide what namespace(s) if any to fall back on when mappings are not found in the project namespace.

Mainly for consistency.

@devholic wondering if you could take this one since you very recently worked on #1148 and know the code well.

krancour:
The point of it is to give admins some degree of choice so that they don't have to keep configuration in the `kargo` (system) namespace if they prefer not doing so -- which is why #1041 used the approach it did.

PR #1190, opened 2023-11-30
refactor: allow users to set global service account namespaces

devholic:
Resolves https://github.com/akuity/kargo/issues/1186


krancour:
@devholic it seems some changes might be missing from the PR?

charts/kargo/templates/api/configmap.yaml
```diff
@@ -30,6 +30,9 @@ data:
   {{- end }}
   {{- if .Values.api.oidc.enabled }}
   OIDC_ENABLED: "true"
+  {{- if .Values.api.oidc.globalServiceAccounts.enabled }}
+  GLOBAL_SERVICE_ACCOUNT_NAMESPACES: {{ join "," .Values.api.oidc.globalServiceAccounts.namespaces }}
+  {{- end }}
```
krancour, line 35: What do you think about making this unconditional? i.e. If `.Values.api.oidc.globalServiceAccounts.namespaces` has length zero, the value just ends up being the empty string.

In retrospect, I wish #1041 had done that. I'll open a separate issue or PR to fix that.

charts/kargo/values.yaml
```diff
@@ -112,11 +112,17 @@ api:
     ## @param api.oidc.cliClientID The client ID for the OIDC client used by CLI (optional). Needed by some OIDC providers (such as Dex) that require a separate Client ID for web app login vs. CLI login (`http://localhost`). If Dex is enabled, this value will be ignored and cli client ID will be automatically configured. If Dex is not enabled, and a different client app is configured for localhost CLI login, this should be the client ID configured in the IdP.
     cliClientID:
 
+    globalServiceAccounts:
+      ## @param api.oidc.globalServiceAccounts.enabled Whether to enable the api to look for service accounts in specific namespaces used to share service accounts.
+      enabled: false
```
krancour, line 117: See [this related comment](https://github.com/akuity/kargo/pull/1190/files#r1411408832)

internal/api/option/auth.go
```diff
@@ -272,10 +272,15 @@ func (a *authInterceptor) listServiceAccounts(
 			kubeclient.ServiceAccountsByGroupIndexField: group,
 		})
 	}
-	kargoNamespaces := make(map[string]struct{})
+	allowedNamespaces := make(map[string]struct{})
 	if a.cfg.KargoNamespace != "" {
 		// Kargo namespace is allowed by default
-		kargoNamespaces[a.cfg.KargoNamespace] = struct{}{}
+		allowedNamespaces[a.cfg.KargoNamespace] = struct{}{}
+	}
```
krancour, line 279: We can get rid of this part now. Letting the admin who install specify allows them to name the kargo namespace as an allowed one if they want to, but we no longer have to hard code it.

PR #1201, opened 2023-12-01
switch docs from yarn to pnpm

krancour:
We use pnpm everywhere else in the project.

Figured this makes things just a smidge easier.

Used `pnpm import` for the migration, so we _should_ be pinned to all the exact same versions that `yarn.lock` files pinned us to.


krancour:
@devholic yep... I got the title backwards, didn't I?

Thank you for noticing.

Fixed!

PR #1204, opened 2023-12-01
lower the barriers to entry to run quickstart or to contribute

krancour:
This came out of a dialogue @rbreeze and I had earlier today about the workflows we each tend to use locally.

He mentioned [OrbStack](https://orbstack.dev/) as a drop-in replacement for Docker Desktop optimized for Mac OS.

I couldn't resist trying it and it's blazing fast and extends my battery life significantly when used instead of Docker Desktop.

The best part was I didn't have to mess with a local registry or any port-forwarding, which we need to do with both kind and k3d.

_Everything just worked._

This got me wondering how far Docker Desktop's own Kubernetes support had come. Last time I had used it, it was still rough around the edges.

Although it's not as fast as OrbStack, _everything just worked_ -- again, with no need for any port-forwarding or a local registry.

Apart from these simpler configurations being attractive, documenting setup for Docker Desktop and OrbStack lowers the barrier to entry for contributors and quickstart users alike -- which, in the end, is what I really care about.

__Edit:__ Probably easiest to review this by looking at the Netlify preview.


morey-tech (approved):
LGTM, love this change to enable more users!

PR #1218, opened 2023-12-03
add option to disable liveness and readiness probes

krancour:
This option is useful when developing in a local cluster.

This PR also disables probes when using Tilt to run in a local cluster. It makes things come up quite a bit faster.


rbreeze (approved):
LGTM

PR #1220, opened 2023-12-04
relax cors policy on api server

### LINKED ISSUE #1206: discussion: relax CORS policy on api server, or a certain endpoint thereof
I'll start by saying that imho, there's no real security advantage to strict CORS policies for an API server... in fact, I think relaxing them does a lot for the the prospect of alternative/third-party UIs emerging someday.

Drilling down into specific motivation I have for wanting to relax this, and more than a little bit related to #1205, @rbreeze and I were talking this morning about advantages to running a UI server as a local process (using `vite`), even if the back end happens to be running in-cluster.

To facilitate that, the front-end, when run as a native process, already implements a reverse proxy for whatever API server it is configured to talk to. This skirts CORS issues...

This works great _except for SSO_, because if you happen to be using Dex -- which we do, by default -- the API server's public configuration endpoint, when queried, will return an OIDC issuer URL that points _straight_ to the API server's _own_ reverse proxy for the Dex server... and of course, COORS prevents the browser from talking to the API server without going through its own reverse proxy.

Phew. Complicated.

Anyway... if we were to relax the CORS policy on the API server, either in general, or even just for the paths that are a reverse proxy for Dex, we'd end up with everything, _incuding SSO_ working perfectly with the UI server running as a native process, serving a SPA that talks to an API server elsewhere.

It's actually a very small change.

I can't see any security grounds on which to _not_ do this, but wanted to open it up to discussion in case there is a valid reason for strict CORS policies on an API server that I might be overlooking.

krancour:
Fixes #1206


krancour:
@jessesuen @rbreeze I updated this to be conditional like we discussed this morning.

gdsoumya (approved):

PR #1223, opened 2023-12-04
Allow to specify affinities

Tchoupinax:
Hello,

I made this merge request to add the ability to define pod affinity.

Thank you for your work!
Have a nice day :)

### DIFF
--- a/charts/kargo/Chart.yaml
+++ b/charts/kargo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: kargo
 description: Kargo
 type: application
-version: 0.1.0
+version: 0.1.1
 appVersion: v0.1.0
 sources:
 - https://github.com/akuity/kargo
--- a/charts/kargo/templates/api/deployment.yaml
+++ b/charts/kargo/templates/api/deployment.yaml
@@ -25,6 +25,10 @@ spec:
         secret/checksum: {{ include (print $.Template.BasePath "/api/secret.yaml") . | sha256sum }}
     spec:
       serviceAccount: kargo-api
+      {{- with .Values.api.affinity }}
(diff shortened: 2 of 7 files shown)


krancour:
Hey @Tchoupinax this is a great change and we appreciate it.

This seems to add the capability only for three of five components.

The two that are missed seem to be the Dex server and the garbage collector.

Would you be able to add it for those two as well?

In all cases, can we also please keep the new option right under the corresponding tolerations option in the values.yaml file, just so we're consistently keeping similar settings grouped together!

Thanks so much for this!

Tchoupinax:
Hello @krancour,

I took your review in account, please tell me what do you think.
Sorry for the noice in Git, the rebase with the fork did not happen as expected ^^'

Have a nice day!

charts/kargo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: kargo
 description: Kargo
 type: application
-version: 0.1.0
+version: 0.1.1
```
krancour, line 5: @Tchoupinax this is just a placeholder value. (I wish it could be v0.0.0 or "placeholder" so that would be more obvious, but the linter doesn't allow that.)

The release process will automatically set this correctly.

krancour:
```suggestion
version: 0.1.0
```

krancour (approved):
fyi @Tchoupinax 

I ran codegen, which also updates docs that are derived from metadata in the values.yaml file and amended the PR with that.

LGTM

Thank you so much for this contribution!

Tchoupinax:
Great, thank you for handling this!
Have a nice day :)

PR #1226, opened 2023-12-05
make vite bind to 0.0.0.0

krancour:
Vite serves the UI during development only, so this has no bearing on anything but devx.

Making vite bind to 0.0.0.0 instead of 127.0.0.1 only makes it possible to run vite in a container without affecting the ability to run it as a native process.


rbreeze (approved):
LGTM!

Issue #1197, opened 2023-11-30
UI needs a way to determine freight ID (after aliases were added)


jessesuen:
# Description

I want to know the freight ID (hash), because I want to find the corresponding CR in my namespace. But UI has no way of finding it after aliases were introduced.

# Screenshots

<img width="452" alt="image" src="https://github.com/akuity/kargo/assets/12677113/1347407b-5216-470d-93b4-432f0efb2ffc">

# Version

Tip as of today

# Logs

```
Paste any relevant application logs here.
```

krancour:
Mouseover... added earlier today. Problem is it's not obvious you can do that.

krancour:
@rbreeze do you think if the alias were styled like a link it might prompt people in search of more info to mouseover?

rbreeze:
I think that may be a question @jessesuen could answer better: would an underline have helped you discover it?

An underline wouldn't be my first choice in this case, as IMO it would add a lot of visual cruft since it would affect every alias label. But at the same time, I can't think of a better way to make it clearer.

Two more thoughts:
1. Clicking the alias should copy the ID (or at least, there should be an action menu item to copy the ID)
2. The Kargo CLI should have a feature to allow getting Freight by alias

krancour:
> Clicking the alias should copy the ID (or at least, there should be an action menu item to copy the ID)

Love that idea.

PR #1238, opened 2023-12-06
feat: ability to copy freight id and alias

rbreeze:
CleanShot 2023-12-06 at 12 54 28@2x

Closes https://github.com/akuity/kargo/issues/1197


jessesuen:
Is there a way to see the freight ID without copying it? Sometimes I just want to visually see the ID (e.g. mentally map it to a CR i see from kubectl get freight).

krancour (approved):

PR #1239, opened 2023-12-06
Krancour/digests

### LINKED ISSUE #1078: better support for image digests
After #1058 is merged, we will _always_ have image digests.

If we saved image digests in Freight (and used them, instead of tags in the Freight ID calculations), this would allow us to start supporting the possibility of (optionally) updating kustomize and/or helm configuration not with a tag, but with a digest. (Not all Helm charts will support this, but that's not actually our problem.)

And while I don't recommend that anyone ever use mutable tags (e.g. latest), this change offers the possibility of subscribing to a mutable tag, creating _new_ Freight every time that tag is updated to point to a new digest, and then incorporating that new digest into application configuration.

Even when using _immutable_ tags, some users may prefer promotion mechanisms to use digests over tags simply as a matter of precision. 🖐️

krancour:
Fixes #1078 

Since #1058 we're able to get our hands on image digests (regardless of tag selection strategy).

This PR makes relevant adjustments to CRDs, protobufs, type conversion code, and controller, such that we not only capture this information in Freight (and use it in Freight ID calculation), but also have the option to _use_ the digest instead of the tag in promotion mechanisms.

Note: Most Helm charts do _not_ support specifying an image using a digest, but some do. With Kustomize, this always works.

More work is needed to make it work with Kargo Render, but the PR is big enough already. I will handle the Kargo Render part in a follow-up.

Example of using this with Helm:

```yaml
promotionMechanisms:
  gitRepoUpdates:
  - repoURL: https://github.com/krancour/kargo-demo-gitops.git
    writeBranch: stage/test
    helm:
      images:
      - image: nginx
        valuesFilePath: charts/kargo-demo/stages/test/values.yaml
        key: image.digest
        value: Digest
```

Example of using this with Kustomize:

```yaml
promotionMechanisms:
  gitRepoUpdates:
  - repoURL: https://github.com/krancour/kargo-demo-gitops.git
    writeBranch: stage/uat
    kustomize:
      images:
      - image: nginx
        path: stages/uat
        useDigest: true
```

Not to reviewer(s): A substantial chunk of this PR is generated code isolated to one commit.


api/v1alpha1/stage_types.go
```diff
@@ -6,12 +6,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// +kubebuilder:validation:Enum={Image,Tag}
+// +kubebuilder:validation:Enum={ImageAndTag,Tag,ImageAndDigest,Digest}
```
krancour, line 9: This is a breaking change, fyi.

jessesuen (approved):
Great work! The ability to resolve and render the image digest to the final output will be a great security improvement that many will appreciate. 

Just so I understand, with this PR, the ability to render digest is supported when a user wants Kargo to update an Argo CD spec, but not yet supported with rendered branches because it requires a corresponding change to kargo-render?

jessesuen:
> because it requires a corresponding change to kargo-render?

Ah I just saw https://github.com/akuity/kargo/pull/1240, so I think I answered my own question.

krancour:
@jessesuen

> Just so I understand, with this PR, the ability to render digest is supported when a user wants Kargo to update an Argo CD spec, but not yet supported with rendered branches because it requires a corresponding change to kargo-render?

It will also work if you're using the straight kustomize or helm-based promotion mechanisms without rendering.

Rendered branches were the only case not covered by this PR and only because it required some additional thought.

PR #1245, opened 2023-12-07
fix: correct image links for popular image registries

### LINKED ISSUE #831: Links on image subscriptions are broken in the UI
Xnapper-2023-09-22-11 48 37

UI seems to assume that the image URL can be accessed directly.

rbreeze:
Fixes #831

### DIFF
--- a/ui/src/utils/url.test.ts
+++ b/ui/src/utils/url.test.ts
@@ -21,7 +21,10 @@ const testImages = {
 
   // DockerHub User
   'jitesoft/nginx': 'https://hub.docker.com/r/jitesoft/nginx',
-  'docker.io/jitesoft/nginx': 'https://hub.docker.com/r/jitesoft/nginx'
+  'docker.io/jitesoft/nginx': 'https://hub.docker.com/r/jitesoft/nginx',
+
+  // AWS
+  'public.ecr.aws/karpenter/controller': 'https://gallery.ecr.aws/karpenter/controller'
 } as { [key: string]: string };
 
 describe('urlForImage', () => {
--- a/ui/src/utils/url.test.ts
+++ b/ui/src/utils/url.test.ts
@@ -24,7 +24,9 @@ const testImages = {
   'docker.io/jitesoft/nginx': 'https://hub.docker.com/r/jitesoft/nginx',
 
   // AWS
-  'public.ecr.aws/karpenter/controller': 'https://gallery.ecr.aws/karpenter/controller'
+  'public.ecr.aws/karpenter/controller': 'https://gallery.ecr.aws/karpenter/controller',
+  '602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/coredns':
+    'https://us-west-2.console.aws.amazon.com/ecr/repositories/private/602401143452/eks/coredns'
 } as { [key: string]: string };
 
 describe('urlForImage', () => {
--- a/ui/src/utils/url.test.ts
+++ b/ui/src/utils/url.test.ts
@@ -1,0 +1,33 @@
+import { expect, test, describe } from 'vitest';
+
+import { urlForImage } from './url';
+
+const testImages = {
+  // Quay
+  'quay.io/jitesoft/nginx': 'https://quay.io/jitesoft/nginx',
+
+  // GHCR
+  'ghcr.io/jitesoft/nginx': 'https://ghcr.io/jitesoft/nginx',
+
+  // pkg.dev
+  'us-docker.pkg.dev/cloudrun/container/hello':
+    'https://us-docker.pkg.dev/cloudrun/container/hello',
+
+  // DockerHub Official
+  nginx: 'https://hub.docker.com/_/nginx',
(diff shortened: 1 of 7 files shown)


ui/src/utils/url.ts
```diff
@@ -1 +1,9 @@
-export const urlWithProtocol = (url: string) => `https://${url.replace('https://', '')}`;
+export const urlWithProtocol = (image: string): string | undefined => {
+  if (image.startsWith('quay.io') || image.startsWith('docker.io') || image.startsWith('ghcr.io')) {
+    return `https://${image}`;
+  } else if (image.includes('/')) {
+    return `https://hub.docker.com/r/${image}`;
+  } else {
+    return `https://hub.docker.com/_/${image}`;
+  }
+};
```
krancour, line 9: Some of this looks iffy to me.

I know an image like `docker.io/krancour/kargo`, for instance, cannot be accessed at http://docker.io/krancour/kargo. That's a 404.

It _does_ work for ghcr.io.

```
} else if (image.includes('/')) {
    return `https://hub.docker.com/r/${image}`;
  } else {
    return `https://hub.docker.com/_/${image}`;
  }
 ```
 
 This part also looks iffy because there would be many images that would contain a forward slash that aren't on Docker Hub.

Comment (rbreeze) [reply ui/src/utils/url.ts:9]: For context, I think by nature this feature will have to be a bit iffy, or at least will be brittle and break if your image doesn't hold up to one of the 4 patterns we're accounting for (quay, ghcr, docker official and docker hub)

You're right that `docker.io/krancour/kargo` gives a 404, but wouldn't this image be referenced as just `krancour/kargo`? If it's common to reference it with the `docker.io` prefix I'll change the logic but I didn't think it was. Continuing with the `krancour/kargo` example, this would trip the second `if` condition and redirect properly to Docker Hub.  

You can think of the forward slash as _only_ differentiating between a Docker Hub image or Docker Official image; if we get past the first `if` condition, the image isn't from one of the registries we committed to recognizing for this feature, so we assume it is from Docker itself and just determine if it's official or not. You're right that any image in any registry might have a forward slash in its name, but I don't think it's possible to recognize every possible repository.

I'd like to introduce a new label (something like `kargo.akuity.io/repository`) which would allow someone to bypass this logic. There are some hurdles to overcome to make that work but it would allow customization for unique image repos that we couldn't support natively.

File: ui/tests/api/apimachinery.spec.ts [hunk 2]
```diff
@@ -1,23 +1,22 @@
+import { Timestamp } from '@bufbuild/protobuf';
 import { test, describe, expect } from 'vitest';
 
-import { Time } from '../../src/gen/k8s.io/apimachinery/pkg/apis/meta/v1/generated_pb.js';
-
 describe('metav1.Time', () => {
   test('fromJson', () => {
-    const time = Time.fromJson('2023-05-30T00:00:00Z');
+    const time = Timestamp.fromJson('2023-05-30T00:00:00Z');
```
rbreeze, line 6: This test was erroring because the `Time` type no longer exists

ui/src/utils/url.ts
```diff
@@ -1 +1,32 @@
...
+
+  if (parts[0] === 'library') {
+    parts.shift();
+  }
+
+  image = parts.join('/');
+
+  if (parts.length === 1) {
+    return `${protocol}${dockerHub}/${officialPath}/${image}`;
+  } else if (
```
jessesuen, line 24: I have another one:

public.ecr.aws/karpenter/controller:v0.31.2
can be seen at:
https://gallery.ecr.aws/karpenter/controller

jessesuen: Actually, no-one can easily publish to that repository so no point in adding it.

jessesuen: And here is another:

602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/coredns

will map to:

https://us-west-2.console.aws.amazon.com/ecr/repositories/private/602401143452/eks/coredns

rbreeze: Added support and a test 👍🏻

ui/src/utils/url.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  // DockerHub User
+  'jitesoft/nginx': 'https://hub.docker.com/r/jitesoft/nginx',
+  'docker.io/jitesoft/nginx': 'https://hub.docker.com/r/jitesoft/nginx',
+
+  // AWS
+  'public.ecr.aws/karpenter/controller': 'https://gallery.ecr.aws/karpenter/controller',
+  '602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/coredns':
+    'https://us-west-2.console.aws.amazon.com/ecr/repositories/private/602401143452/eks/coredns'
+} as { [key: string]: string };
```
krancour, line 30: I didn't review the whole PR, but if all of these cases pass, I think you got it right.

krancour (approved):

PR #1248, opened 2023-12-08
chore(devx): make tilt crazy fast

krancour:
I had a lot of conversations with @rbreeze last week about reasons he typically opts to run Kargo components as native processes instead of a local cluster. It was enlightening and inspired me to fix everything that was slowing Tilt down.

If using Tilt to develop in a local cluster, this PR will make image builds faster overall, achieve sub-second restarts on every component, and the UI component will even be live updated as code changes.

How it works:

1. Two new stages are added to the Dockerfile. These two new stages are not involved in the official image that is built and published by our release process. They are only for facilitating local development, and the differences between them and the stage that produces the official image are minor.

    There is one new stage for the back end and one for the front.

    The one for the back end copies a natively/locally built `kargo` binary from the host _instead_ of building the `kargo` binary in-container. (This allows the build to take advantage of your local Go installation's own cache -- a benefit that is lost when building in a container.) It also omits the front end.

    The one for the front end runs vite -- which is how @rbreeze runs the front end locally. It also omits the back end.

1. The updated `Tiltfile` has some new smarts.

    When a back end component is restarted, the `kargo` binary is built natively/locally and _then_ the image is rebuilt, targeting the new back end stage that just copies that binary. This achieves sub-second restarts on back end components.

    For local development only, Tilt serves the front end from a separate component (not the API server.) It's running vite, as previously mentioned. When front end code changes, the changes are instantly synced to the pod running vite. Vite already knows how to apply code changes without restarting. This means the UI is always live updating as changes are made and never requires a restart.

__Known issue:__ SSO will fail locally, but I know how to fix that and will handle it in a follow-up.

https://github.com/akuity/kargo/assets/1821014/c3258e0e-776c-40d5-b377-8032439a8136

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
-bin/
 docs/
 ui/node_modules/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ####################################################################################################
-# builder
+# back-end-builder
 ####################################################################################################
-FROM --platform=$BUILDPLATFORM golang:1.21.4-bookworm as builder
+FROM --platform=$BUILDPLATFORM golang:1.21.4-bookworm as back-end-builder
 
 ARG TARGETOS
 ARG TARGETARCH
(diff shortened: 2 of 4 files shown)


jessesuen (commented):
Do you think you could separate this out this into a different Dockerfile. A technique I've seen and used in the past is to have Dockerfile and Dockerfile.dev to split out the dev related stuff from Dockerfile.

Because now if someone only wants to build a kargo image, then it will unnecessarily build other stuff.

jessesuen (approved):

Issue #1215, opened 2023-12-03
error 500 where 403 would be more clear


krancour:
I've noticed recently that when I sign in using SSO and I am not mapped to any SAs, API calls I make all result in a 500.

They ought to be 403s.

We already do seem to throw the correct sort of Kubernetes status error when this happens.

I believe the issue is that this error gets wrapped many times over as it bubbles back up to the API endpoint that was invoked.

I think we need to give the API server's global error handling the smarts to unwind the error and test if the root case was a k8s `StatusError` with `Reason: metav1.StatusReasonForbidden`.

@devholic would you be willing to look at this?

PR #1266, opened 2023-12-11
feat(api): add interceptor to bubble up underlying error code

devholic:
Resolves https://github.com/akuity/kargo/issues/1215


internal/api/option/error.go
```diff
@@ -0,0 +1,92 @@
...
+	if ok := errors.As(err, &connectErr); !ok {
+		return err
+	}
+	underlyingErr := connectErr.Unwrap()
+	var statusErr *kubeerr.StatusError
+	if ok := errors.As(underlyingErr, &statusErr); ok {
+		return connect.NewError(httpStatusToConnectCode(statusErr.Status().Code), statusErr)
+	}
+	return err
+}
```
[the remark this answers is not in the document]
devholic: @krancour, thank you for the review :)

After reading your review, I've realized that handling errors with interceptors might lead to confusion and make debugging difficult. For example, an interceptor could prioritize certain underlying errors, leaving the business layer (handler functions) with no control over the response code.

IMHO, it would be clearer if, instead of requiring all handler functions to return a `connect.Error` with a specific code and make interceptors to take care of panic-like errors only. WDYT?

https://github.com/akuity/kargo/pull/1266/commits/b823b71688147771f5c54c652c9de76849e86599 introduces `getCodeFromError`, allowing handlers to use it as a fallback (to extract status code) after attempting to handle errors with best effort.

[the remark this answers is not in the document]
krancour: I liked the first attempt -- the filter based one -- better, to be honest.

`getCodeFromError` looks useful, but the problem with it is that one must always _remember_ to use it.

With the filter-based approach, the handler has control over the error code returned if it needs to it, and in cases where the handler chooses not to do that, the filter can make a good inference.

So the first approach was good, imho. Can we just cover the corner cases better?

devholic:
@krancour Changed interceptor to wrap error if the given error is `unknown` connect error or plain go error

internal/api/approve_freight_v1alpha1.go
```diff
@@ -37,7 +38,7 @@ func (s *server) ApproveFreight(
 			return nil, connect.NewError(connect.CodeNotFound,
 				fmt.Errorf("freight %q not found", freightKey.String()))
 		}
-		return nil, connect.NewError(connect.CodeInternal, err)
+		return nil, connect.NewError(connect.CodeUnknown, errors.Wrap(err, "get freight"))
```
krancour, line 41: @devholic I think we have had some miscommunication, which I take responsibility for.

My suggestion wasn't intended to be that handlers should return a connect error with code `connect.CodeUnknown`.

I intended "unknown" in a generic sense. i.e. If a "naked" `error` is returned, I propose the new error-handling logic should make a best effort to figure out what the root cause was and translate that to an appropriate connect error.

imho, the goal should be that when writing a handler, you don't need to do anything special at all in terms of error handling.

devholic: @krancour Changed to return raw error instead of wrapping error with `Unknown` status code.

internal/api/watch_promotion_v1alpha1.go
```diff
@@ -60,17 +60,17 @@ func (s *server) WatchPromotion(
 			}
 			u, ok := e.Object.(*unstructured.Unstructured)
 			if !ok {
-				return errors.Errorf("unexpected object type %T", e.Object)
+				return connect.NewError(connect.CodeInternal, errors.Errorf("unexpected object type %T", e.Object))
```
krancour, line 63: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_promotion_v1alpha1.go
```diff
@@ -60,17 +60,17 @@ func (s *server) WatchPromotion(
...
 			}
 			var promotion *kargoapi.Promotion
 			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &promotion); err != nil {
-				return errors.Wrap(err, "from unstructured")
+				return connect.NewError(connect.CodeInternal, errors.Wrap(err, "from unstructured"))
```
krancour, line 67: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_promotion_v1alpha1.go
```diff
@@ -60,17 +60,17 @@ func (s *server) WatchPromotion(
...
 			}
 			if err := stream.Send(&svcv1alpha1.WatchPromotionResponse{
 				Promotion: typesv1alpha1.ToPromotionProto(*promotion),
 				Type:      string(e.Type),
 			}); err != nil {
-				return errors.Wrap(err, "send response")
+				return connect.NewError(connect.CodeInternal, errors.Wrap(err, "send response"))
```
krancour, line 73: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_promotions_v1alpha1.go
```diff
@@ -55,11 +55,11 @@ func (s *server) WatchPromotions(
 			}
 			u, ok := e.Object.(*unstructured.Unstructured)
 			if !ok {
-				return errors.Errorf("unexpected object type %T", e.Object)
+				return connect.NewError(connect.CodeInternal, errors.Errorf("unexpected object type %T", e.Object))
```
krancour, line 58: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_promotions_v1alpha1.go
```diff
@@ -55,11 +55,11 @@ func (s *server) WatchPromotions(
...
 			}
 			var promotion *kargoapi.Promotion
 			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &promotion); err != nil {
-				return errors.Wrap(err, "from unstructured")
+				return connect.NewError(connect.CodeInternal, errors.Wrap(err, "from unstructured"))
```
krancour, line 62: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_promotions_v1alpha1.go
```diff
@@ -70,7 +70,7 @@ func (s *server) WatchPromotions(
 				Promotion: typesv1alpha1.ToPromotionProto(*promotion),
 				Type:      string(e.Type),
 			}); err != nil {
-				return errors.Wrap(err, "send response")
+				return connect.NewError(connect.CodeInternal, errors.Wrap(err, "send response"))
```
krancour, line 73: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_stages_v1alpha1.go
```diff
@@ -61,17 +61,17 @@ func (s *server) WatchStages(
 			}
 			u, ok := e.Object.(*unstructured.Unstructured)
 			if !ok {
-				return errors.Errorf("unexpected object type %T", e.Object)
+				return connect.NewError(connect.CodeInternal, errors.Errorf("unexpected object type %T", e.Object))
```
krancour, line 64: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_stages_v1alpha1.go
```diff
@@ -61,17 +61,17 @@ func (s *server) WatchStages(
...
 			}
 			var stage *kargoapi.Stage
 			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &stage); err != nil {
-				return errors.Wrap(err, "from unstructured")
+				return connect.NewError(connect.CodeInternal, errors.Wrap(err, "from unstructured"))
```
krancour, line 68: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_stages_v1alpha1.go
```diff
@@ -61,17 +61,17 @@ func (s *server) WatchStages(
...
 			}
 			if err := stream.Send(&svcv1alpha1.WatchStagesResponse{
 				Stage: typesv1alpha1.ToStageProto(*stage),
 				Type:  string(e.Type),
 			}); err != nil {
-				return errors.Wrap(err, "send response")
+				return connect.NewError(connect.CodeInternal, errors.Wrap(err, "send response"))
```
krancour, line 74: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_warehouses_v1alpha1.go
```diff
@@ -61,17 +61,17 @@ func (s *server) WatchWarehouses(
 			}
 			u, ok := e.Object.(*unstructured.Unstructured)
 			if !ok {
-				return errors.Errorf("unexpected object type %T", e.Object)
+				return connect.NewError(connect.CodeInternal, errors.Errorf("unexpected object type %T", e.Object))
```
krancour, line 64: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_warehouses_v1alpha1.go
```diff
@@ -61,17 +61,17 @@ func (s *server) WatchWarehouses(
...
 			}
 			var warehouse *kargoapi.Warehouse
 			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &warehouse); err != nil {
-				return errors.Wrap(err, "from unstructured")
+				return connect.NewError(connect.CodeInternal, errors.Wrap(err, "from unstructured"))
```
krancour, line 68: Is this a place where you meant to return a raw error and missed it?

internal/api/watch_warehouses_v1alpha1.go
```diff
@@ -61,17 +61,17 @@ func (s *server) WatchWarehouses(
...
 			}
 			if err := stream.Send(&svcv1alpha1.WatchWarehousesResponse{
 				Warehouse: typesv1alpha1.ToWarehouseProto(*warehouse),
 				Type:      string(e.Type),
 			}); err != nil {
-				return errors.Wrap(err, "send response")
+				return connect.NewError(connect.CodeInternal, errors.Wrap(err, "send response"))
```
krancour, line 74: Is this a place where you meant to return a raw error and missed it?

krancour:
@devholic it also looks like there might be several remaining cases where a handler does a `apierrors.IsNotFound(err)` test to explicitly return a connect error with `connect.CodeNotFound`.

That explicit handling can be removed now, right?

internal/api/watch_promotion_v1alpha1.go
devholic: No, since these errors are explicit `internal` errors, I changed to return `internal` status code instead of returning raw error.

devholic:
> @devholic it also looks like there might be several remaining cases where a handler does a `apierrors.IsNotFound(err)` test to explicitly return a connect error with `connect.CodeNotFound`.
> 
> That explicit handling can be removed now, right?

@krancour I'm still not sure whether it's okay to remove explicit error handling in handlers.

- Each handler tests will be affected by `interceptors`.
  - If there are changes in `interceptors`, all handler test cases can be affected (where checking status code)
- There can be an inconsistency when writing handler test codes - to check returned error types

internal/api/watch_promotion_v1alpha1.go
krancour: I'm not sure I understand. Isn't the handler's whole purpose to make sense of whatever error is returned to it?

* If it's a connect error, no interpretation is necessary.
* If it's a k8s error, map it to a connect error.
* If it's anything else, it was completely unanticipated and therefore amounts, by default, to an internal service error.

Am I missing something?

devholic: I want to make things clear - I think the `handler`'s purpose is to handle request and return appropriate _response_ with explicit status code to the client.

From that perspective, if the handler returns non-connect error, the error should be treated as some kind of _RuntimeException_ - so the status code should be `unknown` instead of `internal` since the handler doesn't expected such error. In this case, we tried to cast object's type but failed - the handler expected this error so the error code should be `internal`, not `unknown`.

devholic: \+ https://github.com/akuity/kargo/pull/1266#issuecomment-1874249046

Yes, handlers can be tested without interceptor. However, IMHO, returning raw k8s error and mapping error codes in interceptor will cause uncertainty of API responses since the handler doesn't know how the status code will be mapped to. 

> I think the handler's purpose is to handle request and return appropriate response with explicit status code to the client.

Which makes us not able to check response's status code. We can assert error `type` in tests for sure, but I believe it will make hard to write test code since it should deeply know the implementation of called functions/methods in handler function, not the abstracted error.

krancour: Ok... I think there are a few different things going on here that we need to separate and deal with more or less independently:

1. > I think the handler's purpose is to handle request and return appropriate response with explicit status code to the client.

    It makes total sense for handlers to be explicit with the error codes they return _in certain cases where specific failures are anticipated_. As an example, if validation of a request failed, it would make sense for a handler to explicitly return something like "bad request" (I think connect calls it "invalid argument").
    
    In the general case -- i.e. an unanticipated error -- it is onerous for developers to explicitly specify an error code. Requiring them to do so also creates lots of opportunity for human error. _It is too easy to forget._

2. If you accept no. 1, then the role of the interceptor is to try to make sense of unanticipated errors -- which most errors are likely to be.

    > if the handler returns non-connect error, the error should be treated as some kind of RuntimeException
    
    Agree.
    
    > so the status code should be unknown instead of internal since the handler doesn't expected such error.

    This is where we seem to disagree. "I didn't expect this to happen" is different from "I don't know what happened" and "internal server error" has always been the conventional way of communicating to a client that the failure was something unexpected on the server's end. Can you explain why this should be "unknown" instead?

3. I think since there are places in handlers where you are explicitly choosing  "internal server error," you must also perceive there to be some difference between "internal server error" and "unknown," but its not clear to me what's different between the spots where you are explicit and where you are not.

4. I think the interceptor is not currently handling the case of an error that _wasn't_ a Kubernetes error or a connect error. It should, and per no.2, I believe such cases should default to "internal server error."

5. I don't understand why the tests for the interceptor are so complex and standing up an HTTP server and using dummy handlers. The logic that maps errors to to connect errors could be tested by directly invoking the `wrapError()` function. I think the test is also missing many cases -- such as the errors mentioned in no.4 and also errors that have been wrapped using `errors.Wrap()`.

6. Sort of separate from the rest of it:

    > Yes, handlers can be tested without interceptor. However, IMHO, returning raw k8s error and mapping error codes in interceptor will cause uncertainty of API responses since the handler doesn't know how the status code will be mapped to.

    Returning to no.1, interpretation of errors isn't a handler's task.
    
    In general, I'd like us to embrace true _unit_ testing where we can. i.e. Testing components in isolation. Test cases simulate the layer "above" that is (directly, whenever possible) invoking the component/function under test, with dependencies _of_ the component/function under test mocked out (wherever practical). Integration tests have their place, and I am certain that we will eventually build a suite of e2e tests, but the distinct advantage of testing components in isolation, to the extent possible, is that it tremendously narrows down the source of problems and makes it easier to exercise edge cases.

Despite the length of this comment... I think this PR is very close to where it needs to be. I think we just need to come to some agreement on a few nitty gritty details and then things will fall into place.

internal/api/option/error.go
```diff
@@ -0,0 +1,91 @@
...
+func (i *errorInterceptor) wrapError(err error) error {
+	var connectErr *connect.Error
+	if ok := errors.As(err, &connectErr); ok {
+		return err
+	}
+	var statusErr *kubeerr.StatusError
+	if ok := errors.As(err, &statusErr); ok {
+		return connect.NewError(httpStatusToConnectCode(statusErr.Status().Code), statusErr)
+	}
+	return err
```
krancour, line 59: If we get to here, don't we want to default to an "internal server error" code?

internal/api/option/error.go
```diff
@@ -0,0 +1,91 @@
...
+) connect.StreamingHandlerFunc {
+	return func(ctx context.Context, conn connect.StreamingHandlerConn) error {
+		if err := next(ctx, conn); err != nil {
+			return i.wrapError(err)
+		}
+		return nil
+	}
+}
+
+func (i *errorInterceptor) wrapError(err error) error {
```
krancour, line 50: Nit on naming. "wrap" makes sense for things like `WrapStreamingClient` and other functions above, because those functions are, indeed, wrapping a client, handler, etc. with additional functionality.

I'm not sure if it makes sense here with `wrapError`, because that's not what this function does -- or at least it doesn't in all cases. I think it's function would be more clear if it were named something like `toConnectError`.

internal/api/option/error.go
```diff
@@ -0,0 +1,91 @@
...
+	case http.StatusInternalServerError:
+		return connect.CodeInternal
+	case http.StatusNotImplemented:
+		return connect.CodeUnimplemented
+	case http.StatusServiceUnavailable:
+		return connect.CodeUnavailable
+	case http.StatusGatewayTimeout:
+		return connect.CodeDeadlineExceeded
+	default:
+		return connect.CodeUnknown
```
krancour, line 89: I still feel weird about "unknown." I don't think there are many cases of unexpected errors, if any at all, where we don't _know_ what happened. The logs will say.

internal/api/option/error_test.go
```diff
@@ -0,0 +1,120 @@
...
+			opt := connect.WithInterceptors(newErrorInterceptor())
+			mux := http.NewServeMux()
+			handler := newTestErrorInterceptorServer(tc.handlerFunc)
+			mux.Handle(svcv1alpha1connect.NewKargoServiceHandler(handler, opt))
+			srv := httptest.NewServer(mux)
+			srv.EnableHTTP2 = true
+			t.Cleanup(srv.Close)
+
+			cli := svcv1alpha1connect.NewKargoServiceClient(srv.Client(), srv.URL, connect.WithGRPC())
+			_, err := cli.GetVersionInfo(
```
krancour, line 108: Why don't we just call `wrapError()` directly to see what it does for different `tc.err`?

internal/api/option/error.go
krancour: Tiniest nit -- the receiver `i` isn't used in this function. Eliminating it would make it the tiniest bit easier to test it. To be clear, I wouldn't care if this _didn't_ change, but I'm only pointing it out because I think it's a good practice to structure code such that it's maximally easy to test.

internal/api/option/error.go
devholic: Connect will care about this case - ([docs](https://connectrpc.com/docs/go/errors/#working-with-errors))

> Connect handlers attach status codes to errors using the NewError function. Handlers should return coded errors; if they don't, Connect will use the deadline_exceeded code for context.DeadlineExceeded, canceled for context.Canceled, and unknown for all other errors.

And other interceptor _may_ affect status code as we did like `errorInterceptor` - so I think we can just return raw error

internal/api/option/error_test.go
devholic: Since the subject of test is `interceptor`, and I think test code doesn't have to know about implementation details, it would be better to test interceptor flow to make sure it works.

internal/api/option/error.go
krancour: > and unknown for all other errors

This is the sticky point that keeps coming up.

It is illuminating, to see the documentation you linked to and finally realize (💡) that "unknown" may be _conventional_ or at least _common_ for connect when dealing with unanticipated errors.

When you zoom out and stop looking at the implementation detail that is connect, I still find it doesn't make good sense. Being unanticipated isn't the same as being unknown. We'll always know what happened. It will be in the logs.

> And other interceptor may affect status code

Our only other interceptor is the log interceptor and it does not affect status.

If we add another that _may_ affect status, we would have to be careful about that, but isn't it pretty unlikely that we would do so? It's not unlikely that we'd add another interceptor. I think it's unlikely that we'd add one that obscures the error received from the handler.

internal/api/option/error_test.go
krancour: > Since the subject of test is interceptor

With a majority of the logic being in the `toConnectError()` function (formerly `wrapError()`), it seems that if the function were viewed as the test subject instead of the entire type, validating the important logic would be accomplished more simply.

_I'll concede that you squeeze out slightly better test coverage with your approach, so I won't die on this hill._

> I think test code doesn't have to know about implementation details

Kargo uses whitebox testing pretty extensively. Knowledge of the code under test often leads to better coverage of edge cases.

krancour:
Code looks good @devholic. I'm going to test drive e2e Monday morning.

krancour:
I tried this with a variety of error cases just now. Overall, it's working great, but I did notice that 422 (unprocessable content/entity) is something k8s may sometimes return when, for instance, a webhook or other admission control has found something invalid in a resource, but that's a status code the interceptor isn't mapping.

@devholic I think maybe we should handle that case the same as "bad request." I couldn't find a connect error code that was a better mapping than that.

Easy way to see this in action is through `kargo create project FOO`, which will fail because Kubernetes doesn't like caps in resource names. Right now it comes through as an internal error due to lack of mapping mentioned above.

krancour (approved):
This now shows the anticipated error type for every error I've been able to force. 🔥 

Thank you @devholic!

Issue #1036, opened 2023-10-27
Running and Pending promotions show same spinner


jessesuen:
# Description

Here is a Running promotion:
<img width="731" alt="image" src="https://github.com/akuity/kargo/assets/12677113/41c1315d-fbdd-4e9e-8a4c-b5fe1a4ad94b">

Here is a Pending promotion:
<img width="705" alt="image" src="https://github.com/akuity/kargo/assets/12677113/1b9bfd83-9221-4599-a70a-69c50e2a72c5">

And for Promotions that do not have `status.phase`, we show it in UI as `Unknown`:

<img width="701" alt="image" src="https://github.com/akuity/kargo/assets/12677113/dd423e12-2204-4376-91b5-f686bbee8512">

Changes needed:

1. Pending and Running should show different icons. The spinning icon for Pending is misleading because it shows motion, which usually indicates progress.

2. Second, instead of showing `Unknown` when a promo has an empty status.phase, we should show this as `Pending` instead

# Version

20231026-unstable

```
Paste the output from `kargo version` here.
```

# Logs

```
Paste any relevant application logs here.
```

Issue #640, opened 2023-09-05
Pull request (GitHub) promotion mechanism


jessesuen:
Today, we support a git promotion, which will make a commit directly to a branch, followed by the Argo CD sync.

We need to support a promotion that goes through GitOps PR approvals. So instead of doing a direct git commit, the promotion would:

1. Open a pull request against the branch
2. Wait for that PR to be merged or rejected
3. Fail the promotion if the PR was closed without merging
4. Otherwise, once merged, perform Argo CD sync

To start, we will support GitHub, with support for other git providers later.

krancour:
> with support for other git providers later.

If we don't highly prioritize those other providers, this might be a really good area to enlist help from the community.

jessesuen:
@alexmt has questioned what it means for a promotion to be "long-lived". The PR use case is my rationale for the need for long-lived promotion. IMO, I think it's justification enough because the act of waiting for the PR to be merged, and syncing the app, is part of the overall promotion process. Also, with a fire-and-forget-based approach, I don't think we can effectively achieve the semantics of: only one promotion process should be happening at a time.

I think I am willing to cede that testing and analysis is something that probably should *not* be part of Promotion job (in favor of a more [continuous](https://github.com/akuity/kargo/issues/756) approach), but I think the PR create + wait for merge is strong justification for a long-lived promotion job.

@alexmt - since you had ideas of something more event-driven (fire and forget). I will put the onus on you to describe how you propose the UX to be for the PR use case.

krancour:
@jessesuen assigned to you because I _think_ you were already working on this. Correct me if I am wrong.

PR #1274, opened 2023-12-14
feat(promo)!: support pull request promotions

jessesuen:
Resolves https://github.com/akuity/kargo/issues/640
Resolves https://github.com/akuity/kargo/issues/1036

For git promotions, adds the option to create a PR against the writeBranch instead of committing directly to it:

```yaml
  promotionMechanisms:
    argoCDAppUpdates:
    - appName: guestbook-simple-dev
    gitRepoUpdates:
    - kustomize:
        images:
        - image: ghcr.io/jessesuen/guestbook
          path: env/dev
      pullRequest: {}   # create a PR against master instead of committing directly
      repoURL: https://github.com/jessesuen/kargo-simple.git
      writeBranch: master
```

1. Introduced `gitprovider` package for dealing with git providers in a generic way (this will need to be extracted to shared library to be used by kargo-render)
1. Introduced `Failed` promo phase. It's only usage is when a PR is closed without merging, the promo is considered Failed and not Errored.
1. Refreshing a Stage will also refresh its currently running promotion (if it exists)
1. Renamed `Promotion.Status.Error` field to `Message` so that it can communicate non-errors
1. Introduced `Promotion.Status.Metadata` map so long-running promotions can bookkeep
1. Refactored promotion framework so that mechanisms are expected to return a new PromotionStatus, which will be patched against the promo. This means promoMechanisms are also expected to return Successful, Failed, Running phase as part of the returned status
1. When multiple promoMechanisms are used in a composit, we do not proceed to the next mechanism if the first is still running. This is because PRs need to be merged before the argo-cd sync.

## Demo:

https://github.com/akuity/kargo/assets/12677113/77b86c49-a302-40ef-811d-dc4302fef9f3

## TODO:
* unit tests

Not done in this PR:
* kargo-render support
* periodic polling (resync of promo jobs) to monitor when PR is merged. Requires stage refresh
* webhook support to fast-track stage refresh


cmd/controlplane/controller.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/akuity/kargo/internal/controller/stages"
 	"github.com/akuity/kargo/internal/controller/warehouses"
 	"github.com/akuity/kargo/internal/credentials"
+	_ "github.com/akuity/kargo/internal/gitprovider/github"
```
jessesuen, line 22: I am using the technique of anonymous imports so that providers can "register" themselves through their respective init() functions. This avoids an import cycle, or the need for an extra instantiation package that knows about all providers.

internal/kargo/kargo.go
```diff
@@ -114,3 +114,45 @@ func (p PromoWentTerminal) Update(e event.UpdateEvent) bool {
...
 	return false
 }
+
+// IgnoreClearRefreshUpdates is a predicate that filters out update events if
+// the update was only to clear the refresh annotation. This prevents the
+// extra reconciliation that happens when reconcilers respond to refresh by
+// updating the object's annotations.
+type IgnoreClearRefreshUpdates struct {
+	predicate.Funcs
+}
```
jessesuen, line 124: This predicate eliminates the double reconcile that happens when user clicks refresh (which sets the refresh annotation), and then when the controller clears the refresh (which also modifies the annotation)

krancour: Is there any chance of making this piece a separate PR? I'd rubber stamp that right away since I'd love having that improvement as soon as possible.

jessesuen: Yea no problem. This PR may have overstepped a bit. I have a habit of fixing problems when it impacts the current work I'm doing.

krancour:
Is `internal/api/types/v1alpha1/types.go` possibly missing type conversions for some of the new types?

internal/gitprovider/github/github.go
```diff
@@ -0,0 +1,164 @@
...
+	githubRegistration = gitprovider.ProviderRegistration{
+		Predicate: func(repoURL string) bool {
+			u, err := url.Parse(repoURL)
+			if err != nil {
+				return false
+			}
+			// The assumption here is that no one would ever put "github" in
+			// the DNS name of their self-hosted git provider. There is also
+			// the chance that they self host github but give it a name like:
+			// 'git.mycompany.com', in which case we would return false
```
krancour, line 26: Isn't there a good possibility that someone self-hosting GitHub (e.g. GitHub Enterprise) _does_ have the word "github" in a subdomain?

internal/gitprovider/github/github.go
```diff
@@ -0,0 +1,164 @@
...
+
+func init() {
+	gitprovider.RegisterProvider("github", githubRegistration)
+}
+
+type GitHubProvider struct {
+	client *github.Client
+}
+
+func NewGitHubProvider() (*GitHubProvider, error) {
```
krancour, line 43: I think it would be an improvement if this function signature returned a `gitprovider.GitProviderService` interface instead of directly returning a specific implementation.

By doing so, the implementation can also be made unexported and implementation details, therefore, neatly hidden from callers.

internal/gitprovider/registry.go
```diff
@@ -0,0 +1,50 @@
...
+	registeredProviders = map[string]ProviderRegistration{}
+)
+
+// NewGitProviderServiceFromName returns a git provider service by it's registered name
+func NewGitProviderServiceFromName(name string) (GitProviderService, error) {
+	if reg, ok := registeredProviders[name]; ok {
+		return reg.NewService()
+	}
+	return nil, fmt.Errorf("No registered providers with name %q", name)
+}
```
krancour, line 29: This appears to be unused. Is there a scenario where we need to find the correct provider implementation by name rather than by URL?

internal/gitprovider/gitprovider.go
```diff
@@ -0,0 +1,61 @@
+package gitprovider
```
krancour, line 1: 🤔 I could see the entire contents of this package being folded into the existing `git` package.

That package, already has the likes of `git.Repo`, and I think that `git.Provider` and `git.ProviderRegistry`, being the generic things they are, fit neatly in there.

Just a thought aimed at combatting the sprawl of many micro-packages.

Not a hill I would die on though.

internal/controller/git/git.go
```diff
@@ -251,6 +274,30 @@ func (r *repo) CreateOrphanedBranch(branch string) error {
...
 }
 
+func (r *repo) CurrentBranch() (string, error) {
+	resBytes, err := libExec.Exec(r.buildCommand(
+		"branch",
+		"--show-current",
+	))
+	return strings.TrimSpace(string(resBytes)),
+		errors.Wrap(err, "error determining current branch")
+}
```
krancour, line 284: All existing operations that change the current branch (`Checkout()`, for instance) update `repo.currentBranch`, so you can just return that.

internal/controller/git/git.go
```diff
@@ -48,19 +48,27 @@ type Repo interface {
 	Checkout(branch string) error
 	// Commit commits staged changes to the current branch.
 	Commit(message string) error
+	// CommitsHaveDiffs returns whether there is a diff between two commits/branches
+	CommitsHaveDiffs(commit1 string, commit2 string) (bool, error)
```
krancour, line 52: Nit: Since this is described as also being effective for comparing differences between branches, it's possible that the name `RefsHaveDiffs` might be more inclusive of the ways this could be used.

internal/controller/promotion/composite.go
```diff
@@ -41,30 +43,87 @@ func (c *compositeMechanism) GetName() string {
...
+		newStatus = aggregateGitPromoStatus(newStatus, *otherStatus)
+		if newStatus.Phase == kargoapi.PromotionPhaseSucceeded {
+			// We only continue to the next promotion mechanism if the current
+			// mechanism succeeded. This is because a PR must be merged before
+			// performing the ArgoCD sync.
+			continue
+		}
+		// Current promotion mechanism is either Failed/Errored/Running.
+		// We stop executing promotion mechanisms and return the current status.
+		break
```
krancour, line 80: This logic simplifies to:

```
if newStatus.Phase != kargoapi.PromotionPhaseSucceeded {
  break
}
```

internal/controller/promotion/composite.go
```diff
@@ -41,30 +43,87 @@ func (c *compositeMechanism) GetName() string {
...
+		otherStatus, newFreight, err = childMechanism.Promote(ctx, stage, promo, newFreight)
 		if err != nil {
-			return newFreight, errors.Wrapf(
+			return nil, newFreight, errors.Wrapf(
 				err,
 				"error executing %s",
 				childMechanism.GetName(),
 			)
 		}
+		newStatus = aggregateGitPromoStatus(newStatus, *otherStatus)
```
krancour, line 71: I might be overlooking something here, so apologies in advance if I am...

I'm not sure we need this `aggregateGitPromoStatus` function. Since the only way to continue on to another iteration of the loop is for the current iteration to have succeeded, the severity of `otherStatus`'s phase is _always_ going to be >= that of `newStatus`, meaning we could skip any fancy logic for collapsing two statuses into one.

Or am I missing something?

internal/controller/promotion/git.go
```diff
@@ -87,73 +87,113 @@ func (g *gitMechanism) GetName() string {
...
+			return nil, newFreight, err
+		}
+	} else {
+		// For git commit promotions, promotion is successful as soon as the commit is pushed.
+		newStatus.Phase = kargoapi.PromotionPhaseSucceeded
 	}
 
 	if commitIndex > -1 {
 		newFreight.Commits[commitIndex].HealthCheckCommit = commitID
 	}
```
krancour, line 194: There may be another change needed here. If I am reading all new logic above this correctly, in the case of a PR, commitID remains set to the commit at the tip of the PR branch. If `reconcilePullRequest` determines that the PR has been merged, it is the commit now at the tip of the target (stage-specific) branch that should be the new `HealthCheckCommit`. (i.e. The commit that the `Stage` reconciler will be looking for Argo CD Apps to be synced to.)

krancour: Out of the whole PR, this would be my biggest concern.

krancour:
@jessesuen I know some unit test fixes are still outstanding, but from what I've seen of the WIP, this is fantastic. I really like the approach. My comments are mostly small nits or questions. Only [this one](https://github.com/akuity/kargo/pull/1274#discussion_r1430684976) amount to any serious concern.

internal/gitprovider/github/github.go
jessesuen: I think I need to revise this comment but basically it's trying explain that: if we see github in the hostname, we can safely assume that the git provider is github. If the word github **_doesn't_** appear in the hostname, we will return false, and not detect that github is the provider. The comment explains that someone might be hosting GitHub but not incorporate the word "github" into the name and then we would miss it.

> The assumption here is that no one would ever put "github" in the DNS name of their self-hosted git provider.

I'll rephrase this to say:
> We assume that any hostname with the word "github" in the hostname, can use this provider.

internal/gitprovider/registry.go
jessesuen: Oh yes, I added this even though it is unused because once we have 2+ providers, none of them using publically hosted well-known URLs (github.com, gitlab.com, bitbucket.com, etc...), then users can explicitly instantiate it by name:

This syntax is actually already added in the spec:

```yaml
    gitRepoUpdates:
    - pullRequest:
        github: {}
```

However, it's not yet used because we don't have two providers and so there wasn't a need to instantiate it by name. 

That said, I think I should add the code to instantiate by name, so that the next person to add support for the next provider will know what to do. I will make that change.

internal/gitprovider/gitprovider.go
jessesuen: I think I agree with this, But I'd like to revisit combing this when we extract this out. I think providers may be a subpackage of a git package we split out.

internal/controller/git/git.go
jessesuen: Actually, that was my first thought. But the reason this was necessary was because Checkout() is commonly called with an empty string and so `currentBranch` was being set to empty string.

If you'd like, I can improve checkout to also call `git branch --show-current` when empty string, and then this method becomes an accessor method to `currentBranch`.

internal/controller/promotion/composite.go
jessesuen: You are right, for this case where we short-circuit the composite mechanism as soon as we get a `Failed/Errored/Running` status it is unecessary. But `aggregateGitPromoStatus` is also called in `git::Promote`, where short-circuiting does not happen.

At the very least, `aggregateGitPromoStatus` in composite.go is unnecessary and should be removed. Good catch!

As for git.go, I think I should either short-circuit the loop (and remove`aggregateGitPromoStatus` altogether) or continue aaggregation without short-circuiting. Any preference?

internal/controller/promotion/git.go
jessesuen: Great catch. It is a bug.

internal/gitprovider/github/github.go
krancour: Ah. Ok... now I understand what the intent of the comment was meant to be. 👍 

fwiw, I do I have some _vague_ recollection from previous projects about GitHub Enterprise being a somewhat different beasty API-wise from regular GitHub, but since I cannot recall specifics, that's something we can revisit at a later point if we determine its necessary.

internal/gitprovider/gitprovider.go
krancour: > I think providers may be a subpackage

I agree with that. Interfaces in the existing git package, sub-packages for specific provider implementations.

internal/controller/git/git.go
krancour: Oh... I forgot about that. If checkout is called with the empty string, you're just getting the trunk and you don't necessarily know what it's named.

> I can improve checkout to also call git branch --show-current when empty string

That seems like the right thing to do.

internal/controller/promotion/composite.go
krancour: No preference other than whatever you feel ends up being simpler.

jessesuen: Turns out I still need this because aggregateGitPromoStatus also has the responsibility of merging the metadata map.

jessesuen:
> Is `internal/api/types/v1alpha1/types.go` possibly missing type conversions for some of the new types?

It was. Thank you for pointing it out. Adding it was... tedius/painful, to say the least. Definitely need to do something about https://github.com/akuity/kargo/issues/708.

internal/controller/promotion/git_test.go
```diff
@@ -160,92 +160,92 @@ func TestGitDoSingleUpdate(t *testing.T) {
...
+		// 		status *kargoapi.PromotionStatus,
+		// 		newFreightIn kargoapi.SimpleFreight,
+		// 		newFreightOut kargoapi.SimpleFreight,
+		// 		err error,
+		// 	) {
+		// 		require.Error(t, err)
+		// 		require.Equal(t, "something went wrong", err.Error())
+		// 		require.Equal(t, newFreightIn, newFreightOut)
+		// 	},
+		// },
```
krancour, line 248: Are these supposed to stay commented?

krancour: I see that you probably commented these to isolate the success case.

I took the liberty of uncommenting these and amending the PR.

krancour (approved):
🔥

PR #1279, opened 2023-12-14
fix: eliminate double reconcile during refresh requests

jessesuen:
Currently when someone clicks refresh on a Stage, Warehouse, we get two reconciles:
1. first reconcile happens because refresh annotation was set by kargo-api
2. second reconcile happens because controller patches the objects annotation to clear the refresh annotation

This introduces a predicate which will filter out the event if the update event was to clear the annotation, preventing the second reconcile.


krancour (approved):
Awesome

PR #1282, opened 2023-12-15
feat(api): treat sub and email fields of oidc tokens as distinct things; enable SA mappings by either

### LINKED ISSUE #1278: Consult `rbac.kargo.akuity.io/email` as a scope
# Proposed Feature

In v0.3.0, we will support mapping of `sub` and `groups` OIDC fields:

```yaml
metadata:
  annotations:
    rbac.kargo.akuity.io/sub: jessesuen
    rbac.kargo.akuity.io/groups: myorg:kargoadmin
```

But we should also support `email`, which is a standard OIDC scope.

# Motivation

`sub` is often an opaque user ID. But emails are not. In order to improve the UX of giving individuals access, we need to support the `email` scope.

# Suggested Implementation

Similar implementation as sub:

https://github.com/akuity/kargo/blob/4e28102659e94c151c751ad86344a9f40f829882/internal/kubeclient/indexer.go#L307-L330

krancour:
Fixes #1278

### DIFF
--- a/internal/api/kubernetes/client.go
+++ b/internal/api/kubernetes/client.go
@@ -553,7 +553,7 @@ func getAuthorizedClient(globalServiceAccountNamespaces []string) func(
 			Name:        key.Name,
 		}
 
-		if userInfo.Username != "" {
+		if userInfo.Subject != "" {
 			var namespacesToCheck []string
 			if key.Namespace != "" {
 				namespacesToCheck = make([]string, 0, len(globalServiceAccountNamespaces)+1)
--- a/internal/api/option/auth.go
+++ b/internal/api/option/auth.go
@@ -263,10 +263,10 @@ func (a *authInterceptor) listServiceAccounts(
 ) (map[string]map[types.NamespacedName]struct{}, error) {
 	queries := []libClient.MatchingFields{
 		{
-			kubeclient.ServiceAccountsByOIDCEmailIndexField: c.Subject,
+			kubeclient.ServiceAccountsByOIDCSubjectIndexField: c.Subject,
 		},
 		{
-			kubeclient.ServiceAccountsByOIDCSubjectIndexField: c.Subject,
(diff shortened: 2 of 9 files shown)


internal/api/kubernetes/client.go
```diff
@@ -553,7 +553,7 @@ func getAuthorizedClient(globalServiceAccountNamespaces []string) func(
 			Name:        key.Name,
 		}
 
-		if userInfo.Username != "" {
+		if userInfo.Subject != "" {
```
gdsoumya, line 556: Do we need to consider Email too?

internal/api/option/auth.go
```diff
@@ -260,17 +259,19 @@ func (a *authInterceptor) WrapStreamingHandler(
...
 func (a *authInterceptor) listServiceAccounts(
 	ctx context.Context,
-	username string,
-	groups []string,
+	c claims,
 ) (map[string]map[types.NamespacedName]struct{}, error) {
 	queries := []libClient.MatchingFields{
 		{
-			kubeclient.ServiceAccountsBySubjectIndexField: username,
+			kubeclient.ServiceAccountsByOIDCEmailIndexField: c.Subject,
```
jessesuen, line 266: Is this a bug? Should this be c.Email?

internal/api/kubernetes/client.go
krancour: Good question. "sub" and "email" are both standard claims. We'll either have a value for both or for neither. So we only need to check one to know what kind of scenario we're dealing with.

internal/api/option/auth.go
krancour: 😓 

Nice catch.

krancour: Fixed in b5ad67caa1c15c894428e63020fe1bcdee1093b5

jessesuen (approved):
LGTM!

PR #1288, opened 2023-12-18
fix: helm repo regex to allow -

Brightside56:
This PR needed because Helm repo regex currently isn't allowing `-` in repoURL


krancour (approved):
LGTM pending CI.

krancour:
Thank you for this @Brightside56 !

PR #1289, opened 2023-12-18
feat: display warehouse and its subscription as separate nodes in the UI

### LINKED ISSUE #1275: DAG: Add warehouse node to the left of subscription nodes
Currently, subscriptions are shown as their own nodes, but there is a lot of redundant information when they belong to the same warehouse. This has caused some confusion. To fix this, we will add a new parent node for subscriptions that contains warehouse information, while simplifying the subscription nodes.

rbreeze:
Fixes #1275 

With subscription nodes expanded:
CleanShot 2023-12-18 at 14 34 27@2x

With subscription nodes collapsed:
CleanShot 2023-12-18 at 14 34 18@2x

### DIFF
--- a/ui/src/features/project/project-details/nodes/repo-node.module.less
+++ b/ui/src/features/project/project-details/nodes/repo-node.module.less
@@ -4,6 +4,7 @@
   display: flex;
   flex-direction: column;
   padding: 2px;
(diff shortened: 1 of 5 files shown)


krancour (approved):
🎉

PR #1310, opened 2023-12-20
refactor(api)!: update annotations used for user -> sa mappings

krancour:
Encountered this while working on docs...

Since these annotations accept multiple comma-delimited values, it seems to makes more sense if the keys end with plural nouns to reflect that.

### DIFF
--- a/api/v1alpha1/annotations.go
+++ b/api/v1alpha1/annotations.go
@@ -3,7 +3,7 @@ package v1alpha1
 const (
 	AnnotationKeyRefresh = "kargo.akuity.io/refresh"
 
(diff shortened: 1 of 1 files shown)


gdsoumya (approved):

krancour:
After offline discussion with @jessesuen, we've decided there's a benefit to keeping these annotations such that they match the corresponding OIDC claim names.

Closing.

jessesuen (changes requested):
As discussed, I prefer to keep these named the same as OIDC scope names. Because then in the future, when we add additional support for other scopes, even custom or nonstandard, then we have a convention that didn't break.

PR #1315, opened 2023-12-21
fix(controller): when updating a values.yaml, make sure image tags that look like a number are still treated as strings

### LINKED ISSUE #876: When Kargo Updates Helm Values ,is not seting quote on the tag value
when Kargo updates helm chart values (i.e image.tag)

it does not set quotes on the values, if as part of the build the tag is fully composed out of numbers helm thinks that it is an integer and it breaks the deployment 

> Error: values don't meet the specifications of the schema(s) in the following chart(s): app: - image.tag: Invalid type. Expected: > string, given: integer

after Kargo updated the git
```YAML
image:
  ## Name of the image you want to build/deploy
  ## Defaults to the release name
  name: quay.io/gotham/batman
  ## Tag that you want to build/deploy
  tag: 3108617
```

How it should set it
```YAML
image:
  ## Name of the image you want to build/deploy
  ## Defaults to the release name
  name: quay.io/gotham/batman
  ## Tag that you want to build/deploy
  tag: '3108617'
```

maksimstankevic:
Fixes #876 

image

### DIFF
--- a/internal/controller/promotion/helm.go
+++ b/internal/controller/promotion/helm.go
@@ -125,7 +125,7 @@ func buildValuesFilesChanges(
 	tagsByImage := map[string]string{}
 	digestsByImage := make(map[string]string, len(images))
 	for _, image := range images {
-		tagsByImage[image.RepoURL] = "'" + image.Tag + "'"
+		tagsByImage[image.RepoURL] = image.Tag
 		digestsByImage[image.RepoURL] = image.Digest
 	}
 	changesByFile := make(map[string]map[string]string, len(imageUpdates))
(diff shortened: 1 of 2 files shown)


internal/controller/promotion/helm.go
```diff
@@ -125,7 +125,7 @@ func buildValuesFilesChanges(
 	tagsByImage := map[string]string{}
 	digestsByImage := make(map[string]string, len(images))
 	for _, image := range images {
-		tagsByImage[image.RepoURL] = image.Tag
+		tagsByImage[image.RepoURL] = "'" + image.Tag + "'"
```
krancour, line 128: This will break things farther down on lines 155 to 161.

Just change line 160.

internal/controller/promotion/helm_test.go
```diff
@@ -321,8 +321,8 @@ func TestBuildValuesFilesChanges(t *testing.T) {
 		t,
 		map[string]map[string]string{
 			"fake-values.yaml": {
-				"fake-key":        "fake-url:fake-tag",
-				"second-fake-key": "second-fake-tag",
+				"fake-key":        "fake-url:'fake-tag'",
```
krancour, line 324: This test case demonstrates the problem I mentioned [here](https://github.com/akuity/kargo/pull/1315/files#r1434538580).

`fake-url:'fake-tag'` isn't a valid image identifier.

maksimstankevic: @krancour thank you, I fixed it https://github.com/akuity/kargo/pull/1315/commits/df36c0f753b648d7bb18953b637d694642200d42
I still had to keep single quotes in the tagOnly test case as otherwise it fails, is it OK?

krancour (approved):
Thank you @maksimstankevic!

PR #1316, opened 2023-12-21
fix(ui): hovering on a DAG node would highlight other nodes with same freight

rbreeze:
Before:
CleanShot 2023-12-21 at 12 38 04@2x

After: 
CleanShot 2023-12-21 at 12 37 46@2x


krancour (approved):
Test drove. LGTM

PR #1317, opened 2023-12-21
chore(dependencies): upgrade controller runtime and all things k8s

krancour:
Starting v0.4.0 off with a 💥.


jessesuen (approved):
Beautiful!

internal/controller/promotions/promotions.go
```diff
@@ -84,7 +84,13 @@ func SetupReconcilerWithManager(
 		pqs:         reconciler.pqs,
 	}
 	promoWentTerminal := kargo.NewPromoWentTerminalPredicate(logger)
-	if err := c.Watch(&source.Kind{Type: &kargoapi.Promotion{}}, priorityQueueHandler, promoWentTerminal); err != nil {
+	if err := c.Watch(
+		source.Kind(kargoMgr.GetCache(),
+			&kargoapi.Promotion{},
+		),
```
krancour, line 90: @jessesuen this is what I was talking about... the fact that you have to pick a cache (which is a wrapper around a client) when you call `Watch` makes me suspect we may be able to create controllers that reconcile using one client, but also watch things using another. (Useful for sharding.)

PR #1339, opened 2023-12-28
feat(chart): support additional labels

nlamirault:


krancour:
@nlamirault I adjusted the metadata on the `values.yaml` file and ran codegen, which updates the `README.md` based on the metadata.

I don't want to rush into merging this, however. I wanted to asked for some clarification on the intent behind this PR.

I can definitely see the scenario where one may wish to apply additional labels to deployments or services, for whatever reasons -- reporting, identifying an owner, etc...

As you've implemented it, those additional labels do not extend to the pods managed _by_ the deployments. For the same reasons I could see wanting to add labels of your own to deployments, I could imagine wanting those same labels on pods.

Basically, the question boils down to why one and not the other?

nlamirault:
it was a mistake @krancour 
I thought deployment template use `labels`
So I make a fix.

krancour:
@nlamirault thanks for clarifying the intent and fixing it.

I did fix an issue with the webhook server feployment, but everything else looks great.

Thank you for the contribution!

krancour (approved):

PR #1340, opened 2023-12-28
fix: using both digest and tag for Freight.ID generation to admit freights…

maksimstankevic:
… based off same image tagged differently

### DIFF
--- a/api/v1alpha1/freight_types.go
+++ b/api/v1alpha1/freight_types.go
@@ -50,6 +50,12 @@ func (f *Freight) UpdateID() {
 	for _, image := range f.Images {
 		artifacts = append(
 			artifacts,
+			// Note: This isn't the usual image representation using EITHER :<tag> OR @<digest>.
+			// It is possible to have found an image with a tag that is already known, but with a
+			// new digest -- as in the case of "mutable" tags like "latest". It is equally possible to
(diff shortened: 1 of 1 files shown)


api/v1alpha1/freight_types.go
```diff
@@ -50,7 +50,7 @@ func (f *Freight) UpdateID() {
 	for _, image := range f.Images {
 		artifacts = append(
 			artifacts,
-			fmt.Sprintf("%s@%s", image.RepoURL, image.Digest),
+			fmt.Sprintf("%s:%s@%s", image.RepoURL, image.Tag, image.Digest),
```
krancour, line 53:
```suggestion
			// Note: This isn't the usual image representation using EITHER :<tag> OR @<digest>.
			// It is possible to have found an image with a tag that is already known, but with a
			// new digest -- as in the case of "mutable" tags like "latest". It is equally possible to
			// have found an image with a digest that is already known, but has been re-tagged.
			// To cover both cases, we incorporate BOTH tag and digest into the canonical
			// representation of an image used when calculating Freight ID.
			fmt.Sprintf("%s:%s@%s", image.RepoURL, image.Tag, image.Digest),
```

krancour:
Thank you @maksimstankevic !

krancour (approved):

Issue #1341, opened 2023-12-28
all cli commands fail if config file doesn't exist already


christianh814:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

Running `kargo version` when you're not logged in returns an error

Example

```shell
$ kargo version
Error: get version info from server: unavailable: unsupported protocol scheme ""
```

# Steps to Reproduce

First:

```
rm -rf ~/.config/kargo
```

Then 

```
kargo version
```

# Expected behavior

There should be a `--client` flag that returns the version of the client when not logged into a Kargo server

Example:

```
kargo version --client
```

krancour:
It's not just `version`. It's actually _all_ commands that fail with this if you don't have an existing config file. Even `login` fails.

It appears to have been introduced in #1179 

cc @devholic

bygui86:
Hi @krancour when do you plan/think to release kargo cli v0.3.1 to fix this issue?
It's quite critical as it's not possible to test the [quickstart guide](https://kargo.akuity.io/quickstart) because not possible to create any project.
Is there an alternative way to create a Kargo Project? maybe YAML?

krancour:
@bygui86 I expect v0.3.1 to be released today.

PR #1345, opened 2024-01-02
fix(cli): handle CLI config file not found error

devholic:
Fixes https://github.com/akuity/kargo/issues/1341

Currently, the CLI entrypoint and some commands assumes that the config file exists in the home directory. So all commands and config-related commands (e.g. `config set`, `config unset`) are failing if the config file does not exist.

This commit fixes this issue by returning an default config if the config file does not exists in the CLI entrypoint and creates the config file if necessary in config-related commands.

### DIFF
--- a/internal/cli/client/client.go
+++ b/internal/cli/client/client.go
@@ -26,7 +26,7 @@ func GetClientFromConfig(ctx context.Context, opt *option.Option) (
 	}
 	cfg, err := config.LoadCLIConfig()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "load cli config")
 	}
 	skipTLSVerify := opt.InsecureTLS || cfg.InsecureSkipTLSVerify
 	if cfg, err =
--- a/internal/cli/client/client.go
+++ b/internal/cli/client/client.go
@@ -28,6 +28,11 @@ func GetClientFromConfig(ctx context.Context, opt *option.Option) (
 	if err != nil {
 		return nil, errors.Wrap(err, "load cli config")
 	}
+	if cfg.APIAddress == "" || cfg.BearerToken == "" {
+		return nil, errors.New(
+			"seems like you are not logged in; please use `kargo login` to authenticate",
+		)
+	}
 	skipTLSVerify := opt.InsecureTLS || cfg.InsecureSkipTLSVerify
 	if cfg, err =
 		newTokenRefresher().refreshToken(ctx, cfg, skipTLSVerify); err != nil {
--- a/internal/cli/cmd/config/set.go
+++ b/internal/cli/cmd/config/set.go
@@ -20,7 +20,7 @@ func newSetCommand() *cobra.Command {
 kargo config set project my-project
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cfg, err := config.EnsureCLIConfig()
+			cfg, err := config.LoadCLIConfig()
 			if err != nil {
 				return errors.Wrap(err, "ensure cli config")
 			}
--- a/internal/cli/cmd/config/set.go
+++ b/internal/cli/cmd/config/set.go
@@ -20,9 +20,9 @@ func newSetCommand() *cobra.Command {
 kargo config set project my-project
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cfg, err := config.LoadCLIConfig()
+			cfg, err := config.EnsureCLIConfig()
 			if err != nil {
-				return errors.Wrap(err, "load cli config")
+				return errors.Wrap(err, "ensure cli config")
 			}
 
 			key := strings.ToLower(args[0])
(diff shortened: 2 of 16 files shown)


krancour:
@devholic this PR creates a new problem. Empty config obviously contains no credentials, so now any commands that attempts to connect to the API server fails like this:

```
go run cmd/cli/*.go get projects
Error: get client from config
exit status 1
```

Which isn't very clear that the issue is simply that the user needs to log in.

I believe the only two commands for which we _don't_ need to be logged in are `version` (in which case, you get no server version info) and `login`.

Can we just check for and ignore `ErrConfigNotFound` in those two cases?

devholic:
> @devholic this PR creates a new problem. Empty config obviously contains no credentials, so now any commands that attempts to connect to the API server fails like this:
> 
> ```
> go run cmd/cli/*.go get projects
> Error: get client from config
> exit status 1
> ```
> 
> Which isn't very clear that the issue is simply that the user needs to log in.
> 
> I believe the only two commands for which we _don't_ need to be logged in are `version` (in which case, you get no server version info) and `login`.
> 
> Can we just check for and ignore `ErrConfigNotFound` in those two cases?

@krancour 

I believe ignoring `ErrConfigNotFound` for the specific commands will not resolve this issue since users can modify CLI config file directly without Kargo CLI.

Instead, checking required fields (`APIAddress` and `BearerToken`) in `GetClientFromConfig` function and showing the detailed error to the user more will be more clear and work even if the user removes some required field from the config file. WDYT? (https://github.com/akuity/kargo/pull/1345/commits/33cfede663849c8f24675b8f2326226dbb1f9f31)

internal/cli/client/client.go
```diff
@@ -26,7 +26,12 @@ func GetClientFromConfig(ctx context.Context, opt *option.Option) (
...
 	cfg, err := config.LoadCLIConfig()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "load cli config")
+	}
+	if cfg.APIAddress == "" || cfg.BearerToken == "" {
+		return nil, errors.New(
+			"seems like you are not logged in; please use `kargo login` to authenticate",
+		)
 	}
```
krancour, line 35: This condition will never occur. The previous action was to load the config, and that will have failed with an error if it didn't exist.

I think you can safely remove this condition for now.

Later, we should revisit the fact that config is being re-loaded when building a client. We shouldn't be doing that, but that is something we can deal with later.

devholic: No, it can occur by removing `apiAddress` or `bearerToken` field in Kargo config file.

```
❯ cat ~/.config/kargo/config            
project: kargo-demo

❯ ./cli get projects
Error: get client from config: seems like you are not logged in; please use `kargo login` to authenticate
```

internal/cli/cmd/config/set.go
```diff
@@ -20,9 +20,9 @@ func newSetCommand() *cobra.Command {
 kargo config set project my-project
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cfg, err := config.LoadCLIConfig()
+			cfg, err := config.EnsureCLIConfig()
```
krancour, line 23: This doesn't seem necessary. Can't we just let this fail if no config file exists? If no config file exists, the user is not logged in, so changing anything else in the configuration seems meaningless.

internal/cli/cmd/config/unset.go
```diff
@@ -20,9 +20,9 @@ func newUnsetCommand() *cobra.Command {
 kargo config unset project my-project
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cfg, err := config.LoadCLIConfig()
+			cfg, err := config.EnsureCLIConfig()
```
krancour, line 23: Same as above.

internal/cli/client/client.go
krancour: Oh. I see.

internal/cli/cmd/config/set.go
krancour: > but I think our UX strategy for CLI is to have same interface (or behavior) with kubectl

I'm not against that, but we're currently not very similar to kubectl in terms of configuration. Our config is a single context only and tracks the server you're currently logged into.

By contrast, kubectl's supports many contexts and switching between contexts.

With kubectl, if you `kubectl config set ...`, that is applied to the current context, which is _not forgotten_ when you switch to some other context. With `kargo`, when you `kargo login` to a different server, all the old config is forgotten...

Which is what essentially makes it (currently) meaningless to set configuration values for a server you're not logged into.

I think we should have a conversation at some point about whether we want our CLI to support multiple contexts like kubectl does. I see reasons to... I also see reasons not to. To the extent that Kargo is supposed to be a single pane of glass over many clusters, it might not be common for users to interact with more than one Kargo.

krancour (approved):
LGTM

PR #1350, opened 2024-01-03
refactor(cli): don't load config twice for commands that use a client

krancour:
Ever since #1179, we've been loading CLI config up front since it may contain a user's preference for a default project.

I noticed today that commands requiring a client for interacting with the Kargo API server continue to _reload_ the configuration just-in-time.

This change passed the config that is loaded up front to the "constructors" for each of the CLI's sub-commands so configuration is now only ever loaded once per invocation of the CLI.


rbreeze (approved):
LGTM!

PR #1351, opened 2024-01-03
refactor(cli): flag cleanup

krancour:
This PR fixes two things I noticed today:

1. There's a lot of unnecessary complexity in how we're setting up widely used flags for various CLI sub-commands. A little bit of simplification makes this much easier to follow.

1. The login command was doing flags very differently from how all other commands were, so I fixed it to be more consistent.


rbreeze (approved):
LGTM! 🙏🏻🎉

PR #1364, opened 2024-01-04
fix: configuring kubernetesDatabaseConfig with env vars

lrotim:
As kubernetesDatabaseConfig has private vars, envconf didn't fill their content from defined env vars


krancour:
@lrotim great catch! Great fix! Thank you!

krancour (approved):

PR #1365, opened 2024-01-04
feat(controller): use optional third kubeconfig for rollouts kube client

### LINKED ISSUE #1359: add supoort for a third kube config / client
Currently, a Kargo controller is configured to use _two_ different Kubernetes clients:

1. One that interacts with Kargo resources
2. One that interacts with Argo CD resources

This is meant to accommodate cases where Kargo resources and Argo CD resources don't live in the same cluster, although, in the average case, they do, and both clients actually point to the _same_ cluster.

Since the introduction of verification processes powered by Argo Rollouts AnalysesTemplates / AnalysisRuns, we've been assuming that the cluster where Argo CD resources live is _also_ running the Argo Rollouts controller _and_ is a suitable place to execute user-defined workloads, but it's gradually becoming clear that this is not always a safe assumption.

Cases where it is not:

* Kargo controllers that do not integrate with Rollouts, but _do_ integrate with Argo CD
    * Opting out of Rollouts integration entirely isn't possible yet, but will be soon
* Kargo controllers that do not integrate with Argo CD, but _do_ integrate with Rollouts
    * Opting out of Argo CD integration entirely isn't possible yet, but will be soon
    * Note: Rollouts doesn't depend on Argo CD
* Kargo controllers that integrate with a managed Argo CD service (i.e. SaaS) -- which is not likely to have Rollouts installed and is _very_ unlikely to permit execution of user-defined workloads

Scenarios such as these can be accommodated by support for a third Kubernetes client:

_3. One that interacts with AnalysisRun resources on a cluster that runs a Rollouts controller and is suitable for hosting user-defined workloads_

In the average case, as before with two client, all _three_ clients will point to the same cluster.

### LINKED ISSUE #1363: Execute `AnalysisRun` in target cluster
# Checklist

# Proposed Feature

Love that `AnalysisTemplate` support got added. However, the cluster our Argo CD runs on (and in the future, Kargo) doesn't necessarily have access to the in-cluster prometheus endpoint, or to in-cluster `Service` endpoints needed to run e2e tests.

It would be awesome if Analysis could be executed on the **destination** cluster instead of in the Kargo cluster/namespace. I know I could work around this by essentially invoking a Job with a kubeconfig that in turn invokes another Job on the remote cluster. But that seems... kludgy.

# Motivation

Run in-cluster end-to-end tests against deployed services. And being able to access in-cluster `Service` endpoints such as prometheus, which might not necessarily be publicly exposed via Ingress.

# Suggested Implementation

Kargo could borrow Cluster Secrets from Argo CD? Though that would make it difficult in light of future "remote argo cd support".

It'll also muddy the lines between Argo CD and Kargo even further.

Here's an alternative suggestion - add AnalysisTemplate support to Argo CD post-sync hooks :)

krancour:
Fixes #1359
Fixes #1363

cc @jessesuen @gdsoumya this should solve some problems for you both.

### DIFF
--- a/charts/kargo/README.md
+++ b/charts/kargo/README.md
@@ -20,18 +20,28 @@
 
 ### KubeConfigs
 
-Optionally point to Kubernetes Secrets containing kubeconfig for a remote
-Kubernetes cluster hosting Kargo resources and/or a remote Kubernetes cluster
-hosting Argo CD resources. This is useful for cases where the Kargo
-controller is running somewhere other than the cluster(s) it is managing. The
-config for Kargo and Argo CD can be the same, different, or omitted entirely.
-When omitted, the controller will fall back on in-cluster configuration.
-In the average case, these settings should be left alone.
+Optionally point to Kubernetes Secrets containing kubeconfig for:
 
-| Name                       | Description                                                                                              | Value       |
-| -------------------------- | -------------------------------------------------------------------------------------------------------- | ----------- |
-| `kubeconfigSecrets.kargo`  | Kubernetes `Secret` name containing kubeconfig for a remote Kubernetes cluster hosting Kargo resources   | `undefined` |
-| `kubeconfigSecrets.argocd` | Kubernetes `Secret` name containing kubeconfig for a remote Kubernetes cluster hosting Argo CD resources | `undefined` |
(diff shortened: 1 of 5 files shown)


gdsoumya (approved):
LGTM

PR #1366, opened 2024-01-04
feat(controller): support git repos with self-signed cert

### LINKED ISSUE #1224: Support for Self signed https git repo
# Proposed Feature

Currently, kargo supports only https secure git repositories. In an air-gapped environment, insecure self-signed https endpoints may be required. 

# Motivation

For testing Kargo in air-gapped environments.

krancour:
Fixes #1224

I verified this against gitea in a local cluster with ingress and a self-signed cert.

Worked as expected.

cc @geowalrus4gh, I think you were looking for this.


gdsoumya (approved):
LGTM

PR #1367, opened 2024-01-04
feat(ui): ability to rename freight from UI

### LINKED ISSUE #1329: add ability to change freight aliases via ui
The CLI can already do this, but we overlooked it for the UI.

### LINKED ISSUE #1371: UI: Mutations in Modals are broken
Discovered when working on https://github.com/akuity/kargo/pull/1367

Edit Stage and Create Stage modals both return the following error after calling the API:

```
{
    "code": "unauthenticated",
    "message": "no token provided"
}
```

rbreeze:
Fixes #1371
Fixes #1329 

CleanShot 2024-01-04 at 15 14 29@2x

Also improved the Freight Label tooltip:
CleanShot 2024-01-04 at 15 14 02@2x

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -4,7 +4,6 @@ import { ConfigProvider } from 'antd';
 import { BrowserRouter, Route, Routes } from 'react-router-dom';
 
 import { transport } from '@ui/config/transport';
-import { ModalContextProvider } from '@ui/features/common/modal/modal-context';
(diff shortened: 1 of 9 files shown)


krancour:
This is failing for me in all browsers.

Screenshot 2024-01-05 at 11 11 08 AM

This is while logged in as admin.

The same operation worked via CLI, so I think the back end is ok.

krancour (approved):
🔥

PR #1376, opened 2024-01-06
chore(controller): test order of precedence in credentials db

moensch:
Adding coverage to validate order of precedence in credentials.Get()


krancour:
Thank you so much @moensch! Looks good at a glance, but I'll take a closer look on Monday!

internal/credentials/credentials_test.go
```diff
@@ -27,6 +28,244 @@ func TestNewKubernetesDatabase(t *testing.T) {
...
+			secrets:  []client.Object{secretInGlobalExact, secretInGlobalPrefix},
+			expected: secretInGlobalExact,
+			found:    true,
+		},
+		{
+			name:     "argocd exact before prefix",
+			secrets:  []client.Object{secretInArgoCDNamespaceExact, secretInArgoCDNamespacePrefix},
+			expected: secretInArgoCDNamespaceExact,
+			found:    true,
+		},
```
krancour, line 228: @moensch are we missing a test case for "namespace before global"?

krancour: I added the missing test case.

krancour (approved):

krancour:
Thank you @moensch!

PR #1382, opened 2024-01-09
feat: add ability to entirely opt-out of argo cd and rollouts integrations

### LINKED ISSUE #1379: rollouts scheme may need to be registered for api server
AnalysisTemplates are typically defined in the Kargo cplane, so the API server needs to be able to apply those.

krancour:
Fixes #1379

Highlights:

* Allows Argo CD integration to be explicitly disabled for the controller. In which case:
    * Application reconciler does not start
    * Argo CD-based promotion mechanisms will error
    * Argo CD Application state cannot be factored into Stage health 
* Allows Argo Rollouts integration to be explicitly disabled for the controller. In which case:
    * AnalysisRun reconciler does not start
    * Verifications will fail with an error since they're defined using AnalysisTemplates and executed as AnalysisRuns
    * VerificationInfo has been expanded with new fields for better surfacing such errors
* Allows Argo Rollouts integration to be explicitly disable for the API server, in which case:
    * API server cannot apply manifests for AnalysisTemplate resources

WIP because I am still addressing some of @jessesuen feedback.


jessesuen:
Instead of (or in addition to) the `ROLLOUTS_INTEGRATION_ENABLED` variable to indicate if integration should explicitly be enabled, could we make this inferred by ability to list analysisruns/analysistemplates CRs (even if none are found) during startup?

The reason I think this is desired is because coordinating `ROLLOUTS_INTEGRATION_ENABLED` across many shards will be hard to maintain and it would be better to infer this with less required upfront configuration.

I'm okay with having the option of `ROLLOUTS_INTEGRATION_ENABLED` be explicit (to allow things to break when things aren't expected), but in the case where it is omitted, I feel we can rely on inference for a better UX.

jessesuen:
This is an example of how we avoid a knob for Istio in Argo Rollouts by trying to detect if an Istio VirtualService is even present. If not, we don't attempt to start the informer.

https://github.com/argoproj/argo-rollouts/blob/master/utils/istio/istio.go#L16-L22

```golang
func DoesIstioExist(dynamicClient dynamic.Interface, namespace string) bool {
	_, err := dynamicClient.Resource(GetIstioVirtualServiceGVR()).Namespace(namespace).List(context.TODO(), metav1.ListOptions{Limit: 1})
	if err != nil {
		return false
	}
	return true
}
```

api/v1alpha1/stage_types.go
```diff
@@ -673,7 +711,17 @@ type AnalysisRunArgument struct {
...
 type VerificationInfo struct {
-	AnalysisRun AnalysisRunReference `json:"analysisRun"`
+	// Phase describes the current phase of the Verification process. Generally,
+	// this will be a reflection of the underlying AnalysisRun's phase, however,
+	// there are exceptions to this, such as in the case where an AnalysisRun
+	// cannot be launched successfully.
+	Phase VerificationPhase `json:"phase,omitempty"`
+	// Message may contain additional information about why the verification
+	// process is in its current phase.
+	Message string `json:"message,omitempty"`
```
krancour, line 721: These are the main changes to the resources types.

It's _always_ been possible that something could go wrong with a verification other than an AnalysisRun failing or erroring, so really, this is the way things should have been done from the start. Added Phase and Message fields for reporting such problems.

The specific problem that is more _likely_ to occur, starting with this PR, is that verification is requested, but the controller doesn't have Rollouts integration enabled. This means an AnalysisRun cannot even be launched.

charts/kargo/templates/api/cluster-role.yaml
```diff
@@ -76,4 +76,12 @@ rules:
...
       - patch
       - update
+{{- if .Values.api.rollouts.integrationEnabled }}
+  - apiGroups:
+      - argoproj.io
+    resources:
+      - analysistemplates
+    verbs:
+      - "*"
+{{- end }}
```
krancour, line 86: This fixes #1379

cmd/controlplane/api.go
```diff
@@ -156,6 +158,11 @@ func newSchemeForAPI() (*runtime.Scheme, error) {
 	if err := kubescheme.AddToScheme(scheme); err != nil {
 		return nil, errors.Wrap(err, "add Kubernetes api to scheme")
 	}
+	if types.MustParseBool(os.GetEnv("ROLLOUTS_INTEGRATION_ENABLED", "true")) {
```
krancour, line 161: This is a spot where we might try to make an inference from the presence or absence of Rollouts CRDs.

cmd/controlplane/api.go
```diff
@@ -156,6 +158,11 @@ func newSchemeForAPI() (*runtime.Scheme, error) {
...
+		if err := rollouts.AddToScheme(scheme); err != nil {
+			return nil, errors.Wrap(err, "add argo rollouts api to scheme")
+		}
```
krancour, line 164: Fixes #1179

cmd/controlplane/controller.go
```diff
@@ -106,7 +106,7 @@ func newControllerCommand() *cobra.Command {
 			}
 
 			var argocdMgr manager.Manager
-			{
+			if types.MustParseBool(os.GetEnv("ARGOCD_INTEGRATION_ENABLED", "true")) {
```
krancour, line 109: This is a spot where we might try to make an inference from the presence or absence of Rollouts CRDs.

cmd/controlplane/controller.go
```diff
@@ -169,7 +169,7 @@ func newControllerCommand() *cobra.Command {
 			}
 
 			var rolloutsMgr manager.Manager
-			{
+			if types.MustParseBool(os.GetEnv("ROLLOUTS_INTEGRATION_ENABLED", "true")) {
```
krancour, line 172: This is a spot where we might try to make an inference from the presence or absence of Rollouts CRDs.

internal/controller/promotion/argocd.go
```diff
@@ -76,6 +79,14 @@ func (a *argoCDMechanism) Promote(
 		return promo.Status.WithPhase(kargoapi.PromotionPhaseSucceeded), newFreight, nil
 	}
 
+	if a.argocdClient == nil {
+		return promo.Status.WithPhase(kargoapi.PromotionPhaseFailed), newFreight,
+			errors.New(
+				"Argo CD integration is disabled on this controller; cannot perform " +
+					"promotion",
+			)
+	}
```
krancour, line 88: It isn't just the Application reconciler that we cannot start when Argo CD is disabled. We also cannot carry out Argo CD-based promotion mechanisms.

internal/controller/stages/health.go
```diff
@@ -24,6 +24,15 @@ func (r *reconciler) checkHealth(
...
 	}
 
+	if r.argocdClient == nil && len(argoCDAppUpdates) > 0 {
+		h.Status = kargoapi.HealthStateUnknown
+		h.Issues = []string{
+			"Argo CD integration is disabled on this controller; cannot assess" +
+				" the health or sync status of Argo CD Applications",
+		}
+		return &h
+	}
```
krancour, line 34: Cannot factor Application state into Stage health if Argo CD integration is disabled.

internal/controller/stages/stages.go
```diff
@@ -70,12 +70,12 @@ type reconciler struct {
 	startVerificationFn func(
 		context.Context,
 		*kargoapi.Stage,
-	) (*kargoapi.VerificationInfo, error)
+	) *kargoapi.VerificationInfo
```
krancour, line 73: Function signature changed...

If something goes wrong, we report it through the VerificationInfo's Phase and Message fields. What we don't want is the error to present as a _Stage_ error.

internal/controller/stages/verification.go
```diff
@@ -20,7 +20,15 @@ import (
 func (r *reconciler) startVerification(
 	ctx context.Context,
 	stage *kargoapi.Stage,
-) (*kargoapi.VerificationInfo, error) {
+) *kargoapi.VerificationInfo {
```
krancour, line 23: Once again, if something goes wrong, we now report it through the VerificationInfo's Phase and Message fields. What we don't want a verification problem to look like a _Stage_ problem.

internal/credentials/credentials.go
```diff
@@ -71,9 +71,9 @@ type Database interface {
 // utilizes a Kubernetes controller runtime client to retrieve credentials
 // stored in Kubernetes Secrets.
 type kubernetesDatabase struct {
-	kargoClient client.Client
-	argoClient  client.Client // nil if credential borrowing is not enabled
-	cfg         KubernetesDatabaseConfig
+	kargoClient  client.Client
+	argocdClient client.Client // nil if credential borrowing is not enabled
```
krancour, line 75: You see changes like this throughout. "argoClient" was too ambiguous, because Rollouts is also and Argo project. `argocdClient` is more specific.

api/v1alpha1/stage_types.go
```diff
@@ -485,7 +523,7 @@ type FreightReference struct {
 	Charts []Chart `json:"charts,omitempty"`
 	// VerificationInfo is information about any verification process that was
 	// associated with this Freight for this Stage.
-	VerificationInfo *VerificationInfo `json:"verificationResult,omitempty"`
+	VerificationInfo *VerificationInfo `json:"verificationInfo,omitempty"`
```
krancour, line 526: This was an oops to begin with.

krancour:
> Instead of (or in addition to) the ROLLOUTS_INTEGRATION_ENABLED variable to indicate if integration should explicitly be enabled, could we make this inferred by ability to list analysisruns/analysistemplates CRs (even if none are found) during startup?
>
> The reason I think this is desired is because coordinating ROLLOUTS_INTEGRATION_ENABLED across many shards will be hard to maintain and it would be better to infer this with less required upfront configuration.
>
> I'm okay with having the option of ROLLOUTS_INTEGRATION_ENABLED be explicit (to allow things to break when things aren't expected), but in the case where it is omitted, I feel we can rely on inference for a better UX.

@jessesuen, I addressed this in new commit 12bfb2883cfc8b95a8a3b3bc5cf77cd32853cbbd

I think there is a benefit to explicitly disabling Argo CD and/or Rollouts integrations if they are undesired because the controller is granted slightly fewer permissions that way.

But your point about configuration for many shards being difficult is a good one. And we've also seen instances of our own colleagues being surprised at the controller going into a crash loop when they forgot to install one of the dependencies.

So where I landed was:

* If you explicitly disable an integration, it is truly disabled without further condition. If you, for instance, had Argo CD or Argo Rollouts on a cluster and really wanted _not_ to enable those integrations, this is what you'd have to do.

* If you _don't_ explicitly opt-out, there's a sanity check in case you rolled with the default by accident. So if you start up with Rollouts integration (for example) enabled, but Rollouts CRDs not installed, then you'll see a warning, but the controller will function without error, just as if that feature had been explicitly disabled.

Does all that seem ok?

Willing to tweak this further if you think I've missed the mark.

jessesuen:
> but Rollouts CRDs not installed, then you'll see a warning, but the controller will function without error, just as if that feature had been explicitly disabled.

Yep! That is effectively the same as what I was asking for.

jessesuen (approved):

PR #1388, opened 2024-01-11
feat!: fully support project CRD

### LINKED ISSUE #558: Kargo Project onboarding
In Kargo, a project is really just a Namespace under the covers. But along with a Project, we might need to create other standard Kargo things inside that namespace, or do other conveniences. For example:

* Create some common, pre-defined Kargo-specific RBAC, such as an kargo "admin", "viewer", Role/RoleBinding/ServiceAccount.
* Give the `kargo:kargo-api` ServiceAccount permissions necessary access to the namespace, such as impersonating a ServiceAccount in the namespace (for the purposes of SSO)
* Maybe some default PromotionPolicies?
* Others?

This issue is about creating a convenience command or API to do wrap up all the things we want to do in a Kargo Project into a single command/API.

The motivation is to improve the user experience by reducing the amount of repetitious steps during project onboarding

### LINKED ISSUE #667: Project CRD
The project list view is a bit sad:

image

I wonder if we could fill out the project information with things like:
* number of stages
* aggregate health
* whether there is a promotion is in progress

Unfortunately, we need this information to be populated/cached/stored somewhere for the UI to be efficient.

Options I see:
* dedicated cache (like redis)
* annotation in project namespace (requires patch privileges to namespaces)
* ProjectMetadata CRD (which would be a child of a namespace)
* Project CRD (which would be an owner of a namespace)

krancour:
Fixes #558 
Fixes #667 

## Summary

This PR relieves the API server of responsibility (and permissions) for creating specially labeled namespaces that we count as "projects."

Project creation via API server now entails creation of a Project resource. The controller reconciles new Project resources by creating the corresponding namespace. Project resources are set as the owner of the associated namespace, so Project deletion takes the associated namespace with it. The reverse is also true. Namespaces have a finalizer that, upon deletion, is only cleared by the controller deleting the associated Project resource.

Follow-up PRs will add to this to create things like ServiceAccounts, Roles, and RoleBindings in the project namespaces.

## Breakdown

About 2/3 of this is generated code, so please don't be intimidated by the size.

Narrowly scoped commits cover each of the following:

* CRD changes: Add Project status field with phase and message (for communicating reconciliation errors)
* Corresponding protobuf changes: Mirror the CRD changes
* Run codegen
* Corresponding type conversion changes
* Update API server endpoints to operate on Project resources instead of operating directly on namespaces
* Refactor "apply" endpoints on API server: Necessary so the API server is smart enough to pick Projects out of manifests and apply them first, just as would normally happen with a namespace, since a Project _is_ a proxy for a namespace
* Corresponding CLI changes: Align with the protobuf changes
* New webhook: Validate new Project resources by checking the a namespace with the same name doesn't already exist
* Controller changes: Reconcile new Project CRs and deleted Namespace CRs
* Garbage collector changes: Trivial change -- adapts to a renamed constant in the api/v1alpga1 package
* Corresponding chart changes: Set up new webhook + changes to permissions for API server and controller
* UI changes: Trivial -- adapt to Project now being a root resource type
* Run `go mod tidy`: Refactor of API server "apply" endpoints removed a dependency
* Corresponding devx changes: Enable debug level logging for management controller when running with Tilt

## What doesn't change

Anywhere in the code where we validate the existence of a Project, we still do so by validating the existence of the corresponding namespace, since that's usually what we actually care about.

This PR does not address #1389. That will be addressed in a follow-up.


api/v1alpha1/labels.go
```diff
@@ -2,10 +2,12 @@ package v1alpha1
 
 const (
 	AliasLabelKey   = "kargo.akuity.io/alias"
-	LabelProjectKey = "kargo.akuity.io/project"
 	FreightLabelKey = "kargo.akuity.io/freight"
+	ProjectLabelKey = "kargo.akuity.io/project"
```
krancour, line 6: Changed this symbol just for consistency with the others.

api/service/v1alpha1/service.proto
```diff
@@ -385,25 +385,20 @@ message DeletePromotionPolicyResponse {
   /* explicitly empty */
 }
 
-message Project {
-  string name = 1;
-  google.protobuf.Timestamp create_time = 2;
-}
```
krancour, line 391: This moved to `types.proto` where we have the protobuf definitions of anything that's also a CRD.

charts/kargo/templates/webhooks/webhooks.yaml
```diff
@@ -65,6 +65,21 @@ webhooks:
...
+    service:
+      namespace: {{ .Release.Namespace }}
+      name: kargo-webhooks-server
+      path: /validate-kargo-akuity-io-v1alpha1-project
+  rules:
+  - scope: Cluster
+    apiGroups: ["kargo.akuity.io"]
+    apiVersions: ["v1alpha1"]
+    resources: ["projects"]
+    operations: ["CREATE"]
```
krancour, line 81: For the moment, only need to validate creates.

cmd/controlplane/webhooks.go
```diff
@@ -81,17 +82,20 @@ func newWebhooksServerCommand() *cobra.Command {
...
 				return errors.Wrap(err, "setup Promotion webhook")
 			}
 			if err = promotionpolicy.SetupWebhookWithManager(mgr); err != nil {
 				return errors.Wrap(err, "setup PromotionPolicy webhook")
 			}
-			if err = freight.SetupWebhookWithManager(mgr); err != nil {
-				return errors.Wrap(err, "setup Freight webhook")
+			if err = stage.SetupWebhookWithManager(mgr); err != nil {
+				return errors.Wrap(err, "setup Stage webhook")
 			}
```
krancour, line 99: Besides adding a webhook, I reordered these to be alphabetical.

internal/kubeclient/manifest/manifest.go
```diff
@@ -1,73 +0,0 @@
...
-	deserializer := codecs.UniversalDeserializer()
-	factory := resmap.NewFactory(provider.NewDefaultDepProvider().GetResourceFactory())
-
-	return func(data []byte) (cluster, namespaced []*unstructured.Unstructured, err error) {
-		resMap, err := factory.NewResMapFromBytes(data)
-		if err != nil {
-			return nil, nil, errors.Wrap(err, "new resmap from data")
-		}
-		cluster = make([]*unstructured.Unstructured, 0, resMap.Size())
-		for _, r := range resMap.ClusterScoped() {
```
krancour, line 38: To make this recognize Project as a cluster-scoped resource required jumping through a lot of hoops that were more easily avoided by refactoring this whole parser into the new `splitYAML()` function that was adapted from GitOps Engine.

internal/webhook/project/webhook.go
```diff
@@ -0,0 +1,96 @@
...
+		namespace,
+	); err != nil {
+		if client.IgnoreNotFound(err) == nil {
+			// This is good. The namespace doesn't already exist.
+			return nil, nil
+		}
+		return nil, apierrors.NewInternalError(
+			errors.Wrapf(err, "error getting namespace %q", project.Name),
+		)
+	}
```
jessesuen, line 71: Since you also have logic in the Project controller to fail project initialization if the namespace already exists, does the webhook serve to provide immediate feedback instead of delayed feedback of a pre-existing namespace?

jessesuen (approved):

internal/webhook/project/webhook.go
krancour: Yes. That's it exactly. Fail synchronously if we can spot the problem sooner.

internal/api/create_resource_v1alpha1.go
```diff
@@ -15,26 +15,15 @@ func (s *server) CreateResource(
...
-				},
-			})
-			continue
-		}
-		res = append(res, s.createResource(ctx, obj))
+	resources := append(projects, otherResources...)
+	res := make([]*svcv1alpha1.CreateResourceResult, 0, len(resources))
+	for _, r := range resources {
+		resource := r // Avoid implicit memory aliasing
+		res = append(res, s.createResource(ctx, &resource))
```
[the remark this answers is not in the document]
krancour: It _is_ possible, but so far I've seen it only once out of probably 50+ times I have done this.

I'd been contemplating how to make that more foolproof. I could probably add logic here to wait for project readiness before moving on to non-project resources.

I'm going to open a follow-up issue.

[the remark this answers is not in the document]
krancour: Right here so we don't lose track:

https://github.com/akuity/kargo/issues/1415

PR #1406, opened 2024-01-19
feat: use finalizer to ensure proper cleanup after stage deletion

### LINKED ISSUE #1191: undo freight verifications on stage deletion
Follow up to #1189.

#1189 adds manual approvals and had the foresight to also undo those approvals when a Stage is deleted.

That prevents implicit approval of a piece of Freight for a _new_ Stage that may just happen to reuse the name of an old Stage that has since been deleted, but might have nothing _else_ in common with it.

For the same reason, we should undo verifications on Stage deletion.

krancour:
Fixes #1191


rbreeze (approved):
LGTM!

PR #1428, opened 2024-01-25
refactor(controller): improve application and analysis run event handling

### LINKED ISSUE #1318: see what new tricks controller runtime can do
Follow up on [this comment](https://github.com/akuity/kargo/pull/1317/files#r1434557927).

I believe it _might_ be possible now to create controllers that reconcile using one client and still watch related resources using _a different_ client, which will be useful for sharded topologies wherein one client is used for communication with the Kargo control plane and another is used for communication with the shard. 🤞

krancour:
Fixes #1318

What I had hoped was possible was. After a recent controller runtime upgrade, we are able to use two different clients for watching resources and for adding related resources to the work queue. 😄 

This means we can remove the dedicated reconcilers for Argo CD Applications and Argo Rollouts AnalysisRuns in favor of new event handlers on the Stage controller. When an Application changes health or sync state, or when an AnalysisRun changes phase, the corresponding Stages are enqueued for reconciliation.

This is much more straightforward than monkeying around with annotations to induce reconciliation of Stages, which may live on a different cluster than the Applications or AnalysisRuns.

__Additional note:__ All the logic is transplanted from the old predicates to the new event handlers without any significant modification. That means even existing TODOs remain TODOs.


jessesuen (approved):
This is great!

PR #1431, opened 2024-01-26
refactor: create project namespaces synchronously

### LINKED ISSUE #1415: make kargo apply slightly more foolproof wrt async project reconcilation
When using `kargo apply -f ...`, referencing a file that contains both Projects _and_ resources that belong _to_ the Project, there is a small possibility of problems due to the fact that Projects are asynchronously reconciled as namespaces. If that doesn't happen fast enough, the `apply` command may experience failures.

We apply logic already prioritizes Project apply over all other resources, so we can probably account for this by waiting for Project readiness before moving on to other resources.

https://github.com/akuity/kargo/pull/1388#discussion_r1464126640

krancour:
Fixes #1415 

#1388 began creating the namespace associated with a Project asynchronously as part of Project reconciliation.

Although rarely encountered, it became possible for creation of resources _in_ the Project's namespace to fail because the Project was not yet reconciled for the first time.

This change absolves the Project reconciler of responsibility for namespace creation and makes that task the responsibility of the Project validation webhook. This webhook used to at least check for availability of the namespace, but now it creates it. Namespace creation is synchronous now, which eliminates the potential for the failure described.

### DIFF
--- a/internal/webhook/project/webhook.go
+++ b/internal/webhook/project/webhook.go
@@ -3,17 +3,22 @@ package project
 import (
 	"context"
 
+	"github.com/containerd/log"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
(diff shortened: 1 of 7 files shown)


internal/webhook/project/webhook.go
```diff
@@ -3,17 +3,22 @@ package project
 import (
 	"context"
 
+	"github.com/containerd/log"
```
devholic, line 6:
```suggestion
	log "github.com/sirupsen/logrus"
```

devholic (commented):
Seems good to me with a minor import fix

internal/webhook/project/webhook.go
krancour: 🤦 Great catch @devholic.

devholic (approved):

PR #1433, opened 2024-01-26
feat(cli): Add kargo dashboard command

### LINKED ISSUE #549: command to open ui in default browser
Low priority, but it would be very convenient to have a CLI command that opens the UI in the system's default browser if you are already logged in to a server.

webstradev:
This should close #549 .

Just to note this will not work in development when using tilt (it won't break but it will open the API Url which in the Tilt method is not currently the URL for the dashboard).

### DIFF
--- a/internal/cli/cmd/dashboard/dashboard.go
+++ b/internal/cli/cmd/dashboard/dashboard.go
@@ -1,0 +1,29 @@
+package dashboard
+
+import (
+	"github.com/bacongobbler/browser"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/akuity/kargo/internal/cli/config"
+)
+
+func NewCommand(cfg config.CLIConfig) *cobra.Command {
+	return &cobra.Command{
+		Use:     "dashboard",
+		Short:   "Open the Kargo Dashboard in your default browser.",
+		Example: "kargo logout",
+		RunE: func(*cobra.Command, []string) error {
+			if cfg.APIAddress == "" {
+				return errors.New(
+					"seems like you are not logged in; please use `kargo login` to authenticate",
+				)
+			}
+
+			return errors.Wrap(
+				browser.Open(cfg.APIAddress),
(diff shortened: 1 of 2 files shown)


internal/cli/cmd/dashboard/dashboard.go
```diff
@@ -0,0 +1,29 @@
...
+
+func NewCommand(cfg config.CLIConfig) *cobra.Command {
+	return &cobra.Command{
+		Use:     "dashboard",
+		Short:   "Open the Kargo Dashboard in your default browser.",
+		Example: "kargo logout",
+		RunE: func(*cobra.Command, []string) error {
+			if cfg.APIAddress == "" {
+				return errors.New(
+					"it seems like you are not logged in - Please use kargo login first",
```
krancour, line 19: There's a similar error message one other place in the code.

I think when we re-organize the CLI (which I alluded was coming), we'll make sure to DRY this up, but for now, I just want to make this message match the other.

```suggestion
					"seems like you are not logged in; please use `kargo login` to authenticate",
```

internal/cli/cmd/dashboard/dashboard.go
```diff
@@ -0,0 +1,29 @@
...
+				)
+			}
+
+			return errors.Wrap(
+				browser.Open(cfg.APIAddress),
+				"error opening dashboard in defaultbrowser",
```
krancour, line 25:
```suggestion
				"error opening dashboard in default browser",
```

krancour:
@webstradev I made two minor suggestions about wording, but all the code LGTM.

webstradev:
> @webstradev I made two minor suggestions about wording, but all the code LGTM. Thank you!

Thanks for that, resolved both comments ( rebased off main as well to keep the history clean)

krancour:
@webstradev thanks!

For future reference, we squash merge, so don't worry too much about commit history.

krancour (approved):
🔥

PR #1434, opened 2024-01-27
feat(chart): add possibility to change pathPrefix for Ingress resource

Brightside56:
Ingress controllers like AWS LoadBalancer controller don't support `/` path as wildcard path when using `ImplementationSpecific` pathType, therefore possibility to set another value is needed. I would say `pathType: Prefix` would be even more safe default, but I'm not sure if it's acceptable


krancour (approved):
LGTM @Brightside56. Thank you! I know you've also given some guidance to those having trouble with ALBs on Discord, so thank you for that as well!

PR #1446, opened 2024-01-31
feat: allow warehouse git subs to select commits by tag

### LINKED ISSUE #816: optionally constrain git repo subscriptions by tags
Subscriptions to Git repos are currently very different from subscriptions to image or chart repos.

With images, it's possible, but unlikely that every commit to your source code repo results in a new release of the image. For most cases, a recent _batch_ of changes will be deemed release-worthy and the corresponding image(s) will be built and published to the image repo.

Similar is true for charts (consumed from chart repos; not charts consumed directly from Git). It's possible, but unlikely that every commit to a chart's own source code repo results in a new release of the chart. For most cases, a recent _batch_ of changes will be deemed release-worthy and the corresponding chart will be built and published to the chart repo.

But if you subscribe a stage to the `main` branch of a Git repo containing manifests, _every_ commit to that branch enters your pipeline. Speaking personally, I don't love that. As with images and charts, I'd like to decide that a recent _batch_ of changes is release-worthy and then release it into my pipeline.

I think this is easy-ish to do using semantics similar to those used by the image subscriptions (which are borrowed in their entirety from Argo CD Image Updater). I'm not advocating that manifests need to be tagged with semantic versions, but suppose they were. It's easy to see that you _could_ then subscribe to manifests with a version constraint like `^1.24.0`. Again... using all the same semantics as Image Updater, we could support a broad range of sensible schemes ranging from "just give me the newest" to "give me the last in a lexical ordering of all tags matching this pattern."

To be clear... this would all be optional, but the end result would be a way to prevent _every. little. change._ from entering your pipelines if that isn't what you want.

cc @jessesuen

krancour:
Fixes #816 

This is much smaller than it looks once you ignore codegen.

This works similarly to image selection by tag:

* Select latest commit from a given branch (existing behavior)
* Select lexically last tag (useful if tags contain date/time stamps)
* Select newest tag
* Select latest tag that looks like a semantic version -- semver constaints also supported
* Selection can be further constrained by:
    * Regex of allowed tags
    * List of forbidden tags

WIP because I am still working out the kinks.


christianh814 (approved):

api/v1alpha1/warehouse_types.go
```diff
@@ -59,13 +69,48 @@ type GitSubscription struct {
...
+	//+kubebuilder:validation:Optional
+	AllowTags string `json:"allowTags,omitempty"`
+	// IgnoreTags is a list of tags that must be ignored when determining the
+	// newest commit of interest. No regular expressions or glob patterns are
+	// supported yet. The value in this field only has any effect when the
+	// CommitSelectionStrategy is Lexical, NewestTag, or SemVer. This field is
+	// optional.
+	//
+	//+kubebuilder:validation:Optional
+	IgnoreTags []string `json:"ignoreTags,omitempty"`
```
jessesuen, line 113: I question the need for ignoreTags, when:
1. allowTags is a regular expression. And exclusion would be possible with a "negative lookahead":
https://stackoverflow.com/a/2078953
2. The fact that ignoreTags is a list, whereas allowTags is a regexp, making them not symetrical.

jessesuen: Discussed offline. Image tag selection already has this feature and so it's worthwhile to offer the same with git.

jessesuen (approved):
Great feature!

krancour:
@jessesuen thank you so much for reviewing!

PR #1452, opened 2024-02-03
chore(controller): change image selection func and variable names for clarity

krancour:
Over the course of other recent PRs, variables and functions involved in the Warehouse reconciler's selection of commits and charts have been changed to avoid use of the term "latest," which can be misleading, since user-defined selection criteria often results in selection of a commit or chart that is _not_ the most recent.

This PR applies similar name changes to variables and functions involved in the Warehouse reconciler's selection of container images.

There are no functional changes in this PR. This is purely for readability and clarity of code.

### DIFF
--- a/internal/controller/warehouses/images.go
+++ b/internal/controller/warehouses/images.go
@@ -14,7 +14,7 @@ import (
 	"github.com/akuity/kargo/internal/logging"
 )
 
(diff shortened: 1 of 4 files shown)


gdsoumya (approved):
LGTM

PR #1460, opened 2024-02-05
feat(chart): Support for adding EnvVars to the dex pod

sennerholm:
For example to add a CliettSecret from a secret already created in the cluster 

Sample:
```
api:
    host: <fqdn>
    oidc:
      enabled: true
      dex:
        enabled: true
        envVars:
          - name: CLIENT_SECRET
            valueFrom:
              secretKeyRef:
                name: github-dex
                key: dex.github.clientSecret

        connectors:
          - type: github
            id: github
            name: GitHub
            config:
              clientID: <theClientIdFromGithub>
              clientSecret: "$CLIENT_SECRET"
              redirectURI: https://<fqdn>/dex/callback
              orgs:
              - name: <YourGithubOrg>
```

Inspired by the config to the upstream dexidp chart at: https://artifacthub.io/packages/helm/dex/dex


sennerholm:
If you have any testcases for the helm chart, please point me in the right direction and I will try to update those.

eobw (approved):

krancour:
Thanks @sennerholm!

This solves something I knew was a problem and hadn't really gotten around to dealing with yet.

For the exact reason that connector configs can contain sensitive information, the chart puts all connector config in a secret... but that doesn't really achieve much. Your connector config _still_ appears in plain text in your Kargo installation's `values.yaml` file -- sensitive info an all -- which makes it unsuitable for storing in a git repo.

Your approach opens the possibility of secrets being managed independently of the chart through some other solution, like Sealed Secrets, for instance, and then merely referenced from the connector config.

Love it!

If you don't mind, I may amend this PR with some changes to the wording in the docs, but I'm leaving your approach exactly as is.

sennerholm:
Hi

Improve it the way you want. The important part for me is to get the
functionality in.

Sincerely

Den fre 16 feb. 2024 20:43Kent Rancourt ***@***.***> skrev:

krancour (approved):

krancour:
Thanks again @sennerholm! This really is a great improvement.

PR #1462, opened 2024-02-06
chore!: clean up things having to do with image selection

krancour:
Something has been bugging me ever since we switched off of Image Updater in #1058...

Both in the Warehouse CRD and way, deep down, in the `image` package that replaced Image Updater, we speak of "tag selection strategies." This is not accurate. Users define criteria for selecting an _image_. The criteria are often (but not always) based on tags, but what is being selected is an image.

This PR corrects these inaccuracies. Examples:

* TagSelectionStrategy becomes ImageSelectionStrategy
* Tag struct renamed to Image
* Other interface, function, and struct names renamed for clarity.

There is one breaking change.

Before:

```yaml
kind: Warehouse
metadata:
  name: kargo-demo
  namespace: kargo-demo
spec:
  subscriptions:
  - image:
      repoURL: some/image
      tagSelectionStrategy: NewestBuild # For example
```

After:

```yaml
kind: Warehouse
metadata:
  name: kargo-demo
  namespace: kargo-demo
spec:
  subscriptions:
  - image:
      repoURL: some/image
      imageSelectionStrategy: NewestBuild # For example
```

Apart from this, there are no changes that aren't purely for code clarity.

I have vetted this end-to-end and, more importantly, the integration tests that test all strategies against Docker Hub and GHCR are still passing. (These do not run during CI because they are too heavy and can lead to rate limiting if executed without DockerHub credentials.)


internal/image/digest_selector.go
```diff
@@ -0,0 +1,82 @@
+package image
```
krancour, line 1: This file used to be `digest_tag_selector.go`. With the rename of many variables and functions, GitHub seems not to recognize this as a renamed file and thinks it is new.

This applies to all the selectors.

internal/image/image.go
```diff
@@ -7,23 +7,23 @@ import (
...
-// Tag is a representation of a container image Tag.
-type Tag struct {
-	Name      string
+// Image is a representation of a container image.
+type Image struct {
+	Tag       string
 	Digest    digest.Digest
 	CreatedAt *time.Time
 	semVer    *semver.Version
 }
```
krancour, line 16: This change actually illustrates the motivation for the entire PR.

`Digest` is _not_ an attribute of a tag. It is an attribute of an _image_.

internal/image/selector.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"context"
+	"regexp"
+
+	"github.com/pkg/errors"
+)
+
+// SelectionStrategy represents a strategy for selecting a single image from a
+// container image repository.
+type SelectionStrategy string
```
krancour, line 12: Another good example of the motivations for this PR. This was formerly `TagSelectionStrategy`.

`image.TagSelectionStrategy` being used to select a "tag" (which is clearly an image) made little sense compared to:

`image.SelectionStrategy`, which properly uses package name for context, doesn't "stutter," and selects an `Image`.

jessesuen (approved):
LGTM. Always in favor of fixing inaccurate terminology before it's too late.

PR #1463, opened 2024-02-06
fix(controller): add new nil checks in stages reconciler initialization

krancour:
Fixes a nil pointer dereference that @gdsoumya discovered.

These some controller managers can be nil when Argo CD and/or Argo Rollouts integrations are disabled. In those cases, we don't need the corresponding watches anyway.


gdsoumya (approved):
LGTM

PR #1464, opened 2024-02-07
feat(controller): simplify and improve rollouts integration

### LINKED ISSUE #1442: Support ROLLOUTS_INSTANCE_ID when creating AnalysisRuns
# Proposed Feature

When creating AnalysisRuns, the kargo controller should have the option to label that AnalysisRun with the `controller-instance-id` label supported in Argo Rollouts. e.g.

```yaml
metadata:
  labels:
    argo-rollouts.argoproj.io/controller-instance-id: foo
```

# Motivation

There may be multiple argo rollouts controllers acting on a cluster for different purposes (e.g. shard). The instance-id option allows them to co-exist.

# Suggested Implementation

Introduce `ROLLOUTS_INSTANCE_ID` env var. When set, then the AnalysisRun `argo-rollouts.argoproj.io/controller-instance-id` label will be set with that value.

krancour:
Fixes #1442

Adds the ability to configure a Kargo controller to add the `argo-rollouts.argoproj.io/controller-instance-id` label to any AnalysisRuns it creates and stores in the Kargo control plane cluster.

Tangentially, also allows user installing Kargo to specify a namespace to put AnalysisRuns instead of coupling that to whether or not the controller is running in a shard.


gdsoumya (approved):
LGTM

Issue #1270, opened 2023-12-12
Freight filtering rules


jessesuen:
# Proposed Feature

Subscribing to a git repository can be, in many cases, very noisy and will produce inconsequential freight. Examples include:
* mono repos where there are other directories of unrelated apps
* changes to README.md or unrelated files

Warehouses should allow users to define rules so that only when certain paths are changed, should new freight be generated. e.g. suppose for a kustomize layout, maybe only changes to these directories should produce freight:

```
apps/
  guestbook/
    base/
    overlays/
```

# Motivation

This will improve usability and filter out noise in the Freightline

# Suggested Implementation

Warehouse API spec should introduce a new stanza on filtering rules. Before generating new Freight, it will see if the filtering rules apply and produce Freight accordingly.

NOTE: if we support git webhooks, the git webhook will have information about files that changed. This would avoid a git checkout. But I consider this an optimization and we shouldn't rely on this because we might miss a webhook.

maksimstankevic:
@krancour @jessesuen I was looking for a new, bit bigger issue, can I pick this one? I have limited free time off work and not sure how long this will take me but if it's fine - really interested to work on this.

krancour:
@maksimstankevic have at it. We're happy for your assistance!

If I may suggest, simply on account of this being a somewhat complex change, if @jessesuen or myself can review the additions to the Warehouse CRD before you move ahead with corresponding changes to the controller, I think it would help to make sure we're all on the same page early in the process.

maksimstankevic:
@krancour @jessesuen could you please check and comment, https://github.com/akuity/kargo/compare/main...maksimstankevic:kargo:feature/1270. I thought we could use similar approach as with AllowTags/IgnoreTags in ImageSubscription but use list of regexps in both GitSubscription's ScanPaths and IgnorePaths for more flexibility.

krancour:
@maksimstankevic I am so sorry for the delayed review. I think that is a good starting point.

maksimstankevic:
@krancour, hey, I submitted https://github.com/akuity/kargo/pull/1477 for this one, please, take a look when you have a sec

krancour:
This was closed by #1477

PR #1477, opened 2024-02-14
feat(api/controller): support filtering rules Git subscriptions

maksimstankevic:
https://github.com/akuity/kargo/issues/1270


krancour:
Many apologies for the delay getting to this -- 👀 -- top priority for tomorrow.

internal/controller/warehouses/git.go
```diff
@@ -139,14 +148,46 @@ func (r *reconciler) selectTagAndCommitID(
 ) (string, string, error) {
 	if sub.CommitSelectionStrategy == kargoapi.CommitSelectionStrategyNewestFromBranch {
 		// In this case, there is nothing to do except return the commit ID at the
-		// head of the branch.
-		commit, err := r.getLastCommitIDFn(repo)
+		// head of the branch unless there are scanPaths/ignorePaths configured to
+		// handle.
+		commit, err1 := r.getLastCommitIDFn(repo)
```
hiddeco, line 153: Think the error should be dealt with before further looking at the scan or ignore paths. Reason for this is that otherwise the `commit` you are making use of may actually not be available.

In addition, all `err<num>` can be changed to `err`.

internal/controller/warehouses/git.go
```diff
@@ -94,13 +94,22 @@ func (r *reconciler) selectCommitMeta(
...
+		cloneShallowness = 1
+	}
 	repo, err := git.Clone(
 		sub.RepoURL,
 		*creds,
 		&git.CloneOptions{
 			Branch:                sub.Branch,
 			SingleBranch:          true,
-			Shallow:               true,
+			Shallow:               cloneShallowness,
```
hiddeco, line 112: (If my `Depth` comment is taken into account)

```suggestion
			Depth:               &cloneDepth,
```

internal/controller/warehouses/git.go
```diff
@@ -231,6 +272,72 @@ func ignores(tagName string, ignore []string) bool {
...
+		return true, nil
+	}
+	return false, nil
+}
+
+// matchesRegexList is a general purpose function iterating given slice of regexp
+// strings (regexList) to check if any of them match the stringToMatch string, if
+// a regexp fails to compile it returns false and error, if match is found it
+// returns true and if match is not found it returns false.
+func matchesRegexList(stringToMatch string, regexList []string) (bool, error) {
```
hiddeco, line 322: This is quite a costly function, as recompilation of regular expressions happens repeatedly for every `diffPath`, while `scanPaths` and `ignorePaths` are static.

Given this, I would restructure the code in such a way that the lists of regular expressions are compiled once before attempting to match them against diff paths.

hiddeco (commented):
Thanks for your contribution @maksimstankevic! 🍏 

Have made a couple of comments and suggestions, but as I am quite new to the project I would wait for the final suggestions from @kencochrane before making any adjustments.

internal/controller/git/git.go
```diff
@@ -113,10 +117,9 @@ type CloneOptions struct {
 	Branch string
 	// SingleBranch indicates whether the clone should be a single-branch clone.
 	SingleBranch bool
-	// Shallow indicates whether the clone should be with a depth of 1. This is
-	// useful for speeding up the cloning process when all we care about is the
-	// latest commit from a single branch.
-	Shallow bool
+	// Shallow indicates depth of cloning. This is useful for speeding up
+	// cloning process.
+	Shallow uint8
```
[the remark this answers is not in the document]
maksimstankevic: agree

internal/controller/warehouses/git.go
```diff
@@ -94,13 +94,22 @@ func (r *reconciler) selectCommitMeta(
...
 	}
+	// for cases when ScanPaths and/or IgnorePaths filters are used we want to have 2
+	// commits cloned to be able to get actual DiffPaths of the last commit to compare
+	// against the filters
+	var cloneShallowness uint8
+	if sub.ScanPaths != nil || sub.IgnorePaths != nil {
+		cloneShallowness = 2
+	} else {
+		cloneShallowness = 1
+	}
```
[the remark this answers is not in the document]
maksimstankevic: agree

internal/controller/warehouses/git.go
maksimstankevic: agree

internal/controller/warehouses/git.go
maksimstankevic: agree

maksimstankevic:
@hiddeco thanks for your review! Agree with all points, @krancour , @kencochrane, please, give me a go if you agree as well with proposed changes.

krancour:
@maksimstankevic I'm going to look at this myself today as well, but I agree with all of @hiddeco's proposed changes. (He's a new maintainer on our team, btw. 🎉 )

internal/controller/git/git.go
```diff
@@ -113,10 +117,9 @@ type CloneOptions struct {
 	Branch string
 	// SingleBranch indicates whether the clone should be a single-branch clone.
 	SingleBranch bool
-	// Shallow indicates whether the clone should be with a depth of 1. This is
-	// useful for speeding up the cloning process when all we care about is the
-	// latest commit from a single branch.
-	Shallow bool
+	// Shallow indicates depth of cloning. This is useful for speeding up
```
krancour, line 120:
```suggestion
	// Shallow indicates depth of cloning. This is useful for speeding up the
```

internal/controller/git/git.go
```diff
@@ -65,6 +66,9 @@ type Repo interface {
 	// GetDiffPaths returns a string slice indicating the paths, relative to the
 	// root of the repository, of any new or modified files.
 	GetDiffPaths() ([]string, error)
+	// GetDiffPathsForCommitID returns a string slice indicating the paths, relative to the
+	// root of the repository, of any new or modified files for a given commit ID.
```
krancour, line 70:
```suggestion
	// root of the repository, of any files that are new or modified since the given commit ID.
```

internal/controller/warehouses/git.go
```diff
@@ -94,13 +94,20 @@ func (r *reconciler) selectCommitMeta(
 	if sub.CommitSelectionStrategy == "" {
 		sub.CommitSelectionStrategy = kargoapi.CommitSelectionStrategyNewestFromBranch
 	}
+	// for cases when ScanPaths and/or IgnorePaths filters are used we want to have 2
+	// commits cloned to be able to get actual DiffPaths of the last commit to compare
+	// against the filters
+	var cloneDepth uint8 = 1
+	if sub.ScanPaths != nil || sub.IgnorePaths != nil {
```
krancour, line 101:
```suggestion
	if len(sub.ScanPaths) != 0 || len(sub.IgnorePaths) != 0 {
```

This will avoid unnecessarily increasing the clone depth if one of these is non-nil, but still empty.

internal/controller/git/git.go
```diff
@@ -336,6 +340,24 @@ func (r *repo) GetDiffPaths() ([]string, error) {
 	return paths, nil
 }
 
+func (r *repo) GetDiffPathsForCommitID(commitId string) ([]string, error) {
+	resBytes, err := libExec.Exec(r.buildCommand("show", "--name-status", "--format=format:", commitId))
```
krancour, line 344: I was understanding the purpose of this function to be finding all paths that have had any changes since the provided commit ID. Is that correct?

If so, the command `git show --name-status --format=format:<commit id here>` does not seem to accomplish that.

I am getting the correct results from `git diff --name-only <commit id here>..HEAD`

internal/controller/warehouses/git.go
```diff
@@ -139,14 +146,47 @@ func (r *reconciler) selectTagAndCommitID(
...
+				errors.Wrapf(err, "error determining commit ID at head of branch %q in git repo %q",
+					sub.Branch,
+					sub.RepoURL)
+		}
+		// below if statement handles scanPaths/ignorePaths filters if configured in a
+		// git subscription
+		if sub.ScanPaths != nil || sub.IgnorePaths != nil {
+
+			// getting actual diffPaths for last commit
+			diffs, err := r.getDiffPathsForCommitIDFn(repo, commit)
```
krancour, line 163: I understood the intent of this function call to be finding all paths that have changed since `commit`, but the value of `commit` here is that of the HEAD. Wouldn't a comparison of HEAD to HEAD always yield no diffs.

internal/controller/git/git.go
maksimstankevic: No, never-mind, this was getting diffs of a given commit compared to previous one, and now that I understand my approach was wrong, this does not matter, see my next comment.

internal/controller/warehouses/git.go
maksimstankevic: My approach was wrong, now I'm getting it - we need to obtain diffs between the commit that produced last Freight and the HEAD as you say, so Depth of clone loses its sense probably as the mentioned Freight-producing commit is at unknown Depth... Thanks a lot for your review @krancour ! I will reiterate.

maksimstankevic:
My approach was wrong, now I'm getting it - we need to obtain diffs between the commit that produced last Freight and the HEAD as you say, so Depth of clone loses its sense probably as the mentioned Freight-producing commit is at unknown Depth... Thanks a lot for your review @krancour ! I will reiterate.

krancour:
> Depth of clone loses its sense

@maksimstankevic right... that was my next bit of feedback was that depth required to figure this out correctly cannot be predicted, but want to make sure we were on the same page with _what_ diffs we need to find first... sounds like we now are.

Can't tell you how much we appreciate your effort on this!

api/v1alpha1/warehouse_types.go
```diff
@@ -216,6 +230,8 @@ type WarehouseStatus struct {
 	// ObservedGeneration represents the .metadata.generation that this Warehouse
 	// was reconciled against.
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// FreightReference refers to the last Freight produced by this Warehouse
+	FreightReference string `json:"freightReference,omitempty"`
```
krancour, line 234: @maksimstankevic there is a `FreightReference` type. I propose something like this:

```suggestion
	LastFreight *FreightReference `json:"lastFreight,omitempty"`
```

The benefits of this are twofold:

1. No need to make a separate call later to retrieve Freight.
2. Guards against the possibility that the last Freight shipped from the Warehouse has since been deleted.

maksimstankevic: @krancour, thanks, absolutely, I'm completely new to this k8s programming so still need to wrap my mind around a lot of ways :) I will refactor that once testing is successful. But I have a couple of points to raise, during my work on this I saw 2 instances of what I believe are bugs, can you give your take on those:

1)   I think it should not be possible to create a Warehouse with duplicated Git type subscriptions, verification should be added I suppose, below is an excerpt from a Warehouse running in my DEV env:
```
  uid: 46e4a742-0a97-48d5-8f9a-a719922820cb
spec:
  subscriptions:
  - git:
      branch: helm
      commitSelectionStrategy: NewestFromBranch
      repoURL: https://github.com/maksimstankevic/kargo-demp-876.git
  - git:
      commitSelectionStrategy: NewestFromBranch
      repoURL: https://github.com/maksimstankevic/kargo-demp-877.git
  - git:
      commitSelectionStrategy: NewestFromBranch
      repoURL: https://github.com/maksimstankevic/kargo-demp-877.git
status:
  freightReference: 70c665c38cf6f238e7fb0a6b7763ea902b29547d  
```
2) the interesting one: per my testing I see that in case of multiple Git supscriptions in a Warehouse a valid failure to get a new commit from one of Git repo subscriptions blocks normal progress of others, meaning new valid commits in other Git subscriptions will not be able to produce new Freight if one Git subscription fails to select a commit due to valid reasons (scanPaths/ignorePaths or Tag configurations): https://github.com/akuity/kargo/blob/c9aebcd37e20d40f75c31504297126fafc926968/internal/controller/warehouses/git.go#L58C1-L60C29
Don't you think that in such case we should just let the failing Git repo subscription to stay where it was is but if others have new valid commits - we should produce new Freight?

krancour: > I think it should not be possible to create a Warehouse with duplicated Git type subscriptions, verification should be added I suppose

Additional validations seldom hurt. I would suggest opening a new issue for it and not dealing with it here.

> per my testing I see that in case of multiple Git supscriptions in a Warehouse a valid failure to get a new commit from one of Git repo subscriptions blocks normal progress of others

This is by design. If you think of a piece of Freight as a box containing all the artifacts required for your application, including image, manifests, etc., a failure to fetch any one of those means you have an incomplete set.

> Don't you think that in such case we should just let the failing Git repo subscription to stay where it was

I guess this suggests that if I'm looking for versions of x and y and I find x, but don't find y, we should use the last known version of y... the problem is that if there were a version of y that fit your current constraints, _it would have been found_, meaning the last known version of y _does not_ fit your constraints. The main scenario where I can see this happening is when someone has _changed_ their constraints. They previously wanted ^1.0.0 of y and now want ^2.0.0. If nothing matching ^2.0.0 was found, it's incorrect to move forward with the old y that matched ^1.0.0.

maksimstankevic: yes, that makes sense, for validation of duplications I'll open an issue, that was my thought initially

maksimstankevic: @krancour I pushed the changes, please, review when you get a chance

krancour: @maksimstankevic at a glance, this is looking like it's almost there. I promise a more thorough review tomorrow (Tuesday).

Let me say again how much we appreciate this. This was an ambitious feature to bite off and it's coming along nicely!

internal/controller/warehouses/warehouses.go
```diff
@@ -151,6 +155,7 @@ func newReconciler(
 	r.getLatestFreightFromReposFn = r.getLatestFreightFromRepos
 	r.selectCommitsFn = r.selectCommits
 	r.getLastCommitIDFn = r.getLastCommitID
+	r.getDiffPathsSinceCommitIDFn = r.getDiffPathsSinceCommitID
```
krancour, line 158: Tests should probably assert that this is set by `newReconciler()`.

internal/controller/warehouses/warehouses.go
```diff
@@ -238,6 +243,7 @@ func (r *reconciler) syncWarehouse(
 ) (kargoapi.WarehouseStatus, error) {
 	status := *warehouse.Status.DeepCopy()
 	status.ObservedGeneration = warehouse.Generation
+	status.LastFreight = warehouse.Status.LastFreight
```
krancour, line 246: I'm not sure I understand the reason for this line. Isn't a copy of `LastFreight` included in the deep copy that was already performed?

internal/controller/warehouses/warehouses.go
```diff
@@ -290,10 +302,18 @@ func (r *reconciler) getLatestFreightFromRepos(
...
+		for _, commit := range warehouse.Status.LastFreight.Commits {
+			repoCommitMappings[commit.RepoURL+"#"+commit.Branch] = commit.ID
+		}
+	}
+
 	selectedCommits, err := r.selectCommitsFn(
 		ctx,
 		warehouse.Namespace,
 		warehouse.Spec.Subscriptions,
+		repoCommitMappings,
```
krancour, line 316: I think things would be tidier and more testable overall if we passed the `warehouse.Status.LastFreight` here and let `selectCommits()` build the map.

maksimstankevic:
@krancour thank you for reviewing! I will fix the last bits you commented about and handle merge conflicts, will try to make that happen until the end of this week.

internal/controller/warehouses/warehouses.go
maksimstankevic: I believe this is already done:

https://github.com/akuity/kargo/blob/407be79fb9dccaf77264043bfca37795cd3ca4dd/internal/controller/warehouses/warehouses_test.go#L43

krancour: Don't know how I missed that! 👍

internal/controller/warehouses/warehouses.go
maksimstankevic: removed:
https://github.com/akuity/kargo/pull/1477/commits/953eed13f54be0cc23c93edcf84962335162850b

internal/controller/warehouses/warehouses.go
maksimstankevic: refactored:
https://github.com/akuity/kargo/pull/1477/commits/b8667c672b0f4819c7522de56e0859e11d6595ba

internal/controller/warehouses/git.go
```diff
@@ -55,7 +56,16 @@ func (r *reconciler) selectCommits(
 			logger.Debug("found no credentials for git repo")
 		}
 
-		gm, err := r.selectCommitMetaFn(ctx, *s.Git, repoCreds)
+		repoCommitMappings := map[string]string{}
+		if lastFreight != nil {
+			for _, commit := range lastFreight.Commits {
+				repoCommitMappings[commit.RepoURL+"#"+commit.Branch] = commit.ID
+			}
+		}
```
krancour, line 64: Would it be better to build this map just once before the for loop?

maksimstankevic: it sure would! 🤦‍♂️ deemed it too small a fix after a long working day 🙂 Thanks @krancour 
https://github.com/akuity/kargo/pull/1477/commits/304f7ca24d1c8d1cadc16b512a35377b796f9298

maksimstankevic:
> @maksimstankevic this looks amazing! I commented inline with just small bits of feedback -- nothing major.
> 
> An unfortunate thing we do have to contend with here is a lot of merge conflicts. Most of them seem to arise from our codegen process having undergone an extreme makeover in #1515. Those are probably not resolvable by hand.
> 
> If you want to resolve them yourself, the easiest approach would be:
> 
> 1. Squash everything you've done to date in this PR into a single commit
> 2. Remove all changes to the following generated files:
>    
>    * api/v1alpha1/types.proto
>    * zz_generated.deepcopy.go
>    * charts/kargo/crds/kargo.akuity.io_warehouses.yaml
>    * internal/api/types/v1alpha1/types.go
>    * pkg/api/v1alpha1/types.pb.go
>    * ui/src/gen/schema/warehouses.kargo.akuity.io_v1alpha1.json
>    * ui/src/gen/v1alpha1/types_pb.ts
> 3. Rebase on `main` -- `internal/controller/warehouses` will have some conflicts to resolve, but I don't think they'll be too bad compared to generated code conflicts
> 4. Re-run `make hack-codegen`
> 5. Force push
> 
> Alternatively, I am happy to run point on resolving those because it's my slow review that put this behind #1515 in the first place. 🙏
> 
> Just let me know how you want to proceed.

@krancour After I rebased from main hack-codegen started failing in my setup because of some recent changes... I will continue troubleshooting and try to get it pushed as soon as I can.

krancour:
@maksimstankevic 😬 if coming from the old codegen process to the new, you might have to manually delete the `gomodcache` to unstick some permissions issues.

```
docker volume rm gomodcache
```

If that doesn't get you moving, lmk and I will prioritize figuring out what's borked.

maksimstankevic:
> @maksimstankevic 😬 if coming from the old codegen process to the new, you might have to manually delete the `gomodcache` to unstick some permissions issues.
> 
> ```
> docker volume rm gomodcache
> ```
> 
> If that doesn't get you moving, lmk and I will prioritize figuring out what's borked.

@krancour yeah I figured out, I actually pushed the rebased changes already, now it's failing on linter... https://github.com/akuity/kargo/actions/runs/8350487015/job/22856994599?pr=1477
maybe add exceptions for that? Not really sure how to resolve...

maksimstankevic:
@krancour I found out about deprecated "github.com/pkg/errors" https://github.com/vmware-tanzu/community-edition/issues/4850, will fix it tomorrow

krancour:
Ya. We moved off of that within the past week.

maksimstankevic:
@krancour everything is passing eventually :) can you take a look?

krancour:
@hiddeco can you be a second set of 👀 on this (again) please? 🙏

api/v1alpha1/warehouse_types.go
```diff
@@ -126,6 +126,20 @@ type GitSubscription struct {
 	// should be ignored when connecting to the repository. This should be enabled
 	// only with great caution.
 	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty" protobuf:"varint,7,opt,name=insecureSkipTLSVerify"`
+	// ScanPaths is a list of regular expressions that can optinally be used to
```
hiddeco, line 129:
```suggestion
	// ScanPaths is a list of regular expressions that can optionally be used to
```

internal/controller/git/git.go
```diff
@@ -316,6 +319,24 @@ func (r *repo) GetDiffPaths() ([]string, error) {
 	return paths, nil
 }
 
+func (r *repo) GetDiffPathsSinceCommitID(commitId string) ([]string, error) {
+	resBytes, err := libExec.Exec(r.buildCommand("diff", "--name-only", commitId+"..HEAD"))
+	if err != nil {
+		return nil,
+			fmt.Errorf("error getting diffs since commit %q %w", commitId, err)
+	}
+	paths := []string{}
```
hiddeco, line 328: Nit:

```suggestion
	var paths []string
```

internal/controller/warehouses/git.go
```diff
@@ -26,8 +26,17 @@ func (r *reconciler) selectCommits(
...
 	namespace string,
 	subs []kargoapi.RepoSubscription,
+	lastFreight *kargoapi.FreightReference,
 ) ([]kargoapi.GitCommit, error) {
 	latestCommits := make([]kargoapi.GitCommit, 0, len(subs))
+
+	repoCommitMappings := map[string]string{}
+	if lastFreight != nil {
+		for _, commit := range lastFreight.Commits {
+			repoCommitMappings[commit.RepoURL+"#"+commit.Branch] = commit.ID
```
hiddeco, line 36: Nit: as the size of the map always equals the number of commits, you could do:

```suggestion
	var repoCommitMappings map[string]string
	if lastFreight != nil {
		repoCommitMappings = make(repoCommitMappings, len(lastFreight.Commits))
		for _, commit := range lastFreight.Commits {
			repoCommitMappings[commit.RepoURL+"#"+commit.Branch] = commit.ID
```

hiddeco (dismissed):
I have left some comments, but none of them (except for the typo in the API spec _maybe_) are really blockers. I primarily wrote them as constructive feedback to help you grow.

Excellent job @maksimstankevic! 🥇

krancour:
One other thing I was thinking about -- @maksimstankevic and @hiddeco, I'd like to hear both your opinions -- I personally _do_ like "scanPaths", but can't help but wondering if "includePaths" would be more symmetric with "ignorePaths." Include and ignore?

wdyt?

I don't feel strongly one way or the other, but felt it was worth bringing up.

hiddeco:
Yes, now that you are mentioning it I think this would be a good suggestion.

hiddeco:
(Another thing I really like, is the simplicity of `.gitignore` where `!` just becomes an "include". But this likely does not really pair well with how things currently are.)

maksimstankevic:
> One other thing I was thinking about -- @maksimstankevic and @hiddeco, I'd like to hear both your opinions -- I personally _do_ like "scanPaths", but can't help but wondering if "includePaths" would be more symmetric with "ignorePaths." Include and ignore?
> 
> wdyt?
> 
> I don't feel strongly one way or the other, but felt it was worth bringing up.

That was my initial thought (use include instead of scan), however at least for me personally it lacked intuitivity. I actually never liked the include/exclude wording in all kind of yaml pipelines triggering configurations out there and this was my first opportunity to bring a change :)  Include into where, exclude from where? If you take scan/ignore paired to a list of paths it's crystal clear what you intend to instruct the robot to do. But I'm happy to change it (together with all great @hiddeco amendments above) as it's 1 vote against 2 :) But then, I think, it should be includePaths/excludePaths not includePaths/ignorePaths?

Let me know your verdict @krancour, in case we go with the changes I'll try to close it tonight.

hiddeco:
> it should be includePaths/excludePaths not includePaths/ignorePaths?

This actually makes sense to me as well... But I will let @krancour decide.

---

One other note: as we would really like to include this in the release we hope to put out tomorrow, and (only) if you are comfortable with it, I would be more than willing to take this out of your hands to finish the last bits on time @maksimstankevic.

However, I also absolutely do understand it if you want to address things yourself!

maksimstankevic:
> One other note: as we would really like to include this in the release we hope to put out tomorrow, and (only) if you are comfortable with it, I would be more than willing to take this out of your hands to finish the last bits on time

@hiddeco I will soon be back home and implement all your suggestions and I hope in the meantime @krancour will share his final thought on filter field names so I'll add it too and by tomorrow morning you will have it. I like to finish what I started. :)

krancour:
> it should be includePaths/excludePaths not includePaths/ignorePaths?

Ah. Yes. include/exclude pair more nicely than include/ignore.

> If you take scan/ignore paired to a list of paths it's crystal clear what you intend to instruct the robot to do.

That _is_ crystal clear to me. _In context_ (subscriptions), includePaths/excludePaths feels equally clear. "The subscription _includes these paths and excludes these paths_."

To unblock us, I'm opting for includePaths/excludePaths just because it feels slightly more conventional.

We're pumping the brakes a bit on breaking changes after this release, but this isn't _truly_ set in stone until v1.0.0, so if we discover in hindsight that this was a poor decision, it can be reconsidered.

maksimstankevic:
> rename scanPaths/ignorePaths to includePaths/excludePaths and some re…

@krancour @hiddeco I pushed https://github.com/akuity/kargo/pull/1477/commits/41162b9cd6a0433478db1c48614b90ba331a8151

hiddeco:
Took this for a test drive and found an issue.

If you create a new branch with `git checkout --orphan <branch>` and add a single commit, the code around getting the commit metadata / set of changes will fail when an inclusion and/or exclusion rule has been configured:

> error getting latest Freight from repositories: error syncing git repo subscriptions: error determining latest commit ID of git repo "https://github.com/hiddeco/kargo-filter-example.git": error selecting commit from git repo "https://github.com/hiddeco/kargo-filter-example.git": commit "484a36189efb7c56b16a387799c60d26b537b30a" not applicable due to includePaths/excludePaths configuration in repo "https://github.com/hiddeco/kargo-filter-example.git"

hiddeco:
Another note around UX. I would expect the following to work (ignore everything, but include specific directory), but it doesn't. Not sure if that's great:

```
      Exclude Paths:
        .*
      Include Paths:
        guestbook/.*
```

maksimstankevic:
> Another note around UX. I would expect the following to work (ignore everything, but include specific directory), but it doesn't. Not sure if that's great:
> 
> ```
>       Exclude Paths:
>         .*
>       Include Paths:
>         guestbook/.*
> ```

This is by design. Please refer to the docs: https://github.com/akuity/kargo/pull/1477/files#diff-f57379e50bcccb7724767399090312223069b636d3292358d83f638b13aa22eb:~:text=//%20IncludePaths%20is%20a,rep%2Cname%3DexcludePaths%22%60. "includePaths" is the main filter, setting it to whatever subset of paths has same effect as enabling excludePaths on the rest of paths, secondly ignorePaths has higher weight than includePaths and if it happens so that result sets after matching includePaths/excludePaths overlap 100% - this means commit is not applicable.

@krancour when I talked about intuitivity for the old names (scanPaths/ignorePaths) I meant that those were very meaningful for the logics behind, when you set scanPaths - you want only those to be scanned and nothing else, when you additionally set ignorePaths - you want to limit a broad result from scanPaths.

This logic was there from beginning and if we want to change it, I will need time.

maksimstankevic:
> Took this for a test drive and found an issue.
> 
> If you create a new branch with `git checkout --orphan <branch>` and add a single commit, the code around getting the commit metadata / set of changes will fail when an inclusion and/or exclusion rule has been configured:
> 
> > error getting latest Freight from repositories: error syncing git repo subscriptions: error determining latest commit ID of git repo "https://github.com/hiddeco/kargo-filter-example.git": error selecting commit from git repo "https://github.com/hiddeco/kargo-filter-example.git": commit "484a36189efb7c56b16a387799c60d26b537b30a" not applicable due to includePaths/excludePaths configuration in repo "https://github.com/hiddeco/kargo-filter-example.git"
> 
> Second issue appears to be that when a second commit is added, it will first recognize it as new Freight but then sequentially fail:
> 
> ```
> Spec:
>   Subscriptions:
>     Git:
>       Branch:                     manifests
>       Commit Selection Strategy:  NewestFromBranch
>       Include Paths:
>         guestbook/.*
>       Repo URL:  https://github.com/hiddeco/kargo-filter-example.git
> Status:
>   Last Freight:
>     Commits:
>       Branch:           manifests
>       Id:               ebb349f4970130427c16938a8263bed34ed4975e
>       Message:          Update guestbook manifest
>       Repo URL:         https://github.com/hiddeco/kargo-filter-example.git
>     Name:               cbe7a9fedb7457ec3d94c6df74edbbde6a82e4be
>   Observed Generation:  7
> ```
> 
> ```
> Spec:
>   Subscriptions:
>     Git:
>       Branch:                     manifests
>       Commit Selection Strategy:  NewestFromBranch
>       Include Paths:
>         guestbook/.*
>       Repo URL:  https://github.com/hiddeco/kargo-filter-example.git
> Status:
>   Last Freight:
>     Commits:
>       Branch:           manifests
>       Id:               ebb349f4970130427c16938a8263bed34ed4975e
>       Message:          Update guestbook manifest
>       Repo URL:         https://github.com/hiddeco/kargo-filter-example.git
>     Name:               cbe7a9fedb7457ec3d94c6df74edbbde6a82e4be
>   Message:              error getting latest Freight from repositories: error syncing git repo subscriptions: error determining latest commit ID of git repo "https://github.com/hiddeco/kargo-filter-example.git": error selecting commit from git repo "https://github.com/hiddeco/kargo-filter-example.git": commit "ebb349f4970130427c16938a8263bed34ed4975e" not applicable due to includePaths/excludePaths configuration in repo "https://github.com/hiddeco/kargo-filter-example.git"
>   Observed Generation:  7
> ```
> 
> The root cause of this appears to be that there are no mechanics in place to guard against `baseCommit == HEAD`, or handling the empty diff path result when this happens.

I will check this.

krancour:
I think the existing logic is ok. I'll give my rationale...

If I don't overthink it, my intuition would be:

Everything is implicitly included unless specific things to include are enumerated, in which case, only those are included.

Nothing is implicitly excluded unless specific things to exclude are enumerated.

Combined, the rule would be "is it implicitly or explicitly included and not explicitly excluded?"

So @hiddeco your example seems odd. I would think if you only want to include one specific path, it would be sufficient to just list it as included and leave the excluded field alone.

To put all of this in context, let's think about how this feature is likely to be used:

I have a monorepo and only want to produce new freight when resources belonging to a specific app or microservice have changed. Include is useful. Exclude need not be used.

OR

I render manifests into specific dirs in main (instead of a stage-specific branch). So I want to exclude those dirs from triggering freight production (or else I create a loop). Exclude is useful and include need not be used.

OR

Combine the two cases. I want to subscribe to config for a certain app or microservice, but ignore rendered manifests that live somewhere under that app. Include apps/foo, but exclude apps/foo/stages.

I think this tracks with what @maksimstankevic described to us earlier:

> you want only those to be scanned and nothing else, when you additionally set ignorePaths - you want to limit a broad result from scanPaths.

The issue with the orphaned branch is interesting...

I may not have understood completely, but my takeaway was that there's a problem here if the last commit that produced freight and the current commit don't have common history and diffs between the two can therefore not be determined. I think this would only ever occur in a few cases:

Someone has edited the warehouse and pointed it at a different branch -- or a different repo even.

Some naughty developer force-pushed to main or something.

I think in either of these cases, we just need to detect that the two commits are not comparable and default to treating the selected commit as something that should trigger production of new freight.

maksimstankevic:
> > Took this for a test drive and found an issue.
> > If you create a new branch with `git checkout --orphan <branch>` and add a single commit, the code around getting the commit metadata / set of changes will fail when an inclusion and/or exclusion rule has been configured:
> > > error getting latest Freight from repositories: error syncing git repo subscriptions: error determining latest commit ID of git repo "https://github.com/hiddeco/kargo-filter-example.git": error selecting commit from git repo "https://github.com/hiddeco/kargo-filter-example.git": commit "484a36189efb7c56b16a387799c60d26b537b30a" not applicable due to includePaths/excludePaths configuration in repo "https://github.com/hiddeco/kargo-filter-example.git"
> > 
> > 
> > Second issue appears to be that when a second commit is added, it will first recognize it as new Freight but then sequentially fail:
> > ```
> > Spec:
> >   Subscriptions:
> >     Git:
> >       Branch:                     manifests
> >       Commit Selection Strategy:  NewestFromBranch
> >       Include Paths:
> >         guestbook/.*
> >       Repo URL:  https://github.com/hiddeco/kargo-filter-example.git
> > Status:
> >   Last Freight:
> >     Commits:
> >       Branch:           manifests
> >       Id:               ebb349f4970130427c16938a8263bed34ed4975e
> >       Message:          Update guestbook manifest
> >       Repo URL:         https://github.com/hiddeco/kargo-filter-example.git
> >     Name:               cbe7a9fedb7457ec3d94c6df74edbbde6a82e4be
> >   Observed Generation:  7
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > Spec:
> >   Subscriptions:
> >     Git:
> >       Branch:                     manifests
> >       Commit Selection Strategy:  NewestFromBranch
> >       Include Paths:
> >         guestbook/.*
> >       Repo URL:  https://github.com/hiddeco/kargo-filter-example.git
> > Status:
> >   Last Freight:
> >     Commits:
> >       Branch:           manifests
> >       Id:               ebb349f4970130427c16938a8263bed34ed4975e
> >       Message:          Update guestbook manifest
> >       Repo URL:         https://github.com/hiddeco/kargo-filter-example.git
> >     Name:               cbe7a9fedb7457ec3d94c6df74edbbde6a82e4be
> >   Message:              error getting latest Freight from repositories: error syncing git repo subscriptions: error determining latest commit ID of git repo "https://github.com/hiddeco/kargo-filter-example.git": error selecting commit from git repo "https://github.com/hiddeco/kargo-filter-example.git": commit "ebb349f4970130427c16938a8263bed34ed4975e" not applicable due to includePaths/excludePaths configuration in repo "https://github.com/hiddeco/kargo-filter-example.git"
> >   Observed Generation:  7
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The root cause of this appears to be that there are no mechanics in place to guard against `baseCommit == HEAD`, or handling the empty diff path result when this happens.
> 
> I will check this.

As far as I can tell this is working correctly. Detailed steps below.


1. Produce a freight as a starting point in one branch:
```
maksimstankevic@MaksimsMacBookAir kargo-876                     (helm) $ git commit -am "last commit in helm that produced freight"
[helm c7cd3b2] last commit in helm that produced freight
 1 file changed, 1 insertion(+), 1 deletion(-)
maksimstankevic@MaksimsMacBookAir kargo-876                     (helm) $ git push
```
resulting Warehouse:
```
maksimstankevic@MaksimsMacBookAir kargo                     (feature/1270-2) $ k get warehouses kargo-demo -o yaml|grep -A 10 status
status:
  lastFreight:
    commits:
    - branch: helm
      id: c7cd3b22b033de6417f9b847e14937bd533f1c84
      message: last commit in helm that produced freight
      repoURL: https://github.com/maksimstankevic/kargo-demp-876.git
    name: df2030e68117868489bc1f8f25b1509b8f895549
  observedGeneration: 3
```

2. Create an orphan branch to follow @hiddeco, push one commit there and reconfigure Warehouse's only Git Subcription to the new branch:
```
maksimstankevic@MaksimsMacBookAir kargo-876                     (helm) $ git checkout --orphan orphan_branch
Switched to a new branch 'orphan_branch'
maksimstankevic@MaksimsMacBookAir kargo-876 $ git commit -am "1st commit in orphan branch"
[orphan_branch (root-commit) a8fdde3] 1st commit in orphan branch
 12 files changed, 487 insertions(+)
.
.
maksimstankevic@MaksimsMacBookAir kargo-876                     (orphan_branch) $ git push --set-upstream origin orphan_branch
Enumerating objects: 20, done.
```
After warehouse is changed to use new orphan branch it produces a new freight because it does not have a baseCommit (this Git Repo Subscription is new and did not have a commit that produced Freight yet):
```
maksimstankevic@MaksimsMacBookAir kargo                     (feature/1270-2) $ k get warehouses kargo-demo -o yaml|grep -A 10 status
status:
  lastFreight:
    commits:
    - branch: orphan_branch
      id: a8fdde393265ae213e93aa9170cd59ef0536e4fe
      message: 1st commit in orphan branch
      repoURL: https://github.com/maksimstankevic/kargo-demp-876.git
    name: 53e40061fac4015f04773f729fcd41ea3d12fdf6
  observedGeneration: 4
```
Above special case is commented in the code: https://github.com/akuity/kargo/pull/1477/files#diff-2c205d00808318c7f1543995081bf9667196a96456ea71f026b910dd0dc1eadc:~:text=//%20In%20case%20includePaths,without%20applying%20filters.

On next reconciliations, naturally, the Warehouses gets the error attached that it could not get latest commit because there are no diffs between last commit and HEAD which are the same in this case:

```
maksimstankevic@MaksimsMacBookAir kargo                     (feature/1270-2) $ k get warehouses kargo-demo -o yaml|grep -A 15 status
status:
  lastFreight:
    commits:
    - branch: orphan_branch
      id: a8fdde393265ae213e93aa9170cd59ef0536e4fe
      message: 1st commit in orphan branch
      repoURL: https://github.com/maksimstankevic/kargo-demp-876.git
    name: 53e40061fac4015f04773f729fcd41ea3d12fdf6
  message: 'error getting latest Freight from repositories: error syncing git repo
    subscriptions: error determining latest commit ID of git repo "https://github.com/maksimstankevic/kargo-demp-876.git":
    error selecting commit from git repo "https://github.com/maksimstankevic/kargo-demp-876.git":
    commit "a8fdde393265ae213e93aa9170cd59ef0536e4fe" not applicable due to includePaths/excludePaths
    configuration in repo "https://github.com/maksimstankevic/kargo-demp-876.git"'
  observedGeneration: 4
``` 

Next commit (valid per filter rules) produces new Freight on first reconciliation and all the next ones fail again due to the same reason above (no diffs). Not adding any more command output for brevity.

I believe this is following the design. What do you think @krancour ?

hiddeco:
> On next reconciliations, naturally, the Warehouses gets the error attached that it could not get latest commit because there are no diffs between last commit and HEAD which are the same in this case

This is precisely the thing I find odd. It's already up-to-date, so why should it complain about filtering rules instead of it telling that there are simply no new changes?

maksimstankevic:
> The issue with the orphaned branch is interesting...
> 
> I may not have understood completely, but my takeaway was that there's a problem here if the last commit that produced freight and the current commit don't have common history and diffs between the two can therefore not be determined. I think this would only ever occur in a few cases:
> 
> Someone has edited the warehouse and pointed it at a different branch -- or a different repo even.
> 
> Some naughty developer force-pushed to main or something.
> 
> I think in either of these cases, we just need to detect that the two commits are not comparable and default to treating the selected commit as something that should trigger production of new freight.

@krancour this is already being handled by "repo#branch - commit" mappings, such a new wild first commit in a reconfigured repo is not there in our LastFreight history - so it produces Freight by default, please take a look at my testing above.

maksimstankevic:
> > On next reconciliations, naturally, the Warehouses gets the error attached that it could not get latest commit because there are no diffs between last commit and HEAD which are the same in this case
> 
> This is precisely the thing I find odd. It's already up-to-date, so why should it complain about filtering rules instead of it telling that there are simply no new changes?

I will leave this to be answered by @krancour, because I took the concept from existing commit tag filtering rules that as well result in an error getting last commit when rules are not satisfied.

hiddeco:
> After warehouse is changed to use new orphan branch it produces a new freight because it does not have a baseCommit (this Git Repo Subscription is new and did not have a commit that produced Freight yet):

After a good night of sleep, I can not reproduce the first issue I reported anymore either. So sorry about creating noise, as I probably messed something up due to being tired.

> [..] And I tend to agree with that because it is an error - we cannot get a new Freight from the Warehouse "black box" in given circumstances and we clearly explain why the error is happening in the logs - due to tag or path filtering rules.

But it's not a black box in this case, you _can_ know that HEAD == base commit. Which means the expectation is there to be no divergence, because there simply are not any changes which should be filtered. At present, I find the combination of last freight having a reference to the same commit the error complains about highly confusing. Because it makes me think something is wrong with my filtering configuration that I need to address (which actually is not the case, it's just waiting for a new HEAD).

```
Status:
  Last Freight:
    Commits:
      Branch:           manifests-2
      Id:               566ef4865767f9876ed6f4a88068449043d80d14
      Message:          Init
      Repo URL:         https://github.com/hiddeco/kargo-filter-example.git
    Name:               91d8238ac6da37f4a5b38a86f06fb7be150d86f5
  Message:              error getting latest Freight from repositories: error syncing git repo subscriptions: error determining latest commit ID of git repo "https://github.com/hiddeco/kargo-filter-example.git": error selecting commit from git repo "https://github.com/hiddeco/kargo-filter-example.git": commit "566ef4865767f9876ed6f4a88068449043d80d14" not applicable due to includePaths/excludePaths configuration in repo "https://github.com/hiddeco/kargo-filter-example.git"
  Observed Generation:  1
```

hiddeco (approved):
As the error is more of a user experience thing. I am inclined to say that this can be merged as is, and we can do refinements later.

maksimstankevic:
> As the error is more of a user experience thing. I am inclined to say that this can be merged as is, and we can do refinements later.

@hiddeco I see your point now and agree actually, I sent a tiny change, now logs are clean on reconciliations when latest commit == baseCommit
thanks for pointing that out!
Please let me know if that is ok

hiddeco:
Think this addresses it nicely, thanks 🙇

krancour:
@maksimstankevic and @hiddeco great to sit down this morning and see you guys have hashed everything out!

I want to raise one final UX thing that I am sorry to say I overlooked until just now...

I get that it was likely modeled after tag filters, but are regular expressions really the best way to do path matching? I have some doubts about that. At least it doesn't feel like something I have commonly encountered.

wdyt?

hiddeco:
I was on the fence about this myself as well. Based on experience in another project, we used a `.gitignore` format in the form of:

```
spec:
  ignore: |
    */
    !guestbook/
```

This worked really well, _except_ that extensive ignore rules can become complex in cases where you have patterns in what you want to include (e.g. `stages/{a,b,c}`).

maksimstankevic:
> @maksimstankevic and @hiddeco great to sit down this morning and see you guys have hashed everything out!
> 
> I want to raise one final UX thing that I am sorry to say I overlooked until just now...
> 
> I get that it was likely modeled after tag filters, but are regular expressions really the best way to do path matching? I have some doubts about that. At least it doesn't feel like something I have commonly encountered.
> 
> wdyt?

My 2 cents, I am working as devops and part-time developer for 10 years now and when I saw this issue open I knew it has to absolutely be regexps in both filters right from the start. It gives me as a user the flexibility I need and if it hadn't this support I would bomb you with CRs for implementing it. I'm pretty sure most devops and developers know how to use regexp and that should not be an obstacle. I don't feel it's hitting performance either, but I don't pretend to be an expert here.

krancour:
@maksimstankevic I think my worry is that someone who doesn't read the docs might intuitively do something like `foo/*` which is a pretty common way of saying "everything in `foo/`," but when compiled as a regex, matches `foobar/`.

hiddeco:
What I suggest to overcome this issue (and also to potentially allow adding e.g. glob support later) is to require an identifier to be specified.

For example:

```
includePaths:
- regexp:<actual expression>
```

Based on experience with these kind of identifiers, I would both allow `regexp` and `regex`, as there seem to be two groups of people.

maksimstankevic:
@krancour yes I understand that perfectly. And I still need regexp here as a user.

Anything you decide, I can work on it. Not before the release though, I actually need a break, can that be done as a refinement after 0.0.5?

krancour:
@maksimstankevic you've been so accommodating. We're intent on getting this into v0.5.0. At the start of our current release cycle, we were fully prepared to be implementing this ourselves and it was a surprise to see a community member step up.

@hiddeco will take this the last mile and change nothing other than requiring a prefix to signal use of a regex and will otherwise assume items in the lists to be full paths or path prefixes.

All credit for this PR goes to you. You're going to end up with a big thank you in the release notes.

krancour (approved):
This gets an enthusiastic LGTM from me.

@hiddeco kept it as regex matching only, but simply requiring the "regex" or "regexp" prefix.

We can add prefix and glob based path matching in v0.6.0.

krancour:
@maksimstankevic thank you so much for the extensive work on this!

maksimstankevic:
> This gets an enthusiastic LGTM from me.
> 
> @hiddeco kept it as regex matching only, but simply requiring the "regex" or "regexp" prefix.
> 
> We can add prefix and glob based path matching in v0.6.0.

@hiddeco @krancour thanks!

I'd like to work on adding the other 2 matching methods if possible...

krancour:
@maksimstankevic that sounds great! Thank you.

PR #1481, opened 2024-02-16
fix(controller): simplify exec of kargo edit set image command

### LINKED ISSUE #1476: `kustomize edit set image` should not set `newName`
# Description

When Kargo sets an image in the kustomization.yaml as part of image promotion, it will add `newName`, even though the newName is the same as name. e.g.:

```yaml
images:
- name: ghcr.io/jessesuen/guestbook
  newName: ghcr.io/jessesuen/guestbook # << not desirable
  newTag: 0.2.0
```

This is because kargo runs:

```
kustomize edit set image ghcr.io/jessesuen/guestbook=ghcr.io/jessesuen/guestbook:0.2.0
```

Instead of that command, kargo should run the following:

```
kustomize edit set image ghcr.io/jessesuen/guestbook:0.2.0
```

Fix would go here:
https://github.com/akuity/kargo/blob/main/internal/kustomize/kustomize.go#L18-L28

# Version

0.3.0

krancour:
Fixes #1476

### DIFF
--- a/internal/controller/promotion/kustomize.go
+++ b/internal/controller/promotion/kustomize.go
@@ -41,7 +41,7 @@ func selectKustomizeUpdates(updates []kargoapi.GitRepoUpdate) []kargoapi.GitRepo
 // kustomizer is a helper struct whose sole purpose is to close over several
 // other functions that are used in the implementation of the apply() function.
 type kustomizer struct {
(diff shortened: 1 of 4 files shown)


gdsoumya (approved):
LGTM

Issue #1443, opened 2024-01-31
Allow Kargo to adopt existing namespaces


jessesuen:
# Proposed Feature

The Project CRD is being introduced in 0.4. I upgraded my instance to a nightly build with this feature. I then tried to backfill projects for my existing kargo namespaces. But when I try to create project for an existing kargo namespace created in 0.3, I hit the following error:

```
Error from server (Conflict): error when creating "my-project.yaml": admission webhook "project.kargo.akuity.io" denied the request: Operation cannot be fulfilled on projects.kargo.akuity.io "my-project": cannot create Project "my-project" because namespace "my-project" already exists
```

When creating a project for a namespace that already exists, we should allow this to go through as long as the namespace has the `kargo.akuity.io/project: "true"` project label. Otherwise, I have no way of creating projects, short of deleting the admission webhook.

# Motivation

We need to be more flexible to the possibility that namespaces may already exist. There are two use cases:
1. v0.3 to v0.4 upgrade case where Kargo namespaces were already created
2. User has an existing namespace, which they would now like to use for Kargo purposes

# Suggested Implementation

Relax the webhook/management controller to allow project creation if `kargo.akuity.io/project: "true"` is set on the namespace. If so we would "adopt" the existing namespace.

NOTE: the adoption behavior is up for debate. I do wonder if users would prefer that Namespaces be created explicitly/out-of-band from Kargo.

PR #1482, opened 2024-02-16
refactor: make management controller able to create project namespaces again

krancour:
Since #1431 we rely on a webhook to create Project namespaces synchronously. Although this was a valid decision made for well-documented reasons, one undesired side-effect has been difficulty bootstraping Kargo with a default Project already in place:

* If the Project is applied before the webhook registration, Project creation succeeds, but creation of the underlying namespace does not take place. If the manifests contain additional resources that belong in that namespace, their creation will fail.

* If the webhook registration is applied before the Project, Project creation will fail since the webhook server isn't running yet. Creation of the underlying namespace does not take place and, as above, if the manifests contain additional resources that belong in that namespace, their creation will fail.

Some unsatisfactory solutions we explored included:

* Ensuring the webhook server is ready _before_ attempting to create the Project.

    This was deemed to be difficult, or at least annoying.

* "Create projects the hard way" by directly creating an appropriately labeled namespace.

    This loses out on the benefits of _having_ a Project resource. Those aren't significant now, but will be of increasing consequence over time.

We settled on _partially_ rolling back #1431, allowing for a Project's underlying namespace to be created synchronously by a webhook _or_ asynchronously by the management controller. This ability enables a better option:

* "Create projects the hard way" by directly creating an appropriately labeled namespace, but _avoid_ losing out on the additional benefits of having a proper Project resource by _also_ defining that. It must _precede_ registration of the webhook, but that's easy enough to ensure. When Kargo comes up, the Project resource will be reconciled and will "adopt" the corresponding namespace at that time (a process we're already using in the webhook since #1443).

cc @gdsoumya

### DIFF
--- a/charts/kargo/templates/management-controller/cluster-roles.yaml
+++ b/charts/kargo/templates/management-controller/cluster-roles.yaml
@@ -12,6 +12,7 @@ rules:
   resources:
   - namespaces
   verbs:
+  - create
   - get
   - list
   - patch
--- a/internal/controller/management/projects/projects.go
+++ b/internal/controller/management/projects/projects.go
@@ -3,9 +3,16 @@ package projects
 import (
 	"context"
 
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


gdsoumya (approved):
LGTM

PR #1483, opened 2024-02-16
refactor(controller)!: remove option to borrow credentials from argo cd

krancour:
It is only _superficially_ that the creds one manages for Kargo and for Argo CD might look like the DRY violations that "credential borrowing" was originally conceived of to remediate.

Argo CD only _ever_ needs R/O creds, while Kargo _typically_ requires R/W. In most cases, "borrowing" creds from Argo CD therefore requires giving Argo CD permissions it doesn't need.

As discussed with @jessesuen @gdsoumya offline, we are removing the credential borrowing feature to dissuade violations of the principle of least privilege.


jessesuen (approved):
LGTM!

PR #1486, opened 2024-02-17
feat: grant API server permissions to manage project secrets on project create

krancour:
## Problem

If you use `kubectl apply -f ...` on a manfiest that looks like this, everything works great:

```yaml
kind: Project
metadata:
  name: kargo-demo
---
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: kargo-demo
  namespace: kargo-demo
  labels:
    kargo.akuity.io/secret-type: repository
stringData:
  type: git
  url: https://github.com/<username>/kargo-demo-gitops.git
  username: <username>
  password: <PAT>
```

If you use `kargo apply -f ...` instead, Secret creation will fail because the Kargo API server is not authorized to perform any CRUD operations on Secrets.

The API server currently lacks this ability because, with an inability to anticipate all the Project namespaces that might ever exist, giving it this ability would require granting permission to manage all Secrets cluster-wide. That is an unacceptable level of permissions to grant to a component that is internet-facing.

## Solution

The solution requires just-in-time expansion of the API server's permissions to manage Secrets _in specific Project namespaces only, as those Projects are created._ Moreover, this has to take place synchronously or else there is still a high likelihood of failure when applying the example manifest in the previous section.

This can be implemented by expanding the responsibilities of the existing admission review webhook for Projects, which is already used to address the similar concern of synchronous Project namespace creation.

The same just-in-time expansion of the API server's permissions to manage Secrets _in specific Project namespaces only_, must _also_ be handled asynchronously in order to accommodate edge cases in which a Project may have been created prior to webhook registration, in which case, the system will have some "catching up" to perform upon startup. #1482 provides more context for how such a scenario might unfold.

Because Kubernetes forbids subjects from delegating permissions that they themselves do not have, both the webhook server and management controller will require all permissions on all Secrets cluster-wide -- the very permission we wished _not_ to grant to the API server. This is more acceptable, however, because, unlike the API server, the webhook server and management controller are _not_ internet facing components.

## Complication

In the course of working on this PR, I discovered that if a webhook creates a resource (namespace, in this case) and assigns ownership to a _new_ resource under review (Project, in this case), and if the webhook doesn't complete fast enough, the resource under review won't formally exist yet, and any resource that references it as its owner will appear orphaned _and will be deleted by Kubernetes._

The webhook _already_ establishes new Project resources as the owners of their corresponding namespaces. Until now, the webhook has always (lucky for us) completed quickly enough to avoid the scenario described above. With the expansion of the webhook's responsibilities, it has become impossible to complete fast enough to avoid that scenario.

The solution is simply to remove the responsibility for assigning ownership from the webhook and leave it up to the Project reconciler only to handle that asynchronously.

## Detailed Break Down

1. Enabled debug logs for webhook server when running in Tilt

1. Added ClusterRole for managing Secrets

1. Gave webhook server all permissions on all secrets

1. Gave webhook server permission to create RoleBindings

1. Refactored Project webhook with no functional changes -- prep work for what follows

1. Fixed ownership issue

1. Made Project webhook synchronously create a RoleBinding that allows the API server to manage Secrets in the Project's namespace

1. Gave management controller all permissions on all secrets

1. Gave management controller permission to create RoleBindings

1. Made the Project reconciler asynchronously create a RoleBinding that allows the API server to manage Secrets in the Project's namespace

1. Corresponding Tiltfile changes -- just categorizing new resources to keep everything tidy in the Tilt dashboard

1. Remove one unused field in API server config that I noticed while working on this

---

cc @dhpup as this addresses something that has been a pain point for you recently.


charts/kargo/templates/common/cluster-roles.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.rbac.installClusterRoles }}
```
gdsoumya, line 1: Thinking if this should be part of the api server template folder as this is needed by api server only?

gdsoumya (approved):
Rest looks good!

charts/kargo/templates/common/cluster-roles.yaml
krancour: I'm good with that. Will move.

krancour: Fixed in 3624f193f568efc041b6b8e657322cd816dc6289

PR #1501, opened 2024-02-20
docs: update quickstart for v0.4

krancour:
Before a release, I usually recruit help to validate, among other things, that the updated quickstart works flawlessy against a release candidate.

This, unfortunately, requires reviewers to be aware of which parts of the quickstart may require small tweaks. As a for instance, one cannot install the "latest" version of the Kargo chart, because Helm counts something like `v0.4.0-rc.1` as pre-release and will only install it if you explicitly request it. The commands are buried in a script, which needs to be downloaded and modified instead of simply curl|bash'ed...

It's a pain, frankly.

I'm trying something different this time, for the convenience of all who help to validate this. I've updated all those little things _for you_ in a temporary commit that I will remove before this PR gets merged. 😄 

Just follow [the quickstart in the Netlify preview of this PR](https://deploy-preview-1501.kargo.akuity.io/quickstart) exactly.

### DIFF
--- a/docs/docs/20-quickstart.mdx
+++ b/docs/docs/20-quickstart.mdx
@@ -220,10 +220,10 @@ UAT, and production.
 
 ### Create Argo CD `Application` Resources
 
-In this step, we will create three Argo CD `Application` resources that deploy
-the sample application at three different stages of its lifecycle, with three
-slightly different configurations, to three different namespaces in our local
-cluster:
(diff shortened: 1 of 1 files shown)


sennerholm:
@krancour, do you want more testers of this?

krancour:
@sennerholm the more the merrier. Always love the extra validation!

Issue #872, opened 2023-09-28
Support git repos with http/non-ssl URL


krancour:
From #867

Support for git repos with http/non-ssl URL has been requested.

fwiw, I think there are a lot of security implications to consider before deciding to proceed with this.

krancour:
Discussed with @jessesuen.

My initial concern here was that I don't want to see credentials passed over an insecure connection. The counterpoint is that devs may sometimes run local git servers without SSL. I was reluctant to enable the latter by opening the door to the former.

The middle ground we agreed on is similar to what Kubernetes does -- allow non-SSL connections, but refuse to pass credentials over them.

So implementation here looks like:

* Relax a few regexes
* Make the credential DB log a warning and return nil if/when it looks up creds for a URL beginning with `http://`

This is sufficiently easy that I'm adding this to hacktoberfest.

hiddeco:
> Make the credential DB log a warning and return nil if/when it looks up creds for a URL beginning with `http://`

If taken literally (i.e. without taking the type into account), this would appear to have an effect on Helm repositories as well. As it seems HTTP endpoints are currently allowed here: https://github.com/akuity/kargo/blob/ba55091734a723b0d53dc4610b826ed1cbdb7e36/api/v1alpha1/warehouse_types.go#L193

While I do not necessarily think this would be bad (the same logic around _not_ wanting to pass credentials over an insecure wire could apply here), it would be breaking.

krancour:
> it would be breaking

Breaking changes are ok for the time being, but that may end soon.

We're in "get 'em in while you still can" territory.

PR #1504, opened 2024-02-21
feat: allow connecting to HTTP Git repositories

hiddeco:
Fixes: #872

This adds support for connecting to Git repositories over an (insecure) HTTP wire as described in https://github.com/akuity/kargo/issues/872#issuecomment-1761972854, while taking into account the effect it would have on HTTP Helm repositories as observed in https://github.com/akuity/kargo/issues/872#issuecomment-1956378143.

Effectively, this means that connecting to any HTTP endpoint is now allowed as long as this does not require authentication. This includes "classic" Helm repositories served over HTTP, where basic authentication used to be supported in the past.

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -10,6 +10,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/akuity/kargo/internal/git"
+	"github.com/akuity/kargo/internal/logging"
 )
 
 const (
@@ -99,6 +100,15 @@ func (k *kubernetesDatabase) Get(
 	repoURL string,
 ) (Credentials, bool, error) {
 	creds := Credentials{}
+
+	// If we are dealing with an insecure HTTP endpoint (of type Git),
+	// refuse to return any credentials
+	if credType == TypeGit && strings.HasPrefix(repoURL, "http://") {
+		logger := logging.LoggerFromContext(ctx).WithField("repoURL", repoURL)
+		logger.Warnf("refused to get credentials for insecure HTTP endpoint")
+
+		return creds, false, nil
+	}
 
 	var secret *corev1.Secret
 	var err error
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -101,9 +101,9 @@ func (k *kubernetesDatabase) Get(
 ) (Credentials, bool, error) {
 	creds := Credentials{}
 
-	// If we are dealing with an insecure HTTP endpoint (of type Git),
+	// If we are dealing with an insecure HTTP endpoint (of any type),
 	// refuse to return any credentials
-	if credType == TypeGit && strings.HasPrefix(repoURL, "http://") {
+	if strings.HasPrefix(repoURL, "http://") {
 		logger := logging.LoggerFromContext(ctx).WithField("repoURL", repoURL)
 		logger.Warnf("refused to get credentials for insecure HTTP endpoint")
 
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -195,7 +195,7 @@ type GitRepoUpdate struct {
 	// RepoURL is the URL of the repository to update. This is a required field.
 	//
 	//+kubebuilder:validation:MinLength=1
-	//+kubebuilder:validation:Pattern=`^https://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
+	//+kubebuilder:validation:Pattern=`^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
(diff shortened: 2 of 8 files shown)


internal/credentials/credentials.go
```diff
@@ -100,6 +101,15 @@ func (k *kubernetesDatabase) Get(
 ) (Credentials, bool, error) {
 	creds := Credentials{}
 
+	// If we are dealing with an insecure HTTP endpoint (of type Git),
```
krancour, line 104: See [comment](https://github.com/akuity/kargo/issues/872#issuecomment-1957427712)

Ok with the breaking change here. We can get away with it for a short while longer and should take advantage while we can.

krancour: Apart from this, the rest LGTM

krancour (approved):
🔥

geowalrus4gh:
waiting for v0.5.0 🤞

geowalrus4gh:
tried the latest [0.4.1-unstable-20240223] but getting the below error 
`spec.promotionMechanisms.gitRepoUpdates[0].repoURL in body should match '^https://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$',`

geowalrus4gh:
Yes, I installed a fresh Kargo helm installation after deleting old CRDs. Here is the extract from the new CRD installed. 
`repoURL:
                          description: URL is the repository's URL. This is a required
                            field.
                          minLength: 1
                          pattern: ^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$
`

krancour:
@geowalrus4gh what is the command you used to apply? Did you use kubectl or kargo CLI?

geowalrus4gh:
i used helm command to install. 
`helm upgrade --install kargo --create-namespace -n kargo oci://<harbor-url>/ghcr/akuity/kargo-charts-unstable/kargo --version 0.4.1-unstable-20240223 -f values.yaml --wait`

I copied the crds from the github and applied, then it worked. 

But now am getting an authentication failed error. 
Also, there is a warning shown like this

`time="2024-02-24T10:14:06Z" level=warning msg="refused to get credentials for insecure HTTP endpoint" freight=a508d8e8281d5b453cb4719f2e9f1b47c6f2afd9 namespace=sample-kargo promotion=dev.01hqd9q685kx5fmnw1t42jdcgf.a508d8e repoURL="http://bitbucket.company.com/scm/samp/sample-env.git" stage=dev
`
Does that mean, the repo secret for the http URL is not being read by kargo?

krancour:
That is correct. It was a compromise we arrived at in order to support unsecured connections -- we won't pass credentials over them.

tal-hason:
@krancour, How are you doing?

I didn't fully understand what that meant?!

if I want to use a HTTP git repo the Kargo Controller will not use the creds stored in the secret.
so how will it do a push to the remote!?

krancour:
Hey @tal-hason, good to hear from you!

> if I want to use a HTTP git repo the Kargo Controller will not use the creds stored in the secret.

Making sure you spotted the missing "S". (Easy to overlook.) This is strictly for HTTP and not HTTPS. Kargo is refusing to pass credentials over an unsecure connection.

Practically speaking, this shouldn't be much of a limitation. The only time I can think of where you would conceivably access a repo over an unsecure connection would be during local development, in which case you can just skip the creds entirely and configure the repo to allow pushes without authenticating.

krancour:
> Also, you don't have to pass the creds in the HTTP url.

> Store it in the .gitconfig and in the .git-credentails and the git will use it.

Well, both of these are technically true. The credentials are passed as a _header_. And on an insecure connection the header is perfectly accessible to anyone or anything reading the request. This is exactly why the decision was made in the first place to disallow the use of creds over an unsecure connection. After even a single use, you have to assume the credentials could have been compromised.

At the time we made that decision, there was some precedent for it (or so we thought?), as we could recall kubectl disallowing the use of credentials over an unsecure connection. Just now, I've been searching for any reference to that and keep coming up empty-handed, perhaps partly because as far back as v1.22.0, unsecure connections to the API server aren't even possible. So if our recollection was correct, it was in the _distant_ past and since then, if anything, Kubernetes has moved in a _more_ dogmatically secure-by-default position.

Anyway -- that's all background.

We discussed at our team meeting this morning and all agreed that we will consider adding this as some kind of highly discouraged option in the future _if_ we hear of more users / use cases that require this. Even as an opt-in, we don't want to relax security if this use case is an extreme outlier.

Please feel free to open a new issue. If the issue attracts a lot of attention, it will strengthen the case for this option to exist and may also accelerate the timeframe for getting it done.

tal-hason:
@krancour ,thanks as always on the detailed answer 😃.

I can tell you the I going at least 3-4 customers that have the potential to work with Kargo that have unsecured git server, do to that their network is disconnected.

I understand the secure by design concept and constraints, but security at the end is the customer risk to take.

I will Open an issue and see what is what.

PR #1505, opened 2024-02-21
fix: transform uname arch into valid Go arch

hiddeco:
On an AMD64 system, `uname -m` will return `x86_64` instead of the Go accepted `amd64`. Detect when this is returned (or provided), and overwrite it with the correct value.

This ensures the `build-cli` Make target actually works on AMD64 Linux systems.


krancour:
Nice catch. We accounted for this in CLI installation docs and missed it here.

krancour (approved):
lgtm

Issue #1016, opened 2023-10-23
Helm option to disable creation of secret


jessesuen:
# Proposed Feature

I agree with, and understand the rationale for https://github.com/akuity/kargo/pull/829, which forgoes a default password hash and token signing key.

However, in my case, I don't want kargo to manage the secret, because I want to create the secret myself through an ExternalSecret.

# Motivation

Right now I am forced to workaround the issue by create dummy values for admin password hash, and then use kustomize to remove the secret, and add the External Secret 

# Suggested Implementation

Many charts have a `create: true` option for secrets. I am proposing we add one ourselves:

```yaml
api:
  secret:
    # creates the `kargo-api` secret. Set to false if the `kargo-api` Secret is created by some other means
    create: true
```

krancour:
I'm sorry I let this one slide for so long. This is an obvious week point in our security. Sure -- we can create secrets for you, but the value that goes _in_ that secret still has to be in plaintext in a values.yaml file somewhere, and that makes it unsuitable to store in a repo. 😢 

The proposed feature here would allow users to manage their own secret "out of band" using whatever secret-management they prefer -- Sealed Secrets, for instance.

This is overdue and small, so I will add it to v0.5.0.

Community help is welcome, but I will assign to myself if we approach the v0.5.0 release without this complete.

The only thing I would change from the initial proposal would be instead of using a boolean to say you've provided the secret yourself, let's use a string that points to a secret. If it's left empty, we'll create the secret like we do now, but otherwise use the secret that was referenced.

PR #1507, opened 2024-02-21
feat(chart): allow defining Secret reference

hiddeco:
#1016

By configuring `api.secret.name`, the chart will no longer generate the Secret by itself but rather assume the referenced Secret is created out of band (using e.g. Sealed Secrets).

This effectively makes `.adminAccount.passwordHash`, `.adminAccount.password` and `.adminAccount.tokenSigningKey` optional when `api.secret.name` is specified.


krancour (approved):
🔥

PR #1515, opened 2024-02-22
refactor(api)!: make Kubebuilder types protobuf compatible

### LINKED ISSUE #708: Return native kube API types instead of a proto conversion

devholic:
Resolves #708 

---

**🚨 THIS PR CONTAINS BREAKING CHANGES (NON-BACKWARD COMPATIBLE) 🚨**

## Background

Current implementation of converting between Go types (Kubebuilder) and Protobuf types (API Request / Response) may cause unintended bugs since the conversion code should be generated by hand.

This PR removes hand-written type conversion code by making Kubebuilder types protobuf compatible.

## Changes

Now `codegen` target will generate _intermediate_ protobuf definition from the Kubebuilder types using [go-to-protobuf (part of kubernetes/code-generator)](https://github.com/kubernetes/code-generator/tree/v0.29.1/cmd/go-to-protobuf).

Since we assume Kubebuilder structs as a source of truth, we will not use generated protobuf code from above. Instead, [`hack/codegen/prototag`](https://github.com/akuity/kargo/blob/7e798210f4a4b8de8a7a46ae431c7044e22c465a/hack/codegen/prototag/main.go) will extract `protobuf` tag from the generated code then inject into the Kubebuilder structs. This will help us not to  write/manage `protobuf` tags.

### Limitations

- `json` support

`gogo/protobuf` extends `google/protobuf` to support various features. However these extended feature doesn't work on cross-platform environment - for example, json encoding/decoding for some messages (e.g. metav1.Time, `one_of` message) fails between API and UI.

To make things simple, this PR makes binary-encoded protobuf as a first-class citizen, and drops `json` support from API for the consistency.


hack/install-tools.sh
```diff
@@ -0,0 +1,61 @@
...
+  workdir=$(dirname "$0")
+
+  local programs
+  programs=$(get_imports "${workdir}/tools.go")
+
+  for program in ${programs}; do
+    version=$(get_module_version "${program}")
+
+    echo "Installing ${program}@${version}..."
+    go install "${program}@${version}"
```
hiddeco, line 57: By no means a blocker, but an observation as I have been diving into the build/devx related parts of Kargo the past day. :cherries: 

---

When you are working on multiple Kubernetes related projects, `controller-gen` tends to be on different versions across these projects quite a lot. In addition, assuming that e.g. a contributor does not already have _some_ version of a tool on their system (and us overwriting it) is quite invasive. 

Given this, I am wondering if there have been chats about collecting dependency tools in a directory relative to the project root, e.g. `./hack/bin`. Instead of globally installing them on a developer their system.

A common trick to pull this off is to set `GOBIN` to a newly created temporary directory before running `go install`, and to then copy any binaries from there. See for example: https://github.com/DataDog/datadog-operator/blob/a4cf7a31e463b69dce523f9ef41aab2aaa5c1421/Makefile#L83-L108

.github/workflows/ci.yaml
```diff
@@ -109,15 +111,26 @@ jobs:
...
+      uses: arduino/setup-protoc@v3
+      with:
+        version: "25.3"
+    - name: Install linter
+      env:
+        GOLANGCI_LINT_VERSION: 1.54.2
+      working-directory: /usr/local/bin
+      run: |
+        curl -sSfL https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64.tar.gz \
+        | tar xvz golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64/golangci-lint --strip-components=1
```
krancour, line 127: I'm curious why we need this linter here.

Makefile
```diff
@@ -66,7 +66,11 @@ lint-go:
 
 .PHONY: lint-proto
 lint-proto:
+	# Vendor go dependencies to build protobuf definitions
+	go mod vendor
```
krancour, line 70: We should probably add `vendor/` to the `.gitignore` file if we're going to do this.

Makefile
```diff
@@ -66,7 +66,11 @@ lint-go:
...
 	buf lint api --error-format=$(BUF_LINT_ERROR_FORMAT)
+	# Clean up vendor directory
+	rm -r $(CURDIR)/vendor
```
krancour, line 73: Cleanup won't happen if the previous line fails.

Makefile
```diff
@@ -124,10 +135,13 @@ nightly-cli:
 ################################################################################
 # Code generation: To be run after modifications to API types                  #
 ################################################################################
+APIMACHINERY_PKGS=k8s.io/apimachinery/pkg/util/intstr,+k8s.io/apimachinery/pkg/api/resource,+k8s.io/apimachinery/pkg/runtime/schema,+k8s.io/apimachinery/pkg/runtime,k8s.io/apimachinery/pkg/apis/meta/v1,k8s.io/api/core/v1,k8s.io/api/batch/v1
 
 .PHONY: codegen
-codegen:
-	buf generate api
+codegen: deps-tools codegen-proto codegen-controller codegen-ui codegen-docs
```
krancour, line 141: The `hack-codgen` target runs `make codegen` inside a container that already has all the tools pre-installed.

You should probably update `hack-codegen` to run `make codegen-proto codegen-controller codegen-ui codegen-docs` instead.

api/v1alpha1/groupversion_info.go
```diff
@@ -12,16 +12,16 @@ import (
 
 var (
 	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{
-		Group:   "kargo.akuity.io",
-		Version: "v1alpha1",
+	GroupVersion	= schema.GroupVersion{
+		Group:		"kargo.akuity.io",
+		Version:	"v1alpha1",
```
krancour, line 17: Where did all this extra space come from? This doesn't look gofmt'ed.

Same question in a lot of other places.

charts/kargo/crds/kargo.akuity.io_projects.yaml
```diff
@@ -71,8 +71,6 @@ spec:
                       minLength: 1
                       pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
                       type: string
-                  required:
-                  - stage
```
krancour, line 75: It doesn't look right that this got removed.

Same comment applies in other places.

ui/src/gen/schema/projects.kargo.akuity.io_v1alpha1.json
```diff
@@ -31,9 +31,6 @@
                 "type": "string"
               }
             },
-            "required": [
-              "stage"
-            ],
```
krancour, line 36: Similar comment made about CRD definitions. This looks like it shouldn't be removed.

I am wondering if `//+kubebuilder:validation:Required` is being ignored?

Makefile
devholic: `.gitignore` already has `vendor/`: https://github.com/akuity/kargo/blob/ebf0dfc8224ba2414999da14f8ffdaf95c2eee3c/.gitignore#L7

.github/workflows/ci.yaml
devholic: I used it for formatting updated Kubebuilder code. Changed to using `go/format` in `prototag` util in https://github.com/akuity/kargo/pull/1515/commits/656c6154338b3edbe011ff93effe7deb53da9d14. It will also resolve https://github.com/akuity/kargo/pull/1515#discussion_r1500090417.

api/v1alpha1/groupversion_info.go
devholic: Resolved in https://github.com/akuity/kargo/pull/1515#discussion_r1500519314

Makefile
```diff
@@ -137,11 +149,48 @@ codegen:
...
+	cp -R ${GOPATH}/src/github.com/akuity/kargo/api $(CURDIR)
+
+	# Generate protobuf code (Go)
+	buf generate api
+	# Inject generated protobuf tag to Kubebuilder structs
+	go run $(CURDIR)/hack/codegen/prototag/main.go \
+		-src-dir=$(CURDIR)/pkg/api/v1alpha1 \
+		-dst-dir=$(CURDIR)/api/v1alpha1
+	# Clean up intermediate file
+	rm -r $(CURDIR)/pkg/api/v1alpha1
```
devholic, line 180: @krancour As far as I know, there is no same concept with Go's defer. Could you recommend any way to clean up intermediate files even though the previous commands are failed?

hack/install-tools.sh
krancour: He opened #1525, so let's count this as resolved.

Makefile
krancour: I looked. idk how I missed it. So sorry!

ui/src/gen/schema/projects.kargo.akuity.io_v1alpha1.json
krancour: > I didn't noticed that the omitempty has presedence over kubebuilder's
Required marker

That has tripped me up before as well, as recently as yesterday.

Makefile
krancour: TIL

hack/codegen/prototag/main.go
```diff
@@ -0,0 +1,86 @@
+package main
```
krancour, line 1: Can we get some comments in here to document what this tool is used for?

Makefile
devholic: I just found that `.DELETE_ON_ERROR` doesn't work against directories... https://savannah.gnu.org/bugs/?16372

I resolved it by using bash `trap`: https://github.com/akuity/kargo/pull/1515/commits/c1135af5e095657f5720a434f9ab60725282519d

hack/codegen/prototag/main.go
devholic: Add comments: https://github.com/akuity/kargo/pull/1515/commits/003b4607215101c2fd9c70b2c7d8c9bcb096f0d2

krancour:
@devholic this is awesome work!

I'm fine with everything I see here, but still want to test-drive it just to make sure I really understand how the new codegen process feels. I will do on Tuesday.

I also just want to check with @jessesuen and @gdsoumya that they have no specific objections to using only binary-encoded protobufs.

krancour:
@gdsoumya and @devholic did you connect re: the binary-only change?

devholic:
@krancour Yes, we discussed about the change and it will not be a problem if binary format used between just user client (CLI/UI) and Kargo API server :) (cc @gdsoumya)

krancour:
@devholic this works perfectly!

Slower than the old process, but a very small price to pay for no longer having to write type conversions by hand!

🔥 🔥 🔥

krancour (approved):

PR #1517, opened 2024-02-22
fix(devx): mount permission issues on Linux

### LINKED ISSUE #1506: In-container CLI build fails to obtain VCS metadata (on a Linux host)
After tackling #1505, I discovered `make hack-build-cli` continues to fail due to Go failing to obtain VCS metadata within the container (on an Ubuntu host).

```console
$ make hack-build-cli
docker build -f Dockerfile.dev -t kargo:dev-tools .
[+] Building 1.6s (6/6) FINISHED                                                                                                                   docker:default
 => [internal] load build definition from Dockerfile.dev                                                                                                     0.0s
 => => transferring dockerfile: 1.40kB                                                                                                                       0.0s
 => [internal] load metadata for docker.io/library/golang:1.22.0-bookworm                                                                                    1.5s
 => [internal] load .dockerignore                                                                                                                            0.0s
 => => transferring context: 63B                                                                                                                             0.0s
 => [1/2] FROM docker.io/library/golang:1.22.0-bookworm@sha256:925fe3fa28ba428cf67a7947ae838f8a1523117b40e3e6b5106c378e3f97fa29                              0.0s
 => CACHED [2/2] RUN go install github.com/bufbuild/buf/cmd/buf@v1.26.1     && go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0     && cd  0.0s
 => exporting to image                                                                                                                                       0.0s
 => => exporting layers                                                                                                                                      0.0s
 => => writing image sha256:7f68b5f5568115748089b97b25e6267277c009b7376c56366d883efb740100fc                                                                 0.0s
 => => naming to docker.io/library/kargo:dev-tools                                                                                                           0.0s
docker run -it --rm -v gomodcache:/go/pkg/mod -v /home/h/projects/github.com/akuity/kargo/:/workspaces/kargo -v /workspaces/kargo/ui/node_modules -w /workspaces/kargo kargo:dev-tools sh -c 'GOOS=linux GOARCH=amd64 make build-cli'
CGO_ENABLED=0 go build \
        -ldflags "-w -X github.com/akuity/kargo/internal/version.version= -X github.com/akuity/kargo/internal/version.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ') -X github.com/akuity/kargo/internal/version.gitCommit= -X github.com/akuity/kargo/internal/version.gitTreeState=" \
        -o bin/kargo-linux-amd64 \
        ./cmd/cli
error obtaining VCS status: exit status 128
        Use -buildvcs=false to disable VCS stamping.
make: *** [Makefile:100: build-cli] Error 1
make: *** [Makefile:211: hack-build-cli] Error 2
```

The quick fix for this would be to add something like the following, which should be harmless for the intended usage/users of the command (contributors):

```patch
diff --git a/Makefile b/Makefile
index 71f8799..cbc36ba 100644
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,7 @@ DOCKER_CMD := $(CONTAINER_RUNTIME) run \
        -v $(dir $(realpath $(firstword $(MAKEFILE_LIST)))):/workspaces/kargo \
        -v /workspaces/kargo/ui/node_modules \
        -w /workspaces/kargo \
+       -e "GOFLAGS=-buildvcs=false" \
        kargo:dev-tools
 
 .PHONY: hack-build-dev-tools
```

However, YMMV here as a better solution could potentially be achieved if the command does work as expected on other types of machines.

hiddeco:
Fixes #1506 

This addresses an issue observed on Linux systems where files written to mount paths would end up being owned by root, or files could not be read because they were _not_ owned by the container's user.

It is achieved by creating a user within the development image with the same UID and GID as the caller on the host, while ensuring mount target paths do exist for this user and group before the container starts.

This effectively makes the `hack-build-cli` target work while also delivering smoother operations with `hack-codegen`.

In the future, we should likely rethink the way we work with tooling dependencies and their versions. As this problem was harder to tackle due to the `codegen` target performing system-global installations.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -22,3 +22,25 @@ RUN go install github.com/bufbuild/buf/cmd/buf@${BUF_VERSION} \
     && apt-get update && apt-get install nodejs -y \
     && curl -fsSL https://get.pnpm.io/install.sh | bash \
     && mv /root/.local/share/pnpm/pnpm /usr/bin
+
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+
+# To ensure mounts end up with the right permissions on Linux systems,
+# create a non-root user with the provided UID and GUID.
+RUN addgroup --gid ${GROUP_ID} user \
+    && adduser --disabled-password --gecos '' --uid ${USER_ID} --gid ${GROUP_ID} user \
+    && mkdir -p /workspaces/kargo/ui/node_modules \
+    && chown -R ${USER_ID}:${GROUP_ID} /workspaces/kargo
+
+USER user
+
+# Configure user writable Go Mod cache path and prepare it to ensure
+# it does not end up being owned by root when mounted
+ENV GOMODCACHE=/home/user/gocache
+RUN mkdir -p $GOMODCACHE
+
+# Configure user writable "global" NPM bin directory,
+# and add to path
+ENV NPM_CONFIG_PREFIX=/home/user/.npm-global
+ENV PATH=$NPM_CONFIG_PREFIX/bin:$PATH
--- a/Makefile
+++ b/Makefile
@@ -152,15 +152,22 @@ codegen:
 DOCKER_CMD := $(CONTAINER_RUNTIME) run \
 	-it \
 	--rm \
-	-v gomodcache:/go/pkg/mod \
+	-v gomodcache:/home/user/gocache \
 	-v $(dir $(realpath $(firstword $(MAKEFILE_LIST)))):/workspaces/kargo \
 	-v /workspaces/kargo/ui/node_modules \
 	-w /workspaces/kargo \
 	kargo:dev-tools
 
+DEV_TOOLS_BUILD_OPTS =
+ifeq ($(GOOS),linux)
+	DEV_TOOLS_BUILD_OPTS += --build-arg USER_ID=$(shell id -u)
+	DEV_TOOLS_BUILD_OPTS += --build-arg GROUP_ID=$(shell id -g)
+endif
+
 .PHONY: hack-build-dev-tools
 hack-build-dev-tools:
-	$(CONTAINER_RUNTIME) build -f Dockerfile.dev -t kargo:dev-tools .
+	$(CONTAINER_RUNTIME) build $(DEV_TOOLS_BUILD_OPTS) \
+ 		-f Dockerfile.dev -t kargo:dev-tools .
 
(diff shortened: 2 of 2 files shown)


Dockerfile.dev
```diff
@@ -22,3 +22,25 @@ RUN go install github.com/bufbuild/buf/cmd/buf@${BUF_VERSION} \
...
+
+ARG USER_ID
+ARG GROUP_ID
+
+# To ensure mounts end up with the right permissions on Linux systems,
+# create a non-root user with the UID and GID from the caller on the host
+RUN addgroup --gid ${GROUP_ID} user \
+    && adduser --disabled-password --gecos '' --uid ${USER_ID} --gid ${GROUP_ID} user \
+    && mkdir -p /workspaces/kargo/ui/node_modules \
+    && chown -R ${USER_ID}:${GROUP_ID} /workspaces/kargo
```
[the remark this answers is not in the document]
hiddeco: The latter part (around macOS and WSL2) typically should not be an issue, as the mount there tends to be a network drive which does not suffer from the same problems.

Given this, one option would likely be to only provide the build arguments when Linux arch is detected. While defaulting to some `1xxx` UID/GUID when none are provided.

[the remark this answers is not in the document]
krancour: sgtm

Makefile
```diff
@@ -152,15 +152,22 @@ codegen:
...
 	--rm \
-	-v gomodcache:/go/pkg/mod \
+	-v gomodcache:/home/user/gocache \
 	-v $(dir $(realpath $(firstword $(MAKEFILE_LIST)))):/workspaces/kargo \
 	-v /workspaces/kargo/ui/node_modules \
 	-w /workspaces/kargo \
 	kargo:dev-tools
 
+DEV_TOOLS_BUILD_OPTS =
+ifeq ($(GOOS),linux)
```
krancour, line 162: `GOOS` is often undefined. Would it make sense to use `uname` here instead?

hiddeco: `$(GOOS)` != `${GOOS}`, and actually is based on `uname` already. See: https://github.com/akuity/kargo/blob/2bb3b3895bfb2d0d3bf77242a5e584b05aed1c8f/Makefile#L43 😄

krancour: Apologies... I overlooked that.

krancour (approved):
🔥

Issue #1479, opened 2024-02-16
chore: hew closer to controller runtime best practices


krancour:
Strategic follow-up to #1475 

I know this is vague, but I suspect the root problem from #1475 can probably be handled through more careful adherence to controller runtime best practices. This requires a bit more research, so I am self-assigning.

krancour:
Pulling this into v0.6.0 because @hiddeco and I have agreed this is overdue and practically necessary to enable further feature development.

krancour:
@hiddeco unless you disagree, I see this issue as involving an on-going effort. I want to kick it to v0.7.0 so it doesn't have the appearance of holding up v0.6.0

krancour:
As this is just an ongoing thing, I am bumping it to v0.9.0.

krancour:
This is an on-going thing until complete. Kicking it to v1.1.0.

krancour:
Bumping to v1.6.0, but @hiddeco, maybe you can comment on how much progress we've made on this.

afaik, we're using conditions almost across the board now and most reconcilers have been almost completely re-written since this was opened.

Any opinion on outstanding things to do here? Or ought we close this?

hiddeco:
Sorry, your message got lost between notifications.

I think it is safe to close this, and I will keep an eye on the approaches we take in the future (and new things introduced upstream) so we stay aligned.

PR #1524, opened 2024-02-23
fix(controller): do not return non-0 result and err

### LINKED ISSUE #1522: warning message from controller
New warning message that's popping up in controller logs as of recently:

```
time="2024-02-22T21:20:22Z" level=info msg="Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: [https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler"](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler%22) Stage="{uat kargo-demo}" controller=stage controllerGroup=kargo.akuity.io controllerKind=Stage name=uat namespace=kargo-demo reconcileID="\"3cb63bd3-e9d3-424f-a305-2e11e7239bed\""
```

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

hiddeco:
Fixes #1522

This avoids the warning from being logged, while also moving towards more "controller-runtime best practices" as documented in #1479 (albeit not entirely related).

In general, I am of opinion that construction of the `ctrl.Result` is best done at return time to provide a clear view on what the expected requeue behavior would be without having to mentally maintain a list of conditionals which could have affected the returned result.

In case the preference would still be to apply any defaults, I think this would be better done by defining a `defer` which (conditionally) modifies the `res` and/or `err` return values based on e.g. end state of the reconciled object.

### DIFF
--- a/internal/controller/management/namespaces/namespaces.go
+++ b/internal/controller/management/namespaces/namespaces.go
@@ -6,7 +6,6 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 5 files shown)


devholic (approved):
Seems good to me 👍

krancour:
> In general, I am of opinion that construction of the ctrl.Result is best done at return time to provide a clear view on what the expected requeue behavior would be without having to mentally maintain a list of conditionals which could have affected the returned result.

Well... put that in the category of "I wish I'd thought of that first." 🤦‍♂️

PR #1534, opened 2024-02-26
fix(ui): UI does not properly handle shards deep linking

rpelczar:


hiddeco (approved):
LGTM based on https://github.com/akuity/kargo/blob/5c66b9a1288522e1b7346dfdcdcdd55239177546/api/v1alpha1/labels.go#L7

Issue #1163, opened 2023-11-22
improve cli consistency


krancour:
While working on #1162, I became more acutely aware of how internally inconsistent the CLI is in a few key regards:

* Some commands are of the form `<verb> <resource>`, while others are of the form `<resource> <verb>`

* ~~At least two commands take project as a positional argument, while most others take project as a flag.~~
    This was fixed in #1320

* There is not a consistent style for usage text, command examples, etc.

    Examples:

    ```
    kargo login -h
    Log in to a Kargo API server

    Usage:
      kargo login server-address [flags]

    ...
    ```
    
    ```
    Usage:
      kargo stage promote [flags]

    Examples:
    kargo stage promote (PROJECT) (NAME) [(--freight=)freight-id]
    ```

    ```
    Usage:
      kargo create project (NAME) [flags]

    Examples:

    # Create project
    kargo create project my-project
    ```

This issue calls for settling on and implementing a standardized style for command layout and usage/help text.

At the same time, I would like to consider the pitfalls of using multiple positional arguments. Taking `kargo stage promote (PROJECT) (NAME)` as an example, it is all too easy for a user to forget the correct order of the arguments and transpose them accidentally.

krancour:
> `kargo update freight <name> --alias=<alias>`

My only reservation is that it might not be immediately obvious that it's not for _general purpose Freight updates_. (Freight is actually immutable apart from the alias.) Having said that, it seems like a clear improvement over what's there now, so I'd say let's count it as a step in the right direction, but continue thinking about how we can improve it further.

fwiw, I don't totally love that `config set` command -- at least not in its present form -- because it's not clear what all the supported keys are. (It's only "project" that is supported, but you'd have to dig into code to know that. Bad UX.) If we modeled it after `kubectl`, we could change it to `kargo config set-project`. wdyt?

PR #1535, opened 2024-02-26
refactor(cli)!: "stage management" sub-commands

### LINKED ISSUE #1352: cli: refactor "stage management" sub-commands
Related to #1163, but specifically about two particular commands.

From #1163:

> Some commands are of the form `<verb>` `<resource>`, while others are of the form `<resource>` `<verb>`

The `stage` sub-command is one specific area where there is room for improvement.

The `stage` sub-command has the following sub-commands of its own:

* disable-auto-promotion
* enable-auto-promotion
* promote
* promote-subscribers

A major theme of the v0.4.0 release is "project management" and calls for both the creation of a `Project` CRD and additional `PromotionPolicy` options. (See [roadmap](https://kargo.akuity.io/roadmap#v040).

~~I foresee `PromotionPolicy` collapsing into the new `Project` CRD, in which case, dedicated `disable-auto-promotion` and `enable-auto-promotion` commands begin to make even less sense than they do already.~~ __This is done.__ The ability of `Freight` to be auto-promoted to any given `Stage`s in a given project to would simply be managed through edits to `Project` resources. i.e. These sub-commands will likely go away in this development cycle.

That leaves `promote` and `promote-subscribers`. There is room for two improvements here:

1. They could be collapsed into a single sub-command, which could be used as follows:

    `kargo promote --freight <freight> --stage <stage>`
    
    OR
    
    `kargo promote --freight <freight> --subscribers-of <stage>`
    
    Where precisely _one_ of `--stage` or `--subscribers-of` must be specified.

1. They can be lifted out from under the `stage` sub-command to fix a violation of our verb-first approach.

    Besides improving overall consistency, this also just makes more sense... promotions always involve a piece of Freight being promoted _to_ a Stage and it was arbitrary that these commands were buried under "stage management" when Freight is an equally important part of the promotion process.

hiddeco:
Fixes #1352

This refactors the `kargo stage promote` and `kargo stage promote-subscribers` commands into a single `kargo promote` command with `--stage` and `--subscribers-of` flags. In addition, it removes the removes the `kargo stage enable-auto-promotion` and `kargo stage disable-auto-promotion` commands, as this is now managed (declaratively) on the Project.

Effectively, this means for the promotion of a Freight to a Stage, one can now run:

```shell
kargo promote --project <project> --freight <freight-id> --stage <stage>
```

While for the promotion of a Freight to subscribers of a Stage, one must run:

```shell
kargo promote --project <project> --freight <freight-id> --subscribers-of <stage>
```


docs/docs/20-quickstart.mdx
```diff
@@ -613,9 +613,8 @@ thereby forming a _pipeline_:
 1. `uat` subscribes to `test`
 1. `prod` subscribes to `uat`.
 
-We leave it as an exercise to the reader to use the `kargo stage promote`
-command to progress the `Freight` from `stage` to `uat` and again from `uat` to
-`prod`.
+We leave it as an exercise to the reader to use the `kargo promote` command
+to progress the `Freight` from `stage` to `uat` and again from `uat` to `prod`.
```
krancour, line 617:
```suggestion
to progress the `Freight` from `test` to `uat` and again from `uat` to `prod`.
```

This wasn't your mistake, but it's a good time to fix it.

hiddeco: Ah, great 🎣. Fixed!

internal/cli/cmd/promote/promote.go
```diff
@@ -0,0 +1,129 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+
+			project := opt.Project
+			if project == "" {
+				return errors.New("project is required")
+			}
+
+			if freight == "" {
+				// TODO: Get latest available freight if empty
```
krancour, line 51: I know this comment moved here from another file, but this TODO can go away. Scenarios where we can safely infer the desired freight are few and far between. It is better that we leave the flag required.

internal/cli/cmd/promote/promote.go
```diff
@@ -0,0 +1,129 @@
...
+				return errors.New("freight is required")
+			}
+
+			if stage != "" && subscribersOf != "" {
+				return errors.New("stage and subscribers-of can not be supplied simultaneously")
```
krancour, line 56: I wonder a little if the user's mistake is more intuitively understood if we use "--stage and --subscribers-of" in the error message; i.e. with the leading "--".

Note that I don't feel strongly one way or the other. I'm only posing the question. I defer to your judgement. Of course, whichever we select, we should try to honor that convention throughout the CLI. See #1163

internal/cli/cmd/promote/promote.go
```diff
@@ -0,0 +1,129 @@
...
+				if printerErr != nil {
+					return errors.Wrap(printerErr, "new printer")
+				}
+				for _, p := range res.Msg.GetPromotions() {
+					kubeP := typesv1alpha1.FromPromotionProto(p)
+					_ = printer.PrintObj(kubeP, opt.IOStreams.Out)
+				}
+				return promoteErr
+			default:
+				return errors.New("stage or subscribers-of is required")
```
krancour, line 114: Not a hill I would die on by any means, but I think it might be easier to understand the function overall if this problem of invalid input were dealt with up front before getting into the real logic.

internal/cli/option/option.go
```diff
@@ -66,3 +66,12 @@ func MinimumNArgs(n int) cobra.PositionalArgs {
 		return nil
 	}
 }
+
+// NoArgs is a wrapper around cobra.NoArgs to additionally print usage string
```
krancour, line 70: Nice

krancour (approved):
LGTM. I don't see anything I commented on as a blocker.

internal/cli/cmd/promote/promote.go
hiddeco: I felt the same when typing this error message, but left it as an exercise for when flags in the broader sense are refactored. As I feel there is much to gain in how we currently glue them into commands and validate them. But to settle on an approach, I need a broader picture of the other commands.

The primary problem I have at present with this approach as a whole is that they are variables attached to flags, and the error message going into the details of what sets them may make it actually prone to mistakes later on...

internal/cli/cmd/promote/promote.go
hiddeco: I think (again) this would be better dealt with by finding a nicer approach to input validation, as the switch statement (if it remains) would still require an error safeguard.

internal/cli/cmd/promote/promote.go
krancour: Ok. So defer to when tackling consistency issues more broadly. sgtm.

internal/cli/cmd/promote/promote.go
krancour: Fine with that.

krancour (approved):

krancour:
This feels like a big improvement. Thank you!

PR #1539, opened 2024-02-26
fix(cli): don't involve kubeconfig in reading bytes from files

krancour:
This is like #1528, but for `main`. (#1528 was for `release-0.4`.)

This is more thoroughly tested and documented than #1528, and I have also moved the logic to a different package that seems it will be a better home for it.


internal/yaml/yaml.go
```diff
@@ -4,13 +4,63 @@ import (
...
+		}
+		var fileBytes []byte
+		if fileInfo.IsDir() {
+			var dirEntries []os.DirEntry
+			if dirEntries, err = os.ReadDir(filename); err != nil {
+				return nil, err
+			}
+			dirFiles := make([]string, 0, len(dirEntries))
+			for _, dirEntry := range dirEntries {
+				dirFiles = append(dirFiles, path.Join(filename, dirEntry.Name()))
```
hiddeco, line 39:
```suggestion
				dirFiles = append(dirFiles, filepath.Join(filename, dirEntry.Name()))
```

internal/yaml/yaml.go
```diff
@@ -4,13 +4,63 @@ import (
...
+			}
+			if fileBytes, err = Read(dirFiles); err != nil {
```
hiddeco, line 41: This infinite recursion makes me slightly uncomfortable, especially because we use `os.Stat` (which follow symlinks) while _also_ reading files endlessly without some limit.

Because of this, I think someone could create a file, and then a symlink to "jump back" to the root of the directory, causing the memory to endlessly fill up.

To guard against this, I would at least use a `bytes.Buffer` with an upper limit or maybe even ignore symlinks and/or prevent infinite recursion by other means.

krancour: I overlooked the possibility of symlinks. 🤦‍♂️

krancour: I fixed this. We follow symlinks, but ignore any symlink that evaluates to a path that contains itself.

There is still no explicit limit on how much we'll read, but reading infinitely is no longer a possibility.

internal/yaml/yaml_test.go
```diff
@@ -1,13 +1,185 @@
...
+	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
 )
 
+func TestRead(t *testing.T) {
+	abDir, err := os.MkdirTemp("", "ab")
```
hiddeco, line 15: Nit: with the number of test cases, it may be appealing to put them in individual descriptive `t.Run` cases to more easily spot what breaks.

hiddeco (approved):

internal/yaml/yaml_test.go
krancour: I typically do for table-driven tests. I didn't in this particular case since the setup of each case builds on the setup of the previous case.

I realized only after you said something that I could leave setup _outside_ of `t.Run` and  wrap `t.Run` only around groups of assertions and that wouldn't prevent iteratively building the more complex cases.

All set and thank you for a good idea.

hiddeco: Only spotted this now 🤦 

But if you use `t.TempDir()`, you can get away with not doing any `defer` calls to clean things up _and_ the clean up will still succeed if the test case fails, instead of it accumulating directories in temp.

krancour: Forgot about that!

Also found another edge case involving symlinks... switching this to draft for now.

krancour: Ok... switched to `t.TempDir()`. Thanks again for the reminder.

Two edge cases I found:

1. If we end up reading a symlink that points to a directory _containing_ what we're reading, the existing logic for preventing infinite recursion will work, but duplicates may be read in the process.

2. If we deliberately did `yaml.Read("foo", "foo")`, we'll get all the resources twice.

Both problems can be solved by keeping track of paths we've read and short-circuiting reads of things already read.

krancour: These two new edge cases are addressed in 26d3e66cbae702a8cbb58748b357e6d2786dd022

hiddeco (approved):

PR #1540, opened 2024-02-26
Fix garbage-collector affinity indentation

sthomson-wyn:
Garbage collector affinities are under indented, resulting in invalid yaml


krancour:
Looks to have been superseded by #1541 👍

PR #1542, opened 2024-02-27
refactor(crds)!: consistently use message field instead of error field

krancour:
@jessesuen and I agreed quite some time ago that status subresources would use `Message` fields instead of `Error` fields to preserve the potential for communicating things that are not strictly errors.

We have, since then, consistently used `Message` in _new_ status subresource types, but Stages and Warehouses were still using `Error`.

This was a low-hanging consistency improvement that I wanted to make before breaking changes become more difficult.

cc @rbreeze or @rpelczar, although there are open issues for surfacing info from these fields, I can not find any place in the UI where the field names that are changing are currently referenced. You might want to double-check me.

There are no other changes in this PR.


hiddeco (approved):
Without taking the details of the UI into account, this looks good to me.

💯 for consistency.

PR #1547, opened 2024-02-27
refactor(cli): restructure commands, option flags and validation

hiddeco:
Touches into #1163

This pull request achieves multiple things:

1. Commands no longer make use of lengthy in-line `func`s to perform an action. Instead, they are built around "option" `struct`s with methods to add flags, parse command arguments, validate, and run the actual command logic. By doing this, we introduce a separation of concerns while also allowing things to (hypothetically) be tested without relying on further Cobra functionalities.
2. Flags are now registered as required, mutually exclusive, taking file and/or directory inputs, etc. using Cobra features. Which improves out-of-the-box validation, while also enriching the shell completion instructions. For more information around this, see e.g. https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagRequired

#### Example boilerplate new command

```go
struct fooOptions {
	SomeArg  string
	SomeFlag string
}

func NewCommand() *cobra.Command {
	cmdOpts := &fooOptions{}

	cmd := &cobra.Command{
		Use:   "foo (ARG) --some-flag=value",
		Args:   option.ExactArgs(1),
		Short: "Perform a demo operation",
		// Other Cobra fields omitted for brevity...
		RunE: func(cmd *cobra.Command, args []string) error {
			cmdOpts.complete(args)

			if err := cmdOpts.validate(); err != nil {
				return err
			}

			return cmdOpts.run()
		},
	}

	// Register the option flags on the command.
	cmdOpts.addFlags(cmd)

	return cmd
}

func (o *fooOpts) addFlags(cmd *cobra.Command) {
	cmd.Flags().StringVar(&o.SomeFlag, "some-flag", "", "An example flag which requires a value")

	if err := cmd.MarkFlagRequired("some-flag"); err != nil {
		panic(errors.Wrap(err, "could not mark some-flag flag as required"))
	}
}

func (o *fooOpts) complete(args []string) {
	o.SomeArg = strings.TrimSpace(args[0])
}

func (o *fooOpts) validate() error {
	var errs []error

	if o.SomeArg = "" {
		errs = append(errs, errors.New("arg is required"))
	}

	if o.SomeFlag == "" {
		errs = append(errs, errors.New("some-flag is required"))
	}

	return errors.Join(err...)
}

func (o *fooOpts) run() error {
	// Do something intelligent with arg and flag...

	return nil
}
```

### DIFF
--- a/internal/cli/cmd/apply/apply.go
+++ b/internal/cli/cmd/apply/apply.go
@@ -1,6 +1,7 @@
 package apply
 
 import (
+	"context"
 	goerrors "errors"
 	"fmt"
 
@@ -22,8 +23,37 @@ import (
 
 type applyOptions struct {
 	*option.Option
+	Config config.CLIConfig
 
 	Filenames []string
+}
+
+func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
+	cmdOpts := &applyOptions{
+		Option: opt,
+		Config: cfg,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "apply [--project=project] -f (FILENAME)",
+		Short: "Apply a resource from a file or from stdin",
+		Example: `
+# Apply a stage using the data in stage.yaml
+kargo apply -f stage.yaml
+`,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if err := cmdOpts.validate(); err != nil {
+				return err
+			}
+
+			return cmdOpts.run(cmd.Context())
+		},
+	}
+
+	// Register the option flags on the command.
+	cmdOpts.addFlags(cmd)
+
+	return cmd
 }
 
 // addFlags adds the flags for the apply options to the provided command.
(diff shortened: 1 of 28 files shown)


internal/cli/cmd/apply/apply.go
```diff
@@ -19,8 +19,54 @@ import (
...
+func newApplyOptions(opt *option.Option) *applyOptions {
+	return &applyOptions{Option: opt}
+}
+
+// AddFlags adds the flags for the apply options to the provided command.
+func (o *applyOptions) AddFlags(cmd *cobra.Command) {
+	o.PrintFlags.AddFlags(cmd)
+
+	// TODO: Factor out server flags to a higher level (root?) as they are
+	//   common to almost all commands.
```
krancour, line 38: Ya... it's tricky. _Almost but not all_ commands use them.

Personally, I'd rather repeat ourselves sometimes that have inapplicable flags show up on commands. i.e. If we had to choose, UX wins over DRY, imho.

internal/cli/cmd/apply/apply.go
```diff
@@ -19,8 +19,54 @@ import (
...
 )
 
+type applyOptions struct {
+	*option.Option
+
+	Filenames []string
+}
+
+// newApplyOptions creates a new options struct.
+func newApplyOptions(opt *option.Option) *applyOptions {
```
krancour, line 29: Nit: I'm a big fan of constructor-like functions if they're either exported for use outside the package _or_ are used internally, but are still performing non-trivial initialization. Since this one is neither of those scenarios and is also called only from one place, I'd just as soon kill it. _But it is not a hill I would die on._

internal/cli/cmd/apply/apply.go
```diff
@@ -19,8 +19,54 @@ import (
...
+	return &applyOptions{Option: opt}
+}
+
+// AddFlags adds the flags for the apply options to the provided command.
+func (o *applyOptions) AddFlags(cmd *cobra.Command) {
```
krancour, line 34: I'm not dead set against this, but I am curious about the rationale for extracting this logic from `NewCommand()` into its own function.

I ask because I think the code is _slightly_ easier to follow the closer we keep all the flag setup to the actual definition of the command... _but_ I suspect you had reasons for this.

internal/cli/cmd/apply/apply.go
```diff
@@ -19,8 +19,54 @@ import (
...
+	o.PrintFlags.AddFlags(cmd)
+
+	// TODO: Factor out server flags to a higher level (root?) as they are
+	//   common to almost all commands.
+	option.InsecureTLS(cmd.PersistentFlags(), o.Option)
+	option.LocalServer(cmd.PersistentFlags(), o.Option)
+
+	option.Filenames(cmd.Flags(), &o.Filenames, "Filename or directory to use to apply the resource(s)")
+
+	if err := cmd.MarkFlagRequired(option.FilenameFlag); err != nil {
```
krancour, line 44: These are new since the last time I went deep on Cobra. I am super excited about how much this cleans things up.

internal/cli/cmd/apply/apply.go
hiddeco: A problem I have observed though is that we sometimes ourselves forget to actually wire them in (see e.g. https://github.com/akuity/kargo/blob/5912e5a9dfd2912fdd6a46e1b35655b6225c4454/internal/cli/cmd/approve/approve.go in current `main`, which AFAIK should have the flags set).

By registering them once at root and loading them into some config which is passed to downward commands, this could be avoided.

internal/cli/cmd/apply/apply.go
krancour: Nit: It could be unexported.

internal/cli/cmd/apply/apply.go
```diff
@@ -19,8 +19,54 @@ import (
...
+		panic(errors.Wrap(err, "could not mark filename flag as filename"))
+	}
+	if err := cmd.MarkFlagDirname(option.FilenameFlag); err != nil {
+		panic(errors.Wrap(err, "could not mark filename flag as dirname"))
+	}
+}
+
+// Validate performs validation of the options. If the options are invalid, an
+// error is returned.
+func (o *applyOptions) Validate() error {
```
krancour, line 57: Kudos for this...

I've frequently seen this pattern of using a struct to encapsulate options and have rarely seen much value in it...

Coupled with the decision to extract validation into its own function, that pattern begins to make sense.

You've converted me.

Nit: Could be unexported.

krancour:
@hiddeco overall, I love the approach. I commented on "apply" with some nits, questions, and kudos. I know this is only a draft right now, but I see nothing blocking so far.

internal/cli/cmd/apply/apply.go
krancour: Is there a way to do it at root and _unregister_ them in spots?

hiddeco: I think you can theoretically use `MarkHidden` to hide them from the help menu for specific commands.

internal/cli/cmd/apply/apply.go
hiddeco: The goal is to move into a direction in which the actual logic of running the command is driven by just the options struct, making it easier to test things in isolation (without requiring a full end-to-end setup).

The construction of the command then just becomes gluing various building blocks together, in which case I think a clear separation of concerns becomes desirable.

I also think it helps "drive by contributors" to know where to put what, while providing guardrails to keep the code structured and tidy. If they want to add some functionality based on a value, they now have clear steps to follow:

1. Add the field to the struct.
2. Register the flag in the designated area.
3. Implement logic which consumes the value.

Lastly, it makes options "reusable". In case you have hypothetical options which are e.g. a subset of other options for multiple commands. You can simply embed the options struct and register the flags in multiple commands simultaneously.

krancour: > The goal is to move into a direction in which the actual logic of running the command is driven by just the options struct, making it easier to test things in isolation (without requiring a full end-to-end setup).

Agree and wish we'd done things this from the start.

> gluing various building blocks together

Agree... with the hope of keeping the glue easy to follow, which was why I raised this question. My only worry had been that breaking flag registration out into its own function undermined keeping the glue in one place. In my head, something like this might tick all your boxes and mine:

```golang
func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
	cmdOpts := &applyOptions{Option: opt}

	cmd := &cobra.Command{
		Use:   "apply [--project=project] -f (FILENAME)",
		Short: "Apply a resource from a file or from stdin",
		PreRunE: func(cmd *cobra.Command, _ []string) error {
			return cmdOpts.Validate()
		},
		RunE: func(cmd *cobra.Command, _ []string) error {
			// Run should be easy to test without involving Cobra.
			return run(cmd.Context(), cmdOpts)
		},
	}

	opt.PrintFlags.AddFlags(cmd)
	// etc...
	if err := cmd.MarkFlagDirname(option.FilenameFlag); err != nil {
		panic(errors.Wrap(err, "could not mark filename flag as dirname"))
	}

	return cmd
}
```

You can look at this one `NewCommand` func now for quite a concise overview of the whole command. Admittedly, having the logic in-line as we previously did stood in the way of that even more than breaking flag registration into its own function ever did. Extracting that logic makes things look less insane.

wdyt?

hiddeco: Not strongly opposed to it, although I think that with a bigger number of flags and/or markers you can end up in the same position as the in-line run. An example:

```go
func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
	cmdOpts := newPromotionOptions(opt)

	cmd := &cobra.Command{
		Use:   "promote [--project=project] --freight=freight-id [--stage=stage] [--subscribers-of=stage]",
		Short: "Manage the promotion of freight",
		Args:  option.NoArgs,
		Example: `
# Promote a freight to a stage for a specific project
kargo promote --project=my-project --freight=abc123 --stage=dev

# Promote a freight to subscribers of a stage for a specific project
kargo promote --project=my-project --freight=abc123 --subscribers-of=dev

# Promote a freight to a stage for the default project
kargo config set project my-project
kargo promote --freight=abc123 --stage=dev

# Promote a freight to subscribers of a stage for the default project
kargo config set project my-project
kargo promote --freight=abc123 --subscribers-of=dev
`,
		PreRunE: func(cmd *cobra.Command, args []string) error {
			return cmdOpts.Validate()
		},
		RunE: func(cmd *cobra.Command, args []string) error {
			return run(cmd.Context(), cmdOpts)
		},
	}

	// This is where the messier part starts.
	o.PrintFlags.AddFlags(cmd)

	option.InsecureTLS(cmd.PersistentFlags(), o.Option)
	option.LocalServer(cmd.PersistentFlags(), o.Option)

	option.Project(cmd.Flags(), &o.Project, o.Project,
		"The Project the Freight belongs to. If not set, the default project will be used.")
	option.Freight(cmd.Flags(), &o.Freight, "The ID of the Freight to promote.")
	option.Stage(cmd.Flags(), &o.Stage, fmt.Sprintf("The Stage to promote the Freight to. If set, --%s "+
		"must not be set.", option.SubscribersOfFlag))
	option.SubscribersOf(cmd.Flags(), &o.SubscribersOf, fmt.Sprintf("The Stage from which the subscribers "+
		"will be used to promote the Freight to. If set, --%s must not be set.", option.StageFlag))

	if err := cmd.MarkFlagRequired(option.FreightFlag); err != nil {
		panic(errors.Wrap(err, "could not mark freight flag as required"))
	}
	cmd.MarkFlagsOneRequired(option.StageFlag, option.SubscribersOfFlag)
	cmd.MarkFlagsMutuallyExclusive(option.StageFlag, option.SubscribersOfFlag)

	return cmd
```

To me, the above looks like something which I still may need to break up into smaller pieces to improve readability, which I think the `AddFlags` method does in the right way by encapsulating the registration with the options themselves (just like `Validate`).

But as said, I do not have extremely strong feelings about this and would be happy to remove and/or circle back to it once we both observe of it becoming a problem.

krancour: I see your point. There may be no perfect solution here and I don't want to let perfect be the enemy of good. I am also not strongly attached to my opinion. I think stick with what you originally proposed and we'll just keep an eye out for opportunities to improve this further as this evolves.

internal/cli/cmd/apply/apply.go
hiddeco: I actually agree, and this was mainly added to wrap my head around factoring things out at the beginning. Have removed them.

internal/cli/cmd/apply/apply.go
krancour: Putting some closure on this. We spoke offline and explored some tweaks to the pattern that look like they're just what we're looking for.

krancour (approved):
I have not reviewed every single line, but reviewed enough of the commands to assert that the new pattern that's emerging is a major improvement.

My one nit is that `run()` as a func of an "options" type seems slightly counter-intuitive. I like an example you showed me yesterday where e.g. instead of `applyOptions`, it may have been `applyRun`. Or even something like `applyCmdImpl` might be nice.

This is not in any way a blocker or a hill that I would die on.

hiddeco:
> My one nit is that `run()` as a func of an "options" type seems slightly counter-intuitive. [..]

When I started implementing that, I ran into some issues:

- `fooRun#run` felt odd ("run" twice)
- `fooCmd#run` felt better, but would conflict with e.g. `cmd := &cobra.Command{}`

Given it "runs" things with the options, I then felt that this convention was not that bad and stopped looking for alternatives.

PR #1548, opened 2024-02-27
feat: manage mappings for pre-defined service accounts

### LINKED ISSUE #1368: Predefined ServiceAccounts to improve SSO/RBAC UX
# Proposed Feature

One of the first things people should do with their Kargo instance is to configure SSO and stop using the built-in admin user. It will be common for orgs to have SSO user(s) that is equivalent to the admin user. Today, this process is:

1. GLOBAL_SERVICE_ACCOUNT_NAMESPACES: kargo
2. create a `kargo-admin` ServiceAccount
3. create `kargo-admin` ClusterRoleBinding to `kargo-api` ClusterRole
4. annotate kargo-admin ServiceAccount with a SSO annotations

I feel that Kargo Helm chart should have a convenience option to do 1-3 out of the box to reduce the upfront steps and have a standard ServiceAccount that we could refer to in docs.

In addition, we could introduce a `kargo-viewer` built-in ServiceAccount and ClusterRole/ClusterRoleBinding, that is ready to be annotated for giving readonly access to SSO users.

# Motivation

Getting started will be easier.

# Suggested Implementation

Simply vend these resources in our Helm chart, with a helm values option for additional annotations.

krancour:
Fixes #1368

Adds two new `ServiceAccount` resources (`kargo-admin` and `kargo-viewer`) if the API server and OIDC are both enabled. (Without both of those enabled, it seems we would not need these.) How these are mapped to claims like `sub`, `email`, or `groups` can be specified at the time of Kargo installation.

Each of these new `ServiceAccount`s are bound to appropriate cluster-level permissions. For `kargo-admin`, those permissions are a near-complete subset of the API server's own permissions. (Access to create `SubjectAccessReviews` is not needed.) For `kargo-viewer`, those permissions are an even narrower subset of the API server's permissions.

The `kargo-admin` SA does not get any permissions the API server itself does not also get. This includes _not_ having access to `Secret` resources cluster-wide. Since #1486, we dynamically expand the API server's permissions to include access to `Project`-specific `Secret`s as `Project`s are created. This PR uses the same mechanism to dynamically expand the `kargo-admin` `ServiceAccount`'s permissions as well.

cc @gdsoumya


jessesuen:
I see that these will hardwire to the install namespace (.Release.Namespace).

There was an idea (perhaps even proposed by you) where these ServiceAccounts did not necessarily have to reside in the kargo installation namespace. The idea being `kargo` install namespace would remain purely for the kargo workloads, and configuration could be managed elsewhere.

WDYT about making this configurable? e.g. I'd like to install these in like a `kargo-global` namespace.

krancour:
When I proposed that, my intent was to avoid people messing with the kargo namespace. Or to at least avoid normalizing that. It bothers me less here because it's the chart managing the SA and not a person.

That said, the decision to put these SAs in the kargo ns here had more to do with needing to put them in a ns that I know for sure will exist.

It's unconventional, but I guess I could make the chart manage a second namespace. I'll tinker with that tomorrow.

jessesuen:
Offline discussion:

Because the controller has the logic to dynamically expand permissions for project-specific secrets, I think it makes sense to treat these ServiceAccounts as special, always existing. So I've reversed my thoughts on this and think that `kargo-admin` and `kargo-viewer `should be special, well-known ServiceAccounts that people can bind SSO claims to as an operator-level decision, and reside in the kargo install namespace.

jessesuen (approved):

PR #1552, opened 2024-02-29
refactor(controller)!: improved cred storage

### LINKED ISSUE #1546: proposal: update credential storage strategy
Argo CD has always used the label `argocd.argoproj.io/secret-type` with value `repository` or `repo-creds` on Secrets to indicate that they are either:

1. Credentials for one specific repository (`repository`)
1. Credentials multiple repositories, all starting with the same URL prefix (`repo-creds`)

The label name changed, but Kargo copied this strategy almost exactly. Even at the time I did that, I felt this was confusing and have since found that others are also confused by it.

Confusion stems from:

* _Both_ label values indicate a Secret represents credentials, but one has "creds" in its name and the other doesn't. 🤔 
* One label values indicates a Secret represents something akin to a "credential template," but nothing in the label value hints at such behavior.
* Apart from these labels, a Secret's `data.type` field is used to with a value like `git`, `helm`, or `image` to denote what kind of repository the credentials are for. Overuse of "type" contributes to confusion.

I'm sure Argo CD has rational reasons for why it is this way, but seeing the confusion it can cause at times, and given that credential management features (#1523, #1444, #1531) are currently high priorities, I think we're at the right juncture to ask whether we were wrong to blindly copy the Argo CD approach and whether we want to make a small course correction.

Note that #1483 removed the option to "borrow" credentials from Argo CD. Doing so may have eliminated the only remaining incentive for emulating Argo CD's credential strategy, oddities and all.

What I propose is:

1. Eliminating the `kargo.akuity.io/secret-type` and its two values.
2. Begin using Secret's _own_ [type field](https://kubernetes.io/docs/concepts/configuration/secret/#secret-types) with one of the following values to indicate the type of repository the credentials are for:
    * `kargo.akuity.io/git-creds`
    * `kargo.akuity.io/image-creds`
    * `kargo.akuity.io/chart-creds`
3. Support _two_ fields for specifying repository / repositories the credentials are for:
    * `repoURL` -- a single specific repository
    * `repoURLPattern` -- a regular expression identifying repositories the credentials are for

Implementation-wise, none of this is very complicated.

@jessesuen wdyt?

krancour:
Fixes #1546 

@jessesuen I promised I'd look specifically for your buy-in, but I think none of what's in this PR should be a surprise based on our offline conversation earlier today.

Best way to review the new strategy -- which diverges from Argo CD's only in small, but meaningful ways -- might be to look at the [doc preview](https://deploy-preview-1552.kargo.akuity.io/how-to-guides/managing-credentials).

@hiddeco I'd value your feedback as well.

@wanghong230 fyi... I saw you found dealing with credential secrets a little clunky the other day. The old strategy was copied from Argo CD, but @jessesuen and I agreed there was no reason we _had_ to copy it, so hopefully these changes address some of the frustrations you experienced.

### DIFF
--- a/docs/docs/30-how-to-guides/20-managing-credentials.md
+++ b/docs/docs/30-how-to-guides/20-managing-credentials.md
@@ -41,7 +41,7 @@ The `name` of such a `Secret` is inconsequential and may follow any convention
 preferred by the user.
 
 :::info
-Kargo uses Kubernetes `Namespace`s to demarcate project boundaries. `Secret`s
+Kargo uses Kubernetes `Namespace`s to mark project boundaries. `Secret`s
 representing credentials will typically exist in the same `Namespace` as the
 `Stage` resources that will require them. There are exceptions to this, which
 are covered in the next section.
--- a/docs/docs/30-how-to-guides/20-managing-credentials.md
+++ b/docs/docs/30-how-to-guides/20-managing-credentials.md
@@ -5,71 +5,87 @@ sidebar_label: Managing credentials
 
 # Managing Credentials
 
-To manage the progression of changes from stage to stage, Kargo will
-often require read/write permissions on private GitOps repositories and
-read-only permissions on private container image and/or Helm chart repositories.
+To manage the progression of freight from stage to stage, Kargo will often
+require read/write permissions on private GitOps repositories and read-only
+permissions on private container image and/or Helm chart repositories.
 
 This section presents an overview of how these credentials can be managed.
 
 ## Credentials as Kubernetes `Secret` Resources
 
-Kargo borrows its general credential-management approach from Argo CD, meaning
-that credentials are stored as Kubernetes `Secret` resources containing
-specially-formatted data. These secrets generally take the following form:
+:::caution
+Kargo formerly borrowed its general credential-management approach from Argo CD,
+but has since diverged.
+:::
+
+Kargo expects any credentials it requires to have been stored as specially
+labeled Kubernetes `Secret` resources containing specially-formatted data. These
+`Secret`s take the following form:
 
 ```yaml
 apiVersion: v1
 kind: Secret
 metadata:
   name: <name>
-  namespace: <namespace>
+  namespace: <project namespace>
   labels:
-    kargo.akuity.io/secret-type: <secret type>
+    kargo.akuity.io/cred-type: <cred type>
 stringData:
-  type: <type>
-  url: <repo url>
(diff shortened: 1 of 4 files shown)


jessesuen (approved):

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -5,71 +5,87 @@ sidebar_label: Managing credentials
...
 preferred by the user.
 
 :::info
-Kargo uses Kubernetes `Namespace`s to demarcate project boundaries, so related
-`Stage` resources always share a `Namespace`.
-
-Secrets representing credentials will typically exist in the same `Namespace` as
-the `Stage` resources that will require them. There are exceptions to this rule,
-which are covered in the next section.
+Kargo uses Kubernetes `Namespace`s to demarcate project boundaries. `Secret`s
```
hiddeco, line 44: Nit: when writing documentation, it is often best to use simple English (as much as possible) to help non-native speakers grasp things easily.

I think "demarcate" may be a word some people will have a harder time to understand, and I would suggest to change it to something like "define" or "mark off".

krancour:
@hiddeco 

> you often want to disable caching these object kinds, as you tend to be interested in only a small subset while production size clusters have a lot of them

I've had thoughts along those lines before, but rather than disable the cache entirely for secrets, I had wanted to research whether the underlying informer can be configured to use a label selector so only secrets we care about are ever cached.

docs/docs/30-how-to-guides/20-managing-credentials.md
krancour: Agree, and I'm usually more mindful of that and this is a place where I probably overestimated familiarity with that word. Happy to fix this.

krancour, line 44:
```suggestion
Kargo uses Kubernetes `Namespace`s to mark project boundaries. `Secret`s
```

krancour: Fixed in 08c28a8895c3c3af034e9b6eb125241bc255add4

hiddeco:
@krancour that's actually an option since not too long ago, see https://github.com/kubernetes-sigs/controller-runtime/blob/7ad4b4c318f8628bcab1c2b436e86f9dbd91d852/designs/cache_options.md#example-usages

krancour:
Awesome!

@hiddeco any chance you want to follow up on that?

PR #1557, opened 2024-02-29
feat(controller): watch only Secrets w/ Kargo labels

### LINKED ISSUE #1554: Explore caching Secrets with specific labels only
A production Kubernetes cluster can potentially have thousands of Secrets, which at present are all cached due to default controller-runtime behavior.

As we are typically only interested in a subset of Secrets with specific labels (see #1546 and #1552), it is worth exploring if we can make use of the [controller-runtime cache configuration options](https://github.com/kubernetes-sigs/controller-runtime/blob/7ad4b4c318f8628bcab1c2b436e86f9dbd91d852/designs/cache_options.md#example-usages) to limit the caching to Secrets that have Kargo related labels attached.

hiddeco:
Fixes #1554 

This configures the Kargo manager client to only watch Secrets with a `kargo.akuity.io/cred-type` label set to a recognized type.

When testing this with 100 randomly generated Secrets in the `default` namespace using:

```shell
for i in {1..100}; do head -c 500000 /dev/urandom | base64 | kubectl create secret generic random-$i --from-file=my-file=/dev/stdin; done
```

This makes the memory consumption as reported by kube-metrics go from 199Mi before the patch to 14Mi afterwards.

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 	"github.com/akuity/kargo/internal/git"
 	"github.com/akuity/kargo/internal/logging"
 )
@@ -31,10 +32,6 @@ const (
 	TypeHelm Type = "helm"
 	// TypeImage represents credentials for an image repository.
 	TypeImage Type = "image"
-
-	// CredentialTypeLabelKey is the key for a label used to identify the type
-	// of credentials stored in a Secret.
-	CredentialTypeLabelKey = "kargo.akuity.io/cred-type" // nolint: gosec
 )
 
 // Credentials generically represents any type of repository credential.
@@ -164,7 +161,7 @@ func (k *kubernetesDatabase) getCredentialsSecret(
 		&client.ListOptions{
 			Namespace: namespace,
 			LabelSelector: labels.Set(map[string]string{
-				CredentialTypeLabelKey: credType.String(),
+				kargoapi.CredentialTypeLabelKey: credType.String(),
 			}).AsSelector(),
 		},
 	); err != nil {
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -19,6 +19,11 @@ import (
 // Type is a string type used to represent a type of Credentials.
 type Type string
 
+// String returns the string representation of a Type.
+func (t Type) String() string {
+	return string(t)
+}
+
 const (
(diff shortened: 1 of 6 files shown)


krancour:
I'm obviously in favor of this.

It's also making me anxious about something else... The number of Stages, Freight, and Promotions in a cluster are likely to to make numbers like _100_ look tiny. Presumably the cache is going to get us into quite a bit of trouble with those as well?

hiddeco:
> The number of Stages, Freight, and Promotions in a cluster are likely to to make numbers like 100 look tiny. Presumably the cache is going to get us into quite a bit of trouble with those as well?

I would be much less worried about this. The Secrets I am creating in the example above are exceptionally large (on purpose) to quickly see the effect it would have on the memory usage. While typically some of these Secrets exist on clusters (e.g. people putting binaries in them, Helm history, etc.), this is an aggravated example.

For our object kinds, while there may be many — I do not expect them to ever reach this size. Therefore (and based on experience with building other controllers), I do not think this would become a problem anytime soon.

hiddeco:
> For our object kinds, while there may be many — I do not expect them to ever reach this size. Therefore (and based on experience with building other controllers), I do not think this would become a problem anytime soon.

Further tested this by creating 100 projects, with each containing one Warehouse (spawning one Freight). Total memory consumption peaked at ~6.4MiB above previous baseline, dropping again to ~4.1MiB above previous baseline once it had reconciled everything.

I did not check if this included Go's garbage collector kicking in, but I think it underlines that we shouldn't immediately worry.

internal/credentials/credentials.go
```diff
@@ -27,9 +32,9 @@ const (
 	// TypeImage represents credentials for an image repository.
 	TypeImage Type = "image"
 
-	// credentialTypeLabelKey is the key for a label used to identify the type
+	// CredentialTypeLabelKey is the key for a label used to identify the type
 	// of credentials stored in a Secret.
-	credentialTypeLabelKey = "kargo.akuity.io/cred-type" // nolint: gosec
+	CredentialTypeLabelKey = "kargo.akuity.io/cred-type" // nolint: gosec
```
hiddeco, line 37: Wondering if this actually shouldn't be moved into the API package.

krancour: I was initially keeping it close to the _one_ place it was used.

Now that it's called for in more than one place, I agree.

I _also_ have a need to move it in a PR I am working on.

krancour:
@hiddeco 

> The Secrets I am creating in the example above are exceptionally large

I overlooked `500000` 😌

internal/credentials/credentials.go
hiddeco: Moved it in 0606a09

krancour (approved):

PR #1559, opened 2024-03-01
chore(cli): update all usages and examples

### LINKED ISSUE #1163: improve cli consistency

hiddeco:
Fixes #1163 

This updates all command usages, examples, etc. to use the same syntax, which matches the syntax used for `kubectl` commands.

In addition, where applicable various variations of command options have been documented in places where a command works with a series of options.

### DIFF
--- a/internal/cli/cmd/apply/apply.go
+++ b/internal/cli/cmd/apply/apply.go
@@ -35,11 +35,15 @@ func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:   "apply [--project=project] -f (FILENAME)",
+		Use:   "apply -f FILENAME",
 		Short: "Apply a resource from a file or from stdin",
+		Args:  option.NoArgs,
 		Example: `
 # Apply a stage using the data in stage.yaml
 kargo apply -f stage.yaml
+
+# Apply the YAML resources in the stages directory
+kargo apply -f stages/
 `,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			if err := cmdOpts.validate(); err != nil {
--- a/internal/cli/cmd/approve/approve.go
+++ b/internal/cli/cmd/approve/approve.go
@@ -33,10 +33,10 @@ func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
 		Short: "Manually approve freight for promotion to a stage",
 		Args:  option.NoArgs,
 		Example: `
-# Approve a freight in a project to the QA stage
+# Approve a piece of freight for the QA stage
 kargo approve --project=my-project --freight=abc1234 --stage=qa
 
-# Approve a freight in the default project to the QA stage
+# Approve a piece of freight for the QA stage in the default project
 kargo config set-project my-project
 kargo approve --freight=abc1234 --stage=qa
 `,
--- a/internal/cli/cmd/approve/approve.go
+++ b/internal/cli/cmd/approve/approve.go
@@ -29,9 +29,17 @@ func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:     "approve --project=project --freight=freight --stage=stage",
-		Short:   "Manually approve freight for promotion to a stage",
-		Example: "kargo approve --project=project --freight=abc1234 --stage=qa",
+		Use:   "approve [--project=project] --freight=freight --stage=stage",
+		Short: "Manually approve freight for promotion to a stage",
+		Args:  option.NoArgs,
+		Example: `
+# Approve a freight in a project to the QA stage
+kargo approve --project=my-project --freight=abc1234 --stage=qa
+
+# Approve a freight in the default project to the QA stage
+kargo config set-project my-project
+kargo approve --freight=abc1234 --stage=qa
+`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			if err := cmdOpts.validate(); err != nil {
 				return err
--- a/internal/cli/cmd/dashboard/dashboard.go
+++ b/internal/cli/cmd/dashboard/dashboard.go
@@ -21,7 +21,7 @@ func NewCommand(cfg config.CLIConfig) *cobra.Command {
 		Short: "Open the Kargo Dashboard in your default browser",
 		Args:  option.NoArgs,
 		Example: `
-# Open the Kargo Dashboard
+# Open the Kargo Dashboard in the browser
 kargo dashboard
 `,
 		RunE: func(*cobra.Command, []string) error {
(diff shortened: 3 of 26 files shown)


internal/cli/cmd/apply/apply.go
```diff
@@ -35,11 +35,15 @@ func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:   "apply [--project=project] -f (FILENAME)",
+		Use:   "apply -f FILENAME",
```
krancour, line 38: No opinion here -- more of a question. Would it generally be more clear to use long flags in the usage strings?

hiddeco: Think it depends on the flag type, for `--project` I think it's a good choice. However, for files `-f` is so common that I think people can figure this out based on the context itself.

internal/cli/cmd/approve/approve.go
```diff
@@ -29,9 +29,17 @@ func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
...
 
 	cmd := &cobra.Command{
-		Use:     "approve --project=project --freight=freight --stage=stage",
-		Short:   "Manually approve freight for promotion to a stage",
-		Example: "kargo approve --project=project --freight=abc1234 --stage=qa",
+		Use:   "approve [--project=project] --freight=freight --stage=stage",
+		Short: "Manually approve freight for promotion to a stage",
+		Args:  option.NoArgs,
+		Example: `
+# Approve a freight in a project to the QA stage
```
krancour, line 36: Tricky grammar thing that I find myself stumbling over frequently.

"Freight" is a [mass noun](https://www.grammar-monster.com/glossary/mass_noun.htm) and inherently represents an unspecified volume of -- well, freight.

Grammatically, "a freight" is incorrect. So is "freights." This is pretty esoteric, especially for ESL speakers, so I try not to be dogmatic about this. If it's grammatically incorrect in code or a comment, I don't think it matters. For user-facing things like docs and CLI help, I think we need to exercise more than the usual amount of care.

The tricks I have found help in those scenarios are:

1. Drop the indefinite article "a"
2. Preface "freight" with "a piece of"

So, personally, I would rephrase this either of the following:

* Approve freight in a project for the QA stage
* Approve a piece of freight in a project for the QA stage

Separately, with nearly everything in Kargo being project-scoped, I wonder if we may wish to avoid using the phrase "in a project" unless we truly need to clarify. (e.g. Credentials can exist in a project or globally, but that's an exception.)

So in the end, maybe something like this?

* Approve freight for the QA stage
* Approve a piece of freight for the QA stage

This same comment probably applies in other places.

Also, please, please forgive me if this seems overly pedantic. That's not my intention at all. I think with the task at hand being very much about improving consistency throughout the CLI, these are the exact sort of things we need to hammer out and then whatever we decide, stick with it.

internal/cli/cmd/dashboard/dashboard.go
```diff
@@ -16,9 +17,13 @@ func NewCommand(cfg config.CLIConfig) *cobra.Command {
...
-		Use:     "dashboard",
-		Short:   "Open the Kargo Dashboard in your default browser.",
-		Example: "kargo logout",
+		Use:   "dashboard",
+		Short: "Open the Kargo Dashboard in your default browser",
+		Args:  option.NoArgs,
+		Example: `
+# Open the Kargo Dashboard
+kargo dashboard
+`,
```
krancour, line 26: Do you think we should specify that users will have to log in to the dashboard separately from the UI?

I'm torn. I feel like it's like "Caution. Coffee may be hot." Like I think that _shouldn't_ be that surprising, but I can somehow still imagine someone opening an issue about it.

internal/cli/cmd/approve/approve.go
hiddeco: > Also, please, please forgive me if this seems overly pedantic.

I actually really like you thinking this through, so I appreciate the suggestions.

---

Within the context of promotion, I think:

> Approve a piece of freight in a project for the QA stage

would fit best because it clarifies it being a singular unit.

I also agree with your suggestion around dropping "a project".

internal/cli/cmd/apply/apply.go
krancour: I can live with that.

krancour:
Apart from the esoteric grammar things, this is perfect.

internal/cli/cmd/dashboard/dashboard.go
hiddeco: Reading this again, I was personally more thinking it would require "[...] in the default browser" as a comment 😅 

I think users should not be surprised about having to login again.

krancour: I was already torn. I'm fine with whatever side of the fence you landed on.

internal/cli/cmd/approve/approve.go
```diff
@@ -33,10 +33,10 @@ func NewCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Command {
 		Short: "Manually approve freight for promotion to a stage",
 		Args:  option.NoArgs,
 		Example: `
-# Approve a freight in a project to the QA stage
+# Approve a piece of freight for the QA stage
 kargo approve --project=my-project --freight=abc1234 --stage=qa
 
-# Approve a freight in the default project to the QA stage
+# Approve a piece of freight for the QA stage in the default project
```
krancour, line 39: In many cases, I am wondering about the necessity of the second example for the default project.

I have no strong opinion about it, but think it's something that is worth some consideration.

Any thoughts?

krancour (approved):

internal/cli/cmd/approve/approve.go
hiddeco: Well, personally I only found out about this setting once I had seen it in one of the commands where it was documented. So I figured I would add it everywhere for now, until we have extensive documentation in other places.

PR #1561, opened 2024-03-02
refactor(api, cli)!: improved efficiency of cli get commands and robust support for aliases

### LINKED ISSUE #1098: improve network efficiency of cli commands

### LINKED ISSUE #1560: freight-oriented cli commands and api endpoints should more consistently support aliases
Some do. Some don't. All should.

krancour:
Fixes #1098
Fixes #1560

## Overview

I was adding new API endpoints to help solve #1098 and among these was one for get Freight. There has been a known, but undocumented issue that not all Freight-related endpoints support aliases as an alternative to their name -- which is a not-so-human-friendly sha. I ended up opening #1560. I implemented the new get endpoint with proper support for that, then decided I'd better fix the others while at it...

## Detailed Break Down

* Added protobuf defs for new endpoints. Re-arranged the protobuf defs slightly so that endpoint names and messages are a little more alphabetically ordered. Just trying to impose some additional order on this file. More remains to be done. I tried not to get carried away and limit the change to existing endpoints "near" the new ones I was working on.

* As I went to implement the new endpoints, I found existing validation functions very awkward to work with. Some were very hyper-specific like "validate Project and Stage not empty," or "validate Project and Warehouse name not empty," etc. I generalized these to "validate field not empty," which is less specific, but therefore more reusable.

* Updated existing endpoints to use the refactored validators. There are no functional changes to the existing endpoints.

* Added the new endpoints I set out to add -- get Project and get Freight. (Part of #1098)

* Made corresponding CLI sub-command improvements to leverage the new endpoints and solve . (Part of #1098)

* Redefined existing protobuf defs for existing Freight-related API endpoints. (Part of #1560)

* Refactored existing Freight-related API endpoints to use aliases. (Part of #1560)

* Made corresponding CLI sub-command improvements to leverage alias support in the Freight-related API endpoints. (Part of #1560)

* Finally, all changes to generated code are isolated to their own commit. This is accounts for most of the changes in this PR.


internal/api/delete_freight_v1alpha1.go
```diff
@@ -2,39 +2,58 @@ package api
 
 import (
 	"context"
+	"fmt"
 
 	"connectrpc.com/connect"
 	"github.com/pkg/errors"
```
hiddeco, line 8: Curiosity question: what's the reason `github.com/pkg/errors` is extensively used throughout the code base while it has been marked as deprecated / archived, and Go's own error package has been quite enriched in more recent versions?

internal/cli/cmd/get/promotions.go
```diff
@@ -40,21 +39,28 @@ func newGetPromotionsCommand(cfg config.CLIConfig, opt *option.Option) *cobra.Co
...
 		Short:   "Display one or many promotions",
 		Example: `
-# List all promotions in the project
+# List all promotions in my-project
 kargo get promotions --project=my-project
 
-# List all promotions in JSON output format
+# List all promotions in my-project in JSON output format
 kargo get promotions --project=my-project -o json
+# List all promotions for the QA stage in my-project
```
hiddeco, line 47:
```suggestion

# List all promotions for the QA stage in my-project
```

internal/api/delete_freight_v1alpha1.go
krancour: Habit? I think some of the big improvements came in in 1.2 iirc and we haven't explored them adequately.

The subject of it being deprecated was discussed once before and judged a non-issue at the time given the relative simplicity of the package.

I'm unopposed to moving off of it as long as Go can wrap and unwind errors now.

btw... off topic... working on this PR, I think there is a lot more that can be done to improve how API requests are validated, but this PR is big enough already. Let's keep it in mind for another area for improvement when time permits. I'll open an issue when I'm not out of pocket.

hiddeco: Wrapping is now possible using `fmt.Errorf("more context: %w"`, and the appropriate internal things are available to unwrap them. I will create an issue for it on Monday.

I will explore the API validation, it's an area I haven't got into yet although I noticed that _some_ errors were both caught by the CLI and the API while exploring other things.

krancour: At one time, I used `fmt.Errorf` a lot. It used to not do anything special to make the error unwrappable. Just dug into source code and found this:

```golang
// If the format specifier includes a %w verb with an error operand,
// the returned error will implement an Unwrap method returning the operand.
// If there is more than one %w verb, the returned error will implement an
// Unwrap method returning a []error containing all the %w operands in the
// order they appear in the arguments.
```

Blame says special support for %w was added five years ago. Support for multiple %w was added 2 years ago. I missed the memo because I had an established habit of using `github.com/pkg/errors` by then.

Totally in favor of the switch when time permits.

internal/api/promote_stage_v1alpha1.go
```diff
@@ -12,24 +13,41 @@ import (
...
+	if err := validateFieldNotEmpty("name", stageName); err != nil {
+		return nil, err
+	}
+
+	freightName := req.Msg.GetFreight()
+	freightAlias := req.Msg.GetFreightAlias()
+	if (freightName == "" && freightAlias == "") || (freightName != "" && freightAlias != "") {
+		return nil, connect.NewError(
+			connect.CodeInvalidArgument,
+			errors.New("exactly one of freightName or freightAlias should not be empty"),
```
hiddeco, line 37: Nit: I think the "freight name" is called "freight" within `PromoteStageRequest`? Given this, I would make the error message

> exactly one of freight or freightAlias should not be empty

to correctly refer to the field.

hiddeco (approved):

internal/api/promote_stage_v1alpha1.go
krancour: Nice catch. I believe this accidentally snuck in because this was all extracted from another branch with more extensive changes. Fixing.

krancour: Fixed in 8f3aea757c1b0ab5531df769ebb5d77eb136b1d2

PR #1566, opened 2024-03-04
feat(api,cli): add cred mgmt api end endpoints and corresponding cli sub-commands

### LINKED ISSUE #1444: UI for managing credentials
# Proposed Feature

Kargo needs credentials for things like:
* git repositories
* container repositories

Since they are Secrets, it becomes tricky to manage via GitOps. The UI should provide an imperative way to:
1. input credentials into a project namespace
2. view what credentials are available (not the actual sensitive content) and for which repositories

# Motivation

Although managing credentials declaratively with things like external-secrets would be preferred, having an imperative way to do this would help with people getting started with kargo, developer self-service, etc...

Plus, having visibility in the UI to what credentials are available would be useful in and of itself.

# Suggested Implementation

In the project view, have a credentials tab where this could be inputted and listed, but without exposing/returning any sensitive content.

### LINKED ISSUE #1531: CLI for managing credentials
# Proposed Feature

This is the CLI equivalent to https://github.com/akuity/kargo/issues/1444

# Motivation

For all the same reasons in https://github.com/akuity/kargo/issues/1444

Currently, inputting a credential is a bit painful. They can do this today, but it requires knowledge of the Secret format. e.g.:

```yaml
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: github-creds
  namespace: kargo-simple
  labels:
    kargo.akuity.io/secret-type: repository
stringData:
  type: git
  url: https://github.com/akuity/kargo-simple.git
  username: username
  password: github_pat_REPLACEME
```

A convenience CLI option would be useful

```
kargo credentials add \
    --project my-project \
    --type repository \
    --url https://github.com/akuity/kargo-simple.git \
    --username jessesuen
```

It should then prompt from stdin, or accept pipes.

krancour:
Fixes #1444 
Fixes #1531

Credentials are unique among everything Kargo deals with because they are really Secrets in disguise.

After this PR, credentials can still be managed using properly labeled Secret manifests and the `create -f`, `apply -f`, and `delete -f` sub-commands, but new options will be available as well:

* `get credentials ...`: Lists credentials if none are specified. Otherwise, fetches individual credentials. The values returned are _Secrets_ with decoded `stringData` and the value of the `password` field therein redacted server side.

* `create credentials ...`: Single arg used for credential name; flags used for other attributes. Importantly, using this to create credentials, unlike the existing options, does not require one to know the proper formatting of the underlying Secret's payload. Password can be supplied interactively instead of via flag for those who do not want passwords in their shell history.

* `update credentials ...`: Works the same as the above. 

* `delete credentials ...`: Multiple args specify creds to delete.


internal/api/get_credentials_v1alpha1.go
```diff
@@ -0,0 +1,63 @@
...
+	}), nil
+}
+
+func redactPassword(secret corev1.Secret) corev1.Secret {
+	secret.StringData = make(map[string]string, len(secret.Data))
+	for k, v := range secret.Data {
+		if k == "password" {
+			secret.StringData[k] = "*** REDACTED ***"
+		} else {
+			secret.StringData[k] = string(v)
```
hiddeco, line 58: Assuming someone for some reason decides to mix their Kargo credentials together with something else, this would leak the _other_ (non Kargo) secrets. Given this, I am wondering if we shouldn't work with a whitelist instead.

krancour: Nice catch. imho, I should be checking for the existence of the credentials label here and treating the credential as not found if a secret is found lacking that label.

wdyt?

krancour: Oh... also note that users cannot retrieve secrets this way that they aren't already authorized to get.

krancour: Fixed in https://github.com/akuity/kargo/pull/1566/commits/c8b95af925db036dc83823b6aa0df182db391a5a

hiddeco: With this patch, I think it's still odd (even with authorization checks) that we handle the password field with a mask while allowing other values to be plain text. I would either expect it to not have any masks, or to mask any fields except the ones which we expect to contain non-sensible values (e.g. `repoURL`, `username`, etc.)

krancour: We expect Secrets that are credentials to have a payload that's structured a certain way. At present `password` is the only field we _expect_ that is sensitive.

But I'm starting to see your point, because suppose I had the following:

```
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: kargo-demo
  namespace: kargo-demo
  labels:
stringData:
  username: krancour
  passwrd: '12345'
```

__Notice that I fat-fingered the key for the password field.__

And created it like so:

```
kargo apply -f creds.yaml
```

Then `kargo get credentials -p kargo-demo kargo-demo` will show me the "passwd" unredacted.

So your point behind white-listing fields _not_ to redact is more clear to me now and I agree. Sorry it took me two tries to see the value in that.

I suppose it is still not _entirely_ foolproof, as nothing stops someone from, for instance, accidentally putting the password in the `username` field. (Whoops!) But I'm not sure there's anything that can be done about mistakes of that nature.

krancour: Fixed in f4a8ecf54ab1c9c5fd9d3becd6df49d40e22ebaf

internal/cli/cmd/create/credentials.go
```diff
@@ -0,0 +1,201 @@
...
+		return errors.Wrap(err, "get client from config")
+	}
+
+	var repoType string
+	if o.Git {
+		repoType = "git"
+	} else if o.Helm {
+		repoType = "helm"
+	} else if o.Image {
+		repoType = "image"
```
hiddeco, line 180: Any reason for not using the exported types from the `credentials` package instead of redefining strings?

internal/api/get_credentials_v1alpha1.go
```diff
@@ -0,0 +1,88 @@
...
+		switch k {
+		case "repoURL", "repoURLPattern", "username":
+			secret.StringData[k] = string(v)
+		default:
+			secret.StringData[k] = "*** REDACTED ***"
+		}
+	}
+	secret.Data = nil
+	return secret
+}
```
jessesuen, line 88: There was one corner case that bit us in Argo CD. When the Secret was applied via Kubectl directly to Kube API server, we forgot that the sensitive data was additionally included in the `last-applied-configuration` annotation. And so the secret data could leak through that annotation and added logic to strip that specific annotation.

Next, another user reported another config management tool (I forget which), had the equivalent of `last-applied-configuration` (which Argo CD did not know about), and so we had to remove that.

That said, labels, given their character constraints, will likely never be used the same way as annotations and contain sensitive data.

krancour: > `last-applied-configuration`

Nice though @jessesuen 

I can blank that out.

krancour: > another config management tool (I forget which), had the equivalent of last-applied-configuration

There may not be much reason to not just blank out all annotations when retrieving creds in this manner.

krancour: @jessesuen see 56be70b48e97f619c2182785c5834d68bed5514b

hiddeco: Is there user value in redacting the annotation field values? Or would simply stripping all annotations be a simpler and less noisy approach.

hiddeco: (Another config management tool having the equivalent of last-applied-configuration is KPT Config Sync by the way, using `configmanagement.gke.io/declared-config`).

krancour: But this gets me thinking now about what is the _right_ way for someone to update creds.

The counterpart to `kargo get creds` is `kargo update creds`. The update endpoint carries out an update operation... I should change it to a patch.

Changing this to draft for a bit...

krancour: Updated in 47be6461fa8a0e451d2b9f1bbf4f03baa68fc7e9

It's still an update (not a patch), but it now makes specific changes to a secret instead of overwriting it wholesale.

internal/api/get_credentials_v1alpha1.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// If this isn't labeled as repository credentials, return not found.
+	var isCredentials bool
+	if secret.Labels != nil {
+		_, isCredentials = secret.Labels[kargoapi.CredentialTypeLabelKey]
+	}
+	if !isCredentials {
+		return nil, connect.NewError(
+			connect.CodeNotFound,
+			errors.Errorf("secret %q exists, but is not labeled as credentials", name),
```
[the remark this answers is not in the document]
krancour: I think that's a great idea.

internal/cli/cmd/create/credentials.go
krancour: Fixed in e729af4fee2c4c64342a15513b1e5650aa847909

internal/cli/cmd/create/credentials.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	kargoSvcCli, err := client.GetClientFromConfig(ctx, o.Config, o.Option)
+	if err != nil {
+		return errors.Wrap(err, "get client from config")
+	}
+
+	var repoType string
+	if o.Git {
+		repoType = string(credentials.TypeGit)
```
hiddeco, line 179: We are starting to cast this to a string in so many places now that I wonder if it wouldn't be beneficial to simply add a `String()` method to the type.

internal/cli/cmd/delete/credentials.go
```diff
@@ -0,0 +1,124 @@
...
+	cfg config.CLIConfig,
+	opt *option.Option,
+) *cobra.Command {
+	cmdOpts := &deleteCredentialsOptions{
+		Option: opt,
+		Config: cfg,
+	}
+
+	cmd := &cobra.Command{
+		Use:     "credentials [--project=project] [NAME ...]",
```
hiddeco, line 36:
```suggestion
		Use:     "credentials [--project=project] (NAME ...)",
```

(We require _at least_ one)

internal/cli/cmd/create/credentials.go
```diff
@@ -0,0 +1,204 @@
...
+				RepoUrlPattern: o.RepoURLPattern,
+				Username:       o.Username,
+				Password:       o.Password,
+			},
+		),
+	); err != nil {
+		return errors.Wrap(err, "create credentials")
+	}
+
+	return nil
```
hiddeco, line 203: We should probably print here what we created, see the project command for further inspiration on how to do this with recent changes made to `main`:

https://github.com/akuity/kargo/blob/34bcb82cb0696c3b390e3321f786ff21c2702071/internal/cli/cmd/create/project.go#L128-L132

internal/cli/cmd/delete/credentials.go
```diff
@@ -0,0 +1,124 @@
...
+				&v1alpha1.DeleteCredentialsRequest{
+					Project: o.Project,
+					Name:    name,
+				},
+			),
+		); err != nil {
+			resErr = goerrors.Join(resErr, errors.Wrap(err, "Error"))
+			continue
+		}
+		_, _ = fmt.Fprintf(o.IOStreams.Out, "Credentials Deleted: %q\n", name)
```
hiddeco, line 121: With recent changes to `main`, the way we print this has changed. See this command for further inspiration:

https://github.com/akuity/kargo/blob/34bcb82cb0696c3b390e3321f786ff21c2702071/internal/cli/cmd/delete/project.go#L106-L118

internal/cli/option/flag.go
```diff
@@ -98,11 +142,25 @@ func OldAlias(fs *pflag.FlagSet, stage *string, usage string) {
 	fs.StringVar(stage, OldAliasFlag, "", usage)
 }
 
+func Password(fs *pflag.FlagSet, password *string, usage string) {
```
hiddeco, line 145:
```suggestion
// Password adds the PasswordFlag to the provided flag set.
func Password(fs *pflag.FlagSet, password *string, usage string) {
```

internal/cli/option/flag.go
```diff
@@ -70,10 +95,29 @@ func FreightAlias(fs *pflag.FlagSet, stage *string, usage string) {
...
+// Image adds the ImageFlag to the provided flag set.
+func Image(fs *pflag.FlagSet, image *bool, usage string) {
+	fs.BoolVar(image, ImageFlag, false, usage)
+}
+
 func InsecureTLS(fs *pflag.FlagSet, opt *Option) {
 	fs.BoolVar(&opt.InsecureTLS, "insecure-skip-tls-verify", false, "Skip TLS certificate verification")
 }
 
+func InteractivePassword(fs *pflag.FlagSet, changePasswordInteractively *bool, usage string) {
```
hiddeco, line 117:
```suggestion
// InteractivePassword adds the InteractivePasswordFlag to the provided flag set.
func InteractivePassword(fs *pflag.FlagSet, changePasswordInteractively *bool, usage string) {
```

hiddeco (commented):
Tested this extensively, and it works really well! I have made a couple of suggestions based on recent changes and other observations after reading the code thoroughly.

I also caused some trouble with #1574 (sorry!), but it should be straight forward by looking at other existing commands to add the required changes.

internal/cli/cmd/create/credentials.go
```diff
@@ -0,0 +1,204 @@
...
+	opt *option.Option,
+) *cobra.Command {
+	cmdOpts := &createCredentialsOptions{
+		Option: opt,
+		Config: cfg,
+	}
+
+	cmd := &cobra.Command{
+		Use: `credentials [--project=project] NAME \
+    (--git | --helm | --image) \
```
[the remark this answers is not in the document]
krancour: `--type=<type>` was my first instinct, but I figured three separate flags that don't require values were going to work better with completions.

I don't foresee the number of credential types growing dramatically, but if it grows by even one, then there could be a benefit to this being dynamic -- otherwise one ends up in a position where they _need_ the latest client in order to create credentials of a newer type.

Right now, I feel inclined to keep `--git`, `--helm`, and `--image`, but _also_ add `--type=<type>` and use "one of" semantics for the four of those flags.

internal/cli/cmd/create/credentials.go
```diff
@@ -0,0 +1,204 @@
...
+    --repo-url=repo-url | --repo-url-pattern=repo-url-pattern) \
+    -username=username \
+    [--password=password]`,
```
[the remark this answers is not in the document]
krancour: Eagle eye. Thank you!

internal/cli/cmd/create/credentials.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// run creates the credentials in the project based on the options.
+func (o *createCredentialsOptions) run(ctx context.Context) error {
+	for {
+		if o.Password != "" {
+			break
+		}
+		prompt := &survey.Password{
+			Message: "Repository password",
```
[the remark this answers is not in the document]
krancour: So far, yes. If in the future we encounter a scenario where credentials might be for something other than a repo, then the `repo-url` and `repo-url-pattern` fields are a bigger issue than this is.

internal/cli/cmd/create/credentials.go
krancour: I can get down with that.

internal/cli/cmd/delete/credentials.go
krancour: I probably copied that from "get" and overlooked. Thanks again for the eagle eye!

internal/cli/cmd/create/credentials.go
krancour: Yeah... I realized this while in the midst of merging the latest from main into here. I plan to.

internal/cli/cmd/create/credentials.go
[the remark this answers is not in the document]
krancour: Fixed in a051595b97541cab94bb7e8f3be72f480a7f8022

internal/cli/option/flag.go
krancour: Fixed in a051595b97541cab94bb7e8f3be72f480a7f8022

internal/cli/option/flag.go
krancour: Fixed in a051595b97541cab94bb7e8f3be72f480a7f8022

internal/cli/cmd/create/credentials.go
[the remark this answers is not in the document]
krancour: Fixed in b01861180331708aac4eb438cc602c85e4f0cd73

internal/cli/cmd/create/credentials.go
krancour: Fixed in a few places in 4985b45b7500b87ec57414a54fbcac1ab13469c0, however calling `.String()` doesn't work when defining a constant.

i.e. `const GitFlag = credentials.TypeGit.String()` doesn't work.

Choices are leave it alone or make affected flags vars instead of consts.

I opted for leaving it alone.

internal/cli/cmd/create/credentials.go
krancour: I've added this in e7fc174a7c92447e1095d8d8878c603ac0bfe3b5 but it's still done the old way...

I'll merge shortly and will make sure everything is consistent with your recent changes.

krancour: Fixed in c73f394aca28048395ae334e36b655467ba105a2

internal/cli/cmd/delete/credentials.go
krancour: Fixed in c73f394aca28048395ae334e36b655467ba105a2

krancour:
@hiddeco c73f394aca28048395ae334e36b655467ba105a2 should bring these commands up to date with other recent CLI improvements.

internal/cli/cmd/delete/credentials.go
```diff
@@ -0,0 +1,146 @@
...
+	// Set the input/output streams for the command.
+	io.SetIOStreams(cmd, cmdOpts.IOStreams)
+
+	return cmd
+}
+
+// addFlags adds the flags for the get credentials options to the provided
+// command.
+func (o *deleteCredentialsOptions) addFlags(cmd *cobra.Command) {
+	o.PrintFlags.AddFlags(cmd)
```
hiddeco, line 80:
```suggestion
	o.ClientOptions.AddFlags(cmd.PersistentFlags()))
	o.PrintFlags.AddFlags(cmd)
```

internal/cli/cmd/get/credentials.go
```diff
@@ -0,0 +1,190 @@
...
+	// Set the input/output streams for the command.
+	io.SetIOStreams(cmd, cmdOpts.IOStreams)
+
+	return cmd
+}
+
+// addFlags adds the flags for the get credentials options to the provided
+// command.
+func (o *getCredentialsOptions) addFlags(cmd *cobra.Command) {
+	o.PrintFlags.AddFlags(cmd)
```
hiddeco, line 88:
```suggestion
	o.ClientOptions.AddFlags(cmd.PersistentFlags())
	o.PrintFlags.AddFlags(cmd)
```

internal/cli/cmd/create/credentials.go
```diff
@@ -0,0 +1,221 @@
...
+	// Set the input/output streams for the command.
+	io.SetIOStreams(cmd, cmdOpts.IOStreams)
+
+	return cmd
+}
+
+// addFlags adds the flags for the get credentials options to the provided
+// command.
+func (o *createCredentialsOptions) addFlags(cmd *cobra.Command) {
+	o.PrintFlags.AddFlags(cmd)
```
hiddeco, line 115:
```suggestion
	o.ClientOptions.AddFlags(cmd.PersistentFlags())
	o.PrintFlags.AddFlags(cmd)

```

internal/cli/cmd/update/credentials.go
```diff
@@ -0,0 +1,202 @@
...
+	// Set the input/output streams for the command.
+	io.SetIOStreams(cmd, cmdOpts.IOStreams)
+
+	return cmd
+}
+
+// addFlags adds the flags for the get credentials options to the provided
+// command.
+func (o *updateCredentialsOptions) addFlags(cmd *cobra.Command) {
+	o.PrintFlags.AddFlags(cmd)
```
hiddeco, line 103:
```suggestion
	o.ClientOptions.AddFlags(cmd.PersistentFlags())
	o.PrintFlags.AddFlags(cmd)
```

hiddeco (approved):
Besides a tiny nit on client option flag registration, this looks good to me. 💯

internal/cli/cmd/update/credentials.go
krancour: Seems I overlooked these when bringing the new cmds up to date with the pattern. Will for sure fix these before merging.

PR #1570, opened 2024-03-04
custom authz decisions on approvals and promotions

### LINKED ISSUE #1222: Add RBAC for manually approving Freight to a Stage
# Proposed Feature

We introduced the ability to approve freight for a certain stage. This needs to be protected by RBAC. 

# Motivation

Manual approving freight should be considered a privilege similar/or same as to promoting.

# Suggested Implementation

In our RBAC webhook check, we should perform the same check we do for promote against a freight, making sure user has the `promote` privilege for stage, if we notice that the user is modifying a Freight for a certain stage.

### LINKED ISSUE #1567: bug: ability to promote not enforced properly when using API server
We have long performed a subject access review in a validating webhook when a new Promotion is created.

It asserts that the user has the custom "promote" verb on the Stage they are attempting to promote _to_.

This works fine for anyone using `kubectl` to create Promotion resources directly, but if you're using the API server, it doesn't work correctly...

The reason is that the API server will assert an SSO user can create Promotions before using the API server's own service account to do so. The "user" that the webhook server sees is the API server -- _and it has broad permissions to promote._

Similar to #1222, this is a very rare case where we need to build some additional authz logic directly into an API server endpoint.

I will probably address both this issue and #1222 in tandem.

krancour:
Fixes #1222 
Fixes #1567

### DIFF
--- a/internal/api/approve_freight_v1alpha1.go
+++ b/internal/api/approve_freight_v1alpha1.go
@@ -6,7 +6,8 @@ import (
 
 	"connectrpc.com/connect"
 	"github.com/pkg/errors"
-	kubeerr "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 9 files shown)


jessesuen (approved):
Nice work!

hiddeco (approved):

Issue #1569, opened 2024-03-04
Value of local API server feature in CLI


hiddeco:
At present, the `kargo` CLI supports a "local API server" mode for certain commands when `--local-server` is provided. By passing this flag, a local instance of the API server is temporarily spawned using the KubeConfig on the local system.

Since #1353, the registration of the `--local-server` flag has already been moved to commands where it makes sense instead of it being a global persistent flag. However, the setup for the server has remained at root level — making it a slightly awkward feature as the consumption happens at a higher level than the registration.

While this could potentially be refactored to at least abstract the local server away, and only spawn it in places where it is strictly required. Doing this in a nice manner, while dealing with the temporary listener and configuration of the address would continue to be challenge. In addition, one could arguably do about anything the API server is capable of by just using `kubectl` to either get, list or modify resources.

Given the above, I am wondering what the use case was/is for this feature, and if this still applies today and outweighs the technical challenges it introduces?

cc: @jessesuen

krancour:
I've always questioned the value of this local API server option...

Specifically, I question how often it will be that someone _isn't_ running the API server in the Kargo control plane. If someone isn't doing so, they're also not using the UI. I feel someone running Kargo in this manner already intends to just do everything via `kubectl`.

I am +1 on dropping this, but want to hear other perspectives.

I agree that it's worth asserting how valuable this is or isn't before putting any effort into cleaning up how it works.

jessesuen:
The value of the local API server was motivated by the following use case where:
* Kargo API server is **_not_** exposed in a shared manner
* End users are given access to a portion of the cluster via kubeconfig
* Users could still have a UI to do their thing

And so the local kargo dashboard option was intended to enable a cluster tenant user, who has read/write permissions to just a portion of the cluster (one or more namespaces), to run a local UI so they have an interface for managing their kargo environments. They could also avoid involve another team (e.g. cluster operator) who might control ingress to the cluster.

This design and access model is most analogous to the (now defunct) https://octant.dev/ project. Both Argo CD and Workflows have a similar local UI capability inspired by Octant.

That said, I'm not quite sure how popular these features are, and I'm okay eliminating the feature if it's causing us consternation.

But would that mean we build `kargo` CLI separate from `kargo-api` binaries?

krancour:
@jessesuen the use case you described makes sense, but cannot be done with the feature as currently implemented.

At present, adding `--local-server` to a command launches an ephemeral server that lives for the life of the command it's carrying out.

[...]

PR #1574, opened 2024-03-05
refactor(cli): remove `--local-server` flag and untangle root cmd logic

hiddeco:
- Reduce the typical parameters required to construct a (sub-)command to `config.CLIConfig` and `genericiooptions.IOStreams`.
- Properly implement `genericiooptions.IOStreams` in all places, by removing calls to e.g. `fmt.Println` while also wiring it into the `cobra.Command` using the available setters.
- Better utilization of `genericclioptions.PrintFlags` and the available default "name" printer when no output is configured and the command is printing information about operations performed on/for Kargo objects. This effectively changes output like:

  ```console
  $ kargo apply -f kargo          
  Stage Created: "kargo-advanced/ab-test-b"
  Stage Created: "kargo-advanced/prod-east"
  Project Updated: "/kargo-advanced"
  AnalysisTemplate Updated: "kargo-advanced/test"
  Stage Updated: "kargo-advanced/dev"
  Stage Updated: "kargo-advanced/staging"
  Stage Updated: "kargo-advanced/ab-test-a"
  Stage Updated: "kargo-advanced/prod"
  Stage Updated: "kargo-advanced/prod-west"
  Stage Updated: "kargo-advanced/prod-central"
  Warehouse Updated: "kargo-advanced/guestbook"
  ```

  to

  ```console
  $ kargo apply -f kargo                                   
  stage.kargo.akuity.io/ab-test-b created
  stage.kargo.akuity.io/prod-east created
  project.kargo.akuity.io/kargo-advanced updated
  analysistemplate.argoproj.io/test updated
  stage.kargo.akuity.io/dev updated
  stage.kargo.akuity.io/staging updated
  stage.kargo.akuity.io/ab-test-a updated
  stage.kargo.akuity.io/prod updated
  stage.kargo.akuity.io/prod-west updated
  stage.kargo.akuity.io/prod-central updated
  warehouse.kargo.akuity.io/guestbook updated
  ```
- Based on the discussion in #1569: removal of `--local-server` flag (which based on and the further discovery of https://github.com/akuity/kargo/issues/1569#issuecomment-1980596897, appears to have been broken for quite some time).


internal/cli/cmd/create/create.go
```diff
@@ -57,21 +61,22 @@ kargo create project my-project
 	// Register the option flags on the command.
 	cmdOpts.addFlags(cmd)
 
+	// Set the input/output streams for the command.
+	cmd.SetIn(cmdOpts.IOStreams.In)
+	cmd.SetOut(cmdOpts.IOStreams.Out)
+	cmd.SetErr(cmdOpts.IOStreams.ErrOut)
```
krancour, line 67: Just an observation: This seems to be universal for all commands.

internal/cli/util/util.go
```diff
@@ -0,0 +1,13 @@
+package util
```
krancour, line 1: Nit: "util" packages are something I try to avoid because they have a tendency to become "junk drawers" for every little thing that we can't decide where else to put it.

Note: We already have a few such packages that are "util" packages in all but name and long-term, I'd like to untangle those.

I'd rather see this in a package like `cli/io` or something, because that name will make it less of a magnet for homeless code.

Better yet...

I recognize this `cli/util` package only even exists because of the need to share certain functionality across all commands, which are _each_ in their own package...

I would propose that the packages under `internal/cli` are mostly vanity packages (_and_ micro-packages) and we would do well to collapse them into a single package, in which case, shared functionality like `SetIOStreams()` no longer requires its own package to be shared across all the commands.

I'll still say this is a "nit" because I wouldn't let it block this PR, and I think any package re-structuring could happen in a follow-up PR instead of piggy-backing on this one.

cmd/cli/root.go
```diff
@@ -30,99 +21,44 @@ import (
...
-						LocalMode: true,
-					},
-					client,
-					client,
-				)
-				go srv.Serve(ctx, l) // nolint: errcheck
-				opt.LocalServerAddress = fmt.Sprintf("http://%s", l.Addr())
-			}
-			return nil
-		},
```
krancour, line 79: Not a blocker... more of a what if...

Knowing that much of this will shortly re-surface under a new `kargo server` command, what would you think about moving this code under a new, _hidden_ command and we'll work work on that as time permits and unhide it when it's ready?

Dealer's choice.

krancour (approved):
Had some discussion points, but nothing is blocking.

LGTM!

internal/cli/util/util.go
hiddeco: > Nit: "util" packages are something I try to avoid because they have a tendency to become "junk drawers" for every little thing that we can't decide where else to put it.

Yes, I agree and your `io` suggestion is actually a better way of dealing with this. The primary reason I put it into a separate place was because I didn't want to continue to pollute the `option` package which historically appeared to have been used as a "there is no better place for this" dump yard.

> [..] we would do well to collapse them into a single package [..]

The fear I have with this is that all sub-packages together may be a bit too much to continue to offer a clear separation of concerns, which could potentially open us up a temptation to not factor things out at all but rather move things into private functions. Which long-term is not a great solution either.

Given this, I will rename it to `io` before merge and let the idea simmer for a bit.

cmd/cli/root.go
hiddeco: This is actually a great suggestion to keep the code around, even if it would abort as soon as the command is run.

I will implement this with a TODO note with a rough outline of what the future should bring.

internal/cli/util/util.go
krancour: > option package which historically appeared to have been used as a "there is no better place for this" dump yard.

Yep. One example of existing packages that are "util" in all but name.

> The fear I have with this is that all sub-packages together may be a bit too much to continue to offer a clear separation of concerns, which could potentially open us up a temptation to not factor things out at all but rather move things into private functions

I agree there is some logical limit to how much one lets a package sprawl before deciding it _should_ be broken up. That limit is for sure subjective. imho, the number of commands we're dealing with doesn't approach that limit.

As for common things being factored out into private functions (or types) -- I don't think that's _inherently_ terrible. 🤷‍♂️ I think there are a lot of concerns that are exclusive to a _class_ of problem -- e.g. the CLI options are exclusively a concern for CLI commands -- and keeping things of that nature close to the one place they're used feels reasonable. It doesn't preclude using other approaches for things deemed more broadly reusable.

> and let the idea simmer for a bit

💯

There's room for more discussion perhaps, and at any rate, I see no reason to rush into that sort of decision.

PR #1577, opened 2024-03-05
fix(verification): inherit AnalysisTemplate args

### LINKED ISSUE #1573: Default args not carried over from AnalysisTemplate when creating AnalysisRuns
# Description

When Kargo creates an AnalysisRun from an AnalysisTemplate, if the Stage does not "pass" args, then the "default" args as defined in the template will not be used.

e.g., if this is my template:

```yaml
kind: AnalysisTemplate
metadata:
  name: test
spec:
  args:
  - name: foo
    value: bar
  metrics:
  - name: test
    provider:
      job:
        spec:
```

And my Stage does not "pass" args, then the AnalysisRun that is created will not have args. e.g.:

```yaml
kind: AnalysisRun
metadata:
  name: dev.01hr6e2hap3be7mfks5qy04jcj.9ed4919
spec:
  metrics:
  - name: test
    provider:
      job:
        spec:
```

In Argo Rollouts, the arg values in the AnalysisTemplate are supposed to act as a "default," and used if the caller does not supply them. See docs:

https://argo-rollouts.readthedocs.io/en/stable/features/analysis/#analysis-template-arguments

# Version

v0.4.1

hiddeco:
Fixes #1573

The main cause appears to have been a mix-up in order of parameters.

### DIFF
--- a/internal/controller/stages/verification.go
+++ b/internal/controller/stages/verification.go
@@ -262,7 +262,7 @@ func (r *reconciler) buildAnalysisRun(
 			Value: &arg.Value,
 		}
 	}
-	mergedArgs, err := mergeArgs(template.Spec.Args, rolloutsArgs)
+	mergedArgs, err := mergeArgs(rolloutsArgs, template.Spec.Args)
 	if err != nil {
 		return nil, errors.Errorf("error merging arguments")
 	}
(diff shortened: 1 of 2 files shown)


jessesuen (approved):
TY!

Issue #1578, opened 2024-03-05
Add the ability to set insecureSkipTLSVerify for Images in a Warehouse


christianh814:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

In a `Warehouse`, users can set `insecureSkipTLSVerify` for Git repos subscriptions. The same option should also be available for image repos, which there [is no option currently](https://doc.crds.dev/github.com/akuity/kargo/kargo.akuity.io/Warehouse/v1alpha1@v0.4.2#spec-subscriptions-image)

# Motivation

Internal repos often times have self signed certificates

# Suggested Implementation

Similar to the Git repo's implementation of the `insecureSkipTLSVerify` subscription.

PR #1591, opened 2024-03-08
feat(controller): allow skipping TLS verification for Image repositories

hiddeco:
Fixes: #1578

This adds a `.insecureSkipTLSVerify` field to `Image` repository objects, allowing the user to disable the verification of TLS certificates in the same manner as Git.

In addition, it solves an issue where Go's HTTP default transport was being used for things which should run in a contained manner. This could _theoretically_ result in security related issues due to a malicious dependency having the option to overwrite the global configuration, and would typically have been marked as a low undefined behavior finding during a security audit ([I have been there before](https://fluxcd.io/flux-security-report-with-review-2023.pdf#h.nr7usjsjajp)).


internal/controller/warehouses/images.go
```diff
@@ -58,6 +58,7 @@ func (r *reconciler) selectImages(
 			sub.AllowTags,
 			sub.IgnoreTags, // TODO: KR: Fix this
```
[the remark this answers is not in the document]
krancour:
```suggestion
			sub.IgnoreTags,
```

internal/controller/warehouses/images.go
```diff
@@ -109,17 +110,19 @@ func getImageRefs(
 	allowTagsRegex string,
 	ignoreTags []string,
 	platform string,
+	insecureSkipVerify bool,
```
krancour, line 113: This was my mistake and not yours, but maybe worth fixing while touching this code?

We should probably just take the whole sub as an arg instead of all the fields individually.

internal/image/selector.go
```diff
@@ -64,6 +64,9 @@ type SelectorOptions struct {
 	// Creds holds optional credentials for authenticating to the image
 	// repository.
 	Creds *Credentials
+	// InsecureSkipVerify is an optional flag, that if set to true, will disable
+	// verification of the image repository's TLS certificate.
+	InsecureSkipVerify bool
```
krancour, line 69: Nit: The field name isn't clear that _cert_ verification is the verification it skips. I'd also be ok with dropping the "insecure" prefix here because I see that as more of a scare word for _user facing_ things. (So it's good that it is in the corresponding field in the CRD.)

None of this is a hill I'd die on. :100:% dealer's choice.

internal/image/repository_client.go
```diff
@@ -132,11 +135,18 @@ func newRepositoryClient(
 	image := reg.normalizeImageName(reference.Path(repoRef))
 	apiAddress := strings.TrimSuffix(reg.apiAddress, "/")
 
+	httpTransport := cleanhttp.DefaultTransport()
```
krancour, line 138: Nice improvement. Thank you!

krancour (approved):
Made some comments for your consideration, but nothing that blocks approval.

LGTM and thank you!

internal/image/selector.go
hiddeco: I changed it to `InsecureSkipTLSVerify` as this appears to be the common name in the code base.

internal/controller/warehouses/images.go
hiddeco: This had caught my 👀 as well, and I have changed it now.

Issue #1549, opened 2024-02-28
Unsynced ressources forever with ArgoCd


stephaneetje:
Since version 0.4.0 of the chart, argocd sees some changes that never can be made, and kargo remains OutOfSync forever.
Here is the screenshot with the concerned ressources 

<img width="1148" alt="Capture d’écran 2024-02-28 à 10 23 18" src="https://github.com/akuity/kargo/assets/73826402/15f833f7-3c58-4af4-86e8-e6495ed82d66">

krancour:
@stephaneetje that's not a lot of information to go on. Do you have an error message?

stephaneetje:
I don't have any error message, if i modify manually the ressources for the requested change with `name: kargo...` with kubectl , it saves successfuly but automaticaly delete the entry, so no error.

stephaneetje:
With latest version, the problem is still here only for the 2 cluster roles :

<img width="1199" alt="Capture d’écran 2024-02-29 à 09 27 08" src="https://github.com/akuity/kargo/assets/73826402/99a43c30-95aa-4eab-bcd1-c6b462974788">

krancour:
Backtracking here to make sure I am clear on the context... you are using Argo CD to manage the installation of Kargo. Is that correct?

> argocd sees some changes that never can be made

There is no explanation for why surfaced anywhere in the Argo CD UI?

stephaneetje:
You perfectly understood! 
I see absolutly no error, it just syncs successfully, but remains outofsync, as if the change gets ignored by kubernetes.

<img width="1278" alt="Capture d’écran 2024-02-29 à 15 28 12" src="https://github.com/akuity/kargo/assets/73826402/7451aaa8-3963-42be-bf27-4d95b0df4edc">

stephaneetje:
You perfectly understood!
I see absolutly no error, it just syncs successfully, but remains outofsync, as if the change gets ignored by kubernetes.

<img width="1207" alt="Capture d’écran 2024-02-29 à 15 30 33" src="https://github.com/akuity/kargo/assets/73826402/737f70fd-611b-4307-a05f-b5d75b5b5e13">

krancour:
Your last sync was to commit b802577. Argo CD is aware of 1190e9f, but _hasn't_ synced to it because you have auto-sync disabled (which is fine). Your original screen shot is showing you what has changed between b802577 and 1190e9f.

All you need to do is click sync.

stephaneetje:
[...]

PR #1594, opened 2024-03-09
fix(chart): preventing rendering confimaps key as nil

ftmiro:
When running the Quick start installation script, a user may face the following error:

```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: [unknown object type "nil" in ConfigMap.data.GLOBAL_CREDENTIALS_NAMESPACES, unknown object type "nil" in ConfigMap.data.ROLLOUTS_ANALYSIS_RUNS_NAMESPACE, unknown object type "nil" in ConfigMap.data.ROLLOUTS_CONTROLLER_INSTANCE_ID]
```

The error is caused by these keys being rendered as `nil`:

```
helm template kargo oci://ghcr.io/akuity/kargo-charts/kargo \
--set api.service.type=NodePort \
--set api.adminAccount.password=admin \
--set api.adminAccount.tokenSigningKey=iwishtowashmyirishwristwatch |grep ROLLOUT
```

```
  ROLLOUTS_INTEGRATION_ENABLED: "true"
  ROLLOUTS_INTEGRATION_ENABLED: "true"
  ROLLOUTS_ANALYSIS_RUNS_NAMESPACE:
  ROLLOUTS_CONTROLLER_INSTANCE_ID:
```

### DIFF
--- a/charts/kargo/templates/controller/configmap.yaml
+++ b/charts/kargo/templates/controller/configmap.yaml
@@ -15,7 +15,9 @@ data:
   {{- if .Values.kubeconfigSecrets.kargo }}
   KUBECONFIG: /etc/kargo/kubeconfigs/kubeconfig.yaml
   {{- end }}
+  {{- if .Values.controller.globalCredentials.namespaces }}
   GLOBAL_CREDENTIALS_NAMESPACES: {{ join "," .Values.controller.globalCredentials.namespaces }}
+  {{- end }}
   ARGOCD_INTEGRATION_ENABLED: {{ quote .Values.controller.argocd.integrationEnabled }}
   {{- if .Values.controller.argocd.integrationEnabled }}
   {{- if .Values.kubeconfigSecrets.argocd }}
@@ -29,7 +31,11 @@ data:
   {{- if .Values.kubeconfigSecrets.rollouts }}
   ROLLOUTS_KUBECONFIG: /etc/kargo/kubeconfigs/rollouts-kubeconfig.yaml
   {{- end }}
+  {{- if .Values.controller.rollouts.analysisRunsNamespace }}
   ROLLOUTS_ANALYSIS_RUNS_NAMESPACE: {{ .Values.controller.rollouts.analysisRunsNamespace }}
+  {{- end}}
+  {{- if .Values.controller.rollouts.controllerInstanceID }}
   ROLLOUTS_CONTROLLER_INSTANCE_ID: {{ .Values.controller.rollouts.controllerInstanceID }}
+  {{- end }}
   {{- end }}
 {{- end }}
--- a/charts/kargo/templates/controller/configmap.yaml
+++ b/charts/kargo/templates/controller/configmap.yaml
@@ -15,9 +15,7 @@ data:
   {{- if .Values.kubeconfigSecrets.kargo }}
   KUBECONFIG: /etc/kargo/kubeconfigs/kubeconfig.yaml
   {{- end }}
-  {{- if .Values.controller.globalCredentials.namespaces }}
-  GLOBAL_CREDENTIALS_NAMESPACES: {{ join "," .Values.controller.globalCredentials.namespaces }}
-  {{- end }}
+  GLOBAL_CREDENTIALS_NAMESPACES: {{ quote (join "," .Values.controller.globalCredentials.namespaces) }}
   ARGOCD_INTEGRATION_ENABLED: {{ quote .Values.controller.argocd.integrationEnabled }}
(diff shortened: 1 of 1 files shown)


krancour:
> When running the Quick start installation script...

> ```
> helm template kargo oci://ghcr.io/akuity/kargo-charts/kargo \
> --set api.service.type=NodePort \
> --set api.service.nodePort=31444 \
> --set api.adminAccount.password=admin \
> --set api.adminAccount.tokenSigningKey=iwishtowashmyirishwristwatch |grep ROLLOUT
> ```

The quickstart scripts do not use this command.

The resulting YAML is also valid _and_ accepted by k8s.

I would take a guess that you encountered this while installing Kargo via Argo CD, in which case, [this is a known issue](https://github.com/akuity/kargo/issues/1549).

I'm not resistant to "fixing" this, but as stated in the other issue, there are likely dozens of _other_ spots in the chart where this same issue is possible and no one has noticed yet because they haven't turned tweaked the relevant settings. Before scouring the chart to find them all, I'd like to know why Argo CD is choking on valid YAML and what else can be done about that.

@morey-tech I know you looked at #1549 previously. Have you any further insight into why Argo CD doesn't like this?

You can pipe the output of that `helm template` command into `kubectl apply` and there is no problem whatsoever.

krancour:
Ok... I've read through https://github.com/argoproj/argo-cd/issues/15566 and understand the underlying Argo CD issue.

I think we can square this away, but there are definitely many more places in the chart affected than just these two.

I also want to weigh the possibility of quoting the possibly null values rather than using a whole lot of conditionals to cover the possibility of any value being undefined.

ftmiro:
> > When running the Quick start installation script...
> 
> > ```
> > helm template kargo oci://ghcr.io/akuity/kargo-charts/kargo \
> > --set api.service.type=NodePort \
> > --set api.service.nodePort=31444 \
> > --set api.adminAccount.password=admin \
> > --set api.adminAccount.tokenSigningKey=iwishtowashmyirishwristwatch |grep ROLLOUT
> > ```
> 
> The quickstart scripts do not use this command.
> 
> The resulting YAML is also valid _and_ accepted by k8s.
> 
> I would take a guess that you encountered this while installing Kargo via Argo CD, in which case, [this is a known issue](https://github.com/akuity/kargo/issues/1549).
> 
> I'm not resistant to "fixing" this, but as stated in the other issue, there are likely dozens of _other_ spots in the chart where this same issue is possible and no one has noticed yet because they haven't tweaked the relevant settings. Before scouring the chart to find them all, I'd like to know why Argo CD is choking on valid YAML and what else can be done about that.
> 
> @morey-tech I know you looked at #1549 previously. Have you any further insight into why Argo CD doesn't like this?
> 
> You can pipe the output of that `helm template` command into `kubectl apply` and there is no problem whatsoever.

Indeed, the quick start script uses `helm install`. I'm using the `kind.sh` script, so, to understand what was the issue, I used the `helm template` command to inspect the generated manifests. I believe ArgoCD was not the one creating the issue since I see the error being thrown by a `helm install`.

I agree that this PR tries to fix only 1 specific error and not every other possible similar problem. I also prefer to pipe to `|quote` to guarantee an empty string.

ftmiro:
> Ok... I've read through [argoproj/argo-cd#15566](https://github.com/argoproj/argo-cd/issues/15566) and understand the underlying Argo CD issue.
> 
> I think we can square this away, but there are definitely many more places in the chart affected than just these two.
> 
> I also want to weigh the possibility of quoting each possibly null value rather than using a whole lot of conditionals to account for them.

I also believe is much better to set it as an empty string than add a lot of `ifs`. My concern doing this is that I'm not sure about the config KEYS that can cause a side effect if passed as empty, for example, the key SHARD_NAME is currently conditionally added to the config map:

```
  {{- if .Values.controller.shardName }}
  SHARD_NAME: {{ .Values.controller.shardName }}
  {{- end }}
```

ftmiro:
Hi @krancour , I updated my local Kubernetes cluster to 1.29 and repeated the quickstart script, and the kargo chart 0.4.3 was installed with no issues. I think the problem I was having was observed only in the Kubernetes version previous than 1.27.

krancour:
> I think the problem I was having was observed only in the Kubernetes version previous than 1.27.

@ftmiro that is interesting insight. Up to now, this was believed to be strictly an Argo CD issue. It is enlightening to know that k8s version may also play a role.

krancour:
#1598 was a more holistic alternative to this (accounted for all fields that we thought could be problematic). It has since been merged, so I'm closing this.

PR #1598, opened 2024-03-11
chore(chart): be more careful about undefined values

### LINKED ISSUE #1549: Unsynced ressources forever with ArgoCd

krancour:
Fixes #1549

Alternative to #1594

Some users who install Kargo via Argo CD have noted difficulties syncing that are likely attributable to https://github.com/argoproj/argo-cd/issues/15566

While #1594 proposed two specific changes to the most noteworthy occurrences of the problem, this PR attempts a more holistic remedy.

1. In Go code, define default namespace for Argo CD _in only one place._ This information is used by multiple components, and each previously set their own default (although they all set it the same). Remedying this seemed necessary for what follows...

1. In Go code, make the default ttl for admin tokens 24h. This differed from the default set in the chart itself. Once again, this seemed necessary for what follows...

1. Chart modification according to the following framework:

    * Add quotes around anything where there is:
        * No default value in `values.yaml`.
        * A default value from `values.yaml`, but overriding it as undefined would be acceptable because Go code will catch the invalid configuration.
        * A default value from `values.yaml`, but overriding it as undefined would be acceptable because Go code will provide its own default value.

    * Add a default value _in relevant templates_ where:
        * A default value from `values.yaml` exists, but if overridden as undefined, would be problematic.

    * Leave things alone / unquoted, un-defaulted if:
        * There is already a check for nil/empty values preventing any configuration issues.
        * Kubernetes will provide its own default.

I will annotate a few individual changes to clarify the rationale.

### DIFF
--- a/charts/kargo/templates/api/cert.yaml
+++ b/charts/kargo/templates/api/cert.yaml
@@ -9,7 +9,7 @@ metadata:
     {{- include "kargo.api.labels" . | nindent 4 }}
 spec:
   dnsNames:
-  - {{ .Values.api.host }}
+  - {{ quote .Values.api.host }}
   issuerRef:
     kind: Issuer
     name: kargo-selfsigned-cert-issuer
--- a/charts/kargo/templates/api/configmap.yaml
+++ b/charts/kargo/templates/api/configmap.yaml
@@ -9,7 +9,7 @@ metadata:
     {{- include "kargo.api.labels" . | nindent 4 }}
 data:
   KARGO_NAMESPACE: {{ .Release.Namespace }}
-  LOG_LEVEL: {{ .Values.api.logLevel }}
+  LOG_LEVEL: {{ quote .Values.api.logLevel }}
   {{- if .Values.kubeconfigSecrets.kargo }}
   KUBECONFIG: /etc/kargo/kubeconfig.yaml
   {{- end }}
@@ -26,8 +26,8 @@ data:
   {{- else }}
   ADMIN_ACCOUNT_TOKEN_ISSUER: http://{{ .Values.api.host }}
   {{- end }}
-  ADMIN_ACCOUNT_TOKEN_AUDIENCE: {{ .Values.api.host }}
-  ADMIN_ACCOUNT_TOKEN_TTL: {{ .Values.api.adminAccount.tokenTTL }}
+  ADMIN_ACCOUNT_TOKEN_AUDIENCE: {{ quote .Values.api.host }}
+  ADMIN_ACCOUNT_TOKEN_TTL: {{ quote .Values.api.adminAccount.tokenTTL }}
   {{- end }}
   {{- if .Values.api.oidc.enabled }}
   OIDC_ENABLED: "true"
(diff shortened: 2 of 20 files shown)


charts/kargo/templates/api/cert.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     {{- include "kargo.api.labels" . | nindent 4 }}
 spec:
   dnsNames:
-  - {{ .Values.api.host }}
+  - {{ quote .Values.api.host }}
```
krancour, line 12: If this ends up as `""`, let cert manager worry about it.

charts/kargo/templates/api/configmap.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     {{- include "kargo.api.labels" . | nindent 4 }}
 data:
   KARGO_NAMESPACE: {{ .Release.Namespace }}
-  LOG_LEVEL: {{ .Values.api.logLevel }}
+  LOG_LEVEL: {{ quote .Values.api.logLevel }}
```
krancour, line 12: `values.yaml` sets this to `INFO`. If a user set it to `""`, that should be ok because the Go code provides its own default.

Quotes added to avoid https://github.com/argoproj/argo-cd/issues/15566

Same change applied in other configmaps as well.

charts/kargo/templates/api/configmap.yaml
```diff
@@ -18,16 +18,16 @@ data:
   TLS_CERT_PATH: /etc/kargo/tls.crt
   TLS_KEY_PATH: /etc/kargo/tls.key
   {{- end }}
-  PERMISSIVE_CORS_POLICY_ENABLED: {{ quote .Values.api.enablePermissiveCORSPolicy }}
+  PERMISSIVE_CORS_POLICY_ENABLED: {{ default "false" (quote .Values.api.enablePermissiveCORSPolicy) }}
```
krancour, line 21: This is defaulted to `false` in `values.yaml`. If a user should change that to `""`, this default in the template prevents a problem.

charts/kargo/templates/api/configmap.yaml
```diff
@@ -18,16 +18,16 @@ data:
...
   ADMIN_ACCOUNT_ENABLED: "true"
   {{- if or .Values.api.tls.enabled (and .Values.api.ingress.enabled .Values.api.ingress.tls.enabled) }}
   ADMIN_ACCOUNT_TOKEN_ISSUER: https://{{ .Values.api.host }}
   {{- else }}
   ADMIN_ACCOUNT_TOKEN_ISSUER: http://{{ .Values.api.host }}
   {{- end }}
-  ADMIN_ACCOUNT_TOKEN_AUDIENCE: {{ .Values.api.host }}
-  ADMIN_ACCOUNT_TOKEN_TTL: {{ .Values.api.adminAccount.tokenTTL }}
+  ADMIN_ACCOUNT_TOKEN_AUDIENCE: {{ quote .Values.api.host }}
+  ADMIN_ACCOUNT_TOKEN_TTL: {{ quote .Values.api.adminAccount.tokenTTL }}
```
krancour, line 30: Either of these could reasonably have been left undefined by a user. Go code will catch the issue.

Quotes added to avoid https://github.com/argoproj/argo-cd/issues/15566.

charts/kargo/templates/api/ingress.yaml
```diff
@@ -16,10 +16,10 @@ spec:
   ingressClassName: {{ .Values.api.ingress.ingressClassName }}
   {{- end }}
   rules:
-  - host: {{ .Values.api.host }}
+  - host: {{ quote .Values.api.host }}
     http:
       paths:
-      - pathType: {{ .Values.api.ingress.pathType }}
+      - pathType: {{ .Values.api.ingress.pathType | default "ImplementationSpecific" }}
```
krancour, line 22: `ImplementationSpecific` is the default in `values.yaml`, but this must never be blank. Default added here to account for the possibility of a user changing it to `""`.

charts/kargo/templates/argocd/role-binding.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: kargo-controller
-  namespace: {{ .Values.controller.argocd.namespace }}
+  namespace: {{ .Values.controller.argocd.namespace | default "argocd" }}
```
krancour, line 6: This is already defaulted to `argocd` in `values.yaml`, but problems occur if a user overrides it to `""`.

Go code will default to `argocd` if this is unspecified...

The problem this change actually avoids is the resource being created in the cluster's _default_ namespace when this value is unspecified.

krancour: Same comment applies in several other places.

charts/kargo/templates/dex-server/secret.yaml
```diff
@@ -27,10 +27,10 @@ stringData:
       http: 0.0.0.0:5558
 
     oauth2:
-      skipApprovalScreen: {{ .Values.api.oidc.dex.skipApprovalScreen }}
+      skipApprovalScreen: {{ .Values.api.oidc.dex.skipApprovalScreen | default "true" }}
```
krancour, line 30: This prevents invalid configuration if a user blanks out the setting in `values.yaml`.

charts/kargo/templates/api/configmap.yaml
krancour: Change removed... Go code provides its own default and will also error on values that don't parse as a boolean.

hiddeco (approved):
Based on the issue references (and additional comments), this all makes sense to me! 🥇

PR #1599, opened 2024-03-11
chore(deps): drop `github.com/pkg/errors`

hiddeco:
In favor of Go's built-in error wrapping since 1.13 (~2019).

As discussed previously in https://github.com/akuity/kargo/pull/1561#discussion_r1510081840


krancour (approved):
Gave this a cursory once over. Based mainly on prior discussion, LGTM.

PR #1606, opened 2024-03-12
fix: bug loading rest config

krancour:
Fixes a bug that snuck in with #1599. I'm not sure how many other similar mistakes may exist.


hiddeco (approved):
Argh, I discovered this mistake myself and thought I had corrected them all. So this should be the only one, or at least one of the very few.

Issue #1581, opened 2024-03-06
add verification management endpoints


krancour:
This is half-baked, but my initial thoughts are that we _at least_ need:

* A way to re-run verification. Presumably this is for a failed verification. imho, it should be ok to re-run successful verifications, but a failed re-run should not "unverify" the Freight in that Stage, because that Freight may already have been passed downstream after the previously successful verification process.

* Abort a verification. Especially because verifications can involve user-defined processes, it is conceivable that one could get "stuck." We would benefit from the ability to manually abort.

hiddeco:
How would we go with a scenario where the template contains an error, which is then later updated to correct the mistake?

Personally, I would assume in this case Kargo would be smart enough to re-issue a verification with the newly built AnalysisRun _without_ requiring manual user intervention. At present, this appears to not be the case.

krancour:
I think that is something to think about...

I would not oppose the notion of reconciling AnalysisTemplates in the management controller and, on change, retrying any failed verification that depends on that template.

Since multiple Stages can all use the same AnalysisTemplate for verification, we may need to think through whether this creates the possibility of a thundering herd. At present, I don't think it does because Stages can only reference AnalysisTemplates in their own namespaces, so the number of Stages associated with a given template should be small.

This could become a problem to watch out for if we eventually open the possibility of sharing AnalysisTemplates "globally." Some orgs will have a standard verification process that they'll want to share widely across projects, so I foresee this as likely.

jessesuen:
> A way to re-run verification

Agree we should provide a way to simply run ad-hoc verifications. For a given stage, sometimes people will want to just kick off a test/analysis just to see if that environment is working that day.

> Abort a verification.

Agree. Analysis might be an expensive K8s Job that user can't wait for it to finish.

NOTE: in argo rollouts, to terminate an AnalysisRun, `spec.terminate` just needs to be set to true.

https://github.com/argoproj/argo-rollouts/blob/master/pkg/kubectl-argo-rollouts/cmd/terminate/terminate.go#L16

PR #1611, opened 2024-03-13
feat(api/cli): add verification management endpoints

hiddeco:
Fixes: #1581

This pull requests adds mechanisms to:

1. Reverify the Freight of a Stage
2. Abort a (long-)running verification for the Freight of a Stage

Both are implemented through annotations (`kargo.akuity.io/reverify` and `kargo.akuity.io/abort`), allowing the behavior to be triggered outside of the API/CLI by annotating the Stage with the respective annotation using the ID of the existing verification information for the Stage as the value.

In addition to this, a `kargo verify stage (NAME) [--abort]` command has been added to make the features available to CLI users.

---

There is more to cover here in the future, e.g. keeping track of a "stack" of `VerificationInfo` objects to provide an historical overview, and to detect changes as mentioned in https://github.com/akuity/kargo/issues/1581#issuecomment-1981166069. However, they are out of scope for this pull request and should be handled as separate feature requests.


internal/controller/stages/stages.go
```diff
@@ -706,30 +713,42 @@ func (r *reconciler) syncNormalStage(
...
-					status.CurrentFreight.VerificationInfo = r.startVerificationFn(ctx, stage)
+		if stage.Spec.Verification != nil {
+			// Confirm if a rerun of verification is requested. If so, clear the
+			// verification info and start the verification process again.
+			info := status.CurrentFreight.VerificationInfo
+			if info != nil && info.AnalysisRun != nil && info.Phase.IsTerminal() {
+				if v, ok := stage.GetAnnotations()[kargoapi.AnnotationKeyReconfirm]; ok && v == info.AnalysisRun.Name {
+					logger.Debug("reconfirming verification")
+					status.Phase = kargoapi.StagePhaseVerifying
+					status.CurrentFreight.VerificationInfo = nil
```
hiddeco, line 724: @krancour this has ended up slightly different than we discussed offline. As requeueing would have caused the annotation to be removed prematurely, making it impossible to determine later on if we should (not) take existing AnalysisRuns for the Stage/Freight into account in `startVerification`.

api/v1alpha1/annotations.go
```diff
@@ -3,6 +3,8 @@ package v1alpha1
 const (
 	AnnotationKeyRefresh = "kargo.akuity.io/refresh"
 
+	AnnotationKeyReconfirm = "kargo.akuity.io/reconfirm"
```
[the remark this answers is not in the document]
krancour: > an alternative I had in mind myself was "reverify"

I like that alternative as well. "confirm" isn't nomenclature we've used anywhere else yet.

api/service/v1alpha1/service.proto
```diff
@@ -32,6 +32,7 @@ service KargoService {
   rpc PromoteStage(PromoteStageRequest) returns (PromoteStageResponse);
   rpc PromoteSubscribers(PromoteSubscribersRequest) returns (PromoteSubscribersResponse);
   rpc RefreshStage(RefreshStageRequest) returns (RefreshStageResponse);
+  rpc RequestStageVerification(RequestStageVerificationRequest) returns (RequestStageVerificationResponse);
```
krancour, line 35: Two things here.

1. I don't think I would count verification specifically as a Stage-oriented operation. It's really verification of a piece of Freight _with respect to_ a Stage. (Side note: I've been meaning for a long time to rename `PromoteStage` for the same reason. I'll finally make an issue for it at least so we don't lose track.)

2. Much more of a nit: I suppose it's _accurate_ that this endpoints _requests_ re-verification, but the corresponding message types, `RequestStageVerificationRequest` and `RequestStageVerificationResponse` sound awkward to me.

In light of both of the above, could I suggest tentatively renaming this to `Reverify`, with corresponding message types as `ReverifyRequest` and `ReverifyResponse`?

api/v1alpha1/stage_types.go
```diff
@@ -726,17 +726,19 @@ type AnalysisRunArgument struct {
 // VerificationInfo contains information about the currently running
 // Verification process.
 type VerificationInfo struct {
+	// ID is the identifier of the Verification process.
+	ID string `json:"id,omitempty" protobuf:"bytes,1,opt,name=id"`
```
hiddeco, line 730: @krancour do we have any rules around (not) restructuring the protobuf order while we are still `< v1.0.0`?

internal/controller/stages/stages.go
```diff
@@ -706,30 +742,60 @@ func (r *reconciler) syncNormalStage(
...
+						if v, ok := stage.GetAnnotations()[kargoapi.AnnotationKeyAbort]; ok && v == newInfo.ID {
+							log.Debug("aborting verification")
+							if err := r.abortVerificationFn(ctx, stage); err != nil {
+								return status, fmt.Errorf(
+									"error aborting verification for Stage %q in namespace %q: %w",
+									stage.Name,
+									stage.Namespace,
+									err,
+								)
+							}
```
hiddeco, line 784: I am on the fence here about returning an error, or returning a `VerificationInfo` struct with the error.

api/v1alpha1/stage_types.go
krancour: Not really at this stage, although I've occasionally done it the right way instead of renumbering.

Buuuuut... that was before #1515 

Did you renumber these by hand?

I believe that _since_ #1515, you can omit protobuf info from new fields in the struct and the codgen process will add it for you and _should_ get it right.

internal/controller/stages/stages.go
```diff
@@ -536,6 +558,14 @@ func (r *reconciler) Reconcile(
 	if clearRefreshErr != nil {
 		logger.Errorf("error clearing Stage refresh annotation: %s", clearRefreshErr)
 	}
+	clearReconfirmErr := kargoapi.ClearStageReverify(ctx, r.kargoClient, stage)
```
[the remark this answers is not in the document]
krancour: Both sgtm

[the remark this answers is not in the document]
krancour: btw... We _do_ have a generic `PatchStatus`. It's a more recent phenomenon that we have to also patch annotations sometimes. So I agree that a generic func for this is called for.

internal/controller/stages/stages.go
krancour: > Another interesting thing I have noticed is that if you provide Argo Rollouts with a terminate request, it appears that in some cases the AnalysisRun will end up as Successful. I wonder if this is actually what we want to happen, or if we should construct a custom VerificationInfo with data which reflects the abort operation having taken place.

I think we have separate field for VerificationInfo phase and then, under AnalysRunReference, it has its own phase.

So we should be able to capture that the AnalysisRun succeeded and _still_ record that the verification was aborted.

In determining the VerificationInfo's phase, I agree that the attempt to abort takes precedence over the fact that the attempt raced the AnalysisRun and lost with the AnalysisRun succeeding.

krancour: > As in the current implementation, we do not distinguish an "object not found" error from e.g. "we temporary can't reach the Kubernetes API server". Which means that even if we could eventually recover from the error, the user has to rerun the verification because the controller has given up on the previous AnalysisRun.

Right. We make no attempt currently to distinguish between errors where a retry helps and one where it doesn't.

I support however you might wish to improve that.

internal/cli/cmd/verify/stage.go
```diff
@@ -0,0 +1,134 @@
...
+kargo config set-project my-project
+kargo verify stage my-stage
+
+# Abort the verification of a stage's current freight
+kargo verify stage --project=my-project my-stage --abort
+
+# Abort the verification of a stage in the default project
+kargo config set-project my-project
+kargo verify stage my-stage --abort
+`,
```
krancour, line 50: Nit: We need to fix this in a bunch of places. We're getting an extra line break at the end of the examples section (two instead of one), so we should probably move these ending ticks to the end of the previous line.

Not a blocker. Just a thing I've noticed -- and definitely not unique to this PR.

internal/controller/stages/stages.go
hiddeco: > In determining the VerificationInfo's phase, I agree that the attempt to abort takes precedence over the fact that the attempt raced the AnalysisRun and lost with the AnalysisRun succeeding.

It is actually not a race, it is the AnalysisRun reflecting a success while it was the result from a termination request. I am not sure if this is default behavior for Kargo Rollouts, or if it depends on the command which is used to perform the analysis and the potential exit code it gives when it e.g. receives a `SIGKILL`. FWIW, I am testing things with a simple `sleep 600; exit 0;`.

I covered a potential race by checking the current state of the AnalysisRun _before_ adhering to the abort request, which prevents an unnecessary patch operation.

> We make no attempt currently to distinguish between errors where a retry helps and one where it doesn't.
>
> I support however you might wish to improve that.

I think improving this is too big of a change for the current scope of the PR, as it would require bubbling up Kubernetes errors (or translating them into our own typed errors). Where as at present, we have a pattern of translating not found to a `nil` object while returning other errors.

Given this, I am inclined to keep the pattern of reflecting errors in `VerificationInfo` for now. To address this in full in a follow up PR.

internal/cli/cmd/verify/stage.go
hiddeco: It also actually has a leading newline everywhere. We should probably just do:

```
Example: `# Verify a stage
kargo verify stage --project=my-project my-stage`,
```

api/v1alpha1/stage_types.go
hiddeco: I renumbered them by hand because it felt weird to not have an identifier at the beginning of the payload. I will see what the codegen process makes of it when I let the computer do the work.

krancour: afaik, it looks at the highest numbered field, adds one, and that's its starting point for numbering new fields, regardless of where they're positioned in the struct. i.e. It could still be at the top (where I agree an identifier belongs), but end up as a much larger number. But I'm basing that on what I've seen it do so far and not knowledge of the underlying implementation.

internal/controller/stages/stages.go
krancour: > It is actually not a race, it is the AnalysisRun reflecting a success while it was the result from a termination request.

Ah. I was mistakenly thinking the AnalysisRun succeeded between the time the abort annotation was added and the time the Stage was reconciled.

This behavior seems peculiar indeed.

@jessesuen would probably be able to give us some insight on this...

@jessesuen you know anything about aborted AnalysisRuns having phase Success?

internal/cli/cmd/verify/stage.go
krancour: > It also actually has a leading newline everywhere.

Funny... I noticed the double blank line after examples and never noticed that the _one_ blank line between "Examples:" and the actual examples was still one blank line more than what is between all the other help section headings and their content.

Good eye.

internal/cli/cmd/verify/stage.go
```diff
@@ -0,0 +1,134 @@
...
+type verifyStageOptions struct {
+	Config        config.CLIConfig
+	ClientOptions client.Options
+
+	Project string
+	Name    string
+	Abort   bool
+}
+
+func newVerifyStageCommand(cfg config.CLIConfig) *cobra.Command {
```
krancour, line 27: This is another inconsistency we can fix when time permits. (Not specifically introduced here.)

These constructor-like funcs for building cobra commands are sometimes named like `newNounCommand` and other times are named like `newVerbNounCommand`.

I'll open a separate issue.

internal/cli/cmd/verify/stage.go
```diff
@@ -0,0 +1,134 @@
...
+		Example: `
+# Rerun the verification of the stage's current freight
+kargo verify stage --project=my-project my-stage
+
+# Rerun the verification of a stage in the default project
+kargo config set-project my-project
+kargo verify stage my-stage
+
+# Abort the verification of a stage's current freight
+kargo verify stage --project=my-project my-stage --abort
```
krancour, line 45: I'm a little iffy on `--abort` as a flag on `kargo verify stage`.

We've gotten a lot better about consistently adhering to `kargo verb nound` and this feels like a departure from that.

I get why you did it this way. There's currently nothing else "abort" applies to, which would make it feel like some awkward one-off sub-command...

I feel we will soon have _other_ things that can be aborted as well. I know @jessesuen has mentioned aborting promotions a time or two before...

We might want to plan ahead for that and consider making `abort` its own subcommand after all.

internal/controller/stages/stages.go
krancour: I might be missing something here.

Here's how I thought this would work:

1. If the annotation is present and matched current verification info then blank that info out and put the phase back to Verifying. Early return.

2. Clear the annotations only if we successfully updated State status. Meaning:

    1. On success, next reconcile will lack the annotations and also have blank verification info and phase Verifying -- it will look like we're starting fresh.

    2. On failure, next reconcile will still have the annotations and will try again to blank out the verification info and set status to Verifying.

Is the issue right now mainly that we're not clearing the annotations _conditionally_ on status update success?

internal/controller/stages/stages.go
krancour: Thinking about this more, this shouldn't block the PR.

As long as the verification phase correctly identifies that the verification was aborted, I don't think the phase of the analysis run, _even if it is arguably incorrect_, matters anywhere else in Kargo.

internal/controller/stages/stages.go
krancour: I think this is the only thing that may be blocking. Everything else lgtm.

internal/controller/stages/stages.go
jessesuen: > it appears that in some cases the AnalysisRun will end up as Successful

Is it some cases or all?

> it is the AnalysisRun reflecting a success while it was the result from a termination request

I seem to recall that this was designed behavior for AnalysisRun. There is such thing as an "indefinite" analysisrun that needs to stopped manually / programmatically. 

Let me check on this.

jessesuen: Yes, when we terminate an AnalysisRun, we consider it a Successful run:

https://github.com/argoproj/argo-rollouts/blob/master/analysis/analysis.go#L581-L585
https://github.com/argoproj/argo-rollouts/blob/master/analysis/analysis.go#L610-L614

It's because Argo Rollouts will terminate indefinite/background AnalyisRuns when the update is completed. So in the case of Rollouts, termination is a happy path (for background analysis)

I'm thinking about how Kargo should handle this. I think in the case of Kargo, termination is more of a user decision, whereas in Argo Rollouts, the controller deciding to stop the run because the update completed.

Given that, perhaps an terminated analysis run should be treated as failed analysis.

internal/controller/stages/stages.go
hiddeco: The problem is that if we blank out the verification information, and then return to have the annotation removed. `startVerification` is designed to look for existing AnalysisRuns _before_ creating a new one: https://github.com/akuity/kargo/blob/f6fd5d15daa6e1688c282bb88a91dba331e16480/internal/controller/stages/verification.go#L35-L71

With an early return, we do not have anything at hand which allows us to differentiate as the annotation has been removed. Given this, I opted for not returning early. Which allows us to conditionally skip looking for existing AnalysisRuns.

api/v1alpha1/stage_types.go
hiddeco: You were right, and I amended my commit to let the computer do its thing.

internal/cli/cmd/verify/stage.go
hiddeco: What would the full abort version be called in this case? As with `kargo abort verification --stage <stage>`, I feel it would be counterintuitive from `kargo verify stage`.

internal/cli/cmd/verify/stage.go
hiddeco: What I have see some projects do is adding a tiny wrapper which adds `strings.TrimSpace` (e.g. `printDescription`), as this allows you to continue to maintain the visible advantage of having everything start at the same point in-code and/or allow you to move it to constants defined outside of the struct.

internal/controller/stages/stages.go
krancour: > and then return to have the annotation removed. startVerification is designed to look for existing AnalysisRuns before creating a new one

I'd forgotten about this. That was to account for the scenario that we successfully launched an AnalysisRun and _then_ encountered an error updating status...

That would have safeguarded against duplicate runs being created.

Problem understood, but I guess I am still a little confused by how this _does_ work now. 👀 some more. Standby.

krancour: Ok... I get it now.

We don't need a second reconciliation attempt in order for the new verification attempt to proceed. After you've blanked out the old into, the _current_ reconcilation attempt proceeds with starting verification anew.

There might be an edge case here... I am still 👀

internal/controller/stages/verification.go
```diff
@@ -32,41 +35,54 @@ func (r *reconciler) startVerification(
...
-		return &kargoapi.VerificationInfo{
-			Phase: kargoapi.VerificationPhase(analysisRuns.Items[0].Status.Phase),
-			AnalysisRun: &kargoapi.AnalysisRunReference{
-				Name:      analysisRuns.Items[0].Name,
-				Namespace: analysisRuns.Items[0].Namespace,
-				Phase:     string(analysisRuns.Items[0].Status.Phase),
+	// If the stage does not have a reverification annotation, check if there is
+	// an existing AnalysisRun for the Stage and Freight. If there is, return
+	// the status of this AnalysisRun.
+	if _, ok := stage.GetAnnotations()[kargoapi.AnnotationKeyReverify]; !ok {
```
krancour, line 41: What happens here if we previously succeeded in restarting verification, but failed to clear the annotation?

It looks to me like that would result in a duplicate AnalysisRun.

internal/controller/stages/stages.go
krancour: Continuing this convo in a more [relevant spot](https://github.com/akuity/kargo/pull/1611/files#r1526446585)

internal/controller/stages/verification.go
hiddeco: In that case (assuming the status update patch _did_ work), we would not end up here.

krancour: 🤦‍♂️

internal/controller/stages/verification.go
```diff
@@ -224,6 +251,54 @@ func (r *reconciler) getVerificationInfo(
...
+	}
+
+	// Return a new VerificationInfo with the same ID and a message indicating
+	// that the verification was aborted. The Phase will be set to Failed, as
+	// the verification was not successful.
+	// We do not use the further information from the AnalysisRun, as this
+	// will indicate a "Succeeded" phase due to Argo Rollouts behavior.
+	return &kargoapi.VerificationInfo{
+		ID:      stage.Status.CurrentFreight.VerificationInfo.ID,
+		Phase:   kargoapi.VerificationPhaseFailed,
```
krancour, line 297: I think we should introduce an Aborted phase, then there are three distinct and non-overlapping negative outcomes:

* Kargo encountered a problem (error)
* Verification completed with a negative result (failure)
* Verification was aborted

wdyt?

hiddeco: This crossed my mind as well, so more than happy to add it!

krancour (approved):

internal/cli/cmd/verify/stage.go
krancour: Ya... now that you say that... that seems clunky because `verification` isn't a resource -- nor do I think it needs to be.

Obviously this was non-blocking anyway.

PR #1612, opened 2024-03-13
refactor: add alias field to freight; add default alias in webhook

### LINKED ISSUE #1604: proposal: create freight aliases in defaulting webhook
Currently, random Freight aliases are assigned by the Warehouse reconciler during Freight creation.

Working on #1600 today, I had the experience of scripting creation of a lot of Freight for testing purposes.

Generally, users do not create Freight yet and it is only done by the Warehouses, though we expect that may change eventually.

Having scripted Freight creation, I got no aliases and I found myself wishing the defaulting webhook had set the aliases for me.

cc @jessesuen

krancour:
Fixes #1604


internal/api/update_freight_alias_v1alpha1.go
```diff
@@ -99,9 +99,10 @@ func (s *server) patchFreightAlias(
 ) error {
 	patchBytes := []byte(
 		fmt.Sprintf(
-			`{"metadata":{"labels":{"%s":"%s"}}}`,
+			`{"metadata":{"labels":{%q:%q}},"alias":%q}`,
```
hiddeco, line 102: Is the "alias" part here actually correct?

hiddeco (approved):

PR #1615, opened 2024-03-13
fix: make warehouses not own freight

### LINKED ISSUE #1602: proposal: freight should not be owned by warehouses
While working on #1600, I stumbled on something I had forgotten...

Freight are currently owned by the Warehouse that creates them.

I don't love this, as it means that deleting a Warehouse takes all its Freight with it.

This feels too dangerous, frankly. I'm tentatively considering this a bug.

I believe we stop creating this ownership relationship between Warehouses and Freight.

Some components do index Freight by Warehouse, so we _would_ have to add an `warehouse` field to Freight to indicate where it came from.

cc @jessesuen for thoughts.

krancour:
Fixes #1602


api/v1alpha1/freight_types.go
```diff
@@ -29,6 +29,12 @@ type Freight struct {
 	// present, the defaulting webhook will choose an available alias and assign
 	// it to both the field and label.
 	Alias string `json:"alias,omitempty" protobuf:"bytes,7,opt,name=alias"`
+	// Warehouse is the name of the Warehouse that created this Freight. This is a
+	// required field. TODO: It is not clear yet how this field should be set in
+	// the case of user-defined Freight.
+	//
+	//+kubebuilder:validation:Required
```
hiddeco, line 36: Nit:

```suggestion
	// +kubebuilder:validation:Required
```

krancour: Taking the suggestion, but the missing space between `//` and `+` is pervasive in this package.

I know the reason for this is that the first few types were created using `kubebuilder create api ...` and it didn't include that space. I blindly continued followed that precedent.

I'm unopposed to a follow-up chore to change this more broadly. idk why, but I strongly agree it's nicer to look at with the space included.

hiddeco: Ah, it has been ages since I set up things using `kubebuilder`. In the past, once we had bootstrapped a project I never looked back and just wrote new API structures by hand. But this explains a lot.

krancour: > In the past, once we had bootstrapped a project I never looked back and just wrote new API structures by hand.

That's the same route we've taken. Haven't touched it in forever... until last night when I generated a test CRD to verify that was where this weird precedent initially came from.

hiddeco (approved):

PR #1616, opened 2024-03-14
chore: exclude generated code from coverage report

krancour:
Massive amounts of generated code are ruining our coverage numbers.

In fact #1515 reduced us from ~50% to ~20%.

After a bit of research, grep is the most direct way to fix this.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,6 +25,11 @@ jobs:
           ${{ runner.os }}-go-
     - name: Run unit tests
       run: make test-unit
+    - name: Remove generated code from report
+      run: |
+        grep -v .pb.go coverage.txt | grep -v zz_generated | grep -v service.connect.go > coverage.tmp
+        rm coverage.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -25,6 +25,11 @@ jobs:
           ${{ runner.os }}-go-
     - name: Run unit tests
       run: make test-unit
+    - name: Remove generated code from report
+      run: |
+        grep -v .pb.go coverage.txt | grep -v zz_generated | grep -v service.connect.go > coverage.tmp
+        rm coverage.txt
```
jessesuen, line 31: Pretty certain the `rm coverage.txt` line is unnecessary.

jessesuen (approved):

.github/workflows/ci.yaml
krancour: You're right. Thanks!

krancour:
```suggestion
```

PR #1618, opened 2024-03-14
chore: calculate unstable version as X.Y+1.0 instead of X.Y.Z+1

jessesuen:
Unstable nightly (which is built off of main) is currently being produced as `v0.4.1-unstable-20240313`, even though there is an officially released v0.4.2, v0.4.3. This will bump the Y (in X.Y) rather than current Z in (X.Y.Z).

Here is how it will calculate nightly version after this change:

```
$ LATEST_VERSION=$(go list -m -versions github.com/akuity/kargo | awk '{print $NF}' | awk -F "." '{print $1"."$2".0"}')

$ NEW_VERSION=$(awk 'BEGIN {FS=OFS="."} {$2++; print}' <<< "${LATEST_VERSION}")

$ echo "unstable-version=${NEW_VERSION}-unstable-$(date +'%Y%m%d')"
unstable-version=v0.5.0-unstable-20240313
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -34,7 +34,8 @@ jobs:
       id: unstable-version
       run: |
         set -xue
-        LATEST_VERSION=$(go list -f {{.Version}} -mod=mod -m github.com/akuity/kargo@main | cut -f1 -d'-')
+        # Once Kargo goes to v2, we need to replace github.com/akuity/kargo with github.com/akuity/kargo/v2 on the next line
+        LATEST_VERSION=$(go list -m -versions github.com/akuity/kargo | awk '{print $NF}' | awk -F "." '{print $1"."$2".0"}')
         NEW_VERSION=$(awk 'BEGIN {FS=OFS="."} {$2++; print}' <<< "${LATEST_VERSION}")
         echo "unstable-version=${NEW_VERSION}-unstable-$(date +'%Y%m%d')" >> $GITHUB_OUTPUT
     - name: Set up QEMU
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -35,7 +35,7 @@ jobs:
       run: |
         set -xue
         LATEST_VERSION=$(go list -f {{.Version}} -mod=mod -m github.com/akuity/kargo@main | cut -f1 -d'-')
-        NEW_VERSION=$(awk 'BEGIN {FS=OFS="."} {$3++; print}' <<< "${LATEST_VERSION}")
+        NEW_VERSION=$(awk 'BEGIN {FS=OFS="."} {$2++; print}' <<< "${LATEST_VERSION}")
         echo "unstable-version=${NEW_VERSION}-unstable-$(date +'%Y%m%d')" >> $GITHUB_OUTPUT
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v3


34fathombelow (approved):
LGTM, this changes makes sense.

.github/workflows/release.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
       run: |
         set -xue
         LATEST_VERSION=$(go list -f {{.Version}} -mod=mod -m github.com/akuity/kargo@main | cut -f1 -d'-')
-        NEW_VERSION=$(awk 'BEGIN {FS=OFS="."} {$3++; print}' <<< "${LATEST_VERSION}")
+        NEW_VERSION=$(awk 'BEGIN {FS=OFS="."} {$2++; print}' <<< "${LATEST_VERSION}")
```
krancour, line 38: This is quirky...

`go list -f {{.Version}} -mod=mod -m github.com/akuity/kargo@main | cut -f1 -d'-'` seems to, currently find `v0.4.0-rc.1.0.20240313162750-34c46f624f5a`... 

i.e. It is finding tonight's nightly. I assume it's actually finding the last semver-looking tag applied.

Overall, this is yielding `v0.5.0` as the value for `NEW_VERSION`. I think that differs from the X.Y (no patch version) that you had intended.

I think you also may be getting lucky that the _current_ nightly naming scheme has the patch version always set to 0.

If the last semver-ish looking tag applied had been `v0.4.4-abcdefg` instead, your `NEW_VERSION` would end up being `v0.5.4`, which is _definitely_ not what you are looking for.

Example:

```
LATEST_VERSION=$(printf v0.4.4-abcdefg | cut -f1 -d'-')
awk 'BEGIN {FS=OFS="."} {$2++; print}' <<< ${LATEST_VERSION}
```

I would think we'd want to use the latest _official_ release as the starting point.

I _think_ this is how we could do it:

```
LATEST_VERSION=$(go list -m -versions github.com/akuity/kargo | awk '{print $NF}' | awk -F "." '{print $1"."$2}')
NEW_VERSION=$(awk 'BEGIN {FS=OFS="."} {$2++; print}' <<< "${LATEST_VERSION}")
```

krancour: The suggested change would get you from `v0.4.4` (the latest official release) to `v0.5.0`

jessesuen: I think you're right. I tweaked your suggestion to add a `.0` at the end so that it becomes v0.5.0 so that the version will be a semantic version. Here is the result:

```
$ LATEST_VERSION=$(go list -m -versions github.com/akuity/kargo | awk '{print $NF}' | awk -F "." '{print $1"."$2".0"}')

$ NEW_VERSION=$(awk 'BEGIN {FS=OFS="."} {$2++; print}' <<< "${LATEST_VERSION}")

$ echo "unstable-version=${NEW_VERSION}-unstable-$(date +'%Y%m%d')"
unstable-version=v0.5.0-unstable-20240313
```

jessesuen: Also, when I tried `go list -m -versions github.com/akuity/kargo` against other go repositories (e.g. Argo CD), I learned that it will only return up to v1.x versions. As soon as we reach v2.x, we need to use `github.com/akuity/kargo/v2`. I left a comment for our future selves.

hiddeco (approved):

krancour (approved):

.github/workflows/release.yaml
krancour: > I tweaked your suggestion to add a .0 at the end so that it becomes v0.5.0 which will make it a semantic version

I was thrown off by the PR title.

Verifying intent -- we want X.Y.Z+1 --> X.Y+1.0 ?

krancour: > How is it even finding that? No such tag seems to even exist.

Man, the CLI help for `go list` is terrible.

GitHub Copilot was more helpful:

> The version v0.4.0-rc.1.0.20240314020641-0484299a739b you're seeing is a pseudo-version.
>
> ...
>
> * The base version is v0.4.0-rc.1
>
> ...
>
> This pseudo-version is generated because you're asking for the main branch with github.com/akuity/kargo@main, and the main branch does not have a semver-compatible tag that matches the latest commit.

Indeed... `v0.4.0-rc.1` was the last tag for a commit on the main branch.

`go list -m -versions github.com/akuity/kargo` on the other hand, just gives all semver tags from the repo, in order.

Confusion resolved.

jessesuen: > Verifying intent -- for an existing X.Y.Z, we want to find X.Y+1.0 ?

Yes, that was what I was aiming for.

Issue #1617, opened 2024-03-14
configmap/checksum and secret/checksum constantly changing


jessesuen:
# Description

I noticed that every time I updated from kargo nightly chart, the configmap or secret checksum would always change. This is despite me never changing anything with the kargo configmap/secret. I found out why:

```
        configmap/checksum: {{ include (print $.Template.BasePath "/api/configmap.yaml") . | sha256sum }}
        secret/checksum: {{ include (print $.Template.BasePath "/api/secret.yaml") . | sha256sum }}
```

The checksum includes the entire contents of configmap and secret yaml, instead of just the data field. This means it will checksum the annotations/labels, which are always changing with each chart release.

We should improve this to only sha256sum against the data fields, so that the checksum remains stable.

# Version

0.4.1-nightly

PR #1620, opened 2024-03-14
fix(chart): enhance ConfigMap/Secret checksum annotation logic

hiddeco:
Fixes: #1617 

By picking only the `data` or `stringData` subset for the respective ConfigMap or Secret to calculate the checksum. Ensuring only actual value changes update the annotation value.


krancour (approved):
🔥

PR #1623, opened 2024-03-14
refactor(api)!: rename promotion endpoints

### LINKED ISSUE #1622: rename two poorly named endpoints
A promotion is an operation that involves Freight _and_ one or more Stages.

The thing _being_ promoted is the Freight. The Stage or Stages are _where it's being promoted to_.

In light of that, the names for these two API endpoints don't make sense:

```
rpc PromoteStage(PromoteStageRequest) returns (PromoteStageResponse);
rpc PromoteSubscribers(PromoteSubscribersRequest) returns (PromoteSubscribersResponse);
```

Indeed... I'm not sure what it means to "promote a Stage." 🤔 

While we're still in "hurry up and break things before it's too late" mode, I'd like to move these out from under the "Stage operations" section of `service.proto` and into the "Freight operations" section (just to keep things nicely organized) and change their names to `PromoteToStage` and `PromoteToStageSubscribers`, respectively.

I believe these names will be far more clear about what these endpoints do.

krancour:
Fixes #1622


internal/api/promote_to_stage_subscribers_v1alpha1_test.go
```diff
@@ -15,21 +15,23 @@ import (
...
-func TestPromoteSubscribers(t *testing.T) {
+func TestPromoteToStageSubscribers(t *testing.T) {
 	testCases := []struct {
 		name       string
-		req        *svcv1alpha1.PromoteSubscribersRequest
+		req        *svcv1alpha1.PromoteToStageSubscribersRequest
 		server     *server
 		assertions func(
-			*connect.Response[svcv1alpha1.PromoteSubscribersResponse],
+			*testing.T,
```
hiddeco, line 24: At this rate we will be done changing all of them in no time 😄

krancour: Trying to do them every time I touch a test for other reasons. 😄

Issue #1536, opened 2024-02-26
Project creation wizard


jessesuen:
There is currently no ability to create a new Project via the UI.

@jessesuen "wizard" was your wording. imho, at _create_ time, we probably don't need more than a very simple dialog to ask for project name.

How about after drilling into a Project, much as when drilling into a GitHub repo, we have a little gear button or tab to open and edit Project configuration?

wdyt?

PR #1626, opened 2024-03-14
feat(ui): Project creation modal

rpelczar:
#1536 

image

image


jessesuen:
Because project creation is mostly just a name, can we also have the default creation UI simply be a text field of the project name? We can still keep this YAML editor as a second tab/option.

Other resources like Warehouse, AnalysisTemplate, can just be YAML create wizard.

krancour:
@rpelczar this works great, except that I notice the list of projects isn't refreshed after the modal closes.

It's also not surfacing an error on an attempt to create a new project with a project name that already exists... but I've discovered that is a back end issue and I'll fix that right away.

krancour:
> It's also not surfacing an error on an attempt to create a new project with a project name that already exists... but I've discovered that is a back end issue and I'll fix that right away.

I have fixed this in #1649, but when I test this PR on top of #1649, I notice that even though I can see the error being returned now, it still isn't being surfaced in the UI.

krancour (approved):

PR #1627, opened 2024-03-15
chore: address all linter issues

hiddeco:
Discovered a lot of things reported by `golangci-lint run` (without `--new`) were actually due to `t *testing.T` not being used in assertion functions. So figured I would take the opportunity to address using this properly, while also solving the small remainder of other issues.


krancour (approved):
🚀

Issue #1582, opened 2024-03-06
add shard field to StageSpec


krancour:
In #1537 @jessesuen wrote:

> For something this important, I don't totally love relying on a label that, let's be honest, people are going to have to look up... and that's if they're even aware of its existence in the first place.

> I'd be in favor of a separate issue for adding a proper shard field to the StageSpec. The label is still useful for internal reasons, but we could modify the webhook to make the label always match the value in the new shard field.

I agree.

Note that I do not believe other sharded resource types need this to the same extent that Stages do, but I'm willing for that notion to be challenged.

PR #1637, opened 2024-03-15
feat: add convenience shard field on warehouse spec

### LINKED ISSUE #1631: ux: Add shard field to Warehouse
Similar to #1582, we should add a shard field to Warehouses, because it's a pain to look up the correct label.

krancour:
Fixes #1631 

This is similar to #1582, but for Warehouses.


internal/webhook/warehouse/webhook.go
```diff
@@ -61,6 +62,22 @@ func newWebhook(kubeClient client.Client) *webhook {
...
+func (w *webhook) Default(_ context.Context, obj runtime.Object) error {
+	warehouse := obj.(*kargoapi.Warehouse) // nolint: forcetypeassert
+
+	// Sync the convenience shard field with the shard label
+	if warehouse.Spec.Shard != "" {
+		if warehouse.Labels == nil {
+			warehouse.Labels = make(map[string]string, 1)
+		}
+		warehouse.Labels[kargoapi.ShardLabelKey] = warehouse.Spec.Shard
+	} else if warehouse.Labels != nil && warehouse.Labels[kargoapi.ShardLabelKey] != "" {
```
hiddeco, line 74: Nit: I think the `!= nil` check is redundant. As checking the value of a `nil` map is actually safe (xref: https://go.dev/play/p/XiztcCXuNdt), only assignments cause panics.

hiddeco (approved):

internal/webhook/warehouse/webhook.go
krancour: TIL

Fixed in 70d776e8a7a53dbf476593a04c5ef7f0a849e4ab

Issue #1638, opened 2024-03-15
Maintain verification history for a Stage


hiddeco:
This came up during an offline discussion with @krancour while working on #1611.

Currently, a Stage only offers information about the most recent verification performed on the current Freight. However, users now have the ability to rerun verifications for a Stage. Additionally, users may also want to access historical verification information for Freight verified at a previous time.

Considering this, it might be beneficial to implement a `VerificationInfoStack` on the Stage, which would offer a history of verifications for up to N items.

krancour:
I've always believed we needed this and really didn't add it initially because there was no re-triability yet...

I'm going to assign this to the current milestone because this seems like very important companion functionality to the retries.

PR #1645, opened 2024-03-18
feat(controller): record verification history Stage

hiddeco:
Fixes: #1638

This adds a `VerificationHistory` field to the `FreightReference` of a Stage, containing a list of the last 10 verification attempts. Technically, I would say this addition makes the `VerificationInfo` field obsolete. But I kept it for backwards compatibility reasons.

To ensure the AnalysisRun reference for a verification item is always kept around whenever we can (either in the error message, or as `AnalysisRunReference`), aborting an AnalysisRun will now maintain the reference information.


krancour:
> Technically, I would say this addition makes the VerificationInfo field obsolete. But I kept it for backwards compatibility reasons.

I think that's the right move. Stages have a stack of FreightReferences in their history and the top of that stack always matches what's in the CurrentFreight field. So technically, the CurrentFreight field is redundant... but it was useful to be able to see what the _current_ Freight was without having to know with certainty whether the history used stack or queue semantics... I think it's the same thing here with the verifications.

api/v1alpha1/stage_types.go
```diff
@@ -746,6 +749,39 @@ type VerificationInfo struct {
...
+type VerificationInfoStack []VerificationInfo
+
+// UpdateOrAppend updates the VerificationInfo with the same ID as the provided
+// VerificationInfo or appends the provided VerificationInfo to the stack if no
+// such VerificationInfo is found.
+//
+// The order of existing items in the stack is preserved, and new items without
+// a matching ID are appended to the top of the stack. If the stack grows beyond
+// 10 items, the bottom items are removed.
+func (v *VerificationInfoStack) UpdateOrAppend(info ...VerificationInfo) {
```
krancour, line 761: Nit: `UpdateOrPush` might describe the functionality more accurately.

krancour: This is non-blocking.

krancour (approved):
Test drove and it works very smoothly!

Issue #1640, opened 2024-03-15
Distinguish permanent API errors from transient ones


hiddeco:
We do at present not distinguish "not found" errors (permanent) from e.g. "the Kubernetes API server temporary can not be reached" (transient). Because of this, a Stage's verification process may fail prematurely while the controller could _theoretically_ automatically recover it, if given the time.

As manually recovering from it is both cumbersome to a user, and potentially a waste of computing power used by the AnalysisRun. I think we can do a better job at distinguishing these type of errors, and prevent giving up on transient ones by e.g. requeueing and not erasing AnalysisRun references, etc.

xref: https://github.com/akuity/kargo/pull/1611#discussion_r1525229572

---

**Note:** While I have only observed this to happen for a Stage's verification process, this may actually apply to more areas of Kargo.

krancour:
I think we've made progress on this and there's more to be made still, but I think that, like #1479, this is an on-going effort that we can kick from release to release until we feel satisfied.

krancour:
Deferring as this is something we're just tracking until satisfied.

krancour:
@hiddeco do you see this as something we still need to be tracking?

hiddeco:
I believe we have addressed this issue in most places now.

PR #1650, opened 2024-03-19
feat(verification): improve transient error handling

hiddeco:
Addresses (part of) #1640

This improves the handling of transient Kubernetes API errors which may occur during the verification step of a Stage's current Freight.

It does this by selectively returning errors which should be retried, causing the reconciler to not leave the `Verifying` phase and allowing it to retry the action it performed on the next reconciliation attempt.

To allow the reconciler to distinguish a creation error from a failed attempt to retrieve results, closer inspection is performed on the existence of an `AnalysisRun` reference in the `VerificationInfo`.


krancour:
This LGTM after merge conflicts are resolved.

I may be PR'ing something similar for Promotions shortly.

hiddeco:
Has been rebased.

krancour (approved):

PR #1655, opened 2024-03-20
feat(cli): add more commands to display config

### LINKED ISSUE #1614: additional kargo config commands
Currently we can only do `kargo config set-project <project>`.

I'd like if we also had `kargo config view` (view the whole config file) and `kargo config get-project`.

This would make it behave more like `kubectl` overall and would also stop users from having to hunt down and cat their config file just to see what their current default project is.

hiddeco:
Fixes #1614

### DIFF
--- a/internal/cli/cmd/config/view.go
+++ b/internal/cli/cmd/config/view.go
@@ -1,0 +1,98 @@
+package config
+
+import (
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	sigyaml "sigs.k8s.io/yaml"
+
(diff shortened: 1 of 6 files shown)


internal/cli/cmd/config/view.go
```diff
@@ -0,0 +1,97 @@
...
+	io.SetIOStreams(cmd, cmdOpts.IOStreams)
+
+	return cmd
+}
+
+// addFlags adds the flags for the get freight options to the provided command.
+func (o *viewOptions) addFlags(cmd *cobra.Command) {
+	o.PrintFlags.AddFlags(cmd)
+
+	cmd.Flags().BoolVar(&o.RawByteData, "raw", o.RawByteData, "Display raw byte data and sensitive data")
```
krancour, line 65: TIL `kubectl config view` offers this same option. Props for thinking of this.

krancour (approved):
LGTM. I assume you'll amend this to use your new `template.Example()` func. LGTM stands after doing so.

PR #1657, opened 2024-03-20
chore(cli): tidy all command examples

hiddeco:
As discovered in https://github.com/akuity/kargo/pull/1611#pullrequestreview-1937442294

**Note:** Would be wise to get this in before #1655, so I can utilize the new helper there.


krancour (approved):
I really love the approach. The indentation is a nice touch as well, because that was entirely lacking before. Very enthusiastic LGTM!

PR #1658, opened 2024-03-20
refactor(controller): restructure controlplane commands

hiddeco:
This is the "control plane" version of #1547.

While it does move things into a better direction by creating a better separation of concerns, this continues to be a candidate to be revised again when we have time.

### DIFF
--- a/cmd/controlplane/controller.go
+++ b/cmd/controlplane/controller.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


krancour:
At a glance this looks fantastic, and I agree it's a step in the right direction with room for further improvement.

I do want to put this off to v0.6.0 though.

cmd/controlplane/controller.go
```diff
@@ -32,305 +33,367 @@ import (
...
+	ShardName  string
+	KubeConfig string
+
+	ArgoCDEnabled       bool
+	ArgoCDKubeConfig    string
+	ArgoCDNamespaceOnly bool
+
+	RolloutsEnabled      bool
+	RolloutsKubeConfig   string
+	AnalysisRunNamespace string
```
krancour, line 46: This option is gone since #1706

cmd/controlplane/controller.go
```diff
@@ -32,305 +33,367 @@ import (
...
+		// control plane cluster) and project namespaces do not exist in the
+		// shard clusters. We need a place to put them, so for now we allow
+		// the user to specify a namespace that that exists on each shard for
+		// this purpose. Note that the namespace does not need to be the same
+		// on every shard. This may be one of the weaker points in our tenancy
+		// model and can stand to be improved.
+		cacheOpts.DefaultNamespaces = map[string]cache.Config{
+			o.AnalysisRunNamespace: {},
+		}
+	}
```
krancour, line 302: This went away with #1706

cmd/controlplane/controller.go
krancour, line 46: Both of these went away in #1706.

AnalysisRuns now live exclusively in the Kargo control plane, so there's never a need for a separate client nor a need to fall back on using any namespace that isn't a Project's own namespace.

cmd/controlplane/controller.go
```diff
@@ -32,305 +33,367 @@ import (
...
+			Scheme: scheme,
+			Metrics: server.Options{
+				BindAddress: "0",
+			},
+			Cache: cacheOpts,
+		},
+	)
+}
+
+func (o *controllerOptions) setupRolloutsManager(ctx context.Context) (manager.Manager, error) {
```
krancour, line 250: Not needed since #1706

krancour:
Apart from conflicts with #1706, this looks good.

hiddeco:
Think I managed to rebase this correctly, but a careful look with a second pair of 👀 would be much appreciated.

krancour:
@hiddeco it looked good before rebase... I'm just going to run it.

krancour:
Works flawlessly. Startup logs look just like they did before. A few e2e cases all check out.

krancour (approved):
🔥

Issue #1656, opened 2024-03-20
misc doc updates for upcoming v0.5.0


krancour:
I'm still looking for more items, but I found two things already in the docs that need to be updated and were overlooked until now:

- [x] `Freight` and `FreightReference` types have lost their `id` field since #1609, but it still appears in a few places in docs

- [x] Docs mention a `verificationResult` field, which is actually `verificationInfo`. I believe the composition of that type _might_ have changed, and we're also recording `VerificationHistory` now as well since #1645.

- [x] `Freight` has a new `alias` field, which is a convenience over the alias label. #1612

- [x] Warehouse `scanPaths` and `ignorePaths` are a major new feature being introduced in #1477. That PR doesn't contain documentation, so we will have to add it separately.

I'll update this list if/as I find more items.

krancour:
@hiddeco if you have the bandwidth, would you be able to look at this?

hiddeco:
Yes no problem, will tackle this tomorrow morning.

krancour:
Will re-open if more items are found.

PR #1666, opened 2024-03-21
chore(docs): update documentation for v0.5.0

hiddeco:
~~Fixes:~~ Partly addresses: #1656

- Remove mentions of `id` field for Freight.
- Change `verificationResult` to `verificationInfo`.
- Document newly introduced `verificationHistory` field.
- Document `alias` field and label for Freight.
- Small fixes of things I found along the way.

### DIFF
--- a/docs/docs/15-concepts.md
+++ b/docs/docs/15-concepts.md
@@ -170,7 +170,7 @@ For each `Stage`, the Kargo controller will periodically check for `Freight`
 resources that are newly qualified for promotion to that `Stage`.
 
 For any `Stage` subscribed directly to a `Warehouse`, _any_ new `Freight`
-resource from that `Warehouse` is tacitly consider to have been _verified_
+resource from that `Warehouse` is tacitly considered to have been _verified_
 upstream, and is therefore immediately qualified for promotion to such a
 `Stage`.
 
@@ -404,7 +404,7 @@ A `Stage` resource's `status` field records:
 * History of `Freight` that has been deployed to the `Stage`. (From most to
   least recent.)
 
-* The health status any any associated Argo CD `Application` resources.
+* The health status of any associated Argo CD `Application` resources.
 
 * The status of any in-progress of completed verification processes.
 
@@ -414,18 +414,27 @@ For example:
 status:
   phase: Steady
   currentFreight:
-    id: 47b33c0c92b54439e5eb7fb80ecc83f8626fe390
     images:
     - repoURL: nginx
       tag: 1.25.3
     commits:
     - repoURL: https://github.com/example/kargo-demo.git
       id: 1234abc
-    verificationResult:
+    name: 47b33c0c92b54439e5eb7fb80ecc83f8626fe390
+    verificationInfo:
       analysisRun:
         namespace: kargo-demo
         name: test.ab85b188-0ad5-43d9-a36d-ddcf63666183.47b33c0
         phase: Successful
+      id: 69219d8d-cf5e-414e-8ee9-7d3e3a7c3f15
+      phase: Successful
+    verificationHistory:
+    - analysisRun:
+        namespace: kargo-demo
+        name: test.ab85b188-0ad5-43d9-a36d-ddcf63666183.47b33c0
+        phase: Successful
+      id: 69219d8d-cf5e-414e-8ee9-7d3e3a7c3f15
+      phase: Successful
   health:
     argoCDApps:
     - healthStatus:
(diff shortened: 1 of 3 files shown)


docs/docs/15-concepts.md
```diff
@@ -464,11 +476,16 @@ A single `Freight` resource references one or more versioned artifacts, such as:
...
+A `Freight` resource's `metadata.name` field is a SHA1 hash of a canonical
+representation of the artifacts referenced by the `Freight` resource. (This is
+enforced by an admission webhook.) The `metadata.name` field is therefore a
+"fingerprint", deterministically derived from the `Freight`'s contents.
+
+To provide a human-readable identifier for a `Freight` resource, a `Freight`
+resource has an `alias` field and `kargo.akuity.io/alias` label. This alias is
+a human-readable string that is unique within the `Project` to which the
+`Freight` belongs. While it can be set manually (and changed), by default,
+Kargo will automatically generate a unique alias for each `Freight` resource.
```
krancour, line 488: I think we may need an info box or something to mention that the alias field is a convenience (so no one ever needs to go looking for the label) and maybe something about precedence:

- If field is specified, webhook syncs the label value to the field value
- If field is unspecified, webhook syncs the field field value to the label value

fwiw, that means removing an alias _entirely_ requires you to clear both of these, but I think that is something people will rarely, if ever, do. We probably don't need to call that out. 🤷

docs/docs/20-quickstart.mdx
```diff
@@ -300,7 +300,7 @@ the previous section.
 
 1. Next, we'll create several Kargo resource:
 
-    1. A `Project`, which, when reconciled, will effect all boilerplate project
+    1. A `Project`, which, when reconciled, will affect all boilerplate project
```
krancour, line 303: "effect" was actually correct here.

https://www.merriam-webster.com/grammar/affect-vs-effect-usage-difference

Sorry... there's no convenient anchor on that page, but refer to the "A Few Rare Exceptions" section.

docs/docs/30-how-to-guides/15-working-with-freight.md
```diff
@@ -110,12 +93,12 @@ f5f87aa23c9e97f43eb83dd63768ee41f5ba3766   mortal-dragonfly   35s
...
 The Kargo UI, to make efficient use of screen real estate, displays aliases
-only, but a `Freight` resource's `id` can always be discovered by hovering over
-its alias.
+only, but a `Freight` resource's `name` can always be discovered by hovering
+over its alias.
 :::
 
 :::note
-Some, but not all, Kargo CLI commands will accepts `Freight` aliases as an
+Some, but not all, Kargo CLI commands will accept `Freight` aliases as an
```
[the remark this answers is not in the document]
krancour: https://github.com/akuity/kargo/pull/1561

krancour:
Overall, great! I made a few comments, but nothing huge.

I did add some new items to #1656 and will probably continue doing so, so this might not entirely close #1656.

docs/docs/15-concepts.md
hiddeco: As this is a concepts page, I was worried about putting the information in the wrong place. To solve this, I have added it to the section about updating aliases and added references to this page instead.

I think this would potentially be good to do in other places as well in the future, to ensure we reduce the number of places where things are documented twice.

docs/docs/15-concepts.md
```diff
@@ -464,11 +476,22 @@ A single `Freight` resource references one or more versioned artifacts, such as:
...
+
+:::note
+For more information on aliases, refer to the [aliases](./30-how-to-guides/15-working-with-freight.md#aliases)
+and [updating aliases](./30-how-to-guides/15-working-with-freight.md#updating-aliases)
+sections of the "Working with Freight" how-to guide.
+:::
```
krancour, line 494: 👍

docs/docs/30-how-to-guides/15-working-with-freight.md
```diff
@@ -138,7 +120,21 @@ kargo update freight \
...
+
+:::info
+The `alias` field is a convenient way to update the `Freight` resource's
+`kargo.akuity.io/alias` label, which causes a webhook to sync the field value
+to the label value. The precedence rules for syncing between the field and
+label values are as follows:
+
+- If the field is specified, the webhook syncs the value to the label value.
+- If the field is unspecified, the webhook syncs the label value to the field
+  value.
```
krancour, line 133: This feels backwards to me.

> If the field is specified, the webhook syncs the value to the label value.

This sounds to me like the label's value is assumed by the field.

> If the field is unspecified, the webhook syncs the label value to the field
  value.

This sounds to me like the label assumes the field's value.

I think "sync" might be interpreted differently by native English speakers v ESL speakers, or may even interpreted differently _within_ each of those groups.

I propose the following:

```suggestion
- If the field has a non-empty value, the label will assume the field's value.
- If the field has an empty value, the field will assume the label's value.
  value.
```

hiddeco: Reading it twice, it also actually feels backwards to me now as well due to the _"**to** the [..] value"_.

I will update it with your suggestions to avoid this.

docs/docs/30-how-to-guides/15-working-with-freight.md
```diff
@@ -138,7 +120,21 @@ kargo update freight \
...
+
+:::info
+The `alias` field is a convenient way to update the `Freight` resource's
+`kargo.akuity.io/alias` label, which causes a webhook to sync the field value
+to the label value. The precedence rules for syncing between the field and
+label values are as follows:
+
+- If the field has a non-empty value, the label will assume the field's value.
+- If the field has an empty value, the field will assume the label's value.
+  value.
```
krancour, line 133:
```suggestion
```

krancour: I hope you don't mind that I fixed this. Figured it was not-controversial and would speed up approval. LGTM

krancour (approved):
📚  Thank you for this!

PR #1671, opened 2024-03-22
feat(chart)!: manage CRDs as part of release

hiddeco:
This introduces a new change to the chart which allows the Custom Resource Definitions to be managed as part of the release, effectively allowing upgrades. Which is not possible with [Helm's built-in Custom Resource Definition management](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds).

The behavior can be opted-out of by setting `crds.create` to `false`, causing them to be excluded from the Helm release manifests which effectively causes a fallback to Helm's default behavior.

With this change, existing installations will have to adopt the Custom Resource Definitions by configuring the following annotations and label on the resources:

```sh
RELEASE=<Helm release name>
NAMESPACE=<Helm release namespace>

for name in $(kubectl get crds -o name | grep kargo.akuity.io); do
   echo "Adopting $name"
   kubectl annotate $name meta.helm.sh/release-name=$RELEASE
   kubectl annotate $name meta.helm.sh/release-namespace=$NAMESPACE
   kubectl label $name app.kubernetes.io/managed-by=Helm
done
```

### DIFF
--- a/charts/kargo/templates/crds.yaml
+++ b/charts/kargo/templates/crds.yaml
@@ -1,0 +1,5 @@
+{{- if .Values.crds.install -}}
+{{- range $path, $bytes := .Files.Glob "crds/*.yaml" }}
+{{ $.Files.Get $path }}
+{{- end }}
+{{- end }}
--- a/charts/kargo/templates/crds.yaml
+++ b/charts/kargo/templates/crds.yaml
@@ -1,5 +1,11 @@
 {{- if .Values.crds.install -}}
 {{- range $path, $bytes := .Files.Glob "crds/*.yaml" }}
-{{ $.Files.Get $path }}
+{{- $manifest := $.Files.Get $path | fromYaml }}
+{{- if $.Values.crds.keep }}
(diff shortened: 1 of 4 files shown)


charts/kargo/templates/crds.yaml
```diff
@@ -0,0 +1,11 @@
...
+{{- range $path, $bytes := .Files.Glob "crds/*.yaml" }}
+{{- $manifest := $.Files.Get $path | fromYaml }}
+{{- if $.Values.crds.keep }}
+{{- $newAnnotations := dict "helm.sh/resource-policy" "keep" | merge $manifest.metadata.annotations }}
+{{- $_ := set $manifest.metadata "annotations" $newAnnotations }}
+{{- end }}
+{{- $manifest | toYaml  }}
+{{- print "\n---\n" }}
+{{- end }}
+{{- end }}
```
krancour, line 11: This is super slick.

krancour:
This LGTM and my only thought for a possible improvement is probably better left for a follow-up... we could make your little adoption script available in the project's `hack/` dir.

krancour (approved):

PR #1673, opened 2024-03-22
chore(chart): provide CRD adoption script

hiddeco:
Follow up on #1671 as suggested in https://github.com/akuity/kargo/pull/1671#issuecomment-2015162352, but a bit smarter.


krancour (approved):

krancour:
Test drove it and it works quite nicely.

PR #1682, opened 2024-03-26
feat(api): add GetAnalysisRun endpoint

hiddeco:


cmd/controlplane/api.go
```diff
@@ -135,48 +153,139 @@ func newClientForAPI(ctx context.Context, r *rest.Config, scheme *runtime.Scheme
...
+	serverCfg config.ServerConfig,
+) (kubernetes.Client, error) {
+	kubeClientOptions := kubernetes.ClientOptions{
+		NewInternalClient: func(context.Context, *rest.Config, *runtime.Scheme) (client.Client, error) {
+			return internalClient, nil
+		},
+	}
+	if serverCfg.OIDCConfig != nil {
+		kubeClientOptions.GlobalServiceAccountNamespaces = serverCfg.OIDCConfig.GlobalServiceAccountNamespaces
+	}
```
hiddeco, line 289: @krancour I could not figure out if this is applicable to the Argo Rollouts specific client.

charts/kargo/templates/api/configmap.yaml
```diff
@@ -60,4 +60,10 @@ data:
   ARGOCD_URLS: {{ range $key, $val := .Values.api.argocd.urls }}{{ $key }}={{ $val }},{{- end }}
   {{- end }}
   ROLLOUTS_INTEGRATION_ENABLED: {{ quote .Values.api.rollouts.integrationEnabled }}
+  {{- if .Values.controller.rollouts.integrationEnabled }}
```
krancour, line 63: It bugs me that we're using a `.Values.controller...` setting for an API server concern...

I suspect we may eventually remove support for two separate clients for Kargo control plane and Rollouts control plane, so I wouldn't waste time fixing this. Just calling out that I recognize the peculiarity of this and think we can let it slide for now.

krancour: Actually, I take that back. It's easy enough to fix this. Sorry for the flip-flopping.

krancour: Also, we'll still need this flag even if we remove support for the second client, so ya... any initial thought I had that this was temporary was completely off.

cmd/controlplane/api.go
krancour: Oh wow. Good question. This throws a monkey wrench into things... 🤦‍♂️ 

Question is... how do can we ask this wrapper to perform a subject access review when the SAs live in one cluster (Kargo control plane) and the Rollouts control plane lives in another cluster...

I think the answer is, we don't. It would be too complicated and too dangerous to have RBAC in one cluster dictating permissions in another cluster.

I was already fairly certain that we'd eventually kill off the separate Rollouts client, and I think this is the impetus that seals the deal.

I say rip out the extra client... back this branch up just in case.

@gdsoumya just want to check that you agree with me here. This is why I had pinged you on Slack... in the controller, we left a separate client for Rollouts, acknowledging we probably wouldn't need it. Here's @hiddeco has been _adding_ a client for Rollouts to the API server, but that's starting to make less and less sense and present new problems. Would you agree we should just not do this?

charts/kargo/templates/api/configmap.yaml
hiddeco: Is there a scenario in which API and controller are configured differently? If not, we could move `.rollouts` up under `.Values` itself, with fallback to the previous fields where applicable.

hiddeco: Addressed the remaining `.controller` reference by amending c885b167f1f660f1f194422a42d7eba068b75df3

cmd/controlplane/api.go
hiddeco: Ripped it out again in 8c055fa69ff05f0e8a91497216b10cb493663685

charts/kargo/templates/api/configmap.yaml
krancour: > Is there a scenario in which API and controller are configured differently?

In a single Helm release... no.

API server is only ever installed on the control plane. Controllers can be more distributed, but any API server and controller that are co-located would probably agree on what integrations are enabled.

I agree we could move it up, although it is a breaking change... 🤔

internal/api/server.go
```diff
@@ -136,12 +144,14 @@ func NewServer(
 	cfg config.ServerConfig,
 	kubeClient kubernetes.Client,
 	internalClient client.Client,
+	rolloutsEnabled bool,
```
krancour, line 147: I suspect longer term we will find more things we need to pass in and might opt for some kind of config object here, but I think this is suitable for now.

krancour (approved):

internal/api/server.go
hiddeco: Yes, or `ServerConfig` being utilized better. As I am not sure what to think of passing config _and_ options.

krancour: 🤦‍♂️ I overlooked the `ServerConfig` that was already there... I might prefer if this bool were a field in there instead of loose.

hiddeco: I'll do a quick follow up tomorrow as I initially didn't want to touch it due to it dealing with loading user defined settings.

But I think a way I do feel comfortable with is to overwrite / downgrade these settings if they are not feasible (due to e.g. the CRDs not being present).

krancour: It's probably possible though for the CRDs to be present on the cluster and the integration is still not desired.

PR #1683, opened 2024-03-26
feat: add v0.4.x --> v0.5.x upgrade reconcilers

### LINKED ISSUE #1681: proposal: v0.4.x --> v0.5.x upgrade tool
There are a few breaking changes in v0.5.0 that are somewhat difficult to contend with.

1. CRDs are now fully managed by the chart. This means you need to "adopt" existing CRDs. @hiddeco already wrote a script for that that's in the `hack/` dir.

1. `Freight` has lost its `ID` field. This isn't actually a problem. Corresponding to this, however, the `ID` field of the `FreightReference` type was renamed to `Name`.

1. `Freight` no longer have an ownership reference to the `Warehouse` that created them. They have a `Warehouse` field instead. This is used in building a "freight by warehouse" index that is used in several places. Notably, it's used for finding available `Freight` for any `Stage` that subscribes _directly_ to a `Warehouse`.

I propose the following to pave the way for these changes in existing Kargo installations:

1. Put the `ID` field back in the `FreightReference` type. We can allow it to co-exist with the `Name` field until v0.6.0.

1. Implement a binary to be run by the same use who run `helm upgrade`. If we assume this, then we can also assume they have sufficient permissions to permit that binary to do all of the following:

    1. Find all existing Kargo CRDs and add necessary labels and annotations to indicate that they are managed by Helm.

    1. Step through all `Stage`s cluster-wide. For each `FreightReference` (`CurrentFreight` and `History`) copy the value of the `ID` field to the `Name` field.

    1. Step through all `Freight` to copy the name of the owning `Warehouse` from the ownership reference to the new `Warehouse` field, then delete the ownership reference. The Kargo webhook server ordinarily enforces the immutability of `Freight`. The `ValidatingWebhookConfiguration` that enforces that would have to be temporarily patched.

One limitation of this approach is that it doesn't especially help anyone who is gitops'ing their Kargo installation.

To that end, a second alternative is to build most of the same functionality (CRD change wouldn't be possible) into a throwaway controller -- or even add it temporarily to the startup logic of the existing management controller since, even in a sharded topology, that controller runs exclusively on the control plane. (This would be preferable to ensuring that many distributed controllers _all_ handle upgrades of the resources they reconcile.)

__Edit:__ Since all of this work is temporary, I also believe we should perform it in the `release-0.5` branch after it diverges from `main`. This way `main` isn't polluted with these short-term concerns.

krancour:
Fixes #1681

__Note:__ This PR is currently using a placeholder branch equivalent (for the moment) to `main` to avoid accidental merge to `main`. My intention is that this change is never merged to `main` and is merged _only_ into the `release-0.5` branch once it exists. (We're not ready for that branch to diverge from `main` just yet.)

These changes, in conjunction with #1673 provide a smooth upgrade path from v0.4.x to v0.5.x.

How it works __(revised 2024-03-28)__:

* Starts new "upgrade reconcilers" in the management controller. The management controller is guaranteed to run only in the Kargo control plane. It handles upgrades for non-sharded types that are not reconciled by the normal controller(s):
    * Updates old `Freight` that _lack_ a special v0.5.0 compatibility label:
        * Copies value from `kargo.akuity.io/alias` label to new `Alias` field
        * Copies Warehouse name found in an owner ref to new `Warehouse` field
        * `Warehouse` field is temporarily mutable just for v0.5.x
        * Only does patches to avoid racing with with the `Stage` reconciler and API server, both of which may patch `FreightStatus` with approval or verification details.
        * Adds v0.5.0 compatibility label to avoid further reconciliation
    * Updates credentials (`Secret`s) with _old_ label (`kargo.akuity.io/secret-type`)
        * Transforms values in `Data` field
        * Adds new label (`kargo.akuity.io/cred-type`)
        * Removes old label (`kargo.akuity.io/secret-type`) to prevents further reconciliation
* Adds logic to existing reconcilers in regular controller(s):
    * Updates old `Stage`s that _lack_ a special v0.5.0 compatibility label:
        * Copies value from `kargo.akuity.io/shard` label to new `Shard` field
        * Copies `ID` --> `Name` for all `FreightReference`s (`CurrentFreight` and `History` fields of `StageStatus`).
        * Adds v0.5.0 compatibility label to avoid further reconciliation
        * Requeus `Stage` for normal processing
    * Updates old `Warehouse`s that _lack_ a special v0.5.0 compatibility label:
        * Copies value from `kargo.akuity.io/shard` label to new `Shard` field
        * Adds v0.5.0 compatibility label to avoid further reconciliation
        * Requeus `Warehouse` for normal processing

This PR is broken into narrowly-scoped commits that should be easiest to review individually.


internal/controller/management/upgrade/credentials.go
```diff
@@ -0,0 +1,123 @@
...
+
+const (
+	SecretTypeLabelKey  = "kargo.akuity.io/secret-type" // nolint: gosec
+	repoLabelValue      = "repository"
+	repoCredsLabelValue = "repo-creds" // nolint: gosec
+)
+
+// credentialsReconciler reconciles credentials (Secrets) to upgrade them from
+// v0.4.0-compatible to v0.5.0-compatible.
+type credentialsReconciler struct {
```
hiddeco, line 28: My worry with updating the Secrets for the user is that you end up fighting a GitOps agent.

Given this, I am wondering if supporting both formats while emitting some sort of warning while the "legacy" format is encountered wouldn't be the better approach.

internal/controller/management/upgrade/common.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const v050CompatibilityLabelKey = "kargo.akuity.com/v0.5.0-compatible"
```
hiddeco, line 14: How would we delete this label in the future?

hiddeco (commented):
I am wondering if the simpler approach over introducing new reconcilers would be to add logic at the beginning of existing reconcilers to conditionally handle the migration of data, after which it would requeue.

internal/controller/management/upgrade/common.go
krancour: I had been thinking we don't care if it stays.

internal/controller/management/upgrade/credentials.go
krancour: > My worry with updating the Secrets for the user is that you end up fighting a GitOps agent.

That's a really good point.

> Given this, I am wondering if supporting both formats while emitting some sort of warning while the "legacy" format is encountered wouldn't be the better approach.

The logical place to surface this would be in a Warehouse's `status.Message`. If things are working, I don't think people will ever look there. They'll never see the warning.

What would you think about removing this conversion due to the gitops issue and letting things _fail_ when the old format is encountered? That seems more likely to prompt someone to look at the Warehouse's `status.Message` field.

This means we need to at least _look_ for older credentials, but don't need to use them.

krancour:
> I am wondering if the simpler approach over introducing new reconcilers would be to add logic at the beginning of existing reconcilers to conditionally handle the migration of data, after which it would requeue.

That is a good thought, but Freight had no reconciler at all prior to this and in a sharded topology, migrating all Stages would require upgrading _every_ controller. I don't want to take it for granted that would occur in a timely fashion, so I'd feel better having all of this handled within the control plane.

internal/controller/management/upgrade/credentials.go
krancour: > My worry with updating the Secrets for the user is that you end up fighting a GitOps agent.

Alternatively, that may not be the end of the world. People will notice that and fix it.

krancour: @jessesuen do you have any opinions about upgrading credential formats?

hiddeco: I would think people look at their logs at least once in a while in combination with reading the release notes, making this the least invasive and user friendly approach in my opinion.

But I can live with any option you think is good. Just wanted to highlight it as an alternative.

krancour: > in combination with reading the release notes

☝️ Leveraging those might be my own preference.

Instead of "Hey, we broke all your creds." It's, "Hey, we broke all your creds and then fixed them. If you're gitops'ing your projects, you might see some odd behavior, but you know how to fix it."

Overall, since we're not GA, I don't want to _overthink_ the upgrade path from v0.4.x to v0.5.x. I'm happy as long as we don't leave anyone in a degraded state that they cannot bound back from and I think we've at least ticked those boxes.

jessesuen: > @jessesuen do you have any opinions about upgrading credential formats?

I don't have as much of a strong opinion here, since credentials are an easier and much more obvious thing to manually fix, than how we broke freights. Migration of credentials would be nice to have, IMO but not as required as the other stuff we broke.

krancour: I mean... it's done already. It's purely a matter of whether you think we ought _not_ touch secrets, which is a perspective I could understand.

internal/controller/management/upgrade/freight.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	if freight == nil {
+		// Ignore if not found. This can happen if the Freight was deleted after the
+		// current reconciliation request was issued.
+		return ctrl.Result{}, nil // Do not requeue
+	}
+
+	// None of these things should really happen
+	if len(freight.OwnerReferences) == 0 {
+		logger.Warning("skipping Freight with no OwnerReferences")
```
hiddeco, line 63: If you have existing (or new?) Freight created by a `v0.5.0` release (or current `main`), this will actually (continuously) happen each time the management controller is started:

```
time="2024-03-28T12:03:43Z" level=warning msg="skipping Freight with no OwnerReferences" freight=319c6e315e081836c1ef597b08cc22811c84af50 namespace=kargo-advanced
time="2024-03-28T12:03:43Z" level=debug msg="reconciling Freight" freight=7baad20879af2b1c2e1c039120d52ca60cdf0218 namespace=kargo-advanced
time="2024-03-28T12:03:43Z" level=warning msg="skipping Freight with no OwnerReferences" freight=7baad20879af2b1c2e1c039120d52ca60cdf0218 namespace=kargo-advanced
```

krancour:
@hiddeco large chunks of this are re-written based on your feedback.

hiddeco (approved):
Think we ended up in a great middle ground with this updated version, and I can't find any new pitfalls in the new approach. Test drove it as well, and it appears to do what's expected from it.

krancour:
Thank you @hiddeco. btw... idk about Flux, but I verified through experimentation that if you're managing Kargo projects via Argo CD, it _won't_ fight with the upgrade reconcilers over any of these changes. Thanks also to @jessesuen for explaining why that is so. Much appreciated.

hiddeco:
> idk about Flux, but I verified through experimentation that if you're managing Kargo projects via Argo CD, it won't fight with the upgrade reconcilers over any of these changes

Flux would in cases where fields are removed and still while still being declared in the Git (with a few exceptions, if configured), as that's the ultimate source of truth.

hiddeco (approved):

krancour:
@hiddeco would you mind looking at https://github.com/akuity/kargo/pull/1683/commits/800ef8ed3e103692d7bfe65ef3a921609dfb9940 and letting me know if you think it's sensible?

I discovered through extensive testing that if attempting to patch resources with values for _new_ fields before the CRDs were updated, it was bad news.

This change seems to resolve it nicely, but a sanity check is always good.

hiddeco:
My only worry with the change would be that the back-off from the error would be too rapid at the beginning, and then too much delayed later on, depending on _why_ it takes some time for the CustomResourceDefinition to be updated. Given this, I wonder if it would make sense to log the error instead and requeue on a fixed interval.

krancour:
Experimentation shows this all resolves very quickly, but I don't mind changing it to a fixed interval.

This race is _probably_ not going to occur in the wild anyway, because there's no way the CRDs don't update faster than the new image can be pulled. I've managed to run into that condition a number of times simply by switching back and forth between v0.4.5 and main. Our Tiltfile is tunes to replace and restart binaries without rebuilding images and restarting containers, and this also takes advantage of the local Go build cache. Long story short... my controller sometimes upgrades faster than my CRDs. I doubt this happens anywhere other than in local dev, so it's really done out of an abundance of caution.

Also, it has been a little bit of a surprise that patching a field that doesn't exist (which is what happens if the CRDs aren't updated yet) succeeds without error _and_ without the intended effect, which is another reason this is necessary.

hiddeco (approved):

krancour:
fyi... recent force push is only because I rebased on `release-0.5` and re-ran codegen.

PR #1689, opened 2024-03-27
fix: display warehouses even if no stages are subscribed to it

rbreeze:


krancour (approved):

krancour:
I rebased this on `main` before test-driving it. It works perfectly, but I do notice that when I create a new Warehouse from the UI, it doesn't show up right away. If I refresh, it _does_ show up -- meaning this PR does what it was meant to do.

I'll open a separate issue for getting the view to refresh after Warehouse creation.

krancour:
https://github.com/akuity/kargo/issues/1697

PR #1691, opened 2024-03-27
feat(api): add get, list and delete endpoints for AnalysisTemplates

### LINKED ISSUE #1688: add get, list, and delete endpoints for Rollouts AnalysisTemplates
Required to enable #1686

hiddeco:
Depends on #1690
Required for #1686
Fixes #1688


api/service/v1alpha1/service.proto
```diff
@@ -68,13 +69,18 @@ service KargoService {
...
 
   /* Credential APIs */
+
   rpc CreateCredentials(CreateCredentialsRequest) returns (CreateCredentialsResponse);
   rpc DeleteCredentials(DeleteCredentialsRequest) returns (DeleteCredentialsResponse);
   rpc GetCredentials(GetCredentialsRequest) returns (GetCredentialsResponse);
   rpc ListCredentials(ListCredentialsRequest) returns (ListCredentialsResponse);
   rpc UpdateCredentials(UpdateCredentialsRequest) returns (UpdateCredentialsResponse);
 
   /* AnalysisRun APIs */
```
krancour, line 79: Nit: Since these endpoints are for both runs and templates now, we could change this:

```suggestion
  /* Analysis APIs */
```

internal/api/delete_analysistemplate_v1alpha1.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func (s *server) DeleteAnalysisTemplate(
+	ctx context.Context,
+	req *connect.Request[svcv1alpha1.DeleteAnalysisTemplateRequest],
+) (*connect.Response[svcv1alpha1.DeleteAnalysisTemplateResponse], error) {
+	if !s.cfg.RolloutsIntegrationEnabled {
+		return nil, connect.NewError(
+			connect.CodeUnimplemented,
+			errors.New("Argo Rollouts integration is not enabled"),
```
krancour, line 21: 👍

internal/api/delete_analysistemplate_v1alpha1.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil, err
+	}
+	if at == nil {
+		err = fmt.Errorf("AnalysisTemplate %q not found in namespace %q", name, project)
+		return nil, connect.NewError(connect.CodeNotFound, err)
+	}
+
+	if err := s.client.Delete(ctx, at); err != nil {
+		return nil, fmt.Errorf("delete AnalysisTemplate: %w", err)
+	}
```
krancour, line 53: Do we have to get before delete? Can we just blindly delete? If it's not found, the error interceptor will correctly interpret the error and the client will still get end up receiving `connect.CodeNotFound`.

krancour:
Just some very minor things... preemptively approving.

krancour (approved):

internal/api/delete_analysistemplate_v1alpha1.go
hiddeco: Well, this matches what my intuition was telling me as well. But I then looked at the other delete commands and noticed they did a get before deleting the object, so I followed their example instead.

hiddeco: It's actually done in the right way in other places as well, except for things that have an alias (where we need to look up, although it _could_ be more efficient if only a name is provided). Given this, I addressed it in [c6bf4b8](https://github.com/akuity/kargo/pull/1691/commits/c6bf4b80e0343634b3287699020abe4f42519970)

Issue #1684, opened 2024-03-26
helm chart subscriptions should ignore tags that are not semver


krancour:
In #1678, @hoopty noted that it's possible for OCI repos being used to store Helm charts to defy the expectation that the repository only contains tags that are valid semantic versions.

When this is the case, an error like the following may result:

```
error getting latest Freight from repositories: error syncing chart repo subscriptions:
error searching for latest version of chart in repository "oci://registry-1.docker.io/bitnamicharts/rabbitmq":
error determining latest version of chart "" from repository "oci://registry-1.docker.io/bitnamicharts/rabbitmq":
error parsing version "sha256-4730c3c5fe00ed4cf6ffd139ed081f5a27342a959fbc89ef53d2b5c5341ff7af":
Invalid Semantic Version
```

This issue supersedes #1678 and calls for chart subscription logic to ignore tags that are not parsable as semantic versions.

This is a very small change right here:

https://github.com/akuity/kargo/blob/0b43b26e0d3c5e01b57b1db2914cf9e67e5408d7/internal/helm/helm.go#L176-L181

jacobnguyenn:
@krancour can i take this one :)

krancour:
@jacobnguyenn all yours. Thank you!

PR #1695, opened 2024-03-27
fix: helm chart subscriptions ignore tags that are not semver

jacobnguyenn:
fixes https://github.com/akuity/kargo/issues/1684.

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -172,12 +172,14 @@ func getChartVersionsFromOCIRepo(
 // version will be returned. The empty string will be returned when the provided
 // list of versions is nil or empty.
 func getLatestVersion(versions []string, constraintStr string) (string, error) {
-	semvers := make([]*semver.Version, len(versions))
-	for i, version := range versions {
-		var err error
-		if semvers[i], err = semver.NewVersion(version); err != nil {
-			return "", fmt.Errorf("error parsing version %q: %w", version, err)
+	semvers := make([]*semver.Version, 0, len(versions))
+	for _, version := range versions {
+		if semverVersion, err := semver.NewVersion(version); err == nil {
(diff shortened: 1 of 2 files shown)


internal/helm/helm.go
```diff
@@ -172,13 +172,15 @@ func getChartVersionsFromOCIRepo(
 // version will be returned. The empty string will be returned when the provided
 // list of versions is nil or empty.
 func getLatestVersion(versions []string, constraintStr string) (string, error) {
-	semvers := make([]*semver.Version, len(versions))
-	for i, version := range versions {
-		var err error
-		if semvers[i], err = semver.NewVersion(version); err != nil {
-			return "", fmt.Errorf("error parsing version %q: %w", version, err)
+	var semvers []*semver.Version
```
krancour, line 175: Since we expect, under normal circumstances, that this _will_ grow to size `len(versions)`, we're typically going to resize this `len(versions)` times.

If we initialize this way, the initial size will remain zero, but the capacity will be `len(versions)`:

```suggestion
	semvers := make([]*semver.Version, 0, len(versions))
```

internal/helm/helm_test.go
```diff
@@ -116,9 +116,17 @@ func TestGetLatestVersion(t *testing.T) {
...
+			},
+		},
+		{
+			name:     "success with invalid version ignored",
+			unsorted: []string{"not-semantic", "1.0.0"},
+			assertions: func(t *testing.T, latest string, err error) {
+				require.NoError(t, err)
+				require.Equal(t, "1.0.0", latest)
 			},
 		},
```
krancour, line 131: I think these two cases can just be combined.

krancour:
Thank you @jacobnguyenn! I added just some very minor feedback. Overall, looks great!

krancour (approved):

PR #1674, opened 2024-03-23
feat: make pr-based promotions work with kargo render

krancour:
Fixes #1330

~~Depends on https://github.com/akuity/kargo-render/pull/268~~

This PR:

* Upgrades us to Kargo Render v0.1.0-rc.38, which has three important, new features:
    * Can take input from a local repository instead of a remote repository
    * Can write output to a local path instead of a branch in a remote repository
    * Does not refuse to manage an existing branch that lacks Kargo Render metadata, _as long as the branch is 100% empty_
* Fixes a bug in Kargo's existing PR-based promotions where the base branch is assumed to exist
    * We now create it if it doesn't
* Make Kargo Render promotion mechanism behave much more like the Helm and Kustomize-based ones:
    * Points Kargo Render at the path of the already-cloned repo, which will already be sitting on the desired commit

Taken together, these remove the limitation of Kargo Render and PR-based promotions not working together.

PR #1696, opened 2024-03-27
always pass credentials to kargo render

krancour:
Prior to #1674 we _did_ always pass credentials to Kargo Render. We stopped because Kargo proper now handles all the checking out and committing -- or so I thought...

I had forgotten that Kargo Render will, even if given a local repo that is already checked out to the correct source commit, check if the target branch exists remotely and attempt to pull from it. It does this because it wants up-to-date metadata from the target branch, if it exists.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,7 +68,7 @@ RUN GRPC_HEALTH_PROBE_VERSION=v0.4.15 && \
 # - supports development
 # - not used for official image builds
 ####################################################################################################
-FROM ghcr.io/akuity/kargo-render:v0.1.0-rc.38 as back-end-dev
+FROM ghcr.io/akuity/kargo-render:v0.1.0-rc.39 as back-end-dev
 
 USER root
 
@@ -103,7 +103,7 @@ CMD ["pnpm", "dev"]
 # - the official image we publish
 # - purposefully last so that it is the default target when building
 ####################################################################################################
-FROM ghcr.io/akuity/kargo-render:v0.1.0-rc.38 as final
+FROM ghcr.io/akuity/kargo-render:v0.1.0-rc.39 as final
 
 USER root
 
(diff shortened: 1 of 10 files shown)


Dockerfile
```diff
@@ -103,7 +103,7 @@ CMD ["pnpm", "dev"]
 # - the official image we publish
 # - purposefully last so that it is the default target when building
 ####################################################################################################
-FROM ghcr.io/akuity/kargo-render:v0.1.0-rc.38 as final
+FROM ghcr.io/akuity/kargo-render:v0.1.0-rc.39 as final
```
krancour, line 106: This fixes a bug where Kargo render sometimes pushed a new, empty target branch to the remote even when instructed _not_ to commit anything.

internal/kargo-render/render.go
```diff
@@ -57,13 +62,19 @@ func RenderManifests(req Request) error { // nolint: revive
...
 		"json",
 	}
 	for _, image := range req.Images {
 		cmdTokens = append(cmdTokens, "--image", image)
 	}
 	cmd := exec.Command(cmdTokens[0], cmdTokens[1:]...) // nolint: gosec
+	cmd.Env = append(
+		os.Environ(),
+		fmt.Sprintf("KARGO_RENDER_REPO_PASSWORD=%s", req.RepoCreds.Password),
+	)
```
krancour, line 77: One of two key changes. Most everything else is just updates to function signatures to get these creds down to here.

internal/controller/promotion/render.go
```diff
@@ -97,6 +99,7 @@ func (r *renderer) apply(
 		Images:       images,
 		LocalInPath:  workingDir,
 		LocalOutPath: writeDir,
+		RepoCreds:    repoCreds,
```
krancour, line 102: One of two key changes. Most everything else is just updates to function signatures to get these creds down to here.

hiddeco (approved):
Tested this with a scenario which was previously failing, and it works great!

PR #1700, opened 2024-03-27
fix(cli): remove duplicate `create project` cmd

hiddeco:
The interesting thing is that this became apparent due to Bash auto-completion showing odd behavior, as it would complete `kargo create project` with the short description added `(Create a project)`.


krancour (approved):
I bet this was a botched attempt at resolving merge conflicts at some point.

PR #1694, opened 2024-03-27
feat(ui): Verifications visibility on stages

rpelczar:
image

image

PR #1705, opened 2024-03-28
feat(api): add `StartTime` field to `VerificationInfo`

hiddeco:
This came up during a chat with @rpelczar, and would allow displaying a timestamp column in #1694. Giving the user a clearer idea about when a verification was (re)run.


api/v1alpha1/stage_types.go
```diff
@@ -719,6 +719,8 @@ type AnalysisRunArgument struct {
 type VerificationInfo struct {
 	// ID is the identifier of the Verification process.
 	ID string `json:"id,omitempty" protobuf:"bytes,4,opt,name=id"`
+	// Timestamp is the time at which the Verification process was initiated.
+	Timestamp *metav1.Time `json:"timestamp,omitempty" protobuf:"bytes,5,opt,name=timestamp"`
```
krancour, line 723: I'm wondering if some field name other than `Timestamp` might be more self-descriptive in terms of _what_ the timestamp represents. Something like `StartTime` maybe?

hiddeco: Yes, I was struggling with finding a good name for this field as `Creation` seemed wrong. But `StartTime` seem a great suggestion.

hiddeco: The only nit I have now, after changing it, is that at times the verification process _technically_ has not started (but e.g. runs into an "immediate" error).

krancour: In the event of an error, I think people are very likely to understand this as the start of an _attempt_.

I'm happy with it _if you are_.

LGTM'ing, but if you want to noodle on it more, feel free.

PR #1706, opened 2024-03-28
refactor(controller)!: remove possibility of separate rollouts manager/client

krancour:
Since we've become committed to AnalysisRuns used for verification purposes always living within the Kargo control plane, the door is now closed on the possibility of Kargo controllers using a separate client for communication with a different cluster that hosts AnalysisRun resources.

This PR removes code that was already rendered dead by that decision.

Note: This _doesn't_ mean AnalysisRuns cannot _execute_ elsewhere now. That continues to be achievable by setting `controller.rollouts.controllerInstanceID` on a Kargo controller.

cc @jessesuen and @gdsoumya


charts/kargo/templates/controller/cluster-roles.yaml
```diff
@@ -93,26 +93,4 @@ rules:
...
-metadata:
-  name: kargo-controller-rollouts
-  labels:
-    {{- include "kargo.labels" . | nindent 4 }}
-    {{- include "kargo.controller.labels" . | nindent 4 }}
-rules:
-- apiGroups:
-  - argoproj.io
-  resources:
-  - analysisruns
```
hiddeco, line 109: I think these permissions are still required by the verification process?

krancour: Odd... this all _worked_. Let me look into that.

krancour: Oh... I removed those last minute _after_ I saw it working. Fixing.

hiddeco (approved):
🚀

PR #1710, opened 2024-03-28
feat: Credentials management page

rbreeze:
List View:
Screenshot 2024-03-28 at 14 20 11

Create:
Screenshot 2024-03-28 at 14 20 19

Edit:
Screenshot 2024-03-28 at 14 20 27

Delete:
Screenshot 2024-03-28 at 14 20 40

With the addition of tabs, I've slightly redesigned and refactored the freightline view as well (the banner text is now outside of the grey area):

Screenshot 2024-03-28 at 14 20 54
Screenshot 2024-03-28 at 14 21 02


jessesuen (approved):
Approval based on screenshots

krancour:
@rbreeze repoURL and repoURL pattern are supposed to be mutually exclusive. Is it at all possible that we could just make this _one_ field with a checkbox for "is regular expression" or something?

rbreeze:
@krancour Oh, there is a front end validation check to make sure that they are mutually exclusive. But I can see how the UI is confusing -- I'll work on an improvement to that part

ui/src/features/project/credentials/delete-credentials-modal.tsx
rpelczar, line 1: Maybe you can use the existing component? `useConfirmModal`

ui/src/features/project/credentials/create-credentials-modal.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+const labelForKey = (s: string) =>
+  s
+    .split('')
+    .map((c, i) => (c === c.toUpperCase() ? (i !== 0 ? ' ' : '') + c : c))
+    .join('')
+    .replace(/^./, (str) => str.toUpperCase())
+    .replace('Url', 'URL');
+
+const constructDefaults = (init?: Secret) => {
```
rpelczar, line 68: Can you move all the above functions to a separate file (utils)?

hiddeco:
> `repoURL` and `repoURLPattern` are supposed to be mutually exclusive. Is it at all possible that we could just make this _one_ field with a checkbox for "is regular expression" or something?

This makes me wonder if we can't do more on the back-end to allow both an "exact" match and regular expression in just the `repoURL` field. Either by allowing e.g. `regexp:<regex>` like we have been doing in #1270, or by automagically detecting special regex characters.

krancour:
> This makes me wonder if we can't do more on the back-end to allow both an "exact" match and regular expression in just the repoURL field. Either by allowing e.g. regexp:<regex> like we have been doing in https://github.com/akuity/kargo/issues/1270, or by automagically detecting special regex characters.

After #1270 set that precedent, I can get down with that. I'll work on that now so it gets into v0.5.0... I _don't_ want to break the credential format twice.

@rbreeze those two fields can be combined now, I guess, but I think a checkbox for "is regex" would still be a nicer UX than requiring people to somehow know they can optionally prefix the URL with "regex:".

rbreeze:
Updated UI to reflect discussed changes:

Screenshot 2024-03-29 at 17 00 21
Screenshot 2024-03-29 at 17 00 29
Screenshot 2024-03-29 at 17 00 34

PR #1712, opened 2024-03-29
refactor!: slight change to credential format

krancour:
This is a small, but breaking UX improvement to the credential format. It combines the formerly separate fields `repoURL` and `repoURLPattern` into a single `repoURL` field and _adds_ a new field `repoURLIsRegex` which indicates the value of the `repoURL` is a regular expression (if `true`) or not (if absent or anything but `true`).

I want this in v0.5.0 so that we don't break the credential format in v0.5.0 and _again_ in some later release.

When this is merged, I will updated the upgrade logic in #1683 accordingly.

cc @hiddeco for review

cc @rbreeze this affects/unblocks work on #1710

cc @gdsoumya I promised I would alert you to any further breaking changes


api/service/v1alpha1/service.proto
```diff
@@ -495,7 +495,7 @@ message UpdateCredentialsRequest {
   string name = 2;
   string type = 3;
   string repo_url = 4 [json_name = "repoURL"];
-  string repo_url_pattern = 5 [json_name = "repoURLPattern"];
+  bool repo_url_is_regex = 5 [json_name = "repoURLIsRegex"];
```
rbreeze, line 498: If regex must be prefixed with `regex:`, is this boolean necessary?

krancour: Granted, it might be a tiny bit leaky. Because now when building an edit/update modal, I suppose you have to go in the other direction and  _look_ for the "regex:" prefix to know whether the checkbox should be pre-checked or not...

Maybe I ought to keep this as a boolean all the way through to the secret instead of using a prefix ever/at all.

hiddeco: What's the advantage of having to remember a field to check for a boolean over having to check a field value for a prefix?

rbreeze: @hiddeco I think the advantage of a field over the prefix is IDE auto-complete and type checking, and simpler logic (no need to parse a string)... @krancour my opinion is to go with the boolean field and no prefix, but I don't feel strongly either way (except that we should choose one or the other and not both)

hiddeco: I am happy to provide a boolean in the API for the reasons mentioned, but I think doing it all the way through the Secret makes you loose the advantage of having to remember a single field to work with while doing things declaratively (which is what my suggestion was aiming to improve).

krancour: > makes you loose the advantage of having to remember a single field to work with while doing things declaratively / without a CLI or UI

That's a fair point.

This is a unique area of the codebase because we're trying superimpose structure and a nicer UX on top of something that is inherently very generic (Secrets). It seems we can either optimize for the portion of the experience we fully control (Kargo CLI and UI) or for the part of it we don't -- i.e. `kubectl` or GitOps'ing a project, where you are forced to work with naked Secrets.

We need to make a decision and I am very narrowly ruling in favor of optimizing for the part we control. By a narrow margin, I think it's the option that keeps more logic out of the clients and just in the back end.

Also, I think people working with naked Secrets are going to end up having to consult docs regardless, so not knowing off the top of their heads that there's a separate field to indicated "regexiness" isn't likely to be a huge problem.

hiddeco: > Also, I think people working with naked Secrets are going to end up having to consult docs regardless, so not knowing off the top of their heads that there's a separate field to indicated "regexiness" isn't likely to be a huge problem.

I think in combination with the subscription filtering feature, it's quite easy to remember you have to just prefix something with `regex:` once you have used it once or twice. Which is why I am generally in favor of "pick one way of doing things and stick with it".

But I come from a world where Kubernetes API design (and the YAML) was the core foundation of everything we built, so my opinion/view may differ wildly here from Kargo's world where things tend to be more CLI/UI driven.

internal/api/create_credentials_v1alpha1.go
```diff
@@ -18,7 +19,7 @@ type credentials struct {
 	name           string
 	credType       string
 	repoURL        string
-	repoURLPattern string
+	repoURLISRegex bool
```
hiddeco, line 22: Nit:

```suggestion
	repoURLIsRegex bool
```

krancour: That was a mistake. Thanks.

Will change it locally so I can use ID "rename symbol" functionality.

krancour: Fixed in https://github.com/akuity/kargo/pull/1712/commits/11744d906112ac41a61f4d801c6d91af4e5cc990

internal/cli/cmd/create/credentials.go
```diff
@@ -50,7 +50,7 @@ func newCredentialsCommand(cfg config.CLIConfig, streams genericiooptions.IOStre
 	cmd := &cobra.Command{
 		Use: `credentials [--project=project] NAME \
     (--git | --helm | --image) \
-    (--repo-url=repo-url | --repo-url-pattern=repo-url-pattern) \
+    --repo-url=repo-url [--regex] \
```
hiddeco, line 53: This really smoothens the CLI flag usage.

hiddeco (approved):

PR #1715, opened 2024-03-29
fix(cli): restore ability to read manifests from stdin

### LINKED ISSUE #1713: can no longer pipe inline manifest into kargo apply command
@dhpup has reported the following no longer works on CLI built from `main` and I can confirm:

```
cat <<EOF | kargo apply -f -
kind: Project
metadata:
  name: kargo-demo
EOF
```

```
Error: read manifests: lstat /Users/krancour/Code/akuity/kargo/-: no such file or directory
```

@hiddeco -- wondering if you could look at this?

krancour:
Fixes #1713 and reverts/replaces what was done in #1539


hiddeco (approved):
I count not depending on `kubectl` still as a major improvement, even if this is a second attempt 💯

PR #1718, opened 2024-03-30
fix(ui): password should be optional when editing credentials

### LINKED ISSUE #1717: ui: edit credentials modal doesn't need to require password
The way the update credentials API works, it will only update fields that you you include in the update request.

So if a password isn't passed, it will just keep the old one.

This means we don't have to require it in the edit modal.

Is there a way we could visually indicate in the password field that a password _exists_ without actually populating the field?

This way, the user would be free to populate the field with a new value if they choose, but they could also leave it blank, which would keep the existing password.

This doesn't have to get into v0.5.0

rbreeze:
Fixes #1717


krancour:
@rbreeze what would you think about using `••••••••` as the help/example text instead of `admin12345`?

I am hoping / wondering if that will make it more clear that there's an existing value and we're just not showing it to you.

Mocked up:

Screenshot 2024-03-30 at 6 47 28 PM

rbreeze:
Good idea! I'll make that change

Issue #1628, opened 2024-03-15
add FreightReference to PromotionStatus


krancour:
Since the lifecylce of Freight and Promotions are fully decoupled from one another, it is possible that an older Promotion may stick around longer than some older piece of Freight that it references.

If we add a FreigthReference field to PromotionStatus, every Promotion will retain some details of the Freight it referenced, even with that Freight having been deleted.

This is similar to how/why we keep FreightReferences in StageStatus's History field.

jacobnguyenn:
@krancour  can I take this one?

krancour:
@jacobnguyenn it's yours! Thank you!

Issue #1613, opened 2024-03-13
Move Argo CD sync & wait logic to promotion


jessesuen:
# Proposed Feature

Currently, the Argo CD sync happens in the promotion, but waiting for the sync operation to complete is currently happening in the Stage reconciler. I believe that the wait should instead be part of the promotion. Instead of the Stage reconciler doing the sync and wait, I propose this logic be moved to the Promotion reconciler.

# Motivation

Earlier, we did not do this because Promotions were not long-lived. But now that they are, I feel it would make more sense for sync + wait to be part of the promotion. This will align better with our future desire for promotions to integrate with other tools aside from Argo CD, which might have asynchronous, long-lived deployment processes.

Moving it to Promotion will make it more visible when the sync failed, because the promotion would fail too.

# Suggested Implementation

* When a promotion job issues a sync, it does not complete the Promotion immediately like it does now. It leaves it in Running.
* We have a predicate for Argo CD such that when an operation completes, we enqueue the promotion job instead of the stage. When the Promotion job sees that it is completed successfully, the promotion is Successful. If the operation fails, the promotion is considered failed.
* We will need a way to terminate Promotions, which in the case of Argo CD, will terminate the operation. This can be a separate issue.

# Open Issues

There is also the question of whether or not waiting for an Application to become Healthy should be part of promotion. I suggest we leave this in Stage reconciler, because this is more like a verification than it is a promotion.

krancour:
I most agree with this, but tentatively think it that checking sync state should happen in _both_ places. While an Application's sync state might be distinct from its health, this feels less true for a Kargo Stage. Detecting that an Application _isn't_ synced to the current Freight indicates the _Stage_ is unhealthy (even if the App is). I feel that's an important check to continue doing periodically.

wdyt?

jessesuen:
> think that checking sync state should happen in both places

I didn't consider moving the sync state check into promotion. I think that would be fine to have that in promotion as well, because we could clearly indicate why the promotion job failed in the status message, and have a record of it.

hiddeco:
Based on a lot of reading today, I believe this is an opportunity to also address #1451.

PR #1721, opened 2024-04-01
feat: add FreightReference to PromotionStatus

jacobnguyenn:
Resolve https://github.com/akuity/kargo/issues/1628


jacobnguyenn:
@krancour can you kindly take a look :)

api/v1alpha1/promotion_types.go
```diff
@@ -96,6 +96,9 @@ type PromotionStatus struct {
 	// Metadata holds arbitrary metadata set by promotion mechanisms
 	// (e.g. for display purposes, or internal bookkeeping)
 	Metadata map[string]string `json:"metadata,omitempty" protobuf:"bytes,3,rep,name=metadata" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
+	// History is a stack of recent Freight. By default, the last ten Freight are
+	// stored.
+	History FreightReferenceStack `json:"history,omitempty" protobuf:"bytes,4,rep,name=history"`
```
krancour, line 101: A Promotion would only ever reference a single piece of Freight.

internal/controller/promotions/promotions.go
```diff
@@ -379,6 +379,7 @@ func (r *reconciler) promote(
 				err,
 			)
 		}
+		newStatus.PromotedFreight = &nextFreight
```
hiddeco, line 382: I believe we could also benefit from having this reference _before_ the promotion in full is actually successful.

For example, to be able to facilitate #1613 in which the Git promotion may produce a "health check commit" which we later want to use to ensure an Argo CD Application has successfully been synced.

krancour: Agreed. This should be captured at the earliest opportunity.

hiddeco:
@jacobnguyenn as this pull request would be a very welcome addition for some other things I want to factor out of the Promotion reconciler, I was wondering if are still interested in incorporating the feedback from https://github.com/akuity/kargo/pull/1721#pullrequestreview-1973563446.

I would otherwise also be happy to do the last bits for you, but wanted to give you a chance to address things yourself. (This doesn't have to be tomorrow, I can wait a little longer! 🍒)

jacobnguyenn:
> @jacobnguyenn as this pull request would be a very welcome addition for some other things I want to factor out of the Promotion reconciler, I was wondering if are still interested in incorporating the feedback from [#1721 (review)](https://github.com/akuity/kargo/pull/1721#pullrequestreview-1973563446).
> 
> I would otherwise also be happy to do the last bits for you, but wanted to give you a chance to address things yourself. (This doesn't have to be tomorrow, I can wait a little longer! 🍒)

thanks @hiddeco , i have been busy the last few days. I will get back to this today :)

internal/controller/promotions/promotions.go
```diff
@@ -389,7 +389,7 @@ func (r *reconciler) promote(
 		)
 	}
 	if targetFreight == nil {
-		return nil, fmt.Errorf("Freight %q not found in namespace %q", promo.Spec.Freight, promo.Namespace)
+		return nil, fmt.Errorf("freight %q not found in namespace %q", promo.Spec.Freight, promo.Namespace)
```
hiddeco, line 392: Even though linters complain about it, "Freight" is on purpose here because it's the name of the object kind (same applies to other places where this was changed).

api/v1alpha1/promotion_types.go
```diff
@@ -96,6 +96,8 @@ type PromotionStatus struct {
 	// Metadata holds arbitrary metadata set by promotion mechanisms
 	// (e.g. for display purposes, or internal bookkeeping)
 	Metadata map[string]string `json:"metadata,omitempty" protobuf:"bytes,3,rep,name=metadata" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
+	// PromotedFreight is the detail of the piece of freight that was referenced by this promotion.
+	PromotedFreight *FreightReference `json:"promotedFreight,omitempty" protobuf:"bytes,4,opt,name=promotedFreight"`
```
hiddeco, line 100: On the fence about the naming of this field, as it is also set when the promotion has not completed and/or failed. While the current name may make you think it's the result of a successful promotion.

Given this, I wonder if it maybe should just be named `Freight`?

cc: @krancour

internal/controller/promotions/promotions.go
```diff
@@ -422,6 +421,13 @@ func (r *reconciler) promote(
...
 	}
+	// record the freight reference to the promotion's status
+	err = kubeclient.PatchStatus(ctx, r.kargoClient, &promo, func(status *kargoapi.PromotionStatus) {
+		status.PromotedFreight = &targetFreightRef
+	})
+	if err != nil {
+		return nil, err
+	}
 
 	newStatus, nextFreight, err := r.promoMechanisms.Promote(ctx, stage, &promo, targetFreightRef)
```
hiddeco, line 432: If I recall correctly, `Promote` may actually modify the `FreightRef` with more accurate information based on the outcome from the promotion mechanisms. Given this, I think only patching the Status _before_ promoting doesn't suffice.

internal/controller/promotions/promotions.go
jacobnguyenn: agreed.

internal/controller/promotions/promotions.go
jacobnguyenn: Can you please kindly help to point out where it was mentioned?

api/v1alpha1/promotion_types.go
jacobnguyenn: Agreed. FreightRef maybe?

internal/controller/promotions/promotions.go
hiddeco: An example of where this happens is during Git promotion, where a `HealthCheckCommit` gets added to the `FreightRef`: https://github.com/akuity/kargo/blob/38ae6e6fedcfcc8da320e912347d27fae94f5d02/internal/controller/promotion/git.go#L209-L211

api/v1alpha1/promotion_types.go
krancour: It would be fine to just call it `Freight`. I don't think there would be any ambiguity in doing so.

internal/controller/promotions/promotions.go
krancour: Spot on. The health check commit is the main thing that may _change_ in the course of promotion.

To be perfectly honest, it's not going to matter to any health check logic if we don't update it the `PromotionStatus`, but it may cause confusion among human users if they notice the discrepancy between the FreightReference held by the Stage's CurrentFreight and History fields and this one.

krancour: i.e. So we _should_ make sure it gets set here.

api/v1alpha1/promotion_types.go
jacobnguyenn: Agreed.

internal/controller/promotion/git.go
```diff
@@ -119,7 +119,7 @@ func (g *gitMechanism) Promote(
 		); err != nil {
 			return nil, newFreight, err
 		}
-		newStatus = aggregateGitPromoStatus(newStatus, *otherStatus)
+		newStatus = aggregateGitPromoStatus(newStatus, *otherStatus).WithFreight(&newFreight)
```
hiddeco, line 122: I am not sure setting this here is the best thing we can do. As this would mean that for every promotion mechanism which does something with the Freight, we would have to remember to include it in the status.

What I would do instead, is persist the `FreightReference` to the `PromotionStatus` within `internal/controller/promotions/promotions.go` after `r.promoMechanisms.Promote` has been called. As `nextFreight` will contain the updated Freight data.

jacobnguyenn: > What I would do instead, is persist the `FreightReference` to the `PromotionStatus` within `internal/controller/promotions/promotions.go` after `r.promoMechanisms.Promote` has been called. As `nextFreight` will contain the updated Freight data.

I see, but is it more convenient to add freight reference right beside the place we update its details? Also, it is pretty unclear which promote mechanism cause freight reference update even if we update `PromotionStatus` in `nternal/controller/promotions/promotions.go`

jacobnguyenn: > > I see, but is it more convenient to reflect freight reference changes right beside the place we update its details?
> 
> Can you eleborate more on this question, I am not sure I am following the details of it.
> 


I'm just wondering if it is more clear to add the new detail of FreightReference to the Promotion Status. In that way, after calling r.promoMechanisms.Promote, the caller will have the latest FreightReference data.

> > Also, it is pretty unclear which promotion mechanism caused freight reference update even if we update `PromotionStatus` in `internal/controller/promotions/promotions.go`
> 
> Who and/or why would we have the desire to know what promotion mechanism changed something? Or how is the current approach solving this better?

For the second part, since we're doing an aggregate promote mechanism, will it be useful (or not) to know that which promote mechanism changed the FreightReference detail.

krancour: I believe the easiest place to assign the FreightReference to the PromotionStatus would be right here:

https://github.com/akuity/kargo/blob/5287d0e47578ff09af88e76eb091f28affdc81b1/internal/controller/promotions/promotions.go#L455-L458

A superior, but more involved option might be to modify the signature of the `Mechanism` interface's `Promote` function so that promotion mechanisms do not return PromotionStatus and modified FreightReference separately, but just return the PromotionStatus with the FreightReference already set.

I think I would suggest taking the easier approach and either @hiddeco or myself will do the rest of the housekeeping in a follow-up.

api/v1alpha1/promotion_types.go
```diff
@@ -96,6 +96,8 @@ type PromotionStatus struct {
 	// Metadata holds arbitrary metadata set by promotion mechanisms
 	// (e.g. for display purposes, or internal bookkeeping)
 	Metadata map[string]string `json:"metadata,omitempty" protobuf:"bytes,3,rep,name=metadata" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
+	// Freight is the detail of the piece of freight that was referenced by this promotion.
+	Freight *FreightReference `json:"promotedFreight,omitempty" protobuf:"bytes,4,opt,name=freight"`
```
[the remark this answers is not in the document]
hiddeco: You can run `make hack-codegen` which will replace the files.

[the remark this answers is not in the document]
krancour: > You can run make hack-codegen which will replace the files.

Sometimes merge conflicts prevent you from easily getting to a spot where you can do this.

I usually isolate all codegen activity to its own commit so that in the even of a merge or rebase, I can remove the old codegen commit ahead of time to prevent conflict with `main` and then a new codegen run should work fine.

But please don't waste a lot of time on that...

I would be happy to straighten out the codegen once everything else about the PR looks right.

internal/controller/promotions/promotions.go
```diff
@@ -437,6 +436,7 @@ func (r *reconciler) promote(
 	if err != nil {
 		return nil, err
 	}
+	newStatus = newStatus.WithFreight(&targetFreight)
```
krancour, line 439:
```suggestion
	newStatus = newStatus.WithFreight(&nextFreight)
```

krancour:
@jacobnguyenn don't worry about conflicts. I'll run codegen tomorrow to resolve them then look at this with fresh eyes.

I _think_ this might be all set.

krancour:
Just needs codegen rerun and then this is 🚀 

I will square that away on Monday.

hiddeco (approved):
Resolved the conflicts, this LGTM. Thanks @jacobnguyenn 🙇

internal/controller/promotions/promotions.go
```diff
@@ -458,6 +457,7 @@ func (r *reconciler) promote(
 	if err != nil {
 		return nil, err
 	}
+	newStatus = newStatus.WithFreight(&nextFreight)
```
krancour, line 460: Nit: Personally, I'm not sure I would delegate such a trivial assignment as `newStatus.Freight = nextFreight` to this new `WithFreight()` function.

hiddeco: I thought the same but figured we could sort it with the other suggestions we would tackle ourselves.

PR #1727, opened 2024-04-02
feat(ui): extend IntOrString to support JsonValue

devholic:
`intstr.IntOrString` relies on `gogo/protobuf` for JSON encoding/decoding, which is incompatible with the protobuf dependency used in our UI.

This commit extends `IntOrString` to encode/decode to/from JSON value to make compatible with
`gogo/protobuf`.

Related:
- https://github.com/akuity/kargo/issues/708
- https://github.com/akuity/kargo/pull/1515

### DIFF
--- a/ui/hack/generate-proto-extensions.mjs
+++ b/ui/hack/generate-proto-extensions.mjs
@@ -3,6 +3,36 @@
  */
 import { Project } from 'ts-morph';
 
+const formatCodeSettings = { indentSize: 2, convertTabsToSpaces: true };
+
+function overrideFromJson(classDecl, bodyText) {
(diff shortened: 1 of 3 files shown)


rpelczar (approved):

jessesuen:
I'm curious if Connect-generated APIs for other languages (golang, python, etc...) would face the same issue and require custom handling.

devholic:
> I'm curious if Connect-generated APIs for other languages (golang, python, etc...) would face the same issue and require custom handling.

Actually, it's a protobuf problem... It will happen to all clients to dealing with gogo/protobuf types 😢

PR #1730, opened 2024-04-02
fix: support helm in freightline

### LINKED ISSUE #1280: Warehouse helm subscription is not working
# Description

In the document, it is specified that kargo supports helm subscription for the warehouse. But when i tried this use case, freight box is detected when a new chart version comes up. But no version information is displayed. Also when promoted, stages shows promoted but no commits there in the git branch. Here is the warehouse.yaml

    kind: Warehouse
    metadata:
      name: kargo-demo
      namespace: kargo-demo
    spec:
      subscriptions:
      - chart:
          registryURL: oci://docker.io
          name: xyz/kargo-demo-app-charts
          semverConstraint: ^0.1.0

# Screenshots
image

# Version

```
v0.2.1
```

# Logs

```
No special logs
```

rbreeze:
Fixes #1280

CleanShot 2024-04-02 at 11 18 27@2x

### DIFF
--- a/ui/src/features/freightline/freight-contents.tsx
+++ b/ui/src/features/freightline/freight-contents.tsx
@@ -1,5 +1,5 @@
 import { faDocker, faGit } from '@fortawesome/free-brands-svg-icons';
-import { IconDefinition } from '@fortawesome/free-solid-svg-icons';
+import { IconDefinition, faAnchor } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Tooltip } from 'antd';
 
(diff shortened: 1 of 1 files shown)


rpelczar (approved):

krancour:
@rbreeze this is non-blocking, but would the Helm logo be a more intuitive thing to use here than the anchor? Or are we already using the anchor to represent Helm charts in other places?

rbreeze:
In an ideal world I'd also prefer the "real" Helm logo, but unfortunately FontAwesome doesn't have a built-in icon for Helm... eventually we could consider adding a custom PNG (and same thing for the Argo logo), but for now I figured it's more trouble than it's worth. I've been using the anchor icon in other PRs to represent Helm up to this point too

krancour:
> but for now I figured it's more trouble than it's worth

Totally fine with that. And even with it being an anchor instead of the Helm logo, this is a major improvement. Thanks for this!

Issue #1205, opened 2023-12-01
option to run (some) components as native processes


krancour:
Before we lose track of it...

In a conversation with @rbreeze earlier today, we talked about using a process manager like goreman to make it easier to build and run Kargo components from source _as native processes._

Look before we leap..

There are some considerations here:

Kargo's always going to have a dependency on k8s. Running Kargo components natively _outside_ a cluster -- which could be remote for all we care -- offers a _speed_ advantage.

On the flip side -- we'll never be able to get the webhook server working a a native process, because it needs to be accessible to the k8s cluster's own API server. If forced to run at least one component in the cluster, why not run them all in the cluster?

Last, the UI in particular would benefit tremendously from (optionally) being run as a native process. There is no way the Kargo container image can ever be rebuilt as fast as `vite` can hot reload modified UI code. _Maybe this is the main problem to solve for._

jessesuen:
I will push for this. I already do all my development by running kargo controller and API server with `go run`. I think it's the easiest/fastest development workflow and is especially helpful when I need to step through code with an IDE.

You're right that webhook is not possible to do without something like telepresence proxying back requests to desktop, but for API and controller work, it works wonderfully.

jessesuen:
In case it helps other, here's how I start kargo in VSCode. We'd need the same env var conveniences for the goreman script.

```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "api-server",
            "type": "go",
            "request": "launch",
            "mode": "auto",
            "program": "${workspaceFolder}/cmd/controlplane",
            "args": [
                "api"
            ],
            "env": {
                "UI_DIR": "${workspaceFolder}/ui/build",
                "LOG_LEVEL": "DEBUG",
                "KUBECONFIG": "/Users/jesse/.kube/config",
                //"DEX_ENABLED": "true",
                //"DEX_SERVER_ADDRESS": "http://localhost/dex",
                "ARGOCD_KUBECONFIG": "/Users/jesse/.kube/config",
                "ADMIN_ACCOUNT_TOKEN_SIGNING_KEY": "abc123",
                "ADMIN_ACCOUNT_TOKEN_AUDIENCE": "localhost",
                "ADMIN_ACCOUNT_TOKEN_ISSUER": "localhost",
[...]

PR #1738, opened 2024-04-02
chore: add a couple of conveniences for easily running Kargo API locally

rbreeze:
Related to #1205

Run `. ./hack/import-env.sh` to import environment variables from `.env.dev` to your local shell

Run `make start-api-local` to start the API locally


jessesuen (commented):
Is the `.env.dev` useful outside of start-api-local? If not, WDYT about creating a `./hack/start-api-local.sh` which has all the envs in the same file (can be overridden from environment). 

Then the Makefile target just runs that?

hack/start-api.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# password hash is 'admin'
+
+ADMIN_ACCOUNT_ENABLED=true \
+ADMIN_ACCOUNT_TOKEN_AUDIENCE=kargo.demo.akuity.io \
+ADMIN_ACCOUNT_TOKEN_ISSUER=http://kargo.demo.akuity.io \
```
jessesuen, line 6:
```suggestion
ADMIN_ACCOUNT_TOKEN_AUDIENCE=localhost \
ADMIN_ACCOUNT_TOKEN_ISSUER=localhost \
```

I see that this is set to `kargo.demo.akuity.io`, which is our demo instance URL. I'm pretty sure this doesn't matter since I believe these values are arbitrary. Suggest updating it to `localhost`

jessesuen (approved):

Issue #1231, opened 2023-12-06
Kargo should emit Kubernetes Events on interesting events


jessesuen:
# Proposed Feature

When interesting events happen, we should emit Kubernetes events.

Some events we should handle:
* User initiated a promotion
* User approved a Freight for promotion
* Promotion Succeeded/Failed/Error
* Freight became verified in a Stage

These events should then be shown in the UI in an Events tab.

# Motivation

These Events can then be persisted and shown alongside other interesting things

# Suggested Implementation

Kubernetes has an eventing library that API server and controller would use to emit events. However, its default behavior is to deduplicate multiple or reoccurring events into a single event and increase a counter of occurrences in the Event object. We need to make sure that we disable that when it makes sense since we want individual records of events. The deduplication behavior loses individual timestamps of each event and instead records first and last occurrence.

```[tasklist]
### Tasks
- [ ] https://github.com/akuity/kargo/issues/1818
- [x] Backend
```

jessesuen:
From discord:
> What about DORA metrics? Kargo knows when commit was made (commit which is related to warehouse artifacts) and when it reached production 

The above question got me thinking...

Since the Kargo API/controller has full context on Freight, Stages, Users, Promotions etc..., we should ensure that the events that we omit are enriched with useful information that could be mined later. For example, to address the above use case, we could pass along metadata around (as annotations to the Event) so that the Events can be associated together.

Here are some ideas on annotations that we would want to preserve:

```yaml
metadata:
  annotations:
    meta.kargo.akuity.io/initiator: jesse@akuity.io
    meta.kargo.akuity.io/freight: abc123
    meta.kargo.akuity.io/stage: prod
    meta.kargo.akuity.io/promotionName: prod.01hgjfdt6d22wcq1h6neez0enr.abc123
    meta.kargo.akuity.io/promotionCreationTime: 2023-12-05T12:00:00
    meta.kargo.akuity.io/promotionCompletionTime: 2023-12-05T12:01:00
    meta.kargo.akuity.io/stageVerificationTime: 2023-12-05T12:05:00
```

geowalrus4gh:
what about exposing a prometheus compatible metrics endpoint so that all information is made available in the endpoint?

geowalrus4gh:
DORA metrics are something we are looking for. Kargo must be the super candidate to provide these details.

krancour:
@gdsoumya or @devholic do you believe this issue is now fully addressed?

PR #1739, opened 2024-04-02
feat: emit Kubernetes events on Kargo events

devholic:
> Part of https://github.com/akuity/kargo/issues/1231

This PR let kargo components (API or controllers) to emit Kubernetes events when some notable events are happened.

For now, it emits on following Kargo events:

- User created Promotion
- User approved a Freight for Promotion
- Freight Verification Completed in a Stage
- Promotion Completed (Succeded/Failed/Errored)


internal/cli/cmd/server/server.go
```diff
@@ -99,6 +100,7 @@ func (o *serverOptions) run(ctx context.Context) error {
 		},
 		client,
 		client,
+		&fakekubeclient.EventRecorder{},
```
devholic, line 103: Q. Should the local server emit events?

rbreeze: IMO, yes. Will be useful for testing

internal/controller/promotions/promotions.go
```diff
@@ -106,16 +118,20 @@ func SetupReconcilerWithManager(
 }
```
krancour, line 118: Nit: I see that the task of building up the `reconcilerName` is accomplished a bit differently here and in the Stage reconciler. I am pretty sure that discrepancy is because the Stage reconciler has a `ReconcilerConfig` type that helped you deal with things in a nicer way and no such type exists here for the Promotion reconciler...

I like the way it was done for the Stage reconciler better, and even if it means introducing a new `ReconcilerConfig` type here for the Promotions reconciler, I'd be in favor of that.

charts/kargo/templates/api/cluster-role.yaml
```diff
@@ -23,6 +23,13 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - create
+      - patch
```
krancour, line 32: I am curious about the permission to patch Events. Aren't we only _creating_ events?

Same question goes for the controller.

internal/api/approve_freight_v1alpha1_test.go
```diff
@@ -345,20 +360,27 @@ func TestApproveFreight(t *testing.T) {
 			},
 			assertions: func(
 				t *testing.T,
+				er *fakekubeclient.EventRecorder,
```
krancour, line 363: Nit: `er` looks too close to the symbol commonly used for errors -- `err`. I suspect this will be the source of confusion now and then unless one reading the code brings extra scrutiny to bear on what they're reading. I would name this differently, but like I said... this is a nit. Dealer's choice.

krancour:
I made a few comments. Overall, this looks good.

One thing I notice is missing here is that if a user created a promotion via `kubectl`, there will be no event published for that.

I understand why this is being done in the API server -- because there is a better ability to capture user identity there -- but I don't think I'd want to completely ignore Promotions created via `kubectl`.

Here is what I propose:

A webhook can access Kubernetes end-user information, like user or SA name. Modify the existing validating webhook for Promotions so that on create, _if valid_, and _not_ created by the API server or controller, the Promotion created event is published from there. If the API server is the user creating the Promotion, you can ignore, with the assumption that the API server did or will publish its own event with information about the OIDC user. Similarly, if the controller is the user creating the Promotion (which only happens on auto-promotion), you can assume that the controller did or will publish its own event.

internal/api/approve_freight_v1alpha1.go
```diff
@@ -112,6 +114,26 @@ func (s *server) ApproveFreight(
 		return nil, fmt.Errorf("patch status: %w", err)
 	}
 
+	eventAnnotations := map[string]string{
```
hiddeco, line 117: Nit: I like how this is factored out in other places.

internal/api/approve_freight_v1alpha1_test.go
hiddeco: I historically have ended up with using either `events` or `recorder` in these kind of situations.

internal/controller/stages/stages.go
```diff
@@ -917,6 +967,23 @@ func (r *reconciler) syncNormalStage(
...
+			kargoapi.AnnotationKeyEventActor:         kargoapi.FormatEventControllerActor(r.cfg.Name()),
+			kargoapi.AnnotationKeyEventProject:       promo.Namespace,
+			kargoapi.AnnotationKeyEventPromotionName: promo.Name,
+			kargoapi.AnnotationKeyEventFreightAlias:  latestFreight.Alias,
+			kargoapi.AnnotationKeyEventFreightName:   latestFreight.Name,
+			kargoapi.AnnotationKeyEventStageName:     promo.Spec.Stage,
+		},
+		corev1.EventTypeNormal,
+		kargoapi.EventReasonPromotionCreated,
+		"Automatically promote for Stage %q",
```
hiddeco, line 983:
```suggestion
		"Automated promotion for Stage %q",
```

or

```suggestion
		"Automatically promoted Freight for Stage %q",
```

maybe?

internal/api/approve_freight_v1alpha1_test.go
devholic: Changed to use `recorder`: https://github.com/akuity/kargo/pull/1739/commits/dc92e6044fb6b7216728c29a255c14fe327ce0a3

internal/controller/stages/stages.go
devholic: Updated :)
https://github.com/akuity/kargo/pull/1739/commits/3ba27ff252089a0c3a6391e154d7e569aadebb54

internal/webhook/admission.go
```diff
@@ -0,0 +1,22 @@
...
+
+type IsRequestFromKargoControlplaneFn func(admission.Request) bool
+
+func IsRequestFromKargoControlplane(
+	knownServiceAccounts map[types.NamespacedName]struct{},
+) IsRequestFromKargoControlplaneFn {
+	return func(req admission.Request) bool {
+		for account := range knownServiceAccounts {
+			if serviceaccount.MatchesUsername(account.Namespace, account.Name, req.UserInfo.Username) {
+				return true
```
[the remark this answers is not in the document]
krancour: @gdsoumya what would the regex-based match be needed for? It looks like @devholic included relevant changes to the chart to populate webhook config with the exact details of the SAs used by the API server and controller.

But I assume you have something in mind...

internal/webhook/config/config.go
```diff
@@ -0,0 +1,38 @@
+package config
+
+import (
+	"strings"
+
+	"github.com/kelseyhightower/envconfig"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type WebhookConfig struct {
```
krancour, line 10: Nit: idk about this needing its own package. Can we move it up one level? Or was this choice made to break a would-be import cycle?

internal/webhook/freight/webhook.go
```diff
@@ -211,6 +235,19 @@ func (w *webhook) ValidateUpdate(
...
+	if err != nil {
+		return nil, fmt.Errorf("get admission request from context: %w", err)
+	}
+	// Record Freight approved events if the request doesn't come from Kargo controlplane.
+	if !w.isRequestFromKargoControlplaneFn(req) {
+		for approvedStage := range newFreight.Status.ApprovedFor {
+			if _, ok := oldFreight.Status.ApprovedFor[approvedStage]; !ok {
+				w.recordFreightApprovedEvent(req, newFreight, approvedStage)
+			}
+		}
```
krancour, line 249: I think it's good that this is here and I wouldn't change it, but want to mention that I'm not sure if approval is actually possible except via API. I've never had luck mutating status directly from `kubectl`. iirc, there is some limitation to controller runtime that prevents it.

krancour:
Apart from [one nit](https://github.com/akuity/kargo/pull/1739#discussion_r1558184530), I am happy with this. Nice work, @hoon. 🔥 

I'll wait to see what @gdsoumya says [this](about https://github.com/akuity/kargo/pull/1739#discussion_r1558154822) before approving.

internal/webhook/admission.go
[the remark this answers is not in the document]
krancour: That makes sense.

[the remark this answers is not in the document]
krancour: @gdsoumya had you been thinking of the regex being defined by the chart? Or had you been thinking of hard-coding one that would therefore establish a specific convention for how control plane SAs are named?

internal/webhook/config/config.go
devholic: I followed how `api` server organize config, but no preferences over it. Will move under `webhook` :)

internal/webhook/freight/webhook.go
hiddeco: This is _technically_ possible using `kubectl edit --subresource=status`.

devholic: @krancour @hiddeco I found that I miss grant webhook-server to record event 😓
After granting it in https://github.com/akuity/kargo/pull/1739/commits/41165fe15de6a24f98c165f35dc4814c0962209c and as @hiddeco mentioned - using `kubectl edit --subresource=status` records event:

krancour (approved):

PR #1743, opened 2024-04-02
fix(docs): minor fixes to the quickstart

krancour:
Adds the shard column to some of the sample output.

Also stops using Freight's `id` field, which is empty for all _new_ Freight.

Instead of using Freight's `metadata.name` in place of `id`, the docs now opt for using the Freight's `alias`, which is now supported just as well as `metadata.name` as a means of identifying a piece of Freight.

These fixes also need to be cherry-picked into the `release-0.5` branch.

cc @christianh814

### DIFF
--- a/docs/docs/20-quickstart.mdx
+++ b/docs/docs/20-quickstart.mdx
@@ -418,8 +418,8 @@ the previous section.
     Sample output:
 
     ```shell
-    NAME         AGE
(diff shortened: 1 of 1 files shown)


rbreeze (approved):

christianh814 (approved):
LGTM!

Issue #1207, opened 2023-12-01
UI: Surface promotion status in DAG


rbreeze:
When a promotion fails, it should be obvious from the DAG view

rbreeze:
Discussed with @jessesuen and decided that we will add a new field `status.lastPromotion` to accomplish this

krancour:
I need to revisit the back end changes needed to enable this. They're not a high priority for me atm, but I wanted to note that it hasn't been forgotten. Assigning myself as co-owner of this issue with @rbreeze since progress is blocked on me.

PR #1748, opened 2024-04-03
feat: add lastPromotion field to Stage status

rbreeze:
Necessary for https://github.com/akuity/kargo/issues/1207


api/v1alpha1/stage_types.go
```diff
@@ -670,6 +672,15 @@ type PromotionInfo struct {
 	Freight FreightReference `json:"freight" protobuf:"bytes,2,opt,name=freight"`
 }
 
+type LastPromotionInfo struct {
+	// Name is the name of the Promotion
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+	// Freight is the freight being promoted
+	Freight FreightReference `json:"freight" protobuf:"bytes,2,opt,name=freight"`
+	// Status is the status of the promotion
+	Status PromotionStatus `json:"status" protobuf:"bytes,3,opt,name=status"`
```
krancour, line 681: I wonder if we could just add this field to the existing `PromotionInfo` type instead of making a whole new type that so closely resembles it.

krancour:
At a glance, looks good, but I need to look closer tomorrow.

krancour:
@rbreeze and I actually discussed some small modifications yesterday that I don't think have made it in yet.

internal/controller/stages/stages.go
```diff
@@ -696,6 +703,36 @@ func (r *reconciler) syncNormalStage(
...
+			return status, err
+		}
+
+		freight, err := r.getFreightFn(ctx, r.kargoClient, types.NamespacedName{
+			Namespace: stage.Namespace,
+			Name:      promo.Spec.Freight,
+		})
+		if err != nil {
+			return status, err
+		}
```
krancour, line 722: Can this be copied from `currentPromotion.Freight` instead of being retrieved here?

PR #1752, opened 2024-04-03
fix(ui): could not click dropdown without also opening freight view

### LINKED ISSUE #1749: ui: clicking "hamburger menu" on a piece of freight also triggers freight view
It seems the logic must pop open the freight view for a click anywhere within a piece of Freight's borders, but that should probably exclude the boundaries of the "hamburger."

rbreeze:
Fixes #1749


krancour:
Test drove this and found that this stops the Freight view from opening when the hamburger is clicked, but once selecting an item _from_ the hamburger menu, the Freight view still does open.

krancour (approved):
🔥 Thank you @rbreeze !

PR #1753, opened 2024-04-03
feat(controller): wait for Application operation to finish before marking Promotion as successful

hiddeco:
Fixes: #1613

This pull request adds logic to the Argo CD Application promotion mechanism to allow it to wait for the operation it initiated to finish, before marking the Promotion as successful. It achieves this by inspecting the operation state of an Application before triggering an update, to confirm an operation for the Promotion is not already running and/or has produced the desired revision compared to the Freight.


internal/controller/promotion/argocd_test.go
```diff
@@ -175,6 +345,252 @@ func TestArgoCDPromote(t *testing.T) {
 	}
 }
 
+func TestArgoCDMustPerformUpdate(t *testing.T) {
+	testCases := []struct {
```
hiddeco, line 349: @krancour I couldn't resist to show how I would write these tests compared to how they're typically done in the code base.

jessesuen:
> Fixes: https://github.com/akuity/kargo/pull/1753

Do you mean it fixes https://github.com/akuity/kargo/issues/1613?

krancour (approved):

Issue #1733, opened 2024-04-02
Kargo objects should have a description field


jessesuen:
# Proposed Feature

When inputing/managing credentials (or other objects), I would like the ability to add some notes about the credential (e.g. expiration, URL to the site where it can be managed). Kargo should introduce a description field on credentials so users can add their notes.

# Motivation

It would be useful to have some more context about credential it refers to so it can be managed easier.

# Suggested Implementation

Our secrets can recognize annotation where we allow users to capture a description. This description is not sensitive, and would be shown in UI in our credentials tab. I prefer annotation over something in `data` because I believe some secrets types have predefined fields in `data` that do not allow for extra fields. Also, data field in secrets need to be base64 decoded before viewed.

hiddeco:
I can imagine the same could be said for Stages, or even Projects?

jessesuen:
> I can imagine the same could be said for Stages, or even Projects?

Great idea! Updated title

krancour:
Note that for Stages and Projects (and Warehouses, too, for that matter) I'd rather see this be a field than an annotation.

I understand the motivations for it being an annotation on a Secret, but absent those, a field is more straightforward.

jessesuen:
For something like AnalysisTemplates, we also can't control the spec and it would need to be an annotation. So Secrets wouldn't be the only inconsistency.

Does that change your position if we had to have Secrets/AnalysisTemplates use annotations but Warehouse, Stages, Projects use `spec.description`?

krancour:
Very good call out @jessesuen. Yes. That changes my thinking. Annotations across the board.

Note that this is already implemented for credentials/`Secret`s as of #1772 -- merged just earlier today.

If we're going the annotation route, what's left of this is only a UI task.

Removing API and CLI labels.

krancour:
There is a lot of UI work remaining for v0.6.0. I'm making the decision that this is some of the less critical UI work and am deferring it to v0.7.0.

PR #1772, opened 2024-04-05
feat(cli/api/ui): add a description to credentials

zhenya-khvan-form3:
Partially addresses https://github.com/akuity/kargo/issues/1733

Note | Screenshot
-- | --
Two secrets, one with description and one without | image
`kubectl` command listing secrets | image


internal/api/update_credentials_v1alpha1.go
```diff
@@ -90,6 +92,15 @@ func applyCredentialsUpdateToSecret(
 	secret *corev1.Secret,
 	credsUpdate credentialsUpdate,
 ) {
+	if credsUpdate.description != "" {
+		if secret.Annotations == nil {
+			secret.Annotations = make(map[string]string)
```
krancour, line 97: Nit: You can initialize this with a capacity of 1 to avoid an resizing when the annotation is added to it.

This is non-blocking.

krancour:
@zhenya-khvan-form3 thanks for this! And overall, it's in great shape.

✅  I'm quite happy with the server-side of this.

For UI, I think the description _if present_ would be better displayed as a thin, slightly differently colored (very light gray?) row immediately below the credential and _spanning all columns_. I think this will be a nicer way to display wordier descriptions that probably can't be comfortably jammed into one column.

I'm asking @rbreeze to please have a look over the UI portion of this as well.

Last, I think there needs to be an optional `--description` flag added to both the `kargo create credentials` and `kargo update credentials` CLI commands.

krancour:
@zhenya-khvan-form3 I think that was a giant step in the right direction!

It looks like the description is expandable/collapsible? I would probably not bother with that unless @rbreeze says we should.

I think we could probably also reduce the top and bottom margins/padding in the description row so that it's very thin.

> I'm happy to do the CLI change, I didn't do it initially just to make sure the rest was sound.

👍 It's sound. This is a really nicely done PR.

internal/cli/cmd/create/credentials.go
```diff
@@ -120,6 +122,7 @@ func (o *createCredentialsOptions) addFlags(cmd *cobra.Command) {
 		cmd.Flags(), &o.Project, o.Config.Project,
 		"The project in which to create credentials. If not set, the default project will be used.",
 	)
+	option.Description(cmd.Flags(), &o.Description, "Description of the credential.")
```
krancour, line 125:
```suggestion
	option.Description(cmd.Flags(), &o.Description, "Description of the credentials.")
```

krancour:
@rbreeze can you look at the UI aspects of this PR please?

rbreeze (approved):
LGTM, this is great! 

I agree with @krancour that we don't need expandable/collapsable descriptions, but in this case, the implementation with Antd is _easier_ than if we were to fix it in place. Additionally, since each row is expanded by default, the UX isn't much different but users can hide the descriptions if they want, which I think is good.

For padding on the description row, I agree we can shrink it, but I won't block this PR for it because 1. Antd doesn't have an immediately obvious way to do it and 2. I don't foresee users adding dozens of credentials, so we probably won't be pressed for space here. We can always follow up with a PR to shrink the padding if necessary

krancour:
Ok. Let's take this as is. @rbreeze would you mind opening the follow-up issue for the padding on the description?

PR #1777, opened 2024-04-05
feat(cli): add --no-headers flag to all get commands.

### LINKED ISSUE #1751: cli: add --no-headers flag
`kubectl` has a useful `--no-headers` flag that you can use on any get operation. It is enormously useful when you're going to pipe the output into `grep`, `awk`, etc.

Without `--no-headers`:

```shell
$ kubectl get namespaces | head -n 5

NAME                 STATUS   AGE
kube-system          Active   6h52m
kube-public          Active   6h52m
default              Active   6h52m
kube-node-lease      Active   6h52m
```

With `--no-headers`:

```shell
$ kubectl get namespaces --no-headers | head -n 5

kube-system          Active   6h52m
kube-public          Active   6h52m
default              Active   6h52m
kube-node-lease      Active   6h52m
cert-manager         Active   6h52m
```

I'd love to see us add a `--no-headers` flag to the `kargo` CLI as well.

webstradev:
Fixes #1751

So I hope this way of solving it is okay, please be critical because I haven't worked much on cobra before.

I've added the flag to the options pacakge just to keep it clean, I believe we have done that for all flags as far as I can tell.

I set the --no-headers flag as a persistent flag on the get (root) command using the same options pattern.

In the options structs of all subcommands I have embedded the getOptions (kept it unexported for now), so that we can easily add more get options to all subcommands. **Struct embedding felt natural here to me as we are using it across all get commands, but happy to amend that if needed** just means we get o.getOptions.NoHeaders instead of o.NoHeaders in a few places.

For the printTable function I ended up just adding a no headers bool as well, the alternative would have been being able to pass the whole printoptions struct, but it would have meant a bunch of if statements on every subcommand, plus I don't see the use currently for that considering we are only specifying one print option.

Note: I haven't doen the tests yet because I wanted to make sure the implementation is alright first. Although I'm note sure if there are any tests for these aspects of the CLI.

```
webstradev@Eriks-MacBook-Pro kargo % kargo get project
NAME         PHASE   AGE
my-project   Ready   26h
webstradev@Eriks-MacBook-Pro kargo % kargo get project --no-headers
my-project   Ready   26h
webstradev@Eriks-MacBook-Pro kargo % 
```

Closes #1751

Let me know what you think!

### DIFF
--- a/internal/cli/cmd/get/get.go
+++ b/internal/cli/cmd/get/get.go
@@ -38,7 +38,7 @@ kargo get promotions --project=my-project --stage=my-stage
 `),
 	}
 
-	cmdOpts := getOptions{}
+	cmdOpts := &getOptions{}
 
 	cmdOpts.addFlags(cmd)
 
--- a/internal/cli/cmd/get/get.go
+++ b/internal/cli/cmd/get/get.go
@@ -38,17 +38,17 @@ kargo get promotions --project=my-project --stage=my-stage
 `),
 	}
 
-	cmdOpts := &getOptions{}
-
-	// Register subcommands.
-	cmd.AddCommand(newGetCredentialsCommand(cfg, streams))
-	cmd.AddCommand(newGetFreightCommand(cfg, streams))
-	cmd.AddCommand(newGetProjectsCommand(cfg, streams))
-	cmd.AddCommand(newGetPromotionsCommand(cfg, streams))
-	cmd.AddCommand(newGetStagesCommand(cfg, streams))
-	cmd.AddCommand(newGetWarehousesCommand(cfg, streams))
+	cmdOpts := getOptions{}
 
 	cmdOpts.addFlags(cmd)
+
+	// Register subcommands.
+	cmd.AddCommand(newGetCredentialsCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetFreightCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetProjectsCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetPromotionsCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetStagesCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetWarehousesCommand(cfg, streams, cmdOpts))
 
 	return cmd
 }
--- a/internal/cli/cmd/get/get.go
+++ b/internal/cli/cmd/get/get.go
@@ -17,6 +17,10 @@ import (
 	"github.com/akuity/kargo/internal/cli/templates"
 )
 
+type getOptions struct {
+	NoHeaders bool
+}
+
 func NewCommand(cfg config.CLIConfig, streams genericiooptions.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "get TYPE [NAME ...]",
@@ -34,6 +38,8 @@ kargo get promotions --project=my-project --stage=my-stage
 `),
 	}
 
+	cmdOpts := &getOptions{}
+
 	// Register subcommands.
 	cmd.AddCommand(newGetCredentialsCommand(cfg, streams))
 	cmd.AddCommand(newGetFreightCommand(cfg, streams))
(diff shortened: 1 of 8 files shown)


webstradev:
The one thing I have yet to do is update any documentation, examples, cli descriptiosn etc etc, I wanted to have someone look at the implementation first before I do that.

internal/cli/option/flag.go
```diff
@@ -147,6 +147,11 @@ func NewAlias(fs *pflag.FlagSet, stage *string, usage string) {
 	fs.StringVar(stage, NewAliasFlag, "", usage)
 }
 
+// NoHeaders adds the NoHeadersFlag to the provided flag set.
+func NoHeaders(fs *pflag.FlagSet, noHeaders *bool) {
+	fs.BoolVar(noHeaders, "no-headers", false, "Don't print headers when printing resource tables")
```
krancour, line 152: I know that nailing this help text down was part of what you were saving for last, but I don't see anything else that needs changing... so...

My only suggestion here is tweaking the text to be a little more like this:

```
kubectl get -h | grep no-headers -A 1
    --no-headers=false:
	When using the default or custom-column output format, don't print headers (default print headers).
```

We don't have any custom-column output format (nor am I inclined to add this at the moment), so we could maye tweak this to:

```suggestion
	fs.BoolVar(noHeaders, "no-headers", false, "When using the default output format, don't print headers (default print headers).")
```

krancour: Oh... despite not being referenced in multiple placed, I would still maybe create a constant for `"no-headers"` purely for consistency with the other flags.

krancour:
@webstradev apart from my feedback about help wording and creating a constant for the flag, this whole PR looks rock solid to me.

I do want @hiddeco to give it a once-over as well.

internal/cli/cmd/get/get.go
```diff
@@ -34,20 +38,30 @@ kargo get promotions --project=my-project --stage=my-stage
...
 	cmd.AddCommand(newGetStagesCommand(cfg, streams))
 	cmd.AddCommand(newGetWarehousesCommand(cfg, streams))
+
+	cmdOpts.addFlags(cmd)
+
 	return cmd
 }
 
+func (o *getOptions) addFlags(cmd *cobra.Command) {
+	option.NoHeaders(cmd.PersistentFlags(), &o.NoHeaders)
```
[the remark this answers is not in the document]
krancour: Ah. Good eye @hiddeco.

I like no. 2, personally. no. 1 would work, but I don't think I'd care for it being handled so much differently from `streams`. Consistency is a virtue.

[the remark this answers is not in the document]
webstradev: Yeah I like number two better as well. will sort that now

internal/cli/option/flag.go
webstradev: This should resolve both your comments @krancour 
https://github.com/akuity/kargo/pull/1777/commits/1df2965113b9ff0f93d0d2eb59a4510c1adfa78c

internal/cli/cmd/get/get.go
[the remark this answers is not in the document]
webstradev: @hiddeco this should resolve this:
https://github.com/akuity/kargo/pull/1777/commits/9bd37eeda943a857b1a2bf1dc9897387bffcca39.

And it is good that we did it, because I missed the flag registration on one resource so that would have been an annoying one considering we don't have tests for this.

internal/cli/cmd/get/get.go
```diff
@@ -34,20 +38,30 @@ kargo get promotions --project=my-project --stage=my-stage
...
-	cmd.AddCommand(newGetProjectsCommand(cfg, streams))
-	cmd.AddCommand(newGetPromotionsCommand(cfg, streams))
-	cmd.AddCommand(newGetStagesCommand(cfg, streams))
-	cmd.AddCommand(newGetWarehousesCommand(cfg, streams))
+	cmd.AddCommand(newGetCredentialsCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetFreightCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetProjectsCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetPromotionsCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetStagesCommand(cfg, streams, cmdOpts))
+	cmd.AddCommand(newGetWarehousesCommand(cfg, streams, cmdOpts))
```
hiddeco, line 51: As the values of `cmdOpts` are set at runtime, the constructors should take a pointer I think. Otherwise the propagated value won't be available to the downward commands.

krancour: Test drove and this observation is correct.

webstradev: Ow that's odd I gave it a try and it did seem to work fine but ill give it a second look

webstradev: well actually properly rebuilding the cli to test will probably help, have pushed a fix to amend.

krancour: fwiw, when actively working on the cli, I do this:

```
alias kargo='go run cmd/cli/*.go'
```

krancour (approved):
🔥 Thank you so much for this @webstradev!

internal/cli/cmd/get/get.go
webstradev: ow that is probably better for me to do than what I have now

PR #1778, opened 2024-04-05
feat(controller): support filtering Git subscriptions using globs and path prefixes

maksimstankevic:
…e Git Subscription

### DIFF
--- a/internal/controller/warehouses/git.go
+++ b/internal/controller/warehouses/git.go
@@ -20,6 +20,7 @@ import (
 const (
 	regexpPrefix = "regexp:"
 	regexPrefix  = "regex:"
+	globPrefix   = "glob:"
 )
 
 type gitMeta struct {
@@ -309,7 +310,7 @@ func ignores(tagName string, ignore []string) bool {
 // filters match one or more commit diffs and new Freight is
 // to be produced. It returns false otherwise.
 func matchesPathsFilters(includePaths []string, excludePaths []string, diffs []string) (bool, error) {
-	includePathsRegexps, includePathsGlobs, err := sortGlobsAndRegexps(includePaths)
+	includePathsRegexps, includePathsGlobs, includePathsPrefixes, err := sortFilters(includePaths)
 	if err != nil {
 		return false, fmt.Errorf(
 			"error compiling includePaths regexps: %w",
@@ -317,7 +318,7 @@ func matchesPathsFilters(includePaths []string, excludePaths []string, diffs []s
 		)
 	}
 
-	excludePathsRegexps, excludePathsGlobs, err := sortGlobsAndRegexps(excludePaths)
+	excludePathsRegexps, excludePathsGlobs, excludePathsPrefixes, err := sortFilters(excludePaths)
 	if err != nil {
 		return false, fmt.Errorf(
 			"error compiling excludePaths regexps: %w",
@@ -341,16 +342,18 @@ func matchesPathsFilters(includePaths []string, excludePaths []string, diffs []s
 				err,
 			)
 		}
+		matchesIncludePrefixes := matchesPrefixList(diffPath, includePathsPrefixes)
+		matchesExcludePrefixes := matchesPrefixList(diffPath, excludePathsPrefixes)
 		// matchesIncludePaths case is a bit different from matchesExcludePaths
 		// in the way that if includePaths string array is empty - it matches
 		// ANY change so we need to have a check for that
 		matchesIncludePaths := len(includePaths) == 0 || matchesRegexpList(
 			diffPath,
 			includePathsRegexps,
-		) || matchesIncludeGlobs
+		) || matchesIncludeGlobs || matchesIncludePrefixes
 		matchesExcludePaths := matchesRegexpList(diffPath,
 			excludePathsRegexps,
-		) || matchesExcludeGlobs
+		) || matchesExcludeGlobs || matchesExcludePrefixes
 		// combined filter decision, positive for matching includePaths and
 		// unmatching excludePaths
 		if matchesIncludePaths && !matchesExcludePaths {
@@ -363,26 +366,31 @@ func matchesPathsFilters(includePaths []string, excludePaths []string, diffs []s
 	return false, nil
 }
 
-// sortGlobsAndRegexps handles sorting of a slice of strings to globs
-// and regexps based on prefix, regexps are compiled into a slice of
+// sortFilters handles sorting of a slice of strings to regexps,
+// globs and prefixes, regexps are compiled into a slice of
 // *regexp.Regexp additionally
-func sortGlobsAndRegexps(regexpStrings []string) (regexps []*regexp.Regexp, globs []string, err error) {
+func sortFilters(regexpStrings []string) (regexps []*regexp.Regexp, globs []string, prefixes []string, err error) {
 	regexpsSlice := make([]*regexp.Regexp, 0, len(regexpStrings))
 	globsSlice := make([]string, 0, len(regexpStrings))
+	prefixesSlice := make([]string, 0, len(regexpStrings))
 	for _, regexpString := range regexpStrings {
 		switch {
 		case strings.HasPrefix(regexpString, regexpPrefix):
 			regexpString = strings.TrimPrefix(regexpString, regexpPrefix)
 		case strings.HasPrefix(regexpString, regexPrefix):
 			regexpString = strings.TrimPrefix(regexpString, regexPrefix)
-		default:
+		case strings.HasPrefix(regexpString, globPrefix):
+			regexpString = strings.TrimPrefix(regexpString, globPrefix)
 			globsSlice = append(globsSlice, regexpString)
+			continue
+		default:
+			prefixesSlice = append(prefixesSlice, regexpString)
 			continue
 		}
 
 		regex, err := regexp.Compile(regexpString)
 		if err != nil {
-			return nil, nil, fmt.Errorf(
+			return nil, nil, nil, fmt.Errorf(
 				"error compiling string %q into a regular expression: %w",
 				regexpString,
 				err,
@@ -390,7 +398,7 @@ func sortGlobsAndRegexps(regexpStrings []string) (regexps []*regexp.Regexp, glob
 		}
 		regexpsSlice = append(regexpsSlice, regex)
 	}
-	return regexpsSlice, globsSlice, nil
+	return regexpsSlice, globsSlice, prefixesSlice, nil
 }
 
 // matchesRegexpList is a general purpose function iterating given slice of
@@ -424,6 +432,19 @@ func matchesGlobList(stringToMatch string, globList []string) (bool, error) {
 		}
 	}
 	return false, nil
+}
+
+// matchesPrefixList is a general purpose function iterating given slice of
+// strings (prefixList) to check if any of prefixes match stringToMatch string,
+// if match is found it returns true, if match is not found it
+// returns false
+func matchesPrefixList(stringToMatch string, prefixList []string) bool {
+	for _, prefix := range prefixList {
+		if strings.HasPrefix(stringToMatch, prefix) {
+			return true
+		}
+	}
+	return false
 }
 
 // selectLexicallyLastTag sorts the provided tag name in reverse lexicographic
--- a/internal/controller/warehouses/git.go
+++ b/internal/controller/warehouses/git.go
@@ -2,7 +2,6 @@ package warehouses
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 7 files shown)


maksimstankevic:
@hiddeco @krancour as this is a continuation of includePaths/excludePaths story, tagging you both to take a look at my changes for added glob matching when you have a second

krancour:
@maksimstankevic thanks for this!

This looks pretty good. I've got just a few miscellaneous observations and questions.

* I think the comments on `IncludePaths` and `ExcludePaths` fields need to be updated to reflect the new options (and the codegen re-run).

* I'm not sure whether this really should be based _only_ on globs. As it stands currently, I think people would be quite surprised to find that `foo/bar` isn't a match for the pattern `foo/`. @maksimstankevic and @hiddeco what are each of your thoughts about this?

* I only just noticed this now and didn't notice in #1477 -- does this path filtering only work with the `NewestFromBranch` commit selection strategy? That appears to be the case and I'm now wondering why that is. (We can open a separate issue for this if we determine this was a mistake.)

maksimstankevic:
> * I think the comments on `IncludePaths` and `ExcludePaths` fields need to be updated to reflect the new options (and the codegen re-run).
**-----My bad, will work on it with after next point is finalised.** 
> * I'm not sure whether this really should be based _only_ on globs. As it stands currently, I think people would be quite surprised to find that `foo/bar` isn't a match for the pattern `foo/`. @maksimstankevic and @hiddeco what are each of your thoughts about this?
**-----I agree, but do we combine prefix matching and glob matching under the same unprefixed string array (that holds only globs currently)? Or we introduce new "regex/p:"-like prefix for either "prefix:" or "glob:" and leave the other one unprefixed?**
> * I only just noticed this now and didn't notice in [feat(api/controller): support filtering rules Git subscriptions #1477](https://github.com/akuity/kargo/pull/1477) -- does this path filtering only work with the `NewestFromBranch` commit selection strategy? That appears to be the case and I'm now wondering why that is. (We can open a separate issue for this if we determine this was a mistake.)
**-----I somehow thought back then that tag filtering should not interfere with commit filtering. Now I have less confidence it was a correct approach but I agree if it's decided to add includePaths/excludePaths filtering to CommitSelectionStrategies involving tags - that needs to be done in a separate issue**

krancour:
Ok. So second point is the one most in need of addressing. imho, yes... prefix and glob match should go together, but I am willing to hear arguments against that. My argument in favor of it is that anyone accustomed to working in a shell environment is likely to find that to be the least surprising behavior. I contend most people would never think to do `foo/*` and would assume that `foo/` suffices.

hiddeco:
One argument against combining them would be that there could be potential conflicts or overlaps in certain scenarios, such as when dealing with literals versus special characters, nested patterns, or character escaping.

To solve this, one would have to start defining (and documenting) rules around precedence. Which in the end may make it easier from both a technical and user perspective to create a separation between the two.

maksimstankevic:
Yeah, I agree, in my eyes that would be an overcomplication for a simple topic. I would go with "regex:/regexp:" prefix for regexps, "glob:" prefix for globs, these first 2 being the advanced options for folks wishing to use them (like me 😄 ) and strings without prefix would be for exact matches and string prefixes like "foo/" to match "foo/bar".

krancour:
I'm a little confused because I feel as if we all probably use prefixes _and_ globs on a regular basis in our shells without ever having to specify which we're using.

hiddeco:
There are small details to take into account here:

1. Globbing in the shell is a two-step feature. The globbing happens before feeding the input to the command, and it's not a feature of the CLI you are working with.
2. `.gitignore` is a special format, and _not_ 1:1 equal to glob https://git-scm.com/docs/gitignore#_pattern_format.

krancour:
> Globbing in the shell is a two-step feature. The globbing happens before feeding the input to the command, and it's not a feature of the CLI you are working with.

Right... but that's an implementation detail. You don't _think_ about that when you write commands like `mv foo/ bar/` or `mv foo/*.go bar/` you just expect either way of doing things to work.

hiddeco:
But that implementation detail means that globbing > prefix matches. Which means there is an order of precedence.

hiddeco:
> That may technically be so, but how often do you stop and think about this?

Not very often, but the shell also present me for different things to e.g. do literal string interpretations. Assuming I have a directory called `foo?`, I can do `ls "foo?"` to ensure it wouldn't be parsed as a glob. Using a declarative API, this would probably have to become `foo\?` as the input is a literal string already, and it needs to be escaped _somehow_.

These kind of details make me worry about edge cases for directory structures and (not so) great people do while naming things. In addition to it creating technical hurdles over simply using an identifier, as you can't e.g. say "if it contains one of these special characters, it's a glob".

> Would you not find it surprising if foo/ and foo/* (for example) didn't both work with our default matching strategy?

It would depend on what the API tells me the features are, if the default is just simple prefix matching and there also is a regular expression feature with an identifier. I wouldn't, I think.

krancour:
> It would depend on what the API tells me the features are

This is a fair point, but I also want to default to the least surprising behavior for the benefit of those who didn't read the manual.

_That_ is the main thing I care about.

If we think prefix matching would be the least surprising default, let's go with that and use the `glob:` prefix for glob-matching.

maksimstankevic:
Thanks @hiddeco @krancour, I'll do it with "glob:" for globs and no prefix for prefix and exact string matching.

krancour:
@maksimstankevic as always, thank you for your patience and your efforts!

maksimstankevic:
@hiddeco @krancour pushed it, please review when you can

internal/controller/warehouses/git.go
```diff
@@ -303,19 +306,19 @@ func ignores(tagName string, ignore []string) bool {
 	return false
 }
 
-// pathsFilterPositive returns true when IncludePaths and/or ExcludePaths
+// matchesPathsFilters returns true when IncludePaths and/or ExcludePaths
 // filters match one or more commit diffs and new Freight is
 // to be produced. It returns false otherwise.
 func matchesPathsFilters(includePaths []string, excludePaths []string, diffs []string) (bool, error) {
```
krancour, line 312: This function looks like it could be written more simply and more efficiently as well.

Something like the following pseudocode would probably be easier to understand and perform better:

```
for diff in diffs:
  if !match(diff, includePrefixes) && !match(diff, includeGlobs) && !match(diff, includeRegexes):
    // Diff not selected by any of the patterns
    continue
  // If we get to here, the diff was selected by one of the patterns. Check if it's un-selected.
  if match(diff, excludePrefixes) || match(dif, excludeGlobs) || match(diff, excludeRegexes):
    // Diff was unselected by one of the patterns
    continue
  // If we get to here, we have found one path that was selected by the include patterns
  // and not un-selected by the exclude patterns. That's enough to say we found a change
  // that should be allowed to produce Freight.
  return true
```

The benefits of this implementation are that it returns from the entire function as soon as it finds even one path they is selected and not un-selected. It also leverages short-circuiting in the boolean logic so that we don't ever check against more patterns than we need to.

maksimstankevic: @krancour absolutely, change pushed, I need to improve on my code-owner spirit :)

hiddeco: Think in addition to this, it may be good to make this interface based. I outlined an approach like this in https://github.com/hiddeco/kargo/commit/43573b7d87d348530b585cb51d316156876c247b.

The advantage of this is that include and exclude rules can maintain their declared order of precedence.

krancour: @maksimstankevic it looks like these changes don't avoid as many unnecessary comparison as I had hoped.

What I had in mind was more like this:

```golang
	for _, diffPath := range diffs {
		if len(includePathsRegexps) > 0 || len(includePathsGlobs) > 0 ||
			len(includePathsPrefixes) > 0 {
			// If none of the above had been non-empty, the path is implicitly included
			matches, err := matchesGlobs(diffPath, includePathsGlobs)
			if err != nil {
				return false, fmt.Errorf(
					"syntax error in include glob patterns: %w",
					err,
				)
			}
			if !matches && !matchesPrefix(diffPath, includePathsPrefixes) &&
				!matchesRegexpList(diffPath, includePathsRegexps) {
				// diffPath didn't match any of the include filters
				continue
			}
		}
		// If we get to here, the path was either implicitly or explicitly included
		matches, err := matchesGlobs(diffPath, excludePathsGlobs)
		if err != nil {
			return false, fmt.Errorf(
				"syntax error in exclude glob patterns: %w",
				err,
			)
		}
		if matches || matchesPrefix(diffPath, excludePathsPrefixes) ||
			matchesRegexpList(diffPath, excludePathsRegexps) {
			// diffPath matched an exclude filter
			continue
		}
		// If we get to here, the path was not explicitly excluded
		return true, nil
	}
	return false, nil
```

internal/controller/warehouses/git.go
```diff
@@ -303,75 +306,97 @@ func ignores(tagName string, ignore []string) bool {
...
 
-// compileRegexps is a general purpose function taking a slice of strings
-// as argument and compiling them into a slice of *regexp.Regexp. It
-// returns the compiled regexps in case of success and if it fails to compile
-// a string - nil and error is returned
-func compileRegexps(regexpStrings []string) (regexps []*regexp.Regexp, err error) {
+// sortFilters handles sorting of a slice of strings to regexps,
+// globs and prefixes, regexps are compiled into a slice of
+// *regexp.Regexp additionally
+func sortFilters(regexpStrings []string) (regexps []*regexp.Regexp, globs []string, prefixes []string, err error) {
```
krancour, line 370: `regexpStrings` is probably not an ideal name for this argument given that the input can also contain glob patterns and prefixes.

Note that we also don't use named returns anywhere else in the code base.

internal/controller/warehouses/git.go
```diff
@@ -389,6 +414,37 @@ func matchesRegexpList(stringToMatch string, regexpList []*regexp.Regexp) bool {
...
+	return false, nil
+}
+
+// matchesPrefixList is a general purpose function iterating given slice of
+// strings (prefixList) to check if any of prefixes match stringToMatch string,
+// if match is found it returns true, if match is not found it
+// returns false
+func matchesPrefixList(stringToMatch string, prefixList []string) bool {
+	for _, prefix := range prefixList {
+		if strings.HasPrefix(stringToMatch, prefix) {
```
krancour, line 441: This doesn't look right to me...

This would have `foobar` matching `foo` and I don't think that's what we really intended. Perhaps "prefix" is a poor word choice? Because all along, I thought we've been talking about something more like this:

```golang
func matchesPrefix(path string, prefixes []string) (bool, error) {
	for _, prefix := range prefixes {
		relPath, err := filepath.Rel(prefix, path)
		if err != nil {
			return false, err
		}
		if !strings.Contains(relPath, "..") {
			return true, nil
		}
	}
	return false, nil
}
```

internal/controller/warehouses/git.go
krancour: Actually... hold off on that... @hiddeco and I are discussing another alternative.

krancour: After some back and forth, @hiddeco and I agreed something [along these lines](https://gist.github.com/krancour/876d774469bb36d098798f8c5dfc6521) will check both boxes of avoiding unnecessary comparisons and making the code easier to follow overall.

maksimstankevic: > After some back and forth, @hiddeco and I agreed something [along these lines](https://gist.github.com/krancour/876d774469bb36d098798f8c5dfc6521) will check both boxes of avoiding unnecessary comparisons and making the code easier to follow overall.

@hiddeco @krancour I will refactor and test it out this week

krancour: Thanks @maksimstankevic. We didn't thoroughly test it. It was a "something like this should work." I did run the existing tests though and saw only minor issues that are probably fixed easily. Didn't go too deep.

maksimstankevic: @krancour @hiddeco [along these lines](https://gist.github.com/krancour/876d774469bb36d098798f8c5dfc6521) worked out of the box, no work for me other than testing, changes pushed. Elegant and scalable design.

krancour:
@maksimstankevic I think this looks pretty great. Thanks for the hard work.

At a later date, when documenting the changes, I may adjust the comments on the `IncludePaths` and `ExcludePaths` fields a bit, but this functionality all looks 👍 to me.

LGTM unless @hiddeco has anything more to say about it. (@hiddeco go ahead and merge if you're good.)

krancour (approved):

maksimstankevic:
> Also happy with how this looks, thanks @maksimstankevic! 🍒
> 
> As _another_ potential follow up, I think we could improve the tests a bit to provide some guarantees when people mix-and-match patterns. But that's absolutely non-blocking for now.

Just filed a new issue for this: https://github.com/akuity/kargo/issues/1862, will take it and add thorough testing

thanks @hiddeco !

PR #1779, opened 2024-04-05
feat(cli): add --no-header option to applicable cli usage text

krancour:
Follow-up to #1777 

My apologies that I didn't notice this was missing.

cc @webstradev


webstradev:
Ah yes I was still meaning to add this, was wondering if we need to do anything to signify it is a bool/optional but i guess the name is quite clear

krancour:
> Ah yes I was still meaning to add this

No worries. I think @hiddeco and I accidentally steered you away from this with _other_ changes were discussing and then we all overlooked it. nbd

devholic (approved):

PR #1784, opened 2024-04-06
fix(ui): when serializing resources as json, interpret zero value timestamps as undefined

### LINKED ISSUE #1783: preview modal for analysistemplates and analysis runs both fail
Both fail with errors similar to the following:

```
Uncaught Error: cannot encode field github.com.akuity.kargo.internal.controller.rollouts.api.v1alpha1.AnalysisRun.spec to JSON: cannot encode field github.com.akuity.kargo.internal.controller.rollouts.api.v1alpha1.AnalysisRunSpec.metrics to JSON: cannot encode field github.com.akuity.kargo.internal.controller.rollouts.api.v1alpha1.Metric.provider to JSON: cannot encode field github.com.akuity.kargo.internal.controller.rollouts.api.v1alpha1.MetricProvider.job to JSON: cannot encode field github.com.akuity.kargo.internal.controller.rollouts.api.v1alpha1.JobMetric.metadata to JSON: cannot encode field k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta.creationTimestamp to JSON: invalid date
    writeMessage chunk-6HYTNFIP.js:1350
    toJson chunk-6HYTNFIP.js:149
    AnalysisRunModal analysis-run-modal.tsx:35
    React 18
    batchCalls notifyManager.ts:60
    notifyFn notifyManager.ts:17
    flush notifyManager.ts:72
    flush notifyManager.ts:71
    batchNotifyFn notifyManager.ts:20
    flush notifyManager.ts:70
    setTimeout handler*scheduleFn notifyManager.ts:22
    flush notifyManager.ts:69
    batch notifyManager.ts:36
    dispatch_fn query.ts:601
    setData query.ts:206
    onSuccess query.ts:483
    resolve retryer.ts:103
    promise callback*run retryer.ts:153
    createRetryer retryer.ts:204
    fetch query.ts:469
    executeFetch_fn queryObserver.ts:322
    onSubscribe queryObserver.ts:96
    subscribe subscribable.ts:14
    useBaseQuery useBaseQuery.ts:77
    React 16
    show modal-context.tsx:34
    show use-modal.ts:26
    onClick verifications.tsx:114
    React 
```

krancour:
Fixes #1783 

When serializing resources as JSON, this change will interpret zero value timestamps (i.e. `{seconds: 0, nanos: 0}`) as `undefined` instead of throwing a parsing error.

@rpelczar this solves the problem that we were debugging together earlier.

@hoon I think you'd be the best person to review this. 🙏

### DIFF
--- a/ui/hack/generate-proto-extensions.mjs
+++ b/ui/hack/generate-proto-extensions.mjs
@@ -98,7 +98,11 @@ function extendTime(src) {
       z = "." + nanosStr + "Z";
     }
   }
-  return new Date(ms).toISOString().replace(".000Z", z);
+  try {
+    return new Date(ms).toISOString().replace(".000Z", z);
+  } catch {
+    return undefined;
+  }
 `);
 
   // Add helper methods
--- a/ui/src/gen/k8s.io/apimachinery/pkg/apis/meta/v1/generated_pb.ts
+++ b/ui/src/gen/k8s.io/apimachinery/pkg/apis/meta/v1/generated_pb.ts
@@ -2938,7 +2938,11 @@ export class Time extends Message<Time> {
         z = "." + nanosStr + "Z";
       }
     }
-    return new Date(ms).toISOString().replace(".000Z", z);
+    try {
+      return new Date(ms).toISOString().replace(".000Z", z);
+    } catch {
+      return undefined;
+    }
   }
 
   toDate(): Date {


jessesuen (approved):

devholic:
@krancour Does the zero value timestamps throws error? 

When I try to write a test code against this change (and the previous code), the zero value treated as `1970-01-01T00:00:00Z`, not `undefined` 🤔

```
AssertionError: expected '1970-01-01T00:00:00Z' to be undefined
 ❯ tests/api/apimachinery.spec.ts:28:27
     26|   test('zeroValue', () => {
     27|     const time = new Timestamp({seconds: BigInt(0), nanos: 0});
     28|     expect(time.toJson()).toBeUndefined();
       |                           ^
     29|   });
     30| });

```

krancour:
@devholic I should have been more careful in my use of the phrase "zero value" here.

I think, from Go's perspective, the zero value of `time.Time` (or any of the wrappers around it) is epoch (`1970-01-01T00:00:00Z`).

There are edge cases where a JS client sees a Time of `{"seconds": 0, "nanos": 0}` and this cannot be parsed correctly. (See #1782)

I assume these `{"seconds": 0, "nanos": 0}` values are the result of something lost in translation when the Go resource types are serialized as protobufs and then deserialized as generated JS types.

devholic:
@krancour I still don't get it...

> There are edge cases where a JS client sees a Time of {"seconds": 0, "nanos": 0} and this cannot be parsed correctly.

If so, shouldn't we change `fromBinary` or `fromJson` instead of `toJson`? Since `fromJson` and `fromBinary` are the only methods that involved with parsing logics.

krancour:
toJson gets called by the UI in several places where they are taking an object and trying to prepare its JSON representation for editing.

But I see your point, that this could possibly be fixed in the Go --> proto serialization or the proto --> JS serialization. That is probably preferable, in fact. Perhaps you might want to PR that alternative?

devholic:
Let me clarify: Should the zero value (`{"seconds": 0, "nanos": 0}`) should be treated as `null`? 

Because wrapping date formatting logic (https://github.com/krancour/kargo/blob/bef13801db1ef6e7c3d957cbf5e652fcca9e6fe1/ui/src/gen/k8s.io/apimachinery/pkg/apis/meta/v1/generated_pb.ts#L2941-L2945
) doesn't affect anything since `new Date(0)` is a valid date (which is `1970-01-01T00:00:00Z`).  Instead, we should return `null` (or `undefined`) early when the both `seconds` and `nanos` are zero.

krancour:
It does _something_ because it fixed #1783.

If you want to see, roll back to before this fix and #1783 should be easily reproduced.

Issue #1419, opened 2024-01-24
Standardize on go-logr/logr instead of sirupsen/logrus


jessesuen:
# Proposed Feature

`sirupsen/logrus` is in maintenance mode. `go-logr/logr` is a logging API/wrapper that is agnostic to the actual underlying implementation. 

Many go libraries, including `sigs.k8s.io/controller-runtime/pkg/log`, are starting to standardize on the generic `go-logr/logr` interface, and expect you to set a logr instance if you want log output. e.g.

```golang
package logging

import (
	runtimelog "sigs.k8s.io/controller-runtime/pkg/log"
	"github.com/go-logr/logr"
)


func init() {
	var mylogr logr.Logger
	runtimelog.SetLogger(mylogr)
}
```

NOTE: as I mentioned, `go-logr/logr` is just an interface, we still need to choose the implementation. This can still be `sirupsen/logrus`, or we can choose zap, klog. Recommend moving off `sirupsen/logrus` since it is deprecated, but the first step is to move to the `go-logr/logr` interface.

# Motivation

`sirupsen/logrus` is no longer maintained

# Suggested Implementation

krancour:
💯 on all of this.

webstradev:
Is this specifically for the logging package or for all logging done throughout Kargo.

krancour:
Everywhere, but I wanted to take a closer look at this first. (Or you can?) I have no great love for logrus, but one very important feature of it that I don't want to lose is the ability to add key/val pairs to log entries. Not all loggers have that, unfortunately, and I like keeping the logs at least semi-structured.

@webstradev were you interested in working on this?

webstradev:
Yeah I'm happy to take a look at the implications of this.

One other consequence of switching to logr is not having warning, and fatal log level's.

Any reason

krancour:
> Have you considered slog which was introduced in go 1.21?"

I'd be open to a cook-off, but would also want to hear @jessesuen's opinion.

webstradev:
FYI I'm not very opinionated on the implementation, i have yet to use slog myself other than a quick mess around but it might be worth considering.

krancour:
@webstradev we're super appreciative of the assistance.

bakunowski:
@webstradev Did you get a chance to look at this? I've dug around a little and got a provisional PR, but don't want to step on your toes.

webstradev:
No I haven't had the time yet so feel free to take it over.

krancour:
[...]

PR #1791, opened 2024-04-07
refactor: use go-logr/logr with log/slog backend

bakunowski:
Addresses #1419

This PR standardises on the `go-logr/logr` interface with `log/slog` as the actual logging implementation.

# Level names

There's a little bit of logic on top of the `globalLogger` to achieve clearer `level` field names in log messages. By default `logr` negates the `V()` for `slog`, effectively achieving:

| logr    | slog      |
| --------| ----------|
| V(1)    | -1        |
| V(2)    | -2        |
| ...     | ...       |
| V(4)    | -4 (DEBUG)|

In practice, this means that the following logs:

```go
log.Info("0")
log.V(1).Info("", "slog-level", "1")
log.V(2).Info("", "slog-level", "2")
log.V(3).Info("", "slog-level", "3")
log.V(4).Info("", "slog-level", "4")
```

result in the following log output:

```sh
time=2006-01-02T15:04:05.000-07:00 level=INFO msg=0
time=2006-01-02T15:04:05.000-07:00 level=DEBUG+3 msg="" slog-level=1
time=2006-01-02T15:04:05.000-07:00 level=DEBUG+2 msg="" slog-level=2
time=2006-01-02T15:04:05.000-07:00 level=DEBUG+1 msg="" slog-level=3
time=2006-01-02T15:04:05.000-07:00 level=DEBUG msg="" slog-level=4
```

Using `ReplaceAttr` from `slog.HandlerOptions` allows us to modify it, getting rid of the `+{number}` bit from `level` field. The same logs achieve the following output:

```sh
time=2006-01-02T15:04:05.000-07:00 level=INFO msg=0
time=2006-01-02T15:04:05.000-07:00 level=DEBUG msg="" slog-level=1
time=2006-01-02T15:04:05.000-07:00 level=TRACE msg="" slog-level=2
time=2006-01-02T15:04:05.000-07:00 level=TRACE msg="" slog-level=3
time=2006-01-02T15:04:05.000-07:00 level=TRACE msg="" slog-level=4
```

Currently the lowest defined level is `LevelTrace` at `V(2)` or `slog.Level(-2)`, meaning that any logs with a higher `logr` level will never be shown.

# Warning level

Since `logr` maps to `slog.Level()` by negating the `V()` field, it's impossible to log at warning level. I've added PR comments to highlight instances of `log.Warn` and tried to replace them with what made sense to me.

### DIFF
--- a/cmd/controlplane/api.go
+++ b/cmd/controlplane/api.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"net"
 
-	log "github.com/sirupsen/logrus"
+	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -22,6 +22,7 @@ import (
 	"github.com/akuity/kargo/internal/api/kubernetes"
 	rollouts "github.com/akuity/kargo/internal/controller/rollouts/api/v1alpha1"
 	"github.com/akuity/kargo/internal/kubeclient"
+	"github.com/akuity/kargo/internal/logging"
 	"github.com/akuity/kargo/internal/os"
 	versionpkg "github.com/akuity/kargo/internal/version"
 )
@@ -32,12 +33,12 @@ type apiOptions struct {
 	Host string
 	Port string
 
-	Logger *log.Logger
+	Logger logr.Logger
 }
 
 func newAPICommand() *cobra.Command {
 	cmdOpts := &apiOptions{
-		Logger: log.StandardLogger(),
+		Logger: logging.LoggerFromContext(context.Background()),
 	}
 
 	cmd := &cobra.Command{
@@ -64,10 +65,10 @@ func (o *apiOptions) complete() {
 
 func (o *apiOptions) run(ctx context.Context) error {
 	version := versionpkg.GetVersion()
-	o.Logger.WithFields(log.Fields{
-		"version": version.Version,
-		"commit":  version.GitCommit,
-	}).Info("Starting Kargo API Server")
+	o.Logger.WithValues(
+		"version", version.Version,
+		"commit", version.GitCommit,
+	).Info("Starting Kargo API Server")
 
 	cfg := config.ServerConfigFromEnv()
 
@@ -84,7 +85,7 @@ func (o *apiOptions) run(ctx context.Context) error {
 	case !cfg.RolloutsIntegrationEnabled:
 		o.Logger.Info("Argo Rollouts integration is disabled")
 	case !argoRolloutsExists(ctx, clientCfg):
-		o.Logger.Warn(
+		o.Logger.Info(
 			"Argo Rollouts integration was enabled, but no Argo Rollouts " +
 				"CRDs were found. Proceeding without Argo Rollouts integration.",
 		)
@@ -97,11 +98,11 @@ func (o *apiOptions) run(ctx context.Context) error {
 		o.Logger.Info("admin account is enabled")
 	}
 	if cfg.OIDCConfig != nil {
-		o.Logger.WithFields(log.Fields{
-			"issuerURL":   cfg.OIDCConfig.IssuerURL,
-			"clientID":    cfg.OIDCConfig.ClientID,
-			"cliClientID": cfg.OIDCConfig.CLIClientID,
-		}).Info("SSO via OpenID Connect is enabled")
+		o.Logger.WithValues(
+			"issuerURL", cfg.OIDCConfig.IssuerURL,
+			"clientID", cfg.OIDCConfig.ClientID,
+			"cliClientID", cfg.OIDCConfig.CLIClientID,
+		).Info("SSO via OpenID Connect is enabled")
 	}
 
 	srv := api.NewServer(cfg, kubeClient, internalClient, recorder)
--- a/cmd/controlplane/api.go
+++ b/cmd/controlplane/api.go
@@ -38,7 +38,7 @@ type apiOptions struct {
 
 func newAPICommand() *cobra.Command {
 	cmdOpts := &apiOptions{
-		Logger: logging.LoggerFromContext(context.Background()),
+		Logger: logging.DefaultLogger(),
 	}
 
 	cmd := &cobra.Command{
--- a/cmd/controlplane/api.go
+++ b/cmd/controlplane/api.go
@@ -65,10 +65,11 @@ func (o *apiOptions) complete() {
 
 func (o *apiOptions) run(ctx context.Context) error {
 	version := versionpkg.GetVersion()
-	o.Logger.WithValues(
+	o.Logger.Info(
+		"Starting Kargo API Server",
 		"version", version.Version,
 		"commit", version.GitCommit,
-	).Info("Starting Kargo API Server")
+	)
 
 	cfg := config.ServerConfigFromEnv()
 
@@ -98,11 +99,12 @@ func (o *apiOptions) run(ctx context.Context) error {
 		o.Logger.Info("admin account is enabled")
 	}
 	if cfg.OIDCConfig != nil {
-		o.Logger.WithValues(
+		o.Logger.Info(
+			"SSO via OpenID Connect is enabled",
 			"issuerURL", cfg.OIDCConfig.IssuerURL,
 			"clientID", cfg.OIDCConfig.ClientID,
 			"cliClientID", cfg.OIDCConfig.CLIClientID,
-		).Info("SSO via OpenID Connect is enabled")
+		)
 	}
 
 	srv := api.NewServer(cfg, kubeClient, internalClient, recorder)
--- a/cmd/controlplane/controller.go
+++ b/cmd/controlplane/controller.go
@@ -76,14 +76,11 @@ func (o *controllerOptions) complete() {
 func (o *controllerOptions) run(ctx context.Context) error {
 	version := versionpkg.GetVersion()
 
-	startupLogEntry := o.Logger.WithValues(
-		"version", version.Version,
-		"commit", version.GitCommit,
-	)
+	keyVals := []any{"version", version.Version, "commit", version.GitCommit}
 	if o.ShardName != "" {
-		startupLogEntry = startupLogEntry.WithValues("shard", o.ShardName)
+		keyVals = append(keyVals, "shard", o.ShardName)
 	}
-	startupLogEntry.Info("Starting Kargo Controller")
+	o.Logger.Info("Starting Kargo Controller")
 
 	promotionsReconcilerCfg := promotions.ReconcilerConfigFromEnv()
 	stagesReconcilerCfg := stages.ReconcilerConfigFromEnv()
--- a/cmd/controlplane/controller.go
+++ b/cmd/controlplane/controller.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"sync"
 
-	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -26,6 +25,7 @@ import (
 	"github.com/akuity/kargo/internal/controller/stages"
 	"github.com/akuity/kargo/internal/controller/warehouses"
 	"github.com/akuity/kargo/internal/credentials"
+	"github.com/akuity/kargo/internal/logging"
 	"github.com/akuity/kargo/internal/os"
 	"github.com/akuity/kargo/internal/types"
 	versionpkg "github.com/akuity/kargo/internal/version"
--- a/cmd/controlplane/controller.go
+++ b/cmd/controlplane/controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -41,12 +42,12 @@ type controllerOptions struct {
 	ArgoCDKubeConfig    string
 	ArgoCDNamespaceOnly bool
 
-	Logger *log.Logger
+	Logger logr.Logger
 }
 
 func newControllerCommand() *cobra.Command {
 	cmdOpts := &controllerOptions{
-		Logger: log.StandardLogger(),
+		Logger: logging.DefaultLogger(),
 	}
 
 	cmd := &cobra.Command{
@@ -75,12 +76,12 @@ func (o *controllerOptions) complete() {
 func (o *controllerOptions) run(ctx context.Context) error {
 	version := versionpkg.GetVersion()
 
-	startupLogEntry := o.Logger.WithFields(log.Fields{
-		"version": version.Version,
-		"commit":  version.GitCommit,
-	})
+	startupLogEntry := o.Logger.WithValues(
+		"version", version.Version,
+		"commit", version.GitCommit,
+	)
 	if o.ShardName != "" {
-		startupLogEntry = startupLogEntry.WithField("shard", o.ShardName)
+		startupLogEntry = startupLogEntry.WithValues("shard", o.ShardName)
 	}
 	startupLogEntry.Info("Starting Kargo Controller")
 
@@ -149,7 +150,7 @@ func (o *controllerOptions) setupKargoManager(
 	}
 	if stagesReconcilerCfg.RolloutsIntegrationEnabled {
 		if argoRolloutsExists(ctx, restCfg) {
-			log.Info("Argo Rollouts integration is enabled")
+			o.Logger.Info("Argo Rollouts integration is enabled")
 			if err = rollouts.AddToScheme(scheme); err != nil {
 				return nil, fmt.Errorf(
 					"error adding Argo Rollouts API to Kargo controller manager scheme: %w",
@@ -157,7 +158,7 @@ func (o *controllerOptions) setupKargoManager(
 				)
 			}
 		} else {
-			log.Warn(
+			o.Logger.Info(
 				"Argo Rollouts integration was enabled, but no Argo Rollouts " +
 					"CRDs were found. Proceeding without Argo Rollouts integration.",
 			)
@@ -217,7 +218,7 @@ func (o *controllerOptions) setupArgoCDManager(ctx context.Context) (manager.Man
 	// Application resources in a single namespace, so we will use that
 	// namespace when attempting to determine if Argo CD CRDs are installed.
 	if !argoCDExists(ctx, restCfg, argocdNamespace) {
-		o.Logger.Warn(
+		o.Logger.Info(
 			"Argo CD integration was enabled, but no Argo CD CRDs were found. " +
 				"Proceeding without Argo CD integration.",
 		)
--- a/cmd/controlplane/garbage_collector.go
+++ b/cmd/controlplane/garbage_collector.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 
-	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -17,6 +16,7 @@ import (
 	"github.com/akuity/kargo/internal/api/kubernetes"
 	"github.com/akuity/kargo/internal/garbage"
 	"github.com/akuity/kargo/internal/kubeclient"
+	"github.com/akuity/kargo/internal/logging"
 	"github.com/akuity/kargo/internal/os"
 	versionpkg "github.com/akuity/kargo/internal/version"
 )
--- a/cmd/controlplane/garbage_collector.go
+++ b/cmd/controlplane/garbage_collector.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -24,12 +25,12 @@ import (
 type garbageCollectorOptions struct {
 	KubeConfig string
 
-	Logger *log.Logger
+	Logger logr.Logger
 }
 
 func newGarbageCollectorCommand() *cobra.Command {
 	cmdOpts := &garbageCollectorOptions{
-		Logger: log.StandardLogger(),
+		Logger: logging.DefaultLogger(),
 	}
 
 	cmd := &cobra.Command{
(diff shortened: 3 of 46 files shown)


cmd/controlplane/api.go
```diff
@@ -81,7 +82,7 @@ func (o *apiOptions) run(ctx context.Context) error {
 	case !cfg.RolloutsIntegrationEnabled:
 		o.Logger.Info("Argo Rollouts integration is disabled")
 	case !argoRolloutsExists(ctx, clientCfg):
-		o.Logger.Warn(
+		o.Logger.Info(
```
bakunowski, line 85: `Warn()` replaced with `Info()`

cmd/controlplane/controller.go
```diff
@@ -94,7 +95,7 @@ func newControllerCommand() *cobra.Command {
 							)
 						}
 					} else {
-						log.Warn(
```
bakunowski, line 97: `Warn()` replaced with `Info()`

cmd/controlplane/controller.go
```diff
@@ -188,7 +189,7 @@ func newControllerCommand() *cobra.Command {
 						return fmt.Errorf("error initializing Argo CD Application controller manager: %w", err)
 					}
 				} else {
-					log.Warn(
```
bakunowski, line 191: `Warn()` replaced with `Info()`

internal/controller/stages/stages.go
```diff
@@ -826,14 +825,14 @@ func (r *reconciler) syncNormalStage(
 	// Stop here if we have no chance of finding any Freight to promote.
 	if stage.Spec.Subscriptions == nil ||
 		(stage.Spec.Subscriptions.Warehouse == "" && len(stage.Spec.Subscriptions.UpstreamStages) == 0) {
-		logger.Warn(
```
bakunowski, line 829: `Warn()` replaced with `V(1).Info()` (DEBUG)

internal/controller/warehouses/git.go
```diff
@@ -142,7 +142,7 @@ func (r *reconciler) selectCommitMeta(
 	msg, err := repo.CommitMessage(selectedCommit)
 	if err != nil {
 		// This is best effort, so just log the error
-		logger.Warnf("failed to get message from commit %q: %v", selectedCommit, err)
```
bakunowski, line 145: `Warn()` replaced with `Error()`, since it includes an `err`

internal/credentials/credentials.go
```diff
@@ -110,8 +109,8 @@ func (k *kubernetesDatabase) Get(
 	// If we are dealing with an insecure HTTP endpoint (of any type),
 	// refuse to return any credentials
 	if strings.HasPrefix(repoURL, "http://") {
-		logger := logging.LoggerFromContext(ctx).WithField("repoURL", repoURL)
-		logger.Warnf("refused to get credentials for insecure HTTP endpoint")
```
bakunowski, line 114: `Warn()` replaced with `Info()`

internal/credentials/credentials.go
```diff
@@ -220,10 +219,10 @@ func (k *kubernetesDatabase) getCredentialsSecret(
 		}
 		regex, err := regexp.Compile(string(patternBytes))
 		if err != nil {
-			logger.WithFields(log.Fields{
-				"namespace": namespace,
-				"secret":    secret.Name,
-			}).Warn("failed to compile regex for credential secret")
```
bakunowski, line 226: `Warn()` replaced with `Info()`

cmd/controlplane/api.go
```diff
@@ -31,12 +32,12 @@ type apiOptions struct {
...
 	Port string
 
-	Logger *log.Logger
+	Logger logr.Logger
 }
 
 func newAPICommand() *cobra.Command {
 	cmdOpts := &apiOptions{
-		Logger: log.StandardLogger(),
+		Logger: logging.LoggerFromContext(context.Background()),
```
bakunowski, line 40: I'm not sure if this is a good way to handle creating a new logger here.

krancour: @hiddeco and I are working on refactoring `cmd/controlplane` a bit. We'll probably revisit this later, but this line is ok for now.

krancour: Actually... re-evaluating this, if you could please set this to `log.Log` instead, that would be 👍 

The desire is actually to have "startup" logging be unaffected by the log level selected by whoever installed Kargo. We want info-level messages here, even if the log level is otherwise set to exclude those.

Since `logr` doesn't have a default concrete implementation of the `logr.Logger` interface, the best option here might be to extend our own `logging` package with something like `logging.DefaultLogger()` that would return something like a slog implementation hard-coded to info-level.

This would actually be used by _all_ the `cmd/controlplane` cobra commands.

cmd/controlplane/controller.go
```diff
@@ -44,12 +44,13 @@ func newControllerCommand() *cobra.Command {
 			version := versionpkg.GetVersion()
 			shardName := os.GetEnv("SHARD_NAME", "")
 
-			startupLogEntry := log.WithFields(log.Fields{
-				"version": version.Version,
-				"commit":  version.GitCommit,
-			})
+			log := logging.LoggerFromContext(ctx)
```
krancour, line 47: See [this comment](https://github.com/akuity/kargo/pull/1791/files#r1556679011) -- same concern.

cmd/controlplane/garbage_collector.go
```diff
@@ -31,10 +31,11 @@ func newGarbageCollectorCommand() *cobra.Command {
 			ctx := cmd.Context()
 
 			version := versionpkg.GetVersion()
-			log.WithFields(log.Fields{
-				"version": version.Version,
-				"commit":  version.GitCommit,
-			}).Info("Starting Kargo Garbage Collector")
+			log := logging.LoggerFromContext(ctx)
```
krancour, line 34: See [this comment](https://github.com/akuity/kargo/pull/1791/files#r1556679011) -- same concern.

cmd/controlplane/management_controller.go
```diff
@@ -31,10 +31,11 @@ func newManagementControllerCommand() *cobra.Command {
 
 			version := versionpkg.GetVersion()
 
-			log.WithFields(log.Fields{
-				"version": version.Version,
-				"commit":  version.GitCommit,
-			}).Info("Starting Kargo Management Controller")
+			log := logging.LoggerFromContext(ctx)
```
krancour, line 34: See [this comment](https://github.com/akuity/kargo/pull/1791/files#r1556679011) -- same concern.

cmd/controlplane/webhooks.go
```diff
@@ -35,10 +35,11 @@ func newWebhooksServerCommand() *cobra.Command {
 			ctx := cmd.Context()
 
 			version := versionpkg.GetVersion()
-			log.WithFields(log.Fields{
-				"version": version.Version,
-				"commit":  version.GitCommit,
-			}).Info("Starting Kargo Webhooks Server")
+			log := logging.LoggerFromContext(ctx)
```
krancour, line 38: See [this comment](https://github.com/akuity/kargo/pull/1791/files#r1556679011) -- same concern.

internal/api/kubernetes/client.go
```diff
@@ -578,15 +578,15 @@ func GetRestConfig(ctx context.Context, path string) (*rest.Config, error) {
 	// is empty, but will issue a warning that we can suppress by checking for
 	// that condition ourselves and calling rest.InClusterConfig() directly.
 	if path == "" {
-		logger.Debug("loading in-cluster REST config")
+		logger.V(1).Info("loading in-cluster REST config")
```
krancour, line 581: I feel uncertain about anywhere where we see something like `logger.V(1)`. Doesn't the call to `V()` _change_ the verbosity level? Except for in the startup logs, (as mentioned [here](https://github.com/akuity/kargo/pull/1791/files#r1556679011)) we want to use the level specified by whoever installed Kargo. i.e. If obtained from `logging.LoggerFromContext`, it should already be configured to the correct verbosity.

internal/api/server.go
```diff
@@ -228,9 +227,9 @@ func (s *server) Serve(ctx context.Context, l net.Listener) error {
 		}
 	}()
 
-	log.WithFields(logrus.Fields{
-		"tls": s.cfg.TLSConfig != nil,
-	}).Infof("Server is listening on %q", l.Addr().String())
+	log.WithValues(
+		"tls", s.cfg.TLSConfig != nil,
+	).Info("Server is listening", "port", l.Addr().String())
```
krancour, line 232: This doesn't look quite right.

internal/api/option/log.go
```diff
@@ -81,18 +82,19 @@ func (i *logInterceptor) WrapStreamingHandler(
...
+			level = slog.LevelInfo
 			fields["connect.code"] = connect.CodeOf(err).String()
 		}
 		logging.
 			LoggerFromContext(ctx).
-			WithFields(fields).
-			Log(level, "finished streaming call")
+			WithValues(fields).
+			V(int(level)).
+			Info("finished streaming call")
```
krancour, line 97: I'm not sure this accomplishes the same thing as the old call. This change seems to use a logger with a particular verbosity, but doesn't seem to change the level of the message itself.

internal/controller/promotion/composite.go
```diff
@@ -54,7 +54,7 @@ func (c *compositeMechanism) Promote(
 	newFreight = *newFreight.DeepCopy()
 
 	logger := logging.LoggerFromContext(ctx)
-	logger.Debugf("executing %s", c.name)
+	logger.V(1).Info("executing", "name", c.name)
```
krancour, line 57: It seems as if the level of many of the messages has changed.

cmd/controlplane/controller.go
```diff
@@ -44,12 +44,13 @@ func newControllerCommand() *cobra.Command {
...
+			startupLogEntry := log.WithValues(
+				"version", version.Version,
+				"commit", version.GitCommit,
+			)
 			if shardName != "" {
-				startupLogEntry = startupLogEntry.WithField("shard", shardName)
+				startupLogEntry = startupLogEntry.WithValues("shard", shardName)
 			}
 			startupLogEntry.Info("Starting Kargo Controller")
```
krancour, line 55: I think all of this can be written more cleanly with logr than with logrus. Something like this?

```
keyVals := []any{"version", version.Version, "commit", version.GitCommit}
if shardName != "" {
	keyVals = append(keyVals, "shard", shardName)
}
log.Info("Starting Kargo Controller", keyVals...)
```

internal/controller/promotions/promoqueues.go
```diff
@@ -68,20 +67,20 @@ func (pqs *promoQueues) initializeQueues(ctx context.Context, promos kargoapi.Pr
...
-			logger.WithFields(log.Fields{
-				"stage":     stage.Name,
-				"namespace": stage.Namespace,
-				"depth":     pq.Depth(),
-			}).Debug("Stage-specific Promotion queue initialized")
+			logger.WithValues(
+				"stage", stage.Name,
+				"namespace", stage.Namespace,
+				"depth", pq.Depth(),
+			).V(1).Info("Stage-specific Promotion queue initialized")
```
krancour, line 83: I think this is another case where logr gives us a cleaner way of doing things:

```
logger.Info(
  "Stage-specific Promotion queue initialized",
  "stage", stag.name,
  "namespace", stage.Namespace,
  "depth", pq.Depth(),
)
```

internal/controller/stages/stages.go
```diff
@@ -567,7 +566,7 @@ func (r *reconciler) Reconcile(
 	if err == nil {
 		err = errors.Join(updateErr, clearErr)
 	}
-	logger.Debug("done reconciling Stage")
+	logger.V(1).Info("done reconciling Stage")
```
krancour, line 569: More cases of levels that seem to have changed?

internal/controller/stages/watches.go
```diff
@@ -92,10 +91,11 @@ func (v *verifiedFreightEventHandler) Update(
 				LabelSelector: v.shardSelector,
 			},
 		); err != nil {
-			logger.Errorf(
-				"Failed list Stages downstream from Stage %v in namespace %q",
-				evt.ObjectOld,
-				newFreight.Namespace,
+			logger.Error(
+				nil,
```
krancour, line 95: I know it wasn't included before, but `err` probably should be included. Mind fixing it while we're already touching this?

internal/api/server.go
krancour: Clarifying... `"port"` and `l.Addr().String()` don't look right as a key/val pair because `l.Addr().String()` may often evaluate to something like `0.0.0.0:8080`. Maybe we could use `"bindAddress"` as the key instead.

krancour:
@bakunowski thank you so much for tackling this.

I left a number of comments -- and many of them are applicable in many places other than where I pointed them out. I didn't want to get carried away pointing out every instance of a general something I had feedback on...

Please don't let the number of comments fool you into thinking this PR is anything less than awesome. I am so happy to see progress on this! This is, unfortunately, just what happens when some cross-cutting concern like logging receives a broad update.

We're on the right track!

internal/logging/context.go
```diff
@@ -15,39 +17,107 @@ import (
...
 	return context.WithValue(ctx, loggerContextKey{}, logger)
 }
 
-// LoggerFromContext extracts a *log.Entry from the provided context.Context and
-// returns it. If no *log.Entry is found, a global, error-level *log.Entry is
-// returned.
-func LoggerFromContext(ctx context.Context) *log.Entry {
+// LoggerFromContext extracts a logr.Logger from the provided context.Context and
+// returns it. If no logr.Logger is found, a global logr.Logger is returned.
+func LoggerFromContext(ctx context.Context) logr.Logger {
```
hiddeco, line 90: Wondering if at this point, it wouldn't simply be better to make use of the tooling that comes with controller-runtime: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log#pkg-index

krancour: @hiddeco I've had the same thought previously and have held off so-far because our `logging` package also does some logger initialization based on environment variables. If we stop relying on that, we need to come up with another way of doing that. I'm open to it, but think it should be handled in a follow-up.

bakunowski:
> @bakunowski thank you so much for tackling this.
> 
> I left a number of comments -- and many of them are applicable in many places other than where I pointed them out. I didn't want to get carried away pointing out every instance of a general something I had feedback on...
> 
> Please don't let the number of comments fool you into thinking this PR is anything less than awesome. I am so happy to see progress on this! This is, unfortunately, just what happens when some cross-cutting concern like logging receives a broad update.
> 
> We're on the right track! 🙏

Thanks for the detailed review! Sounds good, I didn't expect this to go down easy 😄 

Sorry about the delay in getting back - last week has been pretty busy. I'll try to address the comments and make changes where it's clear what they should be.

cmd/controlplane/api.go
bakunowski: Let me know if something like [74f006b](https://github.com/akuity/kargo/pull/1791/commits/74f006bf5d9c0967717666186a6482f605a5c11c) makes sense. Since `slog.HandlerOptions` is nil here, it defaults to `LevelInfo`, but I don't know how verbose we want to be in the comments. [Docs](https://pkg.go.dev/golang.org/x/exp/slog#HandlerOptions)

cmd/controlplane/webhooks.go
bakunowski: I'll handle these after a rebase, looks like stuff has moved around a bit.

internal/api/server.go
bakunowski: Should be addressed in [54d2579](https://github.com/akuity/kargo/pull/1791/commits/54d257913008f1415189d1137eddd2cd44b1d237)

cmd/controlplane/controller.go
bakunowski: Thanks, modified in [f9266aa](https://github.com/akuity/kargo/pull/1791/commits/f9266aa3804467d1eb443382fadfdf27be5064ec)

internal/controller/promotions/promoqueues.go
bakunowski: I think I got all of them here [c4c69fb](https://github.com/akuity/kargo/pull/1791/commits/c4c69fb3b0145755839faabbe93bc0c8f1a81c68)

internal/controller/stages/watches.go
bakunowski: Makes sense! [9a2cc02](https://github.com/akuity/kargo/pull/1791/commits/9a2cc028bbb2bdbaf4b4918cf068d1a847cea0e6)

internal/api/option/log.go
bakunowski: Sorry, this one slipped me. I think the level is right, as I attempted to explain [here](https://github.com/akuity/kargo/pull/1791#discussion_r1569044829), but AFAIK there's no way to specify an error level log with `V()`.

I think the changes in [aa1a00d](https://github.com/akuity/kargo/pull/1791/commits/aa1a00dcc1d63480271ae7b336a786119784f938) should put us back on track.

cmd/controlplane/api.go
bakunowski: I've modified the rest of `cmd/controlplane` after the rebase: [aa4336be](https://github.com/akuity/kargo/pull/1791/commits/aa4336be30b3c984dbba124304d9f22995e4b6ee)

krancour: @bakunowski you got it. Exactly what I wanted -- `LevelInfo` in the initial phases of startup, no matter what the user-configured log level actually is.

internal/api/kubernetes/client.go
krancour: 🤦‍♂️ You're right. Although I'm comfortable with the notion of an abstraction over the underlying logger (mainly from my Java days), having not used logr before, I wasn't at all prepared for how wildly different the interface is from what I am accustomed to. My initial reaction is that needing to remember numeric levels as I code seems painful in comparison to `.Debug()` or `.Warn()`. Let me cozy up with the logr docs and try to wrap my head around why this is supposed to be better before commenting further.

krancour: @hiddeco pointed out that realistically, we'd use constants we defined ourselves for the levels.

I'm still left feeling that `logger.V(logging.DEBUG).Info(...)` is all at once wordier _and_ less expressive than `logger.Debug(...)`.

_I'm totally bought into the benefits of logr, but just cannot get over the interface._

Wondering what @jessesuen and @hiddeco might think about a thin wrapper around `logr.Logger` in our existing `logging` package. `logging.DefaultLogger()` and `logging.LoggerFromContext(ctx)` could be modified to return something that looks like this:

```golang
type Logger struct {
	*logr.Logger
}

func (l *Logger) Debug(msg string, keysAndValues ...any) {
	l.Logger.V(1).Info(msg, keysAndValues...)
}

// etc...

func (l *Logger) WithValues(keysAndValues ...any) *Logger {
	newLogger := l.Logger.WithValues(keysAndValues...)
	return &Logger{Logger: &newLogger}
}
```

To be clear... I'm _not_ proposing an "abstraction over an abstraction." I'm proposing fully taking advantage of the logr abstraction (i.e. we can still swap logging back ends no problem), but merely enhancing their interface _without obscuring access_ in case we sometimes need to "drop-down" to it.

hiddeco: On one hand, I feel averse of this idea because it does not allow you to [log errors](https://github.com/go-logr/logr/blob/master/logr.go#L294-L302) at debug level, which a real `logr.Logger` implementation does allow you to do.

I also think (and know based on experience), the problem can largely be worked around by using a modular approach in your application architecture. Allowing you to e.g. inject a `logr.Logger` instance at a predefined level, so that you have to think less about the level which you are logging at.

On the other hand, is the question how often you really have the desire to log an error at debug level, and [isn't this approach really uncommon in our landscape](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b918c3636f5dcc030bbbc5191e757e37df87c682/pkg/logger/logger.go).

krancour: > it does not allow you to [log errors](https://github.com/go-logr/logr/blob/master/logr.go#L294-L302) at debug level

Keep in mind, we wouldn't lose access to the underlying `logr.Logger`.

> Allowing you to e.g. inject a logr.Logger instance at a predefined level

To be clear, my concern so much about obtaining a logger that logs at a certain level of verbosity. My concern is the logr interface makes it harder to express the relative importance of an individual message.

krancour: And lol at that link being almost exactly what I described.

internal/logging/context.go
```diff
@@ -15,39 +17,113 @@ import (
...
+					a.Value = slog.StringValue(ERROR)
+				}
+			}
+			return a
+		},
+	}
+
+	globalLogger = logr.FromSlogHandler(slog.NewTextHandler(os.Stderr, opts))
+
 	SetKLogLevel(os.GetEnvInt("KLOG_LEVEL", 0))
```
hiddeco, line 77: I wonder how much of https://github.com/go-logr/logr#using-a-logrlogsink-as-backend-for-slog is applicable here...

internal/logging/context.go
```diff
@@ -15,39 +17,113 @@ import (
...
-	globalLogger.Logger.SetLevel(level)
+	programLevel.Set(level)
+
+	opts := &slog.HandlerOptions{
+		Level: programLevel,
+		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
+			// Customize the name of the level key and the output string,
+			// including custom level values. Meaning:
+			// log.V(1).Info() becomes level=DEBUG instead of level=DEBUG+3
+			// log.V(2).Info() becomes level=TRACE instead of level=DEBUG+2
```
hiddeco, line 55: It would probably be wise to align this with the upstream levels of our arguably biggest dependency. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use

internal/api/kubernetes/client.go
jessesuen: I don't have a philosophical attachment to the principles of `go-logr`. A lot of the stuff in the go-logr [FAQ](https://github.com/go-logr/logr?tab=readme-ov-file#why-v-levels) is quite principled. I'm not a go-logr purist so we don't have to fully accept their interface as-is. The main use case I was concerned with was that we didn't have a way of setting a logger to `sigs.k8s.io/controller-runtime/pkg/log`, which wants a `logr.Logger`.

I'm fine with creating the wrapper that embeds logr.Logger, especially if it maintains our code readability while also addressing the initial use case. The underlying `logr.Logger` is always available if we wish to use the lower interface.

FWIW `logger.Debug(...)` is definitely preferrable over `logger.V(logging.LevelDebug).Info(...)` and so if we had to switch all our logging statements to the 2nd version, that might be reason enough to create our wrapper.

krancour: Thanks @jessesuen.

@bakunowski are you game for adding a wrapper like the one @hiddeco [linked to](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b918c3636f5dcc030bbbc5191e757e37df87c682/pkg/logger/logger.go) in our existing `logging` package?

krancour:
As we're approaching a v0.6.0 release and there's a sizeable amount of work outstanding here, this is being deferred to v0.7.0.

krancour:
Closing in favor of #2073

PR #1794, opened 2024-04-08
fix(ui): provide initialization value to `metav1.Time` properties

devholic:
Fixes https://github.com/akuity/kargo/issues/1783

UI fails to render `metav1.Time` because properties are omitted during deserialization. However, those omitted number properties should be treated as `zero` by the protobuf speiciation (https://protobuf.dev/programming-guides/proto2/#optional).

This commit narrows down the fix provided in https://github.com/akuity/kargo/issues/1783 by ensuring omitted (undefined) properties to be treated as `zero` by providing initialization value to the properties. Also make sure to `metav1.Time` to be serialized to `null` when the both of properties are `zero` (or undefined) - same as the Go implementation.

### DIFF
--- a/ui/hack/generate-proto-extensions.mjs
+++ b/ui/hack/generate-proto-extensions.mjs
@@ -56,6 +56,18 @@ function extendTime(src) {
   // Extend Time class
   const time = src.getClassOrThrow('Time');
 
+  // Set default value as zero for scalar fields.
+  // From https://protobuf.dev/programming-guides/proto2/#optional:
+  // For numeric types, the default value is zero.
+  time
+    .getPropertyOrThrow('seconds')
+    .setType('bigint')
+    .setInitializer('protoInt64.zero')
+  time
+    .getPropertyOrThrow('nanos')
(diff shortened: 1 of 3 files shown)


krancour:
@devholic have you verified this can handle an `AnalysisTemplate` lile the following?

```yaml
kind: AnalysisTemplate
metadata:
  name: kargo-demo
  namespace: kargo-demo-07
spec:
  metrics:
  - name: test
    provider:
      job:
        metadata:
          annotations:
            foo: bar # annotations defined here will be copied to the Job object
          labels:
            foo: bar # labels defined here will be copied to the Job object
        spec:
          backoffLimit: 1
          template:
            spec:
              containers:
              - name: test
                image: alpine:latest
                command:
                - sleep
                - "20"
              restartPolicy: Never
```

devholic:
@krancour Yes, I verified the UI works like below:

image

krancour:
@devholic thank you so much for improving on my hacky solution!

krancour (approved):

PR #1796, opened 2024-04-08
fix: change placement of freightline tooltips

### LINKED ISSUE #1773: UI: Pop up detail should not cover the Yes / No button
image

I prefer the pop up detail to be moved down, not to cover the YES / NO button.

rbreeze:
Using `placement='bottom'` does not work immediately and would likely require fiddling with CSS for an hour or two... but `placement='right'` works with a one line change. I'd prefer bottom placement but it's not worth the refactoring

Fixes #1773

CleanShot 2024-04-08 at 08 27 32@2x

### DIFF
--- a/ui/src/features/common/freight-label.tsx
+++ b/ui/src/features/common/freight-label.tsx
@@ -37,6 +37,7 @@ export const FreightLabel = ({ freight }: { freight?: Freight }) => {
     >
       {alias || id ? (
         <Tooltip
+          placement='right'
           title={
             <>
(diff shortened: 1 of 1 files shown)


jessesuen (approved):

krancour:
@rbreeze did we verify there is acceptable behavior when the tooltips pop up for Freight on the far right of the Freightline?

rbreeze:
Yep! It automatically switches it to the left hand side.

CleanShot 2024-04-08 at 10 50 36@2x

krancour (approved):

PR #1798, opened 2024-04-08
fix(chart): add AnalysisRun permissions for users

hiddeco:
Without these being present on the roles, it appears that the get functionalities in the UI for AnalysisRuns do not work as expected.

### DIFF
--- a/charts/kargo/templates/users/cluster-roles.yaml
+++ b/charts/kargo/templates/users/cluster-roles.yaml
@@ -64,6 +64,14 @@ rules:
   - analysistemplates
   verbs:
   - "*"
+- apiGroups:
+  - argoproj.io
+  resources:
+  - analysisruns
+  verbs:
+  - get
+  - list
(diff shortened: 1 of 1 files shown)


charts/kargo/templates/users/cluster-roles.yaml
```diff
@@ -64,6 +64,14 @@ rules:
...
   verbs:
   - "*"
+- apiGroups:
+  - argoproj.io
+  resources:
+  - analysisruns
+  verbs:
+  - get
+  - list
+  - watch
```
hiddeco, line 74: I do not think this is _technically_ required.

krancour: It probably is not, but so far I have been in the habit of granting get, list, and watch together for anyone/anything deserving read permissions. For most SAs, all three are required, or else the informer that keeps the client's cache up to date will fail.

PR #1802, opened 2024-04-08
fix: pass version info to UI during build

rbreeze:


Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN pnpm install
 
 COPY ["ui/", "."]
 
-RUN NODE_ENV='production' pnpm run build
+RUN NODE_ENV='production' VERSION=${VERSION} pnpm run build
```
krancour, line 46: For this to work, you need to include `ARG VERSION` somewhere around line 37. (i.e. Between `FROM` and `RUN`)

rbreeze: Oh right, the UI builder is separate... thank you

krancour (approved):

PR #1804, opened 2024-04-09
fix: images table did not load intermittently

### LINKED ISSUE #1775: Every refresh several times you will see "No images available"
https://github.com/akuity/kargo/assets/2680725/edca62e7-1a75-4cc2-bae9-0e7739af4092

rbreeze:
Fixes #1775


ui/src/features/project/pipelines/images.tsx
```diff
@@ -122,12 +122,20 @@ export const Images = ({ projectName, stages }: { projectName: string; stages: S
...
   const [showHistory, setShowHistory] = useLocalStorage(`${projectName}-show-history`, false);
 
+  useEffect(() => {
+    setImageURL(images.keys().next().value as string);
+  }, [images]);
+
+  useEffect(() => {
+    setCurImage(images.get(imageURL));
+  }, [imageURL]);
+
```
rpelczar, line 135: I think you can use `useMemo` here.

rpelczar (changes requested):

ui/src/features/project/pipelines/images.tsx
rbreeze: We can for `curImage` but not `imageURL` because we need to `setImageURL` elsewhere as well

Issue #1451, opened 2024-02-02
bug: app sync error not being factored into stage health


krancour:
@jessesuen pointed out some time ago that some applications exist in a state of perpetual drift, and as such, when determining stage health, we transitioned from looking at sync state to simply verifying sync was complete as evidenced by lack of an operation...

This has worked relatively well, _but_ the change doesn't account for the possibility that there was an _error_ syncing. So the scenario I am running into now is that syncing has complete, but with an error. No resources exist within the app, so Argo CD judges the App to be healthy. Kargo sees a healthy App with no pending operation and thinks all is well.

The change I propose is this:

Return to looking at sync state. If it's synced, great. If it's completed with an error, factor this into stage health. If it's in-progress, _then_ look at whether there is a pending operation or not.

@jessesuen can you sanity check my thought process here before I moved ahead with the fix?

jessesuen:
> Return to looking at sync state. If it's synced, great. If it's completed with an error, factor this into stage health. If it's in-progress, then look at whether there is a pending operation or not.

I'm not sure I'm following the proposal. Argo CD terminology is regrettably confusing (we overused the term sync), so I would like to clarify some of the statements. Argo CD has the following:

* Sync status - a living status whether or not the app is `Synced` or `OutOfSync` with git. Earlier we decided Sync status was irrelevant to Kargo because we can have apps that are perpetually OutOfSync.
* Health status - a living whether or not the application is healthy. Health can flip at any time (e.g. pod restarts). One of: `Progressing`, `Healthy`, `Suspended`, `Degraded`. Health status is relevant to Kargo because we need an application to reach Healthy in order to consider the stage verified
* Operation phase - the phase of the current or last sync operation. This is either `Terminating`, `Failed`, `Error`, `Succeeded`. Kargo does not currently examine operation phase, but I think you are proposing that it should.

This statement is what confused me:
> Return to looking at sync state. If it's synced, great. If it's completed with an error, factor this into stage health. 

I think you are referring to operation phase, since you used the term "completed". But you also referred to it as sync state.

If you are suggesting we start factoring operation phase into the equation, I agree that it may address the issue.

Here's what I'm thinking.
* I still don't think Sync status has any bearing on Kargo and can be ignored.
* A Stage's Health status is a direct mirror of application health status
* Stage's phase is affected by Application Health Status + Operation Phase

[...]

PR #1807, opened 2024-04-09
refactor(controller): improve Argo CD Application health check by taking conditions into account

hiddeco:
Fixes: #1451

Due to how Argo CD determines the "health" state of an Application, certain issues may go unnoticed at present due to it not reflecting certain types of failures as a health issue but rather as a condition in the status of an Application.

This pull request is an attempt to solve this, resulting in the following order to determine if an application (_little a_) is healthy based on the information of an Argo CD Application:

1. The Application does not have pre-flight related error conditions. If it does, it's not healthy and we should report it as "issues".
2. If we have a desired revision, and it's not synced to this revision. It's not healthy.
3. If it hasn't ran into an issue by now, the reported "health" for the Application can be trusted as the best trustworthy health we can get for the application.


internal/argocd/health.go
```diff
@@ -0,0 +1,277 @@
...
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	argocd "github.com/akuity/kargo/internal/controller/argocd/api/v1alpha1"
+	"github.com/akuity/kargo/internal/git"
+)
+
+// healthErrorConditions are the v1alpha1.ApplicationConditionType conditions
+// that indicate an Argo CD Application is unhealthy.
+var healthErrorConditions = []argocd.ApplicationConditionType{
+	argocd.ApplicationConditionComparisonError,
+	argocd.ApplicationConditionInvalidSpecError,
```
hiddeco, line 22: I could use someone with Argo expertise to tell me if there are any other errors which would fall into this category.

internal/controller/stages/stages_test.go
```diff
@@ -354,13 +352,7 @@ func TestSyncNormalStage(t *testing.T) {
...
 			reconciler: &reconciler{
 				hasNonTerminalPromotionsFn: noNonTerminalPromotionsFn,
-				checkHealthFn: func(
-					context.Context,
-					kargoapi.FreightReference,
-					[]kargoapi.ArgoCDAppUpdate,
-				) *kargoapi.Health {
-					return nil
-				},
+				appHealth:                  &mockAppHealthEvaluator{},
```
krancour, line 355: I know you've argued in favor of this approach for some time. I maintain that there's little _technical_ difference between injecting a function directly vs injecting an implementation of an interface that _has_ only one function...... but I'll concede that this is a helluva lot easier on the eyes. 👍

hiddeco: To play advocate of the devil against myself, the way the other function injections could be made easier to the eyes is by creating a mocking function wrapper. E.g.

```go
func mockFunc(expectedReturn string) func(context.Context, ...) {
    return func(context.Context, ...) {
      return expectedReturn
    }
}
```

but at that point... you are almost creating the same LOC as doing mock implementations.

internal/argocd/health.go
```diff
@@ -0,0 +1,278 @@
...
+// errors.Join.
+type compositeError interface {
+	// Unwrap returns the wrapped errors.
+	Unwrap() []error
+}
+
+// ApplicationHealth is a health evaluator for Argo CD Applications.
+type ApplicationHealth struct {
+	Client client.Client
+}
```
krancour, line 35: Nit: This type is a teensy bit brittle because it requires callers outside this package to have knowledge of how to initialize it. (Not that this isn't a pervasive problem in Go due to its lack of actual constructors.)

I would have a slight preference that we do this:

```
type AppHealthEvaluator interface {
	EvaluateHealth(context.Context, kargoapi.FreightReference, []kargoapi.ArgoCDAppUpdate) *kargoapi.Health
}

type appHealthEvaluator struct {
	client client.Client
}

func NewAppHealthEvaluator(c client.Client) AppHealthEvaluator {
	return &appHealthEvaluator {
		client: c,
	}
}

// etc...
```

internal/controller/stages/stages.go
```diff
@@ -43,6 +44,10 @@ func ReconcilerConfigFromEnv() ReconcilerConfig {
 	return cfg
 }
 
+type appHealthEvaluator interface {
+	EvaluateHealth(context.Context, kargoapi.FreightReference, []kargoapi.ArgoCDAppUpdate) *kargoapi.Health
+}
```
krancour, line 49: Would go away if we did [this](https://github.com/akuity/kargo/pull/1807/files#r1558305789).

internal/argocd/health.go
```diff
@@ -0,0 +1,278 @@
...
+
+	return &health
+}
+
+// GetApplicationHealth assesses the health of an Argo CD Application by looking
+// at its conditions, health status, and sync status. Based on these, it returns
+// an overall health state, the Argo CD Application's health status, and its sync
+// status. If it can not (fully) assess the health of the Argo CD Application, it
+// returns an error with a message explaining why.
+func (h *ApplicationHealth) GetApplicationHealth(
```
krancour, line 101: These are the interesting bits. Will 👀 more later this evening.

internal/argocd/health.go
hiddeco: I can live with this suggestion.

The reason for the current code is because I expected the caller to be aware of the structure — which is a constructor by itself. While I did not expect there to be multiple implementations from the side of the `argocd` package — which would be reason to define an interface on this side.

This approach aligns with Go's philosophy of favoring composition over inheritance, which promotes defining interfaces on the side of the user. However, this is also frowned upon by many (including myself at times), and I do agree that your approach may be the nicer way of doing it here.

hiddeco: Moved things around in 3742ef807e93b86de49b15f33c91f52c5e100a08

internal/argocd/health.go
```diff
@@ -0,0 +1,289 @@
...
+				"the health of multi-source Application %q in namespace %q",
+			key.Name,
+			key.Namespace,
+		)
+		return kargoapi.HealthStateUnknown, healthStatus, syncStatus, err
+	}
+
+	// Check for any error conditions. If these are found, the application is
+	// considered unhealthy as they may indicate a problem which can result in
+	// e.g. the health status reporting a false positive.
```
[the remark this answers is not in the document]
hiddeco: Maybe something like this then?

> [..] which can result in e.g. the health status result to become unreliable.

internal/argocd/health.go
```diff
@@ -0,0 +1,289 @@
...
+	argocd.ApplicationConditionComparisonError,
+	argocd.ApplicationConditionInvalidSpecError,
+}
+
+// compositeError is an interface for wrapped standard errors produced by
+// errors.Join.
+type compositeError interface {
+	// Unwrap returns the wrapped errors.
+	Unwrap() []error
+}
```
krancour, line 30: 🤯 that this is not in the stdlib.

krancour:
All of this LGTM. Kudos on thinking to look at conditions _first_.

I defer to @jessesuen and/or @alexmt as better able to judge whether this seems to accomplish a comprehensive assessment of `Application` health.

I can at least verify that it solves the problems I was encountering when I opened #1451.

internal/argocd/health.go
jessesuen: I cannot think of any others. We can start with this and improve it later if any others arise.

krancour (approved):
LGTM after conflicts are resolved.

Issue #782, opened 2023-09-18
Kargo Project page randomly ordered


jessesuen:
Visit kargo project page with 2+ projects. Then click refresh in your browser a bunch of times. The order in which projects are listed will be random each time.

jessesuen:
I filed this under API, instead of UI, because if/when we do pagination, the API server is responsible for returning items in a predictable order.

krancour:
Easy to fix.

PR #1810, opened 2024-04-10
fix(api): order Projects and AnalysisTemplates by name

hiddeco:
Fixes: #782

### DIFF
--- a/internal/api/list_analysistemplates_v1alpha1.go
+++ b/internal/api/list_analysistemplates_v1alpha1.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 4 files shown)


krancour (approved):
LGTM. As a follow-up, I am thinking we really ought to also list Freight and Promotions by date/time -- newest first. It may not affect the UI, but the CLI will benefit from it. @hiddeco wanna take that, by any chance, or at least open an issue for it?

hiddeco:
@krancour I actually checked on those as well, figured they were already sorted in the frontend, and then decided to not do it. But the CLI benefiting from it is actually a good point, and I will follow up.

PR #1811, opened 2024-04-10
chore(deps): bump golangci-lint to 1.57.2

hiddeco:
While dealing with some deprecations.


.golangci.yaml
```diff
@@ -74,17 +75,15 @@ linters-settings:
...
+      - shadow
   lll:
     # max line length, lines longer will be reported. Default is 120.
     # '\t' is counted as 1 character by default, and can be changed with the tab-width option
     line-length: 120
     # tab width in spaces. Default to 1.
     tab-width: 2
-  maligned:
-    # print struct with more effective memory layout or not, false by default
-    suggest-new: true
```
hiddeco, line 87: This has not been working for quite some time, but could theoretically be replaced with the `fieldalignment` option from `govet`. However, I do wonder if the minimal number of bytes this saves is really that important compared to having ordered fields in a struct.

.golangci.yaml
```diff
@@ -94,28 +93,16 @@ linters-settings:
     # make an issue if func has more lines of code than this setting and it has naked returns; default is 30
     max-func-lines: 30
   unparam:
-    # call graph construction algorithm (cha, rta). In general, use cha for libraries,
-    # and rta for programs with main packages. Default is cha.
-    algo: cha
```
hiddeco, line 99: Not an option anymore, see: https://golangci-lint.run/usage/linters/#unparam

.golangci.yaml
```diff
@@ -94,28 +93,16 @@ linters-settings:
...
     # XXX: if you enable this setting, unparam will report a lot of false-positives in text editors:
     # if it's called for subdir of a project it can't find external interfaces. All text editor integrations
     # with golangci-lint call it on a directory with the changed file.
     check-exported: false
-  unused:
-    # treat code as a program (not a library) and report unused exported identifiers; default is false.
-    # XXX: if you enable this setting, unused will report a lot of false-positives in text editors:
-    # if it's called for subdir of a project it can't find funcs usages. All text editor integrations
-    # with golangci-lint call it on a directory with the changed file.
-    check-exported: false
```
hiddeco, line 111: This has not been working for quite some time, and has thus historically been running with default. See: https://golangci-lint.run/usage/linters/#unused

.golangci.yaml
```diff
@@ -94,28 +93,16 @@ linters-settings:
...
   revive:
     # based on https://github.com/mgechev/revive#recommended-configuration
     ignore-generated-header: false
     severity: warning
     confidence: 0.8
-    error-code: 0
-    warning-code: 0
```
hiddeco, line 118: Invalid options, see: https://golangci-lint.run/usage/linters/#revive

krancour (approved):
Feel free to copy these improvements over to Kargo Render also.

PR #1815, opened 2024-04-10
feat: add event details to track DORA metrics

devholic:
Part of https://github.com/akuity/kargo/issues/1231

This commit introduces following annotations to collect deployment-related metrics.

- Freight approved:
  - event.kargo.akuity.io/freight-create-time
- Freight verification:
  - event.kargo.akuity.io/verification-start-time
  - event.kargo.akuity.io/verification-complete-time
- Promotion created:
  - event.kargo.akuity.io/freight-commits
  - event.kargo.akuity.io/freight-images
  - event.kargo.akuity.io/freight-charts


internal/controller/stages/verification.go
```diff
@@ -86,9 +91,10 @@ func (r *reconciler) startVerification(
 			logger.Debug("AnalysisRun already exists for Freight")
 			latestAnalysisRun := analysisRuns.Items[0]
 			return &kargoapi.VerificationInfo{
-				ID:        uuid.NewString(),
-				StartTime: ptr.To(latestAnalysisRun.CreationTimestamp),
-				Phase:     kargoapi.VerificationPhase(latestAnalysisRun.Status.Phase),
+				ID:           uuid.NewString(),
+				StartTime:    ptr.To(latestAnalysisRun.CreationTimestamp),
+				CompleteTime: latestAnalysisRun.Status.CompletedAt(),
```
devholic, line 96: @gdsoumya AnalysisRun's CompletedAt time will be bubbled up to `VerificationInfo` and annotated to make consistent with start time

Brightside56:
@devholic I would like to know also when every single artifact of freight was produced or earliest artifact

krancour:
@Brightside56 I'm not sure (yet?) that that level of detail is appropriate for a Freight created event. It also may not be the best way to surface that information. Let's discuss it in a separate issue please so we don't derail this PR.

api/v1alpha1/event.go
```diff
@@ -83,6 +98,30 @@ func NewPromotionCreatedEventAnnotations(actor string, p *Promotion, f *Freight)
 	}
 	if f != nil {
 		annotations[AnnotationKeyEventFreightAlias] = f.Alias
+		if len(f.Commits) > 0 {
```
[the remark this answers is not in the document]
krancour: I think that's fine. Kubernetes itself has existing precedent for structured data (and JSON in particular) in the last applied annotation. And this data will certainly be small by comparison to that. I see no issue.

api/v1alpha1/stage_types.go
```diff
@@ -163,6 +163,13 @@ type StageSpec struct {
 	Verification *Verification `json:"verification,omitempty" protobuf:"bytes,3,opt,name=verification"`
 }
 
+func (s *StageSpec) HasVerification() bool {
+	if s.Verification != nil && len(s.Verification.AnalysisTemplates) > 0 {
```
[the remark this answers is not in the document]
krancour: `s.Verification` and `len(s.Verification.AnalysisTemplates)` only specify _user-defined_ verification.

_Some kind of verification always takes place._ i.e. The verification phase is never bypassed. It's simply that in some cases, the extent of the verification me simply be waiting for the Stage to become healthy.

internal/controller/stages/stages.go
```diff
@@ -587,7 +594,10 @@ func (r *reconciler) Reconcile(
 func (r *reconciler) syncControlFlowStage(
 	ctx context.Context,
 	stage *kargoapi.Stage,
+	nowFn func() time.Time,
```
krancour, line 597: Don't pass this in... get it from `r.nowFn` when you need it.

internal/controller/stages/stages.go
```diff
@@ -682,7 +696,9 @@ func (r *reconciler) syncControlFlowStage(
 func (r *reconciler) syncNormalStage(
 	ctx context.Context,
 	stage *kargoapi.Stage,
+	nowFn func() time.Time,
```
krancour, line 699: Same as [here](https://github.com/akuity/kargo/pull/1815/files#r1563362958)

krancour:
I've looked at most of this and spotted only minor issues.

@hiddeco should 👀 also.

internal/controller/stages/stages.go
devholic: Resolved in https://github.com/akuity/kargo/pull/1815/commits/fff47409b15ecc2dd8ebed60443f9b17cd8b4aae

internal/controller/stages/stages.go
devholic: Resolved in https://github.com/akuity/kargo/pull/1815/commits/fff47409b15ecc2dd8ebed60443f9b17cd8b4aae

devholic:
@Brightside56

> I would like to know also when every single artifact of freight was produced or earliest artifact. If I understand correctly, with current implementation Kargo will emit freight creation date, but will not take into account that some artifacts were created earlier. For example, say I have two artifacts for a release: application build and templates update. But... templates were upgraded a week after docker build - Kargo will not take it into account

As @krancour mentioned, it would be better to discuss on a separate issue because this PR is for emitting events with the data we already have :)

api/v1alpha1/stage_types.go
```diff
@@ -741,6 +741,8 @@ type VerificationInfo struct {
 	// AnalysisRun is a reference to the Argo Rollouts AnalysisRun that implements
 	// the Verification process.
 	AnalysisRun *AnalysisRunReference `json:"analysisRun,omitempty" protobuf:"bytes,3,opt,name=analysisRun"`
+	// CompleteTime is the time at which the Verification process was completed.
+	CompleteTime *metav1.Time `json:"completeTime,omitempty" protobuf:"bytes,6,opt,name=completeTime"`
```
hiddeco, line 745: Wondering if this should not be:

```suggestion
	// FinishTime is the time at which the Verification process finished.
	FinishTime *metav1.Time `json:"finishTime,omitempty" protobuf:"bytes,6,opt,name=finishTime"`
```

Both because start/finish seems a better pair, and because I think "complete" would imply it ended _successfully_. While it also appears to be set in case of failures.

hiddeco: Ah, reading more it appears to come from the `Completed` method on the AnalysisRun, which also has a nested `FinishedAt`. 😬

Looking back, I like `StartedAt` and `FinishedAt` more than what I introduced with `StartTime`. So I guess you can just pick your poison here...

api/v1alpha1/annotations.go
```diff
@@ -12,12 +12,19 @@ const (
...
-	AnnotationKeyEventPromotionName       = "event.kargo.akuity.io/promotion-name"
-	AnnotationKeyEventPromotionCreateTime = "event.kargo.akuity.io/promotion-create-time"
-	AnnotationKeyEventFreightAlias        = "event.kargo.akuity.io/freight-alias"
-	AnnotationKeyEventFreightName         = "event.kargo.akuity.io/freight-name"
-	AnnotationKeyEventStageName           = "event.kargo.akuity.io/stage-name"
-	AnnotationKeyEventAnalysisRunName     = "event.kargo.akuity.io/analysis-run-name"
+	AnnotationKeyEventActor                    = "event.kargo.akuity.io/actor"
+	AnnotationKeyEventProject                  = "event.kargo.akuity.io/project"
+	AnnotationKeyEventPromotionName            = "event.kargo.akuity.io/promotion-name"
+	AnnotationKeyEventPromotionCreateTime      = "event.kargo.akuity.io/promotion-create-time"
```
hiddeco, line 18:
```suggestion
	AnnotationKeyEventPromotionCreatedTime      = "event.kargo.akuity.io/promotion-created-time"
```

feels more natural.

```suggestion
	AnnotationKeyEventPromotionCreatedAt      = "event.kargo.akuity.io/promotion-created-at"
```

feels more specific.

Same for the `freight-create-time`, `verification-start-time`, etc. bellow.

api/v1alpha1/stage_types.go
devholic: I prefer `_time` because `_at` is a little be ambiguous to me (in perspective of date or time), so I'll go with `finishTime`

hiddeco (commented):
In terms of functionality, it absolutely looks great! Just a few comments around API design choices, and a tiny nit to improve testability. Nothing I couldn't live with however, and none blocking if you happen to not agree :-).

internal/controller/stages/stages.go
```diff
@@ -67,6 +69,8 @@ type reconciler struct {
 
 	// Loop guard:
 
+	nowFn func() time.Time
```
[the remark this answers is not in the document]
devholic: I'm not sure if it's good to have global variable even if it's private, I think there can be a lot of possibilities to be accessed/changed across the same package 🤔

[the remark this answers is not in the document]
hiddeco: The same can be said for the field in your struct. The only potential real problem I can see is parallel testing.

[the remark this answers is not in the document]
devholic: I'll stick to the struct field for now, but feel free to change it :)

PR #1816, opened 2024-04-11
docs: quote uname-arch

blakepettersson:
In zsh this command fails, which is remedied by a bit of quoting.


hiddeco (approved):

krancour:
Happy to merge this, but for everyone's benefit...

An error such as the following:

```shell
zsh: parse error near `)'
```

Is actually due to a setting in [ohmyzsh](https://ohmyz.sh/) that sometimes mangles pasted text.

To resolve _that_ problem (which may bite you in other places than Kargo), find the following in `~/.zshrc`:

```
# DISABLE_MAGIC_FUNCTIONS="true"
```

Uncomment it, save, and reload your shell.

PR #1820, opened 2024-04-11
fix: pass environment variables to sub processes

zkhvan:
This PR allows sub processes to have access to the parent environment variables (usually set on the container).

For example, certain environments use an HTTP_PROXY to forward traffic. In order to support this use-case, Kargo needs to forward the container's environment variables to the sub processes.

The updated processes are `git`, `helm` and `kustomize`. `kargo-render` is already passing the `os.Environ()` down to the sub process (see [here](https://github.com/akuity/kargo/blob/2d9101eebdbe4ddb1b121f2cdbe885a42721e1c4/internal/kargo-render/render.go#L73-L77))

Let me know if you'd like me to create an issue for this also.

### DIFF
--- a/internal/kustomize/kustomize.go
+++ b/internal/kustomize/kustomize.go
@@ -1,6 +1,7 @@
 package kustomize
 
 import (
+	"os"
 	"os/exec"
 
 	libExec "github.com/akuity/kargo/internal/exec"
@@ -22,6 +23,7 @@ func buildSetImageCmd(dir, fqImageRef string) *exec.Cmd {
 		"image",
 		fqImageRef,
 	)
+	cmd.Env = append(cmd.Env, os.Environ()...)
 	cmd.Dir = dir
 	return cmd
 }
--- a/internal/controller/git/git.go
+++ b/internal/controller/git/git.go
@@ -556,12 +556,8 @@ func (r *repo) setupAuth(repoCreds RepoCredentials) error {
 
 func (r *repo) buildCommand(arg ...string) *exec.Cmd {
 	cmd := exec.Command("git", arg...)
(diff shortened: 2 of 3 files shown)


internal/kustomize/kustomize.go
```diff
@@ -22,6 +23,7 @@ func buildSetImageCmd(dir, fqImageRef string) *exec.Cmd {
 		"image",
 		fqImageRef,
 	)
+	cmd.Env = append(cmd.Env, os.Environ()...)
```
zkhvan, line 26: This is actually the default when `cmd.Env == nil` but I've added it here explicitly so that any further `env` customization will be additive.

krancour:
Thanks @zkhvan. This is how things were always _supposed_ to work from the beginning.

imho, it's _surprising behavior_ that `cmd.Env = append(cmd.Env, "FOO=bar")` isn't _always_ additive. i.e. Surprising that `cmd.Env` starts life as `nil` (and _not_ as a copy of `os.Environ()`) and more surprising still that `cmd.Env` of `nil` is treated the same as `os.Environ()` behind the scenes.

It's far too easy to do `cmd.Env = append(cmd.Env, "FOO=bar")` and not realize you're _losing_ a bunch env vars in the process.

We did fix that WRT to Kargo, as you pointed out, and overlooked it in the other places.

🙏 Thank you so much for finding and fixing this.

krancour (approved):

zkhvan:
I think it's one of those scenarios where the default behaviour is predictable, but as soon as you customize any paramaters, they want you to _read the docs_. It's also something I didn't expect, but learned while looking into this.

Is this something that can be cherry-picked into the next `0.5.x` release?

PR #1821, opened 2024-04-11
feat: inject reconcile target details to Promotion controller logger

devholic:
This PR adds Promotion details (namespace and name) to logger to let controller log with contexts to track on.

### DIFF
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -164,7 +164,13 @@ func (r *reconciler) Reconcile(
 	ctx context.Context,
 	req ctrl.Request,
 ) (ctrl.Result, error) {
-	logger := logging.LoggerFromContext(ctx)
+	logger := logging.LoggerFromContext(ctx).
+		WithFields(log.Fields{
+			"namespace": req.NamespacedName.Namespace,
+			"promotion": req.NamespacedName.Name,
+		})
+	ctx = logging.ContextWithLogger(ctx, logger)
+	logger.Debug("reconciling Promotion")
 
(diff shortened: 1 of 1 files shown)


internal/controller/promotions/promotions.go
```diff
@@ -185,8 +185,6 @@ func (r *reconciler) Reconcile(
 		return ctrl.Result{}, fmt.Errorf("error initializing Promotion queues: %w", err)
 	}
 
-	ctx = logging.ContextWithLogger(ctx, logger)
-
```
krancour, line 189: Good eye @devholic. As written, this is indeed redundant...

It was _supposed_ to look like this:

```
logger = logging.LoggerFromContext(ctx).WithFields(log.Fields{
	"namespace": req.NamespacedName.Namespace,
	"promotion": req.NamespacedName.Name,
})
logger.Debug("reconciling Promotion")
```

Instead of removing this line, could we add this additional information? It's valuable context for all other log messages produced by the various functions that this one will call into.

krancour: Note that all the other reconcilers have lines like these. It's a mistake that they were absent here.

devholic: It would be much better :) Followed your suggestion in https://github.com/akuity/kargo/pull/1821/commits/b6099de10a5d262867f35b6c575efe15b2eb4b11 (and updated PR title/desc too)

krancour: Thank you!

krancour (approved):

Issue #1785, opened 2024-04-06
bug: lots of empty strings and collections in yaml editors


krancour:
When JS objects get serialized as JSON, there isn't anything equivalent to Go's `omitempty` option and that leads to seeing a lot of empty strings, collections, etc. in all our YAML editors.

While this may not be harmful, Kubernetes users are definitely not accustomed to seeing resources represented this way.

Example:

<img width="972" alt="Screenshot 2024-04-06 at 1 08 21 AM" src="https://github.com/akuity/kargo/assets/1821014/28a23d9a-8e87-4341-ae67-f68bc8a755d1">

I am wondering if running objects through something like this before stringifying might be a good idea?

```typescript
function cleanObject(obj: any) {
  for (let key in obj) {
    if (obj[key] === "" || obj[key] === null || obj[key] === undefined || (Array.isArray(obj[key]) && obj[key].length === 0)) {
      delete obj[key];
    } else if (typeof obj[key] === 'object' && obj[key] !== null) {
      cleanObject(obj[key]);
    }
  }
  return obj;
}
```

webstradev:
There is an easier way to do this than cleaning the object @krancour .

JSON.stringify can take a replacer function as an argument and then you can set anything you want to undefined which should hide it from the stringify result.

krancour:
Thanks @webstradev! I am admittedly not a front-end guy. I dabble if it's a tiny change, but otherwise put my trust in @rpelczar and @rbreeze. 👍 on this suggestion!

krancour:
Since most languages don't have anything comparable to Go's `omitempty`, the agreement we came to internally was that we would add an option (new request field) to applicable endpoints, for the benefit of non-Go client, to return YAML or JSON as bytes or a string in a new response field.

I believe this will mainly be applicable to _get_ endpoints, so that the back-end can return the JSON or YAML of a single object in preparation for display as a "live manifest" or in preparation for editing in a YAML editor.

* AnalysisRun
* AnalysisTemplate
* Freight
* Promotion
* Stage
* Warehouse

hiddeco:
This got addressed for AnalysisRuns in #1863. But as discovered there (see https://github.com/akuity/kargo/pull/1863#issuecomment-2064491693), the other modals make use of the returned list response at present to display the data.

To solve the issue for these, they need to start using the get endpoints instead.

PR #1827, opened 2024-04-12
feat(api): support raw JSON/YAML response for Get endpoints

hiddeco:
Addresses backend part of: https://github.com/akuity/kargo/issues/1785#issuecomment-2048105883

This adds the capability of requesting the raw JSON or YAML byte for an object, instead of a typed version, while getting objects from the API. When the raw bytes are requested, the typed object is not included in the response to minimize the size of the response data.

While the comment only suggested a selection of the get endpoints, I ended up doing all of them as credentials (Kubernetes Secrets) and Projects include type and object metadata which have quite a few `,omitempty` fields as well.

As the logic to produce the YAML and JSON makes use of Go's generics feature, it should be relatively straight forward to add this to any future get endpoints.


api/service/v1alpha1/service.proto
```diff
@@ -210,10 +216,12 @@ message ListStagesResponse {
 message GetStageRequest {
   string project = 1;
   string name = 2;
+  RawFormat format = 3;
 }
 
 message GetStageResponse {
-  github.com.akuity.kargo.api.v1alpha1.Stage stage = 1;
+  optional github.com.akuity.kargo.api.v1alpha1.Stage stage = 1;
+  optional bytes raw = 2;
```
devholic, line 224: I think we can make this more safer by leveraging [oneOf](https://protobuf.dev/programming-guides/proto3/#oneof) here 🤔 

```suggestion
  oneof data {
    optional github.com.akuity.kargo.api.v1alpha1.Stage stage = 1;
    optional bytes raw = 2;
  }
```

krancour:
Love how you did this @hiddeco. I agree with @devholic's requested change and LGTM after that.

api/service/v1alpha1/service.proto
hiddeco: This is a great suggestion, will adjust it everywhere.

hiddeco: Added in d34283b2d8ef06f2cddb0acb26c366d14a572c2e. One thing to note is the construction of the response is now slightly more awkward, and kind of hard to work around as the interface for `.Result` is private.

devholic (approved):

PR #1828, opened 2024-04-12
feat: add current Promotion name to AnalysisRun label

devholic:
Currently, there is no way to figure out the AnalysisRun is triggered by whether Promotion or reverification.

This PR adds the last Promotion name to the AnalysisRun label if it's not triggered by reverification.

cc @jessesuen @gdsoumya

### DIFF
--- a/internal/controller/stages/verification.go
+++ b/internal/controller/stages/verification.go
@@ -365,7 +365,7 @@ func (r *reconciler) buildAnalysisRun(
 		numAnnotations = len(stage.Spec.Verification.AnalysisRunMetadata.Annotations)
 	}
 	// Kargo will add up to three labels of its own, so size the map accordingly
-	lbls := make(map[string]string, numLabels+3)
+	lbls := make(map[string]string, numLabels+4)
 	annotations := make(map[string]string, numAnnotations)
 	if stage.Spec.Verification.AnalysisRunMetadata != nil {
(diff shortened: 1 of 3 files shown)


internal/controller/stages/verification.go
```diff
@@ -377,6 +377,13 @@ func (r *reconciler) buildAnalysisRun(
 	}
 	lbls[kargoapi.StageLabelKey] = stage.Name
 	lbls[kargoapi.FreightLabelKey] = stage.Status.CurrentFreight.Name
+	// Add Promotion name if the AnalysisRun is triggered by Promotion.
+	// We can determine it by checking existence of Reverify key in annotations.
+	if _, ok := stage.GetAnnotations()[kargoapi.AnnotationKeyReverify]; !ok {
+		if stage.Status.LastPromotion != nil {
+			lbls[kargoapi.PromotionLabelKey] = stage.Status.LastPromotion.Name
+		}
+	}
```
devholic, line 386: @krancour @hiddeco Not sure if it's okay to checking with the existence of Reverify key (Since verification is started after promotion is completed) 👀

krancour: This should be ok.

One small warning... @rbreeze has detected that `LastPromotion` currently only gets set for Promotions that have _succeeded_ and doesn't get set otherwise.

I am working on fixing that. I don't think you should be too impacted by it anyway, since the verification process won't even kick off for a Promotion that failed. So this is just sort of an fyi.

PR #1829, opened 2024-04-12
refactor(controller): do not short-circuit re-promotion to current freight

### LINKED ISSUE #1826: do not short-circuit re-promotion to stage's current freight
At present, if you promote a Stage to a piece of Freight that is _already_ its current piece of Freight, it will short-circuit and succeed immediately with no action taking place.

If the re-promotion is occurring because some sort of undesirable drift has occurred behind Kargo's back, this short-circuiting prevents Kargo from setting things right again. 😞 

We will stop this short-circuiting and permit the promotion mechanisms to fully do their job.

Some promotion mechanisms may, themselves, short-circuit. For instance, when writing to Git, if the repo/branch is found to _already_ be in the desired state, no new commit or PR is required. For a case like integration with an Argo CD App, there will be no short-circuiting and Kargo will cause Apps to refresh and sync.

cc @wanghong230

krancour:
Fixes #1826

@hiddeco I tapped into your re-verification process to trigger re-verification after a successful re-promotion.

### DIFF
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -351,7 +351,6 @@ func (r *reconciler) promote(
 	logger := logging.LoggerFromContext(ctx)
 	stageName := promo.Spec.Stage
 	stageNamespace := promo.Namespace
-	freightName := promo.Spec.Freight
 
 	stage, err := kargoapi.GetStage(
 		ctx,
@@ -368,13 +367,6 @@ func (r *reconciler) promote(
 		return nil, fmt.Errorf("could not find Stage %q in namespace %q", stageName, stageNamespace)
 	}
 	logger.Debug("found associated Stage")
-
-	if stage.Status.CurrentFreight != nil && stage.Status.CurrentFreight.Name == freightName {
-		return &kargoapi.PromotionStatus{
-			Phase:   kargoapi.PromotionPhaseSucceeded,
-			Message: "Stage already has the desired Freight",
-		}, nil
-	}
 
 	targetFreight, err := kargoapi.GetFreight(
 		ctx,
@@ -408,6 +400,8 @@ func (r *reconciler) promote(
 		)
 	}
 
+	logger = logger.WithField("targetFreight", targetFreight.Name)
+
 	targetFreightRef := kargoapi.FreightReference{
 		Name:      targetFreight.Name,
 		Commits:   targetFreight.Commits,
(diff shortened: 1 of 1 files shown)


internal/controller/promotions/promotions.go
```diff
@@ -439,23 +433,34 @@ func (r *reconciler) promote(
 		// so we are safe from race conditions and can just update the status
 		// TODO: remove all patching of Stage status out of promo reconciler
 		if err = kubeclient.PatchStatus(ctx, r.kargoClient, stage, func(status *kargoapi.StageStatus) {
-			// control-flow Stage history is maintained in Stage controller.
-			// So we only modify history for normal Stages.
-			// (Technically, we should prevent creating promotion jobs on
-			// control-flow stages in the first place)
```
krancour, line 445: I'm removing this comment, which seems as if it was somewhat misplaced and just opening an issue for it instead.

krancour: Created #1830

internal/controller/promotions/promotions.go
```diff
@@ -439,23 +433,34 @@ func (r *reconciler) promote(
...
-
 			if newStatus.Phase == kargoapi.PromotionPhaseSucceeded {
-				// Only push promotion to Stage status history if the promotion succeeded
-				status.Phase = kargoapi.StagePhaseVerifying
-				status.CurrentPromotion = nil
-				if stage.Spec.PromotionMechanisms != nil {
+				// Handle specific things that need to happen on success.
+				// 1. Trigger re-verification for re-promotions.
+				// 3. Otherwise, update the current freight and history.
+				// 2. Update the phase to Verifying and clear the current promotion.
```
hiddeco, line 442: Order seems off here.

internal/controller/promotions/promotions.go
```diff
@@ -439,23 +433,34 @@ func (r *reconciler) promote(
...
+				if status.CurrentFreight != nil &&
+					status.CurrentFreight.Name == targetFreight.Name {
+					if err = kargoapi.ReverifyStageFreight(
```
hiddeco, line 445: While this absolutely does the job for now, I can't help but wonder if the Stage reconciler could become more self-determined.

internal/controller/promotions/promotions.go
krancour: 🤦‍♂️ I swear I can count. Will fix.

internal/controller/promotions/promotions.go
krancour: Agreed. @jessesuen and I have both wanted for a long time that each reconciler will only ever mutate the objects it reconciles and not touch others. That's a larger bit of refactoring that needs to happen at some point. Right now, my biggest concern is getting the functionality correct and then we can circle back to figure out how to clean things up a bit. I know you an I both have various improvements in the works that are chipping away at these things.

internal/controller/promotions/promotions.go
krancour: Fixed in recent force push.

hiddeco (approved):

PR #1832, opened 2024-04-12
feat(chart): allow volume(mount)s on Dex server

### LINKED ISSUE #1805: support mounting arbitrary volumes to dex server
This issue generalizes what was discussed in #1793.

Our chart already has the capability of adding arbitrary env vars to the Dex server. This is useful if one wants to manage certain `Secret`s "out-of-band" using Sealed Secrets or similar and then reference them from their Dex configuration.

As it turns out, some Dex connectors (Google being a good example) have need to read separate configuration from the file system.

To facilitate such use cases, we should add the ability to the chart to mount arbitrary volumes to the Dex server.

This should be done in as consistent a manner as possible with how we've done this for Dex server env vars.

hiddeco:
Fixes #1805


charts/kargo/values.yaml
```diff
@@ -196,6 +196,19 @@ api:
       #       name: github-dex
       #       key: dex.github.clientSecret
 
+      ## @param api.oidc.dex.volumes Add additional volumes to Dex pods. This is convenient for cases where api.oidc.dex.connectors needs to reference to mounted data from a Secret that is managed "out of band" with a secret management solution such as Sealed Secrets.
```
krancour, line 199:
```suggestion
      ## @param api.oidc.dex.volumes Add additional volumes to Dex pods. This is convenient for cases where api.oidc.dex.connectors needs to reference mounted data from a Secret that is managed "out of band" with a secret management solution such as Sealed Secrets.
```

charts/kargo/values.yaml
```diff
@@ -196,6 +196,19 @@ api:
...
+      volumes: []
+      # - name: google-json
+      #   secret:
+      #     defaultMode: 420
+      #     secretName: kargo-google-groups-json
+
+      ## @param api.oidc.dex.volumeMounts Add additional volume mounts to Dex pods. This is convenient for cases where api.oidc.dex.connectors needs to reference to mounted data from a Secret that is managed "out of band" with a secret management solution such as Sealed Secrets.
```
krancour, line 206:
```suggestion
      ## @param api.oidc.dex.volumeMounts Add additional volume mounts to Dex pods. This is convenient for cases where api.oidc.dex.connectors needs to reference mounted data from a Secret that is managed "out of band" with a secret management solution such as Sealed Secrets.
```

krancour:
LGTM after the two small changes I suggested (which require code/doc gen be re-run, unfortunately).

krancour (approved):

Issue #1728, opened 2024-04-02
cli: add `kargo server` command


krancour:
Extracted from #1569:

> Jesse:
> * Kargo API server is not exposed in a shared manner
> * End users are given access to a portion of the cluster via kubeconfig
> * Users could still have a UI to do their thing

This would be _different_ from the bygone `--local-server` option because it only did this:

> Kent: adding --local-server to a command launches an ephemeral server that lives for the life of the command it's carrying out.

webstradev:
Is this already done? I seem to recall seeing this command when I was last playing around with The CLI

hiddeco:
This is still an open issue, and I kind-of documented a blueprint for what is required here in https://github.com/akuity/kargo/blob/v0.5.1/internal/cli/cmd/server/server.go#L69-L79

krancour:
> //The server should be started with a Kubernetes client which does NOT 
//   make use of an authorization wrapper. 

This seems as if it's already taken care of.

> It should allow the user to visit the UI in their browser.

I actually solved _part_ of this problem locally and hadn't PR'ed it yet. Will open.

The rest certainly does remain outstanding...

A _big_ gotcha is the UI currently doesn't accommodate the possibility of no authn at all. I'm going to open that specific item as a separate issue.

krancour:
Opened #1834 for the UI to hopefully gain the possibility of skipping the login page.

#1833 embeds UI assets in both the CLI and control plane binaries.

hanshal101:
Is this still a valid issue?

krancour:
@hanshal101 no one has really been pushing for it. I believe this is something we could still entertain if someone really wants it and wants to do the work. That said, #1833 is a pre-requisite for this and also isn't a priority, because the only real reason to tackle #1833 would be to clear the way for this. i.e. Someone would probably have to take ownership of both those issues in order for there to be some movement here.

hanshal101:
Hey @krancour I would like to work on this. Since I'm new to this community could you please guide me on further steps?

krancour:
@hanshal101 do you personally have a use case for this or are you just looking to grab an issue to work on?

This issue was opened to replace similar functionality (`--local-serve`) that was removed, but as I said, I haven't heard of anyone really pushing to get this issue resolved.

The thing I want to avoid is implementing this "just because."

And as I mentioned, #1833 is a prereq for this and seems to me to be a much harder issue than this one.

hanshal101:
[...]

PR #1833, opened 2024-04-12
refactor: embed ui assets in binaries

krancour:
This small change embeds all UI assets into the CLI and control plane binaries, constituting a minor step toward #1728, specifically the "It should allow the user to visit the UI in their browser" part.

Although it makes the server launched by the CLI capable of serving the UI, the UI still doesn't allow for the possibility of not needing to authn.

As I said... a minor step.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -181,6 +181,26 @@ jobs:
         key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
         restore-keys: |
           ${{ runner.os }}-go-
+    - name: Install pnpm
+      uses: pnpm/action-setup@v3
+      with:
+        version: 8
+        run_install: false
+    - name: Install nodejs
+      uses: actions/setup-node@v4
+      with:
+        node-version: "20.12.1"
+        cache: "pnpm"
+        cache-dependency-path: "**/pnpm-lock.yaml"
+    - name: Build UI
+      env:
+        VERSION: ${{ github.ref_name }}
+      working-directory: ./ui
+      run: |
+        pnpm install
+        NODE_ENV=production pnpm run build
+        rm -rf ../internal/api/ui
+        mv build ../internal/api/ui
     - name: Build CLI
       env:
         GOFLAGS: -buildvcs=false
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -189,7 +189,7 @@ jobs:
     - name: Install nodejs
       uses: actions/setup-node@v4
       with:
-        node-version: "20.12.1"
+        node-version: "20.12.2"
         cache: "pnpm"
         cache-dependency-path: "**/pnpm-lock.yaml"
     - name: Build UI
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yaml
```diff
@@ -181,6 +181,26 @@ jobs:
...
+        cache-dependency-path: "**/pnpm-lock.yaml"
+    - name: Build UI
+      env:
+        VERSION: ${{ github.ref_name }}
+      working-directory: ./ui
+      run: |
+        pnpm install
+        NODE_ENV=production pnpm run build
+        rm -rf ../internal/api/ui
+        mv build ../internal/api/ui
```
krancour, line 203: This builds the UI and copies it into a location where it will be included in the CLI build.

This grows the size of the CLI by ~25M, but that is the price for serving the UI from a local server built into the CLI.

Dockerfile
```diff
@@ -1,3 +1,19 @@
...
+WORKDIR /ui
+COPY ["ui/package.json", "ui/pnpm-lock.yaml", "./"]
+ARG VERSION
+
+RUN pnpm install
+
+COPY ["ui/", "."]
+
+RUN NODE_ENV='production' VERSION=${VERSION} pnpm run build
+
```
krancour, line 16: This is moved here, unaltered from a few lines down. This stage needs to precede the back-end build now.

Dockerfile
```diff
@@ -110,7 +111,6 @@ USER root
 
 COPY --from=back-end-builder /kargo/bin/ /usr/local/bin/
 COPY --from=tools /tools/ /usr/local/bin/
-COPY --from=ui-builder /ui/build /ui/build
```
krancour, line 113: No need to copy these files over anymore, because they're already baked into the binary.

internal/api/server.go
```diff
@@ -26,13 +27,15 @@ import (
...
-	httputil "github.com/akuity/kargo/internal/http"
 	"github.com/akuity/kargo/internal/logging"
 	"github.com/akuity/kargo/pkg/api/service/v1alpha1/svcv1alpha1connect"
 )
 
 var (
 	_ svcv1alpha1connect.KargoServiceHandler = &server{}
+
+	//go:embed all:ui
+	ui embed.FS
```
krancour, line 38: Note that this is the reason for the empty `internal/api/ui/.keep` file. It is necessary to avoid compilation errors.

.github/workflows/release.yaml
hiddeco: 1. Shouldn't "unstable" include this as well?
2. I think we now build the UI twice, once in the image and another time for the CLI. It may be better to instead extract this data from the image?

krancour: Don't t recall off the top of my head, but thought we were not publishing unstable CLI nightly. If I'm wrong (quite possible) then I agree. I will double check this when able.

krancour: Ok. Glad you pointed this out. We _were_ publishing the nightly CLI builds. Should be fixed now. I also slightly cleaned up _how_ we were doing those, because we weren't using a matrix-based strategy like we were for the official one.

hiddeco (approved):

PR #1837, opened 2024-04-13
feat: prioritize tag over commit hash in git promotion when available

vavdoshka:
When using tags for git promotion I want to preserve the context of tags in the freights, as well as propagate the tag further down to ArgoCD app revision instead of raw commit hash.

So if Tag info is present in the commit metadata (should be the case for all strategies except `NewestFromBranch`) it will be used as a label in freight cards and its info panel:
image

And in ArgoCD context the target revision would be set to the freight's tag:
image

One caveat is that this might change the behavior for existing users in their freight and ArgoCD apps. It would probably be best to make this behavior opt-in, but I'm not sure what the right approach would be.


krancour:
The back-end aspects of this LGTM.

@rbreeze can you have a look at the front-end aspects of this PR please?

rbreeze (approved):
UI LGTM

krancour (approved):

Issue #1830, opened 2024-04-12
prevent creation of Promotions referencing control flow Stages


krancour:
Preventing these from being created in the first place will, at places in the promotion process, spare us from having to either check or assume that we're dealing with a normal Stage and not a control flow Stage.

PR #1842, opened 2024-04-15
feat: ensure Promotion's Stage has promo mechanism

hiddeco:
Fixes: #1830


krancour:
@hiddeco this looks good, but I think a necessary complement to this may be a change to the "promote to subscribers" endpoint so that we don't attempt creating Promos for control flow Stages that are destined to fail validation.

hiddeco:
@krancour given that we attempt to promote to subscribers while collecting errors instead of failing the operation as a whole, isn't this actually a good way to make "invalid" subscribers surface?

krancour:
@hiddeco consider the possibility that a Stage has many subscribers, some of which are normal stages and some of which are control flow Stages. I think "promote to subscribers" should succeed without any errors by only creating promotions for normal Stages.

krancour (approved):

PR #1843, opened 2024-04-15
feat(controller/charts): support optional commit signing

### LINKED ISSUE #1758: Signing commits when writing changes to Git
# Checklist

## Proposed Feature

Kargo to support signing commits when writing changes to a Git repository. There are three currently supported mechanisms[^1][^2]:

- GPG
- SSH
- S/MIME

## Motivation

Organizations that want to enforce [commit signing](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-signed-commits) on all repositories and branches wouldn't be able to leverage Kargo without adding exceptions for stage-specific branches.

## Suggested Implementation

Depending on which signing mechanisms Kargo wants to leverage, the complexity can vary. As an example for GPG signing, a potential approach could be:

- Update `kargo-render` image to include `gpg` and `gpg-agent`
- Add Name, Email and GPGPrivateKey to ~~Kargo's credential database type~~ a system-level Kargo identity
- Update Git authentication setup to configure commit signing with GPG and import the GPG private key

[^1]: https://git-scm.com/docs/gitformat-signature
[^2]: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

zkhvan:
Fixes #1758


zkhvan:
@krancour this is ready for review 🙏

charts/kargo/templates/controller/deployment.yaml
```diff
@@ -44,18 +44,32 @@ spec:
...
         - configMapRef:
             name: kargo-controller
-        {{- if or .Values.kubeconfigSecrets.kargo .Values.kubeconfigSecrets.argocd }}
+        {{- if or .Values.kubeconfigSecrets.kargo .Values.kubeconfigSecrets.argocd .Values.controller.committer.signingKeySecret.name }}
         volumeMounts:
+        {{- if .Values.controller.committer.signingKeySecret.name }}
+        - mountPath: /etc/kargo/committer
+          name: committer
+          readOnly: true
+        {{- end }}
```
krancour, line 53: Nit: I'd prefer this were the last volumeMount so that the order of the volumeMounts is the same as what's on line 47. I know it's such a little thing, but it templates are hard to read, so every little bit helps.

charts/kargo/templates/controller/deployment.yaml
```diff
@@ -44,18 +44,32 @@ spec:
...
           {{- toYaml .Values.controller.resources | nindent 10 }}
-      {{- if or .Values.kubeconfigSecrets.kargo .Values.kubeconfigSecrets.argocd }}
+      {{- if or .Values.kubeconfigSecrets.kargo .Values.kubeconfigSecrets.argocd .Values.controller.committer.signingKeySecret.name }}
       volumes:
+      {{- if .Values.controller.committer.signingKeySecret.name }}
+      - name: committer
+        secret:
+          secretName: {{ .Values.controller.committer.signingKeySecret.name }}
+          defaultMode: 0644
+      {{- end }}
```
krancour, line 71: Similar nit as [here](https://github.com/akuity/kargo/pull/1843/files#r1570916323)

charts/kargo/values.yaml
```diff
@@ -268,6 +268,16 @@ controller:
     ## @param controller.globalCredentials.namespaces List of namespaces to look for shared credentials.
     namespaces: []
 
+  committer:
```
krancour, line 271: I doubt this is the last git client customization option we'll provide, so it may be better to name this `gitClient` because it seems like a bucket we'll put more things in in the future.

charts/kargo/values.yaml
```diff
@@ -268,6 +268,16 @@ controller:
...
+    ## @param controller.committer.name Specifies the name of the Kargo controller committer.
+    name: "Kargo Render"
+    ## @param controller.committer.email Specifies the email of the Kargo controller committer.
+    email: "kargo-render@akuity.io"
+
+    signingKeySecret:
+    ## @param controller.committer.signingKeySecret.name Specifies the name of an existing `Secret` which contains the committer's signing key. The value should be accessible under `.data.signingKey` in the same namespace as Kargo.
+      name: ""
```
krancour, line 279: We can probably do without nesting `signingKeySecret.name`. Can `signingKeySecret` just be a string field instead of an object? _Or_ were you anticipating that there may be other settings that get added here in the future?

charts/kargo/templates/controller/deployment.yaml
zkhvan: Good call, updated:

[refactor(charts): order volumes and mounts more logically](https://github.com/akuity/kargo/pull/1843/commits/5e41a69a00a3ac71e2f51afe7c201dc49119b7c8)

charts/kargo/templates/controller/deployment.yaml
zkhvan: Updated

[refactor(charts): order volumes and mounts more logically](https://github.com/akuity/kargo/pull/1843/commits/5e41a69a00a3ac71e2f51afe7c201dc49119b7c8)

internal/controller/git/git.go
```diff
@@ -30,6 +30,17 @@ type RepoCredentials struct {
 	Password string `json:"password,omitempty"`
 }
 
+// CommitUser represents the user for committing changes to a git repository.
+type CommitUser struct {
```
hiddeco, line 34: I think this should probably be called `Author` (or `CommitAuthor`). This pairs better with the method you introduced, and avoids using new terminology for something which has already a well established name within the domain (Git).

hiddeco: Thinking this over once more... The terminology of Git is three fold.

You can have a "user", which can either be;

1. the author of a patch.
2. the committer of a patch.
3. the author of an annotated tag.

On a desktop computer, this is configured as `git config user.name | git config user.email`. Given this, I would just call the struct `User` or `GitUser`, which makes it suitable to be used for any of the above (and beyond just commits).

charts/kargo/values.yaml
krancour: sgtm

internal/controller/git/git.go
```diff
@@ -480,18 +495,56 @@ func (r *repo) WorkingDir() string {
...
+		author.Email = "kargo-render@akuity.io"
+	}
+
+	cmd = r.buildGitCommand("config", "--global", "user.email", author.Email)
+	cmd.Dir = r.homeDir // Override the cmd.Dir that's set by r.buildGitCommand()
+	if _, err := libExec.Exec(cmd); err != nil {
+		return fmt.Errorf("error configuring git user email: %w", err)
+	}
+
+	if author.SigningKeyPath != "" {
```
hiddeco, line 519: Git has support for signing commits with [SSH keys](https://docs.gitlab.com/ee/user/project/repository/signed_commits/ssh.html). I think we should take this into account, and should not automatically assume keys to be of a `gpg` type.

internal/controller/git/git.go
zkhvan: I will go with `User`, as the package is `git`, the usage will be `git.User`. Thanks for the suggestion!

krancour:
@zkhvan first off, this PR looks fantastic overall.

Something I am thinking about though... the `repo.SetAuthor()` func feels odd to me...

I can see this two ways. Either:

1. Author info, including key is a commit option that should be passed at the time of commit _OR_
2. Since you're doing `git config --global` calls, we could treat this a bit more like "client's identity" and include it with new `ClientOptions` that could be passed to the `git.Clone()` func.

I feel as if no.2 feels a little bit cleaner.

So something like this:

```
type ClientOptions struct {
  Name string
  Email string
  Credentials *RepoCredentials
  SigningKeyPath string
}
```

Signature of `git.Clone()` changes:

```
func Clone(repoURL string, clientOpts *ClientOptions, cloneOpts *CloneOptions) (Repo, error)
```

Signatue (_and_ name of `setupAuth()` func changes):

```
func (r *repo) setupClient(opts *ClientOptions) error
```

Obviously, some of these changes will have a little bit of a ripple effect, but shouldn't be too bad and, overall, I think this will be nice because after a call to `git.Clone()`, the repo you get back is 100% ready to go (even for commits) with no additional function calls necessary.

These small changes would help keep the `git` package intuitive to use.

krancour:
Something else I was curious about is I don't see where `user.signingkey` ever gets set. Doesn't that need to be set to the id of the gpg key that gets imported? Else, I'm not sure how git knows _what_ gpg key to sign with.

My own `~.gitconfig` looks something like this:

```
[commit]
	gpgSign = true
[user]
	email = <email here>
	signingkey = <id here>. # <---- where is this being set?
	name = Kent Rancourt
```

zkhvan:
> Something I am thinking about though... the `repo.SetAuthor()` func feels odd to me...

Initially, I had a `repo.Configure(...RepoOption)` type of signature, but couldn't find other use-cases for doing repo-specific configuration other than setting author. It looked a bit silly as:

```go
repo.Configure(
   repo.SetAuthor(author),
)
```

thus, I opted to keep it simpler with just the one method.

> 2. Since you're doing `git config --global` calls, we could treat this a bit more like "client's identity" and include it with new `ClientOptions` that could be passed to the `git.Clone()` func.

I considered something similar to this (updating `setupAuth` to setup the signing key), the only _strange_ part for me was that author information isn't really required when cloning a repository for read-only purposes, with the Warehouse being the primary example. I didn't want to **unnecessarily** configure the name, email and signing key.

I think updating `setupAuth` to `setupClient` with `ClientOptions` will be a good idea and allows the consumer to configure how they want to leverage the repo during cloning. Happy to make this change! :)

> Something else I was curious about is I don't see where user.signingkey ever gets set. Doesn't that need to be set to the id of the gpg key that gets imported? Else, I'm not sure how git knows what gpg key to sign with.

This is only necessary if the user's name and email don't match the key. Otherwise, Git will automatically detect the GPG key based on your available trustdb (at least in my experience). My `gitconfig` looks something like:

```config
[user]
	name = "Zhenya Khvan"
	email = "<email address>"

[commit]
	gpgSign = true
```

internal/controller/git/git.go
krancour: Let's actually handle ssh key signing in a follow-up and for now, just be sure we don't paint ourselves into a corner that will make that harder later.

krancour:
> I didn't want to unnecessarily configure the name, email and signing key.

Fair point. I think we could still keep this seamless behind `git.Clone()` and just don't pass those options when the Warehouse clones a repo.

krancour:
> This is only necessary if the user's name and email don't match the key.

TIL!

Thanks @zkhvan. If this is the case, maybe we should just document the expectation in the `values.yaml` that these must match.

zkhvan:
I rebased on `main` since there was a conflict on the `Dockerfile`.

The new changes are after the commit:

[refactor(charts/controller): generalize committer into gitClient](https://github.com/akuity/kargo/pull/1843/commits/fc0b1556e146f07d3f141fdcdabbd5dd00c038a8)

I think it's ready for review!

Also, I made a small unrelated tweak since my Tilt was having an issue with `pnpm` with the latest changes: [fix(build): match ui-dev pnpm version with package manager](https://github.com/akuity/kargo/pull/1843/commits/5f3316a9e6faf95c1b641c9321cd6a5806c285f8). I can revert this if it's just me and I might have to clear some cache...

```log
     Building image
     [ui-dev 1/6] FROM docker.io/library/node:20.12.2@sha256:844b41cf784f66d7920fd673f7af54ca7b81e289985edc6cd864e7d05e0d133c
     [background] read source files 5.26MB [done: 73ms]
     [ui-dev 5/6] RUN pnpm install
     [ui-dev 4/6] COPY [ui/package.json, ui/pnpm-lock.yaml, ./] [cached]
     [ui-dev 3/6] WORKDIR /ui [cached]
     [ui-dev 2/6] RUN npm install --global pnpm [cached]
       →  ERR_PNPM_BAD_PM_VERSION  This project is configured to use v9.0.2 of pnpm. Your current pnpm is v9.0.3
       → 
       → If you want to bypass this version check, you can set the "package-manager-strict" configuration to "false" or set the "COREPACK_ENABLE_STRICT" environment variable to "0"
     
     ERROR IN: [ui-dev 5/6] RUN pnpm install
```

internal/controller/git/git.go
zkhvan: I went with the first option, defaulting to `gpg`. But all the mechanisms are there to allow us to add SSH signing support in a follow-up.

Dockerfile
```diff
@@ -90,7 +106,7 @@ CMD ["/usr/local/bin/kargo"]
 ####################################################################################################
 FROM --platform=$BUILDPLATFORM docker.io/library/node:20.12.2 AS ui-dev
 
-RUN npm install --global pnpm
+RUN npm install --global pnpm@9.0.2
```
krancour, line 109: This is fixed in `main` already, but we fixed it by upgrading the version of pnpm that our `package.json` is compatible with.

zkhvan: I think a new version just got released: https://github.com/pnpm/pnpm/releases/tag/v9.0.3

internal/controller/git/git.go
```diff
@@ -480,24 +511,68 @@ func (r *repo) WorkingDir() string {
...
-// SetupAuth configures the git CLI for authentication using either SSH or the
-// "store" (username/password-based) credential helper.
-func (r *repo) setupAuth(repoCreds RepoCredentials) error {
-	// Configure the git client
-	cmd := r.buildCommand("config", "--global", "user.name", "Kargo Render")
-	cmd.Dir = r.homeDir // Override the cmd.Dir that's set by r.buildCommand()
+// setupClient configures the git CLI for authentication using either SSH or
+// the "store" (username/password-based) credential helper.
+func (r *repo) setupClient(opts *ClientOptions) error {
+	if opts.User != nil {
```
krancour, line 517: I think we need another nil check for `opts` itself. We can return right away if it's nil.

zkhvan: Gotcha, added. I opted to set a default value, in-case there's any other "setup" behaviour that needs to happen by default, it won't be accidentally skipped.

[fix(controller): set default client options](https://github.com/akuity/kargo/pull/1843/commits/25df2fe97b5052f0a3a434b306c20bd4bbfe6967)

charts/kargo/values.yaml
zkhvan: [refactor(charts/controller): generalize committer into gitClient](https://github.com/akuity/kargo/pull/1843/commits/fc0b1556e146f07d3f141fdcdabbd5dd00c038a8)

internal/controller/git/git.go
zkhvan: [refactor(controller): rename commit user to user](https://github.com/akuity/kargo/pull/1843/commits/01f9006f35b3f7d17209d55c11fa343a8c047535)

charts/kargo/values.yaml
zkhvan: Already have an example 😄 

[feat(charts): add signing key type to indicate how to sign commits wi…](https://github.com/akuity/kargo/pull/1843/commits/08134a4cba6301de5dbe24b30e7617ff65cf2cc1)

krancour (approved):

PR #1848, opened 2024-04-16
fix(promotions): log event for Argo CD Application

### LINKED ISSUE #1845: Kargo initiated Argo CD syncs do not emit Events
# Proposed Feature

When Kargo performs an Argo CD sync on an application, it does so by setting the `.operation` field of the Application (the same way that Argo CD API and controller do)

However, because we set this operation ourselves, it misses the logic that the argocd-application-controller and argocd-server currently have to emit the corresponding Kubernetes Event, which users rely on for audit records.

When Kargo is the one initiating the sync, it should also emit this Event. 

# Motivation

Without this, the sync will not show up in event history

# Suggested Implementation

Emit a Kubernetes event in the same format as Argo CD during sync.

hiddeco:
Fixes #1845

### DIFF
--- a/internal/controller/promotion/argocd.go
+++ b/internal/controller/promotion/argocd.go
@@ -49,7 +49,7 @@ type argoCDMechanism struct {
 		patch client.Patch,
 		opts ...client.PatchOption,
 	) error
-	logAppEventFn func(ctx context.Context, app *argocd.Application, reason, message string)
+	logAppEventFn func(ctx context.Context, app *argocd.Application, user, reason, message string)
 }
 
 // newArgoCDMechanism returns an implementation of the Mechanism interface that
@@ -178,8 +178,7 @@ func (a *argoCDMechanism) doSingleUpdate(
 			app.Spec.Sources[i] = source
 		}
 	}
-	app.ObjectMeta.Annotations[argocd.AnnotationKeyRefresh] =
-		string(argocd.RefreshTypeHard)
+	app.ObjectMeta.Annotations[argocd.AnnotationKeyRefresh] = string(argocd.RefreshTypeHard)
 	app.Operation = &argocd.Operation{
 		InitiatedBy: argocd.OperationInitiator{
 			Username:  "kargo-controller",
@@ -207,8 +206,7 @@ func (a *argoCDMechanism) doSingleUpdate(
 		app.Operation.Sync.Revisions = []string{app.Spec.Source.TargetRevision}
 	}
 	for _, source := range app.Spec.Sources {
-		app.Operation.Sync.Revisions =
-			append(app.Operation.Sync.Revisions, source.TargetRevision)
+		app.Operation.Sync.Revisions = append(app.Operation.Sync.Revisions, source.TargetRevision)
 	}
 	if err = a.argoCDAppPatchFn(
 		ctx,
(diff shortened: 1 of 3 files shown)


internal/controller/promotion/argocd.go
```diff
@@ -215,9 +219,46 @@ func (a *argoCDMechanism) doSingleUpdate(
...
 	return nil
 }
 
+func (a *argoCDMechanism) logAppEvent(ctx context.Context, app *argocd.Application, reason, message string) {
+	logger := logging.LoggerFromContext(ctx)
+
+	t := metav1.Time{Time: time.Now()}
+	event := corev1.Event{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%v.%x", app.Name, t.UnixNano()),
```
hiddeco, line 234: As we are working with a shim of the Argo CD API, we would have to introduce more structures to be able to use the same annotations as being set in https://github.com/argoproj/argo-cd/blob/820f4d861a7789f299143ed89816001091abf923/util/argo/audit_logger.go#L119-L120

In addition, because we are working with a partial spec. I was not sure if we should include it as being done in: https://github.com/argoproj/argo-cd/blob/820f4d861a7789f299143ed89816001091abf923/util/argo/audit_logger.go#L125

internal/controller/promotion/argocd.go
```diff
@@ -215,9 +219,46 @@ func (a *argoCDMechanism) doSingleUpdate(
 	}
 	logging.LoggerFromContext(ctx).WithField("app", app.Name).
 		Debug("patched Argo CD Application")
+
+	a.logAppEventFn(ctx, app, "Promotion", "Promotion triggered a sync of this Application resource.")
```
hiddeco, line 223: Do we want this message to mimic the message Argo CD itself would emit (https://github.com/argoproj/argo-cd/blob/44894e9e438bca5adccf58d2f904adc63365805c/server/application/application.go#L1887-L1895), or do we want it to be Kargo specific?

krancour:
@jessesuen I hate to bug you, but your 👀 might be best on this particular issue.

internal/controller/promotion/argocd.go
hiddeco: While I do agree that's easier for Event consumers, it does feel like that underlines the current shortcoming in Argo CD itself which @krancour took note of in https://github.com/akuity/kargo/issues/1845#issuecomment-2059138876.

krancour: Still not an expert on k8s events yet, but there must be some way to distinguish that, even though it looks just like an Argo CD event, Kargo wrote it, right?

IF that is so the event already has a metaphorical asterisk next to it.

hiddeco: You can't really, as you produce the event _for_ an involved object. Which is the Argo CD Application in this case.

The only sign we give people is the injection of an `kargo-controller` user at the beginning of the message string.

internal/controller/promotion/argocd.go
```diff
@@ -217,24 +215,46 @@ func (a *argoCDMechanism) doSingleUpdate(
...
 	return nil
 }
 
-func (a *argoCDMechanism) logAppEvent(ctx context.Context, app *argocd.Application, reason, message string) {
-	logger := logging.LoggerFromContext(ctx)
+func (a *argoCDMechanism) logAppEvent(ctx context.Context, app *argocd.Application, user, reason, message string) {
+	logger := logging.LoggerFromContext(ctx).WithField("app", app.Name)
+
+	// xref: https://github.com/argoproj/argo-cd/blob/44894e9e438bca5adccf58d2f904adc63365805c/server/application/application.go#L2145-L2147
+	// nolint:lll
```
hiddeco, line 241: The fact that I had to put this here has refueled my dislike of certain linters.

internal/controller/promotion/argocd.go
krancour: So basically the only telltale sign that it came from us and not from Argo CD itself is in an unstructured field?

hiddeco: That's correct.

internal/controller/promotion/argocd.go
jessesuen: I'm checking with some folks, I think for completeness and the way we parse these events (outside of Kargo), we may end up wanting the other annotations.

krancour: @jessesuen I think we need to make a decision soon on if this gets into v0.6.0 or gets deferred.

jessesuen: Confirmed that the annotations are not needed.

jessesuen (approved):

Issue #1851, opened 2024-04-16
small bit of promotion reconciler cleanup


krancour:
Follow up to #1739 

I notice in #1739 that the main `Reconcile` function in the Promotions reconciler is now retrieving Freight to get its alias.

This raises a question, and the answer dictates what action this issue requires...

Q: Aliases are mutable. Given this fact, is this something we want to bother capturing in the promotion event?

If no, then let's remove this.

If yes, let's take the Freight that's retrieved in `Reconcile` and pass it in the call to `r.promoteFn()` so that function, which also retrieves the Freight, will no longer need to. I don't want to retrieve the Freight twice in the course of a single reconcilation loop if that's something we can avoid.

PR #1853, opened 2024-04-16
feat: refactor kargo events

gdsoumya:
This PR :
1. Fixes https://github.com/akuity/kargo/issues/1851
2. Adds freight creation time to all promotion/freight events

### DIFF
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -215,7 +215,12 @@ func (r *reconciler) Reconcile(
 		Name:      promo.Spec.Freight,
 	})
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("get freight: %w", err)
+		return ctrl.Result{}, fmt.Errorf(
+			"error finding Freight %q in namespace %q: %w",
+			promo.Spec.Freight,
+			promo.Namespace,
+			err,
+		)
 	}
 
 	logger = logger.WithFields(log.Fields{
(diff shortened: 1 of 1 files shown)


internal/controller/promotions/promotions.go
```diff
@@ -67,7 +67,7 @@ type reconciler struct {
 		types.NamespacedName,
 	) (*kargoapi.Stage, error)
 
-	promoteFn func(context.Context, kargoapi.Promotion) (*kargoapi.PromotionStatus, error)
+	promoteFn func(context.Context, kargoapi.Promotion, *kargoapi.Freight) (*kargoapi.PromotionStatus, error)
```
devholic, line 70: As @krancour pointed out, I think it makes more sense to show the status (alias in here) of freight at the time the user queries events. (Because there can be a possibilities to mislead wrong reference). So I think it would be better to remove Freight alias from event itself.

devholic (changes requested):

internal/controller/promotions/promotions.go
```diff
@@ -399,14 +401,6 @@ func (r *reconciler) promote(
...
 	logger.Debug("found associated Stage")
 
-	targetFreight, err := kargoapi.GetFreight(
-		ctx,
-		r.kargoClient,
-		types.NamespacedName{
-			Namespace: promo.Namespace,
-			Name:      promo.Spec.Freight,
-		},
-	)
```
krancour, line 409: I think the err check and nil check below this need to be removed as well.

krancour: Actually... I just noticed that this nil check is missing from the Reconcile method where you obtain the Freight that you pass to this function. It should be added.

krancour: We spoke offline and I understand now that Freight being not found would have no bearing on whether you emit the event or not, so leaving the nil check here makes sense to me.

krancour (approved):

internal/controller/promotions/promotions.go
krancour: My mention that alias was mutable was an fyi... if you think there's value to recording the alias by which the freight was known at the time of the promotion, I defer you your judgement. As long as the actual freight name is recorded as well, all is good.

devholic (approved):

PR #1854, opened 2024-04-16
fix: rewrite and test git url normalization logic

### LINKED ISSUE #1852: Adding Credentials with a repo with a trailing slash prevents promotion
# Checklist

* [ ] I've pasted the output of `kargo version`.

# Description

So working on a SaaS demo and running into a promotion issues.

This doesn't work (gives me "username/password" is invalid)

```shell
kargo create credentials --project=kargo-demo kargo-demo-repo --git --repo-url=https://github.com/christianh814/kargo-simple-demo/ --username=christianh814 --password=${TOKEN}
```

This, works, for me (gives a successful promotion of freigh)

```shell
kargo create credentials --project=kargo-demo kargo-demo-repo --git --repo-url=https://github.com/christianh814/kargo-simple-demo --username=christianh814 --password=${TOKEN}
```

# Steps to Reproduce

Create creds with the Kargo CLI with a trailing slash on SAAS

# Version

v0.5.1

# Logs

SaaS doesn't give me logs

krancour:
Fixes #1852

@jessesuen do you think I should PR this to Argo CD also?

Note to self: Backport this to `release-0.5` after it is merged to `main`.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -9,7 +9,15 @@ import (
 
 var scpSyntaxRegex = regexp.MustCompile(`^((?:[\w-]+@)?[\w-]+(?:\.[\w-]+)*)(?::(.*))?$`)
 
-// NormalizeGitURL normalizes a Git URL for purposes of comparison.
+// NormalizeGitURL normalizes Git URLs of the following forms:
+//
+//   - http[s]://[proxy-user@proxy-pass:]host.xz[:port][/path/to/repo[.git][/]]
+//   - ssh://[user@]host.xz[:port][/path/to/repo[.git][/]]
+//   - [user@]host.xz[:path/to/repo[.git][/]]
+//
+// This is useful for the purposes of comparison and also in cases where a
+// canonical representation of a Git URL is needed. Any URL that cannot be
+// normalized will be returned as-is.
 func NormalizeGitURL(repo string) string {
 	origRepo := repo
 	repo = strings.ToLower(repo)
@@ -18,13 +26,11 @@ func NormalizeGitURL(repo string) string {
 	if strings.HasPrefix(repo, "http://") || strings.HasPrefix(repo, "https://") {
 		repoURL, err := url.Parse(repo)
 		if err != nil {
-			panic(fmt.Errorf("error normalizing HTTP/S URL %s: %w", origRepo, err))
+			return origRepo
 		}
 		if len(repoURL.Query()) > 0 {
-			panic(fmt.Errorf(
-				"error normalizing HTTP/S URL %s: query parameters are not permitted",
-				origRepo,
-			))
+			// Query parameters are not permitted
+			return origRepo
 		}
 		repoURL.User = nil // Remove user info if there is any
 		repoURL.Path = strings.TrimSuffix(repoURL.Path, "/")
@@ -37,13 +43,11 @@ func NormalizeGitURL(repo string) string {
 		// repo = strings.TrimPrefix(repo, "ssh://")
 		repoURL, err := url.Parse(repo)
 		if err != nil {
-			panic(fmt.Errorf("error normalizing SSH URL %s: %w", origRepo, err))
+			return origRepo
 		}
 		if len(repoURL.Query()) > 0 {
-			panic(fmt.Errorf(
-				"error SSH URL %s: query parameters are not permitted",
-				origRepo,
-			))
+			// Query parameters are not permitted
+			return origRepo
 		}
 		repoURL.Path = strings.TrimSuffix(repoURL.Path, "/")
 		repoURL.Path = strings.TrimSuffix(repoURL.Path, ".git")
@@ -53,10 +57,8 @@ func NormalizeGitURL(repo string) string {
 	// URLS of the form [user@]host.xz[:path/to/repo[.git][/]]
 	matches := scpSyntaxRegex.FindStringSubmatch(repo)
 	if len(matches) != 2 && len(matches) != 3 {
-		panic(fmt.Errorf(
-			"error normalizing URL: %s does not appear to be a valid HTTP/S, SSH, or SCP-style URL",
-			origRepo,
-		))
+		// This URL doesn't appear to be in a format we recognize
+		return origRepo
 	}
 	userHost := matches[1]
 	var path string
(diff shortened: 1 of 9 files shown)


hiddeco (commented):
We should probably back fill tests for this so we can provide guarantees on supported formats. But for a hotfix, this looks good to me.

hiddeco (approved):

krancour:
> We should probably back fill tests for this so we can provide guarantees on supported formats. But for a hotfix, this looks good to me.

I'd be happy to include that in this PR actually.

The genesis of this is that the function was cribbed verbatim from Argo CD and there was no test over there either. I didn't bother writing one here because I considered it adequately battle-hardened already, but now that we've found a fault in it... yep... I will add the tests and will PR this fix _and_ the tests to Argo CD.

krancour:
> We should probably back fill tests for this

This was well worth doing, because it turns out the existing function was pretty buggy.

I will offer these improvements to Argo CD as well.

jessesuen:
For Kargo, this LGTM. 

I just did a search of how `NormalizeGitURL` is used in Argo CD. I think there might be one place of concern:

https://github.com/argoproj/argo-cd/blob/db615ed1c55dde361287726eb2748b4a3fb2d0b1/pkg/apis/application/v1alpha1/app_project_types.go#L401-L423

Argo CD calls `NormalizeGitURL` when deciding if a git repo is permitted in a project. I'm having trouble understanding if removing vs. preserving the slash would change the behavior of what is permitted in the project or not. But it could be addressed in an Argo CD PR.

jessesuen (approved):

internal/git/git.go
```diff
@@ -1,57 +1,42 @@
...
 )
 
-// NormalizeGitURL normalizes a git URL for purposes of comparison, as well as preventing redundant
-// local clones (by normalizing various forms of a URL to a consistent location).
-// Prefer using SameURL() over this function when possible. This algorithm may change over time
-// and should not be considered stable from release to release
+// https://regex101.com/r/bJCECT/1
+var sshURLRegex = regexp.MustCompile(`^(?:ssh://)?((?:[\w-]+@)[\w-]+(?:\.[\w-]+)*(?::\d+)?)(?::(.*))?`)
+
+// NormalizeGitURL normalizes a git URL for purposes of comparison.
```
hiddeco, line 13: Nit:

```suggestion
// NormalizeGitURL normalizes a Git URL for purposes of comparison.
```

hiddeco (approved):

krancour:
@hiddeco and I had some discussion and I think we agreed in these things:

* I was missing support for URLs of the form `ssh://[user@]host[:port][/path]`
* SCP-style syntax is `[user@]host[:path]` and makes no provision for port number
* Query params are not allowed

I also realized that if any user/password info for an outbound proxy happened to be present in an http/s URL, that really should be stripped from the normalized representation. Hopefully no one would ever use such a URL anyway. Supposing someone did, we need to acknowledge that `https://foo:bar@github.com/example/repo` and `https://bat:baz@github.com/example/repo` both still _do_ point to the same thing.

Have a look at the updated test cases for an updated and comprehensive look at what the latest changes support and don't.

There is also another improvement in the latest commit. Unlike the old logic, we don't return an empty string when we cannot normalize a URL, and unlike the logic proposed previously, we also no longer panic when we cannot normalize a URL. What we do now for anything that cannot be normalized for any reason is simply return the input, unaltered...

This makes sense because it now means:

* `git.NormalizeURL("something non-normalizable") == git.NormalizeURL("something non-normalizable")`
* `git.NormalizeURL("something non-normalizable") != git.NormalizeURL("something ELSE non-normalizable")`

Asking for this to be re-reviewed, please.

internal/git/git.go
```diff
@@ -1,57 +1,78 @@
...
-	}
-	normalized := repoURL.String()
-	return strings.TrimPrefix(normalized, "ssh://")
-}
+var scpSyntaxRegex = regexp.MustCompile(`^((?:[\w-]+@)?[\w-]+(?:\.[\w-]+)*)(?::(.*))?$`)
 
-var sshURLRegex = regexp.MustCompile("^(ssh://)?([^/:]*?)@[^@]+$")
+// NormalizeURL normalizes Git URLs of the following forms:
+//
+//   - http[s]://[proxy-user@proxy-pass:]host.xz[:port][/path/to/repo[.git][/]]
```
krancour, line 14:
```suggestion
//   - http[s]://[proxy-user:proxy-pass@]host.xz[:port][/path/to/repo[.git][/]]
```

internal/git/git_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func TestNormalizeGitURL(t *testing.T) {
+	testCases := map[string]string{
+		// Anything we can't normalize should be returned as-is
+		"https://not a url":                      "https://not a url",
+		"http://github.com/example/repo?foo=bar": "http://github.com/example/repo?foo=bar",
+		"ssh://not a url":                        "ssh://not a url",
+		"ssh://github.com/example/repo?foo=bar":  "ssh://github.com/example/repo?foo=bar",
+		"not even remotely a url":                "not even remotely a url",
+		// URLs of the form http[s]://[proxy-user@proxy-pass:]host.xz[:port][/path/to/repo[.git][/]]
```
krancour, line 17:
```suggestion
		// URLs of the form http[s]://[proxy-user:proxy-pass@]host.xz[:port][/path/to/repo[.git][/]]
```

hiddeco (approved):
Improvements appear to check all the boxes, and really happy with it not causing a panic anymore. 💯

internal/git/git_test.go
```diff
@@ -0,0 +1,90 @@
...
+		"ssh://git@github.com/example/repo/":         "ssh://git@github.com/example/repo",     // 1001
+		"ssh://git@github.com/example/repo.git":      "ssh://git@github.com/example/repo",     // 1010
+		"ssh://git@github.com/example/repo.git/":     "ssh://git@github.com/example/repo",     // 1011
+		"ssh://git@localhost:2222/example/repo":      "ssh://git@localhost:2222/example/repo", // 1100
+		"ssh://git@localhost:2222/example/repo/":     "ssh://git@localhost:2222/example/repo", // 1101
+		"ssh://git@localhost:2222/example/repo.git":  "ssh://git@localhost:2222/example/repo", // 1110
+		"ssh://git@localhost:2222/example/repo.git/": "ssh://git@localhost:2222/example/repo", // 1111
+		// SCP-style URLs of the form [user@]host.xz[:path/to/repo[.git][/]]
+		"git.example.com":     "ssh://git.example.com",
+		"git@git.example.com": "ssh://git@git.example.com",
```
krancour, line 71:
```suggestion
		"git.example.com":      "ssh://git.example.com",
		"git.example.com/":     "ssh://git.example.com",
		"git@git.example.com":  "ssh://git@git.example.com",
		"git@git.example.com/": "ssh://git@git.example.com",
```

PR #1856, opened 2024-04-16
refactor(controller): rework annotation mechanics

hiddeco:
Related to: #1479 
Also addresses a bit of: #1231 (tracking who aborted a verification request)

This pull request refactors the approach to working with annotations that express desires, such as "refresh this resource" or "abort this verification." It eliminates the need to delete an annotation once a request has been fulfilled, by introducing carefully scoped predicates and comparisons against the Status of a resource to determine if a request is new.

There are multiple advantages to this approach:

1. It reduces the number of API calls during reconciliation, as we no longer need to issue a second PATCH operation to remove an annotation.
2. It removes the need to ignore update events from the PATCH operation mentioned in point 1.
3. It makes the annotations compatible with GitOps, as they can now be safely declared in the desired state (e.g., due to a YAML export at an unfortunate moment) without retriggering the request on every sync.
4. We avoid queuing for _any_ annotation change.

#### `kargo.akuity.io/refresh`

Instead of deleting the annotation at the end of the reconciliation, the token value is mirrored to the `.status.lastHandledRefresh` field of the resource. This allows the initiator of the refresh request to wait until the value in the Status field matches the token of their request or equals the current annotation value.

To enable a reconciler to detect new refresh requests, a `RefreshRequested` predicate is introduced. This predicate signals that a resource should be reconciled if the annotation is added or if the annotation value has changed compared to the previous annotation value.

#### `kargo.akuity/reverify`

Instead of having two separate annotations to put in a request to reverify the current Freight of a Stage and to keep track of the actor who put in the request. The values have been unified into the `kargo.akuity/reverify` annotation which is now expected to contain a JSON object with all information with the following structure:

```go
type VerificationRequest struct {
	// ID is the identifier of the VerificationInfo for which the request is
	// being made.
	ID string `json:"id,omitempty"`
	// Actor is the user who initiated the request.
	Actor string `json:"actor,omitempty"`
	// ControlPlane is a flag to indicate if the request has been initiated by
	// a control plane.
	ControlPlane bool `json:"controlPlane,omitempty"`
}
```

To not create a burden for e.g. `kubectl` users, a Stage can still be annotated using a simple ID string value in which case the webhook will take care of transforming this into the expected JSON payload (including information about the actor).

During reconciliation, the data from the annotation value is compared to the `.currentFreight.verficationHistory` to determine if a reverification should be performed. In addition, when it's determined that a verification attempt is due to a reverify request, the actor of this request is mirrored to the `.verificationInfo.actor` field. When a Kubernetes Event is emitted for the verification process, the value of this field is used instead of annotation data to include metadata about the actor with the Event.

To enable a reconciler to detect new reverify requests, a `ReverifyRequested` predicate is introduced. This predicate signals that a resource should be reconciled if the annotation is added or if the ID in the annotation payload has changed compared to the previous ID.

#### `kargo.akuity.io/abort`

This annotation now relies on the same mechanics as introduced for `kargo.akuity/reverify`, and allows for tracking the actor which initiated the request.

To enable a reconciler to detect new abort requests, an `AbortRequested` predicate is introduced. This predicate signals that a resource should be reconciled if the annotation is added or if the ID in the annotation payload has changed compared to the previous ID.

Only small modifications were required to the reconciliation logic itself, as the abort request was already taken into account only when the verification process is still running _and_ the value matches the ID of the current verification process. Since the former condition will be false as soon as the abort request is fulfilled, this minimized the changes to only mirroring the actor from the `VerificationRequest` to the `VerificationInfo` in the Status.

---

@devholic: specific commit of interest to you is 6cd501356dc51a3734396d6e2cd6aac5841fbf12.
@rbreeze, @rpelczar: I tried to make the frontend changes myself here... They are all related to the `kargo.akuity.io/refresh` annotation, and mainly involve waiting for the annotation value to equal the field in the Status – instead of the annotation being removed.


krancour:
@hiddeco despite the size, this looks very straightforward. Love this improvement.

LGTM, but I still want @rbreeze or @rpelczar to look at the UI parts.

krancour (approved):

hiddeco:
Before I merge, it would be great to have an approval from @devholic as well. Given I (carefully) reconstructed bits he has been eagerly working on, and I want to be sure I do not introduce any unforeseen side-effects.

devholic (approved):
Thanks for the awesome work! 🚀 LGTM

rpelczar (approved):
UI LGTM

PR #1859, opened 2024-04-17
fix: create a new Event on every correlated event

devholic:
Part of https://github.com/akuity/kargo/issues/1231

The default Kubernetes event recorder aggregates (patches) Event instead of creating them for correlated events. Since we heavily depend on Kubernetes events to track Kargo events and metrics, it has been difficult to track exact events.

This commit fixes this behavior by creating a custom event recorder that creates a new Event for every emitted event.

---

#### References

- [client-go/tools/record/event.go](https://github.com/kubernetes/client-go/blob/aa7909e7d7c0661792ba21b9e882f3cd6ad0ce53/tools/record/event.go#L288-L304)


internal/kubeclient/event.go
```diff
@@ -0,0 +1,139 @@
...
+	newEventHandlerFn func(event *corev1.Event) func() error
+}
+
+// NewEventRecorder returns a new event recorder that records all events
+// without aggregation, even the given event is correlated.
+//
+// NOTE: This recorder must be used with caution as it creates a new Event
+// on every event without throttling / spam filtering features - which are
+// included in the event recorder from k8s.io/client-go. This could lead
+// to performance issues with the Kubernetes API server.
```
krancour, line 34: I don't know a _ton_ about k8s events. This note gives me pause. Are we absolutely sure we're using it with the requisite caution?

internal/kubeclient/event.go
```diff
@@ -0,0 +1,139 @@
+package kubeclient
```
krancour, line 1: This package has been driving me a little crazy for a while. I think we've had a tendency to dump a lot of unrelated k8s things in here.

Right now I see indexers, rountrippers, helpers for patching resources, error-handling stuff.

I might prefer that we put this code somewhere else, but I don't have any specific concerns about the implementation -- just the location.

krancour:
Noted a few minor things, but high level, this LGTM.

krancour (approved):

internal/kubeclient/event.go
devholic: By default, Kubernetes API server will delete newly created events after 1 hours (configured by [`--event-ttl` flag](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)).
I think it will be fine in most cases since we are trying to emit event once - so it depends on user's usage.

But we can talk about this more in separate issue for sure 🙏

internal/kubeclient/event.go
devholic: There are already lots of packages in the `internal` package now, I think it would be better to group `kubernetes` related resources into one. So created a new package named `kubernetes` and moved event related packages under `kubernetes/event` in https://github.com/akuity/kargo/pull/1859/commits/a8e1c2853f624cc100d62d4a414828e0f1106708 :)

PR #1863, opened 2024-04-18
fix(ui): use YAML response option for AnalysisRun

hiddeco:
Part of: #1785

### DIFF
--- a/ui/src/features/stage/analysis-run-modal.tsx
+++ b/ui/src/features/stage/analysis-run-modal.tsx
@@ -6,6 +6,7 @@ import yaml from 'yaml';
 import YamlEditor from '@ui/features/common/code-editor/yaml-editor-lazy';
 import { ModalProps } from '@ui/features/common/modal/use-modal';
 import { getAnalysisRun } from '@ui/gen/service/v1alpha1/service-KargoService_connectquery';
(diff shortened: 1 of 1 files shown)


hiddeco:
The other ones appear to be building their "view" from the list endpoints, not the get endpoints...

krancour:
> The other ones appear to be building their "view" from the list endpoints, not the get endpoints...

That seems odd. Do we know why we're not using get for those?

@rpelczar or @rbreeze?

hiddeco:
The reason I think, is that to display an overview of things we already list the objects. Which means that it's "cheaper" to pass the data from the list response to the modal.

For AnalysisRuns however, the list is built up from reference data from the Stage. Which means we need to get the further data to display the modal.

krancour:
> The reason I think, is that to display an overview of things we already list the objects. Which means that it's "cheaper" to pass the data from the list response to the modal.

I don't mind the minor expense if it buys us more up-to-date data and cleaner YAML.

hiddeco:
I don't mind either, but I fear it's beyond my Typescript skill level to make that happen quickly and/or in a neat way 😅

rpelczar (approved):

krancour:
> I fear it's beyond my Typescript skill level to make that happen quickly and/or in a neat way

You and me both, @hiddeco.

Just open a follow-up and let's merge this.

PR #1865, opened 2024-04-18
refactor: use normalized URLs in constructing freight id

krancour:
Follow-up to #1854.

I've realized that we should use normalized URLs when constructing Freight IDs.

This way, if someone updates a subscription in a Warehouse to use a new URL that is semantically equivalent to the old URL, we won't end up with a new piece of Freight just because of that.

### DIFF
--- a/api/v1alpha1/freight_types.go
+++ b/api/v1alpha1/freight_types.go
@@ -7,6 +7,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/akuity/kargo/internal/git"
+	"github.com/akuity/kargo/internal/helm"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


hiddeco (approved):
```
  ______________
< Good thinking! >
  --------------
         \   ^__^ 
          \  (oo)\_______
             (__)\       )\/\\
                 ||----w |
                 ||     ||
```

jessesuen:
It's a good idea. Is a consequence of this that after upgrading, we might generate one more extra freight because the git URL normalized to something different than before?

krancour:
> It's a good idea. Is a consequence of this that after upgrading, we might generate one more extra freight because the git URL normalized to something different than before?

@jessesuen good question.

The recent updates _mainly_ affect ssh URLs and validation on the Warehouses isn't even letting those in the door at the moment. We'll eventually have all the ducks in the row to support it.

The only possibility of a new Freight you didn't expect is if you subscribe to `https://github.com/something-like/this/`, because that will now normalize as `https://github.com/something-like/this` (no trailing slash, see #1852)

PR #1868, opened 2024-04-18
fix: ensure old object to be assigned in webhook

devholic:
Controller runtime doesn't decode old object, which causes `req.oldObject.Object` to be always nil.

This commit fixes this issue by manually decoding old object with the `req.oldObject.Raw`.

### DIFF
--- a/internal/webhook/stage/webhook.go
+++ b/internal/webhook/stage/webhook.go
@@ -25,7 +25,8 @@ var (
 )
 
 type webhook struct {
-	client client.Client
+	client  client.Client
+	decoder *admission.Decoder
 
 	// The following behaviors are overridable for testing purposes:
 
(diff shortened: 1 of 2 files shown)


internal/webhook/stage/webhook.go
```diff
@@ -92,8 +99,13 @@ func (w *webhook) Default(ctx context.Context, obj runtime.Object) error {
 	}
 
 	var oldStage *kargoapi.Stage
-	if req.OldObject.Object != nil {
-		oldStage = req.OldObject.Object.(*kargoapi.Stage) // nolint: forcetypeassert
+	// We need to decode old object manually since controller-runtime doesn't decode it for us.
+	if req.Operation == admissionv1.Update ||
+		req.Operation == admissionv1.Delete {
```
hiddeco, line 104:
```suggestion
	if req.Operation == admissionv1.Update {
```

would probably be better, as we do not need the decoded object on delete.

hiddeco (approved):

PR #1871, opened 2024-04-18
feat: display events in UI

### LINKED ISSUE #1818: UI tab for Kargo Events

rbreeze:
Fixes #1818 

CleanShot 2024-04-18 at 15 47 57@2x
CleanShot 2024-04-18 at 15 48 03@2x


api/service/v1alpha1/service.proto
```diff
@@ -579,3 +583,11 @@ message DeleteAnalysisTemplateRequest {
 message DeleteAnalysisTemplateResponse {
   /* explicitly empty */
 }
+
+message ListProjectEventsRequest {
+  string project = 1;
```
devholic, line 588: nit: We may need to introduce pagination in the future since we emit events on every Kargo events 👀

devholic (commented):
Q. Should we apply filter to show Kargo events only? Or do we want to show every events (including Kubernetes events) in the project(namespace)?

krancour:
@rbreeze @devholic where do we stand on this PR? I believe this was demo'ed with great success on Friday.

api/service/v1alpha1/service.proto
rbreeze: IMO we should delay pagination for now (esp. given the 0.6 deadline looming). Events are ephemeral anyways so it's unlikely there will be overload.

devholic:
@krancour sorry for the late reply. Both @rbreeze and I agreed to list Kargo related events only in the `ListProjectEvents` handler.

The change (https://github.com/akuity/kargo/pull/1871/commits/76bb46514763e9e2db07b539ff23ae28c78140ea) introduces new field indexer for listing events by the `involvedObject`'s API group.

internal/api/list_project_events_v1alpha1.go
```diff
@@ -0,0 +1,59 @@
...
+	var eventsList corev1.EventList
+	if err := s.client.List(
+		ctx,
+		&eventsList,
+		client.InNamespace(req.Msg.GetProject()),
+		// List Kargo related events only
+		client.MatchingFieldsSelector{
+			Selector: fields.OneTermEqualSelector(
+				kubeclient.EventsByInvolvedObjectAPIGroupIndexField,
+				kargoapi.GroupVersion.Group,
```
devholic, line 40: As @rbreeze pointed out API version can be changed later on, so I decided to use `Group` only instead of `GroupVersion`, WDYT?

internal/kubeclient/indexer.go
```diff
@@ -36,6 +39,25 @@ const (
...
+
+func indexEventsByInvolvedObjectAPIGroup(obj client.Object) []string {
+	event := obj.(*corev1.Event) // nolint: forcetypeassert
+	// Ignore invalid APIVersion
+	gv, _ := schema.ParseGroupVersion(event.InvolvedObject.APIVersion)
+	if gv.Empty() {
+		return nil
+	}
+	return []string{gv.Group}
+}
```
devholic, line 59: Another idea for here is to create `kargoEvent` field with (string formatted) boolean value to distinguish Kargo related events or not. It might be helpful when we introduce other API group other than `kargo.akuity.io` 🤔

devholic (approved):

internal/kubeclient/indexer.go
rbreeze: That's a good idea, but I think in the interest of getting this into 0.6, we should make this change as a later improvement

PR #1876, opened 2024-04-19
Added annotations: {} to api service.yaml and api.service in values.yaml

ed-boykin:
The specific need to add annotations to the API service manifest was to add an annotation to indicate an internal or external load balancer when running on AKS.


ed-boykin:
Forgot the DCO and borked the rebase to add it.

PR #1874, opened 2024-04-19
feat(chart): Add optional env and envFrom parameters for all deployments and cronjobs

MPritsch:
Adds the possibility to add env values or envFrom values for all deployments and cronjobs. We use kargo on a cluster behind a webproxy and needed a way to configure http_proxy and no_proxy variables in the kargo-controller. Currently we work around this by patching the deployment with kustomize. With this PR it can be done via helm-values.

While I technically only need these values on the kargo-controller, I wanted to add the env and envFrom values to all deployments to keep it consistent.

I tried to stay close to how argocd lets you configure these values in their helm charts. https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd

Used this values file locally to check that the functionality works:
```yaml
global:
  env:
    - name: GLOBAL_ENV
      value: "========="

api:
  env:
    - name: EXTRA_ENV_API
      value: "-----------------"
  envFrom:
    - configMapRef:
        name: EXTRA_CM_API
  oidc:
    enabled: true
    dex:
      enabled: true
      env:
        - name: EXTRA_ENV_DEX
          value: "-----------------"
      envFrom:
        - configMapRef:
            name: EXTRA_CM_DEX
  adminAccount:
    passwordHash: $2a$12$qG6iGJgTpQbZo3dRi3q1a.gYg8Na8PF8f263VcdcQNN.squppsuSG
    tokenSigningKey: iwishtowashmyirishwristwatch

controller:
  env:
    - name: EXTRA_ENV_CONTROLLER
      value: "-----------------"
  envFrom:
    - configMapRef:
        name: EXTRA_CM_CONTROLLER

garbageCollector:
  env:
    - name: EXTRA_ENV_GARBAGE_COLLECTOR
      value: "-----------------"
  envFrom:
    - configMapRef:
        name: EXTRA_ENV_GARBAGE_COLLECTOR

managementController:
  env:
    - name: EXTRA_ENV_MANAGEMENT_CONTROLLER
      value: "-----------------"
  envFrom:
    - configMapRef:
        name: EXTRA_ENV_MANAGEMENT_CONTROLLER

webhooksServer:
  env:
    - name: EXTRA_ENV_WEBHOOK
      value: "-----------------"
  envFrom:
    - configMapRef:
        name: EXTRA_CM_WEBHOOK
```

With this command:
```yaml
helm template kargo . --namespace kargo --create-namespace  -f test.values.yaml > test.yaml
```

Best regards. Thanks for your work.

PR #1877, opened 2024-04-19
feat(chart): Added annotation: to api.service and default api.service.annotations value to values.yaml

ed-boykin:
Added a section to charts/kargo/templates/api/service.yaml to add 0 to many annotations to the api service.
When hosting Kargo in a managed K8s system, in this case AKS, an annotation can be added to create an internal or externally facing managed azure load balancer for ingress.


krancour:
Hi @ed-boykin. Thanks for this!

I understand your need, but want to suggest a more general-purpose approach to this so that arbitrary annotations can be added to any/all Kargo pods, as there may be reasons beyond those you are encountering to do such a thing.

If I may suggest, @MPritsch has taken a very nice approach to doing something similar for adding arbitrary env vars to any/all Kargo pods over in #1874.

If you could emulate that approach, I think this would be something we could get into v0.6.0.

Thanks again!

charts/kargo/values.yaml
```diff
@@ -123,6 +123,8 @@ api:
     type: ClusterIP
     ## @param api.service.nodePort [nullable] Host port the `Service` will be mapped to when `type` is either `NodePort` or `LoadBalancer`. If not specified, Kubernetes chooses.
     # nodePort:
+    ## @param api.service.annotations Annotations to add to the api service.
```
krancour, line 126:
```suggestion
    ## @param api.service.annotations Annotations to add to the API server's service. 
```

krancour: With the change above, this LGTM. You do need to [run codegen](https://kargo.akuity.io/contributor-guide/hacking-on-kargo#executing-code-generation) after any edits to `values.yaml`.

If you prefer, I can run it for you.

Thanks!

ed-boykin: Thanks Kent. If you wouldn't mind running it. Go isn't a strength for me and I'd probably not do it right.

krancour (approved):

krancour:
Thank you @ed-boykin!

PR #1878, opened 2024-04-19
fix(api): allow ui refresh to deep links

zkhvan:
After #1833, installing the chart into a cluster will cause links to nested paths to return 404.

### Reproduction

- Install `0.6.0-unstable-20240419`
- Navigate to Kargo Dashboard
- Create new project (kargo-demo)
- Navigate to new project (https://host/project/kargo-demo)
- Refresh and you'll see:
  `404 page not found`

Since assets are served through the API server, the router needs to redirect the URLs to `index.html` since the UI is a single page application.

@krancour I figured you'd have context on this, so tagging you. I'm not sure if this is something you've encountered and were thinking of an alternative solution.

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -199,7 +199,7 @@ func (s *server) Serve(ctx context.Context, l net.Listener) error {
 	mux.Handle(grpchealth.NewHandler(NewHealthChecker(), opts))
 	path, svcHandler := svcv1alpha1connect.NewKargoServiceHandler(s, opts)
 	mux.Handle(path, svcHandler)
-	dashboardHandler, err := s.newDashboardRequestHandler()
+	dashboardHandler, err := newDashboardRequestHandler()
 	if err != nil {
 		return fmt.Errorf("error initializing dashboard handler: %w", err)
 	}
@@ -260,7 +260,9 @@ func (s *server) Serve(ctx context.Context, l net.Listener) error {
 	}
 }
 
-func (s *server) newDashboardRequestHandler() (http.HandlerFunc, error) {
+func newDashboardRequestHandler() (http.HandlerFunc, error) {
+	const indexHTML = "index.html"
+
 	uiFS := fs.FS(ui)
 	uiFS, err := fs.Sub(uiFS, "ui")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


krancour:
Nice find. I'll review shortly!

krancour:
A note about how this escaped my attention...

When we run Kargo from source with `tilt up`, the UI _isn't_ served from the API server like it normally would be, because front-end development is faster if the UI is served from its own container running vite, which can hot-reload UI changes...

The embedded-ness of the UI was tested by running the `kargo server` command. I could not progress past the login page because the local server starts with support for client-authn only and the UI doesn't support that yet. (#1728 is still WIP.)

So... being unable to progress past the login page, I never actually did see anything deeper than the login page served from the embedded FS. Especially since assets loaded, it didn't occur to me that anything was amiss.

Installing the unstable chart does indeed permit me to easily reproduce the problem.

internal/api/server.go
```diff
@@ -255,3 +259,52 @@ func (s *server) Serve(ctx context.Context, l net.Listener) error {
...
+		// enumerating files in the directory)
+		if info.IsDir() {
+			httputil.SetNoCacheHeaders(w)
+			http.ServeFileFS(w, req, uiFS, "index.html")
+			return
+		}
+
+		handler.ServeHTTP(w, req)
+	}, nil
+}
```
krancour, line 310:
```suggestion
func newDashboardRequestHandler() (http.HandlerFunc, error) {
	const indexHTML = "index.html"

	uiFS := fs.FS(ui)
	uiFS, err := fs.Sub(uiFS, "ui")
	if err != nil {
		return nil, fmt.Errorf("error initializing UI file system: %w", err)
	}

	handler := http.FileServer(http.FS(uiFS))
	return func(w http.ResponseWriter, req *http.Request) {
		if req.Method != http.MethodGet {
			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
			return
		}

		path := filepath.Clean(req.URL.Path)
		if path == "/" {
			http.ServeFileFS(w, req, uiFS, indexHTML)
			return
		}

		f, err := uiFS.Open(strings.TrimPrefix(path, "/"))
		if f != nil {
			defer f.Close()
		}
		if os.IsNotExist(err) {
			// When the path doesn't match an embedded file, serve index.html
			httputil.SetNoCacheHeaders(w)
			http.ServeFileFS(w, req, uiFS, indexHTML)
			return
		}
		if err != nil {
			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
			return
		}

		// If we get to here, the path exists in the embedded file system

		info, err := f.Stat()
		if err != nil {
			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
			return
		}
		if info.IsDir() {
			// Serve index.html to prevent enumerating files in the directory
			httputil.SetNoCacheHeaders(w)
			http.ServeFileFS(w, req, uiFS, indexHTML)
			return
		}

		// Path is a file
		handler.ServeHTTP(w, req)
	}, nil
}
```

internal/api/server.go
```diff
@@ -195,11 +199,11 @@ func (s *server) Serve(ctx context.Context, l net.Listener) error {
 	mux.Handle(grpchealth.NewHandler(NewHealthChecker(), opts))
 	path, svcHandler := svcv1alpha1connect.NewKargoServiceHandler(s, opts)
 	mux.Handle(path, svcHandler)
-	uiFS := fs.FS(ui)
-	if uiFS, err = fs.Sub(uiFS, "ui"); err != nil {
-		return fmt.Errorf("error initializing UI file system: %w", err)
+	dashboardHandler, err := s.newDashboardRequestHandler()
```
krancour, line 202:
```suggestion
	dashboardHandler, err := newDashboardRequestHandler()
```

internal/api/server.go
zkhvan: Ah, forgot to `Close()` the file 🤦 

In this section of the suggestion, is there a particular reason to not set the cache headers?

```go
if path == "/" {
	http.ServeFileFS(w, req, uiFS, indexHTML)
	return
}
```

krancour:
@zkhvan I proposed minor tweaks for readability and if they're cool with you then this PR lgtm.

internal/api/server.go
zkhvan: @krancour I took the suggestion, with the slight change of adding the `httputil.NoCacheHeaders(w)` to that one line.

krancour: 🤦‍♂️ Thank you!

krancour (approved):

PR #1892, opened 2024-04-23
chore(deps): remove Masterminds/semver@v1.5.0

zkhvan:
Fixes inconsistency in using two versions of the `Masterminds/semver` package. The API for v3 is identical, but with bug fixes and improvements to the semver mechanisms.


krancour:
Thank you! We _meant_ to be using v3 all along. Nice catch.

krancour (approved):

krancour (approved):

PR #1895, opened 2024-04-25
chore: minor test cleanup

krancour:
After a long session of more difficult coding, I wanted to zen out over something mindless...

This has been on my radar for a couple weeks. We haven't been taking advantage of `require.ErrorContains()` until now. It combines a check for a non-nil error with a check that the error message contains a particular fragment of text. This saves us a few lines of code here and there and makes the assertions just a tiny bit more expressive.


hiddeco (approved):
Had observed this as well, but the mindless task wasn't luring me as much 💯

PR #1898, opened 2024-04-26
feat(ui): Edit AnalysisTemplate

### LINKED ISSUE #1785: bug: lots of empty strings and collections in yaml editors

rpelczar:
Fixes #1785 

In this PR I also introduced an option to hide Managed Fields inside the Yaml Editor.

image

image


ui/src/features/stage/edit-stage-modal.tsx
```diff
@@ -29,24 +28,31 @@ const formSchema = z.object({
...
 export const EditStageModal = ({ visible, hide, projectName, stageName }: Props) => {
-  const { data } = useQuery(listStages, { project: projectName });
-  const stage = data?.stages.find((item) => item.metadata?.name === stageName);
+  const { data, isLoading } = useQuery(getStage, {
+    project: projectName,
+    name: stageName,
+    format: RawFormat.YAML
+  });
+
+  const manifest = new TextDecoder().decode(
```
rbreeze, line 37: This is the second time this logic is used - maybe worth creating a utility function.

rbreeze (approved):
LGTM, just one small nitpick

Issue #1899, opened 2024-04-26
AnalysisRun RawFormat still shows omitempty fields


jessesuen:
# Description

@rpelczar was showing me [the new UI](https://github.com/akuity/kargo/pull/1898) that uses the new RawFormat fields that was recently introduced. However, I noticed that the YAML still had fields that should have been excluded with the omitempty.

Here is the object from kubectl:

```
rafalpelczar@Rafas-MacBook-Pro ~ % kubectl get analysistemplates -n kargo-advanced -o yaml cat-fact 
apiVersion: argoproj.io/v1alpha1
kind: AnalysisTemplate
metadata:
  creationTimestamp: "2024-04-05T23:13:55Z"
  generation: 6
  name: cat-fact
  namespace: kargo-advanced
  resourceVersion: "1082610"
  uid: 371f714e-1cf4-4a7f-b3ee-9ef2ecb26600
spec:
  metrics:
  - count: 3
    failureLimit: 3
    interval: 10s
    name: cat-fact
    provider:
      web:
        jsonPath: '{$.status.verified}'
        url: https://cat-fact.herokuapp.com/facts/58e00b5f0aac31001185ed24
    successCondition: result == true
```

But here is the object as returned from API. Notice that `spec.metrics[].provider.web.authentication` is present here, but not in kubectl:

```
apiVersion: argoproj.io/v1alpha1
kind: AnalysisTemplate
metadata:
  creationTimestamp: "2024-04-05T23:13:55Z"
  generation: 6
  managedFields:
  - apiVersion: argoproj.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        .: {}
        f:metrics: {}
    manager: kargo
    operation: Update
    time: "2024-04-26T21:08:57Z"
  name: cat-fact
  namespace: kargo-advanced
  resourceVersion: "1082610"
  uid: 371f714e-1cf4-4a7f-b3ee-9ef2ecb26600
spec:
  metrics:
  - count: 3
    failureLimit: 3
    interval: 10s
    name: cat-fact
    provider:
      web:
        authentication:
          oauth2: {}
          sigv4: {}
        jsonPath: '{$.status.verified}'
        url: https://cat-fact.herokuapp.com/facts/58e00b5f0aac31001185ed24
    successCondition: result == true
```

# Screenshots

![image](https://github.com/akuity/kargo/assets/12677113/4226d690-a2e8-4cd7-851b-3e9c909e86d6)

[...]

PR #1903, opened 2024-04-29
fix(api): use Unstructured to return raw response

hiddeco:
Fixes: #1899

This pull request ensures that the raw object data returned by the Kargo API matches the data from the Kubernetes API server. Which, unlike the JSON and YAML serializers, purges empty structures.

### DIFF
--- a/internal/api/get_analysisrun_v1alpha1.go
+++ b/internal/api/get_analysisrun_v1alpha1.go
@@ -6,8 +6,12 @@ import (
 	"fmt"
 
 	"connectrpc.com/connect"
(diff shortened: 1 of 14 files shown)


devholic (approved):
LGTM 👍

jessesuen (approved):
Strange that this technique is necessary. Do you know why Kubernetes API doesn't suffer the same problem given that the `authentication` field is not a pointer?

hiddeco:
@jessesuen I searched hard and looked at the Kubernetes SIG YAML library and the serializers in API machinery but could not figure out what causes the API to drop the empty objects. If I happen to run into the answer, I will revise what I introduced here.

Issue #1731, opened 2024-04-02
add project-level user & permission management


krancour:
Add to:

* API
* UI
* CLI

We need the ability to define Roles and RoleBindings and mappings of OIDC users to ServiceAccounts...

In thinking about how to make this a nice UX that is not complex, I think an abstraction could help, but I'd like some buy-in from @jessesuen on this...

Mapping OIDC users to SAs _is already a binding of sorts_. If a "Kargo role" were an abstraction over a 1:1:1 of SA:RoleBinding:Role it would only require users to define membership (based on OIDC claims) and permissions. Such an object would bind very cleanly to a UI form and eliminate the need to juggle many separate elements.

It would also make for a pleasant UX in the CLI. Examples:

* `kargo create role foo`
* `kargo grant --email kent@example.io --role foo`
* `kargo revoke --email kent@example.io --role foo`
* `kargo grant --group 12345 --role foo`
* `kargo revoke --group 12345 --role foo`
* `kargo grant --verb read --type stage --role foo`
* `kargo revoke --verb read --type stage --role foo`
* `kargo grant --verb promote --type stage --resource prod`
* `kargo revoke --verb promote --type stage --resource prod`

_All of this would reconcile as pure RBAC resources._

jessesuen:
Is the proposal to introduce a CRD that is the abstraction over the 1:1:1 of SA:RoleBinding:Role? Or are you suggesting the API server provides the abstraction?

Given the two options, I hesitate to introduce a CRD to wrap existing objects. Instead, I propose presenting a unified concept at our API layer that makes RBAC easier to reason about and manage.

krancour:
I had been thinking CRD, but the idea of doing it in the API server is another approach we could consider.

For me, the pro of doing it with the CRD was that even if not using the API server and doing everything with `kubectl` or with GitOps, the abstraction would still be available.

However we approach it, I'd like to be able to do _something_ like this (doesn't need to be _exactly_ this):

```
apiVersion: kargo.akuity.io/v1alpha1
kind: Role
metadata:
  name: prod-promoter
  namespace: kargo-demo
users:
  groups:
  - some-oidc-group-claim
  - some-other-oidc-group-claim
rules:
- apiGroups:
  - kargo.akuity.io/v1alpha1
  resources:
  - stages
  resourceNames:
  - prod
  verbs:
  - promote
```

If I'm slinging YAML, having the above reconcile as the following is a significant convenience:

```
apiVersion: v1
[...]

PR #1906, opened 2024-04-29
feat!: add "Kargo Role" management endpoints and cli commands

krancour:
These are the back end and CLI bits of #1731

__Don't freak out -- Most of this PR is generated code.__

The PR is implemented as a series of narrowly-scoped commits:

1. Update `service.proto` with definitions for new endpoints
2. Run codegen
3. Implement new endpoints -- the endpoints themselves are actually very thin, with all the real work done by a well-tested "Kargo Role database" in the new `github.com/akuity/kargo/internal/api/rbac` package
4. Implement new CLI commands -- note some small feats of slight-of-hand in order to adapt non-k8s-resource "Kargo Roles" for use with cli-runtime printers
5. Miscellaneous related changes

    This bit is breaking because it renames the "well-known" `kargo-secret-manager` ClusterRole as `kargo-project-admin`. Until now, this has been assigned to the API server in each Project namespace when a Project is created in order to dynamically grant the API server permission to manage Secrets in each new project, _without_ granting the same permission at a cluster scope. With this PR, the same now needs to be done for management of ServiceAccounts, Roles, and RoleBindings in each Project namespace. Since the permissions in question are no longer only Secret-related, the name has changed to reflect that.

Some general notes on strategy:

* "Kargo Roles" are basically ServiceAccounts in a Project namespaces + any Roles they are bound to.
* Users can, of course, continue to GitOps ServiceAccounts, Roles, and RoleBindings directly.
* "Kargo Roles" are read-only _unless the following criteria are met_:
    * ServiceAccount must have only one RoleBinding, binding it to only one Role. Anything more complex than this, and Kargo cannot reasonably manage it.
    * All components of the ServiceAccount/Role/RoleBinding trio must be explicitly annotated to mark them as Kargo-managed. This prevents Kargo from attempting to manage ServiceAccounts/Roles/RoleBindings that a user only intended to GitOps.
    * When Kargo creates or updates Kargo-managed ServiceAccount/Role/RoleBinding trios, it will normalize all PolicyRules. This is essentially required to make operations like "grant" and "revoke" tenable. This normalization is another reason Kargo will refuse to attempt to manage any ServiceAccount/Role/RoleBinding that isn't explicitly annotated.

Follow-ups:

* I will create one more endpoint that permits the UI to query for whether a given Kargo Role is "manageable." i.e. Whether it meets the criteria described above. This will be useful for @rbreeze so that read-only Kargo Roles can be displayed in a non-editable manner, while those that Kargo _can_ manage can be bound to a form and edited by users.

  __Edit:__ @jessesuen and I had motivations for making this a separate endpoint, but I believe the way that the rest of this shaped up, those motivations may no longer be relevant. It should be pretty easy to have every "Kargo Role" returned from the API server denote whether it's manageable or not.

__Edit 2:__ Getting or listing "Kargo Roles" will now reflect whether one is Kargo-managed or not.

* Because of the breaking change (`kargo-secret-manager` --> `kargo-project-admin`), I will create an upgrade reconciler in the management controller (similar to what we did for handling v0.4.0 --> v0.5.0 breaking changes) and it will ensure the API server's ServiceAccount is bound to `kargo-project-admin` in every Project namespace.


internal/api/rbac/roles.go
```diff
@@ -0,0 +1,842 @@
...
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "Role",
+			Name:     name,
+		},
+	}
+}
+
+func dedupeStringSlice(s []string) []string {
```
hiddeco, line 756: Think this can be replaced by `slices.Compact` since Go 1.22: https://pkg.go.dev/slices#Compact

internal/api/rbac/roles.go
```diff
@@ -0,0 +1,842 @@
...
+	sa corev1.ServiceAccount,
+	roles []rbacv1.Role,
+	rbs []rbacv1.RoleBinding,
+) (*rbacv1.Role, *rbacv1.RoleBinding, error) {
+	gr := schema.GroupResource{
+		Group:    sa.GetObjectKind().GroupVersionKind().Group,
+		Resource: strings.ToLower(sa.GetObjectKind().GroupVersionKind().Kind),
+	}
+	if !isKargoManaged(&sa) {
+		return nil, nil, kubeerr.NewConflict(
```
hiddeco, line 799: Am wondering if this should be `kubeerr.NewForbidden`? As at present, this results in a slightly confusing error message in certain scenarios:

```console
$ kargo delete role default -p kargo-advanced
Error: already_exists: Operation cannot be fulfilled on serviceaccount "default": ServiceAccount is not Kargo-managed
```

internal/api/rbac/roles.go
krancour: Ooh. Looks like that will look as long as the list is sorted first. And I am sorting lists before de-duping them, so that should work.

internal/api/rbac/roles.go
krancour: Which will translate as a 403, right? I worry people will mis-interpret that as an authz failure.

hiddeco: `NewInvalid` then maybe, which yields an [HTTP 422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422).

krancour: I need to double-check, but iirc, connect lacks some status codes we take for granted and invalid is one of them. That said, this may speak to an opportunity to improve the error handling filter. I will look more carefully when I get to my desk.

krancour: I refreshed my memory on this.

Our error-handling filter maps k8s StatusErrors with code 409 (conflict) to `connect.CodeAlreadyExists` because it's the closest match for a 409 that exists in connect.

That seems to me to be close to correct, because 409 conflict _is_ how I would typically handle something already existing. In fact, kuberr.NewAlreadyExists wraps a 409.

So the long and short of it is while I _think_ the circumstances here do meet the definition of a conflict, we also know the technical constraints we're dealing with will obscure that error and make it look like something it is not.

I agree with your suggestion that 422 (invalid) would be a good alternative, but `kubeerr.NewInvalid()` looks fairly onerous, requiring you to pinpoint exactly which _fields_ have issues, and I'm not sure the circumstances here permit us to do that.

I think another acceptable option could be 400 (bad request), which is also arguably true, when a user has requested an update or delete of something Kargo cannot manage. (Especially given that Kargo Roles now have a field indicating whether they are Kargo-managed or not, a client has the ability to know ahead of time that they ought not make a request to mutate such a Kargo Role.) I think this is the direction I am inclined to take... unless you talk me out of it. 😄

krancour: We talked offline and agreed on 400.

hiddeco (approved):
Test drove this extensively, with happy and not-so-happy paths, and could not find any major issues or faulty behavior besides the two minor nits that have now been addressed.

Excellent job @krancour 🥇

Issue #1869, opened 2024-04-18
new feature docs


krancour:
This is an issue to keep track of new features that weren't documented as they were added so that we don't forget to circle back on them. I am happy to run point.

- [ ] #1843 
- [x] #1778
       #1873
- [x] #1731 

Please, anyone add to this list if/as you become aware of more.

hiddeco:
I have been wondering if we should start documenting some annotations as well, to allow people to drive certain mechanics (e.g. refreshes and reverifications) using `kubectl`.

This would both function as a way to explain _how_ things work, in addition to providing documentation for the lowest level API available.

krancour:
I think that's a great idea, but would want to track it separately from this.

krancour:
Closing this. Two of three issues received formal doc updates. The third (#1843) only involves new chart options, which are well-documeted inline in the chart's `values.yaml` file and in the chart's generated `README.md`.

PR #1913, opened 2024-04-30
docs: document role management cli commands

krancour:
Part of #1869


christianh814 (approved):

krancour:
Holding off on merge, because formatting of some of the sample output will be affected by #1917

PR #1918, opened 2024-05-01
feat(controller): Apply Warehouse Git subscription path filters to all CommitSelectionStrategy types

### LINKED ISSUE #1896: Apply Warehouse Git subscription path filters to all CommitSelectionStrategy types
This was raised by @krancour: "I only just noticed this now and didn't notice in [#1477](https://github.com/akuity/kargo/pull/1477) -- does this path filtering only work with the NewestFromBranch commit selection strategy? That appears to be the case and I'm now wondering why that is. (We can open a separate issue for this if we determine this was a mistake.)"

I started implementation of path filters aimed at only the NewestFromBranch CommitSelectionStrategy. Now after rethinking I believe it should not be restricted to the mentioned CommitSelectionStrategy but instead apply to all of the strategies. No matter if tag filters are used in a Git subscription - that should not prevent the option of additionally using path filters. In fact having both filtering types combined makes it more flexible in my opinion.

@krancour @hiddeco what do you think? If approved I'll take it.

maksimstankevic:
Fixes #1896

### DIFF
--- a/internal/controller/warehouses/git.go
+++ b/internal/controller/warehouses/git.go
@@ -169,23 +169,88 @@ func (r *reconciler) selectTagAndCommitID(
 	baseCommit string,
 ) (string, string, error) {
 
-	continueWithoutTag := false
-	var filteredTags []string
-	var err error
-	// get filtered list of repo tags only if strategy differs from NewestFromBranch
-	if sub.CommitSelectionStrategy != kargoapi.CommitSelectionStrategyNewestFromBranch {
-		filteredTags, err = r.getFilteredTags(repo, sub)
+	if sub.CommitSelectionStrategy == kargoapi.CommitSelectionStrategyNewestFromBranch {
+		// In this case, there is nothing to do except return the commit ID at the
+		// head of the branch unless there are includePaths/excludePaths configured to
+		// handle.
+		commit, err := r.getLastCommitIDFn(repo)
 		if err != nil {
-			return "", "", fmt.Errorf("error getting applicable tags: %w", err)
+			return "", "",
+				fmt.Errorf("error determining commit ID at head of branch %q in git repo %q: %w",
+					sub.Branch,
+					sub.RepoURL,
+					err,
+				)
 		}
+		// In case includePaths/excludePaths filters are configured in a git subscription
+		// below if clause deals with it. There is a special case - Warehouse has not
+		// produced any Freight yet, this is sorted by creating Freight based on last
+		// commit without applying filters.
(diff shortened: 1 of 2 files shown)


internal/controller/warehouses/git.go
```diff
@@ -169,88 +169,23 @@ func (r *reconciler) selectTagAndCommitID(
 	baseCommit string,
 ) (string, string, error) {
 
-	if sub.CommitSelectionStrategy == kargoapi.CommitSelectionStrategyNewestFromBranch {
-		// In this case, there is nothing to do except return the commit ID at the
-		// head of the branch unless there are includePaths/excludePaths configured to
-		// handle.
-		commit, err := r.getLastCommitIDFn(repo)
+	continueWithoutTag := false
```
krancour, line 172: @maksimstankevic I'm finding the revised logic of this function a bit hard to follow. fwiw, it does look logically correct. At a quick glance, I think just some slight re-ordering of the previous logic might be easier to read. Something along these lines:

```golang
func (r *reconciler) selectTagAndCommitID(
	repo git.Repo,
	sub kargoapi.GitSubscription,
	baseCommit string,
) (string, string, error) {

	var selectedTag, selectedCommit string
	if sub.CommitSelectionStrategy == kargoapi.CommitSelectionStrategyNewestFromBranch {
		// Set selectedCommit
	} else {
		// Get tags

		// Filter tags

		// Bail if no tags

		switch sub.CommitSelectionStrategy {
		case kargoapi.CommitSelectionStrategyLexical:
			// Set selectedTag
		case kargoapi.CommitSelectionStrategyNewestTag:
			// Set selectedTag
		case kargoapi.CommitSelectionStrategySemVer:
			// Set selectedTag
		default:
			// Unknown strategy
		}

		// Bail if no tag selected

		// Checkout the selected tag

		// Set selectedCommit
	}

	// When we get to here, we have selected the correct commit and, if applicable,
	// correct tag, regardless of what the selection strategy was

	// Bail if selectedCommit == baseCommit

	// Getting diffs since baseCommit
	
	// Apply path filers

	return selectedTag, selectedCommit, nil
}
```

maksimstankevic: done

krancour:
@maksimstankevic this looks great. Only nit is that this comment might not be relevant anymore since there's no early return for this selection strategy.

https://github.com/akuity/kargo/blob/fbea61f1384cc226922f2b6657f0fa27bd1fbdf2/internal/controller/warehouses/git.go#L175C3-L177C13

krancour (approved):
🚀 Thank you @maksimstankevic !

PR #1919, opened 2024-05-01
refactor(controller): avoid patching Stage from Promotions reconciler

hiddeco:
This pull request changes the way Stages keep track of their current Freight, by removing the patching of the Stage from the Promotion reconciler and replacing it with logic to allow the Stage itself to self-determine what its current Freight is based on the Promotions existing for it.

From a bird's-eye view, it performs the following steps during the reconciliation of a Stage:

1. Retrieve all Promotions associated with the Stage.
2. Arrange them according to specific criteria:
   - Prioritize Running promotions, further sorted by their ULID in ascending order.
   - Next consider other non-terminal promotions, further sorted by their ULID in ascending order.
   - Finally, consider terminal promotions, sorted by their ULID in descending order.
3. Check if the first Promotion in the ordered list is non-terminal. If so, it signifies that the Stage is in a Promoting phase, and the CurrentPromotion must reflect this. If not, CurrentPromotion remains empty, indicating that the Stage is in a Steady phase.
4. Identify any newly terminated Promotions since the last reconciliation by comparing their ULID with the LastPromotion information of the Stage. If created after LastPromotion, they are considered new.
5. Proceed through the list of newly terminated Promotions from oldest to newest. Update the LastPromotion information for the Stage accordingly. If the Promotion was successful, update the Freight history and CurrentFreight for the Stage to reflect this Promotion.

In addition to this:

- The logic around when the Stage is marked as Verifying has changed. This is now set at the point where the reconciler notices the CurrentFreight does not have any VerificationInfo and the current phase is Steady (and _not_ e.g. Promoting).
- The health checks for the Stage will run regardless of any phase of the Stage (and there being e.g. a Promotion in a Running phase).


krancour:
@hiddeco do you think we should remove the now-unused `PromoWentTerminal` predicate?

hiddeco:
It is still being used by the queue for Promotions itself.

krancour:
> It is still being used by the queue for Promotions itself.

Ah. I see it now. 👍

krancour (approved):

Issue #1922, opened 2024-05-01
bug in health check making promotions fail


krancour:
Preparing for the release, testing a variety of e2e cases [from here](https://github.com/krancour/kargo-examples), I've noticed a large number of Promotions failing with an error "unable to determine desired revision," and I've traced it back to a bug in the recently revised health check logic.

https://github.com/akuity/kargo/blob/a6ced295abf9ad46c9b30d6bed7b40fba3136d74/internal/controller/promotion/argocd.go#L223-L227

`desiredRevision` comes from iterating over the Freight to find an artifact that is a match for whatever is in the Argo CD Application's source. If there is no match, `desiredRevision` is the empty string.

As of recently, this condition is treated as an error, but it should not be. The previous behavior was to conclude that if we don't _know_ what the Application should be synced to, we don't factor that into the health check. I believe we need to return to that.

A valid example that can reproduce this issue is [this one](https://github.com/krancour/kargo-examples/blob/main/01-argocd-driven/02-helm-driven/02-image-only/02-with-git-repo/kargo.yaml) wherein the Warehouse subscribes to an image repository and the Application's source points to a GitOps repo to which there is no subscription. The result is Kargo has no notion of what the Application should be synced to and that is incorrectly interpreted as an error.

PR #1923, opened 2024-05-01
fix(promotions): only check revision match sync result if found

hiddeco:
Fixes: #1922


krancour (approved):

krancour:
@maksimstankevic this looks great. Only nit is that this comment might not be relevant anymore since there's a lot more that goes on after this conditional now.

https://github.com/akuity/kargo/blob/fbea61f1384cc226922f2b6657f0fa27bd1fbdf2/internal/controller/warehouses/git.go#L175C3-L177C13

PR #1933, opened 2024-05-02
docs: document minor and major release procedures

krancour:
To account for the possibility that I will not always be the one to do it, this PR adds formal documentation for the procedures to orchestrate a minor or major release.


hiddeco (approved):
This LGTM, and appears to match the notes you shared with me offline.

Issue #1935, opened 2024-05-02
Allow automatic backporting of pull requests


hiddeco:
At present, we backport pull requests to release branches by hand. As a relatively small team, this is quite time-consuming work that could be better spent doing other things.

In the past, I have automated this using a GitHub workflow which is triggered when a label is added to a pull request using a GitHub Action like [`korthout/backport-action`](https://github.com/korthout/backport-action). Besides saving people time, the additional advantage of this is that it lowers the bar to consider something to be included in a patch release.

PR #1936, opened 2024-05-02
feat(devx): allow automated backporting of PRs

hiddeco:
Fixes: #1935

If merged, pull requests can be backported automagically if labeled with `backport/<release-branch>`.

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -1,0 +1,45 @@
+name: Backport
+
+on:
+  pull_request_target:
+    types: [closed, labeled]
+
+permissions:
+  contents: read
+
+jobs:
+  pull-request:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    if: github.repository_owner == 'akuity' && github.event.pull_request.merged && (github.event_name != 'labeled' || startsWith('backport/', github.event.label.name))
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Create backport PRs
+        uses: korthout/backport-action@v2
+        # xref: https://github.com/korthout/backport-action#inputs
+        with:
+          # Use token to allow workflows to be triggered for the created PR
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          # Match labels with pattern `backport/<target-branch>`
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Create backport PRs
+        uses: korthout/backport-action@v2
+        # xref: https://github.com/korthout/backport-action#inputs
+        with:
+          # Use token to allow workflows to be triggered for the created PR
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
hiddeco, line 27: Not sure if this GitHub secret is available to this repository, but I noticed it was used in some others within this GitHub organization.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Create backport PRs
+        uses: korthout/backport-action@v2
+        # xref: https://github.com/korthout/backport-action#inputs
+        with:
+          # Use token to allow workflows to be triggered for the created PR
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          # Match labels with pattern `backport/<target-branch>`
+          label_pattern: '^backport\/([^ ]+)$'
```
devholic, line 27: Shouldn't it be
```suggestion
          label_pattern: '^backport\/(.+)$'
```
? 👀

hiddeco: That would also match spaces, which are not allowed as branch names.

devholic (approved):

.github/workflows/backport.yaml
krancour, line 27: Q: Where does `$target_branch` get set? I see a capture group in this regex, but I do not see where or how `$target_branch` takes on that value.

hiddeco: The capture group is automatically taken as the target branch, see: https://github.com/korthout/backport-action?tab=readme-ov-file#label_pattern

krancour: In that case, this LGTM.

krancour (approved):

PR #1937, opened 2024-05-02
fix(api): fix bug in kargo role update logic

krancour:
Fixes a bug @rbreeze found just now where wholesale update of a "Kargo Role" could add to claim annotations on the underlying ServiceAccount, but couldn't remove values from them.

I also updated a test case to surface the issue and prove it's fixed.


rbreeze (approved):
🙏🏻

PR #1939, opened 2024-05-02
feat(ui): Permissions Management

### LINKED ISSUE #1920: ui: project-level user / role management
Using the abstraction introduced in #1906

rbreeze:
Fixes #1920 

CleanShot 2024-05-02 at 15 06 36@2x
CleanShot 2024-05-02 at 15 08 43@2x
CleanShot 2024-05-02 at 15 07 04@2x
CleanShot 2024-05-02 at 15 03 52@2x
CleanShot 2024-05-02 at 15 04 05@2x
CleanShot 2024-05-02 at 15 06 48@2x
CleanShot 2024-05-02 at 15 10 11@2x


jessesuen:
I didn't think about how an RBAC UI might might look like until now, but now that I've seen it. I have suggestions on ways to simplify the UX. 

Here are things I'd like to simplify:

1. This UI supports arbitrary K8s resources with text fields for apiVersion/group/kind. I'm proposing to limit this to Kargo-specific GVKs so that users don't have to type groups, apiVersions, resources, etc... In other words, when adding a rule, they would only be able to select from: Stages, Promotions, Warehouses, Secrets, AnalysisTemplates, and we would format this with the proper GVK in the rule before submitting to API.

2. Similar to 1, verbs are well known: `create, update, delete, patch, list, promote` (promote only applies to stages). Instead of a text field, a checkbox of predefined verbs would be helpful.

3. The emails/sub/groups can be under a subtitle called "OIDC bindings". Instead of `subs`, let's call it `Subjects`.

4. Though a form is nice, I still think a YAML editor would be desired. 


Let me know if any of this is feasible in short timeframe of v0.6. Otherwise, I'm willing for this to be done as a follow-up in v0.7

rbreeze:
Updated interface:

CleanShot 2024-05-02 at 21 59 11@2x

ui/src/features/project/roles/rule-editor.tsx
```diff
@@ -0,0 +1,103 @@
...
+                options={availableResources.map((r) => ({ value: r, label: r }))}
+              />
+            )}
+          </FieldContainer>
+          <FieldContainer control={control} name='resourceNames' className='w-full'>
+            {({ field }) => (
+              <MultiStringEditor
+                value={field.value}
+                onChange={field.onChange}
+                placeholder='my-pod'
```
jessesuen, line 88:
```suggestion
                placeholder='my-stage'
```

ui/src/features/project/roles/rule-editor.tsx
```diff
@@ -0,0 +1,119 @@
...
+  'serviceaccounts',
+  // rbac.authorization.k8s.io
+  'rolebindings',
+  'roles',
+  // kargo.akuity.io
+  'freights',
+  'stages',
+  'promotions',
+  'warehouses',
+  'secret',
```
jessesuen, line 25: I think secret should be removed here because its in core.

ui/src/features/project/roles/rule-editor.tsx
```diff
@@ -0,0 +1,119 @@
...
+  // argoproj.io
+  'analysisruns',
+  'analysistemplates'
+];
+const availableVerbs = ['create', 'update', 'delete', 'patch', 'list', 'promote'];
```
jessesuen, line 30: This seems to be missing `get`

krancour: Can we also have `*` as a verb? You do not need to expand it on the front end. The back end already does.

Also... idk if this is too much to ask, but the promote verb is only applicable to Stages. Is it possible to only have that option if Stage is selected as the resource type?

krancour (approved):
🔥

PR #1942, opened 2024-05-03
feat(controller):create viewer role in project namespace automatically on project creation

krancour:
Draft


internal/controller/management/projects/projects.go
```diff
@@ -341,6 +357,64 @@ func (r *reconciler) ensureProjectAdminPermissions(
...
+			{
+				APIGroups: []string{kargoapi.GroupVersion.Group},
+				Resources: []string{"freights", "promotions", "stages", "warehouses"},
+				Verbs:     []string{"get", "list", "watch"},
+			},
+			{
+				APIGroups: []string{rolloutsapi.GroupVersion.Group},
+				Resources: []string{"analysisruns", "analysistemplates"},
+				Verbs:     []string{"get", "list", "watch"},
+			},
```
jessesuen, line 396: How about secrets? I think the secrets tab may not function.

krancour: Oh... so I was thinking reading secrets is probably _beyond_ what an average viewer should be able to do.

jessesuen: Yes, let's not add this to this PR. I'm thinking we can do something special in Kargo API about returning redacted / non-sensitive secrets to anyone with the viewer role in the project.

krancour:
As discussed offline, I will amend this PR to also create a project level admin role.

internal/controller/management/projects/projects.go
```diff
@@ -341,6 +364,194 @@ func (r *reconciler) ensureProjectAdminPermissions(
...
+	roles := []*rbacv1.Role{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "admin",
+				Namespace: project.Name,
+				Annotations: map[string]string{
+					rbacapi.AnnotationKeyManaged: rbacapi.AnnotationValueTrue,
+				},
+			},
+			Rules: []rbacv1.PolicyRule{
```
hiddeco, line 417: I slightly worry about the maintainability of this, as it's a separate place from where the other permissions are kept up-to-date, which already has a history of being forgotten at times.

krancour: > as it's a separate place from where the other permissions are kept up-to-date

I think you're referring to the fact that the project webhook also manages some project-level permissions...

That wasn't ever intended to manage _all_ project-level permissions; rather it was intended to manage the very few for which there was a necessity of creating them synchronously. This is what enables you to define a Project at the top of a manifest and subsequently define resources that go in that Project's namespace. The webhook doesn't manage anything that isn't required to pull that off.

krancour: Spoke offline and learned this is _not_ was @hiddeco was referring to. We're working it out offline.

hiddeco: To recap what I meant: there is an overlap between the things defined here and the RBAC-related things that are defined in the chart. As we have a history of forgetting about updating permissions in the chart (#1907, #1798), I worry that this will be _another_ thing that is prone to bugs, as it is likely to go unnoticed during reviews.

However, both @krancour and I do not directly have a better solution for this, and we have agreed to revisit it in the future.

internal/controller/management/projects/projects.go
```diff
@@ -341,6 +364,194 @@ func (r *reconciler) ensureProjectAdminPermissions(
...
+				project.Name,
+				err,
+			)
+		}
+	}
+
+	roles := []*rbacv1.Role{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "admin",
```
hiddeco, line 411:
```suggestion
				Name:      adminRoleName,
```

internal/controller/management/projects/projects.go
```diff
@@ -341,6 +364,194 @@ func (r *reconciler) ensureProjectAdminPermissions(
...
+				{ // Full access to AnalysisTemplates
+					APIGroups: []string{rolloutsapi.GroupVersion.Group},
+					Resources: []string{"analysistemplates"},
+					Verbs:     []string{"*"},
+				},
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "viewer",
```
hiddeco, line 468:
```suggestion
				Name:      viewerRoleName,
```

hiddeco (approved):
Besides the two tiny nits, this looks good to me.

internal/controller/management/projects/projects.go
```diff
@@ -341,6 +364,194 @@ func (r *reconciler) ensureProjectAdminPermissions(
...
+	project *kargoapi.Project,
+) error {
+	logger := logging.LoggerFromContext(ctx).WithFields(log.Fields{
+		"project":   project.Name,
+		"name":      project.Name,
+		"namespace": project.Name,
+	})
+
+	const adminRoleName = "admin"
+	const viewerRoleName = "viewer"
```
jessesuen, line 378: Should we consider naming these `kargo-admin` and `kargo-viewer` in case this clashes with someone installing Kargo into a cluster that is purposed for workloads as well?

internal/controller/management/projects/projects.go
krancour: Ah... meant to do this and missed it. Thanks.

internal/controller/management/projects/projects.go
krancour: But these are in project namespaces. Do we ever expect workloads in project namespaces?

jessesuen: It's not that I expect workloads in kargo namespaces/projects. Oftentimes, there's just some mechanism (like [kyverno](https://kyverno.io/docs/writing-policies/generate/)) which might auto-create stuff in namespace.

Since `admin` and `viewer` are common names, I suggested the `kargo-` prefix to avoid potential for someone installing kargo into a cluster that already has `admin` and/or `viewer` as a role that might get created in all namespaces.

This is not a hill I'd die on.

krancour: Ah... ok... that makes sense, so happy to change this.

krancour, line 378:
```suggestion
	const adminRoleName = "kargo-admin"
	const viewerRoleName = "kargo-viewer"
```

jessesuen (approved):

PR #1947, opened 2024-05-03
refactor: improve ux for managing roles

krancour:
This PR is all about UX.

* Make the "Kargo Roles" DB auto-detect the correct groups for each resource type that can legitimately be used as part of a Kargo project. This spares users from additional details that they really ought not get hung up on.
* Make the "Kargo Roles" DB tolerant of singular resource names having been used instead of plural ones. It will just fix them.
* Re-run codegen since requests to grant and revoke permissions no longer need group to be specified.
* Remove --resource-group flag from CLI's grant and revoke commands.
* Update docs accordingly


api/rbac/v1alpha1/role_types.go
```diff
@@ -27,10 +27,9 @@ type RoleResources struct {
 }
 
 type ResourceDetails struct {
-	ResourceGroup string   `json:"resourceGroup,omitempty"`
```
krancour, line 30: No longer needs to be specified.

internal/api/rbac/policy_rules.go
```diff
@@ -49,27 +52,25 @@ func BuildNormalizedPolicyRulesMap(
 ) (map[string]rbacv1.PolicyRule, error) {
 	rulesMap := make(map[string]rbacv1.PolicyRule)
 	for _, rule := range rules {
-		for _, group := range rule.APIGroups {
-			group = strings.TrimSpace(group)
-			if group == "*" {
-				return nil, fmt.Errorf("wildcard APIGroup is not allowed")
```
krancour, line 55: Now we don't even _look_ at the group since we auto-detect the correct one or fail of we don't recognize the resource type.

internal/api/rbac/policy_rules.go
```diff
@@ -49,27 +52,25 @@ func BuildNormalizedPolicyRulesMap(
...
+		for _, resource := range rule.Resources {
+			var err error
+			resource, err = normalizeResourceTypeName(resource)
+			if err != nil {
+				return nil, err
 			}
-			for _, resource := range rule.Resources {
-				resource = strings.TrimSpace(resource)
-				if resource == "*" {
-					return nil, fmt.Errorf("wildcard Resource is not allowed")
```
krancour, line 60: Now `*` is just an unrecognized resource type.

internal/api/rbac/policy_rules.go
```diff
@@ -150,3 +151,57 @@ func buildRule(
 	}
 	return rule
 }
+
+// nolint: goconst
+func normalizeResourceTypeName(resource string) (string, error) {
```
krancour, line 156: Postel's rule. Map incorrect, singular names to correct plural names. Strip whitespace. Ensure lower case.

internal/api/rbac/policy_rules.go
```diff
@@ -150,3 +151,57 @@ func buildRule(
...
+		return "analysisruns", nil
+	case "analysistemplate", "analysistemplates":
+		return "analysistemplates", nil
+	default:
+		return "", fmt.Errorf("unknown resource type %q", resource)
+	}
+}
+
+// nolint: goconst
+func getGroupName(resourceType string) string {
```
krancour, line 193: Infer group name from resource type.

internal/api/rbac/policy_rules_test.go
```diff
@@ -10,61 +10,22 @@ import (
...
-			rules,
-		)
+		require.ErrorContains(t, err, "unknown resource type")
 	})
 
 	t.Run("multiple resources expand", func(t *testing.T) {
 		rules, err := NormalizePolicyRules([]rbacv1.PolicyRule{
 			{
 				APIGroups: []string{""},
-				Resources: []string{"pods", "services"},
```
krancour, line 67: Can't use pods as a resource type in these tests anymore since it isn't recognized as valid anymore. There's no reason for pod permissions to be added to roles in a Kargo project.

internal/api/rbac/roles.go
```diff
@@ -633,20 +653,20 @@ func ResourcesToRole(
...
-	}
-
-	if isKargoManaged(sa) &&
-		(len(roles) == 0 || (len(roles) == 1 && isKargoManaged(&roles[0]))) &&
-		(len(rbs) == 0 || (len(rbs) == 1 && isKargoManaged(&rbs[0]))) {
-		kargoRole.KargoManaged = true
+	// Since we cannot make any assumptions that they only contain resource types
+	// we recognize, or that they don't do something really unusual like using a
+	// wildcard resource type, never attempt to normalize rules if any of the
+	// underlying resources are not Kargo-managed.
```
krancour, line 664: Non-Kargo-managed rules are read-only, so it's fine to just show unormalized rules. The whole rule normalization process is really to facilitate easy grant/revoke of specific permissions.

internal/api/rbac/policy_rules.go
jessesuen: The one out of this list that gives me pause is: 
> Map incorrect, singular names to correct plural names

Instead of normalizing the plurality, I think I would prefer rejecting the non-plural or unsupported resources (with list of accepted ones). This goes back to being equal to kubernetes rbac rules. With this in place, a user might think they are doing the right thing (because when they apply it via API it works), but the moment they try the same rule with k8s, it would fail.

krancour: Does it make any difference to know that `kargo get role <name> --as-kubernetes-resources` exists?

It's specifically for the eventuality of making the leap from managing roles imperatively to gitopsing them.

krancour: Oh... also... I think they _can't_ get it wrong via UI and via CLI, users do things like this:

```shell
kargo grant --role <name> --resource-type <type> --verb get --verb list
```

This is where I was most expecting that incorrect singular variants of resource type names would occur.

jessesuen: > but the moment they try the same rule with k8s, it would fail.

Actually that would remain true with whitespace, and non-lowercase... So perhaps I am suggesting we be strict about this.

hiddeco (approved):

jessesuen (approved):
As discussed, only feedback was being strict about not accepting non-plural for consistency sake. But other than that LGTM.

internal/api/rbac/policy_rules.go
krancour: We hammered this out offline and the PR has been amended to validate resource types instead of normalizing them.

Issue #1974, opened 2024-05-08
Kargo CLI should infer options from the `~/.config/kargo/config` file


christianh814:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

Kargo should infer config options from the  `~/.config/kargo/config` file.

Example, my login timed out but I still needed to provide the URL.

```
$ kargo get projects
Error: get client from config: error refreshing token: your token is expired; please use `kargo login` to re-authenticate

$ kargo login
Usage:
  kargo login SERVER_ADDRESS (--admin | --kubeconfig | --sso) [flags]
# ...snip

Error: accepts 1 arg(s), received 0

$ kargo login --admin
Usage:
  kargo login SERVER_ADDRESS (--admin | --kubeconfig | --sso) [flags]
# ...snip

Error: accepts 1 arg(s), received 0

$ cat ~/.config/kargo/config
apiAddress: https://0vqsqw4hezevubyw.kargo.akuity.cloud/
bearerToken: eyJhbGciOiJIUz......ST8bdzu4XfkXXrcqX1d5Q
```

# Motivation

User experience could be improved by inferring things that were already set.

# Suggested Implementation

Use the config file to infer configurations, these inferences should be overriden by what's passed via the CLI when applicable.

PR #1975, opened 2024-05-08
feat: use config value if no address passed to login command

devholic:
Fixes https://github.com/akuity/kargo/issues/1974

This commit uses the address in config if no address is passed to the login command.

### DIFF
--- a/internal/cli/cmd/login/login.go
+++ b/internal/cli/cmd/login/login.go
@@ -38,6 +38,7 @@ const defaultRandStringCharSet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRST
 var assets embed.FS
 
 type loginOptions struct {
+	Config        libConfig.CLIConfig
 	InsecureTLS   bool
 	UseAdmin      bool
 	UseKubeconfig bool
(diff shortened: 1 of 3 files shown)


internal/cli/cmd/login/login.go
```diff
@@ -47,12 +48,16 @@ type loginOptions struct {
...
+	cfg libConfig.CLIConfig,
+) *cobra.Command {
+	cmdOpts := &loginOptions{
+		Config: cfg,
+	}
 
 	cmd := &cobra.Command{
-		Use:   "login SERVER_ADDRESS (--admin | --kubeconfig | --sso)",
-		Args:  option.ExactArgs(1),
+		Use:   "login [SERVER_ADDRESS] (--admin | --kubeconfig | --sso)",
```
devholic, line 59: Would it better to show config address like below?

```suggestion
		Use:   "login [SERVER_ADDRESS | <default address here>] (--admin | --kubeconfig | --sso)",
```

hiddeco: The current is fine, as the default address is not _actually_ being used as an argument but as a fallback.

Instead, it may be better to instruct people how to make use of the default address in the `Example` section.

devholic: Added an example for omitted server address: https://github.com/akuity/kargo/pull/1975/commits/c327f72bc0988e058438070c06db921640f197a4

internal/cli/cmd/login/login.go
```diff
@@ -47,17 +48,24 @@ type loginOptions struct {
...
+		Args:  option.MaximumNArgs(1),
 		Short: "Log in to a Kargo API server",
 		Example: templates.Example(`
 # Log in using SSO
 kargo login https://kargo.example.com --sso
 
+# Last logged in address will be used if the server address is not provided
```
hiddeco, line 66:
```suggestion
# Log in (again) using the last used server address
```

maybe? To keep the style/sentence structure the same.

Issue #1946, opened 2024-05-03
Helm subchart promotion does not seem to use credentials


tmsdce:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

Hi everyone, this project looks very promising !

I'm giving a first shot at the helm subchart promotion mechanism and I'm getting 401 errors during the `helm dependency update` phase. It is as if promotion does not use the provided credentials whereas Warehouse's subscription does use it correctly. I saw https://github.com/akuity/kargo/issues/1281 but I don't know if it is the exact same issue as it should have been fixed in https://github.com/akuity/kargo/pull/1450 

# Steps to Reproduce

Not sure if this can be reproduced accurately but here's my setup

- I have a repo organised as follows

```
└── stages
    ├── prod
    │   ├── Chart.yaml
    │   └── values.yaml
    ├── test
    │   ├── Chart.yaml
    │   └── values.yaml
    └── uat
        ├── Chart.yaml
        └── values.yaml
```

- Each `Chart.yaml` file has the following dependency defined. The `app-for-kargo` chart is published on our harbor registry.  Proper credentials are needed so that the chart can be pulled.

```
dependencies:
  - name: app-for-kargo
    version: 0.1.0
    repository: oci://my.harbor.registry/helm
```

- I have defined a `Warehouse` that subscribes to both the git repo (for when a value file is updated for example) and the chart repository (for when a new release of the chart is published). The subscription part **works fine** as a freight is generated showing both the latest commit of the repo and the latest version of the helm chart. Credentials are provided as secrets and correctly used by the warehouse controller. The secret for the Harbor OCI registry uses `oci://my.harbor.registry/helm/app-for-kargo` for the `repoURL` field

```
---
apiVersion: kargo.akuity.io/v1alpha1
kind: Warehouse
metadata:
  name: kargo-demo
  namespace: kargo-demo
spec:
  subscriptions:
  - git:
      repoURL: https://my.git.repo/kargo-demo.git
      includePaths:
      - regex:^stages/.*$
  - chart:
      repoURL: oci://my.harbor.registry/helm/app-for-kargo
      semverConstraint: ^0
```

- I defined a `Stage` named `test` like this

```
---
apiVersion: kargo.akuity.io/v1alpha1
kind: Stage
metadata:
  name: test
[...]

PR #1977, opened 2024-05-08
fix(promotion): prepare credentials for Helm chart dependency update

hiddeco:
Fixes: #1946

### DIFF
--- a/internal/controller/promotion/helm.go
+++ b/internal/controller/promotion/helm.go
@@ -82,7 +82,7 @@ func (h *helmer) apply(
 			filepath.Join(workingDir, file),
 			changes,
 		); err != nil {
-			return nil, fmt.Errorf("error updating values in file %q: %w", file, err)
+			return nil, fmt.Errorf("updating values in file %q: %w", file, err)
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


internal/controller/promotion/helm.go
```diff
@@ -98,6 +102,9 @@ func (h *helmer) apply(
 		if err = h.setStringsInYAMLFileFn(chartYAMLPath, changes); err != nil {
 			return nil, fmt.Errorf("error updating dependencies for chart %q: %w", chart, err)
 		}
+		if err = h.prepareDependencyCredentialsFn(context.TODO(), homeDir, chartYAMLPath, namespace); err != nil {
+			return nil, fmt.Errorf("error resolving credentials for chart %q: :%w", chart, err)
```
devholic, line 106: nit: I think `error` prefix can be duplicated in outer context. How about removing it? (Ref - [Uber go style guide - error wrapping](https://github.com/uber-go/guide/blob/master/style.md#error-wrapping)
```suggestion
			return nil, fmt.Errorf("resolve credentials for chart %q: :%w", chart, err)
```

devholic (approved):
LGTM 👍

internal/controller/promotion/helm.go
hiddeco: Ah, the Uber go style guide is where this is coming from. I wondered about it before, after observing this style of errors in other places in the code base.

I agree it's beneficial here (although in _some_ other places, it could use more context to answer _why_ questions specific to the returned error).

I will make some adjustments, thanks.

internal/controller/promotion/helm.go
```diff
@@ -90,16 +94,19 @@ func (h *helmer) apply(
...
+		return nil, fmt.Errorf("preparing changes to affected Chart.yaml files: %w", err)
 	}
 	for chart, changes := range changesByChart {
 		chartPath := filepath.Join(workingDir, chart)
 		chartYAMLPath := filepath.Join(chartPath, "Chart.yaml")
 		if err = h.setStringsInYAMLFileFn(chartYAMLPath, changes); err != nil {
-			return nil, fmt.Errorf("error updating dependencies for chart %q: %w", chart, err)
+			return nil, fmt.Errorf("setting dependency versions for chart %q: %w", chart, err)
+		}
+		if err = h.prepareDependencyCredentialsFn(context.TODO(), homeDir, chartYAMLPath, namespace); err != nil {
```
krancour, line 105: Other than needing to tweak some function signatures and function calls, is there anything preventing us from getting at the right context here?

krancour (approved):
🔥

internal/controller/promotion/helm.go
hiddeco: Not really, but the size of the constructor is starting to feel outrageous and a mixed bag of not-always-so-relevant things.

PR #1978, opened 2024-05-08
feat: allow updating role description via api

rbreeze:


internal/api/rbac/roles.go
```diff
@@ -132,6 +133,13 @@ func (r *rolesDatabase) Create(
 		)
 	}
 
+	// Append the description annotation to the Role if it exists
+	if annotations := kargoRole.GetAnnotations(); annotations != nil {
+		if description, ok := annotations[kargoapi.AnnotationKeyDescription]; ok {
+			sa.Annotations[kargoapi.AnnotationKeyDescription] = description
+		}
+	}
```
hiddeco, line 141:
```suggestion
	if description, ok := kargoRole.Annotations[kargoapi.AnnotationKeyDescription]; ok {
		sa.Annotations[kargoapi.AnnotationKeyDescription] = description
	}
```

The `ok` check already covers for the annotations potentially being `nil`.

internal/api/rbac/roles.go
```diff
@@ -579,6 +587,13 @@ func (r *rolesDatabase) Update(
 	replaceClaimAnnotation(sa, rbacapi.AnnotationKeyOIDCSubjects, kargoRole.Subs)
 	replaceClaimAnnotation(sa, rbacapi.AnnotationKeyOIDCEmails, kargoRole.Emails)
 	replaceClaimAnnotation(sa, rbacapi.AnnotationKeyOIDCGroups, kargoRole.Groups)
+	if annotations := kargoRole.GetAnnotations(); annotations != nil {
```
hiddeco, line 590: This check is also redundant, `delete` can safely operate on `sa.Annotations` (as long as it's a map, but it's allowed to be a `nil` map).

internal/api/rbac/roles.go
```diff
@@ -632,6 +647,12 @@ func ResourcesToRole(
 		},
 	}
 
+	if annotations := sa.GetAnnotations(); annotations != nil {
```
hiddeco, line 650: Same here.

hiddeco:
Can you please add some dummy descriptions to the success tests to ensure we have coverage for the most basic happy path?

hiddeco (approved):
Looks good to me, thanks @rbreeze! 🍰

Issue #1822, opened 2024-04-11
UI: Show more information in Promotion Table view


rbreeze:
Currently, we don't show much detail in the Promotions Table of the Stage panel. 

We should display:
- URL (to PR or commit)
- More information about Freight, or at least link to more detail (name isn't helpful)
- Information about the Argo CD Application status (if applicable)

![CleanShot 2024-04-11 at 14 24 55@2x](https://github.com/akuity/kargo/assets/6250584/e52b2fc3-aac0-4c6a-af25-8600c25e6719)

rbreeze:
TLDR; #1979 is the best we will get for this without some questionable API decisions.

In #1979, I added links to Freight Details in this table. I also added a Tooltip to the Freight ID column, such that when you hover over it, the Freight alias appears.

Implementation-wise, this is accomplished by doing an ad-hoc API request every time a user hovers. This is necessary because the `ListPromotions` API does not contain freight information besides just the freight name (ID), nor should it IMO (would require additional K8S API calls in the endpoint to accomplish). Likewise, the UI shouldn't call the `GetFreight` API repeatedly on this page.

@krancour @hiddeco do you have thoughts here?

hiddeco:
The only optimization I can think of is temporarily caching the frontend response as the Freight is _almost_ always static (the single exception being: if someone manually changes the alias).

Providing e.g. a special endpoint for a more lightweight response would not yield a better result for the backend, which would still have to retrieve the object to provide a compressed view.

krancour:
> ListPromotions API does not contain freight information besides just the freight name (ID), nor should it IMO

I agree with this. Aliases are _mutable_ and if we included information about aliases in a FreightReference, we'd have to track down and fix those any time a piece of Freight's alias changes. It's simply not worth it.

We could, theoretically, add a more bespoke API endpoint just for use with this view, where you supply a list of Freight IDs and get back the corresponding list of aliases.

I wouldn't object if you wanted to do that.

rbreeze:
@hiddeco @krancour I agree with you both. 

> We could, theoretically, add a more bespoke API endpoint just for use with this view, where you supply a list of Freight IDs and get back the corresponding list of aliases.

TBH, I don't think it's worth it. The tooltip solution isn't as nice as if we had the alias inline, but I think it gets 90% of the way there. I'll go ahead and close this issue but feel free to reopen if you feel we should do more here

PR #1979, opened 2024-05-09
feat(ui): display freight contents in freight details page

rbreeze:
Screenshot 2024-05-08 at 17 28 16

Also adds links to Freight Details in Promotions table, and a tooltip showing the Freight alias (related to #1822):
Screenshot 2024-05-08 at 17 28 05

### DIFF
--- a/ui/src/features/freight/freight-details.tsx
+++ b/ui/src/features/freight/freight-details.tsx
@@ -64,7 +64,7 @@ export const FreightDetails = ({ freight }: { freight?: Freight }) => {
                   children: (
                     <>
                       <div className='mb-4'>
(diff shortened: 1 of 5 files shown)


hiddeco:
This looks good to me, except that I am wondering if we should use the term 'artifacts' instead of 'contents'.

hiddeco:
Double-checking my suggestion, it would also better align with how we describe Freight ourselves in the documentation:

> Freight is an important Kargo concept. A single "piece of freight" is a set of references to one or more versioned artifacts, [...]
> — https://kargo.akuity.io/how-to-guides/working-with-freight

rbreeze:
@hiddeco thanks for the suggestion! changed ✅

PR #1983, opened 2024-05-09
fix(ui): Better display of long labels

rbreeze:
- When freight alias is too long to fit in Freightline box, break on the hyphen and overflow with an ellipsis
- Make Freightline boxes slightly taller+wider
- When freight aliases are too wide to fit in Stage nodes, truncate with ellipsis instead of overflowing
- Make Pipeline nodes slightly wider
- Add human-readable timestamp to Stage nodes
- Add full timestamp to Stage node tooltips

Screenshot 2024-05-09 at 09 17 18
Screenshot 2024-05-09 at 09 17 51


hiddeco (approved):
The only minor nit I have is that the hamburger menu thingy in the top right corner will now not have any padding with the artifact metadata (first item in the freightline).

rbreeze:
@hiddeco Oh this is true in `main` today as well... day zero issue

Screenshot from `main`:
Screenshot 2024-05-09 at 10 25 40

Issue #761, opened 2023-09-14
Manually input freight


jessesuen:
I have predefined images in a container repository. Currently, the only way these images can be detected and entered as Freight by kargo, is if we subscribed to that container repository for a stage before those images were created.

That is all well and good for an app developer deploying their service continuously in a SaaS. But in my use case, the images I care about promoting, were developed by someone else and existed long before I started using Kargo. So more of an operator use case than an app developer.

I would like the ability to manually input freight with images. Even better, is if we had an API/UI feature that presented container images (not yet been created as freight), and allowed me to convert them as freight.

The [Freight CRD](https://github.com/akuity/kargo/issues/631) would probably help here

krancour:
I agree we need this capability, but if you have an immediate need to do something like this, I do think, it can be done today if the third-party image is semvered.

jessesuen:
This use case was brought up by @wanghong230 today. When connecting a git/container repository to a warehouse, we only fetch the most recent freight. He proposed that some last N freight be populated.

Though I think a manual ability would be sufficient to suit this use case.

krancour:
> He proposed that some last N freight be populated/backfilled.

The tricky thing about that is that a piece of Freight is a _combination_ of other artifacts.

Say you fetch the last n imagess and the last m commits, then there are n * m different combinations of those and some number of them are probably invalid. i.e. HEAD~4 might be completely incompatible with the latest image.

> Though I think a manual ability would be sufficient to suit this use case.

I think this is the only reasonable way to reach back in time to pull something arbitrarily old into the pipeline.

krancour:
This plays into #1680 and the new idea of a "freight lab."

Brightside56:
> This plays into https://github.com/akuity/kargo/issues/1680 and the new idea of a "freight lab."
And this helps to resolve my issue with many artifacts for components of single microservice

hiddeco:
We discussed this during an offline meeting today and concluded there that to make this work nicely and help address the "freight lab" idea. It should be possible for a user to have an overview of artifacts that are not part of a Freight but are available to a subscription of a Warehouse.

While it was opted there that we potentially could utilize Redis for this as a caching backend, I also briefly explored the idea of running a Kubernetes API extension server to _not_ have to deal with another side dependency we would have to protect and/or invent authentication mechanisms for.

[...]

PR #1984, opened 2024-05-09
feat(warehouse): artifact discovery and manual Freight creation

hiddeco:
Addresses the majority of the backend part of #761 

- [x] Introduce `FreightCreation` field to `Warehouse` spec to allow controlling how `Freight` is created from the `Warehouse`. Supported modes are `Automatic` (default) and `Manual`.
- [x] Introduce `DiscoveredArtifacts` field to the `Warehouse` status with an overview of newly discovered artifacts for the subscriptions of the `Warehouse`.
- [x] Introduce `[]ImageDiscoveryResult` to the `DiscoveredArtifacts` of the `Warehouse` with a list of the latest 20 tags for each `ImageSubscription` (according to the `ImageSelectionStrategy`) when `FreightCreation` is set to `Manual`.
- [x] Introduce `[]ChartDiscoveryResult` to the `DiscoveredArtifacts` of the `Warehouse` with a list of the latest 20 versions for each `ChartSubscription` (according to the `SemverConstraint`) for each `ChartSubscription` when `FreightCreation` is set to `Manual`.
- [ ] Introduce `[]GitDiscoveryResult` to the `DiscoveredArtifacts` of the `Warehouse` with a list of the latest 20 Git references (according to the `CommitSelectionStrategy`) for each `GitSubscription` when `FreightCreation` is set to `Manual`.
- [ ] Write tests for the newly introduced reconciler logic.
- [ ] Make bits of code more efficient (see in-code `// TODO(hidde)`)
- [ ] Revise if we are better off _always_ discovering the artifacts — and then creating the `Freight` (if automated). This would allow the UI to provide a better overview of the `Warehouse`, even if the `Freight` is automatically created.


api/v1alpha1/warehouse_types.go
```diff
@@ -55,13 +55,30 @@ type WarehouseSpec struct {
 	// empty, the defaulting webhook will set the value of this field to the value
 	// of the shard label.
 	Shard string `json:"shard,omitempty" protobuf:"bytes,2,opt,name=shard"`
+	// FreightCreation describes how Freight is created by this Warehouse.
+	// This field is optional. When left unspecified, the field is implicitly
+	// treated as if its value were "Automatic".
+	//
+	// +kubebuilder:default=Automatic
+	FreightCreation FreightCreation `json:"freightCreation" protobuf:"bytes,3,opt,name=freightCreation"`
```
hiddeco, line 63: I am not too keen on how this field has been named. Happy to take suggestions.

api/v1alpha1/warehouse_types.go
```diff
@@ -267,6 +284,127 @@ type WarehouseStatus struct {
...
+	Tag string `json:"tag,omitempty" protobuf:"bytes,3,opt,name=tag"`
+
+	// Subject is the subject of the commit (i.e. the first line of the commit
+	// message).
+	// +optional
+	Subject string `json:"subject,omitempty" protobuf:"bytes,4,opt,name=subject"`
+
+	// Author is the author of the commit.
+	// +optional
+	Author string `json:"author,omitempty" protobuf:"bytes,5,opt,name=author"`
```
hiddeco, line 344: This is primarily named "author" (the person who wrote the code) because that's what the provisional field in `GitCommit` is called. _However_, I can not help but wonder if for Kargo, we are more interested in the "committer" (the person who published it)?

api/v1alpha1/warehouse_types.go
```diff
@@ -267,6 +284,127 @@ type WarehouseStatus struct {
...
+
+	// CreatedAt is the time at which the commit or annotated tag was created.
+	// +optional
+	CreatedAt *metav1.Time `json:"createdAt,omitempty" protobuf:"bytes,6,opt,name=createdAt"`
```
hiddeco, line 348: In Git concepts, this would be named a `CreatorDate`. Based on my observations, Kargo does however lean towards not doing 1:1 copying of these type of terms, and because of this I opted for the more Kubernetes-like `CreatedAt`.

api/v1alpha1/warehouse_types.go
krancour: Only thing I can think of that might be a very slight improvement is `FreightCreationPolicy`.

api/v1alpha1/warehouse_types.go
```diff
@@ -267,6 +284,127 @@ type WarehouseStatus struct {
...
+}
+
+// GitDiscoveryResult represents the result of a Git discovery operation for a
+// GitSubscription.
+type GitDiscoveryResult struct {
+	// RepoURL is the repository URL of the GitSubscription.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
+	RepoURL string `json:"repoURL" protobuf:"bytes,1,opt,name=repoURL"`
```
krancour, line 310: Question applies in several places. I'm curious what the value of the kubebuilder markers might be, if any, for things like this, which (admittedly early in my review) seem only to be system-generated and not user-generated.

api/v1alpha1/warehouse_types.go
```diff
@@ -267,6 +284,127 @@ type WarehouseStatus struct {
...
+	// GitSubscription. An empty list indicates that the discovery operation was
+	// successful, but no commits matching the GitSubscription criteria were found.
+	// +optional
+	Commits []DiscoveredCommit `json:"commits,omitempy" protobuf:"bytes,2,rep,name=commits"`
+}
+
+// DiscoveredCommit represents a commit discovered by a Warehouse for a
+// GitSubscription.
+type DiscoveredCommit struct {
+	// ID is the identifier of the commit. This typically is a SHA-1 hash.
```
krancour, line 322: "typically?" Is it ever not? (Sincere question.)

api/v1alpha1/warehouse_types.go
hiddeco: 1. If the system is wrong due to a (presumably) human mistake, this is one way to figure this out soon.
2. It documents what can be expected of the value, which has value by itself.

api/v1alpha1/warehouse_types.go
krancour: Good question. I am wondering if it's useful to capture both. Also... cannot commits have multiple authors?

api/v1alpha1/warehouse_types.go
hiddeco: While it has been a multi-year-long process, Git is still paving the way for SHA-256. As a matter of fact, since Git 2.42 it is no longer marked as experimental.

api/v1alpha1/warehouse_types.go
hiddeco: A commit object (as stored in Git) can only have a single author. The functionality you see in e.g. GitHub is based on parsing the `Co-authored-by` field at the bottom of the commit message, which is more a standard than an official feature.

api/v1alpha1/warehouse_types.go
```diff
@@ -267,6 +284,127 @@ type WarehouseStatus struct {
...
+	// +kubebuilder:validation:Pattern=`^[\w.\-\_]+$`
+	Tag string `json:"tag" protobuf:"bytes,1,opt,name=tag"`
+	// Digest is the digest of the image.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]+:[a-f0-9]+$`
+	Digest string `json:"digest" protobuf:"bytes,2,opt,name=digest"`
+	// Platform is the target platform of the image. This field is optional, and
+	// only populated if the ImageSubscription specifies a Platform.
+	// +optional
+	Platform string `json:"platform,omitempty" protobuf:"bytes,3,opt,name=platform"`
```
krancour, line 382: I wonder slightly if this is important information and I also wonder if it could lead to confusion.

In terms of image selection, when a tag points to a multi-arch image's manifest list (v2) or index (OCI) it is the digest of the list/index that is returned (as long as that list or index does indeed reference an image having the desired architecture). It's _not_ the digest of the single-arch image that is returned. This behavior makes sense because 1. That's not what the tag points to and 2. This behavior leaves containerd free to self-select the correct platform out of the multiple ones available per usual.

So the confusion that could arise here is that the digest may reference a multi-arch image while the platform field, with a value copied over from the platform constraint, may promote the idea that this image is for a specific platform only.

api/v1alpha1/warehouse_types.go
```diff
@@ -267,6 +284,127 @@ type WarehouseStatus struct {
...
+	// GitRepoURL is the URL of the Git repository that contains the source
+	// code for this image. This field is optional, and only populated if the
+	// ImageSubscription specifies a GitRepoURL.
+	// +optional
+	GitRepoURL string `json:"gitRepoURL,omitempty" protobuf:"bytes,4,opt,name=gitRepoURL"`
```
krancour, line 387: Any thoughts on this moving up a level to the `ImageDiscoveryResult`? Isn't it going to be the same for _every_ `DiscoveredImage` within a result?

hiddeco: It's tag specific for GitHub URLs, due to:

```go
func getGithubImageSourceURL(gitRepoURL, tag string) string {
	return fmt.Sprintf("%s/tree/%s", git.NormalizeURL(gitRepoURL), tag)
}
```

internal/controller/git/git.go
```diff
@@ -89,8 +118,13 @@ type Repo interface {
 	// GetDiffPaths returns a string slice indicating the paths, relative to the
 	// root of the repository, of any new or modified files.
 	GetDiffPaths() ([]string, error)
```
krancour, line 120: This is not used anywhere. Mind getting rid of it (and its implementation) while you're already touching this file?

internal/controller/git/git.go
```diff
@@ -89,8 +118,13 @@ type Repo interface {
...
-	// GetDiffPathsForCommitID returns a string slice indicating the paths, relative to the
-	// root of the repository, of any files that are new or modified since the given commit ID.
+	// GetDiffPathsForCommitID returns a string slice indicating the paths,
+	// relative to the root of the repository, of any files that are new or
+	// modified in the commit with the given ID.
+	GetDiffPathsForCommitID(commitID string) ([]string, error)
+	// GetDiffPathsSinceCommitID returns a string slice indicating the paths,
+	// relative to the root of the repository, of any files that are new or
+	// modified since the given commit ID.
 	GetDiffPathsSinceCommitID(commitId string) ([]string, error)
```
krancour, line 128: Teeny tiny nit: I read this a few times before I understood the difference between the two functions. Maybe consider emphasizing IN and SINCE? Or don't. Not a hill I would die on.

api/v1alpha1/warehouse_types.go
hiddeco: In hindsight, I think I agree. The reason I added it was because I wanted to allow the UI to potentially information about the architecture constraint. However, this is likely better to add one level higher as it will be the same for every listed result anyway.

internal/controller/git/git.go
```diff
@@ -145,17 +185,44 @@ type ClientOptions struct {
...
+	// FilterBlobless is a filter that excludes blobs from the clone.
+	// When using this filter, the initial Git clone will download all
+	// reachable commits and trees, and only download the blobs for commits
+	// when you do a Git checkout (including the first checkout during the
+	// clone).
+	//
+	// When using a blobless clone, you can still explore the commits of the
+	// repository without downloading additional data. This means that you can
+	// perform commands like `git log`, or even `git log -- <path>` with the
+	// same performance as a full clone.
```
krancour, line 198: Did you mean to say _better_ performance than a full clone?

internal/controller/git/git.go
hiddeco: It's no longer in use, and I can potentially also just remove it...

internal/controller/git/git.go
```diff
@@ -145,17 +185,44 @@ type ClientOptions struct {
...
+	//
+	// Commands like `git diff` or `git blame <path>` require the contents of
+	// the paths to compute diffs, so these will trigger blob downloads the
+	// first time they are run.
+	FilterBlobless = "blob:none"
```
krancour, line 203: 🔥

internal/controller/git/git.go
hiddeco: The operation (`git log`, etc.) does not have better performance. It's performing the same as a full clone. The cloning operation itself _is_ however lighter. (You do less, but get the same benefits.)

internal/controller/git/git.go
```diff
@@ -364,6 +431,23 @@ func (r *repo) GetDiffPaths() ([]string, error) {
 	return paths, nil
 }
 
+func (r *repo) GetDiffPathsForCommitID(commitID string) ([]string, error) {
+	resBytes, err := libExec.Exec(r.buildGitCommand("diff", "--name-only", commitID))
```
krancour, line 435: I am not sure if I am misunderstanding how this func is meant to be used or if there's an actual mistake here, but isn't `git diff --name-only <commit>` exactly equivalent to `git diff --name-only <commit>..HEAD` used by `GetDiffPathsSinceCommitID`?

I presumed the intent behind this function to be returning paths modified _by_ a given commit, but I think this is returning paths modified _since_ that commit. (Same as `GetDiffPathsSinceCommitID`.)

Possible I am misunderstanding the intent.

api/v1alpha1/warehouse_types.go
krancour: Works for me.

api/v1alpha1/warehouse_types.go
krancour: TIL

api/v1alpha1/warehouse_types.go
krancour: Ah... sure... I agree moving it up a level would clear up the confusion.

api/v1alpha1/warehouse_types.go
krancour: I'd forgotten about that.

internal/controller/git/git.go
hiddeco: You are right, and it should be `git diff --name-only <commit>^!`. Great catch 🙇

internal/controller/git/git.go
krancour: Yeah. I understood that to technically be so, but the commentary on the performance of `git log` seemed unnecessary to get the point across that this is a lighter weight way to clone, which left me wondering what exactly was being said here...

Now I think I realize what you were actually trying to do was express that the lighter weight clone doesn't at all _impair_ the performance of `git log`.

I just kinda read this wrong. All good.

internal/helm/helm.go
```diff
@@ -19,57 +19,107 @@ import (
...
-// credentials may be nil for public repositories, but must be non-nil for
-// private repositories.
+// specific chart and the name argument must be empty.
+//
+// The credentials argument may be nil for public repositories, but must be
+// non-nil for private repositories.
+//
+// If no versions of the chart are found in the repository, an error is
+// returned.
 func SelectChartVersion(
```
krancour, line 38: This func is no longer used anywhere.

internal/image/digest_selector.go
```diff
@@ -46,6 +46,10 @@ func (d *digestSelector) Select(ctx context.Context) (*Image, error) {
 
 	ctx = logging.ContextWithLogger(ctx, logger)
 
+	// TODO(hidde): it would be much more efficient to directly attempt
+	// to retrieve the image for the tag, while gracefully handling the
+	// case where it does not exist. This would avoid the need to list
+	// all tags and then iterate over them.
```
krancour, line 52: Nice catch.

internal/image/selector.go
```diff
@@ -38,11 +38,14 @@ const (
 	SelectionStrategySemVer SelectionStrategy = "SemVer"
 )
 
-// Selector is an interface for selecting a single image from a container image
+// Selector is an interface for selecting images from a container image
 // repository.
 type Selector interface {
 	// Select selects a single image from a container image repository.
 	Select(context.Context) (*Image, error)
```
krancour, line 45: I'm wondering if we still need this. It seems it's only ever called from integration tests. Could we discover with a limit of 1 in those tests instead?

krancour: This is probably my biggest finding. If we don't need this and we kill off the implementations of it, we jettison about 150 lines or so.

api/v1alpha1/warehouse_types.go
krancour: I'm fine with either.

internal/image/selector.go
krancour: And if we _do_ kill off `Select()`, should we _then_ consider renaming `Discover()` to `Select()`?

It's a little awkward if the role of a _Selector_ is to _discover_.

We could get this to where a _Selector_ still _selects_ and the only difference from before this PR to after is that is selects more than one.

krancour:
I notice the `!` in the PR title, but I cannot spot what this breaks. I'm missing something.

internal/image/newest_build_selector.go
```diff
@@ -49,6 +52,118 @@ func (n *newestBuildSelector) Select(ctx context.Context) (*Image, error) {
...
+				"digest": image.Digest.String(),
+			}).Trace("discovered image")
+		}
+		logger.Tracef("discovered %d images", limit)
+		return images[:limit], nil
+	}
+
+	// TODO(hidde): this is unnecessarily inefficient, as we're fetching the
+	// image _again_ to check if it matches the platform constraint. We should
+	// refactor this to avoid fetching the image twice.
```
krancour, line 126: Agree we could do better here, but fwiw, the second retrieval of each is by digest and will always be a cache hit.

internal/image/selector.go
hiddeco: I went with the option of killing off the original `Select()` implementation in favor of the `Discover()` one (which is now called `Select()`. As the reason I _originally_ did not make the change to the previous method directly was because I was still figuring out if the original implementation would still be required.

internal/image/newest_build_selector.go
hiddeco: Rephrased the comment a bit to highlight this.

hiddeco:
> I notice the `!` in the PR title, but I cannot spot what this breaks. I'm missing something.

I had to think again about why I thought this was breaking, and initially thought it wasn't, but then recalled the _tiny_ difference.

At present, include and exclude paths for Git subscriptions look at the full diff between `$previousCommit` and HEAD. If this contains changes to included paths, the current HEAD is greenlighted to create Freight of.

With this pull request, this behavior changes, as we look at every individual commit to see if it contains changes that make it eligible to be used as an artifact for Freight. As the current HEAD _may_ not contain an included path, this means that the newest discovered artifact can change compared to the latest Freight we created (before upgrading), causing the Warehouse to create new Freight with the exact commit that contained these changes (instead of HEAD).

krancour:
> causing the Warehouse to create new Freight with the exact commit that contained these changes

Ah. That makes sense. I doubt many people will notice this very slight behavior change and if anyone does, I think they're likely to find the new behavior make more intuitive sense.

Fine with me if you want to put the `!` back in the title so we don't forget about this when reviewing commits to produce v0.7.0 release notes. It _may_ prove to not be worth calling attention to, but at least we won't forget about it.

internal/controller/git/git.go
```diff
@@ -86,25 +121,24 @@ type Repo interface {
...
-	// GetDiffPaths returns a string slice indicating the paths, relative to the
-	// root of the repository, of any new or modified files.
-	GetDiffPaths() ([]string, error)
-	// GetDiffPathsForCommitID returns a string slice indicating the paths, relative to the
-	// root of the repository, of any files that are new or modified since the given commit ID.
-	GetDiffPathsSinceCommitID(commitId string) ([]string, error)
+	// GetDiffPathsForCommitID returns a string slice indicating the paths,
+	// relative to the root of the repository, of any files that are new or
+	// modified in the commit with the given ID.
+	GetDiffPathsForCommitID(commitID string) ([]string, error)
```
krancour, line 127: Nit: Now that we don't need to differentiate between three slightly different functions for finding diff paths, do we want to change this _one_ that we're keeping to `GetDiffPaths()`?

Whatever you decide it find with me.

internal/controller/git/git.go
```diff
@@ -86,25 +121,24 @@ type Repo interface {
...
 	// branch.
 	LastCommitID() (string, error)
-	// ListTags returns a slice of tags in the repository.
-	ListTags() ([]string, error)
+	// ListTagsWithMetadata returns a slice of tags in the repository with metadata
+	// such as commit ID, creator date, and subject.
+	ListTagsWithMetadata() ([]TagMetadata, error)
+	// ListCommitsWithMetadata returns a slice of commits in the current branch
+	// with metadata such as commit ID, commit date, and subject.
+	ListCommitsWithMetadata(limit, skip uint) ([]CommitMetadata, error)
```
krancour, line 138: Nit: Since I don't think we need to differentiate from any functions that list Tags or Commits _without_ metadata, do we want to rename these to just `ListTags()` and `ListCommits()`?

Whatever you decide is fine with me.

internal/controller/git/git.go
```diff
@@ -404,24 +447,121 @@ func (r *repo) LastCommitID() (string, error) {
...
+func (r *repo) ListTagsWithMetadata() ([]TagMetadata, error) {
+	if _, err := libExec.Exec(r.buildGitCommand("fetch", "origin", "--tags")); err != nil {
 		return nil, fmt.Errorf("error fetching tags from repo %q: %w", r.url, err)
 	}
-	tagsBytes, err := libExec.Exec(r.buildGitCommand("tag", "--list", "--sort", "-creatordate"))
+
+	// These formats are quite complex, so we break them down into smaller
+	// pieces for readability.
+	//
+	// They are designed to output the following fields, separated by `|*|`:
```
krancour, line 458: I am curious about the choice of `|*|` as a delimiter. I'm not suggesting it should change. Asking for my own benefit.

krancour (approved):
Left a few new comments this morning, but they are entirely non-blocking, dealer's choice sort of things.

LGTM as is and it stands if you do choose to do anything about the remaining comments.

This PR is absolute 🔥

internal/controller/git/git.go
hiddeco: I wanted something that was _highly unlikely_ to show up in some of the fields, I could think of reasons e.g. `,` or `|` would show up. But I would be quite baffled about this happening for `|*|`.

internal/controller/git/git.go
hiddeco: Agree on this one, will ammend.

internal/controller/git/git.go
hiddeco: (The reason it needs to be a (set of) character(s) is that unlike `git log` formatting, it `git for-each-ref` can't insert tabs or other whitespace-like escape sequences.)

PR #2005, opened 2024-05-13
fix: use akuitybot PAT for backport worfklow

hiddeco:
This should ensure workflows are properly triggered in all scenarios. Follow up on #1936.

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -22,7 +22,7 @@ jobs:
         # xref: https://github.com/korthout/backport-action#inputs
         with:
           # Use token to allow workflows to be triggered for the created PR
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.AKUITYBOT_PAT }}
           # Match labels with pattern `backport/<target-branch>`
           label_pattern: '^backport\/([^ ]+)$'
           # A bit shorter pull-request title than the default


.github/workflows/backport.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
         # xref: https://github.com/korthout/backport-action#inputs
         with:
           # Use token to allow workflows to be triggered for the created PR
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.AKUITYBOT_PAT }}
```
krancour, line 25: Would `permissions.pull-requests: write` possibly work?

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

krancour: Oh... nevermind... I was looking at perms at the top of the file. I see this job already has the one I mentioned.

krancour: Oh... this needs to be enabled at the org _and_ repo level. Repo settings are already correct. Perhaps @kencochrane can check if `Allow GitHub Actions to create and approve pull requests` is checked at the org level.

krancour: @kencochrane were the org settings already like that or did you check that box just now?

kencochrane: I didn't change any values; that is what it had.

krancour: Thanks

krancour (approved):
I am dumbfounded by why the `pull-requests: write` permission isn't working, but am willing to accept this workaround in the interim.

.github/workflows/backport.yaml
hiddeco: @kencochrane I have a feeling the repository needs to be added to the allow list of the organization global secret.

kencochrane: we don't have an `AKUITYBOT_PAT` org secret on the `akuity` GitHub organization; we have one on our `akuityio` organization. That is probably the issue.

kencochrane: We will have to ask @jessesuen if he can add it since I don't have the AKUITYBOT_PAT value to add it.

Issue #1885, opened 2024-04-22
IAM auth to ECR support


jessesuen:
# Proposed Feature

Kargo needs to support IAM auth to ECR

# Motivation

ECR is a common container registry and we should have native support for it.

# Suggested Implementation

Follow guidelines here:
https://docs.aws.amazon.com/AmazonECR/latest/userguide/registry_auth.html

jessesuen:
I think it may be possible to workaround this limitation until Kargo natively supports ECR, with a CronJob that periodically updates an image repository Secret that Kargo understands.

https://linuxhandbook.com/auto-update-aws-ecr-token-kubernetes/
https://skryvets.com/blog/2021/03/15/kubernetes-pull-image-from-private-ecr-registry/

Will try and formulate an official workaround.

joebowbeer:
Can only a service account be used for ECR authentication? This is needed for IRSA or the newer Pod Identity method, and avoids the need for secrets.

For IRSA, an `eks.amazonaws.com/role-arn` annotation would be added to Kargo's service account.
For Pod Identity, the association would be pre-configured in AWS.

Keel implements a cache:

https://github.com/keel-hq/keel/blob/master/extension/credentialshelper/aws/aws.go

krancour:
@joebowbeer I don't think this is something that can realistically be handled at the pod level.

Kargo's credentials are managed at the Project level. This means that two Projects can use two different sets of credentials (with entirely different permissions) for talking to the same registry. It also means Projects that have no business talking to that registry cannot, for lack of credentials. This is essential to Kargo's tenancy model.

If credentials were handled at the pod level, all Projects, even those that do not need them, would have identical permissions on the registry and this would seem to undermine the tenancy model.

joebowbeer:
@krancour I interpret what you're saying to mean that, in Kargo, some inherent container identity can't be used to obtain the AWS creds.

I'm saying that there should be some method other than the less secure secretRef to obtain the creds.

ESO, for example, can obtain AWS credsFromSecretRef or credsFromServiceAccount (arn annotation)

https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/aws/auth/auth.go

There's an open issue to support pod identity directly, which has collected comments similar to yours:

https://github.com/external-secrets/external-secrets/issues/2951

A workaround involving two associated roles is described here:

https://containscloud.com/2024/03/24/integrating-aws-secrets-manager-to-eks-using-external-secrets/

[...]

PR #2018, opened 2024-05-15
refactor: use container image repo client from go-containerregistry

### LINKED ISSUE #1488: distribution/distribution v3 deprecates packages we use
I had to close #1461 because it breaks the build.

We use [distribution/distribution](https://github.com/distribution/distribution) for interacting with container registries.

In their upcoming v3 release, they've moved the entire `client` package and its sub-packages under `internal/`, making it now inaccessible to us.

This issue discusses:

https://github.com/distribution/distribution/issues/4119

From my reading of that issue, it may be _temporarily_ internal until it's been iteratively improved to a point where it becomes a publicly available and supported package of some future v3.x. But this is speculative.

Guidance seems to be that there is no harm in using v2 packages in the interim, but I also understand there to be a goal of EOL'ing v2 once v3 is GA. (In reality, we are actually using a "v3" that references the last commit to trunk prior to this change.)

This leaves me scratching my head as to what the official/recommended registry client is between v2 EOL and the speculative re-(re-)lease of a publicly available replacement.

I don't think there's much to do at this point other than keep an eye on the situation. If worse comes to worst, we have some options like forking it or copying their internal package into our own repo.

I just didn't want to lose track of this.

krancour:
Fixes #1488

This is a giant step toward solving #1885 since go-containerregistry supports quite a variety of authn options, including the ability to emulate ACR, ECR, and GCR Docker credentials helpers. Those new authn methods are not included in this PR. Those will come in a follow-up.

~~Note that I prefer we merge #1984 before this since I know there will be merge conflicts between the two PRs and I volunteer as tribute.~~ All set.

### DIFF
--- a/internal/image/creds.go
+++ b/internal/image/creds.go
@@ -1,34 +1,12 @@
 package image
 
-import "net/url"
-
 // Credentials represents the credentials for connecting to a private image
-// repository. It implements the
-// distribution/V3/registry/client/auth.CredentialStore interface.
+// repository.
 type Credentials struct {
 	// Username identifies a principal, which combined with the value of the
 	// Password field, can be used for reading from some image repository.
 	Username string
 	// Password, when combined with the principal identified by the Username
 	// field, can be used for reading from some image repository.
-	Password      string
-	refreshTokens map[string]string
-}
-
-// Basic implements distribution/V3/registry/client/auth.CredentialStore.
-func (c Credentials) Basic(*url.URL) (string, string) {
-	return c.Username, c.Password
-}
-
-// RefreshToken implements distribution/V3/registry/client/auth.CredentialStore.
-func (c Credentials) RefreshToken(_ *url.URL, service string) string {
-	return c.refreshTokens[service]
-}
-
-// SetRefreshToken implements
-// distribution/V3/registry/client/auth.CredentialStore.
-func (c Credentials) SetRefreshToken(_ *url.URL, service, token string) {
-	if c.refreshTokens != nil {
-		c.refreshTokens[service] = token
-	}
+	Password string
 }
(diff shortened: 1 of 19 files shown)


internal/image/creds.go
```diff
@@ -1,7 +1,5 @@
 package image
 
-import "net/url"
-
 // Credentials represents the credentials for connecting to a private image
 // repository. It implements the
 // distribution/V3/registry/client/auth.CredentialStore interface.
```
hiddeco, line 5: I _think_ this is no longer accurate?

krancour: Correct. Nice catch. Will fix.

krancour: Fixed.

internal/image/digest_selector.go
```diff
@@ -36,52 +38,33 @@ func newDigestSelector(
...
-		logger.WithFields(log.Fields{
-			"tag":    image.Tag,
-			"digest": image.Digest.String(),
-		}).Trace("found image")
-		return []Image{*image}, nil
+		return nil, fmt.Errorf("error retrieving image with tag %q: %w", tag, err)
+	}
+
+	if image == nil {
+		logger.Trace("image with tag did not match platform constraints", tag)
```
hiddeco, line 64:
```suggestion
		logger.Trace("image with tag did not match platform constraints")
```

krancour: Thanks... I was using `logger.Tracef()` for a hot minute and then added `tag` to the structured part of the log instead. Looks like I botched it when setting this back to `logger.Trace()`. I'm a bit surprised no linter caught this.

hiddeco: Linter typically only knows about the std libraries accepting format strings, which means it will just think this is a normal string here without looking further at what's in it.

hiddeco (approved):
PRs with a negative LOC are always great, if they also deal with deprecated dependencies it's even better.

_Three_ birds, one stone IMHO. 💯

PR #2020, opened 2024-05-16
fix(ui): improper alignment of empty label

rbreeze:


hiddeco (approved):
This fixes the issue I was seeing, thanks @rbreeze 🙇

PR #2024, opened 2024-05-16
fix(ui): correct logic for determining which freight are promo eligible

rbreeze:


hiddeco:
In its current form, this addresses the issue I had with the [kargo-advanced](https://github.com/akuity/kargo-advanced/) example where I couldn't promote Freight in the `staging` Stage to `ab-test-a`. It does however not solve the scenario in which none of the Stages have Freight, and I want to promote my first-ever Freight into the `dev` Stage.

image

hiddeco (approved):
The other issue should have been solved by #2026.

Issue #2027, opened 2024-05-16
chore: refactor last freight


krancour:
The `WarehouseStatus` type has a `LastFreight` field that is no longer used anywhere since #1984.

We could kill this off completely, although @hiddeco has suggested that at least keeping the ID of the last Freight created by the Warehouse might be useful to human users who are just looking to see if their warehouse is doing anything.

It's tricky to repurpose a field name because etcd will have structured data for existing Warehouses where we now want a string instead, so we agreed we will name it something like `LastFreightID` or similar.

PR #2034, opened 2024-05-17
feat(api)!: replace LastFreight in Warehouse w/ ID 

hiddeco:
Fixes: #2027
Follow-up to #1984 

This is breaking because it removes the `.status.lastFreight` field from the `Warehouse` API. Other than this, it will not cause any issues as the reconciler will automatically set the new field value (after an upgrade), even if the Freight already exists.


api/v1alpha1/warehouse_types.go
```diff
@@ -282,8 +282,9 @@ type WarehouseStatus struct {
 	// ObservedGeneration represents the .metadata.generation that this Warehouse
 	// was reconciled against.
 	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"varint,4,opt,name=observedGeneration"`
-	// LastFreight refers to the last Freight produced by this Warehouse
-	LastFreight *FreightReference `json:"lastFreight,omitempty" protobuf:"bytes,5,opt,name=lastFreight"`
+	// LastFreightID is a reference to the system-assigned identifier (name) of
+	// the most recent Freight produced by the Warehouse.
+	LastFreightID string `json:"lastFreightID,omitempty" protobuf:"bytes,8,opt,name=lastFreightID"`
```
hiddeco, line 287: @krancour we initially discussed this to be called `LastFreightReference`, but I did not like this because it could cause confusion with the `FreightReference` type.

krancour: Yep. I like `LastFreightID` better anyway.

krancour (approved):

Issue #2029, opened 2024-05-16
Workload Identity auth to Google Artifact Registry


jessesuen:
# Proposed Feature

Kargo should support workload identity auth to GCR.

This issue is the GCR equivalent of https://github.com/akuity/kargo/issues/1885.

# Motivation

Users wish to avoid using long-lived static credentials to GCR.

krancour:
https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr

PR #2037, opened 2024-05-17
feat(controller): add rudimentary ecr and artifact registry support

krancour:
Follow up to #2018. Does not quite solve #1885 or #2029 because this does not use workload identity.

As noted in #1885, a possible workaround was to use a CronJob to periodically exchange AWS access key ID and secret access key for a short-lived ECR authorization token (12 hours). The token is of the form `base64(AWS:<temporary-passsword>)`.

Similarly, a GCP service account key (which is structured data) can be exchanged for a GCP access token (one hour) for Artifact Registry. (I believe it works for GCR as well, although GCR is deprecated and as of a few days ago, new repos cannot be created in GCR.) This is an oauth2 access token, which can also be presented as a password (along with username `oauth2accesstoken`) when using basic auth.

This PR introduces functionality that seamlessly makes these exchanges. It also caches tokens to avoid excessive calls to the AWS and GCP APIs.

_Why are we doing this first and not jumping straight to using workload identity?_ 

Good question. Workload identity will not help a Kargo instance running _outside_ of AWS to access ECR  and it will not help a Kargo instance running _outside_ of GCP to access Artifact Registry -- which makes _this_ feature a lowest common denominator / universal solution for ECR and Artifact Registry integration.

Note: ACR (Azure Container Registry) does not require anything special because ACR provides for long-lived, user-managed tokens, with or without an expiration date.

I will circle back to add workload identity-based options for all three major cloud providers in a follow-up(s). See #1885 and #2029.

Docs to be included in a follow-up as well.


internal/credentials/ecr/ecr.go
```diff
@@ -0,0 +1,151 @@
...
+		return decodeAuthToken(entry.(string)) // nolint: forcetypeassert
+	}
+
+	encodedToken, err := c.getAuthTokenFn(region, accessKeyID, secretAccessKey)
+	if err != nil {
+		return "", "", fmt.Errorf("error getting ECR auth token: %w", err)
+	}
+
+	// Cache the encoded token
+	c.tokenCache.Set(cacheKey, encodedToken, cache.DefaultExpiration)
```
hiddeco, line 102: `DefaultExperiation` equals to forever. Shouldn't this be a little bit less than the TTL of the token itself (e.g. 10h)?

krancour: I don't `cache.DefaultExpiration` believe it does equal forever.

https://github.com/patrickmn/go-cache/blob/46f407853014144407b6c2ec7ccc76bf67958d93/cache.go#L29-L32

```golang
// For use with functions that take an expiration time. Equivalent to
// passing in the same expiration duration as was given to New() or
// NewFrom() when the cache was created (e.g. 5 minutes.)
DefaultExpiration time.Duration = 0
```

The implementation of the `Set()` function also supports the accuracy of that comment.

https://github.com/patrickmn/go-cache/blob/46f407853014144407b6c2ec7ccc76bf67958d93/cache.go#L54-L56

```golang
if d == DefaultExpiration {
	d = c.defaultExpiration
}
```

And I did indeed set the default expiration for `c.tokenCache` to `10*time.Hour`:

```golang
tokenCache: cache.New(
	// Tokens live for 12 hours. We'll hang on to them for 10.
	10*time.Hour, // Default ttl for each entry
	time.Hour,    // Cleanup interval
)
```

krancour:
@hiddeco note that I've amended what you already reviewed to add Google Artifact Registry support.

internal/credentials/credentials.go
```diff
@@ -221,10 +231,40 @@ func (k *kubernetesDatabase) getCredentialsSecret(
 	return matchingSecret, nil
 }
 
-func secretToCreds(secret *corev1.Secret) Credentials {
+func (k *kubernetesDatabase) secretToCreds(
+	ctx context.Context, credType Type, secret *corev1.Secret,
+) (Credentials, error) {
+	if credType == TypeImage {
+		// If the cred type is image, we'll try to derive username and password
```
hiddeco, line 238: Wonder if, at some point, it may be better to identify the type of credentials we are dealing with based on some user-defined value. Instead of the brute-force approach that's being taken now.

hiddeco (approved):

PR #2038, opened 2024-05-17
feat(warehouse): allow discovery interval and limit configurations

hiddeco:
This pull request adds the following things (as documented in the follow-up  section of #1984):

1. The option to define a custom interval at which the Warehouse attempts to discover new artifacts (instead of the `5m0s` default).

   ```yaml
   spec:
     interval: 1h0m
   ```
2. The option (for each subscription) to limit artifacts (instead of the default of `20`), with an upper limit of `100`.

   ```yaml
   spec:
     subscriptions:
     - git:
         discoveryLimit: 5
     - chart:
         discoveryLimit: 100
     - image:
         discoveryLimit: 25
   ```


api/v1alpha1/warehouse_types.go
```diff
@@ -55,6 +55,14 @@ type WarehouseSpec struct {
 	// empty, the defaulting webhook will set the value of this field to the value
 	// of the shard label.
 	Shard string `json:"shard,omitempty" protobuf:"bytes,2,opt,name=shard"`
+	// Interval is the interval at which the Warehouse should discover new
+	// artifacts. This field is optional. When left unspecified, the field is
+	// implicitly treated as if its value were "5m0s".
+	//
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(s|m|h))+$"
```
hiddeco, line 63: This is a common workaround for the issue described here: https://github.com/kubernetes/apiextensions-apiserver/issues/56

I took the further liberty to drop `ms` from the regex, as we probably do not want people to go below the seconds (and the level of precision in combination with e.g. minutes has little value).

internal/controller/warehouses/git.go
```diff
@@ -90,9 +90,13 @@ func (r *reconciler) discoverCommits(
 
 			for _, meta := range tags {
 				discovered = append(discovered, kargoapi.DiscoveredCommit{
-					ID:          meta.CommitID,
-					Tag:         meta.Tag,
-					Subject:     meta.Subject,
+					ID:  meta.CommitID,
+					Tag: meta.Tag,
+					// A decent subject length for a commit message is 50 characters
+					// (based on the 50/72 rule). We are nice people, and allow a
```
krancour, line 96: > We are nice people

😂 ❤️

krancour (approved):
Nice and clean.

hiddeco:
@krancour when I woke up on Saturday I actually had a small improvement in mind, might be worth thinking about this so I can potentially implement it tomorrow.

Besides the `discoveryLimit` on subscription items, it could be an idea to also have an object global fallback default:

```yaml
discovery:
  limit: 30
subscriptions:
- discoveryLimit: 100 # overwrite
```

Which could then also possibly result in:

```yaml
discovery:
  interval: 1h0s # now nested
```

Issue #1525, opened 2024-02-23
Store (versioned) tool binaries relative to project root directory


hiddeco:
Based on my comment in https://github.com/akuity/kargo/pull/1515#discussion_r1499139130

At present, we are doing a mixture of assuming certain tool binaries are present (globally) on a contributor their system, in combination with an additional assumption of them having a certain version.

In #1515, this is somewhat improved -- by at least ensuring there is a way to install the tools at their expected version. However, as this is still done globally this can create the new issue of it unexpectedly downgrading existing tools on a user their system.

To further improve this, I would suggest to install tool binaries relative to the project root directory (e.g. `/hack/bin`, see referred comment). Which can be combined with some Make magic to ensure the binaries are updated any time a version is updated in `go.mod`.

For some of the Node dependencies, a likewise approach could likely be used as I had to explore this to solve #1517.

krancour:
I agree this would be a positive improvement.

I'll note that anyone wanting a guarantee of using the correct tools has options for executing most tasks, including unit tests, linting, codegen, and more in containers using `hack-` variants of most make targets.

Even so, there is no harm in improving the situation for those who do not wish to use those containerized options for whatever reason. (Speed could be a factor on some local stacks.)

@hiddeco feel free to assign to yourself if this is something you're especially intent on improving.

PR #2046, opened 2024-05-21
chore: update go minor dependencies + codegen process tweaks

krancour:
Alternative to #2041, which breaks the build.

1. Uses #2041 as a starting point

1. Refactors for compatibility with k8s v1.30.1 and controller-runtime v0.18.2

1. Updates codegen process:

    We've been using tricks in `hack/install-tools.sh` to install the binaries associated with particular package versions enumerated in `go.mod`. This isn't working out so well at the moment, because k8s v1.30.1 (which we do want) forces us on to the latest `go-to-proto`, which has many breaking changes that I've been unable to resolve.

    The workaround I have settled for is something that is also quite straightforward -- simply ensuring correct versions of all codegen-related tooling in `Docker.dev` and in `.gihub/workflows/ci.yaml.` The result is that `go-to-proto` and its own requirements are no longer forced to be the version that comes with k8s v1.30.1.

I invite @devholic and/or @hiddeco to find a better way.


hiddeco:
Have you tried putting a separate `go.mod` in `hack/`? I can imagine this would decouple it from the Kubernetes dependencies in root.

krancour:
> Have you tried putting a separate go.mod in hack/? I can imagine this would decouple it from the Kubernetes dependencies in root.

iirc, when I attempted that, codegen failed because of all the things that were in the main `go.mod` that weren't in that dedicated tooling one.

krancour:
@hiddeco please feel free to amend my PR if you stumble on anything.

hiddeco:
Will play around with it for a bit.

hiddeco:
The alternative workaround I could come up with is: https://github.com/hiddeco/kargo/commit/424030577299ee1352ecea3d44032a09268f686a

This uses a [Go workspace](https://go.dev/ref/mod#workspaces) to force a downgrade for the code-generator libraries. As we may need this eventually anyway to deal with the breaking changes from upstream you could not figure out (like [the removal of trim-path logic](https://github.com/kubernetes/code-generator/commit/81cc0e31682fedd9c0d22ebf01406e2c9a8c824a)), which they removed in favor of proper Go Modules resolving (xref: `https://github.com/kubernetes/kubernetes/pull/123529`).

krancour:
> This uses a [Go workspace](https://go.dev/ref/mod#workspaces) to force a downgrade

I had vaguely suspected workspaces could be a viable approach to this, especially since the breaking changes to the go-to-proto [seem to have been _on account of_ Kubernetes' own workspaces adoption](https://kubernetes.io/blog/2024/04/17/kubernetes-v1-30-release/#go-workspaces-for-k-k-sig-architecture-https-github-com-kubernetes-community-tree-master-sig-architecture).

When I was researching workspaces, the sort of canonical example that was popping up everywhere involved resolving modules to local directories -- useful if you want to hack on a dependency, but frankly, something that could already be done with a `replace` directive. I wasn't quite able to connect the dots on how it could help us here.

👀 on https://github.com/hiddeco/kargo/commit/424030577299ee1352ecea3d44032a09268f686a

Another option -- which I've equally been unable to figure out -- is cracking the correct usage of the newer go-to-protobuf.

krancour:
After playing with workspaces more, we found that the proposed solution results in controller-gen v0.29.5 in Kargo's build path. It might be harmless, but for the interim, keeping the installation of codegen tools separate from Kargo's own build-time dependencies has the benefit of being the most straightforward thing to do.

We can circle back on this at a later time, but I do want to unblock the k8s and controller runtime upgrades sooner rather than later.

hiddeco (approved):
We agreed offline on looking at a way to separate this in a better way, potentially in combination with #1525, after the `v0.7.0` release.

For now, let's just unblock and be done with it.

PR #2048, opened 2024-05-21
docs: slightly reorder roadmap entries

hiddeco:
While looking up some information on the roadmap, it didn't feel right that I had to scroll to the bottom of the page to become aware of what the "next new thing" would be. This is an attempt to solve this annoyance.

### DIFF
--- a/docs/docs/50-roadmap.md
+++ b/docs/docs/50-roadmap.md
@@ -11,62 +11,13 @@ completed per release.
 
 :::caution
 This roadmap tracks only _major_ features and is subject to change at any time,
-for the most up to date information, please see the [GitHub
+for the most up-to-date information, please see the [GitHub
 Project](https://github.com/akuity/kargo/milestones)
 :::
 
-## v0.2.0
+## In Progress
 
-__Status:__ Completed
-
-| Name | Type | Description |
-| ---- | ---- | ----------- |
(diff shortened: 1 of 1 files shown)


christianh814 (approved):

krancour:
The ordering this introduces is:

1. v0.7.0 (current)
1. v0.6.0..v0.1.0 (past)
1. 2. v0.8.0 (future)

This feels strange, because if the intent is to position things that are _currently_ most relevant closest to the top of the page, I'd suggest the _next_ few releases are of greater interest than the _last_ few.

imho, this would be a better order:

1. v0.7.0 (current)
1. v0.8.0 (future)
1. v0.6.0..v0.1.0 (past)

Although it _also_ feels strange, as it's neither fully ascending nor fully descending.

I think this could be solved by breaking the roadmap into two distinct sections:

## The road ahead
### v0.7.0
### v0.8.0
## The road behind us
### v0.6.0
### v0.5.0
### v0.4.0
### etc.

wdyt?

hiddeco:
I agree with your observation, but have a suggestion on the subsection titles. What about just calling them "In progress" and "Completed"?

krancour:
> What about just calling them "In progress" and "Completed"?

Fine with me. As long as we use two sections to put boundaries around the two different sort orders we'd be using, I'm happy.

morey-tech (approved):

PR #2052, opened 2024-05-22
chore: update go-to-protobuf usage

devholic:
> Related: https://github.com/akuity/kargo/pull/2046

This PR updates protobuf codegen to use the latest go-to-protobuf (v0.30.1). 

The new codegen process is done by

- Creating a dummy module in a temporary directory to provide isolated directory to the go-to-protobuf.
- Use vendored proto files instead of generating from Kubernetes types

Reference:

- https://github.com/kubernetes/code-generator/compare/v0.29.5...v0.30.1#diff-028a4953abcbbe704c55b50d01ff8001a791acb61504f409fcd748cf93c3658e


ui/src/gen/k8s.io/api/batch/v1/generated_pb.ts
```diff
@@ -1,3 +1,18 @@
+//
+//Copyright The Kubernetes Authors.
```
hiddeco, line 2: Is this a sign the `--go-header-file` is not working as expected?

krancour:
After looking at this, it's surfacing a concern I've had for some time -- that this process is becoming too complex. I don't believe that the creators of the various tools we're using ever intended for anyone to have to jump through this many hoops, so what is it that we are overlooking?

krancour:
@hiddeco were the problems potentially solved by this PR solved differently through your recent work on managing tools better?

hiddeco:
Not in full, but I have a local copy somewhere that incorporates what's being attempted in this PR on top of the changes from #2449, and I am planning to pick this up as part of #2573.

hiddeco:
This has been addressed by #2573. (Thanks for the work nonetheless, @devholic 🙇)

Issue #1250, opened 2023-12-08
Proposal: File/Directory/Patch based promotions


jessesuen:
# Proposed Feature

There are use cases of promotions where the act of promotion may include copying files or directories from one path to another. Additionally, or alternatively, the promotion may need to modify/patch files in a certain way as part of the promotion.

Conceptually, this is no different than what already happens when Kargo promotes container images on the main branch using kustomize, at different paths:

```
# promote to dev
$ cd env/dev && kustomize edit set image acme/guestbook:v2.0.0

# promote to staging
$ cd env/staging && kustomize edit set image acme/guestbook:v2.0.0

# promote to prod
$ cd env/prod && kustomize edit set image acme/guestbook:v2.0.0
```

For each promotion, you could describe what Kargo is doing as promoting a "patch" to `kustomization.yaml` files from path to path, where the patch is upserting an entry in the `images:` stanza:

```yaml
images:
- name: acme/guestbook
  newTag: v2.0.0
```

This proposal is to generalize the ability to modify files into something that is not kustomize/kubernetes specific, and could apply to anything configuration-related:
* json
* yaml
* hcl

# Motivation

Users of Kargo shouldn't be limited to, or beholden to the choice of tooling Kargo natively decided to support. This feature will expand the scope and capabilities of what Kargo is able to promote, such as Terraform.

# Suggested Implementation

Largely TBD but I imagine we'll need ways to describe:
* filepaths
* exclusion filters
* extraction/parsing logic
* ways to manipulate a configuration file (e.g. yq, jq, sed)

# Open Issues

When we move to support patch-based promotion, I think Kargo has to start treating "rollback" as something **_different_** than "promotion". Because while we may easily know how to apply a patch as part of the promotion, we would not easily know how to "undo" that patch for a rollback. 

Ultimately, I think we may need to require the use of rendered branches to support rollback, and then rollback becomes nothing more than re-applying the contents of an older commit in that branch as the tip.

jessesuen:
I forgot about one use case that we even practice internally. I would like a way for kargo to monitor git-repo-1/path for changes, and then whenever changes are made in that path, copy/replace the directory contents of git-repo-1/path to git-repo-2/path.

We constantly make changes/adjustments to helm chart contents in an **_upstream_** git repo. Let's say:
https://github.com/akuity/kargo/tree/main/charts/kargo

[...]

PR #2054, opened 2024-05-23
feat(promotions): support copying folders and files for Git

hiddeco:
Related to #1250 (and more specifically https://github.com/akuity/kargo/issues/1250#issuecomment-2010724428).

This adds minimal support for copying files or directories into user-defined paths during a Git promotion. The primary goal of this pull request is to get a feel of how it _could_ work, as the issue appears to be more of a collection of desired outcomes rather than concrete descriptions of functionalities.

```yaml
spec:
  promotionMechanisms:
    gitRepoUpdates:
    - repoURL: https://github.com/hiddeco/kargo-test.git
      writeBranch: env/dev
      patches:
      # From a Freight commit
      - copy:
          repoURL: https://github.com/example/charts.git
          source: base/values.yaml
          destination: env/dev/values.yaml
      # Within the same repository
      - copy:
          source: base/values.yaml
          destination: env/dev/values.yaml
      render: {}
```


krancour:
I think we can close this since #2536 has this.

PR #2059, opened 2024-05-24
chore: switch from aws-sdk-go v1 to v2

krancour:
I mistakenly used aws-sdk-go v1 in #2037

v2 is preferable because it helps make shorter work of #1885

PR #2070, opened 2024-05-29
feat(controller): support gcp workload identity federation as auth method for google artifact registry

### LINKED ISSUE #2029: Workload Identity auth to Google Artifact Registry

krancour:
Follow up to #2059 and #2066

Fixes #2029 

This PR makes minor tweaks to ECR support and expands Google Artifact Registry support to include integration with GCP Workload Identity Federation.


internal/credentials/gar/workload_identity_federation.go
```diff
@@ -0,0 +1,135 @@
...
+	"time"
+
+	"cloud.google.com/go/compute/metadata"
+	"github.com/patrickmn/go-cache"
+	"google.golang.org/api/iamcredentials/v1"
+
+	"github.com/akuity/kargo/internal/logging"
+)
+
+var garURLRegex = regexp.MustCompile(`^.+\.pkg\.dev/`)
```
jessesuen, line 16: Though google has deprecated Google Container Registry, the `gcr.io` domain is still valid and [is possible to be hosted by artifact registry](https://cloud.google.com/artifact-registry/docs/transition/gcr-on-ar) by existing gcr.io users.

I think this should be relaxed to also check for `gcr.io`

jessesuen (approved):
Minor nit about valid container repository domain, but other than that LGTM

internal/credentials/gar/workload_identity_federation.go
krancour: Nice catch. Will relax.

krancour: Fixed in https://github.com/akuity/kargo/pull/2070/commits/f648029daf0f199c0f193328fac3ccaf27eaee92

PR #2071, opened 2024-05-29
feat(ui): Freight assembly

### LINKED ISSUE #761: Manually input freight

rbreeze:
Fixes #761

CleanShot 2024-05-29 at 10 49 42@2x

CleanShot 2024-05-30 at 14 45 57@2x
CleanShot 2024-05-30 at 14 46 05@2x
CleanShot 2024-05-30 at 14 46 13@2x
CleanShot 2024-05-30 at 14 45 35@2x

### DIFF
--- a/charts/kargo/templates/api/cluster-role.yaml
+++ b/charts/kargo/templates/api/cluster-role.yaml
@@ -44,6 +44,7 @@ rules:
     resources:
       - freights
     verbs:
+      - create
       - get
       - list
       - patch
--- a/ui/src/config/paths.ts
+++ b/ui/src/config/paths.ts
@@ -7,7 +7,7 @@ export const paths = {
   projectEvents: '/project/:name/events',
   projectRoles: '/project/:name/roles',
   stage: '/project/:name/stage/:stageName',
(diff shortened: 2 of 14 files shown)


krancour (approved):
This works _really well._

My only feedback is that the interface doesn't offer the possibility of creating Freight with references to (very old) artifacts that the Warehouse has not discovered, but I think that is something we ought to address in a follow-up.

@rbreeze I also hope you don't mind that I amended the PR with a tiny permissions change for the API server, which now needs to be able to create Freight resources.

🔥 🔥 🔥

Issue #2068, opened 2024-05-28
bug: errors in controller when AnalysisRun CRD is not installed


krancour:
We meant to be tolerant of scenarios where Rollouts CRDs are not installed, but this seems to have crept in around the same time we started emitting k8s events:

```
time="2024-05-28T21:33:00Z" level=error msg="kind must be registered to the Scheme" error="no kind is registered for the type v1alpha1.AnalysisRun in scheme \"pkg/runtime/scheme.go:100\"" logger=controller-runtime.source.EventHandler
```

After enough of these the controller will restart.

@devholic tentatively assigning to you because I think you will find the underlying cause fastest. Please lmk if you need me to take it up instead.

devholic:
@krancour I couldn't reproduce this problem, so I checked event recording logics. 

The only suspicious part was `getAnalysisRun` is called when recording an `FreightVerificationEvent` in stage controller.

https://github.com/akuity/kargo/pull/2074 may resolves this issue, but as far as I know this should not happen because `VerificationInfo` should not have `AnalysisRun` when the integration is disabled 🤔

krancour:
@devholic I figured out why you cannot reproduce it.

I think you must have explicitly disabled Rollouts integration, _while I did not._ I simply installed into a cluster that lacked Rollouts CRDs.

For both Argo CD and Rollouts, we detect the condition of missing CRDs, even if integration is explicitly enabled. For Argo CD, if the CRDs are not found, we don't create a client to speak to the Argo CD control plane. Everywhere else in the code, we gate Argo CD-specific things on whether that client is nil or non-nil.

We used to do the same for Rollouts when Rollouts used to have its own client, but since #1706 we assume Rollouts resources to live in the Kargo control plane and we eliminated the possibility of it having its own client. From that point forward, we have gated Rollouts-specific things only on whether that integration is explicitly enabled or disabled.

_So it is possible to have Rollouts integration enabled, Rollouts CRDs missing, and for the graceful failure we intended to not work._

#2074 won't resolve this, but since I've got a very clear handle now on what's going on, I don't mind taking this over since I can resolve it quickly.

PR #2074, opened 2024-05-30
fix: skip getAnalysisRun call if integration is disabled

devholic:
May _resolves_ #2068


krancour:
See https://github.com/akuity/kargo/issues/2068#issuecomment-2139646301

PR #2077, opened 2024-05-30
docs: document cred management with cli + specific support for ecr and google artifact registry

### LINKED ISSUE #2067: document new image repo authn options
New registry-specific authentication options exist for image repos:

* ECR
    * Put access key, secret key, and region in credentials (Secret). This is good for cases where ECR is accessed from _outside_ EKS or from an EKS cluster that does not have EKS Pod Identity enabled.
    * Skip the credentials (Secret) and use EKS Pod Identity. I don't want this to become an exercise in documenting how to set up EKS Pod Identity itself. We can link to relevant documentation for that. The important part is that the Kargo controller gets a single IAM role, and that role will need permissions to assume _Project-specific_ roles. This is what limits what images individual Projects can or cannot access. Need to document the trust policies required and the naming conventions assumed for Project-specific roles.
    * IRSA (IAM Roles for [Kubernetes] Service Accounts) is not supported, but may be in the future. EKS Pod Identity is a superior option to this anyhow.
* ACR: At present, we only support long-term creds (username/password) for ACR, which can be created/managed via Azure portal or `az` CLI.
* Google Artifact Registry. (GCR is deprecated.)
    * Put base64-encoded service account key in credentials (Secret) with key `gcpServiceAccountKey`. This is good for cases where Artifact Registry is accessed from _outside_ GKE.
    * Skip the credentials (Secret) and uses GKE Workload Identity Federation. Currently WIP

krancour:
Fixes #2067

### DIFF
--- a/docs/docs/30-how-to-guides/20-managing-credentials.md
+++ b/docs/docs/30-how-to-guides/20-managing-credentials.md
@@ -217,8 +217,8 @@ project `Namespace`s.
 ## Registry-Specific Authentication Options
 
 While many container image registries support authentication using long-lived
-credentials, such as a username plus a password or personal access token, some
-do not.
+credentials, such as a username and password (or personal access token), some
+either require or offer more secure options.
 
 This section provides registry-specific guidance on credential management and
 also covers options for gaining image repository access using workload identity
@@ -315,9 +315,18 @@ options outlined here may be applied.
 
 #### Long-Lived Credentials
 
-Google Artifact Registry does not _directly_ support long-lived credentials,
-however, a Google service account key
-[can be exchanged for a token](https://cloud.google.com/artifact-registry/docs/docker/authentication#token)
+:::caution
+Google Artifact Registry does _directly_ support long-lived credentials
+[as described here](https://cloud.google.com/artifact-registry/docs/docker/authentication#json-key).
+The username `_json_key_base64` and the base64-encoded service account key
+may be stored in the `username` and `password` fields of a `Secret` resource as
+described [in the first section](#credentials-as-kubernetes-secret-resources) of
+this document. Google's own documentation strongly recommends against this
+method however.
+:::
+
+Google documentation recommends
+[exchanging a service account key for an access token](https://cloud.google.com/artifact-registry/docs/docker/authentication#token)
 that is valid for 60 minutes. Kargo can seamlessly make this exchange and will
 cache the resulting token for a period of 40 minutes.
 
@@ -410,9 +419,7 @@ Google Artifact Registry repositories.
 
 ### Azure Container Registry (ACR)
 
-Azure Container Registry is not strictly subject to the same limitations as ECR
-and Google Artifact Registry in terms of an inability to support long-lived
-credentials.
+Azure Container Registry directly supports long-lived credentials.
 
 It is possible to
 [create tokens with repository-scoped permissions](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-repository-scoped-permissions),
@@ -423,6 +430,6 @@ document.
 
 :::info
 Support for authentication to ACR repositories using workload identity, on par
-with Kargo's support for ECR and Google Artifact Registry is likely to be
+with Kargo's support for ECR and Google Artifact Registry, is likely to be
 included in a future release of Kargo.
 :::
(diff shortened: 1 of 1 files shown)


docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -122,3 +122,307 @@ It is important to understand the security implications of this feature. Any
...
+Registry. For authentication to repositories with legacy GCR URLs, the same
+options outlined here may be applied.
+:::
+
+#### Long-Lived Credentials
+
+Google Artifact Registry does not _directly_ support long-lived credentials,
+however, a Google service account key
+[can be exchanged for a token](https://cloud.google.com/artifact-registry/docs/docker/authentication#token)
+that is valid for 60 minutes. Kargo can seamlessly make this exchange and will
```
jessesuen, line 321: > Google Artifact Registry does not _directly_ support long-lived credentials,

Actually, unlike ECR, GAR **_can_** be accessed with a long-lived service account key as a password, so we should change this wording. Credentials look like:

username: `_json_key_base64`
password: base64 encoded of some json like:

```json
{
  "type": "service_account",
  "project_id": "myproject",
  "private_key_id": "xxxxxx",
  "private_key": "-----BEGIN PRIVATE KEY-----\nyyyyyyy==\n-----END PRIVATE KEY-----\n",
  "client_email": "foo@myproject.iam.gserviceaccount.com",
  "client_id": "12345",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://oauth2.googleapis.com/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/foo%40myproject.iam.gserviceaccount.com"
}
```

This is documented at https://cloud.google.com/artifact-registry/docs/docker/authentication#json-key,

We can point them there if they prefer/need to use long-lived credentials and explain that GAR supports two methods of accessing private repos.

TBH, I'm not sure what benefits "Access Token" auth has over "Service Account Key" since Access Token would still need a long lived service account credential anyways.

krancour: 🤦‍♂️ Embarrassed that I missed that part of the docs, and also question the value of access tokens under these circumstances. I will document this as a third option for authenticating to GAR. For my own edification, I might also research if/why exchanging a service account key for an access token might have any benefit over using the key directly.

krancour: Addressed. Directly using the service account token is now listed as an option that works, but which Google discourages.

I have also posted a question about that recommendation on security.stackexchange.com.

https://security.stackexchange.com/questions/277163/why-use-access-token-for-google-artifact-registry-access

jessesuen (approved):

docs/docs/30-how-to-guides/20-managing-credentials.md
krancour: I have much more clarity about this now, and it would have been obvious if I'd re-read my own code after the inital confusion had set in...

I think I allowed the fact that `_json_key_base64` and a base64-encoded service account key _can_ be used as user/pass in basic auth to the registry to mislead me into believing that `gcloud auth print-access-token` (or programmatic equivalent) was using the service account key _in the same way_ -- which is totally not the case.

The service account key is actually used to sign a JWT and it is the JWT that is exchanged for an access token -- meaning that when you request an access token, the service account key does not at any point traverse the wire.

It's right here:

https://github.com/akuity/kargo/blob/088fc4d358d2ed69879e3133d753ea50e1d12a65/internal/credentials/gar/service_account_key.go#L105-L113

This being the case, the benefits of obtaining an access token over using the service account key to access the registry directly are more obvious.

Two action items for me:

1. Without going into exhaustive detail, understanding this much better now compels me to somewhat strengthen my caution against the most direct approach. I will amend this PR accordingly.

2. While researching, I noticed that when requesting the access token, I am requesting a read/write scope, which is quite unnecessary for how Kargo interacts with a registry, which is always in a read-only fashion. I will fix this in a separate PR.

krancour: Docs amended in b94cb5236a3c3737d46876b6eaea9aa0300fcfa6

PR #2088, opened 2024-06-03
feat(controller): allow use of multi sources apps

### LINKED ISSUE #1399: health checks for Applications with multiple sources can probably work now
### Discussed in https://github.com/akuity/kargo/discussions/1398

<sup>Originally posted by **daftping** January 18, 2024</sup>
Seems like the health check for Applications with multiple sources is not supported right now.
https://github.com/akuity/kargo/blob/8b6f58a4ddfc81e004bb782de69a45a575d29a44/internal/controller/stages/health.go#L98-L100
Does anybody know what Argo CD issue this TODO is referring to?

```yaml
  health:
    status: Unknown
    issues:
      - bugs in Argo CD currently prevent a comprehensive assessment of the
        health of multi-source Application "test-app-helm" in
        namespace "argo-cd"
```</div>

### PR DESCRIPTION
**Pull request by akolacz**
Possibly fixes #1399

We're currently using Kargo with multi sources apps in ArgoCD. Because of that, we've noticed that the health status of our ArgoCD applications in Kargo are set to status unknown, and the verification is continuously spinning due to this following code :

```
TODO: We should re-evaluate this soon. It may have been fixed in recent
	//       versions.
	// TODO(hidde): Do we have an upstream reference for this?
	if len(app.Spec.Sources) > 0 {
		err := fmt.Errorf(
			"bugs in Argo CD currently prevent a comprehensive assessment of "+
				"the health of multi-source Application %q in namespace %q",
			key.Name,
			key.Namespace,
		)
		return kargoapi.HealthStateUnknown, healthStatus, syncStatus, err
	}
```

We're using Kargo 0.6.0 and ArgoCD 2.11.2 (both on latest versions available).
We forked the project to remove this test and managed to get our multi sources apps to work.
Here's an ArgoCD application example:

```
kind: Application
metadata:
  annotations:
    kargo.akuity.io/authorized-stage: kargo-test:test
  name: test
  namespace: argocd
spec:
  project: test
  sources:
    - repoURL: test
      chart: test/test
      targetRevision: 1.0.0
      helm:
        valueFiles:
          - $values/test/values.yaml
    - repoURL: https://github.com/test
      targetRevision: main
      ref: values
  destination:
    name: test
    namespace: test
```

Results inside Kargo:

image

The issue seems to be fixed in ArgoCD, maybe one of these:
https://github.com/search?q=repo%3Aargoproj%2Fargo-cd+multi+sources&type=commits

### DIFF
--- a/internal/argocd/health.go
+++ b/internal/argocd/health.go
@@ -160,19 +160,6 @@ func (h *applicationHealth) GetApplicationHealth(
 		}
 	}
 
-	// TODO: We should re-evaluate this soon. It may have been fixed in recent
-	//       versions.
-	// TODO(hidde): Do we have an upstream reference for this?
-	if len(app.Spec.Sources) > 0 {
-		err := fmt.Errorf(
-			"bugs in Argo CD currently prevent a comprehensive assessment of "+
-				"the health of multi-source Application %q in namespace %q",
-			key.Name,
-			key.Namespace,
-		)
-		return kargoapi.HealthStateUnknown, healthStatus, syncStatus, err
-	}
-
 	// Check for any error conditions. If these are found, the application is
 	// considered unhealthy as they may indicate a problem which can result in
 	// e.g. the health status result to become unreliable.
--- a/internal/argocd/health_test.go
+++ b/internal/argocd/health_test.go
@@ -399,56 +399,6 @@ func TestApplicationHealth_GetApplicationHealth(t *testing.T) {
 			},
 		},
 		{
-			name: "error on multiple app sources",
-			key:  types.NamespacedName{Namespace: "fake-namespace", Name: "fake-name"},
-			stage: &kargoapi.Stage{
-				Spec: testStageSpec,
-			},
-			application: &argocd.Application{
-				ObjectMeta: metav1.ObjectMeta{
-					Namespace: "fake-namespace",
-					Name:      "fake-name",
-				},
-				Spec: argocd.ApplicationSpec{
-					Sources: argocd.ApplicationSources{
-						{},
-						{},
-					},
-				},
-				Status: argocd.ApplicationStatus{
-					Health: argocd.HealthStatus{
-						Status:  argocd.HealthStatusHealthy,
-						Message: "fake-message",
-					},
-					Sync: argocd.SyncStatus{
-						Status:    argocd.SyncStatusCodeSynced,
-						Revision:  "fake-revision",
-						Revisions: []string{"fake-revision1", "fake-revision2"},
-					},
-				},
-			},
-			assertions: func(
-				t *testing.T,
-				state kargoapi.HealthState,
-				healthStatus kargoapi.ArgoCDAppHealthStatus,
-				syncStatus kargoapi.ArgoCDAppSyncStatus,
-				err error,
-			) {
-				require.ErrorContains(t, err, "bugs in Argo CD currently prevent a comprehensive assessment")
-
-				require.Equal(t, kargoapi.HealthStateUnknown, state)
-				require.Equal(t, kargoapi.ArgoCDAppHealthStatus{
-					Status:  kargoapi.ArgoCDAppHealthStateHealthy,
-					Message: "fake-message",
-				}, healthStatus)
-				require.Equal(t, kargoapi.ArgoCDAppSyncStatus{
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by krancour** [comment]
@akolacz I will take a look at this, but to the best of my recollection, the big problem with multi-source apps that we ran into was that unlike in the case of a single source app, there was not adequate status information maintained on the Argo CD side for our health checks to confirm that each source was synced to the desired revision of the applicable artifact.

I know at the time that the message about lack of multi-source App support was added, our health checks were not even _trying_ to assess the sync state of multiple sources, so a PR that only removes line and doesn't add any (such as this one) seems it might still be missing some stuff...

However, @hiddeco has also refactored the health checks quite a lot since that time, so it's possible that this is better accounted for now than I believed it to be.

At any rate, we will take a careful look at this PR in the v0.8.0 timeframe. i.e. Starting next week.

Thank you!

**Comment by hiddeco** [comment]
What would help me to review this (quicker) is a simple multi-app example that I can easily use as a testing bed. As my comment and involvement here go as far as attempting to figure out the background of the other comment while refactoring this a bit.

However, with a working example, I can _probably_ combine this with (and enrich!) my knowledge of the further mechanics around health checks, and perform an adequate review.

**Comment by akolacz** [comment]
I send to you these 2 examples of multi sources.

The first example is our use case, multi source with 1 ArgoCD application and the other source is git-hosted values.

```
kind: Application
metadata:
  annotations:
    kargo.akuity.io/authorized-stage: test:multiSourceSingleApp
  name: multiSourceSingleApp
  namespace: argocd
spec:
  project: default
  sources:
    - repoURL: 'https://prometheus-community.github.io/helm-charts'
      chart: prometheus
      targetRevision: 15.7.1
      helm:
        valueFiles:
        - $values/charts/prometheus/values.yaml
    - repoURL: 'https://git.example.com/org/value-files.git'
      targetRevision: dev
      ref: values
  destination:
    name: "in-cluster"
    namespace: default
```

The second example is an ArgoCD application that uses 2 apps under it, one with Nginx and another with Prometheus.

```
kind: Application
metadata:
  annotations:
    kargo.akuity.io/authorized-stage: test:multiSourceMultiApp
  name: multiSourceSingleApp
  namespace: argocd
spec:
  project: default
  sources:
    - repoURL: registry-1.docker.io/bitnamicharts
      chart: nginx
      targetRevision: 15.9.0
    - repoURL: 'https://prometheus-community.github.io/helm-charts'
      chart: prometheus
      targetRevision: 15.7.1
  destination:
    name: "in-cluster"
    namespace: default
```

Let me know if this is what you expected.

akolacz:
Hello @krancour and @hiddeco

I made a full structure with ArgoCD applications and all Kargo ressources needed for this test. 
You will find: 

- a multi sources with 1 apps and Github hosted values.
- a multi sources  with 2 apps and Github hosted values.

You can used it to do your test, I also wrote a README on how to use it.

Repo: <https://github.com/akolacz/argocd-multi-sources>

Hope this will help you for your test.

hiddeco:
I ran some tests with this, and I think this introduces the base functionalities required to assess the general health of an ArgoCD multi-source Application. Where it appears to fall short, is the assessment of finer-grain details around desired revisions – which is currently written to deal with a _single_ revision:

https://github.com/akuity/kargo/blob/e2a8e5f2baad188b3fd42306529b62fd638e302b/internal/argocd/health.go#L178-L186

For this, I am unsure all required information exists in the current Status data, as it appears to be limited to a slice of revisions _without_ identifiers of the e.g. chart it is for.

akolacz:
Hello,

Do you think we can start like this and handle finer-grain later ?
If not, what do you think about adding a configuration option to unlock the feature ?

gnadaban:
Hey @akolacz , I recently opened #2160 to address various issues I found around multi-source apps, check it out.

akolacz:
Hi @gnadaban i will have a look thanks

krancour:
Although #2160 has some issues that still need attention, I think the direction it is moving in of at least attempting to discern desired revisions and check if sources are synced to them is closer to the direction we need to head in. In the interest of keeping all the solutioning for this problem in one place instead of having two competing options, I am closing this one in favor of #2160.

akolacz:
Thx quite agree with you, we need to focus to one solution :)

PR #2096, opened 2024-06-03
fix(chart): provide empty default annotations map for api server ingress

### LINKED ISSUE #2095: `helm template` fails when `api.ingress.annotations` unspecified
# Description

Updated to nightly chart version `0.7.0-unstable-20240603` but `helm template` now fails with:

```
Error: Error: template: kargo/templates/api/ingress.yaml:10:72: executing "kargo/templates/api/ingress.yaml" at <.Values.api.ingress.annotations>: wrong type for value; expected map[string]interface {}; got interface {}
```

I believe it was introduced by this:

```yaml
  {{- with (mergeOverwrite (deepCopy .Values.global.annotations) .Values.api.ingress.annotations) }}
  annotations:
    {{- range $key, $value := . }}
    {{ $key }}: {{ $value | quote }}
    {{- end }}
  {{- end }}
```

(if I remove `.Values.api.ingress.annotations` from the line, it renders properly).

My values stanza is simply:

```yaml
api:
  ingress:
    enabled: true
```

# Version

```
0.7.0-unstable-20240603
```

krancour:
Fixes #2095


hiddeco (approved):

hiddeco:
Requires a codegen rerun.

krancour:
> Requires a codegen rerun.

🤦‍♂️ Fixed.

PR #2105, opened 2024-06-04
fix: Fix missing pod annotations from API server and add support for podLabels in Helm chart

gnadaban:
This PR adds the ability to annotate API pods and label pods with custom labels.


gnadaban:
@krancour I'm not sure if there's anything I missed other than the signoff and signing, let me know and I'll update.

gnadaban:
Ideally all components should also support podAnnotations to support tools like [Bank-Vaults vault-secrets-webhook](https://bank-vaults.dev/docs/mutating-webhook/).

charts/kargo/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: kargo
 description: Kargo
 type: application
-version: 0.1.0
-appVersion: v0.1.0
+version: 0.1.1
+appVersion: v0.6.0
```
krancour, line 6: This should always be left alone. It is automatically set correctly by the release process.

charts/kargo/values.yaml
```diff
@@ -95,6 +96,10 @@ api:
   host: localhost
   ## @param api.logLevel The log level for the API server.
   logLevel: INFO
+
+  ## @param api.podLabels Optional additional labels to apply to pods.
+  podLabels: {}
```
krancour, line 101: Two things:

1. I suggest that this be renamed to just `labels` because we already do something similar to this for annotations and the key for that is just `annotations` (not `podAnnotations`). I think this will just look more consistent.

2. I also suggest relocating this so that it's right above the aforementioned `annotations` key.

This applies to the other occurrences of `podLabels` throughout this PR as well.

krancour: I think we should try to more closely emulate what's already been done with annotations. i.e. It would be nice to see optional "global" labels merged with / potentially overridden by component-level labels.

krancour:
@gnadaban thanks for this. We're certainly amenable to something like this, but I left a few comments about implementation details -- mainly focused about keeping things maximally consistent with precedents already set elsewhere in the chart. If those are addressed, this enhancement to the chart seems pretty useful.

charts/kargo/Chart.yaml
gnadaban: That may be, but if one tries to deploy eg. with ArgoCD from Git directly the missing appVersion causes the deployment to break. Thoughts?

charts/kargo/values.yaml
gnadaban: I'd like you to reconsider the naming. Well-known open-source projects like [cert-manager](https://github.com/cert-manager/cert-manager/blob/c5013dbf9543e9f1c25f1dffed5c78ec28272ac5/deploy/charts/cert-manager/values.yaml#L319), [ArgoCD](https://github.com/argoproj/argo-helm/blob/19ef184d8fe923fdc0ec42411c6272a91138db31/charts/argo-cd/values.yaml#L83) or [external-dns](https://github.com/kubernetes-sigs/external-dns/blob/bed56b7151be415a8af0d7182ed675fd7ced2b67/charts/external-dns/values.yaml#L67) and even [Argo Workflows](https://github.com/argoproj/argo-helm/blob/19ef184d8fe923fdc0ec42411c6272a91138db31/charts/argo-workflows/values.yaml#L117) use `podLabels` and `podAnnotations` to clarify that the labels are intended for the Pod resources. 

While there's no formal standard, me and likely others see this as a clear pattern, and will likely find it more intuitive.

gnadaban: > I think we should try to more closely emulate what's already been done with annotations. i.e. It would be nice to see optional "global" labels merged with / potentially overridden by component-level labels.

I'm not sure that I'd want the same labels on Pods that I'd put on Deployments. For example, a Gatekeeper policy might mandate certain labels on Pods, but not necessarily on all the resources a global labels setting would affect, eg. roles, configmaps, etc.

krancour: > I'm not sure that I'd want the same labels on Pods that I'd put on Deployments.

I said "component-level," not "Deployment," although I think you may have quite reasonably inferred that I had meant "Deployment," because the existing annotations support does indeed apply annotations to the Deployments and _not_ the pods -- a fact I had overlooked _because the comments in the `values.yaml` file misstates= what `global.annotations` and `<component>.annotations` do._ So that is clearly an existing bug. 🤦‍♂️

> use podLabels and podAnnotations to clarify that the labels are intended for the Pod resources.

Given the realization that there is currently a conflict between existing behavior and existing documentation, and given that the naming conventions you suggest would go a long way toward resolving future confusion, I am inclined to agree, despite it being a breaking change _and_ suggest we ultimately land on something like this:

* `global.labels`: Applied to Deployments, CronJobs, etc., amendable/overridable at the component level with `<component>.labels`.
* `global.podLabels`: Applied to templated pod spec of each Deployment, CronJob, etc., amendable/overridable at the component level with `<component>.podLabels`.
* `global.annotations`: Applied to Deployments, CronJobs, etc., amendable/overridable at the component level with `<component>.annotations`.
* `global.podAnnotations`: Applied to templated pod spec of each Deployment, CronJob, etc., amendable/overridable at the component level with `<component>.podAnnotations`.

charts/kargo/Chart.yaml
krancour: If you follow [this](https://kargo.akuity.io/contributor-guide/hacking-on-kargo#iterating-quickly) the chart should just work as is. It also makes it fast and seamless to rebuild and redeploy components as you hack away.

gnadaban: I'll definitely check it out when I have the time, we also use Tilt, though your Tiltfile looks a lot more elaborate and I'll be sure to learn a thing or two! :)

That said, right now my workflow involves getting the deploy working via ArgoCD along with all the configs and manifests, so I must rely on pulling the chart from my fork for the time being.

charts/kargo/values.yaml
gnadaban: I've implemented support for podLabels and podAnnotations with globals/component-level overrides for workload resources, this should be a good-enough improvement for this PR.

I realize that there might be a need for a more significant overhaul if these should extend to _all_ resources in the chart, but I think that's outside the scope of the fix this is intended to be.

gnadaban: I understand this won't make it into a formal release, but we only need this to be merged in Git to be able to use the chart from source. It would mean a lot and we could use it as a baseline at least, even if there are other things to implement/change/update.

hiddeco (approved):
This looks good to me, and appears to check all the boxes of the nits @krancour had.

Thank you for your contribution @gnadaban 🙇

gnadaban:
> This looks good to me, and appears to check all the boxes of the nits @krancour had.
> 
> Thank you for your contribution @gnadaban 🙇

Thanks @hiddeco ! It appears that @krancour 's review is required before this can be merged. Can I leave this in your hands?

krancour:
If @hiddeco is happy, so am I.

krancour (approved):

hiddeco:
Just briefly checking why CI isn't happy with the codegen, as I can't spot it from the diff.

PR #2114, opened 2024-06-04
chore: remove fig integration

krancour:
We never fully got this to work and now Fig is going away 😢 

This will at least remove a failing job from the release process.


rbreeze (approved):
😢

PR #2118, opened 2024-06-05
feat(webhooks): validate Freight during creation

### LINKED ISSUE #2117: Freight creation validating webhook
# Proposed Feature

Today we have a validation webhook that makes Freight immutable. However, it is still possible that the Create Resource API allows someone to create Frieight that:
* may not include an item from each Warehouse subscription
* may include arbitrary images git commits not in the repositories referenced in the Warehouse

Kargo should improve the validation webhook to ensure the above scenarios are rejected during freight creation.

# Motivation

This is a security feature to ensure only things that are intended to be created by the Warehouse are allowed.

# Suggested Implementation

Add checks in validation webhook. Note that we only need to validate create, since we already have a webhook that makes Freight immutable.

hiddeco:
This adds more validation logic for the creation of Freight, ensuring that Freight contains precisely one artifact for each subscription of the Warehouse it references.

Fixes #2117

### DIFF
--- a/internal/webhook/freight/webhook.go
+++ b/internal/webhook/freight/webhook.go
@@ -3,6 +3,7 @@ package freight
 import (
 	"context"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/technosophos/moniker"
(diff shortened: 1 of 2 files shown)


internal/webhook/freight/webhook.go
```diff
@@ -300,3 +332,171 @@ func (w *webhook) recordFreightApprovedEvent(
...
+
+func (a artifactType) FreightPath() string {
+	switch {
+	case a == artifactTypeGit:
+		return "commits"
+	case a == artifactTypeImage:
+		return "images"
+	case a == artifactTypeChart:
+		return "charts"
+	}
```
krancour, line 346:
```suggestion
	switch a {
	case artifactTypeGit:
		return "commits"
	case artifactTypeImage:
		return "images"
	case artifactTypeChart:
		return "charts"
	}
```

krancour (approved):

internal/webhook/freight/webhook.go
krancour: Non-blocking

PR #2130, opened 2024-06-07
docs: 0.7 release chores

krancour:
Undoes #2113 and #2103 so edge docs once again, implicitly, reference latest available binaries and charts.


krancour:
Unsure how that got there. Nice catch. Let's please remove that from main.

Issue #2140, opened 2024-06-11
bug: includePaths breaks discovery if it reaches the root commit


krancour:
```
error discovering artifacts: error discovering commits: error listing commits from git repo "https://github.com/krancour/kargo-demo-gitops-2.git": error getting diff paths for commit "5e48a597589d86abb2e46db8753468209ef33528" in git repo "https://github.com/krancour/kargo-demo-gitops-2.git": error getting diffs for commit "5e48a597589d86abb2e46db8753468209ef33528": error executing cmd [/usr/bin/git diff --name-only 5e48a597589d86abb2e46db8753468209ef33528^ 5e48a597589d86abb2e46db8753468209ef33528]: fatal: ambiguous argument '5e48a597589d86abb2e46db8753468209ef33528^': unknown revision or path not in the working tree.
```

This occurs because `5e48a597589d86abb2e46db8753468209ef33528` is the root commit, which makes `5e48a597589d86abb2e46db8753468209ef33528^` invalid.

hiddeco:
Think we could potentially solve this by resorting to using `git show --pretty="" --name-only <commit ID>`.

PR #2141, opened 2024-06-11
fix: allow obtaining diff paths for root commit

hiddeco:
The previous approach was inherited from when there was a need to be able to obtain the diff paths for a range of commits. With the refactoring in v0.7.0, however, this is no longer a requirement.

Because of this, we can switch to `git show` to obtain the diff paths for a specific commit. Using `--name-only` to print the paths, while further supplying `--pretty=""` to omit any other commit (message) data. Effectively yielding the same output as the previous command, except that it now also works for root commits.

Fixes: #2140

### DIFF
--- a/internal/controller/git/git.go
+++ b/internal/controller/git/git.go
@@ -409,7 +409,7 @@ func (r *repo) HasDiffs() (bool, error) {
 }
 
 func (r *repo) GetDiffPathsForCommitID(commitID string) ([]string, error) {
-	resBytes, err := libExec.Exec(r.buildGitCommand("show", "--pretty=\"\"", "--name-only", commitID))
+	resBytes, err := libExec.Exec(r.buildGitCommand("show", `--pretty=""`, "--name-only", commitID))
 	if err != nil {
 		return nil, fmt.Errorf("error getting diff paths for commit %q: %w", commitID, err)
 	}
--- a/internal/controller/git/git.go
+++ b/internal/controller/git/git.go
@@ -409,7 +409,7 @@ func (r *repo) HasDiffs() (bool, error) {
 }
 
 func (r *repo) GetDiffPathsForCommitID(commitID string) ([]string, error) {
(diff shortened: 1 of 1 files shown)


krancour:
> time="2024-06-11T16:21:54Z" level=error msg="Reconciler error" Warehouse="{test demo}" controller=warehouse controllerGroup=kargo.akuity.io controllerKind=Warehouse error="error discovering artifacts: error discovering commits: error listing commits from git repo \"[https://github.com/krancour/kargo-demo-gitops-2.git\":](https://github.com/krancour/kargo-demo-gitops-2.git/%22:) error getting diff paths for commit \"5e48a597589d86abb2e46db8753468209ef33528\" in git repo \"[https://github.com/krancour/kargo-demo-gitops-2.git\":](https://github.com/krancour/kargo-demo-gitops-2.git/%22:) error getting diff paths for commit \"5e48a597589d86abb2e46db8753468209ef33528\": error executing cmd [/usr/bin/git show --pretty=\"\" --name-only 5e48a597589d86abb2e46db8753468209ef33528]: fatal: invalid --pretty format: \"\"\n" name=test namespace=demo reconcileID="\"5a57ce2e-cbbd-43fc-acab-721b858a0d46\""

krancour:
The git binary on the image is too old. It doesn't have the `--pretty` option on `git show.`

krancour:
https://github.com/akuity/kargo-render/pull/289

hiddeco:
Think the error is not related to this, but to a newline which _somehow_ gets injected into the empty format string. (Updating Alpine is nonetheless a great idea.)

Give me a bit to figure out what the culprit is.

krancour:
> The git binary on the image is too old. It doesn't have the --pretty option on git show

I stand corrected on this matter. I tried it again and it's fine. idk what I was looking at before. 🤷‍♂️

krancour:
Can confirm https://github.com/akuity/kargo/pull/2141/commits/ea3a8f8709f476c5c3162e00f6aeb07a2a7bac79 does the trick.

krancour (approved):

PR #2145, opened 2024-06-11
fix(controller): support parsing github enterprise urls

### LINKED ISSUE #2080: bug?: github prs may not be working for github enterprise
I know @jessesuen had intended for PRs to GitHub Enterprise to work, as evidenced by the predicate used when registering the GitHub implementation of the `GitProviderService` interface:

https://github.com/akuity/kargo/blob/2d4aaf579625fd723ea4b3941c226ec9bea23d5e/internal/gitprovider/github/github.go#L22-L32

However, this function is used several places throughout that implementation and appears that an error will result for any URL not beginning with `https://github.com`:

https://github.com/akuity/kargo/blob/2d4aaf579625fd723ea4b3941c226ec9bea23d5e/internal/gitprovider/github/github.go#L149-L156

krancour:
Fixes #2080

### DIFF
--- a/internal/gitprovider/github/github.go
+++ b/internal/gitprovider/github/github.go
@@ -154,7 +154,7 @@ func parseGitHubURL(repoURL string) (string, string, error) {
 	path = strings.TrimSuffix(path, ".git")
 	parts := strings.Split(path, "/")
 	if len(parts) != 2 {
-		return "", "", fmt.Errorf("could not parse github repository URL %q", u)
+		return "", "", fmt.Errorf("could not extract repository owner and name from URL %q", u)
 	}
 	return parts[0], parts[1], nil
 }
--- a/internal/gitprovider/github/github.go
+++ b/internal/gitprovider/github/github.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	"regexp"
 	"strings"
 
 	"github.com/google/go-github/v56/github"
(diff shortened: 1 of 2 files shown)


internal/gitprovider/github/github.go
```diff
@@ -146,13 +145,18 @@ func convertGithubPR(ghPR *github.PullRequest) *gitprovider.PullRequest {
...
+func parseGitHubURL(repoURL string) (string, string, error) {
+	u, err := url.Parse(repoURL)
+	if err != nil {
+		return "", "", fmt.Errorf("error parsing github repository URL %q: %w", u, err)
+	}
+	path := strings.TrimPrefix(u.Path, "/")
+	path = strings.TrimSuffix(path, ".git")
+	parts := strings.Split(path, "/")
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("could not parse github repository URL %q", u)
```
hiddeco, line 157:
```suggestion
		return "", "", fmt.Errorf("could not extract repository owner and name from URL %q", u)
```

maybe?

hiddeco (approved):
The suggestion is non-blocking.

hiddeco:
Feels like this could potentially be easily backported to 0.7.1?

krancour:
> Feels like this could potentially be easily backported to 0.7.1?

I wouldn't have done a v0.7.1 just for this, but since we're doing one for other reasons, I'm also inclined to include it.

internal/gitprovider/github/github.go
krancour: Yes. I like this wording better. Thanks!

PR #2139, opened 2024-06-11
feat: support github apps as auth method

krancour:
Fixes #759

PR #2149, opened 2024-06-11
docs: DRY things up a bit by swizzling

krancour:
In the course of documenting #2139 and fixing https://github.com/akuityio/docs.akuity.io/issues/491 / https://github.com/akuityio/docs.akuity.io/pull/493 I learned how to [swizzle](https://docusaurus.io/docs/swizzling) Docusaurus 3 correctly.

For reference:

```shell
npm run swizzle @docusaurus/theme-classic MDXComponents -- --eject
```

This change makes `Tabs`, `TabItem`, and `Hlt` available globally, even in files that use the `.md` extension and _not_ `.mdx`.

There's really no point to this other than that it will save effort on future doc additions.

### DIFF
--- a/docs/STYLE_GUIDE.md
+++ b/docs/STYLE_GUIDE.md
@@ -329,8 +329,6 @@ API.
 In these cases, please make use of _tabs_ to prevent mutually exclusive options
 from taking up space and becoming a "wall of text."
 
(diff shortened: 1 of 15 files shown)


hiddeco (approved):
I'm not too fond of the term they picked for this, but I like the outcome.

Issue #1680, opened 2024-03-25
Multiple pipelines


jessesuen:
# Proposed Feature

When talking to various users about kargo use cases, one limitation that became apparent was that Kargo's single freightline could be limiting since it forces all changes to happen as long as they are in the tip. In our git + image repository example, we use a single Warehouse that subscribes to both the git repo and the image repo, then permutates the combination. One feedback was that image promotion may happen at a separate cadence from other promotable configurations, e.g., a feature flag.

Container images are typically the most common and fastest thing to push through to environments. However, a feature flag or manifest change could be much slower and roll out over the course of days or weeks.

Kargo could solve this by allowing the use of multiple freightlines. For example, a user could configure:
* one warehouse purely for image tags
* another warehouse purely for manifest changes, or `feature-flag-values.yaml`

# Motivation

Allowing for multiple freightlines would allow different categories configurations to be promoted faster than others.

# Suggested Implementation

When thinking about this, the warehouse resource seems appropriate as the originator of each Freightline. So there would be a 1:1 relationship between a Warehouse and a Freightline.

The biggest change I imagine, would be to the Kargo Stage spec/status and controller, which would have to be reworked quite extensively to support this.

jessesuen:
Probably depends on https://github.com/akuity/kargo/issues/1613

krancour:
@jessesuen and I had some offline discussion about this yesterday and I just want to capture key snippets of what we discussed and where we landed.

We agreed on three use cases that we care about WRT multiple freightlines:

1. > Kent: implementing a short pipeline that subscribes to new images and updates main and nothing else -- then a longer one that subscribes only to git and picks up where the previous left off.

    Filling in some unspoken blanks here, this means two warehouses, each subscribed to _one_ thing, and each the source of Freight for a distinct Freightline.

1. > Kent: multiple parallel pipelines that are, organizationally, part of the same project, but involve entirely different sets of artifacts

    > for a large service, or even maybe the microservices use case, I can see this being useful... I can see users asking for a project containing more than one pipeline (because of they want to share visibility within same project, or simplify tenancy mangement).

1. > From this issue: Container images are typically the most common and fastest thing to push through to environments. However, a feature flag or manifest change could be much slower and roll out over the course of days or weeks.

    > The biggest change I imagine, would be to the Kargo Stage spec/status and controller, which would have to be reworked quite extensively to support this.
    
[...]

PR #2154, opened 2024-06-12
feat: beginning of support for multiple pipelines/freight per stage

krancour:
Preliminary work on #1680 

Note this isn't headed for main, but for a feature branch @hiddeco will be asynchronously pairing for a bit.

@hiddeco, quick brain dump of what I've gotten done so far:

I basically started with StageSpec changes -- refactoring the subscriptions into FreightSources. Then I went as far as resolving all ripples those changes created.

I have _not_ done anything to StageStatus yet or Promotions, so Stages don't _truly_ have the ability to host multiple pieces of Freight yet.

I have not tested this end-to-end, but I _think_ that except for the UI, it might work.

I leave it entirely up to you if you want to merge this PR and then open your own PR to continue this work, or simply amend this PR.

Lob this whole issue back my way when you tire of it.


api/v1alpha1/stage_types.go
```diff
@@ -163,23 +164,24 @@ type StageSpec struct {
...
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
-	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+	// +kubebuilder:validation:Required
+	Type string `json:"type" protobuf:"bytes,3,opt,name=type"`
+	// WarehouseDirect indicates whether Freight of the desired type may be
+	// obtained directly from its Warehouse of origin. If this field's value is
+	// false, then the value of the UpstreamStages field must be non-empty.
+	// i.e. Between the two fields, at least on source must be specified.
+	WarehouseDirect bool `json:"warehouseDirect,omitempty" protobuf:"varint,1,opt,name=warehouseDirect"`
```
hiddeco, line 179:
```suggestion
	// Origin specifies the origin of Freight to obtain (i.e. specifies what Warehouse
	// the Freight must have originated from). This field is required.
	//
	// +kubebuilder:validation:Required
	Origin string `json:"origin" protobuf:"bytes,3,opt,name=origin"`

	// Direct indicates whether Freight of the desired origin may be
	// obtained directly from its Warehouse of origin. If this field's value is
	// false, then the value of the UpstreamStages field must be non-empty.
	// i.e. Between the two fields, at least one source must be specified.
	Direct bool `json:"direct,omitempty" protobuf:"varint,1,opt,name=direct"`
```

I still feel something like this would be less ambiguous, as "type" could also refer to a "source type". Another, more explicit and descriptive field name, could potentially be `Warehouse`.

krancour: Please feel free to change it. I don't consider _anything_ in this PR to be set in stone.

internal/controller/stages/stages.go
```diff
@@ -884,15 +884,10 @@ func (r *reconciler) syncNormalStage(
...
+	// in descending order.
 	slices.SortFunc(availableFreight, func(lhs, rhs kargoapi.Freight) int {
-		if lhs.CreationTimestamp.Time.Before(rhs.CreationTimestamp.Time) {
-			return 1
-		}
-		if lhs.CreationTimestamp.Time.After(rhs.CreationTimestamp.Time) {
-			return -1
-		}
-		return 0
+		return rhs.CreationTimestamp.Time.Compare(lhs.CreationTimestamp.Time)
```
krancour, line 890: I was looking for this `.Compare()` func yesterday and idk why I didn't find it. Thanks.

👍

krancour:
@hiddeco I think this should be in a re-reviewable state.

hiddeco (approved):
Happy with the progress thus far and the foundation of API changes. Anything else we discover along the way can still be addressed later.

Nice work @krancour 💯

PR #2156, opened 2024-06-13
feat(controller): SSH support for Git repositories in Warehouses

gnadaban:
Changes in PR:
- Add missing user entry that SSH complained about in Dockerfile
- Update Warehouse CRD to allow specifying Git repositories with git@ prefix without https?://
- Modify `git.buildGitCommand` to use `GIT_SSH_COMMAND` in order to use the repo-specific SSH config file if necessary

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,9 +41,9 @@ ARG GIT_COMMIT
 ARG GIT_TREE_STATE
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
-    -ldflags "-w -X ${VERSION_PACKAGE}.version=${VERSION} -X ${VERSION_PACKAGE}.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ') -X ${VERSION_PACKAGE}.gitCommit=${GIT_COMMIT} -X ${VERSION_PACKAGE}.gitTreeState=${GIT_TREE_STATE}" \
-    -o bin/kargo \
-    ./cmd/controlplane \
+      -ldflags "-w -X ${VERSION_PACKAGE}.version=${VERSION} -X ${VERSION_PACKAGE}.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ') -X ${VERSION_PACKAGE}.gitCommit=${GIT_COMMIT} -X ${VERSION_PACKAGE}.gitTreeState=${GIT_TREE_STATE}" \
+      -o bin/kargo \
+      ./cmd/controlplane \
     && bin/kargo version
 
 WORKDIR /kargo/bin
@@ -94,6 +94,7 @@ USER root
 
 COPY bin/controlplane/kargo /usr/local/bin/kargo
 
+RUN adduser -D -H -u 1000 kargo
 USER 1000:0
 
 CMD ["/usr/local/bin/kargo"]
@@ -130,7 +131,7 @@ USER root
 COPY --from=back-end-builder /kargo/bin/ /usr/local/bin/
 COPY --from=tools /tools/ /usr/local/bin/
 
-RUN adduser -D kargo -u 1000
-USER kargo:0
+RUN adduser -D -H -u 1000 kargo
+USER 1000:0
 
 CMD ["/usr/local/bin/kargo"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,9 +41,9 @@ ARG GIT_COMMIT
 ARG GIT_TREE_STATE
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
-      -ldflags "-w -X ${VERSION_PACKAGE}.version=${VERSION} -X ${VERSION_PACKAGE}.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ') -X ${VERSION_PACKAGE}.gitCommit=${GIT_COMMIT} -X ${VERSION_PACKAGE}.gitTreeState=${GIT_TREE_STATE}" \
-      -o bin/kargo \
-      ./cmd/controlplane \
+    -ldflags "-w -X ${VERSION_PACKAGE}.version=${VERSION} -X ${VERSION_PACKAGE}.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ') -X ${VERSION_PACKAGE}.gitCommit=${GIT_COMMIT} -X ${VERSION_PACKAGE}.gitTreeState=${GIT_TREE_STATE}" \
+    -o bin/kargo \
+    ./cmd/controlplane \
     && bin/kargo version
 
 WORKDIR /kargo/bin
@@ -130,6 +130,7 @@ USER root
 COPY --from=back-end-builder /kargo/bin/ /usr/local/bin/
 COPY --from=tools /tools/ /usr/local/bin/
 
-USER 1000:0
+RUN adduser -D kargo -u 1000
+USER kargo:0
 
 CMD ["/usr/local/bin/kargo"]
--- a/api/v1alpha1/warehouse_types.go
+++ b/api/v1alpha1/warehouse_types.go
@@ -104,7 +104,7 @@ type GitSubscription struct {
 	// URL is the repository's URL. This is a required field.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=`^(https?://)?(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d\w-]+)?(/.*)?$`
+	// +kubebuilder:validation:Pattern=`(?:^(https?)://(?:([\w-]+):(.+)@)?([\w-]+(?:\.[\w-]+)*)(?::(\d{1,5}))?(/.*)$)|(?:^([\w-]+)@([\w+]+(?:\.[\w-]+)*):(/?.*))`
 	RepoURL string `json:"repoURL" protobuf:"bytes,1,opt,name=repoURL"`
 	// CommitSelectionStrategy specifies the rules for how to identify the newest
 	// commit of interest in the repository specified by the RepoURL field. This
@@ -354,7 +354,7 @@ type GitDiscoveryResult struct {
 	// RepoURL is the repository URL of the GitSubscription.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=`^(https?://)?(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d\w-]+)?(/.*)?$`
+	// +kubebuilder:validation:Pattern=`(?:^(https?)://(?:([\w-]+):(.+)@)?([\w-]+(?:\.[\w-]+)*)(?::(\d{1,5}))?(/.*)$)|(?:^([\w-]+)@([\w+]+(?:\.[\w-]+)*):(/?.*))`
 	RepoURL string `json:"repoURL" protobuf:"bytes,1,opt,name=repoURL"`
 	// Commits is a list of commits discovered by the Warehouse for the
 	// GitSubscription. An empty list indicates that the discovery operation was
--- a/api/v1alpha1/warehouse_types.go
+++ b/api/v1alpha1/warehouse_types.go
@@ -104,7 +104,7 @@ type GitSubscription struct {
 	// URL is the repository's URL. This is a required field.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=`^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
+	// +kubebuilder:validation:Pattern=`^(https?://)?(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d\w-]+)?(/.*)?$`
 	RepoURL string `json:"repoURL" protobuf:"bytes,1,opt,name=repoURL"`
 	// CommitSelectionStrategy specifies the rules for how to identify the newest
 	// commit of interest in the repository specified by the RepoURL field. This
@@ -354,7 +354,7 @@ type GitDiscoveryResult struct {
 	// RepoURL is the repository URL of the GitSubscription.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=`^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
+	// +kubebuilder:validation:Pattern=`^(https?://)?(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d\w-]+)?(/.*)?$`
 	RepoURL string `json:"repoURL" protobuf:"bytes,1,opt,name=repoURL"`
 	// Commits is a list of commits discovered by the Warehouse for the
 	// GitSubscription. An empty list indicates that the discovery operation was
--- a/charts/kargo/resources/crds/kargo.akuity.io_warehouses.yaml
+++ b/charts/kargo/resources/crds/kargo.akuity.io_warehouses.yaml
@@ -234,7 +234,7 @@ spec:
                           description: URL is the repository's URL. This is a required
                             field.
                           minLength: 1
-                          pattern: ^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$
+                          pattern: (?:^(https?)://(?:([\w-]+):(.+)@)?([\w-]+(?:\.[\w-]+)*)(?::(\d{1,5}))?(/.*)$)|(?:^([\w-]+)@([\w+]+(?:\.[\w-]+)*):(/?.*))
                           type: string
                         semverConstraint:
                           description: |-
@@ -451,7 +451,7 @@ spec:
                         repoURL:
                           description: RepoURL is the repository URL of the GitSubscription.
                           minLength: 1
-                          pattern: ^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$
+                          pattern: (?:^(https?)://(?:([\w-]+):(.+)@)?([\w-]+(?:\.[\w-]+)*)(?::(\d{1,5}))?(/.*)$)|(?:^([\w-]+)@([\w+]+(?:\.[\w-]+)*):(/?.*))
                           type: string
                       required:
                       - repoURL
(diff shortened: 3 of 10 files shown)


gnadaban:
Can I please get a review on this? :)

charts/kargo/resources/crds/kargo.akuity.io_warehouses.yaml
```diff
@@ -14,541 +14,541 @@ spec:
...
-                  the value of this field. If the shard label is present and this field is
-                  empty, the defaulting webhook will set the value of this field to the value
-                  of the shard label.
-                type: string
-              subscriptions:
-                description: |-
-                  Subscriptions describes sources of artifacts to be included in Freight
-                  produced by this Warehouse.
-                items:
+    - additionalPrinterColumns:
```
hiddeco, line 17: The Custom Resource Definition YAML files should not be modified by hand. They are instead generated based on the Go code (and comments) in `api/`.

gnadaban: Addressed.

gnadaban:
@hiddeco is there anything else to do other than updating the API definition comments?
I assume generation will be done during release?

api/v1alpha1/warehouse_types.go
```diff
@@ -104,7 +104,7 @@ type GitSubscription struct {
 	// URL is the repository's URL. This is a required field.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=`^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
+	// +kubebuilder:validation:Pattern=`^(https?://)?(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d\w-]+)?(/.*)?$`
```
krancour, line 107: After tinkering with it a bit, I believe this will be a more complete regex:

```suggestion
	// +kubebuilder:validation:Pattern=`(?:^(https?)://(?:([\w-]+):(.+)@)?([\w-]+(?:\.[\w-]+)*)(?::(\d{1,5}))?(/.*)$)|(?:^([\w-]+)@([\w+]+(?:\.[\w-]+)*):(/?.*))`
```

https://regex101.com/r/DuAJs7/1

api/v1alpha1/warehouse_types.go
```diff
@@ -354,7 +354,7 @@ type GitDiscoveryResult struct {
 	// RepoURL is the repository URL of the GitSubscription.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=`^https?://(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d]+)?(/.*)?$`
+	// +kubebuilder:validation:Pattern=`^(https?://)?(\w+([\.-]\w+)*@)?\w+([\.-]\w+)*(:[\d\w-]+)?(/.*)?$`
```
krancour, line 357:
```suggestion
	// +kubebuilder:validation:Pattern=`(?:^(https?)://(?:([\w-]+):(.+)@)?([\w-]+(?:\.[\w-]+)*)(?::(\d{1,5}))?(/.*)$)|(?:^([\w-]+)@([\w+]+(?:\.[\w-]+)*):(/?.*))`
```

api/v1alpha1/warehouse_types.go
krancour: Codegen would need to be re-run after this change.

Dockerfile
```diff
@@ -130,6 +130,7 @@ USER root
 COPY --from=back-end-builder /kargo/bin/ /usr/local/bin/
 COPY --from=tools /tools/ /usr/local/bin/
 
-USER 1000:0
+RUN adduser -D kargo -u 1000
+USER kargo:0
```
krancour, line 134:
```suggestion
RUN adduser -D -H -u 1000 kargo
USER 1000:0
```

Addition of `-H` prevents creation of a home directory for this user.

I prefer that we keep `USER 1000:0` as it was. `1000` is the conventional/common non-root user ID for containers in Kubernetes. I think keeping it as `USER 1000:0` makes it more instantly recognizable that we're following that.

This `RUN adduser -D -H -u 1000 kargo` also needs to be added on line 97 or else the rest of these changes aren't going to work in our local development environments.

internal/controller/git/git.go
```diff
@@ -165,6 +165,7 @@ type repo struct {
 	dir                   string
 	currentBranch         string
 	insecureSkipTLSVerify bool
+	useSSH                bool
```
krancour, line 168:
```suggestion
```

We don't need this because it's safe to unconditionally set the `GIT_SSH_COMMAND` env var.

internal/controller/git/git.go
```diff
@@ -242,6 +243,7 @@ func Clone(
 		homeDir:               homeDir,
 		dir:                   filepath.Join(homeDir, "repo"),
 		insecureSkipTLSVerify: cloneOpts.InsecureSkipTLSVerify,
+		useSSH:                clientOpts.Credentials.SSHPrivateKey != "",
```
krancour, line 246:
```suggestion
```

internal/controller/git/git.go
```diff
@@ -758,6 +763,9 @@ func (r *repo) buildCommand(command string, arg ...string) *exec.Cmd {
 
 func (r *repo) buildGitCommand(arg ...string) *exec.Cmd {
 	cmd := r.buildCommand("git", arg...)
+	if r.useSSH {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("GIT_SSH_COMMAND=ssh -F %s/.ssh/config", r.homeDir))
+	}
```
krancour, line 768:
```suggestion
	cmd.Env = append(
		cmd.Env,
		fmt.Sprintf(
			"GIT_SSH_COMMAND=ssh -F %s/.ssh/config -i %s/.ssh/id_rsa",
			r.homeDir, r.homeDir,
		),
	)
```

@gnadaban did you ever see this working end-to-end? In all of my experimentation, this will fail if the location of the keyfile isn't also specified.

Dockerfile
```diff
@@ -41,9 +41,9 @@ ARG GIT_COMMIT
 ARG GIT_TREE_STATE
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
-      -ldflags "-w -X ${VERSION_PACKAGE}.version=${VERSION} -X ${VERSION_PACKAGE}.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ') -X ${VERSION_PACKAGE}.gitCommit=${GIT_COMMIT} -X ${VERSION_PACKAGE}.gitTreeState=${GIT_TREE_STATE}" \
-      -o bin/kargo \
-      ./cmd/controlplane \
+    -ldflags "-w -X ${VERSION_PACKAGE}.version=${VERSION} -X ${VERSION_PACKAGE}.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ') -X ${VERSION_PACKAGE}.gitCommit=${GIT_COMMIT} -X ${VERSION_PACKAGE}.gitTreeState=${GIT_TREE_STATE}" \
+    -o bin/kargo \
+    ./cmd/controlplane \
```
krancour, line 46:
```suggestion
      -ldflags "-w -X ${VERSION_PACKAGE}.version=${VERSION} -X ${VERSION_PACKAGE}.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ') -X ${VERSION_PACKAGE}.gitCommit=${GIT_COMMIT} -X ${VERSION_PACKAGE}.gitTreeState=${GIT_TREE_STATE}" \
      -o bin/kargo \
      ./cmd/controlplane \
```

Just avoiding incidental changes.

internal/controller/git/git.go
```diff
@@ -693,14 +695,17 @@ func (r *repo) setupAuth(creds RepoCredentials) error {
 	// If an SSH key was provided, use that.
 	if creds.SSHPrivateKey != "" {
 		sshConfigPath := filepath.Join(r.homeDir, ".ssh", "config")
+		err := os.MkdirAll(filepath.Dir(sshConfigPath), 0700)
+		if err != nil {
+			return fmt.Errorf("error creating SSH config directory %q: %w", sshConfigPath, err)
+		}
+		rsaKeyPath := filepath.Join(r.homeDir, ".ssh", "id_rsa")
```
krancour, line 702:
```suggestion
		sshPath := filepath.Join(r.homeDir, ".ssh")
		if err := os.Mkdir(sshPath, 0700); err != nil {
			return fmt.Errorf("error creating SSH directory %q: %w", sshPath, err)
		}
		sshConfigPath := filepath.Join(sshPath, "config")
		// nolint: lll
		const sshConfig = "Host *\n  StrictHostKeyChecking no\n  UserKnownHostsFile=/dev/null"
		if err :=
			os.WriteFile(sshConfigPath, []byte(sshConfig), 0600); err != nil {
			return fmt.Errorf("error writing SSH config to %q: %w", sshConfigPath, err)
		}

		rsaKeyPath := filepath.Join(sshPath, "id_rsa")
```

api/v1alpha1/warehouse_types.go
krancour: This regex also needs to be changed on the `RepoURL` field of the `GitRepoUpdate` type.

internal/controller/git/git.go
gnadaban: I actually came up with the solution by making this work in our environment, so I can confirm my suggested solution works.

Dockerfile
gnadaban: Auto-formatting the Dockerfile keeps correcting this, maybe it's worth keeping?

api/v1alpha1/warehouse_types.go
gnadaban: Thanks, I completely missed that.

internal/controller/git/git.go
gnadaban: @krancour , in your suggestion you removed the IdentityFile setting from the SSH config, that is likely the reason.

Was that intentional?

internal/controller/git/git.go
gnadaban: I've incorporated some of your suggestions, but have kept the IdentityFile in the `sshConfig`.
If we're always rendering a distinct SSH config we might as well rely on its contents so that we don't have to also pass in the identity file as an option in `GIT_SSH_COMMAND`.

Thoughts?

internal/controller/git/git.go
krancour: Nope. Oversight on my part -- and this explains it. Thanks.

You can ignore this suggestion!

Dockerfile
krancour: This doesn't matter a _ton_ one way or the other. Consider this suggestion take it or leave it.

I'm curious though what is auto-formatting the Dockerfile that way for you. I use the Docker plugin for VS Code and unless I just have it misconfigured, it doesn't do any auto-formatting for me. It would be a useful thing to have.

internal/controller/git/git.go
krancour: Yep. As mentioned in another comment, removing that piece was accidental.

krancour:
What I am now running into is that promotions that push anything to a git repo are failing. It took me a _while_ to notice what was actually going on...

While a promotion is running, it is re-reconciled periodically, usually to check the status of something external, like a pull request or an Argo CD Application...

It seems _all_ parts of the promotion process re-execute on each reconciliation. Arguably, this was by design. As long as the promotion process is idempotent, then running through all the steps each time ensures that if a previous reconciliation attempt had abruptly failed (due to controller restart, for instance) everything will work itself out.

Unfortunately, this also means the the repo gets cloned many times in fairly rapid succession and GitHub, at least, seems to be rate-limiting ssh connections. So what's actually going on is all the steps of the promotion are succeeding, all the way up to forcing a sync/refresh of an Argo CD Application. On subsequent reconciliation attempts (to follow-up on the status of the Argo CD App) we hit the rate limit when trying to clone, and the whole promotion then errors out, despite being, laregly, complete.

I'm going to open a separate issue to see how we can improve this behavior, but I might want to delay merging this until that is resolved because anyone trying ssh + GitHub is likely to run into this.

Dockerfile
gnadaban: I'm using the same VSCode + Docker extension `ms-azuretools.vscode-docker`, and I have my Code configured to format on save, that's all. I'll leave it as you suggested though, I'm done with the changes. :)

```json
"editor.formatOnSave": true
```

gnadaban:
> It seems all parts of the promotion process re-execute on each reconciliation. Arguably, this was by design. As long as the promotion process is idempotent, then running through all the steps each time ensures that if a previous reconciliation attempt had abruptly failed (due to controller restart, for instance) everything will work itself out.

I found this out the hard way as well, and was something I wanted to bring up but I'm not sure if opening an Issue in Github is the right forum. This was causing me grief in the form of constantly changing a template-time generated credential, which I was only able to get around by switching my testing to a different app that doesn't have one. I'd say some optimization in this area is very much warranted, if for nothing else then due to the cross-regional network cost implications of the traffic it causes.

> I'm going to open a separate issue to see how we can improve this behavior, but I might want to delay merging this until that is resolved because anyone trying ssh + GitHub is likely to run into this.

Please CC me on that issue if you don't mind, I'd like to stay in the loop as it's an important feature we'd benefit from as well.

That said, can I please ask that at least this fix is not held back for a tangential reason? I appreciate your concern, but this fix is solid and would unblock more people than would hinder IMO. The method of cloning from Git does not change the fact of the repeated sync if others are already using the HTTPS auth method, right?

krancour:
@gnadaban https://github.com/akuity/kargo/issues/2178

internal/controller/git/git.go
```diff
@@ -692,15 +692,19 @@ func (r *repo) setupAuthor(author User) error {
 func (r *repo) setupAuth(creds RepoCredentials) error {
 	// If an SSH key was provided, use that.
 	if creds.SSHPrivateKey != "" {
-		sshConfigPath := filepath.Join(r.homeDir, ".ssh", "config")
+		sshPath := filepath.Join(r.homeDir, ".ssh")
+		if err := os.Mkdir(sshPath, 0700); err != nil {
+			return fmt.Errorf("error creating SSH directory %q: %w", sshPath, err)
+		}
+		sshConfigPath := filepath.Join(sshPath, "config")
+		rsaKeyPath := filepath.Join(sshPath, "id_rsa") // nolint: lll
```
krancour, line 700:
```suggestion
		rsaKeyPath := filepath.Join(sshPath, "id_rsa")
```

Dockerfile
krancour: TIL. Thanks for that!

krancour:
I've just come to realize that promotions involving a pull request aren't going to work with ssh creds.

krancour:
> I've just come to realize that promotions involving a pull request aren't going to work with ssh creds.

Thinking about this more, if one really wanted to do ssh + PRs:

1. Kargo would need the ability to use separate creds for interacting with the repo and for interacting with the git provider's API.

2. Users would be required to handle _two_ credentials for updates to _one_ repo.

Since I cannot see the sense in no.2, I cannot see the wisdom of bothering with no.1.

The simplest thing may be to say ssh and pull requests _won't_ work together.

I believe this PR should amend the validating webhook for Stages to return an error in the event that any gitRepoUpdate is configured with an ssh URL _and_ wants to use a pull request.

gnadaban:
I'm not sure I understand why SSH couldn't work with gitRepoUpdate, but I'm willing to look into it, and address in a follow-up PR.

The changes included in this PR so far however unblock our team, and at least partially fixes SSH support for those that don't use Git promotion, only ArgoCD. Can we merge this in, so I can rebase #2160 on this, and I can open a follow-up PR to look into either making gitRepoUpdate +SSH work, or amend the webhook as you suggest.

My concern is primarily the time delay between each roundtrip and that this causes our team to have to maintain a fork and an unofficial build to be able to proceed with testing Kargo.

krancour:
> I'm not sure I understand why SSH couldn't work with gitRepoUpdate

The most recent issue I mentioned is with pull request-based promotions, specifically. Pull requests aren't a git feature. It's a (common) feature of individual platforms like GitHub or GitLab. To open them or follow up on their status involves interacting with those platforms' APIs. Those APIs are usually over HTTPS, so an SSH key isn't a useful credential for authenticating to them.

The more general issue -- not specific to PR-based updates -- is the one I raised in #2160.

> The changes included in this PR so far however unblock our team, and at least partially fixes SSH support for those that don't use Git promotion, only ArgoCD.

I appreciate that merging this unblocks you, but especially given that promotions not involving git are not the norm, using this feature will break things for most other users. We're not going to merge something because it was good enough for one user, while opening the door to problems for most others. Who do you think has to deal with the influx of bug reports?

The compromise I would be willing to make here is that if the regex change to `stage_types.go` is undone (and codegen re-run), it would close the door on git-based promotions using SSH. SSH would work fine in Warehouses and that's it.

This way, at least, the inevitable bug reports will be like, "My warehouse can use an SSH URL, but my promotion mechanism can't," and we'll have to explain that it's not implemented/supported yet. Which is, at least, somewhat easier to deal with than explaining why Kargo let them do something that we knew would fail.

gnadaban:
First of all, thanks for your swift and detailed reply, I really appreciate both!

> The most recent issue I mentioned is with pull request-based promotions, specifically. Pull requests aren't a git feature. It's a (common) feature of individual platforms like GitHub or GitLab. To open them or follow up on their status involves interacting with those platforms' APIs. Those APIs are usually over HTTPS, so an SSH key isn't a useful credential for authenticating to them.
This makes a lot of sense, I didn't look into how git-based promotions work, and just assumed it's simply pushing commits to a repo.

> I appreciate that merging this unblocks you, but especially given that promotions not involving git are not the norm, using this feature will break things for most other users. 
I'm not sure I fully see the scope from my admittedly limited understanding of the project, and what you say makes sense. 

Adding SSH-based Warehouse support without git promotion was my original goal with the PR, admittedly not knowing that the PR based promotion is the mainstream for others (it wouldn't be for us, at least not that I can foresee).

> We're not going to merge something because it was good enough for one user, while opening the door to problems for most others. Who do you think has to deal with the influx of bug reports?
That's fair. I don't mean to add to your burdens, I'm just trying to add a small improvement. It is not my intention to break things for others in any way though, so I greatly appreciate your compromise suggestion.

> The compromise I would be willing to make here is that if the regex change to `stage_types.go` is undone (and codegen re-run), it would close the door on git-based promotions using SSH. SSH would work fine in Warehouses and that's it.

I will go ahead and make this suggested change, and hope that I can repay you with other PRs for helping see this through. :)

krancour:
@gnadaban thank you for understanding and also for your patience and all the hard work on this.

This looks good to me now, but I want to give it a final test-drive, which I should be able to get to before EOW.

krancour (approved):

PR #2157, opened 2024-06-13
fix(promo): notice changes to Argo CD source updates

hiddeco:


internal/controller/argocd/api/v1alpha1/application_types.go
```diff
@@ -29,8 +31,34 @@ type ApplicationSource struct {
...
+// Equals compares two instances of ApplicationSource and returns true if
+// they are equal.
+func (source *ApplicationSource) Equals(other *ApplicationSource) bool {
+	if source == nil && other == nil {
+		return true
+	}
+	if source == nil || other == nil {
+		return false
+	}
+	return reflect.DeepEqual(source.DeepCopy(), other.DeepCopy())
```
krancour, line 43: For my own edification, why the deep copies here?

hiddeco: Good question, and I am not so sure myself, but I directly copied these bits from upstream.

internal/controller/promotion/argocd.go
```diff
@@ -233,76 +275,39 @@ func (a *argoCDMechanism) mustPerformUpdate(
...
-			app.Spec.Sources[i] = source
-		}
-	}
+
+	// Initiate a "hard" refresh.
 	app.ObjectMeta.Annotations[argocd.AnnotationKeyRefresh] = string(argocd.RefreshTypeHard)
+
+	// Update the desired source(s) in the Argo CD Application.
+	app.Spec.Source = desiredSource.DeepCopy()
+	app.Spec.Sources = desiredSources.DeepCopy()
```
krancour, line 308: Just another spot where I am curious what the deep copies accomplished.

krancour (approved):
Very comprehensible and easy to follow.

internal/controller/promotion/argocd.go
hiddeco: This is a habit to ensure the Application fields do not unexpectedly become attached to the desired objects. Which may cause surprising bugs much later on when we suddenly do start to modify those things at some point.

PR #2158, opened 2024-06-14
feat(gitlab): support custom self-managed gitlab URLs

### LINKED ISSUE #2079: Support for Gitlab MR's on Private instances
# Proposed Feature

Support the PullRequest promotion Feature for custom Gitlab deployments.

# Motivation

A lot of enterprises are running their git instances. The Promotion with merge requests is a valuable feature, especially in enterprise use cases, where each change to a production environment has to be validated and approved. The Merge Request Process is often already implemented and can be reused but simplified thanks to Kargo.

# Suggested Implementation

I have no suggestion for the implementation of the logic itself, but in the Stage CR it could be implemented like this:

```
  promotionMechanisms:
    gitRepoUpdates:
    - pullRequest:
        gitlab:
          customUrl: gitlab.enterprise.com
      repoURL: https://gitlab.enterprise.com/playground/kargo-demo-app.git
      writeBranch: stage/prod
```

Brightside56:
This adds support of self-managed Gitlab URLs for pullRequest promotions to resolve #2079. Tested (and working) only with self-managed Gitlab instance for now


Brightside56:
@krancour WDYT?

internal/gitprovider/gitlab/gitlab.go
```diff
@@ -27,8 +27,19 @@ var (
...
+				if err != nil {
+					return nil, err
+				}
+				u.Path = ""
+				u.RawQuery = ""
+				u.Fragment = ""
+				return NewGitLabProvider(u.String() + "/api/v4")
+			} else {
+				return NewGitLabProvider("")
+			}
```
krancour, line 42: I feel things would be a bit tidier if you just passed `repoURL` to `NewGitLabProvider()` and let it do the parsing.

krancour:
Overall, I like the approach. Looks like you've got some compilation errors to address, though.

krancour:
I fell down the rabbit hole...

Your build errors are most easily resolved with a small bit of refactoring of the interface common to both git providers...

Specifically, `WithAuthToken()` was the function whose implementation had forced you to use a `gitlab.Client` directly instead of the `GitLabClient` that was used before. (Because you could extract base URL from a `gitlab.Client`.) It was that swap of `GitLabClient` for `gitlab.Client` that broke the build.

That swap could be undone, restoring things to good working order if only the `WithAuthToken()` function were removed from the `GitProviderService` interface. That is possible if any applicable token were passed directly to each implementation's constructor-like function up front.

At this point, now passing the token up front to both implementations, one begins to wonder why the repo URL that you pass up front for the GitLab implementation isn't _also_ passed for the GitHub implementation. Looking into that, it's clear it _should_ have been -- because guess what the GitHub implementation is still not doing... Still not building a client correctly for GitHub Enterprise when the repo URL is not in public GitHub. 🤦‍♂️

If that were fixed, both implementations can parse the repo URL up front to extract and hang on to important values like owner, repo name, project name, etc. Then repo URL is no longer required to be passed to _all the other functions defined by the `GitProviderService` interface._

Bottom line, I tried to resolve your compilation errors and, without meaning to, knocked over this whole row of dominoes.

If you're open to it, I can cherry-pick an additional commit into this branch that builds directly on top of what you've already done. It simplifies the `GitProviderService` interface and their implementations. That simplification made it possible to fix the build errors. I fixed the broken GH Enterprise support while I was at it. _But I won't amend your PR without your consent._

What I am proposing is here, if you care to look: https://github.com/krancour/kargo/commits/feat/gitlab-private-url/

Brightside56:
> At this point, now passing the token up front to both implementations, one begins to wonder why the repo URL that you pass up front for the GitLab implementation isn't also passed for the GitHub implementation. Looking into that, it's clear it should have been -- because guess what the GitHub implementation is still not doing... Still not building a client correctly for GitHub Enterprise when the repo URL is not in public GitHub

Yes, that's much better. `client.WithEnterpriseURLs` should help there. I've seen it was added into go-github some time ago, but I wasn’t sure that anyone would change anything I could refactor in existing code, because/also I don’t have the possibility to test if such changes are working correctly with GHE

> If you're open to it, I can cherry-pick an additional commit into this branch that builds directly on top of what you've already done. It simplifies the GitProviderService interface and their implementations. That simplification made it possible to fix the build errors. I fixed the broken GH Enterprise support while I was at it. But I won't amend your PR without your consent.

I merged your branch to my request, is that fine? If you need to add change something - please do, I'm ok with it

krancour:
> I don’t have the possibility to test if such changes are working correctly with GHE

I'm going to set up a free trial of GHE to test it.

Are you able to test against a privately hosted GitLab?

Brightside56:
> Are you able to test against a privately hosted GitLab?

I can confirm pull request promotions are working for Gitlab 16 latest (self-managed)

krancour:
@Brightside56 

> I can confirm pull request promotions are working for Gitlab 16 latest (self-managed)

Awesome!

I've been trying to test against GHE. I cut some corners and didn't set up a proper cert for my GHE Server.

This made me realize what we have _both_ missed is that the `insecureSkipTLSVerify` property of a `gitRepoUpdate` needs to make it through to our git provider implementations.

I have a fix for that already in progress and I think it could quite easily just be a follow-up.

This LGTM.

Thank you for the contribution and also for your patience with making sure we did something adequately uniform across both providers.

krancour (approved):

PR #2160, opened 2024-06-14
feat(controller): health checks for multi-source argo cd apps

### LINKED ISSUE #1399: health checks for Applications with multiple sources can probably work now

gnadaban:
Fixes #1399

Changes in PR:
- Fix determinining desired revision when application has multiple sources
- Fix detecting revision for apps with multiple sources targeting the same repository
- Add debug log messages to help diagnose issues
- Add missing fields to prevent destroying application settings in Helm sources such as valueFiles, valuesObject, etc.


gnadaban:
@krancour , this is a WIP but I've made some good progress fixing the issues I found.

internal/argocd/revision.go
```diff
@@ -1,35 +1,81 @@
...
 )
 
 // GetDesiredRevision returns the desired revision for the given
 // v1alpha1.Application by traversing the given v1alpha1.FreightReference for
 // a matching source. If no match is found, an empty string is returned.
 func GetDesiredRevision(app *argocd.Application, freight kargoapi.FreightReference) string {
-	switch {
-	case app == nil || app.Spec.Source == nil:
-		// Without an Application, we can't determine the desired revision.
+	ctx := context.Background()
```
hiddeco, line 18: As this is a utility function, I would not introduce logging at this level but rather ensure this is handled by the caller.

gnadaban: When I was debugging, it was super difficult without this logging to isolate why my app was not being considered for update.

If we follow your suggestion and remove logging here, what does it add to the overall solution in exchange for reducing the optional visibility into the decision chain?

krancour: I don't have a strong opinion about whether logging _belongs_ here or not, but whatever is decided, please use a context that was passed to the function instead of grabbing a new background context. The benefit of that is that if a caller has added details to the logger and stashed that updated logger in the context, this function will inherit that logger with those details.

internal/argocd/revision.go
```diff
@@ -1,35 +1,82 @@
...
 			// Join accounts for the possibility that chart.Name is empty.
 			if path.Join(chart.RepoURL, chart.Name) == sourceChart {
 				return chart.Version
 			}
 		}
-	case app.Spec.Source.RepoURL != "":
+	case s.RepoURL != "":
 		// This source points to a Git repository.
-		sourceGitRepoURL := git.NormalizeURL(app.Spec.Source.RepoURL)
+		sourceLogger.Debug("Application source is a Helm chart, without a chart name", "repoURL", s.RepoURL, "path", s.Path)
```
hiddeco, line 78: Is this true? I was under the impression that at this point it's either a chart from an HTTP repository OR a Git repository revision.

internal/controller/promotion/argocd.go
```diff
@@ -510,12 +537,19 @@ func applyArgoCDSourceUpdate(
...
+	updateLogger := logger.WithValues("source", source, "update", update)
+	updateLogger.Debug("About to apply ArgoCD source update")
+
 	if source.Chart != "" || update.Chart != "" {
 		// Infer that we're dealing with a chart repo. No need to normalize the
 		// repo URL here.
 
 		// Kargo uses the "oci://" prefix, but Argo CD does not.
 		if source.RepoURL != strings.TrimPrefix(update.RepoURL, "oci://") || source.Chart != update.Chart {
+			updateLogger.Debug("Kargo uses oci:// prefix, but Argo CD does not, won't update.")
```
hiddeco, line 552: This is not factually correct. The `TrimPrefix` is to make the comparison of a Kargo value against an Argo CD value works (by ensuring we compare the two URLs _without_ the `oci://` prefix). This failing does, however, not mean what is described in the log message — it simply means the update does not apply to this source.

gnadaban: Ah, that makes sense. Any recommendation as to what this message should be then?

internal/argocd/revision.go
gnadaban: @hiddeco , if you're okay with having debug logs, I'd make the requested updates to this. This would really help users debug their own deployments without having to modify and run the code like I have.

hiddeco: I won't be a blocker here.

internal/argocd/revision.go
hiddeco: Yes, which means the current log message does not take it potentially being a Git repository into account?

gnadaban: My bad, I realize I misunderstood your question, but yes, you can have a Helm chart source with a Path instead of a chart name.

internal/controller/promotion/argocd.go
hiddeco: Something like

> Source update not applicable to Application source, skipping

maybe? But I do wonder if logging this at debug level does not cause immense verbosity, as it is common for the mixed bag of source updates to only apply to a subset. Which means `Trace` might be a better level.

gnadaban:
@hiddeco , any other concerns with this one, other than having to wait until #2156 is merged?

internal/argocd/revision.go
hiddeco: We still are not aligned.

My point is that the log message is incorrect... Any Helm chart scenario is handled by the `case s.Chart != ""`. While at the point where you added the log message, it will start looking for commit references.

It should thus instead be something like:

```suggestion
sourceLogger.Debug("Application source is a Git repository", "repoURL", s.RepoURL, "path", s.Path)
```

gnadaban: That makes sense, updated.

internal/controller/argocd/api/v1alpha1/application_types.go
```diff
@@ -66,7 +67,10 @@ const (
 )
 
 type ApplicationSourceHelm struct {
-	Parameters []HelmParameter `json:"parameters,omitempty"`
+	ReleaseName  string          `json:"releaseName,omitempty"`
+	ValueFiles   []string        `json:"valueFiles,omitempty"`
+	ValuesObject any             `json:"valuesObject,omitempty"`
```
gnadaban, line 72: @hiddeco , `codegen` is complaining about the `any` type of this field. I'm not familiar with the constraints of codegen, what should be used here?

gnadaban:
Is there anything else to address for this one @krancour or @hiddeco ?

hiddeco:
This is on our list to look at, but given the (upstream) complexity it may take a while before we have time to validate this end-to-end. Please bear with us until then, thanks 🙇

hiddeco:
That's quite likely, as I think what you are working on compliments what someone else tried to do before in #2088.

gnadaban:
Added health check fix and tests for multi-source app support.

gnadaban:
Hello @hiddeco / @krancour  , can one of you review please?

gnadaban:
Review please?

krancour:
@gnadaban we are working at a furious pace to get v0.8.0 released and this isn't something we need for v0.8.0, so we need to ask for some patience please. 🙏

gnadaban:
Hey @krancour , thanks for the update!
Any chance to include this in v0.8.0? If there's anything I can do to help to make that happen let me know.

It's literally fixing previously broken/unavailable paths, so I don't think it would bother anyone.

kencochrane:
@gnadaban, we have already frozen 0.8.0 and are only accepting critical bug fixes now. The main goal is to get 0.8.0 out the door as stable as possible, and if we keep adding new items to it, it will require that we start our testing all over again. 

The good news is that we release new versions every five weeks, so your change (if ready) should be included in the next release (0.9.0) and will be available before you know it. Thank you for understanding.

gnadaban:
Hey @kencochrane , thanks for your reply, I was not aware that 0.8 was frozen, and can appreciate wanting to avoid scope creep.

At the same time, I can't help but feel disappointed that this also means we'll have to maintain this PR for potentially weeks more, and wait 5+ to get it officially merged, all this for a trivial addition.

Aside from adding debug log messages, which could arguably be invisible, I added

```go
		if len(app.Status.Sync.Revisions) > 0 {
			// App has multiple sources, so we need to check the list of revisions.
			// Note: Apps with multiple sources pointed at the same Git repository can only have the same revision
			// 		 	 for all sources because ArgoCD does not support the alternative, so we only need to check
			//				the first revision match.
			for _, r := range app.Status.Sync.Revisions {
				if r == revision {
					logger.Debug("Found desired revision in list of revisions of multi-source application, app is healthy.",
						"desiredRevision", revision, "currentRevisions", app.Status.Sync.Revisions)
					return kargoapi.HealthStateHealthy, nil
				}
			}
		}
```

I removed 
```go
	// TODO: We should re-evaluate this soon. It may have been fixed in recent
	//       versions.
	// TODO(hidde): Do we have an upstream reference for this?
	if len(app.Spec.Sources) > 0 {
		err := fmt.Errorf(
			"bugs in Argo CD currently prevent a comprehensive assessment of "+
				"the health of multi-source Application %q in namespace %q",
			key.Name,
			key.Namespace,
		)
		return kargoapi.HealthStateUnknown, healthStatus, syncStatus, err
	}
```

And added
```go
	// An application may have more than one source that points to the same Git repository,
	// eg. a Helm and a vanilla manifest source.
	// In that situation it doesn't matter which source's target revision is returned as ArgoCD does not support
	// different target revisions for sources targeting the same repository.
	for i := range app.Spec.Sources {
		desiredRevision := GetDesiredRevisionFromSource(ctx, &app.Spec.Sources[i], freight)
		if desiredRevision != "" {
			return desiredRevision
		}
	}
``` to ` GetDesiredRevision`.

The most frustrating is that every time any of these functions are changed I have to resolve the conflicts and the clock starts again.

**Comment by kencochrane** [comment]
You shouldn’t have to wait 5 weeks to get this merged. If all goes as planned, 0.8.0 goes out the door this week, and then next week we can merge the PRs that are ready, that didn’t make it into 0.8.0. Then it will be available in the nightly builds.

**Comment by gnadaban** [comment]
I have had to essentially reimplement my fixes due to the extensive changes done to the affected files before merging this.
Please try to review it soon so I don't have to again.

**Comment by gnadaban** [comment]
ArgoCD does not support cloning different revisions of the same repository, hence the single desired revision.

**Comment by gnadaban** [comment]
The current implementation only supports a single source per application to sync/subscribe to, which may arguably be the most common/important use-case where everything from chart to values files are in a single location. This implementation does not support the scenario where manifests/charts/resources of a single ArgoCD app are NOT tracked in a single repository, which IMO would be an anti-pattern.

This PR extends that with the case where one might have 2 or more sources, of which we are still interested in changes to a _single_ one being synced.

For the sake of your example, let's assume there's an app with 2 sources, one from an official chart or Git repository that we don't have control over, and another from some repository that we do. Let's call the one we have control over "internal" and the other one "external".

Normally, I'd think users track target revisions of external sources in the same repository as the app whose repo changes we are interested in, eg. via a Chart.yaml dependencies target revision, or a values or kustomization file with specific target revisions.

In my opinion, we don't care about the external source or what it is synced to, as we don't subscribe to freight changes from that source, we care about the internal repository changes, which might very well have eg. values files or manifests that change the revision of the external source as well.

Because of this, I don't think we should care about checking beyond one Git repository match for a single app.
Either way, this PR already unblocks a large set of users who would benefit from at least this level of multi-source app support.

**Comment by krancour** [comment]
As I understand it, subscribing to an official chart repository for templates and default values and to a git repo for value overrides was the main driver behind multi-source apps. Those umbrella charts were often regarded as a workaround.

You may not agree with that and may see it as an anti-pattern, and I won't argue that point. It may largely be a matter of preference. But I will say, as I've told you before, we are not keen on merging functionality that narrowly addresses your use cases, to the exclusion of the general case.

The general case for multiple source is that for every source whose desired revision could be known, we ought to be checking that that is what it is synced to.

**Comment by gnadaban** [comment]
> As I understand it, subscribing to an official chart repository for templates and default values and to a git repo for value overrides was the main driver behind multi-source apps. Those umbrella charts were often regarded as a workaround.
> 

I'm not sure we're on the same page, let me give you an example and let you decide. This I think would be a typical example of using an upstream chart + an internal repo with the values file for it:
```yaml
kind: Application
metadata:
  labels:
    application-group: aws-controllers-k8s
  name: ack-iam-controller
  namespace: argocd
spec:
  project: myproject
  destination:
    namespace: ack-system
    server: https://kubernetes.default.svc
  sources:
    - chart: iam-chart
      repoURL: public.ecr.aws/aws-controllers-k8s
      targetRevision: 1.3.1
      helm:
        releaseName: ack-iam-controller
        valueFiles:
          - $values/config/default/ack-iam-controller.values.yaml
        valuesObject:
          aws:
            region: us-east-1
    - ref: values
      repoURL: git@mygithubserver.net:myorg/myrepo.git
      targetRevision: 3.4.6
  syncPolicy:
    automated:
      selfHeal: true
    syncOptions:
      - CreateNamespace=true
```

I'm not sure one would call this a workaround, or what you mean by "umbrella chart".

> You may not agree with that and may see it as an anti-pattern, and I won't argue that point. It may largely be a matter of preference. But I will say, as I've told you before, we are not keen on merging functionality that narrowly addresses your use cases to the exclusion of the general case.
> 

I would say this approach of using open-source charts from whatever source with your company's specific configuration is very much a de-facto standard, not some esoteric use-case specific to our example. Think of how many are using eg. `cert-manager`, a very popular open source app, for which you need to include eg. Issuer resources with your own configuration, and perhaps some IRSA annotations for an IAM role that grants access to your DNS records. In such a case, you wouldn't want to subscribe to updates to eg. the `cert-manager` chart itself, because it removes determinism from your infrastructure to blindly give up control over when and what to upgrade to. Instead, one would update the revision in their own cadence in the same Git repo the application's values files are in. This is not only my use-case, and I find it difficult to present it as such.

> The general case for multiple source is that for every source whose desired revision could be known, we ought to be checking that that is what it is synced to.

Can you explain to me through an example why would you have 2 separate Git repository sources you want to subscribe changes for in the same application, given my example above?

gnadaban:
@akolacz , can you provide an example for your use-case, and let me know if what @krancour is advocating for is a a requirement for you, eg. do you have an example app with multiple sources where more than one source is from a different repository you control and would want to subscribe to?

krancour:
@gnadaban 

Apologies for the confusion. I might have misunderstood what you meant by all of the following:

> This implementation does not support the scenario where manifests/charts/resources of a single ArgoCD app are NOT tracked in a single repository, which IMO would be an anti-pattern.
>
> For the sake of your example, let's assume there's an app with 2 sources, one from an official chart or Git repository that we don't have control over, and another from some repository that we do. Let's call the one we have control over "internal" and the other one "external".
>
> Normally, I'd think users track target revisions of external sources in the same repository as the app whose repo changes we are interested in, eg. via a Chart.yaml dependencies target revision, or a values or kustomization file with specific target revisions.
>
> In my opinion, we don't care about the external source or what it is synced to, as we don't subscribe to freight changes from that source, we care about the internal repository changes, which might very well have eg. values files or manifests that change the revision of the external source as well.
>
> Because of this, I don't think we should care about checking beyond one Git repository match for a single app.
Either way, this PR already unblocks a large set of users who would benefit from at least this level of multi-source app support.

Isn't the example you gave, which you say is common (and I agree, it is) an instance of what you claimed to be an anti-pattern? Namely, "where manifests/charts/resources of a single ArgoCD app are NOT tracked in a single repository?"

Let's not quibble about patterns or anti-patterns. Let's just run with this use case.

Why do you not care about verifying that you are synced to the correct revisions of _both_ the chart in `public.ecr.aws/aws-controllers-k8s` _and_ values in `git@mygithubserver.net:myorg/myrepo.git`?

krancour:
Yep. I know we're checking for _general_ problems first and moving on to looking at what everything is synced to only once we get past that.

> In the context of Kargo, we check if a Freight from our internal repo is in the synced sources list of the app, because that's one signal we care about.

So you seem only to care whether you're synced as you expect to the repo containing your values file.

Do you think that everyone has this same expectation? Or do you think a lot of people might also care that the _other_ source is synced as they expect as well?

gnadaban:
> So you seem only to care whether you're synced as you expect to the repo containing your values file.
> 

In the context of subscribing to a single Git Warehouse that represents that Git repository, yes, in addition to also having established that the other source is already synced as verified by having no conditions.

> Do you think that everyone has this same expectation? Or do you think a lot of people might also care that the _other_ source is synced as they expect as well?

I expect that very few people today even have Kargo, and that no one today uses 2 Warehouse subscriptions for the same app, yes.

gnadaban:
@krancour I understand your position of wanting to enforce good, robust solutions, I'm just not sure why this particular PR is expected to address every single edge-case all at once. I'm not against improving things, but given my comparably limited time for this I'd hope you or someone from the maintainers can add/update things as you see fit if necessary, like you've done with multi-pipeline support.

I reached out on Discord in June, I'd be more than happy to do a quick call and discuss and try to find common ground to move things forward.

krancour:
> I expect that very few people today even have Kargo, and that no one today uses 2 Git Warehouse subscriptions for the same app, yes.

I don't understand the relevance of this argument. That's not the case we were using as an example. The example involved one source pointed at a Helm chart repo and one pointed at a Git repo.

> I'm just not sure why this particular PR is expected to address every single edge-case all at once

The case we have been discussing is far from an edge case.

If I were subscribed to a Helm chart repo and a Git repo, and updating two corresponding sources of a multi-source App, I would want to know that _both_ updates had the desired effect. In the general case, if any source were worth updating, why would the user not care that the update worked? If _you_ happen not to care about all but one of the sources being synced correctly, it seems to me that _your case_ may actually be the edge case, with the general case still poorly accounted for.

We were previously honest that we couldn't comprehensively assess the health of Stages with associated multi-source Apps. If we were to merge your change as is, there is a real danger that we begin mis-reporting things as healthy when they are not.

You can either figure accept that @hiddeco and I -- the two most senior maintainers of this project -- share this concern and you can work toward remedying it, or you can just close the PR.

This is the second time we've clashed over a PR that suits _your_ specific use case, but does not generalize well. In the previous case, we worked hard to arrive at a compromise, but please let me be clear that pressuring maintainers into merging PRs they are not comfortable with is poor etiquette in the open source community and if you make a habit of it, you may find that future contributions are not welcome.

With that behind us, why don't we talk about if or how it is possible to check whether all sources with known desired revisions are synced according to expectations?

gnadaban:
@krancour  Allow me to respond slightly out-of-order this time.

> We were previously honest that we couldn't comprehensively assess the health of Stages with associated multi-source Apps. We returned unknown with an explanation. If we were to merge your change as is, there is a real danger that we begin mis-reporting things as healthy when they are not.
> 
> You can either accept that @hiddeco and I -- two senior maintainers of this project -- share this concern and you can work toward remedying it, or you can just close the PR.
> 

I understand both your concerns and I also understand that as you are senior maintainers of this project, you have the say in what goes and what doesn't, and as before, will do what I can to accommodate that.

That said, I, and the community that needs this feature may not be convinced that _not supporting_ multi-source apps for fear of potential misrepresentation of health status is better than supporting it at all.

> This is the second time we've clashed over a PR that suits _your_ specific use case, but does not generalize well. In the previous case, we worked hard to arrive at a compromise, but please let me be clear that pressuring maintainers into merging PRs they are not comfortable with is poor etiquette in the open source community and if you make a habit of it, you may find that future contributions are not welcome.
> 

I did not intend to pressure you into accepting something broken, but while I was trying to make this much-needed feature acceptable I did start to feel like being given the runaround for over a month now.

That said, it would be remiss of me to blatantly disregard etiquette in any situation, and if my sense of urgency was pressuring I owe you both an apology. I completely understand if you don't want to accept my contribution for this, or any reason.

I do not think that this PR is the best implementation, by far. Me and our team would obviously benefit from this, but since there are many other issues with Kargo at this moment eg. #2178 , I was hoping to improve things even a bit, one step at a time, without having to rely on a forked version for long.

Respectfully, we are however a potential customer in the process of trying to evaluate Kargo, and by submitting PRs like this, are actively trying to improve things not just for ourselves and the community, but through this project, Akuity. With that in mind, let me try to speak to the use-case you mentioned, with hopes of finding an acceptable solution to this.

> If I were subscribed to a Helm chart repo and a Git repo, and updating two corresponding sources of a multi-source App, I would want to know that _both_ updates had the desired effect. In the general case, if any source were worth updating, why would the user not care that the update worked? If _you_ happen not to care about all but one of the sources being synced correctly, it seems to me that _your case_ may actually be the edge case, with the general case still poorly accounted for.
> 

If we had an app that has a Helm chart repo and a Git repo, IMO an Application would not be subscribed to both changes at the same time because one would want to control the overall target revision of even the Helm chart via the Git repo that contains the application, and any values files. In my mind at least, a single source of truth for an app is always a single Git repository, regardless of how other sources like Helm repos change.

> With that behind us, why don't we talk about if or how it is possible to check whether all sources with known desired revisions are synced according to expectations?

That said, if we want to support multiple sources being synced, in `health.go` we could iterate on a slice of desired revisions returned from `GetDesiredRevision`(s) if need be, provided that we can match sources we care about to FreightReference instances within that function. What do you think?

krancour:
@gnadaban thank you. Let's work through the issues now. 😄 

> That said, I, and the community that needs this feature may not be convinced that not supporting multi-source apps for fear of potential misrepresentation of health status is better than supporting it at all.

I promise you, we want full support for multi-source apps as badly as anyone. Since the very early days of the project, Argo CD App updates can each involve multiple source updates. Granted, without the corresponding health checks also working, using more than one source update per App update has never been practical, but those updates do work, and I hope that reinforces that this has _always_ been on our radar. It's just those pesky health checks -- and we appreciate your willingness to help with that. With this said, any fear we have of misrepresenting health status isn't merely hypothetical and the notion that misrepresenting health may be preferable to not supporting multi-source at all is probably not shared by all users, especially those who may be contemplating Kargo for the delivery of mission critical applications. Any one or more sources not synced to what you think they should be could be a pretty serious problem.

> I completely understand if you don't want to accept my contribution

We're not there. I promise you that I see us as being back on track now.

> I do not think that this PR is the best implementation

The code review process is just meant to tease out the snarls so this can evolve into the right implementation and I'm confident we can get it there together.

> we could iterate on a slice of desired revisions returned from GetDesiredRevision(s) if need be

This is pretty much what we've been wanting...

> provided that we can match sources we care about to FreightReference instances

This is exactly right. On the _Kargo end of things,_ there's definitely some effort involved, but should be relatively straightforward. The `FreightCollection` at the top of the Stage's `Status. FreightHistory` stack either will or will not contain a reference to a specific revision of an artifact from the repo that the source syncs to. For any source where there's nothing corresponding in the `FreightCollection`, we don't care what that source is synced to. For the others, we do.

It is on the Argo CD end of things where I have been skeptical all along (having nothing to do with your PR) that the information we need is present in the App's `Status`. A failure of mine is that in Feb 2023 I didn't more explicitly document what information was either missing or unreliable that lead me to treat a comprehensive snapshot of health for a multi-source app as indeterminable. I'm digging into Argo CD right now to refresh my own memory and it was `Status.History[n].Revisions`, which is a `[]string`, that I had found perplexing because this doesn't tell you anything about what revision goes with which source. To the best of my recollection (which is probably not super reliable), this field also didn't seem that is was being updated correctly. This was in the Argo CD 2.6 time frame. Multi-source Apps were brand new and far from stable then.

Fast forward to the present day and @hiddeco and I have both refactored parts of the health checks numerous times now, for various reasons, and we're no longer looking at `Status.History[n].Revision` at all. Now we're looking at `Status.OperationState.SyncResult.Revision`. There's a corresponding `Revisions` field that was absent from our "knock-off" `SyncOperationResult` type. Your PR (correctly) adds it. I looked at the docs for the _genuine_ `SyncOperationResult` type just now and found this:

```golang
	// Revisions holds the revision this sync operation was performed for respective indexed source in sources field
	Revisions []string `json:"revisions,omitempty" protobuf:"bytes,5,opt,name=revisions"`
```

So it's parallel to (indexed the same as) `Status.OperationState.SyncResult.Sources`! This is surprising because parallel arrays are usually regarded as rather brittle. idk why they did it this way, but it solves the riddle of how revisions in that array can be correlated to individual sources.

So I am fairly convinced at this point that what you are looking to do is possible and mainly hinges on iterating over `Status.OperationState.SyncResult.Sources` (unless `len(sources) == 0`, in which case we only care about `Status.OperationState.SyncResult.Source`), finding the desired revision _for that source_, and as long as we don't come up empty-handed, comparing that to the corresponding entry in `Status.OperationState.SyncResult.Revisions` (or `Status.OperationState.SyncResult.Revision`).

In the end, I believe this to be a slight refactoring of `GetDesiredRevision(...)` in `internal/argocd/revision.go` to accept a `*kargoapi.ArgoCDSourceUpdate` instead of a `*kargoapi.ArgoCDAppUpdate` and a new loop around the two places where  `GetDesiredRevision(...)` is currently invoked, so that it is invoked for every source in `Status.OperationState.SyncResult.Sources` (unless `len(sources) == 0`, in which case it only needs to be called once for `Status.OperationState.SyncResult.Source`).

I hope these details are what is needed to get this across the line.

gnadaban:
Hey @krancour , thanks for the elaborate writeup!

I looked at the `Status.OperationState.SyncResult.Revisions` and found that `Status.Sync.Revisions` to be more accurate.
In my example, the former is actually an incorrect SHA for some reason, while `Status.Sync.Revisions` holds the correct hash.

The [`Status.Sync` field](https://github.com/argoproj/argo-cd/blob/e36248cb224e4d5697ef8db52bad1865ae3885bd/pkg/apis/application/v1alpha1/types.go#L952) definition confirms that it is indeed the current status, so I'm going to check this in my implementation first.
```go
	// Sync contains information about the application's current sync status
	Sync SyncStatus `json:"sync,omitempty" protobuf:"bytes,2,opt,name=sync"`
```

hiddeco:
Can you provide more information on why that SHA is incorrect, or how it may relate to past and/or other state?

As I feel this may relate to the reflections made by Argo not being truthful, which is why we disabled support in the beginning.

krancour:
> I looked at the Status.OperationState.SyncResult.Revisions and found that Status.Sync.Revisions to be more accurate.
In my example, the former is actually an incorrect SHA for some reason, while Status.Sync.Revisions holds the correct hash.

> Can you provide more information on why that SHA is incorrect, or how it may relate to past and/or other state?
>
> As I feel this may relate to the reflections made by Argo not being truthful, which is why we disabled support in the beginning.

> To the best of my recollection (which is probably not super reliable), this field also didn't seem that is was being updated correctly.

Maybe my recall isn't so bad after all.

gnadaban:
Hey guys, so in my specific case the SHA is incorrect in `Status.OperationState.SyncResult.Revisions` because it's pointing to a commit from eg. May which has since been squashed I believe, whereas the hash in `Status.Sync.Revisions` correctly points to the appropriate hash. I don't think I can give you more than that without giving out proprietary information, but I hope it helps.

I suspect that after ArgoCD does an initial sync, it doesn't actually update that information again in those fields.

gnadaban:
Ok, I've an initial implementation of supporting multiple desired revisions, but I could really use your help with a few things:

1. I don't like the way this block is duplicated in `revision.go` `getRevisionFromSource`:
    ```go
        // If there is a source update that targets app.Spec.Source, it might
        // have its own ideas about the desired revision.
        var targetPromoMechanism any
        for i := range update.SourceUpdates {
          sourceUpdate := &update.SourceUpdates[i]
          if sourceUpdate.RepoURL == s.RepoURL && sourceUpdate.Chart == s.Chart {
            targetPromoMechanism = sourceUpdate
            break
          }
        }
        if targetPromoMechanism == nil {
          targetPromoMechanism = update
        }
        desiredOrigin := freight.GetDesiredOrigin(stage, targetPromoMechanism)
    ```

    If I extract this out of the `switch`, it could be rewritten to only consider RepoURL, which might be enough. Am I missing something here?

2. I refactored `TestGetDesiredRevisions` to allow specifying multi-warehouse test cases, can you sanity-check it pls? I am having trouble figuring out why the one test-case is not returning the correct results, but for the love of me I can't output debug logs when running tests that would help identify the codepath. Any chance you could provide a code-snippet as to how to output log messages for testcases, even temporarily? It'd be nice if this was a flag, but I don't know how much complexity it would add to this PR.

3. Something about how ArgoCD has separate `Revision` and `Revisions` fields makes me want to abstract the whole thing away with some helper functions. Would that be bad? It feels too repetitive to iterate and evaluate multiple times in different code paths. I'm thinking some functions that would eg. return all sources with corresponding Freights and current/desired revisions in a single call, that could then be reused. Thoughts?

krancour:
> I suspect that after ArgoCD does an initial sync, it doesn't actually update that information again in those fields.

Ugh. I wonder if we may need to resort to getting certain information from `Status.OperationState` and other info from `Status.Sync`. 🤦‍♂️ 

> I don't like the way this block is duplicated in...

Ya. At a glance, it looks pretty repetitious, but I think the similarities may be superficial.

> it could be rewritten to only consider RepoURL, which might be enough

For a source pointed at a chart repo, chart name matters. Taking that into consideration, the only true commonality between the two cases is `desiredOrigin := freight.GetDesiredOrigin(stage, targetPromoMechanism)`. The logic for identifying the target promotion mechanism and the logic for finding the commit/chart version remain different across the two cases.

But I am never opposed to refactoring code for better efficiency or readability. I tried once before in this spot and although I was able to avoid the superficial appearance of repetition, the resulting code was a bit more convoluted to actually understand. Invoking "clear over clever," I scrapped that change. You are more than welcome to try. If you come up with something better than I did, I'll be happy.

> I refactored TestGetDesiredRevisions to allow specifying multi-warehouse test cases, can you sanity-check it pls?

Happy to.

> Any chance you could provide a code-snippet as to how to output log messages for testcases, even temporarily?

You're _supposed_ to be able to control the log level with the `LOG_LEVEL` env var, but something is fishy with the initialization. I'm seeing _more_ logs than I expect during testing. Let me get back to you on this.

> Something about how ArgoCD has separate...

Yeah. I think there are several spots now where, because of the old 🤷‍♂️ behavior for multi-source apps, we only ever looked at _singular_ `Source`, `Revision`, etc. If we now have several spots where it's "look at `Sources` if it's not empty, otherwise look at `Souce`, and similar for `Revisions` vs `Revisions`, I absolutely agree about finding an efficient way to keep those conditionals in one place if you can.

krancour:
> You're supposed to be able to control the log level with the LOG_LEVEL env var

This does work as intended.

> but something is fishy with the initialization. I'm seeing more logs than I expect during testing. Let me get back to you on this.

My bad. I failed to notice that the logs I was seeing were error messages and I was using the default LOG_LEVEL of INFO.

I also forgot that successful test results are cached and that changing an env var doesn't invalidate the cache.

I'm using VS Code and have now added this in `.vscode/settings.json` and can control the log level to my satisfaction quite easily:

```json
{
  "go.testFlags": ["-count=1", "-v"], // Disable test result caching, but keep verbose output
  "go.testEnvVars": {
    "LOG_LEVEL": "ERROR" // Change as needed
  }
}
```

The equivalent should be possible for other editors/IDEs as well.

gnadaban:
Ah, marvellous, thanks @krancour !

Now, I took a stab at properly supporting multiple source updates in addition to allowing multiple apps, as well as correctly reporting health status of multi-source apps, let me know what you think.

krancour:
> Now, I took a stab at properly supporting multiple source updates in addition to allowing multiple apps, as well as correctly reporting health status of multi-source apps, let me know what you think.

Sure thing. At this stage I really need to test drive because of that whole "do we trust what Argo CD's telling us argument."

I should be able to get to this tomorrow.

internal/argocd/revision.go
```diff
@@ -2,107 +2,224 @@ package argocd
...
+
+		if s.Chart != "" && sourceUpdate.Chart != s.Chart {
+			sourceLogger.Debug("Source update chart does not match application source chart, skipping",
+				"sourceUpdate", sourceUpdate)
+			continue
+		}
+
+		sourceLogger.Debug("Source update matching application source found", "sourceUpdate", sourceUpdate)
+		targetPromoMechanism = sourceUpdate
+		break
```
krancour, line 209: This logic isn't wrong, but it's a bit difficult to follow. Instead of two cases where you want to continue and then dealing with the case where you found what you were looking for as a default, you could write this as just a single case of what to do when you _have_ found what you're looking for. It's fewer cognitive hoops to jump through to arrive at an understanding of what it's doing.

internal/argocd/revision.go
```diff
@@ -2,107 +2,224 @@ package argocd
...
 		}
 		return commit.ID, nil
 	}
-	// If we end up here, no desired revision was found.
+
+	sourceLogger.Debug("Could not determine desired revision for application from this source.")
 	return "", nil
 }
+
+func getDesiredOriginForSource(
```
krancour, line 179: I get what this function is doing, but its _name_ initially confused me since "desired origin" isn't really a function of an ApplicationSource. It's _purely_ a function of an update. Careful reading of the code reveals that you're using the source to help you find which target update to scrutinize further -- which is correct -- but if the function were, for instance, called `getTargetUpdate()` (and returned the target update), then the name of the function would much more closely match what it does and that would eliminate a lot of confusion.

internal/argocd/revision.go
```diff
@@ -2,107 +2,224 @@ package argocd
...
+	// If there is a source update that targets app.Spec.Source, it might
+	// have its own ideas about the desired revision.
+	desiredOrigin := getDesiredOriginForSource(ctx, stage, update, s)
+
 	switch {
-	case app == nil || app.Spec.Source == nil:
-		// Without an Application, we can't determine the desired revision.
-		return "", nil
-	case app.Spec.Source.Chart != "":
-		// This source points to a Helm chart.
```
krancour, line 35: I would maybe not lose this comment.

internal/argocd/revision.go
```diff
@@ -2,107 +2,224 @@ package argocd
...
+			sourceLogger.Error(err, "Error finding chart from repo")
+			return "", fmt.Errorf("error chart from repo %q: %w", s.RepoURL, err)
 		}
 		if chart == nil {
+			sourceLogger.Debug("Chart not found")
 			return "", nil
 		}
 		return chart.Version, nil
-	case app.Spec.Source.RepoURL != "":
-		// This source points to a Git repository.
```
krancour, line 70: I would maybe not lose this comment.

internal/controller/promotion/argocd.go
```diff
@@ -301,22 +312,46 @@ func (a *argoCDMechanism) mustPerformUpdate(
...
 		return "", true, fmt.Errorf("error determining desired revision: %w", err)
-	} else if desiredRevision != "" && status.SyncResult.Revision != desiredRevision {
-		// The operation did not result in the desired revision being applied.
-		// We should attempt to retry the operation.
+	}
+
+	// Check multi-source app for matching revisions.
+	if len(status.SyncResult.Revisions) > 0 {
+		matchingRevisions := libargocd.GetIntersection(status.SyncResult.Revisions, desiredRevisions)
+		if len(matchingRevisions) == len(desiredRevisions) {
```
krancour, line 332: You're never doing anything with `matchingRevisions` other than checking its length. It is possible the question you are looking to have answered (thought it _may involve_ finding the intersection of observed and desired revisions) is a yes/no of "are all source synced to the desired revision?"

internal/argocd/revision.go
```diff
@@ -2,107 +2,224 @@ package argocd
...
+		hash[e] = true
+	}
+	for _, e := range desired {
+		// If elements present in the hashmap then append intersection list.
+		if hash[e] {
+			intersection = append(intersection, e)
+		}
+	}
+	return intersection
+}
```
krancour, line 31: As noted elsewhere, you're returning an array of strings you never use apart from checking its length.

Apart from that, if you desire foo@x and bar@y and _have_ foo@y and bar@x, this just returns x, y, and the logic in the caller that looks at the length of this to see if it matches the length of desired revisions may end up concluding that everything is synced as desired when in fact, it is not.

internal/argocd/revision.go
```diff
@@ -2,107 +2,224 @@ package argocd
...
+		}
+	}
+
+	// An application may have more than one source that points to the same Git repository,
+	// eg. a Helm and a vanilla manifest source.
+	// In that situation it doesn't matter which source's target revision is returned as ArgoCD does not support
+	// different target revisions for sources targeting the same repository.
+	var revisions []string
+	for i := range app.Spec.Sources {
+		s := &app.Spec.Sources[i]
```
krancour, line 87: It is generally more idiomatic do do this:

```suggestion
	for _, s := range app.Spec.Sources {
```

If you've seen other places in the code where we used the index, there was probably a specific reason for it, which is often that the slice is being modified in place as we iterate. That isn't applicable here, so you can get away with the more idiomatic approach.

This applies in several other spots as well.

internal/argocd/revision.go
gnadaban: My code starts with logging the same, is this necessary?
```go
	case s.Chart != "":
		// NB: This has to go first, as the repository URL can also point to
		//     a Helm repository.
		sourceLogger.Debug("Application source is a Helm chart")
```

internal/argocd/revision.go
gnadaban: I'm not sure that's accurate. 

In `health.go` when checking sources, the single-source is explicitly matched, such that we expect the one revision to be matched:

```go
		singleSourceRevision := app.Status.Sync.Revision
		if singleSourceRevision != "" {
			matchingRevisions := GetIntersection([]string{singleSourceRevision}, revisions)

			if len(matchingRevisions) == 1 {
				return kargoapi.HealthStateHealthy, nil
			}
```

And for multiple sources, for all the revisions to be matched:
```go
		multiSourceRevisions := app.Status.Sync.Revisions
		if len(multiSourceRevisions) > 0 {
			// App has multiple sources, so we need to check the list of revisions.
			// Apps with multiple sources pointed at the same Git repository can only have the same revision
			// for all sources because ArgoCD does not support the alternative, so we only need to check
			// the first revision match.
			matchingRevisions := GetIntersection(multiSourceRevisions, revisions)

			if len(matchingRevisions) > 0 && len(matchingRevisions) == len(revisions) {
				logger.Debug("Found all desired revisions in list of revisions of multi-source application, app is healthy.",
					"desiredRevisions", revisions, "currentRevisions", multiSourceRevisions)
				return kargoapi.HealthStateHealthy, nil
			}
```

In both these cases, we also log what has been compared to what, so that an operator trying to diagnose will find the basis for the decisions in the logs.

Now, in `argocd.go`, this is focused on matching the number of desired revisions as well, which should achieve the same:
```go
	// Check multi-source app for matching revisions.
	if len(status.SyncResult.Revisions) > 0 {
		matchingRevisions := libargocd.GetIntersection(status.SyncResult.Revisions, desiredRevisions)
		if len(matchingRevisions) == len(desiredRevisions) {
			appLogger.Debug("Operation completed, no update necessary.", "phase", status.Phase)
			return status.Phase, false, nil
		}

		return "", true, fmt.Errorf(
			"Not all result revisions out of %v match desired revisions %v",
			status.SyncResult.Revisions, desiredRevisions)
	}

	// Check single-source app for matching revision.
	singleSourceRevision := status.SyncResult.Revision
	if singleSourceRevision != "" && len(desiredRevisions) > 0 {
		matchingRevisions := libargocd.GetIntersection([]string{singleSourceRevision}, desiredRevisions)

		if len(matchingRevisions) != 1 {
			// The operation did not result in the desired revision being applied.
			// We should attempt to retry the operation.
			return "", true, fmt.Errorf(
				"operation result revision %q does not match desired revision %v",
				status.SyncResult.Revision, desiredRevisions,
			)
		}
	}
```

Thoughts?

internal/argocd/revision.go
gnadaban: I found plenty of examples for both so I picked one, but I don't have a preference.

internal/controller/promotion/argocd.go
gnadaban: We're also logging this in eg `health.go` line 303 where it matters, I left a more detailed breakdown in reply to your comment under the function, I'm closing this one to avoid flip-flopping between sections.

internal/argocd/revision.go
gnadaban: Ah, the linter just reminded me why: "Implicit memory aliasing in for loop."

```
internal/argocd/revision.go:85:73: G601: Implicit memory aliasing in for loop. (gosec)
                desiredRevision, err := getRevisionFromSource(ctx, cl, stage, update, &s, frght)
```

I'm going to keep my original unless you have a better alternative.

internal/argocd/revision.go
krancour: I think there's a separate convo to be had about logging. @hiddeco and I both felt the amount of new logging introduced in this PR may be excessive. I don't want to get hung up on that right now though, as that's something that can be discussed later after we get this PR working.

When reviewing this PR, purely to aid my own focus, I reviewed it locally with log-related changes removed to reduce the diffs to just those that were functionally relevant. It was from that myopic POV sans the log message that I found myself wishing this comment had remained.

I agree the log message and the comment are redundant and we only need one or the other. Leave it be for now and we can circle back on this later after the important stuff is dealt with.

internal/argocd/revision.go
krancour: The two string slices passed to `GetIntersection` are _just_ version numbers, lacking any other context.

There's what you have and there's what you want.

Let us presume you _have_ `["1.0.0", "1.1.0"]`. Let us further presume the missing context here is that you have _foo_@1.0.0 and _bar_@1.1.0.

Now say we _desire_ foo@1.1.0 and have no opinion about what bar should be. But again, this is furnished without context to your function. It receives `["1.1.0"]`

The logic in your `GetIntersection` algorithm finds the intersection of `["1.0.0", "1.1.0"]` and `["1.1.0"]`, which is `["1.1.0"]`. By the following logic, you are considering all sources to be at their desired revision, but they are not:

```
if len(matchingRevisions) > 0 && len(matchingRevisions) == len(revisions) {
				logger.Debug("Found all desired revisions in list of revisions of multi-source application, app is healthy.",
					"desiredRevisions", revisions, "currentRevisions", multiSourceRevisions)
				return kargoapi.HealthStateHealthy, nil
			}
```

internal/argocd/revision.go
krancour: In that case, we might typically do this:

```
for _, src := range app.Spec.Sources {
  s := src
```

iirc, Go 1.21+ automatically compensate  for implicit memory aliasing, but the linter hasn't got the memo yet. 😉 

Just leave it be.

internal/argocd/revision.go
gnadaban: I see your point. I will see if I can refactor the solution to more accurately represent desired state.

gnadaban: If we wanted to stay aligned with the ArgoCD implementation, we could align by using a shadow array for the desired revisions where:

- If we have a source update, we would return the version appropriate based on freight
- Let missing chart and similar errors bubble up and ultimately treat the app as Unhealthy
- Return source target revision for sources we don't want to update ie. don't have a targetUpdate for

Thoughts?

krancour: Ya. I think you are on the right track.

I don't _like_ all this parallel array business. It's brittle, but it's what Argo CD has left us to work with.

If I understand your last suggestion correctly, it would be something like this:

1. Iterate over all of the App's sources, finding desired revision for each. If desired revision cannot be determined, the corresponding entry in the list of desired revisions will be the empty string.
2. Iterate over the observed revision of each source, which as I understand it, will be ordered the same as the sourced themselves. (Again... super brittle. 🤦‍♂️) If the corresponding desired revision is either the empty string _or_ matches the observed revision, then we're good. And discrepancy where desired revision wasn't the empty string means the App isn't synced correctly.

Sound right?

gnadaban: Hey @krancour , sorry for the delay. I modified the code to mimic shadow-array use in ArgoCD and added some tests to handle multi-source apps accordingly, let me know what you think.

gnadaban:
Hey @krancour do you have an ETA on when this will be reviewed?

internal/argocd/revision.go
```diff
@@ -10,68 +11,163 @@ import (
...
+) ([]string, error) {
+
+	logger := logging.LoggerFromContext(ctx)
+	if app == nil {
+		err := errors.New("Application is nil")
+		logger.Error(err, "Cannot determine desired revision for application, bailing.")
+		return nil, nil
+	}
+
+	appLogger := logger.WithValues("application", app.Name, "namespace", app.Namespace)
```
hiddeco, line 36: Structured logging only makes sense if the same field names are used consistently. As this field is called `app` elsewhere, it should either be `app` here or `application` in other places.

internal/argocd/revision.go
```diff
@@ -10,68 +11,163 @@ import (
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 	argocd "github.com/akuity/kargo/internal/controller/argocd/api/v1alpha1"
 	"github.com/akuity/kargo/internal/controller/freight"
+	"github.com/akuity/kargo/internal/logging"
 )
 
 // GetDesiredRevision returns the desired revision for the given
 // v1alpha1.Application. If that cannot be determined, an empty string is
 // returned.
```
hiddeco, line 19: This should be updated to match the signature change.

hiddeco (commented):
While the logic appears quite sound now (great job! 🙇), I feel we should trim down the number of logged messages (I count 28 additions, covering 40-something LOC) — and if we log them, ensure they match the style of the logs we already have throughout Kargo.

More concretely:

- No capital letter (except for names, i.e. Application), and no period at the end.
- Present participle form (i.e. `evaluating ArgoCD Application health` instead of `about to ...`)

To trim down the number of logged messages, my suggestion continues to be to let the caller handle most of the logging based on a positive or negative outcome — rather than meticulously logging every detail, which is something I would expect at trace level instead.

internal/controller/argocd/api/v1alpha1/application_types.go
```diff
@@ -29,6 +30,11 @@ type ApplicationSource struct {
 	Helm           *ApplicationSourceHelm      `json:"helm,omitempty"`
 	Kustomize      *ApplicationSourceKustomize `json:"kustomize,omitempty"`
 	Chart          string                      `json:"chart,omitempty"`
+	Path           string                      `json:"path,omitempty"`
+}
+
+func (a Application) IsMultisource() bool {
+	return a.Spec.Sources != nil && len(a.Spec.Sources) > 0
```
krancour, line 37: The nil check is not necessary since len of a var with a nil value is 0.

With the complexity of checking whether an App is multisource reduced to nothing more than a _very straightforward_ check that the len of sources is not 0, I would perhaps not even bother with this helper function, but it's not a hill I would die on.

internal/argocd/health.go
```diff
@@ -217,58 +224,123 @@ func (h *applicationHealth) GetApplicationHealth(
...
-func stageHealthForAppSync(app *argocd.Application, revision string) (kargoapi.HealthState, error) {
+func stageHealthForAppSync(
+	ctx context.Context,
+	app *argocd.Application,
+	revisions []string) (kargoapi.HealthState, error) {
+	logger := logging.LoggerFromContext(ctx).WithValues("appName", app.GetName(), "revisions", revisions)
+	logger.Debug("About to determine stage health based on app sync status.")
 	switch {
-	case revision == "":
+	case revisions == nil || (len(revisions) == 1 && revisions[0] == ""):
```
krancour, line 259: Another spot where the nil check is not needed.

internal/argocd/revision.go
```diff
@@ -10,68 +11,163 @@ import (
...
+		if s == nil {
+			err := errors.New("Single-source application source is nil")
+			appLogger.Error(err, "Cannot determine desired revision for application, bailing.")
+			return nil, nil
+		}
+
+		appLogger.Debug("Application source is not nil, checking.", "source", s)
 
 		// If there is a source update that targets app.Spec.Source, it might
 		// have its own ideas about the desired revision.
```
krancour, line 57: Comment is no longer accurate. Needs an update.

internal/controller/promotion/argocd.go
```diff
@@ -334,38 +345,68 @@ func (a *argoCDMechanism) mustPerformUpdate(
...
+	)
+	appLogger.Debug("Operation completed, checking applied revision.", "desiredRevisions", desiredRevisions)
+
+	if err != nil {
+		appLogger.Error(err, "Failed to determine desired revision, will need to retry.")
 		return "", true, fmt.Errorf("error determining desired revision: %w", err)
-	} else if desiredRevision != "" && status.SyncResult.Revision != desiredRevision {
+	}
+
+	if desiredRevisions == nil {
```
krancour, line 376: It probably covers your bases better here if you check for `len(desiredRevisions) == 0` because that presumes less about the implementation of the `GetDesiredRevisions()` function. You know now that it returns a nil in certain cases, but if someone refactored that function in the future and ever returned an empty, but non-nil slice, your current check here would not be sufficient.

internal/controller/promotion/argocd.go
```diff
@@ -608,17 +649,21 @@ func (a *argoCDMechanism) applyArgoCDSourceUpdate(
...
+	logger := logging.LoggerFromContext(ctx)
+	updateLogger := logger.WithValues("source", source, "update", update)
+	updateLogger.Debug("About to apply ArgoCD source update")
 
-		if source.RepoURL != update.RepoURL || source.Chart != update.Chart {
-			// There's no change to make in this case.
+	if source.Chart != "" || update.Chart != "" {
+		// Kargo uses the "oci://" prefix, but Argo CD does not.
+		if source.RepoURL != strings.TrimPrefix(update.RepoURL, "oci://") || source.Chart != update.Chart {
+			updateLogger.Trace("Source update not applicable to Application source, skipping.")
```
krancour, line 659: This was deliberately not done here before.

For simplicity, the "rule" for the updates is to perfectly match the thing you're applying the update to.

This means that for an app source update, you should match the source. This means you should _not_ use the oci:// prefix in your update.

(For a Helm dependency update, you should match the dependency. This means you _should_ use the oci:// prefix in your update.)

So the only time we really need to account for differences in using/not using oci is when comparing sources to _Freight_.

If there's any feeling that this should be re-evaluated, this isn't the PR to introduce the change.

internal/controller/promotion/argocd.go
```diff
@@ -334,38 +345,68 @@ func (a *argoCDMechanism) mustPerformUpdate(
...
+			!desiredSources.Equals(status.SyncResult.Sources) {
+			// The operation did not result in the desired source(s) being applied.
+			// We should attempt to retry the operation.
+			return "", true, fmt.Errorf(
+				"operation result source does not match desired source",
+			)
+		}
+	}
+
 	// Check if the desired revision was applied.
```
krancour, line 360:
```suggestion
	// Check if the desired revisions were applied.
```

internal/argocd/health.go
krancour: Also this seems to treat _one_ empty string "desired revision" as a special case, but I think the special case should _actually_ be "_all_ desired revisions are empty strings."

It may be that the best way to handle that is when _finding_ the desired revisions. If we come up empty-handed for each and every source, we should probably just return nil/empty slice so that everywhere else we don't need to have the logic that checks the whole collection to see if its all empties.

internal/argocd/revision.go
```diff
@@ -10,68 +11,163 @@ import (
...
-		return "", nil
-	case app.Spec.Source.Chart != "":
-		// This source points to a Helm chart.
-		// NB: This has to go first, as the repository URL can also point to
-		//     a Helm repository.
+
+	// An application may have one or more sources.
+	if !app.IsMultisource() {
+		s := app.Spec.Source
+		if s == nil {
```
krancour, line 48: spec.Source and spec.Sources are mutually exclusive, right?

I think this logic can be simplified...

```
if app.Spec.Source != nil {
  // Zeroes in on single source app and doesn't require a second conditional
  // ...
  // return ...
}
// Everything else false through to here

File: internal/argocd/revision.go [hunk 21]
```diff
@@ -10,68 +11,163 @@ import (
...
 		}
+
+		appLogger.WithValues("source", s).Debug("Could not determine desired revision for application from this source.")
+		return nil, nil
+	}
+
+	// An application may have more than one source that points to the same Git repository,
+	// eg. a Helm and a vanilla manifest source.
+	// In that situation it doesn't matter which source's target revision is returned as ArgoCD does not support
+	// different target revisions for sources targeting the same repository.
```
krancour, line 86: Is this comment relevant? This seems like it might be a remnant an earlier iteration of this PR.

internal/argocd/revision.go
```diff
@@ -10,68 +11,163 @@ import (
...
+			appLogger.Trace("Found desired revision for application source.", "revision", desiredRevision, "source", &s)
+		}
+		revisions[i] = desiredRevision
+	}
+
+	appLogger.Debug("Found desired revisions for multi-source application.", "revisions", revisions)
+	return revisions, nil
+}
+
+func getRevisionFromSource(
```
krancour, line 130: This function name really confused me for a bit as it doesn't seem to accurately describe what it does.

I suggest `getDesiredRevisionForSource()`

internal/argocd/health.go
```diff
@@ -217,58 +224,123 @@ func (h *applicationHealth) GetApplicationHealth(
...
+				app.GetNamespace(),
+			)
+
+			return kargoapi.HealthStateUnhealthy, errors.New(msg)
+		}
+
+		multiSourceRevisions := app.Status.Sync.Revisions
+		// Apps with multiple sources pointed at the same Git repository can only have the same revision
+		// for all sources because ArgoCD does not support the alternative.
+		// We follow ArgoCD shadow-array implementation here that preserves the order of app.spec.sources for
```
krancour, line 297: Shadow array isn't a well-defined or well-understood term. I think it would make it easier to understand if we were more explicit that the list of revisions is indexed the same as the list of sources. "Parallel arrays" also would be well-understood nomenclature.

krancour: Applies in two spots, I think.

internal/argocd/revision.go
```diff
@@ -10,68 +11,163 @@ import (
...
+
+		// If there is a source update that targets app.Spec.Source, it might
+		// have its own ideas about the desired revision.
+		targetUpdate := getTargetUpdate(ctx, update, &s)
+		desiredOrigin := freight.GetDesiredOrigin(ctx, stage, targetUpdate)
+		if desiredOrigin == nil {
+			appLogger.WithValues("source", s,
+				"revision", syncedRevisionOfSource).Debug("Could not determine desired origin" +
+				" for application source, using existing revision.")
+			revisions[i] = syncedRevisionOfSource
```
krancour, line 109: This doesn't look right to me. The previous behavior of this function had been to find the (singular) desired revision for the single source App and it returned an empty string if it couldn't determine / had no opinion about the revision we should be synced to. I didn't expect that to change when this was updated to work with multiple sources/revisions. If we can't figure out what's desired for a particular source, the corresponding element in the array of desired revisions should just be an empty string.

internal/argocd/revision.go
krancour: Actually, considering this further, I don't think you need to special-case single source.

```
sources := app.Spec.Sources
if app.Spec.Source != nil {
  sources = []argocd.ApplicationSource{*app.Spec.Source}
}
// Now loop over all sources; no need for a special case
```

krancour:
> I feel we should trim down the number of logged messages

tbh, I would prefer seeing all the new log lines removed. The signal to noise ratio in this PR is low mostly on account of the logging. To be frank, the noise has been a hindrance in reviewing the relatively few substantive changes in this PR.

internal/controller/freight/origins_test.go
```diff
@@ -214,6 +220,54 @@ func TestGetDesiredOrigin(t *testing.T) {
 				return m, &m.ArgoCDAppUpdates[0]
 			},
 		},
+		{
+			name: "ArgoCDAppUpdate for multi-source app with different origins for various sources are correctly identified",
```
krancour, line 224: I'm not sure this new case adds a lot of value. All the existing cases cover the two crucial behaviors we require when discovering the desired origin for a promotion mechanism -- that child mechanisms can inherit from or override their parents.

internal/argocd/revision_test.go
```diff
@@ -5,30 +5,40 @@ import (
...
 	testOrigin := kargoapi.FreightOrigin{
 		Kind: kargoapi.FreightOriginKindWarehouse,
 		Name: "fake-warehouse",
 	}
+	testOrigin2 := kargoapi.FreightOrigin{
+		Kind: kargoapi.FreightOriginKindWarehouse,
+		Name: "another-warehouse",
+	}
+
+	var no_revisions []string
```
krancour, line 26: Apart from the deviation from standard Go variable naming conventions here, I'm not sure I would declare a variable for this at all. The value of this variable is forever `nil` and all the individual test cases read more cleanly if they just _say_ they expect `nil`.

Nit: That being said, expecting `nil` may even unnecessarily brittle. It works now, but very well may break the moment someone refactors the function under test. The important thing we want to assert in many of these cases is that what is returned is _empty_, which could be `nil` or it could be an non-nil, but still empty slice. In the majority of cases, I find it's useful to have an assertion function as part of each test to perform case-specific assertions. For these cases, you'd assert something like `require.Empty(t, result)`.

internal/argocd/revision_test.go
```diff
@@ -99,39 +275,48 @@ func TestGetDesiredRevision(t *testing.T) {
...
+									Origin: &testOrigin,
+								},
+							},
 						},
-					}},
-				},
+					},
+					Status: kargoapi.StageStatus{
+						FreightHistory: testCase.freightHistory,
+					},
```
krancour, line 315: There is no compelling reason to set this. There's actually a comment in `GetDesiredRevisions` about Freight _not_ being obtained from the status and being obtained instead from the last argument to the function because that grants the flexibility to use that function in the context of a Promotion _or_ a health check of a Stage with its current Freight.

Similarly, all the test cases that explicitly build out a State status with Freight history do not need to.

krancour: Although I see this was a flaw in the original test as well.

internal/argocd/health.go
```diff
@@ -217,58 +224,123 @@ func (h *applicationHealth) GetApplicationHealth(
...
-		cooldown := app.Status.OperationState.FinishedAt.Time.Add(10 * time.Second)
-		if duration := time.Until(cooldown); duration > 0 {
-			time.Sleep(duration)
-
-			// Re-fetch the application to get the latest state.
-			if err := h.argoClient.Get(ctx, key, app); err != nil {
-				err = fmt.Errorf("error finding Argo CD Application %q in namespace %q: %w", key.Name, key.Namespace, err)
-				if client.IgnoreNotFound(err) == nil {
-					err = fmt.Errorf("unable to find Argo CD Application %q in namespace %q", key.Name, key.Namespace)
+		if app.Status.OperationState != nil {
```
krancour, line 227: This new nil-check looks like it probably should have been there all along, however, when you added it, you also added the possibility that one or more desired revisions are known, but we still fall through to the logic at the very end of the function that deems the Application healthy without further conditions.

krancour: > but we still fall through to the logic at the very end of the function that deems the Application healthy without further conditions.

Sorry. I read something wrong. This isn't the case.

@hiddeco knows this part of the code better. What's the right thing to do if we don't have any operation state?

internal/argocd/health.go
```diff
@@ -217,58 +224,123 @@ func (h *applicationHealth) GetApplicationHealth(
...
+			return kargoapi.HealthStateUnhealthy, errors.New(msg)
+		}
+
+		multiSourceRevisions := app.Status.Sync.Revisions
+		// Apps with multiple sources pointed at the same Git repository can only have the same revision
+		// for all sources because ArgoCD does not support the alternative.
+		// We follow ArgoCD shadow-array implementation here that preserves the order of app.spec.sources for
+		// the revisions.
+
+		misaligned_sources := make([]string, 0)
```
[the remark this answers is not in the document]
gnadaban: Good catch. Out of curiosity, how come the linter doesn't catch these?

internal/controller/argocd/api/v1alpha1/application_types.go
```diff
@@ -66,7 +72,13 @@ const (
 )
 
 type ApplicationSourceHelm struct {
-	Parameters []HelmParameter `json:"parameters,omitempty"`
+	ReleaseName string   `json:"releaseName,omitempty"`
+	ValueFiles  []string `json:"valueFiles,omitempty"`
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Type=object
+	ValuesObject json.RawMessage `json:"valuesObject,omitempty"`
```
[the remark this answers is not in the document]
gnadaban: If I remove ValuesObject, would it make it impossible to patch it later on?

internal/argocd/health.go
```diff
@@ -136,6 +148,9 @@ func (h *applicationHealth) GetApplicationHealth(
 		}
```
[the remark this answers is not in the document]
gnadaban: In my case, the information these fine-grained details added was quite necessary when diagnosing what is broken and why. I am convinced that debugging is the right level of setting for such, but I'm not against setting some to trace level. 

That however would go against the expectation people usually have when wanting to see "all" logs. Debug is the most verbose in most popular projects, few have a distinct "trace" setting, so I wouldn't think to look for that.

I think that having detailed logging for such a complex machinery is necessary, and as time goes by and the app matures and evolves it will only become more important.

Please advise.

internal/controller/promotion/argocd.go
gnadaban: Nice catch, thanks! :)

internal/controller/argocd/api/v1alpha1/application_types.go
gnadaban: I've no strong preference at this point, removing.

gnadaban:
@krancour , @hiddeco : I've addressed the majority of the asks, and would like your help to finally complete this.

The debug logging messages added in this PR were essential for _debugging_ what was preventing successful sync operations or causing issues, and to understand what parts of the call chain were involved. Understanding what the app does and how it reacts to things is crucial for cluster admins using Kargo. Since there's no equivalent, I'm strongly against complete removal of detailed logging. I'm not opposed to using Trace level for some, albeit I can't recall if I ever had to use such for any open source project, in my experience "debug" is usually what people use for well, debugging. 

Unless there's something broken I'm hesitant to change or refactor anything else at this point, and would ask to forgo changes to parts of the code that are non-essential to this improvement.

As always, your suggestions are greatly appreciated, and I would again like to ask you to help expedite completion.

internal/argocd/health.go
hiddeco: `stageHealthForAppSync` used to error out if there was no `OperationState`, this is because ArgoCD resets the `OperationState` when a new `Operation` is requested (https://github.com/argoproj/argo-cd/blob/473665795c9bcc40e3a60e3bddb7edf76747c974/util/argo/argo.go#L800-L801).

Given this, the lack of an `OperationState` can indicate that a new (sync) operation is being attempted (or has never happened).

gnadaban: So, keep, remove, what is expected?

krancour:
Superseded by #2552

PR #2161, opened 2024-06-15
refactor(controller): cleanup of the credentials packages

krancour:
After implementing a half dozen credential helpers, the commonalities between them have become more clear, so I have finally given them a common interface now and things get quite tidied up this way.

These changes are entirely structural and should not change any functionality.


internal/credentials/kubernetes/database.go
```diff
@@ -0,0 +1,196 @@
...
+
+	for _, helper := range k.credentialHelpers {
+		creds, err := helper(ctx, namespace, credType, repoURL, secret)
+		if err != nil {
+			return credentials.Credentials{}, false, err
+		}
+		if creds != nil {
+			return *creds, true, nil
+		}
+	}
```
krancour, line 123: This, right here, is the crux of what this whole PR is really about.

internal/credentials/kubernetes/ecr/pod_identity.go
```diff
@@ -84,24 +72,25 @@ func NewPodIdentityCredentialHelper(ctx context.Context) PodIdentityCredentialHe
...
-	if p.awsAccountID == "" {
-		// Don't even try if it looks like EKS Pod Identity isn't set up for this
-		// controller.
-		return "", "", nil
+	credType credentials.Type,
+	repoURL string,
+	_ *corev1.Secret,
+) (*credentials.Credentials, error) {
+	if credType != credentials.TypeImage ||
+		p.awsAccountID == "" { // Pod Identity isn't set up for this controller
```
hiddeco, line 86: If having an AWS account ID is a requirement to have a functional credentials helper. Would it be an idea to not register a helper in the database if these requirements are not met?

hiddeco (approved):
Minor comment, but nothing blocking. Nice job on tidying this 👍

internal/credentials/kubernetes/ecr/pod_identity.go
krancour: I had considered that, but narrowly convinced myself that if the common pattern was the helpers deciding when invoked whether they are applicable then I would stick to that for the sake of uniformity. That's not a strongly held opinion, so I am happy to change it. I think it's a similar situation for the GCP WIF stuff.

krancour: Fixed in https://github.com/akuity/kargo/pull/2161/commits/859334b91991207577a57bec5c5bbaa01721abe0

PTAL, if you don't mind.

hiddeco (approved):

Issue #2162, opened 2024-06-15
bug: failed promotions are no longer showing why they failed


krancour:
The `status.message` field seems as if it is no longer being set.

krancour:
Emphasis on _failed_, which is a different phase from _errored_.

I am going to dig deeper. I think it may be one specific scenario that triggers this.

krancour:
It seems this occurs when the Argo CD promotion mechanism fails because the App had an error syncing.

hiddeco:
Fixed by #2179.

PR #2179, opened 2024-06-19
fix(promo): record failure message for Argo CD Application

hiddeco:
This addresses #2162, which unlike the issue indicates, never actually worked in any Kargo version.

### DIFF
--- a/internal/controller/promotion/argocd.go
+++ b/internal/controller/promotion/argocd.go
@@ -118,6 +118,7 @@ func (a *argoCDMechanism) Promote(
 	logger.Debug("executing Argo CD-based promotion mechanisms")
 
 	var updateResults = make([]argocd.OperationPhase, 0, len(updates))
+	var newStatus = promo.Status.DeepCopy()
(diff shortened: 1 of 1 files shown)


internal/controller/promotion/argocd.go
```diff
@@ -156,6 +157,16 @@ func (a *argoCDMechanism) Promote(
 				logger.Info(err.Error())
 			}
 			if phase.Failed() {
+				// Record the reason for the failure if available.
+				if app.Status.OperationState != nil {
```
hiddeco, line 161: This piggybacks on the fact that we short-circuit. In the initial version, I collected the messages together with the phases. But this grew into something unnecessarily complex and over-engineered, which I would like to avoid as long as we do not need it.

krancour:
> which unlike the issue indicates, never actually worked in any Kargo version.

My apologies on that. I think the memory of seeing error messages surface in the status and the UI fooled me into believing I'd seen the same of failure messages.

krancour (approved):

PR #2186, opened 2024-06-21
feat: limit page requested page size to 100 for list tags requests to quay.io

krancour:
Quay is recently returning a hard error if a page size greater than 100 is requested.


krancour:
Quay has backed out the change that prompted this PR. Closing.

PR #2187, opened 2024-06-21
feat: add pipelines support to Stage Freight history

hiddeco:
This adds the bits to the Stage to support "pipelines". Effectively, this means that instead of `.status.history` maintaining the history of a single piece of Freight. It now maintains the history for `n` pieces of Freight in `.status.freightHistory`.

Example:

```yaml
status:
  # ...omitted for brevity
  freightHistory:
  - freight:
      backend:
        images:
        - digest: sha256:5ab5d4bbe7eae255d8ef5c8c095a72c94f186982fabfff84220e4e41097ef994
          repoURL: ghcr.io/example/backend
          tag: v1.2.3
        name: 9e8659a454c1f37eb69fe81190d54ae0df020400
        warehouse: backend
      frontend:
        images:
        - digest: sha256:d6db3faf3bb7908816cbbf923919aeaf589282bff82bd8b3c4d2979fba6dd23c
          repoURL: ghcr.io/example/frontend
          tag: v3.2.1
        name: 78ccafebe9aa5ffc0061b11e0ca3049e53442aa5
        warehouse: frontend
  - freight:
      backend:
        images:
        - digest: sha256:8cb6b1787140f40914a79dfaff4fa59a805acd54916f267ed13b9c6cb11204a6
          repoURL: ghcr.io/example/backend
          tag: v1.2.0
        name: 350878527223a69ae72326694163405d155e29df
        warehouse: backend
      frontend:
        images:
        - digest: sha256:39347186b34714e4753c4a30c3c317c39522ccb7d6eab59d3187327fe093518a
          repoURL: ghcr.io/example/frontend
          tag: v3.1.9
        name: 06658a103e5e9b8d00b6a00bec02ce4fab2df7b4
        warehouse: frontend
```

**Note:** Both `.status.history` and `.status.currentFreight` have been marked as deprecated, in favor of this newly introduced list's (top item). They do however continue to exist to allow migration logic to be implemented.


api/v1alpha1/stage_types.go
```diff
@@ -567,6 +576,60 @@ type FreightReference struct {
 	VerificationHistory VerificationInfoStack `json:"verificationHistory,omitempty" protobuf:"bytes,7,rep,name=verificationHistory"`
 }
 
+// FreightSelection is a collection of FreightReferences, each of which
+// represents a piece of Freight that has been selected for deployment to a
+// Stage.
+type FreightSelection struct {
+	// Items is a map of FreightReference objects, indexed by their Warehouse
+	// origin.
+	Items map[string]FreightReference `json:"items,omitempty" protobuf:"bytes,1,rep,name=items" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
```
krancour, line 585: This was the compromise required to get this to work with protobufs, right? I am thinking we should add some note about it here.

krancour:
Some thoughts that are certainly open to discussion of course:

"items" is bugging me, because when I see it, I am too quickly/easily fooled into thinking the things _under it_ are individual _history entries_. I am fully aware that's not the case, but needing to continuously mentally adjust for that raised the cognitive load of understanding the history by a little bit.

What would you think about renaming `items` as `freight`?

```yaml
status:
  # ...omitted for brevity
  freightHistory:
  - freight:
      backend:
        images:
        - digest: sha256:5ab5d4bbe7eae255d8ef5c8c095a72c94f186982fabfff84220e4e41097ef994
          repoURL: ghcr.io/example/backend
          tag: v1.2.3
        name: 9e8659a454c1f37eb69fe81190d54ae0df020400
        warehouse: backend
      frontend:
        images:
        - digest: sha256:d6db3faf3bb7908816cbbf923919aeaf589282bff82bd8b3c4d2979fba6dd23c
          repoURL: ghcr.io/example/frontend
          tag: v3.2.1
        name: 78ccafebe9aa5ffc0061b11e0ca3049e53442aa5
        warehouse: frontend
  - freight:
      backend:
        images:
        - digest: sha256:8cb6b1787140f40914a79dfaff4fa59a805acd54916f267ed13b9c6cb11204a6
          repoURL: ghcr.io/example/backend
          tag: v1.2.0
        name: 350878527223a69ae72326694163405d155e29df
        warehouse: backend
      frontend:
        images:
        - digest: sha256:39347186b34714e4753c4a30c3c317c39522ccb7d6eab59d3187327fe093518a
          repoURL: ghcr.io/example/frontend
          tag: v3.1.9
        name: 06658a103e5e9b8d00b6a00bec02ce4fab2df7b4
        warehouse: frontend
```

In a similar vein, what would be your thoughts about renaming the `FreightSelection` type as `FreightHistoryEntry`? This change doesn't affect the YAML representation of the history, but when reaind the code, I think `FreightHistoryEntry` might be more intuitively understood?

Neither of these are hills I would die on.

internal/controller/stages/stages.go
```diff
@@ -647,198 +652,196 @@ func (r *reconciler) syncNormalStage(
 	status.ObservedGeneration = stage.Generation
 	status.Health = nil
 
-	if status.CurrentFreight == nil {
+	if currentFreight := status.FreightHistory.Current(); currentFreight == nil || len(currentFreight.Items) == 0 {
```
krancour, line 655: If currentFreight were non-nil, the length of the stack couldn't possibly be 0.

```suggestion
	if currentFreight := status.FreightHistory.Current(); currentFreight == nil {
```

internal/controller/stages/stages.go
```diff
@@ -647,198 +652,196 @@ func (r *reconciler) syncNormalStage(
...
-		if stage.Spec.Verification != nil {
-			// Update the verification history with the current verification info.
-			// NOTE: We do this regardless of the phase of the verification process
-			// and before potentially creating a new AnalysisRun to ensure we add
-			// any previous verification attempt which may have been recorded
-			// before we started tracking history.
-			if status.CurrentFreight.VerificationInfo != nil {
-				status.CurrentFreight.VerificationHistory.UpdateOrPush(*status.CurrentFreight.VerificationInfo)
+		for _, freight := range currentFreight.Items {
+			freightLogger := logger.WithValues("freight", status.FreightHistory.Current())
```
krancour, line 662: Are these k/v pairs correct? `status.FreightHistory.Current()` is a struct and is probably going to dump a lot of noise into the logs.

internal/controller/stages/stages.go
```diff
@@ -851,6 +854,14 @@ func (r *reconciler) syncNormalStage(
 		return status, nil
 	}
 
+	// TODO: Remove this when we have added support for multiple Freight below.
+	if len(stage.Spec.RequestedFreight) > 1 {
+		logger.Info(
+			"Stage requests multiple Freight. Auto-promotion support has however not been implemented for this (yet).",
+		)
+		return status, nil
+	}
```
krancour, line 863: 👍 This will be fun to figure out. Glad to take it out of the equation temporarily.

internal/controller/stages/stages.go
hiddeco: From a pure Go perspective, it theoretically can if we have a bug elsewhere. This ensures that doesn't cause more problems.

api/v1alpha1/stage_types.go
hiddeco: I thought about this, but opted not to because it would leak into the CRD documentation. Where this level of detail may end up confusing the user.

internal/controller/stages/stages.go
hiddeco: Good catch, this does indeed require changes.

api/v1alpha1/stage_types.go
krancour: Fair enough. I have sometimes wished there were a straightforward way to leave type or field-level comments that _don't_ bleed into those docs.

hiddeco: Commit messages are the answer, but this does not really pair well with a squash and merge strategy.

internal/kubeclient/indexer.go
```diff
@@ -87,18 +87,27 @@ func indexStagesByAnalysisRun(shardName string) client.IndexerFunc {
...
 			return nil
 		}
-		return []string{
-			fmt.Sprintf(
-				"%s:%s",
-				stage.Status.CurrentFreight.VerificationInfo.AnalysisRun.Namespace,
-				stage.Status.CurrentFreight.VerificationInfo.AnalysisRun.Name,
-			),
+
+		var analysisRuns []string
```
hiddeco, line 96: This requires more research, as it is allowed (see https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/client#FieldIndexer) but has indications it may not work nicely in every scenario.

krancour (approved):

PR #2197, opened 2024-06-24
feat(ui): ability to collapse old freight in freightline

rbreeze:
In many cases, Freightline are full of dozens or hundreds of old, unused freight that slows down the UI. This PR adds a toggle to hide all freight older than the oldest freight currently in use.

Screenshot 2024-06-24 at 10 48 32
Screenshot 2024-06-24 at 10 48 22

### DIFF
--- a/ui/src/features/freightline/freight-item.tsx
+++ b/ui/src/features/freightline/freight-item.tsx
@@ -14,7 +14,8 @@ export const FreightItem = ({
   mode,
   empty,
   highlighted,
-  onHover
+  onHover,
+  hideLabel
 }: {
   freight?: Freight;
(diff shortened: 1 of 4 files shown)


krancour:
This is super useful!

krancour:
@rbreeze what is the behavior when _no_ Freight are currently in use? i.e. New Project?

rbreeze:
@krancour In this case, the button would have no effect (it wouldn't hide any freight)... so given that, I think it's best to hide the button when no freight are in use. I'll go ahead and add that behavior

krancour (approved):
Test drove and this is 🔥

PR #2201, opened 2024-06-24
feat(ui): add more information to project list page

rbreeze:
Includes new API endpoint that provides a list of Stages for each Project. 

Screenshot 2024-06-24 at 15 48 01
Screenshot 2024-06-24 at 15 48 13


internal/api/list_detailed_projects_v1alpha1.go
```diff
@@ -0,0 +1,45 @@
...
+func (s *server) ListDetailedProjects(
+	ctx context.Context,
+	_ *connect.Request[svcv1alpha1.ListDetailedProjectsRequest],
+) (*connect.Response[svcv1alpha1.ListDetailedProjectsResponse], error) {
+	var list kargoapi.ProjectList
+	if err := s.client.List(ctx, &list); err != nil {
+		return nil, fmt.Errorf("error listing Projects: %w", err)
+	}
+
+	sort.Slice(list.Items, func(i, j int) bool {
```
hiddeco, line 23: We should start making use of the [standard `slices` package](https://pkg.go.dev/slices) which is faster in many cases. I think there are examples elsewhere in the code base for lexicographic sorting.

krancour:
I'm a little bit worried about the size of the responses from the new endpoint. It's well within the realm of possibility that the response will contain thousands of Stages.

I'd feel better about it if we introduced paging to both the new endpoint and the Projects page -- sorry I know that's a fair bit of scope creep.

If we were to do that, I think it would be tenable to make `pageSize` list Stages requests. (One for each Project on the current page -- something I _wouldn't_ have considered with an unbounded number of Project per page.)

Ultimately, that would remove the need for the new endpoint. The extent of the back-end changes would just be the addition of paging to the existing list Project endpoint.

Tangentially, I think there may be some security considerations to weigh here as well. Regardless of whether we were to make the suggested changes above _or not_, there may very likely be an issue of users who are able to list all Projects being unable to list Stages, which makes all of this a tad trickier.

If we were to swap to the paging approach and using existing endpoints, I think we'd just have to make the UI silently ignore 403s from list Stages requests for any given project.

rbreeze:
@krancour Agreed! Updated this PR accordingly, and added simple string filtering in the process

api/rbac/v1alpha1/role_types.go
```diff
@@ -27,13 +27,13 @@ type RoleResources struct {
 }
 
 type ResourceDetails struct {
-	ResourceType string   `json:"resourceType,omitempty"`
-	ResourceName string   `json:"resourceName,omitempty"`
-	Verbs        []string `json:"verbs,omitempty"`
+	ResourceType string   `json:"resourceType,omitempty" protobuf:"bytes,1,opt,name=resourceType"`
+	ResourceName string   `json:"resourceName,omitempty" protobuf:"bytes,2,opt,name=resourceName"`
+	Verbs        []string `json:"verbs,omitempty" protobuf:"bytes,3,rep,name=verbs"`
```
krancour, line 32: I'm trying to figure out why these struct tags were absent before and have now shown up all of the sudden. 🤔

api/service/v1alpha1/service.proto
```diff
@@ -332,11 +332,14 @@ message GetProjectResponse {
 }
 
 message ListProjectsRequest {
-  /* explicitly empty */
+  optional int32 page_size = 1;
+  optional int32 page = 2;
```
krancour, line 336: Thank you for thinking to make these optional! It preventing breaking the CLI!

internal/api/list_projects_v1alpha1.go
```diff
@@ -13,7 +14,7 @@ import (
 
 func (s *server) ListProjects(
 	ctx context.Context,
-	_ *connect.Request[svcv1alpha1.ListProjectsRequest],
+	req *connect.Request[svcv1alpha1.ListProjectsRequest],
```
krancour, line 17: The test are failing because they're passing a nil request. You'll have to update them to pass a legitimate request now.

api/rbac/v1alpha1/role_types.go
krancour: This is non-blocking, btw.

krancour:
@rbreeze took this for a test drive and works very nicely.

Excellent work on this.

I did notice that the list Stages endpoint isn't returning Stages lexically ordered by name. That's not something new you introduced, but the updated Projects page is making that discrepancy more noticeable. (Surprised no one noticed this when using the CLI.)

Screenshot 2024-07-01 at 11 21 25 PM

Would you mind adding a sort here?

https://github.com/akuity/kargo/blob/be15310310e700d1c25da4b092997cf5a8327ed8/internal/api/list_stages_v1alpha1.go#L32-L38

api/rbac/v1alpha1/role_types.go
rbreeze: I was wondering the same thing. Maybe a library used in the tooling was updated recently, but this is the first PR since which had any codegen?

rbreeze:
@krancour Good idea! Updated ListStage endpoint with a sort.

internal/api/list_stages_v1alpha1.go
```diff
@@ -29,6 +30,10 @@ func (s *server) ListStages(
 		return nil, fmt.Errorf("list stages: %w", err)
 	}
 
+	sort.Slice(list.Items, func(i, j int) bool {
+		return list.Items[i].Name < list.Items[j].Name
+	})
```
krancour, line 35: Since Go 1.21, this is the preferred way to sort.

```suggestion
	slices.SortFunc(list.Items, func(a, b kargoapi.Stage) int {
		return strings.Compare(a.Name, b.Name)
	})
```

Note, we do have a lot of places in the code base where it's still done the older way and I'll open a separate chore issue for fixing those.

rbreeze: Ah thanks for catching that! Updated the ListProject sort too. Sorry for the back and forth

krancour (approved):

PR #2203, opened 2024-06-25
chore: cleanup multi-pipeline work

krancour:
This PR puts back a field and some types that were prematurely removed in #2154 and deprecates them instead.

I also added appropriate `// nolint: staticcheck` comments wherever we are temporarily still doing something with a deprecated field...

Although...

@hiddeco if we assume that the `release-0.8` branch will get some release-specific migration logic, do you think we can just remove any/all logic pertaining to deprecated fields with the assumption that the "normal" reconciliation loop will later be amended to operate only on Stages that have already been migrated? This makes sense to me, but wdyt?


hiddeco (approved):
Your suggestion makes sense, but I would put it in place as a last step.

PR #2204, opened 2024-06-25
feat(ui): convenience page for links to CLI binaries

### LINKED ISSUE #2176: CLI download links from UI
# Proposed Feature

The UI should provide convenience links to download the `kargo` CLI for each architecture that we support from GitHub releases.

# Motivation

A user we were helping did not know where to download the kargo CLI.

# Suggested Implementation

Release links can be formulated from version, os, and architecture. e.g:

https://github.com/akuity/kargo/releases/download/v0.7.1/kargo-darwin-amd64

rbreeze:
Fixes #2176

When in Dev mode, links to latest. Otherwise, links to embedded version

CleanShot 2024-06-25 at 09 51 54@2x


ui/src/pages/downloads.tsx
```diff
@@ -0,0 +1,89 @@
...
+      {__UI_VERSION__ === 'development' ? 'Latest version' : __UI_VERSION__}
+    </div>
+    <a
+      href='https://github.com/akuity/kargo/releases'
+      target='_blank'
+      rel='noreferrer'
+      className='mb-6 flex text-xs items-center text-blue-500'
+    >
+      <FontAwesomeIcon icon={faExternalLink} className='mr-2' />
+      VIEW ALL RELEASES
```
hiddeco, line 60: I have been wondering for some time why these types of things use literal capslock, instead of something like `text-transform: capitalize`.

rbreeze: Good question. The reason to use caps lock vs. text transform is for accessibility: if someone uses a screen reader, for example, it will typically not detect the text transform, but if the text is _actually_ capitalized, it will correctly indicate that the text is all-caps.

hiddeco: I was always under the impression this would do the opposite thing for most screen readers, as it will, for example, read certain bits as acronyms (e.g. "IT" becomes "I.T.").

See also:

- https://www.timdunklee.com/notes/use-css-text-transform-for-uppercase-letters-a11y/
- https://www.stefanjudis.com/today-i-learned/text-transforms-affects-screen-readers-too/

rbreeze: TIL! Thanks for the links!

hiddeco (approved):
LGTM, thanks @rbreeze 🥇

krancour:
This is a very nice improvement. Thanks @rbreeze !

PR #2205, opened 2024-06-25
fix(health): sync cooldown before health check 

### LINKED ISSUE #2196: Kargo should not trust Argo CD health too quickly after operation completes
# Description

Users have noticed that Kargo might immediately verify a freight in a Stage before that application reaches a true Healthy state. 

It turns out, Kargo is making an assumption about Argo CD that is often unsafe. The assumption is that when a sync operation (kubectl apply) completes, the application’s health will **_already_** be in a Progressing state (as a result of the apply that just happened). This happens to be true most of the time, but it’s unsafe for the following reason:

For better or worse, Argo CD’s operation reconciler is a separate reconciler from the health/status reconciler. In other words, Argo CD reconciles the .operation field independently and concurrently from the .status health/sync fields. Even worse, these are calculated at different times.

It works most of the time since after a resource is `kubectl applied`, the app gets re-queued in the status reconciler, and so the app health status often gets updated to Progressing even before the sync operation completes. But there is no guarantee. But because the two reconcilers are separate, what might happen is that the Sync operation completes, and the application is still `Healthy`, causing Kargo to immediately consider the Freight verified prematurely.

To fix this, we will have to do something ugly in Kargo: I propose that we have a grace period (~5-10s) **_after an operation_** is completed before we fully trust an app's health. This grace period gives a high percentage chance that the app will enter a progressing state as a result of the kubectl apply that just happened. It’s similar to an ugly sleep hack in Argo CD that also had to be implemented because we might reconcile health too quickly.

https://github.com/argoproj/argo-cd/issues/4669

This problem can currently be worked around by having a dummy PostSync hook (the resource can even be a dummy `ConfigMap`). The reason this works is because the presence of a PostSync hook does the job of waiting for application resources to become healthy (before the hook fires). In other words, the whole waiting for application to become healthy is handled directly by Argo CD operation reconciler (so kargo doesn't have to have to do anything).

# Steps to Reproduce

This happens intermittently during promotions.

# Version

```
v0.7
```

hiddeco:
Fixes #2196


rbreeze (approved):
LGTM

internal/controller/argocd/api/v1alpha1/application_types.go
```diff
@@ -172,6 +172,7 @@ type OperationState struct {
 	Phase      OperationPhase       `json:"phase,omitempty"`
 	Message    string               `json:"message,omitempty"`
 	SyncResult *SyncOperationResult `json:"syncResult,omitempty"`
+	FinishedAt *metav1.Time         `json:"finishedAt,omitempty"`
```
krancour, line 175: This has been PR'ed upstream as well?

hiddeco: This already exists upstream.

internal/argocd/health.go
```diff
@@ -212,6 +214,19 @@ func stageHealthForAppSync(app *argocd.Application, revision string) (kargoapi.H
...
+			// operations (like health checks) have a chance to run after the
+			// sync operation has finished.
+			//
+			// xref: https://github.com/akuity/kargo/issues/2196
+			//
+			// TODO: revisit this when https://github.com/argoproj/argo-cd/pull/18660
+			// 	 is merged and released.
+			time.Sleep(duration)
+		}
+
```
krancour, line 229: @hiddeco and i discussed offline and there are some changes needed here. We need to reload the App after the grace period and double-check its status -- or else requeue the stage for reconciliation.

krancour (changes requested):
https://github.com/akuity/kargo/pull/2205/files#r1655511913

internal/argocd/health.go
hiddeco: Addressed in d5610e5eeb86e155e02ff6ad6eeddb034a113697

krancour (approved):

PR #2206, opened 2024-06-26
refactor: refactor verification process to be suitable for multiple pipelines

krancour:
The main change here is that verification history is moved up a level from `FreightReference` to `FreightHistoryEntry` such that verification attempts, including any associated `AnalysisRun` is now associated with a specific _combination_ of Freight, rather than individual pieces of Freight.

As has been previously discussed, once such a verification attempt succeeds, every constituent piece of Freight is marked as verified, but not claim is made as to the _combination_ being valid.

Ready for review, ~~but I am still validating the changes e2e.~~ ✅ 

Aside: I believe we should consider renaming `FreightHistoryEntry` as `FreightCombination`. `FreightHistoryStack` would _not_ be renamed, but the entries _in_ the stack would be instances of `FreightCombination`.


api/v1alpha1/stage_types.go
```diff
@@ -611,6 +618,24 @@ type FreightHistoryEntry struct {
...
 	Freight map[string]FreightReference `json:"items,omitempty" protobuf:"bytes,1,rep,name=items" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
+	// VerificationHistory is a stack of recent VerificationInfo. By default,
+	// the last ten VerificationInfo are stored.
+	VerificationHistory VerificationInfoStack `json:"verificationHistory,omitempty" protobuf:"bytes,2,rep,name=verificationHistory"`
+}
+
+// FreightNames returns a sorted and comma-delimited list of the names of the
+// Freight in the FreightHistoryEntry. This is useful for when an identifier is
+// needed to represent a precise combination of Freight.
+func (f *FreightHistoryEntry) FreightNames() string {
```
hiddeco, line 629: Wonder if it may be slightly better for the caller to decide if `strings.Join` should be used, and have this return `[]string`.

Alternatively, a separate method could be added which calls `FreightNames` to present a string identifier then.

hiddeco (commented):
Went over all the further changes and do not have any blockers except for my earlier suggestion. I also do agree that `FreightCollection` might be a more descriptive name than the `...Entry` thing we have now.

api/v1alpha1/stage_types.go
krancour: Yeah... reviewing how widely this _isn't_ used, this could really just be a private func closer to the verification logic.

Will move.

hiddeco (approved):

PR #2209, opened 2024-06-26
feat(ui): display last promotion completion time instead of freight creation time

### LINKED ISSUE #2171: UI: Stage Timestamp shows freight creation date, not freight promotion date.
# Checklist

# Proposed Feature

Right now the Timestamp shows the freight creation date, would it make more sense to show when the promotion occured? 

This came as I tested multiple Stages with auto-promote enabled. But left a Stage lingering in a progressing state for a long while (verification). Once the verification was complete, it showed the next Stage with a timestamp that wasn't recent which was what I expected.

Screenshot 2024-06-14 at 9 44 24 AM

# Motivation

Clarity in the UI for end users.

# Suggested Implementation

Should it show promotion date instead?

rbreeze:
Fixes #2171

CleanShot 2024-06-26 at 14 19 29@2x


internal/controller/stages/stages.go
```diff
@@ -1101,14 +1101,15 @@ func (r *reconciler) syncPromotions(
...
-	slices.SortFunc(newPromotions, func(a, b kargoapi.PromotionInfo) int {
+	slices.SortFunc(newPromotions, func(a, b kargoapi.PromotionReference) int {
 		return strings.Compare(a.Name, b.Name)
 	})
 
 	// Update the Stage status with the information about the newly terminated
 	// Promotions, and any new Freight that was successfully promoted.
 	for _, p := range newPromotions {
 		promo := p
+		promo.CompletionTime = &metav1.Time{Time: time.Now()}
```
krancour, line 1112: We can probably be more accurate on the time if we add a `CompletionTime` field to `PromotionStatus` as well and set that when the Promotion actually completes. Then here, we only need to copy it over from the real Promotion to the reference.

krancour:
The back-end parts of this LGTM, but I would appreciate a sanity check from @hiddeco.

api/v1alpha1/stage_types.go
```diff
@@ -685,13 +685,15 @@ type StageList struct {
...
-type PromotionInfo struct {
+type PromotionReference struct {
 	// Name is the name of the Promotion
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
 	// Freight is the freight being promoted
 	Freight FreightReference `json:"freight" protobuf:"bytes,2,opt,name=freight"`
 	// Status is the (optional) status of the promotion
 	Status *PromotionStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+	// CompletionTime is the time at which the Promotion was completed.
+	CompletionTime *metav1.Time `json:"completionTime,omitempty" protobuf:"bytes,4,opt,name=completionTime"`
```
hiddeco, line 696:
```suggestion
	FinishedAt *metav1.Time `json:"finishedAt,omitempty" protobuf:"bytes,4,opt,name=finishedAt"`
```

maybe? To make the `CreatedAt`, etc. from upstream (and align with e.g. Argo CD).

api/v1alpha1/promotion_types.go
```diff
@@ -103,6 +103,8 @@ type PromotionStatus struct {
 	Metadata map[string]string `json:"metadata,omitempty" protobuf:"bytes,3,rep,name=metadata" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
 	// Freight is the detail of the piece of freight that was referenced by this promotion.
 	Freight *FreightReference `json:"freight,omitempty" protobuf:"bytes,5,opt,name=freight"`
+	// CompletionTime is the time when the promotion was completed.
+	CompletionTime *metav1.Time `json:"completionTime,omitempty" protobuf:"bytes,6,opt,name=completionTime"`
```
hiddeco, line 107: Same here.

krancour (approved):
There's a UI linting issue to resolve, but preemptively LGTM'ing.

hiddeco (approved):
LGTM as well, thanks @rbreeze 🙇

PR #2213, opened 2024-06-27
refactor: make origin a struct instead of a string

krancour:
@hiddeco as we have previously discussed, there seems a high probability of additional kinds of Freight origins in the future (i.e. besides Warehouse).

Making this change proactively should prevent that eventuality from breaking our API again.


api/v1alpha1/stage_types.go
```diff
@@ -200,16 +207,50 @@ type StageSubscription struct {
...
+	// +kubebuilder:validation:Required
+	Name string `json:"name" protobuf:"bytes,2,opt,name=name"`
+}
+
+func (f *FreightOrigin) String() string {
+	if f == nil {
+		return ""
+	}
+	return fmt.Sprintf("%s[%s]", f.Kind, f.Name)
+}
```
krancour, line 242: @hiddeco we discussed this briefly offline. Since I am always calling `.String()` explicitly anyway, I am open to renaming this method if you feel strongly that this is an inappropriate implementation of the Stringer interface.

api/v1alpha1/stage_types.go
```diff
@@ -200,16 +207,50 @@ type StageSubscription struct {
...
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name" protobuf:"bytes,2,opt,name=name"`
+}
+
+func (f *FreightOrigin) String() string {
+	if f == nil {
+		return ""
+	}
+	return fmt.Sprintf("%s[%s]", f.Kind, f.Name)
```
hiddeco, line 241: I kind of feel something like

```suggestion
	return fmt.Sprintf("%s/%s", f.Kind, f.Name)
```

might be a better fit. As `[]` is often used for grouping (e.g. API group versions, see `GroupVersions` in Kubernetes' `schema` package), while `/` is a common way for GVK / name joins in the Kubernetes (ecosystem) itself.

krancour: What I had been going for was making it feel like it identifies a specific origin from a collection of origins, but I am equally happy with the alternative you have suggested. Happy to change.

PR #2218, opened 2024-06-28
chore: clarify optionality

blakepettersson:
In the CRD docs `freightCreationPolicy` is required, even though strictly speaking if not specified it will be set to `Automatic`.


hiddeco:
Think the required validation should stay put, as the help we provide of filling in a default is different from the field being optional.

krancour (approved):

krancour:
Thanks @blakepettersson!

PR #2230, opened 2024-07-02
chore: audit controller and management controller permissions

krancour:


charts/kargo/templates/controller/cluster-roles.yaml
```diff
@@ -50,7 +50,6 @@ rules:
   - list
   - patch
   - promote
-  - update
```
krancour, line 53: The only update to Stages involves removing a finalizer and that can quite easily be done with a patch.

charts/kargo/templates/controller/cluster-roles.yaml
```diff
@@ -40,7 +40,6 @@ rules:
   - get
   - list
   - watch
-  - patch
```
krancour, line 43: The controller doesn't need to patch Freight or Promotions at all.

charts/kargo/templates/controller/cluster-roles.yaml
```diff
@@ -57,7 +57,6 @@ rules:
   verbs:
   - get
   - list
-  - patch
```
krancour, line 62: The controller does not need to patch Warehouses at all.

charts/kargo/templates/controller/cluster-roles.yaml
```diff
@@ -67,7 +67,6 @@ rules:
   - warehouses/finalizers
   - warehouses/status
   verbs:
-  - update
```
krancour, line 73: The controller only ever patches status subresources.

charts/kargo/templates/controller/cluster-roles.yaml
```diff
@@ -114,7 +114,6 @@ rules:
   - create
   - get
   - list
-  - patch
```
krancour, line 121: I don't believe there is actually any need to patch AnalysisRuns. Everything works correctly e2e without this.

@hiddeco maybe you can give me a sanity check on this one, however.

charts/kargo/templates/controller/cluster-roles.yaml
```diff
@@ -106,6 +98,14 @@ metadata:
...
     {{- include "kargo.controller.labels" . | nindent 4 }}
 rules:
+- apiGroups:
+  - argoproj.io
+  resources:
+  - analysistemplates
+  verbs:
+  - get
+  - list
+  - watch
```
krancour, line 108: These permissions are better off down here where they will not be granted in the event that Rollouts integration is completely disabled.

charts/kargo/templates/management-controller/cluster-roles.yaml
```diff
@@ -16,7 +16,6 @@ rules:
   - get
   - list
   - patch
-  - update
```
krancour, line 19: Not used at all.

charts/kargo/templates/management-controller/cluster-roles.yaml
```diff
@@ -46,5 +46,4 @@ rules:
   - projects/status
   verbs:
   - patch
-  - update
```
krancour, line 50: This permission was already unused.

api/v1alpha1/helpers.go
```diff
@@ -33,14 +33,6 @@ func AddFinalizer(ctx context.Context, c client.Client, obj client.Object) error
 	return nil
 }
 
-func ClearAnnotations(ctx context.Context, c client.Client, obj client.Object, keys ...string) error {
-	kvs := make(map[string]*string, len(keys))
-	for _, k := range keys {
-		kvs[k] = nil
-	}
-	return patchAnnotations(ctx, c, obj, kvs)
-}
```
krancour: Seems we stopped clearing refresh annotation quite some time ago and that was the only thing we were ever using this for.

charts/kargo/templates/controller/cluster-roles.yaml
krancour: nm. I just noticed the DM you sent me about this hours ago.

It's for aborting a verification's underlying AnalysisRun.

I'll fix this in the morning.

hiddeco (approved):

PR #2232, opened 2024-07-03
feat: make promotions work in multi-pipeline feature branch

krancour:
Notes:

1. A Promotion continues to reference only a single piece of Freight.
1. The Promotion reconciler builds a collection of FreightReferences by taking the latest from history and adding/updating the reference to the piece of Freight that is actively being promoted.
1. It is this collection of FreightReferences that gets passed to all the promotion mechanisms now.
1. We continue to _not_ attempt auto-promotions for Stages that request Freight from multiple origins.

__This works end-to-end!__

Known issues:

1. Verifications fail for Stages with more than one current Freight. 😢 This is because the values we're setting on the `kargo.akuity.io/freight` label of an AnalysisRun is now too long.

1. If n > 1 pieces of Freight _from different origins_ contain artifacts from the same repository, we're not really doing anything yet to disambiguate which of the two a promotion mechanism references. You end up with the artifact from whichever of the n origins is alphabetically first.

These are are fixable, but it would be my preference to address them in a follow-up.


internal/controller/promotions/promotions.go
```diff
@@ -486,3 +492,28 @@ func (r *reconciler) promote(
...
+// targetFreight.
+func (r *reconciler) buildTargetFreightCollection(
+	targetFreight kargoapi.FreightReference,
+	stage *kargoapi.Stage,
+) *kargoapi.FreightCollection {
+	freightCol := &kargoapi.FreightCollection{}
+
+	// We don't simply copy the current FreightCollection because we want to
+	// account for the possibility that some freight contained therein are no
+	// longer requested by the Stage.
```
hiddeco, line 508: Should we be careful about the opposite thing potentially happening? I.e. there is now Freight requested for which we do not have an item in the stack (yet).

krancour: That will happen, but I don't think there's anything special about that scenario as it is essentially no different than a _first-time_ promotion of one piece of Freight to a new Stage that requires n > 1. That's a scenario we had discussed and iirc, we agreed the likely result is reaching a state that will either not become healthy or will fail verification, but that that is also correct behavior for such a scenario. A subsequent promotion will set things right.

_But..._

> we agreed the likely result is reaching a state that will either not become healthy or will fail verification

I'm thinking now about scenarios where this may not be true -- perhaps, for instance, a scenario wherein the Promotion itself cannot succeed without all n Freight. In that case, the Stage's FreightHistory will remain forever empty, making it impossible for any promotion to ever provide the nth of n requested Freight as they will always be providing only the 1st of n.

I will noodle on this for a bit. Please lmk if you have ideas.

krancour: Oh... if we built up the collection of Freight used by the Promotion (apart from the Freight referenced _by_ the Promotion) by getting it from `stage.Status.LastPromotion.Status.FreightCollection` _instead of_ `stage.Status.FreightHistory.Current().Freight` that may solve this. It would allow a second promotion, providing the 2nd of n > 1 Freight to build upon the failure of the first Promotion. For the average case where Promotions succeed, the overall effect of this change would still be identical to the current logic.

wdyt?

hiddeco: I like this idea better, as it would allow you to "fill" Freight until your promotion succeeds.

krancour: Ok. Will do this.

krancour: @hiddeco, this is fixed in 73c5e8e00a3ac36e875d64b4cfd9ab7a36d3db75

hiddeco (approved):

PR #2235, opened 2024-07-03
feat(controller): irsa support

blakepettersson:
Building upon #2066, this PR allows IRSA to be used. The reason why we would want to use this is because there are some environments where EKS Pod Identites cannot be used (e.g AWS GovCloud), or e.g for consistency with existing deployments.


charts/kargo/templates/controller/service-account.yaml
```diff
@@ -7,4 +7,10 @@ metadata:
   labels:
     {{- include "kargo.labels" . | nindent 4 }}
     {{- include "kargo.controller.labels" . | nindent 4 }}
+  {{- with (mergeOverwrite (deepCopy .Values.global.annotations) .Values.controller.annotations) }}
+  annotations:
+    {{- range $key, $value := . }}
+      {{ $key }}: {{ $value | quote }}
+      {{- end }}
+  {{- end }}
```
krancour, line 15: I think this is maybe not how I would approach this. Thus far, this whole strategy of merging global labels and/or annotations with component-specific ones has only been used on workloads (Pods) and things managing workloads (Deployments and CronJobs). Mostly this has been done for the benefit of people who, due to policy or whatever, need workloads labeled or annotated a certain way.

If we approach things in this way, someone is going to add IRSA-related annotations to either `.Values.global.annotations` or `.Values.controller.annotations`, and depending on which, they're going to end up with those annotations added in a whole lot of places where they are not needed -- in which case they are just noise.

I think the best approach might be to add a dedicated `controller.irsa.roleName` or something of that nature and use its value here if and only if non-empty defined.

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -399,8 +398,32 @@ to set up EKS Pod Identity in your EKS cluster and assign an IAM role to the
...
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": "sts:AssumeRole",
+            "Resource": "arn:aws:iam::ACCOUNT_ID:role//kargo-project-*"
+        }
+    ]
+}
+```
```
krancour, line 422: I'm not positive that this is the best way to do this. (Trust policies may be better ways to say what Roles can be assumed by other Roles). But rather than get caught up in that...

imho, we don't want our docs to get overly mired in AWS details. Their docs give better guidance on topics such as these than we are able to anyway. This is why, in the Pod Identity section, I deliberately gave minimal, but sufficient info about the Kargo end of things and then referred readers to AWS docs for further details:

> At this point, an IAM role will be associated with the Kargo controller, however, that controller acts on behalf of multiple Kargo projects, each of which may require access to different ECR repositories. To account for this, when Kargo attempts to access an ECR repository on behalf of a specific project, it will first attempt to [assume an IAM role](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) specific to that project. The name of the role it attempts to assume will always be of the form kargo-project-<project name>. It is this role that should be granted read-only access to applicable ECR repositories.

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -378,17 +378,16 @@ to the required ECR repositories.
...
-of EKS, this method will still work.
+will work regardless of where Kargo is deployed. i.e. if running Kargo outside EKS, this method will still work.
 
-If running Kargo within EKS, you may wish to consider using EKS Pod Identity
+If running Kargo within EKS, you may wish to either consider using EKS Pod Identity or IRSA
 instead.
 :::
 
-#### EKS Pod Identity
+##### EKS Pod Identity (default)
```
krancour, line 387: In an attempt to say more with fewer words, I would consider re-titling this section as "EKS Pod Identity or IAM Roles for Service Accounts (IRSA)" and then make _small_ tweaks to the text of this section. The approach to both is so very similar, that it shouldn't be too hard to cover these two as a pair.

IRSA requires some additional configuration when the controller is installed and those details can be called out in an `:::info` or `:::note` box.

internal/credentials/kubernetes/ecr/pod_identity.go
```diff
@@ -38,20 +38,32 @@ type podIdentityCredentialHelper struct {
...
 	var awsAccountID string
-	if os.Getenv("AWS_CONTAINER_CREDENTIALS_FULL_URI") == "" {
-		logger.Info("AWS_CONTAINER_CREDENTIALS_FULL_URI not set; assuming EKS Pod Identity is not in use")
+	usesIRSA := os.Getenv("AWS_ROLE_ARN") != "" && os.Getenv("AWS_WEB_IDENTITY_TOKEN_FILE") != ""
+	if os.Getenv("AWS_CONTAINER_CREDENTIALS_FULL_URI") != "" {
+		logger.Info("EKS Pod Identity appears to be in use")
+	} else if usesIRSA {
+		logger.Info("AWS_WEB_IDENTITY_TOKEN_FILE and AWS_ROLE_ARN set; assuming IRSA is being used")
+	} else {
+		logger.Info("Neither AWS_CONTAINER_CREDENTIALS_FULL_URI nor AWS_WEB_IDENTITY_TOKEN_FILE and AWS_ROLE_ARN are set; assuming neither EKS Pod Identity nor IRSA are in use")
```
krancour, line 47: This logic appears more complex than it needs to be. Logic-wise, _everything else_ in this credential helper doesn't actually care if you're using Pod Identity or using IRSA. To the Go code, they're the same. So the logic here could really be as simple as "if at least one of these three is set, we're using managed identity of some kind and we're not otherwise."

This also makes me realize this helper should probably be renamed since it will pull double-duty for Pod Identity _and_ IRSA. `managedIdentityCredentialHelper` seems as if it would sufficiently cover both scenarios.

charts/kargo/templates/controller/service-account.yaml
blakepettersson: Yeah, that was actually my first thought 😄 I went with this approach instead, since it seems like there was a convention of doing that thing above... I'll amend this to do something like `controller.serviceAccount.annotations` instead.

internal/credentials/kubernetes/ecr/pod_identity.go
blakepettersson: > Logic-wise, everything else in this credential helper doesn't actually care if you're using Pod Identity or using IRSA. 

That is true, but IMO there should at least be something that indicates whether IRSA or EKS Pod Identity is being used, to aid people in debugging in case something goes wrong.

krancour: Fair. Log it once up front depending on which env var you find. Beyond that, I wouldn't put additional branches in the logic.

We should still rename the component.

charts/kargo/templates/controller/service-account.yaml
blakepettersson: I added a dedicated annotations field for the service account, should this be something which should be merged with `.Values.global.annotations`?

docs/docs/30-how-to-guides/20-managing-credentials.md
blakepettersson: I modified the docs to be a bit more compact, but that the same time tried to also give a few minor pointers in regards to role assumption. What do you think?

charts/kargo/templates/controller/service-account.yaml
krancour: For now I think no. We can revisit if we determine there is a need.

charts/kargo/values.yaml
```diff
@@ -337,6 +337,11 @@ controller:
   ## @param controller.podAnnotations Optional annotations to add to pods. Merges with `global.podAnnotations`, allowing you to override or add to the global annotations.
   podAnnotations: {}
 
+  ## All settings relating to the service account for the controller
+  serviceAccount:
+    ## @param controller.serviceAccount.annotations Optional annotations to add to the service account.
```
krancour, line 342: I think it may be worthwhile mentioning in the comment that this is the appropriate place to include any annotations required by IRSA.

internal/credentials/kubernetes/ecr/managed_iam_credential.go
```diff
@@ -33,27 +33,31 @@ type podIdentityCredentialHelper struct {
 	) (string, error)
 }
 
-// NewPodIdentityCredentialHelper returns an implementation of
+// NewManagedIAMCredentialHelper returns an implementation of
 // credentials.Helper that utilizes a cache to avoid unnecessary calls to AWS.
-func NewPodIdentityCredentialHelper(ctx context.Context) credentials.Helper {
+func NewManagedIAMCredentialHelper(ctx context.Context) credentials.Helper {
```
krancour, line 38: 👍

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -378,27 +378,54 @@ to the required ECR repositories.
...
+If you opt to use EKS Pod Identity, follow
 [this overview](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html#pod-id-setup-overview)
 to set up EKS Pod Identity in your EKS cluster and assign an IAM role to the
 `kargo-controller` `ServiceAccount` within the `Namespace` to which Kargo is (or
 will be) installed.
 
+If you opt to instead use IRSA, follow [this overview](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) 
+to set up IRSA in your EKS cluster and assign an IAM role to the
+`kargo-controller` `ServiceAccount` within the `Namespace` to which Kargo is (or
+will be) installed.
```
krancour, line 404: Suggested edits:

```
If Kargo locates no `Secret` resources matching a repository URL and  is deployed
within an EKS cluster, it will attempt to use
[EKS Pod Identity](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html)
or
[IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)
to authenticate. Both of these rely upon some external setup. Leveraging either
eliminates the need to store ECR credentials in a `Secret` resource.

Follow
[this overview](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html#pod-id-setup-overview)
to set up EKS Pod Identity in your EKS cluster or
[this one](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) to set up IRSA.
For either, you will assign an IAM role to the `kargo-controller` `ServiceAccount`
within the `Namespace` to which Kargo is (or will be) installed.
```

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -378,27 +378,54 @@ to the required ECR repositories.
...
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": "sts:AssumeRole",
+            "Resource": "arn:aws:iam::ACCOUNT_ID:role/kargo-project-*"
+        }
+    ]
+}
```
krancour, line 426: I know we had some disagreement over how in-depth we wanted to go on policy details. I know your concern was that not being explicit enough will result in additional support cases. I think being too prescriptive can yield similar results: Instead of hearing, "I can't figure this AWS stuff out," we'll be hearing, "I tried what the docs said and it didn't work." Worse, is someone more seasoned in IAM than we are complaining, "The way the docs say to do it isn't right," _which strikes me as a distinct possibility, because what you have above is not how I would have done it._ (Which is not to say my way is right and yours is wrong. But if there's more than one way to do it, you can be sure _someone_ with a strong opinion is going to tell you that your way is wrong.)

I guess I am invoking "less is more" here. My gut says the people who would consider using Pod Identity and/or IRSA in the first place are going to already be well-versed in IAM. This won't be their first rodeo. If we give them the few details that are actually Kargo-specific (such as the naming pattern for Project-specific roles), they'll be successful. If I'm proven wrong, we can totally revisit this.

docs/docs/30-how-to-guides/20-managing-credentials.md
krancour: The main thing here is I don't want it to flow like, "First we try this and _then_ we try that," since that isn't the case. If either is set up, it will just work.

internal/credentials/kubernetes/ecr/managed_iam_credential.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/akuity/kargo/internal/logging"
 )
 
-type podIdentityCredentialHelper struct {
+type managedIAMCredentialHelper struct {
```
krancour, line 22: "managed IAM" sounds redundant to me. (__Managed__ Identity and Access __Management__)

I think `managedIdentityCredentialHelper` seems like it accurately covers both Pod Identity and IRSA.

krancour (approved):

PR #2236, opened 2024-07-03
feat(ui): Support multiple pipelines

rbreeze:
Screenshot 2024-07-03 at 14 48 23

Screenshot 2024-07-03 at 14 29 52


krancour:
@rbreeze can you please amend this to uncomment line 37 of the Dockerfile. I commented it in a previous PR because our backend changes had broken the UI so badly that we needed to remove it from the image build. Uncommenting that line will undo that.

hiddeco:
Bit of a nit, but would it be possible in the second screenshot to have the lines go straight instead of the kink they have now?

krancour:
> Bit of a nit, but would it be possible in the second screenshot to have the lines go straight instead of the kink they have now?

We agreed to handle this in a follow-up.

krancour (approved):

PR #2238, opened 2024-07-04
feat: add unique, deterministically calculated ids for freight collections

krancour:
This should address a known issue with some label values for AnalysisRuns being too long in the event of a Stage requesting Freight from multiple origins.


hiddeco (approved):
Beautiful! 💯

Issue #2216, opened 2024-06-28
Stage incorrectly marked as unhealthy when application is suspended


billyshambrook:
# Description

When the application uses a Rollout with a `pause` step, ArgoCD application status is `suspended`. Kargo incorrectly shows this as `unhealthy`.

# Screenshots

# Steps to Reproduce

# Version

```
kargo version
Client Version: v0.7.1
Server Version: v0.7.1
```

# Logs

```
Paste any relevant application logs here.
```

krancour:
I would want @hiddeco to weigh in on this, as this seems to me to be a gray area.

While Application health _influences_ Stage health, there is no direct _equivalence_ between the two. To illustrate, note that some Stages may involve multiple Applications, in which case a Stage can be no healthier than the least healthy App. Also, because a Stages host specific freight that reference specific versions of artifacts, Stages have some notion of what revision each App should be synced to and deviation from those expectations is a health problem. All of this is to say that compared with App health, Stage health reflects a much more comprehensive analysis of what is going on.

On to the question of what should a Stage's health actually be when an associated App is Suspended...

I see a lot of different possibilities and various consequences of each:

1. App is Suspended, so Stage is Suspended.

    This is a health state that Stages don't currently have. It would be new.

    The question I ask myself here is to what extent this is useful/meaningful information at the Stage level. If all Apps associated with the Stage are healthy, apart from one or more of them being Suspended, what does that mean to Kargo? Doesn't it mean the _Stage_ is still attempting to reach a healthy state? And how does that differ from Progressing?

   Which leads directly to the next option...

2. App is Suspended, so Stage is Progressing

    Since I understand that Stage health and App health are different things, this makes the most sense to me, _but I realize that this is probably going to confuse most users._

3. App is Suspended, so Stage is Healthy

    This is true from a certain perspective. Nothing is actually _wrong_. The obvious problem is that when a Stage becomes Healthy for the first time following a Promotion, any associated verification processes are triggered, but they really ought not be until associated Applications are actually Healthy.

    So this is probably a bad option.

4. App is Suspended, so Stage is Unhealthy

    This is where we're at today and this behavior was clearly surprising -- thus why we're here having this conversation.

    This is true from a certain perspective. The Stage has not reached a stable state and it would be premature to move on to executing any verification processes. So if it's not Healthy and not Unhealthy, it's Progressing, right? (No.2 -- which I conceded will likely confuse people.)

[...]

PR #2243, opened 2024-07-05
fix(argocd): mark suspended App as progressing

hiddeco:
Fixes: #2216


krancour (approved):
Thanks for doing this so fast!

Issue #2245, opened 2024-07-06
cli: `kargo get freight` results should have an origin column


krancour:
In the context of multiple pipelines, this seems like a tremendously important bit of information.

We should also add the same column to `kubectl get freight` results.

We should probably also give `kargo get freight` the ability to filter on origin.

@hiddeco is this one you'd be willing to look at?

krancour:
Closed by #2257

PR #2257, opened 2024-07-08
feat(cli): add origin support to `kargo get freight`

hiddeco:
Fixes: #2245

This adds an origin column to `kargo get freight`, and origin kind/name columns to `kubectl get freight` (which can not use the stringify function as it directly reads/prints from the raw data).

In addition, it adds a `--origin` flag to `kargo get freight` to allow filtering Freight by origin(s).


internal/cli/cmd/get/freight.go
```diff
@@ -108,6 +112,11 @@ func (o *getFreightOptions) addFlags(cmd *cobra.Command) {
 	)
 	option.Names(cmd.Flags(), &o.Names, "The name of a piece of freight to get.")
 	option.Aliases(cmd.Flags(), &o.Aliases, "The alias of a piece of freight to get.")
+	option.Origins(cmd.Flags(), &o.Origins, "The origin of the freight to get.")
```
krancour, line 115: I am iffy on this. How do we with the `--origin` flag that we're looking from Freight from a Warehouse named `foo` vs (looking to the future) some other kind of origin (Junction?) named `foo`?

I am wondering if something like `--warehouse foo` (and later `--junction foo`) might be more expressive.

hiddeco: I think it would change to e.g. `Warehouse/foo` to match the identifier we use elsewhere.

If the origin does not contain a `/`, we assume it's a Warehouse.

krancour: I think thus far we've only used the form `Warehouse/foo` for display purposes, but I am ok with what you propose with the caveat that we revisit it if it proves to be a pain point after additional origin types are a reality.

PR #2259, opened 2024-07-08
feat: disambiguate origins

### LINKED ISSUE #2246: feat: ability to disambiguate artifact origin
This is part of the epic that is #1680, but it is a critical piece that is still outstanding, so I want to track it as its own issue.

When a Stage requests pieces of Freight from multiple origins, it is possible that there will be multiple pieces of Freight containing different versions of the same artifact.

Example: A GitOps monorepo has manifests for multiple microservices. One Warehouse subscribes to that repo with includePath "microservice-a". A second Warehouse subscribes to that repo with includePath "microservice-b". All Stages of the microservice application request Freight from both origins.

When such Stages execute Git-based or Argo CD-based promotion mechanisms during the promotion process, it will sometimes be necessary to clarify whether the mechanism operating on an artifact identified by its repo URL means to work with the version found in the microservice-a Freight or the microservice-b Freight.

A similar need for disambiguation arises where health checks are concerned. If a microservice-a and microservice-b App both sync to the same repo, which is the correct targetRevision that each of those should be synced to?

The most sensible thing to do, imho, is to have this qualifying information be optional -- as it will be unnecessary in many, many cases -- with the caveat that if Kargo can't figure things out on its own, it will lead to errors in the promotion or health check processes that will require clarification from users to resolve.

This work is already in progress, but as noted, is one of the larger bits of outstanding work for v0.8.0.

krancour:
Fixes #2246


internal/controller/freight/finder.go
```diff
@@ -0,0 +1,230 @@
+package freight
```
hiddeco, line 1: Non-blocking nit: I think that if we would bring this into a more object-oriented shape, we could get rid of some of the repeated code around resolving the origin, while also preventing having to expose two different clients to the health related code.

krancour: I had thoughts of that as well, but am also trying not to hold up v0.8.0 further. I would be happy to revisit it in the future and I expect our plans for v0.9.0 should give us opportunities to do so.

krancour:
@hiddeco this works end-to-end.

internal/controller/promotion/git.go
```diff
@@ -209,78 +225,73 @@ func (g *gitMechanism) doSingleUpdate(
...
-			}
-		}
+	ctx context.Context,
+	cli client.Client,
+	stage *kargoapi.Stage,
+	update *kargoapi.GitRepoUpdate,
+	newFreight []kargoapi.FreightReference,
+) (string, *kargoapi.GitCommit, error) {
+	desiredOrigin := freight.GetDesiredOrigin(stage, update)
+	fmt.Printf("----> desiredOrigin: %v\n", desiredOrigin)
```
hiddeco, line 285: Left-over from debugging?

krancour: Sure is. Thanks.

hiddeco (approved):

PR #2269, opened 2024-07-09
fix: remediate errant use of deprecated fields

krancour:
This remediates some errant uses of newly deprecated fields and in doing so also happens to address some outstanding todos and fix some previously undetected bugs. (Abort verification and reverification were not working correctly in this feature branch.)

### DIFF
--- a/api/v1alpha1/stage_helpers.go
+++ b/api/v1alpha1/stage_helpers.go
@@ -136,39 +136,28 @@ func ReverifyStageFreight(
 		return err
 	}
 
-	current := stage.Status.FreightHistory.Current()
-	if current == nil || len(current.Freight) == 0 {
+	currentFC := stage.Status.FreightHistory.Current()
+	if currentFC == nil || len(currentFC.Freight) == 0 {
 		return errors.New("stage has no current freight")
 	}
 
-	if len(current.Freight) > 1 {
-		// TODO: This is a naive implementation which does not handle multiple
-		// Freight objects in the history properly. This must be addressed for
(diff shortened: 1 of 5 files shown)


api/v1alpha1/stage_helpers.go
```diff
@@ -136,39 +136,28 @@ func ReverifyStageFreight(
...
-		// TODO: This is a naive implementation which does not handle multiple
-		// Freight objects in the history properly. This must be addressed for
-		// proper "multi-pipeline" support.
-		return errors.New("stage has multiple Freight objects: re-verification not supported at this time")
+	currentVI := currentFC.VerificationHistory.Current()
+	if currentVI == nil {
+		return errors.New("stage has no current verification info")
 	}
 
-	for _, f := range current.Freight {
```
krancour, line 151: This loop pre-dated the VerificationHistory moving up a level from Freight to FreightCollection. The change made for aborting non-terminal verifications is basically the same loop removal.

internal/controller/stages/stages.go
```diff
@@ -694,6 +694,7 @@ func (r *reconciler) syncNormalStage(
 					if req, _ := kargoapi.ReverifyAnnotationValue(stage.GetAnnotations()); req.ForID(currentVI.ID) {
 						logger.Debug("rerunning verification")
 						status.Phase = kargoapi.StagePhaseVerifying
+						currentVI = &kargoapi.VerificationInfo{}
```
krancour, line 697: Absence of this was a bug.

hiddeco (approved):

Issue #2260, opened 2024-07-08
feat: make auto-promotion behave correctly for multiple pipelines


krancour:

hiddeco:
Has been addressed by #2273.

PR #2273, opened 2024-07-09
feat(stage): support multi-pipeline auto-promotion

hiddeco:
Fixes: #2260

This adds support for promoting multiple Freight from different origins to a Stage during auto-promotion, by ensuring there is a Promotion for the latest Freight of each request/origin.

In support of this, the following notable changes were made:

- The `getAvailableFreight` function has been replaced with `getAvailableFreightByOrigin`, which returns a map of Freight grouped by their origin. In addition, tiny bits of the previous logic have been optimized (e.g. stop looking for Freight in other places if `.origin.direct` is set).
- The `FreightReference` field in the `PromotionReference` is now optional, to prevent the `.origin.kind` validation from being triggered in edge-case scenarios.
- The Promotion reconciler confirms the Stage is currently aware of the Promotion before continuing. This prevents race conditions that could lead to compusing the wrong `FreightCollection` during a promotion. In a future iteration, we should decouple this (again), for which I have recorded some notes in a `// TODO`.


krancour (approved):
:fire: This was really straightforward and a test drive (merged with the latest from multi-pipeline) passed with flying colors.

LGTM after the merge conflict in generated code has been resolved.

PR #2285, opened 2024-07-11
feat: if project-scoped ECR auth fails, fallback to default

rbreeze:


internal/credentials/kubernetes/ecr/pod_identity.go
```diff
@@ -158,7 +160,20 @@ func (p *podIdentityCredentialHelper) getAuthToken(
 	output, err := ecrSvc.GetAuthorizationToken(ctx, &ecr.GetAuthorizationTokenInput{})
 	if err != nil {
 		logger.Error(err, "error getting ECR authorization token")
-		return "", nil
+		var ae smithy.APIError
+		if errors.As(err, &ae) {
+			if ae.ErrorCode() != "AccessDenied" {
```
krancour, line 165: Is there a numeric code available by any chance?

krancour: Non-blocking. This LGTM, but I request @hiddeco to take a peek also.

rbreeze: Doesn't appear to be unfortunately... here's the interface: 
```go
type APIError interface {
	error

	// ErrorCode returns the error code for the API exception.
	ErrorCode() string
	// ErrorMessage returns the error message for the API exception.
	ErrorMessage() string
	// ErrorFault returns the fault for the API exception.
	ErrorFault() ErrorFault
}
```

`ErrorFault` is basically an int, but only has 3 values
```go
const (
	FaultUnknown ErrorFault = iota
	FaultServer
	FaultClient
)
```

hiddeco: I think that would only be available if you _assume_ you always make use of an HTTP transport and instead cast it to a `ResponseError` from the AWS SDK itself.

The other option would be to go and look to see if any of the `services/<service>/types` expose a specific thing to describe authentication and/or access denied errors.

krancour: I did some digging and we can skip the interface and go directly after the underlying implementation (`*smithy.OperationError`), but I found the information available from it to be no more structured than what we can get via that interface.

I also looked at the smithy code to see if they had public constant to represent "AccessDenied" and they do not.

So I agree, this is as good as we're going to get on the error handling here.

This is our first direct use of smithy -- it was only used indirectly prior to this.

I think you need to run `go mod tidy` and it will switch it from a transitive dependency to a direct one.

rbreeze: Sounds good. I just pushed the alternative approach using `ResponseError`. 

I believe I already ran `go mod tidy` to address the go.mod issue but if we decide to go with the `ResponseError` approach, we no longer directly use smithy. LMK which approach you prefer, I can rollback my last commit as necessary

internal/credentials/kubernetes/ecr/pod_identity.go
```diff
@@ -158,7 +160,20 @@ func (p *podIdentityCredentialHelper) getAuthToken(
 	output, err := ecrSvc.GetAuthorizationToken(ctx, &ecr.GetAuthorizationTokenInput{})
 	if err != nil {
 		logger.Error(err, "error getting ECR authorization token")
-		return "", nil
+		var re *awshttp.ResponseError
+		if errors.As(err, &re) {
```
krancour, line 164: Oooooh. Nice find.

internal/credentials/kubernetes/ecr/pod_identity.go
```diff
@@ -158,7 +160,20 @@ func (p *podIdentityCredentialHelper) getAuthToken(
...
+			if re.HTTPStatusCode() != 403 {
```
krancour, line 164:
```suggestion
			if re.HTTPStatusCode() != http.StatusForbidden {
```

internal/credentials/kubernetes/ecr/pod_identity.go
```diff
@@ -157,8 +159,21 @@ func (p *podIdentityCredentialHelper) getAuthToken(
 	)
 	output, err := ecrSvc.GetAuthorizationToken(ctx, &ecr.GetAuthorizationTokenInput{})
 	if err != nil {
-		logger.Error(err, "error getting ECR authorization token")
-		return "", nil
+		var re *awshttp.ResponseError
+		if errors.As(err, &re) {
```
krancour, line 163: It looks like we're not handling the alternative -- that this is _not_ a `*awshttp.ResponseError`, in which we probably just want to return the error.

internal/credentials/kubernetes/ecr/pod_identity.go
```diff
@@ -157,8 +159,21 @@ func (p *podIdentityCredentialHelper) getAuthToken(
...
+			if re.HTTPStatusCode() != 403 {
+				return "", err
+			}
+			logger.Info(
```
krancour, line 167:
```suggestion
			logger.Debug(
```

This is likely to happen a lot and it's part of the normal course of business, so we should probably log it at a lower level.

internal/credentials/kubernetes/ecr/pod_identity.go
```diff
@@ -157,8 +160,22 @@ func (p *podIdentityCredentialHelper) getAuthToken(
...
 	if err != nil {
-		logger.Error(err, "error getting ECR authorization token")
-		return "", nil
+		var re *awshttp.ResponseError
+		if !errors.As(err, &re) {
+			return "", err
+		}
+		if re.HTTPStatusCode() != http.StatusForbidden {
+			return "", err
+		}
```
krancour, line 169:
```suggestion
		if !errors.As(err, &re) || re.HTTPStatusCode() != http.StatusForbidden {
			return "", err
		}
```

krancour (approved):

PR #2288, opened 2024-07-12
fix(ui): multiple edges to multiple children

rbreeze:


krancour (approved):
We discussed further improvements offline, but as this is such a great step in the right direction, there's no reason not to merge it.

PR #2295, opened 2024-07-15
fix(webhook): compare to detect Freight mutations 

hiddeco:
This ensures that from now on, the webhook mechanism is no longer affected by changes to the logic of Freight ID generation.

Which previously, could cause "Freight is immutable" errors when e.g. Freight from a (much) older version of Kargo would receive an update to its alias, or get verified for a Stage.

### DIFF
--- a/internal/webhook/freight/webhook.go
+++ b/internal/webhook/freight/webhook.go
@@ -116,17 +116,20 @@ func newWebhook(
 
 func (w *webhook) Default(ctx context.Context, obj runtime.Object) error {
 	freight := obj.(*kargoapi.Freight) // nolint: forcetypeassert
-	// Re-calculate ID in case it wasn't set correctly to begin with -- possible
-	// if/when we allow users to create their own Freight.
-	freight.Name = freight.GenerateID()
 
-	// Sync the convenience alias field with the alias label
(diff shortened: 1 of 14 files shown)


internal/webhook/freight/webhook.go
```diff
@@ -116,17 +116,20 @@ func newWebhook(
...
-	// Sync the convenience alias field with the alias label
 	req, err := admission.RequestFromContext(ctx)
 	if err != nil {
 		return apierrors.NewInternalError(
 			fmt.Errorf("error getting admission request from context: %w", err),
 		)
 	}
+	if req.Operation == admissionv1.Create {
+		// Re-calculate ID in case it wasn't set correctly to begin with -- possible
+		// if/when we allow users to create their own Freight.
```
krancour, line 128: Oh... we can fix this comment. It's not theoretical anymore.

krancour (approved):

internal/webhook/freight/webhook.go
hiddeco, line 128:
```suggestion
		// Re-calculate ID in case it wasn't set correctly to begin with -- possible
		// when users create their own Freight.
```

PR #2309, opened 2024-07-16
feat: add v0.7.x --> v0.8.x upgrade logic

hiddeco:
This PR provides a smooth upgrade path from Kargo v0.7.x to v0.8.x, focusing on migrating Stages and Freight resources to support the multi-pipeline functionality.

Key changes:

1. Stage migration:
   - Adds upgrade logic to the Stage reconciler to migrate Stages from <=0.7.x to 0.8.x compatibility
   - Converts old `Subscriptions` field to new `RequestedFreight` field
   - Migrates `History` to new `FreightHistory` format
   - Adds v0.8 compatibility label to track which resources have been migrated

2. Freight migration:
   - Introduces a new Freight reconciler in the management controller to handle upgrades
   - Migrates `Warehouse` field to new `Origin` field
   - Ensures backward compatibility by syncing `Origin` and `Warehouse` fields
   - Adds v0.8 compatibility label to prevent further migrations

3. Warehouse reconciliation:
   - Prevents duplication of Freight after an upgrade by comparing existing Freight with newly created ones, this is required due to a change in how the Freight name is generated

4. Promotion reconciliation:
   - Adds a check for upgraded Freight in the promotion reconciliation loop
   - Waits for Freight to be updated before proceeding with promotions

5. Webhook updates:
   - Modifies Freight webhook to allow migration from Warehouse to Origin
   - Ensures immutability of Freight while allowing necessary upgrades

6. Chart updates:
   - Adds required permissions for migration to the (management-)controller cluster roles

### DIFF
--- a/internal/controller/management/upgrade/freight.go
+++ b/internal/controller/management/upgrade/freight.go
@@ -116,7 +116,6 @@ func (f *freightReconciler) Reconcile(
 	// Migrate the Warehouse field to the Origin field.
 	freight.Origin.Kind = kargoapi.FreightOriginKindWarehouse
 	freight.Origin.Name = freight.Warehouse // nolint: staticcheck
-	freight.Warehouse = ""                  // nolint: staticcheck
 
 	if err = f.client.Update(ctx, freight); err != nil {
 		return ctrl.Result{}, err
--- a/internal/controller/management/upgrade/freight.go
+++ b/internal/controller/management/upgrade/freight.go
@@ -1,0 +1,132 @@
+package upgrade
+
+import (
+	"context"
+	"errors"
+
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/logging"
+)
+
+// freightReconciler reconciles Freight resources to upgrade them from
+// v0.7.x to v0.8.x.
+type freightReconciler struct {
+	client client.Client
+}
+
+// SetupFreightReconcilerWithManager initializes a freightReconciler and
+// registers it with the provided Manager.
+func SetupFreightReconcilerWithManager(mgr manager.Manager) error {
+	_, err := ctrl.NewControllerManagedBy(mgr).
+		For(&kargoapi.Freight{}).
+		WithEventFilter(
+			predicate.Funcs{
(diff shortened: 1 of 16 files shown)


krancour:
@hiddeco I didn't test drive yet. (I will.)

Logic looks sound.

I do have a minor concern over the expanded permissions required by both controllers in order to do this and I'll DM you about it.

krancour:
@hiddeco do we also need some small modification to the top of the Promotion reconciliation loop to verify any Freight referenced by the Promotion have been upgraded prior to starting/resuming a Promotion?

internal/controller/management/upgrade/freight.go
```diff
@@ -0,0 +1,132 @@
...
+	// If the Warehouse field is not set, we can't migrate the Freight.
+	if freight.Warehouse == "" { // nolint: staticcheck
+		logger.Debug("Freight is missing Warehouse")
+		return ctrl.Result{}, nil
+	}
+
+	// Migrate the Warehouse field to the Origin field.
+	freight.Origin.Kind = kargoapi.FreightOriginKindWarehouse
+	freight.Origin.Name = freight.Warehouse // nolint: staticcheck
+	freight.Warehouse = ""                  // nolint: staticcheck
```
krancour, line 119: @hiddeco I am thinking we shouldn't blank out `freight.Warehouse`. If some shards are still running older controllers, keeping the value in this field would allow them to continue functioning correctly until they are upgraded. wdyt?

hiddeco: Am fine with this if we ensure some validation is in place that it is in sync with `.origin.name` if present.

hiddeco:
@krancour for the sake of safety, that's probably a good idea.

internal/controller/management/upgrade/freight.go
krancour: That's a good idea. I'll handle that.

internal/kubeclient/indexer.go
```diff
@@ -293,6 +293,9 @@ func FreightByWarehouseIndexer(obj client.Object) []string {
 	if freight.Origin.Kind == kargoapi.FreightOriginKindWarehouse {
 		return []string{freight.Origin.Name}
 	}
+	if freight.Warehouse != "" { // nolint: staticcheck
+		return []string{freight.Warehouse} // nolint: staticcheck
+	}
```
krancour, line 298: This ensures the Warehouse reconciler can find Freight that hasn't been upgraded yet when it goes looking to see if Freight equivalent to one its about to create already exists.

krancour:
@hiddeco I made some very modest amendments to the PR. Mind looking at the last three commits?

I also rebased on release-0.8, btw.

internal/controller/warehouses/warehouses.go
```diff
@@ -222,19 +227,39 @@ func (r *reconciler) syncWarehouse(
...
-				err,
-			)
-		} else if err == nil {
-			logger.Debug(
-				"created Freight",
-				"freight", freight.Name,
-				"namespace", freight.Namespace,
-			)
+		var existingFreight kargoapi.FreightList
+		if err = r.listFreightFn(
```
hiddeco, line 231: Great find.

hiddeco:
Changes all look reasonable to me. Think this is ready to be merged.

krancour (approved):

PR #2315, opened 2024-07-18
feat(api)!: Generalise OIDC Claims

### LINKED ISSUE #2119: Customisable claim name for groups
# Checklist

# Proposed Feature

Add a configurable custom claim name for looking at a nominated claim for roles.

# Motivation

I have a custom claim name in argocd for mapping groups, rather than the default "groups" claim in the token.

# Suggested Implementation

Helm chart configuration for overriding the default claim looked at for group membership roles.

BenHesketh21:
fixes #2119 

This is my attempt to unpack claims as a map[string]any and add them as part of the Kargo Role object allowing operators to use any claim as long as it follows the prefix convention I've set out. A couple of things to mention:

## UI and CLI
I deliberately have done the minimum amount of changes to the CLI and UI, meaning they still specifically refer to sub, emails and groups and don't treat it in a generic way. 
There are a couple of reasons for this, I'm not as confident with TS and React and general UI so I wasn't completely sure on the best way to do or if I could do it at all. 
I also felt like there were already a lot of changes just to get the API to support the different claims just from the annotations. 
If you decide that this won't be added without the UI and CLI changes then no problem, I can make a start on it. Just let me know what you think.

## Thank you

Thank you to @krancour for helping me with the context of the problem, I don't think I would have even been able to get started with this without your help.


charts/kargo/values.yaml
```diff
@@ -198,6 +198,12 @@ api:
     clientID:
     ## @param api.oidc.cliClientID The client ID for the OIDC client used by CLI (optional). Needed by some OIDC providers (such as Dex) that require a separate Client ID for web app login vs. CLI login (`http://localhost`). If Dex is enabled, this value will be ignored and cli client ID will be automatically configured. If Dex is not enabled, and a different client app is configured for localhost CLI login, this should be the client ID configured in the IdP.
     cliClientID:
+    ## @param api.oidc.scopes The scopes to send to the OIDC provider. If Dex is enabled, this value will be ignored and the scopes should be in the dex connectors configuration. If Dex is not enabled, this should be set to the scopes you wish to be provided to Kargo by your identity provider.
+    scopes: 
+      - openid
+      - profile
+      - email
+      - groups
```
krancour, line 206: A few things on this bit...

First, nicely documented!

I might prefer that we make this `additionalScopes` with a default value of `[]` and keep the list that you have here hard-coded. I believe we're going to want these four in nearly all cases, so hard-coding these and only requiring users to list _additional_ ones limits the opportunity for mis-configuration.

The other thing:

> If Dex is enabled, this value will be ignored and the scopes should be in the dex connectors configuration. If Dex is not enabled, this should be set to the scopes you wish to be provided to Kargo by your identity provider.

idk if that sounds right.

You're 100% correct that the Dex connector config should specify desired scopes for the upstream IDP.

But this list is for Kargo to tell _clients_ of the IDP what scopes they ought to ask for. (CLI and UI talk _directly_ to the IDP without Kargo as a middle-man.)

In the case where Dex _is_ the IDP the client will talk to, I think this list is still relevant.

internal/api/oidc/config.go
```diff
@@ -18,7 +18,7 @@ type Config struct {
 	// provider for CLI login.
 	CLIClientID string `envconfig:"OIDC_CLI_CLIENT_ID"`
 	// Scopes are the scopes to be requested during the authorization code flow.
-	Scopes []string
+	Scopes string `envconfig:"OIDC_SCOPES"`
```
krancour, line 21: We don't need to change this to a string. envconfig can unpack a comma-delimited string into a slice. Your struct tags you are using here do not even need to change. It will just work.

krancour: Oh... but per [this comment](https://github.com/akuity/kargo/pull/2315/files#r1688313737) maybe this ought to change to:

```suggestion
        DefaultScopes []string
	AdditionalScopes []string `envconfig:"OIDC_ADDITIONAL_SCOPES"`
```

krancour:
@BenHesketh21 overall this appears to be in great shape. It's big though, so I'll review iteratively. I made a few small suggestions already and as things get resolved, I will circle back to on different parts of this.

BenHesketh21:
@krancour Thank you for the review, your comments make perfect sense, I'll work on those and the lack of test coverage.

internal/api/oidc/config.go
BenHesketh21: That's cool that it just figures that out. I've moved them back to `[]string`s and move the envconfig to be used for additional scopes only.

BenHesketh21:
Fixes are in for your comments, hopefully there should be no more merge conflicts too. I'm figuring out the unit test coverage, I'm struggling to interpret the information codecov is providing but i'll figure it out.

krancour:
> I'm struggling to interpret the information codecov is providing but i'll figure it out.

Codecov is just informational and sometimes feels like an old man yelling at kids to get off his lawn. Just ignore it.

That said, creating or updating tests that validate changes to your satisfaction and ours is important. Just don't get hung up on the number.

I'm taking another pass through this PR right now. We can even leave worries over tests until after everything else appears to be in shape. Stay tuned.

api/rbac/v1alpha1/role_types.go
```diff
@@ -11,9 +11,7 @@ type Role struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 	KargoManaged      bool                `json:"kargoManaged,omitempty" protobuf:"varint,2,opt,name=kargoManaged"`
-	Subs              []string            `json:"subs,omitempty" protobuf:"bytes,3,rep,name=subs"`
-	Emails            []string            `json:"emails,omitempty" protobuf:"bytes,4,rep,name=emails"`
-	Groups            []string            `json:"groups,omitempty" protobuf:"bytes,5,rep,name=groups"`
+	Claims            []Claim             `json:"claims,omitempty" protobuf:"bytes,3,rep,name=claims"`
```
krancour, line 14:
```suggestion
	Claims            []Claim             `json:"claims,omitempty" protobuf:"bytes,7,rep,name=claims"`
```

internal/api/user/user.go
```diff
@@ -26,6 +26,7 @@ type Info struct {
 	// Groups are the group claims obtained from credentials that have been
 	// successfully verified by the server's authentication middleware.
 	Groups []string
+	Claims map[string]any
```
krancour, line 29: Should be be removing the `Subject`, `Email`, and `Groups` fields?

internal/cli/cmd/create/role.go
```diff
@@ -154,9 +154,20 @@ func (o *createRoleOptions) run(ctx context.Context) error {
...
+						},
+						{
+							Name:   "emails",
+							Values: o.Emails,
+						},
+						{
+							Name:   "groups",
+							Values: o.Groups,
+						},
+					},
```
krancour, line 170: We should probably add a `--claim` flag that can be used multipe times with values of the form `<claim name>=<value>`. These should be bound to a new options field `Claims []string`. You can iterate those, parsing the entries to get your `[]rbacapi.Claim`. As a convenience for subs, emails, and groups, which are certain to be the most frequently used types of claims, you can keep those flags and keep them bound to their existing options fields. After walking the `[]string` of claims, to get `[]rbacapi.Claim`, you can walk the `Subs`, `Emails`, and `Groups` fields to amend/override what's in the `[]rbacapi.Claim`.


remove the `--sub`, `--email`, and `--group` flags in favor of a `--claim` flag that can be used multipe times with a value of the form `<claim name>=<value>`. This means the `Subs`, `Emails`, and `Groups` fields on the options struct would go away and be replaced by a `Claims []string` of `<claim name>=<value>` pairs. You can iterate over that, parsing entries to arrive at a `[]rbacapi.Claim`. This way, the CLI will work fine for

internal/cli/cmd/grant/grant.go
```diff
@@ -161,9 +161,20 @@ func (o *grantOptions) run(ctx context.Context) error {
...
+					},
+					{
+						Name:   "emails",
+						Values: o.Emails,
+					},
+					{
+						Name:   "groups",
+						Values: o.Groups,
+					},
+				},
```
krancour, line 177: Same situation as described [here](https://github.com/akuity/kargo/pull/2315/files#r1692023211)

internal/cli/cmd/revoke/revoke.go
```diff
@@ -160,9 +160,20 @@ func (o *revokeOptions) run(ctx context.Context) error {
...
+					},
+					{
+						Name:   "emails",
+						Values: o.Emails,
+					},
+					{
+						Name:   "groups",
+						Values: o.Groups,
+					},
+				},
```
krancour, line 176: Same situation as described [here](https://github.com/akuity/kargo/pull/2315/files#r1692023211)

internal/api/option/auth.go
```diff
@@ -250,18 +250,22 @@ func (a *authInterceptor) listServiceAccounts(
...
+		if claimSlice, ok := claim.([]any); ok {
+			for _, claimSliceItem := range claimSlice {
+				if claimSliceItemString, ok := claimSliceItem.(string); ok {
+					queries = append(queries, libClient.MatchingFields{
+						kubeclient.ServiceAccountsByOIDCClaimIndexField: claimSliceItemString,
+					})
+				}
+			}
+		}
 	}
```
krancour, line 269: These queries don't look right. I think the index is wrong also.

The problem is that we're saying, "look for a ServiceAccount with a claim with a certain value," but we're not taking the _kind_ of claim into account. As it stands now, a user could have sub foo and we would consider them having access to to that SA if the SA had _group_ foo.

I think that when the index is built, the values returned need to be of the form "<claim-kind>/<claim-value>" and when searching the index, input needs to take the same form.

internal/kubeclient/indexer.go
```diff
@@ -498,100 +499,47 @@ func indexStagesByWarehouse(obj client.Object) []string {
...
-			refinedSubjects = append(refinedSubjects, subject)
+	refinedClaimValues := []string{}
+	for _, rawClaimValue := range rawClaimValues {
+		claimValues := strings.Split(rawClaimValue, ",")
+		for _, e := range claimValues {
+			if email := strings.TrimSpace(e); email != "" {
+				refinedClaimValues = append(refinedClaimValues, email)
+			}
 		}
 	}
```
krancour, line 541: This is the other side of [this](https://github.com/akuity/kargo/pull/2315/files#r1692044477)

krancour:
@BenHesketh21 I added a new batch of suggestions, but overall, I am still feeling very good about the direction this is headed.

internal/cli/cmd/create/role.go
BenHesketh21: Makes sense, I've added the flag.

internal/api/user/user.go
BenHesketh21: Yes, sorry, I missed them. I've removed them and fixed all the references.

api/rbac/v1alpha1/role_types.go
BenHesketh21: I'm not completely convinced what the number means so thank you for pointing that out.

internal/cli/cmd/revoke/revoke.go
BenHesketh21: Added that flag as above

internal/cli/cmd/grant/grant.go
BenHesketh21: Added that flag as above

BenHesketh21:
@krancour Sorry it's taken me this long, but I've applied most of the fixes you've suggested/pointed out. I'm working on fixing the index queries now.

internal/api/option/auth.go
BenHesketh21: I think I understand the part about not including the claim name as well as the value could result in finding SAs that it shouldn't. 

I'm a bit confused about the last part about being in the form "/". Do you mean that if we want to include the claim name and value it needs to be separated by "/" e.g. group/foo, sub/foo?

krancour:
@BenHesketh21 no worries. Take your time. And if you need any further help, let me know.

BenHesketh21:
Thank you, if you have any guidance on this https://github.com/akuity/kargo/pull/2315#discussion_r1702684498 that would be great, not sure if you had seen it because of my other comment

internal/api/option/auth.go
krancour: Oh wow. Sorry. Formatting issue because I used angled brackets and GitHub's markdown renderer interpreted them as invalid HTML. You correctly interpreted my intent though. `<claim name>/<claim value>` or something along those lines. `<claim name=claim value>`? You get the idea.

BenHesketh21: Yeah that makes sense, I've modified the indexer to look for  "<claimName>/<claimValue>".

internal/kubeclient/indexer.go
BenHesketh21: Fixed the indexer as above

BenHesketh21:
@krancour I think I've addressed everything in your last batch of suggestions. I've resolved the comments but realised that might not be the proper etiquette and I should've waited for you to see and resolve it, so apologies for that.

krancour:
> I've resolved the comments but realised that might not be the proper etiquette

No worries. I don't think we're always consistent about whether marking an issue is resolved is the prerogative of the user who made the comment or the user who believes they have resolved the issue. Not standing on any ceremony WRT this until/unless it becomes clear we must. So again... no worries. 😄 

I will take another pass through this as soon as time permits.

We really appreciate the effort and thoroughness you've brought to bear on this.

krancour:
@BenHesketh21 just letting you know I haven't forgotten about this.

BenHesketh21:
@krancour No worries, I have some conflicts to fix as well so I'll get on them.

BenHesketh21:
The conflicts should be fixed now

api/rbac/v1alpha1/role_types.go
```diff
@@ -32,8 +30,11 @@ type ResourceDetails struct {
...
+type Claim struct {
+	Name   string
+	Values []string
+}
+
 type UserClaims struct {
-	Subs   []string `json:"subs,omitempty" protobuf:"bytes,1,rep,name=subs"`
-	Emails []string `json:"emails,omitempty" protobuf:"bytes,2,rep,name=emails"`
-	Groups []string `json:"groups,omitempty" protobuf:"bytes,3,rep,name=groups"`
+	Claims []Claim `json:"claims,omitempty" protobuf:"bytes,1,rep,name=claims"`
```
krancour, line 39:
```suggestion
	Claims []Claim `json:"claims,omitempty" protobuf:"bytes,4,rep,name=claims"`
```

internal/api/user/user.go
```diff
@@ -17,15 +17,10 @@ type Info struct {
 	// verified as the Kargo API server's admin user. When this is true, all
 	// other fields should have an empty value.
 	IsAdmin bool
-	// Subject is the unique identified of a non-admin user whose credentials have
+	// Claims is the array of claims from an identity provider of a
```
krancour, line 20:
```suggestion
	// Claims is a map of claims from an identity provider of a
```

api/rbac/v1alpha1/role_types.go
```diff
@@ -32,8 +30,11 @@ type ResourceDetails struct {
 	Verbs        []string `json:"verbs,omitempty" protobuf:"bytes,3,rep,name=verbs"`
 }
 
+type Claim struct {
+	Name   string
+	Values []string
+}
```
krancour, line 36: Are struct tags missing? It seems like they should be here.

api/rbac/v1alpha1/role_types.go
```diff
@@ -32,8 +30,11 @@ type ResourceDetails struct {
...
+
 type UserClaims struct {
```
krancour, line 38: Since this type encapsulates only an array of Claims, I am wondering if we really need both UserClaims _and_ Claims or whether we can drop one of them.

api/v1alpha1/event.go
```diff
@@ -67,15 +67,26 @@ func FormatEventUserActor(u user.Info) string {
...
 }
 
+func IsClaimValueNonEmptyString(claimValue any) bool {
+	if claimString, ok := claimValue.(string); ok {
+		if claimString != "" {
+			return true
+		}
+	}
+	return false
+}
```
krancour, line 88: This looks looks like it might be more straightforward:

```golang
func FormatEventUserActor(u user.Info) string {
	email := u.Claims["emails"].(string) // nolint: forcetypeassert
	subject := u.Claims["subs"].(string) // nolint: forcetypeassert
	switch {
	case u.IsAdmin:
		return EventActorAdmin
	case email != "":
		return EventActorEmailPrefix + email
	case subject != "":
		return EventActorSubjectPrefix + subject
	default:
		return EventActorUnknown
	}
}
```

internal/kubeclient/indexer.go
```diff
@@ -498,100 +496,48 @@ func indexStagesByWarehouse(obj client.Object) []string {
...
-	}
-	groups := strings.Split(rawGroups, ",")
-	refinedGroups := make([]string, 0, len(groups))
-	for _, g := range groups {
-		if group := strings.TrimSpace(g); group != "" {
-			refinedGroups = append(refinedGroups, group)
+	rawClaimValues := map[string]string{}
+	for annotationKey, annotationValue := range sa.GetAnnotations() {
+		if strings.HasPrefix(annotationKey, rbacapi.AnnotationKeyOIDCClaimNamePrefix) {
+			rawClaimName := strings.ReplaceAll(annotationKey, rbacapi.AnnotationKeyOIDCClaimNamePrefix, "")
```
krancour, line 520: This seems a bit more conventional:

```suggestion
			rawClaimName := strings.TrimPrefix(annotationKey, rbacapi.AnnotationKeyOIDCClaimNamePrefix)
```

internal/kubeclient/indexer.go
```diff
@@ -498,100 +496,48 @@ func indexStagesByWarehouse(obj client.Object) []string {
...
-	refinedSubjects := make([]string, 0, len(subjects))
-	for _, s := range subjects {
-		if subject := strings.TrimSpace(s); subject != "" {
-			refinedSubjects = append(refinedSubjects, subject)
+	refinedClaimValues := []string{}
+	for rawClaimName, rawClaimValue := range rawClaimValues {
+		claimValues := strings.Split(rawClaimValue, ",")
+		for _, e := range claimValues {
+			if claimValue := strings.TrimSpace(e); claimValue != "" {
+				refinedClaimValues = append(refinedClaimValues, rawClaimName+"/"+claimValue)
```
krancour, line 536: `rawClaimName+"/"+claimValue` feels a little bit brittle because using this index requires you to know this little bit of the indexer's implementation.

It may be best to create some sort of helper function to at least ensure consistent formatting between the indexer and spots where the index is used.

internal/kubeclient/indexer.go
```diff
@@ -498,100 +496,48 @@ func indexStagesByWarehouse(obj client.Object) []string {
...
-// IndexServiceAccountsByOIDCGroups sets up indexing of ServiceAccounts by
-// their OIDC group annotations.
-//
-// It configures the cluster's field indexer to allow querying ServiceAccounts
-// using the ServiceAccountsByOIDCGroupIndexField selector.
-func IndexServiceAccountsByOIDCGroups(ctx context.Context, clstr cluster.Cluster) error {
-	return clstr.GetFieldIndexer().IndexField(
+// It configures the manager's field indexer to allow querying ServiceAccounts
+// using the ServiceAccountsByOIDCClaimIndexField selector.
+func IndexServiceAccountsByOIDCClaim(ctx context.Context, mgr ctrl.Manager) error {
```
krancour, line 504:
```suggestion
func IndexServiceAccountsByOIDCClaims(ctx context.Context, mgr ctrl.Manager) error {
```

internal/kubeclient/indexer.go
```diff
@@ -498,100 +496,48 @@ func indexStagesByWarehouse(obj client.Object) []string {
...
-	if rawGroups == "" {
-		return nil
-	}
-	groups := strings.Split(rawGroups, ",")
-	refinedGroups := make([]string, 0, len(groups))
-	for _, g := range groups {
-		if group := strings.TrimSpace(g); group != "" {
-			refinedGroups = append(refinedGroups, group)
+	rawClaimValues := map[string]string{}
+	for annotationKey, annotationValue := range sa.GetAnnotations() {
```
krancour, line 518: I know you got this pattern from some of the existing code that you replaced, but it was inefficient and if you wouldn't mind improving it by combining the two for loops in this function, I'd be grateful for that.

internal/kubeclient/indexer_test.go
```diff
@@ -855,7 +855,7 @@ func TestIndexStagesByWarehouse(t *testing.T) {
 	}
 }
 
-func TestIndexServiceAccountsOIDCEmail(t *testing.T) {
+func TestIndexServiceAccountsOIDCClaimWithEmails(t *testing.T) {
```
krancour, line 858: With the indexer in question now generically supporting any claim type, we probably don't need these claim-specific tests. Just one test of that indexer for any kind of claim should suffice.

internal/cli/cmd/revoke/revoke.go
```diff
@@ -67,6 +69,11 @@ kargo revoke --project=my-project --role=my-role --resource-type=stage --verb=up
 # Revoke permission to promote to stage dev from my-role
 kargo revoke --project=my-project --role=my-role \
   --resource-type=stage --resource-name=dev --verb=promote
+
+# Revoke my-role from users with a non standard claim and in a specific group
+# Please take note that even though you can specify subs, emails and groups in the generic claim
+# flag, if both are set, the bespoke sub, group, and email flags take precedence
+kargo revoke --project=my-project --role=my-role --claim=group=admins --claim=given_name=alice
```
krancour, line 76: Honestly, I think I'd drop the existing, dedicated --email, --sub, and --group flags. More options sometimes == more confusion, and in this case, the documentation for the command actually becomes a whole lot more concise by dropping those redundant bits.

Same applies for related commands like grant.

internal/cli/cmd/create/role.go
```diff
@@ -145,6 +154,37 @@ func (o *createRoleOptions) run(ctx context.Context) error {
...
+
+	if o.Emails != nil {
+		claims = append(claims, rbacapi.Claim{
+			Name:   "emails",
+			Values: o.Emails,
+		})
+	}
+
+	for _, claimFlagValue := range o.Claims {
+		claimFlagNameAndValue := strings.Split(claimFlagValue, "=")
```
krancour, line 181: We should probably verify that the split results in exactly two tokens.

My best advice is to test for the presence of exactly one "=" in the `validate()` function.

internal/api/rbac/roles.go
```diff
@@ -713,41 +705,47 @@ func RoleToResources(
...
-func amendClaimAnnotation(sa *corev1.ServiceAccount, key string, values []string) {
-	existing := sa.Annotations[key]
-	if existing != "" {
-		values = append(strings.Split(existing, ","), values...)
-	}
-	slices.Sort(values)
-	values = slices.Compact(values)
-	if sa.Annotations == nil {
-		sa.Annotations = map[string]string{}
+func amendClaimAnnotation(sa *corev1.ServiceAccount, prefix string, values []rbacapi.Claim) {
```
krancour, line 719: A each claim has a name and a value, so referring to an array of claims as `values` promotes a small bit of confusion. This applies to a couple other related functions as well.

```suggestion
func amendClaimAnnotation(sa *corev1.ServiceAccount, prefix string, claims []rbacapi.Claim) {
```

internal/api/rbac/roles.go
```diff
@@ -713,41 +705,47 @@ func RoleToResources(
...
+	for _, claim := range values {
+		existing := sa.Annotations[prefix+claim.Name]
+		if existing != "" {
+			claim.Values = append(strings.Split(existing, ","), claim.Values...)
+		}
```
krancour, line 724: Super-nit: Slightly more idiomatic:

```suggestion
		if existing, exists := sa.Annotations[prefix+claim.Name]; exists {
			claim.Values = append(strings.Split(existing, ","), claim.Values...)
		}
```

Might apply in some related functions as well.

krancour:
@BenHesketh21 I added some more feedback, most of which I think should be pretty easy to resolve. Don't mistake my feedback for any degree of dissatisfaction. This was a big chunk of work you bit off and it's looking really fantastic!

I think some of my feedback may move this PR toward a net reduction in LOC, which is my _favorite_ kind of PR.

BenHesketh21:
@krancour I'll get on those points today, no I definitely don't take it that way, I'm still learning so any feedback is more than welcome. Thank you again for the reviews and help.

api/rbac/v1alpha1/role_types.go
BenHesketh21: Fixed

internal/api/user/user.go
BenHesketh21: Fixed

api/rbac/v1alpha1/role_types.go
BenHesketh21: This was removed from the suggestion below.

api/rbac/v1alpha1/role_types.go
BenHesketh21: Yes makes sense, I have removed the Claim struct and just got the Role to use an array of UserClaims

api/v1alpha1/event.go
BenHesketh21: Yes thats much better, fixed.

internal/kubeclient/indexer.go
BenHesketh21: Fixed

internal/kubeclient/indexer.go
BenHesketh21: I added a function called FormatClaim to the indexer.go, I'll leave this unresolved as I think this is what you mean but I'm not 100% sure if it makes sense to live somewhere else or if I've misunderstood.

internal/kubeclient/indexer.go
BenHesketh21: Fixed

internal/kubeclient/indexer.go
BenHesketh21: Yes, that makes sense, didn't even question it when I did it, I've combined the loops which hopefully makes more sense now.

internal/cli/cmd/revoke/revoke.go
BenHesketh21: Yes makes sense, especially when I had to specify in the examples which takes precedence.

BenHesketh21: Removed the email, group and sub flags.

internal/cli/cmd/create/role.go
BenHesketh21: Yep makes sense, I've added that to the validate functions.

internal/api/rbac/roles.go
BenHesketh21: values is definitely one of those words that just becomes meaningless, so absolutely makes sense. Changed.

internal/api/rbac/roles.go
BenHesketh21: That's much better, fixed.

BenHesketh21:
@krancour I've completed all your suggestions, I left a couple of comments unresolved because I'm not as confident in the fix so thought you'd want to see it.

BenHesketh21:
@krancour Also so close to net 0 LOC😂

api/rbac/v1alpha1/annotations.go
```diff
@@ -6,28 +6,14 @@ const (
...
+	// AnnotationKeyOIDCPrefix is an annotation key prefix that can be set on a
+	// ServiceAccount to associate it with a list of claims from an OIDC
 	// provider. The annotation is used to grant permissions to the
-	// ServiceAccount based on the subject claim from a user who authenticates
+	// ServiceAccount based on the claim that is both provided in the suffix and from a user who authenticates
 	// using OIDC.
+	// e.g. rbac.kargo.akuity.io/claim.subs: '123,456'
 	// The value of the annotation should be a comma-separated list.
-	AnnotationKeyOIDCSubjects = "rbac.kargo.akuity.io/sub"
+	AnnotationKeyOIDCClaimNamePrefix = "rbac.kargo.akuity.io/claim."
```
krancour, line 16: Nit: Just want to suggest some minor wordsmithing of the comment.

> // AnnotationKeyOIDCPrefixis the prefix of an annotation key that can be set on a 
// ServiceAccount to associate it with any user authenticated via OIDC and having
// the claim indicated by the full annotation key with any of the values indicated by
// the annotation. The value of the annotation may be either a scalar string value or a
// comma-separated list.

api/service/v1alpha1/service.proto
```diff
@@ -659,11 +659,15 @@ message GetRoleResponse {
   }
 }
 
+message ListUserClaims {
+  repeated github.com.akuity.kargo.api.rbac.v1alpha1.UserClaim user_claims = 1;
+}
```
krancour, line 664: Is this new message type because `oneof` and `repeated` didn't play nicely together?

BenHesketh21: Yes, like the implementation below in this file. Would you like the comment `// Note: oneof and repeated do not work together` to be added?

krancour: Whoops. Overlooked that.

BenHesketh21: Ahh, so you mean we don't want to be doing it this way?

BenHesketh21: Oh okay, no worries then :)

BenHesketh21: I added the comment, just to be clear as why it exists.

api/rbac/v1alpha1/annotations.go
BenHesketh21: Yes definitely makes more sense, fixed.

BenHesketh21:
@krancour I've completed those last two suggestions, just the couple of unresolved comments that if you could confirm if the fixes need work or not that would be great :)

api/rbac/v1alpha1/annotations.go
```diff
@@ -6,28 +6,12 @@ const (
...
-	AnnotationKeyOIDCGroups = "rbac.kargo.akuity.io/groups"
-
-	// AnnotationKeyOIDCSubjects is an annotation key that can be set on a
-	// ServiceAccount to associate it with a list of subjects from an OIDC
-	// provider. The annotation is used to grant permissions to the
-	// ServiceAccount based on the subject claim from a user who authenticates
-	// using OIDC.
-	// The value of the annotation should be a comma-separated list.
-	AnnotationKeyOIDCSubjects = "rbac.kargo.akuity.io/sub"
+	// AnnotationKeyOIDCPrefixis the prefix of an annotation key that can be set on a
```
krancour, line 9:
```suggestion
	// AnnotationKeyOIDCPrefix is the prefix of an annotation key that can be set on a
```

api/rbac/v1alpha1/role_types.go
```diff
@@ -32,8 +30,7 @@ type ResourceDetails struct {
 	Verbs        []string `json:"verbs,omitempty" protobuf:"bytes,3,rep,name=verbs"`
 }
 
-type UserClaims struct {
-	Subs   []string `json:"subs,omitempty" protobuf:"bytes,1,rep,name=subs"`
-	Emails []string `json:"emails,omitempty" protobuf:"bytes,2,rep,name=emails"`
-	Groups []string `json:"groups,omitempty" protobuf:"bytes,3,rep,name=groups"`
+type UserClaim struct {
+	Name   string   `json:"name,omitempty" protobuf:"bytes,4,opt,name=name"`
+	Values []string `json:"values,omitempty" protobuf:"bytes,5,rep,name=values"`
```
krancour, line 35:
```suggestion
	Name   string   `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
	Values []string `json:"values,omitempty" protobuf:"bytes,2,rep,name=values"`
```

BenHesketh21:
@krancour I have ran the codegen after the changes you committed. I must admit though, I'm stumped by the unit test and lint failures, they are saying that `internal/api/rbac/roles.go:666:2: undefined: sort` but it is definitely imported and when I run `make hack-test-unit` and `make hack-lint` locally I don't get the same errors. Only thing I can think is it was a result of the API codegen that was needed but I'm not sure.

krancour:
@BenHesketh21 don't sweat it. #2453 changed a bunch of sorting-related stuff throughout the entire project. I will straighten out any conflicts.

BenHesketh21:
@krancour oh I see, that makes sense. I don't mind fixing them, I'll just update my branch, I forget that GitHub makes a merge branch and that's what the checks are running against.

BenHesketh21:
@krancour There we go, updated my branch, got the same errors and fixed them.

krancour:
@BenHesketh21, @rbreeze and I teamed up to take this the rest of the way. We may have added a lot of commits, but we didn't deviate much at all from where you'd left off. In total, we just:

* Fixed a few bugs and minor style things
* CLI still needed some attention
* @rbreeze, being the wizard he is, got the UI working along with your back end changes
* The standard claims for email and sub are _singular_, so I fixed those
* Docs were updated accordingly

And we tested this extensively.

Thank you for doing the majority of all the heavy lifting on this! You've earned a big mention in the v0.9.0 release notes!

krancour (approved):

PR #2318, opened 2024-07-19
refactor(controller): minor cleanup of aws managed identity credential helper

krancour:
Similar to a 403 when attempting to get an ECR auth token via an assumed IAM role, I believe that a 403 when attempting to get an ECR token via the controller's own IAM role should be treated more like simply coming up empty-handed and treated less like an error. In this way, the caller can move on to checking other credential helpers.


rbreeze (approved):
LGTM

PR #2346, opened 2024-07-24
fix(promo): honor Argo CD sync windows

### LINKED ISSUE #2078: add docs about argocd-update step and Argo CD sync windows
# Description

Kargo replicates the Argo CD server sync logic, but I believe misses out on the sync window check:

https://github.com/argoproj/argo-cd/blob/ef37b3975406ac58bd527990d247071f38198963/server/application/application.go#L1812-L1814

Kargo should honor the sync window of an AppProject, since users often use this to prevent deploys at a certain time.

# Steps to Reproduce

Promote a Stage backed by an Argo CD application and where the AppProject itself is in a window

# Version

```
v0.6.0
```

rbreeze:
Fixes #2078


krancour:
I had been contemplating Kargo having its _own_ windows for things like freight discovery/production, promotions, etc. so things work uniformly for all gitops agents -- this is supposing we support others in the future. But this might be a good interim solution, so let's still consider it. 👀 tomorrow.

Issue #2338, opened 2024-07-24
Kargo OIDC Azure Entra Id is missing code_challenge_methods_supported


BasJ93:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

When configuring Kargo to use Azure Entra Id as OIDC provider, an error is returned stating that there was an error getting the OIDC configuration. This error is due to the `code_challenge_methods_supported` field missing in the metadata. This is a known shortcoming in the Azure OIDC implementation.

The ArgoCD project had discussion [#16649](https://github.com/argoproj/argo-cd/issues/16649) regarding handling this case, and chose to drop the check in the frontend.

# Steps to Reproduce

Configure Azure as the OIDC provider.

# Version

v0.7.1

# Logs
N/A even with debug level enabled.

# Additional question

Is it an option to drop the check in the frontend like it was done in ArgoCD? Or is it an option to add a config flag that disables the check like proposed in [kubelogin #858](https://github.com/int128/kubelogin/issues/858).

If this can not be fixed, what is the recommended way to integrate with Azure as the OIDC provider?

krancour:
I will take a closer look at this.

> If this can not be fixed, what is the recommended way to integrate with Azure as the OIDC provider?

In the meantime, you might look into whether using Dex as a middle-man improves the situation at all.

BasJ93:
I actually pulled the code, and removed the lines in the front-end that run the check. This allows me to authenticate with Azure Entra Id.

For now I'll run our custom build, but I am interested in seeing the solution.

krancour:
I hadn't initially connected the dots that this was PKCE-related.

Argo CD's support for PKCE was actually modeled after Kargo's and implemented by my colleague @Marvin9.

IDPs not supporting (proper?) support for PKCE, even when they otherwise do support OIDC was one of the main drivers for us offering Dex as an optional middle-man.

I am double-checking right now with my own Entra ID tenant that Dex solves the problem and if that is the case, I would rather suggest that to people using Entra over making Entra-specific concessions in the UI.

I'll circle back with what I find and hopefully will post working connector config for you.

krancour:
With these values overrides on `helm install`, OIDC w/ Entra works flawlessly without any need for code modifications:

```yaml
api:
[...]

PR #2352, opened 2024-07-26
docs: Add sample Dex Azure connector.

BasJ93:
As discussed in #2338 I added the Azure connector sample for Dex to the `values.yaml` file.

### DIFF
--- a/charts/kargo/values.yaml
+++ b/charts/kargo/values.yaml
@@ -291,6 +291,15 @@ api:
       #     clientID: <your client ID>
       #     clientSecret: "$CLIENT_SECRET"
       #     redirectURI: <http(s)>://<api.host>/dex/callback
+      ## Azure Example
+      # - id: microsoft
+      #   name: microsoft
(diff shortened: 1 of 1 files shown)


charts/kargo/values.yaml
```diff
@@ -291,6 +291,15 @@ api:
       #     clientID: <your client ID>
       #     clientSecret: "$CLIENT_SECRET"
       #     redirectURI: <http(s)>://<api.host>/dex/callback
+      ## Azure Example
+      # - id: microsoft
+      #   name: microsoft
+      #   type: Microsoft
+      #   config:
+      #     clientID: <your client ID>
+      #     clientSecret: "$ENTRA_CLIENT_SECRET"
```
krancour, line 300: Just for consistency with the other examples:

```suggestion
      #     clientSecret: "$CLIENT_SECRET"
```

krancour (approved):
I made one tiny change, but this is great @BasJ93. Thank you so much!

PR #2360, opened 2024-07-29
chore: add stack dump for better debugging

loafoe:


hiddeco (commented):
I wonder if it would not be better to configure this by globally enabling recovery from panics (see: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/controller#Options -> `RecoverPanic`), but I have to admit that I lack historical insight here (cc: @krancour).

krancour:
@hiddeco I would prefer that as well. I'm not sure why the current panic recovery is limited to this one spot.

hiddeco:
@loafoe do you feel up for making this suggested change?

It should be introduced at these points: https://github.com/search?q=repo%3Aakuity%2Fkargo%20NewManager&type=code. Where the `ctrl.Options` can take a `Controller` struct, which should have something along the lines of:

```go
ctrlcfg.Controller{
	RecoverPanic: ptr.To(true),
}
```

as a value.

loafoe:
Sure I'll have a look

loafoe:
Sure, I'll have a look

loafoe:
@hiddeco @krancour added the controller Config across. However, the panic test function now fails as the`panic` is not recovered. I lack insights into exactly what is happening. Also, no idea where to add the stack dump code, as that is the original intent of this change. Any further pointers appreciated, thx.

hiddeco:
@loafoe the test fails because the wrapping of the panic now happens at the point where `Reconcile` is called, and is thus no longer captured by the test.

However, for this "special" use case, we shell out to a series of other binaries as part of the promotion mechanics. Given this, it would be OK to keep the panic logic in place to surface the error (and be able to persist it to the Kubernetes API).

With the new addition, we can now guarantee the controller will not end up in a crash loop if we make a mistake anywhere else.

cmd/controlplane/garbage_collector.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/config"
```
hiddeco, line 8: Running `golangci-lint --fix` (or `make lint`) should address this.

loafoe:
> However, for this "special" use case, we shell out to a series of other binaries as part of the promotion mechanics. Given this, it would be OK to keep the panic logic in place to surface the error (and be able to persist it to the Kubernetes API).

Okay, I'll bring the wrapper code back and do some linting, thanks

loafoe:
@hiddeco fixed linter issues and all unit tests pass again

hiddeco (approved):
Thanks for following up with the requested changes! 🙇

hiddeco:
With the upgrade to [controller-runtime v0.19.0](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.19.0) (via #2431), this has now become obsolete as it has been enabled by default upstream.

I want to thank you nonetheless for the work you put in @loafoe. 🙇

PR #2373, opened 2024-08-01
fix: avoid unnecessary cli config changes on re-login

### LINKED ISSUE #2363: avoid unnecessary cli config changes on re-login
When you `kargo login ...` you lose any config settings like default project. That makes sense if the API server you are logging into is _different_ from the one you were previously logged into, but if it's the same, we can improve the UX a little by retaining existing config settings.

hanshal101:
Closes #2363 
Fixes   #2363 

## Description
This PR fixes the unnecessary cli config changes on re-login. It loads the `currentConfig` through `libConfig.LoadCLIConfig()` and compares the `APIAdress` to the config `ServerAddress` if true then the config is set to the `currentConfig`

### DIFF
--- a/internal/cli/cmd/login/login.go
+++ b/internal/cli/cmd/login/login.go
@@ -137,6 +137,11 @@ func (o *loginOptions) run(ctx context.Context) error {
 	var bearerToken, refreshToken string
 	var err error
 
+	currentConfig, err := libConfig.LoadCLIConfig()
+	if err != nil {
+		return fmt.Errorf("error loading current configuration: %w", err)
+	}
+
 	switch {
 	case o.UseAdmin:
 		for {
@@ -175,6 +180,11 @@ func (o *loginOptions) run(ctx context.Context) error {
 		// instead. Since we plan not to use the refresh token for such a case,
 		// it's more secure to throw it away immediately.
 		refreshToken = ""
+	}
+
+	// To retain the existing user if he's logging into the same server
+	if currentConfig.APIAddress == o.ServerAddress {
+		o.Config = currentConfig
 	}
 
 	if err = libConfig.SaveCLIConfig(


internal/cli/cmd/login/login.go
```diff
@@ -137,6 +137,11 @@ func (o *loginOptions) run(ctx context.Context) error {
 	var bearerToken, refreshToken string
 	var err error
 
+	currentConfig, err := libConfig.LoadCLIConfig()
+	if err != nil {
+		return fmt.Errorf("error loading current configuration: %w", err)
+	}
+
```
krancour, line 144: This actually shouldn't be necessary. `o.Config` _already_ has the current configuration.

```suggestion
```

internal/cli/cmd/login/login.go
```diff
@@ -177,6 +182,11 @@ func (o *loginOptions) run(ctx context.Context) error {
 		refreshToken = ""
 	}
 
+	// To retain the existing user if he's logging into the same server
```
krancour, line 185: Gender-neutral language is more inclusive.

```suggestion
	// To retain the existing user if they're re-logging into the same server
```

internal/cli/cmd/login/login.go
```diff
@@ -177,6 +182,11 @@ func (o *loginOptions) run(ctx context.Context) error {
...
+	if currentConfig.APIAddress == o.ServerAddress {
+		o.Config = currentConfig
+	}
+
```
krancour, line 189: As noted [here](https://github.com/akuity/kargo/pull/2373/files#r1702128376), `o.Config` _already_ contains the current config, which means this block never has any effect.

```suggestion
```

krancour:
@hanshal101 thank you for the PR. Did you test this at all? The changes you propose do not have the desired effect -- in fact they have _no effect at all on the functionality_, so apart from being ineffective, they're also completely extraneous.

krancour:
Closing this. Please feel free to open a new PR that implements the functionality requested in #2363

Issue #2271, opened 2024-07-09
oidc server specified, certificate required to validate oidc server, where to specify certificate in helm chart?


lknite:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

- ability to specify certificate needed to verify oidc server via helm chart

# Motivation

- after specifying oidc server, client_id, client_secret, etc ... unable to login because certificate verification fails

# Suggested Implementation

implement two methods which may be used:
1. allow a ca-bundle to be provided via a secret:
    - api.oidc.caBundle.secretName
    - api.oidc.caBundle.key
2. allow a ca-bundle to be provided in-line
    - api.oidc.caBundle
- allow the ca-bundle to be one or more certificates

krancour:
@lknite are you using Dex?

lknite:
no, am using keycloak

krancour:
And the failure is occurring right when the API server starts?

lknite:
yes, the pod never becomes ready cause it detects the cert cannot be verified during startup

krancour:
Ok. Got it. I am thinking back to when we first did the ODIC integration and trying to recall why I wouldn't have made provisions for a custom CA bundle right from the get go...

My vague recollection is that the possibility of identity providers _without_ a cert that is trusted by users' browsers didn't seem realistic. After all, if one doesn't trust the IDP's cert without using a custom CA bundle, neither does the other.

Are your users installing a custom CA bundle into their browsers? Or are they just clicking through the warnings?

I guess what it all boils down to was an assumption that, with the possible exception of the included Dex, IDPs being properly secured with widely-trusted certificates was a given.

I'm open to that assumption being challenged. Can you tell me more about your use case?

lknite:
Vault is used via cert-manager to generate certificates within all clusters.  The ca-bundle root certificates and intermediate certificates are pushed out to client systems automatically so that clients can access trust applications running via kubernetes.

Tanzu is used to stand up kubernetes clusters, and there is a place to specify the vault ca-bundle so that it also reaches workernodes of kubernetes clusters.

[...]

PR #2382, opened 2024-08-05
feat(controller,api): adds a cabundle option to the values.yaml file.

lknite:
If enabled, it adds an initContainer to the api/controller deployment which mounts certs provided via either a configMap or secret.

Uses an emptyDir, or optionally a pvc, for the actual certs mount within the controller; this allows the controller to write to the certs folder if needed.


krancour:
What is the point of the init container and the volumes here? I'm not clear about what would be insufficient about _just_ mounting a user-provided secret into the container that contains the trust store.

krancour:
Following up on my previous comment:

The original issue was #2271 and it seemed to be about CA for the OIDC IDP, so in the back of my mind, that was the context/scope. _That_ issue can be solved by adding more CA certs directly to the http client that the Kargo _API server_ uses to communicate with the IDP and that doesn't require anything more than those certs to be mounted somewhere where the API server will look for them. e.g. Secrets do the trick. No init container needed. Very small code change needed.

This PR seems to address the issue of custom CA certs more-broadly, at the system level, which is nice, but also seems to do it for the _controller_ and not the API server, meaning it doesn't solve the original issue.

Perhaps what is required here is some clarification on the scope of what you are trying to solve for.

krancour:
@lknite, ok I get where you were going. It was the scope / area where you implemented the change that confused me a bit. With greater clarity on that now, I will take another look.

lknite:
added to kargo-api as well, now will resolve issue https://github.com/akuity/kargo/issues/2271

lknite:
tried to update other files on my fork, i might have broken things will fork again and redo my changes

lknite:
tried to update my fork, might have messed up things will redo with a new pull request

PR #2388, opened 2024-08-05
fix(cli): retain config changes on re-login

walsm232:
Fixes: https://github.com/akuity/kargo/issues/2363

### DIFF
--- a/internal/cli/cmd/login/login.go
+++ b/internal/cli/cmd/login/login.go
@@ -177,14 +177,16 @@ func (o *loginOptions) run(ctx context.Context) error {
 		refreshToken = ""
 	}
 
-	if err = libConfig.SaveCLIConfig(
-		libConfig.CLIConfig{
-			APIAddress:            o.ServerAddress,
-			BearerToken:           bearerToken,
-			RefreshToken:          refreshToken,
-			InsecureSkipTLSVerify: o.InsecureTLS,
-		},
-	); err != nil {
+	if o.Config.APIAddress != o.ServerAddress {
+		o.Config = libConfig.CLIConfig{}
+	}
+
+	o.Config.APIAddress = o.ServerAddress
+	o.Config.BearerToken = bearerToken
+	o.Config.RefreshToken = refreshToken
+	o.Config.InsecureSkipTLSVerify = o.InsecureTLS
+
+	if err = libConfig.SaveCLIConfig(o.Config); err != nil {
 		return fmt.Errorf("error persisting configuration: %w", err)
 	}
 	return nil


internal/cli/cmd/login/login.go
```diff
@@ -177,14 +177,13 @@ func (o *loginOptions) run(ctx context.Context) error {
...
-			BearerToken:           bearerToken,
-			RefreshToken:          refreshToken,
-			InsecureSkipTLSVerify: o.InsecureTLS,
-		},
-	); err != nil {
+	// Merge new settings with the existing configuration
+	o.Config.APIAddress = o.ServerAddress
+	o.Config.BearerToken = bearerToken
+	o.Config.RefreshToken = refreshToken
+	o.Config.InsecureSkipTLSVerify = o.InsecureTLS
```
krancour, line 184: This looks very close, but I think that if the server we're logging into isn't the same as the one we were previously logged into, then we _don't_ want to merge because we _will_ want to lose any other server-specific settings in the config file at that point, e.g. default project.

Could we conditionally merge or start with a new config from scratch?

walsm232: Ah I think I see what you mean, I will adjust this now and hopefully fix it if I am understanding correctly

walsm232: Adjusted the PR now

krancour: Looks great! Thanks!

internal/cli/cmd/login/login.go
```diff
@@ -177,14 +177,20 @@ func (o *loginOptions) run(ctx context.Context) error {
...
 	}
 
-	if err = libConfig.SaveCLIConfig(
-		libConfig.CLIConfig{
+	if o.Config.APIAddress == o.ServerAddress {
+		o.Config.BearerToken = bearerToken
+		o.Config.RefreshToken = refreshToken
+		o.Config.InsecureSkipTLSVerify = o.InsecureTLS
+	} else {
+		o.Config = libConfig.CLIConfig{
```
krancour, line 185:
```suggestion
	// If re-logging into the same API server, update the following, but leave other settings alone
	if o.Config.APIAddress == o.ServerAddress {
		o.Config.BearerToken = bearerToken
		o.Config.RefreshToken = refreshToken
		o.Config.InsecureSkipTLSVerify = o.InsecureTLS
	} else { // Otherwise, completely replace the configuration
		o.Config = libConfig.CLIConfig{
```

internal/cli/cmd/login/login.go
krancour: Actually... my apologies... this works, but something like the following pseudocode may be more efficient and easier to read and maintain:

```
if not same server:
  cfg = new config
cfg.apiAddress = apiAddress
cfg.bearerToken = bearerToken
# etc...
```

walsm232: Agreed, that would definitely be easier to read. I've updated PR now, hopefully it makes sense - let me know if you think there's any comment(s) needed on it or if it's self-explanatory

krancour (approved):
LGTM @walsm232. Solid first contribution to the project. Thank you!

PR #2389, opened 2024-08-05
feat(ui): move image processing to backend 

rbreeze:
Additionally, when a user creates a project, the images table will now be hidden by default. Once they create a Warehouse and promote a Stage, it will be shown again by default, at which point they can show / hide and it the preference will be stored in LocalStorage as it is now.


internal/api/list_stages_with_images_v1alpha1.go
```diff
@@ -0,0 +1,73 @@
...
+
+	stages := make([]*kargoapi.Stage, len(list.Items))
+	images := make(map[string]*svcv1alpha1.TagMap)
+
+	for idx, stage := range list.Items {
+		stages[idx] = &list.Items[idx]
+
+		for i, freightGroup := range stage.Status.FreightHistory {
+			for _, freight := range freightGroup.Freight {
+				for _, image := range freight.Images {
```
rbreeze, line 40: This level of nesting is far from ideal, but barring some larger refactor (like storing images in a top level field of `StageSpec` for example), this is the only option. Previously, very similar logic was performed by the frontend.

Some context makes this less scary:
- Users typically have on the order of 10 Stages per project. Hundreds would be quite rare.
- `stage.Status.FreightHistory` is by default a list of the 10 latest freight. This maximum is not even user-configurable at the moment, so this particular loop can be considered O(1)
- The size of `freightGroup.Freight` corresponds to the number of `requestedFreight` slots in a Stage. This too will typically be a handful of freight at most
- `freight.Images` also rarely has more than a handful of images

In situations where assumptions 1, 3, or 4 are violated, my guess is users would run into _other_ performance bottlenecks before this API becomes a problem

Makefile
```diff
@@ -60,10 +60,17 @@ endif
 .PHONY: lint
 lint: lint-go lint-proto lint-charts lint-ui
 
+.PHONY: format
+format: format-go format-ui
+
```
krancour, line 65: This is a nice addition. I am curious if something specific prompted this. I take it for granted that our editors format things appropriately, but with an uptick in community contributions, we're probably seeing a wider variety of editors, editor plugins, etc. It's nice to have an editor-agnostic final arbiter of correct formatting.

I wonder if we should consider putting a formatting check into the CI process. We do something similar already for codegen. We re-run codegen and check that it produces no discrepancies with the generated code in the PR. We could tackle that in a follow-up, of course.

rbreeze: It was actually something I've been meaning to add for a long time, and ended up scratching the itch while working on this PR 😄 

For the UI, we already have a formatting check in CI ✅ . I thought there was one for the Go code too, but maybe I'm misremembering

krancour: @rbreeze we lint Go code in CI, but I'm not positive the linter catches all formatting issues.

In general, lint == stop you from making common mistakes, but that sometimes does overlap with formatting issues.

krancour:
@rbreeze thinking about this more and I think this endpoint is a little odd in that it returns something another endpoint already returns (list of Stages) _and_ supplementary information used by the UI only.

If you don't mind making two API calls from the UI instead of one, I think it would be tidier if the UI continued to call out to the original list Stages endpoint and then made a second call to a new endpoint that returns just the image info.

internal/api/list_images.go
```diff
@@ -0,0 +1,72 @@
+package api
```
krancour, line 1: Just to stick with the current naming convention for files in this package, the filename should probably be `list_images_v1alpha1`.

krancour: This is my only feedback. Pre-emptively LGTM'ing.

krancour (approved):

internal/api/list_images.go
rbreeze: 🤦🏻 my intention was to name it that way... didn't have enough coffee yesterday 😆

PR #2391, opened 2024-08-06
feat(controller): allow GitHub App as authn method for image repos in ghcr

### LINKED ISSUE #2383: GitHub App as Image Registry-Specific Authentication Option
# Checklist

# Proposed Feature

Support GitHub App auth as Image Registry-Specific Authentication Option

# Motivation

Some users might be using GHCR as their image registry, currently it is only possible to use Kargo with public ghcr.io or use username/password which isn't a best practice for machine to machine auth

Dhouib-Mohamed:
Fixes #2383 
Accepting credentials.TypeImage in the getCredentials condition
Adding Necessary tests linked to TypeImage to validate te changes

### DIFF
--- a/internal/credentials/kubernetes/github/app_test.go
+++ b/internal/credentials/kubernetes/github/app_test.go
@@ -49,7 +49,25 @@ func TestAppCredentialHelper(t *testing.T) {
 			},
 		},
 		{
+			name:     "cred type is git",
+			credType: credentials.TypeGit,
+			helper:   &appCredentialHelper{},
+			assertions: func(t *testing.T, creds *credentials.Credentials, _ *cache.Cache, err error) {
+				require.NoError(t, err)
+				require.Nil(t, creds)
+			},
+		},
+		{
 			name:     "secret is nil",
+			credType: credentials.TypeImage,
+			helper:   &appCredentialHelper{},
+			assertions: func(t *testing.T, creds *credentials.Credentials, _ *cache.Cache, err error) {
+				require.NoError(t, err)
+				require.Nil(t, creds)
+			},
+		},
+		{
+			name:     "secret is nil - git",
 			credType: credentials.TypeGit,
 			helper:   &appCredentialHelper{},
 			assertions: func(t *testing.T, creds *credentials.Credentials, _ *cache.Cache, err error) {
@@ -59,6 +77,16 @@ func TestAppCredentialHelper(t *testing.T) {
 		},
 		{
 			name:     "no github details provided",
+			credType: credentials.TypeImage,
+			secret:   &corev1.Secret{},
+			helper:   &appCredentialHelper{},
+			assertions: func(t *testing.T, creds *credentials.Credentials, _ *cache.Cache, err error) {
+				require.NoError(t, err)
+				require.Nil(t, creds)
+			},
+		},
+		{
+			name:     "no github details provided - git",
 			credType: credentials.TypeGit,
 			secret:   &corev1.Secret{},
 			helper:   &appCredentialHelper{},
@@ -69,6 +97,20 @@ func TestAppCredentialHelper(t *testing.T) {
 		},
 		{
 			name:     "app ID missing",
+			credType: credentials.TypeImage,
+			secret: &corev1.Secret{
+				Data: map[string][]byte{
+					installationIDKey: []byte(testInstallationIDStr),
+					privateKeyKey:     []byte(testPrivateKey),
+				},
+			},
+			helper: &appCredentialHelper{},
+			assertions: func(t *testing.T, _ *credentials.Credentials, _ *cache.Cache, err error) {
+				//require.ErrorContains(t, err, "must all be set or all be unset")
+			},
+		},
+		{
+			name:     "app ID missing - git",
 			credType: credentials.TypeGit,
 			secret: &corev1.Secret{
 				Data: map[string][]byte{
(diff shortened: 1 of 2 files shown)


internal/credentials/kubernetes/github/app_test.go
```diff
@@ -48,8 +48,26 @@ func TestAppCredentialHelper(t *testing.T) {
...
 		},
+		{
+			name:     "cred type is git",
+			credType: credentials.TypeGit,
+			helper:   &appCredentialHelper{},
+			assertions: func(t *testing.T, creds *credentials.Credentials, _ *cache.Cache, err error) {
+				require.NoError(t, err)
+				require.Nil(t, creds)
+			},
+		},
```
krancour, line 59: I think the number of new test cases is not entirely necessary. We mainly utilize "white-box testing," which means the tests are written with enough awareness of the implementation such that the test cases are engineered to force execution down as many paths in the code as possible. The flip side of this is that it sometimes reduces the number of cases required to test the code thoroughly. In this particular situation, we know that `credType` isn't used anywhere other than in the one conditional you have updated -- meaning that once that conditional is cleared, the value of `credType` has no further bearing on the outcome of any given test case. i.e. Test cases `no github details provided` and `no github details provided - git` are effectively redundant.

I believe, where tests are concerned, you only need two small changes _instead of all these new cases_: Change the name `cred type is not git` to `cred type is not supported` and in that test case, change `credType: credentials.TypeGit,` to `credType: credentials.TypeHelm`.

Note: GHCR _can_ store Helm charts, meaning a GitHub App _should_ be a viable authentication method for a Helm chart stored in GHCR, but I think that can be dealt with in a follow-up PR because it is slightly more involved than this change was.

krancour:
@Dhouib-Mohamed there's still about 81 more changed lines in this PR than I was expecting...

Dhouib-Mohamed:
I created the cache hit and the cache miss testes for both Git and Image because i maybe depends on the type of the credential since it is not just checking the inputs like secret and app id and it contains business logic
Should i createb one case only for git for those as well?

krancour:
@Dhouib-Mohamed if you look at how the cache keys are constructed, the repo type has no bearing on it.

internal/credentials/kubernetes/github/app_test.go
```diff
@@ -40,7 +40,25 @@ func TestAppCredentialHelper(t *testing.T) {
...
+		{
+			name:     "cred type is image",
```
krancour, line 61: These two new cases are still not needed. I know this may seem counter-intuitive, but remember this is white-box testing...

The previous case (cred type not supported) tests for an early return if the cred type is not supported. Subsequent cases only require the cred type _to be a supported one_ so that we avoid that early return and move on to testing other outcomes. Again, with white-box testing, we have the luxury of knowing the implementation and we know that once we're past that potential early return, the cred type has no further bearing on the results.

krancour:
@Dhouib-Mohamed the changes look good. Thank you for your patience with the review process.

Did you happen to test this end-to-end? i.e. Take it for a test drive?

Dhouib-Mohamed:
I tested a basic scenario (creating a credential from the ui) but i am not sure if i covered the necessary use cases
To be frank even now i am not sure about the usage of this function since this is my first interaction with this project
If possible can you give an example of a full use case to test?

krancour:
@Dhouib-Mohamed this code doesn't really have anything to do with creating creds via the UI. This has to do with the back end retrieving temporary access tokens associated with a GitHub App.

If I were testing this, I would write a small, standalone program that instantiates this credential helper and I would pass it a Kubernetes Secret I created myself, programmatically so test whether it returns the temporary token. If it does, I would copy that and test your ability to use that token to log into the Docker CLI and pull a private image from ghcr.

Dhouib-Mohamed:
Got it
Thanks for clearing that out for me
I will go ahead and test it and make any necessary changes

Dhouib-Mohamed:
@krancour I have made the necessary tests and i have concluded that even with adding packages permissions to the Github app it still return a Not Authorized error when trying to interact with Github Packages ( images ).
This is a discussion found explaining the problem : https://github.com/orgs/community/discussions/34324

That's why i think that this PR should be closed for now.

krancour:
@Dhouib-Mohamed that discussion indeed looks like it doesn't work in our favor here. I'm going to a test a little as well and report back.

One thing that I expect is a factor here -- and so I want to ask if you took it into account -- if the App has `packages:read` permission and is installed into a Git repo (and not installed org-wide) then the package (container image) in question _must_ "belong" to that Git repo. (This really isn't the most intuitive thing, honestly. I believe that if this _does_ work, this is probably where most people go wrong.)

So:

1. Did you install at the Git repo level? Or org level?
2. If at the Git repo level, are you certain the package belonged to that Git repo?

As I said, I will experiment as well, but it _is_ quite possible that you are correct and this isn't doable at the moment.

Dhouib-Mohamed:
@krancour So i tested on Git Repo level and on organization level and they both failed
On Git Repo lever i made sure that the package belong the repo ( linked to the repo ) and i tried to link the github app to all the repos or to a specefic repo and they both failed

Do you want me to share the code i created to test this out to simplify your testing?

krancour:
@Dhouib-Mohamed I appreciate your thoroughness! I don't need the code... I think I can do that pretty quickly. It's the setup on the GitHub end of things that is the more time consuming part. 🤦‍♂️

krancour:
@Dhouib-Mohamed thank you so much for the non-trivial amount of effort you put into this. I can independently confirm your findings. Despite all appearances (a package permissions option when configuring a GitHub App), an App installation token simply cannot enable access to a private ghcr.io package repository. ☹️ 

I will close this PR for now, but it is my hope that someday this _will_ work and this PR can be revived.

PR #2393, opened 2024-08-06
feat(ui): ability to collapse all unused freight

### LINKED ISSUE #2386: Option to only show running Freight in timeline
# Proposed Feature

One simple UI feature I thought of while filing https://github.com/akuity/kargo/issues/2385, was a UI option to only show running freight in the UI.

# Motivation

User may not care to see non-running freight, or there is just so much freight to sift through.

# Suggested Implementation

Show a checkbox or toggle to only show running freight.

rbreeze:
Fixes #2386

CleanShot 2024-08-06 at 12 02 50@2x
CleanShot 2024-08-06 at 12 02 42@2x
CleanShot 2024-08-06 at 12 03 07@2x


krancour:
Seeing some issues with this:

1. When I collapse unused, the box that represents the unused Freight says "Old Freight," which is somewhat misleading, since hiding old Freight is a separate feature.

2. When I hide Old Freight, it's not hiding _all_ the old Freight, but just one. The count that it shows in the box that represents the old Freight is off -- reflecting the number that is _should_ have hidden.

N.B.: Don't be fooled by version numbers in this video. I manually created Freight containing references to older images. The chronological order in the Freight timeline is correct.

https://github.com/user-attachments/assets/3663d104-894c-4c4a-a503-a42cd1d135c0

rbreeze:
@krancour Thanks for finding the bug with the old freight collapsing improperly. Just pushed a fix.

WRT the "old freight" label, I chose to label the oldest set of collapsed freight as "old" no matter which collapse mode is chosen. My thinking is that no matter whether all unused freight are hidden or not, that last set is always "old". I can certainly see how this could be confusing though, so I'm happy to change it. WDYT? 

Here's an example of all freight collapsed with the old freight label for the last group:
Screenshot 2024-08-21 at 18 22 33

krancour:
> I chose to label the oldest set of collapsed freight as "old" no matter which collapse mode is chosen

That makes sense!

krancour (approved):

PR #2395, opened 2024-08-06
feat: skip authentication if running kargo server locally

### LINKED ISSUE #1834: ui: option to skip login page
This is related to #1728

When a user runs `kargo server`, they should be able to use the UI served from there without further authn. (The API server will already be running using the user's own kubeconfig.)

Currently, if a server is configured with _neither_ the admin account enabled _nor_ OIDC enabled, the login page says logins are disabled and you should contact the system administrator.

The UI determines whether admin account, ODIC, both, or neither are supported by making a call to the public config endpoint. This is an endpoint that requires no authn and basically advertises the API server's capabilities.

I'd love if the results from that endpoint returned a new option that says no authn is desired. We'd only set this when a launching a server locally from the CLI.

If the UI could respect that new option, causing the login screen to be completely bypassed, that would be awesome.

@rpelczar @rbreeze lmk what you guys think about this.

__Edit:__ When running in this manner, the UI's API client would also not need to pass credentials of any kind to the local API server.

rbreeze:
Fixes #1834


internal/api/config/config.go
```diff
@@ -19,7 +19,7 @@ type StandardConfig struct {
 
 type ServerConfig struct {
 	StandardConfig
-	LocalMode                   bool
+	LocalMode                   bool // this will never be set from the environment
```
krancour, line 22: I think a little more detail here may help.

```suggestion
	// LocalMode is true if the server is running as a non-containerized process
	// that was launched via `kargo server`, which uses the kargo CLI user's
	// current kubeconfig for all cluster interactions.
	LocalMode                   bool // this will never be set from the environment
```

krancour:
@rbreeze it seems I get caught in a bit of a, not-infinite, redirect loop.

Steps I used:

```
# Build UI to be embedded into CLI binary
cd ui
pnpm install
NODE_ENV=production pnpm run build
cd ..
rm -rf internal/api/ui
mv ui/build internal/api/ui

# Start server
go run cmd/cli/*.go server
```

Visit the URL displayed in the server's startup message.

When you enter the loop, you should be able to see the public config endpoint being called over and over again in the server logs.

Not really sure what's going on, but so far, I don't _think_ it's back end.

krancour (approved):
Remaining issues fall more squarely under #1728, so this PR does exactly what was required.

LGTM

Issue #2125, opened 2024-06-05
ui: make show/hide warehouse subs button show/hide only subs for one warehouse


krancour:
Currently if you click the show/hide subscriptions button on any Warehouse, it shows/hides the subscriptions for _all_ Warehouses.

I'm not classifying this as a bug, because I know this was by design, but @rbreeze and I discussed and agreed it might be seen as surprising behavior.

PR #2211, opened 2024-06-27
refactor(ui): hide subscriptions button should be per warehouse

rbreeze:
Fixes #2125

PR #2396, opened 2024-08-06
fix(ui): hide subscriptions per warehouse in DAG

rbreeze:
Fixes https://github.com/akuity/kargo/issues/2125


krancour:
This still has the same problems that we encountered with #2211.

When you have multiple Warehouses, no matter what hide button you click, the subscriptions are hidden from bottom to top and no matter what show button you click, the subscriptions are revealed from top to bottom.

rbreeze:
@krancour and I discussed offline. I changed the mechanism by which this hiding works to solve the problem: before, the approach was to remove the subscription nodes _before_ the DAG was calculated. Now, we render the DAG the same regardless of which subscriptions are hidden, and simply don't render the hidden nodes. Before, the DAG jumped around, and now the DAG is stable.

https://github.com/user-attachments/assets/7cca3452-f860-4afa-9564-965da1869cab

krancour:
Thanks @rbreeze. I tried it out and it works great. The only potential issue I see is that once all the subscriptions are hidden, since they're merely invisible, they're still taking up screen real estate, which may be irksome for some longer DAGs.

I _think_ you were working on doing the DAG differently anyway and this fix is just an interim one? If that's the case, this LGTM.

krancour (approved):

rbreeze:
@krancour I agree, that would be an annoyance for sure. 

You're correct, I'm currently working on rewriting the DAG rendering to take place on the backend, so I can certainly fix this issue as part of that overhaul.

PR #2399, opened 2024-08-06
feat(controller): acr short-lived token support

### LINKED ISSUE #2372: ACR short-lived token support
# Checklist

# Proposed Feature

Adding the ability for "password-less", short-lived token support for freight (OCI Helm charts + Docker images) in ACR. 

# Motivation

This already exists for ECR and GCR repositories, so IMO it would make sense to also have the same for ACR.

# Suggested Implementation

There are some differences between how Azure does IAM and how other clouds handle IAM. Whereas with AWS IAM we derive the IAM role name from the Kargo project name, and with GCP the service account name is also derived from the Kargo project name, doing this in Azure becomes a whole lot trickier.

Everything in Azure is done on the basis of a `ClientID`, so in order to do what we really want to do (= get a short-lived token for ACR) we would first need to lookup the correct `ClientID` by first finding a [Managed Identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-manage-user-assigned-managed-identities?pivots=identity-mi-methods-azp) matching the correct name. Before we get to that point though, we would need to first retrieve a managed identity client that can retrieve said managed identity, and for _that_ we would to get the client by looking it up by a subscription id and resource group name. While all of this _could_ be done, I'd like to propose an alternative.

Instead of doing the steps above, I'd like to instead propose the federation of service account tokens to Azure Managed Identities. It is already the case that every Kargo project corresponds to a Kubernetes namespace, and each of those namespaces currently has two service accounts - `kargo-viewer` and `kargo-admin`. For this example I'm using `kargo-viewer` as the example service account to federate with, but this could potentially be `kargo-admin` or another serviceaccount altogether custom made for token creation.

For a user to take advantage of this, the steps would be to

1. [Create a user-managed-identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-manage-user-assigned-managed-identities?pivots=identity-mi-methods-azp#create-a-user-assigned-managed-identity)
2. [Federate the managed identity](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-create-trust-user-assigned-managed-identity?pivots=identity-wif-mi-methods-azp#kubernetes-accessing-azure-resources) to the relevant Kargo namespace and Kargo service account.
3. Give the relevant permissions (`AcrPull`) to the managed identity in the relevant Azure repositories
4. ???

We still have the issue of assigning the `ClientID` (as well as the `TenantID`) to the correct Kargo project. There are a few ways in which we could do this, but I think the one which would require least work is to retrieve those values from the Kargo project directly. For that I'd like to propose the addition of two annotations to be used for this purpose:

* `azure.workload.identity/client-id` (mandatory for this to work)
* `azure.workload.identity/tenant-id` (optional, fallback would be whatever is set on the Kargo controller pod)

The Kargo controller gets the `ClientID` and `TenantID` from the Kargo project CR and assuming all is good there, starts the OIDC dance:

1. It does a lookup of the relevant Kargo serviceaccount (`kargo-viewer` for this example)
2. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to create a temporary token for said serviceaccount
3. Uses that token to exchange it for an Azure AD Token
4. Exchanges the Azure AD token to retrieve an ACR token scoped for the relevant repo, which is valid for 3h.

There are a few implications in regards to K8s RBAC for the kargo-controller:

* We'd need to grant `read` access for `serviceaccounts` (as well as `list` and `watch` due to the serviceaccount informers requiring that access)
* We'd also need to grant access to the `TokenRequest` API, with `create` privileges for `serviceaccounts/token`.

We'd need to ensure that this is something that's only allowed within `kargo-project` namespaces.

blakepettersson:
Fixes #2372 

Needs a lot of TLC


internal/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,330 @@
...
+	"github.com/akuity/kargo/internal/logging"
+)
+
+const (
+	accessTokenUsername  = "00000000-0000-0000-0000-000000000000"
+	metadataURL          = "http://169.254.169.254/metadata/instance?api-version=2021-02-01"
+	viewerRoleName       = "kargo-viewer"
+	AzureDefaultAudience = "api://AzureADTokenExchange"
+	AnnotationClientID   = "azure.workload.identity/client-id"
+	AnnotationTenantID   = "azure.workload.identity/tenant-id"
```
krancour, line 34: I think we are best off getting tenantID from values.yaml at install time. Similar for EntraID (nee AAD) endpoint URL. I _think_ we can entirely forget about the metadataURL and the steps to retrieve Azure environment if tenantID and EntraID endpoint URL are explicitly configured at install time.

internal/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,330 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/credentials"
+	"github.com/akuity/kargo/internal/logging"
+)
+
+const (
+	accessTokenUsername  = "00000000-0000-0000-0000-000000000000"
```
krancour, line 29: Is the username allowed to be anything?

internal/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,330 @@
...
+	clientID, ok := proj.GetAnnotations()[AnnotationClientID]
+	if !ok {
+		return nil, fmt.Errorf("project is missing annotation: %s", AnnotationClientID)
+	}
+	tenantID, ok := proj.GetAnnotations()[AnnotationTenantID]
+	if !ok {
+		return nil, fmt.Errorf("project is missing annotation: %s", AnnotationTenantID)
+	}
+
+	repoHost, err := url.Parse(repoURL)
```
krancour, line 130: `repoHost` isn't the host. It's a `url.URL`, so I would perhaps reconsider the naming of this var to prevent confusion.

internal/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,330 @@
...
+
+	w := &workloadIdentityCredentialHelper{
+		tenantID:    tenantID,
+		clientID:    clientID,
+		kargoClient: kargoClient,
+		azureEnv:    &env,
+		tokenCache: cache.New(
+			// Access tokens live for three hours. We'll hang on to them for 40 minutes.
+			40*time.Minute, // Default ttl for each entry
+			time.Hour,      // Cleanup interval
```
krancour, line 74: If the tokens live for three hours, I would hang on to them for longer than 40 mins. 2.5 hours seems decent?

internal/credentials/kubernetes/acr/workload_identity.go
blakepettersson: No, from what I can see in the [docs](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#az-acr-login-with---expose-token) the user name needs to be exactly that string when using short-lived tokens.

internal/credentials/kubernetes/acr/workload_identity.go
blakepettersson: > I think we can entirely forget about the metadataURL and the steps to retrieve Azure environment if tenantID and EntraID endpoint URL are explicitly configured at install time.

~That is true, although I don't know how easy it is for a user to find the EntraID endpoint URL...~ EDIT: Nevermind, the endpoints can be found [here](https://learn.microsoft.com/en-us/entra/identity-platform/authentication-national-cloud)

internal/credentials/kubernetes/acr/workload_identity.go
blakepettersson: Renamed it to `repoURL`, and renamed the string which it's constructed from to `repo`. I don't really like the string name but can't think of anything better atm...

internal/credentials/kubernetes/acr/workload_identity.go
blakepettersson: We also need to determine the [management endpoint](https://learn.microsoft.com/en-us/rest/api/azure/). From what I can see it's `https://management.core.windows.net/` when I test this, but I don't know if this will be true for all environments.

internal/credentials/kubernetes/acr/workload_identity.go
krancour: Personally, in a case like this where I have a string I will parse into some other type, I add a suffix to the string var. So here, `repoURLStr` and `repoURL`, but that's, as I said, a personal convention. Feel free to use it or not.

internal/credentials/kubernetes/acr/workload_identity.go
blakepettersson: Ah, these endpoints are cloud specific:

The default Public Cloud has `https://management.core.windows.net`, but the endpoint differ with the other variants (China Cloud, Gov Cloud etc). We could perhaps do what external-secrets does and have the Azure Cloud Type (`AZURECHINACLOUD`, `AZUREPUBLICCLOUD` etc) be set as a Helm value / env variable, and then get those endpoints with a map lookup without needing to call the metadata API?

krancour: I've been racking my brain, because I used to know an easier way to do this. I finally found it... sort of...

If "environment" is configured at install-time with a well-known value like `AzurePublicCloud` or `AzureChinaCloud`:

```golang
import "github.com/Azure/go-autorest/autorest/azure"

// ...

azureEnv, err := azure.EnvironmentFromName(envName)
// ...
```

The returned `azure.Environment` struct includes _some_ of the same info you are obtaining right now using a direct HTTP call to the Azure Instance Metadata Service. We can get both the AAD endpoint and the resource management endpoint.

In the end, I like your existing approach of querying IMDS directly much better. I'd initially found it objectionable because of the direct HTTP request. It'd have been nice if we could have accomplished everything using official Azure SDKs. Although `azure.EnvironmentFromName(envName)` looks tidier, it comes at the price of requiring more configuration at install-time.

So... as long as we're making that call to IMDS, we _might_ be able to get tenantID from there as well. It's not documented, but I _suspect_ it might be there. I will experiment to try and find out.

We can fall back on getting tenantID from the `AZURE_TENANT_ID` env var like you do currently, _but_ if we're able to get the tenant ID without looking at `AZURE_TENANT_ID`, the advantage will be no need to establish a managed identity for the controller itself. You could bother with managed identity for the Projects only. So, if we _can't_ get tenantID from IMDS, there may still be a benefit in having the user provide tenantID at install-time since it helps eliminate the need for the controller to have a managed identity.

Of course, if we want functionality like we have for ECR where we fall back on the controller's own permissions if the Project doesn't have a managed identity, then the controller _does_ need a managed identity.

I think all of this is going to be one giant game of starting with the ideal where no extra configuration is required at install-time and only Projects need managed identity and then we start gracefully falling back on other strategies as needed.

krancour: > So... as long as we're making that call to IMDS, we might be able to get tenantID from there as well. It's not documented, but I suspect it might be there. I will experiment to try and find out.

Nope. It's not there. At least not when I `curl -H "Metadata: true" http://169.254.169.254/metadata/instance?api-version=2021-02-01` from within an Azure VM. 😦 

So I guess this means we do try to get tenantID from the `AZURE_TENANT_ID` env var and if we come up empty-handed because the controller doesn't have its own managed identity, then we fall back on a value provided at install-time -- which we will document as being used only when tenantID cannot be inferred.

krancour: Oh... you were already calling `azure.EnvironmentFromName(envName)`.

krancour:
I amended the PR with 655ce12bf8fb0eb0848aefbd6f762088ce58e481

It's some changes to how we configure and initialize the component.

Please do feel free to revert it if you don't like it.

blakepettersson:
> I amended the PR with [17ecfed](https://github.com/akuity/kargo/commit/17ecfedbe790df060cf1bb45f5033c46c57c0131)
> 
> It's some changes to how we configure and initialize the component.
> 
> Please do feel free to revert it if you don't like it.

This looks great, thanks!

krancour:
@blakepettersson def more to do still. Wanna keep tag-teaming this til we're both happy with it?

krancour:
@blakepettersson 🤦‍♂️ I realized my optimization of using the controller's own managed identity to try to discover the tenantID (in case the admin installing Kargo hasn't bothered to provide it) doesn't work in practice...

_Because you need to know the tenantID to configure workload identity for the controller itself._

i.e. There is no getting around requiring tenantID to be specified at the time the chart is installed. I will make the requisite simplifications to my previous amendment.

krancour:
@blakepettersson I've just realized a major issue with the overall approach...

With EKS and GKE, getting a token for ECR or GAR using those platforms' respective workload identity solutions works as long as the _controller_ in question is running in EKS or GKE.

The solution you and I have been going back and forth on here requires the Kargo _control plane_ to be running in AKS since that is where the Project-specific SAs that are federated with each Project's managed identity will reside.

There are reasons this may be a somewhat damning limitation. I'll ping you off-line.

krancour:
With the realization that this approach requires the Kargo control plane to live in AKS, this isn't going to hold up. I'm going to tentatively close this PR, but we may one day be able to salvage parts of it once we have identified a more acceptable option.

PR #2402, opened 2024-08-08
fix(promotion): sync ArgoCD Application without updates

hiddeco:
Continuation of #2390, which historically actually had much to do with how #1922 was addressed.

The previous assumptions about there not being a desired source ended up being a red herring. As without updates, this will equal the current source in the Application's spec — resulting in a no-op because this branch _likely_ still equals the source from the previous Promotion-triggered sync result.

The actual working solution, is to always sync the Application when no specific updates are provided, at the cost of idempotency due to Kargo only being aware of the e.g. Git reference (and not the commit SHA) it expects the Application to be synced to.


jessesuen (approved):
LGTM

PR #2409, opened 2024-08-09
fix: multi-source support

shmargum:
i tried to keep the changes minimal, best viewed with whitespace changes ignored

this also adds an option to start the kargo controller with a debugger, which maybe doesn't belong here but figured i'd include it anyway


krancour:
Hi @shmargum we appreciate the contribution, but this issue is unfortunately much more complex than it may appear on the surface and literally _months_ of conversation and effort have gone into #2160, so I believe we would much prefer to stay the course with that that already-ongoing effort.

Issue #2423, opened 2024-08-12
Multi source Argo CD applications fail when promotions are applied


danielloader:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

I have been attempting to use Kargo with a multi-source Argo CD application last week that has a relatively simple use-case:
- One helm chart with an upstream application (OpenFGA)
- One kustomization manifest containing auxiliary manifests for this chart e.g. Cloudnative PG cluster, External Secrets, etc.

The screenshots show the pre-promotion state (the default state of the Application) and the post promotion state of the Application. In this scenario moving from 0.2.9 to 0.2.10 on OpenFGAs helm chart, and moving from the HEAD to a fixed commit on the main branch on the kustomization.

On promotion, the targetRevisions are being bumped correctly but the rest of the object is lost causing Argo CD to fail to sync (thankfully) due to nil pointer dereferences.

```
InvalidSpecError: spec.source.repoURL and either source.path, source.chart, or source.ref are required for source {https://github.com/example/platform-services e0b568919388ed0f514bc1ac18dbed344ecbba6c nil &ApplicationSourceKustomize{NamePrefix:,NameSuffix:,Images:[],CommonLabels:map[string]string{},Version:,CommonAnnotations:map[string]string{},ForceCommonLabels:false,ForceCommonAnnotations:false,Namespace:,CommonAnnotationsEnvsubst:false,Replicas:[]KustomizeReplica{},Patches:[]KustomizePatch{},Components:[],LabelWithoutSelector:false,} nil nil }
```

# Screenshots

![image](https://github.com/user-attachments/assets/64ee585b-8387-4ff0-b2c8-3b49efdc3411)
![image](https://github.com/user-attachments/assets/dfc5854f-8a0e-4502-8191-a66c4c0678fb)

# Steps to Reproduce

Here's the stage in question, I'm assuming the warehouse configuration is not pertinent as it's finding/selecting the correct versions for the sources.

```yaml
apiVersion: kargo.akuity.io/v1alpha1
kind: Stage
metadata:
  name: uat
  namespace: service-openfga
spec:
  promotionMechanisms:
    argoCDAppUpdates:
      - appName: openfga-uat
        appNamespace: argocd
        origin:
          kind: Warehouse
          name: openfga
        sourceUpdates:
[...]

PR #2428, opened 2024-08-13
fix(promotion): patch Application w/ Unstructured

hiddeco:
This PR changes the approach to patching Argo CD Applications, using unstructured data. This addresses an issue where fields not present in Kargo's Application shim (which we maintain to not rely on Argo CD and its dependencies) could accidentally be removed during patch operations (see #2423).

It achieves this by modifying and patching the Argo CD Application as an unstructured Kubernetes object, using replace and merge operations where applicable to update the fields Kargo cares about. To ensure merges of list items (i.e. `.spec.sources`) do not result in an unintended result (for example, because an `ApplicationSet` made a change to the list) — both the generation of the objects are compared before the modifying operations, and an optimistic lock is used during the actual patch operation.

### DIFF
--- a/internal/controller/promotion/argocd.go
+++ b/internal/controller/promotion/argocd.go
@@ -11,6 +11,7 @@ import (
 	"github.com/gobwas/glob"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


internal/controller/promotion/argocd.go
```diff
@@ -404,18 +402,23 @@ func (a *argoCDMechanism) syncApplication(
...
-		patch,
-	); err != nil {
+	// Patch the Argo CD Application.
+	if err := a.argoCDAppPatchFn(ctx, app, func(src, dst unstructured.Unstructured) error {
+		// If the resource has been modified since we fetched it, an update
+		// can result in unexpected merge results. Detect this, and return an
+		// error if it occurs.
+		if src.GetGeneration() != dst.GetGeneration() {
+			return fmt.Errorf("unable to update sources to desired revisions: resource has been modified")
+		}
```
krancour, line 412: Wondering if this is a universal concern and should just be built into the `kubeclient.PatchUnstructured()` func?

krancour (approved):

PR #2430, opened 2024-08-15
feat(api): grant API to manage ConfigMap

devholic:
This commit grants API to manage ConfigMap, which allows users to manage ConfigMaps that used in `AnalysisTemplate` through Kargo API.


hiddeco (approved):
This looks good to me, but I would like an additional review from @krancour as we recently very carefully audited them (#2230).

krancour (approved):

PR #2432, opened 2024-08-15
chore(deps): update Go to 1.23

hiddeco:
Testing the waters, as there have been reports about some breaking changes in standard libraries.


krancour (approved):
I am satisfied with this whenever you are @hiddeco. Preemptively LGTM'ing.

hiddeco:
I think this is good to go for now, one thing we should keep in mind ourselves is that Go 1.23 continued to enrich the standard libraries (https://go.dev/doc/go1.23#library), with more iterator utilities for `slices` and `maps`.

PR #2433, opened 2024-08-15
fix(controller): improve logic to determine when syncing an argo cd app is required

### LINKED ISSUE #2427: bug: not checking if app sources are synced correctly in some common cases
Another entry in the saga that is #1922 #2390 #2402.

Prior to #2390, during promotion, for each `ArgoCDAppUpdate` promotion mechanism, we unconditionally checked whether we needed to "update" the corresponding Argo CD Application. Update includes a forced sync. Determining whether this was required involved looking at operation status and, if successful, determining whether Application sources were synced to desired revisions (for any artifact where the desired revision can be determined; i.e. for any artifact in the Freight). A specific benefit to this was that operations that weren't complete blocked the Promotion from being considered completed and if/when an Application was observed to have source(s) not synced to the desired revision, the sync would then be initiated because the App "must update."

#2390 and #2402 are variations on accounting for outlying cases where the desired revision of a source cannot be known. #2402, makes the mistake of equating an `ArgoCDAppUpdate` promo mechanism that has no child `ArgoCDSourceUpdate` mechanisms with the unknowability of desired revisions of each source. In such a case, an App sync is initiated and immediately considered complete with the presumption that we wouldn't know how to judge its completeness accurately even if we did wait.

https://github.com/akuity/kargo/blob/81d085e1823b16738ae234c08d482d1765c2704a/internal/controller/promotion/argocd.go#L138-L155

This ends up bypassing the process that could have determined desired revisions (despite the absence of any `ArgoCDSourceUpdate`s; if it's in the Freight, we know it). The result is that in a number of common cases, we prematurely, and possibly incorrectly, count Promotions as complete and successful when the underlying Application(s) sources correct revisions were knowable, but were not checked.

krancour:
Fixes #2427 

925b765b461bd3fc19fd7e9055698140bd76e5a3 (the second commit) is the actual bug fix and is pretty straightforward. It modifies the sync operations we initiate to include the id of the Freight collection that is being promoted. We check for this information when determining if we need to initiate a sync operation. This allows a promotion to wait for a sync operation to complete in cases where we previously would not have simply for lack of any means to determine whether the completed sync operation was indeed the one we'd been waiting for.

fe15116e9e8447a2f12b9f6da5a392216606b61c (the first commit) changes the interface of promotion mechanisms to be quite a bit cleaner. The bug fix described above required the Argo CD app update promotion mechanism to have access to the Freight collection ID. It could obtain it from `promo.Status.FreightCollection.ID`. Doing this made me realize how _seemingly_ redundant it was that promotion mechanisms were also receiving and returning `[]kargoapi.FreightReference`, since there is already access to the Freight collection. In fact, it was not entirely redundant because promotion mechanisms can make certain modifications to those Freight references (like set health check commit IDs) and return them, and then those modified references become the input to the next promotion mechanism. The situation was quite similar for Promotion status, which could be modified by each mechanism and returned. The more I looked at this, the more confusing it started to look and feel, and I felt it was leaving too much potential for future mistakes. So, while I don't usually love when functions modify an argument in place (Promotion, in this case), updating the promotion mechanisms to get all details from the Promotion argument and modify its status (including Freight collection) in place made everything feel much more straightforward.

### DIFF
--- a/internal/controller/promotion/argocd.go
+++ b/internal/controller/promotion/argocd.go
@@ -27,6 +27,7 @@ const (
 	authorizedStageAnnotationKey = "kargo.akuity.io/authorized-stage"
 
 	applicationOperationInitiator = "kargo-controller"
+	freightCollectionInfoKey      = "kargo.akuity.io/freight-collection"
 )
 
 // argoCDMechanism is an implementation of the Mechanism interface that updates
@@ -52,10 +53,11 @@ type argoCDMechanism struct {
 		argocd.ApplicationSources,
 	) (argocd.OperationPhase, bool, error)
 	syncApplicationFn func(
-		context.Context,
-		*argocd.Application,
-		*argocd.ApplicationSource,
-		argocd.ApplicationSources,
+		ctx context.Context,
+		app *argocd.Application,
+		desiredSource *argocd.ApplicationSource,
+		desiredSources argocd.ApplicationSources,
+		freightColID string,
 	) error
 	getAuthorizedApplicationFn func(
 		ctx context.Context,
@@ -134,25 +136,6 @@ func (a *argoCDMechanism) Promote(
 			return err
 		}
 
-		// If we do not have specific source updates, request a sync of the
-		// Application with its current source(s).
-		if len(update.SourceUpdates) == 0 {
-			if err = a.syncApplicationFn(
-				ctx,
-				app,
-				app.Spec.Source.DeepCopy(),
-				app.Spec.Sources.DeepCopy(),
-			); err != nil {
-				return err
-			}
-
-			// As we have no knowledge of the specifically desired revision(s),
-			// we cannot wait for the update to complete as we would not know
-			// what to wait for.
-			updateResults = append(updateResults, argocd.OperationSucceeded)
-			continue
-		}
-
 		// Build the desired source(s) for the Argo CD Application.
 		desiredSource, desiredSources, err := a.buildDesiredSourcesFn(
 			ctx,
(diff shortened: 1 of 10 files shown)


internal/controller/promotion/argocd.go
```diff
@@ -309,6 +298,31 @@ func (a *argoCDMechanism) mustPerformUpdate(
 		return "", true, nil
 	}
 
+	// Deal with the possibility that the operation was not initiated for the
+	// current freight collection. i.e. Not related to the current promotion.
+	var correctFreightColIDFound bool
+	for _, info := range status.Operation.Info {
+		if info.Name == freightCollectionInfoKey {
+			correctFreightColIDFound = info.Value == freightCol.ID
```
hiddeco, line 306: Should this consider that _hypothetically_, a Freight collection could somehow end up without an ID?

krancour: It would be a nice thing to safeguard against, but I think it's a rabbit hole unto itself.

The ID is updated whenever you push to the collection. But because the collection is an API type, all its fields are exported, else marshaling/unmarshaling doesn't work. That introduces two things that I don't love: 1. You can tamper with the underlying collection without the ID being updated. 2. You can mutate the ID in place so it doesn't match the underlying collection anymore. These would be the only two ways that your hypothetical could occur, and except in test cases where we are engineering various scenarios, we never do anything of the sort.

I definitely do not love the _potential_ for a FreightCollection to be mishandled, but I don't feel we have the luxury of sorting that out at the moment. wdyt?

hiddeco (approved):
This is absolutely better, and much more resilient than what I came up with. 💯 for thinking about the operation info field! 🧠🥇

PR #2437, opened 2024-08-16
feat: make local server work finally

### LINKED ISSUE #1728: cli: add `kargo server` command

krancour:
Fixes #1728

Follows up on #2395 cc @rbreeze 

This makes some modifications to how we construct the custom k8s client used by the API server only. This makes a lot of the boilerplate configuration of that client a bit DRYer so that we can more easily use it in multiple places -- with one of those places being the `kargo server` command.

`kargo server` fully works after this.

Note that to try this, you need to make sure the UI is baked into the CLI, which it typically isn't except by our actual release process.

I did something like this to ensure that:

```console
# Build UI to be embedded into CLI binary             
cd ui
pnpm install
NODE_ENV=production pnpm run build
cd ..
rm -rf internal/api/ui
mv ui/build internal/api/ui

# Start server
go run cmd/cli/*.go server
```


hiddeco (approved):
Tested it, and it works well. I added two Make targets as a little 🎁, so it continues to be easy to try this out locally whenever we need it.

krancour:
Thanks for the 🎁 @hiddeco!

PR #2449, opened 2024-08-20
chore: centralize development tool management

hiddeco:
Fixes: #1525
Prerequisite to addressing: #2052

This pull request introduces a centralized system for managing development tools using a Makefile and Go modules.

It aims to:

- improve consistency and ease of maintenance across the project
- prevent having versions declared in multiple places
- step away from relying on system-global installations as much as we can (NPM and PNMP excluded)
- allow the binaries to be cached, both on the user their local system and for any of the existing `hack-*` targets

The main changes:

- Add `hack/tools.mk` for managing tool installations (which is included in our central `Makefile)
- Create `hack/tools/go.mod` for versioning Go-based tools (which allows them to be updated by Dependabot)
- Update `Makefile` and various scripts to use the newly managed tools
- Update the CI workflow to utilize the Make targets (and cache their results)


krancour (approved):
This is fantastic and I am especially appreciative that it's not disruptive of existing workflows, including container-based ones. (I still feel like running tests and such in a clean room is a useful thing at times.)

🔥 🔥 🔥

PR #2470, opened 2024-08-27
feat: support kustomize image newName override

jessesuen:
Some users will use different image repositories for prod vs. preprod. As part of their CI process, they might retag/push images to a production image repository. And so they would take advantage of kustomize's newName feature like so:

```yaml
images:
- name: docker.io/example/guestbook-preprod
  newName: docker.io/example/guestbook-prod
  newTag: 1.0.0
```

The problem is that kargo will only run `kustomize edit set image <image>:<tag>` which blows away the `newName` setting.

This change introduces `newName` to the kustomize spec, and when specified, will run:

```
kustomize edit set image <image>=<newName>:<tag>
```

otherwise:

```
kustomize edit set image <image>:<tag>
```


api/v1alpha1/stage_types.go
```diff
@@ -423,6 +423,10 @@ type KustomizeImageUpdate struct {
 	//
 	// +kubebuilder:validation:Optional
 	UseDigest bool `json:"useDigest" protobuf:"varint,3,opt,name=useDigest"`
+	// NewName specifies a container image name override when setting the image
+	//
+	// +kubebuilder:validation:Optional
+	NewName string `json:"newName" protobuf:"bytes,5,opt,name=newName"`
```
krancour, line 429: I think omitempty should usually be set on optional fields. Weirdly, controller-gen prioritizes that over the kubebuilder markers. It would probably be ok here, but we should do this for consistency.

krancour: Apart from this, LGTM.

krancour (approved):

api/v1alpha1/stage_types.go
jessesuen: Nice catch. Agree

jessesuen:
Verified the feature is working. newName can be specified next to `image`: 

```yaml
    gitRepoUpdates:
      - repoURL: https://github.com/jessesuen/kargo-simple.git
        insecureSkipTLSVerify: false
        writeBranch: main
        kustomize:
          images:
            - image: ghcr.io/jessesuen/guestbook
              path: env/dev
              useDigest: false
              newName: ghcr.io/akuity/guestbook
```

Resulting git commit from kargo promotion will set the value:

image

danielloader:
Is this also able to work with argoCDAppUpdates sourceUpdates?

krancour:
@danielloader it was only for git-based changes.

PR #2480, opened 2024-08-29
feat(api): add endpoints to get AnalysisTemplate config resources

devholic:
This commit adds endpoints to get env resources - `ConfigMap` and `Secret` - which can be used in the AnalysisTemplate, to allow users to manage those resources in the Kargo UI. The `ConfigMap` and `Secret` must be labeled with `kargo.akuity.io/analysis-env: "true"` to be returned in API. 

### Note

- We may leverage existing `Create/Update/DeleteResource` API to manage `ConfigMap` and `Secret`, same as the `AnalysisTemplate` (cc @gdsoumya)

### Open Questions

- I'm not sure about the `kargo.akuity.io/analysis-env` key naming, could anyone give other suggestions?


hiddeco:
What about just calling the annotation `/ui-managed` (to compliment `/managed`), so that we can eventually utilize it in other places without having to introduce a bazillion annotations?

hiddeco:
The other option I can think of, as it's theoretically not just the UI, is to call it `/api-managed` or `/api-accessible`.

hiddeco:
Talked this through some more offline, and the worry with a `/api-managed` or `/api-accessible` endpoint is that a value of `true` would not allow the caller to "categorize" the ConfigMaps and/or Secrets they retrieve.

Given this, and with an eye on the future, I propose looking for a generic label name with a special _categorized_ value, e.g., `env`. This way, we allow room for future enhancements in which we add new categories, potentially using a generic endpoint that accepts a query parameter to set the value.

internal/api/get_analysis_template_config_map_v1alpha1.go
```diff
@@ -0,0 +1,95 @@
...
+	}, &u); err != nil {
+		if client.IgnoreNotFound(err) == nil {
+			err = fmt.Errorf("ConfigMap %q not found in namespace %q", name, project)
+			return nil, connect.NewError(connect.CodeNotFound, err)
+		}
+		return nil, err
+	}
+
+	if u.GetLabels()[kargoapi.AnalysisEnvLabelKey] != kargoapi.LabelTrueValue {
+		err := fmt.Errorf("ConfigMap %q is not allowed for the AnalysisTemplate", name)
```
hiddeco, line 65: From a security perspective, I think it may be better to ignore its existence.

devholic:
@hiddeco 

> Talked this through some more offline, and the worry with a `/api-managed` or `/api-accessible` endpoint is that a value of `true` would not allow the caller to "categorize" the ConfigMaps and/or Secrets they retrieve.
> 
> Given this, and with an eye on the future, I propose looking for a generic label name with a special _categorized_ value, e.g., `env`. This way, we allow room for future enhancements in which we add new categories, potentially using a generic endpoint that accepts a query parameter to set the value.

If we have generic label name and using categorized value, couldn’t we leverage the Kubernetes’s label selector? For example, `kargo.akuity.io/configmaps: "analysis-env"` and if we add another categorized value (such as) stage-env , the label should be `kargo.akuity.io/configmaps: "analysis-env, stage-env"` but not sure if we can leverage label selector in this case.

hiddeco:
Yes, you can use a label selector, which we have already been doing for credential secrets (and what is being practiced in this PR). However, this does not work when getting a specific object, which requires additional verification.

devholic:
Then I'll choose to use `kargo.akuity.io/analysis-run-template: "config"` for both ConfigMap and Secret - The reason I chose `config` over `env` is that both of them can be used as a volume

WDYT? @hiddeco @gdsoumya

krancour (approved):

Issue #2400, opened 2024-08-07
harden image


krancour:
When time permits, I would like to harden the Kargo image in a similar way to what we did for Kargo Render in https://github.com/akuity/kargo-render/pull/297

PR #2483, opened 2024-08-30
refactor(controller): overhaul git repo interfaces

krancour:
This is a bit of an overhaul of the `internal/controller/git` package that kills a few birds with one stone.

Promotion directives will require support for bare repos with multiple working trees associated with them.

The capabilities of a normal repo with a single working tree vs a bare repo vs a working tree associated _with_ a bare repo are all a little different. It made sense to expose three distinct interfaces -- Repo, BareRepo, and WorkTree. (The Repo interface is a superset of the WorkTree interface while the BareRepo interface lacks all the capabilities of a WorkTree, but has some distinct capabilities like AddWorkTree or RemoveWorkTree.)

Git-based promotion directives will rely heavily on these new interfaces. A clone directive may create a bare clone with one or more working trees.

Subsequent directives, such as commit and push will require the ability to interact with bare repos and working trees _that already exist in the FS_. Thus, obtaining interfaces for existing repos and working trees is another new capability introduced by this PR.

Directives will also introduce the possibility that, for instance, pushing a commit occurs long after cloning a repo, and in the intervening time, the credentials used to clone may have expired and will not be useful for pushing. This introduces a requirement that when obtaining interfaces for repos or working trees that already exist on the file system, the credentials must be refreshed.

This PR also takes #2400 into account. I anticipate the image-hardening effort will soon result in Kargo images that lack a shell. Kargo Render has already been through this in https://github.com/akuity/kargo-render/pull/297. The store credential helper that we previously relied on _requires_ a shell to function properly. So while already ensuring credentials are refreshed whenever an interface is obtained for an existing repo or working tree from the file system, I decided it was advisable to cut over to the custom credential helper strategy that Kargo Render also uses now.

A couple noteworthy aspects of this PR:

* The existing Repo interface _barely_ breaks, so the impact to existing code is minimal.

* The implementation of existing functionality barely changes.

* The PR adds new tests where none existed before to aid in asserting that nothing breaks.

* I've also executed a battery of e2e tests against these changes and everything checks out.


internal/controller/git/work_tree.go
```diff
@@ -0,0 +1,514 @@
...
+			insecureSkipTLSVerify: opts.InsecureSkipTLSVerify,
+		},
+	}
+	res, err := libExec.Exec(w.buildGitCommand("rev-parse", "--git-dir"))
+	if err != nil {
+		return nil, fmt.Errorf("error getting git directory: %w", err)
+	}
+	// The root of the bare repo will be three levels up from the path returned
+	// by git rev-parse --git-dir
+	repoPath := filepath.Join(strings.TrimSpace(string(res)), "..", "..", "..", "repo")
```
hiddeco, line 113: Non-blocking: this feels prone to errors.

hiddeco (approved):
This is excellent! 🔥🔥🔥

internal/controller/git/work_tree.go
krancour: I didn't love this either and I'd be happy to take any suggestion your might have.

krancour: Refactored this in https://github.com/akuity/kargo/pull/2483/commits/b4a47b07829d6b326110b69d352a4bb134a4c5b9

hiddeco (approved):

PR #2486, opened 2024-09-03
feat(controller): more git pkg enhancements

krancour:
This adds a couple more enhancements to our git package:

* On clone, persists preference for ignoring cert errors as config so that preference does not need to be re-specified when loading an existing repo from the FS.

* Add option when pushing to push to a remote branch with a different name than the current local branch.

There are more prerequisite capabilities for the git-based promotion directives.

### DIFF
--- a/internal/controller/git/bare_repo.go
+++ b/internal/controller/git/bare_repo.go
@@ -81,11 +81,10 @@ func CloneBare(
 	}
 	b := &bareRepo{
 		baseRepo: &baseRepo{
-			creds:                 clientOpts.Credentials,
-			dir:                   filepath.Join(homeDir, "repo"),
-			homeDir:               homeDir,
-			insecureSkipTLSVerify: cloneOpts.InsecureSkipTLSVerify,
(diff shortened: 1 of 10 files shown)


hiddeco:
> On clone, persists preference for ignoring cert errors as config so that preference does not need to be re-specified when loading an existing repo from the FS.

Is there a hypothetical scenario where this may change between attempts?

krancour:
> Is there a hypothetical scenario where this may change between attempts?

I thought about that as well, but it seemed that this would require _both_ the validity of a cert to change between directives/attempts _and_ for the desire to ignore/not ignore invalid certs to also change as well. The former seems like an edge case, but possible. The latter seems impossible, however, since I assume the directives from which one Promotion is composed cannot be mutated once that promotion is in-flight.

So I think that when you clone, you're either ignoring or not ignoring cert errors for the life of that clone.

wdyt?

hiddeco:
Think the latter observation solves all of my initial concerns.

hiddeco (approved):

PR #2488, opened 2024-09-04
feat(ui): surface verification errors in DAG 

### LINKED ISSUE #2487: Pipeline view should indicate when most recent verification failed
# Proposed Feature

When verifications fail, they are not visible from the pipeline view. So user does not know anything is wrong unless they click into a Stage.

We should surface a problem indicator in the Stage for when a verification has failed.

# Motivation

Improve the UX for when something is wrong.

# Suggested Implementation

rbreeze:
Fixes #2487

Screenshot 2024-09-04 at 11 10 57


krancour (approved):
🔥

Issue #2491, opened 2024-09-05
Missing Documentation on Previewing Local Changes for Kargo Docs


fykaa:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] This feature is related to improving the documentation process.

# Proposed Feature

Add clear documentation on how to preview local documentation changes in Kargo, particularly for users who prefer using containers to avoid installing Node.js directly.

# Motivation

Contributors often need to preview changes locally before pushing them. However, there's no clear guidance in the current documentation on how to do this, particularly for developers who prefer a containerized environment. Providing this information will make the contribution process smoother and more accessible, especially for new contributors

# Suggested Implementation

- Add a section in the [contributor's guide](https://kargo.akuity.io/contributor-guide/) explaining how to preview local doc changes.
- Include instructions for both:
  - A local setup using Node.js and `pnpm`.
  - A containerized solution (possibly using a pre-built image with necessary tools installed).

PR #2493, opened 2024-09-06
feat: add convenience targets for working on docs

krancour:
@fykaa let me know if these are the sort of things that would have saved you some trouble up front and whether you think you could documents these targets as part of #2491.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,8 @@ IMAGE_PUSH 			?= false
 IMAGE_PLATFORMS 	=
 DOCKER_BUILD_OPTS 	=
 
+DOCS_PORT 				?= 3000
+
 # Intelligently choose to build a multi-arch image if the intent is to push to a
 # container registry (IMAGE_PUSH=true). If not pushing, build an single-arch
 # image for the local architecture. Honor IMAGE_PLATFORMS above all.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -358,3 +360,18 @@ start-controller-local:
...
+# Convenience targets for building and running the documentation site natively #
+# or in a container.                                                           #
+################################################################################
+
+.PHONY: hack-docs
+hack-docs: hack-build-dev-tools
+	$(CONTAINER_RUNTIME) run $(DOCKER_OPTS) -p $(DOCS_PORT):$(DOCS_PORT) kargo:dev-tools make docs
+
+.PHONY: docs
+docs:
```
hiddeco, line 376:
```suggestion
serve-docs:
```

maybe?

krancour: I can get down with that.

krancour: After thinking about it, I do slightly favor `serve-docs` / `hack-serve-docs` despite being a bit more verbose. There are other doc-related targets -- and for different kinds of docs, no less. For instance, there is `codegen-docs`, which deals with generating chart docs from the metadata in the Helm chart's `values.yaml` file. Naming things is hard. Humans, by nature, are easily confused... so I am often willing to sacrifice brevity for clarity.

krancour: 3b999b7849826600f4f88e4fad8521987fa3f69e updates the names of the targets as discussed above.

hiddeco (approved):

PR #2495, opened 2024-09-06
fix(docs): specify pnpm version for building docs

krancour:
Unlike the UI's `package.json`, the `package.json` for the docs site and for the custom plugin it uses were not specifying what version of pnpm to use. As a result, build/deploy of the docs site on Netlify was using their default pnpm version, which was older. Thanks to [a recent PR from Dependapot](https://github.com/akuity/kargo/pull/2469) the plugin's `pnpm-lock.yaml` is no longer compatible with that older pnpm...

This made pnpm _ignore_ the lock file and resolve _different_ versions of some dependencies, which, in turn, was responsible for the recent failed doc site builds on Netlify.

This PR should set things straight again and because the current production docs are built and deployed from the `release-0.8` branch, I am going to also backport this.

cc @fykaa this may solve some recent troubles you've had.

cc @rbreeze for node ecosystem sanity check.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -15,6 +15,7 @@
     "write-heading-ids": "docusaurus write-heading-ids",
     "typecheck": "tsc"
   },
+  "packageManager": "pnpm@9.0.3",
   "dependencies": {
     "@cmfcmf/docusaurus-search-local": "^0.11.0",
     "@docusaurus/core": "^3.0.0",
--- a/docs/plugins/gtag/package.json
+++ b/docs/plugins/gtag/package.json
@@ -7,6 +7,7 @@
     "build": "tsc --build",
     "watch": "tsc --build --watch"
   },
+  "packageManager": "pnpm@9.0.3",
   "dependencies": {
     "@docusaurus/core": "^3.0.0",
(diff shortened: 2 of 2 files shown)


hiddeco:
When I worked on #2449, I looked into streamlining the NodeJS-related dependencies. This has made me wonder if we should not be more opinionated about how we practice our Node versioning (at least for builds), using e.g. [Corepack](https://nodejs.org/api/corepack.html). Although this is, _of course_, part of [new NodeJS drama](https://socket.dev/blog/node-js-takes-steps-towards-removing-corepack).

krancour:
@hiddeco fwiw, [Netlify is using Corepack already](https://docs.netlify.com/configure-builds/manage-dependencies/#pnpm).

It's quite possible we might get some mileage out of it ourselves and we should definitely have a look at that, but I think the minimal changes in this PR are a critical stop-gap to get our doc previews, not to mention official doc deployments, working again.

hiddeco:
Yes, my comment was more of an observation rather than a request to address this now.

hiddeco (approved):

PR #2497, opened 2024-09-06
fix(ui): promote downstream subs in control flow stages was broken

rbreeze:


ui/src/features/project/pipelines/pipelines.tsx
```diff
@@ -443,14 +443,7 @@ export const Pipelines = ({ project }: { project: Project }) => {
...
                             } else {
                               const stageName = node.data?.metadata?.name || '';
-                              // default to current freight when promoting subscribers
-                              state.select(
-                                type,
-                                stageName,
-                                type === FreightTimelineAction.PromoteSubscribers
-                                  ? currentFreight[0].name
-                                  : undefined
-                              );
```
rbreeze, line 453: Defaulting to selecting current freight was the cause of the bug (`currentFreight` was undefined), but did not work in the happy path case anyways because this is not the mechanism by which freight are selected by promotion. So it is safe to remove this.

jessesuen (approved):

Issue #2214, opened 2024-06-27
feat: option for api server not to touch credentials (secrets) at all


krancour:
We're already very careful about this...

Rather than give the API server cluster-scoped r/w on Secrets, we have given that permission to the management controller instead. As Projects are created/deleted, the management controller dynamically expands/contracts the permissions of the API server by creating/deleting RoleBindings that add/remove r/w access to Secrets in the Project namespace only.

This is inherently more secure because unlike the api server, the management controller receives no inbound traffic and therefore has less attackable surface.

_But_ for users who elect to GitOps their credentials (along with everything else), the API server _could_ stand to entirely lose its ability to manage credentials, meaning the API server could, in such a case, not require permission to of any kind on any Secrets _at all._ 🥳 

For mature organizations, this trades functionality that _they wouldn't be using anyway_ for improved security footing.

On the back-end, this is super easy.

@rbreeze ideally, I'd like to figure out a way to hide credential management tabs _entirely_ if an admin opts into disabling that functionality. I was thinking we can add a field to the response from the protected configuration endpoint (_not_ the public one) that could inform the front-end of whether the API server can or cannot manage credentials. wdyt?

https://github.com/akuity/kargo/blob/93940f957ec5eda563f8b163a3fc3d78474e3120/api/service/v1alpha1/service.proto#L21

https://github.com/akuity/kargo/blob/93940f957ec5eda563f8b163a3fc3d78474e3120/api/service/v1alpha1/service.proto#L136-L138

__Edit:__ All of the above could equally be said for the API server's ability to manage SAs/Roles/RoleBindings in Project namespaces.

Brightside56:
@krancour we can continue our discussion here I think. From perspective of infrastructure engineer is see threat not in API server which can have an access to VCS token which can "touch" all production branches, I see treat in kargo-controller which has an access to VCS token in case of sharding

I wouldn't like kargo-controller of shard "staging" or "region-xx" to be able to "touch" all production branches (what would provide a possibility for intruder to deploy anything to anywhere from any cluster)

https://github.com/akuity/kargo/blob/main/charts/kargo/templates/controller/cluster-roles.yaml#L33 - Also controller has RBAC-permissions to manipulate freights/promotions/stages/warehouses

Every Kubernetes cluster (with deployed controller) is a part of root of trust and in case of compromise allows to perform infrastructure/organization takeover

[...]

PR #2517, opened 2024-09-09
feat: implement conditional secret management

fykaa:
Fixes: https://github.com/akuity/kargo/issues/2214

Key changes include:

1. **Configuration updates**:
   - `EnableSecretManagement` flag
   - modified `ClusterRole` and `ConfigMap` to manage secrets based on this flag
   - updated `values.yaml` and `config.go` to incorporate the new setting

2. **API enhancements**:
   - updated `GetConfig` endpoint to include `SecretManagementEnabled` status

3. **Endpoint adjustments**:
   - ensure endpoints for managing secrets return `connect.CodeUnimplemented` when secret management is disabled

cc: @krancour @rbreeze


charts/kargo/templates/api/configmap.yaml
```diff
@@ -18,6 +18,9 @@ data:
   TLS_CERT_PATH: /etc/kargo/tls.crt
   TLS_KEY_PATH: /etc/kargo/tls.key
   {{- end }}
+  {{- if .Values.api.enableSecretManagement }}
+  ENABLE_SECRET_MANAGEMENT: "true"
```
hiddeco, line 22:
```suggestion
  SECRET_MANAGEMENT_ENABLED: "true"
```

would match the style of `PERMISSIVE_CORS_POLICY_ENABLED` and `ADMIN_ACCOUNT_ENABLED` below.

charts/kargo/templates/api/configmap.yaml
```diff
@@ -18,6 +18,9 @@ data:
   TLS_CERT_PATH: /etc/kargo/tls.crt
   TLS_KEY_PATH: /etc/kargo/tls.key
   {{- end }}
+  {{- if .Values.api.enableSecretManagement }}
```
hiddeco, line 21:
```suggestion
  {{- if .Values.api.secretManagement.enabled }}
```

maybe? To match `.api.oidc.enabled`, .api.adminAccount.enabled`, etc.

krancour:
It looks like a few unit tests that use the credential management endpoints may need an update. Their config needs secret management enabled.

charts/kargo/templates/api/configmap.yaml
hiddeco: I now notice there is a `.api.rollouts.integrationEnabled`, so I guess we have already lost this game.

I would prefer things to be uniform at some point, but will not let this block you now. So go for whatever you think is best.

krancour: Let's make an issue for improving the consistency pre-GA.

krancour:
@rbreeze any chance we can get you to amend this PR with the bits to hide credential-management functionality when the config endpoint says its not enabled?

rbreeze:
@fykaa @krancour Sure thing, just pushed the update 👍🏻

krancour:
Added a tiny fix in 92f07be85881f3e7df6d34837893bf809536750a

charts/kargo/values.yaml
```diff
@@ -106,6 +106,9 @@ api:
   ## @param api.podAnnotations Optional annotations to add to pods. Merges with `global.podAnnotations`, allowing you to override or add to the global annotations.
   podAnnotations: {}
 
+  ## @param api.enableSecretManagement Specifies whether secret management is enabled. When enabled, the API server will manage secrets as allowed by its configured permissions. If disabled, the API server will not manage secrets, and related operations will be restricted or result in unimplemented responses. This setting enhances security by limiting the API server's permissions and reducing its attack surface. If secret management is not needed, it is advisable to disable this feature.
```
krancour, line 109: I propose rewording a little bit so that it is more clear that this is a choice one might make if they are GitOps'ing their Kargo Projects -- which is a common thing in larger, more security-conscious organizations.

```suggestion
  ## @param api.enableSecretManagement Specifies whether Secret management is enabled. This affects the API server's ability to manage repository credentials and other Project-level Secrets, such as those used by AnalysisRuns for verification purposes. If using GitOps to manage Kargo Projects declaratively, the API's Secret management capabilities are not needed and can be disabled to effectively reduce the API server's attackable surface.
```

Note that codegen needs to be re-run after this.

krancour (approved):

krancour:
Thank you @fykaa! This is great!

krancour:
And @rbreeze, thanks to you as well!

Issue #2518, opened 2024-09-10
fix: standardize configuration field naming


fykaa:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.

# Proposed Feature

Improve Consistency in Configuration Field names

We've noticed inconsistencies in the [naming of configuration fields](https://github.com/akuity/kargo/blob/main/charts/kargo/values.yaml). 

For example, `api.rollouts.integrationEnabled ` does not follow the same pattern as other fields like `.api.oidc.enabled` or `.api.adminAccount.enabled`. While there’s currently no immediate blocker, aligning these fields for better uniformity before GA would help improve the clarity and maintainability of the configuration.

krancour:
Let's not limit this to the API, but audit this throughout the entire values.yaml file.

krancour:
I am also moving this to v0.9.0 because with any sort of breakage, sooner is always better than later.

PR #2519, opened 2024-09-10
refactor!: standardise configuration fields

fykaa:
Fixes: https://github.com/akuity/kargo/issues/2518

Fields that needed improvements:

`api.rollouts.integrationEnabled`
`controller.argocd.integrationEnabled`
`controller.rollouts.integrationEnabled`
`enablePermissiveCORSPolicy`

### DIFF
--- a/internal/api/config/config.go
+++ b/internal/api/config/config.go
@@ -19,7 +19,7 @@ type StandardConfig struct {
 
 type ServerConfig struct {
 	StandardConfig
-	SecretManagementEnabled      bool
+	SecretManagementEnabled     bool
 	LocalMode                   bool // LocalMode is true if the server is running as a non-containerized process
 	TLSConfig                   *TLSConfig
 	OIDCConfig                  *oidc.Config


jessesuen (dismissed):

charts/kargo/values.yaml
```diff
@@ -148,8 +148,9 @@ api:
     ## @param api.tls.selfSignedCert Whether to generate a self-signed certificate for use by the API server. If `true`, `cert-manager` CRDs **must** be present in the cluster. Kargo will create and use its own namespaced issuer. If `false`, a cert secret named `kargo-api-cert` **must** be provided in the same namespace as Kargo.
     selfSignedCert: true
 
-  ## @param api.enablePermissiveCORSPolicy Whether to enable a permissive CORS (Cross Origin Resource Sharing) policy. This is sometimes advantageous during local development, but otherwise, should generally be left disabled.
-  enablePermissiveCORSPolicy: false
+  ## @param api.permissiveCORSPolicy.enabled Whether to enable a permissive CORS (Cross Origin Resource Sharing) policy. This is sometimes advantageous during local development, but otherwise, should generally be left disabled.
+  permissiveCORSPolicy:
+    enabled: false
```
jessesuen, line 153: Generally, I would agree with `.enabled: true` option in helm charts consistency.

However, for this field, `permissiveCORSPolicy`, I don't think make sense as a containing struct, since there would never be any other fields there other than `enabled`. So I suggest we either:

1. do nothing, and wait until we need support for more cors options

2. introduce a `corsOptions` field, with a convenience for our existing option:

```yaml
corsOptions:
  permissive: false
```

Then, in the future, we could expand this:

```yaml
corsOptions:
  allowedOrigins:
  - foo.com
```

jessesuen: NOTE: the current `enablePermissiveCORSPolicy` option is a [convenience to set the following options](https://github.com/akuity/kargo/blob/88f717019f9e7d72d3dc603c59a3c43444fe568f/internal/api/server.go#L221-L226):

```golang
		handler = cors.New(cors.Options{
			AllowCredentials: true,
			AllowedOrigins:   []string{"*"},
			AllowedMethods:   []string{"DELETE", "GET", "POST", "PUT"},
			AllowedHeaders:   []string{"Authorization", "Content-Type"},
		}).Handler(handler)
```

krancour: And in terms of backwards compatibility, this only affects a feature whose usefulness is limited to local development and a feature that isn't released yet, so this is a safe/simple fix.

krancour:
This is a solid improvement. LGTM and thank you!

krancour (approved):

hiddeco (approved):
This looks great! 🎉

Issue #2448, opened 2024-08-19
feat: limit controller access to control plane secrets


krancour:
For quite some time now, the permissions of the API server are dynamically expanded and contracted as new Projects are created, such that the API server receives access to Secrets in all Project namespaces, but does _not_ receive cluster-scoped access to Secrets.

It plausible and relatively simply to extend this behavior to sharded controllers as well.

Since all sharded controllers must have a SA within the Kargo control plane, if the management controller is configured with the names/namespaces of those SAs, it can dynamically bind those SAs to Project-specific Roles that grant r/o permissions on Project Secrets at the time of Project creation. Since those bindings would live within each Project's own namespace, deletion of a Project will naturally remove the associated binding.

The net result of this improvement would be more tightly-scoped access to control plane Secrets (limited to Project Secrets only) for all sharded/distributed controllers.

Brightside56:
> On (non-delete) Project reconciliation, ensure all SAs labeled as controller SAs have bindings to read Secrets in the Project namespace.

It would be extremely helpful to have possibility to isolate prod shards from non-prod shards from point of access to secrets... But with this improvement I could isolate prod from non-prod shards using separate Kargo projects for prod/non-prod

krancour:
> But with this improvement I could isolate prod from non-prod shards using separate Kargo projects for prod/non-prod

I'm not sure that would work... there's no promoting across projects.

Brightside56:
> I'm not sure that would work... there's no promoting across projects.

As a bare minimum, prod project could be subscribed on a warehouse which provides artifacts similarly as warehouse of non-prod project does.

It's not ideal, but it provides a possibility to reliable isolate shards in case when engineers working with project don't have permissions to edit/create stages.

But if I would like to provide greater degree of freedom for engineers (for example, possibility to create non-prod stages and deploy what they want on these stages) - I would need two separate Kargo instances for prod/non-prod

PR #2522, opened 2024-09-12
feat: limit controller access to control plane secrets

fykaa:
**Issue Reference:**  
Closes: https://github.com/akuity/kargo/issues/2448

This PR limits the `kargo-controller`'s access to project-specific secrets by creating a `ClusterRole` (`kargo-controller-secrets-readonly`) with read-only permissions. Role bindings are dynamically created per project and removed upon project deletion and this ensures least-privileged access for controllers.

**Impact:**  
Tighter scoping of controller access to secrets, more security for sharded controllers.


krancour:
@fykaa, this looks like a good start, but as you know, we've encountered some scope creep that is described in detail here: https://github.com/akuity/kargo/issues/2448#issuecomment-2346781212

internal/controller/management/projects/projects.go
```diff
@@ -392,6 +400,74 @@ func (r *reconciler) ensureAPIAdminPermissions(
 	return nil
 }
 
+func (r *reconciler) ensureControllerPermissions(
+	ctx context.Context,
+	project *kargoapi.Project,
+) error {
+	const roleBindingName = "kargo-controller-secrets-readonly"
```
krancour, line 407: Interesting. In my head, each Project was going to have a RoleBinding per controller SA.

You're actually creating or updating a single RoleBinding and making sure it has _all_ the controller SAs listed as its subjects.

On some level, I quite like that, but I have a worry this will be more complicated than anticipated once you get into part two of this PR...

Remember that we also need to deal with creating/removing bindings as new controller SAs come and go. Once we start doing that, you start to run into possibilities like a new Project and a new SA being reconciled concurrently and racing to update the same binding.

One tempting solution to that problem is to _patch_ the RoleBinding instead of updating it, but patching _lists_ can be quite difficult.

In the end, it may be more straightforward if every Project has one RoleBinding _per controller SA._ I feel that would lead to fewer conflicts and race conditions.

But I also want @hiddeco's opinion on this matter.

krancour: Note that apart from this, this PR is off to a very good start.

krancour: Let's wait to hear from @hiddeco before making any changes. He usually has the best insight into avoiding races in controllers.

krancour: @hiddeco any opinion on just using a separate RoleBinding per SA/Project pair? I cannot see a race that way and wonder if it isn't simpler than your option 1.

krancour: Sorry. Brain is still booting this AM...

> You put one reconciler in control of the reconciliation of the RoleBinding

ikd if this is a viable option. It's not the RoleBindings that need to be reconciled.

RoleBindings need to be updated _or_ created/removed in response to both Projects and controller SAs being added/removed.

hiddeco: You can create a reconciler around an arbitrary resource type since controller-runtime v0.19.0. See:

- https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/typed/main.go
- https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/multiclustersync/main.go

internal/controller/management/projects/projects.go
```diff
@@ -392,6 +400,74 @@ func (r *reconciler) ensureAPIAdminPermissions(
...
+			},
+		}
+
+		if err := r.createRoleBindingFn(ctx, roleBinding); err != nil {
+			if kubeerr.IsAlreadyExists(err) {
+				existingRoleBinding := &rbacv1.RoleBinding{}
+				if err := r.client.Get(ctx, client.ObjectKey{Name: roleBindingName, Namespace: project.Name}, existingRoleBinding); err != nil {
+					return fmt.Errorf("error getting existing RoleBinding: %w", err)
+				}
+				existingRoleBinding.Subjects = roleBinding.Subjects
```
krancour, line 448: I think I would not bother to retrieve the existing RB. I would probably jump straight to updating using the RB struct you've already built. There isn't any justifiable reason that anyone should be tampering with this RB as it belongs to _Kargo_ and not to a user. It should unconditionally look like what we say, so imho, we don't need to be careful about not squashing any modifications a user may have made to the existing RB.

Note that _if_ Projects were more frequently reconciled, I would probably encourage a different approach here, BUT as things currently stand, Projects are only reconciled until they are fully initialized or fail to initialize (without possibility of recovering). This means, on average, each Project is reconciled only once. Creating the RB will be the thing that usually needs to happen here, so its sensible to continue leading off with that, then dealing with the very unlikely case that the RB already existed by updating it, as I said, unconditionally.

internal/controller/management/projects/projects.go
```diff
@@ -392,6 +400,74 @@ func (r *reconciler) ensureAPIAdminPermissions(
...
+
+	// Get all ServiceAccounts labeled as controller ServiceAccounts
+	controllerSAs := &corev1.ServiceAccountList{}
+	if err := r.client.List(ctx, controllerSAs, client.InNamespace(r.cfg.KargoNamespace), client.MatchingLabels{"app.kubernetes.io/component": "controller"}); err != nil {
+		return fmt.Errorf("error listing controller ServiceAccounts: %w", err)
+	}
+
+	// Create/update a RoleBinding for each ServiceAccount
+	for _, sa := range controllerSAs.Items {
+		roleBindingName := fmt.Sprintf("kargo-controller-secrets-readonly-%s", sa.Name)
```
krancour, line 420: I would suggest a slightly different naming convention here:

```suggestion
		roleBindingName := fmt.Sprintf("%-readonly-secrets", sa.Name)
```

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/logging"
+)
+
+const (
+	ComponentLabelKey = "app.kubernetes.io/component"
+	ControllerLabelValue = "controller"
```
krancour, line 25: I think we should generally avoid exporting things that don't need to be referenced from other packages.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+type ReconcilerConfig struct {
+	KargoNamespace string `envconfig:"KARGO_NAMESPACE" required:"true"`
+}
+
+// ServiceAccountReconciler Reconciles for ServiceAccounts
+type ServiceAccountReconciler struct {
```
krancour, line 33: Please don't export this, and please rename it to just `reconciler`.

Go prefers that types not "stutter." This package is `serviceaccounts`. `serviceaccounts.ServiceAccountReconciler` is redundant.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
+package upgrade
```
krancour, line 1: This package should be named `serviceaccounts`.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+			CreateFunc: func(e event.CreateEvent) bool {
+				return hasControllerLabel(e.Object.(*corev1.ServiceAccount))
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return hasControllerLabel(e.Object.(*corev1.ServiceAccount))
+			},
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				oldSA := e.ObjectOld.(*corev1.ServiceAccount)
+				newSA := e.ObjectNew.(*corev1.ServiceAccount)
+				return hasControllerLabel(oldSA) != hasControllerLabel(newSA)
```
krancour, line 53: This is clever.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// SetupReconcilerWithManager initializes a reconciler
+// and registers that reconciler with the provided Manager.
+func SetupReconcilerWithManager(kargoMgr manager.Manager) error {
+	return ctrl.NewControllerManagedBy(kargoMgr).
+		For(&corev1.ServiceAccount{}).
+		WithEventFilter(predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool {
+				return hasControllerLabel(e.Object.(*corev1.ServiceAccount))
```
krancour, line 45: Initial syncs of existing Stages on controller restart are treated as create events, so this filter shouldn't be here or else this could happen:

1. ServiceAccount has appropriate label, receives corresponding permissions.
2. Controller shuts down
3. ServiceAccount has label removed
4. Controller starts up
5. This filter prevents removal of the permissions.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return hasControllerLabel(e.Object.(*corev1.ServiceAccount))
```
krancour, line 48: We're actually not interested in deletes at all.

A "delete" in k8s is initially a soft delete, where the `DeletedAt` field gets set to a non-nil value.

Our reconciler will see that as an _update_.

So the reconcile logic will need to be on the lookout for a non-nil value in that field and treat it like a delete when it sees one.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+
+// ServiceAccountReconciler Reconciles for ServiceAccounts
+type ServiceAccountReconciler struct {
+	cfg ReconcilerConfig
+	client.Client
+}
+
+// SetupReconcilerWithManager initializes a reconciler
+// and registers that reconciler with the provided Manager.
+func SetupReconcilerWithManager(kargoMgr manager.Manager) error {
```
krancour, line 40: Don't you need to be passed a config struct as well?

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+				newSA := e.ObjectNew.(*corev1.ServiceAccount)
+				return hasControllerLabel(oldSA) != hasControllerLabel(newSA)
+			},
+			GenericFunc: func(event.GenericEvent) bool {
+				return false
+			},
+		}).
+		Complete(&ServiceAccountReconciler{
+			Client:         kargoMgr.GetClient(),
+		})
```
krancour, line 61: I recommend creating a contructor-like function like `newReconciler` and using it here. For non-trivial components where there is high potential for mis-configuration, the signature of such a function helps ensure that the calling code has indeed injected anything required and the logic of such a function provides a better/safer guarantee of proper initialization.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+const (
+	ComponentLabelKey = "app.kubernetes.io/component"
+	ControllerLabelValue = "controller"
+)
+
+type ReconcilerConfig struct {
+	KargoNamespace string `envconfig:"KARGO_NAMESPACE" required:"true"`
+}
```
krancour, line 30: You're going to want to add some kind of function like the following to populate one of these structs on-demand:

```
func ReconcilerConfigFromEnv() ReconcilerConfig {
	cfg := ReconcilerConfig{}
	return cfg
}```

File: internal/controller/management/serviceaccount/serviceaccounts.go [hunk 13]
```diff
@@ -0,0 +1,229 @@
...
+	if controllerutil.AddFinalizer(sa, kargoapi.FinalizerName) {
+		if err := r.Update(ctx, sa); err != nil {
+			return ctrl.Result{}, err
+		}
+		logger.Debug("Added finalizer to ServiceAccount", "serviceaccount", sa.Name)
+		return ctrl.Result{}, nil
+	}
+
+	// Handle deletion logic
+	if sa.DeletionTimestamp != nil {
```
krancour, line 96: Ah good. This is already here.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Reconcile handles the reconciliation logic for ServiceAccounts
+func (r *ServiceAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceAccount", req.NamespacedName.Name,
+	)
+	ctx = logging.ContextWithLogger(ctx, logger)
+	logger.Debug("Reconciling ServiceAccounts")
```
krancour, line 71:
```suggestion
	logger.Debug("Reconciling ServiceAccount")
```

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Fetch the ServiceAccount
+	sa := &corev1.ServiceAccount{}
+	if err := r.Get(ctx, req.NamespacedName, sa); err != nil {
+		if kubeerr.IsNotFound(err) {
+			logger.Debug("ServiceAccount not found, deleting role bindings for all projects")
+			// If not found it means the ServiceAccount with the 
+			// controller label was deleted, hence remove Controller
+			// Permissions for the ServiceAccount.
```
krancour, line 80: This comment isn't true because you didn't include any label-based criteria in your `Get()`. Also, I _wouldn't_ add label-based criteria to the `Get()`. At this stage, we don't know if the SA exists or doesn't. If it does, we don't know whether it has the label or not.

We should be getting the ServiceAccount exactly as you do now, drop this comment, and then deal with the following possibilities:

1. Not found --> ensure binding doesn't exist. (This happens if final deletion occurred in-between a create or update event being enqueued and the time the reconciler actually begins processing the reconcilation request.)

1. Found and lacks label OR deletion timestamp is not nil --> ensure the binding doesn't exist; ensure kargo finalizer is cleared.

1. Else (found, has label, has nil deletion timestamp) --> ensure finalizer exists; ensure binding exists.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if controllerutil.AddFinalizer(sa, kargoapi.FinalizerName) {
+		if err := r.Update(ctx, sa); err != nil {
+			return ctrl.Result{}, err
+		}
+		logger.Debug("Added finalizer to ServiceAccount", "serviceaccount", sa.Name)
+		return ctrl.Result{}, nil
+	}
```
krancour, line 93: We shouldn't be doing this unconditionally. As it stand now, you have the possibility of adding the finalizer here only to remove it a few lines down from here.

internal/controller/management/serviceaccount/serviceaccounts.go
krancour: I suggest moving this check up higher in the `Reconcile()` method.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+		return ctrl.Result{}, r.deleteRoleBindingsForAllProjects(ctx, sa)
+	}
+	logger.Debug("Creating role bindings for ServiceAccount",
+		"serviceaccount", sa.Name,
+	)
+	return ctrl.Result{}, r.createRoleBindingForAllProjects(ctx, sa)
+}
+
+// createRoleBindingForAllProjects creates RoleBindings for all projects
+func (r *ServiceAccountReconciler) createRoleBindingForAllProjects(ctx context.Context, sa *corev1.ServiceAccount) error {
```
krancour, line 123: I would call this "ensure" instead of "create."

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+				{
+					Kind:      "ServiceAccount",
+					Name:      sa.Name,
+					Namespace: r.cfg.KargoNamespace,
+				},
+			},
+		}
+
+		// Create or update the RoleBinding		
+		if err := r.Client.Get(ctx, client.ObjectKey{Name: roleBindingName, Namespace: project.Name}, roleBinding); err != nil {
```
krancour, line 164: From here to the end of the method, you have some unnecessary cyclomatic complexity. i.e. Conditions go too many levels deep to easily keep track of. Please refactor this to take advantage of early returns.

1. `err != nil && !kubeerr.IsNotFound(err)` --> return this unanticipated error
1. `err != nil` --> this must have been a 404. Create, but then return when the create is done.
1. At this point, you can treat things unconditionally as an update -- almost -- you _could_ check if the update is necessary, but it's not the end of the world if you don't.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+		err := r.Client.Get(ctx, client.ObjectKey{Name: roleBindingName, Namespace: project.Namespace}, roleBinding)
+		if err != nil {
+			if kubeerr.IsNotFound(err) {
+				logger.Debug("RoleBinding not found, nothing to remove", "roleBinding", roleBindingName, "namespace", project.Namespace)
+				continue // Skip to the next project if not found
+			}
+			return fmt.Errorf("error retrieving RoleBinding %q in namespace %q: %w", roleBindingName, project.Namespace, err)
+		}
+
+		// Delete the RoleBinding if it exists
```
krancour, line 216: Nit on readability. We _already_ dealt with the possibility of the RB not existing, so don't treat this as "if"

```suggestion
		// Delete the RoleBinding
```

internal/controller/management/serviceaccount/serviceaccounts.go
krancour: Although... if you _don't_ check whether the update is necessary, then you could rewind a bit here, skip the get, update unconditionally, and deal with the 404 if it happens. Checking whether the update is necessary is really the only reason to have led off with the get.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,229 @@
...
+	projectList := &kargoapi.ProjectList{}
+	if err := r.Client.List(ctx, projectList); err != nil {
+		return fmt.Errorf("error listing projects: %w", err)
+	}
+
+	for _, project := range projectList.Items {
+		
+		roleBinding := &rbacv1.RoleBinding{}
+
+		err := r.Client.Get(ctx, client.ObjectKey{Name: roleBindingName, Namespace: project.Namespace}, roleBinding)
```
krancour, line 207: Don't bother with the get. Just delete and ignore 404s.

internal/controller/management/projects/projects.go
```diff
@@ -392,6 +400,72 @@ func (r *reconciler) ensureAPIAdminPermissions(
 	return nil
 }
 
+func (r *reconciler) ensureControllerPermissions(
```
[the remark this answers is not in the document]
krancour: > I still need to figure most efficient way to manage RB when Projects are added.

That's what the code below appears to do. Am I missing something?

1. SA added/updated/deleted --> iterate projects and ensure/remove permissions for _that_ SA.

2. Project added/updated (and not fully initialized yet) --> iterate SAs and ensure permissions on _that_ Project.

It seems this PR is already well on the way to covering both of these.

krancour:
You're also going to have to update `cmd/controlplane/management_controller.go` to create a manager for the `kargo` namespace only and use it to initialize the reconciler.

internal/controller/management/serviceaccount/serviceaccounts.go
krancour: It's never _strictly_ needed. It's just a good idea when there's a lot involved in initializing something properly. Right now it may seem there _isn't_ a lot involved in that... but I know how things tend to go with reconcilers over time.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,289 @@
...
+	return ctrl.NewControllerManagedBy(kargoMgr).
+		For(&corev1.ServiceAccount{}).
+		WithEventFilter(
+			predicate.Funcs{
+				CreateFunc: func(e event.CreateEvent) bool {
+					// Allow creation events to be handled for all ServiceAccounts
+					// This ensures any labeled or unlabeled SA gets proper
+					// reconciliation, including on restarts
+					return true
+				},
```
krancour, line 80: iirc, if we don't define `CreateFunc` then no create events are filtered out. I think we only need to specify when there's a chance we do want to ignore something.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,289 @@
...
+					return false
+				},
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					oldSA := e.ObjectOld.(*corev1.ServiceAccount)
+					newSA := e.ObjectNew.(*corev1.ServiceAccount)
+					return hasControllerLabel(oldSA) != hasControllerLabel(newSA) || hasControllerLabel(newSA) && newSA.DeletionTimestamp != nil
+				},
+				GenericFunc: func(event.GenericEvent) bool {
+					return false
+				},
```
krancour, line 91: I don't think there's any possibility of receiving on of these anyway.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,289 @@
...
+func newReconciler(kubeClient client.Client, cfg ReconcilerConfig) *reconciler {
+	r := &reconciler{
+		cfg:    cfg,
+		client: kubeClient,
+	}
+	r.getServiceAccountFn = r.client.Get
+	r.createRoleBindingFn = r.client.Create
+	r.deleteRoleBindingFn = r.client.Delete
+	r.ensureControllerPermissionsFn = r.ensureControllerPermissions
+	r.removeControllerPermissionsFn = r.removeControllerPermissions
```
krancour, line 106: This is one of the compelling reasons to use a constructor-like function as we were previously discussing. This provides a stronger guarantee that initialization was carried out correctly.

This pattern of assigning default implementations of function fields so that they can be overridden for testing purposes is something we should discuss offline tomorrow. We may be moving away from this for _new_ components. Good for now. More to come when we chat.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,289 @@
...
+	logger.Debug("reconciling ServiceAccount")
+
+	sa := &corev1.ServiceAccount{}
+	if err := r.getServiceAccountFn(ctx, req.NamespacedName, sa); err != nil {
+		if kubeerr.IsNotFound(err) {
+			logger.Debug("ServiceAccount not found, deleting RoleBindings for all Projects")
+			// If not found it means the ServiceAccount with the
+			// controller label was deleted, hence remove Controller
+			// Permissions for the ServiceAccount.
+			return ctrl.Result{}, r.removeControllerPermissionsFn(ctx, sa)
```
krancour, line 127: This isn't going to work out the way you hope. If the SA was not found, the `sa` your are passing to `r.removeControllerPermissionsFn()` lacks any details of that SA. It may be that the `removeControllerPermissions()` method needs to be refactored to take a namespaced name as input instead of a SA struct.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,289 @@
...
+				logger.Info("Created RoleBinding for ServiceAccount", "roleBinding", roleBindingName, "namespace", project.Name)
+			} else {
+				return fmt.Errorf("error retrieving RoleBinding %q in namespace %q: %w", roleBindingName, project.Name, err)
+			}
+		} else {
+			// Update RoleBinding
+			if err := r.client.Update(ctx, roleBinding); err != nil {
+				return fmt.Errorf("error updating RoleBinding %q in Project namespace %q: %w", roleBinding.Name, project.Name, err)
+			}
+			logger.Debug("Updated RoleBinding for ServiceAccount", "roleBinding", roleBindingName, "namespace", project.Name)
```
krancour, line 235: I still believe that having bothered with the get above, you you probably check here whether updating this RB is even necessary. (Unnecessary updates just create potential for problems.)

Alternatively, if you _don't_ want to do that, then the get wasn't worth doing. You can jump straight to the create and update if you get an error saying it already exists.

Whichever of these two approaches you prefer is fine with me.

internal/controller/management/serviceaccount/serviceaccounts.go
```diff
@@ -0,0 +1,289 @@
...
+
+		if err := r.deleteRoleBindingFn(ctx, roleBinding); err != nil {
+			if kubeerr.IsNotFound(err) {
+				logger.Debug("RoleBinding not found, skipping deletion", "roleBinding", roleBindingName, "namespace", project.Namespace)
+				continue // Skip to the next project if RoleBinding is not found
+			}
+
+			// Log the error but continue to the next project
+			logger.Error(err, "Failed to delete RoleBinding", "roleBinding", roleBindingName, "namespace", project.Namespace)
+			continue // Continue with the next project even after an error
```
krancour, line 278: This isn't the right thing to do. This can lead to apparent success in cleaning up these RBs when cleanup is, in fact, not complete.

Return the error instead.

When the error is returned from here, the finalizer should not be removed, and the SA should be re-queued for another reconciliation attempt.

krancour:
This is looking much improved, but I'd really like to see unit tests added to these reconcilers. I think you'll find that doing so is a great aid in helping to identify and deal with edge cases.

internal/controller/management/projects/projects.go
```diff
@@ -392,6 +399,66 @@ func (r *reconciler) ensureAPIAdminPermissions(
...
+					roleBinding.Name, sa.Name, project.Name, err)
+			}
+			if err := r.client.Update(ctx, roleBinding); err != nil {
+				return fmt.Errorf("error updating existing RoleBinding %q in Project namespace %q: %w",
+					roleBinding.Name, project.Name, err)
+			}
+			loggerWithSA.Debug("Updated RoleBinding for ServiceAccount", "serviceAccount", sa.Name)
+		} else {
+			loggerWithSA.Debug("Created RoleBinding for ServiceAccount", "serviceAccount", sa.Name)
+		}
```
krancour, line 456:
```suggestion
			loggerWithSA.Debug("Updated RoleBinding for ServiceAccount", "serviceAccount", sa.Name)
			continue
		}
		loggerWithSA.Debug("Created RoleBinding for ServiceAccount", "serviceAccount", sa.Name)

```

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,117 @@
...
+			Name:      saName,
+			Namespace: saNamespace,
+		},
+	}
+
+	// Fake client with ServiceAccount and Project.
+	scheme := runtime.NewScheme()
+	_ = kargoapi.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+	_ = rbacv1.AddToScheme(scheme)
```
krancour, line 59: It's a good practice to assert that even test setup worked as expected.

```suggestion
	err := kargoapi.AddToScheme(scheme)
	require.NoError(t, err)
	err = corev1.AddToScheme(scheme)
	require.NoError(t, err)
	err = rbacv1.AddToScheme(scheme)
	require.NoError(t, err)
```

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,117 @@
...
+	roleBinding := &rbacv1.RoleBinding{}
+	err = kubeClient.Get(context.Background(), types.NamespacedName{
+		Name:      saName + "-readonly-secrets",
+		Namespace: projectName,
+	}, roleBinding)
+	require.NoError(t, err)
+	require.Equal(t, saName+"-readonly-secrets", roleBinding.Name)
+	require.Equal(t, "kargo-controller-secrets-readonly", roleBinding.RoleRef.Name)
+	require.Equal(t, saName, roleBinding.Subjects[0].Name)
+}
```
krancour, line 79: All of the above only tests the path where the RB was created and doesn't test the path where it is updated.

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	err := r.removeControllerPermissions(context.Background(), types.NamespacedName{Name: saName})
+	require.NoError(t, err)
+
+	err = kubeClient.Get(context.Background(), types.NamespacedName{
+		Name:      roleBindingName,
+		Namespace: projectName,
+	}, roleBinding)
+	require.Error(t, err)
+	require.True(t, apierrors.IsNotFound(err), "expected NotFound error but got: %v", err)
```
krancour, line 116: fwiw, we don't often bother with the last argument to require/assert functions, which provides an English description of the failure. It's seldom any more informative than if it were omitted. I don't mind it here, but I figure knowing this may save you some keystrokes in the future.

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,117 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: projectName,
+		},
+	}
+	serviceAccount := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      saName,
+			Namespace: saNamespace,
+		},
+	}
```
krancour, line 53: Do we actually need this for this test?

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Fake client with ServiceAccount and Project.
+	scheme := runtime.NewScheme()
+	_ = kargoapi.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
```
krancour, line 58: Depending on your answer [here](https://github.com/akuity/kargo/pull/2522/files#r1801535893) you may not need this scheme.

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	project := &kargoapi.Project{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: projectName,
+		},
+	}
+
+	scheme := runtime.NewScheme()
+	_ = kargoapi.AddToScheme(scheme)
+	_ = rbacv1.AddToScheme(scheme)
```
krancour, line 100:
```suggestion
	err := kargoapi.AddToScheme(scheme)
	require.NoError(t, err)
	err = rbacv1.AddToScheme(scheme)
	require.NoError(t, err)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	) error
+
+	listProjectFn func(
+		context.Context,
+		client.ObjectList,
+		...client.ListOption,
+	) error
+
+	ensureControllerPermissionsFn func(context.Context, *corev1.ServiceAccount) error
+	removeControllerPermissionsFn func(context.Context, types.NamespacedName) error
```
krancour, line 82: As discussed offline, these can go away as we are avoiding this approach to testing for new components.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+				Name:      roleBindingName,
+				Namespace: project.Name,
+			},
+		}
+
+		if err := r.deleteRoleBindingFn(ctx, roleBinding); err != nil {
+			if kubeerr.IsNotFound(err) {
+				logger.Debug("RoleBinding not found, skipping deletion", "roleBinding",
+					roleBindingName, "namespace", project.Namespace)
+				continue // Skip to the next project if RoleBinding is not found
```
krancour, line 284: I suggest re-wording as this is somewhat misleading. We're not skipping anything. We already attempted the deletion.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+
+			// Return the error to trigger a requeue and stop further cleanup until the issue is resolved.
+			return fmt.Errorf("error deleting RoleBinding %q in Project namespace %q: %w",
+				roleBindingName, project.Namespace, err)
+		}
+
+		logger.Debug("Deleted RoleBinding for ServiceAccount", "roleBinding",
+			roleBindingName, "in namespace", project.Namespace)
+	}
+	logger.Info("Completed deletion of RoleBindings for all Projects")
```
krancour, line 295:
```suggestion
	logger.Debug("Completed deletion of RoleBindings for all Projects")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+// controller has the necessary permissions to manage resources across all Projects.
+func (r *reconciler) ensureControllerPermissions(ctx context.Context, sa *corev1.ServiceAccount) error {
+
+	roleBindingName := fmt.Sprintf("%s-readonly-secrets", sa.Name)
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceaccount", sa.Name,
+		"roleBinding", roleBindingName,
+	)
+	logger.Info("starting to create RoleBindings for all Projects")
```
krancour, line 203:
```suggestion
	logger.Debug("starting to create RoleBindings for all Projects")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+					roleBinding.Name, sa.Name, project.Name, err)
+			}
+			if updateErr := r.updateRoleBindingFn(ctx, roleBinding); updateErr != nil {
+				return fmt.Errorf("error updating existing RoleBinding %q in Project namespace %q: %w",
+					roleBinding.Name, project.Name, updateErr)
+			}
+			logger.Debug("Updated existing RoleBinding for ServiceAccount", "roleBinding",
+				roleBindingName, "namespace", project.Name)
+		} else {
+			logger.Info("Created RoleBinding for ServiceAccount", "roleBinding", roleBindingName, "namespace", project.Name)
```
krancour, line 245:
```suggestion
			logger.Debug("Created RoleBinding for ServiceAccount", "roleBinding", roleBindingName, "namespace", project.Name)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+		}
+	}
+	logger.Info("Completed creating RoleBindings for all Projects")
```
krancour, line 248:
```suggestion
	logger.Debug("Completed creating RoleBindings for all Projects")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+// unauthorized access to resources.
+func (r *reconciler) removeControllerPermissions(ctx context.Context, sa types.NamespacedName) error {
+
+	roleBindingName := fmt.Sprintf("%s-readonly-secrets", sa.Name)
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceaccount", sa.Name,
+		"roleBinding", roleBindingName,
+	)
+	logger.Info("Starting to delete RoleBindings for all Projects")
```
krancour, line 264:
```suggestion
	logger.Debug("Starting to delete RoleBindings for all Projects")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      roleBindingName,
+				Namespace: project.Name,
+			},
+		}
+
+		if err := r.deleteRoleBindingFn(ctx, roleBinding); err != nil {
+			if kubeerr.IsNotFound(err) {
+				logger.Debug("RoleBinding not found, skipping deletion", "roleBinding",
+					roleBindingName, "namespace", project.Namespace)
```
krancour, line 283: I also suggest, a little higher up, around line 272, doing this:

```
projectLogger := logger.WithValues("namespace", project.Name)
```

Then use that throughout the rest of the loop so you get namespace info for free in every message you write with that logger.

Also... `project.Namespace` will always be `""` because Projects themselves are cluster-scoped resources. `project.Name` however, will always match the namespace associated with the Project.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+				continue // Skip to the next project if RoleBinding is not found
+			}
+
+			// Return the error to trigger a requeue and stop further cleanup until the issue is resolved.
+			return fmt.Errorf("error deleting RoleBinding %q in Project namespace %q: %w",
+				roleBindingName, project.Namespace, err)
+		}
+
+		logger.Debug("Deleted RoleBinding for ServiceAccount", "roleBinding",
+			roleBindingName, "in namespace", project.Namespace)
```
krancour, line 293: Example of the benefits of [this](https://github.com/akuity/kargo/pull/2522/files#r1801558400):

```suggestion
		projectLogger.Debug("Deleted RoleBinding")
```

I'll let you adjust the other occurrences.

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,117 @@
...
+	err := r.removeControllerPermissions(context.Background(), types.NamespacedName{Name: saName})
+	require.NoError(t, err)
+
+	err = kubeClient.Get(context.Background(), types.NamespacedName{
+		Name:      roleBindingName,
+		Namespace: projectName,
+	}, roleBinding)
+	require.Error(t, err)
+	require.True(t, apierrors.IsNotFound(err), "expected NotFound error but got: %v", err)
+}
```
krancour, line 117: We should make sure tests also cover the case where the RB already doesn't exist. We want to be sure we _don't_ fail under those circumstances.

internal/controller/management/serviceaccounts/serviceaccounts_test.go
krancour: 👍 Sorry... I briefly had it confused with the corresponding remove method, which only takes a namespaced name.

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// Fake client with ServiceAccount and Project.
+	scheme := runtime.NewScheme()
+	_ = kargoapi.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+	_ = rbacv1.AddToScheme(scheme)
+	kubeClient := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithObjects(serviceAccount, project).
```
krancour, line 62: Remember that this fake client is being built to help you simulate interactions with k8s APIs. The `ensureControllerPermissions()` method won't ever retrieve this SA or any SA. The SA is passed into it. i.e. You don't need to pass it to `WithObjects()` and you also won't need the corev1 scheme.

krancour:
@fykaa I see a lot of recent feedback having been marked resolved with no changes having been made. Are there new commits you accidentally did not push?

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	sa := &corev1.ServiceAccount{}
+	if err := r.getServiceAccountFn(ctx, req.NamespacedName, sa); err != nil {
+		if kubeerr.IsNotFound(err) {
+			logger.Debug("ServiceAccount not found, deleting RoleBindings for all Projects")
+			// If not found it means the ServiceAccount with the
+			// controller label was deleted, hence remove Controller
+			// Permissions for the ServiceAccount.
+			return ctrl.Result{}, r.removeControllerPermissionsFn(ctx, req.NamespacedName)
+		}
+		logger.Error(err, "Failed to get ServiceAccount")
```
krancour, line 150: There's usually no point in logging an error that you're returning. Any error that's returned from a reconciler will get logged, so this just doubles up on the error messages.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+		return ctrl.Result{}, err
+	}
+
+	// Handle ServiceAccount deletion or updates where the
+	// controller label has been removed from the ServiceAccount.
+	// This indicates that the ServiceAccount is no longer managed by the controller,
+	// and we need to clean up any associated RoleBindings.
+	if sa.DeletionTimestamp != nil || !hasControllerLabel(sa) {
+		logger.Debug("Deleting RoleBindings for ServiceAccount", "serviceaccount", sa.Name)
```
krancour, line 159: You already added this information to the logger on lines 135 - 137.

```suggestion
		logger.Debug("Deleting RoleBindings for ServiceAccount")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+		controllerutil.RemoveFinalizer(sa, kargoapi.FinalizerName)
+		if err := r.updateServiceAccountFn(ctx, sa); err != nil {
+			return ctrl.Result{}, err
+		}
+		logger.Debug("Removed finalizer from ServiceAccount", "serviceaccount", sa.Name)
+		return ctrl.Result{}, nil
+	}
+
+	// If we get to here, we had a not found error and we can proceed with
+	// creating the RoleBindings.
```
krancour, line 172:
```suggestion
	// If we get to here, we can proceed with ensuring required RoleBindings exist.
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// Add a finalizer to the ServiceAccount to prevent premature deletion
+	// before RoleBindings are removed. The AddFinalizer function:
+	// - Returns false if the finalizer is already present, avoiding redundant logs.
+	// - Adds the finalizer if not present, ensuring proper cleanup.
```
krancour, line 177:
```suggestion
	// Add a finalizer to ensure this controller always had the opportunity to
	// delete appropriate RoleBindings before final deletion of an SA.
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	if controllerutil.AddFinalizer(sa, kargoapi.FinalizerName) {
+		if err := r.updateServiceAccountFn(ctx, sa); err != nil {
+			return ctrl.Result{}, err
+		}
+		logger.Debug("Added finalizer to ServiceAccount", "serviceaccount", sa.Name)
```
krancour, line 182:
```suggestion
		logger.Debug("Added finalizer to ServiceAccount")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	logger.Debug("Creating RoleBindings for ServiceAccount",
+		"serviceaccount", sa.Name,
+	)
```
krancour, line 188:
```suggestion
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+// controller has the necessary permissions to manage resources across all Projects.
+func (r *reconciler) ensureControllerPermissions(ctx context.Context, sa types.NamespacedName) error {
+
+	roleBindingName := fmt.Sprintf("%s-readonly-secrets", sa.Name)
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceaccount", sa.Name,
+		"roleBinding", roleBindingName,
+	)
+	logger.Debug("starting to create RoleBindings for all Projects")
```
krancour, line 203:
```suggestion
	logger.Debug("Ensuring RoleBindings for all Projects")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+// by creating RoleBindings in each Project's namespace. This function ensures that the
+// controller has the necessary permissions to manage resources across all Projects.
+func (r *reconciler) ensureControllerPermissions(ctx context.Context, sa types.NamespacedName) error {
+
+	roleBindingName := fmt.Sprintf("%s-readonly-secrets", sa.Name)
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceaccount", sa.Name,
+		"roleBinding", roleBindingName,
+	)
```
krancour, line 202:
```suggestion
	logger := logging.LoggerFromContext(ctx).WithValues("roleBinding", roleBindingName)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	return r
+}
+
+// Reconcile is part of the main Kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceAccount", req.NamespacedName.Name,
+	)
```
krancour, line 137:
```suggestion
	logger := logging.LoggerFromContext(ctx).WithValues("serviceAccount", req.NamespacedName.Name)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+			if !kubeerr.IsAlreadyExists(err) {
+				return fmt.Errorf("error creating RoleBinding %q for ServiceAccount %q in Project namespace %q: %w",
+					roleBinding.Name, sa.Name, project.Name, err)
+			}
+			if updateErr := r.updateRoleBindingFn(ctx, roleBinding); updateErr != nil {
+				return fmt.Errorf("error updating existing RoleBinding %q in Project namespace %q: %w",
+					roleBinding.Name, project.Name, updateErr)
+			}
+			logger.Debug("Updated existing RoleBinding for ServiceAccount", "roleBinding",
+				roleBindingName, "namespace", project.Name)
```
krancour, line 243:
```suggestion
			logger.Debug("Updated existing RoleBinding", "namespace", project.Name)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+		} else {
+			logger.Debug("Created RoleBinding for ServiceAccount", "roleBinding", roleBindingName, "namespace", project.Name)
```
krancour, line 245:
```suggestion
			logger.Debug("Created RoleBinding for ServiceAccount", "namespace", project.Name)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	logger.Debug("starting to create RoleBindings for all Projects")
+
+	projectList := &kargoapi.ProjectList{}
+	if err := r.listProjectFn(ctx, projectList); err != nil {
+		return fmt.Errorf("error listing Projects: %w", err)
+	}
+
+	// Loop through each Project to create or update the corresponding RoleBinding.
+	for _, project := range projectList.Items {
+
```
krancour, line 212:
```suggestion

		nsLogger := logger.WithValues("namespace", project.Name)

```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	)
+	return ctrl.Result{}, r.ensureControllerPermissionsFn(ctx, req.NamespacedName)
+}
+
+// ensureControllerPermissions grants the controller ServiceAccount necessary access to all Projects
+// by creating RoleBindings in each Project's namespace. This function ensures that the
+// controller has the necessary permissions to manage resources across all Projects.
+func (r *reconciler) ensureControllerPermissions(ctx context.Context, sa types.NamespacedName) error {
+
+	roleBindingName := fmt.Sprintf("%s-readonly-secrets", sa.Name)
```
krancour, line 197: This same logic occurs in three places. It is probably best to create a small helper function that implements this logic and use that helper in the three places this logic currently occurs. This is meant to prevent the scenario where bugs creep over time due to logic being updated in some locations and not others.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+
+		roleBinding := &rbacv1.RoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      roleBindingName,
+				Namespace: project.Name,
+			},
+			RoleRef: rbacv1.RoleRef{
+				APIGroup: rbacv1.GroupName,
+				Kind:     "ClusterRole",
+				Name:     "kargo-controller-secrets-readonly",
```
krancour, line 221: This occurs in enough places that it probably ought to be a constant. This is meant to prevent the scenario where bugs creep over time due this value being updated in some locations and not others.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+// managed by the controller, ensuring that permissions are cleaned up to prevent
+// unauthorized access to resources.
+func (r *reconciler) removeControllerPermissions(ctx context.Context, sa types.NamespacedName) error {
+
+	roleBindingName := fmt.Sprintf("%s-readonly-secrets", sa.Name)
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceaccount", sa.Name,
+		"roleBinding", roleBindingName,
+	)
```
krancour, line 263:
```suggestion
	logger := logging.LoggerFromContext(ctx).WithValues("roleBinding", roleBindingName)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	logger.Debug("Starting to delete RoleBindings for all Projects")
```
krancour, line 264:
```suggestion
	logger.Debug("Ensuring removal of RoleBindings for all Projects")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+
+	projectList := &kargoapi.ProjectList{}
+	if err := r.listProjectFn(ctx, projectList); err != nil {
+		return fmt.Errorf("error listing Projects: %w", err)
+	}
+
+	for _, project := range projectList.Items {
+
+		projectLogger := logger.WithValues("namespace", project.Name, "roleBinding", roleBindingName)
```
krancour, line 273:
```suggestion
		projectLogger := logger.WithValues("project.namespace", project.Name)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+			}
+
+			// Return the error to trigger a requeue and stop further cleanup until the issue is resolved.
+			return fmt.Errorf("error deleting RoleBinding %q in Project namespace %q: %w",
+				roleBindingName, project.Name, err)
+		}
+
+		projectLogger.Debug("Deleted RoleBinding")
+	}
+	logger.Debug("Completed deletion of RoleBindings for all Projects")
```
krancour, line 295:
```suggestion
	logger.Debug("Ensured removal of RoleBindings for all Projects")
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+
+	logger.Debug("Creating RoleBindings for ServiceAccount",
+		"serviceaccount", sa.Name,
+	)
+	return ctrl.Result{}, r.ensureControllerPermissionsFn(ctx, req.NamespacedName)
+}
+
+// ensureControllerPermissions grants the controller ServiceAccount necessary access to all Projects
+// by creating RoleBindings in each Project's namespace. This function ensures that the
+// controller has the necessary permissions to manage resources across all Projects.
```
krancour, line 194:
```suggestion
// ensureControllerPermissions ensures controllers have necessary access to Secrets
// in all Project namespaces by ensuring the existence of a particular RoleBinding in
// every such namespace.
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	) error
+
+	listProjectFn func(
+		context.Context,
+		client.ObjectList,
+		...client.ListOption,
+	) error
+
+	ensureControllerPermissionsFn func(context.Context, types.NamespacedName) error
+	removeControllerPermissionsFn func(context.Context, types.NamespacedName) error
```
krancour, line 82:
```suggestion
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	r.getServiceAccountFn = r.client.Get
+	r.updateServiceAccountFn = r.client.Update
+	r.createRoleBindingFn = r.client.Create
+	r.updateRoleBindingFn = r.client.Update
+	r.deleteRoleBindingFn = r.client.Delete
+	r.listProjectFn = r.client.List
+	r.ensureControllerPermissionsFn = r.ensureControllerPermissions
+	r.removeControllerPermissionsFn = r.removeControllerPermissions
+	return r
```
krancour, line 128:
```suggestion
	return &reconciler{
		cfg:    cfg,
		client: kubeClient,
	}
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceAccount", req.NamespacedName.Name,
+	)
+	ctx = logging.ContextWithLogger(ctx, logger)
+	logger.Debug("reconciling ServiceAccount")
+
+	sa := &corev1.ServiceAccount{}
+	if err := r.getServiceAccountFn(ctx, req.NamespacedName, sa); err != nil {
```
krancour, line 142:
```suggestion
	if err := r.client.Get(ctx, req.NamespacedName, sa); err != nil {
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+		if kubeerr.IsNotFound(err) {
+			logger.Debug("ServiceAccount not found, deleting RoleBindings for all Projects")
+			// If not found it means the ServiceAccount with the
+			// controller label was deleted, hence remove Controller
+			// Permissions for the ServiceAccount.
+			return ctrl.Result{}, r.removeControllerPermissionsFn(ctx, req.NamespacedName)
```
krancour, line 148:
```suggestion
			return ctrl.Result{}, r.removeControllerPermissions(ctx, req.NamespacedName)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+		return ctrl.Result{}, err
+	}
+
+	// Handle ServiceAccount deletion or updates where the
+	// controller label has been removed from the ServiceAccount.
+	// This indicates that the ServiceAccount is no longer managed by the controller,
+	// and we need to clean up any associated RoleBindings.
+	if sa.DeletionTimestamp != nil || !hasControllerLabel(sa) {
+		logger.Debug("Deleting RoleBindings for ServiceAccount", "serviceaccount", sa.Name)
+		if err := r.removeControllerPermissionsFn(ctx, req.NamespacedName); err != nil {
```
krancour, line 160:
```suggestion
		if err := r.removeControllerPermissions(ctx, req.NamespacedName); err != nil {
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+			return ctrl.Result{}, err
+		}
+		controllerutil.RemoveFinalizer(sa, kargoapi.FinalizerName)
+		if err := r.updateServiceAccountFn(ctx, sa); err != nil {
```
krancour, line 164:
```suggestion
		if err := r.client.Update(ctx, sa); err != nil {
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// If we get to here, we had a not found error and we can proceed with
+	// creating the RoleBindings.
+
+	// Add a finalizer to the ServiceAccount to prevent premature deletion
+	// before RoleBindings are removed. The AddFinalizer function:
+	// - Returns false if the finalizer is already present, avoiding redundant logs.
+	// - Adds the finalizer if not present, ensuring proper cleanup.
+	if controllerutil.AddFinalizer(sa, kargoapi.FinalizerName) {
+		if err := r.updateServiceAccountFn(ctx, sa); err != nil {
```
krancour, line 179:
```suggestion
		if err := r.client.Update(ctx, sa); err != nil {
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+			},
+		}
+
+		// Try to create RoleBinding directly, update if it already exists.
+		if err := r.createRoleBindingFn(ctx, roleBinding); err != nil {
+			if !kubeerr.IsAlreadyExists(err) {
+				return fmt.Errorf("error creating RoleBinding %q for ServiceAccount %q in Project namespace %q: %w",
+					roleBinding.Name, sa.Name, project.Name, err)
+			}
+			if updateErr := r.updateRoleBindingFn(ctx, roleBinding); updateErr != nil {
```
krancour, line 238:
```suggestion
			if updateErr := r.client.Update(ctx, roleBinding); updateErr != nil {
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	roleBindingName := fmt.Sprintf("%s-readonly-secrets", sa.Name)
+
+	logger := logging.LoggerFromContext(ctx).WithValues(
+		"serviceaccount", sa.Name,
+		"roleBinding", roleBindingName,
+	)
+	logger.Debug("Starting to delete RoleBindings for all Projects")
+
+	projectList := &kargoapi.ProjectList{}
+	if err := r.listProjectFn(ctx, projectList); err != nil {
```
krancour, line 267:
```suggestion
	if err := r.client.List(ctx, projectList); err != nil {
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+		projectLogger := logger.WithValues("namespace", project.Name, "roleBinding", roleBindingName)
+
+		roleBinding := &rbacv1.RoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      roleBindingName,
+				Namespace: project.Name,
+			},
+		}
+
+		if err := r.deleteRoleBindingFn(ctx, roleBinding); err != nil {
```
krancour, line 282:
```suggestion
		if err := r.client.Delete(ctx, roleBinding); err != nil {
```

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,115 @@
...
+	require.NotNil(t, kubeClient, r.client)
+
+	require.NotNil(t, r.getServiceAccountFn)
+	require.NotNil(t, r.updateServiceAccountFn)
+	require.NotNil(t, r.createRoleBindingFn)
+	require.NotNil(t, r.updateRoleBindingFn)
+	require.NotNil(t, r.deleteRoleBindingFn)
+	require.NotNil(t, r.listProjectFn)
+	require.NotNil(t, r.ensureControllerPermissionsFn)
+	require.NotNil(t, r.removeControllerPermissionsFn)
```
krancour, line 34:
```suggestion
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,242 @@
...
+	return ctrl.NewControllerManagedBy(kargoMgr).
+		For(&corev1.ServiceAccount{}).
+		WithEventFilter(
+			predicate.Funcs{
+				CreateFunc: func(event.CreateEvent) bool {
+					// Allow creation events to be handled for all ServiceAccounts
+					// This ensures any labeled or unlabeled SA gets proper
+					// reconciliation, including on restarts
+					return true
+				},
```
krancour, line 57:
```suggestion
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,242 @@
...
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					oldSA, _ := e.ObjectOld.(*corev1.ServiceAccount)
+					newSA, _ := e.ObjectNew.(*corev1.ServiceAccount)
+					return hasControllerLabel(oldSA) != hasControllerLabel(newSA) ||
+						hasControllerLabel(newSA) && newSA.DeletionTimestamp != nil
+				},
+				GenericFunc: func(event.GenericEvent) bool {
+					return false
+				},
```
krancour, line 69:
```suggestion
				GenericFunc: func(event.GenericEvent) bool {
					return false
				},
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,242 @@
...
+		WithEventFilter(
+			predicate.Funcs{
+				CreateFunc: func(event.CreateEvent) bool {
+					// Allow creation events to be handled for all ServiceAccounts
+					// This ensures any labeled or unlabeled SA gets proper
+					// reconciliation, including on restarts
+					return true
+				},
+				DeleteFunc: func(event.DeleteEvent) bool {
+					return false
```
krancour, line 59:
```suggestion
					// We are only interested in handling soft deletes
					return false
```

internal/controller/management/projects/projects.go
```diff
@@ -392,6 +406,66 @@ func (r *reconciler) ensureAPIAdminPermissions(
...
+		if err := r.createRoleBindingFn(ctx, roleBinding); err != nil {
+			if !kubeerr.IsAlreadyExists(err) {
+				return fmt.Errorf("error creating RoleBinding %q for ServiceAccount %q in Project namespace %q: %w",
+					roleBinding.Name, sa.Name, project.Name, err)
+			}
+			if err := r.client.Update(ctx, roleBinding); err != nil {
+				return fmt.Errorf("error updating existing RoleBinding %q in Project namespace %q: %w",
+					roleBinding.Name, project.Name, err)
+			}
+			loggerWithSA.Debug("Updated RoleBinding for ServiceAccount", "serviceAccount", sa.Name)
```
krancour, line 460:
```suggestion
			loggerWithSA.Debug("updated RoleBinding for ServiceAccount", "serviceAccount", sa.Name)
```

internal/controller/management/projects/projects.go
```diff
@@ -392,6 +406,66 @@ func (r *reconciler) ensureAPIAdminPermissions(
...
+			continue
+		}
+		loggerWithSA.Debug("Created RoleBinding for ServiceAccount", "serviceAccount", sa.Name)
```
krancour, line 463:
```suggestion
		loggerWithSA.Debug("created RoleBinding for ServiceAccount", "serviceAccount", sa.Name)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
krancour, line 159: A similar message is logged at the start of the method.

```suggestion
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,301 @@
...
+	if sa.DeletionTimestamp != nil || !hasControllerLabel(sa) {
+		logger.Debug("Deleting RoleBindings for ServiceAccount", "serviceaccount", sa.Name)
+		if err := r.removeControllerPermissionsFn(ctx, req.NamespacedName); err != nil {
+			return ctrl.Result{}, err
+		}
+		controllerutil.RemoveFinalizer(sa, kargoapi.FinalizerName)
+		if err := r.updateServiceAccountFn(ctx, sa); err != nil {
+			return ctrl.Result{}, err
+		}
+		logger.Debug("Removed finalizer from ServiceAccount", "serviceaccount", sa.Name)
```
krancour, line 167:
```suggestion
		logger.Debug("removed finalizer from ServiceAccount", "serviceaccount", sa.Name)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
krancour, line 177:
```suggestion
	// Add a finalizer to the ServiceAccount to ensure the opportunity for cleanup
	// upon ServiceAccount deletion.
```

internal/controller/management/serviceaccounts/serviceaccounts.go
krancour, line 182:
```suggestion
		logger.Debug("added finalizer to ServiceAccount", "serviceaccount", sa.Name)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
krancour, line 188: A similar message is logged at the start of the function.

```suggestion
```

krancour:
Closing in favor of #2761

PR #2537, opened 2024-09-16
feat: add git-based promotion directives

krancour:
* clone
* overwrite
* commit
* push
* open pr
* wait for pr

### DIFF
--- a/internal/directives/git_commit_directive.go
+++ b/internal/directives/git_commit_directive.go
@@ -76,7 +76,7 @@ func (g *gitCommitDirective) run(
 		return Result{Status: StatusFailure},
 			fmt.Errorf("error adding all changes to working tree: %w", err)
 	}
-	commitMsg, err := buildCommitMessage(stepCtx.SharedState, cfg)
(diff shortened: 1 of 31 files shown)


internal/directives/git_commit_directive.go
```diff
@@ -32,22 +35,116 @@ func (g *gitCommitDirective) Name() string {
...
+		}
+	}
+	if err = workTree.Commit(commitMsg, commitOpts); err != nil {
+		return Result{Status: StatusFailure},
+			fmt.Errorf("error committing to working tree: %w", err)
+	}
 	return Result{Status: StatusSuccess}, nil
 }
+
+func buildCommitMessage(sharedState State, cfg GitCommitConfig) (string, error) {
```
hiddeco, line 101: I would probably make this a method to ensure we do not get it mixed up with the `generateCommitMessage` methods of the directives which _produce_ Git commit messages as part of their output.

krancour: Fixed in e62495c6b3ec78dc180ffaaff04a8cfdf819a3ad

hiddeco (approved):

PR #2539, opened 2024-09-16
feat(api/controller): Promotion directive steps

hiddeco:
This introduces a `PromotionTemplate` within `Stage` resources, allowing someone to define a sequence of steps (directives) that should be run to promote Freight to a Stage using a Promotion.

```yaml
---
kind: Stage
metadata:
  name: example
spec:
  promotionTemplate:
    spec:
      steps:
      - step: <directive>
      - step: <directive>
```

When this field is defined for a Stage, any previous "Promotion mechanisms" are ignored for newly created Promotions.

---

**Note:** While this pull request makes the feature work from end to end, it still has quite some rough edges, including (nonfunctioning) work directory persistence, result and queue behavior, Stage health checks, etc.


krancour (approved):
+1 to PromotionContext

Issue #1610, opened 2024-03-13
Ability to inflate helm chart to a git path during promotion


jessesuen:
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Today, when using a helm chart subscription, kargo has the following two ways to promote a new chart that is pushed to a helm chart repository:
1. Non-gitops: update a live argocd app spec with the new semvers
2. GitOps: update a subchart in an umbrella chart at a git path

These have the following drawbacks:
* Option 1 is not really gitops, and specific to Argo CD
* While Option 2 is gitops, it forces the use of an umbrella chart, which has the drawback of dependent values being nested under a field in the top values.yaml.

Kargo should provide a third option, which would download and untar the full helm chart to a chosen location in git.

# Motivation

One use case is to avoid the use of the umbrella chart, which seems unnecessary when there is a single chart needs to be promoted.

Users of [Kustomize's chart inflator](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md) would benefit from fully inflated charts in git.

# Suggested Implementation

As an alternative to updating the semvers of a file in git, extract the helm tarball to a specified location.

krancour:
This would be something we can tackle more easily after #2219. i.e. It will just be another built-in step. _Tentatively_ putting this in v0.10.0.

jessesuen:
@hiddeco indicated this could be implemented using `helm pull`. e.g.:

```
helm pull oci://ghcr.io/argoproj/argo-helm/argo-cd --untar
```

So we could just introduce a `helm-pull` step.

That said, if we want to expand this to OCI artifacts, we might want to think bigger.

jessesuen:
> That said, if we want to expand this to OCI artifacts, we might want to think bigger than helm.

Upon further discussion, we think this shouldn't be helm specific (e.g. to support OCI artifacts). So we should introduce:

`oci-download`
`http-download` (or improve existing `http` step)

krancour:
Adding to this that _retagging_ OCI artifacts might be worthwhile as well.

We've heard from a number of users who'd like to retag images at a certain point in their pipeline.

That would work really well combined with [this pattern](https://docs.kargo.io/user-guide/patterns/#pre-pipelines)

koorikla:
[...]

PR #2540, opened 2024-09-18
fix(promotions): multiple stages can sync same argocd app

shmargum:
In the case that two stages are trying to sync the same ArgoCD app (such as an app of apps), there was an infinite cycle of the stages overwriting the `kargo.akuity.io/freight-collection` value in the ArgoCD operation state info.
In my case, this is an ArgoCD promotion mechanism with no source updates (just refreshing an app with gitops updates).

Promotions were never able to complete this way since they were always being requeued.

These changes explicitly prevent this infinite loop by adding stage metadata to the freight collection key.

### DIFF
--- a/internal/controller/promotion/argocd.go
+++ b/internal/controller/promotion/argocd.go
@@ -57,6 +57,7 @@ type argoCDMechanism struct {
 		desiredSource *argocd.ApplicationSource,
 		desiredSources argocd.ApplicationSources,
 		freightColID string,
+		stage *kargoapi.Stage,
 	) error
 	getAuthorizedApplicationFn func(
 		ctx context.Context,
@@ -201,6 +202,7 @@ func (a *argoCDMechanism) Promote(
 			desiredSource,
 			desiredSources,
 			promo.Status.FreightCollection.ID,
+			stage,
 		); err != nil {
 			return err
 		}
@@ -301,7 +303,12 @@ func (a *argoCDMechanism) mustPerformUpdate(
 	// current freight collection. i.e. Not related to the current promotion.
 	var correctFreightColIDFound bool
 	for _, info := range status.Operation.Info {
-		if info.Name == freightCollectionInfoKey {
+		if info.Name == fmt.Sprintf(
+			"%s/%s/%s",
+			freightCollectionInfoKey,
+			stage.ObjectMeta.Namespace,
+			stage.ObjectMeta.Name,
+		) {
 			correctFreightColIDFound = info.Value == freightCol.ID
 			break
 		}
@@ -374,6 +381,7 @@ func (a *argoCDMechanism) syncApplication(
 	desiredSource *argocd.ApplicationSource,
 	desiredSources argocd.ApplicationSources,
 	freightColID string,
+	stage *kargoapi.Stage,
 ) error {
 	// Initiate a "hard" refresh.
 	if app.ObjectMeta.Annotations == nil {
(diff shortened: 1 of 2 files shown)


krancour:
@shmargum thanks for the PR.

I just wanted to respond that it might be a little bit before we get to looking at this. The issue isn't quite as simple as it may seem on the surface.

I think this may work for your use case, but in cases where the Argo CD App updates are actually mutating something _or_ cases where the health checks on a Stage involve asserting that the App is synced to some specific, desired revision, I _think_ this is going to not work.

In general, the notion of multiple Stages updating the same App is not something we anticipated, and tbh, I have my doubts that it's a generally advisable pattern. Stages are, roughly speaking, "environments" and it seems, at the very least, _odd_ that two or more distinct environments would be managed by a single App.

At any rate, I'm not by any means dismissing this PR out of hand. As I said, I just wanted to respond sooner than later and let you know that this is going to require some careful thought on our part before we can commit to taking this.

shmargum:
Just want to give some more context about why I am doing this in the first place.

I'd like to keep all my state in git, and so ArgoCD source updates are out of the question.
This leaves umbrella helm chart as a documented approach for gitops style updates with noted drawbacks in #1610 

We've taken another approach, where we publish helm charts as OCI artifacts, and use the image promotion process to promote the helm chart "image" as a value in git, which configure child apps.

This means that we are also doing 2 ArgoCD updates, first one for the parent app, then a refresh of the child app.
Currently, the parent app holds all the children apps for a service (environments).
I suppose I could work around this further by adding an intermediate level of parent app.  I think a similar approach would be needed for umbrella charts as well.

krancour:
> I'd like to keep all my state in git, and so ArgoCD source updates are out of the question.

Yep. That's ideal. Source updates are not real GitOps.

> Currently, the parent app holds all the children apps for a service (environments).

If you're doing App of Apps, why is Kargo ever having to sync the parent app? I would imagine that the Apps managed by the parent App are fairly static in that their definitions must rarely change. Why doesn't each Stage just sync relevant children?

To be transparent, I don't _completely_ understand the approach you've described, but the notions of multiple Stages all syncing the same App or any Stage syncing a parent in App of Apps still feels vaguely suspect.

krancour:
Ok. I understand better what you are doing now. Thanks for explaining. 😄 

> The parent app which deploys the child apps sets the "target revision" of the child apps.

That is, for sure, a novel approach to mutating the target revisions whilst still being real GitOps.

What I am, perhaps, not understanding is why use that somewhat complex approach to begin with? Would not all your state _still_ be in your repositories if the child apps all tracked the head of various branches and Kargo managed what's at the head of each branch?

krancour:
> In order to achieve both 2 and 3 we need the base helm chart and environment specific values to be independent (we can't bundle those with the base chart or else they would go through the promotion process)

I think you're going to love the upcoming release. We've been building toward doing things exactly like this without so much hassle. In a previous release, we added the possibility of Stages subscribing to Freight from multiple Warehouses. In v0.9.0, we're adding far more flexibility to the promotion processes. It's a bit reminiscent of GitHub Actions.

For your scenario, what you'll be able to do at each Stage is consume multiple pieces of Freight (i.e. artifacts _not_ bundles together in one box) and describe much more explicitly how you would like to combine those artifacts.

krancour:
I'm going to close this as I think we've established (through occurrences besides this one, as well) that multiple Stages attempting to manage a single Argo CD App is asking for trouble.

@shmargum, v0.9.0 has been released now, and I am hopeful that all its new flexibility makes workflows possible for you that won't require you to attempt managing one Application using multiple Stages.

shmargum:
While I still think it is a common pattern to have apps that manage other apps, I also agree that Kargo trying to manage a single app with multiple stages sounds like trouble.
I've actually just went ahead and created a separate parent app per actual app.
This does nearly double the number of ArgoCD Application objects, but hey, things work :)

krancour:
@shmargum I will continue thinking about how to make this use case work.

I'm in complete agreement that it _makes sense._

PR #2545, opened 2024-09-18
feat(controller): add argocd-update directive

krancour:
This adds the `argocd-update` directive.

It's a lot of code, _but_ a vast majority of it is copied from the existing `internal/controller/promotion` package and then _minimally_ adapted to work with the configs for directive, which have been generated from JSON schema.

This means this logic is mostly tried and true and also means we have pretty decent test coverage.

Note there was no attempt to make this code common and shared between the older promotion mechanisms and this directive since it is anticipated that the older promotion mechanisms will be removed within a couple releases anyway.

Unsurprisingly, configuration should look remarkably similar to the existing Argo CD promotion mechanism:

```yaml
steps:
- step: argocd-update
  config:
    apps:
    - name: my-app
      namespace: my-namespace (optional; defaults to controller's configured default -- usually argocd)
      sources:
      - repoURL: https://github.com/example/repo.git
        updateTargetRevision: true
        helm:
          images:
          - repoURL: nginx
            key: image.tag
            value: Tag
        kustomize:
          images:
          - repoURL: nginx
            newName: foo
            useDigest: true
```

### DIFF
--- a/internal/directives/argocd_update_directive.go
+++ b/internal/directives/argocd_update_directive.go
@@ -1,0 +1,1104 @@
+package directives
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/gobwas/glob"
+	"github.com/xeipuuv/gojsonschema"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	libargocd "github.com/akuity/kargo/internal/argocd"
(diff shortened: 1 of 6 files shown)


internal/directives/argocd_update_directive.go
```diff
@@ -0,0 +1,1104 @@
...
+	// sources.
+	message := "initiated sync"
+	if app.Spec.Source != nil {
+		message += " to " + app.Spec.Source.TargetRevision
+	}
+	a.logAppEventFn(
+		ctx,
+		stepCtx,
+		app,
+		"kargo-controller",
```
hiddeco, line 539: Appears like this could make use of `applicationOperationInitiator`?

krancour: Fixes in 18dabd66be1c30110ced578cbf5ea2a54e55d1b1

hiddeco (approved):
This will be extremely high on my list to refactor, as the shape of things makes my hands itch.

But functionality-wise, this looks good to me, and has been proven to work well 💯

PR #2549, opened 2024-09-19
feat(ui): display promotion steps and YAML

rbreeze:
Screenshot 2024-09-30 at 4 57 33 PM (2)
Screenshot 2024-09-30 at 4 56 40 PM (2)

Screenshot 2024-09-30 at 4 58 59 PM (2)


ui/src/features/stage/test-directives.ts
```diff
@@ -0,0 +1,152 @@
+export const testDirectives = [
```
rbreeze, line 1: Examples to test feature while waiting for finalized promotion directives

Marvin9:
@krancour would you please add more context in this task? or we can discuss offline

Marvin9:
@hiddeco @krancour Do we have rough idea at this point, how the community (read non-built-ins) runners will be installed?

krancour:
> Do we have rough idea at this point, how the community (read non-built-ins) runners will be installed?

We had a three part plan for cutting over to step-based (directive-based) promotions, then containerizing the built-in steps, then finally opening things up to third-party step runners.

I think the exact installation/registration process is something that might currently be absent from that plan. Let's connect about this offline.

ui/src/features/common/promotion-directive-step-status/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+import { PromotionStatus } from '@ui/gen/v1alpha1/generated_pb';
+
+// UI concludes from Promotion's status data
+export enum PromotionDirectiveStepStatus {
+  RUNNING,
+  FAILED,
+  SUCCESS
+}
+
+export const getPromotionDirectiveStepStatus = (
```
Marvin9, line 11: @krancour / @hiddeco can you please read this logic if anything is inaccurate?

Basically for any step ran, UI tries to decide what is state of all steps

ui/src/features/common/promotion-directive-step-status/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+) => {
+  if (
+    promotionStatus?.phase === PromotionStatusPhase.RUNNING &&
+    stepNumber === Number(promotionStatus?.currentStep)
+  ) {
+    return PromotionDirectiveStepStatus.RUNNING;
+  }
+
+  // TODO: controller should point 'currentStep' to exact failed step
+  // at the moment, it is defaulted to 0
```
krancour, line 23: I can fix that right now.

internal/directives/simple_engine.go
```diff
@@ -136,7 +136,7 @@ func (e *SimpleEngine) Promote(
 	return PromotionResult{
 		Status:           PromotionStatusSuccess,
 		HealthCheckSteps: healthCheckSteps,
-		CurrentStep:      0,
+		CurrentStep:      int64(len(steps)),
```
hiddeco, line 139: Shouldn't this be `len(steps) - 1`?

ui/src/features/common/promotion-directive-step-status/utils.ts
krancour: Actually, I could only find this being set to zero if the steps couldn't even get started or after success. I updated it so it's _not_ set back to zero on success.

I cannot seem to find where a failed step is getting this set back to zero.

krancour:
@Marvin9 this looks pretty good. The one bug I am observing is that it seems when the current step is pending, previous steps won't show up as successful:

Screenshot 2024-10-01 at 11 03 33 AM

As soon as the last step turns green, they all do. 😄 

I think what we should do is assume success for all steps prior to the current one.

internal/directives/simple_engine.go
krancour: Uh. Sure should.

krancour:
```suggestion
		CurrentStep:      int64(len(steps)-1),
```

krancour:
I think there is still a small issue here. It seems that on failure, all previous steps are being displayed as failed.

Here's a screenshot:

Screenshot 2024-10-02 at 10 01 11 AM

Fundamentally, I believe the UI can assume that if we're on step n > 0, then all steps < n have succeeded.

krancour (approved):

PR #2550, opened 2024-09-19
fix(ui): stage/freight live manifest scrolls infinitely

### LINKED ISSUE #2523: bug(ui): live manifest of stage has ever-expanding text box
When selecting the live manifest tab for a Stage, you will see that the text box where the YAML is displayed grows infinitely tall.

I'm observing this in both Firefox and Chrome (I didn't check other browsers). The issue occurs in both, but is more noticeable in Firefox.

Live manifest views of other resource types seem to be ok.

rbreeze:
Fixes #2523


krancour (approved):

Marvin9 (approved):
Tested this in stage and freight live manifest view and LGTM!

PR #2554, opened 2024-09-22
refactor: make engine and directive interfaces and related types support the notion of healthchecks

krancour:
Refactors the Engine and "directives" (the word actually no longer appears anywhere except in the package name) to generically support the concept of health checks. (Open to a better package name if we can think of one.)

What were formerly called "directives" are now components with varied names (that reflect their functionality) that implement either `PromotionStepRunner`, `HealthCheckStepRunner`, or both.

The built-in registry now supports registration of both types of steps.

On success, promotion steps can return configuration for a corresponding health check. That configuration is completely opaque to the Engine (and will be to the Stage reconciler as well, although those bits will be updated in a follow-up).

The Engine executes health checks in sequence and aggregates their results, including opaque output.

@hiddeco only the last two commits are new since you last looked at this.


internal/directives/engine.go
```diff
@@ -6,42 +6,30 @@ import (
...
+	// Name returns the name of the Directive.
+	Name() string
+	// RunPromotionStep executes an individual step of a user-defined promotion
+	// process using the provided PromotionStepContext. Implementations may
+	// indirectly modify that context through the returned PromotionStepResult to
+	// allow subsequent promotion steps to access the results of its execution.
+	RunPromotionStep(context.Context, *PromotionStepContext) (PromotionStepResult, error)
+	// RunHealthCheckStep executes a health check using the provided
+	// HealthCheckStepContext.
+	RunHealthCheckStep(context.Context, *HealthCheckStepContext) HealthCheckStepResult
```
hiddeco, line 34: Wonder if we should split this up into e.g. `Promotable` and `Healthable` interfaces, which would allow "partial" implementations.

krancour: That's a great idea. Consider it done.

hiddeco (approved):

PR #2564, opened 2024-09-23
docs: improve kargo concepts docs

nitishfy:
Improve Kargo concept docs to make the concepts slightly easier to understand

### DIFF
--- a/docs/docs/15-concepts.md
+++ b/docs/docs/15-concepts.md
@@ -37,10 +37,9 @@ more "prod" stages at the end.
 
 ### What is Freight?
 
-**Freight** is Kargo's second most important concept. Frieight generally refers
-to a collection of artifacts that gets promoted together to the next stage.
-A single "piece of freight" is a set of references to one or more versioned
-artifacts, which may include one or more:
+**Freight** is Kargo's second most important concept. A single "piece of
+freight" is a set of references to one or more versioned artifacts, which may
+include one or more:
 
 * Container images (from image repositories)
 
--- a/docs/docs/15-concepts.md
+++ b/docs/docs/15-concepts.md
@@ -52,10 +52,7 @@ Kargo seeks to progress from one stage to another.
 
 ### What is a Warehouse?
 
-A **warehouse** is a _source_ of freight. It is where a collection of 'Freight'
-gets stored before the 'Freight' is deployed anywhere.
-
-A warehouse subscribes to one or more:
+A **warehouse** is a _source_ of freight. A warehouse subscribes to one or more:
 
 * Container image repositories
 
--- a/docs/docs/15-concepts.md
+++ b/docs/docs/15-concepts.md
@@ -19,8 +19,10 @@ for automatic promotions of new **freight**.
 ### What is a Stage?
 
 When you hear the term “environment”, what you envision will depend
-significantly on your perspective. For a developer, the term "environment" will
-have a different meaning while for a Devops engineer, the meaning will be different.
+significantly on your perspective. A developer, for example, may think of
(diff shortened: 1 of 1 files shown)


docs/docs/15-concepts.md
```diff
@@ -19,10 +19,12 @@ for automatic promotions of new **freight**.
 ### What is a Stage?
 
 When you hear the term “environment”, what you envision will depend
-significantly on your perspective. To eliminate confusion, Kargo avoids the term
-"environment" altogether in favor of **stage**. The important feature of a stage
-is that its name ("test" or "prod," for instance) denotes an application
-instance's _purpose_ and not necessarily its _location_.
+significantly on your perspective. For a developer, the term "environment" will
+have a different meaning while for a Devops engineer, the meaning will be different.
```
krancour, line 23: I think this can be more clear, still.

```suggestion
significantly on your perspective. A developer, for example, may think of
an "environment" as a specific _instance_ of an application they work on,
while a DevOps engineer, may think of an "environment" as a particular segment
of the infrastructure they maintain.
```

docs/docs/15-concepts.md
```diff
@@ -33,9 +37,10 @@ more "prod" stages at the end.
 
 ### What is Freight?
 
-**Freight** is Kargo's second most important concept. A single "piece of
-freight" is a set of references to one or more versioned artifacts, which may
-include one or more:
+**Freight** is Kargo's second most important concept. Frieight generally refers
+to a collection of artifacts that gets promoted together to the next stage.
+A single "piece of freight" is a set of references to one or more versioned
+artifacts, which may include one or more:
```
krancour, line 43: tbh, I feel this change adds more confusion. We should be careful of unnecessarily using words like "generally," as it implies there are exceptions to what you're about to say. Freight isn't _generally_ a collection of artifacts. Freight _is_ a collection of artifacts. There isn't an exception to that.

The bit about Freight being what moves from stage to stage is already stated just a few lines down from here.

imho, I would revert this change.

```suggestion
**Freight** is Kargo's second most important concept. A single "piece of
freight" is a set of references to one or more versioned artifacts, which may
include one or more:
```

docs/docs/15-concepts.md
```diff
@@ -48,7 +53,10 @@ Kargo seeks to progress from one stage to another.
 
 ### What is a Warehouse?
 
-A **warehouse** is a _source_ of freight. A warehouse subscribes to one or more:
+A **warehouse** is a _source_ of freight. It is where a collection of 'Freight'
+gets stored before the 'Freight' is deployed anywhere.
+
+A warehouse subscribes to one or more:
```
krancour, line 59: Even in the real word, I think warehouses are a somewhat fuzzy concept. Is a warehouse a sort of archive where physical things are stored, possibly long-term? Or is it more of a fulfillment/shipping hub? It can be one or the other, or both.

Kargo warehouses are really about fulfillment. I'm not sure it would be accurate to say they _store_ freight. Freight are instances of a CRD. So, really, _Kubernetes_ stores your Freight. The first few sections of this page stay away from technical details. CRDs are not mentioned until later. But I think the statement that warehouses _store_ freight may set the reader up for confusion when they do read about the CRD later.

docs/docs/15-concepts.md
```diff
@@ -48,7 +52,10 @@ Kargo seeks to progress from one stage to another.
...
```
krancour, line 58:
```suggestion
A **warehouse** is a _source_ of freight. A warehouse subscribes to one or more:
```

krancour:
Taking parts of this. TY!

krancour (approved):

Issue #2384, opened 2024-08-05
ui intermittently shows current freight as empty


alexmt:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.

# Description
Looks like https://github.com/akuity/kargo/issues/2094 again. UI intermittently incorrectly shows that the stage has no freight. Browser refresh resolves it.

# Screenshots

<img width="1301" alt="Screenshot 2024-08-05 at 8 42 02 AM" src="https://github.com/user-attachments/assets/b37d4e52-53f0-4b67-8dca-2bc70b20c060">

# Steps to Reproduce

Issue sometimes happens after opening kargo UI, switching to the different tab and back.

# Version

```
v0.8.2
```

rbreeze:
@alexmt I can't seem to reproduce this, can you attempt to reproduce in `main`? This way if it is fixed as I suspect we can isolate the portion that fixed it and back port it. `main` contains a large refactor/redesign PR which I believe fixed the issue so want to confirm this.

danielloader:
Happens to me when I have kargo in a background idle tab, and multiple promotions have taken place in that state and then going back to the tab I can see the freight on the line but it just says empty on the stage - refreshing the page gives the correct information.

PR #2568, opened 2024-09-24
fix(ui): empty freights in some cases

Marvin9:
fixes https://github.com/akuity/kargo/issues/2384

I was able to [reproduce](https://github.com/akuity/kargo/issues/2384#issuecomment-2361824549). I have tested and now UI won't show stages with empty freight if they have freight.

### DIFF
--- a/ui/src/features/project/pipelines/pipelines.tsx
+++ b/ui/src/features/project/pipelines/pipelines.tsx
@@ -237,6 +237,21 @@ export const Pipelines = ({
     return freightMap;
   }, [freightData]);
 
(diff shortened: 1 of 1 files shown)


Marvin9:
@krancour to confirm, we don't have situation where stages have freights but they don't exist. To put it other way, once stage is associated with freights, those freights cannot be deleted.

krancour:
@Marvin9 correct. We use a validating webhook to block deletion of any Freight that is in-use.

krancour (approved):

Issue #2529, opened 2024-09-13
migrate sso user <--> serviceaccount mapping annotations


krancour:
#2315 changed the annotation keys that are used for mapping SSO users to ServiceAccounts.

Before:

* `rbac.kargo.akuity.io/sub`
* `rbac.kargo.akuity.io/email`
* `rbac.kargo.akuity.io/groups`

After:

* `rbac.kargo.akuity.io/claim.<any claim>`

As we've done in previous releases, we can automate the process of compensating for this breaking change.

Let's plan to include migration logic for this in the `release-0.9` branch (once it exists). This functionality does not need to be included in the `main` branch.

@fykaa, @hiddeco or myself will help you get started on this as we get closer to the v0.9.0 release. We have an established pattern for how this is done.

krancour:
@fykaa you can get started on this.

This can be addressed by creating a new reconciler for the _management_ controller. It will have to reconcile ServiceAccounts that have annotations with _any_ of the following keys:

* rbac.kargo.akuity.io/sub
* rbac.kargo.akuity.io/email
* rbac.kargo.akuity.io/groups

_Replace_ them, respectively, with:

* rbac.kargo.akuity.io/claim.sub
* rbac.kargo.akuity.io/claim.email
* rbac.kargo.akuity.io/claim.groups

Although the keys will change, the values should remain the same.

Note that it's not actually _intrinsically_ possible to watch resources based on annotations, but it can effectively be accomplished by watching all ServiceAccounts but using a predicate to filter out ones that do not have any of the older annotation keys. You should be able to find examples of predicates in most of the existing reconcilers.

krancour:
#2579 addressed this, but GH did not close this issue automatically since the PR was merged to a release branch and not main. All set.

PR #2579, opened 2024-09-26
feat: migrate sso user, serviceaccount mapping annotations

fykaa:
Closes: https://github.com/akuity/kargo/issues/2529


krancour:
Note: This must not be merged into `main`. It must be merged into the `release-0.9` branch, which does not exist yet, so `main` is a placeholder for now, I've applied the "do not merge yet" label, and once `release-0.9` exists, we can update this PR to use that branch as the base.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+	OldAnnotationKeySub    = "rbac.kargo.akuity.io/sub"
+	OldAnnotationKeyEmail  = "rbac.kargo.akuity.io/email"
+	OldAnnotationKeyGroups = "rbac.kargo.akuity.io/groups"
+)
+
+// ServiceAccountReconciler Reconciles for ServiceAccounts that need annotation migration
+type ServiceAccountReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+	Logger *logging.Logger
```
krancour, line 29: Let's not have `Logger` be a field. We can use `logging.LoggerFromContext(ctx)` to pluck a properly configured logger out of thin air wherever we need it. (And often the context-bound logger has had additional details added to it by some caller, so our logs tend to be more information-rich if we stick to using that.)

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+const (
+	OldAnnotationKeySub    = "rbac.kargo.akuity.io/sub"
+	OldAnnotationKeyEmail  = "rbac.kargo.akuity.io/email"
+	OldAnnotationKeyGroups = "rbac.kargo.akuity.io/groups"
+)
+
+// ServiceAccountReconciler Reconciles for ServiceAccounts that need annotation migration
+type ServiceAccountReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
```
krancour, line 28: Why are we holding onto this?

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+		WithEventFilter(predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				return hasOldAnnotations(e.ObjectNew.GetAnnotations())
+			},
+			CreateFunc: func(e event.CreateEvent) bool {
+				return hasOldAnnotations(e.Object.GetAnnotations())
+			},
+			GenericFunc: func(e event.GenericEvent) bool {
+				return hasOldAnnotations(e.Object.GetAnnotations())
+			},
```
krancour, line 45: This should probably return false. Generic events probably won't happen for a ServiceAccount, but if one does somehow, it's not striking me as an event we need to be interested in.

Additionally, we definitely don't have anything to do on delete, so we should also add `DeleteFunc` that unconditionally returns false.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+			return ctrl.Result{}, nil
+		}
+		r.Logger.Error(err, "Failed to get ServiceAccount")
+		return ctrl.Result{}, err
+	}
+
+	annotations := sa.GetAnnotations()
+	if annotations == nil {
+		return ctrl.Result{}, nil
+	}
```
krancour, line 65: We can probably leave this early return out. We know this reconciliation wouldn't have been triggered in there first place if there were no annotations and if _somehow_ this code were executed without any annotations defined (e.g. in the context of a unit test, perhaps), everything that follows is still safe. So I think we can lose these three lines to just keep the logic shorter and more readable.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+		delete(annotations, OldAnnotationKeyEmail)
+		updated = true
+	}
+	if val, ok := annotations[OldAnnotationKeyGroups]; ok {
+		annotations[rbacapi.AnnotationKeyOIDCClaim("groups")] = val
+		delete(annotations, OldAnnotationKeyGroups)
+		updated = true
+	}
+
+	if updated {
```
krancour, line 84: I don't think we need a variable to keep track of whether any annotations were updated. Again, this function only would have been called if there were an update to perform, and even if somehow it were called under other circumstances, everything inside this conditional is safe to do.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+		sa.SetAnnotations(annotations)
+		if err := r.Update(ctx, sa); err != nil {
+			r.Logger.Error(err, "Failed to update ServiceAccount annotations")
+			return ctrl.Result{}, err
```
krancour, line 88: I wouldn't bother to explicitly log this here. Any errors returned from the reconciler will be logged automatically, but I would take the context that you're including in the log message and add it to the returned error instead:

```suggestion
			return ctrl.Result{}, fmt.Errorf("failed to update ServiceAccount annotations: %w", err)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+// Reconcile handles the reconciliation logic for migrating annotations
+func (r *ServiceAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	sa := &corev1.ServiceAccount{}
+	err := r.Get(ctx, req.NamespacedName, sa)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		r.Logger.Error(err, "Failed to get ServiceAccount")
+		return ctrl.Result{}, err
```
krancour, line 59: Similar to [here](https://github.com/akuity/kargo/pull/2579/files#r1777293104)

```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to get ServiceAccount: %w", err)
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+
+// ServiceAccountReconciler Reconciles for ServiceAccounts that need annotation migration
+type ServiceAccountReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+	Logger *logging.Logger
+}
+
+// SetupWithManager sets up the reconciler with the manager
+func (r *ServiceAccountReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
krancour, line 33:
```suggestion
// SetupReconcilerWithManager initializes a reconciler for ServiceAccount resources
// and registers it with the provided Manager.
func SetupReconcilerWithManager(mgr manager.Manager) error {
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+	return hasSub || hasEmail || hasGroups
+}
+
+func SetupServiceAccountReconcilerWithManager(mgr ctrl.Manager) error {
+	return (&ServiceAccountReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+		Logger: logging.NewLogger(logging.InfoLevel),
+	}).SetupWithManager(mgr)
+}
```
krancour, line 113: Not needed after [this](https://github.com/akuity/kargo/pull/2579/files#r1777309004)

```suggestion
```

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
...
+	err := r.Get(ctx, req.NamespacedName, sa)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		r.Logger.Error(err, "Failed to get ServiceAccount")
+		return ctrl.Result{}, err
+	}
+
+	annotations := sa.GetAnnotations()
```
krancour, line 62: It should be safe to just modify `sa.Annotations` in place.

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1 @@
+package serviceaccounts
```
krancour, line 1: This particular reconciler should be pretty easy to test. Let's make sure we add some test cases here.

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,113 @@
+package serviceaccounts
```
krancour, line 1: For no other reason than consistency with how we have done upgrade reconcilers in the past, I would probably change the package name to `upgrade`.

internal/controller/management/serviceaccounts/serviceaccounts.go
krancour: Let me clarify... I wasn't objecting to the use of GetAnnotations(). I was meaning to say that there was no need for annotations:= sa.GetAnnotation(). The modifications you subsequently make to annotations could be made directly to sa.Annotations and that would read a bit cleaner.

krancour:
This looks good. Let's try to get some unit tests added and it will be g2g

internal/controller/management/upgrade/serviceaccounts.go
```diff
@@ -0,0 +1,98 @@
...
+
+	rbacapi "github.com/akuity/kargo/api/rbac/v1alpha1"
+	"github.com/akuity/kargo/internal/logging"
+)
+
+// Old annotation keys
+const (
+	OldAnnotationKeySub    = "rbac.kargo.akuity.io/sub"
+	OldAnnotationKeyEmail  = "rbac.kargo.akuity.io/email"
+	OldAnnotationKeyGroups = "rbac.kargo.akuity.io/groups"
```
krancour, line 22: Nit: There's no reason to export these. This comment is non-blocking. It is up to you whether it's worth addressing.

internal/controller/management/upgrade/serviceaccounts.go
```diff
@@ -0,0 +1,98 @@
...
+	if val, ok := sa.Annotations[OldAnnotationKeyEmail]; ok {
+		sa.Annotations[rbacapi.AnnotationKeyOIDCClaim("email")] = val
+		delete(sa.Annotations, OldAnnotationKeyEmail)
+	}
+	if val, ok := sa.Annotations[OldAnnotationKeyGroups]; ok {
+		sa.Annotations[rbacapi.AnnotationKeyOIDCClaim("groups")] = val
+		delete(sa.Annotations, OldAnnotationKeyGroups)
+	}
+
+	sa.SetAnnotations(sa.Annotations)
```
krancour, line 80: Now that you're updating `sa.Annotations` in place, you don't need to do this.

```suggestion
```

internal/controller/management/upgrade/serviceaccounts_test.go
```diff
@@ -0,0 +1,137 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			result := hasOldAnnotations(tt.annotations)
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestServiceAccountReconciler_Reconcile(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = corev1.AddToScheme(scheme)
```
krancour, line 56: It's a good idea in our unit tests to test assumptions as well as outcomes. An failure in test setup could affect the outcome of the test cases. If the setup fails silently, we end up with a harder time tracking down the root cause.

```suggestion
	err := corev1.AddToScheme(scheme)
	require.NoError(t, err)
```

internal/controller/management/upgrade/serviceaccounts_test.go
```diff
@@ -0,0 +1,137 @@
...
+		},
+	}
+
+	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(sa).Build()
+
+	reconciler := &ServiceAccountReconciler{
+		Client: client,
+	}
+
+	ctx := context.TODO()
```
krancour, line 78: Nit: Prefer `context.Background()` if there isn't genuinely a need to circle back and provider a different context.

```suggestion
	ctx := context.Background()
```

internal/controller/management/upgrade/serviceaccounts_test.go
```diff
@@ -0,0 +1,137 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := hasOldAnnotations(tt.annotations)
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
```
krancour, line 137: This seems to duplicate the tests in `TestHasOldAnnotations()`.

Personally, I find this piece of setup for each reconciler difficult to test and not entirely worth the effort. I would suggest just removing this particular test.

```suggestion
```

krancour:
@fykaa this is in great shape. There are only a few very tiny things left to address. Solid work.

krancour (approved):

PR #2580, opened 2024-09-26
feat(controller): use directives engine for Stage health checks

hiddeco:


internal/directives/config.go
```diff
@@ -23,6 +23,11 @@ func (c Config) DeepCopy() Config {
 	return runtime.DeepCopyJSON(c)
 }
 
+// ToJSON marshals the configuration to JSON.
+func (c Config) ToJSON() ([]byte, error) {
+	return json.Marshal(c)
+}
```
krancour, line 29: I'm curious what the rationale for this was. I'm sure you had a reason that you wanted to call this somewhere instead of calling `json.Marshal()` directly?

internal/controller/stages/stages.go
```diff
@@ -439,8 +441,13 @@ func newReconciler(
 	r := &reconciler{
 		kargoClient:  kargoClient,
 		argocdClient: argocdClient,
-		recorder:     recorder,
-		cfg:          cfg,
+		directivesEngine: directives.NewSimpleEngine(
+			nil, // TODO: is this required here?
```
krancour, line 445: That nil is the creds db, right?

I don't want to rule out the idea that third-party directives might require access to credentials.

imho, the best bet is to instantiate just one engine in the controller startup logic and inject that into both of the reconcilers that use it.

internal/directives/config.go
hiddeco: I didn't want the reconciler to end up with too much raw logic when working with the further directives abstractions, and I want to add a check that ensures an empty map actually returns nil instead of an empty JSON object.

api/v1alpha1/promotion_types.go
```diff
@@ -138,10 +138,38 @@ type PromotionStatus struct {
...
+	// HealthChecks contains the health check directives to be executed after
+	// the Promotion has completed.
+	HealthChecks []HealthCheckStep `json:"healthChecks,omitempty" protobuf:"bytes,8,rep,name=healthChecks"`
 	// FinishedAt is the time when the promotion was completed.
 	FinishedAt *metav1.Time `json:"finishedAt,omitempty" protobuf:"bytes,6,opt,name=finishedAt"`
 }
 
+// HealthCheckStep describes a health check directive which can be executed by
+// a Stage to verify the health of a Promotion result.
+type HealthCheckStep struct {
```
krancour, line 150: I think we could probably make the engine use this type directly instead of `directive.HealthCheckStep`. The same can probably be said for promotion steps.

I can handle that _after_ this PR is merged since I am still actively working on the directives package and want to avoid creating merge conflicts over there.

internal/controller/stages/stages.go
```diff
@@ -676,16 +683,40 @@ func (r *reconciler) syncNormalStage(
...
+			// Stage. This is regardless of the phase of the Stage, as the health of the
+			// Argo CD Applications is always relevant.
+			if status.Health = r.appHealth.EvaluateHealth(
+				ctx,
+				stage,
+			); status.Health != nil {
+				logger.WithValues("health", status.Health.Status).Debug("Stage health assessed")
+			} else {
+				logger.Debug("Stage health deemed not applicable")
+			}
```
krancour, line 719: It look like we're doing the classic health check as long as the promotion status has 0 healthchecks defined...

I don't feel 100% positive, but I am thinking this may not be the ideal behavior.

Someone could be using directives instead of classic promotion mechanisms and as long as they haven't updated any Argo CD Apps (since this is the only promotion step that currently generates any health checks steps) then they would have 0 health check steps in that collection, but falling back on the legacy health check process doesn't seem right in this case, especially since it is tightly coupled to the legacy promotion mechanisms, which are not at play in this case.

What would you think about only doing these legacy health checks if the legacy promo mechs are non-nil?

hiddeco: If people would have legacy promotion mechanisms and new directives defined, that would also cause issues because then it could potentially never check the new health checks.

But I get your point, and will restructure the logic a bit and e.g. make use of a switch to avoid the if/else issue.

cmd/controlplane/controller.go
```diff
@@ -273,11 +275,20 @@ func (o *controllerOptions) setupReconcilers(
 	promotionsReconcilerCfg promotions.ReconcilerConfig,
 	stagesReconcilerCfg stages.ReconcilerConfig,
 ) error {
+	var argoCDClient client.Client
+	if argocdMgr != nil {
+		argoCDClient = argocdMgr.GetClient()
+	}
+
+	directivesEngine := directives.NewSimpleEngine(credentialsDB, kargoMgr.GetClient(), argoCDClient)
+	promoMechanisms := promotion.NewMechanisms(kargoMgr.GetClient(), argoCDClient, credentialsDB)
```
krancour, line 284: I actually love seeing the two of these side-by-side here and knowing half of it goes away soon. 😄

krancour (approved):

PR #2583, opened 2024-09-26
refactor: swap step field for uses field

krancour:
@hiddeco are you ok with this?


hiddeco:
What I do not like about this is that it creates ambiguity with Kubernetes resource kinds.

krancour:
We discussed offline and I changed this to `Uses`.

hiddeco (approved):

PR #2604, opened 2024-09-29
feat: keep track of promotion state in promotion status

krancour:
@hiddeco it is best for you to review this one.

I know we had several offline conversations about whether this was worth avoiding or something we'd be forced to do.

Waiting for a PR to merge or close is the thing I ended up really _needing_ this for.

Going back and looking at legacy code, even then, we never really had a way of doing this without storing the PR number in promotion status' metadata field.

fwiw, this is the last thing standing in the way of _all_ of these directive-based examples passing:

https://github.com/krancour/kargo-examples/tree/for-v0.9.0

If you are generally unopposed to this, but wish to change some implementation details, please feel free to amend this PR to maximize the possibility that we cut a release candidate early in the day tomorrow.

__Edit:__ A tangential benefit of this is it provides a little more visibility into what step is holding up a long-running promotion.


api/v1alpha1/promotion_types.go
```diff
@@ -143,6 +143,26 @@ type PromotionStatus struct {
...
+	// CurrentStep is the index of the current promotion step being executed. This
+	// permits steps that have already run successfully to be skipped on
+	// subsequent reconciliations attempts.
+	CurrentStep int64 `json:"currentStep,omitempty" protobuf:"varint,9,opt,name=currentStep"`
+	// State stores the state of the promotion process between reconciliation
+	// attempts.
+	State *apiextensionsv1.JSON `json:"state,omitempty" protobuf:"bytes,10,opt,name=state"`
+}
+
+// GetConfig returns the State field as unmarshalled YAML.
```
hiddeco, line 155:
```suggestion
// GetState returns the State field as unmarshalled YAML.
```

hiddeco (approved):
No real objection from me. I wonder if in the future we continue to want to persist all state to the object (and not e.g. a sub-set which _has_ to survive), but I think this concern can be addressed later.

PR #2617, opened 2024-10-01
feat(controller)!: queue Stages for Application changes using annotation

### LINKED ISSUE #2593: bug: comparison error in health check
I'm seeing many directive-based promotion workflows working e2e now, but I am frequently observing Stages who reported as unhealthy due to an underlying Application having a "comparison error."

The funny thing is that simply refreshing the Stage will re-evaluate health and unstick things.

Since it's the same health check logic running when the comparison error is found and again when it isn't, my assumption is that there really _is_ a comparison error that is cleared in the time between the two checks...

But when I used the legacy promotion mechanisms, I don't encounter this.

I haven't dug into the relevant bits of code yet, but this sounds very much to me as if we were, possibly, promptly re-queueing Stages on failed health if the Stage uses legacy promotion mechanisms and perhaps are _not_ doing that if using promotion steps instead.

@hiddeco is there any chance you'd be willing to dig into this one?

I don't think this needs to block us from cutting a v0.9.0-rc.1, as it's easy enough to work around this issue in the near-term by refreshing a Stage.

hiddeco:
Fixes #2593 

This addresses the above issue by using the `kargo.akuity.io/authorized-stage` annotation to determine which Stage should be queued for reconciliation when a state transition of the Application occurs.

The support for glob expressions in the annotation had to be removed to facilitate this as we require an absolute reference. However, given what people were trying to do with this (see #2540), not allowing this is presumably better to protect people from making this mistake.

I tested this against all the examples in https://github.com/krancour/kargo-examples/tree/for-v0.9.0 (in bulk), and they all eventually reached a healthy state without requiring manual interference.


api/rbac/v1alpha1/annotations.go
```diff
@@ -1,20 +1,22 @@
...
-	// ServiceAccount to associate it with any user authenticated via OIDC and having
-	// the claim indicated by the full annotation key with any of the values indicated by
-	// the annotation. The value of the annotation may be either a scalar string value or a
-	// comma-separated list.
+	// AnnotationKeyOIDCClaimNamePrefix is the prefix of an annotation key that
+	// can be set on a ServiceAccount to associate it with any user authenticated
+	// via OIDC and having the claim indicated by the full annotation key with
+	// any of the values indicated by the annotation. The value of the annotation
+	// may be either a scalar string value or a comma-separated list.
 	AnnotationKeyOIDCClaimNamePrefix = "rbac.kargo.akuity.io/claim."
```
krancour, line 16: Thank you for fixing this.

Dickyvio44 (approved):

Dickyvio44 (approved):

PR #2623, opened 2024-10-01
feat(ui): promotion directives wizard on stage creation step

Marvin9:
https://github.com/user-attachments/assets/ae40fbab-3a54-4c7b-98ec-91582aacf4f7

- This PR simply utilises the JSON schema of directives configuration defined in backend.
- `rjsf` is used to generate dynamic forms from JSON schemas - https://github.com/rjsf-team/react-jsonschema-form/tree/main/packages/antd


krancour:
There may still be time to get this into v0.9.0 if the merge conflicts are resolved.

internal/directives/schemas/git-commit-config.json
```diff
@@ -36,7 +36,6 @@
     "messageFromSteps": {
       "type": "array",
       "description": "TODO",
-      "minItems": 1,
```
Marvin9, line 39: @hiddeco FYI I removed this as this blocked UI form with 0 items of this attribute.

krancour: JSON schema is weird sometimes. Min items only applies if the field is _not_ null and _not_ undefined. So the rule basically says, "If you define this field at all, it should be non-empty." idk how realistic this is, but can you send `null` instead of `[]` of the array is empty?

Marvin9: Agreed and thats the reason I removed [some properties](https://github.com/akuity/kargo/blob/ca6f53cb60f9448423b47bd1275f6ae789b8bb7f/ui/hack/generate-directive-schema.mjs#L66) that conflicts with what the UI library (rsjf) interprets for the schema. I will put back this attribute as it is and skip validation itself for now and come back later when I have generic solution.

Marvin9:
@krancour this one is ready. Should move the milestone to 1.0?

krancour:
@Marvin9 I think we can try and get this in v1.0.0.

Marvin9:
@krancour if you could take a UI look and @rpelczar can look at the code, we are good to merge this

ui/src/features/stage/promotion-steps-wizard/runner-form.module.less
```diff
@@ -0,0 +1,12 @@
...
+    @apply bg-gray-50;
+    @apply p-5;
+    @apply rounded-md;
+    fieldset {
+        // border: 0.5px solid lightgray !important;
+        // box-shadow: 1px 0px 15px -5px rgba(0, 0, 0, 0.1);
+        // padding: 20px 30px !important;
+        // border-radius: 8px;
+        // margin-top: 8px;
+    }
```
rpelczar, line 11:
```suggestion
```

ui/src/features/stage/promotion-steps-wizard/promotion-steps-wizard.tsx
```diff
@@ -0,0 +1,174 @@
...
+                          e.stopPropagation();
+                          setSelectedRunners(
+                            selectedRunners.filter((_, deleteOrder) => deleteOrder !== order)
+                          );
+                        }}
+                      />
+                    </div>
+                  </div>
+                </Card>
+              );
```
rpelczar, line 130: Can we extract this into a separate component?

ui/src/features/stage/create-stage.tsx
```diff
@@ -148,12 +160,32 @@ export const CreateStage = ({
...
+            const yaml = getValues('value');
+
+            try {
+              const stageSpec: Stage = parse(yaml);
+
+              promotionWizardStepsState.setYAML(
+                stringify(stageSpec?.spec?.promotionTemplate?.spec?.steps)
+              );
+            } catch (e) {
+              // explicitly ignore
```
rpelczar, line 177: Don't we want to inform the user that yaml is broken?

ui/src/features/stage/create-stage.tsx
```diff
@@ -109,7 +115,11 @@ export const CreateStage = ({
   const onSubmit = handleSubmit(async (data) => {
     let value = data.value;
     if (tab === 'wizard') {
-      const unmarshalled = stageFormToYAML(watch(), project || '');
+      const unmarshalled = stageFormToYAML(
+        watch(),
```
rpelczar, line 119: This watch here looks bad I think. The intention of usage for this feature is different IMO.

ui/src/features/stage/create-stage.tsx
```diff
@@ -60,7 +66,7 @@ const stageFormToYAML = (data: z.infer<typeof wizardSchema>, namespace: string)
     spec: {
       requestedFreight: data.requestedFreight,
       ...(data.promotionTemplateSteps && {
-        promotionTemplate: { spec: { steps: yaml.parse(data.promotionTemplateSteps) } }
+        promotionTemplate: { spec: { steps: yaml.parse(promotionTemplateSteps) } }
```
rpelczar, line 69: Won't an error be thrown if the schema is invalid?

ui/src/features/common/layout/error-boundary.tsx
```diff
@@ -16,9 +17,15 @@ export class ErrorBoundary extends React.Component<ErrorProps> {
     return { hasError: true };
   }
 
+  componentDidCatch(error: Error) {
+    // eslint-disable-next-line no-console
+    console.error(error);
```
rpelczar, line 22: Not sure if we want to have console errors in the prod build.

rpelczar (approved):
A few minor comments, but apart from that LGTM!

ui/src/features/stage/promotion-steps-wizard/promotion-steps-wizard.tsx
Marvin9: Done

ui/src/features/stage/create-stage.tsx
Marvin9: Agreed. Now It uses `getValues()`

ui/src/features/stage/create-stage.tsx
Marvin9: Interestingly no. We had total control of schema in wizard before it reaches here. The stringified steps are valid so parsing is going to be valid in all cases. But I can see that the beforehand stringify operation is redundant so We can remove it. I will update the code

ui/src/features/stage/create-stage.tsx
Marvin9: This is tricky. If we block user navigating if YAML is invalid, this becomes too aggressive. If we unblock user navigating and show error if YAML is invalid, we already lost broken YAML when we navigate. If we want to keep the broken YAML, this becomes complicated of syncing 2 source of truths (local state and YAML state).

So to simplify at this point, I think the use-case will either be complete raw YAML or complete form or correct YAML to wizard transformation. Broken YAML is rare case I think

Marvin9:
@krancour not sure what to do with codecov CI failing

krancour:
> not sure what to do with codecov CI failing

Sorry for the late reply. It's informational and not required to pass.

PR #2628, opened 2024-10-02
fix(directives): require release name for `helm-template`

hiddeco:
This solves a potential annoyance people could run into as without a release name specified, Helm defaults it to `release-name`. Which causes something like you see in the screenshot below:

image

### DIFF
--- a/internal/directives/schemas/helm-template-config.json
+++ b/internal/directives/schemas/helm-template-config.json
@@ -17,7 +17,8 @@
     },
     "releaseName": {
       "type": "string",
-      "description": "ReleaseName to use for the rendered manifests."
+      "description": "ReleaseName to use for the rendered manifests.",
+      "minLength": 1
     },
     "useReleaseName": {
       "type": "boolean",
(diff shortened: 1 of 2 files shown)


internal/directives/schemas/helm-template-config.json
```diff
@@ -17,7 +17,8 @@
     },
     "releaseName": {
       "type": "string",
-      "description": "ReleaseName to use for the rendered manifests."
+      "description": "ReleaseName to use for the rendered manifests.",
+      "minLength": 1
```
krancour, line 21: I think `releaseName` should also be included in `required` fields on line 6. iirc, `minLength` only kicks in after a field is defined. So `null` or `undefined` remain valid currently.

hiddeco: Oh, in that case I need to improve some other configurations as well.

krancour: Can we fix this on and then circle back to others in a follow-up?

krancour: Fixed in https://github.com/akuity/kargo/pull/2628/commits/d4e765c6e59e5bb9a7bfee6e68ef4967cae99c1a

krancour (approved):

PR #2639, opened 2024-10-02
chore(ui): improve stability

Marvin9:
This PR aims to introduce optimistic updates in `non-watch` APIs so that users don't need to refresh the page at any point of time in the UI.

- For example, create project or create first stage or warehouse, create analysis template

This also addresses bugs in `watch` APIs that didn't work in some cases, which caused users to refresh the page, say after "Promote" action


ui/src/features/project/pipelines/create-warehouse-modal.tsx
```diff
@@ -24,7 +26,16 @@ const formSchema = z.object({
...
   const { mutateAsync, isPending } = useMutation(createResource, {
-    onSuccess: () => hide()
+    onSuccess: (response) => {
+      for (const result of response?.results || []) {
+        if (result?.result?.case === 'createdResourceManifest') {
+          queryCache.warehouse.add(project, [
+            decodeUint8ArrayYamlManifestToJson(result?.result?.value)
+          ]);
+        }
+      }
```
rpelczar, line 36: Why is this needed if we have a watcher for Warehouses?

ui/src/utils/cache.ts
```diff
@@ -0,0 +1,116 @@
+// cache invalidation source-of-truth
```
rpelczar, line 1: I like the concept of cache management utils, but not sure if one big object is a good idea.

WDYT about utils hook inside a feature folder?

ui/src/features/stage/create-stage.tsx
```diff
@@ -84,7 +86,16 @@ export const CreateStage = ({
...
   const { mutateAsync, isPending } = useMutation(createResource, {
-    onSuccess: () => close()
+    onSuccess: (response) => {
+      for (const result of response?.results || []) {
+        if (result?.result?.case === 'createdResourceManifest') {
+          queryCache.stage.add(project || '', [
+            decodeUint8ArrayYamlManifestToJson(result?.result?.value)
+          ]);
+        }
+      }
```
rpelczar, line 96: Same here? We have a watcher for Stages.

ui/src/app.tsx
```diff
@@ -54,6 +55,7 @@ export const App = () => (
           </BrowserRouter>
         </AuthContextProvider>
       </ConfigProvider>
+      <ReactQueryDevtools />
```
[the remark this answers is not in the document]
Marvin9: It is not [included in bundles](https://tanstack.com/query/latest/docs/framework/react/devtools#install-and-import-the-devtools), when NODE_ENV is not 'development'. So yes, its safe to use as it is

ui/src/features/project/pipelines/utils/watcher.ts
```diff
@@ -53,15 +55,21 @@ export class Watcher {
...
     ProcessEvents(
       stream,
-      stages,
+      () => {
+        const data = this.client.getQueryData(
+          createConnectQueryKey(listStages, { project: this.project })
+        );
+
+        return (data as ListStagesResponse)?.stages || [];
+      },
```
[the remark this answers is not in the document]
Marvin9: This is just getting fresh data from state. Previously it used to register watch calls each time there is new data in state and thats how it made sure of accurate data. 

Now with this PR, there are no excessive watch calls register-deregister, so this function needs some mechanism to have fresh data, and this is one of the way to achieve.

ui/src/features/project/pipelines/pipelines.tsx
```diff
@@ -173,17 +173,20 @@ export const Pipelines = ({
...
 
-    return () => watcher.cancelWatch();
-  }, [isLoading, isVisible, name]);
+    watcher.watchStages();
+    watcher.watchWarehouses(refetchFreightData);
+
+    return () => {
+      watcher.cancelWatch();
+    };
+  }, [name, client, isVisible]);
```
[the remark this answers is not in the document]
Marvin9: It does now. It wasn't really watching in particular case (when number of stages span from 0 -> 1, try to perform promotion on first stage and watch won't work) previously. It only render once as long as you are on individual project page

ui/src/features/project/pipelines/create-warehouse-modal.tsx
Marvin9: I think you are right. I resolved a bug with watch after I added this invalidation, now we don't need this anymore.

ui/src/utils/cache.ts
Marvin9: Definitely I can add

ui/src/features/stage/create-stage.tsx
Marvin9: Will update. Same reason as [this](https://github.com/akuity/kargo/pull/2639#discussion_r1786407864)

ui/src/features/project/pipelines/pipelines.tsx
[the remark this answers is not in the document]
Marvin9: Oh and incase if you interpreted `name` or `isVisible` as constantly changing states, it is not. `name` changes when we move to different page and `isVisible` depends on browser tab is active or not

ui/src/features/project/pipelines/utils/watcher.ts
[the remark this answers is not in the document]
rpelczar: You're right, but still, I think we can get the old data from the cache in the updater function. Check this: https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientsetquerydata

ui/src/utils/cache.ts
rpelczar: It's better, but we don't need those hooks in the utils (common) folder.

`analytics-template` we can move to `/features/project/analysis-templates/utils/query-data-cache.ts` or something like this. WDYT?

rpelczar (approved):

ui/src/features/project/pipelines/utils/watcher.ts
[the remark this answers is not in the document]
Marvin9: Makes sense. I will keep this point in mind on 2nd refactor phase at some point. Your suggestion will simplify code

PR #2643, opened 2024-10-03
refactor(directives): `git-overwrite` -> `git-clear`

hiddeco:
After dog feeding, it became apparent that the more natural way of this directive would be to clear a Git branch from contents before running e.g. `kustomize-build`.

By doing this, the previous ("rendered branch") flow of:

1. Clone and checkout `source` and `target`
2. Kustomize build to `output/`
3. "Overwrite" `output/` to `target/`

Changes to:

1. Clone and checkout `source` and `target`
2. Clear contents from `target`
3. Kustomize build to `target`

Effectively causing there to be one directory less to deal with.


Marvin9:
@hiddeco can you update [the name here](https://github.com/akuity/kargo/blob/main/ui/src/features/promotion-directives/registry/use-discover-registries.ts#L60) as well? On a side-note, do you have a better way to bundle built-in registries?

krancour (approved):
🔥 Agree this is a huge improvement. I will update my docs-in-progress accordingly.

hiddeco:
> On a side-note, do you have a better way to bundle built-in registries for accurate UI discovery mechanism?

Not completely sure I follow your question, but is your goal to e.g. gain access to the JSON schemas — or do you want to be able to discover any arbitrary directive that is available?

Marvin9:
> > On a side-note, do you have a better way to bundle built-in registries for accurate UI discovery mechanism?
> 
> Not completely sure I follow your question, but is your goal to e.g. gain access to the JSON schemas — or do you want to be able to discover any arbitrary directive that is available?

any directive that is available. Basically some mechanism, that prevents the necessity to update UI code for any modification in directive registry. I think we can tackle this later on when we make directives pluggable, for now hardcoding it in UI is fine

krancour:
> any directive that is available. Basically some mechanism, that prevents the necessity to update UI code for any modification in directive registry. I think we can tackle this later on when we make directives pluggable, for now hardcoding it in UI is fine

I put some thought into this the other night. I think it very likely that plugins will be registered using a plugin registration CRD. We'll make sure there is an API endpoint that can list those so the UI can learn everything it needs to about available directives.

For now, hard coded is fine.

PR #2648, opened 2024-10-04
refactor(controller): fix slice init length

cuishuang:
The intention here should be to initialize a slice with a capacity of length rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW

### DIFF
--- a/internal/argocd/health.go
+++ b/internal/argocd/health.go
@@ -171,14 +171,14 @@ func (h *applicationHealth) GetApplicationHealth(
 	// e.g. the health status result to become unreliable.
 	if errConditions := filterAppConditions(app, healthErrorConditions...); len(errConditions) > 0 {
 		issues := make([]error, len(errConditions))
-		for _, condition := range errConditions {
-			issues = append(issues, fmt.Errorf(
+		for i, condition := range errConditions {
+			issues[i] = fmt.Errorf(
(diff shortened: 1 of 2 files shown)


internal/argocd/health.go
```diff
@@ -170,7 +170,7 @@ func (h *applicationHealth) GetApplicationHealth(
 	// considered unhealthy as they may indicate a problem which can result in
 	// e.g. the health status result to become unreliable.
 	if errConditions := filterAppConditions(app, healthErrorConditions...); len(errConditions) > 0 {
-		issues := make([]error, len(errConditions))
+		issues := make([]error, 0, len(errConditions))
```
krancour, line 173: Nice catch. But it's not _this_ line that's wrong. It's what comes after it. This is what it should have been:

```golang
		issues := make([]error, len(errConditions))
		for i, condition := range errConditions {
			issues[i] = fmt.Errorf(
				"Argo CD Application %q in namespace %q has %q condition: %s",
				appKey.Name,
				appKey.Namespace,
				condition.Type,
				condition.Message,
			)
		}
```

krancour: Similar for the other spot.

cuishuang: Modified.

Please review again.

krancour:
Thank you @cuishuang!

krancour (approved):

PR #2649, opened 2024-10-04
feat(directives): allow ignore rules for `copy`

hiddeco:
This adds the option to configure `ignore` rules for the `copy` directive, i.e.:

```yaml
steps:
- uses: copy
  config:
    inPath: ./src
    outPath: ./dst
    ignore: |
      # Ignore all *.log files
      *.log
      # Except for the ones in ./src/foo
      !foo/*.log
```

By default, the `.git/` directory is ignored. However, this can be overwritten using:

```yaml
steps:
- uses: copy
  config:
    ignore: |
      # Overwrite default exclude with explicit include
      !.git/
```


krancour:
> By default, the .git/ directory is ignored.

`.git` is a directory, of course, for a typical clone, but for a working tree of a bare repo, `.git` is a file.

By my reading of the code, the pattern you are using is `.git` (and not `.git/`), so I _think_ our bases are covered, but I just wanted to double-check with you.

krancour:
I see you amended the tests to assert that .git dir and .git file are both covered by default.

krancour (approved):

PR #2673, opened 2024-10-06
feat: Adding ClusterAnalysisTemplate support for Stage verifications

### LINKED ISSUE #2481: Stages should support referencing of ClusterAnalysisTemplates
# Proposed Feature

Today we support project (namespace) scoped AnalysisTemplates, but Argo Rollouts also has ClusterAnalysisTemplates. Currently, Kargo Stages can't reference them.

It's been requested that Stages have the ability to reference a globally defined ClusterAnalysisTemplate so that a change to the ClusterAnalysisTemplate can affect all Stages immediately.

# Motivation

A Kargo operator wishes to centrally manage some verification checks that apply to all projects/stages and are not application specific.

# Suggested Implementation

Introduce `clusterScope: true` to the template reference ([matches argo-rollouts syntax](https://argo-rollouts.readthedocs.io/en/stable/features/analysis/#clusteranalysistemplates))

```yaml
  verification:
    analysisTemplates:
    - name: cat-fact
      clusterScope: true 
```

NOTE: there would need to be a new UI component for this as well, since ClusterAnalysisTemplate would be managed outside of a single project.

BenHesketh21:
closes #2481 

# ClusterAnalysisTemplate Support
This PR adds the ability to refer to cluster level analysis templates using the `clusterScope` option (inspired by Argo Rollouts).
The verification process finds all the templates and combines them in one run, in the same way as before but now will also find referenced ClusterAnalysisTemplates if `clusterScope` is set to true.
The UI has also been updated to show both scopes of analysis templates.

[Image from Gyazo](https://gyazo.com/2b04675180ecc4d0d588d9f273c176ac)


krancour:
@BenHesketh21 thank you for your patience with this. I've finally started reviewing this in earnest. What I've looked at so far looks solid, but this is big and there's a lot more to still look at.

Due to the size and complexity, I'm going to ask @hiddeco and @Marvin9 for help with this. Group effort.

@hiddeco since this touches the Stages reconciler, which you are actively refactoring, controller changes here probably need to be on your radar.

@Marvin9 if you don't mind looking at the UI portions of this, it would be a huge help.

I am happy to review the remainder -- e.g. API server changes and such.

We will _try_ to get this into v1.1.0, but it cannot be guaranteed.

Marvin9 (dismissed):
Sorry for late review, Overall UI LGTM! 

Only thing that is blocking is that since Cluster scoped templates are not really project scoped they should not be in individual projects page. I would suggest to have same clickops experience as in project analysis templates but rather on home page (ie. project list page).

Let me know if you have any question

BenHesketh21:
@Marvin9 I've gone with the approach of adding an Analysis button to the left hand menu for now, let me know if you would rather it be done a different way. 
ee926285ac6ed5088b06d6188bb82a1a

krancour:
> Would we want to add a "Analysis Templates" section to the left side menu? or introduce a similar tab system to the project list page?

I would suggest, instead, a "Settings" item on the sidebar (with a gear as the icon) and have the UI for this be under a "Verifications" tab on that settings page. We currently lack any sort of UI for these other "cluster-scoped" things:

* Credentials (Really, these are Secrets in designated namespaces since there's no such thing as a cluster-scoped Secret.)
* ServiceAccounts. (Again, these are actually in designated namespaces.)
* ClusterRoles + ClusterRoleBindings for the aforementioned SAs

As we add those, a tabbed settings page would prevent the left sidebar from getting too noisy. And I think this is a fairly conventional approach anyway.

Marvin9:
I agree with @krancour here. I initially thought put in sidebar and new page but couldn't find a reason to justify its place there (its the only cluster wide settings at the moment). I am fine under "settings" with "verification" tab. I would use [Menu](https://ant.design/components/menu) component from Antd to have vertical tab.

BenHesketh21:
I'll get on that, thank you for the pointers

BenHesketh21:
@Marvin9 One thing I'm not so sure on if you could clarify. I'm looking at a vertical tab using the Menu from Antd. I'm struggling to picture how it would look and where it would go? Here's my thinking so far.

1. It could be like a second sidebar on the left
2. We could move it to the right
3. The verification and others could be children items of the settings button on the left sidebar

Let me know your thoughts, thanks.

BenHesketh21:
I've done the main parts of this, just waiting to see what the tabs should look like if they need to be vertical. I've put it in a horizontal one for now as it was the default. But the settings page with different tabs should be working.
585db1c6f8c5777213543506aad39010

Marvin9:
Sure thing @BenHesketh21 here is how I imagine it would look like considering more settings incoming

image

Let me know if its feasible to you otherwise I would refactor the layout later this PR and for now we can keep the tab as in your screenshot. I will put my comments after we finalise what are we scoping for this PR. Thanks!

krancour:
We're trying to wrap up v1.2 work by 12/20 with plans to release _after_ the holidays.

It might be too ambitious for us to get this merged before then. I'm tentatively moving this to the v1.3 milestone, but if it somehow becomes possible to sneak this into v1.2, I won't hesitate.

Marvin9:
Sorry @BenHesketh21 do you mind if I change the design a bit? just adjusting color, spacing and layout.. Other than that I don't think there is anything that should block UI

BenHesketh21:
@Marvin9 Please be my guest. UI isn't my strongest area.

Marvin9:
Done with the UI and it LGTM! Thanks @BenHesketh21

@krancour it is now ready to review the rest of the things..

api/v1alpha1/stage_types.go
```diff
@@ -621,6 +621,11 @@ type AnalysisTemplateReference struct {
 	//
 	// +kubebuilder:validation:Required
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+	// ClusterScope determines whether the template is an
+	// AnalysisTemplate or a ClusterAnalysisTemplate resource
+	//
+	// +kubebuilder:validation:Optional
+	ClusterScope bool `json:"clusterScope" protobuf:"varint,2,opt,name=clusterScope"`
```
hiddeco, line 628: I would prefer to have this be a `kind` selector, which matches the way we reference `ClusterPromotionTask` resources at present:

https://github.com/akuity/kargo/blob/ef223531d8c4f052aa5ca3aff264684fc97ebd4a/api/v1alpha1/promotion_types.go#L124-L140

hiddeco (commented):
Implementation-wise, this all looks pretty straight-forward and is a job well done.

I just have a minor suggestion in how a `ClusterAnalysisTemplate` is referenced, please see my comment.

api/v1alpha1/stage_types.go
BenHesketh21: I'll get on that, I just took the method used by Argo Rollouts when referencing the different template kinds but I see you way make more sense in this project.

BenHesketh21: @hiddeco I have changed the clusterscope to use a kind reference instead.

hiddeco:
Tested this end-to-end, and everything is working as expected. Great work @BenHesketh21, and you should be done here.

There appears to be a small issue with the UI however, @Marvin9:

image

hiddeco (approved):

PR #2694, opened 2024-10-08
chore: remove legacy promo mechs

krancour:
This is pretty-straightforward -- simply a removal of everything we deprecated in v0.9.0.

All tests are passing and everything verifiably works e2e.

@Marvin9 I notice there are still some references to promotionMechanisms in the UI code, but they mostly seem to be of the form `promotionMechanisms?.` Testing things e2e, I see no adverse effects. Everything in the UI seems to work perfectly still -- but I'm still calling your attention to this because you may still want to take a look and/or clean up those references. Please feel free to amend the PR with related UI changes.

PR #2697, opened 2024-10-08
feat: harden image using apko and wolfi

### LINKED ISSUE #2400: harden image

krancour:
Fixes #2400

This PR applies a hardening strategy previously put to the test on Kargo Render in https://github.com/akuity/kargo-render/pull/297

When building with `make hack-build` or via the official Actions-based CI/release processes, this will build a distroless base image using apko + wolfi and then load that into Docker and than layer the rest of Kargo on top of that. I've verified it works identically to the old image.

I've also verified that changes to the Dockerfile do not disrupt our Tilt-based development workflow.

__Edit:__ I've just realized this will break #2413. Draft until fixed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ VERSION_PACKAGE := github.com/akuity/kargo/internal/version
 CONTAINER_RUNTIME ?= docker
 
 IMAGE_REPO 			?= kargo
-LOCAL_REG_PORT  ?= 5000
+LOCAL_REG_PORT			?= 5000
 BASE_IMAGE 			?= localhost:$(LOCAL_REG_PORT)/$(IMAGE_REPO)-base
 IMAGE_TAG 			?= dev
 IMAGE_PUSH 			?= false
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ VERSION_PACKAGE := github.com/akuity/kargo/internal/version
 CONTAINER_RUNTIME ?= docker
 
 IMAGE_REPO 			?= kargo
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -15,6 +15,8 @@ VERSION_PACKAGE := github.com/akuity/kargo/internal/version
 CONTAINER_RUNTIME ?= docker
 
 IMAGE_REPO 			?= kargo
+LOCAL_REG_PORT  ?= 5000
```
hiddeco, line 18:
```suggestion
LOCAL_REG_PORT			?= 5000
```

charts/kargo/templates/api/deployment.yaml
```diff
@@ -118,7 +118,7 @@ spec:
       {{- if or .Values.api.cabundle.configMapName .Values.api.cabundle.secretName }}
       initContainers:
       - name: parse-cabundle
-        image: {{ include "kargo.image" . }}
+        image: alpine:latest
```
hiddeco, line 121: Probably better to depend on `alpine/curl` latest, which comes with `update-ca-certificates` included. By doing this, we do not have to update and download the package on every Pod start.

krancour: Resolved in https://github.com/akuity/kargo/pull/2697/commits/2ebcab9c50089ba51871c9e9fb532d17f89c24aa

krancour:
This can be merged after #2694.

PR #2702, opened 2024-10-09
chore: use additional scopes while having dex enabled

kpanter:
After upgrade to 0.9.0 SSO groups scope stop to work while having dex enabled (api.oidc.dex.enabled). Env OIDC_ADDITIONAL_SCOPES is not added to api configmap.

### DIFF
--- a/charts/kargo/templates/api/configmap.yaml
+++ b/charts/kargo/templates/api/configmap.yaml
@@ -34,6 +34,7 @@ data:
   {{- end }}
   {{- if .Values.api.oidc.enabled }}
   OIDC_ENABLED: "true"
+  OIDC_ADDITIONAL_SCOPES: {{ join "," .Values.api.oidc.additionalScopes }}
   {{- if .Values.api.oidc.globalServiceAccounts.namespaces }}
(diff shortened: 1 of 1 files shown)


charts/kargo/templates/api/configmap.yaml
```diff
@@ -56,8 +56,8 @@ data:
   {{- if .Values.api.oidc.cliClientID }}
   OIDC_CLI_CLIENT_ID: {{ quote .Values.api.oidc.cliClientID }}
   {{- end }}
-  OIDC_ADDITIONAL_SCOPES: {{ join "," .Values.api.oidc.additionalScopes }}
   {{- end }}
+  OIDC_ADDITIONAL_SCOPES: {{ join "," .Values.api.oidc.additionalScopes }}
```
krancour, line 60: @kpanter thank you! I agree with the change, but would suggest moving this to between existing lines 36 and 37 where I think it will be more obvious that this setting being applied isn't conditional on dex being enabled or not. i.e. It will immediately follow "oidc true". If you don't mind changing that, this will LGTM.

kpanter: @krancour moved OIDC_ADDITIONAL_SCOPES as you suggested.

krancour (approved):

PR #2710, opened 2024-10-10
chore: don't suppress echoing commands of hack-build target

krancour:
In #2697 I had not meant to suppress the commands in the `hack-build` target from being echoed. Seeing the commands is helpful for troubleshooting purposes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,7 +292,7 @@ hack-codegen: hack-build-dev-tools
 # 	make hack-build IMAGE_PLATFORMS=linux/amd64 DOCKER_BUILD_OPTS=--no-cache
 .PHONY: hack-build
 hack-build: build-base-image
-	@{ \
+	{ \
 		$(CONTAINER_RUNTIME) run -d -p 5000:$(LOCAL_REG_PORT) --name tmp-registry registry:2; \
 		trap '$(CONTAINER_RUNTIME) rm -f tmp-registry' EXIT; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -292,14 +292,15 @@ hack-codegen: hack-build-dev-tools
...
-	@{ \
+	{ \
 		$(CONTAINER_RUNTIME) run -d -p 5000:$(LOCAL_REG_PORT) --name tmp-registry registry:2; \
 		trap '$(CONTAINER_RUNTIME) rm -f tmp-registry' EXIT; \
 		docker push $(BASE_IMAGE):latest-amd64; \
 		docker push $(BASE_IMAGE):latest-arm64; \
 		$(CONTAINER_RUNTIME) buildx build \
 			$(DOCKER_BUILD_OPTS) \
 			--network host \
+			--platform linux/amd64 \
```
hiddeco, line 303: This does not appear to match the example usage comment from above.

krancour: This is accidental. Thank you for catching. Coincidentally, I'd put that there temporarily because setting IMAGE_PLATFORMS doesn't seem to have the intended effect. I had intended to troubleshoot that later and open a separate PR.

krancour, line 303:
```suggestion
```

hiddeco: I think the origin of the issue is that we stopped taking `IMAGE_PLATFORMS` into account at some point.

hiddeco (approved):

Issue #2492, opened 2024-09-05
Support git ssh keys as a more secure login mechanism for updating git repos


yagnasrinath:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.

# Proposed Feature
Support git login with ssh keys

# Motivation

http access tokens/password are not allowed to be used in many secure environments. ssh  keys are usually allowed in such secure environments.

krancour:
#2178 is what is blocking this. I expect other changes going into v0.9.0 will incidentally address #2187, and by extension, unblock this.

krancour:
This _might_ already be working in main/release-0.9

I need to test that out before closing.

PR #2711, opened 2024-10-10
fix: relax clone step config validation to allow ssh urls

krancour:
This was too restrictive to permit ssh URLs.

Note that the schemas for `git-open-pr` and `git-wait-for-pr` do not require the same change because you cannot interact with a Git provider's API over ssh, so the http/s URL is _always_ used for those steps.

Part of #2492


hiddeco (approved):

krancour:
I need to fix a test.

PR #2724, opened 2024-10-11
fix: improve freight availability checks

### LINKED ISSUE #2365: bug: need to verify freight being promoted is wanted
Toward the beginning of the promotion process, we do validate that the Freight being promoted is _available_ to the Stage it is being promoted to:

https://github.com/akuity/kargo/blob/fd47846bcddb26e340abc0a1ccb52c44e43d8243/internal/controller/promotions/promotions.go#L471-L478

There is, however, an edge case wherein a piece of Freight could be _available_ to the Stage, but not _wanted._

It would/could happen if a piece of Freight from a particular origin had been manually approved for the Stage in question and since that time the Stage was edited to no longer request Freight from that origin.

Such a promotion should not proceed.

Note: @hiddeco has a recollection of being unable to reproduce this. I will attempt to do so when time permits, but if he's right, we'll just close this.

krancour:
Fixes #2365 

After approaching this problem a few different ways, the thing that came out cleanest was treating the question of whether a piece of Freight is _wanted_ by a given Stage as just another dimension of whether that Freight is _available_ to that Stage.

1b3d577e491eddf6e91b7e463a41ae8c02399419 is some pre-requisite refactoring that made the more prescient refactoring and enhancements in 75d6fcc7d66da56006be5fc371e8c15f965d0de9 a bit easier.

6753bad5a0cb046e891e320508acce410a00356b just updates existing usage of the refactored availability check accordingly.

2559c0c2323e5c544966df8361267e1580e1bc4b adds the availability check to the Promo creation webhook so that availability checks aren't bypassed when using `kubectl` instead of `kargo` or the UI.

### DIFF
--- a/api/v1alpha1/freight_helpers.go
+++ b/api/v1alpha1/freight_helpers.go
@@ -86,34 +86,24 @@ func GetFreightByAlias(
 }
 
 // IsFreightAvailable answers whether the specified Freight is available to the
-// specified Stage having the specified upstream stages. Freight is available
-// if:
-//
-//  1. No upstreamStages are specified
-//     OR
-//  2. The Freight has been verified in ANY of the specified upstream stages
-//     OR
-//  3. The Freight is approved for the specified stage
-//
-// Note: The rationale for returning true when no upstream stages are specified
-// is that some Stages have no upstream Stages (e.g. a Stage that subscribes to
-// a Warehouse), so ANY Freight is available to such a Stage.
-func IsFreightAvailable(
-	freight *Freight,
-	stage string,
-	upstreamStages []string,
-) bool {
-	if len(upstreamStages) == 0 {
+// specified Stage.
+func IsFreightAvailable(stage *Stage, freight *Freight) bool {
+	if stage == nil || freight == nil || stage.Namespace != freight.Namespace {
+		return false
+	}
+	if _, approved := freight.Status.ApprovedFor[stage.Name]; approved {
 		return true
 	}
-	for _, stage := range upstreamStages {
-		if _, ok := freight.Status.VerifiedIn[stage]; ok {
-			return true
-		}
-	}
-	if stage != "" {
-		if _, ok := freight.Status.ApprovedFor[stage]; ok {
-			return true
+	for _, req := range stage.Spec.RequestedFreight {
+		if freight.Origin.Equals(&req.Origin) {
+			if req.Sources.Direct {
+				return true
+			}
+			for _, source := range req.Sources.Stages {
+				if _, verified := freight.Status.VerifiedIn[source]; verified {
+					return true
+				}
+			}
 		}
 	}
 	return false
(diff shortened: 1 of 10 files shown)


internal/api/promote_downstream_v1alpha1.go
```diff
@@ -92,22 +92,26 @@ func (s *server) PromoteDownstream(
 		}
 		return nil, connect.NewError(connect.CodeNotFound, err)
 	}
-	if !s.isFreightAvailableFn(
-		freight,
-		"",                  // approved for not considered
-		[]string{stageName}, // verified in
-	) {
```
krancour, line 99: Awkward usage of `IsFreightAvailable()`, which was written awkwardly to accommodate this.

internal/api/promote_downstream_v1alpha1.go
```diff
@@ -92,22 +92,26 @@ func (s *server) PromoteDownstream(
...
 		return nil, connect.NewError(
 			connect.CodeInvalidArgument,
 			fmt.Errorf(
-				"Freight %q is not available to Stage %q",
+				"Freight %q is not available to Stages down stream from %q",
 				freightName,
 				stageName,
 			),
 		)
 	}
```
krancour, line 112: More direct way of answering "can this Freight move downstream from here?"

internal/api/promote_downstream_v1alpha1.go
```diff
@@ -160,16 +164,23 @@ func (s *server) PromoteDownstream(
...
 	var allStages kargoapi.StageList
 	if err := s.client.List(ctx, &allStages, client.InNamespace(stage.Namespace)); err != nil {
 		return nil, connect.NewError(connect.CodeInternal, err)
 	}
 	var downstreams []kargoapi.Stage
 	for _, s := range allStages.Items {
 		for _, req := range s.Spec.RequestedFreight {
+			if !req.Origin.Equals(&origin) {
+				continue
+			}
```
krancour, line 183: Pretty sure it was an undiscovered bug that this check wasn't here already. I believe the potential has existed for Freight to be promoted to all of a Stage's downstreams _including_ those that may be uninterested in the Freight in question.

api/v1alpha1/freight_helpers.go
```diff
@@ -86,34 +86,24 @@ func GetFreightByAlias(
...
-			return true
-		}
+	if _, approved := freight.Status.ApprovedFor[stage.Name]; approved {
+		return true
 	}
-	if stage != "" {
-		if _, ok := freight.Status.ApprovedFor[stage]; ok {
-			return true
+	for _, req := range stage.Spec.RequestedFreight {
+		if freight.Origin.Equals(&req.Origin) {
```
krancour, line 98: This line is the cornerstone of the PR.

internal/webhook/promotion/webhook.go
```diff
@@ -221,15 +221,38 @@ func (w *webhook) ValidateCreate(
...
+			promo.Name,
+			field.ErrorList{
+				field.Invalid(
+					field.NewPath("spec", "freight"),
+					promo.Spec.Freight,
+					"Freight is not available to this Stage",
+				),
+			},
+		)
+	}
```
krancour, line 252: Now we don't miss out on this check when using `kubectl` instead of `kargo` or the UI.

Issue #1538, opened 2024-02-26
Warehouse creation wizard


jessesuen:
Current the UI has no feature for creating a new Warehouse.

@jessesuen, "wizard" was your wording. Could we guess that initially, this isn't more than was done for creating Stages? i.e. Pop open a YAML editor with example YAML and schema-aware context help?

jessesuen:
> @jessesuen, "wizard" was your wording. Could we guess that initially, this isn't more than was done for creating Stages? i.e. Pop open a YAML editor with example YAML and schema-aware context help?

Yes, YAML editor would be an MVP. 

That said, I think Warehouses and Projects (unlike Stages), have less options to configure, and so I do feel it's possible to create a simplified UX around it (even if it means they are unable to create more advanced Warehouses). The YAML editor will always be our escape hatch for more advanced config.

krancour:
Agree that Stages are the main difficulty and others have potential to be bound to a form or other UI controls. YAML editor for MVP sgtm. Thank you.

Marvin9:
I was about to create an issue for warehouse wizard. I think it would be good to have it in v1.0?

PR #2728, opened 2024-10-11
feat(ui): warehouse creation wizard

Marvin9:
fixes https://github.com/akuity/kargo/issues/1538

https://github.com/user-attachments/assets/4d8ccfd9-848c-44a0-9b70-e69d31849ba9


krancour:
I took this for a test drive. I think it's a great start, but it feels a little rough still.

Some issues I encountered...

This felt awkward because the names `subscription-1`, `subscription-2`, etc. look like some sort of identifiers, which I cannot control. I know they're just indices, but it _felt_ weird.

For reference:

Screenshot 2024-10-11 at 7 36 02 PM

I wonder if we could just not have any heading on each subscription?

The next issue was that the form misled me into believing I can create a subscription that use more than one of git, image, or chart. Ex:

Screenshot 2024-10-11 at 7 39 44 PM

I knew that these were mutually exclusive, so I tried it and ended up with the expected error:

Screenshot 2024-10-11 at 7 38 25 PM

If we could have a drop down where you select to create a git, image _or_ chart subscription, and then have that open one specific sub-form or another, that seems it might prevent the confusion.

The last is that I felt like some sort of help text might be in order for most of these fields.

I still do think this is a great start, but unless @jessesuen feels strongly about it this being in v1.0.0, I might like to hold this for v1.1.0 so that this can be refined a bit more.

Marvin9:
Thanks for the suggestion @krancour I will modify it to make it usable this week

Marvin9:
@krancour @rpelczar would you take a look please? I have updated demo video

ui/src/features/utils/url-query-state/use-url-query-state.ts
```diff
@@ -0,0 +1,45 @@
+// generic solution to use URL query as a state
+
+import queryString from 'query-string';
+import { useCallback, useMemo } from 'react';
+import { useNavigate, useSearchParams } from 'react-router-dom';
+
+type SimpleState = Record<string, string>; /* lets keep it simple string value for now */
+
+export const useURLQueryState = <T extends Record<string, unknown>>() => {
```
Marvin9, line 9: @rpelczar I will use our core approach once I get green light there

krancour:
Sorry for the delay. I promise to get my 👀 on this first thing tomorrow.

krancour:
This works _really_ well. My only additional feedback is that it might help if fields for subscriptions were organized more in order of importance (or whether they are required). `repoURL` at the top, for instance, would make the forms feel more intuitive, followed perhaps by selection strategy, constraints, etc.

Longer term, (and definitely non-blocking / something for a follow-up) I think we could/should warn about some of the less efficient options. High discovery limits, low intervals, or selection strategies based on time are all inefficient and are best avoided if possible.

Marvin9:
> `repoURL` at the top, for instance, would make the forms feel more intuitive

That is very good point. Required fields on top, I would do it in this PR.

> Longer term, (and definitely non-blocking / something for a follow-up) I think we could/should warn about some of the less efficient options. High discovery limits, low intervals, or selection strategies based on time are all inefficient and are best avoided if possible.

Yes the summary view point is also considerable. As you said lets not block this PR.

krancour:
Latest changes lgtm, but will still defer to @rpelczar in case he has any outstanding feedback still.

Marvin9:
@rpelczar its ready to review now

rpelczar (changes requested):
In my opinion, the search params mechanism is unnecessarily complicated for such a simple case. This solution is also not consistent to Create Stage drawer.
I'd like to have a new route for the creation wizard, and I don't think we need search params for tabs inside the wizard.

Marvin9:
Discussed with this @rpelczar. I think from the users POV it improves a lot (save your form states on refresh and you can bookmark it as well). But we agreed that for now we will stay consistent with create stage and keep URL clean.

rpelczar (approved):

Marvin9:
@krancour your stamp required

PR #2730, opened 2024-10-11
feat(ui): freight history in stage details

Marvin9:
Screenshot 2024-10-11 at 4 21 45 PM (2)

### DIFF
--- a/ui/src/features/stage/freight-history.tsx
+++ b/ui/src/features/stage/freight-history.tsx
@@ -4,6 +4,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { useQueryClient } from '@tanstack/react-query';
 import { Empty } from 'antd';
 import classNames from 'classnames';
+import { useMemo } from 'react';
 import { generatePath, Link, useNavigate } from 'react-router-dom';
 
 import { paths } from '@ui/config/paths';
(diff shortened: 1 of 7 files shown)


ui/src/features/stage/freight-history.tsx
```diff
@@ -0,0 +1,164 @@
...
+  freightHistory?: StageStatus['freightHistory'];
+  // freight hash name which is active at the moment
+  // you can get this from lastPromotion in stage status
+  // usually last one is active but we have to consider multi-pipeline case
+  currentActiveFreight?: string;
+}) => {
+  const queryClient = useQueryClient();
+  const navigate = useNavigate();
+
+  const { freightHistoryPerWarehouse, freightMap } = useMemo(() => {
```
Marvin9, line 46: @rpelczar I have moved this to useMemo

ui/src/features/stage/freight-history.tsx
```diff
@@ -0,0 +1,164 @@
...
+      for (const [warehouseIdentifier, freightReference] of Object.entries(items)) {
+        if (!freightHistoryPerWarehouse[warehouseIdentifier]) {
+          freightHistoryPerWarehouse[warehouseIdentifier] = [];
+        }
+
+        freightHistoryPerWarehouse[warehouseIdentifier].push(freightReference);
+      }
+    }
+
+    const freightData = queryClient.getQueryData(
```
Marvin9, line 67: > Why not regular fetch? I think you assumed there are data. Maybe it will work always in our case, but not sure if it is a good practice. You can set a stale property for data in query configuration, so it won't generate a new request. Also, you are sure that if the data are not available, it will be fetched. Let me know what you think.

You are right, this will work in our case. I didn't want to make the query for the data we already have and I wanted to avoid adding context provider or prop drilling to have that data. The closest I can do is [ensureQueryData](https://tanstack.com/query/v4/docs/reference/QueryClient/#queryclientensurequerydata) but since its async I have to wrap in effect. I will look whether to make query with invalidation or use ensureQueryData and add state

krancour:
Test drove and this worked quite nicely. 🔥 

LGTM as long as @rpelczar feels all his feedback was adequately addressed.

krancour (approved):

rpelczar (approved):
LGTM

PR #2747, opened 2024-10-14
fix: git username/email fixes

### LINKED ISSUE #2700: bug(git-commit): author name and email are not taken into account
When using the following promotionTemplate, the produced commit is from the `"Kargo"` user. I would expect the commit to be from the `"foo"` user
Changing the value of `author.name` or `author.email` does not seem to change anything.

```yaml
  promotionTemplate:
    spec:
      steps:
        - uses: git-clone
          config:
            repoURL: https://github.com/mygitopsrepo.git
            checkout:
              - branch: main
                path: ./repo
        - uses: helm-update-image
          as: update-image
          config:
            path: ./path/to/values.yaml
            images:
              - image: myimagename
                fromOrigin:
                  kind: Warehouse
                  name: mywarehousename
                key: image.tag
                value: Tag
        - uses: git-commit
          as: commit
          config:
            path: ./repo
            author:
              name: foo
              email: bar@guestbook.com
            messageFromSteps:
              - update-image
        - uses: git-push
          config:
            path: ./repo
            targetBranch: main
```

Besides, I fear that the environment variables provided to the controller via the `.Values.controller.gitClient` chart value are not honored anymore, so signing commits is no longer possible.

krancour:
Fixes #2700 -- which is actually two separate bugs. See in-line comments.


charts/kargo/values.yaml
```diff
@@ -377,9 +377,9 @@ controller:
 
   gitClient:
     ## @param controller.gitClient.name Specifies the name of the Kargo controller (used when authoring Git commits).
-    name: "Kargo Render"
+    name: "Kargo"
     ## @param controller.gitClient.email Specifies the email of the Kargo controller (used when authoring Git commits).
-    email: "kargo-render@akuity.io"
+    email: "no-reply@kargo.io"
```
krancour, line 382: These seemed worth updating.

internal/controller/git/base_repo.go
```diff
@@ -12,6 +12,11 @@ import (
 	libExec "github.com/akuity/kargo/internal/exec"
 )
 
+const (
+	defaultUsername = "Kargo"
+	defaultEmail    = "no-reply@kargo.io"
+)
```
krancour, line 18: Used in multiple places now. Time to make them consts.

internal/controller/git/work_tree.go
```diff
@@ -199,6 +199,17 @@ func (w *workTree) Commit(message string, opts *CommitOptions) error {
...
+		name := opts.Author.Name
+		if name == "" {
+			name = defaultUsername
+		}
+		email := opts.Author.Email
+		if email == "" {
+			email = defaultEmail
+		}
+		cmdTokens = append(cmdTokens, "--author", fmt.Sprintf("%s <%s>", name, email))
+	}
```
krancour, line 212: This logic has been missing all along. Until the `git-commit` step, we didn't actually have any scenario where optional authorship information was passed to this method.

internal/directives/git_cloner.go
```diff
@@ -28,14 +29,34 @@ func init() {
...
+		SigningKeyPath string `envconfig:"GITCLIENT_SIGNING_KEY_PATH"`
+	}{}
+	envconfig.MustProcess("", &cfg)
+	return git.User{
+		Name:           cfg.Name,
+		Email:          cfg.Email,
+		SigningKeyType: git.SigningKeyType(cfg.SigningKeyType),
+		SigningKeyPath: cfg.SigningKeyPath,
+	}
+}
```
krancour, line 51: The retired git-based promo mechs were configured similarly and this was accidentally not included in #2537

hiddeco (approved):

Issue #2662, opened 2024-10-04
Ability to terminate a Running Promotion


jessesuen:
# Proposed Feature

When hitting https://github.com/akuity/kargo/issues/2661, it became apparent that we need the ability to terminate promotions since they might be long-lived, or bugs might cause them to get stuck running.

# Motivation

I no longer want a Promotion to run.

# Suggested Implementation

Introduce a `terminate` field in a Promotion, that would be set by API server by a user interaction:

```
spec:
  terminate: {}
```

Note that terminate is a struct instead of a boolean, because we might want different termination options, e.g. gracefully close PRs that were opened (SIGTERM), or do nothing like a `kill -9` (SIGKILL).

hiddeco:
I am unsure if this should be declared in-spec, or be more like the [current abort annotation](https://github.com/akuity/kargo/blob/8d4e58e21e15f0a65a6dad5b3dac089b2285559f/api/v1alpha1/annotations.go#L23-L27) we have for the verification process of a Stage.

krancour:
Verifications aren't themselves a CR, so an annotation on a Stage to kill a verification feels like it makes sense.

If we did go an annotation-based route here, couldn't we apply that annotation directly to the Promotion instead of on the Stage?

hiddeco:
> If we did go an annotation-based route here, couldn't we apply that annotation directly to the Promotion instead of on the Stage?

Yes, that's actually what I was suggesting.

hiddeco:
This is now possible using `kubectl`, `kargo` and the API thanks to #2479. The bit that's missing is the integration into the UI.

PR #2749, opened 2024-10-15
feat: allow terminating Promotions 

hiddeco:
Backend implementation of #2662. 

This allows non-terminal Promotions (e.g. pending, running) to be terminated per user request using the `kargo.akuity.io/abort` annotation (which we already use to abort verifications for a Stage). 

Using `kubectl`:

```console
$ kubectl annotate promotions <promotion> kargo.akuity.io/abort=terminate
```

Using `kargo`:

```console
$ kargo promote --abort --name=<promotion>
```


internal/cli/cmd/promote/promote.go
```diff
@@ -122,13 +128,18 @@ func (o *promotionOptions) addFlags(cmd *cobra.Command) {
 			option.StageFlag,
 		),
 	)
+	option.Abort(cmd.Flags(), &o.Abort, false, fmt.Sprintf(
+		"Abort a non-terminal promotion. If set, --%s must be set.", option.NameFlag,
+	))
```
krancour, line 133: `verifyStageOptions` could use this also. Would you mind updating that so that we're more consistent on how the abort flag is defined between these two commands?

internal/webhook/stage/webhook.go
```diff
@@ -127,10 +127,10 @@ func (w *webhook) Default(ctx context.Context, obj runtime.Object) error {
...
 
-		if verReq, ok := kargoapi.AbortAnnotationValue(stage.Annotations); ok {
+		if verReq, ok := kargoapi.AbortVerificationAnnotationValue(stage.Annotations); ok {
 			var oldVerReq *kargoapi.VerificationRequest
 			if oldStage != nil {
-				oldVerReq, _ = kargoapi.AbortAnnotationValue(oldStage.Annotations)
+				oldVerReq, _ = kargoapi.AbortVerificationAnnotationValue(oldStage.Annotations)
 			}
 			// If the abort request has changed, enrich the annotation with the
 			// actor and control plane information.
```
krancour, line 136: Do you suppose we ought to do this for the Promotions webhook also?

krancour:
Two small suggestions, but preemptively LGTM'ing.

krancour (approved):

internal/cli/cmd/promote/promote.go
hiddeco: Ah, doh. That was the whole idea.

internal/webhook/stage/webhook.go
hiddeco: Yes, great catch.

internal/cli/cmd/promote/promote.go
hiddeco: Now available in b4f3b660cfd4c57eb8fb94fb2ad3db82b4771470.

internal/webhook/stage/webhook.go
hiddeco: Addressed in cd45c93de0133d93d42d384e87899d264593c72c.

PR #2750, opened 2024-10-15
fix(controller)!: commit message build fix + fail on obvious misconfiguration

### LINKED ISSUE #2699: bug(promotion): helm-update-image directive does not produce commitMessage when there is no update
According to [the documentation](https://kargo.akuity.io/references/promotion-steps#helm-update-image-output), the `helm-update-image` step should produce a `commitMessage`

However, my git-commit action does not find any when looking up that step's outputs. I get the following error message even if the previous steps were successful: 
```bash
failed to run step "git-commit": error building commit message: no commit message found in output from step with alias "update-image"; cannot construct commit message
```
image

Corresponding promotionTemplate (trying to update a single file on main):
```yaml
    steps:
      - uses: git-clone
        config:
          repoURL: https://github.com/org/repo.git
          checkout:
            - branch: main
              path: ./src
      - uses: helm-update-image
        as: update-image
        config:
          path: ./src/path/to/values.yaml
          images:
            - image: repoURL/imageName
              key: image.tag
              value: Tag
      - uses: git-commit
        as: commit
        config:
          path: ./src
          messageFromSteps:
            - update-image
      - uses: git-push
        config:
          path: ./src
          targetBranch: main
```

Not sure if it's my promotionTemplate (and how I try to do the commit) that's wrong as I could not find an example for my use-case or if it's an actual bug

krancour:
Fixes #2699 and then some.

The first commit simply addresses the realistic scenario that a previous step that might have produced a fragment of a commit message may not have.

The user who reported #2699 eventually determined that their issue was caused by misconfiguration. We've often heard reports of users fat-fingering something, resulting in a step in their process effectively being a no-op, and we've often discussed failing hard when this happens. I couldn't find an open issue for this, but felt this was a good opportunity to address it.

So the second commit deals with all the following scenarios that are obvious evidence of misconfiguration:

* Cannot find an artifact that is expected to be in a Promotion's target Freight.
* Key in a YAML file whose value is to be update not found.
* Chart dependency whose version is to be updated not found.
* Argo CD source update that does not match any of the Application's sources.

This is a breaking change only in that it changes promotion process behavior to be stricter, which is for the better.

### DIFF
--- a/internal/directives/argocd_revisions.go
+++ b/internal/directives/argocd_revisions.go
@@ -2,6 +2,7 @@ package directives
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -114,7 +115,7 @@ func (a *argocdUpdater) getDesiredRevisionForSource(
 			chartName,
 		)
 		if err != nil {
-			if _, ok := err.(freight.NotFoundError); ok {
+			if errors.As(err, &freight.NotFoundError{}) {
 				return "", nil
 			}
 			return "",
@@ -133,7 +134,7 @@ func (a *argocdUpdater) getDesiredRevisionForSource(
 			src.RepoURL,
 		)
 		if err != nil {
-			if _, ok := err.(freight.NotFoundError); ok {
+			if errors.As(err, &freight.NotFoundError{}) {
 				return "", nil
 			}
 			return "",


internal/controller/freight/finder.go
```diff
@@ -19,6 +19,7 @@ func FindCommit(
 	desiredOrigin *kargoapi.FreightOrigin,
 	freight []kargoapi.FreightReference,
 	repoURL string,
+	mustFind bool,
```
krancour, line 22: Most of the time, it's a problem if we fail to find the artifact we're looking for. Most callers will set this to true. There are some instances where the caller is more tolerant of something not being found, in which case they can set this to false and deal with the nil artifact themselves.

internal/directives/argocd_revisions.go
```diff
@@ -112,6 +112,7 @@ func (a *argocdUpdater) getDesiredRevisionForSource(
 			stepCtx.Freight.References(),
 			repoURL,
 			chartName,
+			false, // mustFind = false means we'll deal with a nil result ourselves
```
krancour, line 115: This is a case where we don't care if we don't find the artifact. It's completely plausible that a source references a chart repo or git repo that our Warehouse(s) don't subscribe to.

internal/directives/argocd_revisions_test.go
```diff
@@ -60,12 +60,6 @@ func Test_argoCDUpdater_getDesiredRevisions(t *testing.T) {
 							RepoURL: "example.com",
 							Chart:   "fake-chart",
 						},
-						{
-							// This has no matching artifact in the Freight. We should not
-							// know what revision we want.
-							RepoURL: "https://example.com",
-							Chart:   "yet-another-fake-chart",
-						},
```
krancour, line 68: This case is just an error now.

internal/directives/argocd_updater.go
```diff
@@ -87,7 +87,7 @@ type argocdUpdater struct {
 		update *ArgoCDAppSourceUpdate,
 		desiredRevision string,
 		src argocd.ApplicationSource,
-	) (argocd.ApplicationSource, error)
+	) (argocd.ApplicationSource, bool, error)
```
krancour, line 90: This will now report whether the update matched the source.

internal/directives/argocd_updater.go
```diff
@@ -717,27 +736,30 @@ func (a *argocdUpdater) applyArgoCDSourceUpdate(
 	update *ArgoCDAppSourceUpdate,
 	desiredRevision string,
 	source argocd.ApplicationSource,
-) (argocd.ApplicationSource, error) {
+) (argocd.ApplicationSource, bool, error) {
```
krancour, line 739: This will now report whether the update matched the source.

internal/directives/argocd_updater.go
```diff
@@ -342,6 +344,23 @@ func (a *argocdUpdater) buildDesiredSources(
...
+						"for source with repoURL %s",
+					app.Name, app.Namespace, srcUpdate.RepoURL,
+				)
+			}
+			return nil, fmt.Errorf(
+				"no source of Argo CD Application %q in namespace %q matched update "+
+					"for source with repoURL %s and chart %q",
+				app.Name, app.Namespace, srcUpdate.RepoURL, srcUpdate.Chart,
+			)
 		}
```
krancour, line 364: If we looped through all sources, attempting to apply this update, and the update didn't match any of those sources, this is evidence of misconfiguration.

internal/directives/helm_chart_updater.go
```diff
@@ -160,20 +160,26 @@ func (h *helmChartUpdater) processChartUpdates(
...
+				updateUsed = true
 				break
 			}
 		}
+		if !updateUsed {
+			return nil, fmt.Errorf(
+				"no dependency in Chart.yaml matched update with repository %s and name %q",
+				update.Repository, update.Name,
+			)
+		}
```
krancour, line 182: If we looped through all the dependencies and found none that matched the update, this is evidence of misconfiguration.

internal/controller/freight/finder.go
krancour: Fixed in c2b1b014

internal/directives/argocd_revisions.go
```diff
@@ -114,12 +114,12 @@ func (a *argocdUpdater) getDesiredRevisionForSource(
 			chartName,
 		)
 		if err != nil {
+			if _, ok := err.(freight.NotFoundError); ok {
```
hiddeco, line 117:
```suggestion
			if errors.Is(freight.NotFoundError) {
```

(applies to all of them)

hiddeco (approved):

PR #2760, opened 2024-10-16
feat(ui): abort promotion

Marvin9:
fixes https://github.com/akuity/kargo/issues/2662

https://github.com/user-attachments/assets/305e2d20-c2da-44a9-b4d0-6116798b5d67


internal/controller/promotions/promotions.go
```diff
@@ -198,6 +198,7 @@ func newReconciler(
 	}
 	r.getStageFn = kargoapi.GetStage
 	r.promoteFn = r.promote
+	r.terminatePromotionFn = r.terminatePromotion
```
Marvin9, line 201: @hiddeco ^

hiddeco: Thanks 🙇

Marvin9:
@hiddeco @krancour I think "Promotion terminated on <actor> request" message would be more useful

hiddeco:
I don't think that in all cases the "actor" will be meaningful because it could be a controller identity. But happy to change it if @krancour agrees.

Marvin9:
> it could be a controller identity

You mean controller aborts promotion itself in some cases?

krancour:
> You mean controller aborts promotion itself in some cases?

When does it happen when not initiated by a user?

> I don't think that in all cases the "actor" will be meaningful because it could be a controller identity.

imho, if it's meaningful in _some_ cases, I think it's worth it.

Marvin9:
Thanks @krancour @hiddeco this was an idea, and not blocking this PR

ui/src/features/common/promotion-directive-step-status/utils.ts
```diff
@@ -30,7 +33,7 @@ export const getPromotionDirectiveStepStatus = (
 
   if (
     promotionStatus?.phase !== PromotionStatusPhase.SUCCEEDED &&
-    stepNumber > Number(promotionStatus?.currentStep)
+    stepNumber >= Number(promotionStatus?.currentStep)
```
Marvin9, line 36: @hiddeco can you confirm what is index of promotion step after it transition to Aborted state?

hiddeco: It will freeze.

Marvin9: Thanks for confirming. The change in this line is correct!

ui/src/features/stage/promotion-details-modal.tsx
```diff
@@ -188,6 +234,22 @@ export const PromotionDetailsModal = ({
...
+            </Button>
+            <Button
+              danger
+              icon={<FontAwesomeIcon icon={faStopCircle} className='text-lg' />}
+              onClick={confirmAbortRequest}
+            >
+              Abort
+            </Button>
+          </>
+        )
```
rpelczar, line 251: I think Close button should always be visible

ui/src/features/stage/promotion-details-modal.tsx
```diff
@@ -188,6 +234,22 @@ export const PromotionDetailsModal = ({
...
+            <Button onClick={hide} type='primary'>
+              Close
+            </Button>
+            <Button
+              danger
+              icon={<FontAwesomeIcon icon={faStopCircle} className='text-lg' />}
+              onClick={confirmAbortRequest}
+            >
+              Abort
+            </Button>
```
rpelczar, line 249: I wonder if this place is the best. On the right side, next to the cancel button usually we have confirmation/main action button. I like the danger color, but WDYT about moving this button to the left side?

ui/src/features/stage/promotion-details-modal.tsx
```diff
@@ -188,6 +234,22 @@ export const PromotionDetailsModal = ({
       onOk={hide}
       onCancel={hide}
       cancelButtonProps={{ hidden: true }}
+      footer={
+        canAbortPromotion(promotion) && (
+          <>
+            <Button onClick={hide} type='primary'>
```
rpelczar, line 240: Can we change the type to the default as everywhere for the cancel button?

ui/src/features/stage/promotions.tsx
```diff
@@ -121,6 +142,11 @@ export const Promotions = () => {
       render: (_, promotion) => {
         const annotation = promotion.metadata?.annotations['kargo.akuity.io/create-actor'];
         const email = annotation ? annotation.split(':')[1] : 'N/A';
+
+        if (!email) {
+          return annotation;
+        }
+
         return email;
```
rpelczar, line 150:
```suggestion
        return email || annotation;
```

ui/src/features/stage/promotion-details-modal.tsx
Marvin9: Its default everywhere else so I changed to default here

ui/src/features/stage/promotion-details-modal.tsx
Marvin9: Makes sense but everywhere else we have action buttons from right to left. So if we change it here, we have to change at lots of places

ui/src/features/stage/promotion-details-modal.tsx
Marvin9: Done. Thanks

rpelczar (approved):

PR #2761, opened 2024-10-16
feat!: limit controller secret access only to project namespaces

krancour:
Fixes: #2448 

Supersedes #2522 

Don't be scurred... most of this is tests.

This is intended to mitigate sharded controllers (distributed to clusters other than the control plane's) having cluster-wide read on Secrets in the control plane's cluster. This will dynamically expand and contract the permissions of the Kargo controller(s) in a similar fashion to what's been done for the API server since #1486:

* On Project create, add a RoleBinding per Kargo controller SA to grant r/o Secret access within the Project's namespace.
* Nothing special to do on Project delete. Project delete takes the namespace with it, including any such RoleBindings.

* On create/update of SA _in Kargo's own namespace_, ensure appropriate RoleBinding exists for every Project _if the SA is labeled as a controller_.
* On delete of SA _in Kargo's own namespace_, ensure removal of appropriate RoleBinding for every Project.
* On update of SA _in Kargo's own namespace_, ensure removal of appropriate RoleBinding for every Project _if the SA is no-longer labeled as a controller_.

Finalizers are added/removed as appropriate used to ensure necessary cleanup always occurs _and_ that cleanup doesn't occur unnecessarily. 

These changes are verified to work e2e.

Breaking change: If you have designated some namespaces as "global credential stores" using `controller.globalCredentials.namespaces` at install-time, you must now _manually_ ensure controller(s) have access to Secrets in those namespaces. The chart cannot do this because the namespaces may not exist at the time Kargo is installed. This also cannot be done dynamically with the management controller because if values are removed from the list and the management controller is restarted, it will not know what namespaces have been removed from the list and require clean up.

@fykaa laid most of the groundwork for this. Thank you!

@gdsoumya requesting your 👀 on this, if you don't mind.


charts/kargo/README.md
```diff
@@ -145,7 +145,7 @@ the Kargo controller is running.
...
+| `controller.globalCredentials.namespaces`    | List of namespaces to look for shared credentials. Note that as ov v1.0.0, the Kargo controller does not have cluster-wide access to Secrets. The controller receives read-only permission for Secrets on a per-Project basis as Projects are created. If you designate some namespaces as homes for "global" credentials, you will need to manually grant the controller permission to read Secrets in those namespaces.                                                                                                                                                                                                                                                                                                        | `[]`                |
```
krancour, line 148: cc @gdsoumya

cmd/controlplane/controller.go
```diff
@@ -172,29 +170,23 @@ func (o *controllerOptions) setupKargoManager(
...
+			Client: client.Options{
+				Cache: &client.CacheOptions{
+					// The controller does not have cluster-wide permissions, to
+					// get/list/watch Secrets. Its access to Secrets grows and shrinks
+					// dynamically as Projects are created and deleted. We disable caching
+					// here since the underlying informer will not be able to watch
+					// Secrets in all namespaces.
+					DisableFor: []client.Object{&corev1.Secret{}},
+				},
+			},
```
krancour, line 189: cc @hiddeco. Thanks for helping me figure this out.

cmd/controlplane/management_controller.go
```diff
@@ -118,6 +128,15 @@ func (o *managementControllerOptions) setupManager(ctx context.Context) (manager
...
 			},
+			Cache: cache.Options{
+				ByObject: map[client.Object]cache.ByObject{
+					&corev1.ServiceAccount{}: {
+						Namespaces: map[string]cache.Config{
+							os.GetEnv("KARGO_NAMESPACE", "kargo"): {},
+						},
+					},
+				},
+			},
```
krancour, line 139: Another thank you to @hiddeco.

internal/controller/labels.go
```diff
@@ -63,18 +62,3 @@ func GetShardRequirement(shard string) (*labels.Requirement, error) {
...
-	req, err := labels.NewRequirement(kargoapi.CredentialTypeLabelKey, selection.In, []string{
-		credentials.TypeGit.String(),
-		credentials.TypeHelm.String(),
-		credentials.TypeImage.String(),
-	})
-	if err != nil {
-		return nil, fmt.Errorf("error creating credentials label selector: %w", err)
-	}
-	return req, nil
-}
```
krancour, line 80: No longer needed since the manager won't cache Secrets at all anymore.

internal/controller/management/projects/projects.go
```diff
@@ -28,8 +28,12 @@ import (
 	"github.com/akuity/kargo/internal/logging"
 )
 
+const controllerReadSecretsClusterRoleName = "kargo-controller-read-secrets"
+
 type ReconcilerConfig struct {
-	KargoNamespace string `envconfig:"KARGO_NAMESPACE" required:"true"`
+	KargoNamespace                     string `envconfig:"KARGO_NAMESPACE" default:"kargo"`
+	ControllerServiceAccountLabelKey   string `envconfig:"CONTROLLER_SERVICE_ACCOUNT_LABEL_KEY" default:"app.kubernetes.io/component"` // nolint: lll
+	ControllerServiceAccountLabelValue string `envconfig:"CONTROLLER_SERVICE_ACCOUNT_LABEL_VALUE" default:"controller"`
```
krancour, line 36: cc @gdsoumya. Please lmk offline if this will be configurable enough for you.

internal/controller/management/projects/projects.go
```diff
@@ -96,6 +100,8 @@ type reconciler struct {
 
 	ensureAPIAdminPermissionsFn func(context.Context, *kargoapi.Project) error
 
+	ensureControllerPermissionsFn func(context.Context, *kargoapi.Project) error
```
krancour, line 103: @hiddeco, this is only for this existing reconciler so that @fykaa and I didn't have to re-write all the tests. The tests for the new reconciler are done differently.

hiddeco:
This is on my list to review tomorrow before you start your day.

internal/controller/management/serviceaccounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,629 @@
+package serviceaccounts
```
hiddeco, line 1: I am so incredibly happy with how there isn't any callback wizardry here 💯🥇

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,239 @@
...
+	)
+	logger.Debug("ensuring non-existence of necessary RoleBinding in all Project namespaces")
+
+	projectList := &kargoapi.ProjectList{}
+	if err := r.client.List(ctx, projectList); err != nil {
+		return fmt.Errorf("error listing Projects: %w", err)
+	}
+
+	for _, project := range projectList.Items {
+		projectLogger := logger.WithValues("project.namespace", project.Name)
```
hiddeco, line 208: Is the `project.namespace` here intentional, or should it be `project` OR `namespace`?

internal/controller/management/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,239 @@
...
+		// ServiceAccount access to read Secrets in all Project namespaces.
+		if controllerutil.AddFinalizer(sa, kargoapi.FinalizerName) {
+			if err := r.client.Update(ctx, sa); err != nil {
+				return ctrl.Result{}, fmt.Errorf(
+					"error adding finalizer to ServiceAccount %q in namespace %q: %w",
+					sa.Name, sa.Namespace, err,
+				)
+			}
+			logger.Debug("added finalizer to ServiceAccount")
+			if err := r.ensureControllerPermissions(ctx, req.NamespacedName); err != nil {
```
hiddeco, line 121: Do we only need to run this if `controllerutil.AddFinalizer(sa, kargoapi.FinalizerName) == true`?

internal/controller/management/projects/projects.go
```diff
@@ -441,6 +452,87 @@ func (r *reconciler) ensureAPIAdminPermissions(
...
+					"error adding finalizer to controller ServiceAccount %q in namespace %q: %w",
+					sa.Name, sa.Namespace, err,
+				)
+			}
+		}
+
+		roleBindingName := getRoleBindingName(sa.Name)
+		saLogger := logger.WithValues(
+			"serviceAccount", sa.Name,
+			"serviceAccount.namespace", sa.Namespace,
```
hiddeco, line 491: Given it's structured logging, I wonder if we shouldn't make this structured. I.e.:

```suggestion
			"serviceAccount": types.NamespacedName{
				Name: sa.Name,
				Namespace: sa.Namespace,
			},
```

charts/kargo/README.md
krancour: It learns the namespaces from an env var at startup. So it knows what "global" namespaces _should_ have bindings, but it has no way to know what things might have _previously_ been on that list, removed before being restarted, and now ought to _not_ have those bindings. So we can ensure existence easily. We cannot ensure removal.

internal/controller/management/projects/projects.go
krancour: I'd been consciously avoiding that until now just out of a innate sense that it would make log messages much uglier to look at. Let me try it and see how I feel about the result.

internal/controller/management/serviceaccounts/serviceaccounts.go
krancour: You caught a 🐞

internal/controller/management/serviceaccounts/serviceaccounts.go
krancour: It was intentional. I was trying to distinguish between `serviceAccount.namespace`  and `project.namespace`, which would be the NS wherein the RB is being created.

I suppose how we deal with this depends on the resolution of [this other comment](https://github.com/akuity/kargo/pull/2761/files#r1804533377).

👀

internal/controller/management/serviceaccounts/serviceaccounts.go
krancour: Fixed in a837a61b9a1961cdb6b27d584170b6f2b333a1a7

internal/controller/management/projects/projects.go
krancour: Discussed offline and we agreed on a different approach altogether.

internal/controller/management/projects/projects.go
krancour: I tried and it's worse than I thought.

This:

```
logger := logging.LoggerFromContext(ctx).WithValues("serviceAccount", req.NamespacedName)
logger.Debug("reconciling ServiceAccount")
```

Yields messages like this:

```
time="2024-10-17T14:44:01Z" level=debug msg="reconciling ServiceAccount" serviceAccount="{kargo-controller kargo}"
```

Notice that values are shown, but keys are not.

This may be a specific quirk of the underlying logger implementation, but until we change that, I'm not sure that this is an improvement.

wdyt?

hiddeco: I forgot we do not have JSON logging enabled by default. Given this, the current option does indeed seem better.

charts/kargo/README.md
hiddeco, line 148:
```suggestion
| `controller.globalCredentials.namespaces`    | List of namespaces to look for shared credentials. Note that as of v1.0.0, the Kargo controller does not have cluster-wide access to Secrets. The controller receives read-only permission for Secrets on a per-Project basis as Projects are created. If you designate some namespaces as homes for "global" credentials, you will need to manually grant the controller permission to read Secrets in those namespaces.                                                                                                                                                                                                                                                                                                        | `[]`                |
```

internal/controller/management/projects/projects.go
krancour: Definitely something to improve soon-ish.

charts/kargo/README.md
krancour: Fixed in 5e3f652850c7e1ca479cba253b3467fc5d024311

hiddeco (approved):

Issue #2693, opened 2024-10-07
bug: if a step-based promotion fails, the stage get stuck in the promoting phase until a new promotion succeeds


krancour:
This doesn't seem like the right thing to do.

After a Promotion fails or errors, a Stage may be _between_ states -- neither the beginning state nor the target state. (Whether this is true or not would really depend on the particulars of the user-defined promotion process.) imho, this calls for us to move the Stage out of the Promoting phase and into some new phase like "Indeterminate."

@hiddeco any thoughts here?

P.S. Let's _not_ have this issue snowball into one about automated rollbacks. Assuming we will do that at some point in the future, recognizing and categorizing these conditions as called for by this issue is certainly a prerequisite.

hiddeco:
Is the Stage actually still reporting it is promoting (which would be surprising to me, as I think we look at the terminated state of the last promotion)?

Or is this more about the Stage saying it's on Freight X, while it's actually "partially" on Freight Y?

krancour:
It still says phase is `Promoting`, which isn't deliberate. But in terms of fixing it, it seems wrong to put it to `Verifying` or `Steady`.

hiddeco:
Based on the code that synchronizes the Promotions for a Stage, I would expect the Stage to be in a `Steady` phase as soon as the Promotion reaches a terminal state. 🤔

krancour:
So there are two things here.

1. I agree with your interpretation of what the code says _should_ be happening and I want to figure out how I've observed something different.

2. I'm questioning whether what we think the code says _should_ be happening is ideal to begin with. If a Promotion failed, depending on the particulars of the process, "steady" might be a poor descriptor for the currrent state.

Don't worry about this more until I can reproduce it and give some more details about how it happened.

krancour:
After a few days, I have been unable to reproduce this, but I maintain that we need to make some kind of Stage status change after a failed Promo. Steady + Healthy doesn't seem appropriate. Depending on the particulars of the user-defined promotion process, the Stage may be in an invalid state and we don't really have a way to know.

hiddeco:
> I have been unable to reproduce this, but I maintain that we need to make some kind of Stage status change after a failed Promo.

I think the answer to this would be to add Conditions to the Stage, as these are far superior to the current "phase" reflection we do. Every moving element could then get a condition type, allowing you to provide a much better insight into what particular things failed.

krancour:
> I think the answer to this would be to add Conditions to the Stage

[...]

Issue #2152, opened 2024-06-12
Stage shows Running verifications even when all are Failed


jessesuen:
# Description

All my AnalysisRuns are Failed:

```
$ k get ar
NAME                                     STATUS   AGE
dev.01j02dz4tbp75sf7zmvgkw3d3j.9e5f6ab   Failed   44h
dev.01j02hycta7xsjz7xczamptk6c.9d167fb   Failed   43h
dev.01j047c0f8qcgq8fyj3mn6sa0r.ab22279   Failed   28h
dev.01j03w1h1njwvj432nbh8m7xvk.a72c52c   Failed   31h
dev.01j03wxxntp3a3xm5pbpdzzzkd.bd13218   Failed   31h
dev.01j047080w47q6bvgdaqc11tws.d8df953   Failed   28h
dev.01j046aze40vzfvbwfc3mpvpc4.1bda081   Failed   28h
dev.01j03z1dzstgnx5tw85ty3apxr.97465d1   Failed   30h
dev.01j05679g7kccfehb2spejqcx7.151cea4   Failed   19h
dev.01j059yc9xmrej9hna0yspwxsz.3e53240   Failed   18h
dev.01j05me1tzty93z5mgvgpykvp3.1128f45   Failed   15h
```

Yet the UI still shows some as Running:

<img width="835" alt="image" src="https://github.com/akuity/kargo/assets/12677113/6ffa8d56-2040-4dc7-bf22-c25570bd414b">

# Steps to Reproduce

I have a feeling https://github.com/akuity/kargo/issues/2142 may have something to do with this (e.g. we are only processing the last one)

# Version

```
v0.7.0
```

# Logs

```
Paste any relevant application logs here.
```

rbreeze:
@jessesuen and I spoke about this offline and it looks like this may be a backend bug. The UI uses the raw `Status` of the AnalysisRun to generate the icon, so there shouldn't be a way for the mismatch between CLI/UI above to occur. It's possible this was already fixed inadvertently

Marvin9:
Correct @jessesuen , not sure but this could be related to https://github.com/akuity/kargo/issues/2128

UI relies on stage's freight histroy which has verification history. When user runs multiple promotions for same stage, and verifications for those run in parallel without waiting on one-another. Some of the previous verifications status freeze in stage's status but if you open the details of the verification from the UI, the status is correct and exactly what you see when you run `k get ar` for the particular analysis.

There are multiple solutions to this that I can think of.

[...]

Issue #2128, opened 2024-06-06
Keep enqueued promotion in pending state while analysis template is running for the previous promotion


shabbirsaifee92:
# Checklist

* [ x ] I've searched the issue queue to verify this is not a duplicate feature request.
* [ x ] I've pasted the output of `kargo version`, if applicable.
* [ x ] I've pasted logs, if applicable.

kargo version: 0.6.0

# Proposed Feature
**Current Behavior**
When a Promotion (lets call it PromotionA) creates a PR, and another Promotions (PromotionB) is enqueued, it stays in pending state while the PR is open. 
When the PR is merged for PromotionA, a new PR is created for PromotionB while analysis templates are running for PromotionA

<img width="1272" alt="Screenshot 2024-06-06 at 2 12 07 PM" src="https://github.com/akuity/kargo/assets/24420264/40844aaf-92b5-4a3a-95fb-6610688b9fc7">
<img width="1512" alt="Screenshot 2024-06-06 at 2 10 36 PM" src="https://github.com/akuity/kargo/assets/24420264/c2306eb7-3bf4-48c9-ba59-87f809fb7676">

**Suggested/Desired behavior**
PromotionB should be in pending state and only be moved to Running once the verification for previous Promotion is completed and successful.
 
# Motivation
Verifications is an important part of the the delivery process. In many scenarios these templates would be reconciling with argo to ensure that the deployment is actually successful. With the current behavior a user can merge another change while the analysis for the previous promotion is still not complete which makes things confusing. 

# Suggested Implementation

krancour:
I agree completely.

I think @hiddeco and I will have to put our heads together on this one.

hiddeco:
I would argue this could mean that verification becomes part of the Promotion. This does not have to mean verification can't also be run for a Stage, but that they can be triggered from the Promotion (to confirm it was promoted successfully) and (manually) for the Stage (to confirm things are still as they should be).

The easier way, however, would be to check if there are any AnalysisRuns in a non-terminated state before actually starting to work on the Promotion.

shabbirsaifee92:
I'd probably also suggest that any enqueued promotions should be cancelled if the verification fails for the ongoing promotion.

krancour:
> I would argue this could mean that verification becomes part of the Promotion.

[...]

PR #2762, opened 2024-10-17
fix(stage): wait for running verification to finish before next Promotion

hiddeco:
Fixes: #2152
Fixes: #2128 
Fixes: #2693

💡With the Stage mechanism determining what Promotion should be allowed to be processed next, we can get rid of the prioritized queue sometime in the future.

### DIFF
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -22,6 +22,9 @@ const (
 	StagePhasePromoting StagePhase = "Promoting"
 	// StagePhaseVerifying denotes a Stage that is currently being verified.
 	StagePhaseVerifying StagePhase = "Verifying"
+	// StagePhaseFailed denotes a Stage that is in a failed state. For example,
+	// the Stage may have failed to promote or verify its Freight.
+	StagePhaseFailed StagePhase = "Failed"
 )
 
 type VerificationPhase string
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -140,7 +140,7 @@ func SetupReconcilerWithManager(
 	// If Argo CD integration is disabled, this manager will be nil and we won't
 	// care about this watch anyway.
 	if argocdMgr != nil {
-		if err := c.Watch(
+		if err = c.Watch(
 			source.Kind(
 				argocdMgr.GetCache(),
 				&argocd.Application{},
@@ -165,7 +165,7 @@ func SetupReconcilerWithManager(
 		pqs:         reconciler.pqs,
 	}
 	promoWentTerminal := kargo.NewPromoWentTerminalPredicate(logger)
-	if err := c.Watch(
+	if err = c.Watch(
 		source.Kind(
 			kargoMgr.GetCache(),
 			&kargoapi.Promotion{},
@@ -174,6 +174,17 @@ func SetupReconcilerWithManager(
 		),
 	); err != nil {
 		return fmt.Errorf("unable to watch Promotions: %w", err)
+	}
+
+	// Watch Stages that acknowledge their next Promotion and enqueue it
+	if err = c.Watch(
+		source.Kind(
+			kargoMgr.GetCache(),
+			&kargoapi.Stage{},
+			&PromotionAcknowledgedByStageHandler[*kargoapi.Stage]{},
+		),
+	); err != nil {
+		return fmt.Errorf("unable to watch Stages: %w", err)
 	}
 
 	return nil
@@ -326,7 +337,9 @@ func (r *reconciler) Reconcile(
 	// collection it is promoting.
 	if stage.Status.CurrentPromotion == nil || stage.Status.CurrentPromotion.Name != promo.Name {
 		logger.Debug("Stage is not awaiting Promotion", "stage", stage.Name, "promotion", promo.Name)
-		return ctrl.Result{Requeue: true}, nil
+		// Our watch will catch this and requeue the Promotion when the Stage
+		// acknowledges it. Which typically should be faster.
+		return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
 	// Update promo status as Running to give visibility in UI. Also, a promo which
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -274,10 +274,7 @@ func (r *reconciler) Reconcile(
 		return ctrl.Result{}, nil
 	}
 
-	if promo.Status.Phase == kargoapi.PromotionPhaseRunning {
-		// anything we've already marked Running, we allow it to continue to reconcile
-		logger.Debug("continuing Promotion")
-	} else {
+	if promo.Status.Phase == "" {
 		// promo is Pending. Try to begin it.
 		if !r.pqs.tryBegin(ctx, promo) {
 			// It wasn't our turn. Mark this promo as Pending (if it wasn't already)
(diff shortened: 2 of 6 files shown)


api/v1alpha1/stage_types.go
```diff
@@ -22,6 +22,9 @@ const (
 	StagePhasePromoting StagePhase = "Promoting"
 	// StagePhaseVerifying denotes a Stage that is currently being verified.
 	StagePhaseVerifying StagePhase = "Verifying"
+	// StagePhaseFailed denotes a Stage that is in a failed state. For example,
+	// the Stage may have failed to promote or verify its Freight.
+	StagePhaseFailed StagePhase = "Failed"
```
hiddeco, line 27: @krancour happy for this to be named differently, but I think this accomplishes what you were after.

internal/controller/promotions/promotions.go
```diff
@@ -274,31 +285,19 @@ func (r *reconciler) Reconcile(
...
-			if promo.Status.Phase != kargoapi.PromotionPhasePending {
-				err = kubeclient.PatchStatus(ctx, r.kargoClient, promo, func(status *kargoapi.PromotionStatus) {
-					status.Phase = kargoapi.PromotionPhasePending
-				})
-				return ctrl.Result{}, err
-			}
-			return ctrl.Result{}, nil
+	if promo.Status.Phase == "" {
+		// Mark the Promotion as pending.
+		if promo.Status.Phase != kargoapi.PromotionPhasePending {
```
krancour, line 290: Won't the second conditional _always_ be true if the first was?

internal/controller/stages/stages.go
```diff
@@ -1005,8 +1005,32 @@ func (r *reconciler) syncPromotions(
...
+	if curPromotion := status.CurrentPromotion; curPromotion != nil {
+		if curPromotion.Name != latestPromo.Name || latestPromo.Status.Phase.IsTerminal() {
+			status.CurrentPromotion = nil
+		}
+	}
+
+	// If there is any ongoing verification, we need to let it finish before we
+	// can continue with acknowledging the new Promotion.
+	if curFreightCol := status.FreightHistory.Current(); curFreightCol != nil {
+		for _, verification := range curFreightCol.VerificationHistory {
```
krancour, line 1022: Would it make sense to jump right to checking the last one instead of iterating over them all?

internal/controller/stages/stages.go
```diff
@@ -1005,8 +1005,32 @@ func (r *reconciler) syncPromotions(
 	// current state of the Stage.
 	slices.SortFunc(promotions, kargoapi.ComparePromotionByPhaseAndCreationTime)
 
+	// The latest Promotion is the first element in the sorted list.
+	latestPromo := promotions[0]
```
krancour, line 1009: Is latestPromo misleading? If I understand the sort order correctly, it's running promos first, then non-terminal, then terminal. And each of those categories are internally sorted by ULID, which implicitly sorts by creation date. Have I got that right? IF I have got that right, then `promotions[0]` doesn't feel "latest" (most recently created?) to me.

internal/controller/promotions/promotions.go
```diff
@@ -176,6 +176,17 @@ func SetupReconcilerWithManager(
...
+	if err = c.Watch(
+		source.Kind(
+			kargoMgr.GetCache(),
+			&kargoapi.Stage{},
+			&PromotionAcknowledgedByStageHandler[*kargoapi.Stage]{},
+		),
+	); err != nil {
+		return fmt.Errorf("unable to watch Stages: %w", err)
+	}
+
```
krancour, line 189: Nit: If this were moved up to somewhere around line 140, I feel like it would be a little easier to understand where this fits in with everything else.

internal/controller/stages/stages.go
krancour: I believe the logic is sound. It's just the name I found confusing.

krancour: What about something like `priorityPromo`? As, I believe as sorted, `promotions[0]` _is_ the highest priority?

internal/controller/stages/stages.go
hiddeco: The reason I picked this approach is because it makes it independent from an assumption that two verification attempts can't be running at the same time.

As the length of the list is limited to 10 items anyway, I think the cost of the loop is negligible.

internal/controller/stages/stages.go
hiddeco: `highestPrioPromo` then maybe?

internal/controller/promotions/promotions.go
hiddeco: Yes, great catch.

internal/controller/stages/stages.go
```diff
@@ -1005,18 +1005,41 @@ func (r *reconciler) syncPromotions(
 	// current state of the Stage.
 	slices.SortFunc(promotions, kargoapi.ComparePromotionByPhaseAndCreationTime)
 
+	// The Promotion with the highest priority (i.e. a Running or Pending phase)
+	// is the one that we will consider for the current state of the Stage.
+	highestPrioPromo := promotions[0]
+
+	// If the latest Promotion does not match the current Promotion, or is in a
```
krancour, line 1012:
```suggestion
	// If the highest priority Promotion does not match the current Promotion, or
	// is in a terminal phase, then the current Promotion is no longer valid.
```

internal/controller/stages/stages.go
```diff
@@ -1005,18 +1005,41 @@ func (r *reconciler) syncPromotions(
...
+			if !verification.Phase.IsTerminal() {
+				logger.WithValues("verification", verification.ID).Debug(
+					"Stage has a running verification: waiting for it to complete before promoting new Freight",
+				)
+				return status, nil
+			}
+		}
+	}
+
 	// If the latest Promotion is in a running phase, the Stage is promoting.
```
krancour, line 1033:
```suggestion
	// If the highest priority Promotion is in a running phase, the Stage is
	// promoting.
```

krancour:
Test drove and this works quite well, except for one thing, which is that I find after a promo fails, I need to refresh the Stage to queue it for reconciliation and make it acknowledge the new highest priority promo as the current promo.

Something up with the stage reconciler's watch on promo status changes maybe?? A guess.

hiddeco:
I have been unable to reproduce what you describe, but did find a UI bug which I took note of in #2773.

internal/controller/stages/stages.go
```diff
@@ -1005,18 +1005,43 @@ func (r *reconciler) syncPromotions(
...
-	// If the latest Promotion is in a running phase, the Stage is promoting.
-	if latestPromo := promotions[0]; !latestPromo.Status.Phase.IsTerminal() {
-		logger.WithValues("promotion", latestPromo.Name).Debug("Stage has a running Promotion")
+	// The Promotion with the highest priority (i.e. a Running or Pending phase)
+	// is the one that we will consider for the current state of the Stage.
+	highestPrioPromo := promotions[0]
+
+	// If the highest priority Promotion does not match the current Promotion, or
+	// is in a terminal phase, then the current Promotion is no longer valid.
+	// terminal phase, then the current Promotion is no longer valid.
```
hiddeco, line 1014:
```suggestion
```

internal/controller/promotions/watches.go
```diff
@@ -239,3 +239,78 @@ func (u *UpdatedArgoCDAppHandler[T]) Update(
...
+		return
+	}
+
+	if oldStage.Status.CurrentPromotion == nil && newStage.Status.CurrentPromotion == nil {
+		return
+	}
+
+	if oldStage.Status.CurrentPromotion != nil && newStage.Status.CurrentPromotion == nil {
+		return
+	}
```
krancour, line 299: This logic reduces to:

```suggestion
	if newStage.Status.CurrentPromotion == nil {
		return
	}
```

internal/controller/promotions/watches.go
```diff
@@ -239,3 +239,78 @@ func (u *UpdatedArgoCDAppHandler[T]) Update(
...
+
+	if oldStage.Status.CurrentPromotion == nil ||
+		oldStage.Status.CurrentPromotion.Name == newStage.Status.CurrentPromotion.Name {
```
krancour, line 302: This seems responsible for the issues I've been observing:

```suggestion
	if oldStage.Status.CurrentPromotion == nil ||
		oldStage.Status.CurrentPromotion.Name != newStage.Status.CurrentPromotion.Name {
```

krancour (approved):

PR #2789, opened 2024-10-21
feat: add argocd application info to promotion event

hanxiaop:
This PR adds the applications updated during promotion to the event, so we can get the related application info. Also, I’ve moved `NewPromotionEventAnnotations` from the api directory to internal/api since it’s only used in internal packages, to avoid cyclic imports.

The sample results will be something like:
```
...
 metadata:
    annotations:
      event.kargo.akuity.io/actor: admin
      event.kargo.akuity.io/applications: '[{"name":"kargo-demo-test"}]'
      event.kargo.akuity.io/freight-alias: billowing-shark
...
```


internal/api/event.go
```diff
@@ -0,0 +1,101 @@
+package api
```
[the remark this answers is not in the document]
hanxiaop: There are two main reasons:
1. The function is only used in internal packages. 
2. If I keep it in the original directory, importing `ArgoCDUpdateConfig` from the internal package will cause `import cycle not allowed`, as the `v1alpha1/` package is also imported in the internal package.

internal/api/event.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+		}
+	}
+
+	var apps []AppReference
+	for _, step := range p.Spec.Steps {
+		if step.Uses != "argocd-update" || step.Config == nil {
+			continue
+		}
+		var cfg directives.ArgoCDUpdateConfig
```
krancour, line 79: We're trying really hard to make built-in directives "not special." Ideally, the rest of Kargo would have as little awareness of their internals as it will eventually have of the internals of _third-party_ directives.

That being said, I understand this might seem like the only practical way to get the information you're trying to include in the events. This, unfortunately, wouldn't be the first spot where we compromised on this ideal because it seemed like the only way to get something done.

Ultimately, I think this ties into a _much broader_ question of how coupled do we want to be or _not_ want to be to Argo CD. Implementation details aside, just including this information in the Promotion event to begin with says something about Kargo's relationship to Argo CD being special in some way. I think this may be a question for @jessesuen.

internal/api/event.go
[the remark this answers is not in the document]
krancour: I understand the import cycle you're trying to avoid, but I don't think this package is a good home for this. (idk if its old home was any better, tbh.)

"API" is overloaded. I wish _this_ package had been named `apiserver` instead, as this package is all API server endpoints, interceptors, etc. and this functionality you're moving to here is used much more broadly -- by the controller, by the webhooks, etc.

Is it possible this might be better off living in the existing `event` package (e.g. Live alongside things like `func NewRecorder()`)?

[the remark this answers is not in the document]
hanxiaop: @krancour I tried placing it in the existing `event` package, but realized it is better for k8s event type. Since I couldn’t find a better location, I’ve refactored some files into internal folders in https://github.com/akuity/kargo/pull/2808. PTAL if this makes sense. Thank you!

internal/api/event.go
hanxiaop: > We're trying really hard to make built-in directives "not special." Ideally, the rest of Kargo would have as little awareness of their internals as it will eventually have of the internals of third-party directives.

The intention of maintaining flexibility and avoiding any special logics for built-in and third-party directives makes sense to me. However I'm still a bit unclear about the design here. From my perspective, this little awareness is from code level instead of user level. Users still see the results the directives processed, but developers are limited to access the output of runners they have written.

> Ultimately, I think this ties into a much broader question of how coupled do we want to be or not want to be to Argo CD. Implementation details aside, just including this information in the Promotion event to begin with says something about Kargo's relationship to Argo CD being special in some way.

I would also like to know the idea behind https://github.com/akuity/kargo/pull/2545. As a Kargo newbie, when I saw the `argocd-update` step in`Promotion` resource, I would expect there are some ArgoCD integrations with Kargo. Otherwise it seems we might not need this directive from the lower-level implementations. Seems like this can also be related to how we can integrate with projects that manage workloads?

internal/kubernetes/event/event.go
```diff
@@ -140,3 +144,93 @@ func (e *sink) Patch(event *corev1.Event, data []byte) (*corev1.Event, error) {
 	err := e.client.Patch(e.ctx, event, libClient.RawPatch(types.StrategicMergePatchType, data))
 	return event, err
 }
+
+type AppReference struct {
```
hiddeco, line 148: Is there any reason not to use a `types.NamespacedName` here?

hanxiaop: `types.NamespacedName` does not omit an empty namespace value. If the namespace is empty, it will be something like:
```
{
  "Namespace" : "",
  "Name" : "app-name"
}
```
instead of 
```
{
  "Name" : "app-name"
}
```

hiddeco: Is it desirable for the namespace to be empty? I would think that in terms of traceability, it would be good to always include the namespace as it will otherwise be hard to derive the default without knowing how the controller is further configured.

hanxiaop: Makes sense to me. Revised to `types.NamespacedName` with a default namespace value to provide better understanding on the namespace without any additional context.

internal/api/event.go
hanxiaop: @hiddeco I tried to move the entire `api/v1alpha1/event.go` file and kubernetes events related files to `internal/event` in https://github.com/akuity/kargo/pull/2808. However as you can see, this would require moving multiple files since some helpers rely on functions within that file. Moving just one function to internal/event doesn’t seem to make much sense. I’d also like to wait for consensus on the best location for this function.

internal/kubernetes/event/event.go
```diff
@@ -140,3 +145,92 @@ func (e *sink) Patch(event *corev1.Event, data []byte) (*corev1.Event, error) {
 	err := e.client.Patch(e.ctx, event, libClient.RawPatch(types.StrategicMergePatchType, data))
 	return event, err
 }
+
+// NewPromotionEventAnnotations returns annotations for a Promotion related event.
+// It may skip some fields when error occurred during serialization, to record event with best-effort.
+func NewPromotionEventAnnotations(
```
hiddeco, line 151: Please move this to `internal/event/promotion.go`.

hiddeco: Plus maybe rename it to `NewPromotionAnnotations`, as `event` already should be enough to indicate it's about events.

hiddeco: Additional candidate to be moved from the public API package is `FormatEventUserActor` (-> `FormatUserActor`).

hanxiaop: @hiddeco Thank you for the suggestion! I've moved this function to `internal/event/promotion.go`.

> Additional candidate to be moved from the public API package is FormatEventUserActor (-> FormatUserActor).

I've tried this but it still caused the import cycle issue like:
```
0.602   imports github.com/akuity/kargo/api/v1alpha1
0.602   imports github.com/akuity/kargo/internal/event
0.602   imports github.com/akuity/kargo/api/v1alpha1: import cycle not allowed
```
I think we could only move `NewPromotionAnnotations` here for minimal changes. We can discuss further improvements if necessary and I'd be very happy to contribute more.

hiddeco: That's fair.

krancour (approved):

PR #2808, opened 2024-10-22
chore: refactor some files from the api directory to internal

hanxiaop:


krancour:
What is the goal of this PR? I would never make a claim that everything is organized 100% optimally as is, but these changes move a lot of important functionality out of a package where it made sense for that functionality to live into a nondescript "helpers" package. Helper packages are almost always a code smell. With some careful thought, there is almost always a better answer than moving functionality into a package whose name does not reflect a clear purpose. It becomes a virtual certainty that such a package will become a "junk drawer" over time.

hanxiaop:
> What is the goal of this PR?

@krancour sorry for not writing the description and causing confusion. I agree that `helpers` is not a great name for the moved files. I was trying to move just [event.go](https://github.com/akuity/kargo/blob/main/api/v1alpha1/event.go) file to a more general event directory that includes both kargo related events and k8s events, which could also unblock [PR #2789](https://github.com/akuity/kargo/pull/2789). However, I realized this isn’t possible since the functionality in those helper files in the kargo api direcotry also rely on some functionality in that file. I think it would be better to separate the client-related functionality from the API definitions to improve modularity and reusability.

hanxiaop:
Those functionalities you mentioned are coupled with some functions in `event.go`, but it is not a mandatory refactor as I was just trying to separate useful functionality from the API definitions. I'll close this PR and go with your suggestion in https://github.com/akuity/kargo/pull/2789#discussion_r1809632547 with the minimum refactor. Thank you for the review!

Issue #2831, opened 2024-10-24
OCI repository with Chart and Image : warehouse display both (no matter if it's an chart or image warehouse)


flbla:
# Checklist

* [X] I've searched the issue queue to verify this is not a duplicate bug report.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `kargo version`.
* [X] I've pasted logs, if applicable.

# Description

I have an OCI repository with Charts and Images in it.
When I create a warehouse (image or chart) it display charts AND images versions.

# Screenshots

![image](https://github.com/user-attachments/assets/33d5fc3a-0293-4629-87f3-c31f4696fbd1)

![image](https://github.com/user-attachments/assets/525dd1a2-c9b6-4c7f-9eea-2d19763c65cc)

![image](https://github.com/user-attachments/assets/7748bbf6-a2bf-4bcc-916b-d2db3f559052)

# Steps to Reproduce

An OCI repo with charts and images in it.
Create a chart warehouse connected to this repo.
=> Images version are also displayed

# Version

```
/ $ kargo version
{
  "Version": "v1.0.1",
  "BuildDate": "2024-10-22T00:33:07Z",
  "GitCommit": "9fc3e3b417de09d4fd561ceeb1d87cece989bd4a",
  "GitTreeDirty": false,
  "GoVersion": "go1.23.2",
  "Compiler": "gc",
  "Platform": "linux/amd64"
}
```

hiddeco:
I suspect you are subscribed to images using something like a tag selector. When Kargo selects images using such a selector, it relies solely on the tag list provided by the registry API, which is a simple list containing strings. Because of this, it has no idea about the image type (which would require at least one additional request **per listed tag**).

While I see us benefiting from an image type selector complementary to the existing platform selector at some point, I would not see us enabling this by default, as it changes the complexity from `O(1)` to `O(n)`.

Until we have such a feature, I can see a couple of ways for you to work around this problem for container images:

1. Use an `allowTags` regex expression, as I can see a pattern of `v1.0.0.` versus `1.0.0`.
2. Not putting all your eggs in the same basket, i.e. have separate image repositories for Helm charts versus images. 

---

For the Helm chart discovery part, the quick fix would be for us to ignore anything that isn't a strict SemVer (which is how `helm` itself determines what is and isn't a chart): https://github.com/helm/helm/blob/v3.16.1/pkg/registry/client.go#L682-L690

I will draft a pull request to enforce this.

PR #2837, opened 2024-10-24
fix(helm): handle OCI version constraints and metadata

hiddeco:
Fixes the second bit of https://github.com/akuity/kargo/issues/2831#issuecomment-2435595674

---

When working with OCI-based Helm charts, version handling needs to be
stricter than with classic HTTP repositories:

1. OCI chart versions must be strict SemVer (e.g. "1.0.0" not "v1.0")
2. OCI registry tags cannot contain "+" characters used in SemVer
   build metadata

To solve this:

- Add strict SemVer validation for OCI chart versions
- Convert between OCI's "_" and SemVer's "+" for build metadata

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"os"
-	"os/exec"
 	"slices"
 	"strings"
 
@@ -15,8 +13,6 @@ import (
 	"oras.land/oras-go/pkg/registry"
 	"oras.land/oras-go/pkg/registry/remote"
 	"oras.land/oras-go/pkg/registry/remote/auth"
-
-	libExec "github.com/akuity/kargo/internal/exec"
 )
 
 // DiscoverChartVersions connects to the specified Helm chart repository and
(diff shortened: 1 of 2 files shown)


internal/helm/helm.go
```diff
@@ -185,10 +183,25 @@ func getChartVersionsFromOCIRepo(
 
 // versionsToSemVerCollection converts a slice of versions to a semver.Collection.
 // Any versions that cannot be parsed as SemVer are ignored.
-func versionsToSemVerCollection(versions []string) semver.Collection {
+func versionsToSemVerCollection(versions []string, isOCI bool) semver.Collection {
+	newSemver := semver.NewVersion
+	if isOCI {
+		// Image tags produced by Helm are STRICT SemVer, meaning that they
```
krancour, line 189: Nit: Do we really want to say "Image tags" here? Applies in a few other places as well.

Totally non-blocking.

internal/helm/helm.go
```diff
@@ -223,84 +242,6 @@ func filterSemVers(semvers semver.Collection, semverConstraint string) (semver.C
 	return filtered, nil
 }
 
-// Login runs `helm registry login` or `helm repo add` for the provided
-// repository. The provided homePath is used to set the HOME environment
-// variable, as well as the XDG_* environment variables. This ensures that Helm
-// uses the provided homePath as its configuration directory, and allows for
-// isolation.
-func Login(homePath, repository string, credentials Credentials) error {
```
krancour, line 231: Dead code. Nice find.

krancour (approved):
Presuming you checked this out e2e, LGTM.

internal/helm/helm.go
hiddeco: Changed it to "OCI artifact tags".

hiddeco:
@krancour I now thoroughly tested it end-to-end, and for our use case, it is better to stay true to the "Helm version". Reason being that ArgoCD and other bits where we utilize the version want the proper SemVer version, and not the registry one.

krancour:
I'm not sure I followed that. Could you give an example?

hiddeco:
Assuming there is a chart with `1.0.0+build.1`, this would get pushed to an OCI registry as `1.0.0_build.1`.

When the Warehouse is gathering chart versions, it now has two options:

1. Store the version reference to equal the tag.
2. Store the version reference to equal the strict SemVer.

The problem, if you go with option 1, is that if Freight is created with this and later consumed by a Promotion, this format is not valid to e.g. ArgoCD or any other consumer, as Helm always expects to receive strict SemVers when e.g. doing `helm pull`, etc.

Given this, it is better to store the OCI tag as a proper SemVer.

krancour:
Ok. Then I understood it already after all. This appears to be what you're doing? I think I was confused because I'd somehow misinterpreted your previous comment as second-guessing this change.

hiddeco:
I was not! But my initial  (now overwritten) approach was doing option 2. We are all good here! 😄

PR #2841, opened 2024-10-25
fix(docs): kustomize build folder of uat & prod

ntheanh201:
Currently, when make a promotion from test to uat or uat to prod, the kustomize still build the manifest of test stage, it makes the application cannot start because of the same nodePort, and the html content is not correct


ntheanh201:
Can you check it? @krancour

hiddeco (approved):
Thank you @ntheanh201 🙇

PR #2846, opened 2024-10-25
fix: Clicking multiple times on same freight not working

ddeath:
Fix: Clicking multiple times on same freight not working

Having components defined inside another components is not the best practice especially without useMemo or something similar since it is quite easy to trigger a lot of re-renders.

In this instance it was triggering loop somehow (not sure 100% what it was causing, my tip is that tooltip) which ends up in:
Screenshot 2024-10-25 at 21 12 56

Since there is no really good reason to have it (FreightContentItem) defined inside another component I extracted it as separate component which fixes the issue.

You can replicate the issue by clicking on highlighted area multiple times:
Screenshot 2024-10-25 at 21 45 25

Note that after first click it will open details of Freight, after second click it will close it and on third click it will do nothing (it should open details again but it does not)... This happens only if you keep mouse in highlighted area. If you go up and down (git / docker) then it kind of works....

With this commit you can click as many times you want on the same spot and it will keep opening and closing those details without hitting:
```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```

### DIFF
--- a/ui/src/features/freight-timeline/freight-content-item.tsx
+++ b/ui/src/features/freight-timeline/freight-content-item.tsx
@@ -16,33 +16,35 @@ export const FreightContentItem = (props: {
   highlighted: boolean;
   linkClass: string;
 }) => {
-  const { horizontal, dark, highlighted, overlay, title, icon, href, children, linkClass} = props
-  return <Tooltip
-    className={classNames('min-w-0 flex items-center justify-center my-1 rounded', {
-      'flex-col p-1 w-full': !horizontal,
-      'mr-2 p-2 max-w-60 flex-shrink': horizontal,
-      'bg-black text-white': dark,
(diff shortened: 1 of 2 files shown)


Marvin9 (approved):
LGTM! Please run `pnpm --dir ui lint:fix`

ddeath:
@Marvin9 done.

One suggestion would be to create one command which, would verify everything including linter so we can run it locally without knowing which parts needs to be checked since Github checks do not fail automatically.

Marvin9:
> @Marvin9 done.
> 
> One suggestion would be to create one command which, would verify everything including linter so we can run it locally without knowing which parts needs to be checked since Github checks do not fail automatically.

Sure thing. There is already a make command `make format` but agree that should be in doc if not already

PR #2848, opened 2024-10-25
refactor: create separate reconciler for control flow Stage

hiddeco:


internal/controller/stages/control_flow_stages.go
```diff
@@ -0,0 +1,603 @@
...
+// watches on the required objects.
+func (r *ControlFlowStageReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	// Configure client and event recorder using manager.
+	r.client = mgr.GetClient()
+	r.eventRecorder = libEvent.NewRecorder(ctx, mgr.GetScheme(), mgr.GetClient(), r.cfg.Name())
+
+	// This index is used to find all Freight that are directly available from
+	// a Warehouse. It is used to find Freight that can be sourced directly from
+	// the Warehouse for the control flow Stage.
+	if err := mgr.GetFieldIndexer().IndexField(
```
hiddeco, line 59: This is a matter of personal preference, but I never liked the way the registration was factored out because it hid two important things from me when I wanted to know what a reconciler cared about:

1. The index field that's important to the reconciler.
2. The actual thing that decides how the value of this field is determined.

Given this, I took the liberty of not using the abstraction but rather doing the registration by hand here.

internal/controller/stages/control_flow_stages.go
```diff
@@ -0,0 +1,603 @@
...
+) *ControlFlowStageReconciler {
+	return &ControlFlowStageReconciler{
+		cfg: cfg,
+	}
+}
+
+// SetupWithManager sets up the control flow Stage reconciler with the given
+// controller manager. It registers the reconciler with the manager and sets up
+// watches on the required objects.
+func (r *ControlFlowStageReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
```
hiddeco, line 51: Even if I like to be idiomatic and puristic at times, this is how controller-runtime projects do it — and would fit more into what we have set out for ourselves with #1479.

https://xkcd.com/927/

internal/controller/stages/event_handlers.go
```diff
@@ -317,61 +343,71 @@ func (u *updatedArgoCDAppHandler[T]) Generic(
...
-		)
-		logger.Debug(
-			"enqueued Stage for reconciliation",
-			"namespace", stage.Namespace,
-			"stage", stage.Name,
-			"app", newApp.Name,
-		)
+	// If the Stage is a control flow Stage, there is no need to reconcile it.
+	if stage.IsControlFlow() {
+		return
```
hiddeco, line 394: Previously, this would always result in a reconciliation — even if there wasn't any actual interest.

internal/controller/stages/event_handlers.go
```diff
@@ -478,6 +513,11 @@ func (p *phaseChangedAnalysisRunHandler[T]) Update(
 			)
 		}
 		for _, stage := range stages.Items {
+			// If the Stage is a control flow Stage, there is no need to reconcile it.
+			if stage.IsControlFlow() {
+				continue
```
hiddeco, line 518: Previously, this would always result in a reconciliation — even if there wasn't any actual interest.

internal/controller/stages/control_flow_stages.go
```diff
@@ -0,0 +1,603 @@
...
+		*status = newStatus
+	}); err != nil {
+		// Prioritize the reconcile error if it exists.
+		if reconcileErr != nil {
+			logger.Error(err, "failed to update Stage status after reconciliation error")
+			return ctrl.Result{}, reconcileErr
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to update Stage status: %w", err)
+	}
+	return ctrl.Result{}, reconcileErr
```
hiddeco, line 215: I need to test this more thoroughly, but I believe there is no actual reason for us to requeue on an interval here because we should be informed by a watcher or change event.

internal/controller/stages/control_flow_stages.go
krancour: Far from being a hill I would die on, I think the advantage of the other approach was that a caller in another package could obtain a fully-initialized reconciler after a single function call. With this approach, it requires two, which creates the possibility of obtaining a _partially-initialized_ reconciler.

As I said, it's not a hill I would die on. Merely explaining the rationale behind the prior approach.

internal/controller/stages/control_flow_stages.go
krancour: Excellent rationale. 👍

internal/controller/stages/control_flow_stages.go
hiddeco: Tested this, and what's expected from it is triggered just fine without it.

krancour (approved):

Issue #2793, opened 2024-10-21
UI does not redirect to login page 


alexmt:
# Description

UI does not redirect user to login page if token is invalid.

# Screenshots

<img width="1711" alt="Screenshot 2024-10-21 at 9 04 03 AM" src="https://github.com/user-attachments/assets/10b93b0f-4e3d-47a7-9093-b58dac7a0dc4">

# Steps to Reproduce

1. Navigate to kargo main page with invalid token
2. See empty projects page and "invalid token" error message

I expected to be redirected to login page (even better redirected to sso login )

# Version

v1.0.0

PR #2866, opened 2024-10-28
chore(ui): perform token validity check

Marvin9:
fixes https://github.com/akuity/kargo/issues/2793

Thanks @krancour for help. We couldn't exactly find the reason but token validity check would fix it.


krancour:
When we talked this through yesterday, I thought we had concluded that a new endpoint wasn't necessary. i.e. I thought that 401s from _existing_ API calls just needed to be handled with a redirect to the login page.

Marvin9:
Discussed with @krancour and we think its enough to handle API failure due to auth checks in UI itself. All the normal cases are handled beforehand by interceptor

ui/src/config/transport.ts
```diff
@@ -55,6 +55,12 @@ export const newErrorHandler = (handler: (err: ConnectError) => void): Intercept
 
       handler(err);
 
+      // in rare cases, token is invalid but UI could not detect it beforehand
+      // CodeUnauthenticated <- to ease the global code search
+      if (err instanceof ConnectError && err?.message.includes('unauthenticated')) {
+        logout();
+      }
```
rpelczar, line 62: Better practice is to check the status code. Can you change it?

rpelczar (approved):

Issue #2836, opened 2024-10-24
Warehouse does not reconcile


wmiller112:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

This is a continuation of this [issue](https://github.com/akuity/kargo/issues/2544). I continue to experience warehouses not respecting spec.interval or reconciling when changed. I've observed the same behavior across ~20 different projects. It does not seem related to git/image repo because when the reconcile does finally trigger, it executes in a matter of seconds, as can be seen in the screenshot of logs. The reconcile occurs at fairly regular interval of close to 15-20 minutes when the interval is set to any time, ranging from 2m to 10m. Triggering a sync via gui adds annotation to warehouse, but even this does not trigger the reconcile. It eventually reconciles on the same interval.

# Screenshots

![Screenshot 2024-10-24 at 8 59 21 AM](https://github.com/user-attachments/assets/3dec2b69-37e1-40ae-8a5c-8f77caf3a091)

# Steps to Reproduce

- Create a warehouse with spec similar to

```
spec:
  freightCreationPolicy: Automatic
  interval: 2m0s
  subscriptions:
    - git:
        branch: main
        commitSelectionStrategy: NewestFromBranch
        discoveryLimit: 20
        includePaths:
          - regex:path/to/manifests/.*
        repoURL: https://github.com/<org>/<repo>
    - image:
        allowTags: ^production-[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}-[0-9]{2}-[0-9]{2}Z-[0-9a-f]{40}$
        discoveryLimit: 20
        gitRepoURL: https://github.com/<org>/<repo>
        imageSelectionStrategy: Lexical
        repoURL: xxxxxxxxxx.dkr.ecr.us-west-2.amazonaws.com/<repo>
```
- Click refresh on the warehouse in gui
- Watch it spin until it eventually reconciles

I'd expect immediately after clicking reconcile, the modified warehouse would be seen by the controller and reconcile - triggering a freight fetch.

# Version

```
Client Version: v1.0.3
Server Version: v1.0.3
```

krancour:
@wmiller112 first, my apologies. I really thought we'd found it and nailed it in #2544 and #2754.

[...]

Issue #2544, opened 2024-09-18
Delayed freight discovery


wmiller112:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

Frequently I've found that new image tags will be available that match the specified warehouse configuration, but they will not be discovered in a timely manor. The warehouse is configured with spec.interval of 2m0s, but with debug logs enabled, I see the warehouse reconcile loop happens far less frequently, but its often random. Clicking the refresh button also does not seem to reliably trigger the check for new artifacts.

# Steps to Reproduce

- Create a warehouse with spec similar to
```
spec:
  freightCreationPolicy: Automatic
  interval: 2m0s
  subscriptions:
    - git:
        branch: main
        commitSelectionStrategy: NewestFromBranch
        discoveryLimit: 20
        includePaths:
          - regex:path/to/manifests/.*
        repoURL: https://github.com/<org>/<repo>
    - image:
        allowTags: ^production-[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}-[0-9]{2}-[0-9]{2}Z-[0-9a-f]{40}$
        discoveryLimit: 20
        gitRepoURL: https://github.com/<org>/<repo>
        imageSelectionStrategy: Lexical
        repoURL:xxxxxxxxxx.dkr.ecr.us-west-2.amazonaws.com/<repo>
```
- Push a tag matching the allowTags schema
- Monitor debug logs in controller to observe the refresh does not happen at the specified `interval`
- Click `refresh` on the warehouse, confirm the annotation is added with updated ts
- Continue to monitor debug logs to see that frequently a refresh will not happen within specified `interval` of clicking refresh
- Eventually a reconcile loop will run, new tag discovered, and the rest of the process works as expected

# Version

```
Client Version: v0.8.4
Server Version: v0.8.4
```

# Logs

```
time="2024-09-18T15:48:08Z" level=debug msg="reconciling Warehouse" namespace=<project> warehouse=sources
time="2024-09-18T15:48:08Z" level=debug msg="obtained credentials for git repo" namespace=<project>repo="https://github.com/<org>/<repo>" warehouse=sources
time="2024-09-18T15:48:20Z" level=debug msg="discovered commits" count=20 namespace=<project> repo="https://github.com/<org>/<repo>" warehouse=sources
[...]

PR #2872, opened 2024-10-29
feat: resource and concurrency settings

krancour:
This PR is a bundle of resource utilization and concurrency improvements @hiddeco and I agreed on after extensively troubleshooting #2544 and #2836 with @wmiller112.

* Sets GOMAXPROCS on all components to match (explicit or implicit) CPU limit on the container. i.e. GOMAXPROCS = cores available rounded up to the nearest integer. This prevents Go from backing goroutines with a number of OS threads exceeding the number of cores available, which is a condition that results in losing compute time to avoidable context switches. This value is logged at startup for all components to assist in troubleshooting efforts when necessary.

* Sets GOMEMLIMIT (soft memory limit) on all components to match (explicit or implicit) memory limits on the container. This help Go to optimize GC. This value is logged at startup for all components to assist in troubleshooting efforts when necessary.

* Allows `MaxConcurrentReconciles` to be set on a per-reconciler basis in both the controller(s) and management controller. There is a default of four for each. For Warehouses especially, the ability to tune this parameter may help decrease wait times for reconciliation of Warehouses overall when some Warehouses, due either to their configuration or to rate limits, take a long time to reconcile. The maximum number of concurrent reconciles for is logged at startup for all reconcilers to assist in troubleshooting efforts when necessary.


cmd/controlplane/controller.go
```diff
@@ -3,12 +3,13 @@ package main
 import (
 	"context"
 	"fmt"
+	"runtime"
```
hiddeco, line 6:
```suggestion
	stdruntime "runtime"
```

might be better, as we utilize `k8s.io/apimachinery/pkg/runtime` more than the stdlib.

charts/kargo/values.yaml
```diff
@@ -377,6 +377,21 @@ controller:
...
+      maxConcurrentReconciles: 4
+    promotions:
+      ## @param controller.reconcilers.promotions.maxConcurrentReconciles specifies the maximum number of Promotion resources the controller can reconcile concurrently.
+      maxConcurrentReconciles: 4
+    stages:
+      ## @param controller.reconcilers.stages.maxConcurrentReconciles specifies the maximum number of (non-control flow) Stage resources the controller can reconcile concurrently.
+      maxConcurrentReconciles: 4
+    warehouses:
+      ## @param controller.reconcilers.warehouses.maxConcurrentReconciles specifies the maximum number of Warehouse resources the controller can reconcile concurrently.
+      maxConcurrentReconciles: 4
```
hiddeco, line 393: Would it be an idea to have a default fallback, in case you just want to scale lazily?

cmd/controlplane/controller.go
krancour: That sgtm.

charts/kargo/values.yaml
krancour: Not sure I follow? It defaults to 4 in the code as well. I debated leaving it undefined here, but decided this was more clear.

hiddeco: I mean something like: `{{ default .Values.reconcilers.global.maxConcurrentReconciles .Values.reconcilers.stages.maxConcurrentReconciles }}`. Changing the global value would cause all values to change _unless_ they have a kind specific value set.

hiddeco: Not to keen on the `.global` here by the way, but you get the point.

charts/kargo/templates/api/deployment.yaml
```diff
@@ -59,9 +59,19 @@ spec:
...
+          - name: GOMEMLIMIT
+            valueFrom:
+              resourceFieldRef:
+                containerName: api
+                resource: limits.memory
+          - name: GOMAXPROCS
+            valueFrom:
+              resourceFieldRef:
+                containerName: api
+                resource: limits.cpu
```
hiddeco, line 72: Think this may need `if` wrapping to ensure `limits.memory` is set?

charts/kargo/values.yaml
krancour: I get your meaning now. I can do something like that without the nasty global bit.

charts/kargo/templates/api/deployment.yaml
krancour: I tried it. With no limits set, they come through as all CPU and memory on the node.

charts/kargo/values.yaml
krancour: Addressed in f7c2294bfc357a01fb84dde318adb1aebb231f22

hiddeco (approved):
💯

Issue #2845, opened 2024-10-25
git-open-pr : "Another open merge request already exists for this source branch" : retrieve the ID of the MR/PR and go next step


flbla:
# Checklist

* [X] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

When a MR is already opened retrieve it's ID and continue with next step

# Motivation

Stage is failing but it should not

Brightside56:
same on 0.9.1

krancour:
@hiddeco, give me a sanity check?

The legacy promo mechs in v0.9.x and earlier used to store the promotion number in the Promotion's bygone `metadata` map -- a precursor to promotion steps' shared state. It _had_ to do this, because the legacy promo mechs replayed in their entirety on every reconciliation and there had to be some way of knowing the PR in question had already been opened.

For reference:

https://github.com/akuity/kargo/blob/e746b7b0e7500b79602619c0eaa6da97a7851f24/internal/controller/promotion/git.go#L218-L223

We currently have no equivalent to this check in the `git-open-pr` step and that is what needs to be rectified to resolve this issue.

Promotion steps only repeat themselves if we're _forced_ to start a Promotion over from the beginning because we lost its working directory -- which would pretty much only be because the controller restarted while the Promotion was in-progress.

You can see this right here:

https://github.com/akuity/kargo/blob/c462c76edfd53821dbf7d5ec8aca24ed1fca6cf1/internal/controller/promotions/promotions.go#L463-L480

You'll notice we clear the shared state if we determine we're starting over. If we reset the starting step back to zero, but _don't_ clear the shared state right here, the `git-open-pr` step would be able to self-determine that it's already done what it needs to do. (Perhaps, for good measure, we _update_ the PR in this case in case any previous step that has now re-played yielded different results than the first time through. Such a thing could happen, for instance, with Helm charts that render anything that is random or time-based.)

I cannot think of any reason _not_ to make this change, but as I said... I wanted a quick sanity check. lmk

hiddeco:
I think there are two scenarios we need to consider:

1. The scenario you described with Promotion replays due to a lost working directory.
2. A second variation where a previous Promotion failed/terminated after PR creation, which can happen because:

   - Users can use static branches for PRs
   - Our branch names are tied to Project/Stage name, not specific Promotions
[...]

PR #2881, opened 2024-10-30
feat(controller)!: assign a default alias to promotion steps that have none

krancour:
Part of #2845 

This PR make the step execution engine assign a default alias to any promotion step that doesn't explicitly have an alias specified.

This makes it such that _all_ steps leave behind outputs in the shared state map where that was only possible before for steps that had explicitly been assigned an alias.

This creates the possibility of modifying some step implementations, in a follow-up, to consult any output they themselves may have produced in a previous execution.

This is braking only insofar as it reserves step aliases of the form `^step-\d+$` for system use.

### DIFF
--- a/api/v1alpha1/promotion_types.go
+++ b/api/v1alpha1/promotion_types.go
@@ -1,8 +1,6 @@
 package v1alpha1
 
 import (
-	"regexp"
-
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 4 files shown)


internal/directives/simple_engine.go
```diff
@@ -13,6 +15,8 @@ import (
 	"github.com/akuity/kargo/internal/credentials"
 )
 
+var forbiddenStepAliasRegex = regexp.MustCompile(`^step-\d+$`)
```
hiddeco, line 18: Maybe better to make this publicly accessible and have the webhook depend on it?

krancour: I didn't want to make the webhook dependent on this package or vice versa, but since you've called me out on this, I probably _should_ move this to `api/v1alpha1`. Pretty much everything already depends on that and that is arguably a better location for this anyway.

Fix incoming.

hiddeco (approved):

api/v1alpha1/promotion_types.go
```diff
@@ -33,6 +35,10 @@ const (
 	PromotionPhaseAborted PromotionPhase = "Aborted"
 )
 
+// ReservedStepAliasRegex is a regular expression that matches step aliases that
+// are reserved for internal use.
```
hiddeco, line 39:
```suggestion
// ReservedStepAliasRegex is a regular expression that matches step aliases that
// are reserved for internal use.
//
// +protobuf=false
```

hiddeco (approved):

PR #2894, opened 2024-11-04
fix(stages): auto-promote Freight from other origins

hiddeco:
Found while refactoring the Stage reconciler, and seems important enough to justify a patch for 1.0.x.


krancour (approved):
Nice find.

PR #2896, opened 2024-11-04
fix(controller): fix pr already open issue with git-open-pr step

### LINKED ISSUE #2845: git-open-pr : "Another open merge request already exists for this source branch" : retrieve the ID of the MR/PR and go next step

krancour:
Fixes #2845

Follows up on #2878 and #2881 

This makes the `git-open-pr` step's short-circuiting a little more resilient.

Prior to this trio of related fixes, if the controller restarted while a Promotion was still in-progress waiting for a PR to be merged, when the controller came back up and found the Promotion's state gone from the ephemeral disk, it would start the entire user-defined Promotion process over again from step 0, potentially getting into trouble when it came time to open a PR, since a PR identical to the one it would otherwise open already exists. (Same source branch, same commit at the head of that branch, same base branch.)

This PR, the final component of the fix, uses a two-phased approach for the `git-open-pr` step to determine if a PR actually needs to be opened or not:

* Looks at state that may have been left behind by a previous successful execution. If a PR number is found therein, short-circuit as if successful and continue on to the next step.

* For good measure, looks for a PR identical to the one it may open. If found, it will short-circuit as described above. This makes the step more resilient to the remote possibility that _somehow_ a PR identical to the one it would otherwise open was created through other means. It will work with that PR  as if it had opened it itself.

  __Edit:__ This second check is also helpful in the case that someone's got unmerged PRs open when they upgrade from an existing v1.0.x to v1.1.0.

Both of these are resilient to the possibility of an existing PR having been merged or closed while the controller was down/restarting, as they do not even check if it is open, merged, closed, etc. (It is a subsequent `git-wait-for-pr` step's job to worry about that.)

### DIFF
--- a/internal/directives/git_pr_opener.go
+++ b/internal/directives/git_pr_opener.go
@@ -72,6 +72,22 @@ func (g *gitPROpener) runPromotionStep(
 	stepCtx *PromotionStepContext,
 	cfg GitOpenPRConfig,
 ) (PromotionStepResult, error) {
+	// Short-circuit if shared state has output from a previous execution of this
+	// step that contains a PR number.
+	prNumber, err := g.getPRNumber(stepCtx, stepCtx.SharedState)
+	if err != nil {
+		return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored},
+			fmt.Errorf("error getting PR number from shared state: %w", err)
+	}
+	if prNumber != -1 {
+		return PromotionStepResult{
+			Status: kargoapi.PromotionPhaseSucceeded,
+			Output: map[string]any{
+				prNumberKey: prNumber,
+			},
+		}, nil
+	}
+
 	sourceBranch, err := g.getSourceBranch(stepCtx.SharedState, cfg)
 	if err != nil {
 		return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored},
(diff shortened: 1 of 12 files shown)


internal/directives/git_pr_opener.go
```diff
@@ -198,7 +220,49 @@ func (g *gitPROpener) runPromotionStep(
 	}, nil
 }
 
-func getSourceBranch(sharedState State, cfg GitOpenPRConfig) (string, error) {
+// getPRNumber checks shared state for output from a previous execution of this
+// step. If any is found and it contains a PR number, that number is returned.
+// 0 is returned if no PR number is found in the shared state. An error is
```
hiddeco, line 225: Nit: I think returning `-1` would be a more explicit "does not exist".

krancour: I waffled on that myself. Since you're throwing your weight behind that option, I'll happily change it.

krancour: Fixed in 0a593fd914852c4a26ee00cb02ed916624dbedeb

internal/directives/git_pr_opener.go
```diff
@@ -298,15 +367,38 @@ func mustOpenPR(
...
 	}
 	if len(prs) == 0 {
-		return true, nil
+		return nil, nil
 	}
+	// If promotion names are incorporated into PR source branches, it's highly
+	// unlikely that we would have found more than one PR matching the search
+	// criteria. Accounting for the possibility of users specifying their own
+	// source branch names using an expression, there is somewhat more of a
+	// possibility of multiple PRs being found. We sort descending by creation
```
hiddeco, line 379: Is this true? As far as I know, on GitHub, you can only have one open pull request (which we are querying for) for a target/source combination at a time. This makes sense because they otherwise would be literal duplicates from the perspective of the Git patch.

krancour: You're right that only two identical PRs can be _open_. But I tested and you can open one, close it, and then open a new one that's identical.

hiddeco: Yes, but the default behavior of `ListPullRequests` is to only query for _open_ pull requests, which makes me think that the result will always equal one?

krancour: > but the default behavior of ListPullRequests is to only query for open pull requests

Ah... I'm glad you called this out.

I want to find open _and_ closed PRs.

hiddeco: In that case, wouldn't it be better to first sort by state (open > closed) and then by date (in case all are closed)?

krancour: 🤷‍♂️ Maybe? There are _so many_ edge cases lurking here.

You could open PR 1, close it, open an identical 2, close it, then re-open 1.

With my sort, we'd incorrectly get 2 as the top result when 1 is the one we ought to be tracking. Mine cannot contend with this edge case.

By your suggestion, we'd get 1 as the top result while it remains open, but once it's closed/merged, the same sort would give us 2 as the top result.

Maybe the right order is open > merged > closed?

krancour: Sort order is fixed specifically in 19f3e7d44c8c7967088aaba09361ced0bd84b88d

hiddeco (approved):

Issue #2775, opened 2024-10-17
UI should remember zoom preference of pipeline


jessesuen:
# Description

The pipeline zoom in/out feature preference which is not remembered, making it less useful than it could be. 

Another thought is that we could scale the pipeline to fit the screen size by default. Unless they click zoom in/out, which would store their preference of zoom.

# Screenshots

https://github.com/user-attachments/assets/bb45a7c8-74fa-4169-aff4-15766d231e46

# Version

```
v0.9
```

PR #2904, opened 2024-11-06
feat(ui): pipeline UX improvements

Marvin9:
- default fit the whole pipeline into view
- drag like maps and zoom in/out using mouse wheel
- [x] remember zoom preference - https://github.com/akuity/kargo/issues/2775
- [x] Add screenshot/video

https://github.com/user-attachments/assets/14afcf3f-1686-4f36-bf70-1c06338a4cff


jessesuen (approved):
Not a code review, but approval based on behavior demonstrated in video

krancour (approved):
🔥 This is amazing and I also approve based on a review of the functionality.

idk if @rpelczar may wish to review implementation details as well since this appears non-trivial.

ui/src/utils/use-temporary.ts
```diff
@@ -0,0 +1,18 @@
...
+export const useTemporaryBoolean = (timeout = 500 /* milliseconds */) => {
+  const [init, setInit] = useState(false);
+
+  useEffect(() => {
+    if (init) {
+      setTimeout(() => {
+        setInit(false);
+      }, timeout);
+    }
+  }, [init]);
```
rpelczar, line 15: Don't forget clear interval in the cleanup function.

rpelczar (approved):
Looks amazing! I left one comment with one small thing to improve.

rpelczar:
I also think the save button is overkill. I like it, but not sure if it is as useful as the reset button.
This is my personal opinion, so you don't have to do anything with that.

Marvin9:
I think that makes sense @rpelczar, we have reset view anyways so I have removed explicit save view preference and it autosaves now. Let me know if you think otherwise @jessesuen @krancour

02strich:
Btw, this breaks kargo on firefox due to https://developer.mozilla.org/en-US/docs/Web/API/Element/computedStyleMap not being supported there today

hiddeco:
@02strich, thank you for bringing this to our attention. We've noticed this issue internally as well, and will work on a solution. Appreciate the report! 🙇

PR #2905, opened 2024-11-07
refactor: rewrite Stage reconciler

hiddeco:
This refactors the logic of the "regular" Stage reconciler into smaller chunks, with more extensive test coverage.

In addition, it adds conditions to the Stage to allow a better assessment of the state it is in over the previous `Phase`:

- `Reconciling`: Set for as long a reconciliation is in progress on the Stage, and e.g. new changes can be expected.
- `Ready`: Set to `True` when the Stage is in a "steady" state.
- `Promoting`: When `True`, new Freight is actively being Promoted to the Stage.
- `Healthy`: To indicate the health of the Stage based on the result of the health checks, compliments `.status.health`.
- `Verified`: To indicate if the Stage (Freight) has passed verification.


internal/controller/stages/regular_stages.go
```diff
@@ -0,0 +1,1981 @@
...
+					kargoClient: kargoMgr.GetClient(),
+				},
+			),
+		); err != nil {
+			return fmt.Errorf("unable to watch AnalysisRuns: %w", err)
+		}
+	}
+
+	logging.LoggerFromContext(ctx).Info(
+		"Initialized control flow Stage reconciler",
```
krancour, line 277: Should this be the regular Stage reconciler?

internal/controller/stages/regular_stages.go
```diff
@@ -0,0 +1,1981 @@
...
+		newStatus, err = subR.reconcile()
+
+		// Summarize the conditions after each sub-reconciler to ensure that
+		// we have a consistent view of the Stage status.
+		summarizeConditions(stage, &newStatus, err)
+
+		// If an error occurred during the sub-reconciler, then we should
+		// return the error and request an immediate requeue.
+		if err != nil {
+			return newStatus, false, err
```
krancour, line 435: I'm pretty sure you've left requeue as false because a non-nil error will cause a requeue anyway, but to the casual reader, this may appear to contradict the comment above. wdyt about either setting it to true _or_ adding a comment that a non-nil error will trigger a requeue?

internal/controller/stages/regular_stages.go
```diff
@@ -0,0 +1,1981 @@
...
+	// highest priority Promotion is in a terminal phase, then we must have
+	// finished promoting.
+	if currentPromo != nil && (currentPromo.Name != highestPrioPromo.Name || highestPrioPromo.Status.Phase.IsTerminal()) {
+		// Gather information about the Promotions that have terminated since the
+		// last reconciliation.
+		var newPromotions []kargoapi.PromotionReference
+		for _, promo := range promotions.Items {
+			// Update the conditions to reflect that we are no longer promoting.
+			conditions.Delete(&newStatus, kargoapi.ConditionTypePromoting)
+			newStatus.CurrentPromotion = nil
```
krancour, line 540: Should this be done just once before the loop starts?

internal/controller/stages/regular_stages.go
```diff
@@ -0,0 +1,1981 @@
...
+	// for it to complete regardless of health state to ensure we capture the
+	// results.
+	if curFreight := stage.Status.FreightHistory.Current(); curFreight != nil {
+		if curFreight.HasNonTerminalVerification() {
+			logger.Debug(
+				"current Freight has a non-terminal verification: " +
+					"wait for it to complete before allowing new promotions to start",
+			)
+			conditions.Delete(&newStatus, kargoapi.ConditionTypePromoting)
+			return newStatus, hasNonTerminalPromotions, nil
```
krancour, line 621: Ref: [this comment](https://github.com/akuity/kargo/pull/2905/files#r1842750619)

internal/controller/stages/regular_stages.go
```diff
@@ -0,0 +1,1981 @@
...
+			err,
+		)
+	}
+	return nil
+}
+
+// summarizeConditions summarizes the conditions of the given Stage. It sets the
+// Ready condition based on the Promoting, Healthy and Verified conditions.
+// If there is an error, the Ready condition is set to False until the error is
+// resolved.
```
krancour, line 1823: I'm wondering if "ready" is the ideal condition to summarize the other conditions here. We've used ready for Warehouses and Projects, and for those cases, I think it makes sense as it implies "functional."

For a stage, I feel as if it's not obvious what "ready" means. Ready for what?

I am wondering whether something like steady or stable might better convey that all other conditions are nominal.

hiddeco: The reason I picked "Ready" is because it is kind of standardized, and recognized by certain tools when they automatically try to determine if a resource is "in motion" or not (and thus, if they should e.g. wait for it to reach a stable state).

In this context, I see "ready" as "not working on anything new or worth waiting on".

internal/controller/stages/regular_stages.go
hiddeco: If we delete it at the beginning by default, the transition timestamp on the condition will continue to drift, which is why they are removed for individual cases.

internal/controller/stages/regular_stages.go
krancour: Language-wise, I have difficulty interpreting "ready" as meaning "idle", _buuuut_ the point about it being a common convention understood by a lot of tooling is a good one and well-taken. Consider me sufficiently convinced.

krancour:
One thing that tripped me up when reading through the changes, although this is not necessarily a _new_ issue:

The method names `verifyStageFreight()` and `verifyFreightForStage()` are quite similar. The former has a good comment explaining what it does. The latter could probably stand to either be renamed or commented in a way that makes it more clear that it's about _marking_ Freight as verified as opposed to _performing_ verification.

internal/controller/stages/regular_stages.go
```diff
@@ -0,0 +1,1981 @@
...
+		{
+			name: "syncing Promotions",
+			reconcile: func() (kargoapi.StageStatus, error) {
+				status, hasPendingPromotions, err := r.syncPromotions(ctx, stage)
+				if err != nil {
+					err = fmt.Errorf("failed to sync Promotions: %w", err)
+				}
+				if status.CurrentPromotion == nil && hasPendingPromotions {
+					requestRequeue = true
+				}
```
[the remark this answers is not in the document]
krancour: Is this actually resolved?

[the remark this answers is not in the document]
krancour: You had marked it as such, but I couldn't see what, if anything changed that fixed it. Or was it not actually an issue?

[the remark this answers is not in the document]
hiddeco: The bit that "undoes" this is: https://github.com/akuity/kargo/blob/9414bb72f1cc2ae2f25eba03a342a874456b925e/internal/controller/stages/regular_stages.go#L400-L406

The reason I put it there is because it will get you a more accurate result, as it knows about the most recent verification changes at that time.

hiddeco:
> The latter could probably stand to either be renamed or commented in a way that makes it more clear that it's about marking Freight as verified as opposed to performing verification.

I had this same concern, but at the time could not figure out what name would be better. "Marking" however is a good suggestion, will make the change.

internal/controller/stages/regular_stages.go
```diff
@@ -0,0 +1,2030 @@
...
+		}
+	}
+	return newStatus, err
+}
+
+// markFreightVerifiedForStage marks the Freight that is associated with the
+// Stage as verified. If the Freight has already been verified, then no action
+// is taken.
+//
+// Once the Freight has been verified, the Stage can proceed
```
krancour, line 986: Is this line a mistake?

krancour: Non-blocking, so pre-emptively LGTM'ing.

krancour (approved):
🔥 I am massively appreciative of these improvements. 🙏

PR #2906, opened 2024-11-07
feat: Implement livenessProbe variables and readinessProbe variables in helm chart

stefanandres:
It's sometimes needed to configure certain timeouts. This PR makes all the settings available as helm variables.


krancour:
I'm supportive of this change, but there are a few issues that need to be resolved:

- This only adds the options to configure liveness and readiness probes to the API server and the Dex server. For consistency (and because it's useful), the same options should also be added to the controller, management controller, and webhook server.

- Inline documentation/metadata is required for the new fields in `values.yaml`. This isn't because I assume users require much instruction on liveness or readiness probes; rather it is because we generate the chart's `README.md` from metadata in the `values.yaml` file. Without proper metadata having been applied to these new fields, even the _existence_ of these options will not be surfaced in the `README.md`.

- A few other issues, that I will call out inline in the diffs.

charts/kargo/templates/api/deployment.yaml
```diff
@@ -85,8 +85,11 @@ spec:
             - name: h2c
               containerPort: 8080
               protocol: TCP
+
 {{- if .Values.api.probes.enabled }}
-          livenessProbe:
+          {{- with .Values.api.probes.livenessProbe }}
+          livenessProbe: {{- toYaml . | nindent 12 }}
+          {{- end }}
```
krancour, line 92: The following 8 lines are also part of the probe.

charts/kargo/templates/api/deployment.yaml
```diff
@@ -95,8 +98,9 @@ spec:
                 - -tls
                 - -tls-no-verify
 {{- end }}
-            initialDelaySeconds: 10
-          readinessProbe:
+          {{- with .Values.api.probes.readinessProbe }}
+          readinessProbe: {{- toYaml . | nindent 12 }}
+          {{- end }}
```
krancour, line 103: The following 8 lines are also part of the probe.

stefanandres:
Thanks for the review, I'll see that I do the changes. 

I would have changed other deployments if they'd already used probes, but can add placeholder for them as well.

charts/kargo/templates/api/deployment.yaml
```diff
@@ -85,8 +85,11 @@ spec:
             - name: h2c
               containerPort: 8080
               protocol: TCP
+
 {{- if .Values.api.probes.enabled }}
```
krancour, line 89: I think with the probes fully configurable, this option becomes redundant and should be removed from the template and the `values.yaml` file.

It is a breaking change, but it's a low-impact breaking change:

- Exists mainly to support local development.
- Anyone who may have disabled the probes already (in non-local dev envs), if this breaking change were to catch them by surprise, they'd be _gaining_ new probes rather than _losing_ anything. Whatever disruption that might cause should be easily identified and corrected.

I will of course mention a change such as this prominently in the release notes.

charts/kargo/templates/dex-server/deployment.yaml
```diff
@@ -74,25 +74,13 @@ spec:
         resources:
           {{- toYaml .Values.api.oidc.dex.resources | nindent 10 }}
         {{- if .Values.api.oidc.dex.probes.enabled }}
```
krancour, line 76: Same as [this comment](https://github.com/akuity/kargo/pull/2906/files#r1832831666).

charts/kargo/values.yaml
```diff
@@ -144,6 +144,11 @@ api:
   probes:
     ## @param api.probes.enabled Whether liveness and readiness probes should be included in the API server deployment. It is sometimes advantageous to disable these during local development.
     enabled: true
+    livenessProbe:
+      initialDelaySeconds: 10
+
```
krancour, line 149: Nit: Killing this blank line would visually make it more obvious that `readinessProbe` below is a component of this block.

krancour:
> rather it is because we generate the chart's README.md from metadata in the values.yaml file

Running `make codegen-docs` will bring the chart's `README.md` up to date after any modifications to `values.yaml`.

charts/kargo/templates/api/deployment.yaml
krancour: It occurs to me that removing this option will require some slight modifications to `hack/tilt/values.dev.yaml`. You'll just have to nil out the probes for a couple of components.

stefanandres: Alright, I'll just remove the tls-variable for the probe then

From what I see, the probes are already disabled in tilt?

* https://github.com/akuity/kargo/blob/main/hack/tilt/values.dev.yaml#L8
* https://github.com/akuity/kargo/blob/main/hack/tilt/values.dev.yaml#L22

stefanandres: Just to mention it here: I'd personally leave the tls variable in the template. Removing it adds manual toil to deployments (like ours) because you'll be confused why the probe is failing when TLS is disabled.
in our case, we don't use TLS here because our ingress is already doing TLS

stefanandres:
> > rather it is because we generate the chart's README.md from metadata in the values.yaml file
> 
> Running `make codegen-docs` will bring the chart's `README.md` up to date after any modifications to `values.yaml`.

This fails for me, how do I fix that?

```
❯ make codegen-docs
npm install -g @bitnami/readme-generator-for-helm
(node:9158) ExperimentalWarning: CommonJS module /opt/homebrew/lib/node_modules/npm/node_modules/debug/src/node.js is loading ES Module /opt/homebrew/lib/node_modules/npm/node_modules/supports-color/index.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported

added 19 packages in 3s

2 packages are looking for funding
  run `npm fund` for details
npm notice
npm notice New patch version of npm available! 10.9.0 -> 10.9.1
npm notice Changelog: https://github.com/npm/cli/releases/tag/v10.9.1
npm notice To update run: npm install -g npm@10.9.1
npm notice
bash hack/helm-docs/helm-docs.sh
Skipping check for kubeconfigSecrets
INFO: Checking missing metadata...


######
The following errors must be fixed before proceeding
######


ERROR: Missing metadata for key: api.probes.livenessProbe.initialDelaySeconds
ERROR: Missing metadata for key: api.probes.livenessProbe.exec.command
ERROR: Missing metadata for key: api.probes.readinessProbe.initialDelaySeconds
ERROR: Missing metadata for key: api.probes.readinessProbe.exec.command
ERROR: Missing metadata for key: api.oidc.dex.probes.livenessProbe.httpGet.path
ERROR: Missing metadata for key: api.oidc.dex.probes.livenessProbe.httpGet.port
ERROR: Missing metadata for key: api.oidc.dex.probes.livenessProbe.initialDelaySeconds
ERROR: Missing metadata for key: api.oidc.dex.probes.livenessProbe.periodSeconds
ERROR: Missing metadata for key: api.oidc.dex.probes.livenessProbe.timeoutSeconds
ERROR: Missing metadata for key: api.oidc.dex.probes.livenessProbe.successThreshold
ERROR: Missing metadata for key: api.oidc.dex.probes.livenessProbe.failureThreshold
ERROR: Missing metadata for key: api.oidc.dex.probes.readinessProbe.httpGet.path
ERROR: Missing metadata for key: api.oidc.dex.probes.readinessProbe.httpGet.port
ERROR: Missing metadata for key: api.oidc.dex.probes.readinessProbe.initialDelaySeconds
ERROR: Missing metadata for key: api.oidc.dex.probes.readinessProbe.periodSeconds
ERROR: Missing metadata for key: api.oidc.dex.probes.readinessProbe.timeoutSeconds
ERROR: Missing metadata for key: api.oidc.dex.probes.readinessProbe.successThreshold
ERROR: Missing metadata for key: api.oidc.dex.probes.readinessProbe.failureThreshold
/opt/homebrew/lib/node_modules/@bitnami/readme-generator-for-helm/lib/checker.js:93
    throw new Error('ERROR: Wrong metadata!');
    ^

Error: ERROR: Wrong metadata!
    at checkKeys (/opt/homebrew/lib/node_modules/@bitnami/readme-generator-for-helm/lib/checker.js:93:11)
    at getParsedMetadata (/opt/homebrew/lib/node_modules/@bitnami/readme-generator-for-helm/index.js:26:3)
    at runReadmeGenerator (/opt/homebrew/lib/node_modules/@bitnami/readme-generator-for-helm/index.js:52:28)
    at Object.<anonymous> (/opt/homebrew/lib/node_modules/@bitnami/readme-generator-for-helm/bin/index.js:22:1)
    at Module._compile (node:internal/modules/cjs/loader:1546:14)
    at Object..js (node:internal/modules/cjs/loader:1698:10)
    at Module.load (node:internal/modules/cjs/loader:1303:32)
    at Function._load (node:internal/modules/cjs/loader:1117:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:218:24)

Node.js v23.3.0
make: *** [codegen-docs] Error 1
```

charts/kargo/templates/api/deployment.yaml
krancour: > In tilt the probes are already disabled, so we shouldn't need to override the exec there

This suggests some misunderstanding.

> I think with the probes fully configurable, this option becomes redundant and should be removed from the template and the values.yaml file.

"This option," here referred to `api.probes.enabled`. There is no sense in having an `enabled` boolean when having or not having a liveness probe can simply depend on whether `api.probes.livenessProbe` is defined or not. Same can be said for the readiness probes.

Having said that, the following may change my line of thought...

> Just to mention it here: I'd personally leave the tls variable in the template. Removing it adds manual toil to deployments (like ours) because you'll be confused why the probe is failing when TLS is disabled.

That wasn't even on my radar when I initially reviewed, but you're right. I believe this exposes the fact that it's probably not so ideal to put the definitions of the probes _fully_ within the control of the user. It's not _just_ about TLS. It altogether eliminates any (current or future) possibility of probes being configured conditionally based on based on _other_ settings throughout the chart. TLS being enabled or not is just one such example.

This drives me toward thinking that we ought not let users define `livenessProbe` and `readinessProbe` in their entirety, but should, instead, surface individual options for things like `initialDelaySeconds`, `periodSeconds`, `timeoutSeconds`, etc., etc.

stefanandres: > This drives me toward thinking that we ought not let users define livenessProbe and readinessProbe in their entirety, but should, instead, surface individual options for things like initialDelaySeconds, periodSeconds, timeoutSeconds, etc., etc.

I'd argue to go with my first implementation. With this, all options except the exec.command variables would be setable.

krancour: > I'd argue to go with my first implementation. With this, all options except the exec.command variables would be setable.

Have a look at something like https://github.com/akuity/kargo/pull/3041 where we're looking to change how one of the probes is implemented.

Changes such as this would become much more difficult if we leave things _too_ configurable.

krancour:
Closing this PR since it seems to have been abandoned. @stefanandres you are welcome to re-open it at any time if you want to finish it up.

Issue #2883, opened 2024-10-30
docs: update kargo UI snapshot in overview


fykaa:
https://docs.kargo.io/
Kargo's user interface in the overview documentation is Old.

krancour:
Can you provide some more detail here? It's not clear to me what the issue meant to address.

PR #2914, opened 2024-11-11
docs: updated kargo ui snapshot

fykaa:
fixes https://github.com/akuity/kargo/issues/2883

before:
image

after:
P5esNo604TwM3IYa7vyksyaJQ


krancour:
Where are all the little Argo logos coming from? Is there a reason that these should be here?

krancour:
@fykaa can we put some closure on this PR please?

krancour (approved):

PR #2918, opened 2024-11-12
fix(ui): pipeline view issues

Marvin9:
fixes

- replace unsupported `computedStyleMap` in firefox with widely supported method - https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle
- unblock stage node click

### DIFF
--- a/ui/src/features/project/pipelines/utils/use-pipelines-infinite-canvas.ts
+++ b/ui/src/features/project/pipelines/utils/use-pipelines-infinite-canvas.ts
@@ -239,6 +239,7 @@ export const usePipelinesInfiniteCanvas = (conf: pipelineInfiniteCanvasHook) => 
 
     const onCanvasMouseDown = (e: MouseEvent) => {
       // block stage node click
+      // TODO: rewrite HTML structure and refactor this along for better control
       let element = e.target as HTMLElement;
       while (element) {
         if (element?.classList?.contains(stageNodeClassName)) {
--- a/ui/src/features/project/pipelines/utils/use-pipelines-infinite-canvas.ts
+++ b/ui/src/features/project/pipelines/utils/use-pipelines-infinite-canvas.ts
@@ -26,6 +26,14 @@ export const usePipelineViewPrefHook = (project: string, opts?: { onSet?(): void
   return [state, setState] as const;
 };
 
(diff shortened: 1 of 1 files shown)


ui/src/features/project/pipelines/utils/use-pipelines-infinite-canvas.ts
```diff
@@ -240,16 +236,26 @@ export const usePipelinesInfiniteCanvas = (conf: pipelineInfiniteCanvasHook) =>
...
 
     const onCanvasMouseDown = (e: MouseEvent) => {
+      // block stage node click
+      let element = e.target as HTMLElement;
+      while (element) {
+        if (element?.classList?.contains(stageNodeClassName)) {
+          return;
+        }
+        element = element.parentNode as HTMLElement;
+      }
```
rpelczar, line 246: Can't we just handle canvas instead of blocking movement when the stage is detected? It might stop working if we change html structure.

rpelczar (changes requested):

ui/src/features/project/pipelines/utils/use-pipelines-infinite-canvas.ts
Marvin9: Could you expand on it please? I didn't get handle canvas instead blocking movement. stage nodes are in canvas, any mousedown event in stage node will propagate to canvas too and canvas mouse down is interpreted as drag & move.

rpelczar: I mean, you can detect when the user clicks the background instead of the whole pipeline. You can add a new div, place it behind the pipeline, and set an absolute position.

Marvin9: that is good direction but can we do it in different PR, this hooks need bit refactor anyways as there are lots of moving parts.

rpelczar (approved):

ui/src/features/project/pipelines/utils/use-pipelines-infinite-canvas.ts
rpelczar: Can you add a todo comment?

PR #2925, opened 2024-11-14
feat(controller): add functions to EL for finding artifacts in promo's freight collection

krancour:
Before:

```yaml
steps:
- uses: git-clone
  config:
    repoURL: ${{ vars.gitRepo }}
    checkout:
    - fromFreight: true                  # Magic!!!
      fromOrigin:                        # Verbose
        kind: Warehouse
        name: base-manifests
      path: ./base
    - fromFreight: true                  # Magic!!!
      fromOrigin:                        # Verbose
        kind: Warehouse
        name: overlay
      path: ./overlay
# ...
```

After:

```yaml
steps:
- uses: git-clone
  config:
    repoURL: ${{ vars.gitRepo }}
    checkout:
    - commit: ${{ commitFrom(vars.gitRepo, warehouse('base-manifests')).ID }} # Clear and succinct
      path: ./base
    - commit: ${{ commitFrom(vars.gitRepo, warehouse('overlay')).ID }} # Clear and succinct
      path: ./overlay
# ...
```

Cases not requiring origin disambiguation are even more straightforward:

```yaml
steps:
- uses: git-clone
  config:
    repoURL: ${{ vars.gitRepo }}
    checkout:
    - commit: ${{ commitFrom(vars.gitRepo).ID }} # Clear and succinct
      path: ./src
```

Similar functions are defined for accessing images and charts from the Promo's Freight collection.

See doc updates for more usage details and examples.

1b80bf06a87cdda7fdb71ef535424f46a43c413e is the core of this PR.

Most other commits focus on a single promotion step each:

* Update schema with new fields that will be more natural replacements for a handful of existing fields that were workaround for EL not previously existing.
* Update step implementation and tests accordingly.
* Update step docs: Document new fields, deprecate others, update examples

A large chunk of this PR is new tests on schemas that I touched.

_Short term_ this makes schemas slightly more complex and introduces additional conditionals into step implementations, but once we're able to fully remove the more "magical" fields and the logic that supports them, the schemas and step implementations will both be simplified beyond the state they were in _prior_ to this PR. i.e. We get to delete lots of stuff in 1.2.

This branch of my examples is updated to take full advantage of these improvements: https://github.com/krancour/kargo-examples/tree/for-v1.1.0


internal/directives/argocd_updater.go
```diff
@@ -815,28 +815,40 @@ func (a *argocdUpdater) buildKustomizeImagesForAppSource(
...
-		)
-		if err != nil {
-			return nil,
-				fmt.Errorf("error finding image from repo %q: %w", imageUpdate.RepoURL, err)
+		var digest, tag string
+		if imageUpdate.Digest != "" {
+			digest = imageUpdate.Digest
+		} else if imageUpdate.Tag != "" {
+			tag = imageUpdate.Tag
+		} else {
```
hiddeco, line 823: Looks like this might become more readable if we would make it a switch?

hiddeco (approved):
This is really awesome, excellent job @krancour 🙇 🔥

PR #2930, opened 2024-11-15
fix(docs): consistent title case

### LINKED ISSUE #2902: fix: Use title case for everything in the left sidebar/doctree
Here's how it looks now:

Screenshot 2024-11-06 at 9 53 54 AM

I the cases where the title is simply derived from the filename, we don't end up with title case. In cases where it's been explicitly overridden, we _do_ have title case.

Let's consistently use title case for all links in the doc tree.

fykaa:
closes #2902

The current state of the docs adopts the label for each item in the following order of precedence

1. From the `sidebar_label` field in the frontmatter of the `.md` file (if explicitly provided)
2. From the first `# heading` in the `.md` file (if `sidebar_label` is absent)
3. From the filename of the `.md` file (if neither `sidebar_label` nor a `# heading` is present)

Given this behavior I have tried to address the issue by updating the markdown files to explicitly include titlecase in the `sidebar_label` field where it was missing. 

Do we need to add further changes, like automating this process or handling additional edge cases?
Also, should we add monotype texts for kargo resources?

### DIFF
--- a/docs/docs/40-contributor-guide/index.md
+++ b/docs/docs/40-contributor-guide/index.md
@@ -11,6 +11,6 @@ Kargo project.
 
 This guide is decomposed into the following, high-level topics:
 
-* [Hacking on Kargo](./10-hacking-on-kargo.md)
-* [Signing commits](./20-signing-commits.md)
-* [Code of conduct](./30-code-of-conduct.md)
+* [Hacking On Kargo](./10-hacking-on-kargo.md)
+* [Signing Commits](./20-signing-commits.md)
+* [Code Of Conduct](./30-code-of-conduct.md)
--- a/docs/docs/40-contributor-guide/index.md
+++ b/docs/docs/40-contributor-guide/index.md
@@ -1,6 +1,6 @@
 ---
 description: A comprehensive introduction for developers who are looking to get involved with contributing directly to the Kargo project
-sidebar_label: Contributor guide
+sidebar_label: Contributor Guide
 ---
 
 # Kargo Contributor Guide
--- a/docs/docs/40-contributor-guide/index.md
+++ b/docs/docs/40-contributor-guide/index.md
@@ -11,6 +11,6 @@ Kargo project.
 
 This guide is decomposed into the following, high-level topics:
 
-* [Hacking On Kargo](./10-hacking-on-kargo.md)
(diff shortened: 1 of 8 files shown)


docs/docs/40-contributor-guide/index.md
```diff
@@ -11,6 +11,6 @@ Kargo project.
 
 This guide is decomposed into the following, high-level topics:
 
-* [Hacking on Kargo](./10-hacking-on-kargo.md)
-* [Signing commits](./20-signing-commits.md)
-* [Code of conduct](./30-code-of-conduct.md)
+* [Hacking On Kargo](./10-hacking-on-kargo.md)
+* [Signing Commits](./20-signing-commits.md)
+* [Code Of Conduct](./30-code-of-conduct.md)
```
krancour, line 16:
```suggestion
* [Hacking on Kargo](./10-hacking-on-kargo.md)
* [Signing Commits](./20-signing-commits.md)
* [Code of Conduct](./30-code-of-conduct.md)
```

krancour:
> Do we need to add further changes, like automating this process or handling additional edge cases?

Adding new pages or changing a page's title is an infrequent enough thing that the payoff for automating conversion to title case isn't worth the effort.

> Also, should we add monotype texts for kargo resources?

Good question, and frequently a difficult one...

Even in titles, I use monospaced text if I am _unquestionably_ referring to a type, variable, command, etc. There are many cases that aren't so black and white. Take "Working with Freight" as an example. Is this "Freight" as a _concept_ (a bundle of artifacts) or is it the `Freight` type? If you read the page, maybe it's actually a little bit of both. In this sort of case, it's hard to decide what's right and I would err on the side of whatever _looks_ nicer. imho, for titles, if in doubt, don't monospace.

This comes up in a lot of other contexts, btw. Two great examples are "string" and "namespace," both of which arguably _could_ be monospaced nearly everywhere you see them. But they're also frequently used words and _well-understood concepts_. If we go overboard monospacing those everywhere they occur, it needlessly calls attention to insignificant words which is more likely to annoy readers than to improve their understanding of anything.

In the end, there is no hard and fast rule here and in many cases, you just need to use your best judgement.

Here, I would personally err on the side of no.

krancour:
There are still some issues:

Screenshot 2024-11-15 at 11 11 40 AM

PR #2932, opened 2024-11-15
fix(directives): log error to surface reason for ArgoCD update

hiddeco:
Found this while trying to debug an issue with the `argocd-update` step.

### DIFF
--- a/internal/directives/argocd_updater.go
+++ b/internal/directives/argocd_updater.go
@@ -265,6 +265,12 @@ func (a *argocdUpdater) runPromotionStep(
 			continue
 		}
 
+		// Log the error, as it contains information about why we need to
+		// perform an update.
+		if err != nil {
+			logger.Debug(err.Error())
+		}
+
 		// Perform the update.
 		if err = a.syncApplicationFn(
 			ctx,


internal/directives/argocd_updater.go
```diff
@@ -265,6 +265,12 @@ func (a *argocdUpdater) runPromotionStep(
 			continue
 		}
 
+		// Log the error, as it contains information about why we need to
+		// perform an update.
+		if err != nil {
+			logger.Debug(err.Error())
+		}
```
krancour, line 272: I'm somewhat questioning whether an error explains why `mustUpdate` is `true`.

If successfully we've made the determination that we must update, wouldn't `err` be `nil`?

krancour (changes requested):
Not necessarily requesting a change. Just trying to block merge until my question is answered.

internal/directives/argocd_updater.go
krancour: I see. I hadn't quite realized we were treating mismatches as a sort of error and not _just_ returning true. All set.

krancour (approved):

Issue #2834, opened 2024-10-24
docs: need to mention need for manual rolebindings in global credential namespaces


krancour:
This is a requirement that was introduced in v1.0.0 due to us no longer allowing controllers cluster-wide read access to Secrets. The docs do not currently mention this and should.

PR #2934, opened 2024-11-15
docs: manual rolebindings for global namespaces

fykaa:
closes https://github.com/akuity/kargo/issues/2834

### DIFF
--- a/docs/docs/30-how-to-guides/20-managing-credentials.md
+++ b/docs/docs/30-how-to-guides/20-managing-credentials.md
@@ -5,7 +5,7 @@ sidebar_label: Managing credentials
 
 # Managing Credentials
 
-To manage the progression of freight from stage to stage, Kargo will often
+To manage the progression of `Freight` from `Stage` to `Stage`, Kargo will often
 require read/write permissions on private GitOps repositories and read-only
 permissions on private container image and/or Helm chart repositories.
 
@@ -105,6 +105,35 @@ designating one or more `Namespace`s as homes for "global" credentials using the
 Refer to
 [the advanced section of the installation guide](./10-installing-kargo.md#advanced-installation)
 for more details.
+
+While doing this you must ensure that the Kargo controller has the appropriate
+permissions to read `Secret` resources in these namespaces. This is necessary because,
+as of Kargo v1.0.0, the controller does not have cluster-wide access to `Secret`s. Permissions
+are granted dynamically on a per-Project basis as `Projects` are created.
+
+You can do so by creating your own `RoleBinding`s to permit the Kargo
+ controller(s) to read Secret(s) from each of those namespaces:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: kargo-controller-read-secrets-binding
+  namespace: kargo
+subjects:
+- kind: ServiceAccount
+  name: kargo-controller
+  namespace: kargo
+roleRef:
+  kind: Role
+  name: kargo-controller-read-secrets
+  apiGroup: rbac.authorization.k8s.io
+```
+
+OR
+
+(Not recommended) Set `controller.serviceAccount.clusterWideSecretReadingEnabled` 
+setting to `true` in Kargo's Helm chart.
 
 :::note
 Any matching credentials (exact match _or_ pattern match) found in a project's
--- a/docs/docs/30-how-to-guides/20-managing-credentials.md
+++ b/docs/docs/30-how-to-guides/20-managing-credentials.md
@@ -106,29 +106,38 @@ Refer to
 [the advanced section of the installation guide](./10-installing-kargo.md#advanced-installation)
 for more details.
 
-While doing this you must ensure that the Kargo controller has the appropriate
-permissions to read `Secret` resources in these namespaces. This is necessary because,
-as of Kargo v1.0.0, the controller does not have cluster-wide access to `Secret`s. Permissions
-are granted dynamically on a per-Project basis as `Projects` are created.
-
-You can do so by creating your own `RoleBinding`s to permit the Kargo
- controller(s) to read Secret(s) from each of those namespaces:
+:::note
+Operators must manually ensure Kargo controllers receive read-only access
+to `Secret`s in the designated namespaces. For example, if `kargo-global-creds`
+is designated as a global credentials namespace, the following `RoleBinding`
+should be created within that `Namespace`:
 
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: kargo-controller-read-secrets
-  namespace: kargo-global-creds
+    name: kargo-controller-read-secrets
+    namespace: kargo-global-creds
 subjects:
-- kind: ServiceAccount
-  name: kargo-controller
-  namespace: kargo
+    - kind: ServiceAccount
+      name: kargo-controller
+      namespace: kargo
 roleRef:
-  kind: Role
-  name: kargo-controller-read-secrets
-  apiGroup: rbac.authorization.k8s.io
+    kind: Role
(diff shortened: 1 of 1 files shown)


docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -5,7 +5,7 @@ sidebar_label: Managing credentials
 
 # Managing Credentials
 
-To manage the progression of freight from stage to stage, Kargo will often
+To manage the progression of `Freight` from `Stage` to `Stage`, Kargo will often
```
krancour, line 8: I agree with the changes to uppercase here, but probably wouldn't monospace here.

```suggestion
To manage the progression of Freight from Stage to Stage, Kargo will often
```

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -106,6 +106,35 @@ Refer to
...
+are granted dynamically on a per-Project basis as `Projects` are created.
+
+You can do so by creating your own `RoleBinding`s to permit the Kargo
+ controller(s) to read Secret(s) from each of those namespaces:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: kargo-controller-read-secrets-binding
```
krancour, line 121:
```suggestion
  name: kargo-controller-read-secrets
```

In general, let's avoid unnecessarily including information about a resource's type in its name.

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -106,6 +106,41 @@ Refer to
...
+
+You can do so by creating your own `RoleBinding`s to permit the Kargo
+ controller(s) to read Secret(s) from each of those namespaces:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: kargo-controller-read-secrets
+  namespace: kargo
```
krancour, line 122: It is probably not ideal to use the `kargo` namespace as a home for shared creds because:

* Controllers would gain access to Kargo's own _internal_ Secrets
* Users authorized to create global creds would require r/w on Secrets in Kargo's own namespace, which would also give them access to Kargo's own _internal_ Secrets

I don't feel like we need to caution too explicitly against this, but for sure, our example shouldn't imply that is viable.

```suggestion
  namespace: kargo-global-creds
```

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -106,6 +106,41 @@ Refer to
...
+OR
+
+(Not recommended) Set `controller.serviceAccount.clusterWideSecretReadingEnabled` 
+setting to `true` in Kargo's Helm chart.
+
+:::warning
+It is important to understand the security implications of this feature. Any
+credentials stored in a global credentials `Namespace` will be available to
+_all_ Kargo projects.
+:::
```
krancour, line 142: I know you moved this warning up from farther below, but this isn't the explanation for why `controller.serviceAccount.clusterWideSecretReadingEnabled=true` is highly discouraged. I would move this back to where it was...

Better yet, just remove it. _The whole section is about enabling global access to certain credentials._ It probably doesn't need repeating.

As for the warning text that _does_ belong here:

:::warning
Setting `controller.serviceAccount.clusterWideSecretReadingEnabled` setting to
`true` during Kargo installation will grant Kargo controllers cluster-wide read
permission on `Secret` resources.

__This is highly discouraged, especially in sharded environments where this
permission would have the undesirable effect of granting remote Kargo
controllers read permissions on all `Secret`s throughout the Kargo control
plane's cluster -- including `Secret`s having nothing to do with Kargo.__
:::

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -106,6 +106,41 @@ Refer to
 [the advanced section of the installation guide](./10-installing-kargo.md#advanced-installation)
 for more details.
 
+While doing this you must ensure that the Kargo controller has the appropriate
+permissions to read `Secret` resources in these namespaces. This is necessary because,
+as of Kargo v1.0.0, the controller does not have cluster-wide access to `Secret`s. Permissions
+are granted dynamically on a per-Project basis as `Projects` are created.
```
krancour, line 112: I have new thoughts about how to automate this. I will open a separate issue.

These changes are still worthwhile because this is the current state of things.

docs/docs/30-how-to-guides/20-managing-credentials.md
```diff
@@ -106,6 +106,41 @@ Refer to
...
+Setting `controller.serviceAccount.clusterWideSecretReadingEnabled` setting to
+`true` during Kargo installation will grant Kargo controllers cluster-wide read
+permission on `Secret` resources.
+
+__This is highly discouraged, especially in sharded environments where this
+permission would have the undesirable effect of granting remote Kargo
+controllers read permissions on all `Secret`s throughout the Kargo control
+plane's cluster -- including `Secret`s having nothing to do with Kargo.__
+:::
+
```
krancour, line 143: I think we need to visually attract more attention to the requirement of managing these permissions manually by putting it in a note box (implies this is important) and I think we can bury nitty gritty details in an info box (implies you can skip over it):

```suggestion
:::note
Operators must manually ensure Kargo controllers receive read-only access to `Secret`s in the designated namespaces. For example, if `kargo-global-creds` is designated as a global credentials namespace, the following `RoleBinding` should be created within that `Namespace`:

```yaml
kind: RoleBinding
metadata:
  name: kargo-controller-read-secrets
  namespace: kargo-global-creds
subjects:
- kind: ServiceAccount
  name: kargo-controller
  namespace: kargo
roleRef:
  kind: Role
  name: kargo-controller-read-secrets
  apiGroup: rbac.authorization.k8s.io
```
:::

:::info
By default, Kargo controllers lack cluster-wide permissions on `Secret`
resources. Instead, the Kargo _management controller_ dynamically expands
controller access to `Secret`s on a namespace-by-namespace basis as new
`Project`s are created.

_It is because this process does not account for "global" credential namespaces
that these bindings must be created manually by an operator._
:::

:::warning
Setting `controller.serviceAccount.clusterWideSecretReadingEnabled` setting to
`true` during Kargo installation will grant Kargo controllers cluster-wide read
permission on `Secret` resources.

__This is highly discouraged, especially in sharded environments where this
permission would have the undesirable effect of granting remote Kargo
controllers read permissions on all `Secret`s throughout the Kargo control
plane's cluster -- including `Secret`s having nothing to do with Kargo.__
:::

```

File: docs/docs/30-how-to-guides/20-managing-credentials.md [hunk 7]
```diff
@@ -106,6 +106,50 @@ Refer to
...
+kind: RoleBinding
+metadata:
+    name: kargo-controller-read-secrets
+    namespace: kargo-global-creds
+subjects:
+    - kind: ServiceAccount
+      name: kargo-controller
+      namespace: kargo
+roleRef:
+    kind: Role
```
jkleinlercher, line 126: I tested this RoleBinding and for me it only worked if I reference to a 'ClusterRole' because there is no 'Role' kargo-controller-read-secrets but only a 'ClusterRole' kargo-controller-read-secrets

krancour: > while ClusterRole grants permissions across all namespaces

Just fyi, that's not quite what it does here. I think this seems to be a lesser-known use of ClusterRoles, but when a namespaced `RoleBinding` references a `ClusterRole`, it gets the permissions of that `ClusterRole` _but only in that namespace_.

This is used to avoid having to redefine common sets of permissions in every namespace where they happen to be needed.

krancour (approved):
LGTM

PR #2940, opened 2024-11-15
feat: allow configuration of retry attempts for Promotion steps

hiddeco:
We have discovered some edge cases in which a Promotion step may end up in an infinite loop. To remediate this, we can configure the number of (retry) attempts a step is allowed to make before bailing.

This could later be extended to a waiting (or polling) configuration for steps like `git-wait-for-pr`, to e.g., set a deadline for merging a pull request.

With this pull request, **any** step execution (no matter the outcome) is counted as an "attempt", while the `wait-for-pr` and `argocd-update` steps have been allowed an unlimited number of retries (by default) to achieve backwards compatible behavior.

```yaml
steps:
- uses: argocd-update
  retry:
    attempts: 5
  config:
    apps:
    - name: guestbook
       sources:
       - repoURL: https://github.com/example/guestbook.git
```

**Note:** In the future, we probably want to introduce more rigorous handling of errors, to be able to distinguish "fatal" errors from transient ones.


internal/directives/promotions.go
```diff
@@ -27,6 +27,18 @@ type PromotionStepRunner interface {
 	RunPromotionStep(context.Context, *PromotionStepContext) (PromotionStepResult, error)
 }
 
+// RetryableStepRunner is an interface for PromotionStepRunners that can be
+// retried in the event of a failure.
+type RetryableStepRunner interface {
```
krancour, line 32: Super nit: `RetryableStepRunner` seems to imply that this interface is composed of `StepRunner` + additional methods. (In the same way an `io.ReadCloser` is _also_ an `io.Reader` and an `io.Closer`.)

I'm _not_ suggesting that you redefine it as such; rather I am wondering if `Retryable` might be a more apt name for the current definition.

I do not feel strongly about this at all, so completely disregard this if you think I'm bonkers. I may be.

internal/directives/promotions.go
```diff
@@ -78,11 +92,59 @@ type PromotionStep struct {
...
+}
+
+// RecordAttempt records an attempt to execute a step in the provided
+// output.
+func (s *PromotionStep) RecordAttempt(state State, output map[string]any) map[string]any {
+	attempts := s.GetAttempts(state)
+	if output == nil {
+		output = make(map[string]any, 1)
+	}
+	output[stateKeyAttempts] = attempts + 1
```
krancour, line 144: I'm iffy on this. Up to now those states/outputs are entirely user-defined and this starts to add system-defined things to it -- which also makes `attempts` a special output that no one else ought to use. I do know this is a convenient options since this map is persisted. Is there any another option you'd find acceptable?

hiddeco: We could count the number of attempts for a current step in a separate field in the Status, and then reset it after a step finishes. The only small disadvantage I can see with this is that the information about how many attempts it typically takes for a step to succeed gets lost.

hiddeco: Refactored this by adding `CurrentStepAttempt` to the Promotion status.

internal/directives/promotions.go
hiddeco: I didn't like _just_ `Retryable` because it seemed _too_ generic, which made me opt for this name instead.

krancour: Ya. I can see that perspective as well and it's a big part of why I don't feel strongly about this. Let's just leave it be.

api/v1alpha1/promotion_types.go
```diff
@@ -109,6 +109,27 @@ type PromotionVariable struct {
 	Value string `json:"value" protobuf:"bytes,2,opt,name=value"`
 }
 
+// PromotionRetry describes the retry policy for a PromotionStep.
+type PromotionRetry struct {
```
krancour, line 113: Given #2972, I wonder if we should name this `PromotionStepRetry` or even `PromotionStepRetryPolicy` as I think we will probably also end up with Promotion-level retry policies in addition to step-level.

krancour: Non-blocking.

krancour (approved):
🔥 This is awesome!

api/v1alpha1/promotion_types.go
hiddeco: Renamed it to `PromotionStepRetry` to reserve the policy version for e.g. some enum that describes how it should be retried.

PR #2941, opened 2024-11-16
refactor: replace helm-update-image with yaml-update

### LINKED ISSUE #2939: "rename" helm-update-image step to yaml-update
Ref: https://github.com/akuity/kargo/issues/2701#issuecomment-2479949647

This isn't _actually_ renaming. It is actually duplication + deprecation.

Note that the `helm-update-chart` step is unaffected by this because that step is more complex than a simple "set this key to this value."

krancour:
Fixes #2939

This PR:

* Reverts most of what #2925 did to the `helm-update-image` step. _There is no sense improving that step and deprecating individual fields if the entire step is going to be deprecated and replaced. This would only have led to confusion._

* Deprecates `helm-update-image`.

* Adds a new, generic `yaml-update` step that is essentially an improved/stripped down version of the `helm-update-image` step.

### DIFF
--- a/docs/docs/35-references/10-promotion-steps.md
+++ b/docs/docs/35-references/10-promotion-steps.md
@@ -394,6 +394,8 @@ file with new version information which is referenced by the Freight being
 promoted. This step is commonly followed by a [`helm-template`](#helm-template)
 step.
 
+__Deprecated: Use the generic `yaml-update` step instead. Will be removed in v1.2.0.__
+
 ### `helm-update-image` Configuration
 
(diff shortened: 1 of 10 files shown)


hiddeco (approved):

krancour:
Something is wrong here. The commit that undoes the `helm-update-image` changes is missing.

Looking into it.

krancour:
Fixed now.

krancour:
@Marvin9 possible to have both for now and then remove `helm-update-image` in 1.2?

Issue #2953, opened 2024-11-18
Kargo UI switches Warehouse filter when I'm about to promote


jessesuen:
# Description

I filtered my Freight based on a specific Warehouse. But when trying to promote, it switched my filter to my other Warehouse. I expect the filters to remain static, while interacting with UI

https://github.com/user-attachments/assets/4d5697c6-2ebb-4f97-8715-3c0ec43255f2

# Steps to Reproduce

See video

# Version

```
v1.0.3
```

PR #2958, opened 2024-11-18
feat(ui): warehouse selection persistence throughout pipeline 

Marvin9:
fixes https://github.com/akuity/kargo/issues/2953

### DIFF
--- a/ui/src/features/common/utils.ts
+++ b/ui/src/features/common/utils.ts
@@ -28,6 +28,20 @@ export function getCurrentFreight(stage: Stage): FreightReference[] {
     : [];
 }
 
+export function getCurrentFreightWarehouse(stage: Stage) {
+  const freightRef = getCurrentFreight(stage);
+  const isWarehouseKind = freightRef.reduce(
+    (acc, cur) => acc || cur?.origin?.kind === 'Warehouse',
+    false
+  );
+
+  if (isWarehouseKind) {
+    return freightRef[0]?.origin?.name || stage?.spec?.requestedFreight[0]?.origin?.name || '';
+  }
+
+  return '';
+}
+
 export function currentFreightHasVerification(stage: Stage): boolean {
   const collection = stage?.status?.freightHistory[0];
   return (collection && (collection.verificationHistory || []).length > 0) || false;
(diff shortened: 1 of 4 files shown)


ui/src/features/common/utils.ts
```diff
@@ -28,6 +28,39 @@ export function getCurrentFreight(stage: Stage): FreightReference[] {
     : [];
 }
 
+export function getCurrentFreightWarehouse(stage: Stage) {
+  const freightRef = getCurrentFreight(stage);
+  const isWarehouseKind = freightRef.reduce(
+    (acc, cur) => acc || cur?.origin?.kind === 'Warehouse',
+    false
+  );
```
rpelczar, line 36: Can you use this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

ui/src/features/common/utils.ts
```diff
@@ -28,6 +28,39 @@ export function getCurrentFreight(stage: Stage): FreightReference[] {
...
+    if (order === -1) {
+      return LATEST_FREIGHT;
+    }
+
+    return order;
+  }
+
+  // latest freight default
+  return LATEST_FREIGHT;
+}
```
rpelczar, line 62:
```suggestion
  const order = freightsInOrder?.findIndex((freight) => freight?.origin?.name === warehouse);

  return order > -1 ? order : 0;
}
```

rpelczar: It should do the same, and it is simpler.

ui/src/features/freight-timeline/freight-timeline-header.tsx
```diff
@@ -65,6 +66,28 @@ export const FreightTimelineHeader = ({
...
+      onChange={(value) => setSelectedWarehouse(value)}
+      size='small'
+      labelRender={({ label }) => <div className='text-xs font-medium'>{label}</div>}
+      optionRender={(opt) => (
+        <div className='text-xs font-normal w-full h-full flex items-center'>
+          <div>{opt.label}</div>
+        </div>
+      )}
+      options={[
+        ...(Object.keys(warehouses) || []).map((w) => ({ value: w, label: w })),
```
rpelczar, line 82: if the warehouses variable is undefined, this OR operator doesn't make sense here, I guess. It just throws an error.

Marvin9: the default is not undefined or null but empty object thats for sure

Issue #2880, opened 2024-10-30
Document the possible values of Warehouse imageSelectionStrategy


Chewie:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Add an entry in the docs for the possible values of imageSelectionStrategy, namely `Digest`, `Lexical` and `NewestBuild`.

# Motivation

Playing around with Kargo for evaluation, I wanted to try a toy continuous delivery scenario without semver:
- A push on main on the app repo triggers tests and a new image build
- That new image is immediately deployed to a staging cluster
- Some sort of E2E tests and/or human validation is made and triggers the next stage
- The new version is now deployed in production

However, I found it difficult to find the information to represent this scenario. The [key concepts](https://docs.kargo.io/concepts#warehouse-resources) kind of implies that a semver approach is recommended, and while the [CRD Reference](https://doc.crds.dev/github.com/akuity/kargo/kargo.akuity.io/Warehouse/v1alpha1@v1.0.0#spec-subscriptions-image-imageSelectionStrategy) shows an `imageSelectionStrategy` field, it doesn't show what the possible values are.

I was only able to discover the possible values `Digest`, `Lexical` and `NewestBuild` by looking at the old [0.4.0 release notes](https://github.com/akuity/kargo/releases/tag/v0.4.0).

However, I'm not sure if this is an oversight or an information hidden on purpose. If it is the latter, what would be the recommended approach to implement this scenario? On a more general note, I'd be interested to know if the project has strong opinions regarding promotion strategies, in particular those that want to move away from tagging releases.

Cheers!

krancour:
Thanks for opening this issue. You're the second to notice this in as many days.

It wasn't deliberate.

It's a combination of hand-written docs still evolving and us being taken by surprise that the enumerated accepted values for this field were not included in the generated documentation.

For reference, we had incorrectly expected this:

https://github.com/akuity/kargo/blob/f30f66b5b25ef5d943026cd6af8b2f2eef77bee6/api/v1alpha1/warehouse_types.go#L15-L16

To surface here:

https://doc.crds.dev/github.com/akuity/kargo/kargo.akuity.io/Warehouse/v1alpha1@v1.0.0#spec-subscriptions-image-imageSelectionStrategy

PR #2963, opened 2024-11-19
docs: warehouse selection strategies

fykaa:
~closes https://github.com/akuity/kargo/issues/2880~

This addresses the aforementioned problem in its broadest sense.

However, we would prefer to have a separate "Working with Warehouses" guide in the long run, which can be completed within this PR.

### DIFF
--- a/docs/docs/30-how-to-guides/16-working-with-warehouses.md
+++ b/docs/docs/30-how-to-guides/16-working-with-warehouses.md
@@ -1,0 +1,161 @@
+---
+description: Learn how to work effectively with Warehouses
+sidebar_label: Working with Warehouses
+---
+
+# Working with Warehouses
+
+Each Kargo warehouse is represented by a Kubernetes resource of type
+`Warehouse`.
+
+## The `Warehouse` Resource Type
+
+A `Warehouse` resource's most important field is its `spec.subscriptions` field,
+which is used to subscribe to one or more:
+
+* Container image repositories
+
+* Git repositories
+
+* Helm charts repositories
+
+The following example shows a `Warehouse` resource that subscribes to a
+container image repository and a Git repository:
+
+```yaml
+apiVersion: kargo.akuity.io/v1alpha1
+kind: Warehouse
+metadata:
+    name: my-warehouse
+    namespace: kargo-demo
+spec:
+    subscriptions:
+    - image:
+        repoURL: public.ecr.aws/nginx/nginx
+        semverConstraint: ^1.26.0
+    - git:
+        repoURL: https://github.com/example/kargo-demo.git
+```
+:::info
+Kargo uses [semver](https://github.com/masterminds/semver#checking-version-constraints) to handle semantic versioning constraints.
+:::
+
+### Image Subscription
+
+For subscriptions to container image repositories, the `imageSelectionStrategy` field specifies the method for selecting
+the desired image. The available strategies for subscribing to an image repository are:
+
+- `Digest`: Selects the image with the specified digest.
+- `Lexical`: Selects the image with the lexicographically greatest tag.
+- `NewestBuild`: Selects the image with the most recent build time.
+- `SemVer`: Selects the image that best matches a semantic versioning constraint.
+
+### Git Subscription
+
+In subscriptions to Git repositories, the `commitSelectionStrategy` field
+specifies the method for selecting the desired commit.
+The available strategies for subscribing to a git repository are:
+
+- `Lexical`: Selects the commit with the lexicographically greatest reference.
+- `NewestFromBranch`: Selects the most recent commit from a specified branch.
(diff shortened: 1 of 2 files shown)


docs/docs/30-how-to-guides/16-working-with-warehouses.md
```diff
@@ -0,0 +1,161 @@
...
+:::info
+Kargo uses [semver](https://github.com/masterminds/semver#checking-version-constraints) to handle semantic versioning constraints.
+:::
+
+### Image Subscription
+
+For subscriptions to container image repositories, the `imageSelectionStrategy` field specifies the method for selecting
+the desired image. The available strategies for subscribing to an image repository are:
+
+- `Digest`: Selects the image with the specified digest.
```
krancour, line 48: This isn't quite what this one does. This is for people who for whatever reason are ignoring the well-established best practice of not using mutable tags like `latest`.

When you use this one, you _must_ supply a value in the constaint field. That value is the name of a single "mutable" tag.

The strategy will retrieve the latest details for the image tagged that way, including new/updated digest.

docs/docs/30-how-to-guides/16-working-with-warehouses.md
```diff
@@ -0,0 +1,161 @@
...
+- `Lexical`: Selects the image with the lexicographically greatest tag.
```
krancour, line 49: It may be useful to describe the circumstances under which you might use this. It's useful for scenarios where tags incorporate a date/time stamp. With any date/time format that arranges units from most significant to least significant (for example, yyyymmdd), the lexicographically greatest tag will also be the newest. You'd typically pair this with a regex in the `allowTags` field to limit eligibility to just those tags that conform to the expected format. i.e. You don't want to ask the the lexicographically latest tag and end up with `zzzz-my-custom-build` instead of `nightly-20241211`.

docs/docs/30-how-to-guides/16-working-with-warehouses.md
```diff
@@ -0,0 +1,161 @@
...
+- `NewestBuild`: Selects the image with the most recent build time.
```
krancour, line 50: This is incredibly slow because it requires retrieval of metadata for every eligible tag. This makes it easy to blast through your registry's rate limit.

This should probably come with a warning. If it must be used, it is probably best to use `allowTags` to constrain the retrieval of metadata to as few images as possible.

docs/docs/30-how-to-guides/16-working-with-warehouses.md
```diff
@@ -0,0 +1,161 @@
...
+- `SemVer`: Selects the image that best matches a semantic versioning constraint.
```
krancour, line 51: Probably a good place to re-iterate this:

```
:::info
Kargo uses [semver](https://github.com/masterminds/semver#checking-version-constraints) to handle semantic versioning constraints.
:::
```

docs/docs/30-how-to-guides/16-working-with-warehouses.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Git Subscription
+
+In subscriptions to Git repositories, the `commitSelectionStrategy` field
+specifies the method for selecting the desired commit.
+The available strategies for subscribing to a git repository are:
+
+- `Lexical`: Selects the commit with the lexicographically greatest reference.
```
krancour, line 59:
```suggestion
- `Lexical`: Selects the commit with the lexicographically greatest tag.
```

As with the Lexical strategy for image selection, it may we good to explain the circumstances under which this is useful. Also as with that image selection strategy, it's good to limit the acceptable format here so that you don't end up with something like `zzz-custom` over `nightly-20241211`.

docs/docs/30-how-to-guides/16-working-with-warehouses.md
```diff
@@ -0,0 +1,161 @@
...
+- `NewestFromBranch`: Selects the most recent commit from a specified branch.
+- `NewestTag`: Selects the most recent commit associated with a tag.
```
krancour, line 61: Not quite what it does.

Unlike branches, tags are supposed to be immutable, so there should only be _one_ commit associated with each tag.

The purpose of this strategy is to select the commit associated with the newest (most recently created) tag.

Again, constraining the tags that are eligible is a good idea.

docs/docs/30-how-to-guides/16-working-with-warehouses.md
```diff
@@ -0,0 +1,161 @@
...
+- `SemVer`: Selects the commit that best matches a semantic versioning constraint.
```
krancour, line 62: This selects the commit referenced by a _tag_ that best matches the constraint.

krancour:
This is great, overall.

An additional tip is that you can find more information about how different subscription fields like relate to the different strategies by reading field-level comments in `warehouse_types.go`.

PR #2969, opened 2024-11-20
fix: missing enumerated values in generated docs for warehouse

fykaa:
closes https://github.com/akuity/kargo/issues/2880

doc ref: https://github.com/akuity/kargo/pull/2963


krancour:
This is a good start. I'll approve because I'd like to see this information out into the real world sooner rather than later, but we should follow-up in the very near future to document the behavior of each of those strategies.

PR #2973, opened 2024-11-20
fix(api): be more thorough with checking if secret management is enabled

krancour:
cc @Marvin9 thanks for finding.

### DIFF
--- a/internal/api/create_or_update_resource_v1alpha1.go
+++ b/internal/api/create_or_update_resource_v1alpha1.go
@@ -42,6 +42,14 @@ func (s *server) createOrUpdateResource(
 	ctx context.Context,
 	obj *unstructured.Unstructured,
 ) (*svcv1alpha1.CreateOrUpdateResourceResult, error) {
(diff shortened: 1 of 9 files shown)


internal/api/create_or_update_resource_v1alpha1.go
```diff
@@ -42,6 +42,14 @@ func (s *server) createOrUpdateResource(
 	ctx context.Context,
 	obj *unstructured.Unstructured,
 ) (*svcv1alpha1.CreateOrUpdateResourceResult, error) {
+	if obj.GroupVersionKind() == secretGVK && !s.cfg.SecretManagementEnabled {
```
Marvin9, line 45: not very familiar with this comparison but just checking, does it compare deep equal?

Marvin9 (approved):
Thank you!

internal/api/create_or_update_resource_v1alpha1.go
krancour: `==` comparison in Go isn't deep. It checks equality field-by-field. This type is composed of only three `string` fields, so although this comparison isn't deep, it's as deep as it needs to be.

PR #2975, opened 2024-11-20
feat: allow project scoped generic kubernetes secrets

Marvin9:
This PR now allows create generic k8s secrets

TODO:

- [x] API - Update secret
- [x] Tests

Screenshot 2024-11-20 at 6 51 09 PM (2)
Screenshot 2024-11-20 at 6 55 23 PM (2)
Screenshot 2024-11-20 at 6 55 56 PM (2)


Marvin9:
@krancour generic secret creation and listing those is ready. Till I figure out what to do for updates could you start looking at the backend code?

api/v1alpha1/labels.go
```diff
@@ -8,6 +8,8 @@ const (
 	CredentialTypeLabelValueGit   = "git"
 	CredentialTypeLabelValueHelm  = "helm"
 	CredentialTypeLabelValueImage = "image"
+	// TODO: Should we explicitly keep this label or agree that absence of this label should implicitly say that credential is generic?
+	CredentialTypeLabelValueGeneric = "generic"
```
hiddeco, line 12: I think the advantage of this label is that we do not automatically make any Secret managed by Kargo, but instead, continue to allow them to be "opt-in".

jessesuen: I was waffling on what to name this, but `generic` seems appropriate as it is the same term used by kubectl, i.e.

`kubectl create secret generic my-secret`

Marvin9: Alright. I will keep this explicit. Thanks!

Marvin9:
@krancour could you start looking.. I will bring code coverage to threshold

krancour:
Sorry this took me so long.

Overall, this looks fantastic. Test drive went really smoothly.

I have some small bits of feedback, that I think are easy to address.

Screenshot 2024-12-11 at 4 20 35 PM

In the screenshot above, I was initially thrown off by not seeing any kind of "Add Secret" button. It took me a while to realize that clicking "Add Credentials" would pop up a dialog that could be used for adding generic Secrets as well.

I think this could potentially be less confusing if each of the two sections of this page each got their own "Add Credentials" and "Add Secret" buttons.

In the same screen shot, upper right, the alt text for the tab says "Credentials". With Secrets being more broad/generic than credentials, I would suggest making the alt text there "Secrets." Credentials are a _kind_ of Secret. (If it's not a problem, I think the URL of the page could also change to end in "/secrets".

Screenshot 2024-12-11 at 4 25 17 PM

Two things tripped me up in the screenshot above.

The title of the dialog is "Edit Credentials" even when I'm working on a generic Secret. Could it be at all practical to have two _separate_ dialogs? One for creds and one for generic?

When editing an existing Secret, not seeing any `******` in the value fields left me wondering whether I was going to accidentally blank out those fields when I hit Update. I confirmed that this is _not_ a problem, but seeing `******` in those fields would have stopped me from wondering.

Last -- a weird one. If I try to add a new key "foobar" above, the new line disappears the moment I hit the second "o" in "foobar". I _assume_ there's some real time validation that keys aren't being duplicated and after hitting that second "o", it looks like I am duplicating "foo" although I wasn't done typing. If I type any other value _not_ staring with the characters of an existing key, everything works fine.

I think/hope these are all relatively minor tweaks. As I said, overall, this worked great and did exactly what it's supposed to.

Marvin9:
Great suggestions! @krancour you might want to take a look now, except your last point all other review points are addressed. Thanks

Marvin9:
@krancour ready to review again

internal/api/create_credentials_v1alpha1.go
```diff
@@ -34,34 +41,141 @@ func (s *server) CreateCredentials(
...
+		if err != nil {
+			return nil, err
+		}
+
+		return connect.NewResponse(
+			&svcv1alpha1.CreateCredentialsResponse{
+				Credentials: kubernetesSecret,
+			},
+		), nil
+	case kargoapi.CredentialTypeLabelValueGeneric:
```
krancour, line 76: I'm thinking a Secret that _isn't_ labeled as a generic credential should just automatically be treated as such. i.e. Perhaps generic credentials is the _default_ case?

The thing I think I want to avoid is Kargo insisting on a special label for Secret that _aren't_ really special in any way. (The "specific" creds _are_ special, so that doesn't bother me.)

krancour: Or... I bet that label was basically necessary to make the existing credential-management endpoints work for generic Secrets? This actually has got me thinking about whether separate endpoints may be in order for Secrets since credentials are special in many ways that generic Secrets aren't.

For generic Secrets, I think we need the following:

* Get/list endpoints that redact _all_ values and annotations. (Compare to creds, which exempts the values of certain keys from being redacted.)

  Under the hood, credential and generic secret endpoints can share a common redaction func and pass the names of exempt keys.

* Update endpoint that updates/deletes individual keys. The generic update resource endpoint cannot be used for this purpose. Since we redact a lot on the get, the client doesn't _have_ everything it needs to send back for a full update.

  This is another bit that seems it could be shared between cred update and generic secret update endpoints.

* Delete endpoint.

* Create can use the generic create resource endpoint since I don't see any special treatment being required.

krancour:
@Marvin9 overall, this is great. I had some suggestions about the back end parts of this. In a nutshell, the existing credential management endpoints do some very specific things that I don't think apply to generic Secrets.

I'm happy to help on the new endpoints I suggested if you want me to. Just lmk.

krancour:
Sorry... one last thing. Can we make this dialog say "Create Credentials" (but leave the _other_ dialog as "Create Secrets")?

Screenshot 2024-12-23 at 4 22 27 PM

internal/api/create_credentials_v1alpha1.go
Marvin9: > The thing I think I want to avoid is Kargo insisting on a special label for Secret that _aren't_ really special in any way. (The "specific" creds _are_ special, so that doesn't bother me.)

Shading some lights on [previous comments](https://github.com/akuity/kargo/pull/2975#discussion_r1852285190) which seems in favour of labels? Lets agree on either of way and I can update APIs accordingly @krancour

krancour:
@Marvin9 I promise I'll 👀 the latest changes tomorrow. Thank you for the hard work on this.

internal/api/create_credentials_v1alpha1.go
```diff
@@ -34,34 +34,63 @@ func (s *server) CreateCredentials(
...
+	credType := req.Msg.GetType()
+
+	if err := validateFieldNotEmpty("type", credType); err != nil {
+		return nil, err
 	}
 
+	switch credType {
+	case kargoapi.CredentialTypeLabelValueGit,
+		kargoapi.CredentialTypeLabelValueHelm,
+		kargoapi.CredentialTypeLabelValueImage:
```
krancour, line 46: This is more conventional in Go -- return early if there's a validation error and then the rest of the code isn't in the switch.

```suggestion
	switch credType {
	case kargoapi.CredentialTypeLabelValueGit,
		kargoapi.CredentialTypeLabelValueHelm,
		kargoapi.CredentialTypeLabelValueImage:
	default:
		return nil, connect.NewError(
			connect.CodeInvalidArgument,
			errors.New("type should be one of git, helm, image"),
		)
	}
```

internal/api/create_credentials_v1alpha1.go
```diff
@@ -34,34 +34,63 @@ func (s *server) CreateCredentials(
...
+			&svcv1alpha1.CreateCredentialsResponse{
+				Credentials: kubernetesSecret,
+			},
+		), nil
+	}
+
+	return nil, connect.NewError(
+		connect.CodeInvalidArgument,
+		errors.New("type should be one of git, helm, image"),
+	)
```
krancour, line 74: Related to [this comment](https://github.com/akuity/kargo/pull/2975/files#r1910720424)

internal/api/update_credentials_v1alpha1.go
```diff
@@ -33,31 +34,28 @@ func (s *server) UpdateCredentials(
...
-		repoURLISRegex: req.Msg.GetRepoUrlIsRegex(),
-		username:       req.Msg.GetUsername(),
-		password:       req.Msg.GetPassword(),
+	credType := req.Msg.GetType()
+	project := req.Msg.GetProject()
+	name := req.Msg.GetName()
+
+	if err := validateFieldNotEmpty("type", credType); err != nil {
+		return nil, err
 	}
```
krancour, line 43: We shouldn't make `type` required here. Existing clients like the CLI only send project, name, and whatever is being updated. No value is sent for a field not receiving an update.

If this seems strange, it's only because for many of the fields, a "get" gets you a pretty heavily redacted object, so we don't expect most fields to be passed _back_ to us with the update request, because the client doesn't even know their values.

internal/api/update_credentials_v1alpha1.go
```diff
@@ -80,7 +78,29 @@ func (s *server) UpdateCredentials(
 		)
 	}
 
-	applyCredentialsUpdateToSecret(&secret, credsUpdate)
+	switch credType {
```
krancour, line 81: This switch can be improved in the same way described [here](https://github.com/akuity/kargo/pull/2975/files#r1910720424)

internal/api/create_credentials_v1alpha1.go
krancour: Actually... the `validateCredentials()` method is already doing this same check.

internal/api/delete_credentials_v1alpha1.go
```diff
@@ -15,34 +12,16 @@ func (s *server) DeleteCredentials(
...
-	name := req.Msg.GetName()
-	if err := validateFieldNotEmpty("name", name); err != nil {
-		return nil, err
-	}
+	_, err := s.DeleteSecrets(ctx, &connect.Request[svcv1alpha1.DeleteSecretsRequest]{
+		Msg: &svcv1alpha1.DeleteSecretsRequest{
+			Project: req.Msg.GetProject(),
+			Name:    req.Msg.GetName(),
+		},
+	})
```
krancour, line 20: Constructing a protobuf message to pass to another function seems a bit unconventional.

If you want to share Secret deletion logic between two endpoints, can it take project and name as arguments instead?

internal/api/create_secrets_v1alpha1.go
```diff
@@ -0,0 +1,107 @@
...
+
+	kubernetesSecret := s.genericCredentialsToSecret(creds)
+
+	if err := s.client.Create(ctx, kubernetesSecret); err != nil {
+		return nil, fmt.Errorf("create secret: %w", err)
+	}
+
+	redactedSecret := sanitizeSecret(*kubernetesSecret, []string{})
+
+	return redactedSecret, nil
```
krancour, line 90:
```suggestion
	return sanitizeSecret(*kubernetesSecret, []string{}), nil
```

internal/api/create_credentials_v1alpha1.go
```diff
@@ -34,34 +34,63 @@ func (s *server) CreateCredentials(
...
+}
+
+// creates credentials used for known purpose; specifically external subscriptions
+// to private git repo or helm OCI or docker image
+func (s *server) createSpecificCredentials(ctx context.Context, creds specificCredentials) (*corev1.Secret, error) {
```
krancour, line 79: I think linters might only enforce it for exported things, but the generally accepted practice in Go docs is that a comment like this one should start with the name of the variable, function, method, etc. that it is documenting. It seems somewhat silly, but iirc many doc tools in the Go ecosystem assume this convention is observed.

```suggestion
// createSpecificCredentials creates credentials used specifically for
// git, Helm chart, and image container repositories.
func (s *server) createSpecificCredentials(ctx context.Context, creds specificCredentials) (*corev1.Secret, error) {
```

internal/api/create_credentials_v1alpha1.go
```diff
@@ -14,7 +14,7 @@ import (
 	svcv1alpha1 "github.com/akuity/kargo/pkg/api/service/v1alpha1"
 )
 
-type credentials struct {
+type specificCredentials struct {
```
krancour, line 17: I see this was renamed to differentiate it clearly from the new `genericCredentials` type, but I think we could leave this as it was (`credentials`) and rename `genericCredentials` to just `secret` -- which I think is cleaner because it doesn't imply any assumption that all secrets are credentials, which isn't necessarily true.

api/service/v1alpha1/service.proto
```diff
@@ -85,6 +85,13 @@ service KargoService {
   rpc ListCredentials(ListCredentialsRequest) returns (ListCredentialsResponse);
   rpc UpdateCredentials(UpdateCredentialsRequest) returns (UpdateCredentialsResponse);
 
+  /* Secrets APIs */
+
+  rpc ListSecrets(ListSecretsRequest) returns (ListSecretsResponse);
+  rpc CreateSecrets(CreateSecretsRequest) returns (CreateSecretsResponse);
+  rpc UpdateSecrets(UpdateSecretsRequest) returns (UpdateSecretsResponse);
+  rpc DeleteSecrets(DeleteSecretsRequest) returns (DeleteSecretsResponse);
```
krancour, line 93: I think these sound better if they are singular instead of plural. These are, after all creating, updating, or deleting a single k8s Secret resource.

(I am guessing you pluralized these initially to match CreateCredentials, UpdateCredentials, etc., credentials is a peculiar word because it's common in English to use "credentials" even when talking about only one. That doesn't apply to secret/secrets.)

```suggestion
  rpc CreateSecret(CreateSecretRequest) returns (CreateSecretResponse);
  rpc UpdateSecret(UpdateSecretRequest) returns (UpdateSecretResponse);
  rpc DeleteSecret(DeleteSecretRequest) returns (DeleteSecretResponse);
```

internal/api/list_secrets_v1alpha1.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return connect.NewResponse(&svcv1alpha1.ListSecretsResponse{
+		Secrets: secrets,
+	}), nil
+}
+
+func sanitizeSecret(secret corev1.Secret, exemptKeys []string) *corev1.Secret {
+	s := secret.DeepCopy()
+
+	s.StringData = make(map[string]string, len(s.Data))
```
krancour, line 65: Nit: It's better to move this initialization of `s.StringData` to just above where it becomes relevant, i.e. down above the `for` loop that starts on line 76. It keeps things easier to read and comprehend when related things are grouped more closely together.

internal/api/update_secrets_v1alpha1.go
```diff
@@ -0,0 +1,116 @@
...
+	secret := corev1.Secret{}
+	if err := s.client.Get(
+		ctx,
+		types.NamespacedName{
+			Namespace: project,
+			Name:      name,
+		},
+		&secret,
+	); err != nil {
+		return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("secret %s/%s not found", project, name))
```
krancour, line 46: You don't need to be explicit about the not found error code here. The error handling for the entire API server knows how to translate Kubernetes client errors to an appropriate error code.

krancour (approved):

Issue #2968, opened 2024-11-20
Infinite loop when promoting image


stephaneetje:
Hello, 

I've been trying to use kargo to only promote image tags (using helm-image-update) but on all my stages, i get an inifnite loop at the argocd-update step.

Here is my stage:

```
      steps:
        - config:
            checkout:
              - branch: main
                path: ./src
            repoURL: https://github.com/test/test-helm.git
          uses: git-clone
        - as: update-image
          config:
            images:
              - image: europe-docker.pkg.dev/test-001/test/scim
                key: image.tag
                value: Tag
            path: ./src/helm/scim/values-scim-staging-image.yaml
          uses: helm-update-image
        - as: commit
          config:
            messageFromSteps:
              - update-image
            path: ./src
          uses: git-commit
        - config:
            path: ./src
          uses: git-push
        - config:
            apps:
              - name: scim-staging
                sources:
                  - desiredCommitFromStep: commit
                    repoURL: https://github.com/test/test-helm.git
          uses: argocd-update
```

I also tried removing `desiredCommitFromStep` but didn't change anything.
My warehouse is only subscribing an image repository:

```
apiVersion: kargo.akuity.io/v1alpha1
kind: Warehouse
metadata:
  annotations:
    argocd.argoproj.io/sync-wave: "3"
    kargo.akuity.io/color: black
  name: scim-prod
  namespace: scim
spec:
  subscriptions:
  - image:
      imageSelectionStrategy: SemVer
      repoURL: europe-docker.pkg.dev/test-01/test/scim
      discoveryLimit: 10
      allowTags: ^v(\d+\.)?(\d+\.)?(\*|\d+)$
```

You can see in the Argocd app's envents that it is looping
<img width="1188" alt="Capture d’écran 2024-11-20 à 12 38 54" src="https://github.com/user-attachments/assets/53389d9b-ddd1-4ff0-b7b9-e91d6f317fa7">

Discussion has began on [this issue](https://github.com/akuity/kargo/issues/2952), opening another one as asked by @krancour

krancour:
@stephaneetje my apologies, as I seem to have misunderstood the nature of your loop. With the previous discussion having taken place in a thread about `desiredCommit`, I was misled into believing that to have been an actual factor in your problem. If you leave that field alone, as you are currently, AND if there is no commit in your Freight, then when syncing an App, Kargo has no notion of what specific revision that App should observably be synced to once complete.

[...]

PR #2980, opened 2024-11-22
fix(controller)!: do not attempt to infer desired revision when not specified

### LINKED ISSUE #2952: Removing `desiredCommitFromStep: commit` from argocd-update causes infinite sync loop
# Description

I have multiple Stages that are committing to the same branch (e.g. `HEAD`). Per @hiddeco suggestion, I removed the `desiredCommitFromStep: commit` option, since a stage writing to the same HEAD would cause the other to become 

```yaml
        - uses: argocd-update
          config:
            apps:
              - name: kargo-helm-ff-dev
                sources:
                # desiredCommitFromStep: commit     # << removing this causes an infinite sync loop
```

However, when syncing, Kargo went into an infinite loop sync-ing the application over and over again

# Screenshots

image

# Steps to Reproduce

```yaml
kind: Stage
metadata:
  name: dev
  namespace: kargo-helm-ff
  labels:
    kargo.akuity.io/shard: kargo-agent
spec:
  shard: kargo-agent
  requestedFreight:
    - origin:
        kind: Warehouse
        name: guestbook
      sources:
        direct: true
    - origin:
        kind: Warehouse
        name: features
      sources:
        direct: true
  promotionTemplate:
    spec:
      steps:
        - uses: git-clone
          config:
            checkout:
              - fromFreight: true
                path: ./src
              - branch: main
                path: ./out
            repoURL: https://github.com/dhpup/kargo-helm-ff.git
        - uses: helm-update-image
          as: update-image
          config:
            images:
              - image: ghcr.io/dhpup/guestbook
                key: image.tag
                value: Tag
            path: ./out/env/dev/values.yaml
        - uses: copy
          config:
            inPath: ./src/base/feature-flags.yaml
            outPath: ./out/env/dev/feature-flags.yaml
        - uses: git-commit
          as: commit
          config:
            messageFromSteps:
              - update-image
            path: ./out
        - uses: git-push
          config:
            path: ./out
        - uses: argocd-update
          config:
            apps:
              - name: kargo-helm-ff-dev
                sources:
                  - repoURL: https://github.com/dhpup/kargo-helm-ff.git
```

# Version

```
v1.0.3
```

# Logs

```
Paste any relevant application logs here.
```

krancour:
Fixes #2952 #2968

When determining whether an Application sync has completed successfully and again when assessing Application health as a factor in Stage health, the argocd-update step (as did legacy promotion mechanisms before it) often takes revision information into account. i.e. _It often requires ApplicationSources to be observably synced to a specific revision_ (commit ID or chart version).

Until now, when the desired revision for a given ApplicationSource has not been explicitly specified, the argocd-update step has attempted to _infer_ the desired revision by examining the Promotion's FreightCollection. If it found an artifact matching an ApplicationSource's repoURL field (and chart field, when applicable), _that_ revision of the artifact was presumed to be the one that ApplicationSource should be observably synced to.

This behavior has repeatedly tripped up users for a variety of reasons, the most prominent involving a scenario wherein multiple Stages use promotion processes that _update the head of a single, common branch, which is tracked by multiple Applications._ When a Promotion to one such Stage pushes a new commit, deemed its "desired revision," any Application that tracks that branch (and subsequently syncs) will no longer observably be synced to its _own_ "desired revision." A Stage's health checks count this apparent mis-sync as a health problem.

In other scenarios, this behavior has also been implicated in causing the argocd-update step to enter an infinite series of retries for what appear to be failed attempts to sync an Application.

This PR simply eliminates the behavior whereby the argocd-update step attempts to infer desired revision when none is specified. This essentially relegates the step's desiredRevision field (and the deprecated desiredCommitFromStep field) to the status of an "advanced" feature for use in scenarios not prone to the difficulties highlighted above.

This is a breaking change insofar as it changes _behavior_, although it will _not_ actually break anything as it _relaxes_ success criteria. Most users will notice nothing different. It is only users who _like_ the more thorough health checks who may wish to now begin explicitly specifying desired revisions, which is something easily accomplished using v1.1's new EL support.

This change will be prominently mentioned in the v1.1.0 release notes.


jessesuen (approved):
Thank you!

krancour:
> This needs a codegen rerun, but otherwise looks good to me.

Actually looks like codegen verification just timed out in CI... context deadline exceeded.

Kicked it and it ran to completion.

Issue #2514, opened 2024-09-09
Publish the `kargo` package to brew


nitishfy:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

Add a support to install the kargo via brew. Currently, the [installation of kargo CLI](https://docs.akuity.io/kargo-getting-started/access-kargo-instance#access-kargo-using-the-kargo-cli) is by using the `curl` command, We should release the kargo stable release package to brew so that Mac users can install the kargo via brew itself rather than using the `curl` command.

Furthemore, the docs will be in sync with how we [install the Akuity CLI Tool](https://docs.akuity.io/cli/#installation).

cc @krancour

krancour:
@fykaa now that we're GA the timing is right for this.

krancour:
Not stale, but also a very priority.

PR #3011, opened 2024-11-27
feat: Publish kargo to brew

fykaa:
closes https://github.com/akuity/kargo/issues/2514

Added the Kargo CLI to Homebrew. The formula supports macOS (Intel and ARM) and Linux (Intel and ARM) with architecture-specific binaries for version `v1.0.3` currently.
Formula can be found here: https://github.com/akuity/homebrew-tap

TODO: Give Permission to access `akuity/homebrew-tap.git` (for creating a pull request) to github-actions[bot] (`GITHUB_TOKEN`)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -333,6 +333,7 @@ jobs:
       base64-subjects: "${{ needs.combine_hashes.outputs.hashes }}"
       upload-assets: true # Optional: Upload to a new release
       provenance-name: kargo-cli.intoto.jsonl
+
   update-homebrew-formula:
     if: github.event_name == 'release'
     permissions:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -347,9 +347,6 @@ jobs:
 
     - name: Setup Git User
       uses: fregante/setup-git-user@v2
-      with:
-        name: ${{ github.actor }}
-        email: ${{ github.actor }}@users.noreply.github.com
 
     - name: Update formula and push
       run: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -341,34 +341,33 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ publish-image, publish-cli ]
     steps:
-      - name: Checkout bottle Repository
-        uses: actions/checkout@v4
-        with:
-          repository: akuity/homebrew-tap
-          token: ${{ secrets.GITHUB_TOKEN }}
-          path: homebrew-tap
+    - name: Checkout tap repository
+      uses: actions/checkout@v4
+      with:
+        repository: akuity/homebrew-tap
+        token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Update Homebrew formula
-        run: |
-          cd homebrew-tap
-          ./update.sh kargo ${{ github.ref_name }}
+    - name: Setup Git User
+      uses: fregante/setup-git-user@v2
+      with:
+        name: ${{ github.actor }}
+        email: ${{ github.actor }}@users.noreply.github.com
 
-      - name: Configure Git and Commit changes
-        run: |
-          cd homebrew-tap
-          git config user.name "${{ github.actor }}"
-          git config user.email "${{ github.actor }}@users.noreply.github.com"
-          git checkout -b update-kargo-${{ github.ref_name }}
-          git add kargo.rb
-          git commit -m "chore: update Kargo cli to version ${{ github.ref_name }}"
+    - name: Update formula and push
+      run: |
+        git checkout -b update-kargo-${{ github.ref_name }}
+        ./update.sh kargo ${{ github.ref_name }}
+        git add kargo.rb
+        git commit -m "chore: update Kargo cli to version ${{ github.ref_name }}"
+        git push origin update-kargo-${{ github.ref_name }}
 
-      - name: Push changes and open PR
-        uses: peter-evans/create-pull-request@v5
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          commit-message: "Update Kargo formula to version ${{ github.ref_name }}"
-          branch: update-kargo-${{ github.ref_name }}
-          base: main
-          title: "chore: update Kargo formula to version ${{ github.ref_name }}"
-          body: "This PR updates the Homebrew formula for Kargo to version ${{ github.ref_name }}."
-          delete-branch: 'true'
+    - name: Open PR
+      uses: peter-evans/create-pull-request@v5
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        commit-message: "Update Kargo formula to version ${{ github.ref_name }}"
+        branch: update-kargo-${{ github.ref_name }}
+        base: main
+        title: "chore: update Kargo formula to version ${{ github.ref_name }}"
+        body: "This PR updates the Homebrew formula for Kargo to version ${{ github.ref_name }}."
+        delete-branch: 'true'
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -355,9 +355,9 @@ jobs:
 
       - name: Configure Git and Commit changes
         run: |
+          cd homebrew-tap
           git config user.name "${{ github.actor }}"
           git config user.email "${{ github.actor }}@users.noreply.github.com"
-          cd homebrew-tap
           git checkout -b update-kargo-${{ github.ref_name }}
           git add kargo.rb
           git commit -m "chore: update Kargo cli to version ${{ github.ref_name }}"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -335,10 +335,7 @@ jobs:
       provenance-name: kargo-cli.intoto.jsonl
 
   update-homebrew-formula:
-    if: github.event_name == 'release'
-    permissions:
-      contents: write # Needed to push changes
-      pull-requests: write # Needed to create a pull request
+    if: github.event_name == 'release' && github.event.action == 'released' # Runs only for stable releases
     runs-on: ubuntu-latest
     needs: [ publish-image, publish-cli ]
     steps:
@@ -346,7 +343,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: akuity/homebrew-tap
-        token: ${{ secrets.GITHUB_TOKEN }}
+        token: ${{ secrets.TAP_PAT }}
 
     - name: Setup Git User
       uses: fregante/setup-git-user@v2
@@ -365,7 +362,7 @@ jobs:
     - name: Open PR
       uses: peter-evans/create-pull-request@v5
       with:
-        token: ${{ secrets.GITHUB_TOKEN }}
+        token: ${{ secrets.TAP_PAT }}
         commit-message: "Update Kargo formula to version ${{ github.ref_name }}"
         branch: update-kargo-${{ github.ref_name }}
         base: main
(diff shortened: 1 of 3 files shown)


docs/docs/40-contributor-guide/15-release-procedures.md
```diff
@@ -154,6 +154,29 @@ release date -- which is always a Friday.
...
+        ```bash
+        brew install --build-from-source ./kargo.rb
+        kargo --version
+        ```
+    * Commit and push the changes to the Homebrew tap repository:
+        ```bash
+        git add kargo.rb
+        git commit -m "Update Kargo formula to version vM.m.0"
+        git push origin main
+        ```
```
krancour, line 178: I'd rather we automated this. It should just be a new job in the release process.

hiddeco: It may at this point be easier to start looking into something like [GoReleaser](https://goreleaser.com), which has built-in utilities to update Brew formulas (in different repositories: [example](https://github.com/fluxcd/flux2/blob/f42a17de541d6ce2ffe56d0dcba96a26ce581818/.goreleaser.yml#L74-L88)).

krancour: She had been looking at that. I meant to read up as well.

krancour: > I will take a look and see how GitHub Actions can run scripts in a different repository

goreleaser, aside, this isn't hard. The a job in one repo can easily clone a different repo, make changes (by running a script, if that's the best way), and commit. Preferable to committing directly would be opening a PR because that way we can still have branch protections on the tap repo's main branch.

.github/workflows/release.yaml
```diff
@@ -333,3 +333,42 @@ jobs:
...
       provenance-name: kargo-cli.intoto.jsonl
+  update-homebrew-formula:
+    if: github.event_name == 'release'
+    permissions:
+      contents: write # Needed to push changes
+      pull-requests: write # Needed to create a pull request
+    runs-on: ubuntu-latest
+    needs: [ publish-image, publish-cli ]
+    steps:
+      - name: Checkout bottle Repository
```
krancour, line 344: Formatting nit: Throughout the rest of this file, we're not using the extra indent before `-`. We should updated this (and below) to match that.

.github/workflows/release.yaml
```diff
@@ -333,3 +333,42 @@ jobs:
...
+      - name: Update Homebrew formula
+        run: |
+          cd homebrew-tap
+          ./update.sh kargo ${{ github.ref_name }}
+
+      - name: Configure Git and Commit changes
+        run: |
+          cd homebrew-tap
+          git config user.name "${{ github.actor }}"
+          git config user.email "${{ github.actor }}@users.noreply.github.com"
```
krancour, line 360: There's an action that does this: `fregante/setup-git-user@v2`

What you've done works, but I'm aware of several other repositories of ours using this action, so it might be nice to stick with that.

.github/workflows/release.yaml
```diff
@@ -333,3 +333,42 @@ jobs:
...
+          git checkout -b update-kargo-${{ github.ref_name }}
+          git add kargo.rb
+          git commit -m "chore: update Kargo cli to version ${{ github.ref_name }}"
```
krancour, line 363: I'm pretty sure you need to push here. The `create-pull-request` action expects the head branch to already exist on the remote.

.github/workflows/release.yaml
```diff
@@ -333,3 +333,42 @@ jobs:
...
+    needs: [ publish-image, publish-cli ]
+    steps:
+      - name: Checkout bottle Repository
+        uses: actions/checkout@v4
+        with:
+          repository: akuity/homebrew-tap
+          token: ${{ secrets.GITHUB_TOKEN }}
+          path: homebrew-tap
+
+      - name: Update Homebrew formula
```
krancour, line 351: Since this step and the next are both scripts, it's reasonable to combine then into a single step named something like "Update formula and push".

.github/workflows/release.yaml
```diff
@@ -333,3 +333,42 @@ jobs:
...
+      pull-requests: write # Needed to create a pull request
+    runs-on: ubuntu-latest
+    needs: [ publish-image, publish-cli ]
+    steps:
+      - name: Checkout bottle Repository
+        uses: actions/checkout@v4
+        with:
+          repository: akuity/homebrew-tap
+          token: ${{ secrets.GITHUB_TOKEN }}
+          path: homebrew-tap
```
krancour, line 349: If you omit `path`, this action will check out into `$GITHUB_WORKSPACE` and then you can skip the `cd` in your scripts. This is the conventional usage when you are dealing with only one repo.

.github/workflows/release.yaml
krancour, line 344: Tap and bottle are two different things.

```suggestion
      - name: Checkout tap repository
```

.github/workflows/release.yaml
```diff
@@ -333,3 +333,41 @@ jobs:
       base64-subjects: "${{ needs.combine_hashes.outputs.hashes }}"
       upload-assets: true # Optional: Upload to a new release
       provenance-name: kargo-cli.intoto.jsonl
+  update-homebrew-formula:
```
krancour, line 336:
```suggestion

  update-homebrew-formula:
```

.github/workflows/release.yaml
```diff
@@ -333,3 +333,42 @@ jobs:
       base64-subjects: "${{ needs.combine_hashes.outputs.hashes }}"
       upload-assets: true # Optional: Upload to a new release
       provenance-name: kargo-cli.intoto.jsonl
+
+  update-homebrew-formula:
+    if: github.event_name == 'release'
+    permissions:
+      contents: write # Needed to push changes
+      pull-requests: write # Needed to create a pull request
```
krancour, line 341: Something to look out for. I _think_ these are permissions WRT to the current repository. This may not be sufficient to open a PR in a _different_ repository. We'll find out when we try this.

docs/docs/20-quickstart.md
```diff
@@ -489,10 +489,7 @@ the previous section.
     <TabItem value="general" label="Mac, Linux, or WSL" default>
 
     ```shell
-    arch=$(uname -m)
-    [ "$arch" = "x86_64" ] && arch=amd64
-    curl -L -o kargo https://github.com/akuity/kargo/releases/latest/download/kargo-"$(uname -s | tr '[:upper:]' '[:lower:]')-${arch}"
-    chmod +x kargo
+    brew install akuity/tap/kargo
```
krancour, line 492: I've been thinking about this particular change and I think we would be better off keeping the original install instructions until we're supremely confident that the automated formula updates are working as expected. Let's roll this one change back and re-add it at a later date.

.github/workflows/release.yaml
```diff
@@ -333,3 +333,42 @@ jobs:
       base64-subjects: "${{ needs.combine_hashes.outputs.hashes }}"
       upload-assets: true # Optional: Upload to a new release
       provenance-name: kargo-cli.intoto.jsonl
+
+  update-homebrew-formula:
+    if: github.event_name == 'release'
```
krancour, line 338: I think there are additional conditions...

I think we would want to exclude RCs and patch releases _not_ cut from the latest release-1.x branch.

idk how to do this off the top of my head.

You might be able to borrow bits of the logic used by the nightly to determine what the latest stable release is.

krancour (changes requested):

.github/workflows/release.yaml
hiddeco: You are correct, for permissions to another repository you need to use a token that has access to this repository. This also means we do not have to elevate the permissions for the step here.

.github/workflows/release.yaml
krancour: I thought about that as well. I'm open to that over the short-term.

I'm going to test this on my own Kargo fork and private tap in a bit.

.github/workflows/release.yaml
```diff
@@ -333,3 +333,43 @@ jobs:
       base64-subjects: "${{ needs.combine_hashes.outputs.hashes }}"
       upload-assets: true # Optional: Upload to a new release
       provenance-name: kargo-cli.intoto.jsonl
+
+  update-homebrew-formula:
+    if: >
+      github.event_name == 'release' &&
+      startsWith(github.ref, 'refs/tags/v') &&
+      !contains(github.ref, '-') &&
+      github.ref == format('refs/tags/v{0}', steps.latest_release.outputs.version)
```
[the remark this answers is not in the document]
hiddeco: I think you can do something like `github.event_name == 'release' && github.event.action == 'released'`, which would filter out prereleases. As our release candidates are typically marked as a pre-release, see e.g. https://github.com/akuity/kargo/releases/tag/v1.1.0-rc.3

Ref: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

.github/workflows/release.yaml
```diff
@@ -333,3 +333,39 @@ jobs:
...
+      uses: actions/checkout@v4
+      with:
+        repository: akuity/homebrew-tap
+        token: ${{ secrets.HOMEBREW_TOKEN }}
+
+    - name: Setup Git User
+      uses: fregante/setup-git-user@v2
+      with:
+        name: ${{ github.actor }}
+        email: ${{ github.actor }}@users.noreply.github.com
```
hiddeco, line 352: I think this configuration is redundant, based on reading the README of the action.

.github/workflows/release.yaml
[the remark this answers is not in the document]
hiddeco: The only thing my suggestion did not take into account is @krancour's

> [..] and patch releases not cut from the latest release-1.x branch

For this, I think we still need to pull off some version comparisons.

---

However, I also want to note that it's common for Homebrow to e.g. provide minor version taps. I.e., `kargo.rb` (latest), `kargo@1.1.rb` (1.1.x releases), `kargo@1.0.rb` (1.0.x releases), etc.

In the future, we should offer support for this as well. Without this, it would be impossible to update to a patch release for a previous minor version if it's released after a newer minor range.

.github/workflows/release.yaml
krancour: The tap absolutely needs to be a separate repo. Having played with this quite a bit recently, I'm pretty confident that things work most seamlessly when the tap repo is named <org>/homebrew-tap.

(Semi-related: akuity/homebrew-tap is not currently following the conventional layout for a tap, but it does work and we can deal with that separately later.)

Here, you can use `secrets.TAP_PAT` and I'll ensure that secret is configured with a PAT that can write to akuity/homebrew-tap.

.github/workflows/release.yaml
hiddeco: My point was more with this bit from the [action documentation](https://github.com/fregante/setup-git-user?tab=readme-ov-file#usage):

> New commits and tags will be assigned to the [@actions](https://github.com/actions) user. **If you want to customize the user, you don't need this action.**

.github/workflows/release.yaml
```diff
@@ -333,3 +333,55 @@ jobs:
...
+    - name: Update formula and push changes
+      run: |
+        git checkout -b update-kargo-${{ github.ref_name }}
+ 
+        if [[ ${{ github.ref_name }} == ${{ env.LATEST_TAG }} ]]; then
+          ./update.sh kargo ${{ github.ref_name }}
+          git add kargo.rb
+        fi
+        
+        ./update.sh kargo ${{ github.ref_name }} ${{ env.VERSION_SUFFIX }}
```
krancour, line 373: I have some concerns about this script. It seems a bit brittle and also isn't producing formulae that pass the default CI tests included in a new tap (which I added into that repo not long ago since they were absent).

There's been previous mention of using goreleaser to publish the CLI to our homebrew tap. I've played with it a bit and it works _extremely_ well. I am highly in favor of pursuing that further.

The one catch is that the free/open source version of goreleaser cannot publish versioned formulae, but a pro subscription isn't very expensive and I will sort out how we can get one.

krancour:
Closing this because we decided some time ago that we want to do this using goreleaser instead.

Issue #2827, opened 2024-10-23
[Feature Idea/Suggestion] promotionTemplates as a CRD of its own?


pww217:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

I think that it may related to [this issue ](https://github.com/akuity/kargo/issues/2614) which I believe is about reducing the amount of total configuration and boilerplate code for `promotionTemplates`.

But in summary, I wonder if Kargo may be more modular and flexible if promotionTemplates were a CRD like analysisTemplates are.

# Motivation

I've actually been writing a Helm Chart for my own use cases and for the most part, it's quite compact. I want to clean it up a bit more and test it, and then I intend to offer it upstream and see if there's any interest. 

The chart is designed to provision a `Project`, `Warehouses`, and `Stages` dynamically without having to use explicit manifests or Kustomize. This idea was first floated [here](https://github.com/akuity/kargo/discussions/2472). The chart is meant to organize these CRs around a top-level service definition and flesh out the entire pipeline. I want it to be silly simple that my developers can manage it themselves.

Here is a sample of a values file:

```yaml
owner: my-team

# Will override the naming conventions that are by default based on .Release.Name
nameOverride:

# Services definition including both warehouses and stages.
# This is design only for Github, Helm, and Amazon ECR at the moment.
services:
  - name: nginx
    imageURL: docker.io/library/nginx
    imageSelectionStrategy: NewestBuild
    git:
      repoURL: https://github.com/nginx-app.git
      valuesFilePath: infra/values # Path to values file(s) from repo root
      imageValueFile: version.yaml # The file that actually contains the image and/or tag key(s)
      key: image.tag # The key that the image is defined at in the values file.

    # Each stage represents a distinct environment or stop along the pipeline.
    # The source of each must be `direct` (from image Warehouse), or the name of a previous stage.
    stages:
      - name: dev
        autoPromotionEnabled: true
        source: direct # Pulls from the Warehouse
        argoCDAppName: nginx-dev
        verifications:
          [] # Examples of analysisTemplates below
          # - name: unit-test
      - name: staging
        autoPromotionEnabled: true
        source: dev # From previous stage
        argoCDAppName: nginx-staging
        verifications:
          []
          # - name: integration-test
          # - name: end-to-end-test
      - name: prod
        autoPromotionEnabled: false
[...]

PR #3022, opened 2024-11-29
feat(api): make PromotionTemplate dedicated resource

hiddeco:
Fixes: #2827

This makes the `PromotionTemplate` a namespaced resource and allows it to be referenced in a Stage in addition to being embedded. By doing this, and in combination with the [support for variables introduced in Kargo 1.1.x](https://main.docs.kargo.io/references/expression-language/), promotions that are the same across Stages can now avoid duplication.

For example:

```yaml
---
kind: PromotionTemplate
metadata:
  name: promote-to-stage
  namespace: guestbook
spec:
  vars:
  - name: gitRepo
    value: https://github.com/example/guestbook.git
  - name: srcPath
    value: ./src
  - name: outPath
    value: ./out
  - name: targetBranch
    value: stage/${{ ctx.stage }}
  steps:
  - uses: git-clone
    config:
      repoURL: ${{ vars.gitRepo }}
      checkout:
      - fromFreight: true
        path: ${{ vars.srcPath }}
      - branch: ${{ vars.targetBranch }}
        create: true
        path: ${{ vars.outPath }}
  - uses: git-clear
    config:
      path: ${{ vars.outPath }}
  - uses: kustomize-set-image
    as: update-image
    config:
      path: ${{ vars.srcPath }}/base
      images:
      - image: public.ecr.aws/nginx/nginx
  - uses: kustomize-build
    config:
      path: ${{ vars.srcPath }}/stages/${{ ctx.stage }}
      outPath: ${{ vars.outPath }}
  - uses: git-commit
    as: commit
    config:
      path: ${{ vars.outPath }}
      messageFromSteps:
      - update-image
  - uses: git-push
    config:
      path: ${{ vars.outPath }}
      targetBranch: ${{ vars.targetBranch }}
  - uses: argocd-update
    config:
      apps:
      - name: example-${{ ctx.stage }}
        sources:
        - repoURL: ${{ vars.gitRepo }}
          desiredCommit: ${{ outputs.commit.commit }}
---
kind: Stage
metadata:
  name: dev
  namespace: guestbook
spec:
  # ...omitted for brevity
  promotionTemplateRef:
    name: promote-to-stage
---
kind: Stage
metadata:
  name: staging
  namespace: guestbook
spec:
  # ...omitted for brevity
  promotionTemplateRef:
    name: promote-to-stage
```

**Note:** While working on this, I could think of two future improvements people may desire:

1. The ability to provide "last mile" variables, to e.g. overwrite PromotionTemplate defaults and/or provide Stage-specific variables which are hard to derive based on the context. For example, because the folder structure is not set up to equal or incorporate Stage names.
2. Conditional execution of steps, i.e. being able to define an `if` on a step to only execute if `ctx.stage == "dev"`. This would allow a PromotionTemplate to be used for a wider set of Stages and could further avoid duplication and/or a desire to do "compositions" as is being opted in #2827.


jessesuen:
Instead of a single ref, I think we should support chaining together multiple, so they can be composed

```yaml
spec:
  promotionTemplateRefs:
  - name: prechecks-stuff
  - name: gitops-stuff
  - name: notification-stuff
```

krancour:
Safe to close, knowing #3121 is the new direction?

Issue #2069, opened 2024-05-29
Option to delay Freight verification in a Stage


jessesuen:
# Proposed Feature

As soon as a Freight is successfully promoted into a Stage, it is immediately verified to go to downstream Stages. This is often too fast for users. Instead, they wish for there to be a minimum "soak" or "bake" time before it can **_automatically_** go downstream (e.g. 1h, 1d).

We could introduce a `verificationDelay` option in a Stage where a Freight must be running in a Stage for some minimum amount of time, before it is qualified to downstream

# Motivation

Users want automated promotions, but don't want for it to be immediate.

# Suggested Implementation

Delay the marking of a Freight to be verifiedIn a stage until it passes the user specified time.

There is a workaround for now, which is to reference an AnalysisTemplate that has an initialDelay. e.g.:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: AnalysisTemplate
metadata:
  name: soak
spec:
  metrics:
    - name: soak
      count: 1
      initialDelay: 1h
      provider:
        web:
          url: https://example.com # Choose a URL that will return 200
```

csantanapr:
I'm interested on this feature as I'm building POC using Kargo to deploy EKS clusters using Kubernetes Controllers like CAPI/ACK/Crossplane, and soak/bake time is important when rolling out upgrades in waves, with different soak time criteria. for example:
- after env/dev is deployed, we want to wait 6hrs before starting deploying env/test
- after env/test is deployed, we want to wait 3hrs before starting deploying env/prod
Reference: https://mollysheets.com/2024/02/03/cell-based-architecture-lowering-the-blast-radius-by-accepting-continuous-deployment-is-here/

hiddeco:
Your comment (@csantanapr) makes me wonder if our actual desire is not to delay the verification process, but rather to delay the (automatic) creation of the promotion for the next Stage based on a set of options.

I think this would also be the superior option, as it puts the consumer in control over their "delayed rollout" desires instead of the producer. I.e., in a scenario where you have `A -> [B, C]`, both `B` and `C` can define their unique soak desires instead of it being forced upon them by `A`.

PR #3040, opened 2024-12-03
feat(api): track verification and approval times

hiddeco:
This starts tracking the Stage approval and verification timestamps, which is generally useful for insight and tracking purposes but also a prerequisite to #2069.


krancour (approved):
Great improvement!

Issue #3038, opened 2024-12-02
Use builtin Kubernetes feature for gRPC health check


hiddeco:
Since Kubernetes 1.27 (available since April 2023), there is a [built-in gRPC liveness probe available](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe).

We should consider starting to use this to free ourselves from the dependency on [`grpc_health_probe`](https://github.com/grpc-ecosystem/grpc-health-probe), which we have to [bake into our image](https://github.com/akuity/kargo/blob/main/Dockerfile#L69-L71) (and keep updated).

hiddeco:
@krancour @jessesuen what Kubernetes versions are we officially supporting? Anything that is not EOL?

krancour:
It's not something we've formally established yet.

> Anything that is not EOL?

That would be my vote, but that means v1.29+.

I defer to @jessesuen.

hiddeco:
As noted in #3041, we can not fully adopt this until Kubernetes starts to support TLS in the built-in probe.

krancour:
We can just close this now because the protobuf-based API is being phased out anyway.

PR #3041, opened 2024-12-03
chore(chart): start using Kubernetes' builtin gRPC probe

hiddeco:
Related to #3038 

This starts to make use of the [builtin gRPC probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) in an attempt to eventually drop our dependency on `grpc_health_probe`.

Sadly, however, the builtin probe does not support TLS (https://github.com/kubernetes/kubernetes/issues/128365) (or disabling verification), and we can thus not drop the `grpc_health_probe` binary yet.

### DIFF
--- a/charts/kargo/templates/api/deployment.yaml
+++ b/charts/kargo/templates/api/deployment.yaml
@@ -87,24 +87,30 @@ spec:
               protocol: TCP
 {{- if .Values.api.probes.enabled }}
           livenessProbe:
+            {{- if .Values.api.tls.enabled }}
             exec:
               command:
                 - /usr/local/bin/grpc_health_probe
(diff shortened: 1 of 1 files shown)


hiddeco:
This kind of conflicts with what is being attempted in #2906.

krancour:
@hiddeco I will make sure #2906 lands in a spot that won't conflict with this. I don't want the probes _fully_ configurable -- just the various thresholds.

krancour:
Let's get this in ahead of #2906.

krancour (approved):

PR #3042, opened 2024-12-03
fix(controller)!: argocd-update: require desired revision to be specified when using updateTargetRevision

krancour:
@hiddeco and I have both been tripped up by this recently.

The previous behavior was that `updateTargetRevision` had no effect if neither `desiredCommitFromStep` (deprecated) or `desiredRevision` is set.

For @hiddeco and myself, it failed, because the previous target revision in the Applications in question was a placeholder.

In other cases, this would have effectively, but silently been a no-op, which surely would trip other users up.

This is especially important in light of #2980 which stopped attempting to infer desired revision when it's not explicitly set.


hiddeco (approved):

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -1026,7 +1026,7 @@ spec:
...
-| `apps[].sources[].updateTargetRevision` | `boolean` | Y | Indicates whether the target `ApplicationSource` should be updated such that its `targetRevision` field points directly at the desired revision. A `true` value in this field has no effect if `desiredRevision` and `desiredCommitFromStep` are both undefined. |
+| `apps[].sources[].updateTargetRevision` | `boolean` | Y | Indicates whether the target `ApplicationSource` should be updated such that its `targetRevision` field points directly at the desired revision. A `true` value in this field requires exactly one of `desiredCommitFromStep` or `desiredRevision` to be specified. |
```
34fathombelow, line 1029: PTAL at #L1074, s/'desiredCommit: ${{ outputs.commit.commit }}'/'desiredRevision: ${{ outputs.commit.commit }}'

krancour: Nice catch. This is in the merge queue, so I will fix in a separate PR.

PR #3045, opened 2024-12-03
docs: document outline of step options

hiddeco:


docs/docs/35-references/10-promotion-steps.md
```diff
@@ -5,10 +5,93 @@ description: Learn about all of Kargo's built-in promotion steps
...
+  as: alias
+- uses: another-step
+  config:
+    input: ${{ outputs.alias.someOutput }}
+```
+
+:::tip
+If a step does not have an alias, a default alias is generated based on the
+step's position in the `steps` list, starting with `step-0`.
+:::
```
krancour, line 66: Since this was added for internal reasons and I don't think it wise for users to rely on this, I'm questioning whether we're better off not mentioning this detail.

Not a hill I would die on. You decide.

krancour (approved):

docs/docs/35-references/10-promotion-steps.md
hiddeco: Removed it, thanks 🙇

PR #3053, opened 2024-12-04
prevent unsuccessful promo from blocking next promo

### LINKED ISSUE #3050: Manual promotion stays Pending in a Stage
# Description

I have a Stage (~which has auto-promotion enabled~), where a Pending promo will not run. I started a manual promotion, but that promo remains in `Pending`.

# Screenshots

image

# Steps to Reproduce

Based on the logs, I think to reproduce this:
* ~have a Failed promotion for the existing Freight.~ See screenshot. In my case, the previous promotion was aborted.
* Try to promote the same freight again
* ~I also believe that autopromotion may be a second requirement.~ I will try disabling auto-promotion to see if I'm correct. 

EDIT: I disabled auto-promotion but it remains `Pending`, so auto-promotion is not a factor.
EDIT2: I also deleted the previously failed promotions but it still remains `Pending`

# Version

```
v1.2.0-unstable-20241203
```

# Logs

The following is the output when refreshing the stage with a `Pending` promotion:

```
time="2024-12-04T00:39:03Z" level=debug msg="reconciling Stage" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="syncing Promotions" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="current Freight needs to be verified before allowing new promotions to start" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="assessing health" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="Last promotion did not succeed: no health checks to perform" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="verifying Stage Freight" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="Stage has not passed health checks: skipping verification" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="verifying Freight for Stage" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="auto-promoting Freight" controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="found PromotionPolicy associated with Stage" autoPromotionEnabled=true controlFlow=false namespace=guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="Promotion already exists for Freight" controlFlow=false freight=072a46be6f4cb218ac3ed730a9bf502d76e46be2 namespace=guestbook origin=Warehouse/guestbook stage=test
time="2024-12-04T00:39:03Z" level=debug msg="done reconciling Stage" controlFlow=false namespace=guestbook stage=test```

### PR DESCRIPTION
**Pull request by krancour**
Fixes #3050

Draft because this needs quite a serious sanity check from @hiddeco.

@jessesuen if you're in a position to verify this resolves the issue, I built a custom image for you with the fix. It's at `ghcr.io/akuity/kargo-unstable:custom-20241204`.

### DIFF
--- a/internal/controller/stages/regular_stages.go
+++ b/internal/controller/stages/regular_stages.go
@@ -636,11 +636,20 @@ func (r *RegularStageReconciler) syncPromotions(
 			return newStatus, hasNonTerminalPromotions, nil
 		}
 
-		// If we are in a healthy state, the current Freight needs to be verified
-		// before we can allow the next Promotion to start. If we are unhealthy,
-		// then we can allow the next Promotion to start immediately as the
-		// expectation is that the Promotion can fix the issue.
-		if stage.Status.Health == nil || stage.Status.Health.Status != kargoapi.HealthStateUnhealthy {
+		// As long as the last Promotion was successful and the Stage isn't
+		// unhealthy, we need to wait for the current Freight to complete
+		// verification (regardless of outcome) before we can allow the next
+		// Promotion to start.
+		//
+		// N.B.: If the last Promotion failed or was aborted then the current
+		// Freight did not arrive via that Promotion. It may have arrived prior to
+		// v1.1.0 when we began creating Freight history in the event of an implicit
+		// verification. What we are trying to avoid is a scenario where we're
+		// waiting on verification of Freight that might have been verified long ago
+		// and isn't being verified now because the unsuccessful promotion has
+		// prevented further health checks and verifications from running.
+		if (lastPromo != nil && lastPromo.Status != nil && lastPromo.Status.Phase == kargoapi.PromotionPhaseSucceeded) &&
+			(stage.Status.Health == nil || stage.Status.Health.Status != kargoapi.HealthStateUnhealthy) {
 			curVI := curFreight.VerificationHistory.Current()
 			if curVI == nil || curVI.Phase != kargoapi.VerificationPhaseSuccessful {
 				logger.Debug("current Freight needs to be verified before allowing new promotions to start")

### CODE REVIEW

**Comment by hiddeco** [comment]
I think that if a Promotion fails, a Stage _could_ be considered not healthy. However, given the state it is in at that point is uncertain and we have no means to tell it is — which is also what my `TODO` note aims at there.

However, given that `TODO`, I think the solution might be to eventually implement something which allows us to continue to assess health based on the _previous_ Promotion. Thereby making checking the (failure) result of the current Promotion redundant as you would always have a health result.

If you agree with this, I think your current PR does cover for it now and we should eventually address the `TODO` instead. What I would just like to see is a test case added, to ensure we have some sort of regression coverage.

**Comment by krancour** [comment]
@jessesuen I re-pushed to `ghcr.io/akuity/kargo-unstable:custom-20241204` with the new approach.

**Comment by jessesuen** [comment]
Switched to `ghcr.io/akuity/kargo-unstable:custom-20241204`, and it unblocked the Pending promotion. So, without looking at the code, from a functional standpoint, it appears to have worked.

**Review by hiddeco — APPROVED** [review]

**Comment by jvincze84** [comment]
Hi All,
We have the same issue as in the #3050 issue with  version 1.4.0 and 1.4.2.
Error message:

```plain
time="2025-04-17T13:03:32Z" level=debug msg="current Freight needs to be verified before allowing new promotions to start" controlFlow=false namespace=kargo-ms-demo stage=kargo-ms-demo-int
```

If a promotion failed, all subsequent promotions are stuck in Pending phase. 

And if we roll back to "kargo-unstable:custom-20241204" version, we can promote new freight.
Are the new versions include this fix?

Thanks,
Janos Vincze

krancour:
@jvincze84:

> If a promotion failed, all subsequent promotions are stuck in Pending phase.

Can you confirm that the Promotion is _actually_ in a failed state? i.e. Confirm you're not using "failed" in a more casual sense wherein the Promotion _succeeded_ but you weren't happy with the results.

I don't believe a Stage whose last Promotion is actually failed should be attempting any verification that should delay the next Promotion, but we can double-check.

jvincze84:
Hi @krancour,

Yes, I can confirm that the promotion is in errored state..
We are using ArgoCD integration, and the failed step is "arrocd-update".
For example if Argocd failed with manifest validation error, the entire stage stuck in failed phase and unknown health. 
We cannot abort, revert, fix this promotion. The only way is delete and recreate the stage.

krancour:
@jvincze84 if you are able to reliably reproduce this (sound like you are), could you please look at the Stage's status and make note of what the conditions are?

hiddeco:
@jvincze84, thank you for sharing all the information so far. Could you also provide the controller logs for the Stage to which the Freight is promoted (but not picked up)?

jvincze84:
> @jvincze84, thank you for sharing all the information so far. Could you also provide the controller logs for the Stage to which the Freight is promoted (but not picked up)?

Hi,
You can find the logs below:

[kargo-logs.zip](https://github.com/user-attachments/files/19864419/kargo-logs.zip)

Thanks a lot.

jvincze84:
Hi,
I'm wondering if there is any workaround now to avoid deleting the entire stage?
Thanks,
Jvincze84

Issue #3055, opened 2024-12-04
GitHub Apps Credentials - error decoding private key, cannot figure out correct format


jeffmccune:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

Following [GitHub App Authentication](https://docs.kargo.io/how-to-guides/managing-credentials/#github-app-authentication) docs, I can't convince Kargo to accept the credentials.

# Screenshots

<img width="821" alt="Screenshot 2024-12-04 at 12 46 21 PM" src="https://github.com/user-attachments/assets/f4c8b99e-46ca-4956-9420-bc3d9ee1cc8d">

# Steps to Reproduce

1. Create and install a GitHub App as described at [GitHub App Authentication](https://docs.kargo.io/how-to-guides/managing-credentials/#github-app-authentication)
2. Follow the [Kargo Quickstart](https://docs.kargo.io/quickstart/#your-first-kargo-project) to configure the Project, Warehouse, and Stages
3. Create a secret as described in the GitHub App Authentication doc.

Take care to ensure the string values don't have trailing newlines and the base64 pem encoded key is the value:

```bash
kubectl get secret -n kargo-demo github-app-creds -o json \
  | jq --exit-status '.data | map_values(@base64d)'
```

```json
{
  "githubAppID": "1099999",
  "githubAppInstallationID": "57999999",
  "githubAppPrivateKey": "-----BEGIN RSA PRIVATE KEY-----\nMIIE\nREDACTED\n-----END RSA PRIVATE KEY-----",
  "repoURL": "https://github.com/jeffmccune/bank-of-holos.git"
}
```

# Version

```
v1.0.3
```

# Logs

```
failed to run step "git-clone": error getting credentials for https://github.com/jeffmccune/bank-of-holos.git: error getting installation access token: error decoding private key: illegal base64 data at input byte 0
```

hiddeco:
I suspect your private key is not base64 encoded as a literal Secret value (i.e. encoded twice).

jeffmccune:
@hiddeco Do you mean the pem file GitHub provides needs to be base64 encoded a second time?  PEM encoding is Base64 encoding so I figured the Kargo documentation was saying to take care to provide the pem contents.

If Kargo requires double base 64 encoding for some reason, it might be worth an update the docs to clarify.

[...]

PR #3059, opened 2024-12-04
fix(creds): don't require base64 PK for GitHub Apps

hiddeco:
Fixes: https://github.com/akuity/kargo/issues/3055


krancour (approved):
Still can only guess at why I did it that way. I can't dispute how absurd it seems.

The fix though... elegant. 👏

krancour:
As small, straightforward, and useful as this is, I can't think of a reason not to sneak this into v1.1.0 since we're still waiting on one or two larger fixes anyway. Backporting.

PR #3060, opened 2024-12-05
fix: refactor retries

### LINKED ISSUE #3052: undesired behavior from promotion step retries
We may have missed the mark on promotion step retries. I am seeing infinite retries of a `wait-for-pr` step where the PR was merged without being closed, which is a failure condition. This step _does_ default to an infinite number of attempts, but no amount of retries will resolve this failure.

Note: We do _not_ need to go down the rabbit hole of distinguishing between technical errors and logical failures. We also do not need to distinguish between recoverable and unrecoverable errors. Although these are things we will get around to, I believe there is a more fundamental problem with our current implementation of retries.

There are two different scenarios retries need to address:

1. You want to retry when an unanticipated error occurs.

2. You want to _cap_ the retries when no error has occurred, but you've been waiting a long time for something that might never happen, such as waiting for a PR to be merged.

We got no.1 right. If an error/failure occurs and max attempts isn't exhausted, we downgrade the error/failure to "running." The promotion will be requeued and the step will be re-attempted on the Promotion's next reconciliation attempt. The system-wide default used by most built-in steps is one attempt -- no retries.

No.2 is where we seem to have two separate problems:

* The logic that checks if max attempts have been exhausted is only ever triggered by a step that returned an error/failure. i.e. Even with a finite cap on attempts, retries are _effectively_ infinite unless an error occurs.

* Steps that wait for something to happen, such as waiting for a PR to be merged, default to infinite attempts. This makes sense. It can take a long time for a PR to be merged. But the logic that worked so well for case no.1 will downgrade an infinite number of errors/failures to "running," as if the step is just waiting. In fact, even _without_ infinite attempts, any number of attempts greater than one could result in undesired retries after an unrecoverable error/failure.

I can't really spot a way to logic ourselves out of this without adding additional options. I believe the solution may be to define the max number of _total_ attempts and, separately, an error/failure threshold. A good default for _all_ existing steps would be infinite total attempts (as long as the step says its still running/working/waiting) with immediate failure after a single error/failure.

Related: I think a configurable timeout would make sense as well. It could default to no limit for all existing steps, but users could indicate, for instance, that a `wait-for-pr` step fails if the PR isn't merged within 48h. This is a nice-to-have that could be handled in a follow-up.

I'm happy to do the legwork on this issue if my analysis seems accurate and the approach seems sensible to @jessesuen and @hiddeco.

krancour:
Fixes #3052

Never fear... a lot of this is just codegen.

None of this is breaking because the bits that are refactored haven't been released yet. 😄 

@hiddeco this follows the plan you and I discussed offline earlier.

You also get your wish of us tracking start and end time for each step. I bet @Marvin9 could make good use of this information in the future.

I did my very best to preserve the general spirit and structure of #2940.

This is successfully having all steps fail on their first error with no retries by default, but you can make a step retry after an error by explicitly configuring a error threshold > 1.

As long a step says it is running, it will be retried ad infinitum by default, but the argocd-update step has a lower default of 10m, you can make any step have a finite timeout if you explicitly configure a non-nil and non-zero duration.

As a bonus, I found and fixed two other bugs related to step execution:

* Step execution engine was only ever returning health checks on success. This meant that, if in a single run, step 1 succeeded and had health checks to return, but step 2 was running and needed to be requeued, those health checks step 1 wanted to register never make it back to the reconciler.

* The reconciler was only ever doing anything with the healthchecks on promotion success. A single promotion succeeds a maximum of once. This meant that even if the previous bug hadn't existed, the reconciler would ignore all health checks except those returned from its _final_ call out the the engine. Other steps could have succeeded in previous calls and their health checks would be nowhere to be found.

  We never happened to notice this, because the only step that registers health checks currently is argocd-update and that tends to be the _last_ step in a promo process, which meant things always have worked out in its favor despite these bugs.
  
This still needs corresponding doc updates, but I will tackle them in a follow-up.


api/v1alpha1/promotion_types_test.go
```diff
@@ -2,16 +2,53 @@ package v1alpha1
...
+		want     *time.Duration
+	}{
+		{
+			name:     "retry is nil",
+			retry:    nil,
+			fallback: ptr.To(time.Hour),
+			want:     ptr.To(time.Hour),
+		},
+		{
+			name:     "threshold is not set",
```
hiddeco, line 26:
```suggestion
			name:     "timeout is not set",
```

api/v1alpha1/promotion_types_test.go
```diff
@@ -2,16 +2,53 @@ package v1alpha1
...
+			retry:    &PromotionStepRetry{},
+			fallback: ptr.To(time.Hour),
+			want:     ptr.To(time.Hour),
+		},
+		{
+			name: "threshold is set",
```
hiddeco, line 32:
```suggestion
			name: "timeout is set",
```

internal/directives/promotions.go
```diff
@@ -96,16 +101,29 @@ type PromotionStep struct {
...
+	return s.Retry.GetTimeout(fallback)
+}
+
+// GetErrorThreshold returns the number of consecutive times the provided runner
+// must fail to execute the step (for any reason) before retries are abandoned
+// and the entire Promotion is marked as failed. If the runner is a
+// RetryableStepRunner, its threshold is used as the default. Otherwise, the
+// default is 1.
+func (s *PromotionStep) GetErrorThreshold(runner any) uint32 {
+	fallback := uint32(1)
```
hiddeco, line 122:
```suggestion
	const fallback = uint32(1)
```

Micro optimization.

internal/directives/simple_engine_promote.go
```diff
@@ -77,94 +81,129 @@ func (e *SimpleEngine) executeSteps(
...
-				// Update the result to indicate that the step is still running.
-				result.Status = kargoapi.PromotionPhaseRunning
-				result.Message = message.String()
-
-				// Swallow the error if the step failed, as we are still
-				// retrying it.
-				err = nil
+		if stepExecMeta.Status == kargoapi.PromotionPhaseSucceeded {
+			stepExecMeta.FinishedAt = ptr.To(metav1.Now())
+			stepExecMeta.ErrorCount = 0
```
hiddeco, line 127: Do we need to reset the error count? I think it may be valuable to a user to know that a step every time succeeds after `n` erroneous attempts.

internal/directives/simple_engine_promote.go
```diff
@@ -77,94 +81,129 @@ func (e *SimpleEngine) executeSteps(
...
+			if healthCheck := result.HealthCheckStep; healthCheck != nil {
+				healthChecks = append(healthChecks, *healthCheck)
 			}
+			continue // Move on to the next step
 		}
 
-		// Update the state with the step output, regardless of the result.
-		state[step.Alias] = result.Output
+		// Treat errors and logical failures the same for now.
+		// TODO(krancour): These may be handled differently in the future.
```
hiddeco, line 135: I think what you are aiming at here is that we eventually may have terminating errors etc. But I am not sure we will remember this ourselves in 1-2 months, so I would add a "how" here.

hiddeco (commented):
First sweep with a few minor nits. Overall, it looks absolutely great.

internal/directives/simple_engine_promote.go
krancour: That's a good idea.

```suggestion
```

internal/directives/simple_engine_promote.go
krancour:
```suggestion
		// TODO(krancour): In the future, we should fail without retry for a logical failures and unrecoverable errors and retry only those with a chance of recovery.
```

internal/directives/promotions.go
```diff
@@ -96,16 +101,29 @@ type PromotionStep struct {
...
+	return s.Retry.GetTimeout(fallback)
+}
+
+// GetErrorThreshold returns the number of consecutive times the provided runner
+// must fail to execute the step (for any reason) before retries are abandoned
+// and the entire Promotion is marked as failed. If the runner is a
+// RetryableStepRunner, its threshold is used as the default. Otherwise, the
+// default is 1.
+func (s *PromotionStep) GetErrorThreshold(runner any) uint32 {
+	const fallback = uint32(1)
```
krancour, line 122: This can't be const. It potentially takes on a different value a few lines below.

```suggestion
	fallback := uint32(1)
```

hiddeco (approved):
Needs DCO and linter fixes, but otherwise LGTM. Great work @krancour 🙇

Marvin9:
> You also get your wish of us tracking start and end time for each step. I bet @Marvin9 could make good use of this information in the future.

definitely following on this.. took "x" duration or running since "y" duration will add up value

hiddeco:
Actually, think this also needs a slight update of documentation.

krancour:
@hiddeco I mentioned docs will be in a follow up. I'll fix DCO and linter errors. This is what I get for editing directly on github on my phone!

PR #3069, opened 2024-12-05
chore: clear app status op state

krancour:
cc @jessesuen

### DIFF
--- a/internal/directives/argocd_updater.go
+++ b/internal/directives/argocd_updater.go
@@ -536,6 +536,15 @@ func (a *argocdUpdater) syncApplication(
 	for _, source := range app.Spec.Sources {
 		app.Operation.Sync.Revisions = append(app.Operation.Sync.Revisions, source.TargetRevision)
 	}
+	// TODO(krancour): This is a workaround for the Argo CD Application controller
+	// not handling this correctly itself. It is Argo CD's API server that usually
+	// handles this, but we are bypassing the API server here.
+	//
+	// See issue: https://github.com/argoproj/argo-cd/issues/20875
+	//
+	// We can remove this hack once the issue is resolved and all Argo CD versions
+	// without the fix have reached their EOL.
+	app.Status.OperationState = nil
 
 	// Patch the Argo CD Application.
 	if err := a.argoCDAppPatchFn(ctx, stepCtx, app, func(src, dst unstructured.Unstructured) error {


jessesuen (approved):
In principle, Kargo (and even Argo CD API server) should not update Application status when creating the operation. In fact, if status subresources were enabled for the CRD, this would not even be possible.

I take the blame for this, having wrote the argo-cd code that does exactly this.

We'll have to live with ugly hack until Argo CD improves its behavior.

hiddeco (approved):

stephaneetje:
Hello, i just tried the patch in v1.1.0, it is still not working for me, it syncs before it sees the outofsync ,so i need another manual sync... 
FYI, i first tried a patch from argocd that was working : https://github.com/argoproj/argo-cd/pull/20915
Then i took it of to try kargo 1.1 and see if it was fixing the problem, which is not.

krancour:
@stephaneetje we _are_ clearing it only when initiating an operation.

I wonder if you aren't actually running into #3082 now.

stephaneetje:
@krancour i don't think so as i do see the sync. The problem is exactly as before i used argocd's patch. It starts the sync Before we see the app OutOfSync.

> we are clearing it only when initiating an operation

I was thinking more of the time ordering than just the condition. I think it is cleared too early, it should be cleared "when sync is getting started"

krancour:
That is what we are doing. When we update app.Operation, we clear app.Status.OperationState.

internal/directives/argocd_updater.go
```diff
@@ -536,6 +536,15 @@ func (a *argocdUpdater) syncApplication(
...
 	}
+	// TODO(krancour): This is a workaround for the Argo CD Application controller
+	// not handling this correctly itself. It is Argo CD's API server that usually
+	// handles this, but we are bypassing the API server here.
+	//
+	// See issue: https://github.com/argoproj/argo-cd/issues/20875
+	//
+	// We can remove this hack once the issue is resolved and all Argo CD versions
+	// without the fix have reached their EOL.
+	app.Status.OperationState = nil
```
andrii-korotkov-verkada, line 547: @krancour, shall we also do
```
dst.Object["status"] = src.Object["status"]
```
below?

krancour: Ugh. Missed that, although it probably shouldn't be status, but _just_ status.operationState. As previously discussed, we really don't want to be messing with status at all, so we need to be very surgical about what we patch here.

andrii-korotkov-verkada: Was there a fix done for this?

krancour: Look down 👇

stephaneetje: I just tried it and did not change anything, I still see the sync begining Before i see the OutOfstatus

PR #3072, opened 2024-12-05
chore(ui): connectrpc migrate to version 2

Marvin9:
Migration guide that was followed: https://github.com/connectrpc/connect-es/blob/main/MIGRATING.md

Verified with basic operations and the area that code change touches and fix all the potential issues.

Notable changes:

- No more manual patch to generated proto classes. It was used to get `Date` type from `google.protobuf.Timestamp` type. Now its explicit function `k8sApiMachineryTimestampDate`.
- API was changed to generate query.. this still needs better way to figure out how to do it
- There is now `PlainMessage` equivalent homegrown implementation that just works


Marvin9:
@rpelczar can I get some eyes on it please?

ui/src/features/common/commit-info.tsx
```diff
@@ -1,6 +1,7 @@
 import { GitCommit } from '@ui/gen/v1alpha1/generated_pb';
+import { PlainMessage } from '@ui/utils/connectrpc-extension';
 
-export const CommitInfo = ({ commit }: { commit: GitCommit }) => (
+export const CommitInfo = ({ commit }: { commit: PlainMessage<GitCommit> }) => (
```
rpelczar, line 4: I understand the purpose of having PlainMessage, but in some places, it seems redundant.
This is an example.

ui/src/utils/connectrpc-extension.ts
```diff
@@ -0,0 +1,26 @@
+import { Timestamp, timestampDate } from '@bufbuild/protobuf/wkt';
+
+import { Time } from '@ui/gen/k8s.io/apimachinery/pkg/apis/meta/v1/generated_pb';
+
+export const k8sApiMachineryTimestampDate = (t?: Time | PlainMessage<Time> | string) => {
```
rpelczar, line 5: Can we find a better name for it? Do we use it for all dates or just selected ones?

ui/src/utils/connectrpc-extension.ts
rpelczar, line 1: Can we name it `connectrpc-utils.ts`?

rpelczar (changes requested):

rpelczar (approved):
LGTM!

krancour (approved):

PR #3074, opened 2024-12-05
chore(controller): make backoff configurable for tests

hiddeco:
Primarily to be able to overwrite it in tests, as the previous approach caused an unnecessary delay for the unit tests.


krancour (approved):
Very nice improvement!

Issue #2860, opened 2024-10-28
Add Kargo Promotion Step to Remove File and/or Directory


BenjaminNeale-Heritage:
# Checklist

* [X] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature
The promotion directives/steps introduced in Kargo v0.9 included a 'copy' step [1].  The 'copy' step would pair well with a 'remove' step to allow individual files and/or directories to be removed.  The 'git-clear' step [2] is too broad for this purpose.

# Motivation
With the potential that Kargo Render will be archived in the near future, I am hoping to use Kargo to manager cluster infrastructure configuration.  I currently use Kargo Render to generate Git branches for each cluster following the Rendered Manifest Pattern.  Each of these Git branches contains all of the infrastructure configuration (e.g. cert-manager, Kargo, Argo CD, etc) for a single cluster.  An Argo CD ApplicationSet resource on a cluster points to matching branch.

Directory structure of `main` branch.
```
├─ {COMPONENT_NAME}/
│  └─ resources/
│     ├─ base/
│     ├─ overlays/
│        ├─ {CLUSTER_NAME}/
│        └─ ...
└─ ...
```

Example of `main` branch.
```
├─ argo-cd/
│  └─ resources/
│     ├─ base/
│     ├─ overlays/
│        ├─ non-production/
│        ├─ production-01/
│        ├─ production-02/
│        └─ ...
├─ cert-manager/
├─ kargo/
└─ ...
```

Directory structure of a cluster branch (e.g. `cluster/non-production` branch).
```
├─ COMPONENT_NAME/
└─ ...
```

Example of cluster branch.
```
└─ components/
   ├─ argo-cd
   ├─ cert-manager
   ├─ kargo
   └─ ...
```

If Kargo is configured with a separate pipeline for each component then the 'git-clear' step available within Kargo cannot be used because all of the other components are removed from the target branch containing the rendered manifests when one of the components changes.

Here is Kargo Render config that currently achieves the same outcome.
```
configVersion: v1alpha1
branchConfigs:

  # =============== Non-Production Cluster ===============
  - name: cluster/non-production
    appConfigs:
      argo-cd-resources:
        configManagement:
          path: argo-cd/resources/overlays/non-production
        outputPath: components/argo-cd
      cert-manager-resources:
        configManagement:
[...]

PR #2987, opened 2024-11-23
feat(ui): 'http' and 'yaml-update' promotion directives support

Marvin9:
fixes https://github.com/akuity/kargo/issues/2986

PR #3086, opened 2024-12-07
feat: add promotion step to remove a file/directory

nitishfy:
Fixes https://github.com/akuity/kargo/issues/2860

### DIFF
--- a/docs/docs/35-references/10-promotion-steps.md
+++ b/docs/docs/35-references/10-promotion-steps.md
@@ -312,14 +312,8 @@ steps:
 
 #### `delete` Example
 
-The most common usage of this step is to clean up unwanted files
-or directories before proceeding to subsequent steps in the promotion
-process. For instance, you might use the delete step to remove
-intermediate configurations or artifacts generated in earlier steps.
-
-Consider a scenario where a Stage combines content from two working trees
-to render Stage-specific manifests. If temporary files are generated during
-this process, you can use the delete step to clean them up:
+One, common usage of this step is to remove intermediate files
+produced by the promotion process prior to a `git-commit` step:
 
 ```yaml
 vars:
@@ -330,21 +324,20 @@ steps:
   config:
     repoURL: ${{ vars.gitRepo }}
     checkout:
-    - commit: ${{ commitFrom(vars.gitRepo, warehouse("base")).ID }}
+    - commit: ${{ commitFrom(vars.gitRepo) }}
       path: ./src
-    - commit: ${{ commitFrom(vars.gitRepo, warehouse(ctx.stage + "-overlay")).ID }}
-      path: ./overlay
     - branch: stage/${{ ctx.stage }}
       create: true
       path: ./out
-- uses: copy
-  config:
-    inPath: ./overlay/stages/${{ ctx.stage }}/kustomization.yaml
-    outPath: ./src/stages/${{ ctx.stage }}/kustomization.yaml
+
+# Steps that produce intermediate files in ./out...
+
 - uses: delete
   config:
-    path: ./overlay/stages/${{ ctx.stage }}/temp-file.yaml
-# Render manifests to ./out, commit, push, etc...
+    path: ./out/unwanted/file
+- uses: git-commit
+  config:
+    path: ./out
 ```
 
 ### `kustomize-set-image`
--- a/docs/docs/35-references/10-promotion-steps.md
+++ b/docs/docs/35-references/10-promotion-steps.md
@@ -302,7 +302,7 @@ steps:
 
 ### `delete`
 
-`delete` deletes the files or the directory,
+`delete` deletes a file or directory.
 
 #### `delete` Configuration
 
--- a/docs/docs/35-references/10-promotion-steps.md
+++ b/docs/docs/35-references/10-promotion-steps.md
@@ -300,6 +300,53 @@ steps:
 # Render manifests to ./out, commit, push, etc...
 ```
 
+### `delete`
+
+`delete` deletes the files or the directory,
+
+#### `delete` Configuration
+
+| Name      | Type | Required | Description                              |
+|-----------|------|----------|------------------------------------------|
+| `path`    | `string` | Y | Path to the file or directory to delete. |
+
+#### `delete` Example
+
+The most common usage of this step is to clean up unwanted files
+or directories before proceeding to subsequent steps in the promotion
+process. For instance, you might use the delete step to remove
+intermediate configurations or artifacts generated in earlier steps.
+
+Consider a scenario where a Stage combines content from two working trees
+to render Stage-specific manifests. If temporary files are generated during
+this process, you can use the delete step to clean them up:
+
+```yaml
+vars:
+- name: gitRepo
+  value: https://github.com/example/repo.git
+steps:
+- uses: git-clone
+  config:
+    repoURL: ${{ vars.gitRepo }}
+    checkout:
+    - commit: ${{ commitFrom(vars.gitRepo, warehouse("base")).ID }}
+      path: ./src
+    - commit: ${{ commitFrom(vars.gitRepo, warehouse(ctx.stage + "-overlay")).ID }}
+      path: ./overlay
+    - branch: stage/${{ ctx.stage }}
+      create: true
+      path: ./out
+- uses: copy
+  config:
+    inPath: ./overlay/stages/${{ ctx.stage }}/kustomization.yaml
+    outPath: ./src/stages/${{ ctx.stage }}/kustomization.yaml
+- uses: delete
+  config:
+    path: ./overlay/stages/${{ ctx.stage }}/temp-file.yaml
+# Render manifests to ./out, commit, push, etc...
+```
+
 ### `kustomize-set-image`
 
 `kustomize-set-image` updates the `kustomization.yaml` file in a specified
--- a/internal/directives/file_deleter.go
+++ b/internal/directives/file_deleter.go
@@ -1,0 +1,151 @@
+package directives
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+	"github.com/xeipuuv/gojsonschema"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+)
+
+func init() {
+	builtins.RegisterPromotionStepRunner(newFileDeleter(), nil)
+}
+
+// fileDeleter is an implementation of the PromotionStepRunner interface that
+// deletes a file or directory.
+type fileDeleter struct {
+	schemaLoader gojsonschema.JSONLoader
+}
+
+// newFileDeleter returns an implementation of the PromotionStepRunner interface
+// that deletes a file or directory.
+func newFileDeleter() PromotionStepRunner {
+	r := &fileDeleter{}
+	r.schemaLoader = getConfigSchemaLoader(r.Name())
+	return r
+}
+
+// Name implements the PromotionStepRunner interface
+func (f *fileDeleter) Name() string {
+	return "delete"
+}
+
+// RunPromotionStep implements the PromotionStepRunner interface.
+func (f *fileDeleter) RunPromotionStep(
+	ctx context.Context,
+	stepCtx *PromotionStepContext,
+) (PromotionStepResult, error) {
+	// Validate the configuration against the JSON Schema.
+	if err := validate(f.schemaLoader, gojsonschema.NewGoLoader(stepCtx.Config), f.Name()); err != nil {
+		return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored}, err
+	}
+
+	// Convert the configuration into a typed object.
+	cfg, err := ConfigToStruct[DeleteConfig](stepCtx.Config)
+	if err != nil {
+		return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored},
+			fmt.Errorf("could not convert config into %s config: %w", f.Name(), err)
+	}
+
+	return f.runPromotionStep(ctx, stepCtx, cfg)
+}
+
+func (f *fileDeleter) runPromotionStep(
+	_ context.Context,
+	stepCtx *PromotionStepContext,
+	cfg DeleteConfig,
+) (PromotionStepResult, error) {
+	absPath, err := f.resolveAbsPath(stepCtx.WorkDir, cfg.Path)
+	if err != nil {
+		return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored},
+			fmt.Errorf("could not secure join path %q: %w", cfg.Path, err)
+	}
+
+	symlink, err := f.isSymlink(absPath)
+	if err != nil {
+		return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored}, err
+	}
+
+	if symlink {
+		err = os.Remove(absPath)
+		if err != nil {
+			return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored}, err
+		}
+	} else {
+		// Secure join the paths to prevent path traversal.
+		pathToDelete, err := securejoin.SecureJoin(stepCtx.WorkDir, cfg.Path)
+		if err != nil {
+			return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored},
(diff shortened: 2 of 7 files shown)


internal/directives/file_deleter_test.go
```diff
@@ -0,0 +1,88 @@
...
+			assertions: func(t *testing.T, workDir string, result PromotionStepResult, err error) {
+				assert.NoError(t, err)
+				assert.Equal(t, PromotionStepResult{Status: kargoapi.PromotionPhaseSucceeded}, result)
+
+				_, statErr := os.Stat(filepath.Join(workDir, "dirToDelete"))
+				assert.ErrorIs(t, statErr, os.ErrNotExist)
+			},
+		},
+		{
+			name: "fails for non-existent path",
```
hiddeco, line 61: In addition to these tests, can you add:

- One which removes a symlink (should remove the literal symlink, and not the file).
- One which removes a file or directory within a symlink (should traverse into the symlink and remove it).

krancour:
This needs docs.

krancour:
@nitishfy we're looking to have 1.2 work done by 12/20 with plans to release after the winter break. Are you able to implement feedback before then?

internal/directives/file_deleter_test.go
```diff
@@ -0,0 +1,120 @@
...
+			cfg: DeleteConfig{
+				Path: "nonExistentFile.txt",
+			},
+			assertions: func(t *testing.T, _ string, result PromotionStepResult, err error) {
+				assert.Error(t, err)
+				assert.Equal(t, PromotionStepResult{Status: kargoapi.PromotionPhaseErrored}, result)
+			},
+		},
+		{
+			name: "removes symlink only",
```
nitishfy, line 76:
```suggestion
			name: "removes symlink only",
```

@hiddeco I tried debugging to understand why is this test case failing even though the path specified is correct. After debugging i found that, `SecureJoin()` is actually joining the file `input/input.txt` that is linked by the `symlink.txt` instead of joining the file `symlink.txt`, Following is the error message

```
=== RUN   Test_fileDeleter_runPromotionStep
--- FAIL: Test_fileDeleter_runPromotionStep (120.72s)
=== RUN   Test_fileDeleter_runPromotionStep/removes_symlink_only
path to delete: /tmp/Test_fileDeleter_runPromotionStepremoves_symlink_only1504252788/001/input/input.txt
    file_deleter_test.go:100: 
        	Error Trace:	/home/nitish/Documents/Work/Open-Source/kargo/internal/directives/file_deleter_test.go:100
        	            				/home/nitish/Documents/Work/Open-Source/kargo/internal/directives/file_deleter_test.go:115
        	Error:      	Received unexpected error:
        	            	stat /tmp/Test_fileDeleter_runPromotionStepremoves_symlink_only1504252788/001/input.txt: no such file or directory
        	Test:       	Test_fileDeleter_runPromotionStep/removes_symlink_only
    --- FAIL: Test_fileDeleter_runPromotionStep/removes_symlink_only (120.72s)
```

nitishfy (commented):
cc @hiddeco PTAL

internal/directives/file_deleter_test.go
nitishfy: The first test case has been added. I would like to confirm something for the second test case you'd like to be added:

The test shouldn't remove the actual symlink but only the target file(inside some other directory) pointed by a symlink directory. Correct?

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,139 @@
...
+	return f.runPromotionStep(ctx, stepCtx, cfg)
+}
+
+func (f *fileDeleter) runPromotionStep(
+	_ context.Context,
+	stepCtx *PromotionStepContext,
+	cfg DeleteConfig,
+) (PromotionStepResult, error) {
+	// join with a regular join
+	absPath, err := checkAbsPath(stepCtx.WorkDir, cfg.Path)
```
hiddeco, line 60:
```suggestion
	absPath, err := resolveAbsPath(stepCtx.WorkDir, cfg.Path)
```

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return f.runPromotionStep(ctx, stepCtx, cfg)
+}
+
+func (f *fileDeleter) runPromotionStep(
+	_ context.Context,
+	stepCtx *PromotionStepContext,
+	cfg DeleteConfig,
+) (PromotionStepResult, error) {
+	// join with a regular join
```
hiddeco, line 59:
```suggestion
```

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	return PromotionStepResult{Status: kargoapi.PromotionPhaseSucceeded}, nil
+}
+
+func checkSymlink(path string) bool {
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return false
+	}
```
hiddeco, line 91: This should consider that the file can already be deleted, in which case I would not expect an error to be thrown nor a removal to be attempted.

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,139 @@
...
+
+func removePath(path string) error {
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return err
+	}
+
+	if fi.Mode()&os.ModeSymlink != 0 {
+		return os.Remove(path)
+	}
```
hiddeco, line 132: This does not do anything anymore.

internal/directives/file_deleter_test.go
hiddeco: Given path `foo/file.txt`, where `foo` is a symlink to `bar/`. It should delete `bar/file.txt`.

internal/directives/schemas/delete-config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "DeleteConfig",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["Path"],
+  "properties": {
+    "Path": {
```
krancour, line 8: This should be lowercase.

krancour:
I'm not considering it a blocker, but because JSON schema can be tricky to mentally parse, unit tests that assert validations work as intended would be a nice-to-have.

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,139 @@
...
+		if err = removePath(pathToDelete); err != nil {
+			return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored},
+				fmt.Errorf("failed to delete %q: %w", cfg.Path, sanitizePathError(err, stepCtx.WorkDir))
+		}
+	}
+
+	return PromotionStepResult{Status: kargoapi.PromotionPhaseSucceeded}, nil
+}
+
+func checkSymlink(path string) bool {
```
krancour, line 87: As there are quite a number of step runners in this package, and given that we're very likely to break this package up in the future as progress is made on custom/third-party steps, I'm of a mind to make small helper functions such as this one into methods of the runner. I think it will make breaking everything up in the future easier when we can see the following and know that the functionality is used exclusively by a given step runner and not shared among several:

```suggestion
func (f *fileDeleter) checkSymlink(path string) bool {
```

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,139 @@
...
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return false
+	}
+
+	return fi.Mode()&os.ModeSymlink != 0
+}
+
+func checkAbsPath(workDir string, path string) (string, error) {
```
krancour, line 96: Same rationale as [here](https://github.com/akuity/kargo/pull/3086/files#r1888956261).

```suggestion
func (f *fileDeleter) checkAbsPath(workDir string, path string) (string, error) {
```

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,139 @@
...
+	cfg DeleteConfig,
+) (PromotionStepResult, error) {
+	// join with a regular join
+	absPath, err := checkAbsPath(stepCtx.WorkDir, cfg.Path)
+	if err != nil {
+		return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored},
+			fmt.Errorf("could not secure join path %q: %w", cfg.Path, err)
+	}
+
+	if checkSymlink(absPath) {
```
krancour, line 66: Corresponds to the change proposed [here](https://github.com/akuity/kargo/pull/3086/files#r1888956261).

```suggestion
	if f.isSymlink(absPath) {
```

internal/directives/file_deleter.go
krancour: Also if the error _isn't_ the the file/dir/symlink isn't found, it's not right to swallow it and just say "not a symlink."

This method should return such an error.

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,139 @@
...
+	// Get the relative path from base to the requested path
+	// If the requested path tries to escape, this will return
+	// an error or a path starting with "../"
+	relPath, err := filepath.Rel(absBase, absPath)
+	if err != nil {
+		return "", err
+	}
+
+	// Check if path attempts to escape
+	if strings.HasPrefix(relPath, "..") {
```
krancour, line 117: This isn't the right way to check for this.

`sub-dir/../../something-i-should-not-access` accomplishes the same thing as `../something-i-should-not-access`, but the former isn't caught by this logic.

internal/directives/file_deleter.go
krancour:
```suggestion
	absPath, err := f.resolveAbsPath(stepCtx.WorkDir, cfg.Path)
```

internal/directives/file_deleter.go
krancour: Nevermind... seeing where `relPath` comes from, containing "..", but not starting with it is an impossibility.

internal/directives/file_deleter.go
nitishfy: @krancour why would you want the function to have a receiver in this case?

Marvin9:
if you could please update UI code as well - read [this PR](https://github.com/akuity/kargo/pull/2987) for reference (skip the icons)

internal/directives/file_deleter.go
nitishfy: the error returned will be of type `PathError` according to the documentation. Can you provide some more context on why should we specifically care about this error?

krancour: > why should we specifically care about this error

It's more like, "Why _shouldn't_ we?"

As a general rule, you handle what errors you can -- in this case that entails disregarding errors that are _specifically_ that no file/dir exsists at that path -- and you surface any other problem you weren't anticipating or don't know how to handle.

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,142 @@
...
+			fmt.Errorf("could not secure join path %q: %w", cfg.Path, err)
+	}
+
+	symlink, err := f.isSymlink(absPath)
+
+	/* TODO
+	if err != nil {
+
+	}
+	*/
```
krancour, line 71: Please address this todo.

internal/directives/file_deleter.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if symlink {
+		err = os.Remove(absPath)
+		if err != nil {
+			return PromotionStepResult{}, err
```
krancour, line 76:
```suggestion
			return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored}, err
```

internal/directives/schemas/delete-config.json
krancour: You marked this as resolved, but did not fix the issue.

nitishfy: This is resolved now.

internal/directives/file_deleter.go
nitishfy: Got it, please see if this is the correct way to handle this error now.

krancour:
@nitishfy thanks for applying all the feedback!

For docs, just add a new section to `docs/docs/35-references/10-promotion-steps.md`. Follow the format of existing step docs as closely as possible.

I would suggest the best place for this would be under the docs for the copy step.

@hiddeco, apart from docs, I think the only unresolved feedback was yours. Leaving final approval up to you.

internal/directives/file_deleter.go
nitishfy: marking this as resolved

hiddeco:
Apart from the documentation, this also looks good to me.

nitishfy:
Awesome, i'll go ahead and add the docs now.

nitishfy:
Ready for the review.

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -300,6 +300,53 @@ steps:
 # Render manifests to ./out, commit, push, etc...
 ```
 
+### `delete`
+
+`delete` deletes the files or the directory,
```
hiddeco, line 305:
```suggestion
`delete` deletes a file or directory.
```

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -300,6 +300,53 @@ steps:
...
+  config:
+    inPath: ./overlay/stages/${{ ctx.stage }}/kustomization.yaml
+    outPath: ./src/stages/${{ ctx.stage }}/kustomization.yaml
+- uses: delete
+  config:
+    path: ./overlay/stages/${{ ctx.stage }}/temp-file.yaml
+# Render manifests to ./out, commit, push, etc...
+```
+
 ### `kustomize-set-image`
```
krancour, line 350: I'm trying to think of a more realistic example. I've noticed a tendency among users to put too much stock in examples, supposing them to be recommended patterns, which makes it important for the examples to be relatable and realistic...

At the same time, I don't think I've personally run into a need for this step. Looking back to #2860, the original issue was addressable using the existing git-clear step, so unfortunately, there is no inspiration to be found there.

I have absolutely no doubt that this step is generally useful and a completely necessary complement to the copy step... just need to find a realistic example for showcasing it.

I'll think on this a bit and if you don't mind, I'll update the example directly when I come up with something.

Ideas are welcome, of course.

krancour: Other than this, docs look great. We're at like 99% on this PR.

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -300,6 +300,53 @@ steps:
...
+      path: ./out
+- uses: copy
+  config:
+    inPath: ./overlay/stages/${{ ctx.stage }}/kustomization.yaml
+    outPath: ./src/stages/${{ ctx.stage }}/kustomization.yaml
+- uses: delete
+  config:
+    path: ./overlay/stages/${{ ctx.stage }}/temp-file.yaml
+# Render manifests to ./out, commit, push, etc...
+```
```
krancour, line 348: Thought about this and, ironically, the solution is probably to be more hand-wavy about how the file/dir we're deleting came to be:

~~~suggestion
One, common usage of this step is to remove intermediate files
produced by the promotion process prior to a `git-commit` step:

```yaml
vars:
- name: gitRepo
  value: https://github.com/example/repo.git
steps:
- uses: git-clone
  config:
    repoURL: ${{ vars.gitRepo }}
    checkout:
    - commit: ${{ commitFrom(vars.gitRepo) }}
      path: ./src
    - branch: stage/${{ ctx.stage }}
      create: true
      path: ./out

# Steps that produce intermediate files in ./out...

- uses: delete
  config:
    path: ./out/unwanted/file
- uses: git-commit
  config:
    path: ./out
```
~~~

docs/docs/35-references/10-promotion-steps.md
krancour: See https://github.com/akuity/kargo/pull/3086#discussion_r1893250643

docs/docs/35-references/10-promotion-steps.md
krancour: LGTM after this.

krancour: I'm going to take the liberty of committing this suggestion since we're trying really hard to get v1.2 buttoned up.

krancour (approved):
@nitishfy thank you for all the hard work that went into this!

PR #3100, opened 2024-12-09
feat: factor optional "soak time" into freight availability

hiddeco:
Fixes: #2069

With this addition, it is possible to configure a duration for which the Freight of an upstream Stage has to be verified, effectively allowing it to "soak" in this Stage.

```yaml
kind: Stage
metadata:
  name: prod
  namespace: guestbook
spec:
  # ...omitted for brevity 
  requestedFreight:
  - origin:
      kind: Warehouse
      name: guestbook
    sources:
      stages: ["staging"]
      # Freight needs to be verified for 12 hours in "staging"
      # before being allowed to be auto-Promoted to "prod".
      verifiedFor: 12h0m
```


krancour:
This looks like it doesn't comprehensively address the question of a piece of Freight's overall _availability_ to a given Stage. The required soak time isn't, for instance, being enforced in the case of a manual Promotion.

I realize that based on wording of this PR's title and issue #2069, that the scope might have been narrowly about _auto-promotion_, in which case this PR checks that box perfectly, but imho, the soak time should also be enforced for manual promotions, and if someone absolutely needs to manually Promote a piece of Freight whose soak time hasn't elapsed, they should add a manual Approval to it first. (I don't think soak time should apply to manual approvals.)

cc @jessesuen for his opinion on what the scope is supposed to be.

@hiddeco since you're out for a bit, I will take up any remaining work on this pending @jessesuen's clarification.

jessesuen:
> The required soak time isn't, for instance, being enforced in the case of a manual Promotion.
> if someone absolutely needs to manually Promote a piece of Freight whose soak time hasn't elapsed, they should add a manual Approval to it first. (I don't think soak time should apply to manual approvals.)

Thanks for catching this. I agree. We can consider this feature similar to a policy is enforced regardless of auto vs. manual promotion and it would be good to have the guard rail in place for both humans and machines. Manual approval seems like the appropriate way to go around it.

krancour:
Thanks @jessesuen. I think this is relatively easy to resolve.

krancour:
I have this well underway now. I should have something ready late tomorrow.

krancour:
The last four commits I added move Freight availability logic to one central place and it's applied consistently whether you're promoting, auto-promoting, or just querying for available Freight.

I'm changing this to a draft, because there is still an outstanding issue here that I discovered and @jessesuen and I discussed offline. This will currently count a piece of Freight as available to a given Stage if its successful verification merely occurred long-enough ago, which turns out to be insufficient given the following scenario:

Given a pipeline test --> uat, if Freight A were promoted to test and successfully verified and almost immediately afterward, Freight B were promoted to test, Freight A will still become available to uat after the specified time has elapsed, _but Freight A didn't "soak" adequately in test._

This is going to require additional bookkeeping to keep track of how long Freight verified in a given Stage actually _spent_ in that Stage.

My best thinking on how to achieve this, which needs to be sanity checked by @hiddeco:

1. Make Freight status track two new things:
    1. What Stages _currently_ have it and when it got there.
    1. In verification records, track the max duration the Freight spent in that Stage.

1. Make Stage reconciliation look at the status of every piece of Freight in the current Freight collection (top of the history stack) and, only if necessary, update the above information accordingly.

    In the case where a new Freight collection is pushed on to the top of the Freight history stack, updates need to be applied to the union of all Freight in the old top Freight collection and the new one.

    If a piece of Freight's status is updated to reflect that it is no longer in the Stage, the time it spent there should be calculated. If that Freight's status has a verification record for that Stage and the calculated duration recently spent in that Stage exceeds the currently recorded max duration, the max duration should be updated.

1. Freight is available to a given Stage under any of the following conditions:

    1. Freight is approved for the Stage
    1. Freight is verified for the Stage _with no soak time criteria_
    1. Freight is verified for the Stage the recorded max time spent in that Stage meets or exceeds soak time criteria
    1. Freight is verified for the Stage, currently in the Stage, and the elapsed time since it entered the Stage meets or exceeds soak time criteria

A tangential benefit to this is that with every piece of Freight having a record of what Stages are currently using it, the UI, which I believe is currently crunching that information itself, will no longer have to.

@hiddeco might be able to refine this idea or suggest an even better alternative.

krancour:
This is working e2e and now conforms to the requirements that were clarified by @jessesuen in our offline discussions.

krancour:
The merge conflicts are all with codegen. I'll resolve those once this is approved because a rebase on main is probably the best way and I don't want to force push while two maintainers are potentially working on this.

api/v1alpha1/freight_types.go
```diff
@@ -172,10 +174,24 @@ type FreightStatus struct {
...
 // VerifiedStage describes a Stage in which Freight has been verified.
 type VerifiedStage struct {
 	// VerifiedAt is the time at which the Freight was verified in the Stage.
 	VerifiedAt *metav1.Time `json:"verifiedAt,omitempty" protobuf:"bytes,1,opt,name=verifiedAt"`
+	// LongestCompletedSoak represents the longest definite time interval wherein
+	// the Freight was in CONTINUOUS use by the Stage. This value is updated as
+	// Freight EXITS the Stage. If the Freight is currently in use by the Stage,
+	// the time elapsed since the Freight ENTERED the Stage is its current soak
+	// time, which may exceed the value of this field.
+	LongestCompletedSoak *metav1.Duration `json:"longestSoak,omitempty" protobuf:"bytes,2,opt,name=longestSoak"`
```
hiddeco, line 194: I can not help but feel the placement of this is odd, and if we would be better off keeping track of "entering" / "leaving" history to compute this value.

api/v1alpha1/freight_helpers_test.go
```diff
@@ -2,16 +2,38 @@ package v1alpha1
...
+	if !ok {
+		return nil
+	}
+	currentStages := make([]string, len(freight.Status.CurrentlyIn))
+	var i int
+	for stage := range freight.Status.CurrentlyIn {
+		currentStages[i] = stage
+		i++
+	}
+	return currentStages
```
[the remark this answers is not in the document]
krancour: > idiomacy

_What's you just call me??_

j/k

And agree.

api/v1alpha1/freight_types.go
krancour: I had this exact thought, but had trouble justify all the additional bookkeeping because we really only care about this info for Stages in which the Freight has been verified.

Were we to track entering/leaving separately, it'd also be one more thing to clean up during finalization of a deleted Freight -- which isn't a _huge_ deal.

Consider me completely amenable to doing this differently if you can suggest a way that isn't too onerous and doesn't create an accumulation of unimportant data.

api/v1alpha1/stage_helpers.go
```diff
@@ -97,6 +99,92 @@ func GetStage(
...
+	availableFreight = slices.CompactFunc(availableFreight, func(lhs, rhs Freight) bool {
+		return lhs.Name == rhs.Name
+	})
+
+	return availableFreight, nil
+}
+
+// IsFreightAvailable answers whether the specified Freight is available to the
+// Stage.
+func (s *Stage) IsFreightAvailable(freight *Freight) bool {
```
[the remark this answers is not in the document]
krancour: Nice catch.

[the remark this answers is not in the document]
krancour: Oh. Sorry. I misread this.

There wasn't any specific reason.

It's a function for which Stage and Freight are equally important arguments.

I think I could have gone either way on this and this is just where I landed.

If you think there's a compelling reason to do it the other way, I'm unopposed.

api/v1alpha1/freight_helpers_test.go
```diff
@@ -26,10 +26,9 @@ func freightByCurrentStagesIndexer(obj client.Object) []string {
 		return nil
 	}
 	currentStages := make([]string, len(freight.Status.CurrentlyIn))
-	var i int
+	currentStages := make([]string, 0, len(freight.Status.CurrentlyIn))
```
hiddeco, line 29: Something went wrong here when the suggestion was auto-applied.

api/v1alpha1/stage_helpers.go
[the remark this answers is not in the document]
hiddeco: I felt it should be put on the Freight instead of the Stage because, assuming the Freight `!= nil`, it is always available. If it is available _to_ something appears to be the responsibility of the Freight (see also `.IsApprovedFor` and `.IsVerifiedIn`), and not the entity having the desire.

api/v1alpha1/freight_types.go
krancour: We took this offline and struggled to come up with anything that was obviously better. The one potential thing we came up with wasn't better enough to justify the migration logic and deprecated fields that would have to stick around for many released to come. We agreed to come back to this someday if something worth paying that price for presents itself.

api/v1alpha1/freight_helpers_test.go
krancour: Will fix.

krancour:
I removed four codegen commits, rebased on main, re-ran codegen, and force pushed.

krancour (approved):

PR #3107, opened 2024-12-09
feat: custom PR title (#3063)

### LINKED ISSUE #3063: Custom PR Title
# Checklist

# Proposed Feature

It would be really great if we could set a custom PR title via the git-open-pr step.  Either completely override the last commit massage or, if possible, Add to the title.  

# Motivation

As suggested https://github.com/akuity/kargo/issues/3002
I can absolutely forsee someone wanting the ability to customize the title and thus overriding the last commit message

I know a lot of shops will add things like [staging] or [release] at the beginning of the title which might trigger automated workflows as an example.

# Suggested Implementation

muenchdo:
closes #3063


muenchdo:
@krancour Please let me know your thoughts on testing here. My thinking here is that we can both assume expressions in promotion steps are evaluated properly (likely tested elsewhere), and that if I pass any title to a Git provider's `CreatePullRequest`, the pull request will be created with the given title.

krancour:
@muenchdo correct on both counts. 😄

krancour:
@muenchdo I think if you re-run codegen, this looks g2g.

krancour (approved):
LGTM

Thanks for this @muenchdo!

PR #3111, opened 2024-12-10
fix(controller): expressions: fix quote() of a json object

### LINKED ISSUE #3109: JSON body in HTTP promotion step causes schema validation failure
# Checklist

# Description

I'm trying to do a POST request with a JSON body as part of an HTTP promotion step. It fails during the config validation because after evaluating the expressions, the body is no longer a string but a map. 

I've tried things like inlining my JSON body string and multiple YAML multi-line scalars (`>` and `|`), but couldn't get it to work. It might be possible with different quoting, but I'd argue that Kargo should be smart enough to know that an HTTP request body should always be interpreted as a string. This would also preserve readability.

# Screenshots

CleanShot 2024-12-09 at 19 01 08@2x

# Steps to Reproduce

```yaml
kind: Stage
metadata:
  name: in-cluster--beta
  namespace: echoserver
spec:
  requestedFreight:
  - origin:
      kind: Warehouse
      name: echoserver
    sources:
      direct: true
  promotionTemplate:
    spec:
      steps:
      - uses: http
        config:
          method: POST
          url: https://echo.free.beeceptor.com
          body: |
            {
              "body": {
                "content": [
                  {
                    "content": [
                      {
                        "text": "Kargo triggered a promotion to ${{ ctx.stage }}",
                        "type": "text"
                      }
                    ],
                    "type": "paragraph"
                  }
                ],
                "type": "doc",
                "version": 1
              }
            }
```

# Version

```
Client Version: v0.9.0
Server Version: v1.1.1
```

# Logs

```
time="2024-12-10T02:45:33Z" level=error msg="error executing Promotion" error="step execution failed: step 1 met error threshold of 1: failed to run step \"http\": invalid http config: body: Invalid type. Expected: string, given: object" freight=857d4e7e79ff63801a8b675728aa0d0dce1d4327 namespace=echoserver promotion=in-cluster--beta.01jeq7a9a8m0ds3b22tfhz2gh5.857d4e7 stage=in-cluster--beta
```

krancour:
Fixes #3109 

cc @muenchdo thanks for discovering this issue.

### DIFF
--- a/docs/docs/35-references/10-promotion-steps.md
+++ b/docs/docs/35-references/10-promotion-steps.md
@@ -1300,7 +1300,7 @@ with a wide variety of external services.
 | `queryParams` | `[]object` | N | A list of query parameters to include in the request. |
 | `queryParams[].name` | `string` | Y | The name of the query parameter. |
 | `queryParams[].value` | `string` | Y | The value of the query parameter. The provided value will automatically be URL-encoded if necessary. |
-| `body` | `string` | N | The body of the request. |
+| `body` | `string` | N | The body of the request. __Note:__ As this field is a `string`, take care to utilize [`quote()`](./20-expression-language.md#quote) if the body is a valid JSON `object`.  |
 | `insecureSkipTLSVerify` | `boolean` | N | Indicates whether to bypass TLS certificate verification when making the request. Setting this to `true` is highly discouraged. |
 | `timeout` | `string` | N | A string representation of the maximum time interval to wait for a request to complete. _This is the timeout for an individual HTTP request. If a request is retried, each attempt is independently subject to this timeout._ See Go's [`time` package docs](https://pkg.go.dev/time#ParseDuration) for a description of the accepted format. |
 | `successExpression` | `string` | N | An [expr-lang] expression that can evaluate the response to determine success. If this is left undefined and `failureExpression` _is_ defined, the default success criteria will be the inverse of the specified failure criteria. If both are left undefined, success is `true` when the HTTP status code is `2xx`. If `successExpression` and `failureExpression` are both defined and both evaluate to `true`, the failure takes precedence. Note that this expression should _not_ be offset by `${{` and `}}`. See examples for more details. |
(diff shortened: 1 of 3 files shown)


muenchdo:
Thanks for the quick fix! I'd say JSON bodies are such a common use case that it's worth having an explicit example. Maybe [here](https://github.com/akuity/kargo/blob/ad9e16f63ae050a04edc4e2bc3b54980586546e0/docs/docs/35-references/10-promotion-steps.md?plain=1#L1454-L1466)? Or what you [commented](https://github.com/akuity/kargo/issues/3109#issuecomment-2530345238) is also great.

krancour:
> Thanks for the quick fix! I'd say JSON bodies are such a common use case that it's worth having an explicit example.

Totally agree. The example of posting to Slack is probably a good place for this. Knocking this down to a draft until I can fix that in the morning.

krancour:
Fixed the Slack example.

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -1452,8 +1455,8 @@ steps:
     - name: Content-Type
       value: application/json
     body: |
-      {
-        "channel": ${{ vars.slackChannel }},
+      ${{ quote({
```
krancour, line 1458: The existing example was actually wrong without this.

internal/expressions/json_templates.go
```diff
@@ -105,7 +105,13 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
 	}
 	exprOpts = append(exprOpts, expr.Function(
 		"quote",
-		func(a ...any) (any, error) { return fmt.Sprintf(`"%v"`, a[0]), nil },
+		func(a ...any) (any, error) {
+			jsonBytes, err := json.Marshal(a[0])
```
hiddeco, line 109: Should we `json.Valid` before attempting to marshal?

hiddeco (approved):

internal/expressions/json_templates.go
krancour: Discussed offline.

Issue #2788, opened 2024-10-21
kustomize-set-image improvements


Brightside56:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

`kustomize-set-image` requires setting `images[].image` field. Proposed feature is to make field non-required

# Motivation

Requirement to additionally enumerate images for incorporation to manifests (when I for example would like to incorporate all images allowed by `requestedFreight`) makes templating of promotionTemplate non-trivial. 

# Suggested Implementation

When field isn't set - kustomize could set all images originating from freight. Probably... Probably... There could be a possibility to use kustomize build or something different for setting images (including setting images to helm charts) and to reproduce `render: {}` behaviour, because, I wouldn't like my Kargo pipeline to be aware how images should be set and I would like define it in repository with manifests instead

hiddeco:
Sorry for the late check-in, this had escaped my attention.

I can see the value in being able to say: "attempt to update anything that matches this bag of references", instead of having to maintain an explicit (double) declaration of what you want to update. Especially because with #2827 (or the successor proposal we are working on related to this), we imagine it is possible to have a declaration of "common tasks" that can be shared between Projects/Stages/PromotionTemplates, and the image configuration may then be arbitrary (and e.g. Project or Stage-specific).

> There could be a step which allows kustomize build (with Helm plugin) and set images using kustomize.yaml at same time

I think this would be better solved by having a "common task" as I referred to earlier that bundles these two based on your personal preferences.

krancour:
> I can see the value in being able to say: "attempt to update anything that matches this bag of references"

Agree on the usefulness, but calling out the root of a few edge cases...

If a Stage requests Freight from multiple origins (i.e. Warehouses) that each subscribe to the same image, but with different selection criteria, there can be ambiguity over the correct revision of the image to use with (the equivalent of) `kustomize edit set image`.

PR #3115, opened 2024-12-10
feat(directives): optional `images` for `kustomize-set-image`

hiddeco:
With this pull request, setting `images` for the `kustomize-set-image` step is now optional.

When left unspecified, all images from the Freight collection are automatically set as long as there is no ambiguous target image found (due to e.g. two Warehouses having different selectors for the same image). In case of an ambiguous target image, an error is returned and manual configuration of `images` will still be required.

It is worth noting that the generated commit message describes what's being set in the `kustomization.yaml`, and will thus contain a reference to all images even if they do not appear in any of the Kubernetes manifests which are part of the Kustomization overlay. 

Fixes: #2788

### DIFF
--- a/docs/docs/35-references/10-promotion-steps.md
+++ b/docs/docs/35-references/10-promotion-steps.md
@@ -312,7 +312,7 @@ to executing `kustomize edit set image`. This step is commonly followed by a
 | Name | Type | Required | Description |
 |------|------|----------|-------------|
 | `path` | `string` | Y | Path to a directory containing a `kustomization.yaml` file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
-| `images` | `[]object` | Y | The details of changes to be applied to the `kustomization.yaml` file. At least one must be specified. |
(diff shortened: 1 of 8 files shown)


docs/docs/35-references/10-promotion-steps.md
```diff
@@ -312,7 +312,7 @@ to executing `kustomize edit set image`. This step is commonly followed by a
 | Name | Type | Required | Description |
 |------|------|----------|-------------|
 | `path` | `string` | Y | Path to a directory containing a `kustomization.yaml` file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
-| `images` | `[]object` | Y | The details of changes to be applied to the `kustomization.yaml` file. At least one must be specified. |
+| `images` | `[]object` | N | The details of changes to be applied to the `kustomization.yaml` file. When left unspecified, all images from the Freight collection will be set in the Kustomization file. Unless there is a ambiguous image name (for example, due to two Warehouses subscribing to the same repository), which requires manual configuration. |
```
krancour, line 315:
```suggestion
| `images` | `[]object` | N | The details of changes to be applied to the `kustomization.yaml` file. When left unspecified, all images from the Freight collection will be set in the Kustomization file. Unless there is an ambiguous image name (for example, due to two Warehouses subscribing to the same repository), which requires manual configuration. |
```

hiddeco: Thanks, amended the commit.

PR #3118, opened 2024-12-10
feat: PR labels (#3002)

### LINKED ISSUE #3002: PR Labels
# Checklist

kargo v1.0.3

# Proposed Feature
It would be really great if we could set PR labels via the git-open-pr step.  

# Motivation
We like to use them to filter our PRs in github and it would be nice if we didn't have to do this after the PR is created.
Additionally, one could have a github action triggered based on the label being added and the value.  While I don't have an immediate use case for this, I could see how I might use it.

# Suggested Implementation

The git-open-pr promotion step could have a configuration item that is Labels that is of type list of strings.  Each item in the list is added as a label after the PR is opened.  

It would be worth noting that the labels have to pre-exist in your github repo for the workflow to add the label, but that would be on the creator of the workflow to insure.

muenchdo:
closes #3002

### DIFF
--- a/internal/gitprovider/azure/azure.go
+++ b/internal/gitprovider/azure/azure.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/microsoft/azure-devops-go-api/azuredevops/v7"
+	adocore "github.com/microsoft/azure-devops-go-api/azuredevops/v7/core"
 	adogit "github.com/microsoft/azure-devops-go-api/azuredevops/v7/git"
 	"k8s.io/utils/ptr"
 
@@ -94,6 +95,12 @@ func (p *provider) CreatePullRequest(
 		return nil, fmt.Errorf("error getting repository %q: %w", p.repo, err)
 	}
 	repoID := ptr.To(repository.Id.String())
+	labels := make([]adocore.WebApiTagDefinition, 0, len(opts.Labels))
+	for _, label := range opts.Labels {
+		labels = append(labels, adocore.WebApiTagDefinition{
+			Name: &label,
+		})
+	}
 	sourceRefName := ptr.To(fmt.Sprintf("refs/heads/%s", opts.Head))
 	targetRefName := ptr.To(fmt.Sprintf("refs/heads/%s", opts.Base))
 	adoPR, err := gitClient.CreatePullRequest(ctx, adogit.CreatePullRequestArgs{
@@ -102,6 +109,7 @@ func (p *provider) CreatePullRequest(
 		GitPullRequestToCreate: &adogit.GitPullRequest{
 			Title:         &opts.Title,
 			Description:   &opts.Description,
+			Labels:        &labels,
 			SourceRefName: sourceRefName,
 			TargetRefName: targetRefName,
 		},
--- a/internal/gitprovider/github/github.go
+++ b/internal/gitprovider/github/github.go
@@ -41,11 +41,42 @@ func init() {
 	gitprovider.Register(ProviderName, registration)
 }
 
+type githubClient interface {
+	CreatePullRequest(
+		ctx context.Context,
+		owner string,
+		repo string,
+		pull *github.NewPullRequest,
+	) (*github.PullRequest, *github.Response, error)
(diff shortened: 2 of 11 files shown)


muenchdo:
@krancour I went down the rabbit hole of adding tests for the GitHub provider. Please let me know if this is generally going in a direction you like or whether the added complexity for these new tests is too much.

krancour:
> I went down the rabbit hole of adding tests for the GitHub provider. 

I like how you're approaching it, but the thing I worry about is that for something as complex as the GitHub APIs, it's plausible for the mock to accidentally do something that's inconsistent with the real API's behavior with the result being tests that pass but shouldn't.

We eventually need to figure out how to have dedicated repos _for each provider_ perpetually available to test against or have a way to provision repos just-in-time and clean them up afterwards.

For now, I wouldn't let the above get in the way.

muenchdo:
> For now, I wouldn't let the above get in the way.

By that do you mean you wouldn't add tests at all or just leave them as proposed in the PR and then add some real integration tests later?

internal/gitprovider/github/github.go
```diff
@@ -81,10 +112,30 @@ func NewProvider(
...
+}
+
+func (g githubClientWrapper) GetPullRequests(ctx context.Context, owner string, repo string, number int) (*github.PullRequest, *github.Response, error) {
+	return g.client.PullRequests.Get(ctx, owner, repo, number)
+}
+
+func (g githubClientWrapper) AddLabelsToIssue(ctx context.Context, owner string, repo string, number int, labels []string) ([]*github.Label, *github.Response, error) {
+	return g.client.Issues.AddLabelsToIssue(ctx, owner, repo, number, labels)
+}
+
```
krancour, line 138: Instead of creating a client wrapper, could the `githubClient` interface just be the relevant subset of the methods the client already actually has? This way we can just use the existing (unwrapped) client as the implementation of the `githubClient`, which is the interface we'll code to?

krancour: Oh... it's because of this `p.client.PullRequests.Create` 🤦‍♂️ 

That makes my suggestion rather difficult to pull off.

Nevermind.

muenchdo: Believe me, I tried 😄

internal/gitprovider/gitlab/gitlab.go
```diff
@@ -119,6 +119,7 @@ func (p *provider) CreatePullRequest(
 	glMR, _, err := p.client.CreateMergeRequest(p.projectName, &gitlab.CreateMergeRequestOptions{
 		Title:              &opts.Title,
 		Description:        &opts.Description,
+		Labels:             (*gitlab.LabelOptions)(&opts.Labels),
```
krancour, line 122: Man, I wish the GitHub client had made it this easy.

krancour:
> By that do you mean you wouldn't add tests at all or just leave them as proposed in the PR and then add some real integration tests later?

Dealer's choice. I'm fine either way.

btw... #3128 just got merged, so you probably have to tweak that new provider a bit as well. Hopefully that's no too involved.

internal/gitprovider/azure/azure.go
```diff
@@ -94,6 +95,12 @@ func (p *provider) CreatePullRequest(
 		return nil, fmt.Errorf("error getting repository %q: %w", p.repo, err)
 	}
 	repoID := ptr.To(repository.Id.String())
+	labels := make([]adocore.WebApiTagDefinition, 0, len(opts.Labels))
+	for _, label := range opts.Labels {
+		labels = append(labels, adocore.WebApiTagDefinition{
+			Name: &label,
+		})
+	}
```
muenchdo, line 103: Should work based on [this](https://github.com/Azure/azure-devops-cli-extension/pull/1358/files#diff-003bc1ec1705ca0b976c2786c1ecf1e3370683b0dac7ce8d759dde906d71e073R164)

krancour (approved):

krancour:
@diegocaspi thank you for some excellent work on this!

PR #3119, opened 2024-12-11
fix(controller): git-push step: pull --rebase before push

### LINKED ISSUE #3071: git-push subject to race conditions when many stages push to same mono repo
# Description

Two or more stages might attempt to push to same repo at the same time (e.g. using promote downstream button). When this happens, the git-push step might intermittently fail like:

`
step execution failed: failed to run step "git-push": error pushing commits to remote: error pushing branch: error executing cmd [/usr/bin/git push origin HEAD]: remote: Bypassed rule violations for refs/heads/main: remote: remote: - Changes must be made through a pull request. remote: To https://github.com/example/guestbook-deploy.git ! [remote rejected] HEAD -> main (cannot lock ref 'refs/heads/main': is at 74291c2cb80ee3f5c6e5e6e329e13a5a1866ca61 but expected 473aff63b3da037e6c6d4a3de2dabaee3c196c83) error: failed to push some refs to 'https://github.com/example/guestbook-deploy.git'
`

A workaround may be to use the retry feature, but I haven't verified this.

# Steps to Reproduce

Have multiple stages promoting at same time while modifying the same gitops repo. Timing should try to overlap 

# Version

```
v1.2.0-unstable-20241205
```

krancour:
Fixes #3071 

Although this is WIP 👀 would still be helpful, @hiddeco. 🙏 

Apart from the git bits of this, this PR is the catalyst for the step execution engine to start distinguishing between errors/failures that are terminal and ones that can be retried. This seemed like the time, because once a merge conflict that requires manual resolution has been detected, there is no point in performing any retries, no matter what the retry policy says.

I tried to implement what we discussed offline, but I think you'll have some good constructive feedback on this.

r4r

### DIFF
--- a/internal/controller/git/errors.go
+++ b/internal/controller/git/errors.go
@@ -1,0 +1,14 @@
+package git
+
+import (
+	"errors"
+)
+
+// ErrMergeConflict is returned when a merge conflict occurs.
+var ErrMergeConflict = errors.New("merge conflict")
+
+// IsMergeConflict returns true if the error is a merge conflict or wraps one
+// and false otherwise.
+func IsMergeConflict(err error) bool {
+	return errors.Is(err, ErrMergeConflict)
+}
--- a/internal/controller/git/errors.go
+++ b/internal/controller/git/errors.go
@@ -12,3 +12,13 @@ var ErrMergeConflict = errors.New("merge conflict")
 func IsMergeConflict(err error) bool {
 	return errors.Is(err, ErrMergeConflict)
 }
+
+// ErrNonFastForward is returned when a push is rejected because it is not a
+// fast-forward or needs to be fetched first.
+var ErrNonFastForward = errors.New("non-fast-forward")
+
+// IsNonFastForward returns true if the error is a non-fast-forward or wraps one
+// and false otherwise.
+func IsNonFastForward(err error) bool {
+	return errors.Is(err, ErrNonFastForward)
+}
--- a/internal/controller/git/work_tree.go
+++ b/internal/controller/git/work_tree.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -57,6 +58,9 @@ type WorkTree interface {
 	GetDiffPathsForCommitID(commitID string) ([]string, error)
 	// IsAncestor returns true if parent branch is an ancestor of child
 	IsAncestor(parent string, child string) (bool, error)
+	// IsRebasing returns a bool indicating whether the working tree is currently
+	// in the middle of a rebase operation.
+	IsRebasing() (bool, error)
 	// LastCommitID returns the ID (sha) of the most recent commit to the current
 	// branch.
 	LastCommitID() (string, error)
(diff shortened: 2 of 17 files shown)


internal/controller/git/errors.go
```diff
@@ -0,0 +1,14 @@
...
+)
+
+// ErrMergeConflict is returned when a merge conflict occurs.
+var ErrMergeConflict = errors.New("merge conflict")
+
+// IsMergeConflict returns true if the error is a merge conflict or wraps one
+// and false otherwise.
+func IsMergeConflict(err error) bool {
+	return errors.Is(err, ErrMergeConflict)
+}
```
hiddeco, line 14: Why create an abstraction over just plain usage of `errors.Is`?

internal/controller/git/work_tree.go
```diff
@@ -316,6 +320,25 @@ func (w *workTree) IsAncestor(parent string, child string) (bool, error) {
...
+	if _, err = os.Stat(rebaseMerge); err == nil {
+		return true, nil
+	}
+	if res, err = libExec.Exec(w.buildGitCommand("rev-parse", "--git-path", "rebase-apply")); err != nil {
+		return false, fmt.Errorf("error determining rebase status: %w", err)
+	}
+	rebaseApply := filepath.Join(w.dir, strings.TrimSpace(string(res)))
+	if _, err = os.Stat(rebaseApply); err == nil {
+		return true, nil
+	}
```
hiddeco, line 338:
```suggestion
	if _, err = os.Stat(rebaseMerge); !os.IsNotExist(err) {
		if err != nil {
			return false, err
		}
		return true, nil
	}
	if res, err = libExec.Exec(w.buildGitCommand("rev-parse", "--git-path", "rebase-apply")); err != nil {
		return false, fmt.Errorf("error determining rebase status: %w", err)
	}
	rebaseApply := filepath.Join(w.dir, strings.TrimSpace(string(res)))
	if _, err = os.Stat(rebaseApply); !os.IsNotExist(err) {
		if err != nil {
			return false, err
		}
		return true, nil
	}
```

internal/controller/git/errors.go
krancour: It felt slightly more ergonomic for callers from other packages since it obviates the need to reason through whether `errors.Is()` or `errors.As()` is the appropriate approach to simply answering yes/no on whether a merge conflict occurred. In fact, with this, you don't even need to know that `ErrMergeConflict` exists. You just get a simple yes/no.

But I'm not that strongly attached to it.

internal/directives/git_pusher.go
```diff
@@ -116,18 +120,49 @@ func (g *gitPushPusher) runPromotionStep(
...
+		//
+		// This is misleading, as it implies that the total number of attempts may
+		// exceed the value of Steps and that Steps only dictates the maximum number
+		// of adjustments to the interval between retries.
+		//
+		// Reading the implementation of retry.DefaultBackoff reveals that Steps
+		// is indeed the maximum number of attempts.
+		backoff.Steps = int(*cfg.MaxAttempts)
+	} else {
+		backoff.Steps = 50
```
hiddeco, line 147: This default seems pretty high, I think something like `10` would probably be better. Unless you have a specific reason to opt for this value?

krancour: With the following, and ~20 promos running concurrently and writing to the same branch, I was _occasionally_ able to exhaust attempts:

```
wait.Backoff{
	Duration: 1 * time.Second,
	Factor:   2,
	Steps:    10,
	Cap:      2 * time.Minute,
	Jitter:   0.1,
}
```

Being aware of how extensive some users' monorepos are, I presumed 10 to be not _nearly_ enough.

The effective defaults are now:

```
wait.Backoff{
	Steps:    50,
	Duration: 10 * time.Millisecond,
	Factor:   5.0,
	Jitter:   0.1,
}
```

I think I still like defaulting to a reasonably high number of steps, but what I definitely _didn't_ account for here is that `retry.DefaultBackoff` has no cap -- which makes 50 steps absurd.

What do you think about leaving it 50, but capping it at something like 5s?

krancour: Alternatively, I could use a mutex to make the pull + rebase + push a critical section of code.

Given that controllers can be sharded, it won't eliminate the need for retries _entirely_, but it _would_ improve the situation considerably.

krancour: I was using this default backoff because, iirc, you'd suggested it, but I'm rethinking that.

In general, the best way to avoid conflicts will be spacing out the retries of the multiple concurrent promos...

That's got me thinking about a larger starting interval and a larger jitter.

At the same time, increasing the interval by a factor of five at each step seems way too aggressive. Even two seems aggressive.

I'm contemplating something like this:

```
wait.Backoff{
	Steps:    10,
	Duration: time.Second,
	Factor:   1.5,
	Jitter:   0.5,
	Cap:   30 * time.Second,
}
```

I'll try that in the morning to see how it performs.

hiddeco (approved):

PR #3121, opened 2024-12-11
feat: (cluster) Promotion tasks

hiddeco:
Fixes: #2827

This pull request introduces the PromotionTask and ClusterPromotionTask resources to Kargo. These new resource types allow you to streamline Stage resources by bundling commonly shared promotion steps into reusable tasks, which can be referenced in promotion templates.

### `PromotionTask` example

```yaml
---
kind: PromotionTask
metadata:
  name: open-pr-and-wait
  namespace: my-project
spec:
  vars:
  - name: repoURL
  - name: sourceBranch
  - name: targetBranch
    value: main
  steps:
  - uses: git-open-pr
    as: open-pr
    config:
      repoURL: ${{ vars.repoURL }}
      createTargetBranch: true
      sourceBranch: ${{ vars.sourceBranch }}
      targetBranch: ${{ vars.targetBranch }}
  - uses: git-wait-for-pr
    as: wait-for-pr
    config:
      repoURL: ${{ vars.repoURL }}
      prNumber: ${{ task.outputs['open-pr'].prNumber }}
  - uses: compose-output
    as: output
    config:
      mergeCommit: ${{ task.outputs['wait-for-pr'].commit }}
```

### `Stage` example

```yaml
---
kind: Stage
metadata:
  name: staging
  namespace: my-project
spec:
  requestedFreight:
  - origin:
      kind: Warehouse
      name: kargo-demo
    sources:
      direct: true
  promotionTemplate:
    spec:
      vars:
      - name: repoURL
        value: https://github.com/example/repository.git
      - name: srcPath
        value: ./src
      - name: outPath
        value: ./out
      - name: targetBranch
        value: stage/${{ ctx.stage }}
      steps:
      # ...omitted for brevity
      - uses: git-push
        as: push
        config:
          path: ${{ vars.outPath }}
          generateTargetBranch: true
      - task:
          name: open-pr-and-wait
          kind: ClusterPromotionTask
        as: pr
        vars:
        - name: sourceBranch
          value: ${{ outputs.push.branch }}
```

### Full example for testing

https://gist.githubusercontent.com/hiddeco/df2da6a88ad1966642ed6ee94bad9787/raw/a886da867108c4ef970dec2b137e81f6a4b1f6fe/tasks.yaml


internal/indexer/indexer.go
```diff
@@ -168,7 +168,7 @@ func RunningPromotionsByArgoCDApplications(
 				Promotion: promo.Name,
 				Vars:      promo.Spec.Vars,
 			}
-			vars, err := dirStep.GetVars(promoCtx)
+			vars, err := dirStep.GetVars(promoCtx, promoCtx.State)
```
krancour, line 171: Since we only built out enough of promoCtx to get us through this, promotCtx.State will be empty. I suppose that's actually ok in this (pardon the overload) _context?_

internal/kargo/promotion_builder.go
```diff
@@ -0,0 +1,264 @@
+package kargo
```
krancour, line 1: This wasn't introduced by this PR, but this `internal/kargo` package has been bugging me for a while because I don't feel like the package has a clearly defined purpose. (Maybe it does, and I'm failing to see it.)

If I'm not wrong, there's certainly no need to untangle this knot to get this PR merged, but (again, only if I'm right), I wonder if there's any other place that might be appropriate for the new functionality added to the package, in particular `NewPromotionBuilder()`.

internal/kargo/kargo_test.go
```diff
@@ -37,6 +37,17 @@ func TestNewPromotion(t *testing.T) {
 					Name:      "test",
 					Namespace: "kargo-demo",
 				},
+				Spec: kargoapi.StageSpec{
```
krancour, line 40: Smallest of nits (higher up, because it's too far from the change to comment on): Do we want to consider renaming `TestNewPromotion()` to `TestPromotionBuilder()`?

krancour: Oh... actually, I see `TestPromotionBuilder_Build` already exists.

krancour: So now I wonder if there isn't some overlap between these two test funcs?

internal/api/promote_downstream_v1alpha1.go
```diff
@@ -137,19 +137,23 @@ func (s *server) PromoteDownstream(
...
 	for _, downstream := range downstreams {
-		newPromo := kargo.NewPromotion(ctx, downstream, freight.Name)
 		if downstream.Spec.PromotionTemplate != nil &&
 			len(downstream.Spec.PromotionTemplate.Spec.Steps) == 0 {
 			// Avoid creating a Promotion if the downstream Stage has no promotion
 			// steps and is therefore a "control flow" Stage.
 			continue
 		}
-		if err := s.createPromotionFn(ctx, &newPromo); err != nil {
+		newPromo, err := kargo.NewPromotionBuilder(s.client).Build(ctx, downstream, freight.Name)
```
krancour, line 146: This isn't a new problem, but the PR's making me more aware of it...

It seems that now, as before, we build out a Promo's steps from the template in three places:

1. Promote to Stage API endpoint
2. Promote to downstream States API endpoint
3. Auto-Promotion by the Stage reconciler

In all cases, a Promo's steps are defined when it's _created._

I'll avoid the question of whether it's better to copy/expand steps when the Promo is created vs being done by the Promo reconciler when the Promo's _started_ (because I could see it both ways and we may have to confront this eventually), but the weird side effect of the Promo reconciler not doing it is that you can't create a Promo with kubectl unless you defined all the steps yourself. There's no way to say: "This Stage, this piece of Freight, and let the controller fill in the steps."

Should we at least have the Promo reconciler build/expand steps just-in-time if a work on a Promo starts and its steps are empty?

internal/directives/output_composer.go
```diff
@@ -0,0 +1,70 @@
...
+
+// outputComposer is an implementation of the PromotionStepRunner interface
+// that allows composing outputs from previous steps into new outputs.
+//
+// It works based on the PromotionStepContext.Config field allowing to an
+// arbitrary number of key-value pairs to be exported as outputs.
+// Because the values are allowed to be expressions and can contain
+// references to outputs from previous steps, this allows for remapping
+// the outputs of previous steps to new keys, or even combining them
+// into new structures.
```
krancour, line 25: I know docs are coming in a follow-up. In the interim, I think a small example at the tail end of this comment would be helpful.

internal/kargo/kargo_test.go
hiddeco: Ah great catch, this is an old test that can just be removed but continued to work because I wrote something that was... backward compatible.

internal/api/promote_downstream_v1alpha1.go
hiddeco: This is an excellent point, but I wonder if instead of doing this — wouldn't it be better to introduce a modifying webhook?

hiddeco: We would still need some sort of builder, I think. But the inflation would happen as a secondary step in a single place.

internal/indexer/indexer.go
krancour: Explicit nil sgtm, maybe with a brief comment about why that arg is unimportant in this context?

Really this is a nit... so however you want to deal with it is fine, even if it's do nothing.

internal/kargo/promotion_builder.go
```diff
@@ -0,0 +1,259 @@
...
+			Steps:   stage.Spec.PromotionTemplate.Spec.Steps,
+		},
+	}
+	return &promotion, nil
+}
+
+// InflateSteps inflates the Promotion steps by resolving any references to
+// PromotionTasks and expanding them into their individual steps. The inflated
+// steps are then set on the Promotion, replacing the original steps.
+func (b *PromotionBuilder) InflateSteps(ctx context.Context, promo *kargoapi.Promotion) error {
```
krancour, line 92: Making sure I understand the rational for having InflateSteps() separate from Build()...

I presume that by not having Build() automatically inflate, we are preserving the possibility of a Promo having been defined manually and submitted via kubectl, in which case, building isn't necessary, but expansion is?

internal/indexer/indexer.go
hiddeco: Upon closer inspection, I do think we need to pass some state here. In addition, I have also a vague suspicion we may actually need the outputs as well while evaluating the individual fields because _theoretically_, the values can not just originate from variables.

internal/kargo/promotion_builder.go
hiddeco: The only alternative I can think of is inventing a new `promotion` package, but I do not really like this alternative either...

internal/kargo/promotion_builder.go
hiddeco: `Build` is for creating a Promotion based on a Stage (or rather, the Promotion template within). Which we do in several places, either through the API or CLI.

`InflateSteps` is what always happens on the server-side, and ensures that even if a Promotion would be created through `kubectl` (as you suggest) — we still handle the core logic of correctly inflating any references we find into actual steps.

internal/kargo/promotion_builder.go
```diff
@@ -0,0 +1,259 @@
...
+
+	promoVarsMap := make(map[string]kargoapi.PromotionVariable, len(promoVars))
+	for _, v := range promoVars {
+		promoVarsMap[v.Name] = v
+	}
+
+	stepVarsMap := make(map[string]kargoapi.PromotionVariable, len(stepVars))
+	for _, v := range stepVars {
+		stepVarsMap[v.Name] = v
+	}
```
krancour, line 236: Is there any edge case here where we lose anything important through the use of maps?

I'm thinking something like this, which seems dumb _but possible_, might be such a case?

```
vars:
- name: foo
  value: bar
- name: foo
  value: ${{ vars.foo + vars.foo }}
```

krancour: The second would clobber the first in the map, but the result would be an error, I think, instead of "barbar".

internal/directives/promotions.go
```diff
@@ -190,28 +207,63 @@ func (s *PromotionStep) GetConfig(
 
 // GetVars returns the variables defined in the PromotionStep. The variables are
 // evaluated in the context of the provided PromotionContext.
-func (s *PromotionStep) GetVars(promoCtx PromotionContext) (map[string]any, error) {
-	vars := make(map[string]any, len(promoCtx.Vars))
+func (s *PromotionStep) GetVars(
+	promoCtx PromotionContext,
+	state State,
+) (map[string]any, error) {
+	var rawVars = make(map[string]string, len(promoCtx.Vars))
```
krancour, line 214: Same question about maps as [here](https://github.com/akuity/kargo/pull/3121/files#r1893194578).

internal/kargo/promotion_builder.go
krancour: We can put a pin in it. Fortunately there are only a couple packages like this that seem to have collected unrelated bits of functionality over time for lack of a better home.

internal/kargo/promotion_builder.go
krancour: Ok... so I agree with the rationale for keeping them separate, but is there any reason we inflate server side, but we don't build server side?

Could the mutating webhook we discussed build if steps == nil and then inflate unconditionally? As opposed to calling build in several different places?

hiddeco: We discussed this offline, and partly because of debugging reasons and introducing too many sprinkles of magic. It is better to keep the builder in place so you have some (meaningful) YAML to hold on to.

internal/kargo/promotion_builder.go
hiddeco: This is still an issue, the plan is to inflate them as:

1. task-level vars
2. task-step-level vars

Which would allow the task-step-level to overwrite (or utilize) the task-level variable.

The only concern I have with this approach is that the promotion-level variables then can not overwrite the task-level (default) variables, because they are evaluated before any of the variables in the above list (and a task-level variable with the same name would thus overwrite this value).

api/v1alpha1/cluster_promotion_task_types.go
```diff
@@ -0,0 +1,27 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +kubebuilder:resource:scope=Cluster,shortName={clusterpromotask,clusterpromotasks}
+// +kubebuilder:object:root=true
+// +kubebuilder:printcolumn:name=Age,type=date,JSONPath=`.metadata.creationTimestamp`
+
+type ClusterPromotionTask struct {
```
jessesuen, line 9: While I know ClusterPromotionTask is technically accurate, I wonder if we want to term this "GlobalPromotionTask". We already refer to credentials and ServiceAccounts that are to be used in a global capacity as "global" credentials and global serviceaccounts, and this would be along those lines.

The "cluster" is less meaningful in

krancour: @jessesuen I wonder if I can offer a counterpoint. The "global" things we have now (and I don't love that term) such as credentials and SAs are things incapable of living in a cluster scope. Putting them in designated namespaces is essentially a workaround.

Here, we have the opportunity to handle things in a more conventional manner. When a CRD whose instances will exist in the cluster scope needs to be differentiated from a namespace-scoped counterpart, I think "Cluster____" is a very widely observed convention that people will understand intuitively.

krancour (approved):
🔥

PR #3039, opened 2024-12-02
chore(ui): bump `oauth4webapi` major version

Marvin9:
**NOTE: This could potentially be breaking change if API server does not have SSL and Dex is being used. HTTP requests will not be supported to IDPs**

Read all breaking changes here - https://github.com/panva/oauth4webapi/releases/tag/v3.0.0

Old dependabot PR https://github.com/akuity/kargo/pull/3025

PR #3122, opened 2024-12-11
fix(auth): token renew failure

Marvin9:
Not sure if this was issue previously or after recent library major version bump but token renew failures went unnoticed and causing page stuck on loading spinner

### DIFF
--- a/ui/src/features/auth/token-renew.tsx
+++ b/ui/src/features/auth/token-renew.tsx
@@ -79,25 +79,35 @@ export const TokenRenew = () => {
     }
 
     (async () => {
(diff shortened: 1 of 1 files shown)


krancour:
I ran into this also starting just recently.

krancour (approved):

krancour:
Thank you for being right on top of this!

krancour:
Womp womp.

Marvin9:
@krancour this might not necessarily needs to be backported due to the reason that this _might_ be caused by recent major version bump of oauth dependency https://github.com/akuity/kargo/pull/3039, which hasn't been released. I will confirm

Issue #3024, opened 2024-11-30
PR link no longer displayed in the Promotion UI


simonartige:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

With the new promotion steps, which are indeed more flexible now, it seems that we have lost the link to the opened PR in the promotion view on the Kargo UI.
This feature on the UI was very handy as it allowed to quickly see the pull request opened for the promotion.

Looking at the code, we can see the functionality got removed with this change:
https://github.com/akuity/kargo/pull/2694/files#diff-831008a406138fc2b4e0288bf2dfc2fe4c9ed871de1268ee4c67edaa0f7977f2
Which looks like the intended behavior.

We can still find the PR number by browsing in the new promotion details panel 
<img width="648" alt="image" src="https://github.com/user-attachments/assets/f9842204-aa19-46bd-9f7f-a472ab5b3471">
But this requires more navigation and there is no link available - only the PR number

Are there any future plans to show the list of PRs opened in the promotion steps on the UI?

I logged a bug, but this is a functional regression. But I would understand if this is seen as a feature request as this is a feature from the 0.x Kargo version, that was before the new promotion steps mechanism. With the new system, there could be multiple PRs opened in a single promotion, therefore there may be UX challenges to display them. But I would say it is expected to find these PR links somehow when we see this icon 
<img width="226" alt="image" src="https://github.com/user-attachments/assets/405c21ad-19e2-4091-8367-b621025a620c">

# Screenshots

![image](https://github.com/user-attachments/assets/ae09f78b-b000-492a-8663-1733948ce656)

# Steps to Reproduce

Create a promotion with the following configuration:

```
        - uses: git-commit
          config:
            path: ./app
            messageFromSteps:
              - helm-update-image
        - uses: git-push
          as: push
          config:
            path: ./app
            generateTargetBranch: true
        - uses: git-open-pr
          as: open-pr
          config:
            repoURL: https://github.com/x/y
            provider: github
            sourceBranchFromStep: push
            targetBranch: main
        - uses: git-wait-for-pr
          config:
            repoURL: https://github.com/x/y
            provider: github
            prNumberFromStep: open-pr
```
Promote a freight using that configuration
A PR is opened during the promotion
Looking at the promotion history in the stage details, we don't have the link to the PR anymore (see screenshot)

[...]

PR #3129, opened 2024-12-12
feat(ui): PR deep links in Promotion steps and in Promotion lists

Marvin9:
fixes https://github.com/akuity/kargo/issues/3024

Other than fix. This PR does some works for

- Lays very soft foundation for future plugin architecture. UI takes opportunity of learn possible ways to split plugins. Note that this is not final and subject to change based on discussions when time comes but at the very least makes it easier to break into pieces
- Removes the icons in promotion steps.. with deep links, it looked more busy with icons.

https://github.com/user-attachments/assets/bd4d2ad7-a106-458d-871b-6c0ea3995e91

Will cleanup some UI a bit but other than that its ready to review

### DIFF
--- a/ui/src/features/stage/promotions.tsx
+++ b/ui/src/features/stage/promotions.tsx
@@ -21,6 +21,8 @@ import {
 import { KargoService } from '@ui/gen/service/v1alpha1/service_connect';
 import { ListPromotionsResponse } from '@ui/gen/service/v1alpha1/service_pb';
 import { Freight, Promotion } from '@ui/gen/v1alpha1/generated_pb';
+import uiPlugins from '@ui/plugins';
+import { UiPluginHoles } from '@ui/plugins/ui-plugin-hole/ui-plugin-holes';
 
 import { PromotionDetailsModal } from './promotion-details-modal';
 import { hasAbortRequest, promotionCompareFn } from './utils/promotion';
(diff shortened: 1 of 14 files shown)


rpelczar (changes requested):
IMO there is too much shared components, utils and types in plugins. Let's try to separate code as much as you can. Also, I'd like to have more clear structure.
Maybe:

1. plugins
-    common
-    plugin1
-    plugin2
-    index.tsx

Marvin9:
> IMO there is too much shared components, utils and types in plugins. Let's try to separate code as much as you can. Also, I'd like to have more clear structure. Maybe:
> 
> -plugins
> 
> * common
> * plugin1
> * plugin2
> * index.tsx

Done! Thanks for this suggestion. I have use `atoms` instead of `common` as I think it explains as building blocks. But again this isn't final bullet proof and subject to change drastically in future

ui/src/features/stage/promotions.tsx
```diff
@@ -171,6 +173,26 @@ export const Promotions = () => {
           </Link>
         </Tooltip>
       )
+    },
+    {
+      title: 'Extra',
```
Marvin9, line 178: @rpelczar @krancour do you have any better name for this column?

rpelczar: @Marvin9 You can leave it empty.

Marvin9, line 178:
```suggestion
      title: '',
```

PR #3135, opened 2024-12-13
docs: Kargo role matrix

34fathombelow:
This PR introduces a 'Kargo Role Matrix` which outlines the maximum rules required for a ClusterRole/Role.


krancour (approved):
Thanks!

PR #3148, opened 2024-12-17
docs: Fix Argo application source

han-steve:
Currently, the demo repo puts all stages in the master branch. Fixing the application set code to reflect this.


han-steve:
Oh I see this is because Kargo generates the built code in these branches. My bad!

Issue #2961, opened 2024-11-19
json-update step


jessesuen:
# Proposed Feature

We just introduced the `yaml-update` step, which is a great improvement over `helm-update-image`. Now we should add an equivalent `json-update` step to deal with `json` files.

# Motivation

This would enable users who use other config management tools like .jsonnet or grafana tanka. Somone using jsonnet could patch an image tag to a `.json`. Technically they could even patch a `.jsonnet` file, as long as that file was pure json and did not use any jsonnet syntax.

# Suggested Implementation

Step syntax can follow the same syntax as `yaml-update` but understand how to manipulate a JSON file.

As a future improvement, we could consider leaving room for different patching types (JSON Patch, JSON Merge Patch), but I would not consider this required for MVP. Also, I do not think we should support strategic merge patch, which is a can of worms I would not want to go down.

krancour:
Adding a note here that, as we discussed offline, I think it's important to preserve formatting and field order. (We took great care to do so with YAML.) If we don't it creates noise in the diffs.

Fortunately, I think this is accomplished somewhat easily with this lib: https://github.com/tidwall/sjson

PR #3151, opened 2024-12-17
feat: json update promotion step

fykaa:
closes: https://github.com/akuity/kargo/issues/2961

This PR adds json-update promotion step that allows one to automate JSON updates within the Kargo promotion workflows

### DIFF
--- a/docs/docs/35-references/10-promotion-steps.md
+++ b/docs/docs/35-references/10-promotion-steps.md
@@ -579,6 +579,53 @@ steps:
 |------|------|-------------|
 | `commitMessage` | `string` | A description of the change(s) applied by this step. Typically, a subsequent [`git-commit`](#git-commit) step will reference this output and aggregate this commit message fragment with other like it to build a comprehensive commit message that describes all changes. |
 
+### `json-update`
+
+`json-update` updates the values of specified keys in any JSON file.
+
+#### `json-update` Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `path` | `string` | Y | Path to a JSON file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |                      |
+| `updates` | `[]object` | Y | The details of changes to be applied to the file. At least one must be specified. |
+| `updates[].key` | `string` | Y | The key to update within the file. For nested values, use a JSON dot notation path. See [sjson documentation](https://github.com/tidwall/sjson) for supported syntax. |
+| `updates[].value`| `any` | Y | The new value for the key. Typically specified using an expression. Supports strings, numbers, booleans, arrays, and objects. |
+
+#### `json-update` Example
+
+```yaml
+vars:
+- name: gitRepo
+  value: https://github.com/example/repo.git
+steps:
+- uses: git-clone
+  config:
+    repoURL: ${{ vars.gitRepo }}
+    checkout:
+    - commit: ${{ commitFrom(vars.gitRepo).ID }}
+      path: ./src
+    - branch: stage/${{ ctx.stage }}
+      create: true
+      path: ./out
+    - uses: git-clear
+      config:
+        path: ./out
+- uses: json-update
+  config:
+    path: configs/settings.json
+    updates:
+    - key: image.tag
+      value: ${{ imageFrom("my/image").Tag }}
+# Render manifests to ./out, commit, push, etc...
+```
+
+#### `json-update` Output
+
+| Name | Type | Description |
+|------|------|-------------|
+| `commitMessage` | `string` | A description of the change(s) applied by this step. Typically, a subsequent [`git-commit`](#git-commit) step will reference this output and aggregate this commit message fragment with other like it to build a comprehensive commit message that describes all changes. |
+
 ### `yaml-update`
 
 `yaml-update` updates the values of specified keys in any YAML file. This step
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,6 @@ require (
 	github.com/jferrl/go-githubauth v1.1.1
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/klauspost/compress v1.17.11
-	github.com/microsoft/azure-devops-go-api/azuredevops/v7 v7.1.0
 	github.com/oklog/ulid/v2 v2.1.0
 	github.com/otiai10/copy v1.14.0
 	github.com/patrickmn/go-cache v2.1.0+incompatible
@@ -122,7 +121,6 @@ require (
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/shopspring/decimal v1.4.0 // indirect
 	github.com/spf13/cast v1.7.0 // indirect
-	github.com/stretchr/objx v0.5.2 // indirect
 	github.com/valyala/bytebufferpool v1.0.0 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
@@ -199,6 +197,7 @@ require (
 	github.com/prometheus/common v0.55.0 // indirect
 	github.com/prometheus/procfs v0.15.1 // indirect
 	github.com/vbatts/tar-split v0.11.3 // indirect
+	github.com/xanzy/go-gitlab v0.115.0
 	github.com/xlab/treeprint v1.2.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.54.0 // indirect
 	go.opentelemetry.io/otel v1.31.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 	github.com/jferrl/go-githubauth v1.1.1
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/klauspost/compress v1.17.11
+	github.com/microsoft/azure-devops-go-api/azuredevops/v7 v7.1.0
 	github.com/oklog/ulid/v2 v2.1.0
 	github.com/otiai10/copy v1.14.0
 	github.com/patrickmn/go-cache v2.1.0+incompatible
@@ -40,6 +41,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.10.0
 	github.com/technosophos/moniker v0.0.0-20210218184952-3ea787d3943b
+	github.com/tidwall/sjson v1.2.5
 	github.com/valyala/fasttemplate v1.2.2
 	github.com/xeipuuv/gojsonschema v1.2.0
 	gitlab.com/gitlab-org/api/client-go v0.116.0
@@ -121,6 +123,10 @@ require (
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/shopspring/decimal v1.4.0 // indirect
 	github.com/spf13/cast v1.7.0 // indirect
+	github.com/stretchr/objx v0.5.2 // indirect
+	github.com/tidwall/gjson v1.14.2 // indirect
+	github.com/tidwall/match v1.1.1 // indirect
+	github.com/tidwall/pretty v1.2.0 // indirect
 	github.com/valyala/bytebufferpool v1.0.0 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
@@ -197,7 +203,6 @@ require (
 	github.com/prometheus/common v0.55.0 // indirect
 	github.com/prometheus/procfs v0.15.1 // indirect
 	github.com/vbatts/tar-split v0.11.3 // indirect
-	github.com/xanzy/go-gitlab v0.115.0
 	github.com/xlab/treeprint v1.2.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.54.0 // indirect
 	go.opentelemetry.io/otel v1.31.0 // indirect
--- a/internal/directives/json_updater.go
+++ b/internal/directives/json_updater.go
@@ -1,0 +1,150 @@
+package directives
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"slices"
+	"strings"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+	"github.com/tidwall/sjson"
+	"github.com/xeipuuv/gojsonschema"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+)
+
+func init() {
+	builtins.RegisterPromotionStepRunner(newJSONUpdater(), nil)
+}
+
+// jsonUpdater is an implementation of the PromotionStepRunner interface that
+// updates the values of specified keys in a JSON file.
+type jsonUpdater struct {
+	schemaLoader gojsonschema.JSONLoader
+}
+
+// newJSONUpdater returns an implementation of the PromotionStepRunner interface
+// that updates the values of specified keys in a JSON file.
+func newJSONUpdater() PromotionStepRunner {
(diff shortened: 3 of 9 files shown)


docs/docs/35-references/10-promotion-steps.md
```diff
@@ -539,6 +539,61 @@ steps:
...
+`json-update` allows you to update the values of specified keys in any JSON file.
+
+#### `json-update` Configuration
+
+| Field | Type | Required | Description |
+|--------------|------------------|--------------|---------------------------------------------------------------------------------|
+| `path`       | `string`         | Yes          | Path to the JSON file relative to the promotion workspace.                      |
+| `updates`    | `array`          | Yes          | List of key-value pairs to update in the JSON file.                             |
+| `updates.key`| `string`         | Yes          | JSON path (dot-separated) for the key to update.                                |
+| `updates.value`| `any`          | Yes          | Value to set for the specified key. Supports strings, numbers, booleans, arrays, and objects. |
```
krancour, line 553: Edited for consistency with existing step docs:

```suggestion
| Name | Type | Required | Description |
|------|------|----------|-------------|
| `path` | `string` | Y | Path to a JSON file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |                      |
| `updates` | `[]object` | Y | The details of changes to be applied to the file. At least one must be specified. |
| `updates[].key`| `string` | Y | JSON path (dot-separated) for the key to update.                                |
| `updates[].value`| `any` | Y | The new value for the key. Typically specified using an expression. Supports strings, numbers, booleans, arrays, and objects. |
```

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -539,6 +539,61 @@ steps:
...
+- uses: json-update
+  config:
+    path: configs/settings.json
+    updates:
+    - key: app.version
+      value: "v2.0.0"
+    - key: featureFlags.newFeatureEnabled
+      value: true
+    - key: thresholds.memory.limit
+      value: 512
```
krancour, line 583: Something involving an expression would probably be more instructive since this is how the feature is more likely to be used:

```suggestion
    path: configs/settings.json
    updates:
    - key: image.tag
      value: ${{ imageFrom("my/image").Tag }}
```

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -539,6 +539,61 @@ steps:
...
+# Render manifests to ./out, commit, push, etc...
+```
+
+#### `json-update` Output
+
+When `json-update` successfully modifies a JSON file, it generates a `commitMessage` in the following format:
```
krancour, line 589: For consistency with existing step docs:

```suggestion
| Name | Type | Description |
|------|------|-------------|
| `commitMessage` | `string` | A description of the change(s) applied by this step. Typically, a subsequent [`git-commit`](#git-commit) step will reference this output and aggregate this commit message fragment with other like it to build a comprehensive commit message that describes all changes. |
```

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -539,6 +539,61 @@ steps:
 |------|------|-------------|
 | `commitMessage` | `string` | A description of the change(s) applied by this step. Typically, a subsequent [`git-commit`](#git-commit) step will reference this output and aggregate this commit message fragment with other like it to build a comprehensive commit message that describes all changes. |
 
+### `json-update`
+
+`json-update` allows you to update the values of specified keys in any JSON file.
```
krancour, line 544: Edited for consistency with existing step docs:

```suggestion
`json-update` updates the values of specified keys in any JSON file.
```

docs/docs/35-references/10-promotion-steps.md
```diff
@@ -539,6 +539,61 @@ steps:
...
+
+#### `json-update` Output
+
+When `json-update` successfully modifies a JSON file, it generates a `commitMessage` in the following format:
+
+```
+Updated <path>
+  - <key>: <value>
+  - <key>: <value>
+```
```
krancour, line 595: For consistency with existing step docs:

```suggestion
| Name | Type | Description |
|------|------|-------------|
| `commitMessage` | `string` | A description of the change(s) applied by this step. Typically, a subsequent [`git-commit`](#git-commit) step will reference this output and aggregate this commit message fragment with other like it to build a comprehensive commit message that describes all changes. |
```

internal/directives/schemas/json-update-config.json
```diff
@@ -0,0 +1,44 @@
...
+      "properties": {
+        "key": {
+          "type": "string",
+          "description": "The key whose value needs to be updated. For nested values, use a JSON dot notation path.",
+          "minLength": 1
+        },
+        "value": {
+          "type": ["string", "number", "boolean", "object", "array", "null"],
+          "description": "The new value for the specified key."
+        }
```
krancour, line 19: Because the codegen used to create Go structs from these schemas is a bit temperamental, I suggest creating a definition for a type `jsonValue` and using that here. It will improve the generated code so that the struct is called `JsonValue`, which looks less generic than `Value`. It will avoid the need to refactor schemas and/or step implementations in the future if/when the need for some _other_ kind of "value" type arises.

go.mod
```diff
@@ -121,6 +121,10 @@ require (
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/shopspring/decimal v1.4.0 // indirect
 	github.com/spf13/cast v1.7.0 // indirect
+	github.com/tidwall/gjson v1.18.0 // indirect
+	github.com/tidwall/match v1.1.1 // indirect
+	github.com/tidwall/pretty v1.2.1 // indirect
+	github.com/tidwall/sjson v1.2.5 // indirect
```
krancour, line 127: I'm not sure what happened here, but you probably need to re-run `go mod tidy` as this package should have appeared as a direct dependency and not an indirect one.

internal/directives/zz_config_types.go
```diff
@@ -436,3 +450,12 @@ const (
...
 )
+
+// The new value for the specified key.
+type Value struct {
+	AnythingArray []interface{}
+	AnythingMap   map[string]interface{}
+	Bool          bool
+	Double        int64
+	String        string
+}
```
krancour, line 461: This generated code is quirky. Again... the tool we're using for this bit of codegen is temperamental. It's worth playing with the schema more to see if we can't coerce this into something that will work better.

`Double int64` is obviously wrong.

internal/directives/json_updater.go
```diff
@@ -0,0 +1,166 @@
...
+		if update.Value != nil {
+			switch {
+			case update.Value.String != "":
+				value = update.Value.String
+			case len(update.Value.AnythingArray) > 0:
+				value = update.Value.AnythingArray
+			case len(update.Value.AnythingMap) > 0:
+				value = update.Value.AnythingMap
+			case update.Value.Double != 0:
+				value = update.Value.Double
```
krancour, line 82: This means you can never set the value of a number field to 0.

This is a consequence of the quirky code generation, which we will need to work around somehow. It may require some experimentation.

[This comment](https://github.com/akuity/kargo/pull/3151/files#r1889056508) is related.

internal/directives/json_updater.go
```diff
@@ -0,0 +1,166 @@
...
+func (j *jsonUpdater) validate(cfg Config) error {
+	return validate(j.schemaLoader, gojsonschema.NewGoLoader(cfg), j.Name())
+}
+
+func (j *jsonUpdater) runPromotionStep(
+	_ context.Context,
+	stepCtx *PromotionStepContext,
+	cfg JSONUpdateConfig,
+) (PromotionStepResult, error) {
+	updates := make(map[string]any, len(cfg.Updates))
```
krancour, line 70: What is the point of building this map? You're iterating over the list to build the map, then iterating over the map. Seems the former is unnecessary.

krancour:
I think the library you're using already accounts for preserving the formatting of the original JSON, but it would be good to add a test case that verifies that, because this is an incredibly important aspect of this.

internal/directives/schemas/json-update-config.json
krancour: Did a bit of experimentation for you. If you omit the `type` entirely, you'll end up with this:

```
type JSONUpdate struct {
	// The key whose value needs to be updated. For nested values, use a JSON dot notation path.
	Key string `json:"key"`
	// The new value for the specified key.
	Value interface{} `json:"value"`
}
```

And that should be quite a bit more pleasant to work with and should help resolve many of my other comments.

internal/directives/json_updater.go
krancour: See [this comment](https://github.com/akuity/kargo/pull/3151/files#r1889083798) also.

That should help.

internal/directives/json_updater.go
```diff
@@ -0,0 +1,166 @@
...
+	for _, update := range cfg.Updates {
+		var value any
+		if update.Value != nil {
+			switch {
+			case update.Value.String != "":
+				value = update.Value.String
+			case len(update.Value.AnythingArray) > 0:
+				value = update.Value.AnythingArray
+			case len(update.Value.AnythingMap) > 0:
+				value = update.Value.AnythingMap
```
krancour, line 80: I think I would consider initially supporting only scalar values -- strings, numbers, booleans, and nulls. I think that with arrays and objects/maps in the picture, "there be dragons here." Among other potential problems is the difficulty posed in distilling such changes into a commit message -- which I see you're currently not doing.

Another problem with supporting non-scalar values is that although we can preserve formatting choices in the existing file, we won't be able to make the modifications _match_ the formatting choices of the existing file.

i.e. If you start with:

```
{
  "name": "Bart",
  "address": {
    "street": "42 Evergreen Terrace",
    "city": "Springfield"
  }
}
```

And you update address, you could end up with something like this:

```
{
  "name": "Bart",
  "address": { "street": "123 Fake Street", "city": "Springfield" }
}
```

It's not invalid, but I can imagine users being annoyed with it.

Altogether, I think it's better to start off with what we know we can do well and look to improve it over time if there's a demonstrated need for it.

I think just supporting scalars, would address the vast majority user cases.

krancour:
@fykaa this is very good overall. I know I made a lot of comments, but the majority of them can be directly traced back to the quirkiness of the codegen that creates Go structs from JSON schema.

I think with some slight modifications to the schema, everything else will fall into place.

internal/directives/zz_config_types.go
krancour: [This](https://github.com/akuity/kargo/pull/3151#discussion_r1889083798) will help.

internal/directives/json_updater.go
krancour: > I took inspiration for this from the [yaml-updater](https://github.com/akuity/kargo/blob/main/internal/directives/yaml_updater.go#L69)

I suspected this was the case. 😄 

There are two not-entirely-obvious reasons not to emulate that:

1. `yaml.SetStringsInFile()` wants a map as input. The `yaml-update` step _had_ to do that transformation, whereas here, you don't have any need to do so. You have the luxury of just working with the array you already have in-hand.

2. `yaml.SetStringsInFile()` taking a map as input was misguided in the first place. At the time I wrote it, we were only using it to update _always-pre-existing_ keys in a values.yaml or Chart.yaml file. A map of existing keys to new values served that purpose well, but...

  The more-generic `yaml-update` step was added more recently and I should have accounted better for the possibility of it being used to _add new elements to a list_, and in that case, order matters and order is lost in the transformation to a map.
  
  _I'm actually going to fix this today._

> the map will help in data integrity by deduplicating keys

It won't matter. Since the order of the updates will be preserved, the last one wins -- which is identical behavior to what you get if you iterate over a list and transform it to a map.

Marvin9:
if you could please update UI code as well - read [this PR](https://github.com/akuity/kargo/pull/2987) for reference (skip the icons)

internal/directives/json_updater.go
```diff
@@ -0,0 +1,150 @@
...
+
+	for key, value := range changes {
+		updatedContent, setErr := sjson.Set(string(fileContent), key, value)
+		if setErr != nil {
+			return fmt.Errorf("error setting key %q in JSON file: %w", key, setErr)
+		}
+		fileContent = []byte(updatedContent)
+	}
+
+	err = os.WriteFile(absFilePath, fileContent, 0o600)
```
krancour, line 114: Nit: This notation is more common.

```suggestion
	err = os.WriteFile(absFilePath, fileContent, 0600)
```

internal/directives/json_updater.go
```diff
@@ -0,0 +1,150 @@
...
+		value := updates[key]
+		switch v := value.(type) {
+		case string:
+			_, _ = commitMsg.WriteString(fmt.Sprintf("\n- %s: %q", key, v))
+		case bool:
+			_, _ = commitMsg.WriteString(fmt.Sprintf("\n- %s: \"%v\"", key, v))
+		case int, float64:
+			_, _ = commitMsg.WriteString(fmt.Sprintf("\n- %s: %v", key, v))
+		default:
+			_, _ = commitMsg.WriteString(fmt.Sprintf("\n- %s: %v", key, v))
```
krancour, line 145: This is potentially going to be very funky-looking and it was part of why I suggested only supporting scalar values for now.

internal/directives/json_updater.go
```diff
@@ -0,0 +1,150 @@
...
+		}
```
krancour, line 146:
```suggestion
		switch v := value.(type) {
		case string:
			_, _ = commitMsg.WriteString(fmt.Sprintf("\n- %s: %q", key, v))
		default:
			_, _ = commitMsg.WriteString(fmt.Sprintf("\n- %s: %v", key, v))
		}
```

As long as we're only working with scalars, your default case works well for everything _except_ string.

internal/directives/json_updater.go
```diff
@@ -0,0 +1,150 @@
...
+		return fmt.Errorf("error joining path %q: %w", path, err)
+	}
+
+	fileContent, err := os.ReadFile(absFilePath)
+	if err != nil {
+		return fmt.Errorf("error reading JSON file %q: %w", absFilePath, err)
+	}
+
+	for key, value := range changes {
+		updatedContent, setErr := sjson.Set(string(fileContent), key, value)
```
krancour, line 107: Right above here, I'd be verifying that `value` is scalar.

Something like this would work:

```
switch value.(type) {
case int, int8, int16, int32, int64,
		uint, uint8, uint16, uint32, uint64,
		float32, float64,
		string, bool:
default:
		// return an appropriate error
}
```

internal/directives/json_updater.go
krancour: To perhaps add some clarity...

When there's a map involved, you can't do this:

```
config:
    path: configs/settings.json
    updates:
    - key: list.-1
      value: foo
    - key: list.-1
      value: bar
```

But that should be valid.

You can see I fixed how yaml-update does this in #3156

internal/directives/schemas/json-update-config.json
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "JSONUpdateConfig",
+
+  "definitions": {
+
```
hiddeco, line 6:
```suggestion
  "definitions": {
```

internal/directives/json_updater.go
```diff
@@ -0,0 +1,145 @@
...
+			return PromotionStepResult{Status: kargoapi.PromotionPhaseErrored},
+				fmt.Errorf("JSON file update failed: %w", err)
+		}
+
+		if commitMsg := j.generateCommitMessage(cfg.Path, cfg.Updates); commitMsg != "" {
+			result.Output = map[string]any{
+				"commitMessage": commitMsg,
+			}
+		}
+	}
```
krancour, line 81: Nit: I'd remove the conditional. We validated up front that there was at least one update to perform and even if somehow the updates passed into here were nil or empty, it's still perfectly safe to iterate over them.

krancour (approved):

krancour (approved):
Thank you for all the hard work on this!

PR #3152, opened 2024-12-17
feat(controller): make argocd-update step honor Argo CD sync windows

### LINKED ISSUE #2078: add docs about argocd-update step and Argo CD sync windows

krancour:
Fixes #2078  

This is a straightforward way to begin honoring Argo CD sync windows.

Note that in an ideal world, the Argo CD controller would enforce these itself. As things currently stand, sync windows are enforced by the Argo CD API server.

This PR expands our "stubbed out" Argo CD resource types to include AppProject and a subset of its fields, which we only ever _read_.

This also copies sync window enforcement logic directly from Argo CD without any modification except what was required to clear our linters, which are more strict.

The only interesting bit of this PR is what's in ece12c56c826db99de4d4617a913a220f1da68fd.

If a sync window is determined to be closed, the argocd-update step will return with a Running status -- meaning the Promotion will be re-reconciled at a later time and the step will be tried again in hopes of finding the window is now open.

Users should adjust the timeout on the step to exceed the duration of the longest closed window likely to be encountered. This is addressed in the included documentation updates.

cc @jessesuen 

Sharing credit with @rbreeze as well, since this PR is nearly identical to #2346, but with merge conflicts resolved and docs added.


blakepettersson:
>Note that in an ideal world, the Argo CD controller would enforce these itself. As things currently stand, sync windows are enforced by the Argo CD API server.

Unless I misunderstand something Argo CD has done this since argoproj/argo-cd#16492 (from Argo CD 2.11)?

krancour:
@blakepettersson well that's interesting. It certainly appears that way. Embarrassingly... my argo-cd master branch hadn't been synced to upstream since before 2.11, so I was referencing code from before this change. 🤦‍♂️

But...

I'm about 95% certain that I've _recently_ encountered the issue described in #2078, and I think @jessesuen has as well?

This definitely merits a little more research. Thank you for calling attention to it!

krancour:
I should really do less GitHubbing from my phone.

> Embarrassingly... my argo-cd master branch hadn't been synced to upstream since before 2.11, so I was referencing code from before this change.

Back in front of my computer, this turns out not to be the case. I'd pulled the latest from mater just a few days ago.

> I'm about 95% certain that I've recently encountered the issue described in https://github.com/akuity/kargo/issues/2078, and I think @jessesuen has as well?

Wrong again. I'm sure I replicated #2078, _but it couldn't have been recently_. With a little research... #2078 was opened in May, which is long before the fix was introduced on the Argo CD end. The Kargo team was developing against v2.9 at the time. With a fresh attempt this morning, I _cannot_ replicate #2078 with Argo CD v2.13, but _can_ with v.2.9. So it must have been quite some time ago that I saw #2078 happen with my own 👀

Whether this PR should be closed or merged may come down to a question of what's the oldest Argo CD version we wish to support. Alternatively, we can ask what is the likelihood that someone's willing to upgrade their Kargo to the very latest version to get this workaround, but _not_ willing to upgrade their Argo CD to something as relatively old as v2.11 to get the correct fix? It seems small.

My own preference is to close this and open a smaller PR that just amends the docs for the `argocd-update` PR to advise setting the timeout appropriately if there is a possibility of encountering an active deny window.

But I'd like @jessesuen to weight in.

@blakepettersson thanks again for catching this!

krancour:
Discussed with @jessesuen offline. Closing this.

PR #3156, opened 2024-12-19
feat: make yaml-update syntax the same as json-update syntax

krancour:
This complements #3151.

#3151 uses tidwall/sjson for modifying JSON documents whilst preserving their existing formatting.

The keys/paths supported by that library use simple dot-separated notation. Dots can be escaped. Colons are used as a hint that a numeric-looking key part should be treated as a key in an object and not as an index in a sequence. Colons also can be escaped.

Examples:

* foo.bar --> [foo bar]
* foo\.bar --> [foo.bar]
* foo.2 --> [foo 2] (2 is an index in the sequence foo)
* foo.:2 --> [foo 2] (2 is a key in the object foo)
* foo\:bar --> [foo:bar]

-1 can be used as an index to indicate a desire to append to a sequence.

__This is slightly inconsistent with the syntax supported by our yaml-update step.__

* Indices are handled the way Helm does. e.g. foo.[2]
* Negative indices not allowed at all -- appending to a sequence was not possible. (Technically it was, but only if you knew the length of the sequence, which effectively made it useless.)
* No provision for escaped dots

__These slightly different syntaxes were inevitably going to lead to confusion.__

This PR updates the internal `yaml` library we use for editing YAML whilst preserving comments and formatting choices, and by extension the yaml-update step, to use the same syntax as tidwall/sjson and the new json-update step.

__There are no breaking changes in this PR. The legacy syntax still works.__

Since these changes open the possibility of using our internal `yaml` package, and by extension the yaml-update step, to append to a sequence (which it couldn't do before), the notion of using a _map_ of key/value pairs as input to update operations is invalidated. If we can append to a sequence, order matters now...

So to enable all of the above, this PR also contains a small bit of refactoring to transition away from using maps as input to YAML update operations and toward using arrays of yaml.Update structs.

cc @fykaa

### DIFF
--- a/docs/docs/35-references/10-promotion-steps.md
+++ b/docs/docs/35-references/10-promotion-steps.md
@@ -551,7 +551,7 @@ followed by a [`helm-template`](#helm-template) step.
 |------|------|----------|-------------|
 | `path` | `string` | Y | Path to a YAML file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
 | `updates` | `[]object` | Y | The details of changes to be applied to the file. At least one must be specified. |
-| `updates[].key` | `string` | Y | The key to update within the file. For nested values, use a YAML dot notation path. |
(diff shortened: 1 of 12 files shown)


hiddeco (approved):
💯

Issue #2833, opened 2024-10-24
docs: update go/crd docs with semantic version info


fykaa:
ref: https://github.com/akuity/kargo/pull/2521#discussion_r1814010960

krancour:
> We can move this to be inline in the Go/CRD docs.

It turns out it is already mentioned here:

https://github.com/akuity/kargo/blob/1a7ab783e48cd04be0343bf4e1ff689090352fc3/api/v1alpha1/warehouse_types.go#L241-L253

(Although the comment about referring to Image Updater documentation ought to be removed. That is leftover from a time when image selection functionality was borrowed from Image Updater. That is no longer the case.)

This is another spot where semver constraints are mentioned -- this time for commit selection:

https://github.com/akuity/kargo/blob/1a7ab783e48cd04be0343bf4e1ff689090352fc3/api/v1alpha1/warehouse_types.go#L137-L146

Run codegen after updating these comments.

krancour:
@fykaa you could consider addressing #2880 in tandem with this as they are very closely related.

PR #3167, opened 2024-12-20
chore: remove image updater info

fykaa:
closes https://github.com/akuity/kargo/issues/2833

part of it was merged in https://github.com/akuity/kargo/pull/2963

### DIFF
--- a/api/v1alpha1/warehouse_types.go
+++ b/api/v1alpha1/warehouse_types.go
@@ -249,7 +249,7 @@ type ImageSubscription struct {
 	// constraints, which means the latest semantically tagged version of an image
 	// will always be used. Care should be taken with leaving this field
 	// unspecified, as it can lead to the unanticipated rollout of breaking
-	// changes. Refer to Image Updater documentation for more details.
+	// changes.
 	// More info: https://github.com/masterminds/semver#checking-version-constraints
 	//
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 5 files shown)


api/v1alpha1/warehouse_types.go
```diff
@@ -249,8 +249,7 @@ type ImageSubscription struct {
 	// constraints, which means the latest semantically tagged version of an image
 	// will always be used. Care should be taken with leaving this field
 	// unspecified, as it can lead to the unanticipated rollout of breaking
-	// changes. Refer to Image Updater documentation for more details.
-	// More info: https://github.com/masterminds/semver#checking-version-constraints
+	// changes.
```
krancour, line 252: This may have been poorly worded by me initially and it may have caused you some confusion.

Image Updater and masterminds/semver are two different things.

We _do_ want to remove anything that refers readers to Image Updater for more info because we're no longer using Image Updater anywhere in Kargo.

masterminds/semver is a library we use for parsing and comparing semantic version strings. (Image Updater happens to use it as well. It's a popular library.)

At any rate... we should _keep_ references to masterminds/semver docs as they're still the definitive source of information for how version constraints work.

krancour: This applies throughout the PR.

krancour (approved):

PR #3173, opened 2024-12-21
chore: enforce LF line endings in git

Sobuno:
Adding a .gitattributes file to normalize on LF for line endings in text files even if someone has their Git config set up to use CRLF.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
hiddeco, line 1:
```suggestion
* text=auto eol=lf

```

hiddeco (approved):

PR #3174, opened 2024-12-22
fix(ui): link to image repo correctly in Warehouse UI

Sobuno:
Link was pointing relative to the current URL rather than to a repository. Reusing `urlForImage` function that was used for linking in the repo node already.

Before (See bottom of screenshot):
image

Now (See bottom of screenshot):
image


Marvin9 (approved):
LGTM. Thanks

Marvin9:
@Sobuno Could you just fix the UI lint and then it is good to go

Issue #3163, opened 2024-12-19
incorrectly worded error message form argocd-update step causes confusion


krancour:
```
time="2024-12-19T19:03:00Z" level=info msg="current operation was not initiated by \"kargo-controller\" and not by \"\": waiting for operation to complete" freight=b764d720b186a9d73685a01f7600e8a4b0f03727 namespace=akuity-platform
```

This was meant to say something like "was initiated by <actual initiator> and not by kargo-controller."

As worded, that's not the meaning it's conveying.

RohanMishra315:
Hey @krancour should I look into it ?

PR #3192, opened 2024-12-24
fix: incorrectly worded error message in argocd-update is corrected

RohanMishra315:
fixes : #3163
1. Error Message Adjustment:

Changed the phrasing from "was not initiated by %q and not by %q" to "was initiated by %q and not by %q".
This directly states who initiated the operation (status.Operation.InitiatedBy.Username) and clarifies it wasn't initiated by the expected user (applicationOperationInitiator).

### DIFF
--- a/internal/directives/argocd_updater.go
+++ b/internal/directives/argocd_updater.go
@@ -388,8 +388,8 @@ func (a *argocdUpdater) mustPerformUpdate(
 			// NB: We return the current phase here because we want the caller
 			//     to know that an operation is still running.
 			return status.Phase, false, fmt.Errorf(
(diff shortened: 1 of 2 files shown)


krancour:
@RohanMishra315, it looks like there's a test case that is sensitive to the wording in that error and needs to be adjusted accordingly:

https://github.com/akuity/kargo/blob/011f87af218379419d5d44c29eb5d676200b5ebc/internal/directives/argocd_updater_test.go#L981

After that's corrected, LGTM.

Thank you for this!

krancour (approved):
Thank you @RohanMishra315!

PR #3218, opened 2025-01-06
feat: Mount path /tmp as emptyDir volume 

### LINKED ISSUE #3150: Mount path /tmp as emptyDir volume
# Checklist

# Proposed Feature

The path `/tmp` is used in the promotion step *git clone* to save the cloned git repository. Currently, this path is part of the container's root filesystem. I'm proposing to mount an ephemeral emptyDir volume to this path.

# Motivation

We are about to roll out Kargo to our production stage. One of the compliance requirements is to enable the `securityContext` `readOnlyRootFilesystem`. This protects the root filesystem of the container from being changed during runtime and therefore increases security. When we enable this, the promotion step *git clone* fails with the following error:

image

# Suggested Implementation

I see two possible implementation options in the Kargo Helm Chart:
1. Allow to mount volumes to any of the containers. Currently, it is only possible to mount volumes to the dex container.
2. Mount an emptyDir volume in the path `/tmp` to the container by default.

ggogel:
Closes #3150

Mounting an emptyDir volume to the path `/tmp`, which Kargo uses as target for cloning git repositories. This allows enabling the securityContext readOnlyRootFilesystem.

I deployed the chart to our test environment and tested it with the quickstart tutorial from the docs.


charts/kargo/templates/controller/deployment.yaml
```diff
@@ -79,8 +79,9 @@ spec:
         {{- with (concat .Values.global.envFrom .Values.controller.envFrom) }}
           {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or .Values.kubeconfigSecrets.kargo .Values.kubeconfigSecrets.argocd .Values.controller.gitClient.signingKeySecret.name .Values.controller.cabundle.configMapName .Values.controller.cabundle.secretName }}
         volumeMounts:
+        - mountPath: /tmp
+          name: tmp-data
```
krancour, line 84:
```suggestion
          name: tmpData
```

charts/kargo/templates/controller/deployment.yaml
```diff
@@ -127,8 +128,9 @@ spec:
         - name: certs
           mountPath: /tmp/target
       {{- end }}
-      {{- if or .Values.kubeconfigSecrets.kargo .Values.kubeconfigSecrets.argocd .Values.controller.gitClient.signingKeySecret.name .Values.controller.cabundle.configMapName .Values.controller.cabundle.secretName }}
       volumes:
+      - name: tmp-data
```
krancour, line 132:
```suggestion
      - name: tmpData
```

krancour (approved):

krancour (approved):
Thank you @ggogel!

PR #3220, opened 2025-01-06
chore: add api.oidc.forceHttpsInBaseUrl in values.yaml to support external TLS termi…

AdmirHad:
This pull request introduces a new configuration option, forceHttps, to the Kargo Helm chart. This enhancement allows users to enforce the use of https in the kargo.api.baseURL, even when TLS options are disabled for the API and ingress. This is particularly useful for scenarios where the API is exposed behind an AWS Load Balancer that terminates SSL certificates.

Changes Made
_helpers.tpl: Updated the kargo.api.baseURL helper function to include a condition that checks the new forceHttps value. If forceHttps is set to true, the base URL will use https regardless of other TLS settings.

values.yaml: Added the forceHttps configuration option under the api section with a default value of true.

Use Case
This change is designed to support environments where the API is secured by external mechanisms, such as an AWS Load Balancer, and internal TLS settings are not required. By setting forceHttps to true, users can ensure that the API base URL is always constructed with https.

Testing
Deployed the updated Helm chart in a test environment with forceHttps set to true and verified that the API base URL uses https.
Tested with forceHttps set to false to ensure the base URL respects the existing TLS settings.

Documentation
Updated the values.yaml file with comments explaining the new forceHttps option.

### DIFF
--- a/charts/kargo/templates/_helpers.tpl
+++ b/charts/kargo/templates/_helpers.tpl
@@ -92,7 +92,7 @@ app.kubernetes.io/component: webhooks-server
 {{- end -}}
 
 {{- define "kargo.api.baseURL" -}}
-{{- if or .Values.api.tls.enabled (and .Values.api.ingress.enabled (or .Values.api.ingress.tls.enabled .Values.api.ingress.tls.usesControllerCert)) -}}
+{{- if or .Values.api.oidc.forceHttpsInBaseUrl (or .Values.api.tls.enabled (and .Values.api.ingress.enabled (or .Values.api.ingress.tls.enabled .Values.api.ingress.tls.usesControllerCert))) -}}
 {{- printf "https://%s" .Values.api.host -}}
 {{- else -}}
 {{- printf "http://%s" .Values.api.host -}}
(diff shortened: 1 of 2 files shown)


krancour:
@AdmirHad this is almost exactly what `api.ingress.tls.usesControllerCert` is for.

It was added in #3158 to cover the edge case of the ingress controller terminating ssl using its _own_ cert (i.e. not one specific to the Ingress) and using plain http between the ingress controller and the Kargo API server, but really this should work for _any_ load balancer or reverse proxy that's terminating ssl.

This feature is only in `main` at the moment, which may be why you weren't aware of it.

AdmirHad:
Cool, I'll drop the PR.

Can you please tell me when will you release the next helm chart version?

krancour:
> Can you please tell me when will you release the next helm chart version?

We'd hope to have it released by now, but quality supersedes self-imposed deadlines and we've got a couple things holding us up. Best _guess_ is sometime next week.

But... the timing might be fortuitous here.

`api.ingress.tls.usesControllerCert` _does the trick_, but your use case got me thinking about the fact that someone could encounter the need for this even if they're _not_ using ingress.

Since `api.ingress.tls.usesControllerCert` hasn't made it into a release yet, I'm thinking about moving it to something like `api.tls.terminatedUpstream`

Issue #3171, opened 2024-12-21
docs: make multiple versions available from a drop down in the header


krancour:
This was briefly discussed the other day.

Each time we have a minor release, that release's branch becomes the branch that the main docs are served from. Older release branches move to subdomains like release-1-1.docs.kargo.io

We need to add a drop down to the header so that docs for older releases can be selected. It just needs to redirect readers to the correct subdomain.

I think @rpelczar mentioned that Docusaurus may actually have some built-in feature to help with this, although I've not explored that.

rpelczar:
Yes, it has a built-in feature: https://docusaurus.io/docs/versioning

PR #3226, opened 2025-01-08
feat: versioned docs dropdown

fykaa:
closes https://github.com/akuity/kargo/issues/3171

This pull request implements a version dropdown in the Docusaurus configuration (`docusaurus.config.js`) to support versioned documentation. The following changes were made:

1. Added version dropdown to Navbar:
   - Updated `docusaurus.config.js` to include a `docsVersionDropdown` in the navbar letting users to switch between different doc versions

2. Versioned doc creation:
   - Ran `npm run docusaurus docs:version <version>` for each version (`1.0`, `1.1`) to generate versioned docs.
   - Updated the `versions.json` file to include the new versions.

**Next Steps:** update release procedure
- To maintain the latest doc with each minor/ patch update, the process of running `npm run docusaurus docs:version <version>` should be added to the release procedures

image


krancour:
I see that this is based on what @rpelczar had linked you to, but this process is frankly much to onerous and prone to human error. It doesn't satisfy this requirement that I included in #3171:

> Each time we have a minor release, that release's branch becomes the branch that the main docs are served from. Older release branches move to subdomains like release-1-1.docs.kargo.io
>
> We need to add a drop down to the header so that docs for older releases can be selected. It just needs to redirect readers to the correct subdomain.

This would introduce no new complications or process changes.

krancour (changes requested):
See above.

krancour:
> Older release branches move to subdomains like release-1-1.docs.kargo.io

> It's possible that a plugin exists for this. If not, I think this would be a very easy plugin to write.

I looked for one and didn't find one. It doesn't mean it's not our there somewhere.

Rolling our own is really not too bad. It could probably be as simple as:

1. Look at `window.location.hostname` to determine what version the user is currently viewing. If it's `docs.kargo.io`, they're viewing current docs. If it is a release docs subdomain, parse it to learn the major and minor revision numbers. This gives you everything you need to make something like a very slim banner at the top of the page that says something like "You're viewing documentation for v1.1.x". For any other value of `window.location.hostname`, version number is "unknown".

2. You can position a dropdown for "Other Versions" to the right of it. To find the version numbers, you can hit https://api.github.com/repos/akuity/kargo/branches to get the names of all branches. Use a regex to identify all GA release branches and cache it for some short period of time in the browser's local storage. Use that to populate the dropdown.

After this is working to our satisfaction in main, we'd backport it to release-1.1 and release-1.0.

krancour:
@fykaa don't get caught up in building or deploying docs.

Please just assume that for every branch release-x-y with x > 1, the subdomain will exist. If one of them does not, I can fix it.

The reason that the 1.0 docs don't currently exist has to do with the migration we did from docs.kargo.akuity.io to docs.kargo.io. iirc, there were some changes related to the domain change that didn't get backported that far, so I just didn't deploy them. I'll investigate and fix it.

krancour:
> iirc, there were some changes related to the domain change that didn't get backported that far, so I just didn't deploy them

It was dumber than that. Something just got stuck on their end during the migration.

Should be all set now.

All of the following exist:

* docs.kargo.io (currently published from release-1.1)
* release-1-1.docs.kargo.io (published from release-1.1)
* release-1-0.docs.kargo.io (published from release-1.0)
* main.docs.kargo.io ("edge" docs, published from main)

docs/plugins/versioning/ui/components/NavbarDropdown.js
```diff
@@ -0,0 +1,88 @@
...
+      const branches = await response.json();
+      console.log("fetched branches are: ", branches);
+
+      const releaseBranches = branches
+        .map(branch => branch.name)
+        .filter(name => /^release-1/.test(name))
+        .map(name => {
+          const [major, minor] = name.split('.');
+          return {
+            version: `${major}-${minor}`,
```
krancour, line 31:
```suggestion
            version: `v${major}.${minor}`,
```

docs/plugins/versioning/ui/components/NavbarDropdown.js
```diff
@@ -0,0 +1,88 @@
...
+  const [currentVersion, setCurrentVersion] = useState("");
+
+  useEffect(() => {
+    fetchVersions();
+  }, []);
+
+  const fetchVersions = async () => {
+    try {
+      console.log("Before fetching versions")
+      const response = await fetch(githubApiUrl);
```
krancour, line 21: Shouldn't you look in local storage instead of unconditionally making this request?

docs/plugins/versioning/ui/components/NavbarDropdown.js
```diff
@@ -0,0 +1,88 @@
...
+      const branches = await response.json();
+      console.log("fetched branches are: ", branches);
+
+      const releaseBranches = branches
+        .map(branch => branch.name)
+        .filter(name => /^release-1/.test(name))
```
krancour, line 27: This looks brittle. It will break down if/when we ever have `release-2.0`. It will also select branches like `release-11`, which lacks a `.` and will fail the logic below. You'll have to come up with a regex that is a bit more selective.

If you've not used it before, https://regex101.com/ is very helpful.

docs/plugins/versioning/ui/components/NavbarDropdown.js
```diff
@@ -0,0 +1,88 @@
...
+        .map(name => {
+          const [major, minor] = name.split('.');
+          return {
+            version: `${major}-${minor}`,
+            url: `https://${major}-${minor}.docs.kargo.io${window.location.pathname}`
```
krancour, line 32: I would not include the `${window.location.pathname}`. There's no guarantee that some different version of the docs _has_ that path. I would say the likelihood of 404s is exceptionally high.

docs/plugins/versioning/ui/components/NavbarDropdown.js
```diff
@@ -0,0 +1,88 @@
+import React, { useState, useEffect } from 'react';
+import styles from './NavbarDropdown.module.css';
+
+const CACHE_KEY = 'kargo-docs-versions';
+const CACHE_DURATION = 1000 * 60 * 30; // 30 minutes
+const githubApiUrl = 'https://api.github.com/repos/akuity/kargo/branches';
+
+function NavbarDropdown() {
```
krancour, line 8: `NavbarDropdown()` feels rather unspecific as to what this component does. Would it not be better to name this `VersionDropdown`?

docs/plugins/versioning/ui/components/NavbarDropdown.js
```diff
@@ -0,0 +1,88 @@
...
+      releaseBranches.unshift({
+        version: 'current',
+        url: `https://docs.kargo.io${window.location.pathname}`
+      });
+      console.log("These are release branches: ", releaseBranches)
+
+      localStorage.setItem(CACHE_KEY, JSON.stringify({
+        versions: releaseBranches,
+        timestamp: Date.now()
+      }));
```
krancour, line 46: You're setting this and never looking at it. See [other comment](https://github.com/akuity/kargo/pull/3226/files#r1913370317).

docs/plugins/versioning/ui/components/NavbarDropdown.js
```diff
@@ -0,0 +1,88 @@
...
+            version: `${major}-${minor}`,
+            url: `https://${major}-${minor}.docs.kargo.io${window.location.pathname}`
+          };
+        });
+        console.log("These are release branches before unshifting: ", releaseBranches)
+      // Add current version
+      releaseBranches.unshift({
+        version: 'current',
+        url: `https://docs.kargo.io${window.location.pathname}`
+      });
```
krancour, line 40: "Current" might be a poor word choice. I know you mean it as the docs for the _latest_ version, but I think will be too easily misinterpreted as the version of the docs currently selected by the dropdown.

tbh, I wouldn't bother with anything like this.

I would suggest that the title of the dropdown menu item be the _currently selected_ version and when expanded, can unconditionally show all the versions (including the currently selected one).

krancour:
@fykaa while reviewing this, I've come to realize that we don't need a _plugin_ after all. We just need a _component_ and that's far easier to achieve.

I just pushed a couple commits. I didn't change anything you've done (except removing the bit that messed with webpack config because that's not required now). I've just moved some things around.

The other issues I commented on earlier, such as not caching properly in local storage still remain for you to resolve.

docs/src/components/VersionDropdown.js
```diff
@@ -0,0 +1,97 @@
+import React, { useState, useEffect } from 'react';
+
+const CACHE_KEY = 'kargo-docs-versions';
+const CACHE_DURATION = 1000 * 60 * 30; // 30 minutes
+const githubApiUrl = 'https://api.github.com/repos/akuity/kargo/branches';
```
hiddeco, line 5:
```suggestion
const githubApiUrl = 'https://api.github.com/repos/akuity/kargo/branches?protected=true';
```

would piggyback on the fact that we use branch protection for all our release branches, while reducing the size of the payload.

docs/src/components/VersionDropdown.js
```diff
@@ -23,7 +33,7 @@ function VersionDropdown() {
 
       const releaseBranches = branches
         .map(branch => branch.name)
-        .filter(name => /^release-1/.test(name))
+        .filter(name => /^release-(1[.]\d+|[2-9]\d*[.]\d+)$/.test(name))
```
krancour, line 36:
```suggestion
        .filter(name => /^release-([1-9]\d*\.\d+)$/.test(name))
```

docs/src/components/VersionDropdown.js
```diff
@@ -0,0 +1,97 @@
...
+            version: `v${major}.${minor}`,
+            url: `https://${name.replace('.', '-')}.docs.kargo.io`
+          };
+        });
+        console.log("These are release branches before unshifting: ", releaseBranches)
+      // Add current version
+      releaseBranches.unshift({
+        version: 'Latest Version',
+        url: `https://docs.kargo.io`
+      });
```
krancour, line 49:
```suggestion
      console.log("These are release branches before sorting and updating 0 element: ", releaseBranches)
      releaseBranches.sort((a, b) => {
        if (a.version > b.version) {
          return -1;
        }
        if (a.version < b.version) {
          return 1;
        }
        return 0;
      });
      // Overwrite the first element with the latest version
      releaseBranches[0] = {
        version: 'Latest Version',
        url: `https://docs.kargo.io`
      };
```

krancour: I think you're also going to have to look at `window.location` to get the domain and determine which version is currently selected. As it is right now, you will see `Latest Version` in your navbar no matter which one you're actually looking at.

docs/plugins/versioning/ui/components/NavbarDropdown.js
krancour: Please don't worry about that for right now.

krancour (approved):

PR #3234, opened 2025-01-09
Fix warehouse regex for branch names

tal-hason:
the regex expression for the branch name was too limited and didn't accept branch name such as "release-0.3"

updated the regex from:
```
^\w+([-/]\w+)*$
```

to:
```
^\w+([-\/]\w+)*(-\d+\.\d+)?$
```

image


tal-hason:
reopening from PR list

PR #3235, opened 2025-01-09
fix: warehouse branch regex that was to limiting

tal-hason:
in the warehouse when selecting git subscription the regex was:
`^\w+([-/]\w+)*$`

what didn't accept the branch name like: release-0.3

I have updated the warehouse branch regex to: `^\w+([-\/]\w+)*(-\d+\.\d+)?$`

so now it accept branch names:
image

image

### DIFF
--- a/api/v1alpha1/warehouse_types.go
+++ b/api/v1alpha1/warehouse_types.go
@@ -124,7 +124,7 @@ type GitSubscription struct {
 	// subscription is implicitly to the repository's default branch.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=`^\w+([-\/]\w+)*(-\d+\.\d+)?$`
+	// +kubebuilder:validation:Pattern=`^\w+([-/\.]\w+)*$`
 	Branch string `json:"branch,omitempty" protobuf:"bytes,3,opt,name=branch"`
 	// StrictSemvers specifies whether only "strict" semver tags should be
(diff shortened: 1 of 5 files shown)


api/v1alpha1/warehouse_types.go
```diff
@@ -124,7 +124,7 @@ type GitSubscription struct {
 	// subscription is implicitly to the repository's default branch.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Pattern=`^\w+([-/]\w+)*$`
+	// +kubebuilder:validation:Pattern=`^\w+([-\/]\w+)*(-\d+\.\d+)?$`
```
krancour, line 127: I'm supportive of relaxing this, but I don't think it's relaxed _enough_.

There are a lot of valid branch names containing a `.` that this pattern will not match.

`foo.bar` should be acceptable, for instance.

Also, `/` needs to be escaped in PCRE, but not in Go.

I suggest this:

```suggestion
	// +kubebuilder:validation:Pattern=`^\w+([-/\.]\w+)*$`
```

https://regex101.com/r/J7GXoJ/1

Codegen will need to be re-run.

tal-hason: I will extend the regex to be more relaxed 😎

krancour: Thanks @tal-hason! Appreciate this!

tal-hason:
> Thank you so much @tal-hason!

Happy to 😎

Issue #2581, opened 2024-09-26
bug(ui): selecting freight in assembly resets the view


Marvin9:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

When we select freight in freight assembly, page number in the table resets to 1.

# Screenshots

https://github.com/user-attachments/assets/51c85993-e3df-4893-9e2c-d4cc830c32dc

# Steps to Reproduce

- Go to create freight page (You can click on "Assemble Freight" at top right corner in freight timeline), Make sure to select warehouse which has > 10 freights.
- Go to page 2
- Click on any freight
- Page is reset to 1. It destroys the context. The freight is selected however

# Version

```
dev
```

# Logs

```
Paste any relevant application logs here.
```

aayushsrivastava:
@krancour @Marvin9 I am looking for a good first issue to get my feet wet. Can I work on this?

Marvin9:
Sure @aayushsrivastava, assigned to you.

PR #3246, opened 2025-01-11
fix(ui): freight assembly tables keep resetting to page 1

aayushsrivastava:
fixes: https://github.com/akuity/kargo/issues/2581

**RCA**:
1. The ImageTable/CommitTable/ChartTable component gets re-mounted (unmounted and mounted again) when the _select_ function is called. This leads to the loss of any local state in these components (including the active page).
2. The _select_ function modifies a state in the AssembleFreight component.
3. The DiscoveryTable component (which wraps the ImageTable/CommitTable/ChartTable component) is a nested function inside AssembleFreight. Every time AssembleFreight rerenders due to a change in its state, the DiscoveryTable function is redeclared, changing its memory reference. React thinks the component has been replaced, unmounts the old Table, and mounts a new one.

This problem is also described here: https://stackoverflow.com/a/64211013

### DIFF
--- a/ui/src/features/assemble-freight/assemble-freight.tsx
+++ b/ui/src/features/assemble-freight/assemble-freight.tsx
@@ -252,14 +252,18 @@ export const AssembleFreight = ({
   );
 };
 
-const DiscoveryTable = ({ selected, chosenItems, select }: {
-  selected?: DiscoveryResult,
+const DiscoveryTable = ({
+  selected,
+  chosenItems,
+  select
+}: {
+  selected?: DiscoveryResult;
   chosenItems: {
     [key: string]: {
       artifact: DiscoveryResult;
       info: FreightInfo;
     };
-  },
+  };
   select: (item?: FreightInfo) => void;
 }) => {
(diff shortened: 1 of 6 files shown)


aayushsrivastava:
After fixing the original bug, I noticed that the page still resets to 1 when you change the type of artifact. The second commit fixes that issue by calculating and setting the expected page when the table first mounts.

aayushsrivastava:
https://github.com/user-attachments/assets/2575f16f-baea-474c-bc9d-1ef5816e8e13

aayushsrivastava:
This PR is ready for review. FYI: @krancour

Marvin9:
Hello @aayushsrivastava sorry for the delay. This is in good shape. I wanted to append some commits for alternative approach than to calculate page by selected contents, but for some reason couldn't push to your fork's main after adding commits. So I have raised a new PR https://github.com/akuity/kargo/pull/4120 that includes your commits as well. Let me know if you have any questions. And thanks a lot for this fix

PR #3257, opened 2025-01-13
feat: add ability to control semantics for Freight being made available to a Stage

### LINKED ISSUE #3207: feat: opt-in to "all" semantics when determining freight's availability to a given stage
# Checklist

# Note
I'm not actually sure if this is a feature request or a bug - the current functionality caught me by surprise though I could be biased based on my personal use-case.

I've searched the docs, issues, discussions and Discord and I couldn't find any mention of this behavior.  I've also read through parts of the code-base to determine if this feature request is supported but not documented.  I've found reference to the ControlFlow concept, but it doesn't support any extra customization at present.  

Please let me know if this use-case is solvable using existing constructs.

# Proposed Feature
(Optionally?) Freight should only be promoted to a downstream Stage when **_all_** upstream Stages have successfully verified it.

# Motivation
In our environment we have a large number of Stages (each represents a different cluster/environment).  Stages are grouped based on risk profile (think: dev, staging, prod) where we would like each Stage in a 'group' to receive a Freight in parallel. Each Stage then runs their own verification process (potentially different / more exhaustive per-Stage).  Each Stage also has `autoPromotionEnabled` - we are relying on the verifications to ensure a safe promotion.

If a single Stage fails to verify a given Freight, it indicates something needs to be investigated and possibly fixed and as a result, the Freight should not be promoted further as it could fail in more critical environments.

In a simple Kargo Pipeline with no parallel steps, the above behavior works as expected: eg:

```
Warehouse-A -> Stage-Dev -> Stage-Staging -> Stage-Prod
```

In pipelines that have 'parallel' Stages, Freight will be promoted to a downstream Stage if any _single_ upstream Stage has verified it:

```

                            - Stage-StagingA -
Warehouse-A -> Stage-Dev  -|                   |- Stage-Prod
                            - Stage-StagingB -
```
If Freight `aaaaaa` passed verification in `Stage-StagingA` but fails in `Stage-StagingB`, `Stage-Prod` will receive the Freight.

# Suggested Implementation
I am struggling to think of use-cases where the current functionality is the preferred behavior which is making me lean towards this being a bug rather than a feature.  If its a bug, I have a small working POC that changes the [getAvailableFreight](https://github.com/akuity/kargo/blob/bb95eb4b6753fdc9512db97b9281ba16b63d3bac/internal/controller/stages/control_flow_stages.go#L317) method in a ControlFlow Stage to only return Freight verified by all upstream Stages.  Assumably, the same logic could be applied to a regular Stage.

If this is a feature, then I imagine the same logic would be suitable but would need to be configurable - ideally per-Stage.  Though maybe this is only a 'bug' when `autoPromotionEnabled` is set on a Stage and this behavior is otherwise expected?

Let me know if opening a draft PR helps to understand the problem better.

aidan-canva:
Closes #3207 

Adds concept of 'availability strategy' to a Freight Request to allow users to control the semantics used to determine if (verified) Freight is made available to a stage.

Default behavior is to make a Freight available if 'any' of the upstream Stages have verified it. This is backwards compatible with the current state of behavior.  Users can explicitly configure `UpstreamAny` (default behavior) or `UpstreamAll` (enums) as possible values.  `UpstreamAll` will only make Freight available if all of the configure upstream Stages have verified it.  An enum was chosen over a boolean as it is possible other option could be implemented in the future (ie, promote when a majority of upstream have verified the Freight).

This condition is applied on-top of existing Freight Request controls, such as the recently added minimum bake time control.  For that reason, perhaps its better suited to sit closer to `FreightSources` where the `RequiredSoakTime` is configured.


aidan-canva:
FYI @krancour

krancour:
@aidan-canva sorry for the delay. At a glance, this looks fantastic. I'll dig deeper on Monday and maybe have @hiddeco look as well.

api/v1alpha1/stage_helpers.go
```diff
@@ -135,8 +135,9 @@ func (s *Stage) ListAvailableFreight(
 		var listOpts *ListWarehouseFreightOptions
 		if !req.Sources.Direct {
 			listOpts = &ListWarehouseFreightOptions{
-				ApprovedFor: s.Name,
-				VerifiedIn:  req.Sources.Stages,
+				ApprovedFor:          s.Name,
+				VerifiedIn:           req.Sources.Stages,
+				RequireAllVerifiedIn: req.AvailabilityStrategy == FreightAvailabilityStrategyAllUpstream,
```
hiddeco, line 140: Is there a specific reason you made the `RequireAllVerifiedIn` a boolean instead of just passing on the strategy?

api/v1alpha1/stage_types.go
```diff
@@ -128,6 +128,14 @@ type FreightOriginKind string
 
 const FreightOriginKindWarehouse FreightOriginKind = "Warehouse"
 
+// +kubebuilder:validation:Enum={AllUpstream,AnyUpstream}
+type FreightAvailabilityStrategy string
+
+const (
+	FreightAvailabilityStrategyAllUpstream FreightAvailabilityStrategy = "AllUpstream"
+	FreightAvailabilityStrategyAnyUpstream FreightAvailabilityStrategy = "AnyUpstream"
+)
```
hiddeco, line 137: Wonder if brevity might be an improvement here, but open to discussion.

```suggestion
const (
	FreightAvailabilityStrategyAll FreightAvailabilityStrategy = "All"
	FreightAvailabilityStrategyOneOf FreightAvailabilityStrategy = "OneOf"
)
```

hiddeco (commented):
> For that reason, perhaps its better suited to sit closer to `FreightSources` where the `RequiredSoakTime` is configured.

After reviewing your code, I think this is a good observation, and I would feel more comfortable with it sitting at the same level as `RequiredSoakTime`. This would make the setting count for the list item (just like the soak time requirement), which I feel is better.

Other than this and my question/suggestion below, this looks like a straight forward implementation and a job well done. 🌷

api/v1alpha1/stage_types.go
aidan-canva: The explicit use of 'Upstream' in these enums was chosen in case there was a future availability strategy that wasn't referring to upstream Stage state, but practically those use cases will probably need to be implemented as specific values like `RequiredSoakTime` as it will have some kind of configurable 'value' (ie, the soak time).

Maybe you have more context to guide whether that can be a future concern (if at all).  I've taken your suggestion though and applied it to the PR 👍

api/v1alpha1/stage_helpers.go
aidan-canva: Happy to change this over - this is how I originally had it implemented. 

I was probably thinking too much into the (unknown) future values of this enum, but it felt a bit odd adding tight coupling of the `AvailabilityStrategy` into the `ListFreight` method.  With the `bool` it makes it quite clear which states of the value are being handled as `FreightAvailabilityStrategyAllUpstream` is the only value that requires specific consideration (in this implementation). 

I also had the impression that although a public method, making breaking changes to the `ListWarehouseFreightOptions` interface in the future wouldn't be an issue 😄 maybe that's a wrong assumption and is a good reason to just plumb in the raw `AvailabilityStrategy` value now.

Anyway, happy to take your guidance on this 😄

api/v1alpha1/stage_types.go
hiddeco: I have a hard time imagining what enums we would introduce that would not directly relate to stages, at this point. The only possible thing I could think of is to move the `direct` field to an enum selector, but I fear that ship has already sailed.

This makes me wonder however if `FreightAvailabilityStrategy` is the right name for the field/type, or if it should hint towards something like `StageSelectionStrategy`. But I think @krancour is a better judge here, as English isn't my primary language.

api/v1alpha1/freight_helpers.go
```diff
@@ -129,6 +129,19 @@ func (f *Freight) IsVerifiedIn(stage string) bool {
 	return verified
 }
 
+// IsVerifiedInAll returns whether the Freight has been verified in all
+// Stages.
+func (f *Freight) IsVerifiedInAll(stages []string) bool {
+	// NB: This method exists for convenience. It doesn't require the caller to
+	// know anything about the Freight status' internal data structure.
```
krancour, line 136: Nit: We could lose this comment. I know `IsVerifiedIn()` has a comment like this, but it was there to justify its existence when `freight.Status.VerifiedIn[stage]` seems no harder than `freight.IsVerifiedIn(stage)`. `IsVerifiedInAll()` is doing a more obvious bit of work, so I don't think it needs any justification.

api/v1alpha1/stage_types.go
```diff
@@ -128,6 +128,14 @@ type FreightOriginKind string
 
 const FreightOriginKindWarehouse FreightOriginKind = "Warehouse"
 
+// +kubebuilder:validation:Enum={AllUpstream,AnyUpstream}
+type FreightAvailabilityStrategy string
+
+const (
+	FreightAvailabilityStrategyAll   FreightAvailabilityStrategy = "All"
+	FreightAvailabilityStrategyOneOf FreightAvailabilityStrategy = "OneOf"
+)
```
krancour, line 137: There seems to be a discrepancy between the kubebuilder marker and the values of the consts.

api/v1alpha1/stage_types.go
```diff
@@ -264,6 +272,12 @@ type FreightSources struct {
 	// +kubebuilder:validation:Type=string
 	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(s|m|h))+$"
 	RequiredSoakTime *metav1.Duration `json:"requiredSoakTime,omitempty" protobuf:"bytes,3,opt,name=requiredSoakTime"`
+	// AvailabilityStrategy specifies the semantics for how requested Freight is
+	// made available to the Stage. This field is optional. When left unspecified,
+	// the field is implicitly treated as if its value were "AnyUpstream".
+	//
+	// +kubebuilder:default=AnyUpstream
+	AvailabilityStrategy FreightAvailabilityStrategy `json:"availabilityStrategy" protobuf:"bytes,4,opt,name=availabilityStrategy"`
```
krancour, line 280: See https://github.com/akuity/kargo/pull/3257/files#r1939758923

api/v1alpha1/warehouse_helpers.go
```diff
@@ -149,6 +155,32 @@ func (w *Warehouse) ListFreight(
...
+				}
+			}
+			if f.IsVerifiedInAll(opts.VerifiedIn) {
+				filtered = append(filtered, f)
+				continue
+			}
+		}
+
+		freight = filtered
+	}
```
krancour, line 182: Wondering if we can lose this loop entirely by modifying the and'ing relevant field selectors before the queries.

api/v1alpha1/warehouse_helpers.go
```diff
@@ -164,22 +196,42 @@ func (w *Warehouse) ListFreight(
...
+		// Otherwise, include Freight if it has passed the soak time in a single
+		// Stage.
+		if opts.RequireAllVerifiedIn {
+			// If Freight is verified in ALL upstream Stages, then it is
+			// available.
+			if verifiedStages.Equal(sets.New(opts.VerifiedIn...)) {
+				filtered = append(filtered, f)
+			}
+			continue
+		}
```
krancour, line 229: Weren't these already filtered out?

krancour:
Few comments, but mostly looks very good.

api/v1alpha1/warehouse_helpers.go
aidan-canva: Not quite, this second loop/availability check is only run if the `opts.VerifiedBefore != nil` is set and is verifying that **all** of the (verified, at this point) Freight has also passed its minimum `opts.VerifiedBefore` time.  `fake-freight-4` test case in the `warehouse_helper_tests.go` should cover this logic 👍

api/v1alpha1/freight_helpers.go
```diff
@@ -129,6 +129,17 @@ func (f *Freight) IsVerifiedIn(stage string) bool {
 	return verified
 }
 
+// IsVerifiedInAll returns whether the Freight has been verified in all
+// Stages.
+func (f *Freight) IsVerifiedInAll(stages []string) bool {
```
aidan-canva, line 134: Potentially not longer required after the refactor to do the logical ANDing of verified freight to the actual k8s query.

api/v1alpha1/warehouse_helpers.go
aidan-canva: Good call!  I've just done a quick refactor to implement this and tests look to be passing as expected.  It also looks much cleaner with the shift to passing int the raw `AvailabilityStrategy` as it makes it clear which semantics are being handled (or not).

I'll do a bit of manual testing just to verify some things, but if this looks good to you I'll proceed with removing the `IsVerifiedInAll` method and tests (its no longer required) and getting the rest of the CI steps green.

aidan-canva:
Thanks for the feedback @krancour 👍 I've addressed all of the comments and its ready for another review.

api/v1alpha1/warehouse_helpers.go
krancour: Ah. I missed that. Indeed, this this is necessary for checking the soak times. 👍

aidan-canva:
@krancour @hiddeco just a friendly probe - any blockers on getting this merged?  I'll get the last DCO commit signed/commits squashed if I can get the go ahead.

hiddeco (approved):
Thank you very much @aidan-canva 🙇

aidan-canva:
Thanks for the feedback and collaboration!

PR #3275, opened 2025-01-14
docs: split up steps into separate reference pages

hiddeco:
Replacement for #3272 targeting `newdocs`.

This splits the single page with step documentation into multiple pages, accompanied by an auto-generated index.

As a follow-up, the previous "Promotion Step Reference" document will be turned into a broader "Promotion Template Reference" document and paired with a dedicated "Promotion Task Reference" page.

All pages have the same structure and headings, e.g. "Configuration", "Output" (where applicable), and "Examples". This is to ease navigation for the user.


krancour:
I updated the relative URLs to exclude numbers. In some older configuration (it might have been with Docusaurus 2), we seemed to need them, but we no longer do.

docs/docs/60-new-docs/60-user-guide/60-reference-docs/30-promotion-steps/10-git-clone.md
```diff
@@ -0,0 +1,100 @@
...
+| `checkout[].commit` | `string` | N | A specific commit to check out. Mutually exclusive with `branch`, `tag`, and `fromFreight=true`. If none of these is specified, the default branch will be checked out. |
+| `checkout[].tag` | `string` | N | A tag to check out. Mutually exclusive with `branch`, `commit`, and `fromFreight=true`. If none of these is specified, the default branch will be checked out. |
+| `checkout[].fromFreight` | `boolean` | N | Whether a commit to check out should be obtained from the Freight being promoted. A value of `true` is mutually exclusive with `branch`, `commit`, and `tag`. If none of these is specified, the default branch will be checked out. Default is `false`, but is often set to `true`. <br/><br/>__Deprecated: Use `commit` with an expression instead. Will be removed in v1.3.0.__ |
+| `checkout[].fromOrigin` | `object` | N | See [specifying origins](#specifying-origins). <br/><br/>__Deprecated: Use `commit` with an expression instead. Will be removed in v1.3.0.__ |
```
krancour, line 33: All `#specifying-origins` are broken... but (I've just realized) they are in the existing docs as well. It seems that section was prematurely removed in the 1.1 docs. Need to go back to 1.1 to find it: https://release-1-0.docs.kargo.io/references/promotion-steps#specifying-origins

We could fix this _or_ especially since it is only deprecated fields that reference this, we could leave it be or remove these links. I don't have any strong opinion.

krancour: This otherwise LGTM, so I'm approving and I am content with however you decide to deal with this (or not).

krancour (approved):

hiddeco:
@krancour not sure about the relative URLs being an improvement, for a couple of reasons:

1. When you rename a page (or file), you must find all relative "magic" paths that once matched the previous name. This is a step backward for people with a more intelligent IDE (like any IntelliJ one), which knows how to find the previous file reference and replace it with your new one.
2. It is no longer possible to navigate the links through the source, on both GitHub and in my IDE.

If the primary concern is, "whenever I reorder files, I do not want to update references", I suggest we follow the actual advice from Docusaurus (see their warning [here](https://docusaurus.io/docs/sidebar/autogenerated#using-number-prefixes)) and move the ordering to the front matter (as documented [here](https://docusaurus.io/docs/sidebar/autogenerated#autogenerated-sidebar-metadata)).

krancour:
@hiddeco valid points. Let's revert my amendment.

(I'll also adjust #3281)

As to ordering through front matter, I'm less sure about that because it makes it much harder to understand the order just by looking at the file tree, which I think is beneficial.

docs/docs/60-new-docs/60-user-guide/60-reference-docs/30-promotion-steps/10-git-clone.md
hiddeco: Had noticed this as well. Will address this in a follow up PR at some point.

hiddeco:
> As to ordering through front matter, I'm less sure about that because it makes it much harder to understand the order just by looking at the file tree, which I think is beneficial.

Yes, that was also the downside I had observed. Given this, let's just stick with numbering in the file system for now, and see how that goes.

PR #3281, opened 2025-01-15
docs: add faqs

krancour:
cc @hiddeco @jessesuen @kencochrane 

Please feel free to suggest additional questions, challenge the importance of an existing question, edit wording, etc.

Also feel free to amend the PR directly if you please. I do not mind.


kencochrane (approved):

docs/docs/60-new-docs/30-faqs.md
```diff
@@ -4,6 +4,179 @@ sidebar_label: FAQs
...
+### _So “stage” is just another word for “environment?”_
+
+Not exactly, but you could think of it that way if you like.
+
+"Environment" can be an overloaded term to begin with. We didn't want to make
+things worse. Beyond that, we find more flexibility in the term "stage." _Stage
+can mean whatever you need it to._ It could be a cluster. It could be a
+particular instance of your entire application. It could be one of just several
+microservices comprising your application. Or it could be a group of related
+microservices.
```
hiddeco, line 49: I feel like this is missing a "punch".

I.e. it could bring forward that a "stage" is essentially just a promotion target for a set of artifacts ("freight"), and that it's up to the user to assess what this entails for their specific promotion pipeline.

I also feel it maybe should touch into a "stage" being a (hierarchical) "station" along a pipeline.

hiddeco (approved):
One minor comment, but absolutely not blocking.

krancour:
The question we get a lot that I've overlooked is "Do I need to use environment-specific branches?"

That pattern is offputting to many people for completely spurious reasons, so I think it's important to establish up front that _no_ you don't need to do that, and at the same time plant the seed that it's a pattern that shouldn't be so quickly dismissed.

Issue #3282, opened 2025-01-15
Leading \ character in Quickstart promotionTemplates causes error during artifact promotion


dylanfeehan:
# Checklist

* [ x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `kargo version`.
* [x ] I've pasted logs, if applicable.

# Description

In the Quickstart section of Kargo's documentation, the promotionTemplate expressions contain an accidental leading "\" character which causes error during artifact promotion.

# Screenshots

### Using the Kubernetes CLI to apply the manifest which is copied directly from the Quickstart page
![Screenshot 2025-01-14 at 9 15 55 PM](https://github.com/user-attachments/assets/0dccaefb-1e9d-4b8b-bd02-61693af65502)
### Viewing the live manifests in the Kargo UI we can see the leading "\" characters in front of the expressions
![Screenshot 2025-01-14 at 9 16 36 PM](https://github.com/user-attachments/assets/e4e83deb-dfb2-426f-b89a-fcd24455c2ed)
### Promotion error
![Screenshot 2025-01-14 at 9 16 51 PM](https://github.com/user-attachments/assets/c35a2860-8a16-40ae-af3c-b4e8eac6ea66)
### Note "fatal: protocol '\https' is not supported, since the expression resolves to \https://github.com...
![Screenshot 2025-01-14 at 9 16 59 PM](https://github.com/user-attachments/assets/a8693a87-decd-42f1-9063-846c5038457d)
### Fixing only for the first step, git-clone, and showing diff
![Screenshot 2025-01-14 at 9 19 19 PM](https://github.com/user-attachments/assets/5d1f3c6b-0041-4f8e-90e5-ee349f236edc)
### Only the step which I fixed did not error
![Screenshot 2025-01-14 at 9 19 47 PM](https://github.com/user-attachments/assets/10f6c117-135b-4d53-9a1e-26e29b40f025)
### Since I did not update the second step's expressions, it failed due to the leading "\" which screwed up the directory name
![Screenshot 2025-01-14 at 9 20 07 PM](https://github.com/user-attachments/assets/51dca164-b9b4-4858-9b6e-67757092c068)

# Steps to Reproduce

Follow the Quickstart documentation

# Version

n/a

# Logs

n/a

dylanfeehan:
Link to PR: https://github.com/akuity/kargo/pull/3283

krancour:
[...]

PR #3283, opened 2025-01-15
docs: remove leading \ character from all quickstart promotionTemplate expressions

dylanfeehan:
Link to detailed issue: https://github.com/akuity/kargo/issues/3282

### DIFF
--- a/docs/docs/20-quickstart.md
+++ b/docs/docs/20-quickstart.md
@@ -336,43 +336,43 @@ the previous section.
           steps:
           - uses: git-clone
             config:
-              repoURL: \${{ vars.gitopsRepo }}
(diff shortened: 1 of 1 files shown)


dylanfeehan:
Sorry for all the force pushes, I accidentally authored with a GitHub account I use for work and needed to update the Author and Signature

krancour:
This PR _breaks_ the Quickstart. The command for defining the project appears as follows as it was meant to be copied/pasted into a shell as-is:

```console
cat <<EOF | kubectl apply -f -

... manifests here ...

EOF
```

The `\` is necessary escaping to prevent the shell from interpreting the `$` character as it normally would.

PR #3292, opened 2025-01-15
fix: responsive versioned dropdown

fykaa:
added version width and title
image

---
fixed dropdown component
before:
image
after:
image

### DIFF
--- a/docs/src/components/VersionDropdown.css
+++ b/docs/src/components/VersionDropdown.css
@@ -9,10 +9,17 @@
     cursor: pointer;
     padding: 5px 10px;
     font-size: 1rem;
-    font-family: 'Monospace', 'Roboto', 'Open Sans', 'Lato', 'Courier New', monospace;
     font-weight: bold;
+    width: 73px;
+    text-align: center;
 }
 
 .version_dropdown option {
     font-weight: normal; /* Keeps options normal weight */
+}
+
+@media (max-width: 996px) {
+    .navbar__items--right .version_dropdown {
+        display: none;
+    }
 }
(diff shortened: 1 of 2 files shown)


hiddeco:
I wonder if it would not be more useful to display a version for the "latest" version as well, or to at least mention _which_ version is the latest, i.e. `Latest version (v1.2.0)`. The advantage of this is that if someone knows they are on Kargo v1.1.0, they can then automatically tell they are _not_ on latest.

krancour:
> I wonder if it would not be more useful to display a version for the "latest" version as well, or to at least mention which version is the latest, i.e. Latest version (v1.2.0). The advantage of this is that if someone knows they are on Kargo v1.1.0, they can then automatically tell they are not on latest.

Well... this is a bit complicated because I'm not sure we have any reasonable way to infer what branch docs.kargo.io currently points at. If we tried do something like take a guess that it points at the semantically greatest release-x.y branch, we'd end up being wrong anytime a release branch exists for an _upcoming_ release. Two days ago, such a heuristic would have shown 1.1 docs and told people they were looking at 1.2.

But I agree that it would be nice as long as we had a reasonable way to approach it. I think we should put a pin in this and come back to it when we have the luxury of time. I wouldn't want it to block the modest styling improvements in this PR.

krancour:
@fykaa this still doesn't look quite right on mobile...

IMG_2997

hiddeco:
Should the version in the mobile sidebar be either at the top or the very bottom of the menu?

krancour:
I need to re-review the difference from a small viewport. I will do so as soon as possible.

krancour (approved):

PR #3295, opened 2025-01-16
Adding entrypoint

JoshuaCooper:
Removing busybox  and adding entrypoint. This makes helm as your entrypoint so the image can run with an input of a helm chart.

```
➜  kargo git:(main) ✗ docker run kargo:latest-amd64    
The Kubernetes package manager

Common actions for Helm:

- helm search:    search for charts
- helm pull:      download a chart to your local directory to view
- helm install:   upload the chart to Kubernetes
- helm list:      list releases of charts
```

### DIFF
--- a/kargo-base.apko.yaml
+++ b/kargo-base.apko.yaml
@@ -4,7 +4,6 @@ contents:
   repositories:
   - https://packages.wolfi.dev/os
   packages:
-  - busybox # TODO: Remove this if we can figure out how to get by without a shell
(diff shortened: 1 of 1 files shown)


hiddeco:
Can you elaborate on why this would be an improvement? As I have a hard time reasoning why someone would want to use the Kargo image to run `helm`.

krancour:
This PR makes no sense at all.

PR #3299, opened 2025-01-17
feat: Mount path /home/nonroot as emptyDir volume 

### LINKED ISSUE #3298: Mount path /home/nonroot as emptyDir volume
# Checklist

# Proposed Feature

This is a follow-up to #3150. To execute the promotion step `helm-update-chart` with `securityContext` `readOnlyRootFilesystem` enabled we also need to mount the path `/home/nonroot` as tmpDir volume.

# Motivation

When executing the promotion step `helm-update-chart` we get the following error message:

```
step execution failed: step 1 met error threshold of 1: failed to run step "helm-update-chart": failed to update chart dependencies: could not find : chart argo-cd not found in https://artifactory.local/artifactory/api/helm/helm-myorg-virtual: looks like "[https://artifactory.local//artifactory/api/helm/helm-myorg-virtual"](https://artifactory.local/artifactory/api/helm/helm-myorg-virtual%22) is not a valid chart repository or cannot be reached: open /home/nonroot/.cache/helm/repository/IsciyC9Yvs8xutdO0YDeHD1EAgo=-index.yaml: no such file or directory
```

This error occurs because Helm cannot pull the chart to the read-only path `/home/nonroot`.

# Suggested Implementation

Implementation analogous to #3150.

ggogel:
Closes #3298.

I deployed the chart to our test environment and successfully tested the promotion step `helm-update-chart`.

### DIFF
--- a/charts/kargo/templates/controller/deployment.yaml
+++ b/charts/kargo/templates/controller/deployment.yaml
@@ -82,6 +82,8 @@ spec:
         volumeMounts:
         - mountPath: /tmp
           name: tmp-data
+        - mountPath: /home/nonroot
+          name: home-nonroot-data
         {{- if or .Values.kubeconfigSecrets.kargo .Values.kubeconfigSecrets.argocd }}
         - mountPath: /etc/kargo/kubeconfigs
           name: kubeconfigs
@@ -130,6 +132,8 @@ spec:
       {{- end }}
       volumes:
       - name: tmp-data
+        emptyDir: {}
+      - name: home-nonroot-data
         emptyDir: {}
       {{- if or .Values.kubeconfigSecrets.kargo .Values.kubeconfigSecrets.argocd }}
       - name: kubeconfigs


krancour:
@ggogel I see what you're trying to do, but I think this might point to a bug that needs to be resolved differently than this.

Everything that happens in the course of a promotion process is supposed to happen in a promotion-specific workspace under /tmp (as you know from #3298).

If `/home/nonroot/` is being touched for any reason, that strikes me as a bug that needs fixing... and then the problem you're encountering goes away.

cc @hiddeco

ggogel:
@krancour Thanks for pointing that out. We also suspected this might be related to a bug, but we weren’t entirely sure.

After further investigation, it seems the issue is caused by the `helm repo add` command. When executing this inside the container it defaults to the path `/home/nonroot`.

krancour:
It may be best for @hiddeco to sort this out, as he's most familiar with this bit of code. tbh, I don't know off the top of my head why we would even be doing a `helm repo add` as I don't _think_ it's really required for what that step is doing, but he will know better than I.

Issue #3290, opened 2025-01-15
Helm repo config & cache directory environment variables aren't respected during helm-update-chart


WZHGAHO:
# Description

During helm-update-chart promotion step, I received the following error:

```shell
step execution failed: step 1 met error threshold of 1: failed to run
    step "helm-update-chart": failed to update chart dependencies: could not
    find : chart rsp-pulsar-function-example not found in
    https://artifacts.example.com/artifactory/api/helm/rsp-helm-host: looks like
    "https://artifacts.example.com/artifactory/api/helm/rsp-helm-host" is not a
    valid chart repository or cannot be reached: open
    /.cache/helm/repository/ahF2zqceBHME9GAXefZb1vqyn5M=-index.yaml: no such
    file or directory'
``` 

I tried to reproduce by exec'ing into kargo-controller pod, and trying to add Helm repo:

```shell
~ $ helm repo add gabor-test https://artifacts.example.com/artifactory/api/helm/rsp-helm-host
Error: mkdir /.config: permission denied
``` 

It seems if repository config directory is not set, then Kargo takes Helm's default, which is $HOME/.config/helm
Since in our case $HOME is set to /, it tried to create directory /.config. 
Kargo runs with the following user, and it has no permission to write to /

```shell
~ $ ps auxf
PID   USER     TIME  COMMAND
    1 10010700  0:00 /sbin/tini -- /usr/local/bin/kargo controller
    7 10010700  0:09 /usr/local/bin/kargo controller
   28 10010700  0:00 sh
45444 10010700  0:00 ps auxf
~ $ id
uid=1001070000(1001070000) gid=0(root) groups=0(root),1001070000
``` 

Then I tried to set repository config directory to a directory where the user has write permission, and then I could reproduce the problem:

```shell
~ $ helm repo add gabor-test https://artifacts.example.com/artifactory/api/helm/rsp-helm-host --repository-config /tmp/foo
Error: looks like "https://artifacts.example.com/artifactory/api/helm/rsp-helm-host" is not a valid chart repository or cannot be reached: open /.cache/helm/repository/gabor-test-index.yaml: no such file or directory
``` 

When I set also the repository cache directory for Helm, repo add worked:

```shell
~ $ helm repo add gabor-test https://artifacts.example.com/artifactory/api/helm/rsp-helm-host --repository-config /tmp/foo --repository-cache /tmp/bar
"gabor-test" has been added to your repositories
``` 

[...]

PR #3326, opened 2025-01-20
fix(directives): HTTP repo cache `helm-chart-update`

hiddeco:
Fixes: #3299
Fixes: #3298 
Fixes: #3290

---

The reason `/home/nonroot` is being touched, is that Helm has decided to do something odd within their download manager by not taking the configured value into account for the repository cache while trying to resolve a chart dependency for a repository that does not have a cache entry for the `index.yaml`.

Instead, it temporarily caches the index to an `index.yaml` with a random 20-byte name to the [environment variable-based cache path](https://github.com/helm/helm/blob/v3.17.0/pkg/helmpath/lazypath.go#L60), and then removes it as soon as it has resolved the chart URL for the version it is interested in.

Stack trace:

1. https://github.com/helm/helm/blob/v3.17.0/pkg/downloader/manager.go#L315
2. https://github.com/helm/helm/blob/v3.17.0/pkg/downloader/manager.go#L745-L748
3. https://github.com/helm/helm/blob/v3.17.0/pkg/repo/chartrepo.go#L243-L254
4. https://github.com/helm/helm/blob/v3.17.0/pkg/repo/chartrepo.go#L78

---

To solve this issue, we now add all the "classic" repositories (even those that do not require authentication) to our isolated cache and fetch their indexes before running a Helm chart update, which prevents Helm from attempting to download them.

### DIFF
--- a/internal/directives/helm_chart_updater.go
+++ b/internal/directives/helm_chart_updater.go
@@ -2,9 +2,11 @@ package directives
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"io"
+	"net/url"
 	"os"
 	"path"
 	"path/filepath"
@@ -211,7 +213,7 @@ func (h *helmChartUpdater) updateDependencies(
 		}
 	}
 
-	if err = h.loadDependencyCredentials(
+	if err = h.setupDependencyRepositories(
 		ctx,
 		stepCtx.CredentialsDB,
 		registryClient,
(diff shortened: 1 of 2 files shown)


internal/directives/helm_chart_updater.go
```diff
@@ -605,3 +645,27 @@ func backupFile(src, dst string) (err error) {
 	}
 	return nil
 }
+
+// nameForRepositoryURL generates an SHA-256 hash of the repository URL to se
```
krancour, line 649:
```suggestion
// nameForRepositoryURL generates an SHA-256 hash of the repository URL to use
```

Issue #3240, opened 2025-01-09
Parse (grep/yq/jq/sed/awk) step to extract file contents


jessesuen:
# Proposed Feature

It has been requested to extract a portion of a file's contents after cloning, so that the file's contents can be used in a subsequent step. e.g. 

```yaml
- uses: git-clone
  config:
    repoURL: ${{ vars.gitRepo }}
    checkout:
    - commit: ${{ commitFrom(vars.gitRepo, warehouse("base")).ID }}
      path: ./src
- uses: parse-file
  as: app_versions
  config:
    paths:
    - path: config/app1/version.txt
      sed: # some expression to extract output from the file
      output: app1
    - path: config/app2/version.yaml
      yq: ".app2.version"
      output: app2
- uses: yaml-update
  config:
    path: ./src/charts/my-chart/values.yaml
    updates:
    - key: app1.image.tag
      value: ${{ outputs.app_versions.app1 }}
    - key: app2.image.tag
      value: ${{ outputs.app_versions.app2 }}
```

# Motivation

This would allow Kargo to use git repository monitoring feature in a more powerful way. e.g. A user could configure Kargo to watch a single file in for changes, similar to how we watch a OCI/Helm/Git repo for changes, but then act on specific changes to that file, possibly even in a different repo.

We already have a copy step which is close to this and consider this a variation of the`copy` step.

# Suggested Implementation

Implementation will probably be similar to our [`copy`](https://docs.kargo.io/references/promotion-steps/#copy) step, which deals with files on disk.

The big question would be how to extract information from a file. But should be as good as:
* jq
* yq
* grep
* sed
* awk

jq/yq would be great for structured data. but unstructured data may need grep/sed/awk-like syntax.

krancour:
For the http step, we allowed expr-lang expressions to be used to extract stuff from the structured response body.

I'm not 100% against exec'ing out to jq/yq, but reusing the approach the http step used seems like a nice starting point because it introduces no new binaries to the image and has the added benefit of being consistent with something we've already done.

Unstructured data is a different story...

Is there a specific use case for unstructured data being used in a promotion process?

It might be worth tackling structured first and revisiting unstructured later.

jessesuen:
Great suggestion. I forgot that expr already has a very powerful way to extract, and we have precedent with http. I  like the idea of using that over `yq` or `jq` syntax, so we can make a call of expr as the method for extracting structured data.

[...]

PR #3327, opened 2025-01-21
feat: json-parse promotion step

fykaa:
closes https://github.com/akuity/kargo/issues/3240

`parse-file` step allows users to specify file paths and extract specific values using expressions. The `fromExpression` field accepts direct expressions without `${{}}` wrappers, as evaluation happens within the step logic itself.

eg.
```
steps:
  - uses: json-parse
    config:
      path: "path/to/file1.json"
      outputs:
        - fromExpression: "object.version"
          name: "version1"
        - fromExpression: "object.name"
          name: "name1"
```

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/json-parse.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/json-parse.md
@@ -43,7 +43,7 @@ After cloning the repository and clearing the output directory, the `json-parse`
 to extract the image tag from the `Freight` being promoted.
 Using dot notation (image.tag), it extracts the nested value from the JSON file.
 
-```json
+```yaml
 vars:
 - name: gitRepo
   value: https://github.com/example/repo.git
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/json-parse.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/json-parse.md
@@ -1,0 +1,90 @@
+---
+sidebar_label: json-parse
+description: Parses a JSON string and extracts values based on specified expressions.
+---
+
+# `json-parse`
+
+`json-parse` is a utility step that parses a JSON string and extracts values using [expr-lang][] expressions.
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `path` | `string` | Y | Path to a JSON file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `outputs` | `[]object` | Y | A list of rules for extracting values from the parsed JSON. |
+| `outputs[].name` | `string` | Y | The name of the output variable. |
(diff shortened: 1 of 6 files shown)


krancour:
Sorry for the delay in the latest round of reviews here.

The schema is looking pretty good. 👍 

I appreciate us reaching consensus on that before starting on the implementation.

There's another detail that's just occurred to me, and I'd like to invite @hiddeco and @jessesuen to weight in.

Had we been thinking of a single `parse-file` step that can handle both JSON and YAML? Or had we been thinking of two similar, but separate steps? thb, I'd not even considered this question until now.

I think I may have a slight preference for separate `parse-yaml` and `parse-json` steps, with my rationale being: On a technical level, it seems easy enough to combine these, but if we add support for parsing _other_ specific types of files in the future, their general shape might not conform to this schema that happens to work well for both JSON and YAML. At that point, we might regret having named a step so generically as `parse-file`.

I wouldn't die on this hill.

krancour:
> Had we been thinking of a single parse-file step that can handle both JSON and YAML? Or had we been thinking of two similar, but separate steps?

@jessesuen and I discussed this offline and agreed on separate steps (which can share a lot of common code underneath).

They can even share the schema.

This plan gives us the flexibility to add file parsers for additional formats in the future which may not conform to the general "shape" of these two.

hiddeco:
Single-file multi-output is likely the best option to keep things simple (i.e., the second option).

krancour:
Agree with @hiddeco. Most existing steps deal with one file (or one _thing_ -- one HTTP request, one PR, etc.) at a time. `argocd -update` seems like a conspicuous exception to that, but in that case, I think there are good reasons for it.

jessesuen:
@fykaa - can we fast follow this with the `yaml` version of this?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/json-parse.md
```diff
@@ -0,0 +1,77 @@
...
+| `outputs` | `[]object` | Y | A list of rules for extracting values from the parsed JSON. |
+| `outputs[].name` | `string` | Y | The name of the output variable. |
+| `outputs[].fromExpression` | `string` | Y | An [expr-lang](https://expr-lang.org/) expression that can extract the value from the JSON file. Note that this expression should not be offset by `${{` and `}}`. See examples for more details. |
+
+## Expressions
+
+The `fromExpression` field supports [expr-lang](https://expr-lang.org/) expressions.
+
+:::note
+Expressions should _not_ be offset by `${{` and `}}` to prevent pre-processing evaluation by Kargo. The `json_parser` step itself will evaluate these expressions.
```
hiddeco, line 24:
```suggestion
Expressions should _not_ be offset by `${{` and `}}` to prevent pre-processing evaluation by Kargo. The `json-parse` step itself will evaluate these expressions.
```

hiddeco (approved):

PR #3331, opened 2025-01-21
refactor: change generic secret label

krancour:
@jessesuen @gdsoumya this PR changes the label used to indicate a Secret is a generic one as discussed offline.

New Secrets get created with the new label.

To not break backwards compatibility with v1.2.0, updates will swap the legacy label for the new one and list operations will list all Secrets labeled with one or both.

There is a small breaking change, which will be documented in the release notes:

Expressions now have access only to appropriately labeled Secrets. That should have been included in #2975, but was accidentally overlooked.


api/v1alpha1/labels.go
```diff
@@ -8,8 +8,12 @@ const (
 	CredentialTypeLabelValueGit   = "git"
 	CredentialTypeLabelValueHelm  = "helm"
 	CredentialTypeLabelValueImage = "image"
+	CredentialTypeLabelGeneric    = "generic"
 
 	// Project Secrets
+	// Deprecated: Use CredentialTypeLabelGeneric instead. This label should not
+	// be used and won't be documented, but for now we plan to support it
+	// indefinitely for backwards compatibility.
```
hiddeco, line 16: Is there a need to support this indefinitely instead of just phasing it out in 2 minor versions from now?

krancour: I'd prefer that as well, although it doesn't feel especially onerous in this case to continue.

@jessesuen wdyt?

jessesuen: Agree with Hidde. Would rather not keep it around too long

krancour: Ok. Will adjust the comment accordingly.

krancour: Fixed in 27c0c77f089062cde33780bb84efe37f6dba4e39.

jessesuen:
As part of this, let's also get rid of `kargo.akuity.io/analysis-run-template` label, which was never used/documented/etc...

krancour:
> As part of this, let's also get rid of kargo.akuity.io/analysis-run-template label, which was never used/documented/etc...

@jessesuen there are spots in the code where we list _only_ Secrets or ConfigMaps that have that label. When building an AnalysisRun we also refuse references to any ConfigMap or Secret not labeled in that way.

Can you clarify for me how you want those behaviors to change? Are we simply eliminating those constraints? Or, more likely, are we eliminating that constraint for ConfigMaps and for Secrets, we will require the new generic cred label instead?

jessesuen:
> there are spots in the code where we list only Secrets or ConfigMaps that have that label

It must not be enforced with the Job metric. In other words, I can currently create a job that references a secret to be used as an env var, and it currently works. So the label is not really fulfilling it's purpose, if that was the intent.

krancour:
> When building an AnalysisRun we also refuse references to any ConfigMap or Secret not labeled in that way.

> It must not be enforced with the Job metric. In other words, I can currently create a job that references a secret to be used as an env var, and it currently works. So the label is not really fulfilling it's purpose, if that was the intent.

My apologies. I was a bit hasty in my analysis of where this was used and I mis-identified a chunk of code that referenced that label.

This is used exclusively by API endpoints that don't appear ever to have been used.

So that label can go away easily.

jessesuen (approved):

PR #3340, opened 2025-01-22
docs: make old and new docs live side by side

krancour:
This pulls everything from the `newdocs` branch into `main` and moves some existing content in `main` around to put both "Old Docs" and "New Docs" at the same (top) level in the doc tree.

Screenshot 2025-01-22 at 3 12 24 PM


krancour:
> can you add a screenshot of the home page to the PR description?

Sure can. The automatically created preview of the PR would help even more, but the build failed... so I guess I had better straighten that out first.

kencochrane (approved):

PR #3346, opened 2025-01-23
fix(cmd): improve Argo integration checks

hiddeco:
This ensures that if the API server or controller is unable to determine if Argo CD or Argo Rollouts is installed (e.g. because of some transient Kubernetes API error), the respective service will fail to start instead of disabling the integration.

By doing this, the service will restart to get an accurate result once the API server is available again. Providing a more reliable result than the integrations not working for the lifetime of the Pod because of API server flakiness at boot time.

### DIFF
--- a/cmd/controlplane/api.go
+++ b/cmd/controlplane/api.go
@@ -81,17 +81,28 @@ func (o *apiOptions) run(ctx context.Context) error {
 		return fmt.Errorf("error creating Kubernetes client for Kargo API server: %w", err)
 	}
 
-	switch {
-	case !serverCfg.RolloutsIntegrationEnabled:
-		o.Logger.Info("Argo Rollouts integration is disabled")
-	case !argoRolloutsExists(ctx, restCfg):
-		o.Logger.Info(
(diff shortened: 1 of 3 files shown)


cmd/controlplane/api.go
```diff
@@ -81,17 +81,28 @@ func (o *apiOptions) run(ctx context.Context) error {
...
-		o.Logger.Info(
-			"Argo Rollouts integration was enabled, but no Argo Rollouts " +
-				"CRDs were found. Proceeding without Argo Rollouts integration.",
-		)
-		serverCfg.RolloutsIntegrationEnabled = false
-	default:
-		o.Logger.Info("Argo Rollouts integration is enabled")
+	if serverCfg.RolloutsIntegrationEnabled {
+		var exists bool
+		if exists, err = argoRolloutsExists(ctx, restCfg); !exists {
```
krancour, line 86: This only checks for an error if `exists` is `false`. Looking at the implementation of `argoRolloutsExists()`, `exists` _will_ always be `false` if there is an error, but wouldn't it be more idiomatic for the caller not to assume the function has been implemented that way?

krancour (approved):

PR #3363, opened 2025-01-25
fix: fix faulty logic for handling possible double-encoding of github app private key

krancour:
Follows up on #3055 and #3059

cc @hiddeco

The issue was here:

```
if corruptInputErr := new(base64.CorruptInputError); !errors.As(err, &corruptInputErr) {
```

`&corruptInputErr` is a `**base64.CorruptInputError`, which `err` never is.

Rather than just delete the offending `&`, I've refactored this and added a test to be sure this is fixed, but I've also switched over to a different strategy for identifying whether we're dealing with a "raw" PEM-encoded PKCS1 or PKCS8 key (i.e. one not unnecessarily double-encoded). It seems this might actually be more reliable than guessing we have a PEM-encoded key whenever the input happens not to be a valid base64 encoding.

But please lmk if you see any problems with the new approach or if you have any ideas for making it more robust.

### DIFF
--- a/internal/credentials/kubernetes/github/app.go
+++ b/internal/credentials/kubernetes/github/app.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/base64"
-	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/jferrl/go-githubauth"
@@ -137,16 +137,9 @@ func (a *appCredentialHelper) getAccessToken(
 	installationID int64,
 	encodedPrivateKey string,
 ) (string, error) {
-	decodedKey, err := base64.StdEncoding.DecodeString(encodedPrivateKey)
+	decodedKey, err := decodeKey(encodedPrivateKey)
 	if err != nil {
-		if corruptInputErr := new(base64.CorruptInputError); !errors.As(err, &corruptInputErr) {
-			return "", fmt.Errorf("error decoding private key: %w", err)
-		}
-
-		// If the key is not base64 encoded, it may be a raw key. Try using it
-		// as-is. We do this because initially, we required the PEM-encoded key
-		// to be base64 encoded (for reasons unknown today).
-		decodedKey = []byte(encodedPrivateKey)
+		return "", err
 	}
 	appTokenSource, err := githubauth.NewApplicationTokenSource(appID, decodedKey)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


hiddeco:
🤦

The reason I picked the previous (albeit faulty) logic is that we would not have to keep up with the input requirements from the GitHub API, and would typically produce more telling errors (from GitHub itself) when a key would not be double base64 encoded.

For example, if someone now adds a newline in front of their key, they will receive a vague base64 decode error rather than GitHub telling them something about the key being wrong.

krancour:
With your logic, if someone botches the extra encoding -- by missing a digit when copying + pasting, for instance -- the result is a surprising authn error from GitHub instead of a decoding error. But the docs don't _say_ to do that extra encoding, so I'll acknowledge that an incorrect key without extra encoding is a far more likely scenario than a botched encoding and your approach is better at that case. Sorry for not recognizing this initially.

My logic handles the less likely case better. You get a decoding error instead of an authn error.

I would be fine with just switching back to a corrected version of your logic (but keeping the new tests), but am also wondering if there isn't an approach that would handle all cases correctly.

wdyt if: We strip leading and trailing whitespace from the input and then check whether it contains non-base64 characters -- which a PEM encoded key always will. (Checking for this seems more forgiving than checking for the PEM header in the event GH starts using a different key format in the future.) If non-base64 characters are found we're probably not looking at a botched encoding, so we should use the value as is and let authn fail if that value isn't the correct key. On the other hand, if we see only base64 characters, we attempt to decode it. If it fails, it's more likely a botched encoding than an incorrect key, so we return the decoding error wrapped with a message informing the user that the extra encoding is unnecessary.

I think the only scenario that this isn't resilient to would be if GH Apps ever switch to a key format that uses base64 digits only and that does not seem likely at all.

I could be overthinking all of this, but this seems like it would return the correct type of error for all cases.

hiddeco:
I think this solution would be slightly over-engineered to support a use case that we accidentally introduced and primarily keep around to not butcher people their existing setups. 

Personally, I would stick with the current logic of trying to decode first, and use as-is if that fails. It's simpler, handles the backwards compatibility case we need, and people using double base64-encoded keys would have an already working setup anyway. Which dismisses the need for enriched errors in this situation.

However, I can live with that if you feel strongly about your solution.

internal/credentials/kubernetes/github/app.go
```diff
@@ -159,3 +155,28 @@ func (a *appCredentialHelper) getAccessToken(
...
+// whether to return the input as is or surface the decoding error. All other
+// errors are surfaced as is. This function is necessary because we initially
+// required the PEM-encoded key to be base64 encoded (for reasons unknown today)
+// and then we later dropped that requirement.
+func decodeKey(key string) ([]byte, error) {
+	decodedKey, err := base64.StdEncoding.DecodeString(key)
+	if err != nil {
+		if !errors.As(err, new(base64.CorruptInputError)) {
+			return nil, fmt.Errorf("error decoding private key: %w", err)
+		}
```
krancour, line 172: This is the correction to the original logic.

internal/credentials/kubernetes/github/app.go
```diff
@@ -159,3 +155,28 @@ func (a *appCredentialHelper) getAccessToken(
...
+		if base64Regex.MatchString(key) {
+			return nil, fmt.Errorf(
+				"probable corrupt base64 encoding of private key; base64 encoding "+
+					"this key is no longer required and is discouraged: %w", err,
+			)
+		}
```
krancour, line 178: This is the only part that's new.

hiddeco (approved):

Issue #3204, opened 2025-01-02
UI: scaling issue in pipeline view


Marvin9:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [ ] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

Our pipeline view doesn't work well when it comes to scaling. Here are list of issues that directly or indirectly relates

https://github.com/akuity/kargo/issues/3161
https://github.com/akuity/kargo/issues/2979
https://github.com/akuity/kargo/issues/2748
https://github.com/akuity/kargo/issues/2727

All of those come down to our custom graph rendering solution. I discussed this with @rpelczar softly some time back but now I am more leaning to delegate this problem to graph rendering library and keep ourselves focus on what matters. I am also confident that this isn't huge task in itself but still requires discussion on whether to move forward with new library or not.

# Version

```
v1.1.2
```

Marvin9:
Update:

discussed with @rpelczar and we agreed switch to graph library [xyflow](https://github.com/xyflow/xyflow)

krancour:
@Marvin how much of a change is that? (I'm not discouraging this. I'm just curious.)

Marvin9:
@krancour Very less compare to flexibility it brings. I had POC ready with more skeleton nodes done in few hours.

krancour:
@Marvin9 sounds like good news to me. Thanks!

Issue #3161, opened 2024-12-19
Kargo UI always centers and has slow scroll speed


moro-drake:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.

# Description

With the release of v1.1.1 the UI now centers on project space. This is good when the project does not involve too many stages, but when it's bigger - that combined with the slow scroll speed, make it impossible to interact with things that are far out from center.
With a structure like on the screenshot kargo will always center to the middle 'blue' Stage, making it really hard to interact with Warehouses and Stages that part of the  first and last pipeline in the attached picture.

I tested this on Windows PC chrome and edge, asked my colleagues to check it on their Mac - it is the same.

# Screenshots

![image](https://github.com/user-attachments/assets/30a8702d-9f2f-4c93-9068-4406aa21efee)

# Steps to Reproduce

1. Create a project with 5 separate Warehouses and stages, ideally with 2 or more subscriptions to each Warehouse
2. Inspect the UI - the scroll is slow and zooming in / out will always make it center to the 3rd 'pipeline' in the project.

# Version
v1.1.1

krancour:
This was on our radar, but we don't seem to have had an issue for it. Thanks for opening one!

aayushsrivastava:
In case we are open to suggestions...

Scrolling across the graph is a more common action. Zoom is something you adjust once in a while. Therefore, ideally the scrolling should require _lesser effort_ from the end-user than adjusting the zoom level.

Right now, we use vertical scroll for zoom and click-and-drag for scrolling across the graph. 

I think we should use vertical+horizontal scroll for scrolling the graph. For zoom, we can do a pinch and/or CMD+Vertical scroll. **Excalidraw** is probably a good source of inspiration for what I just described.

krancour:
@aayushsrivastava I agree. I think I frequently get tripped up by using the gestures I'm accustomed to for scrolling and finding it adjusts zoom level instead.

@Marvin9 what do you think about this suggestion?

Marvin9:
I agree with the suggestion and it was originally intended to work like that when I did pipeline view improvements. But I am now seeing the limitation of our custom graph UI and issues keep piling up for busy pipeline view. I will open up new issue for longer term solution

Marvin9:
fixed as part of https://github.com/akuity/kargo/issues/3204

PR #3391, opened 2025-01-29
feat(ui): migrate custom graph solution to `xyflow`

Marvin9:
part of https://github.com/akuity/kargo/issues/3204

this will replace our custom graph solution with [xyflow](https://reactflow.dev/) and free us from graph related issues.


krancour:
This is so much more responsive than what we had previously. I ❤️ this.

The only issue I've spotted so far is that we don't seem to be getting multiple differently colored connections between Stages when they request Freight from two different Warehouses.

Screenshot 2025-01-29 at 4 00 15 PM

For reference, this is how it would have rendered before:

multiple-warehouses-54c290ef5be66dce545d7e12c99c7491

Marvin9:
@krancour thanks! I have modified edge and will also implement to animate only the edge belongs to selected warehouse for

rpelczar:
Can we hide this connecting dot? 
image

rpelczar (approved):
Overall it looks great!

Can you remove all the leftovers? I found some functions not used anywhere in the utils folder.

aayushsrivastava:
I just test-drove the new pipeline UI and it is so much better now! Eagerly waiting for v1.3.0 just for this feature.

I am still wondering if it's possible to use more natural gestures for scrolling and zooming as mentioned [here](https://github.com/akuity/kargo/issues/3161#issuecomment-2567842483)? Willing to create a new issue if it needs more deliberation.

Marvin9:
Thanks @aayushsrivastava. Yes please open up new issue, though I think its expected that some people might be used to navigation like maps and some with wheel scroll so will need some discussion

Issue #3294, opened 2025-01-16
Secret management should not require horizontal scrolling in table


jessesuen:
# Description

I couldn't easily delete a secret because the delete/edit buttons were to the right of the entry in the secret table. When a user doesn't have a trackpad, then it's not easily accessible to scroll and delete/edit the secret.

We should stack the project secret management and the credential management, or create tabs. Instead of showing two tables side by side.

# Screenshots

<img width="1624" alt="image" src="https://github.com/user-attachments/assets/62fbe268-91d1-47d7-94a1-65e36ccf9fe9" />

# Version

```
v1.2.0
```

PR #3392, opened 2025-01-29
fix: vertical view for secrets and filter credentials by label

Marvin9:
fixes https://github.com/akuity/kargo/issues/3294 and API of credentials was previously returning generic secrets


krancour (approved):
🔥

PR #3408, opened 2025-01-30
feat: support Expression Language in Stage verification args

### LINKED ISSUE #2482: Pass dynamic information to AnalysisRun
# Proposed Feature

Today, we support hardwired arguments to an AnalysisRun:

```yaml
  verification:
    analysisTemplates:
    - name: cat-fact
    args:
    - name: foo
      value: bar
```

This is useful but somewhat limited. It would be more useful to be able to supply runtime information to the AnalysisRun, for example, information about the freight that it is running verification against.

# Motivation

The analysis may want to use the freight information for its query or test.

# Suggested Implementation

One approach is to introduce some variable feature, e.g.:

```yaml
  verification:
    analysisTemplates:
    - name: cat-fact
    args:
    - name: foo
      value: "${freight.image}"  
```

We are already thinking of variables as part of promotion directives, and this would likely need to come after that.

hiddeco:
Fixes #2482 

This enables the usage of [Expression Language](https://expr-lang.org) in the `spec.verification.args` of a `Stage`, while simultaneously making the `warehouse()`, `commitFrom()`, `imageFrom()`, and `chartFrom()` functions available.

### DIFF
--- a/internal/expressions/functions.go
+++ b/internal/expressions/functions.go
@@ -123,6 +123,10 @@ func warehouse(a ...any) (any, error) {
 		return nil, fmt.Errorf("argument must be string, got %T", a[0])
 	}
 
+	if name == "" {
+		return nil, fmt.Errorf("name must not be empty")
+	}
+
 	return kargoapi.FreightOrigin{
 		Kind: kargoapi.FreightOriginKindWarehouse,
 		Name: name,
--- a/internal/expressions/functions.go
+++ b/internal/expressions/functions.go
@@ -1,0 +1,275 @@
+package expressions
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/expr-lang/expr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/controller/freight"
+)
+
+type exprFn func(params ...any) (any, error)
+
+// FreightFunctions returns a slice of expr.Option containing functions for
+// Freight operations.
+//
+// It provides `warehouse()`, `commitFrom()`, `imageFrom()`, and `chartFrom()`
+// functions that can be used within expressions. The functions operate within
+// the context of a given project with the provided freight requests and
+// references.
+func FreightFunctions(
+	ctx context.Context,
+	c client.Client,
+	project string,
+	freightRequests []kargoapi.FreightRequest,
+	freightRefs []kargoapi.FreightReference,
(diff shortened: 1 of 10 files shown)


internal/expressions/functions.go
```diff
@@ -0,0 +1,279 @@
...
+		new(func(repoURL string, origin kargoapi.FreightOrigin) kargoapi.Image),
+		new(func(repoURL string) kargoapi.Image),
+	)
+}
+
+// ChartFromFunction returns an expr.Option that provides a `chartFrom()`
+// function for use in expressions.
+//
+// The chartFrom function finds Helm charts based on repository URL, optional
+// hart name, and optional origin, using the provided freight requests and
```
krancour, line 93:
```suggestion
// chart name, and optional origin, using the provided freight requests and
```

internal/expressions/functions.go
```diff
@@ -0,0 +1,279 @@
...
+//
+// The returned function uses freight requests and references to locate the
+// appropriate commit within the project context.
+func getCommit(
+	ctx context.Context,
+	cl client.Client,
+	project string,
+	freightReqs []kargoapi.FreightRequest,
+	freightRefs []kargoapi.FreightReference,
+) func(a ...any) (any, error) {
```
krancour, line 147:
```suggestion
) exprFn {
```

krancour:
I have no major concerns. I have just one minor one.

Until now, I'd intentionally kept domain-specific functions like `commitFrom()` out of the `internal/expressions` package. To the extent possible, I try to do that with packages I see as "low level" (`internal/git` being another good example) in hopes that it minimizes the likelihood of encountering import cycles, which are obviously no fun to resolve, as we find new uses for the packages in the future.

If you think there's merit to that, I think we could move domain-specific bits out of `internal/expressions` and into their own package.

If you disagree, I'm not going to fight you on it, as this is overall a slam dunk. 🏀 

LGTM'ing and leaving it up to you.

krancour (approved):

hiddeco:
> If you think there's merit to that, I think we could move domain-specific bits out of internal/expressions and into their own package.

To prevent circular dependencies, I moved the function option bits out into a sub-module. This way we still bundle them with expressions (as I do believe there is a strong relationship between them), but without causing potential issues if we want to use the expression template logic elsewhere (for as long as we do not _also_ require the function options — which would be problematic in all cases).

The reason I picked this over moving them to some place next to `expressions` is that the function signatures and "arbitrary parameter resolving" are _very specific_ to the EL engine.

krancour:
Yep. That's perfect!

PR #3456, opened 2025-02-06
feat: surface likely oidc misconfiguration through ui

### LINKED ISSUE #3455: invalid OIDC info causes kargo api pods to crash
# Checklist

# Description

Invalid OIDC info like an invalid issuer url causes kargo api pods to crash forever. This can be handled more gracefully by the server by showing an error message to the user when they try to login.

# Steps to Reproduce

Set invalid OIDC config like a non-existent issuer url

# Version

```
v1.2.2
```

krancour:
Fixes #3455

#3455 reports the API server crashing during startup if OIDC is misconfigured. tbh, this was the desired behavior. It's hard to tell the difference between misconfiguration (e.g. incorrect issuer URL) and transient errors (common if using Dex and the API server starts faster than Dex), so the k8s way is to let it die and try to start again with hopes of success, however...

1. OIDC misconfiguration is common.

1. If the API server is down, the UI is also inaccessible. If a user doesn't have access to the logs and cannot reach the UI, there is no insight at all into this common problem.

This change allows the only interaction between the API server and the IDP (retrieval of the keyset for token validation) to fail gracefully and be retried just-in-time on future attempts to validate a token, until it succeeds. (And it might never if things are misconfigured.)

This effectively results in the following:

* Invalid issuer URL:

  Keyset retrieval will fail, but the server will not crash.

  * If any request requiring authn is attempted with an _existing_ token, that token would have been issued by an IDP not matching the invalid issuer URL. As such, the token cannot be validated and the user will be asked to log in.
  * If the user attempts to login, they will be doing so with an invalid issuer URL and the _client_ will fail to communicate with the IDP. They will see this, which hints well-enough at ODIC being misconfigured:

    Screenshot 2025-02-06 at 9 54 23 AM

* Transient issue:

  Keyset retrieval will fail, but the server will not crash.

  * If any request requiring authn is attempted with an _existing_ token, keyset retrieval will be retried prior to the attempt to validate.
  * If the user attempts to login, they will be doing so with a valid issuer URL and the authorization code flow w/ PKCE will succeed. This doesn't involve the API server. For validation of the newly-issued token, see previous bullet.

cc @gdsoumya @hiddeco

### DIFF
--- a/internal/api/option/auth.go
+++ b/internal/api/option/auth.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"sync"
 
 	"connectrpc.com/connect"
 	"github.com/coreos/go-oidc/v3/oidc"
@@ -23,6 +24,7 @@ import (
 	"github.com/akuity/kargo/internal/api/config"
 	"github.com/akuity/kargo/internal/api/user"
 	"github.com/akuity/kargo/internal/indexer"
+	"github.com/akuity/kargo/internal/logging"
 )
 
 const authHeaderKey = "Authorization"
@@ -49,7 +51,7 @@ type authInterceptor struct {
 	verifyIDPIssuedTokenFn   func(
 		ctx context.Context,
 		rawToken string,
-	) (claims, bool)
+	) (claims, error)
 	oidcTokenVerifyFn     goOIDCIDTokenVerifyFn
 	oidcExtractClaimsFn   func(*oidc.IDToken) (claims, error)
 	listServiceAccountsFn func(
@@ -66,17 +68,13 @@ func newAuthInterceptor(
 	ctx context.Context,
 	cfg config.ServerConfig,
 	client libClient.Client,
-) (*authInterceptor, error) {
+) *authInterceptor {
 	a := &authInterceptor{
 		cfg:            cfg,
 		internalClient: client,
 	}
 	if cfg.OIDCConfig != nil {
-		var err error
-		a.oidcTokenVerifyFn, err = newMultiClientVerifier(ctx, cfg)
-		if err != nil {
-			return nil, err
-		}
+		a.oidcTokenVerifyFn = newMultiClientVerifier(ctx, cfg)
 	}
 	a.parseUnverifiedJWTFn =
 		jwt.NewParser(jwt.WithoutClaimsValidation()).ParseUnverified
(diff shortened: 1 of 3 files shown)


internal/api/option/auth.go
```diff
@@ -410,29 +418,40 @@ func (a *authInterceptor) authenticate(
...
+	if a.cfg.OIDCConfig == nil {
+		// Really, this method never should have been called under these
+		// circumstances.
+		return claims{}, errors.New("OpenID Connect is not supported")
+	}
 	c := claims{}
 	if a.oidcTokenVerifyFn == nil {
-		return c, false
+		verifierMu.Lock()
+		defer verifierMu.Unlock()
```
hiddeco, line 438: We probably want to unlock the mutex as soon as we can, and there appears to be quite some logic here before the `defer` will be called.

krancour: Initially I was and then opted for clarity. The claim extraction is very quick. They're just raw claims. It doesn't include any of the claims --> SA mapping logic.

krancour: Ah... but there's also the actual verification before that, which is a cryptographic process which may be a tad slow.

I'll fix.

krancour: Fixed in 445cdfe72421ecc83d3eee2d7218cc0ce034f83b

internal/api/option/auth.go
```diff
@@ -435,9 +435,9 @@ func (a *authInterceptor) verifyIDPIssuedToken(
 	c := claims{}
 	if a.oidcTokenVerifyFn == nil {
 		verifierMu.Lock()
-		defer verifierMu.Unlock()
 		if a.oidcTokenVerifyFn == nil {
 			a.oidcTokenVerifyFn = newMultiClientVerifier(ctx, a.cfg)
+			verifierMu.Unlock()
```
hiddeco, line 440: Think this should be something like:

```go
verifierMu.Lock()
if a.oidcTokenVerifyFn == nil {
    a.oidcTokenVerifyFn = newMultiClientVerifier(ctx, a.cfg)
}
verifier := a.oidcTokenVerifyFn
verifierMu.Unlock()

if verifier == nil {
    return c, errors.New(
        "could not validate token, possibly due to a transient network " +
            "error; if the problem persists, check your OpenID Connect " +
            "configuration",
    )
}
```

krancour: 607b9116de35c5c1ae1037ad1c8231254a853f4f

hiddeco (approved):

Issue #3452, opened 2025-02-06
Go Subprocess (exec.Command) Does Not Inherit Environment Variables for the proxy settings from main controller process


abelhoula:
# Checklist

* [x ] I've searched the issue queue to verify this is not a duplicate bug report.
* [ x] I've included steps to reproduce the bug.
* [x ] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description
We encountered an issue where the `exec.Command("git", args...)` subprocess in our Krago controller does not inherit the environment variables from the controller process. This caused our Git client to fail when running in a GitLab environment behind a proxy.

## Steps to Reproduce
1. Set the environment variables in the kargo controller :
   ```sh
   export HTTP_PROXY=http://proxy.example.com
   export HTTPS_PROXY=http://proxy.example.com
2. The Git command fails with a failed to clone repo  The requested URL returned error: 403
3. while opening a shell inside the kargo controller pod & running same command (as the warehouse) works fine 
# Version

```
1.2.0
```

krancour:
Closed by #3452

PR #3461, opened 2025-02-07
feat: add proxy settings to git client

abelhoula:


abelhoula:
related to #3452

hiddeco:
I think the correct patch would be:

```go
	if b.creds != nil && b.creds.Password != "" {
		cmd.Env = append(cmd.Env, "GIT_ASKPASS=/usr/local/bin/credential-helper")
		cmd.Env = append(cmd.Env, fmt.Sprintf("GIT_PASSWORD=%s", b.creds.Password))
	}
	if httpProxy := os.Getenv("HTTP_PROXY"); httpProxy != "" {
		cmd.Env = append(cmd.Env, fmt.Sprintf("http_proxy=%s", httpProxy))
	}
	if httpsProxy := os.Getenv("HTTPS_PROXY"); httpsProxy != ""
		cmd.Env = append(cmd.Env, fmt.Sprintf("https_proxy=%s", httpsProxy))
	}
```

As you can have a proxy set without having credentials set, and you can have a HTTP proxy set without the HTTP/S proxy being set.

internal/controller/git/base_repo.go
```diff
@@ -232,13 +232,20 @@ func (b *baseRepo) buildCommand(command string, arg ...string) *exec.Cmd {
...
 	cmd.Env = append(cmd.Env, fmt.Sprintf("GIT_SSH_COMMAND=ssh -F %s/.ssh/config", b.homeDir))
+
 	if b.creds != nil && b.creds.Password != "" {
-		cmd.Env = append(
-			cmd.Env,
-			"GIT_ASKPASS=/usr/local/bin/credential-helper",
-			fmt.Sprintf("GIT_PASSWORD=%s", b.creds.Password),
-		)
+		cmd.Env = append(cmd.Env, "GIT_ASKPASS=/usr/local/bin/credential-helper")
+		cmd.Env = append(cmd.Env, fmt.Sprintf("GIT_PASSWORD=%s", b.creds.Password))
```
krancour, line 238: This was more efficient as previously written, as it resized the slice just once. Now it's being resized twice. It's a quite a small difference, but if there isn't a good reason for the change, I'd rather see it left alone.

internal/controller/git/base_repo.go
```diff
@@ -232,13 +232,20 @@ func (b *baseRepo) buildCommand(command string, arg ...string) *exec.Cmd {
...
+	}
+
+	if httpProxy := os.Getenv("HTTP_PROXY"); httpProxy != "" {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("http_proxy=%s", httpProxy))
 	}
+
+	if httpsProxy := os.Getenv("HTTPS_PROXY"); httpsProxy != "" {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("https_proxy=%s", httpsProxy))
+	}
```
krancour, line 247: `http_proxy` and `https_proxy` are well-known env vars respected by many tools. I'd prefer that we use those (lower case) instead of `HTTP_PROXY` and `HTTPS_PROXY`. i.e. If they're defined, just copy them as-is into the command's env.

It would probably be a good idea to make these configurable via the chart as well.

krancour (approved):

hiddeco (approved):

PR #3465, opened 2025-02-07
docs: Fix example for Helm warehouses

mimartin12:
While going through the docs and using an example, I noticed that I was getting an error for using
```yaml
spec:
  subscriptions:
  - helm:
      repoURL: https://charts.example.com
      name: my-chart
      semverConstraint: ^1.0.0
```
After noticing that creating a Helm chart warehouse via the UI, it should be
```yaml
spec:
  subscriptions:
  - chart:
      repoURL: https://charts.example.com
      name: my-chart
      semverConstraint: ^1.0.0
```

This simply fixes the example.


krancour (approved):
Thank you!

Issue #3125, opened 2024-12-12
Conditional step execution


muenchdo:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

Add the possibility to add conditions for promotion step executions, similar to what is available in [GitHub Actions](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution). The expressions added with v1.1 would make this very useful.

# Motivation

Right now the main use case I have in mind is being able to run something like a "finalizer" to send a Slack notification when the promotion fails, but I feel like this would be useful in more cases.

# Suggested Implementation

GitHub Actions seems to be a good inspiration here. 

```yaml
steps:
- uses: foo
  alias: fooStep
  config:
    option1: value1
    option2: value2
- uses: bar
  if: ${{ len(outputs.fooStep.someOutput) > 0 }}
  config:
    option1: value1
    option2: value2
```

For my finalizer use case, we would need some [special expressions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions) such as `$always()`.

Looking at #3124, it seems there are technical limitations that make this difficult, but I do think this (especially the finalizer use case) is less of an edge case than loops.

krancour:
You're on the same page as me. I think `if` is fairly easy to do. As for "always," I think we could potentially add something like a "failure policy" that can be configured to allow a process (or designated steps of a process) to continue after a failure. This bit is harder and would require some design work. If we split "always" into a separate issue, "if" on its own can probably get done faster.

PR #3485, opened 2025-02-11
feat: conditional step execution

hiddeco:
Fixes: #3125


api/v1alpha1/promotion_types.go
```diff
@@ -207,6 +207,9 @@ type PromotionStep struct {
 	Task *PromotionTaskReference `json:"task,omitempty" protobuf:"bytes,5,opt,name=task"`
 	// As is the alias this step can be referred to as.
 	As string `json:"as,omitempty" protobuf:"bytes,2,opt,name=as"`
+	// If is an optional expression that, if present, must evaluate to a boolean
+	// value. If the expression evaluates to false, the step will be skipped.
+	If string `json:"if,omitempty" protobuf:"bytes,7,opt,name=if"`
```
krancour, line 212: Would it be worth describing what happens if it doesn't evaluate to a boolean? As a reader who hasn't looked at the code, I might wonder if it's an error or is implicitly false.

krancour:
LGTM with one minor suggestion about docs.

krancour (approved):

krancour:
Also love how the expression support continues to pay dividends!

Issue #3483, opened 2025-02-11
Freight is always empty in verifications tab


alexmt:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

The Verifications tab has Freight column but it is empty.

# Screenshots

<img width="1370" alt="Image" src="https://github.com/user-attachments/assets/3450cbdd-a051-496f-9b5c-ee2f085237bd" />

# Steps to Reproduce

1. Create any stage and promote a freight.
2. See promotions tab
3. The freight column is empty

# Version

```
v1.2.2
```

hiddeco:
I believe this is due to Freight metadata not being fed into the verifications tab, as it appears to be based on a list of `VerificationInfo`s:

- https://github.com/akuity/kargo/blob/v1.2.2/ui/src/features/stage/stage-details.tsx#L42-L56
- https://github.com/akuity/kargo/blob/v1.2.2/ui/src/features/stage/verifications.tsx#L118-L124

PR #3486, opened 2025-02-11
fix(ui): empty freight value in column

Marvin9:
fixes https://github.com/akuity/kargo/issues/3483


krancour (approved):
@Marvin9, thank you for being so quick to fix this!

PR #3516, opened 2025-02-18
fix: invalid type conversion for the `git-wait-for-pr` step

nitishfy:
This PR fixes an issue where Kargo incorrectly interprets a PR number as a boolean if the PR number is "1". This causes the `gh-wait-for-pr` promotion step to fail with the error:

`invalid type, expected number, given boolean`

- If a PR number evaluates to "1", Kargo mistakenly treats it as a boolean (true).
- The PR itself is created successfully and can be merged in GitHub, but Kargo fails to detect it, preventing the promotion from proceeding.

## Before
Screenshot from 2025-02-18 15-04-27

## After
Screenshot from 2025-02-18 15-23-28

### DIFF
--- a/internal/expressions/json_templates.go
+++ b/internal/expressions/json_templates.go
@@ -155,14 +155,19 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
 		}
 		return result, nil
 	}
-	// If the result is parseable as a bool return that.
-	if resBool, err := strconv.ParseBool(result); err == nil {
-		return resBool, nil
-	}
+
+	// NOTE: The result is explicitly parsed as numbers before booleans to prevent values like
+	// "0" and "1" from being misinterpreted as boolean `false` and `true`,
+	// respectively. See: https://github.com/akuity/kargo/pull/3516
+
 	// If the result is parseable as a float64, return that. float64 is used
 	// because it can represent all JSON numbers.
 	if resNum, err := strconv.ParseFloat(result, 64); err == nil {
 		return resNum, nil
+	}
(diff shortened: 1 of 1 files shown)


krancour:
@hiddeco this fix lgtm. Wanted to validate that you agree.

hiddeco (commented):
I agree with the solution but I think it would be wise to add a comment about numbers explicitly being parsed before booleans to avoid detecting `0` and `1` as bools.

internal/expressions/json_templates.go
```diff
@@ -155,15 +155,20 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
...
-	if resBool, err := strconv.ParseBool(result); err == nil {
-		return resBool, nil
-	}
+
+	// NOTE: The result is explicitly parsed as numbers before booleans to prevent values like
+	// "0" and "1" from being misinterpreted as boolean `false` and `true`,
+	// respectively. See: https://github.com/akuity/kargo/pull/3516
+
 	// If the result is parseable as a float64, return that. float64 is used
 	// because it can represent all JSON numbers.
```
krancour, line 164:
```suggestion
	// If the result is parseable as a float64, return that. float64 is used
	// because it can represent all JSON numbers.
	//
	// NB: This is attempted prior to attempting to parse the result as a boolean
	// so that "0" and "1" will be interpreted as numbers.
```

krancour: I don't think we need a reference to the PR living in here forever when the comment explains everything there is to know about why we're doing this quite clearly.

hiddeco (approved):

PR #3520, opened 2025-02-19
fix: Long Stage names result in AnalysisRun creation failures during promotion

### LINKED ISSUE #3202: bug: Unable to create AnalysisRun due to long Stage name
# Checklist

# Description
Execution of stages with long name (>=30 characters) and `spec.verification` configured results in failure to create `AnalysisRun` resources.  This is due to the `kargo.akuity.io/promotion` label value exceeding the 63 character limit.

# Steps to Reproduce
1. Create a `Stage` with a sufficiently long name (>30 characters)
2. Configure a `spec.verification` step that triggers an AnalysisTemplate - any configuration will do.
3. Promote a Freight to the `Stage` with the long name.
4. Note that the verification step fails to create the AnalysisRun resource.

# Version
```
v1.1.2
```

# Logs
Error message:
```
error creating AnalysisRun "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.01jgjn7248tgmb6xq3wxzrqp9s.43776e4" in namespace "kargo-xxxxx": AnalysisRun.argoproj.io "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.01jgjn7248tgmb6xq3wxzrqp9s.43776e4" is invalid: metadata.labels: Invalid value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.01jgjn66w0kkacc9yaajzc0713.56d0e7f": must be no more than 63 characters
```

# Investigation
The long name generation looks to come from the `generateName` method [here](https://github.com/akuity/kargo/blob/main/internal/rollouts/analysis_run_builder.go#L100).  It's being set as a label on the AnalysisRun [here](https://github.com/akuity/kargo/blob/main/internal/controller/stages/regular_stages.go#L1243) from the previous promotion value.

The ULID + Freight ID take a fair chunk of the available real estate - perhaps `generateName()` should simply truncate (or hash + truncate) long prefixes.  Not sure what the implications of diverging from the 1:1 Stage name are though.

For a workaround, I will likely do some internal truncating of the Stage name to keep it <=29 chars though this may impact readability and identification of the stages running (and what environment they represent).  Are there plans to display additional Stage metadata in the UI via something like annotations?

aidan-canva:
Closes #3202 

Long Stage names cause errors during promotions due to calculated labels (`kargo.akuity.io/promotion`) on AnalysisRun resources exceeding the max 63 character limit of a label value.

This solution moves the `kargo.akuity.io/promotion` label to an annotation which has a much larger max length.  It doesn't appear this label key is used for selecting at the moment so its largely an informative value.  

An alternative approach to truncate Stage name was considered (with hashing for uniqueness) but that greatly reduces legibility within the UI (IMO).  Doing it inconsistently might work, but then impacts ability to debug and map relationships with resources easily.

This solution strikes a middle ground to unblock longer Stage names and give time for a more considered approach to naming/labeling conventions if its required.


krancour:
Tentatively assigned to v1.3.0, but might be a good candidate to be deferred.

internal/rollouts/analysis_run_builder.go
```diff
@@ -120,17 +121,24 @@ func (b *AnalysisRunBuilder) buildMetadata(
 	namespace, name string,
 	metadata *kargoapi.AnalysisRunMetadata,
 	extraLabels map[string]string,
+	extraAnnotations map[string]string,
```
hiddeco, line 124:
```suggestion
	extraLabels, extraAnnotations map[string]string,
```

aidan-canva:
Thanks @hiddeco, fixed!

hiddeco (approved):
Double checked, and this should at this point not break anything anywhere. Thank you @aidan-canva ☕

aidan-canva:
Thanks @hiddeco !

PR #3529, opened 2025-02-20
feat: yaml-parse promotion step

fykaa:


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-parse.md
```diff
@@ -0,0 +1,86 @@
+---
+sidebar_label: yaml-parse
+description: Parses a YAML string and extracts values based on specified expressions.
+---
+
+# `yaml-parse`
```
hiddeco, line 6: I have been thinking some more about this, and it might be easier if parse steps were prefixed with `parse-`. I.e., `parse-yaml`, `parse-json`, etc., what do you think?

internal/directives/schemas/yaml-parse-config.json
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "YAMLParseConfig",
+
+  "definitions": {
+    "yamlParse": {
```
hiddeco, line 6: As this appears to be equal to `jsonParse` in the other schema, I wonder if we would benefit from a common type that's shared between the two schemas. By doing this, I suspect we can share much more code between the two parsers.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-parse.md
hiddeco: Given we already have a `yaml-update` step, we should address this at a later moment while simultaneously aliasing that step to e.g. `update-yaml`.

internal/directives/schemas/yaml-parse-config.json
hiddeco: Deferred as it is non-blocking.

hiddeco (approved):

PR #3320, opened 2025-01-19
feat(gitprovider): setup gitea provider

Tchoupinax:
Hello,

As a big user of Gitea, I would love to have this provider included in Kargo!

Inspiration:
- https://github.com/akuity/kargo/issues/2569
- https://github.com/akuity/kargo/pull/3128

Closes https://github.com/akuity/kargo/issues/3321

Issue #2198, opened 2024-06-24
Enable Http git repository authentication to be able to clone and push to unsecured repositorys.



tal-hason:
# Proposed Feature

Add a http authentication check box to the creds section that will enable the end-user to use http git repositorys with basic authentication.

# Motivation

Let customers deicide the secure level they can use duo to their network constraints.

# Suggested Implementation

Add a "http - basic auth" check box to the credentials section that will enable unsecure git connection.

As discussed on:
[PR1504](https://github.com/akuity/kargo/pull/1504)

PR #3530, opened 2025-02-20
fix(gitprovider): use provided scheme for API calls

hiddeco:
Follow up on https://github.com/akuity/kargo/pull/3320#pullrequestreview-2587924812

This ensures that if explicitly defined, a connection over HTTP can still be made instead of silently enforcing HTTPS. This is particularly useful when e.g. running a self-hosted Gitea or GitLab instance in a homelab environment without a proper TLS configuration.

### DIFF
--- a/internal/gitprovider/gitea/gitea.go
+++ b/internal/gitprovider/gitea/gitea.go
@@ -85,7 +85,7 @@ func NewProvider(
 		opts = &gitprovider.Options{}
 	}
 
-	host, owner, repo, err := parseRepoURL(repoURL)
+	scheme, host, owner, repo, err := parseRepoURL(repoURL)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


krancour:
This LGTM on its own, but with what I think is a safe assumption that APIs of all these Git hosting providers don't permit any unauthenticated access (or at least won't allow opening a PR without authenticating), this change might not have any practical effect unless we address #2198 also.

https://github.com/akuity/kargo/blob/a8dfa3453ce4566c947b26adcd74bf114020c664/internal/credentials/kubernetes/database.go#L82-L90

I remain of the opinion that sending credentials over an insecure protocol is playing with fire, but I've come around to the idea of an option in the chart that allows operators to opt into that, as long as it also comes with a sufficiently stern warning.

I'd be equally supportive of this PR being amended with that change or addressing it in a follow-up instead.

For now, LGTM'ing this as is on its own merits.

krancour (approved):

hiddeco:
Will tackle credentials for HTTP in a separate PR.

PR #3536, opened 2025-02-21
docs: documentation for new features

hiddeco:
Adds documentation for:

- #3320
- #2673 
- #3257

### DIFF
--- a/docs/docs/50-user-guide/20-how-to-guides/40-working-with-stages.md
+++ b/docs/docs/50-user-guide/20-how-to-guides/40-working-with-stages.md
@@ -108,14 +108,14 @@ spec:
 
 In the next example, the `prod` `Stage` requests `Freight` that has originated
 from the `my-warehouse` `Warehouse`, but indicates that it will accept such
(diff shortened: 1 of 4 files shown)


docs/docs/50-user-guide/20-how-to-guides/40-working-with-stages.md
```diff
@@ -98,6 +106,30 @@ spec:
...
+spec:
+  requestedFreight:
+  - origin:
+      kind: Warehouse
+      name: my-warehouse
+    sources:
+      stages:
+      - test
+      - uat
+    availabilityStrategy: All
```
krancour, line 129: Nit: If someone is really paying close attention, I think the previous example establishes uat as being downstream from test. That context makes this second example somewhat odd (albeit not invalid) as the pipeline would look like this:

```
test --> uat --> prod
  \___________/
```

With an `All` strategy, that isn't any different from:

```
test --> uat --> prod
```

If we just changed `test` to `qa` or something here, it might be less odd.

krancour (approved):

PR #3538, opened 2025-02-21
docs: Update 50-analysis-templates.md

Utwo:
I think here in order for the verification to pass, you will also need a count.


krancour:
@jessesuen you know Rollouts better than myself or @hiddeco.

Would you be willing to review this one-liner?

Utwo:
> @jessesuen you know Rollouts better than myself or @hiddeco.
> 
> Would you be willing to review this one-liner?

I'm not an expert on Rollouts as well but I've tried it locally and I only managed to make it pass using the above suggestion.

jessesuen (approved):
Nice catch!

PR #3539, opened 2025-02-21
feat: allow opt-in to credentials over HTTP

hiddeco:
Historically, we have not allowed the credentials database to do lookups for HTTP URLs out of security considerations. However, in some scenarios, this may still be useful as long as the user is aware of the implications.

To facilitate this, add an `ALLOW_CREDENTIALS_OVER_HTTP` environment variable which allows looking up credentials for HTTP URLs, and wire this into the chart. 

Follow up to #3530


krancour:
cc @tal-hason, you get your wish from #2198

krancour (approved):
Thank you!

Issue #3548, opened 2025-02-24
A property in CRD marked as required although said to be optional


anttijman:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

Not sure whether this is a bug or issue but decided to create one nevertheless.

We are doing some validation for our manifests in CI. Noticed that it started to fail for kargo stages. It says: `stage-kargo-v1alpha1.json#/properties/spec/properties/requestedFreight/items/properties/sources/required: missing properties: 'availabilityStrategy'`.
Then I went to see whether there were changes to the CRD and found this commit from few days ago: https://github.com/akuity/kargo/commit/e8b8e2cf7875c3777607b78e311f6bd0569021f2#diff-c047caa556db761926b2756556157c65ff10fac49cea571ed076b9873aeca783R301. And to me that seems wrong as the property itself has a default value and should be optional. 
To fix our CI validation issue I would have to add this property to all my stages. Which I can do but does not seem right. 

Also in warehouse CRD there is the same issue with strictSemvers. See for example: https://github.com/akuity/kargo/blob/main/charts/kargo/resources/crds/kargo.akuity.io_warehouses.yaml#L369
We resolved that earlier by specifying strictSemvers for all our warehouses..

hiddeco:
While I do think your concern is valid (and we should mark those fields as optional), I do think you should be aware that you appear to be testing your custom resources against an unreleased version of Kargo. As `availabilityStrategy` has not be made available in an official release (yet).

PR #3550, opened 2025-02-24
fix(api): make `availabilityStrategy` optional 

hiddeco:
Fixes: #3548 

This ensures that the newly introduced `availabilityStrategy` is not forcefully marked as `required` but instead allows it to be left unspecified, and then (code)automagically defaults to `OneOf`.

The issue also mentions a historical problem with the `semverConstraint` field being required (and arguably, being confusing as it defaults to `true` even if the field has no further impact on the configured selection strategy, i.e. it does not target SemVers) — I do not see an immediate path forward here to resolve this, as protobuf isn't familiar with e.g. nullable booleans.

### DIFF
--- a/api/v1alpha1/generated.proto
+++ b/api/v1alpha1/generated.proto
@@ -438,7 +438,7 @@ message FreightSources {
   // made available to the Stage. This field is optional. When left unspecified,
   // the field is implicitly treated as if its value were "OneOf".
   //
-  // +kubebuilder:default=OneOf
+  // +kubebuilder:validation:Optional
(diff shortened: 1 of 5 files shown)


krancour:
> I do not see an immediate path forward here to resolve this, as protobuf isn't familiar with e.g. nullable booleans.

It's unfortunate, but at least there's this already in the field's description:

> This is enabled by default, but only has any effect when the ImageSelectionStrategy is SemVer.

Hopefully confusion is minimal.

krancour (approved):

PR #3578, opened 2025-02-27
fix: require tag or digest for kustomize-set-image

hiddeco:
After `fromOrigin` fields were removed, not having either defined could result in the tag of the `image` unexpectedly being set to `latest`. To avoid this, explicitly require either `tag` or `digest` to be set.

### DIFF
--- a/internal/directives/kustomize_image_setter_test.go
+++ b/internal/directives/kustomize_image_setter_test.go
@@ -61,6 +61,32 @@ func Test_kustomizeImageSetter_validate(t *testing.T) {
 			},
 		},
 		{
+			name: "digest and tag are both not specified",
+			config: Config{
+				"images": []Config{{
+					"image": "fake-image",
+				}},
+			},
+			expectedProblems: []string{
+				"images.0: Must validate one and only one schema (oneOf)",
+				"images.0: tag is required",
+			},
+		},
+		{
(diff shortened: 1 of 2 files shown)


internal/directives/kustomize_image_setter_test.go
```diff
@@ -60,6 +60,32 @@ func Test_kustomizeImageSetter_validate(t *testing.T) {
...
+		{
+			name: "digest and tag are both not specified",
+			config: Config{
+				"images": []Config{{
+					"image": "fake-image",
+				}},
+			},
+			expectedProblems: []string{
+				"images.0: Must validate one and only one schema (oneOf)",
+				"images.0: tag is required",
```
krancour, line 72: I'm trying to wrap my head around why this second error message would exist under these conditions.

Same for line 86.

It may just be a sign that I need more ☕

krancour: That sound familiar now that you've said it.

It might possibly be less confusing to remove the assertion about `"images.0: tag is required"` since that's a bit of weirdness and it's really just the first of the two messages that reflects what we were expecting.

wdyt?

hiddeco: Fine with me, updated in da75949d76970f721a54128a914ba39ad7dd8ed5

Issue #3081, opened 2024-12-06
make promoter info (username and email, at least) available in expressions


Sebast1aan:
# Checklist

* [ X] I've searched the issue queue to verify this is not a duplicate feature request.
* [ X] I've pasted the output of `kargo version`, if applicable.
* [ X] I've pasted logs, if applicable.

# Proposed Feature

When currently using the `git-commit` promotion step, the commits are done with the default user `kargo` (if no explicit `author.name` is set). If a promotion is done manually, it would be useful that this becomes `kargo-$logged-in-user` (such as kargo-admin, in case you log in with the admin account). In the promotion history you can already see who initiated the promotion, but in the git commit log itself there is no trace of it.

# Motivation

This would enable to easily see who initiated a git commit from within kargo.

krancour:
I think the most sensible thing here might be to make promoter information available within expressions, so you can dynamically populate the existing author info fields.

creeram:
@Sebast1aan Do you have any updates on when this feature might be implemented?

krancour:
@creeram I believe @Sebast1aan was making a feature request and not _necessarily_ volunteering to implement the feature. This is assigned to myself, but isn't currently regarded as a priority.

creeram:
@krancour  I'm waiting for this feature to be implemented so I can use it in my configuration. This will allow us to get details in Slack about who performed the release and in which environment. I would appreciate it if you could start working on this.

krancour:
@creeram our priorities aren't _dictated_ by your needs. If you need this so urgently, you can take a crack at implementing it. Would you like me to assign it to you?

creeram:
@krancour Thank you for your response. I completely understand that open-source projects have their own priorities. However, I’d like to mention that not all users are developers or have the skills to contribute code. I appreciate the work you do and just wanted to share that a little understanding in responses can go a long way.

keithfz:
@krancour I can take a look into this if no one else is looking at it. I have a similar desire for this.

For implementation, I see that there is already the annotation for `kargoI.akuity.io/create-actor` -- would it suffice / be acceptable to simply propagate that to the well-known expression ctx? I'm a bit of a beginner to this codebase :)

krancour:
This approach won't yield a lot of detail, but sounds like it is probably a good starting point. cc @hiddeco for some further validation.

krancour:
I believe this was resolved by #3592

PR #3592, opened 2025-03-02
feat: Add a creator field in the promotion ctx, so it's usable in expressions

keithfz:
As a prerequisite to adopting Kargo, we want to be able to include information about the user triggering the promotion in our promotion templates. Specifically so the authenticated user triggering a promotion is tracked in the Git history of our GitOps repo. 

To implement this, I followed the same pattern & logic used by the `Created By` column in the frontend where we parse the `kargo.akuity.io/create-actor` annotation. If it doesn't exist, we return `N/A`. This gets added to the promotion context, and can then be exposed in an expression.

https://github.com/akuity/kargo/blob/29e6fb8cc090f61c9be6c3d66dda70a2ac7a9556/ui/src/features/stage/promotions.tsx#L150-L155

For our use case around auditability in the commit history, this suffices.

Relates to issue: https://github.com/akuity/kargo/issues/3081


internal/controller/promotions/promotions.go
```diff
@@ -476,6 +477,12 @@ func (r *reconciler) promote(
 		}
 	}
 
+	creator := "N/A"
+	actorAnnotation, ok := promo.ObjectMeta.Annotations[kargoapi.AnnotationKeyCreateActor]
+	if ok {
+		creator = strings.Split(actorAnnotation, ":")[1]
```
hiddeco, line 483: We should ensure it actually contains two parts, as otherwise this may result in a `index range out of bound` panic.

kzeto-sl: thanks, yeah shouldn't assume it'll always be in the proper form. I've added a check to ensure the len of the split string is as expected.

keithfz: thanks, makes sense, shouldn't assume that. I've added a check to ensure the string list is the expected length.

hiddeco:
Made some changes to facilitate further growth better, and to simplify potential comparison checks people may want to do at some point:

- Renamed and moved the field from `ctx.creator` to `ctx.meta.promotion.actor`.
- Dropped the `N/A` in case none is found, this makes it much simpler to do e.g. `!= ""` type of checks.
- Slightly modified the extraction logic to incorporate a potential `admin` value that might be set.

keithfz:
Awesome! Changes make sense to me. 

Should we also update the docs accordingly for the new path you’ve added?

https://github.com/akuity/kargo/blob/8d6022532cee2175ef9612a2b43baae14312f5e7/docs/docs/50-user-guide/60-reference-docs/40-expressions.md?plain=1#L146

internal/controller/promotions/promotions_test.go
```diff
@@ -462,6 +463,74 @@ func Test_reconciler_terminatePromotion(t *testing.T) {
 	}
 }
 
+func Test_parseCreateActorAnnotation(t *testing.T) {
+	tests := []struct {
+		name            string
```
keithfz, line 468: Lint error here

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -143,7 +143,7 @@ Expect other useful variables to be added in the future.
 
 | Name | Type | Description |
 |------|------|-------------|
-| `ctx` | `object` | `string` fields `project`, `stage`, and `promotion` provide convenient access to details of a `Promotion`. |
+| `ctx` | `object` | `string` fields `project`, `stage`, `promotion`, and `creator` provide convenient access to details of a `Promotion`. |
```
keithfz, line 146: Should update this to new field?

hiddeco: Yes, this still needs to be updated, but I am struggling to find the best way to address this as we seem to be running into the limitations of using tables.

hiddeco (approved):
Thank you very much @keithfz 🌷

Issue #3552, opened 2025-02-24
Google Artifact Registry credentials should be more flexible


kacpercesarz98:
# Checklist

* [X] I've searched the issue queue to verify this is not a duplicate feature request.
* [X] I've pasted the output of `kargo version`, if applicable.
* [X] I've pasted logs, if applicable.

# Proposed Feature

Improve the flexibility of Google Artifact Registry credentials service account impersonation.

# Motivation

Currently, the hardcoded identity format:
`kargo-project-<kargo project name>@<gcp project name>.iam.gserviceaccount.com` leads to a suboptimal administrator experience.
In scenarios where tens or hundreds of projects are required, administrators must create numerous Service Accounts and IAM permissions, even when all of them ultimately point to the same Artifact Repository. This results in unnecessary complexity and overhead.

# Suggested Implementation

Instead of enforcing a strict identity format, allow experienced users to manage their IAM configurations more flexibly. One possible approach is to enable mapping a Service Account to an Artifact Registry instance rather than binding it directly to a Kargo Project. Other way is to add some field/annotation to the Repository secret with the name of service account which should be impersonated.

krancour:
For AWS/ECR, we have long had a feature where if the project-specific ambient credentials could not be assumed or were assumed and have insufficient permissions, Kargo will fall back on using the ambient credentials of the controller itself.

We don't have this for Google/GAR yet, but I've been meaning to get it caught up to AWS/ECR in this regard.

Without necessarily closing the door to other approaches in the future, would you find this relatively simple improvement helpful?

kacpercesarz98:
> For AWS/ECR, we have long had a feature where if the project-specific ambient credentials could not be assumed or were assumed and have insufficient permissions, Kargo will fall back on using the ambient credentials of the controller itself.
> 
> We don't have this for Google/GAR yet, but I've been meaning to get it caught up to AWS/ECR in this regard.
> 
> Without necessarily closing the door to other approaches in the future, would you find this relatively simple improvement helpful?


Yes it will be definitely helpful and more convenient. If it will not close doors for future improvements it is good enough for now.

kacpercesarz98:
Another issue which I found is lenght limit of GCP service account name:
`The account ID "kargo-project-*" does not have a length between 6 and 30., badRequest`
It force me to create project with not clear name.

remi-gelinas:
[...]

PR #3597, opened 2025-03-03
Add simple fallback to kargo controller IAM.

kacpercesarz98:
Add simple fallback to kargo controller IAM when project-specific credentials could not be assumed for Google Artifact Registry
closes: [issue](https://github.com/akuity/kargo/issues/3552)


kacpercesarz98:
With this approach there is needed to fix or disable caching as using `google.DefaultTokenSource` do not always create fresh token. There are time windows when cached token is already expired.

internal/credentials/kubernetes/gar/workload_identity_federation.go
```diff
@@ -105,32 +111,48 @@ func (w *workloadIdentityFederationCredentialHelper) getCredentials(
...
-			Scope: []string{
-				"https://www.googleapis.com/auth/cloud-platform",
-			},
+			Scope: tokenRequestScope,
 		},
 	).Do()
 	if err != nil {
-		logger.Error(err, "error generating access token")
-		return "", nil
+		logger.Error(err, "Error generating access token; falling back to Application Default Credentials (ADC)")
```
krancour, line 138: I think there needs to be some effort here to determine whether the failure is an auth failure or something else. If it's (for instance) a connectivity issue, we should be returning that error right away.

If it _is_ an auth failure, I also don't believe this should be logged as an error. In your own intended usage, this first token request failing and then falling back on this logic that attempts to use the controller's own identity directly is actually business as usual. You're not going to want to see the logs cluttered with "errors" that reflect things working as intended.

internal/credentials/kubernetes/gar/workload_identity_federation.go
```diff
@@ -105,32 +111,48 @@ func (w *workloadIdentityFederationCredentialHelper) getCredentials(
...
+		if w.tokenSource == nil {
+			tokenSource, err := google.DefaultTokenSource(ctx, tokenRequestScope...)
+			if err != nil {
+				logger.Error(err, "Failed to find Application Default Credentials")
+				return "", false, nil
+			}
+			w.tokenSource = tokenSource
+		}
```
krancour, line 146: This is not concurrency safe.

kacpercesarz98:
> > With this approach there is needed to fix or disable caching as using google.DefaultTokenSource do not always create fresh token. There are time windows when cached token is already expired.
> 
> This warrants some research, because I wonder if this is actually true or if this observation is actually an artifact of this concurrency safety issue:
> 
> https://github.com/akuity/kargo/pull/3597/files#r1979750397

Please verify:
https://github.com/golang/oauth2/blob/master/oauth2.go#L371

krancour:
> Please verify:
> https://github.com/golang/oauth2/blob/master/oauth2.go#L371

I agree that a single `TokenSource`, if its implementation is `reuseTokenSource`, has its own internal cache and will return the same token a number of times.

And through research and experimentation, I can verify that the `TokenSource` in question here is indeed a `reuseTokenSource`.

This can be demonstrated with a simple program:

```golang
package main

import (
	"context"
	"log"
	"os"

	"golang.org/x/oauth2/google"
)

const (
	credsPath         = "/Users/krancour/Downloads/kents-test-project-317520-e3fb3e962f2d.json" // nolint: gosec
	tokenRequestScope = "https://www.googleapis.com/auth/cloud-platform"                        // nolint: gosec
)

func init() {
	if err := os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", credsPath); err != nil {
		log.Fatal(err)
	}
}

func main() {
	ctx := context.Background()
	tokenSource, err := google.DefaultTokenSource(ctx, tokenRequestScope)
	if err != nil {
		log.Fatal(err)
	}
	token1, err := tokenSource.Token()
	if err != nil {
		log.Fatal(err)
	}
	token2, err := tokenSource.Token()
	if err != nil {
		log.Fatal(err)
	}
	if token1.AccessToken == token2.AccessToken {
		log.Println("Tokens are the same")
	} else {
		log.Println("Tokens are different")
	}
}
```

```shell
go run main.go
2025/03/04 12:16:11 Tokens are the same
```

But two different instances of `reuseTokenSource` would return two different tokens, as demonstrated with some small tweaks to the above program:

```
func main() {
	ctx := context.Background()
	tokenSource1, err := google.DefaultTokenSource(ctx, tokenRequestScope)
	if err != nil {
		log.Fatal(err)
	}
	tokenSource2, err := google.DefaultTokenSource(ctx, tokenRequestScope)
	if err != nil {
		log.Fatal(err)
	}
	token1, err := tokenSource1.Token()
	if err != nil {
		log.Fatal(err)
	}
	token2, err := tokenSource2.Token()
	if err != nil {
		log.Fatal(err)
	}
	if token1.AccessToken == token2.AccessToken {
		log.Println("Tokens are the same")
		fmt.Println(token1.AccessToken)
	} else {
		log.Println("Tokens are different")
	}
}
```

```shell
go run main.go
2025/03/04 12:18:05 Tokens are different
```

Your code is currently falling into the first scenario because of the way you're re-using a single `TokenSource` for all requests, however, you haven't written that in a concurrency safe manner. Admittedly, the lack of concurrency safety is a bit academic here, because the race to set the one-reuses `TokenSource` won't have any adverse effects (beyond the side-effect of it invalidating this component's existing caching strategy).

If you were to tackle the concurrency safety issue, academic though it may be, by not reusing a single `TokenSource`, your code would more closely resemble the second example, but more importantly, it would have the side effect of _preserving_ this component's existing caching strategy.

You _could_ take a different approach to solving the concurrency safety issue, or we could just leave it alone, since I acknowledge that it's mainly academic in this instance. In either case, you'd be reusing the same `TokenSource` over and over, which does seem efficient since it has its own built-in cache, but again, we come back to the fact of this breaking the existing caching strategy...

So the thing I am asking myself is why I actually care about preserving the existing caching strategy. I like that the existing strategy is consistent for both controller-level and project-level tokens, which makes this code, overall, easier to comprehend and maintain.

kacpercesarz98:
@krancour I believe I addressed your comments. Could You please verify? If You want to preserve current cache logic let me know and I will refactor this code.

krancour:
> If You want to preserve current cache logic let me know and I will refactor this code.

Yes. It would be my preference that we preserve the original caching strategy by not reusing a single token source.

kacpercesarz98:
@krancour I did some experiment, looks like my first assumption that DefaultTokenSource always return same token if it is still not expired was correct. What You showed in Your test program was working in different way because You was running this program on Your PC. I prepared test environment. And run bellow code in 2 scenarios:
`package main

import (
	"context"
	"fmt"
	"log"

	"golang.org/x/oauth2"
	"golang.org/x/oauth2/google"
)

const (
	tokenRequestScope = "https://www.googleapis.com/auth/cloud-platform" // nolint: gosec
)

func fallbackToControllerIdentity(ctx context.Context) (*oauth2.Token, error) {
	fmt.Println("falling back to Application Default Credentials (ADC)")
	ts, err := google.DefaultTokenSource(ctx)
	if err != nil {
		fmt.Println(err, "Error generating Application Default Credentials")
		return nil, nil
	}
	token, err := ts.Token()
	if err != nil {
		fmt.Println(err, "Error generating access token from Application Default Credentials")
		return nil, nil
	}
	fmt.Println("Generated access token using Application Default Credentials")
	return token, nil
}

func main() {
	ctx := context.Background()
	tokenSource1, err := google.DefaultTokenSource(ctx, tokenRequestScope)
	tokenSource2, err := google.DefaultTokenSource(ctx, tokenRequestScope)
	token1, err := tokenSource1.Token()

	if err != nil {
		log.Fatal(err)
	}

	token2, err := tokenSource2.Token()
	if err != nil {
		log.Fatal(err)
	}
	if token1.AccessToken == token2.AccessToken {
		log.Println("Tokens are the same")
	} else {
		log.Println("Tokens are different")
	}
}
`
1. Scenario run on my local machine:
kacpercesarz@LPM240101K kargo-test % go run main.go
2025/04/11 09:02:14 Tokens are different
2. Run in Pod in GKE cluster with metadata server:
root@pod:/go/test# go run main.go 
2025/04/11 07:01:20 Tokens are the same
Looks like behaviour is different because Application Default Credentials works different in both scenarios. Looks like metadata server is caching token until it is expired. Im looking for some clean scenario for this problem.

kacpercesarz98:
Updated code, as there is no way to force refresh token when using DefaultTokenSource provided solution in my opinion could be good quick win as we preserve cache logic and use consistent style of fetching token. Downside of this is that extra ServiceAccount need to be created(instead of using workloadIdentityPool directly). Expected ServiceAccount name is:
`gcpFallbackResourceNameFormat = "projects/-/serviceAccounts/kargo-controller@%s.iam.gserviceaccount.com"`

kacpercesarz98:
> I'm going to amend your PR with a commit that attempts what I described above. Perhaps you could review that and test drive it? And it can certainly be undone if we end up finding an even better option.

Of course yes, I will be happy to test this.

krancour:
Ok. I had a short debate with myself over whether to use:

1. Just one cache that might contain _both_ short-lived tokens with short TTLs _and_ long-lived token sources with long TTLs

2. A separate cache for each of those.

I ended up opting for the second one.

Unit tests are updated accordingly and passing, but I don't have a GKE cluster with workload identity enabled handy.

@kacpercesarz98 please lmk how it works out in your cluster. 🙏

kacpercesarz98:
So far all looks good, works as expected.

krancour:
Thanks @kacpercesarz98. I won't rush to merge this, but unless you run into some problem with it, this will definitely make it into v1.5.0.

Thanks again for all the research, effort, and testing you've put into this.

krancour (approved):

Issue #3293, opened 2025-01-15
Move generated promo step type definitions to pkg 


jessesuen:
# Proposed Feature

Currently the golang versions of native promotion steps are auto generated to the `internal/directives` directory:
https://github.com/akuity/kargo/blob/main/internal/directives/zz_config_types.go

For example:

```golang
type GitCloneConfig struct {
	// The commits, branches, or tags to check out from the repository and the paths where they
	// should be checked out. At least one must be specified.
	Checkout []Checkout `json:"checkout"`
	// Indicates whether to skip TLS verification when cloning the repository. Default is false.
	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
	// The URL of a remote Git repository to clone. Required.
	RepoURL string `json:"repoURL"`
}
```

We should move these outside of `internal` into `pkg` so they can be publically consumed.

# Motivation

Anyone who is programmatically construct a Kargo object from its CRD types, or the reverse: unmarshalling them into an golang object, do not have access to these types since they are in the `internal` directory.

# Suggested Implementation

Change the path where these are generated and the corresponding import statements.

krancour:
Re-assigning to myself because this aligns closely with other on-going cleanup chores I am already engaged in.

krancour:
This also fits more neatly with work we are planning for 1.4. Moving to that milestone.

PR #3609, opened 2025-03-04
chore: move generated step config types to pkg/x

krancour:
Related to #3293, although I don't consider that issue closed until this package is consumable as its own module that doesn't drag along all other Kargo dependencies.

We will still get that done in the v1.4.0 timeframe, but we're putting it off for just a bit because it breaks some of our codegen and I don't want to make this PR bigger than it already is.


internal/directives/argocd_revisions.go
```diff
@@ -2,6 +2,7 @@ package directives
 
 import (
 	argocd "github.com/akuity/kargo/internal/controller/argocd/api/v1alpha1"
+	"github.com/akuity/kargo/pkg/x/directives/builtins"
```
hiddeco, line 5: We may want to take this opportunity to rename the package to singular (`.../directive/builtin`). As we (probably _I_, actually) appear to have missed the boat there to adhere to community conventions.

krancour: Consider it done

krancour: All set.

Issue #3582, opened 2025-02-27
GHE base url not being used


han-steve:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

I'm trying to hook up Kargo with a Github Enterprise repo. However, after adding the app token, I'm seeing this error on the warehouse page: 
```
Unable to discover artifacts: error discovering commits: 
error obtaining credentials for git repo "https://github.rbx.com/org/deployment": 
error getting installation access token: 
error getting installation access token:
 POST https://api.github.com/app/installations/515/access_tokens: 
401 A JSON web token could not be decoded []
```

Supposedly, this should've been fixed by https://github.com/akuity/kargo/issues/1355, however, in the message above, it's clear that it's trying to post to api.github.com when my repo is at github.rbx.com. The correct url is api.github.rbx.com. 

# Version

This is the Kargo image I'm running: 
```
ghcr.io/akuity/kargo:v1.3.0 
```

# Logs

```
level=error msg="Reconciler error" Warehouse="{ray-infra ray-infra}" controller=wa
rehouse controllerGroup=kargo.akuity.io controllerKind=Warehouse error="error discovering artifacts: error dis
covering commits: error obtaining credentials for git repo \"https://github.rbx.com/Roblox/ml-platform-deploym
ent\": error getting installation access token: error getting installation access token: POST https://api.gith
ub.com/app/installations/515/access_tokens: 401 A JSON web token could not be decoded []" name=ray-infra names
pace=ray-infra reconcileID="\"5c208fc9-1007-42c2-85a1-90178c05b092\""
```

hiddeco:
I believe this _somehow_ may have been solved for actions on pull requests, but not for clone operations. As we do not appear to do anything here to detect or configure enterprise URLs: https://github.com/akuity/kargo/blob/413063fe13d3fc1199bb4a34df9b3ee40a9a536a/internal/credentials/kubernetes/github/app.go#L147-L155

han-steve:
Thanks that's a good catch!

PR #3617, opened 2025-03-06
refactor(credentials): replace `Helper` type with `Provider` interface

hiddeco:
Fixes: #3582

This PR refactors the credential database by replacing the function-based `Helper` type with a more robust `Provider` interface. The work began as a way to support GitHub Enterprise instances, but ended up addressing some deeper design issues along the way.

In short:

- Credential support detection is now separated from credential generation, to primarily improve testing and ensure single responsibility.
- Ensures provider constructors new explicitly return `nil` when they can't function properly, rather than returning a partially initialized instance that would be attempted each time.
- Providers now work with raw data maps instead of Kubernetes Secrets, ensuring Kubernetes does not leak directly into the higher-level credentials abstraction.
- Support for GitHub Enterprise instances has been added based on the repository URL.

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -2,8 +2,6 @@ package credentials
 
 import (
 	"context"
-
-	corev1 "k8s.io/api/core/v1"
 )
 
 const (
@@ -43,10 +41,22 @@ type Credentials struct {
 	SSHPrivateKey string
 }
 
-type Helper func(
-	ctx context.Context,
-	project string,
-	credType Type,
-	repoURL string,
-	secret *corev1.Secret,
-) (*Credentials, error)
+// Provider is an interface for providing credentials for a given type,
+// repository URL and data values.
+type Provider interface {
+	// Supports returns true if the provider can provide credentials for the
+	// given type, repository URL and data values. Otherwise, it should return
+	// false.
+	Supports(credType Type, repoURL string, data map[string][]byte) bool
+
+	// GetCredentials returns the credentials for the given type, repository URL
+	// and data values. If the provider cannot provide credentials for the given
+	// type, repository URL and data, it should return nil.
+	GetCredentials(
+		ctx context.Context,
+		project string,
+		credType Type,
+		repoURL string,
+		data map[string][]byte,
+	) (*Credentials, error)
+}
--- a/internal/credentials/kubernetes/database.go
+++ b/internal/credentials/kubernetes/database.go
@@ -81,7 +81,7 @@ func (k *database) Get(
 	namespace string,
 	credType credentials.Type,
 	repoURL string,
-) (credentials.Credentials, bool, error) {
+) (*credentials.Credentials, error) {
 	// If we are dealing with an insecure HTTP endpoint (of any type),
 	// refuse to return any credentials
 	if !k.cfg.AllowCredentialsOverHTTP && strings.HasPrefix(repoURL, "http://") {
@@ -89,7 +89,7 @@ func (k *database) Get(
 			"refused to get credentials for insecure HTTP endpoint",
 			"repoURL", repoURL,
 		)
-		return credentials.Credentials{}, false, nil
+		return nil, nil
 	}
 
 	var secret *corev1.Secret
@@ -102,7 +102,7 @@ func (k *database) Get(
 		credType,
 		repoURL,
 	); err != nil {
-		return credentials.Credentials{}, false, err
+		return nil, err
 	}
 
 	if secret == nil {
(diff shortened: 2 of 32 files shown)


internal/credentials/credentials.go
```diff
@@ -43,10 +41,22 @@ type Credentials struct {
...
-	repoURL string,
-	secret *corev1.Secret,
-) (*Credentials, error)
+// Provider is an interface for providing credentials for a given type,
+// repository URL and data values.
+type Provider interface {
+	// Supports returns true if the provider can provide credentials for the
+	// given type, repository URL and data values. Otherwise, it should return
+	// false.
+	Supports(credType Type, repoURL string, data map[string][]byte) bool
```
krancour, line 50: All the providers seem to invoke `Supports()` themselves and I don't see anywhere where it's invoked from elsewhere. I'm in agreement with the notion of separating the logic for whether a provider is supported from the actually credential retrieval logic, but I am wondering if this isn't more of an implementation detail that doesn't need to be exposed as part of the `Provider` interface.

internal/credentials/kubernetes/ecr/access_key.go
```diff
@@ -23,118 +21,101 @@ const (
...
-		secretAccessKey string,
-	) (string, error)
+	getAuthTokenFn func(ctx context.Context, region, accessKeyID, secretAccessKey string) (string, error)
 }
 
-// NewAccessKeyCredentialHelper returns an implementation of credentials.Helper
-// that utilizes a cache to avoid unnecessary calls to AWS.
-func NewAccessKeyCredentialHelper() credentials.Helper {
-	a := &accessKeyCredentialHelper{
+func NewAccessKeyProvider() *AccessKeyProvider {
```
krancour, line 30: Applies for all providers... would it be more conventional to return `credentials.Provider` here? If we did so, individual provider implementations also need not be exported.

internal/credentials/kubernetes/database.go
```diff
@@ -53,25 +53,27 @@ func NewDatabase(
...
-	return &database{
-		kargoClient:       kargoClient,
-		credentialHelpers: finalCredentialHelpers,
-		cfg:               cfg,
+
+	for _, p := range credentialProviders {
+		if p != nil {
+			db.credentialProviders = append(db.credentialProviders, p)
+		}
 	}
```
krancour, line 74: Why loop to do this instead of setting them directly around line 68?

hiddeco: Because a constructor may return `nil`. Note that this is no different from the previous implementation, except that the constructors now do return a `nil` value.

internal/credentials/credentials.go
hiddeco: The reason I exported it is because I can think of scenarios in which you may want to put together a set of providers by testing their support without actually wanting to retrieve credentials at that moment.

internal/credentials/kubernetes/ecr/access_key.go
krancour: :thinking: I always had interpreted that as `return <concrete type>` and not so much that the return type in the signature should be concrete. I do think it produces cleaner abstractions where callers don't need to concern themselves _at all_ with the underlying implementation.

krancour:
This was way overdue for some cleanup and broadly speaking, I agree with the changes.

I raised a few questions, but overall looks good.

One thing I would add is that while we're doing this, we might consider changing the signature of the DB itnerface's `Get()` method from:

```golang
Get(ctx context.Context, namespace string, credType credentials.Type, repoURL string (credentials.Credentials, bool, error)
```

which I have always somewhat regretted, to:

```golang
Get(ctx context.Context, namespace string, credType credentials.Type, repoURL string (*credentials.Credentials, error)
```

internal/credentials/credentials.go
krancour: sgtm

internal/credentials/kubernetes/ecr/access_key.go
krancour: > it should be noted that this is not a hard and fast rule. There are times when it makes sense to accept concrete types and return interfaces, or to accept interfaces and return concrete types.

I won't die on this hill, but only want to establish that I'm not nuts.

> If we did so, individual provider implementations also need not be exported.

If we _start_ there with "callers have no need to be aware of the underlying implementation, so I'm not exporting it," then returning an interface makes sense.

However, if you have defensible reasons for the implementations to be exported, then my scenario does not apply.

Either way, not a blocker.

krancour (approved):

PR #3622, opened 2025-03-07
chore: move rollouts api stubs

krancour:
This chore moves the stubbed out Rollouts API types that the service API depends on out of `internal`.

I'll be moving the service API out of `pkg` and into `api` in a follow-up.

The intention is to aggregate useful APIs in one module that doesn't impose many transitive dependencies on those who use it.


krancour:
I have some UI and proto linting issues to address, but I think the rest of this remains reviewable.

krancour:
Linting issues are fixed.

hiddeco (approved):

hiddeco:
Do wonder why the linting exclusion wasn't required before.

krancour:
> Do wonder why the linting exclusion wasn't required before.

It's because linting is enabled for `api/`, but not for `internal/controller/`. It appears it was only ever enabled for `api/` to lint `service.proto`, which is our only hand-written `.proto` file.

```yaml
- path: api
    name: buf.build/akuity/kargo
    lint:
      use:
        - STANDARD
      except:
        - FIELD_NOT_REQUIRED
        - PACKAGE_DIRECTORY_MATCH
        - PACKAGE_NO_IMPORT_CYCLE
      ignore:
        - api/rbac/v1alpha1/generated.proto
        - api/stubs/rollouts/v1alpha1/generated.proto
        - api/v1alpha1/generated.proto
    breaking:
      use:
        - FILE
      except:
        - EXTENSION_NO_DELETE
        - FIELD_SAME_DEFAULT
  - path: internal/controller
    lint:
      disallow_comment_ignores: true
```

PR #3634, opened 2025-03-10
chore: move service api

krancour:
This PR moves the `service` package as part of my broader effort to consolidate APIs in `api/`, which will eventually become its own module.

This checks out e2e.

I broken protobu-related codegen (again) when doing this and finally opted to drop in the replacement process I've been sitting on for about a month.


hack/codegen/proto.sh
```diff
@@ -20,91 +20,126 @@ readonly APIMACHINERY_PKGS=(
...
-
 function main() {
-  echo "Change working directory to temporary build directory..."
-  cd "${build_dir}"
-
-  echo "Prepare build directory..."
-  # Initialize dummy module inside build directory
+  set -x
+
+  # go-to-protobuf is used for generating .proto and .pb.go files from
```
hiddeco, line 26: Doing God's work here with the doc blocks.

krancour:
I have linter errors in the UI. I need to reorder imports so they're alphabetical by path.

hiddeco (approved):
Tested this locally and confirmed the codegen changes to work, and the further structure to look like I expected. Assuming the linter is made happy again, this looks good to me.

PR #3635, opened 2025-03-10
api ingress path match any

johnpekcan:
The `/` path seems too restrictive. Can't load dependencies, `/login` etc.

image
image


johnpekcan:
`Prefix` instead of `ImplementationSpecific` does seem to alleviate this issue

PR #3644, opened 2025-03-13
refactor: untangle the internal/directives package

krancour:
This PR untangles the Gordian Knot that is has been `internal/directives` package.

Before:

```
internal/
├── ...
├── directives/
│   ├── schemas/
│   │    └── schemas, obviously
│   └── LOTS AND LOTS OF STUFF
└── ...
```

After:

```
internal/
├── ...
├── health/
│   ├── checkers/
│   │   └── builtin/
│   │       └── specific, built-in health checkers
│   └── generic logic for checking health
├── ...
├── promotion/
│   ├── runners/
│   │   └── builtin/
│   │       ├── schemas/
│   │       │   └── schemas, obviously
│   │       └── specific, built-in promotion step runners
│   └── generic logic for executing a promotion
└── ...

```

Many things are also renamed as we discussed offline @hiddeco.

I am confident you may still have some good naming improvements to suggest.

Notes:

* Apart from how `promotion.Runner`s and `health.Checker`s are registered, no other logic has changed. Everything checks out e2e.

* ~~I don't entirely love `health.Engine` and will probably rename it to something like `health.ComprehensiveChecker` first thing in the morning, but this is otherwise r4r.~~ 

  ~~Renamed to `health.MultiChecker`~~

  Renamed to `health.AggregatingChecker`


hiddeco (commented):
I do not think the nesting under `controller/` is logical to me, as I would only expect logic there that directly has to do with the lifecycle of Kubernetes custom resources — which isn't the case here.

internal/controller/health/checkers/builtin/init.go
```diff
@@ -0,0 +1,13 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/akuity/kargo/internal/controller/health"
+)
+
+// Initialize registers all built-in health.Checkers with the health package's
+// internal Checker registry.
+func Initialize(argocdClient client.Client) {
+	health.RegisterChecker(newArgocdChecker(argocdClient))
+}
```
hiddeco, line 13: Historically, I have been bitten more than once by global registrations like this as they tend to backfire in a (unexpected) way at some point in time. This is also why, before, registration was detached from dependency injection. It allowed global registration, but without the assumption that all callers have the same dependencies (they could operate with different credentials, or on different clusters, or ...).

I would feel more comfortable if this could continue to be the case, either by injecting the dependencies right before they are going to be used, or by dropping the global register and returning a slice of builtins here to be used at construction time of the engine.

If that's not an option, a last resort would be to wrap it in a `sync.Once` so that it should become obvious very quickly something is wrong.

internal/controller/health/checkers/builtin/argocd.go
```diff
@@ -1,4 +1,4 @@
-package directives
+package builtin
```
hiddeco, line 1: Think the path should be `.../health/checker/builtin/...`.

internal/controller/health/checkers/builtin/argocd.go
```diff
@@ -54,46 +55,66 @@ type compositeError interface {
 	Unwrap() []error
 }
 
-// RunHealthCheckStep implements the Directive interface.
-func (a *argocdUpdater) RunHealthCheckStep(
+type argocdChecker struct {
```
hiddeco, line 58: Arguably, the `Checker` suffix has become redundant.

internal/controller/health/health.go
```diff
@@ -0,0 +1,92 @@
...
+	// Status is the high-level outcome of the HealthCheck executed by a
+	// Checker.
+	Status kargoapi.HealthState
+	// Output is the opaque output of a HealthCheck executed by a
+	// Checker. The Engine will aggregate this output and include it
+	// in the final results of the health check, which will ultimately be included
+	// in StageStatus.
+	Output map[string]any
+	// Issues is a list of issues that were encountered during the execution of
+	// the HealthCheck by a Checker.
```
hiddeco, line 48:
```suggestion
	// the health check by a Checker.
```

internal/controller/health/registry.go
```diff
@@ -0,0 +1,23 @@
...
+// register adds a Checker to the checkerRegistry.
+func (c checkerRegistry) register(checker Checker) {
+	c[checker.Name()] = checker
+}
+
+// getChecker returns the Checker for the health check with the given name. If
+// no Checker is registered with the given name, nil is returned instead.
+func (c checkerRegistry) getChecker(name string) Checker {
+	return c[name]
+}
```
hiddeco, line 23: Order of things is a bit odd, probably easier to read like this:

```suggestion
package health

// checkerRegistry is a registry of Checkers.
type checkerRegistry map[string]Checker

// checkerReg is a registry of Checkers.
var checkerReg = checkerRegistry{}

// register adds a Checker to the checkerRegistry.
func (c checkerRegistry) register(checker Checker) {
	c[checker.Name()] = checker
}

// getChecker returns the Checker for the health check with the given name. If
// no Checker is registered with the given name, nil is returned instead.
func (c checkerRegistry) getChecker(name string) Checker {
	return c[name]
}

// RegisterChecker adds a Checker to the package's internal registry.
func RegisterChecker(checker Checker) {
	checkerReg.register(checker)
}
```

internal/controller/promotion/registry.go
```diff
@@ -0,0 +1,24 @@
+package promotion
```
hiddeco, line 1: Same comment as for the `.../health/registry.go`.

internal/controller/promotion/runners/builtin/init.go
```diff
@@ -0,0 +1,37 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/akuity/kargo/internal/controller/promotion"
+	"github.com/akuity/kargo/internal/credentials"
+)
+
+// Initialize registers all built-in promotion.StepRunners with the promotion
+// package's internal StepRunner registry.
+func Initialize(kargoClient, argocdClient client.Client, credsDB credentials.Database) {
+	builtIns := []promotion.StepRunner{
```
hiddeco, line 13: Same concern around globals here as commented on the health registry.

internal/controller/promotion/promotion.go
```diff
@@ -11,27 +11,26 @@ import (
...
+// execution of an individual Step in a user-defined promotion process.
+type StepRunner interface {
+	// Name returns the name of the StepRunner.
 	Name() string
-	// RunPromotionStep executes an individual step of a user-defined promotion
-	// process using the provided PromotionStepContext. Implementations may
-	// indirectly modify that context through the returned PromotionStepResult to
-	// allow subsequent promotion steps to access the results of its execution.
-	RunPromotionStep(context.Context, *PromotionStepContext) (PromotionStepResult, error)
+	// Run executes an individual Step from a user-defined promotion process using
```
hiddeco, line 25: Related to our discussions offline:

> Run executes [...]

makes me think I may not have been too far off with `StepExecutor`.

internal/controller/health/checkers/builtin/init.go
krancour: > This is also why, before, registration was detached from dependency injection.

> I would feel more comfortable if this could continue to be the case, either by injecting the dependencies right before they are going to be used

I see the benefits of the original pattern and actually like it quite a bit, however, the inability to inject dependencies like the creds database or k8s clients _up front_, did mean that, precisely as you said, those dependencies needed to be provided just-in-time. Our method for doing so had been injecting them, as necessary, into what was then called `PromotionStepContext`.

Undoing _that_ was actually the catalyst for touching the `directives` package in the first place, as fields of a type that live under neither `api/` nor `pkg/` were impairing _other_ refactoring efforts -- namely it complicated relocating `PromotionStepContext` (among a few other key types) to `pkg/...` since `pkg` is _also_ intended to become its own module _without dependencies on Kargo's main module_.

Because I really did like the self-registration we had before, the first thing I actually tried here was injecting things like k8s clients and the creds database (which is _the_ actual problem, btw), just-in-time, as needed, into an actual `context.Context` instead of a `PromotionStepContext`. It worked, but felt too sneaky and made the code harder to follow. This is how I landed where I did instead.

> or by dropping the global register and returning a slice of builtins here to be used at construction time of the engine

That sounds reasonable, but I have one concern about it that I will PM you about.

internal/controller/health/checkers/builtin/argocd.go
krancour: I realized even when I did it that Go preferred the singular here. Strictly in terms of directory structure, however, the singular `checker/` seemed too odd a name for a folder that is clearly a bucket for multiples. I got hung up on that, and it led to this. I don't feel strongly about it. If you think deference to Go conventions trumps my concerns over the clarity of the directory structure, I am happy to change this.

internal/controller/health/checkers/builtin/argocd.go
krancour: I knew there was an argument to be made for that since `checkers/` (which may become `checker/`) is in the import path.

If the package name had been `checker`, I'd have considered `checker.argocd` clear enough, but with the package name being `builtin`, I think `builtin.argocdChecker` is a bit more clear about what it is than `builtin.argocd` is.

I realize this is a bit academic because that type isn't even exported, so you're never going to see `builtin.argocdChecker` anywhere in the code. 🤷‍♂️ 

I'm willing to change it if you feel strongly.

internal/controller/health/registry.go
krancour, line 23: Agree it was odd. I would prefer this very slight alteration to your own suggestion as it keeps `checkerRegistry`'s methods defined immediately after the type def instead of putting the variable declaration in between.

```suggestion
package health

// checkerRegistry is a registry of Checkers.
type checkerRegistry map[string]Checker

// register adds a Checker to the checkerRegistry.
func (c checkerRegistry) register(checker Checker) {
	c[checker.Name()] = checker
}

// getChecker returns the Checker for the health check with the given name. If
// no Checker is registered with the given name, nil is returned instead.
func (c checkerRegistry) getChecker(name string) Checker {
	return c[name]
}

// checkerReg is a registry of Checkers.
var checkerReg = checkerRegistry{}

// RegisterChecker adds a Checker to the package's internal registry.
func RegisterChecker(checker Checker) {
	checkerReg.register(checker)
}
```

krancour: Agree it was odd. I would prefer this very slight alteration to your own suggestion as it keeps `checkerRegistry`'s methods defined immediately after the type def instead of putting the variable declaration in between.

```suggestion
package health

// checkerRegistry is a registry of Checkers.
type checkerRegistry map[string]Checker

// register adds a Checker to the checkerRegistry.
func (c checkerRegistry) register(checker Checker) {
	c[checker.Name()] = checker
}

// getChecker returns the Checker for the health check with the given name. If
// no Checker is registered with the given name, nil is returned instead.
func (c checkerRegistry) getChecker(name string) Checker {
	return c[name]
}

// checkerReg is a registry of Checkers.
var checkerReg = checkerRegistry{}

// RegisterChecker adds a Checker to the package's internal registry.
func RegisterChecker(checker Checker) {
	checkerReg.register(checker)
}
```

krancour:
> I do not think the nesting under controller/ is logical to me, as I would only expect logic there that directly has to do with the lifecycle of Kubernetes custom resources — which isn't the case here.

With the exception of a single regex that is exported from here and used by a webhook, the code in `directives` is exclusively used by the controller and I'm not sure I could see it being used in other contexts. For me that was strong impetus to move it under `controller/`. It just felt slightly more organized this way. If you think there is an argument to be made for this later being used in other contexts, I'm more than happy to move this out from under `controller/`.

hiddeco:
> With the exception of a single regex that is exported from here and used by a webhook, the code in directives is exclusively used by the controller [...]

It being used or not being used by the controller is not my point, as we have many bits in `internal/` that are probably _only_ used by `internal/controller`. My point is that anything that has to do with the actual execution of a promotion is a separate concern than dealing with the `Promotion` object (which is part of the `controller/` logic), by nesting it, you are creating a ramp for entanglement.

internal/controller/health/checkers/builtin/init.go
krancour: Discussed offline and addressed in 329eb50ef3f114679ab47ba72a2aa54b3083915b

krancour:
`internal/controller/{health,promotion}` --> `internal/{health,promotion}` in a701120c69d2c83a584392089d4e4ff1b548eab9

internal/controller/health/checkers/builtin/argocd.go
krancour: 6aed278ece6af78e0ad5bf1be5d7037f07cf0115

internal/health/health.go
```diff
@@ -9,6 +9,15 @@ import (
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 )
 
+// MultiChecker is an interface for components that can execute a comprehensive
+// assessment of Stage health by executing a sequence of health checks and
+// aggregating their results.
+type MultiChecker interface {
```
hiddeco, line 15: `AggregatedChecker` maybe?

hiddeco, line 15: `AggregatedChecker` maybe?

internal/health/health.go
```diff
@@ -9,6 +9,15 @@ import (
...
+	// Check executes the specified sequence of health checks and returns a
+	// kargoapi.Health that aggregates their results.
+	Check(ctx context.Context, project, stage string, criteria []Criteria) kargoapi.Health
```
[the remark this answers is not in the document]
krancour: Ya... that's a weird one.

`StepRunner` implementations (just `argocdUpdater`, at the moment) return the `health.Criteria` for any checks(s) that should be part of the comprehensive assessment of Stage health. It does not populate those fields because, viewed in the context of a `Stage`, they would be pretty redundant.

A `health.Checker` implementation might want this information.

I was passing them as args here and leaving it up to the `health.MultiChecker` implementation to hydrate the `Project` and `Stage` fields in each `Criteria` before passing it on to its respective `Checker`.

This seemed more appropriate than putting the onus of hydrating those fields on the caller of the `MultiChecker`.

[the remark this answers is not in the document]
krancour: Alternatively, I could remove those fields from the `Criteria` and pass them as args all the way down.

internal/health/health.go
krancour: Sure.

internal/health/health.go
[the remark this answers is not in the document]
krancour: That's what I'm going to do.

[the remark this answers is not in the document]
krancour: a2f94e42ee59fc6b86cf69f7d8c6e6e385bba2f4

internal/health/health.go
krancour: Went with `AggregatingChecker`.

2d82e53dc7543ea730632af5e8569c71aa302fab

internal/controller/health/checkers/builtin/argocd.go
krancour: Going to leave this one be.

krancour:
Note to self make doubly sure changes from https://github.com/akuity/kargo/pull/3642 aren't lost in conflict resolution.

hiddeco (approved):
🔥🚀

Issue #3641, opened 2025-03-12
display logs for verifications/analyses that take the form of a job


krancour:
We're not going to get into the business of forwarding and storing logs, but @jessesuen has proposed that we allow operators to specify a URL template that can produce URLs as a function of stage, promotion name, freight name, analysis run name, etc., along with auth headers.

This can be used to enable a new API server endpoint to act as a proxy for logs that are accessible via an https get request.

The UI, and possibly the CLI, can make use of the new endpoint to expose verifications/analyses to users.

Ultimately, this would mean that if you can manage to ship your logs to object storage (for instance), Kargo can display them.

blakestoddard:
I scoop up logs from all pods in a cluster and send them to a central store where they're queryable by pod name (Loki) -- having the name of the job (or the child pod itself) that Argo Rollouts kicks off would be great:

`status.metricResults[...]measurements.metadata.job-name`

```
status:
. . .
  metricResults:
  - consecutiveSuccess: 1
    count: 1
    measurements:
    - finishedAt: "2025-03-11T05:47:26Z"
      metadata:
        job-name: f4a3e538-f183-4d9d-b23f-a5a55ba54bda.integration-test.1
        job-namespace: test-service
. . .
```

antoine-sncf:
In our case we use the jobs to run roboframework tests.
And, as we cannot have an output in kargo we currently use gitlab pages to store the full report but that's not exactly the same use-case.
Having the logs would help in debugging but having a link of this roboframework report from kargo's stage to access this document would be required (is there an issue about this one already?).
_argocd has an external link attribut on applications where we can set fixed links_

krancour:
> we currently use gitlab pages to store the full report

> but that's not exactly the same use-case

I'm not sure how that _isn't_ the same use case, unless perhaps what's being stored are not raw logs, but something different like a report formatted with HTML or something.

Can you clarify?

blakestoddard:
If you use something like [Allure](https://allurereport.org/) and generate an HTML report that gets thrown in S3, you could link to it (not sure how you'd handle auth for that though).

krancour:
As proposed, this is strictly about _logs_ and not reports. Displaying reports _within_ the UI or CLI does not make sense in the same way that displaying raw logs would. Linking to external reports in a separate browser tab seems useful and doable, but also seems like a separate feature from what's proposed here and I'd rather keep the scope of this issue narrow. (Feel free to open a separate issue.)

> where do we store things if it should be private

[...]

PR #3674, opened 2025-03-19
WIP(ui): AnalysisRun logs

Marvin9:
ui prototype for https://github.com/akuity/kargo/issues/3641

in Analysis run modal

Screenshot 2025-03-19 at 1 52 38 PM (2)

Deep link full page

Screenshot 2025-03-19 at 1 53 12 PM (2)

cc @jessesuen @kencochrane @krancour


kencochrane:
Looks good, we don't need "follow logs" though since these won't be live logs.

blakestoddard:
Live tail would be a nice feature if you’re running a long-running integration test and want to know where it’s at in the process.

Marvin9:
closing this for https://github.com/akuity/kargo/pull/3682 because had some resolved merge conflicts that was crashing API server and I couldn't find why

Marvin9:
> Live tail would be a nice feature if you’re running a long-running integration test and want to know where it’s at in the process.

@krancour ^

kencochrane:
I agree, but it makes things way more complicated, and we decided it is better to keep things simple and deliver this feature now, even if it isn't perfect, verse waiting until it is perfect with all the features. 

So, we deliver it with the logs minus live streaming to start, and then we can come back to this when we have time, and add the live streaming feature later. 

In my opinion some logs are better than no logs. If we want live streaming of logs, it won't available in 1.4 and I really want to have logs for this, it is currently a pain without them.

PR #3715, opened 2025-03-25
feat(ui): visualize oci annotation

Marvin9:
head from https://github.com/akuity/kargo/pull/3676

Screenshot 2025-03-26 at 1 06 57 AM (2)

Screenshot 2025-03-26 at 3 16 01 PM (2)


krancour:
@Marvin9 I test drove this with arbitrary annotation foo=bar and didn't see it come through in the UI. Looking at the code, it seems this only displays two specific, well-defined annotations. Have I got that right?

If so, I am wondering if it would be better to display all the annotations that @hiddeco's changes find.

@jessesuen how were you expecting this to work?

Marvin9:
Correct @krancour its because those are the thing that are meaningful, all other annotation you can see it from manifest. I will update at least freight details to show more raw annotation values

Marvin9:
@krancour would you test this again? Now it shows all OCI annotation in freight details

jessesuen:
Sorry for commenting late but I was in agreement with Kent in that we should be able to see all the annotations.

Marvin9:
> Sorry for commenting late but I was in agreement with Kent in that we should be able to see all the annotations.

I have updated UI to show all OCI in freight details. @krancour Please test now

nikolay-te:
This is great! Would these be also eventually be accessible in the freight, e.g. be able to use them during promotions?

hiddeco:
@nikolay-te the [`imageFrom()` expression function](https://docs.kargo.io/user-guide/reference-docs/expressions#imagefromrepourl-freightorigin) will return an object with an `Annotations` field.

krancour:
> I was in agreement with Kent in that we should be able to see all the annotations.

@jessesuen, my apologies, but we need more clarification still...

Is "all" really all just those with [predefined keys](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys)?

Currently this still will not show completely arbitrary ones like `foo: bar`.

ui/src/features/freight-timeline/open-container-initiative.ts
```diff
@@ -0,0 +1,71 @@
...
+  } else if (url.includes('bitbucket.org')) {
+    baseUrl = url
+      .replace(/^git@bitbucket.org:/, 'https://bitbucket.org/')
+      .replace(/^https?:\/\/bitbucket.org\//, 'https://bitbucket.org/')
+      .replace(/\.git$/, '');
+    return `${baseUrl}/commits/${revision}`;
+  }
+
+  return url;
+};
```
rpelczar, line 47: Isn't better to use something similar to this? https://github.com/argoproj/argo-cd/blob/master/ui/src/app/shared/components/urls.ts

ui/src/features/freight-timeline/freight-content-item.tsx
```diff
@@ -54,21 +58,49 @@ export const FreightContentItem = (props: {
...
-        </a>
-      ) : (
-        _children
-      )}
+      <div className='flex gap-2 items-center'>
+        <FontAwesomeIcon icon={icon} style={{ fontSize: '14px' }} className={classNames('px-1')} />
+        {props.artifactSource && (
+          <a
+            href={props.artifactSource}
+            className={classNames('text-blue-500', {
```
rpelczar, line 66: Please don't use colors from tailwind. If you want to have a blue color link, use Typography.Link component.

ui/src/features/freight-timeline/freight-content-item.tsx
```diff
@@ -54,21 +58,49 @@ export const FreightContentItem = (props: {
...
-        })}
-      />
-      {href ? (
-        <a target='_blank' className={linkClass}>
-          {_children}
-        </a>
-      ) : (
-        _children
-      )}
+      <div className='flex gap-2 items-center'>
```
rpelczar, line 61: If posible use Flex compnent.

rpelczar (changes requested):

rpelczar (approved):

jessesuen:
> Currently this still will not show completely arbitrary ones like foo: bar.

I meant all as in **_any_** annotation, including `foo: bar`. I suspect there would be use cases to see annotations outside the standard. If we think seeing too many annotations would be an issue in the future, we could always have a toggle/preference to show only "standard" annotations.

krancour:
Thank you for the clarification @jessesuen!

ui/src/features/freight-timeline/open-container-initiative.ts
Marvin9: I think its fine

ui/src/features/freight-timeline/open-container-initiative.ts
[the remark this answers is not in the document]
Marvin9: I would add suffix. I learnt from @krancour in backend that `util` name is code smell and that kind of stuck to me because truly its hard to know what it contains and also hard to find the file by function name.

krancour (approved):

PR #3723, opened 2025-03-27
fix(api server): take greater care decoding log steams

krancour:
These fixes make the log streaming endpoint more tolerant of different encoding schemes used when retrieving logs from some other source via HTTP.

It's also more careful about the possibility of multi-byte character sequences in any supported encoding spanning the end of one chunk and the start of the next when streaming.

cc @gdsoumya

### DIFF
--- a/internal/server/get_analysisrun_logs_v1alpha1.go
+++ b/internal/server/get_analysisrun_logs_v1alpha1.go
@@ -10,6 +10,9 @@ import (
 
 	"connectrpc.com/connect"
 	"github.com/hashicorp/go-cleanhttp"
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/encoding/unicode"
+	"golang.org/x/text/transform"
 	"k8s.io/apimachinery/pkg/types"
 
 	svcv1alpha1 "github.com/akuity/kargo/api/service/v1alpha1"
@@ -17,6 +20,7 @@ import (
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 	"github.com/akuity/kargo/internal/api"
 	"github.com/akuity/kargo/internal/api/stubs/rollouts"
+	libEncoding "github.com/akuity/kargo/internal/encoding"
 	"github.com/akuity/kargo/internal/expressions"
 	"github.com/akuity/kargo/internal/server/user"
 )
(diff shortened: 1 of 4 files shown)


krancour:
Note to self to open an issue to be more careful about the encoding of responses in the http step.

internal/server/get_analysisrun_logs_v1alpha1.go
```diff
@@ -409,3 +421,62 @@ func (s *server) buildRequest(
...
+					return
+				case <-ctx.Done():
+					return
+				}
+			}
+		}
+	}()
+
+	return chunkCh, errCh, nil
+}
```
hiddeco, line 482: Maintaining a single channel seems slightly cleaner:

```suggestion
// LogChunk represents a chunk of log data or an error.
type LogChunk struct {
	Data  string
	Error error
}

// StreamLogs reads log data from the provided reader, decodes it using the
// specified decoder, and returns a channel that receives chunks of log data.
// The channel is closed when all data has been read or an error occurs.
func StreamLogs(
	ctx context.Context,
	reader *bufio.Reader,
	decoder *encoding.Decoder,
	bufferSize int,
) (<-chan LogChunk, error) {
	// Handle UTF-16 BOM if present
	peekedBytes, err := reader.Peek(2)
	if err != nil && err != io.EOF {
		return nil, fmt.Errorf("error peeking at log stream: %w", err)
	}
	if libEncoding.HasUTF16BOM(peekedBytes) {
		if _, err = reader.Discard(2); err != nil {
			return nil, fmt.Errorf("error discarding BOM: %w", err)
		}
	}

	transformReader := transform.NewReader(reader, decoder)
	resultCh := make(chan LogChunk)

	go func() {
		defer close(resultCh)
		buf := make([]byte, bufferSize)

		for {
			n, err := transformReader.Read(buf)
			if n > 0 {
				select {
				case resultCh <- LogChunk{Data: string(buf[:n])}:
				case <-ctx.Done():
					return
				}
			}
			if err != nil {
				if err == io.EOF {
					return
				}
				select {
				case resultCh <- LogChunk{Error: fmt.Errorf("error reading data: %w", err)}:
				case <-ctx.Done():
				}
				return
			}
		}
	}()

	return resultCh, nil
}
```

PR #3740, opened 2025-03-28
fix: always honor availability strategy

krancour:
#3257 introduced availability strategies (verified + soaked in any upstream vs all upstreams).

The change was properly implemented as new criterion when _listing_ available Freight, but was overlooked in the function that answers whether a single piece of Freight is available to a given Stage.

The effect is that Freight that was not verified and adequately soaked in _all_ upstream stages may sometimes have been deemed available to a Stage if it met availability criteria in _any_ upstream stage.

This PR fixes that.

### DIFF
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -181,15 +181,26 @@ func (s *Stage) IsFreightAvailable(freight *Freight) bool {
 		return true
 	}
 	for _, req := range s.Spec.RequestedFreight {
-		if freight.Origin.Equals(&req.Origin) {
-			if req.Sources.Direct {
-				return true
-			}
-			for _, source := range req.Sources.Stages {
-				if freight.IsVerifiedIn(source) {
-					return req.Sources.RequiredSoakTime == nil ||
-						freight.GetLongestSoak(source) >= req.Sources.RequiredSoakTime.Duration
+		if !freight.Origin.Equals(&req.Origin) {
+			continue
+		}
+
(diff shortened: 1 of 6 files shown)


api/v1alpha1/stage_types.go
```diff
@@ -185,10 +185,25 @@ func (s *Stage) IsFreightAvailable(freight *Freight) bool {
...
+				}
+				return true
+			}
+			// Make sure Freight is verified and soaked in at least one upstream Stage
+			for _, stage := range req.Sources.Stages {
+				_, verified := freight.Status.VerifiedIn[stage]
+				soaked := req.Sources.RequiredSoakTime == nil ||
+					freight.GetLongestSoak(stage) >= req.Sources.RequiredSoakTime.Duration
+				if verified && soaked {
+					return true
```
hiddeco, line 206: It might be nice to make this logic a method on `Freight`, so that we can avoid the logic duplication in this and the above loop.

krancour: That had occurred to me at one point and I think I didn't entertain it very long because in the other spot, a lot of the work is done through field selectors, but this aspect of it isn't, so I shouldn't have dismissed it so quickly. 👀

krancour: Oh... sorry... I misunderstood. You mean the loop immediately above this. Yes. But I think there's even more opportunity than that to DRY things up.

hiddeco (approved):

PR #3751, opened 2025-03-29
initial work on acr authentication

snuxoll:


krancour:
Closing this in favor of #4623, which is a simpler approach to Workload Identity that doesn't attempt to assume/impersonate a Project-specific identity. Since that is a considerably more complex thing to accomplish with Azure than it is with either AWS or GCP, that is the extent of what we are comfortable supporting at this time.

PR #3760, opened 2025-03-30
feat: add public oci repository support to thhelm-template promotion step

koorikla:


koorikla:
https://github.com/akuity/kargo/issues/1610

Issue #3686, opened 2025-03-21
Improve Project details navigation


rpelczar:
# Proposed Feature
Use tabs or something similar for navigation in the project details.

# Motivation
It's not easy to find this navigation, and it's not obvious what is there.

![Image](https://github.com/user-attachments/assets/94d7bfa4-9abc-4c49-96b8-3140bbf6a0a8)

# Suggested Implementation
Replace icons with https://ant.design/components/tabs or add labels to the icons.

PR #3768, opened 2025-03-31
feat(ui): Project Settings Redesign

### LINKED ISSUE #3685: Project settings
If we want to edit or delete a project we have to click the arrow down icon next to the project title which is not the best UX practice.

Image

I want to propose adding completely a new tab dedicated to project settings.
We can move the edit yaml and delete options there, and we get the ability to add more settings in the future.

rpelczar:
Fixes #3685 #3686 

Changes:
- added breadcrumbs
- added project settings
- moved edit/delete project to the project settings
- moved some tabs to the project settings
- small design updates

image

image

image


ui/src/pages/project.tsx
```diff
@@ -105,68 +45,33 @@ export const Project = ({
...
+          </Button>
+          <Button
+            icon={<FontAwesomeIcon icon={faCog} />}
+            onClick={() => navigate(generatePath(paths.projectSettings, { name }))}
+            size='small'
+          >
+            Settings
+          </Button>
+        </Space>
+      </BaseHeader>
```
Marvin9, line 68: Can we have this header outside of normal loading state?

Marvin9 (changes requested):

ui/src/pages/project.tsx
rpelczar: Yes, I wanted to have it, but it needs more changes to avoid a double loader.

image

Marvin9 (approved):
LGTM!

PR #3771, opened 2025-04-02
fix(ui): projects filter ui

mirrajabi:
Two issues with project filter input are fixed. (You may also see it as one issue and a feature)

## 1. The hidden search bar

In the case that the filter result was returned empty, the search bar would disappear, blocking the user from resetting the search. The only fix in this situation was to remove the `projects-filter` cookie to get the search bar back.

The image below shows how it would like before, if you would search for a non-existing project name:
ne

## 2. Hitting the enter button wouldn't change the `projects-filter`

This one was more of an annoying thing because previously you had to manually click on the search button to change the search filter. Now just hitting enter will do that.

I've checked the other interactions with the search bar like selecting different items and all still work as expected.

P.S. I've uploaded a screen recording of how it looks with this PR in [this](https://github.com/akuity/kargo/pull/3771#issuecomment-2773570479) comment below.

### DIFF
--- a/ui/src/features/project/list/project-list-filter.tsx
+++ b/ui/src/features/project/list/project-list-filter.tsx
@@ -26,7 +26,7 @@ export const ProjectListFilter = ({
   const handleKeyDown = (e: React.KeyboardEvent) => {
     if (e.key !== 'Enter') return;
 
-    if (filteredProjects?.length !== 1) {
+    if (filteredProjects?.length !== 1 || !filter) {
       onChange(filter);
       return;
     }
--- a/ui/src/features/project/list/project-list-filter.tsx
+++ b/ui/src/features/project/list/project-list-filter.tsx
@@ -24,11 +24,16 @@ export const ProjectListFilter = ({
   );
 
   const handleKeyDown = (e: React.KeyboardEvent) => {
-    if (e.key === 'Enter' && filteredProjects?.length === 1) {
-      const selectedProject = filteredProjects[0].metadata?.name;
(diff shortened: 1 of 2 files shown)


mirrajabi:
This is how it looks after the improvements:

https://github.com/user-attachments/assets/8333f68a-1a00-450f-85c5-2970929a3500

krancour:
@mirrajabi thanks for this! I agree this can be improved.

I test drove this and found the keypress on enter sometimes behaving in surprising ways.

In this example, I search for "foo", which does not exist. Everything works great. Then I highlight and delete "foo" thinking I'm completely clearing the filter and I hit enter, expecting to once again see all Projects, but I end up being taken directly into my Project's page, even though it does _not_ appear that I've highlighted it in the dynamic results beneath the search box.

https://github.com/user-attachments/assets/02fffb5a-d1da-48c1-a49e-61f1898a205e

mirrajabi:
@krancour Thanks for testing!

There was indeed an issue there. If the search returned exactly one result, hitting enter would return that exact result, regardless of the search input. In this case there was only one project, so an empty search returned exactly that.
I've fixed this in [f4d0517](https://github.com/akuity/kargo/pull/3771/commits/f4d0517901206eac3c16268f9376a411348c4e3f).

krancour:
@mirrajabi thank you so much. Test drove and it lgtm, but I will defer to @rpelczar or @Marvin9 for a quick review of the implementation.

rpelczar (approved):
LGTM

PR #3778, opened 2025-04-03
feat: allow specifying a token for job metric logs access + docs

krancour:
This PR addresses an oversight in the job metric logs streaming functionality where there was no _secure_ way to specify an auth header.

This PR fixes that and adds the missing documentation for the feature as well.

cc @gdsoumya


charts/kargo/templates/api/deployment.yaml
```diff
@@ -75,6 +75,15 @@ spec:
...
           {{- end }}
+          {{- if and .Values.api.rollouts.integrationEnabled .Values.api.rollouts.logs.enabled .Values.api.rollouts.logs.tokenSecret.name .Values.api.rollouts.logs.tokenSecret.key }}
+          - name: ANALYSIS_RUN_LOG_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.api.rollouts.logs.tokenSecret.name }}
+                key: {{ .Values.api.rollouts.logs.tokenSecret.key }}
+          {{- end }}
+          {{- if and .Values.api.oidc.enabled .Values.api.oidc.dex.enabled }}
+          {{- end }}
```
krancour, line 86: Hold up...  not sure where these new lines came from. 👀

krancour: Fixed in latest commit.

Issue #3646, opened 2025-03-13
Expression functions not available in conditionals or vars values passed to tasks


jessesuen:
# Description

If I try to use `commitFrom` function to pass something to a task, e.g.:

```yaml
      steps:
      - task:
          name: render-branch
        vars:
        - name: commit
          value: ${{ commitFrom("https://github.com/jessesuen/kargo-rendered-branches.git").ID }}
```

I get the following error:

`step execution failed: step 0 met error threshold of 1: failed to get step config: error pre-processing promotion variable "commit": reflect: call of reflect.Value.Call on zero Value (1:2) | commitFrom("https://github.com/jessesuen/kargo-rendered-branches.git").ID | .^`

But the same `commitFrom` will work as a normal step:

```yaml
      steps:
      - uses: git-clone
        config:
          repoURL: https://github.com/jessesuen/kargo-rendered-branches.git
          checkout:
          - commit: ${{ commitFrom("https://github.com/jessesuen/kargo-rendered-branches.git").ID }}
            path: ./src
```

# Screenshots

<img width="784" alt="Image" src="https://github.com/user-attachments/assets/d22fe755-7975-4aef-bc87-99005f6bf999" />

# Steps to Reproduce

Try to use `commitFrom` in a vars[].value passed to a task.

# Version

```
v1.3.1
```

jessesuen:
Conditionals (`if`) also lack access to functions.

krancour:
@jessesuen assigning to you only because I think you said you were working on it.

If I'm wrong of if you need to hand it off, please feel free to kick it back to me.

nghazali:
Hey @krancour and @jessesuen 
This problem is repeated in verification as well. Based on [Kargo doc](https://docs.kargo.io/user-guide/how-to-guides/verification#:~:text=value%3A%20%24%7B%7B%20commitFrom(%22https%3A//github.com/example/repo.git%22).ID%20%7D%7D) I expect expressions work in verification args. I hope we have this fix in `1.3.2` as well.  

I have this manifest:

```
  verification:
    analysisTemplates:
      - name: operators-stability-check
    analysisRunMetadata:
    args:
      - name: verification-tag
        value: ${{ imageFrom(vars.verificationUrl, warehouse("verification-images")).Tag }}
```
and I get the same error as reported in this issue:

[...]

PR #3817, opened 2025-04-08
feat: support expr funcs in if conditions and vars

hiddeco:
Fixes: #3646


hiddeco:
@krancour this needs a careful review as the "functions are allowed at variable level" is potentially opening Pandora's box — but it seemed like the better option for reasons listed above.

krancour (approved):
🎉 Nice and clean and straightforward.

krancour:
> is potentially opening Pandora's box

How so?

jessesuen (approved):
Thank you! I had a similar [approach](https://github.com/jessesuen/kargo/commit/1680a662794bed642721b6e4fa28bbe55c46f8e4) but never got around to finishing.

hiddeco:
> How so?

Just for the sake of also mentioning this online:

> The functions are evaluated every time a Promotion is run, so the outcome may change depending on the attempt. This may or may not result in (un)predictable outcomes.

Issue #3647, opened 2025-03-13
FreightFrom Expression functions should return nil if not found instead of erroring


jessesuen:
# Description

I'm trying to create a re-usable promo task that might be called with an image Freight, or it might be called with git Freight. And so I want to conditionally run a step depending on the type of freight we are dealing with. I tried doing this:

```yaml
  - uses: yaml-update
    if: ${{ imageFrom( vars.image ) != nil }}
```

But I get this error:

`step execution failed: error checking if step 1 should be skipped: reflect: call of reflect.Value.Call on zero Value (1:2) | imageFrom( vars.image ) != nil | .^`

I believe this happens because we error if Freight is not found, instead of returning nil without error:

https://github.com/akuity/kargo/blob/9c17446d404a7ee5071d06101232ee3a7aa0e0f1/internal/controller/freight/finder.go#L80-L97

When the freight is not found, we should allow for this to return `nil` without error, since there are use cases where we expect it not to exist.

Alternatively, we may need other helpers to determine if freight exists that doesn't error, but I think returning nil would be preferred.

# Screenshots

<img width="781" alt="Image" src="https://github.com/user-attachments/assets/31b60996-0930-4e14-8414-155d77c6ba11" />

# Steps to Reproduce

```yaml
  - uses: yaml-update
    if: ${{ imageFrom( vars.image ) != nil }}
```

# Version

```
v1.3.1
```

hiddeco:
Think the actual issue is that `if` (and `vars`) at present do not have access to functions.

krancour:
I think I agree with this change, but at the same time, I wonder if the nil pointer dereference you'll end up with instead of a very clear error won't obscure some legitimate user mistakes.

I wonder a little if some variations of the existing functions would allow you to easily do what you're trying to do without obscuring actual mistakes... something like `imageOrNilFrom(...)`...

But I'm really not sure of that either. 🤷‍♂

PR #3820, opened 2025-04-08
fix!: freight from funcs return nil instead of error when freight not found

jessesuen:
Resolves https://github.com/akuity/kargo/issues/3647

The `imageFrom`, `chartFrom`, `commitFrom` expr functions will now return `nil` instead of returning error if a Freight could not be found. This supports the use case where a PromoTask could be re-used for multiple types of artifact promotions (image/commit/chart) using a conditional like:

```yaml
  - uses: yaml-update
    if: ${{ imageFrom( vars.image ) != nil }}
```

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
+++ b/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
@@ -286,7 +286,14 @@ The returned `FreightOrigin` object has the following fields:
 
 The `FreightOrigin` object can be used as an optional argument to the
 `commitFrom()`, `imageFrom()`, or `chartFrom()` functions to disambiguate the
-desired source of an artifact when necessary.
+desired source of an artifact when necessary. These functions return `nil` when
+relevant `Freight` is not found from the `FreightCollection`. 
+
+:::tip
+You can handle `nil` values gracefully in Expr using its
+[nil coalescing](https://expr-lang.org/docs/language-definition#nil-coalescing) and
+[optional chaining](https://expr-lang.org/docs/language-definition#optional-chaining) features.
+:::
 
 ### `commitFrom(repoURL, [freightOrigin])`
 
--- a/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
+++ b/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
@@ -315,6 +315,8 @@ The optional `freightOrigin` argument should be used when a `Stage` requests
 `Freight` from multiple origins (`Warehouse`s) and more than one can provide a
 `GitCommit` object from the specified repository.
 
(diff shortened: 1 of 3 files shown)


hiddeco:
I was wondering today if returning an empty object wouldn't be more convenient as it would prevent you from having to do `&&` type of checks.

jessesuen:
> I was wondering today if returning an empty object wouldn't be more convenient

Hmmm, I'd like to understand more. What would be the equivalent of the following?

```yaml
  - uses: yaml-update
    if: ${{ imageFrom( vars.image ) != nil }}
```

jessesuen:
One advantage I see with nil, is that you could use expr's [nil coalescing](https://expr-lang.org/docs/language-definition#nil-coalescing) syntax. e.g.:

```
commitFrom( vars.commit ) ?? "main"
```

hiddeco:
My initial idea was that doing something like the following would be more straightforward in most cases than `nil` checking (and then possibly having to look at _other_ things as well):

```
${{ commitFrom('https://git.example.com/repo.git').ID != '' }}
```

But I had forgotten about the fact that [optional chaining](https://expr-lang.org/docs/language-definition#optional-chaining) exists in expression language, which makes this as easy when `nil` is returned:

```
${{ commitFrom('https://git.example.com/repo.git')?.ID != '' }}
```

Given this, I am now convinced your suggestion is the best way.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -315,6 +315,8 @@ The optional `freightOrigin` argument should be used when a `Stage` requests
 `Freight` from multiple origins (`Warehouse`s) and more than one can provide a
```
hiddeco, line 315: Probably an idea to add references to [optional chaining](https://expr-lang.org/docs/language-definition#optional-chaining) and [nil coalescing](https://expr-lang.org/docs/language-definition#nil-coalescing) to this document to help people supercharge the `nil` returns.

krancour:
@jessesuen, I had #3647 in v1.4.1 initially, but as this is a (very minor) breaking change, I am thinking this is more appropriately a v1.5 feature unless you have some reason otherwise. wdyt?

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
jessesuen: Will add a note in the docs

jessesuen:
I'm okay with v1.5, since:
* So far I think I'm the only one hitting it
* It'll be in a v1.5 nightly
* It would need to go with https://github.com/akuity/kargo/pull/3817 to be more useful

hiddeco (approved):

PR #3825, opened 2025-04-09
feat: support for Stage-level variables in PromotionTemplate and Verification config

### LINKED ISSUE #3787: Introduce stage-level `spec.vars` referenceable from `verifications`
# Proposed Feature

Stages have a concept of vars, but they are defined at the `spec.promotionTemplate.spec.vars` level. This means that they are not accessible to areas outside `promotionTemplate`, namely `verification`. 

But verification might also need to access these vars and pass them to an AnalysisTemplate. This proposal is to move `vars` to the `.spec` level, so that they can be referenced by verification.

Example:

```yaml
kind: Stage
...
spec:
  vars:
  - name: branch
    value: foo
  promotionTemplate:
    spec:
      steps:
      - uses: git-clone
        config:
          repoURL: https://github.com/example/example.git
          checkout:
          - branch: ${{ vars.branch }}
            path: ./out
  verification:
    analysisTemplates:
    - name: smoke-test
    args:
    - name: branch
      value: ${{ vars.branch }}
```

# Motivation

This part of overall effort to keep things DRY. 
Related issue https://github.com/akuity/kargo/issues/3733.

Combined with https://github.com/akuity/kargo/issues/3733, users would have two ways to keep things DRY:
1. if var is needed across Stages, it would go into a ConfigMap
2. if var needed for single Stage, it would go into the Stage's `spec.vars`

# Suggested Implementation

See above spec proposal. To maintain backward compatibility, we could prefer looking at `spec.vars`, but if those are not defined, then we fall back to `spec.promotionTemplate.vars`

hiddeco:
Resolves #3787


api/v1alpha1/stage_types.go
```diff
@@ -220,6 +220,9 @@ type StageSpec struct {
 	// kargo.akuity.io/shard label with the value of this field. When this field
 	// is empty, the webhook will ensure that label is absent.
 	Shard string `json:"shard,omitempty" protobuf:"bytes,4,opt,name=shard"`
+	// Vars is a list of variables that can be referenced by the PromotionTemplate
+	// and the arguments of the Verification.
```
krancour, line 224: Non-blocking nit. If we want to drive home the notion of "scopes" for the vars, it might help (and may be more future-proof, to boot) if we say these vars can be referenced _anywhere in the Stage spec that supports expressions_ and cite PromotionTemplate and Verification as good examples.

krancour (approved):

PR #3827, opened 2025-04-09
chore(deps): update Helm to v3.17.3

hiddeco:
[Helm `v3.17.3` is a security patch release](https://github.com/helm/helm/releases/tag/v3.17.3), which prioritizes an update and backport.


krancour (approved):
fwiw, I also confirmed that the base image built with apko gets the `helm` v3.17.3 binary.

Issue #3404, opened 2025-01-29
Add a way to map users git email address to a user name for easier understanding who made the change.


kencochrane:
# Proposed Feature

The list of emails in the "Created By" column on the stage details page seems to be coming from the git commits. It would be nice if there were a way to map these git emails to users and display the user's names in case some of the emails aren't obvious who they are. You could still show the emails if needed (Maybe when you hover over, or click on an "i" icon)

<img width="718" alt="Image" src="https://github.com/user-attachments/assets/25ffd224-cf57-45a8-a170-7a1b5d0c01fd" />

# Motivation

Make it created by list more user-friendly and easier to use when someone has weird email addresses for their git commits. This will also make it easier if someone commits code from multiple email addresses.

# Suggested Implementation

None

krancour:
@kencochrane those aren't coming from a commit, but from claims in the ID token of the authenticated user who manually triggered the Promotion. In the case of an auto-promotion, it will show up as "N/A".

Current logic seems to prefer the email claim, followed by the sub claim. Then it gives up if neither of those are found. But those are standard claims, so in practice, it is probably always email.

GitHub user handles would be nice, but we'd only have a claim with that info if GitHub is the IDP.

Since many useful claims (like GitHub handle) _aren't_ standard, the best thing might be to just make it configurable which claim Kargo should use as the username.

wdyt?

kencochrane:
@krancour yeah that seems reasonable, thanks.

PR #3841, opened 2025-04-10
feat: support custom username claim

fuskovic:
Closes https://github.com/akuity/kargo/issues/3404

Allows propogating username claim from `values.yaml` -> server config map -> oidc config.

This helps resolves actors to their oidc username during audit if provided; falling back to the existing default behavior otherwise.


internal/server/user/user.go
```diff
@@ -27,6 +27,9 @@ type Info struct {
 	// ServiceAccountsByNamespace is the mapping of namespace names to sets of
 	// ServiceAccounts that a user has been mapped to.
 	ServiceAccountsByNamespace map[string]map[types.NamespacedName]struct{}
+	// Username is the username of the user. This is typically the email address
+	// of the user, but may be different depending on the identity provider.
```
krancour, line 31: Small suggestion on wording here, just because it feels like this is more a function of configuration than which identity provider is in use:

```suggestion
	// Username is the username of the user. This is often the email address
	// of the user, but may be different depending on configuration.
```

internal/server/option/auth.go
```diff
@@ -404,6 +404,7 @@ func (a *authInterceptor) authenticate(
 				Claims:                     c,
 				ServiceAccountsByNamespace: sa,
 				BearerToken:                rawToken,
+				Username:                   a.cfg.OIDCConfig.UsernameClaim,
```
krancour, line 407: Instead of putting the name of the claim here, let's put the _value_ of the claim.

internal/api/event.go
```diff
@@ -21,7 +21,14 @@ func FormatEventControllerActor(name string) string {
...
 func FormatEventUserActor(u user.Info) string {
-	var email, subject string
+	var email, subject, username string
+	if u.Username != "" && u.Username != "email" {
+		if usernameClaim, ok := u.Claims[u.Username]; ok {
+			if usernameStr, ok := usernameClaim.(string); ok {
+				username = usernameStr
+			}
+		}
+	}
```
krancour, line 31: I think this whole function can be boiled down to something simpler:

```
func FormatEventUserActor(u user.Info) string {
	if u.IsAdmin {
		return kargoapi.EventActorAdmin
	}
	if u.Username != "" {
		return kargoapi.EventActorOidcUsernamePrefix + u.Username
	}
	if emailClaim, ok := u.Claims["email"]; ok {
		if email, ok := emailClaim.(string); ok {
			return kargoapi.EventActorEmailPrefix + email
		}
	}
	if subClaim, ok := u.Claims["sub"]; ok {
		if sub, ok := subClaim.(string); ok {
			return kargoapi.EventActorSubjectPrefix + sub
		}
	}
	return kargoapi.EventActorUnknown
}
```

(fwiw, it was a tad more complex than necessary before you touched it.)

api/v1alpha1/event.go
```diff
@@ -40,5 +40,6 @@ const (
 	EventActorEmailPrefix          = "email:"
 	EventActorSubjectPrefix        = "subject:"
 	EventActorKubernetesUserPrefix = "kubernetes:"
+	EventActorOidcUsernamePrefix   = "oidc-username:"
 	EventActorUnknown              = "unknown actor"
 )
```
krancour, line 45: I think that if, historically, the prefix has been `email:` when we extracted actor information from an `email` claim and `subject:` when we've extracted it from the `sub` claim (ya... there's a slight mismatch there), the right thing to do, in general, would be to use whatever claim the operator specified as the prefix.

I would probably just remove this line altogether and then refer to [this comment]() where you'll see we no longer need this constant.

```suggestion
	EventActorEmailPrefix          = "email:"
	EventActorSubjectPrefix        = "subject:"
	EventActorKubernetesUserPrefix = "kubernetes:"
	EventActorUnknown              = "unknown actor"
)

var EventActorOIDCClaimPrefix = os.GetEnv("OIDC_USERNAME_CLAIM", "email") + ":"
```

internal/api/events_test.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		{
+			name: "oidc-username",
+			user: user.Info{
+				Claims: map[string]any{
+					"oidc-provider": "oidc-username",
+				},
+				Username: "oidc-provider",
+			},
+			expected: kargoapi.EventActorOidcUsernamePrefix + "oidc-username",
```
krancour, line 49:
```suggestion
			expected: kargoapi.EventActorOIDCClaimPrefix + "oidc-username",
```

charts/kargo/values.yaml
```diff
@@ -211,6 +211,8 @@ api:
     ## @param api.oidc.additionalScopes The additional scopes to send to the OIDC provider. This should be set to the scopes you wish to be provided to your identity provider from clients of Kargo, the scopes openid, profile and email are always requested and don't need to be added, this value is intended for any additional ones you require.
     additionalScopes:
     - groups
+    ## @param api.oidc.usernameClaim The claim to use as the username for the user. This is typically set to `email` or `sub`. If you are using Dex, this value will be ignored and the username claim will be automatically configured.
```
krancour, line 214: This is applicable with or without Dex involved.

internal/api/events_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/server/user"
+)
+
+func TestFormatEventUserActorForOidc(t *testing.T) {
```
krancour, line 12: Conventionally, we should use `Test<name of the function under test>`:

```suggestion
func TestFormatEventUserActor(t *testing.T) {
```

internal/api/events_test.go
```diff
@@ -0,0 +1,57 @@
...
+			name: "sub",
+			user: user.Info{
+				Claims: map[string]any{
+					"email": "email@inbox.com",
+				},
+			},
+			expected: kargoapi.EventActorEmailPrefix + "email@inbox.com",
+		},
+		{
+			name: "oidc-username",
```
krancour, line 42: Heads up that other suggestions I made will probably break this test case.

internal/server/option/auth.go
krancour:
```suggestion
				Username:                   c[a.cfg.UsernameClaim],
```

If this somehow ends up being the empty string (like if the claim the operator wanted to use for this purpose didn't actually exist), it's fine, because we know the function that formats actor information will fall back on its old strategy of directly going after `email` and `sub` claims.

internal/server/option/auth_test.go
```diff
@@ -472,6 +472,37 @@ func TestVerifyIDPIssuedTokenFn(t *testing.T) {
...
+						"email":                  "wrong-username@failed-test.io",
+					}, nil
+				},
+			},
+			assertions: func(t *testing.T, c claims, err error) {
+				require.NoError(t, err)
+				require.Equal(t, "oidc_provider_username", c["oidc_username_claim"])
+				require.Equal(t, "correct-username", c["oidc_provider_username"])
+			},
+		},
```
krancour, line 505: I would probably just work this new detail of `Username` being set properly into the existing success test case above.

krancour:
I made a lot of small suggestions, but this was really in pretty decent shape to begin with. Nice work. Should be easy to take it the rest of the way, I think.

internal/server/option/auth_test.go
krancour: Discussed offline. Actually seems this particular test didn't require modifications and `TestAuthenticate()` needs to be amended instead.

internal/api/events_test.go
```diff
@@ -0,0 +1,54 @@
...
+			},
+			expected: kargoapi.EventActorSubjectPrefix + "subject",
+		},
+		{
+			name: "oidc-username",
+			user: user.Info{
+				Username: "oidc-username",
+			},
+			expected: formatOidcUsername("oidc-username"),
+		},
```
krancour, line 47: Nit: I would make this the second test case, between admin and email. It makes the order of the test cases reflect the precedence rules that the function under test follows.

internal/server/option/auth_test.go
```diff
@@ -308,7 +308,8 @@ func TestAuthenticate(t *testing.T) {
 			authInterceptor: &authInterceptor{
 				cfg: config.ServerConfig{
 					OIDCConfig: &libOIDC.Config{
-						IssuerURL: testIDPIssuer,
+						IssuerURL:     testIDPIssuer,
+						UsernameClaim: "oidc-username",
```
krancour, line 312: Nit: Instead of `oidc-username` I would probably use a real claim name like `preferred_username` or a nonsense word like `foo`, only because I think it's too easy to look at `oidc-username` and be fooled into thinking this value _means something_. I'd probably say the same for the claim's value down on line 326.

```suggestion
						UsernameClaim: "oidc-username",
```

internal/server/option/auth.go
```diff
@@ -398,12 +398,25 @@ func (a *authInterceptor) authenticate(
...
+		un, ok := c[a.cfg.OIDCConfig.UsernameClaim]
+		if ok {
+			username, ok = un.(string)
+			if !ok {
+				return ctx, fmt.Errorf(
+					"claim %q must be a string; got %T",
+					a.cfg.OIDCConfig.UsernameClaim,
+					un,
+				)
+			}
```
krancour, line 411: Generally preferred when you can get away with it:

```suggestion
			if username, ok = un.(string); !ok {
				return ctx, fmt.Errorf(
					"claim %q must be a string; got %T",
					a.cfg.OIDCConfig.UsernameClaim,
					un,
				)
			}
```

hiddeco (approved):
🎉🚀 great work @fuskovic!

Issue #3663, opened 2025-03-17
controller: try to make step timeout somewhat more accurate


krancour:
Numerous discussions have taken place over the last few months re: configurability of the retry interval for step retries and timeouts, possibly not being honored.

I'll borrow my own recent explanation for the difficulty behind this from #3515:

> the timeouts are not exact because steps do not continuously retry internally. If the timeout hasn't elapsed, a step that's still running (waiting on something external) is retried on the next reconciliation attempt.
> 
> In general, those attempts are every five minutes, but the next attempt can be sooner if a related resource has a state change that forces the Promotion back onto the queue. It can also be later depending on the depth of the queue.
>
> ...
>
> We're a little bit at the mercy of the controller runtime here since we don't have precise control over the interval before the next reconciliation...
>
> That said, we can probably get closer to the specified timeout by shortening the requeue interval when timeout is sooner than when the next reconciliation attempt would typically be.

If someone wishes to tackle this, it should only be a few lines of code [right around here](https://github.com/akuity/kargo/blob/e2cdf0110122c6aa2740fed01c0cf0aca60cc4e7/internal/controller/promotions/promotions.go#L419-L421).

We can keep the hard coded five minutes, but _shorten it_ by the appropriate amount if the running step's start time + timeout is sooner than now + five minutes.

Note: At this time, I'm still hesitant to apply a similar strategy for making retry intervals configurable because I think our inability to offer much precision there is more problematic than it is with timeouts. Let's, at least, consider it out of scope for this issue.

ashish111333:
@krancour  can I help with this one ?

krancour:
@ashish111333, go for it. And thank you!

krancour:
@ashish111333 if you don't have the bandwidth for this, please let me know, as I believe @fuskovic will have the bandwidth, but I don't want to yank the rug out from under you if you're already working on this.

PR #3851, opened 2025-04-11
feat:  improve step timeout accuracy

fuskovic:
Closes https://github.com/akuity/kargo/issues/3663

Previously we were hard-coding the requeue interval to 5 minutes.

This sets a lower requeue interval if the target timeout is before 5 minutes from now.

### DIFF
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -668,6 +668,7 @@ func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
 	step := p.Spec.Steps[p.Status.CurrentStep]
 	runner := promotion.GetStepRunner(step.Uses)
 	timeout := ptr.To(time.Duration(0))
+	
 	if retryCfg, isRetryable := runner.(pkgPromotion.RetryableStepRunner); isRetryable {
 		timeout = retryCfg.DefaultTimeout()
 	}
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -414,12 +414,8 @@ func (r *reconciler) Reconcile(
 		return ctrl.Result{}, err
 	}
 
-	// If the promotion is still running, we'll need to periodically check on
-	// it.
-	//
-	// TODO: Make this configurable
 	if newStatus.Phase == kargoapi.PromotionPhaseRunning {
-		return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
+		return ctrl.Result{RequeueAfter: calculateRequeueInterval(promo)}, nil
 	}
 	return ctrl.Result{}, nil
 }
@@ -664,4 +660,31 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
 		}
 	}
 	return creator
+}
+
+func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
+	defaultRequeueInterval := 5 * time.Minute
+	for i := range p.Spec.Steps {
+		if p.Status.CurrentStep == int64(i) {
+			step := &promotion.Step{
+				Kind:   p.Spec.Steps[i].Uses,
+				Alias:  p.Spec.Steps[i].As,
+				If:     p.Spec.Steps[i].If,
+				Retry:  p.Spec.Steps[i].Retry,
+				Vars:   p.Spec.Steps[i].Vars,
+				Config: p.Spec.Steps[i].Config.Raw,
+			}
+			var (
+				runner        = promotion.GetStepRunner(step)
+				timeout       = step.GetTimeout(runner)
+				md            = p.Status.StepExecutionMetadata[p.Status.CurrentStep]
+				fiveMinLater  = time.Now().Add(5 * time.Minute)
+				targetTimeout = md.StartedAt.Time.Add(*timeout)
+			)
+			if targetTimeout.Before(fiveMinLater) {
+				return time.Until(targetTimeout)
+			}
+		}
+	}
+	return defaultRequeueInterval
 }
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -666,8 +666,8 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
 func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
 	var (
 		defaultRequeueInterval = 5 * time.Minute
-		apiStep                = p.Spec.Steps[p.Status.CurrentStep]
-		runner                 = promotion.GetStepRunner(apiStep.As)
+		step                   = p.Spec.Steps[p.Status.CurrentStep]
+		runner                 = promotion.GetStepRunner(step.As)
 		timeout                = ptr.To(time.Duration(0))
 	)
 
--- a/internal/controller/promotions/promotions.go
+++ b/internal/controller/promotions/promotions.go
@@ -668,8 +668,9 @@ func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
 	step := p.Spec.Steps[p.Status.CurrentStep]
 	runner := promotion.GetStepRunner(step.Uses)
 	timeout := ptr.To(time.Duration(0))
-	
-	if retryCfg, isRetryable := runner.(pkgPromotion.RetryableStepRunner); isRetryable {
+
+	retryCfg, isRetryable := runner.(pkgPromotion.RetryableStepRunner)
+	if isRetryable {
 		timeout = retryCfg.DefaultTimeout()
 	}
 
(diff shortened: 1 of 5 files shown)


internal/controller/promotions/promotions.go
```diff
@@ -665,3 +661,30 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
 	}
 	return creator
 }
+
+func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
+	defaultRequeueInterval := 5 * time.Minute
+	for i := range p.Spec.Steps {
```
krancour, line 667: Why iterate here when we know which element we're looking for and could grab it directly without looking for it?

internal/promotion/registry.go
```diff
@@ -24,3 +26,7 @@ var stepRunnerReg = stepRunnerRegistry{}
 func RegisterStepRunner(runner promotion.StepRunner) {
 	stepRunnerReg.register(runner)
 }
+
+func GetStepRunner(step *Step) promotion.StepRunner {
```
hiddeco, line 30: Might be better to simply do:

```suggestion
func GetStepRunner(name string) promotion.StepRunner {
```

internal/controller/promotions/promotions.go
```diff
@@ -665,3 +661,30 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
...
+				If:     p.Spec.Steps[i].If,
+				Retry:  p.Spec.Steps[i].Retry,
+				Vars:   p.Spec.Steps[i].Vars,
+				Config: p.Spec.Steps[i].Config.Raw,
+			}
+			var (
+				runner        = promotion.GetStepRunner(step)
+				timeout       = step.GetTimeout(runner)
+				md            = p.Status.StepExecutionMetadata[p.Status.CurrentStep]
+				fiveMinLater  = time.Now().Add(5 * time.Minute)
```
hiddeco, line 681:
```suggestion
				fiveMinLater  = time.Now().Add(defaultRequeueInterval)
```

internal/controller/promotions/promotions.go
```diff
@@ -665,3 +661,30 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
...
+	for i := range p.Spec.Steps {
+		if p.Status.CurrentStep == int64(i) {
+			step := &promotion.Step{
+				Kind:   p.Spec.Steps[i].Uses,
+				Alias:  p.Spec.Steps[i].As,
+				If:     p.Spec.Steps[i].If,
+				Retry:  p.Spec.Steps[i].Retry,
+				Vars:   p.Spec.Steps[i].Vars,
+				Config: p.Spec.Steps[i].Config.Raw,
+			}
```
hiddeco, line 676: Think much of this construction is unnecessary if my comment about `GetStepRunner` is taken into account.

internal/promotion/registry.go
```diff
@@ -24,3 +26,7 @@ var stepRunnerReg = stepRunnerRegistry{}
 func RegisterStepRunner(runner promotion.StepRunner) {
 	stepRunnerReg.register(runner)
 }
+
+func GetStepRunner(name string) promotion.StepRunner {
```
krancour, line 30:
```suggestion
// GetStepRunner returns a StepRunner from the package's internal registry. If
// no StepRunner is registered with the given name, nil is returned instead.
func GetStepRunner(name string) promotion.StepRunner {
```

internal/controller/promotions/promotions.go
```diff
@@ -665,3 +662,27 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
 	}
 	return creator
 }
+
+func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
+	var (
+		defaultRequeueInterval = 5 * time.Minute
+		step                   = p.Spec.Steps[p.Status.CurrentStep]
+		runner                 = promotion.GetStepRunner(step.As)
```
krancour, line 670: As is a user-assigned name for the "instance" of the step. Uses denotes the "kind" of the step, and that's what we should use for the lookup.

```suggestion
		runner                 = promotion.GetStepRunner(step.Uses)
```

internal/controller/promotions/promotions.go
```diff
@@ -665,3 +662,27 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
 	}
 	return creator
 }
+
+func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
+	var (
```
krancour, line 667: Nit: This seems unconventional. It's actually two extra lines of code to put all these vars in a `var` block vs just declaring each of them just-in-time with `:=` instead.

internal/controller/promotions/promotions.go
```diff
@@ -665,3 +662,27 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
...
+		timeout                = ptr.To(time.Duration(0))
+	)
+
+	if retryCfg, isRetryable := runner.(pkgPromotion.RetryableStepRunner); isRetryable {
+		timeout = retryCfg.DefaultTimeout()
+	}
+
+	var (
+		md            = p.Status.StepExecutionMetadata[p.Status.CurrentStep]
+		fiveMinLater  = time.Now().Add(defaultRequeueInterval)
```
krancour, line 680: This seems like an awkward name for the variable, because if we ever change the value of `defaultRequeueInterval` we'll have to change the name of this variable as well.

Perhaps we don't need a variable for this at all (same for `targetTimeout`, perhaps) because the following seems both expressive and easy-enough to read:

`if md.StartedAt.Time.Add(*timeout).Before(time.Now().Add(defaultRequeueInterval)) {`

fuskovic: That's fair.  Should we keep `targetTimeout` though because it's used both in the conditional and `time.Until(targetTimeout)`?

internal/controller/promotions/promotions.go
```diff
@@ -665,3 +662,20 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
 	}
 	return creator
 }
+
+func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
+	defaultRequeueInterval := 5 * time.Minute
+	step := p.Spec.Steps[p.Status.CurrentStep]
+	runner := promotion.GetStepRunner(step.Uses)
+	timeout := ptr.To(time.Duration(0))
+	if retryCfg, isRetryable := runner.(pkgPromotion.RetryableStepRunner); isRetryable {
```
hiddeco, line 671: Absolute nit: now that `var()` is gone, this makes it slightly more pleasant to my eyes to read.

```suggestion
	defaultRequeueInterval := 5 * time.Minute
	step := p.Spec.Steps[p.Status.CurrentStep]
	runner := promotion.GetStepRunner(step.Uses)
	timeout := ptr.To(time.Duration(0))

	if retryCfg, isRetryable := runner.(pkgPromotion.RetryableStepRunner); isRetryable {
```

internal/controller/promotions/promotions.go
```diff
@@ -414,12 +415,8 @@ func (r *reconciler) Reconcile(
 		return ctrl.Result{}, err
 	}
 
-	// If the promotion is still running, we'll need to periodically check on
-	// it.
```
krancour, line 418: We can lose the TODO below, but I think this bit of the comment is worth keeping.

internal/controller/promotions/promotions.go
```diff
@@ -665,3 +662,22 @@ func parseCreateActorAnnotation(promo *kargoapi.Promotion) string {
...
+	defaultRequeueInterval := 5 * time.Minute
+	step := p.Spec.Steps[p.Status.CurrentStep]
+	runner := promotion.GetStepRunner(step.Uses)
+	timeout := ptr.To(time.Duration(0))
+
+	retryCfg, isRetryable := runner.(pkgPromotion.RetryableStepRunner)
+	if isRetryable {
+		timeout = retryCfg.DefaultTimeout()
+	}
+
```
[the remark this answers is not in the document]
krancour: cc @hiddeco for double-check.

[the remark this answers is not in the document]
hiddeco: I actually think we want something like this:

```go

func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
	defaultRequeueInterval := 5 * time.Minute
	step := p.Spec.Steps[p.Status.CurrentStep]
	runner := promotion.GetStepRunner(step.Uses)

	timeout := (&promotion.Step{
		Retry:  step.Retry,
	}).GetTimeout(runner)

	md := p.Status.StepExecutionMetadata[p.Status.CurrentStep]
	targetTimeout := md.StartedAt.Time.Add(*timeout)
	if targetTimeout.Before(time.Now().Add(defaultRequeueInterval)) {
		return time.Until(targetTimeout)
	}
	return defaultRequeueInterval
}
```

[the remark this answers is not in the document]
krancour: Ok... I'm right, but this isn't the ideal way to fix it. @hiddeco has a better idea.

krancour (approved):

PR #3862, opened 2025-04-15
fix(docs): correctly reference outputs

schildwaechter:
Outputs from previous steps are available from `task` variable, see also [Expression Language](https://docs.kargo.io/user-guide/reference-docs/expressions)


hiddeco (approved):

schildwaechter:
Actually it looks like the example doesn't actually work as it's not possible to do this nested lookup.

PR #3868, opened 2025-04-15
fix(chart): give controller cluster-wide configmap reads

krancour:
This was meant to be included in #3865

cc @gdsoumya for the permissions change.


hiddeco (approved):
I always forget about this too.

PR #3869, opened 2025-04-15
feat: mark skipped steps as skipped

### LINKED ISSUE #3834: Mark steps as skipped if "if" evaluates to false
# Checklist

# Proposed Feature
If a step as the an "if" statement and that evaluates to false then the step should me marked as skipped.  

# Motivation
Today the step is marked with a green check box which indicates it was successfully run.  That step hasn't actually been run but there is not indication of this in the UI.

Image

# Suggested Implementation

- I think you just need a different icon there at a minimum.  If you expand the step out there isn't a good indicator that it was skipped.  You can tell because there is no OUTPUT or CONFIG section but that is much too subtle.  It should be clear that the step was skipped.

krancour:
Fixes #3834

This also accomplishes something we've long meant to do: Since their possible states don't overlap _entirely_, this gives steps their own distinct phase type instead of just using promotion phases.

@hiddeco the back end of this is r4r.

@Marvin9 would you possibly be able to amend the PR with related UI changes? 🙏


api/v1alpha1/promotion_types.go
```diff
@@ -36,6 +37,33 @@ const (
 	PromotionPhaseAborted PromotionPhase = "Aborted"
 )
 
+type PromotionStepPhase string
```
hiddeco, line 40: Should we still call this a "phase" if we further refer to it as a "status"?

krancour: I was only mirroring the existing `PromotionPhase` type (which may also be a misnomer).

I'm not married to the name and would happily consider something different.

I had momentarily considered `PromotionStepStatus`, but felt this _sounded_ more like a struct than a string type.

krancour: (Sounds that way to me because k8s has me conditioned to think of "status" as a compound type.)

krancour: Spoke offline and settled on status.

Marvin9:
I will update the UI

PR #3870, opened 2025-04-15
feat: adding GetCommitURL function to gitprovider interface 

### LINKED ISSUE #3814: Links to commits in git subscriptions
# Checklist

# Proposed Feature

This may be a setting I can't find or something but we should be able to click through to the commit on Github/Gitlab for Git subscription types, after using the OCI annotations for container subscriptions (which is super useful) this feels like a nice addition.

# Motivation

It would allow a user to click through and see the commit that the Freight is referring to.

# Suggested Implementation

Build the URL the same way the UI does from the OCI annotations using the RepoURL and commit ID in the freight.
Or for a more future proof way (if/when more git providers come along), add a field to the subscriptions that is like the Github Autolink feature so they would need to add something like:

```
  - git:
      repoURL: https://github.com/some-org/some-repo
      linkPathFormat: /commit/{{ id }}
      strictSemvers: false
      discoveryLimit: 10
```

BenHesketh21:
fixes #3814

I've created the function for the `gitprovider` interface and added it as an output of the `git push` promotion step.

The function itself, I tried a couple of different ways of accounting for different URL formats and felt that regex was the cleanest way to do so but happy to change it if there's a better way I haven't thought of.


BenHesketh21:
I think this is ready for review now, I did need to add some clean up tasks to some unit tests to stop conflicts I was getting when registering new git providers at the beginning of tests with different methods being mocked. I hope this is a good enough method to get around that, if not then I'm happy to find a different way.

internal/gitprovider/gitprovider.go
```diff
@@ -51,6 +51,9 @@ type Interface interface {
 	// to differences in the underlying provider APIs. It is the responsibility of
 	// the caller to sort the results as needed.
 	ListPullRequests(context.Context, *ListPullRequestOptions) ([]PullRequest, error)
+
+	// GetCommitURL get a commit by the repository URL and commit SHA.
```
krancour, line 55:
```suggestion
	// GetCommitURL returns a commit URL inferred from the provided repository URL and commit SHA.
```

internal/promotion/runner/builtin/git_pusher_test.go
```diff
@@ -200,6 +201,30 @@ func Test_gitPusher_run(t *testing.T) {
 	err = workTree.AddAllAndCommit("Initial commit")
 	require.NoError(t, err)
 
+	// Set up a fake git provider
+	fakeGitProviderName := "fake1"
```
krancour, line 205: Elsewhere, in another test, this was handled a bit differently:

```suggestion
	// Cannot register multiple providers with the same name, so this takes
	// care of that problem
	fakeGitProviderName := uuid.NewString()
```

I believe this would obviate the need for the unregister function and any special cleanup.

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -181,11 +186,27 @@ func (g *gitPushPusher) run(
 		return promotion.StepResult{Status: kargoapi.PromotionPhaseErrored},
 			fmt.Errorf("error getting last commit ID: %w", err)
 	}
+
+	// Using git provider to get commit url.
+	gpOpts := &gitprovider.Options{}
+	gitProvider, err := gitprovider.New(workTree.URL(), gpOpts)
```
krancour, line 192: It's ok for this to be nil. The function you're calling will deal with it.

```suggestion
	gitProvider, err := gitprovider.New(workTree.URL(), nil)
```

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -181,11 +186,27 @@ func (g *gitPushPusher) run(
...
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionPhaseErrored},
+			fmt.Errorf("error creating git provider service: %w", err)
+	}
```
krancour, line 196: This may not be how we want to deal with this error.

Everything else this step does can potentially succeed even if we're unable to construct this URL.

I think provider not found is the only thing that can actually go wrong in the call to `gitprovider.New()` so you probably just want to ignore this error entirely, although it's probably best to leave a comment about why ignoring it is the correct thing to do as well as a safe thing to do.

You can omit the URL from the output if no provider is found.

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -181,11 +186,27 @@ func (g *gitPushPusher) run(
...
+
 	return promotion.StepResult{
 		Status: kargoapi.PromotionPhaseSucceeded,
 		Output: map[string]any{
-			stateKeyBranch: pushOpts.TargetBranch,
-			stateKeyCommit: commitID,
+			stateKeyBranch:    pushOpts.TargetBranch,
+			stateKeyCommit:    commitID,
+			stateKeyCommitURL: commitURL,
```
krancour, line 209: This can be omitted if we couldn't construct it.

internal/promotion/runner/builtin/git_pusher.go
krancour: Although...

In other steps that interact with a git provider, there is an optional provider field that can be used to explicitly identify a provider in the cases where the provider cannot possibly be inferred from the repo URL. ex: If `https://my-git.example.com` were the base URL for my GitHub Enterprise, there's nothing in the URL to suggest the GitHub provider ought to be used. In cases such as this, a user can be explicit.

That same option should probably be added to the git pusher's config.

internal/gitprovider/azure/azure.go
```diff
@@ -181,6 +182,25 @@ func (p *provider) ListPullRequests(
...
+	sha string,
+) (string, error) {
+	repoURLRegex := `^(?:(?:\w+://)?(?:\w+@)?)?(dev.azure.com|ssh.dev.azure.com\:v3)[:/]([^/]+/[^.]+)(?:\.git)?$`
+	re := regexp.MustCompile(repoURLRegex)
+	matches := re.FindStringSubmatch(repoURL)
+	if len(matches) != 3 {
+		return "", fmt.Errorf("error processing repository URL: %s: must match regex: %s", repoURL, repoURLRegex)
+	}
+	host := "dev.azure.com"
+	path := matches[2]
```
krancour, line 197: `url.Parse(...)` is probably a better, more straightforward way than a regex to break the URL up into constituent parts like host and path.

If there's a concern about not verifying the URL is a pattern this provider can work with, I wouldn't worry about it... that's basically already been determined by the call to `gitprovider.GetProvider(...)`. If URL x _got_ us this provider, x is a safe URL to use with it.

krancour: Looks like this comment probably applies to all the provider implementations.

internal/promotion/runner/builtin/git_pusher.go
BenHesketh21: Okay I'll add an optional repo URL input to the git pusher and just make it use origin (as it does now) if nothing is set.

internal/promotion/runner/builtin/git_pusher.go
BenHesketh21: Makes sense, remove the error.

internal/gitprovider/azure/azure.go
BenHesketh21: I did try `url.Parse()` first but it doesn't seem to work with SSH URLs e.g. `ssh://git@gitlab.com:akuity/kargo.git`, `url.Parse()` errors with `parse "ssh://git@gitlab.com:akuity/kargo.git": invalid port ":akuity" after host`. So regex was my next thing to try. I suppose we could use `url.Parse()` and then seeing if it is SSH if that errors?

internal/promotion/runner/builtin/git_pusher.go
BenHesketh21: Would it make sense to add some log message to say we couldn't construct the URL? Just thinking that if users are trying to make use of it and it's not appearing in the output, how would they figure out why its being omitted?

krancour: Yes. Good thought. And it's got me thinking now that it may even be good to set the URL to `Unknown` in this case. Without explicitly seeing that it didn't resolve, there might never be a thought to check the logs, so this would call attention to it more explicitly.

internal/gitprovider/azure/azure.go
krancour: Oh, right. We've run into that bit of nastiness before...

`ssh://git@gitlab.com:akuity/kargo.git` is actually not a valid URL. ssh URLs, expect a an optional colon after the hostname to be followed by a port number and not a path.

So either of these are legitimate, but the one above is not:

* `ssh://git@gitlab.com/akuity/kargo.git`
* `ssh://git@gitlab.com:22/akuity/kargo.git`

If `ssh://git@gitlab.com:akuity/kargo.git` _seemed_ valid, it's because you're mixing it up with scp-style syntax, which would be: `git@gitlab.com:akuity/kargo.git`.

It's easy to make this mistake because when you ask GitHub for the ssh URL for a repo, they give it to you in scp-style syntax. They've brainwashed us.

We had a whole issue about this forever ago: https://github.com/akuity/kargo/pull/1854

If you look at the implementation of `internal/git/git.go/git.NormalizeURL()` you'll see how we normalize scp-style syntax as proper ssh URLs.

Long story-short, I believe if you `git.Normalize()` the `repoURL` you can then reliably use `url.Parse()`. Yes -- I know `git.Normalize()` also uses `url.Parse()`, so you're parsing twice, but I'm ok with it, because this normalization process is something we should habitually do before any URL comparisons or manipulations, such as your attempt to infer an http/s URL from an ssh URL or scp-style syntax.

BenHesketh21: I see, makes sense, I'll get on that. Good to know it isn't just me😂

internal/promotion/runner/builtin/git_pusher_test.go
BenHesketh21: I just thought I'd share my reasoning for the `unregister` because I'm a bit stuck. I believe it because the other implementations of `gitprovider` give a name in the options so the tests have no need to mock `Registration.Predicate()`, this one however doesn't have a name, unless we want to add that as an optional input as well as the URL? 

But this is causing `Test_gitPusher_run` to fail as it forces the `gitprovider.New()` to iterate through all the registered providers and check if `Registration.Predicate()` returns true or false, which since the other fake providers don't implement that function, causes a `invalid memory address or nil pointer dereference` error.

This is why I added the unregister function and the test clean up. It's very likely I'm missing something, but I still can't see how this would work without it. Let me know what you think.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -34,13 +34,16 @@ Stages that write to the same branch do not write to the same files.
 | `targetBranch` | `string` | N | The branch to push to in the remote repository. Mutually exclusive with `generateTargetBranch=true`. If neither of these is provided, the target branch will be the same as the branch currently checked out in the working tree. |
 | `maxAttempts` | `int32` | N | The maximum number of attempts to make when pushing to the remote repository. Default is 50. |
 | `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of 'true' is mutually exclusive with `targetBranch`. If neither of these is provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
+| `repoURL` | `string` | N | The URL of a remote Git repository. If this is left blank, the remote URL of respository when it was cloned will be used. |
```
krancour, line 37:
```suggestion
| `repoURL` | `string` | N | The URL of a remote Git repository. If this is left blank, the remote URL of repository when it was cloned will be used. |
```

krancour: Actually, I'm unclear on why this field's been added. I see it's ultimately included in push options. I'm not clear on why that is.

internal/gitprovider/gitprovider.go
```diff
@@ -51,6 +51,9 @@ type Interface interface {
 	// to differences in the underlying provider APIs. It is the responsibility of
 	// the caller to sort the results as needed.
 	ListPullRequests(context.Context, *ListPullRequestOptions) ([]PullRequest, error)
+
+	// GetCommitURL returns a commit URL inferred from the provided repository URL and commit SHA.
+	GetCommitURL(string, string) (string, error)
```
krancour, line 56:
```suggestion
	// GetCommitURL returns a commit URL inferred from the provided repository URL
	// and commit ID.
	GetCommitURL(repoURL string, commitID string) (string, error)
```

krancour: Note on function signatures... with strings, ints, etc. where the type itself doesn't impart a clear understanding of what the argument represents, it is usually best to include a name for the argument, even if it is otherwise not needed. It helps a lot in understanding, in this case, what these two strings represent and which order they belong in.

internal/gitprovider/gitprovider.go
```diff
@@ -126,6 +129,8 @@ type Fake struct {
 		context.Context,
 		*ListPullRequestOptions,
 	) ([]PullRequest, error)
+	// GetCommitURLFn defines the functionality of the GetCommitURL method.
+	GetCommitURLFn func(string, string) (string, error)
```
krancour, line 133:
```suggestion
	// GetCommitURLFn defines the functionality of the GetCommitURL method.
	GetCommitURLFn func(repoURL string, commitID string) (string, error)
```

krancour: Same rationale as [here](https://github.com/akuity/kargo/pull/3870/files#r2060823864).

internal/gitprovider/gitprovider.go
```diff
@@ -151,3 +156,11 @@ func (f *Fake) ListPullRequests(
 ) ([]PullRequest, error) {
 	return f.ListPullRequestsFn(ctx, opts)
 }
+
+// GetCommitURL implements gitprovider.Interface.
+func (f *Fake) GetCommitURL(
+	repoURL string,
+	sha string,
+) (string, error) {
```
krancour, line 164:
```suggestion
func (f *Fake) GetCommitURL(repoURL string, sha string) (string, error) {
```

Project-specific, but we avoid unnecessary line breaks when we can. We start entertaining the idea when a line reaches 80 chars. 120 chars is the hard limit.

internal/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,25 @@ func (p *provider) ListPullRequests(
...
+// GetCommitURL implements gitprovider.Interface.
+func (p *provider) GetCommitURL(
+	repoURL string,
+	sha string,
+) (string, error) {
+	normalizedURL := git.NormalizeURL(repoURL)
+
+	parsedURL, err := url.Parse(normalizedURL)
+	if err != nil {
+		return "Unknown", fmt.Errorf("error processing repository URL: %s: %s", repoURL, err)
```
krancour, line 193: It would be more conventional to return an empty string when you're returning an error.

```suggestion
		return "", fmt.Errorf("error processing repository URL: %s: %s", repoURL, err)
```

Applies in a few other places as well.

internal/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,25 @@ func (p *provider) ListPullRequests(
 	return pts, nil
 }
 
+// GetCommitURL implements gitprovider.Interface.
+func (p *provider) GetCommitURL(
+	repoURL string,
+	sha string,
+) (string, error) {
```
krancour, line 188:
```suggestion
func (p *provider) GetCommitURL(repoURL string, sha string) (string, error) {
```

internal/gitprovider/gitea/gitea.go
```diff
@@ -257,6 +257,23 @@ func (p *provider) ListPullRequests(
 	return prs, nil
 }
 
+// GetCommitURL implements gitprovider.Interface.
+func (p *provider) GetCommitURL(
+	repoURL string,
+	sha string,
+) (string, error) {
```
krancour, line 264:
```suggestion
func (p *provider) GetCommitURL(repoURL string, sha string) (string, error) {
```

internal/gitprovider/gitea/gitea.go
```diff
@@ -257,6 +257,23 @@ func (p *provider) ListPullRequests(
...
+	normalizedURL := git.NormalizeURL(repoURL)
+
+	parsedURL, err := url.Parse(normalizedURL)
+	if err != nil {
+		return "Unknown", fmt.Errorf("error processing repository URL: %s: %s", repoURL, err)
```
krancour, line 269:
```suggestion
		return "", fmt.Errorf("error processing repository URL: %s: %s", repoURL, err)
```

internal/gitprovider/github/github.go
```diff
@@ -268,6 +268,23 @@ func (p *provider) ListPullRequests(
...
+// GetCommitURL implements gitprovider.Interface.
+func (p *provider) GetCommitURL(
+	repoURL string,
+	sha string,
+) (string, error) {
+	normalizedURL := git.NormalizeURL(repoURL)
+
+	parsedURL, err := url.Parse(normalizedURL)
+	if err != nil {
+		return "Unknown", fmt.Errorf("error processing repository URL: %s: %s", repoURL, err)
```
krancour, line 280:
```suggestion
		return "", fmt.Errorf("error processing repository URL: %s: %s", repoURL, err)
```

internal/gitprovider/gitlab/gitlab.go
```diff
@@ -199,6 +199,24 @@ func (p *provider) ListPullRequests(
...
+func (p *provider) GetCommitURL(
+	repoURL string,
+	sha string,
+) (string, error) {
+
+	normalizedURL := git.NormalizeURL(repoURL)
+
+	parsedURL, err := url.Parse(normalizedURL)
+	if err != nil {
+		return "Unknown", fmt.Errorf("error processing repository URL: %s: %s", repoURL, err)
```
krancour, line 212:
```suggestion
		return "", fmt.Errorf("error processing repository URL: %s: %s", repoURL, err)
```

internal/gitprovider/gitlab/gitlab_test.go
```diff
@@ -191,3 +191,36 @@ func TestParseGitLabURL(t *testing.T) {
 		})
 	}
 }
+
+func TestGetCommitURL(t *testing.T) {
+
```
krancour, line 196:
```suggestion
```

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -89,15 +98,21 @@ func (g *gitPushPusher) run(
 		return promotion.StepResult{Status: kargoapi.PromotionPhaseErrored},
 			fmt.Errorf("error loading working tree from %s: %w", cfg.Path, err)
 	}
+	repoURL := ""
+	if cfg.RepoURL == "" {
+		repoURL = workTree.URL()
+	} else {
+		repoURL = cfg.RepoURL
+	}
```
krancour, line 106:
```suggestion
```

krancour: See https://github.com/akuity/kargo/pull/3870/files#r2060819191

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -89,15 +98,21 @@ func (g *gitPushPusher) run(
...
 	creds, err := g.credsDB.Get(
 		ctx,
 		stepCtx.Project,
 		credentials.TypeGit,
-		workTree.URL(),
+		repoURL,
```
krancour, line 111:
```suggestion
		workTree.URL(),
```

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -89,15 +98,21 @@ func (g *gitPushPusher) run(
...
 	)
 	if err != nil {
 		return promotion.StepResult{Status: kargoapi.PromotionPhaseErrored},
-			fmt.Errorf("error getting credentials for %s: %w", workTree.URL(), err)
+			fmt.Errorf("error getting credentials for %s: %w", repoURL, err)
```
krancour, line 115:
```suggestion
			fmt.Errorf("error getting credentials for %s: %w", workTree.URL(),, err)
```

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -133,6 +148,9 @@ func (g *gitPushPusher) run(
 				fmt.Errorf("error getting current branch: %w", err)
 		}
 	}
+	if cfg.RepoURL != "" {
+		pushOpts.RepoURL = repoURL
+	}
```
krancour, line 153:
```suggestion
```

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -181,11 +199,25 @@ func (g *gitPushPusher) run(
 		return promotion.StepResult{Status: kargoapi.PromotionPhaseErrored},
 			fmt.Errorf("error getting last commit ID: %w", err)
 	}
+
+	// Using git provider to get commit url. Continuing even if provider
+	// cannot be implemented as the push will still have succeeded, we
+	// just can't construct the URL.
+	gitProvider, err := gitprovider.New(repoURL, nil)
```
krancour, line 206:
```suggestion
	gitProvider, err := gitprovider.New(workTree.URL(), nil)
```

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -181,11 +199,25 @@ func (g *gitPushPusher) run(
...
+	commitURL := ""
+	if err != nil {
+		commitURL = "Unknown"
+		logger.Info("unable to construct commit URL from %s: %s", repoURL, err)
```
krancour, line 210:
```suggestion
		logger.Info("unable to construct commit URL from %s: %s", workTree.URL(), err)
```

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -181,11 +199,25 @@ func (g *gitPushPusher) run(
...
+	} else {
+		commitURL, _ = gitProvider.GetCommitURL(repoURL, commitID)
```
krancour, line 212:
```suggestion
		workTree.URL(), _ = gitProvider.GetCommitURL(repoURL, commitID)
```

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -181,11 +199,25 @@ func (g *gitPushPusher) run(
...
+	}
+
 	return promotion.StepResult{
 		Status: kargoapi.PromotionPhaseSucceeded,
 		Output: map[string]any{
-			stateKeyBranch: pushOpts.TargetBranch,
-			stateKeyCommit: commitID,
+			stateKeyBranch:    pushOpts.TargetBranch,
+			stateKeyCommit:    commitID,
+			stateKeyCommitURL: commitURL,
```
krancour, line 220: Now that we're running into multiple outputs prefixed with "commit", it is probably better that we structure output like this:

```
branch: <branch>
commit:
  id: <sha>
  url: <commit url>
commitID: <sha> # Retained for backwards compatibility.
```

internal/promotion/runner/builtin/schemas/git-push-config.json
```diff
@@ -23,6 +23,10 @@
     "targetBranch": {
       "type": "string",
       "description": "The target branch to push to. Mutually exclusive with 'generateTargetBranch=true'. If neither of these is provided, the target branch will be the currently checked out branch."
+    },
+    "repoURL": {
+      "type": "string",
+      "description": "The URL to the remote repository. If this is left blank, the repoURL will be the remote used when the repository was cloned."
```
krancour, line 29:
```suggestion
```

krancour, line 29:
```suggestion
    }
```

krancour: See https://github.com/akuity/kargo/pull/3870/files#r2060819191

internal/gitprovider/azure/azure_test.go
```diff
@@ -91,3 +91,36 @@ func TestParseRepoURL(t *testing.T) {
...
+	}
+
+	for _, testCase := range testCases {
+		// call the code we are testing
+		g := provider{}
+		commitURL, err := g.GetCommitURL(testCase.url, testCase.sha)
+		require.NoError(t, err)
+		require.Equal(t, testCase.expectedURL, commitURL)
+	}
+}
```
krancour, line 126: Renaming test case fields adds a little clarity.

```suggestion
func TestGetCommitURL(t *testing.T) {
	testCases := []struct {
		repoURL           string
		sha               string
		expectedCommitURL string
	}{
		{
			repoURL:           "ssh://git@ssh.dev.azure.com/akuity/_git/kargo",
			sha:               "sha",
			expectedCommitURL: "https://dev.azure.com/akuity/_git/kargo/commit/sha",
		},
		{
			repoURL:           "git@ssh.dev.azure.com:v3/akuity/_git/kargo",
			sha:               "sha",
			expectedCommitURL: "https://dev.azure.com/akuity/_git/kargo/commit/sha",
		},
		{
			repoURL:           "http://dev.azure.com/akuity/_git/kargo",
			sha:               "sha",
			expectedCommitURL: "https://dev.azure.com/akuity/_git/kargo/commit/sha",
		},
	}

	for _, testCase := range testCases {
		// call the code we are testing
		g := provider{}
		commitURL, err := g.GetCommitURL(testCase.repoURL, testCase.sha)
		require.NoError(t, err)
		require.Equal(t, testCase.expectedCommitURL, commitURL)
	}
}
```

internal/gitprovider/gitea/gitea_test.go
```diff
@@ -383,3 +383,46 @@ func TestListPullRequests(t *testing.T) {
...
+	}
+
+	for _, testCase := range testCases {
+		// call the code we are testing
+		g := provider{}
+		commitURL, err := g.GetCommitURL(testCase.url, testCase.sha)
+		require.NoError(t, err)
+		require.Equal(t, testCase.expectedURL, commitURL)
+	}
+}
```
krancour, line 428:
```suggestion
func TestGetCommitURL(t *testing.T) {
	testCases := []struct {
		repoURL           string
		sha               string
		expectedCommitURL string
	}{
		{
			repoURL:           "http://gitea.com/akuity/kargo",
			sha:               "sha",
			expectedCommitURL: "https://gitea.com/akuity/kargo/commit/sha",
		},
		{
			repoURL:           "ssh://git@gitea.com/akuity/kargo",
			sha:               "sha",
			expectedCommitURL: "https://gitea.com/akuity/kargo/commit/sha",
		},
		{
			repoURL:           "git@gitea.com:akuity/kargo",
			sha:               "sha",
			expectedCommitURL: "https://gitea.com/akuity/kargo/commit/sha",
		},
		{
			repoURL:           "git@custom.host.com:akuity/kargo",
			sha:               "sha",
			expectedCommitURL: "https://custom.host.com/akuity/kargo/commit/sha",
		},
		{
			repoURL:           "http://custom.host.com/akuity/kargo",
			sha:               "sha",
			expectedCommitURL: "https://custom.host.com/akuity/kargo/commit/sha",
		},
	}

	for _, testCase := range testCases {
		// call the code we are testing
		g := provider{}
		commitURL, err := g.GetCommitURL(testCase.repoURL, testCase.sha)
		require.NoError(t, err)
		require.Equal(t, testCase.expectedCommitURL, commitURL)
	}
}
```

internal/gitprovider/github/github_test.go
```diff
@@ -365,3 +365,41 @@ func TestListPullRequests(t *testing.T) {
...
+	}
+
+	for _, testCase := range testCases {
+		// call the code we are testing
+		g := provider{}
+		commitURL, err := g.GetCommitURL(testCase.url, testCase.sha)
+		require.NoError(t, err)
+		require.Equal(t, testCase.expectedURL, commitURL)
+	}
+}
```
krancour, line 405:
```suggestion
func TestGetCommitURL(t *testing.T) {
	testCases := []struct {
		repoURL           string
		sha               string
		expectedCommitURL string
	}{
		{
			repoURL:           "ssh://git@github.com/akuity/kargo.git",
			sha:               "sha",
			expectedCommitURL: "https://github.com/akuity/kargo/commit/sha",
		},
		{
			repoURL:           "git@github.com:akuity/kargo.git",
			sha:               "sha",
			expectedCommitURL: "https://github.com/akuity/kargo/commit/sha",
		},
		{
			repoURL:           "https://username@github.com/akuity/kargo",
			sha:               "sha",
			expectedCommitURL: "https://github.com/akuity/kargo/commit/sha",
		},
		{
			repoURL:           "http://github.com/akuity/kargo.git",
			sha:               "sha",
			expectedCommitURL: "https://github.com/akuity/kargo/commit/sha",
		},
	}

	for _, testCase := range testCases {
		// call the code we are testing
		g := provider{}
		commitURL, err := g.GetCommitURL(testCase.repoURL, testCase.sha)
		require.NoError(t, err)
		require.Equal(t, testCase.expectedCommitURL, commitURL)
	}
}
```

internal/gitprovider/gitlab/gitlab_test.go
```diff
@@ -191,3 +191,36 @@ func TestParseGitLabURL(t *testing.T) {
...
+	}
+
+	for _, testCase := range testCases {
+		// call the code we are testing
+		g := provider{}
+		commitURL, err := g.GetCommitURL(testCase.url, testCase.sha)
+		require.NoError(t, err)
+		require.Equal(t, testCase.expectedURL, commitURL)
+	}
+}
```
krancour, line 226:
```suggestion
func TestGetCommitURL(t *testing.T) {
	testCases := []struct {
		repoURL         string
		sha         string
		expectedCommitURL string
	}{
		{
			repoURL:         "ssh://git@gitlab.com/akuity/kargo.git",
			sha:         "sha",
			expectedCommitURL: "https://gitlab.com/akuity/kargo/-/commit/sha",
		},
		{
			repoURL:         "git@gitlab.com:akuity/kargo.git",
			sha:         "sha",
			expectedCommitURL: "https://gitlab.com/akuity/kargo/-/commit/sha",
		},
		{
			repoURL:         "http://gitlab.com/akuity/kargo",
			sha:         "sha",
			expectedCommitURL: "https://gitlab.com/akuity/kargo/-/commit/sha",
		},
	}

	for _, testCase := range testCases {
		// call the code we are testing
		g := provider{}
		commitURL, err := g.GetCommitURL(testCase.repoURL, testCase.sha)
		require.NoError(t, err)
		require.Equal(t, testCase.expectedCommitURL, commitURL)
	}
}
```

internal/promotion/runner/builtin/git_pusher_test.go
krancour: Ok... I see what you're talking about now.

I think I can untangle this...

internal/gitprovider/gitprovider.go
BenHesketh21: Yes makes perfect sense. I've added the names for the inputs.

BenHesketh21:
@krancour I've completed the changes from your recent review. Am I right in saying I should also undo the RepoURL input option in the push function in `work_tree.go`?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
BenHesketh21: Sorry, coming back to this I'm seeing just how much I misunderstood this [comment](https://github.com/akuity/kargo/pull/3870#discussion_r2051068236). I'll add another input for the git push step for the _provider_, not the repoURL.

internal/promotion/runner/builtin/git_pusher_test.go
BenHesketh21: This is solved now there is the provider option on the git push step.

BenHesketh21:
@krancour Sorry for the delay, I've added the provider option similar to other steps that create a `gitprovider`, this solved the unit test issue. I believe (probably closer to hope) that this is ready for another review.

internal/promotion/runner/builtin/git_pusher.go
hiddeco: It appears like we now ended up with a `provider` field in the `git-push` configuration (from which we can now defer the provider and construct the URL in case of enterprise-hosted versions). However, the `RepoURL` option in `PushConfig` remains unused.

Is it safe to assume the latter can be removed?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -34,13 +34,16 @@ Stages that write to the same branch do not write to the same files.
...
 | `maxAttempts` | `int32` | N | The maximum number of attempts to make when pushing to the remote repository. Default is 50. |
 | `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of 'true' is mutually exclusive with `targetBranch`. If neither of these is provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
+| `provider` | `string` | The name of the Git provider to use. Currently 'azure', 'bitbucket', 'gitea', 'github', and 'gitlab' are supported. Kargo will try to infer the provider if it is not explicitly specified. |
```
krancour: This row appears to be missing the cell with a `Y` or `N` indicating whether a value is required for this field.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -34,13 +34,16 @@ Stages that write to the same branch do not write to the same files.
...
 
 ## Output
 
 | Name | Type | Description |
 |------|------|-------------|
 | `branch` | `string` | The name of the remote branch pushed to by this step. This is especially useful when the `generateTargetBranch=true` option has been used, in which case a subsequent [`git-open-pr`](git-open-pr.md) will typically reference this output to learn what branch to use as the head branch of a new pull request. |
 | `commit` | `string` | The ID (SHA) of the commit pushed by this step. |
+| `commitURL` | `string` | The URL to the commit that was pushed to the remote repository. |
```
krancour: Reads slightly better.

```suggestion
| `commitURL` | `string` | The URL of the commit that was pushed to the remote repository. |
```

internal/controller/git/work_tree.go
```diff
@@ -538,6 +538,9 @@ type PushOptions struct {
 	// be useful when pushing changes to a remote branch that has been updated
 	// in the time since the local branch was last pulled.
 	PullRebase bool
+	// RepoURL determines the URL of the remote repository to push to. If empty,
+	// origin will be used.
+	RepoURL string
```
krancour: I'm not clear on why this new field is needed.

internal/promotion/runner/builtin/git_pusher.go
```diff
@@ -181,11 +186,26 @@ func (g *gitPushPusher) run(
...
+	// just can't construct the URL.
+	gpOpts := gitprovider.Options{}
+	if cfg.Provider != nil {
+		gpOpts.Name = string(*cfg.Provider)
+	}
+	gitProvider, err := gitprovider.New(workTree.URL(), &gpOpts)
+	var commitURL string
+	if err == nil {
+		commitURL, _ = gitProvider.GetCommitURL(workTree.URL(), commitID)
+	}
```
krancour: I'm concerned about the potential this creates for commitURL to be inexplicably empty in the output with no evidence of anything having gone wrong that would account for it.

Can this attempt at resolving the provider be moved to _before_ the push so that we can then fail harder/faster and not obfuscate the problem?

hiddeco: I think failing hard here would be too aggressive. It's essentially metadata, and the goal of the step is to push, I would not want a step to fail if it can't produce metadata that's likely only required to sugarcoat other things.

hiddeco: What I can do to at least not hide the error in full is to log it if we can't obtain it from the getter.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -34,13 +34,16 @@ Stages that write to the same branch do not write to the same files.
...
 | `maxAttempts` | `int32` | N | The maximum number of attempts to make when pushing to the remote repository. Default is 50. |
 | `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of 'true' is mutually exclusive with `targetBranch`. If neither of these is provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
+| `provider` | `string` | N | The name of the Git provider to use. Currently 'azure', 'bitbucket', 'gitea', 'github', and 'gitlab' are supported. Kargo will try to infer the provider if it is not explicitly specified. |
```
krancour: We might need to explain how this is used, otherwise it feels surprising to (even optionally) be asked for this information. Under normal circumstances, I'd be thinking to myself that pushing to an established remote shouldn't require me to identify the provider. Explaining that it's used to construct a URL to a human-friendly representation of the commit might demystify things.

internal/promotion/runner/builtin/git_pusher.go
krancour: Logging it would be an improvement for sure. I understand not wanting to fail so hard if the push succeeds. My only fear is that we really don't know how dependent downstream steps may be on this info. It's not a hill I'd die on though.

BenHesketh21: This was the balance we were trying to reach, I opted to add the log message but the priority definitely makes sense to not block otherwise successful pushes

krancour (approved):

Issue #3823, opened 2025-04-08
RefreshWarehouse/RefreshStage should not require `patch` privileges


jessesuen:
# Proposed Feature

Refreshing a Warehouse or Stage currently requires `patch` privileges because it works by setting the annotation: `kargo.akuity.io/refresh`. We should provide a way to refresh a Warehouse / Stage but without giving patch privileges on the resource.

# Motivation

It may not be desireable to give an end user or a robot account the privileges to edit a Stage or Warehouse, but want to give them the ability to refresh the warehouse/stage in order to fast-track discovery of artifacts, or update when a PR is merged.

# Suggested Implementation

We have API endpoints for RefreshStage and RefreshWarehouse. We would relax the RBAC check but still proceed to annotating the Stage/Warehouse via `patch`. The question becomes, what privilege *do* we check in lieu of `patch`? I see two options:

### Option 1: if you have `get` privilege, you can refresh. 

This is the approach Argo CD took. This is quick and easy.

### Option 2: introduce a `refresh` verb on Stages/Warehouses.

Harder because it's another verb to manage. But more flexible. For example, we might configure GitHub with a robot account token that can only refresh stages/warehouses, but cannot read them. This would set us up for when we support [webhooks](https://github.com/akuity/kargo/issues/1449) by giving git/image providers the least amount of privilege necessary.

krancour:
Option 1 is nearly free because refresh endpoints get before they patch. All we need to do is bypass the authorizing wrapper around the k8s client and use the internal client directly when patching.

PR #3872, opened 2025-04-16
feat: bypass patch perms for refresh of stages and warehouses

fuskovic:
Closes https://github.com/akuity/kargo/issues/3823

This PR implements implementation option 1 stated in the issue.

By passing the internal client to these refresh methods, the auth wrapper client hooks don't run.

Also, didn't see any coverage for `internal/server/refresh_warehouse_v1alpha1.go` for so added a test file for this as well.

### DIFF
--- a/internal/server/refresh_stage_v1alpha1.go
+++ b/internal/server/refresh_stage_v1alpha1.go
@@ -31,7 +31,7 @@ func (s *server) RefreshStage(
 		Namespace: project,
 		Name:      name,
 	}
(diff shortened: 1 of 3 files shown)


krancour (approved):
🔥 Excellent work @fuskovic.

> Also, didn't see any coverage for internal/server/refresh_warehouse_v1alpha1.go for so added a test file for this as well.

Thank you very much for taking the initiative on that.

Issue #3670, opened 2025-03-18
Apply promotion policy to multiple stages


jessesuen:
# Proposed Feature

It would be useful to set a promotion policy based on a set of labels and/or naming patterns.

Today, a promotion policy is set against explicit names:

```yaml
apiVersion: kargo.akuity.io/v1alpha1
kind: Project
metadata:
  name: example
spec:
  promotionPolicies:
  - stage: test
    autoPromotionEnabled: true
  - stage: uat
    autoPromotionEnabled: true
```

Instead, we could possibly use selectors:

```yaml
spec:
  promotionPolicies:
  - stageSelector:
      matchLabels:
        autopromote: "true"
      matchExpressions:
        - { key: foo, operator: In, values: [bar] }
        - { key: env, operator: NotIn, values: [prod] }
```

or globs/regex:

```yaml
spec:
  promotionPolicies:
  - autoPromotionEnabled: true
    name: glob:prod-*
```

# Motivation

Currently, all the Stages need to be referened in the Project if you want auto promotion

# Suggested Implementation

NOTE: labels are mutable on resources, but names cannot be changed. It could mean that there is a use case for globs/regex vs. labels. So we might want both.

krancour:
See note about security considerations [here](https://github.com/akuity/kargo/issues/3535#issuecomment-2740719966)

Does _not_ mean we can't do this. I just don't want to lose track of this concern.

krancour:
@jessesuen what do you think about just making this a requirement in #3669? It's hard to imagine we'd tackle these two issues separately anyway.

PR #3875, opened 2025-04-16
feat: move Project configuration to ProjectConfig resource

### LINKED ISSUE #3669: Move Project configuration into a single namespace-scoped resource
# Proposed Feature

This is a proposal to introduce a new namespaced-scope CRD, `ProjectConfig`, where all Project configuration would live, including the current `promotionPolicies`. By moving configuration into a namespaced-scope resource, it would enable self-service and more granularity of who can edit project configurations (e.g. promotion policies, and future settings).

# Motivation

Currently, project configuration (currently only `promotionPolicies`) are stored in the `Project` CRD. The challenge is that because Projects are a cluster-scoped resource, only a cluster admin can edit projects. This makes it awkward (though not impossible) to grant access to edit a single Project.

We want a more fine-grained ability to control projects. e.g. there should be the ability for user A to edit project configuration for `project-a`, but not for `project-b`.

As I alluded to, this can _technically_ be solved with current kubernetes RBAC using ClusterRoles that specify resourceNames, e.g.:

```yaml
kind: ClusterRole
metadata:
  name: kargo-project-a-admin
rules:
- apiGroups: kargo.akuity.io
  kind: Project
  resourceNames:
  - project-a
``` 

However, the above approach is not ideal. The ability to create this ClusterRole and ClusterRoleBindings, in and of itself, requires cluster privileges. Many things would become simpler, and possible, if project configuration were namespaced scoped.

# Suggested Implementation

Whenever a `Project` is created, along with its corresponding `Namepsace`, we would immediately create a `ProjectConfig` in the project namespace, called `default`. This would be similar to how the `default` ServiceAccount is created in every namespace. The `default` ProjectConfig is where Kargo would look for all project configurations, such as promotion policies. 

We would then enhance our existing, `kargo-admin` Role, to have `update` `patch` privileges to the `default` `ProjectConfig`.

NOTE: Despite this being a namespaced scope resource, other `ProjectConfig` resources not named `default` would be ignored. There would never be a reason to give `create` or `delete` privileges on `ProjectConfig` resources since our controller would automate the creation of them, and deletion would only be needed when the Project/Namespace is deleted (already handled by cascaded deletion).

hiddeco:
Resolves #3669

This moves the `PromotionPolicy` configuration in `Project` resources to a dedicated (namespaced) `ProjectConfig` resource. The expectation is that there is only one `ProjectConfig` in each `Project` (namespace), with the same name as the `Project` itself — a validation webhook actively enforces this.

For backwards compatibility, the management controller will migrate any existing spec in a `Project` to a `ProjectConfig` resource as long as this resource does not exist, before `nil` writing the spec of the `Project`.

Updates to the spec of a `Project` other than this `nil` write are no longer accepted, and will be actively blocked by a validation webhook to force people to move to a `ProjectConfig`.


api/v1alpha1/project_config_types.go
```diff
@@ -0,0 +1,48 @@
...
+	// specific Stages within the Project.
+	PromotionPolicies []PromotionPolicy `json:"promotionPolicies,omitempty" protobuf:"bytes,1,rep,name=promotionPolicies"`
+}
+
+// PromotionPolicy defines policies governing the promotion of Freight to a
+// specific Stage.
+type PromotionPolicy struct {
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	Stage string `json:"stage" protobuf:"bytes,1,opt,name=stage"`
```
jessesuen, line 30: Are we happy with carrying over the existing spec as-is, or do we want to use this as an opportunity to modify the API to better accommodate feature requests like: https://github.com/akuity/kargo/issues/3670 (e.g should Stage remain a string or be a struct with a `name` field)

jessesuen:
As part of this PR, we should also update the `kargo-admin` and `kargo-viewer` roles to have privileges to this resource.

https://github.com/akuity/kargo/blob/4eff199c482f52f0be1ff5e9efe10122cb402471/internal/controller/management/projects/projects.go#L594

api/v1alpha1/project_config_types.go
hiddeco: My idea was to do something like the following:

```go
// PromotionPolicy defines policies governing the promotion of Freight to a
// specific Stage.
type PromotionPolicy struct {
	// Stage is the name of the Stage to which this policy applies.
	//
	// Deprecated: Use StageSelector instead.
	//
	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
	Stage string `json:"stage" protobuf:"bytes,1,opt,name=stage"`
	// StageSelector is a selector that matches the Stage resource to which
	// this policy applies.
	StageSelector *PromotionPolicySelector `json:"stageSelector,omitempty" protobuf:"bytes,3,opt,name=stageSelector"`
	// AutoPromotionEnabled indicates whether new Freight can automatically be
	// promoted into the Stage referenced by the Stage field. Note: There are may
	// be other conditions also required for an auto-promotion to occur. This
	// field defaults to false, but is commonly set to true for Stages that
	// subscribe to Warehouses instead of other, upstream Stages. This allows
	// users to define Stages that are automatically updated as soon as new
	// artifacts are detected.
	AutoPromotionEnabled bool `json:"autoPromotionEnabled,omitempty" protobuf:"varint,2,opt,name=autoPromotionEnabled"`
}

type PromotionPolicySelector struct {
	// Name is the name of the resource to which this policy applies.
	//
	// It can be used to select a specific resource by name, or to select a
	// resource based on a regex (e.g. `regex:dev-.*`) or glob (e.g. `glob:dev-*`)
	// pattern.
	//
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,4,opt,name=name"`
	// MatchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels
	// map is equivalent to an element of MatchExpressions, whose key field is "key", the
	// operator is "In", and the values array contains only "value". The requirements are ANDed.
	// +optional
	MatchLabels map[string]string `json:"matchLabels,omitempty" protobuf:"bytes,2,rep,name=matchLabels" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
	// matchExpressions is a list of label selector requirements. The requirements are ANDed.
	// +optional
	// +listType=atomic
	MatchExpressions []metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty" protobuf:"bytes,3,rep,name=matchExpressions"`
}
```

hiddeco: See (last bunch of commits in): #3886

krancour (approved):
Needs docs, but otherwise LGTM. Docs can be a follow-up.

hiddeco:
@krancour I added a commit that does the bare minimum of updating the existing documentation. I will make more fine-grain improvements in #3886.

Issue #3750, opened 2025-03-28
Kubelogin Support in the Kargo Controller


sbyrne-teck:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Can the kubelogin binary be installed into the kargo container so that its possible to use a AD managed service principal in the kubeconfig for the sharded Kargo configuration. 

# Motivation

On Azure Kubernetes the security is integrated with Azure Entra ID. Instead of using local service account users we make use of Azure AD managed identities. 

# Suggested Implementation

kubelogin binary added to the Docker image. 
https://github.com/Azure/kubelogin

hiddeco:
Think instead of accumulating binaries in our image that can become outdated, the better option is probably to (allow to) facilitate this through an init container.

sbyrne-teck:
Fair point. Thanks @hiddeco. Will solution it on our side instead.

IvanJosipovic:
I do agree with the sentiment of using Init Containers. However, it might be worth including the 3 Hyperscalers (Azure, AWS, GCP) in the base Kargo image.

Additionally, Init Containers do not appear to be supported.

PR #3879, opened 2025-04-16
chore: Adding init container support for the Controller in Helm

sbyrne13:
There is currently no capability to add custom init container code into the controller.
This is particularly useful for things like adding a binary or tool that is necessary for cloud compatibility.

Also adding volume and volume mount support too so that we can share installed binaries to the controller. 

Some background here - https://github.com/akuity/kargo/issues/3750


krancour:
@sbyrne13 codegen needs to be run after modifications to `values.yaml`. It's how the chart's readme gets updated based on metadata in that file.

sbyrne-teck:
Ah okay - thanks @krancour - done!

krancour:
This LGTM, but requesting @hiddeco to voice any concerns if he has any.

fwiw, I momentarily contemplated whether this capability should be added to all the components, but I think the controller is the only one where it is likely needed, so I'm ok with doing just the one.

sbyrne13:
@krancour - have updated this PR to also include volumes and volumeMounts after some more testing. Please re-review when you have some time. Thanks

charts/kargo/values.yaml
```diff
@@ -381,6 +381,13 @@ controller:
   podLabels: {}
   ## @param controller.podAnnotations Optional annotations to add to pods. Merges with `global.podAnnotations`, allowing you to override or add to the global annotations.
   podAnnotations: {}
+  ## @param controller.initContainers Optional init containers to add to the controller pods. This is rendered as the literal yaml.
```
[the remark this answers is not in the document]
sbyrne-teck: Done

charts/kargo/values.yaml
```diff
@@ -467,6 +474,10 @@ controller:
     # requests:
     #   cpu: 100m
     #   memory: 128Mi
+  ## @param controller.volumes Volumes for the Controller Pods.
+  volumes: []
+  ## @param controller.volumeMounts Volume Mounts for the Controller Pods.
+  volumeMounts: []
```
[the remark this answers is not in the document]
sbyrne-teck: done

pkg/x/promotion/runner/builtin/zz_config_types.go
sbyrne-teck, line 1: @hiddeco & @krancour  - this file was generated when i ran the codegen to populate the README. Not sure if I was supposed to push it here but the codegen ci job failed so not sure?

hiddeco: Some dependency we hadn't pinned started producing not-so-valid things due to an upstream bug. If you rebase once #3944 has landed in `main`, and re-generate, the issue should be gone.

sbyrne-teck: thanks hiddeco - file wasn't generated with rebase.

hiddeco (approved):
Thank you @sbyrne13 🙇

Issue #3450, opened 2025-02-06
provide PR link in `git-open-pr` promotion step output


Marvin9:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

Pull request deep links are not working - https://github.com/akuity/kargo/issues/3024#issuecomment-2639199830

With promotion steps, UI recently changed the way it structures Pull request deep link - https://github.com/akuity/kargo/issues/3024.

It constructs URL from `git-open-pr` step output and config values (`repoURL` from step config and `prNumber` from output). However when the spec contains expression language, it fails to generate URL because now `repoURL` is not raw value. 

The solution is to provide as much as info in promotion steps output (in this case, `repoURL`), it can be either full URL or at the very least, values that will help UI to construct link.

# Screenshots

https://github.com/akuity/kargo/issues/3024#issuecomment-2639199830

# Steps to Reproduce

- Create stage with promotion step `git-open-pr` and promote to that stage
- check the promotion lists in stage details page and open the promotion step modal. Both won't have PR deep link

# Version

```
dev
```

hiddeco:
The benefit of a direct URL to the pull request instead of just the repository would be that we no longer have to reconstruct it in the UI.

Marvin9:
+1 to @hiddeco would be great to have direct PR link so UI don't have to reconstruct for different git providers

Marvin9:
UI needs to use this new field

PR #3884, opened 2025-04-17
feat(controller): provide repoURL in git-open-pr step's output

fykaa:
fixes https://github.com/akuity/kargo/issues/3450


internal/promotion/runner/builtin/git_pr_opener.go
```diff
@@ -85,6 +89,7 @@ func (g *gitPROpener) run(
 			Status: kargoapi.PromotionPhaseSucceeded,
 			Output: map[string]any{
 				stateKeyPRNumber: prNumber,
+				stateKeyRepoURL:  cfg.RepoURL,
```
krancour, line 92: Now that the output is growing, we could probably be more future-proof by doing this:

```
	if prNumber != -1 {
		// We can don't need to check for the existence of prior output from this
		// step because we'd not have gotten to here if it didn't exist.
		output, _ := stepCtx.SharedState.Get(stepCtx.Alias)
		return promotion.StepResult{
			Status: kargoapi.PromotionPhaseSucceeded,
			// Don't need to validate the type assertion on the output here because we
			// we'd not have gotten to here if it wasn't a map[string]any.
			Output: output.(map[string]any), //nolint: forcetypeassert
		}, nil
	}
```

krancour: Also note that this short-circuiting logic is probably not required anymore.

GitHub doesn't let me comment on lines too far above this change, but would you please modify the comment about short-circuiting on (new) line 80 like so:

```
	// Short-circuit if shared state has output from a previous execution of this
	// step that contains a PR number.
	//
	// TODO(krancour): This short-circuiting logic may have outlived its
	// usefulness now that we keep track of the current step.
```

internal/promotion/runner/builtin/git_pr_opener.go
```diff
@@ -159,6 +164,7 @@ func (g *gitPROpener) run(
 			Status: kargoapi.PromotionPhaseSucceeded,
 			Output: map[string]any{
 				stateKeyPRNumber: pr.Number,
+				stateKeyRepoURL:  cfg.RepoURL,
```
krancour, line 167: This isn't the right information to add. You want `pr.URL` and the symbol and value of the key should probably be changed as well on (new) line 29:

```
	stateKeyPRURL = "prURL"
```

internal/promotion/runner/builtin/git_pr_opener.go
```diff
@@ -223,6 +229,7 @@ func (g *gitPROpener) run(
 		Status: kargoapi.PromotionPhaseSucceeded,
 		Output: map[string]any{
 			stateKeyPRNumber: pr.Number,
+			stateKeyRepoURL:  cfg.RepoURL,
```
krancour, line 232: Same as [here](https://github.com/akuity/kargo/pull/3884/files#r2049510891)

internal/promotion/runner/builtin/git_pr_opener.go
krancour: Ah. I see this was already discussed in #3450.

hiddeco, line 167: May I suggest that we use this opportunity to stop calling it a "pr number" and start calling it an ID? I.e., I feel this might be slightly easier to use:

```yaml
output:
  pr:
    id: 1234
    url: https://github.com/org/example/pull/1234
```

krancour: Good thought.

It is a breaking change, so @fykaa docs will need to be updated accordingly. `getPRNumber()` will also need to be modified to look at `pr.id` first and then fall back on `prNumber`, _unless_ we decide to just do away with the [short-circuiting](https://github.com/akuity/kargo/pull/3884#discussion_r2049506582) this step does, in which case, `getPRNumber()` is code we can jettison.

@hiddeco any thoughts on that?

krancour: Although... this is a somewhat nasty breaking change. Anyone _using_ that output in an expression would have to rewrite expressions to look for the new output structure and then fall back on the old one. That seems incredibly onerous.

We could either deprecate without ever removing or live with the way things are. 

@hiddeco?

internal/promotion/runner/builtin/git_pr_opener.go
```diff
@@ -75,17 +79,23 @@ func (g *gitPROpener) run(
...
+	//
+	// TODO(krancour): This short-circuiting logic may have outlived its
+	// usefulness now that we keep track of the current step.
 	prNumber, err := g.getPRNumber(stepCtx, stepCtx.SharedState)
 	if err != nil {
 		return promotion.StepResult{Status: kargoapi.PromotionPhaseErrored},
 			fmt.Errorf("error getting PR number from shared state: %w", err)
 	}
 	if prNumber != -1 {
+		// We can don't need to check for the existence of prior output from this
```
krancour, line 91:
```suggestion
		// We don't need to check for the existence of prior output from this
```

krancour:
Thanks @fykaa. This is in very good shape now.

Let's wait to hear from @hiddeco on a few things before merging this. I'd really like to dump the short-circuiting logic and reach consensus on what the output will look like.

internal/promotion/runner/builtin/git_pr_opener.go
hiddeco: My idea was to maintain the `prNumber` field for some time, and remove it at a later moment.

internal/promotion/runner/builtin/git_pr_opener.go
```diff
@@ -75,17 +75,23 @@ func (g *gitPROpener) run(
 ) (promotion.StepResult, error) {
 	// Short-circuit if shared state has output from a previous execution of this
 	// step that contains a PR number.
+	//
+	// TODO(krancour): This short-circuiting logic may have outlived its
+	// usefulness now that we keep track of the current step.
```
krancour, line 80: @fykaa after further consideration and a conversation with @hiddeco, I would indeed like to see this short-circuiting logic removed.

docs/docs/80-release-notes/100-deprecations.md
```diff
@@ -15,8 +15,9 @@ The table below summarizes features that have been deprecated and either removed
 
 | Feature | Deprecated In | Removed In | Replacement/Notes |
 |---------|---------------|------------|-------------------|
+| `prNumber` of `git-open-pr` Promotion Steps | v1.5.0 | Scheduled for v1.6.0 | Users should migrate to using `pr.id` for referencing pull request numbers. |
```
krancour, line 18:
```suggestion
| `prNumber` of `git-open-pr` Promotion Steps | v1.5.0 | Scheduled for v1.7.0 | Users should migrate to using `pr.id` for referencing pull request numbers. |
```

internal/promotion/runner/builtin/git_pr_opener.go
```diff
@@ -222,50 +211,15 @@ func (g *gitPROpener) run(
 	return promotion.StepResult{
 		Status: kargoapi.PromotionPhaseSucceeded,
 		Output: map[string]any{
-			stateKeyPRNumber: pr.Number,
+			"pr": map[string]any{
+				"id":  pr.Number,
+				"url": pr.URL,
+			},
+			stateKeyPRNumber: pr.Number, // Backward compatibility, this field will be deprecated in the future
```
krancour, line 218:
```suggestion
			stateKeyPRNumber: pr.Number, // TODO: Remove in v1.7.0
```

internal/promotion/runner/builtin/git_pr_opener_test.go
```diff
@@ -206,10 +210,17 @@ func Test_gitPROpener_run(t *testing.T) {
 		},
 	)
 	require.NoError(t, err)
+	// Validate backward compatibility with prNumber
```
krancour, line 213:
```suggestion
	// Validate backward compatibility with prNumber
	// TODO: Remove in v1.7.0
```

PR #3886, opened 2025-04-17
feat: add name and label selectors to PromotionPolicy

hiddeco:
Depends on: #3875 
Resolves: #3670


api/v1alpha1/project_config_types.go
```diff
@@ -0,0 +1,71 @@
...
+	// regex matching (with prefix "regex:"), or glob pattern matching (with
+	// prefix "glob:").
+	//
+	// +optional
+	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+
+	// LabelSelector is a selector that matches the resource to which this policy
+	// applies.
+	*metav1.LabelSelector `json:",inline" protobuf:"bytes,2,opt,name=labelSelector"`
+}
```
jessesuen, line 62: I think name, regex, glob, and label selector will cover all our bases. Great!

krancour: We've discussed before that a specific name rather than regex, glob, or label selectors is the most secure option because it eliminates the possibility of someone who can create/or edit Stages enabling auto-promotion for a Stage that wasn't intended to have it.

Should we be making a note of this in a comment here since however we document the fields here is what ends up in the CRD and generated CRD docs?

api/v1alpha1/project_config_types.go
```diff
@@ -38,6 +46,35 @@ type PromotionPolicy struct {
...
+
+	// LabelSelector is a selector that matches the resource to which this policy
+	// applies.
+	//
+	// NOTE: Using label selectors introduces security risks as users with
+	// appropriate permissions could create new resources with labels that match
+	// the selector, potentially enabling unauthorized auto-promotion.
+	// For sensitive environments, exact Name matching provides tighter control.
+	*metav1.LabelSelector `json:",inline" protobuf:"bytes,2,opt,name=labelSelector"`
+}
```
krancour, line 76: Can we update the comments on the `Name` and `LabelSelector` fields to clarify matters of precedence? Does one take precedence over the other? Are the logically OR'ed?

krancour (approved):

api/v1alpha1/project_config_types.go
hiddeco: They are ANDed.

PR #3888, opened 2025-04-18
feat(controller): collect stats in project status

### LINKED ISSUE #3812: aggregate details to create project "snapshot" in project status subresource
To support #3684, we need to "roll up" certain project stats into the Project CRD's status so that a high-level "snapshot" of each Project is included when listing Projects. These probably include, but aren't necessarily limited to, counts of things like:

* Total number of Warehouses
* Healthy / erroring Warehouses
* Total number of Stages
* Healthy / unhealthy Stages

@rpelczar @Marvin9 please feel free to add here exactly what you want available.

krancour:
Fixes #3812


hiddeco:
After more discussion offline, the event-driven approach seems reasonable to start with. While we're not entirely certain about the frequency of Warehouse and Stage status changes (especially with auto-promotion scenarios), I agree it's a good starting point.

We did discuss several alternatives:

1. Building in-memory metrics updated by Stage/Warehouse reconcilers, then having Project reconcilers "sink" those metrics
2. Having Stage/Warehouse reconcilers putting stats directly into Prometheus metrics
3. Fixed interval collection ([my original suggestion](https://github.com/akuity/kargo/pull/3888#issuecomment-2821185887))

I'm comfortable proceeding with the current implementation. If we encounter performance issues with large Projects or simultaneous status changes causing reconciliation spikes, we have these alternatives ready for further exploration.

rpelczar:
I don't know if we want to show all this data in the Project Tile, but I think I can easily display a healthy stages status if all stages are healthy, same for the warehouse.

Can we also add a number of stages and warehouses? Or do you want ui to sum numbers from all statuses?

Is there anything else you would like to see in the project tile? I have no more ideas.

krancour:
@rpelczar I mentioned the other day, but it may have gotten lost in a sea of other discussions, the stats I'm collecting right now were basically a placeholder. I can give you anything you'd like that supports your vision for the redesigned Project list page.

The actual design is much better left to you, but fwiw I'd been imagining something like this on each item in the Project list. I'm not sure if you're doing cards or a table. With cards, I could imagine this at the bottom of each card. With a table, I could imagine this in smaller text under the Project name in each row.

```
Warehouses: 💚 1/2          Stages:     💚 4/5
```

For something like this, I'd give you total warehouses, healthy warehouses, total stages, and healthy stages.

hiddeco (approved):
Thanks for following up on all my suggestions. This looks great! 🥇

krancour:
> Thanks for following up on all my suggestions.

And thank _you_ for all the sanity checks.

PR #3905, opened 2025-04-23
fix(controller): deal with the possibility of an App having a nil status field when we patch it

krancour:


jessesuen (approved):

internal/promotion/runner/builtin/argocd_updater.go
```diff
@@ -501,9 +501,19 @@ func (a *argocdUpdater) syncApplication(
...
-			src.Object["status"].(map[string]any)["operationState"]
+		// We've once encountered an occasion where the unstructured representation
+		// of the destination App was missing the status field because it had never
+		// yet been reconciled (Application controller was not yet running), so we
+		// are completely bailing on this hack if we find this to be the case. We
+		// check the source object too for good measure, although that should not be
+		// prone to the same problem.
+		_, dstHasStatus := dst.Object["status"]
+		_, srcHasStatus := src.Object["status"]
+		if !dstHasStatus || !srcHasStatus {
```
hiddeco, line 512: Shouldn't this be: `dstHasStatus && srcHasStatus`?

krancour: 🤦‍♂️

PR #3921, opened 2025-04-24
chore(deps/tools): update `golangci-lint` to v2

hiddeco:


hiddeco:
@krancour ping to get this on your radar again. This should _finally_ be good to go.

krancour:
I can't spot why we need to disregard staticcheck so frequently.

hiddeco:
@krancour it is because we sometimes start error messages with an uppercase (because the object kind is a name), while Go recommends to have errors start with a lowercase.

krancour:
Ah. I'm aware, but I failed to recognize that as the common element in most spots where staticcheck was ignored.

LGTM

PR #3927, opened 2025-04-25
feat(runner): add `untar` Promotion step

### LINKED ISSUE #3926: Add untar promotion step
# Checklist

# Proposed Feature

A new untar promotion step for extracting tgz.

# Motivation

We heavily use helm charts, with umbrella charts and dependencies. But we want to keep the dependent charts as raw helm templates and not the tgz.
And the `helm dependency update` as no `--untar` flag see https://github.com/helm/helm/issues/8479

So we came with a plan for our promotion steps:
- helm-update-chart
- untar (the dependency)
- delete  (the tgz archive)

But obviously this can be use for any other extracting needs 😄 

# Suggested Implementation

I'll join a pull request for implementation...

AlexDouze:
close #3926


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/untar.md
```diff
@@ -0,0 +1,102 @@
...
+
+# `untar`
+
+`untar` extracts the contents of a tar file (including gzipped tar files) to a specified location. It automatically detects if the file is gzipped and handles it accordingly.
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `filePath` | `string` | Y | Path to the tar file to extract. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
```
krancour, line 14: `inPath` feels as if it would be more "symmetric" with `outPath`. The `copy` step has already used `inPath` along with `outPath`. I think making that change here would help keep things more consistent overall.

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,255 @@
...
+	} else {
+		// File is not gzipped, reset to beginning of file and try as regular tar
+		_, err = file.Seek(0, 0)
+		if err != nil {
+			return promotion.StepResult{Status: kargoapi.PromotionPhaseErrored},
+				fmt.Errorf("failed to seek in tar file: %w", err)
+		}
+		tarReader = tar.NewReader(file)
+		logger.Debug("treating file as regular tar")
+	}
```
hiddeco, line 121: I think that instead of "brute-forcing" this, the better way would be to check if the first two bytes of the file contain the "magic numbers" `0x1F 0x8B` ([see ID1 and ID2 from the RFC](https://www.rfc-editor.org/rfc/rfc1952#page-5)).

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+
+		// Skip if this is not a regular file, symlink, or directory
+		if header.Typeflag != tar.TypeReg && header.Typeflag != tar.TypeDir && header.Typeflag != tar.TypeSymlink {
+			logger.Debug("skipping non-regular file", "path", header.Name, "type", header.Typeflag)
+			continue
+		}
+
+		// Handle stripping components if specified
+		targetName := header.Name
```
hiddeco, line 149: Further validation is required to ensure it does not attempt to escape the working directory, which can happen with a maliciously crafted compressed artifact. E.g. something like:

```
func validRelPath(p string) bool {
	if p == "" || strings.Contains(p, `\`) || strings.HasPrefix(p, "/") || strings.Contains(p, "../") {
		return false
	}
	return true
}
```

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,255 @@
...
+		case tar.TypeDir:
+			// Create directory
+			if err := os.MkdirAll(targetPath, 0755); err != nil {
+				return promotion.StepResult{Status: kargoapi.PromotionPhaseErrored},
+					fmt.Errorf("failed to create directory %s: %w", targetPath, err)
+			}
+
+		case tar.TypeSymlink:
+			// Create symlink
+			if err := os.Symlink(header.Linkname, targetPath); err != nil && !os.IsExist(err) {
```
hiddeco, line 200: We should (again) ensure that a symlink never attempts to escape outside the working directory.

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,255 @@
...
+	// Extract the tar file
+	stripComponents := int64(0)
+	if cfg.StripComponents != nil {
+		stripComponents = *cfg.StripComponents
+		if stripComponents > 0 {
+			logger.Debug("stripping path components", "count", stripComponents)
+		}
+	}
+
+	for {
```
hiddeco, line 132: We should ensure we cap the maximum size we allow to be extracted from the compressed artifact. This is something which was [recently fixed in Helm](https://github.com/helm/helm/commit/e4da49785aa6e6ee2b86efd5dd9e43400318262b), and is generally a [good practice to apply](https://github.com/fluxcd/pkg/blob/main/tar/tar.go#L28-L29) when multi-tenancy is handled within the same container.

AlexDouze:
👋  @hiddeco
I started to work from the release-1.4 branch. From what I understand i should rather start from main.
And it might be in the v1.6.X release.

Am I rigth?
Thanks

krancour:
> From what I understand i should rather start from main.

Always start from main.

If something needs to be included in a patch release to an existing minor release, we selectively cherry-pick from main.

Everything goes to main first.

The good news is that I don't think there would be any merge conflicts in this case if you cherry-picked your commits from this branch into a new branch based on main. You can even force push that new branch to this one and everything else should just sort itself out.

internal/webhook/project/webhook.go
```diff
@@ -51,7 +51,8 @@ type webhook struct {
 
 func SetupWebhookWithManager(mgr ctrl.Manager, cfg WebhookConfig) error {
 	w := &webhook{
-		cfg: cfg,
+		cfg:    cfg,
+		client: mgr.GetClient(),
```
AlexDouze, line 55: I had nil pointer without that 😄 
And I was enable to create a project....

internal/promotion/runner/builtin/tar_extractor_test.go
```diff
@@ -0,0 +1,509 @@
...
+				&promotion.StepContext{WorkDir: workDir},
+				tt.cfg,
+			)
+			tt.assertions(t, workDir, result, err)
+		})
+	}
+}
+
+// Helper function to create an int64 pointer
+func intPtr(i int64) *int64 {
```
hiddeco, line 507: You can use `ptr.To` from `k8s.io/utils/ptr`.

internal/promotion/runner/builtin/schemas/untar-config.json
```diff
@@ -0,0 +1,29 @@
...
+      "description": "StripComponents is the number of leading components to strip from file names in the archive.",
+      "minimum": 0
+    },
+    "ignore": {
+      "type": "string",
+      "description": "Ignore is a (multiline) string of glob patterns to ignore when extracting files. It accepts the same syntax as .gitignore files.",
+      "minLength": 1
+    }
+  }
+}
```
hiddeco, line 29:
```suggestion
}

```

internal/promotion/runner/builtin/tar_extractor_test.go
```diff
@@ -0,0 +1,509 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/pkg/promotion"
+	"github.com/akuity/kargo/pkg/x/promotion/runner/builtin"
+)
+
+func Test_tarExtractor_run(t *testing.T) {
```
hiddeco, line 20: It would be nice to have a test case for files that are too big as well. I think you can test this doing something like:

```go
        // Mock large file
        largeFileSize := MaxDecompressedFileSize + 1
        hdr := &tar.Header{
            Name: "large_file.bin",
            Mode: 0600,
            Size: largeFileSize,
        }
        require.NoError(t, tw.WriteHeader(hdr))
        
        // Write zeros to the file
        _, err = tw.Write(make([]byte, largeFileSize))
        require.NoError(t, err)
```

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,336 @@
...
+	if header[0] == 0x1F && header[1] == 0x8B {
+		// File is gzipped
+		gzr, err := gzip.NewReader(file)
+		if err != nil {
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("failed to create gzip reader: %w", err)
+		}
+		defer gzr.Close()
+		tarReader = tar.NewReader(gzr)
+		logger.Debug("treating file as gzipped tar based on magic numbers")
```
hiddeco, line 138: (All of) these logs are too verbose for debug level, and we do not tend to log this much information. If we keep them around, they should at least be at `Trace` level.

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,336 @@
...
+		destDir := filepath.Dir(targetPath)
+		if err := os.MkdirAll(destDir, 0755); err != nil {
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("failed to create directory %s: %w", destDir, err)
+		}
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			// Create directory
+			if err := os.MkdirAll(targetPath, 0755); err != nil {
```
hiddeco, line 232: `0o755` can probably be moved to a `const` (and then also be used above). (See also comment below about creating files). I also think we might want to use `0o750` instead.

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,336 @@
...
+
+	// Reset to beginning of file
+	_, err = file.Seek(0, 0)
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("failed to seek in tar file: %w", err)
+	}
+
+	// Check for gzip magic numbers (0x1F 0x8B)
+	if header[0] == 0x1F && header[1] == 0x8B {
```
hiddeco, line 129: Might be worth moving these to constants, e.g.

```go
const (
    gzipID1 = 0x1F
    gzipID2 = 0x8B
)
```

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+
+		destDir := filepath.Dir(targetPath)
+		if err := os.MkdirAll(destDir, 0755); err != nil {
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("failed to create directory %s: %w", destDir, err)
+		}
+
+		switch header.Typeflag {
+		case tar.TypeDir:
```
hiddeco, line 230: [Reading an upstream Go implementation](https://cs.opensource.google/go/x/build/+/e235f28d:internal/untar/untar.go;l=69-79) of this again makes me wonder if we can assume directory entries always go before file entries. I.e., if we can run into file paths for which the parent directory does not exist yet.

internal/promotion/runner/builtin/tar_extractor.go
```diff
@@ -0,0 +1,336 @@
...
+				return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+					fmt.Errorf("extraction aborted: total size would exceed limit of %d bytes", MaxDecompressedTarSize)
+			}
+
+			// Create file
+			outFile, err := os.Create(targetPath)
+			if err != nil {
+				return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+					fmt.Errorf("failed to create file %s: %w", targetPath, err)
+			}
```
hiddeco, line 275: The `os.Create` and `os.Chmod` can be combined as:

```go
os.OpenFile(targetPath, O_RDWR|O_CREATE|O_TRUNC, <permissions>)
```

I also do wonder if we should copy the modification data, as done here: https://cs.opensource.google/go/x/build/+/e235f28d:internal/untar/untar.go;l=114-123

AlexDouze:
@hiddeco should I push further on testing?

hiddeco:
Sorry for the wait, I was pretty busy working on `v1.5.1` things before I had the space to follow up on new `v1.6.0` features.

As I kept you waiting for quite some time, I addressed some things that came to mind without doing another round of nitpicking, but I was unable to push them to your branch. If you could have a look at the following commits (and maybe cherry-pick them, if you agree), that would be perfect:

- https://github.com/akuity/kargo/commit/818dc7e4aad8abb82313f307cfe7dbcee8feeefb
- https://github.com/akuity/kargo/commit/edcc0e7b17256b07eb9c44389904097587541b9e

For the last commit, I think an update to the documentation (both `untar-config.json` and `untar.md`) would be required, to highlight the `outPath` is _replaced_ if it already exists.

hiddeco (approved):
This looks good to me, thanks for the contribution and patience with getting this right! 🌷 

It just needs some conflict resolving before I can merge.

hiddeco:
Code generation still has to be run and committed, as it is detecting divergences.

Issue #3770, opened 2025-04-02
UI: Pipeline view revamp


Marvin9:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Revamp the complete pipeline view. Right now it is good enough but becomes information overdose on scale. This feature aims to solve those issues from scratch so its useful for simple to advanced use cases.

The design revamp includes everything in Pipeline view and Promotion UX. That means Freight timeline, DAG view, list view _(new)_ and Promote stage flow.

Find the low fidelity wireframes below to know more about incoming changes. This issue will be split into the subtasks and in individual issues by sections.

Some notable incoming changes

1. Warehouse filter that can be used to separate by "Pipelines" 
2. Space for Promotion flow
3. Hide downstream stages when number of stages scales
4. Stages List view with useful filters
5. Filters in freight timeline for quick freight searches
6. Freight timeline with option to compact/expand

# Suggested Implementation

![Image](https://github.com/user-attachments/assets/1ed6124c-db90-4cc8-8640-42914cbb77d6)

![Image](https://github.com/user-attachments/assets/c1fbbd47-77bb-4340-8e56-26712b5d8234)

![Image](https://github.com/user-attachments/assets/f65e9b0a-1695-44a7-9878-04996f88dba6)

![Image](https://github.com/user-attachments/assets/32447316-9a3d-449e-8bc3-8eb0a0f4afd0)

![Image](https://github.com/user-attachments/assets/5d917dd4-f616-4dfe-adf7-2b433bb28679)

![Image](https://github.com/user-attachments/assets/70b80e2d-f821-4e71-be7d-963837c5e51c)

hiddeco:
One comment I received more than one time from people approaching us at the booth is that their developers would prefer version metadata (e.g., a Git commit SHA or image tag) to be more prominently visible instead on each "Stage" instead of the alias we show at the moment.

Reason they gave was that these versions are something they already know, and the alias is something they have to "learn" by looking at other parts of the UI.

apenney:
> One comment I received more than one time from people approaching us at the booth is that their developers would prefer version metadata (e.g., a Git commit SHA or image tag) to be more prominently visible instead on each "Stage" instead of the alias we show at the moment.
> 
[...]

PR #3941, opened 2025-04-28
feat(ui): pipeline view revamp

Marvin9:
fixes https://github.com/akuity/kargo/issues/3770

Screenshot 2025-04-28 at 3 44 57 PM

Screenshot 2025-04-28 at 3 45 31 PM

Screenshot 2025-04-28 at 3 46 14 PM

Screenshot 2025-04-28 at 3 54 07 PM (2)


ui/src/pages/project.tsx
```diff
@@ -69,8 +84,23 @@ export const Project = ({
...
+              }
+            : undefined
+        }
         creatingStage={creatingStage}
         creatingWarehouse={creatingWarehouse}
+        warehouseName={warehouseName}
+        freightName={freightName}
+        warehouses={listWarehousesQuery.data?.warehouses || []}
+        freights={getFreightQuery.data?.groups || {}}
+        refetchFreights={getFreightQuery.refetch}
```
rpelczar, line 103: What's the point of passing all those params? Let's get url params directly in the component, and fetch data there too. We don't need a loading state here.

Marvin9: Its either pollute pipelines or project so I chose project due to the fact that it would be easy to split from here for pages in future if we ever want. I kept loading state here because I want data ready before it renders for details

rpelczar: So now we have two loaders, one here and the second one in the pipeline. It doesn't look good.
If there is no need to have that data now in the project component, please move everything to the pipeline component.

Marvin9: I can move stage loader up to project but the sole reason of having those in project component is to have data ready pipeline component on first attempt

rpelczar: I checked this file one more time, and I don't see any reason to pass 11 params if we can have everything inside the pipeline component.

rpelczar:
Can we improve couple more things?

1. Delete this drag&drop option:
image

2. Limit width of freight:
image

3. I think colors should be enabled by default.

ui/src/features/project/pipelines/nodes/stage-node.tsx
```diff
@@ -1,363 +1,337 @@
...
+    if (stagePhase === StagePhase.Promoting) {
+      Phase = (
+        <Link
+          to={generatePath(paths.promotion, {
+            name: props.stage?.metadata?.namespace,
+            promotionId: props.stage?.status?.currentPromotion?.name || ''
+          })}
+        >
+          {Phase}
+        </Link>
```
jmalisz-akuity, line 103: Instead of using assignments with `let` I think stuff like that is a good case for `ConditionalWrapper` generic component:
https://github.com/akuityio/akuity-platform/pull/7221/commits/f1d3808027ced4574c3fdb7cdbac61cdf936a0e7#diff-2439518e51e8f403d0eda180e563cbd0c672ea88d7f63e0cd5e3e2db7127abbe

@rpelczar

ui/src/features/project/pipelines/promotion/mini-graph/use-mini-promotion-graph.ts
```diff
@@ -0,0 +1,162 @@
...
+      edgeHandleIdBuckets = {
+        ...edgeHandleIdBuckets,
+        [type]: {
+          ...edgeHandleIdBuckets[type],
+          [node]: edgeHandleIdBuckets[type][node] + 1
+        }
+      };
+    };
+
+    const useBucket = (type: 'source' | 'target', node: string): string => {
```
jmalisz-akuity, line 53: Please change this to not resemble a hook name to avoid confusion

ui/src/features/project/pipelines/promotion/promotion.tsx
```diff
@@ -0,0 +1,144 @@
...
+        label: 'duration',
+        children: <span title={`${promotionEndTime}`}>{duration}</span>
+      });
+    }
+  }
+
+  promotionDescriptions.push({
+    label: 'created by',
+    children: getPromotionActor(promotion)
+  });
```
jmalisz-akuity, line 99: I see that there is generally a lot of quite complex logic in Kargo components. IMO it would be good if stuff like that was in some hook instead.

jmalisz-akuity (commented):
It looks good overall. I have a few comments, but most of them are more like nitpicks.

Maybe some of those components were moved, so they looked like that before. I didn't work in Kargo yet, so I'm not that familiar with the codebase.

PR #3944, opened 2025-04-28
fix(api): CEL validation `PromotionPolicy`

hiddeco:
Addresses something that went wrong at some point in #3875.


jessesuen (approved):

jessesuen:
Verified the CRDs in this PR applied successfully

PR #3945, opened 2025-04-28
feat:  allow conditional step execution as a function of promotion status 

### LINKED ISSUE #3639: Status Check Functions in conditionals
# Proposed Feature

This is a follow up to https://github.com/akuity/kargo/issues/3125.

We added conditionals, but would now like the ability to run steps based on status conditions (e.g. always, or upon failure).

GitHub Actions solves this by re-using their `if` clause with [Status Check Functions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions).

Examples:

```
if: ${{ success() }}
if: ${{ always() }}
if: ${{ cancelled() }}
if: ${{ failure() }}
```

At first I felt reusing the `if` clause for this purpose seemed a bit too magical and preferred this to be in a separate field. From docs:

> A default status check of success() is applied unless you include one of these functions

However, after thinking about this more, having it in the `if` clause has advantages over introducing it into another field because it can be mixed with other checks. e.g.:

```
if: ${{ failure() && steps.demo.conclusion == 'failure' }}
```

So I am in favor of GitHub's approach to this.

# Motivation

This would be useful in things like notification steps

# Suggested Implementation

Introduce status check functions understood in the `if` clause.

krancour:
Fixes #3639

I am still validating e2e.


internal/promotion/simple_engine.go
```diff
@@ -62,12 +63,12 @@ func (e *simpleEngine) Promote(
...
-	result, err := e.executeSteps(ctx, promoCtx, steps, workDir)
-	if err != nil {
-		return result, fmt.Errorf("step execution failed: %w", err)
+	result := e.executeSteps(ctx, promoCtx, steps, workDir)
+	if result.Status == kargoapi.PromotionPhaseErrored {
+		err = errors.New(result.Message)
 	}
 
-	return result, nil
+	return e.executeSteps(ctx, promoCtx, steps, workDir), err
```
hiddeco, line 71: Think this shouldn't execute twice?

api/v1alpha1/promotion_types.go
```diff
@@ -64,13 +74,37 @@ const (
...
+func (s PromotionStepStatus) Compare(rhs PromotionStepStatus) int {
+	// Define severity levels for each status
+	severity := map[PromotionStepStatus]int{
+		PromotionStepStatusSucceeded: 0,
+		PromotionStepStatusSkipped:   1,
+		PromotionStepStatusRunning:   2,
+		PromotionStepStatusAborted:   3,
+		PromotionStepStatusFailed:    4,
+		PromotionStepStatusErrored:   5,
+	}
```
hiddeco, line 90: I think this can be moved outside the method.

api/v1alpha1/promotion_types.go
```diff
@@ -64,13 +74,37 @@ const (
...
-// IsTerminal returns true if the PromotionPhase is a terminal one.
-func (p *PromotionPhase) IsTerminal() bool {
-	switch *p {
-	case PromotionPhaseSucceeded, PromotionPhaseFailed, PromotionPhaseErrored, PromotionPhaseAborted:
-		return true
+// Compare compares the severity of the current PromotionStepStatus with
+// another. It returns -1 if the current status is less severe than the other, 0
+// if the two statuses are equally severe, or 1 if the current status is more
+// severe than the other.
+func (s PromotionStepStatus) Compare(rhs PromotionStepStatus) int {
```
hiddeco, line 81: Nit: slightly more idiomatic.

```suggestion
func (s PromotionStepStatus) Compare(other PromotionStepStatus) int {
```

internal/promotion/promotion.go
```diff
@@ -193,7 +208,7 @@ func (s *Step) Skip(
 	state promotion.State,
 ) (bool, error) {
 	if s.If == "" {
-		return false, nil
+		s.If = "${{ success() }}"
```
hiddeco, line 211: 1. Was this added because the meaning of `success()` may change at some point, and we want to continue to have consistent behavior?
2. As this takes a pointer to `s`, do we _actually_ want to modify it?
3. If "yes" on 1), I think this defaulting behavior requires documentation and reasoning.

krancour: The default behavior is intended to be that if no expression is given, it's equivalent to `${{ success() }}`. Whether I took myself too literally or just got lazy and/or saw an opportunity for an easy implementation, perhaps we'll never know.

But I don't think it would be too hard to do this differently and it would be a bit efficient to avoid evaluating this "extra" expression. I'll change it.

api/v1alpha1/promotion_types.go
krancour: I'd meant to and forgot. I didn't intend to re-instantiate this map every time. Fixing.

krancour:
@hiddeco all feedback addressed in 4ac91af0db948498a05c7ceb89c7e7e9c238cf56

krancour:
There's an issue lurking somewhere in here where a step can appear to be stuck in a running state even though it completes successfully.

krancour:
@rpelczar this PR has no changes to UI code except for ones that are generated. I can't quite figure out why typechecking is failing. Is this something you'd be able to take a look at?

Please feel free to amend the PR if some small fix is needed.

hiddeco (approved):

krancour:
@rpelczar and @hiddeco thank you both for the assist!

Issue #3848, opened 2025-04-11
Remove deprecated `phase` and `message` fields in favor of conditions


hiddeco:
Some time ago, Conditions were added to the `Project`, `Stage`, and `Warehouse` kinds while existing `Phase` and `Message` fields were marked as deprecated (#2239, #2401, #2905). 

To finalize this change, we should:

- [ ] Remove the deprecated fields from the statuses
- [ ] Remove any migration logic
- [ ] Ensure errors for control flow Stages surface as a Condition

PR #3866, opened 2025-04-15
chore: remove deprecated phase, message fields for Project

0xMALVEE:
Issue: https://github.com/akuity/kargo/issues/3848
Removed deprecated fields phase, message from the Project

PR #3949, opened 2025-04-29
chore(api): remove deprecated `phase` and `message` fields from Project and Stage

hiddeco:
Supersedes: #3866 (thank you @0xMALVEE!)
Resolves: #3848


ui/src/features/common/stage-status/stage-condition-icon.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  // Default state
+  let iconState: IconState = {
+    icon: faCircleMinus,
+    tooltipTitle: '',
+    tooltipMessage: '',
+    iconClass: 'text-gray-400'
+  };
+
+  // Priority: Promoting > Verifying > Reconciling > Failed > Ready
```
Marvin9, line 74: Curious how can multiple condition exist at the same time?

hiddeco: Conditions are a slice of things true for the object, instead of a singular phase. Depending on the condition type and polarity, it's either permanently or temporarily present on the resource (see `api/v1alpha1/conditions.go`).

This follows the Kubernetes pattern where conditions represent different aspects of a resource's state rather than mutually exclusive phases. For example, a `Pod` can be both `PodScheduled=True `and `Ready=False` simultaneously — meaning it's been scheduled to a node but isn't ready to serve traffic yet.

rpelczar (approved):
I don't like that we have to hardcode enums, but I assume we don't have a simple solution for now.

We also might need some performance improvements - for the project with >100 stages, we will have a lot of unnecessary calculations. If you have time, you can check this out - https://react.dev/reference/react/useMemo and try to improve, otherwise we can do this optimization later.

krancour (approved):

PR #3956, opened 2025-04-30
fix compose-out example

EmilyXinyi:
The current example of referencing the output of `compose-output` in a `http` step's for Slack message will fail with a "cannot fetch output" error. However, dropping the `task` works. (ie. `outputs['pr-link'].url`)


hiddeco:
We need to find a better way to document this, as #3862 changed this because it was confusing from the perspective of writing a `PromotionTask` (which is what the `compose-output` step initially was designed for).

A good starting point would probably be to add a note block that informs people about `task` versus `task`-less usage depending on if the `compose-output` step is used within a PromotionTask or Promotion.

PR #3957, opened 2025-04-30
docs(oidc-setup): fix wrong example (viewer→viewerS)

Horgix:
The doc mentions `api.oidc.viewers` everywhere, but the example snippet uses `api.oidc.viewers` (note the missing S).

The expected key being `viewers` is confirmed by the Helm Chart parameters doc: https://github.com/akuity/kargo/blob/f001566652b001b9c2fec628d2df130e145a5ddb/charts/kargo/README.md?plain=1#L116


krancour (approved):
Nice catch @Horgix. Thank you!

Issue #3720, opened 2025-03-26
git-merge-pr step


jessesuen:
# Proposed Feature

I spoke with an organization that preferred Kargo to merge the PR it created rather than their end users. e.g. Their use case would have promo steps something like:

1. git-open-pr
2. HTTP stuff 
3. git-merge-pr
4. deploy

# Motivation

In their use case, there was no need for the PR that kargo created, to be a human interaction point, but rather for information/automation purposes, and so they wanted the PR open/close to be fully automated.

# Suggested Implementation

Introduce a `git-merge-pr` step that effectively clicks the merge button given a PR number. 

File the issue for tracking interest.

krancour:
Note that @fykaa has already taken a stab at what this might look like:

https://github.com/akuity/kargo/pull/3958/files

But there are still open questions.

michaelasper:
One note: Github supports enabling auto merge, could be a nice way to support this by toggling auto merge if it's Github PR. If it's not in the original design, I would be open to adding it later on as a flag

krancour:
> Github supports enabling auto merge

That raises another interesting dimension to all of this. In creating an abstraction over many different services/providers, the tendency (sometimes for better, other times for worse) is to be constrained by the lowest common denominator.

i.e. If GitHub can do this and all the others can't, it makes it awfully hard to incorporate.

This isn't to say we can't/won't. Just noting what some challenges are here.

fykaa:
@krancour 
> How is this step to behave when conditions do not permit merge to be "clicked?" Fail immediately? Retry on the next reconciliation attempt, with a long timeout, like we have for wait-for-pr?

design considerations I think can be made: 
- default behaviour: fail the step immediately with a clear message (like "Merge blocked due to required reviews or checks")
- optional retry mode: a config like `retryUntilMergeable: true` to make it behave like `wait-for-pr` with a long timeout and periodic retries
- merge conflicts: fail with an error saying the PR cannot be merged due to conflicts
- having a customizable failure policy: support `onFailure: [fail | retry | ignore]`

> Does it attempt to use any admin-authority associated with the credentials to override branch protections (i.e. "Merge without waiting for requirements to be met (bypass rules)")?

this can be _opt-in only_ maybe via a config like `bypassProtections: true`. The docs can clearly state that github credentials must have admin or bypass rights for this to work

> Which also raises the question of whether automatic merging might only need to be an option on the existing wait-for-pr step instead of being its own step.

[...]

PR #3958, opened 2025-04-30
feat: git-merge-pr promotion step

fykaa:
closes https://github.com/akuity/kargo/issues/3720

This adds a new `git-merge-pr` step to Kargo's promotion pipeline, letting automated merging of pull requests previously opened via the `git-open-pr` step.

### What this does

- Merges a given PR by PR number and repo URL
- Supports optional merge methods: `merge`, `squash`, `rebase`
- Waits for required status checks before merging (configurable)
- Optional cleanup of source branch post-merge

### Notes

- Provider is inferred automatically from the `repoURL`, consistent with other Git steps
- Auth and API interaction follow the same model as `git-open-pr`

### DIFF
--- a/pkg/promotion/runner/builtin/git_pr_merger.go
+++ b/pkg/promotion/runner/builtin/git_pr_merger.go
@@ -3,6 +3,7 @@ package builtin
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/xeipuuv/gojsonschema"
 
@@ -111,14 +112,31 @@ func (g *gitPRMerger) run(
 			fmt.Errorf("error creating git provider service: %w", err)
 	}
 
-	// Try to merge the PR
-	mergedPR, merged, err := gitProv.MergePullRequest(ctx, cfg.PRNumber)
-	if err != nil {
-		// Only actual errors (auth, network, invalid PR, closed but not merged, etc.) reach here
-		return promotion.StepResult{Status: kargoapi.PromotionStepStatusFailed},
-			&promotion.TerminalError{
-				Err: fmt.Errorf("error merging pull request %d: %w", cfg.PRNumber, err),
-			}
+	// Try to merge the PR using a primitive retry loop. PRs are often ready to
+	// merge moments after being opened, but not quite immediately. Accounting
+	// for this internally avoids the scenario where a Promotion needs to wait
+	// for its next regularly scheduled reconciliation to merge a PR that could
+	// have been merged already if we were patient for just a few seconds.
+	var mergedPR *gitprovider.PullRequest
+	var merged bool
+	const maxMergeAttempts = 3
+	for i := range maxMergeAttempts {
+		if mergedPR, merged, err = gitProv.MergePullRequest(
+			ctx, cfg.PRNumber,
+		); err != nil {
+			// Only actual errors (auth, network, invalid PR, closed but not merged,
+			// etc.) reach here
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusFailed},
+				&promotion.TerminalError{
+					Err: fmt.Errorf("error merging pull request %d: %w", cfg.PRNumber, err),
+				}
+		}
+		if merged {
+			break
+		}
+		if i < maxMergeAttempts {
+			time.Sleep(time.Second * 5)
+		}
 	}
 
 	if !merged {


internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
```diff
@@ -0,0 +1,47 @@
...
+        "description": "Skip TLS verification when interacting with the Git provider. Default is false."
+      },
+      "deleteBranchAfterMerge": {
+        "type": "boolean",
+        "description": "If true, deletes the source branch after merging the PR. Default is false."
+      },
+      "waitForChecks": {
+        "type": "boolean",
+        "description": "If true, waits for all required status checks to pass before merging. Default is true."
+      },
```
krancour, line 40: I believe specific permissions are required (at least in GitHub) to bypass branch protections. We probably need to call that out and also define what the behavior is if you configure this to not wait, but lack the necessary permissions to do so.

internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
```diff
@@ -0,0 +1,47 @@
...
+      "maxWaitSeconds": {
+        "type": "integer",
+        "description": "Maximum number of seconds to wait for checks to pass. Default is 600 (10 minutes).",
+        "minimum": 1
+      }
```
krancour, line 45: This looks like it would be problematic.

Including this in step configuration seems to imply the step itself is going to wait _internally_ and we ought not do that. "Idling" steps unnecessarily consumes a portion of the controller's finite concurrency. With enough of that going on, the system can become live-locked.

This should be configured using [retries](https://docs.kargo.io/user-guide/reference-docs/promotion-templates#step-retries) instead. (The step can specify its own default timeout that overrides the system-wide default.) Using this approach, Promotions that are just _waiting_ for something external to happen do not consume capacity.

internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
```diff
@@ -0,0 +1,42 @@
...
+      "provider": {
+        "type": "string",
+        "description": "The name of the Git provider to use. Currently 'azure', 'bitbucket', 'gitea', 'github', and 'gitlab' are supported. Kargo will try to infer the provider if it is not explicitly specified.",
+        "enum": ["azure", "bitbucket", "gitea", "github", "gitlab"]
+      },
+      "mergeMethod": {
+        "type": "string",
+        "description": "The merge strategy to use. If not specified, the provider's default will be used.",
+        "enum": ["merge", "squash", "rebase"]
+      },
```
krancour: A bit of research may be required here. Off the top of my head, I don't know if the APIs of all the providers we currently support do, in turn, support this option. Even if they all do, do they all permit specifying your preference when making the API call to _merge_ the PR? Or do some support it by allowing you to specify your preference when making the API call to _open_ the PR (and then merge just does what was specified when the PR was opened)?

I'm not suggesting we don't support these options. I'm questioning what the right way to do it might be. Might we, for instance, need the same choice to be available on the `git-open-pr` step to accommodate providers that do things one way as well as those doing it the other way?

internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
```diff
@@ -0,0 +1,42 @@
...
+      "insecureSkipTLSVerify": {
+        "type": "boolean",
+        "description": "Skip TLS verification when interacting with the Git provider. Default is false."
+      },
+      "deleteBranchAfterMerge": {
```
krancour: Similar question to [here](https://github.com/akuity/kargo/pull/3958/files#r2238188381).

internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
krancour: Another possibility is that we might need to carve out provider-specific config sections if the capabilities of each vary drastically.

internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
```diff
@@ -0,0 +1,42 @@
...
+        "type": "boolean",
+        "description": "If true, deletes the source branch after merging the PR. Default is false."
+      },
+      "waitForChecks": {
```
krancour: Being honest about what I don't know... do all the providers support something like "checks?" Or are our expectations of what they all support mis-calibrated by our familiarity with GitHub?

krancour:
Sorry it's taken so long to look at this.

I commented on a few specific bits of the schema for the steps, but in general, I don't feel as if we have clear answers yet to the concerns raised in #3720.

In general, we will either need to cater to the lowest common denominator in terms of features _or_ figure out how to support provider-specific options. This is going to require quite a bit of research on your part to understand what each provider's API can and cannot do.

If we don't have clear answers for those things yet or can't get them quickly, then we might consider a more minimally viable approach to this feature. i.e. Make it less configurable and we declare the feature to be beta, which I think is entirely reasonable if there are a lot of unanswered questions.

Moving this to the v1.8.0 milestone.

internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
krancour: As we're now adopting a very simple initial approach as discussed [here](https://github.com/akuity/kargo/issues/3720#issuecomment-3272807235), we don't need to make the specific merge method configurable at this point in time.

internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
krancour: This is another thing we can count as out of scope for this initial implementation.

internal/promotion/runner/builtin/schemas/git-merge-pr-config.json
krancour: Per conversation [here](https://github.com/akuity/kargo/issues/3720#issuecomment-3272807235), I propose changing this to `wait`. It will not be dependent on platform-specific features (checks a GitHub thing), and all it will do is cause the step to return `RUNNING` instead of `FAILED` if the PR is not yet mergeable. The promotion will idle for a few minutes and the merge will be attempted again on the next reconciliation.

pkg/promotion/runner/builtin/git_pr_merger.go
```diff
@@ -0,0 +1,340 @@
...
+	for _, pattern := range allRetryablePatterns {
+		if strings.Contains(errStr, strings.ToLower(pattern)) {
+			return true
+		}
+	}
+
+	// For any other errors, be conservative and don't retry
+	// unless we explicitly know they're retryable
+	return false
+}
```
krancour: Analyzing the body of error messages in this way is brittle and it's also forcing a large amount of provider-specific knowledge (for _every_ provider) onto this step runner.

Except in tests, a better option for understanding the nature of an error is creating specific error wrapper types(s) that callers can check for to understand more clearly how to handle the error, but here, even that doesn't feel right...

As a rule of thumb, don't rely on errors to convey information _when everything is working as expected_. It's an _expected_ condition that a PR might not be in a mergeable state, _yet_. That probably shouldn't be treated as an error.

In light of that, a better option may be to make the `Merge()` method return a `bool`, in addition to what it returns now, indicating if the merge has been performed or not. Then inability to merge due to current state returns `false` _and no error_. Actual problems can return an error.

internal/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,83 @@ func (p *provider) ListPullRequests(
...
+		switch strings.ToLower(opts.MergeMethod) {
+		case "squash":
+			mergeStrategy = adogit.GitPullRequestMergeStrategyValues.Squash
+		case "rebase":
+			mergeStrategy = adogit.GitPullRequestMergeStrategyValues.Rebase
+		default:
+			// Default to NoFastForward for "merge" or any other method
+			mergeStrategy = adogit.GitPullRequestMergeStrategyValues.NoFastForward
+		}
+	}
```
krancour: It seems all the providers have code similar to this. I'm wondering why we're bothering with this given that we're not surfacing this option in the step's configuration.

If our first iteration of the merge step is "merge or die," shouldn't we just merge without specifying and accept the remote repo's default merge strategy?

fykaa: right now the implementation already defers to the repo's default by passing an empty merge method. hence in practice, we're already doing the 'merge or die' behavior. The only reason providers expose `MergeMethod` is for cases where we want deterministic behavior across repos (like always squash), or for testing. It isn't serving an immediate purpose, I'm good with simplifying this to just use provider defaults for the first iteration, and we can always reintroduce the explicit merge strategy later if consistency requirements come up

krancour: There's nothing inconsistent about using whatever is the repo's default as long as that's what we're _always_ doing.

It's within the realm of possibility that we need this eventually, but at this stage, I would like to keep this PR as simple as we possibly can.

pkg/promotion/runner/builtin/git_pr_merger.go
```diff
@@ -0,0 +1,339 @@
...
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusFailed},
+			&promotion.TerminalError{
+				Err: fmt.Errorf("pull request %d is closed but not merged", cfg.PRNumber),
+			}
+	}
+
+	// Try to merge the PR
+	mergedPR, err := gitProv.MergePullRequest(ctx, cfg.PRNumber, &gitprovider.MergePullRequestOpts{
+		CommitTitle:   fmt.Sprintf("Merge PR #%d", cfg.PRNumber),
+		CommitMessage: "Automatically merged by Kargo promotion step",
```
fykaa: Right now this step uses a hardcoded commit title and message, since the config doesn't expose any way to override this, is that considered idiomatic for our workflow, or should we support configurable commit messages?

pkg/promotion/runner/builtin/git_pr_merger.go
```diff
@@ -0,0 +1,339 @@
...
+	})
+	if err != nil {
+		// If waiting is enabled and the error suggests the PR is not ready to merge,
+		// return RUNNING to retry later
+		if cfg.Wait && g.isRetryableError(err) {
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusRunning}, nil
+		}
```
fykaa: when `wait` is true, the code will keep retrying indefinitely until the PR is either merged or closed. Is it acceptable for us to allow infinite retries, or should we build in a timeout or max retry limit to avoid steps getting stuck forever?

krancour: Same as other steps with similar behavior, it's handled at a higher level than the step runner's own logic. No need to worry about it here.

pkg/promotion/runner/builtin/git_pr_merger.go
krancour: Discussed offline. We neither want to make this user-configurable at this point in time nor provide this optional info to any of the Git hosting providers in the merge API calls.

pkg/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,79 @@ func (p *provider) ListPullRequests(
...
+	ctx context.Context,
+	id int64,
+) (*gitprovider.PullRequest, bool, error) {
+	gitClient, err := adogit.NewClient(ctx, p.connection)
+	if err != nil {
+		return nil, false, fmt.Errorf("error creating Azure DevOps client: %w", err)
+	}
+
+	// Get the current PR to check its status and get the last merge source commit
+	currentPR, err := gitClient.GetPullRequest(ctx, adogit.GetPullRequestArgs{
```
krancour: "current" is an odd word to be using here because that's a temporal or relative term. There's no previous PR. There's no next PR. Isn't this just _the_ PR?

pkg/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,79 @@ func (p *provider) ListPullRequests(
...
+	if currentPR == nil {
+		return nil, false, fmt.Errorf("pull request %d not found", id)
+	}
+
+	status := ptr.Deref(currentPR.Status, adogit.PullRequestStatusValues.NotSet)
+	mergeStatus := ptr.Deref(currentPR.MergeStatus, adogit.PullRequestAsyncStatusValues.NotSet)
+
+	switch status {
+	case adogit.PullRequestStatusValues.Completed:
+		pr, convertErr := convertADOPullRequest(currentPR)
```
krancour: It's idiomatic to call an error `err`. It's a rare circumstance that you _need_ to deviate from that. I don't see what's forcing you to do that here, so it would be best to be idiomatic.

pkg/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,79 @@ func (p *provider) ListPullRequests(
...
+		return pr, true, nil
+	case adogit.PullRequestStatusValues.Abandoned:
+		return nil, false, fmt.Errorf("pull request %d is abandoned", id)
+	case adogit.PullRequestStatusValues.Active:
+		// If merge status shows conflicts or policies not met → retry later
+		if mergeStatus == adogit.PullRequestAsyncStatusValues.Conflicts ||
+			mergeStatus == adogit.PullRequestAsyncStatusValues.Queued ||
+			mergeStatus == adogit.PullRequestAsyncStatusValues.RejectedByPolicy ||
+			mergeStatus == adogit.PullRequestAsyncStatusValues.Failure {
+			return nil, false, nil
```
krancour, line 229: Few things in this case...

According to docs, draft PRs are still considered active.

By your current logic, an active PR that's a draft, but with `mergeStatus` `succeeded` (which actually just means _mergeable_), you're going to try to merge and you ought not.

So I think you need to be checking `IsDraft`.

The other thing about this case, is that it seems it would be more straigtforward to check for `mergeStatus != adogit.PullRequestAsyncStatusValues.Succeeded` instead of checking for these many other conditions.

pkg/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,79 @@ func (p *provider) ListPullRequests(
...
+	case adogit.PullRequestStatusValues.Completed:
+		pr, convertErr := convertADOPullRequest(currentPR)
+		if convertErr != nil {
+			return nil, false, fmt.Errorf("error converting pull request %d: %w", id, convertErr)
+		}
+		return pr, true, nil
+	case adogit.PullRequestStatusValues.Abandoned:
+		return nil, false, fmt.Errorf("pull request %d is abandoned", id)
+	case adogit.PullRequestStatusValues.Active:
+		// If merge status shows conflicts or policies not met → retry later
```
krancour: I would encourage you to avoid talking about "retries."

This component doesn't know anything about retries.

All it does is tries to merge and indicates if it was successful or not. Any decisions about retries are made at a much higher level.

pkg/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,79 @@ func (p *provider) ListPullRequests(
...
+		if mergeStatus == adogit.PullRequestAsyncStatusValues.Conflicts ||
+			mergeStatus == adogit.PullRequestAsyncStatusValues.Queued ||
+			mergeStatus == adogit.PullRequestAsyncStatusValues.RejectedByPolicy ||
+			mergeStatus == adogit.PullRequestAsyncStatusValues.Failure {
+			return nil, false, nil
+		}
+	default:
+		// Draft or not mergeable yet → retry later
+		return nil, false, nil
```
krancour: See previous comment. Active and draft are not mutually exclusive. Active and not yet mergeable are _also_ not mutually exclusive. So the conclusions drawn by this comment aren't correct and it will mislead people who try to understand this code in the future.

I would just remove the comment.

pkg/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,79 @@ func (p *provider) ListPullRequests(
...
+	}
+
+	// Try to merge
+	updatedPR, err := gitClient.UpdatePullRequest(ctx, adogit.UpdatePullRequestArgs{
+		Project:       &p.project,
+		RepositoryId:  &p.repo,
+		PullRequestId: ptr.To(int(id)),
+		GitPullRequestToUpdate: &adogit.GitPullRequest{
+			Status:                ptr.To(adogit.PullRequestStatusValues.Completed),
+			LastMergeSourceCommit: currentPR.LastMergeSourceCommit,
```
krancour: It wasn't initially clear to me before what this was for. I had to look it up. I understand now that it's a concurrency control mechanism. The merge won't proceed if the PR has been amended between the time you retrieved it and the time you attempt to merge it. It is probably worth adding a brief comment explaining this.

pkg/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,79 @@ func (p *provider) ListPullRequests(
...
+		},
+	})
+	if err != nil {
+		// Network/auth errors → terminal
```
krancour: This is probably too narrow a description of what might land you here... I think _any_ unanticipated error is going to land you here. I would just remove the comment.

pkg/gitprovider/bitbucket/bitbucket.go
```diff
@@ -289,6 +296,66 @@ func (p *provider) ListPullRequests(
...
+
+	// Check if PR is already merged
+	if currentPR.State == prStateMerged {
+		return toProviderPR(currentPR, currentPRResp), true, nil
+	}
+
+	// Check if PR is closed/declined
+	if currentPR.State == prStateDeclined || currentPR.State == prStateSuperseded {
+		return nil, false, fmt.Errorf("pull request %d is closed but not merged (state: %s)", id, currentPR.State)
+	}
```
krancour: I'm not positive, but I believe Bitbucket Server/Data Center also has `CLOSED`...

But instead of checking for `DECLINED`, `SUPERSEDED`, and `CLOSED`, maybe it's better to just check for _not_ `OPEN`?

pkg/gitprovider/bitbucket/bitbucket.go
```diff
@@ -289,6 +296,66 @@ func (p *provider) ListPullRequests(
...
+
+	// Check if PR is not in open state
+	if currentPR.State != prStateOpen {
+		// PR is not ready to merge yet (pending checks, conflicts, etc.)
+		return nil, false, nil
+	}
+
+	// Attempt to merge the PR
```
krancour, line 349: If we get to here, unless I'm reading things wrong, we don't actually know anything other than that the PR is open, right? We don't actually know anything about whether it's mergeable or not. Is there a way to check on that?

pkg/gitprovider/gitea/gitea.go
```diff
@@ -257,6 +276,50 @@ func (p *provider) ListPullRequests(
...
+		return &pr, true, nil
+	}
+
+	// Check if PR is not in open state
+	if currentPR.State != gitea.StateOpen {
+		// Not ready to merge
+		return nil, false, nil
+	}
+
+	_, err = p.client.MergePullRequest(ctx, p.owner, p.repo, int(id), &gitea.MergePullRequestOption{})
```
krancour: Similar to my feedback on Bitbucket... at this point, don't we _only_ know that the PR is open. I don't think we know that it's mergeable.

There appears to be a `Mergeable` property you can check.

pkg/gitprovider/github/github.go
```diff
@@ -268,6 +295,62 @@ func (p *provider) ListPullRequests(
...
+		pr := convertGithubPR(*currentPR)
+		return &pr, true, nil
+	}
+
+	// Check if PR is closed but not merged - this is a terminal error
+	if ptr.Deref(currentPR.State, prStateClosed) != prStateOpen {
+		return nil, false, fmt.Errorf("pull request %d is closed but not merged", id)
+	}
+
+	mergeResult, _, err := p.client.MergePullRequest(
```
krancour, line 324: As with some of the other providers, I think you have access to more details here regarding whether the PR is actually mergeable or not, which you should be checking.

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -198,6 +205,45 @@ func (p *provider) ListPullRequests(
...
+		return &pr, true, nil
+	}
+
+	// Check if MR is open
+	if currentMR.State != "opened" {
+		// MR is closed or locked; cannot merge
+		return nil, false, nil
+	}
+
+	// Merge the MR
```
krancour, line 233: I think you want to check `DetailedMergeStatus` first?

pkg/gitprovider/github/github.go
```diff
@@ -268,6 +295,62 @@ func (p *provider) ListPullRequests(
...
+		ctx,
+		p.owner,
+		p.repo,
+		int(id),
+		"", // Empty string - let GitHub use its default Commit Message
```
krancour:
```suggestion
		"", // Let GitHub use its default Commit Message
```

pkg/gitprovider/azure/azure.go
fykaa: we are assigning PR at multiple places across the codebase, hence I will use `adoPR` for the Azure DevOps pull request object (the native type) and use `pr` only for the final converted `gitprovider.PullRequest` that gets returned. i think this is cleaner than using "current"

pkg/gitprovider/azure/azure.go
fykaa: it was a variable shadowing lint issue, I will make `pr` a function scoped variable, to make this happen

pkg/gitprovider/bitbucket/bitbucket.go
fykaa: I can't find `CLOSED` PR state on their documentation. I do see that declined and superseded is referred as closed in bitbucket.

krancour: In this particular case, it was Copilot who told me about Bitbucket Server having CLOSED. Copilot is not trustworthy, which is why I said I'm not positive...

However, it's also not uncommon (unfortunately) for self-hosted versions of these platforms to have small differences from the SaaS versions. And they typically have their own _separate_ documentation that might be hard to find...

At any rate, if you check for not OPEN, you don't need to be explicitly checking for CLOSED (or the other statuses) anyway.

pkg/gitprovider/bitbucket/bitbucket.go
krancour: Note that for many of the other platforms, there ended up being information about mergeability right in the PR, but I wasn't easily spotting any such thing here. I might not have looked hard enough. Or it might involve additional API calls.

fykaa: Bitbucket's API does not provide a direct `mergeable` boolean field unlike others. I have added it to others. for bitbucket, I can only attempt a merge and catch errors indicating mergeability issues

pkg/gitprovider/azure/azure.go
```diff
@@ -181,6 +181,83 @@ func (p *provider) ListPullRequests(
...
+		if err != nil {
+			return nil, false, fmt.Errorf("error converting pull request %d: %w", id, err)
+		}
+		return pr, true, nil
+	case adogit.PullRequestStatusValues.Abandoned:
+		return nil, false, fmt.Errorf("pull request %d is abandoned", id)
+	case adogit.PullRequestStatusValues.Active:
+		if ptr.Deref(adoPR.IsDraft, false) {
+			// Draft PRs should not be merged
+			return nil, false, nil
```
krancour, line 225: For the benefit of future readers of this code:

```suggestion
		// Draft PRs can have a merge status of `succeeded`, but aren't actually
		// mergable, so we explicitly check for draft status.
		if ptr.Deref(adoPR.IsDraft, false) {
			return nil, false, nil
```

pkg/gitprovider/bitbucket/bitbucket.go
krancour: Ok. I fell down the rabbit hole and I've spent quite a bit of time now trying to figure out how to do this. API docs don't cover this. Posts in various forums complain about this and all suggested workarounds I have seen are very brittle. There's an [open feature request](https://jira.atlassian.com/browse/BCLOUD-22014) that's almost old enough to be starting kindergarten. Unfortunately, in my experience, it's not uncommon that Bitbucket is missing features common to their competitors.

The way I'd like to handle this is:

1. It seems it's easy enough to check whether the PR is a draft, so we can at least not attempt the merge under that condition.

1. Keep the rest of this logic as-is.

1. Add a comment/TODO on what is now line ~~+339~~ +332 indicating that we have no choice here but to attempt the merge and hope for the best until the Bitbucket API improves in ways that facilitate a comprehensive check for merge eligibility.

1. In our docs, explicitly state that the "wait" option is unreliable for repositories hosted in Bitbucket.

1. Open a follow-up issue in our repo to track this.

pkg/gitprovider/gitea/gitea.go
krancour: Ok. Please check `Mergeable`. There is no `Draft` field in the `PullRequest` struct, although Gitea's API docs indicate the field exists.

I've opened a PR to surface the value of that field in the struct:

https://gitea.com/gitea/go-sdk/pulls/731

Please make a TODO comment in the code noting that we're not yet able to check for draft status and include the URL above.

pkg/gitprovider/github/github.go
krancour: Confirmed `Draft` and `Mergeable` fields are present in the struct.

pkg/gitprovider/gitlab/gitlab.go
krancour: You're looking for a value of `mergeable` in that field.

According to docs, that field have a value of `draft_status` if the MR were a draft, so there is no need to separately look at the `Draft` field.

https://docs.gitlab.com/api/merge_requests/#merge-status

krancour:
@fykaa this needs docs as well.

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -219,20 +218,22 @@ func (p *provider) MergePullRequest(
...
 		return &pr, true, nil
-	}
 
-	// Check if MR is open
-	if glMR.State != "opened" {
-		// MR is closed or locked; cannot merge
+	case glMR.State != "opened":
+		return nil, false, nil
+
+	case glMR.DetailedMergeStatus != "can_be_merged":
```
krancour: `can_be_merged` seems to be a possible value of the `merge_status_field`, which is not surfaced in the struct. We are looking at the `detailed_merge_status` field, and documentation seems to indicate `mergeable` is the value we should be looking for.

Feel free to double-check me.

Ref: https://docs.gitlab.com/api/merge_requests/#merge-status

```suggestion
	case glMR.DetailedMergeStatus != "mergeable":
```

pkg/gitprovider/github/github.go
```diff
@@ -300,49 +300,49 @@ func (p *provider) MergePullRequest(
...
 		pr := convertGithubPR(*ghPR)
 		return &pr, true, nil
-	}
 
-	// Check if PR is closed but not merged - this is a terminal error
-	if ptr.Deref(ghPR.State, prStateClosed) != prStateOpen {
+	case ptr.Deref(ghPR.State, prStateClosed) != prStateOpen:
 		return nil, false, fmt.Errorf("pull request %d is closed but not merged", id)
+
+	case ghPR.Mergeable == nil:
```
krancour: Let's just treat this the same as `false`, please.

krancour: You also still need to check `Draft`.

pkg/promotion/runner/builtin/git_pr_merger.go
```diff
@@ -0,0 +1,156 @@
...
+		}, nil
+	}
+
+	// If PR is closed but not merged, it's a terminal error
+	if !pr.Open {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusFailed},
+			&promotion.TerminalError{
+				Err: fmt.Errorf("pull request %d is closed but not merged", cfg.PRNumber),
+			}
+	}
```
krancour: How necessary is this? All the providers are retrieving the PR by its ID and I think they would be more reliable at inspecting the current state of the PR.

pkg/gitprovider/gitlab/gitlab.go
fykaa: `DetailedMergeStatus` field already handles draft status detection, when a merge request is in draft status, the `DetailedMergeStatus` will be "draft_status" instead of "can_be_merged". I think the current implementation doesn't require draft detection, or I can add an extra `glMR.Draft` check

pkg/gitprovider/github/github.go
fykaa: I skipped checking for draft since github's `Mergeable` field already returns false for draft PRs, then the explicit draft check (ghPR.Draft != nil && *ghPR.Draft) would be redundant.

pkg/promotion/runner/builtin/git_pr_merger.go
fykaa: this helps avoid unnecessary merge attempt when the PR is already in a terminal state (merged/ closed) and also keeps the errors handling consistent for invalid PRs.

And for already-merged PRs which could be a common case in retry scenarios, this could help avoid the more expensive merge operation entirely

And it also keeps things a bit consistent with `git-wait-pr` promo step implementation

pkg/gitprovider/gitlab/gitlab.go
krancour: That's what I said. 😉

All good.

pkg/gitprovider/bitbucket/bitbucket.go
```diff
@@ -289,6 +296,75 @@ func (p *provider) ListPullRequests(
...
+	if bbPR.Draft {
+		return nil, false, fmt.Errorf("cannot merge pull request %d: pull request is in draft state", id)
+	}
+
+	// TODO: The Bitbucket API lacks comprehensive merge eligibility checks. We
+	// cannot reliably determine if a PR is mergeable due to conflicts, failing
+	// checks, or other blocking conditions before attempting the merge. This means
+	// we have no choice but to attempt the merge and hope for the best.
+	// See: https://bitbucket.org/site/master/issues/5814/
+	// This limitation makes the "wait" option unreliable for Bitbucket repositories.
```
krancour: What is: https://bitbucket.org/site/master/issues/5814/ ?

That appears to have nothing to do with this problem.

```suggestion
	// TODO: The Bitbucket API lacks comprehensive merge eligibility checks. We
	// cannot reliably determine if a PR is mergeable due to conflicts, failing
	// checks, or other blocking conditions before attempting the merge. This
	// means we have no choice but to attempt the merge and hope for the best.
	//
	// See: https://jira.atlassian.com/browse/BCLOUD-22014
	//
	// This limitation makes the "wait" option unreliable for Bitbucket
	// repositories.
```

pkg/gitprovider/gitea/gitea.go
```diff
@@ -257,6 +276,54 @@ func (p *provider) ListPullRequests(
...
+	case giteaPR.State != gitea.StateOpen:
+		return nil, false, nil
+
+	case !giteaPR.Mergeable:
+		return nil, false, nil
+	}
+
+	// TODO(fykaa): We should also check if the PR is in draft status before merging.
+	// The Gitea API includes a draft field, but the go-sdk doesn't expose it yet.
+	// See: https://gitea.com/gitea/go-sdk/pulls/731
```
krancour:
```suggestion
	// TODO(fykaa): We should also check if the PR is in draft status before
	// merging. The Gitea API includes a draft field, but the go-sdk doesn't
	// expose it yet.
	//
	// See: https://gitea.com/gitea/go-sdk/pulls/731
```

pkg/promotion/runner/builtin/git_pr_merger.go
krancour: > this helps avoid unnecessary merge attempt when the PR is already in a terminal state

All your `MergePullRequest()` implementations are (correctly) already doing that.

> and also keeps the errors handling consistent for invalid PRs.

You're weighting this too heavily. If a closed PR is encountered, all your `MergePullRequest()` implementations emit errors that are crystal clear about what happened and that's the main thing that matters.

> And it also keeps things a bit consistent with git-wait-pr promo step implementation

That is a poor comparison because that's a far, far simpler step. There are many fewer provider-specific details to analyze, so it's perfectly reasonable to retrieve the information, transform it to a provider-neutral representation and _then_ make a decision. Here, you have a lot of provider-specific details to examine, which are all inaccessible to you after a PR has been transformed into a provider-neutral representation.

In your current state, `GetPullRequest()` retrieves the PR from the provider and transforms it to the provider-neutral representation. You are checking certain things, then if you clear those checks, `MergePullRequest()` is retrieving the PR from the provider _again_, checking many of the same things you already checked, checking _additional_ things, and then doing the transformation _again_.

You are quite unnecessarily doubling up on retrievals and transformations. There is no reason to call an external API twice when once will do. It also introduces the possibility of a scenario where the state of the PR changes between the first retrieval and the second retrieval, meaning decisions in the step and in the provider won't be made based on a consistent state.

Please change this.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,73 @@
+---
+sidebar_label: git-merge-pr
+description: Merges a specified open pull request.
```
krancour: "Specified open" just doesn't sound right. We can ditch one word or the other and I don't think the meaning will be any less clear.

```suggestion
description: Merges an open pull request.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,73 @@
...
+---
+
+# `git-merge-pr`
+
+`git-merge-pr` merges a specified open pull request. This step commonly follows a
+[`git-open-pr` step](git-open-pr.md) or can be used independently to merge existing
+pull requests.
```
krancour: "or can be used independently to merge existing pull requests" may technically be true, but I think it feels like an invitation to all sorts of bizarre scenarios. Who or what would have opened that PR? We know that the ideal is that Kargo implements the entirety of one's promotion process. I just want our language to not open the door to things we don't recommend.

```suggestion
`git-merge-pr` merges an open pull request. This step commonly follows a
[`git-open-pr` step](git-open-pr.md).
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `repoURL` | `string` | Y | The URL of a remote Git repository. |
+| `provider` | `string` | N | The name of the Git provider to use. Currently `azure`, `bitbucket`, `gitea`, `github`, and `gitlab` are supported. Kargo will try to infer the provider if it is not explicitly specified. |
+| `insecureSkipTLSVerify` | `boolean` | N | Indicates whether to bypass TLS certificate verification when interfacing with the Git provider. Setting this to `true` is highly discouraged in production. |
+| `prNumber` | `integer` | Y | The pull request number to merge. |
+| `wait` | `boolean` | N | If `true`, the step will return a running status instead of failing when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. Default is `false`. Note: The `wait` option is unreliable for repositories hosted on Bitbucket due to API limitations. |
```
krancour:
```suggestion
| `wait` | `boolean` | N | If `true`, the step will return a running status instead of failing when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. Default is `false`. Note: The `wait` option is unreliable for repositories hosted by Bitbucket due to API limitations. |
```

I would also link to docs about timeouts here.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,73 @@
...
+In this example, a pull request is merged immediately if it's ready. If the pull
+request is not ready to merge (e.g., due to pending checks or conflicts), the step
+will fail.
+
+```yaml
+steps:
+  - uses: git-merge-pr
+    config:
+      repoURL: https://github.com/example/repo.git
+      prNumber: 42
```
krancour: I will remind you again that this is the syntax we have settled on throughout the project. Please fix it in other locations as well.

```suggestion
steps:
- uses: git-merge-pr
  config:
    repoURL: https://github.com/example/repo.git
    prNumber: 42
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,73 @@
...
+      wait: true
+```
+
+### Custom Git Provider
+
+This example demonstrates specifying a custom Git provider for `git-merge-pr`. This
+is useful when the provider cannot be inferred from the `repoURL`, such as with
+self-hosted GitLab instances.
```
krancour: Please figure out how to reword this or even drop this section entirely.

This is _purely_ for instructing Kargo on what kind of provider it's interfacing with when that cannot be inferred. The world "custom," however, strongly implies this is for using providers that aren't otherwise supported, which is not the case.

pkg/promotion/runner/builtin/git_pr_merger.go
fykaa: Got it, I've updated the implementation to remove the duplicate retrieval and checks. Now it only relies on the provider's MergePullRequest() logic for handling state

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,72 @@
...
+    repoURL: https://github.com/example/repo.git
+    prNumber: 42
+    wait: true
+```
+
+### Explicit Git Provider
+
+In some cases, `git-merge-pr` cannot determine the Git provider from the `repoURL` alone,
+such as with self-hosted GitLab instances. In these cases, you can explicitly specify
+the provider:
```
krancour: It feels awkward that's we're calling out GitLab specificall here. It's true of course, but this is really a problem for any of the supported Git hosting platforms when they are self-hosted, so calling out GitLab specifically feels as if it has the unintended effect of de-emphasizing the ubiquity of the problem.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,72 @@
+---
+sidebar_label: git-merge-pr
+description: Merges an open pull request.
+---
+
+# `git-merge-pr`
```
krancour, line 7:
```suggestion
<span class="tag beta"></span>
# `git-merge-pr`
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `repoURL` | `string` | Y | The URL of a remote Git repository. |
+| `provider` | `string` | N | The name of the Git provider to use. Currently `azure`, `bitbucket`, `gitea`, `github`, and `gitlab` are supported. Kargo will try to infer the provider if it is not explicitly specified. |
+| `insecureSkipTLSVerify` | `boolean` | N | Indicates whether to bypass TLS certificate verification when interfacing with the Git provider. Setting this to `true` is highly discouraged in production. |
+| `prNumber` | `integer` | Y | The pull request number to merge. |
+| `wait` | `boolean` | N | If `true`, the step will return a running status instead of failing when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. Default is `false`. Note: The `wait` option is unreliable for repositories hosted by Bitbucket due to API limitations. |
```
krancour: I'm worried that the warning about Bitbucket isn't visible enough. Can we please break this out into a caution/warning box below the table?

pkg/gitprovider/gitea/gitea.go
```diff
@@ -257,6 +276,56 @@ func (p *provider) ListPullRequests(
...
+		return nil, false, fmt.Errorf("pull request %d not found", id)
+	}
+
+	switch {
+	case giteaPR.HasMerged:
+		pr := convertGiteaPR(*giteaPR)
+		return &pr, true, nil
+
+	case giteaPR.State != gitea.StateOpen:
+		return nil, false, nil
```
krancour: For three out of five providers, you're treating a PR that hasn't been merged but is now in a non-open state as an error, which seems correct. Is there a reason that you're handling that differently here and for GitLab?

pkg/gitprovider/github/github.go
```diff
@@ -268,6 +295,59 @@ func (p *provider) ListPullRequests(
...
+		return nil, false, fmt.Errorf("unexpected nil merge result")
+	}
+
+	updatedPR, _, err := p.client.GetPullRequests(ctx, p.owner, p.repo, int(id))
+	if err != nil {
+		return nil, false, fmt.Errorf("error getting pull request %d after merge: %w", id, err)
+	}
+	if updatedPR == nil {
+		return nil, false, fmt.Errorf("unexpected nil pull request after merge")
+	}
```
krancour, line 345: I'm raising this here because I'm most familiar with GitHub and I know this is a possibility here. It may be a possibility for other providers as well, though I am less certain of it...

GitHub repos can be using a merge queue. i.e. A PR might not _actually_ be merged immediately after "clicking" merge as we are here.

In the case of GitHub, at least, we probably need to account for that.

I think this requires too much investigation and effort in to realistically have that addressed on time for this feature to be included in v1.8.0, so, I have suggested [here](https://github.com/akuity/kargo/pull/3958/files#r2388088062) that we declare this step as beta.

Please open a follow-up issue for v1.8.1 and assign to yourself.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
fykaa: I tried to use this example to keep things consistent with how the [git-open-pr documentation addresses it as a Custom Git Provider](https://docs.kargo.io/user-guide/reference-docs/promotion-steps/git-open-pr#custom-git-provider)
Should I reword it there as well?

pkg/gitprovider/gitea/gitea.go
fykaa: Thanks for catching that! I did try to review it before committing but must have overlooked this case. I'll get the fixes in right away

pkg/gitprovider/github/github.go
fykaa: that is right this can delay the actual merge, and the PR state may not update immediately. I've updated this to beta for v1.8.0. I'll explore checking queue status via polling or events and update our logic wrt Github's merge-queue api behaviour

I've opened this issue to track this: https://github.com/akuity/kargo/issues/5127

krancour: Thank you!

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -0,0 +1,78 @@
...
+
+```yaml
+steps:
+- uses: git-merge-pr
+  config:
+    repoURL: https://gitlab.example.com/example/repo.git
+    provider: gitlab
+    prNumber: 123
+    wait: true
+```
```
krancour: I know there was previous discussion about this section. I hadn't liked "custom," and you did change that. Thank you.

Few things...

I now realize the original title and text of this section came from existing docs for the `git-open-pr` step. I don't like "custom" there any more than I like it here, for the same reasons, it sounds like something it isn't.

The example used here (and there) is also not a great one. The URL has "gitlab" in it, so Kargo actually would have no trouble inferring the provider, making it unnecessary to be explicit about it in this case. It would be necessary for something like `https://source-control.example.com/example/repo.git` because the URL contains no hint at all as to what platform hosts the repo.

Anyway... I've decided on a "less is more" approach. I think the table higher up on this page adequately describes the `provider` option. Let's remove this entire section please, and please do the same for the existing `git-open-pr` docs.

🙏 


```suggestion
```

krancour (approved):

PR #3968, opened 2025-05-01
chore: pipeline view updates

Marvin9:
- Numbers in freight boxes should only be visible for > 2
- Stage details can be opened by click on node as well
- Stage stacking filter is only visible when it can be used (ie. not in small pipelines)
- Remove "1 Minute" timerange option


rpelczar:
1. I think we should disable the stack button instead of hiding it.
2. I don't like a clickable entire stage node, because we have some clickable elements inside. Maybe we can add button next to 3 dots menu?

rpelczar:
Updated stage node navigation:
image

rpelczar (approved):

Issue #1449, opened 2024-02-02
Warehouse webhook


jessesuen:
# Proposed Feature

When a new git commit or container image is pushed, we would like the Warehouse to refresh immediately to produce the new freight.

# Motivation

Currently, we have to wait several minutes until the next poll interval before a new image is noticed, or when we click "refresh" on the warehouse.

# Suggested Implementation

Introduce an endpoint in Kargo API, which when hit, it will refresh the warehouse.

In terms of implementation, I see two approaches:

1. we could introduce a global endpoint, `https://kargo.example.com/webhook`, which is configured in the container registries. Then teach this endpoint to understand [docker hub](https://docs.docker.com/docker-hub/webhooks/), [quay.io](https://docs.quay.io/guides/notifications.html), [ghcr.io](https://docs.github.com/en/webhooks/webhook-events-and-payloads#package) and other popular container registry webhooks. The endpoint would interpret the payload to refresh the appropriate warehouse.

2. instead of teaching Kargo about all payloads, we would support a convention where the warehouse information is encoded into webhook URL itself, e.g.

https://kargo.example.com/webhook/my-project/my-warehouse

The payload would not matter and could be discarded, because all we care about is refreshing the warehouse.

jessesuen:
Related https://github.com/akuity/kargo/issues/647

krancour:
@jessesuen do you have any thoughts about how to secure this? If we don't this opens Kargo to DOS attacks.

jessesuen:
It's a valid concern. I have a few ideas, not mutually exclusive:

1. Unguessable URLs

The URL could be something like: 
`https://kargo.example.com/webhook/my-project/my-warehouse/ddfd21a9-d0c9-4e4d-ab8c-0d0fffdee6f5`

where `ddfd21a9-d0c9-4e4d-ab8c-0d0fffdee6f5` is a random UUID generated at the time webhook support was enabled for the Warehouse/Project. This in and of itself is not foolproof. URLs might still be logged. But at least this decreases the guessability of a valid webhook receiver URL, and would be sufficient for most purposes.

2. Rate limiting

Webhook refreshes should not occur more frequently than some reasonable number (e.g. once per second). Any requests above the limit could be silently ignored, held up, and have some exponential backoff.

3. Minimize the impact

[...]

PR #3969, opened 2025-05-01
feat: add indexer func for warehouses by subscription repo url

fuskovic:
Related to #1449 

Depends on #3985


krancour:
@fuskovic let's get #3985 merged first and then you can rebase this PR and amend it make the external webhooks server build this index on startup.

If we do things in that order, the new bits of code in this PR don't have the appearance of being dead/unused in the interim.

internal/indexer/indexer.go
```diff
@@ -443,3 +447,42 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
...
+	}
+	return repoURLs
+}
+
+func apppendIfNotExists(ss []string, repoURL string) []string {
+	if !slices.Contains(ss, repoURL) {
+		return append(ss, repoURL)
+	}
+	return ss
+}
```
krancour, line 488: If the same URL were to appear in the list more than once, which is unlikely to begin with, I don't believe it would actually harm anything. I'd probably just kill this bit off.

internal/indexer/indexer.go
```diff
@@ -443,3 +447,42 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
...
+			repoURLs = apppendIfNotExists(repoURLs,
+				helm.NormalizeChartRepositoryURL(sub.Chart.RepoURL),
+			)
+		}
+		if sub.Image != nil && sub.Image.RepoURL != "" {
+			repoURLs = apppendIfNotExists(repoURLs,
+				// The normalization of Helm chart repository URLs can also be used here
+				// to ensure the uniqueness of the image reference as it does the job of
+				// ensuring lower-casing, etc. without introducing unwanted side effects.
+				helm.NormalizeChartRepositoryURL(sub.Image.RepoURL),
```
krancour, line 476: I'm not clear on why we're normalizing an image URL using the chart URL normalization logic. I'm not aware of any place else in the code where we've applied any kind of normalization at all to image URLs.

fuskovic: I saw similar [here](https://github.com/akuity/kargo/blob/main/internal/webhook/kubernetes/warehouse/webhook.go#L299) for image validation logic by @hiddeco but maybe it doesn't apply here?

krancour: Alright. I'd overlooked that spot. There's no one right answer here, probably -- or at least not an easy one.

I understand why @hiddeco did that there. He's trying to ensure that a single Warehouse does not have multiple subscriptions to the same repository (something we don't currently support) and normalizing in the way he is closes the loophole whereby you could subscribe to the same repository twice in one Warehouse simply by using variations of the same URL.

On the flip side there are a lot of other places where we're _not_ doing any kind of image URL normalization -- for instance, [here](https://github.com/akuity/kargo/blob/714a640b50937bff5a966ef84ffac79ee9e3ba97/internal/credentials/kubernetes/database.go#L175-L180). We probably should start... but we should _also_ probably have dedicated image URL normalization logic instead of reusing the chart URL normalization logic.

Let's leave things as you have them for now, although I would suggest rewording the comment as a TODO that states the chart URL normalization logic is adequate for the interim, but that it should be replaced with dedicated image URL normalization logic in the future. If you wouldn't mind also opening an issue so we don't neglect to follow up on that, that would be great.

Last, as you continue work on #1449, don't forget to apply the same normalization as you are here before using this index.

internal/indexer/indexer.go
```diff
@@ -38,6 +40,8 @@ const (
 	StagesByWarehouseField      = "warehouse"
 
 	ServiceAccountsByOIDCClaimsField = "claims"
+
+	WarehouseRepoURLIndexKey = "subscriptions.repoURL"
```
krancour, line 44: This feels like it would be a bit more consistent with those above:

```suggestion
	WarehousesBySubscribedURLsField = "subscribedURLs"
```

internal/indexer/indexer.go
fuskovic: I opened https://github.com/akuity/kargo/issues/3999. It's rather vague at the moment because we probably need to do some triaging around normalization criteria and define scope. I otherwise pushed the update to the TODO and included a link to the issue there as well.

krancour: Thank you!

internal/indexer/indexer.go
```diff
@@ -38,6 +40,8 @@ const (
 	StagesByWarehouseField      = "warehouse"
 
 	ServiceAccountsByOIDCClaimsField = "claims"
+
+	WarehouseRepoURLIndexKey = "subscribedURLs"
```
krancour, line 44:
```suggestion
	WarehousesBySubscribedURLsField = "subscribedURLs"
```

internal/indexer/indexer.go
```diff
@@ -443,3 +447,35 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
 	}
 	return refinedClaimValues
 }
+
+// WarehousesByRepoURL is a client.IndexerFunc that indexes Warehouses by the
+// the RepoURLs they are associated with.
```
krancour, line 452:
```suggestion
// the RepoURLs they subscribe to.
```

internal/indexer/indexer.go
```diff
@@ -443,3 +447,35 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
 	}
 	return refinedClaimValues
 }
+
+// WarehousesByRepoURL is a client.IndexerFunc that indexes Warehouses by the
```
krancour, line 451:
```suggestion
// WarehousesBySubscribedURLs is a client.IndexerFunc that indexes Warehouses by the
```

internal/indexer/indexer.go
```diff
@@ -443,3 +447,35 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
...
+// the RepoURLs they are associated with.
+func WarehousesByRepoURL(obj client.Object) []string {
```
krancour, line 453:
```suggestion
func WarehousesBySubscribedURLs(obj client.Object) []string {
```

internal/indexer/indexer.go
```diff
@@ -443,3 +447,35 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
...
+		if sub.Chart != nil && sub.Chart.RepoURL != "" {
+			repoURLs = append(repoURLs,
+				helm.NormalizeChartRepositoryURL(sub.Chart.RepoURL),
+			)
+		}
+		if sub.Image != nil && sub.Image.RepoURL != "" {
+			repoURLs = append(repoURLs,
+				// TODO(fuskovic): the chart URL normalization logic is adequate for the interim,
+				// but that it should be replaced with dedicated image URL normalization logic in the future.
+				// See https://github.com/akuity/kargo/issues/3999
```
krancour, line 475:
```suggestion
				// TODO(fuskovic): This chart URL normalization logic is adequate for
				// normalizing image URLs in the near term, but should eventually be
				// replaced with dedicated image URL normalization logic.
				// See https://github.com/akuity/kargo/issues/3999
```

cmd/controlplane/external_webhooks.go
```diff
@@ -75,7 +77,15 @@ func (o *externalWebhooksServerOptions) run(ctx context.Context) error {
 		return fmt.Errorf("error creating Kubernetes client: %w", err)
 	}
 
-	// TODO: Add all indices required by the external webhooks server
+	err = cluster.GetFieldIndexer().IndexField(
+		ctx,
+		&kargoapi.Warehouse{},
+		indexer.WarehouseRepoURLIndexKey,
```
krancour, line 83:
```suggestion
		indexer.WarehousesBySubscribedURLsField,
```

cmd/controlplane/external_webhooks.go
```diff
@@ -75,7 +77,15 @@ func (o *externalWebhooksServerOptions) run(ctx context.Context) error {
...
+		indexer.WarehousesByRepoURL,
```
krancour, line 84:
```suggestion
		indexer.WarehousesBySubscribedURLs,
```

krancour (approved):
🔥

PR #3974, opened 2025-05-02
chore: pipeline view updates

Marvin9:
- Less timerange option in filters
- white text color when color is used
- promotion manifest in promotion modal


ui/src/features/project/pipelines/freight/filter-timerange-utils.ts
```diff
@@ -59,10 +59,7 @@ export const timerangeToLabel = (timerange: timerangeTypes) => {
 };
 
 export const timerangeOrderedOptions: timerangeTypes[] = [
-  '15-minutes',
-  '30-minutes',
   '1-hour',
-  '12-hours',
```
rpelczar, line 65: Can we also delete it from the types and timerangeToDate function?

rpelczar (approved):

PR #3985, opened 2025-05-05
feat: add external webhooks server component

krancour:
Related to #1449.

This PR adds some scaffolding for the new "external webhooks server." @fuskovic will flesh that out further over a series of follow-up PRs.

Up to now, "webhook server" has referred to the Kargo component that serves the logic for defaulting/validating _Kubernetes_ webhooks. Now that we will also be receiving webhooks from external platforms like GitHub, Docker Hub, GitLab, etc., we need to start distinguishing clearly between the components serving those two, distinctly different sorts of webhooks. To that end, the existing webhook server is being rebranded as the "kubernetes webhooks server" while the new component will be known as the "external webhooks server." Code and the chart are refactored accordingly, although there are some limits to this reorganization. To minimize the likelihood of any backwards incompatibilities, _no existing resources are renamed._ e.g. Any resource named `kargo-webhooks-server` retains its existing name. The `webhooksServer` section of `values.yaml` is also untouched for the same reason.

This PR also incorporates ngrok into our `Tiltfile` to facilitate easily connecting external platforms to the external webhooks server running locally with Tilt.


Tiltfile
```diff
@@ -183,9 +202,12 @@ k8s_resource(
 k8s_resource(
   new_name = 'crds',
   objects = [
+    'clusterpromotiontasks.kargo.akuity.io:customresourcedefinition',
     'freights.kargo.akuity.io:customresourcedefinition',
+    'projectconfigs.kargo.akuity.io:customresourcedefinition',
     'projects.kargo.akuity.io:customresourcedefinition',
     'promotions.kargo.akuity.io:customresourcedefinition',
+    'promotiontasks.kargo.akuity.io:customresourcedefinition',
```
krancour, line 210: This is incidental cleanup that seemed ok to tackle while already touching this file.

charts/kargo/templates/common/cert-issuer.yaml
```diff
@@ -1,5 +1,5 @@
 
-{{- if or (and .Values.api.enabled .Values.api.ingress.enabled .Values.api.ingress.tls.enabled .Values.api.ingress.tls.selfSignedCert) (and .Values.webhooksServer.enabled .Values.webhooksServer.tls.selfSignedCert) (and .Values.api.enabled .Values.api.oidc.enabled .Values.api.oidc.dex.enabled .Values.api.oidc.dex.tls.selfSignedCert) }}
+{{- if or (and .Values.api.enabled .Values.api.tls.enabled .Values.api.tls.selfSignedCert) (and .Values.api.enabled .Values.api.ingress.enabled .Values.api.ingress.tls.enabled .Values.api.ingress.tls.selfSignedCert) (and .Values.api.enabled .Values.api.oidc.enabled .Values.api.oidc.dex.enabled .Values.api.oidc.dex.tls.selfSignedCert) (and .Values.externalWebhookServer.enabled .Values.externalWebhookServer.tls.enabled .Values.externalWebhookServer.tls.selfSignedCert) (and .Values.externalWebhookServer.enabled .Values.externalWebhookServer.ingress.enabled .Values.externalWebhookServer.ingress.tls.enabled .Values.externalWebhookServer.ingress.tls.selfSignedCert) (and .Values.webhooksServer.enabled .Values.webhooksServer.tls.selfSignedCert) }}
```
krancour, line 2: Apart from enabling the cert issuer in the event of new external-webhook-server-related scenarios, there's a small bug fix in here...

`(and .Values.api.enabled .Values.api.tls.enabled .Values.api.tls.selfSignedCert)` was previously unaccounted for.

Tiltfile
```diff
@@ -1,3 +1,7 @@
+if 'ENABLE_NGROK_EXTENSION' in os.environ and os.environ['ENABLE_NGROK_EXTENSION'] == '1':
```
hiddeco, line 1: Not sure if this is actually required, as you still have to manually start the ngrok tunnel from within the UI as far as I know.

Tiltfile
```diff
@@ -1,3 +1,7 @@
...
+  v1alpha1.extension_repo(name = 'default', url = 'https://github.com/tilt-dev/tilt-extensions')
```
hiddeco, line 2: [Given it listens to `:4040` by default](https://github.com/tilt-dev/tilt-extensions/tree/master/ngrok#custom-port), I am not sure this works for the further configuration. Or at least, I could not get it to work.

Tiltfile
fuskovic: @hiddeco the ngrok resources(operator, config, and status) don't start without enabling this for me

hiddeco: My point is that no outbound tunnel is created if you enable this by default. I.e., if you remove this line, you would have a single way to enable this (by pressing the button in the UI), versus now having to set both the environment variable _and_ press the button.

Tiltfile
fuskovic: Clicking "start ngrok" initially failed for me too:

```
+ ngrok start --config /var/folders/_v/d5c4bh0j3yq7yjm8_59f3jz40000gn/T//tilt-ngrok.nk1GHlCa4 --all
ERROR:  authentication failed: Your account is limited to 1 simultaneous ngrok agent sessions.
ERROR:  You can run multiple simultaneous tunnels from a single agent session by defining the tunnels in your agent configuration file and starting them with the command `ngrok start --all`.
ERROR:  Read more about the agent configuration file: https://ngrok.com/docs/secure-tunnels/ngrok-agent/reference/config
ERROR:  You can view your current agent sessions in the dashboard:
ERROR:  https://dashboard.ngrok.com/agents
ERROR:  
ERROR:  ERR_NGROK_108
ERROR:  https://ngrok.com/docs/errors/err_ngrok_108
ERROR:  
```

Then after clicking `trigger update` on the resource `ngrok:status` on the detail page it seemed to start ok

Screenshot 2025-05-06 at 9 35 48 AM

This behavior seems to be consistent.

Tiltfile
fuskovic: For me when I don't enable this extension, the "start ngrok" button doesn't render at all, along with the ngrok resources I mentioned.

fuskovic: @krancour Spoke with @hiddeco offline. I see what he means. No harm in just always turning on these resources regardless of exporting that env var.

krancour: Admittedly, I copied the condition from an older project of mine. I must have had some rational reason for it being conditional then, but as I can no longer remember what that was, I'm happy to make this unconditional.

krancour: Fixed in 36b4b712f115248efc90c8515fbc7ec0d44a254b

hiddeco:
Other weird thing: I keep running into HTTP 418 issues using Orbstack — while I _did_ log in to `ngrok`. The configured token is correct, as I tested it using `ngrok http`. Clicking `trigger update` or restarting Tilt does not appear to have any effect, and I do wonder if this is because Tilt somehow runs under a different user and cannot find the `ngrok` configuration file with the token.

image

Tiltfile
samcrichard: Sam from ngrok here! Can we offer you some help troubleshooting?

hiddeco:
The problem is that `ngrok config add-authtoken` (on macOS) creates a config file in `/Users/<username>/Library/Application Support/ngrok/ngrok.yml` while the Tilt plugin set to:

> `--default_config_file=~/.config/ngrok/my-default-config.yml`: Sets the path to the default configuration file for Ngrok. Default `~/.config/ngrok/ngrok.yml`.

Moving that file to `~/.config/ngrok/ngrok.yml` makes it work, but breaks `ngrok` itself.

---

However, once the agent is running (on `:4040`), and I enable the tunnels for either the `api` or `external-webhook-server`. I still can not reach any of the endpoints using the URLs advertised on `http://localhost:4040/inspect/http`.

Tiltfile
krancour: Actually... I remember now why I had the conditional. It was to keep all ngrok integration strictly opt-in so that Tilt wouldn't unconditionally fail if `ngrok` wasn't installed.

hiddeco: Maybe better to then check for the presence of the command? https://docs.tilt.dev/snippets.html?local#snip_require_local_tool

krancour:
@samcrichard any assistance is certainly welcome. Not sure how familiar you are with the ngrok extension for Tilt. I suspect our issue are more with the extension than with ngrok itself.

krancour:
The ngrok extension for Tilt hasn't been touched in years. I think it was written for an older version of the ngrok agent, and I am now thinking that might be the source of some of the trouble here.

It may be better to leave ngrok integration out of the Tiltfile and if/when we're hacking away locally on the external webhooks server, we can just start the ngrok agent manually.

hiddeco (approved):

fuskovic (approved):

Makefile
```diff
@@ -417,6 +417,10 @@ hack-uninstall-argocd: install-helm
 hack-uninstall-cert-manager: install-helm
 	$(HELM) delete cert-manager --namespace cert-manager
 
+.PHONY: hack-ngrok
+hack-ngrok:
+	ngrok http --hostname=$(KARGO_EXTERNAL_WEBHOOKS_SERVER_HOSTNAME) 30083
```
hiddeco, line 422: Can you put this at the top of this file as `KARGO_EXTERNAL_WEBHOOKS_SERVER_HOSTNAME ?=`

krancour: Fixed in 8a2611cc89559318bb10986f8cee10f12cbc8044

Issue #3964, opened 2025-05-01
bug: errant reference to erstwhile magical behavior of kustomize-set-image step


krancour:
This bit of docs references some "magical" behavior from the pre-expression days.

WRT to the `tag` field:

> A tag naming a specific revision of image. Mutually exclusive with digest. If none of these are specified, the tag specified by a piece of Freight referencing image will be used as the value of this field.

And WRT `digest`:

> A digest naming a specific revision of image. Mutually exclusive with tag. If none of these are specified, the tag specified by a piece of Freight referencing image will be used as the value of tag.

https://docs.kargo.io/user-guide/reference-docs/promotion-steps/kustomize-set-image/#configuration

But this magical behavior was deprecated in v1.1.0 and removed in v1.3.0, as documented here:

https://docs.kargo.io/release-notes/v1.3.0/

We need to get the `kustomize-set-image` docs caught up to reality.

PR #3992, opened 2025-05-06
docs: fix errant reference to magical behavior of kustomize-set-image

fykaa:
closes https://github.com/akuity/kargo/issues/3964


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/kustomize-set-image.md
```diff
@@ -17,8 +17,8 @@ to executing `kustomize edit set image`. This step is commonly followed by a
...
-| `images[].tag` | `string` | N | A tag naming a specific revision of `image`. Mutually exclusive with `digest`. If none of these are specified, the tag specified by a piece of Freight referencing `image` will be used as the value of this field.                                                                                                                          |
-| `images[].digest` | `string` | N | A digest naming a specific revision of `image`. Mutually exclusive with `tag`. If none of these are specified, the tag specified by a piece of Freight referencing `image` will be used as the value of `tag`.                                                                                                                               |
+| `images[].tag` | `string` | N | A tag naming a specific revision of `image`. Mutually exclusive with `digest`. Users must explicitly specify this field or use an [expression](../../60-reference-docs/40-expressions.md#imagefromrepourl-freightorigin).                                                                                                                          |
```
krancour, line 20: The wording here is awkward (same of the other field, below). It doesn't sound right that "Users must explicitly specify this field _**OR**_ use an expression." Isn't it more the case that it would be explicitly specified _**often using**_ an expression?

"must explicitly specify" also makes this sound required, when the real requirement is that _one and only one_ of the `tag` or `digest` field is specified.

krancour (approved):

Issue #3221, opened 2025-01-07
fix: enumerate allowed values in generated references


fykaa:
wrt the comment : https://github.com/akuity/kargo/issues/2880#issuecomment-2488332266
related PR: https://github.com/akuity/kargo/pull/2969

# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.

# Description

The current [generated reference](https://doc.crds.dev/github.com/akuity/kargo) lacks enumeration of allowed values for specific fields. This issue proposes adding metadata tweaks to enumerate these allowed values

PR #3997, opened 2025-05-06
chore: document accepted values for certain crd fields

fykaa:
closes https://github.com/akuity/kargo/issues/3221

### DIFF
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -309,7 +309,7 @@ type FreightSources struct {
 	//    one upstream Stage to be considered available for promotion.
 	// - "": Treated the same as "OneOf".
 	//
-	// +kubebuilder:validation:Enum=All;OneOf;""
+	// +kubebuilder:validation:Optional
 	AvailabilityStrategy FreightAvailabilityStrategy `json:"availabilityStrategy,omitempty" protobuf:"bytes,4,opt,name=availabilityStrategy"`
 }
 
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -300,6 +300,7 @@ type FreightSources struct {
 	// AvailabilityStrategy specifies the semantics for how requested Freight is
 	// made available to the Stage. This field is optional. When left unspecified,
 	// the field is implicitly treated as if its value were "OneOf".
+	// Accepted Values: "All", "OneOf", ""
 	//
 	// +kubebuilder:validation:Optional
 	AvailabilityStrategy FreightAvailabilityStrategy `json:"availabilityStrategy,omitempty" protobuf:"bytes,4,opt,name=availabilityStrategy"`
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -303,9 +303,11 @@ type FreightSources struct {
 	//
 	// Accepted Values:
 	//
-	// - "All": All requested Freight must be available before proceeding.
-	// - "OneOf": At least one requested Freight must be available before proceeding.
-	// - "": Treated the same as "OneOf" (for backward compatibility).
+	// - "All": Freight must be be verified and, if applicable, soaked in all
+	//   upstream Stages to be considered available for promotion.
+	// - "OneOf": Freight must be verified and, if applicable, soaked in at least
+	//    one upstream Stage to be considered available for promotion.
+	// - "": Treated the same as "OneOf".
 	//
 	// +kubebuilder:validation:Enum=All;OneOf;""
 	AvailabilityStrategy FreightAvailabilityStrategy `json:"availabilityStrategy,omitempty" protobuf:"bytes,4,opt,name=availabilityStrategy"`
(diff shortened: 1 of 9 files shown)


api/v1alpha1/stage_types.go
```diff
@@ -300,6 +300,7 @@ type FreightSources struct {
 	// AvailabilityStrategy specifies the semantics for how requested Freight is
 	// made available to the Stage. This field is optional. When left unspecified,
 	// the field is implicitly treated as if its value were "OneOf".
+	// Accepted Values: "All", "OneOf", ""
```
krancour, line 303: Is this the only place where we use enumerated values without defining what they are?

It is probably worthwhile to explain, as we do for image selection strategies, what the meaning of each value is.

krancour: > I don't see explanation in the imageSelectionStrategy:

🤦‍♂️ It's defined somewhere internally:

https://github.com/akuity/kargo/blob/dfc75c46d0467ba1e98fec7abf9907976bfd0185/internal/image/selector.go#L13-L39

These explanations really ought to be copied into `warehouse_types.go` so that they'll show up in the generated docs.

Looks like similar is true for commit selection strategies are also missing explanations of their values.

api/v1alpha1/stage_types.go
```diff
@@ -301,7 +301,13 @@ type FreightSources struct {
 	// made available to the Stage. This field is optional. When left unspecified,
 	// the field is implicitly treated as if its value were "OneOf".
 	//
-	// +kubebuilder:validation:Optional
+	// Accepted Values:
+	//
+	// - "All": All requested Freight must be available before proceeding.
+	// - "OneOf": At least one requested Freight must be available before proceeding.
+	// - "": Treated the same as "OneOf" (for backward compatibility).
```
krancour, line 308:
```suggestion
--------------------------------------------------------------------------------
	// - "All": Freight must be be verified and, if applicable, soaked in all
	//   upstream Stages to be considered available for promotion.
	// - "OneOf": Freight must be verified and, if applicable, soaked in at least
	//    one upstream Stage to be considered available for promotion.
	// - "": Treated the same as "OneOf".
```

api/v1alpha1/warehouse_types.go
```diff
@@ -67,9 +67,14 @@ type WarehouseSpec struct {
 	// FreightCreationPolicy describes how Freight is created by this Warehouse.
 	// This field is optional. When left unspecified, the field is implicitly
 	// treated as if its value were "Automatic".
-	// Accepted values: Automatic, Manual
+	//
+	// Accepted values:
+	//
+	// - "Automatic": Freight is automatically created in response to detected changes in the source repository.
+	// - "Manual": Freight must be created explicitly by a user or an external system.
```
krancour, line 74:
```suggestion
--------------------------------------------------------------------------------
	// - "Automatic": New Freight is created automatically when any new artifact
	//   is discovered.
	// - "Manual": New Freight is never created automatically.
```

api/v1alpha1/warehouse_types.go
```diff
@@ -112,9 +117,22 @@ type GitSubscription struct {
...
+	// - `"NewestFromBranch"`: Selects the latest commit on the specified branch,
+	//   or the default branch if none is specified. This is the default strategy.
+	//
+	// - `"SemVer"`: Selects the commit pointed to by the tag with the highest
+	//   semantic version. An optional constraint can narrow the version range.
+	//
+	// - `"Lexical"`: Selects the commit referenced by the lexicographically greatest
+	//   tag. Useful when tags encode timestamps or dates.
+	//
+	// - `"NewestTag"`: Selects the commit pointed to by the most recently created tag.
```
krancour, line 132:
```suggestion
	// - "NewestFromBranch": Selects the latest commit on the specified branch, or
	//   the default branch if none is specified. This is the default strategy.
	//
	// - "SemVer": Selects the commit referenced by the the semantically greatest
	//   tag. An optional constraint can narrow the acceptable range of versions.
	//
	// - "Lexical": Selects the commit referenced by the lexicographically
	//   greatest tag. Useful when tags embed a _leading_ date or timestamp. An
	//   optional constraint can narrow acceptable tags.
	//
	// - "NewestTag": Selects the commit referenced by the most recently created
	//   tag. An optional constraint can narrow acceptable tags.
```

charts/kargo/resources/crds/kargo.akuity.io_stages.yaml
```diff
@@ -286,15 +286,27 @@ spec:
...
                         availabilityStrategy:
+                          allOf:
+                          - enum:
+                            - All
+                            - OneOf
+                            - ""
+                          - enum:
+                            - All
+                            - OneOf
+                            - ""
```
krancour, line 297: I notice this is duplicated now. I think I know why... I'll link to here from the spot where I think it happened.

api/v1alpha1/stage_types.go
```diff
@@ -301,6 +301,13 @@ type FreightSources struct {
 	// made available to the Stage. This field is optional. When left unspecified,
 	// the field is implicitly treated as if its value were "OneOf".
 	//
+	// Accepted Values:
+	//
+	// - "All": Freight must be be verified and, if applicable, soaked in all
+	//   upstream Stages to be considered available for promotion.
+	// - "OneOf": Freight must be verified and, if applicable, soaked in at least
+	//    one upstream Stage to be considered available for promotion.
+	// - "": Treated the same as "OneOf".
```
krancour, line 310: Sorry. I know this means running codegen again, but these are all more readable if we leave a line between the free-form comments and the kubebuilder markers.
```suggestion
	// - "": Treated the same as "OneOf".

```

krancour (approved):

PR #4001, opened 2025-05-07
feat: improve step alias error messages

### LINKED ISSUE #3402: Add step numbers to the left of the steps on the promotion details page
# Proposed Feature

Can we add step numbers to the left of the steps on the promotion details page? It would make it easier to determine which one the error is referring to (I know there is already and X, but the numbers would also be nice because it uses the step numbers in the error message).

Speaking about the error messages, the step numbers are not correct for human reading since it is 0-based. we should add 1 to the step number in the error message so that it is accurate.

Image

# Motivation

It will make it easier to read and use.

# Suggested Implementation

- Add numbers next to the steps
- fix the step numbers in the error message so it is correct (add 1, since it is 0-based)

krancour:
Fixes #3402

This PR assigns generated step aliases as a `promotionTemplate` and tasks it references are expanded instead of doing so just-in-time.

This allows for slightly easier and easier-to-follow bookkeeping in terms of step execution metadata.

More significantly, it allows any error messages we build that involve steps to use aliases instead of indices. Indices, according to #3402, were confusing because the UI didn't display them and the question of being 0-based vs 1-based was possibly also the source of some confusion. This change alleviates all of that.

Screenshot 2025-05-01 at 5 10 24 PM

### DIFF
--- a/api/v1alpha1/promotion_types.go
+++ b/api/v1alpha1/promotion_types.go
@@ -288,9 +288,9 @@ func (s *PromotionStep) GetAlias(i int) string {
 	case s.As != "":
 		return s.As
 	case s.Task != nil:
-		return fmt.Sprintf("task-%d", i)
+		return fmt.Sprintf("task-%d", i+1)
 	default:
-		return fmt.Sprintf("step-%d", i)
+		return fmt.Sprintf("step-%d", i+1)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


jessesuen (approved):
Love these small QoL improvements

rpelczar (approved):

Issue #4016, opened 2025-05-08
Kargo v1.5 UI lost ability to delete freight


jessesuen:
# Description

It is no longer possible to delete Freight in UI in v1.5

# Screenshots

In v1.4:
<img width="231" alt="Image" src="https://github.com/user-attachments/assets/0830cc01-7dad-47d2-9d81-d8cdfcf033dd" />

In v1.5:
<img width="266" alt="Image" src="https://github.com/user-attachments/assets/986c9a18-c754-4174-8f5e-e7964b1abb86" />

# Version

```
kargo-unstable:v1.6.0-unstable-20250505
```

# Logs

```
Paste any relevant application logs here.
```

Issue #4011, opened 2025-05-07
bug: warehouse filter "x" doesn't remove warehouse name from filter box


krancour:
I can place my cursor in the box and remove the filter criteria by backspacing, but clicking the x does not seem to have the effect one would presume.

![Image](https://github.com/user-attachments/assets/28846f9b-36f3-493b-aa41-75635c556198)

rpelczar:
https://github.com/user-attachments/assets/cbdd8ead-886d-4925-9f89-b587967478b2

I think it's related. I can click hidden filter checkboxes.

Issue #4010, opened 2025-05-07
bug: warehouses are no longer color coded


krancour:
Prior to the pipeline view revamp, the border on a Warehouse was a color match for the line flowing out of it and into/through the Stages. This made it easy to visually identify the flow of artifacts from a given Warehouse through the Stages.

The colored border is now missing.

![Image](https://github.com/user-attachments/assets/0234fe0e-be8a-456d-b18c-471d78d3407f)

Issue #4009, opened 2025-05-07
bug: stage with no freight mistakenly shows verification in progress


krancour:
![Image](https://github.com/user-attachments/assets/956e3d02-1740-443e-ba6b-7bcc0efaa292)

PR #4017, opened 2025-05-08
fix: pipeline view bugs

Marvin9:
fixes

https://github.com/akuity/kargo/issues/4016
https://github.com/akuity/kargo/issues/4011
https://github.com/akuity/kargo/issues/4010
https://github.com/akuity/kargo/issues/4009

### DIFF
--- a/ui/src/features/project/pipelines/freight/delete-freight-modal.tsx
+++ b/ui/src/features/project/pipelines/freight/delete-freight-modal.tsx
@@ -1,0 +1,55 @@
+import { useMutation } from '@connectrpc/connect-query';
+import { Alert, Modal, message } from 'antd';
+import { useParams } from 'react-router-dom';
+
+import { ModalComponentProps } from '@ui/features/common/modal/modal-context';
+import { getAlias } from '@ui/features/common/utils';
+import { deleteFreight } from '@ui/gen/api/service/v1alpha1/service-KargoService_connectquery';
+import { Freight } from '@ui/gen/api/v1alpha1/generated_pb';
+
+export const DeleteFreightModal = ({
+  visible,
+  hide,
+  onDelete,
+  freight
+}: ModalComponentProps & { onDelete: () => void; freight: Freight }) => {
+  const { name: project } = useParams();
+  const { mutate: deleteAction, isPending } = useMutation(deleteFreight, {
+    onSuccess: () => {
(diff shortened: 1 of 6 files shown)


ui/src/features/project/pipelines/freight/freight-card.tsx
```diff
@@ -72,6 +78,22 @@ export const FreightCard = (props: FreightCardProps) => {
 
   let CardContent: ReactNode;
 
+  const items: ItemType[] = [];
+
+  if (!props.inUse) {
```
jmalisz-akuity, line 83: I'm not sure how it was working before, but I think that adding delete option only when freight isn't in use is confusing. I think it would be better if it was always there, but if it's in use it should be disabled (and maybe have a tooltip that the freight is currently used).

jmalisz-akuity (commented):
Looks good, but @rpelczar should probably also take a look as he has better context

jmalisz-akuity (approved):

jmalisz-akuity (approved):

hiddeco (approved):

ui/src/features/project/pipelines/freight/delete-freight-modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+        banner
+        message={
+          <div>
+            Are you sure you want to delete freight{' '}
+            <span className='font-semibold'>{alias ? alias : freight?.metadata?.name}</span>?
+          </div>
+        }
+        className='mb-4'
+        showIcon={false}
+      />
```
rpelczar, line 52: Be consistent, please. We don't use error alerts for confirmation modals.

ui/src/features/project/pipelines/pipelines.tsx
```diff
@@ -177,7 +177,7 @@ export const Pipelines = (props: { creatingStage?: boolean; creatingWarehouse?:
             />
 
             <div className='w-full h-full relative'>
-              <Flex gap={12} className='absolute z-10 top-2 right-2 left-2'>
+              <Flex gap={12} className='absolute z-50 top-2 right-2 left-2'>
```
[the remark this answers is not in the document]
rpelczar: Ohh, I understand now. It is a fix to this: https://github.com/akuity/kargo/issues/4011

PR #4028, opened 2025-05-08
fix: make error message more consistent

fuskovic:
Follow up for https://github.com/akuity/kargo/pull/4014/files#r2080156596

### DIFF
--- a/cmd/controlplane/external_webhooks.go
+++ b/cmd/controlplane/external_webhooks.go
@@ -78,7 +78,7 @@ func (o *externalWebhooksServerOptions) run(ctx context.Context) error {
 	}
 
 	if err = kargoapi.AddToScheme(cluster.GetClient().Scheme()); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/controlplane/external_webhooks.go
```diff
@@ -78,7 +78,7 @@ func (o *externalWebhooksServerOptions) run(ctx context.Context) error {
 	}
 
 	if err = kargoapi.AddToScheme(cluster.GetClient().Scheme()); err != nil {
-		return fmt.Errorf("failed to add custom kargo type to runtime scheme: %w", err)
+		return fmt.Errorf("error adding Kargo API to Kargo manager scheme: %w", err)
```
krancour, line 81: The controller uses multiple managers and clients, so its error messages differentiate between errors setting up the scheme for one vs the other. We don't need that differentiation here.

```suggestion
		return fmt.Errorf("error adding Kargo API to scheme: %w", err)
```

Issue #4013, opened 2025-05-07
feat: there's currently nowhere to view or edit the ProjectConfiguration CR


krancour:

krancour:
Closed by #4036

PR #4036, opened 2025-05-09
feat: ProjectConfig ui and API

Marvin9:
https://github.com/akuity/kargo/issues/4013

Screenshot 2025-05-09 at 12 07 27 PM


Marvin9:
@rpelczar I have put the config below project yaml and in general tab as it was originally split from that but let me know if it makes sense to have it in different dedicated tab.

hiddeco (commented):
Backend part of this looks 👍

ui/src/features/project/settings/views/general/general-settings.tsx
```diff
@@ -1,41 +1,82 @@
...
+  newErrorHandler((err) => {
+    if (err.code === Code.NotFound) {
+      // ignore
+      return;
+    }
+
+    const errorMessage = err instanceof ConnectError ? err.rawMessage : 'Unexpected API error';
+    notification.error({ message: errorMessage, placement: 'bottomRight' });
+  })
+);
```
rpelczar, line 46: Can you move this into a separate file?
Also, please add a comment on why we need this.

rpelczar (changes requested):

Marvin9:
@krancour @hiddeco ProjectConfig does get created automatically? If yes then when I create new project and hit this API I always get not found

ui/src/features/project/settings/views/general/general-settings.tsx
rpelczar, line 1: I think you forgot to revert the changes in this file.

rpelczar (approved):
UI LGTM

PR #4038, opened 2025-05-09
docs: update 20-openid-connect.md

Utwo:
Fix docs for the OpenID connector.


docs/docs/40-operator-guide/40-security/20-openid-connect.md
```diff
@@ -87,7 +87,7 @@ When installing Kargo with Helm, all options related to OIDC are grouped under
         If you needed to register two separate clients, use the client ID
         associated with the UI.
```
krancour, line 88:
```suggestion
        If you needed to register two separate separate clients with your IDP,
        additionally set `api.oidc.cliClientID` to the second client's ID.
```

krancour (approved):

PR #4047, opened 2025-05-09
chore(controller): remove redundant condition in http step logic

krancour:
This condition was already handled a few lines above.


fuskovic (approved):
LGTM

PR #4048, opened 2025-05-09
fix(api): allow ssh://git@example:1234/repo.git urls

mogsie:
Fixes #4039

PR #4050, opened 2025-05-10
fix(api): verify RepoURL validation

mogsie:
From [this comment](https://github.com/akuity/kargo/pull/4048#discussion_r2082458194):

> Anything you are willing to contribute to move the needle on making regexes more reliable and more thoroughly tested is welcome.

I had a go at this:

* It uses go's AST to pull in the source code of the types file
* The field in question gets a "// +akuity:IsGitRepoURL" comment
* The test reads the source code, looks for the comment
* Extracts the regexp
* Runs the regexp against a table test

I think it pointed out some interesting edge cases (see the table tests)

It also verifies that the two RepoURLs that point to git are identical.

### DIFF
--- a/api/v1alpha1/regexp_test.go
+++ b/api/v1alpha1/regexp_test.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
-	"github.com/akuity/kargo/api/v1alpha1/testhelper"
+	testingPkg "github.com/akuity/kargo/api/testing"
 )
 
 // ValidateResourceExpression checks the regular expression identified by marker
@@ -20,7 +20,7 @@ import (
 func ValidateResourceExpression(t *testing.T, marker string, testCases map[string]bool) {
 	expression, err := findExpression(marker)
 	require.NoError(t, err)
-	testhelper.ValidateRegularExpression(t, expression, testCases)
+	testingPkg.ValidateRegularExpression(t, expression, testCases)
 }
 
 func TestBranchPattern(t *testing.T) {
--- a/api/v1alpha1/regexp_test.go
+++ b/api/v1alpha1/regexp_test.go
@@ -1,0 +1,294 @@
+package v1alpha1_test
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/akuity/kargo/api/v1alpha1/testhelper"
+)
+
+// ValidateResourceExpression checks the regular expression identified by marker
+// against the provided test cases.
+func ValidateResourceExpression(t *testing.T, marker string, testCases map[string]bool) {
+	expression, err := findExpression(marker)
+	require.NoError(t, err)
+	testhelper.ValidateRegularExpression(t, expression, testCases)
+}
+
+func Test_Pattern_Branch(t *testing.T) {
+	testCases := map[string]bool{
+		"":             false,
+		"foo/bar":      true,
+		"foo.bar":      true,
+		"release-0.58": true,
+		"/foo":         false,
+		"foo/":         false,
+		"foo//bar":     true,
+		".foo":         false,
+		"foo.":         false,
+		"foo..bar":     true,
+	}
+
+	ValidateResourceExpression(t, "Branch", testCases)
+}
+
+func Test_Pattern_Digest(t *testing.T) {
+	testCases := map[string]bool{
+		"":                        false,
+		"sha256:":                 false,
+		"sha256:1234567890abcdef": true,
+		":1234567890abcdef":       false,
+		"sha256:xyz":              false,
+		"xyz:1234567890abcdef":    true,
+		"sha256:1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef": true,
+	}
+
+	ValidateResourceExpression(t, "Digest", testCases)
+}
+
+func Test_Pattern_Duration(t *testing.T) {
+	testCases := map[string]bool{
+		"":  false,
+		"s": false,
+		".": false,
+		// simple cases
+		"10s": true,
+		"20h": true,
+		// Decimal cases
+		"1.5s": true,
+		".5s":  false,
+		"0.5s": true,
+		"1.5":  false,
+		"1.5h": true,
+		"1.5m": true,
+		"1,4s": false,
+		// edge cases
+		"1":    false,
+		"0":    false,
+		"-1":   false,
+		"-1s":  false,
+		"+33s": false,
+	}
+
+	ValidateResourceExpression(t, "Duration", testCases)
+}
+
+func Test_Pattern_HelmRepoURL(t *testing.T) {
+	testCases := map[string]bool{}
+
+	ValidateResourceExpression(t, "HelmRepoURL", testCases)
+}
+
+func Test_Pattern_GitRepoURL(t *testing.T) {
+	testCases := map[string]bool{
+		"":             false,
+		":":            false,
+		"/etc/passwd":  false,
+		"//etc/passwd": false,
+		"https:":       false,
+
+		"https://not a url":                      false,
+		"http://github.com/example/repo?foo=bar": true,
+		"ssh://not a url":                        false,
+		"ssh://github.com/example/repo?foo=bar":  true,
+		"not even remotely a url":                false,
+		// URLs of the form http[s]://[proxy-user:proxy-pass@]host.xz[:port][/path/to/repo[.git][/]]
+		"https://github.com":          false,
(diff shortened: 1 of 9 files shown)


krancour:
@mogsie this is a neat approach. I love it! I wonder if it could be generalized to test not just repo URL regexes, but any regexes we have on resource types.

I'm thinking something like put some identifier in the comment that correlates to a table of valid and invalid example strings. Then your existing test function needs only minor tweaks to match comments to a tables and rip through the examples.

The nice thing about this would be that testing new or updated regexes requires no complex code. You only need to add a comment and example strings.

Bonus points if the actual table processing/testing is a separate utility function from the AST stuff so that it can be invoked manually from test suites elsewhere in the code base that involve regexes that _aren't_ on resource types.

mogsie:
I'm on it!  I just wanted early feedback if something like this was in line with your thinking.  I'll generalize it a bit and add some tests to another object.

krancour:
> I just wanted early feedback if something like this was in line with your thinking.

Totally. I really like this.

mogsie:
I ended up with a separate file that tests these expressions, since the expressions themselves aren't strictly tied to a _types.go file.

It now does the following:
* All comments in all **_types.go** files are loaded.
* Patterns must now use backticks.
* If the marker is used more than once, all usages (patterns) with the same marker must be identical.  It reports the location (filename+linenumber) of the violations.
* Patterns are then evaluated against the set of tests.

api/v1alpha1/regexp_test.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	expression, err := findExpression("Duration")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for tt, expected := range testCases {
+		t.Run(tt, func(t *testing.T) {
+			require.Equal(t, expected, expression.MatchString(tt))
```
mogsie, line 87: > Bonus points if the actual table processing/testing is a separate utility function from the AST stuff

I thought about it, and came to the conclusion that extracting a utility function to run these tests would just make it a bit too magical; wouldn't be worth it.

In theory, you could lump all of these regex tests in one huge table-of-table tests.  I will do it if you say so, but I thought this was more readable.

mogsie:
I found a branch regexp101 links in 3235, but the results were all over the place.  I won't edit the patterns, but willl adjust the tests to at least document what the current state of things are.

Branch:

``` go
	testCases := map[string]bool{
		"":             false,
		"foo/bar":      true,
		"foo.bar":      true,
		"release-0.58": true,
		"/foo":         false,
		"foo/":         false,
		"foo//bar":     true,
		".foo":         false,
		"foo.":         false,
		"foo..bar":     true,
	}
```

api/v1alpha1/regexp_test.go
```diff
@@ -0,0 +1,347 @@
...
+		})
+	}
+}
+
+func Test_Pattern_HelmRepoURL(t *testing.T) {
+	testCases := map[string]bool{}
```
mogsie, line 102: I couldn't find any PR that mentions any test cases for helm chart URLs.  They could be added here or later.  The test _still has value_ because it ensures that all uses of HelmRepoURL are in sync.

api/v1alpha1/regexp_test.go
krancour: I'm also not in favor of anything that feels too magical, but was thinking something like the following is not. It's maximally DRY and `ValidateExpression()` can be used for testing regexes elsewhere in the code that have nothing to do with CRDs.

```
func Test_Pattern_Branch(t *testing.T) {
	ValidateResourceExpression(
		t, 
		"Branch",
		map[string]bool{
			"":             false,
			"foo/bar":      true,
			"foo.bar":      true,
			"release-0.58": true,
			"/foo":         false,
			"foo/":         false,
			"foo//bar":     true,
			".foo":         false,
			"foo.":         false,
			"foo..bar":     true,
		},
	)
}

func ValidateResourceExpression(t *testing.T, marker string, testCases map[string]bool) {
	expression, err := findExpression(marker)
	require.NoError(t, err)
	ValidateExpression(t, expression, testCases)
}

func ValidateExpression(t *testing.T, *regexp.Regexp, map[string]bool) {
	for tt, expected := range testCases {
		t.Run(tt, func(t *testing.T) {
			require.Equal(t, expected, expression.MatchString(tt))
		})
	}
}
```

mogsie: Thanks!  Is there a package I could place the ValidateRegularExpression func?  I guess there are various regex'es around the codebase that could use this.  It could happen in another PR of course.

krancour: The tricky thing about where to put this is that Kargo is actually composed of multiple modules and things are deliberately set up so that the api module is the lowest level with the fewest dependencies. If we put this anywhere but there, best case, we'll probably drag a bunch of transitive dependencies into that module, and worst case, we'll create circular dependencies...

So while it's not completely ideal, I would say leave this in `api/v1alpha1`. Pretty much _everything_ throughout the entire rest of Kargo depends on that package anyway.

Maybe just leave a TODO comment that this isn't the ideal place for this and that longer-term we should look for a new home for it.

cc @hiddeco in case he has any ideas.

mogsie: I tried depending on it, but it seems difficult to depend on test files in other packages.  In addition, the vendoring caught me out.  I ended up duplicating the simple function to verify the `nonForwardRegex` from git/work_tree.go

krancour: You would need to define `ValidateResourceExpression()` in a file not ending with `_test.go`. For lack of a better option, this is common for test-related things that are used across multiple packages.

api/v1alpha1/regexp.go
```diff
@@ -0,0 +1,20 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// ValidateRegularExpression ensures that the provided regular expression
+// matches the expected results for the given test cases. Each key in the
+// testCases is tested against the expression, and the result must match the
+// value in the map of testcases.
+func ValidateRegularExpression(t *testing.T, expression *regexp.Regexp, testCases map[string]bool) {
```
hiddeco, line 14: I get how the constraints are cornering us into doing this, but I would at least like to see this in `v1alpha1/testing` to clearly indicate it is not part of the regular API.

mogsie: I totally agree.  "testing" is unfortunate though, since it would conflict with the testing package that is imported in all tests.  I'll called it "testhelper"

hiddeco: `testutils` would be more idiomatic I think, `testing` also really wouldn't be a problem as you can just alias it as `apitesting` during import in the couple of places where we use it.

api/v1alpha1/regexp_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+// ValidateResourceExpression checks the regular expression identified by marker
+// against the provided test cases.
+func ValidateResourceExpression(t *testing.T, marker string, testCases map[string]bool) {
+	expression, err := findExpression(marker)
+	require.NoError(t, err)
+	testhelper.ValidateRegularExpression(t, expression, testCases)
+}
+
+func Test_Pattern_Branch(t *testing.T) {
```
krancour, line 26: Nit: I'm usually reluctant to use snake case in test function names. I'll do it, but only when I feel _forced_ to because, for instance, multiple types in the same package have functions with the same name and `Test_Foo_Hello()` and `Test_Bar_Hello()` end up being _much_ more clear than `TestFooHello()` and `TestBarHello()`, which would easily fool anyone into thinking we're testing functions named `FooHello()` and `BarHello()`.

This is a preference and not even close to a hill I would die on, but I'd like to hear from @hiddeco about how strongly (or not) he feels this should be `TestBranchPattern()`.

Obviously applies to other functions in this same file.

krancour:
Out of an (over-)abundance of risk aversion, I am moving this to v1.6.0 since some regexes did change slightly.

api/v1alpha1/regexp_test.go
hiddeco: I only use snake cases for method testing (e.g. `TestFoo_Bar` for method `Bar` from `Foo`), or in case the function I am testing is private (e.g. `Test_foo` for function `foo()`).

krancour: These are easy enough to change because nothing calls them. Fix incoming and then this LGTM.

krancour (approved):
Yeoman's work here @mogsie. Can't thank you enough!

krancour:
Whoops. Missed the package rename.

krancour:
There we go.

krancour (approved):

api/v1alpha1/regexp_test.go
mogsie: That was just me not quite understanding the use of underscores when I started this little endeavour. Thanks.

Issue #3877, opened 2025-04-16
Support JSON Array Responses in HTTP Step


RonakShah92:
Title: Support JSON Array Responses in HTTP Step

# Checklist
* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature
Add support for handling JSON array responses in the HTTP step. Currently, the HTTP step expects all JSON responses to be objects (map[string]any), but many REST APIs return arrays of objects.

Example error when receiving an array response:
```step execution failed: step 0 met error threshold of 1: failed to run step "http": error building expression context from HTTP response: json: cannot unmarshal array into Go value of type map[string]interface {```

# Motivation
Many REST APIs return arrays by default, especially when fetching lists of resources. Some examples:
- GitLab API's merge request endpoints (`/api/v4/projects/{id}/merge_requests`)
- GitHub API's PR endpoints
- Most REST APIs that return collections

This limitation prevents users from using common REST API endpoints directly in their promotion workflows.

# Suggested Implementation
1. Modify the `buildExprEnv` function in `http_requester.go` to handle both object and array responses:

```go
if len(bodyBytes) > 0 && contentType == contentTypeJSON {
    // Try unmarshaling as array first
    var bodyArray []any
    if err = json.Unmarshal(bodyBytes, &bodyArray); err == nil {
        env["response"].(map[string]any)["body"] = bodyArray
    } else {
        // Fall back to object unmarshal
        body := map[string]any{}
        if err = json.Unmarshal(bodyBytes, &body); err != nil {
            return nil, err
        }
        env["response"].(map[string]any)["body"] = body
    }
}
```

2. Update the expression syntax to support array access:
```yaml
outputs:
  - name: firstMrTitle
    fromExpression: "response.body[0].title"  # Access first item in array
  - name: allTitles
    fromExpression: "response.body[*].title"  # Get all titles
```

This would maintain backward compatibility while adding support for array responses.

krancour:
Thanks for raising this. Agree. This should be pretty easily done, although I don't believe any updates to how expressions are evaluated is required. In fact, we don't even define the syntax. It's all expr-lang.

petersbattaglia:
+1 - expressing interest in this feature as well.

krancour:
Maintainers' priorities are largely driven by customers. +1'ing adds needless clutter tho these threads and doesn't change our prioritization.

[...]

PR #4059, opened 2025-05-11
Handle array json response in buildExprEnv

LeonLow97:
# Changes

- In `buildExprEnv`, added `json.Unmarshal` into type `any`, essentially an `interface{}`. Then, I added a type switch clause in Go to attempt to cast the interface type to either `map[string]any` or `[]any`. I used a switch statement in case we want to add different types in the future.
- Added error handling if the interface cannot be cast to any of the mentioned types, i.e., `map[string]any` or `[]any`
- I added unit tests, and all have been passed.
- Updated documentation to demonstrate handling of JSON response of type array


LeonLow97:
Resolves #3877

LeonLow97:
It seems that the unit test did not meet the code coverage requirements, I will be adding error checks to resolve this issue.

internal/promotion/runner/builtin/http_requester.go
```diff
@@ -227,12 +227,20 @@ func (h *httpRequester) buildExprEnv(
...
+		if err := json.Unmarshal(bodyBytes, &parsedBody); err != nil {
+			return nil, fmt.Errorf("failed to parse response: %w", err)
+		}
+
+		// Unmarshal into map[string]any or []any
+		switch parsedBody.(type) {
+		case map[string]any, []any:
+			env["response"].(map[string]any)["body"] = parsedBody
+		default:
+			return nil, fmt.Errorf("unexpected type when unmarshaling response: %w", err)
```
hiddeco, line 240: Do not think `err` actually holds something at this point, and I think you meant to do something like:

```suggestion
			return nil, fmt.Errorf("unexpected type when unmarshaling response: %T", parsedBody)
```

LeonLow97: You are right! I made a mistake there and have resolved it in my latest commit 👍

PR #4079, opened 2025-05-13
feat(ui): soak time

### LINKED ISSUE #3438: Visualize Soak Time in Pipeline
# Proposed Feature

The recently introduced soak time feature doesn't yet visualize in the pipeline. 

Since the time remaining requires two pieces of information: the stage and freight, this will likely need to be context aware. e.g. The user would first select the promote option on a stage, then freight timeline might show a timer on something that is not yet eligible due to the delay.

However, I'm open to other ways to show this.

# Motivation

UI parity on feature.

Marvin9:
fixes #3438 

Screenshot 2025-05-13 at 5 01 24 PM

- [x] Promote to downstream calculation


krancour (approved):

krancour:
This is great!

ui/src/features/project/pipelines/freight/duration-to-seconds.ts
```diff
@@ -0,0 +1,24 @@
...
+    years * 365.25 * 24 * 60 * 60 +
+    months * 30.44 * 24 * 60 * 60 +
+    weeks * 7 * 24 * 60 * 60 +
+    days * 24 * 60 * 60 +
+    hours * 60 * 60 +
+    minutes * 60 +
+    seconds;
+
+  return Math.floor(totalSeconds);
+};
```
rpelczar, line 24: We don't need this. Please use https://date-fns.org/v4.1.0/docs/differenceInMilliseconds

ui/src/features/project/pipelines/freight/use-soak-time-counter.ts
```diff
@@ -0,0 +1,69 @@
...
+          }
+        }
+
+        if ((newCounter?.minutes || 0) < 0) {
+          newCounter.minutes = 59;
+
+          if (newCounter.hours !== undefined) {
+            newCounter.hours -= 1;
+          }
+        }
```
rpelczar, line 37: We don't need to update every digit here, just recalculate time left using https://date-fns.org/v4.1.0/docs/differenceInMilliseconds

rpelczar (changes requested):

ui/src/features/project/pipelines/freight/duration-to-seconds.ts
Marvin9: Sorry I didn't get that. I don't need difference, I need seconds and not necessarily to find difference, besides I don't have exact Date because its duration, so I can't convert it without reference point either

ui/src/features/project/pipelines/freight/use-soak-time-counter.ts
Marvin9: I don't understand how that would help here. I think with that, at some point we still need to do some transformations to have formatted date

krancour:
@rpelczar and I spoke offline and he indicated his feedback can be addressed in a follow-up.

ui/src/features/project/pipelines/freight/use-soak-time-counter.ts
rpelczar: We want to include this PR in 1.5, so let's discuss it later.

PR #4087, opened 2025-05-13
refactor: limit expressions secret() func to secrets labeled as creds

krancour:
See conversation [here](https://github.com/akuity/kargo/pull/4062#discussion_r2087679481).

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
+++ b/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
@@ -285,7 +285,10 @@ config:
 
 The `secret()` function returns the `Data` field of a Kubernetes `Secret` with
 the specified name from the `Project`'s namespace decoded into a
-`map[string]string`. If no such `Secret` exists, an empty map is returned.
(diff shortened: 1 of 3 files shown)


hiddeco (commented):
I actually liked the fact that you could reference any Secret in a PromotionTemplate, as I think the Secrets managed by Kargo are different in the sense that they always provide access to some subscription. Whereas with Promotions, I can see there being a desire to have arbitrary "secret" things that are not necessarily "credentials" or a desire to reference a Secret managed by something else.

On the other hand, I can also see the validity of wanting to "allow" Kargo to access a Secret. But I am not sure labeling them as a "generic credential" is then the most straight forward choice in terms of wording.

krancour:
@hiddeco that label is not new though. It also wasn't my first choice for wording. For better or worse, the API server won't touch Secrets that lack that label, so the situation is that only secrets with that label can be created or viewed through the UI. The awkward bit this PR is trying to resolve is that it's strange for this function to have access to Secrets that are essentially invisible to the rest of Kargo and cannot be maintained via Kari's UI.

hiddeco (approved):

PR #4100, opened 2025-05-15
fix(controller): make yaml-update quote numeric-looking strings as necessary

### LINKED ISSUE #4099: yaml-update step isn't correctly quoting number like fields
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate bug report.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `kargo version`.

# Description
When passing through a number that is quoted it doesn't end up quoted in the yaml output

# Steps to Reproduce
Create a pipeline with a step similar to:
```yaml
steps:
    - uses: yaml-update
      as: update-image
      vars:
      - name: testValue
        value: main-123456
      config:
        path: values.yaml
        updates:
        - key: version
          value: ${{ vars.TestValue | split("-") | last() | quote() }}
      - uses: helm-template
        config:
          path: ./src/helm
          releaseName: test
          outPath: ./out/manifests.yaml
          valuesFiles:
          - values.yaml
```
If you have a helm chart using the version key it will be expressed in scientific notation indicating the value is unquoted when it is put in the yaml file, even though it is quoted in the value.

# Version
Kargo version 1.4.2

# Logs

```
Paste any relevant application logs here.
```

krancour:
Fixes #4099

#4099 was probably introduced when a different bug was closed by #3794.

This fix does the following:

- Adds a utility function for quoting strings intended for use as values in YAML files as necessary.

  i.e. If you have a string that looks like a number, you get back the original string enclosed in quotes. Anything else, including non-string values are returned as-is.

- Make the `internal/yaml` package able to set non-string values.

  It only did strings before. If it was a string that looked like a number, you got it without any extra quotes that would hint to a YAML parser that the value should be treated as a string. Because it _only_ worked with strings there wasn't any way to _know_ whether the value should be treated as a string or other type. It _used_ to rely on extra quotes around the numeric looking string value, but #3794 closed the door on that.

  By supporting other types we can know what the actual intent is. We can use the new utility (previous bullet) to quote numeric-looking strings only and all other values are used as-is.

- Update the JSON schema for the yaml-update promotion step's config to not require values to be a string and run codegen.

- At this point, the bug was fixed, however I realized the commit message fragments composed by the yaml-update step that are added to the step's output unconditionally quoted all values in its summary of changes. This created a slight impedance between the summary and what the step actually did. This was also fixed by using the new utility (first bullet).

### DIFF
--- a/internal/yaml/quote.go
+++ b/internal/yaml/quote.go
@@ -29,8 +29,7 @@ func QuoteIfNecessary(val any) any {
 	}
 	// If valStr is valid JSON, return its unmarshaled value. This could be an
 	// object, array, or null.
-	var parsedOther any
-	if err := json.Unmarshal([]byte(valStr), &parsedOther); err == nil {
+	if json.Valid([]byte(valStr)) {
 		return fmt.Sprintf("%q", valStr)
 	}
 	// If we get to here, just return the string.
--- a/internal/yaml/quote.go
+++ b/internal/yaml/quote.go
@@ -1,0 +1,38 @@
+package yaml
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+)
+
+// QuoteIfNecessary takes a value and returns it as-is if it is not a string. If
+// it is a string, it ascertains whether a YAML parser would interpret it as
+// another type. If so, it returns the string with additional quotes around it.
+// If not, it returns the string as-is.
+func QuoteIfNecessary(val any) any {
+	valStr, ok := val.(string)
+	if !ok {
+		return val
+	}
+	// If valStr is parseable as a float64, return that. float64 is used because
+	// it can represent all JSON numbers.
+	//
+	// NB: This is attempted prior to attempting to parse valStr as a boolean so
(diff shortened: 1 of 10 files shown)


internal/yaml/quote.go
```diff
@@ -0,0 +1,38 @@
...
+		return fmt.Sprintf("%q", valStr)
+	}
+	// If valStr is parseable as a bool return that.
+	if _, err := strconv.ParseBool(valStr); err == nil {
+		return fmt.Sprintf("%q", valStr)
+	}
+	// If valStr is valid JSON, return its unmarshaled value. This could be an
+	// object, array, or null.
+	var parsedOther any
+	if err := json.Unmarshal([]byte(valStr), &parsedOther); err == nil {
```
hiddeco, line 33: Can't we just use `json.Valid` here?

krancour: Forgot that existed. Will fix.

hiddeco (approved):

PR #4127, opened 2025-05-17
feat: roadmap info update

fykaa:
The roadmap hasn’t been updated in a while. This PR updates it to reflect the latest released versions (1.3.0, 1.4.0, 1.5.0) and marks 1.6.0 as the upcoming release with a placeholder for new features

### DIFF
--- a/docs/docs/100-roadmap.md
+++ b/docs/docs/100-roadmap.md
@@ -17,28 +17,40 @@ Project](https://github.com/akuity/kargo/milestones)
 
 ## In Progress
 
-### v1.3.0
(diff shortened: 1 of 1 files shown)


krancour:
I know this has gone without an update for some time, but I'd like to run point on this.

The description of what was accomplished in 1.4.0 is incorrect. It is what had been planned far in advance, but we pivoted and did _not_ do those things.

The description of 1.5 is _too_ detailed. These aren't release notes. The intent is a very, _very_ high level overview of where we've been and where we're headed.

I will try my best to square this away today.

PR #4168, opened 2025-05-20
fix(controller): make stage spec.vars referenceable from verifications

### LINKED ISSUE #4124: Stage spec.vars not referenceable  from verifications
# Description

In v1.5 we added Stage spec level vars, so that verifications could reference them. However, when I try to reference it and pass it to an AnalysisTemplate, I get error:

`error building AnalysisRun for Stage "vars" and Freight collection "3cf57b82db812e4abedfdedf3393676225a6b2d0" in namespace "kargo-demo-18": build spec: build arguments: evaluate argument "pokemon": cannot fetch pokemon_1 from <nil> (1:7) | vars.pokemon_1 | ......^`

# Screenshots

Image

# Steps to Reproduce

Run example here:
https://github.com/akuity/kargo-examples/pull/3/files

```yaml
kind: Stage
metadata:
  name: vars
  namespace: kargo-demo-18
spec:
  vars:
  - name: pokemon_1
    value: pikachu
  requestedFreight:
  - origin:
      kind: Warehouse
      name: vars
    sources:
      direct: true
  promotionTemplate:
    spec:
      vars:
      - name: pokemon_2
        value: charmander
      steps:
      - task:
          name: promo-process
        vars:
        - name: pokemon_3
          value: bulbasaur
  verification:
    analysisTemplates:
    - name: pokemon-query
    args:
    - name: pokemon
      value: ${{ vars.pokemon_1 }}
```

# Version

```
v1.5.0
```

krancour:
Fixes #4124

### DIFF
--- a/internal/controller/stages/regular_stages.go
+++ b/internal/controller/stages/regular_stages.go
@@ -1341,6 +1341,7 @@ func (r *RegularStageReconciler) startVerification(
 				),
 				exprfn.DataOperations(ctx, r.client, stage.Namespace)...,
 			),
+			Vars: stage.Spec.Vars,
 		},
 	}
 	for _, freightRef := range freight.Freight {
--- a/internal/rollouts/analysis_run_builder_test.go
+++ b/internal/rollouts/analysis_run_builder_test.go
@@ -637,7 +637,7 @@ func TestAnalysisRunBuilder_buildArgs(t *testing.T) {
 				{Name: "param1", Value: "${{ vars.test }}"},
 			},
 			exprCfg: &ArgumentEvaluationConfig{
-				Vars: []ArgumentVariable{
+				Vars: []kargoapi.ExpressionVariable{
 					{Name: "test", Value: "value1"},
 					// Ensure overriding and expression evaluation works.
 					{Name: "test", Value: "${{ vars.test + '-suffix'}}"},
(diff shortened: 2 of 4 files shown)


internal/controller/stages/regular_stages.go
```diff
@@ -1342,6 +1342,7 @@ func (r *RegularStageReconciler) startVerification(
 				exprfn.DataOperations(ctx, r.client, stage.Namespace)...,
 			),
 		},
+		rollouts.WithVars(stage.Spec.Vars),
```
hiddeco, line 1345: There is no need to create a `WithVars`, you can just set `Vars` in `rollouts.WithArgumentEvaluationConfig`.

hiddeco: When doing this, I think the type of `Vars` should change to `kargoapi.ExpressionVariable`, and the `ArgumentVariable` can be dropped then. That's what was supposed to happen in my original PR.

krancour: 🤦‍♂️

krancour: There _is_ a conversion from `[]kargoapi.ExpressionVariable` to `[]rollouts.ArgumentVariable` that needs to happen item-by-item. Does that influence your thinking at all?

hiddeco: See my second comment :-).

krancour: I think they crossed in the mail.

krancour: Should be all set now.

hiddeco (approved):

Issue #4072, opened 2025-05-13
panic in the promotions!


krancour:
A few times now, we've seen cases where a panic while executing some promotion step has not only caused the entire promotion to fail, but it has also prevented proper bookkeeping in terms of recording the outcome of those promotion steps that _did_ run successfully. This creates the illusion that the panic has occurred at step 0.

I thought we'd improved this recently with modification to the step execution engine to always update promotion status _between_ steps, but I saw this problem present itself again today and that change seems not to have paid off.

More investigation is required, but since this problem has now obstructed the troubleshooting of other problems multiple times now, we should prioritize this.

alexmt:
Noticed this in our controller:

[...]

PR #4169, opened 2025-05-20
fix(runner): step panic recovery within engine

hiddeco:
Branched off from another local branch that was getting out of hand with other engine-related refactoring.

This should address (at least a part of) #4072, and it works because we catch the panic within the engine. By doing this, we can still construct a `Result` to return to the reconciler, which will then propagate this into the actual status of the Promotion.

This detachment (or separation of concerns) is the primary reason why [the `recover()` within the reconciler itself](https://github.com/akuity/kargo/blob/e53abcca8a37c4d623400391c69359d154f8725e/internal/controller/promotions/promotions.go#L308-L318) did not have an effect. Arguably, this also means it is a potential candidate for removal, as I doubt it will be of much use after this has been merged.

### DIFF
--- a/internal/controller/promotions/promotions_test.go
+++ b/internal/controller/promotions/promotions_test.go
@@ -569,14 +569,14 @@ func Test_parseCreateActorAnnotation(t *testing.T) {
 func Test_calculateRequeueInterval(t *testing.T) {
 	testStepKindWithoutTimeout := "fake-step-without-timeout"
 	promotion.RegisterStepRunner(
-		&pkgPromotion.MockStepRunner{Nm: testStepKindWithoutTimeout},
+		&pkgPromotion.MockStepRunner{StepName: testStepKindWithoutTimeout},
 	)
 
 	testStepKindWithTimeout := "fake-step-with-timeout"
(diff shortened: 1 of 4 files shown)


internal/controller/promotions/promotions_test.go
```diff
@@ -569,14 +569,14 @@ func Test_parseCreateActorAnnotation(t *testing.T) {
 func Test_calculateRequeueInterval(t *testing.T) {
 	testStepKindWithoutTimeout := "fake-step-without-timeout"
 	promotion.RegisterStepRunner(
-		&pkgPromotion.MockStepRunner{Nm: testStepKindWithoutTimeout},
+		&pkgPromotion.MockStepRunner{StepName: testStepKindWithoutTimeout},
```
krancour, line 572: That's nicer than `Nm` I was tired when I did that and grumpy that `Name` conflicted with `Name()`. Thank you.

krancour (commented):
🔥 I think I would close #4072 with this and we can re-open it later if we notice panics coming from elsewhere.

krancour (approved):

Issue #4089, opened 2025-05-14
Remove older/redundant promotion view


jessesuen:
# Description

We have a newer promotion view in v1.5, but we still show the older one when clicking on a promo after a Stage.

Also, the new view lacks the button to Abort a promotion, and needs to be added in that view.

# Screenshots

https://github.com/user-attachments/assets/ee952066-bf51-4dde-89dc-d62d5c4fd206

# Version

```
v1.5.0-rc.3
```

PR #4175, opened 2025-05-21
chore(ui): replace old promotion modal and add abort promotion in new one

Marvin9:
fixes https://github.com/akuity/kargo/issues/4089

@rpelczar this (drawer over drawer) or redirect to existing url (in which case we will loose stage drawer)

Screenshot 2025-05-21 at 2 24 01 PM


krancour:
If we release v1.5.1 tomorrow, do we think this will make it in?

Marvin9:
> @Marvin9 I think it's better to redirect user to a separate route with a new drawer. In this case user won't lose the stage drawer, because we have the browser's back button.

But that will refetch data (promotion list) so users have to wait

Marvin9:
> If we release v1.5.1 tomorrow, do we think this will make it in?

Mostly yes

rpelczar (approved):

PR #4181, opened 2025-05-21
refactor(runner): promotion engine improvements

hiddeco:
This refines the "simple" Promotion engine to address technical debt and improve maintainability. As the features in the engine have grown, several architectural weaknesses have emerged — particularly around state management, component coupling, and code structure. This doesn't completely redesign the system, but makes small improvements to reduce complexity and support future development.

- The `promotion.Context` now acts as the single source of truth for state, and methods no longer accept separate state parameters. Previously, I had avoided this and imagined the context to be read-only. However, with the introduction of `StepExecutionMetadata` we started making direct modifications to `promotion.Context` anyway, and maintaining separate state parameters created an inconsistent pattern that was both confusing and error-prone. 
- Instead of determining if output should be added under the "namespace" of a task by being aware of specific step name (e.g. `compose-output`), the engine now looks if the step implements a `TaskLevelOutputStepRunner` interface. This provides better (albeit — not great) separation of concerns and removes step specific knowledge from the engine itself.
- The lengthy `executeSteps` method has been broken down into smaller methods to improve readability. While I still believe there is much room for improvement, this is a small incremental step to (hopefully) something better in the future.

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -204,7 +204,7 @@ func RunningPromotionsByArgoCDApplications(
 
 			// As step-level variables are allowed to reference to output, we
 			// need to provide the state.
-			vars, err := dirStep.GetVars(ctx, cl, nil, promoCtx, promoCtx.State)
+			vars, err := dirStep.GetVars(ctx, cl, nil, promoCtx)
 			if err != nil {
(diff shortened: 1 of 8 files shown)


krancour (approved):
This is beautiful.

PR #4184, opened 2025-05-21
feat: cache `secret()` and `configmap()` results

hiddeco:


krancour:
I agree we could benefit from some caching here, but...

To execute one step, we:

* Check if it should be skipped:
  * This requires that we first evaluate the vars, which involves:
    * One call to `expressions.EvaluateTemplate()` for _each_ process-level var.
    * Another for _each_ step-level var.
  * A call to `expressions.EvaluateTemplate()` to evaluate the `if` expression.
* Prepare the step context:
  * This requires that we, again, evaluate the vars, which again involves:
    * One call to `expressions.EvaluateTemplate()` for _each_ process-level var.
    * Another for _each_ step-level var.
  * Evaluate the step's config.
* Finally, actually execute the step.

_If I am reading the changes correctly_ (leaving room for self-doubt here), each cache lives just for the duration of a single execution of `expressions.EvaluateTemplate()` or `expressions.EvaluateJSONTemplate()`... and that would mean there are _many different caches_ involved in all of the above.

If `secret()` or `configmap()` were invoked with the same arguments multiple times during the evaluation of the step's config, the values returned would be consistent, which is for sure a small win, _however_, those results aren't guaranteed to agree with all the _other_ invocations of those same functions (with the same arguments) that may have occurred in all the of the previous bits of evaluation (vars, `if` expression, etc.) Weirder still, the vars the step's config are evaluated against were, themselves, all evaluated using different caches.

What I'm building toward is that I think even with this change, inconsistent results within the end-to-end execution of a step are possible. I think this could be solved with a small tweak whereby instead of saying `true` "I want to use a cache" when evaluating expressions, we (optionally) _provide_ the cache. If the cache were created at the start of the step execution loop (inside of the loop) we would get consistent results throughout all the many expression evaluations described above.

Tangent:

In describing all of the above, I'm noting two obvious inefficiencies:

* We could probably evaluate process-level vars just once before the engine's main loop starts executing steps.

* We probably don't need to evaluate step level vars before evaluating a step's `if` expression. I'd not actually expect that expression to have access to those.

Fixing these two things would significantly reduce the total number of expression evaluations performed.

krancour (approved):
🔥 Great change.

PR #4185, opened 2025-05-22
chore: docusaurus version 3.7.0

fykaa:
Dependabot has raised vulnerability alerts related to the current Docusaurus version, so we're upgrading it to address those issues.


krancour:
This broke the docs build, but resolving that is low priority.

krancour:
This will need a rebase, although this remains low priority.

hiddeco (approved):

Issue #4113, opened 2025-05-15
Warehouse exclusion filter by git author


jessesuen:
# Proposed Feature

Similar to how we can exclude paths, we should additionally have a include/exclude filter where Freight may or may not be generated, based on attributes of the git commit, such as git author.

To make this flexible, perhaps we can use expr in some way so that the filter can be more generic.

# Motivation

This would provide another way to have Kargo avoid acting on automated commits from systems, including Kargo itself.

I'm working on a kustomize based rendered branch pattern, and found that I could not exclude the `kustomization.yaml` file from my git Warehouse, because that file might be edited by humans or by machines. And I want to exclude commits made by machines (i.e. kargo).

PR #4193, opened 2025-05-22
feat: support filtering Git commits and tags using expressions

hiddeco:
Fixes: #4113


docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -344,6 +352,114 @@ strategies are:
...
+:::tip
+You can test your expressions using the
+[expr-lang playground](https://expr-lang.org/playground).
+
+The playground allows you to evaluate expressions against sample data and
+see the results in real-time. This is especially useful for debugging and
+validating your expressions before applying them to your `Warehouse` resources.
+:::
+
+##### Tag Expression Filtering
```
jessesuen, line 381: Can you explain why a `tagExpressionFilter` is necessary in addition to `commitExpressionFilter`? I thought everything is a commit at the end of the day and we would just need `commitExpressionFilter`. I noticed the fields that are available in both of them (`tag`, `commitID`, `author`, etc...) are identical. Would they ever be different?

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -344,6 +348,165 @@ strategies are:
...
+:::
+
+:::warning
+Invalid expressions will cause the subscription to fail. Always test your
+expressions to ensure they evaluate correctly with your repository's data.
+:::
+
+:::tip
+You can test your expressions using the
+[expr-lang playground](https://expr-lang.org/playground).
```
krancour, line 370: I somehow didn't know this existed. This is useful.

internal/controller/git/work_tree.go
```diff
@@ -460,6 +460,11 @@ type TagMetadata struct {
 	// Subject is the subject (first line) of the commit message associated
 	// with the tag.
 	Subject string
+	// Tagger is the person who created the tag, in the format "Name <email>".
+	// This field is only populated for annotated tags.
+	Tagger string
+	// Annotation is the annotation of the tag, if it is an annotated tag.
+	Annotation string
```
krancour, line 467: Not a new problem, but I notice that this struct is missing a field for the `commitDate` of the commit pointed to by the tag.

Would now be a good time to rectify that?

krancour: Other than this, LGTM.

hiddeco: I am not sure what is best. The way Git deals with it, is that it has the concept of a "creator date" which will either point to the date of the commit or the date of the annotated tag. I.e., it gives you a single field that is the most detailed.

If we were to move away from this, I would think we would deprecate the `CreatorDate` (including in the `DiscoveredCommit` API object) in favor of a `CommitterDate` and `TaggerDate` (following Git idioms) or `CommitDate` and `TagDate` (following slightly more natural language).

krancour: Ah... I'd not dug deep enough into this to realize that it wasn't simply something that was available and we forgot to read it.

I would say we can reconsider later if/how this can be improved.

I don't want that to hold this up.

LGTM

krancour (approved):

internal/controller/git/work_tree.go
krancour: Err... after the lint issue is fixed.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
hiddeco: I will create a separate PR to also add this to the expressions reference page.

Issue #4173, opened 2025-05-21
UI: timeline does not auto-refresh to show new freight


fykaa:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

The timeline view does not refresh automatically when a new `Freight` is created. Users have to manually refresh the page to see the latest timeline updates. This affects usability and may confuse users expecting live updates.

Subsequently, we may also need to update the Quickstart to remove the note after this is addressed. The note makes the UI seems less stable than it actually is.

supersedes https://github.com/akuity/kargo/issues/4106

# Steps to Reproduce

1. Trigger a new `Freight` (e.g., through promotion or pipeline run)
2. Wait and observe the timeline view
3. Notice that the new `Freight` does not appear
4. Manually refresh the page
5. New `Freight` entry becomes visible

# Version

```
v1.5.0
```

Marvin9:
> Trigger a new Freight (e.g., through promotion or pipeline run)

Can you clarify this step please?

Marvin9:
We don't have watch freights endpoint. @krancour I assume that it isn't straightforward (as watch promotion for example) because of grouping and other filters?

krancour:
I _think_ it could still be fairly straightforward. I'll look as soon as I have a chance.

I apologize because I don't think I'd realized we didn't already have that and that it was the underlying reason for this.

krancour:
@Marvin9 I can do this quickly. How would you want it to work? Do you just want to watch all Freight for a given Project? Or do you also want to (optionally?) specify a Warehouse (from your filters) and have that criteria applied server-side?

Marvin9:
Thanks @krancour only by the project will work for now

krancour:
@fykaa re-assigning the back end parts of this to you. This should be quick and easy as this can be modeled after existing watch endpoints, which are pretty simple.

PR #4199, opened 2025-05-22
feat: add watchFreights API endpoint

fykaa:
part of https://github.com/akuity/kargo/issues/4173

cc: @Marvin9


api/service/v1alpha1/service.proto
```diff
@@ -61,6 +61,7 @@ service KargoService {
   rpc ApproveFreight(ApproveFreightRequest) returns (ApproveFreightResponse);
   rpc DeleteFreight(DeleteFreightRequest) returns (DeleteFreightResponse);
   rpc GetFreight(GetFreightRequest) returns (GetFreightResponse);
+  rpc WatchFreights(WatchFreightsRequest) returns (stream WatchFreightsResponse);
```
krancour, line 64: This comes up from time to time... Freight is a [mass noun](https://en.wikipedia.org/wiki/Mass_noun).

Grammatically speaking, this should be `WatchFreight` (no "s").

```suggestion
  rpc WatchFreights(WatchFreightsRequest) returns (stream WatchFreightsResponse);
```

krancour: Implementation looks fine, so LGTM after this is fixed.

krancour:
There are also linting errors.

api/service/v1alpha1/service.proto
krancour: Almost anywhere you see Freights in code or docs, it is accidental because we've tried to be grammatically correct all along. Wherever we encounter it, I'd like us to fix it, as long as it's not breaking. We don't need to hunt down all such occurrences, rather they can be fixed as we happen upon them while working on other things. There are a few places, like the CLI where "freights" is tolerated due to Postel's law, and there may be a place or two where some lower level package automatically pluralizes a word and doesn't get this one correct. In most cases, we have the ability to be grammatically correct.

krancour (approved):

krancour:
Thank you!

PR #4203, opened 2025-05-22
docs: switch to Algolia for search

hiddeco:
This moves from the quite awkward "local search" to the service from Algolia.

It may need some tweaking in the future, but  I count this as a big step forward in terms of actually being able to find things again.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -50,12 +50,6 @@ const config = {
         anonymizeIP: true,
       },
     ],
-    [
-      require.resolve("@cmfcmf/docusaurus-search-local"),
-      {
-        indexBlog: false,
-      },
-    ],
     'docusaurus-plugin-sass'
   ],
 
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -114,6 +114,12 @@ const config = {
         defaultMode: 'light',
       },
       metadata: [{name: 'akuity, argoproj, argo cd, argo workflows, argo events, argo rollouts, kubernetes, gitops, devops', content: 'akuity, documentation, developer documentation'}],
+      algolia: {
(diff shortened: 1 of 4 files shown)


docs/docusaurus.config.js
```diff
@@ -114,6 +108,12 @@ const config = {
         defaultMode: 'light',
       },
       metadata: [{name: 'akuity, argoproj, argo cd, argo workflows, argo events, argo rollouts, kubernetes, gitops, devops', content: 'akuity, documentation, developer documentation'}],
+      algolia: {
+        appId: '3SQ7LK6WD9',
+        apiKey: '5627b8c2efd5b28a5b70c6660cb2b0f3',
```
krancour, line 113: Few questions here:

1. Is this safe to commit in a public repo? I'm assuming it may be. That's common with Google Analytics integration and I guess this is similar?

2. How does this work with the multiple versions of the docs site we host simultaneously (one for each minor revision of Kargo)? They can all use the same ID? Or no? Are they all using the same search index? Or are they each indexed separately?

hiddeco: 1. Yes, this API key is [safe to commit](https://support.algolia.com/hc/en-us/articles/18966776061329-Can-the-search-API-key-be-public).
2. The search results on the version subdomains will, at this moment, return results for the latest version. I have yet to figure out how to avoid this. But given this won't be a problem until we release 1.6.0, I would not like to make this a blocker given the value provided as soon as we merge this.

krancour: I agree it should not block this -- even if we didn't figure it out for a while. Doing a search in an older version of the docs and getting results from the newest version is a little weird, but it's not anything I'm going to lose sleep over.

krancour (approved):

PR #4207, opened 2025-05-22
fix(docs): center landing page and styling tweaks

rpelczar:


krancour:
Never occurred to me there was an invisible TOC there. This is a good change, but it has this side effect:

Screenshot 2025-05-22 at 5 35 18 PM

Any ideas about how to fix the symmetry here?

hiddeco:
@krancour working on this with Rafal offline, we want to:

- Maintain 2/2 rows for first section
- Keep second section single row of 3
- Realign the "hero", as it looks slightly weird now

hiddeco (approved):

hiddeco:
Thanks for the icing on the 🍰 @rpelczar 🙇

krancour:
Something may be wrong...

In when I look at the preview in Firefox, this still has three cards on the top row and in Chrome things look like this:

Screenshot 2025-05-22 at 5 56 24 PM

krancour:
> Something may be wrong...

Eh... now it seems my results vary quite a bit by viewport size.

hiddeco:
@krancour any chance it's your browser cache? I didn't see this in my Firefox.

krancour:
> any chance it's your browser cache?

I suspected that at first and did a hard refresh.

I'm a little bit wondering if we need to clear Netlify's build cache.

krancour:
This is after clearing the cache. I get the same behavior if I build the doc site locally.

https://github.com/user-attachments/assets/797857d5-6189-4474-bdd1-70aecf4afe75

rpelczar:
@krancour You mean the last item is not centered? This is intentional.
We tried making the first section 2/2 and the second 3 items/row, but we think it looks worse.

image

Personally, I don't think it's wrong that we align tiles to the left.

krancour:
@rpelczar no... the thing I thought looks wrong is that as the viewport grows horizontally, the top row is getting longer and the second row is getting shorter, instead of staying 2x2.

hiddeco:
@krancour that's kind of the price you pay for a responsive grid. I made some changes so that smaller screens will get a nice 2/2 view due to the padding, but this is the best we will get for now.

krancour:
fwiw, the weird issues I was experiencing went away after #4185.

PR #4214, opened 2025-05-23
chore(api)!: rename `CredentialTypeLabelGeneric` and document labels

hiddeco:
- Rename `CredentialTypeLabelGeneric` to `CredentialTypeLabelValueGeneric`
- Document all labels

### DIFF
--- a/api/v1alpha1/labels.go
+++ b/api/v1alpha1/labels.go
@@ -1,22 +1,53 @@
 package v1alpha1
 
 const (
+	// AliasLabelKey is used to identify the alias of a resource.
+	// For example, Freight may have an alias that is easier to remember than
+	// the computed name of the resource. For details about the behavior of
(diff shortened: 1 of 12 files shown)


fykaa:
Could you help me understand why we're renaming `CredentialTypeLabelGeneric` to `CredentialTypeLabelValueGeneric` Just trying to get the context here 🤔

the rest lgtm!

fykaa (approved):

krancour:
Is this breaking only because of anyone that happened to be using `CredentialTypeLabelGeneric` in their own Go code already?

I know api/v1alpha1 is accessible, but as we don't actually _publish_ that module, I don't feel strongly that we need to regard this as breaking... or at least I don't think this rises to the level of a mention in the v1.6.0 release notes.

Issue #4122, opened 2025-05-16
Can't apply existing Kargo Project configs to Kubernetes after upgrading to 1.5.0


ThreePinkApples:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

After upgrading to 1.5.0, our existing projects that and does not have a `ProjectConfig` fails to apply (sync through ArgoCD). We have our projects defined in yaml files in a git repository, synced through ArgoCD. We do not use the Kargo UI to do any configurations other than promotions.

```
error when patching "/dev/shm/4234176703": admission webhook "project.kargo.akuity.io" denied the request: Project.kargo.akuity.io "examplename" is invalid: spec: Forbidden: deprecated field: create a ProjectConfig named "examplename" with the config instead
```

# Steps to reproduce
Try to apply a Kargo Project yaml directly to kubernetes that uses the old `spec` directly on the Project and not in a ProjectConfig

# Version

```
{
  "Version": "v1.5.0",
  "BuildDate": "2025-05-15T22:29:42Z",
  "GitCommit": "61a1c24b7971bb19fcb430c737ca48e28912200c",
  "GitTreeDirty": false,
  "GoVersion": "go1.24.3",
  "Compiler": "gc",
  "Platform": "linux/amd64"
}
```

jessesuen:
In v1.5, Project configuration was moved from `spec` to a new separate namespace resource, `ProjectConfig` (See https://github.com/akuity/kargo/issues/3669). So you would gitops a resource, like:


```
---
apiVersion: kargo.akuity.io/v1alpha1
kind: Project
metadata:
   name: example
---
apiVersion: kargo.akuity.io/v1alpha1
kind: ProjectConfig
metadata:
  name: example
  namespace: example
spec:
  promotionPolicies:
  - stageSelector:
      name: test
    autoPromotionEnabled: true
  - stageSelector:
      name: uat
    autoPromotionEnabled: true
```

ThreePinkApples:
Yes, but I would have expected the old config to work until full removal in 1.7.0 so we can upgrade to 1.5 first, and then do the changes a little later when it is unlikely that we'd revert back to 1.4.4. But not that big of a deal, a search-replace and do it all in one commit makes reverting easy enough

krancour:
[...]

PR #4215, opened 2025-05-23
fix(mgmt-controller): do not erase Project spec after migration to ProjectConfig

hiddeco:
Resolves: #4122

This provides a friendlier automated migration path from a `Project` to a `ProjectConfig` by labeling the `Project` with `kargo.akuity.io/migrated: true` after migration, instead of erasing the spec.


api/v1alpha1/labels.go
```diff
@@ -16,6 +16,12 @@ const (
 	ShardLabelKey             = "kargo.akuity.io/shard"
 	StageLabelKey             = "kargo.akuity.io/stage"
 
+	// MigratedLabelKey is a label set on a resource that has successfully been
+	// migrated, indicating that no further actions are required. This label is
+	// set to "true" when the migration is complete. An example of this is the
+	// migration of the ProjectSpec to a ProjectConfig resource.
+	MigratedLabelKey = "kargo.akuity.io/migrated"
```
krancour, line 23: Would it be worth having the key be more specific? I'm thinking about the possibility of a resource that may undergo multiple different sorts of migrations over the course of many Kargo releases. If we use a key as generic as this, we'll have trouble identifying which migrations a resource is or is not in need of.

krancour:
To fully resolve #4122, don't we also need to temporarily resume allowing the deprecated project.spec field to be set?

api/v1alpha1/labels.go
krancour: How will we denote that a resource has undergone multiple different migrations over time?

If we wanted them comma separated or structured as JSON or something along those lines, we should probably introduce some helpers for setting/checking migration status.

internal/api/annotations.go
```diff
@@ -95,3 +95,38 @@ func AbortPromotionAnnotationValue(annotations map[string]string) (*kargoapi.Abo
 	}
 	return &req, ok
 }
+
+// HasMigrationAnnotationValue checks if the AnnotationKeyMigrated annotation
+// is present in the provided annotations map and if it contains the specified
+// migration type as a key with a value of true.
+func HasMigrationAnnotationValue(annotations map[string]string, migrationType string) bool {
```
krancour, line 102: I'm wondering it if might be more ergonomic for this to accept a `client.Object` instead of a `map[string]string`, so that the question reads more like "has this resource undergone this migration?" (And rename the func accordingly, of course.)

To complement, a similar change could be made to `AddMigrationAnnotationValue()`, with the benefit that the annotation map can be created if it doesn't already exist, which might save callers of that function some effort.

Issue #3818, opened 2025-04-08
refactor: dump all claims from identity token on user details page


krancour:
The user details page in the UI seems to be checking for the existence of specific claims in the user's identity token and displaying their values if they exist, but I believe it would be more informative to dump _all_ the claims from the user's identity token.

krancour:
Bumping to v1.6.0.

Marvin9:
we can have raw view to show all claim dumps and keep human readable as it is. What do you think?

krancour:
This is what it looks like now:

![Image](https://github.com/user-attachments/assets/5637d74c-6ac3-476e-9c10-5eba3c2eb750)

imo, any logic that might exist currently for special treatment of Issuer, Email, or Groups could just go away and the raw claims could be dumped into a table that otherwise looks just like the one we've got now.

Sound ok?

PR #4216, opened 2025-05-23
feat(ui): dump all claims from identity token on user details page

Marvin9:
fixes https://github.com/akuity/kargo/issues/3818

Screenshot 2025-05-23 at 3 16 11 PM

### DIFF
--- a/ui/src/features/auth/jwt-utils.ts
+++ b/ui/src/features/auth/jwt-utils.ts
@@ -38,3 +38,31 @@ export const getUserEmail = (user?: JWTInfo | null) => {
 
   return meta;
 };
+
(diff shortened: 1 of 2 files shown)


rpelczar:
@Marvin9 can we wrap the content with the Card component?

krancour:
@Marvin9 I ❤️ this. This is going to help so many people troubleshooting RBAC issues.

Only feedback I have is that users not very familiar with JWT / OAuth / OIDC may not intuitively understand exactly what this table is showing them.

Is there a chance we could add a title or label or something to make it clear that these are claims from the user's identity token?

krancour:
I'm going to move this back to v1.5.1. This is too valuable not to have available as soon as possible.

PR #4224, opened 2025-05-23
feat(ui): watch freights

Marvin9:
fixes https://github.com/akuity/kargo/issues/4173

### DIFF
--- a/ui/src/features/project/pipelines/use-watch-freights.ts
+++ b/ui/src/features/project/pipelines/use-watch-freights.ts
@@ -17,7 +17,7 @@ export const useWatchFreights = (project: string) => {
     const watchFreights = async () => {
       const promiseClient = createClient(KargoService, transportWithAuth);
 
-      const stream = promiseClient.watchFreights(
+      const stream = promiseClient.watchFreight(
         {
           project
         },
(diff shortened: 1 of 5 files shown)


ui/src/features/project/pipelines/use-watch-freights.ts
```diff
@@ -0,0 +1,80 @@
...
+            },
+            transport: transportWithAuth
+          });
+
+          client.setQueryData(queryFreightKey, currentFreights);
+        }
+      }
+    };
+
+    watchFreights();
```
fykaa, line 76: Can we rename this to `watchFreight` to match the backend method name? imo keeping this consistent across the stack will help with traceability and reduce confusion when jumping between the backend and frontend code

krancour: I agree this is preferable. Freight is a [mass noun](https://en.wikipedia.org/wiki/Mass_noun) so `freights` always sounds very awkward and we're trying to avoid it throughout the code (and docs) except in places where a (dumb) pluralization library or something forces it on us.

krancour (approved):

PR #4229, opened 2025-05-24
feat: Support disabling deletion of adopted namespace

Kavinraja-G:


docs/docs/50-user-guide/20-how-to-guides/20-working-with-projects.md
```diff
@@ -243,6 +243,15 @@ spec:
   # ...
 ```
 
+### Preventing Namespace Deletion
+
+By default, when a Project is deleted, Kargo will attempt to delete the corresponding Namespace.
+However, in cases where an existing Namespace is adopted by a Project, you may wish to prevent Kargo
```
krancour, line 249: It is probably useful not _only_ in cases of adoption.

internal/controller/management/projects/projects.go
```diff
@@ -480,23 +480,30 @@ func (r *reconciler) ensureNamespace(ctx context.Context, project *kargoapi.Proj
 		// uncommon scenario where an organization has its own controller that
 		// creates and initializes namespaces to ensure compliance with
 		// internal policies. Such a controller might already own the namespace.
```
krancour, line 482: Even prior to this PR, this comment was misplaced. Could you please relocate it to just above where the Project is (now conditionally) made an owner?

docs/docs/50-user-guide/20-how-to-guides/20-working-with-projects.md
Kavinraja-G: Yeah I did thought of that, but if the namespace is created by Kargo and user wants to keep in this case that will leave the namespace orphaned?

internal/controller/management/projects/projects.go
```diff
@@ -480,23 +480,30 @@ func (r *reconciler) ensureNamespace(ctx context.Context, project *kargoapi.Proj
...
+		if project.Labels[kargoapi.KeepNamespaceLabelKey] != kargoapi.LabelTrueValue &&
+			ns.Labels[kargoapi.KeepNamespaceLabelKey] != kargoapi.LabelTrueValue {
+
+			updated, err := r.ensureFinalizerFn(ctx, r.client, ns)
+			if err != nil {
+				return fmt.Errorf("error ensuring finalizer on namespace %q: %w", project.Name, err)
+			}
+			if updated {
+				logger.Debug("added finalizer to namespace")
+			}
```
krancour, line 495: I'm not sure this should be conditional. This is for ensuring things work in the _opposite_ direction. i.e. Ensures that if you delete the Project's namespace, the Project goes with it. I don't feel as if that is behavior that needs to change in order to prevent the namespace from being deleted when the Project is.

api/v1alpha1/labels.go
```diff
@@ -15,6 +15,7 @@ const (
 	ProjectLabelKey           = "kargo.akuity.io/project"
 	ShardLabelKey             = "kargo.akuity.io/shard"
 	StageLabelKey             = "kargo.akuity.io/stage"
+	KeepNamespaceLabelKey     = "kargo.akuity.io/keep-namespace"
```
krancour, line 18: This really should be an annotation.

Kavinraja-G: Fixed https://github.com/akuity/kargo/pull/4229/commits/332d481507c22e94f215a3107491ee86b7e409dd

docs/docs/50-user-guide/20-how-to-guides/20-working-with-projects.md
Kavinraja-G: But do we want that to happen? This will intentionally make the controller to orphan, maybe there can be scenarios where user wants to keep and delete the Project alone. I can include this scenario if we are okay to have.

internal/controller/management/projects/projects.go
```diff
@@ -476,27 +476,35 @@ func (r *reconciler) ensureNamespace(ctx context.Context, project *kargoapi.Proj
...
-			return fmt.Errorf(
-				"error patching namespace %q with project %q as owner: %w",
-				project.Name,
-				project.Name,
-				err,
-			)
+
+		// Before setting project as the owner for the existing namespace check for `keep-namespace`
+		// label in either project/namespace if it's true we don't want to make project as the owner
+		if project.Annotations[kargoapi.AnnotationKeyKeepNamespace] != kargoapi.AnnotationTrueValue &&
```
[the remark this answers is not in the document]
Kavinraja-G: Thanks for the brief explanation!

> Namespace has no remaining relationship to the Project
In that case, is that ideal to have the `keep-namespace` annotation only on Project and not on the namespace?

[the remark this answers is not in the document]
Kavinraja-G: Done! lmk, if you have any suggestions. Thanks again, for your inputs.

docs/docs/50-user-guide/20-how-to-guides/20-working-with-projects.md
krancour: Oh... there's no orphan... in your latest commits, you already included the logic to remove the ownership reference.

internal/controller/management/projects/projects.go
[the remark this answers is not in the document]
krancour: Nicely done. At a glance, this looks almost entirely correct.

For good measure, I would not overlook this part:

> Ownership should be unconditional (mainly for bookkeeping purposes), _but can be configured to prevent cascading deletion of dependents._

In `ensureNamespace()`, I believe this requires you to set `ownerRef.Controller = nil`.

You will also need to make sure existing ownership references are updated.

Look for `logger.Debug("namespace exists and is already owned by this Project")`. You see we return early right after that, but you'll need to check if `ownerRef.Controller != nil` and if that's the case, you'll need to fix it.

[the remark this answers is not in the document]
Kavinraja-G: Ahh, I missed this part. Thanks! Fixed in https://github.com/akuity/kargo/pull/4229/commits/7afe3179bb8146ad82b030dc634eaf4d16480e23

internal/controller/management/projects/projects.go
```diff
@@ -467,6 +532,9 @@ func (r *reconciler) ensureNamespace(ctx context.Context, project *kargoapi.Proj
 		for _, ownerRef := range ns.OwnerReferences {
 			if ownerRef.UID == project.UID {
 				logger.Debug("namespace exists and is already owned by this Project")
+				if ownerRef.Controller != nil {
+					ownerRef.Controller = nil
+				}
 				return nil
 			}
 		}
```
krancour, line 540:
```suggestion
		for i, ownerRef := range ns.OwnerReferences {
			if ownerRef.UID == project.UID {
				logger.Debug("namespace exists and is already owned by this Project")
				if ownerRef.Controller != nil {
					logger.Debug("owner reference requires update")
					ns.OwnerReferences[i].Controller = nil // Update in place
					if err = r.patchOwnerReferencesFn(ctx, r.client, ns); err != nil {
						return fmt.Errorf(
							"error patching namespace %q owner references: %w",
							project.Name,	err,
						)
					}
					logger.Debug("updated owner reference")
				}
				return nil
			}
		}
```

internal/controller/management/projects/projects.go
```diff
@@ -464,9 +529,20 @@ func (r *reconciler) ensureNamespace(ctx context.Context, project *kargoapi.Proj
...
+					ns.OwnerReferences[i].Controller = nil // Update in place
+					if err = r.patchOwnerReferencesFn(ctx, r.client, ns); err != nil {
+						return fmt.Errorf(
+							"error patching namespace %q owner references: %w",
+							project.Name,	err,
+						)
+					}
+					logger.Debug("updated owner reference")
+				}
 				return nil
```
krancour, line 546: This early return isn't new, but it's not sitting right with me and I think while we're already touching this logic, it's the right time to fix it.

There was a time when Projects were reconciled only once and once everything was initialized, they weren't reconciled again until deletion. That changed fairly recently.

An early return from here bypasses the logic further down for ensuring the existence of the finalizer on the namespace and we really want to ensure that unconditionally.

We may need to slightly re-order the sequence of events in this method. (All the right code is here, and I think things just need to move around.)

I will tinker with this if you don't mind @Kavinraja-G.

krancour: I think that would do the trick, but as I'm looking at how this method has now grown, I think it's probably possible to refactor it to be more readable and concise. I haven't looked hard enough to figure out yet what might be possible.

Kavinraja-G: Thanks @krancour, lmk if you need any help. Also is it possible to expedite this change to the upcoming releases?

krancour:
Just a note that I haven't forgotten about this. I did start playing with it and got sidetracked by some more urgent things.

My immediate observation was that breaking the delete/cleanup logic out into its own method makes it quite a bit easier both to comprehend and to test.

@Kavinraja-G if you had the bandwidth to do that little bit of refactoring, it would move this along a bit.

Kavinraja-G:
> Just a note that I haven't forgotten about this. I did start playing with it and got sidetracked by some more urgent things.
> 
> My immediate observation was that breaking the delete/cleanup logic out into its own method makes it quite a bit easier both to comprehend and to test.
> 
> @Kavinraja-G if you had the bandwidth to do that little bit of refactoring, it would move this along a bit.

@krancour Sure, did some initial refactoring https://github.com/akuity/kargo/pull/4229/commits/63328adbd6bc5fb6c5fdaab2ea68fe4d633a2c86, how about now?

Kavinraja-Ganesan:
@krancour bumping this.. 👀 Apologies if you are out this week.

docs/docs/50-user-guide/20-how-to-guides/20-working-with-projects.md
```diff
@@ -243,6 +243,15 @@ spec:
...
 ```
 
+## Preventing Namespace Deletion
+
+By default, when a Project is deleted, Kargo will attempt to delete the corresponding Namespace.
+However, there are scenarios where you may want to retain the namespace after the associated Project is removed
+whether it was created by Kargo or adopted from an existing setup.
+
+To achieve this, you can apply the following annotation to the `Namespace` or the corresponding `Project` resource
+with `kargo.akuity.io/keep-namespace: "true"`
```
krancour:
```suggestion
By default, when a Project is deleted, Kargo will attempt to delete the
corresponding Namespace. However, there are scenarios where you may want to
retain the namespace after the associated Project is removed whether it was
created by Kargo or adopted from an existing setup.

To achieve this, you can apply the following annotation to the `Namespace` or
the corresponding `Project` resource with
`kargo.akuity.io/keep-namespace: "true"`.
```

krancour:
@hiddeco if you're happy, I'm happy.

PR #4242, opened 2025-05-26
feat: added ability to specify GCP SA email to be used by the controller in a GKE cluster. Also added ability to specify other optional annotations if needed

ilyakazm:
# Add GCP IAM and Custom Annotations Support for Controller ServiceAccount

## Summary

- Added support for specifying a GCP IAM Service Account (`gcpSa`) for the controller’s Kubernetes `ServiceAccount`. When set, the appropriate GKE annotation is added.
- Added support for arbitrary user-defined annotations on the controller `ServiceAccount` via the `annotations` field.
- Updated Helm template, `values.yaml`, and documentation (`README.md`) to reflect these new options.

## Details

- `controller.serviceAccount.gcpSa`:  
  Allows specifying a GCP IAM Service Account email for GKE Workload Identity integration.

- `controller.serviceAccount.annotations`:  
  Allows users to add any custom annotations to the controller `ServiceAccount`.

Both new parameters are documented in `values.yaml` and the chart's `README.md`.

## Why

These changes make it easier to use the chart in GKE environments and provide flexibility for users to add any required `ServiceAccount` annotations.


charts/kargo/README.md
```diff
@@ -153,7 +153,9 @@ the Kargo controller is running.
...
+| `controller.serviceAccount.gcpSa`                                  | Specifies the email of a GCP IAM Service Account to be used by the controller in a GKE cluster. If set, the corresponding annotation will be added to the ServiceAccount. | `""` |
+| `controller.serviceAccount.annotations`                            | Additional annotations to add to the controller ServiceAccount. | `{}` |
+| `controller.serviceAccount.clusterWideSecretReadingEnabled`          | Specifies whether the controller's ServiceAccount should be granted read permissions to Secrets CLUSTER-WIDE in the Kargo control plane's cluster. Enabling this is highly discouraged and you do so at your own peril. When this is NOT enabled, the Kargo management controller will dynamically expand and contract the controller's permissions to read Secrets on a Project-by-Project basis.                                                                                                                                                                                                                                                                                                                               | `false`             |
```
hiddeco, line 158: Instead of creating a setting specifically for GCP. I think it would be better to just accept annotations being set.

ilyakazm: Fair point, indeed! I've removed the GCP specific one and just made it `iamRole` + any other custom annotations if such present. Also updated docs/comments to reflect that. Thanks for the review!

krancour:
Is there a reason this is being implemented only for the controller's SA?

krancour:
> similar to how AWS was done

I _somewhat_ wish that had been done differently way back then, but there is also a small difference between that and this... IRSA (older approach) doesn't work at all without that annotation on the controller's SA (although the newer approach, EKS Pod Identity does). Having a dedicated setting for it seemed to make sense if IRSA is in use and absolutely requires that annotation.

Contrasting the above with Workload Identity Federation, the annotation you're trying to set _isn't_ required, which makes the benefits of this PR feel arbitrary and deserving of being quite generalized.

All that said... if you are under the impression that this annotation _is_ required to make WIF work, that contradicts my understanding and means there is a flaw in our documentation -- although that particular bit of docs was updated not more than a month ago (at most) and I did personally validate the instructions at that time and never had to set this annotation.

Generalizing this would be welcome, but if instead we can determine that you don't need this annotation after all, that might be the fastest path to unblocking yourself.

krancour:
"ASAP" often comes across as very demanding. Putting that behind us, I have no interest in arguing and I'm glad it seems we can move past that.

> I'm surprised it is not already, knowing that a specific case was added for AWS at some point and no one asked to generalized it.

Note that I did explain how that came to be:

https://github.com/akuity/kargo/pull/4242#issuecomment-2913492063

charts/kargo/values.yaml
```diff
@@ -96,7 +101,12 @@ api:
   host: localhost
   ## @param api.logLevel The log level for the API server.
   logLevel: INFO
-
+  ## Service Account specific settings
+  serviceAccount:
+    ## @param api.serviceAccount.annotations Additional annotations to add to the API server ServiceAccount.
+    annotations: {}
+    # foo: bar
+    # another: value
```
krancour, line 109:
```suggestion
    # another: value

```

charts/kargo/values.yaml
```diff
@@ -26,7 +26,12 @@ global:
   podLabels: {}
   ## @param global.podAnnotations Annotations to add to pods.
   podAnnotations: {}
-
+  ## Service Account global settings
+  serviceAccount:
```
krancour, line 30: I appreciate you having done it this way. I regret that `labels` and `annotations` above were added without specifying _what_ those are applied to -- which led to the necessity of the ugly `podLabels` and `podAnnotations`. I'd love to change that if it weren't such a breaking change. At any rate, thank you for not perpetuating that ugly pattern.

charts/kargo/values.yaml
```diff
@@ -96,7 +101,12 @@ api:
   host: localhost
   ## @param api.logLevel The log level for the API server.
   logLevel: INFO
-
+  ## Service Account specific settings
```
krancour, line 104:
```suggestion
  ## ServiceAccount specific settings
```

charts/kargo/values.yaml
```diff
@@ -342,6 +352,12 @@ api:
       annotations: {}
       ## @param api.oidc.dex.securityContext Security context for Dex server pods. Defaults to `global.securityContext`.
       securityContext: {}
+      ## Service Account specific settings
```
krancour, line 355:
```suggestion
      ## ServiceAccount specific settings
```

charts/kargo/values.yaml
```diff
@@ -393,9 +409,13 @@ controller:
   serviceAccount:
     ## @param controller.serviceAccount.iamRole Specifies the ARN of an AWS IAM role to be used by the controller in an IRSA-enabled EKS cluster.
     iamRole: ""
+    ## @param controller.serviceAccount.annotations Additional annotations to add to the controller ServiceAccount.
+    annotations: {}
+    # foo: bar
+    # another: value
+
```
krancour, line 416:
```suggestion
```

charts/kargo/values.yaml
```diff
@@ -393,9 +409,13 @@ controller:
...
     ## @param controller.serviceAccount.clusterWideSecretReadingEnabled Specifies whether the controller's ServiceAccount should be granted read permissions to Secrets CLUSTER-WIDE in the Kargo control plane's cluster. Enabling this is highly discouraged and you do so at your own peril. When this is NOT enabled, the Kargo management controller will dynamically expand and contract the controller's permissions to read Secrets on a Project-by-Project basis.
     clusterWideSecretReadingEnabled: false
```
krancour, line 418:
```suggestion
    clusterWideSecretReadingEnabled: false

```

charts/kargo/values.yaml
```diff
@@ -580,6 +600,12 @@ externalWebhooksServer:
     # nodePort:
     ## @param externalWebhooksServer.service.annotations Annotations to add to the external webhook server's service. Merges with `global.annotations`, allowing you to override or add to the global annotations.
     annotations: {}
+  ## Service Account specific settings
```
krancour, line 603:
```suggestion
  ## ServiceAccount specific settings
```

charts/kargo/values.yaml
```diff
@@ -612,6 +638,12 @@ managementController:
     serviceAccounts:
       ## @param managementController.reconcilers.serviceAccounts.maxConcurrentReconciles optionally overrides the maximum number of ServiceAccount resources the management controller can reconcile concurrently.
       maxConcurrentReconciles:
+  ## Service Account specific settings
```
krancour, line 641:
```suggestion
  ## ServiceAccount specific settings
```

krancour:
This LGTM, although I may follow up myself with another PR to cleanup how/where the new `serviceAccount` section is positioned under each component. I'd like it to be pretty consistent, but there are already too many inconsistencies with how other sections are laid out from component-to-component, so it's not worth dealing with here.

krancour:
@ilyakazm this is totally generally useful and appreciated.

In the end, what was standing in the way of WIF working? It would be a good thing to know in case a doc clarification might be in order.

krancour:
Superseded by #4264

krancour:
> In any case, many thanks for adding WIF support in 1.5.0

Were you already on 1.5.0 when running into these issues? Falling back from a Project-specific GSA to the controller's GSA was new in 1.5.0.

> I'm wondering if a controller, after several attempts to authenticate with project-specific SA could fall back to its own KSA/GSA to get the token?

It should already be doing this. Whatever token it finds gets cached for some period. The cache key includes the Project name, so you might see attempts by multiple Projects to get a Project-specific GSA and then fall back to the controller's GSA. Or you might see one Project go through this anew each time the token expires.

If you have any logs you want to share, they could be helpful.

ilyakazm:
> Were you already on 1.5.0 when running into these issues? Falling back from a Project-specific GSA to the controller's GSA was new in 1.5.0.

Yes, I upgraded Kargo chart as soon as I saw 1.5.0 is out to test the WIF stuff.

> It should already be doing this. Whatever token it finds (regardless of being Project-specific or the controller's) gets cached for some period. The cache key includes the Project name, so you might see attempts by multiple Projects to get a Project-specific GSA and then fall back to the controller's GSA. Or you might see one Project go through this anew each time the token expires.

Awesome, good to know! Totally might be a cache issue here, indeed. We also suspected that at some point, tried to restart controller pod few times to see if this will force cache to clear - maybe  I was just too impatient and indeed just needed to wait a bit longer for token to refresh 🤷 Maybe deletion of that SA in question coincided with token renewal, not too sure.

As to logs - the logline that Antoine [shared](https://github.com/akuity/kargo/pull/4242#issuecomment-2916527269) earlier, was pretty useful as it pointed us to a thought that controller is likely still trying to use project-specific SA to get a token instead of falling back to its own SA - beside it I couldn't not spot anything noteworthy on other loglines

Issue #3286, opened 2025-01-15
Ability to do `helm dependency update`-equivalent without specifying a known subchart


jvmdc:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

Client Version: v1.1.2

# Proposed Feature
The ability to do `helm dependency update` without specifying a specific subchart.

# Motivation
Our `Chart.yaml` file has the following structure: 

```
apiVersion: v2
name: <Some name here>
type: application
version: 0.1.0

dependencies:
  - alias: <some alias>
    name: <subDependencyName>
    version: "0.1.0"
    repository: <Helm repo URL - Not OCI repository btw.>
```
Currently I am trying to setup a single-freight warehouse that simply points at the Git repository hosting the above `Chart.yaml` file (and `values.yaml` etc.). 
My PromotionTemplate steps are the following:
1. git-clone Git repository's main branch (hosting above file) to `src` dir, clone `dev` branch to `out` directory
2. git-clear `dev` branch
3. helm-template

The promotion will fail at the 3rd step because the dependency has not been fetched: ```step execution failed: step 2 met error threshold of 1: failed to run step "helm-template": missing chart dependencies: found in Chart.yaml, but missing in charts/ directory: <subDependencyName>```

A workaround that I can apply if I know my subDependency's name, repository and version is to add another step:
1. git-clone Git repository's main branch (hosting above file) to `src` dir, clone `dev` branch to `out` directory
2. git-clear `dev` branch
3. helm-update-chart (with an entry for the known subDependency in the `charts` array)
4. helm-template

This does require me to know the subDependency when I build my PromotionTemplate - As my setup is meant as a very generic setup across the organization, that is probably not feasible.

I had hoped to do it using `helm-update-chart` without any entries in the `charts` array, but that fails with an error: ```step execution failed: step 2 met error threshold of 1: failed to run step "helm-update-chart": invalid helm-update-chart config: charts: Array must have at least 1 items```

I could workaround the issue by having a dummy chart whose version I never change that I reference in `helm-update-chart`, but that would be a bit clunky.

# Suggested Implementation
Allow having an empty `charts` array in the `helm-update-chart` - Based on a brief glance at the code, that should still update dependencies.

krancour:
To summarize succinctly, you don't have the tarballs for the subcharts checked in -- which is fine and common.

[...]

PR #4245, opened 2025-05-26
feat(runner): allow building dependencies before Helm template

Brightside56:
Resolves https://github.com/akuity/kargo/issues/3286

I've copy-pasted some methods from helm-chart-update to helm-template just to be sure it works (and it works, at least helm-template). `buildDependencies` method is same as `updateDependencies` from helm_chart_updater.go with following exceptions:
- I've replaced manager.Update() with manager.Build() and I'm not sure if it's acceptable (but from what I know - it should have same effect)
- I've added lockFileExists as workaround, because helm-template tries to restore not existing Chart.lock backup in case there are no dependency updates

I've commented out `missing dependencies` test, because it seems to be not actual and ends up with 404 error

I would really appreciate any hints and ideas regarding how to factor out code which can be reused between helm-chart-update/helm-template steps


hiddeco:
FYI: I was initially busy getting `v1.5.1` out the door, but I picked this up today and expect to have a polished version soon that incorporates the abstraction I mentioned. Thanks for your work on this! 🙇

internal/helm/dependency_manager.go
```diff
@@ -0,0 +1,653 @@
...
+		case strings.HasPrefix(dep.Repository, "oci://"):
+			repository := NormalizeChartRepositoryURL(dep.Repository)
+			host := hostForRepositoryURL(repository)
+
+			if _, exists := hosts[host]; exists {
+				// We already logged in to this host, skip it
+				continue
+			}
+
+			credURL := "oci://" + path.Join(repository, dep.Name)
```
krancour, line 428: Number one thing I'd have worried we'd get wrong, but you nailed it.

krancour (approved):
🔥

Brightside56:
> As far as I am able to review my own code, this looks good to me now. Thank you @Brightside56! 🙇
> 
> @krancour if you can have an additional 👀, that would be great.

Thank! Sorry for being unavailable

Issue #4265, opened 2025-05-28
The Regular Stage with no verifications generates error logs during promotion


boh-dan:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

The Regular Stage with no verifications generates error logs during promotion(re-promoting freight).

# Steps to Reproduce

1. Create a Regular Stage without verification steps
2. Promote/Re-promote a freight to the Stage
3. Check Kargo Controller logs

Expected behaviour:
no errors

Actual behavior
level=error msg="error triggering re-verification" error="current stage verification info has no ID"

# Version

```
1.5.0
```

# Logs

time="2025-05-28T03:11:15Z" level=error msg="error triggering re-verification" error="current stage verification info has no ID" freight=aef86d6487fd7ab69d41752dd4cb873df3489aac namespace=addon-xxx promotion=stage-name-01jwady6w464rnkz0ppdyp7wq5.aef86d6 stage=stage-name targetFreight=aef86d6487fd7ab69d41752dd4cb873df3489aac

#Investigation

I've checked the `verificationHistory` of the latest promoted Freight in the Stage status:
```
"verificationHistory": [
   {
      "finishTime": "2025-05-21T10:28:48Z",
      "phase": "Successful",
      "startTime": "2025-05-21T10:28:48Z"
    }
]
```
It is created [here](https://github.com/akuity/kargo/blob/main/internal/controller/stages/regular_stages.go#L1068 ) and doesn't contain ID field.

PR #4269, opened 2025-05-28
fix(controller): set ID on (no-op) VerificationInfo

hiddeco:
Fixes: #4265

### DIFF
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -704,17 +704,23 @@ type AnalysisRunArgument struct {
 // process.
 type VerificationInfo struct {
 	// ID is the identifier of the Verification process.
-	ID string `json:"id,omitempty" protobuf:"bytes,4,opt,name=id"`
+	//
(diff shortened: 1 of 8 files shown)


api/v1alpha1/stage_types.go
```diff
@@ -704,17 +704,23 @@ type AnalysisRunArgument struct {
 // process.
 type VerificationInfo struct {
 	// ID is the identifier of the Verification process.
-	ID string `json:"id,omitempty" protobuf:"bytes,4,opt,name=id"`
+	//
+	// +kubebuilder:validation:Required
+	ID string `json:"id" protobuf:"bytes,4,opt,name=id"`
```
krancour, line 709: I'm sure you're two steps ahead of me, but what's the rationale for validation on fields that are only populated by the system?

hiddeco: The system would have alerted me that it was missing, and this bug would have never existed if it were already in place.

krancour (approved):

Issue #4237, opened 2025-05-26
Cannot promote a freight that has manual approval but is in a soaking window


mwalsh2atlassian:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

In this workflow, I am wanting to bypass my soaking window, so I have manually approved the freight. However when I try to promote this freight, it fails because the upstream provider has not met the soaking requirements.

Note: v1.4.X had similar issues with soaking time, but the freight's themselves had a "promote" button that allowed requesting the API to try promoting the freight regardless of what the UI thought should work, and this would allow the freight to go ahead.

# Screenshots

<img width="502" alt="Image" src="https://github.com/user-attachments/assets/8a19e2ab-58f6-4175-9486-93bd3d5bf3d0" />

# Steps to Reproduce

1. Create 2 stages, one that depends on the other with a requiredSoakingTime. 
2. Test that manual approval works to the 2nd stage when the freight is not active in the first stage
3. Promote something new to the first stage, manually approve it, then try to promote this freight.

# Version

1.5.0

# Logs

Marvin9:
UI should probably check approved in stages which it isn't already. Definitely a bug

krancour:
@Marvin9 we can probably do better than having the UI look at fields to determine whether a piece of Freight is available for promotion to a given Stage. This feels brittle because there's a lot that goes into making that determination and we probably do not want all that logic repeated in the UI. I've long wanted to refactor the "query Freight API," but at a glance, I believe it can retrieve a list of all the Freight available for a given Stage. Is that not being used?

Marvin9:
@krancour that will be real help. We use query freight to list available freight, but it doesn't really say **why** freight is not available.

krancour:
> So for that reason, we do soak time calculation in UI and it failed in this case.

Why is that?

I understand if you're doing some calculations on soak time since I think you now display the time remaining until soak criteria are met... but I assumed this to be mostly cosmetic?

i.e., If the query Freight API call is already returning just Freight available to a given Stage, with soak time having already been factored in, my expectation is that for any piece of Freight, any calculations you do (even if flawed) won't override availability/unavailability as determined by the back end.

krancour:
> I thought what you meant was to expand the API to why it is not available

[...]

PR #4270, opened 2025-05-28
fix(ui): soak time issue

Marvin9:
https://github.com/akuity/kargo/issues/4237

prioritise backend calculation if freight is available then it is available


krancour (approved):

Marvin9:
@rpelczar i addressed what we discussed offline

Issue #4227, opened 2025-05-23
CEL for webhook receivers constrains max receivers to 5


fuskovic:
# Problem

Webhook receivers have to have a unique name field.

We're currently using the following CEL to accomplish this:

```
	// +kubebuilder:validation:MaxItems=5
	// +kubebuilder:validation:XValidation:message="WebhookReceiverConfig must have a unique name",rule="self.all(i, size(self.filter(j, i.name == j.name)) == 1)"
```

Unfortunately, this is too computationally expensive e.g.

```
Forbidden: x-kubernetes-validations estimated rule cost total for entire OpenAPIv3 schema exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)]
```

So we added the following to reduce the cost factor:

```
 +kubebuilder:validation:MaxItems=5
```

Unfortunately, this limits max configurable webhook receivers to 5.

# Suggested Solution

A CEL validation rule with a cost factor that lets us configure more receivers than we can currently is a welcome improvement but ideally we have something that's not constrained.

If this can't be accomplished via CEL an alternative could be to implement a custom validation rule in Go for project configs [here](https://github.com/akuity/kargo/blob/main/internal/webhook/kubernetes/projectconfig/webhook.go)

PR #4276, opened 2025-05-29
feat: validate webhook receiver name uniqueness + add webhook receiver name to hashed path

fuskovic:
Closes https://github.com/akuity/kargo/issues/4227

Screenshot 2025-05-29 at 10 15 30 AM

Also Closes https://github.com/akuity/kargo/issues/4280

### DIFF
--- a/internal/webhook/external/path.go
+++ b/internal/webhook/external/path.go
@@ -8,11 +8,11 @@ import (
 
 // GenerateWebhookPath generates a unique path for a webhook based on the
 // provided webhook receiver name, project name, kind, and token.
-// The path is formatted as "/webhook/{provider}/{hash}" where the hash is
+// The path is formatted as "/webhook/{kind}/{hash}" where the hash is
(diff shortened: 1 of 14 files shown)


internal/webhook/external/path.go
```diff
@@ -6,9 +6,13 @@ import (
 	"strings"
 )
 
-func GenerateWebhookPath(project, provider, token string) string {
-	input := []byte(project + token)
+// GenerateWebhookPath generates a unique path for a webhook based on the
+// provided webhook receiver name, project name, kind, and token.
+// The path is formatted as "/webhook/{kind}/{hash}" where the hash is
+// a SHA-256 hash of the concatenated webhook receiver name, project name, and token.
+func GenerateWebhookPath(name, project, kind, token string) string {
```
krancour, line 13: Suggest "secret" instead of "token" because it's more generic. Individual receiver types might pass a secret here that isn't a token, but is something else.

krancour (approved):

Issue #4258, opened 2025-05-27
Add Ping Event Support for Github Webhook Receivers


fuskovic:
# Problem

After the initial configuration of a Github webhook, a ping event is sent to the webhook to validate it works.

We don't do anything with ping events at this time.

# Suggested Solution

After parsing the event with `go-github`, type-switch against the returned event and return a 200 if the event-type is a ping.

# Motivation

A webhook receiver returning a non-200 to Github for a ping event doesn't functionally change anything but, from a UX perspective I think it's nice to see that initial contact was made successfully after configuration.

PR #4278, opened 2025-05-29
feat: add ping event support

fuskovic:
Closes https://github.com/akuity/kargo/issues/4258

Screenshot 2025-05-29 at 3 47 58 PM

### DIFF
--- a/internal/webhook/external/github.go
+++ b/internal/webhook/external/github.go
@@ -132,6 +132,7 @@ func githubHandler(
 		case *gh.PushEvent:
 			repoWebURL := e.GetRepo().GetHTMLURL()
 			logger = logger.WithValues("repoWebURL", repoWebURL)
+			ctx = logging.ContextWithLogger(ctx, logger)
 			result, err := refreshWarehouses(ctx, c, namespace, repoWebURL)
 			if err != nil {
 				xhttp.WriteErrorJSON(w,
--- a/internal/webhook/external/github.go
+++ b/internal/webhook/external/github.go
@@ -59,10 +59,13 @@ func githubHandler(
 		// TODO(fuskovic): eventually switch on event type to perform
 		// different actions (e.g. refresh Promotion on PR merge)
 		eventType := r.Header.Get("X-GitHub-Event")
-		if eventType != "push" && eventType != "ping" {
-			xhttp.WriteErrorJSON(w,
+		switch eventType {
+		case "ping", "push":
+		default:
+			xhttp.WriteErrorJSON(
+				w,
 				xhttp.Error(
-					fmt.Errorf("only ping + push events are supported"),
+					fmt.Errorf("event type %q is not supported", eventType),
 					http.StatusNotImplemented,
 				),
 			)
--- a/internal/webhook/external/github.go
+++ b/internal/webhook/external/github.go
@@ -27,12 +27,9 @@ func githubHandler(
 ) http.HandlerFunc {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
-		logger := logging.LoggerFromContext(ctx).
-			WithValues("path", r.URL.Path)
-
-		logger.Debug("retrieving secret",
-			"secret-name", secretName,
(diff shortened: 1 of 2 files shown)


internal/webhook/external/github.go
```diff
@@ -59,10 +59,10 @@ func githubHandler(
 		// TODO(fuskovic): eventually switch on event type to perform
 		// different actions (e.g. refresh Promotion on PR merge)
 		eventType := r.Header.Get("X-GitHub-Event")
-		if eventType != "push" {
+		if eventType != "push" && eventType != "ping" {
```
krancour, line 62: This conditional will start to look unwieldy as we handle more events (like PRs being merged) and the same goes for the error message a few lines down. I would suggest that a switch here will be cleaner over time and that the error message can just say the event isn't supported instead of listing all those that are, so that message won't need to be updated for each new event.

```
switch(eventType) {
case: "ping", "push":
default:
	xhttp.WriteErrorJSON(
		w,
		xhttp.Error(fmt.Errorf("event type %q is not supported", eventType),
		http.StatusNotImplemented,
	)
	return
}
```

internal/webhook/external/github.go
```diff
@@ -114,53 +114,57 @@ func githubHandler(
...
+						repo,
+					),
+				},
 			)
-			return
-		}
+		case *gh.PushEvent:
+			repo := e.GetRepo().GetHTMLURL()
+			logger.Debug("source repository retrieved", "name", repo)
+			ctx = logging.ContextWithLogger(ctx, logger)
```
krancour, line 133: I know this line was present before, but what's it for? This really only needs to be done if you've added more context to the logger and want to propagate it down the call stack. I can't spot where you've added any additional context, although there are some spots in here where I think you could/should (and then you can keep this line, but probably move it to right after wherever you add context to the logger).

ex: Early in the method, you may want to add k/v pairs for receiver type and URL (or just the path maybe). After determining the repository the event originated from, that's another k/v pair you could add.

internal/webhook/external/github.go
```diff
@@ -114,53 +114,57 @@ func githubHandler(
...
+						"ping event received, webhook is configured correctly for %s",
+						repo,
+					),
+				},
 			)
-			return
-		}
+		case *gh.PushEvent:
+			repo := e.GetRepo().GetHTMLURL()
+			logger.Debug("source repository retrieved", "name", repo)
```
krancour, line 132: Being careful to not reduce the signal:noise ratio too much, I don't think this message is particularly valuable. I'd cut this.

krancour: Also related to signal:noise ratio: I'm on the fence about this, but I think many (not all) of the things being logged as errors throughout the handler would be more appropriate as debug-level messages.

Error messages are useful when something is actually wrong with our software, but in cases where an expected header is missing, signature couldn't be verified, etc., our software is working correctly and the information is really only useful to someone who is actively troubleshooting.

cc @hiddeco for any opinion he may have on this.

internal/webhook/external/github.go
```diff
@@ -59,10 +61,13 @@ func githubHandler(
...
-		if eventType != "push" {
-			xhttp.WriteErrorJSON(w,
+		switch eventType {
+		case "ping", "push":
+		default:
+			xhttp.WriteErrorJSON(
+				w,
 				xhttp.Error(
-					fmt.Errorf("only push events are supported"),
+					fmt.Errorf("event type %q is not supported", eventType),
```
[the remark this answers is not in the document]
krancour: Or even _just_ `%s`.

internal/webhook/external/github.go
```diff
@@ -27,7 +27,9 @@ func githubHandler(
 ) http.HandlerFunc {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
-		logger := logging.LoggerFromContext(ctx)
+		logger := logging.LoggerFromContext(ctx).
+			WithValues("path", r.URL.Path)
```
krancour, line 31: 👍

Issue #4172, opened 2025-05-21
UI: freight history improvements


Marvin9:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/0a091c80-d0a2-4bcb-b95a-9a6c05fcd247" />

- ~[ ] Rename "Freight History" to "Deployments"~
- [ ] Add time of promotion of that freight and freight creation time
- [ ] Add "Promote" action button

# Motivation

Ease of use and less navigation

krancour:
I think this new view would be fine, except for:

> Rename "Freight History" to "Deployments"

We're trying really hard to not use the word "deployments" because it makes Kargo sound like a GitOps agent. We want to keep the focus on _promotions_ which is really about managing desired state. Deployment's left to Argo CD.

Other than that, this all seems fine.

PR #4291, opened 2025-05-30
feat(ui): freight history improvements

Marvin9:
fixes https://github.com/akuity/kargo/issues/4172

Screenshot 2025-06-04 at 4 37 06 PM

### DIFF
--- a/ui/src/features/project/pipelines/freight/short-version-utils.ts
+++ b/ui/src/features/project/pipelines/freight/short-version-utils.ts
@@ -1,10 +1,10 @@
-export const shortVersion = (version: string = '') => {
-  if (version.length <= 12) {
+export const shortVersion = (version: string = '', length = 12) => {
+  if (version.length <= length) {
     return version;
   }
 
-  const prefix = version.slice(0, 6);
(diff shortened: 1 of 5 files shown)


rpelczar:
@Marvin9 I like that we added more data, but now I don't like the overall look of the table.
What about this:
- delete id col (not sure if we really need id on this table)
- do we really need promoted and created dates? Maybe we can have just one and add the title to the column?
- limit number of characters for aliases (+ ...)

rpelczar (commented):
@Marvin9 I like that we added more data, but now I don't like the overall look of the table.
What about this:
- delete id col (not sure if we really need id on this table)
- do we really need promoted and created dates? Maybe we can have just one and add the title to the column?
- limit number of characters for aliases (+ ...)

Marvin9:
Done @rpelczar I've kept promoted at though

rpelczar:
@krancour can you also take a look at this? Just check the screenshot.

Marvin9:
Merging it. @krancour feel free to comment and I will resolve in next PR

PR #4292, opened 2025-05-30
Update azure git provider to work with self hosted server

b3go:
Applying the same logic as github and gitlab git-provider in the azure provider.

The provider will accept if the ProviderName 'azure' is part of the hostname.

### DIFF
--- a/internal/gitprovider/azure/azure.go
+++ b/internal/gitprovider/azure/azure.go
@@ -79,6 +79,8 @@ func NewProvider(
 }
 
 // CreatePullRequest implements gitprovider.Interface.
+const errConvertPullRequestFmt = "error converting pull request %d: %w"
+
 func (p *provider) CreatePullRequest(
 	ctx context.Context,
 	opts *gitprovider.CreatePullRequestOpts,
@@ -119,7 +121,7 @@ func (p *provider) CreatePullRequest(
 	}
 	pr, err := convertADOPullRequest(adoPR)
 	if err != nil {
-		return nil, fmt.Errorf("error converting pull request %d: %w", adoPR.PullRequestId, err)
+		return nil, fmt.Errorf(errConvertPullRequestFmt, adoPR.PullRequestId, err)
 	}
 	return pr, nil
 }
@@ -143,7 +145,7 @@ func (p *provider) GetPullRequest(
 	}
 	pr, err := convertADOPullRequest(adoPR)
 	if err != nil {
-		return nil, fmt.Errorf("error converting pull request %d: %w", id, err)
+		return nil, fmt.Errorf(errConvertPullRequestFmt, id, err)
 	}
 	return pr, nil
 }
@@ -174,7 +176,7 @@ func (p *provider) ListPullRequests(
 	for _, adoPR := range *adoPRs {
 		pr, err := convertADOPullRequest(&adoPR)
 		if err != nil {
-			return nil, fmt.Errorf("error converting pull request %d: %w", adoPR.PullRequestId, err)
+			return nil, fmt.Errorf(errConvertPullRequestFmt, adoPR.PullRequestId, err)
 		}
 		pts = append(pts, *pr)
 	}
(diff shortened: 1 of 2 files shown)


hiddeco:
Can you please sign off on your commit? This is a requirement for us to be able to merge it. For instructions on how to do this, please see: https://github.com/akuity/kargo/pull/4292/checks?check_run_id=43187750239

internal/gitprovider/azure/azure.go
```diff
@@ -119,7 +121,7 @@ func (p *provider) CreatePullRequest(
 	}
 	pr, err := convertADOPullRequest(adoPR)
 	if err != nil {
-		return nil, fmt.Errorf("error converting pull request %d: %w", adoPR.PullRequestId, err)
+		return nil, fmt.Errorf(errConvertPullRequestFmt, adoPR.PullRequestId, err)
```
krancour: I appreciate that you were trying to eliminate what must feel like some obvious DRY exceptions, but this change makes this significantly less readable. One now needs to look _elsewhere_ to see the format string that the args following it are being applied to. Format strings are much more comprehensible when they are close to where they are being used.

b3go:
Hey guys, thanks for looking into it. I will incorporate your feedback as soon as I am back from vacation!

I am actually not very happy about the hostname/provider name logic to step into the self hosted url validation but couldn't make anything better up, without a bigger refactoring. If you have any ideas let me know. I will gladly change it.

krancour:
@b3go unless I am missing something, it seems this only makes the predicate work for self-hosted Azure devops, but nothing appears to be adjusting the client connection URL for self-hosted. It appears it will still, unconditionally, be of the form `https://dev.azure.com/org/<project>/_git/<repo>`.

b3go:
@krancour true, the path is the same on self hosted and cloud services. For older instances, which where upgraded from when it was still called team foundation server there seems to sometimes be a base route /tfs, which would be the only difference to the cloud service urls.

On the self hosted server org is called collection but that's about it. API is the same

krancour:
> the path is the same on self hosted and cloud services

But hostname/domain is not.

b3go:
> > the path is the same on self hosted and cloud services
> 
> But hostname/domain is not.

I probably misunderstood what you wanted to tell me then, sorry.

I thought about putting the call to `parseSelfHostedRepoUrl` at the end of the `parseRepoUrl` method, without the if. That would allow any url to be accepted and wouldn't enforce users to host their server with azure somewhere in the hostname. That would be more flexible but I am unsure if it could be problematic.

krancour:
@b3go you've updated how org, project, and repo name are extracted from a repo URL, but look at how those results are being used:

https://github.com/akuity/kargo/pull/4292/files#diff-c2a6f499400ce5b01e38e813062669873d033b1da374ce5b17e6dcdbb07227e4R66-R78

No matter what self-hosted URL you give it, the base URL of the API that the provider will connect to is _always_ be `https://dev.azure.com/<org>` and that obviously doesn't look like the correct base URL for your _self-hosted_ Azure DevOps' API server.

Apart from this, you also have not updated the predicate:

https://github.com/akuity/kargo/pull/4292/files#diff-c2a6f499400ce5b01e38e813062669873d033b1da374ce5b17e6dcdbb07227e4R32-R38

Without updating the predicate, this provider will never be applied for URLs like this one: `https://azure-devops.example.com/<collection>/<project>/_git/<repo>`

b3go:
Well looks like I was pretty naive at that. Should have given it more thought. Thanks for getting me on the right track. I will fix my PR when I am back!

b3go:
This bothered me the whole vacation, usually I don't deliver such botched work. I haven't tested it yet but I incorporated the feedback, fixed the provider connection url and signed of the commits

b3go:
Unfortunately I am currently unable to test this in my environment. My company has installed a major update for the corporate proxy and now the container images can't build as a lot of downloads get blocked. Don't know when this will be fixed.

But something else: the codecov check fails because the coverage decreases by 0,08% even though it shows 100% coverage for my patch. Any ideas where I can have a look to fix this?

krancour:
> But something else: the codecov check fails because the coverage decreases by 0,08% even though it shows 100% coverage for my patch. Any ideas where I can have a look to fix this?

That won't block the PR from being merged.

b3go:
I was finally able to test this in my local environment. It won't work unfortunately. I have outlined the reason in this Discussion: https://github.com/akuity/kargo/discussions/4598

The open-pr promotion step needs to clone the git repository, on Azure DevOps Server this currently only works with user:pw and not PAT. The gitprovider uses this go repo for accessing the Azure Devops Api to create PRs: [here](https://github.com/microsoft/azure-devops-go-api). This can only open connections using PATs. So either Kargo won't be able to clone the repo, or create the pr.

As I've stated in the discussion the only way to make this work is to pass the http.extraheader in git clone/push. I would be happy to contribute to that but need some feedback where to start.

PR #4307, opened 2025-06-02
fix: missing forward slash

fuskovic:
I noticed the webhook receiver URL rendering with a missing forward slash in the protocol scheme.

e.g.

`https:/...`

instead of:

`https://...`

I found `path.Join` was removing it and the URL rendered correctly after using `url.JoinPath` instead.

Screenshot 2025-06-02 at 7 30 00 AM

### DIFF
--- a/internal/controller/management/projectconfigs/project_configs.go
+++ b/internal/controller/management/projectconfigs/project_configs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 	"time"
 
 	"github.com/kelseyhightower/envconfig"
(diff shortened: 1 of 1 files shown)


fykaa (approved):

internal/controller/management/projectconfigs/project_configs.go
```diff
@@ -277,7 +276,7 @@ func (r *reconciler) newWebhookReceiver(
 			string(secret),
 		),
 	}
-	wr.URL = path.Join(r.cfg.ExternalWebhookServerBaseURL, wr.Path)
+	wr.URL = r.cfg.ExternalWebhookServerBaseURL + wr.Path
```
krancour: I think url.JoinPath() might cover some edge cases better than concatenation would. tbh, I didn't realize url.JoinPath() existed until just now. There are probably many more spots where we could have taken advantage of that.

hiddeco (approved):

Issue #4277, opened 2025-05-29
Freights never receive LongestCompletedSoak status. Only current freights can be promoted when soaking is required


m-messiah:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

The issue is continue of closed https://github.com/akuity/kargo/issues/4238: When promoting out of a task, the LongestCompletedSoak field is never applied to a freight. The bugfix in https://github.com/akuity/kargo/pull/4261 added the run of RemoveCurrentStage, but now the method `RemoveCurrentStage` panics with nilpointer itself because LongestCompletedSoak is still empty:
```
if soak > f.VerifiedIn[stage].LongestCompletedSoak.Duration {
	f.VerifiedIn[stage] = VerifiedStage{
		LongestCompletedSoak: &metav1.Duration{Duration: soak},
	}
}
```

# Steps to Reproduce

1. Have 2 stages without verifications created on the older version.
2. Updated kargo
3. Promote new freight to the stage - it stuck in verifying state and controller is full of nil-pointer panic errors.

# Version

```
v1.5.1
```

# Logs

```
time="2025-05-29T14:30:17Z" level=error msg="Reconciler error" Stage="{canary default}" controller=stage controllerGroup=kargo.akuity.io controllerKind=Stage error="panic: runtime error: invalid memory address or nil pointer dereference [recovered]" name=canary namespace=default reconcileID="\"1da66638-ac9c-4739-8f28-f7bc79536245\""
[...]

PR #4308, opened 2025-06-02
fix(api): nil pointers in `RemoveCurrentStage`

hiddeco:
Fixes: #4277

- Ensure `LongestCompletedSoak` is set based it does not have a value.
- Preservation of other fields, to avoid erasing the verification timestamp.

@krancour: please carefully check if this matches the idea you had when you originally implemented this, as I suspect (based on the API documentation) that the core of this bug originates from some mistake made at the time.


krancour:
Sorry for the delay on this. I'd added `do not merge yet` since @hiddeco had specifically asked me to look at it and I'd not yet gotten around to it. I lost track of the fact that I was the one holding this up.

LGTM

krancour (approved):

PR #4000, opened 2025-05-06
feat: webhook receivers

fuskovic:
Related to https://github.com/akuity/kargo/issues/1449

Implements:
- Reconciler for project configs
- Kube client indexer for indexing project configs by webhook receiver path
- Generic warehouse refresh helper with continue-on-error semantics
- IO pkg for limit reading
- HTTP handler for routing webhook events
- HTTP handler for Github webhook events

Updates:
- project configs to include webhook receivers
- `internal/http` to include utils for json formatting, and error-handling.
- go-github dep from `v56` -> `v71`

PR #4314, opened 2025-06-02
webhook receiver type enums

fuskovic:
Refactors `WebhookReceiverType` and `WebhookReceiverSecretKey` into enum types per: 

https://github.com/akuity/kargo/pull/4000#discussion_r2112004749

### DIFF
--- a/api/v1alpha1/project_config_types.go
+++ b/api/v1alpha1/project_config_types.go
@@ -5,15 +5,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const (
-	WebhookReceiverTypeGitHub = "GitHub"
-	// TODO(fuskovic): Add more receiver enum types(e.g. Dockerhub, Quay, Gitlab, etc...)
-)
-
-const (
-	WebhookReceiverSecretKeyGithub = "token"
-)
-
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:scope=Namespaced
 // +kubebuilder:subresource:status
--- a/internal/api/webhook_receivers.go
+++ b/internal/api/webhook_receivers.go
@@ -1,0 +1,30 @@
+package api
+
+// WebhookReceiverType is an enum representing the type of a webhook receiver.
+// It is used to identify the platform or service that the webhook receiver is
+// associated with, such as GitHub or Quay.
+type WebhookReceiverType string
+
+func (w WebhookReceiverType) String() string {
+	return string(w)
+}
+
+const (
+	WebhookReceiverTypeGitHub WebhookReceiverType = "GitHub"
+	WebhookReceiverTypeQuay   WebhookReceiverType = "Quay"
+	// TODO(fuskovic): Add more receiver enum types(e.g. Dockerhub, Quay, Gitlab, etc...)
+)
+
+// WebhookReceiverSecretKey is an enum representing the key used in a secret
+// for a webhook receiver. It is used to identify the specific key that contains
+// the secret data required for the webhook receiver to function properly.
+type WebhookReceiverSecretKey string
+
+func (w WebhookReceiverSecretKey) String() string {
+	return string(w)
+}
+
+const (
+	WebhookReceiverSecretKeyGithub WebhookReceiverSecretKey = "token"
+	WebhookReceiverSecretKeyQuay   WebhookReceiverSecretKey = "quay-secret"
+)
--- a/internal/controller/management/projectconfigs/project_configs.go
+++ b/internal/controller/management/projectconfigs/project_configs.go
@@ -20,6 +20,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/api"
 	"github.com/akuity/kargo/internal/conditions"
 	"github.com/akuity/kargo/internal/controller"
 	"github.com/akuity/kargo/internal/kubeclient"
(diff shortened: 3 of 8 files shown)


internal/api/webhook_receivers.go
```diff
@@ -0,0 +1,30 @@
...
+type WebhookReceiverSecretKey string
+
+func (w WebhookReceiverSecretKey) String() string {
+	return string(w)
+}
+
+const (
+	WebhookReceiverSecretKeyGithub WebhookReceiverSecretKey = "token"
+	WebhookReceiverSecretKeyQuay   WebhookReceiverSecretKey = "quay-secret"
+)
```
krancour: Contrasted with `WebhookReceiverType`, `WebhookReceiverSecretKey` seems as if it adds more inconvenience than it does value. The former gives use something we can annotate to indicate enumerated/accepted values. That helps with validation and docs generated from the CRDs. None of that seems applicable for these keys, which still ought to be constants, but just constant strings.

internal/api/webhook_receivers.go
```diff
@@ -0,0 +1,30 @@
+package api
+
+// WebhookReceiverType is an enum representing the type of a webhook receiver.
+// It is used to identify the platform or service that the webhook receiver is
+// associated with, such as GitHub or Quay.
+type WebhookReceiverType string
+
+func (w WebhookReceiverType) String() string {
+	return string(w)
+}
```
krancour: This isn't really needed. Variables of type `WebhookReceiverType` will already be treated like plain strings in most places and in the few places they aren't, typcasting is no harder than invoking `.String()`.

internal/api/webhook_receivers.go
```diff
@@ -0,0 +1,21 @@
...
+
+const (
+	WebhookReceiverTypeGitHub WebhookReceiverType = "GitHub"
+	WebhookReceiverTypeQuay   WebhookReceiverType = "Quay"
+	// TODO(fuskovic): Add more receiver enum types(e.g. Dockerhub, Quay, Gitlab, etc...)
+)
+
+const (
+	WebhookReceiverSecretKeyGithub string = "token"
+	WebhookReceiverSecretKeyQuay   string = "quay-secret"
```
krancour:
```suggestion
	WebhookReceiverSecretKeyGithub = "token"
	WebhookReceiverSecretKeyQuay   = "quay-secret"
```

internal/api/webhook_receivers.go
```diff
@@ -0,0 +1,21 @@
+package api
+
+// WebhookReceiverType is an enum representing the type of a webhook receiver.
+// It is used to identify the platform or service that the webhook receiver is
+// associated with, such as GitHub or Quay.
+type WebhookReceiverType string
```
krancour: This (and its values) really ought to stay in `project_config_types.go` so they can be annotated similarly to here to aid with validation and generated docs:

https://github.com/akuity/kargo/blob/b031b380a32bdc01d09f80d224f75a2b52e35f60/api/v1alpha1/warehouse_types.go#L15-L16

krancour: ~~The keys for the data maps can stay put.~~

https://github.com/akuity/kargo/pull/4314/files#r2122066126

internal/api/webhook_receivers.go
```diff
@@ -0,0 +1,17 @@
...
+
+const (
+	WebhookReceiverTypeGitHub WebhookReceiverType = "GitHub"
+	WebhookReceiverTypeQuay   WebhookReceiverType = "Quay"
+	// TODO(fuskovic): Add more receiver enum types(e.g. Dockerhub, Quay, Gitlab, etc...)
+)
+
+const (
+	WebhookReceiverSecretKeyGithub = "token"
+	WebhookReceiverSecretKeyQuay   = "quay-secret"
```
krancour, line 5: The won't need to be exported.

krancour: Oh... whoops... they don't need to be exported if they move to `webhook/external`. I don't think they are/will be used anywhere else.

fuskovic: They're still needed in the management controller for what is currently `getProviderConfig`

api/v1alpha1/project_config_types.go
```diff
@@ -5,6 +5,18 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// WebhookReceiverType is an enum representing the type of a webhook receiver.
+// It is used to identify the platform or service that the webhook receiver is
+// associated with, such as GitHub or Quay.
+//
+// +kubebuilder:validation:Enum={GitHub}
```
krancour, line 12: I suggested adding this, but I'm realizing now that we don't actually ask users to identify a receiver's type by name anywhere in their config. They identify the receiver type only by setting one and only one of the receiver config's fields to a non-nil value. 🤦‍♂️ 

This realization makes me question the value of `WebhookReceiverType` being defined here in the `api/v1alpha1` package, which was my idea based on the same flawed premise. Perhaps it was indeed better closer to where it was being used. 🤦‍♂️ 🤦‍♂️ 

Backing up even further to #4000, I'm not sure we necessarily even needed any sort of broadly accessible constants for these at all in the first place...

Aren't these ultimately used in only _one_ place? Used only in constructing receiver URLs?

I have a fair amount of refactoring I'll be PR'ing tomorrow to make webhook receiver logic reusable across ProjectConfig and (new) ClusterConfig reconcilers. Let's wait and see how that shakes out before making any more decisions about where these constants should be defined or whether we even need them at all.

krancour:
After #4326 we have no need for constants representing different types of receivers.

PR #4318, opened 2025-06-03
feat: file writer promotion step

### LINKED ISSUE #4267: File Writer Step
# Checklist

# Proposed Feature

File Writer 

File Writer is a promotion step that writes a to a file.
Proposed step definition:
```yaml
- uses: write
  config:
    outFile: ./out/changelog.json
    contents: |
      ${{ quote({
        "message": vars.message,
      }) }}
```

Using os.WriteFile semantics, this would behave as follows:

> WriteFile writes data to the named file, creating it if necessary. If the file does not exist, WriteFile creates it with permissions perm (before umask); otherwise WriteFile truncates it before writing, without changing permissions. Since WriteFile requires multiple system calls to complete, a failure mid-operation can leave the file in a partially written state.

# Motivation

Currently I have a need to update/write to  a file. I cannot use file copy as the contents must come from elsewhere e.g. a compose output or a http response.

As part of my promotion pipeline, I would greatly benefit from the flexiblity of writing to a file as opposed to using yaml or json update steps. In addition, it would be helpful to be able to use expression syntax to template the contents of the file in this way.

benduncan-tote:
Implementation of file writer promotion step

closes #4267

### DIFF
--- a/internal/promotion/runner/builtin/file_writer.go
+++ b/internal/promotion/runner/builtin/file_writer.go
@@ -1,0 +1,100 @@
+package builtin
+
+import (
+	"context"
+	"fmt"
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/pkg/promotion"
+	"github.com/akuity/kargo/pkg/x/promotion/runner/builtin"
+	securejoin "github.com/cyphar/filepath-securejoin"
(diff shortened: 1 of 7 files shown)


krancour:
@jessesuen is this a feature you want? #4267 is a proposal that was never triaged, discussed, or accepted.

benduncan-tote:
Happy to take another look at a later date if this feature gets accepted.

Issue #4150, opened 2025-05-19
add warehouse-refreshing webhook support for popular platforms


krancour:
This is an umbrella issue for several follow-ups to #4000.

We will need handlers _and docs_ for inbound webhooks from the following platforms, loosely grouped here by priority:

- High:
  - [ ] Docker Hub
  - [ ] GitHub (included in #4000)
  - [ ] GitHub Enterprise Cloud / GitHub Enterprise Server (this probably only involves tweaks to GH support)
  - [ ] GHCR (this probably only involves supporting another event from GH)
  - [ ] Quay
  
- Medium:
  - [ ] BitBucket / BitBucket Server
  - [ ] ECR
  - [ ] GitLab / GitLab self-hosted
  - [ ] GAR

- Low:
  - [ ] ACR
  - [ ] Artifactory
  - [ ] CloudEvents / Generic
  - [ ] Gitea

Ideally, we get through at least all the high and medium priority ones for v1.6.0 and can finish off the lower priority ones in v1.7.0 if we must.

@jessesuen feel free to modify the relative priorities as you see fit.

nikolay-te:
Would it make sense to also have a generic webhook receiver so that users can plug in any source not listed above?

krancour:
@nikolay-te that's already on the list.

nikolay-te:
🤦‍♂ haha, I see it now, I saw CloudEvents and stopped reading.
I'm going to get some more coffee, thanks! :)

nitishfy:
After discussing with Kargo team, I'll be working on GHCR.

krancour:
We nailed all the high priority ones. Moving this to the v1.7.0 milestone for what's left of it.

krancour:
Closing this in favor of discrete issues tracking remaining work items with varying levels of priority.

See:

#4486
#4487
#4488
#4489

Issue #2426, opened 2024-08-12
feature request: Customisable freight aliases


joaoestrela:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.

# Proposed Feature

Customisable freight aliases

# Motivation

With the introduction of Multi-Pipelines it would be nice to allow some customisation to freight aliases to better distinguish. It is possible to rename but generated recognisable aliases would be a great improvement.

# Suggested Implementation

1. Add a new string key to the Warehouse CRD that allows the addition of a prefix or suffix to the respective generated freights.
2. Add a boolean key to the Warehouse CRD that enables the prefixing or suffixing of the Warehouse name to the respective generated freights.

krzwiatrzyk:
Fully agree, it would be awesome to automatically generate freights with an alias equal image version (from semver) like that:
![image](https://github.com/user-attachments/assets/37b1bfd2-c7af-49fc-94f9-30ec3af3fda7)

krancour:
Bear in mind that a piece of Freight is a _collection_ of artifacts. A piece of Freight with just one artifact in it is the _simple_ case. It may seem intuitive in that case that the Freight alias would just match the tag of your Docker image, for instance, but that doesn't generalize well. The moment you have two or more artifacts in a piece of Freight (which is common, btw) that strategy falls apart. _Some kind_ of customization ability seems likely, eventually, but adopting a strategy of mirroring version information of a single artifact doesn't seem likely when it doesn't hold up for the common case.

antoine-sncf:
Contributing to the use case

**Possible situations to identify** 
1. Single subscription: use the item tag (image/chart)
3. Single image & Single chart : option to chose the image or chart as tag source (shouldn't -image- be the more popular option?)
4. Multiple subscription ( <x> chart and/or <y> image). Option to select a specific freight item as tag source (same as 3 in fact but with full subscription name ? could be merged with only 4)
5. External info (http call to a custom service?)
6. External trigger ? In my case using git  mono repo & -trunk based- development : when the project tags the main repo, our CI creates the images with their own semver (possibly) or maybe the same as the tag. We want the freight to be created only when the CI is finished (because we don't want the auto-assembly to trigger while it's still building possibly "many" images). 
So the question here are: who is sending the tag information 
- inside the CI ? does it sends all freights or le the auto discover handle it ?)
[...]

PR #4322, opened 2025-06-03
feat: quay webhook receiver support

fuskovic:
Related to: https://github.com/akuity/kargo/issues/4150

### DIFF
--- a/internal/webhook/kubernetes/projectconfig/webhook_test.go
+++ b/internal/webhook/kubernetes/projectconfig/webhook_test.go
@@ -583,7 +583,7 @@ func Test_webhook_ValidateCreate(t *testing.T) {
 				assert.Equal(t, "spec.webhookReceivers[0]", statusErr.ErrStatus.Details.Causes[0].Field)
 				assert.Equal(t, metav1.CauseTypeForbidden, statusErr.ErrStatus.Details.Causes[0].Type)
 				assert.Contains(t, statusErr.ErrStatus.Details.Causes[0].Message,
-					"only one webhook receiver can be defined at a time, found 3: [GitHub GitLab Quay]")
+					"cannot define a receiver that is of more than one type, found 3: [GitHub GitLab Quay]")
 			},
 		},
 	}


internal/controller/management/projectconfigs/project_configs_test.go
```diff
@@ -279,6 +279,87 @@ func TestReconciler_syncWebhookReceivers(t *testing.T) {
...
+					fake.NewClientBuilder().
+						WithScheme(scheme).
+						WithObjects(
+							&kargoapi.ProjectConfig{
+								ObjectMeta: metav1.ObjectMeta{
+									Name:      "fake-name",
+									Namespace: "fake-namespace",
+								},
+								Spec: kargoapi.ProjectConfigSpec{},
+							},
```
fykaa: is there a reason we are using an extra `ProjectConfig` stub with the name "fake-name"? curious if that stub is needed for something specific

fuskovic: No. This was an oversight. Deleted.

internal/webhook/external/quay_test.go
```diff
@@ -0,0 +1,186 @@
...
+		{
+			"name": "repository",
+			"repository": "mynamespace/repository",
+			"namespace": "mynamespace",
+			"docker_url": "quay.io/mynamespace/repository",
+			"homepage": "https://quay.io/repository/mynamespace/repository",
+			"updated_tags": [
+			  "latest"
+			]
+		  }
```
fykaa: would it make sense to simplify the payload like this, assuming we don’t currently need the other fields for our tests? if, I'd also trim down the docker hub test payloads to only include the required fields
but that said, if we anticipate needing fields like name, repository, namespace, or updated_tags in the near future (for validation or richer logging) then it might be worth keeping them in place. wdyt

```suggestion
		{
			"docker_url": "quay.io/mynamespace/repository"
		  }
```

fuskovic: Sure does. Done!

krancour:

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	xhttp "github.com/akuity/kargo/internal/http"
+	"github.com/akuity/kargo/internal/io"
+	"github.com/akuity/kargo/internal/logging"
+)
+
+const (
+	QuaySecretDataKey = "secret"
```
krancour: Doesn't need to be exported. If you saw it exported for GitHub, that was a unique exception because the ProjectConfig and ClusterConfig reconciler tests use that receiver in some of their tests.

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+) ([]string, error) {
+	secretValue, ok := secretData[QuaySecretDataKey]
+	if !ok {
+		return nil,
+			errors.New("Secret data is not valid for a Quay WebhookReceiver")
+	}
+	return []string{string(secretValue)}, nil
+}
+
+func (q *quayWebhookReceiver) GetHandler() http.HandlerFunc {
```
krancour, line 74:
```suggestion

// GetHandler implements WebhookReceiver.
func (q *quayWebhookReceiver) GetHandler() http.HandlerFunc {
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := logging.LoggerFromContext(ctx).WithValues("path", r.URL.Path)
+		ctx = logging.ContextWithLogger(ctx, logger)
```
krancour: The router will already add this to the context for all receivers.

```suggestion
		logger := logging.LoggerFromContext(ctx)
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+		logger.Debug("identifying source repository")
+
+		const maxBytes = 2 << 20 // 2MB
+		b, err := io.LimitRead(r.Body, maxBytes)
```
krancour: We're letting these limits be different for different senders so that if/when they need adjustment, they can be adjusted individually. GitHub for instance claims messages can be as large as 25 MB! (We left it at 2 MB for now, but this illustrates why we'd want to set these on a per sender basis.)

```suggestion
		b, err := io.LimitRead(r.Body, quayWebhookBodyMaxBytes)
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+		if err != nil {
+			xhttp.WriteErrorJSON(w,
+				xhttp.Error(
+					fmt.Errorf("failed to read request body: %w", err),
+					http.StatusRequestEntityTooLarge,
+				),
+			)
```
krancour: This is a new trick that io.ReadLimit got in #2426 -- it handles the errors message if the problem it encounters is the body being too large so that it's consistent across receivers.

```suggestion
			if errors.Is(err, &io.BodyTooLargeError{}) {
				xhttp.WriteErrorJSON(
					w,
					xhttp.Error(err, http.StatusRequestEntityTooLarge),
				)
				return
			}
			xhttp.WriteErrorJSON(w, err)
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+			return
+		}
+
+		payload := struct {
+			// format: quay.io/mynamespace/repository
+			RepoWebURL string `json:"docker_url"`
+		}{}
+
+		if err = json.Unmarshal(b, &payload); err != nil {
+			logger.Error(err, "failed to unmarshal request body")
```
krancour: This is one of those things that is not worth logging as an error. It was a bad request. Nothing is wrong that anyone looking at logs needs to know about.

```suggestion
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+			xhttp.WriteErrorJSON(w,
+				xhttp.Error(
+					fmt.Errorf("failed to unmarshal request body: %w", err),
+					http.StatusBadRequest,
+				),
```
krancour: We only need to respond with what's wrong, not where/how we choked.

```suggestion
				xhttp.WriteErrorJSON(
					w,
					xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
				)
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+			)
+			return
+		}
+
+		if payload.RepoWebURL == "" {
+			logger.Debug("missing repository web URL in request body")
```
krancour:
```suggestion
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+					fmt.Errorf("failed to read request body: %w", err),
+					http.StatusRequestEntityTooLarge,
+				),
+			)
+			return
+		}
+
+		payload := struct {
+			// format: quay.io/mynamespace/repository
+			RepoWebURL string `json:"docker_url"`
```
krancour: It would be better to name this field consistent with what they actually call it in the payload (while still observing Go naming conventions).

```suggestion
			dockerURL string `json:"docker_url"`
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+			xhttp.WriteErrorJSON(w,
+				xhttp.Error(
+					fmt.Errorf("missing repository web URL in request body"),
+					http.StatusBadRequest,
+				),
+			)
+			return
+		}
+
+		logger = logger.WithValues("repoWebURL", payload.RepoWebURL)
```
krancour:
```suggestion
		logger = logger.WithValues("repoURL", payload.dockerURL)
```

krancour: "repoWebURL" was some vestigial remnant of when we were, at one time, prior to #2326 reading info from a different (wrong) field in GitHub payload.

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,157 @@
...
+		ctx = logging.ContextWithLogger(ctx, logger)
+		result, err := refreshWarehouses(ctx, q.client, q.project, payload.RepoWebURL)
+		if err != nil {
+			xhttp.WriteErrorJSON(w,
+				xhttp.Error(err, http.StatusInternalServerError),
+			)
```
krancour: The behavior of `xhttp.WriteErrorJSON` is already to treat everything that isn't otherwise qualified as a `http.StatusInternalServerError`, so all cases of an unclassified error do not require any special wrapping.

```suggestion
			xhttp.WriteErrorJSON(w, err)
```

api/v1alpha1/project_config_types.go
```diff
@@ -120,6 +127,18 @@ type GitHubWebhookReceiverConfig struct {
 	SecretRef corev1.LocalObjectReference `json:"secretRef" protobuf:"bytes,1,opt,name=secretRef"`
 }
 
+type QuayWebhookReceiverConfig struct {
```
krancour, line 145:
```suggestion
// QuayWebhookReceiverConfig describes a webhook receiver that is compatible
// with Quay.io payloads.
type QuayWebhookReceiverConfig struct {
```

api/v1alpha1/project_config_types.go
```diff
@@ -120,6 +127,20 @@ type GitHubWebhookReceiverConfig struct {
...
+// with Quay.io payloads.
+type QuayWebhookReceiverConfig struct {
+	// SecretRef contains a reference to a Secret. For Project-scoped webhook
+	// receivers, the referenced Secret must be in the same namespace as the
+	// ProjectConfig.
+	//
+	// For cluster-scoped webhook receivers, the referenced Secret must be in the
+	// designated "cluster Secrets" namespace.
+	//
+	// The Secret's data map is expected to contain a `secret` key.
```
krancour: Comments on these types eventually wind up in some generated CRD docs...

More info about what this is supposed to match would be helpful. For GitHub, we're saying this:

```
	// The Secret's data map is expected to contain a `secret` key whose value is
	// the shared secret used to authenticate the webhook requests sent by GitHub.
	// For more information please refer to GitHub documentation:
	//   https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries
```

```suggestion
	// The Secret's data map is expected to contain a `secret` key.
```

fuskovic: Quay doesn't support a shared secret. So I suppose I can be explicit about that here. 

e.g.

`quay does not support a shared secret. However, webhook receivers account for this by always supporting a secret on our side; allowing us to always generate a secure webhook receiver URL`

Thoughts?

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,151 @@
...
+	return []string{string(secretValue)}, nil
+}
+
+// GetHandler implements WebhookReceiver.
+func (q *quayWebhookReceiver) GetHandler() http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := logging.LoggerFromContext(ctx)
+		logger.Debug("identifying source repository")
+
```
krancour, line 79: In many cases where the body is too large, we can spare ourselves the effort of the io.LimitRead. (However, we cannot trust that the `Content-Length` is always set, so falling back on the io.LimitRead is still imporant.)

```suggestion

		// Early check of Content-Length if available
		if contentLength := r.ContentLength; contentLength > githubWebhookBodyMaxBytes {
			xhttp.WriteErrorJSON(
				w,
				xhttp.Error(
					fmt.Errorf("content exceeds limit of %d bytes", githubWebhookBodyMaxBytes),
					http.StatusRequestEntityTooLarge,
				),
			)
			return
		}

```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,151 @@
...
+func (q *quayWebhookReceiver) getSecretValues(
+	secretData map[string][]byte,
+) ([]string, error) {
+	secretValue, ok := secretData[quaySecretDataKey]
+	if !ok {
+		return nil,
+			errors.New("Secret data is not valid for a Quay WebhookReceiver")
+	}
+	return []string{string(secretValue)}, nil
+}
```
krancour: We don't appear to be doing anything with `secretValue`. Do we need to? Or do we not need this info here?

I've not read up on Quay webhooks, but either this plays a role in authenticating the request somehow, or if it doesn't (e.g. case where the unguessable URL is is own implicit authentication mechanism), then we don't need to do anything with it. Not sure which of those it is.

internal/webhook/external/quay_test.go
```diff
@@ -0,0 +1,120 @@
...
+			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
+				require.Equal(t, http.StatusRequestEntityTooLarge, rr.Code)
+				res := map[string]string{}
+				err := json.Unmarshal(rr.Body.Bytes(), &res)
+				require.NoError(t, err)
+				require.Contains(t, res["error"], "content exceeds limit")
+			},
+		},
+		{
+			name:       "success -- push event",
```
krancour: Nit: Not worth qualifying as "push event" when it's the only event type we're responding to.

Separate from that, I'd suggest adding a partial success test case like the one in #4352

```suggestion
			name:       "success",
```

internal/webhook/external/quay.go
fuskovic: Quay doesn't support an auth mechanism.

api/v1alpha1/project_config_types.go
fuskovic: I've updated it to reflect the above.

krancour: Ok. Similar situation to Docker Hub, huh?

This is what #4351 says. Perhaps just adapt this.

```
	// The Secret's data map is expected to contain a `secret` key whose value
	// does NOT need to be shared directly with Docker Hub when registering a
	// webhook. It is used only by Kargo to create a complex, hard-to-guess URL,
	// which implicitly serves as a shared secret. For more information about
	// Docker Hub webhooks, please refer to the Docker documentation:
	//   https://docs.docker.com/docker-hub/webhooks/
```

internal/webhook/external/quay.go
krancour: Ok... Just drop this whole stanza then.

fuskovic: Should we leave a comment? If so,  how should it read?

```
	// Because Quay does not provide support for a secret
	// during configuration, the unguessable URL is it's 
	// own implicit authentication mechanism. 
	// So we don't need to do anything here.
```
Thoughts?

fuskovic: I've removed the logic but kept the function signature ( so the interface is still implemented ).
I've added a comment that reflects the above. Let me know if you want it nixed or updated.

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,161 @@
...
+// getSecretValues implements WebhookReceiver.
+func (q *quayWebhookReceiver) getSecretValues(
+	_ map[string][]byte,
+) ([]string, error) {
+	// Because Quay does not provide support for a secret
+	// during configuration, the unguessable URL is it's
+	// own implicit authentication mechanism.
+	// So we don't need to do anything here.
+	return nil, nil
+}
```
krancour: The comment is true up until the not needing to do anything part. This method is called by the code that constructs the hard-to-guess URL. It needs the value of the `secret` key because that's essentially a salt for the hash.

```suggestion
func (q *quayWebhookReceiver) getSecretValues(
	secretData map[string][]byte,
) ([]string, error) {
	secretValue, ok := secretData[quaySecretDataKey]
	if !ok {
		return nil,
			errors.New("Secret data is not valid for a Quay.io WebhookReceiver")
	}
	return []string{string(secretValue)}, nil
}
```

internal/webhook/external/quay.go
```diff
@@ -0,0 +1,161 @@
...
+			return
+		}
+
+		logger = logger.WithValues("repoURL", payload.DockerURL)
+		ctx = logging.ContextWithLogger(ctx, logger)
+		result, err := refreshWarehouses(ctx, q.client, q.project, payload.DockerURL)
+		if err != nil {
+			xhttp.WriteErrorJSON(w, err)
+			return
+		}
```
krancour: Nit: Slightly easier to follow what's going on if we bring a `repoURL` variable into play because it's _the_ thing we put all this effort into finding and Warehouses are indexed by repo URLs. It sort of connects the dots.

```suggestion
		repoURL := payload.DockerURL

		logger = logger.WithValues("repoURL", repoURL)
		ctx = logging.ContextWithLogger(ctx, logger)

		result, err := refreshWarehouses(ctx, q.client, q.project, repoURL)
		if err != nil {
			xhttp.WriteErrorJSON(w, err)
			return
		}
```

internal/webhook/external/quay_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	const testToken = "mysupersecrettoken"
+	testSecretData := map[string][]byte{
+		quaySecretDataKey: []byte(testToken),
+	}
+
+	for _, testCase := range []struct {
+		name       string
+		client     client.Client
+		secretData map[string][]byte
```
krancour: This field isn't relevant for any of these test cases because this particular handler doesn't do anything with the secret. The secret's only used in url construction and that has no relevance to testing the handler

```suggestion
```

krancour:
A few small things to fix and then this LGTM.

Although, this will probably face some merge conflicts and have to grapple with improved validation after #4332 is merged.

fuskovic:
> A few small things to fix and then this LGTM.
> 
> Although, this will probably face some merge conflicts and have to grapple with improved validation after #4332 is merged.

Addressed the merge conflicts and refactored the CEL validation into Go.

I manually tested as well though so you can see what the error message looks like for this edge case.

validation_test

`Error: apply resource: invalid_argument: admission webhook "projectconfig.kargo.akuity.io" denied the request: ProjectConfig.kargo.akuity.io "kargo-demo" is invalid: spec.webhookReceivers[0]: Forbidden: only one webhook receiver can be defined at a time, found 2: [GitHub GitLab]`

internal/webhook/kubernetes/external/validation.go
```diff
@@ -11,6 +11,48 @@ import (
...
+			receivers = append(receivers, "Quay")
+		}
+		if len(receivers) > 1 {
+			errs = append(errs, field.Forbidden(
+				f.Index(i),
+				fmt.Sprintf(
+					"only one webhook receiver can be defined at a time, found %d: %s",
+					len(receivers),
+					receivers,
+				),
```
krancour: This message seems misleading. You can register more than one receiver at a time. What you cannot do is define one receiver that is of more than one type.

internal/webhook/kubernetes/projectconfig/webhook_test.go
```diff
@@ -546,6 +546,46 @@ func Test_webhook_ValidateCreate(t *testing.T) {
 					"webhook receiver name already defined at spec.webhookReceivers[0]")
 			},
 		},
+		{
+			name: "mutually exclusive webhook receiver config",
```
krancour: We've tended toward naming test cases for the circumstances they exercise, so I might suggest:

```suggestion
			name: "webhook receiver contains config for multiple receiver types",
```

krancour (approved):

PR #4323, opened 2025-06-03
chore(runner): tidy `http` step internals

hiddeco:
- Use the `io.LimitRead` function that was introduced to support the webhook logic.
- Consume status codes from the `http` library.
- Use `mime.ParseMediaType()` to parse the content type.
- Move constants out of method.


internal/promotion/runner/builtin/http_requester.go
```diff
@@ -24,6 +24,8 @@ import (
 const (
 	contentTypeHeader = "Content-Type"
```
krancour, line 25: Non-blocking nit.

This constant is now only used by tests, I think. Perhaps move it?

internal/promotion/runner/builtin/http_requester.go
```diff
@@ -272,7 +254,7 @@ func (h *httpRequester) wasRequestSuccessful(
 	default:
 		// The client automatically follows redirects, so we consider only
 		// 2xx status codes successful.
-		return statusCode >= 200 && statusCode < 300, nil
+		return statusCode >= http.StatusOK && statusCode < http.StatusMultipleChoices, nil
```
krancour: I generally agree with using status constants over int literals where we can, but in the case of specifying a _range_ as we are here, I wonder if doing so makes this harder to understand.

hiddeco: Do not have a strong opinion on this in this specific scenario, and am happy to revert it.

internal/promotion/runner/builtin/http_requester.go
hiddeco: Yeah, good catch. Will address.

internal/promotion/runner/builtin/http_requester.go
krancour: I'm very narrowly in favor of reverting just where ranges are concerned.

Preemptively approving though.

krancour (approved):

internal/promotion/runner/builtin/http_requester.go
hiddeco: It is still used in-code on L210.

PR #4324, opened 2025-06-03
feat: dockerhub webhook receiver

fykaa:
Part of https://github.com/akuity/kargo/issues/4150

this PR introduces initial support for receiving and processing webhook events from DockerHub in the Kargo webhook server


internal/webhook/external/dockerhub_test.go
```diff
@@ -0,0 +1,203 @@
...
+                "owner": "someuser",
+                "repo_name": "someuser/somerepo",
+                "repo_url": 
+					"https://registry.hub.docker.com/u/someuser/somerepo/",
+                "star_count": 0,
+                "status": "Inactive"
+            }
+        }
+    `))
+}
```
fuskovic: Same as the other PR's. Just need the fields we're exercising + `bytes.NewBufferString`

internal/webhook/external/dockerhub.go
```diff
@@ -0,0 +1,144 @@
...
+		IsTrusted       bool    `json:"is_trusted"`
+		Name            string  `json:"name"`
+		Namespace       string  `json:"namespace"`
+		Owner           string  `json:"owner"`
+		RepoName        string  `json:"repo_name"`
+		RepoURL         string  `json:"repo_url"`
+		StarCount       int     `json:"star_count"`
+		Status          string  `json:"status"`
+	} `json:"repository"`
+}
```
fuskovic: This looks excessive since we're not interested or using a large majority of this. I think we can just use an anonymous struct to target what we need.

internal/webhook/external/dockerhub.go
```diff
@@ -0,0 +1,144 @@
...
+		if len(b) == 0 {
+			logger.Debug("empty request body")
+			xhttp.WriteErrorJSON(w,
+				xhttp.Error(
+					fmt.Errorf("empty request body"),
+					http.StatusBadRequest,
+				),
+			)
+			return
+		}
```
fuskovic: I'm not sure this is necessary but if we're going to keep it seems easy to add a test case for it.

internal/webhook/external/dockerhub.go
```diff
@@ -0,0 +1,144 @@
...
+		if payload.Repository.RepoName == "" {
+			logger.Debug("missing repository name in Docker Hub request body")
+			xhttp.WriteErrorJSON(w,
+				xhttp.Error(
+					fmt.Errorf("missing repository name in request body"),
+					http.StatusBadRequest,
+				),
+			)
+			return
+		}
```
fuskovic, line 66: Is it possible to get a webhook event from dockerhub that's missing the repo name?

fuskovic:
I would love to see a demo for this in Slack. The PR's we've been doing for this can be time-intensive to setup. Since we have to do it ourselves anyway during implementation we can make things easier for ourselves when we review by just recording our manual testing.

internal/webhook/external/dockerhub.go
```diff
@@ -0,0 +1,144 @@
...
+					fmt.Errorf("empty request body"),
+					http.StatusBadRequest,
+				),
+			)
+			return
+		}
+
+		payload := BuildPayload{}
+
+		if err = json.Unmarshal(b, &payload); err != nil {
```
nitishfy, line 53: Irrespective of the fact whether you plan to use the library (I mentioned in other comment) or not, I think it would be good to add this logic for parsing into a separate function. This way we can have unit tests for parsing.

internal/webhook/external/dockerhub.go
fykaa: It’s pretty unlikely since docker hub's webhook payloads always includes the `repo` object and its `repo_name` field by default. that said, I still think we should validate incoming payloads, just to be safe. theres always a chance the sender is misconfigured or buggy or someone could manually POST malformed data to the endpoint OR if docker hub changes their payload format in the future (very rare but ig possible)

krancour:
Closing in favor of #4351

PR #4326, opened 2025-06-04
feat: add support for cluster-scoped webhook receivers defined by new clusterconfig resource

krancour:
Adds support for cluster-scoped webhook receivers via a new ClusterConfig CRD.

This also contains quite a bit of refactoring of #4000 with an aim toward making certain parts reusable across both ProjectConfig and ClusterConfig reconcilers. Some of that could also be reused for routing/handling, so the refactor is sizable...

This refactoring also creates a `WebhookReceiver` interface to be implemented by receivers for all supported platforms. `NewWebhookReceiver()` is the _only_ switch anywhere on which kind of receiver has been configured. It instantiates an appropriate implementation of `WebhookReceiver`. These can/should be composed of a `baseWebhookReceiver`, which implements all the boilerplate for a receiver. This leaves very little for each individual implementation of `WebhookReceiver` to actually implement.

cc @fuskovic you asked to be tagged when I added support for cluster-scoped webhook receivers...


api/v1alpha1/cluster_config_types.go
```diff
@@ -0,0 +1,65 @@
...
+	c.Conditions = conditions
+}
+
+// +kubebuilder:object:root=true
+
+// ClusterConfigList contains a list of ClusterConfigs.
+type ClusterConfigList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+	Items           []ClusterPromotionTask `json:"items" protobuf:"bytes,2,rep,name=items"`
```
hiddeco:
```suggestion
	Items           []ClusterConfig `json:"items" protobuf:"bytes,2,rep,name=items"`
```

api/v1alpha1/cluster_config_types.go
```diff
@@ -0,0 +1,65 @@
...
+	// Conditions contains the last observations of the ClusterConfig's current
+	// state.
+	//
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchMergeKey:"type" patchStrategy:"merge" protobuf:"bytes,1,rep,name=conditions"`
+	// WebhookReceivers describes the status of Project-specific webhook receivers.
+	WebhookReceivers []WebhookReceiver `json:"webhookReceivers,omitempty" protobuf:"bytes,2,rep,name=webhookReceivers"`
```
hiddeco:
```suggestion
	// WebhookReceivers describes the status of cluster-level webhook receivers.
	WebhookReceivers []WebhookReceiver `json:"webhookReceivers,omitempty" protobuf:"bytes,2,rep,name=webhookReceivers"`
```

api/v1alpha1/cluster_config_types.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (c *ClusterConfig) GetStatus() *ClusterConfigStatus {
+	return &c.Status
+}
+
+// ClusterConfigSpec describes cluster-level Kargo configuration.
+type ClusterConfigSpec struct {
+	// WebhookReceivers describes Project-specific webhook receivers used for
+	// processing events from various external platforms
```
hiddeco:
```suggestion
	// WebhookReceivers describes cluster-level webhook receivers used for
	// processing events from various external platforms.
```

internal/webhook/external/receiver.go
```diff
@@ -0,0 +1,111 @@
...
+// generateWebhookPath generates a unique path for inbound webhooks based on the
+// provided Project name, receiver name, receiver type, and secret values. The
+// path is formatted as "/webhook/{receiverType}/{hash}" where the hash is a
+// SHA-256 hash of the concatenated project name, receiver name, and secret
+// values.
+func generateWebhookPath(
+	project string,
+	name string,
+	receiverType string,
+	secretVals []string,
```
hiddeco:
```suggestion
	secretValues ...string,
```

internal/webhook/external/receiver.go
```diff
@@ -0,0 +1,111 @@
...
+	var targetKeys []string
+	switch {
+	case cfg.GitHub != nil:
+		receiverType = kargoapi.WebhookReceiverTypeGitHub
+		secretName = cfg.GitHub.SecretRef.Name
+		targetKeys = []string{kargoapi.WebhookReceiverSecretKeyGithub}
+	default:
+		return "", nil, errors.New(
+			"webhook receiver config has no configuration for a known receiver type",
+		)
```
hiddeco: I assume the idea is still that this will be refactored eventually? As I feel this should be put at a higher level.

api/v1alpha1/cluster_config_types.go
krancour: Well that's obviously a mistake. Nice catch.

internal/webhook/external/receiver.go
krancour: I intended to untangle this further and missed it.

internal/io/io.go
```diff
@@ -32,7 +53,7 @@ func LimitRead(r io.ReadCloser, limit int64) ([]byte, error) {
 			)
 		}
 		if n > 0 {
-			return nil, fmt.Errorf("content exceeds limit of %d bytes", limit)
+			return nil, newBodyTooLargeError(limit)
```
hiddeco: Nit: using a constructor here seems kind of an overkill.

internal/webhook/external/receiver.go
```diff
@@ -0,0 +1,182 @@
...
+	getSecretName() string
+	// getSecretValues extracts a list of receiver-specific values from the
+	// provided Secret data.
+	getSecretValues(map[string][]byte) ([]string, error)
+	// setSecretData sets the Secret data for this receiver. This is used to
+	// later when handling inbound webhooks.
+	setSecretData(map[string][]byte)
+	// setDetails sets the details of the WebhookReceiver in the form of
+	// kargoapi.WebhookReceiverDetails.
+	setDetails(kargoapi.WebhookReceiverDetails)
```
hiddeco, line 32: Having an interface with private methods feels slightly awkward.

internal/webhook/external/receiver.go
```diff
@@ -0,0 +1,182 @@
...
+	// configuration provided.
+	var receiver WebhookReceiver
+	switch {
+	case cfg.GitHub != nil:
+		receiver = newGitHubWebhookReceiver(c, project, cfg)
+	default:
+		return nil, errors.New(
+			"WebhookReceiverConfig has no configuration for a known receiver type",
+		)
+	}
```
hiddeco: When the number of implementations grows, I think we would benefit from a registry/factory pattern. It might be worth taking note of this here.

internal/webhook/external/github.go
```diff
@@ -6,57 +6,66 @@ import (
...
+	token, ok := secretData["token"]
+	if !ok {
+		return nil,
+			errors.New("Secret data is not valid for a GitHub WebhookReceiver")
+	}
+	return []string{string(token)}, nil
+}
+
+// GetHandler implements WebhookReceiver.
+func (g *githubWebhookReceiver) GetHandler() http.HandlerFunc {
```
hiddeco, line 76: It feels like this could be broken up even more, e.g.

```go
func (g *githubWebhookReceiver) GetHandler() http.HandlerFunc {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        ctx := r.Context()

        if err := g.validateRequest(r); err != nil {
            g.handleError(w, err)
            return
        }

        event, err := g.parseWebhook(r)
        if err != nil {
            g.handleError(w, err)
            return
        }

        g.handleEvent(ctx, w, event)
    })
}
```

internal/webhook/external/github.go
```diff
@@ -6,57 +6,66 @@ import (
...
+// GetDetails implements WebhookReceiver.
+func (g *githubWebhookReceiver) getReceiverType() string {
+	return "github"
+}
+
+// getSecretValues implements WebhookReceiver.
+func (g *githubWebhookReceiver) getSecretValues(
+	secretData map[string][]byte,
+) ([]string, error) {
+	token, ok := secretData["token"]
```
hiddeco: Should `token` be a `const`?

internal/webhook/external/github.go
```diff
@@ -71,83 +80,74 @@ func githubHandler(
 			return
 		}
 
+		logger = logger.WithValues("eventType", eventType)
+		ctx = logging.ContextWithLogger(ctx, logger)
+
 		const maxBytes = 2 << 20 // 2MB
```
hiddeco: If `token` becomes a `const`, it might be worth moving this into that block as well (instead of within the method).

internal/webhook/external/github.go
krancour: This isn't as easy to break up as it may seem. Validating the request and extracting information from the payload often _both_ require reading the body of the request, and because the body is an `io.ReadCloser`, you can really only do that once. We could get clever by doing a limit read on the whole body up front and replacing `r.Body` before each of the validate and parse method calls. I'm not sure that's preferable to what we're doing now.

internal/webhook/external/github.go
krancour: It should. Will do.

Separately, I have reservations about it being called "token." When registering a webhook in GH, they call it "secret."

I'll have to double-check docs to see if they make frequent use of the word "token" in describing the registration process, but even if they did, my gut says matching what it's called on the registration form would remain more intuitive.

Any thoughts?

internal/webhook/external/receiver.go
krancour: It's not common, for sure, but I'll try to give some insight into my thought process and if you can propose a cleaner way to address my concerns, I'll be open to it...

If you look at how new `NewReceiver()` initializes any `WebhookReceiver` implementation, it has to use a _partially initialized_ receiver to retrieve some information that it uses to construct other information, which it then hands back to the receiver, at which point it is _fully initialized._

These methods exist only to facilitate that two-phased initialize. I don't think they're of much use from outside the package. In fact, I'd consider it a leaky abstraction if they _were_ exposed, as I don't think anyone else has any justifiable reason to be calling them.

If you can propose a cleaner way to do this sort of two-phased initialization, I'm happy to hear it.

internal/io/io.go
krancour: Happy to remove.

internal/webhook/external/receiver.go
krancour: Agree. I was just happy to get it down to a single switch that needs to be updated instead of three.

I'll add a comment and I'll be happy to revisit it in a follow-up.

internal/webhook/external/github.go
krancour: Separately, I was also thinking the 2MB limit also seems ridiculously big. It's copy/pasta from the original limit reading code... but apparently the max body size of an inbound webhook from GH _is a whopping 25MB._

I'm really on the fence about lowering this, raising it, or just leaving it alone.

Any thoughts?

internal/webhook/external/github.go
hiddeco: They use "token" in their [titles](https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries#creating-a-secret-token), but then `secret`/`payload` in the field names of their [example](https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries#testing-the-webhook-payload-validation). If you look at the SDK, it is [called a `secretToken`](https://pkg.go.dev/github.com/google/go-github/v72/github#ValidateSignature) 🤷.

---

~Unrelated, but worth noting now that I looked at the SDK: we may be interested in using [`ValidatePayload`](https://pkg.go.dev/github.com/google/go-github/v72/github#ValidatePayload).~ This would have the same reader problems.

internal/webhook/external/github.go
hiddeco: I would leave it alone for now until someone complains about it being too small. Lowering it would only increase the chance of someone complaining, and I think we are sufficiently guarded by the URL being unguessable for most scenarios to worry about 2MB being a problem.

internal/webhook/external/receiver.go
hiddeco: After this explanation, I can live with it as I can't come up with a less awkward alternative myself :-).

internal/webhook/external/github.go
krancour: Let's hear it for consistency.

Would you agree that in the absence of any consistency, aligning with the field name on the registration form would be the most sensible?

hiddeco: I am fine with this.

internal/webhook/external/receiver.go
krancour: Actually, I did the registry.

internal/webhook/external/receiver.go
```diff
@@ -0,0 +1,185 @@
...
+) (WebhookReceiver, error) {
+	// Pick an appropriate WebhookReceiver implementation based on the
+	// configuration provided.
+	var receiver WebhookReceiver
+	for _, registration := range webhookReceiverRegistry {
+		if registration.predicate(cfg) {
+			receiver = registration.factory(c, project, cfg)
+			break
+		}
+	}
```
hiddeco: Nit: feels like if `webhookReceiverRegistry` had its own type, this could be moved to a method on that type.

internal/webhook/external/registry.go
```diff
@@ -0,0 +1,53 @@
...
+// instantiate.
+var webhookReceiverRegistry = map[string]webhookReceiverRegistration{}
+
+// registerWebhookReceiver is invoked once for each implementation of
+// WebhookReceiver upon package initialization to associate a
+// webhookReceiverPredicate with a webhookReceiverFactory.
+func registerWebhookReceiver(
+	receiverType string,
+	predicate webhookReceiverPredicate,
+	factory webhookReceiverFactory,
```
hiddeco: Nit: any reason for this to not just accept a `webhookReceiverRegistration`?

internal/webhook/external/receiver.go
krancour: I can get down with that.

internal/webhook/external/registry.go
krancour: We pass registration structs directly in other places where we've used this pattern and I almost emulated it, but then convinced myself that this was actually tidier. It's a tiny bit less repetitious for each caller and has the benefit that you cannot accidentally pass a registration that's missing one of these two requirements elements. Is there a compelling reason I overlooked for the way it's been done elsewhere?

hiddeco: I think the struct nicely incapsulates the fields, reducing the number of arguments. While the validation can stay put to catch any missing fields.

krancour: Eh. I forgot the validation those weren't nil was there. You're right that we certainly don't need both.

I'll change this.

internal/webhook/external/receiver.go
krancour: I cleaned up the registry in eb321cbcdaac2c6e346b1310a50847c5194d5513, making `register()` a method of the registry.

I could not turn `NewReceiver()` into a method. It's called from outside the package (by ProjectConfig and ClusterConfig reconcilers), so making that function a method of the registry would require exposing the registry outside the package.

internal/webhook/external/registry.go
krancour: Fixed in eb321cbcdaac2c6e346b1310a50847c5194d5513

PR #4327, opened 2025-06-04
feat: add ghcr webhook support to Kargo

nitishfy:
Part of https://github.com/akuity/kargo/issues/4150

The following PR adds the webhook support for GHCR.

### DIFF
--- a/internal/webhook/external/github_test.go
+++ b/internal/webhook/external/github_test.go
@@ -87,7 +87,7 @@ func TestGithubHandler(t *testing.T) {
 					Build()
 			},
 			req: func() *http.Request {
-				b := newBody()
+				b := newGitHubEventBody("push")
 				req := httptest.NewRequest(http.MethodPost, url, b)
 				req.Header.Set("Content-Type", "application/json")
 				req.Header.Set("X-Hub-Signature-256", sign(t, testSecret, b.Bytes()))
@@ -159,7 +159,7 @@ func TestGithubHandler(t *testing.T) {
 					Build()
 			},
 			req: func() *http.Request {
-				b := newBody()
+				b := newGitHubEventBody("push")
 				req := httptest.NewRequest(http.MethodPost, url, b)
 				req.Header.Set("Content-Type", "application/json")
 				req.Header.Set("X-Hub-Signature-256", sign(t, "fakesecret", b.Bytes()))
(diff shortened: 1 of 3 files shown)


internal/webhook/external/github_test.go
```diff
@@ -674,22 +761,100 @@ func sign(t *testing.T, s string, b []byte) string {
...
 	"head_commit": {
-	  "id": "f12cd167152d80c0a2e28cb45e827c6311bba910"
+		"id": "f12cd167152d80c0a2e28cb45e827c6311bba910"
 	}
-  }	
+}
 `))
+	case "package":
+		return bytes.NewBuffer([]byte(`
+{
```
fuskovic, line 786: The sample event payload has a lot of unnecessary fields. Since this is just for testing, reduce it down to only the field you need. You can do the same for the push event too. I originally included some fields that I thought we would need but then no longer needed.

Also, I think if we use `bytes.NewBufferString` we can get rid of the type-cast.

nitishfy: I've only put the required fields from the [docs itself](https://docs.github.com/en/webhooks/webhook-events-and-payloads#package). I think there are few fields that are still missing.

fuskovic: You don't need to add all those fields. Just the ones our handler cares about.

fuskovic: For example, you only need the `package` object, and in that object you only need the `name` and `namespace` fields.

nitishfy: I've removed the extra fields. I thought anyone reading the payload would get to know that these are the required fields that a payload should have. However, they can refer to the docs for it.  Thanks!

fuskovic: The payload is sent to us by Github so it will always have the required fields. No problem!

internal/webhook/external/github_test.go
```diff
@@ -674,22 +761,43 @@ func sign(t *testing.T, s string, b []byte) string {
 	)
 }
 
-func newBody() *bytes.Buffer {
-	return bytes.NewBuffer([]byte(`
+func newGitHubEventBody(eventType string) *bytes.Buffer {
```
fuskovic, line 764: For the `push` event, looks like we just need the `repository` object for `html_url`.

For the `package` event, looks like we just need the `package` object for  `name` and `namespace`.

With that said I think we can greatly simplify this if we just narrow it down to that.

@fykaa and I are doing the same on our PR's as well.

I think you can get rid of the `default` case too but if you decide to keep it we can at least use `bytes.NewBufferString`

krancour:

krancour:
Closing in favor of #4353

PR #4328, opened 2025-06-04
feat:gitlab webhook receiver support

fuskovic:
Related to https://github.com/akuity/kargo/issues/4150


krancour:

internal/webhook/external/gitlab.go
```diff
@@ -0,0 +1,143 @@
...
+			xhttp.WriteErrorJSON(w,
+				xhttp.Error(
+					fmt.Errorf("failed to read request body: %w", err),
+					http.StatusRequestEntityTooLarge,
+				),
+			)
+			return
+		}
+
+		e, err := gl.ParseWebhook(eventType, b)
```
nitishfy, line 90:
```suggestion
		e, err := gl.ParseWebhook(eventType, b)
```

I think we should write the function for signature validation before parsing. Let's not skip it.

krancour: @nitishfy what signature are you talking about?

I see nothing indicating that inbound requests from GitLab are signed in any way.

Authentication is done by checking for an expected shared secret in the `X-Gitlab-Token` header, which has already been done above.

https://docs.gitlab.com/user/project/integrations/webhooks/

krancour:
Closing in favor of #4332, which is the same logic, but compatible with recent refactoring.

PR #4329, opened 2025-06-04
feat: default to common api ingress for webhook

gdsoumya:
This PR exposes the external webhook service through the kargo-api ingress by default, but allows for a separate ingress if `externalWebhooksServer.ingress.enabled` is set to 'true'.


charts/kargo/templates/external-webhooks-server/deployment.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.webhooksServer.enabled }}
+{{- if .Values.externalWebhooksServer.enabled }}
```
gdsoumya, line 1: there was a bug that used admission webhook server value instead

PR #4332, opened 2025-06-05
feat: add gitlab webhook receiver support

krancour:
Supersedes #4328

Like #4328, but refactored for compatibility with #4326.

All credit to @fuskovic for actually working out the details of integrating with GitLab. I only adapted it.


internal/webhook/external/gitlab.go
```diff
@@ -0,0 +1,175 @@
...
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/git"
+	xhttp "github.com/akuity/kargo/internal/http"
+	"github.com/akuity/kargo/internal/io"
+	"github.com/akuity/kargo/internal/logging"
+)
+
+const (
+	gitlab                    = "gitlab"
+	gitLabSecretDataKey       = "gitlab-secret"
```
krancour: I'm questioning using `gitlab-secret` as the key. We should take a look at docs and the actual webhook registration form and hopefully (unlike in the case of GitHub) they agree/use consistent language for what this _thing_ is called -- "token," "secret token," "shared secret," etc. and we should align this key with that. If the language is inconsistent across the registration form and the docs, we should do as we did with GitHub and align with the form, as this will be most intuitive / easily understood for the majority of users.

api/v1alpha1/project_config_types.go
```diff
@@ -115,8 +120,26 @@ type GitHubWebhookReceiverConfig struct {
...
+// with GitLab payloads.
+type GitLabWebhookReceiverConfig struct {
+	// SecretRef contains a reference to a Secret. For Project-scoped webhook
+	// receivers, the referenced Secret must be in the same namespace as the
+	// ProjectConfig.
+	//
+	// For cluster-scoped webhook receivers, the referenced Secret must be in the
+	// designated "cluster Secrets" namespace.
+	//
+	// The secret is expected to contain a `gitlab-secret` key containing the
```
krancour: See https://github.com/akuity/kargo/pull/4332/files#r2132639128

internal/webhook/external/gitlab.go
krancour: @fuskovic this is for you to research and make a recommendation on.

api/v1alpha1/project_config_types.go
```diff
@@ -94,10 +94,15 @@ type WebhookReceiverConfig struct {
...
-	// +kubebuilder:validation:Required
+	// TODO(fuskovic): Make this mutually exclusive with configs for other
+	// platforms.
 	GitHub *GitHubWebhookReceiverConfig `json:"github,omitempty" protobuf:"bytes,2,opt,name=github"`
+	// GitLab contains the configuration for a webhook receiver that is compatible
+	// with GitLab payloads.
+	//
+	// TODO(fuskovic): Make this mutually exclusive with configs for other
+	// platforms.
+	GitLab *GitHubWebhookReceiverConfig `json:"gitlab,omitempty" protobuf:"bytes,3,opt,name=gitlab"`
```
krancour: @fuskovic can you work out the mutual exclusivity of these two field, ideally using CEL?

internal/webhook/external/gitlab.go
fuskovic: Gitlab refers to it as a `secret-token` in both their docs and the creation form...so I'll change it to match that.

api/v1alpha1/project_config_types.go
```diff
@@ -102,7 +102,7 @@ type WebhookReceiverConfig struct {
 	//
 	// TODO(fuskovic): Make this mutually exclusive with configs for other
 	// platforms.
-	GitLab *GitHubWebhookReceiverConfig `json:"gitlab,omitempty" protobuf:"bytes,3,opt,name=gitlab"`
+	GitLab *GitLabWebhookReceiverConfig `json:"gitlab,omitempty" protobuf:"bytes,3,opt,name=gitlab"`
```
krancour, line 103: 🤦‍♂️ Nice catch. That was a botched resolution of a merge conflict.

api/v1alpha1/project_config_types.go
```diff
@@ -134,7 +134,7 @@ type GitLabWebhookReceiverConfig struct {
 	// For cluster-scoped webhook receivers, the referenced Secret must be in the
 	// designated "cluster Secrets" namespace.
 	//
-	// The secret is expected to contain a `gitlab-secret` key containing the
+	// The secret is expected to contain a `secret-token` key containing the
```
krancour, line 135: 👍 Appears consistent with the form:

Screenshot 2025-06-09 at 3 12 59 PM

api/v1alpha1/project_config_types.go
fuskovic: I've pushed this. Here's a screenshot for the manual test.

Screenshot 2025-06-10 at 10 45 39 AM

fuskovic (approved):

api/v1alpha1/project_config_types.go
```diff
@@ -88,16 +88,19 @@ type PromotionPolicy struct {
 
 // WebhookReceiverConfig describes the configuration for a single webhook
 // receiver.
+//
+// +kubebuilder:validation:XValidation:message="WebhookReceiverConfig must have exactly one of github or gitlab set",rule="has(self.github) ? !has(self.gitlab) : has(self.gitlab)"
```
krancour, line 92: @fuskovic I sent you a message offline. I can see more clearly now why you were reluctant about this line. It works great for two fields, but the expression becomes extremely cumbersome for three or more.

Since it's working for now, we can merge this and we'll just have to fall back on using a validating webhook for this when one of the other in-flight PRs adds a third field.

krancour: You can do the honors.

PR #4334, opened 2025-06-05
fix(chart): external webhook server base url

gdsoumya:
This PR fixes a bug introduced in #4329 that calculated the wrong base url for external webhooks server


hiddeco (commented):
I think that it could potentially be better to rewrite the functions into something like:

```

{{/*
Helper function to determine if TLS should be used for a service.
*/}}
{{- define "kargo.service.useTLS" -}}
{{- $service := . -}}
{{- or (and $service.ingress.enabled $service.ingress.tls.enabled) (and (not $service.ingress.enabled) $service.tls.enabled) $service.tls.terminatedUpstream -}}
{{- end -}}

{{/*
Helper function to generate base URL for a service.
*/}}
{{- define "kargo.service.baseURL" -}}
{{- $service := . -}}
{{- if include "kargo.service.useTLS" $service -}}
{{- printf "https://%s" $service.host -}}
{{- else -}}
{{- printf "http://%s" $service.host -}}
{{- end -}}
{{- end -}}

{{/*
Generate the base URL for the API service.
*/}}
{{- define "kargo.api.baseURL" -}}
{{- include "kargo.service.baseURL" .Values.api -}}
{{- end -}}

{{/*
Generate the base URL for the external webhook server.
*/}}
{{- define "kargo.externalWebhooksServer.baseURL" -}}
{{- $webhookServer := .Values.externalWebhooksServer -}}
{{- if include "kargo.service.useTLS" $webhookServer -}}
{{- printf "https://%s" $webhookServer.host -}}
{{- else if and (not $webhookServer.ingress.enabled) (not $webhookServer.tls.enabled) (not $webhookServer.tls.terminatedUpstream) -}}
{{- include "kargo.api.baseURL" . -}}
{{- else -}}
{{- printf "http://%s" $webhookServer.host -}}
{{- end -}}
{{- end -}}
```

hiddeco (approved):

PR #4338, opened 2025-06-05
feat: github enterprise webhook receiver support

fykaa:
part of #4150 

built on top of #4326 (that needs to be merged before this.)


krancour:
This is well implemented, but I am somewhat questioning the need for this feature.

The URLs are hard to guess and a shared secret is required to produce a valid signature.

Anyone who has that information and intends to maliciously spoof webhooks originating from GitHub Enterprise can certainly set the `X-GitHub-Enterprise-Host` correctly.

i.e. I'm not sure what value this adds.

cc @hiddeco for thoughts.

hiddeco:
I do not think the intent of the header is for validating the origin of the request (nor does it appear [to be do documented as such](https://docs.github.com/en/enterprise-server@3.16/webhooks/webhook-events-and-payloads#delivery-headers)), I rather think it is to allow the receiver of the request to determine how they should talk back to the GitHub API when they receive a request / webhook payload.

Given this, I also doubt the usefulness, as the HMAC / signature header should be sufficient.

krancour:
@fykaa it seems as if nothing special is required to make the existing GitHub webhook support work with GitHub Enterprise. I'm closing this and you can update #4150 to indicate there's nothing to do here.

PR #3955, opened 2025-04-29
feat(controller): add targetFreight to Promotion Variables

johnpekcan:
Addresses https://github.com/akuity/kargo/issues/3915

```yaml
spec:
  steps:
    - as: argocd-update
      config:
        apps:
          - name: ${{ vars.argoAppName }}
            namespace: ${{ vars.argoNamespace }}
            sources:
              - desiredRevision: ${{ commitFrom(vars.gitRepo, warehouse(ctx.targetFreight.origin.name)).ID }}
                helm:
                  images:
                    - key: my-app.imageTag
                      value: ${{ imageFrom(vars.artifactoryRepo, warehouse(ctx.targetFreight.origin.name)).Tag }}
                repoURL: ${{ vars.gitRepo }}
                updateTargetRevision: true
      uses: argocd-update
```

Now, we can promote "hotfixes" or "releases" from a separate warehouses 🎉

PR #4345, opened 2025-06-06
feat: ability to access freight metadata in template and add target freight to step ctx

gdsoumya:
This PR:

1. Adds ability to access freight metadata as part of step templates
2. Passed target freight info to step ctx

### DIFF
--- a/internal/expressions/function/functions.go
+++ b/internal/expressions/function/functions.go
@@ -190,7 +190,7 @@ func freightMetadata(
 			Name:      freightRefName,
 		}, &freightData); err != nil {
 			if kubeerr.IsNotFound(err) {
-				return nil, fmt.Errorf("freight %s not found in project %s", freightRefName, project)
+				return nil, nil
 			}
 			return nil, fmt.Errorf("failed to get freight %s: %w", freightRefName, err)
 		}
@@ -201,7 +201,7 @@ func freightMetadata(
 			return nil, fmt.Errorf("failed to get metadata %s from freight %s: %w", key, freightRefName, err)
 		}
 		if !found {
-			return nil, fmt.Errorf("metadata %s not found in freight %s", key, freightRefName)
+			return nil, nil
 		}
 
 		return data, nil
--- a/internal/expressions/function/functions_test.go
+++ b/internal/expressions/function/functions_test.go
@@ -1191,7 +1191,7 @@ func Test_freightMetadata(t *testing.T) {
 			objects: []client.Object{testFreight},
 			args:    []any{testFreightName, "non-existent-key"},
 			assertions: func(t *testing.T, result any, err error) {
-				assert.ErrorContains(t, err, "metadata non-existent-key not found in freight")
+				assert.NoError(t, err)
 				assert.Nil(t, result)
 			},
 		},
@@ -1200,7 +1200,7 @@ func Test_freightMetadata(t *testing.T) {
 			objects: []client.Object{}, // No freight objects
 			args:    []any{testFreightName, "deployment-config"},
 			assertions: func(t *testing.T, result any, err error) {
-				assert.ErrorContains(t, err, "freight fake-freight not found in project")
+				assert.NoError(t, err)
 				assert.Nil(t, result)
 			},
 		},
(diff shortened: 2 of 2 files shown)


krancour:
I think the `freightMetadata()` would be more generally useful if it took _two_ arguments.

To get metadata from the Promotion's "target freight", you could use: 

`freightMetadata(ctx.TargetFreightRef.Name, 'some-key')`.

But this could equally be used for accessing metadata from _any_ piece of Freight.

An added benefit is this would rectify a nuanced problem with how this PR currently does things. Technically, when evaluated in the context of a Promotion `freightMetadata()` is retrieving metadata from the target Freight (the Freight actively being promoted), but when evaluated in the context of a verification/analysis, `freightMetadata()` is retrieving metadata from the last updated piece of Freight in the Stage's current FreightCollection...

Yes... these (at least for now) will always be the same thing, but unless we were to take great care in explaining the nuanced difference to users, we will create the impression that verification/analysis has access to a Promotion's target Freight, which reinforces a false narrative already subscribed to by many users, which we are trying to combat -- that verification is part of the promotion process. It isn't. It's a point-in-time verification of a Stage's current state.

So, I think we could kill two birds with one stone by accepting two arguments. `freightMetadata()` would be more flexible _and_ we wouldn't be reinforcing what is already a common misunderstanding. As a bonus, it's a lot less new code to implement and maintain. wdyt?

internal/expressions/function/functions.go
```diff
@@ -143,6 +144,87 @@ func ChartFrom(
...
+		exists := false
+		for _, ref := range freightRefs {
+			if ref.Name == freightRefName {
+				exists = true
+				break
+			}
+		}
+		if !exists {
+			return nil, fmt.Errorf("freight reference %s not found for stage", freightRefName)
+		}
```
krancour: I would suggest blindly going after whatever Freight was specified as that makes this function not dependent on a collection of Freight refs, and thereby usable in other contexts. (i.e. Could still be used in verifications.)

gdsoumya: Updated!

internal/expressions/function/functions.go
```diff
@@ -143,6 +144,70 @@ func ChartFrom(
...
+
+		freightData := kargoapi.Freight{}
+
+		if err := c.Get(ctx, client.ObjectKey{
+			Namespace: project,
+			Name:      freightRefName,
+		}, &freightData); err != nil {
+			if kubeerr.IsNotFound(err) {
+				return nil, fmt.Errorf("freight %s not found in project %s", freightRefName, project)
```
krancour: It might be better to return nil instead of an error. Handling errors within an expression is hard, but handling an unexpected nil using [nil-coalescing](https://expr-lang.org/docs/language-definition#nil-coalescing) is easier.

cc @hiddeco for confirmation of this.

pkg/promotion/promotion.go
```diff
@@ -163,6 +163,8 @@ type StepContext struct {
 	// responsible for finding them and furnishing them directly to each
 	// StepRunner.
 	Freight kargoapi.FreightCollection
+	// TargetFreightRef is the actual Freight that triggered this Promotion.
+	TargetFreightRef kargoapi.FreightReference
```
krancour, line 167: I think we don't need this, as it turns out.

https://github.com/akuity/kargo/issues/3955 already seems to have this covered as of a few days ago.

internal/expressions/function/functions.go
gdsoumya: @krancour because we want a specific key from the metadata we need to error out if the freight itself is not found because we also return error if metadata key is not found. Imo in this case the promo should fail instead of continuing because the expected data is not there.

krancour: Ideally, I think, missing Freight and missing key _both_ would return nil.

We're used to doing `if err != nil` all over the place in our Go code, but handling errors is much harder in an expression.

If someone were savvy enough to use nil coalescing, they could deal with the nil result elegantly.

If someone _doesn't_ do that, there will be a nil pointer dereference, which will panic, and I believe we have everything in place at this point that that panic will be recovered and the _step_ (not the promo) will fail.

(Step failure no longer guarantees promotion failure. We added the ability in 1.5 to conditionally execute steps _in the event that a previous one has failed_. This allows for writing steps with error-handling logic.)

Bottom line is returning a nil, _will_ make things fail in the way you believe they ought to.

pkg/promotion/promotion.go
krancour: Discussed offline and this should stay.

internal/expressions/function/functions.go
gdsoumya: Makes sense, I'll change it to return nil

gdsoumya: Updated

Issue #4266, opened 2025-05-28
docs: improve README for clarity and first Impressions


fykaa:
### summary

[the current README](https://github.com/akuity/kargo/blob/main/README.md) could be updated to better guide new users/contributors. since this is the first page people see, a few small improvements would help with clarity, structure and onboarding

this issue tracks a few suggestions to clean up and improve the content without making big changes

### To-do list

* [x] add a short **one-line summary** to explain what Kargo is in plain language.
* [ ] update the kargo UI snapshot to the most updated one from v1.5.1
* [ ] clean up the top of the README:
  * [ ] keep only essential badges (like CI, docs, discord), remove netlify codecov?
  * [ ] consider moving the second logo and some badges lower.
* [ ] Improve the **Getting started** section by briefly describing what to expect before linking to the Quickstart

  * [ ] I think we can update the youtube video to a more updated one? say one from here https://docs.kargo.io/resources/#kargo-ga-webinar-
* [ ] remove the line about the docs being “very new”, imo we are GA now, and docs is in great form

feel free to add more...

krancour:
>  keep only essential badges (like CI, docs, discord), remove netlify codecov?

> consider moving the second logo and some badges lower.

Please leave the badges alone.

fykaa:
netlify badge currently shows cancelled, I am not sure if we need that to be on the readme.

krancour:
> netlify badge currently shows cancelled, I am not sure if we need that to be on the readme.

That was a transient condition and shouldn't be seen as impetus to not have the badge.

krancour:
The README was recently updated in #4998

Issue #4114, opened 2025-05-15
docs: update all for new ui


krancour:
Similar to #4111, but (only slightly) less urgent and broader in scope...

The UI has been significantly overhauled, but the docs haven't caught up to it yet.

Besides updating screenshots, we should also make sure all directions presented to the reader are accurate for the revised UI.

PR #4350, opened 2025-06-06
chore: update readme (kargo dashboard)

fykaa:
part of https://github.com/akuity/kargo/issues/4266


krancour:
I'm not sure I agree with the updated image being placed in `ui/public`. Those are assets belonging to the UI itself.

`docs/static/img/` might, _temporarily_, be a better home.

My guess is that when you complete work on #4114 this image or a similar one will be used somewhere and, as we tend to do, will live in the appropriate `docs/docs/<some path>/img/` directory and the `README.md` can be updated to point to it in its new home.

fykaa:
> My guess is that when you complete work on https://github.com/akuity/kargo/issues/4114 this image or a similar one will be used somewhere and, as we tend to do, will live in the appropriate docs/docs/<some path>/img/ directory and the README.md can be updated to point to it in its new home

I'll update the README to point to it if I find an image similar to this one in that PR.

krancour (approved):

Issue #3999, opened 2025-05-06
Standardized Image URL normalization


fuskovic:
# Proposed Feature

There are a lot of places where we're not doing any kind of image URL normalization -- for instance, [here](https://github.com/akuity/kargo/blob/714a640b50937bff5a966ef84ffac79ee9e3ba97/internal/credentials/kubernetes/database.go#L175-L180). We probably should start.

# Motivation

For context see https://github.com/akuity/kargo/pull/3969#discussion_r2076261351

krancour:
Adding some detail here, there's at least one spot where the chart URL normalization logic was re-used for image URL normalization -- which is safe, but awkward.

We're really good about consistently normalizing git repo URLs and chart URLs before any URL comparisons. We should do similar for image URLs, but should develop image-specific logic for doing so.

krancour:
#4351 took care of this.

PR #4351, opened 2025-06-07
feat: add docker hub webhook receiver

krancour:
Supersedes #4324 

This is @fykaa's logic from #4324 refactored for compatibility with #4326.


internal/webhook/external/dockerhub.go
```diff
@@ -0,0 +1,157 @@
...
+			)
+			return
+		}
+
+		// TODO(krancour): We should be normalizing this. We do not yet have any
+		// normalization logic for image URLs, and we've gotten away with it
+		// until now, but it's important here because, for example:
+		//   - docker.io/example/repo == example/repo
+		//   - docker.io/nginx == library/nginx == nginx
+		repoURL := payload.Repository.RepoName
```
krancour: @fykaa I'd like you to work this out please.

internal/webhook/external/dockerhub.go
```diff
@@ -71,6 +77,84 @@ func (d *dockerhubWebhookReceiver) getSecretValues(
...
+// canonical representation of a Docker Hub image reference is needed. Any reference
+// that cannot be normalized will return an error.
+//
+// Examples:
+//
+//	"nginx"                    -> "docker.io/library/nginx:latest"
+//	"user/repo:v1.0"           -> "docker.io/user/repo:v1.0"
+//	"docker.io/library/nginx"  -> "docker.io/library/nginx:latest"
+//	"nginx@sha256:..."         -> "docker.io/library/nginx@sha256:..."
+func normalizeDockerImageRef(ref string) (string, error) {
```
krancour: You will have better luck and much less code if instead of using regex for this, you use `name.ParseReference()` from the `github.com/google/go-containerregistry/pkg/name` package.

Quick standalone program to show you how it even accounts for all the idiosyncrasies for you:

```
package main

import (
	"fmt"
	"log"

	"github.com/google/go-containerregistry/pkg/name"
)

func main() {
	examples := []string{
		"nginx",
		"library/nginx",

		"docker.io/nginx",
		"docker.io/library/nginx",

		"index.docker.io/nginx",
		"index.docker.io/library/nginx",
	}
	for i, example := range examples {
		ref, err := name.ParseReference(example)
		if err != nil {
			log.Fatal(err)
		}
		fmt.Printf("%d. %s:\n", i+1, example)
		fmt.Printf("\tRegistry: %s\n", ref.Context().Registry.Name())
		fmt.Printf("\tRepository: %s\n", ref.Context().RepositoryStr())
	}
}
```

Some more notes:

1. I think when normalizing, I'd prefer that we go in the _other_ direction. i.e. If, after parsing, registry hostname is `index.docker.io`, drop it. If that's the hostname _and_ `ref.Context().RepositoryStr()` starts with `library/`, drop `library/`.

    I'd consider this preferable because, by way of example, `nginx` is a much more ubiquitous method of referencing the official nginx repo than any other equivalent.

1. There are other places in the code that will benefit from this logic, so you need to move this function to some more common package and you need to export it. My suggestion would be putting it in the `internal/image` package.

1. For the purposes of this PR, you need to apply this normalization in two places:

    i. When building the `WarehousesBySubscribedURLs()` index.

    i. Here in the receiver, before doing the list Warehouses operation that uses that index.

internal/image/docker.go
```diff
@@ -0,0 +1,88 @@
...
+// canonical representation of a Docker Hub image reference is needed. Any reference
+// that cannot be normalized will return an error.
+//
+// Examples:
+//
+//	"nginx"                    -> "docker.io/library/nginx:latest"
+//	"user/repo:v1.0"           -> "docker.io/user/repo:v1.0"
+//	"docker.io/library/nginx"  -> "docker.io/library/nginx:latest"
+//	"nginx@sha256:..."         -> "docker.io/library/nginx@sha256:..."
+func NormalizeRef(ref string) (string, error) {
```
krancour: This would be more consistent and more intuitively understood.

```suggestion
func NormalizeURL(repoURL string) (string, error) {
```

internal/webhook/external/dockerhub.go
krancour: Adding a fourth bit here:

4. The output of this should be the URL for a _repository_ (a "bucket" in a registry, if you will). It should never be the URL for a specific image. i.e. The output should never include a tag or digest as those reference specific images _within_ the repository and that is not what we want.

internal/image/docker.go
fykaa: this function should also address part of this issue: https://github.com/akuity/kargo/issues/3999

internal/webhook/external/dockerhub.go
fykaa: addressed in ebdf0453980b266dcded0bf9d8fd7c634373c22e 2fb73b7272e49466713bb10bcbc687bb76b012b7 21dd157e706378a0280da9d5c41df013d9d4b437

internal/image/docker.go
```diff
@@ -0,0 +1,36 @@
...
+
+	// Compose normalized reference (no tag or digest)
+	var sb strings.Builder
+	if reg != "" {
+		sb.WriteString(reg)
+		sb.WriteString("/")
+	}
+	sb.WriteString(repo)
+	return sb.String(), nil
+}
```
krancour: This can be streamlined a bit further.

```suggestion
// NormalizeURL normalizes image repository URLs. Notably, hostnames docker.io
// and index.docker.io, if present, are dropped. The optional /library prefix
// for official images from Docker Hub, if included, is also dropped. Valid,
// non-Docker Hub repository URLs will be returned unchanged.
//
// This is useful for the purposes of comparison and also in cases where a
// canonical representation of a repository URL is needed. Any URL that cannot
// be normalized will be returned as-is.
func NormalizeURL(repoURL string) (string) {
	parsed, err := name.ParseReference(repoURL, name.WeakValidation)
	if err != nil {
		return repoURL
	}

	reg := parsed.Context().Registry.Name()
	repo := parsed.Context().RepositoryStr()

	// For all images from Docker Hub, reg will be index.docker.io after parsing.
	if reg == "index.docker.io" {
		return strings.TrimPrefix(repo, "library/")		
	}

	return fmt.Sprintf("%s/%s", reg, repo)
}
```

Not returning an error and just returning the non-normalizable URL as-is is how we normalize elsewhere and it's safe because two occurrences of the same input that cannot be normalized for whatever reason will both be returned unchanged, and when compared, will be equal, just as they should be.

You'll have to adapt the unit tests and other calls to this function for the revised signature.

internal/image/docker_test.go
```diff
@@ -0,0 +1,39 @@
...
+		// Official images, index.docker.io registry
+		{"index.docker.io/library/redis:7", "redis"},
+		// Custom repo in docker.io
+		{"docker.io/myuser/myapp:dev", "myuser/myapp"},
+		// Other registries
+		{"gcr.io/myproj/app:tag", "gcr.io/myproj/app"},
+		{"quay.io/org/repo:tag", "quay.io/org/repo"},
+		// No tag/digest
+		{"docker.io/library/busybox", "busybox"},
+		{"gcr.io/myproj/app", "gcr.io/myproj/app"},
```
krancour: I would suggest cutting test cases involving tags or digests.

We both know that if you should happen to pass a full image URL to this function, what you get back is just the normalized _repo_ URL, but since that is definitely not how this function is meant to be used, I think explicitly testing that "off label" use will prompt confusion for people who don't understand the nuances well.

internal/image/docker.go
krancour, line 1: The filename `docker.go` feels too specific for the general-purpose functionality the file contains. Even if Docker Hub is the main reason we had to do this, this function is going to be blindly called on any/all repo URLs.

Same goes for `docker_test.go` as well.

I might suggest `normalization.go` and `normalization_test.go` as alternatives.

internal/webhook/external/dockerhub.go
```diff
@@ -0,0 +1,158 @@
...
+		if err = json.Unmarshal(body, &payload); err != nil {
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
+			)
+			return
+		}
+
+		// Normalize the repo name
+		repoURL, err := image.NormalizeURL(payload.Repository.RepoName)
```
krancour: Please add this to the Quay and GitHub receivers as well. (The GitHub receiver handles package events from ghcr.io.)

We know this function basically does nothing for non-Docker Hub URLs, but I don't want us to take it for granted that we won't someday expand the normalization logic.

krancour:
This is looking good @fykaa. Only some very minor cleanup remains.

internal/image/normalization_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	for _, tc := range tests {
+		got := NormalizeURL(tc.input)
+		require.Equal(t, tc.expected, got, "input: %s", tc.input)
+	}
+
+	// Invalid input: should return input as-is
+	input := "not a valid ref"
+	got := NormalizeURL(input)
+	require.Equal(t, input, got)
```
krancour: This can be a normal test case. It doesn't need to be separate from the others. i.e. The input/expected formula you're using for all the others works for this one, too.

internal/webhook/external/github.go
krancour, line 1: Please undo all the changes to this file.

When I requested that you update this, I didn't realize that #4353 wasn't merged yet.

The part that requires modification is not present here. (So the part you updated isn't correct.)

If you back out the changes to this file, resolving merge conflicts will be easier, and then the changes can be applied in the correct spot.

Please let me resolve the merge conflicts because there have been a number of gotchas as we merge all these receivers.

Just back out these changes and I'll handle the rest.

internal/image/normalization_test.go
```diff
@@ -0,0 +1,31 @@
...
+		{"docker.io/library/busybox", "busybox"},
+		// Official images, index.docker.io registry
+		{"index.docker.io/library/redis", "redis"},
+		// Custom repo in docker.io
+		{"docker.io/myuser/myapp", "myuser/myapp"},
+		// Other registries
+		{"gcr.io/myproj/app", "gcr.io/myproj/app"},
+		{"quay.io/org/repo", "quay.io/org/repo"},
+		// Invalid input: should return input as-is
+		{"not a valid ref", "not a valid ref"},
```
krancour:
```suggestion
		// Official images on Docker Hub
		{"index.docker.io/library/busybox", "busybox"},
		{"index.docker.io/busybox", "busybox"},
		{"docker.io/library/busybox", "busybox"},
		{"docker.io/busybox", "busybox"},
		{"library/busybox", "busybox"},
		{"busybox", "busybox"},

		// Other images on Docker Hub
		{"index.docker.io/example/repo", "example/repo"},
		{"docker.io/example/repo", "example/repo"},
		{"example/repo", "example/repo"},

		// Images from other registries
		{"ghcr.io/example/repo", "ghcr.io/example/repo"},
		{"quay.io/example/repo", "quay.io/example/repo"},

		// Input that cannot be normalized (invalid URL)
		{"invalid url", "invalid url"},
```

internal/image/normalization.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		return repoURL
+	}
+	reg := parsed.Context().Registry.Name()
+	repo := parsed.Context().RepositoryStr()
+	// For all images from Docker Hub, reg will be index.docker.io after parsing.
+	if reg == "index.docker.io" {
+		return strings.TrimPrefix(repo, "library/")
+	}
+	return fmt.Sprintf("%s/%s", reg, repo)
```
hiddeco: I believe another variation you can hypothetically run into is `registry-1.docker.io`, see: https://github.com/moby/moby/blob/v24.0.2/registry/config.go#L35-L41

krancour: Yes... that looks familiar. I'll add that to the test cases.

internal/image/normalization_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+func TestNormalizeURL(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected string
+	}{
+		// Official images on Docker Hub
+		{"index.docker.io/library/busybox", "busybox"},
+		{"index.docker.io/busybox", "busybox"},
+		{"docker.io/library/busybox", "busybox"},
```
krancour, line 19:
```suggestion
		{"registry-1.io/library/busybox", "busybox"},
		{"registry-1.io/busybox", "busybox"},
		{"docker.io/library/busybox", "busybox"},
```

internal/image/normalization_test.go
```diff
@@ -14,8 +14,8 @@ func TestNormalizeURL(t *testing.T) {
 		// Official images on Docker Hub
 		{"index.docker.io/library/busybox", "busybox"},
 		{"index.docker.io/busybox", "busybox"},
-		{"registry-1.io/library/busybox", "busybox"},
-		{"registry-1.io/busybox", "busybox"},
+		{"registry-1.docker.io/library/busybox", "busybox"},
+		{"registry-1.docker.io/busybox", "busybox"},
```
krancour, line 18: Whoops. Thank you for fixing that.

internal/image/normalization.go
```diff
@@ -22,8 +22,10 @@ func NormalizeURL(repoURL string) string {
 	}
 	reg := parsed.Context().Registry.Name()
 	repo := parsed.Context().RepositoryStr()
-	// For all images from Docker Hub, reg will be index.docker.io after parsing.
-	if reg == "index.docker.io" {
+	// For all images from Docker Hub, reg will be one of docker.io, index.docker.io,
```
krancour: The way I had it before was true. AFTER parsing, it's always index.docker.io.

fykaa: the tests were failing because the implementation only normalized Docker Hub images when the registry is exactly `index.docker.io` 

but, the parser does not convert `registry-1.docker.io` to `index.docker.io`. It only does this for `docker.io`

fykaa: after parsing
for most docker hub images, reg will be `index.docker.io`
for images that explicitly uses `registry-1.docker.io` reg will be `registry-1.docker.io`

internal/image/normalization.go
```diff
@@ -0,0 +1,31 @@
...
+// be normalized will be returned as-is.
+func NormalizeURL(repoURL string) string {
+	parsed, err := name.ParseReference(repoURL, name.WeakValidation)
+	if err != nil {
+		return repoURL
+	}
+	reg := parsed.Context().Registry.Name()
+	repo := parsed.Context().RepositoryStr()
+	// For all images from Docker Hub, reg will be "index.docker.io" after parsing,
+	// except when the input explicitly uses "registry-1.docker.io", which is preserved.
```
krancour:
```suggestion
	// For all images from Docker Hub, reg will be "index.docker.io" after
	// parsing, except when the input explicitly uses "registry-1.docker.io",
	// which is preserved.
```

internal/image/normalization.go
krancour: Ack. I'd seen `docker.io` always become `index.docker.io` and it was wrong of me to assume the same would be true for `registry-1.docker.io`. Thank you for this.

hiddeco (approved):

PR #4352, opened 2025-06-07
chore: add new github webhook receiver test case and minor cleanup

krancour:
While working on #4351, I realized #4326 missed the partial success test-case for refreshing Warehouses in response to push events.


krancour:
Closing because #4353 also contains similar fixes and I'd rather take those than deal with the merge conflict.

PR #4353, opened 2025-06-07
feat: add ghcr.io package event support to github webhook receiver

krancour:
Supersedes #4327

This adapts @nitishfy's updates to the GitHub webhook receiver to be compatible with the substantial refactoring that occurred in #4326 

This is built on top of changes in #4352 and should be rebased after that is merged.


internal/webhook/external/github.go
```diff
@@ -162,39 +164,49 @@ func (g *githubWebhookReceiver) GetHandler() http.HandlerFunc {
...
-						"error": fmt.Sprintf("failed to refresh %d of %d warehouses",
-							result.failures,
-							result.successes+result.failures,
-						),
-					},
-				)
-				return
-			}
+			repoURL = git.NormalizeURL(e.GetRepo().GetHTMLURL())
+		case *gh.PackageEvent:
```
krancour: @nitishfy we should probably only respond to actions `published` and `updated`. `deleted` is another possible action for which I think there is nothing for us to do.

internal/webhook/external/github.go
```diff
@@ -162,39 +164,51 @@ func (g *githubWebhookReceiver) GetHandler() http.HandlerFunc {
...
-					},
-				)
-				return
-			}
+			repoURL = git.NormalizeURL(e.GetRepo().GetHTMLURL())
+		// nolint: lll
+		case *gh.PackageEvent:
+			// Construct the repoURL from the name and namespace fields of the
+			// payload.
+			//   https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry
```
krancour: @nitishfy I didn't remove this reference that was in your original changes in #4327, but I was slightly perplexed as to the value it's supposed to add here. It doesn't seem as if it relates to constructing the repoURL from the name and namespace fields.

Am I missing something? Would a link to https://docs.github.com/en/webhooks/webhook-events-and-payloads#package be more useful?

This actually gets me thinking that we should probably have a link to the most relevant bit of documentation for each supported event type, i.e. we should add for ping and push as well.

nitishfy: This was the actual url that should be added here - https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images

I changed the URL because lint was complaining about maximum characters in a line. From the above link, the docs says:

> Replace NAMESPACE with the name of the personal account or organization to which you want the image to be scoped.

internal/webhook/external/github.go
nitishfy: I think the `package` event is going to respond to the actions you mentioned provided you have specified the relevant permissions to your token. 

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic

internal/webhook/external/github.go
krancour: There's a way around the line length limit without resorting to changing URLs. Look at line 174.

> Replace NAMESPACE with the name of the personal account or organization to which you want the image to be scoped.

Ok... but that page has nothing at all to do with webhooks. Those are only instructions regarding a placeholder used in the previous line, which was:

> docker push ghcr.io/NAMESPACE/IMAGE_NAME:latest

This does not speak to a description of the fields in the payload.

internal/webhook/external/github.go
krancour: `package` is the event type. The payload contains an `action` field. We should be looking at that and only responding to relevant ones.

internal/webhook/external/github.go
nitishfy: > Ok... but that page has nothing at all to do with webhooks. Those are only instructions regarding a placeholder used in the previous line, which was:

Even though the page has nothing to do with the webhooks, it gives us an idea of the basic structure for how the URL is going to look like with those placeholders in place.

> This does not speak to a description of the fields in the payload.

I couldn't find any helpful docs that speaks about the description of the fields in the payload. However, I did notice the `package_url` field under `package files` initally too while writing the logic and thought to strip it down:

`"package_url": "ghcr.io/nitishfy/guestbook:v0.2.0"`

I still think our way of constructing the Repo URL is correct provided this is the only way that makes a URL valid.

internal/webhook/external/github.go
nitishfy: I didn't understand what you mean by only responding to actions `published` and `updated`. We are already responding to these actions.

internal/webhook/external/github.go
krancour: I didn't say this was an incorrect way to piece the URL together. I said this bit of documentation does not seem to be an appropriate reference.

krancour: Actually... I take it back... it's possible that this is not the right way. It needs more research.

Does GitHub Enterprise or GitHub Enterprise Cloud include a registry? If it does, those URLs will not start with `ghcr.io`.

This is similar to this [other comment](https://github.com/akuity/kargo/pull/4354#discussion_r2134148878) on the Bitbucket PR.

nitishfy: Reading the docs [here](https://docs.github.com/ja/enterprise-cloud@latest/packages/working-with-a-github-packages-registry/working-with-the-container-registry#container-registry-%E3%81%AE-url)

> `octocorp.ghe.com` If you access GitHub through a different domain, such as ghcr.io, https://containers.SUBDOMAIN.ghe.comreplace "ghcr.io" with , where SUBDOMAINis a unique subdomain for your enterprise.

nitishfy: I think there are two cases here:

1. When using the GitHub Enterprise Cloud (Github.com), URL will be `ghcr.io/<organization>/<image>:<tag>`
2. If an organization uses a self-hosted GitHub Enterprise Server instance, the registry URL does not start with ghcr.io. Instead, it uses your enterprise's domain in the following format `https://containers.<subdomain>.<your-enterprise-domain>/<organization>/<image>:<tag>`

I think when a user is using the enterprise server instance, there's presence of `enterprise` field in the webhook payload. Also,
- If the URLs (such as `repository.url`, `organization.url`, etc.) point to `github.com`, the event is from GitHub Enterprise Cloud.
- If the URLs point to your organization's custom domain (e.g., github.mycompany.com), the event is from GitHub Enterprise Server.

nitishfy: For both the enterprise cloud and enterprise server, there's the `enterprise` field in the webhook payload object for `package` event. If this is present, it simply means the user is using the enterprise version else the free/pro version. Also the delivery headers in this case contain `X-GitHub-Enterprise-Host` header.

nitishfy: This makes me think if we should have added the support for GitHub enterprise webhook configuration.

krancour: Up to this point, nothing special needed to be done for GitHub enterprise. The push events have a `repository.clone_url` field which is a full URL for the repository. We didn't need to do anything special to piece it together ourselves.

It is only (so far) for these package events where _lack_ of a full repo URL anywhere in the payload (as far as I have seen thus far) forces us to piece the URL together ourselves, which forces us to account for Enterprise.

krancour: At any rate, this PR is assigned to you now. Please work out and implement the correct logic for this.

nitishfy: How about now? I pushed changes and followed this logic:

- Fetch the repository URL 
- Check if the hostname is `github.com`
- If it exists, use `ghcr.io` as the registry; otherwise, construct the new repo URL.

The only thing I'm concerned about here is if it is possible for a user who is using an enterprise server to keep using github.com instead of using their own custom domain.

internal/webhook/external/github.go
krancour: @nitishfy but we _don't_ need to handle `deleted` actions.

nitishfy: I tested the existing code for the `deleted` event. Surprisingly, the webhook didn't even trigger when the package was deleted. The controller showed the following logs:
```
time="2025-06-10T17:58:09Z" level=error msg="Reconciler error" Warehouse="{guestbook kargo-helm}" controller=warehouse controllerGroup=kargo.akuity.io controllerKind=Warehouse error="error discovering artifacts: error discovering images: error discovering latest images \"ghcr.io/nitishfy/guestbook\": error discovering newest applicable images \"ghcr.io/nitishfy/guestbook\": error listing tags: error listing tags for repo URL ghcr.io/nitishfy/guestbook: GET https://ghcr.io/token?scope=repository%3Anitishfy%2Fguestbook%3Apull&service=ghcr.io: DENIED: requested access to the resource is denied" name=guestbook namespace=kargo-helm reconcileID="\"cc99cf50-16b2-4199-b712-be3cfca766c2\""
```

Having said that, I don't think we need to do anything specifically to handle the `delete` action.

internal/webhook/external/github.go
```diff
@@ -153,7 +157,41 @@ func (g *githubWebhookReceiver) GetHandler() http.HandlerFunc {
...
+			switch pkg.GetPackageType() {
+			// These are the only types of packages we care about.
+			case ghcrPackageTypeContainer, ghcrPackageTypeDocker:
+			default:
+				xhttp.WriteResponseJSON(w, http.StatusOK, nil)
+				return
+			}
+			var ref name.Reference
+			if ref, err = name.ParseReference(
+				*pkg.GetPackageVersion().PackageURL,
```
fuskovic: If we use the `GetX` methods that Github provides we don't have to dereference + we get the nil check for free

```suggestion
				pkg.GetPackageVersion().GetPackageURL(),
```

krancour: Nice catch @fuskovic. Agree.

PR #4339, opened 2025-06-05
feat: add bitbucket webhook support

nitishfy:
Part of https://github.com/akuity/kargo/issues/4150

PR #4354, opened 2025-06-07
feat: add bitbucket webhook receiver

krancour:
Supersedes #4339 

This adapts @nitishfy's logic (with some changes) for compatibility with the substantial refactoring that occurred in #4326


internal/webhook/external/bitbucket.go
```diff
@@ -0,0 +1,188 @@
...
+		payload := bb.RepoPushPayload{}
+		if err = json.Unmarshal(body, &payload); err != nil {
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
+			)
+			return
+		}
+
+		repoURL := payload.Repository.Links.HTML.Href
```
krancour, line 181: @nitishfy this is definitely not the correct field to be retrieving the repo URL from.

It might _happen_ to work with `git clone`, but officially, this is a link the repository's _home page_.

Looking at docs, I actually cannot find a better place to get this info from. What I _can_ find is `full_name`, which for a repo such as `https://bitbucket.org/example/repo`, is `example/repo`...

It's tempting to prefix the `full_name` with `https://bitbucket.org`, except that won't work out correctly for webhooks originating from Bitbucket Datacenter.

I'd like you to please figure out where we can either get the repo URL directly _or_ where we can get the host/domain name of the server the webhook originated from so that we can construct the repo URL properly. Hopefully I'm just overlooking something in the docs.

If you can't find either of the above, then circle back and we'll discuss what the next best option is.

internal/webhook/external/bitbucket.go
```diff
@@ -0,0 +1,188 @@
+package external
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	bb "github.com/go-playground/webhooks/v6/bitbucket"
```
nitishfy:
```suggestion
	bb "github.com/go-playground/webhooks/v6/bitbucket"
```
@krancour I personally think this library offers us everything that we need to make the webhook receiver for BB work out. Even if there are some things left that the library doesn't provide (for eg. validating signatures), we can write that ourselves. Having said that, one good thing about this ilbrary is that it is easier to add new events for the webhook to listen for. They have the support for docker and gitlab too. I'm not sure if you want other PRs to use the same library but if we plan to do so, I think a significant part of our code can be reduced.

nitishfy: TL;DR 

Should we use the same library for Docker/Gitlab to maintain consistency? In future, if we want to add the support for new events, this will be easier to handle.

krancour: I actually have an unfavorable opinion of this library and I feel more inclined toward not using it at all.

nitishfy: But why is that?

krancour: I started to explain in great detail and it's not worth it. The high-level answer is there is no way of using this library that doesn't require us to duplicate some bit of effort that is already performed by `hook.Parse()` (which is why, compared to your original implementation in #4339, I'm not using it here). In the end, we're using precisely two things from the library: the `RepoPushEvent` constant and `RepoPushPayload{}` type. It's about five lines of code to make our own constant and our own struct that has only the fields we need. That's much cheaper and easier than accepting additional transitive dependencies in order to use a library that's giving us almost no value.

internal/webhook/external/bitbucket.go
krancour: Looking at actual payload from this event, I can now confidently say that there is no better place from which to get the repo's URL (one that would work with `git clone`) than `payload.Repository.Links.HTML.Href`.

The data we would wish is there just isn't.

As mentioned previously, we could piece the URL together by combining the protocol and hostname from `payload.Repository.Links.HTML.Href` with a `/` and the value of the `repository.full_name` field...

So far as I can tell, the result of that will be the same as if we'd just used the value of `payload.Repository.Links.HTML.Href` directly. There _might_ be some as-yet-unidentified edge cases involving BitBucket Data Center (self-hosted enterprise offering) where this doesn't hold up.

Reluctantly, I agree with getting the repo URL from `payload.Repository.Links.HTML.Href`, but since it looks so peculiar, some comments are in order.

internal/webhook/external/bitbucket.go
```diff
@@ -0,0 +1,196 @@
...
+		}{}
+		if err = json.Unmarshal(body, &payload); err != nil {
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
+			)
+			return
+		}
+
+		repoURL := payload.Repository.Links.HTML.Href
```
krancour, line 181:
```suggestion
		// Note: It may seem peculiar to obtain the repo URL from the payload's
		// repository.links.html.href field, however, there is no better option.
		//
		// A naive option would be combining https://bitbucket.org/ with the value
		// of the payload's repository.full_name field, but that does not hold up
		// for events originating from BitBucket Data Center, which will utilize a
		// custom domain name.
		//
		// A slightly better approach would be to parse the protocol and hostname
		// from the value of the payload's repository.links.html.href field and
		// combine that with the value of the payload's repository.full_name field,
		// however, in all (currently known) cases, that yields the same result as
		// simply using the value of the repository.links.html.href field directly.
		//
		// TODO(krancour): There are very likely some yet-to-be-identified edge
		// cases where this choice does not hold up.
		repoURL := payload.Repository.Links.HTML.Href
```

api/v1alpha1/project_config_types.go
```diff
@@ -114,9 +119,24 @@ type GitHubWebhookReceiverConfig struct {
...
+	// SecretRef contains a reference to a Secret. For Project-scoped webhook
+	// receivers, the referenced Secret must be in the same namespace as the
+	// ProjectConfig.
+	//
+	// For cluster-scoped webhook receivers, the referenced Secret must be in the
+	// designated "cluster Secrets" namespace.
 	//
-	// The value of the token key goes in the "Secret" field when registering a
-	// GitHub App or webhook in the GitHub UI.
+	// The Secret's data map is expected to contain a `bitbucket-secret` key whose
```
krancour: Screenshot 2025-06-10 at 11 04 44 PM

The registration form does call this field `secret`, but I don't believe prefixing that with `bitbucket-` adds any value here. In fact, it looks strange.

@nitishfy let's please change this to just `secret`.

Codegen will need to be re-run.

internal/webhook/external/bitbucket.go
nitishfy: I did the research on my end and I couldn't find a better field to retrieve the repository link other than from Href for Bitbucket cloud. Taking a look at [how Argo CD does that](https://github.com/argoproj/argo-cd/blob/7496ede7c286c84fe3f66154dc348376b7cbe74d/util/webhook/webhook.go#L198), they also fetch the link from the same field. It looks like the payload has a cloned URL fied for the bitbucket server. Note that I didn't test it but the following pieces of code gave me an idea:

- https://github.com/argoproj/argo-cd/blob/7496ede7c286c84fe3f66154dc348376b7cbe74d/util/webhook/webhook.go#L254
- https://github.com/gfleury/go-bitbucket-v1/blob/master/api_response.go#L71

Since the [official documentation](https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html#Eventpayload-repo-push) doesn't mention anything about it, I'm unsure if we should use this field. Having said that, I modifed our source code to include the clone field but it's clearly evident that the cloned field is never available for the Bitbucket cloud which was why it was showing as nil. For Bitbucket server, I didn't test it because I'm not on the enterprise plan.

krancour: > It looks like the payload has a cloned URL fied for the bitbucket server.

> [official documentation](https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html#Eventpayload-repo-push) doesn't mention anything about it

I cannot find anywhere where this is documented either. What has led you to believe that field might exist in push event payloads originating from Bitbucket Server / Data Center?

[The code you linked to](https://github.com/argoproj/argo-cd/blob/7496ede7c286c84fe3f66154dc348376b7cbe74d/util/webhook/webhook.go#L254) is not handling a push event, so I'd not consider this any evidence for the existence of that field.

> I'm unsure if we should use this field.

Yes. That's pretty much how I feel as well.

> Having said that, I modifed our source code to include the clone field

I can't see where you've done this. No such change seems to be in this PR. At any rate, that's ok. I am so dubious of that field's existence that I would simply not bother.

I think we can consider this dilemma resolved.

internal/webhook/external/bitbucket.go
```diff
@@ -0,0 +1,212 @@
...
+			)
+			return
+		}
+
+		// Note: It may seem peculiar to obtain the repo URL from the payload's
+		// repository.links.html.href field, however, there is no better option.
+		//
+		// A naive option would be combining https://bitbucket.org/ with the value
+		// of the payload's repository.full_name field, but that does not hold up
+		// for events originating from BitBucket Data Center, which will utilize a
```
krancour: This was my bad. They don't capitalize the second b.

```suggestion
		// for events originating from Bitbucket Data Center, which will utilize a
```

PR #4365, opened 2025-06-09
feat: support fallback to local cluster when searching for cred secrets

krancour:
@jessesuen @hiddeco this is a possible solution to the extremely outlying use case we discussed offline of wanting to fall back on searching for creds in the local cluster if none are discovered in the control plane's cluster.

The big, BIG caveat to this is that managing credential Secrets on each shard and the permissions that allow the controller on each shard to access those Secrets is completely left to the operator. There is no feature of Kargo's chart or Kargo itself to facilitate any of this.

### DIFF
--- a/cmd/controlplane/controller.go
+++ b/cmd/controlplane/controller.go
@@ -13,6 +13,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	libCluster "sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
@@ -42,9 +43,9 @@ import (
 type controllerOptions struct {
 	ShardName string
 
-	KubeConfig string
-	QPS        float32
-	Burst      int
+	ControlPlaneKubeConfig string
+	QPS                    float32
+	Burst                  int
 
 	ArgoCDEnabled       bool
 	ArgoCDKubeConfig    string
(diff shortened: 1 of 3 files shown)


cmd/controlplane/controller.go
```diff
@@ -247,7 +248,65 @@ func (o *controllerOptions) setupKargoManager(
...
+	}
+
+	scheme = runtime.NewScheme()
+	if err = corev1.AddToScheme(scheme); err != nil {
+		return nil, nil, stagesReconcilerCfg, fmt.Errorf(
+			"error adding Kubernetes core API to local client scheme: %w",
+			err,
+		)
+	}
+	localCluster, err := libCluster.New(
```
gdsoumya, line 274: should be check if LOCAL_CLUSTER_CREDS_FALLBACK != false before doing this?

gdsoumya: nvm I see the check is there before

cmd/controlplane/controller.go
```diff
@@ -247,7 +248,65 @@ func (o *controllerOptions) setupKargoManager(
...
+		os.GetEnv("LOCAL_CLUSTER_CREDS_FALLBACK", "false") != "true" {
+		return mgr, nil, stagesReconcilerCfg, err
+	}
+
+	// Build a separate client for the local cluster...
+	if restCfg, err =
+		kubernetes.GetRestConfig(ctx, o.ControlPlaneKubeConfig); err != nil {
+		return nil, nil, stagesReconcilerCfg,
+			fmt.Errorf("error loading REST config for local cluster client: %w", err)
+	}
```
gdsoumya: this says separate client for local but uses the same kubeconfig as controlplane? This should only be "" right which would automatically load the config of mounted sa?

krancour: Great catch. Fixed in e292c08b09c6aae5e2ebb298b3711bfa018ca951

krancour:
Weird linting failures that don't happen locally, even in a clean room container.

krancour:
The doc preview failures seem to be fallout from today's DNS outages.

krancour:
I nudged Netlify and the docs preview built fine. It hasn't updated the checks apparently.

https://684b67acf506d36b6f29b553--docs-kargo-io.netlify.app/

I'm satisfied that there is no problem here. Merging.

Issue #4156, opened 2025-05-20
UI(pipeline): alternative view than DAG view


Marvin9:
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

One of the leftover from pipeline revamp (https://github.com/akuity/kargo/issues/3770) is list view of pipeline

![Image](https://github.com/user-attachments/assets/b53387e8-431c-4508-a04b-ddb20c82e5a1)

nikolay-te:
How would parallel stages be represented in this list view? Maybe one level nesting/indentation?
Regardless, it's great so see this, the current DAG view can get very unwieldy when you have a large number of stages.

Marvin9:
Just the flat list of stages with more and more useful filters

PR #4371, opened 2025-06-11
feat(ui): stage list view

Marvin9:
https://github.com/akuity/kargo/issues/4156

Screenshot 2025-06-18 at 5 35 16 PM


ui/src/features/project/pipelines/freight/source-catalogue-utils.ts
```diff
@@ -33,6 +33,54 @@ export const catalogueFreights = (freights: Freight[]) => {
...
+export const filterFreightByVersion = (versions: string[]) => (_freight: Freight) => {
+  const freight = { ..._freight };
+
+  if (!versions.length) {
+    return freight;
+  }
+
+  freight.images = freight.images?.filter((image) => versions.includes(image.tag));
+  freight.commits = freight.commits?.filter((commit) => versions.includes(commit.tag));
+  freight.charts = freight.charts?.filter((chart) => versions.includes(chart.version));
```
jmalisz-akuity, line 75: It feels like you don't intend to modify the passed `_freight`, but I think that you do there

ui/src/features/project/pipelines/list/columns/phase-column.tsx
```diff
@@ -0,0 +1,101 @@
+import { Button, Flex } from 'antd';
+import Checkbox from 'antd/es/checkbox/Checkbox';
```
jmalisz-akuity: I would suggest to import from `antd` if possible. In the past I've encountered issues with notifications when they were imported directly from file rather than just `antd`.

ui/src/features/project/pipelines/freight/source-catalogue-utils.ts
Marvin9: It is cloned

ui/src/features/project/pipelines/list/columns/phase-column.tsx
Marvin9: There is a reason I subconsciously import like that. In the past I encountered importing directly from `antd` loads all component of antd which increases bundle size. Not an big issue though at the moment

ui/src/features/project/pipelines/freight/source-catalogue-utils.ts
jmalisz-akuity: You are right, I'm sorry. I thought that the clone wasn't deep enough, but I see now that it is.

ui/src/features/project/pipelines/list/columns/version-column.tsx
```diff
@@ -0,0 +1,215 @@
...
+              (firstFreight?.charts?.slice(0, 2)?.length +
+                firstFreight?.commits?.slice(0, 2)?.length +
+                firstFreight?.images?.slice(0, 2)?.length)}{' '}
+            more
+          </Typography.Text>
+        )}
+      </>
+    );
+  },
+  filterDropdown: () => {
```
jmalisz-akuity: I think that those filters could use a reset button, so you don't need to do it manually

ui/src/features/project/pipelines/list/columns/version-column.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+      return opts;
+    }, [cataloguedFreights]);
+
+    return (
+      <Flex style={{ padding: 16 }} vertical gap={16}>
+        <Flex align='center' gap={8}>
+          <label>Source: </label>
+          <Select
+            className='min-w-[580px] ml-auto'
```
jmalisz-akuity: This width makes it look weird on smaller screens. We probably should make those select narrower, like maybe 300px.

ui/src/features/project/pipelines/list/list-view.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+type PipelineListViewProps = {
+  stages: Stage[];
+  warehouses: Warehouse[];
+  freights: Freight[];
+  className?: string;
+  project: string;
+};
+
+export const PipelineListView = (props: PipelineListViewProps) => {
```
jmalisz-akuity, line 28: In this view the `Create` button lacks spacing, please add it

ui/src/features/project/pipelines/list/columns/stage-column.tsx
```diff
@@ -0,0 +1,65 @@
...
+            Search
+          </Button>
+        </Space>
+      </div>
+    );
+  },
+  onFilter: (value, record) => {
+    return !!record?.metadata?.name?.toLowerCase?.()?.includes((value as string).toLowerCase());
+  },
+  filteredValue: [filter?.stage || ''],
```
jmalisz-akuity: I think that because of this, this filter is always shown as used. Wouldn't something like that work better?
```suggestion
  filteredValue: filter?.stage ? [filter.stage] : null,
```

ui/src/features/project/pipelines/list/columns/version-column.tsx
Marvin9: Good idea

ui/src/features/project/pipelines/list/columns/version-column.tsx
Marvin9: I didn't want to trim the dropdown content. I will see if this is solved with dynamic width by screen

ui/src/features/project/pipelines/list/columns/stage-column.tsx
Marvin9: You beat me to it. It was annoying me. Good catch!

ui/src/features/project/pipelines/list/columns/version-column.tsx
jmalisz-akuity: I think that this still needs fixing, as for me this filter gets cropped already at `1024px`. Which isn't that small.

image

ui/src/features/project/pipelines/list/applied-filters.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return null;
+  }
+
+  return (
+    <div className={classNames(props.className, 'space-y-2')}>
+      <span className='text-xs mr-2 font-bold'>Applied filters: </span>
+      {appliedFilters.map((appliedFilter, idx) => (
+        <Tag key={idx}>
+          {appliedFilter.key}: {appliedFilter.value}
+        </Tag>
```
jmalisz-akuity: Please add a close button to the tag for removing the filter. Unless it will be hard, I think it's nice to have it.

Because of that, the date filter probably should be kept in the single tag.

ui/src/features/project/pipelines/list/use-applied-filters.ts
```diff
@@ -0,0 +1,75 @@
...
+      appliedFilters.push({
+        key: 'start time',
+        value: filters.lastPromotion[0].toString()
+      });
+
+      appliedFilters.push({
+        key: 'end time',
+        value: filters.lastPromotion[1].toString()
+      });
+    }
```
jmalisz-akuity: Please remove the time and timezone data from those. We don't need them, because we don't allow to set it.

Marvin9:
@jmalisz-akuity I agree that it should be consistent but on the other hand multi-select deserves apply/reset because we are not sure when they finalised the selection, while name or date search is simple to predict. Its because of consistency otherwise search filter is more easier if it is out of table (for example if it is above table)

ui/src/features/project/pipelines/list/columns/version-column.tsx
Marvin9: Please check now

jmalisz-akuity: There's still an issue for small mobile screens, but we have those in other places as well. So I think that's fine for now.

ui/src/features/project/pipelines/list/list-view.tsx
jmalisz-akuity: It seems that changes there caused issues with scroll and padding on the regular view. I think it's because previously this bar used absolute, please take a look at it.

Main branch:
image

This branch:
image

jmalisz-akuity:
I generally think that this isn't a problem, we could trigger checkbox filter when someone checks it. This should be consistent with for example Material Design guidelines: https://m3.material.io/components/checkbox/guidelines#aa0a76fe-1046-4369-8425-a723b53d6418

One thing that I hadn't noticed before is that this table is on the same page as regular view... I think that it would be better to completely separate those two and use URL based filters. But I also see that we are missing this already in promotion filters...

I think, that I would leave it as it is for now and create an issue to change that in the future.

rpelczar:
I think we shouldn't add any filters to this new view. We already have simple filters above the table, and if we think that we need more filters, we should add something there. We can keep sorting. ArgoCD also doesn't have filters in the table.

Marvin9:
> This should be consistent with for example Material Design guidelines

Right, so we have let them select multiple checkboxes and then apply when they are ready

> those two and use URL based filters

That is very good iteration I will try to add

> I think we shouldn't add any filters to this new view

I would put this in good to have category and slight improvements, specially version filters so lets keep it.. which directs me to important use case that people are really tied up with freight `alias`, I will add that as well

ui/src/features/project/pipelines/list/list-view.tsx
Marvin9: Done

ui/src/features/project/pipelines/list/columns/version-column.tsx
```diff
@@ -28,6 +35,8 @@ export const versionColumn = (
 ): ColumnType<Stage> => ({
   title: 'Version',
   render: (_, stage) => {
+    const dictionaryContext = useDictionaryContext();
+    const freightTimelineControllerContext = useFreightTimelineControllerContext();
```
jmalisz-akuity, line 16: Using contexts here seems to cause React errors about rule of hooks - the order of hooks changed after re-render.

I'm not sure exactly why this happens, but maybe it's better to pass the context through props.

jmalisz-akuity (commented):
I'm generally for filters in this table, especially since we already took time to implement them. But I'll let you decide about this @Marvin9, @rpelczar.

Regarding the checkbox filters, the Material Design guidelines I've linked before describe usage when checkbox is turning something on or off - without additional "apply" action. I think it corresponds to our usage of checkboxes here as turning the filter on or off. I would prefer for those filters to be consistent and I think we could just use all of them without additional "apply" action. But I think it's fine as it is currently, so I won't argue for this further.

ui/src/features/project/pipelines/list/list-view.tsx
jmalisz-akuity: There was still some jerkiness when switching between views, so I've adjusted the padding. It's smooth now.

shamsalmon:
thanks for adding this guys. one request... could you add something in the project annotations to give the ability to set a default view for a project? If you would like I can take a shot at adding this.

Marvin9:
> the project annotations to give the ability to set a default view for a project

I would leave that upto @krancour / @hiddeco if we want to annotate project or project config

rpelczar:
@shamsalmon @Marvin9 Can't we just save the latest view in localStorage? First time you always see DAG view, but once you switch to the list view, you will see this view by default for a particular project on the current browser.

shamsalmon:
> @shamsalmon @Marvin9 Can't we just save the latest view in localStorage? First time you always see DAG view, but once you switch to the list view, you will see this view by default for a particular project on the current browser.

I would personally prefer a way to make a project default to the listview without ever seeing the DAG

Marvin9:
> Can't we just save the latest view in localStorage?

I generally saw a pattern that an admin wants to set default view for all team members using kargo so they don't have to

ui/package.json
```diff
@@ -80,6 +80,7 @@
     "monaco-editor": "^0.52.2",
     "monaco-yaml": "^5.4.0",
     "oauth4webapi": "^3.5.1",
+    "rc-picker": "^4.11.3",
```
rpelczar: You can remove this

ui/src/features/project/pipelines/pipelines.tsx
```diff
@@ -255,39 +267,55 @@ export const Pipelines = (props: { creatingStage?: boolean; creatingWarehouse?:
...
-                <Button
-                  icon={<FontAwesomeIcon icon={faDocker} />}
-                  onClick={() =>
-                    setPreferredFilter({
-                      ...preferredFilter,
-                      images: !preferredFilter?.images
-                    })
-                  }
-                />
+                {pipelineView === 'graph' && (
```
rpelczar: Can we keep this for both views?

ui/src/features/project/pipelines/pipelines.tsx
```diff
@@ -199,10 +202,19 @@ export const Pipelines = (props: { creatingStage?: boolean; creatingWarehouse?:
...
 
             <div className='w-full h-full relative'>
-              <Flex gap={12} className='absolute z-10 top-2 right-2 left-2'>
+              <Flex
+                gap={12}
+                className={classNames(
+                  'z-10 top-2 right-2 left-2',
+                  pipelineView === 'graph' ? 'absolute' : 'pt-2 px-2'
+                )}
+                align='center'
```
rpelczar: I think it looks worse than before. Can we align it to the top?

jmalisz-akuity: I assume that you mean the filters inside card? This align there doesn't seem to do anything, probably should be removed.

Also a small nitpick, this tag inside select seems a bit cropped.
image

rpelczar: Nope, I mean this: 
image

Marvin9:
> If we don't use filters, then we might want to keep sorting

Sorting is there for stage names and last promotion

Issue #4315, opened 2025-06-02
prep work for webhook receivers UI


fuskovic:
# Proposed Feature

We've started discussing adding UI around webhook receivers. 

~~Since the config for webhook receivers is rolled up under project configs, and there is no connect RPC support for performing CRUD on project configs, so it looks like we'll need those before we can get to the UI.~~

krancour:
Don't these exist?

https://github.com/akuity/kargo/blob/b031b380a32bdc01d09f80d224f75a2b52e35f60/api/service/v1alpha1/service.proto#L56-L57

Perhaps it was the absence of create and update endpoints that led to this issue? For most types the UI uses a generic endpoint for creating and updating, so unless @Marvin9 asks us for something specific, I _think_ our bases are covered.

Marvin9:
No specific asks at the moment but highly depends on how we want to do the webhook UI cc @rpelczar in which case I believe there would be separate webhook related endpoints

krancour:
@Marvin9, @jessesuen had mentioned maybe a single form for both the webhook configuration and whatever Secret it needs to reference, so perhaps that may call for a more specialized endpoint.

rpelczar:
@Marvin9 and I discussed this. First, we will try to implement this feature without any API changes. We want to extend the current ProjectConfig setting by adding a preset option for webhooks.

PR #4375, opened 2025-06-12
feat(ui): webhook receiver UI

Marvin9:
fixes https://github.com/akuity/kargo/issues/4315

Screenshot 2025-06-12 at 3 16 29 PM

Screenshot 2025-06-12 at 3 18 35 PM

- [ ] handling edge cases (ie. duplicate secret, duplicate webhook, failure fallbacks)
- [ ] ease of URL view from UI

### DIFF
--- a/ui/src/features/project/settings/views/project-config/project-config.tsx
+++ b/ui/src/features/project/settings/views/project-config/project-config.tsx
@@ -1,4 +1,6 @@
 import { useMutation, useQuery } from '@connectrpc/connect-query';
+import { faPlus } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { Button, Card, Flex, message } from 'antd';
 import type { JSONSchema4 } from 'json-schema';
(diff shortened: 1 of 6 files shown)


Marvin9:
@krancour @hiddeco trying to understand why is [this](https://github.com/akuity/kargo/blob/main/ui/src/gen/api/v1alpha1/generated_pb.ts#L2162) generated schema for ProjectConfig spec is `receivers` and not `webhookReceivers`

hiddeco:
I believe the protobuf tag in the annotation of the field is wrong: https://github.com/akuity/kargo/blob/main/api/v1alpha1/project_config_types.go#L37

Marvin9:
@fuskovic / @krancour is it safe to change this name to `webhookReceivers` or there are more after changes that we need to accommodate?

ui/src/features/project/settings/views/project-config/webhooks.tsx
```diff
@@ -0,0 +1,72 @@
...
+        },
+        {
+          key: 'url',
+          title: 'Webhook URL',
+          width: '60%',
+          render: (_, record) => <WebhookURLColumn details={record} />
+        }
+      ]}
+    />
+  );
```
rpelczar: Please wrap this with a card component.

krancour:
> I believe the protobuf tag in the annotation of the field is wrong: https://github.com/akuity/kargo/blob/main/api/v1alpha1/project_config_types.go#L37

@Marvin9 can fix the mistake referenced above as part of this PR?

Just need to re-run codegen afterwards is all.

🙏

PR #4377, opened 2025-06-12
fix: Updates label selector in deployment

Plork:
Updates the label selector used in the management controller deployment to use the `kargo.labels` helper function.

This ensures consistency and proper label application across the chart.


krancour (approved):

krancour:
Thank you @Plork!

PR #4378, opened 2025-06-12
Remove `\` from yaml in QuickStart

MatthewKennedy:
`\` in the yaml seems to be problematic for me.


krancour:
@MatthewKennedy I assume you didn't copy the entire contents of the samples and paste them into your shell. That is not plain YAML. It is YAML embedded in a shell command and the escaping is therefore necessary.

Issue #4312, opened 2025-06-02
AnalysisRun phase changes don't trigger stage controller reconcile


AlanLonguet:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

AnalysisRun phase changes doesn't trigger stage controller reconcile. AnalysisRun are shown as "Verifying" until the stage gets requeued automatically by the defaultRequeue of 5min.

# Screenshots

![Image](https://github.com/user-attachments/assets/7e5c1187-fd0c-46d6-9e71-fd0dec0e344a)

# Steps to Reproduce

- Use a sharded kargo controller
- We're using a ClusterAnalysisTemplate based on a metric provider
- Launching through the UI or automated by promotion makes no difference on our side

# Version

Using kargo helm chart without overrides (1.5.1)

# Logs

In the logs the EventSource is started correctly, however we don't see the [handler](https://github.com/akuity/kargo/blob/main/internal/controller/stages/event_handlers.go#L489) getting called at all (with debug logs).

```
time="2025-06-02T15:11:48Z" level=info msg="Starting EventSource" controller=stage controllerGroup=kargo.akuity.io controllerKind=Stage source="kind source: *v1alpha1.AnalysisRun"
```

Our hypothesis revolves around this [predicate](https://github.com/akuity/kargo/blob/main/internal/controller/stages/regular_stages.go#L180) masking the phase status change because the generation didn't change but that might be something else.

hiddeco:
Have been unable to reproduce this, and can see the reconciliation of the Stage getting queued in the logs:

```
time="2025-06-03T10:16:06Z" level=debug msg="enqueued Stage for reconciliation" analysisRun=test.01jwtmh6jvcgsngse8j1r94gcr.1a2e85b namespace=kargo-demo-01 stage=test
```

Could you provide a minimal viable example that I can use to attempt to reproduce it?

AlanLonguet:
Oo, well I'll post the Stage and ClusterAnalysisTemplate we're using below, maybe this is a setup issue on our side, is there anything on argo-rollouts or on kargo (e.g. RBAC) that would not permit the events to be properly received ?

```
apiVersion: kargo.akuity.io/v1alpha1
kind: Stage
metadata:
  annotations:
    kargo.akuity.io/color: '#882255'
  labels:
[...]

PR #4390, opened 2025-06-13
fix: sharded kargo setup analysisRun indexer

AlanLonguet:
The indexer was using the rollout controller instance ID instead of the shardName, looking at the underlying function the parameter is called shardName and is used to be checked against the shard label.

This fixes https://github.com/akuity/kargo/issues/4312

I tested it on our live setup and we had no more waiting time after the verification was successful.

### DIFF
--- a/internal/controller/stages/regular_stages.go
+++ b/internal/controller/stages/regular_stages.go
@@ -277,7 +277,7 @@ func (r *RegularStageReconciler) SetupWithManager(
 			ctx,
 			&kargoapi.Stage{},
 			indexer.StagesByAnalysisRunField,
-			indexer.StagesByAnalysisRun(r.cfg.RolloutsControllerInstanceID),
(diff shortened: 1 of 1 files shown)


AlanLonguet:
I can't handle the DCO stuff as I'm not on my own setup but can do in about 8 hours

krancour:
Nice find @AlanLonguet. This was probably so elusive because most of the time the rollouts controller instance ID is the same as the shard name. I guess in your case that must not have been so. (Which is fine. There's nothing wrong with that. And it helped uncover this problem!)

Thanks!

I took care of DCO for you.

AlanLonguet:
@krancour thanks for handling the DCO !

Our setup can be considered elusive as we have our kargo shards in the same cluster and for now using a single rollout instance mainly for querying metrics, so sharded kargo and non-sharded rollouts which clashed with what the indexer was attempting to do.

Our setup might change in the future but that wait on verifications started to be a bit  inconvenient :D

PR #4391, opened 2025-06-13
feat: allow configuration of minimum Warehouse interval

hiddeco:
This adds support for configuring a minimum reconciliation interval for Warehouse resources via the `MIN_WAREHOUSE_RECONCILIATION_INTERVAL` environment variable, which accepts duration strings (e.g., `5m`, `30s`, `1h`).

When set, any Warehouse with a `.spec.interval` lower than this minimum will have the minimum value enforced instead. This helps prevent overly aggressive reconciliation that could impact system performance.

This can be configured using the new Helm chart value:

```yaml
controller:
  reconcilers:
    warehouses:
      minReconciliationInterval: 5m
```

(NB: @krancour, note that I slightly changed your suggestion here to add more context — you may need to update the documentation you are working on.)


api/v1alpha1/warehouse_types.go
```diff
@@ -39,6 +43,29 @@ type Warehouse struct {
...
+
+	interval := w.Status.DiscoveredArtifacts.DiscoveredAt.
+		Add(effectiveInterval).
+		Sub(metav1.Now().Time)
+
+	if interval > 0 {
+		return interval
+	}
+
+	return 0
```
hiddeco: I have been wondering if this is the correct approach, as an interval of `0` means "do not requeue at all" — which seems like it could cause issues with getting stuck?

charts/kargo/values.yaml
```diff
@@ -449,6 +449,8 @@ controller:
     warehouses:
       ## @param controller.reconcilers.warehouses.maxConcurrentReconciles optionally overrides the maximum number of Warehouse resources the controller can reconcile concurrently.
       maxConcurrentReconciles:
+      ## @param controller.reconcilers.warehouses.minReconciliationInterval optionally sets the minimum reconciliation interval for Warehouse resources. Accepts duration format (e.g., "5m", "1h", "30s"). If a Warehouse specifies an interval lower than this minimum, the minimum value will be enforced instead. If not set, no minimum is enforced.
```
krancour, line 452: Enforcing no minimum if not set seems as if it may be asking for trouble. It's easy mor me to imagine an operator overlooking this setting and many developers, impatient with the amount of time their Warehouses require to detect new artifacts, doing the exact wrong thing by turning the interval down quite low on every Warehouse.

imho, we should default this to 5m.

api/v1alpha1/warehouse_types.go
krancour: I agree that looks wrong. (And I know it was my mistake that just got copied to here.)

Can we just return a very small interval in this case?

krancour: Actually... shouldn't we base this on the last time we _looked_ for artifacts rather than the last time we successfully found some? (Again, I think this was my mistake.)

krancour: Closing the loop. I realize now that `w.Status.DiscoveredArtifacts.DiscoveredAt` does reflect the last time we _looked_ and not just the last time we found something.

internal/controller/warehouses/warehouses.go
```diff
@@ -28,8 +28,9 @@ import (
 )
 
 type ReconcilerConfig struct {
-	ShardName               string `envconfig:"SHARD_NAME"`
-	MaxConcurrentReconciles int    `envconfig:"MAX_CONCURRENT_WAREHOUSE_RECONCILES" default:"4"`
+	ShardName                 string        `envconfig:"SHARD_NAME"`
+	MaxConcurrentReconciles   int           `envconfig:"MAX_CONCURRENT_WAREHOUSE_RECONCILES" default:"4"`
+	MinReconciliationInterval time.Duration `envconfig:"MIN_WAREHOUSE_RECONCILIATION_INTERVAL"`
```
krancour, line 33:
```suggestion
	MinReconciliationInterval time.Duration `envconfig:"MIN_WAREHOUSE_RECONCILIATION_INTERVAL" default:"5m"`
```

hiddeco: I prefer setting this via the chart only to allow an empty value still to be set if people are sure they do not want to set a global default.

krancour: Since there's a default set on `WarehouseSpec.Interval`, I guess I am ok with this.

I'm happy as long as there's _something_ ensuring that blanking this out in the chart doesn't instantly make everything go nuts, and it seems that will not happen.

👍

krancour (approved):

krancour:
Fixed a gci formatting issue.

PR #4398, opened 2025-06-14
refactor: DRY up webhook receivers

krancour:
All receivers had identical logic for verifying the request body size didn't exceed some maximum and for returning the results of fully or partially successful Warehouse refreshes.

This PR resolves that.

### DIFF
--- a/internal/webhook/external/bitbucket_test.go
+++ b/internal/webhook/external/bitbucket_test.go
@@ -2,9 +2,6 @@ package external
 
 import (
 	"bytes"
-	"context"
-	"encoding/json"
-	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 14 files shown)


internal/webhook/external/bitbucket_test.go
```diff
@@ -246,14 +157,19 @@ func TestBitbucketHandler(t *testing.T) {
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
+			requestBody, err := io.ReadAll(testCase.req().Body)
+			require.NoError(t, err)
+			defer testCase.req().Body.Close()
```
hiddeco:
```suggestion
			t.Cleanup(func() {
				_ = testCase.req().Body.Close()
			})
```

Same applies to other tests with this structure.

internal/webhook/external/receiver.go
```diff
@@ -71,6 +74,11 @@ func (b *baseWebhookReceiver) GetDetails() kargoapi.WebhookReceiverDetails {
 	return b.details
 }
 
+// getMaxRequestBodyBytes
```
fuskovic: unfinished comment?

krancour: Indeed. Nice catch.

internal/webhook/external/bitbucket_test.go
krancour: Fixed in 10b5b844e7ea72dcd066a046195c8215929d8518

internal/webhook/external/receiver.go
krancour: Fixed in 10b5b844e7ea72dcd066a046195c8215929d8518

Issue #4396, opened 2025-06-14
UI freezes when opening a project with a "long" pipeline.


boh-dan:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

UI freezes when opening a project with a "long" pipeline.
The pipeline contains 50 stages that are grouped by 4 stages and connected into a chain. See diagram (pipeline_after) below. The idea is to promote a Freight to a group of 4 Stages in parallel, wait for verifications, and move to the next group. 

We noticed the issue after changing the pipeline shape. See diagrams below.
**Pipeline before:**
![Image](https://github.com/user-attachments/assets/ff5e6f97-3d84-41ed-9fa1-a8b7567464c0)

**Pipeline after:**
![Image](https://github.com/user-attachments/assets/cfe250d4-c78f-453b-87a5-01c512982218)

# Screenshots

<img width="1085" alt="Image" src="https://github.com/user-attachments/assets/3e6531e2-5fcc-47ec-a4f5-69e25a383de4" />

# Steps to Reproduce
1. Create a pipeline with 50 stages grouped in groups of 4 and connected in a long chain similar to the diagram *Pipeline after* mentioned above.
2. Open Pipeline in Kargo UI

# Version
1.5.1

# Logs

Nothing in logs

# Investigation
I've run the JS profiler on page load and found a culprit:

<img width="823" alt="Image" src="https://github.com/user-attachments/assets/89104c4d-f586-4843-a1ba-05571c5c22e6" />
You can see that it takes more than 60s seconds to load a page.

<img width="386" alt="Image" src="https://github.com/user-attachments/assets/47bea4f6-e5fa-4bdf-bc46-bb3a05371066" />
The last screenshot shows minified code, which I believe corresponds to this code: https://github.com/akuity/kargo/blob/main/ui/src/features/project/pipelines/graph/stack-nodes.ts#L9. This code traverses the graph of stages.

[...]

PR #4402, opened 2025-06-16
fix: Ensure each node is only traversed once using visited set to prevent UI freezing for large pipelines

wesley-yep-canva:
Fixes https://github.com/akuity/kargo/issues/4396

A pipeline with 59 nodes is doing 690,794 iterations as nodes are being traversed multiple times. 
This change adds a set of visited nodes to reduce the number of iterations. Tested locally and it reduces the time from over 12 sec to less than 1 second.

### DIFF
--- a/ui/src/features/project/pipelines/graph/stack-nodes.ts
+++ b/ui/src/features/project/pipelines/graph/stack-nodes.ts
@@ -11,14 +11,6 @@ export const stackNodes = (
   graph: graphlib.Graph,
   stageByName: Record<string, Stage>
 ) => {
-  if (afterNodes.length === 0) {
-    return {
-      stackNodes: [],
-      ignoreList: new Set<string>(),
-      graph
-    };
-  }
-
   const sources = graph.sources();
 
   const stackNodes: Array<{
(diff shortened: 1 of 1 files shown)


Marvin9:
Thanks! I was just looking at the issue. Will test and review

ui/src/features/project/pipelines/graph/stack-nodes.ts
```diff
@@ -20,21 +20,31 @@ export const stackNodes = (
   }> = [];
 
   const ignoreList = new Set<string>();
+  const afterNodesSet = new Set(afterNodes);
+  const processedParents = new Set<string>();
+  const visited = new Set<string>();
 
   // @ts-expect-error it is string array
```
Marvin9, line 34: There one more thing we can do, do not iterate at all if `afterNodes` is empty

ui/src/features/project/pipelines/graph/stack-nodes.ts
```diff
@@ -20,21 +20,31 @@ export const stackNodes = (
   }> = [];
 
   const ignoreList = new Set<string>();
+  const afterNodesSet = new Set(afterNodes);
```
Marvin9: Doesn't seem worth to create set for this. The array list is going to be very small anyways

Marvin9 (changes requested):
couple of minor changes and it is good to merge. Thanks!

ui/src/features/project/pipelines/graph/stack-nodes.ts
wesley-yep-canva: Makes sense, added check for this now

Marvin9 (approved):
LGTM, Thanks! 🚀

PR #4407, opened 2025-06-16
helm-pull to untar helm charts from oci registries

shabbirsaifee92:


koorikla:
I created a similar one into the helm template step which imo is more similar to how cli is usually used
https://github.com/akuity/kargo/pull/3760


But looks like there is a bigger plan around this - https://github.com/akuity/kargo/issues/1610

shabbir-saifee-cmm:
> I created a similar one into the helm template step which imo is more similar to how cli is usually used #3760
> 
> But looks like there is a bigger plan around this - #1610

Yeah agreed. I am just implementing a quick POC to test out some ideas i have. 
Any idea how is it progressing? I am not seeing any action on https://github.com/akuity/kargo/issues/1610 since march.

PR #4408, opened 2025-06-16
feat: add generic webhook receiver

krancour:
Example usage:

```yaml
kind: ProjectConfig
metadata:
  name: webhooks-demo
  namespace: webhooks-demo
spec:
  webhookReceivers:
  - name: my-receiver
    generic:
      warehouseRefresh:
        predicate: request.header('X-Kargo-Git-Event') == 'push'
        repoURL:   request.body.repository.repo_name
      secretRef:
        name: my-receiver-secret
---
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: my-receiver-secret
  namespace: webhooks-demo
data:
  secret: c295bGVudCBncmVlbiBpcyBwZW9wbGUK
```


internal/webhook/external/generic.go
```diff
@@ -0,0 +1,154 @@
...
+
+		var parsedBody any
+		if err := json.Unmarshal(requestBody, &parsedBody); err != nil {
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
+			)
+			return
+		}
+		env["request"].(map[string]any)["body"] = parsedBody // nolint: forcetypeassert
```
fuskovic: Wdyt of introducing something like: https://github.com/tidwall/gjson?

krancour: expr-lang is used so broadly throughout Kargo, including for parsing response bodies in the http promotion step, that it would probably be more confusing than helpful to use something else here.

api/v1alpha1/project_config_types.go
```diff
@@ -160,6 +163,69 @@ type DockerHubWebhookReceiverConfig struct {
...
+	// +optional
+	ImageRepoURL string `json:"imageRepoURL,omitempty" protobuf:"bytes,3,opt,name=imageRepoURL"`
+	// ChartRepoURL is a user-defined expression that should return a Helm
+	// chart repository URL for which all subscribed Warehouses should be
+	// refreshed. If an empty string is returned, no Warehouses will be refreshed.
+	// If a non-empty string is returned, Helm chart repo URL normalization
+	// will be applied before the URL is used to find Warehouses to refresh.
+	//
+	// +optional
+	ChartRepoURL string `json:"chartRepoURL,omitempty" protobuf:"bytes,4,opt,name=chartRepoURL"`
```
krancour: Having separate fields for each of the supported repo types does two things:

1. It allows specifying separate repo URL extraction logic for each kind of repo, making it possible, for instance, for one generic receiver to process push events for multiple types of repos.

2. It allows us to know what kind of normalization process to apply to the URL, as the logic for normalizing each of Git, image, and chart URLs is different.

internal/webhook/external/generic.go
```diff
@@ -0,0 +1,186 @@
...
+
+		program, err := expr.Compile(g.cfg.ArtifactPush.Predicate)
```
hiddeco: If I am understanding this correctly, a predicate is always required and it's not sufficient to declare e.g. `artifactPush` without one?

krancour: As things stand currently, yes.

I had thought about blindly jumping into seeing if the other expressions evaluate to anything, but that seems to invite in the possibility that one or more of those expressions do coincidentally evaluate to something, but the actual event we're being notified of is not one of interest.

By way of example, ignore for a moment that we have a dedicated receiver for GitHub, and imagine we're using a generic receiver to handle incoming webhooks from there. Nearly all event types include `repository.url` in the payload. The predicate `request.header('X-GitHub-Event') == 'push'` would prevent us from blindly reading `repository.url` from an event that turned out to be, for instance, someone adding a comment to an issue.

Please note that I think this works well, but I'm not entirely convinced that there isn't a more ergonomic option I am overlooking, so if you have ideas, I'm totally open to them.

krancour:
There's too much discussion to be had about this for this to reasonably make it into 1.7.0. Deferring to v1.8.0.

krancour:
@fuskovic you've worked extensively on the webhook receivers and have worked quite a bit with expressions recently. You're sitting right at the intersection of the things involved in this...

I'm going to turn this over to to you. Feel free to either build on what I've done or scrap it and start fresh if that's what ends up making more sense, but either way, don't start any work on this until we have 1.8 buttoned up _and_ until you, @hiddeco, and I can meet to hammer out what a reasonable UX looks like here... previous conversation between @hiddeco and myself was offline, but revealed that there are a lot small concerns that aren't addressed by this PR.

krancour:
Closing in favor of #5305

PR #4411, opened 2025-06-16
docs: github webhook receivers

fuskovic:
Related to: https://github.com/akuity/kargo/issues/4150

PR #4417, opened 2025-06-17
docs: gitlab webhook receivers

fuskovic:
Related to: https://github.com/akuity/kargo/issues/4150


krancour:
Let's get feedback on #4411 resolved first and then tweak this and docs for other receivers to match afterwards.

krancour (approved):

PR #4422, opened 2025-06-18
feat: make image/oci webhook receivers also work for charts

krancour:
Warehouses subscribed to charts in Docker Hub could not successfully be refreshed.

For GitHub/GHCR and Quay, this _did_ work, but only coincidentally. We didn't design for it.

We now check the media type conveyed by payloads from GitHub/GHCR and Docker Hub to determine what kind of normalization logic to apply to the URL we extract from the payload before requesting a refresh of Warehouses subscribed to that URL.

Quay doesn't send media type information, so we normalize the URL _both_ as if it were an image URL and as if it were a chart URL, then refresh Warehouses subscribed to either. This currently results in two identical URLs, but I don't want to take it for granted that normalization logic will never change. To account for this, the refresh logic is also updated to de-dupe URLs.

### DIFF
--- a/internal/webhook/external/github.go
+++ b/internal/webhook/external/github.go
@@ -166,17 +166,10 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc 
 				return
 			}
 			manifest := pkg.GetPackageVersion().GetContainerMetadata().GetManifest()
-			// Determine if the package was a Helm chart
-			if cfgAny, ok := manifest["config"]; ok {
-				if cfg, ok := cfgAny.(map[string]any); ok {
-					if mediaTypeAny, ok := cfg["media_type"]; ok {
-						if mediaType, ok := mediaTypeAny.(string); ok {
-
-							if mediaType == helmChartMediaType {
-								repoURL = helm.NormalizeChartRepositoryURL(ref.Context().Name())
-							}
-						}
-					}
+			// Determine if the package is a Helm chart
+			if cfg, ok := manifest["config"].(map[string]any); ok {
+				if mediaType, ok := cfg["media_type"].(string); ok && mediaType == helmChartMediaType {
+					repoURL = helm.NormalizeChartRepositoryURL(ref.Context().Name())
 				}
 			}
 			if repoURL == "" {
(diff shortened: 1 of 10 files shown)


internal/webhook/external/github.go
```diff
@@ -164,7 +165,18 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+			repoURLs = append(
+				repoURLs,
+				repoURL,
+				// This covers our bases in the event that the package event originated
+				// from an OCI repository that actually contains Helm charts. All OCI
+				// Helm chart repositories in Kargo start with oci://. So we're going to
+				// try to refresh Warehouses subscribed to both, e.g.,
+				// ghcr.io/example/repo AND oci://ghcr.io/example/repo. Only one of the
+				// two should actually work.
+				helm.NormalizeChartRepositoryURL(fmt.Sprintf("oci://%s", repoURL)),
```
hiddeco: Shouldn't this be:


```suggestion
				fmt.Sprintf("oci://%s", helm.NormalizeChartRepositoryURL(repoURL)),
```

As `NormalizeChartRepositoryURL` strips the `oci://` prefix, and adding it seems otherwise redundant.

krancour: Disregard. I hate how I did this whole thing.

internal/webhook/external/github.go
```diff
@@ -164,7 +165,24 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+					if mediaTypeAny, ok := cfg["media_type"]; ok {
+						if mediaType, ok := mediaTypeAny.(string); ok {
+
+							if mediaType == helmChartMediaType {
+								repoURL = helm.NormalizeChartRepositoryURL(ref.Context().Name())
+							}
+						}
+					}
+				}
+			}
```
hiddeco: Might be slightly neater to do something like:

```go
func extractMediaType(manifest map[string]any) (string, bool) {
    config, ok := manifest["config"].(map[string]any)
    if !ok {
        return "", false
    }
    mediaType, ok := config["media_type"].(string)
    return mediaType, ok
}
```

krancour: Ok. That's a neat trick. It didn't occur to be that I could do a retrieval and a type coercion in one fell swoop.

Fixed in 13b26915738f59117195f405213776d3f3a013bc

hiddeco (approved):

Issue #4433, opened 2025-06-20
Add `-trimpath` and update `-ldflags` for `kargo` and `credential-helper` for production builds


zeraye:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

I would like to add:
- `-trimpath` flag to go build (for both `kargo` and `credential-helper`),
- `s` parameter to `-ldflags` flag.

# Motivation

Reduce binary size. Below are binary sizes after building `kargo` using `make build-cli` on the `main` branch on my `linux/amd64` pc:

| Flags | Binary size (bytes) 
| - | - |
| no changes | 65193996 |
| `-trimpath` | 65136602 |
| added `s` to `-ldflags` | 56479928 |
| `-trimpath` + added `s` to `-ldflags` | 56422584 |

I understand we want to have more debugging information for the development builds, but for the production one it shouldn't be necessary. I expect there is a reason these flags are not present, but I couldn't find information why.

hiddeco:
While I do agree `-trimpath` should be added, adding `-s` would complicate the support we can provide to users because it would obfuscate [the information we request from people](https://docs.kargo.io/contributor-guide/debugging-kargo/#enabling-pprof-endpoints) when they are running into performance issues or other hard-to-trace bugs.

PR #4451, opened 2025-06-23
chore: build w/ `-trimpath` and `-ldflags -w`

hiddeco:
Related to #4433

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,10 +45,13 @@ ARG GIT_COMMIT
 ARG GIT_TREE_STATE
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -45,10 +45,13 @@ ARG GIT_COMMIT
 ARG GIT_TREE_STATE
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
+      -trimpath \
+      -ldflags "-w -s" \
```
hiddeco, line 49: Note: `-s` is fine here, as I expect the chances for us to have the desire to `pprof` this are minimal as opposed to the controller binary.

krancour (approved):

PR #4458, opened 2025-06-23
feat: gitea webhook receiver

### LINKED ISSUE #4489: add webhook receiver for gitea
Broken out from #4150

Add a webhook receiver for refreshing Warehouses in response to commits being pushed to Gitea.

fuskovic:
Closes #4489

### DIFF
--- a/internal/webhook/external/gitea.go
+++ b/internal/webhook/external/gitea.go
@@ -16,7 +16,9 @@ import (
 )
 
 const (
-	giteaSecretDataKey = "secret"
+	giteaEventTypeHeader = "X-Gitea-Event"
+	giteaSignatureHeader = "X-Hub-Signature-256"
+	giteaSecretDataKey   = "secret"
 
 	gitea = "gitea"
 )
@@ -84,7 +86,7 @@ func (g *giteaWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc {
 			return
 		}
 
-		eventType := r.Header.Get("X-Gitea-Event")
+		eventType := r.Header.Get(giteaEventTypeHeader)
 		switch eventType {
 		case "push":
 		default:
@@ -101,7 +103,7 @@ func (g *giteaWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc {
 		logger = logger.WithValues("eventType", eventType)
 		ctx = logging.ContextWithLogger(ctx, logger)
 
-		sig := r.Header.Get(gh.SHA256SignatureHeader)
+		sig := r.Header.Get(giteaSignatureHeader)
 		if sig == "" {
 			xhttp.WriteErrorJSON(
 				w,
--- a/internal/webhook/external/gitea_test.go
+++ b/internal/webhook/external/gitea_test.go
@@ -58,7 +58,7 @@ func TestGiteaHandler(t *testing.T) {
 			secretData: testSecretData,
 			req: func() *http.Request {
 				req := httptest.NewRequest(http.MethodPost, testURL, nil)
-				req.Header.Set("X-Gitea-Event", "nonsense")
+				req.Header.Set(giteaEventTypeHeader, "nonsense")
 				return req
 			},
 			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
@@ -75,7 +75,7 @@ func TestGiteaHandler(t *testing.T) {
 			secretData: testSecretData,
 			req: func() *http.Request {
 				req := httptest.NewRequest(http.MethodPost, testURL, nil)
-				req.Header.Set("X-Gitea-Event", "push")
+				req.Header.Set(giteaEventTypeHeader, "push")
 				return req
 			},
 			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
@@ -88,8 +88,8 @@ func TestGiteaHandler(t *testing.T) {
 			secretData: testSecretData,
 			req: func() *http.Request {
 				req := httptest.NewRequest(http.MethodPost, testURL, nil)
-				req.Header.Set("X-Gitea-Event", "push")
-				req.Header.Set("X-Hub-Signature-256", "totally-invalid-signature")
+				req.Header.Set(giteaEventTypeHeader, "push")
+				req.Header.Set(giteaSignatureHeader, "totally-invalid-signature")
 				return req
 			},
 			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
@@ -103,8 +103,8 @@ func TestGiteaHandler(t *testing.T) {
 			req: func() *http.Request {
 				bodyBuf := bytes.NewBuffer([]byte("invalid json"))
 				req := httptest.NewRequest(http.MethodPost, testURL, bodyBuf)
-				req.Header.Set("X-Hub-Signature-256", sign(bodyBuf.Bytes()))
-				req.Header.Set("X-Gitea-Event", "push")
+				req.Header.Set(giteaSignatureHeader, sign(bodyBuf.Bytes()))
+				req.Header.Set(giteaEventTypeHeader, "push")
 				return req
 			},
 			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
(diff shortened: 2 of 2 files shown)


api/v1alpha1/project_config_types.go
```diff
@@ -120,6 +120,26 @@ type WebhookReceiverConfig struct {
 	// Quay contains the configuration for a webhook receiver that is compatible
 	// with Quay payloads.
 	Quay *QuayWebhookReceiverConfig `json:"quay,omitempty" protobuf:"bytes,4,opt,name=quay"`
+	// Gitea contains the configuration for a webhook receiver that is compatible
+	// with Gitea payloads.
+	Gitea *GiteaWebhookReceiverConfig `json:"gitea,omitempty" protobuf:"bytes,7,opt,name=gitea"`
+}
+
+type GiteaWebhookReceiverConfig struct {
```
krancour, line 130:
```suggestion
// GiteaWebhookReceiverConfig describes a webhook receiver that is compatible
// with Gitea payloads.
type GiteaWebhookReceiverConfig struct {
```

krancour: Will need to re-run codegen.

internal/webhook/external/gitea.go
```diff
@@ -0,0 +1,145 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/git"
+	xhttp "github.com/akuity/kargo/internal/http"
+	"github.com/akuity/kargo/internal/logging"
+)
+
+const (
+	GiteaSecretDataKey = "secret"
```
krancour: No need to export this. Most of the receivers don't. The GitHub receiver does, but it's an exception. A few other packages use it. If you modeled this closely on that receiver, this is probably just copy/pasta.

```suggestion
	giteaSecretDataKey = "secret"
```

internal/webhook/external/gitea.go
```diff
@@ -0,0 +1,145 @@
...
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(errors.New("unauthorized"), http.StatusUnauthorized),
+			)
+			return
+		}
+
+		payload := struct {
+			Repo struct {
+				URL string `json:"html_url"`
```
krancour: I didn't dig too deep, but at a glance, it looks like `clone_url` may be a better place to get this from.

Ref: https://docs.gitea.com/usage/webhooks#example

```suggestion
				URL string `json:"clone_url"`
```

internal/webhook/external/gitea_test.go
```diff
@@ -0,0 +1,179 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/indexer"
+)
+
+const giteaSigningKey = "mysupersecrettoken"
```
krancour: Since the majority of the receivers rely on a signing key and need one to test with, there is a `testSigningKey` already defined in `common_test.go` and we can reuse that.

```suggestion
```

internal/webhook/external/gitea_test.go
```diff
@@ -0,0 +1,179 @@
...
+		} `json:"repository"`
+	}{Repo: struct {
+		Name string `json:"html_url"`
+	}{Name: "https://gitea.com/example/repo"}}
+
+	testScheme := runtime.NewScheme()
+	require.NoError(t, kargoapi.AddToScheme(testScheme))
+
+	testSecretData := map[string][]byte{
+		GiteaSecretDataKey: []byte(giteaSigningKey),
```
krancour:
```suggestion
		GiteaSecretDataKey: []byte(testSigningKey),
```

internal/webhook/external/gitea_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	const testProjectName = "fake-project"
+
+	var validPushEvent = struct {
+		Repo struct {
+			Name string `json:"html_url"`
+		} `json:"repository"`
+	}{Repo: struct {
+		Name string `json:"html_url"`
+	}{Name: "https://gitea.com/example/repo"}}
```
krancour: Nesting anonymous struct types can be clunky. (Notice you had to repeat `json:"html_url"` twice to make this work?) For the small chunk of JSON you want to get by marshaling this, I would just define it as a string literal instead.

```suggestion
	const giteaWebhookRequestBodyImage = `
{
	"repository": {
		"clone_url": "https://gitea.com/example/repo.git"
	}
}`
```

internal/webhook/external/gitea_test.go
```diff
@@ -0,0 +1,179 @@
...
+				req.Header.Set("X-Gitea-Event", "push")
+				return req
+			},
+			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
+				require.Equal(t, http.StatusBadRequest, rr.Code)
+				require.JSONEq(t, `{"error":"invalid request body"}`, rr.Body.String())
+			},
+		},
+		{
+			name:       "success -- push event",
```
krancour: Nit: It's the only success case.

```suggestion
			name:       "success",
```

internal/webhook/external/gitea_test.go
krancour: The multi-line string in my suggestion is an eyesore though, only because it makes the indenting look all off. If you don't care, then disregard. The other option is to move this const up to somewhere around line 20 and then the indenting won't look off at all. Whatever you want to do is fine.

krancour:
Only some very minor things to address. This really looks fantastic!

api/v1alpha1/project_config_types.go
```diff
@@ -120,6 +120,26 @@ type WebhookReceiverConfig struct {
 	// Quay contains the configuration for a webhook receiver that is compatible
 	// with Quay payloads.
 	Quay *QuayWebhookReceiverConfig `json:"quay,omitempty" protobuf:"bytes,4,opt,name=quay"`
+	// Gitea contains the configuration for a webhook receiver that is compatible
+	// with Gitea payloads.
+	Gitea *GiteaWebhookReceiverConfig `json:"gitea,omitempty" protobuf:"bytes,7,opt,name=gitea"`
```
krancour, line 125: Note: There are PRs open currently for three new receivers and all three are competing for field index `7`. So this is just a reminder to merge the latest from main and update this index as needed.

internal/webhook/external/gitea_test.go
```diff
@@ -0,0 +1,174 @@
...
+				require.JSONEq(t, `{"error":"missing signature"}`, rr.Body.String())
+			},
+		},
+		{
+			name:       "invalid signature",
+			secretData: testSecretData,
+			req: func() *http.Request {
+				req := httptest.NewRequest(http.MethodPost, testURL, nil)
+				req.Header.Set("X-Gitea-Event", "push")
+				req.Header.Set("X-Hub-Signature-256", "totally-invalid-signature")
```
krancour: Let's make constants for these, please.

internal/webhook/external/gitea.go
```diff
@@ -0,0 +1,145 @@
...
+					http.StatusNotImplemented,
+				),
+			)
+			return
+		}
+
+		logger = logger.WithValues("eventType", eventType)
+		ctx = logging.ContextWithLogger(ctx, logger)
+
+		sig := r.Header.Get(gh.SHA256SignatureHeader)
```
krancour: Related to https://github.com/akuity/kargo/pull/4458/files#r2210733437

I wouldn't use `gh.SHA256SignatureHeader` here. I immediately found myself asking if the header Gitea uses really is the same as the one GitHub uses. _It is_, but I had to go look at docs to verify that. To anyone else reading the code in the future, this just looks like a mistake.

I think this is a little different than the situation a few lines down where we have the note about using go-github's signature validation because it's good and it's generic enough. There, writing a short comment was more expedient than implementing our own signature validation. Here, it's more expedient to just make our own `giteaSignatureHeader` constant.

PR #4464, opened 2025-06-24
perf: Use git log -- <path1> <path2>, instead of iterating over batches of commits

boh-dan:
## Context

Git commit discovery takes a significant amount of time because Kargo Controller pulls commits using `git log --skip=% --max-count=%`, which means it needs to spawn a new OS process to get a small group of commits (in our case, 3 commits). It might be "expensive" if the desired commits are located 1000 commits below the HEAD.

## Fix

Instead of listing all the commits and checking if the commit contains subscription `IncludePaths`, use native `git log -- <path1> <path2>`.

This PR fixes - https://github.com/akuity/kargo/issues/4465

### DIFF
--- a/internal/controller/git/work_tree.go
+++ b/internal/controller/git/work_tree.go
@@ -72,7 +72,7 @@ type WorkTree interface {
 	ListTags() ([]TagMetadata, error)
 	// ListCommits returns a slice of commits in the current branch with
 	// metadata such as commit ID, commit date, and subject.
-	ListCommits(limit, skip uint) ([]CommitMetadata, error)
+	ListCommits(limit, skip uint, paths []string) ([]CommitMetadata, error)
 	// CommitMessage returns the text of the most recent commit message associated
 	// with the specified commit ID.
 	CommitMessage(id string) (string, error)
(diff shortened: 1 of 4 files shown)


krancour:
I understand your problem and what you want to do, but this change breaks things badly. The list of paths to include can be written in a variety of dynamic ways (globs, regular expressions, expr-lang expressions) that `git log -- path1 path2` does not understand.

hiddeco:
To add to what Kent is saying.

The clone that is performed has the `blob:none`[^1][^2] filter applied, which means that the size of the clone should be substantially smaller, even if it has to iterate through (the logs of) dozens of commits.

The only potential improvement I can think of to reduce the number of processes is to increase the limit beyond the discovery limit; however, this could hypothetically result in larger memory usage for non-monolithic repositories that do not require such a high number of iterations.

[^1]: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-code--filterltfilter-specgtcode
[^2]: https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/

boh-dan:
@krancour @hiddeco thanks for the comments. Last week I've tried to create another PR with a new field `CommitBatchSize` in GitSubscription. The problem is that I can't run `make hack-codegen` it fails on `buf ...` command with rate limit error, probably because I don't have a premium membership.

Is there a way to bypass this issue?

hiddeco:
@boh-dan the rate limit should go away after a cooldown period of +-1h.

Alternatively, logging in using `./hack/bin/buf login` and then running `make codegen` should make it go away as well. The signup for this is without any cost.

boh-dan:
@krancour @hiddeco Fixed this issue in another [PR](https://github.com/akuity/kargo/pull/4552) by adding a new field CommitBatchSize in GitSubscription.

PR #4493, opened 2025-06-25
feat: `helm-update-chart` with `Chart.yaml` SemVer ranges

hiddeco:
This was already a direct by-effect when running the `helm-update-chart` step with a `Chart.yaml` file that had both exact SemVers and ranges defined.

By making the declaration of exact updates optional, dependencies can now be updated according to the declared ranges in the `Chart.yaml` file without requiring any further attachment to e.g., Freight. This can be useful when you want the latest version available for your dependencies anytime you make use of the chart, without being explicitly subscribed to the updates through Freight.

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/helm-update-chart.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/helm-update-chart.md
@@ -16,7 +16,7 @@ referenced by the Freight being promoted. This step is commonly followed by a
 | Name | Type | Required | Description |
 |------|------|----------|-------------|
 | `path` | `string` | Y | Path to a Helm chart (i.e. to a directory containing a `Chart.yaml` file). This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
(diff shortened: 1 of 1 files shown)


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/helm-update-chart.md
```diff
@@ -16,7 +16,7 @@ referenced by the Freight being promoted. This step is commonly followed by a
 | Name | Type | Required | Description |
 |------|------|----------|-------------|
 | `path` | `string` | Y | Path to a Helm chart (i.e. to a directory containing a `Chart.yaml` file). This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
-| `charts` | `[]string` | Y | The details of dependency (subschart) updates to be applied to the chart's `Chart.yaml` file. |
+| `charts` | `[]string` | N | The details of dependency (subchart) updates to be applied to the chart's `Chart.yaml` file. When left unspecified, charts will be updated according to the SemVer ranges declared in the `Chart.yaml` file. |
```
krancour: This was wrong to begin with, but let's fix it while already touching this row.

```suggestion
| `charts` | `[]object` | N | The details of dependency (subchart) updates to be applied to the chart's `Chart.yaml` file. When left unspecified, charts will be updated according to the SemVer ranges declared in the `Chart.yaml` file. |
```

krancour (approved):

PR #4497, opened 2025-06-25
feat: add jira promo step doc

gdsoumya:


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
+---
+sidebar_label: jira
+description: Integrates with Jira to manage issues, comments, and track promotion workflows.
+---
+
+# `jira`
+
+:::info Enterprise Feature
+This extension is only available for AKP Enterprise Kargo, you can use this feature in Kargo v1.6+ offered on [Akuity Platform](https://akuity.io/akuity-platform).
```
kencochrane: Instead of saying "AKP Enterprise Kargo" We should say "Akuity Promote"

gdsoumya: Sure. Fixed

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
+---
+sidebar_label: jira
```
jessesuen, line 2: Is there an icon we can use to indicate this as an enterprise feature?

gdsoumya: Not sure @krancour  / @kencochrane any ideas? Also any ideas how we can possibly add it to the sidebar?

fykaa: we can perhaps have a _lock icon_ 

that simply goes with `sidebar_label: jira 🔒`

image

but i also feel emojis often look informal, another way is we can add a sidebar class and use CSS to prepend a professional enterprise icon instead of using the emo

krancour: My preference would be something like this in the headings of applicable pages. I think we can leave the sidebar alone for now (mostly because I think that might be a much harder thing to tackle since it's all driven off of headings and front matter).

Screenshot 2025-06-30 at 2 10 50 PM

@rpelczar or @Marvin9 can give some better guidance on the CSS, but the example above was done like so:

Added to `docs/src/css/custom.css`:

```css
.enterprise-only::after {
  content: "Enterprise Only";
  display: inline-block;
  margin-left: 0.5em;
  padding: 0.1em 0.5em;
  background: #1976d2;
  color: #fff;
  font-size: 0.3em;
  border-radius: 0.3em;
  vertical-align: middle;
  font-weight: normal;
}
```

Heading on relevant page modified like so:

```html
# Jira <span className="enterprise-only"></span>
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+description: Integrates with Jira to manage issues, comments, and track promotion workflows.
+---
+
+# `jira`
+
+:::info Enterprise Feature
+This extension is only available in Akuity Promote, you can use this feature in Kargo v1.6+ offered on [Akuity Platform](https://akuity.io/akuity-platform).
+:::
+
+The `jira` extension provides comprehensive integration with Jira, allowing you to create, update, delete, and search for issues, manage comments, and track promotion workflows. This extension is particularly useful for maintaining traceability between your deployment processes and project management activities.
```
krancour: With tables and long URLs being notable exceptions, we try to observe a (soft) limit of 80 chars per line. It may seem silly, but it works out a lot better when reviewing PRs from a mobile device or viewing many docs or pieces of code side-by-side on a larger screen.

krancour: As a matter of terminology, is "Akuity Promote" synonymous with "Kargo Enterprise?"

I just want to be sure that we're being clear and consistent so readers know how they could gain access to this feature (and others to come).

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+The extension supports various operations including issue management, comment handling, and status tracking, making it a powerful tool for promotion workflows that require coordination with project management systems.
+
+## Credentials Configuration
+
+All Jira operations require proper authentication credentials stored in a Kubernetes secret.
+
+| Name                     | Type     | Required | Description                                                                                                                |
+| ------------------------ | -------- | -------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `credentials.secretName` | `string` | Y        | Name of the secret containing the Jira credentials.                                                                        |
+| `credentials.namespace`  | `string` | N        | Namespace for the credentials secret. Can be project or cluster secret namespace. If empty, the project namespace is used. |
```
krancour: This feels like rough UX.

I assume we don't allow consuming a Secret from any arbitrary namespace. (We shouldn't.) If this is the case and there are only two namespaces where the creds might be found -- either the Project's own namespace or the cluster secrets namespace -- then can we just ask the user to indicate project-scoped or cluster-scoped? Especially in the case of cluster-scoped, this would allow the user to use the cluster-scoped credentials without needing to know which namespace is designated as the cluster secrets namespace.

The other danger I sense here is that, assuming this doesn't permit using Secrets from any arbitrary namespace, users can easily be misled into thinking it does. And although that would be a Very Bad Thing™, I guarantee some user, disappointed to find out that _doesn't_ work, will end up asking for it, and it's a thing we're going to have to say no to on security grounds.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+
+The referenced secret should contain the following keys:
+
+- `domain`: The domain of your Jira instance or Jira api (e.g., `https://yourcompany.atlassian.net`)
+- `username`: Your Jira username or email
+- `password`: Your Jira API token or password
+
+:::info Content Formatting
+When using `description` or `body` fields, use plain text as Jira does not render markdown. For rich formatting, use the ADF (Atlassian Document Format) alternatives `adfDescription` or `adfBody` instead.
```
krancour: Up to this point in the doc neither `description` nor `body` fields have been mentioned. I'm actually confused, so readers will be also.

Re: ADF, is there any external resource we can link to that covers it?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+      createIssue:
+        projectKey: DEPLOY
+        summary: "Deploy ${{ imageFrom(vars.imageRepo).Tag }} to ${{ ctx.stage.name }}"
+        description: "Deploying ${{ imageFrom(vars.imageRepo).RepoURL }}:${{ imageFrom(vars.imageRepo).Tag }} to ${{ ctx.stage.name }} environment. Promotion ID: ${{ ctx.promotion.name }}. Freight: ${{ ctx.targetFreight.name }}."
+        issueType: Task
+        assigneeEmail: devops@company.com
+        labels:
+          - deployment
+          - "${{ ctx.stage.name }}"
+          - "release-${{ imageFrom(vars.imageRepo).Tag }}"
```
krancour: In code and docs, Kargo has tried to standardize on this convention WRT to YAML lists:

```suggestion
        labels:
        - deployment
        - "${{ ctx.stage.name }}"
        - "release-${{ imageFrom(vars.imageRepo).Tag }}"
```

krancour: May apply in other spots.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Another matter of terminology, I think we could afford to be less generic than "extension" and say "This _promotion step_..."

imho, this is more straightforward for the reader.

Applies in a few other spots throughout this doc.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+## Credentials Configuration
+
+All Jira operations require proper authentication credentials stored in a Kubernetes secret.
+
+| Name                     | Type     | Required | Description                                                                                                                |
+| ------------------------ | -------- | -------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `credentials.secretName` | `string` | Y        | Name of the secret containing the Jira credentials.                                                                        |
+| `credentials.namespace`  | `string` | N        | Namespace for the credentials secret. Can be project or cluster secret namespace. If empty, the project namespace is used. |
+
+The referenced secret should contain the following keys:
```
krancour:
```suggestion
The referenced `Secret` should contain the following keys:
```

Throughout the existing docs, we try to upcase and use monospacing for Kubernetes resource types.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+
+```yaml
+# Access the default issue key
+issueKey: ${{ freightMetadata(ctx.targetFreight.name, 'jira-issue-key') }}
+
+# Access a custom issue key (when issueAlias was used during creation)
+issueKey: ${{ freightMetadata(ctx.targetFreight.name, 'my-custom-alias') }}
+```
+
+## Multi-Stage Workflow Example
```
krancour: I like this wording of "Multi-Stage Workflow" and want to suggest rewording "Freight Metadata Integration" as "Multi-Stage Workflow" and nesting this example within that section.

My rationale is that working with one issue across multiple Stages/Promotions is going to be the thing users care about and potentially go looking for information on. The metadata is just an implementation detail. imho, focusing on the use case instead of the implementation will make it easier for readers to find relevant info faster.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: I initially mentioned Akuity Enterprise Kargo and Ken suggested to make it Akuity Promote so I am guessing it is the more appropriate term.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: It will be in the next parts where we have the cofngiuration info/examples. I put it at the top because it's better to clarify it before we suddenly start using it without explaining how it works. Will add external link to ADF.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: 1.6 is already released I wouldn't want to break this now. I think the documentation makes it clear what to expect and the user won't be asking for other ns because it's ee only and they cannot create any random ns.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Updated `extension` to `promotion step` everywhere

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Makes sense, updated.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: updated

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: updated

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Rephrased it to be more descriptive and less confusing. Also added link

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
+---
+sidebar_label: jira
+description: Integrates with Jira to manage issues, comments, and track promotion workflows.
+---
+
+# `jira`
+
+:::info Enterprise Feature
+This promotion step is only available in Akuity Promote, you can use this feature in Kargo v1.6+ offered on [Akuity Platform](https://akuity.io/akuity-platform).
+:::
```
krancour: @kencochrane this is a spot we were talking about offline where we need to nail down wording to be more clear.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+
+The referenced `Secret` should contain the following keys:
+
+- `domain`: The domain of your Jira instance or Jira api (e.g., `https://yourcompany.atlassian.net`)
+- `username`: Your Jira username or email
+- `password`: Your Jira API token or password
+
+:::info Content Formatting
+The Jira configuration supports setting issue and comment content using `description` or `body` fields. These are plain text fields that do not support special formatting such as Markdown. For rich formatting capabilities, use the ADF (Atlassian Document Format) alternatives `adfDescription` or `adfBody`. For more information about ADF structure and formatting, see the [Atlassian Document Format documentation](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/).
+:::
```
krancour: We should probably move this down because it currently looks to be part of the credentials section.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+This promotion step supports various operations including issue management, comment handling, and status tracking, making it a powerful tool for promotion workflows that require coordination with project management systems.
+
+## Credentials Configuration
+
+All Jira operations require proper authentication credentials stored in a Kubernetes `Secret`.
+
+| Name                     | Type     | Required | Description                                                                                                                  |
+| ------------------------ | -------- | -------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| `credentials.secretName` | `string` | Y        | Name of the `Secret` containing the Jira credentials.                                                                        |
+| `credentials.namespace`  | `string` | N        | Namespace for the credentials `Secret`. Can be project or cluster secret namespace. If empty, the project namespace is used. |
```
krancour:
```suggestion
| `credentials.namespace`  | `string` | N        | Namespace containing the credentials `Secret`. This can be the Project's namespace or the designated cluster secrets namespace. If empty, the Project's namespace is assumed. |
```

krancour: I can't remember where we landed on it, but if the cluster secrets namespace isn't configurable in Akuity Promote and this feature is exclusive to Akuity Promote, should we say here what the name of the the cluster secrets namespace actually is?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+description: Integrates with Jira to manage issues, comments, and track promotion workflows.
+---
+
+# `jira`
+
+:::info Enterprise Feature
+This promotion step is only available in Akuity Promote, you can use this feature in Kargo v1.6+ offered on [Akuity Platform](https://akuity.io/akuity-platform).
+:::
+
+The `jira` promotion step provides comprehensive integration with Jira, allowing you to create, update, delete, and search for issues, manage comments, and track promotion workflows. This is particularly useful for maintaining traceability between your deployment processes and project management activities.
```
krancour: We've been very careful to avoid "deployment language" in Kargo so as to maintain a crisp delineation between the responsibilities of Kargo and the responsibilities of GitOps agents like Argo CD. Compared to deployments, promotion is much more "behind the scenes," putting the desired state in the desired state.

This wording change might apply in some other spots as well.

```suggestion
The `jira` promotion step provides comprehensive integration with Jira, allowing you to create, update, delete, and search for issues, manage comments, and track promotion workflows. This is particularly useful for maintaining traceability between your promotion processes and project management activities.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+| `deleteIssue.issueKey`       | `string`  | Y        | The Jira Issue Key (e.g., EXT-123).            |
+| `deleteIssue.deleteSubtasks` | `boolean` | N        | If true, all subtasks will be deleted as well. |
+
+#### Output
+
+This step does not produce any output values.
+
+#### Example
+
+This example deletes a Jira issue and all its subtasks when a deployment is rolled back.
```
krancour: Apart from [this](https://github.com/akuity/kargo/pull/4497/files#r2178475136), I'm also confused by the use of "rolled back" here. Up to now Kargo "rolls back" promotions by promoting forward to older, more stable Freight.

Not saying that won't ever change, but for now, this sentence leaves me with the false impression that rollouts are some kind of formal event _and_ that such an event triggers this step, which is also not the case.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+      credentials:
+        secretName: jira-credentials
+      deleteIssue:
+        issueKey: "${{ freightMetadata(ctx.targetFreight.name, 'jira-issue-key') }}"
+        deleteSubtasks: true
+```
+
+### Search Issues
+
+Searches for Jira issues using JQL (Jira Query Language).
```
krancour: Probably worth a link to JQL docs?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+| `searchIssue.jql`            | `string`  | Y        | The JQL query to search for issues.                                                           |
+| `searchIssue.expectMultiple` | `boolean` | N        | If true, expects multiple results. If false, expects single result and fails with >1 results. |
+| `searchIssue.fields`         | `array`   | N        | List of fields to include in search results.                                                  |
+| `searchIssue.expands`        | `array`   | N        | List of fields to expand in search results.                                                   |
+
+#### Output
+
+| Name    | Type     | Description                                                                 |
+| ------- | -------- | --------------------------------------------------------------------------- |
+| `issue` | `object` | The found Jira issue object containing all requested fields and expansions. |
```
krancour, line 188: It's not clear from this description what happens if `expectMultiple` was `true` and multiples were found. It leaves me guessing.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+          config:
+            apps:
+              - name: test-app
+                namespace: argocd
+                sources:
+                  - repoURL: https://github.com/company/app-config.git
+                    kustomize:
+                      images:
+                        - repoURL: public.ecr.aws/nginx/nginx
+                          tag: ${{ imageFrom("public.ecr.aws/nginx/nginx").Tag }}
```
krancour: This is [pseudo GitOps](https://docs.kargo.io/user-guide/patterns/#pseudo-gitops). I know it's an easy/accessible example, but if we're responsible about it, it probably needs some kind of disclaimer. It might be better to stick a `# ...` comment here to make it clear that other steps that weren't really relevant to the example were omitted.

May apply in other spots as well.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+        # Cleanup on failure
+        - as: on-failure-cleanup-issue
+          uses: jira
+          if: ${{ failure() }}
+          config:
+            credentials:
+              secretName: jira
+            deleteIssue:
+              issueKey: ${{ outputs['create-deployment-ticket'].key }}
+              deleteSubtasks: true
```
krancour: Seeing it in this context, I think updating the [standalone example I commented on](https://github.com/akuity/kargo/pull/4497/files#r2178488575) to more closely resemble this might resolve that comment.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+
+---
+apiVersion: kargo.akuity.io/v1alpha1
+kind: Stage
+metadata:
+  name: stage
```
krancour: Nit: Using `stage` as a Stage name invites confusion. I would suggest something different like `qa` or `uat`. Even `staging` would get the point across clearly while sidestepping any confusion.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+| `createIssue.description`    | `string` | N        | The description of the issue. Supports markdown formatting.                                                 |
+| `createIssue.adfDescription` | `object` | N        | ADF (Atlassian Document Format) content for complex formatting. Alternative to `description`.               |
+| `createIssue.issueType`      | `string` | N        | The type of issue to create (e.g., 'Bug', 'Task', 'Story').                                                 |
+| `createIssue.assigneeEmail`  | `string` | N        | Email of the user to assign the issue to.                                                                   |
+| `createIssue.labels`         | `array`  | N        | Labels to add to the issue for categorization.                                                              |
+| `createIssue.customFields`   | `object` | N        | Custom fields to set. Keys should match Jira custom field IDs.                                              |
+| `createIssue.issueAlias`     | `string` | N        | Override for the freight metadata key used to reference the created issue id. Defaults to `jira-issue-key`. |
```
krancour, line 51: I know feedback reflects on a detail of the actual implementation we're documenting here, but it feels like an odd choice to include "key" in the default key's name. If there were any chance of changing this, I'd be in favor of doing so.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Although I wonder if this is a realistic pattern...

Assuming many users will emulate what they see here, is deleting the Jira issue really a good way to respond to a promotion failure?

I see other Stages' promotion processes have similar cleanup steps. Do we really want to have an issue tracking successful promotion through several Stages and then throw all that information away because of a hiccup farther downstream?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+      steps:
+        # Wait for manual approval to proceed to staging
+        - as: wait-approval
+          uses: jira
+          config:
+            credentials:
+              secretName: jira
+            waitForStatus:
+              issueKey: ${{ freightMetadata(ctx.targetFreight.name, 'jira-issue-key') }}
+              expectedStatus: STAGED
```
krancour: The workflow is odd. We wait for a STAGED status in order to promote _into_ staging? That feels backwards. Same is true for prod further down. We wait for a RELEASED status before promoting into prod. These statuses sound very much like the thing that's being done has _already_ been done. I assume all these statuses are arbitrary or semi-arbitrary? (I haven't used Jira in a long time.) Can we just use different statuses to add clarity? Something like APPROVED_FOR_STAGING?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Nvm... I see now that subsequent Stages are cleaning up a _comment_ on failure.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+# Access the default issue key
+issueKey: ${{ freightMetadata(ctx.targetFreight.name, 'jira-issue-key') }}
+
+# Access a custom issue key (when issueAlias was used during creation)
+issueKey: ${{ freightMetadata(ctx.targetFreight.name, 'my-custom-alias') }}
+```
+
+### Example
+
+This comprehensive example demonstrates how to use the Jira promotion step across multiple stages in a promotion pipeline, tracking a single issue from creation through production deployment:
```
krancour: I really like this example because it ties everything else together nicely. I still think the example would be improved by avoiding "deployment" language. A deployment is rather low level and env-specific, while the issue that's created here and updated throughout the course of many promotions is much bigger picture, tracking the progression of artifacts through several envs. i.e. I favor changing some language, but structurally, I think the example is great!

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,651 @@
...
+      sources:
+        direct: true
+  promotionTemplate:
+    spec:
+      vars:
+        - name: imageRepo
+          value: public.ecr.aws/nginx/nginx
+      steps:
+        # Create initial deployment ticket
+        - as: create-deployment-ticket
```
krancour: I'm wondering what happens if/when Freight is _re_-promoted to this Stage. We get a branch new issue and the ID of that new issue replaces the ID of the old issue in the metadata?

I think there are all kinds of weird things that might happen.

1. Promote to test: New issue
1. Promote to stage: Update issue
1. Promote _something else_ to test and decide you don't like it
1. Promote test back to the older Freight: New issue? Do we now, essentially lose track of all the successful promotions that already happened downstream from test?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Making this example closer to what you do with this step in the larger/comprehensive example would clear things up a lot.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
jessesuen: Good suggestion, though I would just say `Enterprise` and not `Enterprise Only`.

jessesuen (commented):
Given that this is our first iteration of JIRA support, can we add a `Beta` label to this feature so that we have some freedom to change things after real-world usage and feedback?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Looking at the implementation, I think perhaps we ought to make "create issue" fall back on adding a comment to the existing issue instead if the Freight's metadata already has an issue under the key... but this would change how we might wish to deal with failures in promotions to test. Instead of cleaning up by _deleting_ the issue, we could clean up by adding a comment. Other stages that update the issue with a comment, likewise, could clean up by adding an additional comment (instead of one they already created)...

i.e. If we start to look at a single Jira ticket as a write-only ledger of activity recorded in the comments, non-linear flows will become a bit easier.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: We probably need a "beta" label as well.

krancour: > just say Enterprise

Or "Akuity Promote?" It's not entirely clear yet what is the correct/consistent way to reference the commercial offering from these OSS docs.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Actually... another concern popped into my head. Other features using the "cluster secrets namespace" use that namespace as a repository for Secrets referenced by, as the name implies, _cluster-scoped things._

The way it's being used here seems to be quite a departure from that. It seems this is being used in a way that is more consistent with what "global credential namespaces" had been intended for -- i.e. sharing across Projects.

As it stands now, it seems referencing cluster-scoped secrets from a Project-scoped resource creates a possible vulnerability by allowing any Project to access Secrets belonging to the "Kargo admin" (my shorthand for whoever maintains cluster-scoped Kargo resources).

gdsoumya: Actually the secret does allow specifying global credential ns too, but because we had this convo about not using global cred ns in some places I thought it's best to support all 3 types of ns we have. project, global and cluster secrets. I only put cluster secret ns as that seemed more suitable but I will add global creds ns for the doc too. 

I can see how jira creds could be part of cluster secret ns as users might want to reuse it with their jira webhooks in the future. As for security unlike a generic secret read this expects a specific jira cred with very specific fields so it cannot be misused. We should keep cluster secret ns as an option too imo.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Yes, the stages only cleanup the resources they create so only the first stage deletes the issues while others delete the comments so that we don't have any dangling entries in jira.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: jira itself refers to the data we are storing here as `issue key` so I think it makes sense. We aren't storing the whole issue content just the identifier which is referred to as `issue key`

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: We can make the changes, the statuses are user defined columns of the project board. like TODO, IN PROGRESS, DONE. Usually that's how it is a state that reflects the current status of the ticket, so how I thought this would work is test stage creates the ticket with TESTING status(default initial status), then once it's successfully tested a  user manually accepts it to be STAGED. And then similarly once the staged status is successful(issue comment is made and labels are updated) the user marks the ticket ready for RELEASE. You are right that the promotion is the actual action of promoting to that status but the project board is more of a state the user assumes the ticket to be in before they go ahead and perform promo/verification etc. It' is a very manual process, if the promo fails we do not put the comments/labels needed so user would know something failed in the env the ticket is already in (STAGE, PROD etc.) and then perform whatever manual fix is needed and retry promo.

gdsoumya: Let me know what you think of this ^^ if you think we need to still change the status names I'll do that.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: > Promote test back to the older Freight: New issue? Do we now, essentially lose track of all the successful promotions that already happened downstream from test?

That would indeed happen, I initially had logic to error out the step if the freight already has a jira key, user would have to set an override flag in the step to get past the error and manually force the overwrite. But then I thought if we did do this it won't really change the fact that user wants to overwrite the jira key or the fact that the user might want to re-promote so I removed that logic and always override now.

I think this can be dealt with promo step conditions like `if freightMetadata(ctx.targetFreight.name, <issue-key>) exists skip issue creation step` for eg. Not sure how to exactly frame this w.r.t to kargo for us but I would prefer to not include any custom handling in the low level step action itself and let user figure out a way they prefer to deal with this. Either maybe they don't create a ticket with promo step and just use comment step to update existing ticket. Or they use the conditional approach to skip re-creating issue.

I will though add this behavior to the create action in the doc so user is aware of it, but let me know what you think of the above approach.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: refactored all deployment -> promotion

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Updated.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: refactored

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
Marvin9: Screenshot 2025-07-02 at 12 55 17 PM

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: changed rollback to failure

gdsoumya: updated example to be similar to comprehensive eg.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Updated comment on the step to specify it's for demo and not ideal for practical use

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: renamed to uat

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Updated the description to include cluster secret ns and global ns references.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
kencochrane: How about this.
> This promotion step is only available with Akuity Promote version v1.6 and above on [Akuity Platform](https://akuity.io/akuity-platform).

docs/docusaurus.config.js
```diff
@@ -32,6 +33,29 @@ const config = {
...
+                  beta: tags.isBeta(item),
+                  enterprise: tags.isEnterprise(item)
+                };
+
+                return item;
+              });
+            }
+
+            return addBadges(sidebarItems);
+          },
```
krancour, line 58: Wow! I wouldn't have had the patience to figure out how to do this the right way in Docusaurus. This is amazing!

docs/src/css/custom.css
```diff
@@ -71,3 +71,35 @@ main.landing-page {
...
+  display: inline-block;
+  padding: 0.1em 0.5em;
+  border-radius: 0.3em;
+  vertical-align: middle;
+  font-weight: normal;
+  font-size: small;
+}
+
+.enterprise::after {
+  content: "Enterprise";
```
krancour:
```suggestion
  content: "Enterprise";
```

@kencochrane says we can use the term "Kargo Promote" (everywhere, including these labels) instead of "Enterprise."

I think it would be fine to leave the class name as `enterprise`. That insulates us from extra work if the product name changes.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Gotcha. I agree completely that if Jira calls it an "issue key" then aligning with that terminology makes complete sense. We did something similar in webhook receivers where there is something called "secret-key" it's _not_ the key in the map. It's what the external platform _calls_ the shared key used for symmetric encryption.

💯

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+| `credentials.secretName` | `string` | Y        | Name of the `Secret` containing the Jira credentials.                                                                        |
+| `credentials.namespace`  | `string` | N        | Namespace containing the credentials `Secret`. This can be the Project's namespace, the cluster secrets namespace (`kargo-cluster-secrets`) or any of the configured global credential namespaces. If empty, the Project's namespace is assumed.|
+
+The referenced `Secret` should contain the following keys:
+
+- `domain`: The domain of your Jira instance or Jira api (e.g., `https://yourcompany.atlassian.net`)
+- `username`: Your Jira username or email
+- `password`: Your Jira API token or password
+
+## Issue Management
```
krancour, line 33: More of a question than a comment. Does Jira use the terms "issue" and "ticket" completely interchangeably? It seems we're using "issue" in some places, but "ticket" in others (mostly in examples).

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+
+#### Output
+
+| Name  | Type     | Description                                           |
+| ----- | -------- | ----------------------------------------------------- |
+| `key` | `string` | The key/id of the created Jira issue (e.g., EXT-123). |
+
+#### Example
+
+This example creates a new Jira issue to track a promotion, assigns it to a team member, and adds relevant labels.
```
krancour, line 61: Just a big thank you for the effort that clearly went into switching from "deployment language" to "promotion language."

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+steps:
+- uses: jira
+  as: create-promotion-issue
+  config:
+    credentials:
+      secretName: jira-credentials
+    createIssue:
+      projectKey: DEPLOY
+      summary: "Deploy ${{ imageFrom(vars.imageRepo).Tag }} to ${{ ctx.stage }}"
+      description: "Deploying ${{ imageFrom(vars.imageRepo).RepoURL }}:${{ imageFrom(vars.imageRepo).Tag }} to ${{ ctx.stage }} environment. Promotion ID: ${{ ctx.promotion }}. Freight: ${{ ctx.targetFreight.name }}."
```
krancour: These references to "deployment" seem to be stragglers.

Here's a suggestion that fixes it, but I'm not sure if the project name is realistic or not. Feel free to alter my suggestion further:

```suggestion
      projectKey: GUESTBOOK
      summary: "Promote ${{ imageFrom(vars.imageRepo).Tag }} to ${{ ctx.stage }}"
      description: "Promoting ${{ imageFrom(vars.imageRepo).RepoURL }}:${{ imageFrom(vars.imageRepo).Tag }} to ${{ ctx.stage }} environment. Promotion ID: ${{ ctx.promotion }}. Freight: ${{ ctx.targetFreight.name }}."
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+- uses: jira
+  config:
+    credentials:
+      secretName: jira-credentials
+    updateIssue:
+      issueKey: DEPLOY-123
+      status: "IN PROGRESS"
+      summary: "Deploy ${{ imageFrom(vars.imageRepo).Tag }} to ${{ ctx.stage }} - IN PROGRESS"
+      addLabels:
+      - deploying
```
krancour: Feel free to tweak further:

```suggestion
      issueKey: GUESTBOOK-123
      status: "IN PROGRESS"
      summary: "Promote ${{ imageFrom(vars.imageRepo).Tag }} to ${{ ctx.stage }} - IN PROGRESS"
      addLabels:
      - in-progress
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+### Search Issues
+
+Searches for Jira issues using JQL ([Jira Query Language](https://support.atlassian.com/jira-service-management-cloud/docs/use-advanced-search-with-jira-query-language-jql/)).
+
+#### Configuration
+
+| Name                         | Type      | Required | Description                                                                                   |
+| ---------------------------- | --------- | -------- | --------------------------------------------------------------------------------------------- |
+| `searchIssue.jql`            | `string`  | Y        | The JQL query to search for issues.                                                           |
+| `searchIssue.expectMultiple` | `boolean` | N        | If true, expects multiple results and returns the first matching result. If false, expects single result and fails with >1 results. |
```
krancour, line 180: Just a note that this is why I've _completely_ given up on trying to make tables look pretty in the source. Change one row and nothing lines up anymore unless you commit to changing all other rows and that juice isn't worth the squeeze. 🤷‍♂️ 

No action needed.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Much clearer now!

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+
+```yaml
+steps:
+- uses: jira
+  as: find-open-promotions
+  config:
+    credentials:
+      secretName: jira-credentials
+    searchIssue:
+      jql: 'project = DEPLOY AND status != "Done" AND labels = "${{ ctx.stage }}-promotion" AND created >= -7d'
```
krancour: Feel free to change for consistency with other suggestions if you tweak them as well.

```suggestion
      jql: 'project = GUESTBOOK AND status != "Done" AND labels = "${{ ctx.stage }}-promotion" AND created >= -7d'
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+      expectMultiple: true
+      fields:
+      - summary
+      - status
+      - assignee
+      - created
+      expands:
+      - changelog
+# Use search results in subsequent steps to notify team
+# Note: This is just an example of using search outputs and may not be syntactically valid
```
krancour, line 212: Is it specifically the JQL that might not be syntactically valid? I assume so, but could we be specific about that so that no one's left guessing about exactly which part of the example might be invalid?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+description: Integrates with Jira to manage issues, comments, and track promotion workflows.
+---
+
+<span class="tag enterprise"></span>
+<span class="tag beta"></span>
+# `jira`
+
+:::info Enterprise Feature
+This promotion step is only available in Akuity Promote, you can use this feature in Kargo v1.6+ offered on [Akuity Platform](https://akuity.io/akuity-platform).
+:::
```
krancour: Taken directly from @kencochrane's suggestion [above](https://github.com/akuity/kargo/pull/4497/files#r2180939930):

```suggestion
:::info
This promotion step is only available with Akuity Promote version v1.6 and above on [Akuity Platform](https://akuity.io/akuity-platform).
:::
```

Note we may wish to make "Akuity Promote" a link in the future as well when it has an official product page.

@kencochrane if it does already, lmk.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+      - deploying
+      - "${{ ctx.stage }}-promotion"
+      customFields:
+        customfield_10000: "${{ ctx.stage }} Environment"
+        customfield_10001: "${{ ctx.promotion }}"
+```
+
+### Delete Issue
+
+Deletes a Jira issue and optionally its subtasks.
```
krancour:
```suggestion
Deletes a Jira issue and, optionally, its subtasks.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+| `updateIssue.issueType`      | `string` | N        | Updated issue type.                                               |
+| `updateIssue.assigneeEmail`  | `string` | N        | Email of the user to assign the issue to.                         |
+| `updateIssue.status`         | `string` | N        | Status to set for the issue (e.g., 'IN PROGRESS', 'DONE').        |
+| `updateIssue.addLabels`      | `array`  | N        | Labels to add to the issue.                                       |
+| `updateIssue.removeLabels`   | `array`  | N        | Labels to remove from the issue.                                  |
+| `updateIssue.customFields`   | `object` | N        | Custom fields to update.                                          |
+
+#### Output
+
+This step does not produce any output values.
```
krancour: "Output values" sounds redundant, although it's not a hill I would die on. Take it or leave it. Applies in other places as well.

```suggestion
This step does not produce any output.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+apiVersion: kargo.akuity.io/v1alpha1
+kind: Stage
+metadata:
+  name: test
+  namespace: kargo-proj
+spec:
+  requestedFreight:
+  - origin:
+      kind: Warehouse
+      name: nginx
```
krancour, line 377: I suggested the project name `GUESTBOOK` in a few places, so I wonder if we should make the Warehouse name and the image repo it subscribes to match.

It _feels_ slightly better imho because the idea of moving the latest revision of homegrown software through a pipeline probably resonates with most readers more readily than the idea of moving a piece of infra through a pipeline (even though that's a perfectly valid thing to do).

I intend to update the quickstart similarly for the same reason.

Just a thought. Up to you if you think this is worth it or not.


```suggestion
      name: nginx
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+            secretName: jira
+          deleteIssue:
+            issueKey: ${{ outputs['create-promotion-ticket'].key }}
+            deleteSubtasks: true
+
+---
+apiVersion: kargo.akuity.io/v1alpha1
+kind: Stage
+metadata:
+  name: uat
```
krancour, line 484: Much less prone to confusion with `Stage`. Appreciate it!

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+    sources:
+      stages:
+        - test
+  promotionTemplate:
+    spec:
+      vars:
+      - name: imageRepo
+        value: public.ecr.aws/nginx/nginx
+      steps:
+        # Wait for manual approval to proceed to staging
```
krancour: Just a straggler:

```suggestion
        # Wait for manual approval to proceed to UAT
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+            - name: uat-app
+              namespace: argocd
+              sources:
+              - repoURL: https://github.com/company/app-config.git
+                kustomize:
+                  images:
+                  - repoURL: ${{ vars.imageRepo }}
+                    tag: ${{ imageFrom(vars.imageRepo).Tag }}
+
+      # Update ticket with staging progress
```
krancour:
```suggestion
      # Update ticket with UAT progress
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+        value: public.ecr.aws/nginx/nginx
+      steps:
+      # Create initial promotion ticket
+      - as: create-promotion-ticket
+        uses: jira
+        config:
+          credentials:
+            secretName: jira
+          createIssue:
+            projectKey: KD
```
krancour:
```suggestion
            projectKey: GUESTBOOK
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+            issueType: Task
+            summary: "Deploy Release ${{ imageFrom(vars.imageRepo).Tag }}"
```
krancour:
```suggestion
            summary: "Promoting Release ${{ imageFrom(vars.imageRepo).Tag }}"
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+                  text: "${{ ctx.project }}"
+                  marks:
+                  - type: code
+            labels:
+            - "automated-promotion"
+            - "env-${{ ctx.stage }}"
+            - "release-${{ imageFrom(vars.imageRepo).Tag }}"
+            - "project-${{ ctx.project }}"
+
+      # Update application, directly updates app for demonstration. Not ideal for practical purposes.
```
krancour: 👍 Appreciate the disclaimer.

```suggestion
      # Update the Argo CD Application directly. Not ideal for practical purposes.
```

Same suggestion in other place _if_ you want it.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+This promotion step supports various operations including issue management, comment handling, and status tracking, making it a powerful tool for promotion workflows that require coordination with project management systems.
+
+## Credentials Configuration
+
+All Jira operations require proper authentication credentials stored in a Kubernetes `Secret`.
+
+| Name                     | Type     | Required | Description                                                                                                                  |
+| ------------------------ | -------- | -------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| `credentials.secretName` | `string` | Y        | Name of the `Secret` containing the Jira credentials.                                                                        |
+| `credentials.namespace`  | `string` | N        | Namespace containing the credentials `Secret`. This can be the Project's namespace, the cluster secrets namespace (`kargo-cluster-secrets`) or any of the configured global credential namespaces. If empty, the Project's namespace is assumed.|
```
krancour: In an offline conversation with @jessesuen, he suggested simply not documenting this field for now. Without doing so, no one will be aware of it (unless they're following along right now 🤣). Since it default's to the Project's namespace, it's fine for people to omit it. This would buy us some time to nail down possible implementation changes in an upcoming release, essentially without breakage.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+      steps:
+        # Wait for manual approval to proceed to staging
+      - as: wait-approval
+        uses: jira
+        config:
+          credentials:
+            secretName: jira
+          waitForStatus:
+            issueKey: ${{ freightMetadata(ctx.targetFreight.name, 'jira-issue-key') }}
+            expectedStatus: UAT
```
krancour, line 508: Since I had some concern about re-promotions and generally non-linear workflows... does this effectively make re-promotion to UAT impossible if the Freight has already progressed beyond UAT and may now have a status such as PROD?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Marking this as resolved and we can move the convo to [here](https://github.com/akuity/kargo/pull/4497#discussion_r2181199754).

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -0,0 +1,659 @@
...
+
+```yaml
+steps:
+- uses: jira
+  config:
+    credentials:
+      secretName: jira-credentials
+    waitForStatus:
+      issueKey: "${{ freightMetadata(ctx.targetFreight.name, 'change-request-key') }}"
+      expectedStatus: "Approved"
```
krancour, line 332: Are Jira statuses all caps?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: Yup. I understand your point. There's sort of a natural impedance between Jira tracking current state and using that state as a trigger for (other) state changes (in Kargo).

This is a detail we probably don't need to obsess over at the moment. Resolved as far as I am concerned.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: > Not sure how to exactly frame this w.r.t to kargo for us but I would prefer to not include any custom handling in the low level step action itself and let user figure out a way they prefer to deal with this.

Ya... I agree with us not getting overly opinionated about how to deal with this sort of thing. Let users decide. At the same time, I think we should _also_ avoid requiring too much boilerplate from users that must be _just so_ in order to escape the unconditional new ticket footgun.

But if we're calling the feature beta, I think we'll learn more about realistic usage patterns by using it and by observing what kind of things other users end up doing with it that we may not even have imagined.

I think we could let this go for now.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Actually Jira refers to their issues as "Issue" but I guess ticket is more common word that we use with Jira. I'll update to use issue instead of ticket.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Oh no it's the http step to notify I guess that's just an example. I copied it from another example I think not sure if the syntax is correct for the slack endpoint call. The jira query is valid

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: No, the step looks at the current state of the issue, there's no historical state in jira data so if it doesn't match then it won't be possible to re-promote. though it should be simple to just move the ticket to the old status and then change it back to get past this.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: our logic is actually case insensitive, so doesn't matter how the user names their status it will work.

docs/src/css/custom.css
gdsoumya: @Marvin9 ^^ 1 last time

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Good idea, removed

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: I changed all project key to PROMOTE, as guestbook seemed very specific.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Leaving it as nginx as I changed the project name to PROMOTE instead of guestbook.

docs/docusaurus.config.js
gdsoumya: Yes! @Marvin9 did his magic.

docs/src/css/custom.css
Marvin9: Done

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: > seemed very specific.

Specific had been my intent, but I am inferring now that perhaps we were viewing the hypothetical project from different angles.

I saw the project as being app-specific and oriented around all things having to do with that app. This could include features, support tickets, etc., and tracking promotions through the stages of the app's pipeline is just one type of issue tracked in that project.

But perhaps you see the `PROMOTE` project as not being app-specific, rather a project dedicated to tracking promotions through the stages of many/all apps' pipelines.

If I'm correctly understanding how you saw this, then 👍. I think both approaches are probably valid and it's a preference or organizational choice that might reflect other choices such as using or not use monorepos, central/siloed devops org, etc., etc.

I think we're all good.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour: If something's made it all the way through to prod and I needed to re-promote it to a lesser environment, like `test` (which had since moved on to newer Freight), I'd hate having to put the status of my issue back to something that no longer reflects that it was cleared for / has made it to prod.

I'm wondering if there's a cleaner way to do this. Issues can only have one status, but they can have many labels...

I wonder if it might be better to demonstrate a pattern of waiting for a particular _label_ to be present. However, I know that would involve adding a new feature to the step, so I think what we're showing here might be the best we can do for now, and I will open an issue in the appropriate repo to track the idea of a feature that waits on the presence (or absence) of a certain label. (Maybe something generic where an expression can be used to examine arbitrary properties of the issue might be even more useful.)

I'm fine with this, as is, given the constraints we're working with. We can revisit in a future release.

krancour (approved):

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
gdsoumya: Yup you are right I did think of this as a generic project board for multiple different apps.

gdsoumya:
@krancour I think we need to backport this right or else it won't make it to 1.6 release docs?

Issue #4488, opened 2025-06-24
add webhook receiver for artifactory


krancour:
Broken out from #4150

Add support for refreshing Warehouses in response to artifacts being pushed to Artifactory.

PR #4498, opened 2025-06-25
feat: artifactory(jfrog) webhook receiver

fuskovic:
Part of #4488


api/v1alpha1/project_config_types.go
```diff
@@ -120,6 +120,10 @@ type WebhookReceiverConfig struct {
 	// Quay contains the configuration for a webhook receiver that is compatible
 	// with Quay payloads.
 	Quay *QuayWebhookReceiverConfig `json:"quay,omitempty" protobuf:"bytes,4,opt,name=quay"`
+	// Artifactory contains the configuration for a webhook receiver that is compatible
+	// with JFrog Artifactory payloads.
```
krancour:
```suggestion
	// Artifactory contains the configuration for a webhook receiver that is
	// compatible with JFrog Artifactory payloads.
```

api/v1alpha1/project_config_types.go
```diff
@@ -120,6 +120,10 @@ type WebhookReceiverConfig struct {
...
+	Artifactory *ArtifactoryWebhookReceiverConfig `json:"artifactory,omitempty" protobuf:"bytes,7,opt,name=artifactory"`
+	// Path is the path to the
```
krancour: Looks misplaced?

api/v1alpha1/project_config_types.go
```diff
@@ -219,6 +223,26 @@ type QuayWebhookReceiverConfig struct {
...
+	// receivers, the referenced Secret must be in the same namespace as the
+	// ProjectConfig.
+	//
+	// For cluster-scoped webhook receivers, the referenced Secret must be in the
+	// designated "cluster Secrets" namespace.
+	//
+	// The Secret's data map is expected to contain a `secret-token` key whose value is
+	// the shared secret used to authenticate the webhook requests sent by
+	// JFrog Artifactory. For more information please refer to the
+	// JFrog Artifactory documentation:
```
krancour:
```suggestion
	// The Secret's data map is expected to contain a `secret-token` key whose
	// value is the shared secret used to authenticate the webhook requests sent
	// by JFrog Artifactory. For more information please refer to the JFrog
	// Artifactory documentation:
```

api/v1alpha1/project_config_types.go
```diff
@@ -120,6 +120,10 @@ type WebhookReceiverConfig struct {
 	// Quay contains the configuration for a webhook receiver that is compatible
 	// with Quay payloads.
 	Quay *QuayWebhookReceiverConfig `json:"quay,omitempty" protobuf:"bytes,4,opt,name=quay"`
+	// Artifactory contains the configuration for a webhook receiver that is compatible
+	// with JFrog Artifactory payloads.
+	Artifactory *ArtifactoryWebhookReceiverConfig `json:"artifactory,omitempty" protobuf:"bytes,7,opt,name=artifactory"`
```
krancour: Note: There are PRs open currently for three new receivers and all three are competing for field index `7`. So this is just a reminder to merge the latest from main and update this index as needed.

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+			predicate: func(cfg kargoapi.WebhookReceiverConfig) bool {
+				return cfg.Artifactory != nil
+			},
+			factory: newArtifactoryWebhookReceiver,
+		},
+	)
+}
+
+// artifactoryWebhookReceiver is an implementation of WebhookReceiver that handles
+// inbound webhooks from Jfrog Artifactory.
```
krancour:
```suggestion
// artifactoryWebhookReceiver is an implementation of WebhookReceiver that
// handles inbound webhooks from Jfrog Artifactory.
```

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+type artifactoryWebhookReceiver struct {
+	*baseWebhookReceiver
+}
+
+// newArtifactoryWebhookReceiver returns a new instance of artifactoryWebhookReceiver.
```
krancour:
```suggestion
// newArtifactoryWebhookReceiver returns a new instance of
// artifactoryWebhookReceiver.
```

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+	return &artifactoryWebhookReceiver{
+		baseWebhookReceiver: &baseWebhookReceiver{
+			client:     c,
+			project:    project,
+			secretName: cfg.Artifactory.SecretRef.Name,
+		},
+	}
+}
+
+// GetDetails implements WebhookReceiver.
```
krancour:
```suggestion
// getReceiverType implements WebhookReceiver.
```

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+		ctx := r.Context()
+		logger := logging.LoggerFromContext(ctx)
+
+		token, ok := a.secretData[artifactorySecretDataKey]
+		if !ok {
+			xhttp.WriteErrorJSON(w, nil)
+			return
+		}
+
+		sig := r.Header.Get("x-jfrog-event-auth")
```
krancour: We use this in a enough places in the tests that we may as well make a constant for this.

This is a low-hanging bit of fruit that we can fix in other receivers as well. (I'll open a separate PR. No need to pork barrel here.)

Also, perusing JFrog docs, I sometimes see `x-jfrog-event-auth` and at other times, I see `X-Jfrog-Event-Auth`. I sent myself a webhook to see what I actually received and it was `X-Jfrog-Event-Auth`. Headers are case insensitive (when accessed this way, at least), but I think it would still be best to make the case match what we know we will actually receive.

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+		if sig == "" {
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(
+					errors.New("missing signature"),
+					http.StatusUnauthorized,
+				),
```
krancour:
```suggestion
				xhttp.Error(errors.New("missing signature"), http.StatusUnauthorized),
```

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+		if !hmac.Equal([]byte(sig), []byte(computedSig)) {
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(
+					errors.New("unauthorized"),
+					http.StatusUnauthorized,
+				),
+			)
+			return
+		}
```
krancour, line 117: Would the signature validation from GitHub, which was generic enough that we've reused it in some other receivers, work here as well?

Note that I have not actually tried this:

```suggestion
		// Note: github.com/google/go-github/v71/github has a great implementation
		// of HMAC signature validation that isn't GitHub-specific, so we've opted
		// to use it here for Artifactory as well.
		if err := gh.ValidateSignature(sig, requestBody, signingKey); err != nil {
			xhttp.WriteErrorJSON(
				w,
				xhttp.Error(errors.New("unauthorized"), http.StatusUnauthorized),
			)
			return
		}
```

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+
+		logger.Info("Received Artifactory webhook",
+			"raw event", string(requestBody),
+		)
```
krancour: Info's a bit high a level to be logging this. If Kargo is installed with defaults, info-level logging is enabled and all the request bodies are going to show up in logs. If I were to keep this at all, I'd probably knock it all the way down to trace level.

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+
+		var payload artifactoryEvent
+		if err := json.Unmarshal(requestBody, &payload); err != nil {
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(errors.New("invalid request body"),
+					http.StatusBadRequest,
+				),
```
krancour:
```suggestion
				xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
```

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+			)
+			return
+		}
+
+		switch payload.EventType {
```
krancour: If we're only accepting one specific event type, I would not use a switch here. I'd move the default case up to a conditional above here with an early return. Doing so keeps the happy path below indented one level less, which helps combat the phenomenon known as "cylcomatic complexity."

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+		case "pushed":
+			parsed, err := url.Parse(payload.Origin)
+			if err != nil {
+				xhttp.WriteErrorJSON(
+					w,
+					xhttp.Error(
+						fmt.Errorf("failed to parse origin URL: %w", err),
+						http.StatusBadRequest,
+					),
```
krancour: I might just leave it at this since this is what we've done thus far for webhook payloads that are no good for any reason. Really, this should never happen with legitimate webhooks delivered from the corresponding platform. If it does happen, it was almost certainly on account of a failed attempt to spoof a webhook and we don't need to worry about giving lots of detail.

```suggestion
					xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
```

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+				payload.Data.ImageName,
+			)
+			if err != nil {
+				xhttp.WriteErrorJSON(
+					w,
+					xhttp.Error(
+						fmt.Errorf("failed to construct repository URL: %w", err),
+						http.StatusBadRequest,
+					),
+				)
```
krancour: This shouldn't happen and if it somehow does, I think we ought to count this as an internal error. To have gotten this far, the request was good.

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+			// Payloads from Artifactory contain no information about media type, so we
+			// normalize the URL BOTH as if it were an image repo URL and as if it were
+			// a chart repository URL. These will coincidentally be the same, but by
+			// doing this, we safeguard against future changes to normalization logic.
+			// Note: The refresh logic will dedupe the URLs, so this does not create
+			// the possibility of a double refresh.
+			repoURLs := []string{
+				image.NormalizeURL(repoURL),
+				helm.NormalizeChartRepositoryURL(repoURL),
+			}
```
krancour: Correct approach. Just curious if you're tried pushing a helm chart to Artifactory to verify how this all plays out.

internal/webhook/external/artifactory.go
```diff
@@ -0,0 +1,196 @@
...
+						http.StatusBadRequest,
+					),
+				)
+				return
+			}
+			repoURL, err := url.JoinPath(
+				parsed.Hostname(),
+				payload.Data.RepoKey,
+				payload.Data.ImageName,
+			)
```
krancour: This looks like it might not be universally accurate.

Artifactory repos seem hierarchical in ways that most repos are not.

I created a repo called `test-repo` and then did the following:

```shell
docker pull busybox:latest
docker tag busybox:latest trial30g6c2.jfrog.io/test-repo/foo/bar/busybox:latest
docker push trial30g6c2.jfrog.io/test-repo/foo/bar/busybox:latest
```

The webhook payload I received was:

```
{
  "domain": "docker",
  "event_type": "pushed",
  "data": {
    "repo_key": "test-repo",
    "path": "foo/bar/busybox/latest/manifest.json",
    "name": "manifest.json",
    "sha256": "2278daae78b8899e6c7192a8a6f49763b9e0cde6fcf94de34f115df4401b955f",
    "size": 527,
    "image_name": "busybox",
    "tag": "latest",
    "platforms": [],
    "image_type": "docker"
  },
  "subscription_key": "test-webhook",
  "jpd_origin": "https://trial30g6c2.jfrog.io",
  "source": "jfrog/jfrt@01jzq1wpac4msw1djjwqwh09w0"
}
```

The logic above would determine the URL to be `trial30g6c2.jfrog.io/test-repo/busybox`, which is incorrect.

It may be necessary to do something like split the value of the `path` field on `/`, discard the last part, then do something like `url.JoinPath(parsed.Hostname(), payload.Data.RepoKey, pathParts...)`

internal/webhook/external/artifactory.go
krancour: Actually... now that I've tried this...

It seems the value of `data.image_type` is `docker` for an image and `oci` otherwise.

internal/webhook/external/artifactory.go
krancour: Actually... small plot twist...

I was playing more with Artifactory and while it was very easy to fire of `pushed` events in response to an image being pushed to an image repo or a chart being pushed to an (OCI) chart repo (underneath, they're _both_ just OCI), it was a bit harder to figure out how to fire off any kind of event when a chart was pushed to a legacy (HTTP/S) chart repository. I eventually came to understand that `Artifact` --> `Artifact was deployed` was the right event if I want to hear about pushes to a legacy chart repo. (imho, "deployed" is an odd way to describe something being _published_ to a repository. To me, "deployed" sounds much more like an artifact being put into use somewhere, but I digress.)

This is where things get a little bit more interesting. You can actually fire off webhooks for `Artifact` --> `Artifact was deployed` events from repos of any kind. i.e. Works for all three of: image repos, OCI chart repos, and legacy chart repos.

_I briefly considered that we might want to support `domain: artifact` with `event_type: deployed`._

Here's what a `domain: docker` `event_type: pushed` event looks like:

```json
{
  "domain": "docker",
  "event_type": "pushed",
  "data": {
    "repo_key": "test-repo",
    "path": "foo/bar/busybox/latest/manifest.json",
    "name": "manifest.json",
    "sha256": "2278daae78b8899e6c7192a8a6f49763b9e0cde6fcf94de34f115df4401b955f",
    "size": 527,
    "image_name": "busybox",
    "tag": "latest",
    "platforms": [],
    "image_type": "docker"
  },
  "subscription_key": "test-webhook",
  "jpd_origin": "https://trial30g6c2.jfrog.io",
  "source": "jfrog/jfrt@01jzq1wpac4msw1djjwqwh09w0"
}
```

All the info we need is present. And as noted in another comment, `data.image_type` can even help us distinguish between images and charts. `docker` == image. `oci` == chart.

In contrast, a `domain: artifact` `event_type: deployed` event looks like:

```json
{
  "domain": "artifact",
  "event_type": "deployed",
  "data": {
    "repo_key": "test-repo",
    "path": "foo/bar/busybox/latest/manifest.json",
    "name": "manifest.json",
    "sha256": "2278daae78b8899e6c7192a8a6f49763b9e0cde6fcf94de34f115df4401b955f",
    "size": 527
  },
  "subscription_key": "test-webhook",
  "jpd_origin": "https://trial30g6c2.jfrog.io",
  "source": "jfrog/jfrt@01jzq1wpac4msw1djjwqwh09w0"
}
```

As you can see, it's quite similar, but lacks a few critical pieces of information. We're missing `image_name` and `tag`. We _could_ extract that info from the `path`. We're also missing `image_type`, which makes it harder to distinguish between images and charts. We also have no way of knowing whether the repo this event originated from was an OCI one or a legacy chart repo.

So... in the end, I think supporting `domain: docker` with `event_type: pushed` only was the right thing to do. 

Why am I bringing all this up then?

1. I think to cover our bases, we should be looking at `domain` _in addition to_ `event_type` and rejecting anything that isn't _both_ `domain: docker` and `event_type: pushed` as unsupported. This way if _other (future) types of repos_ ever issue `pushed` events, we won't act on them.

1. I think people who are more familiar with Artifactory than we are and have been using it for some time are possibly already accustomed to using `Artifact` --> `Artifact was deployed` webhooks (i.e. `domain: artifact` with `event_type: deployed`). For reasons explained above, we'd be better off supporting only `domain: docker` with `event_type: pushed`. This impacts docs (I'll link to here from the corresponding doc PR):

    * We should clearly state that we support webhooks _only_ from image repos and (OCI) chart repos and we _do not_ support webhooks from any other kind of repos, _including_ legacy (HTTP/S) chart repos. (I'm ok with this because legacy chart repos are increasingly falling out of fashion in favor of OCI anyway.)

    * Even though your directions already explain exactly what event to subscribe to, we should be explicit that we do not respond to `domain: artifact` with `event_type: deployed`.

internal/webhook/external/artifactory.go
fuskovic: It doesn't work here because it expects the prefix `<hash-algo>=` to be prepended.

internal/webhook/external/artifactory.go
fuskovic: Updated the conditional to ensure events match both constraints (`domain: docker` + `event_type: pushed`) and I've updated the docs PR to be explicit about what we don't support and respond to.

internal/webhook/external/artifactory.go
fuskovic: This is why I had to use a different helper to sign artifactory requests in the tests instead of the existing one

internal/webhook/external/artifactory_test.go
```diff
@@ -0,0 +1,313 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/indexer"
+)
+
+const artifactorySigningKey = "mysupersecrettoken"
```
krancour: There's a `testSigningKey` in `common_test.go` that we can use instead of this. Your `signWithArtifactory()` is, in fact, already using `testSigningKey`. As it stands, I think we lucked into things working because `testSigningKey` and `artifactorySigningKey` happen to have the same value.

Most other receiver tests are using the common `testSigningKey` already. For the few that aren't we should probably fix that. (Separate PR.)

```suggestion
```

internal/webhook/external/common_test.go
```diff
@@ -16,3 +16,9 @@ func sign(content []byte) string {
 		hex.EncodeToString(mac.Sum(nil)),
 	)
 }
+
+func signWithArtifactory(content []byte) string {
+	mac := hmac.New(sha256.New, []byte(testSigningKey))
+	_, _ = mac.Write(content)
+	return hex.EncodeToString(mac.Sum(nil))
+}
```
krancour: Since this signing function is, at least for now, unique to Artifactory, I'd consider moving it into `artifactory_test.go`.

Alternatively, and perhaps better, you could leave it here, but rename it to something like `signWithoutAlgoPrefix()` and then it seems we could potentially reuse it in other future cases and being here in `common_test.go` would be fine because it would no longer seem so Artifactory-specific.

krancour:
@fuskovic thank for all the very hard work on this. There are two very tiny things to resolve, plus you probably need to merge the latest from main and re-run codegen in order to resolve merge conflicts in generated code and then this LGTM.

I've tested this extensively e2e and it works great!

Preemptively LGTM'ing!

krancour (approved):

PR #4507, opened 2025-06-25
feat(chart): make it possible to trust cert when not using cert-manager

ricoapon:
# Problem
When not using `cert-manager`, the certificate from the user is not automatically trusted. The value `caBundle` is not set in the Helm chart.

See #4389 for the discussion.

# Solution

Set `caBundle` in `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration` only when `.Values.webhooksServer.tls.selfSignedCert` is set to false.


charts/kargo/templates/kubernetes-webhooks-server/webhooks.yaml
```diff
@@ -1,4 +1,12 @@
 {{- if .Values.webhooks.register }}
+{{/* Resolve caBundle only when secret with certificate is present on the cluster. */}}
+{{- $caBundle := "" }}
+{{- if not .Values.webhooksServer.tls.selfSignedCert }}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace "kargo-webhooks-server-cert" | default dict }}
+  {{- if and (hasKey $secret "data") (hasKey $secret.data "tls.crt") }}
+    {{- $caBundle = get $secret.data "tls.crt" }}
```
ricoapon: I choose to not throw an error if the lookup failed. If you would throw an error, then running `helm template` locally would always fail. That would be very annoying as a developer.

charts/kargo/templates/kubernetes-webhooks-server/webhooks.yaml
```diff
@@ -17,6 +25,9 @@ webhooks:
       namespace: {{ .Release.Namespace }}
       name: kargo-webhooks-server
       path: /mutate-kargo-akuity-io-v1alpha1-freight
+    {{- if not .Values.webhooksServer.tls.selfSignedCert }}
```
ricoapon: You could choose to change this to "if `caBundle` is not empty. I think this is cleaner, since running `helm template` will now still show the key.

krancour:
Thank you for this @ricoapon. This was a known issue (due to another discussion), but didn't have its _own_ issue yet.

You just saved me a lot of work.

@hiddeco do you mind reviewing this?

charts/kargo/templates/kubernetes-webhooks-server/webhooks.yaml
hiddeco: Might be an idea to make this a helper function?

ricoapon: What would be the added value? This file is the only place where this is used. I would expect a helper function to be used throughout the chart.

krancour: `locate` isn't particularly GitOps friendly. I'd like to avoid it.

ricoapon: @krancour Where do you see `locate`? Did you mean `lookup`?

Also, what would be the alternative? I cannot see any other solution.

krancour: Sorry... It's late. I did mean `lookup`.

This is quite tricky.

One option you have, which requires no modifications to the chart, would be to use a cert that's already trusted by the Kubernetes API server.

If you're intent on bringing your own cert that _isn't_ already trusted by the the Kubernetes API server, the chart modification that would reliably work would be adding a `webhooks.caBundle` field in the `values.yaml`. You would have to specify its value at install-time.

That's sort of the price of admission to use your own cert since for this particular use case. For reasons I don't have a handle on, and quite unfortunately, imho, Kubernetes doesn't let you specify a CA bundle by referencing a ConfigMap or Secret. _That would really take care of all our problems._

The alternative is simply to _not_ bring your own cert. There are tricks I've used in the past to make the chart generate a self-signed CA, use that to sign a cert, and insert all relevant details into both a `kubernetes.io/tls` Secret and the webhook registrations. But this is complex and inefficient in comparison to using cert-manager for this and either way what you end up with is a self-signed cert, so depending on cert-manager to do all of this is the better option.

Help me understand...

Are you resistant to the notion of self-signed certs? Or are you resistant to using cert-manager?

idk if it changes anything for you, but it's perfectly secure to use a self-signed cert for this use case. This webhook server isn't internet-facing and the fact of being self-signed is really no problem at all since the k8s API server, courtesy of the webhook registrations, will have a CA bundle it can use to validate the cert. If you're going to do this, cert-manager is the quick and easy way of accomplishing that.

So what is it that's actually standing in your way?

ricoapon: Those are some good questions. I am actually surprised though, since you seem to ask about my perspective on the certificates. I thought there was already an idea on how the Helm chart should be used without cert-manager. I just tried to follow the existing approach. I will share what I thought was the idea and what my opinion is. I hope you can fill in the details on how you think it should work.

> Are you resistant to the notion of self-signed certs? Or are you resistant to using cert-manager?

I believe that a dependency like `cert-manager` should not be mandatory. It is the default, which is totally fine. The tool `cert-manager` is mature and used by many companies. I mentioned this also in my team and we now also consider installing it (unrelated to Kargo). So I really do think that relying on this approach is a good one.

If a user does not want `cert-manager` for whatever reason (probably only a small group), Kargo should still function. This is the context to which my PR applies. Additionally, this means `webhooksServer.tls.selfSignedCert` is set to false.

> Kubernetes doesn't let you specify a CA bundle by referencing a ConfigMap or Secret. That would really take care of all our problems.

Yes, the only reason why we have this discussion at all is because this functionality is not present in Kubernetes. Maybe we should also look at other big Helm charts to see how they solve this problem before making a decision here. I have not done this analysis. Maybe you are familiar with other approaches?

### What I thought was the approach
The documentation of `webhooksServer.tls.selfSignedCert` states:
> If `false`, a cert secret named `kargo-webhooks-server-cert` **must** be provided in the same namespace as Kargo.

This leads me to believe that the user that manages the Helm chart should, in whatever way possible, create the certificate themselves. It doesn't matter if it is self-signed or not. We don't have this knowledge. Just that it needs to exist. It could be that the user creates their own Certificate object for `cert-manager` (strange to do this if the chart supports it, but it is a valid case right?). It could be that they request a certificate from a provider (stuff like CloudFlare, Xolphin, you name it) and upload it to the cluster. Or, most likely, they create their own self-signed certificate.

I consider all these scenarios to be "equal" from the perspective of the chart: the user must arrange it all themselves and we don't care how it is created. The chart will just make this one work. There is only one requirement that the chart documents: the name of the secret must be `kargo-webhooks-server-cert`. The rest is up to the user.

Following this line of thought, it would be logical to ensure that the certificate should be added to the caBundle. We cannot guarantee that the certificate that is created is already trusted by Kubernetes, and we have also no way to know this. Hence my change: lookup the certificate from the cluster and add it to the `caBundle`.

### Modifications to this approach
I choose to not add any additional configuration options in the `values.yaml`. We could of course do this. We could create a boolean `isTrustedByKubernetes` or something, where only if it is false that we should add the `caBundle`. The only added value for this would be to lock the `lookup` behind a specific boolean, which I think adds no value at all.

In line with your other suggestion: we could also add the certificate to the `values.yaml` in some option like `certBase64` or something. We could do the lookup if this value is empty. I am not a big fan of adding the certificate to the Helm chart. This would mean I have duplicate configuration. I would have to configure the Secret, which contains the certificate, and then copy-paste the content into the `values.yaml`. No matter how much I agree that `lookup` is not the nicest thing, this would also not be user-friendly. This change is an approach I can get behind though!

### My opinion on this approach
I think this is the most logical approach. The chart indicates: the user has to figure out the details of the certificate itself. But we also need to ensure that the certificates work for Kargo. So adding it to caBundle is required.

### Creating the certificate in the cart
Instead of having the user create the certificate, we could choose to have the chart do this for the user. That would be user-friendly and convenient. However, the entire chart has been built with `cert-manager` as a provider. This is clearly documented.  Creating a certificate in the Helm chart is a complex part that would cost a lot of effort and has almost no added value. `cert-manager` was created to exactly resolve this problem. If the user choose not to use this, it is their problem to figure out the certificate.

hiddeco: > There is only one requirement that the chart documents: the name of the secret must be `kargo-webhooks-server-cert`.

There is a secondary, implicit requirement, and that is that the Secret has to exist before the installation of Kargo. Otherwise, the data from it can not be read. This creates a chicken-egg problem that I do not necessarily like.

Additionally, the `caBundle` should contain the CA certificates from `ca.crt`, not the server certificate from `tls.crt`. The caBundle field expects the certificate authority that can validate the webhook server's certificate, which would mean that we should look at the `ca.crt`.

> In line with your other suggestion: we could also add the certificate to the values.yaml in some option like `certBase64` or something. We could do the lookup if this value is empty. I am not a big fan of adding the certificate to the Helm chart.

If we were to do this, it would be better to accept a `caBundle` and use the [`b64enc` function](https://helm.sh/docs/chart_template_guide/function_list/#encoding-functions) from Helm to transform that into base64.

krancour: > I thought there was already an idea on how the Helm chart should be used without cert-manager.

> I believe that a dependency like cert-manager should not be mandatory.

> If a user does not want cert-manager for whatever reason (probably only a small group), Kargo should still function.

> The documentation of webhooksServer.tls.selfSignedCert states:
>
>> If false, a cert secret named kargo-webhooks-server-cert must be provided in the same namespace as Kargo.
>
> This leads me to believe that the user that manages the Helm chart should, in whatever way possible, create the certificate themselves.

You are 100% correct about all of this and I share your opinions.

For the Kargo API server and for the recently added _external_ webhooks server, this approach of bringing your own Secret through whatever means you please works very well. For the registration of validating/mutating webhooks, which requires the Kubernetes API server to trust the (internal) webhooks server's certificate, the same approach doesn't hold up quite as well. It works, but _only_ if the Kubernetes API server's existing trust store can validate the cert. You could say this was either an accidental oversight or a limitation we failed to recognize. I take responsibility for this, as this is an issue I've had to resolve in other charts in the past and I should have recognized the oversight/limitation.

How I've dealt with this in the past:

> Instead of having the user create the certificate, we could choose to have the chart do this for the user.

I've done this and it's exactly the less-palatable approach I was previously describing. [Sprig](https://masterminds.github.io/sprig/) is the library of functions available within Helm chart templates. You can see that I added CA/cert generation/signing functions to that library [way back in 2017](https://github.com/Masterminds/sprig/pull/58). This was added for the benefit of the long-ago defunct [Kubernetes Service Catalog](https://github.com/kubernetes-retired/service-catalog) project, which faced the exact dilemma we're discussing. You can see where this functionality was used in that project's chart:

https://github.com/kubernetes-retired/service-catalog/blob/b37daf589e7797e209028c087d072817934d5d7c/charts/catalog/templates/webhook-register.yaml#L1-L5

https://github.com/kubernetes-retired/service-catalog/blob/b37daf589e7797e209028c087d072817934d5d7c/charts/catalog/templates/webhook-register.yaml#L13

What I don't care for about this is that:

* The certificate changes on every upgrade.
* Being non-deterministic, it's not GitOps-friendly. If this chart were installed via Argo CD, the live state of the cluster would _never_ completely match a fresh execution of `helm template` and your App would therefore exist in a state of perpetual drift -- which isn't the end of the world, but it's a thing you need to work around.

cert-manager, however, can essentially accomplish the above _without_ the drawbacks I cited.

I've also, in the past, used your `lookup`-based approach for avoiding re-generation of self-signed certs on every `helm upgrade`:

https://github.com/brigadecore/brigade/blob/21d51b12f495aef427a7f5fc43c3236bf6382e87/charts/brigade/templates/apiserver/ingress-cert-secret.yaml#L1-L32

(And you can see there that when the Secret doesn't already exist, it falls back on the previously discussed cert-generation functions.)

This `lookup`-based approach is, as I mentioned, not particularly GitOps friendly because it requires a connection to your cluster in order to achieve deterministic results. I've learned to avoid `lookup`.

If it matters, I worked for Deis, the company that created helm and was later acquired by MS. Although my contributions to Helm itself have been minimal, I've been writing charts since before Helm was a tool known to the public. It would be fair to say that I've put a lot of thought into this particular problem through the years. (Which makes it all the more embarrassing that I failed to immediately recognize that the "bring your own Secret" approach doesn't really work well when you need to register webhooks.)

Copping to the possibility that the state of the art in this area had evolved and I missed the memo, I posed this question to GitHub Copilot:

> is there a way to specify the ca bundle by referencing a secret? that would be much cleaner than having to actually include the encoded content of the bundle in this manifest

It's response was long, but the most salient bit of it was:

> **Common Practice**
> - Use automation (e.g., a controller, cert-manager’s `webhook` injector, or a CI/CD pipeline) to keep the `caBundle` up to date in the webhook configuration.
> - Tools like [cert-manager](https://cert-manager.io/docs/concepts/webhook/) can automate this injection for you.

And this is exactly what we're doing.

I think you start to see that opting out of cert-manager isn't quite as easy a proposition as we'd meant for it to be.

So where does it leave us?

1. We can add the caveat that opting out of cert-manager requires the (internal) webhook server's certificate to already be trusted by the Kubernetes API server. In truth, this is easier said than done because managed Kubernetes distros like EKS, GKE, AKS, etc. don't let you amend the Kubernetes API server's trust store.

1. We could allow the caBundle to be (optionally) specified via `webhooks.caBundle`.

1. We could do _both_ of the above.

I don't know what better options there are.

krancour: Aside: Ability to specify CA bundles via reference to a Secret or ConfigMap was first requested 6.5 years ago: https://github.com/kubernetes/kubernetes/issues/72944

ricoapon: I am honered to be working with such experts in the field! I am glad you have so much experience we can rely on. And everybody makes mistakes as well, it happens. Let's just move forward and see how we can fix this.

For me, the conclusion of this discussion is that we think very much alike. We agree with each others opinions and understand that there is actually no "good" solution because Kubernetes doesn't support it. Let's focus on getting the PR done.

> So where does it leave us?
> 
> 1. We can add the caveat that opting out of cert-manager requires the (internal) webhook server's certificate to already be trusted by the Kubernetes API server. In truth, this is easier said than done because managed Kubernetes distros like EKS, GKE, AKS, etc. don't let you amend the Kubernetes API server's trust store.
> 2. We could allow the caBundle to be (optionally) specified via `webhooks.caBundle`.
> 3. We could do _both_ of the above.
> 
> I don't know what better options there are.

I am a big fan option 3. This will make the chart function in scenario's where users don't have `cert-manager`, but we should also clearly communicate to others that this approach has its caveats.

I will make the Helm chart change. Thank you @hiddeco for the suggestion of using the `b64enc`. I will implement it like that.

I will document the scenario as best as I can in the property `.Values.webhooksServer.tls.cert` in the `values.yaml`. I think we should add some proper documentation on the website as well. Especially here: https://docs.kargo.io/operator-guide/advanced-installation/common-configurations/#ingress-tls

Do both of you agree that this is the right place to document (and that we do it in both `values.yaml` and website)? All three of us probably have our own preferences on documentation. We can go back and forth many times on small changes. Can you maybe help by giving a first suggestion or a direction? That would save us a lot of time reviewing in the PR. I am very willing to make all changes. A nudge into the right direction is good enough for me :)

charts/kargo/README.md
```diff
@@ -59,8 +59,8 @@ Optionally point to Kubernetes Secrets containing kubeconfig for:
 2. A remote cluster hosting Argo CD resources
 
 3. A remote cluster that is running Argo Rollouts and is a suitable location
-to execute user-defined verification processes in the form of Argo
-Rollouts AnalysisRuns
```
ricoapon: This change was included when generating docs. I did not do anything with it. Seems ok to me.

charts/kargo/values.yaml
```diff
@@ -866,3 +866,5 @@ webhooksServer:
   tls:
     ## @param webhooksServer.tls.selfSignedCert  Whether to generate a self-signed certificate for the controller's built-in webhook server. If `true`, `cert-manager` CRDs **must** be present in the cluster. Kargo will create and use its own namespaced issuer. If `false`, a cert secret named `kargo-webhooks-server-cert` **must** be provided in the same namespace as Kargo. There is no provision for webhooks without TLS.
     selfSignedCert: true
+    ## @param webhooksServer.tls.cert  PEM-encoded TLS certificate (which should also be stored in `kargo-webhooks-server-cert`) to trust when `webhooksServer.tls.selfSignedCert` is **false**. When that cert is *already* trusted by the cluster, ignore this parameter. When it is *not* trusted (e.g. self-signed), configure the cert here.
+    cert: ""
```
ricoapon: Small detail: the certificate should be added in the `values.yaml` like this:
```
webhooksServer:
  tls:
    selfSignedCert: false
    cert: |
      -----BEGIN CERTIFICATE-----
      ...
      -----END CERTIFICATE-----
```

If `>` is used instead of `|`, the new-lines are removed. This is incorrect PEM format and also very hard to detect as the error if one doesn't know about it. Not sure if it would be worth a comment in the documentation or as an example maybe? Just thought it would be good to mention. Maybe one of you have an idea how to integrate this knowledge into the documentation.

charts/kargo/templates/kubernetes-webhooks-server/webhooks.yaml
krancour: Glad we're converging on something that could work.

In terms of where to add the options and how to document them, the `values.yaml` has _two_ relevant sections here and I think they both need very small tweaks. There's `webhooks` and `webhooksServer`. (And `externalWebhooksServer` is something else entirely.) Why are `webhooks` and `webhooksServer` separate sections? At some point it seemed slightly more probably than it has turned out to be in practice that the webhooks might be registered in one cluster with the server running on another (or outside Kubernetes entirely?)

The `webhooks` section currently has only one option under it: `webhooks.register`. This is what will cause the resources (webhook registrations) in the `charts/kargo/kubernetes-webhooks-server/webhooks.yaml` template to be rendered or not. I think this section would be the most appropriate place to add a `caBundle` option that defaults to either undefined or an empty string. In its presence, its value can be encoded with `b64enc` and included in each registration. In its absence, there's nothing to do.

I think this new field would be the right place to document the conditions/caveats. It can be left alone if you're using cert-manager (recommended) OR if your Kubernetes API server is already going to trust whatever cert you're using (possible, but unlikely). Users could be directed to `webhooksServer.tls` for further details.

In `webhooksServer.tls.selfSignedCert.tls.selfSignedCert` I think you could revise the language there to more strongly recommend cert-manager and to mention that if opting to bring your own cert instead, unless it's already trusted by the Kubernetes API server (again, unlikely) that you _must_ then provide a CA bundle using the `webhooks.caBundle` setting. I _might_ go so far as to mention that webhook registrations cannot specify a CA bundle by referencing a Secret, so the _content_ of the bundle must be provided directly.

Aside: I've volunteered as tribute to sort this out the right way: https://github.com/kubernetes/kubernetes/issues/72944#issuecomment-3013462859

krancour: And thank you very, very much @ricoapon for this great discussion and your willingness to help improve our chart.

charts/kargo/README.md
hiddeco: I think this should go away if you rebase it, as I believe `main` already has this change.

charts/kargo/values.yaml
hiddeco: Wouldn't it be better to call this `caBundle`? As I still believe it should/could contain a bundle (including intermediate certificates) rather than only the certificate presented by the server.

krancour: > If `>` is used instead of `|`, the new-lines are removed. This is incorrect PEM format and also very hard to detect as the error if one doesn't know about it.

I'm not sure we need to go very deep into giving a YAML lesson -- but I won't say I've never been burned by that. I think we can nudge folks in the right direction by example. Something like this:

```yaml
## @section Webhooks
webhooks:
  ## @param webhooks.register Whether to create `ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration` resources.
  register: true
  ## @param webhooks.cabundle ... blah blah blah
  caBundle:
  # caBundle: |
  #   -----BEGIN CERTIFICATE-----
  #   ...
  #   -----END CERTIFICATE-----
```

charts/kargo/templates/kubernetes-webhooks-server/webhooks.yaml
ricoapon: Thank you as well @krancour for the discussion! I am very happy with the whole experience.

And cool that you volunteered to implemented the change. Hopefully something will come out of it!

hiddeco (approved):
This looks good to me! Thanks for following up with all our suggestions 🙇

krancour:
Nicely done @ricoapon. Thank you so much!

PR #4509, opened 2025-06-25
fix: external webhooks server fixes

krancour:
This PR addresses quite a number of bugs, all chart-related.

I found these while exercising different configuration permutations that aren't ordinarily exercised by our dev-time configuration when we `tilt up`.

I will add comments inline with the changes to better explain the necessity of each change.

### DIFF
--- a/hack/quickstart/kind.sh
+++ b/hack/quickstart/kind.sh
@@ -19,7 +19,7 @@ nodes:
   - containerPort: 31444 # Kargo dashboard
     hostPort: 31444
   - containerPort: 31445 # External webhooks server
-    hostPort: 31444
+    hostPort: 31445
   - containerPort: 30081 # test application instance
     hostPort: 30081
   - containerPort: 30082 # UAT application instance


charts/kargo/templates/NOTES.txt
```diff
@@ -107,6 +110,122 @@ Ready to get started?
...
+{{- $https = or .Values.api.ingress.tls.enabled .Values.api.tls.terminatedUpstream }}
+{{- $selfSignedCert = and .Values.api.ingress.tls.enabled .Values.api.ingress.tls.selfSignedCert (not .Values.api.tls.terminatedUpstream) }}
+{{- $url = printf "%s/webhook" (include "kargo.api.baseURL" .) }}
+
+⚙️  You've configured Kargo's external webhooks server to be accessible through
+   an Ingress controller.
+
+   If DNS and your Ingress controller are configured correctly, the external
+   webhooks server server is reachable at:
+
```
krancour: These lines carefully account for the scenario where one Ingress (that of the API server) is being used for both the API server _and_ the external webhooks server.

krancour: Eh. I don't like how I did this. Refactoring.

charts/kargo/templates/_helpers.tpl
```diff
@@ -55,9 +55,10 @@ Generate the base URL for the API service.
 Generate the base URL for the external webhook server.
 */}}
 {{- define "kargo.externalWebhooksServer.baseURL" -}}
+{{- $apiService := .Values.api -}}
 {{- $webhookService := .Values.externalWebhooksServer -}}
-{{- if and (not $webhookService.ingress.enabled) (not $webhookService.tls.enabled) (not $webhookService.tls.terminatedUpstream) -}}
```
krancour, line 59: These conditions seemed incorrect.

charts/kargo/templates/_helpers.tpl
```diff
@@ -55,9 +55,10 @@ Generate the base URL for the API service.
...
-{{- include "kargo.api.baseURL" . -}}
+{{- if and (not $webhookService.ingress.enabled) $apiService.enabled $apiService.ingress.enabled -}}
+{{- printf "%s/webhooks" (include "kargo.api.baseURL" .) -}}
```
krancour, line 61: If API Ingress is enabled for the API server and the external webhooks server does not have its own dedicated ingress, the API server ingress will route `/webhook` to the external webhooks server.

We were not accounting for thins in constructing the external webhook server's base URL.

charts/kargo/templates/api/ingress.yaml
```diff
@@ -34,12 +34,12 @@ spec:
               {{- end }}
       {{- if and .Values.externalWebhooksServer.enabled (not .Values.externalWebhooksServer.ingress.enabled) }}
       - pathType: {{ .Values.api.ingress.pathType | default "ImplementationSpecific" }}
-        path: /webhook
+        path: /webhooks
```
krancour, line 37: Not a hill I will die on, but I think plural feels more natural here.

charts/kargo/templates/api/ingress.yaml
```diff
@@ -34,12 +34,12 @@ spec:
...
         backend:
           service:
             name: kargo-external-webhooks-server
             port:
-              {{- if .Values.api.tls.enabled }}
+              {{- if .Values.externalWebhooksServer.tls.enabled }}
```
krancour, line 42: This was incorrectly determining whether to use TLS for the connection from the Ingress controller to the external webhook server endpoints based on whether the _API_ server endpoints used TLS.

charts/kargo/templates/external-webhooks-server/cert.yaml
```diff
@@ -13,5 +13,5 @@ spec:
   issuerRef:
     kind: Issuer
     name: kargo-selfsigned-cert-issuer
-  secretName: kargo-external-webhooks-server
+  secretName: kargo-external-webhooks-server-cert
```
krancour, line 16: This is more consistent with how we did things for the API server and the admission control webhooks. Naming this differently also preserves the possibility of using the name `kargo-external-webhooks-server` for a more general purpose repository of external webhook server secrets in the future.

charts/kargo/templates/external-webhooks-server/deployment.yaml
```diff
@@ -82,42 +82,43 @@ spec:
           {{- toYaml . | nindent 8 }}
         {{- end }}
         ports:
-        - containerPort: 9443
```
krancour, line 85: This was copy pasta from the admission control webhooks server, which unconditionally uses TLS. Switching to 8080 makes the external webhooks server more consistent with the API server. The external webhook server actually resembles the API server much more than it does the admission control webhooks server.

charts/kargo/templates/external-webhooks-server/deployment.yaml
```diff
@@ -82,42 +82,43 @@ spec:
...
-          name: webhooks-server
```
krancour, line 86: This port name wasn't very enlightening as to the protocol used by this port's listener. This is more informative.

charts/kargo/templates/external-webhooks-server/deployment.yaml
```diff
@@ -82,42 +82,43 @@ spec:
...
-        {{- end }}
         {{- with .Values.externalWebhooksServer.securityContext | default .Values.global.securityContext }}
         securityContext:
           {{- toYaml . | nindent 10 }}
         {{- end }}
         resources:
           {{- toYaml .Values.externalWebhooksServer.resources | nindent 10 }}
-      {{- if or .Values.externalWebhooksServer.tls.enabled .Values.kubeconfigSecrets.kargo  }}
+      {{- if or .Values.kubeconfigSecrets.kargo .Values.externalWebhooksServer.tls.enabled }}
       volumes:
```
krancour, line 101: All volumes/volumeMounts have been updates to more closely resemble the API server instead of the admission control webhooks server.

krancour: Without these changes, the external webhooks server was failing to find its cert when TLS is enabled.

hack/quickstart/install.sh
```diff
@@ -45,4 +45,6 @@ helm install kargo \
   --set api.service.nodePort=31444 \
   --set api.adminAccount.passwordHash='$2a$10$Zrhhie4vLz5ygtVSaif6o.qN36jgs6vjtMBdM6yrU1FOeiAAMMxOm' \
   --set api.adminAccount.tokenSigningKey=iwishtowashmyirishwristwatch \
+  --set externalWebhooksServer.service.type=NodePort \
+  --set externalWebhooksServer.service.nodePort=31445 \
```
krancour, line 49: The external webhooks server does not really play a role in the quickstart, but as long as we're not going out of our way to disable it, we may as well make it accessible.

(Personally, I find this quite convenient as I typically run through the quickstart prior to a release and then roll right into testing other features, which now includes webhooks.)

internal/webhook/external/receiver.go
```diff
@@ -155,33 +155,46 @@ func getDetails(
 	receiverName string,
 	secretValues ...string,
 ) (kargoapi.WebhookReceiverDetails, error) {
```
krancour, line 157: This function is updated to find any path included in the baseURL and use it as part of the path we are building. We need to do this or we'll else we have problems when trying to match the path of an inbound webhook request to a receiver.

internal/webhook/external/receiver.go
```diff
@@ -155,33 +155,46 @@ func getDetails(
...
-	// Warning: Changes to this line could alter URLs that users may already be
-	// using!!!
+) (string, error) {
+	if basePath == "" {
+		basePath = "/"
+	}
 	input := []byte(project + receiverName + strings.Join(secretValues, ""))
 	h := sha256.New()
 	_, _ = h.Write(input)
-	return fmt.Sprintf("/webhook/%s/%x", receiverType, h.Sum(nil))
```
krancour, line 186: I removed the static `webhook/` from the path we build.

There are two scenarios...

1. The external webhooks server lives on its own subdomain -- something like webhooks.kargo.example.com. In this case, `https://webhooks.kargo.example.com/webhooks/...` looks redundant.

2. The API server's Ingress routes `webhooks/` to the external webhooks server. In this case, `https://kargo.example.com/webhooks/webhooks/...` also looks pretty redundant.

hack/quickstart/kind.sh
```diff
@@ -18,6 +18,8 @@ nodes:
     hostPort: 31443
   - containerPort: 31444 # Kargo dashboard
     hostPort: 31444
+  - containerPort: 31445 # External webhooks server
+    hostPort: 31444
```
hiddeco:
```suggestion
    hostPort: 31445
```

krancour: 🤦‍♂️ Of all the things I tested, running through this with kind is one of the few I didn't. Great catch.

krancour:
> Based on how far I was able to render this as a human, combined with some manual testing, this looks good to me.

I appreciate the effort.

fwiw, I meticulously tested every combination of Ingresses, Service types, and TLS on/off to ensure webhooks remained deliverable in all cases.

PR #4520, opened 2025-06-27
fix: application recorded in promotion event's annotation may be incorrect

hanxiaop:
Before the fix, the annotation may look something like this:
```
    annotations:
      event.kargo.akuity.io/actor: admin
      event.kargo.akuity.io/applications: '[{"Namespace":"argocd","Name":"kargo-demo-${{
        ctx.stage }}"}]'
```
which was directly recorded before the evaluation with defined variables.

### DIFF
--- a/internal/event/promotion.go
+++ b/internal/event/promotion.go
@@ -126,7 +126,9 @@ func NewPromotionAnnotations(
 						"name": f.Origin.Name,
 					}
 				}
-				env["ctx"].(map[string]any)["targetFreight"] = targetFreight
+				if ctx, ok := env["ctx"].(map[string]any); ok {
+					ctx["targetFreight"] = targetFreight
+				}
 			}
 
 			if evaled, err := expressions.EvaluateTemplate(string(data), env); err == nil {
--- a/internal/event/promotion.go
+++ b/internal/event/promotion.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"maps"
+	"sort"
 	"time"
 
 	"k8s.io/apimachinery/pkg/types"
@@ -71,39 +72,7 @@ func NewPromotionAnnotations(
 		}
 	}
 
-	var apps []types.NamespacedName
-	for _, step := range p.Spec.Steps {
-		if step.Uses != "argocd-update" || step.Config == nil {
-			continue
-		}
-		var cfg builtin.ArgoCDUpdateConfig
-		if err := json.Unmarshal(step.Config.Raw, &cfg); err != nil {
-			logger.Error(err, "unmarshal ArgoCD update config")
-			continue
-		}
-		for _, app := range cfg.Apps {
-			namespacedName := types.NamespacedName{
-				Namespace: app.Namespace,
-				Name:      app.Name,
-			}
-			if namespacedName.Namespace == "" {
-				namespacedName.Namespace = libargocd.Namespace()
-			}
-			apps = append(apps, namespacedName)
-		}
-	}
-
-	if len(apps) == 0 {
-		return annotations
-	}
-
-	data, err := json.Marshal(apps)
-	if err != nil {
-		return annotations
-	}
-
-	var result string
-	env := map[string]any{
+	baseEnv := map[string]any{
 		"ctx": map[string]any{
 			"project":   p.GetNamespace(),
 			"promotion": p.GetName(),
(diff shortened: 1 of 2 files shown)


internal/event/promotion.go
```diff
@@ -95,7 +96,30 @@ func NewPromotionAnnotations(
...
+					"stage":     p.Spec.Stage,
+				},
+				"vars": func() map[string]any {
+					vars := map[string]any{}
+					for _, v := range p.Spec.Vars {
+						vars[v.Name] = v.Value
+					}
+					return vars
+				}(),
+			}
```
hiddeco: I wonder if this is sufficient to solve _all_ issues, as I believe that technically anything available in the env (including e.g. outputs from steps) can be used in the config.

hanxiaop: @hiddeco agree, this could be flaky if the vars don't align with the defined envs. I can see another option: populating this app info in the promotion status's state field. However, the problem is that it would only work for update events, as create events wouldn't contain the status I think. WDYT about adding some kind of annotations to the promotion so that this info can be used for every event? Or do you have any other ideas? Thank you!

hiddeco: Can you illustrate the type of annotations that you would add? Trying to understand the full picture.

hanxiaop: After looking at the code again, I found the annotation still doesn't work. I only want to get the exact parsed app name from the promotion, but I've found that the real app names are only available when the step is executing(after the context is prepared). This means if I want to handle the app name for all cases, only events recorded after `prepareStepContext` would have the info, and the creation event will miss the app info, which is also not perfect.

hiddeco: Yes, that's what I believe to be true as well.

I think our best option is to provide as much as we can provide to render the config,  while failing gracefully (by publishing the event without ArgoCD application metadata) in cases the information is not available.

hanxiaop: I've made a fix based on https://docs.kargo.io/user-guide/reference-docs/expressions/#pre-defined-variables, please take another look. Thanks!

internal/event/promotion.go
```diff
@@ -95,7 +96,54 @@ func NewPromotionAnnotations(
...
+						},
+					},
+				},
+				"vars": func() map[string]any {
+					vars := map[string]any{}
+					for _, v := range p.Spec.Vars {
+						vars[v.Name] = v.Value
+					}
+					return vars
+				}(),
```
krancour: I think there are two pretty important things that don't seem to be accounted for here.

1. Step-level vars are also a thing. They can override Promotion-level vars. This means you will actually have to build the list of updated Apps step-by-step, evaluating expressions step-by-step as you go.

2. Vars at both the Promotion-level and step-level can be defined using expressions referencing previously defined vars.

    You can see there is some relatively complex logic for calculating the full set of vars for a given step:

    https://github.com/akuity/kargo/blob/22bf5d0e9c19d3ec91e875cd163e96acf08e6f59/internal/promotion/promotion.go#L362-L411

    I believe that may need to be refactored a tiny bit to be more reusable in this context as well.

There are obviously other things (e.g. various functions) that rely on promotion context that cannot be added to the env quite as easily here, but I know @hanxiaop and @hiddeco seem to have agreed on evaluating App names that are defined using an expressions on a "best effort" basis, so I think we can ignore the unavailability of those functions for now, but getting the vars correct is definitely essential here.

cc @hiddeco for any additional input.

krancour: Or...

There might be an altogether different approach that works. If an argocd-update step completes successfully, it registers a healthcheck to be performed periodically as part of Stage reconciliation.

You could look at the health checks instead of the steps to find all Apps that were updated. The healthcheck config is always static. There's no need to evaluate any expressions at all.

I would actually suggest this approach over the other, but the caveat is those health checks are only registered when an App is updated/synced _successfully_. I think that may be ok for your purposes here, but I'm not positive.

cc @hiddeco again.

hiddeco: I had not thought about the health checks getting added, and I love this idea for cases where the Application is updated/synced successfully.

I do worry, though, that we soon will need this information when, e.g., a Promotion itself is created — and I think having a "best-effort" fallback in this case would make it a perfect pair.

krancour: I agree... it's why I mentioned the caveat. If the current need is just have an event when an App is successfully updated, the healthcheck trick would get us quite far with almost no effort. I think we could hold off on doing anything more sophisticated until we know there's some other event we're actually interested in publishing, with Promotion creation you cited being a prime example. Another example would be if we wanted to create an event for an App that _failed_ to update. I guess the question really is exactly _what_ we care about capturing right now.

hanxiaop: @krancour I've added a commit to evaluate with step-by-step logic in https://github.com/akuity/kargo/pull/4520/commits/ecc56e23c437e6c793002c6c1b493242fbe3b5ef, please take another look when you have time. Thank you!

krancour: @hanxiaop sorry for the delay. The changes in ecc56e23c437e6c793002c6c1b493242fbe3b5ef do not appear to be correct.

Currently, you are still building out _just one_ environment for expression evaluation and then evaluating your serialized list of Apps all at once. You cannot do that.

I will give an example of why this does not work:

Suppose you have two steps where _both_ update App "${{ someApp }}". In the first step, `someApp` is defined as `"my-app"`, while in the second step, `someApp` is defined as `"your-app"`.

By your current logic, you will start by building a list of Apps that ends up serializing something like this:

```
[
  {
    name: "${{ someApp }}",
    namespace: "argocd",
  },
  {
    name: "${{ someApp }}",
    namespace: "argocd",
  }
]
```

Then you are building out _one_ environment that expressions are evaluated against and you're evaluating the entire templated JSON above all at once. The logic in `calculateVars` is incorrect, but we can set that aside for a moment. Whether it were correct or incorrect, `someApp` can only have _one_ value when you are building only one env and evaluating only once. Your result is going to end up being something like:

```
[
  {
    name: "your-app",
    namespace: "argocd",
  },
  {
    name: "your-app",
    namespace: "argocd",
  }
]
```

The above is clearly wrong.

What you need to do is _do not_ build a list of Apps with unevaluated expressions in their names up front. Start with an empty list of Apps, then go over the steps one-by-one, building out the env _for each step_, evaluating that step's config, and _only then_ extracting the names of any Apps updated by that step and adding them to the list of Apps. Repeat for the next step.

You probably need to sort and de-dupe the resulting list of Apps when you are done.

hanxiaop: @krancour thank you for the explanation. I misunderstood the steps and thought the `argocd-update` step only existed once. I've pushed a new commit and added a [test case with multiple steps](https://github.com/akuity/kargo/pull/4520/commits/f6dc139f56fe147bbccd65c622f015023e4bcb9f#diff-d340be45d9cdb5df6fe5df5f3fce267bc995b5e5bbf3adbad2ff63ffd35a8842R155-R218). Let me know if the output makes sense.

krancour: It looks much improved. Thank you. 🙏 

I found an opportunity to optimize a little bit using more specialized support for JSON templates that already exists in our `expressions` package. I pushed it in 9d22875694047c144faade00ce1f2f6686acc930.

2832982c373e310b9ea93d71d5c8cbba1ac67590 contains some small tweaks to make the tests more readable. (Whitespace is irrelevant in JSON, so JSON config was more readable if we allow a whole bunch of extra indentation in the strings. It results in config that is indented properly WRT the test cases.)

There is another slightly more involved optimization that I think is possible, but I'd be interested in whether we think it's worth it...

Since we're building as much of the expression evaluation environment as we can and it's not _exactly_ what is used during the actual promotion, there is a possibility we get better, more resilient results by unmarshaling the _unevaluated_ config for each step and then for each `argocd-update` step, evaluating all the apps' `name` and `namepace` fields _individually_. The advantage this has over evaluating the entire config at once is that if some other, _unrelated_ part of a step's configuration cannot be evaluated correctly due to the partial env we're using, it doesn't prevent us from successfully evaluating the specific parts the config we actually care about.

It would look something like this:

```golang
for _, step := range p.Spec.Steps {
		if step.Uses != "argocd-update" || step.Config == nil {
			continue
		}
		var cfg builtin.ArgoCDUpdateConfig
		if err := json.Unmarshal(step.Config.Raw, &cfg); err != nil {
			logger.Error(err, "unmarshal ArgoCD update config")
			continue
		}
		for _, app := range cfg.Apps {
			namespacedName := types.NamespacedName{
				Namespace: app.Namespace,
				Name:      app.Name,
			}


			// These more targeted field evaluations might give us more resiliency...
			if strings.Contains(namespacedName.Namespace, "${{") ||
				strings.Contains(namespacedName.Name, "${{") {
				stepEnv := make(map[string]any)
				maps.Copy(stepEnv, baseEnv)
				stepVars := calculateStepVars(step, stepEnv)
				setVar(stepEnv, stepVars)
				var ok bool
				namespaceAny, err :=
					expressions.EvaluateTemplate(namespacedName.Namespace, stepEnv)
				if err != nil {
					logger.Error(err, "error evaluating expression")
					continue
				}
				if namespacedName.Namespace, ok = namespaceAny.(string); !ok {
					logger.Error(err, "expression did not evaluate to a string")
					continue
				}
				appNameAny, err :=
					expressions.EvaluateTemplate(namespacedName.Name, stepEnv)
				if err != nil {
					logger.Error(err, "error evaluating expression")
					continue
				}
				if namespacedName.Name, ok = appNameAny.(string); !ok {
					logger.Error(err, "expression did not evaluate to a string")
					continue
				}
			}


			if namespacedName.Namespace == "" {
				namespacedName.Namespace = libargocd.Namespace()
			}
			if _, exists := appSet[namespacedName]; !exists {
				appSet[namespacedName] = struct{}{}
				allApps = append(allApps, namespacedName)
			}
		}
	}
```

Last, @hiddeco has worked on optimizing the other spots where we uses this `expression` package more recently than I have, so I'm curious if he has any thoughts on whether there is more DRY way to build as much of the context/env as possible.

hanxiaop: Thank you for the improvements! The optimization makes sense to me. I'll see if @hiddeco has any thoughts and revise the code.

krancour: > The optimization makes sense to me.

I pushed that last optimization. Tests still pass.

This LGTM.

krancour (approved):

internal/event/promotion.go
hanxiaop: Thank you!

Issue #4487, opened 2025-06-24
add webhook receiver for azure


krancour:
Broken out from #4150

Support for "Azure" means notification of push events from Azure DevOps (Git) repositories _and potentially_ notification of push events from ACR (OCI) repositories.

It's not yet known whether ACR directly supports delivering notifications to external systems via webhooks. Some research will be required. If it does not, chances are that anything that can be jury rigged to enable that will be best served by the generic webhook receiver (#4486).

krancour:
Closed by #4536

PR #4522, opened 2025-06-27
feat: azure webhook receiver

fuskovic:
Related to: https://github.com/akuity/kargo/issues/4487

Includes support for both:
- Azure Container Registry
- Azure DevOps


Marvin9 (commented):
Reminder for me to add [UI support](https://github.com/akuity/kargo/blob/main/ui/src/features/project/settings/views/project-config/webhook/webhook-receivers.tsx)

api/v1alpha1/project_config_types.go
```diff
@@ -120,6 +120,9 @@ type WebhookReceiverConfig struct {
 	// Quay contains the configuration for a webhook receiver that is compatible
 	// with Quay payloads.
 	Quay *QuayWebhookReceiverConfig `json:"quay,omitempty" protobuf:"bytes,4,opt,name=quay"`
+	// Azure contains the configuration for a webhook receiver that is compatible
+	// with Azure Container Registry (ACR) and Azure DevOps payloads.
+	Azure *AzureWebhookReceiverConfig `json:"azure,omitempty" protobuf:"bytes,7,opt,name=azure"`
```
krancour: Note: There are PRs open currently for three new receivers and all three are competing for field index `7`. So this is just a reminder to merge the latest from main and update this index as needed.

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,181 @@
...
+			predicate: func(cfg kargoapi.WebhookReceiverConfig) bool {
+				return cfg.Azure != nil
+			},
+			factory: newAzureWebhookReceiver,
+		},
+	)
+}
+
+// azureWebhookReceiver is an implementation of WebhookReceiver that handles
+// inbound webhooks from azure.
```
krancour: This cuts through to this crucial bit of info without requiring a detailed reading of the code:

```suggestion
// azureWebhookReceiver is an implementation of WebhookReceiver that handles
// inbound webhooks from both Azure Container Registry (ACR) and Azure DevOps.
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,181 @@
...
+type azureWebhookReceiver struct {
+	*baseWebhookReceiver
+}
+
+// newazureWebhookReceiver returns a new instance of azureWebhookReceiver.
```
krancour:
```suggestion
// newAzureWebhookReceiver returns a new instance of azureWebhookReceiver.
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,181 @@
...
+	return &azureWebhookReceiver{
+		baseWebhookReceiver: &baseWebhookReceiver{
+			client:     c,
+			project:    project,
+			secretName: cfg.Azure.SecretRef.Name,
+		},
+	}
+}
+
+// GetDetails implements WebhookReceiver.
```
krancour:
```suggestion
// getReceiverType implements WebhookReceiver.
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,181 @@
...
+// getHandler implements WebhookReceiver.
+func (q *azureWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := logging.LoggerFromContext(ctx)
+		logger.Debug("identifying source repository")
+
+		logger.Info("received Azure webhook event",
+			"requestBody", string(requestBody),
+		)
```
krancour: Logged at too high a level. This should be debug-level if we keep it at all.

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,181 @@
...
+				),
+			)
+			return
+		}
+	})
+}
+
+// events can come from Azure Container Registry or Azure DevOps.
+// azureEvent is a struct that represents the payload of an event from either
+// of these services as there are no overlapping fields between the two
```
krancour, line 151: By widely observed Go convention, comments on functions should start with the function name:

```suggestion
// azureEvent represents the payload of an event from either Azure Container
// Registry or Azure DevOps.
// of these services as there are no overlapping fields between the two
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,181 @@
...
+	} `json:"target"` // For Azure Container Registry
+	Request struct {
+		Host string `json:"host"`
+	} `json:"request"` // For Azure Container Registry
+	Resource struct {
+		Repository struct {
+			RemoteURL string `json:"remoteUrl,omitempty"`
+		} `json:"repository"`
+	} `json:"resource"` // For Azure DevOps
+}
```
krancour: I know we previously spoke about this offline and I'd said I wouldn't have approached things this way, but I understood your reasoning. Looking at the code now, I think I understand your reasoning even more. At the switch up on line 96, you're trying to figure out what kind of event you're dealing with (and from which service). Since that information is included in the request body, you can't make that decision before parsing the body. By using the approach you currently are, you're avoiding having to parse twice -- once into a `map[string]any` just to find out what kind of event you're dealing with and then _again_ to unpack the body into something well-typed. Is this it, more or less?

I empathize with the dilemma, but having said that, I had a visceral negative reaction when I looked at the type def. Having to say field by field which service the field is associated with is screaming that there's got to be a better way.

I got hands on and sent myself some webhooks from both ACR and Azure DevOps to see what they look like in both cases -- and lucky us. There's a header we can look at. Requests from ACR will contain the header `User-Agent: AzureContainerRegistry/1.0` and requests from ADO will contain the header `VSServices/20.258.36217.10 (w3wp.exe)`. imho, if you look at the `User-Agent` header and unmarshal into an ACR-specific event type when the value starts with `AzureContainerRegistry` and unmarshal into an ADO-specific event type when the value starts with `VSServices` (it's deliberate that I'm removing version information from the equation), then you'll be able to proceed having only parsed the body once and having something much more ergonomic now in-hand.

Sound ok? I think this change is fairly small and will end up feeling cleaner.

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,188 @@
...
+		baseWebhookReceiver: &baseWebhookReceiver{
+			client:     c,
+			project:    project,
+			secretName: cfg.Azure.SecretRef.Name,
+		},
+	}
+}
+
+// getReceiverType implements WebhookReceiver.
+func (q *azureWebhookReceiver) getReceiverType() string {
```
krancour: Guessing the Quay receiver was used as a sort of template. This will have to change in a few spots.

```suggestion
func (a *azureWebhookReceiver) getReceiverType() string {
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,188 @@
...
+		default:
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(
+					fmt.Errorf("user agent %s is not supported", r.UserAgent()),
+					http.StatusNotImplemented,
+				),
+			)
+			return
+		}
```
krancour: The nested switch/cases are particularly hard to follow. Can I suggest a single switch in this method that just handles ACR/ADO/unsupported and for the ACR or ADO case, just invoke another more-specific handler method? The icing on that cake is that the smaller ACR and ADO handlers will, individually, be easier to test.

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,188 @@
...
+	})
+}
+
+// For more information on the payload schema for Azure Container Registry, see:
+//
+//	Azure Container Registry
+//		https://learn.microsoft.com/en-us/azure/container-registry/container-registry-webhook-reference#payload-example-image-push-event
+//
+// nolint:lll
+type acrEvent struct {
```
krancour, line 197: By widely observed convention, this comment should start with the name of the type. Same for `azureDevOpsEvent`.

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,188 @@
...
+			logger = logger.WithValues("repoURL", repoURL)
+			ctx = logging.ContextWithLogger(ctx, logger)
+			refreshWarehouses(ctx, w, q.client, q.project, repoURL)
+		default:
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(
+					fmt.Errorf("user agent %s is not supported", r.UserAgent()),
+					http.StatusNotImplemented,
+				),
```
krancour: This isn't really a matter of us not having implemented something -- or more specifically, its not a matter of us not having implemented something we might reasonably have been expected to. So I would say the real problem is we were sent something we shouldn't have been. i.e. It's a bad request.

(Aside, we're already using `http.StatusNotImplemented`) in three other places where I think I'm starting to regret it for the same reason and I will take it upon myself to fix those.)

I might also avoid mention of the user agent header because it will be in no way obvious to anyone looking at this message what role that header plays in us deciding if we can handle the request or not. (In fact we're only looking at it for lack of any better option.) I would maybe try something along these lines:

```suggestion
				xhttp.Error(
					errors.New(
						"request does not appear to have originated from a supported service",
					),
					http.StatusBadRequest,
				),
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// For information on payload schemas for Azure DevOps, see:
+//
+//	Azure DevOps
+//		https://learn.microsoft.com/en-us/azure/devops/service-hooks/services/webhooks?view=azure-devops#resource-details-to-send
+//
+// nolint:lll
+type azureDevOpsEvent struct {
+	EventType string `json:"eventType,omitempty"` // For Azure DevOps
```
krancour: I'd drop this comment because it's now contextually obvious that this is related to Azure DevOps. I guess this is just an accidental remnant from a previous iteration.

```suggestion
	EventType string `json:"eventType,omitempty"`
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+			map[string]string{
+				"msg": "ping event received, webhook is configured correctly",
+			},
+		)
+	default:
+		xhttp.WriteErrorJSON(
+			w,
+			xhttp.Error(
+				fmt.Errorf("event type %s is not supported", event.Action),
+				http.StatusNotImplemented,
```
krancour:
```suggestion
				http.StatusBadRequest,
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// getSecretValues implements WebhookReceiver.
+func (a *azureWebhookReceiver) getSecretValues(
+	secretData map[string][]byte,
+) ([]string, error) {
+	secretValue, ok := secretData[azureSecretDataKey]
+	if !ok {
+		return nil,
+			errors.New("secret data is not valid for an Azure WebhookReceiver")
```
krancour: Looks like @hiddeco improved these error messages recently across all the existing receivers. This would make this one match:

```suggestion
			fmt.Errorf("missing data key %q for Azure WebhookReceiver", azureSecretDataKey)
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+
+// getHandler implements WebhookReceiver.
+func (a *azureWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := logging.LoggerFromContext(ctx)
+		logger.Debug("received azure event", "userAgent", r.UserAgent())
+
+		switch {
+		// format is AzureContainerRegistry/<version>
```
krancour:
```suggestion
		// Format is AzureContainerRegistry/<version>
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+					),
+					http.StatusBadRequest,
+				),
+			)
+			return
+		}
+	})
+}
+
+func (a *azureWebhookReceiver) handleAcrEvent(
```
krancour: The widely observed convention in Go is to not mix case on an acronyms. If it starts small, keep going with small. If it starts, big, keep going with big.


```suggestion
func (a *azureWebhookReceiver) handleACREvent(
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+	logger = logger.WithValues("repoURL", repoURL)
+	ctx = logging.ContextWithLogger(ctx, logger)
+	refreshWarehouses(ctx, w, a.client, a.project, repoURL)
+}
+
+func resolveAcrRepoURLs(e acrEvent) []string {
+	repoURL := fmt.Sprintf("%s/%s", e.Request.Host, e.Target.Repository)
+	var repoURLs []string
+	switch e.Target.MediaType {
+	case imageMediaType:
```
krancour, line 174: We should probably also include `application/vnd.oci.image.manifest.v1+json` in this case.

There are existing receivers where we should fix that and I volunteer to do so.

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+		repoURLs = append(repoURLs, image.NormalizeURL(repoURL))
+	case helmChartMediaType:
+		repoURLs = append(repoURLs, helm.NormalizeChartRepositoryURL(repoURL))
+	default:
+		repoURLs = append(repoURLs,
+			image.NormalizeURL(repoURL),
+			helm.NormalizeChartRepositoryURL(repoURL),
+		)
+	}
```
krancour, line 183: If we don't recognize the media type, I think we actually have _nothing_ to do.

I'd suggest changing the return type of this function from `[]string` to `string`, since after this change, we'd return one URL _at most_.

krancour: Eh... when the return type changes, don't forget to de-pluralize the function name.

internal/webhook/external/azure_test.go
```diff
@@ -0,0 +1,257 @@
...
+	testScheme := runtime.NewScheme()
+	require.NoError(t, kargoapi.AddToScheme(testScheme))
+
+	testCases := []struct {
+		name       string
+		client     client.Client
+		req        func() *http.Request
+		assertions func(*testing.T, *httptest.ResponseRecorder)
+	}{
+		{
```
krancour, line 35: Can we add one more test case above here for when the User-Agent header doesn't reflect a service whose events we can handle?

krancour:
Only a few very small things remaining. This is quite excellent overall!

internal/webhook/external/azure.go
krancour: Ah... there are even more. Just ignore this bit of feedback and I will write a function that checks for all valid image media types.

internal/webhook/external/azure.go
krancour: Ok... if we can get #4621 merged first, you can rebase this PR on main or merge the latest from main into it and you'll have access to an improved way of normalizing OCI repo URLs that takes all relevant media types into account.

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/akuity/kargo/internal/helm"
+	xhttp "github.com/akuity/kargo/internal/http"
+	"github.com/akuity/kargo/internal/image"
+	"github.com/akuity/kargo/internal/logging"
+)
+
+const (
+	azureSecretDataKey = "secret"
+	azure              = "azure"
+)
```
krancour: Tests are going to need to use these values. May as well make them consts.

```suggestion
const (
	azureSecretDataKey = "secret"
	azure              = "azure"

	acrPingEvent = "ping"
	acrPushEvent = "push"

	acrUserAgentPrefix         = "AzureContainerRegistry"
	azureDevOpsUserAgentPrefix = "VSServices"
)
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+		ctx := r.Context()
+		logger := logging.LoggerFromContext(ctx)
+		logger.Debug("received azure event", "userAgent", r.UserAgent())
+
+		switch {
+		// format is AzureContainerRegistry/<version>
+		case strings.Contains(r.UserAgent(), "AzureContainerRegistry"):
+			a.handleAcrEvent(ctx, w, requestBody)
+		// Format is VSServices/<version>
+		case strings.Contains(r.UserAgent(), "VSServices"):
```
krancour:
```suggestion
		case strings.HasPrefix(r.UserAgent(), azureDevOpsUserAgentPrefix):
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+// getHandler implements WebhookReceiver.
+func (a *azureWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := logging.LoggerFromContext(ctx)
+		logger.Debug("received azure event", "userAgent", r.UserAgent())
+
+		switch {
+		// format is AzureContainerRegistry/<version>
+		case strings.Contains(r.UserAgent(), "AzureContainerRegistry"):
```
krancour:
```suggestion
		case strings.HasPrefix(r.UserAgent(), acrUserAgentPrefix):
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+		return
+	}
+
+	switch event.Action {
+	case "push":
+		repoURLs := resolveAcrRepoURLs(event)
+		logger := logging.LoggerFromContext(ctx)
+		logger = logger.WithValues("repoURLs", repoURLs)
+		ctx = logging.ContextWithLogger(ctx, logger)
+		refreshWarehouses(ctx, w, a.client, a.project, repoURLs...)
```
krancour: #4621 adds a method for more foolproof media type detection and OCI repo URL normalization. Need to merge `main` into here after it's #4621 is merged and then this can work nicely:

```suggestion
		repoURL := normalizeOCIRepoURL(
			fmt.Sprintf("%s/%s", event.Request.Host, eevent.Target.Repository),
			event.Target.MediaType,
		)
		logger := logging.LoggerFromContext(ctx)
		logger = logger.WithValues("repoURL", repoURL)
		ctx = logging.ContextWithLogger(ctx, logger)
		refreshWarehouses(ctx, w, a.client, a.project, repoURL)
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+	case helmChartMediaType:
+		repoURLs = append(repoURLs, helm.NormalizeChartRepositoryURL(repoURL))
+	default:
+		repoURLs = append(repoURLs,
+			image.NormalizeURL(repoURL),
+			helm.NormalizeChartRepositoryURL(repoURL),
+		)
+	}
+	return repoURLs
+}
```
krancour: See https://github.com/akuity/kargo/pull/4522/files#r2206104235

```suggestion
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+			),
+		)
+		return
+	}
+	repoURL := git.NormalizeURL(event.Resource.Repository.RemoteURL)
+	logger := logging.LoggerFromContext(ctx)
+	logger = logger.WithValues("repoURL", repoURL)
+	ctx = logging.ContextWithLogger(ctx, logger)
+	refreshWarehouses(ctx, w, a.client, a.project, repoURL)
+}
```
krancour: Are there only certain events we should be handling here?

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	switch event.Action {
+	case "push":
+		repoURLs := resolveAcrRepoURLs(event)
+		logger := logging.LoggerFromContext(ctx)
+		logger = logger.WithValues("repoURLs", repoURLs)
+		ctx = logging.ContextWithLogger(ctx, logger)
+		refreshWarehouses(ctx, w, a.client, a.project, repoURLs...)
+	case "ping":
```
krancour:
```suggestion
	case acrPingEvent:
```

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,219 @@
...
+			w,
+			xhttp.Error(errors.New("invalid request body"),
+				http.StatusBadRequest,
+			),
+		)
+		return
+	}
+
+	switch event.Action {
+	case "push":
```
krancour:
```suggestion
	case acrPushEvent:
```

internal/webhook/external/azure_test.go
```diff
@@ -0,0 +1,257 @@
...
+			name: "malformed request body",
+			req: func() *http.Request {
+				req := httptest.NewRequest(
+					http.MethodPost,
+					testURL,
+					bytes.NewBuffer([]byte("invalid json")),
+				)
+				req.Header.Set("User-Agent", "AzureContainerRegistry/1.0.0")
```
krancour: If we have a constant for this, we can set the header's value to just the prefix and that will be enough to satisfy the handler.

```suggestion
				req.Header.Set("User-Agent", acrUserAgentPrefix)
```

internal/webhook/external/azure_test.go
```diff
@@ -0,0 +1,257 @@
...
+				indexer.WarehousesBySubscribedURLsField,
+				indexer.WarehousesBySubscribedURLs,
+			).Build(),
+			req: func() *http.Request {
+				req := httptest.NewRequest(
+					http.MethodPost,
+					testURL,
+					newAzurePayload("ping", ""),
+				)
+				req.Header.Set("User-Agent", "AzureContainerRegistry/1.0.0")
```
krancour:
```suggestion
				req.Header.Set("User-Agent", acrUserAgentPrefix)
```

internal/webhook/external/azure_test.go
```diff
@@ -0,0 +1,257 @@
...
+				indexer.WarehousesBySubscribedURLsField,
+				indexer.WarehousesBySubscribedURLs,
+			).Build(),
+			req: func() *http.Request {
+				req := httptest.NewRequest(
+					http.MethodPost,
+					testURL,
+					newAzurePayload("push", imageMediaType),
+				)
+				req.Header.Set("User-Agent", "AzureContainerRegistry/1.0.0")
```
krancour:
```suggestion
				req.Header.Set("User-Agent", acrUserAgentPrefix)
```

internal/webhook/external/azure_test.go
```diff
@@ -0,0 +1,257 @@
...
+				indexer.WarehousesBySubscribedURLsField,
+				indexer.WarehousesBySubscribedURLs,
+			).Build(),
+			req: func() *http.Request {
+				req := httptest.NewRequest(
+					http.MethodPost,
+					testURL,
+					newAzurePayload("push", helmChartMediaType),
+				)
+				req.Header.Set("User-Agent", "AzureContainerRegistry/1.0.0")
```
krancour:
```suggestion
				req.Header.Set("User-Agent", acrUserAgentPrefix)
```

internal/webhook/external/azure_test.go
krancour: We also use "User-Agent" in a lot of places. It's unfortunately that there aren't constants for common headers in the `http` package. Let's add a constant for this in our `xhttp` package.

internal/webhook/external/azure_test.go
```diff
@@ -0,0 +1,257 @@
...
+				indexer.WarehousesBySubscribedURLsField,
+				indexer.WarehousesBySubscribedURLs,
+			).Build(),
+			req: func() *http.Request {
+				req := httptest.NewRequest(
+					http.MethodPost,
+					testURL,
+					newAzurePayload("git.push", ""),
+				)
+				req.Header.Set("User-Agent", "VSServices/1.0.0")
```
krancour:
```suggestion
				req.Header.Set("User-Agent", azureDevOpsUserAgentPrefix)
```

internal/webhook/external/azure_test.go
```diff
@@ -0,0 +1,257 @@
...
+				},
+			}).getHandler(requestBody)(w, testCase.req())
+
+			testCase.assertions(t, w)
+		})
+	}
+}
+
+func newAzurePayload(event, mediaType string) *bytes.Buffer {
+	switch event {
```
krancour, line 249: This switch also can use constants I suggested.

```suggestion
	switch event {
```

internal/webhook/external/azure.go
krancour: Looking at the tests and also #4536, I think `git.push`? That should probably be a constant as well.

internal/webhook/external/azure.go
```diff
@@ -0,0 +1,222 @@
...
+		)
+		return
+	}
+
+	switch event.Action {
+	case acrPushEvent:
+		repoURL := normalizeOCIRepoURL(
+			fmt.Sprintf("%s/%s", event.Request.Host, event.Target.Repository),
+			event.Target.MediaType,
+		)
```
krancour, line 131: Is there media type info in the payload? We should try to determine if the push was an image or a chart because their URLs use different normalization processes.

krancour: Nevermind. This is using the new function already. I'm blind.

krancour (approved):

PR #4524, opened 2025-06-27
fix: replace release trigger from created to published

fuskovic:


krancour:
We discovered the hard way today that when a draft release is created, the release created event doesn't fire. It also doesn't fire when the draft is finally published. (Riddle me this. When can you create a release without _creating a release?_)

This seems to be the correct fix.

LGTM

Thank you for jumping on this @fuskovic!

krancour (approved):

PR #4533, opened 2025-06-30
feat(mgmt-controller): rely on GC of NS through owner ref and finalizer

hiddeco:
This avoids requiring explicit delete permissions for namespaces for the management controller role by relying purely on the cascade delete for owner references, and the existence of the finalizer to still allow this reference to be dropped when the namespace has to be kept around.

Follow up to: #4229 #4499

### DIFF
--- a/charts/kargo/templates/management-controller/cluster-roles.yaml
+++ b/charts/kargo/templates/management-controller/cluster-roles.yaml
@@ -13,7 +13,6 @@ rules:
   - namespaces
   verbs:
   - create
-  - delete
   - get
   - list
   - patch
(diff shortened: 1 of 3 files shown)


krancour:
To verify I understand this correctly, instead of removing the owner reference _or_ deleting the namespace, we're just conditionally removing the owner reference and then allowing everything else to sort itself out.

🤦‍♂️ It's so obvious. idk how I overlooked this at the time #4229 was merged.

I am racking my brain right now to remember if there was actually a reason it wasn't done this way.

krancour (approved):
LGTM after merge conflicts are resolved.

Issue #3606, opened 2025-03-04
Mutable tag subscriptions are rejected on apply


remi-gelinas:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

When subscribing to an image repository with mutable tags, the Kargo admission webhook rejects image subscriptions with the `semverConstraint` field containing the mutable tag name (described in the [docs](https://docs.kargo.io/user-guide/how-to-guides/working-with-warehouses#image-selection-strategies)). As far as I can tell, this means v1.3.0 fully breaks support for mutable tag subscriptions.

EDIT: This is also present on v1.2.3

# Steps to Reproduce

Apply a `Warehouse` with an image subscription listening to new digests from a mutable tag:
```
---
apiVersion: kargo.akuity.io/v1alpha1
kind: Warehouse
metadata:
  name: image
spec:
  subscriptions:
    - image:
        repoURL: <any valid repo>
        imageSelectionStrategy: Digest
        semverConstraint: main # Any mutable tag
```

It will be rejected by the Kargo admissions webhook because the `semverConstraint` value is not valid:

```
 one or more objects failed to apply, reason: admission webhook "warehouse.kargo.akuity.io" denied the request: Warehouse.kargo.akuity.io "image" is invalid: spec.subscriptions[0].image.semverConstraint: Invalid value: "main"
```

# Version

```
{
  "Version": "v1.3.0",
  "BuildDate": "2025-02-25T20:20:33Z",
  "GitCommit": "f645f54c5cc72efc122ffd602544eec1bdee1656",
  "GitTreeDirty": false,
  "GoVersion": "go1.24.0",
  "Compiler": "gc",
  "Platform": "linux/amd64"
}
```

# Logs

```
Paste any relevant application logs here.
```

remi-gelinas:
I figured it probably just slipped through as it's not a recommended use case at all. Unfortunately, I've got a few legacy projects using Kargo that *do* publish to mutable tags, and I appreciated that Kargo did at least support the use case.

Is there any known workaround for v1.3.0 @krancour, or should I be looking to roll back my upgrade to keep feature parity until something is figured out?

krancour:
> should I be looking to roll back my upgrade to keep feature parity until something is figured out?

[...]

PR #4541, opened 2025-07-02
refactor: deprecate `SemverConstraint` field and introduce generic `constraint` field

nitishfy:
Closes https://github.com/akuity/kargo/issues/3606

## After

works as expected
```
subscriptions:
  - image:
      imageSelectionStrategy: Digest
      constraint: latest
```

Error: `digest selection strategy requires a constraint`
```
subscriptions:
  - image:
      imageSelectionStrategy: Digest
      semVerConstraint: latest
```

Error: `Invalid value: "latest"`
```
subscriptions:
  - image:
      imageSelectionStrategy: SemVer
      semVerConstraint: latest
```

---


internal/webhook/kubernetes/warehouse/webhook.go
```diff
@@ -207,12 +207,21 @@ func (w *webhook) validateImageSub(
...
+	// semVer image strategy is used
+	if sub.SemverConstraint != "" && sub.ImageSelectionStrategy == kargoapi.ImageSelectionStrategySemVer {
+		if err := validateSemverConstraint(
+			f.Child("semverConstraint"),
+			sub.SemverConstraint,
+		); err != nil {
+			errs = field.ErrorList{err}
+		}
 	}
+	// TODO(@nitishfy) we should conditionally validate the
```
nitishfy:
```suggestion
	// TODO(@nitishfy) we should conditionally validate the
```
Apart from this, the docs needs to be updated. I'm thinking to resolve the review comments first before proceeding to docs straight.

api/v1alpha1/warehouse_types.go
```diff
@@ -363,20 +363,27 @@ type ImageSubscription struct {
...
 	//
+	// Deprecated: Use constraint instead.
+	//
 	// +kubebuilder:validation:Optional
 	SemverConstraint string `json:"semverConstraint,omitempty" protobuf:"bytes,4,opt,name=semverConstraint"`
+	// Constraint specifies constraints on what new image versions are permissible.
+	// This field takes precedence over the deprecated SemverConstraint field
+	// and can be used with any image strategy.
+	// +kubebuilder:validation:Optional
+	Constraint string `json:"constraint,omitempty" protobuf:"bytes,5,opt,name=constraint"`
```
hiddeco: This should be: 

```
`json:"constraint,omitempty" protobuf:"bytes,11,opt,name=constraint"`
```

Which would prevent all the other fields from getting renumbered, as this heavily breaks API compatibility with versions before this change.

internal/controller/warehouses/images.go
```diff
@@ -130,10 +130,20 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
-			f,
-			"semverConstraint", sub.SemverConstraint,
-		)
+		// if the constraint field has been used, append its logs
+		// otherwise fallback to appending deprecated field logs
+		// only if semVer strategy is used
+		if sub.Constraint != "" {
+			f = append(
+				f,
+				"Constraint", sub.Constraint,
```
hiddeco:
```suggestion
				"constraint", sub.Constraint,
```

api/v1alpha1/warehouse_types.go
```diff
@@ -363,20 +363,27 @@ type ImageSubscription struct {
...
 	// AllowTags is a regular expression that can optionally be used to limit the
 	// image tags that are considered in determining the newest version of an
 	// image. This field is optional.
 	//
 	// +kubebuilder:validation:Optional
-	AllowTags string `json:"allowTags,omitempty" protobuf:"bytes,5,opt,name=allowTags"`
+	AllowTags string `json:"allowTags,omitempty" protobuf:"bytes,6,opt,name=allowTags"`
```
fykaa: changing the protobuf field number is technically a breaking change

there's a risk that older clients/deserializers might misinterpret the wire data, leading to hard-to-debug issues

I read this last week, it's a very interesting story about the same: https://medium.com/@sanhdoan/protocol-buffers-the-importance-of-reserved-and-deprecated-fields-bbb8d7d3211d

api/v1alpha1/warehouse_types.go
krancour: As a rule, do not write _any_ of the `protobuf` part of a struct tag by hand. Codegen will complete it for you and is much less prone to making this particular mistake.

krancour: I think the description of this field needs to be expanded to describe what is expected/supported in this field (or whether it's even used at all) for each of the selection strategies.

api/v1alpha1/warehouse_types.go
krancour: Just to be certain it is not overlooked, this applies to _all_ the fields that have incorrectly been re-numbered.

internal/controller/warehouses/images.go
```diff
@@ -130,10 +130,20 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
+			)
+		} else if sub.SemverConstraint != "" && sub.ImageSelectionStrategy == kargoapi.ImageSelectionStrategySemVer {
+			f = append(
+				f,
+				"semverConstraint", sub.SemverConstraint,
+			)
+		}
```
krancour: The code would be much easier to follow if you replaced this conditional by splitting the switch statement's current single case for `ImageSelectionStrategySemVer` and `ImageSelectionStrategyDigest` into two separate cases.

internal/controller/warehouses/images.go
```diff
@@ -162,3 +172,16 @@ func imageSelectorForSubscription(
...
+func determineConstraint(sub *kargoapi.ImageSubscription) string {
+	if sub.Constraint != "" {
+		return sub.Constraint
+	}
+	if sub.ImageSelectionStrategy == kargoapi.ImageSelectionStrategySemVer && sub.SemverConstraint != "" {
+		return sub.SemverConstraint
+	}
+
+	return ""
+}
```
krancour: This isn't really worthy of a dedicated function. This is a miniscule amount of not-very-complex logic that is used in only one place. Just put this in `imageSelectorForSubscription()`.

internal/webhook/kubernetes/warehouse/webhook.go
```diff
@@ -207,12 +207,21 @@ func (w *webhook) validateImageSub(
...
+	// Give precedence to the constraint field
+	// However, validate semVerConstraint field if
+	// semVer image strategy is used
+	if sub.SemverConstraint != "" && sub.ImageSelectionStrategy == kargoapi.ImageSelectionStrategySemVer {
+		if err := validateSemverConstraint(
+			f.Child("semverConstraint"),
+			sub.SemverConstraint,
+		); err != nil {
+			errs = field.ErrorList{err}
+		}
```
krancour: To avert confusion, we should regard the old field and the new field as mutually exclusive. You should validate that they're not _both_ non-empty. Then if _one_ is non-empty, you can move on to validating that one.

The mutual exclusivity part can possibly be done with CEL.

cc @hiddeco

internal/webhook/kubernetes/warehouse/webhook.go
krancour: This is important. Did you mean to have addressed this already? Or is this PR supposed to be in a draft state?

internal/controller/warehouses/images.go
```diff
@@ -130,10 +130,20 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
 	}
 	switch sub.ImageSelectionStrategy {
 	case kargoapi.ImageSelectionStrategySemVer, kargoapi.ImageSelectionStrategyDigest:
-		f = append(
-			f,
-			"semverConstraint", sub.SemverConstraint,
-		)
+		// if the constraint field has been used, append its logs
+		// otherwise fallback to appending deprecated field logs
+		// only if semVer strategy is used
```
krancour: I'm not suggesting any specific rewording here because this comment probably needs to be rewritten anyway or may go away entirely after [this](https://github.com/akuity/kargo/pull/4541/files#r2180635097) is addressed, but this is just a reminder to pay attention to grammar, capitalization, and punctuation. Some projects may be more lenient about such things, but we've been very good about it up to now, so if for no other reason than consistency, we should continue.

internal/webhook/kubernetes/warehouse/webhook.go
nitishfy: please see now.

internal/webhook/kubernetes/warehouse/webhook.go
hiddeco: Something like the following should work:

```go
// ImageSubscription defines a subscription to an image repository.
//
// +kubebuilder:validation:XValidation:message="ImageSubscription cannot have both semverConstraint and constraint set",rule="!(has(self.semverConstraint) && has(self.constraint))"
type ImageSubscription struct {
```

api/v1alpha1/warehouse_types.go
```diff
@@ -363,8 +365,15 @@ type ImageSubscription struct {
 	// changes.
 	// More info: https://github.com/masterminds/semver#checking-version-constraints
 	//
+	// Deprecated: Use constraint instead.
+	//
 	// +kubebuilder:validation:Optional
-	SemverConstraint string `json:"semverConstraint,omitempty" protobuf:"bytes,4,opt,name=semverConstraint"`
+	SemverConstraint string `json:"semverConstraint,omitempty" protobuf:"bytes,12,opt,name=semverConstraint"`
```
hiddeco:
```suggestion
	SemverConstraint string `json:"semverConstraint,omitempty" protobuf:"bytes,4,opt,name=semverConstraint"`
```

api/v1alpha1/warehouse_types.go
```diff
@@ -376,7 +385,7 @@ type ImageSubscription struct {
 	// supported yet. This field is optional.
 	//
 	// +kubebuilder:validation:Optional
-	IgnoreTags []string `json:"ignoreTags,omitempty" protobuf:"bytes,6,rep,name=ignoreTags"`
+	IgnoreTags []string `json:"ignoreTags,omitempty" protobuf:"bytes,7,rep,name=ignoreTags"`
```
hiddeco:
```suggestion
	IgnoreTags []string `json:"ignoreTags,omitempty" protobuf:"bytes,6,rep,name=ignoreTags"`
```

api/v1alpha1/warehouse_types.go
```diff
@@ -387,11 +396,11 @@ type ImageSubscription struct {
 	// not unheard of.
 	//
 	// +kubebuilder:validation:Optional
-	Platform string `json:"platform,omitempty" protobuf:"bytes,7,opt,name=platform"`
+	Platform string `json:"platform,omitempty" protobuf:"bytes,8,opt,name=platform"`
```
hiddeco:
```suggestion
	Platform string `json:"platform,omitempty" protobuf:"bytes,7,opt,name=platform"`
```

api/v1alpha1/warehouse_types.go
```diff
@@ -387,11 +396,11 @@ type ImageSubscription struct {
...
 	// InsecureSkipTLSVerify specifies whether certificate verification errors
 	// should be ignored when connecting to the repository. This should be enabled
 	// only with great caution.
-	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty" protobuf:"varint,8,opt,name=insecureSkipTLSVerify"`
+	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty" protobuf:"varint,9,opt,name=insecureSkipTLSVerify"`
```
hiddeco:
```suggestion
	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty" protobuf:"varint,8,opt,name=insecureSkipTLSVerify"`
```

api/v1alpha1/warehouse_types.go
```diff
@@ -401,7 +410,7 @@ type ImageSubscription struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=100
 	// +kubebuilder:default=20
-	DiscoveryLimit int32 `json:"discoveryLimit,omitempty" protobuf:"varint,9,opt,name=discoveryLimit"`
+	DiscoveryLimit int32 `json:"discoveryLimit,omitempty" protobuf:"varint,11,opt,name=discoveryLimit"`
```
hiddeco:
```suggestion
	DiscoveryLimit int32 `json:"discoveryLimit,omitempty" protobuf:"varint,9,opt,name=discoveryLimit"`
```

api/v1alpha1/warehouse_types.go
```diff
@@ -363,8 +365,15 @@ type ImageSubscription struct {
...
+	// Deprecated: Use constraint instead.
+	//
 	// +kubebuilder:validation:Optional
-	SemverConstraint string `json:"semverConstraint,omitempty" protobuf:"bytes,4,opt,name=semverConstraint"`
+	SemverConstraint string `json:"semverConstraint,omitempty" protobuf:"bytes,12,opt,name=semverConstraint"`
+	// Constraint specifies constraints on what new image versions are permissible.
+	// This field takes precedence over the deprecated SemverConstraint field
+	// and can be used with any image strategy.
+	// +kubebuilder:validation:Optional
+	Constraint string `json:"constraint,omitempty" protobuf:"bytes,13,opt,name=constraint"`
```
hiddeco:
```suggestion
	Constraint string `json:"constraint,omitempty" protobuf:"bytes,11,opt,name=constraint"`
```

hiddeco (commented):
Protobuf numbering is still off.

internal/controller/warehouses/images.go
```diff
@@ -147,12 +156,18 @@ func imageSelectorForSubscription(
 	sub kargoapi.ImageSubscription,
 	creds *image.Credentials,
 ) (image.Selector, error) {
+	var constraint string
+	if sub.Constraint != "" {
+		constraint = sub.Constraint
+	} else if sub.ImageSelectionStrategy == kargoapi.ImageSelectionStrategySemVer && sub.SemverConstraint != "" {
+		constraint = sub.SemverConstraint
+	}
```
hiddeco:
```suggestion
	constraint := sub.Constraint
	if constraint == "" && sub.ImageSelectionStrategy == kargoapi.ImageSelectionStrategySemVer {
		constraint = sub.SemverConstraint
	}
```

In Go, there is a common pattern of "assign default, then override if needed".

By following this pattern, you simplify the logic by reducing from 7 lines to 4 lines while ordering the conditions more intuitively by checking if the primary value is missing first, then checking fallback conditions.

This makes it easier for developers to understand (at a later moment) because it follows the natural thought process of "do I have what I need, and if not, what are my alternatives?"

internal/webhook/kubernetes/warehouse/webhook.go
```diff
@@ -207,11 +207,25 @@ func (w *webhook) validateImageSub(
...
+			); err != nil {
+				errs = field.ErrorList{err}
+			}
+		} else if sub.Constraint != "" {
+			if err := validateSemverConstraint(
+				f.Child("constraint"),
+				sub.Constraint,
+			); err != nil {
+				errs = field.ErrorList{err}
+			}
```
hiddeco:
```suggestion
		if sub.SemverConstraint != "" {
			if err := validateSemverConstraint(
				f.Child("semverConstraint"),
				sub.SemverConstraint,
			); err != nil {
				errs = field.ErrorList{err}
			}
		}

		if sub.Constraint != "" {
			if err := validateSemverConstraint(
				f.Child("constraint"),
				sub.Constraint,
			); err != nil {
				errs = field.ErrorList{err}
			}
```

Now that you have CEL in place, you can just check the value if it is set. No need to complicate this with `else if` type of things.

Additionally, I would: `errs = append(errs, err)` to prevent the list from being reset if logic is ever introduced above this block.

api/v1alpha1/warehouse_types.go
```diff
@@ -306,6 +306,8 @@ type GitSubscription struct {
 }
 
 // ImageSubscription defines a subscription to an image repository.
+//
+// +kubebuilder:validation:XValidation:message="ImageSubscription cannot have both semverConstraint and constraint set",rule="!(has(self.semverConstraint) && has(self.constraint))"
```
krancour: I feel as if this would be the more standard language for expressing this:

```suggestion
// +kubebuilder:validation:XValidation:message="semverConstraint and constraint fields are mutually exclusive",rule="!(has(self.semverConstraint) && has(self.constraint))"
```

However, my experience with CEL is low up to this point, so cc @hiddeco because idk what best practices we should observe here. (For instance, as with godocs, is it highly encouraged that the first word of the error is exactly the name of the type?)

internal/controller/warehouses/images.go
```diff
@@ -129,11 +129,20 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
-	case kargoapi.ImageSelectionStrategySemVer, kargoapi.ImageSelectionStrategyDigest:
-		f = append(
-			f,
-			"semverConstraint", sub.SemverConstraint,
-		)
+	case kargoapi.ImageSelectionStrategySemVer:
+		if sub.Constraint != "" {
+			f = append(
+				f,
+				"constraint", sub.Constraint)
```
krancour: idk why the original code split this command across multiple lines, but let's not blindly maintain that. Our goal should be maximum readability. In this case, that means this:

```suggestion
			f = append(f, "constraint", sub.Constraint)
```

internal/controller/warehouses/images.go
```diff
@@ -129,11 +129,20 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
+		} else if sub.SemverConstraint != "" {
+			f = append(
+				f,
+				"semverConstraint", sub.SemverConstraint)
```
krancour:
```suggestion
			f = append(f, "semverConstraint", sub.SemverConstraint)
```

internal/controller/warehouses/images.go
```diff
@@ -129,11 +129,20 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
-		f = append(
-			f,
-			"semverConstraint", sub.SemverConstraint,
-		)
+	case kargoapi.ImageSelectionStrategySemVer:
+		if sub.Constraint != "" {
+			f = append(
+				f,
+				"constraint", sub.Constraint)
+		} else if sub.SemverConstraint != "" {
```
krancour: This change would preserve the original behavior where lack of a constraint is still explicitly recorded, which I think is a beneficial detail.

```suggestion
		} else {
```

internal/controller/warehouses/images.go
```diff
@@ -129,11 +129,20 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
+			f = append(
+				f,
+				"semverConstraint", sub.SemverConstraint)
+		}
+	case kargoapi.ImageSelectionStrategyDigest:
+		if sub.Constraint != "" {
+			f = append(f, "constraint", sub.Constraint)
+		}
```
krancour, line 145: Similar to [this comment](https://github.com/akuity/kargo/pull/4541/files#r2185696503), this wasn't conditional before and I think there's a benefit to continuing to explicitly show in logs when the strategy is unconstrained.

```suggestion
		if sub.Constraint != "" {
			f = append(f, "constraint", sub.Constraint)
		}
```

internal/webhook/kubernetes/warehouse/webhook.go
krancour: @nitishfy for clarity, @hiddeco's last suggestion isn't included in his diff.

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -261,8 +261,9 @@ func TestValidateSpec(t *testing.T) {
 							RepoURL: "bogus",
 						},
 						Image: &kargoapi.ImageSubscription{
-							SemverConstraint: "bogus",
-							Platform:         "bogus",
+							ImageSelectionStrategy: kargoapi.ImageSelectionStrategySemVer,
```
krancour: I don't mind this line having been added for clarity, but just as a fyi, when no strategy is specified, behavior default to the semver strategy.

krancour:
I took the liberty of converting this to draft so that it is not accidentally merged without docs. Arguably, docs are _the_ most important part of a deprecation, so they're not something I would allow to be forgone now and addressed in a follow-up.

internal/webhook/kubernetes/warehouse/webhook_test.go
nitishfy: If I remove the strategy specified, it defaults to an empty string, due to which the test case will fail since `validateImageSub()` doesn't handle a scenario when a strategy has not been specified. I thought this might be configured on some other component of Kargo. Should I add an additional check stating that if the strategy is not specified, the default strategy should point to semver?

internal/webhook/kubernetes/warehouse/webhook.go
nitishfy: updated. PTAL.

api/v1alpha1/warehouse_types.go
hiddeco: There is no real guidance on what language to use, but this is what I observed being used in other well-known projects in the space.

internal/webhook/kubernetes/warehouse/webhook_test.go
krancour: > If I remove the strategy specified, it defaults to an empty string, due to which the test case will fail since validateImageSub() doesn't handle a scenario when a strategy has not been specified.

If you made changes to the validation method that are supposed to be non-breaking and it broke this test, then you should fix the method instead of altering the test case.

As it stands, your changes to the validation method are going to break everyone who is not explicitly specifying a selection strategy and the test was trying to tell you that.

krancour:
[This](https://github.com/akuity/kargo/pull/4541/files#r2193212266) need fixing and then please add docs.

internal/webhook/kubernetes/warehouse/webhook_test.go
nitishfy: Please take a look now.

internal/webhook/kubernetes/warehouse/webhook.go
```diff
@@ -207,14 +207,19 @@ func (w *webhook) validateImageSub(
 	seen uniqueSubSet,
 ) field.ErrorList {
 	var errs field.ErrorList
+	// If no strategy has been specified
+	// default to the semVer strategy
```
krancour, line 211:
```suggestion
	// If no strategy has been specified default to the SemVer strategy
```

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -356,16 +420,17 @@ func TestValidateSubs(t *testing.T) {
 			},
 		},
 		{
-			name: "invalid subscriptions",
+			// TODO(@nitishfy) we can remove this test case once
+			// this deprecated field is removed
```
krancour, line 424:
```suggestion
			// TODO(nitishfy): Remove this test case once this deprecated field is
			// removed.
```

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -445,15 +569,16 @@ func TestValidateSub(t *testing.T) {
 		assertions func(*testing.T, kargoapi.RepoSubscription, field.ErrorList)
 	}{
 		{
-			name: "invalid subscription",
+			// TODO(@nitishfy) we can remove this test case once
+			// this deprecated field is removed
```
krancour, line 573:
```suggestion
			// TODO(nitishfy): Remove this test case once this deprecated field is
			// removed.
```

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -596,12 +778,13 @@ func TestValidateImageSub(t *testing.T) {
 		assertions func(*testing.T, field.ErrorList)
 	}{
 		{
-			name: "invalid",
+			// TODO(@nitishfy) we can remove this test case once
+			// this deprecated field is removed
```
krancour, line 782:
```suggestion
			// TODO(nitishfy): Remove this test case once this deprecated field is
			// removed.
```

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -253,17 +253,18 @@ func TestValidateSpec(t *testing.T) {
 			},
 		},
 		{
-			name: "invalid",
+			// TODO(@nitishfy) we can remove this test case once
+			// this deprecated field is removed
+			name: "invalid when using SemVerConstraint field (deprecated)",
```
krancour, line 258:
```suggestion
			name: "invalid SemVerConstraint",
```

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -316,6 +317,69 @@ func TestValidateSpec(t *testing.T) {
 				)
 			},
 		},
+		{
+			name: "invalid when using constraint field",
```
krancour: It's unnecessary to create dedicated test cases for testing individual fields and it adds quite a bit of clutter.

Amend the original invalid and valid test cases with additional invalid or valid subscriptions, respectively, and update assertions accordingly.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -109,6 +109,13 @@ strategies are:
   semantically greatest tag. All tags that are not valid semantic versions are
   ignored.
 
+    :::note
+    Starting Kargo v1.7, the `semverConstraint` field is deprecated for
+    container image subscriptions. It is recommended to use the `constraint`
+    field instead. When using the `SemVer` strategy, `constraint` should be a
+    valid semantic version range.
+    :::
```
krancour: I would reconsider how you present this information.

When `semverConstraint` is eventually removed, and so too is this note, the previous paragraph (starting on line 106) will require adjustment to reference `constraint` instead of `semverConstraint` and to explain that `constraint` must be a valid semantic version range when used with the `SemVer` strategy.

It would be better to make those revisions _now_. Apart from guarding against the possibility of eventually removing the note and forgetting to revise the paragraph above, it makes more sense to show how things _should_ be done right off the bat instead of showing how things _shouldn't_ be done and immediately following up with a note saying not to.

Revise the paragraph starting on line 106 so that it's accurate both now and into the future. Update the note to mention the existence of the `semverConstraint` field that is now deprecated and scheduled for removal in v1.9.0.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -159,7 +166,9 @@ strategies are:
 
     __Unintuitively, the mutable tag name must be specified using the
     `semverConstraint` field.__ Importantly, the _digest_ will change every time
-    the tag is updated.
+    the tag is updated. Since the `semverConstraint` field is deprecated
+    starting v1.7, it is recommended to use the `constraint` field that should
+    be set to a valid mutable tag name.
```
krancour: This is another case where you should show the _right_ way to do things first.

internal/webhook/kubernetes/warehouse/webhook.go
```diff
@@ -207,11 +207,32 @@ func (w *webhook) validateImageSub(
...
+
+		if sub.Constraint != "" {
+			if err := validateSemverConstraint(
+				f.Child("constraint"),
+				sub.Constraint,
+			); err != nil {
+				errs = append(errs, err)
+			}
+		}
 	}
```
krancour: A bit more straightforward:

```suggestion
	var errs field.ErrorList
	if sub.ImageSelectionStrategy == kargoapi.ImageSelectionStrategySemVer || sub.ImageSelectionStrategy == "" {
		var err *field.Error
		// Validate the Constraint field XOR the deprecated SemverConstraint field
		// is a valid semantic version range:
		if sub.Constraint != "" {
			err = validateSemverConstraint(f.Child("constraint"), sub.Constraint)
		} else if sub.SemverConstraint != "" { // nolint:staticcheck
			err = validateSemverConstraint(
				f.Child("semverConstraint"),
				sub.SemverConstraint, // nolint:staticcheck
			)
		}
		if err != nil {
			errs = append(errs, err)
		}
	}
```

internal/controller/warehouses/images.go
```diff
@@ -129,16 +129,18 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
-	case kargoapi.ImageSelectionStrategySemVer, kargoapi.ImageSelectionStrategyDigest:
-		f = append(
-			f,
-			"semverConstraint", sub.SemverConstraint,
-		)
+	case kargoapi.ImageSelectionStrategySemVer:
+		if sub.Constraint != "" {
+			f = append(f, "constraint", sub.Constraint)
+		} else {
+			f = append(f, "semverConstraint", sub.SemverConstraint) //nolint:staticcheck
```
krancour, line 136: These k/v pairs are adding context to all the discovery logging that's about to happen. If everything else has been done correctly, the discovery process is only aware of a constraint and not really aware of _how_ that constraint came to be set. So I think I would probably use `constraint` as the key here instead of `semverConstraint`.

```suggestion
			f = append(f, "constraint", sub.SemverConstraint) //nolint:staticcheck
```

internal/controller/warehouses/images.go
```diff
@@ -129,16 +129,18 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
 	switch sub.ImageSelectionStrategy {
-	case kargoapi.ImageSelectionStrategySemVer, kargoapi.ImageSelectionStrategyDigest:
-		f = append(
-			f,
-			"semverConstraint", sub.SemverConstraint,
-		)
+	case kargoapi.ImageSelectionStrategySemVer:
+		if sub.Constraint != "" {
+			f = append(f, "constraint", sub.Constraint)
+		} else {
```
krancour, line 135:
```suggestion
		} else if sub.SemverConstraint != "" { // nolint:staticcheck
```

internal/controller/warehouses/images.go
```diff
@@ -147,12 +149,16 @@ func imageSelectorForSubscription(
 	sub kargoapi.ImageSubscription,
 	creds *image.Credentials,
 ) (image.Selector, error) {
+	constraint := sub.Constraint
+	if constraint == "" && sub.ImageSelectionStrategy == kargoapi.ImageSelectionStrategySemVer {
```
krancour: This conditional is actually incorrect.

The legacy behavior was that `SemverConstraint` defined the constraints regardless of the selection strategy.

As written, you're falling back to looking at `SemverConstraint` only if `Constraint` is empty _and_ the selection strategy is Semver.

This means anyone using the another strategy (notably Digest) and dependent on the legacy behavior would _not_ pick up their existing constraint from the SemverConstraint field.

It should be this:

```suggestion
	if constraint == "" {
```

internal/controller/warehouses/images.go
```diff
@@ -129,16 +129,18 @@ func imageDiscoveryLogFields(sub kargoapi.ImageSubscription) []any {
...
+	case kargoapi.ImageSelectionStrategySemVer:
+		if sub.Constraint != "" {
+			f = append(f, "constraint", sub.Constraint)
+		} else {
+			f = append(f, "semverConstraint", sub.SemverConstraint) //nolint:staticcheck
+		}
+	case kargoapi.ImageSelectionStrategyDigest:
+		if sub.Constraint != "" {
+			f = append(f, "constraint", sub.Constraint)
+		}
```
krancour: Same explanation as [here](https://github.com/akuity/kargo/pull/4541/files#r2195308012):

```suggestion
	case kargoapi.ImageSelectionStrategySemVer, kargoapi.ImageSelectionStrategyDigest, "":
		constraint := sub.Constraint
		if constraint == "" {
			constraint = sub.SemverConstraint // nolint:staticcheck
		}
		f = append(f, "constraint", sub.Constraint)
```

api/v1alpha1/warehouse_types.go
```diff
@@ -363,8 +365,15 @@ type ImageSubscription struct {
 	// changes.
 	// More info: https://github.com/masterminds/semver#checking-version-constraints
 	//
+	// Deprecated: Use constraint instead.
```
krancour: In other spots where we've done this, I notice we've capitalized the field name to match Go syntax, rather than how it presents in YAML.

```suggestion
	// Deprecated: Use Constraint instead. This field will be removed in v1.9.0.
```

api/v1alpha1/warehouse_types.go
```diff
@@ -363,8 +365,15 @@ type ImageSubscription struct {
...
+	//
 	// +kubebuilder:validation:Optional
 	SemverConstraint string `json:"semverConstraint,omitempty" protobuf:"bytes,4,opt,name=semverConstraint"`
+	// Constraint specifies constraints on what new image versions are permissible.
+	// This field takes precedence over the deprecated SemverConstraint field
+	// and can be used with any image strategy.
+	// +kubebuilder:validation:Optional
```
krancour:
```suggestion
	// Constraint specifies constraints on what new image versions are
	// permissible. Acceptable values for this field vary contextually by
	// ImageSelectionStrategy. The field is optional and is ignored by some
	// strategies. When non-empty, the value in this field takes precedence over
	// the value of the deprecated SemverConstraint field.
	//
	// +kubebuilder:validation:Optional
```

api/v1alpha1/warehouse_types.go
```diff
@@ -306,6 +306,8 @@ type GitSubscription struct {
 }
 
 // ImageSubscription defines a subscription to an image repository.
+//
+// +kubebuilder:validation:XValidation:message="semverConstraint and constraint fields are mutually exclusive",rule="!(has(self.semverConstraint) && has(self.constraint))"
```
krancour, line 310: Note additional doc changes required on lines +326 and +327.

krancour: We weren't validating it before, but we should have been... we should validate that if the selection strategy is Digest, that a constraint (or a semver constraint, for now) has also been specified. Should be doable with CEL, I believe.

internal/controller/warehouses/images.go
nitishfy: Now that we have refactored the logic, I don't think it's required anymore. WDYT?

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
nitishfy: PTAL now.

internal/controller/warehouses/images.go
krancour: The original, unintuitive, behavior was that `SemverConstraint` was used to specify a mutable tag for the `Digest` selection strategy. Until `SemverConstraint` is removed, that original behavior needs to remain _unless_ the user has transitioned to using the new, more general `Constraint` field. Please have a closer look at this conditional and try to understand how it does not facilitate that requirement.

krancour: Hmmm... although that original behavior perhaps rarely worked as advertised since the validating webhook rejected anything that wasn't a valid semver constraint.

But there are also edge cases where it would have worked if the mutable tag happened to be a valid semver constraint. That sounds silly, but it can happen. You need look no farther than the official nginx image to find tags like `1.0` that are mutable. As new 1.0.x are released, they are _also_ pushed to `1.0`.

So it's for the benefit of these edge cases that we need to fall back to `SemverConstraint` regardless of the strategy.

internal/webhook/kubernetes/warehouse/webhook_test.go
krancour: This still needs to be fixed.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
krancour: Your changes did not address my concern.

nitishfy: sorry, i missed modifying the existing paragraph. Can you please review now?

internal/controller/warehouses/images.go
nitishfy: Hmm, that makes sense. I've pushed the changes.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
nitishfy: Please see now.

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -402,14 +434,16 @@ func TestValidateSub(t *testing.T) {
 		assertions func(*testing.T, kargoapi.RepoSubscription, field.ErrorList)
 	}{
 		{
-			name: "invalid subscription",
+			// TODO(@nitishfy) remove this testcase once the deprecated field is removed
+			name: "invalid subscription when using semverconstraint(deprecated) field",
```
nitishfy, line 438:
```suggestion
			name: "invalid subscription when using semverconstraint(deprecated) field",
```
Added a separate test case for the deprecated field in order to stay consistent with the current test case architecture. Currently, it simply takes a subscription. Hence, I thought rather than modifying the subscription field to a slice, i can simply add another test case assuring that the test case structure remains consistent. I'm honestly not sure if it's the best approach here although modifying the subs to a slice is pretty straightforward approach but I'm happy to adjust changes as suggested.

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -589,6 +683,45 @@ func TestValidateImageSub(t *testing.T) {
 				)
 			},
 		},
+		{
+			name: "invalid subscription when using constraint field",
```
nitishfy: Same comment as [here](https://github.com/akuity/kargo/pull/4541/files#r2221600082).

internal/webhook/kubernetes/warehouse/webhook_test.go
```diff
@@ -458,6 +492,64 @@ func TestValidateSub(t *testing.T) {
...
```
nitishfy: Added a separate test case for the deprecated field in order to stay consistent with the current test case architecture. Currently, it simply takes a subscription. Hence, I thought rather than modifying the subscription field to a slice, i can simply add another test case assuring that the test case structure remains consistent. I'm honestly not sure if it's the best approach here although modifying the subs to a slice is pretty straightforward approach. I'm happy to adjust changes as suggested.

nitishfy:
One thing to note is that I accidentally messed up my git commit history and some of the commits got deleted too. I've added new commits that contains all the fixes for the changes that were requested. 
cc @krancour / @hiddeco

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -104,11 +104,19 @@ field specifies a method for selecting the desired image. The available
...
   constraint specified, the strategy simply selects the image with the
   semantically greatest tag. All tags that are not valid semantic versions are
   ignored.
 
+    :::note
+    Prior to Kargo v1.7, the `semverConstraint` field was used
+    to define semantic version constraints for container image
+    subscriptions. This field is now deprecated and will be removed
+    in v1.9.0. Users should use the `constraint` field instead, which
+    supports the same syntax and semantics when used with the `SemVer` strategy.
```
krancour:
```suggestion
    If the `constraint` field is empty, any constraints defined by the
    deprecated `semverConstraint` field will be applied instead. The
    `semverConstraint` field will be removed in v1.9.0.
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -104,11 +104,19 @@ field specifies a method for selecting the desired image. The available
 strategies are:
 
 - `SemVer`: Selects the image with the tag that best matches a semantic
-  versioning constraint specified by the `semverConstraint` field. With no
+  versioning constraint specified by the `constraint` field. With no
```
krancour, line 107:
```suggestion
  versioning constraint specified by the `constraint` field. If no such
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -104,11 +104,19 @@ field specifies a method for selecting the desired image. The available
...
   constraint specified, the strategy simply selects the image with the
```
krancour:
```suggestion
  versioning constraint specified by the `constraint` field. If no such
  constraint is specified, the strategy simply selects the image with the
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -157,9 +165,17 @@ strategies are:
 - `Digest`: This selects the image _currently_ referenced by some "mutable tag,"
    such as `latest`.
```
krancour:
```suggestion
- `Digest`: This selects the image _currently_ referenced by some "mutable tag"
   (such as `latest`) specified by the `constraint` field.
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -157,9 +165,17 @@ strategies are:
...
 
-    __Unintuitively, the mutable tag name must be specified using the
-    `semverConstraint` field.__ Importantly, the _digest_ will change every time
-    the tag is updated.
+    The mutable tag name must be specified using the `constraint` field. This field
+    should be set to a valid tag name (e.g., "latest"). The selected image will
+    track the digest associated with that tag, which may change over time as new
+    images are pushed.
```
krancour:
```suggestion
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -157,9 +165,17 @@ strategies are:
...
+
+    :::note
+    Prior to Kargo v1.7, the `semverConstraint` field was used to
+    specify the mutable tag name. This field is now deprecated
+    and will be removed in v1.9.0. Users should use the `constraint` field
+    instead.
```
krancour:
```suggestion
    If the `constraint` field is empty, the name of the mutable tag will
    (unintuitively) be determined by the value of the deprecated
    `semverConstraint` field. The `semverConstraint` field will be removed in
    v1.9.0.
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -154,12 +160,16 @@ strategies are:
...
 
-- `Digest`: This selects the image _currently_ referenced by some "mutable tag,"
-   such as `latest`.
+- `Digest`: This selects the image _currently_ referenced by some "mutable tag"
+   (such as `latest`) specified by the `constraint` field.
 
-    __Unintuitively, the mutable tag name must be specified using the
-    `semverConstraint` field.__ Importantly, the _digest_ will change every time
-    the tag is updated.
+
```
krancour:
```suggestion
```

krancour (approved):

PR #4544, opened 2025-07-02
fix(runner)!: improve error handling `http` step

hiddeco:
This PR fixes a bug where `http` Promotion steps that failed due to failure criteria being met would never retry, even when configured with thresholds.

The issue was that these failures were incorrectly wrapped as `TerminalError`, causing the engine to treat them as unrecoverable errors rather than retryable failures. This meant that transient HTTP issues (like temporary server errors, network timeouts, or unexpected responses) would immediately fail instead of being retried according to the step's error threshold.

### DIFF
--- a/internal/promotion/runner/builtin/http_requester.go
+++ b/internal/promotion/runner/builtin/http_requester.go
@@ -95,18 +95,31 @@ func (h *httpRequester) run(
 		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
 			fmt.Errorf("error building expression context from HTTP response: %w", err)
 	}
-	success, err := h.wasRequestSuccessful(cfg, resp.StatusCode, env)
+
+	// Evaluate success and failure criteria
+	successResult, err := h.evaluateSuccessCriteria(cfg, env)
 	if err != nil {
 		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
 			fmt.Errorf("error evaluating success criteria: %w", err)
 	}
-	failure, err := h.didRequestFail(cfg, resp.StatusCode, env)
+
+	failureResult, err := h.evaluateFailureCriteria(cfg, env)
 	if err != nil {
 		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
 			fmt.Errorf("error evaluating failure criteria: %w", err)
 	}
+
+	// Determine outcome based on criteria evaluation results
 	switch {
-	case success && !failure:
+	case failureResult != nil && *failureResult:
+		// Failure criteria met: terminal failure
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusFailed},
+			&promotion.TerminalError{Err: fmt.Errorf(
+				"HTTP (%d) response met failure criteria",
+				resp.StatusCode,
+			)}
+	case successResult != nil && *successResult:
+		// Success criteria met: success
 		outputs, err := h.buildOutputs(cfg.Outputs, env)
 		if err != nil {
 			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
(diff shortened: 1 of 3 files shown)


internal/promotion/runner/builtin/http_requester.go
```diff
@@ -53,12 +53,13 @@ func (h *httpRequester) Run(
...
 	if err := h.validate(stepCtx.Config); err != nil {
-		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored}, err
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			&promotion.TerminalError{Err: err}
 	}
 	cfg, err := promotion.ConfigToStruct[builtin.HTTPConfig](stepCtx.Config)
 	if err != nil {
 		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
-			fmt.Errorf("could not convert config into http config: %w", err)
+			&promotion.TerminalError{Err: fmt.Errorf("could not convert config into http config: %w", err)}
```
hiddeco, line 62: This can be done in all other steps as well, and there is probably more low-hanging fruit to be found there.

internal/promotion/runner/builtin/http_requester.go
```diff
@@ -235,16 +233,22 @@ func (h *httpRequester) wasRequestSuccessful(
...
+			return false, &promotion.TerminalError{
+				Err: fmt.Errorf("error compiling success expression %q: %w", cfg.SuccessExpression, err),
+			}
 		}
 		successAny, err := expr.Run(program, env)
 		if err != nil {
-			return false, fmt.Errorf("error evaluating success expression: %w", err)
+			return false, &promotion.TerminalError{
+				Err: fmt.Errorf("error evaluating success expression %q: %w", cfg.SuccessExpression, err),
+			}
```
krancour: If the failure evaluating the expression was attributable to applying the expression to an unexpected response, it's possible that evaluation of the expression might succeed on retry. I think that possibility may also be applicable to the expression evaluating as a non-bool (below). Also applicable in the logic of the `didRequestFail()` method.

I agree it's terminal (above) if the expression cannot be compiled. Nothing will change that.

hiddeco: Fixed!

internal/promotion/runner/builtin/http_requester.go
```diff
@@ -235,16 +233,18 @@ func (h *httpRequester) wasRequestSuccessful(
 	case cfg.SuccessExpression != "":
 		program, err := expr.Compile(cfg.SuccessExpression)
 		if err != nil {
-			return false, fmt.Errorf("error compiling success expression: %w", err)
+			return false, &promotion.TerminalError{
+				Err: fmt.Errorf("error compiling success expression %q: %w", cfg.SuccessExpression, err),
+			}
```
krancour: I'm not positive I agree with this in light of some other changes in this same PR. Or I agree with it partially, but not entirely.

If there were any kind of terminal _error_ in evaluating success or failure, we've already returned that on (new) lines 99 or 104. That means that when we get to here, we're here because the response fulfilled specific failure criteria or _didn't_ fulfill specific success criteria. If those criteria are definitive, then this _is_ a terminal condition.

The exception (and hence why I don't entirely disagree with this change) is when no explicit success or failure criteria were provided, then success or failure was only a function of the response's status code and that may change on retry -- not terminal. I think it's _only_ this situation that the original code didn't deal correctly.

In this PR's current state, we'll end up retrying even after definitive failure criteria have been met. We're exchanging the problem of not retrying when we should for a problem of retrying when we shouldn't.

I feel like getting this right may require a more comprehensive refactor of how we judge success, failure, and retriability.

krancour: idk how, but I added this comment in the wrong place. This should be on line +121

krancour: DM'ed you with a truth table and some ideas.

krancour:
Looks great. Only question is about the milestone. I feel like this changes existing behavior just barely enough that this maybe doesn't belong in a patch release, but I could be convinced otherwise. Your call @hiddeco.

hiddeco:
@krancour I think I agree. Also given that the issue we were presented with is hypnotically solvable by setting an explicit `false` for the failure condition.

I will add an `!` to the PR title though, as this is technically breaking.

PR #4547, opened 2025-07-03
feat: validate no duplicate step aliases

krancour:
This PR validates that Stages, PromotionTasks, and ClusterPromotionTasks do not contain any duplicate step aliases.

I've validated this end-to-end in addition to the unit test coverage.

### DIFF
--- a/internal/webhook/kubernetes/promotiontask/webhook.go
+++ b/internal/webhook/kubernetes/promotiontask/webhook.go
@@ -2,6 +2,7 @@ package promotiontask
 
 import (
 	"context"
+	"errors"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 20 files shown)


internal/webhook/kubernetes/promotiontask/webhook.go
```diff
@@ -34,18 +34,50 @@ func SetupWebhookWithManager(
...
+	if err := libWebhook.ValidateProject(
+		ctx,
+		w.client,
+		promotionTaskGroupKind,
+		task,
+	); err != nil {
 		return nil, err
 	}
-	return nil, nil
+	return w.validateCreateOrUpdate(task)
```
hiddeco: Should these be joined together? That allows the user to see as many issues on first apply as possible. (The same applies to the other validators).

krancour: The thing is that the error returned from ValidateProject() is never a field.Error. So it's very hard to combine it with field.Errors from the rest of the validation. Fixing this would require refactoring ValidateProject(), which is probably long overdue, but a lot of things touch that, so the scope of this somewhat urgent change grows dramatically if we go down that road.

hiddeco (approved):

internal/webhook/kubernetes/promotiontask/webhook.go
krancour: Eh... I fell down the rabbit hole anyway.

See 9fc38c87556e7f0f0b110823b8f7c968aff3489b

I rewrote all the "top level" validation like `ValidateCreate()` and `ValidateUpdate()` to be much more thorough about surfacing as many problems as possible in one go.

Tests are now much improved as well.

PR #4552, opened 2025-07-05
Add CommitBatchSize to the GitSubscription

boh-dan:
## Context

Git commit discovery takes a significant amount of time because Kargo Controller pulls commits using `git log --skip=% --max-count=%`, which means it needs to spawn a new OS process to get a small group of commits (in our case, 3 commits). It might be "expensive" if the desired commits are located 1000 commits below the HEAD.

## Fix
Add CommitBatchSize to control the number of commits fetched in one batch.


hiddeco (commented):
I wonder if instead of making this a configurable option, we could do it internally by "dynamically doubling". I.e., every time we can not find sufficient matches based on the current batch size, we double it until we reach an upper limit of 1000.

Taking 1000 commits as an example:

- Current approach: 50 processes (50 × 20 commits each)
- Dynamic doubling: 6 processes (20→40→80→160→320→380 commits)
- 88% fewer processes

While with 5000 commits, the effect would be even better:

- Current approach: 250 processes
- Dynamic doubling: 10 processes
- 96% fewer processes

This would (significantly) improve performance while maintaining the current behavior for smaller repos, without introducing another option users would need to understand.

boh-dan:
@hiddeco nice idea, just pushed the fix

hiddeco:
@boh-dan thanks, this looks great. The only thing it doesn't take into account is the upper limit.

I would suggest something like the following:

```go
for skip, batchSize := uint(0), uint(limit); ; skip, batchSize = skip+batchSize, min(batchSize*2, 1000) { // nolint: gosec
    commits, err := r.listCommitsFn(repo, uint(batchSize), skip) // nolint: gosec
    
    // ...omitted for brevity 
}
```

boh-dan:
@hiddeco fixed

hiddeco (approved):
This looks good to me, thank you @boh-dan! 🙇

Issue #4490, opened 2025-06-24
warehouse refreshing optimizations


krancour:
#4000 and #4150 got us off to a good start with refreshing Warehouses in response to notifications of artifacts having being pushed to repositories delivered via webhook.

Now we can optimize...

Currently, we have Warehouses indexed by the URLs of the repositories to which they subscribe. After parsing a webhook payload and determining the repository of origin, we lookup Warehouses subscribed to that repository and we refresh them all.

In many cases, this will prove quite unnecessary, and may even prompt _many_ unwanted artifact discovery cycles -- which are expensive.

If, for example, a Warehouse subscribed to pushes _to the main branch_ of `https://github.com/example/repo.git`, but Kargo was notified of a push to _some other branch_ of that same repository, there's no reason to refresh that Warehouse. An alternate example would be a Warehouse subscribed to semantically versioned tags of some Docker Hub repository like `example/repo`. If `example/repo:my-custom-image` were pushed to that repository, there's no reason to refresh that Warehouse.

The existing "Warehouses by subscribed repos" index seems like it will narrow the field of candidate Warehouses sufficiently that we can iterate over the results simply checking whether any of the Warehouses subscriptions _really_ care about the artifact that's been pushed.

PR #4560, opened 2025-07-08
feat: refresh optimizations

fuskovic:
Closes: https://github.com/akuity/kargo/issues/4490

### DIFF
--- a/internal/webhook/external/refresh.go
+++ b/internal/webhook/external/refresh.go
@@ -27,6 +27,7 @@ func refreshWarehouses(
 	w http.ResponseWriter,
 	c client.Client,
 	project string,
+	rc refreshConstraint,
 	repoURLs ...string,
 ) {
 	logger := logging.LoggerFromContext(ctx)
@@ -63,6 +64,9 @@ func refreshWarehouses(
 	})
 	warehouses = slices.CompactFunc(warehouses, func(lhs, rhs kargoapi.Warehouse) bool {
 		return lhs.Namespace == rhs.Namespace && lhs.Name == rhs.Name
+	})
+	warehouses = slices.DeleteFunc(warehouses, func(w kargoapi.Warehouse) bool {
+		return !rc.needsRefresh(ctx, w.Spec.Subscriptions)
 	})
 
 	logger.Debug("found Warehouses to refresh", "count", len(warehouses))
--- a/internal/controller/warehouses/git.go
+++ b/internal/controller/warehouses/git.go
@@ -211,11 +211,11 @@ func (r *reconciler) discoverBranchHistory(repo git.Repo, sub kargoapi.GitSubscr
 	}
 
 	// Compile include and exclude path selectors.
-	includeSelectors, err := getPathSelectors(sub.IncludePaths)
+	includeSelectors, err := GetPathSelectors(sub.IncludePaths)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing include selector: %w", err)
 	}
-	excludeSelectors, err := getPathSelectors(sub.ExcludePaths)
+	excludeSelectors, err := GetPathSelectors(sub.ExcludePaths)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing exclude selector: %w", err)
 	}
@@ -257,7 +257,7 @@ func (r *reconciler) discoverBranchHistory(repo git.Repo, sub kargoapi.GitSubscr
 						err,
 					)
 				}
-				if !matchesPathsFilters(includeSelectors, excludeSelectors, diffPaths) {
+				if !MatchesPathsFilters(includeSelectors, excludeSelectors, diffPaths) {
 					continue
 				}
 			}
@@ -320,11 +320,11 @@ func (r *reconciler) discoverTags(repo git.Repo, sub kargoapi.GitSubscription) (
 	}
 
 	// Compile include and exclude path selectors.
-	includeSelectors, err := getPathSelectors(sub.IncludePaths)
+	includeSelectors, err := GetPathSelectors(sub.IncludePaths)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing include selector: %w", err)
 	}
-	excludeSelectors, err := getPathSelectors(sub.ExcludePaths)
+	excludeSelectors, err := GetPathSelectors(sub.ExcludePaths)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing exclude selector: %w", err)
 	}
@@ -341,7 +341,7 @@ func (r *reconciler) discoverTags(repo git.Repo, sub kargoapi.GitSubscription) (
 				err,
 			)
 		}
-		if matchesPathsFilters(includeSelectors, excludeSelectors, diffPaths) {
+		if MatchesPathsFilters(includeSelectors, excludeSelectors, diffPaths) {
 			filteredTags = append(filteredTags, meta)
 		}
 
@@ -361,7 +361,7 @@ func filterTags(tags []git.TagMetadata, ignoreTags []string, allow string) ([]gi
 	}
 	filteredTags := make([]git.TagMetadata, 0, len(tags))
 	for _, tag := range tags {
-		if ignores(tag.Tag, ignoreTags) || !allows(tag.Tag, allowRegex) {
+		if Ignores(tag.Tag, ignoreTags) || !Allows(tag.Tag, allowRegex) {
 			continue
 		}
 		filteredTags = append(filteredTags, tag)
@@ -466,18 +466,18 @@ func evaluateCommitExpression(
 	}
 }
 
-// allows returns true if the given tag name matches the given regular
+// Allows returns true if the given tag name matches the given regular
 // expression or if the regular expression is nil. It returns false otherwise.
-func allows(tagName string, allowRegex *regexp.Regexp) bool {
+func Allows(tagName string, allowRegex *regexp.Regexp) bool {
 	if allowRegex == nil {
 		return true
 	}
 	return allowRegex.MatchString(tagName)
 }
 
-// ignores returns true if the given tag name is in the given list of ignored
+// Ignores returns true if the given tag name is in the given list of ignored
 // tag names. It returns false otherwise.
-func ignores(tagName string, ignore []string) bool {
+func Ignores(tagName string, ignore []string) bool {
 	for _, i := range ignore {
 		if i == tagName {
 			return true
@@ -486,7 +486,7 @@ func ignores(tagName string, ignore []string) bool {
 	return false
 }
 
-func getPathSelectors(selectors []string) (pattern.Matcher, error) {
+func GetPathSelectors(selectors []string) (pattern.Matcher, error) {
 	if len(selectors) == 0 {
 		return nil, nil
 	}
@@ -502,7 +502,7 @@ func getPathSelectors(selectors []string) (pattern.Matcher, error) {
 	return matchers, nil
 }
 
-func matchesPathsFilters(include, exclude pattern.Matcher, diffs []string) bool {
+func MatchesPathsFilters(include, exclude pattern.Matcher, diffs []string) bool {
 	for _, path := range diffs {
 		// If include is nil, all paths are implicitly included
 		// Otherwise, check if the path matches the include pattern
--- a/internal/controller/warehouses/git_test.go
+++ b/internal/controller/warehouses/git_test.go
@@ -1004,7 +1004,7 @@ func TestAllows(t *testing.T) {
 			require.Equal(
 				t,
 				testCase.allowed,
-				allows(testCase.tag, testCase.regex),
+				Allows(testCase.tag, testCase.regex),
 			)
 		})
 	}
@@ -1035,7 +1035,7 @@ func TestIgnores(t *testing.T) {
 			require.Equal(
 				t,
 				testCase.ignored,
-				ignores(testCase.tag, testCase.ignore),
+				Ignores(testCase.tag, testCase.ignore),
 			)
 		})
 	}
@@ -1331,12 +1331,12 @@ func TestMatchesPathsFilters(t *testing.T) {
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
-			includeSelectors, err := getPathSelectors(testCase.includePaths)
+			includeSelectors, err := GetPathSelectors(testCase.includePaths)
 			require.NoError(t, err)
-			excludeSelectors, err := getPathSelectors(testCase.excludePaths)
+			excludeSelectors, err := GetPathSelectors(testCase.excludePaths)
 			require.NoError(t, err)
 
-			matchFound := matchesPathsFilters(includeSelectors, excludeSelectors, testCase.diffs)
+			matchFound := MatchesPathsFilters(includeSelectors, excludeSelectors, testCase.diffs)
 			testCase.assertions(t, matchFound)
 		})
 	}
--- a/internal/webhook/external/refresh_constraint.go
+++ b/internal/webhook/external/refresh_constraint.go
@@ -1,0 +1,319 @@
+package external
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"slices"
+	"strconv"
+
+	"github.com/Masterminds/semver/v3"
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/controller/git"
+	libSemver "github.com/akuity/kargo/internal/controller/semver"
+	"github.com/akuity/kargo/internal/controller/warehouses"
+	"github.com/akuity/kargo/internal/logging"
+	"github.com/expr-lang/expr"
+)
+
+type refreshConstraint struct {
+	Git *struct {
+		Tag        string
+		Branch     string
+		Diffs      []string
+		Expression string
+	}
+	Image *struct {
+		Tag string
+	}
+	Chart *struct {
+		Tag string
+	}
+}
+
+func (rc refreshConstraint) needsRefresh(
+	ctx context.Context,
+	subs []kargoapi.RepoSubscription,
+	repoURLs ...string,
+) bool {
+	return slices.ContainsFunc(
+		filterSubsByRepoURL(subs, repoURLs...),
+		func(sub kargoapi.RepoSubscription) bool {
+			return rc.matches(ctx, sub)
+		},
+	)
+}
+
+// filterSubsByRepoURL deletes all subscriptions from subs that do not
+// match any of the provided repository URLs.
+func filterSubsByRepoURL(
+	subs []kargoapi.RepoSubscription,
+	repoURLs ...string,
+) []kargoapi.RepoSubscription {
+	containsRepoURL := func(sub kargoapi.RepoSubscription) bool {
+		return sub.Image != nil && slices.Contains(repoURLs, sub.Image.RepoURL) ||
+			sub.Git != nil && slices.Contains(repoURLs, sub.Git.RepoURL) ||
+			sub.Chart != nil && slices.Contains(repoURLs, sub.Chart.RepoURL)
+	}
+	return slices.DeleteFunc(subs, func(sub kargoapi.RepoSubscription) bool {
+		return !containsRepoURL(sub)
+	})
+}
+
+func (rc refreshConstraint) matches(
+	ctx context.Context,
+	sub kargoapi.RepoSubscription,
+) bool {
+	return rc.matchesGitConstraint(ctx, sub.Git) ||
+		rc.matchesImageConstraint(sub.Image) ||
+		rc.matchesChartConstraint(sub.Chart)
+}
+
+func (rc refreshConstraint) matchesGitConstraint(
+	ctx context.Context,
+	sub *kargoapi.GitSubscription,
+) bool {
+	if rc.Git == nil || sub == nil {
+		return false
+	}
+	switch sub.CommitSelectionStrategy {
+	case kargoapi.CommitSelectionStrategyLexical:
+		return rc.matchesLexicalConstraint(ctx, rc.Git.Tag, sub)
+	case kargoapi.CommitSelectionStrategySemVer:
+		return rc.matchesSemVerConstraint(ctx, sub)
+	case kargoapi.CommitSelectionStrategyNewestTag:
+		return rc.matchesNewestTagConstraint(ctx, sub)
+	default:
+		return rc.matchesNewestFromBranchConstraint(ctx, sub)
+	}
+}
+
+func (rc refreshConstraint) matchesSemVerConstraint(
+	ctx context.Context,
+	sub *kargoapi.GitSubscription,
+) bool {
+	logger := logging.LoggerFromContext(ctx)
+
+	constraint, err := semver.NewConstraint(sub.SemverConstraint)
+	if err != nil {
+		logger.Debug("failed to parse semver constraint",
+			"tag", rc.Git.Tag,
+			"constraint", sub.SemverConstraint,
+			"error", err.Error(),
+		)
+		return false
+	}
+
+	version := libSemver.Parse(rc.Git.Tag, sub.StrictSemvers)
+	if version == nil {
+		logger.Debug("tag is not semver formatted",
+			"tag", rc.Git.Tag,
+			"constraint", sub.SemverConstraint,
+			"strictSemvers", sub.StrictSemvers,
+		)
+		return false
+	}
+
+	if !constraint.Check(version) {
+		logger.Debug("tag does not satisfy semver constraint",
+			"tag", rc.Git.Tag,
+			"constraint", sub.SemverConstraint,
+		)
+		return false
+	}
+	return matchesPathFilters(ctx, sub, rc.Git.Diffs) &&
+		matchesExpression(ctx,
+			sub.ExpressionFilter,
+			git.TagMetadata{
+				// TODO(Faris): Fill this in with actual tag metadata
+			},
+			rc.Git.Expression,
+		)
+}
+
+func (rc refreshConstraint) matchesLexicalConstraint(
+	ctx context.Context,
+	tag string,
+	sub *kargoapi.GitSubscription,
+) bool {
+	logger := logging.LoggerFromContext(ctx)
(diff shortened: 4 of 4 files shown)


internal/webhook/external/bitbucket.go
```diff
@@ -155,7 +149,88 @@ func (b *bitbucketWebhookReceiver) getHandler(requestBody []byte) http.HandlerFu
...
+
+// newBitBucketCodeChange creates a new codeChange instance from a Bitbucket push event.
+// It extracts the metadata from the event payload. This is used downstream to determine which Warehouses
+// should be refreshed in response to the event based on the commit selection
+// strategy configured for the Warehouse.
+//
+// See the Bitbucket Push event payload documentation for more details:
+//
+//	https://support.atlassian.com/bitbucket-cloud/docs/event-payloads/#Push
+func newBitBucketCodeChange(e *bitBucketPushEvent) *codeChange {
```
krancour: Rather than provide feedback on all the receivers individually, I'm going to dial in on the Bitbucket receiver for now (since it's first alphabetically) and we can see where it leads us and how that informs changes to the other receivers and/or the general approach...

internal/webhook/external/bitbucket.go
```diff
@@ -155,7 +149,88 @@ func (b *bitbucketWebhookReceiver) getHandler(requestBody []byte) http.HandlerFu
...
+	hc := e.Push.Changes[0].New
```
krancour: What does `hc` stand for? After looking at this code for quite some time, I'm still not sure. A more descriptive name would probably add quite a bit of clarity. All the other receivers have a variable with this name as well.

internal/webhook/external/bitbucket.go
```diff
@@ -155,7 +149,88 @@ func (b *bitbucketWebhookReceiver) getHandler(requestBody []byte) http.HandlerFu
...
+
+	// expected format "Name <email>".
+	author := fmt.Sprintf("%s <%s>",
+		e.Actor.Name,
+		e.Actor.EmailAddress,
+	)
```
krancour: According to the Bitbucket docs, actor isn't the same thing as author. (It's also not the same thing as committer.) The author is the person (or thing) that wrote the code. The committer is the person or thing that made a commit. (These aren't always the same, but in the simplest cases, they are. This isn't Bitbucket specific, btw. This is universal in Git.) Actor, according to the docs, is the person or thing that _pushed_ the commits (plural) to Bitbucket. You can make 100 commits to a feature branch and be the author (and committer) of every single one of them, open a PR (MR in Bitbucket, I think), and if _I_ then to merge that branch into another, _I'm_ the actor.

Even if these were equivalent, I don't believe we should be filtering on this information. Pretend for a moment that actor == author...

During discovery, a Warehouse will look backwards through _all_ commits (starting with a given ref) to find ones we're interested in, up to a configurable limit. Authorship information can, of course, be taken into account at that time using a user-defined expression as filter criteria. But here, you have in-hand information for just a _single_ author and you're using that in manufacturing some commit metadata a few lines down form here. You're later evaluating any user-defined expression against that _one_ piece of information to decide whether discovery should run...

I'll give you a scenario. 100 commits are pushed to a branch (a branch we care about). You pushed them, but I authored them. If the expression only selects commits authored by me, but you evaluated that expression _once_ against commit metadata that says _you_ were the author, the we're not going to refresh / run discovery, and we won't find those 100 commits of interest until the Warehouse runs a regularly scheduled discovery cycle on its own.

To _accurately_ decide if discovery should run, you need to evaluate that expression against all 100 of those commits. i.e. You'd basically need to do discovery in order to decide if we should do discovery. 😬 

Even supposing it made sense to do that... you don't actually have all 100 commits in-hand. The Bitbucket docs say that if a push event reflects more than five commits, the list of commits is truncated.

I believe the situation is going to be similar for every other (Git) receiver as well. You're seldom going to have enough information in-hand to accurately decide if discovery should run, and even if you did, you'd basically have to run discovery just to determin that.

All of the above extends to other aspects of commit metadata as well. For instance, you're capturing a single subject (commit message) and similar to the above, using that in manufacturing some commit metadata that you evaluate a user-defined expression against. If that expression is selecting commits whose subject that start with `feat:` and there are 99 new commits with subjects all starting with `feat:`, but the _one_ newest commit has a subject starting with `fix:` we're going to incorrectly _not_ run discovery.

In the end, I think what is called for here is some drastic simplification in how we determine if discovery should run based on the (limited) data we have in hand. I think this will exclusively involve looking at branch or tag names.

fuskovic: Per our offline conversation: I've completely omitted path + expression filter checks since none of the git providers provide all the information needed to properly evaluate any expressions for commits and tags ( and partially hydrated tag or commit metadata can result in false-negatives).

In these cases we will simply give the refresh the benefit of the doubt and let the downstream controller properly evaluate expression + path filters during discovery. I've pushed the refactor.

krancour:
@fuskovic fyi, I'm going to finish up what's left of this while you focus on receivers.

fuskovic:
Per offline conversation w/ @krancour, reassigning this back to myself since refactor of shared filtering logic was completed in https://github.com/akuity/kargo/pull/4671

internal/webhook/external/refresh.go
```diff
@@ -68,6 +69,11 @@ func refreshWarehouses(
 	warehouses = slices.CompactFunc(warehouses, func(lhs, rhs kargoapi.Warehouse) bool {
 		return lhs.Namespace == rhs.Namespace && lhs.Name == rhs.Name
 	})
```
krancour: I think after #4671, this "eligibility checker" concept can be replaced with a simple function that filters the Warehouses we've already found...

Loop through a Warehouse's subscriptions looking for one that matches any of the repoURLs you have in hand. When you find one, exchange that sub for a image/commit/chart selector and call it's "MatchesX" (tag, ref, etc.) method to see if it matches some "qualifying" bit of information you also have in hand.

What exactly this "qualifying bit of information" is is context-dependent. If the event was a push to an image repo, it's a tag. If it was a push to a git repo, it's a ref -- i.e. `refs/heads/<branch>` for a push to a branch or `refs/heads/<tags>` for a push of a new tag. If it was a push to an OCI chart repo, it's a tag (which will always be a semantic version since Helm requires that).

The signature for the `refreshWarehouses()` function probably changes to something like:

```
func refreshWarehouses(
	ctx context.Context,
	w http.ResponseWriter,
	c client.Client,
	project string,
  qualifier string,
	repoURLs ...string,
)
```

internal/webhook/external/refresh.go
```diff
@@ -69,6 +74,23 @@ func refreshWarehouses(
...
 
+	if qualifier != "" {
+		var refreshEligibleWarehouses []kargoapi.Warehouse
+		for _, wh := range warehouses {
+			shouldRefresh, err := needsRefresh(ctx, wh.Spec.Subscriptions, qualifier, repoURLs...)
+			if err != nil {
+				// log the error but obscure the details from the response
+				logger.Error(err, "failed to evaluate if warehouse needs refresh", "warehouse", wh.Name)
+				xhttp.WriteErrorJSON(w, fmt.Errorf("failed to evaluate if refresh is needed for warehouse %q", wh.Name))
```
krancour: This message won't ever be seen, so we probably don't need to bother. If we don't use `xhttp.Error()` to wrap `err` with the addition of a specific status code, `xhttp.WriteErrorJSON()` is treating it as a 500 and not divulging anything about what went wrong. (Which is the right thing to do.)

```suggestion
				xhttp.WriteErrorJSON(w, err)
```

internal/webhook/external/refresh.go
```diff
@@ -69,6 +74,23 @@ func refreshWarehouses(
 		return lhs.Namespace == rhs.Namespace && lhs.Name == rhs.Name
 	})
 
+	if qualifier != "" {
+		var refreshEligibleWarehouses []kargoapi.Warehouse
```
krancour: We can set the capacity to be equal to the total number of Warehouses in-hand so that we don't need to resize with every append.

```suggestion
		refreshEligibleWarehouses := make([]kargoapi.Warehouse, 0, len(warehouses))
```

internal/webhook/external/refresh.go
```diff
@@ -103,3 +125,61 @@ func refreshWarehouses(
 		},
 	)
 }
+
+func needsRefresh(
+	_ context.Context,
```
krancour: If we're not using this, it probably doesn't need to be part of this func's signature.

```suggestion
	_ context.Context,
```

internal/webhook/external/refresh.go
```diff
@@ -103,3 +125,61 @@ func refreshWarehouses(
...
+	for _, s := range subs {
+		switch {
+		case s.Git != nil:
+			selector, err := commit.NewSelector(*s.Git, nil)
+			if err != nil {
+				return nil, fmt.Errorf("error creating commit selector for Git subscription %q: %w",
+					s.Git.RepoURL, err,
+				)
+			}
+			shouldRefresh = selector.MatchesRef(qualifier)
```
krancour: If we do the early return in each case instead of after the switch, it's a bit easier to follow. Applies to the other cases as well.

```suggestion
			return selector.MatchesRef(qualifier), nil
```

internal/webhook/external/refresh.go
```diff
@@ -103,3 +125,61 @@ func refreshWarehouses(
...
+				)
+			}
+			shouldRefresh = selector.MatchesVersion(qualifier)
+		}
+		if shouldRefresh {
+			// exit early if we already found a match
+			return &shouldRefresh, nil
+		}
+	}
+	return &shouldRefresh, nil
```
krancour: Related to https://github.com/akuity/kargo/pull/4560/files#r2229848911

```suggestion
	}
	return false, nil
```

internal/webhook/external/refresh.go
```diff
@@ -103,3 +125,61 @@ func refreshWarehouses(
...
+}
+
+// filterSubsByRepoURL deletes all subscriptions from subs that do not
+// match any of the provided repository URLs; omitting them from processing.
+func filterSubsByRepoURL(subs []kargoapi.RepoSubscription, repoURLs ...string) []kargoapi.RepoSubscription {
+	containsRepoURL := func(sub kargoapi.RepoSubscription) bool {
+		return sub.Image != nil && slices.Contains(repoURLs, sub.Image.RepoURL) ||
+			sub.Git != nil && slices.Contains(repoURLs, sub.Git.RepoURL) ||
```
krancour: You forgot to normalize `sub.Image.RepoURL` and `sub.Git.RepoURL`. (You remembered for `sub.Chart.RepoURL`.)

btw, we know, contextually, the way this is used, `repoURLs` will already all be normalized. We should probably document that expectation in the comment for this func.

krancour: Also some extra parens here would make the logic easier to follow without needing to do any head scratching about the precedence of `&&` vs `||`. i.e.:

```
(sub.Image != nil && slices.Contains(...)) ||
(sub.Git != nil && slices.Contains(...)) ||
(sub.Chart != nil && slices.Contains(...))
```

internal/webhook/external/refresh.go
```diff
@@ -103,3 +125,61 @@ func refreshWarehouses(
 		},
 	)
 }
+
+func needsRefresh(
```
krancour: For some reason `shoudlRefresh()` feels better here.

krancour:
This seems to be on the right track. I've looked at the general refresh logic and not yet looked at new info we're picking out of the payloads. I will follow up on that tomorrow.

internal/webhook/external/refresh.go
fuskovic: The problem with the case-scoped early return is that if the selector returns false on the match, we skip evaluating other subs. This is incorrect because all subs need to be evaluated before we can return false. This is because if even a single sub for the warehouse matches then we need to return true. This is why the early return after the switch happens only if it's true.

internal/webhook/external/refresh.go
fuskovic: It's gotta be a bool ptr so can't return `false` here and I don't think we should nix the post-switch early return for reasons stated in https://github.com/akuity/kargo/pull/4560#discussion_r2230536902

internal/webhook/external/refresh.go
krancour: 🤦‍♂️ Sorry... you are totally correct! The sentiment may still apply even if the specific line of code was wrong. This might still be more readable overall:

```suggestion
			if selector.MatchesRef(qualifier) {
						return true, nil
			}
```

The thing is this stops the reader from having to skip ahead in the code to see what when `shouldRefresh` is `true`.

However, this isn't _so_ much more readable that it's any kind of hill I'd die on.

Up to you whether you want to change this or not and I'm fine either way.

internal/webhook/external/refresh.go
krancour: The thread in https://github.com/akuity/kargo/pull/4560#discussion_r2230536902 aside...

Why is this function returning a bool pointer to begin with? It's a bit unusual. Why not a bool?

internal/webhook/external/refresh.go
krancour: Actually... before we go too far down this rabbit hole, there might be a broader thing to address. Hold off and I'll DM you.

internal/webhook/external/refresh.go
fuskovic: I wanted to distinguish between "we do not need to refresh" and "we could not determine if we need to refresh"

internal/webhook/external/refresh.go
fuskovic: My rationale was to try and keep things DRY. e.g. instead of three code blocks( 1 for each case ) of:

```go
if matches := selector.Matches(); matches {
    return &matches, nil
 }
```

I just wanted to write it one time and have the same effect. Also not a hill I want to die on just wanted to explain my reasoning.

krancour: Ya... that's exactly why I wouldn't die on my hill either.

internal/webhook/external/github.go
```diff
@@ -184,16 +189,17 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+			qualifier = e.GetRef()
 			// TODO(krancour): GetHTMLURL() gives us a repo URL starting with
 			// https://. By refreshing Warehouses using a normalized representation of
 			// that URL, we will miss any Warehouses that are subscribed to the same
 			// repository using a different URL format.
 			repoURL = git.NormalizeURL(e.GetRepo().GetCloneURL())
 		}
 
-		logger = logger.WithValues("repoURL", repoURL)
+		logger = logger.WithValues("repoURL", repoURL, "qualifier", qualifier)
```
krancour: I think I'd find the logs a little bit more helpful if we could be more specific than "qualifier" here. Can we use "ref" for Git pushes and "tag" for package pushes? The easiest way to do so would probably be to add one or the other directly in their respective cases in the switch above here.

krancour: Applies to other receivers as well, although most of them are more straightforward on account of not dealing with multiple types of events.

internal/webhook/external/github.go
```diff
@@ -166,13 +167,17 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+			// Determine if the package is a Helm chart
 			if cfg, ok := manifest["config"].(map[string]any); ok {
 				if mediaType, ok := cfg["media_type"].(string); ok {
 					repoURL = normalizeOCIRepoURL(ref.Context().Name(), mediaType)
 				}
+			} else {
+				repoURL = image.NormalizeURL(ref.Context().Name())
 			}
-
+			qualifier = v.GetVersion()
```
krancour: Looking at actual payloads, it seems this may be incorrect.

This seems to be the right way:

```suggestion
			qualifier = v.GetContainerMetadata().GetTag().GetName()
```

internal/webhook/external/gitlab.go
```diff
@@ -123,7 +123,7 @@ func (g *gitlabWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 			}
 			logger = logger.WithValues("repoURL", repoURL)
```
krancour:
```suggestion
			logger = logger.WithValues(
				"repoURL", repoURL,
				"ref", e.Ref,
			)
```

internal/webhook/external/quay.go
```diff
@@ -101,10 +101,12 @@ func (q *quayWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc {
 			image.NormalizeURL(payload.DockerURL),
 			helm.NormalizeChartRepositoryURL(payload.DockerURL),
 		}
-
-		logger = logger.WithValues("repoURLs", repoURLs)
+		// Quay sends a []UpdatedTags in the payload, following convention,
+		// this means we should omit setting a qualifier for Quay given
+		// that some elements may match some selectors and others may not.
```
krancour: I researched this a bit and tinkered with it a bit as well. There's very little documentation about this, but it seems Quay may possibly bundle multiple push events into a single webhook request, but I've tried lots of concurrent pushes to a single repo and have been unable to trigger that behavior.

I see no reason not to expect that in _at least_ the common case, there is only one item in `UpdatedTags`.

What I'd like to suggest is that when there is indeed only one item in `UpdatedTags`, we should go ahead and use that one tag as the qualifier and fall back on using no qualifier if there is more than one -- which I'm willing to bet happens very rarely, if ever.

internal/webhook/external/bitbucket.go
```diff
@@ -156,10 +156,13 @@ func (b *bitbucketWebhookReceiver) getHandler(requestBody []byte) http.HandlerFu
 		// TODO(krancour): There are very likely some yet-to-be-identified edge
 		// cases where this choice does not hold up.
 		repoURL := payload.Repository.Links.HTML.Href
-
-		logger = logger.WithValues("repoURL", repoURL)
+		// BitBucket sends a []Changes in the payload, following convention,
+		// this means we should omit setting a qualifier for BitBucket given
+		// that elements could differ here e.g. one RefUpdate element could be a tag
+		// and another could be a branch
```
krancour: Dug into this... I can easily cause changes to have multiple entries that represent pushes to different refs. For instance, I can push to multiple branches with a single `git push command`.

This is making me start to believe that we need to support multiple qualifiers.

I would rather spend the extra time checking additional qualifiers than spending time on a possibly unnecessary discovery cycle. The latter is more expensive than the former.

That changes my guidance on [quay](https://github.com/akuity/kargo/pull/4560/files#diff-c8b105c7915915b0dcd8005dc4bca18d344c5a43ff1f556c3a3c30db26ed2bdb)

internal/webhook/external/refresh.go
fuskovic: Which seemed like a good idea at the time, but probably unnecessary since it's implied when err is not nil.

internal/webhook/external/artifactory.go
```diff
@@ -181,9 +182,12 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
 			)
 			return
 		}
-		logger = logger.WithValues("repoURL", repoURL)
+		logger = logger.WithValues(
+			"repoURL", repoURL,
+			"tag", payload.Data.Tag,
+		)
```
krancour: Looks like this was already accounted for on line +136

```suggestion
```

internal/webhook/external/azure.go
```diff
@@ -215,8 +223,19 @@ type acrEvent struct {
...
+		RefUpdates []struct {
+			Name string `json:"name,omitempty"`
+		} `json:"refUpdates,omitempty"`
 		Repository struct {
 			RemoteURL string `json:"remoteUrl,omitempty"`
 		} `json:"repository"`
 	} `json:"resource"`
 }
+
+func (event azureDevOpsEvent) getQualifiers() []string {
```
krancour: I would name this `getRefs()`. I think it adds a lot of clarity if we are specific about what our qualifiers actually are up until the moment we hand them off to the more generic refresh code. Same goes for the method of the same name for the Bitbucket webhook receiver.

krancour: A comment on the method explaining that it extracts all references mentioned by the event might be helpful.

internal/webhook/external/bitbucket.go
```diff
@@ -122,15 +122,7 @@ func (b *bitbucketWebhookReceiver) getHandler(requestBody []byte) http.HandlerFu
...
-		payload := struct {
-			Repository struct {
-				Links struct {
-					HTML struct {
-						Href string `json:"href"`
-					} `json:"html"`
-				} `json:"links"`
-			} `json:"repository"`
-		}{}
+		payload := new(bitBucketPushEvent)
```
krancour: Can't see a good reason for this to be a pointer, but apart from that, this probably causes a bit of an issue on +126 where `&payload` would now be a `**bitBucketPushEvent` instead of a `*bitBucketPushEvent`.

```suggestion
		payload := bitBucketPushEvent{}
```

krancour:
@fuskovic just a few very tiny things left and then this LGTM.

internal/webhook/external/artifactory.go
fuskovic: Looks like `tag` is but `repoURL` isn't (  just the components that make it up ) so I think I'd like to continue appending `repoURL` here.

fuskovic: Rationale being:

1. follows existing convention
2. could be a useful means of debugging for a possible future url parsing/concatenation edge case we may not have accounted for

krancour: You're right. Tag was and url wasn't. Your fix is more correct than my suggestion. 👍

internal/webhook/external/refresh.go
```diff
@@ -26,6 +30,7 @@ func refreshWarehouses(
 	w http.ResponseWriter,
 	c client.Client,
 	project string,
+	qualifiers []string,
 	repoURLs ...string,
```
krancour: I know that the way we landed on the current signature for this function was incremental. We used to accept just one qualifier, but multiple repoURLs and that's why repoURLs originally came last -- because you can only have one variadic argument and it must be last...

But now that we accept multiple qualifiers in addition to the multiple urls, I feel like it would be more intuitive for repoURLs to come before qualifiers.

Could we swap the order here?

```suggestion
	repoURLs []string,
	qualifiers ...string,
```

This is non-blocking, but I do think it's a good idea.

internal/webhook/external/github.go
```diff
@@ -167,13 +169,19 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+			v := pkg.GetPackageVersion()
+			manifest := v.GetContainerMetadata().GetManifest()
+			// Determine if the package is a Helm chart
 			if cfg, ok := manifest["config"].(map[string]any); ok {
 				if mediaType, ok = cfg["media_type"].(string); ok {
 					repoURLs = getNormalizedImageRepoURLs(ref.Context().Name(), mediaType)
 				}
+			} else {
+				repoURLs = append(repoURLs, image.NormalizeURL(ref.Context().Name()))
 			}
```
krancour: Media type info shouldn't really ever be missing afaik, but if it somehow were, a weakness with what's already in main (my fault) is repoURLs will be empty... so I appreciate that you're trying to improve this.

I think, however, it's also wrong that if we somehow don't know the media type, that we'd assume we're dealing with an image.

I think this might be more correct than either what I had here before or what you have here now:

```suggestion
			var mediaType string
			if cfg, ok := manifest["config"].(map[string]any); ok {
				mediaType, _ = cfg["media_type"].(string)
			}
			repoURLs = getNormalizedImageRepoURLs(ref.Context().Name(), mediaType)
```

The above makes a best effort to determine media type and it's resilient to:

* `"config"` being missing the `mainfest` map.
* The value of `"config"` not being a `map[string]any`.
* `"media_type"` being missing from the `cfg` map.
* The value o`"media_type"` not being a string.

In all of the above cases, `mediaType` will be the empty string.

If the media type is not recognized by `getNormalizedImageRepoURLs()` -- and this includes a media type of `""` -- then we'll get back the URL normalized both ways, which is what we really want.

internal/webhook/external/bitbucket.go
```diff
@@ -156,10 +148,39 @@ func (b *bitbucketWebhookReceiver) getHandler(requestBody []byte) http.HandlerFu
 		// TODO(krancour): There are very likely some yet-to-be-identified edge
 		// cases where this choice does not hold up.
 		repoURL := payload.Repository.Links.HTML.Href
-
-		logger = logger.WithValues("repoURL", repoURL)
+		refs := payload.getRefs()
+		logger = logger.WithValues("repoURL", repoURL, "refs", refs)
```
krancour: Nit: `.WithValues()` is awkward the way it uses a variadic argument to specify key/val pairs. If we treat calls to `.WithValues()` more like we're defining a map, the pairs, as well as any mistakes (like a key missing a value) become visually more obvious.

Non-blocking.

```suggestion
		logger = logger.WithValues(
			"repoURL", repoURL,
			"refs", refs,
		)
```

internal/webhook/external/refresh.go
```diff
@@ -99,7 +123,43 @@ func refreshWarehouses(
...
+		case s.Image != nil && image.NormalizeURL(s.Image.RepoURL) == repoURL:
+			selector, err := image.NewSelector(*s.Image, nil)
+			if err != nil {
+				return false, fmt.Errorf("error creating image selector for Image subscription %q: %w",
+					s.Image.RepoURL, err,
+				)
+			}
+			shouldRefresh = slices.ContainsFunc(qualifiers, selector.MatchesTag)
+		case s.Chart != nil && helm.NormalizeChartRepositoryURL(s.Chart.RepoURL) == repoURL:
+			selector, err := chart.NewSelector(*s.Chart, nil)
```
krancour, line 152: @fuskovic when we talked offline... is this the spot where you were mentioning the selector couldn't be initialized if the repo URL didn't have a leading `oci://`?

krancour:
@fuskovic I commented inline with some very minor recommendations, all of which you may ignore if you choose... except for this, which is the only blocker (and it's 50% my fault, fwiw):

https://github.com/akuity/kargo/pull/4560/files#r2243755366

internal/webhook/external/refresh.go
fuskovic: yeah exactly

Issue #4538, opened 2025-06-30
Helm Chart - Accept extraObjects for arbitrary k8s resources creation


patpicos:
# Checklist

* [X] I've searched the issue queue to verify this is not a duplicate feature request.
* [X] I've pasted the output of `kargo version`, if applicable.
* [X] I've pasted logs, if applicable.

# Proposed Feature

Add a new attribute to the helm chart to allow creation of arbitrary kubernetes resources during the deployment. This is useful when needing to deploy additional manifest through gitops tools like ArgoCD and avoid having an Application for a few resources then one Application driven from Helm Chart
Reference implementation: https://github.com/cloudability/metrics-agent/pull/267
# Motivation

During installation, I see two use cases already:
- ConfigMap creation for the caBundle to be used by the controller
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: gitlab-ca
  namespace: kargo
data:
  gitlab-ca.crt: |
<omitted>
```
- Deployment of RoleBinding when doing global credentials. However, I would prefer an opinionated approach to this feature on the helm chart, but `extraObjects` would do the trick. For this use case, I would need to create the Namespace, RoleBinding, and perhaps a series of ExternalSecrets resources to source my secrets from AWS Secret Manager
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
    name: kargo-controller-read-secrets
    namespace: kargo-global-creds
subjects:
    - kind: ServiceAccount
      name: kargo-controller
      namespace: kargo
roleRef:
    kind: ClusterRole
    name: kargo-controller-read-secrets
    apiGroup: rbac.authorization.k8s.io
```

krancour:
Given that what you're requesting isn't actually a feature Helm provides, I'm not sure this feels like the right place for this.

I've had this exact need many times myself and my current favorite approach is to leverage Argo CD multi-source applications.

I do this in my own home lab where I install charts using one source, and both the `values.yaml` with the specific configuration for the chart _and_ the additional resources required to make everything work being another source. In my case, most of those other things are Cert or SealedSecret resources.

Would that work for you?

Here's an example of how I do it:

https://github.com/krancovia/lab-gitops/blob/main/apps/traefik.yaml

https://github.com/krancovia/lab-gitops/blob/main/traefik/manifests/cert.yaml

patpicos:
[...]

PR #4565, opened 2025-07-08
feat: add support for extra Kubernetes manifests in values.yaml

patpicos:
Solves https://github.com/akuity/kargo/issues/4538


charts/kargo/values.yaml
```diff
@@ -97,7 +97,8 @@ webhooks:
 ## the Kargo controller is running.
 ## @descriptionEnd
 ## @skip kubeconfigSecrets
-kubeconfigSecrets: {}
+kubeconfigSecrets:
+  {}
```
krancour: I appreciate the removal of errant spaces on what should have been blank lines (as on line 51, for instance), because that should be completely non-controversial, but this particular change and others like it seem like they're unnecessarily bundling unrelated enforcement of an opinion into the PR. It's easier to approve this PR without such changes.

```suggestion
kubeconfigSecrets: {}
```

charts/kargo/templates/extra-manifests.yaml
```diff
@@ -0,0 +1,4 @@
+{{ range .Values.extraObjects }}
+---
+{{ tpl (toYaml .) $ }}
+{{ end }}
```
krancour: I like how the Argo CD chart provides the option for `extraObjects` entries to be strings. Depending on the complexity of any templates you are embedding, that is essentially a necessity.

https://github.com/argoproj/argo-helm/blob/bd8552e70850021b2591d1fc037600813e325498/charts/argo-cd/templates/extra-manifests.yaml#L1-L8

charts/kargo/values.yaml
```diff
@@ -872,3 +884,18 @@ webhooksServer:
     #   -----BEGIN CERTIFICATE-----
     #   ...
     #   -----END CERTIFICATE-----
+
+# Extra K8s manifests to deploy
```
```suggestion
## @param extraObjects is an array describing additional, arbitrary Kubernetes resources to include when rendering this chart. Items in the array may be YAML objects or strings. Either may be templated. Templates will be evaluated against the same set of values as the rest of the chart.
```

charts/kargo/templates/extra-manifests.yaml
patpicos: adjusted to argocd upstream approach.

charts/kargo/values.yaml
patpicos: blame "Format on Save" :)

patpicos:
@krancour ready for re-review

charts/kargo/values.yaml
```diff
@@ -872,3 +883,18 @@ webhooksServer:
...
+#   metadata:
+#     name: aws-store-cloudability
+#     annotations:
+#       argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+#       argocd.argoproj.io/sync-wave: "99"
+#   spec:
+#     provider:
+#       aws:
+#         service: SecretsManager
+#         region: us-west-2
```
krancour: imho, using a third-party resource type here is distracting. Using something ubiquitous like a ConfigMap would probably be a more accessible example. I think there's also an opportunity here to show how custom templates can be used.

```suggestion
# - apiVersion: v1
#   kind: ConfigMap
#   metadata:
#     name: custom-cm-1
#   data:
#     host: '{{ .Values.api.host }}'
# - |
#   apiVersion: v1
#   kind: ConfigMap
#   metadata:
#     name: custom-cm-2
#   data:
#     host: {{ .Values.api.host }}
```

charts/kargo/values.yaml
```diff
@@ -872,3 +872,20 @@ webhooksServer:
     #   -----BEGIN CERTIFICATE-----
     #   ...
     #   -----END CERTIFICATE-----
+
+## @param extraObjects is an array describing additional, arbitrary Kubernetes resources to include when rendering this chart. Items in the array may be YAML objects or strings. Either may be templated. Templates will be evaluated against the same set of values as the rest of the chart.
+extraObjects: []
+# - apiVersion: external-secrets.io/v1beta1
```
```suggestion
```

krancour (approved):

charts/kargo/values.yaml
patpicos: im good either way. I had an ES resource because it was a typical use case for me and using the sync-waves in argo.

PR #4578, opened 2025-07-09
fix(git): parsing of annotated Git tags

gnadaban:
This fixes an issue that prevents refreshing Git repository warehouses that have tags with annotations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ CERT_MANAGER_CHART_VERSION 	:= 1.16.1
 BUF_LINT_ERROR_FORMAT	?= text
 GO_LINT_ERROR_FORMAT 	?= colored-line-number
 
+GO_TEST_ARGS ?=
+
 VERSION_PACKAGE := github.com/akuity/kargo/pkg/x/version
 
 # Default to docker, but support alternative container runtimes that are CLI-compatible with Docker
@@ -138,7 +140,7 @@ test-unit: install-helm
 				-race \
 				-coverprofile=coverage.txt \
 				-covermode=atomic \
-				./... $${GO_TEST_ARGS}; \
+				./... $(GO_TEST_ARGS); \
 			cd - > /dev/null; \
 		done; \
 	}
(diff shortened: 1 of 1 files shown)


internal/controller/git/work_tree.go
```diff
@@ -467,6 +467,41 @@ type TagMetadata struct {
 	Annotation string
 }
 
+func ParseTagMetadataLine(line []byte) (TagMetadata, error) {
```
hiddeco: Think this function can be made private, as there is no usage of it outside the package itself.

internal/controller/git/work_tree.go
```diff
@@ -467,6 +467,41 @@ type TagMetadata struct {
...
+			CommitID:    string(parts[1]),
+			Subject:     string(parts[2]),
+			Author:      string(parts[3]),
+			Committer:   string(parts[4]),
+			CreatorDate: creatorDate,
+			Tagger:      string(parts[6]),
+			Annotation:  string(bytes.Join(parts[7:], []byte("|*|"))),
+		}, nil
+	}
+	return TagMetadata{}, fmt.Errorf("unexpected number of fields: %q", line)
```
hiddeco: This can be reduced to something like the following:

```go
if l := len(parts); l < 6 || l == 7 {
	return TagMetadata{}, fmt.Errorf("unexpected number of fields: %q", line)
}

creatorDate, err := time.Parse("2006-01-02 15:04:05 -0700", string(parts[5]))
if err != nil {
	return TagMetadata{}, fmt.Errorf("error parsing creator date %q: %w", parts[5], err)
}

metadata := TagMetadata{
	Tag:         string(parts[0]),
	CommitID:    string(parts[1]),
	Subject:     string(parts[2]),
	Author:      string(parts[3]),
	Committer:   string(parts[4]),
	CreatorDate: creatorDate,
}

if len(parts) >= 8 {
	metadata.Tagger = string(parts[6])
	metadata.Annotation = string(bytes.Join(parts[7:], []byte("|*|")))
}

return metadata, nil
```

hiddeco (approved):

Issue #4566, opened 2025-07-08
ui: live manifest of promo can be stale


krancour:
If I sit on the promotion view and wait for a promotion to finish and then click the YAML tab, the YAML appears to be stale, as it does not reflect up-to-date step execution metadata.

A few times now, this has fooled me into thinking some bookkeeping was wrong, only to discover that if I left the promotion view and came back, the YAML looks as I expect it to. Correct status immediately following a promotion is also verifiable via CLI.

I think we may need to refresh promotion YAML just-in-time.

https://github.com/user-attachments/assets/863da01e-2a89-490b-b1d1-33074c47ed42

PR #4592, opened 2025-07-10
fix(ui): stale promotion yaml in live promotion view

Marvin9:
fixes https://github.com/akuity/kargo/issues/4566


krancour:
@Marvin9 thank you for fixing this so quickly!

Issue #2996, opened 2024-11-25
custom/third-party promotion steps


krancour:
This has been on the roadmap for quite some time, but there hadn't been an issue for it.

Maintainers have been discussing this internally for quite some time.

Due to the extensive security implications of this, there is a fair amount of prerequisite work to be completed. In particular, we need a viable strategy for isolating the execution of third-party functionality to a safe/limited context. This likely involves isolating promotion processes or even individual steps to their own containers or even their own pods, although other options are being weighed as well.

jeffmccune:
Could you lay out the security implications at a high level?  Does the configuration of the promotion process originate from untrusted sources?

danielloader:
Implementing something like Envoys [ext_authz filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter) where you have a HTTP or gRPC contract to fulfil would be ideal for me for a break glass option. Ideally provide a reference implementation in a couple languages for a simple server that can implement the contract.

Similar to how Crossplane implements [composition functions](https://docs.crossplane.io/latest/guides/write-a-composition-function-in-go/) too. 

Definitely pros and cons doing this vs other ways of jamming executables/etc into a container.

In _theory_ these functions shouldn't be too latency sensitive so you shouldn't need the complexity of a sidecar type deployment, just a regular old webhook to wherever you're hosting the function call processor (AWS Lambda Function URL, pod/deployment in a k8s cluster, plain old binary hosted on a VM etc).

krancour:
@danielloader Argo CD also does something similar to that for CMPs.

This will not work in Kargo's case.

User defined promotion processes depend on a workspace in the file system where all of the repo cloning, file manipulation, etc. performed by your promotion steps is carried out. A simple API call to a third party would lack access to that workspace and would therefore be virtually useless.

We need each Promotion's workspace to be accessible to all built-in and third-party steps of a promotion process and that constrains the design choices quite a bit.

@hiddeco and @jessesuen and I have been whiteboarding options that satisfy this constraint for a while. We'll be getting more serious about weighing those and selecting one in 1Q25.

krancour:
This is no longer something we're currently entertaining.

michaelasper:
@krancour Any chance we could get details on why this is no longer being considered? As far as extensibility, are any first party promotion steps gonna be added to help? A docker run step would be pretty nice

jessesuen:
[...]

PR #4603, opened 2025-07-11
WIP: Proposal: "Custom" step runners via grpc

djosephsen:
Intent of this PR is to begin a conversation about implementing "custom" steps via GRPC.  This feature is roughly analogous to [crossplane composite functions](https://docs.crossplane.io/latest/concepts/compositions/#write-a-composition-function)

The general idea is that end-users instantiate custom code, wrapped in a grpc server, running in a pod adjacent to Kargo.  Then:
1. The user configures Kargo to use a "custom step"
```
steps = [{
          uses = "custom-grpc"
           config = {
           serverURL = "davesCustomCode.svc.cluster.local"
            }}]
```
2. Kargo runs the step, which serializes the stepContext into protobuf and ships it to the `davesCustomCode` pod for processing
3. the remote pod receives the stepContext, makes whatever modifications it makes, and returns it along with a stepResult 
4. the runner packs the returned freight into the stepContext it was given, and passes through the result. 

This arrangement would allow end-users to implement whatever transformation they needed, in whatever programming language they're familiar with (which supports grpc).

This pr contains an example of the grpc steprunner along with the proto necessary for serializing, I'm happy to iterate, and act on whatever feedback the project has.  If this feature were merged, my team could use it to collapse multiple mutation webhooks and have a much more supportable story.  

Re: the client side, also happy to work up a client-side sdk, or at least some codegen tools to implement these runners in golang and python.  I have a nascent client-server project I used to test the proto exchange if anyone is interested in that as well.


krancour:
While we expect we're likely to pursue something like this in the future, we've indicated in other issues / PRs that have broached this same issue, most notably [here](https://github.com/akuity/kargo/issues/2996#issuecomment-2797085675), that we do not intend to pursue this at this time. It's a business decision as much as it is a technical one, so despite the generous contribution (thank you), this isn't something we're going to sink any time or effort into reviewing.

lukasmrtvy:
@krancour, in other words, promotion steps which are behind paywall like https://docs.kargo.io/user-guide/reference-docs/promotion-steps/jira, available only on the Akuity Platform, would be easily replaceable with this PR, right?

PR #4621, opened 2025-07-14
refactor: minor webhook improvements

krancour:
Minor improvements I discovered could be made to a few existing webhook receivers while reviewing #4487:

* Responds to events we cannot handle with 400 bad request instead of 501 not implemented. This feels more appropriate. The issue isn't that we didn't implement something -- or more precisely it isn't that we didn't implement something we might reasonably have been expected to. The issue is simply that we were sent something we should not have been sent.

* Does a more thorough job of evaluating media types when responding to a push to an OCI repo, and makes fewer assumptions when doing so.

* Uses constants for headers and event types.

### DIFF
--- a/internal/webhook/external/gitlab.go
+++ b/internal/webhook/external/gitlab.go
@@ -15,14 +15,6 @@ import (
 )
 
 const (
-	azureSecretDataKey = "secret"
-	azure              = "azure"
-
-	acrUserAgentPrefix         = "AzureContainerRegistry"
-	azureDevOpsUserAgentPrefix = "VSServices"
-)
-
-const (
 	gitlab              = "gitlab"
 	gitLabSecretDataKey = "secret-token"
 


internal/webhook/external/media_types.go
```diff
@@ -0,0 +1,36 @@
...
+	"application/vnd.docker.distribution.manifest.list.v2+json",
+	"application/vnd.docker.distribution.manifest.v2+json",
+}
+
+func isContainerImageMediaType(mediaType string) bool {
+	return slices.Contains(containerImageMediaTypes, mediaType)
+}
+
+func isHelmChartMediaType(mediaType string) bool {
+	return mediaType == "application/vnd.cncf.helm.config.v1+json"
```
hiddeco: This could be a `const`?

krancour: Fixed in c4e2c2fb73ad04d354b18e3a46b9968547fa1b6d

fuskovic (approved):

PR #4623, opened 2025-07-15
feat: Adding support for Azure Container Registry Workload ID

sbyrne13:
This change is introducing support for workload identity authentication against an Azure Container Registry for the warehouse reconciliation loop.

It includes all documentation changes that I could see as required as well as go functionality similar to the existing ECR implementation. I have also added tests to cover this new code.

All testsuites are passing and I have verified a build of the controller against a personal Azure deployment. 

Any questions please let me know.


krancour:
Thanks for this!

Note the existence of #3751, which has been opened and unreviewed for quite some time. That PR attempts full parity with our ECR and GAR support by first attempting to assume a Project-specific identity before falling back on the controller's own identity. Azure does not make this nearly as easy as AWS and GCP do, however, and #3751 was submitted after a protracted conversation of headache-inducing complexity on that topic (see #2372) which did not produce an agreed-upon approach. Resource constraints and internal priorities have prevented further research into the topic.

At a glance, this PR does not seem to attempt assuming a Project-specific identity and is therefore considerably more straightforward that #3751, and as it at least gets us halfway to decent ACR support w/ Workload Identity, I'm happy to review when time permits.

Thanks again!

sgpbyrne:
Thanks for the feedback @krancour. Yeah I saw there had been some discussion and attempts at this before. 

This push is more aligned with the ECR code layout and structure with an attempt to keep things consistent with the existing approaches in Kargo. I could simplify this further if you would like as all of the caching pieces that I added to align with the other implementations is available as part of the sdk. Let me know when you get some time to review if you would like me to do this simplification or if the current approach is good. 

Regarding project specific identity - is this something you would like added later or what is your thoughts on this? Its definitely possible to build it on top of this as the base but it is not something I considered to be honest as we tend to follow the single identity per deployment/pod approach. This means we don't need to deal with switching client ids etc based on the need. Keda has a nice solution for this but they had to add a specific CRD to allow for the switching of the client IDs [Keda CRD](https://github.com/kedacore/keda/blob/main/config/crd/bases/keda.sh_triggerauthentications.yaml). In the kargo world this could be an CR deployed in the project that just tells the controller what Client ID to use for its warehouse. 

I would have a preference towards introducing just the current content of this PR for now and introducing project level stuff as a phase 2.

krancour:
@sbyrne-teck there is no need to entertain assuming/impersonating a project specific identity as a "phase two" at this time.

ghost:
Hey @krancour  - is it possible to consider including this in the 1.8.0 milestone?

fykaa:
@sbyrne13 We'll aim to get this included in the `v1.9.0` milestone. In the meantime, could you handle the merge conflicts so I can review the PR end-to-end without any blockers?

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,170 @@
...
+	logger := logging.LoggerFromContext(ctx)
+
+	// Try to create a DefaultAzureCredential which supports workload identity
+	credential, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		logger.Info("Azure workload identity not available", "error", err.Error())
+		return nil
+	}
+
+	logger.Info("Azure workload identity credential provider initialized")
```
fykaa, line 58: I think it will help if we log which credential type in the chain was used

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,170 @@
...
+
+	p := &WorkloadIdentityProvider{
+		tokenCache: cache.New(
+			// ACR refresh tokens are typically valid for 3 hours. We'll cache them for 2.5 hours.
+			150*time.Minute, // Default ttl for each entry
+			30*time.Minute,  // Cleanup interval
+		),
```
fykaa, line 64: I feel we can consider adding explicit expiration tracking as well with something like
```
type cachedToken struct {
    Token string
    ExpiresAt time.Time
}
```
wdyt?

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,170 @@
...
+		Password: accessToken,
+	}, nil
+}
+
+// getAccessToken returns an ACR refresh token using Azure workload identity.
+func (p *WorkloadIdentityProvider) getAccessToken(ctx context.Context, registryName string) (string, error) {
+	logger := logging.LoggerFromContext(ctx).WithValues("registryName", registryName)
+
+	// Get Azure AD access token with the standard ACR scope
+	scope := "https://containerregistry.azure.net/.default"
```
fykaa: lets make this a constant: `ACRScope`

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,170 @@
...
+	var (
+		registryName = matches[1]
+		cacheKey     = tokenCacheKey(registryName, project)
+	)
+
+	// Check the cache for the token
+	if entry, exists := p.tokenCache.Get(cacheKey); exists {
+		token := entry.(string) // nolint: forcetypeassert
+		return &credentials.Credentials{
+			Username: "00000000-0000-0000-0000-000000000000", // ACR username for token auth
```
fykaa: lets make this a constant: `ACRTokenUsername`

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,170 @@
...
+	// If we didn't get a token, we'll treat this as no credentials found
+	if accessToken == "" {
+		return nil, nil
+	}
+
+	// Cache the token
+	p.tokenCache.Set(cacheKey, accessToken, cache.DefaultExpiration)
+
+	return &credentials.Credentials{
+		Username: "00000000-0000-0000-0000-000000000000", // ACR username for token auth
```
fykaa: same constant can be used here https://github.com/akuity/kargo/pull/4623/files#r2437936343

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,170 @@
...
+		logger.Info("Azure workload identity not available", "error", err.Error())
+		return nil
+	}
+
+	logger.Info("Azure workload identity credential provider initialized")
+
+	p := &WorkloadIdentityProvider{
+		tokenCache: cache.New(
+			// ACR refresh tokens are typically valid for 3 hours. We'll cache them for 2.5 hours.
+			150*time.Minute, // Default ttl for each entry
```
fykaa: I was thinking it would be cleaner to use `ACRTokenValidityHours` (3), `CacheTTLMinutes` (150) and `CleanupIntervalMin` (30) as constants and then use them here

```suggestion
			CacheTTLMinutes*time.Minute, // Default ttl for each entry
```

fykaa:
@sbyrne-teck also, to help speed up e2e testing on my side, could you please share the workflow or steps you used to test this locally or against your Azure setup (if you've already done so)? That would save me quite a bit of time in verifying the changes

krancour:
> to test this locally

You can't really.

> or against your Azure setup

@fykaa it's not entirely trivial to set up an EKS cluster with workload identity. I will help with this. Please remind me on Monday.

pkg/credentials/kubernetes/acr/common.go
fykaa, line 1: we also need to move the acr implementation to a correct location under `pkg`

fykaa: done

pkg/credentials/kubernetes/acr/workload_identity.go
fykaa: added logs

pkg/credentials/kubernetes/acr/workload_identity.go
fykaa: I think we can even keep the current approach as its simpler, cleaner and follows other patterns in the codebase

ghost:
Hi @fykaa. Apologies for the delayed response and thanks for looking into this. I see you have already implemented the changes - is there anything you would like me to do here?

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,199 @@
...
+	ACRTokenValidityHours = 3
+	// CacheTTLMinutes is how long we cache ACR tokens before refreshing them.
+	// Set to 2.5 hours to ensure we refresh before the 3-hour token expiry.
+	CacheTTLMinutes = 150
+	// CleanupIntervalMinutes is how often the cache cleanup runs
+	CleanupIntervalMinutes = 30
+	// ACRTokenUsername is the fixed username used for ACR token authentication
+	ACRTokenUsername = "00000000-0000-0000-0000-000000000000"
+	// ACRScope is the Azure AD scope required for ACR authentication
+	ACRScope = "https://containerregistry.azure.net/.default"
```
krancour: There is no reason for any of these to be exported.

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,199 @@
...
+		return false
+	}
+
+	if credType != credentials.TypeImage && credType != credentials.TypeHelm {
+		return false
+	}
+
+	if credType == credentials.TypeHelm && !strings.HasPrefix(repoURL, "oci://") {
+		return false
+	}
```
krancour: This needs to be removed because, since #4995, repoURLs are normalized before ever being passed to the providers and normalization for an OCI chart repo URL will have removed the leading `oci://`.

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,199 @@
...
+
+	var (
+		registryName = matches[1]
+		cacheKey     = tokenCacheKey(registryName, project)
+	)
+
+	// Check the cache for the token
+	if entry, exists := p.tokenCache.Get(cacheKey); exists {
+		token := entry.(string) // nolint: forcetypeassert
+		logging.LoggerFromContext(ctx).Debug(
```
krancour: We should really assign the logger retrieved from the context to a local `logger` variable.

pkg/credentials/kubernetes/acr/workload_identity.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to exchange Azure AD token for ACR refresh token: %w", err)
+	}
+
+	if refreshTokenResp.RefreshToken == nil {
+		return "", fmt.Errorf("received empty ACR refresh token")
+	}
+
+	logger.Debug("successfully obtained ACR refresh token")
+	return *refreshTokenResp.RefreshToken, nil
```
krancour: This is going to need some kind of comment explaining it. The Azure nomenclature here is unintuitive. This "refresh token" is really, for all intents and purposes, an access token.

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -341,8 +341,40 @@ controller restart clears the cache.
...
+identity with the `kargo-controller` `ServiceAccount` within the namespace in 
+which Kargo is (or will be) installed.
+
+:::note
+To use workload identity, you will need to annotate the `kargo-controller` 
+`ServiceAccount` with the client ID of the managed identity using the 
+`azure.workload.identity/client-id` annotation. This can be done via the 
+`controller.serviceAccount.annotations` setting in Kargo's Helm chart at 
+installation.
+:::
```
krancour: According to https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview?tabs=dotnet#service-account-annotations:

> All annotations are optional. If the annotation isn't specified, the default value will be used.

One section down from there, at https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview?tabs=dotnet#pod-labels, the following is said about the `azure.workload.identity/use` label:

> This label is required in the pod template spec. Only pods with this label are mutated by the azure-workload-identity mutating admission webhook to inject the Azure specific environment variables and the projected service account token volume.

Through testing, I have confirmed that the annotation mentioned here is _not_ required and that the label _is_.

My suggestion is that we ditch this notes about the optional annotation and add a warning box (harder to miss) with a note about the label.

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -341,8 +341,40 @@ controller restart clears the cache.
...
+If Kargo locates no `Secret` resources matching a repository URL and is deployed
+within an AKS cluster with workload identity enabled, it will attempt to use it 
+to authenticate. Leveraging this eliminates the need to store ACR credentials
+in a `Secret` resource.
+
+Follow 
+[this guide](https://learn.microsoft.com/en-us/azure/aks/workload-identity-deploy-cluster)
+to set up workload identity in your AKS cluster. You will associate a managed 
+identity with the `kargo-controller` `ServiceAccount` within the namespace in 
+which Kargo is (or will be) installed.
```
krancour: We ought to add a caution box (maybe even a danger box) here encouraging the reader to follow this guide as precisely as possible. I made a number of missteps when testing this, and I've done this before... and also used to work at MS in the Azure org. There is a veritable minefield of things that can go wrong here if you miss even one detail of these instructions.

I also want to avoid the quagmire of documenting lots and lots of Azure stuff instead of deferring to their docs, but some kind of checklist to help readers be certain they didn't miss anything critical in the article linked to above may be in order here.

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
fykaa: Yes, I can confirm too that the label `azure.workload.identity/use` was required during testing

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
fykaa: added the checklist and the danger note

krancour:
I did quite a bit more cleanup on this. Some notable changes:

* Regex needed to be fixed to _not_ allow leading `oci://`.

* It was not necessary to factor Project name into token cache keys. Unlike similar components for Amazon and Google, this credential provider makes no attempt to assume/impersonate a Project-specific identity. This means that all Projects access ACR repos _as the controlller_. i.e. Tokens for any given repo can and should be shared across Projects.

* The above means tokens in the cache can be keyed off of _one_ simple value -- the globally unique registry name extracted from the normalized repo URL. Such values are not sensitive, meaning there's also no need whatsoever to hash them.

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
krancour: > Through testing, I have confirmed that the annotation mentioned here is not required 

I must have been mistaken. In my latest round of tests, I can not make things work without this annotation.

Azure docs do claim this annotation is not required, however, after more research, in practice, this is often not the actually the case.

krancour (approved):

PR #4624, opened 2025-07-15
docs: Fix disabling Argo Rollouts integration in common config docs

gazab:
Changes the docs to use the correct Helm values for disabling ArgoCD Rollouts integration. Stumbled upon this when trying to disable it myself.

Correct Helm values variables:
- https://github.com/akuity/kargo/blob/main/charts/kargo/values.yaml#L321-L323
- https://github.com/akuity/kargo/blob/main/charts/kargo/values.yaml#L490-L492


krancour (approved):
Thank you!

PR #4634, opened 2025-07-17
added a YAML Merger PromotionStep

prune998:
This PR add a new Promotion `Step` that caan merge multiple YAML files into a single file.

This is useful in the scenario where, for example, multiple promotions  (ex: one per cluster) merges content in the same repo/branch.
Then Helm will pick up the values from multiple locations, but updating `global` values could impact many clusters at the same time.

With this step, we are able to pre-merge `global` values with more specific values and only commit one single `values.yaml` file for each specific cluster. No more `globals`.


prune998:
Any way to get someone trigger the CI build ?
thx

internal/yaml/yaml_test.go
```diff
@@ -169,3 +169,184 @@ characters:
...
+			},
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			b, err := MergeYAMLFiles(testCase.inputs)
+
+			// fmt.Println(b)
+			// fmt.Println(err)
+			// fmt.Println("------")
```
fuskovic:
```suggestion
```

internal/yaml/yaml_test.go
```diff
@@ -169,3 +169,184 @@ characters:
...
+			// Note: This YAML is invalid because one line is indented with a tab
+			inputs: []string{},
+			assertions: func(t *testing.T, output string, err error) {
+				require.ErrorContains(t, err, "empty input list provided")
+				require.Empty(t, output)
+			},
+		},
+		{
+			name: "empty input YAML",
+			// Note: This YAML is invalid because one line is indented with a tab
```
fuskovic: This doesn't look like it applies here

internal/yaml/yaml.go
```diff
@@ -146,3 +149,31 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
...
+
+// mergeYAMLFiles merges a list of yaml strings
+func MergeYAMLFiles(inputs []string) (string, error) {
+	if len(inputs) == 0 {
+		return "", fmt.Errorf("empty input list provided")
+	}
+
+	mergedNode, err := kyaml.Parse(inputs[0])
+	if err != nil {
+		return "", err
```
fuskovic:
```suggestion
		return "", fmt.Errorf("error parsing input: %w", err)
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// write yaml file
+	outPath, err := securejoin.SecureJoin(stepCtx.WorkDir, cfg.OutPath)
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("could not secure join outPath %q: %w", cfg.OutPath, err)
+	}
+
+	if err := os.MkdirAll(filepath.Dir(outPath), 0o700); err != nil {
+		return failure, fmt.Errorf("Error creating directory structure %s: %w", filepath.Dir(outPath), err)
```
fuskovic: For consistency:

```suggestion
		return failure, fmt.Errorf("error creating directory structure %s: %w", filepath.Dir(outPath), err)
```

There are other places where this applies as well.

internal/yaml/yaml.go
```diff
@@ -146,3 +149,31 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
...
+	}
+
+	for i := 1; i < len(inputs); i++ {
+		patchNode, err := kyaml.Parse(inputs[i])
+		if err != nil {
+			if err == io.EOF {
+				continue
+			}
+			return "", err
```
fuskovic:
```suggestion
			return "", fmt.Errorf("error parsing input at index %d: %w", i, err)
```

internal/yaml/yaml.go
```diff
@@ -146,3 +149,31 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
...
+		}
+		mergedNode, err = merge2.Merge(patchNode, mergedNode, kyaml.MergeOptions{ListIncreaseDirection: 1})
+		if err != nil {
+			return "", err
```
fuskovic:
```suggestion
			return "", fmt.Errorf("error merging node at index %d: %w", i, err)
```

fuskovic:
Stopped by for a drive-by review. 

Concept looks reasonable to me. 

Related to https://github.com/akuity/kargo/issues/4766

Linter needs to be satisfied. 

I'll defer the rest to @hiddeco and @krancour

prune998:
thanks for the review @fuskovic. Everything should be resolved + rebased now. 
I'll keep watching the workflows when they run, in case any other issue is detected.

prune998:
Anyone to review this ?

krancour:
@prune998 I know this isn't what anyone wants to hear, but our priorities are elsewhere right now and there may be a bit of a wait for a thorough review.

Some discussion may also be in order. I see this is meant to address #4766, but no maintainer triaged that feature request or moved it out of proposal status. It's a gamble putting the effort into a feature without those conversations happening first.

In all probability, this _is_ a worthwhile feature and it _will_ be reviewed eventually and _probably_ merged, but some patience may be required.

If it makes you feel any better, I have PRs open to some other projects that are old enough to be starting kindergarten. Our backlog isn't _that_ bad, thankfully.

krancour:
Small addendum... there are some notoriously well-known gotchas when merging YAML, so this isn't a review I'd rush through and that contributes to the above.

prune998:
Thanks for the comment @krancour . I understand you're busy with your customers features.
The way Kargo is designed forces us to embed the `promotionStep` functions inside the code, which creates friction. Got it.

What you could do, would be 
- have someone do a quick review, ensuring that it works and it is useful, and more importantly, is not a security or stability risk
- mark that function as alpha, not supported
- add me as support of the feature and assign me comments/issues if any
- release the feature

This way, your team will not be much impacted, the features is released, and you have a "feature maintainer". 

This is just a proposition and you can refuse it. I'll wait, or keep using my forked version... but it will be a miss :)
Thanks for your time and effort anyway.

Side note on YAML merging: yeah, there's a ton of gotchas... this is why I used a "well known" library to do it and not try to re-implement it myself. The library I used is also opinionated. But it is intended for K8s yaml style. I also tested others, with varying results... Anyway, they all have the same limitations (biggest one being no merge of lists, latest list takes precedence), like Helm does, so people knowing Helm will not be disturbed.
Once I'm sure you're interrested in this feature and the code, I can add more examples in the docs and test cases in the code to better demonstrate all sort of merges. I just don't want to spend too much time if you are to refuse it. All I can say is that it is working "like Helm" for me at the moment.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,71 @@
...
+description: Merge multiple YAML file into a single file.
+---
+
+# `yaml-merge`
+
+`yaml-merge` merges multiple YAML files into a single file. This step
+most often used to merge multiple Helm values.yaml files into a single
+file and is commonly followed by a [`helm-template` step](helm-template.md).
+YAML files are merged in order, so the first one is the base, and all
+subsequent files are "overlays", modifying the default values.
```
krancour: One comment -- two reasons for making it:

`helm-template` step, just like the `helm template` command allows you to specify multiple `values.yaml` files. So:

1. On account of that, merging multiple `values.yaml` files prior to a `helm-template` step probably _won't_ be a particularly common use for this. I do still think the proposed step is _generally_ useful. It would be fine to document what the step does without calling out any specific, common use.

1. I suspect your own use case actually varies from the above, but on the odd chance that it does not, I just wanted you to know about existing options already available to you.

internal/promotion/runner/builtin/schemas/yaml-merge-config.json
```diff
@@ -0,0 +1,28 @@
...
+    "outPath": {
+      "type": "string",
+      "description": "OutPath is the path to the merged YAML file to created or updated.",
+      "minLength": 1
+    },
+    "strict": {
+      "type": "boolean",
+      "description": "Strict will cause the directive to fail if the input path does not exist.",
+      "default": false
+    }
```
krancour: I think I know why you want this option, but I'm contemplating three things:

* It feels a bit out of the ordinary. `helm` wouldn't like it if you gave it a path to a non-existent file as input. (Which, I think, is actually why you want this?) Generalizing, I can think of very few commands, tools, functions, etc. that are even optionally tolerant of a filename used as input referencing a file that doesn't exist.

  I'm not saying no to this, but I think I'm right to question whether this is generally useful of if it caters to a very niche use case.

* Is `strict` the right name for this option? I appreciate that you've documented it clearly, however, ideally, field names would be at least _somewhat_ self-documenting so all those folks who don't read documentation can easily infer what it means. I'm wondering if there's a different name we could consider that is more transparent about what it does.

* As mentioned in a previous bullet, I think the behavior of ignoring a file that doesn't exist is _unusual_. Having it as the default and needing to opt _into_ more conventional behavior feels wrong.

I'd like a second opinion on much of this from @hiddeco, but he is out for a couple weeks.

hiddeco: My experience is that people like to be able to do this to allow the drop-in of an optional file that is capable of overwriting any other path. They can then instruct their team to put this file into the specific directory they manage to provide their custom values, or set custom settings for a specific microservice or instance without having to duplicate (Kargo) configuration.

I think this feature can be useful, but I also like strict declarative behavior to prevent surprises that otherwise can go unnoticed. Because of this, I would not set this as a global configuration option, but rather allow it to be set _per path_. Additionally, I would not mark is as `strict` but instead declare the path as `optional`.

For example:

```yaml
config:
  inPaths:
  # Base configuration (required)
  - path: "config/base-config.yaml"
    optional: false # Default value

  # Environment-specific config (required)
  - path: "config/environments/production.yaml"
  
  # Team-specific overrides (optional)
  - path: "config/team-overrides.yaml"
    optional: true
  
  # Instance-specific customizations (optional)
  - path: "config/instance-specific.yaml"
    optional: true

  outPath: "config/merged-config.yaml"
```

prune998: Thanks for the comments and analysis. 

I used `strict` as it is already used the same way in the [delete](https://github.com/akuity/kargo/blob/main/internal/promotion/runner/builtin/schemas/delete-config.json#L14) step.

I added this feature to replicate how [argocd}(https://argo-cd.readthedocs.io/en/latest/user-guide/helm/#values-files) do it :

```
If Helm is passed a non-existing value file during template expansion, it will error out. Missing values files can be ignored (meaning, not passed to Helm) using the --ignore-missing-value-files.
```

Skipping non-existing files is supported by ArgoCD, I think it was an option added in late 2.6+ versions.  

Would it be OK to rename the `strict` keyword to `ignoreMissingFiles` ? 

@hiddeco I also don't feel the `strict` setting should be per-file... it's a global behaviour, either you require all the files or you don't. 

The goal for me is to be able to create a single ArgoCD configuration, listing all the possible files, and leave each project to use them or not. We're using ApplicationSets, so it is even more important.

Going back to my example, in a Git project, I can have multiple `services`, deploying in multiple `envs` and in multiple `clusters` in each env. So the structure is something like:

```
.
├── global
│   └── values.yaml //apply to ALL services
└── service-A
  ├── base
  │   └── values.yaml // apply to ALL envs for service-A
  ├── Chart.yaml
  ├── prod
  │   ├── values.yaml // apply to ALL production clusters
  │   ├── prod-cluster-1
  │   │   └── values.yaml // apply to a single cluster prod-cluster-1
  │   ├── prod-cluster-2
  │   │   └── values.yaml
  ├── qa
  │   ├── values.yaml
  │   ├── qa-cluster
  ├── stg
  │   ├── stg-cluster-1
  │   │   └── values.yaml
  │   ├── values.yaml
  │   ├── stg-cluster-2
  │   │   └── values.yaml
  ├── templates
      ├── monitoring.yaml
```

All my Applications (from Appset) are built the same, but all files are not defined in each project.

That means that all my Kargo Promotions have to also copy all those files in my `deployment` branch.
It also means that changing the values in file `global/values.yaml` will update ALL the ArgoCD applications at once because they all depends on it. This is the issue i'm trying to solve.

By pre-merging the files, The ArgoCD App will only have to read the more specific file that contains the merge of everything else. 
So when I change the global file, nothing happen, as this file will not be part of the final files copied to the `deployment` branch. 
Instead, the values in this global files will be merged in the more specific `service-A/prod/prod-cluster-1/values.yaml` when I trigger the promotion.

I know there may be different ways of achieving this kind of scenario. But I feel this is a valid one, and this is my solution to make it happen :)

I updated the code to reflect all your comments + rebase.
Please let me know what you think.

Thanks !

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
prune998: I reworked that and removed reference to Helm

prune998: I also explained my use case in another comment. Not sure if you want that use-case to appear in the docs ?

krancour: Your use case is a bit complex. (This is not to say invalid. It just takes some effort to understand.) I'm not positive we'd be doing readers any favors by asking them to understand that as a means to seeing the utility of this step. i.e. Don't want to demo simple functionality with a complex use case.

pkg/promotion/runner/builtin/schemas/yaml-merge-config.json
```diff
@@ -0,0 +1,28 @@
...
+    "outPath": {
+      "type": "string",
+      "description": "OutPath is the path to the merged YAML file to created or updated.",
+      "minLength": 1
+    },
+    "ignoreMissingFiles": {
+      "type": "boolean",
+      "description": "allow directive to pass even if an input file does not exist.",
+      "default": false
+    }
```
krancour, line 26: Just want to say thank you for this. I like this a lot better that "strict" and being false so it's opt-in feels correct. Really appreciate it.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
prune998: > Your use case is a bit complex. (This is not to say invalid. It just takes some effort to understand.) I'm not positive we'd be doing readers any favors by asking them to understand that as a means to seeing the utility of this step. i.e. Don't want to demo simple functionality with a complex use case.

I updated the docs to simply explain how it work and not describe my use case. I'm not sure readers needs a complex use case here. Basic use case is simple: you have many yaml files and you want to merge them into one.

The description of the use case for reviewers of the PR/issue is at https://github.com/akuity/kargo/pull/4634#discussion_r2316155956

Please let me know if you feel better with this example.

pkg/promotion/runner/builtin/schemas/yaml-merge-config.json
prune998: it feels better to me too

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
+---
+sidebar_label: yaml-merge
+description: Merge multiple YAML files into a single file.
```
krancour: For consistency with the descriptions of other steps.

```suggestion
description: Merges multiple YAML files into a single file.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+---
+
+# `yaml-merge`
+
+`yaml-merge` merges multiple YAML files into a single file.
+YAML files are merged in order.The first file in the list
```
krancour:
```suggestion
YAML files are merged in order. The first file in the list
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+is the source, and all subsequent files are applied over it.
+
+When `ignoreMissingFiles` is false (default), the Task will fail
+if any file from `inPaths` does not exist.
+
+
```
krancour, line 14: Most markdown linters don't like double line breaks.

```suggestion
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+description: Merge multiple YAML files into a single file.
+---
+
+# `yaml-merge`
+
+`yaml-merge` merges multiple YAML files into a single file.
+YAML files are merged in order.The first file in the list
+is the source, and all subsequent files are applied over it.
+
+When `ignoreMissingFiles` is false (default), the Task will fail
```
krancour: Because "Task" (e.g. PromotionTask) has a different meaning, this would be more clear:

```suggestion
When `ignoreMissingFiles` is false (default), the step will fail
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+if any file from `inPaths` does not exist.
+
+
+:::note
+Merging is done with usual constrains:
+- new objects are added
+- object with same name are modified
+- lists are replaced by latest version (no merge)
+- null values delete the object
```
krancour: The wording of this looks a bit inaccurate. I asked Copilot to reword more accurately and it gave me this:

```suggestion
Merging is performed as follows:

- **Scalar values:** If both documents define a scalar (string, number, boolean)
  at the same key, the value from the second document overrides the first.

- **Mapping (object) values:** If both documents define a mapping at the same
  key, the mappings are merged recursively.

- **Sequence (array) values:** If both documents define a sequence at the same
  key, the sequence from the second document replaces the first (no merging).

- **Keys present only in one document:** Keys that exist in only one document
  are included as-is in the result.

- **Null values:** If a key is set to `null` in the second document, it removes
  or overrides the value from the first document.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+:::
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `inPaths` | `[]string` | Y | Paths to a YAML files. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `outPath` | `string`   | Y | The path to the output file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
```
krancour: `inFiles` and `outFile` strike me as possibly better, but I'm not positive. There may even be better alternatives than those. cc @hiddeco for any input.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+| `ignoreMissingFiles` | `bool` | N | When set to true, the directive will skip input files that does not exist. Defaults to `false`. |
+
+## Output
+
+| Name | Type | Description |
+|------|------|-------------|
+| `commitMessage` | `string` | A description of the change(s) applied by this step. Typically, a subsequent [`git-commit` step](git-commit.md) will reference this output and aggregate this commit message fragment with other like it to build a comprehensive commit message that describes all changes. |
```
krancour: I know this typo was copied from elsewhere in the docs. I'll fix other occurrences of it in a separate PR, but let's set this one straight.

```suggestion
| `commitMessage` | `string` | A description of the change(s) applied by this step. Typically, a subsequent [`git-commit` step](git-commit.md) will reference this output and aggregate this commit message fragment with others like it to build a comprehensive commit message that describes all changes. |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Examples
+
+### Common Usage
+
+In this example, three Helm value files, one global, one more specific
+for the QA environment and a last one specific to the cluster, are merged
+into a new single file, that is then commited to the deployment branch.
```
krancour: Clarity and terminology. (We avoid the word "deploy.")

```suggestion
In the following example, multiple Helm values files (one "base", a second with environment-specific overrides, and a third with cluster-specific overrides) are
merged into a new, single file, which is then committed to a Stage-specific
branch:
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+
+This pattern is commonly used when you need to merge multiple value files
+into a final `values.yaml` file, to be used by Helm.
```
krancour: I think the previous bit of text got the point across well enough and we can cut this:

```suggestion
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+      path: ./out
+- uses: git-clear
+  config:
+    path: ./out
+- uses: yaml-merge
+  config:
+    inPaths:
+      - ./src/charts/my-chart/values.yaml
+      - ./src/charts/qa/values.yaml
+      - ./src/charts/qa/cluster-a/values.yaml
```
krancour: We favor this syntax throughout the project and this change would also match the syntax used earlier in the same example:

```suggestion
    - ./src/charts/my-chart/values.yaml
    - ./src/charts/qa/values.yaml
    - ./src/charts/qa/cluster-a/values.yaml
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+    outPath: ./out/charts/my-chart/values.yaml
+# Render manifests to ./out, commit, push, etc...
+```
+
+
```
krancour:
```suggestion
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,104 @@
...
+Given the sample input YAMLs:
```
krancour:
```suggestion
Given the following sample input files:
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/xeipuuv/gojsonschema"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/internal/yaml"
+	"github.com/akuity/kargo/pkg/promotion"
+	"github.com/akuity/kargo/pkg/x/promotion/runner/builtin"
+)
+
+// yamlMerger is an implementation of the promotion.StepRunner interface that
+// updates do a merge of the multiple YAML files.
```
krancour:
```suggestion
// yamlMerger is an implementation of the promotion.StepRunner interface that
// merges multiple YAML files.
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+type yamlMerger struct {
+	schemaLoader gojsonschema.JSONLoader
+}
+
+// newYAMLMerger returns an implementation of the promotion.StepRunner interface
+// that updates the values of specified keys in a YAML file.
```
krancour:
```suggestion
// newYAMLMerger returns an implementation of the promotion.StepRunner interface
// merges multiple YAML files.
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+		return failure, fmt.Errorf("could not convert config into %s config: %w", y.Name(), err)
+	}
+
+	return y.run(ctx, stepCtx, cfg)
+}
+
+// validate validates yamlMerger configuration against a JSON schema.
+func (y *yamlMerger) validate(cfg promotion.Config) error {
+	return validate(y.schemaLoader, gojsonschema.NewGoLoader(cfg), y.Name())
+}
```
krancour: All existing steps have converged on this pattern:

```suggestion
// Run implements the promotion.StepRunner interface.
func (y *yamlMerger) Run(
	ctx context.Context,
	stepCtx *promotion.StepContext,
) (promotion.StepResult, error) {
	cfg, err := y.convert(stepCtx.Config)
	if err != nil {
		return promotion.StepResult{
			Status: kargoapi.PromotionStepStatusFailed,
		}, &promotion.TerminalError{Err: err}
	}
	return y.run(ctx, stepCtx, cfg)
}

// convert validates yamlMerger configuration against a JSON schema and
// converts it into a builtin.YAMLMergeConfig struct.
func (y *yamlMerger) convert(
	cfg promotion.Config,
) (builtin.YAMLMergeConfig, error) {
	return validateAndConvert[builtin.YAMLMergeConfig](y.schemaLoader, cfg, f.Name())
}
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (y *yamlMerger) run(
+	_ context.Context,
+	stepCtx *promotion.StepContext,
+	cfg builtin.YAMLMergeConfig,
+) (promotion.StepResult, error) {
+
+	result := promotion.StepResult{Status: kargoapi.PromotionStepStatusSucceeded}
+	failure := promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored}
```
krancour: I know this is a nitpicky one, but we've generally avoided this pattern of instantiating one or more probable results up front and returning them at various points in the function because it often requires one reading/editing the code to mentally track what their state may be at any given early return point. Instantiating an appropriate `promotion.StepResult` just-in-time (as you're already doing in most places, actually) would be preferable.

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, path := range cfg.InPaths {
+		inPath, err := securejoin.SecureJoin(stepCtx.WorkDir, path)
+		if err != nil {
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("could not secure join inPath %q: %w", path, err)
+		}
+
+		inBytes, err := os.ReadFile(inPath)
+		if err != nil {
+			// we skip if file does not exist
```
krancour:
```suggestion
			// Skip if file does not exist and the config has opted into ignoring
			// that.
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+				continue
+			}
+			return failure, fmt.Errorf(
+				"error reading file %q: %w",
+				inPath,
+				err,
+			)
+		}
+
+		// we skip if file is empty
```
krancour:
```suggestion
		// Skip if file is empty.
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+func (y *yamlMerger) run(
+	_ context.Context,
+	stepCtx *promotion.StepContext,
+	cfg builtin.YAMLMergeConfig,
+) (promotion.StepResult, error) {
+
+	result := promotion.StepResult{Status: kargoapi.PromotionStepStatusSucceeded}
+	failure := promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored}
+
+	mergedFiles := []string{} // keep track of files actually merged
```
krancour: It would be best to size this slice appropriately up front:

```suggestion
	// Keep track of files actually merged.
	mergedFiles := make([]string, 0, len(cfg.InPaths))
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	if err = os.MkdirAll(filepath.Dir(outPath), 0o700); err != nil {
+		return failure, fmt.Errorf("error creating directory structure %s: %w", filepath.Dir(outPath), err)
+	}
+	if err = os.WriteFile(outPath, []byte(outYAML), 0o600); err != nil {
+		return failure, fmt.Errorf("error writing to file %s: %w", outPath, err)
+	}
+
+	// add commit msg
```
krancour:
```suggestion
	// Add a commit message fragment to the step's output.
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	// merge YAML files
+	outYAML, err := yaml.MergeYAMLFiles(yamlData)
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("could not merge YAML files: %w", err)
+	}
+
+	// write yaml file
```
krancour:
```suggestion
	// Write out a single YAML file.
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+		// we skip if file is empty
+		if len(inBytes) == 0 {
+			continue
+		}
+
+		mergedFiles = append(mergedFiles, path)
+		yamlData = append(yamlData, string(inBytes))
+	}
+
+	// merge YAML files
```
krancour: Nit: You don't have files in-hand here. You're merging data that's in-memory.

```suggestion
	// Merge
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+	if commitMsg := y.generateCommitMessage(cfg.OutPath, mergedFiles); commitMsg != "" {
+		result.Output = map[string]any{
+			"commitMessage": commitMsg,
+		}
+	}
+	return result, nil
+}
+
+func (y *yamlMerger) generateCommitMessage(path string, fileList []string) string {
+	if len(path) <= 1 {
```
krancour: I think you're checking the length of the wrong argument?

```suggestion
	if len(fileList) == 0 {
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+		return "no YAML file merged"
```
krancour:
```suggestion
		return "no YAML files merged"
```

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,152 @@
...
+
+	var commitMsg strings.Builder
+	_, _ = commitMsg.WriteString(fmt.Sprintf("Merged YAML files to %s\n", path))
+	for _, file := range fileList {
+		_, _ = commitMsg.WriteString(
+			fmt.Sprintf(
+				"\n- %s",
+				file,
+			),
+		)
```
krancour: A bit more readable. We don't "go vertical" with args until we pass the soft limit of 80 chars.

```suggestion
		_, _ = commitMsg.WriteString(fmt.Sprintf("\n- %s", file))
```

internal/yaml/yaml.go
```diff
@@ -146,3 +149,31 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
 	}
 	return 0, 0, fmt.Errorf("key path not found")
 }
+
+// mergeYAMLFiles merges a list of yaml strings
```
krancour:
```suggestion
// MergeYAMLFiles merges a list of yaml strings.
```

internal/yaml/yaml.go
```diff
@@ -146,3 +149,31 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
...
+func MergeYAMLFiles(inputs []string) (string, error) {
+	if len(inputs) == 0 {
+		return "", fmt.Errorf("empty input list provided")
+	}
+
+	mergedNode, err := kyaml.Parse(inputs[0])
```
krancour: The docs on `kyaml.Parse()` suggest:

```
To parse multiple resources, consider a kio.ByteReader
```

I've not used kyaml directly before and I've not dug too deep into this suggestion, but something about the suggestion resonates...

It had already felt a little odd that we're taking YAML bytes and turning them to strings to pass to here. It feels equally odd that we're _returning_ a string from here. Bytes in and bytes out seems more conventional for this sort of thing and that note above in their docs hints that there are optimizations to be had here if we're doing things that way.

krancour: Put another way: Consider that we _read bytes_ from the input files and we _write bytes_ to the output file. If kyaml can operate directly on the bytes using readers, writers, buffers, etc., there's no reason to ever bring strings into the equation at all, especially given how inefficient string manipulation tends to be with lots and lots of allocations required.

internal/promotion/runner/builtin/yaml_merger.go
prune998: not sure why I made that this way :)

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
krancour: I do just want to hear what @hiddeco thinks if he has a moment at some point.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
krancour: It seems you kindly beat me to fixing it in those other locations. I very much appreciate that!

internal/promotion/runner/builtin/yaml_merger.go
krancour: No worries at all. I admit I'm a stickler for maximizing readability. Thanks for putting up with me.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
krancour: But let's consider it non-blocking. If everything else outstanding is cleared up before @hiddeco weighs in, then `inFiles` and `outFile` (as you now have them) will be just fine.

internal/promotion/runner/builtin/yaml_merger.go
krancour: Non-blocking.

internal/yaml/yaml.go
krancour: This is the only remaining thing I would consider a blocker.

The PR is in excellent shape!

internal/promotion/runner/builtin/yaml_merger.go
prune998: ok, done

internal/yaml/yaml.go
prune998: I reworked it. using `kio` stuff is really meant to parse multiple yaml documents in a single file, and was adding too much complexity.
I made the decision to read/modify/write the file in the `MergeYAMLFiles` func. 
So the builtin `step` function only sanitize the data.

This way we get rid of string -> bytes manipulations and the code is even smaller.
I re-worked the tests to even better check file generation.

I'm resolving this comment. Please feel free to add more :)

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
hiddeco: Not a strong preference, but I believe `inPath(s)` are used in other steps (e.g. `untar`, which may make the API more predictable for Kargo as a whole.

krancour:
@prune998 at a glance, a83788f6da6ad041e54ba989bb356de8229f5cd1 seems like a great change. I want to look closer on Monday. Barring any surprises, I think we'll be all set then.

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,134 @@
...
+func (y *yamlMerger) run(
+	_ context.Context,
+	stepCtx *promotion.StepContext,
+	cfg builtin.YAMLMergeConfig,
+) (promotion.StepResult, error) {
+
+	// sanity check
+	if len(cfg.InFiles) == 0 || cfg.OutFile == "" {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored}, fmt.Errorf("inFiles and OutFile must not be empty")
+	}
```
krancour: Wouldn't validation against the schema already have caught either of these issues?

internal/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,134 @@
...
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored}, fmt.Errorf("could not secure join input file %q: %w", path, err)
+		}
+
+		// only add existing files
+		_, err = os.Stat(inFile)
+		if err != nil {
+			if cfg.IgnoreMissingFiles {
+				continue
+			}
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored}, fmt.Errorf("input file not found:  %s", inFile)
```
krancour: We don't actually know that "file not found" was the problem because we haven't checked `os.IsNotExist(err)`. I think we ought to do so we don't produce an error message indicating "file not found" when the problem may have been something else entirely.

internal/yaml/yaml.go
```diff
@@ -146,3 +148,37 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
 	}
 	return 0, 0, fmt.Errorf("key path not found")
 }
+
+// MergeYAMLFiles merges a list of YAML files.
+func MergeYAMLFiles(inputPaths []string, outputPath string) error {
```
krancour: We would usually prefer not to repeat context that can be inferred from the package name. This will be used as `yaml.MergeYAMLFiles()`, which sounds redundant. `yaml.MergeFiles()` is no less clear.

```suggestion
// MergeFiles merges a list of YAML files.
func MergeFiles(inputPaths []string, outputPath string) error {
```

internal/yaml/yaml.go
```diff
@@ -146,3 +148,37 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
...
+	// read all other YAML file and apply the patch
+	for i := 1; i < len(inputPaths); i++ {
+		patchNode, err := kyaml.ReadFile(inputPaths[i])
+		if err != nil {
+			return fmt.Errorf("error parsing input file %s: %w", inputPaths[i], err)
+		}
+
+		mergedNode, err = merge2.Merge(patchNode, mergedNode, kyaml.MergeOptions{ListIncreaseDirection: 1})
+		if err != nil {
+			return fmt.Errorf("error merging node at index %d: %w", i, err)
```
krancour: This error message would probably be more useful if instead of talking about nodes it included the path of the file that couldn't be merged.

internal/promotion/runner/builtin/yaml_merger.go
prune998: change to
```
"error accessing file %s: %w", inFile, err
```

prune998:
I guess I finally answered all the comments !

prune998:
I just updated the tests and linter comments

prune998:
@krancour, rebased and updated with the latest move to `pkg` folder. Can you please trigger pipelines see if we can merge ?
thx !

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,141 @@
...
+// newYAMLMerger returns an implementation of the promotion.StepRunner interface
+// that merges multiple YAML files.
+func newYAMLMerger(promotion.StepRunnerCapabilities) promotion.StepRunner {
+	return &yamlMerger{schemaLoader: getConfigSchemaLoader(stepKindYAMLMerge)}
+}
+
+// Name implements the promotion.StepRunner interface.
+func (y *yamlMerger) Name() string {
+	return "yaml-merge"
+}
```
krancour: No longer required by the recently refactored StepRunner interface. Nothing outside the runner will call it now. I suggest removing it and using `stepKindYAMLMerge` directly anywhere that this runner needs to use it internally.

```suggestion
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,141 @@
...
+			Status: kargoapi.PromotionStepStatusFailed,
+		}, &promotion.TerminalError{Err: err}
+	}
+	return y.run(ctx, stepCtx, cfg)
+}
+
+// convert validates yamlMerger configuration against a JSON schema and
+// converts it into a builtin.YAMLMergeConfig struct.
+func (y *yamlMerger) convert(cfg promotion.Config) (builtin.YAMLMergeConfig, error) {
+	return validateAndConvert[builtin.YAMLMergeConfig](y.schemaLoader, cfg, y.Name())
```
krancour:
```suggestion
	return validateAndConvert[builtin.YAMLMergeConfig](y.schemaLoader, cfg, stepKindYAMLMerge)
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// validate validates yamlMerger configuration against a JSON schema.
+func (y *yamlMerger) validate(cfg promotion.Config) error {
+	return validate(y.schemaLoader, gojsonschema.NewGoLoader(cfg), y.Name())
```
krancour:
```suggestion
	return validate(y.schemaLoader, gojsonschema.NewGoLoader(cfg), stepKindYAMLMerge)
```

krancour:
@fykaa please make sure this is in good shape and be sure to see it working end-to-end.

Best effort to get this into v1.8.0, but not guaranteed.

prune998:
@fykaa I just applied the latest requested changes + rebased... 
thx for your help !

prune998:
any news @krancour or @fykaa ?
thx !

fykaa:
@prune998 
I'm testing things end-to-end to make sure it's solid

prune998:
OK thanks. Let me know if I can help.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,106 @@
...
+  or overrides the value from the first document.
+:::
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `inFiles` | `[]string` | Y | Paths to a YAML files. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `outFile` | `string`   | Y | The path to the output file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `ignoreMissingFiles` | `bool` | N | When set to true, the directive will skip input files that does not exist. Defaults to `false`. |
```
fykaa:
```suggestion
| `ignoreMissingFiles` | `bool` | N | When set to `true`, the directive will skip input files that does not exist. Defaults to `false`. |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,106 @@
...
+|------|------|-------------|
+| `commitMessage` | `string` | A description of the change(s) applied by this step. Typically, a subsequent [`git-commit` step](git-commit.md) will reference this output and aggregate this commit message fragment with others like it to build a comprehensive commit message that describes all changes. |
+
+## Examples
+
+### Common Usage
+
+In the following example, multiple Helm values files (one "base", a second with
+environment-specific overrides, and a third with cluster-specific overrides) are
+merged into a new, single file, which is then committed to a Stage-specific
```
fykaa: nit
```suggestion
merged into a new, single file, which is then committed to a `Stage`-specific
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,136 @@
...
+// convert validates yamlMerger configuration against a JSON schema and
+// converts it into a builtin.YAMLMergeConfig struct.
+func (y *yamlMerger) convert(cfg promotion.Config) (builtin.YAMLMergeConfig, error) {
+	return validateAndConvert[builtin.YAMLMergeConfig](y.schemaLoader, cfg, stepKindYAMLMerge)
+}
+
+// validate validates yamlMerger configuration against a JSON schema.
+func (y *yamlMerger) validate(cfg promotion.Config) error {
+	return validate(y.schemaLoader, gojsonschema.NewGoLoader(cfg), stepKindYAMLMerge)
+}
```
fykaa: validation is already happening in the `convert()` function via `validateAndConvert()`, so there's no need for a separate `validate()` method imo... unless there's a specific use case where you need to validate configuration without converting it (which doesn't appear to be the case here)
```suggestion
```

fykaa (changes requested):
I've tested this feature end to end, both with and without the `ignoreMissingFiles` field and everything works as expected: image

Merged file:
image

@prune998 Just a few minor nits I mentioned above once those are fixed, I think we're good to go

prune998:
@fykaa I just rebased and applied your comments. Thanks for the review !

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,131 @@
...
+	result := promotion.StepResult{Status: kargoapi.PromotionStepStatusSucceeded}
+	if commitMsg := y.generateCommitMessage(cfg.OutFile, filePaths); commitMsg != "" {
+		result.Output = map[string]any{
+			"commitMessage": commitMsg,
+		}
+	}
+	return result, nil
+}
+
+func (y *yamlMerger) generateCommitMessage(path string, fileList []string) string {
```
fykaa: I think `generateCommitMessage` should have tests as well

pkg/promotion/runner/builtin/yaml_merger_test.go
```diff
@@ -0,0 +1,209 @@
...
+	// 	t.Run(tc.name, func(t *testing.T) {
+	// 		err := runner.validate(tc.config)
+	// 		if tc.expectedError == "" {
+	// 			assert.NoError(t, err)
+	// 		} else {
+	// 			assert.Error(t, err)
+	// 			assert.Contains(t, err.Error(), tc.expectedError)
+	// 		}
+	// 	})
+	// }
```
fykaa: we can get rid of this
```suggestion
```

pkg/yaml/yaml.go
```diff
@@ -146,3 +148,37 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
 	}
 	return 0, 0, fmt.Errorf("key path not found")
 }
+
+// MergeFiles merges a list of YAML files.
+func MergeFiles(inputPaths []string, outputPath string) error {
+	if len(inputPaths) == 0 || outputPath == "" {
+		return fmt.Errorf("inFiles and OutFile must not be empty")
```
fykaa:
```suggestion
		return fmt.Errorf("inFiles and outFile must not be empty")
```

pkg/yaml/yaml.go
```diff
@@ -146,3 +148,37 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
...
+		patchNode, fileErr := kyaml.ReadFile(inputPaths[i])
+		if fileErr != nil {
+			return fmt.Errorf("error parsing input file %s: %w", inputPaths[i], fileErr)
+		}
+
+		mergedNode, err = merge2.Merge(patchNode, mergedNode, kyaml.MergeOptions{ListIncreaseDirection: 1})
+		if err != nil {
+			return fmt.Errorf("error merging file %s: %w", inputPaths[i], err)
+		}
+	}
```
fykaa: I feel maybe this can optimise the logic, 
```suggestion
	// read first YAML file
	mergedNode, err := kyaml.ReadFile(inputPaths[0])
	if err != nil {
		return fmt.Errorf("error parsing first input file: %w", err)
	}

	// read all other YAML files and merge them
	for _, inputPath := range inputPaths[1:] {
		patchNode, err := kyaml.ReadFile(inputPath)
		if err != nil {
			return fmt.Errorf("error parsing input file %s: %w", inputPath, err)
		}

		mergedNode, err = merge2.Merge(patchNode, mergedNode, kyaml.MergeOptions{ListIncreaseDirection: 1})
		if err != nil {
			return fmt.Errorf("error merging file %s: %w", inputPath, err)
		}
	}
```

pkg/yaml/yaml.go
```diff
@@ -146,3 +148,37 @@ func findScalarNode(node *yaml.Node, keyPath []string) (int, int, error) {
...
+
+	// read all other YAML file and apply the patch
+	for i := 1; i < len(inputPaths); i++ {
+		patchNode, fileErr := kyaml.ReadFile(inputPaths[i])
+		if fileErr != nil {
+			return fmt.Errorf("error parsing input file %s: %w", inputPaths[i], fileErr)
+		}
+
+		mergedNode, err = merge2.Merge(patchNode, mergedNode, kyaml.MergeOptions{ListIncreaseDirection: 1})
```
fykaa: `ListIncreaseDirection` is a magic number I feel should be documented/comment or use a named constant

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,131 @@
...
+			if cfg.IgnoreMissingFiles {
+				continue
+			}
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("error accessing file %s: %w", inFile, err)
+
+		}
+		filePaths = append(filePaths, inFile)
+
+	}
```
fykaa: if all files are missing and `IgnoreMissingFiles` is true, this will pass an empty slice to `yaml.MergeFiles` which will fail

fykaa: I think we should check if we have any files to merge after filtering

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,131 @@
...
+func (y *yamlMerger) convert(cfg promotion.Config) (builtin.YAMLMergeConfig, error) {
+	return validateAndConvert[builtin.YAMLMergeConfig](y.schemaLoader, cfg, stepKindYAMLMerge)
+}
+
+func (y *yamlMerger) run(
+	_ context.Context,
+	stepCtx *promotion.StepContext,
+	cfg builtin.YAMLMergeConfig,
+) (promotion.StepResult, error) {
+
```
fykaa:
```suggestion
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,131 @@
...
+	// Secure join the input paths to prevent path traversal attacks.
+	filePaths := []string{}
+	for _, path := range cfg.InFiles {
+		inFile, err := securejoin.SecureJoin(stepCtx.WorkDir, path)
+		if err != nil {
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("could not secure join input file %q: %w", path, err)
```
fykaa:
```suggestion
				fmt.Errorf("invalid input file path %q: %w", path, err)
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		// only add existing files
+		_, err = os.Stat(inFile)
+		if err != nil {
+			if cfg.IgnoreMissingFiles {
+				continue
+			}
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("error accessing file %s: %w", inFile, err)
```
fykaa:
```suggestion
				fmt.Errorf("input file not found: %s", path)
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		filePaths = append(filePaths, inFile)
+
+	}
+
+	// Secure join the output path to prevent path traversal attacks.
+	outFile, err := securejoin.SecureJoin(stepCtx.WorkDir, cfg.OutFile)
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("could not secure join outFile %q: %w", cfg.OutFile, err)
```
fykaa:
```suggestion
			fmt.Errorf("invalid output file path %q: %w", cfg.OutFile, err)
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	// ensure output path fully exist
```
fykaa:
```suggestion
	// ensure output path fully exists
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,131 @@
...
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("error creating directory structure %s: %w", filepath.Dir(outFile), err)
+	}
+
+	// Merge files
+	err = yaml.MergeFiles(filePaths, outFile)
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("could not merge YAML files: %w", err)
+	}
```
fykaa:
```suggestion
    if err = yaml.MergeFiles(filePaths, outFile); err != nil {
		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
			fmt.Errorf("could not merge YAML files: %w", err)
	}
```

pkg/promotion/runner/builtin/yaml_merger.go
prune998: OK.

I was in the impression that an empty list would be ok to fail during the merge, but you're right, if IgnoreMissingFiles is set, it should allow that case and terminate clean.

 I added.

```
	// simply stop here if list is empty and we ignore missing files
	if cfg.IgnoreMissingFiles && len(filePaths) == 0 {
		result := promotion.StepResult{Status: kargoapi.PromotionStepStatusSucceeded}
		result.Output = map[string]any{
			"commitMessage": "nothing to merge",
		}

		return result, nil
	}
```

so:

- if `IgnoreMissingFiles` is set, we stop with success on empty list
- if not, well, that will never happen because a missing file will be trapped in the for loop

prune998: I reworked it... I actually only  rune the merge if there are files in the list, else I go directly to the return message

prune998:
All requested modifications applied.
Thanks for the full review @fykaa

fykaa:
thanks @prune998 I'll do a final cleanup before approving this PR

fykaa (approved):
I have reviewed this PR and it works correctly for both cases - with and without `ignoreMissingFiles`.  in its current implementation, it will error when all files in `inFiles` are missing and there's nothing to merge.

Deferring to @krancour for the final merge.

prune998:
all set !! Many thanks @fykaa. Hope to see that merged soon ! Thanks all for you help

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
```diff
@@ -0,0 +1,106 @@
...
+  or overrides the value from the first document.
+:::
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `inFiles` | `[]string` | Y | Paths to a YAML files. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `outFile` | `string`   | Y | The path to the output file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `ignoreMissingFiles` | `bool` | N | When set to `true`, the directive will skip input files that does not exist. Defaults to `false`. |
```
krancour:
```suggestion
| `ignoreMissingFiles` | `bool` | N | When set to `true`, the directive will skip input files that do not exist. Defaults to `false`. |
```

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,135 @@
...
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("invalid input file path %q: %w", path, err)
+		}
+
+		if _, err = os.Stat(inFile); err != nil {
+			if os.IsNotExist(err) && cfg.IgnoreMissingFiles {
+				continue
+			}
+			return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+				fmt.Errorf("input file %q not found: %w", path, err)
```
krancour: It doesn't appear that this is the only possible cause of the error at this point, although I'll concede it's the likely cause. I think the entire conditional beginning on line +74 needs to be restructed something like this:

```
f _, err = os.Stat(inFile); err != nil {
	if os.IsNotExist(err) {
		cfg.IgnoreMissingFiles {
			continue
		}
		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
			fmt.Errorf("input file %q not found: %w", path, err)
	}
	return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
		fmt.Errorf("unexpected error with file %q: %w", path, err)
}
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
krancour: I also wonder if we need to call out here that at least one file _does_ need to be found, because the behavior seems to be that if none are found, it's an error...

But we also might want to reconsider the that behavior. If someone has set `ignoreMissingFiles` and provides a list of `n` files, do we think they expect it to ignore only _actually_ ignore `n - 1` missing files?

Do we want to consider changing this behavior to just return an empty document without failing?

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		filePaths = append(filePaths, inFile)
+	}
+
+	// Require at least one input file to merge
```
krancour: See https://github.com/akuity/kargo/pull/4634/files#r2441164554

This behavior feels slightly surprising.

pkg/promotion/runner/builtin/yaml_merger.go
```diff
@@ -0,0 +1,135 @@
...
+	outFile, err := securejoin.SecureJoin(stepCtx.WorkDir, cfg.OutFile)
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("invalid output file path %q: %w", cfg.OutFile, err)
+	}
+
+	// Ensure output directory exists
+	if err = os.MkdirAll(filepath.Dir(outFile), 0o700); err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("error creating directory %q: %w", filepath.Dir(outFile), err)
```
krancour: Nit: Since the directory very well may exist already, I would have this say there was an error "ensuring the existence" of the directory instead of saying it was an error creating it.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
krancour: I'll add, I understand that you need a starting point to succeed at merging these files, but if we have no files, we don't _need_ to merge. An empty document seems like a sensible result.

fykaa: returning Empty file seems sensible to me when `ignoreMissingFiles` is true
If `ignoreMissingFiles=false`: error ("no input files found to merge")
If `ignoreMissingFiles=true`: create an empty output file instead of erroring

prune998: I agree to the last statement.
Usually, you will try to merge many files that may not exist, like `base/values.yaml`, `<env>/values.yaml`, `<env>/<cluster>/values.yaml`... if there's no specific config, only the chart's `values.yaml` will be used, so you end up generating an empty "override" value file. 

I think this is the behaviour that is implemented right now, right ? So nothing to do here, except maybe being more descriptive in the docs ?

fykaa: > except maybe being more descriptive in the docs ?

the docs are updated to mention this along with the flag information. I believe that explains most of the things

prune998:
are we good to merge then ?

krancour (approved):

prune998:
@krancour any chance this will be released at some point ? It's been merged for quite some time.
thx !

PR #4636, opened 2025-07-17
feat: derive image creation time from annotations and labels

### LINKED ISSUE #4635: Incorrect image creation date can lead to wrong "NewestBuild" image selection
# Checklist

# Description

When Kargo selects the newest image from a container repository, it relies on the image's creation timestamp. However, due to a known Docker behavior, the creation timestamp embedded in an image's configuration (config.Created) may not reflect the actual build time.
As described in the Moby issue moby/moby#34786, if a docker build command produces an image whose layers are identical to an existing image, Docker will reuse the cached layers, including the original Created timestamp from the cached image config. This can cause confusion in CI/CD pipelines where images are sometimes rebuilt using cache and pushed using a new tag (tipically using the commit hash). A newly pushed image might be ignored by Kargo in favor of an older one simply because the build process resulted in no layer changes, and therefore the creation timestamp was not updated. Even worse is that this is non deterministic and sometimes it picks the latest image, sometimes it doesn't which generates even more confusion for everyone.

To address this, we should prioritize the official Open Containers Initiative (OCI) annotation for the image creation timestamp, which is designed to be a reliable source of truth, independent of Docker's caching behavior.
The [`org.opencontainers.image.created`](https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys) label is automatically set by modern container build tools and reflects the true build time.

I suggest that the fix should involve updating the image metadata retrieval logic to:
1. Check for the org.opencontainers.image.created label in the image's configuration.
2. If the label is present and contains a valid RFC3339 timestamp, use it as the definitive creation date.
3. If the label is not present or is invalid, fall back to the existing config.Created field to ensure backward compatibility.

This change should a more robust and accurate method for determining the newest image, aligning with modern container standards and preventing unexpected behavior in automated workflows.

# Screenshots

# Steps to Reproduce

1. Rebuild an existing docker image using a previous image cached locally and push a new tag
2. Kargo might pick up the newest build, or not.

Here's the diff of the docker images `config` retrieved using `crane` cli that shows this happening. The newer image has an older created date:
```diff
--- /Users/xx/repos/xx/staging-35039f5.json	2025-07-15 17:08:53
+++ /Users/xx/repos/xx/staging-c0ee80f.json	2025-07-15 17:08:50
@@ -18,18 +18,18 @@
     ],
     "WorkingDir": "/app/worker",
     "Labels": {
-      "org.opencontainers.image.created": "2025-07-15T14:00:29.218Z",
+      "org.opencontainers.image.created": "2025-07-15T15:23:30.730Z",
       "org.opencontainers.image.licenses": "",
-      "org.opencontainers.image.revision": "35039f5caab36de7fecd1de69f26629e8da66ef6",
+      "org.opencontainers.image.revision": "c0ee80ff2e24186fd85ea72d0a6fb33454710d2c",
       "org.opencontainers.image.source": "https://github.com/xx/xx",
       "org.opencontainers.image.title": "xx",
       "org.opencontainers.image.url": "https://github.com/xx/xx",
-      "org.opencontainers.image.version": "staging-35039f5"
+      "org.opencontainers.image.version": "staging-c0ee80f"
     },
     "ArgsEscaped": true
   },
-  "created": "2025-07-15T13:53:47.257942839Z",
+  "created": "2025-07-15T13:02:33.484606296Z",
   "history": [
     {
       "created": "2024-12-05T13:16:16Z",
@@ -141,23 +141,23 @@
       "comment": "buildkit.dockerfile.v0"
     },
     {
-      "created": "2025-07-15T13:53:46.355439647Z",
+      "created": "2025-07-15T13:02:29.045778917Z",
       "created_by": "COPY --chown=node:node /app/ . # buildkit",
       "comment": "buildkit.dockerfile.v0"
     },
     {
-      "created": "2025-07-15T13:53:46.355439647Z",
+      "created": "2025-07-15T13:02:29.045778917Z",
       "created_by": "USER node",
       "comment": "buildkit.dockerfile.v0",
       "empty_layer": true
     },
     {
-      "created": "2025-07-15T13:53:47.257942839Z",
+      "created": "2025-07-15T13:02:33.484606296Z",
       "created_by": "WORKDIR /app/worker",
       "comment": "buildkit.dockerfile.v0"
     },
     {
-      "created": "2025-07-15T13:53:47.257942839Z",
+      "created": "2025-07-15T13:02:33.484606296Z",
       "created_by": "CMD [\"node\" \"--import=./dist/instrumentation.js\" \"--enable-source-maps\" \"--dns-result-order=ipv4first\" \"dist/worker.js\"]",
       "comment": "buildkit.dockerfile.v0",
       "empty_layer": true
@@ -178,7 +178,7 @@
       "sha256:2f8a98d500e67efad71d8efd184796f3ab4ce82cb3d58c4cda6bba18061606ec",
       "sha256:e4999a05086398766b51c03ffedf9adc07be64f0443478d576db1385db5257a1",
       "sha256:a66adb5805b2857de72334bd70a3cbc69320c8170ec5ab30d0da2c170e99ca39",
-      "sha256:ffaf858f6036d81e30ff486c349bec0200b428d50278ad4491992f27c51acce9",
+      "sha256:23603882ae732bc706eb04bcf85683a160e582f2bfc7f0742ebbc7c5093c59e2",
       "sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef"
     ]
   }

``` 

# Version

```
1.6.1
```

# Logs

```
N/A
```

bmbferreira:
Fixes #4635 

Added functionality to retrieve the image creation time from the OCI annotation "org.opencontainers.image.created" if present in the image labels. Updated the test suite to validate this behavior.

### DIFF
--- a/internal/image/repository_client.go
+++ b/internal/image/repository_client.go
@@ -444,36 +444,35 @@ func (r *repositoryClient) getImageFromV1Image(
 	}
 
 	return &image{
-		Digest:      digest,
-		CreatedAt:   getCreationTime(cfg, manifest.Annotations),
+		Digest: digest,
+		CreatedAt: getCreationTime(
+			[]map[string]string{
+				manifest.Annotations,
+				cfg.Config.Labels,
+			},
+			&cfg.Created.Time,
+		),
 		Annotations: manifest.Annotations,
 	}, nil
 }
 
-func getCreationTime(cfg *v1.ConfigFile, annotations map[string]string) *time.Time {
-	// Define the order of precedence: annotations first, then labels
-	sources := []struct {
-		data map[string]string
-		keys []string
-	}{
-		{annotations, []string{ociCreatedAnnotation, legacyBuildDateAnnotation}},
-		{cfg.Config.Labels, []string{ociCreatedAnnotation, legacyBuildDateAnnotation}},
-	}
+func getCreationTime(sources []map[string]string, fallback *time.Time) *time.Time {
+	keys := []string{ociCreatedAnnotation, legacyBuildDateAnnotation}
 
 	for _, source := range sources {
-		if source.data != nil {
-			for _, key := range source.keys {
-				if createdStr, ok := source.data[key]; ok {
-					if created, err := time.Parse(time.RFC3339, createdStr); err == nil {
-						return &created
-					}
+		if source == nil {
+			continue
+		}
+		for _, key := range keys {
+			if createdStr, ok := source[key]; ok {
+				if created, err := time.Parse(time.RFC3339, createdStr); err == nil {
+					return &created
 				}
 			}
 		}
 	}
 
-	// Fall back to the config's Created field
-	return &cfg.Created.Time
+	return fallback
 }
 
 // rateLimitedRoundTripper is a rate limited implementation of


internal/image/repository_client.go
```diff
@@ -437,11 +441,22 @@ func (r *repositoryClient) getImageFromV1Image(
...
 	return &Image{
 		Digest:      digest,
-		CreatedAt:   &cfg.Created.Time,
+		CreatedAt:   getCreationTime(cfg),
 		Annotations: manifest.Annotations,
 	}, nil
 }
 
+func getCreationTime(cfg *v1.ConfigFile) *time.Time {
+	if cfg.Config.Labels != nil {
```
hiddeco: I think that extracting this from the labels is (slightly) wrong. The https://specs.opencontainers.org/image-spec/annotations/ refers to _annotations_, which can be found in the annotations field in the [Image Manifest](https://specs.opencontainers.org/image-spec/manifest/?v=v1.0.1#image-manifest-property-descriptions) and [Content Descriptor](https://specs.opencontainers.org/image-spec/descriptor/?v=v1.0.1#properties).

The labels would typically contain [Label Schema](https://label-schema.org) labels, as also referred to by the [annotations spec](https://specs.opencontainers.org/image-spec/annotations/#back-compatibility-with-label-schema).

To solve this, I think it would be best to create a function that looks for `org.opencontainers.image.created` and `org.label-schema.build-date` in both the annotations and the labels (in that order), while returning on the first match it finds (or by falling back to `cfg.Created.Time` if it does not find anything).

bmbferreira: Thanks a lot for the feedback @hiddeco and for the quick review. I implemented your suggestion which makes total sense IMO. I also added a few more test cases, please let me know if anything else is needed.

internal/image/repository_client.go
```diff
@@ -437,11 +445,37 @@ func (r *repositoryClient) getImageFromV1Image(
...
+						return &created
+					}
+				}
+			}
+		}
+	}
+
+	// Fall back to the config's Created field
+	return &cfg.Created.Time
+}
```
hiddeco: It might be a bit cleaner to do something like this:

```suggestion
func getCreationTime(sources []map[string]string, fallback *time.Time) *time.Time {
	keys := []string{ociCreatedAnnotation, legacyBuildDateAnnotation}
	
	for _, source := range sources {
		if source == nil {
			continue
		}
		for _, key := range keys {
			if createdStr, ok := source[key]; ok {
				if created, err := time.Parse(time.RFC3339, createdStr); err == nil {
					return &created
				}
			}
		}
	}

	return fallback
}
```

bmbferreira: done! Thank you

hiddeco (commented):
Thanks for following up with the comments @bmbferreira. I want to give this some manual testing later this week, but the changes themselves look good to me for now. 🙌

bmbferreira:
> Thanks for following up with the comments @bmbferreira. I want to give this some manual testing later this week, but the changes themselves look good to me for now. 🙌

ok! thank you so much for checking this so quickly. please let me know if anything else is needed 🙇‍♂️

koushyk:
@bmbferreira this change breaks imageSelectionStrategy: NewestBuild when org.opencontainers.image when created label comes with a base image from the FROM of the Dockerfile, and is not overridden by the build pipeline.

bmbferreira:
> @bmbferreira this change breaks imageSelectionStrategy: NewestBuild when org.opencontainers.image when created label comes with a base image from the FROM of the Dockerfile, and is not overridden by the build pipeline. 

I think it’s fair to say it’s not any more broken than it was before. The fix is easy.

PR #4641, opened 2025-07-18
feat: Allow to optionally add initContainers to webhooks server pods

ilyakazm:
# Add Support for Init Containers in Webhooks Server Deployment

## Context
When running Kargo controller as a shard on another cluster, the webhooks server instance must also be deployed alongside the controller and be able to make a "call home" to the management controller (similar to how the shard operates). This requires the webhooks server to have kubeconfig access to authenticate with the management cluster.

In our specific use case with GKE, we use the `gke-auth-plugin` to generate credentials dynamically. To make this plugin available to the webhooks server pod, we need a way to set up the necessary authentication components before the main container starts. Init containers provide the perfect solution for this setup phase.

While this addresses our specific GKE authentication scenario, the init container support could be valuable for other use cases where pre-initialization steps are needed for the webhooks server. I don't know if there are potential use cases for having initContainers on other Kargo components so I kept changes scoped to Webhooks Server only. Hope that's ok.

## Changes
Added conditional support for init containers in the webhooks server deployment template
The init containers are only included when `Values.webhooksServer.initContainers` is defined
Maintains backward compatibility - no changes for existing deployments that don't specify init containers

## Usage
Users can now specify init containers in their values file:
```yaml
webhooksServer:
  initContainers:
    - name: setup-auth
      image: my-setup-image
      # ... other init container configuration
```

This change enables more flexible deployment scenarios where the webhooks server needs additional setup before starting, particularly for complex authentication scenarios in multi-cluster environments.


krancour:
While I'm not specifically opposed to the idea of arbitrary initContainers, I'm also not inherently in support of it until I can understand the motivation a little better:

> When running Kargo controller as a shard on another cluster, the webhooks server instance must also be deployed alongside the controller and be able to make a "call home" to the management controller (similar to how the shard operates). This requires the webhooks server to have kubeconfig access to authenticate with the management cluster.

This is not true. Lack of comprehensive coverage of sharding is a known shortcoming of documentation at present, so any misunderstanding in this matter is understandable.

The only component to interact with the webhooks server is the Kubernetes API server of the Kargo control plane's underlying cluster. The webhooks server should be deployed only as part of the Kargo control plane.

I'm not sure if there is any confusion between the "webhooks server" and the "external webhooks server." The former serves Kargo-specific Kubernetes mutating/defaulting and validating/admission control webhooks. The latter is the more recently introduced server that accepts webhook requests from a variety of external platforms. Although, _both_ of these components should only be deployed as part of the control plane.

Can you help me to better understand what it is you're trying to do? Or is this only a matter of misundertstanding in terms of the webhook server's status as a control plane-only component?

ilyakazm:
Hi @krancour, thanks a lot for reviewing this PR!

When I originally opened it (about a month ago), I was actively working on deploying shards across environments. At the time, I couldn’t get them to function as expected — largely due to a misunderstanding on my part about how the architecture should work.

While digging into available documentation, I came across some references to the webhook server, which led me to believe it was a required component alongside the shard controller. Because all Kargo CRs in our setup live in a remote cluster where the control plane runs, I configured the webhook server with a kubeconfig to authenticate with that cluster — using an init container to install the GKE auth plugin so it could access and operate on those CRs. I created this PR thinking that others might need a similar setup.

Your explanation helped me realize that this assumption was incorrect. I’ve since disabled the webhook server on all clusters where shards are deployed and confirmed everything works fine with just the shard running — no webhook server needed.

So in short: I misread the situation and went down the wrong path, which led to this PR. If you think it doesn’t add value, feel free to close or reject it — no hard feelings on my end!

For additional context, here’s what our multi-cluster setup looks like:

- We have three isolated environments: QA, Staging, and Production.
- Each has an “infra” cluster where we run all operational tooling (ArgoCD, cert-manager, monitoring, etc.), including Kargo.
- The control plane runs in the Production “infra” cluster, and all Kargo CRs (Project, Stage, Warehouse, etc.) are deployed there.
- We run one shard per non-prod environment (QA and Staging), colocated with the respective Argo Rollouts shard.
- The primary reason for using shards is post-promotion verification — these are environment-specific and need to run in the same environment as the services they test. So each promotion stage is assigned to a shard running in the corresponding env’s infra cluster, ensuring verification jobs have the required network access to those services.

Thanks again for your time and the clarification — it helped a lot!

krancour:
You're doing things exactly correctly. One centralized Kargo control plane with a Kargo controller (shard) running "alongside" each Argo CD control plane. 👍 

> So in short: I misread the situation and went down the wrong path, which led to this PR.

No worries at all. Even mistakes are useful because eventually someone will happen upon this thread and learn something from it.

> If you think it doesn’t add value, feel free to close or reject it — no hard feelings on my end!

I'm going to close it _for now_. I'm increasingly reluctant to merge anything unless there's a clearly recognized need for it -- otherwise it's just more to maintain and support and another opportunity for someone to configure something incorrectly. I can see it being within the realm of possibility that a proven need for arbitrary init containers on any Kargo component might present itself in the future, in which case I'll re-open this.

Thanks!

PR #4642, opened 2025-07-18
docs: `Cluster` -> `ClusterConfig`

voidlily:
this doc previously mentioned a `Cluster` crd, but it's actually `ClusterConfig`


hiddeco (approved):

hiddeco (approved):
Thank you @voidlily 🙇

PR #4662, opened 2025-07-22
feat: add `http-download` step

hiddeco:
Related to #1610


krancour:
@fykaa could you review this please?

internal/promotion/runner/builtin/http_downloader.go
```diff
@@ -0,0 +1,318 @@
...
+	ctx context.Context,
+	stepCtx *promotion.StepContext,
+	cfg builtin.HTTPDownloadConfig,
+) (promotion.StepResult, error) {
+	absOutPath, err := d.prepareOutputPath(stepCtx.WorkDir, cfg.OutPath, cfg.AllowOverwrite)
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored}, err
+	}
+
+	resp, err := d.performHTTPRequest(cfg)
```
fykaa, line 83: i feel adding a log statement here to indicate that the download is starting and from which URL can help

internal/promotion/runner/builtin/http_downloader.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	client, err := d.buildHTTPClient(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("error creating HTTP client: %w", err)
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("error sending HTTP request: %w", err)
```
fykaa, line 148: this might make debugging easier
```suggestion
		return nil, fmt.Errorf("error sending HTTP request to %q: %w", cfg.URL, err)
```

internal/promotion/runner/builtin/http_downloader.go
```diff
@@ -0,0 +1,318 @@
...
+func (d *httpDownloader) checkFileOverwrite(absOutPath, outPath string, allowOverwrite bool) error {
+	if !allowOverwrite {
+		if _, err := os.Stat(absOutPath); err == nil || !os.IsNotExist(err) {
+			if err != nil {
+				return fmt.Errorf("error checking destination file: %w", err)
+			}
+			return &promotion.TerminalError{
+				Err: fmt.Errorf("file already exists at %s and overwrite is not allowed", outPath),
+			}
+		}
```
fykaa, line 129: this is slightly awkward logic since `err == nil || !os.IsNotExist(err)` is trying to capture both "file exists" and "error" cases, then checking `err != nil` again inside the block, which can be confusing and redundant imho.



```suggestion
		if _, err := os.Stat(absOutPath); err == nil {
		    // file exists
		    return &promotion.TerminalError{
				Err: fmt.Errorf("file already exists at %s and overwrite is not allowed", outPath),
			}
                 } else if !os.IsNotExist(err) {
                     // stat error
                     return fmt.Errorf("error checking destination file: %w", err)			
		}
```

internal/promotion/runner/builtin/http_downloader.go
```diff
@@ -0,0 +1,318 @@
...
+	return &http.Client{
+		Transport: httpTransport,
+		Timeout:   timeout,
+	}, nil
+}
+
+// validateResponse checks the HTTP response status and content length.
+func (d *httpDownloader) validateResponse(resp *http.Response) error {
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("HTTP request failed with status %d", resp.StatusCode)
```
fykaa, line 204:
```suggestion
		return fmt.Errorf("HTTP request failed: %s (%d)", resp.Status, resp.StatusCode)
```

internal/promotion/runner/builtin/http_downloader.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	// Check file size limit using Content-Length header if available
```
fykaa, line 207: should we consider explicitly handling `ContentLength == -1` (unknown).
right now it silently skips the size check, relying only on io.LimitReader

internal/promotion/runner/builtin/http_downloader.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// copyResponseToFile copies response content to the file with size limits and
+// context cancellation.
+func (d *httpDownloader) copyResponseToFile(ctx context.Context, resp *http.Response, f *os.File) error {
+	limitedReader := io.LimitReader(resp.Body, maxDownloadSize)
+
+	// Stream data with context cancellation support
+	var bytesDownloaded int64
+	buf := make([]byte, downloadBufferSize)
```
fykaa, line 270: non blocking: if this ever becomes performance-sensitive (like repeated downloads) could we consider using a `sync.Pool` for the buffer

internal/promotion/runner/builtin/http_downloader.go
hiddeco: `net/http` will typically return [a `url.Error`](https://pkg.go.dev/net/url#Error) which already contains this information.

internal/promotion/runner/builtin/http_downloader.go
hiddeco: `resp.Status` and `resp.StatusCode` pretty much contain the same information, except that `StatusCode` is the shorter version, e.g. `200 OK` versus `200`. Given this, and the fact that people should typically be either aware or can look up the status code to get the in-depth explanation, what would this add?

internal/promotion/runner/builtin/http_downloader.go
hiddeco: What would "explicitly handling" do? If the content length is unknown, that typically means the server does not know upfront and, for example, streams you whatever it has. This is completely fine, as long as it does not go outside the boundaries we set with `io.LimitReader`.

internal/promotion/runner/builtin/http_downloader.go
krancour: I won't deny the clarity of the suggestion, although it won't pass the linter without some further tweaks.

But I also don't think the original way it was written is unclear. In this scenario, "file exists" _is_ an error condition, so all the error conditions neatly fall into one block.

I'd probably leave this be.

hiddeco:
@fykaa can you please let me merge my own pull requests that are features? I still wanted to add the logging.

internal/promotion/runner/builtin/http_downloader.go
fykaa: this PR addresses this https://github.com/akuity/kargo/pull/4728/commits/cadab2d11aefeac351751bc7be00bd400944ea3a

fykaa:
@hiddeco got it, sorry about merging it earlier! I thought it was good to go, but I totally understand now

PR #4671, opened 2025-07-22
refactor(controller): comprehensive refactoring of aritfact selection logic for broader reuse outside discovery

krancour:
#4560 aims to do some preliminary check of whether a Warehouse would likely find new artifacts if it were to run a full discovery cycle (which can be expensive) in order to avoid unnecessary discovery cycles. (Example: An image tagged `foobar` was pushed to some repo, but the Warehouse subscribed to that repo is interested _only_ in semantically tagged images.) This requires that some of the image/commit/chart selection logic be accessible from outside the discovery process.

This is a holistic refactor of code for selecting images/commits/charts so that each has its own dedicated package, a simple `Selector` interface with `MatchesTag()` (or in the case of commits, `MatchesRef()`) in addition to `Select()` (full discovery). These changes are entirely structural. Logic hasn't changed.

With this refactor, #4560 will be able to make better determinations about whether or not to refresh a Warehouse _without_ the need to replicate portions of the discovery logic in ways that would prove brittle over time.

A secondary benefit to this refactor is it establishes a much more uniform approach to discovery of different types of artifacts.

Note that image selection is also significantly simplified. The package previously relied more heavily on internal types that were nearly 1:1 identical to types in the Kargo API. That had been done at a point in time with an eye toward the discovery logic being usable _beyond Kargo_, but I'm making a call that that juice wasn't worth the squeeze as we've yet to do such a thing and I'd prefer the simplicity going forward of relying on the Kargo API as a foundation that cross-cuts packages such as those containing these selectors.

cc @fuskovic once merged, this should radically simplify #4560.


internal/controller/git/commit/common.go
```diff
@@ -0,0 +1,63 @@
...
+		// If we reach here, the path is included and not excluded
+		return true
+	}
+	// None of the paths match our criteria
+	return false
+}
+
+func shortenString(str string, length int) string {
+	if length >= 0 && len(str) > length {
+		return str[:length] + "..."
```
hiddeco: I would expect this to be 

```suggestion
		return str[:length-3] + "..."
```

to ensure we end up with `length`.

(I also realize this might be a copy of my own code).

internal/controller/git/commit/lexical_selector.go
```diff
@@ -0,0 +1,86 @@
...
+		l.getLoggerContext(),
+		"selectionStrategy", kargoapi.CommitSelectionStrategyLexical,
+	)
+	ctx = logging.ContextWithLogger(ctx, logger)
+
+	repo, err := l.clone(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// Sometimes tests return nil repos.
```
hiddeco: Then arguably the tests should be changed.

internal/controller/git/commit/lexical_selector_test.go
```diff
@@ -0,0 +1,388 @@
...
+						return nil, errors.New("something went wrong")
+					},
+				},
+			},
+			assertions: func(t *testing.T, _ []kargoapi.DiscoveredCommit, err error) {
+				require.ErrorContains(t, err, "something went wrong")
+			},
+		},
+		{
+			name: "error filtering tags by diff paths",
```
fuskovic: Looks like you accidentally added this test twice.

internal/controller/git/commit/lexical_selector.go
fuskovic: I think the tests are intentionally returning nil where the clone func is a no-op. Since the `Close` [signature](https://github.com/akuity/kargo/blob/main/internal/controller/git/repo.go#L166) hangs off a pointer receiver you could just move the nil-check there and then safely call `defer repo.Close`.

fuskovic: e.g.

```go
func (r *repo) Close() error {
	if r == nil {
		return nil
	}
	return os.RemoveAll(r.homeDir)
}
```

fuskovic:
The commented out tests, did you mean to delete those since they've essentially been moved to another location?

internal/controller/git/commit/common.go
krancour: Yep... copied verbatim, but happy to fix.

internal/controller/git/commit/lexical_selector.go
krancour: @hiddeco existing problem, but happy to fix. @fuskovic I like your idea. That's the route I'm going. 🙏

krancour:
> The commented out tests, did you mean to delete those since they've essentially been moved to another location?

Correct... fixing.

internal/controller/git/commit/lexical_selector.go
krancour: Err... got too excited there. `repo` is of type `git.Repo`, which is an interface. Adding the suggestion to a specific implementation of that interface doesn't have the suggested effect.

krancour: The root of this is that we have no mockable implementation of `git.Repo`. Never bit the bullet on that because the interface has ~30 methods. Perfect busy work for Copilot. See cbf031040d2b3fd377678ab5d3e5d760a7d646c4.

Aside: There are probably dozens or even hundreds of other places we can make use of this.

93aad9666c382105760036098db6363939bef09b takes things the rest of the way and actually uses `git.MockRepo`.

hiddeco (approved):

PR #4672, opened 2025-07-22
feat(ui): add a new Tabs component to sync tab state with URL query param

rpelczar:


ui/src/features/project/pipelines/promotion/freight-details.tsx
```diff
@@ -1,10 +1,11 @@
 import { toJson } from '@bufbuild/protobuf';
```
Marvin9, line 1: There is also Promotion and its YAML tab above, should we URL that too?

rpelczar: I've seen this, and not sure if we should handle all tab sections. I added this to this section only, because there might be more tabs, as it is a part of extensions.

Marvin9 (approved):

PR #4674, opened 2025-07-23
fix(controller): health assessment fixes

### LINKED ISSUE #4673: bug: app health "cool down" period not enforced when desire revision(s) are not specified
### Discussed in https://github.com/akuity/kargo/discussions/4593

<sup>Originally posted by **AlanLonguet** July 10, 2025</sup>
I need a bit of help to determine if what we're observing is intended or not.

Our current promotion process revolves around committing the new version of an artifact in a repository that is directly watched by ArgoCD so it uses git-commit/git-push/argocd-update steps.

We've noticed that the argocd-update step only checks that the Application is correctly synced and does not wait for the full rollout of our pods which made us add verification steps that checks the rollout has happened.

We were wondering if that's intended behavior or an issue with our own setup.</div>

[Especially relevant comment](https://github.com/akuity/kargo/discussions/4593#discussioncomment-13856753):

> While working for a bit on #4595/#4608, I stumbled upon a more probable explanation for your issue @AlanLonguet.
>
>> My best guess is that the workaround that was applied to correct for https://github.com/akuity/kargo/issues/2196 is insufficient in your case.
>
> Apparently this "cool down" period is, mistakenly, not enforced when desired revision(s) are not specified by the `argocd-update` step. I would presume you have not specified any desired revision(s), as few users seem to do so.
>
> I'm converting this discussion to an issue, and I've got a fix ready to go.

krancour:
Fixes #4673 among other problems.

Per #4673, the post-op "cool down" period before App health status is deemed reliable was not being enforced in the (very common) case where desired revisions are not specified by the health check (because they were not specified by the argocd-update step that created the health check).

Apart from fixing that, this PR also fixes some other problems I've noted while working on this section of code:

1. We probably shouldn't count App health status as reliable if the current/last operation is in-progress or has any status other than Succeeded.

2. Pausing for the cool down period to elapse before deeming App health reliable is inadequate if we don't then repeat all checks we'd previously cleared to determine what new health problems may have arisen _during_ the cool down period. New error conditions could have been introduced. An entirely new operation may have started.

The changes in this PR amount to the following:

1. If the current/last operation is in any state other than Succeeded, App health is immaterial and Stage health is automatically Unknown.

2. If last operation Succeeded, but fewer than 10 seconds ago, App health is immaterial and Stage health is automatically Unknown.

3. No other aspect of App health is examined until/unless the last operation Succeeded more than 10 seconds prior.

4. "Cool down," no longer occurs synchronously within the health check logic.

5. The (regular) Stage reconciler now treats Unknown Stage heath as an error condition, prompting the Stage to be requeued for reconciliation while observing a progressive backoff. This allows "cool down" to happen naturally and without blocking other pending Stage reconciliations.

### DIFF
--- a/internal/health/checker/builtin/argocd.go
+++ b/internal/health/checker/builtin/argocd.go
@@ -19,6 +19,8 @@ import (
 
 const applicationStatusesKey = "applicationStatuses"
 
+var appHealthCooldownDuration = 10 * time.Second
+
 // ArgoCDHealthInput is the input for a health check associated with the the
 // argocd-update step.
 type ArgoCDHealthInput struct {
@@ -161,9 +163,9 @@ var healthErrorConditions = []argocd.ApplicationConditionType{
 
 // getApplicationHealth assesses the health of an Argo CD Application by looking
 // at its conditions, health status, and sync status. Based on these, it returns
-// an overall health state, the Argo CD Application's health status, and its sync
-// status. If it can not (fully) assess the health of the Argo CD Application, it
-// returns an error with a message explaining why.
+// an overall health state and the Argo CD Application's health status. If it
+// can not (fully) assess the health of the Argo CD Application, it returns an
+// error with a message explaining why.
 func (a *argocdChecker) getApplicationHealth(
 	ctx context.Context,
 	appKey client.ObjectKey,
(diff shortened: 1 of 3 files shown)


internal/health/checker/builtin/argocd_test.go
```diff
@@ -458,7 +458,7 @@ func Test_argocdUpdater_getApplicationHealth(t *testing.T) {
 				Namespace: testApp.Namespace,
 				Name:      testApp.Name,
 			},
-			[]string{"fake-version", "fake-commit", "another-fake-commit"},
+			nil,
```
krancour: This change to the test case demonstrably prevents the cool down period from being enforced, until the corresponding changes to the code under test are introduced.

internal/health/checker/builtin/argocd.go
```diff
@@ -218,49 +218,60 @@ func (a *argocdChecker) getApplicationHealth(
...
+	// since operation completion. Note: This workaround assumes no/minimal clock
+	// drift between the Kargo controller and the Argo CD Application controller.
+	//
+	// TODO(krancour): This workaround can be revisited if/when
+	// https://github.com/argoproj/argo-cd/pull/21120 is merged, as (for newer
+	// versions of Argo CD) it will allow us to accurately determine whether an
+	// Application's health was last assessed before or after the most recently
+	// completed operation.
+	if app.Status.OperationState != nil && !app.Status.OperationState.FinishedAt.IsZero() {
+		coolDownPeriod := 10 * time.Second
```
hiddeco: Is there a specific reason you made this code _less_ defensive? I.e., if `FinishedAt` is now zero, it appears like we will look at the health status anyway instead of applying a cooldown period?

hiddeco (approved):

Issue #4629, opened 2025-07-16
Requeued reconciliations don't check for shards on warehouses.


johnthompson-ybor:
I have a sharded setup, I had a shard on a warehouse spec, the shard successfully updated it.  I _removed_ the shard from the spec, and the control plane controller updated it, successfully, as I intended.  Several minutes later, the shard started trying to update it again.

I believe that the controller doesn't check shards again after a requeued reconciliation is triggered.

Rebooting the sharded controller seems to have stopped it from happening again, so it seems to only be about requeueing.

You may want to add a check at the beginning of the reconciliation loop to double check for the shard label and bail if it doesn't match any more.

krancour:
Nice catch. Thank you!

PR #4677, opened 2025-07-23
fix: check shard before warehouse + stage reconciles

fuskovic:
Closes: https://github.com/akuity/kargo/issues/4629


internal/controller/stages/regular_stages.go
```diff
@@ -322,6 +322,11 @@ func (r *RegularStageReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	if !inScope(r.cfg.Name(), stage) {
```
krancour: `r.cfg.Name()` won't give you what you're thinking it will...

For a controller configured for shard `foo`, it will return `stage-controller-foo`. Meanwhile, the label you'd be comparing to would, if present, have a value like, simply `foo`.

I think you need to just use `r.cfg.ShardName` directly.

That said, I would also do that without passing the shard name as an argument. If you make `inScope()` a method of the reconciler, it will be able to access `r.cfg.ShardName` itself without it needing to be passed around. In general, I would advocate for not passing around static configuration that never changes after initialization, which is the case here:

```suggestion
	if !r.inScope(stage) {
```

internal/controller/stages/regular_stages.go
```diff
@@ -1976,6 +1981,11 @@ func (r *RegularStageReconciler) clearAnalysisRuns(ctx context.Context, stage *k
 	return nil
 }
 
+func inScope(targetShard string, stage *kargoapi.Stage) bool {
+	shard, labeled := stage.GetLabels()[kargoapi.LabelKeyShard]
+	return !labeled || shard == targetShard
```
krancour: This logic looks like it might not be correct.

You're deeming a Stage in-scope if it lacks the shard label. If it lacks that label, it should actually be resolved only by the "default controller" (which is just the controller that has no shard name).

So the correct logic is simply this:

```
	return stage.GetLabels()[kargoapi.LabelKeyShard] == targetShard
```

Taking https://github.com/akuity/kargo/pull/4677/files#r2226596810 into account, the whole method becomes something like:

```
func (r *RegularStageReconciler) inScope(targetShard string, stage *kargoapi.Stage) bool {
	return stage.GetLabels()[kargoapi.LabelKeyShard] == r.cfg.ShardName
}
```

🤔 Simple as that is... maybe it's better to kill this func/method entirely and just use `if stage.GetLabels()[kargoapi.LabelKeyShard] != r.cfg.ShardName { ... }` where you would have called this.

internal/controller/stages/regular_stages_test.go
```diff
@@ -6577,3 +6578,104 @@ func Test_buildFreightSummary(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcile_ShardMatching(t *testing.T) {
```
krancour: It would probably be both more succinct and more clear if a "shard doesn't match" test case were added to the existing `TestRegularStageReconciler_Reconcile()` tests. It would logically fall after the "ignores control flow stage" case.

internal/controller/stages/regular_stages.go
```diff
@@ -322,6 +322,11 @@ func (r *RegularStageReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	if stage.GetLabels()[kargoapi.LabelKeyShard] != r.cfg.ShardName {
+		logger.Debug("ignoring Stage because it is not in scope")
```
krancour: I think this wording is more clear in meaning that scope. Applies in the Warehouse reconciler as well.

```suggestion
		logger.Debug("ignoring Stage because it is is not assigned to this shard")
```

krancour:
@fuskovic I've just realized that we should probably add this to the control flow Stages reconciler as well.

internal/controller/stages/regular_stages_test.go
```diff
@@ -78,6 +78,36 @@ func TestRegularStageReconciler_Reconcile(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "default",
+					Name:      "test-stage",
+					Labels: map[string]string{
+						kargoapi.LabelKeyShard: "wrong-shard",
+					},
+				},
+				Spec: kargoapi.StageSpec{
+					Shard: "correct-shard",
+					PromotionTemplate: &kargoapi.PromotionTemplate{
```
krancour, line 101:
```suggestion
					// Specify some minimal promotion process to get this Stage past the
					// logic that verifies this is not a control flow Stage.
					PromotionTemplate: &kargoapi.PromotionTemplate{
```

internal/controller/warehouses/warehouses_test.go
```diff
@@ -1002,3 +1006,97 @@ func TestShouldDiscoverArtifacts(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcile_ShardMatching(t *testing.T) {
```
krancour: I would make this more general since we don't yet have any other tests for the `Reconcile()` method. I'm not suggesting that you add test cases beyond these two at this time. I'm only suggesting that by naming this `TestReconcile()` instead, we're establishing a home for all the other tests that don't yet exist and might be added here one day.

```suggestion
func TestReconcile(t *testing.T) {
```

internal/controller/warehouses/warehouses_test.go
```diff
@@ -1002,3 +1006,97 @@ func TestShouldDiscoverArtifacts(t *testing.T) {
...
+									Labels: map[string]string{
+										kargoapi.LabelKeyShard: "wrong-shard",
+									},
+								},
+							},
+						).Build(),
+					cfg: ReconcilerConfig{ShardName: "right-shard"},
+					// Intentionally not setting any xFns because we should never reach them
+					// because we will exit before any reconciliation logic is executed.
+					// With that said, there is nothing that needs to be asserted here
```
krancour, line 1040: I think this is something we don't need to call out. I think it's a given that we don't mock out things we don't expect to be called.

internal/controller/warehouses/warehouses_test.go
```diff
@@ -1002,3 +1006,97 @@ func TestShouldDiscoverArtifacts(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			r := tt.reconciler()
+			logger := logging.NewLogger(logging.DebugLevel)
+			ctx := logging.ContextWithLogger(t.Context(), logger)
+			if tt.shouldMatch {
+				require.Panics(t, func() {
+					r.Reconcile(ctx, tt.req) //nolint:errcheck
+				})
+				return
+			}
```
krancour: This is could be a head scratcher for anyone who didn't read the comments starting on lines 1038 and 1071. I might suggest that the explanations offered there be moved down here...

But I think there's something even better we can potentially do here...

If you make an injectable `syncWarehouseFn`, your test cases can:

1. For non-match: Not bother mocking it out because it won't be called and then just assert the "zero" result.

1. For match: Mock out `syncWarehouseFn` and `patchStatusFn` with trivial implementations that do nothing and don't return an error, then assert a "non-zero" result.

If you do that, combined with [this](https://github.com/akuity/kargo/pull/4677/files#r2238149545), we'll be well on our way to more thorough tests for this method and we'll also have eliminated the head-scratching moment over the relation between matching and panicking.

internal/controller/stages/regular_stages.go
```diff
@@ -322,6 +322,11 @@ func (r *RegularStageReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	if stage.GetLabels()[kargoapi.LabelKeyShard] != r.cfg.ShardName {
+		logger.Debug("ignoring Stage because it is is not assigned to this shard")
+		return ctrl.Result{}, nil
+	}
```
krancour: As we discussed offline, #4708 changes the logic required here, but luckily also introduces a new predicate, which you can probably just use directly right here.

fwiw, I would probably make that predicate a field of the reconciler and initialize it just once when the reconciler is initialized.

krancour: https://github.com/akuity/kargo/blob/main/internal/controller/predicates.go

internal/controller/predicates.go
```diff
@@ -48,10 +48,10 @@ func (i ResponsibleFor[T]) Generic(e event.TypedGenericEvent[T]) bool {
 	if obj == nil {
 		return false
 	}
-	return i.amResponsible(obj)
+	return i.AmResponsible(obj)
 }
 
-func (i ResponsibleFor[T]) amResponsible(obj client.Object) bool {
+func (i ResponsibleFor[T]) AmResponsible(obj client.Object) bool {
```
hiddeco: If this becomes a public method, I feel the following would be better:

```suggestion
func (i ResponsibleFor[T]) IsResponsibleFor(obj client.Object) bool {
```

internal/controller/stages/control_flow_stages.go
```diff
@@ -33,6 +33,7 @@ type ControlFlowStageReconciler struct {
 	cfg           ReconcilerConfig
 	client        client.Client
 	eventRecorder record.EventRecorder
+	labelChecker  controller.ResponsibleFor[kargoapi.Stage]
```
krancour: I would name this more specifically is all:

```suggestion
	shardPredicate  controller.ResponsibleFor[kargoapi.Stage]
```

krancour: Everything else LGTM, so preemptively LGTM'ing!

krancour (approved):

internal/controller/warehouses/warehouses_test.go
```diff
@@ -1002,3 +1007,148 @@ func TestShouldDiscoverArtifacts(t *testing.T) {
...
+				NamespacedName: types.NamespacedName{
+					Name:      "test-warehouse",
+					Namespace: "test-namespace",
+				},
+			},
+			assertions: func(t *testing.T, r ctrl.Result, err error) {
+				require.NoError(t, err)
+				require.False(t, r.IsZero(), "expected further reconciliation after shard match")
+			},
+		},
```
krancour, line 1142: Since we only added the two test cases we really needed to cover the change, we might put a TODO comment here calling for more test cases.

krancour:
⛹️‍♂️ Slam dunk

PR #4679, opened 2025-07-24
feat: add `oci-download` step 

hiddeco:
Related to #1610

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-download.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-download.md
@@ -18,10 +18,10 @@ Downloads are limited to 100MB to prevent resource exhaustion.
 
 | Name | Type | Required | Description |
 |------|------|----------|-------------|
-| `imageRef` | `string` | Y | Reference to the OCI artifact to pull. Supports both tag format `registry/repository:tag` and digest format `registry/repository@sha256:digest`. |
+| `imageRef` | `string` | Y | Reference to the OCI artifact to download. Supports both tag format `registry/repository:tag` and digest format `registry/repository@sha256:digest`. For Helm OCI artifacts, the `oci://` prefix is supported (e.g., `oci://registry/repository:tag`) and will use Helm-specific credential lookup. |
 | `outPath` | `string` | Y | Path to the destination file where the extracted artifact will be saved. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
 | `allowOverwrite` | `boolean` | N | Whether to allow overwriting an existing file at the specified path. If `false` and the file exists, the download will fail. Defaults to `false`. |
-| `mediaType` | `string` | N | MediaType of the layer to pull. Selects the first layer matching this type. If not specified, selects the first layer available. |
+| `mediaType` | `string` | N | Media type of the layer to download. Selects the first layer matching this type. If not specified, selects the first layer available. |
 | `insecureSkipTLSVerify` | `boolean` | N | Whether to skip TLS verification when downloading the artifact. Defaults to `false`. |
 
 ## Examples
@@ -37,6 +37,21 @@ steps:
 - uses: oci-download
   config:
     imageRef: registry.example.com/charts/my-app:1.0.0
+    outPath: ./charts/my-app-1.0.0.tgz
+    mediaType: application/vnd.cncf.helm.chart.content.v1.tar+gzip
+```
+
+### Downloading a Helm Chart with OCI Protocol
+
+This example shows downloading a Helm chart using the `oci://` prefix, which
+ensures that [Helm-specific credentials](../../50-security/30-managing-credentials.md)
+are used for authentication.
+
+```yaml
+steps:
+- uses: oci-download
+  config:
+    imageRef: oci://registry.example.com/charts/my-app:1.0.0
     outPath: ./charts/my-app-1.0.0.tgz
     mediaType: application/vnd.cncf.helm.chart.content.v1.tar+gzip
 ```
@@ -93,14 +108,14 @@ directly with Stage-specific values before being committed to a Git repository.
 
 ```yaml
 vars:
-- name: manifestRepo
+- name: gitRepo
   value: https://github.com/example/manifests.git
-- name: chartRegistry
-  value: registry.example.com/charts/my-app
+- name: chart
+  value: oci://registry.example.com/charts/my-app
 steps:
 - uses: git-clone
   config:
-    repoURL: ${{ vars.manifestRepo }}
+    repoURL: ${{ vars.gitRepo }}
     checkout:
     - branch: stage/${{ ctx.stage }}
       create: true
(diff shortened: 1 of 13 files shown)


internal/io/fs/move.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"fmt"
+	"os"
+)
+
+// SimpleAtomicMove performs an atomic move operation from src to dst.
+// If the destination already exists, it removes it before attempting the move.
+// It returns an error if the move operation fails for any reason, including
+// if the source file does not exist or if the destination cannot be created.
+func SimpleAtomicMove(src, dst string) error {
```
hiddeco: Minimal copy because https://github.com/akuity/kargo/pull/4662 does the same (but includes tests, etc.)

krancour:
@fykaa if you could review this for @hiddeco, that would be a big help. 🙏

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-download.md
```diff
@@ -0,0 +1,144 @@
...
+:::
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `imageRef` | `string` | Y | Reference to the OCI artifact to download. Supports both tag format `registry/repository:tag` and digest format `registry/repository@sha256:digest`. For Helm OCI artifacts, the `oci://` prefix is supported (e.g., `oci://registry/repository:tag`) and will use Helm-specific credential lookup. |
+| `outPath` | `string` | Y | Path to the destination file where the extracted artifact will be saved. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `allowOverwrite` | `boolean` | N | Whether to allow overwriting an existing file at the specified path. If `false` and the file exists, the download will fail. Defaults to `false`. |
+| `mediaType` | `string` | N | Media type of the layer to download. Selects the first layer matching this type. If not specified, selects the first layer available. |
```
jessesuen, line 24: What is the consequence of omitting `mediaType` in the common use cases? The mediaType field is something I don't think users will know to supply correctly. So if this is needed to be specified correctly in the common use cases, I feel this would be a point of friction.

For example, when I compare this to `oras pull` (https://oras.land/docs/commands/oras_pull), I don't see mediatype as an option users need to worry about.

hiddeco: It usually is not required to specify this (and the first layer picked will be the right one), but it may be required if you e.g. are after some in-toto layer that is attached to an image. Or if you do something obnoxious where you bundle your manifests with your container image as a secondary layer.

jessesuen: Thanks for the explanation. I would suggest the docs be modified slightly to have this as an example of an abnormal case. When reading the docs, I came under the impression that it was needed in the common case. I suggest removing it from all examples except one, with an explanation of the scenario where it might be required.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-download.md
```diff
@@ -0,0 +1,144 @@
...
+saved to a local file. This is useful when you need to download charts that are
+distributed as OCI artifacts.
+
+```yaml
+steps:
+- uses: oci-download
+  config:
+    imageRef: registry.example.com/charts/my-app:1.0.0
+    outPath: ./charts/my-app-1.0.0.tgz
+    mediaType: application/vnd.cncf.helm.chart.content.v1.tar+gzip
```
jessesuen: The "Downloading a Helm Chart" examples showing mediaType is why I had concerns. People will copy these examples, and so we'll be promoting the use of this field unintentionally.

hiddeco: I have updated the examples.

internal/promotion/runner/builtin/oci_downloader.go
```diff
@@ -0,0 +1,365 @@
...
+// the configuration.
+func (d *ociDownloader) buildHTTPTransport(cfg builtin.OCIDownloadConfig) *http.Transport {
+	httpTransport := cleanhttp.DefaultTransport()
+	if cfg.InsecureSkipTLSVerify {
+		httpTransport.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: true, // nolint: gosec
+		}
+	}
+	return httpTransport
+}
```
fykaa: Nit: I think this order is more intuitive, since `resolveImage()` calls it this way.
```suggestion
// buildHTTPTransport creates a new HTTP transport with TLS settings based on
// the configuration.
func (d *ociDownloader) buildHTTPTransport(cfg builtin.OCIDownloadConfig) *http.Transport {
	httpTransport := cleanhttp.DefaultTransport()
	if cfg.InsecureSkipTLSVerify {
		httpTransport.TLSClientConfig = &tls.Config{
			InsecureSkipVerify: true, // nolint: gosec
		}
	}
	return httpTransport
}

// getAuthOption retrieves and configures authentication for the registry.
func (d *ociDownloader) getAuthOption(
	ctx context.Context,
	stepCtx *promotion.StepContext,
	ref name.Reference,
	credType credentials.Type,
) (remote.Option, error) {
	repoURL := ref.Context().String()

	// NB: Some credential database implementations expect the URL to be
	// prefixed with "oci://".
	if credType == credentials.TypeHelm {
		repoURL = "oci://" + repoURL
	}

	creds, err := d.credsDB.Get(ctx, stepCtx.Project, credType, repoURL)
	if err != nil {
		return nil, fmt.Errorf("error obtaining credentials for image repo %q: %w", repoURL, err)
	}

	if creds != nil && (creds.Username != "" || creds.Password != "") {
		return remote.WithAuth(&authn.Basic{
			Username: creds.Username,
			Password: creds.Password,
		}), nil
	}

	return nil, nil
}
```

internal/promotion/runner/builtin/oci_downloader.go
```diff
@@ -0,0 +1,365 @@
...
+			}
+		}
+		return fmt.Errorf("failed to copy layer content: %w", err)
+	}
+
+	return nil
+}
+
+// findTargetLayer finds the appropriate layer based on media type preference.
+func (d *ociDownloader) findTargetLayer(
```
fykaa: I think `findTargetLayer` method might fit well between `extractLayerToFile` and `writeLayerToFile` because `extractLayerToFile` is the high-level orchestration method that calls `findTargetLayer`

internal/promotion/runner/builtin/oci_downloader.go
```diff
@@ -0,0 +1,366 @@
...
+	targetLayer, err := d.findTargetLayer(img, manifest, mediaType)
+	if err != nil {
+		return fmt.Errorf("failed to find target layer: %w", err)
+	}
+
+	return d.writeLayerToFile(targetLayer, absOutPath)
+}
+
+
+// findTargetLayer finds the appropriate layer based on media type preference.
```
fykaa:
```suggestion
}

// findTargetLayer finds the appropriate layer based on media type preference.
```

krancour (approved):

PR #4682, opened 2025-07-25
fix(chart): use .Values.image.pullSecrets in garbage collector CronJob

BWagenerGenerali:
We noticed the Kargo garbage collector was failing with imagePullBackoff error.

The cause was that it's missing imagePullSecrets in the CronJob.

### DIFF
--- a/charts/kargo/templates/garbage-collector/cron-job.yaml
+++ b/charts/kargo/templates/garbage-collector/cron-job.yaml
@@ -52,6 +52,10 @@ spec:
           affinity:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+          {{- with .Values.image.pullSecrets }}
+          imagePullSecrets:
+            {{- toYaml . | nindent 8 }}
+          {{- end }}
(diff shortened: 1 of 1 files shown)


krancour:
There's no change in this PR?

Apart from that, why should any image pull secret be required to download the public Kargo image?

BWagenerGenerali:
Thank you for pointing that out! I pushed the desired changes again.

The use-case for this is usually some sort of regulatory or security related requirement. Users might use a pull-through registry or corporate artifact registries which require authentication.

A common pattern then is to override the image source as well as the required pull secrets:

```yaml
image:
  repository: my-orgs-artifact-registry.my-org.com/akuity/kargo
  pullSecrets:
    - name: my-pull-secret
```

Currently `.Values.image.pullSecrets` is being used in all workloads except in the garbage collection CronJob (e.g. https://github.com/akuity/kargo/blob/main/charts/kargo/templates/management-controller/deployment.yaml#L53-L55)

PR #4697, opened 2025-07-28
Watch the tls certs for changes update the served certs

### LINKED ISSUE #4617: Feature Request: Reload kargo-api TLS certificate when it is updated
# Checklist

# Proposed Feature
We are currently using Kargo v1.5.0 installed via the helm chart.  Our helm settings for SSL are as follows:
```yaml
  tls:
    selfSignedCert: false
  service:
    type: LoadBalancer
```

This creates a cert-manager certificate with 90 day validity.  The certificate is correctly refreshed, however when the secret is changed the new TLS cert isn't loaded into the kargo-api resulting in an expired certificate being presented.  Manually restarting the deployment results in the new certificate being presented.

The proposed feature is to watch the certificate file for updates and then reload the server cert.

# Motivation
Without this feature cluster operators would need to manually track the certificate expiry and then reload the api server themselves.  This is a not insignificant admin overhead, and can result in access to the service being disrupted if the certificate isn't rotated correctly.

dbason:
Fixes #4617

Watches the certificate file locations (in a manner that is compatible with how k8s links and mounts secrets).  Instead of passing the certificate files directly to the http server it uses GetCertificate instead, and updates the certificate when the files are changed.


krancour:
I'd hold off on doing this until what's already here as been reviewed / agreed on, but the external webhooks server is probably in need of this exact same feature.

dbason:
@krancour is there anything else I need to do to get approval for the CI to run?

dbason:
```
internal/server/server.go:253:21: G402: TLS MinVersion too low. (gosec)
			srv.TLSConfig = &tls.Config{
				GetCertificate: certLoader.GetCertificate,
			}
1 issues:
```
This hasn't actually functionally changed from what the code was doing previously but I think it's being picked up now because I'm explicitly setting the `TLSConfig`.  I can update this.

```
=== RUN   TestDirectoryWatcher
    directorywatcher_test.go:56: 
        	Error Trace:	/__w/kargo/kargo/internal/server/certwatcher/directorywatcher_test.go:56
        	Error:      	Should be false
        	Test:       	TestDirectoryWatcher
        	Messages:   	channel should be closed
--- FAIL: TestDirectoryWatcher (0.05s)
FAIL
coverage: 87.5% of statements
```
This passed locally when I was testing but might be flaky so I'll check on it.

krancour:
> is there anything else I need to do to get approval for the CI to run?

They ran. Wonder if @hiddeco authorized it maybe. 🤷‍♂️ 

> This passed locally when I was testing

It's a little bit slower, but `make hack-test-unit` (as opposed to `make test-unit`) should execute the tests in as close as we can get to the same environment they're executed in in CI.

dbason:
I've updated the minimum TLS version to be 1.3 and updated the flaking test to be more resilient.

dbason:
Is there anything else I can do to move this forward @krancour ?

dbason:
@krancour we ran into this again recently in our infrastructure.  I've rebased and updated the logging package but it would be great to get some idea how how you guys want to handle the certs and push it into the external webhook controller as well.

Issue #4562, opened 2025-07-08
event type registry_package is not supported


danielloader:
# Description

I've configured Github to send webhooks on new image pushes to GHCR to Kargo.

These webhooks are rejected as shown in the screenshot of the error, permissions of the app to create webhooks shown below that.

# Screenshots

<img width="1494" height="1010" alt="Image" src="https://github.com/user-attachments/assets/f7d7ee28-9691-4ade-bdcd-4b5f95d36ec7" />

<img width="1696" height="542" alt="Image" src="https://github.com/user-attachments/assets/180cd034-9bd7-49be-8b06-c27b17c1fac3" />

# Steps to Reproduce

Follow the [guide for webhooks for github](https://docs.kargo.io/user-guide/reference-docs/webhook-receivers/github/#webhooks-from-a-github-app) and enable registry package webhooks.

# Version

```
v1.6.0
```

# Logs

_No logs for this seen in the kargo-external-webhooks-server pod(s)._

hiddeco:
The `Registry Package` is indeed not supported, but I get why you selected this as the screenshot shows the wrong example. You should instead select the `Package` option, which we do support (and which is the [recommended](https://docs.github.com/en/webhooks/webhook-events-and-payloads#registry_package) event by GitHub).

@fuskovic any chance you can update the screenshot to not give the false impression to people about what they should select?

danielloader:
Thanks for the quick reply, but I am having some trouble here...

<img width="1500" height="2460" alt="Image" src="https://github.com/user-attachments/assets/cc2d8ec1-fd98-4eb2-bb1b-ff3aac463425" />

I feel blind here but which event is the correct one to assign?

hiddeco:
That's odd, as GitHub [lists them as available for GitHub Applications](https://docs.github.com/en/webhooks/webhook-events-and-payloads#package). Are you on github.com or on an (outdated?) GitHub Enterprise instance?

hiddeco:
What I do observe in the documentation though is this specific note:

> To install this event on a GitHub App, the app must have at least read-level access for the "Packages" repository permission.

Can you double check if this is the case?

danielloader:
I can confirm Packages is selected and Access: Read level.

hiddeco:
Based on [this discussion](https://github.com/orgs/community/discussions/24636), it appears GitHub is not being honest about their support within their own documentation.

[...]

PR #4699, opened 2025-07-28
[WIP] feat: support `registry_package` event for Kargo GitHub webhook

nitishfy:
Closes https://github.com/akuity/kargo/issues/4562


internal/webhook/external/github.go
```diff
@@ -137,44 +145,118 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
-			if ref, err = name.ParseReference(
-				pkg.GetPackageVersion().GetPackageURL(),
-			); err != nil {
+			pkgURL := pkg.GetPackageVersion().GetPackageURL()
+
+			// GitHub sometimes sends package URLs with a trailing colon
+			// and no tag (e.g., "ghcr.io/user/image:"). Such strings are
+			// not valid OCI image references and will cause name.ParseReference()
+			// to fail. We trim the trailing colon to avoid parsing errors
+			// and optionally allow fallback handling (e.g., ":latest").
```
nitishfy: Screenshot 2025-07-30 at 10 58 21 PM

Screenshot 2025-07-30 at 11 00 46 PM

internal/webhook/external/github.go
```diff
@@ -92,8 +95,9 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 
 		eventType := r.Header.Get(gh.EventTypeHeader)
 		switch eventType {
-		case githubEventTypePackage, githubEventTypePing, githubEventTypePush:
+		case githubEventTypePackage, githubEventTypePing, githubEventTypePush, githubEventTypeRegistryPackage:
 		default:
+			logger.WithValues("eventType", eventType).Info("unsupported event type")
```
krancour: Info is the default log level for Kargo. Is this really information that an operator viewing Kargo's logs needs to see? It's not at all indicative of anything having gone wrong within Kargo. There's nothing to act on. Isn't it just cluttering the logs with noise?

Meanwhile, whoever send the bad request has already received a response clearly indicating the request was bad and why.

I wouldn't log this at all.

krancour: Or _if_ were to, I'd do it at the debug level.

krancour: Applies in several other spots.

nitishfy: Those were added just for testing error last time. I will remove those instead.

internal/webhook/external/github.go
```diff
@@ -137,44 +145,118 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+			// not valid OCI image references and will cause name.ParseReference()
+			// to fail. We trim the trailing colon to avoid parsing errors
+			// and optionally allow fallback handling (e.g., ":latest").
+			if strings.HasSuffix(pkgURL, ":") {
+				logger.WithValues("pkgURL", pkgURL).Info("trimming invalid trailing colon")
+				// nolint:staticcheck
+				pkgURL = strings.TrimSuffix(pkgURL, ":") // nolint:staticcheck
+			}
+
+			if ref, err = name.ParseReference(pkgURL); err != nil {
```
krancour: What's the point of parsing this using something so low-level like `name.ParseReference()` when the normalization logic is going to take care of everything required to convert `pkgURL` to a canonical representation?

internal/webhook/external/github.go
```diff
@@ -137,44 +145,118 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 		var mediaType string
 
 		switch e := event.(type) {
-		case *gh.PackageEvent:
+		case *gh.RegistryPackageEvent:
```
krancour: These cases were alphabetized before. Could we stick with that please?

(Note: I am not at all saying that switch cases, or anything, must _always_ be arranged alphabetically. There's often a case for arranging things in order or precedence, importance, chronologically, etc. In many cases where there _isn't_ any other ordering that makes sense, alphabetical is a convenient default ordering to fall back to. So if you see something is already arranged alphabetically, it's best to stick to it.)

internal/webhook/external/github.go
```diff
@@ -197,7 +279,12 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 			"mediaType", mediaType,
 		)
 		ctx = logging.ContextWithLogger(ctx, logger)
+		for _, val := range repoURLs {
+			fmt.Println(val)
+		}
```
krancour: Did you mean to remove this?

internal/webhook/external/github.go
nitishfy: Please take a look now.

internal/webhook/external/github.go
```diff
@@ -92,7 +96,7 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 
 		eventType := r.Header.Get(gh.EventTypeHeader)
 		switch eventType {
-		case githubEventTypePackage, githubEventTypePing, githubEventTypePush:
+		case githubEventTypePackage, githubEventTypePing, githubEventTypePush, githubEventTypeRegistryPackage:
```
krancour: Just slightly more readable than the long line:

```suggestion
		case githubEventTypePackage,
			githubEventTypePing,
			githubEventTypePush,
			githubEventTypeRegistryPackage:
```

internal/webhook/external/github.go
nitishfy: ah yes, i didn't think that case. Removed the logic now.

internal/webhook/external/github.go
nitishfy: removed.

internal/webhook/external/github.go
nitishfy: done.

internal/webhook/external/github.go
```diff
@@ -190,6 +189,45 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+			}
+			pkgURL := pkg.GetPackageVersion().GetPackageURL()
+			// GitHub sometimes sends package URLs with a trailing colon
+			// and no tag (e.g., "ghcr.io/user/image:"). Such strings are
+			// not valid OCI image references and will cause name.ParseReference()
+			// to fail. We trim the trailing colon to avoid parsing errors
+			// and optionally allow fallback handling (e.g., ":latest").
+			if strings.HasSuffix(pkgURL, ":") {
+				pkgURL = strings.TrimSuffix(pkgURL, ":") + ":latest"
+			}
```
krancour:
```suggestion
			pkgURL := strings.TrimSuffix(pkg.GetPackageVersion().GetPackageURL(), ":")
```

internal/webhook/external/github.go
```diff
@@ -190,6 +189,45 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+				return
+			}
+			switch pkg.GetPackageType() {
+			// These are the only types of packages we care about.
+			case ghcrPackageTypeContainer, ghcrPackageTypeDocker:
+			default:
+				xhttp.WriteResponseJSON(w, http.StatusOK, nil)
+				return
+			}
+			pkgURL := pkg.GetPackageVersion().GetPackageURL()
```
krancour:
```suggestion
```

internal/webhook/external/github.go
```diff
@@ -3,11 +3,10 @@ package external
 import (
 	"errors"
 	"fmt"
+	gh "github.com/google/go-github/v74/github"
 	"net/http"
-
-	"github.com/google/go-containerregistry/pkg/name"
-	gh "github.com/google/go-github/v71/github"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
```
krancour: This isn't gci.

internal/webhook/external/github.go
nitishfy: removed this entirely.

internal/webhook/external/media_types.go
```diff
@@ -81,6 +81,13 @@ var probableContainerImageMediaTypes = []string{
 // repository and as if it represented a Helm chart repository. It is
 // exclusively the caller's responsibility to deal with this possibility.
 func getNormalizedImageRepoURLs(repoURL, mediaType string) []string {
+	// Some GitHub package URLs may incorrectly include a trailing colon
+	// (e.g., "ghcr.io/user/image:"), which is not a valid OCI reference.
+	// Append ":latest" to ensure the reference is parsable.
+	if strings.HasSuffix(repoURL, ":") {
```
nitishfy: moved the logic of parsing the reference in the normalization function itself. This removes the requirement of having `pkgURL` and checking it's trailing colon since all this is getting done inside the normalization logic now. Lastly, added the test too to verify this.

internal/webhook/external/github.go
```diff
@@ -160,17 +165,11 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
-			); err != nil {
-				xhttp.WriteErrorJSON(w, err)
-				return
-			}
+			pkgURL := pkg.GetPackageVersion().GetPackageURL()
 			manifest := pkg.GetPackageVersion().GetContainerMetadata().GetManifest()
 			if cfg, ok := manifest["config"].(map[string]any); ok {
 				if mediaType, ok = cfg["media_type"].(string); ok {
-					repoURLs = getNormalizedImageRepoURLs(ref.Context().Name(), mediaType)
+					repoURLs = getNormalizedImageRepoURLs(pkgURL, mediaType)
```
krancour:
```suggestion
					repoURLs = getNormalizedImageRepoURLs(
						pkg.GetPackageVersion().GetPackageURL(),
						mediaType,
					)
```

internal/webhook/external/github.go
```diff
@@ -160,17 +165,11 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
 				return
 			}
-			var ref name.Reference
-			if ref, err = name.ParseReference(
-				pkg.GetPackageVersion().GetPackageURL(),
-			); err != nil {
-				xhttp.WriteErrorJSON(w, err)
-				return
-			}
+			pkgURL := pkg.GetPackageVersion().GetPackageURL()
```
krancour:
```suggestion
```

internal/webhook/external/media_types.go
```diff
@@ -81,6 +81,13 @@ var probableContainerImageMediaTypes = []string{
...
+		repoURL = strings.TrimSuffix(repoURL, ":") + ":latest"
+	}
+
```
krancour: Not crazy about this being _here_. This function was not GitHub-specific before and now it is. Even where GitHub is concerned, this seems only to be a concern for one specific event type. It was better the way it was before where you were dealing with it in the one specific place where it was relevant.

internal/webhook/external/github.go
```diff
@@ -160,17 +165,10 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
-				pkg.GetPackageVersion().GetPackageURL(),
-			); err != nil {
-				xhttp.WriteErrorJSON(w, err)
-				return
-			}
 			manifest := pkg.GetPackageVersion().GetContainerMetadata().GetManifest()
 			if cfg, ok := manifest["config"].(map[string]any); ok {
 				if mediaType, ok = cfg["media_type"].(string); ok {
-					repoURLs = getNormalizedImageRepoURLs(ref.Context().Name(), mediaType)
+					repoURLs = getNormalizedImageRepoURLs(pkg.GetPackageVersion().GetPackageURL(), mediaType)
```
krancour:
```suggestion
					repoURLs = getNormalizedImageRepoURLs(
						pkg.GetPackageVersion().GetPackageURL(),
						mediaType,
					)
```

internal/webhook/external/media_types.go
nitishfy: There's no harm putting this here. If we ever plan to support a webhook support for some other registry (i know we support most of them now) and if the problem was as specific as GitHub (though the chances are less), we'll have to make sure that we repeat ourselves in each switch case.

krancour: The thing is this function was _expected_ to fail when handed invalid input. Now you're making the function _correct_ invalid input that occurs in one very specific scenario. Does it work? Yes. But it's not very clean.

hiddeco: This is an abnormality and I would introduce this to the place where you know this abnormality occurs as well. The chance exist otherwise that unexpected behavior is introduced where you have a different intent.

nitishfy: please take a look now.

internal/webhook/external/github.go
```diff
@@ -190,6 +192,45 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+				return
+			}
+			pkgURL := pkg.GetPackageVersion().GetPackageURL()
+			// GitHub sometimes sends package URLs with a trailing colon
+			// and no tag (e.g., "ghcr.io/user/image:"). Such strings are
+			// not valid OCI image references and will result in empty
+			// repo URLs. We trim the trailing colon to avoid parsing errors
+			// and optionally allow fallback handling (e.g., ":latest").
+			if strings.HasSuffix(pkgURL, ":") {
+				pkgURL = strings.TrimSuffix(pkgURL, ":") + ":latest"
```
hiddeco: There is no need to trim if you already know it has the suffix.

```suggestion
				pkgURL += "latest"
```

would do.

nitishfy: resolved.

internal/webhook/external/github.go
```diff
@@ -190,6 +192,45 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+			}
+			pkgURL := pkg.GetPackageVersion().GetPackageURL()
+			// GitHub sometimes sends package URLs with a trailing colon
+			// and no tag (e.g., "ghcr.io/user/image:"). Such strings are
+			// not valid OCI image references and will result in empty
+			// repo URLs. We trim the trailing colon to avoid parsing errors
+			// and optionally allow fallback handling (e.g., ":latest").
+			if strings.HasSuffix(pkgURL, ":") {
+				pkgURL += "latest"
+			}
```
krancour: Why are you putting a latest tag on the end of this?

We want a URL for the _repository_. You are incorrectly constructing a URL for a specific image _within_ the repository -- one that may not even exist.

And when normalized, the tag will be stripped away anyway.

You're manipulating the URL in a totally nonsensical way and then it's almost immediately undone by existing logic.

Stripping the trailing colon, if it exists is the only thing you should be doing here.

```suggestion
			pkgURL := strings.TrimSuffix(pkg.GetPackageVersion().GetPackageURL(), ":")
```

With this change, line +219 can be removed as well.

internal/webhook/external/github.go
```diff
@@ -23,6 +23,9 @@ const (
 	githubEventTypePackage = "package"
```
krancour, line 27:
```suggestion
	// githubEventTypePackage corresponds to a package push event when the webhook
	// is registered directly at the repository-level. i.e. When the webhook is
	// not registered indirectly via a GitHub App that's been installed into the
	// repository.
	githubEventTypePackage = "package"
```

internal/webhook/external/github.go
nitishfy: resolved.

internal/webhook/external/github.go
```diff
@@ -23,6 +23,9 @@ const (
...
 	githubEventTypePing    = "ping"
 	githubEventTypePush    = "push"
+	// githubEventTypeRegistryPackage is similar to package event
+	// but is only valid for GitHub Apps
+	githubEventTypeRegistryPackage = "registry_package"
```
krancour:
```suggestion
	// githubEventTypeRegistryPackage corresponds to a package push event when the
	// webhook is registered indirectly via a GitHub App that's been installed
	// into the repository. i.e. When the webhook is not registered directly at
	// the repository-level.
	githubEventTypeRegistryPackage = "registry_package"
```

internal/webhook/external/github_test.go
```diff
@@ -319,6 +357,83 @@ func TestGithubHandler(t *testing.T) {
 				require.JSONEq(t, `{"msg":"refreshed 1 warehouse(s)"}`, rr.Body.String())
 			},
 		},
+		{
+			name:       "success -- registry_package event - image",
```
krancour: Nit: It's never easy to come up with test names when you're testing a very specific scenario. I don't generally feel picky about the format, however, we should remain consistent with what was used by other test cases in this suite.

```suggestion
			name:       "success -- registry_package event -- image",
```

internal/webhook/external/github_test.go
```diff
@@ -319,6 +357,83 @@ func TestGithubHandler(t *testing.T) {
...
+				req.Header.Set(gh.SHA256SignatureHeader, sign(bodyBytes))
+				return req
+			},
+			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
+				require.Equal(t, http.StatusOK, rr.Code)
+				require.JSONEq(t, `{"msg":"refreshed 1 warehouse(s)"}`, rr.Body.String())
+			},
+		},
+		{
+			name:       "success -- registry_package event - chart",
```
krancour: Same as [here](https://github.com/akuity/kargo/pull/4699/files#r2245761801):

```suggestion
			name:       "success -- registry_package event -- chart",
```

internal/webhook/external/github.go
nitishfy: One question if a url something like `ghcr.io/user/image:latest` is not valid, how come `ghcr.io/user/image:v7.1` can be valid?

krancour:
@nitishfy can I assume you've seen this work end-to-end with actual payloads from GitHub?

internal/webhook/external/github.go
krancour: Depends what you mean by "invalid," I guess.

Both of those are valid URLs, _for specific images within the `ghcr.io/user/image` repository._

Neither of those is a URL for _the repository itself_. In that sense, you could say neither is a valid _repository URL._

nitishfy:
> @nitishfy can I assume you've seen this work end-to-end with actual payloads from GitHub?

@krancour Yes!
Screenshot 2025-07-31 at 9 25 51 PM

krancour:
> Yes!

Amazing. Thank you!

LGTM after the two final, very small bits of feedback are accounted for.

nitishfy:
Comments resolved, hopefully we can get this merged for v1.7 now :)

krancour (approved):

internal/webhook/external/github.go
nitishfy: okay so we can't officially call them as `package_url` even though GitHub writes it as `package_url`, correct?

nitishfy:
merge conflicts resolved.

internal/webhook/external/github.go
```diff
@@ -20,9 +20,18 @@ const (
 
 	github = "github"
 
+	// githubEventTypePackage is type of event corresponds to a package push when the webhook is
+	// registered directly at the repository-level. i.e. It's when the webhook
+	// is not registered indirectly via a GitHub App that's been installed into
+	// the repository.
```
krancour:
```suggestion
	// githubEventTypePackage corresponds to a package push event when the webhook
	// has been registered directly at the repository-level. i.e. When the webhook
	// has not been registered indirectly via a GitHub App that's been installed
	// into the repository.
```

internal/webhook/external/github.go
```diff
@@ -20,9 +20,18 @@ const (
...
 	githubEventTypePackage = "package"
 	githubEventTypePing    = "ping"
 	githubEventTypePush    = "push"
+	// githubEventTypeRegistryPackage corresponds to a package push event when the
+	// webhook is registered indirectly via a GitHub App that's been installed
+	// into the repository. i.e. When the webhook is not registered directly at
+	// the repository-level.
```
krancour:
```suggestion
	// githubEventTypeRegistryPackage corresponds to a package push event when the
	// webhook has been registered indirectly via a GitHub App that's been
	// installed into the repository. i.e. When the webhook has not been
	// registered directly at the repository-level.
```

internal/webhook/external/github.go
```diff
@@ -197,6 +205,41 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
+			// These are the only types of packages we care about.
+			case ghcrPackageTypeContainer, ghcrPackageTypeDocker:
+			default:
+				xhttp.WriteResponseJSON(w, http.StatusOK, nil)
+				return
+			}
+			// GitHub sometimes sends package URLs with a trailing colon
+			// and no tag (e.g., "ghcr.io/user/image:"). Such strings are
+			// not valid OCI image references and will result in empty
+			// repo URLs. We trim the trailing colon to avoid parsing errors
```
krancour:
```suggestion
			// GitHub sometimes sends package URLs with a trailing colon and no tag
			// (e.g., "ghcr.io/user/image:"). Such strings are not valid OCI image
			// references. We trim the trailing colon to avoid parsing errors.
```

internal/webhook/external/github.go
```diff
@@ -20,9 +20,18 @@ const (
 
 	github = "github"
 
+	// githubEventTypePackage corresponds to a package push event when the webhook
+	// has been registered directly at the repository-level. i.e. When the webhook
+	// is not registered indirectly via a GitHub App that's been installed into
+	// the repository.
```
krancour:
```suggestion
	// has not been registered indirectly via a GitHub App that's been installed
	// into the repository.
```

krancour:
This doesn't account for changes included in #4560.

I will take it from here.

krancour (approved):

krancour:
Tests keep timing out. I think accounting for changes in #4560 broke something else here.

I will investigate and fix.

PR #4704, opened 2025-07-28
fix: take greater care when examining media types

krancour:
My initial pass at determining whether a URL should be normalized as if it is an image repo URL or a chart repo URL was flawed.

What complicates matters is that not all platforms that send us webhooks are consistent or clear about what any media type information they may give us is actually _for_.

Examples:

- GitHub separately gives us the media types for both the manifest and for the config blob!
- DockerHub gives us the media type for the config blob only.
- ACR gives us the media type for the manifest only.
- Quay tells us nothing at all about media types.

This PR is a comprehensive refactor and fix of the logic for determining correct normalization on the basis of media type. Note that when no inference can be drawn, we do what we've always done in the case of Quay and normalize both ways and let logic downstream determine if any warehouses at all subscribe to either of those.

Note this is a large factor in difficulties @nitishfy has encountered in implementing https://github.com/akuity/kargo/issues/4562

### DIFF
--- a/internal/webhook/external/media_types.go
+++ b/internal/webhook/external/media_types.go
@@ -2,45 +2,93 @@ package external
 
 import (
 	"slices"
+	"strings"
 
 	"github.com/akuity/kargo/internal/helm"
 	"github.com/akuity/kargo/internal/image"
 )
 
 const (
-	ociImageIndexMediaType    = "application/vnd.oci.image.index.v1+json"
-	ociImageManifestMediaType = "application/vnd.oci.image.manifest.v1+json"
+	ociImageIndexMediaType      = "application/vnd.oci.image.index.v1+json"
+	ociImageManifestMediaType   = "application/vnd.oci.image.manifest.v1+json"
+	ociImageConfigBlobMediaType = "application/vnd.oci.image.config.v1+json"
 
-	dockerManifestListMediaType = "application/vnd.docker.distribution.manifest.list.v2+json"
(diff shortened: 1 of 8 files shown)


internal/webhook/external/media_types.go
```diff
@@ -2,45 +2,93 @@ package external
...
+func getNormalizedImageRepoURLs(repoURL, mediaType string) []string {
+	if strings.HasPrefix(mediaType, "application/vnd.cncf.helm") {
+		return []string{helm.NormalizeChartRepositoryURL(repoURL)}
+	}
+	if slices.Contains(probableContainerImageMediaTypes, mediaType) {
+		return []string{image.NormalizeURL(repoURL)}
+	}
+	return []string{
+		image.NormalizeURL(repoURL),
+		helm.NormalizeChartRepositoryURL(repoURL),
```
hiddeco, line 96: Is there a chance for these to be the same, and should we e.g. use `slices.Compact()`?

krancour: More than a chance. _At the moment_, I guarantee they will be the same. I may be making some unrelated changes soon for a separate issue and then I guarantee they _won't_ be the same...

But the real point is that whether they're always the same or always not the same, we expect reliability here and should assume the possibility of being the same always exists...

The logic that actually _does_ the refreshing sorts and dedupes these URLs, but compacting here and now seems like a good optimization.

Consider it done.

krancour: Addressed in 308e187146d01dbb96ba966b29487d4cf41a41cc

PR #4708, opened 2025-07-29
feat(controller): allow a controller for a specific shard to also be the default controller

krancour:
Requested by @jessesuen and @gdsoumya.

This is meant to be fully backwards compatible for users who may upgrade their chart without any updates to their values.

This is still a first draft, but I think it wouldn't be too soon to get some feed back @hiddeco. 🙏 I'll DM you to discuss in the morning.


krancour:
Added some tests and tried this out e2e. Discovered some indices needed adjustment and have now accounted for those as well. Should be r4r.

charts/kargo/values.yaml
```diff
@@ -433,7 +433,10 @@ controller:
...
+  ## @param controller.isDefault When running multiple controllers backed by a single underlying control plane, designating this controller as the default will cause it to operate on resources not assigned to a specific shard. If controller.shardName is undefined, this controller will be considered the default **regardless** of the value of this field (as that was the behavior prior to the introduction of this field). If controller.shardName **is** defined, this controller will not be considered the default **unless, additionally** this field is true. i.e. A controller is effectively considered the default if !controller.shardName || controller.isDefault. If controller.shardName is defined **and** this field is true, this controller will operate **both** on resources explicitly assigned to it **as well as** those not assigned to a specific shard.
```
hiddeco: Does this support backtick blocks? If it does, I feel it might be better to do something like:

```suggestion
  ## @param controller.isDefault When running multiple controllers backed by a single underlying control plane, designating this controller as the default will cause it to operate on resources not assigned to a specific shard. If `controller.shardName` is undefined, this controller will be considered the default **regardless** of the value of this field (as that was the behavior prior to the introduction of this field). If `controller.shardName` **is** defined, this controller will not be considered the default **unless, additionally** this field is true. i.e. A controller is effectively considered the default if `!controller.shardName || controller.isDefault`. If `controller.shardName` is defined **and** this field is true, this controller will operate **both** on resources explicitly assigned to it **as well as** those not assigned to a specific shard.
```

to improve readability.

The same applies to the block below.

krancour: Probably... I'll look... sometimes when I'm editing comments in values.yaml, I forget it's translated to markdown and not plain text.

hiddeco (approved):
Assuming my comments around the chart documentation are taken care of, this looks good to me.

charts/kargo/values.yaml
krancour: Fixed in f27c1e98d8129496d85c8b5a10e880960d4713c9

Glad we looked at this... as it turns out, a much bigger problem was unescaped `||` in my pseudo-expressions. They screw up the table rows. You'd think the surrounding backticks would account for this, but alas, they do not.

I resolved that by changing the syntax for those pseudo-expressions to more closely resemble the syntax used by chart templates (which makes more sense anyway).

PR #4710, opened 2025-07-29
chore(api)!: remove deprecated ProjectSpec

hiddeco:
As noted for `v1.5.0`, this was scheduled for removal in `v1.7.0`: https://docs.kargo.io/release-notes/v1.5.0#new-deprecations


hiddeco:
@Marvin9 can you please check if I handled the `// deprecation` in the UI properly? I assume that the newly reflected field in the Stage status should be sufficient — but it would be great if you could confirm.

fuskovic (approved):

Issue #4369, opened 2025-06-10
UI: promotion step alias near proximity to promotion step name


Marvin9:
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Bring promotion step alias right beside promotion step name

<img width="1186" alt="Image" src="https://github.com/user-attachments/assets/5143d6d9-7bfb-43c2-b7dd-5593ee345771" />

# Motivation

When there are multiple same steps, alias are helpful to distinguish by its purpose

PR #4727, opened 2025-07-31
feat(ui): promo step alias name position near promo step name

Marvin9:
fixes https://github.com/akuity/kargo/issues/4369

Screenshot 2025-07-31 at 1 44 08 PM (2)

### DIFF
--- a/ui/src/features/stage/promotion-step.tsx
+++ b/ui/src/features/stage/promotion-step.tsx
@@ -105,6 +105,8 @@ export const Step = ({
     )
     .map((plugin) => plugin.DeepLinkPlugin?.PromotionStep?.render);
 
+  const shortenStepName = step?.as?.length > 25 ? step.as.slice(0, 25) + '...' : step.as;
+
   return {
     className: classNames('', {
       'border-green-500': progressing,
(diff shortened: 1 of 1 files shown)


hiddeco:
Should the horizontal padding be reduced a bit? It feels like it takes up quite some space now.

Marvin9:
@hiddeco thats intentional fixed width so that all boxes are even. I am pretty sure @rpelczar might have some ideas for that

rpelczar:
@Marvin9, I see two other options. Let's keep a fixed width for this new col, but without 100% width for the tag component. We can also move the alias name next to the name.

Marvin9:
@rpelczar like this?

Screenshot 2025-08-04 at 6 43 15 PM (2)

If we move alias, then its the same problem, uneven boxes that we have to fix width to make it even

hiddeco:
@Marvin9 I like this already better than the first screenshot.

ui/src/features/stage/promotion-step.tsx
```diff
@@ -105,6 +105,8 @@ export const Step = ({
     )
     .map((plugin) => plugin.DeepLinkPlugin?.PromotionStep?.render);
 
+  const shortenStepName = step?.as?.length > 25 ? step.as.slice(0, 25) + '...' : step.as;
```
rpelczar, line 108: Can you use CSS for this?

rpelczar (approved):

hiddeco:
We can deal with the CSS changes in a patch release. Merging this as it's one of the few things left for 1.7.0.

hiddeco:
We can deal with the CSS changes in a patch release. Merging this as it's one of the few things left for 1.7.0.

Marvin9:
I actually didn't push the change of the screenshot @hiddeco. Raising a PR and we can backport

PR #4728, opened 2025-07-31
chore: address nits `http-download` step

hiddeco:
Follow up to #4662


test.out
krancour, line 1: I assume this was committed accidentally.

hiddeco: Yes, absolutely. Which reminds me I need to add it to my global `.gitignore`.

PR #4734, opened 2025-07-31
enable project create/delete "self-service"

### LINKED ISSUE #3963: Self-Serviceable Projects
# Proposed Feature

Kargo should provide a way to "self-service" Project creation/administration/deletion. This would mean an end user could create a new project, be the admin of that project, without also having privileges to access/delete other projects. 

Today, this is not easy to do because we use fully utilize K8s RBAC with respect to CRUD on projects, which is a cluster-scoped resource. For example, to give self-service privileges to only delete a single project named `foo` and not others, a user would have to create a `ClusterRole` that would give `delete` privileges to that project by name using `resourceNames`:

```yaml
kind: ClusterRole
metadata:
  name: project-foo-admin
rules:
- apiGroups: ["kargo.akuity.io"]
  resources: ["projects"]
  verbs: ["get", "delete", "update", "patch"]
  resourceNames: ["foo"]
```

# Motivation

We wish to give end users more self-serviceability when it comes to managing their own projects. Without a feature like this, either a Kargo administrator would always need to be involved in project creation/deletion, or all Kargo end-users users would need to be given project create/delete privileges. 

# Suggested Implementation

I think several techniques can be employed to achieve the desired UX:

1. `kargo-management-controller` automation

The `kargo-management-controller` would automate the creation of project-specific RBAC. For example, it would:
* Create a `ClusterRole`, `kargo-project-admin-foo`, which has privileges to delete the single project, `foo`
* Create a `RoleBinding` from namespaced-scoped `kargo-admin` `ServiceAccount` to the `kargo-project-admin-foo` `ClusterRole`
* Annotate the `kargo-admin` ServiceAccount with OIDC e-mail claims of the user who created the project

2. `kargo-project-creator` built-in `ClusterRole`

A Kargo admin would decide if they want their users to self-service project creation. If yes, they would bind OIDC claims to a new built-in `ClusterRole` / `ServiceAccount` named `kargo-project-creator`. This role would only have the ability to create projects.

3. ProjectConfig CRD

Not sure if this will be necessary, but now that we have a namespace-scoped ProjectConfig CRD, I think it could be used as part of the self-service story. For example, changes to the `ProjectConfig` CRD itself might automate management of kubernetes RBAC. 

4. Validating Webhooks

I hope not to resort to this, but if necessary, we could use our validating webhook to decide if certain requests could go through, enforce immutability of certain fields, check claims, etc...

krancour:
Closes #3963 

The commits in this PR are narrowly scoped into digestable chunks that can be reviewed independently of one another.

1. Chart now create a new `ServiceAccount`, `ClusterRole`, and `ClusterRoleBinding`, all called `kargo-project-creator`.

    These give the `kargo-project-creator` `ServiceAccount` the same permissions as the `kargo-user` `ServiceAccount` _plus_ permission to create `Project`s.

    As with the existing `kargo-admin`, `kargo-user`, and `kargo-viewer` `ServiceAccount`s, the chart offers operator the ability to map users to the new `kargo-project-creator` `ServiceAccount` using any OIDC claims they wish.

1. API endpoints for creating/applying resources now have a special case for `Project` creation: they annotate the Project with information about the user who created it.

    It's done in these API endpoints and not elsewhere (e.g. a defaulting webhook) because _this_ is where we have the most information about the user available to us.

1. If it didn't already exist, the management controller already was creating a `kargo-admin` `ServiceAccount` (among other things) in a `Project`'s associated namespace when it reconciles a `Project`.

    Now when that `ServiceAccount` is created, the creator annotation on the `Project` is inspected and that `ServiceAccount` is automatically annotated such that it is mapped to the creator.

    When reconciling `Project` resources, the management controller will also now, if it doesn't already exist, create a Project-specific `ClusterRole` that permits updating and deleting this one `Project`. It will also, if it doesn't already exist, create a Project-specific `ClusterRoleBinding` to bind the `kargo-admin` `ServiceAccount` within the `Project`'s associated namespace to this new Project-specific `ClusterRole`.

    The net effect is that any user mapped to the `kargo-admin` `ServiceAccount` within the `Project`'s associated namespace is now able to update and delete that `Project`. _Because that mapping now automatically includes the user who created the `Project`, the `Project` can now be update and delete it by that user._

    The management controller will clean up these new cluster-scoped, yet still Project-specific, resources when a `Project` is deleted.

1. The authorizing Kubernetes client wrapper used by the API has, historically, only looked in "global" `ServiceAccount` namespaces for `ServiceAccounts` associated with the user when testing for permission to perform an operation on a _cluster-scoped_ resource.

    Now, there is a special case for `Project` resources. `SeviceAccount`s within the `Project`'s associated namespace will also be considered.

1. Last, if someone were to use a command like `kargo create -f my-project.yaml`, they would encounter some difficulty. The `Project` resource would be created, but because the creator's status as an admin of that `Project` is established asynchronously by the management controller, the creation of all other resources meant to reside _within_ the Project's associated namespace would almost certainly fail for lack or proper permissions.

    One possible approach to resolving this would have been similar to what we did to synchronously ensure the existence of a `Project`'s associated namespace as the `Project` is created -- by using a webhook.

    Due to the number of resources involved, replicating too much of the management controller's responsibilities synchronously in a webhook was unattractive.

    The approach taken was to add a special case for `Project` resources to the create/apply endpoints of the Kargo API server. If a `Project` is created successfully by an API call, we know the user had permission to do so and will shortly, thanks to the management controller, be made an admin of that new `Project`. With this being the case, all further creates/applies of resources within that `Project`'s associated namespace _within that same API call_ are performed using elevated permissions. (It uses the API server's own internal client.)

I will try to create a short demo video tomorrow.


charts/kargo/templates/api/configmap.yaml
```diff
@@ -50,8 +50,8 @@ data:
   {{- if .Values.api.oidc.cliClientID }}
   OIDC_CLI_CLIENT_ID: {{ quote .Values.api.oidc.cliClientID }}
   {{- end }}
-  OIDC_USERNAME_CLAIM: {{ quote .Values.api.oidc.usernameClaim | default "email" }}
   {{- end }}
+  OIDC_USERNAME_CLAIM: {{ quote .Values.api.oidc.usernameClaim | default "email" }}
```
krancour, line 54: This fixes a bug where `OIDC_USERNAME_CLAIM` was not getting set if you were using Dex.

internal/controller/management/projects/projects.go
```diff
@@ -134,6 +135,30 @@ type reconciler struct {
...
+		client.Object,
+		...client.DeleteOption,
+	) error
+
+	deleteClusterRoleBindingFn func(
+		context.Context,
+		client.Object,
+		...client.DeleteOption,
+	) error
 }
```
krancour, line 163: I know we're moving away from this method of mocking out specific bits of functionality during tests, but this reconciler is stuck in the past and I wasn't about to refactor the whole thing.

charts/kargo/templates/users/service-accounts.yaml
```diff
@@ -41,3 +41,17 @@ metadata:
...
+  name: kargo-viewer
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "kargo.labels" . | nindent 4 }}
+  {{- if .Values.api.oidc.viewers.claims }}
+  annotations:
+    {{- range $claim, $values := .Values.api.oidc.viewers.claims }}
+    rbac.kargo.akuity.io/claim.{{ $claim }}: {{ quote (join "," $values) }}
+    {{- end }}
+  {{- end }}
```
hiddeco:
```suggestion
  {{- end }}

```

charts/kargo/values.yaml
```diff
@@ -171,20 +171,20 @@ api:
         # groups:
         # - kargo-admin
 
-    viewers:
-      ## @param api.oidc.viewers.claims Subjects having any of these claims will automatically receive read-only access to all Kargo resources. This does not include any access to Secrets.
+    projectCreators:
+      ## @param api.oidc.projectCreators.claims Subjects having any of these claims will automatically receive the permissions of the karo-user role (see `api.oidc.users`) **plus** permission to create new `Project`s. When a `Project`` is created by such a user via the CLI or UI (i.e. through the API and not through `kubectl`) they will automatically receive admin permissions within `that`` Project as well as permission to update and delete the cluster-scoped Project resource itself.
```
hiddeco: I _think_ this is supposed to be this?

```suggestion
      ## @param api.oidc.projectCreators.claims Subjects having any of these claims will automatically receive the permissions of the karo-user role (see `api.oidc.users`) **plus** permission to create new `Project`s. When a `Project` is created by such a user via the CLI or UI (i.e. through the API and not through `kubectl`) they will automatically receive admin permissions within that `Project` as well as permission to update and delete the cluster-scoped `Project` resource itself.
```

internal/controller/management/projects/projects.go
```diff
@@ -992,6 +1041,61 @@ func (r *reconciler) ensureDefaultUserRoles(
 		rbLogger.Debug("created RoleBinding in project namespace")
 	}
 
+	// This ClusterRole allows those bound to it to update and delete one specific
+	// Project. This is necessary since Projects are cluster-scoped, meaning this
+	// permission cannot be defined anywhere except at the cluster-level.
+	//
+	// TODO(krancour): There's an obvious problem here if
+	// len("kargo-project-admin-") + len(project.Name) > max resource name length.
```
hiddeco: The strategy I have always used for this is to do something like the following:

```go
const (
	maxNameLength   = 256
	shortHashLength = 8
)

func shortenName(name string) string {
    if len(name) <= maxNameLength {
        return name
    }
    
    sum := fmt.Sprintf("%x", sha256.Sum256([]byte(name)))
    prefix := name[:maxNameLength-(shortHashLength+1)]
    
    // Remove trailing dashes to avoid double dashes
    prefix = strings.TrimRight(prefix, "-")
    
    return prefix + "-" + sum[:shortHashLength]
}
```

charts/kargo/values.yaml
krancour: You're right. I've no idea what happened here.

internal/controller/management/projects/projects.go
krancour: See https://github.com/akuity/kargo/pull/4734/commits/5056c1e6c4f1ba4b5d93dfa9699da6bbeeb58bda

hiddeco (approved):

PR #4736, opened 2025-08-01
fix(ui): infinite render in yaml view

Marvin9:
caused by https://github.com/akuity/kargo/pull/4725


jmalisz-akuity (approved):
Should be safer, but I think that we should just remove the option to hide managed fields from editors that can be changed.  Supporting that complicates code.

Marvin9:
@krancour let us know if you still think of any need of this, otherwise we can remove in later pr

PR #4740, opened 2025-08-01
fix(api/webhook): move part of Stage step validation to webhook

hiddeco:
Follow up to #4732

As `if` is a reserved CEL keyword, the expression is not allowed to reference a field with that name in Kubernetes versions below 1.32.x. Because of this, we need to temporarily resort to validating it using a webhook, and reintroduce it once these versions have reached EOL.

### DIFF
--- a/internal/webhook/kubernetes/stage/webhook.go
+++ b/internal/webhook/kubernetes/stage/webhook.go
@@ -40,7 +40,11 @@ type webhook struct {
 		client.Object,
 	) error
 
-	validateSpecFn func(*field.Path, kargoapi.StageSpec) field.ErrorList
+	validateSpecFn                  func(*field.Path, kargoapi.StageSpec) field.ErrorList
+	validatePromotionStepTaskRefsFn func(
+		*field.Path,
+		[]kargoapi.PromotionStep,
+	) field.ErrorList
 
 	isRequestFromKargoControlplaneFn libWebhook.IsRequestFromKargoControlplaneFn
 }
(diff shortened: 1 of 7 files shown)


internal/webhook/kubernetes/stage/webhook.go
```diff
@@ -234,3 +251,54 @@ func (w *webhook) validateRequestedFreight(
...
+		//
+		// xref: https://github.com/akuity/kargo/pull/4732
+		// xref: https://github.com/kubernetes/kubernetes/pull/126977
+		if step.If != "" {
+			errs = append(
+				errs,
+				field.Forbidden(
+					f.Index(i).Child("if"),
+					"PromotionTemplate step referencing a task cannot have an 'if' condition",
+				),
```
hiddeco, line 284: This uses "forbidden" because "invalid" would reflect the value, which appeared to be impractical for the config validation below. This still produces a nice error, however:

```console
$ kubectl apply -f kargo.yaml
project.kargo.akuity.io/kargo-demo-06 created
projectconfig.kargo.akuity.io/kargo-demo-06 created
warehouse.kargo.akuity.io/kargo-demo created
promotiontask.kargo.akuity.io/promo-process created
stage.kargo.akuity.io/test created
stage.kargo.akuity.io/prod created
analysistemplate.argoproj.io/kargo-demo created
The Stage "uat" is invalid: spec.promotionTemplate.spec.steps[0].if: Forbidden: PromotionTemplate step referencing a task cannot have an 'if' condition
```

PR #4742, opened 2025-08-01
fix: shorten Stage label value exceeding max length

### LINKED ISSUE #4600: Stage can't be deleted because Stage name is too long
# Checklist

# Description

As documented in Kubernetes:

```
In Kubernetes, resource names generally have a limit of 253 characters. However, some resources or features may have stricter limits, such as labels which have a 63 character limit
```

When creating a Kargo Stage, if `rollouts` are enabled, some `AnalysisRuns` may be created.
When deleting the stage, Kargo controller will try to delete all `AnalysisRuns` with an `label` = Kargo `stage` name, even if none is defined in the Stage.

see in:

- https://github.com/akuity/kargo/blob/main/internal/controller/stages/regular_stages.go#L1960
- https://github.com/akuity/kargo/blob/main/internal/controller/stages/control_flow_stages.go#L569

(BTW, this should probably be refactored, 2 identical functions may lead to problems).

The controller will receive an error from the K8s `client.DeleteAllOf)` like:

```
unable to parse requirement: values[0][kargo.akuity.io/stage]: Invalid value: \"stg-promotion-argocd-demo-project-6-test-staging-us-central-cluster-cluster\": must be no more than 63 characters
```

and the deletion on the `Stage` will stay pending.
To solve the issue and delete the `Stage`, it is needed to edit it and remove the `finalizer` set on it.

# What should happen:

Deleting a `Stage` should not block.

# Possible solutions

- Because Kargo is using `Labels` to link child elements like `AnalysisRuns`, it should reject all `Stages` created with a name longer than 63 chars.
- Truncate the label length to 63 chars, but there is a risk of generating many identical labels.
- Generate a 63 char long hash of the name and use it in the label in place of the stage name
- Check if an `AnalysisRuns` was created and don't try to delete it if not needed (that would partially solve the pending deletion issue, not the long-name issue)

# Screenshots

none

# Steps to Reproduce

- create a `Stage` with a name longer than 63 chars
- delete the stage

# Version

```
1.6.1
```

# Logs

```
error handling deletion of Stage: error deleting AnalysisRuns for Stage \"stg-promotion-argocd-demo-project-6-test-staging-us-central-cluster-cluster\" in namespace \"argocd-demo-project-argocd-demo-project-6\": unable to parse requirement: values[0][kargo.akuity.io/stage]: Invalid value: \"stg-promotion-argocd-demo-project-6-test-staging-us-central-cluster-cluster\": must be no more than 63 characters" name=stg-promotion-argocd-demo-project-6-test-staging-us-central-cluster-cluster namespace=argocd-demo-project-argocd-demo-project-6 reconcileID="\"8993b1a6-4bd0-4107-8481-b42f4902631f\""
```

hiddeco:
Fixes #4600

In a follow-up, the garbage collection logic will be factored out and/or owner references will be introduced to allow Kubernetes itself to handle the lifecycle.

### DIFF
--- a/internal/server/get_analysisrun_logs_v1alpha1.go
+++ b/internal/server/get_analysisrun_logs_v1alpha1.go
@@ -345,15 +345,22 @@ func (s *server) getStageFromAnalysisRun(
 	ctx context.Context,
 	run *rolloutsapi.AnalysisRun,
 ) (*kargoapi.Stage, error) {
-	stageName, ok := run.Labels[kargoapi.LabelKeyStage]
+	// The (full) name of the Stage is stored in the AnalysisRun's annotations
+	// when the label value contains a (hash-)shortened version of the Stage
+	// name. Check for this value first, and if it is not found, check the
+	// labels. If neither is found, return an error.
+	stageName, ok := run.Annotations[kargoapi.AnnotationKeyStage]
 	if !ok {
-		return nil, connect.NewError(
-			connect.CodeNotFound,
-			fmt.Errorf(
-				"AnalysisRun %q in namespace %q has no stage label",
-				run.Name, run.Namespace,
-			),
-		)
+		stageName, ok = run.Labels[kargoapi.LabelKeyStage]
+		if !ok {
+			return nil, connect.NewError(
+				connect.CodeNotFound,
+				fmt.Errorf(
+					"AnalysisRun %q in namespace %q has no stage label",
+					run.Name, run.Namespace,
+				),
+			)
+		}
 	}
 	stage, err := api.GetStage(
 		ctx,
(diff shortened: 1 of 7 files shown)


internal/server/get_analysisrun_logs_v1alpha1.go
```diff
@@ -345,15 +345,22 @@ func (s *server) getStageFromAnalysisRun(
 	ctx context.Context,
 	run *rolloutsapi.AnalysisRun,
 ) (*kargoapi.Stage, error) {
-	stageName, ok := run.Labels[kargoapi.LabelKeyStage]
+	// The (full) name of the Stage is stored in the AnalysisRun's annotations
+	// when the label value contains a (hash-)shortened version of the Stage
+	// name. Check for this value first, and if it is not found, check the
+	// labels. If neither is found, return an error.
+	stageName, ok := run.Annotations[kargoapi.AnnotationKeyStage]
```
hiddeco, line 352: As before this change, Stages with a name >63 characters were not possible at all. This change is fully backwards compatible.

fuskovic (approved):

PR #4764, opened 2025-08-05
feat(ui): pipeline filters must persist even if not in URL query

Marvin9:


bhavyabhut (approved):

ui/src/features/project/pipelines/use-persist-filters.ts
```diff
@@ -17,16 +16,16 @@ export const usePersistPreferredFilter = (
...
-      try {
-        updatePreferredFilter(JSON.parse(preferredFilterLocal));
-      } catch (e) {
-        // silent
-      }
-    }
-  }, []);
+  if (filters) {
+    return JSON.parse(filters);
+  }
```
rpelczar: Isn't it better to keep try/catch here?

Marvin9: Done

Issue #4739, opened 2025-08-01
UI Bug - Deleting Freight does not update UI (needs refresh)


stephenpope:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description
After deleteing `Freight` using the "Delete Freight" option does not refresh, it is still on screen and the page needs to be refreshed before it is removed.

# Screenshots

<img width="302" height="142" alt="Image" src="https://github.com/user-attachments/assets/884b424d-e608-4c6a-b50d-14e4bfcc92df" />

# Steps to Reproduce

Delete a `Freight` item using the menu

# Version

```
1.6.2
```

PR #4772, opened 2025-08-06
fix(ui): auto refresh on delete

bhavyabhut:
https://github.com/akuity/kargo/issues/4739


ui/src/features/project/pipelines/freight/delete-freight-modal.tsx
```diff
@@ -14,9 +19,22 @@ export const DeleteFreightModal = ({
   freight
 }: ModalComponentProps & { onDelete: () => void; freight: Freight }) => {
   const { name: project } = useParams();
+  const queryClient = useQueryClient();
+
   const { mutate: deleteAction, isPending } = useMutation(deleteFreight, {
     onSuccess: () => {
       message.success('Freight successfully deleted');
+
+      queryClient.invalidateQueries({
```
Marvin9: No need of this if we are handling on watch stream right?

Issue #4031, opened 2025-05-08
docs: working with warehouses doesn't have private repo section


fuskovic:
# Proposed Feature

I think this should be explicit with examples.

# Motivation

Reduce friction for new Kargo users

krancour:
Probably all that is needed is a well placed link to https://docs.kargo.io/user-guide/security/managing-credentials

PR #4844, opened 2025-08-12
docs: working with private repositories

fykaa:
closes https://github.com/akuity/kargo/issues/4031

### DIFF
--- a/docs/docs/40-operator-guide/40-security/40-managing-credentials.md
+++ b/docs/docs/40-operator-guide/40-security/40-managing-credentials.md
@@ -6,12 +6,11 @@ sidebar_label: Managing Credentials
 # Managing Credentials
 
 Kargo `Warehouse`s frequently require read-only access to private repositories.
-Promotion processes often require this as well, and may even require read/write
+Promotion processes often require this as well, and may also require read/write
 access to Git repositories.
 
-This page focuses on an operator's role in providing Kargo projects with necessary
-credentials and configuring various authentication options, including ambient
-credentials that require minimal configuration within Kargo itself.
+This section focuses on an operator's role in providing Kargo Projects with
+necessary credentials.
 
 :::info
 __Not what you were looking for?__
--- a/docs/docs/50-user-guide/50-security/30-managing-credentials.md
+++ b/docs/docs/50-user-guide/50-security/30-managing-credentials.md
@@ -6,11 +6,11 @@ sidebar_label: Managing Credentials
 # Managing Credentials
 
 Kargo `Warehouse`s frequently require read-only access to private repositories.
-Promotion processes often require this as well, and may even require read/write
+Promotion processes often require this as well, and may also require read/write
 access to Git repositories.
 
-This page provides a comprehensive overview of the authentication
-options available for managing and using such credentials in Kargo projects.
+This section presents an overview of how users can manage and use such
+credentials within their Kargo Projects.
 
 :::info
 __Not what you were looking for?__


docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -671,6 +671,19 @@ Helm chart repository subscriptions can be defined using the following fields:
         semverConstraint: ^1.0.0
   ```
 
+
```
krancour: Markdown linters usually doesn't like double line breaks -- at least the one I'm using does not.

```suggestion
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -671,6 +671,19 @@ Helm chart repository subscriptions can be defined using the following fields:
...
+When your `Warehouse` needs to access private repositories (Git repositories,
+container image registries, or Helm chart repositories), you'll need to configure
+appropriate credentials.
+
+Kargo uses specially labeled Kubernetes `Secret` resources to store repository
+credentials. These credentials are matched to repositories by URL and repository type.
+
+For a more detailed information on creating and managing repository credentials
+refer to [Managing Credentials](../50-security/30-managing-credentials.md).
```
krancour: First, I like the placement of this new section. 👍 

I probably wouldn't enumerate the different kinds of repositories here. That's been done already on this page and we want to avoid being too repetitive.

There are also various sorts of credentials that don't involve Secrets at all. For instance, repos in ECR can be accessed using Pod Identity or IRSA. Repos in GAR can be accessed using Workload Identity Federation. (These are both examples of so-called "ambient credentials" that require zero configuration on the Kargo end as long as everything is configured correctly on the AWS or GKE end. No Secrets involved.)

If we attempt to make the language here more inclusive of these other (very common, btw) scenarios, we'll end up duplicating too much of the information we'd rather the reader got by just following the provided link. So maybe "less is more" and we're better of just not going into much detail here. i.e. Maybe we just don't need to mention Secrets at all.

These changes probably boil the whole section down to something _very_ short and sweet like this (feel free to tweak further):

```suggestion
## Working with Private Repositories

Frequently, `Warehouse`s require access to private repositories, in which case
appropriate credentials must be made available in some form. The many available
authentication options are covered in great detail on the
[Managing Credentials](../50-security/30-managing-credentials.md) page.
```

If the reader follows the link, the first bit of text they'll see says:

> To orchestrate the promotion of Freight from Stage to Stage, Kargo will often require read/write permissions on private Git repositories and read-only permissions on private container image or Helm chart repositories.

Those are probably my words and I regret that Warehouses aren't explicitly mentioned there. Please reword that section so that readers who arrive there from the "Working with Warehouses" page will feel some sense of continuity on that topic.

I suggest something along the lines of:

```
Kargo `Warehouse`s frequently require read-only access to private repositories.
Promotion processes often require this as well, and may even require read/write
access to Git repositories. This page presents a comprehensive overview of the
many options available for authenticating to various types of repositories.
```

There is also a more operator-focused version of that page (operators have a lot more setup to do than users do when it comes to ambient credentials). If we could reword the introductory text of that page in the same way, that would further improve continuity for readers who jump from "Working with Warehouses" to the user-focused "Managing Credentials" page over to the operator-focused "Managing Credentials" page.

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -5,12 +5,13 @@ sidebar_label: Managing Credentials
 
 # Managing Credentials
 
-To orchestrate the promotion of `Freight` from `Stage` to `Stage`, Kargo will
-often require read/write permissions on private Git repositories and read-only
-permissions on private container image or Helm chart repositories.
+Kargo `Warehouse`s frequently require access to private repositories -
+read/write for Git repositories and read-only for container image or Helm chart
+repositories.
```
krancour: I don't mind that this is different from the text I'd suggested, but this isn't quite accurate...

Warehouses never write anything to anywhere. It's promotion processes that might push changes to a Git repo. (And in the future might also do things like push new tags to an image repo.)

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -5,12 +5,13 @@ sidebar_label: Managing Credentials
 
 # Managing Credentials
 
-To orchestrate the promotion of `Freight` from `Stage` to `Stage`, Kargo will
-often require read/write permissions on private Git repositories and read-only
-permissions on private container image or Helm chart repositories.
+Kargo `Warehouse`s frequently require read-only access to private repositories.
+Promotion processes often require this as well, and may even require read/write
+access to Git repositories.
```
krancour:
```suggestion
Kargo `Warehouse`s frequently require read-only access to private repositories.
Promotion processes often require this as well, and may also require read/write
access to Git repositories.
```

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -5,12 +5,13 @@ sidebar_label: Managing Credentials
...
 
-This section focuses on an operator's role in providing Kargo projects with
-necessary credentials.
+This page focuses on an operator's role in providing Kargo projects with necessary
+credentials and configuring various authentication options, including ambient
+credentials that require minimal configuration within Kargo itself.
```
krancour: Providing credentials and configuring authn options sounds sort of redundant. As I try to reword it to be less so, I'm finding it difficult to work in a mention of ambient credentials in the shorter, sweeter sentence. Even if possible with further wordsmithing effort, it's left me questioning why we're now mentioning ambient credentials here. There's a whole section dedicated to it. I'm not sure mentioning it in this little introductory blurb adds much value. In the end, I think the best thing may be to remove this change.

```suggestion
This section focuses on an operator's role in providing Kargo Projects with
necessary credentials.
```

docs/docs/50-user-guide/50-security/30-managing-credentials.md
```diff
@@ -5,12 +5,12 @@ sidebar_label: Managing Credentials
 
 # Managing Credentials
 
-To orchestrate the promotion of `Freight` from `Stage` to `Stage`, Kargo will
-often require read/write permissions on private Git repositories and read-only
-permissions on private container image or Helm chart repositories.
+Kargo `Warehouse`s frequently require read-only access to private repositories.
+Promotion processes often require this as well, and may even require read/write
+access to Git repositories.
```
krancour:
```suggestion
Kargo `Warehouse`s frequently require read-only access to private repositories.
Promotion processes often require this as well, and may also require read/write
access to Git repositories.
```

docs/docs/50-user-guide/50-security/30-managing-credentials.md
```diff
@@ -5,12 +5,12 @@ sidebar_label: Managing Credentials
...
 
-This section presents an overview of how users can manage and use such
-credentials within their Kargo projects.
+This page provides a comprehensive overview of the authentication
+options available for managing and using such credentials in Kargo projects.
```
krancour: The operator-focused version of this page is clear about its operator-focus.

I don't like this change loses the mention of users (as opposed to operators), which makes the intended audience less clear.

I'd suggest undoing this particular change.

```suggestion
This section presents an overview of how users can manage and use such
credentials within their Kargo Projects.
```

PR #4846, opened 2025-08-12
fix(step)!: tighten validation `desiredRevision`

hiddeco:
This sets the minimum length of the `desiredRevision` field to 40 characters (the lenght of a SHA1 commit reference) to signal to users that an absolute reference must be configured here (and not e.g., a branch name).

This will work with OCI in ArgoCD >=3.1 as well, as an image tag is resolved to a digest in this case (e.g., `sha256:abc123...`).


krancour:
Would it be worth updating the description text on those fields to be explicit that the values are expected to be immutable references to a specific revision? i.e. A full-length commit SHA (or digest) and _not_ something "floating," such as a branch name?

PR #4847, opened 2025-08-13
feat(event)!: Adds concrete event types

thomastaylor312:
This uses cloudevents to make sure it is easier to support arbitrary data types and also declares a discrete data type for the currently existing event.

This is just a Go API change and not a change to actual behavior. The existing method of publishing as a Kubernetes event is unchanged, but does give more flexibilty for managing/consuming the event


api/v1alpha1/event.go
```diff
@@ -1,22 +1,23 @@
...
+	AnnotationKeyEventFreightCharts          = AnnotationKeyEventPrefix + "freight-charts"
+	AnnotationKeyEventStageName              = AnnotationKeyEventPrefix + "stage-name"
+	AnnotationKeyEventAnalysisRunName        = AnnotationKeyEventPrefix + "analysis-run-name"
+	AnnotationKeyEventVerificationPending    = AnnotationKeyEventPrefix + "verification-pending"
+	AnnotationKeyEventVerificationStartTime  = AnnotationKeyEventPrefix + "verification-start-time"
+	AnnotationKeyEventVerificationFinishTime = AnnotationKeyEventPrefix + "verification-finish-time"
+	AnnotationKeyEventApplications           = AnnotationKeyEventPrefix + "applications"
 )
 
 const (
```
thomastaylor312, line 23: I would like to rename these to `EventType*` rather than `EventReason*` and have them use a new concrete type alias `type EventType string` to make things more clear. This is entirely optional and I didn't do it because I wasn't sure if we were ok breaking the API of this module

internal/api/event.go
```diff
@@ -46,18 +46,18 @@ func FormatEventKubernetesUserActor(u authnv1.UserInfo) string {
...
 }
 
-func NewFreightApprovedEventAnnotations(actor string, f *kargoapi.Freight, stageName string) map[string]string {
-	annotations := map[string]string{
-		kargoapi.AnnotationKeyEventProject:           f.Namespace,
-		kargoapi.AnnotationKeyEventFreightCreateTime: f.CreationTimestamp.Format(time.RFC3339),
-		kargoapi.AnnotationKeyEventFreightAlias:      f.Alias,
-		kargoapi.AnnotationKeyEventFreightName:       f.Name,
-		kargoapi.AnnotationKeyEventStageName:         stageName,
+func NewFreightApprovedEvent(actor string, f *kargoapi.Freight, stageName string) event.PromotionEvent {
```
thomastaylor312: I could also split this out as its own event if we prefer, rather than using a single event type to rule them all. I figured I'd keep it mostly the same as it currently is though

internal/controller/promotions/promotions.go
```diff
@@ -117,7 +116,7 @@ func SetupReconcilerWithManager(
 
 	reconciler := newReconciler(
 		kargoMgr.GetClient(),
-		libEvent.NewRecorder(ctx, kargoMgr.GetScheme(), kargoMgr.GetClient(), cfg.Name()),
+		k8sevent.NewKubernetesEventSender(libEvent.NewRecorder(ctx, kargoMgr.GetScheme(), kargoMgr.GetClient(), cfg.Name())),
```
thomastaylor312: I did this for compatibility, but I was wondering if I should actually have the public constructor for this (and other controller things I changed it in) take an `event.Sender` as an argument so it has to be passed in (and can be overridden) at the time of creation

internal/api/event.go
thomastaylor312: Answering my own question here: After looking at it more, I think we do want two different events. There will be duplicate fields between them, but they will work better as separate events

api/v1alpha1/event.go
krancour: DM'ed you.

pkg/event/freight.go
```diff
@@ -0,0 +1,179 @@
...
+	Project                string               `json:"project"`
+	Name                   string               `json:"name"`
+	StageName              string               `json:"stageName"`
+	Message                string               `json:"message"`
+	Actor                  *string              `json:"actor,omitempty"`
+	FreightCreateTime      time.Time            `json:"freightCreateTime"`
+	FreightAlias           *string              `json:"freightAlias,omitempty"`
+	FreightCommits         []kargoapi.GitCommit `json:"freightCommits,omitempty"`
+	FreightImages          []kargoapi.Image     `json:"freightImages,omitempty"`
+	FreightCharts          []kargoapi.Chart     `json:"freightCharts,omitempty"`
```
krancour: I feel like prefixing any field names with `Freight` "stutters." Being fields of the `Freight` type, isn't "Freight" implied? Also there might be a specific reason you did this that is escaping me.

pkg/go.mod
```diff
@@ -5,32 +5,124 @@ go 1.24.3
 replace github.com/akuity/kargo/api => ../api
 
 require (
+	github.com/akuity/kargo v1.7.2
```
krancour: The way things were before: `akuity/kargo` depends on `akuity/kargo/pkg` and `akuity/kargo/api`. `akuity/kargo/pkg` depends only on `akuity/kargo/api`.

I think this line hints that something snuck into `akuity/kargo/pkg` that is importing something from `akuity/kargo`. i.e. Some lower level thing is now importing some higher level thing. (This seems circular and I'm surprised `go mod` didn't totally barf on this.)

Why it matters: I'm sure that this was accidental, but it probably dragged along a lot of indirect dependencies that `akuity/kargo/pkg` does not otherwise need.

I haven't pinpointed the offending import yet.

pkg/event/freight.go
```diff
@@ -0,0 +1,179 @@
...
+	"time"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/google/uuid"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+)
+
+// FreightEvent represents the data available for various events related to freights.
+type FreightEvent struct {
```
krancour: I notice there's no `EventType` field and `EventType` is specified only when invoking this type's `ToCloudEvent()` method. Was there some specific rationale for doing it that way?

pkg/event/freight.go
thomastaylor312: No it was a copy paste artifact. I started from the promotion one and didn’t edit these names. I’ll do that

pkg/event/freight.go
thomastaylor312: Yes because the type only really matters in context of a cloud event. For this type data it doesn’t matter in context of this data

pkg/go.mod
thomastaylor312: I’ll look for it. I honestly thought it would have barfed too, which is why I didn’t notice this

pkg/event/freight.go
thomastaylor312: I can also include it on the type if you want as I can argue it either way

krancour: Well... before we make any final decision about that... here's a more general one about the composition of this type. It seems many of the fields aren't necessarily relevant to _all_ Freight-related events.

For example, Freight being created doesn't involve a Stage and therefore wouldn't need `StageName`. (Oddly that doesn't seem to be an event we actually track and I can't figure out why we wouldn't. So this might be a poor example, but it's a forward-looking one because we probably should/will care about that event someday.)

A better example might be that `VerificationStartTime`, `VerificationFinishTime`, `AnalysisRunName`, and `AnalysisTriggeredByPromotion` all seem irrelevant to a `FreightApproved` event.

So... Is the idea to just have strongly typed fields here for any/all of the bits of data that _may_ be relevant for _any_ sort of Freight-related event? And in many cases, some fields will remain empty?

thomastaylor312: I was trying to keep this as close to how it was, but my first inclination was just to have a separate concrete event for each event type (with a few shared types for a couple of the events). If you're open to that, I'm down to make more specific events

krancour: I think my main worry is over events looking rather amorphous. And that harkens back to my question about whether event type should be a field or not. Right now, I don't think I could look at a `FreightEvent` and recognize what kind of Freight-related event it actually represents unless I could infer that from careful examination of what fields are set and which remain empty.

> I'm down to make more specific events

As long as it's not too onerous, I think that makes sense. A lot of the fields obviously _are_ common to all Freight-related events, so probably 80% of all Freight-related events' fields can be accounted for by composing the events from a "base Freight event."

(Similar for Promotion-related events.)

But if this starts to look like way too much code, which it may, I could equally see heading in an entirely different direction, like (for example; and not at all thought all the way through), just creating functions like `NewFeightApprovedEvent(freight, stage)` that spit out a cloud event without involving any intermediate type with strongly typed fields. Other functions might take different args depending on what's relevant. `NewFreightCreatedEvent(freight)`, for instance, wouldn't need Stage details. (Again, I know this is an event we don't currently track, but it's a decent enough example of an event not needing data that other events do need.)

pkg/go.mod
thomastaylor312: Found it: the `expressions/function` dir is pulling from `internal` (which I also thought wouldn't work). Namely

```
"github.com/akuity/kargo/internal/controller/freight"
"github.com/akuity/kargo/internal/kargo"
```

thomastaylor312: Ok, I fixed this. Had to pull a few more things into `pkg`. As before, I kept it as surgical as possible and only pulled what was needed. Let me know if there are any things you don't like

pkg/event/event.go
```diff
@@ -0,0 +1,441 @@
...
+
+// AnnotationMarshaler is an interface for any type that can marshal out to our custom annotation
+// types
+type AnnotationMarshaler interface {
+	MarshalAnnotations() map[string]string
+}
+
+// Common is a struct that contains fields common to all events.
+type Common struct {
+	Project string  `json:"project"`
```
krancour, line 55: Just checking that we intend that this _can_ be empty for events that are not Project-scoped? (I know we have no such events yet, but I imagine we will before we know it.)

pkg/event/event.go
```diff
@@ -0,0 +1,441 @@
...
+}
+
+func (c *Common) SetMessage(message string) {
+	c.Message = message
+}
+
+// Promotion is a struct that contains common fields for promotion-related events.
+type Promotion struct {
+	*Freight
+	Name         string                 `json:"promotionName"`
```
krancour: No strong opinion, but I notice it's `promotionName` over the wire. I know it stutters, but naming it `PromotionName` here might avoid confusion elsewhere in the code. I could see this being mistaken for the name of the _promotion event_. But as I said... I don't feel strongly.

Edit: Same question applies to Freight.

krancour: Should we move this type to the `promotion.go` file here in the same dir?

Similar question for `Freight`.

pkg/event/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,90 @@
...
+	"encoding/json"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/pkg/event"
+)
+
+// FromKubernetesEvent converts a Kubernetes Kargo event to a CloudEvent.
```
krancour: Think this comment is no longer true.

krancour: Actually, I think CloudEvents seems to still be mentioned in a number of other comments as well.

pkg/event/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,90 @@
...
+func FromKubernetesEvent(evt corev1.Event) (event.Meta, error) {
+	var parsedEvent event.Meta
+	var err error
+
+	switch kargoapi.EventType(evt.Reason) {
```
krancour, line 18: This feels like something we're going to forget to update. I don't have any bright ideas though.

pkg/event/event.go
thomastaylor312: Yes, I can add omitempty if you are worried about it

pkg/event/event.go
thomastaylor312: I was trying to avoid stuttering because I definitely think of this as the name of the promotion. Events usually are referred to in terms of Types or IDs and not names, so I lean toward keeping it as is

pkg/event/kubernetes/kubernetes.go
thomastaylor312: Yep, I have (or had) a grumpy comment about "my kingdom for a real enum". The only thing I've found that might be helpful is this linter: https://github.com/nishanths/exhaustive

pkg/event/event.go
thomastaylor312: I've moved the types as well as suggested

pkg/event/event.go
krancour: Was more worried about how it will function when not defined than I was over whether the field name still appears in the event when its value is empty.

All good.

krancour (approved):

Issue #4833, opened 2025-08-11
account for normalization when matching regex to urls in "credentials database"


krancour:
Derived from #4703.

In the context of a Warehouse's subscriptions to various sorts of repositories, Kargo _searches_ for appropriate credentials. The search involves looking for a Secret in the Project's namespace (or in one of the "global" credentials namespaces) that is either an "exact" match, by _normalized_ URL, for the repository in question _OR_ is configured with a regular expression that matches the _normalized_ URL of the repository in question.

The aforementioned normalization process, used prior to URL comparisons throughout all of Kargo, is meant to prevent minor/inconsequential differences from causing two otherwise logically equivalent URLs from being judged _not equal._

Where credential lookup is concerned, normalization works just fine for direct comparison of two logically equivalent URLs. This is to say that if `url1` and `url2` are logically equivalent, it is guaranteed that (pseudocode) `normalized(url1) == normalized(url2)`.

This breaks down where regular expressions are concerned because the logic used to normalize a full URL cannot be applied to normalize a regular expression. This results in a case where a sensible-looking regular expression such as `^oci://foo/bar/` will _not_ match the URL `oci://foo/bar/something` due to the fact that `^oci://foo/bar/` does not match the _normalized_ representation of `oci://foo/bar/something`, which happens to be `foo/bar/something` (no leading protocol).

The short-sighted way to fix this would be documenting the fact that URLs are normalized before we begin searching for credentials, and calling out that, for instance, `^oci://foo/bar/` will _not_ match the URL `oci://foo/bar/something` as a result of this. I consider it short-sighted because this invites the user to understand something that they'd otherwise be content remaining oblivious to _and_ it would limit our ability to _change_ the normalization process (which is fundamentally an implementation detail) in the future.

The more strategic fix I would propose is that where regular expressions are concerned, a match be attempted against _both_ the normalized and original URLs. This would not require that we share/document details of the normalization process, which will spare the users that concern and preserve our freedom to alter the normalization process in the future.

Necessary changes should be minor and isolated to `internal/credentials/kubernetes/database.go` (and its corresponding tests).

While already touching this bit of code, it would also be good to begin normalizing image repository URLs. It is relatively recently that we began normalizing those and that isn't currently accounted for in this cred-searching code.

@fykaa please DM me with any questions, but I think this should be pretty straightforward.

krancour:
[...]

PR #4852, opened 2025-08-13
fix: account for normalization when matching regex to urls

fuskovic:
Closes: https://github.com/akuity/kargo/issues/4833


internal/credentials/kubernetes/database.go
```diff
@@ -232,7 +238,7 @@ func (k *database) getCredentialsSecret(
 		case credentials.TypeHelm:
 			secretURL = helm.NormalizeChartRepositoryURL(secretURL)
 		}
```
krancour, line 242: I think the switch above needs a new case for image credentials similar to line +194

krancour: Other than that, nailed it.

Issue #4839, opened 2025-08-11
feat(chart): opt out of chart creating/managing the "cluster secrets" namespace


krancour:
Derived from discussion in #4765

Discussion in #4765 aside, a likely reason for opting out of having this namespace created/managed by the chart would be the same reason that many users prefer Projects to "adopt" existing namespaces rather than having their underlying namespaces created by Kargo itself -- organizational requirements, policy engines, etc. might require namespaces to be labeled or annotated in certain ways.

Let's just add a new boolean defaulted to true that indicates whether the chart should create the "cluster secrets" namespace or not.

PR #4853, opened 2025-08-13
feat: allow option to opt out of cluster secrets namespace creation

fuskovic:
Closes https://github.com/akuity/kargo/issues/4839

### DIFF
--- a/charts/kargo/templates/api/configmap.yaml
+++ b/charts/kargo/templates/api/configmap.yaml
@@ -21,7 +21,9 @@ data:
   {{- if .Values.api.secretManagementEnabled }}
   SECRET_MANAGEMENT_ENABLED: "true"
   {{- end }}
+  {{- if .Values.global.createClusterSecretsNamespace }}
   CLUSTER_SECRETS_NAMESPACE: {{ .Values.global.clusterSecretsNamespace }}
+  {{- end }}
   PERMISSIVE_CORS_POLICY_ENABLED: {{ quote .Values.api.permissiveCORSPolicyEnabled }}
   {{- if .Values.api.adminAccount.enabled }}
   ADMIN_ACCOUNT_ENABLED: "true"
--- a/charts/kargo/templates/api/configmap.yaml
+++ b/charts/kargo/templates/api/configmap.yaml
@@ -21,9 +21,7 @@ data:
   {{- if .Values.api.secretManagementEnabled }}
   SECRET_MANAGEMENT_ENABLED: "true"
   {{- end }}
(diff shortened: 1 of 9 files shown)


charts/kargo/templates/api/configmap.yaml
```diff
@@ -21,7 +21,9 @@ data:
   {{- if .Values.api.secretManagementEnabled }}
   SECRET_MANAGEMENT_ENABLED: "true"
   {{- end }}
+  {{- if .Values.global.createClusterSecretsNamespace }}
   CLUSTER_SECRETS_NAMESPACE: {{ .Values.global.clusterSecretsNamespace }}
+  {{- end }}
```
krancour: We don't want this to be conditional.

It's plausible that the operator or some automation they're leveraging has created the namespace "out of band" and they therefore neither need nor desire for the chart to _manage_ the namespace. But in such a case, the namespace may not be named `kargo-cluster-secrets` and the flexibility to specify the name of the namespace (and have it make its way into the `CLUSTER_SECRETS_NAMESPACE`) will still be required.

(Aside: I'm not concerned with the edge case where someone doesn't want such a namespace to exist at all. As far as I am concerned, it's required for Kargo to function properly and if someone really wants that namespace to not exist, they're essentially opting into handicapping all the features that rely on it -- at this time that's only cluster webhook receivers. i.e. I'm unconcerned, for now, with any sort of graceful failure in that edge case.)

Removal of this conditional applies in some other templates in this PR as well.

charts/kargo/values.yaml
```diff
@@ -19,7 +19,8 @@ image:
 global:
   ## @param global.clusterSecretsNamespace Indicates a namespace where Secrets associated with cluster-scoped resources can be located.
   clusterSecretsNamespace: kargo-cluster-secrets
-
+  ## @param global.createClusterSecretsNamespace Indicates whether the `clusterSecretsNamespace` should be created by Kargo.
```
krancour: Especially since Kargo controllers do manage a lot of resources, including namespaces (Project namespaces, specifically) I think the vagueness of "by Kargo" opens the door to some potential confusion. We can sidestep that by being more clear that this setting only affecting behavior of the chart itself.

```suggestion
  ## @param global.createClusterSecretsNamespace Indicates whether the `clusterSecretsNamespace` should be managed by the chart.
```

krancour:
Need to rerun codegen, but preemptively LGTM.

johnpekcan:
Any chance we can get this out sooner than 1.8.0? 😅  @krancour

Issue #3989, opened 2025-05-06
document how to access vars, step outputs, etc. in http step's success/failure expressions


dbason:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

This may be an incorrect assumption on my part but I would have thought vars and outputs were available in the `successExpression` and `failureExpressions` runs in the http step

# Screenshots

![Image](https://github.com/user-attachments/assets/cb982e40-fe2f-42e9-920d-07a688c52861)

# Steps to Reproduce

Use the `vars` or `outputs` objects/maps in the expression fields e.g:
```yaml
failureExpression: response.body.stage.status.lastPromotion.freight.name == vars.freightName && response.body.stage.status.lastPromotion.status.phase == "Errored"
successExpression: response.body.stage.status.lastPromotion.freight.name == vars.freightName && response.body.stage.status.lastPromotion.status.phase == "Succeeded"
```

# Version
Not using the kargo CLI but version if v.1.4.1

# Logs

```
Paste any relevant application logs here.
```

dbason:
For context I'm trying to create a sort of fan in step where we evaluate the result of the promotion from all of the previous fanout steps.  We only want to advance a pipeline if all of the fanout steps succeeded (this is for regional data centres).  We're choosing to fan out in this situation to allow the regional datacentres to be deployed to in parallel

krancour:
> This may be an incorrect assumption on my part

It is, but it's a forgiveable one. 🤪

Those two fields are specifically for defining an expression that picks apart the response. 

You'll notice that the docs instruct you _not_ to offset those expressions with `${{ }}` as you do in other places. Everywhere where you see `${{ }}`, those expressions are evaluated _before_ configuration is passed to a step. i.e. Steps only see pre-evaluated config and are entirely unaware of expressions.

You're instructed to avoid `${{ }}` here because this is a separate / secondary usage of our expression evaluation capability that is performed by the step itself after having received a response. The context used in evaluating those expressions is really nothing except the response.

There's an easy solution to this, however, which should probably be documented:

```
failureExpression: response.body.stage.status.lastPromotion.freight.name == '${{ vars.freightName }}' && response.body.stage.status.lastPromotion.status.phase == "Errored"
[...]

PR #4855, opened 2025-08-13
docs: how to access `vars` & `outputs` in `http` step's success/failure expressions

fykaa:
closes https://github.com/akuity/kargo/issues/3989


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/http.md
```diff
@@ -73,6 +73,36 @@ is structured as follows:
...
+```
+
+At runtime, `${{ vars.expectedStatus }}` will be replaced with its actual value
+(`'completed'`), so the `http` step's expression sees:
+
+```yaml
+successExpression: response.body.status == 'completed'
+```
+:::
+
```
krancour: A bit higher up, there's an existing note box explaining that `successExpression` and `failureExpression` should not be enclosed in `${{ }}` and it gives a partial explanation of why. I think that's the right place to insert this new information, and that the existing explanation can be improved and expanded ever so slightly as a segue to your example.

krancour (approved):

Issue #4835, opened 2025-08-11
Introduce Stage `status.metadata` along with setter/getters


jessesuen:
# Proposed Feature

Spawned from https://github.com/akuity/kargo/issues/4702

This proposal is to:

1. Introduce a `status.metadata` on a Stage to hold user-defined metadata
2. Add the ability to set or read this metadata on a stage in a promotion step with a `set-stage-metadata` step and `stageMetadata()` helper respectively.
3. Add the ability to read this metadata when invoking an AnalysisTemplate, and in promotion steps.

NOTE: we already have a `freightMetadata()` helper, so if we are following that precedent, we would add a `stageMetadata()` helper.

# Motivation

Verification is a process that is considered independent of promotion, and we don't have a good way of passing along information between those two processes. By introducing `status.metadata`, the promotion could set metadata on a Stage itself, so that it can be read by another process (e.g. verification or even a subsequent promotion).

jessesuen:
Another use case for Stage metadata, is that this information could be something users could configure to show in the UI. This would allow a level of customizability of what Kargo shows in the pipeline. We can/should spawn a different or follow-up issue for the UI support if we think it's a good idea.

krancour:
Closely related to #4837 and quite probably to be tackled in unison, so assigning to @fykaa. I will assist with both as needed.

Issue #4837, opened 2025-08-11
Introduce `set-freight-metadata` Step


jessesuen:
# Proposed Feature

[A while back](https://github.com/akuity/kargo/pull/4272), we introduced the `status.metadata` field on a Freight along with `freightMetadata()` helper to access it. Currently, the ability to set Freight metadata is only available through built-in kargo steps, such as the JIRA step.

It would be useful for end-users to manage Freight metadata as part of their promotion process by introducing a `set-freight-metadata` step. 

# Motivation

Adding the ability to perform bookkeeping at the Freight level will open up more interesting use cases. For example, with the JIRA step, it's possible to update an existing issue that is tied to a release with the ["Multi-Stage Workflow"](https://main.docs.kargo.io/user-guide/reference-docs/promotion-steps/jira/#multi-stage-workflow). With a `set-freight-metadata` step, this use case could be replicated using http steps.

krancour:
This should have been closed by #4861

PR #4861, opened 2025-08-14
feat: `set-metadata` promo step, deprecate two-arg `freightMetadata` format, add `stageMetadata` handling

fykaa:
closes https://github.com/akuity/kargo/issues/4835

this PR does the following things:
- deprecate two-argument form of the `freightMetadata()` function
- introduce `stageMetadata` expression function
- addition of metadata map to `StageStatus`
- doc changes to jira.md ragarding `freightMetadata()`
- updates tests for `freightMetadata()` and adds tests for `stageMetadata()`
- adds a new promotion step `set-metadata` , and docs/ tests for the same


internal/expressions/function/functions.go
```diff
@@ -217,6 +232,70 @@ func freightMetadata(
...
+			return nil, fmt.Errorf("stage name must not be empty")
+		}
+
+		key, ok := a[1].(string)
+		if !ok {
+			return nil, fmt.Errorf("second argument must be string, got %T", a[1])
+		}
+		if key == "" {
+			return nil, fmt.Errorf("metadata key must not be empty")
+		}
```
krancour: Based on reading the implementation, I can infer usage in an expression looks something like `stageMetadata('test', 'my-info')`.

I'm wondering if UX would be nicer by taking just one argument, like so, `stageMetadata('test')`, and just returning the whole metadata map. The user can access info within the map using expressions like `stageMetadata('test')['my-info']` or `stageMetadata('test').foobar`.

jessesuen: +1 to what @krancour said. 

Unfortunately, I just noticed `freightMetadata()` accepts a second parameter, so I think @fykaa was just following its lead.  Is it too late to change `freightMetadata()` behavior, or make the second parameter optional? I would like these two helper to be symmetrical.

krancour: I forgot `freightMetadata()` was already implemented. It would be non-breaking to make the second parameter optional in `freightMetadata()`. For symmetry, we'd have to make it optional here as well. Examples and docs can favor the one param form.

jessesuen: I would just make `freightMetadata()` a single argument, in case we want to reserve the second optional parameter for some other purpose.

Agree we should stop documenting the second parameter for `freightMetadata()` and even deprecate it for symmetry.

fykaa: to keep the changes backward-compatible and symmetrical, we can have both `freightMetadata()` and `stageMetadata()` accept either one or two arguments. If called with a single argument, they return the entire metadata map; if called with two, they return the value for the specified key

documentation and examples can favor the single-argument form.
and we add a deprecation note for the second parameter in docs.

krancour: @jessesuen I'll DM you about this.

krancour: @jessesuen and I spoke offline.

@fykaa here's the direction we'll head in, please:

1. Make the second argument to `freightMetadata()` optional. When only one arg is provided, only return the map.

1. List the second arg as deprecated in the v1.8 release notes. We will remove the optional second arg in v1.10. (We typically remove deprecated things two minors later.)

1. Adjust docs to _only_ address the usage with the single argument. We don't want to promote the usage we're deprecating.

1. Implement `stageMetadata()` without the second argument from the beginning.

i.e. We'll do `stageMetadata()` the way we prefer right now and make `freightMetadata()` match it over time.

Also fwiw, @jessesuen believes the blast radius for this breaking change is very small.

krancour:
@fykaa please address #4837 in this same PR.

internal/expressions/function/functions.go
```diff
@@ -23,10 +24,10 @@ type exprFn func(params ...any) (any, error)
...
 // Freight operations.
 //
-// It provides `warehouse()`, `commitFrom()`, `imageFrom()`, and `chartFrom()`
-// functions that can be used within expressions. The functions operate within
-// the context of a given project with the provided freight requests and
-// references.
+// It provides `warehouse()`, `commitFrom()`, `imageFrom()`, `chartFrom()` and
+// `stageMetadata()` functions that can be used within expressions. The functions
+// operate within the context of a given project with the provided freight requests
+// and references.
```
krancour: This change does not appear accurate, so I would recommend reverting it:

```suggestion
// It provides `warehouse()`, `commitFrom()`, `imageFrom()`, and `chartFrom()`
// functions that can be used within expressions. The functions operate within
// the context of a given project with the provided freight requests and
// references.
```

internal/expressions/function/functions.go
```diff
@@ -43,6 +44,21 @@ func FreightOperations(
...
+// to access metadata stored on Stage resources.
+func StageOperations(
+	ctx context.Context,
+	c client.Client,
+	project string,
+) []expr.Option {
+	return []expr.Option{
+		StageMetadata(ctx, c, project),
+	}
+}
```
krancour: A few problems here. You've probably created this for "symmetry" with the `FreightOperations()` function above, but...

1. You're not calling this from anywhere, so the functions this is supposed to make available for use in expressions (within certain contexts) are not being made available for use in any expressions anywhere.

1. I actually am not sure that I love the setup for the `freightMetadata()` expression function having been included in `FreightOperations()` function to begin with. i.e. You're emulating something that probably shouldn't have been done in the first place.

  Before the call to `FreightMetadata()` had been added to `FreightOperations()`, `FreightOperations()` was focused on registering expression functions that had to do with _finding artifacts within the specific collection of Freight involved in a promotion_. `freightMetadata()`, by contrast is for the relatively straightforward retrieval of data from a known source.

  Bottom line: I think the setup for the `freightMetadata()` expression function (i.e. the call to `FreightMetadata()`) would have found a better home in `DataOperations()` function. I also think that's the appropriate place for the new `stageMetadata()` expression function to be set up (i.e. the appropriate place to call `StageMetadata()`).

So, please:

1. Remove the call to `FreightMetadata()` from `FreightOperations()` and add it to `DataOperations()`.
2. Remove the `StageOperations()` function.
3. Call `StageMetadata()` from `DataOperations()`.

api/v1alpha1/stage_types_test.go
```diff
@@ -972,3 +974,157 @@ func TestChartDeepEquals(t *testing.T) {
...
+
+	t.Run("get non-existent key", func(t *testing.T) {
+		status := StageStatus{}
+		var result string
+		found, err := status.GetMetadata("non-existent", &result)
+		require.False(t, found)
+		require.NoError(t, err)
+		require.Equal(t, "", result)
+	})
+}
```
krancour: I know you seem to be emulating how upserting and retrieval of Freight metadata had been tested, but I'd rather we not do so. There was some akwardness in there that I'd rather we not repeat. (And if possible, I'd like us to fix the existing problems.)

What is awkward is that upsert and get are being tested together as a pair instead of testing each in isolation. If we had test cases for _just_ upserting, it wouldn't be hard to assert that, post-upsert, the updated status appears as it ought to. Similarly, if we had test cases for _just_ getting, it wouldn't be hard to assert that for a given status, the getter retrieved the data that it ought to have. Structuring the tests in this way should also easily allow for engineering all kinds of specific scenarios and thereby negating the need for separate "edge case tests" in another function.

If you could please improve these tests, I'd like to see the same improvements extended to upserting and retrieving Freight metadata.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -353,12 +353,16 @@ Use the `freightMetadata()` template function to retrieve issue keys stored by p
 
 ```yaml
 # Access the default issue key
-issueKey: ${{ freightMetadata(ctx.targetFreight.name, 'jira-issue-key') }}
+issueKey: ${{ freightMetadata(ctx.targetFreight.name).jira-issue-key }}
```
krancour: Dashes in attributes are something to be careful of. expr allows it, but it feels extremely awkward because most languages wouldn't permit this.

I would suggest favoring this alternative syntax here:

```suggestion
issueKey: ${{ freightMetadata(ctx.targetFreight.name)['jira-issue-key'] }}
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
```diff
@@ -353,12 +353,16 @@ Use the `freightMetadata()` template function to retrieve issue keys stored by p
...
 
 # Access a custom issue key (when issueAlias was used during creation)
-issueKey: ${{ freightMetadata(ctx.targetFreight.name, 'my-custom-alias') }}
+issueKey: ${{ freightMetadata(ctx.targetFreight.name).my-custom-alias }}
 ```
 
+:::note
+The two-argument form (`freightMetadata(ctx.targetFreight.name, 'jira-issue-key')`) is deprecated as of `v1.8` and will be removed in `v1.10`. Please migrate to the single-argument form as shown above.
+:::
```
krancour: It feels like there ought to have been a better place to address this, such as the docs for the `freightMetadata()` function itself, but it turns out those don't exist. 🤦‍♂️. Let's fix that, please. Docs for `freightMetadata()` and `stageMetadata()` need to be added to `docs/docs/50-user-guide/60-reference-docs/40-expressions.md`

krancour:
@fykaa please also include the new steps for updating Freight and Stage metadata in this PR.

internal/expressions/function/functions.go
```diff
@@ -161,7 +171,8 @@ func FreightMetadata(
 	return expr.Function(
 		"freightMetadata",
 		freightMetadata(ctx, c, project),
-		new(func(freightRefName, key string) any),
+		new(func(freightRefName string) map[string]any),
+		new(func(freightRefName, key string) any), // Deprecated
```
krancour: [Learned the hard way](https://github.com/akuity/kargo/pull/2998) that when expression functions have overloaded signatures, they should be arranged from longest/most complex to shortest/least complex. It may not matter here, but I have dug deep enough to know if it does or doesn't. I would play it safe and swap the order of these.

internal/expressions/function/functions_test.go
```diff
@@ -1159,7 +1165,16 @@ func Test_freightMetadata(t *testing.T) {
 		assertions func(t *testing.T, result any, err error)
 	}{
 		{
-			name:    "successful metadata retrieval - string map",
+			name:    "successful metadata retrieval, single arg - string map",
```
krancour: Not a blocker, but something I strongly recommend: When using whitebox testing, as we are here, meaning the tests have knowledge of how the thing they are testing is constructed, it is helpful to those who come along later to review or modify this code if test cases are laid out in the same order in which they'd be dealt with in the code under tests. In this case, it would probably mean starting with cases like too many or two few arguments and ending with a success case involving two arguments.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/jira.md
krancour, line 1: Very nicely done, and thank you also for all the formatting fixes!

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -371,9 +371,54 @@ relevant `Freight` is not found from the `FreightCollection`.
...
+  # Access metadata values using ['key-name'] syntax
+  issueKey: ${{ freightMetadata(ctx.targetFreight.name)['jira-issue-key'] }}
+  customValue: ${{ freightMetadata(ctx.targetFreight.name)['my-custom-key'] }}
+```
+
+:::note
+The two-argument form (`freightMetadata(ctx.targetFreight.name, 'jira-issue-key')`)
+is deprecated as of `v1.8` and will be removed in `v1.10`. Please migrate to the
+single-argument form as shown above.
+:::
```
krancour: The definite article "the" implies familiarity, making "the two-argument form" sound a little bit as if this is something that was previously discussed, but it wasn't. I would consider rewording things a bit here, to something like "An optional second argument is supported, but is deprecated as of ..." It might also be helpful to, briefly, note how the return values of the two argument and one argument forms differ. It will be helpful to those who do need to migrate.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -371,9 +371,54 @@ relevant `Freight` is not found from the `FreightCollection`.
...
+
+Example:
+
+```yaml
+config:
+  # Access metadata values using ['key-name'] syntax
+  envType: ${{ stageMetadata(ctx.stage)['environment-type'] }}
+  region: ${{ stageMetadata(ctx.stage)['aws-region'] }}
+```
+
```
krancour, line 438: Since I think these functions have good odds of sometimes returning nil, expanding the examples to briefly demonstrate expr-lang features like [optional chaining](https://expr-lang.org/docs/language-definition#optional-chaining) and [nil coalescing](https://expr-lang.org/docs/language-definition#nil-coalescing) would be extraordinarily helpful. There's no need to do a deep dive. You can also link to the language definition as well.

krancour: Same for `freightMetadata()` obviously.

krancour:
This PR is looking great. Still need to add the steps, but everything that's already here looks fantastic. My recommendations so far are very few and very small. 🔥

internal/promotion/runner/builtin/schemas/set-metadata-config.json
```diff
@@ -0,0 +1,39 @@
...
+            "type": "string",
+            "description": "Name of the resource to update metadata for",
+            "minLength": 1
+          },
+          "values": {
+            "type": "object",
+            "description": "Key-value pairs to set as metadata on the resource",
+            "minProperties": 1,
+            "additionalProperties": {
+              "type": ["string", "number", "boolean", "null"]
```
krancour: Is it deliberate that this rules out objects? I think the underlying code would support it just fine? Or am I overlooking something?

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -371,7 +371,78 @@ relevant `Freight` is not found from the `FreightCollection`.
...
+```yaml
+config:
+  # Access metadata values using ['key-name'] syntax
+  issueKey: ${{ freightMetadata(ctx.targetFreight.name)['jira-issue-key'] }}
+
+  # Using nil coalescing (??) to provide default values if metadata is missing
+  tier: ${{ stageMetadata(ctx.stage)['tier'] ?? "default-tier" }}
+
+  # Using optional chaining (?.) with nil coalescing for nested values
+  nested: ${{ stageMetadata(ctx.stage)?.config?.settings?.timeoutSeconds ?? 300 }}
```
krancour: This example feels confusing because it starts out with `issueKey`, looking very much like real-world usage of the Jira step, but then deviates from that with `tier` and `nested` looking like more contrived/constructed/nonsensical examples.

I _don't_ actually have a problem with contrived/constructed/nonsensical examples. It keeps the focus on the expressions instead of the step in which the expressions are being used and gets the point across pretty well, I think. So in that regard, I rather like that, and there are plenty of other such examples throughout the docs.

The thing that I think will end up confusing though is the mixing and matching of a practical, real-world example (Jira step / `issueKey`) with the more contrived examples, and especially since the Jira step is essentially behind a paywall, I think the above example will feel more approachable/accessible if we _don't_ involve it.

My suggesting is the update line +393 to be more like the other two examples -- more generic.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -371,7 +371,78 @@ relevant `Freight` is not found from the `FreightCollection`.
...
+```yaml
+config:
+  # Access metadata values using ['key-name'] syntax
+  region: ${{ stageMetadata(ctx.stage)['aws-region'] }}
+
+  # Using nil coalescing (??) to provide default values if metadata is missing
+  tier: ${{ stageMetadata(ctx.stage)['tier'] ?? "default-tier" }}
+
+  # Using optional chaining (?.) with nil coalescing for nested values
+  nested: ${{ stageMetadata(ctx.stage)?.config?.settings?.timeoutSeconds ?? 300 }}
```
krancour, line 436: Just a note that _this_ is a great example that addresses my worries [in this other comment](https://github.com/akuity/kargo/pull/4861/files#r2310549389).

internal/expressions/function/functions.go
```diff
@@ -171,29 +182,19 @@ func freightMetadata(
 	project string,
 ) exprFn {
 	return func(a ...any) (any, error) {
-		if len(a) != 2 {
-			return nil, fmt.Errorf("expected 2 argument, got %d", len(a))
+		if len(a) != 1 && len(a) != 2 {
+			return nil, fmt.Errorf("expected 1 argument, got %d", len(a))
```
krancour, line 189: Just a note, I love that this says "expected 1" instead of "expected 1 or 2," as I do not like the idea of presenting an option that drives people toward already-deprecated usage. Assuming that was intentional, nicely done!

internal/expressions/function/functions_test.go
```diff
@@ -1159,124 +1165,258 @@ func Test_freightMetadata(t *testing.T) {
 		assertions func(t *testing.T, result any, err error)
 	}{
 		{
-			name:    "successful metadata retrieval - string map",
-			objects: []client.Object{testFreight},
-			args:    []any{testFreightName, "deployment-config"},
+			name: "no arguments",
```
krancour, line 1165: I appreciate you taking my advice. It's much easier now to follow how each of these exercise all the logical branches in the code under test.

internal/expressions/function/functions_test.go
```diff
@@ -1159,124 +1165,258 @@ func Test_freightMetadata(t *testing.T) {
...
+			Namespace: testProject,
+		},
+		Status: kargoapi.StageStatus{},
+	}
+
+	// Add metadata to the stage status
+	err := testStage.Status.UpsertMetadata("deployment-config", testMetadata)
+	assert.NoError(t, err)
+
+	tests := []struct {
```
krancour, line 1346: Would you consider reordering these similarly to how you reordered the test cases for `FreightMetadata()`?

internal/expressions/function/functions_test.go
```diff
@@ -1159,124 +1165,258 @@ func Test_freightMetadata(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testStageName,
+			Namespace: testProject,
+		},
+		Status: kargoapi.StageStatus{},
+	}
+
+	// Add metadata to the stage status
+	err := testStage.Status.UpsertMetadata("deployment-config", testMetadata)
+	assert.NoError(t, err)
```
krancour: I know you're emulating existing code in `Test_stageMetadata()`, but it would be preferable not to rely on `UpsertMetadata()` here.

Ideal unit tests ("unit" meaning "one") isolate the code under test as best as possible from _other code_. This way, when a test fails, you can assume it is failing on account of the code you're testing and _not_ on account of some other code that you're not actively testing. This is the reason unit tests so extensively leverage mocks and fake implementations of interfaces or functions.

If you hardcoded the metadata here instead of relying on `UpsertMetadata()` to function correctly (it has its own test elsewhere to ensure it does), you would know immediately that any failures in these test cases can be pinned on the `stageMetadata()` implementation and not anything else.

If you'd be so kind, I'd love if the existing tests for `freightMetadata()` we also amended to use hardcoded metadata instead of relying on `UpsertMetadata()`.

internal/promotion/runner/builtin/schemas/set-metadata-config.json
krancour, line 1: Requesting @jessesuen's buy-in on this schema.

I think it's pretty good, but as constructed, it would require multiple uses of this step if you needed to update the metadata of multiple resources. You could not, for instance, update the metadata of (arbitrary example) a piece of Freight and _two_ Stages all at once. You'd need three separate steps to do that.

I am fine with that, but want to make sure it comports with @jessesuen's notions of what the UX ought to be here.

internal/promotion/runner/builtin/metadata_setter.go
```diff
@@ -0,0 +1,148 @@
...
+func (s *metadataSetter) convert(cfg promotion.Config) (builtin.SetMetadataConfig, error) {
+	return validateAndConvert[builtin.SetMetadataConfig](s.schemaLoader, cfg, s.Name())
+}
+
+func (s *metadataSetter) run(
+	ctx context.Context,
+	_ *promotion.StepContext,
+	cfg builtin.SetMetadataConfig,
+) (promotion.StepResult, error) {
+	for _, update := range cfg.Updates {
```
krancour, line 64: This loop has a significant inefficiency. You're retrieving the resource that is to be updated over and over again _within_ the loop -- once for each update made to the _one_ resource.

It would be much more efficient to retrieve the Freight or the Stage _once_ up-front before you begin looping over the various changes that need to be made to that resource.

Note: You could create a one-off interface like `metadataUpsertable` here within this function. Both Freight and Stages _already_ implement that. So you could have just a single switch up-front, before the loop to retrieve Freight or Stage, assigning what you find to a variable of type `metadataUpsertable` and that can be what the loop operates on.

internal/expressions/function/functions.go
fykaa: yes, this was intentional

internal/promotion/runner/builtin/schemas/set-metadata-config.json
fykaa: the underlying code would support complex objects just fine unless there's a specific reason to restrict to primitive values only, I have updated this to `"additionalProperties": true`

internal/promotion/runner/builtin/schemas/set-metadata-config.json
jessesuen: The spec makes me think otherwise. The `updates` field is a list, so wouldn't something like this be supported?

```yaml
steps:
- uses: set-metadata
  config:
    updates:
    - kind: Freight
      name: abc123
      values:
        aa: bb
    - kind: Stage
      name: def456
      values:
        cc: dd

```

jessesuen: @fykaa - Is my understanding correct?

fykaa: Yes, your observation is correct. `set-metadata` step is already designed to handle multiple updates in a single step through the updates array in its configuration.

there is also a "valid kitchen sink" test case that verifies this capability

and the current implementation processes the updates array in a loop

internal/promotion/runner/builtin/metadata_setter.go
fykaa: i agree with the inefficiency you pointed out, this was a draft, and I was primarily focused on getting the basic logic and flow working first. I see that retrieving the resource repeatedly in the loop is unnecessary, and using a common interface like `metadataUpsertable` to fetch it once up-front makes a lot of sense. I'll work on refactoring it that way

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -371,7 +371,78 @@ relevant `Freight` is not found from the `FreightCollection`.
...
+
+This returns a map containing all metadata key-value pairs stored in the `Freight`
+resource.
+
+Example:
+
+```yaml
+config:
+  # Access metadata values using ['key-name'] syntax
+  category: ${{ freightMetadata(ctx.targetFreight.name)['deployment-category'] }}
```
krancour: I know this is very picky and that I also sound like a broken record, but I prefer to avoid the word "deployment" at all costs (even in contrived examples) so that we continue to emphasize Kargo's role as a "promoter" and not a "deployer."

This change won't reduce the value of the example any:

```suggestion
  category: ${{ freightMetadata(ctx.targetFreight.name).category }}
```

I also changed the syntax, because with the dash removed from the key, the dot syntax seems more straightforward. The next example does show the alternative, square bracket syntax and we'll leave that be so that we maintain some variety in the examples.

api/v1alpha1/freight_types_test.go
```diff
@@ -470,147 +468,208 @@ func TestFreightStatus_AddApprovedStage(t *testing.T) {
...
 
-			case bool:
-				var result bool
-				found, err := status.GetMetadata(key, &result)
-				assert.True(t, found)
-				assert.NoError(t, err)
-				assert.Equal(t, expected, result)
+	testCases := []struct {
+		name         string
+		metadata     map[string]apiextensionsv1.JSON
```
krancour, line 571: You're using the same metadata in every test case, so there's really no need for `metadata` to be a field of the test case struct.

krancour: Same applies to the similar tests for Stage metadata retrieval.

api/v1alpha1/freight_types_test.go
```diff
@@ -470,147 +468,208 @@ func TestFreightStatus_AddApprovedStage(t *testing.T) {
...
+	}
+}
 
-			case int:
-				var result int
-				found, err := status.GetMetadata(key, &result)
-				assert.True(t, found)
-				assert.NoError(t, err)
-				assert.Equal(t, expected, result)
+func TestFreightStatus_GetMetadata(t *testing.T) {
```
krancour, line 560: I'll start with the disclaimer that I know a lot of what's in this PR is emulating existing things, fixing existing things that should not be emulated, etc.

I spent a lot of time looking at this test. Trying too hard to make these tests table-driven is complicating them. Having a `target` field of type `any` in the test cases is forcing you (and others reading the code) to grapple more than necessary with pointers and esoteric things like what happens when you unmarshal into a variable whose declared type is `any`, but whose concrete type is something else. This has also led you to write the `derefAny()` function, which was actually the "smell" that sent me looking much more closely at all of this.

In short, these test cases don't model how the method is actually used.

If you structure the tests in a non-table-driven fashion, more like the following, the tests will be simpler and will _much_ more closely model _potential_ real-world usage of the `GetMetadata()` method:

```golang
status := FreightStatus{
	Metadata: map[string]apiextensionsv1.JSON{
		"string-key": {Raw: []byte(`"test-value"`)},
		"int-key":    {Raw: []byte(`42`)},
		"bool-key":   {Raw: []byte(`true`)},
		"slice-key":  {Raw: []byte(`[1,2,3]`)},
		"struct-key": {Raw: []byte(`{"name":"test-user","age":30,"active":true}`)},
	},
}

t.Run("get string value", func(t *testing.T) {
	var value string
	found, err := status.GetMetadata("string-key", &value)
	require.NoError(t, err)
	require.True(t, found)
	require.Equal(t, "test-value", value)
})

t.Run("get int value", func(t *testing.T) {
	var value int
	found, err := status.GetMetadata("int-key", &value)
	require.NoError(t, err)
	require.True(t, found)
	require.Equal(t, 42, value)
})

// ...
```

Now, notice I said "potential" real-world use. Something else that's jumped out at me is that test cases like those above do not actually emulate the _one_ way you're actually using this method. This is from `functions.go`:

```golang
var data any
found, err := freightData.Status.GetMetadata(key, &data)
```

This got me thinking more deeply about the method's signature. If _this_ is how the method is being used, a signature like `GetMetadata(key string) (any, err)` would be more straightforward. I almost suggested refactoring in this way, then I thought better of it...

I think it's reasonable to suppose that _we_ may use this function eventually in ways other than enabling the `getFreightMetadata()` expression function. It's totally conceivable that we will eventually want to use this to unmarshal metadata into strongly typed variables. So, I think the implementation of the `GetMetadata()` function is fine and the test cases do indeed test _potential_ (future) real-world use...

However, your test cases are _not_ testing the one way you're _actually_ using this currently. So, I would encourage you to also add a few test cases that look something like these:

```golang
t.Run("get any value -- string", func(t *testing.T) {
	var value any
	found, err := status.GetMetadata("string-key", &value)
	require.NoError(t, err)
	require.True(t, found)
	require.Equal(t, "test-value", value)
})

t.Run("get any value -- map", func(t *testing.T) {
	var value any
	found, err := status.GetMetadata("struct-key", &value)
	require.NoError(t, err)
	require.True(t, found)
	require.Equal(
		t,
		map[string]any{
			"name":   "test-user",
			"age":    float64(30),
			"active": true,
		},
		value,
	)
})
```

I know this is a lot, but I will feel much more confident in the implementation of the `GetMetadata()` method if the test cases are more realistic.

All of this applies equally to the tests for the `StageStatus` type's `GetMetadata()` method.

internal/promotion/runner/builtin/schemas/set-metadata-config.json
krancour: My apologies. I definitely misread the schema. All good.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
+---
+sidebar_label: set-metadata
+description: Updates metadata on Stage or Freight resources during the promotion process.
+---
+
+# `set-metadata`
+
+`set-metadata` updates metadata on `Stage` or `Freight` resources during the
+promotion process. This step allows you to attach arbitrary key-value pairs to the
```
krancour: Nit: I think this is the more conventional way to denote key/value:

```suggestion
promotion process. This step allows you to attach arbitrary key/value pairs to the
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+status of these resources, which can be useful for tracking deployment information,
```
krancour: Reword this somehow, please to not use the word "deployment." Kargo does not deploy.

```suggestion
status of these resources, which can be useful for tracking deployment information,
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+version details, or any other relevant metadata.
+
+## Configuration
+
+| Name               | Type       | Required | Description                                                                                                                                                                                                 |
+| ------------------ | ---------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `updates`          | `[]object` | Y        | List of metadata updates to apply to various resources.                                                                                                                                                      |
```
krancour:
```suggestion
| `updates`          | `[]object` | Y        | List of metadata updates to apply.                                                                                                                                                      |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+| `updates[].kind`   | `string`   | Y        | Kind of resource to update metadata for. Must be either "`Stage`" or "`Freight`".                                                                                                                            |
```
krancour: This looks odd.

```suggestion
| `updates[].kind`   | `string`   | Y        | Kind of resource to update metadata for. Must be either `Stage` or `Freight`.                                                                                                                            |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+| `updates[].name`   | `string`   | Y        | Name of the resource to update metadata for.                                                                                                                                                                 |
+| `updates[].values` | `object`   | Y        | Key-value pairs to set as metadata on the resource. Must contain at least one key-value pair. This field supports various types of values including strings, numbers, booleans, arrays, and nested objects. |
```
krancour:
```suggestion
| `updates[].values` | `object`   | Y        | Key/value pairs to set as metadata on the resource. Must contain at least one key/value pair. This field supports various types of values including strings, numbers, booleans, arrays, and nested objects. |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Examples
+
+### Common Usage
+
+In this example, metadata is added to both a `Stage` and a `Freight` resource. This
+pattern is commonly used to track deployment information or add context about the
```
krancour: Don't say "deployment."

```suggestion
pattern is commonly used to track deployment information or add context about the
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+              components:
+                - name: "frontend"
+                  version: "2.1.0"
+                  status: "healthy"
+                - name: "backend"
+                  version: "1.5.0"
+                  status: "healthy"
+            environment:
+              region: "us-west-2"
+              cluster: "staging-01"
```
krancour: Reminder that we've standardized on this syntax throughout the code and docs.

Additionally:

* Don't use the word "deployment"
* It feels strange to track things like version information and health in metadata since Kargo already continuously tracks these things. Some kind of nonsense example would probably be just fine. foo and bar get the point across and won't ever trip anyone up thinking about whether it's realistic usage and something they should replicate in their own promotion processes.

```suggestion
steps:
- uses: set-metadata
  config:
    updates:
    - kind: Stage
      name: staging
      values:
        deployment:
          timestamp: "xyz"
          components:
          - name: "frontend"
            version: "2.1.0"
            status: "healthy"
          - name: "backend"
            version: "1.5.0"
            status: "healthy"
        environment:
          region: "us-west-2"
          cluster: "staging-01"
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Using Metadata in Other Contexts
+
+Once metadata is set (as shown in the examples above), it can be retrieved using
+[`freightMetadata()`](../40-expressions.md#freightmetadatafreightname) and
+[`stageMetadata()`](../40-expressions.md#stagemetadatastagename) expressions in other
+steps or verification processes. To access the metadata in different contexts:
```
krancour:
```suggestion
Once metadata is set (as shown in the examples above), it can be retrieved using
[`freightMetadata()`](../40-expressions.md#freightmetadatafreightname) and
[`stageMetadata()`](../40-expressions.md#stagemetadatastagename) functions
available within expressions used in other promotion steps or in verification
processes.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+                  status: "healthy"
+                - name: "backend"
+                  version: "1.5.0"
+                  status: "healthy"
+            environment:
+              region: "us-west-2"
+              cluster: "staging-01"
+```
+
+### Using Metadata in Other Contexts
```
krancour: Avoiding "other" because we haven't yet shown how to use it at all.

```suggestion
### Using Metadata
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+            ctx.stage,
+            environment.region,
+            len(deployment.components),
+            deployment.components[0].name,
+            deployment.components[0].version,
+            deployment.components[1].name,
+            deployment.components[1].version
+          )
+        }}
+```
```
krancour: Since you've linked to the expression docs above, I would leave it to those pages to cover use of these functions.

```suggestion
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -371,7 +371,78 @@ relevant `Freight` is not found from the `FreightCollection`.
...
+
+### `freightMetadata(freightName)`
+
+The `freightMetadata()` function retrieves metadata stored in a `Freight` resource.
+It has one required argument:
+
+- `freightName` (Required): The name of the `Freight` resource
+
+This returns a map containing all metadata key-value pairs stored in the `Freight`
+resource.
```
```suggestion
The `freightMetadata()` function retrieves the map of all metadata stored in a
`Freight` resource. It has one required argument:

- `freightName` (Required): The name of the `Freight` resource
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
```diff
@@ -0,0 +1,121 @@
...
+          name: production
+          values:
+            deployedBy: "user@example.com"
+            version: "1.0.0"
+        - kind: Freight
+          name: my-app-freight
+          values:
+            deployed: true
+            deploymentStatus: "success"
+```
```
krancour: Fix array syntax and don't say "deploy."

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-metadata.md
krancour: I found "key-value" in a lot of spots, but probably missed some. Please be on the lookout for it.

api/v1alpha1/freight_types_test.go
fykaa: I initially chose the table-driven approach to maintain consistency with the codebase's existing patterns, I agree that in this case it led to complexity, especially around type handling. I've refactored both `FreightStatus` and `StageStatus` GetMetadata tests to use discrete test cases. that was able to remove the `derefAny()` helper

krancour: Thank you. More important than the removal of the helper, the resulting tests more closely resemble how the method would actually be used.

fwiw, table-driven tests are often convenient, but we don't need to dogmatically follow that pattern when it gets our way.

Thanks again for fixing this.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -371,7 +371,75 @@ relevant `Freight` is not found from the `FreightCollection`.
...
+`Freight` resource. It has one required argument:
+
+- `freightName` (Required): The name of the `Freight` resource
+
+Example:
+
+```yaml
+config:
+  # Access metadata values using ['key-name'] syntax
+  category: ${{ freightMetadata(ctx.targetFreight.name).category }}
```
krancour: Whoops. My bad. You took my suggestion, but the comment is not accurate now.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -371,7 +371,75 @@ relevant `Freight` is not found from the `FreightCollection`.
...
+[optional chaining](https://expr-lang.org/docs/language-definition#optional-chaining)
+features.
+:::
+
+:::note
+An optional second argument (`freightMetadata(freightName, 'key-name')`) is supported
+but deprecated as of `v1.8` and will be removed in `v1.10`. While the two-argument
+form returns a single value for the specified key, the single-argument form returns
+the complete metadata map. To migrate, use the single-argument form with map access
+syntax: `freightMetadata(freightName)['key-name']`.
```
krancour: The other syntax works too, as you know. I don't want this statement to mislead people into thinking otherwise (despite your many excellent examples to the contrary).

internal/promotion/runner/builtin/metadata_setter.go
```diff
@@ -0,0 +1,147 @@
...
+		if _, exists := updatesByResource[key]; !exists {
+			updatesByResource[key] = &updateGroup{
+				kind:   string(u.Kind),
+				values: make(map[string]any),
+			}
+		}
+		for k, v := range u.Values {
+			updatesByResource[key].values[k] = v
+		}
+	}
```
krancour: I appreciate you fixing the problem I previously raised of Stages or Freight being re-retrieved for every values change, but I think we may now have over-optimized.

Correct me if I am wrong, but compared to simply ranging over `cfg.Updates` directly starting on line +89, doesn't this optimization only provide any benefit in the unusual case that someone had unnecessarily done this:

```yaml
config:
  updates:
    - kind: Stage
      name: my-stage
      values:
        foo: bar
    - kind: Stage
      name: my-stage
      values:
        bat: baz
```

instead of this:

```yaml
config:
  updates:
    - kind: Stage
      name: my-stage
      values:
        foo: bar
        bat: baz
```

?

krancour:
@fykaa thank you for your diligence on this. This is looking great, and you've cleaned up a lot of existing things that were a bit "off" in the process.

There are just a few very minor things left and then this LGTM.

One question, however: Can I assume you've seen the new step and the new expression functions work in an end-to-end example? If so, could you shared that example here to speed up my own test drive?

internal/promotion/runner/builtin/metadata_setter.go
fykaa: I used this approach for API efficiency, since it minimized `GET` API calls by grouping updates for same resources, and all metadata changes for a resource were applied in one `UPDATE` , it helped to batch operations on same reources.
But I get that it might increase memory complexity to O(unique resources), and users are expected to group updates for same resources in one entry anyways...

krancour: Yup. I _totally_ get where you were going with it, but at some point we have to ask whether all the extra looping is really better than a few extra gets that would only even happen in rare, odd uses of the feature. And even if it were, is the optimization also significant enough to justify the additional cognitive burden to understanding this code?

krancour:
@fykaa can you provide a working example I can try?

fykaa:
Yeah for sure, I just want to do a final test on my side. My Docker is acting up right now, fixing that first and then I'll definitely share the example here

fykaa:
Tested it end-to-end. This is the config that should work

```
kind: Project
metadata:
  name: kargo-demo
---
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: kargo-demo-repo
  namespace: kargo-demo
  labels:
stringData:
  repoURL: ${GITOPS_REPO_URL}
  username: ${GITHUB_USERNAME}
  password: ${GITHUB_PAT}
---
kind: Warehouse
metadata:
  name: kargo-demo
  namespace: kargo-demo
spec:
  subscriptions:
  - image:
      constraint: ^1.26.0
      discoveryLimit: 5
---
kind: PromotionTask
metadata:
  name: demo-promo-process
  namespace: kargo-demo
spec:
  vars:
  - name: gitopsRepo
    value: ${GITOPS_REPO_URL}
  - name: imageRepo
    value: public.ecr.aws/nginx/nginx
  steps:
  - uses: git-clone
    config:
      repoURL: ${GITOPS_REPO_URL}
      checkout:
      - branch: main
        path: ./src
      - branch: stage/${{ ctx.stage }}
        create: true
        path: ./out
  - uses: git-clear
    config:
      path: ./out
  - uses: git-commit
    as: commit
    config:
      path: ./out
      message: ${{ task.outputs.update.commitMessage }}
  - uses: git-push
    config:
      path: ./out
  - uses: set-metadata
    as: set-meta
    config:
      updates:
      - kind: Stage
        name: ${{ ctx.stage }}
        values:
          promotedBy: ${{ ctx.promotion }}
          lastImage: ${{ imageFrom("public.ecr.aws/nginx/nginx").Tag }}
          customField: "stage-metadata"
      - kind: Freight
        name: ${{ ctx.targetFreight.name }}
        values:
          deployed: true
          version: ${{ imageFrom("public.ecr.aws/nginx/nginx").Tag }}
          customField: "freight-metadata"
  - uses: git-commit
    as: meta-commit
    config:
      path: ./out
      message: |
        Promotion metadata:
        Stage: ${{ quote(stageMetadata(ctx.stage)) }}
        Freight: ${{ quote(freightMetadata(ctx.targetFreight.name)) }}
---
kind: Stage
metadata:
  name: test
  namespace: kargo-demo
spec:
  requestedFreight:
  - origin:
      kind: Warehouse
      name: kargo-demo
    sources:
      direct: true
  promotionTemplate:
    spec:
      steps:
      - task:
          name: demo-promo-process
        as: promo-process
---
kind: Stage
metadata:
  name: uat
  namespace: kargo-demo
spec:
  requestedFreight:
  - origin:
      kind: Warehouse
      name: kargo-demo
    sources:
      stages:
      - test
  promotionTemplate:
    spec:
      steps:
      - task:
          name: demo-promo-process
        as: promo-process
---
kind: Stage
metadata:
  name: prod
  namespace: kargo-demo
spec:
  requestedFreight:
  - origin:
      kind: Warehouse
      name: kargo-demo
    sources:
      stages:
      - uat
  promotionTemplate:
    spec:
      steps:
      - task:
          name: demo-promo-process
        as: promo-process
```

krancour:
@hiddeco it's not a high priority, but I might need you to look at this. I had the review mostly covered, but them ended up doing a fair amount of work on this PR myself.

The main thing I need feedback on is https://github.com/akuity/kargo/pull/4861/commits/32e371ca118516b3b4eb61b191cb1cebbcc3af8c.

hiddeco:
@krancour I believe that fetching the object as an unstructured type would bypass the cache as well.

krancour:
> I believe that fetching the object as an unstructured type would bypass the cache as well.

@hiddeco is there any reason you had in mind that fetching an unstructured object would be preferable?

Or were you only offering it as another option without saying you'd prefer it?

hiddeco:
No specific preference, it only occurred to me because that would save us from working with two client instances, which might become confusing and/or prone to errors.

krancour:
> I believe that fetching the object as an unstructured type would bypass the cache as well.

I verified this is the case.

> it only occurred to me because that would save us from working with two client instances

This is indeed cleaner.

https://github.com/akuity/kargo/pull/4861/commits/b8d8b86c9f3ca708fa7482af26e94c2691a36449 uses this approach.

PR #4869, opened 2025-08-15
fix: stuck in reconciling state when aborting promo

fuskovic:
https://github.com/user-attachments/assets/adb6e821-0ca4-47b2-a963-28bbd6a09747


krancour:
Superseded by #4871

PR #4872, opened 2025-08-16
fix(argocd-update): update sources instead of source when Application specifies both

### LINKED ISSUE #4856: `argocd-update` should update `sources` instead of `source` when Application specifies both
# Description

`argocd-update` supports updating `spec.source` for single-source apps, as well as `spec.sources` for multi-source apps. It chooses to update one or the other based on this line of code:

```golang
	// Update the desired source(s) in the Argo CD Application.
	if app.Spec.Source != nil {
		app.Spec.Source = desiredSources[0].DeepCopy()
	} else {
		app.Spec.Sources = desiredSources.DeepCopy()
	}
```

Basically, it concludes that the Application is a single-source app if `spec.source` is not nil. However, from the Argo CD [docs](https://argo-cd.readthedocs.io/en/stable/user-guide/multiple_sources/), it is actually possible to specify both `sources` and `source`. When this happens, `source` will be ignored.

> When you specify the sources field, Argo CD will ignore the source (singular) field.

This results in Kargo updating a field that is completely ignored by Argo CD.

Now, I would argue that specifying both `source` and `sources` is a bad practice and IMO should be a validation error. But it is what it is, and Kargo should account for this possibility.

The fix we need is to first check if `spec.sources` is specified and update that, otherwise update `spec.source`. 

# Steps to Reproduce

Create an app with both source and sources

# Version

```
v1.7.2
```

octonawish-akcodes:
fix #4856

### DIFF
--- a/internal/promotion/runner/builtin/argocd_updater_test.go
+++ b/internal/promotion/runner/builtin/argocd_updater_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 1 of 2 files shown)


krancour:
@octonawish-akcodes thank you for voluntarily doing this, but please note this work was already assigned to someone on my team. Luckily she had not yet started work on it (that I know of). I will re-assign to you, but please be careful in the future so neither you nor anyone else wastes their time doing something that someone else is already on top of.

internal/promotion/runner/builtin/argocd_updater_test.go
```diff
@@ -1028,6 +1029,176 @@ func Test_argoCDUpdater_syncApplication(t *testing.T) {
...
+					src.Object = map[string]any{
+						"spec": map[string]any{
+							"sources": []any{
+								map[string]any{
+									"repoURL":        "https://example.com/repo.git",
+									"targetRevision": "v2.0.0",
+								},
+							},
+						},
+					}
```
krancour, line 1052: None of the new test cases are actually validating the change. As proof, you can undo changes to `argocd_updater.go` and observe that new and existing test cases all still pass.

The problem right here is that you're defining the source object here and you're defining it properly, so of course the assertions below will pass in all cases.

What you should be doing here instead is asserting that the _actual_ source object that was handed to this mocked out patch function had been modified in the expected manner by the caller.

Similar problems in other new test cases.

If you get it right, the new test cases will fail without your change applied and they will pass when the change is applied.

krancour:
@octonawish-akcodes were you planning to complete this or should I re-assign it?

nitishfy:
Ping @octonawish-akcodes , are you willing to work on this? If not, I can take this up from here.

nitishfy:
we can close this PR in favour of https://github.com/akuity/kargo/pull/5072

PR #4883, opened 2025-08-17
feat: Harbor webhook receiver

### LINKED ISSUE #4858: Support for receiving Harbor webhooks
# Checklist

# Proposed Feature

Support for receiving webhook notifications from Harbor, which is still a relatively popular container registry. Relevant documentation:
https://goharbor.io/docs/2.13.0/working-with-projects/project-configuration/configure-webhooks/

Maybe this was discussed as part of https://github.com/akuity/kargo/issues/4150 but it's not mentioned in the issue itself, so I thought I'd ask.

# Motivation

Harbor is popular. Other registries are explicitly supported. 

Counterpoint: we can use the upcoming generic webhook support to get working Harbor integration.

I may be willing to provide a PR for this if you are interested in accepting the feature, but I understand that any feature is also a maintenance burden for you.

ahockersten:
Support for receiving webhooks from Harbor. Resolves #4858.

Currently this only listens to the `PUSH_ARTIFACT` event. It **might** be useful to also handle the `SCANNING_COMPLETED` event, in case Harbor is configured in such a way that images cannot be pulled until they have gone through scanning. I have opted not to do this however, since I don't know enough about how Harbor works to figure out if this makes sense. It should be easy to add support in the future if needed.


ahockersten:
# How I verified this locally

Since getting Harbor and Kargo up and running for testing this was a project that took longer than writing the code, here is a detailed breakdown of the steps I took so that anyone (including me 😅) who wants to verify or debug this code now or in the future will have an easier time.

This worked on my Linux machine, if using WSL or MacOS you might need to tweak these instructions.

Setup a cluster for Kargo, with an extra port opening for Harbor:

```sh
echo '    - containerPort: 30090 # Harbor
      hostPort: 30090' >> hack/kind/cluster.yaml

make hack-kind-up
```

Add `harbor` to your `/etc/hosts` file. While you **could** use externalUrl `localhost`, this will become problematic later on because setting the `repoURL` to `localhost` in you `Warehouse` means the call to `refreshWarehouses` won't find the warehouses properly. 

```sh
sudo echo "127.0.0.1 harbor" >> /etc/hosts
```

Setup a local Harbor instance:
```sh
helm repo add harbor https://helm.goharbor.io
helm fetch harbor/harbor --untar
helm install release harbor/ --set expose.type=nodePort --set expose.tls.enabled=false --set expose.nodePort.ports.http.nodePort=30090 --set externalURL=http://harbor:30090
```

Start Kargo:
```sh
tilt up
```

Add a test image to Harbor:
```sh
docker pull alpine:3.18.0
docker login harbor:30090 -u admin -p Harbor12345
docker tag library/alpine:3.18.0 harbor:30090/library/alpine:3.18.0
docker push harbor:30090/library/alpine:3.18.0
```

Add resources:
```yaml
echo '
kind: Project
metadata:
  name: kargo-webhooks-harbor-verify
---
apiVersion: v1
kind: Secret
metadata:
  name: harbor-wh-secret
  namespace: kargo-webhooks-harbor-verify
  labels:
    kargo.akuity.io/cred-type: generic
data:
  auth-header: c2VjcmV0 # == secret
---
kind: ProjectConfig
metadata:
  name: kargo-webhooks-harbor-verify
  namespace: kargo-webhooks-harbor-verify
spec:
  webhookReceivers:
  - name: harbor-wh-receiver
    harbor:
      secretRef:
        name: harbor-wh-secret
---
kind: Warehouse
metadata:
  name: warehouse
  namespace: kargo-webhooks-harbor-verify
spec:
  freightCreationPolicy: Automatic
  interval: 5m0s
  subscriptions:
    - image:
        discoveryLimit: 20
        imageSelectionStrategy: SemVer
        repoURL: harbor:30090/library/alpine
        strictSemvers: true
' > resources.yaml
```

Apply this to the cluster:
```sh
kubectl apply -f resources.yaml
```

Fetch the URL to use:
```sh
 kubectl get projectconfigs kargo-webhooks-harbor-verify \
  -n kargo-webhooks-harbor-verify \
  -o=jsonpath='{.status.webhookReceivers[0].url}'
```

Go to http://harbor:30090/harbor/projects/1/webhook and login to Harbor with user `admin` and password `Harbor12345`.

Add a new webhook. Set these values, while keeping the rest at defaults:
- Name: `kargo-webhooks-harbor-verify`
- Endpoint URL: `<output from above, with hostname and port changed to "kargo-external-webhooks-server.kargo.svc.cluster.local">`
- Auth Header: `secret`

Now, test the webhook by pushing the image to Harbor again. You can repeat this as needed, Harbor will trigger the `PUSH_ARTIFACT` event even if the push was effectively a noop.
```sh
docker push harbor:30090/library/alpine:3.18.0
```

## Cleanup

Remove harbor from your `/etc/hosts` file:
```sh
sudo sed -i '/harbor/d' /etc/hosts
```

ahockersten:
OK I feel a bit dumb now. The Harbor demo server on https://demo.goharbor.io/ could probably be used for verification rather than setting up your own instance. I thought it couldn't be used, since webhooks require project admin rights, but I just checked and you are, in fact, allowed to create projects on that demo server (which will grant you admin rights).

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/harbor/index.md
```diff
@@ -0,0 +1,114 @@
...
+:::info
+"Refreshing" a `Warehouse` resource means enqueuing it for immediate
+reconciliation by the Kargo controller, which will execute the discovery of
+new artifacts from all repositories to which that `Warehouse` subscribes.
+:::
+
+## Configuring the Receiver
+
+A Harbor webhook receiver must reference a Kubernetes `Secret` resource
+with a `auth-header` key in its data map. This
```
krancour:
```suggestion
with an `auth-header` key in its data map. This
```

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/harbor/index.md
```diff
@@ -0,0 +1,114 @@
...
+
+    5. Check <Hlt>Verify Remote Certificate</Hlt>.
+
+    6. Click <Hlt>Add</Hlt>
+
+3. The webhook should now be registered with Harbor. You can test it by pushing
+   an artifact to the repository, and verifying in Harbor's logs that the
+   webhook was received successfully by Kargo.
+
+   ![Webhooks View After Adding Webhook](./img/webhooks-view-after-adding-webhook.png "Webhooks View After Adding Webhook")
```
krancour: Very minor formatting fixes:

```suggestion
1. Navigate to
   `https://<harbor-instance>/harbor/projects/<project_id>/webhook`, where
   `<harbor-instance>` and `<project_id>` have been replaced with a Harbor
   instance and project for which you are an administrator.

1. Click <Hlt>New Webhook</Hlt>.

    ![Webhooks View](./img/webhooks-view.png "Webhooks View")

1. Complete the <Hlt>Add Webhook</Hlt> form:

    ![Add Webhook](./img/add-webhook.png "Add Webhook")

    1. Enter a descriptive name in the <Hlt>Name</Hlt> field.

    1. Check <Hlt>Event type</Hlt> <Hlt>Artifact pushed</Hlt>. Uncheck all other
       event types.

    1. Complete the <Hlt>Endpoint URL</Hlt> field using the URL
       [for the webhook receiver](#retrieving-the-receivers-url).

    1. Complete the <Hlt>Auth Header</Hlt> field using shared secret
       [generated above](#configuring-the-receiver).

        :::note
        Use the unencoded value; not the base64-encoded value.
        :::

    1. Check <Hlt>Verify Remote Certificate</Hlt>.

    1. Click <Hlt>Add</Hlt>

1. The webhook should now be registered with Harbor. You can test it by pushing
   an artifact to the repository, and verifying in Harbor's logs that the
   webhook was received successfully by Kargo.

   ![Webhooks View After Adding Webhook](./img/webhooks-view-after-adding-webhook.png "Webhooks View After Adding Webhook")
```

krancour: > The webhook should now be registered with Harbor. You can test it by pushing
   an artifact to the repository, and verifying in Harbor's logs that the
   webhook was received successfully by Kargo.

@ahockersten I don't have much prior experience with Harbor. While I intend to test-drive this e2e, you might able to answer this form me already... Is there any other way to test/verify this? I'm guessing it would not be entirely typical for users to have access to Harbor's logs. And if there isn't any other way, that's fine. I'm just asking. We've found that the capabilities when it comes to testing webhooks and validating their delivery vary widely from platform to platform.

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/harbor/index.md
```diff
@@ -0,0 +1,114 @@
...
+
+```shell
+kubectl get projectconfigs kargo-demo \
+  -n kargo-demo \
+  -o=jsonpath='{.status.webhookReceivers}'
+```
+
+## Registering with Harbor
+
+1. Navigate to
```
krancour, line 78: I'll be getting cozy with the Harbor docs at some point as well as testing e2e, but I am curious whether you may know anything about options for registering at different "levels?" Capabilities vary widely by platform, but for many we've seen options for org-level webhook registrations as well as repo-level.

internal/webhook/external/harbor.go
```diff
@@ -0,0 +1,174 @@
...
+	return harbor
+}
+
+// getSecretValues implements WebhookReceiver.
+func (q *harborWebhookReceiver) getSecretValues(
+	secretData map[string][]byte,
+) ([]string, error) {
+	secretValue, ok := secretData[harborSecretDataKey]
+	if !ok {
+		return nil, errors.New("secret data is not valid for a Harbor WebhookReceiver")
```
krancour: I know this was probably copied and pasted from another receiver and then modified. Some receivers have a more informative error message for this than others do. (I eventually want to fix them all.) As a new receiver, let's just use the more informative message here from the get go:

```suggestion
		return nil, fmt.Errorf(
			"missing data key %q for Harbor WebhookReceiver",
			harborSecretDataKey,
		)
```

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/harbor/index.md
ahockersten: Obviously my docs are unclear. :) You can see logs in Harbor's webui for specific webhook runs in the same place where you register the webhooks. Look under the heading "Webhook Job Histories" here: https://goharbor.io/docs/main/working-with-projects/project-configuration/configure-webhooks/

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/harbor/index.md
ahockersten: I have only been able to find documentation for creating webhooks on the "project" level but I don't have any prior experience with Harbor webhooks so it's possible I've missed something (Harbor's docs for webhooks aren't great either, so it's also possible documentation for global webhooks is missing, if they exist).

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/harbor/index.md
ahockersten: :+1:

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/harbor/index.md
ahockersten: I took your suggestions verbatim. Thank you!

I've tried making the documentation slightly clearer regarding where in Harbor logs can be reviewed, feel free to suggest further improvements if you have them.

krancour:
> The Harbor demo server on https://demo.goharbor.io/ could probably be used for verification...

@ahockersten thanks for this tip! This will make my test drive a whole lot easier!

internal/webhook/external/harbor.go
ahockersten: Makes sense!

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/harbor/index.md
ahockersten: I've now looked through my local Harbor instance's global settings, and I can't find any way of setting webhooks besides doing it through projects, so I strongly suspect this is not supported.

internal/webhook/external/harbor.go
```diff
@@ -0,0 +1,178 @@
...
+				w,
+				xhttp.Error(errors.New("invalid request body"),
+					http.StatusBadRequest,
+				),
+			)
+			return
+		}
+
+		// Only handle artifact push events for now
+		if payload.Type != "PUSH_ARTIFACT" {
```
krancour: I would suggest defining a constant `harborEventTypePush` with this value since it's used many times throughout the tests.

internal/webhook/external/harbor.go
```diff
@@ -0,0 +1,178 @@
...
+				} `json:"resources"`
+				Repository struct {
+					DateCreated  int64  `json:"date_created"`
+					Name         string `json:"name"`
+					Namespace    string `json:"namespace"`
+					RepoFullName string `json:"repo_full_name"`
+					RepoType     string `json:"repo_type"`
+				} `json:"repository"`
+			} `json:"event_data"`
+		}{}
```
krancour: I would recommend leaving out fields we're uninterested in.

internal/webhook/external/harbor.go
```diff
@@ -0,0 +1,178 @@
...
+			if res.ResourceURL != "" {
+				// Normalize URLs both as image and Helm chart repositories
+				// since Harbor supports both Docker images and Helm charts
+				repoURLs = append(repoURLs, image.NormalizeURL(res.ResourceURL))
+				repoURLs = append(repoURLs, helm.NormalizeChartRepositoryURL(res.ResourceURL))
+			}
+			if res.Tag != "" {
+				tags = append(tags, res.Tag)
+			}
+		}
```
krancour: So this is tricky...

Here's an example push event from the Harbor docs:

```
{
  "type": "PUSH_ARTIFACT",
  "occur_at": 1680501893,
  "operator": "harbor-jobservice",
  "event_data": {
    "resources": [
      {
        "digest": "sha256:954b378c375d852eb3c63ab88978f640b4348b01c1b3456a024a81536dafbbf4",
        "tag": "sha256:954b378c375d852eb3c63ab88978f640b4348b01c1b3456a024a81536dafbbf4",
        "resource_url": "localhost/harbor/alpine@sha256:954b378c375d852eb3c63ab88978f640b4348b01c1b3456a024a81536dafbbf4"
      }
    ],
    "repository": {
      "date_created": 1680501893,
      "name": "alpine",
      "namespace": "harbor",
      "repo_full_name": "harbor/alpine",
      "repo_type": "private"
    }
  }
}
```

From this structure, it would appear that an event could list multiple images that were pushed to _one_ repo. You are correct to be looping through `payload.EventData.Resources` and picking out the tags. It's a different story for `resource_url`. Those aren't URLs for the repository. They're URLs for specific images. 

i.e.:

* Repo: `localhost/harbor/alpine`
* Image: `localhost/harbor/alpine@sha256:954b378c375d852eb3c63ab88978f640b4348b01c1b3456a024a81536dafbbf4`.)

We want the repo URL. As it so happens, if you normalize the image URL, what you will be left with _is_ the repo URL, however... recalling that these events can list multiple images that were pushed to _one_ repo, every single one of those `resource_url`s is going to start with `localhost/harbor/alpine`.

So you're building a list of repo URLs that is just `localhost/harbor/alpine` over and over again.

Now...

It would seem most correct to retrieve the repo URL from info in the `repository` block of the data, but nothing in there looks like a _full_ URL. i.e. You don't see `localhost` anywhere in there.

So you're getting the repoURL from the best place you possibly can... but you really only need it once. So this is what I would suggest:

```suggestion
		var repoURLs []string
		var tags []string
		if len(payload.EventData.Resources) > 0 {
			// Payloads from Harbor contain no information about media type, so we
			// normalize the URL BOTH as if it were an image repo URL and as if it
			// were a chart repository URL. These will coincidentally be the same, but
			// by doing this, we safeguard against future changes to normalization
			// logic. Note: The refresh logic will dedupe the URLs, so this does not
			// create the possibility of a double refresh.
			repoURLs = []string{
				image.NormalizeURL(res.ResourceURL,
				helm.NormalizeChartRepositoryURL(res.ResourceURL),
			}
			tags = make([]string, len(payload.EventData.Resources))
			for i, res := range payload.EventData.Resources {
				tags[i] = res.Tag
			}
		}
```

internal/webhook/external/harbor.go
ahockersten: Sure. Done!

internal/webhook/external/harbor.go
ahockersten: Makes sense

internal/webhook/external/harbor.go
ahockersten: Good find! Yes, I understand. I did make the observation that this would include the repo twice when receiving push events for multiple tags, but I decided against doing anything further since it would get deduplicated inside refreshWarehouse.

Your solution is cleaner though. I have taken it with some slight modifications to make it compile.

ahockersten:
@krancour  I have gone through your comments and I believe I have addressed all of them. I've left the ones that might need verification from you commented but unresolved.

Thank you for the review!

internal/webhook/external/harbor.go
krancour: Sorry my suggestion didn't compile. It's the hazard of writing code in a little textbox on GH.

krancour: 🤦‍♂️

internal/webhook/external/harbor_test.go
```diff
@@ -0,0 +1,416 @@
...
+			},
+			"repository": map[string]interface{}{
+				"date_created":   1680501893,
+				"name":           "my-chart",
+				"namespace":      "charts",
+				"repo_full_name": "charts/my-chart",
+				"repo_type":      "private",
+			},
+		},
+	}
```
krancour: I would consider removing any superfluous data to improve the signal:noise ratio.

krancour: Oh... the linter also doesn't like `interface{}` as it turns out.

```suggestion
	validPushArtifactEvent := map[string]any{
		"type":     harborEventTypePush,
		"event_data": map[string]any{
			"resources": []map[string]any{
				{
					"tag":          "v1.0.0",
					"resource_url": "harbor.example.com/library/alpine:v1.0.0",
				},
			},
		},
	}

	validPushArtifactEventMultipleResources := map[string]any{
		"type":     harborEventTypePush,
		"event_data": map[string]any{
			"resources": []map[string]any{
				{
					"tag":          "v1.0.0",
					"resource_url": "harbor.example.com/library/alpine:v1.0.0",
				},
				{
					"tag":          "v1.1.0",
					"resource_url": "harbor.example.com/library/alpine:v1.1.0",
				},
			},
		},
	}

	validHelmChartPushEvent := map[string]any{
		"type":     harborEventTypePush,
		"event_data": map[string]any{
			"resources": []map[string]any{
				{
					"tag":          "v1.0.0",
					"resource_url": "harbor.example.com/charts/my-chart:v1.0.0",
				},
			},
		},
	}
```

krancour:
An e2e test drive went smoothly.

internal/webhook/external/harbor_test.go
ahockersten: Suggestion taken, with some additional noise removed as well.

Issue #4752, opened 2025-08-04
UI: make graph nodes adapt dynamic sizes


Marvin9:
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

We have classic layout/rendering loop problem — layout needs dimensions, but dimensions come from rendering. So we cap the dimension for layout to fixed and visually too it is fixed. This causing overflow issues when content gets bigger or uneven nodes connection.

I propose we take granular approach.

Two-Pass Layout (Measure-Then-Layout)

1. First Pass – Render to Measure Only

- Render nodes off-canvas or invisible, just to measure them.
- Use refs or ResizeObserver to capture actual width/height.

2. Second Pass – Dagre Layout

- Feed the measured dimensions into Dagre.
- Apply computed positions to nodes.
- Render graph normally with correct layout.

# Motivation

- If we want to show more data in stage nodes, it will adapt sizes. Right now we'd have to increase fixed dimension manually for all nodes. Basically content overflow when its more than the size of node is resolved
- More natural and visual appealing of graphs

PR #4898, opened 2025-08-19
feat(ui): graph nodes adapt to dynamic sizing

Marvin9:
https://github.com/akuity/kargo/issues/4752


ui/src/features/project/pipelines/nodes/node-size-source-of-truth.module.less
jmalisz-akuity, line 1: Please add `max-width` (maybe double the `min-width`) for some edge cases. Seems to work alright with the first pass dummy render.

ui/src/features/project/pipelines/graph/layout-graph.ts
```diff
@@ -5,7 +5,9 @@ import { RepoSubscription, Stage, Warehouse } from '@ui/gen/api/v1alpha1/generat
 
 import { repoSubscriptionIndexer, stageIndexer, warehouseIndexer } from './node-indexer';
 import { repoSubscriptionLabelling, stageLabelling, warehouseLabelling } from './node-labeling';
-import { repoSubscriptionSizer, stageSizer, warehouseSizer } from './node-sizer';
+// import { repoSubscriptionSizer, stageSizer, warehouseSizer } from './node-sizer';
```
jmalisz-akuity: Remove

jmalisz-akuity (changes requested):
Sometimes the nodes and connection disappear. It's kinda flaky, I cannot readily reproduce it. It seems to happen more on smaller screen sizes.

Marvin9:
> Sometimes the nodes and connection disappear after initial render (try refreshing the page a few times). It's kinda flaky, I cannot readily reproduce it. It seems to happen more on smaller screen sizes.

Spent a lot amount of time but couldn't really reproduce. I will try reproducing and fix in separate PR if I couldn't reproduce while this merge

jmalisz-akuity (approved):
Well it still happens for me every few renders. I've also looked into it, but couldn't find anything out of ordinary. Later I found this issue in `react-flow`, I think it's the same problem:
https://github.com/xyflow/xyflow/issues/4888

Regardless, since it happens rarely I think we can leave it as it is.

Issue #4718, opened 2025-07-30
Generate ConnectRPC API as part of docs


jessesuen:
# Proposed Feature

We have a ConnectRPC API but users are unsure on how to use it. We should:
1. Generate the ConnectRPC API docs from the protos
2. Have some examples/code snippets with cURL or go (importing kargo as a library) 

# Motivation

Some users want to interface with Kargo without resorting to UI or CLI.

fuskovic:
I noticed we have an existing buf BSR account at https://buf.build/akuity ( I don't have access to this account ) and use buf extensively for codegen.

Is that how we want to publish our API docs?

e.g.

<img width="644" height="581" alt="Image" src="https://github.com/user-attachments/assets/2e65d93b-15f8-4790-94c4-23786306a266" />

krancour:
We're taking an incremental approach to this, so I am closing this in favor of several more narrowly scoped issues I just opened.

PR #4901, opened 2025-08-19
feat: api docs

fuskovic:
Part of: https://github.com/akuity/kargo/issues/4718

Updates local dev toolchain to support protoc-gen-doc.

Introduces `api-docs.templ` which we can use to edit how docs are laid out.

`make hack-codegen` + `make hack-serve-docs` for testing


krancour:
@fuskovic this is a great start. As far as examples go, it would indeed be great to have some, _but_ I'd suggest even higher priority than that, we probably need a human-friendly description of each endpoint and message. We're good about that in Go code. I guess we dropped the ball on the protobuf definitions.

docs/docs/90-api-documentation.md
```diff
@@ -0,0 +1,3069 @@
+# Protocol Documentation
```
krancour: I know this file is generated, obviously, but do we have control over this heading text somehow? "Protocol Documentation" feels like odd wording for what this is.

api-docs.templ
```diff
@@ -0,0 +1,104 @@
+# API Documentation
```
fuskovic: This file comes from https://github.com/pseudomuto/protoc-gen-doc/blob/master/resources/markdown.tmpl. MIT license.

The title differs from the original. Placing our own copy in this repo gives us more control over how information is laid out in the final rendered file.

fuskovic:
Docusauras was complaining about missing closing tags and failing to compile. Several comments in the code base were changed by removing symbols like`<` and `>`. `make hack-serve-docs` works locally for me now. CI is complaining about broken links (`/api-documentation`). It works locally so I need to dig in further into that.

docs/docs/90-api-documentation.md
```diff
@@ -0,0 +1,6140 @@
+# API Documentation
+<a name="top"></a>
+
+## Table of Contents
```
fykaa: It is better to remove lines 4–349 since Docusaurus already generates a table of contents in the sidebar, so this section is redundant

docs/docs/90-api-documentation.md
```diff
@@ -0,0 +1,6140 @@
...
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| project | [string](#string) |  |  |
+| name | [string](#string) |  |  |
+
+
+
+
+
+
```
fykaa: Extra blank lines could cause linting issues

docs/docs/90-api-documentation.md
fykaa: something like this 
image

docs/docs/90-api-documentation.md
fuskovic: Not seeing a clear way to solve this by modifying `api-docs.templ` but also not seeing any lint issues for this.

fuskovic:
Ok, I found the cause of the broken links.

Problem 1 (solved):

First, it was the Kube API types. The proto compiler requires those be in the same dir when performing doc generation but, we don't need to generate docs for the Kube API types so we omit those. The problem with this was we generated broken links when we had something like:

Screenshot 2025-08-21 at 2 28 48 PM

So I fixed this by updating the template to omit wrapping Kube API types with anchor tags: https://github.com/akuity/kargo/pull/4901/commits/b3f2432166eb45990cef6652068c39ee957a6362

Problem 2 (solved):

The second issue was the regex pattern in kube validation pattern rules. Consider the pattern: 

`[a-z0-9]([-a-z0-9]*[a-z0-9])` this resolved to a markdown link due to the format.

So I stripped the square brackets so docusauras wouldn't think it was a link: https://github.com/akuity/kargo/pull/4901/commits/57d7e4d78bf7663a7bac304d4f8f7c4b4eb97505

# Conclusion

All linters satisfied + broken links fixed.

fuskovic:
> @fuskovic this is a great start. As far as examples go, it would indeed be great to have some, _but_ I'd suggest even higher priority than that, we probably need a human-friendly description of each endpoint and message. We're good about that in Go code. I guess we dropped the ball on the protobuf definitions.

As discussed offline, I can follow up with a Copilot assigned issue for generating comments and polish that as needed.

api/v1alpha1/warehouse_types.go
```diff
@@ -233,9 +233,9 @@ type GitSubscription struct {
 	// For NewestFromBranch (commit filtering):
 	//   - `id`: The ID (sha) of the commit.
 	//   - `commitDate`: The commit date of the commit.
-	//   - `author`: The author of the commit message, in the format "Name <email>".
+	//   - `author`: The author of the commit message, in the format "Name email".
```
krancour: I understand the angled brackets are probably causing some trouble for doc generation.

If the angled brackets here were indicative of some kind of placeholder, as they often tend to be, I think this change wouldn't be so bad, but unfortunately, that's not the case here and the angled brackets are quite literally expected to be included in the string the user provides as the value of this field.

Example from an actual git commit: `Author: Kent Rancourt <kent.rancourt@gmail.com>`.

So, unfortunately, I think we're significantly reducing the accuracy of the docs with this change. Is there any possibility of encoding/escaping in the templates to account for this instead?

fuskovic: For doc generation no but for docusauras compilation yes. It interprets <email> as a tag and fails to compile due to what it thinks is the absence of a closing tag.

I fixed using `&lt;` + `&gt;` : https://github.com/akuity/kargo/pull/4901/commits/fbe346653b0f9e677d39ec1f343504264ead7a69

api-docs.templ
```diff
@@ -0,0 +1,82 @@
...
+## {{.Name}}
+{{.Description}}
+
+{{range .Messages}}
+<a name="{{.FullName | anchor}}"></a>
+
+### {{.LongName}}
+{{.Description}}
+
+{{if .HasFields}}
```
krancour: I notice the markdown generated from this template has a lot of extra line breaks. It's not a huge problem, by any means, but lots of markdown linters don't like that. fwiw, it's easy to avoid by favoring `{{- if ... -}}` and `{{- end -}}` over `{{ if ... }}` and `{{ end }}`. The dashes on the left and right eat up the extra whitespace to the left and right of the expression.

So, while this isn't a huge problem, it's one we can fix relatively easily, so I think we ought to.

hack/codegen/proto.sh
```diff
@@ -105,6 +105,42 @@ function main() {
...
+    "api/stubs/rollouts/v1alpha1/generated.proto" \
+    "k8s.io/api/core/v1/generated.proto" \
+    "k8s.io/api/batch/v1/generated.proto" \
+    "k8s.io/api/rbac/v1/generated.proto" \
+    "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1/generated.proto" \
+    "k8s.io/apimachinery/pkg/util/intstr/generated.proto" \
+    "k8s.io/apimachinery/pkg/api/resource/generated.proto" \
+    "k8s.io/apimachinery/pkg/runtime/schema/generated.proto" \
+    "k8s.io/apimachinery/pkg/runtime/generated.proto" \
+    "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto"
```
krancour: Help me understand how this works...

I notice that all the k8s.io APIs are not included in the generated docs. I think that's the right thing to do, but I'm wondering what actually prevents them from appearing there. Is it the `:k8s.io` at the end of line +125?

The reason I'm asking is that I think we should consider `api/stubs/rollouts/v1alpha1/generated.proto` _also_ not being included in the final product.

These docs already feel a bit like drinking from a fire hose, so if we can improve the signal:noise ratio by omitting things where we can reasonably claim they're documented better elsewhere, I think we ought to do so.

api-docs.templ
krancour, line 1: Wondering if there is a better place to put this template than in the root of the repo.

hack/codegen/proto.sh
fuskovic: > but I'm wondering what actually prevents them from appearing there. Is it the `:k8s.io` at the end of line +125?

Yes

```
You can also exclude proto files that match specific path expressions. This is done by passing a second option delimited by :. For example, you can pass any number of comma separated patterns as the second option:
```
source: https://github.com/pseudomuto/protoc-gen-doc

> The reason I'm asking is that I think we should consider api/stubs/rollouts/v1alpha1/generated.proto also not being included in the final product.

Will do

api-docs.templ
fuskovic: `docs/api/docs.templ` ?

fuskovic:
@krancour shaved ~2900 lines off the diff by removing superfluous line breaks + omitting doc generation for rollout stubs. More than 2K of them from the former.

api-docs.templ
krancour: That sgtm.

api/v1alpha1/warehouse_types.go
krancour: Nice solution. Thanks!

api-docs.templ
krancour: Sorry if I wasn't clear. Expanded use of `{{-` and `-}}` _throughout the entire template_ would eliminate many or all of the double line breaks in the rendered Markdown.

krancour:
The way that kubebuilder markers are bleeding through feels awkward. Is there anything that can be done to remove those from the final, rendered Markdown?

If we are able to stop short of anything quite as heavy handed as suppressing all comments that contain `+kubebuilder`, that would be ideal. Maybe it is possible to suppress the remainder of any comment after the first `// +` is encountered?

api-docs.templ
```diff
@@ -0,0 +1,81 @@
...
+
+### {{.Name}}
+{{.Description}}
+
+| Method Name | Request Type | Response Type | Description |
+| ----------- | ------------ | ------------- | ------------|
+{{range .Methods -}}
+  | {{.Name}} | [{{.RequestLongType}}](#{{.RequestFullType | anchor}}){{if .RequestStreaming}} stream{{end}} | [{{.ResponseLongType}}](#{{.ResponseFullType | anchor}}){{if .ResponseStreaming}} stream{{end}} | {{nobr .Description}} |
+{{end}}
+{{end}} <!-- end services -->
```
krancour: My gut says that people looking at these docs will be _primarily_ interested in knowing what services exist, what they do, etc. and that the messages that comprise the requests to and responses are secondary by comparison...

So, I think we should consider moving these up higher -- _above_ messages.

api-docs.templ
```diff
@@ -0,0 +1,81 @@
+# API Documentation
+<a name="top"></a>
+
+{{- range .Files}}
+{{$file_name := .Name}}
+<a name="{{.Name | anchor}}"></a>
+<p align="right"><a href="#top">Top</a></p>
+
+## {{.Name}}
```
krancour: Breaking this doc by filenames of .proto files. I'm not sure anyone would _care_, for instance, that the services, messages, etc. are defined in a file named `api/service/v1alpha1/service.proto`. It may be more important that we clearly demarcate the boundaries of whole APIs. Fortunately, we can probably get away with using the .proto filenames as proxies for whole APIs. Can we maybe just put logic in the template to translate filenames like `api/service/v1alpha1/service.proto` to nicer API names like `service/v1alpha1`? Or something like that?

api-docs.templ
fuskovic: Doing it throughout the entire template causes various elements to break but, I've made the adjustments needed to render a result that yields no double line breaks. Which I believe is the intention here.

fuskovic:
> The way that kubebuilder markers are bleeding through feels awkward. Is there anything that can be done to remove those from the final, rendered Markdown?
> 
> If we are able to stop short of anything quite as heavy handed as suppressing all comments that contain `+kubebuilder`, that would be ideal. Maybe it is possible to suppress the remainder of any comment after the first `// +` is encountered?

I've added some logic to split comments/descriptions into multiple lines, and omit lines containing kubebuilder: https://github.com/akuity/kargo/pull/4901/commits/4689bd66cb2c1d06ef436ddebe85da3abe59eed2

This way we still get the part of the comment/description we want

krancour (approved):
This is a great start. I'll open a few follow-up issues for incremental improvements, but this is g2g.

Issue #4902, opened 2025-08-19
Kargo login --sso doesn't work with windows cli


gladavius:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

I configured my kargo instance to manage sso authentication based on google workspace with dex.

When connecting with the cli from wsl (ubuntu 22.04) everything works well, I get properly directed to the google auth page and then back to kargo.

But when trying to do the same from windows I got an error:
Error: error opening system default browser: exit status 1

You can also see a screenshot of what happen in my browser

# Screenshots

<img width="1236" height="356" alt="Image" src="https://github.com/user-attachments/assets/cf6cafbe-6379-4460-b457-53d5195e3094" />

# Steps to Reproduce

- Download windows kargo cli fom ui
- kargo login https://my-kargo-url.com --sso

# Version

```
Client Version: v1.7.2
```

# Logs

In firefox logs I clearly see that parameters are not properly transmitted to te browser
```
firefox.moz_log:2025-08-19 20:59:31.603000 UTC - [Parent 36396: Main Thread]: E/nsHttp   Referer: https://xxxxxxxxx.com/dex/auth/google?client_id=xxxxxxxxxx-cli
```
whereas when launched from wsl
```
firefox.moz_log:2025-08-19 21:01:47.049000 UTC - [Parent 36396: Main Thread]: E/nsHttp uri=https://xxxxxxxxxx.com/dex/auth?client_id=xxxxxxxxxxxxx.com-cli&code_challenge=xxxxxxxxxxxxxxxxxxx&code_challenge_method=S256&redirect_uri=http%3A%2F%2Flocalhost%3A42527%2Fauth%2Fcallback&response_type=code&scope=openid+profile+email+groups+offline_access&state=xxxxxxxxxxxxxxxxxxxx
```
# Suggested solution

I've been able to workaround this issue by replacing the lib *github.com/bacongobbler/browser* by *github.com/pkg/browser*

I can open a pull request with this fix unless someone else has a better solution to propose

krancour:
> I've been able to workaround this issue by replacing the lib github.com/bacongobbler/browser by github.com/pkg/browser

> I can open a pull request with this fix unless someone else has a better solution to propose

I'd be fine with that as long as we're certain it:

1. Doesn't break on the OSes where it was already working.
[...]

PR #4903, opened 2025-08-19
fix(cli): replace lib bacongobbler/browser by pkg/browser to fix issue on windows cli

gladavius:
#4902


krancour:
I will test it on Mac OS when I'm next at my desk.

There is another usage of the old lib in `dashboard.go`, which should also be fixed, else we're using one library to do this in one place and a different one to do it in another.

gladavius:
> I will test it on Mac OS when I'm next at my desk.
> 
> There is another usage of the old lib in `dashboard.go`, which should also be fixed, else we're using one library to do this in one place and a different one to do it in another.

done

krancour:
Confirmed it runs fine on Mac OS and also compiles without CGO. I did run `go mod tidy` to remove the no-longer-used dependency on the old package.

LGTM

Issue #4680, opened 2025-07-25
Quality of life updates for managing versions and freights (create from freight and diff)


Blongep:
# Checklist

* [✓] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

I'm working on a platform engineering solution that uses multiple fairly large freights (4 Warehouses, 3 of which watch more than 5 sources). To help with our workflow and update process, we would need two QoL features:

- Freight from freight:
Have a way for the interface to manually create a freight from a different one. It would open the freight creation overlay with the versions from the other freight, allowing for specific updates or rollback.

- Freight diff:
Similar to the previous feature, the interface would have a way to compare two freights to see which versions are different

# Motivation

Let's take a complete use case:
I have a warehouse with several sources including our core-api docker image.
I have three stages:
- Development
- Integration
- Production
Each of these render yaml manifests to an environment-specific branch which is deployed via argocd.

Let's say I build a new core-api image version; 1.0.2.
The new freight is promoted to dev automatically.
Testing passes, I am looking to promote to integration. I want to make sure that the only version that changes is this particular one: i diff the new freight against the freight currently in integration. 
If the core-api version is the only difference, I promote. 
If not, I can create a new freight from the one currently in integration. The freight creation form opens with corresponding version already selected. I just have to change the core-api one and confirm.

These two feature put togeher would allow for much easier fine-grained version control and partial updates/rollbacks.

qtchaos:
Diffing is something I would definitely like to see, especially *if* support for promoting multiple freights at once gets added. For example promoting all freights in QA to prod could be done with a single click "Promote multiple to downstream" which could open up a view with the diff for each version.

Brightside56:
> Freight from freight:
Have a way for the interface to manually create a freight from a different one. It would open the freight creation overlay with the versions from the other freight, allowing for specific updates or rollback.

Over the last year I have seen something like ten people who who are experiencing difficulties due to the lack of this functionality - https://github.com/akuity/kargo/issues/2988.

In my case, I solved it in a crude way. We have simple release policy, and according to that policy one set of artifacts is delivered to all environments after testing. We commit new versions for updated artifacts to `main` and Kargo is used only to render manifests to the right branches and update ArgoCD applications. It's sad, but still better than specify version for each image when assembling freight in UI or writing CI-pipeline for promotion

[...]

Issue #2988, opened 2024-11-23
ui: option to copy existing freight as starting point for assembly


krancour:
From https://github.com/akuity/kargo/issues/2979#issuecomment-2492701879

> In case of freight assembly I often need to create freight that differs from the previous one by one/two artifacts from 5-10, but for this I need to search and select proper version for each artifact. Option "create freight like this" could help a lot

Brightside56:
Could anyone to provide hints or short guidance how to implement it, provided that such a feature can be accepted? Thanks!

PR #4905, opened 2025-08-20
feat(ui): create freight like this

Brightside56:
Addresses https://github.com/akuity/kargo/issues/2988 and https://github.com/akuity/kargo/issues/4680 (partially). 

I have very modest skills in UI, but I am ready to spend time and finish this if someone can help

Screenshot 2025-08-20 at 04 58 27
Screenshot 2025-08-20 at 04 58 43
Screenshot 2025-08-20 at 05 00 15


Marvin9:
Thanks. Functionality looks good to me but code needs heavy refactor. I will follow up with some commits

Marvin9:
@Brightside56 I have refactored code and updated logics and bit of UI. Feel free to test it

Brightside56:
@Marvin9 Thank you! Everything seems to be great except situations when auto-populated artifact is not on first page - Kargo will still open first page. It's not critical, but IMO it would better to visualize selection

Screenshot 2025-09-04 at 01 45 31

Marvin9:
`useDetectPage`does that @Brightside56 but there is a bug. Thanks for catching it! Will include fix in next commit

Brightside56:
LGTM

ui/src/features/project/pipelines/freight/freight-card.tsx
```diff
@@ -121,6 +122,21 @@ export const FreightCard = (props: FreightCardProps) => {
             <Dropdown
               menu={{
                 items: [
+                  {
+                    key: 'similar-freight',
+                    label: 'Create similar freight',
+                    icon: <FontAwesomeIcon icon={faMagicWandSparkles} />,
```
rpelczar: Can we use Plus icon.

rpelczar (approved):
LGTM

I wonder if "similar" is the right term. Isn't "based on" more accurate?

Marvin9:
> I wonder if "similar" is the right term. Isn't "based on" more accurate?

"based on" sounds bit weird. What about "Duplicate freight" or "Copy freight" or "Clone freight"?

rpelczar:
> based on" sounds bit weird. What about "Duplicate freight" or "Copy freight" or "Clone freight"?

I like "Clone freight". We should also update at least the query parameter.

Marvin9:
Ok so "Clone freight" and query parameter would be ?clone-freight=\<alias\>. @rpelczar

krancour:
@rpelczar I see you approved, but didn't merge. I didn't want to go ahead and do so in case there was some reason you hadn't. Leaving it up to you.

Issue #4793, opened 2025-08-07
forceString for helm values in helm template step


johnthompson-ybor:
I'm trying to set a value in a helm template step which is json, and that breaks the sort of naive helm template --set logic that the helm template step uses -- Argocd has a forceString option for helm values, and I think a similar option is needed here.

johnthompson-ybor:
https://argo-cd.readthedocs.io/en/latest/user-guide/application-specification/ -- their spec documentation is here.

hiddeco:
Have you tried making use of [`quote()`](https://docs.kargo.io/user-guide/reference-docs/expressions/#quotevalue) or [`unsafeQuote()`](https://docs.kargo.io/user-guide/reference-docs/expressions/#unsafequotevalue)?

johnthompson-ybor:
That won't help.  It already _is_ a string, it's just an error because of the way that `helm template` is being used by kargo.  It tries to parse commas, etc... the only way to fix it is use --set-string instead of --set when running the helm-template.

johnthompson-ybor:
It has to do with the the way that helm template handles anything that looks like json when you 'set' a value -- if it looks like json it'll try to parse it as json, doesn't really matter how you try to escape it -- what you need to do is to run helm template with --set-literal when forceString is set.

johnthompson-ybor:
{"issuer":"foo","authorizationEndpoint":"bar","tokenEndpoint":"baz","userInfoEndpoint":"bat","secretName":"foo","authenticationRequestExtraParams":{"audience":"foo","organization-id":"foo"}

Try setting this is a value in a helm template, as an example --depending on how you escape it, it'll either just error out completely or try and parse it as json badly instead of treating it like a string (which will only be noticeable if you try to use it as an annotation, for example).  --set-literal just works the way you'd expect.

johnthompson-ybor:
I'm going to give you a simpler test case:

name: foo
value: foo,bar,baz,bat

This will fail on the helm template step, quoting or not quoting.  You can escape the commas and it will work, but having a forceString option that causes helm to use the correct --set-literal flag instead of --set will fix this issue completely.

johnthompson-ybor:
I opened a PR for this.

PR #4921, opened 2025-08-21
feat(promotion): allow forcing literal string for `helm-template` values

johnthompson-ybor:
This allows similar behavior to argo cd values so that when you use force string on a value is treats it as a string literal instead of something to be parsed further.

This is relevant for any value which contains commas, for example, or any value which is roughly json shaped, which helm will otherwise try to parse.

```
- name: comma_delimited_list
   value: foo,bar,baz
```

Will cause an error in the current code.

```
- name: comma_delimited_list
   value: foo,bar,baz
   forceString: true
``` 

Will work appropriately after this PR is merged.  This basically duplicates the existing ArgoCD spec for key/value pairs in helm charts.

### DIFF
--- a/pkg/promotion/runner/builtin/helm_template_runner.go
+++ b/pkg/promotion/runner/builtin/helm_template_runner.go
@@ -187,18 +187,22 @@ func (h *helmTemplateRunner) composeValues(
 		}
 		valueOpts.ValueFiles = append(valueOpts.ValueFiles, absValuesPath)
 	}
-	
+
 	// Process setValues, separating forceString values from regular values
 	for _, setValue := range cfg.SetValues {
-		if setValue.ForceString {
-			// When forceString is true, use --set-literal behavior
-			valueOpts.LiteralValues = append(valueOpts.LiteralValues, fmt.Sprintf("%s=%s", setValue.Key, setValue.Value))
-		} else {
-			// Default behavior uses --set which allows type inference
-			valueOpts.Values = append(valueOpts.Values, fmt.Sprintf("%s=%s", setValue.Key, setValue.Value))
+		if setValue.Literal {
+			// When literal is true, use --set-literal behavior.
+			valueOpts.LiteralValues = append(
+				valueOpts.LiteralValues,
+				fmt.Sprintf("%s=%s", setValue.Key, setValue.Value),
+			)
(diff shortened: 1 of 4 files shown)


johnthompson-ybor:
This fixes: https://github.com/akuity/kargo/issues/4793

krancour:
@johnthompson-ybor please be patient. We have a few team members on PTO, so we have a bit of a backlog on reviews. Reviews are also prioritized on the basis of our own, internal priorities. We will get to this when we are able.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/helm-template.md
```diff
@@ -31,6 +31,7 @@ commonly preceded by a [`git-clear` step](git-clear.md) and followed by
 | `setValues` | `[]object` | N | Allows for amending chart configuration inline as one would with the `helm template` command's `--set` flag. |
 | `setValues[].key` | `string` | N | The key whose value should be set. For nested values, use dots to delimit key parts. e.g. `image.tag`. |
 | `setValues[].value` | `string` | N | The new value for the key. |
+| `setValues[].forceString` | `boolean` | N | Whether to force the value to be treated as a string. When true, uses `--set-literal` instead of `--set`. This is `false` by default. |
```
hiddeco, line 34: Would it be more intuitive to name this `literal` to match the Helm flag?

```suggestion
| `setValues[].literal` | `boolean` | N | Whether to force the value to be treated as a string. When true, uses `--set-literal` instead of `--set`. This is `false` by default. |
```

johnthompson-ybor:
I wanted to match the argocd spec

krancour:
> I wanted to match the argocd spec

As a general rule, we try to match the vernacular of the tool or tech we're integrating with.

Argo CD doesn't have any relevance to the `helm-template` step, but `helm` has.

krancour:
This doesn't seem like it will make the cut for v1.8.0.

Moving to v1.9.0.

@johnthompson-ybor as this feature was unsolicited, the PR is likely to eventually be closed if it's been abandoned.

hiddeco:
Closing in favor of #5470. Thank you @johnthompson-ybor 🙇

Issue #4753, opened 2025-08-04
UI: more info in stage node


Marvin9:
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Stage node should show more info based on stage status and health

1. When promoting, it can show current promotion step and deep link (if available)

<img width="1525" height="816" alt="Image" src="https://github.com/user-attachments/assets/6deb25b5-f51e-42f5-9ed6-e97b473c22ed" />

2. When promotion failed, it can show the failed step

<img width="1530" height="738" alt="Image" src="https://github.com/user-attachments/assets/41d5b915-f353-4fa8-a7a1-ae07e72bd749" />

3. When stage is failed state due to verification failure, it can show summary and/or logs link

<img width="2382" height="804" alt="Image" src="https://github.com/user-attachments/assets/cc8f69eb-6e1c-497c-89f3-ad89252abd69" />

# Motivation

PR #4923, opened 2025-08-22
feat(ui): stage node links

Marvin9:
https://github.com/akuity/kargo/issues/4753

Screenshot 2025-08-27 at 4 00 18 PM (2)
Screenshot 2025-08-27 at 4 00 10 PM (2)


Marvin9:
> I'm just worried we want to add too much information to the node

True but its not always crowded, only when there are some action required, and I believe it is useful in that case

PR #4932, opened 2025-08-25
feat: exclude control flow stages from health summary stats

Marvin9:


internal/controller/management/projects/stats_test.go
```diff
@@ -228,10 +272,41 @@ func Test_reconciler_collectStats(t *testing.T) {
...
+					},
+				},
+			).Build(),
+			assertions: func(t *testing.T, status kargoapi.ProjectStatus, err error) {
+				require.NoError(t, err)
+				require.Nil(t, conditions.Get(&status, kargoapi.ConditionTypeHealthy))
+				stats := status.Stats
+				require.Equal(t, int64(0), stats.Stages.Count)
+			},
+		},
```
krancour: I would say that since the previous test case has also been modified to include a control flow stage and assert that the control flow stage is not counted, our bases are already well covered and this new case doesn't add much.

internal/controller/management/projects/stats_test.go
```diff
@@ -171,6 +171,17 @@ func Test_reconciler_collectStats(t *testing.T) {
...
 						Namespace: testProject,
 					},
+					Spec: kargoapi.StageSpec{
+						PromotionTemplate: &kargoapi.PromotionTemplate{
+							Spec: kargoapi.PromotionTemplateSpec{
+								Steps: []kargoapi.PromotionStep{
+									{
+										Uses: "git-clone",
+									},
+								},
```
krancour: I recommend that we always try to use the absolute minimum configuration required when trying to engineer specific scenarios in our test cases. This doesn't need to pass validation, and the mere presence of a single step make this _not_ a control flow stage, so we can get by with saying "ya there's a step" and we don't need to provide any further detail. Same suggestion applies in a few other spots.

```suggestion
								Steps: []kargoapi.PromotionStep{{}},
```

krancour:
Thank you, @Marvin9! I made a few small suggestions, but preemptively LGTM.

krancour (approved):

internal/controller/management/projects/stats_test.go
Marvin9: Agreed but control flow stage condition satisfies only when steps length > 0

krancour: `[]kargoapi.PromotionStep{{}},` is one step. It's just a step with no details filled out.

Marvin9: Oh okay got it!

PR #4933, opened 2025-08-25
fix(auth): redirect back to link after login

Marvin9:


krancour:
@Marvin9 I was just test driving this and, unless I am misunderstanding what the improvement is here, I think it's not working for me.

If I am not logged in, paste a deep link URL like `http://localhost:30082/project/kargo-demo` into the address bar, then complete login, I am still landing on the usual home page / list of Projects instead of being redirected to where I expected.

I tried with multiple browsers, too, fwiw.

Marvin9:
Weird. Tested in safari as well. Can you try again?

https://github.com/user-attachments/assets/29480bad-2263-46a4-808b-1aac3b023657

rpelczar (approved):
LGTM

krancour:
@Marvin9 I figured something out... it works if you are logging in as admin, but isn't working if you log in using SSO.

Marvin9:
Fixed @krancour PTAL. Also @rpelczar since you worked on this flow

krancour:
@Marvin9 perfect! 🔥

ui/src/features/auth/oidc-login.tsx
```diff
@@ -84,6 +85,7 @@ export const OIDCLogin = ({ oidcConfig }: Props) => {
     sessionStorage.setItem(codeVerifierKey, code_verifier);
     const state = generateRandomState();
     sessionStorage.setItem(stateKey, state);
+    sessionStorage.setItem(platformRedirectKey, window.location.search);
```
rpelczar, line 88: Why do we need to store url in sessionStorage? Can't we add this to redirectURI?

Marvin9: No. That was what I tried at first but it will give OIDC error due to redirect uri mismtach

Issue #4899, opened 2025-08-19
artifactory webhook receiver: honor an optional, custom http header containing registry base url info


ggogel:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

We've set up an Artifactory webhook on the cluster level. In the debug logs of the external webhook server, we can see the webhook request from our Artifactory, including the payload, but Kargo doesn't find any warehouses to refresh.

# Steps to Reproduce

1. Follow the [Cluster Level Configuration](https://docs.kargo.io/operator-guide/cluster-configuration#triggering-artifact-discovery-using-webhooks) to setup the Clusterconfig with an `artifactory` webhook.
```
apiVersion: kargo.akuity.io/v1alpha1
kind: ClusterConfig
metadata:
  name: cluster
spec:
  webhookReceivers:
    - name: artifactory-cluster-receiver
      artifactory:
        secretRef:
          name: artifactory-cluster-secret
```
2. Follow [Registering with Artifactory](https://docs.kargo.io/user-guide/reference-docs/webhook-receivers/artifactory/#registering-with-artifactory)

3. Set up a Warehouse:
```
apiVersion: kargo.akuity.io/v1alpha1
kind: Warehouse
metadata:
  name: myorg-tcd
spec:
  subscriptions:
  - image:
      repoURL: artifactory.mycompany.local/docker-myorg-local/myorg-tcd/myorg-tcd-frontend
      discoveryLimit: 20
      imageSelectionStrategy: NewestBuild
      insecureSkipTLSVerify: true
  - image:
      repoURL: artifactory.mycompany.local/docker-myorg-local/myorg-tcd/myorg-tcd-backend
      discoveryLimit: 20
      imageSelectionStrategy: NewestBuild
      insecureSkipTLSVerify: true
  interval: 6h0m0s
```

4. Enable debug logs for the webhook receiver.

5. Warehouse should refresh, but doesn't.  `msg="found Warehouses to refresh" count=0`

One thing I noticed:
`repoURL` from the webhook payload does not include the full URL.

# Version

```
v1.7.3
```

# Logs

```
[...]

PR #4945, opened 2025-08-27
feat: support self-hosted artifactory users

fuskovic:
Closes: https://github.com/akuity/kargo/issues/4899

### DIFF
--- a/internal/webhook/external/artifactory.go
+++ b/internal/webhook/external/artifactory.go
@@ -27,6 +27,7 @@ const (
 	artifactoryAuthHeader      = "X-Jfrog-Event-Auth"
 	artifactorySecretDataKey   = "secret-token"
 	artifactory                = "artifactory"
+	artifactoryRepoURLHeader   = "X-Kargo-Repo-URLs"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/artifactory/index.md
```diff
@@ -141,6 +141,18 @@ kubectl get projectconfigs kargo-demo \
...
 
+    1. Optionally, in the <Hlt>Headers</Hlt> section, you can set a custom `X-Kargo-Repo-URLs` header.
+
+        :::info
+        The value of `X-Kargo-Repo-URLs` can either be a single repository URL
+        or a comma-separated list of repository URLs. If set, any warehouses 
+        with subscriptions to the designated repository URL(s) will be
+        refreshed. This can be useful for repositories with unconventional 
+        naming schemes or self-hosted instances.
+        :::
```
krancour, line 152: A few things I would worry about:

If you register one webhook for many repos, you may have a lot of different repos to list here as part of the header value. Apart from being onerous, my worry is that it leads to unnecessary refresh of many Warehouses.

If wishing to avoid the problem above, it would compel a user to register many webhooks for one repo apiece, with just one repo URL in the header. This also seems onerous.

I'd initially been thinking we were only going to provide a way to relay the information that may be missing from the request's `jpd_origin` field. i.e. A way to say, "here's the base domain name for this entire registry" and leave it at that.

But before we get to far into what the implementation options are here, I think we should wait for a response to [this comment](https://github.com/akuity/kargo/issues/4899#issuecomment-3228943202), because if an empty `jdp_origin` field is the result of misconfiguration (or if not _mis_-configuration, still a problem that can be resolved _via_ configuration) then we may not need a new header. We may be able to get by with simply adding a note to the docs about the need for `jdp_origin` to not be empty and how to configure that correctly.

We _may_ still need to do something with a new header, but let's table it until we have an answer to that comment.

fuskovic: We had a user that had the repo key prepended as the subdomain. The one that was building a middleware service to intercept and modify the request before proxying to Kargo. The `jdp_origin` field was set in his situation but that still doesn't solve his problem. Or other users in a similar situation for that matter. I think it's still a viable option for these kind of situations since they have no alternative.

krancour: We spoke offline, but to summarize here for anyone following along, I agree that what you've PR'ed here solves for tricky cases that my original idea doesn't. So I'm not completely closed off to the idea. But I do want us to get more information about configuration (or mis-configuration, as the case may be) of self-hosted Artifactory and, ideally, get hands-on experience with it too before we commit to moving in any specific direction here.

fuskovic: We've received the additional configuration information on self-hosted instances and I've documented them in https://github.com/akuity/kargo/pull/5103 to help solve cases of missing `jdp_origin` field.

In the other cases where users are using virtual repositories and/or repo keys as subdomains I feel this PR still stands as a viable solution. Wdyt?

fuskovic: Per our offline conversation, will table this for now as a custom base url configuration _may_ satisfy both edge cases described in https://github.com/akuity/kargo/issues/4899

krancour: I'll close this for now to get it off our radar, but if we need to come back to this, we know it's here.

PR #4962, opened 2025-08-30
Support bitbucket server

### LINKED ISSUE #4944: Bitbucket Server external webhooks
# Support external webhook payloads from Bitbucket Server

Today, the [bitbucket payload](https://github.com/akuity/kargo/blob/e84907a1c3136b444090046eb5b2906c99356939/internal/webhook/external/bitbucket.go#L25) only suppors the _Bitbucket Cloud_ event of "repo:push".  The Bitbucket Server / Datacentre edition uses "repo:refs_changed" and has a slightly different structure of the events.

Kargo ought to support Bitbucket Server in addition.

# Documentation

* [Bitbucket Cloud event payload](https://support.atlassian.com/bitbucket-cloud/docs/event-payloads/#Push)
* [Bitbucket Data Center event payload](https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html#Eventpayload-repo-push) (version 9.6)

The main difference is in where to find the name of the repository

| What changed | Bitbucket Cloud | Bitbucket Data Center |
|--------|--------|--------|
| Event name | `repo:push` | `repo:refs_changed` | 
| Name of pushed branch | `push.changes[].new.name` | `changes[].ref.displayId` |
| Repository links | `repository.links.html.href` | `repository.links.clone[].href` or `repository.links.self[].href` | 

I'm not quite sure what the _repository links_ is used for, since the Bitbucket Cloud variant would return `"https://api.bitbucket.org/bitbucket/bitbucket"` and in Datacentre, you (often?) get two clone links (ssh and https) and a self link.  Not entirely sure what its purpose is in it all.  Should we try to refresh them all?

```
"links": {
  "clone": [
    {
      "href": "ssh://git@example.com:7997/project/repository.git",
      "name": "ssh"
    },
    {
      "href": "https://example.com/scm/project/repository.git",
      "name": "http"
    }
  ],
  "self": [
    {
      "href": "https://example.com/projects/project/repos/repo/browse"
    }
  ]
}
```

I'm interested in supporting this with code if this is desirable.  It's all in a nice bitbucket.go file, and I think few modifications are necessary for the code to support both code paths.

mogsie:
Expand the Bitbucket external webhooks to support bitbucket  "Data Center" edition (aka Bitbucket Server).

The format is slightly different, so the JSON payload is captured into a different struct based on the type of event:

* "repo:push" is from Bitbucket cloud
* "repo:refs_changed" is from Bitbucket Server / Data Center.

Resolves #4944.


internal/webhook/external/bitbucket.go
```diff
@@ -22,7 +23,8 @@ const (
 	bitbucketEventHeader     = "X-Event-Key"
 	bitbucketSignatureHeader = "X-Hub-Signature"
 
-	bitbucketPushEvent = "repo:push"
+	bitbucketPushEvent       = "repo:push"
+	bitbucketServerPushEvent = "repo:refs_changed" // Bitbucket Server (Data Center
```
mogsie: I could call this "bitbucketRefsChangedEvent" instead, but I felt it made sense to name it after the product.

Should I rename the `bitbucketPushEvent` to `bitbucketCloudPushEvent` for completeness too?

internal/webhook/external/bitbucket.go
```diff
@@ -179,11 +177,75 @@ type bitBucketPushEvent struct {
...
+}
+
+// getRepoURLs extracts the Bitbucket "clone" URLs. Typically,
+// this includes both ssh and https URLs.
+func (b bitBucketServerPushEvent) getRepoURLs() []string {
+	repoURLs := []string{}
+	for _, link := range b.Repository.Links.Clone {
+		// Internally, the warehouse subscription normalizes URLs by stripping
+		// .git suffixes, so we do the same here.
+		hrefWithoutGitSuffix := strings.TrimSuffix(link.Href, ".git")
```
mogsie: This was required for the warehouses to recognise the URLs.  :shrug:

fuskovic: For normalizing git URL's, we use `git.NormalizeURL`.

internal/webhook/external/bitbucket.go
fuskovic: I think I prefer `bitbucketRefsChangedEvent`. To provide additional context, you can add a comment detailing the discrepancy for Bitbucket Server; adding a link to the [payload schema](https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html).

internal/webhook/external/bitbucket.go
```diff
@@ -179,11 +177,75 @@ type bitBucketPushEvent struct {
...
+		Ref struct {
+			DisplayID string `json:"displayId"`
+		} `json:"ref"`
+	} `json:"changes"`
+	Repository struct {
+		Links struct {
+			Clone []struct {
+				Href string `json:"href"`
+			} `json:"clone"`
+		} `json:"links"`
```
fuskovic, line 239: Interestingly enough, in the Bitbucket Server docs, I'm not seeing a `Repository.Links` array in the push event payload schema but I do see it for other event types. 

I do however see `Repository.Name`. 

Was there a good reason for not using that?

Could I ask you to provide a sample event payload so I could see what `Repository.Name` yields?

internal/webhook/external/bitbucket.go
nitishfy: Please add a better comment at the top explaining this field.

internal/webhook/external/bitbucket.go
mogsie: Of course.  @krancour had the same question in #4944.  The documentation states that Repository is a common entity, but I cannot find the common entity documentation at all.  I can only find it for Bitbucket cloud, but that's no help.

*Edit 2*: The Bitbucket documentation page states at the top:

> The following payloads contain some of the common entity types – User, Repository,  Comment, and Pull Request – which have consistent representations in all the payloads where they appear. For example, the actor property in the repo:refs_changed payload is a representation of the event's user. 

*Edit*: As for `Repository.Name` I was essentially just mimicking what the Bitbucket Cloud code was doing; I assume it is matching the links of repository URLs against warehouse subscription URLs, and both http and ssh are possible to use in subscriptions, so I figured it would be wise to invalidate both.  Repository name is not enough, is it?

<details>
  <summary>Here's a redacted payload from our system with some irrelevant data omitted.</summary>

```
{
  "date": "2025-08-27T15:02:14+0200",
  "actor": {
    "emailAddress": "example@my-org.example.com",
    "displayName": "Example Lastname",
    "name": "example",
    "active": true,
    "links": {
      "self": [
        {
          "href": "https://jira.my-org.example.com/stash/users/example"
        }
      ]
    },
    "id": 1234,
    "type": "NORMAL",
    "slug": "example"
  },
  "toCommit": {
    "committer": {
      "emailAddress": "example@my-org.example.com",
      "displayName": "Example Lastname",
      "name": "example",
      "active": true,
      "links": {
        "self": [
          {
            "href": "https://jira.my-org.example.com/stash/users/example"
          }
        ]
      },
      "id": 1234,
      "type": "NORMAL",
      "slug": "example"
    },
    "committerTimestamp": 1756299732000,
    "author": {
      "emailAddress": "example@my-org.example.com",
      "displayName": "Example Lastname",
      "name": "example",
      "active": true,
      "links": {
        "self": [
          {
            "href": "https://jira.my-org.example.com/stash/users/example"
          }
        ]
      },
      "id": 1234,
      "type": "NORMAL",
      "slug": "example"
    },
    "authorTimestamp": 1756299732000,
    "id": "269e80376eb881afe62ebe71daa6b9d523db19b7",
    "displayId": "269e80376eb",
    "message": "full commit message ...",
    "parents": [
      {...}
    ]
  },
  "eventKey": "repo:refs_changed",
  "changes": [
    {
      "ref": {
        "id": "refs/heads/feature/EXAMPLE-BRANCH",
        "displayId": "feature/EXAMPLE-BRANCH",
        "type": "BRANCH"
      },
      "fromHash": "0000000000000000000000000000000000000000",
      "toHash": "269e80376eb881afe62ebe71daa6b9d523db19b7",
      "refId": "refs/heads/feature/BRANCH-1234",
      "type": "ADD"
    }
  ],
  "commits": [
    {...},
    {...},
    {...},
  ],
  "repository": {
    "archived": false,
    "public": false,
    "hierarchyId": "6f4952ae7a067ffb045f",
    "name": "somerepo",
    "forkable": true,
    "project": {
      "public": false,
      "name": "Example Project Name",
      "description": "Description of the project.",
      "links": {
        "self": [
          {
            "href": "https://jira.my-org.example.com/stash/projects/SOMEPROJ"
          }
        ]
      },
      "id": 925,
      "type": "NORMAL",
      "key": "SOMEPROJ"
    },
    "links": {
      "clone": [
        {
          "name": "http",
          "href": "https://jira.my-org.example.com/stash/scm/someproj/somerepo.git"
        },
        {
          "name": "ssh",
          "href": "ssh://git@bitbucket.my-org.example.com:7999/someproj/somerepo.git"
        }
      ],
      "self": [
        {
          "href": "https://jira.my-org.example.com/stash/projects/SOMEPROJ/repos/somerepo/browse"
        }
      ]
    },
    "id": 2438,
    "scmId": "git",
    "state": "AVAILABLE",
    "slug": "somerepo",
    "statusMessage": "Available"
  }
}

```
</details>

internal/webhook/external/bitbucket.go
mogsie: I renamed as follows:

* bitbucketServerPushEvent to bitbucketRefsChangedEvent
* The two structs (oddly capitalized) "bitBucketServerPushEvent" to "bitbucketRefsChangedEvent**Payload**"

And added a comment to each of these strings to indicate what they're used for.

mogsie:
> ⚠️ Report is 4 commits behind head on main.

Do I have to worry about this and keep on rebasing it?

internal/webhook/external/bitbucket.go
fuskovic: I see. Yes, I've reviewed [#4944](https://github.com/akuity/kargo/issues/4944) and understand.

I could not find a common `Repository` definition. 

I would typically assume the `Repository` definition for cloud [here](https://support.atlassian.com/bitbucket-cloud/docs/event-payloads/#Repository) would be identical to the one for server solely based on the "common entities" language. However, in that definition I notice a `full_name` field and a `website` field. This is curious to me because:

1. I wonder if you omitted those from your example? If not, that could mean the `Repository` object defined for cloud is indeed not identical to server. Can you confirm?

2. Those fields together seem like could have potential for resolving the URL based on how those fields are described. e.g.

**website**
```
The URL to the repository's website when the code is hosted for a specific website.
```

**full_name**
```
The workspace and repository slugs joined with a '/'.
```

mogsie: > I could not find a common Repository definition.

No, neither could I, only the one in the Cloud documentation, which doesn't match the Bitbucket Data Center examples (or the real payloads I see).

I can confirm that neither `website` nor `full_name` appear in the Repository structure. I of course did not remove anything from repository, since it's central to this discussion.  The only thing I did omit were the details on each commit (`{...}`) which included things like committers, authors, parents and so on.

Aren't these URLs used to be compare with the _git_ repoURL of a warehouse?

fuskovic: > Aren't these URLs used to be compare with the git repoURL of a warehouse?

Yes

> I can confirm that neither website nor full_name appear in the Repository structure. I of course did not remove anything from repository, since it's central to this discussion. The only thing I did omit were the details on each commit ({...}) which included things like committers, authors, parents and so on.

Thanks!

internal/webhook/external/bitbucket.go
```diff
@@ -179,11 +185,75 @@ type bitBucketPushEvent struct {
...
+	} `json:"repository"`
+}
+
+// getRepoURLs extracts the Bitbucket "clone" URLs. Typically,
+// this includes both ssh and https URLs.
```
fuskovic: ssh and https details aren't necessary if we explain the repo URLs are already normalized.

```suggestion
// getRepoURLs returns a set of normalized repoURLs from the event.
```

internal/webhook/external/bitbucket_test.go
```diff
@@ -40,6 +40,26 @@ func TestBitbucketHandler(t *testing.T) {
 		}
 	}`
 
+	const pushEventRequestBodyBitbucketServer = `
+	{
+		"actor": {
+			"name": "admin",
+			"emailAddress": "admin@example.com"
+		},
+		"changes": [{"ref": {"displayId": "refs/heads/main"}}],
```
fuskovic: From your sample payload in [this comment](https://github.com/akuity/kargo/pull/4962/files#r2316754455):

```
"ref": { 
    "id": "refs/heads/feature/EXAMPLE-BRANCH", 
    "displayId": "feature/EXAMPLE-BRANCH", 
},
```

But in this test payload you have a `refs/heads/` prefix in your `displayId` value which will not be present according to your sample live payload.

This would fail to accurately match refs because this ref would be considered invalid without that prefix: https://github.com/akuity/kargo/blob/main/internal/controller/git/commit/newest_from_branch_selector.go#L71

fuskovic: So it looks like we should be using the `id` here when collecting refs. Not the `displayId`.

fuskovic (changes requested):
Other than the wrong field being used for refs.

I'd also like to see some screenshots or a video showing this working end-to-end.

krancour:
> Do I have to worry about this and keep on rebasing it?

It shouldn't be necessary.

internal/webhook/external/bitbucket.go
krancour: Skimming this, I think it just underscores the importance of good documentation. 🤷‍♂️

internal/webhook/external/bitbucket.go
```diff
@@ -88,9 +98,19 @@ func (b *bitbucketWebhookReceiver) getHandler(requestBody []byte) http.HandlerFu
 			return
 		}
 
+		// Bitbucket Cloud and Data Center provide slightly different event
+		// payloads. This interface describes what we need from either.
+		var payload interface {
+			getRefs() []string
+			getRepoURLs() []string
+		}
```
krancour, line 106: Love this.

mogsie:
> Other than the wrong field being used for refs, I'd also like to see some screenshots or a video showing this working end-to-end.

I'd be glad to run it in our internal environment. Is there an easy way for me to build or grab an image that is a drop-in replacement?

internal/webhook/external/bitbucket_test.go
mogsie: Sorry about this.  This was sloppy from my side.  I didn't use "real" test data when writing the tests, just restructured the existing test payload to match the testdata warehouses that were part of the existing test.  I will fix by using more real-world payloads.

internal/webhook/external/bitbucket.go
mogsie: I'm sorry for being somewhat unfamiliar with the codebase, but I don't understand the concept of normalization.  Is normalization able to handle the fact that "https://domain1.example.org/stash/blah.git" is somehow _the same repository_ as "ssh://git@otherdomain.example.org/blah.git"?

I realize that when the domain is *bitbucket.org* that you can make out-of-band assumptions about the structure of URLs.  But Bitbucket Data Center lets admins use _different hosts_ for http and clone URLs.

krancour:
I do something like this on the rare occasion that I need to make a custom build available to a remote cluster:

```
IMAGE_PUSH=true IMAGE_REPO=krancour/kargo IMAGE_TAG=custom make hack-build
```

internal/webhook/external/bitbucket.go
krancour: You've got the general idea correct.

If two URLs are different, but equivalent, their normalized forms should be exactly equal. So, in practice, URLs are always normalized before any kind of lookup, search, comparison, etc.

The classic example I use is that `https://github.com/example/repo` and `https://github.com/example/repo.git` both represent the same resource (repository). 

Things get even more interesting with images, where `busybox` could also be written as `index.docker.io/library/busybox` (and about a half dozen other ways as well).

Ironically, the example you cited, comparing equivalent https and ssh URLs, is actually _not_ yet accounted for by normalization. That case has been a bit of a head scratcher precisely because of what you said. Certain assumptions can be made for popular cloud-based platforms, but don't necessarily hold up in the general case.

mogsie:
Thanks @krancour. I managed to build it and put it into our test environment, and managed to get _some_ events through, even some that "refreshed 1 warehouse."  However, some events failed to get through; and some responded (correctly) with refreshing "0 warehouses" (e.g. due to pushes to other branches).  I'm not sure if I made some mistake in the set-up, or if there's an internal error in our Bitbucket somehow, or something wrong with the cluster's ingress.  I'll do more testing on Monday.  The only change I made was to replace the image tag with my custom image; is that enough?

**Edit** Quick update to mention that my ingress is logging (nginx) response code 499, which is (IIRC) a response code indicating that the client disconnected before the response was available.  Perhaps this is an (unrelated) problem with the external webhooks receiver?

**Edit 2**: 

I made a commit to my repository at 20:07:33 (using the web UI, so the commit date is the same as when it was available for Bitbucket to act on). 

The webhook request is dated (I'm in UTC +0200) in the Bitbucket diagnostics.

```
{
    "date": "2025-09-08T20:07:33+0200",
    "actor": {
```

The ingress logs the HTTP 499 after 20 seconds, so presumably something timed out.

```
11.22.33.44 - - [08/Sep/2025:18:07:53 +0000] "POST /webhooks/bitbucket/REDACTEDHEXSTRING HTTP/1.1" 499 0 "-" "Atlassian HttpClient 4.2.1 / Bitbucket-9.4.9 (9004009) / Default" kong_request_id: "40c5a830ed946c7e26f741474bfdac78"
```

It seems as if the webhook is synchronously refreshing the warehouse(s), in a for-loop.  I _believe_ the timeout might cause the context to be cancelled.  I can try to increase verbosity to see if I can see what is happening.  It's odd because I expect only one or two warehouses to be refreshed.

**Edit 3:**

I reproduced my problem with KLOG_LEVEL set to 8, and as far as I could tell, the external webhooks receiver did not log anything.  Only outbound requests seem to be logged (e.g. querying the k8s API), including retrieving the webhook secret, and PATCHing the warehouse.  They are missing when this problem occurs.  Therefore, I'm not sure the webhook makes it through to the external webhooks receiver at all.  The problem may well be in our networking infrastructure.

When the freights _are_ refreshed (`{"msg": "refreshed 1 warehouse(s)"}`), the latency is generally in the tens of milliseconds, as measured by bitbucket, and 3-5 milliseconds measured by our ingress.

mogsie:
The problems were indeed on our networking.  It's running quite smoothly now with my patched 1.7.3 webhook image.

> Other than the wrong field being used for refs, I'd also like to see some screenshots or a video showing this working end-to-end.

Here's a [video of the thing running end-to-end](https://drive.google.com/file/d/1dlK2qL9YnubsIm12AIZu1Xa5RJamNGPN/view?usp=sharing), as requested @fuskovic

krancour:
Thanks @mogsie!

PR #4974, opened 2025-09-02
feat(promotion): add support for submodules to `git-clone` step

### LINKED ISSUE #4755: Support submodules in git-clone
# Checklist

* [x ] I've searched the issue queue to verify this is not a duplicate feature request.

# Proposed Feature

Currently, the git-clone step does not seem to support submodules. Please add the --recurse-submodules flag as default behavior, or at least add a boolean config option to the git-clone step which would allow us to toggle submodule initialization or not.

# Motivation

Without this, we are forced to manually checkout any submodules in the main repo through a separate set of git-clone steps and we must also translate the .gitmodules file into a yaml or json so it can be parsed in Kargo steps. It would be much simpler to enable support for initializing submodules in git-clone.

kyounghunJang:
## Summary
Add submodule support to git-clone step

## Related Issue
fixes #4755

## Description
Currently, the git-clone step does not support submodules.  
Users must manually checkout submodules in separate steps and parse `.gitmodules`,  
which introduces unnecessary complexity.  

This PR adds support for initializing submodules directly within git-clone.

## Proposed Changes
- Added a new boolean config option `recurseSubmodules` (default: false)

### DIFF
--- a/pkg/promotion/runner/builtin/git_cloner.go
+++ b/pkg/promotion/runner/builtin/git_cloner.go
@@ -188,8 +188,7 @@ func (g *gitCloner) run(
 				)
 		}
 		if cfg.RecurseSubmodules {
-			err = worktree.UpdateSubmodules()
-			if err != nil {
+			if err = worktree.UpdateSubmodules(); err != nil {
 				return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
 					fmt.Errorf("error updating submodules for worktree at %s: %w", path, err)
 			}
--- a/pkg/promotion/runner/builtin/git_cloner_test.go
+++ b/pkg/promotion/runner/builtin/git_cloner_test.go
@@ -405,14 +405,16 @@ func Test_gitCloner_run_with_submodules(t *testing.T) {
 	)
 	require.NoError(t, service.Setup())
 	server := httptest.NewServer(service)
-	defer server.Close()
+	t.Cleanup(server.Close)
 
 	// Create submodule remote repo and push a file
 	subRepoURL := fmt.Sprintf("%s/sub.git", server.URL)
 	subRepo, err := git.Clone(subRepoURL, nil, nil)
 	require.NoError(t, err)
-	defer subRepo.Close()
-	err = os.WriteFile(filepath.Join(subRepo.Dir(), "sub.txt"), []byte("sub"), 0600)
+	t.Cleanup(func() {
+		_ = subRepo.Close()
+	})
+	err = os.WriteFile(filepath.Join(subRepo.Dir(), "sub.txt"), []byte("sub"), 0o600)
 	require.NoError(t, err)
 	err = subRepo.AddAllAndCommit("Initial commit sub", nil)
 	require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


hiddeco:
I think additional changes would be required to ensure credentials are properly passed, especially when each Git repository requires a different set of credentials. Which is a hypothetical possibility.

kyounghoonJang:
@hiddeco 
Hello, thank you for the feedback.

Currently, this PR is designed to only allow cloning submodules that share the same credentials as the main repository. I fully recognize and agree with the need to support multiple sets of credentials.
If we were to support multiple authentications, I think the following changes would be required. 

1. Currently, buildGitCommand handles HTTP using GIT_ASKPASS and GIT_PASSWORD. This is expected to only work for a single set of credentials. Therefore, I think this should be changed to use the .git-credentials mechanism instead.

2. To obtain the submodule RepoUrl, we need to perform a bare clone, then parse the .modules file, and store that information into the ClientOptions struct under a Credentials field (or SubCredentials). After that, we should fetch the correct credentials from the secret database.

3. I believe we also need a module that writes the retrieved secret information into .git-credentials.

Please let me know if my thoughts and design approach are correct. Thanks!

krancour:
fwiw, I think doing this with the limitation of only using the same credentials as the main repository seems like it _might_ be a reasonable first step. I say _might_ because the devil's always in the details and I have not looked at this very closely.

> Currently, buildGitCommand handles HTTP using GIT_ASKPASS and GIT_PASSWORD. This is expected to only work for a single set of credentials. Therefore, I think this should be changed to use the .git-credentials mechanism instead.

Only a word of caution that messing with any of that shouldn't be undertaken lightly and I would prefer something like that be done by a maintainer. There are a lot of esoteric reasons that git auth in Kargo is implemented as it is. There are certain approaches that don't work within a container based on a distroless image, and admittedly, I have a pipe dream of one day not even including a shell in that image, and that further restricts which approaches are viable. I definitely do not want us to fall down this particular rabbit hole at this time.

kyounghoonJang:
@krancour 
Hello, thank you for your feedback.
If modifying the logic is too complex, I also believe that, as reflected in the current PR, it would be preferable to allow submodules to be fetched when they share the same authentication as the main repository. I have completed testing to confirm that submodules can be fetched successfully in cases where the authentication matches the main repository.

etetar:
@krancour  @kyounghunJang  Hi! Thank you for working on this. Any chance this will make it into 1.7.6 or 1.8.0?

internal/promotion/runner/builtin/schemas/git-clone-config.json
```diff
@@ -9,6 +9,10 @@
       "type": "boolean",
       "description": "Indicates whether to skip TLS verification when cloning the repository. Default is false."
     },
+    "recurseSubmodules": {
+      "type": "boolean",
+      "description": "Indicates whether to recursively clone submodules. Default is false."
```
hiddeco: Given Kent's comments, please ensure this indicates any credentials have to be valid for the submodule.

kyounghoonJang: @hiddeco 
Thank you for the feedback! I’ve made the changes.

PR #4979, opened 2025-09-03
docs: add AWS Cognito OIDC configuration guide

jessesuen:
This adds an AWS Cognito OIDC configuration guide under the OpenID Connect section. We can add more over time.


docs/docs/40-operator-guide/40-security/20-openid-connect/30-aws-cognito/_category_.json
krancour, line 1: As is, this file is having no effect.

I think you had probably intended for this to be one level higher at `docs/docs/40-operator-guide/40-security/20-openid-connect/_category_.json`?

That would hide the children of the OIDC docs by default, which I think had been your intent?

docs/docs/40-operator-guide/40-security/20-openid-connect/30-aws-cognito/index.md
```diff
@@ -0,0 +1,58 @@
...
+# AWS Cognito OIDC Configuration Guide
+
+This guide explains how to configure AWS Cognito as an OIDC provider for Kargo.
+
+## Step 1. Create a User Pool
+
+Cognito requires a **User Pool** for authentication.  Create a new User Pool 
+(or use an existing one if already set up).
+
+![Create User Pool](./img/create-user-pool.png)
```
krancour: When I test drove these directions, my experience was quite a bit different than this.

I had no existing user pools. When I landed on the Cognito home page, the only _obvious_ path was a big orange button labeled "Get started for free in less than five minutes":



Screenshot 2025-09-03 at 5 05 14 PM



If I click that, the initial setup it guides me through does not resemble where your docs and screen captures go from here...



It looks like this:



Screenshot 2025-09-03 at 4 57 49 PM



I found this form _very_ confusing, as it seemed to put creation of my app client _ahead_ of creating the directory (user pool). I think, in order to avoid readers accidentally going down this awkward path that doesn't match the docs, we may need to be more explicit about how to navigate to the page that lists user pools and sets readers on a path that matches what you have documented.

The way to reliably get to the point you're starting from is by clicking "User pools" in the sidebar menu that is _collapsed by default:_



Screenshot 2025-09-03 at 5 03 08 PM



My suggestion would be that we include expansion of the sidebar and selecting "User pools" from there in the first step.

krancour: Actually... wait... that awkward form is the same one you're referencing. Your screenshot was just zoomed in more on the form.

So creating a pool _always_ sends you down this awkward path of creating a pool and app all at once. ☹️ 

_However..._ if you're adding an app to an existing pool, the form is different:

Screenshot 2025-09-03 at 5 18 35 PM

So I still think people will get confused by:

1. The two different forms they will complete if it's a new pool vs adding an app to an existing pool.

1. The peculiarity of the app being created before/with the pool, if you are indeed creating a new pool.

🤦‍♂️ 

If you don't mind, I will make an attempt of my own to add some clarity to this first step.

krancour:
I fell way down the Cognito rabbit hole.

I refactored the directions so there are two versions of the early part of the directions depending on whether you are creating a new user pool or adding a client app to an existing user pool, since the flows for these are very different from one another.

Both sets of directions leave you in the same spot, and then we resume with one common set of directions.

I also modified formatting for consistency with how we walk users through steps in external systems (e.g. webhook registrations) throughout the rest of the docs.

Also fixed links that were accidentally broken on other pages when `docs/docs/40-operator-guide/40-security/20-openid-connect.md` was renamed to `docs/docs/40-operator-guide/40-security/20-openid-connect/index.md`.

Issue #4140, opened 2025-05-19
ui: icons are missing in the available freight list


caspar-ds:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

In v1.4 it was easy to see which version was for which type of freight as an icon was displayed next to the version identifier. v1.5 seems to be missing these icons.

# Screenshots

v1.4

![Image](https://github.com/user-attachments/assets/e786fac8-e8ee-4118-857f-1d8ce84d7d34)

v1.5 

![Image](https://github.com/user-attachments/assets/91fa2951-388f-411b-8934-ec6cb8bc2850)

# Steps to Reproduce

* Create a warehouse with multiple subscriptions (e.g. image/chart/repo)
* Note that there is nothing to indicate what type of artifact each version is from

# Version

```
Client Version: v1.5.0
```

# Logs

N/A

Marvin9:
For more sources, you can use filters. Otherwise we can have icons back cc @rpelczar 

<img width="827" alt="Image" src="https://github.com/user-attachments/assets/ee886f87-3ae0-4105-90c6-db64be7a6142" />

blakestoddard:
IMO the order of precedence for what to display for commits in a freight should be tag then commit. (one of my users reported the tag -> commit change in 1.5.x today, too)

caspar-ds:
Have removed the request for semver to be displayed and added to own issue here as they are two distinct problems and arguably the missing semver is a bigger issue:
https://github.com/akuity/kargo/issues/4564

PR #4983, opened 2025-09-03
feat(ui): show artifact icons and have filter for it

Marvin9:
fixes https://github.com/akuity/kargo/issues/4140

Screenshot 2025-09-03 at 4 19 44 PM (2)

### DIFF
--- a/ui/src/features/project/pipelines/freight/freight-artifact.tsx
+++ b/ui/src/features/project/pipelines/freight/freight-artifact.tsx
@@ -17,7 +17,6 @@ import { shortVersion } from './short-version-utils';
 type FreightArtifactProps = {
   artifact: GitCommit | Chart | Image;
   expand?: boolean;
-  showArtifactIcons?: boolean;
 };
 
 export const FreightArtifact = (props: FreightArtifactProps) => {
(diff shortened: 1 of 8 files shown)


ui/src/features/project/pipelines/freight/freight-artifact.tsx
```diff
@@ -10,12 +10,14 @@ import {
...
 import { Chart, GitCommit, Image } from '@ui/gen/api/v1alpha1/generated_pb';
 
+import { ArtifactIcon } from './artifact-icon';
 import { humanComprehendableArtifact } from './artifact-parts-utils';
 import { shortVersion } from './short-version-utils';
 
 type FreightArtifactProps = {
   artifact: GitCommit | Chart | Image;
   expand?: boolean;
+  showArtifactIcons?: boolean;
```
rpelczar: If this is coming from context, why can't we use this directly without passing props?

Marvin9: Nothing specific just wanted to keep this component from side-effects just incase if we want to reuse anywhere else than pipeline

rpelczar:
@Marvin9  I don't like the placement where we can enable or disable it. Also, I'm not sure if we really need this to be configurable. 
We decided to get rid of icons because we wanted to save some space, and we thought it was redundant. If people are asking about bringing it back, maybe let's just add it. You can set a smaller size for these icons. 
Maybe we can also delete the external link icon and make all tags clickable?

WDYT?

jmalisz-akuity:
I'm also for just adding those icons back without the setting. I feel that most of the users would keep it enabled at all times, unless they don't know that it exists 😄

Issue #4988, opened 2025-09-03
Warehouse filter keep getting reset


alexmt:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

The Warehouses filter keep getting reset after I'm attempting to promote any freight to any stage. This make it very inconvenient to use UI with multi-freight projects. Bug seems to be a regression on 1.7.4

# Steps to Reproduce

* Navigate to a project with multiple warehouses
* Select a single warehouse in a warehouses filter
* Click "Promote" button of any stage and select any freight
* The  warehouses filter selection resets to empty

# Version

```
1.7.4
```

jessesuen:
I think this got fixed in https://github.com/akuity/kargo/pull/4991

PR #4991, opened 2025-09-04
fix(ui): filter issues

Marvin9:
https://github.com/akuity/kargo/issues/4988

There were 2 source of truths for filters (query params and local storage) that, if one of not properly updated, will cause issues when that source of truth is prioritised (it is prio in sub-pages of project because it is ugly to inject params in each and every subpage) so we rely on local storage in those cases.

This PR will make sure that local storage is also being updated as filters are changed, and being prioritised more than query params

https://github.com/user-attachments/assets/620d9ca2-bc64-424a-8e44-ec2c66f4a03c


Marvin9:
Good catch. Thanks @jmalisz-akuity

PR #4993, opened 2025-09-05
docs: Drop Duplicate `Freight` Output 

ianstanton:
This pull request makes a minor update to the documentation by removing a redundant sample output block from the 'Working with Freight' user guide.


krancour (approved):
Thank you!

Issue #3172, opened 2024-12-21
Scope GitHub App installation token to a set of repositories per project


Sobuno:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

When using a global GitHub App installation token, introduce some way to define a set of repositories accessible by individual Projects which the installation token will then be scoped to.

# Motivation
In a large organisation with many GitOps repositories which would be separated into many Kargo projects owned by different teams, we want to ensure no (accidental or malicious) cross-team (write!) access to other GitOps repositories.

Using Kargo today, that means using either a GitHub App per project or a PAT:

Per https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/registering-a-github-app#about-registering-github-apps, an organization can only create 100 GitHub Apps (although there's no limit to the amount that can be installed in the organization), meaning this will at some point not scale.

PAT: As PATs are user-specific, they are out of the question; users leave, have their permisisons revoked (e.g. due to moving teams) and all commits will be attributed to that user - Alternatively you can setup non-human accounts (i.e. service accounts) with these PATs, but that will cost a seat in the organization; at a price of USD 21 per seat per month in an enterprise, 100 service accounts (equivalent to the maximum GitHub Apps that can be created by an organization) would cost USD 2100 per month.

Having a global set of credentials in the form of one GitHub App that is appropriately scoped in a project to the repositories needed will allow scaling without needing to manage multiple set of credentials and without allowing cross-team access.

# Suggested Implementation
(Below is inspired by Pipelines As Code's approach: https://pipelinesascode.com/docs/guide/repositorycrd/)

[The endpoint for obtaining an installation token](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app) supports sending in a list of repositories or repository_ids when requesting the token - If this is done, the installation token returned will be scoped to those repositories, meaning this could be done. Permissions can also be scoped although not on a repository-by-repository basis but only on a token-by-token basis.

Introduce a new namespaced custom resource "GitRepository" or similar containing a very simple specification: `repoUrl`.

[...]

PR #4995, opened 2025-09-05
feat(controller): narrow scope of github app installation access tokens

nitishfy:
Closes https://github.com/akuity/kargo/issues/3172


internal/credentials/kubernetes/database.go
```diff
@@ -138,11 +138,25 @@ clientLoop:
...
+		dataCopy := make(map[string][]byte)
+		for key, val := range secret.Data {
+			dataCopy[key] = val
+		}
+		if secret.Annotations != nil {
+			if val, ok := secret.Annotations[kargoapi.AnnotationProjectReposKey]; ok && val != "" {
+				dataCopy[kargoapi.AnnotationProjectReposKey] = []byte(val)
+			}
+		}
+		data = dataCopy
```
krancour: These are details relevant to one, specific credentials provider and don't belong here. This component is meant to be quite dumb and just range over credential providers, trying each one _without knowing anything about how any of them work._

I understand _why_ you did this, though. There wasn't already any way to make data from the Secret's annotations available to the credential providers, but this isn't a good way to solve that.

What needs to happen is the `credentials.Provider` interface needs to be modified to create a way for this sort of data to be uniformly passed to any/all credential providers without the caller needing to know anything about how individual credential providers work.

For illustrative purposes, the easiest (but wrong) way to do that would be to change this:

```golang
type Provider interface {
	// ...
	GetCredentials(
		ctx context.Context,
		project string,
		credType Type,
		repoURL string,
		data map[string][]byte,
	) (*Credentials, error)
}
```

To this:

```golang
type Provider interface {
	// ...
	GetCredentials(
		ctx context.Context,
		project string,
		credType Type,
		repoURL string,
		secret *corev1.Secret,
	) (*Credentials, error)
}
```

This would work, but also couples all the credential providers to Kubernetes Secrets when they weren't before. tbh, modifying the `credentials.Provider` interface in a way that gives you what you need _without_ also adding this new coupling seems to me to actually be the hardest issue to solve for in this PR.

The best idea I can come up with so far is modifying the interface like this:

```golang
type Provider interface {
	// ...
	GetCredentials(
		ctx context.Context,
		project string,
		credType Type,
		repoURL string,
		data map[string][]byte,
    metadata map[string][]byte,
	) (*Credentials, error)
}

```

Before you act on that, I want to solicit opinions from @hiddeco and @thomastaylor312 in case one of them can think of something better.

krancour: Oh... I just thought of another option, but it's also not ideal.

We could _not_ modify the interface and have users put the new mappings directly in the Secret's data to begin with, instead of putting them in an annotation.

The only thing I don't love about it is the difficulty of viewing and modifying those mappings if they're included in the Secret's data.

internal/credentials/kubernetes/database.go
```diff
@@ -138,11 +138,25 @@ clientLoop:
...
 	}
 
+	logging.LoggerFromContext(ctx).Info("data value in secret", "data:", data)
```
krancour: I know other comments I made could possibly result in this line going away entirely, but my guidance on this is generally applicable and something to remember going forward...

1. Don't log things like this at info level. This information is of no interest unless you are actively developing or troubleshooting. Outside of that context, this is noise.
2. **This would create a vulnerability. Do not ever write secrets to logs.**

internal/credentials/kubernetes/database.go
```diff
@@ -138,11 +138,25 @@ clientLoop:
...
 	for _, p := range k.credentialProviders {
 		creds, err := p.GetCredentials(ctx, namespace, credType, repoURL, data)
+		logging.LoggerFromContext(ctx).Info("creds value", "creds", creds)
```
krancour: Same as previous comment: This is both noisy and unsafe.

internal/credentials/kubernetes/github/app.go
```diff
@@ -69,17 +74,48 @@ func (p *AppCredentialProvider) Supports(
...
 
+	// Extract the pproect-repos JSON from annotation
+	// in data or directly from data key
+	projectReposJSON, hasProjectRepos := data[kargoapi.AnnotationProjectReposKey]
+
+	var allowedRepos []string
+	unrestricted := true
+
+	if hasProjectRepos && len(projectReposJSON) > 0 {
+		logger.Debug("1")
```
krancour: I assume you didn't mean to commit this line.

internal/credentials/kubernetes/github/app.go
```diff
@@ -69,17 +74,48 @@ func (p *AppCredentialProvider) Supports(
...
 	if !p.Supports(credType, repoURL, data) {
 		return nil, nil
 	}
 
+	// Extract the pproect-repos JSON from annotation
+	// in data or directly from data key
+	projectReposJSON, hasProjectRepos := data[kargoapi.AnnotationProjectReposKey]
+
+	var allowedRepos []string
+	unrestricted := true
```
krancour: As a general rule, booleans that are "not something" are confusing. Every time someone reads something like `!unrestricted` it needs to be mentally converted to "restricted."

If you call the variable `restricted` and default it to false, the meaning remains the same, but the code is easier to understand.

```suggestion
	var restricted bool
```

internal/credentials/kubernetes/github/app.go
```diff
@@ -69,17 +74,48 @@ func (p *AppCredentialProvider) Supports(
...
+		if err := json.Unmarshal(projectReposJSON, &projectRepoMap); err != nil {
+			// If JSON is invalid, we mark as restricted (deny by default)
+			logger.Debug("error unmarshalling project-repos JSON")
+			unrestricted = false
+		} else {
+			// Look up repos for the specific project
+			repos, found := projectRepoMap[project]
+			if !found || len(repos) == 0 {
+				// project not allowed any repos, deny creds
+				return nil, fmt.Errorf("no repositories allowed for project %q", project)
```
krancour: This shouldn't be an error. If you return an error, you'll stop the caller from moving on to checking other providers -- one of which may work! You just want to return nil, indicating you found no creds.

internal/credentials/kubernetes/github/app.go
```diff
@@ -98,11 +134,34 @@ func (p *AppCredentialProvider) GetCredentials(
...
+			}
+		}
+
+		if !found {
+			// repo not allowed for this project
+			return nil, fmt.Errorf("repository %q is not allowed for project %q", repoName, project)
+		}
+	} else {
+		allowedRepos = []string{repoName}
+	}
```
krancour:
```suggestion
	if restricted && !slices.Contains(allowedRepos, repoName) {
		return nil, nil
	}
```

internal/credentials/kubernetes/github/app.go
```diff
@@ -98,11 +134,34 @@ func (p *AppCredentialProvider) GetCredentials(
...
+
 	return p.getUsernameAndPassword(
 		clientID,
 		installID,
 		string(data[privateKeyKey]),
 		baseURL,
+		allowedRepos,
```
krancour: I don't think there is any circumstance where the installation token you're requesting needs to request anything more (or less!) than the single repo for which we were search for creds.

```suggestion
		repoName,
```

internal/credentials/kubernetes/github/app.go
```diff
@@ -198,13 +268,14 @@ func tokenCacheKey(
...
+	allowedRepos []string,
 ) string {
 	return fmt.Sprintf(
 		"%x",
 		sha256.Sum256([]byte(
 			fmt.Sprintf(
-				"%s:%s:%d:%s",
-				baseURL, clientID, installationID, encodedPrivateKey,
+				"%s:%s:%d:%s:%s",
+				baseURL, clientID, installationID, encodedPrivateKey, strings.Join(allowedRepos, ","),
```
krancour: See https://github.com/akuity/kargo/pull/4995/files#r2328386620

If we scope all installation tokens to just the one repo we're trying to access, then the cache key needs to include that repoURL, which means, baseURL doesn't need to be included anymore.

```suggestion
				repoURL, clientID, installationID, encodedPrivateKey, strings.Join(allowedRepos, ","),
```

You'll have to make corresponding adjustments to other bits of code to make `repoURL` available here _in place of_ both `baseURL` and `allowedRepos`.

internal/credentials/kubernetes/github/app.go
```diff
@@ -244,3 +315,15 @@ func extractBaseURL(fullURL string) (string, error) {
 	}
 	return u.Scheme + "://" + u.Host, nil
 }
+
+// extractRepoName returns the repository name from a Git repository URL.
+// It trims the optional ".git" suffix and then extracts the last path segment.
+func extractRepoName(repoURL string) string {
+	trimmed := strings.TrimSuffix(repoURL, ".git")
```
krancour: There's not much point in doing this because throughout the entire process of searching for suitable credentials, the repoURL can be assumed to be normalized. That's done way up front before we start iterating over individual credential providers. Normalization will have already removed any trailing `.git`.

If you're thinking, "there's no harm in being certain of this," you're right, but _just_ removing the trailing `.git` isn't enough. There are other things to account for. (Consider for instance that https://github.com/akuity/kargo.git/ is a valid URL where just trimming `.git` doesn't work. For it to work, your _next_ line needs to move up above this one.) So now we're talking about replicating a larger portion of the normalization logic and it's not worth it. If you don't like assuming that the URL was, indeed, normalized in advance, re-normalize right here, but don't write your own "poor man's" normalization.

internal/credentials/kubernetes/github/app.go
krancour: And passing repoName means you _don't_ need to pass `baseName`.

internal/credentials/kubernetes/database.go
nitishfy: ah sorry, this was meant for debugging locally. I'll remove these logs.

internal/credentials/kubernetes/github/app.go
nitishfy: sorry, that was meant for debugging. Please take a look now.

internal/credentials/kubernetes/github/app.go
nitishfy: I think we shouldn't be returning here `nil, nil` and instead return an error. For eg, if you take a look at a step like `git-clone`, [it doesn't check for a case when creds are not found](https://github.com/akuity/kargo/blob/main/internal/promotion/runner/builtin/git_cloner.go#L103) from the db. It relies on the error to stop the cloning. If we simply return an error as nil here, it can clone a public repository even though a project doesn't allow access to that public repository. For eg, imagine `kargo-simple` as a public repository and you've configured your annotation to make sure `proj-a` CANNOT access `kargo-simple` repo. If we are returning nil here, the `git-clone` will successfully be able to clone the `kargo-simple` repo, abusing the annotation we configured.

internal/credentials/kubernetes/github/app.go
```diff
@@ -69,17 +75,51 @@ func (p *AppCredentialProvider) Supports(
...
+		allowedRepos []string
+		restricted   bool
+	)
+
+	if hasProjectRepos && len(projectReposJSON) > 0 {
+		var projectRepoMap map[string][]string
+		if err := json.Unmarshal(projectReposJSON, &projectRepoMap); err != nil {
+			// If JSON is invalid, we mark as restricted (deny by default)
+			logger.Debug("error unmarshalling project-repos JSON")
+			restricted = true
```
nitishfy: @krancour shouldn't we return an error here instead of denying by default?

nitishfy:
hmmm, trying to figure out why the codegen and tests are failing even though they pass on my local machine...

internal/credentials/kubernetes/database.go
thomastaylor312: I am coming at this without too much context, but couldn't we make this generic by introducing another interface? I don't know if the things we'll be fetching from annotations are all concrete. If they are, the interface could look something like this (which we could add more methods to later if we needed):

```go
type CredentialSource interface {
	Data() map[string][]byte
	ProjectRepos() []string
}
```

If they aren't concrete or need more flexibility, it could be this instead:

```go
type CredentialSource interface {
	Data() map[string][]byte
	// Name is not good, but is just to illustrate
	AdditionalContext() map[string]string
}
```

If we go with this latter one, then I would say that the kubernetes implementation would trim the `kargo.akuity.io` off of the annotation key for each value so that it isn't Kubernetes specific (i.e. you'd use `project-repos` as the key in the map no matter which implementation of the interface you're using).

Then all you'd have to have is something like this for the secret:

```go
type SecretCredentialSource struct {
	data map[string][]byte
	// Impl for massaged map of additional context or for ProjectRepose depending on choice above
}
```

Then the `Provider` Interface could be updated to:

```go
type Provider interface {
	// Supports returns true if the provider can provide credentials for the
	// given type, repository URL and data values. Otherwise, it should return
	// false.
	Supports(credType Type, repoURL string, data map[string][]byte) bool

	// GetCredentials returns the credentials for the given type, repository URL
	// and data values. If the provider cannot provide credentials for the given
	// type, repository URL and data, it should return nil.
	GetCredentials(
		ctx context.Context,
		project string,
		credType Type,
		repoURL string,
		source CredentialSource,
	) (*Credentials, error)
}
```

This keeps the interface non-Kubernetes specific _and_ as an added benefit, makes it so we could theoretically add other sources in the future

internal/credentials/kubernetes/github/app.go
krancour: > If we simply return an error as nil here, it can clone a public repository even though a project doesn't allow access to that public repository.

This feature is meant to control the scope of what a Project can do _with the credentials_. It's not meant to say what the Project can do with the public repo _without_ the use of those credentials.

And as I mentioned previously, returning nil here is "nothing found; keep looking at other providers." Returning an error stops that in its tracks. We don't want that.

krancour:
> hmmm, trying to figure out why the codegen and tests are failing even though they pass on my local machine...

If you're running into "it worked on my machine," run the `hack-` equivalents of the make targets. It will isolate the tasks to clean-room containers that mimic the containers used in CI as closely as possible. This is the exact reason that option exists.

So, ideally, you run those and encounter the same failures you are seeing in CI and then you can fix them more easily.

internal/credentials/kubernetes/github/app.go
krancour: Here, yes. (If this somehow came from one of my suggested code edits, I apologize.  Editing on GH is prone to mistakes.) In other spots, I've mentioned that not finding a credit isn't the same as an error. But this is a case where we can guess that there was _supposed_ to be a policy here telling us something important, but we can't guess what it was. Moving on to the next provider isn't the right thing to do. The right thing is to surface the error and let the user straighten it out.

internal/credentials/kubernetes/github/app.go
```diff
@@ -198,15 +259,16 @@ func tokenCacheKey(
...
 	return fmt.Sprintf(
 		"%x",
 		sha256.Sum256([]byte(
 			fmt.Sprintf(
-				"%s:%s:%d:%s",
-				baseURL, clientID, installationID, encodedPrivateKey,
-			),
-		)),
+				"%s:%s:%d:%s:%s",
+				baseURL, clientID, installationID, encodedPrivateKey, repoName),
```
krancour: The value of `baseURL` is a subset of `repoName`. You don't need it to keep this key unique now that you're including `repoName` in the key.

internal/credentials/kubernetes/github/app.go
nitishfy: So is the following behaviour expected?

The annotation scopes the `proj-a` to `repo-a` and `repo-b`. The promotion step in `proj-a` tries to clone the `repo-c` and the cloning is successfull even though annotation supported limiting the scope of `proj-a` to only `repo-a` and `repo-b`.

nitishfy: I get your point that we should try to look for other credential provider but when i tested it locally, it feels a bit off when i have only one cred provider such as git and even though the project limits the scope to repo-a and repo-b, it can still clone repo-c.

internal/credentials/kubernetes/database.go
krancour: @thomastaylor312 I'm not sure that solves the problem at hand, which isn't _really_ about having a uniform way to pass data to the providers. That's secondary.

The real problem is what data is passed to all the providers, or perhaps more precisely, where it comes from. As this PR currently stands, the caller is retrieving and passing on one extra, very specific bit of data used by only one provider.

We need to solve _that_ more generically. I think the best we can do for now (I don't think the feature this is meant to enable is worthy of us spending a lot of time on this), is go with my first suggestion.

Instead of retrieving _only_ the data map from the Secret and passing that `map[string][]byte` to each provider, we should _additionally_ retrieve the annotation map from the Secret and pass that `map[string]string` to every provider as "metadata." (I did mistakenly refer to the metadata as `map[string][]byte` in a previous comment, but it is actually `map[string]string`.)

What is different between this PR's current approach and this is that the current approach retrieves specific data from the annotations and the proposed approach retrieves the whole map and lets the providers worry about what specific data they need to extract from it.

This does require changing the provider interface to this:

```golang
type Provider interface {
	// ...
	GetCredentials(
		ctx context.Context,
		project string,
		credType Type,
		repoURL string,
		data map[string][]byte,
    metadata map[string][]string,
	) (*Credentials, error)
}
```

@nitishfy, this is the approach I want us to use.

api/v1alpha1/annotations.go
```diff
@@ -69,4 +69,26 @@ const (
...
+	//   }
+	//
+	// If the annotation is present on a Secret, Kargo will enforce that the project
+	// must be explicitly granted access to the target repository for credentials
+	// to be returned. If the project is missing or has no allowed repos, credential
+	// requests for that project and repo will be denied.
+	//
+	// If the annotation is missing or empty, no per-project repository restrictions
+	// will be applied (full access within matching Secret scope).
+	AnnotationProjectReposKey = "kargo.akuity.io/project-repos"
```
krancour: Since this has only to do with the scope of GitHub App installation tokens, I think we'd do well to use a more specific key here. I'm worried `kargo.akuity.io/project-repos` might incorrectly imply broader scope than what this is actually for. Most users won't have read this comment in the code to understand otherwise. Many won't even have read the docs.

`kargo.akuity.io/github-token-scopes` seems reasonable.

krancour: For the same reason, the name of the constant should be changed similarly.

internal/credentials/kubernetes/database.go
nitishfy: resolved.

api/v1alpha1/annotations.go
nitishfy: changed.

internal/credentials/kubernetes/database.go
nitishfy: one question - instead of passing metadata of type `map[string][]string`, why can't we simply pass the annotations field, something like:
```go
type Provider interface {
	// ...
	GetCredentials(
		ctx context.Context,
		project string,
		credType Type,
		repoURL string,
		data map[string][]byte,
      annotations map[string]string,
	) (*Credentials, error)
}
```

nitishfy:
> [return error when repo is not found](https://github.com/akuity/kargo/pull/4995/commits/3ef3409f9ab5ed59cd6dcd1dd62f698d70b46f47)

Just realised this is the wrong commit message, apologies!

internal/credentials/kubernetes/database.go
krancour: You can. With my apologies, `map[string][]string` was a typo. I edited `map[string][]byte` and missed the removal of the empty square brackets.

Thank you for checking.

nitishfy: okay thanks! I tried to make it work with `map[string][]string` but the code looks complicated reading it. I'll revert my changes and get back to using `map[string]string`.

nitishfy: Updated again.

docs/docs/50-user-guide/50-security/30-managing-credentials.md
```diff
@@ -347,6 +347,31 @@ create or install a GitHub App.
...
+   when accessing `repo-1` or `repo-2` repos. The project `kargo-simple` can use it
+   only for `repo-3` repo. Any other project or repo will result in no credentials
+   being returned.
+
+   If the annotation is missing or empty, no per-project restrictions are applied,
+   and the secret is valid for any repo that matches `repoURL`. If the JSON is invalid,
+   Kargo will log an error and fail credential resolution for this secret without moving
+   to next credential provider.
+   :::
+
```
krancour: @nitishfy, very nicely written. Something to clarify probably (if the code doesn't already match this, it will by the time we're done), is that if a token is returned at all it will _alwayss_ be scoped only to the repo for which we were looking for creds. i.e. If a Project is mapped to A, B, and C, and is trying to access A, the token it gets will be scoped to A only and _not_ to B and/or C. And this will be true even when there are no restrictions specified. Any one token can be used to access precisely one repo.

I think this also changes the guidance we give starting down around line 420.

I've not thought carefully about the precise wording for any of this. I leave that to you.

nitishfy: thank you! Please take a look now.

docs/docs/50-user-guide/50-security/30-managing-credentials.md
```diff
@@ -366,6 +366,13 @@ create or install a GitHub App.
    only for `repo-3` repo. Any other project or repo will result in no credentials
    being returned.
 
+   Even if a project is mapped to multiple repositories, the token returned by Kargo
+   is always scoped to exactly one repository, the one you requested credentials for.
+   For example, if `kargo-demo` is mapped to `repo-1` and `repo-2` but is requesting
+   access to `repo-1`, Kargo will return a token that works only for `repo-1`.
+   It will not work for `repo-2`. This behavior applies even when the
+   `kargo.akuity.io/github-token-scope` annotation is missing or empty.
+
```
krancour: 👍

docs/docs/50-user-guide/50-security/30-managing-credentials.md
```diff
@@ -390,33 +390,52 @@ App is that the App's permissions are not tied to a specific GitHub user.
...
+For convenience’s sake, it may be tempting to register a single GitHub App,
+select a broad set of repositories when installing that App, then create a
+single set of [global credentials](#global-credentials), _however_, this will
+have the undesirable effect of granting _all_ Kargo projects access to _all_ of
+the selected repositories.
+
+Alternatively, you might consider registering a _separate_ GitHub App for each
+Kargo project, selecting a narrower set of repositories when installing each
+App, then creating corresponding Secrets in individual projects namespaces.
```
krancour:
```suggestion
App, then creating corresponding Secrets in individual project namespaces.
```

docs/docs/50-user-guide/50-security/30-managing-credentials.md
```diff
@@ -390,33 +390,52 @@ App is that the App's permissions are not tied to a specific GitHub user.
...
+For convenience’s sake, it may be tempting to register a single GitHub App,
+select a broad set of repositories when installing that App, then create a
+single set of [global credentials](#global-credentials), _however_, this will
+have the undesirable effect of granting _all_ Kargo projects access to _all_ of
+the selected repositories.
+
+Alternatively, you might consider registering a _separate_ GitHub App for each
+Kargo project, selecting a narrower set of repositories when installing each
+App, then creating corresponding Secrets in individual project namespaces.
+While this better adheres to the principle of least permissions, it can be
```
krancour:
```suggestion
While this better adheres to the principle of least privilege, it can be
```

docs/docs/50-user-guide/50-security/30-managing-credentials.md
```diff
@@ -390,33 +390,52 @@ App is that the App's permissions are not tied to a specific GitHub user.
...
+App, then creating corresponding Secrets in individual project namespaces.
+While this better adheres to the principle of least privilege, it can be
+onerous to manage. Worse, because GitHub organizations are limited to
+registering 100 GitHub Apps each, the approach does not scale beyond 100
+projects.
+
+Beginning with Kargo v1.8.0, a third, experimental (stability not guaranteed)
+approach builds upon the first, by adding an optional annotation to the
+[global credentials](#global-credentials) `Secret` containing a map that
+constrains the scopes (repositories) available to each Project.
```
krancour:
```suggestion
constrains the scopes (repositories) available to each project.
```

krancour (approved):

PR #4999, opened 2025-09-05
feat(logging)!: Remove logrus and expose logging in `pkg`

### LINKED ISSUE #4829: feat request: Log Format JSON
### Discussed in https://github.com/akuity/kargo/discussions/4782

<sup>Originally posted by **nicolasbriere1** August  7, 2025</sup>
Hi, 

I didn't find any documentation about this, but is it possible to switch the log format of Kargo to JSON ? 

Our observability stack handles better this format :) 

Thanks</div>

thomastaylor312:
This takes some learnings we've had from other projects and folds them into the logging implementation, as well as making it accessible from `pkg`. I marked as breaking because we are swapping out the logger, so logging format might look a little different. Mostly doing this as a heads up more than out of any particular worry

Closes #4829


pkg/logging/logger.go
```diff
@@ -0,0 +1,237 @@
...
+
+// Error logs a message at the error level.
+func (l *Logger) Error(err error, msg string, keysAndValues ...any) {
+	l.callStackHelper()
+	l.logger.Error(err, msg, keysAndValues...)
+}
+
+// Warn logs a message at the warn level.
+func (l *Logger) Warn(msg string, keysAndValues ...any) {
+	// NOTE(thomastaylor312): Yep, logr be dumb. There is no WARN, so it just shows up as Info. If
```
krancour: If there were one log level I wouldn't care about losing entirely, it would be warn. In general, people don't pay adequate heed to warnings. If it's important enough to someone _should_ pay attention, it maybe ought to be an error. If it's innocent enough to safely ignore, then it can be info.

I would be equally fine with keeping it or losing it. Dealer's choice.

Either way, this whole PR LGTM.

I will ping you offline about one possible, optional feature that could be added with about three lines of code.

krancour:
@thomastaylor312 I reviewed your latest improvements and test drove them as well. Prior LGTM stands.

While test-driving, I realized we needed log format to be tunable via the chart. I quickly added that and amended the PR... so you may want to review my contributions to this before merging.

Issue #4978, opened 2025-09-03
OIDC Claims Mapping do not support special characters in key


jessesuen:
# Proposed Feature

IdPs  might return a claims with special characters in the key. For example, AWS Cognito returns `:` in their `cognito:groups` claim:

```json
{
  "cognito:groups": [
    "devops"
  ],
  "email_verified": true,
  "iss": "https://cognito-idp.us-west-2.amazonaws.com/us-west-2_XXXXXXXXXX",
  "email": "user@example.com",
...
}
```

Unfortunately, we can't formulate the proper annotation to map this claim to a ServiceAccount :

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    rbac.kargo.akuity.io/claim.cognito:groups: devops  # disallowed by k8s
```

Because it will fail with the error:
`Invalid value: \"rbac.kargo.akuity.io/claim.cognito:groups\": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')`

I think we will need to introduce a new annotation key, one accepted by Kubernetes, where the annotation value would be the YAML configuration that could accept special characters, and multiple mappings.

For example:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    # old way
    rbac.kargo.akuity.io/claim.email: user@example.com

    # new way
    rbac.kargo.akuity.io/claims: |
      'cognito:groups': devops
      email: user@example.com
```

# Motivation

Currently, services like AWS Cognito cannot fully be supported.

krancour:
I'm reclassifying this as a bug, because I believe we can accommodate this with a minor "fix" that doesn't rise to the level of a new feature. If I turn out to be wrong, I'll reclassify again as and enhancement.

PR #5002, opened 2025-09-06
feat: support special characters for OIDC mappings 

fuskovic:
Closes: https://github.com/akuity/kargo/issues/4978


krancour:
This is pretty different from the direction I'd initially been thinking, but I think you're headed in a _better_ direction.

I _think_ I'd originally gotten the idea from a conversation with @jessesuen that we might get away with something like a simple character substitution. It's best to illustrate with an example:

To map users with claim `foo:bar: bat` to a ServiceAccount, one would use the annotation `rbac.kargo.akuity.io/claim.foo_bar: bat`. (We'd have documented that.) Such a ServiceAccount would end up indexed as `claims: ["foo_bar/bat"]`.

When authenticating a user whose ID token contains the `foo:bar: bat` claim, we would programmatically apply the substitution. We'd have ended up looking for ServiceAccounts where the indexed `claims` field contains `foo_bar/bat` and things would work out great.

I liked the simplicity of this idea, but after researching it more, I think it's not safe. JWT claim names can legally contain nearly any Unicode character. If all characters not permitted in a Kubernetes annotation key were replaced with `_`, then `foo:bar`, `foo@bar`, `foo#bar`, etc. _all_ get encoded as `foo_bar`. When authenticating a user with the claim `foo@bar: bat`, we'd once again be searching for ServiceAccounts where the indexed `claims` field contains `foo_bar/bat`... and we'd get a hit, despite the user actually having the wrong claim.

I think the reality of an IdP using _both_ claims `foo:bar` and `foo@bat` (for instance) is _extremely_ low. But where access controls are concerned, I don't want to rely on "that will probably never happen."

I thought about how we could possibly use more discreate substitutions to avoid that problem and with the characters permitted in a Kubernetes annotation key being so limited and with the length of those keys limited to 63 characters with the prefix `rbac.kargo.akuity.io/claim.` taking up 27 of those, I couldn't see any such scheme possibly succeeding.

I am writing all of this mainly to document the fact of due diligence having been performed on what might seem to some to be a more obvious approach. So again... I like the direction this PR is headed in. I'll suggest a few tweaks in-line, but overall, I am on board!

charts/kargo/templates/users/service-accounts.yaml
```diff
@@ -11,6 +11,10 @@ metadata:
     {{- range $claim, $values := .Values.api.oidc.admins.claims }}
     rbac.kargo.akuity.io/claim.{{ $claim }}: {{ quote (join "," $values) }}
     {{- end }}
+    rbac.kargo.akuity.io/claims: |
+    {{- range $claim, $values := .Values.api.oidc.admins.claimsv2 }}
+      {{ quote ($claim) }}: {{ quote (join "," $values) }}
+    {{- end }}
```
krancour: It would be safe to handle all the claims the new way. That will not break anything.

charts/kargo/values.yaml
```diff
@@ -171,6 +171,9 @@ api:
         # - bob@examples.com
         # groups:
         # - kargo-admin
+      ## @param api.oidc.admins.claimsv2 Subjects having any of these claims will automatically be Kargo admins.
+      # NOTE: This is a map of string to list of strings and should be used when keys contain special characters such as colons (:).
+      claimsv2: {}
```
krancour: See https://github.com/akuity/kargo/pull/5002/files#r2328346418

We could safely ditch this new field and just use the new annotation approach with the keys/vals from the existing `api.oidc.admins.claims` setting.

This might have been what you were talking about here?

> I may refactor this and bake the functionality right into claims

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -101,6 +101,26 @@ A user may be mapped to multiple `ServiceAccount` resources. A user's effective
...
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: admin
+  namespace: kargo-demo
+  annotations:
+    rbac.kargo.akuity.io/claims: |
+      'cognito:groups': devops
+      email: user@example.com
```
krancour: Because it's _extremely_ common for annotation values to contain JSON, I think we'd be better off using JSON for these mappings.

charts/kargo/templates/users/service-accounts.yaml
krancour: Clarifying: We'll still _allow_ the old way, but internally, our chart could get away with only using the new way on the SAs it manages.

fuskovic: When you refer to the _old way_ you mean: `rbac.kargo.akuity.io/claim.{{ $claim }}` ?

If so, you mentioned [here](https://github.com/akuity/kargo/pull/5002#discussion_r2328351497) we want to accept JSON input.

It's not obvious to me how we expect to do both. Iterating over the json fields with special characters for keys will still yield `rbac.kargo.akuity.io/claim.cognito:groups: devops` (which will be invalid)

fuskovic:
@krancour nixed `claimsv2` and started nixing the old `rbac.kargo.akuity.io/claim.{{ $claim }}` pattern but not completely yet.

I'd like some feed back on recent changes.

charts/kargo/templates/users/service-accounts.yaml
fuskovic: resolving per offline chat

api/rbac/v1alpha1/annotations.go
```diff
@@ -15,6 +15,12 @@ const (
 	// may be either a scalar string value or a comma-separated list.
 	AnnotationKeyOIDCClaimNamePrefix = "rbac.kargo.akuity.io/claim."
 
+	// AnnotationKeyOIDCClaims is an annotation key that can be set on a
+	// ServiceAccount to associate it with any user authenticated via OIDC and
+	// having any of the claims indicated by the value of the annotation. The
+	// value is expected to be a map[string]any.
```
krancour: For clarity:

```suggestion
	// having any of the claims indicated by the value of the annotation. The
	// value is expected to be a string representation of a JSON object containing
	// claim names as keys mapped to claim values represented as strings or lists
	// of strings.
	//
	// {"email": "kilgore@kilgore.trout", "groups": ["devops", "maintainers"]}
```

charts/kargo/templates/users/service-accounts.yaml
```diff
@@ -50,8 +44,6 @@ metadata:
     {{- include "kargo.labels" . | nindent 4 }}
   {{- if .Values.api.oidc.viewers.claims }}
   annotations:
-    {{- range $claim, $values := .Values.api.oidc.viewers.claims }}
-    rbac.kargo.akuity.io/claim.{{ $claim }}: {{ quote (join "," $values) }}
-    {{- end }}
+    rbac.kargo.akuity.io/claims: {{ .Values.api.oidc.viewers.claims | toJson | quote}}
   {{- end }}
-{{- end }}
+{{- end }}
```
krancour: It's standard practice (rooted in POSIX standards) to end files with a line break, so I'd rather keep it.

```suggestion
{{- end }}

```

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -85,21 +83,46 @@ In the following example, the `ServiceAccount` resource is mapped to all of:
...
-    rbac.kargo.akuity.io/claim.groups: devops,kargo-admin
+    rbac.kargo.akuity.io/claims:
+      sub: 
+        - alice
+        - bob
+      email: carl@example.com
+      groups: 
+        - devops
+        - kargo-admin
+      "special:key": value
```
krancour, line 103: This isn't a valid annotation. The value should be a _string representation_ of a JSON object.

krancour: This isn't a valid annotation. The value should be a string representation of a JSON object.

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -85,21 +83,46 @@ In the following example, the `ServiceAccount` resource is mapped to all of:
...
+metadata:
+  name: admin
+  namespace: kargo-demo
+  annotations:
+    rbac.kargo.akuity.io/claims:
+      {
+        "sub": ["alice", "bob" ],
+        "email": carl@example.com,
+        "special:key": value
+      }
```
krancour: Same as [here](https://github.com/akuity/kargo/pull/5002/files#r2335258645).

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -157,7 +181,8 @@ metadata:
   name: kargo-viewer
   namespace: kargo
   annotations:
-    rbac.kargo.akuity.io/claim.groups: developers
+    rbac.kargo.akuity.io/claims:
+      groups: developers
```
krancour: Same as [here](https://github.com/akuity/kargo/pull/5002/files#r2335258645).

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -74,9 +74,7 @@ project namespaces only (i.e. only those labeled with
 `kargo.akuity.io/project: "true"`). _This section focuses on the exceptions to
 that rule._
 
-ServiceAccount resources may be mapped to users through the use of annotations
-whose key begins with `rbac.kargo.akuity.io/claim.`. The value of the annotation
-may be a single value, or a comma-delimited list of values.
+ServiceAccount resources may be mapped to users through the use of annotation claims defined under the `rbac.kargo.akuity.io/claims` key. The value of the annotation can be a standard YAML map or a JSON object.
```
krancour: "annotation claims" doesn't sound quite right. Also annotation values can only be strings.

```suggestion
ServiceAccount resources may be mapped to users through use of the
`rbac.kargo.akuity.io/claim` annotation. Its value must be a string
representation of a JSON object containing claim names as keys mapped to claim
values represented as strings or lists.
```

docs/docs/50-user-guide/50-security/20-access-controls/index.md
krancour, line 1: This doc needs a lot of the same changes suggested on the corresponding operator doc.

docs/docs/50-user-guide/60-reference-docs/70-annotations.md
```diff
@@ -23,6 +23,7 @@ programmatically with Kargo.
 | `kargo.akuity.io/refresh` | `Warehouse`, `Stage`, `Promotion` | A string that is unique or at least unlikely to repeat, such as a UUID or a timestamp of "now" | Triggers reconciliation of the resource when its value changes. |
 | `kargo.akuity.io/reverify` | `Stage` | Either a plain string (verification ID from `.status.verifications[*].id` of the `Stage`) or a JSON object with `id` (required), `actor`, and `controlPlane` fields. If a JSON object is provided, it is parsed as a `VerificationRequest`. | Triggers re-verification of a previously completed verification for the current `Freight`. |
 | `rbac.kargo.akuity.io/claim.<name>` | `ServiceAccount` | Any valid OIDC claim value (e.g., `sub`, `email`, or `groups`) | Maps an OIDC claim to a `ServiceAccount`, enabling user-to-ServiceAccount mappings. For more details, refer to the access control sections of the [Operator Guide](../../40-operator-guide/40-security/30-access-controls.md) and [User Guide](../50-security/20-access-controls/index.md). |
```
krancour, line 25: I want to stop short of deprecating this, but I think it's worth mentioning that the new `rbac.kargo.akuity.io/claims` annotation is the recommended way of doing this, although I don't think you need to get into why that is.

hack/tilt/values.dev.yaml
```diff
@@ -26,11 +26,22 @@ api:
...
     admins:
+      # JSON format with arrays
       claims:
-        email:
-        # This email claim is hard-coded in the Dex mockCallback connector
-        - kilgore@kilgore.trout
+        {
+          "email": ["kilgore@kilgore.trout", "user@inbox.com"],
+          "cognito:groups": ["devops"]
+        }
```
krancour: I know these are changes you are just using for testing and intend to revert, so this is only an FYI as a point of reference. YAML is a superset of JSON that enables additional (simpler) syntax, so all valid JSON is _also_ valid YAML. I mention this because I think you may have used this syntax thinking it tests a different path through the code than that taken by the data starting on line +40, but it doesn't. i.e. Feel free to stick exclusively to the nicer syntax. It won't make this any less thorough.

internal/indexer/indexer.go
```diff
@@ -476,17 +476,38 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
...
-			claimValues := strings.Split(rawClaimValue, ",")
-			for _, e := range claimValues {
+			for e := range strings.SplitSeq(rawClaimValue, ",") {
 				if claimValue := strings.TrimSpace(e); claimValue != "" {
 					refinedClaimValues = append(refinedClaimValues, FormatClaim(rawClaimName, claimValue))
 				}
 			}
 		}
+		if strings.HasPrefix(annotationKey, rbacapi.AnnotationKeyOIDCClaims) {
+			claimsMap := make(map[string]any)
```
krancour: DM'ed you about this. You _can_ change this to `map[string][]string` as you suggested earlier.

internal/indexer/indexer.go
```diff
@@ -476,17 +476,38 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
...
 			}
-			claimValues := strings.Split(rawClaimValue, ",")
-			for _, e := range claimValues {
+			for e := range strings.SplitSeq(rawClaimValue, ",") {
 				if claimValue := strings.TrimSpace(e); claimValue != "" {
 					refinedClaimValues = append(refinedClaimValues, FormatClaim(rawClaimName, claimValue))
 				}
 			}
 		}
+		if strings.HasPrefix(annotationKey, rbacapi.AnnotationKeyOIDCClaims) {
```
krancour:
```suggestion
		if annotationKey == rbacapi.AnnotationKeyOIDCClaims {
```

internal/indexer/indexer.go
```diff
@@ -476,17 +476,38 @@ func ServiceAccountsByOIDCClaims(obj client.Object) []string {
...
+					}
+				}
+			}
+		}
 	}
+
 	if len(refinedClaimValues) == 0 {
 		return nil
 	}
+	slices.Sort(refinedClaimValues)
```
krancour: I'm not sure we need to do this.

krancour:
This is a spot where we programmatically annotate a SA:

https://github.com/akuity/kargo/blob/9d5ebfb7cd4f0b369e3f610af4509abb20153576/internal/controller/management/projects/projects.go#L818

This should switch the the new method of mapping users to SAs.

It is fine that there will be existing SAs that are mapped the old way. We just need to use the new way going forward.

__Edit:__

Another spot:

https://github.com/akuity/kargo/blob/9d5ebfb7cd4f0b369e3f610af4509abb20153576/internal/server/rbac/roles.go#L745-L794

Those functions will have to be able to read the old and new styles of annotations. It can _write_ just the new style.

krancour:
I'm almost positive this is the case, but also need to confirm with @Marvin9 that the parts of the UI for managing "Kargo roles" don't ever mess with ServiceAccounts directly and instead exclusively use the role management endpoints exposed by the API server.

internal/indexer/indexer.go
fuskovic: It's mostly for test assertions. Since we're unmarshalling into a map and maps are unordered, removing this causes the tests to fail intermittently.

api/rbac/v1alpha1/annotations.go
```diff
@@ -15,6 +15,16 @@ const (
...
 	AnnotationKeyOIDCClaimNamePrefix = "rbac.kargo.akuity.io/claim."
 
+	// AnnotationKeyOIDCClaims is an annotation key that can be set on a
+	// ServiceAccount to associate it with any user authenticated via OIDC and
+	// having any of the claims indicated by the value of the annotation. The
+	// value is expected to be a string representation of a JSON object containing
+	// claim names as keys mapped to claim values represented as strings or lists
+	// of strings.
+	//
+	// {"email": "kilgore@kilgore.trout", "groups": ["devops", "maintainers"]}
```
krancour: Relates to the what was discussed offline about how scalar strings don't need to be supported after all:

```suggestion
	// claim names as keys mapped to claim values represented as lists of strings.
	//
	// {"email": ["kilgore@kilgore.trout"], "groups": ["devops", "maintainers"]}
```

internal/indexer/indexer.go
krancour: I'll look at the tests... I hadn't yet. But in the worst case, you can make the tests sort before you make your assertions so that we're not sorting unnecessarily in the running system.

fuskovic:
Note to self to also update `listServiceAccounts` in the auth interceptor

krancour:
> Note to self to also update listServiceAccounts in the auth interceptor

Are you sure about that? iirc, that just uses the index and you've already straightened out the indexer.

fuskovic:
I refactored `replaceXX`, `amendXX`, and `dropXX` OIDC claim helpers such that we can read the old way and new but we always write the changes the new way

api/rbac/v1alpha1/annotations.go
```diff
@@ -15,6 +19,15 @@ const (
 	// may be either a scalar string value or a comma-separated list.
 	AnnotationKeyOIDCClaimNamePrefix = "rbac.kargo.akuity.io/claim."
 
+	// AnnotationKeyOIDCClaims is an annotation key that can be set on a
+	// ServiceAccount to associate it with any user authenticated via OIDC and
+	// having any of the claims indicated by the value of the annotation. The
+	// value is expected to be a string representation of a JSON object containing
+	// claim names as keys mapped to claim values represented as lists of strings.
+	//
+	// {"email": ["kilgore@kilgore.trout"], "groups": ["devops", "maintainers"]}
```
krancour:
```suggestion
	// For example:
	//
	//   `{"email": ["kilgore@kilgore.trout"], "groups": ["devops", "maintainers"]}`
```

internal/server/rbac/roles.go
```diff
@@ -743,54 +744,75 @@ func claimListToMap(claims []rbacapi.Claim) map[string][]string {
...
-			if _, exists := claims[name]; !exists {
-				delete(sa.Annotations, k)
-			}
+		if strings.HasPrefix(k, rbacapi.AnnotationKeyOIDCClaimNamePrefix) {
+			delete(sa.Annotations, k)
 		}
 	}
+	// We need to also remove any claims set the new way.
+	// e.g. using rbac.kargo.akuity.io/claims annotation.
+	delete(sa.Annotations, rbacapi.AnnotationKeyOIDCClaims)
```
krancour: I wouldn't bother with the delete unless I had to because the replacement claims are the empty set. And in that case, I would also return early:

```suggestion
	if len(claims) == 0 {
		delete(sa.Annotations, rbacapi.AnnotationKeyOIDCClaims)
		return
	}
```

internal/server/rbac/roles.go
```diff
@@ -743,54 +744,75 @@ func claimListToMap(claims []rbacapi.Claim) map[string][]string {
...
 	if sa.Annotations == nil {
 		sa.Annotations = map[string]string{}
 	}
-	// Add or update the annotations for the new claims
-	for name, values := range claims {
-		sa.Annotations[rbacapi.AnnotationKeyOIDCClaim(name)] = strings.Join(values, ",")
-	}
+	b, _ := json.Marshal(claims)
```
krancour: We shouldn't swallow this error. Just fyi, this method didn't previously return an error only because there was nothing within it that could result in one. (i.e. The reason wasn't because the caller of this method wasn't interested.) Now that there is, we should wrap and return it per usual.

internal/server/rbac/roles.go
```diff
@@ -743,54 +744,75 @@ func claimListToMap(claims []rbacapi.Claim) map[string][]string {
...
+			delete(sa.Annotations, rbacapi.AnnotationKeyOIDCClaim(name))
 		}
 		slices.Sort(values)
-		values = slices.Compact(values)
-		if sa.Annotations == nil {
-			sa.Annotations = map[string]string{}
-		}
-		sa.Annotations[rbacapi.AnnotationKeyOIDCClaim(name)] = strings.Join(values, ",")
+		newClaims[name] = slices.Compact(append(newClaims[name], values...))
+	}
```
krancour: It seems the result here would leave the old annotations for existing mappings alone if they weren't being amended with new values. I think it would be cleaner if we loaded all mappings done the old way into the `newClaims` map and we can delete all those older annotations. After that we can range over additions and insert them where they belong. In the end, we'll be left with the new style only. It seems cleaner, as for a human looking at the annotations, it will be easier to understand than mentally building the claim mappings from the union of what's in the one new style annotation and any remaining old ones.

internal/server/rbac/roles.go
krancour: Same thing in the amend and drop methods.

internal/server/rbac/roles.go
```diff
@@ -743,54 +744,75 @@ func claimListToMap(claims []rbacapi.Claim) map[string][]string {
...
+	if sa.Annotations == nil {
+		sa.Annotations = map[string]string{}
 	}
+	// set everything under the "rbac.kargo.akuity.io/claims" annotation
+	b, _ := json.Marshal(newClaims)
+	sa.Annotations[rbacapi.AnnotationKeyOIDCClaims] = string(b)
 }
 
 func dropFromClaimAnnotations(sa *corev1.ServiceAccount, claims map[string][]string) {
```
krancour: Similar to the amend function, I think it would be cleaner if we didn't leave any of the old annotations behind when we're done. And in this case, even moreseo than with amend, the logic will be much easier to follow if we begin with a comprehensive map and _then_ remove what we need to.

internal/server/rbac/roles.go
```diff
@@ -743,54 +744,75 @@ func claimListToMap(claims []rbacapi.Claim) map[string][]string {
...
+	if _, ok := sa.Annotations[rbacapi.AnnotationKeyOIDCClaims]; ok {
+		_ = json.Unmarshal([]byte(sa.Annotations[rbacapi.AnnotationKeyOIDCClaims]), &jsonStyleClaims)
+	}
 	for name, values := range claims {
-		slices.Sort(values)
-		values = slices.Compact(values)
-		values = removeFromStringSlice(strings.Split(sa.Annotations[rbacapi.AnnotationKeyOIDCClaim(name)], ","), values)
-		if len(values) == 0 {
+		if prefixStyleValues, ok := sa.Annotations[rbacapi.AnnotationKeyOIDCClaim(name)]; ok {
+			values = removeFromStringSlice(strings.Split(prefixStyleValues, ","), values)
```
krancour: `slices.DeleteFunc()` seems to have existed when I wrote this, but it was new enough that I'd not "gotten the memo."

I would use that instead and delete that function.

api/rbac/v1alpha1/annotations.go
```diff
@@ -30,3 +43,28 @@ func OIDCClaimNameFromAnnotationKey(key string) (string, bool) {
...
+	var result []string
+	for name, values := range claims {
+		for _, v := range values {
+			result = append(result,
+				FormatClaim(name, strings.TrimSpace(v)),
+			)
+		}
+	}
+	return result, nil
+}
```
krancour: There are enough other places in the PR where we need to read old and new style annotations and combine their mappings into one comprehensive collection that I think we might do ourselves a big favor by refactoring this to take a `map[string]string` as input and return a `map[string][]string`.

I do know there's a small side effect to that. Other places where you loop through the old style annotations, you delete them as you go, and I recognize that if we refactor this as I've suggested that we will need to take an extra pass through the annotations after to delete old ones. Especially since these collections tend to be very small, the extra pass is a price I am willing to pay in exchange for code that's easier to understand. You could make a helper function for the cleanup as well. It'd be used in at least two places.

Name can change to `OIDCClaimsFromAnnotationValues()` (plural).

internal/server/rbac/roles.go
krancour: Ties into [this](https://github.com/akuity/kargo/pull/5002/files#r2345221047).

api/rbac/v1alpha1/annotations.go
```diff
@@ -30,3 +45,28 @@ func OIDCClaimNameFromAnnotationKey(key string) (string, bool) {
...
+func OIDCClaimsFromAnnotationValue(value string) ([]string, error) {
+	claims := make(map[string][]string)
+	if err := json.Unmarshal([]byte(value), &claims); err != nil {
+		return nil, fmt.Errorf("unmarshaling OIDC claims from annotation value: %w", err)
+	}
+	var result []string
+	for name, values := range claims {
+		for _, v := range values {
+			result = append(result,
+				FormatClaim(name, strings.TrimSpace(v)),
```
krancour: This seems as if it's not right. This is going to return results in a way specially formatted for use with the index we've built of SAs by claim. Other than building or using that index, I don't think there is any other context where we want claims to look like this, and I believe I see this reflected in some of the tests where you are expecting values like `email/kilgore@kilgore.trout` where we ought not to be.

api/rbac/v1alpha1/annotations.go
```diff
@@ -46,23 +49,50 @@ func OIDCClaimNameFromAnnotationKey(key string) (string, bool) {
...
 
-// OIDCClaimsFromAnnotationValue parses the value of an
-// rbac.kargo.akuity.io/claims annotation and returns the corresponding list of
-// formatted and sorted claims.
-func OIDCClaimsFromAnnotationValue(value string) ([]string, error) {
+// OIDCClaimsFromAnnotationValue parses the values of both the
+// rbac.kargo.akuity.io/claims and rbac.kargo.akuity.io/claim.<name> annotations
+// and consolidates them into a single map where the value of each key is sorted and deduped.
+func OIDCClaimsFromAnnotationValues(annotations map[string]string) (map[string][]string, error) {
 	claims := make(map[string][]string)
```
krancour, line 60: We can avoid an unnecessary allocation here.

```suggestion
	var claims map[string][]string
```

api/rbac/v1alpha1/annotations.go
```diff
@@ -46,23 +49,50 @@ func OIDCClaimNameFromAnnotationKey(key string) (string, bool) {
...
+		slices.Sort(v)
+		claims[k] = slices.Compact(v)
+	}
+	return claims, nil
+}
+
+// SetOIDCClaimsAnnotations sets the rbac.kargo.akuity.io/claims annotation on
+// the given ServiceAccount to the JSON representation of the given claims map.
+// It also removes any existing rbac.kargo.akuity.io/claim.<name> annotations.
+func SetOIDCClaimsAnnotations(sa *corev1.ServiceAccount, claims map[string][]string) error {
```
krancour: Good call making a reusable function for this also.

Only suggestion is this is more accurate if it's `SetOIDCClaimsAnnotation()` (singular) and there's only one claims annotation in the new claims --> SA mapping scheme.

```suggestion
// SetOIDCClaimsAnnotation sets the rbac.kargo.akuity.io/claims annotation on
// the given ServiceAccount to the JSON representation of the given claims map.
// It also removes any existing rbac.kargo.akuity.io/claim.<name> annotations.
func SetOIDCClaimsAnnotation(sa *corev1.ServiceAccount, claims map[string][]string) error {
```

I am a bit iffy on this function also cleaning up the old annotations, as it feels like a distinctly different responsibility, but I cannot see any circumstance where we want to write out mappings using the new annotation _without_ also cleaning up the older annotations, so this definitely makes sense. Let's leave it be for now, but it's something we might come back to.

api/rbac/v1alpha1/annotations.go
fuskovic: You have to make the alloc here because making entries into `nil` map will cause a panic. 

You can repro this by running `TestOIDCClaimsFromAnnotationValues/old_style`

krancour: Oh... I zeroed in on the `json.Unmarshal()` call below, for which you don't need to pre-allocate the map, _but I forgot that it's conditional_... if the new annotation isn't used, you reach the code starting on line +64 with a var that's still nil.

🤦‍♂️ Nothing to see here. Carry on.

api/rbac/v1alpha1/annotations.go
```diff
@@ -30,3 +48,55 @@ func OIDCClaimNameFromAnnotationKey(key string) (string, bool) {
...
+		}
+	}
+	return nil
+}
+
+// FormatClaim formats a claims name and values to be used by the
+// IndexServiceAccountsByOIDCClaims index.
+func FormatClaim(claimName, claimValue string) string {
+	return claimName + "/" + claimValue
+}
```
krancour: I understand why you moved this here. It probably seemed the right thing to do when you were using this function more broadly, but since that was a mistake that's since been corrected, this is back to used only when creating or using the SA by claims index. I would probably put this back in the `indexer` package where it originally came from.

internal/indexer/indexer_test.go
```diff
@@ -824,14 +825,55 @@ func TestServiceAccountsByOIDCClaims(t *testing.T) {
 			},
 			expected: []string{"email/fake-email", "email/fake-email-2"},
 		},
+		{
```
krancour, line 828: I would consider renaming the existing tests above here to better reflect what they're now covering. "ServiceAccount has no OIDC email" is probably more properly "has no claim mappings" or something along those lines. "ServiceAccount has OIDC email" could probably stand to be more clear that it's testing the old style of mappings.

On that note, we could probably add a test case below that verifies the index is built correctly for an SA that mixes old and new style mappings.

api/rbac/v1alpha1/annotations.go
```diff
@@ -15,6 +22,17 @@ const (
 	// may be either a scalar string value or a comma-separated list.
 	AnnotationKeyOIDCClaimNamePrefix = "rbac.kargo.akuity.io/claim."
```
krancour, line 25: I want to stop short of deprecating it, because I just don't see a real need to do so, but could we add a comment here indicating that a string representation of a JSON object in the AnnotationKeyOIDCClaims is the preferred method of doing this going forward?

api/rbac/v1alpha1/annotations.go
```diff
@@ -30,3 +48,49 @@ func OIDCClaimNameFromAnnotationKey(key string) (string, bool) {
 	}
 	return strings.TrimPrefix(key, AnnotationKeyOIDCClaimNamePrefix), true
 }
+
+// OIDCClaimsFromAnnotationValue parses the values of both the
+// rbac.kargo.akuity.io/claims and rbac.kargo.akuity.io/claim.<name> annotations
+// and consolidates them into a single map where the value of each key is sorted and deduped.
```
krancour: I'm putting myself in the shoes of a new contributor, or even my future self who will have forgotten most of this, and think it would be useful context if the "new style" vs "old style" that appear in other comments below were more explicitly explained here.

```suggestion
// OIDCClaimsFromAnnotationValue parses the values of the newer, preferred
// rbac.kargo.akuity.io/claims annotation as well as the values of older
// annotations with keys of the form rbac.kargo.akuity.io/claim.<name> and
// consolidates them into a single map of claim names to (deduped) claim values
// describing the set of users mapped to some ServiceAccount.
```

api/rbac/v1alpha1/annotations.go
```diff
@@ -30,3 +48,49 @@ func OIDCClaimNameFromAnnotationKey(key string) (string, bool) {
...
+	for k, v := range claims {
+		slices.Sort(v)
+		claims[k] = slices.Compact(v)
+	}
+	return claims, nil
+}
+
+// SetOIDCClaimsAnnotation sets the rbac.kargo.akuity.io/claims annotation on
+// the given ServiceAccount to the JSON representation of the given claims map.
+// It also removes any existing rbac.kargo.akuity.io/claim.<name> annotations.
```
krancour: Similar to https://github.com/akuity/kargo/pull/5002/files#r2353184706

```suggestion
// SetOIDCClaimsAnnotation marshals the provided map of claims to claim values
// and writes its string representation as the value of the ServiceAccount's
// rbac.kargo.akuity.io/claims annotation, which is the newer, preferred method
// of mapping users to ServiceAccounts. This function also removes
// any annotations with keys of the form rbac.kargo.akuity.io/claim.<name> (the
// older method) from the ServiceAccount.
```

internal/server/rbac/roles.go
```diff
@@ -742,55 +755,46 @@ func claimListToMap(claims []rbacapi.Claim) map[string][]string {
...
-	}
-	if sa.Annotations == nil {
-		sa.Annotations = map[string]string{}
-	}
-	// Add or update the annotations for the new claims
-	for name, values := range claims {
-		sa.Annotations[rbacapi.AnnotationKeyOIDCClaim(name)] = strings.Join(values, ",")
-	}
+func replaceClaimAnnotations(sa *corev1.ServiceAccount, claims map[string][]string) error {
+	return rbacapi.SetOIDCClaimsAnnotation(sa, claims)
```
krancour: We may as well drop this function and call `rbacapi.SetOIDCClaimsAnnotation()` directly.

internal/server/rbac/roles.go
```diff
@@ -742,55 +755,46 @@ func claimListToMap(claims []rbacapi.Claim) map[string][]string {
...
+	}
 	for name, values := range claims {
-		if existing, exists := sa.Annotations[rbacapi.AnnotationKeyOIDCClaim(name)]; exists {
-			values = append(strings.Split(existing, ","), values...)
-		}
-		slices.Sort(values)
-		values = slices.Compact(values)
-		if sa.Annotations == nil {
-			sa.Annotations = map[string]string{}
+		if _, exists := existingClaims[rbacapi.AnnotationKeyOIDCClaim(name)]; exists {
```
krancour: Calling `rbacapi.AnnotationKeyOIDCClaim()` seems its going to give you back an old style annotation key, but the map of existing claims have _just_ the claim names as the keys.

```suggestion
		if _, exists := existingClaims[name]; exists {
```

docs/docs/50-user-guide/50-security/20-access-controls/index.md
```diff
@@ -90,9 +89,13 @@ metadata:
...
   annotations:
-    rbac.kargo.akuity.io/claim.sub: alice,bob
-    rbac.kargo.akuity.io/claim.email: carl@example.com
-    rbac.kargo.akuity.io/claim.groups: devops,kargo-admin
+    rbac.kargo.akuity.io/claims: |
+      {
+        "sub": ["alice", "bob" ],
+        "email": "carl@example.com",
+        "groups": ["devops", "kargo-admin"],
+        "special:key": "value"
```
krancour: Did we end up not supporting scalar claim values?

```suggestion
        "special:key": ["value"]
```

fuskovic: yeah we agreed everything has to be an array e.g. `map[string][]string`

fuskovic: I'll address this in other areas it applies too

docs/docs/50-user-guide/50-security/20-access-controls/index.md
```diff
@@ -72,16 +72,15 @@ mapped. This search is mostly limited to `ServiceAccount` resources in Kargo
 project namespaces only (i.e. only those labeled with
 `kargo.akuity.io/project: "true"`).
 
-ServiceAccount resources may be mapped to users through the use of annotations
-whose key begins with `rbac.kargo.akuity.io/claim.`. The value of the annotation
-may be a single value, or a comma-delimited list of values.
+ServiceAccount resources may be mapped to users through the use of annotation claims defined under the `rbac.kargo.akuity.io/claims` key. The value of the annotation can be a standard YAML map or a JSON object.
```
krancour: I know that especially with the multi-line syntax used starting on line +93, this appears to be an object, but `|` signals that a multi-line _string_ follows, so it is in fact a string representation of an object.

```suggestion
ServiceAccount resources may be mapped to users via the
`rbac.kargo.akuity.io/claims` annotation, whose value is a string representation
of a JSON or YAML object with claim names as its keys and lists of claim values
as its values.
```

fuskovic: Updated this in both places

docs/docs/50-user-guide/50-security/20-access-controls/index.md
```diff
@@ -90,9 +92,13 @@ metadata:
...
-    rbac.kargo.akuity.io/claim.email: carl@example.com
-    rbac.kargo.akuity.io/claim.groups: devops,kargo-admin
+    rbac.kargo.akuity.io/claims: |
+      {
+        "sub": ["alice", "bob" ],
+        "email": "carl@example.com",
+        "groups": ["devops", "kargo-admin"],
+        "special:key": ["value"]
+      }
 ```
```
krancour, line 101: I wonder if we should briefly acknowledge the older syntax here in an info box so that a user who encounters it in the wild will be aware that any mappings expressed that way are, in fact, still in-force.

fuskovic: We could do something like:

> We also support the format `rbac.kargo.akuity.io/claim.<name>`.  However, this format does not allow for special characters; like having a colon in the key. e.g. `special:key`. With that said, the JSON format in the example above is the recommended format from now on.

Wdyt?

krancour: I think we need to be explicit that the effective mappings are the _union_ of mappings defined using both the old and new style. (That should be a rare occurrence, but it could happen. You made a point to test such cases, after all.) If we're not explicit about this, it opens the door to head scratching over which method takes precedence. Neither, as it turns out, but we should make sure that's clear.

Feel free to tweak this further:

```
Mappings specified using annotations with keys of the form
`rbac.kargo.akuity.io/claim.<name>` with comma-delimited values are also
supported for reasons of backwards compatibility. The effective mapping is
therefore is the union of mappings defined using such annotations with any
mappings defined using the newer, recommended `rbac.kargo.akuity.io/claims`
annotation.
```

krancour (approved):
🔥 Thank you for the hard work and patience on this.

@jessesuen will be happy he can use `cognito:groups` now.

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -123,13 +138,9 @@ api:
...
     admins:
-      claims:
-        groups:
-        - devops
+      claims: '{"groups":["devops"]}'
     viewer:
-      claims:
-        groups:
-        - developers
+      claims: '{"groups":["developers"]}'
```
jessesuen: Now that we have `kargo-user` ServiceAccount, I recommend we update docs to use `kargo-user` instead of `kargo-viewer` since `kargo-user` is a more limited, and a safer default.

krancour: There is `projectCreators` also.

krancour: We can mention all four roles, but I don't think the examples need to extend as far as showing the "behind the scenes" part for all four.

fuskovic: Should those examples extend to `admins`, `projectCreators`, and `users` only?

krancour: imho, show all four in the snippet of `values.yaml`, but when we get to the bit that reads "Behind the scenes, the configuration above..." we don't need to show the annotations on all four of the affected ServiceAccounts. It seems like just showing one would get the point across that what you put in the `values.yaml` just makes its way to an annotation on some chart-managed SA. (We currently show two, and even that seems like it's unnecessary to get the point across.)

fuskovic: https://github.com/akuity/kargo/pull/5002/commits/160e326b48bd0e94594597f197086a51e8766621

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -92,11 +94,24 @@ metadata:
...
   annotations:
-    rbac.kargo.akuity.io/claim.sub: alice,bob
-    rbac.kargo.akuity.io/claim.email: carl@example.com
-    rbac.kargo.akuity.io/claim.groups: devops,kargo-admin
+    rbac.kargo.akuity.io/claims: |
+      {
+        "sub": ["alice", "bob" ],
+        "email": "carl@example.com",
+        "groups": ["devops", "kargo-admin"],
+        "special:key": ["value"]
```
krancour: It's great that non-alphanumeric characters are allowed by the changes in this PR, but I don't think we need to go out of our way to show an example of it. It should be a relatively rare thing.

```suggestion
        "special:key": ["value"]
```

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -74,16 +74,18 @@ project namespaces only (i.e. only those labeled with
...
+as its values.
 
 In the following example, the `ServiceAccount` resource is mapped to all of:
 
 * Users with a `sub` claim identifying them as either `alice` or `bob`.
 * A user with the `email` claim `carl@example.com`.
-* All users with a `groups` claim  containing _either_ the `devops` or
+* All users with a `groups` claim containing _either_ the `devops` or
   `kargo-admin` group.
+* Users with a `special:key` claim.
```
krancour: Related to https://github.com/akuity/kargo/pull/5002/files#r2356463219

```suggestion
```

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -110,31 +125,41 @@ As previously mentioned, _most_ access controls are managed at the project level
 by project admins, however, there are two ways in which an operator can also
 map users to `ServiceAccount` resources.
 
-#### `api.oidc.admins` / `api.oidc.viewers`
+#### `api.oidc.admins` / `api.oidc.viewers` / `api.oidc.users` / `api.projectCreators`
```
krancour: This heading didn't feel _too_ bad when there were only two settings, but with four now, this is starting to look unwieldy.

I think we could simplify the heading to something like this:

```suggestion
#### Built-in System Roles
```

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -110,31 +125,41 @@ As previously mentioned, _most_ access controls are managed at the project level
...
+
+The `api.oidc.admins`, `api.oidc.projectCreators`, `api.oidc.users`, and
+`api.oidc.viewers` configuration options of the Kargo Helm chart permit an 
+operator to map users with specific claims to _system-wide_ admin, 
+project creator, user, and viewer roles respectively.
```
krancour: I think we should make this a table showing the settings and a description of the permissions they include.

docs/docs/40-operator-guide/40-security/30-access-controls.md
```diff
@@ -110,31 +123,41 @@ As previously mentioned, _most_ access controls are managed at the project level
...
 api:
   oidc:
     # ... omitted for brevity ...
     admins:
-      claims:
-        groups:
-        - devops
+      claims: '{"groups":["devops"]}'
+    projectCreators:
+      claims: '{"groups":["sales"]}'
```
krancour: This is an odd example -- it would be odd for sales to touch Kargo at all. Maybe change this to something like "team-leads"?

```suggestion
    projectCreators:
      claims: '{"groups":["sales"]}'
```

PR #5015, opened 2025-09-08
fix: remove dangling : prefix when username claim env is not set

gdsoumya:


krancour:
@gdsoumya DM'ed you.

While the bit of code you're patching here is problematic for _other_ reasons, ever since #4734, `OIDC_USERNAME_CLAIM` should reliably be set anytime OIDC is in use...

So I wonder if there's actually something else going on that caused the problem you're attempting to solve for to have surfaced.

Issue #4954, opened 2025-08-28
disclaimer about service/v1alpha1 stability


krancour:
Follow up to #4901

Per @jessesuen, add disclaimer to generated API docs generated from .proto files that there are no stability guarantees for the `service/v1alpha1` API.

PR #5020, opened 2025-09-09
docs: generate disclaimer for service/v1alpha1

fuskovic:
closes: https://github.com/akuity/kargo/issues/4954


docs/api/docs.templ
```diff
@@ -15,9 +15,18 @@
...
 
-## {{.Name | trimPrefix "api/" | trimSuffix "/service.proto" | trimSuffix "/generated.proto" }}
-{{.Description}}
+{{- if eq $name "service/v1alpha1" }}
+
+:::warning
+Stability is not guaranteed.
+:::
+
+{{- end }}
```
krancour: Just so we don't lose track... we spoke offline.

We got a better handle on what is actually _in_ these generated docs: The table at the top is service names with their input and output types and _everything_ else on the page are all types.

I think we concluded all we need to do is "drag" the `service/v1alpha1` heading and the disclaimer up above the table.

I haven't looked at the template close enough to know whether that's actually easily done or not... It's possible that is trickier than it sounds. 🤷‍♂️

krancour (approved):
Small change. Big improvement. 🔥🔥🔥

PR #5025, opened 2025-09-10
feat(ui): Drag and Drop Promotion

### LINKED ISSUE #4894: Improve Promotion Workflow with Drag-and-Drop
# Current behavior
Currently, promoting a freight requires several manual clicks. Additionally, unapproved freight for specific stages needs extra manual approval, which slows down the process.

# Proposed Improvement
Introduce a drag-and-drop feature that allows users to quickly initiate a promotion of any freight to a selected stage. This would reduce the number of clicks and speed up the workflow.

Benefits:
- Faster freight promotion
- More intuitive workflow
- Less manual approval required for specific stages
- Better user experience

Drop Area mockup:
Image

rpelczar:
Fixes #4894 

https://github.com/user-attachments/assets/016e8e3f-bd05-4715-8e1b-2481d20a7fd5

### DIFF
--- a/ui/src/features/common/confirm-modal/confirm-modal.tsx
+++ b/ui/src/features/common/confirm-modal/confirm-modal.tsx
@@ -1,32 +1,42 @@
 import { Modal, ModalFuncProps } from 'antd';
+import React from 'react';
 
 export interface ConfirmProps {
   title: string | React.ReactNode;
-  onOk: () => void;
+  onOk: () => Promise<unknown> | void;
   hide?: () => void;
   content?: string | React.ReactNode;
 }
 
 export const ConfirmModal = ({
   onOk,
+  onCancel,
   title = 'Are you sure?',
   content,
   hide,
-  visible,
   ...props
 }: ConfirmProps & ModalFuncProps) => {
-  const onConfirm = () => {
-    onOk();
+  const [loading, setLoading] = React.useState(false);
+
+  const confirm = async () => {
+    setLoading(true);
+    await onOk();
+    hide?.();
+    setLoading(false);
+  };
+
+  const cancel = () => {
+    onCancel?.();
     hide?.();
   };
 
   return (
     <Modal
-      open={visible}
-      onCancel={hide}
+      onCancel={cancel}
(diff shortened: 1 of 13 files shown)


ui/src/features/common/confirm-modal/confirm-modal.tsx
```diff
@@ -1,32 +1,42 @@
...
-  const onConfirm = () => {
-    onOk();
+  const [loading, setLoading] = React.useState(false);
+
+  const confirm = async () => {
+    setLoading(true);
+    await onOk();
+    hide?.();
+    setLoading(false);
+  };
```
jmalisz-akuity: Add try catch and set loading to false on `onOk` rejection as well

ui/src/features/project/pipelines/nodes/stage-node.tsx
jmalisz-akuity, line 1: Stage nodes that shouldn't be droppable (they are from other warehouse) are still animating even if they don't have the promote box visible. I think it's not a big issue, but if it could be tweaked then I think it would be nice.

ui/src/features/project/pipelines/freight/freight-card.tsx
```diff
@@ -244,66 +265,83 @@ export const FreightCard = (props: FreightCardProps) => {
...
+              }
             >
               Select
             </Button>
+          </Tooltip>
+        ) : (
+          <div
+            {...listeners}
+            {...attributes}
+            className='bg-gray-100 rounded text-center cursor-pointer hover:bg-gray-200'
```
jmalisz-akuity: The hover style sometimes disappears when dragging depending on where you start it and moves

rpelczar:
> There's some kind of bug:
> 
> 1. Hide part of the graph.
> 2. Use drag & drop to promote remaining stage (for me it was the first one).
> 3. Cancel promotion.
> 4. Open the hidden graph.
> 5. The promotion screen appears again.

@jmalisz-akuity good catch, but it's not related to this PR, and the same bug you can find on the latest main. I'm going to open an issue.

PR #5035, opened 2025-09-11
chore: update inline OIDC docs

bhavyabhut:

PR #5038, opened 2025-09-11
fix(docs): document dex callbacks

krancour:
Alternative to #5035

This surfaces the same missing information that #5035 would have, but is more particular about preserving the existing structure of the page and not mixing Dex details into sections where there is a clearly stated assumption that you're working with a compatible IDP.

### DIFF
--- a/docs/docs/40-operator-guide/40-security/20-openid-connect/index.md
+++ b/docs/docs/40-operator-guide/40-security/20-openid-connect/index.md
@@ -90,7 +90,7 @@ When installing Kargo with Helm, all options related to OIDC are grouped under
 1. Only if your IDP supports _both_ OIDC and PKCE:
 
     :::caution
-    If your IDP does not support _both_ OIDC and PKCE,
+    If your IDP does not support _both_ OIDC and PKCE, the
     [Adapting Incompatible IDPs](#adapting-incompatible-idps) section provides
     alternate instructions for this step.
     :::


docs/docs/40-operator-guide/40-security/20-openid-connect/index.md
```diff
@@ -58,18 +58,27 @@ To enable integration with your IDP, you will first need to register Kargo as a
 client. Carefully consult your IDP's documentation for instructions on how to do
 so.
 
-The callback URLs you will need to register are:
+The callback URL(s) you will need to register depend on whether your IDP
+supports both OIDC and PKCE:
 
-- `https://<hostname for api server>/login` (for the UI)
-- `https://localhost/auth/callback` (for the CLI)
+**With a compatible identity provider:**
```
krancour, line 64: Because the current structure of the page aggregates all Dex-related details in an "Adapting Incompatible IDPs" section, with no mention of "Dex" having been made up to this point on the page, (compared to #5035) this drives the selection of URLs based on protocols/compatibility instead of a term that's not yet been defined.

docs/docs/40-operator-guide/40-security/20-openid-connect/index.md
```diff
@@ -80,6 +89,12 @@ When installing Kargo with Helm, all options related to OIDC are grouped under
 
 1. Only if your IDP supports _both_ OIDC and PKCE:
 
+    :::caution
+    If your IDP does not support _both_ OIDC and PKCE,
+    [Adapting Incompatible IDPs](#adapting-incompatible-idps) section provides
+    alternate instructions for this step.
+    :::
```
krancour: Since this section leads with "Only if your IDP supports _both_ OIDC and PKCE," we do well to link to the alternate instructions you should reference if it does _not_. Having this here is preferable to mixing Dex details directly into this section, as the configuration with vs without Dex are really quite different.

docs/docs/40-operator-guide/40-security/20-openid-connect/index.md
```diff
@@ -80,6 +89,12 @@ When installing Kargo with Helm, all options related to OIDC are grouped under
 
 1. Only if your IDP supports _both_ OIDC and PKCE:
 
+    :::caution
+    If your IDP does not support _both_ OIDC and PKCE,
```
krancour:
```suggestion
    If your IDP does not support _both_ OIDC and PKCE, the
```

Issue #4959, opened 2025-08-29
Freight creation filters that work across subscriptions


jessesuen:
# Proposed Feature

Currently, when creating Freight, we provide some options/filters for when to create it. For example:
* whether it should be automatic or manual
* image tag constraints and imageSelectionStrategy
* expressionFilter / path filters for git (author, subject)
* etc...

One use case we can't currently solve is when there are multiple subscriptions, we may be subjected to race conditions when it discovers an artifact in one subscription that is incompatible with the latest artifact in another subscription.

Here is one such problematic scenario. Suppose we were combining two images, frontend and backend. The frontend and backend have compatibility requirements and must have matching tags. 

```yaml
spec:
  freightCreationPolicy: Automatic
  subscriptions:
  - image:
      repoURL: ghcr.io/mycompany/frontend
  - image:
      repoURL: ghcr.io/mycompany/backend
```

The following race might happen:
1. CI pushes frontend:v2 (but hasn't finished pushing backend:v2)
2. Kargo Warehouse discovers frontend:v2, but combines it with backend:v1 to create a buggy Freight that mixes v1 with v2.
3. Freight is now created, but should never be promoted because of an incompatibility

Currently, we advise people to use a "gatekeeper stage" (https://docs.kargo.io/user-guide/patterns/#gatekeeper-stage), but the desire is for the problematic Freight to never be created in the first place.

This proposal is to allow some additional filters on whether or not the Freight should be generated, but would work across subscriptions. For example, one filter could ensure that the tags must always be equal:

```yaml
spec:
  freightCreationPolicy: Automatic
  subscriptions:
  - image:
      repoURL: ghcr.io/mycompany/frontend
  - image:
      repoURL: ghcr.io/mycompany/backend
  freightCreationFilters:
  - expression: subscriptions[0].tag == subscriptions[1].tag
```

The expressions might also have access to metadata, OCI annotations of images, etc... for even deeper controls

# Motivation

It would prevent problematic Freight being generated caused by unlucky, mis-timed discovery. This problem is exacerbated when users are using webhooks. Because we would discover Freight even faster, and cause incompatible combinations.

fuskovic:
Took a look at this today. 

An expression like this:

```
  subscriptions:
  - image:
      repoURL: ghcr.io/mycompany/frontend
  - image:
      repoURL: ghcr.io/mycompany/backend
  freightCreationFilters:
  - expression: subscriptions[n].tag == subscriptions[n].tag
```

[...]

PR #5047, opened 2025-09-12
feat: freight creation criteria expressions

fuskovic:
Closes: https://github.com/akuity/kargo/issues/4959


fuskovic:
I'd like to do docs for this in a follow up if that's alright.

cc: @krancour

internal/controller/warehouses/warehouses.go
```diff
@@ -646,6 +680,52 @@ func validateDiscoveredArtifacts(
...
+// no expression is defined, or no discovered artifacts are present.
+// A non-nil error is returned if there was an error evaluating the expression.
+func freightCreationFilterSatisfied(
+	ctx context.Context,
+	wh *kargoapi.Warehouse,
+	artifacts *kargoapi.DiscoveredArtifacts,
+) (bool, error) {
+	logger := logging.LoggerFromContext(ctx)
+
+	if wh.Spec.FreightCreationFilters.Expression == "" {
```
thomastaylor312: Super nit: It might be worth doing a `TrimSpace` on the expression before checking for empty. This is a nit because it is highly unlikely someone will do `expression: "       "`, but it could happen if someone is templating it or something else. If I remember correctly, passing empty anything to `expr.Compile` will then result in an error

internal/controller/warehouses/warehouses_test.go
```diff
@@ -901,7 +1050,10 @@ func TestValidateDiscoveredArtifacts(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			result := validateDiscoveredArtifacts(tc.warehouse, tc.newStatus)
+			logger, err := logging.NewLogger(logging.DebugLevel, logging.DefaultFormat)
```
thomastaylor312: Nit: This can make the test output fairly verbose, I'd recommend setting to `ErrorLevel`

internal/controller/warehouses/warehouses_test.go
```diff
@@ -1155,3 +1307,315 @@ func TestReconcile(t *testing.T) {
...
+						Versions: []string{"v1.1.0"},
+					},
+				},
+			},
+			expected:    false,
+			errExpected: false,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			logger, err := logging.NewLogger(logging.DebugLevel, logging.DefaultFormat)
```
thomastaylor312: Same nit here around log level

internal/expressions/function/warehouse_functions.go
```diff
@@ -0,0 +1,302 @@
...
+				)
+				for i, dr := range artifacts.Git {
+					if dr.RepoURL != repoURL {
+						continue
+					}
+					logger.Debug("checking discovered git artifact",
+						"index", i,
+						"numCommits", len(dr.Commits),
+					)
+					for _, c := range dr.Commits {
```
thomastaylor312: Just checking here as I'm not as familiar with the warehouse code: Is the slice of commits/images (this comment applies to both) guaranteed to be in any order? If it is, we should just use that (I thought it would be ordered by default from oldest to newest). If it isn't, would it be better if we did ensure discovered freight is ordered by time both for humans (looking at the data and seeing which is the latest) and for use cases such as this?

internal/expressions/function/warehouse_functions.go
```diff
@@ -0,0 +1,302 @@
...
+					logger.Debug("discovered image artifact",
+						"index", i,
+						"numImageRefs", len(dr.References),
+					)
+					for _, ref := range dr.References {
+						if latestImgRef == nil {
+							latestImgRef = &ref
+							continue
+						}
+						if ref.CreatedAt.After(latestImgRef.CreatedAt.Time) {
```
thomastaylor312: Related to my other comment: We probably can't use `CreatedAt` as the way we sort things here. Based on the docs on the struct:

```go
// CreatedAt is the time the image was created. This field is optional, and
// not populated for every ImageSelectionStrategy.
CreatedAt *metav1.Time
```

Because it is optional, that means this code could end up with a nil pointer exception _and_ it means we can't assume it will always be set. So we might have to go with the ordering in the slice

internal/expressions/function/warehouse_functions.go
```diff
@@ -0,0 +1,302 @@
...
+					logger.Debug("checking discovered chart artifact",
+						"index", i,
+						"numVersions", len(dr.Versions),
+					)
+					v := getLatestVersion(dr)
+					if latestChartVersion == nil {
+						latestChartVersion = v
+						continue
+					}
+					if v.GreaterThan(latestChartVersion) {
```
thomastaylor312: So take this example scenario:

There is a Stage "dev" and Stage "prod". Software Foo rolls out a 2.0 version of their chart and software. I want to deploy 2.0 to "dev" so that I can start testing it, but I don't want to roll out to "prod" yet. Now let's say while I'm testing, 1.9.1 and 2.0.1 roll out with a security patch. I want to roll out 1.9.1 to "prod" but I don't want to roll out 2.0 to prod yet.

This is probably totally contrived and maybe a bad practice for laying out stages/promotions because I'm still relatively new to the ecosystem, so please tell me if that is the case (i.e. maybe you need to have a warehouse for `Foo-old` and `Foo-new`). However, if this is a case we need to support, then just taking latest version might not work

api/v1alpha1/warehouse_types.go
```diff
@@ -121,6 +125,12 @@ const (
 	FreightCreationPolicyManual FreightCreationPolicy = "Manual"
 )
 
+// FreightCreationFilters defines filters that must be satisfied for Freight to be created.
+type FreightCreationFilters struct {
+	// Expression is an expr-lang expression that must evaluate to true for Freight to be created.
```
thomastaylor312: Nit: For clarity sake, you might want to note that the expression should be wrapped in `${{` and `}}`

thomastaylor312 (dismissed):
Most of the core business logic looks good, but I have at least one blocking concern with the finding of the latest image as well as some of the other sorting of "latest" discovered artifacts

internal/expressions/function/warehouse_functions.go
thomastaylor312: Maybe 1.x and 2.x isn't the best comparison, but maybe something like 1.8 and 1.9 in the example above

internal/expressions/function/warehouse_functions.go
fuskovic: I would think so but I'm not 100% sure but yeah makes sense to me. We could sort here ourselves too just to be sure.

internal/expressions/function/warehouse_functions.go
fuskovic: I believe these concerns are addressed further upstream during artifact discovery by utils like semver and lexical selectors.

cc: @krancour can you keep me honest here

api/v1alpha1/warehouse_types.go
krancour: Well... I'm not sure that it should be, actually.

I think `${{ }}` is fantastic for contexts where it is necessary to distinguish between static and dynamic elements of a large string -- i.e. a template or a large blob of configuration like we use for promotions steps.

But in contexts where an expression and _only_ an expression is expected, `${{ }}` doesn't feel like it serves any real purpose.

api/v1alpha1/warehouse_types.go
```diff
@@ -108,6 +108,10 @@ type WarehouseSpec struct {
 	//
 	// +kubebuilder:validation:MinItems=1
 	Subscriptions []RepoSubscription `json:"subscriptions" protobuf:"bytes,1,rep,name=subscriptions"`
+
+	// FreightCreationFilters defines filters that must be satisfied for Freight to be created.
+	// +kubebuilder:validation:Optional
+	FreightCreationFilters FreightCreationFilters `json:"freightCreationFilters" protobuf:"bytes,5,opt,name=freightCreationFilters"`
```
krancour: This should be a pointer so that nil clearly represents "no filters."

krancour: I think the docs on this field also need to explain whatever relationship it has to the FreightCreationPolicy field. i.e. Maybe call out that this field has no effect when the policy isn't automatic.

internal/controller/warehouses/warehouses.go
```diff
@@ -629,6 +633,36 @@ func validateDiscoveredArtifacts(
...
+			&metav1.Condition{
+				Type:               kargoapi.ConditionTypeHealthy,
+				Status:             metav1.ConditionTrue,
+				Reason:             "FreightCreationFilterNotSatisfied",
+				Message:            "freight creation filter expression not satisfied; skipping freight creation",
+				ObservedGeneration: warehouse.GetGeneration(),
+			},
+		)
+		return false
+	}
```
krancour: I'm not clear on why we're checking the filters in this function that seems only intended to validate discovery results instead of somewhere around line +289 where there is the existing check for creating Freight or not based on Freight creation policy. These filters are just new criteria for that same decision.

pkg/expressions/function/functions.go
krancour, line 1: This renaming doesn't feel quite right. There's a lot of stuff in this file having nothing to do with Freight.

internal/expressions/function/freight_functions.go
```diff
@@ -79,22 +79,13 @@ func StatusOperations(
...
 }
 
-// Warehouse returns an expr.Option that provides a `warehouse()` function
-// for use in expressions.
-//
-// The warehouse function creates a v1alpha1.FreightOrigin of kind
-// v1alpha1.Warehouse with the specified name.
-func Warehouse() expr.Option {
-	return expr.Function("warehouse", warehouse, new(func(name string) kargoapi.FreightOrigin))
-}
```
krancour, line 94: This should really stay close to the (current) `imageFrom()`, `chartFrom()` etc. functions, as it's very closely related. This is used for disambiguating which Warehouse the artifact should have come from in cases where there might be Freight from multiple Warehouses containing artifacts from one common repo.

api/v1alpha1/warehouse_types.go
thomastaylor312: Good to know, I was working off the impression that was mostly what we did for expr-lang.

pkg/expressions/function/functions.go
fuskovic: Do you prefer it all stay in the same file with the original name?

fuskovic: > Do you prefer it all stay in the same file with the original name?

Ended up just doing this

api/v1alpha1/warehouse_types.go
```diff
@@ -121,6 +127,12 @@ const (
 	FreightCreationPolicyManual FreightCreationPolicy = "Manual"
 )
 
+// FreightCreationFilters defines filters that must be satisfied for Freight to be created.
+type FreightCreationFilters struct {
+	// Expression is an expr-lang expression that must evaluate to true for Freight to be created.
+	Expression string `json:"expression,omitempty" protobuf:"bytes,1,opt,name=expression"`
+}
```
krancour: I know #4959 used the term "filters," but after more reflection, "criteria" seems a more accurate term.

```suggestion
// FreightCreationCriteria defines criteria that must be satisfied for Freight
// to be created automatically from new artifacts following discovery.
type FreightCreationCriteria struct {
	// Expression is an expr-lang expression that must evaluate to true for
	// Freight to be created automatically from new artifacts following discovery.
	Expression string `json:"expression,omitempty" protobuf:"bytes,1,opt,name=expression"`
}
```

api/v1alpha1/warehouse_types.go
```diff
@@ -108,6 +108,12 @@ type WarehouseSpec struct {
 	//
 	// +kubebuilder:validation:MinItems=1
 	Subscriptions []RepoSubscription `json:"subscriptions" protobuf:"bytes,1,rep,name=subscriptions"`
+
+	// FreightCreationCriteria defines criteria that must be satisfied for Freight to be created.
+	// This field has no effect when the FreightCreationPolicy isn't automatic.
```
krancour: Probably better to say no effect when `Manual` since it seems that `""` is implicitly treated as `Automatic`.

```suggestion
	// FreightCreationCriteria defines criteria that must be satisfied for Freight
	// to be created automatically from new artifacts following discovery. This
	// field has no effect when the FreightCreationPolicy is `Manual`.
```

pkg/controller/warehouses/warehouses.go
```diff
@@ -284,6 +287,35 @@ func (r *reconciler) syncWarehouse(
...
+				&metav1.Condition{
+					Type:   kargoapi.ConditionTypeHealthy,
+					Status: metav1.ConditionFalse,
+					Reason: "FreightCreationCriteriaError",
+					Message: fmt.Sprintf(
+						"failed to evaluate freight creation criteria: %s", err.Error(),
+					),
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
+			)
```
krancour, line 321: If I am reading existing code properly, this sort of error should additionally make the Ready false.

pkg/controller/warehouses/warehouses.go
```diff
@@ -284,6 +287,35 @@ func (r *reconciler) syncWarehouse(
...
+			conditions.Set(
+				&status,
+				&metav1.Condition{
+					Type:               kargoapi.ConditionTypeHealthy,
+					Status:             metav1.ConditionTrue,
+					Reason:             "FreightCreationCriteriaNotSatisfied",
+					Message:            "freight creation criteria not satisfied; skipping freight creation",
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
+			)
```
krancour: While I agree that failure to evaluate Freight creation criteria amounts to a Warehouse health and readiness _problem_, I don't think I would agree that successfully evaluating the criteria and finding it to be false speaks to Warehouse health or readiness.

But I might like @hiddeco to weigh in on that since his grasp of idiomatic use of conditions is better than my own.

internal/controller/warehouses/warehouses.go
```diff
@@ -646,6 +678,57 @@ func validateDiscoveredArtifacts(
...
+
+	if artifacts == nil || (len(artifacts.Git) == 0 && len(artifacts.Images) == 0 && len(artifacts.Charts) == 0) {
+		logger.Debug("no artifacts discovered")
+		return true, nil
+	}
+
+	ctx = logging.ContextWithLogger(ctx, logger.WithValues("filterExpression", expression))
+	program, err := expr.Compile(expression, function.WarehouseOperations(ctx, wh, artifacts)...)
+	if err != nil {
+		return false, fmt.Errorf("error compiling freight creation filter expression: %w", err)
```
krancour: This is just one spot where I notice we're still saying "filter" instead of "criteria." We should fix that here as well as other remaining uses of "filter."

internal/controller/warehouses/warehouses.go
```diff
@@ -646,6 +678,57 @@ func validateDiscoveredArtifacts(
...
+	}
+	logger.Debug("evaluated freight creation filter expression", "result", result)
+
+	switch result := result.(type) {
+	case bool:
+		return result, nil
+	default:
+		parsedBool, err := strconv.ParseBool(fmt.Sprintf("%v", result))
+		if err != nil {
+			return false, err
```
krancour: It would be helpful to wrap this error. Without doing so, I'm not sure it will be clear that the problem is that the provided expression did not evaluate to a bool.

pkg/expressions/function/functions.go
```diff
@@ -36,13 +39,31 @@ func FreightOperations(
 ) []expr.Option {
 	return []expr.Option{
 		Warehouse(),
-		CommitFrom(ctx, c, project, freightRequests, freightRefs),
-		ImageFrom(ctx, c, project, freightRequests, freightRefs),
-		ChartFrom(ctx, c, project, freightRequests, freightRefs),
+		CommitFromFreight(ctx, c, project, freightRequests, freightRefs),
+		ImageFromFreight(ctx, c, project, freightRequests, freightRefs),
+		ChartFromFreight(ctx, c, project, freightRequests, freightRefs),
```
krancour, line 46: I 100% get why you renamed these. It's to avoid conflict with the _other_, new implementations of `commitFrom()`, `imageFrom()`, etc. that are used in different contexts than these are.

The thing I'm worried about is `CommitFromFreight()`, for instance, doesn't really convey that it's used in the context of a _Promotion_ and `CommitFromWarehouse()` doesn't really convey well that it's used in the context of Freight creation criteria.

I don't really have any specific ideas in mind yet about how to better organize different "palettes" of functions for use in different contexts, but it's something I think we need to grapple with now so this doesn't become too confusing.

It would be a good area for you to focus on and propose some options.

krancour: This "palette" notion is already present to some extent in the existence of functions like `DataOperations()` or `FreightOperations()`. In different contexts, different sets of related functions are being selected for addition to the environment an expression is evaluated against, but as the number of functions grows, that strategy seems like it needs some refinement.

internal/expressions/function/functions.go
```diff
@@ -374,12 +443,77 @@ func getCommit(
...
+		}
+
+		logger := logging.LoggerFromContext(ctx).WithValues(
+			"repoURL", repoURL,
+			"warehouse", wh.Name,
+		)
+
+		var latestCommit *kargoapi.DiscoveredCommit
+		for _, s := range wh.Spec.Subscriptions {
+			if s.Git != nil && s.Git.RepoURL == repoURL && len(artifacts.Git) != 0 {
```
krancour: These URLs should be normalized before comparison. Same goes for the other functions.

pkg/controller/warehouses/warehouses.go
fuskovic: My rationale was to provide some sort of observability into why freight may not be created. 

To help with trouble shooting and maybe even surface it in the UI if necessary.

internal/git/git.go
```diff
@@ -76,3 +76,9 @@ func NormalizeURL(repo string) string {
 	}
 	return fmt.Sprintf("ssh://%s/%s", userHost, pathURL.String())
 }
+
+// RepoURLsEqual returns true if the two provided Git repository URLs are
+// equivalent after normalization.
+func RepoURLsEqual(repoURL1, repoURL2 string) bool {
+	return NormalizeURL(repoURL1) == NormalizeURL(repoURL2)
+}
```
krancour: I really like the convenience of this, but this may not be optimal. In the majority of cases where we're doing these comparisons, we have one repo URL in-hand and we're iterating over a list and trying to match it to something in the list. We'd usually normalize the one we have in-hand once up front, before the loop. By using this function, we'd unnecessarily double the number of normalizations we perform. Since the normalization processes involve URL parsing, regular expressions, and string manipulations, the added cost is significant for a large n.

pkg/controller/warehouses/warehouses.go
krancour: That makes sense, but perhaps that's a separate, new condition.

cc @hiddeco

pkg/expressions/function/functions.go
krancour: Let's keep the revised names (`CommitFromFreight`, etc.) They seem accurate enough.

I'll comment elsewhere on names I think should change...

internal/expressions/function/functions.go
```diff
@@ -108,19 +130,35 @@ func CommitFrom(
...
 		new(func(repoURL string) kargoapi.GitCommit),
 	)
 }
 
-// ImageFrom returns an expr.Option that provides an `imageFrom()` function for
+// CommitFromWarehouse returns an expr.Option that provides a `commitFrom()` function
+// for use in expressions.
+//
+// The commitFrom function finds the latest Git commit based on repository URL.
+func CommitFromWarehouse(
```
krancour: This would probably, more accurately, be `CommitFromDiscoveredArtifacts`.

Similar changes needed for other functions.

internal/expressions/function/functions.go
```diff
@@ -40,9 +43,28 @@ func FreightOperations(
...
+		FreightMetadata(ctx, c, project),
+	}
+}
+
+// WarehouseOperations returns a slice of expr.Option containing functions for
+// Warehouse operations.
+//
+// It provides `warehouse()`, `commitFrom()`, `imageFrom()`, and `chartFrom()`
+// functions that can be used within expressions.
+func WarehouseOperations(
```
krancour: Let's call this `DiscoveredArtifactsOperations`.

pkg/expressions/function/functions.go
```diff
@@ -40,9 +43,27 @@ func FreightOperations(
...
+		FreightMetadata(ctx, c, project),
+	}
+}
+
+// DiscoveredArtifactsOperations returns a slice of expr.Option containing functions for
+// Warehouse operations.
```
krancour:
```suggestion
// DiscoveredArtifactsOperations returns a slice of expr.Option containing
// functions for retrieving artifacts from a Warehouse's discovered artifacts.
```

pkg/expressions/function/functions.go
```diff
@@ -40,9 +43,27 @@ func FreightOperations(
...
+//
+// It provides `commitFrom()`, `imageFrom()`, and `chartFrom()`
+// functions that can be used within expressions.
```
krancour:
```suggestion
// It provides `commitFrom()`, `imageFrom()`, and `chartFrom()` functions for
// use in the context of expressions defining criteria that permit or block
// automatic Freight creation after artifact discovery. These functions behave
// identically to functions of the same names used within the context of a
// Promotion process, however, they are implemented differently since they
// resolve artifacts from different data.
```

pkg/expressions/function/functions.go
```diff
@@ -108,19 +129,35 @@ func CommitFrom(
...
-		getCommit(ctx, c, project, freightReqs, freightRefs),
+		getCommitFromFreight(ctx, c, project, freightReqs, freightRefs),
 		new(func(repoURL string, origin kargoapi.FreightOrigin) kargoapi.GitCommit),
 		new(func(repoURL string) kargoapi.GitCommit),
 	)
 }
 
-// ImageFrom returns an expr.Option that provides an `imageFrom()` function for
+// CommitFromDiscoveredArtifacts returns an expr.Option that provides a `commitFrom()` function
+// for use in expressions.
```
krancour:
```suggestion
// CommitFromDiscoveredArtifacts returns an expr.Option that provides a
// `commitFrom()` function for use, specifically, in expressions that define
// criteria that permit or block automatic Freight creation after artifact
// discovery.
```

pkg/expressions/function/functions.go
```diff
@@ -129,19 +166,35 @@ func ImageFrom(
...
-		getImage(ctx, c, project, freightReqs, freightRefs),
+		getImageFromFreight(ctx, c, project, freightReqs, freightRefs),
 		new(func(repoURL string, origin kargoapi.FreightOrigin) kargoapi.Image),
 		new(func(repoURL string) kargoapi.Image),
 	)
 }
 
-// ChartFrom returns an expr.Option that provides a `chartFrom()` function for
+// ImageFromDiscoveredArtifacts returns an expr.Option that provides an `imageFrom()` function for
+// use in expressions.
```
krancour:
```suggestion
// ImageFromDiscoveredArtifacts returns an expr.Option that provides an
// `imageFrom()` function for use, specifically, in expressions that define
// criteria that permit or block automatic Freight creation after artifact
// discovery.
```

pkg/expressions/function/functions.go
```diff
@@ -150,22 +203,30 @@ func ChartFrom(
...
 }
 
-// FreightMetadata returns an expr.Option that provides a `freightMetadata()` function for use in expressions.
-//
-// Usage:
-//   - `freightMetadata(freightRefName)` returns the entire metadata map for the Freight.
-//   - `freightMetadata(freightRefName, key)` returns the value for the given key
-//     (DEPRECATED; will be removed in v1.10).
+// ChartFromDiscoveredArtifacts returns an expr.Option that provides a `chartFrom()` function for
+// use in expressions.
```
krancour:
```suggestion
// ChartFromDiscoveredArtifacts returns an expr.Option that provides a
// `chartFrom()` function for use, specifically, in expressions that define
// criteria that permit or block automatic Freight creation after artifact
// discovery.
```

pkg/expressions/function/functions.go
```diff
@@ -498,12 +559,79 @@ func getCommit(
...
+
+		repoURL = urls.NormalizeGit(repoURL)
+
+		logger := logging.LoggerFromContext(ctx).WithValues(
+			"repoURL", repoURL,
+			"warehouse", wh.Name,
+		)
+
+		var latestCommit *kargoapi.DiscoveredCommit
+		for _, s := range wh.Spec.Subscriptions {
```
krancour: Do we really need to iterate over the subscriptions? It looks like we could actually skip right to iterating over `artifacts.Git`, or am I missing something?

Similar question for the other functions.

pkg/expressions/function/functions.go
```diff
@@ -498,12 +559,79 @@ func getCommit(
...
+					if urls.NormalizeGit(ca.RepoURL) != repoURL {
+						continue
+					}
+					logger.Debug("checking discovered git artifact",
+						"index", i,
+						"numCommits", len(ca.Commits),
+					)
+					slices.SortFunc(ca.Commits, func(a, b kargoapi.DiscoveredCommit) int {
+						return a.CreatorDate.Compare(b.CreatorDate.Time)
+					})
```
krancour: Not sure why we're sorting. Discovery should have sorted these already. iirc the 0 element is the best match for the selection criteria.

pkg/expressions/function/functions.go
```diff
@@ -600,6 +805,77 @@ func getChart(
...
+							continue
+						}
+						if sv.GreaterThan(latestSemver) {
+							latestChart = &kargoapi.Chart{
+								RepoURL: caRepoURL,
+								Name:    ca.Name,
+								Version: sv.String(),
+							}
+						}
+					}
```
krancour: Similar question to [here](https://github.com/akuity/kargo/pull/5047/files#r2377074103) with an additional question about why semver...

The selection criteria may not have been semver based.

So as in the other case, we should trust that the discovery process ordered the artifacts correctly.

pkg/expressions/function/functions.go
```diff
@@ -498,12 +552,60 @@ func getCommit(
...
+		}
+		return &kargoapi.GitCommit{
+			RepoURL:   repoURL,
+			ID:        latestCommit.ID,
+			Branch:    latestCommit.Branch,
+			Tag:       latestCommit.Tag,
+			Message:   latestCommit.Subject,
+			Author:    latestCommit.Author,
+			Committer: latestCommit.Committer,
+		}, nil
```
krancour: I realize now that you may have had the wrong impression about this data structure and I'm sorry I didn't catch on sooner. There will be at most _one_ element in `artifacts.Git` that is a match for the specified URL. Its `commits` field will list all relevant commits that discovery found for that repo, with the 0th element being the best match. After you've found that, you can stop looking. If you were to continue looking, you wouldn't find anything more.

We also want to return nil instead of an error if we don't find what we're looking for. Dealing with nils is easy to do in an expression, while an error, much less so.

Last, the `DiscoveredCommit` type looks quite similar to the `GitCommit` type. The former actually has _more_ information than the latter. The only thing the former lacks that the latter has is the `RepoURL` field, but that doesn't matter much because the user using this function in an expression already knows the repo URL -- they gave it to us a input. Long story short, we can return the `DiscoveredCommit` without unpacking it into a `GitCommit`.

So you want something more like this:

```suggestion
		if artifacts == nil {
			return nil, nil
		}
	
		repoURL = urls.NormalizeGit(repoURL)
	
		for _, ca := range artifacts.Git {
			if urls.NormalizeGit(ca.RepoURL) != repoURL {
				continue
			}
			if len(ca.Commits) > 0 {
				return commits[0]
			}
		}
	
		return nil, nil
```

pkg/expressions/function/functions.go
```diff
@@ -498,12 +551,41 @@ func getCommit(
...
+		}
+
+		repoURL = urls.NormalizeGit(repoURL)
+		for _, ca := range artifacts.Git {
+			if urls.NormalizeGit(ca.RepoURL) != repoURL {
+				continue
+			}
+			if len(ca.Commits) > 0 {
+				return ca.Commits[0], nil
+			}
```
krancour, line 578: Realized this was missing from my previous suggestion. We could give up if we got to here.

```suggestion
			}
			return nil, nil
```

pkg/expressions/function/functions.go
```diff
@@ -541,12 +624,41 @@ func getImage(
...
+		}
+
+		repoURL = urls.NormalizeImage(repoURL)
+		for _, ia := range artifacts.Images {
+			if urls.NormalizeImage(ia.RepoURL) != repoURL {
+				continue
+			}
+			if len(ia.References) > 0 {
+				return ia.References[0], nil
+			}
```
krancour, line 650:
```suggestion
			}
			return nil, nil
```

pkg/expressions/function/functions.go
```diff
@@ -600,6 +713,39 @@ func getChart(
...
+func getChartFromDiscoveredArtifacts(artifacts *kargoapi.DiscoveredArtifacts) exprFn {
+	return func(a ...any) (any, error) {
+		if len(a) != 1 {
+			return nil, fmt.Errorf("expected 1 argument, got %d", len(a))
+		}
+
+		repoURL, ok := a[0].(string)
+		if !ok {
+			return nil, fmt.Errorf("first argument must be string, got %T", a[0])
+		}
```
krancour, line 725: There should be an optional second argument that you're not currently accounting for: name.

"Classic" (http/s) chart repos work differently than OCI repos. One repo can have many, differently named charts in it, so for such a repo, name needs to be specified.

pkg/expressions/function/functions.go
```diff
@@ -150,22 +197,29 @@ func ChartFrom(
...
-//     (DEPRECATED; will be removed in v1.10).
-//
-// The second argument is deprecated as of v1.8. Prefer using the single-argument form.
+// The chartFrom function finds the latest Helm charts based on repository URL.
+func ChartFromDiscoveredArtifacts(artifacts *kargoapi.DiscoveredArtifacts) expr.Option {
+	return expr.Function(
+		"chartFrom",
+		getChartFromDiscoveredArtifacts(artifacts),
+		new(func(repoURL string) kargoapi.Chart),
+		new(func(repoURL string, chartName string) kargoapi.Chart),
```
krancour: From previous experience with expr-lang, it's safest for longer, more complex signatures to precede shorter, less complex ones.

For reference: https://github.com/akuity/kargo/pull/2998#issue-2693283605

```suggestion
		new(func(repoURL string, chartName string) kargoapi.Chart),
		new(func(repoURL string) kargoapi.Chart),
```

pkg/expressions/function/functions.go
```diff
@@ -600,6 +712,50 @@ func getChart(
...
+		}
+
+		repoURL = urls.NormalizeChart(repoURL)
+		for _, ca := range artifacts.Charts {
+			if urls.NormalizeChart(ca.RepoURL) != repoURL {
+				continue
+			}
+			if chartName != "" && ca.Name != chartName {
+				continue
+			}
```
krancour:
```suggestion
			if urls.NormalizeChart(ca.RepoURL) != repoURL || ca.Name != chartName {
				continue
			}
```

pkg/expressions/function/functions_test.go
```diff
@@ -232,7 +234,104 @@ func Test_getCommit(t *testing.T) {
 	}
 }
 
-func Test_getImage(t *testing.T) {
+func Test_getCommitFromDiscoveredArtifacts(t *testing.T) {
+	for _, tc := range []struct {
+		name       string
+		warehouse  *kargoapi.Warehouse
```
krancour: This doesn't seem to be needed. Same for other new tests.

pkg/expressions/function/functions_test.go
```diff
@@ -641,6 +827,131 @@ func Test_getChart(t *testing.T) {
...
+		},
+		{
+			name: "success with repo URL and name",
+			warehouse: &kargoapi.Warehouse{
+				Spec: kargoapi.WarehouseSpec{
+					Subscriptions: []kargoapi.RepoSubscription{
+						{
+							Chart: &kargoapi.ChartSubscription{
+								Name:    "kargo",
+								RepoURL: "oci://ghcr.io/akuity/kargo-charts/kargo",
```
krancour: Using an OCI repo URL here isn't technically hurting the test, but `name` is only ever used in conjunction with a classic http/s chart repository, so using an https URL here would paint a picture of more realistic usage.

```suggestion
								RepoURL: "https://charts.example.com",
```

pkg/expressions/function/functions_test.go
```diff
@@ -232,7 +234,104 @@ func Test_getCommit(t *testing.T) {
...
+		},
+		{
+			name:      "invalid arg type",
+			warehouse: nil,
+			args:      []any{1},
+			assertions: func(t *testing.T, result any, err error) {
+				require.Nil(t, result)
+				require.ErrorContains(t, err, "first argument must be string, got int")
+			},
+		},
```
krancour, line 259: Should we additionally be testing correct error handling when the optional second arg is not a string?

krancour:
This needs docs. The only ones currently included seem to be the generated service/v1alpha1 reference docs.

pkg/controller/warehouses/warehouses.go
```diff
@@ -646,6 +690,61 @@ func validateDiscoveredArtifacts(
...
+		logger.Debug("no freight creation criteria expression")
+		return true, nil
+	}
+
+	if artifacts == nil || (len(artifacts.Git) == 0 && len(artifacts.Images) == 0 && len(artifacts.Charts) == 0) {
+		logger.Debug("no artifacts discovered")
+		return true, nil
+	}
+
+	logger.WithValues("criteriaExpression", expression)
```
krancour: This line is having no effect.

It would need to be:

```
logger = logger.WithValues("criteriaExpression", expression)
```

That said, there's only one log message that is (conditionally) written after this point in the code, so you could just add this key/val pair to the log output on line +732.

Somewhat related, and not isolated to this PR, I think we're possibly adding too much noise to the logs lately. It's not a hill I'd die on, but you might consider knocking these messages down to trace level.

pkg/controller/warehouses/warehouses.go
```diff
@@ -646,6 +690,61 @@ func validateDiscoveredArtifacts(
...
 }
 
+// freightCreationCriteriaSatisfied evaluates the freight creation criteria
+// expression, if defined, and returns true if the expression is satisfied,
+// no expression is defined, or no discovered artifacts are present.
+// A non-nil error is returned if there was an error evaluating the expression.
+func freightCreationCriteriaSatisfied(
+	ctx context.Context,
+	wh *kargoapi.Warehouse,
+	artifacts *kargoapi.DiscoveredArtifacts,
```
krancour: It seems awkward to pass `artifacts` as a separate argument when the Warehouse actually includes them.

fuskovic: It doesn't include them. After discovery, the artifacts are assigned to a new status object. Not the existing one.

pkg/controller/warehouses/warehouses.go
krancour: This is really the main thing that needs some resolution at this point.

hiddeco: I agree with @krancour. Successfully evaluating criteria and finding them not satisfied indicates the Warehouse is functioning correctly, as it can create Freight, but the current artifacts don't justify it.

For the `Healthy` condition: This should reflect whether the Warehouse itself is working properly. Successfully evaluating criteria (regardless of outcome) shows it's healthy.

For the `Ready` condition: This should indicate whether the Warehouse is ready to perform its function. Since it can create Freight when warranted, it should be `Ready=True` even when criteria aren't currently satisfied.

Given this, I would lean towards introducing a new `CriteriaSatisfied` condition with normal polarity (`True` when we have matches and `False` if we don't).

pkg/controller/warehouses/warehouses.go
krancour: You're right. Nice catch. I might still suggest a tweak here. If you know that a Warehouse includes discovered artifacts, it's too easy to make the mistake I just did. The signature looks awkward unless you're keenly aware of the context in which the caller is invoking the function. i.e. You need to know that the discovered artifacts being passed in are actually _different/new_ as compared to any (older) discovered artifacts that may be embedded in the Warehouse before the signature starts to make sense.

The only part of the Warehouse you're using here is `wh.Spec.FreightCreationCriteria`, and especially with the name of the function being `freightCreationCriteriaSatisfied()`, this would feel more straightforward and less prone to confusion if you took just the criteria as an argument instead of a whole Warehouse.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -443,11 +443,52 @@ You can handle `nil` values gracefully in Expr using its
 features.
 :::
 
-### `commitFrom(repoURL, [freightOrigin])`
+### `commitFrom`
 
-The `commitFrom()` function returns a corresponding `GitCommit` object from the
-`Promotion` or `Stage` their `FreightCollection`. It has one required and one
-optional argument:
+The `commitFrom` function signature varies based on where it's being used.
```
krancour: Just so it sounds less scary, think we should work in "slightly":

```suggestion
The signature and return value of the `commitFrom()` function vary slightly
with the context in which it's used.
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -443,11 +443,52 @@ You can handle `nil` values gracefully in Expr using its
...
+
+In the context of `DiscoveredArtifacts`, the function signature is:
+
+`commitFrom(repoURL)`
+
+It has one required argument.
```
krancour: Seems to be how we did this in other places:

```suggestion
It has one required argument:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -443,11 +443,52 @@ You can handle `nil` values gracefully in Expr using its
...
+  - git:
+      repoURL: https://github.com/example/backend.git
+  freightCreationCriteria:
+  - expression: |
+  commitFrom('https://github.com/example/frontend.git').tag == 
+  comitFrom('https://github.com/example/backend.git').tag
+```
+
+In all other contexts such as `Promotion`, `Stage`, or `FreightCollection`,
+the function signature is:
```
```suggestion
In all other contexts, such as promotion and verification processes, the
function signature is:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -443,11 +443,52 @@ You can handle `nil` values gracefully in Expr using its
...
 
-### `commitFrom(repoURL, [freightOrigin])`
+### `commitFrom`
 
-The `commitFrom()` function returns a corresponding `GitCommit` object from the
-`Promotion` or `Stage` their `FreightCollection`. It has one required and one
-optional argument:
+The `commitFrom` function signature varies based on where it's being used.
+
+In the context of `DiscoveredArtifacts`, the function signature is:
```
krancour: This might not be well understood. Its meaning would probably be more clear if we talked about "Freight creation criteria":

```suggestion
In the context of an optional expression evaluated to determine whether criteria
have been met for automatic creation of a `Freight` resource following a
`Warehouse`'s artifact discovery process, the function signature is:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -484,16 +525,17 @@ config:
   commitID: ${{ commitFrom("https://github.com/example/repo.git", warehouse("my-warehouse")).ID }}
 ```
 
-### `imageFrom(repoURL, [freightOrigin])`
+### `imageFrom`
 
-The `imageFrom()` function returns a corresponding `Image` object from the
-`Promotion` or `Stage` their `FreightCollection`. It has one required and
-one optional argument:
+The `imageFrom` function signature varies based on where it's being used.
```
```suggestion
The signature of the `imageFrom()` function varies slightly with the context in
which it's used.
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -484,16 +525,17 @@ config:
   commitID: ${{ commitFrom("https://github.com/example/repo.git", warehouse("my-warehouse")).ID }}
 ```
 
-### `imageFrom(repoURL, [freightOrigin])`
+### `imageFrom`
```
```suggestion
### `imageFrom()`
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -443,11 +443,52 @@ You can handle `nil` values gracefully in Expr using its
 features.
 :::
 
-### `commitFrom(repoURL, [freightOrigin])`
+### `commitFrom`
```
```suggestion
### `commitFrom()`
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -484,16 +525,17 @@ config:
...
-The `imageFrom()` function returns a corresponding `Image` object from the
-`Promotion` or `Stage` their `FreightCollection`. It has one required and
-one optional argument:
+The `imageFrom` function signature varies based on where it's being used.
 
-- `repoURL` (Required): The URL of a container image repository.
-- `freightOrigin` (Optional): A `FreightOrigin` object (obtained from
-  [`warehouse()`](#warehousename)) to specify which `Warehouse` should provide
-  the image information.
+In the context of `DiscoveredArtifacts`, the function signature is:
```
```suggestion
In the context of an optional expression evaluated to determine whether criteria
have been met for automatic creation of a `Freight` resource following a
`Warehouse`'s artifact discovery process, the function signature is:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -484,16 +525,17 @@ config:
...
+
+`imageFrom(repoURL)`
+
+It has one required argument.
```
```suggestion
It has one required argument:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -522,27 +593,62 @@ config:
   imageTag: ${{ imageFrom("public.ecr.aws/nginx/nginx", warehouse("my-warehouse")).Tag }}
 ```
 
-### `chartFrom(repoURL, [chartName], [freightOrigin])`
+### `chartFrom`
```
```suggestion
### `chartFrom()`
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -522,27 +593,62 @@ config:
...
 
-The `chartFrom()` function returns a corresponding `Chart` object from the
-`Promotion` or `Stage` their `FreightCollection`. It has one required and two
-optional arguments:
+The `chartFrom` function signature varies based on where it's being used.
```
```suggestion
The signature of the `chartFrom()` function varies slightly with the context in
which it's used.
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -504,9 +546,38 @@ The returned `Image` object has the following fields:
...
+  comitFrom('ghcr.io/example/backend').tag
+```
+
+:::info
+The returned `Image` object is the same for all contexts. Regardless of whether its used for
+`DiscoveredArtifacts`, `Promotion`, `Stage`, or `FreightCollection` resources.
+:::
+
+In all other contexts such as `Promotion`, `Stage`, or `FreightCollection`, the
+function signature is:
```
```suggestion
In all other contexts, such as promotion and verification processes, the
function signature is:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -522,27 +593,62 @@ config:
...
 
-### `chartFrom(repoURL, [chartName], [freightOrigin])`
+### `chartFrom`
 
-The `chartFrom()` function returns a corresponding `Chart` object from the
-`Promotion` or `Stage` their `FreightCollection`. It has one required and two
-optional arguments:
+The `chartFrom` function signature varies based on where it's being used.
+
+In the context of `DiscoveredArtifacts`, the function signature is:
```
```suggestion
In the context of an optional expression evaluated to determine whether criteria
have been met for automatic creation of a `Freight` resource following a
`Warehouse`'s artifact discovery process, the function signature is:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -504,9 +546,38 @@ The returned `Image` object has the following fields:
...
+      repoURL: ghcr.io/example/backend
+  freightCreationCriteria:
+  - expression: |
+  commitFrom('ghcr.io/example/frontend').tag == 
+  comitFrom('ghcr.io/example/backend').tag
+```
+
+:::info
+The returned `Image` object is the same for all contexts. Regardless of whether its used for
+`DiscoveredArtifacts`, `Promotion`, `Stage`, or `FreightCollection` resources.
```
```suggestion
The returned `Image` object is the same in all contexts.
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -522,27 +593,62 @@ config:
...
+      repoURL: oci://example.com/my-other-chart
+  freightCreationCriteria:
+  - expression: |
+  chartFrom('oci://example.com/my-chart').version == 
+  chartFrom('oci://example.com/my-other-chart').version
+```
+
+:::info
+The returned `Chart` object is the same for all contexts. Regardless of whether its used for
+`DiscoveredArtifacts`, `Promotion`, `Stage`, or `FreightCollection` resources.
```
```suggestion
The returned `Chart` object is the same in all contexts.
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -522,27 +593,62 @@ config:
...
+:::
+
+In all other contexts such as `Promotion`, `Stage`, or `FreightCollection`, 
+it has one required and two
+optional arguments:
```
```suggestion
In all other contexts, such as promotion and verification processes, the
function signature is:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -522,27 +593,62 @@ config:
...
+
+`chartFrom(repoURL, [chartName], [freightOrigin])`
+
```
krancour, line 647:
```suggestion

It has one required and two optional arguments:

```

krancour:
Code LGTM. It's just doc tweaks now.

In addition to the updated expression docs, which look pretty good, btw, I think we need to also update the "Working with Warehouses" doc.

I would add a new section (with `##` level heading) after the existing `Working with Private Repositories` section. Something like `Automatic Freight Creation Criteria`.

It doesn't need to be super detailed about the functions. Just explain what the feature is for and then link to the expressions doc.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -443,11 +443,55 @@ You can handle `nil` values gracefully in Expr using its
...
+  freightCreationPolicy: Automatic
+  subscriptions:
+  - git:
+      repoURL: https://github.com/example/frontend.git
+  - git:
+      repoURL: https://github.com/example/backend.git
+  freightCreationCriteria:
+  - expression: |
+  commitFrom('https://github.com/example/frontend.git').tag == 
+  comitFrom('https://github.com/example/backend.git').tag
```
krancour: At the very least these need to be indented another two levels. Separately, I'm doubtful that splitting a comparison across two lines like this is a valid expression, although I've not tried it. I think this would make the most sense:

```suggestion
  - expression: |
      commitFrom('https://github.com/example/frontend.git').tag == comitFrom('https://github.com/example/backend.git').tag
```

Same in other spots.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -683,6 +683,16 @@ appropriate credentials must be made available in some form. The many available
...
+## Automatic Freight Creation Criteria
+
+For more granular control over the creation of your `Freight`, you can use 
+`Freight Creation Criteria` which allows you to define a condition (in the form
+of an expression) that must evaluate to `true` in order to proceed with minting
+new `Freight`. This can be useful if you need to compare information from 
+different artifacts among your `Warehouse` subscriptions. Check out the 
+[Expression Language Reference](../60-reference-docs/40-expressions.md) for more 
+details.
+
```
krancour, line 720: As mentioned, I think we ought not go into great detail about the expressions here. The link you included covers it, but I think an example here would still be a good idea. I think the examples you used in the expression reference doc are good and one could be reused here. (I like the image one best, because it feels like the one that will be most readily understood.) You could introduce the example with some text explaining the hypothetical use case of back end and front end versions always needing to match.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -683,6 +683,30 @@ appropriate credentials must be made available in some form. The many available
...
+For more granular control over the creation of your `Freight`, you can define 
+`Freight Creation Criteria` in the form of an expression. One potential use case
+could be backend and frontend versions needing to match.
+
+Example:
+
+```yaml
+spec:
+  freightCreationPolicy: Automatic
```
krancour, line 707: 🤔 Not a new problem, but this made me realize that `freightCreationPolicy` isn't currently mentioned anywhere in the docs.

Can you introduce this section by explaining that, by default, Warehouses create new Freight following a discovery cycle if it found any new artifacts, and the fact that this can we disabled?

It would be a good segue into your "For more granular control..."

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -683,6 +683,43 @@ appropriate credentials must be made available in some form. The many available
...
+```yaml
+spec:
+  freightCreationPolicy: Manual
+```
+
+## Automatic Freight Creation Criteria
+
+For more granular control over the creation of your `Freight`, you can define 
+`Freight Creation Criteria` in the form of an expression. One potential use case
+could be backend and frontend versions needing to match.
```
krancour, line 703: These two sections are so short and they cover similar/related subject matter, so I think we should combine them:


```suggestion
## Automatic Freight Creation

By default, `Warehouse`s create new `Freight` following a discovery cycle if 
new artifacts are found. This can be disabled by setting `freightCreationPolicy`
to `Manual`.

Example:

```yaml
spec:
  freightCreationPolicy: Manual
```

For more granular control over the creation of your `Freight`, you can define 
`Freight Creation Criteria` in the form of an expression. One potential use case
could be backend and frontend versions needing to match.
```

Comment (krancour) [inline docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md]: These two sections are so short and they cover similar/related subject matter, so I think we should combine them:

~~~suggestion
## Automatic Freight Creation

By default, `Warehouse`s create new `Freight` following a discovery cycle if 
new artifacts are found. This can be disabled by setting `freightCreationPolicy`
to `Manual`.

Example:

```yaml
spec:
  freightCreationPolicy: Manual
```

For more granular control over the creation of your `Freight`, you can define 
`Freight Creation Criteria` in the form of an expression. One potential use case
could be backend and frontend versions needing to match.
~~~

pkg/controller/warehouses/warehouses.go
```diff
@@ -284,6 +287,77 @@ func (r *reconciler) syncWarehouse(
...
+					Message:            "freight creation criteria not satisfied; skipping freight creation",
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
+				&metav1.Condition{
+					Type:               kargoapi.ConditionTypeReconciling,
+					Status:             metav1.ConditionFalse,
+					Reason:             "FreightCreationCriteriaNotSatisfied",
+					Message:            "freight creation criteria not satisfied; skipping freight creation",
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
```
krancour, line 355: I believe it's more idiomatic to delete the Reconciling condition if we're done.

pkg/controller/warehouses/warehouses.go
```diff
@@ -284,6 +287,77 @@ func (r *reconciler) syncWarehouse(
...
+					Type:   kargoapi.ConditionTypeReady,
+					Status: metav1.ConditionFalse,
+					Reason: "FreightCreationCriteriaError",
+					Message: fmt.Sprintf(
+						"failed to evaluate freight creation criteria: %s", err.Error(),
+					),
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
+			)
+			return status, fmt.Errorf("failed to evaluate freight creation criteria: %w", err)
```
krancour: If validation of discovered artifacts fails (existing code, quite a bit higher up from here), there is this:

```
	// Validate the discovered artifacts.
	if !validateDiscoveredArtifacts(warehouse, &status) {
		// Remove the reconciling condition and return early if the validation
		// failed. We do not return an error here, to prevent a requeue loop
		// which would cause unnecessary pressure on the upstream sources.
		conditions.Delete(&status, kargoapi.ConditionTypeReconciling)
		return status, nil
	}
```

I wonder if this isn't a similar scenario. Continued efforts to reconcile aren't going to make an invalid expression valid.

cc @hiddeco

pkg/controller/warehouses/warehouses.go
```diff
@@ -284,6 +287,77 @@ func (r *reconciler) syncWarehouse(
...
+		conditions.Set(
+			&status,
+			&metav1.Condition{
+				Type:               kargoapi.ConditionTypeCriteriaSatisfied,
+				Status:             metav1.ConditionTrue,
+				Reason:             "FreightCreationCriteriaSatisfied",
+				Message:            "freight creation criteria satisfied; creating freight",
+				ObservedGeneration: warehouse.GetGeneration(),
+			},
+		)
```
krancour: imho, we should delete the condition when the criteria are satisfied.

People will be interested in knowing why a Warehouse _didn't_ create Freight from recently discovered artifacts, but I don't think anyone will care about this condition under other circumstances.

pkg/controller/warehouses/warehouses.go
```diff
@@ -304,6 +378,12 @@ func (r *reconciler) syncWarehouse(
 				Reason:  "AwaitingFreightCreation",
 				Message: "Freight creation from latest artifacts is in progress",
 			},
+			&metav1.Condition{
+				Type:    kargoapi.ConditionTypeReady,
+				Status:  metav1.ConditionTrue,
+				Reason:  "ArtifactsDiscovered",
+				Message: conditions.Get(&status, kargoapi.ConditionTypeHealthy).Message,
+			},
```
krancour: I don't understand this. You're setting Ready to false and then immediately setting it to true.

pkg/controller/warehouses/warehouses.go
fuskovic: Good catch and that makes sense to me. We can return a nil error if the expression is invalid.  However, I think it would be a good idea to bubble this up in the UI. Otherwise it will be confusing for reconciliation to be finished with no errors but no new freight minted. Wdyt?

fuskovic: Nvm, I pushed the change and tested locally and it the error makes it's way through. e.g.

Screenshot 2025-09-30 at 3 24 49 PM
Screenshot 2025-09-30 at 3 25 02 PM

krancour:
Test driving again these are the conditions I observe when Freight creation criteria are met:

```
status:
  conditions:
    - lastTransitionTime: 2025-10-01T18:45:20Z
      message: Successfully discovered 3 images and 4 charts from 2 subscriptions
      reason: ArtifactsDiscovered
      status: "True"
      type: Ready
    - lastTransitionTime: 2025-10-01T18:45:20Z
      message: Successfully discovered 3 images and 4 charts from 2 subscriptions
      observedGeneration: 2
      reason: ArtifactsDiscovered
      status: "True"
      type: Healthy```
```

This is not a new problem, but I think the changes in this PR may compound it. Is having run a successful discovery really what makes Warehouse ready? Is it really what also makes it healthy? _Especially_ considering that the Warehouse went on to do _something else_ after discovery -- namely mint new Freight.

When Freight creation criteria are _not_ met:

```
status:
  conditions:
    - lastTransitionTime: 2025-10-01T18:45:20Z
      message: freight creation criteria not satisfied; skipping freight creation
      observedGeneration: 1
      reason: FreightCreationCriteriaNotSatisfied
      status: "True"
      type: Ready
    - lastTransitionTime: 2025-10-01T18:45:20Z
      message: Successfully discovered 3 images and 4 charts from 2 subscriptions
      observedGeneration: 1
      reason: ArtifactsDiscovered
      status: "True"
      type: Healthy
    - lastTransitionTime: 2025-10-01T18:45:26Z
      message: freight creation criteria not satisfied; skipping freight creation
      observedGeneration: 1
      reason: FreightCreationCriteriaNotSatisfied
      status: "False"
      type: CriteriaSatisfied
```

Similar questions apply. Is not having met the criteria really what makes this Warehouse ready? Identical to the previous case: Is having successfully run discovery really what makes it healthy, especially when the Warehouse went on to do other things afterwards?

cc @hiddeco for some input on what he would expect idiomatic conditions to be for these scenarios:

* Discovery failed
* Discovery succeeded, automatic Freight creation disabled
* Discovery succeeded, evaluation of Freight creation criteria fails
* Discovery succeeded, Freight creation criteria not met
* Discovery succeeded, Freight creation criteria met, creation failed
* Discovery succeeded, Freight creation criteria met, creation succeeded

hiddeco:
I think the following can be said about the condition types:

- `Healthy`: "Is the Warehouse controller functioning correctly?"
- `Ready`: "Has the Warehouse completed its reconciliation successfully?"
- `CriteriaSatisfied`:  "Do artifacts warrant Freight creation?"

Additionally, I think `Ready` (when it's `True`) should reflect a summary of the _primary_ thing the Warehouse is designed to do, which is the discovery of artifacts. Anything that relates to `Freight` can now be put in `CriteriaSatisfied`. Not having any criteria defined means that _anything_ warrants the creation of Freight, which still allows you to use it to reflect information.

---

**Discovery failed (any reason - transient or persistent):**

- `Healthy=False` (reason: `DiscoveryFailed`, message: `<descriptive error>`)
- `Ready=False` (reason: `DiscoveryFailed`, message: `<descriptive error>`)
- `CriteriaSatisfied` absent/not evaluated

**Discovery succeeded, automatic Freight creation disabled:**

- `Healthy=True` (reason: ReconciliationSucceeded)
- `Ready=True` (reason: `ArtifactsDiscovered`, message: `Successfully discovered X images and Y charts from Z subscriptions`)
- `CriteriaSatisfied` absent/not evaluated

**Discovery succeeded, evaluation of Freight creation criteria fails:**

- `Healthy=False` (reason: `CriteriaEvaluationFailed`, message: `<descriptive error>`)
- `Ready=False` (reason: `CriteriaEvaluationFailed`, message: `<descriptive error>`)
- `CriteriaSatisfied=Unknown` (or absent)

**Discovery succeeded, Freight creation criteria not met:**

- `Healthy=True` (reason: `ReconciliationSucceeded`)
- `Ready=True` (reason: `ArtifactsDiscovered`, message: `Successfully discovered X images and Y charts from Z subscriptions`)
- `CriteriaSatisfied=False` (reason: `InsufficientChanges`, message: `None of the discovered artifacts could match Freight creation criteria <X-Y-Z>`)

**Discovery succeeded, criteria met, Freight creation failed:**

- `Healthy=False` (reason: `FreightCreationFailed`, message: `<descriptive error>`)
- `Ready=False` (reason: `FreightCreationFailed`, message: `<descriptive error>`)
- `CriteriaSatisfied=True` (reason: `CriteriaMet` or `SufficientChanges`)

**Discovery succeeded, criteria met, Freight creation succeeded:**

- `Healthy=True` (reason: `ReconciliationSucceeded`)
- `Ready=True` (reason: `ArtifactsDiscovered`, message: `Successfully discovered X images and Y charts from Z subscriptions`)
- `CriteriaSatisfied=True` (reason: `FreightCreated`, message: `Created Freight xyz123 from discovered artifacts`)

fuskovic:
@hiddeco @krancour 

Invalid Expression

invalid_expression

Criteria Unmet

fixed_not-Met

Criteria Met

satisfied

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -683,6 +683,41 @@ appropriate credentials must be made available in some form. The many available
...
+
+Example:
+
+```yaml
+spec:
+  freightCreationPolicy: Manual
+```
+
+For more granular control over the creation of your `Freight`, you can define 
+`Freight Creation Criteria` in the form of an expression. One potential use case
```
krancour:
```suggestion
`Freight` creation criteria in the form of an expression. One potential use case
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -683,6 +683,41 @@ appropriate credentials must be made available in some form. The many available
...
+could be backend and frontend versions needing to match.
```
krancour: Nit: Just searched the docs to check and seems that everywhere these appear up to now, we've kept "back end" and "front end" as two words each.

I think other places below where you've used `backend` or `frontend` as image names, it would be reasonable to leave them as they are.

```suggestion
could be back end and front end versions needing to match.
```

pkg/controller/warehouses/warehouses.go
```diff
@@ -304,6 +382,13 @@ func (r *reconciler) syncWarehouse(
...
+			&metav1.Condition{
+				Type:               kargoapi.ConditionTypeCriteriaSatisfied,
+				Status:             metav1.ConditionTrue,
+				Reason:             "CriteriaMet",
+				Message:            "freight creation criteria satisfied",
+				ObservedGeneration: warehouse.GetGeneration(),
+			},
 		)
 
 		// Build a Freight from the latest discovered artifacts.
```
krancour: Not something you did, but while we're here... "a Freight" isn't grammatically correct.

Alternatively, "a Freight resource" would be acceptable as well.

```suggestion
		// Build Freight from the latest discovered artifacts.
```

pkg/controller/warehouses/warehouses.go
```diff
@@ -284,6 +288,80 @@ func (r *reconciler) syncWarehouse(
...
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
+			)
+			conditions.Delete(&status, kargoapi.ConditionTypeReconciling)
+			// return a nil error to avoid a requeue loop since subsequent
+			// retries are not going to make the expression any more valid.
+			return status, nil
+		}
+		if !criteriaSatisfied {
+			logger.Info("freight creation criteria not satisfied; skipping freight creation")
```
krancour: This is too high a priority. It should be debug or even trace.

Same with the other new occurrence of `logger.Info()` in this file.

```suggestion
			logger.Debug("freight creation criteria not satisfied; skipping freight creation")
```

hiddeco:
@fuskovic this looks like a great step forward. Some nits (partly because of my own limited instructions):

- I think the error for an invalid expression contains an error wrap that is redundant (`failed to evaluate ... error evaluating`). Additionally, from a UX perspective, the non-capitalized message is slightly awkward (but inherited from Go's error style) — I think it might be worth prettifying this.
- When criteria are not met, I think the `Healthy` reason/message can be the same as when they are met.

pkg/controller/warehouses/warehouses.go
```diff
@@ -259,6 +262,7 @@ func (r *reconciler) syncWarehouse(
 					ObservedGeneration: warehouse.GetGeneration(),
 				},
 			)
+			conditions.Delete(&status, kargoapi.ConditionTypeCriteriaSatisfied)
```
krancour: imho, this is misplaced. As it stands, we're clearing this condition only if there's been an error during discovery. I think it would make more sense to clear it up front at the same time we initially set Reconciling to true. This way, the CriteriaSatisfied condition remains absent throughout discovery.

cc @hiddeco for some confirmation of this.

pkg/controller/warehouses/warehouses.go
```diff
@@ -284,6 +288,74 @@ func (r *reconciler) syncWarehouse(
...
+					Message:            fmt.Sprintf("Criteria evaluation failed: %s", err.Error()),
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
+				&metav1.Condition{
+					Type:               kargoapi.ConditionTypeCriteriaSatisfied,
+					Status:             metav1.ConditionUnknown,
+					Reason:             "FreightCreationCriteriaError",
+					Message:            fmt.Sprintf("Criteria evaluation failed: %s", err.Error()),
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
```
krancour: imho, we can do without setting this to unknown. If the condition is absent from the start as I suggest [here](https://github.com/akuity/kargo/pull/5047/files#r2400069587), I would just let it continue to be absent.

My rationale is that the Healthy and Ready conditions already are reflecting that an error was encountered whilst attempting to evaluate the criteria. With those both being very clear about that point, I don't think anyone will wonder whether criteria were met or not or need to be told we don't know. It should be pretty obvious from the other conditions.

cc @hiddeco for confirmation.

pkg/controller/warehouses/warehouses.go
```diff
@@ -284,6 +288,74 @@ func (r *reconciler) syncWarehouse(
...
+				&metav1.Condition{
+					Type:               kargoapi.ConditionTypeReady,
+					Status:             metav1.ConditionTrue,
+					Reason:             "ArtifactsDiscovered",
+					Message:            conditions.Get(&status, kargoapi.ConditionTypeHealthy).Message,
+					ObservedGeneration: warehouse.GetGeneration(),
+				},
+			)
+			conditions.Delete(&status, kargoapi.ConditionTypeReconciling)
+			return status, nil
```
krancour: I think there's more stuff happening here than is necessary.

I would consider not updating Healthy or Ready conditions here and not deleting the reconciling conditions...

I would set the CriteriaSatisfied condition and then _not_ return early...

I'd put an `else` after this that isn't currently there, handle Freight creation within that block and then, only after _both_ the existing `if` and the new `else`, tidy up all the conditions to reflect success.

imho, all of this code becomes easier to follow if all successful paths through the code lead to conditions being set consistently in one place.

cc @hiddeco

krancour: Adding to this...

There's an existing spot in `validateDiscoveredArtifacts()` that updates Healthy to true and imho, that also should be removed, leaving final handling of success to one single chunk of code at the end of the `syncWarehouse()` function.

PR #5057, opened 2025-09-15
feat(controller): make registries admin-configurable

### LINKED ISSUE #1139: make registries admin-configurable
Supersedes #1106

Follow-up to #1058

Most of the time, details for how to talk to a given image registry can be inferred from an image's "prefix," but not always:

* Sometimes the API address is different from the hostname.

    Example: Images with no prefix _or_ with hostname `docker.io` in their prefix (e.g. `nginx` or `docker.io/nginx`) have an API address https://registry-1.docker.io

* Sometimes a registry has a "default namespace."

    Example: Images with no prefix or only a hostname in their prefix (e.g. `nginx` or `docker.io/nginx`) need to be prefixed with `library/` when interacting with the API. e.g. `nginx` becomes `library/nginx`.

* Rate limits cannot be inferred.

For well known, public registries like Docker Hub or Quay, it's sensible to hard-code accommodations for such idiosyncrasies, but when such idiosyncrasies arise in self-hosted registries, we need some mechanism for users to define the rules themselves.

I recommend that registries be configured _globally_ by a Kargo admin and _not_ be configured at the project level. #1058 already treats each registry configuration as global singletons so that:

* The registry cache is shared across projects

* Many projects all talking to the same registry are rate-limited as a pool instead of individually.

Ultimately, I believe this boils down to permitting registry idiosyncrasies to be defined by a Kargo admin either/both:

1. In a `values.yaml` file used during Kargo installation/upgrade. (The chart would incorporate this into a `ConfigMap`.)

1. A manually maintained `ConfigMap`.

johnpekcan:
Attempts to fix #1139 

This fix attempts passes in a list of possible root domain names where custom registries may live.
If the detected `imagePrefix` contains an item from the custom registry list -- use a custom rate limit, otherwise default to `20` (no change from what it is hardcoded as now).


krancour:
This appears to address _only_ the rate limit aspect of #1139. Were we to accept this approach to that one bit, without thinking holistically about the rest of it, we'd almost certainly have to incur deprecations and breaking changes when we fully resolve the issue at some point in the future.

Why I'm pretty certain we'd need an entirely different approach if we were looking at this holistically: There's only so much that you can do with environment variables. If we were to allow any number of registries to be configured with not only their own, individual rate limits, but _also_ things like a default namespace, or the API base address for cases where it cannot be properly inferred from an image URL's prefix, we're not likely to stuff that complex data structure into an environment variable. Instead, we'd be looking at options like a ConfigMap mounted to the file system, or perhaps more ideally, we'd add fields to the somewhat recently added ClusterConfig type that allow this to be runtime configurable instead of install-time configurable.

The bottom line is that I'd rather see more discussion and deliberation over how to proceed with #1139 instead of rushing into a partial solution.

johnpekcan:
That makes sense. I certainly wouldn't want to introduce a change knowing we'd have to do breaking change some point later. I'll try to noodle on it some more and see if I come up with a fix that's closer to your vision. Will let y'all know if/when i get to it. Cheers!

krancour:
> I'll try to noodle on it some more 

@johnpekcan if you want to propose what this config might look like if added to the ClusterConfig type, that would probably be a good jumping off point for something that would address this comprehensively.

johnpekcan:
@krancour , i'm quite happy with the way custom registries are used in argocd image updater: 
https://github.com/argoproj/argo-helm/blob/main/charts/argocd-image-updater/templates/configmap.yaml#L45-L47

I wouldn't be opposed to something like this 😇 
It was very easy to jot down a few lines in yaml and set my rate limiting.

krancour:
@johnpekcan it's been a while since I've looked at image updater in detail, but _if_ my memory serves, something quite like what they do would be nice, _but_ because we have the ClusterConfig and ProjectConfig types, I think those are two places we should consider "hanging" that config off of, with the obvious benefit of having it there instead of in a ConfigMap being that it will have well-defined structure and context help will work.

But again, if memory serves, where we land could look pretty similar.

PR #5067, opened 2025-09-17
refactor(api-server): dry up api endpoints

krancour:
I was bugged non-DRYness of the API endpoints and redundant checks for desired format (endpoint checks and `objectOrRaw()` also does).

This also ensures GVK info is not lost when marshaling a structured object to JSON or YAML.

### DIFF
--- a/internal/server/common.go
+++ b/internal/server/common.go
@@ -72,25 +72,55 @@ func splitYAML(
 	return projects, otherResources, nil
 }
 
-// objectOrRaw returns either the object or the raw representation of the object
-// based on the format.
-func objectOrRaw[T client.Object](obj T, format svcv1alpha1.RawFormat) (T, []byte, error) {
+// objectOrRaw takes structured or unstructured objects as input and depending
+// on requested format returns EITHER (but never both) the object serialized in
+// the requested format OR the object converted to the structured object type.
+func objectOrRaw[T client.Object](
+	c client.Client,
+	obj client.Object,
+	format svcv1alpha1.RawFormat,
+	t T,
+) (T, []byte, error) {
+	if _, ok := obj.(*unstructured.Unstructured); !ok {
+		// Structured objects are likely to be missing GVK information, so we add
+		// it in.
+		gvk, err := c.GroupVersionKindFor(t)
+		if err != nil {
+			return *new(T), nil,
+				fmt.Errorf("could not determine GVK for type: %w", err)
+		}
+		obj.GetObjectKind().SetGroupVersionKind(gvk)
+	}
 	switch format {
 	case svcv1alpha1.RawFormat_RAW_FORMAT_JSON:
 		raw, err := json.Marshal(obj)
 		if err != nil {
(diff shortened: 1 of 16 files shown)


internal/server/common.go
```diff
@@ -72,25 +72,55 @@ func splitYAML(
...
-// based on the format.
-func objectOrRaw[T client.Object](obj T, format svcv1alpha1.RawFormat) (T, []byte, error) {
+// objectOrRaw takes structured or unstructured objects as input and depending
+// on requested format returns EITHER (but never both) the object serialized in
+// the requested format OR the object converted to the structured object type.
+func objectOrRaw[T client.Object](
+	c client.Client,
+	obj client.Object,
+	format svcv1alpha1.RawFormat,
+	t T,
```
thomastaylor312, line 82: I think this `t` param (or the `T` return value) is not needed and this function signature could be cleaner. There are two approaches here I can think of:

- Remove `T` from the return signature so it is just `([]byte, error)` and then unmarshal into `t` if you do end up unmarshaling
- Remove `t` and then do this instead of the current `c.GroupVersionKindFor` call:

```go
// Or just move the declaration of `var newObj T` up to the top of the function
var objType T
gvk, err := c.GroupVersionKindFor(objType)
```

internal/server/common_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	testUnstructuredData := map[string]any{
+		"apiVersion": kargoapi.GroupVersion.String(),
+		"kind":       "Stage",
+		"metadata": map[string]any{
+			"name":      testStageName,
+			"namespace": "test-project",
+		},
+	}
+	testUnstructured := &unstructured.Unstructured{Object: testUnstructuredData}
```
thomastaylor312, line 41: Nit: It might be worth having a test where this data doesn't contain GVK info at all to ensure the GVK gets set properly

internal/server/common.go
krancour: > Remove T from the return signature so it is just ([]byte, error) and then unmarshal into t if you do end up unmarshaling

I actually wrote it _exactly_ like that initially, but the thing I couldn't get over was how "asymmetric" it felt. Understanding that you get a non-nil return value _or_ a nil return value, with one of the passed parameters having been modified _instead_ requires you to have read the function's doc string carefully. By contrast, understanding that one of the returned values will be nil and the other non-nil feels like it doesn't require reading the doc at all.

As to your second suggestion, I agree it looks cleaner, but it doesn't work, unfortunately. Because all the types that implement `client.Object` happen to be pointers, `T` is a pointer type, so `var objType T` result in `objType` being `nil` and `c.GroupVersionKindFor(objType)` doesn't like that.

internal/server/common_test.go
krancour: Unstructured objects typically do contain type metadata and it hadn't actually been my intent to add it if it _were_ missing. I'd considered it garbage in garbage out.

Structured objects, on the other hand are _typically_ missing type metadata, so there's little choice but to account for it.

PR #5071, opened 2025-09-18
chore: move off older versions of github.com/google/go-github

krancour:
Noticed we were using three different versions of github.com/google/go-github.

This switches us to uniformly using v74.


rpelczar (approved):

thomastaylor312 (approved):

krancour:
I think after #5079, this is already addressed.

PR #5072, opened 2025-09-18
fix: update `sources` instead of `source` when Application specifies both

nitishfy:
Closes https://github.com/akuity/kargo/issues/4856


internal/promotion/runner/builtin/argocd_updater.go
```diff
@@ -477,12 +477,15 @@ func (a *argocdUpdater) syncApplication(
...
-	if app.Spec.Source != nil {
+	// Set target revisions for sync operation.
+	// If sources are specified, Argo CD ignores source.
+	if len(app.Spec.Sources) > 0 {
+		for _, source := range app.Spec.Sources {
+			app.Operation.Sync.Revisions = append(app.Operation.Sync.Revisions, source.TargetRevision)
+		}
+	} else if app.Spec.Source != nil {
 		app.Operation.Sync.Revisions = []string{app.Spec.Source.TargetRevision}
 	}
```
krancour: @nitishfy I'm starting to question whether we actually need to set `app.Operation.Sync.Revisions` at all.

Docs on that field state this:

https://github.com/argoproj/argo-cd/blob/fbcaf35ab2c26568c7a384be43260861cd55360a/pkg/apis/application/v1alpha1/types.go#L1369-L1371

krancour: Commenting here because I cannot comment on unchanged lines farther down...

There's a conditional on line -539 that needs to change.

pkg/promotion/runner/builtin/argocd_updater_test.go
```diff
@@ -1081,29 +1078,100 @@ func Test_argoCDUpdater_syncApplication(t *testing.T) {
...
+				require.Equal(t, "old-rev-source", patched.Spec.Source.TargetRevision)
+			},
+		},
 	}
 
-	stepCtx := &promotion.StepContext{
-		Freight: kargoapi.FreightCollection{},
-	}
-	// Tamper with the freight collection ID for testing purposes
-	stepCtx.Freight.ID = "fake-freight-collection-id"
```
krancour, line 1099: You moved this up much higher and I'd prefer it remained here. It's a good practice to initialize things closer to where they are first used.

When you moved it, you also removed the important comment about the deliberate massaging of the ID.

internal/promotion/runner/builtin/argocd_updater_test.go
```diff
@@ -1081,29 +1078,100 @@ func Test_argoCDUpdater_syncApplication(t *testing.T) {
...
+				}
+			}
+
+			err := testCase.runner.syncApplication(
+				context.Background(),
+				stepCtx,
+				testCase.app,
+				testCase.desiredSources,
 			)
+			testCase.assertions(t, err, captured)
```
krancour: You don't need to go to these lengths to get a reference to the App. Since the `syncApplication()` method directly modifies the App that's passed to it, you only need to do this:

```suggestion
			err := testCase.runner.syncApplication(
				context.Background(),
				stepCtx,
				testCase.app,
				testCase.desiredSources,
			)
			testCase.assertions(t, err, testCase.app)
```

internal/promotion/runner/builtin/argocd_updater_test.go
```diff
@@ -1021,12 +1021,16 @@ func Test_argoCDUpdater_mustPerformUpdate(t *testing.T) {
...
+		Freight: kargoapi.FreightCollection{ID: "fake-freight-collection-id"},
+	}
+
 	testCases := []struct {
 		name           string
 		runner         *argocdUpdater
 		app            *argocd.Application
 		desiredSources argocd.ApplicationSources
-		assertions     func(*testing.T, error)
+		assertions     func(*testing.T, error, *argocd.Application)
```
krancour: It would be more conventional for an `error` to be the last argument.

```suggestion
		assertions     func(*testing.T, *argocd.Application, error)
```

pkg/promotion/runner/builtin/argocd_updater_test.go
nitishfy: resolved.

internal/promotion/runner/builtin/argocd_updater.go
nitishfy: hmm, makes sense. I have removed that check.

pkg/promotion/runner/builtin/argocd_updater.go
```diff
@@ -533,12 +527,21 @@ func (a *argocdUpdater) syncApplication(
...
-	if app.Spec.Source != nil {
+	switch {
+	case len(app.Spec.Sources) == 1:
+		message += " to " + app.Spec.Sources[0].TargetRevision
+	case len(app.Spec.Sources) > 1:
+		message += fmt.Sprintf(" to %d sources", len(app.Spec.Sources))
+	case app.Spec.Source != nil:
 		message += " to " + app.Spec.Source.TargetRevision
 	}
+
```
nitishfy: @krancour what are your thoughts about it? On a side note, i've added unit tests too for this specific piece of code too for a better understanding.

krancour: This is good. 👍

pkg/promotion/runner/builtin/argocd_updater_test.go
```diff
@@ -1081,29 +1080,190 @@ func Test_argoCDUpdater_syncApplication(t *testing.T) {
...
+			switch {
+			case len(tc.app.Spec.Sources) == 1:
+				message += " to " + tc.app.Spec.Sources[0].TargetRevision
+			case len(tc.app.Spec.Sources) > 1:
+				message += fmt.Sprintf(" to %d sources", len(tc.app.Spec.Sources))
+			case tc.app.Spec.Source != nil:
+				message += " to " + tc.app.Spec.Source.TargetRevision
+			}
+
+			require.Equal(t, tc.expected, message)
```
krancour: This doesn't make any sense. Why are you reproducing this logic here instead of invoking the logic you're actually trying to test? _This doesn't actually test anything._

nitishfy: please take a look now.

pkg/promotion/runner/builtin/argocd_updater.go
```diff
@@ -558,6 +546,25 @@ func (a *argocdUpdater) syncApplication(
 	return nil
 }
 
+// constructMessage construct a human-friendly message about what we're syncing to.
+//   - If there's exactly one source, we include its TargetRevision.
+//   - If there are multiple sources, we just report the count to avoid a noisy log.
+//   - If only the single Source field is set, we log its TargetRevision.
+//     Full details are always available on the Application object itself.
+func constructMessage(app *argocd.Application) string {
```
krancour: Because of the large amount of code in this package, it would be preferable that this were a method of the `*argocdUpdater` type so that its relationship to other functionality in that same component is more obvious, so it isn't ever mistaken for something relevant to a different runner.

Also on account of the large volume of code in this package, please see about tweaking the name and the docstring for clarity and context. The current description still manages to be vague about what this is really used for.

nitishfy: Please take a look now.

krancour: Looks great. Thank you!

PR #5086, opened 2025-09-22
refactor: enable verification log stream

cokeeffe:
When running a Verification we would like to be able to see the logs while the job is running, not just when the job terminates as the time between can be long and visibility into what is happening is not available until after the fact. Here we are remove the checks for the job pod being in a terminated state.


krancour:
@hiddeco and/or @gdsoumya do either of you have a recollection of why we were requiring analysis to be complete? If there was an internal reason for it, please DM me.

krancour:
@gdsoumya and I spoke offline and there are internal reasons we did this, so unfortunately, we cannot accept this change at this time.

cokeeffe:
@krancour could you share the reasons for this? Could it become an optional setting, for example:

```yaml
api:
  rollouts:
    integrationEnabled: true
    logs:
      enabled: true
      waitUntilComplete: true # wait until the job completes before streaming logs
      urlTemplate: https://logs.kargo.example.com/${{project}}/${{analysisRun}}/${{jobName}}/${{container}}
      tokenSecret:
        name: kargo-logs-token
        key: token
      httpHeaders:
        Authentication: "Bearer ${{ token }}"
```

jessesuen:
The reason is that the approach would need to be generalized to work in both the OSS Kargo centralized deployment model, as well as the distributed agent-based deployment model supported in our platform. Live logs are something we definitely would like to do, but it needs to be done in a way that supports both architectures.

PR #5088, opened 2025-09-22
refactor: make task output propagation a promotion step capability

krancour:
@hiddeco, as we discussed, this is a follow-up to #5079 that eliminates the need for a wrapper around the compose-output step that implements the `TaskLevelOutputStepRunner` interface. By doing that we will no longer need to instantiate a runner to know it has this capability.

Please feel free to tweak the wording on the capability's docstring.

### DIFF
--- a/internal/promotion/simple_engine.go
+++ b/internal/promotion/simple_engine.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"slices"
 
 	gocache "github.com/patrickmn/go-cache"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


internal/promotion/simple_engine.go
```diff
@@ -274,15 +275,18 @@ func (e *simpleEngine) shouldSkipStep(
...
+	stepReg promotion.StepRunnerRegistration,
 	result promotion.StepResult,
 ) {
 	// Update the state with the output of the step.
 	promoCtx.State[step.Alias] = result.Output
 
 	// If the step instructs that the output should be propagated to the
 	// task namespace, do so.
-	if p, ok := runner.(promotion.TaskLevelOutputStepRunner); ok && p.TaskLevelOutput() {
+	if stepReg.Metadata != nil && slices.Contains(
```
hiddeco: IIRC, the registration function will always ensure `Metadata` isn't `nil`?

If that is correct, I wonder if we should drop the pointer as that would make our lives a bit easier.

krancour: Good thought. At some point while I'd been working on this, the pointer made more sense, and where I landed in the end, it no longer made as much sense, but I failed to notice.

Fixed in 886bd9194db6bbf02a597a66368014be2227758c

Issue #5087, opened 2025-09-22
UI: Project Favorites


jessesuen:
# Proposed Feature

The Kargo UI should have the ability to favorite one or more projects so that they can get a view of the projects that is a subset of all the projects.

# Motivation

The list of projects might get too large, and the user only cares about a handful of them.

PR #5090, opened 2025-09-23
feat(ui): star projects

Marvin9:
fixes https://github.com/akuity/kargo/issues/5087

https://github.com/user-attachments/assets/71d41aca-f510-4549-8b96-cbfadb0c5232


ui/src/features/project/list/use-star-projects.tsx
```diff
@@ -0,0 +1,18 @@
+import { useLocalStorage } from '@ui/utils/use-local-storage';
+
+export const useStarProjects = () => {
+  const [starred, setStarred] = useLocalStorage('starred-projects', [] as string[]) as [
+    string[],
+    (ids: string[]) => void
+  ];
```
jmalisz-akuity: Please fix the `useLocalStorage` so we can avoid such weird type assignments:
```
export const useLocalStorage = <TValue>(key: string, initialValue?: TValue) => {
  const [storedValue, _setStoredValue] = useState(() => {
    try {
      const item = window.localStorage.getItem(key);
      return item ? JSON.parse(item) : initialValue;
    } catch (_) {
      return initialValue;
    }
  });

  const setStoredValue: typeof _setStoredValue = (storedValue) => {
    _setStoredValue(storedValue);

    if (!storedValue) {
      window.localStorage.removeItem(key);
      return;
    }
    window.localStorage.setItem(key, JSON.stringify(storedValue));
  };

  return [storedValue, setStoredValue] as const;
};
```

ui/src/features/project/list/use-star-projects.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  const toggleStar = (projectId: string) => {
+    if (starred?.includes(projectId)) {
+      setStarred(starred.filter((id) => id !== projectId));
+      return;
+    }
+    setStarred([...(starred || []), projectId]);
+  };
```
jmalisz-akuity, line 12: I would create `Set` from the local storage value and use it there. Performance wise it should be better, but for this use case it probably won't matter. So I'm okay with leaving it as it is.

Marvin9: Yeah I thought of that but TBH it doesn't really make difference unless the operation is in scale of millions and happening very frequently

Marvin9:
> IMO this isn't that useful solution. Starring the project should help when someone has a lot of projects available, but cares only about a few of them. Right now in that case it's likely that those projects won't be available, because filtering is done on FE only on 10 items page.
> 
> The filtering should be moved to BE. We can send the list of starred projects and only return those. If we don't want to change BE, I guess we can load all of the projects by removing pagination. But I think it's a bad idea.

Good point. I have included a [fix](https://github.com/akuity/kargo/pull/5090/commits/cb6d00d5933a8201f254072d0a25987c4bfd2ef7) but I think @krancour might have some ideas as well

krancour:
@Marvin9 I'm happy to do whatever could be done on the back end that would help you.

We have no realistic way of tracking favorites on the back end because we have no place to store user-specific data, so favorites will need to be _remembered_ on the front end, i.e. in local storage.

From there, I think there are two choices:

1. When building the Project list narrowed to favorites, make individual get calls for each favorite Project. This isn't entirely ideal, but it's also probably not as bad as it sounds -- at least it's not very bad unless the number of starred projects is very high. A dozen requests? No big deal. This could be done without any back end changes.

1. I could add new, optional criteria to a list Projects request that narrows the selection to specific Projects that you specify when making the request. This would be relatively easy, but would require a little bit of thoughtfulness in how we approach because there's already an optional `filter` field that's used as part of search. (_Possibly_ instead of a new field, the capabilities of the filter field could be expanded to facilitate this. I'm not really decided yet on implementation. I'm only citing this as one possible approach.)

Let me know what would work best for you and I can coordinate getting it done or give you guidance on how to do it if you want to modify the endpoint yourself.

Marvin9:
Thanks @krancour. I like option 2. But as you might have noticed or maybe not, the UI store in local the unique UID of projects instead of name because thats really unique than names (edge cases but why not). So I added new field in API request that accepts list of UIDs (for UI it is used as list of starred projects to search) and return only those.

You can review backend code..

krancour:
@Marvin9, I'm really sorry. When I responded earlier, I hadn't realized that you already implemented the back end changes. If I had, I'd have reviewed sooner.

Just looked and I think it looks fine.

Good thinking on the UID. Project names, like all k8s resource names are immutable, so I'd have probably used names and it probably would have been fine, but there is an edge case where a Project is _deleted_ and then a _new_ Project is created with the same name, and I suppose, under those circumstances, you hadn't wanted the new Project automatically starred just on account of having reused a name? Seems like the right call.

LGTM!

ui/src/features/project/list/projects-list.tsx
```diff
@@ -45,24 +55,41 @@ export const ProjectsList = () => {
...
-        ))}
+        {data.projects
+          ?.filter((p) => (starredProjectsView ? starred.includes(p?.metadata?.uid || '') : true))
+          .map((proj) => (
+            <ProjectItem
+              key={proj?.metadata?.name}
+              project={proj}
+              starred={starred.includes(proj?.metadata?.uid || '')}
+              onToggleStar={(id) => toggleStar(id)}
+            />
```
jmalisz-akuity: Please remove front-end filtering

jmalisz-akuity (commented):
Alright, then what's left is to remove front-end filtering and we will be good to go

Marvin9:
> I just noticed you tried to fix the problem with be pagination, but it still doesn't work.

You have to run it with backend changes in this PR in order to make it work

> Can we also update the UI a bit? The star button is too big in my opinion. Can we do something like this?

Alright

Marvin9:
Done @rpelczar @jmalisz-akuity

Issue #5045, opened 2025-09-12
Should Stage health evaluated be a ERROR in the kargo-controller?


krancour:
### Discussed in https://github.com/akuity/kargo/discussions/5024

<div type='discussions-op-text'>

<sup>Originally posted by **nissessenap** September 10, 2025</sup>
Hi

Once again, thanks for a great project.

I'm not enabling the ArgoCD integration with Kargo, mostly due to I have environments that are air-gapped, so I can't reach my ArgoCD instances. So instead I just see Kargo as a GitOps tools.

The kargo-controller shows me lots of errors like this:
```
time="2025-09-10T08:09:13Z" level=error msg="Reconciler error" Stage="{customer-exports-staging customer-exports}" controller=stage controllerGroup=kargo.akuity.io controllerKind=Stage error="Stage health evaluated to Unknown" name=customer-exports-staging namespace=customer-exports reconcileID="\"ad800301-2af5-45d0-9eaa-22a51483ff61\""
```

I assume it is since I don't use the argocd-update feature.
https://docs.kargo.io/user-guide/reference-docs/promotion-steps/argocd-update/#health-checks

In my this should be an info, or even if the ArgoCD feature isn't enabled, it should only be debug.
Happy to create a PR around this if you agree.
</div>

---

More context:

* https://github.com/akuity/kargo/discussions/5024#discussioncomment-14386402
* https://github.com/akuity/kargo/pull/4871
* https://github.com/akuity/kargo/pull/4674

@fuskovic, I'm assigning to you because you had looked at this code quite a bit recently, and I think you understand the context, but lmk if you want to discuss at all.

krancour:
I weighted this statement in #5024 too heavily:

> I assume it is since I don't use the argocd-update feature.

Which led me to misspeak:

> It's been the case for some time that if Kargo doesn't know how to check on the Stage's health in some way, it treats the health as Unkown. Without Argo CD in the picture, there really is no way to know.

This is actually not true. Had I double-checked the code, I'd have been reminded that Stages without any health checks defined (i.e. none created as a side effect of the `argocd-update` step) actually default to Healthy. (They must, or else verification is blocked indefinitely.)

Since this whole issue was based on a false assumption, I'm closing this as well as #5093.

I will re-open #5024 and see if we can't get to the bottom of what the real problem was.

PR #5093, opened 2025-09-23
fix: return nil err when there are no healthchecks

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5045

### DIFF
--- a/pkg/controller/stages/regular_stages_test.go
+++ b/pkg/controller/stages/regular_stages_test.go
@@ -1907,7 +1907,7 @@ func TestRegularStageReconciler_assessHealth(t *testing.T) {
 				require.NotNil(t, healthyCond)
 				assert.Equal(t, metav1.ConditionUnknown, healthyCond.Status)
 				assert.Equal(t, "NoHealthChecksDefined", healthyCond.Reason)
-				assert.Contains(t, healthyCond.Message, "No health checks defined for last Promotion")
+				assert.Contains(t, healthyCond.Message, "No health checks defined")
 			},
 		},
 		{


pkg/controller/stages/regular_stages.go
```diff
@@ -458,7 +458,13 @@ func (r *RegularStageReconciler) reconcile(
 					//     the Stage's conditions to reflect that we're still trying to
 					//     reconcile, with subsequent attempts observing a progressive
 					//     backoff. This can be accomplished by returning an error.
-					if lastPromo := status.LastPromotion; lastPromo == nil ||
+					lastPromo := status.LastPromotion
+					if lastPromo != nil && lastPromo.Status != nil && len(lastPromo.Status.HealthChecks) == 0 {
+						// if there are no health checks, then we cannot assess health.
```
krancour: I'd rather like to (additionally?) see this conveyed through some edits to the longer comments above.

We should probably revise it a little to be more general -- that when health is unknown, but _knowable_ it's an error. If it's unknown and unknowable (scenario one OR this new scenario), then it's not an error.

pkg/controller/stages/regular_stages.go
```diff
@@ -458,7 +458,13 @@ func (r *RegularStageReconciler) reconcile(
 					//     the Stage's conditions to reflect that we're still trying to
 					//     reconcile, with subsequent attempts observing a progressive
 					//     backoff. This can be accomplished by returning an error.
-					if lastPromo := status.LastPromotion; lastPromo == nil ||
+					lastPromo := status.LastPromotion
+					if lastPromo != nil && lastPromo.Status != nil && len(lastPromo.Status.HealthChecks) == 0 {
```
krancour: This logic is wrong.

krancour:
```suggestion
					if lastPromo == nil || lastPromo.Status == nil || len(lastPromo.Status.HealthChecks) == 0 {
```

pkg/controller/stages/regular_stages.go
```diff
@@ -437,38 +437,23 @@ func (r *RegularStageReconciler) reconcile(
...
+						// If there was no last promotion or there was but it had no health
+						// checks, then health is not knowable and therefore we do not treat
+						// it as an error.
+						return status, nil
+					}
+
+					if !lp.Status.Phase.IsTerminal() || lp.Status.Phase == kargoapi.PromotionPhaseSucceeded {
+						// If there was a last promotion with health checks, then health is knowable but unknown
+						// due to an unexpected transient error. Therefore we treat it as an error to trigger a requeue
+						// with backoff.
```
krancour: Small tweaks, but overall, I like the way you're documenting this scenario now and same for the above.

```suggestion
						// If there was a last promotion with health checks, then health is
						// knowable, but unknown due to an unexpected, possibly transient
						// condition. Treating this scenario as an error to will requeue the
						// Stage with progressive backoff for repeated failures, as with any
						// other error.
```

pkg/controller/stages/regular_stages.go
```diff
@@ -437,38 +437,23 @@ func (r *RegularStageReconciler) reconcile(
...
 						return status, errors.New("Stage health evaluated to Unknown") // nolint: staticcheck
 					}
 				}
-				// Health assessment was definitive OR scenario 1: Stage health is
-				// unknown specifically because the last Promotion did not succeed.
+				// Health assessment was definitive OR stage health is unknown specifically because the last
+				// Promotion did not succeed.
```
krancour:
```suggestion
				// Health assessment was definitive OR Stage health is unknowable.
```

pkg/controller/stages/regular_stages.go
krancour: Disregard....

The more I look at this PR, the more I think we are (were already) doing this in the wrong place.

The existing `assessHealth()` method on the regular stages reconciler already is looking at lot of the things we're checking on here in great detail and it seems it would be a more appropriate spot to refine the logic around when health is _just_ unknown vs unknown _and_ an error because it was knowable.

imho, we should put _this_ section of code back to how it was prior to #4674. (Just this section of code; only. #4674 should not be reverted because there's other good stuff in there.)

pkg/controller/stages/regular_stages.go
```diff
@@ -828,11 +797,27 @@ func (r *RegularStageReconciler) assessHealth(ctx context.Context, stage *kargoa
 			Status: kargoapi.HealthStateUnknown,
 			Issues: []string{"Cannot assess health because last Promotion did not succeed"},
 		}
-		return newStatus
+		return newStatus, errors.New("cannot assess health because last Promotion did not succeed")
```
krancour: I think health is unknowable in this case.

pkg/controller/stages/regular_stages.go
```diff
@@ -828,11 +797,27 @@ func (r *RegularStageReconciler) assessHealth(ctx context.Context, stage *kargoa
...
+			Reason:             "NoHealthChecksDefined",
+			Message:            "No health checks defined for last Promotion",
+			ObservedGeneration: stage.Generation,
+		})
+		newStatus.Health = &kargoapi.Health{
+			Status: kargoapi.HealthStateUnknown,
+			Issues: []string{"Cannot assess health because there are no health checks defined"},
+		}
+		return newStatus, nil
+	}
```
krancour, line 819: 👍

krancour:
@fuskovic apart from https://github.com/akuity/kargo/pull/5093#discussion_r2376982374, please just give this an e2e test-drive to make sure you don't get stuck reconciling if:

* ~~Stage hasn't had any Promotions yet~~ F.H.
* ~~A Promotion is currently in-progress~~ F.H.
* ~~Last Promotion failed~~ F.H.
* ~~Last Promotion succeeded but didn't have any health checks (you can do this by not involving Argo CD in your promotion process)~~ F.H.

After that sanity check, this looks g2g.

pkg/controller/stages/regular_stages.go
```diff
@@ -828,11 +797,27 @@ func (r *RegularStageReconciler) assessHealth(ctx context.Context, stage *kargoa
...
 
 	// Compose the health check criteria.
 	healthChecks := lastPromo.Status.HealthChecks
+	if len(healthChecks) == 0 {
+		logger.Debug("Last promotion has no health checks: no health checks to perform")
+		conditions.Set(&newStatus, &metav1.Condition{
+			Type:               kargoapi.ConditionTypeHealthy,
+			Status:             metav1.ConditionUnknown,
+			Reason:             "NoHealthChecksDefined",
+			Message:            "No health checks defined for last Promotion",
```
krancour: The "for last Promotion" part doesn't sound accurate. The health checks are a byproduct of a Promotion that are executed periodically against the Stage. "by last Promotion" would be more accurate, but seems unnecessary. I think this is a case where too much information will confuse people more than if we were less specific.

```suggestion
			Message:            "No health checks defined",
```

krancour:
I've discovered a new bug this introduces. If health assesses to Unknown, it will block verification indefinitely.

I'm thinking about the best way to deal with this...

krancour:
Closing this. Explanation here: https://github.com/akuity/kargo/issues/5045#issuecomment-3362114176

Issue #5044, opened 2025-09-12
return 400 instead of 500 when an attempt is made to promote to a control flow stage


krancour:
### Discussed in https://github.com/akuity/kargo/discussions/5040

<div type='discussions-op-text'>

<sup>Originally posted by **TarlyJQ** September 12, 2025</sup>
I have encapsulated the Rustful API using the controller runtime myself, but when calling it, there will be an error
admission webhook \"promotion.kargo.akuity.io\" denied the request: Stage \"dev\" in namespace \"multi-frontend-app\" defines no promotion steps
<img width="1493" height="692" alt="image" src="https://github.com/user-attachments/assets/9d13c531-c045-4a2a-bf4b-9099eacb660b" />
</div>

krancour:
See [this comment](https://github.com/akuity/kargo/discussions/5040#discussioncomment-14385901) for details of cause and possible solutions.

krancour:
This depends on #5126 being straightened out first.

PR #5096, opened 2025-09-23
fix: return 400 when attempting to promote to control flow stage

fuskovic:
Closes:  https://github.com/akuity/kargo/issues/5044

### DIFF
--- a/internal/webhook/kubernetes/promotion/webhook.go
+++ b/internal/webhook/kubernetes/promotion/webhook.go
@@ -191,14 +191,6 @@ func (w *webhook) Default(ctx context.Context, obj runtime.Object) error {
 			promo.Namespace,
 		)
 	}
-	if len(promo.Spec.Steps) == 0 {
-		// nolint:staticcheck
-		return fmt.Errorf(
-			"Stage %q in namespace %q defines no promotion steps",
-			promo.Spec.Stage,
-			promo.Namespace,
-		)
-	}
 
 	// Make sure the Promotion has the same shard as the Stage
 	if stage.Spec.Shard != "" {
@@ -254,6 +246,20 @@ func (w *webhook) ValidateCreate(
 	})
 	if err != nil {
 		return nil, apierrors.NewInternalError(fmt.Errorf("get stage: %w", err))
+	}
+
+	if pt := stage.Spec.PromotionTemplate; pt == nil || len(pt.Spec.Steps) == 0 {
+		return nil, apierrors.NewInvalid(
+			promotionGroupKind,
+			promo.Name,
+			field.ErrorList{
+				field.Invalid(
+					field.NewPath("spec", "stage"),
+					promo.Spec.Stage,
+					"Stage has no promotion steps defined",
+				),
+			},
+		)
 	}
 
 	freight, err := w.getFreightFn(ctx, w.client, types.NamespacedName{
--- a/internal/controller/stages/regular_stages.go
+++ b/internal/controller/stages/regular_stages.go
@@ -436,7 +436,41 @@ func (r *RegularStageReconciler) reconcile(
 		{
 			name: "assessing health",
 			reconcile: func() (kargoapi.StageStatus, error) {
-				return r.assessHealth(ctx, stage)
+				status := r.assessHealth(ctx, stage)
+				if status.Health != nil && status.Health.Status == kargoapi.HealthStateUnknown {
+					// If Stage health has evaluated to Unknown, there are two specific
+					// scenarios between which we must distinguish and handle differently
+					// from one another:
+					//
+					//  1. If Stage health is unknown specifically because the last
+					//     Promotion did not succeed, we know we cannot obtain a more
+					//     definitive assessment of Stage health until a new Promotion has
+					//     restored the Stage to a consistent state by executing to
+					//     completion. In such a case, we're DONE reconciling the Stage,
+					//     for now. We would like the Stage's conditions to reflect that
+					//     and for the next reconciliation attempt to occur after the
+					//     usual interval. This can be accomplished by returning a nil
+					//     error.
+					//
+					//  2. If Stage health is unknown for any other reason, we MAY
+					//     possibly obtain a more definitive assessment simply by
+					//     re-attempting reconciliation. In such a case, we would like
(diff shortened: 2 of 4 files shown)


internal/webhook/kubernetes/promotion/webhook.go
```diff
@@ -256,6 +248,20 @@ func (w *webhook) ValidateCreate(
 		return nil, apierrors.NewInternalError(fmt.Errorf("get stage: %w", err))
 	}
 
+	if pt := stage.Spec.PromotionTemplate; pt == nil || len(pt.Spec.Steps) == 0 {
+		return nil, apierrors.NewInvalid(
```
thomastaylor312, line 252: So in the issue, we talked about returning 400. `NewInvalid` returns 422, but with a nicer error message. This might be nit-y, but I think 422 is generally more for encoding issues, though I could see an argument for it applying here. We should probably decide which is the proper HTTP code we want to return here.

Switching it out is easy though because the `.Code` field is publicly accessible and we can change the code after constructing the nice error with invalid. @krancour did you explicitly want a 400 error or just a 4xx error of any kind?

internal/webhook/kubernetes/promotion/webhook.go
```diff
@@ -191,14 +191,6 @@ func (w *webhook) Default(ctx context.Context, obj runtime.Object) error {
 			promo.Namespace,
 		)
 	}
-	if len(promo.Spec.Steps) == 0 {
```
thomastaylor312, line 194: So I don't think we need to dismiss this so quickly. Looks like we can provide the same error you added below by returning it here. If you take a look at [the code that is calling this handler](https://github.com/kubernetes-sigs/controller-runtime/blob/6422ed031ac07ae85c8bd234657e77ddf5cace6e/pkg/webhook/admission/defaulter_custom.go#L113), you'll see it asserts on `APIStatus` which `StatusError` returned from all of the `apierrors` methods implement

thomastaylor312:
Yep, I'm pretty sure this is in our connect API layer (see [docs](https://connectrpc.com/docs/go/errors/)). We aren't mapping the error returned from `kubeClient.Create` that is used in https://github.com/akuity/kargo/blob/dd4c1b91c2008e37f516c0cc90b819f988f11cbe/pkg/server/promote_to_stage_v1alpha1.go#L125-L127

So I'm thinking we'll need to wrap the error on a create failure with something like this: 

```go
connect.NewError(connect.CodeInvalidArgument, err)
```

Then the connect middleware should wrap that to a 400

krancour:
Even if connect can be implicated in this, I still tend to think there's an issue with the mutating webhook. Those wouldn't ordinarily be responsible for returning things like 400s, so I am pretty sure a 500 is what we get for any error that we don't more carefully classify.

If the API endpoint that attempted the promotion create gets a 500 from its interaction with the Kubernetese API server, that gets passed on.

What really needs to be done to make sure we've got the webhook right is we need to test it using only kubesctl, without our API server in the picture. Once we know we have _that_ part right, we can figure out what's up with connect.

__Edit:__ Whoops. I missed this:

> returns a 403 by default if we just return a plain error

Indeed, that's what I get. It seems an odd choice for it to be a 403.

We could probably stand to improve this.

krancour:
Looking at this more, the promotion endpoints and defaulting webhook both have some peculiar behaviors that I think we may not have intended and I want to get a better handle on them before we go too deep into what the right errors types are for different circumstances.

@hiddeco do you happen to know why our promotion API endpoints, like "promote to stage" build promotions (copy steps as-they-are from the Stage), but don't inflate tasks, while the defaulting webhook doesn't build promotions, but _does_ inflate tasks?

In my mind, the ideal is that whether you interact with Kargo through our API server or through `kubectl` alone, things should work roughly the same. i.e. Either path requires you only to specify a Stage and piece of Freight and then the Promotion is created through _both_ copying steps from the Stage _and_ expanding any tasks references therein.

fuskovic:
@thomastaylor312 @krancour sorry for the confusion but I did not intend on this getting any reviews yet. I should have opened in draft initially. Oversight on my part. 

Since the default webhook never wrapped any returned errors with `apierrors` I assumed it would override any status codes to internal (nor was that it's job; validation I mean) based on this language in the issue:

> As for the 500 which seems as if it should be a 400, the error is being issued from a Kubernetes [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) (also known as a "defaulting webhook"). The role of this component is not to validate inbound requests, but to apply any defaults or other mutations that are required in order for the request to be valid when it reaches the validating webhook, whose job it is (as implied by its name) is validating the inbound request. Kubernetes always invokes webhooks in this order -- defaulting webhooks and then validating webhooks.

So my rationale was to move the check to where we actually did perform validation ( in the validating webhook ).

I have not manually tested this code yet.

Also if connect is involved at all, it's not obvious to me how. 

My understanding is the request is going from:

Default Webhook -> Validating Webhook -> API server

Am I missing something?

fuskovic:
> Yep, I'm pretty sure this is in our connect API layer (see [docs](https://connectrpc.com/docs/go/errors/)). We aren't mapping the error returned from `kubeClient.Create` that is used in
> 
> https://github.com/akuity/kargo/blob/dd4c1b91c2008e37f516c0cc90b819f988f11cbe/pkg/server/promote_to_stage_v1alpha1.go#L125-L127
> 
> So I'm thinking we'll need to wrap the error on a create failure with something like this:
> 
> ```go
> connect.NewError(connect.CodeInvalidArgument, err)
> ```
> 
> Then the connect middleware should wrap that to a 400

I'm not seeing how it's coming from here. Wouldn't `create promotion:` substring be present in the error message in the issue if this were true?

krancour:
> I'm not seeing how it's coming from here...

There's layers of stuff here. The webhook, on its own, is behaving oddly, as mentioned, not _only_ with respect to errors, but also with what it does and does not do. We need to straighten that our first before worrying about how errors from the webhook are (or aren't) being wrapped / translated by the API server.

We really should start with cutting the API server out of the equation and first making sure we're satisfied with success and failure cases when creating a Promotion using kubectl.

krancour:
Closing this because #5126 should really be tackled first, at which point, the merge conflicts in here will probably be bad enough to warrant a fresh start.

PR #5106, opened 2025-09-24
feat(http): add bodytext field to response object to always return response body as string

yiran0427:


pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -289,25 +289,33 @@ func (h *httpRequester) buildExprEnv(
...
+			// Unmarshal into map[string]any or []any
+			switch parsedBody.(type) {
+			case map[string]any, []any:
+				env["response"].(map[string]any)["body"] = parsedBody // nolint: forcetypeassert
+			default:
+				logging.LoggerFromContext(ctx).Debug(
+					"Unexpected JSON type, continuing with bodyText only",
+					"type", fmt.Sprintf("%T", parsedBody),
+				)
+			}
```
krancour, line 318: This doesn't feel right. To have gotten to this chunk of code, the conditional on line +300 evaluated to true. The content type of the response was explicitly JSON _or_ the response has already been determined to be valid JSON (regardless of what the content type said). i.e. If we're in this block of code, it's _expected_ that the body will unmarshal correctly. If it doesn't we should be sticking with the original behavior of returning an error.

yiran0427: oh i see. is it better if i add another case for `contentType == 'application/json'`. then return `bodyText: string(bodyBytes)` and empty body?

krancour: I don't think you need to do anything except revert the changes to these lines.

I _think_ the only change needed to this file is the one on line +296.

PR #5109, opened 2025-09-25
fix(api-server): write debug message instead of returning error when client hangs up on a watch

### LINKED ISSUE #5101: strange api server errors since logging changes
These errors keep showing up in the API server logs since #4999. (Going back one commit prior, these don't show up.)

Given that #4999 was logging related, I tend to wonder if this isn't a new _problem_, but rather an existing one for which we were not previously seeing logs.

cc @thomastaylor312 in case you have any insight.

```
2025-09-24T00:19:48Z	ERROR	option/log.go:88	finished streaming call	{"connect.service": "akuity.io.kargo.service.v1alpha1.KargoService", "connect.method": "WatchStages", "connect.start_time": "2025-09-24T00:19:47Z", "connect.duration": "1.046606062s", "connect.code": "internal", "error": "internal: context canceled"}
github.com/akuity/kargo/internal/server/option.(*logInterceptor).WrapStreamingHandler.func1
	/Users/krancour/Code/akuity/kargo/internal/server/option/log.go:88
connectrpc.com/connect.(*Handler).ServeHTTP
	/Users/krancour/go/pkg/mod/connectrpc.com/connect@v1.18.1/handler.go:237
github.com/akuity/kargo/api/service/v1alpha1/svcv1alpha1connect.NewKargoServiceHandler.func1
	/Users/krancour/Code/akuity/kargo/api/service/v1alpha1/svcv1alpha1connect/service.connect.go:2020
net/http.HandlerFunc.ServeHTTP
	/opt/homebrew/Cellar/go/1.25.1/libexec/src/net/http/server.go:2322
net/http.(*ServeMux).ServeHTTP
	/opt/homebrew/Cellar/go/1.25.1/libexec/src/net/http/server.go:2861
golang.org/x/net/http2/h2c.h2cHandler.ServeHTTP
	/Users/krancour/go/pkg/mod/golang.org/x/net@v0.43.0/http2/h2c/h2c.go:125
github.com/akuity/kargo/internal/server.(*server).Serve.(*Cors).Handler.func2
	/Users/krancour/go/pkg/mod/github.com/rs/cors@v1.11.1/cors.go:289
net/http.HandlerFunc.ServeHTTP
	/opt/homebrew/Cellar/go/1.25.1/libexec/src/net/http/server.go:2322
net/http.serverHandler.ServeHTTP
	/opt/homebrew/Cellar/go/1.25.1/libexec/src/net/http/server.go:3340
net/http.(*conn).serve
	/opt/homebrew/Cellar/go/1.25.1/libexec/src/net/http/server.go:2109
```

krancour:
Fixes #5101 

I can't put my finger on _why_ we're all of the sudden seeing error messages we weren't before, but they're occurring anytime the client hangs up on a watch. As far as I can tell, everything is still working exactly as it ought to, so I think it's sensible to log a debug message when the client hangs up instead of returning an error. After all -- we expect clients to eventually hang up. This is normal.

Only reason I am even bothering to write a debug message instead of just ignoring the context cancellation entirely is on the off chance that there _was_ some legitimate problem that's being overlooked here, this will provide our future selves with a clue as to the cause.

### DIFF
--- a/pkg/server/get_analysisrun_logs_v1alpha1.go
+++ b/pkg/server/get_analysisrun_logs_v1alpha1.go
@@ -21,6 +21,7 @@ import (
 	"github.com/akuity/kargo/pkg/api/stubs/rollouts"
 	libEncoding "github.com/akuity/kargo/pkg/encoding"
 	"github.com/akuity/kargo/pkg/expressions"
+	"github.com/akuity/kargo/pkg/logging"
 	"github.com/akuity/kargo/pkg/server/user"
 )
 
@@ -172,8 +173,9 @@ func (s *server) GetAnalysisRunLogs(
 				return fmt.Errorf("error sending log chunk: %w", err)
 			}
 		case <-ctx.Done():
-			// Context canceled or timed out
-			return ctx.Err()
+			logger := logging.LoggerFromContext(ctx)
+			logger.Debug(ctx.Err().Error())
+			return nil
 		}
 	}
 }
--- a/pkg/server/watch_cluster_config_v1alpha1.go
+++ b/pkg/server/watch_cluster_config_v1alpha1.go
@@ -14,6 +14,7 @@ import (
 	svcv1alpha1 "github.com/akuity/kargo/api/service/v1alpha1"
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 	"github.com/akuity/kargo/pkg/api"
+	"github.com/akuity/kargo/pkg/logging"
 )
 
 func (s *server) WatchClusterConfig(
@@ -38,7 +39,9 @@ func (s *server) WatchClusterConfig(
 	for {
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
+			logger := logging.LoggerFromContext(ctx)
+			logger.Debug(ctx.Err().Error())
+			return nil
 		case e, ok := <-w.ResultChan():
 			if !ok {
 				return nil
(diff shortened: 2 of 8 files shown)


pkg/server/get_analysisrun_logs_v1alpha1.go
```diff
@@ -172,8 +173,9 @@ func (s *server) GetAnalysisRunLogs(
 				return fmt.Errorf("error sending log chunk: %w", err)
 			}
 		case <-ctx.Done():
-			// Context canceled or timed out
-			return ctx.Err()
+			logger := logging.LoggerFromContext(ctx)
+			logger.Debug(ctx.Err().Error())
```
thomastaylor312, line 177: Can we add error as a field and put a message here so we have context around the error? I know we can track down a line, but it is easier as a human to see something like "watch hangup" or something as the log message.

Same thing in all the other places where we added this

krancour: In case you haven't seen it firsthand, this is what the messages currently end up looking like:

```
2025-09-25T16:28:45Z	DEBUG	server/watch_warehouses_v1alpha1.go:57	context canceled	{"connect.service": "akuity.io.kargo.service.v1alpha1.KargoService", "connect.method": "WatchWarehouses", "connect.start_time": "2025-09-25T16:28:44Z"}
```

I think this is already not too far off from what you've suggested, with the main difference only being that I deliberately avoided the words "hang up" because (although I expect the client hanging up to be the typical cause of this) I don't want to assume that's the _only_ reason the context may have been cancelled.

If you think this is _not_ close enough to what you're suggesting, or that it doesn't adequately pinpoint the cause of the I'm still happy to hear that and iterate further.

thomastaylor312: I think this is good

thomastaylor312 (approved):

PR #5128, opened 2025-09-30
fix: remove verified status from Freight on failed re-verification

filip-aipl:
## Problem

**User Story:**

As a platform engineer managing progressive delivery pipelines, I need failed re-verifications to block downstream promotions so that I can ensure only verified Freight progresses through my stages.

**Scenario:**

1. I have a Stage (e.g., `staging`) with verification configured using AnalysisRuns
2. Freight is promoted to `staging` and passes initial verification
3. The Freight is marked as verified in `staging` (`Freight.Status.VerifiedIn["staging"]` is set)
4. I update the verification logic with new/stricter metrics (e.g., add error rate threshold)
5. I trigger re-verification of the current Freight in `staging`
6. The re-verification **fails** with the updated metrics
7. **Expected behavior**: Freight should no longer be available for promotion to downstream stages (e.g., `production`)
8. **Actual behavior**: Freight remained marked as verified, allowing promotion to `production` to proceed

**Root Cause:**

The `markFreightVerifiedForStage()` function only added verified status when verification succeeded, but never removed it when verification failed. This meant that once Freight was marked as verified, it stayed verified regardless
  of subsequent re-verification failures.

## Solution

Modified `markFreightVerifiedForStage()` in `pkg/controller/stages/regular_stages.go` to explicitly handle failed verification phases:

- Check if verification phase is terminal but not successful (`Phase.IsTerminal() && Phase != Successful`)
- When verification fails (Failed, Error, Aborted, or Inconclusive):
  - Remove the Stage from `Freight.Status.VerifiedIn` map
- This ensures `IsVerifiedIn()` checks in downstream stages correctly return `false`
- Downstream promotions are blocked until Freight is successfully re-verified

The implementation uses the existing `IsTerminal()` method for cleaner, more maintainable code that automatically handles any future failure phases.

## Testing

Updated test case `removes verified status when verification fails` to validate:
- Freight with verified status in Stage
- After verification failure, verified status is removed
- Downstream availability check will now correctly fail

## Impact

This fix ensures the verification system behaves as expected:
- Failed re-verification now properly blocks downstream promotions
- Platform teams can safely update verification criteria and re-verify existing Freight
- The verification gate functions correctly as a quality control mechanism

---

**Note:** If this behavior is not what was originally intended (i.e., if failed re-verification should _not_ remove verified status), please let me know and I'll adjust on my end.

### DIFF
--- a/pkg/controller/stages/regular_stages.go
+++ b/pkg/controller/stages/regular_stages.go
@@ -1187,8 +1187,50 @@ func (r *RegularStageReconciler) markFreightVerifiedForStage(
 	curFreight := stage.Status.FreightHistory.Current()
 	if curFreight == nil ||
 		len(curFreight.VerificationHistory) == 0 ||
-		curFreight.HasNonTerminalVerification() ||
-		curFreight.VerificationHistory.Current().Phase != kargoapi.VerificationPhaseSuccessful {
+		curFreight.HasNonTerminalVerification() {
+		return newStatus, nil
+	}
+
+	currentVerification := curFreight.VerificationHistory.Current()
+
+	// If verification failed, remove the verified status to block downstream
+	// promotions.
+	if currentVerification.Phase.IsTerminal() && currentVerification.Phase != kargoapi.VerificationPhaseSuccessful {
+
+		for _, ref := range curFreight.Freight {
+			freight := &kargoapi.Freight{}
+			if err := r.client.Get(ctx, types.NamespacedName{
+				Namespace: stage.Namespace,
+				Name:      ref.Name,
+			}, freight); err != nil {
+				return newStatus, fmt.Errorf(
+					"error getting Freight %q in namespace %q: %w",
+					ref.Name, stage.Namespace, err,
+				)
+			}
+
+			// If the Freight is not verified in this Stage, nothing to remove.
+			if !freight.IsVerifiedIn(stage.Name) {
+				continue
+			}
+
+			// Remove the verified status.
+			if err := kubeclient.PatchStatus(ctx, r.client, freight, func(status *kargoapi.FreightStatus) {
+				if status.VerifiedIn != nil {
+					delete(status.VerifiedIn, stage.Name)
+				}
+			}); err != nil {
+				return newStatus, fmt.Errorf(
+					"error removing verified status from Freight %q in Stage: %w",
+					freight.Name, err,
+				)
+			}
+		}
+		return newStatus, nil
+	}
+
+	// Only proceed if verification was successful.
+	if currentVerification.Phase != kargoapi.VerificationPhaseSuccessful {
 		return newStatus, nil
 	}
 
--- a/pkg/controller/stages/regular_stages_test.go
+++ b/pkg/controller/stages/regular_stages_test.go
@@ -3012,7 +3012,7 @@ func TestRegularStageReconciler_markFreightVerifiedForStage(t *testing.T) {
 			},
 		},
 		{
-			name: "skips verification when last verification is not successful",
+			name: "removes verified status when verification fails",
 			stage: &kargoapi.Stage{
 				ObjectMeta: metav1.ObjectMeta{
 					Namespace: "fake-project",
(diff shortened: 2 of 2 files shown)


krancour:
> If this behavior is not what was originally intended (i.e., if failed re-verification should not remove verified status), please let me know and I'll adjust on my end.

The current behavior was intended, but I agree probably isn't ideal. I'm open to discussing how it can be improved. That being said, I think there's a lot more to discuss here before moving forward...

In your scenario you laid out, which was perfectly plausible, add a 3.5 where the verified Freight has been promoted downstream -- and not necessarily just one "level." It could have gone all the way to prod.

It's not immediately obvious what the "right" way to deal with that should be.

filip-aipl:
> > If this behavior is not what was originally intended (i.e., if failed re-verification should not remove verified status), please let me know and I'll adjust on my end.
> 
> The current behavior was intended, but I agree probably isn't ideal. I'm open to discussing how it can be improved. That being said, I think there's a lot more to discuss here before moving forward...
> 
> In your scenario you laid out, which was perfectly plausible, add a 3.5 where the verified Freight has been promoted downstream -- and not necessarily just one "level." It could have gone all the way to prod.
> 
> It's not immediately obvious what the "right" way to deal with that should be.

Agreed.
Here's my thinking:

If Freight has already been promoted downstream when re-verification fails in "staging", we can:
1. Remove verified status from "staging" (block future promotions FROM here). Implemented in this PR.
2. Remove verified status from downstream stages that received this Freight (acknowledge it's deployed but no longer meets current verification standards). Looking at the code, it seems possible to implement.
3. Leave `CurrentlyIn` unchanged (it's still deployed, we're just marking it as "unverified by current criteria").

This creates a "stop here" moment: the Freight is out there, but we're not making it worse by continuing to promote unverified Freight. Operators must consciously re-verify or manually approve to continue.

If this has already happened, that's on us for not updating verification first. We can't undo it, but we can prevent cascading it further.

What do you think?

krancour:
@filip-aipl I think you're thinking in the right direction, but...

1. Finding all downstream Stages where the Freight was also verified is doable, but not entirely easy.

2. It's not entirely clear that verification in downstream Stages _should_ be rescinded. The verification processes in different Stages are apt to be different. For instance, consider that the battery of tests brought to bear on the software in the uat Stage and a perf Stage are likely quite different.

3. If verification is rescinded for Freight in specific Stages that already are using that Freight, it's surely going to create a head-scratching moment for someone who looks at it and wonders, _"How did this get here? It shouldn't have."_

I'm partly inclined to explore alternative options here, such as allowing a failed re-verification to "taint" Freight, but that idea also requires a great deal of thought.

filip-aipl:
@krancour 
I see your point. What are the next steps here?

Should I close this PR and move the discussion to a new issue to explore the "taint" concept and other alternatives? Or is blocking promotions from the failing Stage (without downstream cascade) acceptable as a first step?

Happy to take either direction - just want to make sure we're aligned on the approach.

carlin-airspace-intelligence:
What I expect from a pipeline above all else: correctness and safety.  Here, I think that touches on both of these points:

>  It's not entirely clear that verification in downstream Stages _should_ be rescinded. The verification processes in different Stages are apt to be different. For instance, consider that the battery of tests brought to bear on the software in the uat Stage and a perf Stage are likely quite different.

The property that I expect a pipeline to enforce: if a freight is in stage `S_n`, then that same freight is verified in stages `S_1 ... S_{n-1}`.  If ever the pipeline finds out that a freight isn't verified in `S_n`, I expect that freight to be blocked from deployment in `S_m` where `m > n`.  

This is the property that's enforced by default during normal operations.  That different stages can have different verification flows doesn't impact this property.  

Likewise

> If verification is rescinded for Freight in specific Stages that already are using that Freight, it's surely going to create a head-scratching moment for someone who looks at it and wonders, "How did this get here? It shouldn't have."

A worse head-scratching moment: during a RCA asking "how did this change get deployed to a prod stage after we saw the change failing upstream [at any time]?".  As a user of this system, I'd much prefer the system making it impossible to taking potentially unsafe options without explicit override.  

Right now: we have a potentially unsafe operation that the system allows by default.  This is what scares me during current flow.

carlin-airspace-intelligence:
Let me walk this back from the more general case.

I'm not going to die on the hill of verification failing `S_n` should now 100% block `S_{n+2}`.  

But!  I do think that if I reverify `S_n` and it now fails, I shouldn't be allowed to promote from `S_n` to `S_{n+1}`.  That is obviously unsafe and allowed by default.

krancour:
@filip-aipl yes... closing this and opening an issue is probably called for.

filip-aipl:
Done, happy to talk there.
https://github.com/akuity/kargo/issues/5144

Issue #5111, opened 2025-09-26
UI: massive amount of subscription nodes look pipeline empty


Marvin9:
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate bug report.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

When there are tons of subscription to warehouse, the default fit view cause the node size very tiny almost invisible.

# Screenshots

<img width="1440" height="752" alt="Image" src="https://github.com/user-attachments/assets/1a585b61-9e17-44f6-9f8d-78899be4818c" />

This is just with 20 subscriptions, with 100 or 200 it become worse

# Steps to Reproduce

Create warehouse with tons of subscriptions

# Version

```
1.7
```

# Logs

```
Paste any relevant application logs here.
```

PR #5137, opened 2025-10-01
feat(ui): exclude repo subscription nodes in fit view

Marvin9:
fixes https://github.com/akuity/kargo/issues/5111

https://github.com/user-attachments/assets/80e0a5ee-5647-47c8-b8d1-ee8879d30f78


rpelczar (changes requested):
@Marvin9 can we add this as a condition if there are more than 5 subscriptions? I think it should be visible by default in most cases.

rpelczar (approved):

Issue #5112, opened 2025-09-26
UI: Improve accessibility in image history section


Marvin9:
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

<img width="452" height="324" alt="Image" src="https://github.com/user-attachments/assets/71bcf9b9-0e8e-4523-88c6-1314e346db83" />

Add contrast border to those boxes

# Motivation

Accessibility

PR #5138, opened 2025-10-01
feat(ui): stage box accessibility

Marvin9:
fixes https://github.com/akuity/kargo/issues/5112

Screenshot 2025-10-01 at 11 36 04 AM

### DIFF
--- a/ui/src/features/project/pipelines/image-history/stage-box.tsx
+++ b/ui/src/features/project/pipelines/image-history/stage-box.tsx
@@ -55,7 +55,7 @@ export const StageBox = memo(
     const style = {
       backgroundColor: baseColor,
       opacity: showHistory && !isHighlighted ? 0.6 : 1,
-      border: isHighlighted ? '2px solid rgba(255,255,255,0.5)' : '2px solid transparent',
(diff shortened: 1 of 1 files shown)


ui/src/features/project/pipelines/image-history/stage-box.tsx
```diff
@@ -55,15 +55,15 @@ export const StageBox = memo(
     const style = {
       backgroundColor: baseColor,
       opacity: showHistory && !isHighlighted ? 0.6 : 1,
-      border: isHighlighted ? '2px solid rgba(255,255,255,0.5)' : '2px solid transparent',
+      border: '2px solid rgba(0,0,0,0.5)',
```
jmalisz-akuity: I think that it looks better, but I would keep the transparent border for not highlighted buttons. It puts more emphasis on the current one. I'm not sure about the static colour, WDYT @rpelczar?

rpelczar: I agree. I also think the black is too strong. Maybe we can set 0.4 opacity?

Issue #5131, opened 2025-09-30
UI: Refresh button shouldn't be greyed out if Warehouse has failure condition


jessesuen:
# Description

It's common for a Warehouse to be misconfigured with credentials that don't work. When this happens, the Warehouse might have an error:

<img width="885" height="182" alt="Image" src="https://github.com/user-attachments/assets/836e4f87-af41-4cc1-b58a-f6f1fc01a88c" />

After fixing the credentials in the secret, users have no way of testing the warehouse to see if the new credentials work, other than waiting several minutes until the next reconciliation. This is because the refresh button is greyed out.

# Steps to Reproduce

1. Misconfigure the credentials to a Warehouse, watch it fail
2. Correct the credentials
3. Needlessly wait until next reconciliation.

# Version

```
v1.7.5
```

jessesuen:
I have a feeling that the UI greys out the Refresh button because of this condition:

```yaml
    - lastTransitionTime: 2025-09-30T21:11:53Z
      message: Discovering artifacts for 1 subscriptions
      observedGeneration: 1
      reason: ScheduledDiscovery
      status: "True"
      type: Reconciling
```

However, we need a way for the users to retry with new credentials once they are fixed.

jessesuen:
I think the UI should use this logic:

* if `metadata.annotations['kargo.akuity.io/refresh']` is set, and is != `status.lastHandledRefresh`, then only then should it grey out the refresh button. Otherwise, it should be possible to click "Refresh".
* We can still show a spinner if the `Reconciling` condition is set, but that should be independent of whether or not the refresh button is greyed out..

PR #5139, opened 2025-10-01
feat(ui): allow refreshing warehouse on error

Marvin9:
fixes https://github.com/akuity/kargo/issues/5131

https://github.com/user-attachments/assets/0e39aca4-fd3e-4275-bb13-755e0c999a18

### DIFF
--- a/ui/src/features/project/pipelines/nodes/warehouse-node.tsx
+++ b/ui/src/features/project/pipelines/nodes/warehouse-node.tsx
@@ -116,7 +116,7 @@ export const WarehouseNode = (props: { warehouse: Warehouse }) => {
       <center>
         <Button
           size='small'
-          icon={<FontAwesomeIcon icon={faRefresh} spin={warehouseState.refreshing} />}
+          icon={<FontAwesomeIcon icon={faRefresh} />}
           onClick={(e) => {
             e.stopPropagation();
             refreshWarehouseMutation.mutate({
@@ -124,7 +124,7 @@ export const WarehouseNode = (props: { warehouse: Warehouse }) => {
               name: props.warehouse?.metadata?.name
             });
           }}
-          disabled={warehouseState.refreshing}
+          loading={warehouseState.refreshing}
         >
           Refresh{warehouseState.refreshing && 'ing'}
         </Button>


ui/src/features/project/pipelines/nodes/warehouse-node.tsx
```diff
@@ -111,15 +111,19 @@ export const WarehouseNode = (props: { warehouse: Warehouse }) => {
...
             refreshWarehouseMutation.mutate({
               project: props.warehouse?.metadata?.namespace,
               name: props.warehouse?.metadata?.name
             });
           }}
+          disabled={
+            warehouseState.refreshing &&
+            props.warehouse?.metadata?.annotations?.['kargo.akuity.io/refresh'] !==
+              props.warehouse?.status?.lastHandledRefresh
+          }
```
jessesuen: Will we be able to click refresh even if the Warehouse is in a stable state? e.g. it is not reconciling? Because that use case also needs to be preserved (my warehouse is configured correctly, and is at a steady state, but I need to refresh it to see new Freight).

jessesuen:
NIT: Can we consider reworking the UI to have "Reconciling" spinner as a separate concept from Refresh? Right now, the spinner, which represents reconciliation, is displayed in the actual button to refresh, which implies that they are the same thing, which they are not.

ui/src/features/project/pipelines/nodes/warehouse-node.tsx
Marvin9: Yep thats correct. User can refresh normally

Marvin9:
> NIT: Can we consider reworking the UI to have "Reconciling" spinner as a separate concept from Refresh? Right now, the spinner, which represents reconciliation, is displayed in the actual button to refresh, which implies that they are the same thing, which they are not.

Sure. I can move it to header of warehouse

Marvin9:
@jessesuen please check updated demo

ui/src/features/project/pipelines/nodes/warehouse-node.tsx
```diff
@@ -111,15 +116,15 @@ export const WarehouseNode = (props: { warehouse: Warehouse }) => {
...
-          loading={warehouseState.refreshing}
+          icon={<FontAwesomeIcon icon={faRefresh} spin={warehouseState.refreshing} />}
           onClick={(e) => {
             e.stopPropagation();
             refreshWarehouseMutation.mutate({
               project: props.warehouse?.metadata?.namespace,
               name: props.warehouse?.metadata?.name
             });
           }}
+          disabled={warehouseState.refreshing}
```
rpelczar: Why did you replace the loading prop with this?

Marvin9: Before my last changes, I needed the icon to spin but not disabled, which is no longer requirement since I moved reconciling state outside button

rpelczar (approved):
LGTM

Issue #5132, opened 2025-09-30
Better validation of image repository URLs


jessesuen:
# Proposed Feature

On more than one occasion, users have inputted credentials for an image repository like the following, where the `https://` protocol was included in the registry URL, e.g.:
* `https://ghcr.io/akuity/kargo`

AFIK, `https://` is never a valid protocol for images.

To make matters worse, the UI even suggests **_using_** `https://` in the example:

<img width="592" height="509" alt="Image" src="https://github.com/user-attachments/assets/621518fd-56c4-48c1-9142-490e3351a648" />

I think we should do the following:
* ensure image credentials repoURL field either omits the protocol, or that protocol is `oci://`. 
* update the UI suggestion so it's a proper example

# Motivation

Users are having a basic problem of inputting credentials, only for it to fail during warehouse discovery, which makes for a poor UX.

jessesuen:
NOTE: I understand that we can only perform validation if the repoURL is not a regular expression. So my proposed validation would only apply if `repoURLIsRegex` is omitted / false.

But if `repoURLIsRegex` is set to true, we **_could_** have validation that the regex itself is a valid expression (e.g. we attempt to compile it).

jessesuen:
A quick win would simply be to update the misleading suggestion in the UI, without having to implement a validating webhook for Secrets

PR #5140, opened 2025-10-01
fix(ui): repoUrl validation

Marvin9:
fixes https://github.com/akuity/kargo/issues/5132

https://github.com/user-attachments/assets/9444e4f8-cf54-45aa-9e9b-6f55a1c4c328

### DIFF
--- a/ui/src/features/project/settings/views/credentials/schema-validator.ts
+++ b/ui/src/features/project/settings/views/credentials/schema-validator.ts
@@ -4,7 +4,7 @@ import { dnsRegex } from '@ui/features/common/utils';
 import { zodValidators } from '@ui/utils/validators';
 
 const imageNameRegex =
-  /^(?:(?:[a-zA-Z0-9.-]+(?::[0-9]+)?\/)?(?:[a-z0-9]+(?:[._-][a-z0-9]+)*\/)*[a-z0-9]+(?:[._-][a-z0-9]+)*)$/;
+  /^(?![a-zA-Z][a-zA-Z0-9+.-]*:\/\/)(\w+([.-]\w+)*(:\d+)?\/)?(\w+([.-]\w+)*)(\/\w+([.-]\w+)*)*$/;
 
 export const createFormSchema = (genericCreds: boolean, editing?: boolean) => {
   let schema = z


jessesuen:
Watched the video. Why wasn't this a valid Helm chart URL?

image

Helm chart repositories can be HTTP (e.g. https://charts.helm.sh/incubator) or OCI (e.g. oci://r.example.com/myuser)

krancour:
"valid Helm chart reference" may invite confusion. http/s repos can contain multiple different charts, so this URL, which is a URL to the _repo_ isn't really a fully qualified reference to a chart.

Marvin9:
So what should be valid references for Helm chart and Container registry?

ui/src/features/project/settings/views/credentials/schema-validator.ts
```diff
@@ -0,0 +1,82 @@
...
+            try {
+              new URL(data.repoUrl);
+            } catch {
+              return false;
+            }
+          }
+
+          return true;
+        },
+        { error: 'repoUrl must be a valid HTTPS URL.', path: ['repoUrl'] }
```
rpelczar:
```suggestion
        { error: 'Repo URL must be a valid HTTPS URL.', path: ['repoUrl'] }
```

please do the same for other validators

rpelczar (approved):
LGTM

krancour:
> So what should be valid references for Helm chart and Container registry?

Sorry... my previous comment was made from my phone, so I didn't go into greater depth.

There are big differences between "classic" http/s Helm chart repositories and OCI repositories.

Where http/s repositories are concerned, you can think of it as one giant bucket that contains revisions of many _different_ charts. But the important point is that a URL for an http/s Helm chart repository is a URL _to the "bucket"_ and not to the chart. So it would be inaccurate to call it a "chart reference."

OCI is different. A registry contains many repositories, where each repository can be thought of as a smaller bucket with its own URL. Typically, all the items in those "buckets" are simply different revisions _of the same thing_. So in this case, a URL for an OCI Helm chart repository is a URL to a bucket (as with http/s), but _is also_ a URL to the chart. In this case, it _would_ be accurate to call it a "chart reference."

So the bottom line is that if the user is using OCI, your current wording is correct guidance, but if the user is using http/s, it is not. Fortunately, there is different wording that will cover both scenarios accurately.

Even though the exact definition of "repository" is _different_ in the case of http/s vs OCI, the following wording would accurately cover both cases:

> repoURL must be a valid Helm chart repository.

jessesuen:
I think @Marvin9 is not hung upon the error messgae, he's just asking what we should use as the regex to validate Helm chart repos. AFAIK, it's valid as long as it's a normal URI that starts with either `oci://`, `https://`, or `http://`, but not sure if other strings are also valid (e.g., omitting the protocol).

krancour:
> I think @Marvin9 is not hung upon the error messgae, he's just asking what we should use as the regex to validate Helm chart repos.

Ahh... if this is the case, then this is correct:

> AFAIK, it's valid as long as it's a normal URI that starts with either oci://, https://, or http://

To answer this:

> but not sure if other strings are also valid (e.g., omitting the protocol).

Omitting the protocol should not be allowed (for a Helm chart). This is another quirk of Helm. We're accustomed to images from OCI repositories having URLs that omit `oci://`, but for whatever reason, Helm _requires_ the leading `oci://`... and Kargo's rule of thumb is that when we integrate with something, we play by their rules. So in this case, it means chart URLs should always lead with the protocol.

(The back end _might_ be tolerant of a missing `oci://` because it gets dropped during normalization, but we shouldn't _rely_ on internal details of the normalization logic because we assume that logic is free to change without adverse effects.)

ui/src/features/project/settings/views/credentials/schema-validator.ts
```diff
@@ -0,0 +1,82 @@
+import { z } from 'zod';
+
+import { dnsRegex } from '@ui/features/common/utils';
+import { zodValidators } from '@ui/utils/validators';
+
+const helmChartRegex =
+  /^(oci:\/\/)?([a-z0-9]+(?:[._-][a-z0-9]+)*\/)*[a-z0-9]+(?:[._-][a-z0-9]+)*$/i;
```
jessesuen: Per discussion, this should allow for:
* `oci://`
* `http://`
* `https://`

ui/src/features/project/settings/views/credentials/schema-validator.ts
```diff
@@ -0,0 +1,82 @@
...
+      repoUrlIsRegex: z.boolean().optional(),
+      username: zodValidators.requiredString,
+      password: editing ? z.string().optional() : zodValidators.requiredString
+    })
+    .check(
+      z.refine(
+        (data) => {
+          if (data.type === 'git' && !data.repoUrlIsRegex) {
+            try {
+              new URL(data.repoUrl);
```
jessesuen, line 28: This is closely related. Kargo supports SSH URLs, but:
1. We don't have UI support for inputting SSH URLs or PEM keys. We should (See: https://github.com/akuity/kargo/issues/5149)
2. I'm not sure if this new validation will suddenly start rejecting existing credentials that are SSH based.

ui/src/features/project/settings/views/credentials/schema-validator.ts
jessesuen: Instead of a regex, should this attempt `new URL(data.repoUrl);` make sure it parses, then check that the protocol is `oci`, `http` or `https`.

ui/src/features/project/settings/views/credentials/schema-validator.ts
```diff
@@ -0,0 +1,82 @@
+import { z } from 'zod';
+
+import { dnsRegex } from '@ui/features/common/utils';
+import { zodValidators } from '@ui/utils/validators';
+
+const helmChartRegex =
+  /^(oci:\/\/)?([a-z0-9]+(?:[._-][a-z0-9]+)*\/)*[a-z0-9]+(?:[._-][a-z0-9]+)*$/i;
+
+const imageNameRegex = /^([a-z0-9]+(?:[._-][a-z0-9]+)*\/)*[a-z0-9]+(?:[._-][a-z0-9]+)*$/i;
```
jessesuen: @krancour for **_image_** repoURLs, is it valid to include protocol in the url (e.g., `oci://ghcr.io/akuity/kargo`) ? Or should it always omit protocol (e.g., `ghcr.io/akuity/kargo`)

@Marvin9 does this support a port? (e.g., `example.com:3000/foo/bar`). I think it doesn't

Marvin9: It does support port now. Waiting for @krancour to confirm if we accept protocol or not..

ui/src/features/project/settings/views/credentials/schema-validator.ts
Marvin9: I will have this in separate PR with API support

ui/src/features/project/settings/views/credentials/schema-validator.ts
Marvin9: @hiddeco can you answer this while @krancour is away?

hiddeco: The validation we apply on the image subscription itself is `(\w+([\.-]\w+)*(:[\d]+)?/)?(\w+([\.-]\w+)*)(/\w+([\.-]\w+)*)*$`.

Marvin9: Is it correct to say that validation should be same for subscription and credential?

hiddeco: Yes, that's correct.

PR #5143, opened 2025-10-01
feat: option to auto-promote Freight currently used immediately upstream

### LINKED ISSUE #4287: Freight selection policy/option to match Freight running in upstream stage
# Proposed Feature

When automatically promoting a Freight, the current behavior is that Kargo will choose the latest Freight that has been verified in the upstream. This works fine in the normal case where we are always producing new builds, and the latest is the one we want.

However, in a rollback scenario, the latest Freight is not desirable when I want to move to an older Freight. Consider the following example:

Image

As shown above, the four stages (`-central` and `-west`) on the right side are configured with auto-promotion. And the `-east` stages are manual.

The following is our current behavior:
* In the happy path, as soon as I manually promote a **_new_** Freight to `prod-us-east` or `prod-eu-east`, the `central` region will automatically get that version, followed by `west`. This is great, and what I would like.
* However, the same is not true if I were to rollback `prod-us-east` or `prod-eu-east` to an **_older_** Freight version. What currently happens is that the `central` and `west` regions would not be updated.

The feature I think we need would be to allow a Stage to be configured to auto-promote whatever Freight is running in the **_immediate_** upstream.

# Motivation

The use case is to provide a way to automatically sync **_the same Freight version_** across many Stages in a controlled manner. In my example, whatever Freight is running in the `east` regions represents the "desired" Freight to be automatically deployed in the downstream regions, regardless of whether there are newer versions or not.

krancour:
Fixes #4287

See preview docs: https://deploy-preview-5143.docs.kargo.io/user-guide/how-to-guides/working-with-stages#auto-promotion


hiddeco (approved):
🔥!

PR #5161, opened 2025-10-06
feat: support setting repo URLs in headers for artifactory webhooks

fuskovic:
Reopens https://github.com/akuity/kargo/pull/4945

Closes: https://github.com/akuity/kargo/issues/4899

### DIFF
--- a/pkg/webhook/external/artifactory.go
+++ b/pkg/webhook/external/artifactory.go
@@ -169,18 +169,6 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
 		var repoURLs []string
 		if repoURLsHeader := r.Header.Get(artifactoryRepoURLHeader); repoURLsHeader != "" {
 			repoURLs = strings.Split(repoURLsHeader, ",")
-			for _, repoURL := range repoURLs {
-				if _, err := url.Parse(repoURL); err != nil {
-					xhttp.WriteErrorJSON(
-						w,
-						xhttp.Error(
-							fmt.Errorf("invalid repo URL %q in %s header", repoURL, artifactoryRepoURLHeader),
-							http.StatusBadRequest,
-						),
-					)
-					return
-				}
-			}
 		} else {
 			originURL, err := url.Parse(payload.Origin)
 			if err != nil {
--- a/pkg/webhook/external/artifactory_test.go
+++ b/pkg/webhook/external/artifactory_test.go
@@ -365,29 +365,6 @@ func TestArtifactoryHandler(t *testing.T) {
 			},
 		},
 		{
-			name:       "invalid repo url in custom header",
-			secretData: testSecretData,
-			req: func() *http.Request {
-				bodyBytes, err := json.Marshal(validImagePushEventWithUnsetOrigin)
-				require.NoError(t, err)
-				req := httptest.NewRequest(
-					http.MethodPost,
-					testURL,
-					bytes.NewBuffer(bodyBytes),
-				)
-				req.Header.Set(artifactoryAuthHeader, signWithoutAlgoPrefix(bodyBytes))
-				req.Header.Set(artifactoryRepoURLHeader, "http://[::1]:namedport")
-				return req
-			},
-			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
-				require.Equal(t, http.StatusBadRequest, rr.Code)
-				require.JSONEq(t,
-					`{"error":"invalid repo URL \"http://[::1]:namedport\" in X-Kargo-Repo-URLs header"}`,
-					rr.Body.String(),
-				)
-			},
-		},
-		{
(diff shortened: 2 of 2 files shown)


pkg/webhook/external/artifactory.go
```diff
@@ -148,41 +149,51 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
...
+					xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
+				)
+				return
+			}
+
+			pathSections := strings.Split(payload.Data.Path, "/")
+			repoURL := strings.Join(
+				append(
+					[]string{originURL.Host, payload.Data.RepoKey},
+					pathSections[:len(pathSections)-2]...,
```
hiddeco, line 194: Think it would be wise to validate if `pathSections` actually contains `>=2` items.

pkg/webhook/external/artifactory.go
```diff
@@ -148,41 +149,51 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
 			return
 		}
 
-		originURL, err := url.Parse(payload.Origin)
-		if err != nil {
+		if payload.Data.ImageType != artifactoryDockerDomain && payload.Data.ImageType != artifactoryChartImageType {
```
hiddeco: I would add the supported types to a `var` slice and use `slices.Contains` here.

pkg/webhook/external/artifactory.go
```diff
@@ -148,41 +149,51 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
...
-		case artifactoryChartImageType:
-			repoURLs = append(repoURLs, urls.NormalizeChart(repoURL))
-		default:
-			xhttp.WriteErrorJSON(
-				w,
-				xhttp.Error(
-					fmt.Errorf("unsupported image type %q", payload.Data.ImageType),
-					http.StatusNotImplemented,
+		if repoURLsHeader := r.Header.Get(artifactoryRepoURLHeader); repoURLsHeader != "" {
+			repoURLs = strings.Split(repoURLsHeader, ",")
```
hiddeco, line 171: Should these URLs still go through validation (e.g. `url.Parse`)? Additionally, should we limit the number of allowed repository URLs?

pkg/webhook/external/artifactory.go
```diff
@@ -26,8 +27,14 @@ const (
 	artifactoryAuthHeader      = "X-Jfrog-Event-Auth"
 	artifactorySecretDataKey   = "secret-token"
 	artifactory                = "artifactory"
+	artifactoryRepoURLHeader   = "X-Kargo-Repo-URLs"
 )
 
+var validImageTypes = []string{
```
hiddeco:
```suggestion
var artifactoryValidImageTypes = []string{
```

pkg/webhook/external/artifactory.go
```diff
@@ -148,41 +155,71 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
 			return
 		}
 
-		originURL, err := url.Parse(payload.Origin)
-		if err != nil {
+		if !slices.Contains(validImageTypes, payload.Data.ImageType) {
```
hiddeco:
```suggestion
		if !slices.Contains(artifactoryValidImageTypes, payload.Data.ImageType) {
```

pkg/webhook/external/artifactory.go
```diff
@@ -148,41 +155,71 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
...
+			repoURLs = append(repoURLs, repoURL)
+		}
+
+		for i, repoURL := range repoURLs {
+			switch payload.Data.ImageType {
+			case artifactoryDockerDomain:
+				repoURLs[i] = urls.NormalizeImage(repoURL)
+			case artifactoryChartImageType:
+				repoURLs[i] = urls.NormalizeChart(repoURL)
+				// no default case because this is already checked above
```
hiddeco: Having an erroring default here would still function as a safeguard if/when the check above accidentally gets removed.

pkg/webhook/external/artifactory.go
```diff
@@ -148,41 +155,79 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
...
+			for _, repoURL := range repoURLs {
+				if _, err := url.Parse(repoURL); err != nil {
+					xhttp.WriteErrorJSON(
+						w,
+						xhttp.Error(
+							fmt.Errorf("invalid repo URL %q in %s header", repoURL, artifactoryRepoURLHeader),
+							http.StatusBadRequest,
+						),
+					)
+					return
```
hiddeco: This will not work because OCI repositories do not have a protocol defined, which will cause the URL parsing to fail. Given this, I would revert this for now.

Issue #5102, opened 2025-09-24
yaml-parse promotion step does not allow parsing YAML files that contain a list


bartpaczkows:
Hi,
The current implementation of the **yaml-parse** promotion step does not allow parsing YAML files that contain a list; only maps are supported. The step is failing when I to try to use it on YAM file which contains a list.

ERROR:
**could not parse YAML file: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal array into Go value of type map[string]interface {}**

# Proposed Feature
My proposed solution is to modify the readAndParseYAML function so that it always returns a map. If the parsed YAML file contains a list at the root, the function should wrap this list in a root element before returning the data structure.
For example, if our input YAML file looks like this:

```
- destination:
    name: demo1
    namespace: one
- destination:
    name: demo2
    namespace: two

```

The returned data should contain a "root" element, like so:

```
root:
- destination:
    name: demo1
    namespace: one
- destination:
    name: demo2
    namespace: two
```

# Motivation
This change would enable us to parse such files and use fromExpression statements like the following:
`filter(root, .destination?.name == "demo1")`
which is currently not possible.

krancour:
Ack on the issue. Not sure about the proposed solution. Further discussion is warranted before anyone implements anything.

kyounghoonJang:
@krancour 
Hi,
what are your thoughts on introducing an optional parameter to the yaml-Parse step? We could call it something like wrapListInKey.

For example, the configuration would look like this:

```
- uses: yaml-parse
  config:
    path: './my-service-list.yaml'
    # If the source file is a list, wrap it in a map using 'serviceList' as the key.
    wrapListInKey: "serviceList"
    outputs:
      - name: firstServiceName
        # Access the data using the key ('serviceList') created by the option above.
        fromExpression: serviceList[0].destination.name
```

krancour:
Actually, I'm now fairly convinced there's no issue here other than _possibly_ a hint about syntax that can be added to documentation to help people through cases such as this one.

Please see this response from me to another, very similar question:

https://github.com/akuity/kargo/discussions/5028#discussioncomment-14386360

If anyone confirms this syntax works for them, then I'll close this.

In the meantime, nobody should be pursuing this issue because I'm fairly certain than not a single code change is required.

bartpaczkows:
[...]

PR #5193, opened 2025-10-10
feat(yaml-parse): support non-map YAML roots via synthetic root key

bartpaczkows:
Enhances yaml-parse step so YAML documents with a top-level list or scalar (string/number/bool/null) are now supported. Non-map roots are wrapped under a synthetic key root to provide a stable object for expression evaluation.


krancour:
In [this comment](https://github.com/akuity/kargo/issues/5102#issuecomment-3368312963), I may have overestimated the possibility of #5102 requiring no code changes, but the notion that expr-lang has an idiomatic way of dealing with this holds. Utilizing a synthetic `root` key is not really the correct approach. The only thing that is really called for are some modifications to the pre-processing of the data.

I'm closing this in favor of #5209.

PR #5195, opened 2025-10-11
chore: Add extraContainer for controller

Crimrose:
The extraContainers feature allows users to inject additional containers into a Helm chart’s Pod specification without modifying the core template.
This is especially useful for extending functionality or adding sidecar containers for logging, monitoring, debugging.

Example:
```
extraContainers:
  - name: metrics-exporter
    image: myrepo/metrics-exporter:latest
    ports:
      - name: metrics
        containerPort: 9090
    resources:
      limits:
        cpu: 100m
        memory: 128Mi
```


hiddeco:
While I do think your PR by itself is sound, I do wonder what the desire is behind allowing this functionality.

If the core reason is to e.g. collect metrics, I think it might be better to expose a Prometheus endpoint on the controller to facilitate this.

Crimrose:
I want to add a sidecar container to handle cleanup tasks, such as removing unused Git repositories in the /tmp directory. Although we already have a step to clean up old files in `/tmp`, many files still remain. Restarting the pod would also clear them, but that approach could disrupt normal operations. Keeping too many files in `/tmp` can lead to high memory usage.

We can automate the cleanup process without affecting the main container’s functionality and this setup could be extended for other automation tasks in the future.

krancour:
@Crimrose similar to @hiddeco I do not necessarily oppose this change, but the motivation you're describing is raising _other_ concerns...

> Although we already have a step to clean up old files in /tmp, many files still remain.

How are you doing that exactly? Promotions processes, by design, don't have access to anything in the FS _except for_ Promotion-specific directories _under_ `/tmp`. Those directories are deleted automatically when a Promotion completes.

Concerns:

1. Security: Are you somehow breaking out and _actually_ deleting things from `/tmp` yourself?

1. GC: If stuff is accumulating in `/tmp`, that sounds as if there's a bug somewhere.

Again, I'm necessarily not opposed to this PR, but I really want to be certain we're not just slapping a band-aid over some concerning problems here instead of treating the root cause.

krancour:
Maintainers, by consensus, recently settled on a policy of not implementing requested chart features (or unsolicited chart feature PRs) if, in our estimation, they are too niche. "Kustomized" Helm is a well-established pattern of using available chart options to get yourself as close to your desired configuration as possible, then performing last mile tweaks using Kustomize. That is how I would suggest approaching this.

I will, however, _happily_ reconsider this PR if it becomes clear that this is not so niche, with many users reporting well-justified reasons they need this.

PR #5208, opened 2025-10-13
refactor!: remove deprecated `SemVerConstraint` field 

nitishfy:
Related to https://github.com/akuity/kargo/pull/4541

### DIFF
--- a/ui/src/features/project/pipelines/warehouse/repo-subscriptions.tsx
+++ b/ui/src/features/project/pipelines/warehouse/repo-subscriptions.tsx
@@ -129,9 +129,9 @@ export const RepoSubscriptions = ({ subscriptions }: Props) => {
                 </Descriptions.Item>
               )}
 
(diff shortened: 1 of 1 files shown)


pkg/image/semver_selector.go
```diff
@@ -47,11 +47,6 @@ func newSemverSelector(
 		strictSemvers:    sub.StrictSemvers,
 	}
 	constraintStr := sub.Constraint
```
krancour: Nit: I would skip this. It made more sense before, but makes less sense with the following five lines now removed. Now it's nothing but an unnecessary allocation.

krancour:
@nitishfy looks like there might have been some UI things still dependent on the removed field.

Please don't hesitate to ask @Marvin9 for help if need be.

Once that's worked out, this will LGTM.

pkg/image/semver_selector.go
nitishfy: done, please take a look now.

PR #5209, opened 2025-10-13
feat(controller): yaml-parse and json-parse improvements

### LINKED ISSUE #5102: yaml-parse promotion step does not allow parsing YAML files that contain a list

krancour:
Fixed #5102

Supersedes #5193

### DIFF
--- a/pkg/promotion/runner/builtin/json_parser.go
+++ b/pkg/promotion/runner/builtin/json_parser.go
@@ -86,44 +86,50 @@ func (jp *jsonParser) run(
 	}, nil
 }
 
-// readAndParseJSON reads a JSON file and unmarshals it into a map.
-func (jp *jsonParser) readAndParseJSON(workDir string, path string) (map[string]any, error) {
-
+// readAndParseJSON reads and unmarshals a JSON document and returns the result.
+func (jp *jsonParser) readAndParseJSON(
+	workDir string,
+	path string,
+) (any, error) {
 	absFilePath, err := securejoin.SecureJoin(workDir, path)
 	if err != nil {
 		return nil, fmt.Errorf("error joining path %q: %w", path, err)
 	}
-
 	jsonData, err := os.ReadFile(absFilePath)
 	if err != nil {
 		return nil, fmt.Errorf("error reading JSON file %q: %w", absFilePath, err)
 	}
-
-	var data map[string]any
+	var data any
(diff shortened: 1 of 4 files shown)


krancour:
@fykaa I'm asking for you to review this PR please due to your familiarity with this section of code. Thanks.

fykaa:
@krancour as a part of this PR and [this comment](https://github.com/akuity/kargo/discussions/5028#discussioncomment-14386360), do you also think we should document it's behavior how `$env` always points to the `root` of the parsed file and that the access style depends on root type.

krancour:
> do you also think we should document it's behavior how $env always points to the root of the parsed file and that the access style depends on root type

It's not unique to these steps. It's not even unique to Kargo. I've been thinking about it for a while now because of nil-coalescing, which is already mentioned in many places throughout the docs, and I haven't really come to any conclusions yet, but I think we need to fundamentally rethink if, how, and/or how much we are going to instruct readers on expr-lang syntax, which is all that this is.

With so many higher priorities, I don't want to grapple with that right now.

fykaa:
Okay! I'm running some final end-to-end tests now, and after that, these changes should be good to go!

fykaa (approved):
I got it working e2e. these changes support array and list roots in the yaml/json parser steps. and the updated type handling allows unmarshalling both objects and arrays at the root level

PR #5217, opened 2025-10-15
fix(events): Add apiVersion to object reference

thomastaylor312:
This fixes an issue from updating how events were created that made it so things don't appear in the UI


thomastaylor312:
Things work now 
image

jessesuen (approved):

Issue #5222, opened 2025-10-15
refactor artifactory webhook receiver repo url customization


krancour:
The impetus behind #5161 is understood, but the implementation promotes a poor UX. I've spoken with @jessesuen about this and we'd like to see #5161 reverted and a different approach brought to bear on the problem it was solving.

Usability issues:

1. If you register a webhook at the org-level (or even just one webhook for multiple, specific repos) in Artifactory, you will need to provide _many_ comma-delimited repo URLs as the value of the custom header. This can obviously be quite onerous if you're dealing with more than just a few repos. _Worse_ the larger that list gets, the larger the performance penalty you will pay. Since Warehouses subscribed to every repo listed in the header value are refreshed, expensive artifact discovery cycles will be triggered for many Warehouses _not subscribed to the repo from which the webhook actually originated_. For a large number of repositories, this almost completely defeats the point of using webhooks -- which is to trigger artifact discovery for select Warehouses _only when we expect discovery to actually find new artifacts_.

2. If you want to avoid the performance penalty inherent to the first option, your only recourse is to register _separate_ webhooks for every repository. While this successfully sidesteps the performance penalty, it is perhaps even more onerous to configure and maintain than the first option.

I'd previously proposed that a better solution to the problem (mismatch between Artifactory virtual repository URLs used by Warehouses and the repository domains, paths, and names embedded in webhook bodies) would be to leverage the (not yet completed) generic webhook receiver to concatenate a hard-configured domain/path prefix with select information from the webhook body (like repo name) to construct virtual repository URLs whose subscribing Warehouses could then be refreshed. This was initially rejected because work on the generic webhook receiver is expected to take quite some time still and we prefer to have a solution for this problem in v1.8.0.

While discussing usability issues in #5161 with @jessesuen yesterday, we arrived at what we believe is an improvement. We are proposing a more focused application of what could be done with the generic receiver (if it were ready). What we want is for the Artifactory webhook receiver to be configurable in a way that describes how to construct virtual repository URLs from information contained in the webhook bodies.

In the _worst_ case, this might look like what I proposed doing with the generic receiver -- the user provides an expression that concatenates a hard-configured prefix with information extracted from the body (like repo name).

[...]

PR #5225, opened 2025-10-16
feat: artifactory webhook virtual repo support

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5222#issuecomment-3408695877

Supercedes + reverts: https://github.com/akuity/kargo/pull/5161


api/v1alpha1/project_config_types.go
```diff
@@ -287,6 +287,14 @@ type ArtifactoryWebhookReceiverConfig struct {
...
 	// +kubebuilder:validation:Required
 	SecretRef corev1.LocalObjectReference `json:"secretRef" protobuf:"bytes,1,opt,name=secretRef"`
+	// VirtualRepoName is the name of the Artifactory virtual repository from
+	// which webhook events can be resolved for. If present, this value will end up
+	// substituting the local or remote repository repoKey in the webhook payload.
+	// For example: if your virtual repoURL is myinstance.jfrog.io/my-virtual-repo-name/path/to/artifact
+	// the "my-virtual-repo-name" part is what should be set here.
+	//
+	// +optional
+	VirtualRepoName string `json:"virtualRepoName" protobuf:"bytes,2,opt,name=virtualRepoName"`
```
krancour: For whatever reason, controller-runtime requires `omitempty` for a field to truly be optional.

```suggestion
	VirtualRepoName string `json:"virtualRepoName,omitempty" protobuf:"bytes,2,opt,name=virtualRepoName"`
```

api/v1alpha1/project_config_types.go
```diff
@@ -287,6 +287,14 @@ type ArtifactoryWebhookReceiverConfig struct {
 	//
 	// +kubebuilder:validation:Required
 	SecretRef corev1.LocalObjectReference `json:"secretRef" protobuf:"bytes,1,opt,name=secretRef"`
+	// VirtualRepoName is the name of the Artifactory virtual repository from
+	// which webhook events can be resolved for. If present, this value will end up
+	// substituting the local or remote repository repoKey in the webhook payload.
+	// For example: if your virtual repoURL is myinstance.jfrog.io/my-virtual-repo-name/path/to/artifact
+	// the "my-virtual-repo-name" part is what should be set here.
```
krancour: I think anyone who hasn't researched this topic to the extent that we have will need a more in-depth and technically precise explanation here.

```suggestion
	// VirtualRepoName is the name of an Artifactory virtual repository.
	//
	// When unspecified, the Artifactory webhook receiver depends on the value of
	// the webhook payload's `data.repo_key` field when inferring the URL of the
	// repository from which the webhook originated, which will always be an
	// Artifactory "local repository." In cases where a Warehouse subscribes to
	// such a repository indirectly via a "virtual repository," there will be a
	// discrepancy between the inferred (local) repository URL and the URL
	// actually used by the subscription, which can prevent the receiver from
	// identifying such a Warehouse as one in need of refreshing. When specified,
	// the value of the VirtualRepoName field supersedes the value of the webhook
	// payload's `data.repo_key` field to compensate for that discrepancy.
	//
	// In practice, when using virtual repositories, a separate Artifactory
	// webhook receiver should be configured for each, but one such receiver can
	// handle inbound webhooks from any number of local repositories that are
	// aggregated by that virtual repository. For example, if a virtual repository
	// `proj-virtual` aggregates container images from all of the `proj`
	// Artifactory project's local image repositories, with a single webhook
	// configured to post to a single receiver configured for the `proj-virtual`
	// virtual repository, an image pushed to
	// `example.frog.io/proj-<local-repo-name>/<path>/image`, will cause that
	// receiver to refresh all Warehouses subscribed to
	// `example.frog.io/proj-virtual/<path>/image`.	
```

pkg/webhook/external/artifactory.go
```diff
@@ -166,58 +167,54 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
...
+				xhttp.Error(errors.New("invalid request body"), http.StatusBadRequest),
+			)
+			return
+		}
+
+		pathSections := strings.Split(payload.Data.Path, "/")
+		if len(pathSections) < 2 {
+			xhttp.WriteErrorJSON(
+				w,
+				xhttp.Error(errors.New("invalid path"), http.StatusBadRequest),
```
krancour: It seems inconsistent to be specific about what's invalid about the request in one case and vague in the other.

```suggestion
				xhttp.Error(
					fmt.Errorf("invalid value %q in payload's jpd_origin field", payload.Origin)
					http.StatusBadRequest,
				),
			)
			return
		}

		pathSections := strings.Split(payload.Data.Path, "/")
		if len(pathSections) < 2 {
			xhttp.WriteErrorJSON(
				w,
				xhttp.Error(
					fmt.Errorf("invalid value %q in payload's data.path field", payload.Data.Path),
					http.StatusBadRequest,
				),
```

pkg/webhook/external/artifactory.go
```diff
@@ -166,58 +167,54 @@ func (a *artifactoryWebhookReceiver) getHandler(requestBody []byte) http.Handler
...
+			payload.Data.RepoKey = a.virtualRepoName
+		}
+
+		repoURL := strings.Join(
+			append(
+				[]string{originURL.Host, payload.Data.RepoKey},
+				pathSections[:len(pathSections)-2]...,
+			),
+			"/",
+		)
```
krancour, line 205: This technically works, of course, but it definitely feels unconventional to tamper with the payload. I'd suggest this is:

```suggestion
		repoName := payload.Data.RepoKey
		if a.virtualRepoName != "" {
			repoName = a.virtualRepoName
		}

		repoURL := strings.Join(
			append(
				[]string{originURL.Host, repoName},
				pathSections[:len(pathSections)-2]...,
			),
			"/",
		)
```

pkg/webhook/external/artifactory_test.go
```diff
@@ -71,6 +65,19 @@ func TestArtifactoryHandler(t *testing.T) {
...
+		Data: artifactoryEventData{
+			Tag:       "v1.0.0",
+			Path:      "foo/bar/test-chart/latest/chart.tgz",
+			ImageType: artifactoryChartImageType,
+			RepoKey:   "test-repo",
+			ImageName: "test-chart",
+		},
+		Origin: "https://artifactory.example.com",
+	}
+
```
krancour: I think with very small tweaks to the new test case, any of the existing test payloads could be reused instead of creating yet another.

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/artifactory/index.md
```diff
@@ -88,6 +88,46 @@ spec:
...
+virtual repository. This is because the webhook event Artifactory sends does
+not include the name of the virtual repository. Instead, it includes
+the repo key for the underlying local/remote repository the event generated 
+from.
+
+If the URL of your virtual repository is
+`https://myinstance.jfrog.io/my-virtual-repo-name/path/to/artifact` than the 
+`my-virtual-repo-name` part is what should be set for the `virtualRepoName`
+field.
+:::
```
krancour: It reads awkwardly when we carve out a whole section for this (which I agree it deserves), but call it a "note." Similarly, the info box feels awkward when it's not supplementing other content.

Here are some tweaks I would suggest. I also pulled some more descriptive wording from my suggestions for `warehouse_types.go`.

```suggestion
### Virtual Repositories

When Warehouses intended to be refreshed by an Artifactory webhook receiver
subscribe to Artifactory
[virtual repositories](https://jfrog.com/help/r/jfrog-artifactory-documentation/virtual-repositories) there will be discrepancies between the URLs the receiver will
infer for the
[local repositories](https://jfrog.com/help/r/jfrog-artifactory-documentation/local-repositories) from which push events have originated and the URLs actually used
by those Warehouses' subscriptions.

To compensate for this, a value can be provided for the Artifactory webhook
receiver configuration's `virtualRepoName` field. When specified, its value
supersedes the local repository name found in the webhook's payload, which
allows the receiver to infer the correct virtual repository URL for which all
subscribed Warehouses should be refreshed.

In practice, when using virtual repositories, a separate Artifactory webhook
receiver should be configured _for each_, but one such receiver can handle
events originating from _any number_ of local repositories that are aggregated by
that virtual repository. For example, if a virtual repository `proj-virtual`
aggregates container images from all of the `proj` Artifactory project's local
image repositories, with a single webhook configured to post to the following
receiver, an image pushed to
`example.frog.io/proj-<local-repo-name>/<path>/image`, will correctly cause that
receiver to refresh all Warehouses subscribed to
`example.frog.io/proj-virtual/<path>/image`.
```

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/artifactory/index.md
```diff
@@ -88,6 +88,46 @@ spec:
...
+metadata:
+  name: kargo-demo
+  namespace: kargo-demo
+spec:
+  webhookReceivers: 
+  - name: artifactory-wh-receiver
+    artifactory:
+      secretRef:
+        name: artifactory-wh-secret
+      virtualRepoName: my-virtual-repo-name
```
krancour: Because you would need one receiver per virtual repo, we should make the example receiver's name reflect the virtual repo it's for.

```suggestion
  webhookReceivers: 
  - name: proj-virtual-wh-receiver
    artifactory:
      secretRef:
        name: artifactory-wh-secret
      virtualRepoName: proj-virtual
```

krancour (changes requested):

PR #5229, opened 2025-10-16
feat: Add proxy-authorization flag to login command

### LINKED ISSUE #3537: Support `Proxy-Authorization` header in `kargo` CLI
# Checklist

# Proposed Feature

Provide a mechanism to supply the `kargo` CLI with a `Proxy-Authorization` header.

# Motivation

I have deployed the kargo API behind [GCP IAP](https://cloud.google.com/iap/docs/concepts-overview) so that it's only accessible to admins. Visiting the web-app is straightforward, since GCP intercepts the request and then forwards me to the UI once I log-in.

I would like to do the same via the CLI, and IAP supports the [Proxy-Authorization](https://cloud.google.com/iap/docs/authentication-howto#authenticating_from_proxy-authorization_header) header so as not to interfere with an application's use of the `Authorization` header. 

# Suggested Implementation

Adding a flag to `kargo login` either allowing arbitrary headers or a `--proxy-authorization` token would be amazing.

onasser1:
Fixes #3537

### DIFF
--- a/pkg/cli/client/client.go
+++ b/pkg/cli/client/client.go
@@ -16,7 +16,9 @@ import (
 )
 
 type Options struct {
-	InsecureTLS bool
+	Credentials      string
+	ProxyCredentials string
+	InsecureTLS      bool
 }
 
 // AddFlags adds the flags for the client options to the provided flag set.
@@ -45,7 +47,11 @@ func GetClientFromConfig(
 	if err != nil {
 		return nil, fmt.Errorf("error refreshing token: %w", err)
 	}
-	return GetClient(cfg.APIAddress, cfg.BearerToken, cfg.ProxyAuthCredentials, skipTLSVerify), nil
+	return GetClient(cfg.APIAddress, &Options{
+		Credentials:      cfg.BearerToken,
+		ProxyCredentials: cfg.ProxyAuthCredentials,
+		InsecureTLS:      skipTLSVerify,
+	}), nil
 }
 
 // GetClient returns a new client for the Kargo API server located at the
@@ -54,62 +60,29 @@ func GetClientFromConfig(
 // requests.
 func GetClient(
 	serverAddress string,
-	credential string,
-	proxyCredential string,
-	insecureTLS bool,
+	opts *Options,
 ) svcv1alpha1connect.KargoServiceClient {
 	httpClient := cleanhttp.DefaultClient()
 
-	if insecureTLS {
+	if opts.InsecureTLS {
 		transport := cleanhttp.DefaultTransport()
 		transport.TLSClientConfig = &tls.Config{
 			InsecureSkipVerify: true, // nolint: gosec
 		}
 		httpClient.Transport = transport
 	}
-	// Both credentials are empty
-	if credential == "" && proxyCredential == "" {
-		return svcv1alpha1connect.NewKargoServiceClient(httpClient, serverAddress)
-	}
 
-	if credential != "" && proxyCredential == "" {
-		return svcv1alpha1connect.NewKargoServiceClient(
-			httpClient,
-			serverAddress,
-			connect.WithClientOptions(
-				connect.WithInterceptors(
-					&authInterceptor{
-						credential: credential,
-					},
-				),
-			),
-		)
+	interceptor := &authInterceptor{}
+	if opts.Credentials == "" {
+		interceptor.credential = opts.Credentials
 	}
-
-	if credential == "" {
-		return svcv1alpha1connect.NewKargoServiceClient(
-			httpClient,
-			serverAddress,
-			connect.WithClientOptions(
-				connect.WithInterceptors(
-					&authInterceptor{
-						proxyCredential: proxyCredential,
-					},
-				),
-			),
-		)
+	if opts.ProxyCredentials == "" {
+		interceptor.proxyCredential = opts.ProxyCredentials
 	}
 
 	return svcv1alpha1connect.NewKargoServiceClient(
 		httpClient,
 		serverAddress,
-		connect.WithClientOptions(
-			connect.WithInterceptors(
-				&authInterceptor{
-					credential:      credential,
-					proxyCredential: proxyCredential,
-				},
-			),
-		),
+		connect.WithClientOptions(connect.WithInterceptors(interceptor)),
 	)
 }
--- a/pkg/cli/cmd/login/login.go
+++ b/pkg/cli/cmd/login/login.go
@@ -242,7 +242,10 @@ func adminLogin(
 	password string,
 	insecureTLS bool,
 ) (string, error) {
-	kargoClient := client.GetClient(serverAddress, "", "", insecureTLS)
+	kargoClient := client.GetClient(serverAddress, &client.Options{
+		Credentials:      "",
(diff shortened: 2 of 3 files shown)


krancour:
@onasser1 at a glance, this looks pretty good. I will review more thoroughly as soon as I am able.

From #3537:

> I wonder, How would GCP IAP for example use that proxy-authorization from kargo?

I really don't know much about GCP IAP, and it's not a rabbit hole I am willing to fall into right now, but as I mentioned in #3537, the Proxy-Authorization header is formally defined in [rfc7235](https://datatracker.ietf.org/doc/html/rfc7235#section-4.4). It's generally useful and not only an IAP thing.

If the header's verifiably being set, I think the objective has been achieved.

krancour:
> If the header's verifiably being set, I think the objective has been achieved.

Although, now that I've said this, I notice you're allowing this to be specified at login time and you are persisting it in Kargo's config file, but I see you're _not_ yet incorporating this value into into the `Proxy-Authorization` header of the outbound requests, so it seems you may still have some work to do here.

pkg/cli/client/client.go
```diff
@@ -55,6 +55,7 @@ func GetClientFromConfig(
 func GetClient(
 	serverAddress string,
 	credential string,
+	proxyCredential string,
 	insecureTLS bool,
```
krancour: As the number of optional details here grows, it's probably better that we think about introducing a `ClientOptions` struct. The signature would change to:

```
func GetClient(
	serverAddress string,
	opts *ClientOptions,
) svcv1alpha1connect.KargoServiceClient
```

It should be tolerant of nil options.

krancour:
@onasser1 this seems to be on the right track. One small bit of feedback: https://github.com/akuity/kargo/pull/5229/files#r2466991433

@fuskovic do you mind giving this a review after the above is accounted for?

@steve-marmalade, this feature was your request. If you were willing to test it against your use case, it would be a big help.

pkg/cli/client/client.go
fuskovic: I was going to suggest functional options here before I saw your comment. 

e.g. `WithCredentials`, `WithProxyCredentials`

```
func GetClient(
	serverAddress string,
	opts ...ClientOption,
) svcv1alpha1connect.KargoServiceClient
```

Thoughts?

pkg/cli/client/client.go
```diff
@@ -66,17 +67,47 @@ func GetClient(
 		}
 		httpClient.Transport = transport
 	}
+	// Both credentials are empty
+	if credential == "" && proxyCredential == "" {
+		return svcv1alpha1connect.NewKargoServiceClient(httpClient, serverAddress)
+	}
+
+	if credential != "" && proxyCredential == "" {
```
krancour: All the logic that follows this can be greatly simplified to something like this:

```golang
interceptor := &authInterceptor{}
if credential == "" {
	interceptor.credential = credential
}
if proxyCredential == "" {
	interceptor.proxyCredential = proxyCredential
}
return svcv1alpha1connect.NewKargoServiceClient(
	httpClient,
	serverAddress,
	connect.WithClientOptions(connect.WithInterceptors(interceptor)),
)
```

pkg/cli/client/client.go
krancour: It's doable, of course, but I'm not sure I like it here. If you look at what the `GetClient()` function is configuring, it's an `http.Client` as well as options for the `svcv1alpha1connect.KargoServiceClient`. Options functions would need to pass around both of these. It's not pretty and if you wanted to tidy it up, you'd need to create some kind of options struct that gets passed around instead, and then we're more or less back to where we started. And because we're not really developing an API here, but simply trying to stop an argument list for something used internally from sprawling, I think we can skip this. It's tidy enough to bind flags directly to option struct fields and call it a day.

pkg/cli/client/client.go
```diff
@@ -16,7 +16,9 @@ import (
 )
 
 type Options struct {
-	InsecureTLS bool
+	Credentials      string
+	ProxyCredentials string
+	InsecureTLS      bool
```
naseyro, line 21: With your naming `ClientOptions`, did you mean to create a totally new struct? I thought the same and tried to create a new `ClientOptions` struct but go linting has complained about the naming.
```
pkg/cli/client/client.go:22:6: exported: type name will be used as client.ClientOptions by other packages, and that stutters; consider calling this Options (revive)
type ClientOptions struct {}
```
So I tried to use the existent `Options` struct and add the credential fields. I was afraid if this would cause a break to the dependencies, but it didn't.

Shall we go with the `Options` struct or I have to create a new struct with a different name than `ClientOptions`?
@fuskovic

krancour: I forgot an Options struct already existed in this package was all.

krancour: The way you're currently doing things (this particular bit) is good.

pkg/cli/cmd/login/login.go
```diff
@@ -205,6 +208,17 @@ func (o *loginOptions) run(ctx context.Context) error {
...
+	if o.UseProxyAuth {
+		for o.ProxyAuthCredentials == "" {
+			prompt := &survey.Password{
+				Message: "Proxy Authorization token",
+			}
+			if err = survey.AskOne(prompt, &o.ProxyAuthCredentials); err != nil {
+				return err
+			}
+		}
+	}
```
krancour, line 220: You really ought to be collecting this information _before_ any login attempt, because if credentials for a proxy are required, they're require _during login_ as well.

krancour: Any login function that interacts with the Kargo API server will need to use this.

`adminLogin()` would need to use it for its interactions with the API server's "public config" endpoint as well as the login endpoint (where password is exchanged for a bearer token).

`ssoLogin()` would need to use it for its interactions with the API server's "public config" endpoint and _if_ the OIDC identity provider's issuer URL has the same hostname as the Kargo API server, it would need to be used for all interactions with the identity provider as well. This is a rather complex thing to do.

I _think_ `kubeconfigLogin()` probably doesn't require any modifications since it has not interactions with the Kargo API server.

krancour:
After looking at this in greater depth, it's much more complicated than I initially anticipated. Combining that with its security implications, _if_ this gets merged, I'd like to personally be the one signing off on it. Please note, however, that this remains a low priority.

pkg/cli/client/auth.go
```diff
@@ -47,3 +55,9 @@ func setAuthHeader(header http.Header, cred string) {
 		header.Set(authHeaderKey, fmt.Sprintf("Bearer %s", cred))
 	}
 }
+
+func setProxyAuthHeader(header http.Header, proxyCred string) {
+	if proxyCred != "" {
+		header.Set(proxyAuthHeaderKey, fmt.Sprintf("Bearer %s", proxyCred))
```
krancour, line 61: I'm not sure it's a given that the credential is a bearer token.

naseyro: So what value for the `<auth-scheme>` should we use here?

krancour:
See https://github.com/akuity/kargo/issues/3537#issuecomment-3893234462

Issue #5170, opened 2025-10-08
allow force push option on git-push promotion step


gdsoumya:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Currently the `git-push` step according to my discussion with @hiddeco doesn't allow us to force push to the mentioned `targetBranch`, but in some scenarios this might be desirable. For eg. in cases where we do not care about the past git history and just want to push the new changes to the branch, like pushing output of rendered manifests which do not depend on the previous output state.

aryasoni98:
@gdsoumya I’m happy to work on this! Could you please assign the issue to me?

thomastaylor312:
Closed in #5239

PR #5239, opened 2025-10-17
feat: allow force push option on git-push promotion step

aryasoni98:
### **Summary**
Successfully implemented the `force` option for the `git-push` promotion step, allowing users to force push to any target branch. This resolves GitHub issue #5170.

### **Key Changes**
- **Added** `force: boolean` configuration option (default: `false`)
- **Updated** core logic to respect force push setting
- **Enhanced** documentation with examples and warnings
- **Added** comprehensive test coverage
- **Maintained** full backward compatibility

### **Usage**
```yaml
- uses: git-push
  config:
    path: ./out
    targetBranch: main
    force: true  # Force push to overwrite remote history
```

### **Benefits**
- ✅ Solves the original issue completely
- ✅ Safe opt-in with clear warnings
- ✅ Perfect for rendered manifests/generated content
- ✅ All tests pass, production-ready

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
@@ -95,28 +95,3 @@ steps:
     generateTargetBranch: true
 # Open a PR and wait for it to be merged or closed...
 ```
-
-### Force Push for Rendered Manifests
-
-In this example, rendered manifests are pushed with force enabled. This is useful when the rendered output completely replaces the previous state and doesn't depend on any previous commits in the branch.
-
-This pattern is common when using tools like `helm-template` or `kustomize-build` to generate Kubernetes manifests, where each promotion generates a fresh set of manifests that should completely replace what was previously in the branch.
-
-```yaml
-steps:
-# Clone, render manifests, etc...
-- uses: helm-template
-  config:
-    chart: ./charts/my-app
-    values: ./values/staging.yaml
-    outPath: ./out
-- uses: git-commit
-  config:
-    path: ./out
-    message: rendered updated manifests for staging
-- uses: git-push
-  config:
-    path: ./out
-    targetBranch: staging
-    force: true  # Force push to completely replace branch content
-```
--- a/pkg/x/promotion/runner/builtin/zz_config_types.go
+++ b/pkg/x/promotion/runner/builtin/zz_config_types.go
@@ -219,6 +219,11 @@ type GitOpenPRConfig struct {
 }
 
 type GitPushConfig struct {
+	// Whether to force push to the target branch, overwriting any existing history. This is
+	// useful for scenarios where you want to completely replace the branch content (e.g.,
+	// pushing rendered manifests that don't depend on previous state). Use with caution as this
+	// will overwrite any commits that exist on the remote branch but not in your local branch.
+	Force bool `json:"force,omitempty"`
 	// Indicates whether to push to a new remote branch. A value of 'true' is mutually exclusive
 	// with 'targetBranch'. If neither of these is provided, the target branch will be the
 	// currently checked out branch.
@@ -238,11 +243,6 @@ type GitPushConfig struct {
 	// The target branch to push to. Mutually exclusive with 'generateTargetBranch=true'. If
 	// neither of these is provided, the target branch will be the currently checked out branch.
 	TargetBranch string `json:"targetBranch,omitempty"`
-	// Whether to force push to the target branch, overwriting any existing history. This is
-	// useful for scenarios where you want to completely replace the branch content (e.g.,
-	// pushing rendered manifests that don't depend on previous state). Use with caution as this
-	// will overwrite any commits that exist on the remote branch but not in your local branch.
-	Force bool `json:"force,omitempty"`
 }
 
 type GitWaitForPRConfig struct {


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -34,6 +34,7 @@ Stages that write to the same branch do not write to the same files.
...
 | `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of 'true' is mutually exclusive with `targetBranch`. If neither of these is provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
+| `force` | `boolean` | N | Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). **Use with caution** as this will overwrite any commits that exist on the remote branch but not in your local branch. When `force: true`, the step will skip the pull/rebase operation to allow overwriting remote history. Default is `false`. |
```
krancour: > This is useful for scenarios where you want to completely replace the branch content

I'm sure legitimate uses will be found for the `force` option, but I don't want _this_ use case used as the canonical example because there's a safer, and well-established way of addressing it by using `git-clear` to wipe out all branch contents before applying other steps.

Is there some other, realistic example of how this might be used?

cc @gdsoumya what exactly had you been hoping to use this for?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -95,3 +96,30 @@ steps:
...
+- uses: git-commit
+  config:
+    path: ./out
+    message: rendered updated manifests
+- uses: git-push
+  config:
+    path: ./out
+    targetBranch: main
+    force: true
+```
```
krancour: Let's see where we land on [this](https://github.com/akuity/kargo/pull/5239/files#r2440627108) first, but it's quite possible that this option doesn't rise to the level of needing its own section in the docs. The description in the table seems as if it may be adequate.

pkg/promotion/runner/builtin/git_pusher_test.go
```diff
@@ -126,6 +126,28 @@ func Test_gitPusher_convert(t *testing.T) {
...
+			},
+		},
+		{
+			name: "force with targetBranch",
+			config: promotion.Config{ // Should be completely valid
+				"path":         "/fake/path",
+				"targetBranch": "fake-branch",
+				"force":        true,
+			},
+		},
```
krancour: You might consider removing this tests case. According to the schema, the value of `targetBranch` doesn't have any implications for that is or isn't an acceptable value for the new field. Nor does the new field have any implications for what is an acceptable value of the `targetBranch` field -- meaning this new test case isn't covering anything not already covered by other cases.

pkg/promotion/runner/builtin/git_pusher_test.go
```diff
@@ -250,3 +272,120 @@ func Test_gitPusher_run(t *testing.T) {
...
+	require.Equal(t, "main", branchName)
+	expectedCommit, err := workTree.LastCommitID()
+	require.NoError(t, err)
+	actualCommit, ok := res.Output[stateKeyCommit]
+	require.True(t, ok)
+	require.Equal(t, expectedCommit, actualCommit)
+	expectedCommitURL := fmt.Sprintf("%s/commit/%s", testRepoURL, expectedCommit)
+	actualCommitURL := res.Output[stateKeyCommitURL]
+	require.Equal(t, expectedCommitURL, actualCommitURL)
+}
```
krancour: This test case actually is not exercising the new functionality. You're setting `force` to `true`, _but there is no existing remote branch with different history._ This test case would have identical results if `force` has been set to `false`.

Since the modification to the git pusher aren't really doing anything more than copying a config setting over to push options, I wouldn't worry about testing this _here_, but I do notice, that force pushing isn't really covered by tests in `git` package itself. (My own oversight, as it turns out.) If effort is going to be applied to asserting force pushing works as intended, I'd rather it be done there, at that lower level.

krancour:
Thanks @aryasoni98. I made some suggestions about improving docs and tests, but overall this looks pretty good.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -34,6 +34,7 @@ Stages that write to the same branch do not write to the same files.
...
 | `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of 'true' is mutually exclusive with `targetBranch`. If neither of these is provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
+| `force` | `boolean` | N | Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you need to rewrite commit history (e.g., after cleaning up commits, removing sensitive data, or synchronizing divergent histories). **Use with caution** as this will overwrite any commits that exist on the remote branch but not in your local branch. When `force: true`, the step will skip the pull/rebase operation to allow overwriting remote history. Default is `false`. |
```
krancour: Thank you for changing this from the original example as I requested (since it was one that had a better, established approach), but this example is also not really landing.

> removing sensitive data

I _guess_ force push is a way of removing sensitive data from a remote repo if it were _accidentally_ committed and pushed, but it's actually not common. It's disruptive to the branch's history and, assuming the sensitive data were credentials, the leaked credentials will need to be rotated anyway, so the disruption to the history isn't usually justifiable. If if it weren't _accidental_, I'd have some tough questions for that user about their ridiculous promotion process that leaks sensitive data every time it's executed.

As to the rest of the example:

> cleaning up commits

> synchronizing divergent histories

I don't know why a promotion process would need to do either of these things. There might be a legitimate reason, but I'm struggling to come up with what it is. 🤷‍♂️ 

The bottom line is that if we're going to offer an example of how this feature is useful, I'd like it to be a _realistic_ example. I'd actually prefer _only describing the feature and offering no example at all_ over offering unrealistic ones.

It may or may not have been obvious, but @gdsoumya, who opened #5170 is a colleague and maintainer, so I'm fairly certain he encountered a need for this option internally. In #5170, he wrote:

> in cases where we do not care about the past git history and just want to push the new changes to the branch, like pushing output of rendered manifests which do not depend on the previous output state

As previously suggested, I think it would be useful for him to expand on what actual use case he found that prompted him to request this feature.

pkg/promotion/runner/builtin/git_pusher.go
```diff
@@ -134,14 +134,22 @@ func (g *gitPushPusher) run(
...
+		// Allow force push if explicitly requested in config
+		Force: cfg.Force,
 	}
 	// If we're supposed to generate a target branch name, do so.
 	if cfg.GenerateTargetBranch {
 		// TargetBranch and GenerateTargetBranch are mutually exclusive, so we're
 		// never overwriting a user-specified target branch here.
 		pushOpts.TargetBranch = fmt.Sprintf("kargo/promotion/%s", stepCtx.Promotion)
+		// Always force push for generated branches to ensure they can be overwritten
 		pushOpts.Force = true
```
krancour, line 155: The original comment could have been better, but the new one isn't really adding any clarity. I would change this:

```suggestion
		pushOpts.TargetBranch = fmt.Sprintf("kargo/promotion/%s", stepCtx.Promotion)
		// Since the name of the generated branch incorporates the Promotion's name,
		// which itself incorporates a UUID, we assume this branch did not exist in
		// the remote repository prior to this Promotion. If it somehow does, the
		// only practical explanation for that would be that, for some reason, the
		// entire promotion process has restarted from step zero AFTER having
		// executed this step successfully on a prior attempt. (This can happen,
		// for instance, if the controller were restarted mid-promotion.) Enabling
		// the force push option here prevents this step from failing under those
		// circumstances, and as long as the reasonable assumption that this branch
		// is specific to this Promotion only holds, it is also safe to do this.
		pushOpts.Force = true
```

pkg/promotion/runner/builtin/schemas/git-push-config.json
```diff
@@ -28,6 +28,11 @@
       "type": "string",
       "description": "The name of the Git provider to use. Currently 'azure', 'bitbucket', 'gitea', 'github', and 'gitlab' are supported. Kargo will try to infer the provider if it is not explicitly specified.",
       "enum": ["azure", "bitbucket", "gitea", "github", "gitlab"]
+    },
+    "force": {
+      "type": "boolean",
+      "description": "Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). Use with caution as this will overwrite any commits that exist on the remote branch but not in your local branch. When true, the step will skip the pull/rebase operation to allow overwriting remote history.",
```
krancour: Just a reminder to update this after we figure out what we're putting [here](https://github.com/akuity/kargo/pull/5239/files#r2443347028).

I also wouldn't mention the pull rebase because that's not an option that's exposed to the user. It's an internal implementation detail.

krancour:
@aryasoni98 just an update on this. Last I had looked, the implementation looked ok. I'm pumping the brakes on this because I remain worried that this feature provides a pretty big foot gun without any realistic use for the feature having yet been described. I previously made comments to that effect a few times, noting that I wanted @gdsoumya (my colleague who first requested this feature) to weigh in with his intended use for this. I don't know why you've marked those comments as resolved without us having heard from him... I still want to hear from him.

aryasoni98:
> @aryasoni98 just an update on this. Last I had looked, the implementation looked ok. I'm pumping the brakes on this because I remain worried that this feature provides a pretty big foot gun without any realistic use for the feature having yet been described. I previously made comments to that effect a few times, noting that I wanted @gdsoumya (my colleague who first requested this feature) to weigh in with his intended use for this. I don't know why you've marked those comments as resolved without us having heard from him... I still want to hear from him.

Got it, @krancour totally understand. The intent was to support automated workflows (like generated manifests) where a clean overwrite is required. The force flag stays false by default to avoid risk. I’ve reached out to @gdsoumya so he can clarify the original use case before we move forward.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
thomastaylor312: I synced up with @hiddeco and @krancour. There is definitely a use case for this feature. Anything that is opening a PR could, correctly, want to overwrite what is in that PR. Imagine a workflow like this:

1. Promote something
2. Look at PR and realize the manifest doesn't look quite right
3. Make change and promote again

If we don't have this feature, that type of workflow has to manually close a PR/delete a branch. So this should be ok to do

thomastaylor312 (approved):
I rebased this and removed a section of documentation @krancour said was not needed. I did validate that a force push works correctly

PR #5242, opened 2025-10-17
fix(controller): Avoid panic when config is not set

thomastaylor312:
When I accidentally munged a step config recently and it was omitted, it triggered a step panic. This makes sure to check if a step's `Config` field is not `nil` before trying to do anything with it


hiddeco:
It should be easy to backport this, only the `pkg/promotion/promotion.go` change is 1.9 only — and would probably require a likewise change in the Promotion reconciler for 1.8.

I can arrange the backport on Monday.

PR #5244, opened 2025-10-18
feat: add ui_for_url expression function for generating Kargo UI URLs (#4585)

### LINKED ISSUE #4585: expose a way to build a url for a given resource
# Checklist

# Proposed Feature

Expose a way to build a URL for a given resource, such as exposing `ctx.UIBaseURL`, or a `url_for` expr function

# Motivation

I want to build slack messages for deployment notification and include links without having to hardcode the url host in the message.

Example uses:

"Freight `${{ ctx.targetFreight.name }}` is being promoted to Stage <${{ ctx.UIBaseURL }}/project/${{ ctx.project }}/stage/${{ ctx.stage }}|${{ ctx.stage }}>"

"Freight `${{ ctx.targetFreight.name }}` is being promoted to Stage <${{ url_for(ctx.stage) }}|${{ ctx.stage }}>" (possibly stage_url_for if they need to be differentiated by type to work properly in expr?)

Gkemhcs:
## 🎯 Overview

This Pr Resolves #4585 , implements the `ui_for_url()` expression function as requested in [issue #4585](https://github.com/akuity/kargo/issues/4585) to enable dynamic URL generation for Kargo UI resources in promotion expressions.

### **Core Implementation**
- **Added `UIForURL` function** in `pkg/expressions/function/functions.go`
  - Supports zero or one argument: `ui_for_url()` or `ui_for_url(stageName)`
  - Uses closure pattern similar to `Secret` function for context access
  - Returns actual URLs, not templates

### **Context Integration**
- **Added `UIBaseURL` to promotion context** in `pkg/promotion/evaluator.go`
  - Makes `ctx.UIBaseURL` available in all expressions
  - Integrated with existing expression evaluation pipeline

### **Function Organization**
- **Created `PromotionOperations`** to group promotion-specific functions
  - Follows same pattern as `DataOperations` and `FreightOperations`
  - Added to all expression evaluation points

## 📝 Usage Examples

```yaml
# Using current stage from promotion context
"Freight ${{ ctx.targetFreight.name }} is being promoted to Stage <${{ ui_for_url() }}|${{ ctx.stage }}>"

# Using specific stage name
"Freight ${{ ctx.targetFreight.name }} is being promoted to Stage <${{ ui_for_url('production') }}|${{ ctx.stage }}>"

# Direct usage in Slack notifications
"Check the stage: ${{ ui_for_url() }}"
"Check production: ${{ ui_for_url('production') }}"
```

**Resolves:** [issue #4585](https://github.com/akuity/kargo/issues/4585)

**Related:** This implementation addresses the exact use case described in the issue for Slack deployment notifications with dynamic UI links.

### DIFF
--- a/pkg/expressions/function/functions.go
+++ b/pkg/expressions/function/functions.go
@@ -91,9 +91,12 @@ func UtilityOperations() []expr.Option {
 
 // PromotionOperations returns a slice of expr.Option containing functions for
 // promotion-specific operations, such as URL generation.
-func PromotionOperations(uiBaseURL, project, stage string) []expr.Option {
+func PromotionOperations(uiBaseURL, project, stage, freightID, promotion string) []expr.Option {
 	return []expr.Option{
-		UIForURL(uiBaseURL, project, stage),
+		ProjectLink(uiBaseURL, project),
+		StageLink(uiBaseURL, project, stage),
+		FreightLink(uiBaseURL, project, freightID),
+		PromotionLink(uiBaseURL, project, promotion),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/expressions/function/functions.go
```diff
@@ -967,3 +975,36 @@ const (
 func getCacheKey(prefix, project, name string) string {
 	return fmt.Sprintf("%s/%s/%s", prefix, project, name)
 }
+
+// UIForURL returns an expr.Option that provides a `ui_for_url()` function for use in expressions.
+func UIForURL(uiBaseURL, project, stage string) expr.Option {
+	return expr.Function(
+		"ui_for_url",
```
krancour: We don't use snake case for functions anywhere in else in the expressions. This should be lower camel case.

pkg/expressions/function/functions.go
```diff
@@ -967,3 +975,36 @@ const (
...
+			if !ok {
+				return nil, fmt.Errorf("argument must be string, got %T", a[0])
+			}
+			if stageName == "" {
+				return nil, fmt.Errorf("stage name must not be empty")
+			}
+			targetStage = stageName
+		}
+
+		return fmt.Sprintf("%s/project/%s/stage/%s", uiBaseURL, project, targetStage), nil
```
krancour, line 1041: This appears to only work for Stages.

We'd like to be able to do the same for Projects, Promotions, and Freight.

I'd like there to be separate functions for each of those.

Gkemhcs: Ok sure @krancour  will make it work for others also , and is below conventions ok  for you while calling in yaml templates ?

**Stage:-**
{{ui_for_stage()}}

**Freight:-**
{{ui_for_freight()}}
 If not, how should be the function  names in yaml template.

hiddeco: I wonder if something like `linkForStage` and `linkForFreight` would be a nicer fluent API.

Gkemhcs: I mean what should be the names of link functions to be defined or used in yaml  promotion task template  yaml file .
 
@hiddeco @krancour

krancour: @hiddeco already suggested good names.

Gkemhcs: Ok sure , will implement accordingly

Gkemhcs: @krancour @hiddeco , able to implement the links for projects,stages, freights and let me know how can i construct the url for promotions as it can't be inferred from ui ?

besides can i proceed to add link for warehouse function also?

krancour: > let me know how can i construct the url for promotions as it can't be inferred from ui ?

Not sure what the UI has to do with this. That's not the _source_ of any information.

As for how to actually get the Promotion's name, make the internal function that registers the expression function receive it as a parameter (same as you do with Project, Stage, etc.) and the expression function can close over it (same as it does for Project, Stage, etc.)

Gkemhcs: @krancour @hiddeco , now could you please check is this ok or not?

Issue #5240, opened 2025-10-17
proposal: static webhook receiver


krancour:
Related to #4486 and #4408

The generic webhook receiver proposed in #4486 is expression-driven. #4408 was a draft attempt at implementing it, but @hiddeco rightfully raised some issues in an offline discussion. We agreed to table it and come back to it later with a new, TBD design. That's currently in the v1.9.0 milestone, _but I also don't want to rush this and get it wrong._

In an offline conversation with @fuskovic, he mentioned an idea, which iirc, he attributed to @hiddeco: a simple webhook receiver that could be configured to refresh a static list of Warehouses. Post to it and it will refresh them all. Per my understanding, this wouldn't look at headers or payload _at all_. Something that looks like the following would do the trick and would leave room for handling refresh of things besides Warehouses in the future:

```
config:
  refresh:
  - project: kargo-demo # include for cluster-scoped receiver, but omit otherwise
    kind: Warehouse
    names:
    - my-warehouse
    - your-warehouse
```

There _are_ obvious problems with this, however. It will be common that Warehouses in the list having nothing to do with the precipitating event will been unnecessarily refreshed. Artifact discovery is "expensive," so that's not ideal. One might sidestep that by defining many separate receivers with shorter lists of Warehouses -- ideally one each. Obviously, that could get quite onerous, which is exactly why the original idea of an expression-driven generic webhook receiver (#4486) still has a lot of value.

What I'm proposing now is that there's room for both of these approaches. One is _very_ easy to implement and fills a gap quickly and well-enough for the short-term so that we don't need to rush the more strategic, but considerably more complex expression-driven receiver.

@jessesuen @hiddeco @fuskovic how does this sound?

fuskovic:
> It will be common that Warehouses in the list having nothing to do with the precipitating event will been unnecessarily refreshed. Artifact discovery is "expensive," so that's not ideal. One might sidestep that by defining many separate receivers with shorter lists of Warehouses -- ideally one each. 

Alternatively, we could accept a `warehouse-name` URL param where the param must be one of the warehouse names in the list.

> What I'm proposing now is that there's room for both of these approaches. One is very easy to implement and fills a gap quickly and well-enough for the short-term so that we don't need to rush the more strategic, but considerably more complex expression-driven receiver.

I agree that there is room for both approaches and that in having both we cover a lot of surface area for who we can support

aj-hadrian:
This could be a good stop gap for us.

[...]

PR #5268, opened 2025-10-22
feat: static webhook receiver

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5240

### DIFF
--- a/pkg/webhook/external/static.go
+++ b/pkg/webhook/external/static.go
@@ -93,12 +93,23 @@ func (s *staticWebhookReceiver) handleRefresh(w http.ResponseWriter, req *http.R
 	targets := s.rule.Targets
 	if t := req.URL.Query().Get("target"); t != "" {
 		logger.Info("filtering refresh to single target", "target", t)
+		filteredTargets := []kargoapi.StaticWebhookTarget{}
 		for _, rt := range s.rule.Targets {
 			if rt.Name == t {
-				targets = []kargoapi.StaticWebhookTarget{rt}
+				filteredTargets = append(filteredTargets, rt)
 				break
 			}
 		}
+		if len(filteredTargets) == 0 {
+			xhttp.WriteErrorJSON(w,
+				xhttp.Error(
+					fmt.Errorf("query param target %q is not specified in target list", t),
+					http.StatusBadRequest,
+				),
+			)
+			return
+		}
+		targets = filteredTargets
 	}
 
 	if failures, err := refreshTargets(ctx, s.client, targets); err != nil {
--- a/pkg/webhook/external/static_test.go
+++ b/pkg/webhook/external/static_test.go
@@ -124,6 +124,68 @@ func TestStaticHandler(t *testing.T) {
 			},
 		},
 		{
+			name: "singled out warehouse not specified in target list",
+			client: fake.NewClientBuilder().WithScheme(testScheme).WithObjects(
+				&kargoapi.Warehouse{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: testProjectName,
+						Name:      "fake-warehouse",
+					},
+					Spec: kargoapi.WarehouseSpec{
+						Subscriptions: []kargoapi.RepoSubscription{{
+							Image: &kargoapi.ImageSubscription{
+								RepoURL: "somegenerichost.com/repo",
+							},
+						}},
+					},
+				},
+				// even though the warehouse exists, it is not in the target list
+				&kargoapi.Warehouse{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: testProjectName,
+						Name:      "other-fake-warehouse",
+					},
+					Spec: kargoapi.WarehouseSpec{
+						Subscriptions: []kargoapi.RepoSubscription{{
+							Image: &kargoapi.ImageSubscription{
+								RepoURL: "somegenerichost.com/otherrepo",
+							},
+						}},
+					},
+				},
+			).WithIndex(
+				&kargoapi.Warehouse{},
+				indexer.WarehousesBySubscribedURLsField,
+				indexer.WarehousesBySubscribedURLs,
+			).Build(),
+			req: func() *http.Request {
+				testURLWithQuery := testURL + "?target=other-fake-warehouse"
+				return httptest.NewRequest(
+					http.MethodPost,
+					testURLWithQuery,
+					bytes.NewBuffer([]byte(`{"foo":"bar"}`)),
+				)
+			},
+			rule: kargoapi.StaticWebhookRule{
+				Action: kargoapi.StaticWebhookActionRefresh,
+				Targets: []kargoapi.StaticWebhookTarget{
+					{
+						Name:      "fake-warehouse",
+						Namespace: testProjectName,
+						Type:      kargoapi.StaticWebhookTargetTypeWarehouse,
+					},
+				},
+			},
+			assertions: func(t *testing.T, rr *httptest.ResponseRecorder) {
+				require.Equal(t, http.StatusBadRequest, rr.Code)
+				require.JSONEq(
+					t,
+					`{"error":"query param target \"other-fake-warehouse\" is not specified in target list"}`,
+					rr.Body.String(),
+				)
+			},
+		},
+		{
 			name: "all targeted warehouses refreshed",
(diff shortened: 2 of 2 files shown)


api/v1alpha1/project_config_types.go
```diff
@@ -342,6 +345,76 @@ type AzureWebhookReceiverConfig struct {
...
+
+	// Rules specifies the rule that define actions to be taken when
+	// webhook events are received.
+	//
+	// +kubebuilder:validation:Required
+	Rule StaticWebhookRule `json:"rule,omitempty" protobuf:"bytes,2,opt,name=rule"`
+}
+
+// StaticWebhookAction represents an action to be performed in response to a webhook event.
+type StaticWebhookAction string
```
krancour, line 372: You should use a `+kubebuilder:validation:Enum` marker here to constrain the allowed values.

api/v1alpha1/project_config_types.go
```diff
@@ -342,6 +345,76 @@ type AzureWebhookReceiverConfig struct {
...
+	//  StaticWebhookTargetType indicates that the target is a Warehouse.
+	StaticWebhookTargetTypeWarehouse StaticWebhookTargetType = "Warehouse"
+)
+
+// StaticWebhookTarget represents a target to which an action should be applied.
+type StaticWebhookTarget struct {
+	// Type specifies the type of the target.
+	//
+	// +kubebuilder:validation:Enum=Warehouse
+	Type StaticWebhookTargetType `json:"type" protobuf:"bytes,1,opt,name=type"`
```
krancour, line 394: It would be more idiomatically k8s to call this `Kind`.

api/v1alpha1/project_config_types.go
```diff
@@ -342,6 +345,76 @@ type AzureWebhookReceiverConfig struct {
...
+	// which implicitly serves as a shared secret.
+	//
+	// +kubebuilder:validation:Required
+	SecretRef corev1.LocalObjectReference `json:"secretRef" protobuf:"bytes,1,opt,name=secretRef"`
+
+	// Rules specifies the rule that define actions to be taken when
+	// webhook events are received.
+	//
+	// +kubebuilder:validation:Required
+	Rule StaticWebhookRule `json:"rule,omitempty" protobuf:"bytes,2,opt,name=rule"`
```
krancour, line 368: Comment says "Rules," but the field name is "Rule".

I'm unsure whether the comment should be fixed or whether the field should be changed to be of type `[]StaticWebhookRule`.

Do we want only one rule per receiver?

On the one hand, if only one rule is allowed, you'll find yourself defining a larger number of receivers. On the other hand, if multiple are allowed, and the list is long, many unnecessary actions are more likely to be taken.

I don't think I (yet?) have a strong opinion one way or the other, but would like to hear your thoughts @fuskovic.

api/v1alpha1/project_config_types.go
```diff
@@ -342,6 +345,76 @@ type AzureWebhookReceiverConfig struct {
...
+
+	// Name specifies the name of the target.
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name" protobuf:"bytes,2,opt,name=name"`
+
+	// Namespace specifies the namespace of the target.
+	//
+	// +kubebuilder:validation:Required
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,3,opt,name=namespace"`
```
krancour, line 404: We spoke about this briefly offline.

The existence of this field seems awkward.

If this receiver is defined within a `ProjectConfig` resource, this field isn't needed. (And fields that exist but must not be used are definitely awkward.)

If the receiver is defined within a `ClusterConfig` resource, the field may serve a purpose, but it begs the question of how sensible/realistic it is to establish a cluster-scoped receiver that will take actions on different resources across many different Projects. Exploring that further...

If this were a GitHub receiver, it's sensible to define at the cluster level. The operator is ostensibly responsible for cluster-level configuration and may have decided they'd like to send all webhooks from their GitHub org to this one receiver, with the knowledge that the receiver is smart enough to figure out what specific actions need to be taken on what specific resources, all based on details of the request. Few, if any, actions would be taken unnecessarily.

In contrast, the static receiver is _indiscriminate_ about which resources it acts on and when. As an operator, would I ever really want to tell the admin of Project A, "send your webhooks here and I'll trigger refreshes of Warehouse A1" while _also_ telling the admin of Project B "send your webhooks _to this same place_ and I'll trigger refreshes of Warehouse B1?" The effect will be A1 is refreshed unnecessarily whenever something Project B-related posts to the receiver and B1 will be refreshed unnecessarily whenever something Project A-related posts to the receiver. It would probably be reasonable to assume that the admins of neither Project A nor B would be very happy about this arrangement. They would very quickly get to, "Can't I just define my own, dedicated static receiver?" And yes they can -- at the Project level...

So I think what this ultimately brings me around to... I don't think the static receiver serves any legitimate purpose at the cluster-level, and in fact probably represents a foot bazooka.

Let's ditch the `Namespace` field here and also ditch the very idea of cluster-level static receivers altogether.

This raises a new question of how to prevent cluster-level static receivers from being defined. Both the Project-level and Cluster-level receivers use the _same_ `WebhookReceiverConfig` type. We could _possibly_ break that up into `ProjectWebhookReceiverConfig` and `ClusterWebhookReceiverConfig` where the latter lacks a field that the former has. On the k8s end, this would be ok. It might break the service/v1alpha1 API though. 😬 We should try and determine for sure whether it would or wouldn't. If it does, our only recourse might be additional validation rules only for cluster-level webhook receivers that disallow the use of the static receiver. (Although it would still be a little awkward because it's a different case of "you must not use this field.")

Last, a lot of this also has me thinking this is a case where we should declare this entire feature beta until we've shaken down the UX over the course of a few release and are satisfied with it. (In fact, if we were to find that the expression-driven webhook receiver, which might not land in v1.9.0 depending on other priorities, handily addresses the use cases the static receiver might have been used for, we might elect to drop the static receiver entirely, and being in a beta state would make it that much easier to do so.)

api/v1alpha1/project_config_types.go
krancour: Ah. I see you've put this marker on the field itself. Has the same effect, afaik. Carry on.

api/v1alpha1/project_config_types.go
```diff
@@ -342,6 +345,76 @@ type AzureWebhookReceiverConfig struct {
...
+
+// StaticWebhookRule defines a rule for processing incoming webhook events.
+type StaticWebhookRule struct {
+	// Action specifies the action to be performed when the rule matches an incoming webhook event.
+	//
+	// +kubebuilder:validation:Enum=Refresh
+	Action StaticWebhookAction `json:"action" protobuf:"bytes,1,opt,name=action"`
+
+	// Targets specifies the list of targets to which the action should be applied.
+	Targets []StaticWebhookTarget `json:"targets,omitempty" protobuf:"bytes,2,rep,name=targets"`
```
krancour, line 415: I have some concern that this might not be adequately future-proof. The existence of the `Action` field is clearly leaving room for _other_ actions in the future, but at the same time, pairing that with a `Targets` field suggests that possible actions will be limited to things that can be accomplished by _only_ naming the resources they need to be performed on.

I'll use a completely made up example to illustrate the point. Let's say we have an action "Pause" that, unsurprisingly, pauses reconciliation of the named resources. Where or how is the duration of the pause specified?

As in similar cases in the past, this leads me away from the idea of a field like `Action` and toward optional fields with (mutually exclusive? not sure) action-specific configuration. Something that would look like this:

```
kind: ProjectConfig
metadata:
  name: kargo-demo
  namespace: kargo-demo
spec:
  webhookReceivers: 
  - name: my-receiver
    static:
      secretRef:
        name: my-secret
      actions:
      - refresh:
          resources:
          - kind: Warehouse
            name: my-warehouse
          - kind: Warehouse
            name: my-other-warehouse
      - pause:
          resources:
          - kind: Stage
            name: dev
          duration: 5m
      - pause:
          resources:
          - kind: Stage
            name: prod
          duration: 1h

```

krancour:
Closing this based on recent session between @fuskovic @hiddeco and myself. I think we came up with a workable plan for the expression-driven generic webhook receiver that also supports static lists of resources to apply actions to quite well. It seems it may be easy enough to implement that this half-measure is probably no longer called for.

Issue #5270, opened 2025-10-22
untar promotion step fails when outPath directory does not exist


kejne:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

When running the `untar` step, it fails if the `outPath` directory is a path where the parent directory does not exist either with a `failed to move` error. See logs

# Steps to Reproduce

Download a chart using the `download-oci` and try to `untar` it into a directory which does not yet exist and where the parent directory does not exist either.

```yaml
...
  steps:
    - as: argocdAws-download
      config:
        imageRef: ${{ vars.argocdAws }}
        outPath: ./tmp/${{ vars.deployment }}/in-target/charts/apps/argocdAws.tgz
      uses: oci-download
    - as: argocdAws-template
      config:
        outPath: ./out/${{ vars.deployment }}/apps/argocdAws
        path: ./tmp/${{ vars.deployment }}/in-target/charts/apps/argocdAws.tgz
        releaseName: argocdaws-${{ vars.deployment }}
        valuesFiles:
      uses: helm-template

# This is successful as the templating before creates the parent directory
    - as: argocdAws-untar
      config:
        inPath: ./tmp/${{ vars.deployment }}/in-target/charts/apps/argocdAws.tgz
        outPath: ./out/${{ vars.deployment }}/apps/argocdAwsHelm
      uses: untar
-     
    - as: cilium-download
      config:
        imageRef: ${{ vars.cilium }}
        outPath: ./tmp/${{ vars.deployment }}/in-target/charts/apps/cilium.tgz
      uses: oci-download

# This FAILS since it's a new parent directory
    - as: cilium-untar
      config:
        inPath: ./tmp/${{ vars.deployment }}/in-target/charts/apps/cilium.tgz
        outPath: ./out/${{ vars.deployment }}/apps/helm/cilium
      uses: untar
```

# Version

```
v1.7.5
```

# Logs

```
[...]

PR #5279, opened 2025-10-23
fix(untar): always create parent directory

kejne:
Fixes: https://github.com/akuity/kargo/issues/5270


hiddeco (approved):
LGTM! Thank you @kejne 🙇

PR #5293, opened 2025-10-27
feat: implement helper for evaluating whether current promo step is running

fuskovic:
Adds a helper for marking promo step to encapsulate some complex boolean logic

### DIFF
--- a/pkg/api/promotion.go
+++ b/pkg/api/promotion.go
@@ -158,8 +158,7 @@ func ComparePromotionPhase(a, b kargoapi.PromotionPhase) int {
 	}
 }
 
-// IsCurrentStepRunning returns true if the current step of the given Promotion is
-// running.
+// IsCurrentStepRunning returns true if the promotion is in a running state and the current step is also running.
 func IsCurrentStepRunning(promo *kargoapi.Promotion) bool {
 	return promo.Status.Phase == kargoapi.PromotionPhaseRunning &&
 		int64(len(promo.Status.StepExecutionMetadata)) == promo.Status.CurrentStep+1 &&


pkg/controller/promotions/promotions.go
```diff
@@ -709,3 +704,14 @@ func calculateRequeueInterval(p *kargoapi.Promotion) time.Duration {
...
+// MarkStepAbortedIfRunning updates the newStatus to mark the current step as
+// aborted if the promotion is running and the current step is running.
+func MarkStepAbortedIfRunning(newStatus *kargoapi.PromotionStatus, promo *kargoapi.Promotion, finishedAt *metav1.Time) {
+	if newStatus.Phase == kargoapi.PromotionPhaseRunning &&
+		int64(len(newStatus.StepExecutionMetadata)) == promo.Status.CurrentStep+1 &&
+		promo.Status.StepExecutionMetadata[promo.Status.CurrentStep].Status == kargoapi.PromotionStepStatusRunning {
+		newStatus.StepExecutionMetadata[promo.Status.CurrentStep].Status = kargoapi.PromotionStepStatusAborted
+		newStatus.StepExecutionMetadata[promo.Status.CurrentStep].FinishedAt = finishedAt
+	}
+}
```
krancour: If the intention is to make this reusable in other contexts, then this isn't the right place for it, as it will force awkward dependencies on this package. This should probably be moved to `pkg/api`.

pkg/api/promotion.go
```diff
@@ -157,3 +157,14 @@ func ComparePromotionPhase(a, b kargoapi.PromotionPhase) int {
 		return 0
 	}
 }
+
+// MarkStepAbortedIfRunning updates the newStatus to mark the current step as
+// aborted if the promotion is running and the current step is running.
+func MarkStepAbortedIfRunning(newStatus *kargoapi.PromotionStatus, promo *kargoapi.Promotion, finishedAt *metav1.Time) {
```
thomastaylor312: I think this would be better as a method on a promotion (or takes a promotion, either way) that was something like `IsCurrentStepRunning`. The issue here is to encapsulate this boolean statement in a single place, not necessarily have something that is "aborting a step" specific

pkg/api/promotion.go
```diff
@@ -157,3 +157,11 @@ func ComparePromotionPhase(a, b kargoapi.PromotionPhase) int {
 		return 0
 	}
 }
+
+// IsCurrentStepRunning returns true if the current step of the given Promotion is
+// running.
```
thomastaylor312: For clarity this should say this returns true if the promotion is in a running state and the current step is also running

PR #5297, opened 2025-10-27
more aggressive retries on argocd-update step

krancour:
This change does three things:

1. Allows step runners to specify a suggested requeue interval as part of a result that indicates the step is still "running" (needs to idle and be tried again later).

2. The argocd-update takes advantage of this. It will suggest requeuing the Promotion after 30 seconds (instead of the five minutes the controller would use otherwise). This should provide some relief from the as-yet-unexplained delays in the Promotion controller being notified of status changes in Argo CD Applications, which have occasionally been observed in the wild.

3. Writes additional debug-style logs at an INFO level (temporarily?) to provide greater insight into what goes on inside the argocd-update step.

cc @jessesuen 

Definitely need @hiddeco to double-check these changes.

### DIFF
--- a/pkg/cli/cmd/update/freight.go
+++ b/pkg/cli/cmd/update/freight.go
@@ -73,7 +73,7 @@ func (o *updateFreightAliasOptions) addFlags(cmd *cobra.Command) {
 		cmd.Flags(), &o.Project, o.Config.Project,
 		"The project the freight belongs to. If not set, the default project will be used.",
 	)
-	option.Name(cmd.Flags(), &o.Name, "The name of the freight to to be updated.")
+	option.Name(cmd.Flags(), &o.Name, "The name of the freight to be updated.")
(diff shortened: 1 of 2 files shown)


pkg/controller/promotions/promotions_test.go
```diff
@@ -629,7 +644,62 @@ func Test_calculateRequeueInterval(t *testing.T) {
...
+			},
+			assertions: func(t *testing.T, requeueInterval time.Duration) {
+				// The request should be requeued according to the suggestion.
+				require.Equal(t, time.Minute, requeueInterval)
+				// Sanity check that the requeue interval is always greater than 0.
+				require.Greater(t, requeueInterval, time.Duration(0))
+			},
+		},
+		{
+			name:                     "timeout occurs before next default interval",
```
krancour:
```suggestion
			name:                     "timeout occurs before next suggested interval",
```

PR #5298, opened 2025-10-28
feat: allow opting into caching image metadata by tag

### LINKED ISSUE #5076: add options to opt into more aggressive image metadata caching (options to cache by tag)
### Discussed in https://github.com/akuity/kargo/discussions/5055

<sup>Originally posted by **nissessenap** September 15, 2025</sup>
Hi

I was very close to migrating from ArgoCD image-updater over to Kargo last week, but sadly we have to cancel that.
The reason is simple, we are using NewestBuild, and it just takes too long time to get images.
At least when I added 250+ warehouses.

I know that you document multiple times that you shouldn't use [newest-build](https://docs.kargo.io/user-guide/how-to-guides/working-with-warehouses/#newest-build), but sadly our current naming standard forces us to use it. Sure we could change our way of working, but as always that takes time...

I spent the morning looking at the image-updater code and the Kargo code and the reason is rather simple, tags are not [cached](https://github.com/nissessenap/kargo/blob/27b4c8c72ee199b148ccd5e541e861f8c0d606e0/internal/image/repository_client.go#L154-L155).

> getImageByTag retrieves an Image by tag. This function uses no cache since tags can be mutable.

I fully understand this reasoning but in many container registries you can disable mutability of tags, and even if you can't, you probably own the container life-cycle within an org so you make sure you don't change tags.

## Suggestion

The suggestion is simple, based on a feature flag/env variable cache image tags.

## Options

My current understanding of webhooks is that they are just a way to tell the warehouse to trigger perform a pool, I assume this is because you need to verify filters etc.
One option could when you are using NewestBuild to assume that the tag that was sent actually is the newst and just use it (of course based on the tag matches your rules).
</div>

---

More context: https://github.com/akuity/kargo/discussions/5055#discussioncomment-14458098

krancour:
Fixes #5076

Highlights:

1. Creates a new, very simple cache abstraction with one implementation based on `github.com/hashicorp/golang-lru/v2`. This replaces `https://github.com/patrickmn/go-cache`. `github.com/hashicorp/golang-lru/v2`, being, as the name implies, an LRU cache, it is better suited to the rest of the changes in this PR.

1. Before, each registry got its own cache. Now there's just one shared cache. Its size is configurable. By default, it holds up to 100,000 entries. Each entry is pretty small.

1. Container image subscriptions can now opt into using cached tags. When choosing this, all image selection strategies, except Digest, will cache image information by tag and _not_ by digest. If one does not opt into this, things work the same as they always have.

1. Operators can set policies that allow/disallow using cached tags or even _require_ cached tags. This way operators who are not confident that users are working with immutable tags can forbid tags from beings cached. Those who are confident that users ARE working with immutable tags can _require_ tag caching.

### DIFF
--- a/docs/docs/40-operator-guide/20-advanced-installation/30-common-configurations.md
+++ b/docs/docs/40-operator-guide/20-advanced-installation/30-common-configurations.md
@@ -487,7 +487,9 @@ tag:
   developers.
 
   :::info
+
   For purposes of backwards compatibility, `Allow` is the default policy.
+
   :::
 
 - `Require`: Container image subscriptions MUST explicitly opt into caching
@@ -502,7 +504,9 @@ tag:
   not opt into caching image metadata by tag will be treated as if they had.
 
   :::info
+
   This is the recommended policy in an immutable-tags-only environment.
+
   :::
 
 Example configuration to allow (but not require) individual container image
--- a/docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
+++ b/docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
@@ -114,21 +114,25 @@ fields:
 
   The default is `false`.
 
-  :::warning Use with caution!
+  :::warning[Use with caution!]
+
   This setting is safest if your tags are known to be "immutable" (i.e., tag
   always references the same image and is never updated to point to a different
   image).
 
   This setting does NOT apply to the `Digest` selection strategy, which assumes
   the one tag it subscribes to is a mutable one.
+
   :::
 
   :::warning
+
   Operators may also choose from a number of policies regarding the caching of
   image metadata using tags as keys. Some of these policies (`Forbid` and
   `Force`) can override an individual container image subscription's choice to
   cache metadata by tag or not. See
   [common configurations](../../40-operator-guide/20-advanced-installation/30-common-configurations.md) for further details.
+
   :::
 
 #### Image Selection Strategies
@@ -248,6 +252,7 @@ strategies are:
   ```
 
   :::tip
+
   If your tags are known to be **immutable** (i.e., a tag always references the
   same image and is never updated to point to a different image), you can use
   the `cacheByTag` field to enable more aggressive caching of image metadata by
@@ -267,10 +272,12 @@ strategies are:
 
   :::
 
-  :::warning Use with caution!
+  :::warning[Use with caution!]
+
   Only enable `cacheByTag` if you are certain that all relevant tags are
   **immutable**. Using this with mutable tags (like `latest`) can cause Kargo
   to select stale images indefinitely.
+
   :::
 
 ### Git Repository Subscriptions
@@ -864,13 +871,15 @@ This enables significantly more aggressive caching of image metadata, which can
 reduce API calls and improve performance by orders of magnitude in repositories
 with large numbers of tags.
 
-:::warning Use with caution!
+:::warning[Use with caution!]
+
 Only enable this option if your tags are known to be **immutable**
 (i.e., a tag always references the same image and is never updated to point
 to a different image).
 
 This setting does not apply to the `Digest` selection strategy, which always
 assumes tags are mutable.
+
 :::
 
 ## Triggering Artifact Discovery Using Webhooks


hiddeco (commented):
In addition to what is being done in this PR, which is excellent, I wonder if this would be a good time to handle the `// TODO` about making the concurrency settings configurable. 

We have had reports internally that adding a zero to the current settings (metadata concurrency and rate limiter) brings down the refresh time from ~15 minutes to just over one.

nikolay-te:
> In addition to what is being done in this PR, which is excellent, I wonder if this would be a good time to handle the `// TODO` about making the concurrency settings configurable.
> 
> We have had reports internally that adding a zero to the current settings (metadata concurrency and rate limiter) brings down the refresh time from ~15 minutes to just over one.

Yes, please 🙏🏻 This made a HUGE difference in our environment. For different reasons we're stuck to using `NewestBuild`, and patching the warehouse shards to have increased rate limits and parallelism helped a lot.

Another potential improvement would be some connection reuse on the client (though it's interesting to see how would this work in a multi-tenant environment), as the other thing we've noticed after bumping these parameters manually is that the querying of the repositories can be very aggressive with multiple tcp connections getting recycled, and also high number of /v2/token requests for docker, etc.

krancour:
> In addition to what is being done in this PR, which is excellent, I wonder if this would be a good time to handle the // TODO about making the concurrency settings configurable.

Are you referring to the rate limit (currently 20 per registry) or the number of goroutines allowed to work on pulling down metadata concurrently (currently capped at 1,000, but shared across all registries)?

Assuming you're referring to the rate limit, I've always been open to making that configurable, but have a strongly held opinions about how it's approached. #1139 is the issue capturing my thoughts about this in more detail. Someone _did_ open a PR not too long ago that was rejected because it only addressed rate limits and not the rest of #1139, which is ok, but it did so in a way that was inadequately future-proof. We'd have been unable to build a comprehensive solution to registry configuration on top of it, which would inevitably have forced a breaking change on us. I'm in no way saying that we cannot make rate limits configurable without addressing the rest of #1139. What I _am_ saying is whatever is done to make rate limits configurable needs to not paint us into a corner in terms of how we'd approach the rest of #1139.

Separate from that, I have some worry that increasing the rate limit isn't the panacea people seem to think it is. There's a natural inclination toward thinking that if you crank that way up, things just get faster. I think there's not enough weight being given to the possibility that when the client isn't self-limiting and the rate limits are being enforced server-side instead, things might actually get _worse_. This isn't a reason not to do it. It's just a concern.

Between my concern and my opinions about there being a (somewhat involved) right way of approaching this, my inclination is to not bundle that improvement into this cache improvement.

Why don't we see about getting @jessesuen's buy-in on the idea of prioritizing #1139 in v1.10.0?

hiddeco:
> Are you referring to the rate limit (currently 20 per registry) or the number of goroutines allowed to work on pulling down metadata concurrently (currently capped at 1,000, but shared across all registries)?

Based on @nikolay-te's report, it is both.

nikolay-te:
> > Are you referring to the rate limit (currently 20 per registry) or the number of goroutines allowed to work on pulling down metadata concurrently (currently capped at 1,000, but shared across all registries)?
> 
> Based on @nikolay-te's report, it is both.

I think it was probably mostly the per repo rate limit that gave us the improvement.
Since it's per registry, and we have only one registry for all images we were seeing image warehouse reconciliations taking up to an hour. This is even when we spread all warehouses on 10 separate kargo shards/instances dedicated to warehouse tasks. Both the Kargo agents and our registry and the caching proxy in front of it were mostly idle, until we increased the rate limits. After that these discoveries that were taking an hour dropped to 1-2 minutes, many were in the seconds.

Having said that, I see now that the caching proxies (nginx, passing through tag list operations, caching only manifests) I've setup is able to hold ALL images in our environment mostly in memory (just below 2G RAM) which makes me think this caching feature has the potential to bring similar improvements maybe even with the default rate limits, except maybe on cold starts.

krancour:
@nikolay-te there's a lot of interesting insight there. Thank you.

Rate limit-related improvements are coming, but they won't be bundled into this PR.

pkg/image/repository_client.go
```diff
@@ -157,7 +162,25 @@ func (r *repositoryClient) getImageByTag(
...
 	platform *platformConstraint,
-) (*image, error) {
+) (*Image, error) {
+	logger := logging.LoggerFromContext(ctx)
+	logger.Trace(
+		"retrieving image",
+		"tag", tag,
+	)
+
+	cacheKey := fmt.Sprintf("%s:%s", r.repoURL, tag)
```
hiddeco, line 161: Should this take the platform constraints into account? Otherwise, I think we could get mismatches?

pkg/image/cache.go
```diff
@@ -0,0 +1,17 @@
+package image
+
+import "github.com/akuity/kargo/pkg/cache"
+
+var imageCache cache.Cache[Image]
+
+func init() {
+	var err error
+	imageCache, err = cache.NewInMemoryCache[Image](100000)
```
hiddeco: Should the size be configurable?

krancour: I thought it was... there might be a commit missing from this PR. Will follow up on that.

pkg/image/repository_client.go
krancour: 🤔 I think you're right. I'll look at that.

pkg/image/cache.go
krancour: ffb1e165da336dc3249f0ba30b3e0bc9bdbd9662 was missing previously. It's configurable now.

pkg/image/repository_client.go
krancour: Yep. Will fix. Converting to draft in the meantime.

krancour: I've resolved this. What we're doing now, in the case of multi-arch images, is we are including metadata for all platforms in a single cache entry.

All subscriptions for a given image will be served by the same cache entry, regardless of whatever platform constraints they may specify, but what is actually returned for a subscription with platform constraints will be a platform-specific representation of what was in the cache.

krancour:
@hiddeco this is ready for another look whenever. There's no rush.

pkg/controller/warehouses/images.go
```diff
@@ -48,6 +48,19 @@ func (r *reconciler) discoverImages(
 			logger.Debug("found no credentials for image repo")
 		}
 
+		if sub.CacheByTag && !r.cfg.AllowCacheByTag {
+			return nil, fmt.Errorf(
+				"image tag caching is disabled by controller configuration",
+			)
+		}
+		if !sub.CacheByTag && r.cfg.RequireCacheByTag {
```
hiddeco: I would kind of expect `r.cfg.RequireCacheByTag` to simply enforce `sub.CacheByTag`. (As in: `sub.CacheByTag = r.cfg.RequireCacheByTag`).

krancour: In many similar situations, I believe I would agree with that. I think this situation may be different.

Here's the scenario I am concerned with:

Some operator says, "Oh ya. All the teams are using immutable tags and I'd really like to not hammer the registries unnecessarily," and decide that this should be required.

But then, as it turns out, some team _isn't_ playing by the rules by which the operator believe everyone had been. They _are_ using mutable tags. If the operator's wishes were silently enforced, this team may be in for a surprise.

By not silently enforcing the operator's wishes, developers end up needing to explicitly acknowledge the behavior / avow that they are not using mutable tags.

hiddeco: Wouldn't it be better to raise this concern then through a warning issued by the API server during apply?

krancour: In a perfect world, yes. Or I'd do it with a validating webhook. I think the problem with both of those is that if an operator starts requiring caching by tag after a warehouse already exists, those approaches won't catch it.

krancour: Considering it further, it doesn't need to be mutually exclusive. I can validate it as you suggest, but also continue doing what we are now to handle the edge case where an operator opt into the policy after Warehousese that don't adhere to it already exist.

Easy fix.

hiddeco: My concern with hard-failing at reconciliation time is that it's a less actionable place to surface this than at apply time. Users discover the problem asynchronously through status conditions rather than getting immediate feedback from an apply.

There's also a subtler issue: when a user sets `cacheByTag: true` just to satisfy the controller policy, we lose the signal of what they actually intended. If the operator later decides to disable `requireCacheByTag`, those Warehouses remain configured with tag caching enabled, even if the user never really wanted that.

This is further complicated by the `allowCacheByTag` option. If an operator first runs with `requireCacheByTag: true`, users are forced to set `cacheByTag: true` on their Warehouses to get them to reconcile. But if the operator later changes their mind and flips to `requireCacheByTag: false` with `allowCacheByTag: false,` all those Warehouses that were coerced into setting `cacheByTag: true` will now start failing. Users who never actually wanted tag caching now have to go back and update all their specs to remove it.

By requiring the setting to be explicit in the spec rather than having the controller enforce the policy transparently, we're encoding policy decisions into user resources, which makes policy changes more disruptive than they need to be.

For handling pre-existing Warehouses that don't comply with a newly-enabled policy, I wonder if a warning event or condition (emitted at a higher level) would work better than blocking reconciliation entirely.

krancour: > when a user sets cacheByTag: true just to satisfy the controller policy, we lose the signal of what they actually intended.

This is true, but if we silently force the policy on them, even with a warning, there's virtually no signal from the operator to the user that they're doing something they should not be, which seems far worse. (I say _virtually_ no signal because people tend not to notice warnings, and if they do, they're seldom given much weight.)

Are some people going to check the box just to comply with the operator's wishes _without_ actually guaranteeing that they're using immutable tags? There's no doubt in my mind that some will, but having checked that box, any unexpected results will be their own fault, which I think is preferable to any unexpected results being _our_ fault.

krancour: Thinking further:

It would be safe to silently enforce it when the admin is prohibiting cache by tag and I am making that change right now.

It's only the inverse scenario of the admin insisting on it and the user not having opted in that remains problematic.

hiddeco: I understand the concern about warnings being ignored, but I'm not sure hard-failing at reconciliation actually produces more informed consent. A user encountering this error is likely to add `cacheByTag: true` to clear the error and move on. They're not necessarily stopping to verify that their tags are actually immutable. It would be more of a "make this error go away" action rather than a meaningful acknowledgement...

thomastaylor312: So I can see the good argument for both approaches here. FWIW, I think I lean a bit more towards what Hidde was saying around having these things surface at apply time. Too many times I've done a k8s thing and then had an issue show up in status and it takes me a second to find (or sometimes way longer than a second depending on what the issue is) if I have to remember that it probably showed up in status and the reason is because I configured a value completely wrong and that could have been caught. 

So regardless of the underlying intent here (explicitly opting in or not in this case), I think the better user experience is to surface at apply time if at all possible because we have the ability to do that without much extra effort (compare that to something like validating an `image` field in a pod where that could require an expensive download to validate)

krancour: @hiddeco I've got us now to where:

1. If operator prohibits caching image metadata by tag, subscriptions that ask to cache by tag have that choice silently overridden at discovery time since, discounting performance impacts, this is safe to do.

1. If operator _requires_ caching image metadata by tag, the Warehouse's existing validating webhook will reject any container image subscription that does not adhere to that policy. The warning text is very explicit about enabling with care.

1. To account for Warehouses that exist prior to an operator hypothetically starting to require caching image metadata by tag, the artifact discovery process will fail if a subscription does not adhere to that policy. Once again, the error text is very clear about enabling with caution.

krancour: Just for the paper trail...

@hiddeco and I spoke offline. I workshopped how different scenarios play out if caching by tag were silently enforced. I've mostly come around to @hiddeco's point of view that the consequences aren't as dire as I imagined originally, _however_, there are some edge cases that leave me still feeling generally uncomfortable with it.

What we agreed on, which is probably a better direction overall anyway, is giving the operator an even finer degree of control over this. Instead of binary choices to allow/forbid or require/not, we will let operators select a "cache by tag policy". Choices:

1. Forbid: Semantically self-explanatory and silently enforced as it is safe to do so.
2. Allow: Semantically self-explanatory and individual container image subscriptions choose whether to cache by tag or not.
3. Require: Always cache by tag, but _require_ subscriptions to explicitly opt into it. This is tantamount to an explicit acknowledgement of the policy. I like to think of it like "TOS." This sacrifices UX a bit for safety's sake.
4. Force: Always cache by tag and silently enforce it. This is a small concession on safety for the sake of better UX.

"Allow" would be the default.

krancour: Accounted for by ccb74dfd3012585f89607cf66bfa4741627df427

docs/docs/40-operator-guide/20-advanced-installation/30-common-configurations.md
```diff
@@ -467,6 +467,65 @@ For a list of resource kinds that can be configured, refer to the
...
+  This is silently enforced. Subscriptions that opt into caching image metadata
+  by tag will be treated as if they had not.
+
+- `Allow` Individual container image subscriptions may choose whether to cache
+  image metadata by tag. This option leaves the decision in the hands of
+  developers.
+
+  :::info
+  For purposes of backwards compatibility, `Allow` is the default policy.
+  :::
```
hiddeco, line 493:
```suggestion
  :::info

  For purposes of backwards compatibility, `Allow` is the default policy.

  :::
```

docs/docs/40-operator-guide/20-advanced-installation/30-common-configurations.md
```diff
@@ -467,6 +467,65 @@ For a list of resource kinds that can be configured, refer to the
...
+  to minimize the possibility of surprise at any stale results from an image
+  discovery process. This option sacrifices some small degree of usability for
+  safety.
+
+- `Force`: Caching image metadata is silently enforced. Subscriptions that to
+  not opt into caching image metadata by tag will be treated as if they had.
+
+  :::info
+  This is the recommended policy in an immutable-tags-only environment.
+  :::
```
hiddeco, line 510:
```suggestion
  :::info

  This is the recommended policy in an immutable-tags-only environment.

  :::
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -108,6 +108,29 @@ fields:
...
+  The default is `false`.
+
+  :::warning Use with caution!
+  This setting is safest if your tags are known to be "immutable" (i.e., tag
+  always references the same image and is never updated to point to a different
+  image).
+
+  This setting does NOT apply to the `Digest` selection strategy, which assumes
+  the one tag it subscribes to is a mutable one.
+  :::
```
hiddeco:
```suggestion
  :::warning[Use with caution!]

  This setting is safest if your tags are known to be "immutable" (i.e., tag
  always references the same image and is never updated to point to a different
  image).

  This setting does NOT apply to the `Digest` selection strategy, which assumes
  the one tag it subscribes to is a mutable one.

  :::
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -108,6 +108,29 @@ fields:
...
+
+  :::warning
+  Operators may also choose from a number of policies regarding the caching of
+  image metadata using tags as keys. Some of these policies (`Forbid` and
+  `Force`) can override an individual container image subscription's choice to
+  cache metadata by tag or not. See
+  [common configurations](../../40-operator-guide/20-advanced-installation/30-common-configurations.md) for further details.
+  :::
```
hiddeco, line 136:
```suggestion
  :::warning

  Operators may also choose from a number of policies regarding the caching of
  image metadata using tags as keys. Some of these policies (`Forbid` and
  `Force`) can override an individual container image subscription's choice to
  cache metadata by tag or not. See
  [common configurations](../../40-operator-guide/20-advanced-installation/30-common-configurations.md) for further details.

  :::
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -224,6 +247,32 @@ strategies are:
         - ^nightly
   ```
 
+  :::tip
+  If your tags are known to be **immutable** (i.e., a tag always references the
```
hiddeco, line 256:
```suggestion
  :::tip

  If your tags are known to be **immutable** (i.e., a tag always references the
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -224,6 +247,32 @@ strategies are:
...
+        - ^nightly
+  ```
+
+  :::
+
+  :::warning Use with caution!
+  Only enable `cacheByTag` if you are certain that all relevant tags are
+  **immutable**. Using this with mutable tags (like `latest`) can cause Kargo
+  to select stale images indefinitely.
+  :::
```
hiddeco:
```suggestion
  :::warning[Use with caution!]

  Only enable `cacheByTag` if you are certain that all relevant tags are
  **immutable**. Using this with mutable tags (like `latest`) can cause Kargo
  to select stale images indefinitely.

  :::
```

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -786,6 +837,42 @@ required for `Warehouse`s to notice new artifacts (of any kind; not just
...
+with large numbers of tags.
+
+:::warning Use with caution!
+Only enable this option if your tags are known to be **immutable**
+(i.e., a tag always references the same image and is never updated to point
+to a different image).
+
+This setting does not apply to the `Digest` selection strategy, which always
+assumes tags are mutable.
+:::
```
hiddeco:
```suggestion
:::warning[Use with caution!]

Only enable this option if your tags are known to be **immutable**
(i.e., a tag always references the same image and is never updated to point
to a different image).

This setting does not apply to the `Digest` selection strategy, which always
assumes tags are mutable.

:::
```

hiddeco (dismissed):
A couple of documentation nits, but the code looks good to me. 🥇

krancour:
> "Add suggestion to batch" seems to be unavailable once a PR is approved. Or at least I think that's the reason it's greyed out. Dismissing only for that reason.

Well, that did NOT do the trick.

krancour:
Ugh. idk why, but the UI shows the button in the "Conversation" tab, but it's greyed out. Over in the "Files changed" tab, it works. Weird UX.

Sorry for the noise.

krancour:
Since there are no changes to this PR since I dismissed @hiddeco's review, I'm using super powers to merge this in good faith.

hiddeco (approved):

PR #5308, opened 2025-10-30
feat(expressions): add semverParse() function for parsing semantic versions

### LINKED ISSUE #5307: add an expression function for parsing semvers
A `semverParse()` function would be useful for parsing semantic version strings.

We already have `semverDiff()`, which returns an indicator of the magnitude of difference between two semantic version strings, but we currently lack a function that would allow you, for instance, to access the major, minor, etc. components of a single semantic version.

The ability to do so would be useful in scenarios where you need to "bump" one of these components.

Deepam02:
Resolves #5307

added  `semverParse()` expression function to parse semantic versions and extract individual components (major, minor, patch, prerelease, metadata) for use in expressions like version bumping workflows.


pkg/expressions/function/functions.go
```diff
@@ -957,6 +974,34 @@ func semverDiff(a ...any) (any, error) {
...
+		return nil, fmt.Errorf("invalid semantic version: %w", err)
+	}
+
+	return map[string]any{
+		"major":      ver.Major(),
+		"minor":      ver.Minor(),
+		"patch":      ver.Patch(),
+		"prerelease": ver.Prerelease(),
+		"metadata":   ver.Metadata(),
+	}, nil
```
krancour: I think it would be reasonable to return the actual `ver` struct instead. Lots of our functions return structs. Once you've called a function form within an expression and have a result in-hand, structs are certainly easier to work with syntax-wise than maps are.

Deepam02: Done!  Changed `semverParse()` to return `*semver.Version` struct directly instead of a map.

pkg/expressions/function/functions.go
```diff
@@ -434,16 +434,15 @@ func SemverDiff() expr.Option {
...
 //
-// The semverParse function parses a semantic version string and returns a map
-// containing its individual components: major, minor, patch, prerelease, and
-// metadata. This is useful for scenarios where you need to inspect or
-// manipulate specific version components, such as bumping the major, minor, or
-// patch version.
+// The semverParse function parses a semantic version string and returns a
+// *semver.Version struct. This allows direct access to version component
+// methods like Major(), Minor(), Patch(), Prerelease(), and Metadata(), as
+// well as utility methods like IncMajor(), IncMinor(), and IncPatch().
```
krancour, line 440: Oh... that's a nice bonus I'd not entirely anticipated. ❤️

pkg/expressions/function/functions_test.go
```diff
@@ -1869,3 +1870,153 @@ func Test_semverDiff(t *testing.T) {
 		})
 	}
 }
+
+func Test_semverParse(t *testing.T) {
```
krancour, line 1874: You've written a lot of test cases here that are re-asserting correct behavior of the `semver.NewVersion()` function. It's a safe bet that they've already tested that quite well. It's not incumbent upon us to duplicate all of that and doing so just adds noise to the codebase without improving coverage.

The only test cases really required here are the basic ones that cover what _this_ function is doing -- checking the number and type of arguments, erroring when `semver.NewVersion()` itself errors and succeeding when it does not.

krancour:
@Deepam02 implementation looks good, but this will need docs as well. You can place them directly above the docs for `semverDiff()` and you can emulate the format of the `semverDiff()` docs as closely as possible.

https://github.com/akuity/kargo/blob/24b5f278f2be704a5557de2f4238728b8ce406ea/docs/docs/50-user-guide/60-reference-docs/40-expressions.md?plain=1#L743

https://docs.kargo.io/user-guide/reference-docs/expressions#semverdiffversion1-version2

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -740,6 +740,66 @@ config:
...
+
+- uses: yaml-parse
+  as: read-version
+  config:
+    path: ./Chart.yaml
+    outputs:
+      - name: currentVersion
+        fromExpression: version
+      - name: parsed
+        fromExpression: semverParse(version)
```
```suggestion
    - name: currentVersion
      fromExpression: version
    - name: parsed
      fromExpression: semverParse(version)
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -740,6 +740,66 @@ config:
...
+
+# Update chart with bumped minor version (e.g., 1.2.3 → 1.3.0)
+- uses: yaml-update
+  config:
+    path: ./Chart.yaml
+    updates:
+      - key: version
+        value: ${{ task.outputs['read-version'].parsed.IncMinor().String() }}
```
```suggestion
    - key: version
      value: ${{ task.outputs['read-version'].parsed.IncMinor().String() }}
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -740,6 +740,66 @@ config:
...
+
+# Or access individual components
+- uses: set-var
+  config:
+    vars:
+      - name: majorVersion
+        value: ${{ task.outputs['read-version'].parsed.Major() }}
+      - name: minorVersion
+        value: ${{ task.outputs['read-version'].parsed.Minor() }}
```
```suggestion
    - name: majorVersion
      value: ${{ task.outputs['read-version'].parsed.Major() }}
    - name: minorVersion
      value: ${{ task.outputs['read-version'].parsed.Minor() }}
```

pkg/expressions/function/functions_test.go
```diff
@@ -1869,3 +1870,61 @@ func Test_semverDiff(t *testing.T) {
...
+			},
+		},
+		{
+			name: "invalid argument type",
+			args: []any{123},
+			assertions: func(t *testing.T, result any, err error) {
+				assert.ErrorContains(t, err, "argument must be string")
+				assert.Nil(t, result)
+			},
+		},
```
krancour:
```suggestion
		{
			name: "no arguments",
			args: []any{},
			assertions: func(t *testing.T, result any, err error) {
				assert.ErrorContains(t, err, "expected 1 argument")
				assert.Nil(t, result)
			},
		},
		{
			name: "too many arguments",
			args: []any{"1.0.0", "2.0.0"},
			assertions: func(t *testing.T, result any, err error) {
				assert.ErrorContains(t, err, "expected 1 argument")
				assert.Nil(t, result)
			},
		},
		{
			name: "invalid argument type",
			args: []any{123},
			assertions: func(t *testing.T, result any, err error) {
				assert.ErrorContains(t, err, "argument must be string")
				assert.Nil(t, result)
			},
		},
		{
			name: "invalid semver",
			args: []any{"invalid"},
			assertions: func(t *testing.T, result any, err error) {
				assert.ErrorContains(t, err, "invalid semantic version")
				assert.Nil(t, result)
			},
		},
		{
			name: "success",
			args: []any{"1.2.3"},
			assertions: func(t *testing.T, result any, err error) {
				assert.NoError(t, err)
				parsed, ok := result.(*semver.Version)
				require.True(t, ok)
				assert.NotNil(t, parsed)
			},
		},

```

pkg/expressions/function/functions_test.go
```diff
@@ -1869,3 +1870,61 @@ func Test_semverDiff(t *testing.T) {
 		})
 	}
 }
+
+func Test_semverParse(t *testing.T) {
+	tests := []struct {
```
krancour:
```suggestion
	testCases := []struct {
```

pkg/expressions/function/functions_test.go
```diff
@@ -1869,3 +1870,61 @@ func Test_semverDiff(t *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := semverParse(tt.args...)
+			tt.assertions(t, result, err)
+		})
+	}
```
krancour:
```suggestion
	for _, testCase := range testCases {
		t.Run(testCase.name, func(t *testing.T) {
			result, err := semverParse(testCase.args...)
			testCase.assertions(t, result, err)
		})
	}
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -740,6 +740,66 @@ config:
...
+    - name: parsed
+      fromExpression: semverParse(version)
+
+# Update chart with bumped minor version (e.g., 1.2.3 → 1.3.0)
+- uses: yaml-update
+  config:
+    path: ./Chart.yaml
+    updates:
+    - key: version
+      value: ${{ task.outputs['read-version'].parsed.IncMinor().String() }}
```
krancour: I didn't try it, but I'm about 99% sure this example won't work 100% of the time.

The `semver.Version` struct that you're stashing in the output of the `yaml-parse` step will, at some point, be marshaled to JSON and persisted (written back to Promotion `status`). If your subsequent `yaml-update` step runs right away, the output from the previous step might still contain the _original_ `semver.Version` struct you stashed and your example might work...

If, for any reason, the `yaml-parse` step does _not_ run right away, perhaps because some step in between these two had to wait for a period of time or retry something a number of times, then when that `yaml-parse` step _does_ run, any shared state it has access to will have been re-hydrated from JSON and there would have been a loss of type information. What you'd be working with is actually a `map[string]any` and _not_ a `semver.Version` struct. The call to the `IncMinor()` method would fail.

The example can be re-written to work, however. The following should work every. cc @hiddeco to double-check my understanding above as well as the proposed change:

```suggestion
# Read current chart version from Char.yaml

- uses: yaml-parse
  as: read-version
  config:
    path: ./src/Chart.yaml
    outputs:
    - name: currentVersion
      fromExpression: version

# Update Chart.yaml with bumped minor version

- uses: yaml-update
  config:
    path: ./src/Chart.yaml
    updates:
    - key: version
      value: ${{ semverParse(task.outputs['read-version'].currentVersion).IncMinor().String() }}
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -740,6 +740,66 @@ config:
...
+
+# Or access individual components
+- uses: set-var
+  config:
+    vars:
+    - name: majorVersion
+      value: ${{ task.outputs['read-version'].parsed.Major() }}
+    - name: minorVersion
+      value: ${{ task.outputs['read-version'].parsed.Minor() }}
```
krancour: `set-var` isn't a step type that exists. I would just drop this example entirely because the one above is (after corrections) as great an example of realistic use as I can imagine. I don't think we need this extra bit.

```suggestion
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
hiddeco: Kent is right, and the version will have to be parsed each time, as done in his revised example.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -740,6 +740,56 @@ config:
...
+:::info
+The function uses the [Semantic Versioning](https://semver.org/) specification
+to parse versions. It supports versions with or without the `v` prefix, as well
+as prerelease and build metadata components.
+:::
+
+Example:
+
+```yaml
+# Read current chart version from Char.yaml
```
```suggestion
# Read current chart version from Chart.yaml
```

pkg/expressions/function/functions_test.go
```diff
@@ -1869,3 +1870,62 @@ func Test_semverDiff(t *testing.T) {
...
+			name: "success",
+			args: []any{"1.2.3"},
+			assertions: func(t *testing.T, result any, err error) {
+				assert.NoError(t, err)
+				parsed, ok := result.(*semver.Version)
+				require.True(t, ok)
+				assert.NotNil(t, parsed)
+			},
+		},
+
```
krancour:
```suggestion
```

PR #5319, opened 2025-10-31
fix(promotion): kustomize-set-image images field optional

### LINKED ISSUE #5316: Unable to omit images field in promotion step kustomize-set-image
# Checklist

# Description

When using kustomize-set-image promotion task. Omitting images fields causes an error. I need this feature since we are migrating apps over which use 0 -> N images. This allows me to continue using a single promotion template with variable amount of images in the pipeline warehouse.

PR which added this feature initially: https://github.com/akuity/kargo/pull/3115

# Screenshots

Field shows as optional yet still causes validation error 
Image

# Steps to Reproduce

Execution a promotion including kustomize-set-image with images field not set

ex
```
  - uses: kustomize-set-image
    as: update-images
    config:
      path: <path-to-my-code>
```

# Version

```
Client Version: v1.7.5
Server Version: v1.7.5
```

# Logs

```
an unrecoverable error occurred: error running step "task-1::update-images": invalid kustomize-set-image config: (root): images is required
```

GooglyBlox:
## Description

This PR fixes the validation error that prevented users from omitting the `images` field in the `kustomize-set-image` promotion step.

## Changes

- Updated `kustomize-set-image-config.json` schema to make `images` field optional (removed from required array)
- Added test case `valid_config_with_only_path_(images_optional)` to validate the fix

## Motivation

When the `images` field is omitted, all images from the Freight collection should be automatically set in the Kustomization file (unless there's an ambiguous match). This feature was intended in the original implementation but was blocked by schema validation.

This allows users to maintain a single promotion template that works with pipelines containing 0 to N images, which is particularly useful during app migrations.

## Testing

All unit tests pass, including:
- New validation test for config with only `path` field
- Existing test for automatic image detection (`automatically_sets_image`)

Fixes #5316

### DIFF
--- a/pkg/x/promotion/runner/builtin/zz_config_types.go
+++ b/pkg/x/promotion/runner/builtin/zz_config_types.go
@@ -455,7 +455,7 @@ type KustomizeSetImageConfig struct {
 	// left unspecified, all images from the Freight collection will be set in the Kustomization
 	// file. Unless there is an ambiguous image name (for example, due to two Warehouses
 	// subscribing to the same repository), which requires manual configuration.
-	Images []Image `json:"images"`
+	Images []Image `json:"images,omitempty"`
 	// Path to the directory containing the Kustomization file.
 	Path string `json:"path"`
 }


GooglyBlox:
Sorry about that, this should fix the check-codegen job failing! 

Would greatly appreciate a second review @hiddeco.

Issue #5345, opened 2025-11-05
Promoting an unverified freight when using the truck button doesn't prompt for approval, it just fails.


dhpup:
# Checklist

* [X] I've searched the issue queue to verify this is not a duplicate bug report.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `kargo version`.
* [X] I've pasted logs, if applicable.

# Description

Promoting an unverified freight when using the truck button doesn't prompt for approval, it just fails. 

# Screenshots

https://github.com/user-attachments/assets/c548bb88-786a-4c5c-824f-884549e41a16

# Steps to Reproduce

Create a freight, drag it to a later stage unvalidated, see approval box, try again with truck promotion and it doesnt prompt for approval. 

# Version

```
Kargo 1.8.0
```

krancour:
That's how it's always worked. So it's janky, but not new. The approval dialog _is_ new. The real issue, I guess, is that drag 'n' drop got that new feature and the other workflow did not. This has been discussed in some other threads as well, notably #5287, but there's a lot going on in this issue...

So I'll keep this narrowly-focused issue open and reframe it as a UI feature request.

PR #5346, opened 2025-11-05
feat(ui): promote unverified freight with manual approval confirmation

Marvin9:
fixes https://github.com/akuity/kargo/issues/5345

### DIFF
--- a/ui/src/features/project/pipelines/nodes/promotion/use-get-promotion-dropdown-items.tsx
+++ b/ui/src/features/project/pipelines/nodes/promotion/use-get-promotion-dropdown-items.tsx
@@ -36,14 +36,17 @@ export const useGetPromotionDropdownItems = (stage: Stage) => {
 
   const showManualApproveModal = useManualApprovalModal();
 
-  const ensureEligibilityBeforeAction = async (
-    freight: string | undefined,
-    onEligible: (eligibleFreight: string) => void,
-    options?: { runAfterApproval?: (eligibleFreight: string) => void }
-  ) => {
+  const ensureEligibilityBeforeAction = async ({
+    freight,
+    onSuccess
+  }: {
+    freight: string | undefined;
+    onSuccess: (eligibleFreight: string) => void;
+  }) => {
     if (!freight) {
       return;
     }
+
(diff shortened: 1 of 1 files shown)


ui/src/features/project/pipelines/nodes/stage-node.tsx
rpelczar, line 1: This component is getting really hard to read and understand.
Can you extract all promotion-related code and move it to the `promotion/node/` folder next to the current folder (`/nodes`)?

Also, I have some feedback on the changes you made.

You created a long chain of functions, which is also hard to read.
I like that you separated logic, but e.g., there is no point in adding this `checkFreightPromotionEligibility` as a separate function.

I'm pretty sure it could be much simpler. The only difference between those two actions is the function at the end.

ui/src/features/project/pipelines/nodes/stage-node.tsx
```diff
@@ -76,15 +80,57 @@ export const StageNode = (props: { stage: Stage }) => {
...
+    if (!isEligible) {
+      showManualApproveModal({
+        freight,
+        projectName,
+        stage: stageName,
+        onApprove: options?.runAfterApproval
+          ? () => {
+              options.runAfterApproval?.(freight);
+            }
+          : undefined
```
rpelczar: This condition is not really helpful. Nothing wrong if you redirect the user here, and simplify the code.

ui/src/features/project/pipelines/promotion/use-manual-approval-modal.tsx
```diff
@@ -52,6 +53,12 @@ export const useManualApprovalModal = () => {
           {
             onSuccess: () => {
               onClose?.();
+
+              if (onApprove) {
+                onApprove();
+                return;
+              }
```
rpelczar: You shouldn't make this condition. Adding `onApprove` method doesn't suggest that there is an action ignored when a property is added. Please make this prop required, and add redirection where the hook is used.

rpelczar (changes requested):
The functionality works as expected, but I'd like to improve code quality.

Marvin9:
Tested and it works as expected. Thanks!

PR #5350, opened 2025-11-05
fix(controller): make git repo retain url suitable for use by external callers of URL() method

### LINKED ISSUE #5344: RegEx URL for credentials don't work for Git Push step unless it's all lowercase
# Checklist

# Description

When using a RegEx URL with camelcase within the credentials, I'm able to clone a repo in promotion, but not able to git push. There is some issue with the normalization logic. 

# Version

```
Kargo 1.8.0
```

# Logs

```
step "task-1::step-6" met error threshold of 1: error running step "task-1::step-6": error pushing commits to remote: error checking for existence of branch
```

krancour:
Fixes #5344

Internally, implementations of `git.Repo` hang on to URLs that often contain `@username`, but callers of the `URL()` method rightfully expect it to return the original URL that was provided to the `git.Clone()` or `git.CloneBare()` function.

### DIFF
--- a/pkg/controller/git/base_repo.go
+++ b/pkg/controller/git/base_repo.go
@@ -17,6 +17,10 @@ import (
 const (
 	defaultUsername = "Kargo"
 	defaultEmail    = "no-reply@kargo.io"
+
+	repoDirConfigKey         = "kargo.repoDir"
+	repoHomeDirConfigKey     = "kargo.repoHomeDir"
+	repoOriginalURLConfigKey = "kargo.repoOriginalURL"
 )
 
 // baseRepo implements the common underpinnings of a Git repository with a
@@ -26,7 +30,11 @@ type baseRepo struct {
 	creds   *RepoCredentials
 	dir     string
 	homeDir string
-	url     string
+	// Store the URL two ways:
+	// 1. Exactly as it was originally provided
+	// 2. Modified in ways required for internal use by this type's methods
+	originalURL string
+	accessURL   string
 }
 
 // ClientOptions represents options for a repository-specific Git client.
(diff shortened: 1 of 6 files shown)


pkg/controller/git/base_repo.go
```diff
@@ -26,7 +26,11 @@ type baseRepo struct {
 	creds   *RepoCredentials
 	dir     string
 	homeDir string
-	url     string
+	// Store the URL two ways. One is for internal use and may contain username@,
+	// while the other lacks that and is suitable for use in error messages and
+	// consumption by external callers of the URL() method.
+	internalURL string
+	externalURL string
```
hiddeco: I find the `internalURL` and `externalURL` naming quite confusing. Would something like `url` (or `operationalURL`) and `canonicalURL` maybe be better?

krancour: Now that you've pointed it out, I agree the names aren't intuitive. You'd need to read the comments to understand. "Canonical" doesn't work because it's too similar to "normalized," and there's no real giant this _is_ normalized. I'll noodle on this a bit. I agree there's a better naming scheme somewhere here that I've overlooked.

jessesuen:
Question: I was helping a user where the `argocd-update` step was passing this as repoURL:
* https://github.com/my-org/my-repo.git

but Argo's Application was configured without the `.git`:
* https://github.com/my-org/my-repo

And so the `argocd-update` step was not updating the app properly. Would this fix have helped?

krancour:
@jessesuen I don't believe so. That sounds like a bug, but likely unrelated to this.

pkg/controller/git/base_repo.go
krancour: Finally addressed this feedback. I've switched to `originalURL` and `accessURL`. I've also changed to storing the original URL directly in git config instead of re-assembling it from `remote.origin.url`.

hiddeco (approved):
Still not sure these are the ultimate names, but I like it better, and naming is hard.

PR #5378, opened 2025-11-12
refactor: generic internal component registries

krancour:
I've got another PR waiting in the wings that requires two new "registries" wherein we register different implementations of an interface with the registry and an appropriate implementation (or factory function for an appropriate implementation) can be retrieved, in some cases, by name, and in other cases, by execution of an implementation-specific predicate function until one implementation reports it is adequate for the purposes at hand.

We've already done this in four other places in this repo (and at least a cpuple in EE as well) and I'm sick of writing interface-specific registries every time we do something like this, so this PR introduces two generic registries (one name-based and the other predicate-based) and transitions four parts of the code base off of their older interface-specific registries and onto these newer ones.

I've marked this high only because I've got a high priority feature already built on top of this.

### DIFF
--- a/pkg/controller/git/commit/selector.go
+++ b/pkg/controller/git/commit/selector.go
@@ -2,7 +2,7 @@ package commit
 
 import (
 	"context"
-	"errors"
+	"fmt"
 
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 	"github.com/akuity/kargo/pkg/controller/git"
@@ -27,14 +27,9 @@ func NewSelector(
 ) (Selector, error) {
 	// Pick an appropriate Selector implementation based on the subscription
 	// provided.
-	reg, found, err := defaultSelectorRegistry.Get(ctx, sub)
+	reg, err := defaultSelectorRegistry.Get(ctx, sub)
 	if err != nil {
-		return nil, err
-	}
-	if !found {
-		// This shouldn't happen because the API doesn't allow opting for any
-		// selector for which for which we've not registered an implementation.
-		return nil, errors.New("no selector found for subscription")
+		return nil, fmt.Errorf("error getting selector factory")
 	}
 	factory := reg.Value
 	return factory(sub, creds)
--- a/pkg/controller/promotions/promotions.go
+++ b/pkg/controller/promotions/promotions.go
@@ -462,7 +462,7 @@ func (r *reconciler) Reconcile(
 	// it.
 	if newStatus.Phase == kargoapi.PromotionPhaseRunning {
 		return ctrl.Result{
-			RequeueAfter: calculateRequeueInterval(promo, suggestedRequeueInterval),
+			RequeueAfter: calculateRequeueInterval(ctx, promo, suggestedRequeueInterval),
 		}, nil
 	}
 	return ctrl.Result{}, nil
@@ -692,6 +692,7 @@ func (r *reconciler) terminatePromotion(
 var defaultRequeueInterval = 5 * time.Minute
 
 func calculateRequeueInterval(
+	ctx context.Context,
 	p *kargoapi.Promotion,
 	suggestedRequeueInterval *time.Duration,
 ) time.Duration {
@@ -706,8 +707,9 @@ func calculateRequeueInterval(
 	}
 
 	step := p.Spec.Steps[p.Status.CurrentStep]
-	reg, found := promotion.DefaultStepRunnerRegistry.Get(step.Uses)
-	if !found {
+	reg, err := promotion.DefaultStepRunnerRegistry.Get(step.Uses)
+	if err != nil {
+		logging.LoggerFromContext(ctx).Error(err, err.Error())
 		return requeueInterval
 	}
 	timeout := step.Retry.GetTimeout(reg.Metadata.DefaultTimeout)
--- a/pkg/image/selector.go
+++ b/pkg/image/selector.go
@@ -2,7 +2,7 @@ package image
 
 import (
 	"context"
-	"errors"
+	"fmt"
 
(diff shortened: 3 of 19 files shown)


pkg/webhook/external/receiver.go
```diff
@@ -91,11 +92,15 @@ func NewReceiver(
...
 	// Pick an appropriate WebhookReceiver implementation based on the
 	// configuration provided.
-	receiverFactory, err := registry.getReceiverFactory(cfg)
+	reg, found, err := defaultWebhookReceiverRegistry.Get(ctx, cfg)
+	if !found { // This shouldn't happen
+		return nil, errors.New("found no suitable webhook receiver")
+	}
 	if err != nil {
 		return nil, err
 	}
```
hiddeco: Believe the `err != nil` and `!found` check should be swapped?

pkg/image/selector.go
```diff
@@ -20,14 +21,19 @@ type Selector interface {
...
 	// provided.
-	selectorFactory, err := selectorReg.getSelectorFactory(sub)
+	reg, found, err := defaultSelectorRegistry.Get(ctx, sub)
 	if err != nil {
 		return nil, err
 	}
-	return selectorFactory(sub, creds)
+	if !found { // This shouldn't happen
+		return nil, errors.New("no selector found for subscription")
+	}
```
hiddeco: Believe the `err != nil` and `!found` check should be swapped?

pkg/controller/git/commit/selector.go
```diff
@@ -20,14 +21,19 @@ type Selector interface {
...
 	// provided.
-	selectorFactory, err := selectorReg.getSelectorFactory(sub)
+	reg, found, err := defaultSelectorRegistry.Get(ctx, sub)
 	if err != nil {
 		return nil, err
 	}
-	return selectorFactory(sub, creds)
+	if !found { // This shouldn't happen
+		return nil, errors.New("no selector found for subscription")
+	}
```
hiddeco: Believe the `err != nil` and `!found` check should be swapped?

pkg/controller/promotions/promotions.go
```diff
@@ -706,8 +706,11 @@ func calculateRequeueInterval(
 	}
 
 	step := p.Spec.Steps[p.Status.CurrentStep]
-	registration := promotion.GetStepRunnerRegistration(step.Uses)
-	timeout := step.Retry.GetTimeout(registration.Metadata.DefaultTimeout)
+	reg, found := promotion.DefaultStepRunnerRegistry.Get(step.Uses)
+	if !found {
+		return requeueInterval
```
hiddeco, line 713: Wouldn't it be more correct to make this return a requeue interval of `0`?

pkg/promotion/registry.go
```diff
@@ -1,60 +1,91 @@
...
-// specified promotion step kind from the package's internal registry. If no
-// such registration exists, nil is returned instead.
-func GetStepRunnerRegistration(
-	stepKind string,
-) *StepRunnerRegistration {
-	return stepRunnerRegistry.GetStepRunnerRegistration(stepKind)
+func MustNewStepRunnerRegistry(
+	registrations ...StepRunnerRegistration,
+) StepRunnerRegistry {
+	return component.MustNewNameBasedRegistry(
```
hiddeco: To maintain 100% backwards compatible behavior, I believe something like this would be required:

```go
	// Validate and normalize DefaultErrorThreshold to system-wide default of 1
	// if not set. This prevents infinite retries and ensures consistent behavior.
	for i := range registrations {
		if registrations[i].Metadata.DefaultErrorThreshold <= 0 {
			registrations[i].Metadata.DefaultErrorThreshold = 1
		}
	}
```

pkg/webhook/external/receiver.go
krancour: 🤦‍♂️

pkg/image/selector.go
krancour: Although this was improperly ordered elsewhere, these actually seem to be in the correct order. But I am adding a better comment here.

pkg/controller/promotions/promotions.go
krancour: 🤔 Looking at the surrounding context, I believe this is correct. Prior to this change, any difficulty calculating an interval was already handled by returning the current value of `requeueInterval`, which will be either the optional, suggested requeue interval (returned by the step) or the reconciler's default.

pkg/image/selector.go
krancour: Unless you _actually_ meant `!found` should be handled first.

pkg/controller/git/commit/selector.go
krancour: https://github.com/akuity/kargo/pull/5378#discussion_r2518850292

pkg/promotion/registry.go
krancour: You're right. But it's not enough...

The registration function also needs to be decorated with the same defaulting behavior so that it also applies to anything registered _after_ this registry is initialized.

Fix incoming...

pkg/webhook/external/receiver.go
krancour: Fixed in b88aca1dffd5f9fce322287f77718a04b9dd7cbf

pkg/promotion/registry.go
krancour: Fixed in b88aca1dffd5f9fce322287f77718a04b9dd7cbf

pkg/image/selector.go
krancour: No longer relevant after latest commit anyway.

hiddeco (approved):

PR #5380, opened 2025-11-12
refactor: credential providers to use generic component registry

krancour:
This is a relatively minor refactoring to make credential providers user the generic predicate-based registry introduced in #5378.

High priority only because there are other things waiting in the wings that require this change.

### DIFF
--- a/pkg/credentials/ssh/ssh.go
+++ b/pkg/credentials/ssh/ssh.go
@@ -1,4 +1,4 @@
-package basic
+package ssh
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


pkg/credentials/ssh/ssh.go
```diff
@@ -0,0 +1,41 @@
+package basic
```
hiddeco: Should probably be `ssh`?

krancour: Nice catch. Fixed in 6e1eb25ea0e4c18843e372ae876dc31402180b80

PR #5382, opened 2025-11-13
refactor: add subscriber interface

krancour:
This is just a bit of internal refactoring to pay down some tech debt. Functionally, there is no change.

@hiddeco and @thomastaylor312 both have the context.

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/90-events/10-event-reference.md
+++ b/docs/docs/50-user-guide/60-reference-docs/90-events/10-event-reference.md
@@ -41,15 +41,16 @@ These fields are included in all Kargo events:
 
 Freight payloads describe the collection of artifacts under evaluation or promotion.
 
-| Field Name   | Type               | Description                                                                       | Optional         |
-| ------------ | ------------------ | --------------------------------------------------------------------------------- | ---------------- |
-| `name`       | String             | Name of the freight object.                                                       | No               |
-| `stageName`  | String             | Stage associated with the freight when the event fired.                           | No               |
-| `createTime` | String (RFC3339)   | Creation timestamp of the freight object.                                         | No               |
-| `alias`      | String             | Human-friendly alias assigned to the freight.                                     | Yes (is pointer) |
-| `commits`    | Array\<GitCommit\> | Git commits that compose the freight (see [GitCommit fields](#gitcommit-fields)). | Yes              |
-| `images`     | Array\<Image\>     | Container images included in the freight (see [Image fields](#image-fields)).     | Yes              |
-| `charts`     | Array\<Chart\>     | Helm charts included in the freight (see [Chart fields](#chart-fields)).          | Yes              |
+| Field Name   | Type                       | Description                                                                                                         | Optional         |
+| ------------ | -------------------------- | ------------------------------------------------------------------------------------------------------------------- | ---------------- |
+| `name`       | String                     | Name of the freight object.                                                                                         | No               |
+| `stageName`  | String                     | Stage associated with the freight when the event fired.                                                             | No               |
+| `createTime` | String (RFC3339)           | Creation timestamp of the freight object.                                                                           | No               |
+| `alias`      | String                     | Human-friendly alias assigned to the freight.                                                                       | Yes (is pointer) |
+| `commits`    | Array\<GitCommit\>         | Git commits that compose the freight (see [GitCommit fields](#gitcommit-fields)).                                   | Yes              |
+| `images`     | Array\<Image\>             | Container images included in the freight (see [Image fields](#image-fields)).                                       | Yes              |
+| `charts`     | Array\<Chart\>             | Helm charts included in the freight (see [Chart fields](#chart-fields)).                                            | Yes              |
+| `artifacts`  | Array\<ArtifactReference\> | Additional arbitrary artifacts included in the freight (see [ArtifactReference fields](#artifactreference-fields)). | Yes              |
 
 #### `GitCommit` Fields
 
@@ -79,6 +80,15 @@ Freight payloads describe the collection of artifacts under evaluation or promot
 | `repoURL`  | String | Helm chart repository URL.                                            | Yes      |
 | `name`     | String | Chart name within the repository (empty for OCI-style references).    | Yes      |
 | `version`  | String | Specific chart version selected for inclusion in the freight payload. | Yes      |
+
+#### `ArtifactReference` Fields
+
+| Field Name         | Type                 | Description                                                                                                                                                                                                                                                                                                       | Optional |
+| ------------------ | -------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| `artifactType`     | String               | Unique type of the artifact.                                                                                                                                                                                                                                                                                      | No       |
+| `subscriptionName` | String               | Name of the subscription that discovered the artifact.                                                                                                                                                                                                                                                            | No       |
+| `version`          | String               | Version identifies a specific revision of this artifact.                                                                                                                                                                                                                                                          | No       |
+| `metadata`         | Map\<String,Object\> | Additional metadata associated with the artifact. It is a mostly opaque collection of attributes. "Mostly" because Kargo may understand how to interpret some documented, well-known, top-level keys. Those aside, this metadata is only understood by a corresponding Subscriber implementation that created it. | Yes      |
 
 ### Freight Verification Fields
 


rpelczar (approved):
UI LGTM

api/v1alpha1/freight_types.go
```diff
@@ -40,6 +40,9 @@ type Freight struct {
 	Images []Image `json:"images,omitempty" protobuf:"bytes,4,rep,name=images"`
 	// Charts describes specific versions of specific Helm charts.
 	Charts []Chart `json:"charts,omitempty" protobuf:"bytes,5,rep,name=charts"`
+	// OtherArtifacts describes specific versions of artifacts other
+	// than Git repository commits, container images, and Helm charts.
+	OtherArtifacts []GenericArtifactReference `json:"otherArtifacts,omitempty" protobuf:"bytes,10,rep,name=otherArtifacts"`
```
thomastaylor312: Nit: Call these `GenericArtifacts` rather than `OtherArtifacts`. (Ironically) In this case, `Generic` is more specific than `Other`, which makes it sound like a junk drawer to me

api/v1alpha1/stage_types.go
```diff
@@ -487,6 +487,9 @@ type FreightReference struct {
 	Images []Image `json:"images,omitempty" protobuf:"bytes,3,rep,name=images"`
 	// Charts describes specific versions of specific Helm charts.
 	Charts []Chart `json:"charts,omitempty" protobuf:"bytes,4,rep,name=charts"`
+	// OtherArtifacts describes specific versions of artifacts other
+	// than Git repository commits, container images, and Helm charts.
+	OtherArtifacts []GenericArtifactReference `json:"otherArtifacts,omitempty" protobuf:"bytes,9,rep,name=otherArtifacts"`
```
thomastaylor312: Nit: Same here around naming

api/v1alpha1/warehouse_types.go
```diff
@@ -145,6 +146,9 @@ type RepoSubscription struct {
 	Image *ImageSubscription `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
 	// Chart describes a subscription to a Helm chart repository.
 	Chart *ChartSubscription `json:"chart,omitempty" protobuf:"bytes,3,opt,name=chart"`
+	// Other describes a subscription to something that is not a Git, container
+	// image, or Helm chart repository.
+	Other *GenericSubscription `json:"other,omitempty" protobuf:"bytes,4,opt,name=other"`
```
thomastaylor312: Nit: Same here with naming

api/v1alpha1/warehouse_types.go
```diff
@@ -510,6 +514,33 @@ type ChartSubscription struct {
 	DiscoveryLimit int32 `json:"discoveryLimit,omitempty" protobuf:"varint,4,opt,name=discoveryLimit"`
 }
 
+// GenericSubscription represents a subscription to something that is not a Git,
+// container image, or Helm chart repository.
+type GenericSubscription struct {
+	// ArtifactKind specifies the kind of artifact this subscription is for.
```
thomastaylor312: I would love this to have a bit more documentation here as to what this string value should be. Is it entirely user defined? Is it a MIME type? I haven't gotten further so maybe it is in docs, but even so it might be worth clarifying here some more

api/v1alpha1/warehouse_types.go
```diff
@@ -566,6 +597,12 @@ type DiscoveredArtifacts struct {
 	//
 	// +optional
 	Charts []ChartDiscoveryResult `json:"charts,omitempty" protobuf:"bytes,3,rep,name=charts"`
+	// OtherResults holds the artifact references discovered by the Warehouse
+	// for all subscriptions that are not to a Git, container image, or Helm
+	// chart repository.
+	//
+	// +optional
+	OtherResults []GenericDiscoveryResult `json:"otherResults,omitempty" protobuf:"bytes,5,rep,name=otherResults"`
```
thomastaylor312: Nit: rename to just `Other`/`Generic` instead of using the `Results` name to align more clearly with the other field names

pkg/api/freight.go
```diff
@@ -19,59 +19,78 @@ import (
 // on its contents and returns it.
 func GenerateFreightID(f *kargoapi.Freight) string {
 	size := len(f.Commits) + len(f.Images) + len(f.Charts)
```
thomastaylor312: This should probably have `len(f.OtherCommits)` as well

pkg/api/freight.go
```diff
@@ -19,59 +19,78 @@ import (
...
+	// known, but with a new digest -- as in the case of "mutable" tags like
+	// "latest". It is equally possible to have found an image with a digest that
+	// is already known, but has been re-tagged. To cover both cases, we
+	// incorporate BOTH tag and digest into the canonical representation of an
+	// image used when calculating Freight ID.
+	return fmt.Sprintf("%s:%s@%s", img.RepoURL, img.Tag, img.Digest)
+}
+
+// genericArtifactHashPart returns a string that uniquely identifies a specific
+// revision\ of an artifact other than a Git commit, container image, or Helm
```
thomastaylor312: Assuming the `\` is unintentional

pkg/expressions/function/functions.go
```diff
@@ -769,6 +808,130 @@ func getChartFromDiscoveredArtifacts(artifacts *kargoapi.DiscoveredArtifacts) ex
...
+		for _, result := range artifacts.OtherResults {
+			if result.SubscriptionName != subName {
+				continue
+			}
+			if len(result.ArtifactReferences) > 0 {
+				artifact = &result.ArtifactReferences[0]
+			}
+		}
+
+		// Unpack details into a map[string]any so they're actually accessible from
```
thomastaylor312: Nit: This requirement should probably be documented in the `Details` field for clarity

api/v1alpha1/freight_types.go
hiddeco: Yes, +1 on this too. Especially because "other" is weird when there are not any "other" present in the piece of Freight (i.e., there are only generic ones).

This is blocking IMHO.

api/v1alpha1/warehouse_types.go
```diff
@@ -163,29 +164,41 @@ func (w *WarehouseSpec) UnmarshalJSON(data []byte) error {
...
+			// Get the single key
+			var key string
+			for k := range rawMap {
+				key = k
+				break // This unnecessary, but it makes it clear what we're doing
+			}
+
+			// Check for known keys (git, image, chart)
+			knownKeys := map[string]bool{"git": true, "image": true, "chart": true}
+			if knownKeys[key] {
```
hiddeco:
```suggestion
			knownKeys := []string{"git", "image", "chart"}
			if slices.Contains(knownKeys, key) {
```

api/v1alpha1/warehouse_types.go
```diff
@@ -163,29 +164,41 @@ func (w *WarehouseSpec) UnmarshalJSON(data []byte) error {
...
+				// Known subscription type - unmarshal normally
+				if err := json.Unmarshal(
+					aux.Subscriptions[i].Raw,
+					&w.InternalSubscriptions[i],
+				); err != nil {
 					return err
 				}
```
hiddeco, line 175: Returning early here would save you from the `} else {` nesting.

api/v1alpha1/warehouse_types.go
```diff
@@ -413,9 +586,9 @@ type ImageSubscription struct {
 	// image tags that are considered in determining the newest version of an
 	// image. This field is optional.
 	//
-	// Deprecated: Use AllowTagsRegexes instead. Beginning in v1.11.0, artifact
+	// Deprecated: Use AllowTagsRegexes instead. Beginning in apiextensionsv1.11.0, artifact
```
thomastaylor312: This and few lines below look like find/replace mistakes

thomastaylor312: Actually in several places here in the file

api/v1alpha1/warehouse_types.go
```diff
@@ -679,6 +882,61 @@ type ChartDiscoveryResult struct {
...
+	ArtifactReferences []ArtifactReference `json:"artifactReferences" protobuf:"bytes,2,rep,name=artifactReferences"`
+}
+
+// ArtifactReference is a reference to a specific version of an artifact.
+type ArtifactReference struct {
+	// Kind specifies the kind of artifact this is. Often it will be a media
+	// type (MIME type) string.
+	//
+	// +kubebuilder:validation:MinLength=1
+	Kind string `json:"kind,omitempty" protobuf:"bytes,1,opt,name=kind"`
```
thomastaylor312: I still don’t love the use of `kind` as it feels overloaded. We have the subscription kind, the gvk kind, and now this kind. If it isn’t going to always be a media type, I’d love to find another name for this. 

Not blocking IMO, just calling out before we commit to this API

pkg/api/freight.go
krancour: Ack. Will fix.

pkg/expressions/function/functions.go
krancour: Also realized that now that details has changed to "metadata," there might be another small tweak needed to these functions.

Will fix both.

api/v1alpha1/warehouse_types.go
krancour: This was copilot's doing, but it's on me that I didn't object to it. Happy to change it.

Personally, I might have done `knownKeys := map[string]struct{}{ ... }`, but since the list is very short, I think your way wins on readability.

Will change as you suggest.

api/v1alpha1/warehouse_types.go
krancour: Yikes. That is precisely what happened. Nice catch.

api/v1alpha1/warehouse_types.go
krancour: Type? Class? These seem overloaded for different reasons. "Ilk" would be novel, but too weird.

I felt "kind" was the one to beat _because_ it felt k8s-ish. If you can suggest something better, I'd use it.

api/v1alpha1/warehouse_types.go
krancour: Actually no... This is inside a loop. Early return here would result in data being lost during the unmarshaling.

pkg/api/freight.go
krancour: Fixed in f0f2cbb88d07ab8cdcb205c3b4f5f4f24c7c5d30

api/v1alpha1/warehouse_types.go
krancour: Fixed in f0f2cbb88d07ab8cdcb205c3b4f5f4f24c7c5d30

api/v1alpha1/warehouse_types.go
krancour: Fixed in f0f2cbb88d07ab8cdcb205c3b4f5f4f24c7c5d30

api/v1alpha1/warehouse_types.go
krancour: We hashed this out offline.

Fixed in ddbb29f5c260201a52aa8459cb571d749fac78ec

pkg/expressions/function/functions.go
krancour: Fixed in b79fcc5cd58ff1eb437a0e1ad8ca440619218318

api/v1alpha1/warehouse_types.go
thomastaylor312: I didn't actually see a fix here, so I unresolved the comment. Totally fine if you keep it as is, but just wanted to call out that I didn't see a change here

krancour: Strange. You're right. "Re"-fixing.

krancour: Fixed for real in 69cfde56f73dc8ff8e3eae6fee66aff976c138b4

pkg/subscription/subscriber.go
```diff
@@ -0,0 +1,39 @@
...
+	ApplySubscriptionDefaults(
+		ctx context.Context,
+		sub *kargoapi.RepoSubscription,
+	) error
+	// ValidateSubscription validates the provided kargoapi.RepoSubscription.
+	ValidateSubscription(
+		ctx context.Context,
+		f *field.Path,
+		sub kargoapi.RepoSubscription,
+	) field.ErrorList
```
krancour, line 26: I will call out myself here that I don't entirely love this package having a dependency on apimachinery, however, since this method is implemented specifically so that a validating webhook can delegate some of its validation responsibilities to it and get back properly formatted errors, I think it makes a certain amount of sense to return field.Errorlist since that's seamlessly usable by the webhook.

krancour:
@Marvin9 back end changes broke the UI again. I'm sorry!

api/v1alpha1/warehouse_types.go
```diff
@@ -117,6 +134,161 @@ type WarehouseSpec struct {
...
+
+// MarshalJSON marshals the WarehouseSpec to JSON, converting the
+// InternalSubscriptions slice into JSON in the Subscriptions field. For
+// Subscription objects, the Kind field is used as the top-level key.
+func (w *WarehouseSpec) MarshalJSON() ([]byte, error) {
+	type warehouseSpecAlias WarehouseSpec
+
+	// Create a copy to avoid modifying the original
+	spec := *w
+	aux := &struct {
```
thomastaylor312, line 197: Super nit: Since we use this inline `struct` twice, should we define it as a private struct here?

api/v1alpha1/warehouse_types.go
```diff
@@ -117,6 +134,161 @@ type WarehouseSpec struct {
...
+						"Chart, or Subscription set, but has %d",
+					i, typesSet,
+				)
+			}
+
+			// If this is a generic Subscription, wrap it with its Kind as the key
+			if sub.Subscription != nil {
+				kind := sub.Subscription.SubscriptionType
+				if kind == "" {
+					return nil, fmt.Errorf("subscription at index %d has empty Kind field", i)
```
thomastaylor312: Should this error message say "empty SubscriptionType" field instead?

api/v1alpha1/warehouse_types.go
```diff
@@ -679,6 +819,63 @@ type ChartDiscoveryResult struct {
...
+	// +kubebuilder:validation:MinLength=1
+	SubscriptionName string `json:"subscriptionName" protobuf:"bytes,2,opt,name=subscriptionName"`
+	// Version identifies a specific revision of this artifact.
+	//
+	// +kubebuilder:validation:MinLength=1
+	Version string `json:"version" protobuf:"bytes,3,opt,name=version"`
+	// Metadata is a JSON object containing a mostly opaque collection of artifact
+	// attributes. (It must be an object. It may not be a list or a scalar value.)
+	// "Mostly" because Kargo may understand how to interpret some documented,
+	// well-known top-level keys. Those aside, this metadata is only understood by
```
thomastaylor312: Super nit: comma after "well-known"

api/v1alpha1/warehouse_types.go
```diff
@@ -679,6 +819,63 @@ type ChartDiscoveryResult struct {
...
+func (g *ArtifactReference) DeepEquals(
+	other *ArtifactReference,
+) bool {
+	if g == nil && other == nil {
+		return true
+	}
+	if g == nil || other == nil {
+		return false
+	}
+	return g.SubscriptionName == other.SubscriptionName &&
```
thomastaylor312: Should this be checking `ArtifactType` as well?

pkg/api/freight.go
```diff
@@ -18,60 +18,78 @@ import (
...
+	// is already known, but has been re-tagged. To cover both cases, we
+	// incorporate BOTH tag and digest into the canonical representation of an
+	// image used when calculating Freight ID.
+	return fmt.Sprintf("%s:%s@%s", img.RepoURL, img.Tag, img.Digest)
+}
+
+// artifactHashPart returns a string that uniquely identifies a specific
+// revision of an artifact.
+func artifactHashPart(ref kargoapi.ArtifactReference) string {
+	return fmt.Sprintf("%s:%s", ref.SubscriptionName, ref.Version)
```
thomastaylor312: This should probably contain `ArtifactType` as well for safety sake

pkg/controller/freight/finder.go
```diff
@@ -284,3 +284,76 @@ func FindChart(
...
+				if sub.Subscription != nil && sub.Subscription.Name == subName {
+					if desiredOrigin != nil {
+						return nil, fmt.Errorf(
+							"multiple requested Freight could potentially provide a "+
+								"%q artifact; please update promotion steps to "+
+								"disambiguate",
+							subName,
+						)
+					}
+					desiredOrigin = &requestedFreight.Origin
```
thomastaylor312, line 335: Nit: I don't love mutating a parameter passed in to the function as a side effect to calling the function. Probably doesn't really matter with how this function is used, but I'd prefer a local variable instead

pkg/controller/warehouses/warehouses.go
```diff
@@ -508,29 +518,48 @@ func (r *reconciler) syncWarehouse(
...
+		case kargoapi.GitDiscoveryResult:
+			discovered.Git = append(discovered.Git, typedRes)
+		case kargoapi.ImageDiscoveryResult:
+			discovered.Images = append(discovered.Images, typedRes)
+		case kargoapi.DiscoveryResult:
+			discovered.Results = append(discovered.Results, typedRes)
+		default:
+			return nil, fmt.Errorf(
+				"subscriber returned unrecognized result type %s",
+				reflect.TypeOf(typedRes),
```
thomastaylor312: Nit: You can just use `%T` rather than having to call `TypeOf` here

pkg/subscription/subscriber.go
```diff
@@ -0,0 +1,39 @@
...
+	// DiscoverArtifacts discovers artifacts according to the parameters of the
+	// provided kargoapi.RepoSubscription. Implementations may return a value of
+	// any type, but in practice, callers can only make sense of:
+	// - kargoapi.ChartDiscoveryResult
+	// - kargoapi.GitDiscoveryResult
+	// - kargoapi.ImageDiscoveryResult
+	// - kargoapi.GenericDiscoveryResult
```
thomastaylor312: I think this is just called `kargoapi.DiscoveryResult` now

pkg/subscription/subscriber.go
```diff
@@ -0,0 +1,39 @@
...
+	DiscoverArtifacts(
+		ctx context.Context,
+		project string,
+		sub kargoapi.RepoSubscription,
+	) (any, error)
```
thomastaylor312, line 38: Nit: I wonder if we could at least add a little bit of a sanity check by having this return a (new) `ArtifactResult` interface. It could have getters for some things like `SubscriptionName` and then as an implementor you'd at least have to implement that interface for things to compile. Feels like it would be easier to restrict now and then open up to `any` down the line if needed

pkg/webhook/kubernetes/warehouse/webhook.go
```diff
@@ -290,38 +273,73 @@ type subscriptionKey struct {
...
+					"subscription for chart %q already exists at %q",
+					sub.Chart.Name, s[*k],
+				)
+			} else {
+				errMsg = fmt.Sprintf("subscription for chart already exists at %q", s[*k])
+			}
+			return field.Invalid(f.Child("chart"), sub.Chart.RepoURL, errMsg)
+		}
+	case sub.Git != nil:
+		k := &subscriptionKey{
```
thomastaylor312: Super nit: instead of getting a pointer to `subscriptionKey` and immediately dereferencing a few lines down, we should just use non-pointer values

pkg/controller/warehouses/warehouses.go
krancour: 🤦‍♂️

api/v1alpha1/warehouse_types.go
krancour: I thought about it, but its purpose is so super specific that I don't want to grapple with how to name that type. I think I would let this be.

api/v1alpha1/warehouse_types.go
krancour: Fixed in baf36cf66acd86efad807fb0579647860fc308ff

api/v1alpha1/warehouse_types.go
krancour: Fixed in baf36cf66acd86efad807fb0579647860fc308ff

api/v1alpha1/warehouse_types.go
krancour: Yes. And it should check compare metadata as well.

Fixed in baf36cf66acd86efad807fb0579647860fc308ff

pkg/api/freight.go
krancour: Validation should be enforcing uniqueness of names across subscriptions of all types, but what you suggest wouldn't hurt.

Fixed in baf36cf66acd86efad807fb0579647860fc308ff

pkg/controller/freight/finder.go
krancour: May leave this one be. We only assign a value to `desiredOrigin` if it was `nil` to begin with, so this doesn't feel appreciably different from something like this, which I think would be regarded as fairly idiomatic. Or, at least, we do this in enough places that I wouldn't get hung up on this one occurrence.

```
func Clone(
	repoURL string,
	clientOpts *ClientOptions,
	cloneOpts *CloneOptions,
) (Repo, error) {
	if clientOpts == nil {
		clientOpts = &ClientOptions{}
	}
	if cloneOpts == nil {
		cloneOpts = &CloneOptions{}
	}
	// ...
}
```

pkg/controller/warehouses/warehouses.go
krancour: Fixed in baf36cf66acd86efad807fb0579647860fc308ff

pkg/subscription/subscriber.go
krancour: Yep. And found one other comment as well where the name change wasn't accounted for. 

Fixed in both places in baf36cf66acd86efad807fb0579647860fc308ff

pkg/webhook/kubernetes/warehouse/webhook.go
krancour: Copy/pasta from older code (where it might have made more sense, contextually).

Fixed in baf36cf66acd86efad807fb0579647860fc308ff

pkg/subscription/subscriber.go
krancour: Sigh...

That was my initial impulse and I'd tried it at first, but it didn't really make sense after I looked at the finished product. The thing is the caller only can work with four specific concrete types representing a discovery result -- the generic `kargoapi.DiscoveryResult` and the different-looking original three. In a perfect world, I wouldn't have to worry about the original three and all subscribers would return generic `kargoapi.DiscoveryResult`s instead of `any`.

If I use an interface as the return type here, a subscriber can return any arbitrary type that implements that interface, _but_ the caller would not find such a result any more acceptable than if it _hadn't_ implemented some specific interface. In short, it doesn't amount to the sort of sanity check we might wish for.

The caller, however, does perform it's own sanity check after the call to `DiscoverArtifacts()`. It asserts that the result is one of the four types it knows how to deal with.

krancour: However, if you can make this work in some way I have overlooked, I'm amenable to it.

thomastaylor312: I think that logic makes sense to me so don’t worry about changing this.

api/v1alpha1/warehouse_types.go
```diff
@@ -117,6 +134,163 @@ type WarehouseSpec struct {
 	FreightCreationCriteria *FreightCreationCriteria `json:"freightCreationCriteria,omitempty" protobuf:"bytes,5,opt,name=freightCreationCriteria"`
 }
 
+var knownKeys = []string{"chart", "git", "image"}
```
hiddeco: Nit:

```suggestion
var knownSubscriptionKeys = []string{"chart", "git", "image"}
```

would probably a bit more descriptive.

krancour: Renamed it to `legacySubscriptionTypes` in 9a466f43618ae0e9c0efdd0b9855e1804da9baf4

krancour:
Holy merge conflicts, Batman!

I'll deal with these tomorrow.

It's mainly on account of things merged earlier today in #5298. It looks worse than it is.

pkg/expressions/function/functions.go
```diff
@@ -769,6 +808,131 @@ func getChartFromDiscoveredArtifacts(artifacts *kargoapi.DiscoveredArtifacts) ex
...
+		var artifact *kargoapi.ArtifactReference
+		for _, result := range artifacts.Results {
+			if result.SubscriptionName != subName {
+				continue
+			}
+			if len(result.ArtifactReferences) > 0 {
+				artifact = &result.ArtifactReferences[0]
+			}
+		}
+
```
hiddeco, line 914: Hypothetically, we could have a `nil` artifact here if there are no matches.

Believe we should do:

```go
if artifact == nil {
    return nil, nil
}
```

pkg/demo/subscriber.go
```diff
@@ -0,0 +1,247 @@
+package datesub
```
hiddeco: Do we want to remove this before merge?

krancour: Yes. It's probably been overlooked as the commit history has grown on this, but some commits are marked as "remove me," so there's a rebase coming.

pkg/expressions/function/functions.go
krancour: Nice catch. That's was a blatant oversight. Will fix.

krancour: Fixed in https://github.com/akuity/kargo/pull/5382/commits/6afb82cbf3912510d8371517f28dd7168c1d11db

api/v1alpha1/freight_types.go
```diff
@@ -40,6 +40,9 @@ type Freight struct {
 	Images []Image `json:"images,omitempty" protobuf:"bytes,4,rep,name=images"`
 	// Charts describes specific versions of specific Helm charts.
 	Charts []Chart `json:"charts,omitempty" protobuf:"bytes,5,rep,name=charts"`
+	// Artifacts describes specific versions of artifacts other
+	// than Git repository commits, container images, and Helm charts.
+	Artifacts []ArtifactReference `json:"artifacts,omitempty" protobuf:"bytes,10,rep,name=artifacts"`
```
thomastaylor312, line 45: One last thing I just remembered is that the event types (`Freight` in particular) needs to be updated to have a new `Artifacts` field that uses the `ArtifactReference` type

docs/docs/90-api-documentation.md
```diff
@@ -1868,6 +1868,17 @@ RawFormat specifies the format for raw resource representation.
...
+<a name="github-com-akuity-kargo-api-v1alpha1-ArtifactReference"></a>
+
+### ArtifactReference
+ ArtifactReference is a reference to a specific version of an artifact.
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| artifactType | [string](#string) |  ArtifactType specifies the type of artifact this is. Often, but not always, it will be the media type (MIME type) of the artifact referenced by this ArtifactReference.   |
+| subscriptionName | [string](#string) |  SubscriptionName is the name of the Subscription that discovered this artifact.   |
+| version | [string](#string) |  Version identifies a specific revision of this artifact.   |
+| metadata | k8s.io.apiextensions_apiserver.pkg.apis.apiextensions.v1.JSON |  Metadata is a JSON object containing a mostly opaque collection of artifact attributes. (It must be an object. It may not be a list or a scalar value.) "Mostly" because Kargo may understand how to interpret some documented, well-known, top-level keys. Those aside, this metadata is only understood by a corresponding Subscriber implementation that created it.  +optional |
```
thomastaylor312, line 1880: Nit: Looks like the autogen captured the `+optional` tag in the text. I see that other places in this doc, so maybe this might be part of a bigger change later

pkg/server/get_warehouse_v1alpha1.go
```diff
@@ -65,3 +74,19 @@ func (s *server) GetWarehouse(
...
+// custom marshaling is not applied when serializing a Warehouse as part of
+// a protobuf message, so this helper can be used to compensate for that.
+func prepareOutboundWarehouse(w *kargoapi.Warehouse) error {
+	type specAlias kargoapi.WarehouseSpec
+	specJSON, err := json.Marshal(w.Spec)
+	if err != nil {
+		return err
+	}
+	newSpec := specAlias{}
+	if err = json.Unmarshal(specJSON, &newSpec); err != nil {
```
thomastaylor312, line 87: Not a huge deal or blocker for this PR but for this method and its inverse I saw earlier, this could add a decent amount of overhead to every warehouse request as JSON serialization and deserialization can only be made so fast. Even though it is annoying, it would probably be better to have logic here that converts all the internal subscriptions to the external representation (it could even reuse the logic from the marshal/unmarshal by popping that out into its own function)

We likely don't this API too much and we're going to be moving to more of a REST-ish API anyway, so this is probably fine to leave. Just calling out in case we care.

docs/docs/90-api-documentation.md
krancour: I noticed this at some point in the past. Agree it's both wrong _and_ something for another time.

pkg/server/get_warehouse_v1alpha1.go
krancour: Agree completely, and think you're correct that as we move away from connectrpc, this problem will probably right itself.

Issue #5267, opened 2025-10-22
UI crashes with TypeError when viewing projects with multiple stages or multiple image subscriptions in warehouse


tal-hason:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

The Kargo UI crashes with `TypeError: Cannot read properties of undefined (reading 'includes')` when accessing the **freight view** of certain projects. The main project view (stages and warehouses) works correctly, but clicking on freight or navigating to freight-related pages causes the crash.

The error occurs specifically for projects that have multiple stages or multiple image subscriptions in their warehouses. The error happens in the minified JavaScript at `DMe (index-Gx-3j6kW.js:1964:26155)` and causes the entire UI to become unresponsive.

# Screenshots

N/A - The UI crashes completely, showing only the error in the browser console.

# Steps to Reproduce

1. Deploy Kargo v1.8.0 with projects that have:
   - Multiple stages in a single project, OR
   - Warehouses with multiple image subscriptions
2. Access the Kargo UI
3. Navigate to the problematic project
4. **The main project view works fine** - you can see stages and warehouses
5. **Click on freight or navigate to freight-related pages**
6. The UI crashes with the TypeError

**Working projects:** Projects with single warehouses and single image subscriptions work fine for all views.

**Failing projects:** Projects with multiple stages or multiple image subscriptions work for the main view but crash when accessing freight views.

# Version

```
Server Version: ghcr.io/akuity/kargo:v1.8.0
```

# Logs

## Browser Console Error:
```
[Violation] 'setTimeout' handler took 269ms
TypeError: Cannot read properties of undefined (reading 'includes')
    at DMe (index-Gx-3j6kW.js:1964:26155)
    at xne (index-Gx-3j6kW.js:1964:26116)
    at index-Gx-3j6kW.js:1989:90819
    at Array.map (<anonymous>)
    at OLt (index-Gx-3j6kW.js:1989:90632)
    at loe (index-Gx-3j6kW.js:482:34587)
    at Loe (index-Gx-3j6kW.js:482:63698)
    at wFe (index-Gx-3j6kW.js:482:74484)
    at XFe (index-Gx-3j6kW.js:482:109030)
    at L5t (index-Gx-3j6kW.js:482:108064)

// Followed by multiple API call failures:
[...]

PR #5386, opened 2025-11-14
fix: UI crashing by image annotations array https://github.com/akuity/kargo/issues/5267

tal-hason:
### Problem

The UI was crashing with `TypeError: Cannot read properties of undefined (reading 'includes')` when:
1. Image annotations lacked the `org.opencontainers.image.source` field
2. Image `repoURL` values were undefined, null, or malformed (e.g., empty strings, paths with only slashes)

### Root Cause

Two functions were calling string methods on potentially undefined values:

1. **`urlForImage()`** in `utils/url.ts`:
   - Accessed `parts[0]` after `shift()` operations without verifying the array still had elements
   - Called `.includes()`, `.toLowerCase()`, and `.endsWith()` on potentially undefined values
   - Didn't handle edge cases where `split('/')` produced empty strings

2. **`getGitCommitURL()`** in `open-container-initiative-utils.ts`:
   - Called `.includes()` on `url` parameter without null checks
   - Function signature didn't reflect that `url` could be undefined

### Solution

Added defensive null checks and validation:

**`utils/url.ts`:**
- Filter out empty strings from `split('/')` results
- Validate `parts[0]` exists and is a non-empty string before calling string methods
- Store `firstPart` in a variable after validation to ensure type safety
- Return early for empty/malformed inputs

**`open-container-initiative-utils.ts`:**
- Added null check in `getImageSource()` before calling `getGitCommitURL()`
- Updated `getGitCommitURL()` signature to accept `string | undefined | null`
- Added type validation before calling `.includes()` on URL

### Files Changed

- `kargo/ui/src/utils/url.ts`
- `kargo/ui/src/features/freight-timeline/open-container-initiative-utils.ts`

### Testing

The fix handles edge cases including:
- Images with missing OCI source annotations
- Undefined/null/empty `repoURL` values
- Malformed image paths (e.g., `"/"`, `"//"`, `"docker.io"`)
- Images with only `revision` but no `source` in annotations

### Related Issues

Fixes crashes related to:
- https://github.com/akuity/kargo/issues/5267
- Missing OCI source annotations in freight images
- Malformed or missing image repository URLs

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/ui/src/features/project/pipelines/freight/freight-card.tsx
+++ b/ui/src/features/project/pipelines/freight/freight-card.tsx
@@ -2,7 +2,6 @@ import { useDraggable } from '@dnd-kit/core';
 import { faDocker, faGithub } from '@fortawesome/free-brands-svg-icons';
 import {
   faAnchor,
-  faArrowsLeftRightToLine,
   faCheck,
   faEllipsis,
   faPlus,
@@ -17,7 +16,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Button, Divider, Dropdown, Flex, Tooltip, Typography } from 'antd';
 import classNames from 'classnames';
 import { Duration, formatDistance, formatDuration } from 'date-fns';
-import { ReactNode, useEffect, useMemo, useRef } from 'react';
+import { useEffect, useMemo, useRef } from 'react';
 import { generatePath, useNavigate } from 'react-router-dom';
 
 import { paths } from '@ui/config/paths';
@@ -45,13 +44,9 @@ type FreightCardProps = {
   className?: string;
   promote?: boolean;
   onReviewAndPromote?(): void;
-  onExpand(): void;
   soakTime?: Duration;
   promotionEligible?: boolean;
   isPromotionEligibleLoading?: boolean;
-
-  // count of stacked freights
-  count?: number;
 };
 
 export const FreightCard = (props: FreightCardProps) => {
@@ -98,8 +93,6 @@ export const FreightCard = (props: FreightCardProps) => {
 
   const soakTimeFormatted = useMemo(() => (soakTime ? formatDuration(soakTime) : ''), [soakTime]);
 
-  let CardContent: ReactNode;
-
   const { attributes, listeners, setNodeRef, transform } = useDraggable({
     id: props.freight.metadata?.name || 'name',
     data: { originName: props.freight.origin?.name }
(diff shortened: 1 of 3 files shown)


ui/src/features/freight-timeline/open-container-initiative-utils.ts
```diff
@@ -22,13 +22,17 @@ export const getImageSource = (annotation: Annotation) => {
...
   }
 
   return getGitCommitURL(url, revision);
 };
 
-export const getGitCommitURL = (url: string, revision: string) => {
+export const getGitCommitURL = (url: string | undefined | null, revision: string) => {
+  if (!url || typeof url !== 'string') {
+    return url || '';
+  }
```
Marvin9, line 34: this is unnecessary. the function being used already ensures url is valid otherwise it won't let create the warehouse with git subscription on the first place

tal-hason: Without this the UI is still crashing.

I tested it in my cluster and only after updating this the UI crash stopped.

ui/src/utils/url.ts
```diff
@@ -6,35 +6,54 @@ const dockerHub = 'hub.docker.com';
 const officialPath = '_';
 const userPath = 'r';
 
-export const urlForImage = (image: string): string => {
-  const parts = image.split('/');
+export const urlForImage = (image: string | undefined | null): string => {
```
Marvin9, line 9: same here. repoURL for docker image subscription is validated when created so there is no chance of it being not valid. this is also unnecessary

tal-hason: As noted, only after combing the 3 change in both files ,only then the UI stopped crashing.

We can remove changes and test again.

Marvin9, line 9: Also there is no need to change this function, we already got the root cause fix merged in main. Please test the changes against that.

If I reverse engineer the failure of this function the only possibility is if you have any of URL exactly like this

```
docker.io/library
```

ui/src/features/freight-timeline/open-container-initiative-utils.ts
Marvin9: That sounds strange. This only fails when you have any freight which has git subscription with EMPTY repoURL. The thing is you can't even create a subscription like that!

tal-hason: So maybe the issue is much deeper, maybe when a warehouse created It don't have. A clear separation between the subscription types?

Marvin9:
@tal-hason please build against latest main and try to test that. If that doesn't work then let me know. Thanks for collaboration

ui/src/utils/url.ts
Marvin9: I want to merge these changes but I am not yet convinced that these are necessary. So I am trying to dig deeper. The only reasoning for those failures are already caught (https://github.com/akuity/kargo/issues/5267#issuecomment-3532532484) and PR is merged for the same

ui/src/features/freight-timeline/open-container-initiative-utils.ts
Marvin9: You can dig into manifests of Stage and Freight. Look specifically for `stage.status.freightHistory` in Stage. And somewhere in `commits` in Freight. See if repoURL is missing in any of those

Marvin9:
@tal-hason did you get a chance to test?

tal-hason:
Hi @Marvin9 , not yet busy week, i will try to check it tomorrow, or Friday out of work hours :-)

Marvin9:
Closing it for now. Feel free to reopen it once you test and still find an issue. Thanks!

Issue #5299, opened 2025-10-28
allow yaml-update promotion step to create keys when missing


mihuross:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

yaml-update promotion step is now (v1.8.0) able to update keys that are present in yaml document, but cannot create a new ones, when they are missing.

# Motivation

I would like to be able to include metadata from promotion like freight name in the yaml document even though the yaml doesn't contain prepared placeholder for it (thus reducing some boilerplate, when configuring source yaml document first time).

krancour:
This is a duplicate of #3752, which is closed only because it went stale a long time ago.

I will leave this one open rather than revive that one.

For starters, I very much want to do this. It is, however, quite technically difficult (or, at least, it was at the time #3752 was opened). For to save everyone some clicks, here's the detailed explanation from that previous thread:

> This is really hard to do, and probably not for the reasons you might guess.
>
> When messing around with a YAML file, two very important things are preserving the author's comments and formatting choices.
>
> This means the simple/easy thing of just unmarshaling the YAML, making changes to the structured data, and then marshaling it back to YAML isn't an option.
>
> The current implementation works by finding the exact row/col position in the file of the value needing to be updated and just substitutes the new value for the old.
>
> Anything that can't work like that poses a bit of a problem.
>
> I'm certain there are packages out there that may be able to help with doing all of this more elegantly. Kustomize, for instance has no trouble doing things like this, but their YAML bits are internal and not easily reusable. Any help in identifying a library that allows modifications to YAML files while preserving formatting and comments would be a step toward being able to do something like this. (Note that I'd rather the solution were not exec'ing to yq.)

What I'm going to do for now is assign this to @fykaa to research what newer/improved/better options may now exist for modifying YAML files while preserving comments and formatting choices.

I really cannot guarantee this is something we can do right now, but we're at least going to give it a look and find out.

mihuross:
Ah, sorry for the double issue, I was really searching only in open issues. My expectation is that this is completely minor issue, hence no motivation to put effort into it. I have some POC for the modifying the yaml so the comments and other formatting are preserved with standard yaml library, so I think it should be possible to do it. If you want to, I can try to create a PR for the feature into kargo (I'm probably the most motivated person here).

krancour:
That's something we'd love to take a look at.

PR #5394, opened 2025-11-16
feat(yaml-updater): add ability to append to mapping keys

mihuross:
If the path points to non existing key in mapping node and would otherwise be a terminal one, add ability to append that key with new value.

I was trying to find a usable implementation somewhere else, but ultimately failed to do so :(

This is a best approach I could come up with, that is minimal and supports adding a new key to mapping node while trying to preserve indentations. This can still screw up formatting rules though (e.g. if formater requires specific key ordering in mapping). For the similar reason, append to empty mapping node is not possible, since it would have to guess what is the correct indentation step.

Adding ability to append to lists might work in similar way. 

Related to https://github.com/akuity/kargo/issues/5299

As described in the issue, I think this is still rather an edge case to solve (although very useful for me), so adding extra code just for this, might not be a good trade of.

### DIFF
--- a/pkg/yaml/yaml.go
+++ b/pkg/yaml/yaml.go
@@ -66,19 +66,24 @@ func SetValuesInBytes(inBytes []byte, updates []Update) ([]byte, error) {
 	}
 
 	type change struct {
-		col   int
-		value any
+		col     int
+		value   any
+		key     string
+		newitem bool
 	}
 	changesByLine := map[int]change{}
 	for _, update := range updates {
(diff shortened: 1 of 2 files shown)


krancour:
Closing this since:

1. The linked issue was closed some time ago.
2. Prior discussion had established that any acceptable solution to this would require preserving existing comments and formatting choices in the modified file. The description states that this PR does not achieve that. So based on that alone, this isn't something we would merge.

Issue #5149, opened 2025-10-02
Git Credential UI should inputting support SSH


jessesuen:
# Proposed Feature

Kargo supports git username/password auth as well as SSH. But the UI only shows input option for username and password.

See: https://docs.kargo.io/user-guide/security/managing-credentials/#repository-credentials-as-secret-resources

We need the following changes in UI
1. Option for git repos to be input using SSH-style URLs, e.g. `git@github.com:example/repo.git`
2. In lieu of username/password, they would need a text box to input a A PEM-encoded SSH private key:

# Motivation

It's not apparent from UI that Kargo supports SSH git credentials

krancour:
Before we dive head first into this, I think there is more to consider here. I've brought this up before offline, but regrettably didn't open an issue.

Credentials come in so many shapes. It's not _just_ SSH keys that aren't supported by the UI.

Ref: https://docs.kargo.io/user-guide/security/managing-credentials#other-forms-of-credentials

Support in the CLI is equally poor.

Things like this work:

```
kargo create credentials \
  --project kargo-demo my-credentials \
  --git \
  --repo-url https://github.com/example/kargo-demo.git \
  --username my-username \
  --password my-personal-access-token
```

But similar flags don't exist for managing any other sort of credentials. It's easy to see we'd get into "flag hell" if we tried to extend what we see above to work with other kinds of credentials.

At present, the only way to create or update anything other than username/password (or PAT, which is really just a password), is to create the manifest and apply it to the Kargo control plane using `kubectl apply` or `kargo apply`.

Instead of solving this only for SSH and only in the UI, I believe we need a generic approach and need to address it in the CLI as well as the UI.

My suggestion when this had been discussed once before was that we could create "Mad Libs" for different kinds of credentials and users can just "fill in the blanks."

On the CLI end, that might look something like:

```
kargo create credentials my-creds --template ssh
```

The SSH key would be collected interactively and a Secret manifest posted to the generic create endpoint. Update would work similarly. This approach would work equally well for things like authenticating using a GitHub App. The keys for which we interactively collect values would simply be different.

On the UI end, you could pick a template from a drop down and it determines what fields (keys) will appear in a form.

This could even work _without_ templates, so the benefits are extended to any kind of credential that we don't even really known the details of. `kargo create credentials my-creds` with no template specified could interactively collect arbitrary key/value pairs from the user.

tbh, this isn't a lot of work, and would even allow us to deprecate some existing jankiness.

cc @jessesuen

[...]

PR #5404, opened 2025-11-19
feat(ui): ssh private key support

Marvin9:
https://github.com/akuity/kargo/issues/5149

Screenshot 2025-11-19 at 5 29 24 PM (2)

### DIFF
--- a/ui/src/features/project/settings/views/credentials/create-credentials-modal.tsx
+++ b/ui/src/features/project/settings/views/credentials/create-credentials-modal.tsx
@@ -74,7 +74,7 @@ type Props = ModalComponentProps & {
 };
 
 export const CreateCredentialsModal = ({ project, onSuccess, editing, init, ...props }: Props) => {
-  const [isSSHRepoUrl, setIsSSHRepoUrl] = useState(init?.stringData['sshPrivateKey'] !== undefined);
+  const [isSSHRepoUrl, setIsSSHRepoUrl] = useState(init?.stringData['username'] === '');
 
   const { control, handleSubmit, watch } = useForm({
     defaultValues: { ...constructDefaults(init, props.type === 'generic' ? props.type : 'git') },
--- a/ui/src/features/project/settings/views/credentials/credentials.tsx
+++ b/ui/src/features/project/settings/views/credentials/credentials.tsx
@@ -46,14 +46,7 @@ export const CredentialsSettings = () => {
   const specificCredentials: Secret[] = listCredentialsQuery.data?.credentials || [];
   const genericCredentials: Secret[] = listSecretsQuery.data?.secrets || [];
 
-  const { show: showCreate } = useModal((p) => (
-    <CreateCredentialsModal
-      type='repo'
-      project={name || ''}
-      onSuccess={listCredentialsQuery.refetch}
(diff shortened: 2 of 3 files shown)


api/service/v1alpha1/service.proto
```diff
@@ -1064,6 +1064,7 @@ message CreateCredentialsRequest {
   string username = 6;
   // password is the password or token for authentication.
   string password = 7;
+  string ssh_private_key = 9;
```
jessesuen, line 1067: We also need to eventually support [GitHub App Authentication](https://docs.kargo.io/user-guide/security/managing-credentials/#github-authentication-options). Which has the fields:
* `githubAppClientID`
* `githubAppPrivateKey`
* `githubAppInstallationID`

Is continually expanding `CreateCredentialsRequest` API the sustainable approach to this?

Should UI switch to the Apply endpoint and just create `Secret` resources with baked-in understanding of the various fields (username/password/privatekey/githubXXXX/etc...)? Then the ConnectRPC API layer never needs to updated.

Marvin9: Oh. I thought its only `sshPrivateKey`. You are right, we should start using generic secrets.

Are you suggesting is to remove CreateCredentials approach? even for normal credentials?

Marvin9: cc @jessesuen @krancour

krancour: @jessesuen I had a plan already. It's written down in some issue somewhere. I just need to find it. Credentials come in all different shapes. What I want to do is have Secret templates for all different kinds of creds and you fill in the blanks for whatever type of cred like you're doing a Mad Lib. Adding support for a new kind of credential would be as easy as providing a new template. In the event that a template does not exist for some sort of credential, the option would remain to just give us the cred type and data block.

But, imho, that more comprehensive approach to UI/CLI support for different kinds of credentials ought not to block this ssh-key-specific incremental improvement.

Having said that, I've reached the point of discouraging people from using SSH with Kargo for the simple reason that none of the providers' APIs accept ssh keys as credentials. So when you choose to use ssh creds for basic git operations like cloning, pushing, etc., you force yourself into having to maintain a second set of creds for the same repo(s) so that you can open and merge PRs, etc.

Let's chat about it face to face in the coming week.

Marvin9: @krancour are we going with this or should close PR for now?

krancour:
@Marvin9 I think @jessesuen should make this call. We had a conversation yesterday about deprecating ssh support in v1.10.0 and maybe removing it many releases from now. Probably no sooner than v1.14.0, if ever. So I'm uncertain whether we want to introduce _new_ ssh-related features.

Small tangent, which I've mentioned before: I think how we define creds, in general, is in need of an overhaul, which metadata of some sort describing the different "shapes" of creds so that the UI and CLI alike can better accommodate the many creds that _aren't_ simply username and password. I'm going to put together what I'll call a "creds v2" proposal, but this is a long-term thing.

In the meantime, let's have @jessesuen's make the call on whether to include this feature or not.

krancour:
Closing in favor of #5978

Issue #5127, opened 2025-09-29
git-merge-pr: add merge queue handling to Github provider


fykaa:
current Github provider implementation in `MergePullRequest` assumes that calling the Github API's merge endpoint immediately results in a merged pull request. But when Github repositories use merge queues, there can be a delay between initiating the merge and the actual merge completion

## Current Behavior

After calling `client.MergePullRequest()`, the code immediately fetches the updated PR and expects it to be in a merged state. Ref https://github.com/akuity/kargo/pull/3958#discussion_r2388157523

## Expected Behavior

The implementation should account for merge queue scenarios where:
1. The merge request is accepted but the PR enters a queue
2. The actual merge happens asynchronously
3. The PR state may not immediately reflect as "merged"

When merge queues are enabled, the function may return incorrect merge status or fail entirely, causing issues with Kargo's PR automation workflows

TODO: Research Github's merge queue API behavior and response codes and determine if other git providers have similar queuing mechanisms. We might consider polling strategies or webhook-based approaches for queued merges

Note: The feature is being marked as beta partly due to this limitation.

krancour:
@fykaa small bit of clarification... when I mentioned this earlier, I used GitHub as a prime example because I knew with certainty that it has this feature.

Some of the other providers _may_ have something similar. That will require some research.

Just mentioning this so that the issue isn't scoped too narrowly to _only_ GitHub.

fykaa:
> Some of the other providers may have something similar. That will require some research.

Yes, the issue TODO mentions we need to check the same behavior across other Git providers as well

krancour:
Sorry... my comment was mainly driven by the title, which I've now changed for clarity.

fykaa:
We can choose Polling over Webhooks 
FOR 
- simpler implementation, as we will have to set up the infrastructure for receiving webhooks
- also polling can be more reliable in restricted environments

We can choose Webhooks over Polling
FOR
- webhooks would be better if we want real time updates
- [GitHub says this about webhooks](https://docs.github.com/en/developers/webhooks-and-events/webhooks/about-webhooks) . They recommend [Webhooks over polling](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#avoid-polling) for their platform

fykaa:
3 out of 5 providers have asynchronous merge mechanisms that will cause issues with the current `MergePullRequest` implementation:

[...]

PR #5405, opened 2025-11-19
docs(git-merge-pr): note api limitation for merge queues in git-merge-pr step

fykaa:
closes https://github.com/akuity/kargo/issues/5127

this PR adds `Queued` field to gitprovider.PullRequest struct to track queued state and updates GitHub provider to detect HTTP 202 responses and set Queued flag

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
@@ -10,19 +10,21 @@ description: Merges an open pull request.
 `git-merge-pr` merges an open pull request. This step commonly follows a
 [`git-open-pr`](git-open-pr.md) step.
 
-:::caution
+:::important
 This step only executes synchronous merges. It can neither initiate an
 asynchronous merge by placing a PR on a merge queue (or similar), nor can it
 recognize when an open PR is already _in_ a merge queue (having been placed
 there by someone or something else), and thus cannot wait for an aynchronous
 merge in-progress to complete.
+:::
 
-If a repository _enforces_ merge queues, the behavior when attempting a
-synchronous merge depends permissions.
-
-- GitHub: If a merge queue is required, the PR will either be merged directly if the
-  token has bypass permissions, or it will fail with an error if it does not.
-- GitLab: This step will always merge directly without using merge trains.
+:::caution
+__GitHub__ repositories can be configured with branch protection rules that
+require PRs to be merged via a merge queue. When such a rule is in place, the
+results of the `git-merge-pr` step attempting a synchronous merge will depend
+upon permissions. With sufficient permissions to bypass branch protection rules,
+the merge queue will be bypassed. Without such permissions, the step's attempt
+to merge will fail.
 :::
 
 ## Configuration
@@ -64,9 +66,8 @@ steps:
 ### Merge with Wait
 
 This example demonstrates merging a pull request with waiting enabled. If the pull
-request is not immediately ready to merge (e.g., due to pending CI checks, required
-reviews, or temporary conflicts), the step will return a running status and Kargo
-will retry it on the next reconciliation.
+request is not yet mergeable for any reason, the step will return a running
+status and Kargo will retry it on the next reconciliation.
 
 ```yaml
 steps:
@@ -76,9 +77,3 @@ steps:
     prNumber: 42
     wait: true
 ```
-
-:::note
-If the repository requires using a merge queue and the token lacks bypass permissions,
-the step will fail regardless of the `wait` setting, as the step does not add PRs to
-merge queues.
-:::
--- a/pkg/promotion/runner/builtin/schemas/git-merge-pr-config.json
+++ b/pkg/promotion/runner/builtin/schemas/git-merge-pr-config.json
@@ -27,7 +27,7 @@
       },
       "wait": {
         "type": "boolean",
-        "description": "If true, the step will return RUNNING instead of FAILED when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. This step performs synchronous merges only and does not add PRs to merge queues. Default is false."
+        "description": "If true, the step will return RUNNING instead of FAILED when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. Default is false."
       }
     }
 }
--- a/ui/src/gen/directives/git-merge-pr-config.json
+++ b/ui/src/gen/directives/git-merge-pr-config.json
@@ -32,7 +32,7 @@
   },
   "wait": {
    "type": "boolean",
-   "description": "If true, the step will return RUNNING instead of FAILED when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. This step performs synchronous merges only and does not add PRs to merge queues. Default is false."
+   "description": "If true, the step will return RUNNING instead of FAILED when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. Default is false."
   }
  }
 }


pkg/gitprovider/github/github_test.go
thomastaylor312, line 1: There should be a test in here that exercises the new behavior of returning a 202

pkg/gitprovider/github/github.go
```diff
@@ -321,7 +321,7 @@ func (p *provider) MergePullRequest(
 	}
 
 	// Merge the PR
-	mergeResult, _, err := p.client.MergePullRequest(
+	_, resp, err := p.client.MergePullRequest(
```
thomastaylor312: Is there a case where we would need the merge result here? For example, could the result be `nil` with no `err` and another response code? I don't think so, but wanted to confirm that there isn't a case that should be handled here

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -242,7 +242,21 @@ func (p *provider) MergePullRequest(
...
+		return &pr, true, nil
+	}
+
+	// GitLab merge trains keep the MR in "opened" state when queued for merging.
+	// The GitLab API returns HTTP 200 (not 202) with the MR object in "opened" state.
+	// We detect queuing by checking if the MR state is still "opened" after a
+	// successful AcceptMergeRequest call when pipeline or policy checks are pending.
+	if resp != nil && updatedMR.State == "opened" {
+		pr.Queued = true
+	}
```
thomastaylor312: This is _probably_ ok, but it might be more robust to check when the state is still `opened` that the PR exists in the merge train using the merge train API. I just worry it could get rejected from the merge queue or some other failure could happen.

I think I'll consider this optional, depending on the amount of effort required to do the check of the merge queue

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -242,7 +242,21 @@ func (p *provider) MergePullRequest(
...
-	return &pr, true, nil
+	if pr.Merged {
+		return &pr, true, nil
+	}
+
+	// GitLab merge trains keep the MR in "opened" state when queued for merging.
+	// The GitLab API returns HTTP 200 (not 202) with the MR object in "opened" state.
+	// We detect queuing by checking if the MR state is still "opened" after a
+	// successful AcceptMergeRequest call when pipeline or policy checks are pending.
+	if resp != nil && updatedMR.State == "opened" {
```
thomastaylor312: Nit: make `opened` a constant or use an existing const (if it exists) from the go client

pkg/gitprovider/gitlab/gitlab_test.go
```diff
@@ -376,6 +410,9 @@ func TestMergePullRequest(t *testing.T) {
 			if tc.expectPR {
 				require.NotNil(t, pr)
 				require.Equal(t, tc.id, pr.Number)
+				if tc.name == "merge queued (merge train)" {
```
thomastaylor312: This is going to be brittle if we change test names. I would add a new `expectQueued` field instead. This would be good as well so you can test the negative (that it _wasn't_ queued) in the other tests

pkg/gitprovider/gitprovider.go
```diff
@@ -118,6 +118,17 @@ type PullRequest struct {
 	HeadSHA string `json:"headSHA"`
 	// CreatedAt is the time the pull request was created.
 	CreatedAt *time.Time `json:"createdAt"`
+	// Mergeable indicates whether the underlying hosting provider considers
+	// the PR mergeable (checks passed, no conflicts, etc.). A nil value
+	// indicates the provider could not determine mergeability.
+	Mergeable *bool `json:"mergeable,omitempty"`
+	// Draft indicates whether the PR is a draft (true) or not (false).
+	Draft *bool `json:"draft,omitempty"`
```
thomastaylor312: Mergeable makes sense why it is a pointer, but I don't think this one needs to be a pointer as it will always be either true or false

pkg/promotion/runner/builtin/git_pr_merger.go
```diff
@@ -112,6 +112,56 @@ func (g *gitPRMerger) run(
...
+	if currentPR.Merged {
+		return promotion.StepResult{
+			Status: kargoapi.PromotionStepStatusSucceeded,
+			Output: map[string]any{stateKeyCommit: currentPR.MergeCommitSHA},
+		}, nil
+	}
+
+	// If the provider indicates the PR is already queued for merge, don't
+	// attempt to re-request the merge. Respect the 'wait' config.
+	if currentPR.Queued {
```
thomastaylor312: I didn't see anything that set this value in `GetPullRequest` of the github or gitlab providers. Did I miss something?

If not, this should be added and a test to check for it as well

pkg/promotion/runner/builtin/git_pr_merger.go
```diff
@@ -134,12 +184,27 @@ func (g *gitPRMerger) run(
 		if merged {
 			break
 		}
-		if i < maxMergeAttempts {
+		if i < maxMergeAttempts-1 {
```
thomastaylor312: Why was this added? The `range` up above is an exclusive range, so if the number is 10, then it does 0-9

pkg/promotion/runner/builtin/git_pr_merger_test.go
thomastaylor312, line 1: I'd like to see some new tests here exercising the new `if` branches around `Queued`

pkg/gitprovider/github/github.go
fykaa: from my research [github merge API](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#merge-a-pull-request) has three possible outcomes:

http 200 + *github.PullRequestMergeResult - merge succeeded synchronously
http 202 + *github.PullRequestMergeResult - merge queued/async
`err` - merge failed

looking at the go-github library's implementation, when there's an error, err will be non-nil. When there's no error, we get either 200 or 202. and response status code is in resp.StatusCode (which we are already checking)

i removed the merge result because:

- the result contains SHA (which we are getting from the updated PR anyway)
- it contains merged boolean (which we're getting from `updatedPR.MergedAt != nil`)
- it contains message string (just descriptive text)

pkg/gitprovider/gitlab/gitlab.go
fykaa: basically, I'm just using the [mergeRequestsService](https://docs.gitlab.com/api/merge_requests/) API since it's available in the free tier. The [merge train API](https://docs.gitlab.com/api/merge_trains/) falls under the premium and ultimate plans, so if we call it and get a 403 Forbidden because the feature isn't enabled, we will need a fallback.

I can make the **merge train API** call and:
- If 200: Trust the status field to determine if queued
- If 404: Not queued
- If 403: Fall back to checking state == "opened" (current behavior)

pkg/promotion/runner/builtin/git_pr_merger.go
fykaa: that was because GitHub doesn't expose merge queue status in the GET PR endpoint

fykaa: but if we still need it, I'll have to use Github's graphql api through mergeQueueEntry field. But I feel it would add additional dependency using graphql client, and complexity due to mixed REST/graphql usage in same provider.

fykaa: for gitlab though, I have added it.

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -156,6 +172,21 @@ func (p *provider) GetPullRequest(
...
+	// Check if MR is in merge train when state is opened
+	if pr.Open {
+		mergeTrain, mtResp, err := p.mergeTrainClient.GetMergeRequestOnAMergeTrain(
+			p.projectName, int(id),
+		)
+		if err == nil && mtResp != nil && mtResp.StatusCode == 200 && mergeTrain != nil {
+			if mergeTrain.Status != "merged" && mergeTrain.Status != "skip_merged" {
+				pr.Queued = true
+			}
+		} else if mtResp != nil && mtResp.StatusCode == 403 {
```
hiddeco: Is there a possibility that the `HTTP 403` error could conflict with someone providing invalid API credentials?

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -156,6 +172,21 @@ func (p *provider) GetPullRequest(
 		return nil, fmt.Errorf("unexpected nil merge request")
 	}
 	pr := convertGitlabMR(glMR.BasicMergeRequest)
+	// Check if MR is in merge train when state is opened
+	if pr.Open {
+		mergeTrain, mtResp, err := p.mergeTrainClient.GetMergeRequestOnAMergeTrain(
+			p.projectName, int(id),
+		)
+		if err == nil && mtResp != nil && mtResp.StatusCode == 200 && mergeTrain != nil {
+			if mergeTrain.Status != "merged" && mergeTrain.Status != "skip_merged" {
```
hiddeco:
```suggestion
			if mergeTrain.Status != mrStateMerged && mergeTrain.Status != "skip_merged" {
```

Plus an additional constant for `skip_merged`?

pkg/gitprovider/gitlab/gitlab.go
fykaa: yes, http 403 can also means authentication failure, valid token but lacking required scopes, GitLab instance blocking the request OR that the merge trains feature is disabled

pkg/gitprovider/gitprovider.go
```diff
@@ -118,6 +118,16 @@ type PullRequest struct {
 	HeadSHA string `json:"headSHA"`
 	// CreatedAt is the time the pull request was created.
 	CreatedAt *time.Time `json:"createdAt"`
+	// Mergeable indicates whether the underlying hosting provider considers
+	// the PR mergeable (checks passed, no conflicts, etc.). A nil value
+	// indicates the provider could not determine mergeability.
```
krancour: > A nil value indicates the provider could not determine mergeability.

I think this might be misinterpreted as something other than what you were trying to convey.

This could reasonably be interpreted as any one given provider attempting to assert mergeability and failing to do so. i.e. GitHub responds with 🤷‍♂️

I suspect what you mean to convey is that some Git hosting providers just have no concept of this, so this field is, in their cases, unused?

It's possible this could be made more clear here.

But this is a nit.

pkg/gitprovider/github/github.go
```diff
@@ -244,6 +244,8 @@ func (p *provider) GetPullRequest(
 		return nil, fmt.Errorf("unexpected nil pull request")
 	}
 	pr := convertGithubPR(*ghPR)
+	// NOTE: GitHub doesn't provide merge queue status in the Get PR response.
+	// The Queued field will only be set after attempting to merge (MergePullRequest).
```
krancour: > The Queued field will only be set after attempting to merge (MergePullRequest).

Can you clarify what `Queued` field you are referring to? I cannot find any such field.

pkg/gitprovider/github/github.go
krancour: I fell deep down this rabbit hole this afternoon. I am less than satisfied with the docs for GitHub REST APIs, so I had to resort to a whole lot of experimentation to learn the mechanics of all of this.

I need to challenge you on some of this, or at minimum, get some clarification.

> http 202 + *github.PullRequestMergeResult - merge queued/async

I see no evidence of this possibility in the docs. My own experiments also do not support this conclusion. As far as I've been able to discern, if a merge queue is required and the token has adequate permissions to do so, this operation _bypasses_ the merge queue entirely. If the token lacks those permissions, the operation results in a 405 and an error message along the lines of "Changes must be made through the merge queue."

Can you provide references for the notion of a merge operation putting something on the queue and a 202 being returned. None of the links you provided said anything of the sort, unless I'm blind and am not seeing it (which is possible). Alternatively, can you point me to any experiments that led you to this notion?

The reason I'm so rigorously challenging this idea is that **_IF_** the merge operation never puts anything on a queue, and that's the only API we're dealing with here (i.e. _not_ using the graphql to access separate merge queue APIs), I'm not sure there is actually any scenario possible here where this step needs to contend with asynchronous merges. (With respect to GitHub, that is. I've not looked critically at the other providers.)

If this is the case, we could go in one of two directions. One would be that we do start using additional APIs to figure out if the PR _should_ be merged with a merge queue, and then make that happen. The other, easier route is to say explicitly in the docs that GitHub merge queues are not supported and that with sufficient permissions, merge queues will be bypassed, and without those permissions, any attempted merge will fail. We'd need to subject the other providers to further scrutiny to see if we can get away with the same for them or if they force our hand on it.

Since this issue has already taken up a lot more time than I'd have liked, I believe I favor the second option. Again, all of this is predicated on whether you can either refute or corroborate my findings.

Somewhat, but not entirely, separate from all of this, the experiments I undertook while trying to learn the mechanics of PRs + merge queues also led me to question whether the step, both in this PR and as originally implemented, might be merging things it ought not.

Again, the docs fall short, but as far as I have been able to determine, `mergeable` only conveys whether a PR is mergeable _purely from a Git perspective_. i.e. No divergent history from the target branch, merge conflicts, etc. _It seems that branch protections, checks, etc. have no bearing at all on the value of that field._ The separate field `mergeable_state` has more information about whether the merge can proceed _from a GitHub perspective_. I notice we're not ever looking at that field. As with the merge queue, the story here is similar -- with sufficient permissions, checks can be bypassed and the merge will proceed. Without them, the merge is blocked.

We should not rely on the token _not_ having sufficient permissions to bypass required checks -- it may. So we really ought to be looking at that field to identify cases where it would be premature to attempt a merge.

krancour: You and I are meeting Tuesday, so please spend some time playing experimenting to see if you reach the same conclusions as me.

Here's some standalone code I used in my experiments.

```
package main

import (
	"context"
	"flag"
	"fmt"
	"log"
	"os"

	"github.com/google/go-github/v57/github"
	"golang.org/x/oauth2"
)

const (
	owner = ""
	repo  = ""
	token = ""
)

func main() {
	prNumber := flag.Int("pr", 0, "PR number to merge")
	flag.Parse()

	if *prNumber == 0 {
		fmt.Println("Usage: merge-pr -pr <number>")
		os.Exit(1)
	}

	// Create authenticated client
	ctx := context.Background()
	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
	tc := oauth2.NewClient(ctx, ts)
	client := github.NewClient(tc)

	// Get a PR
	pr, _, err := client.PullRequests.Get(ctx, owner, repo, *prNumber)
	if err != nil {
		log.Fatalf("Error fetching PR: %v", err)
	}

	fmt.Printf("State: %v\n", pr.GetState())
	fmt.Printf("Mergeable: %v\n", pr.GetMergeable())
	fmt.Printf("Mergeable State: %v\n", pr.GetMergeableState())

	// Attempt to merge the PR
	_, resp, err := client.PullRequests.Merge(ctx, owner, repo, *prNumber, "merging", &github.PullRequestOptions{
		MergeMethod: "squash",
	})

	if err != nil {
		log.Fatalf("Error merging PR: %v", err)
	}

	if resp.StatusCode != 200 {
		log.Fatalf("Merge failed with status code: %d", resp.StatusCode)
	}

	if pr, _, err = client.PullRequests.Get(ctx, owner, repo, *prNumber); err != nil {
		log.Fatalf("Error fetching PR: %v", err)
	}

	fmt.Printf("State: %v\n", pr.GetState())
	fmt.Printf("Mergeable: %v\n", pr.GetMergeable())
	fmt.Printf("Mergeable State: %v\n", pr.GetMergeableState())
}

```

pkg/gitprovider/github/github.go
fykaa: the `Queued` field is on the `gitprovider.PullRequest` struct (defined in `gitprovider.go`), since its not on the github API's pull request object, I've made it more clear now where `pr.Queued` refers to the field on the returned `gitprovider.PullRequest` struct

fykaa:
the test and lint checks broke after merging the [gitlab/client-go dependency](https://github.com/akuity/kargo/pull/5437) update. The gitlab SDK now expects `int64` for merge request parameters, but interface still uses `int`. I'm fixing it now.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -10,6 +10,20 @@ description: Merges an open pull request.
 `git-merge-pr` merges an open pull request. This step commonly follows a
 [`git-open-pr`](git-open-pr.md) step.
 
+:::caution
+This step only supports synchronous merges due to API limitations of the Git providers.
+It does not work with merge queues or wait for queued merges to finish. If a
+repository enforces merge queues, the behavior depends on the token permissions.
```
krancour: I would be a lot more specific than "does not work with."

Something like:

```suggestion
This step only executes synchronous merges. It can neither initiate an
asynchronous merge by placing a PR on a merge queue (or similar), nor can it
recognize when an open PR is already _in_ a merge queue (having been placed
there by someone or something else), and thus cannot wait for an aynchronous
merge in-progress to complete.

If a repository _enforces_ merge queues, the behavior when attempting a
synchronous merge depends permissions.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -10,6 +10,20 @@ description: Merges an open pull request.
...
+
+- GitHub: If a merge queue is required, the PR will either be merged directly if the
+  token has bypass permissions, or it will fail with an error if it does not.
+- GitLab: If a merge train is required, the merge will either fail or can be bypassed
+  based on permissions.
```
krancour: GitLab actually has no feature that requires PRs to be merged using a merge train. It is always optional.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -10,6 +10,20 @@ description: Merges an open pull request.
...
+
+If your repo depends on merge queues, you must either give the service account
+permission to bypass them or add PRs to the queue manually outside of Kargo.
```
krancour: That's not really a great recommendation. We explained the limits of the feature. We don't need to suggest workarounds as feeble as this, which probably create more problems than they solve.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -60,3 +75,9 @@ steps:
     prNumber: 42
     wait: true
 ```
+
+:::note
+If the repository requires using a merge queue and the token lacks bypass permissions,
+the step will fail regardless of the `wait` setting, as the step does not add PRs to
+merge queues.
+:::
```
krancour: Is this true? Under many circumstances, the step does not even attempt to merge and would therefore not fail.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
fykaa: since using the GitLab merge requests API endpoint to merge which bypasses the merge train queue entirely, even if merge trains are enabled on the project, I can remove this caution for gitlab?

krancour: Nevermind what the API can or cannot do. GitLab has no way of _requiring_ that MRs be merged using a merge train. So yes. You can remove that warning because its entire premise is incorrect.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -10,6 +10,21 @@ description: Merges an open pull request.
...
+there by someone or something else), and thus cannot wait for an aynchronous
+merge in-progress to complete.
+
+If a repository _enforces_ merge queues, the behavior when attempting a
+synchronous merge depends permissions.
+
+- GitHub: If a merge queue is required, the PR will either be merged directly if the
+  token has bypass permissions, or it will fail with an error if it does not.
+- GitLab: This step will always merge directly without using merge trains.
+:::
```
krancour: I'm still not understanding why we're mentioning GitLab here at all.

We said, "If a repository enforces merge queues..." and we then say something about GitLab, where repos cannot be configured to enforce the use of a merge train.

GitHub is the only Git hosting provider for which we need to offer any guidance on this.

```suggestion
:::important
This step only executes synchronous merges. It can neither initiate an
asynchronous merge by placing a PR on a merge queue (or similar), nor can it
recognize when an open PR is already _in_ a merge queue (having been placed
there by someone or something else), and thus cannot wait for an aynchronous
merge in-progress to complete.
:::

:::caution
__GitHub__ repositories can be configured with branch protection rules that
require PRs to be merged via a merge queue. When such a rule is in place, the
results of the `git-merge-pr` step attempting a synchronous merge will depend
upon permissions. With sufficient permissions to bypass branch protection rules,
the merge queue will be bypassed. Without such permissions, the step's attempt
to merge will fail.
:::
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -60,3 +76,9 @@ steps:
     prNumber: 42
     wait: true
 ```
+
+:::note
+If the repository requires using a merge queue and the token lacks bypass permissions,
+the step will fail regardless of the `wait` setting, as the step does not add PRs to
+merge queues.
+:::
```
krancour: Upon further thought, I think it's better to just get rid of this. We already said enough about behavior when a merge queue is required.

```suggestion
```

pkg/promotion/runner/builtin/schemas/git-merge-pr-config.json
```diff
@@ -27,7 +27,7 @@
       },
       "wait": {
         "type": "boolean",
-        "description": "If true, the step will return RUNNING instead of FAILED when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. Default is false."
+        "description": "If true, the step will return RUNNING instead of FAILED when the PR is not yet mergeable. The merge will be retried on the next reconciliation until it succeeds or times out. This step performs synchronous merges only and does not add PRs to merge queues. Default is false."
```
krancour: I would undo this change.

I _do_ understand why you may have felt compelled to mention something about this here: It's because an asynchronous merge, if the step could _theoretically_ do it, would require further waiting after initiating the merge, and you may have wanted readers to know that's something that `wait` does not actually wait for.

But with that being said, I don't think it's necessary to cover such theoretical scenarios. Since the step never initiates async merges, it never needs to worry about the fact that it cannot wait for them. (Except in the _extreme_ edge case where someone or something else initiated an async merge.) imho, this is one of those situations where giving people more information than what they need doesn't accomplish anything other than possibly confusing them.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -49,8 +64,9 @@ steps:
 ### Merge with Wait
 
 This example demonstrates merging a pull request with waiting enabled. If the pull
-request is not immediately ready to merge, the step will return a running status and
-Kargo will retry it later.
+request is not immediately ready to merge (e.g., due to pending CI checks, required
+reviews, or temporary conflicts), the step will return a running status and Kargo
+will retry it on the next reconciliation.
```
krancour: imo, this reads better if shortened.

```suggestion
request is not yet mergeable for any reason, the step will return a running
status and Kargo will retry it on the next reconciliation.
```

Issue #5105, opened 2025-09-24
support http plain text response body


yiran0427:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable. v1.7.5
* [] I've pasted logs, if applicable.

# Proposed Feature
the http promotion step should be able to handle response body that is plain text. 

# Motivation

use case: i need to curl `http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity` for a auth token, and the response body here is only a string token.

current behaviour: 
step configuration
```
outputs:
- fromExpression: response.body
  name: token
```
output:
```
{
 "token": {}
}
```

i have a PR opened https://github.com/akuity/kargo/pull/5106

codyja:
I think I may be hitting the same issue here trying to POST to a microsoft webhook api. It's response is just the number 1 during a successful POST. 

> step "promotion::step-4" met error threshold of 1: error running step "promotion::step-4": error building expression context from HTTP response: unexpected type when unmarshaling response: float64

Curl verbose http response from POSTing manually: 

> < HTTP/2 200
> < content-length: 1
> < content-type: text/plain; charset=utf-8

codyja:
I think https://github.com/akuity/kargo/pull/5406 may help here. I can call an API now that responds with a single string and text/plain.

PR #5406, opened 2025-11-19
feat(promotion): support JSON primitives, YAML, plain text, and content type overwrites in `http` step

codyja:
I wanted to interact with an api that responds with text/plain vs application/json. I'm attempting to add that support here.

Fixes: #5105


pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -298,18 +299,25 @@ func (h *httpRequester) buildExprEnv(
...
-		switch parsedBody.(type) {
-		case map[string]any, []any:
-			env["response"].(map[string]any)["body"] = parsedBody // nolint: forcetypeassert
-		default:
-			return nil, fmt.Errorf("unexpected type when unmarshaling response: %T", parsedBody)
+	if len(bodyBytes) > 0 {
+		// Handle text/plain responses by storing the raw text
+		if contentType == contentTypeTextPlain {
+			env["response"].(map[string]any)["body"] = string(bodyBytes) // nolint: forcetypeassert
+		} else if contentType == contentTypeJSON || json.Valid(bodyBytes) {
```
hiddeco: I think this should be switched around; otherwise, the `json.Valid` will not catch JSON that is wrongly marked as `text/plain`.

codyja: Ah good catch, I missed that. Pushed an update and let me know what you think. Thanks

hiddeco:
@codyja, thanks for following up on my comment. As this made the tests fail due to primitive types now being detected as JSON (and e.g., integer strings then being cast to floats due to JSON parsing internals), we discussed this internally, and I made some additional changes.

The end result is something you should be more than happy with, and it makes the step in general more feature-complete and flexible.

codyja:
The latest changes make a lot more sense and look great to me @hiddeco!

thomastaylor312 (approved):

PR #5425, opened 2025-11-23
docs: update CLI auth callback URL to use HTTP

Demivan:
Kargo CLI is, understandably, setting an http url as a redirect.

Using https in the OIDC configuration fails CLI login with unauthorized redirect url.


yafanasiev:
Another caveat when setting up CLI with SSO is that CLI will choose a local port at random, which does not play well with redirect URL in IDP (at least this is the case for Okta). So I wonder if docs should suggest choosing a port and running `kargo login` with `--port` parameter.

Issue #4778, opened 2025-08-06
Deprecate global credential namespaces in favor of cluster secrets


jessesuen:
# Proposed Feature

Today, we have two concepts of system-wide Kargo Secrets.

## Global Credential Namespaces

This holds git, image, and chart Secrets for a `Warehouse` and `git-clone` step to **_implicitly_** used. These secrets are not referenced by name, and are implicitly used if the repo URLs match the Warehouse config or git-clone steps.

Currently, a user is allowed to specify one or multiple namespaces as a global credential namespace. In practice, it is useful to create an empty Kargo Project, for the sole purpose of housing global credentials, just to have the convenience of managing said secrets in the Kargo UI/API/CLI.

## Cluster Secrets

Cluster secrets were introduced when we added cluster-wide webhook receivers. These are Secrets live in a **_single_** hardwired namespace (by default, it is `kargo-cluster-secrets`). Currently, the only use of cluster secrets is the cluster webhook receiver. e.g.:

```yaml
apiVersion: kargo.akuity.io/v1alpha1
kind: ClusterConfig
metadata:
  name: cluster
spec:
  webhookReceivers:
  - name: my-first-receiver
    github:
      secretRef:
        name: my-first-secret
```

Notice that it references the Secret by name, but doesn't specify any namespace. This is because we only allow a single namespace for "cluster secrets".

The existence of both Global Credential Namespaces and Cluster Secrets is now causing confusion.

I am proposing we:
1. Deprecate Global Credential Namespaces
2. Utilize Cluster Secrets as the way to configure global credentials
3. Improve the Cluster Secrets UI to allow inputting `git`, `helm`, `chart` Secret types (today it only supports a generic secret)

# Motivation

The way we do global/cluster secrets is confusing.

krancour:
The tricky thing here that I'd love to discuss when I'm back... global credentials namespace(s) and the cluster secret namespaces were conceived for distinctly different purposes, so combining them into one may cause some new problems. I'll DM you.

jessesuen:
Thanks for reminding me about the difference in intended purpose between the two. I do see a way forward to unify the concepts, while maintaining security/privacy, providing a consistent UX/UI for management, and simplifying things overall.

I still maintain we should have a single concept of "Cluster Secrets", building upon our `kargo-cluster-secrets` (currently only used by cluster webhook feature). What I propose now is simply a slight amendment of my original proposal to deprecate global credential namespaces.

[...]

Issue #4777, opened 2025-08-06
Introduce `sharedSecret()` and `sharedConfigMap()` helpers


jessesuen:
# Proposed Feature

Today, steps can reference a project level Secret like:

```yaml
- uses: http
  config:
    url: https://example.com/api/foo
    headers:
    - name: Authorization
      value: Bearer ${{ secret('my-secret').token }}
```

However, if this step is defined in a ClusterPromotionTask, then the `my-secret` has to be propagated to all projects.

Now that we have a concept of "cluster secrets" (by default is the `kargo-cluster-secrets` namespace), I feel we need to introduce a ~`clusterSecret()`~ `sharedSecret()` helper so that a ClusterPromotionTask (or even a PromotionTask) can reference it.

# Motivation

ClusterPromotionTasks cannot take advantage of Secrets unless it is propagated everywhere

jessesuen:
> Now that we have a concept of "cluster secrets"

@krancour reminded me that `kargo-cluster-secrets` was not intended to be used for shareable secrets, but rather, secrets that need to be referenced in cluster-wide configuration, namely the cluster webhooks

We wish to simplify all our concepts of secrets, which will be the effort in https://github.com/akuity/kargo/issues/4778. This issue has a dependency on what we decide there.

jessesuen:
I just remembered that we also have `configMap()` helper implemented in https://github.com/akuity/kargo/issues/3733.

I imagine we would also want kargo admins to be able to define variables DRY-ly at the cluster scope, and so I think we will also  `clusterConfigMap()` in addition to `clusterSecret()`

Alternatively, we could expand the `ClusterConfig` object to define cluster-scoped variables, but then they wouldn't be configmaps, and you would lose out on ConfigMap-like UX (e.g., referencing a configmap by name, using different configmaps for different purposes)

krancour:
`clusterSecret()` and `clusterConfigMap()` should be no problem at all after #4778 including the issue of Secret "visibility" is licked.

jessesuen:
Should this be named `sharedSecret()` and `sharedConfigMap()` so as not to confuse the secret as a secret needed by cluster-level configuration, such as cluster webhooks?

jessesuen:
This was discussed internally and to go along with `sharedSecret()` helper, we also need update various promotion steps (`jira`, `gha-`, etc...) with a `sharedSecretName` field to reference the shared secret as opposed to the project secret. e.g.:

```yaml
credentials:
[...]

PR #5428, opened 2025-11-24
feat: shared resources namespace

fuskovic:
Deprecates `GLOBAL_CREDENTIALS_NAMESPACES` in favor of new `SHARED_RESOURCES_NAMESPACE`.

Related to: https://github.com/akuity/kargo/issues/4778

Demo showing successful discovery using shared secret:

#### Secret

```
apiVersion: v1
kind: Secret
metadata:
  name: my-shared-secret
  labels:
stringData:
  repoURL: git@github.com:fuskovic/wh-test-repo.git
  sshPrivateKey: redacted
 ```
 
 #### Warehouse
 
 ```
kind: Warehouse
metadata:
  name: my-warehouse
  namespace: kargo-demo
spec:
  interval: 30m
  subscriptions:
  - git:
      repoURL: git@github.com:fuskovic/wh-test-repo.git
      branch: main
      strictSemvers: false
 ```
 
 

https://github.com/user-attachments/assets/203d0452-6628-4373-a773-d250a5cb887e

### DIFF
--- a/ui/src/features/settings/cluster-config/webhook/use-create-webhook-mutation.ts
+++ b/ui/src/features/settings/cluster-config/webhook/use-create-webhook-mutation.ts
@@ -5,9 +5,9 @@ import { parse, stringify } from 'yaml';
 
 import { queryCache } from '@ui/features/utils/cache';
 import {
-  createClusterSecret,
+  createSystemSecret,
   createOrUpdateResource,
-  deleteClusterSecret
+  deleteSystemSecret
 } from '@ui/gen/api/service/v1alpha1/service-KargoService_connectquery';
 import { ClusterConfig } from '@ui/gen/api/v1alpha1/generated_pb';
 import { PartialRecursive } from '@ui/utils/connectrpc-utils';
@@ -24,12 +24,12 @@ type createWebhookPayload = {
 
 export const useCreateWebhookMutation = (opts?: { onSuccess?: () => void }) => {
   const createOrUpdateMutation = useMutation(createOrUpdateResource);
-  const createClusterSecretMutation = useMutation(createClusterSecret);
-  const deleteClusterSecretMutation = useMutation(deleteClusterSecret);
+  const createSystemSecretMutation = useMutation(createSystemSecret);
+  const deleteSystemSecretMutation = useMutation(deleteSystemSecret);
 
   return useReactQueryMutation({
     mutationFn: async (payload: createWebhookPayload) => {
-      await createClusterSecretMutation.mutateAsync({
+      await createSystemSecretMutation.mutateAsync({
         name: payload.secret.name,
         data: payload.secret.data
       });
@@ -76,7 +76,7 @@ export const useCreateWebhookMutation = (opts?: { onSuccess?: () => void }) => {
           manifest: textEncoder.encode(stringify(clusterConfig))
         });
       } catch (e) {
-        await deleteClusterSecretMutation.mutateAsync({
+        await deleteSystemSecretMutation.mutateAsync({
           name: payload.secret.name
         });
 
--- a/ui/src/features/settings/cluster-secret/cluster-secret.tsx
+++ b/ui/src/features/settings/cluster-secret/cluster-secret.tsx
@@ -6,35 +6,35 @@ import { Button, Card, Flex, Table, Tag } from 'antd';
 import { useConfirmModal } from '@ui/features/common/confirm-modal/use-confirm-modal';
 import { useModal } from '@ui/features/common/modal/use-modal';
 import {
-  deleteClusterSecret,
+  deleteSystemSecret,
   getConfig,
-  listClusterSecrets
+  listSystemSecrets
 } from '@ui/gen/api/service/v1alpha1/service-KargoService_connectquery';
 
-import { CreateClusterSecretModal } from './create-cluster-secret-modal';
+import { CreateSystemSecretModal } from './create-system-secret-modal';
 
 export const ClusterSecret = () => {
-  const listClusterSecretsQuery = useQuery(listClusterSecrets);
+  const listSystemSecretsQuery = useQuery(listSystemSecrets);
   const confirm = useConfirmModal();
 
   const getConfigQuery = useQuery(getConfig);
   const config = getConfigQuery.data;
 
   const createSecretModal = useModal((p) => (
-    <CreateClusterSecretModal {...p} onSuccess={listClusterSecretsQuery.refetch} />
+    <CreateSystemSecretModal {...p} onSuccess={listSystemSecretsQuery.refetch} />
   ));
 
-  const deleteSecretsMutation = useMutation(deleteClusterSecret, {
-    onSuccess: () => listClusterSecretsQuery.refetch()
+  const deleteSecretsMutation = useMutation(deleteSystemSecret, {
+    onSuccess: () => listSystemSecretsQuery.refetch()
   });
 
   return (
     <Card
       title={
         <>
-          Cluster Secret{' '}
+          System Secrets{' '}
           <Tag className='text-xs ml-2' color='blue'>
-            namespace: {config?.clusterResourcesNamespace}
+            namespace: {config?.systemResourcesNamespace}
           </Tag>
         </>
       }
@@ -48,9 +48,9 @@ export const ClusterSecret = () => {
       <Table
         className='my-2'
         scroll={{ x: 'max-content' }}
-        dataSource={listClusterSecretsQuery.data?.secrets || []}
+        dataSource={listSystemSecretsQuery.data?.secrets || []}
         rowKey={(record) => record?.metadata?.name || ''}
-        loading={listClusterSecretsQuery.isLoading}
+        loading={listSystemSecretsQuery.isLoading}
         pagination={{ defaultPageSize: 10, hideOnSinglePage: true }}
         size='small'
         columns={[
@@ -88,8 +88,8 @@ export const ClusterSecret = () => {
                   size='small'
                   onClick={() =>
                     createSecretModal.show((p) => (
-                      <CreateClusterSecretModal
-                        onSuccess={listClusterSecretsQuery.refetch}
+                      <CreateSystemSecretModal
+                        onSuccess={listSystemSecretsQuery.refetch}
                         init={record}
                         {...p}
                       />
--- a/ui/src/features/settings/cluster-secret/create-system-secret-modal.tsx
+++ b/ui/src/features/settings/cluster-secret/create-system-secret-modal.tsx
@@ -9,8 +9,8 @@ import { ModalComponentProps } from '@ui/features/common/modal/modal-context';
 import { dnsRegex } from '@ui/features/common/utils';
 import { SecretEditor } from '@ui/features/project/settings/views/credentials/secret-editor';
 import {
-  createClusterSecret,
-  updateClusterSecret
+  createSystemSecret,
+  updateSystemSecret
 } from '@ui/gen/api/service/v1alpha1/service-KargoService_connectquery';
 import { Secret } from '@ui/gen/k8s.io/api/core/v1/generated_pb';
 import { zodValidators } from '@ui/utils/validators';
@@ -20,13 +20,13 @@ const createFormSchema = z.object({
   data: z.array(z.array(z.string()))
 });
 
-type CreateClusterSecretModalProps = ModalComponentProps & {
+type CreateSystemSecretModalProps = ModalComponentProps & {
   init?: Secret;
   onSuccess?(): void;
 };
 
-export const CreateClusterSecretModal = (props: CreateClusterSecretModalProps) => {
-  const createClusterSecretForm = useForm({
+export const CreateSystemSecretModal = (props: CreateSystemSecretModalProps) => {
+  const createSystemSecretForm = useForm({
     defaultValues: {
       name: props.init?.metadata?.name || '',
       data: Object.entries(props.init?.stringData || {})
@@ -36,21 +36,21 @@ export const CreateClusterSecretModal = (props: CreateClusterSecretModalProps) =
 
   const editing = !!props.init;
 
-  const createClusterSecretsMutation = useMutation(createClusterSecret, {
+  const createSystemSecretsMutation = useMutation(createSystemSecret, {
     onSuccess: () => {
       props.hide();
       props.onSuccess?.();
     }
   });
 
-  const updateClusterSecretMutation = useMutation(updateClusterSecret, {
+  const updateSystemSecretMutation = useMutation(updateSystemSecret, {
     onSuccess: () => {
       props.hide();
       props.onSuccess?.();
     }
   });
 
-  const onSubmit = createClusterSecretForm.handleSubmit((values) => {
+  const onSubmit = createSystemSecretForm.handleSubmit((values) => {
     const data: Record<string, string> = {};
 
     if (values?.data?.length > 0) {
@@ -60,13 +60,13 @@ export const CreateClusterSecretModal = (props: CreateClusterSecretModalProps) =
     }
 
     if (editing) {
-      return updateClusterSecretMutation.mutate({
+      return updateSystemSecretMutation.mutate({
         ...values,
         data
       });
     }
 
-    return createClusterSecretsMutation.mutate({
+    return createSystemSecretsMutation.mutate({
       ...values,
       data
     });
@@ -81,7 +81,7 @@ export const CreateClusterSecretModal = (props: CreateClusterSecretModalProps) =
       onOk={onSubmit}
       width='612px'
     >
-      <FieldContainer control={createClusterSecretForm.control} name='name' label='Name'>
+      <FieldContainer control={createSystemSecretForm.control} name='name' label='Name'>
         {({ field }) => (
           <Input
             value={field.value as string}
@@ -90,7 +90,7 @@ export const CreateClusterSecretModal = (props: CreateClusterSecretModalProps) =
           />
         )}
       </FieldContainer>
-      <FieldContainer control={createClusterSecretForm.control} name='data' label='Data'>
+      <FieldContainer control={createSystemSecretForm.control} name='data' label='Data'>
         {({ field }) => (
           <SecretEditor secret={field.value as [string, string][]} onChange={field.onChange} />
         )}
--- a/ui/src/pages/settings.tsx
+++ b/ui/src/pages/settings.tsx
@@ -31,10 +31,10 @@ const settingsViews = {
     component: ClusterPromotionTasks
   },
   clusterSecret: {
-    label: 'Cluster Secrets',
+    label: 'System Secrets',
     icon: faAsterisk,
     component: ClusterSecret,
-    path: 'cluster-secrets'
+    path: 'system-secrets'
   }
 };
 

### CODE REVIEW

File: pkg/credentials/kubernetes/database.go [hunk 1]
```diff
@@ -98,6 +100,22 @@ clientLoop:
...
+				repoURL,
+			); err != nil {
+				return nil, err
+			}
+			if secret != nil {
+				break clientLoop
+			}
+		}
 		// Check global credentials namespaces for credentials
 		for _, globalCredsNamespace := range k.cfg.GlobalCredentialsNamespaces {
```
hiddeco: Given the deprecation, I wonder if it wouldn't be better to do something like:


```suggestion
        // TODO: Remove this when GlobalCredentialsNamespaces is removed.
        sharedResourceNamespaces := k.cfg.GlobalCredentialsNamespace
        if k.cfg.SharedResourcesNamespace != "" {
                // Takes precedence over deprecated GlobalCredentialsNamespaces.
                sharedResourceNamespaces = []string{k.cfg.SharedResourcesNamespace}
        }
        // Check global credentials namespaces for credentials
        for _, ns := range sharedResourceNamespaces {
```

charts/kargo/templates/controller/configmap.yaml
```diff
@@ -23,7 +23,11 @@ data:
   {{- if .Values.kubeconfigSecrets.kargo }}
   KUBECONFIG: /etc/kargo/kubeconfigs/kubeconfig.yaml
   {{- end }}
-  GLOBAL_CREDENTIALS_NAMESPACES: {{ quote (join "," .Values.controller.globalCredentials.namespaces) }}
+  {{- if ( len .Values.controller.globalCredentials.namespaces ) }} 
+  SHARED_RESOURCES_NAMESPACE: {{ quote ( index .Values.controller.globalCredentials.namespaces 0 ) }}
```
hiddeco: I think this should be removed now, as otherwise only the first entry on the list will be checked due to the code changes.

krancour:
Even though what @fuskovic is doing in this PR is 100% in-line with what I'd proposed in #4778, I am having some second thoughts.

The "global credentials namespaces" supported credential lookup by type + URL rather than by direct reference. The corollary to that was that these credentials essentially became ambient. Developers working within individual projects did not actually have any way of _seeing_ the credentials. _That was deliberate_, however, taking #4777 into account, it seems we are on the verge of making these credentials _directly accessible_ by developers and I think that weakens Kargo's security posture quite a bit.

None of this is to say that we don't need to do _something_ about #4778. There is very clearly confusion between "global credentials namespaces" and "cluster secrets namespace." I do not find it confusing, personally, but I do see a litany of _other_ reasons to rethink how it all works now.

I think it would be foolish for us to charge ahead with these changes without stopping for a group sanity check, so imho, we should put a pin in this and thunderdome it after Thanksgiving.

charts/kargo/values.yaml
```diff
@@ -21,6 +21,8 @@ global:
   clusterSecretsNamespace: kargo-cluster-secrets
   ## @param global.createClusterSecretsNamespace Indicates whether the `clusterSecretsNamespace` should be managed by the chart.
   createClusterSecretsNamespace: true
+  ## @param global.sharedResourcesNamespace Indicates a namespace where shared resources can be located.
+  sharedResourcesNamespace: shared-resources-namespace
```
jessesuen: Should this be prefixed with `kargo-`? Also, the `-namespace` suffix seems unnecessary. So I guess I'm proposing: `kargo-shared-resources` as the default.

charts/kargo/templates/shared-resources-namespace/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+    name: {{ include "sharedResources.namespace" . | quote }}
```
krancour: You'll have to make this conditional like we do for the cluster resources namespace. For various reasons, some users may prefer not to have the chart create this namespace.

pkg/cli/option/flag.go
```diff
@@ -106,6 +106,8 @@ const (
 	// RoleFlag is the flag name for the role flag.
 	RoleFlag = "role"
 
+	// SharedFlag is the flag name for the shared flag.
+	SharedFlag = "shared"
```
krancour, line 110: This was just an artifact of conflict resolution I think. nbd

```suggestion
	SharedFlag = "shared"

```

hack/tilt/values.dev.yaml
```diff
@@ -33,11 +33,6 @@ api:
         - kilgore@kilgore.trout
 controller:
   logLevel: DEBUG
-  globalCredentials:
-    namespaces:
-    - kargo-shared-credentials
-  imageCache:
-    cacheByTagPolicy: Force
```
krancour, line 40: I'm not sure how these got removed. It probably had something to do with a rebase or merge at some point. At any rate, these two lines should stay.

hack/tilt/shared-creds.yaml
krancour, line 1: I believe this file can be removed completely. When we were doing global credentials _namespaces_ (plural), the chart had no responsibility for creating those namespaces or managing their permissions. It was entirely the operator (admin's) responsibility. This file was here to augment our local dev setup so that all of this was automatically accounted for when we `tilt up`. We shouldn't need this anymore.

If removing does end up breaking anything, then we missed some permissions that should have been added by the chart.

Tiltfile
```diff
@@ -139,8 +142,7 @@ k8s_resource(
     'kargo-controller-read-secrets:clusterrole',
     'kargo-controller-rollouts:clusterrole',
     'kargo-controller-rollouts:clusterrolebinding',
-    'kargo-controller-read-secrets:rolebinding',
-    'kargo-shared-credentials:namespace'
+    'kargo-controller-read-secrets:rolebinding'
```
krancour: I think this is related to the `hack/tilt/shared-creds.yaml` file that was just removed.

charts/kargo/templates/shared-resources-namespace/role-bindings.yaml
krancour, line 1: Like the shared resources namespace itself, all of this should probably be contingent on `.Values.global.createSharedResourcesNamespace` otherwise installation will potentially fail for anyone who has opted out of the chart creating that namespace for them (unless they created it in advance, but I don't want to assume they have).

charts/kargo/templates/shared-resources-namespace/roles.yaml
krancour, line 1: Same as https://github.com/akuity/kargo/pull/5428/files#r2665148639

charts/kargo/values.yaml
```diff
@@ -25,6 +25,10 @@ global:
   createClusterResourcesNamespace: true
   ## @param global.createClusterSecretsNamespace **Deprecated, as it is directly related to `clusterSecretsNamespace`, which also is deprecated. See docs for `clusterSecretsNamespace` for more details.** Indicates whether the namespace specified by `clusterSecretsNamespace` should be managed be created when installing the chart.
```
```suggestion
  ## @param global.createClusterSecretsNamespace **Deprecated, as it is directly related to `clusterSecretsNamespace`, which also is deprecated. See docs for `clusterSecretsNamespace` for more details.** Indicates whether the namespace specified by `clusterSecretsNamespace` should be managed by the chart.
```

Tiltfile
fuskovic: I think we still need this no? This role is associated with every project/namespace service account.

charts/kargo/values.yaml
```diff
@@ -23,8 +23,12 @@ global:
...
   createClusterResourcesNamespace: true
-  ## @param global.createClusterSecretsNamespace **Deprecated, as it is directly related to `clusterSecretsNamespace`, which also is deprecated. See docs for `clusterSecretsNamespace` for more details.** Indicates whether the namespace specified by `clusterSecretsNamespace` should be managed be created when installing the chart.
+  ## @param global.createClusterSecretsNamespace **Deprecated, as it is directly related to `clusterSecretsNamespace`, which also is deprecated. See docs for `clusterSecretsNamespace` for more details.** Indicates whether the namespace specified by `clusterSecretsNamespace` should be created when installing the chart.
   createClusterSecretsNamespace: true
+  ## @param global.sharedResourcesNamespace Indicates a namespace where shared resources can be located.
+  sharedResourcesNamespace: kargo-shared-resources
+  ## @param global.createSharedResourcesNamespace Indicates whether the `sharedResourcesNamespace` should be managed by the chart.
+  createSharedResourcesNamespace: true
```
krancour: I would propose re-ordering lines from +19 down to here as follows because it keeps related fields closer to one another, separates unrelated groups of fields from one another, and in the case of deprecated fields, shows a logical (_chronological_ actually) progression from old to new.

```
global:
  ## @param global.clusterSecretsNamespace **Deprecated in favor of `clusterResourcesNamespace`.** Designates a namespace to contain Secrets associated with cluster-scoped resources. For example, ClusterConfig is a cluster-scoped resource. Cluster-scoped webhook receivers, defined as part of that resource, each must reference a Secret, however, no such thing as a cluster-scoped Secret exists within Kubernetes. To work around that, all Secrets referenced by a ClusterConfig resource must be located in a designated namespace. NOTE: The namespace designated for this purpose is NOT the place to put Secrets that you want to share across all Projects. This namespace is strictly for Secrets that must be referenced by cluster-scoped resources. **Again, this is deprecated in favor of `clusterResourcesNamespace`.** The name of the setting _and_ its default value are both changing to better reflect the intent to use the namespace not only for Secrets, but for _all_ types of namespaced resources that lack a cluster-scoped analog. Kargo versions >= v1.9.0 and < v1.12.0 will automatically sync Secrets from the namespace specified by `clusterSecretsNamespace` to the namespace specified by `clusterSecretsNamespace` (if they are different). That migration utility will be removed from v1.12.0, by which time operators who GitOps their ClusterConfig will need to have updated their manifests accordingly. Upgrades to v1.12.0 will FAIL if `clusterSecretsNamespace` remains defined.
  ## @param global.createClusterSecretsNamespace **Deprecated, as it is directly related to `clusterSecretsNamespace`, which also is deprecated. See docs for `clusterSecretsNamespace` for more details.** Indicates whether the namespace specified by `clusterSecretsNamespace` should be created when installing the chart.
  createClusterSecretsNamespace: true

  ## @param global.clusterResourcesNamespace Designates a namespace to contain resources associated with cluster-scoped resources and for which no cluster-scoped analog exists. For example, ClusterConfig is a cluster-scoped resource. Cluster-scoped webhook receivers, defined as part of that resource, each must reference a Secret, however, no such thing as a cluster-scoped Secret exists within Kubernetes. To work around that, all Secrets referenced by a ClusterConfig resource must be located in a designated namespace. NOTE: The namespace designated for this purpose is NOT the place to put resources that you want to share across all Projects. This namespace is strictly for namespaced resources that must be referenced by other, cluster-scoped resources.
  clusterResourcesNamespace: kargo-cluster-resources
  ## @param global.createClusterResourcesNamespace Indicates whether the namespace specified by `clusterResourcesNamespace` should be managed be created when installing the chart.
  createClusterResourcesNamespace: true
  
  ## @param global.sharedResourcesNamespace Indicates a namespace where shared resources can be located.
  sharedResourcesNamespace: kargo-shared-resources
  ## @param global.createSharedResourcesNamespace Indicates whether the `sharedResourcesNamespace` should be managed by the chart.
  createSharedResourcesNamespace: true
  
  ## @param global.labels Labels to add to all resources.
  labels: {}
  ## @param global.annotations Annotations to add to all resources.
  annotations: {}
  ## @param global.podLabels Labels to add to all pods.
  podLabels: {}
  ## @param global.podAnnotations Annotations to add to pods.
  podAnnotations: {}
```

Tiltfile
krancour: You're right. I got my wires crossed somehow. Carry on.

Tiltfile
krancour, line 1: We need to kill lines 70 - 72 because they reference a file we deleted.

Tiltfile
krancour: Actually... no... I was right.

krancour:
```
No object identified by the fragment "kargo-controller-read-secrets:rolebinding" could be found...
```

krancour: Bindings to `kargo-controller-read-secrets` are managed dynamically by the management controller. They're not part of the chart, so Tilt doesn't see them.

charts/kargo/templates/controller/configmap.yaml
```diff
@@ -23,7 +23,7 @@ data:
   {{- if .Values.kubeconfigSecrets.kargo }}
   KUBECONFIG: /etc/kargo/kubeconfigs/kubeconfig.yaml
   {{- end }}
-  GLOBAL_CREDENTIALS_NAMESPACES: {{ quote (join "," .Values.controller.globalCredentials.namespaces) }}
+  SHARED_RESOURCES_NAMESPACE: {{ include "sharedResources.namespace" . | quote }}
```
krancour: The more I am thinking about it, it would be sensible for us to do the same here as we did for cluster secrets --> cluster resources: Keep both env vars, used the new one throughout the code, _except_ for one place -- a migration reconciler. With a few very small tweaks, the migration reconciler we already are using for cluster secrets --> cluster resources can be reused here as well. (Make source and destination namespaces injectable and instantiate two of that reconciler.)

Note that because installation explicitly fails when `.Values.controller.globalCredentials.namespaces` contains > 1 value, this is a straightforward migration from one, singular namespace to another.

We can adopt the same deprecation plan for this as we have for cluster secrets --> cluster resources. The migration reconciler will go away in v1.12.0 and install/upgrade will _fail_ if the old settings remain defined.

This would leave us in a much better place than indefinitely allowing a deprecated field to take priority over its strategic replacement.

fuskovic:
### Migration reconciler testing

I pushed the migration reconciler changes described here https://github.com/akuity/kargo/pull/5428#discussion_r2665554775 (e2e test below).

I kept the `shared-creds.yaml` and all related files and code for the test ( which have all been removed in the diff ) so if you want to run the test yourself you have to put those back.

https://github.com/user-attachments/assets/e3fd8c33-f5c0-47ce-be9d-653f30310126

secret I used for the test:

```
apiVersion: v1
kind: Secret
metadata:
  name: my-shared-secret
  namespace: kargo-shared-credentials
  labels:
stringData:
  message: "hello world"
```

charts/kargo/templates/shared-resources-namespace/role-bindings.yaml
```diff
@@ -0,0 +1,20 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: kargo-shared-resources-admin
+subjects:
+- kind: ServiceAccount
+  namespace: {{ .Release.Namespace }}
+  name: kargo-admin
+- kind: ServiceAccount
+  name: kargo-controller
+  namespace: kargo
```
fuskovic, line 19: After the last round of suggested changes I ran into the following error testing discovery:

```
failed to get git creds for git@github.com:fuskovic/wh-test-repo.git in shared namespace "kargo-shared-resources": secrets is forbidden: User "system:serviceaccount:kargo:kargo-controller" cannot list resource "secrets" in API group "" in the namespace "kargo-shared-resources"
```

So I added this subject to fix it.

charts/kargo/templates/management-controller/configmap.yaml
```diff
@@ -15,6 +15,10 @@ data:
   {{- if .Values.controller.serviceAccount.clusterWideSecretReadingEnabled }}
   MANAGE_CONTROLLER_ROLE_BINDINGS: "false"
   {{- end }}
+  {{- with .Values.controller.globalCredentials.namespaces }}
+  GLOBAL_CREDENTIALS_NAMESPACES: "{{ first . }}"
```
krancour: I think for clarity, I would drop the trailing "S".

cmd/controlplane/management_controller.go
```diff
@@ -136,13 +147,23 @@ func (o *managementControllerOptions) run(ctx context.Context) error {
...
+
+	if sharedResourcesCfg.SourceNamespace != sharedResourcesCfg.DestinationNamespace {
+		if err := secrets.SetupReconcilerWithManager(
+			ctx,
+			kargoMgr,
+			sharedResourcesCfg,
+		); err != nil {
+			return fmt.Errorf("error setting up Secrets reconciler for shared resources namespace: %w", err)
 		}
 	}
```
krancour: For both of these conditional, I think I would probably add additional criteria -- that the source namespace is not "".

cmd/controlplane/management_controller.go
```diff
@@ -182,6 +204,15 @@ func (o *managementControllerOptions) setupManager(
 			err,
 		)
 	}
+	namespaceCacheConfigs := map[string]cache.Config{
+		clusterResourcesCfg.SourceNamespace:      {},
```
krancour: Let's make this source namespace conditional.

pkg/controller/management/secrets/secrets.go
```diff
@@ -54,6 +48,7 @@ func SetupReconcilerWithManager(
 	cfg ReconcilerConfig,
 ) error {
 	err := ctrl.NewControllerManagedBy(kargoMgr).
+		Named(cfg.ControllerName).
 		For(&corev1.Secret{}).
 		WithEventFilter(
```
krancour, line 58: I know the `Reconcile()` method already checks explicitly that the resource is from the source namespace and not any other, and that's a great sanity check which we should leave, but I realized we can make this a tad more efficient like this:

```
        WithEventFilter(
            predicate.And(
                predicate.Funcs{
                    DeleteFunc: func(event.DeleteEvent) bool {
                        // We're not interested in any deletes
                        return false
                    },
                },
                // Only reconcile Secrets from the source namespace
                predicate.NewPredicateFuncs(func(obj client.Object) bool {
                    return obj.GetNamespace() == cfg.SourceNamespace
                }),
            ),
        ).
```

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -103,21 +103,23 @@ contain the following keys:
...
 Credentials are generally managed at the project level by project admins, but
 in cases where one or more sets of credentials are needed widely across many or
-all Kargo projects, an operator may opt into designating one or more namespaces
-as containing "global" credentials, accessible to all projects. It is then the
-operator's responsibility to create and manage such credentials as well.
-
-When Kargo searches for repository credentials, these additional namespaces are
-searched only _after_ finding no matching credentials in the project's own
+all Kargo projects, an operator may opt into designating a dedicated namespace
+for "global" credentials (and other resources), accessible to all projects.
```
krancour: Think we need to scrub this remaining use of "global."

```suggestion
for "global" credentials (and other resources), accessible to all projects.
```

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -103,21 +103,23 @@ contain the following keys:
 
 Credentials are generally managed at the project level by project admins, but
 in cases where one or more sets of credentials are needed widely across many or
-all Kargo projects, an operator may opt into designating one or more namespaces
-as containing "global" credentials, accessible to all projects. It is then the
-operator's responsibility to create and manage such credentials as well.
-
-When Kargo searches for repository credentials, these additional namespaces are
-searched only _after_ finding no matching credentials in the project's own
+all Kargo projects, an operator may opt into designating a dedicated namespace
```
krancour: It's not really opt-in anymore.

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -103,21 +103,23 @@ contain the following keys:
...
+for "global" credentials (and other resources), accessible to all projects.
+This namespace is referred to as the `shared-resources-namespace`.
```
krancour: Suggest not monospacing this, but adding of emphasis.

```suggestion
This namespace is referred to as the _**shared resources namespace**_. 
```

cmd/controlplane/management_controller.go
fuskovic: Is this necessary for `CLUSTER_RESOURCES_NAMESPACE` ? We default to `kargo-cluster-resources` if it's empty.

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -103,21 +103,23 @@ contain the following keys:
...
-
-When Kargo searches for repository credentials, these additional namespaces are
-searched only _after_ finding no matching credentials in the project's own
+all Kargo projects, an operator may opt into designating a dedicated namespace
+for "global" credentials (and other resources), accessible to all projects.
+This namespace is referred to as the `shared-resources-namespace`. 
+It is then the operator's responsibility to create and manage such credentials 
+as well.
+
+When Kargo searches for repository credentials, the `shared-resources-namespace`
```
krancour:
```suggestion
When Kargo searches for repository credentials, the shared resources namespace
```

cmd/controlplane/management_controller.go
fuskovic: [This](https://github.com/akuity/kargo/pull/5428#discussion_r2673597080) applies here too

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -103,21 +103,23 @@ contain the following keys:
...
 
 
 :::info[Precedence]
 
-When Kargo searches for repository credentials in a "global" namespace, it
-_first_ iterates over all appropriately labeled `Secret`s _without_
-`repoURLIsRegex` set to `true` looking for a `repoURL` value matching the
-repository URL exactly.
+When Kargo searches for repository credentials in the 
+`shared-resources-namespace`, it_first_ iterates over all appropriately labeled
```
krancour:
```suggestion
shared resources namespace, it_first_ iterates over all appropriately labeled 
```

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -138,14 +140,13 @@ credentials, except possibly that they exist.
 
 ### Enabling Global Credentials
 
-To designate one or more namespaces as containing "global" credentials, list
-them under the Kargo Helm chart's `controller.globalCredentials.namespaces`
-option at installation time.
+A `shared-resources-namespace` can be designated under the Kargo Helm chart's 
+`global.sharedResourcesNamespace` option at installation time.
```
krancour: Not opt-in anymore. But you can say here what the default name for that namespace is.

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -138,14 +140,13 @@ credentials, except possibly that they exist.
 
 ### Enabling Global Credentials
```
krancour:
```suggestion
### Enabling Shared Credentials
```

docs/docs/40-operator-guide/40-security/40-managing-credentials.md
```diff
@@ -138,14 +140,13 @@ credentials, except possibly that they exist.
...
+A `shared-resources-namespace` can be designated under the Kargo Helm chart's 
+`global.sharedResourcesNamespace` option at installation time.
 
 Operators must also manually ensure Kargo controllers receive read-only access
-to `Secret`s in the designated namespaces. For example, if `kargo-global-creds`
-is designated as a global credentials namespace, the following `RoleBinding`
-should be created within that namespace:
+to `Secret`s in the designated `shared-resources-namespace`. For example, 
+if `kargo-global-creds` is designated as the `shared-resources-namespace`, 
+the following `RoleBinding` should be created within that namespace:
```
krancour: This was true for global credentials namespaces, but is not true for the (now singular) shared resources namespace.

(If you care about the technical reason, it had to do with the difficult of having the management controller dynamically expand and contract admin access to a _variable_ set of namespaces managed neither by the management controller itself nor the Kargo chart.)

This can just be removed.

cmd/controlplane/management_controller.go
fuskovic: Nvm, I conflated with the destination namespace.

cmd/controlplane/management_controller.go
fuskovic: Nvm: https://github.com/akuity/kargo/pull/5428#discussion_r2673669202

charts/kargo/templates/management-controller/configmap.yaml
```diff
@@ -15,6 +15,10 @@ data:
   {{- if .Values.controller.serviceAccount.clusterWideSecretReadingEnabled }}
   MANAGE_CONTROLLER_ROLE_BINDINGS: "false"
   {{- end }}
+  {{- with .Values.controller.globalCredentials.namespaces }}
+  GLOBAL_CREDENTIALS_NAMESPACE: "{{ first . }}"
```
krancour, line 24: Have we, somewhere along the line, lost the explicit chart-rendering failure if multiple global creds namespaces had been defined?

cmd/controlplane/management_controller.go
```diff
@@ -182,6 +206,17 @@ func (o *managementControllerOptions) setupManager(
 			err,
 		)
 	}
+	namespaceCacheConfigs := map[string]cache.Config{
+		clusterResourcesCfg.DestinationNamespace: {},
+		sharedResourcesCfg.DestinationNamespace:  {},
```
krancour: Realizing now that no access to these is required if no migrations are happening. i.e. these should probably be guarded by the newly improved conditionals you have below.

PR #5433, opened 2025-11-28
feat: cluster resources namespace

fuskovic:
Deprecates `CLUSTER_SECRETS_NAMESPACE` in favor of new `CLUSTER_RESOURCES_NAMESPACE`.

Related to: https://github.com/akuity/kargo/issues/4778

Migration reconciler manual testing:

https://github.com/user-attachments/assets/548dbf90-e9d1-4798-98d2-60461f13fd98


Plork:
When doing this is it maybe possible to add annotations to the namespace? In case of permitted regex for for example external-secrets.

```yaml
          external-secrets.io/allowed-secret-store: ""
          external-secrets.io/permitted-key-regex: ""
```

When creating the namespace it would help if we can provide these.

krancour:
@Plork that seems like a separate feature request.

charts/kargo/templates/api/configmap.yaml
```diff
@@ -22,7 +22,11 @@ data:
   {{- if .Values.api.secretManagementEnabled }}
   SECRET_MANAGEMENT_ENABLED: "true"
   {{- end }}
-  CLUSTER_SECRETS_NAMESPACE: {{ .Values.global.clusterSecretsNamespace }}
+  {{- if .Values.global.clusterResourcesNamespace }}
+  CLUSTER_RESOURCES_NAMESPACE: {{ quote .Values.global.clusterResourcesNamespace }}
+  {{- else }}
+  CLUSTER_RESOURCES_NAMESPACE: {{ quote .Values.global.clusterSecretsNamespace }}
+  {{- end }}
```
krancour: This needs to be the other way around. To not break existing installations, the older setting key needs to be prioritized. A comment to this effect would help here as well. I cannot remember the syntax for a comment in a Go template, but it exists.

charts/kargo/templates/cluster-resources-namespace/namespace.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if (or .Values.global.createClusterSecretsNamespace .Values.global.createClusterResourcesNamespace) }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  {{- if .Values.global.clusterResourcesNamespace }}
+  name: {{ quote .Values.global.clusterResourcesNamespace }}
+  {{- else }}
+  name: {{ quote .Values.global.clusterSecretsNamespace }}
+  {{- end }}
```
krancour: Same as [here](https://github.com/akuity/kargo/pull/5433/files#r2627511423)

It might be worth making a helper that resolves to the correct value.

charts/kargo/templates/controller/configmap.yaml
```diff
@@ -13,7 +13,11 @@ data:
   {{- end }}
   LOG_LEVEL: {{ quote .Values.controller.logLevel }}
   LOG_FORMAT: {{ quote .Values.controller.logFormat }}
-  CLUSTER_SECRETS_NAMESPACE: {{ .Values.global.clusterSecretsNamespace }}
+  {{- if .Values.global.clusterResourcesNamespace }}
+  CLUSTER_RESOURCES_NAMESPACE: {{ quote .Values.global.clusterResourcesNamespace }}
+  {{- else }}
+  CLUSTER_RESOURCES_NAMESPACE: {{ quote .Values.global.clusterSecretsNamespace }}
+  {{- end }}
```
krancour: Ok... third spot where this needs to change. I said maybe a helper was in order. Now I think it's _definitely_ in order. I won't comment on further occurrences of this logic. Assume they all need to be fixed and would benefit from a helper.

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
 
 ## @section Global Parameters
 global:
-  ## @param global.clusterSecretsNamespace Indicates a namespace where Secrets associated with cluster-scoped resources can be located.
+  ## @param global.clusterResourcesNamespace Indicates a namespace where cluster-scoped resources can be located.
```
krancour: Wording is wrong here. Cluster-scoped resources don't live in a namespace. This namespace if for resources _associated_ with cluster-scoped resources and for which no cluster-scoped analog exists. Secrets are obviously the canonical example. ClusterConfig, which is cluster-scoped, references Secrets. In a perfect world, something like ClusterSecret would exist, so this namespace is a workaround for that.

charts/kargo/templates/api/configmap.yaml
hiddeco: `{{/* a comment */}}`

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
...
+  clusterResourcesNamespace: kargo-resources-namespace
+  ## @param global.clusterSecretsNamespace DEPRECATED( Use clusterResourcesNamespace instead ) Indicates a namespace where Secrets associated with cluster-scoped resources can be located.
   clusterSecretsNamespace: kargo-cluster-secrets
```
krancour, line 23: As mentioned [here](https://github.com/akuity/kargo/pull/5433/files#r2627511423), I believe this needs to be prioritized in order to not break existing installations. That means for `clusterResourcesNamespace` to _ever_ take effect, this field needs to go away...

But then that presents a problem also as it turns out... if it goes away, existing installations that _didn't_ override the default value will break because they'd have shared resources in the original namespace, but the chart would start using the one specified by `clusterResourcesNamespace`...

This is quite a tricky thing to get correct without breaking anyone... I'm starting to wonder if it's even possible.

cc @hiddeco for any ideas.

krancour:
All changes to Go code LGTM. We just need to focus on getting the chart changes right.

krancour:
> All changes to Go code LGTM. We just need to focus on getting the chart changes right.

No longer the case after this afternoon's offline discussions.

@fuskovic I amended this PR with a commit that adds a temporary Secret migration reconciler to the management controller. I propose removing it in v1.12.0, which is one release longer than we typically go between deprecating something and removing it entirely.

There's still a lot of work to be done here, though:

1. Adjust Go code to (apart from the migration reconciler) exclusively work with the "cluster _resources_ namespace." Only the migration reconciler needs to know or care about the old "cluster secrets namespace."

1. Adjust the chart so that rather than prioritize one of `clusterSecretsNamespace` or `clusterResourcesNamespace` over the other, we simply support _both_ (for now), with the values of `clusterSecretsNamespace` ultimately making its way into the `CLUSTER_SECRETS_NAMESPACE` env var and the value of `clusterResourcesNamespace` ultimately making its way into the `CLUSTER_RESOURCES_NAMESPACE` namespace. (Guess we don't need that new helper after all.)

1. Ensure `clusterSecretsNamespace` is clearly marked as being deprecated, with a further note that Kargo will automatically sync Secrets from one namespace to the other until v1.12.0, at which time, the migration reconciler _and_ support for the `clusterSecretsNamespace` will both be removed. At that point, users will have had a substantial amount of time to adapt. Also deprecate `createClusterSecretsNamespace`.
  
1. Make the chart conditionally (based on the value of `createClusterResourcesNamespace`) create the the namespace specified by `clusterResourcesNamespace`.

1. Find docs related to "cluster secrets namespace" and update it to outline that we're simply generalizing from this being limited to Secrets to being a suitable home for _any_ namespaced resource that must be referenced by a cluster-scoped resource. It will need to be just as clear as the comments in `values.yaml` as to what's deprecated, when it's being removed, and what the migration reconciler will do to ease the transition in the meantime.

1. We will need to prepare a similar statement to be featured prominently in the release notes.

1. We can talk with @rpelczar about the possibility of surfacing this important information in a dismissable banner in the UI for managing cluster resources (which might not even exist yet) as well.

1. Create issues to remind ourselves of what needs to be removed in v1.12.0 and add them to that milestone.

Last, I think there are potentially some weird edge cases that I still need to think through more carefully for the temporary migration reconciler... I just didn't want to leave anyone with the impression that it's 100% g2g. I'll worry about this piece of the puzzle.

charts/kargo/values.yaml
fuskovic: Marking this as resolved since we're going to be keeping both for now per your latest comment [here](https://github.com/akuity/kargo/pull/5433#issuecomment-3667742052).

cmd/controlplane/management_controller.go
```diff
@@ -183,6 +192,11 @@ func (o *managementControllerOptions) setupManager(ctx context.Context) (manager
 							o.KargoNamespace: {},
 						},
 					},
+					&corev1.Secret{}: {
+						Namespaces: map[string]cache.Config{
+							os.GetEnv("CLUSTER_RESOURCES_NAMESPACE", "kargo-cluster-resources"): {},
```
krancour:
```suggestion
							os.GetEnv("CLUSTER_SECRETS_NAMESPACE", "kargo-cluster-secrets"): {},
```

krancour: This was 100% my screw up when I transferred work from another branch to this one.

krancour:
Adding a note that my original migration reconciler neglected to add finalizers to original Secrets.

fuskovic:
> Adding a note that my original migration reconciler neglected to add finalizers to original Secrets.

@krancour 

Added finalizers and adjusted the behavior to reflect what we discussed offline. 

Had to delete a couple of the original tests you added since the new behavior makes them obsolete.

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
 
 ## @section Global Parameters
 global:
-  ## @param global.clusterSecretsNamespace Indicates a namespace where Secrets associated with cluster-scoped resources can be located.
+  ## @param global.clusterResourcesNamespace is for resources associated with cluster-scoped resources and for which no cluster-scoped analog exists.
+  clusterResourcesNamespace: kargo-cluster-resources
+  ## @param global.clusterSecretsNamespace DEPRECATED( Use clusterResourcesNamespace instead ) Indicates a namespace where Secrets associated with cluster-scoped resources can be located. Kargo will automatically sync Secrets from one namespace to the other until v1.12.0, at which time, the migration reconciler and support for the clusterSecretsNamespace will both be removed.
```
hiddeco: Nit. Styling of the deprecation notice is strange.

```suggestion
  ## @param global.clusterSecretsNamespace **Deprecated in favor of `clusterResourcesNamespace`.** Indicates a namespace where Secrets associated with cluster-scoped resources can be located. Kargo will automatically sync Secrets from one namespace to the other until v1.12.0, at which time, the migration reconciler and support for the clusterSecretsNamespace will both be removed.
```

or something Markdownsy might be better.

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
...
   clusterSecretsNamespace: kargo-cluster-secrets
-  ## @param global.createClusterSecretsNamespace Indicates whether the `clusterSecretsNamespace` should be managed by the chart.
+  ## @param global.createClusterResourcesNamespace Indicates whether the `clusterResourcesNamespace` should be managed by the chart.
+  createClusterResourcesNamespace: true
+  ## @param global.createClusterSecretsNamespace DEPRECATED( Use createClusterResourcesNamespace ) Indicates whether the `clusterSecretsNamespace` should be managed by the chart.
```
hiddeco: Likewise comment as above.


```suggestion
  ## @param global.createClusterSecretsNamespace **Deprecated in favor of `createClusterResourcesNamespace`.** Indicates whether the `clusterSecretsNamespace` should be managed by the chart.
```

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
 
 ## @section Global Parameters
 global:
-  ## @param global.clusterSecretsNamespace Indicates a namespace where Secrets associated with cluster-scoped resources can be located.
+  ## @param global.clusterResourcesNamespace is for resources associated with cluster-scoped resources and for which no cluster-scoped analog exists.
```
hiddeco:
```suggestion
  ## @param global.clusterResourcesNamespace Namespace for resources (like Secrets) associated with cluster-scoped resources and for which no cluster-scoped analog exists.
```

might be more indicative of what type of resources this is used for.

pkg/controller/management/secrets/secrets.go
```diff
@@ -0,0 +1,204 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf(
+			"error getting destination Secret %q in namespace %q: %w",
+			srcSecret.Name, r.cfg.DestinationNamespace, err,
+		)
+	}
+
+	if err = r.client.Delete(ctx, destSecret); err != nil {
+		return fmt.Errorf(
```
hiddeco, line 206: Missing a `client.IgnoreNotFound(err)` type of check.

cmd/controlplane/management_controller.go
```diff
@@ -134,6 +135,14 @@ func (o *managementControllerOptions) run(ctx context.Context) error {
 		}
 	}
 
+	if err := secrets.SetupReconcilerWithManager(
```
hiddeco: Might be worth checking if `secretsCfg.SourceNamespace != secretsCfg.DestinationNamespace` to prevent doing unnecessary work?

cmd/controlplane/management_controller.go
```diff
@@ -183,6 +192,12 @@ func (o *managementControllerOptions) setupManager(ctx context.Context) (manager
 							o.KargoNamespace: {},
 						},
 					},
+					&corev1.Secret{}: {
+						Namespaces: map[string]cache.Config{
+							os.GetEnv("CLUSTER_SECRETS_NAMESPACE", "kargo-cluster-secrets"):     {},
+							os.GetEnv("CLUSTER_RESOURCES_NAMESPACE", "kargo-cluster-resources"): {},
```
hiddeco: Is there a specific reason we are redefining this here with its own set of defaults versus obtaining this from (one of the) config(s)?

pkg/controller/management/secrets/secrets_test.go
```diff
@@ -0,0 +1,277 @@
...
+				WithObjects(withFinalizer(testSrcSecret)).Build(),
+			assertions: func(t *testing.T, c client.Client, err error) {
+				require.NoError(t, err)
+				dest := &corev1.Secret{}
+				err = c.Get(t.Context(), types.NamespacedName{
+					Namespace: testDestNamespace,
+					Name:      testSecretName,
+				}, dest)
+				require.NoError(t, err)
+				require.Equal(t, testDestSecret2.Data, testSrcSecret.Data)
```
hiddeco:
```suggestion
				require.Equal(t, testDestSecret2.Data, dest.Data)
```

cmd/controlplane/management_controller.go
```diff
@@ -134,6 +135,17 @@ func (o *managementControllerOptions) run(ctx context.Context) error {
 		}
 	}
 
+	secretsCfg := secrets.ReconcilerConfigFromEnv()
```
hiddeco: If we would move this to `L91`.

cmd/controlplane/management_controller.go
hiddeco: Then we can do this here:


```suggestion
func (o *managementControllerOptions) setupManager(ctx context.Context, secretsCfg secrets.ReconcilerConfig) (manager.Manager, error) {
```

cmd/controlplane/management_controller.go
```diff
@@ -167,7 +179,7 @@ func (o *managementControllerOptions) setupManager(ctx context.Context) (manager
 			err,
 		)
 	}
-
+	secretsCfg := secrets.ReconcilerConfigFromEnv()
```
hiddeco: And remove this:

```suggestion
```

hack/tilt/values.dev.yaml
```diff
@@ -1,5 +1,8 @@
 global:
+  clusterResourcesNamespace: kargo-cluster-resources
   clusterSecretsNamespace: kargo-cluster-secrets
+  createClusterResourcesNamespace: true
+  createClusterSecretsNamespace: true
```
krancour: Isn't true the default?

pkg/controller/management/projects/projects.go
```diff
@@ -45,7 +45,7 @@ type ReconcilerConfig struct {
 	ManageControllerRoleBindings bool   `envconfig:"MANAGE_CONTROLLER_ROLE_BINDINGS" default:"true"`
 	KargoNamespace               string `envconfig:"KARGO_NAMESPACE" default:"kargo"`
 	MaxConcurrentReconciles      int    `envconfig:"MAX_CONCURRENT_PROJECT_RECONCILES" default:"4"`
-	ClusterSecretsNamespace      string `envconfig:"CLUSTER_SECRETS_NAMESPACE"`
+	ClusterResourcesNamespace    string `envconfig:"CLUSTER_RESOURCES_NAMESPACE"`
```
krancour: This seems not to be used and can go away entirely.

charts/kargo/templates/api/configmap.yaml
```diff
@@ -22,7 +22,8 @@ data:
   {{- if .Values.api.secretManagementEnabled }}
   SECRET_MANAGEMENT_ENABLED: "true"
   {{- end }}
-  CLUSTER_SECRETS_NAMESPACE: {{ .Values.global.clusterSecretsNamespace }}
+  CLUSTER_SECRETS_NAMESPACE: {{ quote .Values.global.clusterSecretsNamespace }}
+  CLUSTER_RESOURCES_NAMESPACE: {{ quote .Values.global.clusterResourcesNamespace }}
```
krancour: Isn't the management controller the only component that needs to know about both of these (so it can do the migration)? I believe all other components just work the new way with the assumption that the management controller has taken care of the migration.

Same comment applies to a few other components as well.

pkg/controller/management/secrets/secrets.go
```diff
@@ -0,0 +1,204 @@
...
+// SetupReconcilerWithManager initializes a reconciler for Secret resources
+// and registers it with the provided Manager.
+func SetupReconcilerWithManager(
+	ctx context.Context,
+	kargoMgr manager.Manager,
+	cfg ReconcilerConfig,
+) error {
+	err := ctrl.NewControllerManagedBy(kargoMgr).
+		For(&corev1.Secret{}).
+		Complete(newReconciler(kargoMgr.GetClient(), cfg))
```
krancour, line 66: You can add this, because we're not at all interested in actual deletes. ("Delete," as far as we are concerned is an update with `DeletedTimestamp` being non-nil.)

```suggestion
		For(&corev1.Secret{}).
		WithEventFilter(
			predicate.Funcs{
				DeleteFunc: func(event.DeleteEvent) bool {
					// We're not interested in any deletes
					return false
				},
			},
		).
		Complete(newReconciler(kargoMgr.GetClient(), cfg))
```

pkg/controller/management/secrets/secrets.go
```diff
@@ -0,0 +1,204 @@
...
+	srcSecret := &corev1.Secret{}
+	if err := r.client.Get(
+		ctx,
+		types.NamespacedName{
+			Namespace: r.cfg.SourceNamespace,
+			Name:      req.Name,
+		},
+		srcSecret,
+	); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
krancour, line 120:
```suggestion
    // Ignore if not found. This can happen if the Secret was deleted after the
		// current reconciliation request was issued.
		return ctrl.Result{}, client.IgnoreNotFound(err)
```

pkg/controller/management/secrets/secrets.go
```diff
@@ -0,0 +1,204 @@
...
+	logger.Debug("done reconciling Secret")
+
+	return ctrl.Result{}, nil
+}
+
+// handleDelete handles the deletion of a source Secret by deleting the
+// corresponding destination Secret and removing the finalizer from the source
+// Secret to unblock its deletion.
+func (r *reconciler) handleDelete(ctx context.Context, srcSecret *corev1.Secret) error {
+	logger := logging.LoggerFromContext(ctx)
```
krancour, line 187:
```suggestion
	// If the Secret does not have the finalizer, there is nothing to do.
	if !controllerutil.ContainsFinalizer(srcSecret, kargoapi.FinalizerName) {
		return nil
	}

	logger := logging.LoggerFromContext(ctx)
```

pkg/controller/management/secrets/secrets_test.go
```diff
@@ -0,0 +1,277 @@
...
+							...client.GetOption,
+						) error {
+							return fmt.Errorf("something went wrong")
+						},
+					},
+				).Build(),
+			assertions: func(t *testing.T, _ client.Client, err error) {
+				require.ErrorContains(t, err, "something went wrong")
+			},
+		},
```
krancour, line 111: @fuskovic, this was my oversight when I wrote the first draft of the migration reconciler, but we should probably have a separate test case for when the error getting the source is a not found error, because we're explicitly ignoring such an error. Mind fixing my oversight?

pkg/controller/management/secrets/secrets_test.go
```diff
@@ -0,0 +1,277 @@
...
+						Name:      testSecretName,
+						DeletionTimestamp: &metav1.Time{
+							Time: metav1.Now().Time,
+						},
+					},
+				})).Build(),
+			assertions: func(t *testing.T, _ client.Client, err error) {
+				require.ErrorContains(t, err, "error getting destination Secret")
+			},
+		},
```
krancour, line 233: Shouldn't we be ignoring it when the destination doesn't exist? It means the cleanup we were out to perform is (for whatever reason) not necessary. I think we can claim that as a success instead of returning an error.

pkg/controller/management/secrets/secrets_test.go
```diff
@@ -0,0 +1,277 @@
...
+					},
+				).Build(),
+			assertions: func(t *testing.T, _ client.Client, err error) {
+				require.ErrorContains(t, err, "something went wrong")
+			},
+		},
+		{
+			name: "source exists, destination does not; success creating destination",
+			client: fake.NewClientBuilder().WithScheme(testScheme).
+				WithObjects(withFinalizer(testSrcSecret)).Build(),
```
krancour, line 176: Because the scenario for this test case seems to be an initial copy, I think we'd do well to not start with the finalizer already on the source Secret _and_ then assert that it was added.

krancour:
In terms of docs, I believe there's only one relevant place that needs an update as part of this PR:

https://github.com/akuity/kargo/blob/ef1bd7ed4b50198f182b0990fec3643e972fb2e2/docs/docs/40-operator-guide/35-cluster-configuration.md?plain=1#L78-L88

I would recommend rewording that note box accordingly and adding a second box (maybe a warning box to attract more attention to it) and note that this has changed from previous releases and very briefly explain the automated migration and note when that migration and support for the old namespace will be removed).

Along similar lines... please write a short blurb about the change from "cluster secrets" to "cluster resources" -- why we're doing it and why and fire that off to @fykaa. She will incorporate it into release notes.

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
 
 ## @section Global Parameters
 global:
-  ## @param global.clusterSecretsNamespace Indicates a namespace where Secrets associated with cluster-scoped resources can be located.
+  ## @param global.clusterResourcesNamespace Namespace for resources (like Secrets) associated with cluster-scoped resources and for which no cluster-scoped analog exists.
+  clusterResourcesNamespace: kargo-cluster-resources
+  ## @param global.clusterSecretsNamespace **Deprecated in favor of `clusterResourcesNamespace`.** Indicates a namespace where Secrets associated with cluster-scoped resources can be located. Kargo will automatically sync Secrets from one namespace to the other until v1.12.0, at which time, the migration reconciler and support for the clusterSecretsNamespace will both be removed.
   clusterSecretsNamespace: kargo-cluster-secrets
-  ## @param global.createClusterSecretsNamespace Indicates whether the `clusterSecretsNamespace` should be managed by the chart.
+  ## @param global.createClusterResourcesNamespace Indicates whether the `clusterResourcesNamespace` should be managed by the chart.
```
```suggestion
  ## @param global.createClusterResourcesNamespace Indicates whether the namespace specified by `clusterResourcesNamespace` should be managed be created when installing the chart.
```

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
...
+  createClusterResourcesNamespace: true
+  ## @param global.createClusterSecretsNamespace **Deprecated in favor of `createClusterResourcesNamespace`.** Indicates whether the `clusterSecretsNamespace` should be managed by the chart.
```
krancour: Nit: It's not technically true that this is deprecated in favor of `createClusterResourcesNamespace`. `clusterSecretsNamespace` _is_ deprecated in favor of `clusterResourcesNamespace`, but this field is deprecated because there's no need for it once `clusterSecretsNamespace` is gone.

```suggestion
  ## @param global.createClusterSecretsNamespace **Deprecated, as it is directly related to `clusterSecretsNamespace`, which also is deprecated. See docs for `clusterSecretsNamespace` for more details.** Indicates whether the namespace specified by `clusterSecretsNamespace` should be managed be created when installing the chart.
```

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
 
 ## @section Global Parameters
 global:
-  ## @param global.clusterSecretsNamespace Indicates a namespace where Secrets associated with cluster-scoped resources can be located.
+  ## @param global.clusterResourcesNamespace Namespace for resources (like Secrets) associated with cluster-scoped resources and for which no cluster-scoped analog exists.
```
krancour: In the interest of clearing up confusion that we know existed with the old name, I want to suggest being much more detailed here. I want to leave no room for misunderstanding what this namespace is for.

```suggestion
  ## @param global.clusterResourcesNamespace Designates a namespace to contain resources associated with cluster-scoped resources and for which no cluster-scoped analog exists. For example, ClusterConfig is a cluster-scoped resource. Cluster-scoped webhook receivers, defined as part of that resource, each must reference a Secret, however, no such thing as a cluster-scoped Secret exists within Kubernetes. To work around that, all Secrets referenced by a ClusterConfig resource must be located in a designated namespace. NOTE: The namespace designated for this purpose is NOT the place to put resources that you want to share across all Projects. This namespace is strictly for namespaced resources that must be referenced by other, cluster-scoped resources.
```

charts/kargo/values.yaml
```diff
@@ -17,9 +17,13 @@ image:
...
+  clusterResourcesNamespace: kargo-cluster-resources
+  ## @param global.clusterSecretsNamespace **Deprecated in favor of `clusterResourcesNamespace`.** Indicates a namespace where Secrets associated with cluster-scoped resources can be located. Kargo will automatically sync Secrets from one namespace to the other until v1.12.0, at which time, the migration reconciler and support for the clusterSecretsNamespace will both be removed.
```
```suggestion
  ## @param global.clusterSecretsNamespace **Deprecated in favor of `clusterResourcesNamespace`.** Designates a namespace to contain Secrets associated with cluster-scoped resources. For example, ClusterConfig is a cluster-scoped resource. Cluster-scoped webhook receivers, defined as part of that resource, each must reference a Secret, however, no such thing as a cluster-scoped Secret exists within Kubernetes. To work around that, all Secrets referenced by a ClusterConfig resource must be located in a designated namespace. NOTE: The namespace designated for this purpose is NOT the place to put Secrets that you want to share across all Projects. This namespace is strictly for Secrets that must be referenced by cluster-scoped resources. **Again, this is deprecated in favor of `clusterResourcesNamespace`.** The name of the setting _and_ its default value are both changing to better reflect the intent to use the namespace not only for Secrets, but for _all_ types of namespaced resources that lack a cluster-scoped analog. Kargo versions >= v1.9.0 and < v1.12.0 will automatically sync Secrets from the namespace specified by `clusterSecretsNamespace` to the namespace specified by `clusterSecretsNamespace` (if they are different). That migration utility will be removed from v1.12.0, by which time operators who GitOps their ClusterConfig will need to have updated their manifests accordingly. Upgrades to v1.12.0 will FAIL if `clusterSecretsNamespace` remains defined.
```

pkg/controller/management/secrets/secrets.go
```diff
@@ -0,0 +1,221 @@
...
+	if ok, err := api.EnsureFinalizer(ctx, r.client, srcSecret); ok || err != nil {
+		logger.Debug("ensured finalizer on source Secret; requeuing")
+		return ctrl.Result{RequeueAfter: 100 * time.Millisecond}, err
+	}
+
+	destSecret := srcSecret.DeepCopy()
+
+	// Remove the finalizer that was just copied over so that deletes of new
+	// secrets won't be blocked by anything.
+	controllerutil.RemoveFinalizer(destSecret, kargoapi.FinalizerName)
```
krancour, line 140: `controllerutil.RemoveFinalizer()` returns an `error` that you're not capturing. If you were, I suspect you'd find this call is actually returning a non-nil error.

`controllerutil.RemoveFinalizer()` seems to go as far as patching the resource. If this is a first-time sync of the source resource, this would fail with a not-found error.

Rather than starting to capture the error, I think we don't need to use `controllerutil.RemoveFinalizer()` and can get away with simply removing our finalizer if it's found in the list of finalizers.

docs/docs/40-operator-guide/35-cluster-configuration.md
```diff
@@ -80,13 +80,26 @@ receiver vary, and are documented on
...
+:::warning
+
+Prior to `v1.9`, _cluster-scoped_ resources lived in `kargo-cluster-secrets`(
+  or the namespace designated by`global.clusterSecretsNamespace` if the default
+  was overwritten
+). Now, _cluster-scoped_ resources live in `kargo-cluster-resources` by default.
+In addition to this change, a migration reconciler has been introduced to move 
+any existing `Secret` resources from `kargo-cluster-secrets` to 
+`kargo-cluster-resources`. `kargo-cluster-secrets` will be deprecated entirely
+in a future release.
```
krancour:
```suggestion
Prior to `v1.9`, `Secret`s referenced by `ClusterConfig` were contained in the namespace designated by`global.clusterSecretsNamespace`, with a default of `kargo-cluster-secrets`.

Kargo v1.9.0 changed the name of the setting to `global.clusterResourcesNamespace` and its default value to `kargo-cluster-resources` to better reflect the intent to use the namespace not only for `Secret`s, but for _all_ types of namespaced resources that lack a cluster-scoped analog.

Kargo versions >= v1.9.0 and < v1.12.0 will automatically sync `Secret`s from the namespace specified by `global.clusterSecretsNamespace` to the namespace specified by `global.clusterSecretsNamespace` (if they are different). That migration utility will be removed from v1.12.0, by which time operators who GitOps their `ClusterConfig` will need to have updated their manifests accordingly. Upgrades to v1.12.0 will **fail** if `global.clusterSecretsNamespace` remains defined.
```

pkg/controller/management/secrets/secrets.go
fuskovic: `controllerutil.RemoveFinalizer` doesn't return an error or make a patch. I think you're conflating with `api.RemoveFinalizer`?

krancour: 🤦‍♂️ I sure did confuse those two. Thank you. Nothing to see here.

krancour:
This is working beautifully and I believe at this point we've left no stone unturned.

🔥 Great work!

Issue #5434, opened 2025-11-28
Promotion step `kustomize-build` does not adhere to `kustomize`'s filename standard


lindhe:
Files generated by the `kustomize-build` promotion step are named differently than files generated by `kustomize` in CLI. I expected `kustomize-build` to behave pretty much like the CLI tool does, so I was surprised by this discrepancy. It becomes an issue for me because it prevents me from rendering manifests manually, should I need to.

Today, Kargo's `kustomize-build` seems to use the pattern `[<namespace>-]<kind>-<name>.yaml`:

https://github.com/akuity/kargo/blob/f1e86a3c07debaa5e4736fea5d013a598763810c/pkg/promotion/runner/builtin/kustomize_builder.go#L131-L134

Upstream `kustomize build --output ./subdir/ .` seems to be using the pattern `[<namespace>_]<api_group>_<api_version>_<kind>_<name>.yaml`:

https://github.com/kubernetes-sigs/kustomize/blob/17a06a72be7fa8e3fd50b5536c2fb32f8a4126cf/kustomize/commands/build/writer.go

I'd really like to see Kargo's `kustomize-build` adhere to the same filename patterns as upstream Kustomize does.

If you think this sounds like a good idea but lack the bandwidth for it right now, I'd be happy to _attempt_ a PR to fix this.

Would you consider it a breaking change?

PR #5442, opened 2025-12-01
feat(promotion): allow output format configuration `kustomize-build`

hiddeco:
Fixes: #5434

By adding the option to define an `outputFormat` as part of the `kustomize-build` step configuration. When set to `kustomize`, this will produce the same output as `kustomize-build -o dir/` (i.e., `[namespace_]group_version_kind_name.yaml`).

The default continues to be the "Kargo invented" format of `[namespace-]kind-name`.

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/kustomize-build.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/kustomize-build.md
@@ -18,6 +18,7 @@ preceded by a [`git-clear`](git-clear.md) step and followed by
 |------|------|----------|-------------|
 | `path` | `string` | Y | Path to a directory containing a `kustomization.yaml` file. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
 | `outPath` | `string` | Y | Path to the file or directory where rendered manifests are to be written. If the path ends with `.yaml` or `.yml` it is presumed to indicate a file and is otherwise presumed to indicate a directory. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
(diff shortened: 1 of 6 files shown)


lindhe:
Very nicely done! Smooth work making this a non-breaking change! And thanks a lot, this _really_ helps me out a great deal! 👍

lindhe:
FYI: I was just about to post an issue here complaining that the `kustomize` output format is broken, but turns out it's Kustomize that's broken! 😕 

I don't have the full picture yet, but there are many user reports of setting of namespace being broken since version `5.8.0`. For reference, see this and linked issues: https://github.com/kubernetes-sigs/kustomize/issues/6014

lindhe:
Hi!

Circling back to this one. Kustomize is not broken!

I've spent the whole day trying to understand the impact of https://github.com/kubernetes-sigs/kustomize/pull/5940 introduced in 5.8.1 and I have finally come around to believing in the change. I think 5.8.1 has correct behavior, but the change did cause some side-effects in a Helm chart for me (which does not use `.Release.Namespace` correctly), which threw me off.

And it even seems like Kargo is already aware of this changed behavior, to some extent. I found this, but don't know really what it means: https://github.com/akuity/kargo/blob/08332466561a86a9f7bc74d4c88511f0f189d44f/go.mod#L8

Currently, it seems like Kargo has the 5.7.1 behavior of overwriting `.metadata.namespace` for all resources, which is a huge issue anytime Helm templates uses `.Release.Namespace` anywhere other than in the `.metadata.namespace` field, for example inside `ConfigMap.data`. As mentioned, this was fixed in Kustomize 5.8.1, but based on the comment linked above I guess Kargo has pinned the old behavior for some reason?

Any idea if Kargo will move on to Kustomize 5.8.1 or if we're stuck for some technical reason?

krancour:
Please open a new issue.

PR #5443, opened 2025-12-01
fix(yaml): Fixes an issue with the yaml updater failing on long lines

thomastaylor312:
This is a two pronged fix: First, we weren't checking for an error in the scanner when we were reading lines. So now we check the error and return it. Second, we needed to allocate a larger buffer if updating a file with a really long line

### DIFF
--- a/pkg/yaml/yaml.go
+++ b/pkg/yaml/yaml.go
@@ -11,6 +11,14 @@ import (
 	"go.yaml.in/yaml/v3"
 )
 
+// The maximum size of the input buffer for processing YAML files. Given that the max size of a
+// Kubernetes manifest (the most common thing we're dealing with) is 1 MB, this number is chosen to
+// assume that the max value of any given key would be just shy of that number (imagine a config map
+// with a single large value which is a config file for something in the container). Please note
+// that we do not allocate a buffer of this size up front; this is just the maximum size that we
+// will allow when processing input.
+const maxBufferSize = 1024 * 1024 // 1 MB
+
 // Update represents a discrete update to be made to a YAML document.
 type Update struct {
 	// Key is the dot-separated path to the field to update.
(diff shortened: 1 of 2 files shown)


pkg/yaml/yaml.go
```diff
@@ -86,6 +94,11 @@ func SetValuesInBytes(inBytes []byte, updates []Update) ([]byte, error) {
 
 	scanner := bufio.NewScanner(bytes.NewBuffer(inBytes))
 	scanner.Split(bufio.ScanLines)
+	// Create an initial buffer of 100B which should be plenty for most lines. This means we won't
+	// immediately allocate a buffer on the first scan. If we encounter a line that exceeds this
+	// size, the buffer will be grow up to maxBufferSize.
```
hiddeco: Ultra nit: technically, we do allocate (100 bytes) — but not the max capacity.

thomastaylor312: Oh good catch. I updated after I wrote this

pkg/yaml/yaml_test.go
```diff
@@ -101,6 +101,36 @@ characters:
...
+			},
+		},
+		{
+			name: "really long lines still work",
+			// nolint:lll
+			inBytes: []byte(`
+characters:
+- name: Anakin
+  affiliation: Light side
+  temptation: "` + strings.Repeat("Did you ever hear the tragedy of Darth Plagueis The Wise? I thought not. It's not a story the Jedi would tell you. It's a Sith legend. Darth Plagueis was a Dark Lord of the Sith, so powerful and so wise he could use the Force to influence the midichlorians to create life...He had such a knowledge of the dark side that he could even keep the ones he cared about from dying. The dark side of the Force is a pathway to many abilities some consider to be unnatural. He became so powerful...the only thing he was afraid of was losing his power, which eventually, of course, he did. Unfortunately, he taught his apprentice everything he knew, then his apprentice killed him in his sleep. Ironic. He could save others from death, but not himself.", 1000) + `"
```
alexmt, line 116: ✌️

PR #5305, opened 2025-10-29
feat: generic webhook receiver

fuskovic:
Closes: https://github.com/akuity/kargo/issues/4486

PR #5446, opened 2025-12-03
feat: consolidate RefreshX funcs into a single func

fuskovic:
Addresses https://github.com/akuity/kargo/pull/5305#discussion_r2582572059

CLI test:

cli_test

UI test:

https://github.com/user-attachments/assets/065ee848-7df4-4063-9426-413835faaaf4

### DIFF
--- a/pkg/server/refresh_v1alpha1.go
+++ b/pkg/server/refresh_v1alpha1.go
@@ -48,10 +48,10 @@ func (t RefreshResourceType) NameEqualsProject() bool {
 
 // refreshObjectFactories maps resource types to their object factory functions.
 var refreshObjectFactories = map[string]func() client.Object{
-	strings.ToLower(string(RefreshResourceTypeClusterConfig)): func() client.Object { return new(kargoapi.ClusterConfig) },
-	strings.ToLower(string(RefreshResourceTypeProjectConfig)): func() client.Object { return new(kargoapi.ProjectConfig) },
-	strings.ToLower(string(RefreshResourceTypeStage)):         func() client.Object { return new(kargoapi.Stage) },
-	strings.ToLower(string(RefreshResourceTypeWarehouse)):     func() client.Object { return new(kargoapi.Warehouse) },
+	strings.ToLower(string(RefreshResourceTypeClusterConfig)): func() client.Object { return &kargoapi.ClusterConfig{} },
+	strings.ToLower(string(RefreshResourceTypeProjectConfig)): func() client.Object { return &kargoapi.ProjectConfig{} },
+	strings.ToLower(string(RefreshResourceTypeStage)):         func() client.Object { return &kargoapi.Stage{} },
+	strings.ToLower(string(RefreshResourceTypeWarehouse)):     func() client.Object { return &kargoapi.Warehouse{} },
 }
 
 func getRefreshObjectFactory(resourceType string) (func() client.Object, error) {


fuskovic:
@hiddeco I was originally going to deprecate + reserve the deleted messages + methods per gRPC standards but I spoke with @krancour offline and he explained it was ok to just delete them.

pkg/server/refresh_v1alpha1.go
```diff
@@ -0,0 +1,119 @@
...
+	if r.ResourceType != svcv1alpha1.RefreshResourceType_REFRESH_RESOURCE_TYPE_PROJECT_CONFIG {
+		if err := validateFieldNotEmpty("name", r.GetName()); err != nil {
+			return nil, err
+		}
+		o.SetName(r.GetName())
+	} else {
+		o.SetName(r.GetProject())
+	}
+	return &o, nil
+}
```
krancour, line 158: I see lots of switch/case here. Is there not some way to do all of this more generically?

krancour:
> @hiddeco I was originally going to deprecate + reserve the deleted messages + methods per gRPC standards but I spoke with @krancour offline and he explained it was ok to just delete them.

A word of explanation. Other, unrelated changes are going to force those who upgrade to v1.9.0 on the server-side to _also_ update their CLI. This being the case, I don't _think_ there's any reason to not take advantage of the opportunity that presents.

api/service/v1alpha1/service.proto
```diff
@@ -1840,3 +1785,33 @@ message DeleteClusterSecretRequest {
 message DeleteClusterSecretResponse {
   /* explicitly empty */
 }
+
+// RefreshResourceType is an enumeration of resource types that can be refreshed.
+enum RefreshResourceType {
```
hiddeco: I am not a big fan of numeric enums because they can become hard to manage over time due to issues around ordering and removal.

I believe an acceptable alternative in Protobuf is to simply make this a string and handle the validation on the server side.

krancour: I concur with this.

pkg/server/refresh_v1alpha1.go
fuskovic: There's no way around needing to resolve the concrete type from a string that satisfies the `client.Object` implementation ( that I can see ) . Even if `getClientObject` was generic, you would still need to do the switch somewhere else.

fuskovic: Ended up using a map

pkg/cli/cmd/refresh/refresh.go
```diff
@@ -132,37 +102,24 @@ func (o *refreshOptions) run(ctx context.Context) error {
...
-		case refreshResourceTypeProjectConfig:
+		case server.RefreshResourceTypeProjectConfig:
 			err = waitForProjectConfig(ctx, kargoSvcCli, o.Project)
-		case refreshResourceTypeStage:
+		case server.RefreshResourceTypeStage:
 			err = waitForStage(ctx, kargoSvcCli, o.Project, o.Name)
-		case refreshResourceTypeWarehouse:
+		case server.RefreshResourceTypeWarehouse:
 			err = waitForWarehouse(ctx, kargoSvcCli, o.Project, o.Name)
 		}
```
fuskovic, line 123: If this PR is viable, I can explore consolidating `waitForX` into a single func in a separate PR if that's something we're interested in.

api/service/v1alpha1/service.proto
```diff
@@ -1840,3 +1785,19 @@ message DeleteClusterSecretRequest {
...
+  // leave blank if refreshing a project or cluster config.
+  string name = 2;
+  // resource_type is the kind of resource to refresh.
+  // should be one of: ProjectConfig, ClusterConfig, Warehouse, or Stage.
+  string resource_type = 3;
+}
+
+message RefreshResourceResponse {
+  google.protobuf.Any resource = 1;
+}
```
hiddeco, line 1803: Nit:

```suggestion

}
```

pkg/server/refresh_v1alpha1.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	rt := RefreshResourceType(r.GetResourceType())
+	object, ok := ObjectRefreshResourceType[rt]
+	if !ok {
+		return nil, connect.NewError(
+			connect.CodeInvalidArgument,
+			fmt.Errorf("unsupported refresh kind: %s", rt),
+		)
+	}
+	object.SetNamespace(om.GetNamespace())
```
hiddeco: This directly modifies the object in the map. The map should either contain a factory function, or you should do an `object.DeepCopy()` before modifying it.

pkg/server/refresh_v1alpha1.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (s *server) getObjectMeta(ctx context.Context, r *svcv1alpha1.RefreshResourceRequest) (*metav1.ObjectMeta, error) {
+	var o metav1.ObjectMeta
+	rt := RefreshResourceType(r.GetResourceType())
+	if !rt.RequiresName() && !rt.RequiresProject() {
+		o.SetName(api.ClusterConfigName)
+		return &o, nil
+	}
+
+	if rt.RequiresProject() {
```
hiddeco: I think it would be better to ask the scheme if the object is namespaced: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/apiutil#IsObjectNamespaced

pkg/server/refresh_v1alpha1.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	if rt.RequiresName() {
+		if err := validateFieldNotEmpty("name", r.GetName()); err != nil {
+			return nil, err
+		}
+		o.SetName(r.GetName())
+		return &o, nil
+	}
+	o.SetName(r.GetProject())
```
hiddeco: An object not requiring a name appears to be an exception (and is the atypical thing in general). I would model things around this instead.

pkg/server/refresh_v1alpha1.go
hiddeco: Or even simpler (without requiring an object): https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/apiutil#IsGVKNamespaced

pkg/server/refresh_v1alpha1.go
hiddeco: The alternative is to just create a function with a `switch` statement. No strong opinion here.

pkg/server/refresh_v1alpha1.go
hiddeco: Actually, even more descriptive would probably something like `NameEqualsProject()` be.

api/service/v1alpha1/service.proto
hiddeco: 🤦 

I meant

```suggestion
}

```

pkg/server/refresh_v1alpha1.go
fuskovic: I wanted something the `cli` module could use. Unfortunately, using these utilities is awkward in the CLI due to there being no clear way (that I know of) to source a REST mapper from the `kargoServiceCLI`. I found it much simpler to just expose an `IsNamespaced` method that simply checks if we're dealing with a `ClusterConfig`

pkg/cli/cmd/refresh/refresh.go
```diff
@@ -84,12 +84,13 @@ func (o *refreshOptions) validate() error {
...
-		errs = append(errs, errors.New("project is required"))
-	}
-
-	if !o.ResourceType.NameEqualsProject() && o.Name == "" {
-		errs = append(errs, errors.New("name is required"))
+	if o.ResourceType.IsNamespaced() {
+		if o.Project == "" {
+			errs = append(errs, errors.New("project is required"))
+		}
+		if !o.ResourceType.NameEqualsProject() && o.Name == "" {
```
hiddeco: I believe this check should be outside of `o.ResourceType.IsNamespaced()`.

pkg/server/refresh_v1alpha1.go
```diff
@@ -0,0 +1,245 @@
...
+// They are camel-cased versions of the Kargo resource kinds for compatibility
+// purposes with Kubernetes REST mappers.
+const (
+	RefreshResourceTypeClusterConfig RefreshResourceType = "ClusterConfig"
+	RefreshResourceTypeProjectConfig RefreshResourceType = "ProjectConfig"
+	RefreshResourceTypeStage         RefreshResourceType = "Stage"
+	RefreshResourceTypeWarehouse     RefreshResourceType = "Warehouse"
+)
+
+func init() {
```
hiddeco: Still debating if I find this over-engineered or justified...

pkg/cli/cmd/refresh/refresh.go
fuskovic: Moving this outside of `o.ResourceType.IsNamespaced()` causes CLI failures for `clusterconfigs`

hiddeco: But the current rule as-is implies that only a namespaced object must have a name. Which doesn't seem correct?

fuskovic: It is correct that only a namespaced object needs to have a name since we don't accept a name parameter for the refresh cluster config sub command.

fuskovic: Resolving per offline discussion

PR #5447, opened 2025-12-03
feat(httproute): add Gateway API HTTPRoute support

abelhoula:


charts/kargo/templates/api/httproute.yaml
```diff
@@ -0,0 +1,46 @@
...
+        - path:
+            type: PathPrefix
+            value: /webhooks
+      backendRefs:
+        - kind: Service
+          name: kargo-external-webhooks-server
+          port: {{ $webhookPort }}
+          weight: 1
+    {{- end }}
+{{- end }}
```
hiddeco:
```suggestion
{{- end }}

```

charts/kargo/templates/external-webhooks-server/httproute.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /
+      backendRefs:
+        - kind: Service
+          name: kargo-external-webhooks-server
+          port: {{ $webhookPort }}
+          weight: 1
+{{- end }}
```
hiddeco:
```suggestion
{{- end }}

```

krancour:
@abelhoula thanks for working on this and for your patience waiting for an in-depth review. I think this is a really worthwhile feature, ideally, I'd like to see this land in v1.11.0.

My main ask would be for better parity with what's already supported with `Ingress` today. Our policy is for the chart to be more secure than not by default. This means I'd like to keep `api.tls.enabled: true` as a default, as it is today, which is incompatible with an `HTTPRoute` unless it uses a `BackendTLSPolicy` to re-establish TLS between the gateway and the backend. I realize that many people, in practice, terminate TLS at the gateway and that's that. And that's fine. But I want people to opt into relaxing security instead of opting into tightening it.

So, basically, what I'm saying is instead of `api.tls.enabled: true` with `api.httproute.enabled: true` leading to invalid configuration, I'd rather that combination just lead automatically to the creation of a `BackendTLSPolicy` to make everything work correctly with TLS end-to-end. This would require surfacing options in `api.httpRoute.tls` like those in `api.ingress.tls`.

Other varied feedback I have is more targeted, so I'll leave comments inline with the diffs.

charts/kargo/values.yaml
```diff
@@ -454,6 +454,22 @@ api:
     ## @param api.ingress.pathType You may want to use `Prefix` for some controllers (like AWS LoadBalancer Ingress controller), which don't support `/` as wildcard path when pathType is set to `ImplementationSpecific`
     pathType: ImplementationSpecific
 
+  httproute:
+    ## @param api.httproute.enabled Whether to enable HTTPRoute by creating an HTTPRoute resource. By default, this is disabled. Enabling HTTPRoute is advanced usage.
+    ## @param externalWebhooksServer.httproute.enabled -- Enable Gateway API HTTPRoute.
+    ## Gateway API support is in EXPERIMENTAL status. Support depends on your Gateway controller implementation.
```
krancour, line 460: I think it's GA now, but you can double-check me on that.

charts/kargo/values.yaml
```diff
@@ -454,6 +454,22 @@ api:
...
+    ## See the [Gateway API documentation](https://gateway-api.sigs.k8s.io/) for details. By default, this is disabled.
+    enabled: false
```
krancour, line 462: Somewhere in the comment above, it may be worth mentioning that setting this to `true` requires you to have the Gateway API CRDs installed on your cluster.

charts/kargo/values.yaml
```diff
@@ -771,6 +787,21 @@ externalWebhooksServer:
     ## @param externalWebhooksServer.ingress.pathType You may want to use `Prefix` for some controllers (like AWS LoadBalancer Ingress controller), which don't support `/` as wildcard path when pathType is set to `ImplementationSpecific`
     pathType: ImplementationSpecific
 
+  httproute:
```
krancour, line 790: Same comments from the other `httproute` section apply here as well.

charts/kargo/templates/api/httproute.yaml
```diff
@@ -0,0 +1,46 @@
...
+  annotations:
+    {{- range $key, $value := . }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
+spec:
+  parentRefs:
+    {{- with .Values.api.httproute.parentRefs }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
```
krancour, line 22: I would suggest doing it like this so that empty/undefined `.Values.api.httproute.parentRefs` doesn't leave us with a key without a value here.

```suggestion
  {{- with .Values.api.httproute.parentRefs }}
  parentRefs:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

krancour:
This is an enhancement I'm really supportive of if/when the feedback's been addressed. I'll leave it open, but I'm removing it from the v1.11.0 milestone since this is not critical for that release and progress on the PR seems to have stalled.

@abelhoula if you don't intend to resume work on this, please let me know so I have find someone else to take it over.

PR #5451, opened 2025-12-05
fix(gitprovider): Normalize Git URL before parsing in registration

aslafy-z:
The URLs are now normalized before parsing.

With urls in the form `git@example.com:org/repo.git`, `url.Parse` fails at correctly parsing and throws an error: https://play.golang.com/p/M7w-slmNgaB.

### DIFF
--- a/pkg/gitprovider/azure/azure_test.go
+++ b/pkg/gitprovider/azure/azure_test.go
@@ -125,3 +125,49 @@ func TestGetCommitURL(t *testing.T) {
 		})
 	}
 }
+
+func TestRegistrationPredicate(t *testing.T) {
+	testCases := []struct {
+		name     string
+		repoURL  string
+		expected bool
+	}{
+		{
+			name:     "invalid URL",
+			repoURL:  "not-a-url",
+			expected: false,
+		},
+		{
+			name:     "unsupported host",
+			repoURL:  "https://github.com/org/repo",
+			expected: false,
+		},
+		{
+			name:     "modern URL format",
+			repoURL:  "https://dev.azure.com/myorg/myproject/_git/myrepo",
+			expected: true,
+		},
+		{
+			name:     "modern URL format with .git suffix",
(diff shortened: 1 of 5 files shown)


krancour:
Reviewed from my phone and hope I didn't miss anything due to the small viewport, but it seems in all cases, the URL is being parsed and only the hostname is any sort of factor in determining whether a provider supports a the URL. It's not clear how normalizing the URLs -- which will never change the hostname -- would make any difference at all to any of this logic.

But, I presume you encountered some scenario that compelled you to open this PR, so could you please explain how this PR solves the concrete problem you ran into?

aslafy-z:
Hello @krancour, 
Thank you for your quick feedback. 
With urls in the form `git@example.com:org/repo.git`, `url.Parse` fails at correctly parsing and throws an error: https://play.golang.com/p/M7w-slmNgaB.

fuskovic (approved):
> With urls in the form `git@example.com:org/repo.git`, `url.Parse` fails at correctly parsing and throws an error: https://play.golang.com/p/M7w-slmNgaB.

I've confirmed normalizing the git URL does indeed fix the playground error.

https://play.golang.com/p/ipF2gvHI-iH

I'm going to approve and start the checks.

Will defer merge to @krancour

aslafy-z:
@krancour I added tests for registration.Predicate to accomodate codecov and rebased the PR, fixed DCO. Please give a look.

krancour:
This appears related to #5452 and I am rejecting it for related reasons.

The APIs used by these providers do not use SSH. Yes... as discussed in #5452, some of the providers do contain unreliable logic that infer HTTPS equivalents of SSH URLs -- but that logic is basically unreachable right now because none of the promotion steps that use these providers allow SSH URLs in the door in the first place (nor should they, which is why #5452 was closed).

Without #5452, this change has no effect. In a hypothetical world where #5452 and this PR were both merged, broadening the range of SSH URLs "supported" by code that integrates with APIs that _don't even use SSH_ is not an improvement.

PR #5452, opened 2025-12-05
feat(promotion): allow ssh urls into pr builtins

aslafy-z:
Remove 'uri' validation for repoURI as it does not match `git@example.com:org/repo.git` syntax.

### DIFF
--- a/pkg/promotion/runner/builtin/schemas/git-merge-pr-config.json
+++ b/pkg/promotion/runner/builtin/schemas/git-merge-pr-config.json
@@ -8,8 +8,7 @@
       "repoURL": {
         "type": "string",
         "description": "The URL of the remote Git repository containing the pull request.",
-        "minLength": 1,
-        "format": "uri"
+        "minLength": 1
       },
       "prNumber": {
         "type": "integer",
--- a/pkg/promotion/runner/builtin/schemas/git-open-pr-config.json
+++ b/pkg/promotion/runner/builtin/schemas/git-open-pr-config.json
@@ -21,8 +21,7 @@
     "repoURL": {
       "type": "string",
       "description": "The URL of a remote Git repository to clone.",
-      "minLength": 1,
-      "format": "uri"
+      "minLength": 1
     },
     "sourceBranch": {
       "type": "string",
--- a/pkg/promotion/runner/builtin/schemas/git-wait-for-pr-config.json
+++ b/pkg/promotion/runner/builtin/schemas/git-wait-for-pr-config.json
@@ -22,8 +22,7 @@
     "repoURL": {
       "type": "string",
       "description": "The URL of a remote Git repository to clone.",
-      "minLength": 1,
-      "format": "uri"
+      "minLength": 1
     }
   }
 }
--- a/ui/src/gen/directives/git-merge-pr-config.json
+++ b/ui/src/gen/directives/git-merge-pr-config.json
@@ -7,8 +7,7 @@
   "repoURL": {
    "type": "string",
    "description": "The URL of the remote Git repository containing the pull request.",
-   "minLength": 1,
-   "format": "uri"
+   "minLength": 1
   },
   "prNumber": {
    "type": "integer",
--- a/ui/src/gen/directives/git-open-pr-config.json
+++ b/ui/src/gen/directives/git-open-pr-config.json
@@ -26,8 +26,7 @@
   "repoURL": {
    "type": "string",
    "description": "The URL of a remote Git repository to clone.",
-   "minLength": 1,
-   "format": "uri"
+   "minLength": 1
   },
   "sourceBranch": {
    "type": "string",
--- a/ui/src/gen/directives/git-wait-for-pr-config.json
+++ b/ui/src/gen/directives/git-wait-for-pr-config.json
@@ -27,8 +27,7 @@
   "repoURL": {
    "type": "string",
    "description": "The URL of a remote Git repository to clone.",
-   "minLength": 1,
-   "format": "uri"
+   "minLength": 1
   }
  }
 }


krancour:
There may be a question here of exactly how useful an SSH url actually is in this context. All the providers' APIs are over HTTP/S. Even if you could infer details from an SSH url that allowed building a client for any provider's API, you surely need to authenticate to that API. When you go searching for credentials for an SSH URL, what you're going to end up with is an SSH key. To the best of my knowledge, and I am happy to be corrected if I am wrong, none of these APIs will accept that as a credential.

aslafy-z:
You're right, these APIs does not handle an ssh key. 
In my case, I'm cloning with SSH and my repo endpoint looks like: `git@example.com:group/repo.git`.

I tried giving an HTTPs url in there, but these tasks clone before doing anything with the API. At the time, I messed up my token permission, which prevented me from cloning using HTTPS. After looking at the code, I've seen that these tasks are well aware that the URL can be a SSH endpoint instead of an HTTPS one, they do a [best-effort "conversion"](https://github.com/akuity/kargo/blob/main/pkg/gitprovider/gitlab/gitlab.go#L86-L98).

 In this PR, I removed the 'uri' validator as `git@example.com:group/repo.git` did not match it but the plugin stills works with it.
 
In the meantime workarounded by using `ssh://git@example.com/group/repo.git` syntax instead.

I also finally understood why my HTTPS clone was not working at the start.

krancour:
I'm afraid these changes still don't make a great deal of sense.

Supposing we did merge this, these steps would allow ssh urls as input. _But then what?_

I suppose at least one provider already attempts a "best effort conversion" to an https equivalent as you pointed out, but that's a whole other can of worms. There's probably a half-dozen other open issues that request (or require as an enabler) some way to map an ssh url to its https equivalent. There really isn't a reliable way to do that. The last thing I'd want to do is take the naive approach used in the spot you reference and expand its use to many other places. We'll come back to this.

Back to the "now what?" The step now has an ssh url in-hand. Absent the elusive, reliable ssh --> http conversion discussed above, what is the next step? It's looking up credentials using the ssh url as input. This will find an ssh key. That key is useless for authenticating to any of the APIs.

So, now let's pretend we _did_ have reliable url conversions already accounted for. Or pretend we grudgingly expanded use of the naive approach. OR even pretend that instead of doing such a conversion automatically, you just handed us the https URL. Any of these three get you to the same place. You have an https url that the step will use as input to find credentials. You might now get back some type of credential other than an ssh key -- something you can actually use to authenticate to one of these APIs. Fantastic, right?

Or not...

Because your insistence on using ssh for basic git operations like cloning, pushing, etc. means you need to provide an ssh key for all of those operations _but you also have to provide some other kind of credentials for all the operations involving provider-specific APIs._ You force yourself to use two credentials where you could have used one instead. Even if that doesn't bother you, it's not really the user experience I want to promote.

So I'm sure this isn't what you want to hear, and some others might take umbrage with it as well, but my honest recommendation is that people _not_ use ssh repo URLs with Kargo. By doing so, you save yourself from defining two creds where one would have sufficed. You spare yourself from looking for workarounds for the places where ssh urls don't work. You spare the project from feature requests that we can't do a really good job of because of the infeasibility of reliably converting ssh urls to https.

If it were entirely up to me, I'd rip ssh support out of Kargo entirely. I can't see that it's doing anyone any favors.

aslafy-z:
Fair enough. Can I migrate all my warehouses to https without losing history?
Also, I think it would be nice to expose this view in the initial path of the docs.

aslafy-z:
Fair enough. Can I migrate all my warehouses to https without losing history?
Also, I think it would be nice to expose this view in the initial path of the docs.

krancour:
You won't lose history, per se, but because subscription URLs are used as a sort of identifier in many places, you _might_ encounter difficulties re-promoting older Freight to your stages, but I think that's going to be entirely dependent on the particulars of the promotion process used by each stage.

> Also, I think it would be nice to expose this view in the initial path of the docs.

That's a fair suggestion. I will open an issue for it.

krancour:
Opened #5457

aslafy-z:
@krancour I still think this should be merged to align the behavior of the different modules. It works with `ssh://git@example.com/org/repo.git` uris, not with the currently supported `git@example.com:org/repo.git` syntax.

krancour:
@aslafy-z there's still that "but then what?" question.

If we let ssh urls in the door, it's still a problem that looking up creds by URL is going to find you an ssh key and you can't use that to authenticate to any of the APIs.

Unless I'm missing something, this PR does not actually make these steps any more usable with ssh urls than they already are.

aslafy-z:
@krancour 

They do in fact. Today, it works with `ssh://git@example.com/org/repo.git` which gets normalized to `https://gitlab.com/org/repo.git`, and credentials are looked up for the later.
However, `git@gitlab.com:org/repo.git` (which is normalized to the same https url) does not pass the openapi validation.
It looks like a bug to me, as the implementation supports it.

krancour:
If that ssh URL is normalizing as https, it is a serious bug. As we've discussed, we are specifically not doing transformations across protocols because we cannot do so reliably.

According to the unit tests for normalization, urls of the form `ssh://git@example.com/org/repo.git` do not normalize to `https://gitlab.com/org/repo.git` as you claim. In fact, `https://gitlab.com/org/repo.git` does not even look like a normalized https URL, as it has a trailing `.git` that normalization would have stripped.

https://github.com/akuity/kargo/blob/main/pkg/urls/git_test.go

I'm willing to consider the possibility that you're observing a bug in normalization that the tests didn't catch.

aslafy-z:
"normalized" is not the right term, my bad. 

I was mentioning this:

https://github.com/akuity/kargo/blob/main/pkg%2Fgitprovider%2Fgitlab%2Fgitlab.go#L279

https://github.com/akuity/kargo/blob/main/pkg%2Fgitprovider%2Fgithub%2Fgithub.go#L385

- url gets normalized
- https url gets discovered from that normalized url

aslafy-z:
@krancour Should I open an issue about this?

krancour:
Ok... so some of the Git provider implementations make some unreliable inferences about what the http/s equivalent of an ssh URL (specifically of the form `ssh://git@example.com/org/repo.git`), but it doesn't change much.

I'll repeat what I have been saying from the start. When these promotion steps try to find a credential using an SSH URL, the only thing they might get back is an SSH key and you cannot use that as a credential with any of the Git providers' APIs.

It doesn't matter if a provider, using the unreliable logic you have cited, concludes that it can support an SSH URL (of whatever form) if you are then going to hand it a credential that it cannot use.

This is probably not what you want to hear, but this isn't going to be merged and isn't up for continued debate. The reliable thing to do is not use SSH if you need to interact with your Git hosting provider through these APIs, as these steps do.

PR #5454, opened 2025-12-05
feat(gitprovider/gitlab): Support for squash merge and fast forward

aslafy-z:
Currently, `gitprovider.PullRequest.MergeCommitSHA` is set only from `glMR.MergeCommitSHA` by the GitLab provider.

This works for merge strategies that create a merge commit, but it breaks for repositories configured to use fast-forward merges without a merge commit: in those cases, the commit information stays empty even though the merge request is successfully merged.

GitLab’s API exposes three relevant fields on merge requests:

- `merge_commit_sha`: SHA of the merge commit (when one exists).
- `squash_commit_sha`: SHA of the squash commit.
- `sha`: SHA of the head commit in the source branch.

According to the documentation, a squashed merge is supposed to populate `squash_commit_sha`, and (depending on merge method) may or may not create a merge commit. However, there are known inconsistencies in GitLab’s current implementation:

- Squash merges currently populate `merge_commit_sha` with the squashed commit and leave `squash_commit_sha` unset in the webhook payloads.
- There is ongoing work to backfill and correctly populate `squash_commit_sha` and to fix related webhook behavior.

See:

- https://gitlab.com/gitlab-org/gitlab/-/issues/415449
- https://gitlab.com/gitlab-org/gitlab/-/issues/294257
- https://gitlab.com/groups/gitlab-org/-/epics/20145

To handle both the current behavior and the intended future behavior, this PR updates `convertGitlabMR` to compute an “effective merge commit SHA” only when the merge request is in the `merged` state, using the following precedence:

1. `glMR.MergeCommitSHA` if it is set  
   - Covers today’s GitLab behavior for classic merge commits, and for squash merges where GitLab currently puts the squash commit into `merge_commit_sha`.
2. `glMR.SquashCommitSHA` if `MergeCommitSHA` is empty but `SquashCommitSHA` is set  
   - Covers the future / intended behavior and fast-forward-with-squash cases where there is no merge commit but a squash commit exists.
3. `glMR.SHA` if both `MergeCommitSHA` and `SquashCommitSHA` are empty  
   - Covers fast-forward-without-squash, where no merge or squash commit is created and the target branch is advanced to the source tip.

For unmerged merge requests, `MergeCommitSHA` remains empty.

This makes the GitLab provider resilient to both current and future GitLab implementations while ensuring `MergeCommitSHA` accurately reflects the commit that actually landed on the target branch for merged requests.

### DIFF
--- a/pkg/gitprovider/gitlab/gitlab.go
+++ b/pkg/gitprovider/gitlab/gitlab.go
@@ -270,12 +270,12 @@ func convertGitlabMR(glMR gitlab.BasicMergeRequest) gitprovider.PullRequest {
 			mergeCommit = glMR.MergeCommitSHA
 		case glMR.SquashCommitSHA != "":
 			// Fast-forward with squash:
-			// No merge commit exists, but a squash commit SHA may be available.
+			// No merge_commit_sha exists; fall back to squash_commit_sha.
 			//
 			// Note: GitLab does not reliably populate squash_commit_sha in all
-			// relevant API/webhook payloads yet. We still prefer it when present,
-			// and keep this branch to future-proof against incoming GitLab changes
-			// that aim to backfill and correctly populate squash_commit_sha.
+			// relevant API/webhook payloads yet. We keep this branch to
+			// future-proof against incoming GitLab changes that aim to backfill
+			// and correctly populate squash_commit_sha.
 			//
 			// References:
 			// - https://gitlab.com/gitlab-org/gitlab/-/issues/415449


pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -260,12 +260,19 @@ func (p *provider) GetCommitURL(repoURL string, sha string) (string, error) {
 }
 
 func convertGitlabMR(glMR gitlab.BasicMergeRequest) gitprovider.PullRequest {
+	mergeCommit := glMR.MergeCommitSHA
+	if mergeCommit == nil {
+		mergeCommit = glMR.SquashCommitSHA
+	}
+	if mergeCommit == nil {
+		mergeCommit = glMR.SHA
+	}
```
hiddeco: I think it would be more logical to do something like:


```suggestion
		var merged = glMR.State == "merged"
		var mergeCommit string
		if merged {
				switch {
				case glMR.MergeCommitSHA != "":
					mergeCommit = glMR.MergeCommitSHA
				case glMR.SquashCommitSHA != "":
					mergeCommit = glMR.SquashCommitSHA
				default:
					// Fast-forward merge: no merge commit exists,
					// the source HEAD is now on target
					mergeCommit = glMR.SHA
				}	
		}
```

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -260,12 +260,19 @@ func (p *provider) GetCommitURL(repoURL string, sha string) (string, error) {
...
 	return gitprovider.PullRequest{
 		Number:         glMR.IID,
 		URL:            glMR.WebURL,
 		Open:           isMROpen(glMR),
 		Merged:         glMR.State == "merged",
-		MergeCommitSHA: glMR.MergeCommitSHA,
+		MergeCommitSHA: mergeCommit,
```
hiddeco:
```suggestion
		Merged:         merged,
		MergeCommitSHA: mergeCommit,
```

krancour:
@fuskovic because of your familiarity with all the webhook receivers, I'd like you to look at this PR as well, please. Please consult the GitLab docs and/or experiment a bit with their API to verify this is indeed how things work on their end.

pkg/gitprovider/gitlab/gitlab.go
aslafy-z: I applied this pattern

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -260,12 +260,30 @@ func (p *provider) GetCommitURL(repoURL string, sha string) (string, error) {
...
+	if merged {
+		switch {
+		case glMR.MergeCommitSHA != "":
+			// Regular merge (with or without squash commit),
+			// GitLab gives us the merge commit.
+			mergeCommit = glMR.MergeCommitSHA
+		case glMR.SquashCommitSHA != "":
+			// Fast-forward with squash: no merge commit,
+			// but a squash commit exists.
+			mergeCommit = glMR.SquashCommitSHA
```
fuskovic, line 283: Since Gitlab doesn't populate this field yet, it would be useful to have the comment reflect that this logic is intended to future-proof incoming changes. Links to relevant issues and discussions for aforementioned incoming changes could also be useful here.

fuskovic (changes requested):
I think some additional context in the comments for the future-proofing behavior could be useful.

pkg/gitprovider/gitlab/gitlab.go
aslafy-z: @fuskovic I added a comment as requested. Please let me know if it fits you.

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -260,12 +260,39 @@ func (p *provider) GetCommitURL(repoURL string, sha string) (string, error) {
...
+		case glMR.MergeCommitSHA != "":
+			// Regular merge (with or without squash):
+			// GitLab gives us the merge commit SHA.
+			mergeCommit = glMR.MergeCommitSHA
+		case glMR.SquashCommitSHA != "":
+			// Fast-forward with squash:
+			// No merge commit exists, but a squash commit SHA may be available.
+			//
+			// Note: GitLab does not reliably populate squash_commit_sha in all
+			// relevant API/webhook payloads yet. We still prefer it when present,
```
fuskovic: > We still prefer it when present

The sub-test `prefers MergeCommitSHA when present` and the switch case evaluating the `merge_commit_sha` first implies, the `squash_commit_sha` is _NOT_ necessarily always preferred.

So I think we just need to change the comment to read that we "fallback to the `squash_commit_sha` when `merge_commit_sha` is not present". Which is consistent with the sub-test `falls back to SquashCommitSHA when MergeCommitSHA is empty`.

Although, since that's pretty self-explanatory from just looking at the code we can just delete "We still prefer it when present".

On a side note, I've confirmed that the existing behavior does populate `merge_commit_sha` with the `squash_commit_sha` as stated in the PR description.

squash_commit

Just to make sure that we didn't need to have the `squash_commit_sha` be the first case in the switch.

aslafy-z: I updated the comment in 662d0e8b3670838bddcba6dd8bf57696fa3e061a. Let me know if it fits you.

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -260,12 +260,40 @@ func (p *provider) GetCommitURL(repoURL string, sha string) (string, error) {
...
+	if merged {
+		switch {
+		case glMR.MergeCommitSHA != "":
+			// Regular merge (with or without squash):
+			// GitLab gives us the merge commit SHA.
+			mergeCommit = glMR.MergeCommitSHA
+		case glMR.SquashCommitSHA != "":
+			// Fast-forward with squash:
+			// No merge commit exists; fall back to squash_commit_sha when
+			// merge_commit_sha is empty.
```
fuskovic:
```suggestion
			// Fast-forward with squash:
			// No merge_commit_sha exists; fall back to squash_commit_sha.
```

aslafy-z: See 764b04535b7d1d0d29fce0cf13a98622e741e809

krancour:
I'm going to close this in favor of #5581 since that one essentially adds this feature for all Git hosting providers.

We'll be mindful of the possibility that there may be bits of GitLab-specific stuff in this PR that's worth porting over to #5581.

PR #5456, opened 2025-12-06
fix(webhook): gitlab ssh url warehouse not refreshed

aslafy-z:
GitLab equivalent fix of https://github.com/akuity/kargo/pull/5432 

https://docs.gitlab.com/user/project/integrations/webhook_events/#push-events

### DIFF
--- a/pkg/webhook/external/gitlab.go
+++ b/pkg/webhook/external/gitlab.go
@@ -120,7 +120,10 @@ func (g *gitlabWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc 
 		case *gl.PushEvent:
 			var repoURLs []string
 			if e.Repository != nil {
-				repoURLs = []string{urls.NormalizeGit(e.Repository.GitHTTPURL)}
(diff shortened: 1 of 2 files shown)


pkg/webhook/external/gitlab.go
```diff
@@ -120,7 +120,10 @@ func (g *gitlabWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 		case *gl.PushEvent:
 			var repoURLs []string
 			if e.Repository != nil {
-				repoURLs = []string{urls.NormalizeGit(e.Repository.GitHTTPURL)}
+				repoURLs = []string{
+					urls.NormalizeGit(e.Repository.GitHTTPURL),
+					urls.NormalizeGit(e.Repository.GetSSHURL),
```
krancour: There's a typo here... this couldn't possibly have compiled.

aslafy-z: Thanks for pointing it out. I've fixed the typo and the code now compiles correctly.

PR #5460, opened 2025-12-07
fix: typo on stage health change message

Sobuno:
Message was identical to the message for Warehouse health change  - Probably a copy-paste error.


hiddeco (approved):
Thank you!

PR #5469, opened 2025-12-08
feat: add draft support for gitea and improve draft detection for other providers

fykaa:
this resolves the todo item related to https://gitea.com/gitea/go-sdk/pulls/731

### DIFF
--- a/pkg/gitprovider/gitea/gitea.go
+++ b/pkg/gitprovider/gitea/gitea.go
@@ -297,12 +297,7 @@ func (p *provider) MergePullRequest(
 	case giteaPR.State != gitea.StateOpen:
 		return nil, false, fmt.Errorf("pull request %d is closed but not merged", id)
 
-	case giteaPR.Draft:
-		// Draft PRs cannot be merged
-		return nil, false, nil
-
-	case !giteaPR.Mergeable:
-		// Not ready to merge yet (conflicts, failing checks, etc.)
+	case giteaPR.Draft || !giteaPR.Mergeable:
 		return nil, false, nil
 	}
 
--- a/pkg/gitprovider/github/github.go
+++ b/pkg/gitprovider/github/github.go
@@ -316,12 +316,7 @@ func (p *provider) MergePullRequest(
 	case ptr.Deref(ghPR.State, prStateClosed) != prStateOpen:
 		return nil, false, fmt.Errorf("pull request %d is closed but not merged", id)
 
-	case ptr.Deref(ghPR.Draft, false):
-		// Draft PRs cannot be merged
-		return nil, false, nil
-
-	case ghPR.Mergeable == nil || !*ghPR.Mergeable:
-		// Not ready to merge yet (conflicts, failing checks, etc.)
+	case ptr.Deref(ghPR.Draft, false) || !ptr.Deref(ghPR.Mergeable, false):
 		return nil, false, nil
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/gitprovider/bitbucket/bitbucket.go
```diff
@@ -333,7 +333,7 @@ func (p *provider) MergePullRequest(
 
 	// Check if PR is draft - cannot merge draft PRs
 	if bbPR.Draft {
-		return nil, false, fmt.Errorf("cannot merge pull request %d: pull request is in draft state", id)
```
krancour, line 336: Was this an unrelated bug that you happened to notice while working on this PR?

pkg/gitprovider/gitea/gitea.go
```diff
@@ -297,16 +297,15 @@ func (p *provider) MergePullRequest(
 	case giteaPR.State != gitea.StateOpen:
 		return nil, false, fmt.Errorf("pull request %d is closed but not merged", id)
 
+	case giteaPR.Draft:
+		// Draft PRs cannot be merged
+		return nil, false, nil
+
 	case !giteaPR.Mergeable:
+		// Not ready to merge yet (conflicts, failing checks, etc.)
 		return nil, false, nil
```
krancour: imho, the extra case isn't worth it just to have a different comment. We can do this with one case and it's adequately self-documenting:

```suggestion
	case giteaPR.Draft || !giteaPR.Mergeable:
		return nil, false, nil
```

pkg/gitprovider/github/github.go
```diff
@@ -316,7 +316,12 @@ func (p *provider) MergePullRequest(
...
 		return nil, false, fmt.Errorf("pull request %d is closed but not merged", id)
 
-	case ghPR.Mergeable == nil || !*ghPR.Mergeable || (ghPR.Draft != nil && *ghPR.Draft):
+	case ptr.Deref(ghPR.Draft, false):
+		// Draft PRs cannot be merged
+		return nil, false, nil
+
+	case ghPR.Mergeable == nil || !*ghPR.Mergeable:
+		// Not ready to merge yet (conflicts, failing checks, etc.)
 		return nil, false, nil
```
krancour: If you're going to use `ptr.Deref()` for checking on draft status, it would be odd to not also use it for checking mergeability almost immediately afterward. Also, it's not worth these being separate cases for the sake of different comments. This statement is adequately self-documenting.

```suggestion
	case ptr.Deref(ghPR.Draft, false) || !ptr.Deref(ghPR.Mergeable, false)
		return nil, false, nil
```

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -226,7 +226,12 @@ func (p *provider) MergePullRequest(
 	case glMR.State != "opened":
 		return nil, false, fmt.Errorf("pull request %d is closed but not merged", id)
 
+	case glMR.Draft:
+		// Draft MRs cannot be merged
+		return nil, false, nil
+
 	case glMR.DetailedMergeStatus != "mergeable":
+		// Not ready to merge yet (conflicts, failing checks, etc.)
 		return nil, false, nil
```
krancour:
```suggestion
	case glMR.Draft || glMR.DetailedMergeStatus != "mergeable":
		return nil, false, nil
```

krancour: Also, was it a bug that draft status wasn't already being checked here?

pkg/gitprovider/bitbucket/bitbucket.go
fykaa: Yes, because we shouldn't be throwing an error if a PR is in draft state.

pkg/gitprovider/gitlab/gitlab.go
fykaa: Yes

Issue #5012, opened 2025-09-08
support alternative start and end sequences for expressions


johnthompson-ybor:
I'm trying to set a value like this: 

${{quote("{{.domain}}")

That should literally render {{.domain}}, but instead i get an error about an unterminated literal:

 literal not terminated (1:20)\n |  quote(\"{{ .domain \n......^

I've tried a variety of ways to escape it, but nothing seems to work.

krancour:
The `}}` are being misinterpreted as closing the `${{`.

We use [fasttemplate](https://github.com/valyala/fasttemplate) for picking expressions out of strings that might mix static content with expressions.

Some research would be required to better understand how, or even if, fasttemplate can deal with that.

krancour:
After a bit of research:

I don't think the usage of `quote()` is really a factor in this. I think the following would equally trigger the same problem:

```
obj: ${{ {'foo': 'bar', 'nested': {'bat': 'baz'}} }}
```

That is a legitimate expression that evaluates to an object, but the `}}` that closes out the definition of that object would be misinterpreted by fasttemplate as closing out the whole expressions. `{'foo': 'bar', 'nested': {'bat': 'baz'` would then be the partial and invalid expression that would not evaluate correctly.

I cannot find any sensible way around this that doesn't involve giving users a scheme that permits `}}` to explicitly be escaped. If this whole expression were written as follows, there would be no confusion over where the expression ends and a '}' for '\}' substitution can be made before the expression is evaluated.

```
obj: ${{ {'foo': 'bar', 'nested': {'bat': 'baz'\}\} }}
```

The potential trouble I forsee with this is that `\}` could reasonably be used to escape a `}` within a regular expression. So it seems this approach would actually create a whole new bug just waiting for the right expression to come along and expose it.

We could, perhaps, use a less-likely character or combination of characters as an escape sequence.

krancour:
Here's what could work. I was inspired by markdown. In markdown, you typically use triple-backticks as code fences. Sometimes you need to use triple-backticks _within_ a block of code. For such cases, you can use a triple-tilde to start and end the fence instead.

Example:

~~~
```
This whole thing was fenced off with tildes and that's why you see the backticks.
```
~~~

[...]

PR #5473, opened 2025-12-09
feat: support alternative delimiters `${%` for expressions

fykaa:
closes https://github.com/akuity/kargo/issues/5012

this introduces and alternative delimiter `${%` to fix issues where fasttemplate tries to evaluate `}}` as end tag even after using quote().

The choice of delimiter determines which sequences can appear inside the expression:
- If one starts with `${{`, then `}}` cannot appear in the expression, but `${%` can
- If one start with `${%`, then `%}` cannot appear in the expression, but `${{` can

### DIFF
--- a/pkg/expressions/json_templates.go
+++ b/pkg/expressions/json_templates.go
@@ -196,9 +196,9 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
 }
 
 func evaluateTemplate(
-		template string,
-		env map[string]any,
-		exprOpts ...expr.Option,
+	template string,
+	env map[string]any,
+	exprOpts ...expr.Option,
 ) (string, error) {
 	var result strings.Builder
 
--- a/pkg/expressions/json_templates_test.go
+++ b/pkg/expressions/json_templates_test.go
@@ -306,8 +306,8 @@ value13: | # This is a string
 			},
 		},
 		{
-			name: "alternative delimiters allow other end delimiter characters "+
-							"within expression",
+			name: "alternative delimiters allow other end delimiter characters " +
+				"within expression",
 			// This test asserts that choosing an alternative delimiter for an
 			// expression solves the problem of another end delimiter characters
 			// sequence needing to appear within the expression.


pkg/expressions/json_templates.go
```diff
@@ -99,14 +105,29 @@ func evaluateExpressions(collection any, env map[string]any, exprOpts ...expr.Op
...
+// Standard delimiters are ${{ and }}. When expressions need to contain }}
+// (e.g., nested objects or Go template syntax), alternative delimiters $~~ and ~~
+// can be used instead.
 func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Option) (any, error) {
-	if !strings.Contains(template, "${{") {
+	// Determine which delimiter pair to use
+	startTag := "${{"
+	endTag := "}}"
+	hasStandardDelim := strings.Contains(template, "${{")
+	hasAltDelim := strings.Contains(template, "$~~")
```
fykaa: this needs a more intelligent delimiter detection since this doesn't checks if `$~~` is a starting delimeter, working on it.

fykaa:
updating the tests and docs for it rn

fykaa:
I have tested the changes and it is working end to end.

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,166 @@ value13: | # This is a string
...
+			name:         "standard delimiter with missing closing tag",
+			jsonTemplate: `{ "AString": "${{ aString " }`,
+			assertions: func(t *testing.T, _ []byte, err error) {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "Cannot find end tag")
+			},
+		},
+		{
+			name:         "standard delimiter cannot handle }} in expression body",
+			jsonTemplate: `{ "AString": "${{ quote(\"{{.domain}}\") }}" }`,
```
krancour: I would try to leave `quote()` out of the tests if I were you. Despite how the issue was originally reported, you'll see I had quickly concluded that using the `quote()` function wasn't a factor. Knowing that, if we don't involve `quote()`, these tests are all much easier to understand.

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,166 @@ value13: | # This is a string
...
+			jsonTemplate: `{ "AString": "$~~ 'test' ~~ and ${{ aString }}" }`,
+			assertions: func(t *testing.T, jsonOutput []byte, err error) {
+				require.NoError(t, err)
+				parsed := testStruct{}
+				require.NoError(t, json.Unmarshal(jsonOutput, &parsed))
+				// $~~ appears first (index 0), so alternative delimiters are used for entire value
+				// This means ${{ is treated as literal text, not as a delimiter
+				require.Equal(t, "test and ${{ aString }}", parsed.AString)
+			},
+		},
```
krancour: The results you are expecting in these two cases are not what I think anyone would intuitively expect them to be. `${{ expr1 }}-$~~ expr2 ~~` should have _both_ expressions evaluated.

krancour: Looking at this more closely, you probably need to ditch `fasttemplate` in order to make this work. Replacing what `fasttemplate` does in a way that can handle this correctly only takes about 50 lines of code. It may not perform quite as well as `fasttemplate`, but I don't expect the difference to be big enough to matter.

pkg/expressions/json_templates.go
```diff
@@ -99,14 +105,36 @@ func evaluateExpressions(collection any, env map[string]any, exprOpts ...expr.Op
...
+// (e.g., nested objects or Go template syntax), alternative delimiters $~~ and ~~
+// can be used instead. The choice of
+// delimiter determines which sequences can appear inside the expression:
+//   - If you start with ${{, then }} cannot appear in the expression, but $~~ can
+//   - If you start with $~~, then ~~ cannot appear in the expression, but ${{ can
 func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Option) (any, error) {
-	if !strings.Contains(template, "${{") {
+	// Determine which delimiter pair to use by checking which start tag appears first
+	startTag := "${{"
+	endTag := "}}"
```
hiddeco: These should be constants (and used throughout instead of individual string redeclarations).

krancour:
`${% %}` is definitely preferable. In fact, it eliminates a lot of weird edge cases because the end token isn't a subset of the start token. I hadn't actually loved `$~~ ~~` despite it being my own initial suggestion, but I wasn't imaginative enough to come up with something better. Things like `$[[ ]]` or `$(( ))` have the same problems as `${{ }}`. `${% %}` looks like a solid choice.

pkg/expressions/json_templates_test.go
fykaa: oh true, I should have guessed that.

pkg/expressions/json_templates.go
```diff
@@ -182,30 +198,80 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
...
+
+		pos = exprStart + endIdx + len(endDelim)
+	}
+
+	return result.String(), nil
+}
+
+// getExpressionEvaluator returns a function that evaluates input as a single
+// expr-lang expression with the provided map as the environment.
+func getExpressionEvaluator(env map[string]any, exprOpts ...expr.Option) func(string) (string, error) {
```
krancour: There is no longer any reason for this to be a function that returns a function. That bit of indirection was necessary only to produce a function compatible with fasttemplate that closed over `env` and `exprOpts`. That's no longer a concern. This can be refactored to simply be a function that takes `env` and `exprOpts` args directly.

pkg/expressions/json_templates.go
```diff
@@ -182,30 +198,80 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
...
+
+		expression := template[exprStart : exprStart+endIdx]
+		evaluated, err := evaluator(expression)
+		if err != nil {
+			return "", err
+		}
+		result.WriteString(evaluated)
+
+		pos = exprStart + endIdx + len(endDelim)
+	}
```
krancour: This would probably be easier to read and reason about if much of what this loop does were extracted into a helper resembling the following. It would have the added benefit of making it very easy to add new delimiters should we ever desire to.

```
type supportedDelimiterSet struct {
	start string
	end   string
}

var supportedDelimiters = []supportedDelimiterSet{
	{start: stdStartDelim, end: stdEndDelim},
	{start: altStartDelim, end: altEndDelim},
}

// nextExpression finds the next expression in the template string, returning
// the part of the template before the expression, the expression itself, and
// the part after the expression. If no expression is found, the entire template
// is returned as the "after" part, with empty strings for the "before" and
// "expression" parts.
func nextExpression(template string) (string, string, string, error) {
	var delim *supportedDelimiterSet
	delimStartIdx := -1
	for _, d := range supportedDelimiters {
		if i := strings.Index(template, d.start); i >= 0 {
			if delimStartIdx < 0 || i < delimStartIdx {
				delimStartIdx = i
				delim = &d
			}
		}
	}
	if delim == nil {
		return "", "", template, nil
	}
	exprStartIdx := delimStartIdx + len(delim.start)
	exprRelDelimEndIdx := strings.Index(template[exprStartIdx:], delim.end)
	if exprRelDelimEndIdx < 0 {
		return "", "", "", fmt.Errorf(
			"unclosed expression: expected %q but reached end of template",
			delim.end,
		)
	}
	exprEndIdx := exprStartIdx + exprRelDelimEndIdx
	delimEndIdx := exprStartIdx + exprRelDelimEndIdx + len(delim.end)
	return template[:delimStartIdx], // Template part preceding the expression
		template[exprStartIdx:exprEndIdx], // The expression itself
		template[delimEndIdx:], // Template part following the expression
		nil
}
```

krancour: btw... I benchmarked it and it is less performant than what you have now, but negligibly so.

krancour: Also note the rewording of the error. This error is ultimately user-facing, so "not found in tail" wasn't sitting right. It speaks to an implementation detail that the user isn't privy to.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -51,13 +51,32 @@ config:
...
+```
+The above example will be evaluated as follows:
+```yaml
+config:
+  message: Updated image to {{.tag}}
+```
+
+You cannot use `}}` within standard delimiters `${{ }}`, nor can you use `%}`
+within alternative delimiters `${% %}`. Choose the appropriate delimiter for each
+expression based on which characters you need to include.
```
krancour: Well... you _can_ but it will result in Kargo misinterpreting the boundaries of your expressions. I would change "cannot" to "must not" and be more explicit about the consequences.

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -51,13 +51,32 @@ config:
...
+```yaml
+config:
+  # This would fail with standard delimiters due to }} in the string
+  message: ${% "Updated image to {{.tag}}" %}
+```
+The above example will be evaluated as follows:
+```yaml
+config:
+  message: Updated image to {{.tag}}
+```
```
krancour: I don't really like Go templates being cited as the example here when a user is, by a wide margin, more likely to encounter this when creating an expression that either evaluates as a JSON object with some nesting or when creating an expression that evaluates as a string representation of such an object.

Please make the example JSON-based.

pkg/expressions/json_templates.go
```diff
@@ -182,30 +198,80 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
 	return result, nil
 }
 
-// getExpressionEvaluator returns a fasttemplate.TagFunc that evaluates input
-// as a single expr-lang expression with the provided map as the environment.
-func getExpressionEvaluator(env map[string]any, exprOpts ...expr.Option) fasttemplate.TagFunc {
-	return func(out io.Writer, expression string) (int, error) {
+// evaluateMultiDelimiterTemplate evaluates a template that may contain multiple
+// expressions with different delimiter types (${{ }} and ${% %}).
+func evaluateMultiDelimiterTemplate(template string, evaluator func(string) (string, error)) (string, error) {
```
krancour: I think we're putting too much emphasis on the support for multiple delimiters. It isn't as if some variant of this exists that supports only a single delimiter and this distinguishes one from another. This is simply how our templates will work now.

```suggestion
func evaluateTemplate(template string, evaluator func(string) (string, error)) (string, error) {
```

krancour:
You will also need to run `go mod tidy` since this PR removes a direct dependency.

pkg/expressions/json_templates.go
```diff
@@ -98,15 +120,22 @@ func evaluateExpressions(collection any, env map[string]any, exprOpts ...expr.Op
...
+// expressions with different delimiter types.
+//
+// Standard delimiters are ${{ and }}. When expressions need to contain }}
+// (e.g., nested objects or Go template syntax), alternative delimiters ${% and %}
+// can be used instead. Multiple expressions with different delimiters can coexist
+// in the same template string.
 func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Option) (any, error) {
-	if !strings.Contains(template, "${{") {
+	// Check if template contains any expressions
+	if !strings.Contains(template, stdStartDelim) && !strings.Contains(template, altStartDelim) {
```
krancour: Shouldn't this check all `supportedDelimiters` now?

pkg/expressions/json_templates.go
```diff
@@ -182,31 +207,94 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
...
-// as a single expr-lang expression with the provided map as the environment.
-func getExpressionEvaluator(env map[string]any, exprOpts ...expr.Option) fasttemplate.TagFunc {
-	return func(out io.Writer, expression string) (int, error) {
-		program, err := expr.Compile(expression, exprOpts...)
-		if err != nil {
-			return 0, err
-		}
-		result, err := expr.Run(program, env)
+// evaluateTemplate evaluates a template that may contain multiple
+// expressions with different delimiter types (${{ }} and ${% %}).
```
krancour: I would, again, remove emphasis on the support for multiple delimiters. It's a minor facet of what this does. Also, with the layer of indirection that was previously required by fasttemplate being gone now (the function returning a function that closed over certain data), I think the function name `evaluateTemplate()` is adequately self-documenting, and we can just drop this comment altogether.

```suggestion
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,177 @@ value13: | # This is a string
 				require.Equal(t, "42\n", parsed["value13"])
 			},
 		},
+		{
+			name:         "standard delimiter with missing closing tag",
```
krancour: I believe the way the code has now been restructured, there's actually no difference in the paths through the code for "standard" vs "alternative" delimiters. If any set of delimiters works, all the sets of delimiters work. Yuo can drop a number of test cases here.

pkg/expressions/json_templates.go
```diff
@@ -1,17 +1,31 @@
...
+
+type supportedDelimiterSet struct {
+	start string
+	end   string
+}
+
+var supportedDelimiters = []supportedDelimiterSet{
+	{start: stdStartDelim, end: stdEndDelim},
+	{start: altStartDelim, end: altEndDelim},
+}
```
krancour: 1. There's no longer any good reason to have constants for the start/end delimiters.

2. Nit: I would also recommend moving these type defs to just above `nextExpression()`. It's the _only_ place any of this is used. When something is used only in one place, readability is often improved by keeping its definition close to that one place.

```suggestion
type supportedDelimiterSet struct {
	start string
	end   string
}

var supportedDelimiters = []supportedDelimiterSet{
	{start: "${{", end: "}}"},
	{start: "${%", end: "%}"},
}
```

pkg/expressions/json_templates.go
```diff
@@ -23,6 +37,14 @@ import (
...
 // enclosed in quotes.
 //
+// Standard delimiters are ${{ and }}. When expressions need to contain }}
+// (e.g., nested objects or Go template syntax like {{.domain}}), alternative
+// delimiters ${% and %} can be used instead. Both delimiter types can coexist
+// in the same value, and all expressions will be evaluated. For example:
+//   - Standard:    "${{ quote('hello') }}"
+//   - Alternative: "${% '{{.domain}}' %}" (when }} appears in the expression)
+//   - Mixed:       "${{ ctx.stage }} deployed ${% '{{.tag}}' %}"
+//
```
krancour: I would, again, not dive so deep into this in internally-facing comments, especially when the set of supported delimiters is now much more flexible. I'd not want to update all these comments if another delimiter were added -- and that's if I even remembered to.

Even existing comments mentioning the original delimiters could stand to be genericized to just speak of "supported delimiters."

pkg/expressions/json_templates.go
```diff
@@ -1,19 +1,26 @@
...
+var supportedDelimiters = []supportedDelimiterSet{
+	{start: "${{", end: "}}"},
+	{start: "${%", end: "%}"},
+}
+
 // EvaluateJSONTemplate evaluates a JSON byte slice, which is presumed to be a
-// template containing expr-lang expressions offset by ${{ and }}, using the
+// template containing expr-lang expressions offset by supported delimiters, using the
 // provided environment as context. The evaluated JSON is returned as a new byte
 // slice, ready for unmarshaling.
```
krancour:
```suggestion
// template containing expr-lang expressions, offset by supported delimiters,
// using the provided environment as context. Multiple (different) supported
// delimiters can be used within a single template. The evaluated JSON is
// returned as a new byte slice, ready for unmarshaling.
```

pkg/expressions/json_templates.go
```diff
@@ -23,15 +30,21 @@ import (
 // Since the template itself must be valid JSON, all expressions MUST be
 // enclosed in quotes.
 //
+// Multiple delimiter types can coexist in the same value, and all expressions will
+// be evaluated. For example:
+//   - "${{ quote('hello') }}"
+//   - "${% '{{.domain}}' %}"
+//   - "${{ ctx.stage }} deployed ${% '{{.tag}}' %}"
+//
```
krancour:
```suggestion
```

pkg/expressions/json_templates.go
```diff
@@ -23,15 +30,21 @@ import (
...
 // non-string JSON, it will be treated as such. This ensures the possibility of
 // expressions being used to construct any valid JSON value, despite the fact
 // that expressions must, themselves, be contained within a string value. This
 // does mean that for expressions which may evaluate as something resembling a
 // valid non-string JSON value, the user must take care to ensure that the
-// expression evaluates to a string enclosed in quotes. e.g. ${{ true }} will
-// evaluated as a bool, but ${{ quote(true) }} will be evaluated as a string.
+// expression evaluates to a string enclosed in quotes. For example, an expression
+// evaluating to true will result in a bool, but quote(true) will result in a string.
 // This behavior should be intuitive to anyone familiar with YAML.
```
krancour:
```suggestion
// expression evaluates to a string enclosed in quotes. For example, an
// expression evaluating to true will result in a bool, but quote(true) will
// result in a string.
// This behavior should be intuitive to anyone familiar with YAML.
```

pkg/expressions/json_templates.go
```diff
@@ -97,16 +110,24 @@ func evaluateExpressions(collection any, env map[string]any, exprOpts ...expr.Op
 }
 
 // EvaluateTemplate evaluates a single template string with the provided
-// environment. Note that a single template string can contain multiple
-// expressions.
+// environment. A single template string can contain multiple
+// expressions with different supported delimiter types.
```
krancour:
```suggestion
// environment. A single template string can contain multiple expressions offset
// by any of the supported delimiters.
```

pkg/expressions/json_templates.go
```diff
@@ -182,31 +199,92 @@ func EvaluateTemplate(template string, env map[string]any, exprOpts ...expr.Opti
...
-// getExpressionEvaluator returns a fasttemplate.TagFunc that evaluates input
-// as a single expr-lang expression with the provided map as the environment.
-func getExpressionEvaluator(env map[string]any, exprOpts ...expr.Option) fasttemplate.TagFunc {
-	return func(out io.Writer, expression string) (int, error) {
-		program, err := expr.Compile(expression, exprOpts...)
-		if err != nil {
-			return 0, err
-		}
-		result, err := expr.Run(program, env)
+func evaluateTemplate(template string, env map[string]any, exprOpts ...expr.Option) (string, error) {
```
krancour:
```suggestion
func evaluateTemplate(
		template string,
		env map[string]any,
		exprOpts ...expr.Option,
) (string, error) {
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+			name:         "missing closing tag",
+			jsonTemplate: `{ "AString": "${{ aString " }`,
+			assertions: func(t *testing.T, _ []byte, err error) {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "unclosed expression")
+			},
+		},
+		{
+			name:         "closing delimiter in expression body terminates early",
+			jsonTemplate: `{ "AString": "${{ \"{{.domain}}\" }}" }`,
```
krancour:
```suggestion
			jsonTemplate: `{ "AString": "${{ '{{.domain}}' }}" }`,
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+			assertions: func(t *testing.T, _ []byte, err error) {
+				// This fails because }} appears in the expression body
+				// The parser finds the first }} and terminates, leaving unterminated string
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "literal not terminated")
+			},
+		},
+		{
+			name:         "different delimiter type allows problematic characters",
+			jsonTemplate: `{ "AString": "${% \"{{.domain}}\" %}" }`,
```
krancour:
```suggestion
			jsonTemplate: `{ "AString": "${% '{{.domain}}' %}" }`,
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+			assertions: func(t *testing.T, jsonOutput []byte, err error) {
+				require.NoError(t, err)
+				parsed := map[string]any{}
+				require.NoError(t, json.Unmarshal(jsonOutput, &parsed))
+				require.Equal(t, "{{.domain}}", parsed["AString"])
+			},
+		},
+		{
+			name:         "delimiter starts can appear as literals in expressions",
+			jsonTemplate: `{ "AString": "${{ \"This string contains literal ${% inside it\" }}" }`,
```
krancour:
```suggestion
			jsonTemplate: `{ "AString": "${{ 'This string contains literal ${% inside it' }}" }`,
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+			assertions: func(t *testing.T, jsonOutput []byte, err error) {
+				require.NoError(t, err)
+				parsed := testStruct{}
+				require.NoError(t, json.Unmarshal(jsonOutput, &parsed))
+				require.Equal(t, "This string contains literal ${% inside it", parsed.AString)
+			},
+		},
+		{
+			name:         "both delimiter types can coexist in same value",
```
krancour:
```suggestion
			name:         "multiple delimiters can coexist in the same template",
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+			jsonTemplate: `{ "AString": "foo ${{ aString }} and ${% \"bar\" %}" }`,
```
krancour:
```suggestion
			jsonTemplate: `{ "AString": "foo ${{ aString }} and ${% 'bar' %}" }`,
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+				require.NoError(t, err)
+				parsed := testStruct{}
+				require.NoError(t, json.Unmarshal(jsonOutput, &parsed))
+				// Both expressions are evaluated
+				require.Equal(t, "foo hello and bar", parsed.AString)
+			},
+		},
+		{
+			name:         "nested delimiter start inside expression - final closer is literal",
+			jsonTemplate: `{ "AString": "${{ \"text ${%\" }}%}" }`,
```
krancour:
```suggestion
			jsonTemplate: `{ "AString": "${{ 'text ${%' }}%}" }`,
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+				require.NoError(t, err)
+				parsed := testStruct{}
+				require.NoError(t, json.Unmarshal(jsonOutput, &parsed))
+				// ${{ expr }} is evaluated, then "%}" remains as literal text
+				require.Equal(t, "text ${%%}", parsed.AString)
+			},
+		},
+		{
+			name:         "delimiter wrapping another delimiter expression",
+			jsonTemplate: `{ "AString": "${% \"outer\" %}${{ aString }}%}" }`,
```
krancour:
```suggestion
			jsonTemplate: `{ "AString": "${% 'outer' %}${{ aString }}%}" }`,
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
 				require.Equal(t, "42\n", parsed["value13"])
 			},
 		},
+		{
+			name:         "missing closing tag",
```
krancour:
```suggestion
			name:         "missing closing delimiter",
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "unclosed expression")
+			},
+		},
+		{
+			name:         "closing delimiter in expression body terminates early",
+			jsonTemplate: `{ "AString": "${{ '{{.domain}}' }}" }`,
+			assertions: func(t *testing.T, _ []byte, err error) {
+				// This fails because }} appears in the expression body
+				// The parser finds the first }} and terminates, leaving unterminated string
```
krancour:
```suggestion
			name: "closing delimiter in expression body terminates early",
			// This test is a control to validate it is a problem when the same
			// character sequence as the closing delimiter is meant to appear in
			// the body of an expression. Such a sequence should be misinterpreted
			// as terminating the expression, which having been cut short, will lead
			// to an error (or at least an unanticipated result) when evaluated.
			jsonTemplate: `{ "AString": "${{ '{{.domain}}' }}" }`,
			assertions: func(t *testing.T, _ []byte, err error) {
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "literal not terminated")
+			},
+		},
+		{
+			name:         "different delimiter type allows problematic characters",
+			jsonTemplate: `{ "AString": "${% '{{.domain}}' %}" }`,
+			assertions: func(t *testing.T, jsonOutput []byte, err error) {
```
krancour:
```suggestion
			name: "alternative delimiters allow other end delimiter characters "+
							"within expression",
			// This test asserts that choosing an alternative delimiter for an
			// expression solves the problem of another end delimiter characters
			// sequence needing to appear within the expression.
			jsonTemplate: `{ "AString": "${% '{{.domain}}' %}" }`,
			assertions: func(t *testing.T, jsonOutput []byte, err error) {
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+		{
+			name:         "delimiter starts can appear as literals in expressions",
+			jsonTemplate: `{ "AString": "${{ 'This string contains literal ${% inside it' }}" }`,
+			assertions: func(t *testing.T, jsonOutput []byte, err error) {
+				require.NoError(t, err)
+				parsed := testStruct{}
+				require.NoError(t, json.Unmarshal(jsonOutput, &parsed))
+				require.Equal(t, "This string contains literal ${% inside it", parsed.AString)
+			},
+		},
```
krancour: I'd remove this case because I don't think this was ever in doubt, but including it here may inadvertently plant the seen in one's mind that nested expressions are permitted, and they're not. i.e. I think this deserves a "better off left alone" approach.

```suggestion
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+			name:         "nested delimiter start inside expression - final closer is literal",
+			jsonTemplate: `{ "AString": "${{ 'text ${%' }}%}" }`,
+			assertions: func(t *testing.T, jsonOutput []byte, err error) {
+				require.NoError(t, err)
+				parsed := testStruct{}
+				require.NoError(t, json.Unmarshal(jsonOutput, &parsed))
+				// ${{ expr }} is evaluated, then "%}" remains as literal text
+				require.Equal(t, "text ${%%}", parsed.AString)
+			},
+		},
```
krancour: This appears to test the same thing as the [other](https://github.com/akuity/kargo/pull/5473/files#r2620359471) test cases I suggested removing. This should go, too.

```suggestion
```

pkg/expressions/json_templates_test.go
```diff
@@ -284,6 +284,77 @@ value13: | # This is a string
...
+			name:         "delimiter wrapping another delimiter expression",
+			jsonTemplate: `{ "AString": "${% 'outer' %}${{ aString }}%}" }`,
+			assertions: func(t *testing.T, jsonOutput []byte, err error) {
+				require.NoError(t, err)
+				parsed := testStruct{}
+				require.NoError(t, json.Unmarshal(jsonOutput, &parsed))
+				// ${% expr %} is evaluated first, then ${{ expr }}, then "%}" is literal
+				require.Equal(t, "outerhello%}", parsed.AString)
+			},
+		},
```
krancour: This test does not seem to match its description. Also, as noted [here](https://github.com/akuity/kargo/pull/5473/files#r2620359471), I don't want to give the false impression that nested expressions are supported. Cut this.

```suggestion
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -51,13 +51,38 @@ config:
   message: ${{ "Hello, world!" }}
 ```
 
-The above example will be evaluated as the following:
+The above example will be evaluated as follows:
```
krancour: Super nit and esoteric English thing here. As originally written, it meant "the following is the result of the evaluation."

Written as proposed, the meaning is a bit different. It sounds very much like what follows is _how_ it's evaluated; rather than _what_ it evaluates to. i.e. It now sounds to be introducing a _process_.

```suggestion
The above example will be evaluated as the following:
```

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -51,13 +51,38 @@ config:
...
+  jsonData: '{"metadata": {"labels": {"app": "nginx"}}}'
+```
+
+You must not use `}}` within standard delimiters `${{ }}`, nor `%}` within
+alternative delimiters `${% %}`, as this will cause Kargo to misinterpret the
+expression boundaries and terminate the expression prematurely. Choose the
+appropriate delimiter for each expression based on which characters you need to
+include.
+:::
+
```
~~~suggestion
:::info What if?
If an expression must contain the literal character sequence `}}`, that sequence
can be mistaken as marking the end of the expression, cutting the expression
short, and resulting in an error or, at least, unexpected results.
Inconveniently, the `}}` character sequence is a common one occurring in the
definition of JSON objects or string representations of such an objects.

The following would fail, because the expression would be misidentified as
`{"labels": {"app": "nginx"`, which is not valid JSON:

```yaml
config:
  jsonObject: ${{ {"labels": {"app": "nginx"}} }}
```

As a workaround for the above, a secondary set of delimiters, `${%` and `%}`,
are supported as well. Opt for these delimiters when an expression must contain
the literal character sequence `}}`. Similarly, always opt for `${{` and `}}`
as delimiters when an expression must contain the literal character sequence
`%}`.

```yaml
config:
  jsonData: %{{ {"labels": {"app": "nginx"}} }}
```
:::

~~~

krancour (approved):

PR #5478, opened 2025-12-09
feat(helm): allow overriding parse-cabundle init container securityCo…

sstarcher:
…ntext

Add securityContext configuration option for the parse-cabundle init container under api.cabundle and controller.cabundle. This allows users in restricted environments (PodSecurity Standards, Kyverno policies) to override the default root security context.

The default behavior (runAsUser: 0) is preserved for backward compatibility.


hiddeco:
Thanks for submitting this PR.

You need to run `make codegen` and sign-off your commit to ensure all checks pass.

sstarcher:
Verified I can't set any of these restrictions without breaking cabundle setup.

Issue #5484, opened 2025-12-11
fix: port conflict bw development cluster and quickstart


fykaa:
the quickstart script and development cluster configurations use overlapping port mappings, which causes confusion when developers switch between quickstart and development workflows.

Current Port Mappings:

quickstart clusters (`hack/quickstart/kind.sh`):
- 30081 → test application instance
- 30082 → uat application instance
- 30083 → prod application instance

development clusters (`hack/kind/cluster.yaml` and `hack/k3d/cluster.yaml`):
- 30080 → ArgoCD
- 30081 → Kargo API server
- 30082 → Kargo API server proxy (UI)

so when someone creates a quickstart cluster (`kargo-quickstart`) first, then tries to use Tilt for development without switching clusters, accesses `localhost:30082` expecting the Kargo UI per the [contributor guide](https://github.com/akuity/kargo/blob/main/docs/docs/60-contributor-guide/10-hacking-on-kargo.md?plain=1#L375) and gets the UAT application instead

fykaa:
I have [proposed a quick solution here](https://github.com/akuity/kargo/pull/5485) that is to change the quickstart script to use non-overlapping ports for application instances:
- 30084 → test application instance
- 30085 → UAT application instance  
- 30086 → prod application instance

this keeps the 30080-30082 range reserved for development cluster infrastructure (argocd, kargo api, kargo ui)

hiddeco:
Any reason to not just move the application related ports to a different range? E.g. `32080-32083`?

PR #5485, opened 2025-12-11
chore: port conflict for development clusters

fykaa:
closes https://github.com/akuity/kargo/issues/5484

### DIFF
--- a/docs/docs/20-quickstart/index.md
+++ b/docs/docs/20-quickstart/index.md
@@ -735,8 +735,8 @@ We leave it as an exercise to the reader to use the dashboard to progress the
 :::info
 The `uat` and `prod` instances of our site should be accessible at:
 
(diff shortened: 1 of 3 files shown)


fykaa:
if approved, updates should be reflected here: https://raw.githubusercontent.com/akuity/kargo/main/hack/quickstart/kind.sh and https://raw.githubusercontent.com/akuity/kargo/main/hack/quickstart/k3d.sh

krancour:
I'm ok with this in principle, but agree with @hiddeco's suggestion [here](https://github.com/akuity/kargo/issues/5484#issuecomment-3643249022) to start using a much different range for the app ports.

If we put more space between the ports used for "system" things like Argo CD, Kargo, etc. and those used for the demo application workloads, we will eliminate the need to shuffle things around further in the future if we need additional ports for system things.

PR #5487, opened 2025-12-11
chore(docs): admonitions formatting and titles

hiddeco:
1. Add newlines in [admonitions](https://docusaurus.io/docs/next/markdown-features/admonitions).
2. Swap `__` "titles" for real admonition titles.

### DIFF
--- a/docs/api/docs.templ
+++ b/docs/api/docs.templ
@@ -4,7 +4,7 @@
 {{- range .Files}}
 {{$file_name := .Name}}
 <a name="{{.Name | anchor}}"></a>
-<p align="right"><a href="#top">Top</a></p>
+<p class="text--right"><a href="#top">Top</a></p>
 
 ## {{ .Name | trimPrefix "api/" | trimSuffix "/service.proto" | trimSuffix "/generated.proto" }}
 
--- a/docs/docs/90-api-documentation.md
+++ b/docs/docs/90-api-documentation.md
@@ -2,7 +2,7 @@
 <a name="top"></a>
 
 <a name="api_service_v1alpha1_service-proto"></a>
-<p align="right"><a href="#top">Top</a></p>
+<p class="text--right"><a href="#top">Top</a></p>
 
 ## service/v1alpha1
 <a name="akuity-io-kargo-service-v1alpha1-KargoService"></a>
@@ -1762,7 +1762,7 @@ RawFormat specifies the format for raw resource representation.
  <!-- end enums -->
 
 <a name="api_rbac_v1alpha1_generated-proto"></a>
-<p align="right"><a href="#top">Top</a></p>
+<p class="text--right"><a href="#top">Top</a></p>
 
 ## rbac/v1alpha1
 <a name="github-com-akuity-kargo-api-rbac-v1alpha1-Claim"></a>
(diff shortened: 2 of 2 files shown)


docs/docs/40-operator-guide/10-basic-installation.md
```diff
@@ -8,8 +8,7 @@ description: Learn how to do a basic installation of Kargo using Helm
 Installing Kargo's __cluster-side components__ (e.g. controllers and API server)
 with __default configuration__ is quick and easy.
 
-:::info
-__Not what you were looking for?__
+:::info Not what you were looking for?
```
krancour: I was today years old when I found out you can do this. I really like it.

docs/docs/40-operator-guide/40-security/25-service-accounts.md
```diff
@@ -8,14 +8,15 @@ Kargo `ServiceAccount`s enable programmatic access to Kargo APIs without
 requiring a user to share short-lived credentials with non-human agents, such as
 a CI process.
 
-:::note
-__Not what you were looking for?__
+
+:::note Not what you were looking for?
```
krancour: Not a _new_ problem, but definitely one that can be fixed while you're already touching these -- some of the "Not what you were looking for?" admonitions are `note` and some are `info`. We should probably be consistent by picking just one.

Semantically speaking, I like `note` when it's more information that the user really _should_ pay attention to and `info` when it's more information that's very much supplemental and can be skipped.

Oddly, of those two, I find `info` to be the more visually prominent one. imho, on account of that we should use settle on `info` for "Not what you were looking for?" throughout.

We _could_ make a custom `<nwywlf>` (half-joking about the name) tag to keep these consistent throughout. It would also keep the wording of the title consistent. This is a nice to have.

In the end, I don't feel strongly about what we do, other than believing that whatever we do should be maximally consistent.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -797,13 +829,15 @@ The remainder of this section focuses on configuring webhook receivers at the
 Project level.
 
 :::info
+
 **Not what you were looking for?**
```
krancour: I think you meant to move this up into the title.

hiddeco: Ah, this missed my searches because it was the only one with `**`.

docs/docs/40-operator-guide/40-security/25-service-accounts.md
hiddeco: I think with the custom title, we can just (ab)use `info` for visibility.

docs/api/docs.templ
```diff
@@ -4,15 +4,17 @@
 {{- range .Files}}
 {{$file_name := .Name}}
 <a name="{{.Name | anchor}}"></a>
-<p align="right"><a href="#top">Top</a></p>
+<p class="text--right"><a href="#top">Top</a></p>
```
hiddeco, line 7: Does anyone know why this exists?

hiddeco: Ah, it's to navigate to the top of the page at the beginning/end of a section. But it feels like this is kind of easier to do through the ToC on the right-hand side?

PR #5506, opened 2025-12-15
chore(ui): adjust shortVersion function length check

Demivan:
`shortVersion` adds `...` in the middle, making some strings longer.
`15.12.25-1030` Docker tag, for example, becomes `15.12....5-1030`.
This change makes it so `shortVersion` only applies when it actually shortens the string.

### DIFF
--- a/ui/src/features/project/pipelines/freight/short-version-utils.ts
+++ b/ui/src/features/project/pipelines/freight/short-version-utils.ts
@@ -1,5 +1,5 @@
 export const shortVersion = (version: string = '', length = 12) => {
-  if (version.length <= length) {
+  if (version.length <= length + 3) {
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
Good catch, thanks! 🙇

ui/src/features/project/pipelines/freight/short-version-utils.ts
```diff
@@ -1,5 +1,5 @@
 export const shortVersion = (version: string = '', length = 12) => {
-  if (version.length <= length) {
+  if (version.length <= length + 2) {
```
hiddeco:
```suggestion
  if (version.length <= length + 3) {
```

Issue #5492, opened 2025-12-12
`set-freight-alias` step


jessesuen:
# Proposed Feature

It's come up a few times where users would like to update the Freight alias with something more meaningful to their org. To help with this, Kargo could introduce a `set-freight-alias` step where users could set the alias to a value of their choosing, say, during a pre-processing Stage.

# Motivation

Updating to a more useful freight alias would help with visibility.

nitishfy:
I'm happy to work on this!

krancour:
Discussed offline. Please write the schema first and get some buy-in from @hiddeco and myself before moving ahead to implementing the corresponding logic.

krancour:
@jessesuen if you can weigh in on how much of a priority this is, it will help.

jessesuen:
This would be nice to have for v1.9, but best effort, since I don't want to hold up the release

krancour:
I'm afraid there's no way this will squeeze into 1.9. I was just trying to gauge whether this was a responsible use of any bandwidth @nitishfy has. If you'd have regarded it as nice to have for 1.9, I'm happy to drop this in the 1.10 milestone.

@nitishfy, have at it.

PR #5516, opened 2025-12-17
feat: add `set-freight-alias` promotion step

nitishfy:
Closes https://github.com/akuity/kargo/issues/5492

### DIFF
--- a/pkg/promotion/runner/builtin/set_freight_alias.go
+++ b/pkg/promotion/runner/builtin/set_freight_alias.go
@@ -92,7 +92,7 @@ func (s *freightAliasSetter) run(
 		return promotion.StepResult{
 				Status: kargoapi.PromotionStepStatusFailed,
 			}, &promotion.TerminalError{
-				Err: fmt.Errorf("Freight %q not found in project %q", cfg.Name, stepCtx.Project), //nolint:revive
+				Err: fmt.Errorf("freight %q not found in project %q", cfg.Name, stepCtx.Project),
 			}
 	}
 
--- a/pkg/promotion/runner/builtin/set_freight_alias_test.go
+++ b/pkg/promotion/runner/builtin/set_freight_alias_test.go
@@ -1,7 +1,6 @@
 package builtin
 
 import (
-	"context"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -19,57 +18,57 @@ import (
 func Test_setFreightAlias_convert(t *testing.T) {
 	tests := []validationTestCase{
 		{
-			name:   "both freightName and newAlias are not specified",
+			name:   "both name and alias are not specified",
 			config: promotion.Config{},
 			expectedProblems: []string{
-				"(root): freightName is required",
-				"(root): newAlias is required",
+				"(root): name is required",
+				"(root): alias is required",
 			},
 		},
 		{
-			name: "freightName is not specified",
+			name: "name is not specified",
 			config: promotion.Config{
-				"newAlias": "stable",
+				"alias": "stable",
 			},
 			expectedProblems: []string{
-				"(root): freightName is required",
+				"(root): name is required",
 			},
 		},
 		{
-			name: "freightName is empty",
+			name: "name is empty",
 			config: promotion.Config{
-				"freightName": "",
-				"newAlias":    "new-alias",
+				"name":  "",
+				"alias": "new-alias",
 			},
 			expectedProblems: []string{
-				"freightName: String length must be greater than or equal to 1",
+				"name: String length must be greater than or equal to 1",
 			},
 		},
 		{
-			name: "newAlias is not specified",
+			name: "alias is not specified",
 			config: promotion.Config{
-				"freightName": "fake-freight-id",
+				"name": "fake-freight-id",
 			},
 			expectedProblems: []string{
-				"(root): newAlias is required",
+				"(root): alias is required",
 			},
 		},
 		{
-			name: "newAlias is empty",
+			name: "alias is empty",
 			config: promotion.Config{
-				"freightName": "fake-freight-id",
-				"newAlias":    "",
+				"name":  "fake-freight-id",
+				"alias": "",
 			},
 			expectedProblems: []string{
-				"newAlias: String length must be greater than or equal to 1",
+				"alias: String length must be greater than or equal to 1",
 			},
 		},
 		{
 			name: "unknown field is not allowed",
 			config: promotion.Config{
-				"freightName": "fake-freight-id",
-				"newAlias":    "new-alias",
-				"unexpected":  "nope",
+				"name":       "fake-freight-id",
+				"alias":      "new-alias",
+				"unexpected": "nope",
 			},
 			expectedProblems: []string{
 				"(root): Additional property unexpected is not allowed",
@@ -78,8 +77,8 @@ func Test_setFreightAlias_convert(t *testing.T) {
 		{
 			name: "valid minimal config",
 			config: promotion.Config{
-				"freightName": "fake-freight-id",
-				"newAlias":    "new-alias",
+				"name":  "fake-freight-id",
+				"alias": "new-alias",
 			},
 		},
 	}
@@ -100,7 +99,7 @@ func Test_setFreightAlias_run(t *testing.T) {
 		otherFreight = "freight-id-2"
 		testProject  = "test-project"
 		oldAlias     = "old-alias"
-		newAlias     = "new-alias"
+		alias        = "new-alias"
 	)
 
 	tests := []struct {
@@ -114,7 +113,7 @@ func Test_setFreightAlias_run(t *testing.T) {
 			client: fake.NewClientBuilder().WithScheme(scheme).Build(),
 			cfg: builtin.SetFreightAliasConfig{
 				Name:  testFreight,
-				Alias: newAlias,
+				Alias: alias,
 			},
 			assertions: func(t *testing.T, res promotion.StepResult, _ client.Client, err error) {
 				require.ErrorContains(t, err, "not found")
@@ -139,7 +138,7 @@ func Test_setFreightAlias_run(t *testing.T) {
 				Build(),
 			cfg: builtin.SetFreightAliasConfig{
 				Name:  testFreight,
-				Alias: newAlias,
+				Alias: alias,
 			},
 			assertions: func(t *testing.T, res promotion.StepResult, _ client.Client, err error) {
 				require.ErrorContains(t, err, "failed to patch alias")
@@ -166,7 +165,7 @@ func Test_setFreightAlias_run(t *testing.T) {
 				Build(),
 			cfg: builtin.SetFreightAliasConfig{
 				Name:  testFreight,
-				Alias: newAlias,
+				Alias: alias,
 			},
 			assertions: func(t *testing.T, res promotion.StepResult, c client.Client, err error) {
 				require.NoError(t, err)
@@ -183,8 +182,8 @@ func Test_setFreightAlias_run(t *testing.T) {
 				)
 				require.NoError(t, err)
 
-				require.Equal(t, newAlias, freight.Alias)
-				require.Equal(t, newAlias, freight.Labels[kargoapi.LabelKeyAlias])
+				require.Equal(t, alias, freight.Alias)
+				require.Equal(t, alias, freight.Labels[kargoapi.LabelKeyAlias])
 			},
 		},
 	}


charts/kargo/templates/controller/cluster-roles.yaml
```diff
@@ -44,13 +44,23 @@ rules:
...
+  - kargo.akuity.io
+  resources:
+  - freights
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - patch
+  - update
```
nitishfy, line 63: The controller requires permissions to patch the freight resource once the alias has been updated in order to reflect the updated value to the freight label.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,51 @@
...
+previous deployment in your project. Its alias is `staging-old`, and you want to
+rename it to `archived` to reflect its status:
+
+```yaml
+steps:
+- uses: set-freight-alias
+  config:
+    freightID: "freight-id-456"
+    newAlias: "archived
+```
```
hiddeco: I find this example odd, as I think the most common case is to rename Freight that is actively being promoted — and I do not think it's clear from this example how one would obtain the Freight ID.

The reason you have to supply the Freight ID rather than its current alias is that that's the single immutable identifier, and it prevents unintentional side-effects if other Stages also perform mutations on the Freight alias.

I.e., if `dev` renames Freight to `now-in-dev` and `staging` would mutate Freight to `now-in-staging` — you could be mutating the wrong Freight _without_ an ID selector because the Freight in `dev` may not be what you're actively promoting to `staging`.

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,151 @@
...
+) (promotion.StepResult, error) {
+	// This step intentionally targets a Freight specified explicitly via
+	// `freightID` instead of implicitly operating on
+	// `stepCtx.TargetFreightRef`.
+	//
+	// While most promotion steps act on the Freight currently being promoted,
+	// updating a Freight alias is a project-scoped mutation rather than a
+	// promotion-scoped one. There are valid use cases where a user may want to
+	// update the alias of a different Freight in the same project that is not
+	// currently being promoted
```
hiddeco: This comment explains why we require an explicit selector (rather than implicit `stepCtx.TargetFreightRef`), but that's not the interesting design decision, as you'd need an explicit selector regardless of whether you wanted to allow targeting other Freight.

What actually needs explaining is why we require an ID (or name) specifically, rather than accepting an alias as an alternative. The reason is that aliases are mutable: if multiple Stages mutate aliases concurrently, an alias selector could resolve to the wrong Freight. The ID is the only immutable identifier.

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+	}
+
+	if len(freightList.Items) > 1 || (len(freightList.Items) == 1 && freightList.Items[0].Name != freight.Name) {
+		return promotion.StepResult{
+				Status: kargoapi.PromotionStepStatusFailed,
+			}, &promotion.TerminalError{
+				Err: fmt.Errorf("alias %q is already in use by another Freight in project %q", cfg.NewAlias, stepCtx.Project),
+			}
+	}
```
hiddeco: This shouldn't be required, as it would be captured by the validation webhook. Additionally, it's vulnerable to TOCTOU races: the alias could be claimed between the `List` and the `Patch`.

As the webhook enforces uniqueness, just let the `Patch` fail and surface that error.

pkg/promotion/runner/builtin/schemas/set-freight-alias-config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "SetFreightAliasConfig",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["freightID","newAlias"],
+  "properties": {
+    "freightID": {
```
hiddeco: I forgot we internally call this "Freight name", even when it looks like an "ID". We should probably rename this to `freightName`.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
krancour: I agree with @hiddeco's comments about how _not_ to demonstrate this, but demonstrating using aliases like "now-in-staging" might promote some pretty problematic usage as well. If you go back to @jessesuen's original issue, I think the use case he presented seems realistic and not like an abuse of the feature. I would lean into that.

nitishfy: Please take a look now.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,51 @@
...
+:::
+
+## Examples
+
+### Common Usage
+
+In many organizations, Freight identifiers are opaque and not meaningful to humans
+(e.g. commit SHAs or generated IDs). Teams often want to assign human-friendly,
+organization-specific aliases to Freights to make them easier to reason about
+across Stages and environments.
```
krancour: None of this is accurate. Neither name nor alias are ever a commit SHA and the opaqueness of names and aliases is not a "many organizations" thing. It's universally true in Kargo that Freight names are hashes of a canonical representation of the Freight, making them, essentially a fingerprint. Aliases are initially random.

Separate from that, "Freights" is not a word. Weird English thing: Freight is a [mass noun](https://en.wikipedia.org/wiki/Mass_noun). "Freight" can be singular or plural depending on context and sometimes you may need to add additional words to make context more clear, such as "a single Freight resource." (The word "Freights" may appear some places in code or config if it's a case of some technology we use not understanding these esoteric pluralization rules, but anywhere you might see "Freights" in docs, it is grammatically incorrect and should be fixed.)

Please also monospace resource kinds.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Examples
+
+### Common Usage
+
+In Kargo, the name of a `Freight` resource is a hash derived from a canonical
+representation of that `Freight`, making it effectively a stable fingerprint.
+While this ensures immutability and uniqueness, these names are not intended to
+be human-friendly. Additionally, `Freight` aliases are initially assigned random
+values.
```
krancour: I would wordsmith this further. "not _intended_ to be human-friendly" makes it sound as if we "inflicted" this on users, when it's actually just a natural consequence of a sensible design. I would just say it's "not human-friendly" without speaking to any intentions.

> Additionally, `Freight` aliases are initially assigned random
values.

As for the above, it feels like this isn't adequately connected to the preceding thought. We should be saying that human-friendly aliases are assigned _because_ Freight names are not human friendly, but they are random and sometimes users might like to modify those aliases to have semantic meaning.

nitishfy: please take a look now.

snuggie12:
If I understand this correctly this does not enable getting the alias as well? There's two issues requesting that and I was hoping the set solution would also fix that.

nitishfy:
> If I understand this correctly this does not enable getting the alias as well?

No. You can simply fetch the actual freight name using `${{ ctx.targetFreight.name }}` and reference it in the `set-freight-alias` step that will automatically set that freight's alias to new alias.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
+---
+sidebar_label: set-freight-alias
+description: Updates the alias of a `Freight` resource in a project.
```
krancour: Backticks render literally instead here. "In a project" seems awkward.

Image

```suggestion
description: Updates the alias of a Freight resource.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
...
+---
+
+# `set-freight-alias`
+
+`set-freight-alias` updates the alias of a `Freight` resource in a project.
```
krancour:
```suggestion
`set-freight-alias` updates the alias of a `Freight` resource.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Configuration
+
+| Name          | Type     | Required | Description                                                 |
+|---------------|----------|----------|-------------------------------------------------------------|
+| `freightName` | `string` | Y | The immutable name (or ID) of a `Freight` resource to update. |
```
krancour: I know what you were trying to say here. We, too often, use Freight "name" and "ID" interchangeably. Really, there is no ID. It's _just_ name. I get that you were trying to make a concession to how we often talk about Freight, but this also opens the door to misinterpretation. Anyone who _isn't_ aware that "Freight ID" is a colloquial, but inaccurate way of saying "Freight name," can be mislead into thinking the parenthetical is referencing a _second_ way of identifying which Freight resource to update. I'd just leave that parenthetical out.

__Edit:__ `freightID` does appear at places in the code, but it's mostly an invisible implementation detail. It's not user-facing language.

```suggestion
| `freightName` | `string` | Y | The name of a `Freight` resource to update. |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
...
+| `newAlias`    | `string` | Y | The desired new alias to set on the `Freight`.                |
+
+:::note
+
+Alias uniqueness is enforced by the Kargo API. If the requested alias is already
+in use by another `Freight` in the project, this step will fail.
```
krancour: Just a reminder that "freight" is a [mass noun](https://en.wikipedia.org/wiki/Mass_noun).

The word is basically plural by default, so "a Freight" never makes sense grammatically. It needs additional language to make it singular. "A piece of Freight," or the one we use more commonly, "a Freight resource."

It's a common mistake, so nbd, but file it in the 🧠 because the more of us who are aware of the idiosyncrasy and on the lookout for it, the better.

```suggestion
Alias uniqueness is enforced by the Kargo API. If the requested alias is already
in use by another `Freight` resource in the project, this step will fail.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
...
+immutability and uniqueness, but the resulting names are not human-friendly.
+To improve usability, Kargo assigns each `Freight` a human-friendly alias, which
+is initially generated with a random value.
+
+In many cases, teams want to replace this randomly assigned alias with one that
+has clear semantic meaning within their organization. Assigning an
+organization-specific alias can make a `Freight` easier to reason about across
+Stages and environments, improving clarity in promotion pipelines, dashboards,
+and operational workflows—without sacrificing the safety guarantees provided by
+immutable identifiers.
```
krancour: I don't think this level of detail belongs in the examples.

It would be more appropriate to lift this up to the page's introductory text.

I might also suggest reusing more accurate wording from here:

https://docs.kargo.io/user-guide/how-to-guides/working-with-freight#names

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
...
+
+For example, a team might want to label a `Freight` with an alias that reflects its
+intended purpose or lifecycle state within the organization:
+
+```yaml
+steps:
+- uses: set-freight-alias
+  config:
+    freightName: ${{ ctx.targetFreight.name }}
+    newAlias: "candidate-for-staging"
```
krancour: I would strongly urge against using this as an example. 95% of Kargo is about controlling and observing what is where. Using aliases for this purpose is unnecessary. It also feels dirty because it couples the name by which we know a particular set of artifacts to _where it is_ in the pipeline instead of _what it is._ Imagine a conversation. "How's arctic-fox doing?" I'm asking you how a specific set of artifacts is working out. I'm asking where it is in the pipeline maybe. "How's candidate-for-staging doing?" This doesn't convey any real meaning.

I want to avoid this particular example in particular because _if we show it, people will do it._

All this said, I struggle to find realistic uses for this step, although I know it's one that's often requested. The best, most realistic thing I can come up with (which also doesn't create the massive potential for name collision that hard coded `candidate-for-staging` does) is a scenario where Freight resources each reference _precisely one_ artifact and the user might like to have a simple Stage at the beginning of the pipeline that changes every Freight's alias to match the version/revision of the one artifact it references.

Something like this:

```suggestion
    newAlias: ${{ imageFrom('some/repo').Tag }}
```

pkg/promotion/runner/builtin/schemas/set-freight-alias-config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "SetFreightAliasConfig",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["freightName", "newAlias"],
```
krancour: "freight" and "new" are both redundant given the context.

```suggestion
  "required": ["name", "alias"],
```

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,139 @@
...
+	//
+	// Freight aliases are mutable and may be updated independently by multiple
+	// Stages within the same project. If alias-based selection were allowed,
+	// concurrent alias mutations could cause this step to resolve and mutate the
+	// wrong Freight.
+	//
+	// Using an immutable identifier guarantees deterministic behavior and avoids
+	// unintended cross-Stage side effects, even when multiple Stages update Freight
+	// aliases concurrently.
+	freight, err := api.GetFreightByNameOrAlias(
```
krancour: `api.GetFreight()` would be more straightforward here since there's no intention of possibly retrieving by alias.

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,139 @@
...
+			}, &promotion.TerminalError{
+				Err: fmt.Errorf("failed to fetch Freight %q in project %q: %w", cfg.FreightName, stepCtx.Project, err),
+			}
+	}
+
+	if freight == nil {
+		return promotion.StepResult{
+				Status: kargoapi.PromotionStepStatusFailed,
+			}, &promotion.TerminalError{
+				Err: fmt.Errorf("freight %q not found in project %q", cfg.FreightName, stepCtx.Project),
```
krancour: Linters might not like this, so there might be a `// nolint` you need to add here, but throughout the project, we defy the norm of "errors don't start with uppercase letters" if the leading word is a resource type.

```suggestion
				Err: fmt.Errorf("Freight %q not found in project %q", cfg.FreightName, stepCtx.Project),
```

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,139 @@
...
+			kargoapi.LabelKeyAlias,
+			cfg.NewAlias,
+			cfg.NewAlias,
+		),
+	)
+	patch := client.RawPatch(types.MergePatchType, patchBytes)
+
+	// Alias uniqueness is enforced by the admission webhook.
+	// Do not pre-check for conflicts here, as doing so would be
+	// vulnerable to TOCTOU races.
```
krancour: I would not bother to say this only because it implies the webhook is invulnerable to that same problem, but in fact, it is not. (There's nothing at all that can be done about that though because there's no way of blocking all _other_ Freight updates while updating this Freight resource's alias. 🤷‍♂️)

```suggestion
```

krancour:
Impl's good. Most comments are about docs.

pkg/promotion/runner/builtin/set_freight_alias.go
nitishfy: makes sense!

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -1,55 +1,43 @@
...
     freightName: ${{ ctx.targetFreight.name }}
-    newAlias: "candidate-for-staging"
-```
-
-In this example, the `Freight` currently being promoted is identified using its
-immutable name, ensuring the correct resource is targeted even in the presence
-of concurrent promotions or alias mutations. The alias is then updated to a
-human-readable value that is meaningful within the organization.
+    newAlias: ${{ imageFrom('some/repo').Tag }}
+```
```
krancour, line 54: If you could describe what you're doing here, it would be helpful. Something about this type of Freight having only one artifact and changing aliases to match the version of that one artifact.

This wouldn't work for Freight with multiple artifacts because you'd almost certainly have frequent conflicts, so I think it's important to describe the use case for which your example is actually safe and sensible.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,43 @@
...
+with one of their own choosing. This can make it easier to identify a
+particularly important (or problematic) `Freight` resource as it progresses
+through the `Stage`s of a pipeline.
+
+## Configuration
+
+| Name          | Type     | Required | Description                                                 |
+|---------------|----------|----------|-------------------------------------------------------------|
+| `freightName` | `string` | Y | The name of a `Freight` resource to update. |
+| `newAlias`    | `string` | Y | The desired new alias to set on the `Freight`.                |
```
krancour: These were changed, as suggested, elsewhere, but missed here.

```suggestion
| `name` | `string` | Y | The name of a `Freight` resource to update. |
| `alias`    | `string` | Y | The desired new alias to set on the `Freight`.                |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,43 @@
...
+## Examples
+
+### Common Usage
+
+```yaml
+steps:
+- uses: set-freight-alias
+  config:
+    freightName: ${{ ctx.targetFreight.name }}
+    newAlias: ${{ imageFrom('some/repo').Tag }}
```
krancour:
```suggestion
    name: ${{ ctx.targetFreight.name }}
    alias: ${{ imageFrom('some/repo').Tag }}
```

pkg/promotion/runner/builtin/schemas/set-freight-alias-config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "SetFreightAliasConfig",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["name", "alias"],
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "The immutable name of the Freight resource to update. This must be the Freight's name (its stable identifier), not an alias.",
```
krancour:
```suggestion
      "description": "The name of the Freight resource to update.",
```

pkg/promotion/runner/builtin/schemas/set-freight-alias-config.json
```diff
@@ -0,0 +1,19 @@
...
+      "minLength": 1
+    },
+    "alias": {
+      "type": "string",
+      "description": "The new human-readable alias to assign to the Freight. Aliases must be unique within the project.",
```
krancour:
```suggestion
      "description": "The new alias to assign to the Freight. Aliases must be unique within the project.",
```

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,131 @@
...
+					promotion.StepCapabilityAccessControlPlane,
+				},
+			},
+			Value: newSetFreightAlias,
+		},
+	)
+}
+
+// setFreightAlias is an implementation of the promotion.StepRunner interface
+// that updates alias on Freight resources.
```
krancour:
```suggestion
// that updates aliases of Freight resources.
```

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,131 @@
...
+type setFreightAlias struct {
```
krancour: This is inconsistent with how all the other built-in step runners are named. They've all been named as a noun; not a verb.

```suggestion
type freightAliasSetter struct {
```

krancour: Please rename the file accordingly as well. The JSON schema's filename can remain a verb as that will be consistent with existing precedent.

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (s *setFreightAlias) run(
+	ctx context.Context,
+	stepCtx *promotion.StepContext,
+	cfg builtin.SetFreightAliasConfig,
+) (promotion.StepResult, error) {
+	// Use the immutable name to avoid
+	// resolving the wrong Freight if an alias
+	// changes concurrently.
```
krancour: This is just the normal way of finding any resource. I don't think it needs and qualification.

```suggestion
```

pkg/promotion/runner/builtin/set_freight_alias.go
```diff
@@ -0,0 +1,131 @@
...
+			}, &promotion.TerminalError{
+				Err: fmt.Errorf("failed to fetch Freight %q in project %q: %w", cfg.Name, stepCtx.Project, err),
+			}
+	}
+
+	if freight == nil {
+		return promotion.StepResult{
+				Status: kargoapi.PromotionStepStatusFailed,
+			}, &promotion.TerminalError{
+				// nolint
```
krancour: Please dismiss a specific finding or at least finding from a specific linter. This dismisses everything.

krancour:
This looks good. There are a handful of small doc changes remaining and some things in the implementation that need renaming. Everything else is good.

pkg/promotion/runner/builtin/set_freight_alias_test.go
```diff
@@ -0,0 +1,206 @@
...
+				Name:  testFreight,
+				Alias: newAlias,
+			},
+			assertions: func(t *testing.T, res promotion.StepResult, c client.Client, err error) {
+				require.NoError(t, err)
+				require.Equal(t, kargoapi.PromotionStepStatusSucceeded, res.Status)
+
+				freight := &kargoapi.Freight{}
+				err = c.Get(
+					context.Background(),
```
krancour: `t.Context()` is a better choice than `context.Background()`. There are many existing tests still using `context.Background()`, but we've been using `t.Context()` in new tests and we're often replacing `context.Background()` with `t.Context()` if we happen to be touching any given `*_test.go` file for _other_ changes of any kind.

```suggestion
					t.Context(),
```

pkg/promotion/runner/builtin/set_freight_alias_test.go
```diff
@@ -0,0 +1,206 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			step := &freightAliasSetter{
+				kargoClient: tt.client,
+			}
+
+			stepCtx := &promotion.StepContext{
+				Project: testProject,
+			}
+
+			res, err := step.run(context.Background(), stepCtx, tt.cfg)
```
krancour:
```suggestion
			res, err := step.run(t.Context(), stepCtx, tt.cfg)
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
...
+`set-freight-alias` updates the alias of a `Freight` resource.
+
+When a `Warehouse` produces a new `Freight` resource, it computes a SHA-1 hash
+from a canonical representation of the artifacts referenced by that resource.
+This becomes the value of the `metadata.name` field — a unique "fingerprint" of
+the collection of artifacts. While every `Freight` resource is also automatically
+assigned a human-friendly alias, users may sometimes wish to override that alias
+with one of their own choosing. This can make it easier to identify a
+particularly important (or problematic) `Freight` resource as it progresses
+through the `Stage`s of a pipeline.
```
krancour:
```suggestion
When a `Warehouse` produces a new `Freight` resource, it computes a SHA-1 hash
from a canonical representation of the artifacts referenced by that resource.
This unique "fingerprint" for that collection of artifacts becomes the value of
the `metadata.name`. Because this is not a very human-friendly name, every
`Freight` resource is also automatically assigned a human-friendly alias. Users
may sometimes wish to update aliases at various points in their pipelines. This
can make it easier to identify a particularly important (or problematic)
`Freight` resource as it progresses through the `Stage`s of a pipeline.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
...
+in use by another `Freight` resource in the project, this step will fail.
+
+:::
+
+## Examples
+
+When a `Freight` resource references only a single artifact, such as a
+container image, its alias can be updated to reflect that artifact's version.
+This makes it easy to identify the `Freight` resource at a glance as it progresses
+through a pipeline.
```
krancour:
```suggestion
In this example, a `Freight` resource that references only a single container
image and no other artifacts is updated to reflect the image's tag. This makes
it easier to identify such resources and make inferences about their payload.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/set-freight-alias.md
```diff
@@ -0,0 +1,55 @@
...
+
+:::caution
+
+This pattern is only safe when the `Freight` resource contains a single artifact. If the
+`Freight` resource references multiple artifacts (e.g., several images or Git commits),
+different `Freight` resources could easily produce the same alias, causing
+conflicts that fail the promotion step.
```
krancour:
```suggestion
This example is safe only because the `Freight` resource contains only a single
artifact. If that were not the case, contention over a single alias would arise
when two or more Freight resources referenced the same version of the container
image.
```

PR #5521, opened 2025-12-19
feat: allow dynamic loading of ArgoCD URLs from secrets

### LINKED ISSUE #4971: Allow dynamic loading of argocd urls
# Checklist

# Proposed Feature

Allow for easier passing of urls instead of requirement to redeploy whole helm chart. E.g. looking for secrets with specific labels.

# Motivation

With sharded architecture, onboarding new shard requires redeployment of the kargo chart on hub to load additional argocd url. This could be simplified with secrets as those can be created and deleted much easier, without adding the need to redeploy whole app.

ksawerykarwacki:
## Summary

Add support for dynamically loading ArgoCD shard URLs from Kubernetes secrets without requiring Helm chart redeployment.

- Secrets labeled with `kargo.akuity.io/argocd-shard=true` in the Kargo namespace are watched and used to configure ArgoCD shard URLs
- Dynamic shards override static (Helm) configuration with the same name
- When a secret is deleted, any static configuration is restored
- Backwards compatible - existing Helm configuration continues to work unchanged

### Secret Format

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: argocd-shard-production
  namespace: kargo
  labels:
    kargo.akuity.io/argocd-shard: "true"
type: Opaque
stringData:
  name: "production"                      # Shard identifier (empty for default)
  url: "https://argocd.example.com"       # ArgoCD base URL
```

Closes #4971


ksawerykarwacki:
Any chances that soemone will look at this? @krancour, @hiddeco? 

This is small change but lack of it is rather frustrating in the sharded deployement.

krancour:
I understand the value of the proposed change, but am perplexed by the approach. Why would Secrets be used for this purpose? In general, I'd like us to be doing a lot less with Secrets when it's feasible to do so. I don't see what's sensitive about this configuration and it predicates the proper functioning of this feature on relatively privileged level of access. Even if the API server might already have that, this shouldn't become one of the _reasons_ it does.

Why wouldn't this have been art of `ClusterConfiguration` instead?

PR #5522, opened 2025-12-19
feat(promotion): add ignoreMissingValueFiles to helm-template step

### LINKED ISSUE #4349: ignoreMissingValueFiles from ArgoCD for helm-template step
# Proposed Feature

A port of functionality similar to `ignoreMissingValueFiles: true`. I'd like this so I can have optional environment specific values overlays but not mandatory to have empty placeholder yaml for every stage in my kargo project.

# Motivation

I have values.yaml and ${stage}.yaml(s) for every stage in my project currently just so I could override anything on the stage specific basis, but when you have many stages it's a lot of empty files I don't need to exist and if I ever add a stage I have to backfill empty placeholders for every single helm template chart upstream.

ksawerykarwacki:
## Summary
- Adds `ignoreMissingValueFiles` boolean option to the `helm-template` promotion step
- When enabled, missing value files are silently skipped instead of causing an error
- Eliminates the need for empty placeholder YAML files per stage

Closes #4349

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/helm-template.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/helm-template.md
@@ -22,6 +22,7 @@ commonly preceded by a [`git-clear` step](git-clear.md) and followed by
 | `useReleaseName` | `boolean` | N | Whether to use the release name in the output path (instead of the chart name). This is `false` by default, and only has an effect when `outPath` is set to a directory. |
 | `namespace` | `string` | N | Optional namespace to use when rendering the manifests. This is commonly omitted. GitOps agents such as Argo CD will generally ensure the installation of manifests into the namespace specified by their own configuration. |
 | `valuesFiles` | `[]string` | N | Helm values files (apart from the chart's default `values.yaml`) to be used when rendering the manifests.  |
+| `ignoreMissingValueFiles` | `boolean` | N | Whether to skip value files that do not exist instead of returning an error. Useful for optional stage-specific overlays. This is `false` by default. |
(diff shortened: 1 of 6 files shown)


krancour:
I'm not specifically opposed to this, but I do vaguely recall some prior discussion about this, or else it was a very similar conversation. Whether it was on this exact topic or just similar one, I think @hiddeco may have once expressed some reluctance to the idea of steps that wrap a tool (like helm) having options that the tool itself lacks.

`helm template` cannot do this.

It is quite possible I am mis-remembering the aforementioned conversation, so let's have @hiddeco weigh in.

fwiw, there is another way to do this in v1.9.0. There's a step that merges a list of YAML files and it already has an option to ignore any that are missing. It would be possible to use that first to merge your YAML files (from most general to most specific) and then use helm template on the one resulting file.

ksawerykarwacki:
I missed that there is new step for merging YAML (i see that it has this flag implemented). 

My reasoning was to bring Kargo on par with ArgoCD, as some folks (including myself) run mix of both  (e.g. ArgoCD managed apps for dev, PoCs, etc. and rendered manifests via Kargo for staging and prod). Of course I can wrap `yaml-merge` and `helm-template` under single promotion task `my-helm-template` to achieve the same but will flesh out other limitations (vars only support strings so you cannot pass list of value files to it) and will make things more convoluted just for sake of keeping the action as limited as the original source of inspiration (`helm template` command)

@hiddeco it is your call then if it is worth to have this or not

hiddeco:
My concern with a flag at a global level like this is that it opens the gates for incidents where e.g. a typo results in something to go unnoticed because any path is allowed to be missing.

I would be more open to the idea of allowing specific paths to be optional. But the problem with this (and the current API) is that the list of values files is just a simple list (and you can't add fields per list item).

hiddeco:
I do believe we had the same discussion for the `yaml-merge` step though, and as it was added at a global level there my concern isn't particularly strong if we follow our own existing API patterns.

hiddeco (approved):
Thank you! 🙇

PR #5524, opened 2025-12-20
refactor: clean up logging package

### LINKED ISSUE #5523: bug: trace level logging is broken
This is very complicated when you dive into the code, but I'll try and keep the explanation simple.

We have our own logging interface we use instead of logr because, to be blunt, the ergonomics of logr are terrible. _Under the hood_, we use logr with zap as its underlying sink. logr is still used as a middle man between our interface and zap so that if we ever, hypothetically, wanted to replace zap, we just plug in a different sink and we're good to go. In reality, this is probably not quite true because there's a fair amount in our logging package that knows about zap. It has to, because...

For whatever reason, zap doesn't have a TRACE level. @thomastaylor312, when switching the sink from logrus to zap, tried to compensate for this, but it doesn't quite work. This is where I'll gloss over some details, because they're mind numbing and I'm actually struggling to keep it all straight. Suffice it to say that if we're going to play games to make something trace-like work with zap, we've already lost the benefits of using logr as a middle man. Plugging in any other log sink would require ripping out a bunch of zap-specific workarounds...

So here are the choices that I see in front of us:

1. Forget about trace-level logging entirely:
     
    * Switch all Trace() calls to Debug() OR
    * Just remove all Trace() calls OR
    * Some combination of those options depending on the perceived value of the message

    I don't love this option because we do have some areas of the code base where there are log messages that are helpful to see at development time, but are so voluminous that I personally don't want to see them unless I am working on one of those specific areas of the code and have opted into it. If we drop those messages, it will make troubleshooting complex areas of the code more difficult. If we change them to debug-level, the debug logs will be cluttered with a large volume of messages that we very seldom care about.

1. Use a different sink. 

    I don't have any special attachment to zap. I know it's popular. Presumably @thomastaylor312 had reasons he selected zap. I will assume for now that @thomastaylor312 would not prefer this option, but he can correct me if I am wrong.

1. Cut out log4r.

    Because of the workarounds required for zap (which aren't working), log4r isn't _really_ making it trivial to swap in different sinks, which was our reason for using it. Assuming we _didn't_ dump zap and zap-specific concessions continued to exist in our own logging code, we could cut out the middle layer that's providing no tangible value (under these circumstances) and it would make the required workarounds far, far easier to reason over and we can get some kind of trace-like logging working once again.

1. Keep everything as it currently is, but invest a lot of time in reasoning over the complexity to make zap-specific workarounds work as intended.

My preference is for no.3.

I definitely do not want to spend a lot of time bike-shedding over this, because we definitely have higher priorities, but I do kind of want my trace logs back.

krancour:
Fixes #5523 and resolves some other log-related things as well.

Request review from @thomastaylor312 since we had offline conversations about this.

This PR removes logr as the middleman between our own ergonomic logging interface and zap. The primary impetus as discussed in #5523 was that coercing zap to support TRACE-level-like logging, which it normally does not, was significantly easier without logr in between us.

TRACE logging works now.

We don't lose anything by doing this. logr was already not really doing anything for us and anywhere where we really _need_ a `logr.Logger`, perhaps to pass to a third-party framework or library, our own `logging.Logger` now has a `Logr()` method that will return our logger's underlying `zap.Logger` wrapped as a `logger.Logr()`. 🍰

I noticed that output from any new `logging.Logger`s created from scratch was not properly synchronized with output from the global `Logger`, (or output from klog or controller-runtime) because new `Logger`s created from scratch were not using the same sync writer shared by all those other ones. I fixed that while I was at it.

I also found and remediated one lingering use of logrus.

I am including this in v1.9.0 and marking it high priority only because the repaired TRACE logging will be an asset in finishing up work on other, specific features in this release.

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -95,6 +95,10 @@ func init() {
 // Logger is a simple wrapper around zap.Logger that provides a more ergonomic
 // API.
 type Logger struct {
+	// Note(krancour): Carrying around a zap.SugaredLogger may possibly incur a
+	// very small performance penalty, but it spares us from having to duplicate
+	// some of that type's functionality, such as the conversion of context in the
+	// form of []any to []zap.Field. I consider this a worthwhile trade-off.
 	logger *zap.SugaredLogger
 }
 
@@ -121,7 +125,18 @@ func NewLogger(level Level, format Format) (*Logger, error) {
 }
 
 func newLoggerInternal(level Level, format Format) (*Logger, error) {
-	if level < TraceLevel || level > DiscardLevel {
+	if level == DiscardLevel {
+		// Note(krancour): Building a leveled logger with the level set higher than
+		// LevelFatal may actually work, but comments within zap's source code
+		// suggest that was intended to be invalid. Rather than relying on the
+		// observed behavior to never change, we take zap at its word and just
+		// return a no-op logger here when the caller has requested a logger of
+		// DiscardLevel.
+		return &Logger{
+			logger: zap.NewNop().Sugar(),
+		}, nil
+	}
+	if level < TraceLevel || level > ErrorLevel {
 		return nil, fmt.Errorf("invalid log level: %d", level)
 	}
 	// Re-parsing the format we were given has the side effects of validating and
--- a/pkg/logging/logger_test.go
+++ b/pkg/logging/logger_test.go
@@ -2,6 +2,7 @@ package logging
 
 import (
 	"errors"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -17,7 +18,11 @@ func TestGlobalLogger(t *testing.T) {
 func TestNewDiscardLoggerOrDie(t *testing.T) {
 	logger := NewDiscardLoggerOrDie()
 	require.NotNil(t, logger)
-	require.Equal(t, zapcore.FatalLevel+1, logger.logger.Level())
+	require.Equal(
+		t,
+		"zapcore.nopCore",
+		reflect.TypeOf(logger.logger.Desugar().Core()).String(),
+	)
 }
 
 func TestNewLoggerOrDie(t *testing.T) {


pkg/logging/context.go
```diff
@@ -0,0 +1,18 @@
+package logging
+
+import "context"
+
+// ContextWithLogger returns a context.Context that has been augmented with
+// the provided *Logger.
+func ContextWithLogger(ctx context.Context, logger *Logger) context.Context {
+	return context.WithValue(ctx, loggerContextKey{}, logger)
```
thomastaylor312, line 10: Nit: `loggerContextKey` should go in this file

pkg/logging/logger.go
```diff
@@ -37,218 +39,194 @@ const (
...
 func (l *Logger) Trace(msg string, keysAndValues ...any) {
-	l.callStackHelper()
-	l.logger.V(2).Info(msg, keysAndValues...)
+	// Note(krancour): Zap doesn't actually have a trace level, but WE have a
+	// trace level that is numerically below debug. A zap.Logger that was
+	// initialized with that numeric level WILL answer true to
+	// Enabled(zapcore.Level(TraceLevel)). Neat, right? So if this pseudo-trace
+	// level is enabled, we log at debug level with a "TRACE: " prefix.
+	if l.skipped.Core().Enabled(zapcore.Level(TraceLevel)) {
+		l.skipped.Debug("TRACE: "+msg, toZapFields(keysAndValues...)...)
```
thomastaylor312: Nit: Rather than manually encoding here, could we build the `Core` using a [`LevelEncoder`](https://pkg.go.dev/go.uber.org/zap@v1.27.1/zapcore#LevelEncoder)? We should serialize out the trace level properly and then delegate to the underlying `Level` methods for all other calls (see [here](https://pkg.go.dev/go.uber.org/zap@v1.27.1/zapcore#Level.CapitalString)) to all other levels

pkg/logging/logger.go
```diff
@@ -37,218 +39,194 @@ const (
...
-// returns it. If no *Logger is found, a global *Logger is returned.
-func LoggerFromContext(ctx context.Context) *Logger {
-	if logger := ctx.Value(loggerContextKey{}); logger != nil {
-		return ctx.Value(loggerContextKey{}).(*Logger) // nolint: forcetypeassert
+// WithValues adds key-value pairs to a logger's context.
+func (l *Logger) WithValues(keysAndValues ...any) *Logger {
+	logger := l.logger.With(toZapFields(keysAndValues...)...)
+	return &Logger{
+		logger:  logger,
+		skipped: logger.WithOptions(zap.AddCallerSkip(1)),
```
thomastaylor312: Rather than carrying around 2 logger instantiations and then mostly using `skipped`, could we just create a single logger with the `AddCallerSkip` and then if we need to use a "non-skip" logger (like for `Logr`), we can call `AddCallerSkip(-1)`. Based on [the code](https://github.com/uber-go/zap/blob/7b755a3910491932656b01f2013b8bf41e74d4e8/options.go#L108) this should work

krancour: Good call. And I also discovered that the _one_ we carry around may as well be a sugared logger since that spares us from implementing and testing `toZapFields()`. Just as you suggested we can `AddCallerSkip(-1)` as needed (which is rarely), we can also `Desugar()` as needed.

Fixed in https://github.com/akuity/kargo/pull/5524/commits/90cd7520af13b52918a4ac69635dd373eed5f87b

pkg/logging/logger.go
krancour: So true story... copilot had told me there wasn't a way to add custom levels to zap, so I'd not tried very hard to do so. The lie detector has determined... that was a lie.

Your suggestion worked out really nicely and now trace logs look like there were built right into zap.

Screenshot 2025-12-22 at 11 30 49 AM

😍

krancour: Fixed in https://github.com/akuity/kargo/commit/90cd7520af13b52918a4ac69635dd373eed5f87b

pkg/logging/context.go
krancour: Agree. Accidentally left behind in the initial shuffle.

Fixed in https://github.com/akuity/kargo/commit/90cd7520af13b52918a4ac69635dd373eed5f87b

pkg/logging/logger.go
```diff
@@ -17,238 +15,210 @@ import (
...
 
 const (
-	ErrorLevel = iota
-	InfoLevel
-	DebugLevel
-	TraceLevel
+	// Note(krancour): Numerically speaking, zap supports levels above or below
+	// those for it has defined constants. This is how we implement our own
+	// Discard and Trace levels.
+	DiscardLevel Level = Level(zapcore.FatalLevel + 1)
```
thomastaylor312, line 26: This is invalid, so we should probably set this under TraceLevel. See https://github.com/uber-go/zap/blob/7b755a3910491932656b01f2013b8bf41e74d4e8/zapcore/level.go#L57-L60

pkg/logging/logger.go
```diff
@@ -17,238 +15,210 @@ import (
...
-	case TraceLevel:
-		// There is no TraceLevel in zap, so we use DebugLevel
-		return zapcore.DebugLevel
-	default:
-		return zapcore.InfoLevel
-	}
+// Logger is a simple wrapper around zap.Logger that provides a more ergonomic
+// API.
+type Logger struct {
+	logger *zap.SugaredLogger
```
thomastaylor312, line 102: Nit: Just for posterity sake in case it ever matters, can you document in a comment here why we chose a sugared logger? I think the reasoning of "we'll take the performance hit to not be maintaining code for converting to zap fields" is sound and we don't need to eek out every bit of performance with a non-sugared logger.

I just feel like this would be one of those things down the line that someone will be like "why did we choose sugared" when suddenly we do have some performance concerns

pkg/logging/logger.go
krancour: Nice catch. So far, it works (tests use this and they're not panicking). But I agree it's better to be on the safe side.

I'm not sure that defining it as one less than TraceLevel is the right thing though. I think that would have the opposite of the intended effect. Trace is log everything, while discard is log nothing.

It may be that the better approach is for `DiscardLoggerOrDie()` to return a logger with a no-op core.

pkg/logging/logger.go
krancour: You got it.

pkg/logging/logger.go
thomastaylor312: Yeah probably

PR #5538, opened 2025-12-23
feat(chart): allow `priorityClassName` configuration

apt-itude:
This is essential configuration to ensure that Kargo components are given appropriate scheduling priority according to the standards you've established for your cluster. Without being able to configure the priority class, Kargo pods may frequently be evicted in order to schedule pods that do have a priority class configured.

This my first PR to this repository, so please provide any contribution guidance that I need to follow!


krancour:
@apt-itude thanks for the contribution and sorry for the delay getting around to reviewing it.

This looks pretty good. The only change I would really as for: If, for each component, you could move `priorityClassName` down to somewhere around where affinity and tolerations are optionally specified, I think it would be a more natural place for it.

Only other thing is you need to run `make hack-cogeden` after chart changes. Among other things, it refreshes the chart's own README.md.

apt-itude:
> @apt-itude thanks for the contribution and sorry for the delay getting around to reviewing it.
> 
> This looks pretty good. The only change I would really as for: If, for each component, you could move `priorityClassName` down to somewhere around where affinity and tolerations are optionally specified, I think it would be a more natural place for it.
> 
> Only other thing is you need to run `make hack-cogeden` after chart changes. Among other things, it refreshes the chart's own README.md.

@krancour Done!

PR #5558, opened 2026-01-09
fix: make the ui service on 30082 available

DavidS-ovm:
Without this change I was not able to complete the https://docs.kargo.io/contributor-guide/hacking-on-kargo/ guide (using kind).

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -67,6 +67,7 @@ k8s_yaml(
 # Normally the API server serves up the front end, but we want live updates
 # of the UI, so we're breaking it out into its own separate deployment here.
 k8s_yaml('hack/tilt/ui.yaml')
+k8s_yaml('hack/tilt/ui-service.yaml')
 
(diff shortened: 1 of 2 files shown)


krancour:
@DavidS-ovm I am assuming you did not use `make hack-kind-up` as directed in those docs? If you do, kind is launched with all necessary port-forwarding already established. I tested it just now and it worked as intended.

DavidS-ovm:
> @DavidS-ovm I am assuming you did not use `make hack-kind-up` as directed in those docs? If you do, kind is launched with all necessary port-forwarding already established. I tested it just now and it worked as intended.

There was some trashing, but I'm pretty sure I used the makefile:
```
✦2 ❯ history | grep kind
33358  make hack-kind-up
```

DavidS-ovm:
I've retried this morning with fresh eyes and could not reproduce this either. Thanks for double checking for me.

PR #5562, opened 2026-01-10
chore: Upgrade k8s to 1.35

thomastaylor312:
Ok, so this one is a doozy, and full disclaimer, I hate it. But it is what is going to have to work.

## Background

Due to the [upcoming removal of the gogo proto dependency](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5589-gogo-dependency), we have to enable a build flag to reenable the V1 proto message marker method `ProtoMessage` so things can work with the connect API. However, because types no longer have a `Descriptor` method, we would get a panic when Marshaling out any message that had a kubernetes type using bare `map` types because there is no registered type for them. We work around this already in our code with special `protobuf_key` and `protobuf_val` tags for our maps. Even worse, the k8s `ObjectMeta` type works just fine mostly because of a quirk of message handling (detailed later)

## What this actually does

I tried messing around with the underlying `Codec` used by ConnectRPC, but the main problem is you can't update the descriptor at runtime. So, as far as I could tell, I couldn't use `protoreflect` to try to get around this and the `proto.Marshal` method doesn't work like `MarshalJSON` where it recurses down each type in a struct and calls its `MarshalJSON` method. So the easiest (and very very hacky) solution was to copy the types that we had problems with (Secret and ConfigMap, as well as ObjectMeta) from the k8s proto files and add them to our service. We do have `ServiceAccount` but this isn't really used by the UI and I don't think it has any top level maps other than labels (which may work in this case?)

Please note that our types work purely because of the quirk described below where labels and annotations are just skipped and assumed empty when returning them as a protobuf type. This doesn't affect our UI because we use the `RawFormat` option to get the yaml bytes back, which returns the full object

## The gnarly technical quirk, here there be dragons

I asked Claude to help me summarize it and its breakdown was perfect, so I present it here in its entirety:

Summary: Why ObjectMeta Works but ConfigMap Doesn't

**The Key Discovery**

The root cause is in how aberrantLoadMessageDescReentrant handles different Go types:

```go
func aberrantLoadMessageDescReentrant(t reflect.Type, name protoreflect.FullName) protoreflect.MessageDescriptor {
    // ...
    if t.Kind() != reflect.Ptr || t.Elem().Kind() != reflect.Struct {
        return md  // Returns EMPTY descriptor for non-pointer types!
    }
    // ... process fields only for *struct types ...
}
```

For pointer types (*struct): The function processes all struct fields, including map fields that need protobuf_key/protobuf_val tags.

For value types (struct, not pointer): The function returns an empty descriptor without processing any fields.

**How This Affects ConfigMap vs Stage**

ConfigMap (*v1.ConfigMap):
1. ConfigMap is referenced in service.pb.go as a pointer type
2. aberrantLoadMessageDescReentrant processes ConfigMap's fields
3. ConfigMap's Data field is a map[string]string without protobuf_key/protobuf_val tags
4. This creates a malformed map entry descriptor (fields have number 0 instead of 1 and 2)
5. During needsInitCheck, when processing the Data field:
  - fd.IsMap() returns true
  - fd.MapValue() tries to find field #2, but there's no field #2 → returns nil
  - fd.Message() is called on nil → PANIC

Stage (*v1alpha1.Stage):
1. Stage is referenced as a pointer type
2. aberrantLoadMessageDescReentrant processes Stage's fields
3. Stage's fields are:
  - ObjectMeta - embedded as value type (struct, not pointer)
  - Spec StageSpec - embedded as value type
  - Status StageStatus - embedded as value type
4. For each embedded struct field, when aberrantLoadMessageDescReentrant is called, the type is a struct (not pointer), so it short-circuits and returns an empty descriptor
5. These empty descriptors have no fields
6. During needsInitCheck, when recursing into ObjectMeta, Spec, or Status, there are no fields to iterate
7. ObjectMeta's Labels and Annotations map fields are never seen → No panic

**Visual Comparison**

```
ConfigMap (fails):
├── metadata (ObjectMeta) - struct type → empty descriptor ✓
├── Data map[string]string - map type → MALFORMED map entry → PANIC ✗
└── BinaryData map[string][]byte - map type → MALFORMED map entry → PANIC ✗

Stage (works):
├── ObjectMeta - struct type → empty descriptor (hides Labels/Annotations maps) ✓
├── Spec (StageSpec) - struct type → empty descriptor ✓
└── Status (StageStatus) - struct type → empty descriptor ✓
```

**The Irony**

ObjectMeta works not because it's special, but because:
1. It's always embedded as a value type (not pointer)
2. The aberrant loading code accidentally skips non-pointer types
3. This bug/quirk hides ObjectMeta's problematic map fields from the init check

ConfigMap fails because it has direct map fields at the top level of its struct. These maps don't have protobuf_key/protobuf_val tags, and since ConfigMap is processed as a pointer type, these fields are encountered and their malformed descriptors cause the panic.

### DIFF
--- a/ui/src/features/common/settings/access/api-tokens/api-tokens-list.tsx
+++ b/ui/src/features/common/settings/access/api-tokens/api-tokens-list.tsx
@@ -7,7 +7,7 @@ import { format } from 'date-fns';
 
 import { useModal } from '@ui/features/common/modal/use-modal';
 import { listAPITokens } from '@ui/gen/api/service/v1alpha1/service-KargoService_connectquery';
-import { Secret } from '@ui/gen/k8s.io/api/core/v1/generated_pb';
+import { Secret } from '@ui/gen/api/service/v1alpha1/service_pb';
 
 import { CreateAPITokenModal } from './create-api-token-modal';
 import { DeleteAPITokenButton } from './delete-api-token-button';
--- a/ui/src/features/project/settings/views/config-maps/config-maps.tsx
+++ b/ui/src/features/project/settings/views/config-maps/config-maps.tsx
@@ -11,7 +11,7 @@ import {
   deleteResource,
   listConfigMaps
 } from '@ui/gen/api/service/v1alpha1/service-KargoService_connectquery';
-import { ConfigMap } from '@ui/gen/k8s.io/api/core/v1/generated_pb';
+import { ConfigMap } from '@ui/gen/api/service/v1alpha1/service_pb';
 
 import { UpsertConfigMapsModal } from './upsert-config-maps';
 
--- a/ui/src/features/project/settings/views/credentials/create-credentials-modal.tsx
+++ b/ui/src/features/project/settings/views/credentials/create-credentials-modal.tsx
@@ -14,7 +14,7 @@ import {
   updateRepoCredentials,
   updateGenericCredentials
 } from '@ui/gen/api/service/v1alpha1/service-KargoService_connectquery';
-import { Secret } from '@ui/gen/k8s.io/api/core/v1/generated_pb';
+import { Secret } from '@ui/gen/api/service/v1alpha1/service_pb';
 
 import { createFormSchema } from './schema-validator';
 import { SecretEditor } from './secret-editor';
--- a/ui/src/features/project/settings/views/credentials/credentials.tsx
+++ b/ui/src/features/project/settings/views/credentials/credentials.tsx
@@ -19,7 +19,7 @@ import {
   listRepoCredentials,
   listGenericCredentials
 } from '@ui/gen/api/service/v1alpha1/service-KargoService_connectquery';
-import { Secret } from '@ui/gen/k8s.io/api/core/v1/generated_pb';
+import { Secret } from '@ui/gen/api/service/v1alpha1/service_pb';
 
 import { CreateCredentialsModal } from './create-credentials-modal';
 import { CredentialTypeLabelKey, CredentialsDataKey, CredentialsType } from './types';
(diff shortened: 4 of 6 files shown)


pkg/controller/git/bare_repo_test.go
```diff
@@ -111,9 +111,8 @@ func TestBareRepo(t *testing.T) {
 	workingTreePath := filepath.Join(rep.HomeDir(), "working-tree")
 	workTree, err := rep.AddWorkTree(
 		workingTreePath,
-		// "master" is still the default branch name for a new repository unless
-		// you configure it otherwise.
-		&AddWorkTreeOptions{Ref: "master"},
+		// Don't assume a default branch name ("main" vs "master").
```
thomastaylor312, line 114: I kept running into issues with the test here, so I made this more flexible. Shouldn't impact the tests themselves, but let me know if I missed something

thomastaylor312:
Please note that the backend is ready to go and I've spot checked functionality, but obviously UI stuff is failing and I need to have a more thorough manual test done. However, it is pretty much ready for review as is otherwise

krancour:
✅ Build flags

❓ Is the hack of defining our own knock-off corev1 types worth it if we plan to transition away from protobufs? I'm not against it, but want to understand the risks.

thomastaylor312:
Putting this context here from an offline convo: turning on just the build flags isn’t sufficient because anything involving k8s objects with maps panics

pkg/controller/git/bare_repo_test.go
krancour: Changes look ok, but I'm perplexed as to why the other changes in this PR would have surfaced any issues in this spot. 🤔

krancour:
@thomastaylor312 looks like there are a bunch of merge conflicts, but I'd say this LGTM otherwise. Caveat: I'm suspicious that the changes in service.proto may actually be breaking, however, I believe if they are, at this point, it's ok...

v1.9.0+ users have CLIs already using the REST API. As far as UI users are concerned: The UI SPA is served from the API server itself. So it's impossible to have a version mismatch between the UI and APIs.

As I said... LGTM pending resolution of merge conflicts.

Also, I think it's best if we get this change in early in the v1.10 cycle so there's plenty of time to detect any unforeseen issues.

krancour:
In our last chat, @thomastaylor312 and I agreed not to do this, unless for some strange reason we didn't have the UI transitioned to the new API before k8s v1.34 is EOL.

PR #5575, opened 2026-01-14
fix: forbidden secret errors

fuskovic:
@rpelczar reported an error while working on https://github.com/akuity/kargo/pull/5574

I investigated and found two errors.

- General -> Access
- Projects -> Secrets

https://github.com/user-attachments/assets/1dc4a108-68f5-4eab-8c48-a671ee4f68e4

This PR fixes both errors.

https://github.com/user-attachments/assets/12ed60e6-eaa9-4158-8f64-2c996e4bacc8

### DIFF
--- a/charts/kargo/templates/users/role-bindings.yaml
+++ b/charts/kargo/templates/users/role-bindings.yaml
@@ -12,3 +12,6 @@ subjects:
 - kind: ServiceAccount
   namespace: {{ .Release.Namespace }}
   name: kargo-admin
+- kind: ServiceAccount
+  namespace: {{ .Release.Namespace }}
+  name: kargo-api


charts/kargo/templates/shared-resources-namespace/role-bindings.yaml
```diff
@@ -17,4 +17,7 @@ subjects:
 - kind: ServiceAccount
   namespace: {{ .Release.Namespace }}
   name: kargo-controller
+- kind: ServiceAccount
+  namespace:  {{ .Release.Namespace }}
+  name: kargo-api
```
fuskovic, line 22: This gives `kargo-api` service account read access to secrets in the `shared-resources-namepace`

charts/kargo/templates/api/cluster-roles.yaml
```diff
@@ -19,6 +19,7 @@ rules:
       - ""
     resources:
       - serviceaccounts
+      - secrets
     verbs:
```
fuskovic, line 22: This fixes the issue of `kargo-api` read-access to `kargo` namespace secrets but I'm not sure this is the *right* way to fix this.

jessesuen: It's not. Since you are modifying the ClusterRole, you are giving `kargo-api` the privilege to read secrets in the **_whole cluster_**. Something we are actively trying to avoid. We only need `kargo-api` to read secrets in the following three namespaces:
1. `kargo-shared-resources`
2. `kargo`
3. `kargo-system-resources`

You already fixed (1) with the RoleBinding to `kargo-shared-resources-admin` Role. You will need to fix (2) and (3) in a similar manner.

jessesuen (changes requested):
kargo-api needs a RoleBinding (not ClusterRoleBinding) to a Role (not a ClusterRole), giving privilege to `kargo` and `kargo-system-resources` namespaces.

charts/kargo/templates/api/cluster-roles.yaml
fuskovic: 3 looks ok: https://github.com/akuity/kargo/blob/main/charts/kargo/templates/system-resources-namespace/role-bindings.yaml#L34

I've fixed 2.

access_test

jessesuen (approved):

charts/kargo/templates/shared-resources-namespace/role-bindings.yaml
krancour: This should probably be only when `.Values.api.secretManagementEnabled` is true.

charts/kargo/templates/users/role-bindings.yaml
```diff
@@ -12,3 +12,6 @@ subjects:
 - kind: ServiceAccount
   namespace: {{ .Release.Namespace }}
   name: kargo-admin
+- kind: ServiceAccount
+  namespace: {{ .Release.Namespace }}
+  name: kargo-api
```
krancour, line 17: I would tweak this a little bit. Both the API server's permissions and the admin's permissions are pretty broad, but they're not quite the same, and I'm reluctant to mix them up like this because it complicates the mental model of who (or what) can do what.

The more targeted approach I'd rather we took to this would be adding a `kargo-api` `Role` and `RoleBinding` to Kargo's own namespace, and conditionally give that new `Role` read-only or full access to `Secret`s in Kargo's own namespace, depending on the value of `.Values.api.secretManagementEnabled`.

charts/kargo/templates/shared-resources-namespace/role-bindings.yaml
fuskovic: Addressed here: https://github.com/akuity/kargo/pull/5585

charts/kargo/templates/users/role-bindings.yaml
fuskovic: Addressed here: https://github.com/akuity/kargo/pull/5585

Issue #5580, opened 2026-01-14
Add a merge method to the git-merge-pr step


bobdoah:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

The current `git-merge-pr`  step always performs a merge commit. Many projects prefer to squash or rebase commits. 

The GitHub API [supports setting the `merge_method`  parameter on the request](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#merge-a-pull-request)

The Gitea SDK [supports a `MergeStyle` parameter](https://gitea.com/gitea/go-sdk/src/commit/060554f46291abccac23a24f5e9bb264f962e21d/gitea/pull.go#L231), which [has the same methods](https://gitea.com/gitea/go-sdk/src/commit/060554f46291abccac23a24f5e9bb264f962e21d/gitea/pull.go#L87)

The Azure DevOps API [supports similar options](https://learn.microsoft.com/en-us/javascript/api/azure-devops-extension-api/gitpullrequestmergestrategy).

The BitBucket API used [lacks this](https://github.com/ktrysmt/go-bitbucket/blob/master/bitbucket.go#L300), but [`merge_strategy` is a supported supported parameter](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/#api-repositories-workspace-repo-slug-pullrequests-pull-request-id-merge-post) on the API request.

# Motivation

This would be necessary squash or rebasing is desired/enforced on a Git repository. This is the case for a project I am working on. Squash commits are enforced. I would like to make use of Kargo, but this is a bit of a blocker.

krancour:
Should have been closed by #5581

PR #5581, opened 2026-01-14
feat: Add squash merge support to git-merge-pr promotion step

bobdoah:
Adds a new 'mergeMethod' configuration option to the git-merge-pr promotion step that allows users to specify the merge strategy when merging pull requests.

This fulfils the feature request I opened in #5580.

### DIFF
--- a/pkg/gitprovider/azure/pr_integration_test.go
+++ b/pkg/gitprovider/azure/pr_integration_test.go
@@ -1,0 +1,59 @@
+//go:build integration && azure
+
+package azure
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/akuity/kargo/pkg/gitprovider"
+	gptest "github.com/akuity/kargo/pkg/gitprovider/testing"
+)
+
+func TestCreateAndMergePullRequest(t *testing.T) {
+	repoURL := gptest.RequireEnv(t, "TEST_AZURE_REPO_URL")
+	token := gptest.RequireEnv(t, "TEST_AZURE_TOKEN")
+	gitUsername := gptest.RequireEnv(t, "TEST_AZURE_USERNAME")
+
+	repoCfg := gptest.RepoConfig{
+		RepoURL:     repoURL,
+		Token:       token,
+		GitUsername: gitUsername,
+	}
+
+	prov, err := NewProvider(repoURL, &gitprovider.Options{Token: token})
+	require.NoError(t, err)
+
+	gptest.RunPRTests(t, repoCfg, prov, []gptest.PRTestCase{
+		{
+			Name:            "unspecified merge method",
+			ExpectedParents: 2, // Default is noFastForward (merge commit)
+		},
+		{
+			Name:            "noFastForward",
+			MergeMethod:     "noFastForward",
+			ExpectedParents: 2,
+		},
+		{
+			Name:            "rebase",
+			MergeMethod:     "rebase",
+			ExpectedParents: 1,
+		},
+		{
+			Name:            "rebaseMerge",
+			MergeMethod:     "rebaseMerge",
+			ExpectedParents: 2,
+		},
+		{
+			Name:            "squash",
+			MergeMethod:     "squash",
+			ExpectedParents: 1,
+		},
+		{
+			Name:           "invalid merge method",
+			MergeMethod:    "bogus",
+			ExpectMergeErr: true,
+		},
+	})
+}
--- a/pkg/gitprovider/bitbucket/pr_integration_test.go
+++ b/pkg/gitprovider/bitbucket/pr_integration_test.go
@@ -1,0 +1,38 @@
+//go:build integration && bitbucket
+
+package bitbucket
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/akuity/kargo/pkg/gitprovider"
+	gptest "github.com/akuity/kargo/pkg/gitprovider/testing"
+)
+
+func TestCreateAndMergePullRequest(t *testing.T) {
+	repoURL := gptest.RequireEnv(t, "TEST_BITBUCKET_REPO_URL")
+	token := gptest.RequireEnv(t, "TEST_BITBUCKET_TOKEN")
+
+	repoCfg := gptest.RepoConfig{
+		RepoURL:     repoURL,
+		Token:       token,
+		GitUsername: gptest.RequireEnv(t, "TEST_BITBUCKET_USERNAME"),
+	}
+
+	prov, err := NewProvider(repoURL, &gitprovider.Options{Token: token})
+	require.NoError(t, err)
+
+	gptest.RunPRTests(t, repoCfg, prov, []gptest.PRTestCase{
+		{
+			Name:            "unspecified merge method",
+			ExpectedParents: 2, // Repo default is merge commit
+		},
+		{
+			Name:           "explicit merge method",
+			MergeMethod:    "squash",
+			ExpectMergeErr: true, // Library limitation
+		},
+	})
+}
--- a/pkg/gitprovider/gitea/pr_integration_test.go
+++ b/pkg/gitprovider/gitea/pr_integration_test.go
@@ -1,0 +1,63 @@
+//go:build integration && gitea
+
+package gitea
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/akuity/kargo/pkg/gitprovider"
+	gptest "github.com/akuity/kargo/pkg/gitprovider/testing"
+)
+
+func TestCreateAndMergePullRequest(t *testing.T) {
+	repoURL := gptest.RequireEnv(t, "TEST_GITEA_REPO_URL")
+	token := gptest.RequireEnv(t, "TEST_GITEA_TOKEN")
+
+	repoCfg := gptest.RepoConfig{
+		RepoURL:     repoURL,
+		Token:       token,
+		GitUsername: gptest.RequireEnv(t, "TEST_GITEA_USERNAME"),
+	}
+
+	prov, err := NewProvider(repoURL, &gitprovider.Options{Token: token})
+	require.NoError(t, err)
+
+	gptest.RunPRTests(t, repoCfg, prov, []gptest.PRTestCase{
+		{
+			Name:            "unspecified merge method",
+			ExpectedParents: 2, // Kargo defaults to "merge"
+		},
+		{
+			Name:            "fast-forward-only",
+			MergeMethod:     "fast-forward-only",
+			ExpectedParents: 1,
+		},
+		{
+			Name:            "merge",
+			MergeMethod:     "merge",
+			ExpectedParents: 2,
+		},
+		{
+			Name:            "rebase",
+			MergeMethod:     "rebase",
+			ExpectedParents: 1,
+		},
+		{
+			Name:            "rebase-merge",
+			MergeMethod:     "rebase-merge",
+			ExpectedParents: 2,
+		},
+		{
+			Name:            "squash",
+			MergeMethod:     "squash",
+			ExpectedParents: 1,
+		},
+		{
+			Name:           "invalid merge method",
+			MergeMethod:    "bogus",
+			ExpectMergeErr: true,
+		},
+	})
+}
--- a/pkg/gitprovider/github/pr_integration_test.go
+++ b/pkg/gitprovider/github/pr_integration_test.go
@@ -1,0 +1,53 @@
+//go:build integration && github
+
+package github
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/akuity/kargo/pkg/gitprovider"
+	gptest "github.com/akuity/kargo/pkg/gitprovider/testing"
+)
+
+func TestCreateAndMergePullRequest(t *testing.T) {
+	repoURL := gptest.RequireEnv(t, "TEST_GITHUB_REPO_URL")
+	token := gptest.RequireEnv(t, "TEST_GITHUB_TOKEN")
+
+	repoCfg := gptest.RepoConfig{
+		RepoURL:     repoURL,
+		Token:       token,
+		GitUsername: gptest.RequireEnv(t, "TEST_GITHUB_USERNAME"),
+	}
+
+	prov, err := NewProvider(repoURL, &gitprovider.Options{Token: token})
+	require.NoError(t, err)
+
+	gptest.RunPRTests(t, repoCfg, prov, []gptest.PRTestCase{
+		{
+			Name:            "unspecified merge method",
+			ExpectedParents: 2, // Default is merge commit
+		},
+		{
+			Name:            "merge",
+			MergeMethod:     "merge",
+			ExpectedParents: 2,
+		},
+		{
+			Name:            "rebase",
+			MergeMethod:     "rebase",
+			ExpectedParents: 1,
+		},
+		{
+			Name:            "squash",
+			MergeMethod:     "squash",
+			ExpectedParents: 1,
+		},
+		{
+			Name:           "invalid merge method",
+			MergeMethod:    "bogus",
+			ExpectMergeErr: true,
+		},
+	})
+}
--- a/pkg/gitprovider/gitlab/pr_integration_test.go
+++ b/pkg/gitprovider/gitlab/pr_integration_test.go
@@ -1,0 +1,50 @@
+//go:build integration && gitlab
+
+package gitlab
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/akuity/kargo/pkg/gitprovider"
+	gptest "github.com/akuity/kargo/pkg/gitprovider/testing"
+)
+
+func TestCreateAndMergePullRequest(t *testing.T) {
+	repoURL := gptest.RequireEnv(t, "TEST_GITLAB_REPO_URL")
+	token := gptest.RequireEnv(t, "TEST_GITLAB_TOKEN")
+
+	repoCfg := gptest.RepoConfig{
+		RepoURL:           repoURL,
+		Token:             token,
+		GitUsername:       gptest.RequireEnv(t, "TEST_GITLAB_USERNAME"),
+		MergeWaitDuration: 10 * time.Second,
+	}
+
+	prov, err := NewProvider(repoURL, &gitprovider.Options{Token: token})
+	require.NoError(t, err)
+
+	gptest.RunPRTests(t, repoCfg, prov, []gptest.PRTestCase{
+		{
+			Name:            "unspecified merge method",
+			ExpectedParents: 2,
+		},
+		{
+			Name:            "merge",
+			MergeMethod:     "merge",
+			ExpectedParents: 2,
+		},
+		{
+			Name:            "squash",
+			MergeMethod:     "squash",
+			ExpectedParents: 2, // GitLab squash still produces a merge commit
+		},
+		{
+			Name:           "unsupported merge method",
+			MergeMethod:    "rebase",
+			ExpectMergeErr: true,
+		},
+	})
+}
--- a/pkg/gitprovider/testing/helpers.go
+++ b/pkg/gitprovider/testing/helpers.go
@@ -1,0 +1,276 @@
+//go:build integration
+
+// Package testing provides shared helpers for gitprovider integration tests.
+package testing
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/akuity/kargo/pkg/controller/git"
+	"github.com/akuity/kargo/pkg/gitprovider"
+)
+
+// RequireEnv returns the value of the named environment variable, or skips the
+// test if it is not set.
+func RequireEnv(t *testing.T, name string) string {
+	t.Helper()
+	v := os.Getenv(name)
+	if v == "" {
+		t.Skipf("%s must be set", name)
+	}
+	return v
+}
+
+// RepoConfig holds the provider-specific constants needed by the shared test
+// helpers.
+type RepoConfig struct {
+	RepoURL     string
+	Token       string
+	GitUsername string
+	// AuthedRepoURL is the clone URL with credentials embedded. Providers that
+	// embed credentials differently (e.g. Azure DevOps) must set this
+	// explicitly. If empty, it is derived as
+	// https://<GitUsername>:<Token>@<host>/<path>.
+	AuthedRepoURL string
+	// MergeWaitDuration is how long to wait after creating a PR before
+	// attempting to merge. Some providers (e.g. GitLab) need longer to compute
+	// mergeability. Defaults to 5 seconds if zero.
+	MergeWaitDuration time.Duration
+}
+
+func (c RepoConfig) clientOpts() *git.ClientOptions {
+	return &git.ClientOptions{
+		Credentials: &git.RepoCredentials{
+			Username: c.GitUsername,
+			Password: c.Token,
+		},
+		User: &git.User{
+			Name:  "Kargo Integration Test",
+			Email: "test@kargo.io",
+		},
+	}
+}
+
+func (c RepoConfig) authedRepoURL() string {
+	if c.AuthedRepoURL != "" {
+		return c.AuthedRepoURL
+	}
+	const prefix = "https://"
+	if !strings.HasPrefix(c.RepoURL, prefix) {
+		return c.RepoURL
+	}
+	return fmt.Sprintf(
+		"%s%s:%s@%s",
+		prefix, c.GitUsername, c.Token,
+		c.RepoURL[len(prefix):],
+	)
+}
+
+func (c RepoConfig) mergeWait() time.Duration {
+	if c.MergeWaitDuration > 0 {
+		return c.MergeWaitDuration
+	}
+	return 5 * time.Second
+}
+
+// PRTestCase defines a single test case for the PR integration test.
+type PRTestCase struct {
+	Name            string
+	MergeMethod     string
+	ExpectedParents int
+	ExpectMergeErr  bool
+}
+
+// RunPRTests is the shared test runner for gitprovider integration tests. It
+// exercises CreatePullRequest and MergePullRequest for each test case.
+func RunPRTests(
+	t *testing.T,
+	cfg RepoConfig,
+	prov gitprovider.Interface,
+	testCases []PRTestCase,
+) {
+	t.Helper()
+	ensureMainBranch(t, cfg)
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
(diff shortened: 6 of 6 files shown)


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
hiddeco, line 74: Can we add a section that documents which providers support what merge method?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -39,6 +39,7 @@ to merge will fail.
 | `provider`              | `string`  | N        | The name of the Git provider to use. Currently `azure`, `bitbucket`, `gitea`, `github`, and `gitlab` are supported. Kargo will try to infer the provider if it is not explicitly specified.                    |
 | `insecureSkipTLSVerify` | `boolean` | N        | Indicates whether to bypass TLS certificate verification when interfacing with the Git provider. Setting this to `true` is highly discouraged in production.                                                   |
 | `prNumber`              | `integer` | Y        | The pull request number to merge.                                                                                                                                                                              |
+| `mergeMethod`           | `string`  | N        | The merge method to use when merging the pull request. Options are `merge` (traditional merge commit), `squash` (squash all commits into one), or `rebase` (rebase and merge). If not specified, the default merge method configured on the repository will be used. |
```
hiddeco: Can we document that not every method is supported by each provider, referring to the section I proposed below?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -83,3 +84,18 @@ steps:
     prNumber: 42
     wait: true
 ```
+
+### Squash Merge
```
hiddeco: I think it would be better to call this "Using a merge method" and demonstrate that you can use it to "squash".

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -224,18 +224,25 @@ func (p *provider) MergePullRequest(
...
+		return nil, false, fmt.Errorf("pull request %d is closed but not merged", opts.Number)
 
 	case glMR.Draft || glMR.DetailedMergeStatus != "mergeable":
 		return nil, false, nil
 	}
 
 	// Merge the MR
+	mrOpts := &gitlab.AcceptMergeRequestOptions{}
+	// GitLab only supports Squash merging as a boolean
+	// Rebases have to be done via a different API call/manually
```
hiddeco: The way we would achieve this is by commenting `/rebase` to trigger the [quick action](https://docs.gitlab.com/user/project/quick_actions/#rebase) if we detect that the branch is out-of-date.

Can you add a note about this for future reference?

bobdoah:
> Thanks for implementing this, and I think it is a fair feature request. I made a couple of suggestions here and there.
> 
> My primary concern, other than what's commented, is that we silently ignore merge methods not supported by the selected provider. I think we should either:
> 
> 1. Return an error when a user specifies a merge method that the provider doesn't support
> 2. Log a warning so users understand their preference wasn't honored
> 
> I'd lean toward option 1, as it's typically better to fail explicitly than to silently do something different from what the user requested.

Hey, you're welcome and thanks for the comments. 

I had considered either of those two options. I didn't see how I might go about option 2. I guess I thought it would fail how it currently does. I can see how option 1 is preferable, and help users. 

Hopefully my latest changes address your comments. Happy to change/reword anything if you've got any other preferences!

bobdoah:
I've made some further changes as the code coverage drop was flagged as a failure.

hiddeco:
I think this is best combined with #5454 as I suspect the resolved SHA for GitLab will otherwise be wrong.

pkg/gitprovider/gitprovider.go
```diff
@@ -57,7 +69,7 @@ type Interface interface {
 	// - *PullRequest: the merged PR if successful
 	// - bool: true if merge was performed, false if PR is not ready to merge
 	// - error: only for actual errors (auth, network, invalid PR, etc.)
-	MergePullRequest(context.Context, int64) (*PullRequest, bool, error)
+	MergePullRequest(context.Context, *MergePullRequestOpts) (*PullRequest, bool, error)
```
krancour: Since PR number is always required, it should remain in the method signature instead of moving to the options struct.

```suggestion
	MergePullRequest(context.Context, int64, *MergePullRequestOpts) (*PullRequest, bool, error)
```

pkg/gitprovider/gitprovider.go
```diff
@@ -96,6 +108,16 @@ type ListPullRequestOptions struct {
 	BaseBranch string
 }
 
+// MergePullRequestOpts encapsulates the options used when merging a pull
+// request.
+type MergePullRequestOpts struct {
+	// Number is the pull request number to merge.
+	Number int64
```
krancour: See https://github.com/akuity/kargo/pull/5581/changes#r3001313609

```suggestion
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -39,19 +39,32 @@ system to access the git repos.
...
+
+The table below documents the supported merge methods for each of the currently supported providers.
+
+|           | Merge              | Squash             | Rebase             |
+|-----------|--------------------|--------------------|--------------------|
+| Azure     | :white_check_mark: | :white_check_mark: | :white_check_mark: |
+| BitBucket | :white_check_mark: | :x:                | :x:                |
+| Gitea     | :white_check_mark: | :white_check_mark: | :white_check_mark: |
+| GitHub    | :white_check_mark: | :white_check_mark: | :white_check_mark: |
+| GitLab    | :white_check_mark: | :white_check_mark: | :x:                |
```
krancour: I think it would be very important here to note what the default behavior is when unspecified. It's probably even worth specifying whether the default is imposed by the Git hosting provider or is one selected by Kargo.


```suggestion
|           | Merge              | Squash             | Rebase             | Default |
|-----------|--------------------|--------------------|--------------------|---|
| Azure     | :white_check_mark: | :white_check_mark: | :white_check_mark: | ? |
| BitBucket | :white_check_mark: | :x:                | :x:                | ? |
| Gitea     | :white_check_mark: | :white_check_mark: | :white_check_mark: | ? |
| GitHub    | :white_check_mark: | :white_check_mark: | :white_check_mark: | ? |
| GitLab    | :white_check_mark: | :white_check_mark: | :x:                | ? |
```

krancour: Actually, I see now in the code that at least some of these don't even have a default. BitBucket for instance is failing right away if the strategy isn't explicitly "merge." I'm not sure it make much sense to require "merge" to be explicitly specified when it's the only one supported. Let users leave it blank and we can apply defaults and complain only then the user explicitly opts for something invalid/unsupported.

pkg/gitprovider/azure/azure.go
```diff
@@ -334,3 +341,16 @@ func parseLegacyRepoURL(u *url.URL) (string, string, string, error) {
...
+	case gitprovider.MergeMethodMerge:
+		return adogit.GitPullRequestMergeStrategyValues.NoFastForward
+	case gitprovider.MergeMethodSquash:
+		return adogit.GitPullRequestMergeStrategyValues.Squash
+	case gitprovider.MergeMethodRebase:
+		return adogit.GitPullRequestMergeStrategyValues.Rebase
+	default:
+		return adogit.GitPullRequestMergeStrategyValues.NoFastForward
+	}
+}
```
krancour: This seems it could have been a map instead of a switch.

pkg/gitprovider/gitea/gitea.go
```diff
@@ -295,22 +295,38 @@ func (p *provider) MergePullRequest(
...
+			style = gitea.MergeStyleMerge
+		case gitprovider.MergeMethodSquash:
+			style = gitea.MergeStyleSquash
+		case gitprovider.MergeMethodRebase:
+			style = gitea.MergeStyleRebase
+		default:
+			style = gitea.MergeStyleMerge
+		}
+		mergeOpts.Style = style
+	}
```
krancour: Another case where a map might be a better approach.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
krancour: Noticing also that invalid strategies are being handled differently in different places. BitBucket fails outright, others fall back on their (undocumented) default.

We really need to make things more consistent.

pkg/gitprovider/github/github.go
```diff
@@ -314,31 +314,35 @@ func (p *provider) MergePullRequest(
...
+		return nil, false, fmt.Errorf("pull request %d is closed but not merged", opts.Number)
 
 	case ptr.Deref(ghPR.Draft, false) || !ptr.Deref(ghPR.Mergeable, false):
 		return nil, false, nil
 	}
 
 	// Merge the PR
+	prOpts := &github.PullRequestOptions{}
+	if opts.MergeMethod != "" {
+		prOpts.MergeMethod = string(opts.MergeMethod)
```
krancour: Is this possible because the string values we use to represent these strategies align perfectly with the ones GitHub uses? If so, it's worth saying that in a comment, otherwise the absence of any kind of logic/mapping here (which most of the others have) ends up looking suspect.

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -224,18 +224,30 @@ func (p *provider) MergePullRequest(
...
+	switch opts.MergeMethod {
+	case gitprovider.MergeMethodSquash:
+		// GitLab only supports Squash merging as a boolean
+		squash := true
+		mrOpts.Squash = &squash
+	case gitprovider.MergeMethodRebase:
+		// Rebases would be done by commenting /rebase to trigger
+		// the quick action if the branch is out-of-date
+		return nil, false, fmt.Errorf("unsupported merge method '%s' for provider", opts.MergeMethod)
+	}
```
krancour, line 250: Aren't there other options that are being ignored here?

pkg/gitprovider/azure/azure.go
bobdoah: Hopefully, I've addressed this in my changes.

pkg/gitprovider/gitea/gitea.go
bobdoah: Similarly addressed in the most recent changes

pkg/gitprovider/github/github.go
bobdoah: Yes, that's correct. That's a fair ask, I've added a comment saying as much.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
bobdoah: I've tried to align the providers, so that now none of them require a strategy and they all return an error if an invalid one is specified. 

The BitBucket API actually supports methods representative of the merge methods here. The library just lacks support. I've put in place a similar map to that used for Gitea and Azure. At the moment, it only supports the method merge. It will return an error for any other method. I've also [raised a PR on go-bitbucket](https://github.com/ktrysmt/go-bitbucket/pull/353), to add support for the merge strategy in the hope that this can eventually be implemented. 

I'm happy to revert to a simpler/dumber method for BitBucket, if this change is undesirable. 

I've added information about the default method for each provider to the table.

krancour:
I spent pretty much the entire day fiddling with this. I needed to see this working correctly against all five supported Git hosting providers. Here's what I uncovered:

1. The full range of supported merge methods/strategies is inconsistent across the providers. Trying to shoehorn all of what each can and cannot do into a uniform choice of merge/rebase/squash is going to create confusion as well as obscure access to some useful options. The right thing to do may be to accept any string value for merge method and, for the most part, simply pass that value through in the API call to the provider, allowing the provider to reject invalid options instead of us doing that ourselves. Overall this is a simplification. I have it done already in another branch.

2. Much more concerning... some of these providers, most of which were community contributions that perhaps didn't receive as much scrutiny as they ought have, are riddled with as-yet-unreported bugs that I've uncovered in my testing. I am reluctant to layer this new feature on top of that unstable foundation. Fortunately, I've been able to address many of these newly discovered issues. 

Even with all this in mind, I don't see any reason that we won't get this into v1.10.0, but I'm going to re-assign this to myself and amend this PR with a lot of bug fixes and integration tests in the morning.

krancour:
@fuskovic I've updated this PR with some refactoring and a slew of bug fixes. (Bugs that were present in main but actively interfered with attempts to integration test this PR.)

Individual commits have detailed descriptions of the changes within.

pkg/gitprovider/gitea/gitea.go
```diff
@@ -202,26 +149,21 @@ func (p *provider) CreatePullRequest(
 	if giteaPR == nil {
 		return nil, fmt.Errorf("unexpected nil pull request")
 	}
+	// TODO(krancour): Add label support. The Gitea SDK's AddIssueLabels expects
+	// label IDs ([]int64), but Kargo's CreatePullRequestOpts.Labels are names
+	// ([]string). A previous implementation attempted this but silently discarded
+	// the labels entirely. To fix properly: list repo labels, match by name to
+	// get IDs, then call AddIssueLabels.
```
fuskovic, line 156: https://github.com/akuity/kargo/issues/6021

pkg/gitprovider/bitbucket/bitbucket.go
```diff
@@ -355,15 +335,26 @@ func (p *provider) MergePullRequest(
...
 	// repositories.
 
-	// Attempt to merge the PR
+	// TODO(krancour): go-bitbucket does not expose the merge method/strategy as
+	// an option. The merge will always use the repository's default merge
+	// strategy, so simply fail if any merge strategy was specified. This is
+	// strictly a limitation of the go-bitbucket library and not the Bitbucket API
+	// itself, so this can be revisited in the future if the library adds support
+	// for this or if we decide to implement this API call ourselves instead of
+	// using the library.
```
fuskovic, line 344: https://github.com/akuity/kargo/issues/6022

pkg/gitprovider/bitbucket/bitbucket.go
```diff
@@ -58,11 +59,11 @@ func init() {
 
 // pullRequestClient defines the interface for pull request operations.
 type pullRequestClient interface {
-	CreatePullRequest(opt *bitbucket.PullRequestsOptions) (any, error)
-	ListPullRequests(opt *bitbucket.PullRequestsOptions) (any, error)
-	GetPullRequest(opt *bitbucket.PullRequestsOptions) (any, error)
-	GetCommit(opt *bitbucket.CommitsOptions) (any, error)
-	MergePullRequest(opt *bitbucket.PullRequestsOptions) (any, error)
+	Create(*bitbucket.PullRequestsOptions) (any, error)
+	Gets(*bitbucket.PullRequestsOptions) (any, error)
```
fuskovic, line 63: Should this be `List`?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
```diff
@@ -52,6 +53,19 @@ The `wait` option is unreliable for repositories hosted by Bitbucket due to API
...
+The table below documents the supported merge methods/strategies for each of the
+currently supported Git hosting providers.
+
+| Provider | Supported Methods | Default |
+| -------- | ----------------- | ------- |
+| Azure | <ul><li>`noFastForward`</li><li>`rebase`</li><li>`rebaseMerge`</li><li>`squash`</li></ul> | First allowed strategy per the target branch's merge type policy; merge commit if no policy is configured |
+| BitBucket | Client does not yet support specifying a merge strategy | The repository's configured default merge strategy |
+| Gitea | <ul><li>`fast-forward-only`</li><li>`manually-merged`</li><li>`merge`</li><li>`rebase`</li><li>`rebase-merge`</li><li>`squash`</li></ul> | `merge` |
+| GitHub | <ul><li>`merge`</li><li>`rebase`</li><li>`squash`</li></ul> | `merge` |
+| GitLab | <ul><li>`merge`</li><li>`squash`</li></ul> | Defers to the merge request and project-level squash settings |
```
fuskovic, line 67: The `Merge Method` in the examples section below uses the same `###` heading. Also I think this reference could be more useful if we moved it closer to the example.  If we did we could then just use a single `### Merge Method` heading. Thoughts?

pkg/gitprovider/azure/azure.go
```diff
@@ -232,27 +271,76 @@ func (p *provider) MergePullRequest(
...
+		}
+		if completedPR == nil {
+			return nil, false,
+				fmt.Errorf("unexpected nil pull request after merge of %d", id)
+		}
+		if ptr.Deref(completedPR.Status, "") ==
+			adogit.PullRequestStatusValues.Completed {
+			break
+		}
+		time.Sleep(time.Second)
```
fuskovic: Does it make sense to check for a cancelled context here?

```suggestion
    select {
    case <-ctx.Done():
        return nil, false, ctx.Err()
    case <-time.After(time.Second):
```

pkg/gitprovider/azure/azure_test.go
```diff
@@ -1,11 +1,405 @@
...
+	createPullRequestFn func(
+		context.Context, adogit.CreatePullRequestArgs,
+	) (*adogit.GitPullRequest, error)
+	getPullRequestFunc func(
+		context.Context, adogit.GetPullRequestArgs,
+	) (*adogit.GitPullRequest, error)
+	getPullRequestsFunc func(
+		context.Context, adogit.GetPullRequestsArgs,
+	) (*[]adogit.GitPullRequest, error)
+	updatePullRequestFn func(
```
fuskovic, line 29: Naming here looks inconsistent with the other funcs

pkg/gitprovider/azure/azure.go
```diff
@@ -68,13 +102,19 @@ func NewProvider(
...
 
 	return &provider{
-		org:        org,
-		project:    project,
-		repo:       repo,
-		connection: connection,
+		org:     org,
+		project: project,
+		repo:    repo,
+		client:  client,
```
krancour, line 121: Note about this change: connection == hard to mock --> poorly tested; client == easy to mock --> tests added

pkg/gitprovider/bitbucket/bitbucket.go
```diff
@@ -58,11 +59,11 @@ func init() {
...
+	Get(*bitbucket.PullRequestsOptions) (any, error)
+	GetCommit(*bitbucket.CommitsOptions) (any, error)
+	Merge(*bitbucket.PullRequestsOptions) (any, error)
```
krancour, line 66: The different names for this interface's methods were the direct reason for a lot of unnecessary code in the wrapper struct.

pkg/gitprovider/bitbucket/bitbucket.go
```diff
@@ -374,6 +365,11 @@ func (p *provider) MergePullRequest(
 		return nil, false, fmt.Errorf("error parsing merged pull request response: %w", err)
 	}
 
+	if mergedBBPR.State != prStateMerged {
+		return nil, false,
+			fmt.Errorf("merge rejected for pull request %d (state: %s)", id, mergedBBPR.State)
+	}
```
krancour, line 379: Note to self to double-check this. Not mergeable is most likely a temporary condition and probably shouldn't be surfaced as an actual error. I think the one caller (git-merge-pr's step runner) will treat an error as a terminal condition while `false` without an error would eventually lead to a retry in a future reconciliation step.

pkg/gitprovider/bitbucket/bitbucket.go
krancour: imho, the name is unfortunate, but the methods on the interface need to match what the implementation already has or else we're back to having to manually write pass-through methods for the wrapper.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-merge-pr.md
krancour: I focused mainly on doc accuracy and didn't challenge @bobdoah's original formatting choices much. But agree it deserves a second look.

pkg/gitprovider/azure/azure_test.go
krancour: Will double-check.

pkg/gitprovider/azure/azure.go
```diff
@@ -68,13 +102,19 @@ func NewProvider(
 		return nil, err
 	}
 	organizationUrl := fmt.Sprintf("https://%s/%s", modernHostSuffix, org)
-	connection := azuredevops.NewPatConnection(organizationUrl, opts.Token)
+	client, err := adogit.NewClient(
+		context.Background(),
```
krancour, line 110: Note to self: Add a comment explaining the choice to use background context here.

pkg/gitprovider/azure/azure.go
```diff
@@ -232,27 +271,76 @@ func (p *provider) MergePullRequest(
...
+		if ptr.Deref(completedPR.Status, "") ==
+			adogit.PullRequestStatusValues.Completed {
+			break
+		}
+		time.Sleep(time.Second)
+	}
+	if ptr.Deref(completedPR.Status, "") !=
+		adogit.PullRequestStatusValues.Completed {
+		return nil, false,
+			fmt.Errorf("pull request %d did not complete after merge", id)
```
krancour: Note to self: Similar to [here](https://github.com/akuity/kargo/pull/5581/changes#r3023714257), this deserves a second look.

pkg/gitprovider/gitea/gitea.go
```diff
@@ -40,41 +51,29 @@ func init() {
...
-		ctx context.Context,
 		owner string,
 		repo string,
-		number int,
-		labels []string,
-	) ([]*gitea.Label, *gitea.Response, error)
+		number int64,
+		opts gitea.MergePullRequestOption,
+	) (bool, *gitea.Response, error)
 }
```
krancour, line 77: This is another case where more carefully defining the interface to match the existing client implementation relieves us of writing client wrapper methods -- in fact, in this case, we can get away without even having a client wrapper at all.

pkg/gitprovider/gitea/gitea.go
```diff
@@ -122,74 +121,22 @@ func NewProvider(
...
-}
-
-func (g giteaClientWrapper) MergePullRequest(
-	_ context.Context,
-	owner string,
-	repo string,
-	number int,
-	opts *gitea.MergePullRequestOption,
-) (*gitea.Response, error) {
-	_, resp, err := g.client.MergePullRequest(owner, repo, int64(number), *opts)
```
krancour, line 167: The `_` was dropping important information on the floor!

pkg/gitprovider/gitea/gitea.go
```diff
@@ -301,14 +248,29 @@ func (p *provider) MergePullRequest(
...
+		p.repo,
+		id,
+		gitea.MergePullRequestOption{Style: gitea.MergeStyle(mergeMethod)},
+	)
+	if err != nil {
 		return nil, false, fmt.Errorf("error merging pull request %d: %w", id, err)
 	}
+	if !merged {
+		return nil, false, fmt.Errorf("merge rejected for pull request %d", id)
+	}
```
krancour, line 271: Similar to [here](https://github.com/akuity/kargo/pull/5581/changes#r3023714257), this needs another look.

pkg/gitprovider/github/github.go
```diff
@@ -326,15 +331,19 @@ func (p *provider) MergePullRequest(
...
 	if err != nil {
 		return nil, false, fmt.Errorf("error merging pull request %d: %w", id, err)
 	}
 	if mergeResult == nil {
 		return nil, false, fmt.Errorf("unexpected nil merge result")
 	}
+	if !ptr.Deref(mergeResult.Merged, false) {
+		return nil, false,
+			fmt.Errorf("merge rejected for pull request %d", id)
+	}
```
krancour, line 346: Same as [here](https://github.com/akuity/kargo/pull/5581/changes#r3023714257)

pkg/gitprovider/gitlab/gitlab.go
```diff
@@ -230,9 +236,20 @@ func (p *provider) MergePullRequest(
 		return nil, false, nil
 	}
 
-	// Merge the MR
+	// Merge the MR. When unspecified, the GitLab API uses a merge commit. Squash
+	// is supported via a boolean flag. Rebase is not supported through the
+	// accept-merge-request API.
```
krancour: Comment can be worded better. The bit about a "rebase" isn't especially relevant anymore now that we've done away with the ternary choice of merge/squash/rebase.

pkg/gitprovider/azure/azure.go
krancour: In this case, we've now moved away from an error to allow the merge to be "retried" later. Short circuiting logic further up should stop a second merge from _actually_ being attempted, unless the async PR completion _still_ isn't complete at that point. If it isn't, there may be a failure, and at that point, it should be expected. The important thing here is we're providing ample opportunity for async PR completion post-merge-request and there is no possibility of an entirely silent failure.

pkg/gitprovider/gitea/gitea.go
krancour: The `false` on line 269 is a clear indicator that the API server did indeed reject the merge -- for whatever reason. An error is the correct behavior here. It's important to note the mergeability checks that are applied earlier in this method. If this request was rejected it _wasn't_ because of a transient "not yet mergeable" condition.

pkg/gitprovider/github/github.go
krancour: As far as I can tell, this shouldn't ever be false after the request to merge returned 200, so this is a truly unexpected condition and does amount to an error.

pkg/gitprovider/bitbucket/bitbucket.go
krancour: As best I can tell, this is correct. As with GitHub, the API docs lead me to believe that if the request was successful, this should be true. If it's false, that's genuinely unexpected and a problem. It's possible we don't even _really_ need to check this. But if this were to happen, it is definitely an error condition and no a scenario where the caller should retry.

krancour:
Ok... I believe no stone's been left unturned at this juncture. Integration testing this exposed a half dozen pre-existing provider bugs that we've squashed as part of this PR. We opened follow-up issues for anything remaining that can still be improved upon.

Thanks @bobdoah for having done all the initial legwork on this. Thanks again, @fuskovic for the reviews.

bobdoah:
> Ok... I believe no stone's been left unturned at this juncture. Integration testing this exposed a half dozen pre-existing provider bugs that we've squashed as part of this PR. We opened follow-up issues for anything remaining that can still be improved upon.
> 
> Thanks @bobdoah for having done all the initial legwork on this. Thanks again, @fuskovic for the reviews.

Amazing! Thanks for taking this on.

PR #5586, opened 2026-01-15
feat: new configmap endpoints

krancour:
These were requested by @rpelczar in support of ongoing UI work.

Typically, the UI creates and updates resources using generic endpoints, but because ConfigMaps, like credentials (Secrets), but _unlike_ most other resources need to be managed at all three of project-level, system-level, and shared, dedicated endpoints that understand the distinction between those are a good idea. _Without this the UI or any other client would need to know the names of system and shared resources namespaces in order to make use of the generic resource endpoints, and those are details I prefer clients not need to concern themselves with.


pkg/server/create_config_map_v1alpha1.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func (s *server) CreateConfigMap(
+	ctx context.Context,
+	req *connect.Request[svcv1alpha1.CreateConfigMapRequest],
+) (*connect.Response[svcv1alpha1.CreateConfigMapResponse], error) {
+	configMap := req.Msg.ConfigMap
+	if configMap == nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument,
+			fmt.Errorf("config_map is required"))
```
fuskovic: Since there's no formatting directive here we should probably just opt for `errors.New`

```suggestion
			fmt.Errorf("config_map is required"))
```

PR #5609, opened 2026-01-20
fix: `.git` repoURL suffix causing repo not found errors

likelymichael:
upon updating to Kargo v1.8.4, we started noticing an error with our `git-clone` step that would take a repoURL formatted like `https://github.com/org/repo.git` and it would fail. we were on v1.7.4 before and did not have this issue. we traced the issue to this [new function](https://github.com/akuity/kargo/pull/4995/changes#diff-daea05af1594afbd1a948080013c72ae2cb3f44170da2756a0d41b6266a704fbR276)  in this [PR](https://github.com/akuity/kargo/pull/4995). 

this change strips `.git` suffix from the extracted repo URL so there is no mismatch to cause the "repo not found" error we've been getting.

### DIFF
--- a/pkg/credentials/github/app_test.go
+++ b/pkg/credentials/github/app_test.go
@@ -26,7 +26,7 @@ func TestNewAppCredentialProvider(t *testing.T) {
 func TestAppCredentialProvider_Supports(t *testing.T) {
 	p := NewAppCredentialProvider()
 
-	const testRepoURL = "https://github.com/example/repo.git"
+	const testRepoURL = "https://github.com/example/repo"
 	// This is a control. Each test case will tweak a clone of this supported map.
 	supportedDataMap := map[string][]byte{
 		clientIDKey:       []byte("client"),
@@ -178,6 +178,17 @@ func TestAppCredentialProvider_Supports(t *testing.T) {
 			},
 			expected: true,
 		},
+		{
+			name:     "valid with .git suffix",
+			credType: credentials.TypeGit,
+			repoURL:  testRepoURL + ".git",
+			getDataMap: func() map[string][]byte {
+				dm := maps.Clone(supportedDataMap)
+				delete(dm, clientIDKey)
+				return dm
+			},
+			expected: true,
+		},
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/credentials/github/app.go
```diff
@@ -280,7 +280,7 @@ func (p *AppCredentialProvider) extractRepoName(repoURL string) string {
 	if len(parts) < 5 {
 		return ""
 	}
-	return parts[len(parts)-1]
+	return strings.TrimSuffix(parts[len(parts)-1], ".git")
```
krancour, line 283: I am concerned that this fix probably swept a larger problem under the rug.

The function's documentation states it assumes the provided `repoURL` has been normalized by the caller.

A normalized git URL would already have had any trailing ".git" removed.

This means some URL wasn't normalized prior to this where it should have been.

What else might be broken on account of that?

likelymichael: why would a normalized git URL have the `.git` removed? it is coming straight through from the request it looks like: https://github.com/likelymichael/kargo/blob/2f0a744c3e623ebfabaa4dc5720aa610ec38369d/pkg/credentials/github/app.go#L95

krancour: > why would a normalized git URL have the .git removed?

That's just how our normalization works. The point of normalization is that for two different but _equivalent_ URLs x and y, normalized(x) == normalized(y). This is crucial for comparisons of URLs. Normalization could equally have worked by ensuring the _presence_ of a trailing `.git` instead of its absence. We inherited the basic normalization algorithm from Argo CD and ensuring the absence of the trailing `.git` was just how that worked and so it's worked the same for Kargo from day one.

> it is coming straight through from the request

If you follow the code from a higher level, you'll find that the URL is normalized quite early in the process of searching for credentials. The URL in that request object _is_ normalized or _should have been_.

So what I'm concerned with is how something that doesn't look normalized snuck through. After quite a bit of digging, here's my best guess:

By design, normalization has no effect on a URL it cannot parse. I believe your URL, although appearing valid, couldn't be parsed for some reason. Looking at the normalization code, the only way I can come up with that such a thing could happen is if a URL accidentally contained leading whitespace _or_ any invisible characters, such as a BOM.

If you are able to check into that on your side, it could prove or disprove this theory.

In short, I think all signs are pointing the the real bug being a normalization process that was inadequately defensive about trimming whitespace and other invisible characters.

krancour: Whether it was the root cause of your issue or not, it definitely _is_ a bug. I opened a separate issue for it: https://github.com/akuity/kargo/issues/5616

likelymichael: Gotcha. Thanks for creating the issue! Should I submit a revert PR, or how do you want to handle this?

krancour: This change is harmless, so we can leave it for now and I will ask @fuskovic to reverse it in #5637

PR #5611, opened 2026-01-20
feat: Add Webhook path filtering

### LINKED ISSUE #5591: Webhook receiver filter on includePaths
# Checklist

```
kargo version                                                                                                                                                                                                                                                                                                                                                                                
Client Version: v1.8.4
Server Version: v1.8.1
```

# Proposed Feature

Add Warehouse filtering on `includePaths` to webhook receivers for ClusterConfig webhooks.

# Motivation

We store our application configuration in a directory called `apps/` in a monorepo. We have plans to migrate more applications to this repository and might arrive at 150+ applications. Currently we have two warehouse configured:
* ECR Warehouse (not relevant per this discussion)
* Git warehouse 
```
spec:
  subscriptions:
    - git:
        repoURL: https://github.com/<owner>/<repo>.git
        commitSelectionStrategy: NewestFromBranch
        includePaths:
          - apps/foo
```

According to the documentation [here](https://docs.kargo.io/operator-guide/cluster-configuration/#receivers-in-action), 
```
A webhook receiver's only job is to extract a repository URL from the webhook request's payload, query for all Warehouse resources across all Projects having subscriptions to that repository, and request each to execute their artifact discovery process.
```
This means that **ALL** Warehouses subscribed to this repository will get triggered to perform an update. 

I would like to see filtering added to the Webhook receiver to notify only specific Warehouses that match the correct `includePaths` from the payload.

I checked that the Github `push` event does include the files:
```
type PushEvent struct {
    // ... other fields
    HeadCommit   *HeadCommit `json:"head_commit,omitempty"`
    Commits      []*HeadCommit `json:"commits,omitempty"`
}

type HeadCommit struct {
    Added     []string `json:"added,omitempty"`
    Removed   []string `json:"removed,omitempty"`
    Modified  []string `json:"modified,omitempty"`
    // ... other fields
}
```

Work would entail:
1. Extract file paths from the webhook payload (github.go:215-225)
```
case *gh.PushEvent:
    repoURLs = []string{urls.NormalizeGit(e.GetRepo().GetCloneURL())}
    ref := e.GetRef()

    // NEW: Extract file paths from HeadCommit
    var filePaths []string
    if e.HeadCommit != nil {
        filePaths = append(filePaths, e.HeadCommit.Added...)
        filePaths = append(filePaths, e.HeadCommit.Modified...)
        filePaths = append(filePaths, e.HeadCommit.Removed...)
    }
```
2. Update `shouldRefresh` in (refresh.go 148-179)
  *  add file path matching
  * check if modified files match `includePaths`
  * respect the `excludePaths`
3. Pass file paths through call chain:
  * update `refreshWarehouses` signature to accept file paths
  * update `shouldRefresh` to evaluate path patterns
4. Add documentation

This would greatly reduce the Warehouse refresh. 

The only other way I can think of tackling this is to cache the github repository and use that across multiple warehouses so that the Github API queries are reduced and Warehouses that share a repository can detect changes instantly.

kwoodson:
# Add webhook-time path filtering for GitHub push events

## Problem

Closes #5591

Currently, when a webhook is received from GitHub for a repository with multiple warehouses (common in monorepo scenarios), **all** warehouses subscribed to that repository are marked for refresh, regardless of their `includePaths` and `excludePaths` configuration. Path filtering only occurs later during the warehouse reconciliation phase when commits are being selected.

This creates unnecessary load in monorepo environments with many warehouses (e.g., 100-150+ warehouses). For example, changing a single file in `apps/project-a` would trigger refresh for all 150 warehouses, even though 149 of them have path filters that would ultimately exclude the change.

## Solution

This PR implements webhook-time path filtering for GitHub push events by:

1. **Extracting file paths from webhook payload**: GitHub `PushEvent` includes a `HeadCommit` object with `Added`, `Modified`, and `Removed` file arrays. We now extract and use these paths.

2. **Filtering warehouses before marking for refresh**: In the `refreshWarehouses()` flow, we now evaluate each warehouse's `includePaths` and `excludePaths` against the changed files before marking it for refresh.

3. **Reusing existing pattern matching logic**: The implementation uses the existing `pkg/pattern.Matcher` to ensure consistency with the path filtering that occurs during commit selection.

### Core Implementation
- **pkg/webhook/external/github.go**: Extract file paths from `HeadCommit` and pass to `refreshWarehouses()`
- **pkg/webhook/external/refresh.go**: 
  - Updated `refreshWarehouses()` and `shouldRefresh()` to accept and evaluate `filePaths`
  - Added `matchesPathFilters()` function using existing pattern matching logic
  - Enhanced logging to show filtering decisions (Debug level for skipped, Info for matched)

### Testing
- **pkg/webhook/external/path_filter_test.go**: 13 test cases for path matching logic
- **pkg/webhook/external/refresh_test.go**: Added `TestShouldRefreshWithPathFilters` with 6 test cases covering:
  - Matching and non-matching paths
  - Excluded paths
  - Multiple paths scenarios
  - No path filters (backward compatibility)
  - Ref matching combined with path filtering
- **pkg/webhook/external/github_test.go**: Added 2 integration tests for end-to-end webhook handling with path filtering

## Performance Impact

**Real-world testing** in a monorepo with 5 warehouses showed:
- **Before**: All 5 warehouses refreshed on every push
- **After**: Only 1 warehouse refreshed (80% reduction)

For environments scaling to 100-150 warehouses, this represents significant reduction in:
- Kubernetes API calls
- Git repository queries
- Overall controller load

## Backward Compatibility

✅ Fully backward compatible:
- Warehouses without `includePaths`/`excludePaths` behave exactly as before (all changes trigger refresh)
- Webhooks without file path information (e.g., from older GitHub versions) work as before
- Non-GitHub webhooks (GitLab, Bitbucket, etc.) are unaffected

## Future Work

- GitLab webhooks also include file paths in `PushEvent.Commits[].Added/Modified/Removed` and could benefit from similar filtering
- Bitbucket and Azure DevOps webhooks don't include file path information in their payloads

## Testing

All tests pass:
```
✅ TestMatchesPathFilters (13 test cases)
✅ TestShouldRefreshWithPathFilters (6 test cases)  
✅ TestGithubHandler (25 test cases including 2 new path filtering tests)
```

Verified in live cluster:
- Only warehouses with matching path filters are refreshed
- Debug logs show filtering decisions for troubleshooting
- Timestamps confirm only targeted warehouses updated

### DIFF
--- a/pkg/webhook/external/github.go
+++ b/pkg/webhook/external/github.go
@@ -219,13 +219,31 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc 
 			ref := e.GetRef()
 			qualifiers = []string{ref}
 			logger = logger.WithValues("ref", ref)
-			commits := e.Commits // nolint: staticcheck
-			if e.HeadCommit != nil {
-				commits = append(commits, e.HeadCommit)
+			// GitHub includes a max of 2048 commits in the push event payload. If
+			// there are more than that, you need to use the GitHub API to retrieve
+			// the full list of commits. We're not going to do that. Instead, for such
+			// a case, we'll let the list of changed files remain empty, the effect of
+			// which will be that all Warehouses subscribed to the repository will be
+			// refreshed instead of just the ones whose filters match the changed
+			// files. This is good enough for this edge case. So we proceed only if
+			// there are fewer than 2048 commits in the payload.
+			//
+			// Also note: e.Commits is deprecated with a note that GitHub will be
+			// removing commit details from event payloads beginning October 7, 2025,
+			// however, as of this writing (March 18, 2026) GitHub is still including
+			// commit details in the payloads. GitHub's own API documentation for the
+			// push event does not indicate these are deprecated.
+			//
+			// See: https://docs.github.com/en/webhooks/webhook-events-and-payloads#push
+			if len(e.Commits) < 2048 { // nolint: staticcheck
+				commits := e.Commits // nolint: staticcheck
+				if e.HeadCommit != nil {
+					commits = append(commits, e.HeadCommit)
+				}
+				changedFiles = collectPaths(commits, func(c *gh.HeadCommit) []string {
+					return slices.Concat(c.Added, c.Modified, c.Removed)
+				})
 			}
-			changedFiles = collectPaths(commits, func(c *gh.HeadCommit) []string {
-				return slices.Concat(c.Added, c.Modified, c.Removed)
-			})
 		case *gh.RegistryPackageEvent:
 			action := e.GetAction()
 			switch action {


pkg/webhook/external/refresh.go
```diff
@@ -226,3 +260,63 @@ func shouldRefresh(
...
+	matchers := make(pattern.Matchers, len(selectors))
+	for i := range selectors {
+		matcher, err := pattern.ParsePathPattern(selectors[i])
+		if err != nil {
+			return nil, fmt.Errorf("parse error path selector %q: %w", selectors[i], err)
+		}
+		matchers[i] = matcher
+	}
+	return matchers, nil
+}
```
fuskovic: This duplicates existing logic in `pkg/controller/git/commit`. 

I suggest exporting the existing logic for reuse to keep things DRY.

kwoodson: I have updated to reuse the commit logic to keep things `DRY`.

krancour:
I took a cursory look at this. I'm unopposed to the notion, but think it may be achievable with fewer changes. I will take a closer look at that as soon as I'm able.

kwoodson:
@krancour Thanks! Happy to see what you come up with.

kwoodson:
@krancour Hello! Any ideas on this one?

kwoodson:
@krancour Gently - bump on this!

I know you and the team have been busy with the v1.9.0 release. We've rolled this out and are enjoying the latest!

Thanks!

jmmclean:
The performance impact from refreshing all warehouses when using a monorepo is not great. I mistakenly created a PR to do basically the same thing https://github.com/akuity/kargo/pull/5866. Thank you for making this PR @kwoodson  - My searching foo is not great, but we both have the same concerns!

It also seems like your changes will only support `github` and will not support `gitea` and `gitlab` (I happen to be gitlab). One other criticism is that if there are multiple commits in a PR, then your implementation will only fetch the headcommit and will miss file changes in other commits.

@krancour and @fuskovic Is it worth taking a deeper look at !5866 as I have addressed iterating over commits as well as supporting gitea and gitlab?

kwoodson:
>  My searching foo is not great, but we both have the same concerns!

Heh, np!

> It also seems like your changes will only support github and will not support gitea and gitlab (I happen to be gitlab). 

@jmmclean I did not implement `gitea` or `gitlab` as I do not have those environments in which to test. I believe these are worth creating as in your case. I am happy to augment the code to include those two cases but would require someone to test the latest.

I'd be happy to collaborate. This feature is **VERY** important to us.

krancour:
@kwoodson I need to ask for some further patience. I understand this is important to you, but our team has its own priorities, many of which are internal and not visible to the community. _We_ want this feature as well and have already included it to the v1.10.0 milestone -- which is the next release. So please understand we do expect to get to this soon, but we're not putting down all the other higher priority things we have going on to work on this.

kwoodson:
@krancour Thanks for the reply. I completely understand that you have priorities and will continue working on what you believe is important for the success of the project.

@jmmclean has a few enhancements around this subject and has added them to his PR that I did not include. Support for `gitea` and `gitlab` in his PR in #5866 . 

What would your recommendation be for this PR? I have tested this specific fix and verified in our environment. Would you like for me to include @jmmclean's updates inside of this PR (doc updates, gitea, gitlab support, multi-commit support)? Would you like to consider his work?

I'd be happy to help test and or combine efforts.

Thanks for your time and consideration. Please advise.

jmmclean:
Before merging this, can we consider adding support for Gitlab and Gitea? per @kwoodson comment below

> I have tested this specific fix and verified in our environment. Would you like for me to include @jmmclean's updates inside of this PR (doc updates, gitea, gitlab support, multi-commit support)? Would you like to consider his work?

fuskovic:
> Before merging this, can we consider adding support for Gitlab and Gitea? per @kwoodson comment below

Support for Gitlab and Gitea has already been added and tested

krancour:
@kwoodson sorry it took us so long to get serious about this.

@fuskovic thank you for the review.

This mostly looks good to me. My biggest concern was how internals of the commit selectors were bleeding through into other packages. I amended the PR with a few commits to address that concern.

I'm now pretty happy with this.

I do have one question about some implementation details, and I'll pose that in-line with the changes.

pkg/webhook/external/github.go
```diff
@@ -217,7 +219,13 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 			ref := e.GetRef()
 			qualifiers = []string{ref}
 			logger = logger.WithValues("ref", ref)
-
+			commits := e.Commits // nolint: staticcheck
```
krancour: @kwoodson this is my one remaining concern. According to the deprecation notice on this field, commit summaries were to have been removed from event payloads on October 7, 2025.

I do not know if GitHub followed through on that or not, but that is what is alleged by the deprecation notice.

krancour: Although GitHubs own docs do not say this is deprecated:

https://docs.github.com/en/webhooks/webhook-events-and-payloads#push

They do, however, say there's a max of 2048 commits in the field and that if there are more than that you need to use the GitHub API to retrieve the details.

Perhaps we ought to just force a refresh if we see 2048 commits there.

krancour: cc @fuskovic

kwoodson: @krancour thanks for reviewing.

GitHub’s current webhook docs still list `commits` on the push payload and do not mark it deprecated. They also say the field contains up to 2048 commits:
https://docs.github.com/en/webhooks/webhook-events-and-payloads#push

Given that cap, I agree the safest behavior is to treat `len(e.Commits) >= 2048` as potentially incomplete path data and fall back to ref-only refresh behavior for that event. That would preserve current behavior while still allowing path filtering for normal-sized pushes.

krancour: https://github.com/akuity/kargo/pull/5611/commits/6aff8db4f280cce9604311bcfbe2452a39d98c45

PR #5614, opened 2026-01-21
feat: implement REST api

krancour:
## Why?

The short version of this sordid affair:

Beginning with Kubernetes 1.35 it is difficult to represent Kubernetes resource types, including our own custom resource types as protocol buffers. Beginning with Kubernetes 1.36, doing so is essentially impossible. This very disruptive change eliminates any realistic possibility of continuing to keep our dependencies on various Kubernetes modules up to date unless/until such time that we transition away from our protocol buffer-based service/v1alpha1 API, which is the UI and CLI's sole means of communication with Kargo's back end.

On security grounds, and for a host of other reasons, we cannot abide no longer keeping our Kubernetes dependencies up to date. This means the time has come to part ways with Connect RPC (which we haven't exactly loved anyway).

This PR is the first phase of a plan that transitions the Kargo API to a RESTful one (or _RESTish_ one, at least) over the course of three release.

## The Plan

1. **Kargo v1.9.0 introduces a RESTful API, but continues serving up the legacy, Connect-based API.**
    * **The Kargo UI will continue to use the legacy, Connect-based API.**
    * **The Kargo CLI will immediately transition to utilizing the new API.**

      Unrelated changes in Kargo v1.9.0 were already set to force adopters of Kargo v1.9.0 to upgrade their CLI. The team has decided to capitalize on that fact and transition the CLI to the new API _now_. By doing so sooner rather than later, we hope to reduce the likelihood of another forced CLI upgrade a few releases from now when the Connect-based API is eventually removed. We will also feel better about the UI's eventual transition to the new API if any issues are shaken down by the CLI first.

2. **Kargo v1.10.0 will, at least partially, transition the Kargo UI to the new RESTful API, but will continue serving up the legacy, Connect-based API.**

    The choice to continue serving the legacy API through v1.10.x is intended to buy time for any users who have built their own integrations with Kargo via that API.

3. **Kargo v1.11.0 will remove the legacy, Connect-based API.**

   The UI will be fully transitioned to the new API.

   Removal of the old API will permit us to resume freely upgrading dependencies on Kubernetes modules.

## The Nitty Gritty Details

Endpoints of the new API will be annotated with Swagger[https://github.com/go-swagger/go-swagger] annotations that permit generation of a `swagger.json` specification. From that specification, API bindings (clients) can be generated for many different languages.

For the benefit of Go code that must integrate with our new API, namely Kargo's own CLI, we will use (swagger)[https://github.com/swaggo/swag] to generate that client, which will be checked into our repository (and usable by other projects).

A Javascript-based client will be generated using [orval](https://orval.dev/) and checked into our repository. This will initially be unused by Kargo itself, but exists as a preview for the benefit of Kargo's UI team who will undertake the work of transitioning the UI to the new API in the v1.10.0 timeframe.

**It is prudent to note that these generated clients account for the bulk of the 92k+ new lines of code included in this PR.**

All of the above accounts only for how API endpoints are formally documented and how clients are generated from that formal documentation. None of it speaks to the actual runtime _implementation_ of the RESTful API. Turning our attention to that, the primary capability Kargo requires that it lacked before is request-routing that supports the use of _path parameters_. Many viable options exist. For now, we have settled on utilizing the [Gin](https://gin-gonic.com/) framework for this purpose. We'll also be using Gin to implement authentication and error-handling middleware.

Importantly, if the team decides for any reason to transition from Gin to some other framework in the future (or shun frameworks entirely), the refactoring would be non-breaking, and mechanical enough in nature so as to be easily accomplished by an AI agent. All this is to say, it is not written in stone that Gin will be used indefinitely.

### Compromises

_There must be some, right?_

Of course, but they're not too bad (imho).

1. Swagger does not deal well with Kubernetes resource types, including our custom resource types.

    To work around this, anywhere that a Kubernetes resource type is an input to, or included in the output of, a RESTful API endpoint, the generated documentation will formally state those fields are of type `any`, but with clarifying remarks as to what the type _really_ is. Any client generated from that documentation will see such a field as an undifferentiated chunk of JSON. Users of such a client will need to be savvy enough to know what types that JSON can be unmarshaled into. In the case of Go code, like that in the CLI, we're aware of the various implementations of `client.Object` into which that JSON can be unpacked.

2. Swagger does not deal well with streaming.

    Several endpoints in the legacy, Connect-based API support streaming. Their RESTful replacements shall as well (using SSE), however, those streaming capabilities will not be formally documented with Swagger. The consequence of this is that clients generated from that documentation will not support that small handful of endpoints (eight).
    
    Due to the small number of affected endpoints, we will tolerate the necessity of custom, "hand-written" Go and JS clients for interaction with those endpoints. In actuality, creating and maintaining that custom client is a perfect task for an AI agent.

### Execution

Let's get down to what's actually in this PR.

First and foremost, it executes on everything above.

Since porting the existing endpoints to RESTful ones was a _massive_ volume of work, this section is essentially a guide to reviewers, since the review task is _also_ mammoth.

1. Anywhere you see "v2" in a path, please understand it to be a placeholder. We will discuss proper versioning of this during one of our weekly meetings.

2. As previously noted, much of the 92k+ new lines of code are generated clients. These additions were deliberately their own PRs so that those changes can be more easily dismissed at review-time. They are confined to:

    * 921fa8dadecc1270eadda90aa8fb2e38a765400d (Go)
    * 0d8d47b7f4093d2d8f101c4e99913788e8c9e5b3 (JS)

4. 7d789eff61bcf1f6b0ed41dbfeb91f536aee2886 implements the code generation process that produced the above. This is worth looking at even if its output is less so.

3. Implementation of the new API (and nothing else) is in 2d67234b7135b89e25f352a0032d931bb61406d4.

    This was a large volume of work. It was impractical to undertake without the assistance of an AI agent. It's important to note that I spent two weeks guiding that agent toward necessary course corrections, reviewing and tweaking its output myself, and testing new endpoints end-to-end. So a lot of this was done by AI, but it's not slop.
    
    All endpoints --  every single one of them -- have tests now. This was not true for the legacy, Connect-based API. The tests do not invoke endpoint functions directly. They deliberately involve the Gin router so that all the routing, extraction of parameters from request paths, etc. is all put to the test along with the endpoint functionality.
    
4. b0f7014971f8f24b7c468cfaaad1f729b0da388c is the "hand-written" Go client for streaming endpoints.

    This is an area that I did not personally scrutinize as closely as the rest.

5. 42164cf8ab58dfc66e86ce810a924a9a9b715e48 refactors the CLI to use the new API exclusively.

    This also is a large volume of work accomplished with a well-attended AI agent.

6. 0d8d47b7f4093d2d8f101c4e99913788e8c9e5b3 adds the generated, "preview" JS client.

    @rpelczar and @Marvin9 should weight in on its acceptability.

7. e52c04d47b91967413bdeb8d169c26a4606af5c3 contains e2e tests for the CLI + API.

8. Further notes:

    * I will add my own comments in-line on any changes I want to call specific team members' attention to.

    * Please let's not allow perfect to be the enemy of the good. With this volume of work, there are bound to be some imperfections. I believe I've done due diligence to ensure the most important details have been well-attended to.

    * CI is broken because the container it's doing codegen in doesn't have Node to facilitate JS client generation. I will fix that.


hack/tilt/values.dev.yaml
```diff
@@ -15,7 +15,9 @@ api:
   rollouts:
     logs:
       enabled: true
-      urlTemplate: https://tinyurl.com/mrxabed8
+      # This is Pride and Prejudice. It's an adequate volume of text to validate
+      # that log streaming via SSE truly works as intended.
+      urlTemplate: https://www.gutenberg.org/files/1342/1342-0.txt
```
krancour, line 20: The original text (which was the opening crawl of The Empire Strikes Back) was too short to fully put streaming to the test. The whole thing fit in one frame.

pkg/client/watch/watch.go
krancour, line 1: This is the "hand written" client.

swagger.json
krancour, line 1: This is generated from the annotations on the endpoints.

pkg/http/errors.go
krancour, line 1: This was already used by the external webhooks. I enhanced it. It's often necessary (in endpoints only) to attach the desired response code to an error. The error-handling middleware knows how to deal with this.

pkg/server/kubernetes/client.go
```diff
@@ -209,11 +190,33 @@ func NewClient(
...
+	ctx context.Context,
+	list libClient.ObjectList,
+	opts ...libClient.ListOption,
+) (watch.Interface, error) {
+	// TODO(krancour): Claude did this and I don't entirely trust it. It seems a
+	// rather hacky workaround for a client that wasn't implementing WithWatch.
+	// I frankly do not think this works.
+	//
+	// Use the cluster client's watch capability directly through the REST client
+	// This delegates to the underlying cached client's watch implementation
```
krancour: The rationale for this change: Without it, I couldn't involve fake clients in tests for endpoints that support streaming.

I trust this now more than I did before. I will remove this comment. This deserves a special look from @hiddeco. If this breaks, so do streaming endpoints of the legacy API (of which their are eight) and a fair amount of UI responsiveness along with it.

pkg/server/rbac/roles.go
```diff
@@ -640,8 +640,12 @@ func (c *rolesDatabase) RevokePermissionsFromRole(
 
 	filteredRules := make([]rbacv1.PolicyRule, 0, len(role.Rules))
 	for _, rule := range role.Rules {
+		ruleResourceName := ""
+		if len(rule.ResourceNames) > 0 {
+			ruleResourceName = rule.ResourceNames[0]
+		}
 		if rule.APIGroups[0] != group || rule.Resources[0] != resourceDetails.ResourceType ||
-			(resourceDetails.ResourceName != "" && rule.ResourceNames[0] != resourceDetails.ResourceName) {
+			(resourceDetails.ResourceName != "" && ruleResourceName != resourceDetails.ResourceName) {
```
krancour, line 648: Fixes an existing bug that surfaced during testing.

pkg/cli/cmd/create/configmaps.go
krancour, line 1: @jessesuen CLI has configmap commands now.

pkg/cli/cmd/create/repo_credentials.go
krancour, line 1: @jessesuen this is a change from the pre-1.9 CLI, which had `credentials` subcommands, which were all focused on repo creds. There was nothing for generic creds.

I renamed `credentials` to `repo-credentials` (with plenty of aliases for things like `repocreds`) so as to be more specific about what these commands manage _and_ to make room for a new crop of generic gredential (`generic-credentials`) management commands that the CLI previously lacked.

pkg/server/rest_router.go
krancour, line 1: This is all the routing.

pkg/server/get_stage_v1alpha1.go
```diff
@@ -65,3 +70,97 @@ func (s *server) GetStage(
...
+// @Summary Retrieve a Stage
+// @Description Retrieve a Stage resource from a project's namespace.
+// @Tags Core, Project-Level
+// @Security BearerAuth
+// @Produce json
+// @Param project path string true "Project name"
+// @Param stage path string true "Stage name"
+// @Success 200 {object} object "Stage custom resource (github.com/akuity/kargo/api/v1alpha1.Stage)"
+// @Router /v2/projects/{project}/stages/{stage} [get]
+func (s *server) getStage(c *gin.Context) {
```
krancour, line 83: This endpoint, both in terms of swagger annotations and implementation is pretty representative of how the new API is implemented.

pkg/server/get_generic_credentials_v1alpha1.go
krancour, line 1: This file is pretty representative of a pattern that repeats many times over in the new API: there are separate endpoints for project-level, system-level, and shared resources. This is sort of a side effect of RESTs use of path params.

`/v2/projects/{project}/generic-credentials/{generic-credentials}`, for instance, embeds project ID information and necessitates a check for the project's existence.

`/v2/shared/generic-credentials/{generic-credentials}`, on the other hand neither contains that information nor necessitates such a check. It does require awareness of which namespace is the shared one -- something the other endpoint was unconcerned with.

Wherever you see this pattern manifest, you will see some minor DRY violations, but on the whole, I've been mindful of minimizing them.

pkg/server/auth_middleware.go
krancour, line 1: This logic straight up duplicates the authentication middleware for our legacy, Connect-based API. I did not bother to keep this DRY because the component this was duplicated from isn't long for this world.

pkg/server/update_resource_v1alpha1.go
```diff
@@ -37,44 +77,101 @@ func (s *server) UpdateResource(
...
+// @id UpdateResource
+// @Summary Update resources
+// @Description Update one or more Kargo resources from YAML or JSON manifests.
+// @Tags Resources
+// @Security BearerAuth
+// @Accept text/plain
+// @Produce json
+// @Param manifest body string true "YAML or JSON manifest(s)"
+// @Success 200 {object} updateResourceResponse "Update results"
+// @Router /v2/resources [patch]
```
krancour: This is NOT an idiomatically correct usage of the PATCH verb.

PUT is already used _for this same path_ for "create or update" (apply).

Using PATCH here was Claude's doing and I willingly rolled with it. I don't love it.

What I've contemplated doing instead is removing this endpoint in favor of the other one, and changing it from "create or update" to just "update" and adding support for an "upsert" query param that could be used to differentiate between the intent to update or the intent to apply.

@thomastaylor312 does that seem reasonable?

pkg/server/get_stage_v1alpha1.go
```diff
@@ -65,3 +70,97 @@ func (s *server) GetStage(
...
+
+// @id GetStage
+// @Summary Retrieve a Stage
+// @Description Retrieve a Stage resource from a project's namespace.
+// @Tags Core, Project-Level
+// @Security BearerAuth
+// @Produce json
+// @Param project path string true "Project name"
+// @Param stage path string true "Stage name"
+// @Success 200 {object} object "Stage custom resource (github.com/akuity/kargo/api/v1alpha1.Stage)"
```
krancour, line 81: This is an example of where the formal documentation is a bit hand wavey about type since Swagger doesn't play nice with k8s resource types, including CRDs.

Practically speaking, what it means is that clients need some foreknowledge of the structure. For the Go client, for example, this means a tacit understanding that the result can be unpacked into a `github.com/akuity/kargo/api/v1alpha1.Stage` struct.

hack/tilt/values.dev.yaml
thomastaylor312: Le sad. 

_starts to look furiously for a copy of the ESB script in its entirety_

pkg/server/update_resource_v1alpha1.go
thomastaylor312: Yeah, I think we remove the PATCH endpoint and use the query param like you described. Normally with the REST-ish APIs I always think of POST being create and PUT being update (and sometimes acting like an upsert). No matter how we organize it, I think we save PATCH for actual patching if we ever need it

krancour: There are several endpoints already using PATCH the correct way. This was the one sticky spot. I'll fix this in the manner discussed.

api/rbac/v1alpha1/role_types.go
```diff
@@ -28,12 +28,12 @@ type ResourceDetails struct {
 	ResourceType string   `json:"resourceType,omitempty" protobuf:"bytes,1,opt,name=resourceType"`
 	ResourceName string   `json:"resourceName,omitempty" protobuf:"bytes,2,opt,name=resourceName"`
 	Verbs        []string `json:"verbs,omitempty" protobuf:"bytes,3,rep,name=verbs"`
-}
+} // @name ResourceDetails
```
thomastaylor312, line 31: For my own edification, why did the generator need these?

pkg/cli/client/client.go
```diff
@@ -24,17 +29,22 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
...
-// the address specified in local configuration, using credentials also
-// specified in the local configuration.
+// versionHeaderTransport wraps an http.RoundTripper and adds the CLI version
+// header to every outbound request.
+type versionHeaderTransport struct {
+	wrapped http.RoundTripper
+}
+
+func (t *versionHeaderTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("X-Kargo-CLI-Version", version.GetVersion().Version)
```
thomastaylor312: Nit: Should this header be an exported `const` somewhere?

pkg/cli/client/client.go
```diff
@@ -45,40 +55,86 @@ func GetClientFromConfig(
...
+		return nil, fmt.Errorf("invalid server address: %w", err)
+	}
+
+	transportCfg := client.DefaultTransportConfig().
+		WithSchemes([]string{u.Scheme}).
+		WithHost(u.Host)
+	apiClient := client.NewHTTPClientWithConfig(strfmt.Default, transportCfg)
 
+	// Get the runtime to configure transport
+	rt := apiClient.Transport.(*httptransport.Runtime) // nolint: forcetypeassert
```
thomastaylor312: Nit: Since this function already returns an error, I'd rather error here if type assertion fails. Because we're using generated code, this is safe now, but if anything updates in the future, we could fail

pkg/cli/cmd/apply/apply.go
```diff
@@ -107,66 +106,60 @@ func (o *applyOptions) run(ctx context.Context) error {
...
-			continue
+	// Print updated resources
+	if len(updatedRes) > 0 {
+		printer, printerErr := o.toPrinter("configured")
+		if printerErr != nil {
+			return fmt.Errorf("new printer: %w", printerErr)
+		}
+		for _, res := range updatedRes {
+			obj := &unstructured.Unstructured{Object: res.UpdatedResourceManifest}
+			_ = printer.PrintObj(obj, o.Out)
```
thomastaylor312, line 161: Nit: Perhaps we should create a simple printer wrapper that does this for us and can still be configured in a similar way as we currently do it. Maybe this is the only place we do it, but I figure I'll find out as I read on

pkg/cli/cmd/apply/apply.go
```diff
@@ -107,66 +106,60 @@ func (o *applyOptions) run(ctx context.Context) error {
...
 	// TODO: Current implementation of apply is not the same as `kubectl` does.
 	// It actually "replaces" resource with the given file.
 	// We should provide the same implementation as `kubectl` does.
-	resp, err := kargoSvcCli.CreateOrUpdateResource(ctx,
-		connect.NewRequest(&kargosvcapi.CreateOrUpdateResourceRequest{
-			Manifest: manifest,
-		}))
+	res, err := apiClient.Resources.CreateOrUpdateResource(
+		resources.NewCreateOrUpdateResourceParams().
+			WithManifest(string(manifest)),
```
thomastaylor312: Mostly a question: Does `WithManifest` require a string or can it take bytes? I try to avoid string typecasting out of paranoia, but this is perfectly valid if it does require a string

pkg/cli/cmd/create/configmaps.go
```diff
@@ -0,0 +1,243 @@
...
+		cmd.Flags(), &o.Project, o.Config.Project,
+		"The project in which to create the ConfigMap. If not set, the default project will be used.",
+	)
+	option.Shared(
+		cmd.Flags(), &o.Shared, false,
+		"Whether to create a shared ConfigMap that can be used across all projects.",
+	)
+	option.System(
+		cmd.Flags(), &o.System, false,
+		"Whether to create a system ConfigMap.",
```
thomastaylor312, line 113: Nit: Might be good to have a little more about when you need system configmaps. Good for first time users and LLMs

pkg/cli/cmd/create/configmaps.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// Parse and validate --set values
+	o.Data = make(map[string]string)
+	for _, kv := range o.SetValues {
+		parts := strings.SplitN(kv, "=", 2)
+		if len(parts) != 2 {
+			errs = append(errs, fmt.Errorf("invalid --set format %q: expected key=value", kv))
+			continue
+		}
+		key, value := parts[0], parts[1]
```
thomastaylor312: We should probably do a basic sanitizing step of running these through `strings.TrimSpace()` before setting them into the configmap

pkg/cli/cmd/create/configmaps.go
```diff
@@ -0,0 +1,243 @@
...
+		errs = append(errs, fmt.Errorf(
+			"one of %s, %s, or %s is required",
+			option.ProjectFlag, option.SharedFlag, option.SystemFlag,
+		))
+	}
+
+	// Parse and validate --set values
+	o.Data = make(map[string]string)
+	for _, kv := range o.SetValues {
+		parts := strings.SplitN(kv, "=", 2)
```
thomastaylor312, line 151: Probably don't need to worry about this, but is `=` a valid character for a key in a configmap? I haven't seen it, but I've seen too many people use k8s like this 

Image

pkg/cli/cmd/create/configmaps.go
```diff
@@ -0,0 +1,243 @@
...
+// run creates the ConfigMap based on the options.
+func (o *createConfigMapOptions) run(ctx context.Context) error {
+	apiClient, err := client.GetClientFromConfig(ctx, o.Config, o.ClientOptions)
+	if err != nil {
+		return fmt.Errorf("get client from config: %w", err)
+	}
+
+	var resp interface {
+		GetPayload() any
+	}
```
thomastaylor312: Nit: Is there a normal interface we can use instead of an inline one?

pkg/cli/cmd/create/create.go
```diff
@@ -117,37 +118,42 @@ func (o *createOptions) run(ctx context.Context) error {
...
+					fmt.Errorf("marshal created manifest: %w", err))
+				continue
+			}
+			var obj unstructured.Unstructured
+			if err := json.Unmarshal(manifestJSON, &obj); err != nil {
+				_, _ = fmt.Fprintf(o.ErrOut, "Error: %s",
+					fmt.Errorf("unmarshal created manifest: %w", err))
+				continue
+			}
+			_ = printer.PrintObj(&obj, o.Out)
```
thomastaylor312, line 155: Nit: I feel like there is an encoder/codec/something in the k8s libraries that can facilitate this without a roundtrip serialize/deserialize. This penalty could add up, especially if there are many results in the payload. This works, so no hard requirement to change, but I'd like to come back to this later

Only commenting on this here, but it does show up in multiple places

pkg/cli/cmd/create/generic_credentials.go
```diff
@@ -0,0 +1,252 @@
...
+
+	// Parse and validate --set values
+	o.Data = make(map[string]string)
+	for _, kv := range o.SetValues {
+		parts := strings.SplitN(kv, "=", 2)
+		if len(parts) != 2 {
+			errs = append(errs, fmt.Errorf("invalid --set format %q: expected key=value", kv))
+			continue
+		}
+		key, value := parts[0], parts[1]
```
thomastaylor312: Same note here around trimming

pkg/cli/cmd/create/project.go
```diff
@@ -109,21 +109,30 @@ func (o *createProjectOptions) run(ctx context.Context) error {
...
-	resp, err := kargoSvcCli.CreateResource(
-		ctx,
-		connect.NewRequest(
-			&kargosvcapi.CreateResourceRequest{
-				Manifest: projectBytes,
-			},
-		),
+	res, err := apiClient.Resources.CreateResource(
+		resources.NewCreateResourceParams().
+			WithManifest(string(projectBytes)),
```
thomastaylor312, line 114: Same note here around string vs bytes. I won't comment on any others I see

pkg/cli/cmd/create/role.go
```diff
@@ -139,28 +139,48 @@ func (o *createRoleOptions) run(ctx context.Context) error {
...
+		},
+		Claims: claims,
+	}
+
+	// Convert role to map[string]any for the request
+	roleJSON, err := json.Marshal(role)
+	if err != nil {
+		return fmt.Errorf("marshal role: %w", err)
+	}
+	var roleMap map[string]any
```
thomastaylor312, line 155: Nit: This may be a place where https://github.com/go-viper/mapstructure is useful, though I'm not sure if it does anything for speed or not. This would also be resolved if we figure out the k8s encoding/decoding stuff I mentioned earlier

pkg/cli/cmd/create/token.go
```diff
@@ -130,24 +132,48 @@ func (o *createTokenOptions) validate() error {
...
+	}
+
+	secretJSON, err := json.Marshal(payload)
 	if err != nil {
-		return fmt.Errorf("get API token: %w", err)
+		return fmt.Errorf("marshal response: %w", err)
+	}
+	var secret *corev1.Secret
+	if err = json.Unmarshal(secretJSON, &secret); err != nil {
+		return fmt.Errorf("unmarshal secret: %w", err)
```
thomastaylor312, line 176: Nit: Only do this step if printing the object, otherwise you can just access the secret token data from the original non-k8s data structure

pkg/cli/cmd/get/generic_credentials.go
```diff
@@ -0,0 +1,294 @@
...
+		"The project for which to list credentials. If not set, the default project will be used.",
+	)
+	option.Shared(
+		cmd.Flags(), &o.Shared, false,
+		"Whether to list shared credentials instead of project-specific credentials.",
+	)
+	option.System(
+		cmd.Flags(), &o.System, false,
+		"Whether to list system credentials instead of project-specific credentials.",
+	)
```
thomastaylor312, line 130: Nit: This logic and validation is copied across several of these new commands. Could this be refactored into a common package somewhere for use across all of these?

pkg/cli/cmd/update/configmaps.go
```diff
@@ -0,0 +1,309 @@
...
+		errs = append(errs, fmt.Errorf(
+			"one of %s, %s, or %s is required",
+			option.ProjectFlag, option.SharedFlag, option.SystemFlag,
+		))
+	}
+
+	// Parse and validate --set values
+	o.Data = make(map[string]string)
+	for _, kv := range o.SetValues {
+		parts := strings.SplitN(kv, "=", 2)
```
thomastaylor312, line 166: Same note here around paranoid validation. This happens in secrets too, so I won't comment again there

pkg/cli/cmd/update/configmaps.go
```diff
@@ -0,0 +1,309 @@
...
+		if len(parts) != 2 {
+			errs = append(errs, fmt.Errorf("invalid --set format %q: expected key=value", kv))
+			continue
+		}
+		key, value := parts[0], parts[1]
```
thomastaylor312: Same note around trimming. This happens in secrets too, so I won't comment again there

pkg/cli/tests/e2e.sh
thomastaylor312, line 1: This is really nice, but also a bit unwieldy because it is so long. Consider breaking up each of the test sections into their own bash file that is then sourced into this top level script. Should make it easier to add tests down the line

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+type Client struct {
+	baseURL    string
+	httpClient *http.Client
+	token      string
+}
+
+// NewClient creates a new watch client.
+func NewClient(baseURL string, httpClient *http.Client, token string) *Client {
+	if httpClient == nil {
+		httpClient = http.DefaultClient
```
thomastaylor312: Should this be using the `cleanhttp` default client instead?

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+	}
+	return &Client{
+		baseURL:    strings.TrimSuffix(baseURL, "/"),
+		httpClient: httpClient,
+		token:      token,
+	}
+}
+
+// WatchStage watches a specific Stage for changes.
```
thomastaylor312: Nit: Might be worth adding to each of the doc strings that the given context should be used to stop the watch so people know to do that when they want to cancel and so we remember when we pass a context not to cancel it

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+	return eventCh, errCh
+}
+
+// readSSEStream reads SSE events from the response body and sends them to the channel.
+func readSSEStream[T any](
+	ctx context.Context,
+	body io.Reader,
+	eventCh chan<- Event[T],
+) error {
+	scanner := bufio.NewScanner(body)
```
thomastaylor312, line 207: Nit: Consider setting a custom `SplitFunc` using `scanner.Split` that handles the double newline. This would technically follow the spec more correctly

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+	for scanner.Scan() {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		line := scanner.Text()
```
thomastaylor312: Just to double check: Do we ever expect to receive a large body over [`MaxScanTokenSize`](https://pkg.go.dev/bufio@go1.25.5#MaxScanTokenSize)? We ran into this before with the yaml updater and I worry about large objects possibly triggering an issue while scanning

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+
+		// Skip empty lines and comments (keepalives)
```
thomastaylor312: Above here, we should be checking `scanner.Err()` and returning that in the error chan if there is one

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+		defer close(errCh)
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+		if err != nil {
+			errCh <- fmt.Errorf("creating request: %w", err)
+			return
+		}
+
+		req.Header.Set("Accept", "text/event-stream")
+		req.Header.Set("Cache-Control", "no-cache")
```
thomastaylor312: In some examples I also see the `Connection` header being set. Do we need that?

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+	var response PromoteToStageResponse
+	if err := json.Unmarshal(respBody, &response); err != nil {
+		return nil, fmt.Errorf("unmarshaling response: %w", err)
+	}
+
+	return response.Promotion, nil
+}
+
+// PromoteDownstream creates promotions for all stages downstream from the specified stage.
+func (c *Client) PromoteDownstream(
```
thomastaylor312: Is there a reason this functions are in the `watch` package? I wasn't seeing how they related to the watch code

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+		defer close(errCh)
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+		if err != nil {
+			errCh <- fmt.Errorf("creating request: %w", err)
+			return
+		}
+
+		req.Header.Set("Accept", "text/event-stream")
+		req.Header.Set("Cache-Control", "no-cache")
```
thomastaylor312: Same note here on if the connection header matters

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+		if c.token != "" {
+			req.Header.Set("Authorization", "Bearer "+c.token)
+		}
+
+		resp, err := c.httpClient.Do(req)
+		if err != nil {
+			errCh <- fmt.Errorf("executing request: %w", err)
+			return
+		}
```
thomastaylor312: Some of this logic should be wrapped in a shared function so as to not duplicate the same thing as in the watch code

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+	var dataLines []string
+
+	for scanner.Scan() {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		line := scanner.Text()
```
thomastaylor312: Check for `scanner.Err()` after this, otherwise if a line gets too long or something else happens, we'll fail silently

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+		if line == "" {
+			if len(dataLines) > 0 {
+				// Join accumulated data lines and send
+				data := strings.Join(dataLines, "\n")
+				select {
+				case logCh <- data:
+				case <-ctx.Done():
+					return ctx.Err()
+				}
+				dataLines = nil
```
thomastaylor312: This should be `dataLines = dataLines[:0]` to reuse the same backing array rather than reallocating every time. Having a bunch of logs stream for a while will consistently increase memory usage and also trigger garbage collection passes. I focus on this here because logs can be noisy and verbose, so this will likely matter here.

For later (don't need to do this now), we probably want a max capacity or some other heuristic for when we instead mark as `nil` so it can get reaped by GC

pkg/client/watch/watch.go
```diff
@@ -0,0 +1,526 @@
...
+// readLogStream reads log entries from an SSE stream. The server sends each
+// line prefixed with "data: " and terminates events with an empty line.
+// Multi-line log chunks are sent as multiple "data:" lines within a single
+// event.
+func readLogStream(
+	ctx context.Context,
+	body io.Reader,
+	logCh chan<- string,
+) error {
+	scanner := bufio.NewScanner(body)
```
thomastaylor312, line 350: I think we're ok without a custom internal buffer here, but I worry about long log lines causing an error for being too long

pkg/server/kubernetes/client.go
```diff
@@ -209,11 +190,33 @@ func NewClient(
 	return c, nil
 }
 
+// watchableClient wraps a client and adds Watch functionality
+type watchableClient struct {
+	libClient.Client
```
thomastaylor312: Rather than embedding `Client`, why not just embed `WithWatch` instead (since it embeds `Client` as well)? `newDefaultInternalClient` returns an `error`, so rather than type asserting this on every call, we can type assert once on creation and then we don't have to type assert every time here. In fact, we might just be able to get away without this type at all.

Obviously this only applies if the other comment you asked for Hidde's feedback on is ok

pkg/server/kubernetes/client.go
```diff
@@ -679,13 +685,43 @@ func (c *client) Watch(
...
-		ri = c.internalDynamicClient.Resource(gvr)
+		listOpts = append(listOpts, libClient.InNamespace(namespace))
+	}
+	if opts.FieldSelector != "" {
+		// Parse the field selector and convert to MatchingFields
+		// For now, we only support the metadata.name field selector
+		if strings.HasPrefix(opts.FieldSelector, "metadata.name=") {
+			name := strings.TrimPrefix(opts.FieldSelector, "metadata.name=")
+			listOpts = append(listOpts, libClient.MatchingFields{metav1.ObjectNameField: name})
+		}
```
thomastaylor312: This feels a bit brittle, especially if `metadata.name` isn't first. Any reason not to use the normal metav1 parsers for this and actually parse the field selector before extracting?

pkg/server/create_generic_credentials_v1alpha1.go
```diff
@@ -54,59 +57,217 @@ func (s *server) CreateGenericCredentials(
...
+	if req.Name == "" {
+		return errors.New("name should not be empty")
+	}
+	if len(req.Data) == 0 {
+		return errors.New("cannot create empty secret")
+	}
+	return nil
+}
+
+func (s *server) genericCredentialsToK8sSecret(
```
thomastaylor312, line 235: Nit: This method should just be `ToK8sSecret` on the `genericCredentials` type

pkg/server/create_resource_v1alpha1.go
```diff
@@ -51,25 +67,117 @@ func (s *server) CreateResource(
...
+	const maxBodyBytes = 4 * 1024 * 1024
+	manifest, err := io.LimitRead(c.Request.Body, maxBodyBytes)
+	if err != nil {
+		if errors.Is(err, &io.BodyTooLargeError{}) {
+			_ = c.Error(libhttp.Error(err, http.StatusRequestEntityTooLarge))
+			return
+		}
+		_ = c.Error(libhttp.Error(err, http.StatusBadRequest))
+		return
+	}
```
thomastaylor312: I'm seeing this in multiple places now, so we should probably make this a helper on `server` or a gin middleware of some kind for body reads

pkg/cli/tests/e2e.sh
krancour: This script was 100% Claude with no contribution from me. I originally hadn't intended to include it in the PR. But since I am now, I agree it deserves some cleanup.

pkg/server/update_resource_v1alpha1.go
krancour: Fixed in 27ee91c14cb490a871750d3bce34dcd44f072d8e

api/rbac/v1alpha1/role_types.go
krancour: It exerts some influence over the naming of generated types in the client. I don't entirely understand how the names of generated types are determined by default, but anywhere you see `@name`, it's fixing something that was unacceptably ugly.

pkg/cli/client/client.go
krancour: Fixed in 2f96f9cee28a742d9d09b2c0a45fdb318e135a61

pkg/cli/client/client.go
krancour: Fixed in 19c78cfade58d6ac4f4b34e5862ec2db55c65a5d

pkg/cli/cmd/apply/apply.go
krancour: I don't disagree, but I'm going to count it as out of scope. This was a pre-new-API thing and I want to avoid trying to make this PR fix every single thing that was already wrong. (We'd be here forever.)

pkg/cli/cmd/apply/apply.go
krancour: I tried. It's a quirk of go-swagger that anything that's `[]byte` in the swagger.json becomes `[]int64` in the generated client. Kind of a pain. So this is the cleanest alternative.

pkg/cli/cmd/create/configmaps.go
krancour: As far as I can think of, there's actually nothing yet that would use system-level ConfigMaps. This is actually here in anticipation of that eventuality. Would you feel strongly about dropping this?

pkg/cli/cmd/create/configmaps.go
krancour: Fixed in cb9a04948110c65314c16baaa8c73e9342c655c1

pkg/cli/cmd/create/configmaps.go
krancour: ConfigMap keys must be valid DNS subdomain names, which would exclude the possibility of "=" being used within a key.

pkg/cli/cmd/create/configmaps.go
krancour: Sort of...

We can move the unmarshaling into the cases and have bytes be the common output of those cases.

Fixed in cc718a11d4db34657f4ae0633bbdca85316cc552

pkg/cli/cmd/create/create.go
krancour: We talked about this offline and agreed that this is something we can afford to optimize at a later date. (But if you happen to crack the code before we merge this, I'm happy to amend the PR with it.)

pkg/cli/cmd/create/generic_credentials.go
krancour: Fixed in https://github.com/akuity/kargo/commit/cb9a04948110c65314c16baaa8c73e9342c655c1

pkg/cli/cmd/create/project.go
krancour: Responded to the other one. This is the best we can do at the moment.

pkg/cli/cmd/create/role.go
krancour: I'm going to leave this for if/when we figure out the aforementioned encoding/decoding.

pkg/cli/cmd/create/token.go
krancour: I don't think I can. The payload is `any`.

pkg/cli/cmd/get/generic_credentials.go
krancour: I'll open a follow-up issue for this.

pkg/cli/cmd/update/configmaps.go
krancour: Fixed.

pkg/cli/cmd/update/configmaps.go
krancour: Fixed

pkg/client/watch/watch.go
krancour: Fixed in af90bf37a979d42ac023d751aa47442b270b62e0

pkg/client/watch/watch.go
krancour: Fixed in b5595ee5b168fb4cd185fdb246a8d92ebb9122f8

pkg/client/watch/watch.go
krancour: Fixed in dd3b76e0a74e96d4ffa11b620d3ecd4047428d7e

pkg/client/watch/watch.go
krancour: Good question. Bumped the max size of a single SSE to 1MB in 6ba4a85f8afe1373e144716bb8b29fe4d2aa6973

pkg/client/watch/watch.go
krancour: Code's been refactored a bit since this comment was made, but in its current state, the loop ends if `scanner.Scan()` returns false and we do check `scanner.Err()` immediately after the loop.

pkg/client/watch/watch.go
krancour: That header only needs to be set on responses.

pkg/client/watch/watch.go
krancour: There's a `--wait` option for those in the CLI.

pkg/client/watch/watch.go
krancour: Just the server needs to set it on responses.

pkg/client/watch/watch.go
krancour: DRYed up in 8635887117cfbe481690db85aefb2d26020117de

pkg/client/watch/watch.go
krancour: I think this is all set now after one of the other refactors. The loop exits when `scanner.Scan()` returns false, and `scanner.Err()` is checked immediately following the loop.

pkg/client/watch/watch.go
krancour: Same as for watching resources, this buffer was increased to 1MB.

jessesuen:
I hit this panic trying the new CLI / REST API:

```
$ ~/kargo-darwin-arm64-rest get token -p kargo-demo
panic: reflect: Call using zero Value argument

goroutine 1 [running]:
reflect.Value.call({0x104b64e60?, 0x104f2b680?, 0x1400006b2c8?}, {0x10447b43a, 0x4}, {0x1400006b498, 0x2, 0x1029678e0?})
	/usr/local/go/src/reflect/value.go:432 +0x1518
reflect.Value.Call({0x104b64e60?, 0x104f2b680?, 0x10653b280?}, {0x1400006b498?, 0x1400006b348?, 0x102963bc4?})
	/usr/local/go/src/reflect/value.go:365 +0x94
k8s.io/cli-runtime/pkg/printers.printRowsForHandlerEntry({0x104f3d1c0, 0x14000734000}, 0x1064c9f80, {0x0, 0x0}, {0x0?, 0x0?}, {0x0, 0x0, 0x0, ...}, ...)
	/home/user/gocache/k8s.io/cli-runtime@v0.34.3/pkg/printers/tableprinter.go:407 +0x120
k8s.io/cli-runtime/pkg/printers.(*HumanReadablePrinter).PrintObj(0x1400006b7d0, {0x0?, 0x0?}, {0x104f3c738?, 0x14000120020?})
	/home/user/gocache/k8s.io/cli-runtime@v0.34.3/pkg/printers/tableprinter.go:161 +0x5a0
github.com/akuity/kargo/pkg/cli/cmd/get.PrintObjects[...]({0x1400006ba60, 0x0, 0x104e95980}, 0x140006fa030?, {{0x104f3c758, 0x14000120018}, {0x104f3c738, 0x14000120020}, {0x104f3c738, 0x14000120028}}, ...)
	/workspaces/kargo/pkg/cli/cmd/get/get.go:142 +0x74c
github.com/akuity/kargo/pkg/cli/cmd/get.(*getTokensOptions).run(0x14000328a80, {0x104f66910?, 0x106538b00?})
	/workspaces/kargo/pkg/cli/cmd/get/tokens.go:187 +0x4dc
github.com/akuity/kargo/pkg/cli/cmd/get.newGetTokensCommand.func1(0x140006f6908, {0x1400051a380?, 0x4?, 0x10447b386?})
	/workspaces/kargo/pkg/cli/cmd/get/tokens.go:87 +0xbc
github.com/spf13/cobra.(*Command).execute(0x140006f6908, {0x1400051a360, 0x2, 0x2})
	/home/user/gocache/github.com/spf13/cobra@v1.10.2/command.go:1015 +0x7d4
github.com/spf13/cobra.(*Command).ExecuteC(0x1400022b508)
	/home/user/gocache/github.com/spf13/cobra@v1.10.2/command.go:1148 +0x350
github.com/spf13/cobra.(*Command).Execute(...)
	/home/user/gocache/github.com/spf13/cobra@v1.10.2/command.go:1071
github.com/spf13/cobra.(*Command).ExecuteContext(...)
	/home/user/gocache/github.com/spf13/cobra@v1.10.2/command.go:1064
main.main()
	/workspaces/kargo/cmd/cli/main.go:22 +0x184
```

pkg/server/delete_repo_credentials_v1alpha1.go
```diff
@@ -22,43 +23,36 @@ func (s *server) DeleteRepoCredentials(
 		return nil, connect.NewError(connect.CodeUnimplemented, errSecretManagementDisabled)
 	}
 
-	logger := logging.LoggerFromContext(ctx)
 	project := req.Msg.GetProject()
-	if project == "" {
-		logger.Debug("no project specified, defaulting to shared resources namespace",
```
thomastaylor312, line 28: There a reason you removed logging here?

pkg/server/kubernetes/client.go
krancour: @hiddeco nevermind about looking at this. I dug into it and this is slop. I'm fixing it. I'm actually making it better than it was before all this.

pkg/server/get_analysisrun_logs_v1alpha1.go
```diff
@@ -498,3 +501,181 @@ func streamLogs(
...
+				}
+			}
+			// Empty line terminates the SSE event
+			if _, err := fmt.Fprint(c.Writer, "\n"); err != nil {
+				logger.Debug("failed to write event terminator", "error", err)
+				return
+			}
+
+			// Flush to ensure the data is sent immediately
+			c.Writer.Flush()
```
thomastaylor312, line 669: Nit: This should probably be encapsulated in a struct that also contains the `logCh` above. There can be a method to call to "close" the SSE stream (or it can be automatically triggered with canceling the context). This is a minimal amount of code to copy around, but would reduce duplication and make any bug fixes easier

pkg/server/get_cluster_config_v1alpha1.go
```diff
@@ -53,3 +58,88 @@ func (s *server) GetClusterConfig(
...
+	// ClusterConfig is cluster-scoped, so no namespace
+	w, err := s.client.Watch(ctx, &kargoapi.ClusterConfig{}, "", opts)
+	if err != nil {
+		logger.Error(err, "failed to start watch")
+		_ = c.Error(fmt.Errorf("watch cluster config: %w", err))
+		return
+	}
+	defer w.Stop()
+
+	c.Writer.Flush()
```
thomastaylor312: I haven't looked at `setSSEHeaders` yet, but I think this `Flush` should be there. My thinking is we want to keep enough of the lower level logic here as easy to use as possible. It is super easy when adding a new feature to forget to add a flush and have an issue, so we should make sure flushes are in there as much as possible.

pkg/server/kubernetes/client.go
krancour: This is completely refactored in 21d319f93f7763df6fd4db31bd31f19fe319adb3 and I'm quite confident in the improvements.

pkg/server/get_role_v1alpha1.go
```diff
@@ -99,3 +102,117 @@ func (s *server) GetRole(
...
+	}
+
+	sa, roles, rbs, err := s.rolesDB.GetAsResources(ctx, false, project, name)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	// The ServiceAccount is the most critical component of a Kargo Role. If one
+	// was not found, the Kargo Role does not exist.
```
thomastaylor312, line 133: Image

pkg/server/kubernetes/client.go
krancour: I did that and one better. Our authorizing client wrapper now implements `client.WithWatch()` directly so it's now less custom.

krancour: That's in 21d319f93f7763df6fd4db31bd31f19fe319adb3

pkg/server/kubernetes/client.go
krancour: This fell out with one of the other refactors.

pkg/server/get_warehouse_v1alpha1.go
```diff
@@ -90,3 +96,96 @@ func prepareOutboundWarehouse(w *kargoapi.Warehouse) error {
...
 }
+
+// @id GetWarehouse
+// @Summary Retrieve a Warehouse
+// @Description Retrieve a Warehouse resource from a project's namespace.
+// @Tags Core, Project-Level
+// @Security BearerAuth
+// @Param project path string true "Project name"
+// @Param warehouse path string true "Warehouse name"
+// @Produce json
```
thomastaylor312, line 105: I don't know if this matters for gin or not, but since passing `watch` could cause this to return `text/event-stream`, will this throw things off?

pkg/server/kubernetes/client.go
hiddeco: That does look better indeed.

pkg/server/list_promotion_tasks_v1alpha1.go
```diff
@@ -50,3 +52,39 @@ func (s *server) ListPromotionTasks(
...
+		ctx, list, client.InNamespace(project),
+	); err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	// Sort ascending by name
+	slices.SortFunc(list.Items, func(lhs, rhs kargoapi.PromotionTask) int {
+		return strings.Compare(lhs.Name, rhs.Name)
+	})
```
thomastaylor312, line 83: Nit: Just had a thought as I saw this again. This could be generalized with making a function that takes anything with `ObjectMeta`.  It would shorten all of these list things to `slices.SortFunc(list.Items, objectNameFunc)`

pkg/server/list_promotions_v1alpha1.go
```diff
@@ -51,3 +56,103 @@ func (s *server) ListPromotions(
...
+
+	c.Writer.Flush()
+
+	keepaliveTicker := time.NewTicker(30 * time.Second)
+	defer keepaliveTicker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Debug("watch context done", "error", ctx.Err())
```
thomastaylor312, line 123: Is there something we need to write to the client to inform it the stream is done or does the closing of the http request by exiting from this function just handle it?

pkg/server/kubernetes/client.go
krancour: Confirmed this is no longer an issue. We use controller-runtime list options directly now and don't need to do any conversion.

pkg/server/create_generic_credentials_v1alpha1.go
krancour: I'd agree, except that this method accesses `s.cfg`.

So the choices are make this a method of genericCredentials and pass the server config or leave it as a method of the server and pass the genericCredentials.

Out of the two, I'd prefer it as it is.

pkg/server/update_config_map_v1alpha1.go
```diff
@@ -56,3 +62,184 @@ func (s *server) validateUpdateConfigMapRequest(
...
+func applyConfigMapUpdateToK8sConfigMap(configMapObj *corev1.ConfigMap, req updateConfigMapRequest) {
+	// Set the description annotation if provided
+	if req.Description != "" {
+		if configMapObj.Annotations == nil {
+			configMapObj.Annotations = make(map[string]string, 1)
+		}
+		configMapObj.Annotations[kargoapi.AnnotationKeyDescription] = req.Description
+	} else {
+		delete(configMapObj.Annotations, kargoapi.AnnotationKeyDescription)
+	}
```
thomastaylor312, line 237: Nit: I think this is the same logic as was in the patch stuff. Consider DRY-ing it up

Makefile
```diff
@@ -445,3 +470,9 @@ hack-serve-docs: hack-build-dev-tools
 .PHONY: serve-docs
 serve-docs:
 	cd docs && pnpm install && pnpm start
+
+################################################################################
+# Swagger/OpenAPI Documentation                                                #
+################################################################################
```
thomastaylor312: This a placeholder for something you haven't added yet?

pkg/cli/cmd/apply/apply.go
thomastaylor312: Yeah I wondered if it was. We can easily fix later

pkg/cli/cmd/create/configmaps.go
thomastaylor312: Nope I wouldn't feel strongly at all. Definitely in favor of keeping it out until actually needed

pkg/cli/cmd/create/configmaps.go
thomastaylor312: I looked this up and it isn't strictly DNS, but basically the same:

> Each key under the data or the binaryData field must consist of alphanumeric characters, `-`, `_` or `.`. The keys stored in data must not overlap with the keys in the `binaryData` field.

pkg/cli/tests/e2e.sh
thomastaylor312: Yeah I figured it was Claude since it likes to add comments that match its implementation plan in files it creates

pkg/client/watch/watch.go
thomastaylor312: Just calling out that this still exists just on a new line

pkg/server/create_generic_credentials_v1alpha1.go
thomastaylor312: Yeah that's fine and makes sense

pkg/server/create_resource_v1alpha1.go
krancour: I'm now using middleware for this, and even better, I'm using it for ALL endpoints now. The limit is higher for `/resources` and lower for everything else.

krancour: I'm also using middleware for checking project existence now.

pkg/server/get_cluster_config_v1alpha1.go
krancour: Fixed in e5a187e7a2c22fcef369144153d58d6271872c93

pkg/server/get_warehouse_v1alpha1.go
krancour: Nope. Gin knows nothing about these annotations.

pkg/server/list_promotion_tasks_v1alpha1.go
krancour: Oh... you're gonna love this. Can't really be done, because `list.Items` are values; not pointers. And only a pointer implements `object.Client`. I tried about a half dozen different ways of working around that before concluding this was actually the best we could manage.

pkg/client/watch/watch.go
krancour: Ok... so this was wrong. I actually have no idea why those are there. The "wait" is actually accomplished by the client initiating a watch and watching until the promotion is complete.

I removed these, which, you are right, do not belong and I refactored the promotion commands to use the generated client.

c40eff0045e99252a6f83377f256bfb53b7bc440

pkg/server/delete_repo_credentials_v1alpha1.go
krancour: idk why this method was touched at all.

1e0b0916c17f3d67f037c079d094d768521b60f2

pkg/cli/cmd/create/configmaps.go
krancour: Oh. The question was do you feel strongly that it _must_ be dropped.

pkg/server/list_promotions_v1alpha1.go
krancour: There would be if there were an actual "end" to the stream, but the semantics of "watch" are that the client watches for as long as it wants and hangs up when it's done.

Makefile
krancour: It's vestigial from some stuff I removed already.

Removed in f1c0539bfc600527281aa34b86b371d98ed39d17

pkg/server/update_config_map_v1alpha1.go
krancour: It's actually slightly different. For update, the new description is whatever was provided. For patch, there's the possibility of the description being nil == no change or "" == remove.

pkg/server/get_analysisrun_logs_v1alpha1.go
krancour: Going to let this one be for now.

pkg/client/watch/watch.go
krancour: I can't spot it.

krancour:
> I hit this panic trying the new CLI / REST API:

@jessesuen fixed

pkg/cli/tests/e2e.sh
krancour: If you don't mind, I'm going to save this for a follow up. I'd also like to get is running as part of CI eventually also. But for now, I'm willing to skip some pleasantries, bells 'n' whistles, etc. so we don't delay 1.9 further.

krancour:
@thomastaylor312 @hiddeco I updated legacy watch endpoints. They were broken because the new and improved watch functionality in our authorizing client wrapper works with structured objects instead of unstructured ones like the dynamic client had. Things are fixed now AND much improved from where those endpoints started.

@jessesuen if you'd observed any problems during testing with the UI not autofreshing, this was why.

krancour:
REST API is now versioned as `v1beta1`, which is what the team agreed to in the latest thunderdome.

pkg/server/list_promotion_tasks_v1alpha1.go
thomastaylor312: Ooof that is just...fun

pkg/server/create_resource_v1alpha1.go
```diff
@@ -111,14 +111,8 @@ func (s *server) CreateResource(
 func (s *server) createResources(c *gin.Context) {
 	ctx := c.Request.Context()
 
-	// Limit request body to 4MB to prevent DoS
-	const maxBodyBytes = 4 * 1024 * 1024
-	manifest, err := io.LimitRead(c.Request.Body, maxBodyBytes)
+	manifest, err := io.ReadAll(c.Request.Body)
```
thomastaylor312, line 116: Nit: you might want to add a comment that body limits are handled in middleware

pkg/client/watch/watch.go
thomastaylor312: Must have pushed something in between when I saw it and when you saw my comment. Just noting that the current logic still has a similar issue, but not blocking

pkg/server/create_resource_v1alpha1.go
krancour: 1be4ad1cb2bc3ea88fce09c93dd7c5a0fd17d21e

mrlubos:
@krancour curious how did you decide to use Orval over @hey-api?

Issue #5610, opened 2026-01-20
github app authentication: transient git clone issues over https


voidlily:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [ ] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

About 15% of the time, git clone operations using github app authentication over https fail with a "404 repo not found". After talking to github support, they had the following to say

> We apologise for the inconvenience caused. Private repository access failures are surfaced as 404's to avoid revealing whether a private repository exists. Our logs for Git clone requests headed to that repository on 2026-01-16 show ~85% of calls were successful; however, the rest were rejected due to an authentication error. To confirm, this isn't anything to do with rate limiting, rather it's a problem with the credentials that are being passed when performing the git clone operation. The most likely causes for this are attempting to authenticate with a bad/expired token, or the token not being presented at all.
 
> At the moment, our logs aren’t granular enough to tell which of those two conditions is occurring. Since GitHub App installation access tokens expire quickly (1hr lifetime), would you be able to add lightweight logging for each git attempt that confirms a token was provided (no token value - just presence/length) and records the token “issued at” and “expires at” timestamps? Given that the majority of clone attempts succeed while a subset intermittently fail, this points to possible reuse of an expired GitHub App installation token, likely due to caching or refresh timing issues. 

They ruled out rate limiting and suggest this is possibly from use of an expired cached auth token.

https://github.com/akuity/kargo/blob/main/pkg/credentials/github/app.go#L61
https://github.com/patrickmn/go-cache/blob/master/cache.go

I looked at the caching implementation and a 40 minute expiration time should be covering the 1 hour token expiry. I wonder if there might be an issue with the cleanup interval being 1 hour and trying to overwrite the token between the expiration interval and cleanup interval? Some sort of write-after-read consistency issue going on?

Maybe some debug logging could help track down this error should it keep happening. I didn't notice any debug logging in that section of the code I could use if I turned on debug logging either.

# Screenshots

# Steps to Reproduce

Don't yet have a minimal reproduction instructions other than the following conditions that are currently true in my setup

* Use github app authentication
* Have ~40 projects, each with a warehouse subscription to a different private repo

I've seen 404 errors on many different repos so it didn't appear isolated to a specific repo.

# Version

```
{
  "Version": "v1.8.4",
  "BuildDate": "2025-11-25T22:29:20Z",
[...]

PR #5615, opened 2026-01-22
feat: add debug logging for github app token expiry

singhvibhanshu:
## Description
This PR introduces debug logging to the GitHub App credential provider to help diagnose intermittent `404 Not Found` errors during git clone operations (Issue #5610).

Currently, Kargo caches GitHub App installation tokens (but only stores the raw token string). This makes it impossible to know from the logs if a token retrieved from the cache is **valid**, **nearing expiration**, or **already expired**.

## What this PR tries to do?
The primary goal is **observability**. By tracking the token's lifecycle, we can verify if the 404 errors are caused by Kargo attempting to use an expired credential that was returned from the internal cache.

## Proposed Changes
- Modified the internal cache to store a `cachedToken` struct containing the `AccessToken`, `IssuedAt`, and ,`ExpiresAt` timestamps, replacing the previous simple `string` storage.
- Added `log.Printf` statements when retrieving a token from the cache. Logs now displays critical metadata like `issued_at`, `expires_at`, and `time_until_expiry`.
- **Note**: The actual token value is **never** logged, only the metadata is logged.
- Updated `pkg/credentials/github/app_test.go` to support the new `struct` based cache implementation approach and ensures no regressions in the auth logic.

## What this PR does NOT do
- It does not fix the underlying `404 error`.
- It does not change the auth behaviour.

## Verification
- [x] Ran `go test ./pkg/gitprovider/github` (All tests passed).
- [x] Ran `make hack-lint-go` (No linting errors).

### DIFF
--- a/pkg/credentials/github/app.go
+++ b/pkg/credentials/github/app.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
 	"net/url"
 	"regexp"
 	"slices"
@@ -35,6 +36,12 @@ const (
 
 var base64Regex = regexp.MustCompile(`^[a-zA-Z0-9+/]*={0,2}$`)
 
+type cachedToken struct {
+	accessToken string
+	issuedAt    time.Time
+	expiresAt   time.Time
+}
+
 func init() {
 	if provider := NewAppCredentialProvider(); provider != nil {
 		credentials.DefaultProviderRegistry.MustRegister(
@@ -54,7 +61,7 @@ type AppCredentialProvider struct {
 		installationID int64,
 		encodedPrivateKey string,
 		repoURL string,
-	) (string, error)
+	) (*cachedToken, error)
 }
 
 // NewAppCredentialProvider returns an implementation of credentials.Provider.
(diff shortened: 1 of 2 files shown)


krancour:
@singhvibhanshu isn't this an experiment you could attempt on your own without this having to be merged into the project?

singhvibhanshu:
hey @krancour!

my thinking was that currently, when users get a 404 error, its very confusing for them because they don't know if the repo is actually missing or if their token just expired.

even though this pr doesnt fix the underlying bug, i thought adding these logs would make life much easier for the users to debug the issue and confirm if its just an expired token.

however, if you feel these changes aren't right for the project or add too much noise, i am happy to close my pr.

krancour:
@singhvibhanshu it depends...

> we can verify if the 404 errors are caused by Kargo attempting to use an expired credential that was returned from the internal cache

Is this research? Like you're getting 404s and speculate this could be the cause and you want to confirm? Or are you already certain expired creds from the cache are causing 404s and you just want the logs to show that?

If it's the latter, it might be the smoking gun in this discusson: https://github.com/akuity/kargo/issues/5616#issuecomment-3801125557

I was already beginning to question whether there was any possibility that we were caching credentials past their expiration and started work on reducing the likelihood of that yesterday. I didn't open a PR yet, but will in a moment. I think I already may have included debug messages about expiry date/time, but if not, I will add them.

So this may end up closed, but the change you're asking for will still land.

In the meantime, I'd still be very interested to know if you've already confirmed this to be the source of 404s. Also... do you happen to be using GitHub _Enterprise_? I've begun to wonder if a token's lifespan is different (or configurable) in Enterprise and it invalidated prior assumptions about how long a token was good for. This suspicion is why the PR I am opening shortly calculates cache TTL dynamically based on the expiry date.

singhvibhanshu:
**Is this research?**
Yes, I am encountering those 404s and strongly suspect the cache is returning expired credentials, but i haven't been able to definitively prove it yet without these logs.

Also, I am using Standard GitHub not the Enterprise one.

It sounds like your upcoming PR regarding dynamic TTL calculation addresses the root cause. If that PR also adds debug logging for expiry, then my PR is definitely redundant.

I'll keep an eye out for your PR and am happy to close this one, once yours is up.
Thanks!

krancour:
Closed in favor of #5684

singhvibhanshu:
Thanks for the update, @krancour!

The dynamic TTL calculation in #5684 definitely looks like the right way to solve the root cause, so I am totally happy with this being closed in favor of that.

On that note, I am eager to keep contributing to Kargo. Are there any other open issues or specific areas of the backlog you would suggest I look at next? I would be happy to pick something else up.

PR #5617, opened 2026-01-23
fix: add needed permissions to shared resources for built-in roles

jessesuen:
When navigating the Kargo UI while logged in as non-admin built-in roles such as `kargo-user`, `kargo-viewer`, `kargo-project-creator`, the UI would error on the various tabs on system settings page due to the inability to view ~serviceaccounts~, shared secrets, and shared configmaps. See screenshots:

EDIT: we do not need non-admin roles to see system settings (serviceaccounts, tokens), at this time. only shared resources.

image

image

It is intended for kargo end users to be able to list shared secrets and configmaps, so that they know the name of the Secret or ConfigMap to reference in promotion templates (via `sharedSecret()` and `sharedConfigMap()` helpers).


jessesuen:
One thing that needs to be discussed is whether non-admin users should be able to visit the `Access` and `System Secrets` tab without error. In other words, should end users know that there exists:
* system secrets and their name
* the built-in kargo roles (kargo-user, kargo-viewer, kargo-admin, kargo-project-creator)
* the names of tokens associated with the above roles

image

Since Kargo API redacts secret data, and also filters non-API token related secrets in responses, allowing this should be safe.

But if non-admin users have no business viewing roles, tokens, or system secrets, I would remove the permission and have the UI hide those tabs from them.

charts/kargo/templates/shared-resources-namespace/roles.yaml
```diff
@@ -1,6 +1,24 @@
...
+metadata:
+  name: kargo-shared-resources-reader
+  namespace: {{ .Values.global.sharedResources.namespace }}
+  labels:
+    {{- include "kargo.labels" . | nindent 4 }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
```
jessesuen: Discussed. We don't want to give this ability directly via k8s RBAC, but rather a specially treated API server endpoint that knows that callers may not have that privilege.

jessesuen:
> But if non-admin users have no business viewing roles, tokens, or system secrets, I would remove the permission and have the UI hide those tabs from them.

The safer choice at the moment is not to add this privilege. So for the time being, I will remove the rbac for this and let it error. We discussed improving UI so that those users won't at least see an error.

jessesuen:
PR has been updated to:
1. not have read privileges on secrets for the non-admin roles, since the Kargo API will be improved with a backend change to allow the redacted form of the secret, without the actual role needing access to it
2. remove the ability for the non-admin roles to view system settings. this will result in an error when visiting that portion of the UI, but we will improve this with front end changes to hide panels that users do not have access to.

PR #5629, opened 2026-01-26
fix(ui): correct format of max-age directive

mogsie:
The cache-control response header returns `public, max-age=720h0m0s`. The max-age directive is not spec-compliant.  After this change, the max-age correctly specifies 2592000.

Anecdotally, this was not a problem in Chrome (it seems to fall back to the Expires header).  It seems to affect Firefox, causing Firefox to download the 6Mb javascript file e.g. /assets/index-LxV79Jo9.js.

This should dramatically improve the experience of using Kargo on Firefox.

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -28,7 +28,7 @@ func SetCacheHeaders(w http.ResponseWriter, maxAge time.Duration, timeUntilExpir
 		return
 	}
 	w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", int(maxAge.Seconds())))
-	w.Header().Set("Expires", time.Now().Add(timeUntilExpiry).Format(time.RFC1123))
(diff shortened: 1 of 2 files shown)


mogsie:
I also noticed that the Expires header has "UTC" in it, when it should have GMT per the RFC. I'll fix that too.

I also need to point out that *not all* configurations of Firefox are affected, but suffice to say, *some* are, and IMHO there's no point in making remotely working people re-download a 6Mb .js file when it's not needed.

krancour (dismissed):

krancour:
> I also noticed that the Expires header has "UTC" in it, when it should have GMT per the RFC. I'll fix that too.

LGTM pending this

mogsie:
Used [`http.TimeFormat`](https://pkg.go.dev/net/http#TimeFormat)! TIL!

mogsie:
Rebased and ready @krancour do I now unassign myself?

krancour:
> do I now unassign myself?

No need.

Thank you very much for this small, but meaningful improvement. 🙏

PR #5634, opened 2026-01-26
feat(cli): allow auth using api token in env var

krancour:
cc @jessesuen

### DIFF
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -67,6 +67,16 @@ type CLIConfig struct {
 	Project string `json:"project,omitempty"`
 }
 
+// NewEnvVarCLIConfig returns a new CLI configuration populated from environment
+// variables.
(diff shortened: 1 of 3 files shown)


pkg/cli/config/config.go
```diff
@@ -83,7 +87,7 @@ func loadCLIConfig(configPath string) (CLIConfig, error) {
 	_, err := os.Stat(configPath)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return cfg, fmt.Errorf("please use `kargo login` to continue: %w", NewConfigNotFoundErr(configPath))
+			return cfg, NewConfigNotFoundErr(configPath)
```
krancour, line 96: Removed the error text here, because there was never any scenario in which it was displayed, since the behavior of the caller is to fall back on default config if the config file doesn't exist.

Issue #5616, opened 2026-01-22
bug in url normalization


krancour:
Creating a new issue for this so I don't lose track of it.

Refer to [this discussion](https://github.com/akuity/kargo/pull/5609/files#r2718462715). 

Although it's not yet verified as the root cause of the issue #5609 was intended to address, investigation into that has left me certain that our normalization processes (the one for Git URLs, at least, but possibly others as well) is not adequately defensive about whitespace and other invisible characters (like a BOM) that can interfere with parsing a URL.

This small program demonstrates:

```golang
package main

import (
	"fmt"

	"github.com/akuity/kargo/pkg/urls"
)

func main() {
	orig := "\ufeffhttps://github.com/org/repo.git"
	norm := urls.NormalizeGit(orig)
	fmt.Printf("Unchanged: %v\n", orig == norm)
}
```

likelymichael:
it's worth noting that, at least from our testing, this issue is not consistent. on the first attempt of promoting to a stage, we'll encounter the error about the URL not being normalized:

```bash
step "step-1" met error threshold of 1: error running step "step-1": error cloning https://github.com/org/repo.git: error cloning repo "https://github.com/org/repo.git" into "/tmp/promotion-88303aa4-620a-48c6-b9d9-5e3123709e4e/repo-310626314/repo": error executing cmd [/usr/bin/git clone --bare https://kargo@github.com/org/repo.git /tmp/promotion-88303aa4-620a-48c6-b9d9-5e3123709e4e/repo-310626314/repo]: Cloning into bare repository '/tmp/promotion-88303aa4-620a-48c6-b9d9-5e3123709e4e/repo-310626314/repo'... remote: Repository not found. fatal: repository 'https://github.com/org/repo.git/' not found
```

however if we retry 1-2 times, it usually will work

krancour:
@likelymichael is that with or without #5609 in play?

likelymichael:
> [@likelymichael](https://github.com/likelymichael) is that with or without [#5609](https://github.com/akuity/kargo/pull/5609) in play?

Both

krancour:
Ok. Research into #5609 at least exposed this bug in the normalization logic, which I still want to fix, but if you encounter this problem intermittently _with or without_ the changes from #5609 in place, that would seem to place the blame for this somewhere other than the normalization logic. 🤔 

**Edit:** Re-reading #5609, you said:

[...]

PR #5637, opened 2026-01-27
fix: make url normalization remove leading and trailing whitespace + all non-printable runes

fuskovic:
Reverts https://github.com/akuity/kargo/pull/5609 and Closes https://github.com/akuity/kargo/issues/5616

### DIFF
--- a/pkg/credentials/github/app.go
+++ b/pkg/credentials/github/app.go
@@ -280,7 +280,7 @@ func (p *AppCredentialProvider) extractRepoName(repoURL string) string {
 	if len(parts) < 5 {
 		return ""
 	}
-	return strings.TrimSuffix(parts[len(parts)-1], ".git")
+	return parts[len(parts)-1]
 }
 
 // extractBaseURL extracts the base URL from a full repository URL. The base
--- a/pkg/credentials/github/app_test.go
+++ b/pkg/credentials/github/app_test.go
@@ -178,17 +178,6 @@ func TestAppCredentialProvider_Supports(t *testing.T) {
 			},
 			expected: true,
 		},
-		{
-			name:     "valid with .git suffix",
-			credType: credentials.TypeGit,
-			repoURL:  testRepoURL + ".git",
-			getDataMap: func() map[string][]byte {
-				dm := maps.Clone(supportedDataMap)
-				delete(dm, clientIDKey)
-				return dm
-			},
-			expected: true,
-		},
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
@@ -536,11 +525,6 @@ func TestAppCredentialProvider_extractRepoName(t *testing.T) {
 		{
 			name:     "GitHub URL",
 			repoURL:  "https://github.com/example/repo",
-			expected: "repo",
-		},
-		{
-			name:     "GitHub URL with .git suffix",
-			repoURL:  "https://github.com/example/repo.git",
 			expected: "repo",
 		},
 		{


pkg/urls/git_test.go
```diff
@@ -9,11 +10,11 @@ import (
 func TestNormalizeGit(t *testing.T) {
 	testCases := map[string]string{
 		// Anything we can't normalize should be returned as-is
-		"https://not a url":                      "https://not a url",
+		"https://not a url":                      "https://notaurl",
```
krancour: We should preserve existing behavior which is still in the description of the exported normalization functions:

> Any URL that cannot be normalized will be returned as-is.

If the output isn't going to be a valid URL, it shouldn't be a _different_ invalid URL from what we started with.

pkg/urls/git.go
```diff
@@ -76,3 +69,19 @@ func NormalizeGit(repo string) string {
...
+		return nil, fmt.Errorf("parsing URL: %w", err)
+	}
+	if len(repoURL.Query()) > 0 {
+		// Query parameters are not permitted
+		return nil, fmt.Errorf("URL contains %d query parameters; not permitted", len(repoURL.Query()))
+	}
+	decodedPath, err := url.PathUnescape(repoURL.Path)
+	if err == nil {
+		repoURL.Path = decodedPath
+	}
```
krancour: `url.Parse()` will have already done this. Same for the other location where you used this.

pkg/urls/sanitize.go
```diff
@@ -0,0 +1,30 @@
...
+func sanitize(repo string) string {
+	decodedPath, err := url.PathUnescape(repo)
+	if err == nil {
+		repo = decodedPath
+	}
+	repo = strings.TrimSpace(strings.ToLower(repo))
+	return strings.Map(func(r rune) rune {
+		if unicode.IsSpace(r) || r == zeroWidthNoBreakSpace || r == zeroWidthSpace || r == noBreakSpace {
+			return -1 // Remove the character
+		}
```
krancour: This should be sufficient:

```suggestion
		if !unicode.IsPrint(r) {
			return -1 // Remove the character
		}
```

pkg/urls/sanitize.go
```diff
@@ -0,0 +1,30 @@
...
+const (
+	zeroWidthNoBreakSpace = '\uFEFF' // BOM
+	zeroWidthSpace        = '\u200B'
+	noBreakSpace          = '\u00A0'
+)
+
+// sanitize performs basic sanitization of a repository URL by trimming
+// leading and trailing whitespace, converting to lowercase, and removing
+// unusual whitespace characters.
+func sanitize(repo string) string {
```
krancour: idk about `sanitize()` as a name for this function. I think the usual connotation to the word "sanitize" in a SWE context is removal of sensitive information. For instance, elsewhere in the codebase, redaction of Secret data is performed with methods having "sanitize" in their names.

But rather than seeking out a better name for this function, I'd propose this: Given my other suggestion below about how to make this even more short and sweet than it was to begin with, I think what's left on lines 23 - 27 ought to just be transplanted to the beginning of your `getURL()` method, which I'd further propose be renamed to `safeParseURL()` or something along those lines because it's a more self-documenting name than `getURL()`, which doesn't tell me much.

pkg/urls/git.go
fuskovic: Call stack for `url.Parse` is -> `url.setFragment` -> `unescape(f, encodeFragment)`.

Where `url.PathUnescape` calls `unescape(s, encodePathSegment)` which does not have the same behavior.

This is clear when test cases related to encoded spaces e.g. `%20` start to fail as a result of this suggestion due to `invalid URL escape "%20"`.

pkg/urls/sanitize.go
fuskovic: Unfortunately it's not. It only counts ` U+0020` as the only spacing character per the description in the std lib comment: `This categorization is the same as [IsGraphic] except that the only spacing character is ASCII space, U+0020`.

fuskovic: It is together though e.g. `if unicode.IsSpace(r) || !unicode.IsPrint(r)`

pkg/urls/sanitize.go
fuskovic: I've reduced the scope and renamed to `rmSpaces`. There are cases where you may want to use this and not `safeParseURL` yet and vice versa so I've opted to keep those as two separate functions; each with it's own narrow scope.

pkg/urls/git_test.go
fuskovic: The existing test cases imply that a repo URL with spaces is invalid and so we should return the original. However, the fix we are trying to implement contradicts this. e.g. repo URLs with white spaces should be normalized. Seems like it needs to be one way or the other. Thoughts?

pkg/urls/git.go
```diff
@@ -76,3 +72,32 @@ func NormalizeGit(repo string) string {
...
+		repo = decodedPath
+	}
+	repo = strings.TrimSpace(strings.ToLower(repo))
+	return strings.Map(func(r rune) rune {
+		if unicode.IsSpace(r) || !unicode.IsPrint(r) {
+			return -1 // Remove the character
+		}
+		return r
+	}, repo)
+}
```
krancour: We're doing too much here. I don't want to jump through hoops to make a valid URL out of an obviously invalid one. I only want to correct for the sort of innocent mistakes that are apt to happen as a result of copy/paste. 

Leading or trailing whitespace is an easy mistake to make when copying/pasting. Invisible characters are probably most likely to be leading or trailing, but really, if you had one _anywhere_ in a string you wouldn't know it was there.

`http://exa ple.com`, on the other hand, is a very different scenario. It's _visibly_ invalid. If we just let the parsing fail, and return the original url, it will eventually show up in some error message exactly as it was originally written and it will be immediately obvious what's wrong and how it should be fixed.

Another subtle thing here is that we don't actually know that removing spaces fixes a url correctly. `http://exa ple.com` seems more likely to be mis-typed `http://example.com` than `http://exaple.com`. We just don't know and shouldn't guess.

I also don't think we need to path unescape. For one, that will already happen automatically when we parse, but the second reason is that, `http://exa%20ple.com` is just as obviously invalid as `http://exa ple.com` is. We need not guess at how to fix it.

Ultimately, I wouldn't do anything more than trim leading and trailing whitespace and remove invisible characters from any position in the string. I wouldn't even do the downcasing here. That can happen elsewhere. I'd make this function short and sweet and 100% about cleaning up a string that is an invalid url, but not _obviously_ invalid.

krancour: Sorry. I didn't see your responses to other comments of mine until after having written this one. I hope that this comment adds some clarity to what's in scope and out of scope. This was really supposed to be a small, very targeted change.

fuskovic:
@krancour I have this in a place now where we are retaining the existing behavior and handling the new edge cases. If something doesn't look necessary I recommend trying to remove it and running the tests.

pkg/urls/helm.go
```diff
@@ -1,21 +1,26 @@
...
 	// repositories.
 	return NormalizeImage(
 		strings.TrimPrefix(
-			strings.ToLower(
-				strings.TrimSpace(repo),
-			),
+			strings.ToLower(repo),
 			"oci://",
 		),
 	)
```
krancour, line 26: `NormalizeImage()` was not previously downcasing the string. It is now -- and I think that was a good call. Since it is, we can skip it here.

pkg/urls/helm.go
```diff
@@ -1,21 +1,26 @@
...
 // NormalizeChart normalizes a chart repository URL for purposes of comparison.
 // Crucially, this function removes the oci:// prefix from the URL if there is
 // one.
 func NormalizeChart(repo string) string {
+	ogRepo := repo
+	repo = trimSpace(repo)
+	// just to check validity
+	if _, err := url.Parse(repo); err != nil {
+		return ogRepo
+	}
```
krancour, line 17: For clarity:

```suggestion
  // This is our only opportunity to return the original URL, so establish
  // whether it is valid prior to calling NormalizeImage().
	if _, err := url.Parse(repo); err != nil {
		return ogRepo
	}
```

Also, this was a clever thing to check. I think I would have overlooked this. 👍

pkg/urls/git.go
```diff
@@ -76,3 +76,15 @@ func NormalizeGit(repo string) string {
...
+// strings.TrimSpace does not.
+func trimSpace(repo string) string {
+	isSpaceOrUnprintable := func(r rune) bool {
+		return unicode.IsSpace(r) || !unicode.IsPrint(r)
+	}
+	return strings.TrimRightFunc(
+		strings.TrimLeftFunc(repo, isSpaceOrUnprintable),
+		isSpaceOrUnprintable,
+	)
+}
```
krancour: Two very small issues here, but easy enough to correct, so we should:

1. It is probably an extreme outlying case that an unprintable rune appears anywhere other than the beginning or end of a string, but it's _possible_ and as written, this won't remove those. We should though, because an unprintable rune _within_ the string will leave us staring at an error message containing what appears to be a valid URL and have us scratching our heads just as much as if the unprintable rune occurred at the head or tail of the URL.

1. The "space" in `trimSpace()` doesn't feel right since this function removes more than spaces. After accounting for no.1, "trim" no longer feels right either since it connotes removal of something from the head or tail (or both), but not from within. I know I might previously have suggested avoiding the word "sanitize" because it's used in other places to reflect redaction of sensitive information, but I can come up with no better descriptor for a function whose job is to remove leading and trailing whitespace only and remove non-printable runes from anywhere in the string.

Accounting for all of the above, this is my suggested implementation:

```golang
/ SanitizeURL removes leading and trailing whitespace only from a string
// presumed to represent a URL. It additionally removes non-printable runes such
// as byte order marks (BOMs) from anywhere in a string. Leading whitespace and
// non-printable runes can easily be copied and pasted without a user realizing
// and are known to interfere with URL parsing.
func SanitizeURL(url string) string {
	return strings.TrimSpace(strings.Map(func(r rune) rune {
		if unicode.IsPrint(r) {
			return r
		}
		return -1
	}, url))
}
```

Since this is a bit more general purpose than just for use with git URLs, I'd probably break this out into a `common.go`. I'd also have it exported because I think we're very likely to find other places where this will come in handy.

pkg/urls/git_test.go
```diff
@@ -81,6 +84,18 @@ func TestNormalizeGit(t *testing.T) {
...
+		" \ufeffhttps://github.com/example/repo":      "https://github.com/example/repo", // Leading BOM
+		"https://github.com/example/repo\ufeff":       "https://github.com/example/repo", // Trailing BOM
+		" https://github.com/example/repo ":           "https://github.com/example/repo", // Leading and trailing spaces
+		"\tssh://github.com/example/repo\t":           "ssh://github.com/example/repo",   // Leading and trailing tabs
+		"\nssh://github.com/example/repo\n":           "ssh://github.com/example/repo",   // Leading and trailing newlines
+		"\rssh://github.com/example/repo\r":           "ssh://github.com/example/repo",   // Leading and trailing carriage returns
+		" \t\nhttps://github.com/example/repo\t\n ":   "https://github.com/example/repo", // Mixed whitespace
+		"\u00A0https://github.com/example/repo\u00A0": "https://github.com/example/repo", // Non-breaking spaces
+		"\u200Bhttps://github.com/example/repo\u200B": "https://github.com/example/repo", // Zero-width spaces
+		"\ufeff\ufeffhttps://github.com/example/repo": "https://github.com/example/repo", // Multiple BOMs
```
krancour: I would consider testing these new cases thoroughly in a new `TestSanitizeURL()` function in `common_test.go` (see https://github.com/akuity/kargo/pull/5637/changes#r2789222721). That will cover our bases thoroughly with less code. You could leave the one or two most likely problem cases here (like leading space and leading BOM). Same comment applies to the other tests.

krancour:
A few easily addressed things remain, but overall, I like this. 👍

pkg/urls/helm.go
fuskovic: Unfortunately, we need it in both places. Consider the existing test`OCI://Repo`. It fails when making this change because the `oci://` prefix never gets trimmed due to it not be the correct case.

fuskovic: Unless you want to trim the `oci://` prefix in `NormalizeImage` but that feels out of place. Thoughts?

krancour: 🤦‍♂️ You're absolutely right. Not sure how I missed that. Nothing the see here.

krancour:
🔥 This is fantastic.

If I could make one additional request, please revert #5609 in this PR.

That PR author's original problem was misdiagnosed. That change should not have and _did not_ resolve their actual problem. The change _did_ resolve a theoretical problem that the user was not actually experiencing, but _this_ PR is a more correct and robust solution to that problem.

PR #5638, opened 2026-01-27
feat(chart): add PodDisruptionBudget and topologySpreadConstraints support

### LINKED ISSUE #5396: PodDisruptionBudget and TopologySpreadContraints for better HA configuration

# Checklist

# Proposed Feature

We configure Kargo for HA requirements (not fully distributed topology) and currently our HA specific settings are:

```
  api:
    replicas: 2
  externalWebhooksServer:
    replicas: 2
  webhooksServer:
    replicas: 2
```

However, for a stable HA configuration we think we would also need a PodDisruptionBudget and TopologySpreadContraints, so that the replicas are spread across different nodes (maybe also AZs) and the PDB takes care that at least on replica is running.

What do you think? Is this already planned? If not - and you think PDBs and TopologySpreadContraints are the things to achieve good HA - I could create a PR for it.

# Motivation

Anyone who wants to set up Kargo in HA mode.

bafulton:
## Summary

Adds support for PodDisruptionBudgets and topology spread constraints to enable robust high-availability deployments.

Closes #5396

## Changes

- Add PDB templates for api, webhooksServer, and externalWebhooksServer
- Add topologySpreadConstraints to deployment templates  
- Add configuration options in values.yaml with sensible defaults
- Update README with new parameter documentation

## Usage Example

```yaml
api:
  replicas: 2
  podDisruptionBudget:
    enabled: true
    minAvailable: 1
  topologySpreadConstraints:
    - maxSkew: 1
      topologyKey: topology.kubernetes.io/zone
      whenUnsatisfiable: ScheduleAnyway
      labelSelector:
        matchLabels:
          app.kubernetes.io/component: api
```

## Test Plan

- [x] `helm lint` passes
- [x] Templates render correctly with PDBs enabled
- [x] Templates render correctly with topologySpreadConstraints configured


krancour:
Note that #5396 was not yet evaluated by any maintainers and remains in a proposal state, as indicated by its labels.

We can take a look at this, but it's going to receive the scrutiny now that it should have received before anyone spent time implementing it.

@hiddeco would you mind having a look at this, but only if you have the bandwidth?

charts/kargo/values.yaml
```diff
@@ -69,6 +69,8 @@ global:
   tolerations: []
   ## @param global.affinity Default affinity for all Kargo pods.
   affinity: {}
+  ## @param global.topologySpreadConstraints Default topology spread constraints for all Kargo pods.
+  topologySpreadConstraints: []
```
krancour: First of all, this PR is very well done. I love it.

This is the one thing I feel pretty hesitant about. Topology spread constraints would need to specify pod selectors and those would, for sure, vary from one component to the next, which I worry might make it impractical to ever have a global default.

bafulton: @krancour, fair callout. I've removed the global default.

bafulton: Feel free to take the branch over if there are other changes you want to make. I just wanted to get the ball rolling!

charts/kargo/templates/external-webhooks-server/pdb.yaml
```diff
@@ -0,0 +1,32 @@
...
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
+spec:
+  {{- with .Values.externalWebhooksServer.podDisruptionBudget.minAvailable }}
+  minAvailable: {{ . }}
+  {{- end }}
+  {{- with .Values.externalWebhooksServer.podDisruptionBudget.maxUnavailable }}
+  maxUnavailable: {{ . }}
+  {{- end }}
```
fuskovic, line 27: Since `minAvailable` defaults to 1,  a user who sets `maxUnavailable: 1` without also clearing `minAvailable` e.g. `minAvailable: ""` will get both fields rendered; which is invalid. 

I think we can solve for this by checking if `maxUnavailable` is set and render it if it is. If not, fallback to `maxUnavailable`. 

Applies in other locations as well.

bafulton:
This whole release process has been a pain! I'm closing the PR on principle. I long ago gave up hoping this chart would be fixed and just rolled my own downstream solution.

@krancour and @fuskovic, if y'all want this, feel free to take it.

PR #5639, opened 2026-01-28
fix(api): Ensure a cred-type is set when creating a secret via the API

thomastaylor312:
When creating cluster secrets via the API, we should have been setting the credential label to generic if not already specified


thomastaylor312:
As a note, this is already fixed with the new secret namespaces in 1.9, hence the PR directly to release-1.8

Issue #5413, opened 2025-11-20
"Promote to downstream" action incorrectly opens "Promote to current stage" dialog


stephaneetje:
When attempting to use the "Promote to downstream" action on a healthy Kargo stage, the UI incorrectly loads the dialog for promoting the source Freight to the current stage, instead of proposing the current stage's content for promotion to its connected downstream stages. This prevents the expected progression in the environment pipeline.

**Steps to Reproduce:**

1. Navigate to a healthy Kargo Stage that is configured with downstream stages.
2. Click the stage's menu (e.g., the three horizontal lines icon).
3. Select the "Promote to downstream" option.

**Expected Behavior:** A promotion dialog should open, listing the connected downstream stages  and allowing the user to initiate the promotion of the current stage's verified Freight to those downstream stages.

**Actual Behavior:** The dialog that opens is titled (or displays content related to) "Promote [Source Freight ID/Alias] to Current Stage Name". This is the screen used for an initial promotion to the current stage, not for a promotion to downstream stages.

# Version
```
1.8.0 to 1.8.3
```

aryasoni98:
@stephaneetje I’m happy to work on this! Could you please assign the issue to me?

stephaneetje:
@aryasoni98  I don't think i have the permissions to assign it (it's greyed)

mortenf1984:
I also see the same issue, also present in 1.8.4

etetar:
Will this get prioritized soon? It blocks multi-promotions for Stage-to-Stage workflows, which seems like a pretty common setup.

We can currently work around this by introducing a temporary control flow stage before the parallel stages, however control flow stages seem to have a bug, where "Instant promote from upstream" is possible on the control flow stage, which always fails. Control flow stages also interfere with auto-promotions (inserting an only-manual stage before one or more auto-promoting stages after it).

stephaneetje:
Hello, is this going to be fixed anytime ? This is a very comon scenario on Kargo...

prune998:
+1 on that, the current behaviour is really weird

alishchytovych:
Version 1.8.6 still has the same issue. A fix would be greatly appreciated.

mortenf1984:
This issue may seem like a small inconvenience, but when we have multiple branching promotions, this becomes annoying and time consuming. Can we please have  fix on this soon? It was working before just fine before 1.8 release

alishchytovych:
1.8.7 has it unfixed too

[...]

PR #5648, opened 2026-01-28
fix(ui): downstream promotion

Marvin9:
fixes https://github.com/akuity/kargo/issues/5413


Marvin9:
@krancour any objections of having backported in 1.9?

PR #5652, opened 2026-01-28
fix(controller): fix bug that wrongly judges workload identity to be available

### LINKED ISSUE #5647: Unable to authenticate to azure container registry after upgrade
# Checklist

# Description

Following an upgrade from v1.8.4 to v1.8.8 the Kargo warehouse is no longer able to authenticate to Azure Container Registry.

```
2026-01-28T09:58:15Z    ERROR   controller/controller.go:474    Reconciler error        {"controller": "warehouse", "controllerGroup": "kargo.akuity.io", "controllerKind": "Warehouse", "Warehouse": {"name":"<service-name>","namespace":"<project-namespace>"}, "namespace": "<project-namespace>", "name": "<service-name>", "reconcileID": "caa9f083-e584-494d-98c5-5041188a03d3", "error": "error discovering artifacts: error discovering artifacts: error obtaining credentials for image repo \"<acr url>/<image-name>\": error getting ACR access token: failed to get Azure AD access token for ACR: DefaultAzureCredential: failed to acquire a token.\nAttempted credentials:\n\tEnvironmentCredential: missing environment variable AZURE_TENANT_ID\n\tWorkloadIdentityCredential: no client ID specified. Check pod configuration or set ClientID in the options\n\tManagedIdentityCredential: failed to authenticate a system assigned identity. The endpoint responded with {\"error\":\"invalid_request\",\"error_description\":\"Multiple user assigned identities exist, please specify the clientId / resourceId of the identity in the token request\"}\n\tAzureCLICredential: executable not found on path\n\tAzureDeveloperCLICredential: executable not found on path\n\tAzurePowerShellCredential: executable not found on path"}
```

We do not use managed identity for Azure Container Registry, instead we use global scoped container registry credentials as per the documentation - https://docs.kargo.io/operator-guide/security/managing-credentials.

I've verified the ACR token is still valid, and that workload identity is disabled on the Kargo pods.

# Steps to Reproduce

- Setup ACR repository with global credentials and add it to warehouse
- Project can no longer connect to container registry with error as posted above

# Version

```
Client Version: v1.8.8
```

krancour:
Fixes #5647

The Azure Workload Identity credential provider was always intended to not register itself with the internal registry of credential providers when Azure Workload Identity is determined to be unavailable. This was not working as intended, essentially _always_ believing Workload Identity to be available.

Historically, this has not been a problem because the basic auth provider, which simply extracts username and password from a Secret (if it exists and has those fields in its data map), was explicitly registered _before_ the broken Azure Workload Identity credential provider. When attempting to resolve a credential request, credential providers are asked, in the order they were registered, "Can you handle this?" Thus, if basic auth credentials existed for an ECR repo, the credential request never made it to the broken Azure Workload Identity credential provider. All was right in the world.

v1.9.0 is using a new internal component registry, which has coincidentally changed the order in in which the cred providers are registered, which isn't inherently a problem. What _is_ a problem is the broken Azure Workload Identity credential provider happens to be registered before the basic auth credential provider now. This wouldn't actually be a problem if it worked, but it doesn't.

This one-liner changes the Azure Workload Identity credential provider's initialization in a way that verifiably fails in an environment without Azure Workload Identity available, which in turn prevents a broken Azure Workload Identity credential provider from appearing before the basic auth credential provider in the internal registry.

### DIFF
--- a/pkg/credentials/acr/workload_identity_test.go
+++ b/pkg/credentials/acr/workload_identity_test.go
@@ -3,6 +3,7 @@ package acr
 import (
 	"context"
 	"errors"
+	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


krancour:
Added tests.

Issue #5663, opened 2026-01-29
UI: subscription nodes are hidden


Marvin9:
# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate bug report.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

Subscription nodes are hidden in all cases

# Screenshots

<img width="1512" height="831" alt="Image" src="https://github.com/user-attachments/assets/d39e287e-72d8-4a8a-a76d-a1be09d6e833" />

# Steps to Reproduce

Visit project page

# Version

```
1.9.0
```

# Logs

```
Paste any relevant application logs here.
```

Marvin9:
Ok I realise having convo with @hiddeco that subscriptions on watch calls are not parsed properly and did fix but then this was changed again [here](https://github.com/krancour/kargo/blob/a02ac625dfe03b28735466412d888efedda1d103/pkg/server/watch_warehouses_v1alpha1.go)

Marvin9:
empty subscriptions in watch warehouse calls

<img width="1131" height="199" alt="Image" src="https://github.com/user-attachments/assets/df90b74d-d784-4144-a523-1914331c6408" />

PR #5664, opened 2026-01-29
fix(api): empty subscriptions

Marvin9:
fixes https://github.com/akuity/kargo/issues/5663


krancour:
Closed in favor of #5668

PR #5684, opened 2026-01-31
fix(controller): make token cache ttl more dynamic

### LINKED ISSUE #5610: github app authentication: transient git clone issues over https

krancour:
Fixes #5610

Supersedes #5615 

In many cases, though not all, we have access to a token's expiry date/time.

For such cases, it is safer to derive cache TTL durations from the expiry than to make assumptions about how long the token is valid for.

This also adds debug level logging with details of token expiry (if known) and corresponding cache entry TTL.

### DIFF
--- a/pkg/credentials/ecr/access_key.go
+++ b/pkg/credentials/ecr/access_key.go
@@ -110,14 +110,7 @@ func (p *AccessKeyProvider) GetCredentials(
 	}
 	logger.Debug("obtained new auth token")
 
-	// Cache the encoded token, preferring a TTL derived from the actual token
-	// expiry when available.
-	ttl := cache.DefaultExpiration
-	if !expiry.IsZero() {
-		if remaining := time.Until(expiry) - tokenCacheExpiryMargin; remaining > 0 {
-			ttl = remaining
-		}
-	}
+	ttl := credentials.CalculateCacheTTL(expiry, tokenCacheExpiryMargin)
 	logger.Debug(
 		"caching auth token",
 		"expiry", expiry,
--- a/pkg/credentials/ecr/managed_identity.go
+++ b/pkg/credentials/ecr/managed_identity.go
@@ -151,14 +151,7 @@ func (p *ManagedIdentityProvider) GetCredentials(
 	}
 	logger.Debug("obtained new auth token")
 
-	// Cache the encoded token, preferring a TTL derived from the actual token
-	// expiry when available.
-	ttl := cache.DefaultExpiration
-	if !expiry.IsZero() {
-		if remaining := time.Until(expiry) - tokenCacheExpiryMargin; remaining > 0 {
-			ttl = remaining
-		}
-	}
+	ttl := credentials.CalculateCacheTTL(expiry, tokenCacheExpiryMargin)
 	logger.Debug(
 		"caching auth token",
 		"expiry", expiry,
--- a/pkg/credentials/expiry.go
+++ b/pkg/credentials/expiry.go
@@ -1,0 +1,22 @@
+package credentials
+
+import (
+	"time"
+
+	"github.com/patrickmn/go-cache"
+)
+
+// CalculateCacheTTL calculates the time-to-live for a cached credential based
+// on the credential's expiry time and a safety margin. If the expiry time is
+// zero (unknown) or the remaining time after subtracting the margin is not
+// positive, it returns cache.DefaultExpiration, deferring to the cache's own
+// default TTL.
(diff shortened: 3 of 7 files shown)


pkg/credentials/ecr/access_key.go
```diff
@@ -93,9 +108,22 @@ func (p *AccessKeyProvider) GetCredentials(
...
 	}
-
-	// Cache the encoded token
-	p.tokenCache.Set(cacheKey, encodedToken, cache.DefaultExpiration)
+	logger.Debug("obtained new auth token")
+
+	// Cache the encoded token, preferring a TTL derived from the actual token
+	// expiry when available.
+	ttl := cache.DefaultExpiration
+	if !expiry.IsZero() {
```
hiddeco: We are doing this in multiple places, probably better to extract it to a helper:

```go
func CalculateCacheTTL(expiry time.Time, margin time.Duration) time.Duration {}
```

krancour: Fixed in https://github.com/akuity/kargo/pull/5684/commits/c0be8f675db69d94dd52aa4cb3e1188da367edaf

thomastaylor312 (approved):

Issue #5703, opened 2026-02-04
feat(chart): Still create Role/Rolebinding when opting out of creating shared/system resources namespace


shiehand:
__Note:__ If you're requesting a Helm chart option that may be very niche and not useful to the community at large, please consider using Kustomize to apply "last mile" tweaks to the output of `helm template` to suit your needs instead.

# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

In the Helm Chart, would like the Role/RoleBinding for the system / shared resources namespace to be still created when `createNamespace` is set to `false`.

We can do this by removing the `if` statement at the [start of the Role/RoleBinding templates](https://github.com/akuity/kargo/blob/5ad52dfbbf20eb1c4e5794b155d15c0949d59229/charts/kargo/templates/shared-resources-namespace/roles.yaml#L1).
Noticed that this was the case in [the `cluster-secrets-namespace` templates](https://github.com/akuity/kargo/blob/5ad52dfbbf20eb1c4e5794b155d15c0949d59229/charts/kargo/templates/cluster-secrets-namespace/roles.yaml#L1).

# Motivation

Similar to https://github.com/akuity/kargo/issues/4839, we want to opt out of the chart creating the new system and shared resources namespace by setting `.Values.global.sharedResources.createNamespace` to false.
We do this because we have a dedicated process in creating namespaces, so we would like to create the namespace separately and have Kargo use it afterwards.

However the [creation of the Role/RoleBinding](https://github.com/akuity/kargo/blob/5ad52dfbbf20eb1c4e5794b155d15c0949d59229/charts/kargo/templates/shared-resources-namespace/roles.yaml#L1) seems to be also tied to the `createNamespace` value, thus when opting out we would have to define the Role/RoleBinding ourselves.
I noticed that [this was not the case](https://github.com/akuity/kargo/blob/5ad52dfbbf20eb1c4e5794b155d15c0949d59229/charts/kargo/templates/cluster-secrets-namespace/roles.yaml#L1) on the `cluster-secrets-namespace` template and I wonder if the change in the new namespaces templates are unintended.

Thanks for developing a great tool guys!

PR #5705, opened 2026-02-05
fix: always create role + rolebinding for shared resources namespace

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5703

This was unintentional as it's supposed to follow the same pattern as the system-resources-namespace e.g. https://github.com/akuity/kargo/blob/main/charts/kargo/templates/system-resources-namespace/roles.yaml + https://github.com/akuity/kargo/blob/main/charts/kargo/templates/system-resources-namespace/role-bindings.yaml

### DIFF
--- a/charts/kargo/templates/shared-resources-namespace/role-bindings.yaml
+++ b/charts/kargo/templates/shared-resources-namespace/role-bindings.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.global.sharedResources.createNamespace }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
@@ -43,5 +42,4 @@ subjects:
 - kind: ServiceAccount
   namespace:  {{ .Release.Namespace }}
   name: kargo-api
-{{- end }}
 {{- end }}


krancour:
Marked this as "needs discussion."

This might be right, but I want us to discuss and put some careful thought into this, because I think there's _also_ an argument to be made that this being conditional wasn't wrong and that in the _other places_ where it is not conditional, it should have been.

Let's hash it out at the next thunderdome.

krancour:
After further consideration, this is indeed the right approach.

The impetus for people choosing not to have the chart create shared + system resource namespaces would be the same as that for not using the `--create-namespace` flag on a `helm install` or the `CreateNamespace` sync option in Argo CD to create Kargo's _own_ namespace -- namely organization-specific methods or policies for how namespaces must be created. 

_But_ just as not using the `--create-namespace` flag on a `helm install` most definitely does _not_ mean that the chart also shouldn't install anything _into_ Kargo's own namespace, the same is true here. _Creation_ of shared + system resource namespaces may be handled externally, but the chart's responsibility for populating them with resources required for Kargo to function correctly remains.

PR #5710, opened 2026-02-07
fix: Fixed chart repoURL pattern regex

neboman11:
## Issue

When trying to create a new Warehouse with a chart Subscription, the form will not allow any submissions, even with proper values entered. It always shows the error `unterminated parenthetical` when hitting "Add Subscription". The issue is the regex that backs that field is missing a closing parenthesis.

Screenshot_20260206_223619

## Solution

Added parenthesis to end of regex to close domain and port clause.

`^(((https?)|(oci))://)(([\\w\\d\\.\\-]+)(:([\\d]+)?)?(/.*)*$`

vs.

`^(((https?)|(oci))://)(([\\w\\d\\.\\-]+)(:([\\d]+)?)?(/.*)*)$`

### DIFF
--- a/pkg/subscription/schemas/chart.json
+++ b/pkg/subscription/schemas/chart.json
@@ -8,7 +8,7 @@
         "repoURL": {
             "type": "string",
             "minLength": 1,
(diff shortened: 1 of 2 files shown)


neboman11:
@hairyhum it looks like some of the CI steps failed because of a GitHub hiccup. Could you rerun them please?

Issue #5630, opened 2026-01-26
Support for tpl on `api.host`


marquesj2-ppb:
__Note:__ If you're requesting a Helm chart option that may be very niche and not useful to the community at large, please consider using Kustomize to apply "last mile" tweaks to the output of `helm template` to suit your needs instead.

# Checklist

* [ ] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Would it be possible to have `api.host` support `tpl` ?

# Motivation

We have base hostname global values injected to our pipeline that we can leverage to construct hostname for all services, instead of having to adapt the pipeline to include kargo specific overrides we would simply be able to `api.host: kargo.{{ .Values.global.internalBaseDomain }}` in the values

krancour:
@marquesj2-ppb with apologies, I am not quite clear on what you're requesting here. Could you elaborate please?

marquesj2-ppb:
We currently install Kargo in multiple clusters and would like for an easy way to set up the `api.host` without having to change our pipelines just for kargo.
We inject global values that have a field like `cluster.internalAppDomain` that would be used by `api.host` if it supported `tpl` on the fields it is used, like https://github.com/akuity/kargo/blob/5ad52dfbbf20eb1c4e5794b155d15c0949d59229/charts/kargo/templates/api/ingress.yaml#L21

We would be able to set the value like
```yaml
api:
  host: kargo.{{ .Values.global.cluster.internalAppDomain }}
```

Alternatively if this value was read from a helper similar to
https://github.com/akuity/kargo/blob/5ad52dfbbf20eb1c4e5794b155d15c0949d59229/charts/kargo/templates/_helpers.tpl#L47-L52

We would be able to override the helper in our side.

krancour:
You're asking our chart to accommodate something specific to your org?

marquesj2-ppb:
Not our org specifically, anyone who wants to use it.
ArgoCD helm chart already supports `tpl` on host

marquesj2-ppb:
I opened a PR with the proposed changes #5718 , wasn't expecting anyone to the the work for us, was following the guide of creating an issue before

krancour:
I'm sorry. I know this is an answer you won't care for, but I struggle to see the incredibly modest benefit of this as justifying the added complexity. This is the exact sort of change we were discouraging with the advisory at the top of the issue template:

[...]

PR #5718, opened 2026-02-10
feat(chart): allow for template on host input

jdcmarques:
## Description
- Allow for hostnames to be helm temptable
  - `api.host`
  - `api.argocd.urls{*}`
  - `externalWebhooksServer.host`
  
Previously the fields must have been set with
```yaml
api:
  host: kargo.example.com
externalWebhooksServer:
  host: webhooks.kargo.example.com
```

We can now do something like:

```yaml
global:
  baseDomain: 'example.com'

api:
  host: "kargo.{{ .Values.global.baseDomain }}"
externalWebhooksServer:
  host: "webhooks.kargo.{{ .Values.global.baseDomain }}"  
```

### Validation
Rendered before and after proposed changes and had no diff in the outcome templates
- `api.oidc.enabled` - `true` and `false`
- `api.oidc.dex.enabled` - `true` and `false`
- `api.tls.enabled.selfSignedCert` - `true` and `false`
- `externalWebhooksServer.tls.enabled` - `true` and `false`
- `externalWebhooksServer.tls.selfSignedCert` - `true` and `false`

Addresses #5630

### DIFF
--- a/charts/kargo/templates/_helpers.tpl
+++ b/charts/kargo/templates/_helpers.tpl
@@ -45,10 +45,24 @@ Generate base URL for a service.
 {{- end -}}
 
 {{/*
+Generate the base HOST for the API service.
+*/}}
+{{- define "kargo.api.baseHost" -}}
+{{- (tpl .Values.api.host .) -}}
+{{- end -}}
+
+{{/*
 Generate the base URL for the API service.
 */}}
 {{- define "kargo.api.baseURL" -}}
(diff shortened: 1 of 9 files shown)


krancour:
Were this proposed chart feature, originally described for #5630, to be accepted, this PR isn't the implementation of it that I would prefer, so I am closing this to keep conversation in one place - #5630 - where we can sort out _if_ we should do this separately from _how_ we would do it.

PR #5726, opened 2026-02-11
feat(ui): infer git provider for self-hosted deep links

### LINKED ISSUE #5480: Deep links to self-hosted github/gitlab should work automatically through inference
# Proposed Feature

Currently, unless `github.com` is the repo URL's domain name, we won't construct a proper deep link to the actual commit, and the Freight will link to the base URL instead of to the commit. So users who self-host github / gitlab will not get proper deep links.

We should expand this so that if the word "github" is in the domain name, we can make an assumption about the how the URL is formed, in order to construct a proper deeplink.

Same should be true for other vendor URL formats, e.g. "gitlab." GitLab has an even greater chance of being self-hosted, with non-'gitlab.com.com` domains.

# Motivation

1. It very common that whenever someone self-hosts github or gitlab, they incorporate "github" and "gitlab" into the domain name (e.g. github.mycompany.com). 
2. Things will just work OOB.

singhvibhanshu:
## Description
Fixes broken deep links for self-hosted and enterprise Git providers (e.g., GitLab Self-Managed, GitHub Enterprise).

Currently, the UI does an exact match on the domain (e.g., `domain === 'github.com'`), which fails for custom domains like `gitlab.internal.net`. This PR updates `getGitCommitURL` to handle these cases by checking if the hostname *includes* the provider keyword rather than requiring an exact match.

## Changes
- **Inference Logic:** The utility now detects provider keywords (e.g., `gitlab`, `github`, `bitbucket`) within custom domains to apply the correct URL formatting.
    - *Example:* `gitlab.internal.net` is detected as GitLab and gets the required `/-/commit/` path.
- **SSH Support:** Added logic to normalize SSH remote URLs (`git@...`) into clickable HTTPS links.
- **Refactor:** Switched to `URL` parsing for safer domain extraction.

## Verification
I verified this locally by spinning up a `tilt` environment and creating a mock Freight pointing to a self-hosted GitLab instance (`gitlab.internal.net`).

**1. Logic Verification:**
Console logs showing the utility correctly identifying providers from custom domains.
Screenshot from 2026-02-11 15-13-15

**2. End-to-End UI Verification:**
- Created a mock Freight pointing to `gitlab.internal.net`.
- Clicking the commit hash successfully navigated to `https://gitlab.internal.net/.../-/commit/...`, confirming the GitLab-specific `/-/` segment was applied.
sa

## Related Issue
Closes #5480

### DIFF
--- a/ui/src/features/freight-timeline/open-container-initiative-utils.ts
+++ b/ui/src/features/freight-timeline/open-container-initiative-utils.ts
@@ -29,29 +29,38 @@ export const getImageSource = (annotation: Annotation) => {
 };
 
 export const getGitCommitURL = (url: string, revision: string) => {
-  let baseUrl;
-
-  if (url.includes('github.com')) {
-    baseUrl = url
-      .replace(/^git@github.com:/, 'https://github.com/')
-      .replace(/^https?:\/\/github.com\//, 'https://github.com/')
-      .replace(/\.git$/, '');
-    return `${baseUrl}/commit/${revision}`;
-  } else if (url.includes('gitlab.com')) {
-    baseUrl = url
-      .replace(/^git@gitlab.com:/, 'https://gitlab.com/')
-      .replace(/^https?:\/\/gitlab.com\//, 'https://gitlab.com/')
-      .replace(/\.git$/, '');
-    return `${baseUrl}/-/commit/${revision}`;
-  } else if (url.includes('bitbucket.org')) {
-    baseUrl = url
-      .replace(/^git@bitbucket.org:/, 'https://bitbucket.org/')
-      .replace(/^https?:\/\/bitbucket.org\//, 'https://bitbucket.org/')
-      .replace(/\.git$/, '');
-    return `${baseUrl}/commits/${revision}`;
+  if (!url) {
+    return '';
   }
 
-  return url;
+  let baseUrl = url;
+
+  if (baseUrl.startsWith('git@')) {
+    baseUrl = baseUrl.replace(':', '/').replace(/^git@/, 'https://');
+  } else if (!baseUrl.startsWith('http://') && !baseUrl.startsWith('https://')) {
+    baseUrl = `https://${baseUrl}`;
+  }
+
+  baseUrl = baseUrl.replace(/\.git$/, '');
+
+  try {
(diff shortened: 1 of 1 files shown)


krancour:
This is an improvement over what we started with, however, there is a more complete/robust implementation of this algorithm on the back end. I'm wondering if it makes sense for the API server to expose a small utility endpoint for constructing navigable commit URLs. I _vaguely_ recall having had a conversation about this at some point in the past. We may not have. If we did, I don't remember the rationale that we landed on for _not_ doing that.

I'm wondering if @rpelczar or @Marvin9 knows why we're doing it this way. Was it that we need to do this often and the UI would be overly chatty with the API server?

Regardless of the answers, front end is @rpelczar's and @Marvin9's domain. You guys decide what's best here and if you happen to prefer a back end endpoint for this, let me know.

Marvin9:
@krancour I don't recall that discussion maybe @rpelczar can. But that should be ideal situation, only if url construction is user provided. Thats for the reason we can't just guess based on the fact that url contains "github" or "gitlab" word

krancour:
@rpelczar and I discussed this morning and he suggested using [git-url-parse](https://www.npmjs.com/package/git-url-parse) for this. It's actually in our `package.json` already, but seems not to be used anywhere yet.

Building out a utility endpoint in our API server to unify how the front end and back end parse and construct git URLs, that's something we'll defer to a later time, if we do it at all.

Marvin9:
I need to check but most probably that package won't detect anything outside of providers domain, what this PR is trying to achieve

singhvibhanshu:
@Marvin9 
I believe that standard parsers typically rely on the hardcoded list of known domains to determine provider type...
Parser will generally consider `gitlab.internal.net` as a generic Git server and will probably fails to apply provider specific formatting.
This PR ensures to detect the provider even when the custom domain is being used, ensuring the generation of correct deep links

rpelczar:
@Marvin9 It covers custom domains. Just look at the test here: https://github.com/argoproj/argo-cd/blob/master/ui/src/app/shared/components/urls.test.ts

krancour:
Closing this since it seems @rpelczar and @Marvin9 prefer a different direction for resolving this.

PR #5730, opened 2026-02-12
test: add unit tests for project webhook validation

singhvibhanshu:
## Explanation
This PR introduces unit tests for the `pkg/webhook/kubernetes/project.go file`. The goal is to ensure the `ValidateProject` function correctly handles validation outcomes and maps internal errors (such as `ErrProjectNotFound`) to the appropriate Kubernetes API errors (like `404 NotFound`).

## Proposed Changes
1. Added `pkg/webhook/kubernetes/project_test.go`.
2. Implemented `TestValidateProject` using the controller-runtime `fake` client to simulate:
    - Success Scenario: Validating an object in an existing, labeled Project namespace.
    - Failure Scenario: Validating an object in a non-existent or invalid namespace, ensuring it returns a proper `apierrors.IsNotFound` result.

## Test Results
```
=== RUN   TestValidateProject
=== RUN   TestValidateProject/success:_project_exists
=== RUN   TestValidateProject/error:_project_not_found_(namespace_missing)
=== RUN   TestValidateProject/error:_namespace_exists_but_is_not_a_project_(label_missing)
--- PASS: TestValidateProject (0.05s)
    --- PASS: TestValidateProject/success:_project_exists (0.05s)
    --- PASS: TestValidateProject/error:_project_not_found_(namespace_missing) (0.00s)
    --- PASS: TestValidateProject/error:_namespace_exists_but_is_not_a_project_(label_missing) (0.00s)
PASS
ok      github.com/akuity/kargo/pkg/webhook/kubernetes  (cached)
```

### DIFF
--- a/pkg/webhook/kubernetes/project_test.go
+++ b/pkg/webhook/kubernetes/project_test.go
@@ -1,0 +1,99 @@
+package webhook
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestValidateProject(t *testing.T) {
+	const testPodName = "test-pod"
+	scheme := runtime.NewScheme()
+	require.NoError(t, corev1.AddToScheme(scheme))
+
+	testCases := []struct {
+		name        string
+		objects     []client.Object
+		targetObj   client.Object
+		expectedErr func(error) bool
+	}{
+		{
+			name: "success: project exists",
+			objects: []client.Object{
+				&corev1.Namespace{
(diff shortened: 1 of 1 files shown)


krancour:
@singhvibhanshu Was there something specific that motivated this PR? While more coverage is always welcome, I want to flag some context that might affect whether this work is worthwhile right now.

The function you're testing delegates to validation.ValidateProject, which implements a very outdated validation approach. Early in Kargo's life, the Project type didn't exist — "Project" was just a specially labeled namespace. This function still checks for that specially labeled namespace, but since the introduction of the Project type quite some time ago, it would be more straightforward to check for that directly.

Rather than testing the current implementation, I'd prefer to see it refactored first, with tests written against the new approach.

To complicate things further: aside from webhooks, most (possibly all) other usage of validation.ValidateProject is in endpoints of the legacy gRPC API or their tests. That entire API is deprecated and slated for removal in v1.12.0.

So the bottom line is this: there are significant changes pending in this area of the codebase, and unless you have a specific reason for adding these tests — like a known bug or failure — I'd suggest leaving this alone for now. It will all be refactored within a few months.

singhvibhanshu:
@krancour thanks for the context!

No, I didn't encounter any specific bug or failure. I was simply just exploring the codebase, noticed this file has missing test, and thought it would be great to contribute in order to fill the gap.

But yeah, totally makes sense not to bother testing deprecated code that's getting removed in the new version soon. I'll go ahead and close this PR.

If there are other areas that actually needs test coverage or some work, please let me know.
I'd happy to help out there instead!

Issue #5671, opened 2026-01-30
CLI download link should not point to "latest", but to its current version instead


jwaibel-pons:
# Checklist

* [x ] I've searched the issue queue to verify this is not a duplicate bug report.
* [x ] I've included steps to reproduce the bug.
* [x ] I've pasted the output of `kargo version`.
* [x ] I've pasted logs, if applicable.

# Description

The links on the CLI download page in the UI point to "latest", which could lead to downloading mismatching versions.

# How to reproduce

This is rather a mild annoyance than a bug. nevertheless it just did cost me some time to figure out what's going on.

We are still running Kargo 1.8.6 here. When I go to the CLI download page, links are pointing to the latest version, e.g. https://github.com/akuity/kargo/releases/latest/download/kargo-linux-amd64

This will download the current version of the CLI (1.9.0 as of today).

As you pointed out in the release notes _"CLI Upgrade Required: Due to unavoidable breaking changes in how Warehouse and Freight types are serialized as protocol buffers, users upgrading the Kargo back end to v1.9.0 must also [upgrade their CLI](https://github.com/akuity/kargo/releases/tag/v1.9.0)."_ the CLI has breaking changes.

And indeed, all I got from the CLI was

`Error: error retrieving public configuration from server: &{false <nil> false} (*models.PublicConfig) is not supported by the TextConsumer, can be resolved by supporting TextUnmarshaler interface`

As the UI knows its version, it should easily be able to put that into the link, like

https://github.com/akuity/kargo/releases/download/v1.8.6/kargo-linux-amd64

# Screenshots

<img width="1180" height="376" alt="Image" src="https://github.com/user-attachments/assets/d55f408d-0414-429c-ae0f-d9defee8f434" />

rpelczar:
I think we can add a version selector with the current version as a default option.

krancour:
@rpelczar I looked into this a little bit just now. There isn't a single, convenient URL that can give you the latest patch release of every n minors, excluding pre-releases. The whole list of releases needs to be downloaded and filtered client side. Not pleasant, but doable.

But while investigating that, I was also reminded that unauthenticated requests to GitHub APIs are rate-limited to 60/hour per IP. In orgs where everyone isn't remote, or in orgs where remote workers all use a VPN, that could actually be a problem because you'll have lots and lots of users that GH sees as all having the same IP.

[...]

PR #5739, opened 2026-02-14
feat(api): add best releases endpoint

krancour:
This is related to #5671 

This is an API endpoint that returns a list of the "best" Kargo releases -- meaning the latest patch of every minor. This includes CLI download links in a map indexed by OS and Arch.

The latest release and the one matching the version of the API server are both highlighted with fields `latest` and `current`, respectively.

Responses look like this:

```json
{
  "releases": [
    {
      "version": "v1.9.2",
      "latest": true,
      "cliBinaries": {
        "darwin": {
          "amd64": "https://github.com/akuity/kargo/releases/download/v1.9.2/kargo-darwin-amd64",
          "arm64": "https://github.com/akuity/kargo/releases/download/v1.9.2/kargo-darwin-arm64"
        },
        "linux": {
          "amd64": "https://github.com/akuity/kargo/releases/download/v1.9.2/kargo-linux-amd64",
          "arm64": "https://github.com/akuity/kargo/releases/download/v1.9.2/kargo-linux-arm64"
        },
        "windows": {
          "amd64": "https://github.com/akuity/kargo/releases/download/v1.9.2/kargo-windows-amd64.exe",
          "arm64": "https://github.com/akuity/kargo/releases/download/v1.9.2/kargo-windows-arm64.exe"
        }
      }
    },
    {
      "version": "v1.8.10",
      "current": true,
      "cliBinaries": {
        "darwin": {
          "amd64": "https://github.com/akuity/kargo/releases/download/v1.8.10/kargo-darwin-amd64",
          "arm64": "https://github.com/akuity/kargo/releases/download/v1.8.10/kargo-darwin-arm64"
        },
        "linux": {
          "amd64": "https://github.com/akuity/kargo/releases/download/v1.8.10/kargo-linux-amd64",
          "arm64": "https://github.com/akuity/kargo/releases/download/v1.8.10/kargo-linux-arm64"
        },
        "windows": {
          "amd64": "https://github.com/akuity/kargo/releases/download/v1.8.10/kargo-windows-amd64.exe",
          "arm64": "https://github.com/akuity/kargo/releases/download/v1.8.10/kargo-windows-arm64.exe"
        }
      }
    },

...

  ]
}
```

@rpelczar or @Marvin9 which ever of you ends up working on the UI changes for #5671, you can amend this PR directly if you want to. Or I can merge it before then. Just let me know whichever you prefer.


Marvin9:
I will amend commit for UI

Marvin9:
Sorry I might have missed something but why do we need an API? We already have construct link, we just need to give current version to that function.

```js
const constructUrl = (platform: string, version?: string) =>
  `https://github.com/akuity/kargo/releases/${version ? '' : 'latest/'}download/${version ? `${version}/` : ''}${platform}`;
```

krancour:
@Marvin9 good question. There are a few reasons, which I failed to capture here. Some are discussed in #5671, but not all.

In terms of functionality, this does more than simply constructing download URLs from a given version number. This finds the _best_ (i.e. latest patch) for every minor release. This way we're always nudging users toward the least buggy and most secure version of the CLI for their current/chosen minor release.

This will also identify which is latest release and which one matches the current UI/API server.

You _could_ do all of that in the front end, but the GitHub API endpoints you would have to access yourself to do all of this are rate limited at 60/hour unless your are authenticated. That could pose a in orgs that are not fully remote or use a VPN, as there may be many more than 60 requests per hour originating from a single public IP.

The back end caches this list for an hour at a time and also (mainly for EE's benefit) offers the option to authenticate to GitHub using a GitHub App to increase the rate limit to 5,000/hour.

krancour:
@thomastaylor312 offered up an alternative to this. We could publish JSON with the "best" versions to GitHub pages and keep it up to date automatically via our release process. The only thing we couldn't easily do this way is set "current" to match the version of the server that user has. I think that would be relatively easy to do on the front end?

@rpelczar @Marvin9 I'm happy to go either way on this, so lmk why of the two approaches is more convenient for you.

__Edit:__ I verified already that there are no CORS issues with GitHub Pages.

Marvin9:
Thanks. I think Github pages would be better. UI knows which version server is running so it won't be an issue for UI.

Screenshot 2026-02-20 at 1 18 16 PM

krancour:
Closed in favor of #5797

PR #5754, opened 2026-02-18
fix(controller): direct api to avoid stale promo object

shamsalmon:
Closes: https://github.com/akuity/kargo/issues/5282

I am sure there is a more elegant solution than this but I wanted to get a solution out there. We are still testing this, but so far looking solid.

I noted on the issue why I believe this is occurring and I can confirm the "direct" api object is correct. However my initial assumptions were wrong.

Here is a timeline:

1. Promotion starts
2. "Running" - Status update sent
3. Promotion works on something that requires a wait (git-wait-for-pr, argocd-sync) - Sends another status update
4. Immediate reconciliation triggers, however informer does not have latest status / metadata.
5. Reconciliation runs against the stale / cached promotion object which has no metadata

Using the apiReader (which already exists) we can ensure that we are getting the real and non cached version of the object.

### DIFF
--- a/pkg/controller/promotions/promotions.go
+++ b/pkg/controller/promotions/promotions.go
@@ -61,6 +61,7 @@ func ReconcilerConfigFromEnv() ReconcilerConfig {
 type reconciler struct {
 	kargoClient    client.Client
 	promoEngine    promotion.Engine
+	apiReader      client.Reader
 	shardPredicate controller.ResponsibleFor[kargoapi.Promotion]
 
 	cfg ReconcilerConfig
@@ -116,6 +117,7 @@ func SetupReconcilerWithManager(
 
 	reconciler := newReconciler(
 		kargoMgr.GetClient(),
+		kargoMgr.GetAPIReader(),
 		k8sevent.NewEventSender(
 			libEvent.NewRecorder(ctx, kargoMgr.GetScheme(), kargoMgr.GetClient(), cfg.Name()),
 		),
@@ -188,12 +190,14 @@ func SetupReconcilerWithManager(
 
 func newReconciler(
 	kargoClient client.Client,
+	apiReader client.Reader,
 	sender event.Sender,
 	promoEngine promotion.Engine,
 	cfg ReconcilerConfig,
 ) *reconciler {
 	r := &reconciler{
 		kargoClient: kargoClient,
+		apiReader:   apiReader,
 		promoEngine: promoEngine,
 		sender:      sender,
 		cfg:         cfg,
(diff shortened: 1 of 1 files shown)


krancour:
@shamsalmon you lost me at step 4. How or why is the Promotion being re-reconciled immediately? I agree this could theoretically cause the problem you've been observing, but I can't quite see how or why immediate re-reconciliation would happen in the first place.

krancour:
Hmmm. Is it from mashing refresh maybe?

krancour:
Or I supposed the potential exists that hitting it even once could possibly cause this.

shamsalmon:
I am sure my users are being aggressive with the refresh button but ive seen this happen just by chance pressing it once as you say.

shamsalmon:
I think several factors probably make this worse, including latency from kargo controller to targeted cluster in a sharded setup. We probably run 50-200 promotions a day (100ish users) and see this happen once or twice a day.

krancour:
I'm fairly certain the refresh has something to do with this. For one, it could mean a promo currently being reconciled is added to the work queue so that there is that immediate re-reconciliation you spoke of. Two, I believe _handling_ of the refresh at the head end of the reconciliation process involves a programmatic immediate requeue. I don't have the code in front of me at the moment, but I plan to dig into whether or how these are individually, or in combination, contributing to this. We can potentially consider your proposed change as a stop gap, but I feel we're really close to the smoking gun here and may be able to do something more strategic.

shamsalmon:
Sounds good! I would dig into this more but I need to switch my focus onto some other priorities. If you do have another idea I would be happy to test it out for you.

EronWright:
@shamsalmon I reviewed the evidence and commentary, and of course this PR.  The use of a direct reader is indeed a possible solution to what is a cache-related issue (see [controller-runtime#1464](https://github.com/kubernetes-sigs/controller-runtime/issues/1464) for more discussion).

Based on the evidence, the likely root cause is the Refresh button being clicked while the promotion is running, especially in an early, time-consuming step like git-clone.  I opened https://github.com/akuity/kargo/pull/5934 with an alternative solution based on the upstream discussion.  If that's accepted, we'll close your PR.

Thanks again for your efforts here.

shamsalmon:
hey @krancour are you planning on adding this to 1.10.0? Would be great to see it added to the release. Let me know if you need anything from me.

EronWright:
Closing in favor of #6029, which targets `main` (this PR was targeting `release-1.9`). The core apiReader approach from this PR is preserved there — thanks @shamsalmon!

krancour:
> this PR was targeting release-1.9

For future reference, it's _very_ hidden, but it's possible to change the target branch of an open PR. If you click the pencil next to the title as if intending to edit the title, there's a mutable drop down under the title for the target branch.

PR #5755, opened 2026-02-18
refactor(devx): streamline local dev setup

krancour:
clone --> running, much faster

@thomastaylor312 this is based on a conversation we had not long ago about making Tilt take over installation of prereqs into the local dev cluster...

* Make targets `hack-kind-up` and `hack-k3-up` now, like most other Make targets of ours, will install suitable versions of their own binary dependencies in `hack/bin/`. This is one less thing a contributor needs to have pre-installed.

  These targets will _no longer_ install prerequisite software into the cluster using `helm`. (Read on. There's a better way now.)

* Users of Orbstack and Docker Desktop previously needed to run `make hack-install-prereqs` to install prerequisite software into the cluster using `helm`. They no longer do. This makes the experience of using these two options more similar to the experience using the other two.

* New make target `make hack-tilt-up` will ensure adequate versions of `tilt` and `helm` are installed into `hack/bin/`. Again -- less the contributor needs to have pre-installed. (You can keep running `tilt up` directly if you don't care about this.)

* `Tiltfile` takes over responsibility for installing prerequisite software into the cluster using `helm`. Care was taken that this doesn't slow down subsequent `make hack-tilt-up` (or direct `tilt up`.)

Assuming they have `make`, a new contributor can get Kargo built from source running on kind or k3d cluster with just two commands. Those using Orbstack or Docker Desktop can get there in just one.

All docs updated accordingly.

Why the backport? I want the doc changes to get to the production docs site right away.

### DIFF
--- a/docs/docs/60-contributor-guide/10-hacking-on-kargo.md
+++ b/docs/docs/60-contributor-guide/10-hacking-on-kargo.md
@@ -228,7 +228,7 @@ this.
 
     The `hack-kind-up` target will ensure the installation of suitable versions
     of `kind` and [ctlptl](https://github.com/tilt-dev/ctlptl#how-do-i-install-it)
-    (used for declarative kind configuration) into `hack/bind/`.
+    (used for declarative kind configuration) into `hack/bin/`.
 
     :::
 
@@ -261,7 +261,7 @@ this.
     The `hack-k3d-up` target will ensure the installation of suitable versions
     of `k3d` and
     [ctlptl](https://github.com/tilt-dev/ctlptl#how-do-i-install-it) (used for
-    declarative k3d configuration) into `hack/bind/`.
+    declarative k3d configuration) into `hack/bin/`.
 
     :::
 
@@ -327,7 +327,7 @@ this.
     :::info
 
     The `hack-tilt-up` target will ensure the installation of a suitable
-    versions of `tilt` and `helm` into `hack/bind/`.
+    versions of `tilt` and `helm` into `hack/bin/`.
 
     When run for the first time on a new development cluster, `make
     hack-tilt-up` will install suitable versions of


docs/docs/60-contributor-guide/10-hacking-on-kargo.md
```diff
@@ -241,10 +226,18 @@ this.
 
     :::info
 
-    While this option is a bit more complex than using Docker Desktop or OrbStack
-    directly, it offers the advantage of being fully-disposable. If your cluster
-    reaches a state you are dissatisfied with, you can simply destroy it and
-    launch a new one.
+    The `hack-kind-up` target will ensure the installation of suitable versions
+    of `kind` and [ctlptl](https://github.com/tilt-dev/ctlptl#how-do-i-install-it)
+    (used for declarative kind configuration) into `hack/bind/`.
```
thomastaylor312: Looks like maybe a fat finger and this should be `hack/bin` and not `hack/bind`

docs/docs/60-contributor-guide/10-hacking-on-kargo.md
```diff
@@ -269,29 +258,24 @@ this.
...
-    reaches a state you are dissatisfied with, you can simply destroy it and
-    launch a new one.
-    :::
-
-    </TabItem>
-    </Tabs>
+    The `hack-k3d-up` target will ensure the installation of suitable versions
+    of `k3d` and
+    [ctlptl](https://github.com/tilt-dev/ctlptl#how-do-i-install-it) (used for
+    declarative k3d configuration) into `hack/bind/`.
```
thomastaylor312: Same here with the path

docs/docs/60-contributor-guide/10-hacking-on-kargo.md
krancour: Fat fingered it once and proceeded to paste it two other places. 🤦‍♂️ 

Fixed in 97c05fd3eca03a8d37b639da9729788565e96dd1

thomastaylor312 (approved):

Issue #4627, opened 2025-07-16
Annotation keep-namespace doesn't work


vietwow:
Hi, I'm deploying kargo chart version 1.6.1 in EKS 1.32

In https://docs.kargo.io/user-guide/how-to-guides/working-with-projects/#preventing-namespace-deletion, it mentioned :
```
By default, when a Project is deleted, Kargo will attempt to delete the corresponding Namespace. However, there are scenarios where you may want to retain the namespace after the associated Project is removed whether it was created by Kargo or adopted from an existing setup.

To achieve this, you can apply the following annotation to the Namespace or the corresponding Project resource with kargo.akuity.io/keep-namespace: "true".
```

Here is my project config :
<img width="1333" height="642" alt="Image" src="https://github.com/user-attachments/assets/42f232bf-cdff-4b7d-86b7-0de867a59ad1" />

when I try to delete the project, then the namespace is still deleted :

<img width="1496" height="662" alt="Image" src="https://github.com/user-attachments/assets/144c1780-303d-4694-8a56-01e07938c60d" />
wait some seconds, argocd recreate all resources :
<img width="1486" height="887" alt="Image" src="https://github.com/user-attachments/assets/140d8a71-9f05-4f2a-9ee3-9f6edbd2e7e8" />

Any ideas on this ? Thank you.

Best Regards,

hiddeco:
Are you sure the Namespace is actually deleted? The first screenshot appears to show an admission webhook error (below the `Syncing` that I would love to learn more about.

vietwow:
> Are you sure the Namespace is actually deleted? The first screenshot appears to show an admission webhook error (below the `Syncing` that I would love to learn more about.

Ya, I know the first image is a bit confuse, I'm also concerning about that webhook error. But in the second image, ns was deleted so ArgoCD re-sync and re-create ns (and other resources), you can see all resources are created in "a few seconds ago". And of course, I tested it several times. Here is another test, resources are deleted :

<img width="1035" height="733" alt="Image" src="https://github.com/user-attachments/assets/48d38970-84e8-409f-81ef-d093ec5c2f36" />

then ArgoCD reconcile it :

<img width="1452" height="722" alt="Image" src="https://github.com/user-attachments/assets/028e8b0f-ea56-4e28-9f7e-fc456c7a5f9f" />

Please let me know if you need any more info or my configs. Thanks.

Best Regards,

vietwow:
Here is logs of controllers and it said ns is terminated : 

[...]

PR #5763, opened 2026-02-19
fix: keep-me annotation not blocking delete of namespace

fuskovic:
Closes https://github.com/akuity/kargo/issues/4627

### Resources used

```yaml
kind: Project
metadata:
  name: keep-me
  annotations:
    kargo.akuity.io/keep-namespace: "true"
---
apiVersion: v1
kind: Namespace
metadata:
  name: keep-me
  labels:
    istio.io/dataplane-mode: ambient
    kargo.akuity.io/project: "true"
```

#### Reproducing the issue

This demo demonstrates the `keep-me` namespace getting deleted when the associated `keep-me` project is deleted with `--cascade=foreground` . This should not happen due to to the presence of the `keep-me` annotation.

https://github.com/user-attachments/assets/2d6017d2-14b8-4daf-8c44-d3ef5091073c

#### The Fix

This demo contains the fix and demonstrates the `keep-me` namespace NOT getting deleted when the associated `keep-me` project is deleted with `--cascade=foreground` .

https://github.com/user-attachments/assets/8396040f-40c9-40a0-a9df-c33dd7dd622f

### DIFF
--- a/pkg/controller/management/namespaces/namespaces.go
+++ b/pkg/controller/management/namespaces/namespaces.go
@@ -140,7 +140,7 @@ func (r *reconciler) Reconcile(
 	}
 	hasProjectOwnerRef, err := controllerutil.HasOwnerReference(ns.OwnerReferences, project, r.client.Scheme())
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to check for owner reference on namespace %q: %w", ns.Name, err)
+		return ctrl.Result{}, fmt.Errorf("failed to check for project owner reference on namespace %q: %w", ns.Name, err)
 	}
 	if hasProjectOwnerRef {
 		logger.Debug("Namespace has project owner reference; removing")
@@ -150,7 +150,7 @@ func (r *reconciler) Reconcile(
 			r.client.Scheme(),
 		)
 		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to remove owner reference from namespace %q: %w", ns.Name, err)
+			return ctrl.Result{}, fmt.Errorf("failed to remove project owner reference from namespace %q: %w", ns.Name, err)
 		}
 		if err := r.patchOwnerReferencesFn(ctx, r.client, ns); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to patch owner references for namespace %q: %w", ns.Name, err)


pkg/controller/management/namespaces/namespaces_test.go
```diff
@@ -71,14 +93,118 @@ func TestReconcile(t *testing.T) {
...
 			reconciler: &reconciler{
 				getNamespaceFn: func(
-					context.Context,
-					types.NamespacedName,
-					client.Object,
-					...client.GetOption,
+					_ context.Context,
+					_ types.NamespacedName,
+					_ client.Object,
+					_ ...client.GetOption,
```
krancour: Nit: Usually `_` is only used in place of an arg you're not using if some of the other args _do_ have names / are being used. When none are used, it's typical to leave these out.

Main an fyi. Non-blocking.

pkg/controller/management/namespaces/namespaces_test.go
```diff
@@ -265,7 +391,7 @@ func TestReconcile(t *testing.T) {
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
-			res, err := testCase.reconciler.Reconcile(context.Background(), ctrl.Request{})
+			res, err := testCase.reconciler.Reconcile(t.Context(), ctrl.Request{})
```
krancour, line 402: Thanks. I also keep fixing this anywhere I find it. Glad you are too.

pkg/controller/management/namespaces/namespaces.go
```diff
@@ -119,6 +127,20 @@ func (r *reconciler) Reconcile(
...
+	// because they have a non-existent project owner reference.
+	// For more information see: https://github.com/akuity/kargo/issues/4627#issuecomment-3821967156
+	if slices.ContainsFunc(ns.OwnerReferences, isProjectOwnerRef) {
+		logger.Debug("Namespace has a Project owner reference, removing.")
+		ns.OwnerReferences = slices.DeleteFunc(ns.OwnerReferences, isProjectOwnerRef)
+		if err := r.patchOwnerReferencesFn(ctx, r.client, ns); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to patch namespace %q owner references: %w", ns.Name, err)
+		}
+		return ctrl.Result{}, nil
+	}
```
krancour: Can't this be done with _just_ `slices.DeleteFunc()`?

krancour: Or... TIL `controllerutil.RemoveOwnerReference()` also exists. We've not used it anywhere else, but this might be the most idiomatic way to do this currently in controller-runtime.

krancour: This is the only blocker. Great work!

PR #5765, opened 2026-02-19
chore: upgrade golangci-lint and address new lint warnings

krancour:


pkg/client/generated/models/create_repo_credentials_request.go
```diff
@@ -23,7 +23,8 @@ type CreateRepoCredentialsRequest struct {
 	// name
 	Name string `json:"name,omitempty"`
 
-	// password
+	// #nosec G117 -- Request data is unmarshaled into this struct, but the struct
```
thomastaylor312, line 26: This (and the other ones in generated files) will be clobbered next time these are generated. I'd have the linter ignore gosec in this generated directory

ui/src/gen/api/v2/models/createRepoCredentialsRequest.ts
thomastaylor312, line 1: Same note here around generated files

swagger.json
```diff
@@ -4712,6 +4712,7 @@
                     "type": "string"
                 },
                 "password": {
+                    "description": "#nosec G117 -- Request data is unmarshaled into this struct, but the struct\nis never marshaled and transmitted to anywhere.",
```
thomastaylor312, line 4715: Just found this. I don't love this solution TBH, but am ok with it if you want to keep it. Probably better to ignore the generated files

pkg/client/generated/models/create_repo_credentials_request.go
krancour: Ah... I didn't place this here. It landed here because the annotated type this was generated _from_ had this comment.

swagger.json
krancour: Related to https://github.com/akuity/kargo/pull/5765#discussion_r2830551399

I didn't put this here. swagger.json, itself, was generated. The source of this was the type this was generated _from_.

krancour: So I'll add that I'd also prefer if codegen didn't put this here, but I didn't find an easy way to avoid it.

thomastaylor312: Ah ok. I had it backwards. Guess it is what it is

Issue #5737, opened 2026-02-13
feat: add git-tag step; make git-push step support pushing tags


krancour:
Split from #3762

1. Add a `tag` option to the existing `git-push` step. It should be mutually exclusive with the `targetBranch` and `generateTargetBranch` options. When it's specified, push that tag.

1. Add a `git-tag` step that tags the commit at the head of the current branch with a user-specified tag name. Be sure to have it support a signing key, just like the `git-commit` step already does.

PR #5767, opened 2026-02-20
fix: pkg/controller/git tests

fuskovic:
I was working on https://github.com/akuity/kargo/issues/5737 and found that the tests in `pkg/controller/git` were actually failing on `main`.

```
 faris@Mac  ~/go/src/github.com/akuity/kargo/pkg/controller/git   main  go test . | grep FAIL:
--- FAIL: TestBareRepo (0.57s)
--- FAIL: TestBareRepo_SparseCheckout (0.63s)
    --- FAIL: TestBareRepo_SparseCheckout/sparse_checkout_includes_only_specified_directories (0.03s)
    --- FAIL: TestBareRepo_SparseCheckout/worktree_without_sparse_includes_all_directories (0.03s)
    --- FAIL: TestBareRepo_SparseCheckout/pattern_validation_rejects_absolute_paths (0.03s)
    --- FAIL: TestBareRepo_SparseCheckout/pattern_validation_rejects_path_traversal (0.03s)
    --- FAIL: TestBareRepo_SparseCheckout/glob_patterns_are_rejected (0.03s)
--- FAIL: TestBareRepo_WithFilter (0.50s)
--- FAIL: TestRepo (0.59s)
    --- FAIL: TestRepo/can_check_if_remote_branch_exists_--_positive_result (0.04s)
--- FAIL: TestWorkTree (0.52s)
```

```
 faris@Mac  ~/go/src/github.com/akuity/kargo/pkg/controller/git   fuskovic/fix-git-tests  go test .
ok      github.com/akuity/kargo/pkg/controller/git      4.294s
```

### DIFF
--- a/pkg/controller/git/bare_repo_test.go
+++ b/pkg/controller/git/bare_repo_test.go
@@ -109,10 +109,9 @@ func TestBareRepo(t *testing.T) {
 	})
 
 	workingTreePath := filepath.Join(rep.HomeDir(), "working-tree")
-	initBranch := defaultInitBranch(t)
 	workTree, err := rep.AddWorkTree(
 		workingTreePath,
-		&AddWorkTreeOptions{Ref: initBranch},
+		&AddWorkTreeOptions{Ref: defaultInitBranch(t)},
 	)
 
 	require.NoError(t, err)
@@ -532,8 +531,6 @@ func Test_validateSparsePatterns(t *testing.T) {
 }
 
 func TestBareRepo_WithFilter(t *testing.T) {
-	initBranch := defaultInitBranch(t)
-
 	testRepoCreds := RepoCredentials{
 		Username: "fake-username",
 		Password: "fake-password",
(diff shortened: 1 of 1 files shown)


fuskovic:
Noticing CI is failing. Looks like there are some environment differences between running locally vs CI. Need to take a closer look.

fuskovic:
I found it was most reliable to dynamically source the initial default branch. Which can vary based on git version, user config etc...

krancour:
This is a good change. Thanks @fuskovic.

I do just want to add to this that the make targets that are part of the suggested workflow are all prefixed with `hack-` and in the case of tests, linters, etc., those targets execute in a container maximally similar to CI. It exists to avoid these head-scratching moments of, "it worked on my machine but not in CI." 🤷‍♂️ 

Being containerized with a bind-mounted fs, `hack-test-unit` is a bit slower than `test-unit`, which runs on the host, so tbh, 99% of the time, I used the faster option. I just wanted everyone to be aware that an easy option for a sanity check exists if/when you encounter what may be environmental differences.

krancour:
Actually... this change appears to introduce the problem it purports to fix. Let's discuss offline.

Issue #5768, opened 2026-02-20
feat: make Kubernetes webhook server port configurable


audip:
## Problem

The Kubernetes webhook server port is hardcoded to 9443 in three places:

1. Go binary — `cmd/controlplane/kubernetes_webhooks.go` line 120:
   `WebhookServer: webhook.NewServer(webhook.Options{Port: 9443})`
2. Deployment template — `charts/kargo/templates/kubernetes-webhooks-server/deployment.yaml` line 85:
   `containerPort: 9443`
3. Service template — `charts/kargo/templates/kubernetes-webhooks-server/service.yaml` line 15:
   `targetPort: 9443`

This makes it impossible to change the webhook server port without forking the project. In environments where port 9443 conflicts with other services running on the same nodes (e.g. other admission webhook servers), operators have no way to resolve the conflict.

## Proposed Solution

Follow the existing pattern used by other configurable options (e.g. `METRICS_BIND_ADDRESS`, `KUBE_API_QPS`):

1. Helm `values.yaml`: Add `webhooksServer.port` defaulting to `9443` for backward compatibility.
2. ConfigMap template: Pass the port to the binary as `WEBHOOK_SERVER_PORT` env var.
3. Deployment template: Use `{{ .Values.webhooksServer.port }}` for `containerPort`.
4. Service template: Use `{{ .Values.webhooksServer.port }}` for `targetPort`.
5. Go code: Read `WEBHOOK_SERVER_PORT` env var (default `"9443"`) and pass it to `webhook.Options{Port: ...}.`

## Use Case

We run Kargo alongside other operators that also use port `9443` for their webhook servers. Being able to configure the port via Helm values (e.g. to `10288`) would allow us to avoid port conflicts without any workarounds.

audip:
@krancour can you review this issue and the PR #6358 that closes as we discussed? Sorry it took me a little while to submit a fresh PR

krancour:
Discussion in #5769 led to an agreement on how to approach.

PR #5769, opened 2026-02-20
feat: make Kubernetes webhook server port configurable

audip:
Fix for #5768 

## Summary

The webhook server port is hardcoded to `9443` in the Go binary, Deployment template, and Service template, making it impossible to change without forking. This PR makes it configurable via a new `webhooksServer.port` Helm value.

### Changes

- **`cmd/controlplane/kubernetes_webhooks.go`**: Added `WebhookServerPort` field to the options struct. Reads `WEBHOOK_SERVER_PORT` env var (default `"9443"`) in `complete()`, consistent with how `KUBE_API_QPS`, `KUBE_API_BURST`, and `METRICS_BIND_ADDRESS` are handled. Uses the field in `webhook.Options{Port: ...}` instead of the hardcoded value.
- **`charts/kargo/values.yaml`**: Added `webhooksServer.port: 9443`.
- **`charts/kargo/templates/kubernetes-webhooks-server/configmap.yaml`**: Added `WEBHOOK_SERVER_PORT` entry from the new Helm value.
- **`charts/kargo/templates/kubernetes-webhooks-server/deployment.yaml`**: Changed `containerPort: 9443` to `containerPort: {{ .Values.webhooksServer.port }}`.
- **`charts/kargo/templates/kubernetes-webhooks-server/service.yaml`**: Changed `targetPort: 9443` to `targetPort: {{ .Values.webhooksServer.port }}`.

### Motivation

When running Kargo alongside other operators that also bind their webhook servers to port 9443, there is no way to resolve the conflict. This change lets operators set a custom port (e.g. `10288`) via Helm values while remaining fully backward compatible — existing installations default to `9443` with no change in behavior.

## Test plan

- [ ] Verify `go build ./cmd/controlplane/` compiles cleanly (confirmed locally)
- [ ] Deploy with default values and confirm webhook server listens on 9443 (no behavioral change)
- [ ] Deploy with `webhooksServer.port: 10288` and confirm webhook server listens on 10288, Service targetPort is 10288, and webhook admission calls succeed through the Service
- [ ] `helm template` with default values produces identical output to current main (except for the new ConfigMap entry)
- [ ] `helm template` with `webhooksServer.port: 10288` produces correct containerPort and targetPort


krancour:
The motivation cited does not add up. It listens on 9443 in its own pod separate from any other webhook servers you may have running. There is no competition for the port.

audip:
> The motivation cited does not add up. It listens on 9443 in its own pod separate from any other webhook servers you may have running. There is no competition for the port.

Hi, thanks for the quick review.

You're absolutely right that under normal circumstances there’s no port contention since the webhook server runs in its own Pod network namespace.

However, in our environment (EKS with Cilium CNI), we run the webhook server with hostNetwork: true. In that configuration, the container shares the node’s network namespace, so ports must be unique across all host-networked Pods on the same node. Since 9443 is a commonly used webhook port (many operators default to it), we can encounter real port conflicts at the node level.

In that setup, the port is no longer isolated per Pod — it becomes a host-level binding — which is why configurability becomes necessary.

This PR:

- Does not change the default behavior (still 9443).
- Follows the existing pattern used for other configurable server settings (e.g., METRICS_BIND_ADDRESS).
- Only introduces flexibility for environments that require non-default networking modes.

Even if hostNetwork is not a common deployment mode, making the port configurable keeps the chart and binary more flexible with minimal complexity and no backward compatibility impact. These are some open-source projects that allow webhook server port to be configurable:
 - github-action-runners helm chart
 - trust-manager helm chart
 - and there are many others like this

Happy to clarify further or adjust the approach if you'd prefer a different configuration mechanism.

krancour:
Thanks for additional context @audip. That clarifies a lot.

I'm afraid I don't know much about Cilium, so excuse me if this question seems remedial. Is it common with Cilium to run webhook servers with host networking?

Is this done for many other workloads as well? Or is it something you're doing specifically for webhook servers?

Just trying to gather more information about how widespread an issue this may be and whether it might justify extending this same degree of port configurability to other Kargo components.

audip:
Great question @krancour . Cilium itself doesn't force `hostNetwork` on other workloads, but when running Cilium in kube-proxy replacement mode (which is a popular mode for EKS & Cilium), admission webhooks often need `hostNetwork: true` to be reachable during node bootstrap — before Cilium has fully initialized pod networking. It's a chicken-and-egg problem: the API server needs to call webhooks for pod admission, but pod networking isn't ready yet.

This mostly affects admission webhook servers specifically, since they're the components the API server must reach synchronously. Regular controllers and app workloads use normal pod networking and don't have this issue.

The port conflict problem is real and common: port 9443 is a popular webhook default shared by cert-manager, OPA Gatekeeper, Kyverno, and others. When multiple webhook servers run with `hostNetwork` on the same node, they collide. This is well-documented across the ecosystem — for example, [cert-manager](https://cert-manager.io/docs/installation/compatibility/#primitives) and [Kyverno](https://kyverno.io/docs/installation/customization/#container-flags) both make their webhook port configurable for exactly this reason.

krancour:
I'm more or less of the opinion that this is the sort of last-mile configuraion folks should consider doing with another config management tool, such as Kustomize.

audip:
> I'm more or less of the opinion that this is the sort of last-mile configuraion folks should consider doing with another config management tool, such as Kustomize.

I agree in general, but in this case the binary itself doesn't expose the port as configurable, so there isn't anything for Kustomize to patch cleanly. The alternative today is effectively forking the repo and maintaining a downstream copy just to change a port, which feels like a pretty poor outcome.

To me, the ideal model is exposing a few configuration levers with sane defaults, and then letting users layer Kustomize or other tooling on top as needed. That's why I'm disappointed this hasn't gained more traction so far.

Hopefully, you can help me support this tiny PR and make the webhook port configurable (this screenshot shows the piece that is actually lacking)

image

krancour:
@audip fair point. Although I won't want to complicate the chart with options that will be used by an extreme minority of users, I'd be willing to take the small change to the Go code only. This would allow you to modify this last-mile as discussed.

audip:
> @audip fair point. Although I won't want to complicate the chart with options that will be used by an extreme minority of users, I'd be willing to take the small change to the Go code only. This would allow you to modify this last-mile as discussed.

Sure that works. I'll update the PR!

PR #5770, opened 2026-02-21
fix: add finalizer to running Promotions to prevent orphaned work dirs

### LINKED ISSUE #5728: bug: temporary promo dirs on controller fs are orphaned if a running promo is deleted
Promotions do not have a finalizer to guarantee cleanup because Promotions stick around after reaching a terminal state. Relying on a finalizer to effect cleanup would result in cleanup not occurring until a Promotion were garbage collected or deleted manually. So cleanup is handled differently... it occurs at the very end of a reconciliation cycle if it ends with the Promo having reached a terminal state. The trouble with this is that cleanup will never occur if a Promotion is deleted while in a Running state.

We should begin adding finalizers to Promotions as they enter the Running state and clearing them after cleanup has been triggered by the transition to a terminal state. If a Promotion is deleted while _still_ in Running state, the still-present finalizer will block final deletion until a reconcilation cycle executes cleanup and clears it.

The end goal is that the transition to a terminal state _and_ deletion of a Running Promotion _both_ will trigger the same cleanup process.

Assigning this to myself as a placeholder to prevent anyone else from picking this up. My true intention is to assign this to a new colleague who is onboarding next week. cc @thomastaylor312

EronWright:
## Summary

Fixes #5728

When a Promotion is deleted while in `Running` state, its temporary working directory (`/tmp/promotion-<UID>`) is never cleaned up. Cleanup currently only happens via a `defer` in `promote()` that checks `IsTerminal()`. If the Promotion object is deleted instead of reaching a terminal phase, the reconciler returns early and cleanup never occurs.

This PR adds a Kubernetes finalizer to Promotions when they enter `Running` state. The finalizer ensures that both normal completion and deletion trigger the same cleanup path.

### Finalizer lifecycle

1. **Added before Running** — The finalizer is added *before* patching the phase to `Running`, preventing a race where a delete could slip in between the status patch and the finalizer addition.
2. **Removed on terminal** — After a Promotion reaches a terminal phase (`Succeeded`, `Failed`, `Errored`, `Aborted`), the work directory is cleaned up and the finalizer is removed.
3. **Removed on deletion** — If a Running Promotion is deleted, `deletePromotion()` sets `Aborted` status, emits a `PromotionAborted` event, cleans up the work directory, and removes the finalizer.
4. **Removed on user-driven abort** — `terminatePromotion()` now also cleans up the work directory and removes the finalizer.

### Deletion handling

The `IgnoreDelete` event filter is removed so the reconciler receives delete events for finalized Promotions. Promotions without a finalizer (e.g., deleted while `Pending`) still vanish immediately — the reconciler finds `promo == nil` and returns.

When a finalized Promotion is deleted, `deletePromotion()` sets `Aborted` status before removing the finalizer. This matters because:
- Other finalizers on the object may keep it alive, so observers (UI, other controllers, audit tools) benefit from seeing a proper terminal state.
- The `PromotionAborted` event provides the same audit trail as user-driven aborts.

### Crash recovery

If the controller crashes between patching terminal status and removing the finalizer, the work directory and finalizer are stranded. On restart, the informer LIST generates synthetic Create events that pass all predicates. The reconciler re-enters, sees the Promotion is terminal, and cleans up the stale finalizer. This is a no-op when the finalizer is already absent.

### Why this doesn't cause reconciliation loops

Patching `metadata.finalizers` does not increment `metadata.generation` (the Promotion CRD uses the status subresource, so only `.spec` changes increment generation). The existing `predicate.Or(GenerationChanged, RefreshRequested, AbortRequested)` filter drops the resulting Update event. Delete events pass through because all three predicates inherit the default `Delete() → true` from `predicate.Funcs`.

## Test plan

- [x] Deleted Running Promotion with finalizer — sets Aborted status, emits event, cleans up workdir, removes finalizer
- [x] Deleted Promotion without finalizer — no-op
- [x] Terminal Promotion with stale finalizer — cleaned up on re-reconcile
- [x] Terminal Promotion without finalizer — no-op
- [x] Pending → Running transition adds finalizer before status patch
- [x] `terminatePromotion` cleans up workdir and removes finalizer
- [x] Existing `TestReconcile` and `Test_reconciler_terminatePromotion` updated and passing
- [x] `go test -v -race ./pkg/controller/promotions/...`
- [x] `make lint-go`

### DIFF
--- a/pkg/controller/promotions/promotions.go
+++ b/pkg/controller/promotions/promotions.go
@@ -31,6 +31,7 @@ import (
 	"github.com/akuity/kargo/pkg/kubeclient"
 	libEvent "github.com/akuity/kargo/pkg/kubernetes/event"
 	"github.com/akuity/kargo/pkg/logging"
+	intpredicate "github.com/akuity/kargo/pkg/predicate"
 	"github.com/akuity/kargo/pkg/promotion"
 )
 
@@ -82,14 +83,14 @@ type reconciler struct {
 		*kargoapi.Freight,
 	) (*kargoapi.PromotionStatus, *time.Duration, error)
 
-	terminatePromotionFn func(
+	handleAbortFn func(
 		context.Context,
 		*kargoapi.AbortPromotionRequest,
 		*kargoapi.Promotion,
 		*kargoapi.Freight,
 	) error
 
-	deletePromotionFn func(
+	handleDeleteFn func(
 		context.Context,
 		*kargoapi.Promotion,
 	) error
@@ -136,6 +137,7 @@ func SetupReconcilerWithManager(
 			IsDefaultController: cfg.IsDefaultController,
 			ShardName:           cfg.ShardName,
 		}).
+		WithEventFilter(intpredicate.IgnoreDelete[client.Object]{}).
 		WithEventFilter(predicate.Or(
 			predicate.GenerationChangedPredicate{},
 			kargo.RefreshRequested{},
@@ -210,8 +212,8 @@ func newReconciler(
 	}
 	r.getStageFn = api.GetStage
 	r.promoteFn = r.promote
-	r.terminatePromotionFn = r.terminatePromotion
-	r.deletePromotionFn = r.deletePromotion
+	r.handleAbortFn = r.handleAbort
+	r.handleDeleteFn = r.handleDelete
 	r.cleanupWorkDirFn = r.cleanupWorkDir
 	return r
 }
@@ -247,12 +249,13 @@ func (r *reconciler) Reconcile(
 
 	if promo.Status.Phase.IsTerminal() {
 		// Clean up any finalizer left on a terminal Promotion (whether deleted or not).
+		// This is a no-op if the finalizer has already been removed.
 		return ctrl.Result{}, r.handleCleanup(ctx, promo)
 	}
 
 	// Handle premature deletion of a non-terminal Promotion.
 	if !promo.DeletionTimestamp.IsZero() {
-		return ctrl.Result{}, r.deletePromotionFn(ctx, promo)
+		return ctrl.Result{}, r.handleDeleteFn(ctx, promo)
 	}
 
 	// Find the Freight
@@ -280,7 +283,7 @@ func (r *reconciler) Reconcile(
 	if req, ok := api.AbortPromotionAnnotationValue(
 		promo.GetAnnotations(),
 	); ok && req.Action == kargoapi.AbortActionTerminate {
-		if err = r.terminatePromotionFn(ctx, req, promo, freight); err != nil {
+		if err = r.handleAbortFn(ctx, req, promo, freight); err != nil {
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, nil
@@ -328,8 +331,6 @@ func (r *reconciler) Reconcile(
 		return ctrl.Result{}, nil
 	}
 
-	// Update promo status as Running to give visibility in UI. Also, a promo which
-	// has already entered Running status will be allowed to continue to reconcile.
 	// Add finalizer to ensure cleanup of the working directory.
 	// This is outside the phase check so that Promotions that were already
 	// Running before the finalizer was introduced also get one.
@@ -337,6 +338,8 @@ func (r *reconciler) Reconcile(
 		return ctrl.Result{}, fmt.Errorf("error adding finalizer to Promotion: %w", err)
 	}
 
+	// Update promo status as Running to give visibility in UI. Also, a promo which
+	// has already entered Running status will be allowed to continue to reconcile.
 	if promo.Status.Phase != kargoapi.PromotionPhaseRunning {
 		if err = kubeclient.PatchStatus(ctx, r.kargoClient, promo, func(status *kargoapi.PromotionStatus) {
 			status.Phase = kargoapi.PromotionPhaseRunning
@@ -563,7 +566,7 @@ func (r *reconciler) promote(
 		stage,
 		promotion.WithActor(api.CreateActorAnnotationValue(&promo)),
 		promotion.WithUIBaseURL(r.cfg.APIServerBaseURL),
-		promotion.WithWorkDir(filepath.Join(os.TempDir(), "promotion-"+string(workingPromo.UID))),
+		promotion.WithWorkDir(promotionWorkDir(workingPromo.UID)),
 	)
 	if err := os.Mkdir(promoCtx.WorkDir, 0o700); err == nil {
 		// If we're working with a fresh directory, we should start the promotion
@@ -660,10 +663,10 @@ func (r *reconciler) buildTargetFreightCollection(
 	return freightCol
 }
 
-// terminatePromotion terminates the given Promotion with a message indicating
+// handleAbort terminates the given Promotion with a message indicating
 // that it was terminated on user request. It does nothing if the Promotion is
 // already in a terminal phase.
-func (r *reconciler) terminatePromotion(
+func (r *reconciler) handleAbort(
 	ctx context.Context,
 	req *kargoapi.AbortPromotionRequest,
 	promo *kargoapi.Promotion,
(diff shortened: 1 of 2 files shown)


pkg/controller/promotions/promotions.go
```diff
@@ -88,6 +88,13 @@ type reconciler struct {
 		*kargoapi.Promotion,
 		*kargoapi.Freight,
 	) error
+
+	deletePromotionFn func(
+		context.Context,
+		*kargoapi.Promotion,
+	) error
+
+	cleanupWorkDirFn func(ctx context.Context, promoUID types.UID)
```
krancour: @EronWright I can't recall if I'd mentioned this. Purely an fyi, we're avoiding this pattern these days, but anywhere it's already in use, it's fine to continue, as you have here. It's the more rational choice than refactoring the tests completely.

pkg/controller/promotions/promotions.go
```diff
@@ -129,7 +136,6 @@ func SetupReconcilerWithManager(
 			IsDefaultController: cfg.IsDefaultController,
 			ShardName:           cfg.ShardName,
 		}).
-		WithEventFilter(intpredicate.IgnoreDelete[client.Object]{}).
```
krancour, line 132: I'm about 99% sure we don't need to remove this. This blocks _actual_ delete events, at which point there's nothing left for us to do. The logical delete events we care about are actually update events affecting `DeletionTimestamp`.

pkg/controller/promotions/promotions.go
```diff
@@ -312,6 +330,13 @@ func (r *reconciler) Reconcile(
 
 	// Update promo status as Running to give visibility in UI. Also, a promo which
 	// has already entered Running status will be allowed to continue to reconcile.
```
krancour: Suggest moving the comment to above +340 so it remains close to the code it was documenting.

pkg/controller/promotions/promotions.go
EronWright: I double-checked and agree that this filters the event that would happen post-finalization.

pkg/controller/promotions/promotions.go
```diff
@@ -237,6 +245,16 @@ func (r *reconciler) Reconcile(
 		return ctrl.Result{}, nil
 	}
 
+	if promo.Status.Phase.IsTerminal() {
+		// Clean up any finalizer left on a terminal Promotion (whether deleted or not).
```
krancour, line 251: Nit: Just to avert possible future confusion, noting here that this is a no-op if the finalizer is already absent might be helpful. You already did note this on the method itself, but I saw this first and it initially made me wonder how much work we were potentially doing unnecessarily, until I dug deeper.

pkg/controller/promotions/promotions.go
```diff
@@ -696,6 +725,85 @@ func (r *reconciler) terminatePromotion(
 	return nil
 }
 
+// deletePromotion aborts the given Promotion with a message indicating
+// that the object was deleted. It assumes that the Promotion is NOT in a terminal phase.
+func (r *reconciler) deletePromotion(ctx context.Context, promo *kargoapi.Promotion) error {
```
krancour: Could we name this something like `handleDelete()`? As is, it sounds like it actually _deletes_ a Promotion, but that's actually something that's already been set in motion externally.

pkg/controller/promotions/promotions.go
```diff
@@ -696,6 +725,85 @@ func (r *reconciler) terminatePromotion(
...
+		return nil
+	}
+	r.cleanupWorkDirFn(ctx, promo.UID)
+	return r.removeFinalizer(ctx, promo)
+}
+
+// cleanupWorkDir removes the temporary working directory for a Promotion.
+// This is safe to call even if the directory does not exist.
+func (r *reconciler) cleanupWorkDir(ctx context.Context, promoUID types.UID) {
+	workDir := filepath.Join(os.TempDir(), "promotion-"+string(promoUID))
```
krancour: This same logic for determining the Promotion's working directory occurs one other place in this package (in this file, in fact). We should probably make that a small helper function so that if we ever decide to change how this works, we don't risk forgetting to update one of the two spots.

pkg/controller/promotions/promotions.go
EronWright: 100% agree, will do.  My rationale was that `terminatePromotion` handles the user request that stems from an annotation case, and `deletePromotion` handles the DELETE variant.  I suppose I could rename `terminatePromotion` to `handleTermination` alongside `handleDeletion`.

EronWright:
## Manual Test Results

Tested the bugfix end-to-end using the **kargo-demo** quickstart (3-stage pipeline: test → uat → prod) on a local OrbStack cluster with Tilt.

### Test A: Normal Promotion (baseline) ✅

Promoted Freight to `test` Stage. Finalizer was added when the Promotion transitioned to Running, and removed after it Succeeded.

Controller log confirms cleanup:
```
promotions/promotions.go:396  promotion  phase=Succeeded
promotions/promotions.go:803  removed finalizer from source Promotion
```

After completion, no finalizer remains:
```yaml
metadata:
  finalizers: []   # empty
status:
  phase: Succeeded
```

### Test B: Delete a Running Promotion ✅

Promoted Freight to `prod`, then deleted the Promotion ~1 second later while it was still Running (on the `argocd-update` step).

The watch output shows the full lifecycle:

**1. Finalizer added, promotion running:**
```yaml
metadata:
  finalizers:
  - kargo.akuity.io/finalizer
  deletionTimestamp: "2026-02-23T19:43:12Z"   # deletion requested
status:
  phase: Running
  stepExecutionMetadata:
  - alias: promo-process::step-7
    startedAt: "2026-02-23T19:43:19Z"
    status: Running
```

**2. Controller aborts the deleted promotion:**
```
promotions/promotions.go:737  aborting deleted Promotion
promotions/promotions.go:803  removed finalizer from source Promotion
```

**3. Final state before deletion:**
```yaml
status:
  message: Promotion deleted
  phase: Aborted
  stepExecutionMetadata:
  # steps 1-6: Succeeded
  - alias: promo-process::step-7
    finishedAt: "2026-02-23T19:43:19Z"
    startedAt: "2026-02-23T19:43:19Z"
    status: Aborted    # running step marked Aborted
```

The object was then fully deleted (finalizer removed → Kubernetes completed the deletion).

**4. UI observes the final status:**

image

**5. Event fired:**

```yaml
apiVersion: v1
count: 1
eventTime: null
firstTimestamp: "2026-02-23T19:43:19Z"
involvedObject:
  kind: Promotion
  name: prod-delete-test-8pps6
  namespace: kargo-demo
kind: Event
lastTimestamp: "2026-02-23T19:43:19Z"
message: Promotion deleted
metadata:
  annotations:
    event.kargo.akuity.io/actor: kubernetes:system:admin
    event.kargo.akuity.io/applications: '[{"Namespace":"argocd","Name":"kargo-demo-prod"}]'
    event.kargo.akuity.io/freight-alias: loitering-panther
    event.kargo.akuity.io/freight-create-time: "2026-02-23T18:56:35Z"
    event.kargo.akuity.io/freight-images: '[{"repoURL":"public.ecr.aws/nginx/nginx","tag":"1.29.5","digest":"sha256:89b944864b72d5d0932f40bfee9c4de19975f9593d8767c8374c645d7b2730c0"}]'
    event.kargo.akuity.io/freight-name: c25813b46b65588cb7e8efe9e2f10cc6fba6d42d
    event.kargo.akuity.io/project: kargo-demo
    event.kargo.akuity.io/promotion-create-time: "2026-02-23T19:43:11Z"
    event.kargo.akuity.io/promotion-name: prod-delete-test-8pps6
    event.kargo.akuity.io/stage-name: prod
  creationTimestamp: "2026-02-23T19:43:19Z"
  name: prod-delete-test-8pps6.1896f7a6e83a070e
  namespace: kargo-demo
  resourceVersion: "277251"
  uid: 720e698b-70c7-45ab-ab0f-bbcf4a153ac0
reason: PromotionAborted
reportingComponent: promotion-controller
reportingInstance: ""
source:
  component: promotion-controller
type: Normal
```

### Test C: Delete a Completed Promotion ✅

Deleted a Succeeded promotion — it was removed instantly with no phase change and no finalizer to process.

No orphaned work directories in the controller pod:
```
$ kubectl exec <controller-pod> -- ls /tmp/ | grep promotion
(no output — clean)
```

### Additional Finding: Missing RBAC

The controller's ClusterRole was missing permissions needed for finalizer management. Added in commit `123bdfd5`:
- `patch` verb on `promotions` resource (needed to add/remove finalizers via merge patch)
- `promotions/finalizers` subresource with `patch` verb

krancour (dismissed):
🔥

EronWright:
Closing in favor of a different approach that doesn't rely on having a finalizer:
https://github.com/akuity/kargo/pull/5805

PR #5782, opened 2026-02-24
feat: add oci-push promotion step

### LINKED ISSUE #3762: Add a `push-image` step
# Proposed Feature

Add a ~~`push-tag` or~~ `push-image` ~~step for images and a `push-tag` step for repositories~~

__Edit(@krancour):__ Further discussion in this thread will be limited to images. Pushing git tags has been broken out to a separate issue. See #5737.

# Motivation

This would to release tags only for versions that have been properly tested (promoted in a testing environment),  the following pattern :

You can also look [at the following reddit discussion 
](https://www.reddit.com/r/ArgoCD/comments/1j9gnsf/comment/mhe2urq/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)
Two warehouses: 

- A, which subscribe using the NewestBuild strategy
- B, which subscribe using the SemVer strategy

1. Artifact (git or image) is pulled from warehouse A using the NewestBuild strategy (ex: reference fe80c...)
2. Artifact is promoted to a testing environment
3. Artifact is tested with an AnalysisRun
4. If the AnalysisRun is valid, push a tag to git repo/image repo (ex: reference 1.0.2)
5. That artifact (1.0.2) is retrieved using the SemVer stategy
6. Artifact is promoted to as staging environment
7. ...
8. Artifact is promoted to a production environment

# Suggested Implementation

A couple important points :

- Make sure that there is a way to specify and/or automatically determine and set what will be the (name) of the tag that will be pushed to git/image registry.
- Make sure that the `push-tag` can happen *after* an AnalysisRun.
- For image registries, we could create either a `push-image` or `push-tag` but the most important bit is the ability to tag the image that has been tested.

EronWright:
## Summary

Adds a new `oci-push` promotion step that copies/retags OCI artifacts (container images and Helm charts) between registries.

- Extracts shared OCI helpers (`parseOCIReference`, `buildOCIRemoteOptions`, etc.) from `oci-download` into `oci_common.go` so both steps reuse the same credential resolution and transport logic
- New step runner registered as `oci-push` with `StepCapabilityAccessCredentials`
- Supports single images, multi-arch image indexes, and Helm charts (via `oci://` prefix)
- Config fields: `srcRef` (source artifact), `destRef` (destination), optional `annotations` and `insecureSkipTLSVerify`
- Optional `annotations` field to set OCI manifest annotations on the pushed artifact, with scoped prefixes (`index:`, `manifest:`) for controlling placement on image indexes vs child manifests
- Custom annotation mutation wrappers (extracted into `mutate.go`) to work around a limitation of `go-containerregistry`
- 1 GiB compressed size limit enforced for cross-repository copies; same-repo retags skip the check since blobs are already present
- Step outputs: `image` (destination ref), `digest` (sha256), `tag`

Closes #3762

## Test plan

- [x] Unit tests pass (`go test -race ./pkg/promotion/runner/builtin/...`)
- [x] Linter passes (`make lint-go` — no issues in changed files)
- [x] E2E: Helm chart cross-repo copy on GHCR with custom annotations
- [x] E2E: nginx image (~70 MB) cross-repo copy — succeeded with size check
- [x] E2E: pulumi/pulumi (~6.1 GiB) cross-repo copy — rejected with `compressed artifact size 6.1 GiB exceeds maximum allowed size of 1.0 GiB`
- [x] E2E: pulumi/pulumi (~6.1 GiB) same-repo retag + annotate — succeeded (size check bypassed)

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
@@ -84,6 +84,13 @@ steps:
   config:
     srcRef: registry.example.com/widget-service@${{ imageFrom("registry.example.com/widget-service").digest }}
     destRef: registry.example.com/widget-service/${{ ctx.stage }}:${{ imageFrom("registry.example.com/widget-service").tag }}
+- uses: kustomize-set-image
+  config:
+    path: ./out
+    images:
+    - image: registry.example.com/widget-service
+      newName: registry.example.com/widget-service/${{ ctx.stage }}
+      digest: ${{ outputs["push-to-stage"].digest }}
 ```
 
 ### Promoting an OCI Helm Chart


EronWright:
## E2E Test Results

Tested the `oci-push` step end-to-end on a local Tilt dev cluster (OrbStack) with a Helm chart hosted on GHCR.

### Setup
- Pushed two chart versions (`0.1.0`, `0.2.0`) to `oci://ghcr.io/eronwright/charts/mychart`
- Created `oci-push-demo` project with:
  - Helm credential secret for GHCR
  - Warehouse subscribing to `oci://ghcr.io/eronwright/charts/mychart` (semver `>=0.1.0`)
  - Stage with an `oci-push` step to retag the chart as `:promoted` with custom annotations

### Promotion

```yaml
steps:
- uses: oci-push
  as: retag-chart
  config:
    imageRef: "oci://ghcr.io/eronwright/charts/mychart:${{ chartFrom('oci://ghcr.io/eronwright/charts/mychart').Version }}"
    destRef: "oci://ghcr.io/eronwright/charts/mychart:promoted"
    annotations:
      io.kargo.promoted-by: "kargo"
      io.kargo.stage: "${{ ctx.stage }}"
```

### Result: ✅ Succeeded

**Step outputs:**
```
image:  ghcr.io/eronwright/charts/mychart:promoted
digest: sha256:f2cca6aef681692050ad7262b90cc3e2b9e3fea7d09013217bca498480fddd54
tag:    promoted
```

**Manifest annotations verified on GHCR:**
```json
{
  "io.kargo.promoted-by": "kargo",
  "io.kargo.stage": "retag",
  "org.opencontainers.image.created": "2026-02-23T16:06:15-08:00",
  "org.opencontainers.image.description": "A minimal chart for testing oci-push",
  "org.opencontainers.image.title": "mychart",
  "org.opencontainers.image.version": "0.2.0"
}
```

EronWright:
## E2E Test Report: Cross-repo Helm chart push with annotations on GHCR

Tested cross-repository OCI push of a Helm chart with annotations to a brand-new GHCR repository (no pre-existing blobs).

**Test setup** (namespace `oci-push-demo`, local Tilt cluster):
- Source: `oci://ghcr.io/eronwright/charts/mychart:0.2.0`
- Destination: `oci://ghcr.io/eronwright/charts/mychart-prod3:0.2.0` (fresh repo, never pushed to)
- Annotations: `io.kargo.promoted-by`, `io.kargo.source-repo`, `io.kargo.test`

**Result**: `Succeeded`

```
Promotion: cross-repo-wrapper-4jrbj
Digest:    sha256:e2264815f65afd471aee8c1df274932d72cccb6ab43013df854d91e8f236cee5
```

**Manifest on GHCR** (`docker manifest inspect`):
```json
{
  "config": { "mediaType": "application/vnd.cncf.helm.config.v1+json" },
  "layers": [{ "mediaType": "application/vnd.cncf.helm.chart.content.v1.tar+gzip" }],
  "annotations": {
    "io.kargo.promoted-by": "kargo",
    "io.kargo.source-repo": "ghcr.io/eronwright/charts/mychart",
    "io.kargo.test": "custom-wrapper",
    "org.opencontainers.image.created": "2026-02-23T16:06:15-08:00",
    "org.opencontainers.image.description": "A minimal chart for testing oci-push",
    "org.opencontainers.image.title": "mychart",
    "org.opencontainers.image.version": "0.2.0"
  }
}
```

- Config and layer blobs transferred correctly (no `MANIFEST_BLOB_UNKNOWN`)
- All three custom annotations applied
- Original source annotations preserved
- Single write (no two-phase workaround)

Previously tested in earlier iterations: retag-in-place (same repo), cross-repo without annotations, image index push. All passed.

EronWright:
## E2E Test: Container Image (multi-arch nginx)

Tested `oci-push` with a multi-arch container image (`public.ecr.aws/nginx/nginx:1.29.5`, Docker manifest list with amd64+arm64) pushed cross-registry to GHCR.

### Pure copy (no annotations)

Digest is **preserved** across registries:

| | Digest |
|---|---|
| Source (`public.ecr.aws`) | `sha256:92038aacb18ce74dd5c454f03351dbd9c6ba27c75fd6c1a7658c899ae11ed496` |
| Destination (`ghcr.io`) | `sha256:92038aacb18ce74dd5c454f03351dbd9c6ba27c75fd6c1a7658c899ae11ed496` |

The pushed image was verified with `docker pull` + `docker run`.

### With scoped annotations

Annotations used:
```yaml
annotations:
  io.kargo.promoted-by: kargo
  io.kargo.source-repo: public.ecr.aws/nginx/nginx
  "index:io.kargo.index-test": index-annotation-value
  "manifest:io.kargo.manifest-test": manifest-annotation-value
```

All digests changed as expected (annotations mutate manifest content):

| Level | Pure copy | With annotations |
|---|---|---|
| **Index** | `sha256:9203...` | `sha256:7126...` |
| **amd64** | `sha256:b137...` | `sha256:3c6d...` |
| **arm64** | `sha256:bde5...` | `sha256:6e1e...` |

Annotation scoping worked correctly:
- **Index manifest**: only `io.kargo.index-test` (prefix stripped)
- **Child manifests**: `io.kargo.promoted-by`, `io.kargo.source-repo` (unprefixed → manifest scope), `io.kargo.manifest-test` (prefix stripped)

EronWright:
Added a 1 GiB size limit for cross-repository copies in `oci-push`. The limit is enforced before any blobs are transferred, using only manifest metadata to compute the compressed artifact size (config + all layers, summed across child images for indexes).

Same-repository retags skip the check since no blob transfer occurs.

Tested against the live cluster:
- **Happy path**: `public.ecr.aws/nginx/nginx:1.29.5` (~70 MB) cross-repo push succeeded
- **Size limit**: `pulumi/pulumi:latest` (~6.1 GiB) was rejected with: `compressed artifact size 6.1 GiB exceeds maximum allowed size of 1.0 GiB`

EronWright:
Verified that same-repo retag + annotate correctly bypasses the size limit:

1. Copied `pulumi/pulumi:latest` (~6.1 GiB) to `ghcr.io/eronwright/test-images/pulumi:latest` using `oras`
2. Created a promotion to retag and annotate within the same repo → **succeeded** (size check skipped since blobs are already present)
3. Verified annotations landed on the child manifests

So the three cases work as expected:
- **Cross-repo copy of large image** → rejected with `compressed artifact size 6.1 GiB exceeds maximum allowed size of 1.0 GiB`
- **Same-repo retag + annotate of large image** → succeeded, no size check
- **Cross-repo copy of normal image** → succeeded, size check passed

EronWright:
Closing this PR for now, in favor of a more limited solution focusing on tagging, not replication.

To recap, this PR offers an `oci-push` step that does remote-to-remote copying and tagging, with a 1GB size cap applied when crossing repos, and with full annotation support.

krancour:
Re-opening after further offline discussions.

pkg/promotion/runner/builtin/oci_common.go
EronWright, line 1: Note that these functions were extracted from oci_downloader.go to oci_common.go for sharing purposes with oci_pusher.go.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
```diff
@@ -0,0 +1,149 @@
...
+registry. This step supports container images and Helm charts stored in OCI
+registries, making it useful for promoting verified artifacts through a pipeline
+— for example, retagging an image with a release version after verification, or
+copying it to a production registry. Multi-arch image indexes are copied in full.
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `imageRef` | `string` | Y | Reference to the source OCI artifact. Supports both tag format `registry/repository:tag` and digest format `registry/repository@sha256:digest`. For Helm OCI artifacts, the `oci://` prefix is supported (e.g., `oci://registry/repository:tag`) and will use Helm-specific credential lookup. |
```
krancour: I'm iffy on `imageRef`. If we're going to the trouble to generalize this as `oci-push` and already plan to use it for charts, and will likely use it for other sorts of artifacts in the future, I think I'd try to be more general here also. `srcRef`? `artifactRef`?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
```diff
@@ -0,0 +1,149 @@
...
+exceed 1 GiB (or as configured by your administrator). For multi-arch image indexes, this includes the sum across all
+child images. Exceeding this limit causes a terminal (non-retryable) error.
+
+This limit is not enforced when `imageRef` and `destRef` refer to the same
+repository (i.e. retagging within the same registry and path), since no blob
+transfer occurs in that case.
+
+## Examples
+
+### Retagging an Image After Verification
```
krancour: General comment that might have minor implications for a few of the examples, which generally appear to be pretty good: Verification is too often mistaken by users as being integral to the promotion process. In fact, it's not. It is validation of specific Freight in a specific Stage _at any time_ and it's independent of any Promotion. It can be repeated on-demand.

(No doubt the confusion arises from the fact that after a Promotion completes, the Freight is not yet verified in that Stage, so verification does kick off almost immediately. But it's neither part of the Promotion itself nor triggered directly by the its completion. The trigger is the Stage having Freight that it hasn't yet validated.)

"Retagging an Image After Verification" seems to me a _valid_ and useful example, but for a user prone to this misunderstanding or one not reading the example critically, this could sound a little bit like "a push step following a verification step," which isn't at all what's going on.

I think the thing that would drive home what realistic usage looks like would be to say something about this one-step promotion process belonging to a dedicated Stage downstream from test and upstream from staging. When Freight is promoted from test to here, the single step just retags and pushes the image. (Or something like that.)

pkg/fmt/bytes.go
```diff
@@ -0,0 +1,17 @@
+package fmt
+
+import "fmt"
+
+// FormatBytes formats a byte count using the largest appropriate binary unit.
+func FormatBytes(b int64) string {
```
krancour: This probably deserves a unit test. I might consider renaming it also. At a glance, it's odd that `FormatBytes()` doesn't take bytes as input. A name like `FormatByteCount()` would be in better agreement with the function signature and the docstring.

pkg/promotion/runner/builtin/oci_pusher.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+// newOCIPusher returns an implementation of the promotion.StepRunner interface
+// that pushes OCI artifacts to a registry. It uses the provided credentials
+// database to authenticate with source and destination registries.
+func newOCIPusher(caps promotion.StepRunnerCapabilities) promotion.StepRunner {
+	return &ociPusher{
+		credsDB:         caps.CredsDB,
+		schemaLoader:    getConfigSchemaLoader(stepKindOCIPush),
+		maxArtifactSize: maxArtifactSizeFromEnv(),
```
krancour: I would probably do this more eagerly so that if it's misconfigured, the panic is at startup and not at the time of a Promotion.

pkg/promotion/runner/builtin/oci_pusher.go
```diff
@@ -0,0 +1,319 @@
...
+
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/xeipuuv/gojsonschema"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/pkg/credentials"
+	libfmt "github.com/akuity/kargo/pkg/fmt"
+	libmutate "github.com/akuity/kargo/pkg/image/mutate"
```
krancour: I wouldn't alias this unless necessary due to also using a `mutate` package from a third party or the stdlib.

pkg/promotion/runner/builtin/oci_pusher.go
```diff
@@ -0,0 +1,319 @@
...
+// with "manifest:" or unprefixed target image manifests.
+type annotationScopes struct {
+	index    map[string]string // applied to the image index manifest
+	manifest map[string]string // applied to each image manifest
+}
+
+// parseAnnotationScopes splits annotation keys by their scope prefix.
+// Keys prefixed with "index:" are routed to the index manifest, keys prefixed
+// with "manifest:" or unprefixed are routed to image manifests.
+func parseAnnotationScopes(annotations map[string]string) annotationScopes {
```
krancour: For better or worse, we've got a lot of unrelated steps in this one package. Maybe we'll break that up someday, but in the meantime, I've found that making functions such as this one a method of the step runner, it proactively heads off naming collisions with what could be totally different but similarly named functions supporting some other step runner.

pkg/image/mutate/mutate.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	return partial.Digest(i)
+}
+
+func (i *image) Size() (int64, error) {
+	if err := i.compute(); err != nil {
+		return -1, err
+	}
+	return partial.Size(i)
+}
```
krancour, line 135: Can you explain what's going on here? By all appearances, `compute()` performs some one-time initialization of the `image` and many methods of `image` call it up front as if to guarantee initialization. This seems a bit unconventional. Is there a reason not to perform all initialization with a constructor-like function like `newImage()`?

krancour: Similar question for `index`. I see `newIndex()` actually exists, but doesn't handle the initialization performed by its own `compute()` method, so it's still piquing my curiosity.

EronWright: The rationale is that it mimics the pattern seen in upstream `go-containerregistry`. This file contains an alternative implementation of the `Image` and `Index` interfaces, to overcome a bug in the upstream implementation that is triggered by non-container images (i.e. Helm charts).

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
EronWright: This is patterned after `oci-download`:

> `imageRef` - Reference to the OCI artifact to download. Supports both tag format `registry/repository:tag` and digest format `registry/repository@sha256:digest`. For Helm OCI artifacts, the `oci://` prefix is supported (e.g., `oci://registry/repository:tag`) and will use Helm-specific credential lookup.

I think `srcRef` would be a good alternative but would rather leave it.

pkg/promotion/runner/builtin/oci_pusher.go
EronWright: I wanted to draw a visual distinction for our specialized mutator, at the usage site.

pkg/promotion/runner/builtin/oci_pusher.go
EronWright: Make sense.

pkg/image/mutate/mutate.go
EronWright: Looking a bit deeper, the rationale for lazy computation is that computing the manifest requires calling `i.base.Manifest()` which can fail, and so the constructor approach would require returning an error from `newImage()`.  So, the `compute()` approach is well-justified.

Filed https://github.com/google/go-containerregistry/issues/2251 for the problem that motivated having our own `image` implementation.

krancour: > Looking a bit deeper...

I had looked as well some time ago and forgotten to report back. Yes. This explanation checks out.

Nothing to see here.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
krancour: Ok... circling back to a lot of prior discussion, I see a lot of choices here (not limited to this specific chain of comments) are motivated by consistency with the existing `oci-download` step. Although it makes me wish we'd made some different choices _a that time_, the appeal to consistency is justifiable, and you can consider me sold.

Specifically in regard to `imageReg` as a field name here, however, I'll make a case for deviating from the `oci-download` step. `imageRef` works nicely there because there's only one ref involved. There's no ambiguity. The semantics here are different, however, with two refs involved, and one of them already designated as `destRef`. To _not_ complement the `destRef` field with a `srcRef` field feels odd and, after some checking around, defies the conventions of every semantically similar thing I've been able to check on -- `mv`, `cp`, `scp`, `docker tag`, `crane copy`. All of these document themselves using source/destination or source/target.

All this being said, let me be clear that this isn't a hill I'm willing to die on. Contemplate it, but consider this feedback non-blocking.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
```diff
@@ -0,0 +1,151 @@
...
+sidebar_label: oci-push
+description: Copies or retags OCI artifacts (container images, Helm charts) between registries.
+---
+
+# `oci-push`
+
+`oci-push` copies or retags OCI artifacts between registries or within the same
+registry. This step supports container images and Helm charts stored in OCI
+registries, making it useful for promoting verified artifacts through a pipeline
+— for example, retagging an image with a release version after verification, or
```
krancour: > retagging an image with a release version after verification

Nit: I'm not sure how this would work exactly. Verification is something that occurs independently of Promotions. Saying this may reinforce a common misconception to the contrary. I don't have specific suggestions on how to reword this at this time, and I would consider this feedback non-blocking, but still perhaps worth some consideration.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
```diff
@@ -0,0 +1,151 @@
...
+copying it to a production registry. Multi-arch image indexes are copied in full.
+
```
krancour, line 14: The introductory text for the `oci-download` step seems to make a point of mentioning that the step supports registry authentication. I don't know if you think that's worth mentioning that here as well. I'm fine with it either way.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/oci-push.md
EronWright: I'm also drawn to the symmetry of `srcRef`/`destRef`, let's make that change.  Update: done.

PR #5785, opened 2026-02-24
feat: git tag step + tag support for git push

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5737
Closes: https://github.com/akuity/kargo/issues/5846

# Demo

First part showcases tagging with annotations and pushing,

Second part showcases tagging, signing the tag with a GPG key and pushing.

https://github.com/user-attachments/assets/2308c1b9-605b-4220-9d68-5b7ae2caee87

### Demo resources

```yaml

kind: ProjectConfig
metadata:
  name: kargo-demo
  namespace: kargo-demo
---
kind: Warehouse
metadata:
  name: my-warehouse
  namespace: kargo-demo
spec:
  interval: 30m
  subscriptions:
  - git:
      repoURL: git@github.com:fuskovic/wh-test-repo.git
      branch: main
      strictSemvers: false
---
apiVersion: v1
kind: Secret
metadata:
  name: my-private-secret
  labels:
stringData:
  repoURL: git@github.com:fuskovic/wh-test-repo.git
  sshPrivateKey: omitted
---
apiVersion: v1
kind: Secret
metadata:
  name: my-gpg-secret
  namespace: kargo-demo
stringData:
  privateKey:  omitted
---
kind: PromotionTask
metadata:
  name: promo-process
  namespace: kargo-demo
spec:
  vars:
  - name: gitRepo
    value: git@github.com:fuskovic/wh-test-repo.git
  - name: outPath
    value: ./out
  steps:
  - uses: git-clone
    config:
      repoURL: ${{ vars.gitRepo }}
      checkout:
      - branch: main
        create: true
        path: ${{ vars.outPath }}
  - uses: git-tag
    config:
      provider: github
      path: ${{ vars.outPath }}
      tag: v1.0.13
      message: "great success!"
      tagger:
        name: "Faris Huskovic"
        email: faris.huskovic@akuity.io
        signingKey: ${{ secret('my-gpg-secret').privateKey }}
  - uses: git-push
    as: push
    config:
      provider: github
      path: ${{ vars.outPath }}
      tag: v1.0.13
---
kind: Stage
metadata:
  name: test
  namespace: kargo-demo
spec:
  requestedFreight:
  - origin:
      kind: Warehouse
      name: my-warehouse
    sources:
      direct: true
  promotionTemplate:
    spec:
      steps:
      - task:
          name: promo-process  
---
kind: Stage
metadata:
  name: uat
  namespace: kargo-demo
spec:
  requestedFreight:
  - origin:
      kind: Warehouse
      name: my-warehouse
    sources:
      stages:
      - test
  promotionTemplate:
    spec:
      steps:
      - task:
          name: promo-process
---
kind: Stage
metadata:
  name: prod
  namespace: kargo-demo
spec:
  requestedFreight:
  - origin:
      kind: Warehouse
      name: my-warehouse
    sources:
      stages:
      - uat
  promotionTemplate:
    spec:
      steps:
      - task:
          name: promo-process
---
```

### DIFF
--- a/pkg/promotion/runner/builtin/git_tagger_test.go
+++ b/pkg/promotion/runner/builtin/git_tagger_test.go
@@ -22,7 +22,8 @@ func Test_gitTagger_convert(t *testing.T) {
 		{
 			name: "path not specified",
 			config: promotion.Config{
-				"tag": "v1.0.0",
+				"tag":     "v1.0.0",
+				"message": "hello",
 			},
 			expectedProblems: []string{
 				"(root): path is required",
@@ -31,8 +32,9 @@ func Test_gitTagger_convert(t *testing.T) {
 		{
 			name: "path is empty string",
 			config: promotion.Config{
-				"path": "",
-				"tag":  "v1.0.0",
+				"path":    "",
+				"tag":     "v1.0.0",
+				"message": "hello",
 			},
 			expectedProblems: []string{
 				"path: String length must be greater than or equal to 1",
@@ -41,7 +43,8 @@ func Test_gitTagger_convert(t *testing.T) {
 		{
 			name: "tag not specified",
 			config: promotion.Config{
-				"path": "/tmp/foo",
+				"path":    "/tmp/foo",
+				"message": "hello",
 			},
 			expectedProblems: []string{
 				"(root): tag is required",
@@ -50,18 +53,40 @@ func Test_gitTagger_convert(t *testing.T) {
 		{
 			name: "tag is empty string",
 			config: promotion.Config{
-				"path": "/tmp/foo",
-				"tag":  "",
+				"path":    "/tmp/foo",
+				"tag":     "",
+				"message": "hello",
 			},
 			expectedProblems: []string{
 				"tag: String length must be greater than or equal to 1",
 			},
 		},
 		{
-			name: "valid config",
+			name: "message not specified",
 			config: promotion.Config{
 				"path": "/tmp/foo",
-				"tag":  "v1.0.0",
+			},
+			expectedProblems: []string{
+				"(root): message is required",
+			},
+		},
+		{
+			name: "message is empty string",
+			config: promotion.Config{
+				"path":    "/tmp/foo",
+				"tag":     "v1.0.0",
+				"message": "",
+			},
+			expectedProblems: []string{
+				"message: String length must be greater than or equal to 1",
+			},
+		},
+		{
+			name: "valid config",
+			config: promotion.Config{
+				"path":    "/tmp/foo",
+				"tag":     "v1.0.0",
+				"message": "hello",
 			},
 		},
 	}


pkg/promotion/runner/builtin/schemas/git-push-config.json
```diff
@@ -22,7 +22,7 @@
     },
     "targetBranch": {
       "type": "string",
-      "description": "The target branch to push to. Mutually exclusive with 'generateTargetBranch=true'. If neither of these is provided, the target branch will be the currently checked out branch."
+      "description": "The target branch to push to. Mutually exclusive with 'generateTargetBranch=true' and 'tag'. If neither of these is provided, the target branch will be the currently checked out branch."
```
krancour: Nit:

```suggestion
      "description": "The target branch to push to. Mutually exclusive with 'generateTargetBranch=true' and 'tag'. If none of these are provided, the target branch will be the currently checked out branch."
```

pkg/promotion/runner/builtin/schemas/git-tag-config.json
```diff
@@ -0,0 +1,18 @@
...
+      "minLength": 1
+    },
+    "tag": {
+      "type": "string",
+      "description": "The tag to create in the repository.",
+      "minLength": 1
+    }
+  },
+  "required": ["path", "tag"]
+}
```
krancour:
```suggestion
}

```

pkg/promotion/runner/builtin/schemas/git-tag-config.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "GitTagConfig",
+  "type": "object",
+  "properties": {
+    "path": {
+      "type": "string",
+      "description": "The path to the local Git repository.",
```
krancour: For consistency with existing steps:

```suggestion
      "description": "The path to a working directory of a local repository.",
```

The `git-clone` step clones a bare repo and the path here isn't to the repo, but to one of possibly many associated working trees, so this is more accurate terminology as well.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -5,9 +5,9 @@ description: Pushes the committed changes in a specified working tree to a speci
 
 # `git-push`
 
-`git-push` pushes the committed changes in a specified working tree to a
-specified branch in the remote repository. This step typically follows a
-[`git-commit` step](git-commit.md) and is often followed by a
+`git-push` can either push committed changes in a specified working tree to a
+specified branch or a new tag to the remote repository. This step typically 
+follows a [`git-commit` step](git-commit.md) and is often followed by a
 [`git-open-pr` step](git-open-pr.md).
```
krancour:
```suggestion
`git-push` can push committed changes or new tags from specified working tree to
the remote repository. This step typically follows a
[`git-commit` step](git-commit.md) and/or [`git-tag` step](git-tag.md) and is
often followed by a [`git-open-pr` step](git-open-pr.md).
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -28,14 +28,24 @@ Stages that write to the same branch do not write to the same files.
...
+
+:::
+
 ## Configuration
 
 | Name | Type | Required | Description |
 |------|------|----------|-------------|
 | `path` | `string` | Y | Path to a Git working tree containing committed changes. |
-| `targetBranch` | `string` | N | The branch to push to in the remote repository. Mutually exclusive with `generateTargetBranch=true`. If neither of these is provided, the target branch will be the same as the branch currently checked out in the working tree. |
+| `targetBranch` | `string` | N | The branch to push to in the remote repository. Mutually exclusive with `generateTargetBranch=true` and `tag`. If neither of these is provided, the target branch will be the same as the branch currently checked out in the working tree. |
```
krancour: Nit:

```suggestion
| `targetBranch` | `string` | N | The branch to push to in the remote repository. Mutually exclusive with `generateTargetBranch=true` and `tag`. If none of these are provided, the target branch will be the same as the branch currently checked out in the working tree. |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -28,14 +28,24 @@ Stages that write to the same branch do not write to the same files.
...
+| `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of 'true' is mutually exclusive with `targetBranch` and `tag`. If neither of these is provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
```
krancour:
```suggestion
| `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of 'true' is mutually exclusive with `targetBranch` and `tag`. If none of these are provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -97,3 +108,20 @@ steps:
     generateTargetBranch: true
 # Open a PR and wait for it to be merged or closed...
 ```
+
+### Pushing Tags
+
+In this example, a tag is pushed to the remote repository. The `git-tag` step 
+typically precedes the [`git-tag`](git-tag.md) in order to create the new tag.
```
krancour: I think `git-tag` preceding `git-tag` was a mistake, but regardless, I'd shorten this quite a bit because I don't think this is (or will be) common usage.

```suggestion
In this example, a new tag is pushed to the remote repository.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,73 @@
...
+
+# `git-tag`
+
+The `git-tag` step creates a new tag in a local Git repository. This step is commonly used to mark specific commits with a tag, which can be useful for versioning or tracking changes in a repository.
+
+## Configuration
+
+| Name   | Type     | Required | Description                                                                 |
+|--------|----------|----------|-----------------------------------------------------------------------------|
+| `path` | `string` | Y        | Path to the local Git repository where the tag should be created. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
```
krancour:
```suggestion
| `path` | `string` | Y        | Path to a working directory of a local repository. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Output
+
+| Name  | Type     | Description                                                                 |
+|-------|----------|-----------------------------------------------------------------------------|
+| `tag` | `string` | The name of the tag that was created by this step. This can be referenced in subsequent steps. |
+
+:::caution
+
+If the specified tag already exists, the git-tag step will fail.
```
krancour:
```suggestion
If the specified tag already exists in the remote repository, the `git-tag` step
will fail.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,73 @@
...
+
+:::
+
+## Examples
+
+### Basic Usage
+
+In this example, the `git-tag` step creates a tag named `v1.0.0` in the local Git repository located at `./out`.
```
krancour: Saying less here makes this more accurate. `./out` is a working tree and while it _is_ the directory the command will run from, the tag gets created in the bare repo, which is not at `./out`.

(If you went and checked another working tree of the same repo, you'd see the new tag also exists there. So this is proof the new tag isn't actually isolated to the `./out` directory.)

```suggestion
In this example, the `git-tag` step creates a tag named `v1.0.0` in a local Git repository.
```

pkg/controller/git/work_tree.go
```diff
@@ -42,6 +42,8 @@ type WorkTree interface {
 	// CreateOrphanedBranch creates a new branch that shares no commit history
 	// with any other branch.
 	CreateOrphanedBranch(branch string) error
+	// CreateTag creates a new tag.
+	CreateTag(tag string) error
```
krancour: imho, this is nicer. Dealer's choice, as this is 0% a hill I would die on.


```suggestion
	// Tag creates a new tag with the specified name.
	Tag(name string) error
```

pkg/controller/git/work_tree.go
```diff
@@ -584,6 +593,9 @@ type PushOptions struct {
 	// be useful when pushing changes to a remote branch that has been updated
 	// in the time since the local branch was last pulled.
 	PullRebase bool
+	// Tag specifies an optional tag to push to the remote repository. Mutually
+	// exclusive with TargetBranch.
```
krancour: Worth noting that `PullRebase == true` would have no effect when a tag is pushed?

pkg/controller/git/work_tree.go
```diff
@@ -622,15 +634,24 @@ func (w *workTree) Push(opts *PushOptions) error {
 			}
 		}
 	}
```
krancour, line 702: I would consider moving all of this into the default case of your switch statement below. Visually, it would become far easier to dismiss all of this as inapplicable for the case of pushing a tag.

pkg/gitprovider/registry.go
```diff
@@ -33,8 +33,10 @@ func New(repoURL string, opts *Options) (Interface, error) {
 		return nil, fmt.Errorf("no registered providers with name %q", opts.Name)
 	}
 	for _, reg := range registeredProviders {
-		if reg.Predicate(repoURL) {
-			return reg.NewProvider(repoURL, opts)
+		if reg.Predicate != nil && reg.NewProvider != nil {
+			if reg.Predicate(repoURL) {
+				return reg.NewProvider(repoURL, opts)
+			}
```
krancour, line 39: Total aside and note to self... I realize this code wasn't ever refactored to use the PredicateBasedRegistry. Out of scope for this PR, but I should fix that.

pkg/promotion/runner/builtin/git_pusher.go
```diff
@@ -238,6 +247,7 @@ func (g *gitPushPusher) run(
 			stateKeyBranch:    pushOpts.TargetBranch,
 			stateKeyCommit:    commitID,
 			stateKeyCommitURL: commitURL,
+			stateKeyTag:       pushOpts.Tag,
```
krancour: Should this be set conditionally?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -46,6 +56,7 @@ Stages that write to the same branch do not write to the same files.
 | `branch` | `string` | The name of the remote branch pushed to by this step. This is especially useful when the `generateTargetBranch=true` option has been used, in which case a subsequent [`git-open-pr`](git-open-pr.md) will typically reference this output to learn what branch to use as the head branch of a new pull request. |
 | `commit` | `string` | The ID (SHA) of the commit pushed by this step. |
 | `commitURL` | `string` | The URL of the commit that was pushed to the remote repository. |
+| `tag` | `string` | The tag that was pushed to the remote repository. |
```
krancour: This is a suggestion -- but if getting the URL of the tag is an enormous pain, this is something I wouldn't bother with. See what's involved first.

```suggestion
| `tag.name` | `string` | If applicable, the tag that was pushed to the remote repository. |
| `tag.url` | `string` | If applicable, the URL of the tag that was pushed to the remote repository. |
```

pkg/promotion/runner/builtin/git_pusher.go
krancour: Note that I am in favor of continuing to unconditionally include the `commitID` and `commitURL` in the output. Those are still 100% valid and useful when a tag was pushed.

pkg/promotion/runner/builtin/git_pusher_test.go
```diff
@@ -150,22 +171,140 @@ func Test_gitPusher_convert(t *testing.T) {
 }
 
 func Test_gitPusher_run(t *testing.T) {
```
krancour, line 173: These tests feel a lot harder to follow now. This is a case where I would recommend _not_ using table-driven testing and instead keep setup as it was, then just put two separate `t.Run()` calls after "Now we can proceed to test gitPusher..."

pkg/promotion/runner/builtin/git_tagger.go
```diff
@@ -0,0 +1,86 @@
...
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("error loading working tree from %s: %w", cfg.Path, err)
+	}
+	if err = workTree.CreateTag(cfg.Tag); err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("error creating tag %s: %w", cfg.Tag, err)
+	}
+	return promotion.StepResult{
+		Status: kargoapi.PromotionStepStatusSucceeded,
+		Output: map[string]any{stateKeyTag: cfg.Tag},
```
krancour: It's possibly useful to include the commit the tag points to in this output.

pkg/promotion/runner/builtin/git_tagger_test.go
```diff
@@ -0,0 +1,152 @@
...
+	// Test creating a tag
+	res, err := runner.run(
+		context.Background(),
+		&promotion.StepContext{WorkDir: workDir},
+		builtin.GitTagConfig{
+			Path: "master",
+			Tag:  "v1.0.0",
+		},
+	)
+	// Verify
```
krancour, line 173: I think we can additionally verify that the tag actually got created. We don't have functionality for listing tags or checking for the existence of a tag, nor do I think we have any other motivation to create that functionality... but we should be able to validate the new tag exists using the `WorkTree` interface's existing `Checkout()` method.

fuskovic:
@krancour I've pushed the suggested changes and updated the demo in the PR description to show the commit information. Some notes:

- The commit URL only populates when the provider is set, so I've updated the output to omit that key/value pair entirely if provider is not provided or if it is and we fail to get the commit URL. 

- You'll notice in the current demo the branch is included in the output for git pushing tags. I'm wondering if we want to keep that behavior or have tag and branch be mutually exclusive in the output?

Alternatively, we can always set all output keys unconditionally so they can always be referenced dynamically in other steps (regardless if they're empty).

Thoughts?

pkg/promotion/runner/builtin/schemas/git-tag-config.json
```diff
@@ -0,0 +1,23 @@
...
+      "minLength": 1
+    },
+    "tag": {
+      "type": "string",
+      "description": "The tag to create in the repository.",
+      "minLength": 1
+    },
+    "provider": {
+      "type": "string",
+      "description": "The name of the Git provider to use. Currently 'azure', 'bitbucket', 'gitea', 'github', and 'gitlab' are supported. Kargo will try to infer the provider if it is not explicitly specified.",
```
krancour: Tagging is a pure Git operation, and a local one too. No Git hosting provider involved, so we shouldn't need this. I see why you're doing it, though. It's to help construct a commit URL as output for this step, but I don't think that's an output that would be expected from this step.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,75 @@
...
+
+| Name  | Type     | Description                                                                 |
+|-------|----------|-----------------------------------------------------------------------------|
+| `tag` | `string` | The name of the tag that was created by this step. This can be referenced in subsequent steps. |
+| `commit` | `string` | The ID (SHA) of the commit pushed by this step. |
+
+:::caution
+
+If the specified tag already exists in the remote repository, the `git-tag` step
+will fail.
```
krancour: Is this true? I think the failure wouldn't occur until trying to push the tag to the remote.

Maybe we just move this warning into the push step's docs?

pkg/promotion/runner/builtin/git_pusher.go
```diff
@@ -224,21 +233,27 @@ func (g *gitPushPusher) run(
...
 		if commitURL, err = gitProvider.GetCommitURL(workTree.URL(), commitID); err != nil {
 			logger.Error(err, "unable to get commit URL from Git provider")
+		} else {
+			output[stateKeyCommitURL] = commitURL
 		}
 	}
-
+	if pushOpts.Tag != "" {
+		output[stateKeyTag] = pushOpts.Tag
+	}
```
krancour: I would probably leave this out. If the step was configured to push a tag, that tag's name is already known, which makes this noise.

pkg/promotion/runner/builtin/git_tagger.go
```diff
@@ -0,0 +1,99 @@
...
+
+	commitID, err := workTree.LastCommitID()
+	if err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored},
+			fmt.Errorf("error getting last commit ID: %w", err)
+	}
+
+	output := map[string]any{
+		stateKeyTag:    cfg.Tag,
+		stateKeyCommit: commitID,
```
krancour: Agree with including commit ID in the output. That's potentially useful information that wasn't known up to this point in the process. As mentioned elsewhere, though, this step would have been configured with the tag name, so we're not revealing anything new by including it in the output.

pkg/controller/git/work_tree.go
```diff
@@ -596,41 +610,50 @@ func (w *workTree) Push(opts *PushOptions) error {
 		opts = &PushOptions{}
 	}
 	targetBranch := opts.TargetBranch
-	if targetBranch == "" {
+	if targetBranch == "" && opts.Tag == "" {
```
krancour: This is the spot in the code I was looking for earlier in our meeting and and couldn't find.

`targetBranch` is only used in your `default` case, so this can move to there.

I might suggest some other, very slight refactoring for readability. You're deciding on a value for `artifact` up front, handling divergent bits of the two cases, then handling common bits, referencing `artifact` only in error messages. Those common bits are so brief (especially since checking for non-fast forward errors isn't required at all in the case of pushing a tag) that I think its would be less code and more readable if we skipped the `artifact` variable and just moved setting the force flag (or not), the exec, and the two different variants of error-handling right into the cases. The whole method end up as completely divergent paths for pushing a tag v pushing to a branch. It will be more readable without being an egregious repetition of anything that's actually the same.

krancour:
Looking good. All of my feedback is minor.

EronWright (commented):
As a stretch goal or follow-up, consider adding support for **tag signing**, since Kargo already supports commit signing.
https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -5,10 +5,10 @@ description: Pushes the committed changes in a specified working tree to a speci
 
 # `git-push`
 
-`git-push` pushes the committed changes in a specified working tree to a
-specified branch in the remote repository. This step typically follows a
-[`git-commit` step](git-commit.md) and is often followed by a
-[`git-open-pr` step](git-open-pr.md).
+`git-push` can push committed changes or new tags from specified working tree to
```
krancour:
```suggestion
`git-push` can push committed changes or new tags from a specified working tree to
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -28,14 +28,24 @@ Stages that write to the same branch do not write to the same files.
 
 :::
 
+
+:::note
+
+For a tag push, an attempt to pull/rebase first is not made as tags are 
+immutable and any existing tag with the same name on the remote would cause 
+the pull/rebase to fail.
```
krancour: imho, I'd state that there is no pull/rebase retry loop for tags without getting into why.

The reason is that, strictly speaking, tags _aren't_ immutable. You can force push them. _You shouldn't_, but you can.

I often worry about what kind of strange doors we open up by saying more than we need to. I don't want to invite weird feature request starting with, "Well, we _do_ force push tags as part of our workflow..."

Which reminds me that we should probably make force push true mutually exclusive with pushing a tag. I'm confident it _can_ work, but I'm also confident it's a foot bazooka. I'd rather we default to not allowing it and have a very high bar for being convinced otherwise.

pkg/controller/git/work_tree.go
```diff
@@ -595,42 +609,54 @@ func (w *workTree) Push(opts *PushOptions) error {
...
-	targetBranch := opts.TargetBranch
-	if targetBranch == "" {
-		var err error
-		if targetBranch, err = w.CurrentBranch(); err != nil {
-			return err
+	args := []string{"push", "origin"}
+	switch {
+	case opts.Tag != "":
+		args = append(args, "tag", opts.Tag)
+		if opts.Force {
```
krancour: See other note about this in the docs. We probably shouldn't do this.

pkg/controller/git/work_tree.go
```diff
@@ -595,42 +609,54 @@ func (w *workTree) Push(opts *PushOptions) error {
...
-		// We only want to pull and rebase if the remote branch exists.
-		if exists {
-			if _, err = libExec.Exec(w.buildGitCommand("pull", "--rebase", "origin", targetBranch)); err != nil {
-				// The error we're most concerned with is a merge conflict requiring
-				// manual resolution, because it's an error that no amount of retries
-				// will fix. If we find that a rebase is in progress, this is what
-				// has happened.
-				if isRebasing, isRebasingErr := w.IsRebasing(); isRebasingErr == nil && isRebasing {
-					return ErrMergeConflict
+	default:
```
krancour: Nice. This is really easy to follow now. Thanks.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -28,6 +28,13 @@ Stages that write to the same branch do not write to the same files.
 
 :::
 
+
+:::note
+
+For a tag push, there is no pull/rebase retry loop for tags
```
krancour: Need to reword. This says "for tags" at both the beginning and end.

pkg/promotion/runner/builtin/schemas/git-push-config.json
```diff
@@ -33,27 +33,42 @@
...
+        "tag": { "enum": ["", null] }
       },
       "required": ["targetBranch"]
     },
     {
       "properties": {
         "generateTargetBranch": { "enum": [false, null] },
-        "targetBranch": { "enum": ["", null] }
+        "targetBranch": { "enum": ["", null] },
+        "tag": { "minLength": 1 }
```
krancour:
```suggestion
        "tag": { "minLength": 1 },
        "force": { "enum": [false, null] }
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -38,10 +45,11 @@ system to access the git repos.
...
+| `tag` | `string` | N | An optional tag to push to the remote repository. Mutually exclusive with `generateTargetBranch` and `targetBranch`. |
+| `force` | `boolean` | N | Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). **Use with caution** as this will overwrite any commits that exist on the remote branch but not in your local branch. Default is `false`. If `tag` is set, `force` has no effect as we do not support force-pushing tags.  |
```
krancour: I would reword this so that instead of presenting `force` as having no effect when pushing a tag, the `force` and `tag` fields are presented as being mutually exclusive.

See https://github.com/akuity/kargo/pull/5785/changes#r2890293218

pkg/promotion/runner/builtin/schemas/git-push-config.json
```diff
@@ -33,27 +33,42 @@
       "type": "boolean",
       "description": "Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). Use with caution as this will overwrite any commits that exist on the remote branch but not in your local branch.",
       "default": false
+    },
+    "tag": {
+      "type": "string",
+      "description": "An optional tag to push to the remote repository. Mutually exclusive with 'generateTargetBranch=true' and 'targetBranch'."
```
krancour:
```suggestion
      "description": "A tag to push to the remote repository. Mutually exclusive with 'generateTargetBranch=true', 'targetBranch', and 'force=true'."
```

pkg/promotion/runner/builtin/schemas/git-push-config.json
```diff
@@ -33,27 +33,42 @@
       "type": "boolean",
       "description": "Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). Use with caution as this will overwrite any commits that exist on the remote branch but not in your local branch.",
```
krancour:
```suggestion
      "description": "Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). Use with caution as this will overwrite any commits that exist on the remote branch but not in your local branch. Mutually exclusive with 'tag'.",
```

pkg/controller/git/work_tree.go
```diff
@@ -592,6 +601,11 @@ type PushOptions struct {
 	// be useful when pushing changes to a remote branch that has been updated
 	// in the time since the local branch was last pulled.
 	PullRebase bool
+	// Tag specifies an optional tag to push to the remote repository. Mutually
+	// exclusive with TargetBranch. Because tags are immutable and any existing
+	// tag with the same name on the remote would cause the pull/rebase to fail,
+	// PullRebase will be ignored if Tag is specified.
```
krancour: In this low-level code, I wouldn't present Tag and TargetBranch as being mutually exclusive. That's something that's enforced on user-defined input at a much higher-level. The correct thing to do here is explain precedence when both of these conflicting fields have non-empty values. According to the code in the `run()` method, `Tag` wins.

pkg/controller/git/work_tree.go
```diff
@@ -603,42 +617,51 @@ func (w *workTree) Push(opts *PushOptions) error {
 	if opts == nil {
 		opts = &PushOptions{}
 	}
-	targetBranch := opts.TargetBranch
-	if targetBranch == "" {
-		var err error
-		if targetBranch, err = w.CurrentBranch(); err != nil {
-			return err
+	args := []string{"push", "origin"}
+	switch {
```
krancour: The refactoring you already did on this switch was great, but since the whole thing is now a simple either/or, I would replace this with an `if` and an early return.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,88 @@
...
+## Configuration
+
+| Name   | Type     | Required | Description                                                                 |
+|--------|----------|----------|-----------------------------------------------------------------------------|
+| `path` | `string` | Y        | Path to a working directory of a local repository. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `tag`  | `string` | Y        | The tag to create. |
+| `author.name` | `string` | N | The committer's name. Defaults to `Kargo` if not provided. |
+| `author.email` | `string` | N | The committer's email address. Defaults to `no-reply@kargo.io` if not provided. |
+| `author.signingKey` | `string` | Y | The GPG signing key for the author. |
+| `author.signingMessage` | `string` | Y | The message to annotate the tag with. |
```
krancour: I'm iffy on the use of "author" here.

On one hand, signing a tag doesn't make you an author. You're more like a "signer."

On the other hand, if we leave it this way, it has a nice symmetry with the commit step.

cc @thomastaylor312 and @EronWright for additional opinions.

pkg/controller/git/base_repo.go
```diff
@@ -188,6 +188,15 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
 				return fmt.Errorf("error configuring commit gpg signing: %w", err)
 			}
 
+			// enable signing for tags as well. You'll notice the capital 'S' in
+			// 'gpgSign' - this is inentional, as git-config uses 'gpgSign' for
+			// tag signing and 'gpgsign' for commit signing.
+			cmd = b.buildGitCommand("config", "--global", "tag.gpgSign", "true")
+			cmd.Dir = homeDir
```
krancour, line 195: #5857 explains why this is necessary. That PR fixes the existing spots, but obviously doesn't account for these new lines.

```suggestion
			cmd.Dir = homeDir
			b.setCmdHome(cmd, homeDir)
```

pkg/controller/git/base_repo.go
```diff
@@ -188,6 +188,16 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
 				return fmt.Errorf("error configuring commit gpg signing: %w", err)
 			}
 
+			// enable signing for tags as well. You'll notice the capital 'S' in
+			// 'gpgSign' - this is inentional, as git-config uses 'gpgSign' for
+			// tag signing and 'gpgsign' for commit signing.
```
krancour: Interesting. TIL that these config keys are case insensitive. In git's own documentation, it's a capital S, so you're correct here, but rather than call out the difference from commit signing configuration a few lines up, I'd fix the commit signing configuration to be consistent with this.

pkg/controller/git/work_tree.go
```diff
@@ -304,6 +306,71 @@ func (w *workTree) CreateOrphanedBranch(branch string) error {
...
+	if _, err := libExec.Exec(w.buildGitCommand(
+		"tag",
+		"-s",
+		tag,
+		"-m",
+		signingMsg,
+	)); err != nil {
+		return fmt.Errorf("error creating signed tag %q", err)
+	}
+	return nil
```
krancour, line 355:
```suggestion
	var homeDir string
	if opts.Author != nil {
		// This signing config is specific to this tag, so we will override
		// repository-level signing config by creating a temporary home
		// directory, setting the tag configuration "globally" within it, and
		// then ensuring the git commit command uses that home directory.
		var err error
		if homeDir, err = os.MkdirTemp(w.homeDir, ""); err != nil {
			return fmt.Errorf(
				"error creating virtual home directory %q for tag command: %w",
				homeDir, err,
			)
		}
		defer func() {
			if cleanErr := os.RemoveAll(homeDir); cleanErr != nil {
				logging.LoggerFromContext(context.TODO()).
					Error(cleanErr, "error removing virtual home directory", "path", homeDir)
			}
		}()
		if err = w.setupAuthor(homeDir, opts.Author); err != nil {
			return fmt.Errorf(
				"error setting up author information for commit command: %w", err,
			)
		}
	}

	signingMsg := "Signed by Kargo"
	if opts.SigningMsg != "" {
		signingMsg = opts.SigningMsg
	}

	cmd := w.buildGitCommand("tag", "-s", tag, "-m", signingMsg)
	if homeDir != "" {
		// Override the home directory set by w.buildGitCommand().
		w.setCmdHome(cmd, homeDir)
	}
	if _, err := libExec.Exec(cmd); err != nil {
		return fmt.Errorf("error creating signed tag %q", err)
	}
	return nil
```

pkg/controller/git/base_repo.go
krancour: Turns out this is case insensitive, but capital S matches git's own docs.

```suggestion
			cmd = b.buildGitCommand("config", "--global", "commit.gpgSign", "true")
```

pkg/controller/git/base_repo.go
```diff
@@ -188,6 +188,16 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
 				return fmt.Errorf("error configuring commit gpg signing: %w", err)
 			}
 
+			// Enable signing for tags as well. You'll notice the capital 'S' in
+			// 'gpgSign' - this is inentional, as git-config uses 'gpgSign' for
+			// tag signing and 'gpgsign' for commit signing.
```
krancour: No longer necessary to call out the different casing because it's changed above.

```suggestion
			// Enable signing for tags as well.
```

pkg/controller/git/work_tree.go
```diff
@@ -304,6 +306,72 @@ func (w *workTree) CreateOrphanedBranch(branch string) error {
...
+			)
+		}
+	}
+
+	signingMsg := "Signed by Kargo"
+	if opts.SigningMsg != "" {
+		signingMsg = opts.SigningMsg
+	}
+
+	cmd := w.buildGitCommand("tag", "-s", tag, "-m", signingMsg)
```
krancour: `-s` (which implies `-a`) will fail in the event no signing key was configured. If we use just `-a` (indicates we're creating an annotated tag), the signing will happen automatically if the signing key _is_ configured.

```suggestion
	cmd := w.buildGitCommand("tag", "-a", tag, "-m", signingMsg)
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -5,10 +5,10 @@ description: Pushes the committed changes in a specified working tree to a speci
...
 # `git-push`
 
-`git-push` pushes the committed changes in a specified working tree to a
-specified branch in the remote repository. This step typically follows a
-[`git-commit` step](git-commit.md) and is often followed by a
-[`git-open-pr` step](git-open-pr.md).
+`git-push` can push committed changes or new tags from a specified working tree to
+the remote repository. This step typically follows a
+[`git-commit` step](git-commit.md) and/or [`git-tag` step](git-tag.md) and is
+often followed by a [`git-open-pr` step](git-open-pr.md).
```
krancour: "can" is reading awkwardly here.

```suggestion
`git-push` pushes committed changes or new tags from a specified working tree to
the remote repository. This step typically follows a
[`git-commit` step](git-commit.md) and/or [`git-tag` step](git-tag.md) and is
often followed by a [`git-open-pr` step](git-open-pr.md).
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -38,10 +45,11 @@ system to access the git repos.
...
-| `force` | `boolean` | N | Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). **Use with caution** as this will overwrite any commits that exist on the remote branch but not in your local branch. Default is `false`. |
+| `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of `true` is mutually exclusive with `targetBranch` and `tag`. If none of these are provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
+| `tag` | `string` | N | An optional tag to push to the remote repository. Mutually exclusive with `generateTargetBranch` and `targetBranch`. |
```
krancour: "optional" is awkward here. It implies that this is like an "add on" to something the step is going to do anyway, but we know it's actually an either or.

```suggestion
| `tag` | `string` | N | An tag to push to the remote repository. Mutually exclusive with `generateTargetBranch` and `targetBranch`. |
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,88 @@
+---
+sidebar_label: git-tag
+description: Creates a new tag for the latest committed changes.
+---
+
+# `git-tag`
+
+The `git-tag` step creates a new tag in a local Git repository. This step is commonly used to mark specific commits with a tag, which can be useful for versioning or tracking changes in a repository.
```
krancour: I think it's important to note it's an annotated tag. The rest of it is awkward because the common use described is it's _only_ use and if we drop the word "commonly," we're defining what a git tag is and why it's useful, which feels unnecessary.

```suggestion
The `git-tag` step creates a new, annotated tag in a local Git repository
referencing the current `HEAD` of a checked-out branch.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
krancour: After more thought, we should use `signer` instead of `author`.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,88 @@
...
+    tag: v1.0.0
+- uses: git-push
+  config:
+    path: ./out
+    tag: v1.0.0
+```
+
+### Creating A Signed Tag
+
+In this example, the `git-tag` step creates a signed tag using the provided author information.
```
krancour: I think we need to make it clear that this may have been configured at the system level by a Kargo admin and that the options show here will override that configuration if it exists.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,88 @@
...
+```yaml
+steps:
+- uses: git-tag
+  config:
+    path: ./out
+    tag: v1.0.0
+    author:
+      name: yourname
+      email: your@inbox.com
+      signingKey: <base64-encoded-ascii-armored-gpg-key>
```
krancour: This will unfortunately invite negligent readers to paste a key directly in the step. We should show the and expression with the `secret()` function being used for this instead.

pkg/controller/git/work_tree.go
```diff
@@ -304,6 +306,72 @@ func (w *workTree) CreateOrphanedBranch(branch string) error {
 	return w.Clean()
 }
 
+// TagOptions represents options for creating a new git tag.
+type TagOptions struct {
+	// Author is the author of the tag. If nil, the default author already
+	// configured in the git repository will be used.
+	Author *User
+	// SigningMsg is the tag message to use if signing key is provided.
```
krancour: This isn't really connected in any way to the presence or absence of a key.

```suggestion
	// SigningMsg is the message with which the tag will be annotated. If empty,
	// a default message will be used.
```

pkg/controller/git/work_tree.go
```diff
@@ -304,6 +306,72 @@ func (w *workTree) CreateOrphanedBranch(branch string) error {
 	return w.Clean()
 }
 
+// TagOptions represents options for creating a new git tag.
+type TagOptions struct {
+	// Author is the author of the tag. If nil, the default author already
+	// configured in the git repository will be used.
+	Author *User
```
krancour: Change to `Signer`.

pkg/controller/git/work_tree.go
```diff
@@ -304,6 +306,72 @@ func (w *workTree) CreateOrphanedBranch(branch string) error {
...
+	// SigningMsg is the tag message to use if signing key is provided.
+	SigningMsg string
+}
+
+// CreateTag creates a new tag with the specified name. If the author field of
+// opts is non-nil, the tag will be signed using the provided author information
+// and signing message. If the author field is nil, the tag will be unsigned.
```
krancour: This isn't quite true. Signing might have been configured by options when the repo was cloned or may have been configured at the system level.

krancour: I would just drop the comment altogether because this just implements a method on an interface and you've already left a succinct and accurate comment on that, and what callers outside this package will see.

```suggestion
```

pkg/controller/git/work_tree.go
```diff
@@ -592,6 +660,10 @@ type PushOptions struct {
 	// be useful when pushing changes to a remote branch that has been updated
 	// in the time since the local branch was last pulled.
 	PullRebase bool
+	// Tag specifies an optional tag to push to the remote repository.
+	// If this field is non-empty along with TargetBranch, the tag will be
+	// pushed instead of a branch, and all other fields will be ignored.
```
krancour: Optional is awkward here because it implies this just piles onto base behavior, but we know it's actually an either/or.

```suggestion
	// Tag specifies a tag to push to the remote repository. If this field and
	// TargetBranch are both non-empty, this field takes precedence and the tag
	// will be pushed -- the branch will not.
```

pkg/controller/git/work_tree.go
krancour:
```suggestion
// TargetBranch specifies a remote branch to push to. If empty, the remote
// branch's name will be assumed to be the same as the branch checked out
// locally. Whether this field is empty or non-empty, if Tag is non-empty,
// it takes precedence and the tag will be pushed -- the branch will not.
```

pkg/promotion/runner/builtin/schemas/git-tag-config.json
```diff
@@ -0,0 +1,44 @@
...
+      "type": "string",
+      "description": "The path to a working directory of a local repository.",
+      "minLength": 1
+    },
+    "tag": {
+      "type": "string",
+      "description": "The tag to create in the repository.",
+      "minLength": 1
+    },
+    "author": {
```
krancour: As mentioned elsewhere, let's go with `signer`.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,89 @@
...
+referencing the current `HEAD` of a checked-out branch.
+
+## Configuration
+
+| Name   | Type     | Required | Description                                                                 |
+|--------|----------|----------|-----------------------------------------------------------------------------|
+| `path` | `string` | Y        | Path to a working directory of a local repository. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `tag`  | `string` | Y        | The tag to create. |
+| `signer.name` | `string` | N | The signer's name. Defaults to `Kargo` if not provided. |
+| `signer.email` | `string` | N | The signer's email address. Defaults to `no-reply@kargo.io` if not provided. |
```
krancour: I just realized -- I don't think name and email are relevant to signing.

krancour: Nevermind... researched it more and these values might appear in the tag's annotation and may also play a role in loading the key.

Leave these be.

krancour: But the descriptions need updates. Those defaults aren't static. If not specified, these value can come from repo-level configuration specified when the repo was cloned, or this also can be configured at the system level.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,97 @@
...
+## Configuration
+
+| Name   | Type     | Required | Description                                                                 |
+|--------|----------|----------|-----------------------------------------------------------------------------|
+| `path` | `string` | Y        | Path to a working directory of a local repository. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `tag`  | `string` | Y        | The tag to create. |
+| `signer.name` | `string` | N | The signer's name. If unspecified, defaults to the repo-level configuration specified when the repo was cloned. Can also be configured at the system level. |
+| `signer.email` | `string` | N | The signer's email address. If unspecified, defaults to the repo-level configuration specified when the repo was cloned. Can also be configured at the system level. |
+| `signer.signingKey` | `string` | Y | The GPG signing key for the signer. |
+| `signer.signingMessage` | `string` | Y | The message to annotate the tag with. |
```
krancour: These also default to repo or system-level config if unspecified.

krancour: You're gonna want to murder me. `signer` was an improvement on `author`, but I've realized there's no guarantee than any signing is happening. We're unconditionally creating an annotated tag, but it might not be signed.

Inspecting an annotated with or without a signature, git formally would call this user the "tagger."

```
git tag -v my-test-tag
object 9aa4f0dfaf22f5ed6720e70734a3a63a71554330
type commit
tag my-test-tag
tagger Kent Rancourt <kent.rancourt@gmail.com> 1772733851 -0500

This is a test.
```

```
git tag -v my-signed-test-tag                     
object 9aa4f0dfaf22f5ed6720e70734a3a63a71554330
type commit
tag my-signed-test-tag
tagger Kent Rancourt <kent.rancourt@gmail.com> 1772733958 -0500

This is a test.
gpg: Signature made Thu Mar  5 13:05:58 2026 EST
gpg:                using RSA key 8E09BE17D094189B079D2D5DBFA8785F5C39E3FE
gpg: Good signature from "Kent Rancourt <kent.rancourt@microsoft.com>" [unknown]
gpg:                 aka "Kent Rancourt <kent.rancourt@gmail.com>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 0BD9 3986 840F C7E8 0E77  5F0A 38E6 6067 946A 0A68
     Subkey fingerprint: 8E09 BE17 D094 189B 079D  2D5D BFA8 785F 5C39 E3FE
```

We probably ought to go with "tagger"

fuskovic: lol nah it's all good. Tagger makes sense.

krancour: TIL I don't currently trust my own key.

krancour:
This LGTM @fuskovic. Just as a sanity check, lmk if you still have this working e2e after the many changes made today.

fuskovic:
@krancour I've updated the demo in the PR description with end-to-end tests following the last round of suggested changes.

pkg/promotion/runner/builtin/schemas/git-commit-config.json
```diff
@@ -22,7 +22,7 @@
         },
         "signingKey": {
           "type": "string",
-          "description": "The GPG signing key for the author."
+          "description": "The GPG signing key for the author. If provided 'email' and 'name' must match the ones in the uid of the associated GPG key."
```
EronWright: Curious how this requirement came about, how it is enforced, and whether we think these are "verified" email addresses.

pkg/promotion/runner/builtin/schemas/git-tag-config.json
```diff
@@ -0,0 +1,44 @@
...
+        },
+        "name": {
+          "type": "string",
+          "description": "The name of the tagger. If unspecified, defaults to the repo-level configuration specified when the repo was cloned. Can also be configured at the system level.",
+          "minLength": 1
+        },
+        "signingKey": {
+          "type": "string",
+          "description": "The GPG signing key for the tagger. If provided 'email' and 'name' must also be provided and must match the 'email' and 'name' in the uid of the associated GPG key." 
+        }
```
EronWright: Consider 'force' attribute.

pkg/promotion/runner/builtin/git_pusher.go
```diff
@@ -154,7 +154,13 @@ func (g *gitPushPusher) run(
 		// branch is specific to this Promotion only holds, it is also safe to do this.
 		pushOpts.Force = true
 	}
-
+	if cfg.Tag != "" {
+		pushOpts.Tag = cfg.Tag
+		// If we're pushing a tag, we should not attempt to pull/rebase first as
+		// tags are immutable and any existing tag with the same name on the remote
+		// would cause the pull/rebase to fail.
+		pushOpts.PullRebase = false
```
EronWright, line 172: Could the force attribute be honored in tag push?

pkg/promotion/runner/builtin/schemas/git-commit-config.json
fuskovic: When you use `git tag -a` , git refers to the config field `user.signingKey` for the GPG key ID. This can be set explicitly using `git config --global user.signingkey YOUR_KEY_ID` however, we don't do that. Instead, the key is sourced implicitly e.g. when git sees that `user.signingKey` is unset, it attempts to look the key up using the format `{{name}} <{{email}}>` which is the format for GPG key uids. With that said, if the `name` and `email` fields don't match the ones in the key uid, the tag will not get signed because the key will not be found; resulting an annotated and unsigned tag. This is more succinctly documented towards the end of `git-tag.md`.

pkg/promotion/runner/builtin/schemas/git-tag-config.json
fuskovic: This is an intentional design decision. Force pushing tags is generally considered a bad idea because tags are intended to be immutable refs. Changing them breaks assumptions many tools and workflows rely on.

pkg/promotion/runner/builtin/git_pusher.go
fuskovic: See https://github.com/akuity/kargo/pull/5785#discussion_r2906930700

pkg/promotion/runner/builtin/schemas/git-commit-config.json
fuskovic: So they are not really _enforced_ per se. They were originally using `git tag -s` (which fails when the key is not found) but @krancour suggested using `git tag -a`; favoring an annotated + unsigned tag over an error/failed-promotion.

In terms of "verified" email addresses this is enforced by goschema validation in the schema config via `"format": "email"`

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-commit.md
```diff
@@ -113,3 +113,40 @@ steps:
...
+      name: Kargo
+      email: kargo@example.com
+      signingKey: ${{ secret('my-gpg-secret').privateKey }}
+```
+
+:::note
+
+If `author.signingKey` is provided but `author.name` and `author.email` do not
+match the email and name in the uid of the associated GPG key, commit tag will 
+not get signed.
```
krancour:
```suggestion
If `author.signingKey` is provided, but `author.name` and `author.email` do not
match the key's UID, the commit will fail.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,107 @@
...
+The `git-tag` step creates a new, annotated tag in a local Git repository
+referencing the current `HEAD` of a checked-out branch.
+
+## Configuration
+
+| Name   | Type     | Required | Description                                                                 |
+|--------|----------|----------|-----------------------------------------------------------------------------|
+| `path` | `string` | Y        | Path to a working directory of a local repository. This path is relative to the temporary workspace that Kargo provisions for use by the promotion process. |
+| `tag`  | `string` | Y        | The tag to create. |
+| `message` | `string` | N | The message to annotate the tag with. If unspecified, defaults to the repo-level configuration specified when the repo was cloned. Can also be configured at the system level. |
```
krancour: This doesn't sound right. I don't think we have repo-level or system-level tag messages.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,107 @@
...
+      name: yourname
+      email: your@inbox.com
+      signingKey: ${{ secret('my-gpg-secret').privateKey }}
+```
+
+:::note
+
+If `tagger.signingKey` is provided but `tagger.name` and `tagger.email` do not
+match the email and name in the uid of the associated GPG key, the tag will not
+get signed.
```
krancour:
```suggestion
If `tagger.signingKey` is provided, but `tagger.name` and `tagger.email` do not
match the key's UID, tagging will fail.
```

pkg/controller/git/base_repo.go
krancour:
```suggestion
			cmd = b.buildGitCommand("config", "--global", "commit.gpgSign", "true")
```

krancour: I think we'd uncovered at some point that this is case insensitive, but this is the way it's documented.

pkg/promotion/runner/builtin/schemas/git-commit-config.json
krancour: > Curious how this requirement came about, how it is enforced, and whether we think these are "verified" email addresses.

It came about organically. From the time Kargo first supported signing with a gpg key (at the time, only a system-level key was supported), we were not setting `user.signingkey`. We _were_ setting `user.name` and `user.email` and importing the gpg key. Given name and email, the correct key can be inferred. If there's a mismatch, the correct key won't be found and signing will fail.

Given that many platforms, even when the signing key is known to them, won't treat a commit as verified if there's a mismatch between the committer info and the signing key, failing _up front_ when these are mismatched really wasn't the worst thing in the world.

> With that said, if the name and email fields don't match the ones in the key uid, the tag will not get signed because the key will not be found; resulting an annotated and unsigned tag.

As far as I can tell, testing directly with `git`, if `tag.gpgSign` is `true` and a key matching the tagger's name and email address isn't found, you don't end up with an annotated and unsigned tag. The whole command fails.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-commit.md
```diff
@@ -113,3 +113,40 @@ steps:
...
+In this example, the `git-commit` step creates a signed commit using the
+provided `author.signingKey` by sourcing it from an existing secret in the same 
+namespace using the [`secret()`](../40-expressions.md#secretname) expression 
+function.
+
+:::note
+
+Author signing information may have been configured at the system level by a 
+Kargo admin. If a system-level configuration exists, the example shown below 
+would override it.
```
krancour, line 128:
```suggestion
Author signing information may have been configured at the system level by a 
Kargo admin. If system-level configuration exists, the example shown below 
would override it.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,107 @@
...
+
+In this example, the `git-tag` step creates a signed tag using the provided 
+`tagger.signingKey` by sourcing it from an existing secret in the same namespace
+using the [`secret()`](../40-expressions.md#secretname) expression function.
+
+:::note
+
+Tagger signing information may have been configured at the system level by a 
+Kargo admin. If a system-level configuration exists, the example shown below 
+would override it.
```
krancour, line 84:
```suggestion
Tagger signing information may have been configured at the system level by a 
Kargo admin. If system-level configuration exists, the example shown below 
would override it.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-commit.md
```diff
@@ -113,3 +113,40 @@ steps:
...
+steps:
+# Update Kustomize manifests, etc...
+- uses: git-commit
+  config:
+    path: ./out
+    message: ${{ outputs['update-image'].commitMessage }}
+    author:
+      name: Kargo
+      email: kargo@example.com
+      signingKey: ${{ secret('my-gpg-secret').privateKey }}
```
krancour:
```suggestion
      name: Me
      email: me@example.com
      signingKey: ${{ secret('my-gpg-secret').privateKey }}
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
```diff
@@ -0,0 +1,107 @@
...
+steps:
+- uses: git-tag
+  config:
+    path: ./out
+    tag: v1.0.0
+    message: legitness
+    tagger:
+      name: yourname
+      email: your@inbox.com
+      signingKey: ${{ secret('my-gpg-secret').privateKey }}
```
krancour:
```suggestion
    message: My example tag
    tagger:
      name: Me
      email: me@example.com
      signingKey: ${{ secret('my-gpg-secret').privateKey }}
```

pkg/promotion/runner/builtin/schemas/git-commit-config.json
krancour:
```suggestion
          "description": "The GPG signing key for the author. If provided, 'email' and 'name' must match the key's UID." 
```

pkg/promotion/runner/builtin/schemas/git-push-config.json
```diff
@@ -7,7 +7,7 @@
   "properties": {
     "generateTargetBranch": {
       "type": "boolean",
-      "description": "Indicates whether to push to a new remote branch. A value of 'true' is mutually exclusive with 'targetBranch'. If neither of these is provided, the target branch will be the currently checked out branch."
+      "description": "Indicates whether to push to a new remote branch. A value of 'true' is mutually exclusive with 'targetBranch' and 'tag'. If neither of these is provided, the target branch will be the currently checked out branch."
```
krancour:
```suggestion
      "description": "Indicates whether to push to a new remote branch. A value of 'true' is mutually exclusive with 'targetBranch' and 'tag'. If none of these are provided, the target branch will be the currently checked out branch."
```

pkg/promotion/runner/builtin/schemas/git-push-config.json
```diff
@@ -31,29 +31,45 @@
     },
     "force": {
       "type": "boolean",
-      "description": "Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). Use with caution as this will overwrite any commits that exist on the remote branch but not in your local branch.",
+      "description": "Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). Use with caution as this will overwrite any commits that exist on the remote branch but not in your local branch. Mutually exclusive with 'tag'.",
       "default": false
+    },
+    "tag": {
+      "type": "string",
+      "description": "A tag to push to the remote repository. Mutually exclusive with 'generateTargetBranch=true', 'targetBranch', and 'force=true'."
```
krancour:
```suggestion
      "description": "A tag to push to the remote repository. Mutually exclusive with 'generateTargetBranch=true' and 'targetBranch'. If none of these are provided, the target branch will be the currently checked out branch. 'tag' is also mutually exclusive with 'force=true'."
```

pkg/promotion/runner/builtin/schemas/git-tag-config.json
```diff
@@ -0,0 +1,44 @@
...
+      }
```
krancour: These are not quite consistent with what the commit step is doing. There, if author is specified, email and name are required attributes. Tagger should behave the same.

```suggestion
      "properties": {
        "email": {
          "type": "string",
          "description": "The email of the tagger.",
          "format": "email"
        },
        "name": {
          "type": "string",
          "description": "The name of the tagger.",
          "minLength": 1
        },
        "signingKey": {
          "type": "string",
          "description": "The GPG signing key for the tagger. If provided, 'email' and 'name' must match the key's UID." 
        }
      },
      "required": ["name", "email"]
```

pkg/promotion/runner/builtin/schemas/git-tag-config.json
```diff
@@ -0,0 +1,44 @@
...
+      "description": "The tag to create in the repository.",
+      "minLength": 1
+    },
+    "path": {
+      "type": "string",
+      "description": "The path to a working directory of a local repository.",
+      "minLength": 1
+    }
+  },
+  "required": ["path", "tag"]
```
krancour: I would make message required because any hard coded default message is going to be about as useful as a hard coded commit message would be.

```suggestion
  "required": ["path", "tag", "message"]
```

pkg/controller/git/work_tree.go
```diff
@@ -304,6 +306,63 @@ func (w *workTree) CreateOrphanedBranch(branch string) error {
...
+	cmd := w.buildGitCommand("tag", "-a", tag, "-m", msg)
+	if homeDir != "" {
+		// Override the home directory set by w.buildGitCommand().
+		w.setCmdHome(cmd, homeDir)
+	}
+	if _, err := libExec.Exec(cmd); err != nil {
+		return fmt.Errorf("error creating signed tag %q", err)
+	}
+	return nil
+}
```
krancour, line 356: See https://github.com/akuity/kargo/pull/5785/changes#r2908744212

```suggestion
// TagOptions represents options for creating a new git tag.
type TagOptions struct {
	// Tagger is the tagger of the tag. If nil, the default tagger already
	// configured in the git repository will be used.
	Tagger *User
}

func (w *workTree) CreateTag(tag, msg string, opts *TagOptions) error {
	if opts == nil {
		opts = &TagOptions{}
	}

	var homeDir string
	// This signing config is specific to this tag, so we will override
	// repository-level signing config by creating a temporary home
	// directory, setting the tag configuration "globally" within it, and
	// then ensuring the git tag command uses that home directory.
	if opts.Tagger != nil {
		var err error
		if homeDir, err = os.MkdirTemp(w.homeDir, ""); err != nil {
			return fmt.Errorf(
				"error creating virtual home directory %q for tag command: %w",
				homeDir, err,
			)
		}
		defer func() {
			if cleanErr := os.RemoveAll(homeDir); cleanErr != nil {
				logging.LoggerFromContext(context.TODO()).
					Error(cleanErr, "error removing virtual home directory", "path", homeDir)
			}
		}()
		if err = w.setupAuthor(homeDir, opts.Tagger); err != nil {
			return fmt.Errorf(
				"error setting up author information for tag command: %w", err,
			)
		}
	}

	cmd := w.buildGitCommand("tag", "-a", tag, "-m", msg)
	if homeDir != "" {
		// Override the home directory set by w.buildGitCommand().
		w.setCmdHome(cmd, homeDir)
	}
	if _, err := libExec.Exec(cmd); err != nil {
		return fmt.Errorf("error creating signed tag %q", err)
	}
	return nil
}
```

pkg/promotion/runner/builtin/schemas/git-clone-config.json
```diff
@@ -20,7 +20,7 @@
     },
     "author": {
       "type": "object",
-      "description": "Default authorship information for any commits made to the cloned repository. If provided, this overrides any system-level defaults. Note: Configuration of the `git-commit` step can override this information.",
+      "description": "Default authorship information for any commits made to the cloned repository. If provided, this overrides any system-level defaults. Note: Configuration of the `git-commit` or `git-tag` step can override this information.",
```
krancour:
```suggestion
      "description": "Default authorship information for any commits made to the cloned repository. If provided, this overrides any system-level defaults. Note: Configuration of the `git-commit` and `git-tag` steps can override this information.",
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-tag.md
krancour: Ok... so there _was_ a hard coded one after all, but see:

https://github.com/akuity/kargo/pull/5785/changes#r2908744212

```suggestion
| `message` | `string` | Y | The message with which to annotate the tag. |
```

pkg/promotion/runner/builtin/schemas/git-commit-config.json
fuskovic: > As far as I can tell, testing directly with git, if tag.gpgSign is true and a key matching the tagger's name and email address isn't found, you don't end up with an annotated and unsigned tag. The whole command fails.

You're right. I was conflating with an end-to-end test I did where the signing key wasn't provided at all.

PR #5787, opened 2026-02-24
fix: add Aggregator field to DatadogMetric stub struct

bafulton:
## What changed and why

The `DatadogMetric` stub struct in `api/stubs/rollouts/v1alpha1/analysis_types.go` was missing the `Aggregator` field that exists in the upstream Argo Rollouts type ([reference](https://github.com/argoproj/argo-rollouts/blob/master/pkg/apis/rollouts/v1alpha1/analysis_types.go)).

When Kargo reads an `AnalysisTemplate` containing `aggregator: sum` and builds an `AnalysisRun` spec, Go silently drops the unknown field during JSON deserialization. The resulting `AnalysisRun` has no `aggregator`, so Argo Rollouts defaults to `aggregator="last"` when calling the Datadog v2 scalar API.

For `as_count()` metrics, `aggregator="last"` returns `null` because it takes only the final bucket of the query window (typically empty for count metrics). This causes Argo Rollouts to record null measurements throughout the `AnalysisRun`, leading to incorrect verification results.

## How it was tested

Confirmed via direct Datadog v2 scalar API calls:
- `aggregator=last` + `as_count()` → `{"values": [null]}`
- `aggregator=sum` + `as_count()` → correct aggregated count

## Notes for reviewers

This is a single-field addition to bring the stub in sync with the upstream Argo Rollouts type. No logic changes.


krancour:
@bafulton you need to re-run codegen after this change.

bafulton:
@krancour, done. 🙇‍♂️

Issue #5788, opened 2026-02-25
Ability to distribute shared secrets across projects


jessesuen:
# Proposed Feature

In v1.9, we introduced "[Shared Resources](https://docs.kargo.io/operator-guide/security/managing-secrets#shared-secrets)", which are `ConfigMap`s and `Secret`s that live in `kargo-shared-resources`.  Promotion steps can reference with the `sharedSecret()` `sharedConfigMap()` helpers.

However, shared secrets do not help with `AnalysisTemplate`s, which do not have the ability to reference a secret outside of its current namespace. It expects the secret to live in the namespace of the AnalysisRun, and may even be mounted by a Job Pod.

For shared secrets to be useful to AnalysisRuns, Kargo should support an option to **_distribute_** a shared secret to one or more namespaces. This would be similar in concept to ESO's `ClusterExternalSecret`.

# Motivation

Currently, (Cluster)AnalysisTemplates cannot make use of a centrally managed Secret. **_Something_** has to perform the distribution to all the Project namespaces.

# Open Issues

What should we do if there is a conflict in the secret name? e.g. a project user created a secret with the same name as the shared/distributed one. In this case, I feel the right thing to do is to preserve the existing secret, and not clobber it.

The UI/API should detect when a secret is managed centrally, and also prevent modifications to it.

jessesuen:
## Implementation Notes

* When a user decides to replicate a secret, they will set the annotation: `kargo.akuity.io/replicate-to: "*"`
* Management-controller monitors secrets in `kargo-shared-resources`. If there are any with this annotation, it:
  * adds a `kargo.akuity.io/replicated` finalizer to the secret (so cleanup can happen)

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-shared-secret
  namespace: kargo-shared-resources
  annotations:
    kargo.akuity.io/replicate-to: "*"
  finalizers:
  - kargo.akuity.io/replicated
data:
  foo: dmFsdWUtMg0KDQo=
```

Once the central secret is marked with the `kargo.akuity.io/replicate-to` annotation, the management-controller starts replicating the secret to all projects. The secret that it creates should be additionally labeled with metadata that allows for efficient detection of out-of-sync versions:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-shared-secret
  namespace: my-project
  label:
    kargo.akuity.io/replicated-from: my-shared-secret
[...]

PR #5789, opened 2026-02-25
feat: option to replicate shared resources to all project namespaces

jessesuen:
Partially resolves https://github.com/akuity/kargo/issues/5788

Adds a new resource replication reconciler to the management controller. `Secrets` & `ConfigMaps` in the `kargo-shared-resources` Namespace, annotated with `kargo.akuity.io/replicate-to: "*"` are automatically copied into every Project namespace. For example:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-shared-secret
  annotations:
    kargo.akuity.io/replicate-to: "*"
data:
  foo: dmFsdWUtMg0KDQo=
```

This enables Argo Rollouts AnalysisTemplates (including Job Pods) to consume them, as it is not possible for cross-namespace secret references.

Key behaviors:
* Source resource gets `kargo.akuity.io/finalizer` finalizer on first reconcile; deletion is blocked until all managed replicas are cleaned up.
* Replicated resources carry `replicated-from` and `replicated-sha` labels; the SHA label is a 16-char truncated SHA-256 of the resource data.
* The `replicated-sha` label is used to detect and optimize when content updates are necessary to be copied to replicated resources. If sha is the same, the update is skipped.
* User-created resources with the same name (no replicated-from label) are left untouched (conflict avoidance).
* Removing the replicate-to annotation triggers the same cleanup path as source secret deletion.
* A Project watch handler re-enqueues all annotated source secrets when a new Project appears, ensuring immediate replication.

Remaining work:
* Documentation
* Mutating Webhook should prevent modifications to replicated resources
* UI checkbox option for replicating a resource

### DIFF
--- a/pkg/controller/management/replication/replication.go
+++ b/pkg/controller/management/replication/replication.go
@@ -423,7 +423,9 @@ func hashMetadata(h hash.Hash, lbls, annotations map[string]string) {
 	h.Write([]byte("annotations"))
 	annotationKeys := make([]string, 0, len(annotations))
 	for k := range annotations {
-		if k != kargoapi.AnnotationKeyReplicateTo && k != lastAppliedConfigAnnotation {
+		if k != kargoapi.AnnotationKeyReplicateTo &&
+			k != kargoapi.AnnotationKeyReplicatedAt &&
+			k != lastAppliedConfigAnnotation {
 			annotationKeys = append(annotationKeys, k)
 		}
 	}


jessesuen:
In addition to the unit tests, this was manually tested end-to-end to cover various corner cases, including:
* adding the annotation after creation
* removing the annotation after creation
* removing the annotation while the controller was down, ensuring it cleaned up when it came back up
* creating a project when controller was down, ensuring Secrets/Configmaps were created when it came back up
* modifying the contents of the Secret/Configmap

api/v1alpha1/annotations.go
```diff
@@ -95,4 +95,15 @@ const (
...
+	// AnnotationKeyReplicateTo is an annotation key that can be set on a resource
+	// in the shared resources namespace to indicate that it should be replicated
+	// to other namespaces. The only supported value is "*", which replicates the
+	// to all Project namespaces.
+	AnnotationKeyReplicateTo = "kargo.akuity.io/replicate-to"
+
+	// AnnotationValueReplicateToAll is the annotation value for
+	// AnnotationKeyReplicateTo that causes a resource to be replicated to all
+	// Project namespaces.
+	AnnotationValueReplicateToAll = "*"
```
jessesuen, line 108: I am considering introducing an annotation, `kargo.akuity.io/replicated-at: 2026-02-25T01:49:00Z` timestamp to indicate when the resource was replicated. This might help end user (or even kargo devs) debug when things changed (e.g., if an admin updated the Secret/ConfigMap from underneath an AnalysisRun).

pkg/controller/management/replication/replication.go
```diff
@@ -0,0 +1,455 @@
...
+	src client.Object,
+	namespace string,
+	sourceHash string,
+	existing client.Object,
+) error {
+	logger := logging.LoggerFromContext(ctx).WithValues("namespace", namespace)
+
+	if existing == nil {
+		// Either no replica exists yet, or it already has the current SHA
+		// (filtered out by the List). Try to create; handle AlreadyExists.
```
EronWright: Is it necessary to call Create in the case where the replicated object exists with a matching SHA? If you were to not filter it, and compare the SHA here (which I see you do anyway later), could the Create call be avoided?

pkg/controller/management/replication/replication.go
```diff
@@ -0,0 +1,455 @@
...
+		return ctrl.Result{}, fmt.Errorf("error listing replicated resources: %w", err)
+	}
+	existingByNamespace := make(map[string]client.Object)
+	for _, item := range r.adapter.getItems(existingList) {
+		existingByNamespace[item.GetNamespace()] = item
+	}
+
+	// Sync to each project namespace.
+	for ns := range projectNamespaces {
+		if err := r.syncToProjectNamespace(ctx, srcObj, ns, sourceHash, existingByNamespace[ns]); err != nil {
```
EronWright, line 227: nit: parallelize

pkg/controller/management/replication/replication.go
```diff
@@ -0,0 +1,455 @@
...
+	// include out-of-date replicas, but check defensively before updating.
+	if existing.GetLabels()[kargoapi.LabelKeyReplicatedSHA] == sourceHash {
+		logger.Debug("replicated resource is already up to date; skipping")
+		return nil
+	}
+
+	// Source changed — update the replica.
+	r.adapter.copyFields(existing, src)
+	existing.SetLabels(replicaLabels(src, sourceHash))
+	existing.SetAnnotations(replicaAnnotations(src))
```
EronWright, line 279: nit: this approach would squash any extra labels or annotations placed directly on the replicated object.  Consider using SSA to easily retain such things.

pkg/controller/management/replication/replication.go
```diff
@@ -0,0 +1,455 @@
...
+	if existing == nil {
+		// Either no replica exists yet, or it already has the current SHA
+		// (filtered out by the List). Try to create; handle AlreadyExists.
+		destObj := r.adapter.newObject()
+		destObj.SetName(src.GetName())
+		destObj.SetNamespace(namespace)
+		destObj.SetLabels(replicaLabels(src, sourceHash))
+		destObj.SetAnnotations(replicaAnnotations(src))
+		r.adapter.copyFields(destObj, src)
+		if err := r.client.Create(ctx, destObj); err != nil {
```
EronWright, line 256: nit: set the `FieldManager`

pkg/controller/management/replication/replication.go
jessesuen: I think you are right about this. I think this logic used to make sense before I added an optimization in the list query. I will revisit this.

pkg/controller/management/replication/replication.go
jessesuen: It was intentional. In fact, a follow-up improvement I plan to make to this is to prevent modifications to replicated objects by end-users (see remaining work I listed in description).

krancour:
First, this looks really good.

I want to note for the record that I looked at how this compares to the existing secret migration reconciler to see if there was any strategic misalignment or opportunities to consolidate. The two have surface-level similarities, but are ultimately different enough that I don't think consolidation is justified.

At this point, I need to validate an assumption. For migration, when a migrated resource is modified, there are no further attempts to sync it to the source. You touch it, it's yours now. I assume that is _not_ how we want this feature to work and that replicas shouldn't drift from the originals.

_If_ my assumption is correct, there's currently a blind spot in the implementation. Drift isn't ever corrected. (The reason it isn't is because drift detection relies on a label that's unlikely to have been updated if/when the replica was modified.) I have specific modifications in mind to account for this, but I'll hold off on offering them until my assumption is validated or challenged.

jessesuen:
I should have called this out more in the design, but you are correct: drift is not corrected for.

The intention is that these are replicated secrets, and users are not supposed to touch them. If they somehow do modify a replicated secret (which I would like to prevent in other ways), then they should understand the replicated secret may not be stable: their modification will get lost sometime in the future (if the source is updated, or we actually decide to implement drift correction).

The lack of drift correction was partly a conscious choice. I tried to optimize the controller to be as lightweight as possible. Although this feature is necessary for some use cases, I don't imagine it will be used heavily. And so I did not want it to incur an ongoing burden on the Kube API or controller resource usage for instances with a lot of projects/secrets.

That said, accounting on the fact that the implementation does not correct drift, my plan was to follow this up with:
* UI and API feature to disable modifications if `replicated-from` is present on the Secret/ConfigMap
* Alternatively, a webhook could do the same, but was hoping not to have to go the webhook route.

krancour:
@jessesuen gotcha.

> which I would like to prevent in other ways

I was already going to suggest that as well.

> * UI and API feature to disable modifications if replicated-from is present on the Secret/ConfigMap

This is what I'd initially been thinking as well, but given that replicated resources would remain mutable via kubectl, it would leave me still wanting to implement some drift correction. I understand the concern about chatiness with the API server and controller resource utilization, but I have a simple approach in mind that doesn't run into either of those problems. Happy to describe it if that possibility interests you.

> * Alternatively, a webhook could do the same, but was hoping not to have to go the webhook route.

I'm not sure why this hadn't already occurred to me. If we took this approach, I think it's actually quite a bit less work and also ensures immutability of replicated resources well-enough that I'd no longer be concerned about drift correction. What were the reasons you were hoping to avoid this approach?

Either way, I agree the UI can easily detect replicas and not present those resources as editable.

jessesuen:
> What were the reasons you were hoping to avoid this approach?

I've gone back and forth on it (you can see in my original description, I actually planned for the webhook to be a follow-up work item. I mistakenly said MWC but meant VWC).

My main concern is my general consternation about webhooks. When webhook servers are down, and the VWC/MWCs are still present, things get hard to clean up (it's a similar gripe I have with finalizers, but to a lesser extent).

Now that I'm thinking about this a little more, I think webhooks are the way to go. There are a few techniques to mitigate my concerns:
* Scope it to only Secrets/ConfigMaps with `replicated-from` label
* Scope the `DELETE` operation policy with `failurePolicy: Ignore` so that things can still be deleted even if the webhook service is down. `UPDATE` operations can remain `failurePolicy: Always`. That way, we'll get the protection we want 99% of the time, but it's not the end of the world if the webhook server can't block a deletion.

So if you feel that webhooks are the right approach, I can follow up this PR with another one that adds webhook protection.

krancour:
Thanks @jessesuen. We're on the same page.

> Scope it to only Secrets/ConfigMaps with replicated-from label

That's exactly what I would have done.

> Scope the DELETE operation policy with failurePolicy: Ignore so that things can still be deleted even if the webhook service is down. UPDATE operations can remain failurePolicy: Always. That way, we'll get the protection we want 99% of the time, but it's not the end of the world if the webhook server can't block a deletion.

Hadn't thought of that and it's a great idea.

> So if you feel that webhooks are the right approach, I can follow up this PR with another one that adds webhook protection.

I'm also happy to follow up on that myself -- unless you specifically _want_ to.

pkg/controller/management/replication/replication.go
krancour: We don't use SSA anywhere (yet), so setting this wouldn't be especially meaningful.

pkg/controller/management/replication/replication.go
krancour: I fixed this in 883946f6d16689da0dd988aaeed98901c75cda4c

pkg/controller/management/replication/replication.go
krancour: That seem like it's possibly a premature optimization. Let's come back to that if we come to believe there's an actual bottleneck here.

api/v1alpha1/annotations.go
krancour: I added it in c7cb658968b0bfcba6f3ad19450110c8fcc31e0e

pkg/controller/management/replication/replication.go
```diff
@@ -0,0 +1,433 @@
...
+	for k := range lbls {
+		if k != kargoapi.LabelKeyReplicatedFrom && k != kargoapi.LabelKeyReplicatedSHA {
+			labelKeys = append(labelKeys, k)
+		}
+	}
+	sort.Strings(labelKeys)
+	for _, k := range labelKeys {
+		h.Write([]byte(k))
+		h.Write([]byte(lbls[k]))
+	}
```
krancour, line 421: There is a possibility of hash collissions here: `a: bc` would hash the same as `ab: c`. `a: b, c: de` would hash the same as `a: bc, d: e`. This could be resolved by inserting separators between keys and values and between k/v pairs as well.

Applies in two other places also.

krancour: Resolved in a98dde495d5a58bc2024d6f7bd51a7bdee6b5410

jessesuen: This function should also be updated to exclude the replicated-at annotation (that was just added) from the computed hash.

krancour:
Opened https://github.com/akuity/kargo/issues/5954 so we don't lose track of the follow-ups we agreed to.

jessesuen:
> I'm also happy to follow up on that myself -- unless you specifically want to.

Oh yes, that'd be wonderful. I haven't had time to revisit this PR, let alone think about the follow up tasks.

pkg/controller/management/replication/replication.go
krancour: @jessesuen double-check me here... we hash _source_ metadata and that new annotation appears on the replicas. Or am I missing something?

krancour: This may be the only thing left. Everything else LGTM.

jessesuen: Oh yes. We do only ever call the hash function on the source.

But if we were ever to run the hash function on both the source and destination for comparison's sake (we don't currently), my recommendation would make this function safe to use for that purpose.

krancour: That seems reasonable.

krancour: Fixed in 40bb87481b587e5f812191c60815cd5cec8386ce

krancour:
This LGTM now, but since the last few commits have my grubby fingerprints all over them, @jessesuen, you might want to double-check that you're still happy. Leave it to you to merge if all is good.

krancour:
The PR is missing docs, but I'll add that to the follow-up issue I already opened.

PR #5795, opened 2026-02-25
fix: revert #5767

krancour:
Further analysis of #5767 shows it to be _introducing_ the issue it purports to resolve.

The key insight into why it was right to begin with is that the `git` CLI is not being used to create test repositories. `gitkit` is being used for this purpose and it always creates default branches named `master`. This introduces a problem that wasn't there before if the unrelated `git config init.defaultBranch` returns anything other than `master`.

If the change in #5767 resolved any difficulties encountered whilst working on #5737, that has to have been coincidental. I believe we'll need to dig deeper for the root cause of any interference with #5737.


EronWright:
## Analysis: This revert reintroduces test failures on macOS with Apple Git

The premise of this revert — that gitkit always creates default branches named `master` — is incorrect. Gitkit calls `git init --bare` under the hood and **inherits the system's git configuration**, including `init.defaultBranch`.

### Root Cause

Apple Git (shipped with Xcode Command Line Tools) sets `init.defaultBranch = main` in its system-level gitconfig:

```
$ cat /Library/Developer/CommandLineTools/usr/share/git-core/gitconfig
[credential]
	helper = osxkeychain
[init]
	defaultBranch = main
```

This means on any macOS system using Apple Git (without a separate homebrew git installation), `git init --bare` creates repos with `main` as the default branch — not `master`.

### Reproducing the failures

```
$ git --version
git version 2.50.1 (Apple Git-155)

$ git config --show-origin init.defaultBranch
file:/Library/Developer/CommandLineTools/usr/share/git-core/gitconfig	main

$ cd /tmp && git clone --depth 1 --branch main https://github.com/akuity/kargo.git
$ cd kargo && go test -v -race -count=1 ./pkg/controller/git/...
```

**5 tests fail:**
- `TestBareRepo` — `fatal: invalid reference: master` when adding a worktree
- `TestBareRepo_SparseCheckout`
- `TestBareRepo_WithFilter`
- `TestRepo`
- `TestWorkTree`

The error is always the same: tests try `git worktree add ... master` but the gitkit-created repo has `main` as its only branch.

### Why CI passes

The CI environment uses `golang:1.26.0-trixie` (Debian) where `init.defaultBranch` is not configured, so git defaults to `master`. This is why the tests pass in CI but fail locally on macOS.

### Why #5767 was correct

The `defaultInitBranch(t)` helper queries the actual git config to determine what branch name `git init --bare` will produce, matching what gitkit actually creates. This makes the tests work correctly regardless of the system's git configuration.

### Suggested fix

Revert this revert (i.e., re-apply #5767), or apply an equivalent fix that makes the tests resilient to different `init.defaultBranch` configurations.

krancour:
Much of the analysis here is correct, but the conclusion is not, because there's a missing piece of the puzzle here.

All of Kargo's git operations occur with `HOME` overridden to ensure configuration that's isolated from your own system configuration. So the helper introduced by #5767 was making an incorrect determination because it was based on _your_ git config and not the isolated config.

PR #5805, opened 2026-02-26
fix(controller): best-effort cleanup of promotion working directories

### LINKED ISSUE #5728: bug: temporary promo dirs on controller fs are orphaned if a running promo is deleted

EronWright:
## Summary

Fixes #5728

- Cleans up temporary working directories (`/tmp/promotion-<UID>`) when promotions reach a terminal state, are aborted, or are deleted
- Uses a shared informer delete handler for immediate deletions (no finalizer needed), and a `DeletionTimestamp` check for deferred deletions when another controller's finalizer keeps the object alive
- Moves workdir cleanup out of the `promote()` defer so it runs after the status patch, ensuring the directory is removed only once the promotion is definitively terminal

## Test plan

- [x] `go test -v -race ./pkg/controller/promotions/...` passes
- [x] `make lint-go` clean (no new issues)
- [x] Deploy to dev cluster and verify working directories are cleaned up after promotions complete
- [x] Verify deletion of a Promotion removes its working directory

### DIFF
--- a/pkg/controller/promotions/promotions.go
+++ b/pkg/controller/promotions/promotions.go
@@ -741,7 +741,11 @@ func setupDeleteCleanup(
 			if !shardPredicate.IsResponsible(meta) {
 				return
 			}
-			cleanupFn(ctx, meta.GetUID())
+			cleanupCtx := logging.ContextWithLogger(ctx, logger.WithValues(
+				"namespace", meta.GetNamespace(),
+				"promotion", meta.GetName(),
+			))
+			cleanupFn(cleanupCtx, meta.GetUID())
 		},
 	})
 	if err != nil {
@@ -779,7 +783,7 @@ func promotionWorkDir(promoUID types.UID) string {
 func (r *reconciler) cleanupWorkDir(ctx context.Context, promoUID types.UID) {
 	workDir := promotionWorkDir(promoUID)
 	logger := logging.LoggerFromContext(ctx)
-	logger.Debug("removing promotion working directory", "uid", promoUID, "path", workDir)
+	logger.Debug("removing promotion working directory", "path", workDir)
 	if err := os.RemoveAll(workDir); err != nil {
 		logger.Error(err, "could not remove working directory", "path", workDir)
 	}


EronWright:
## Manual Test Results

Deployed to local dev cluster and tested both cleanup paths.

### Test 1: Terminal state cleanup
Created a promotion that errored (git auth failure). Controller cleaned up the workdir on terminal state:
```
2026-02-26T18:49:31Z  DEBUG  promotions/promotions.go:786  removing promotion working directory  {"namespace": "kargo-demo", "promotion": "test-cleanup-mcpdx", "path": "/tmp/promotion-695af726-359b-4311-876a-ea61e7601530"}
```

### Test 2: Delete event cleanup
Deleted the same promotion. Informer handler fired and cleaned up (no-op since already removed):
```
2026-02-26T18:49:40Z  DEBUG  promotions/promotions.go:786  removing promotion working directory  {"namespace": "kargo-demo", "promotion": "test-cleanup-mcpdx", "path": "/tmp/promotion-695af726-359b-4311-876a-ea61e7601530"}
```

### Verification
Confirmed no `promotion-*` directories remain on the controller container filesystem:
```
$ kubectl exec -n kargo kargo-controller-... -- ls /tmp/ | grep promotion
(empty)
```

Issue #5799, opened 2026-02-25
Browser titlebar should include project name


jessesuen:
# Proposed Feature

User request: simple request to incorporate the current project name in the browser title bar so that users know which tab to click for which project.

# Motivation

When multiple tabs are open, users don't know which to click

PR #5811, opened 2026-02-27
feat(ui): page titles

Marvin9:
fixes https://github.com/akuity/kargo/issues/5799

### DIFF
--- a/ui/src/features/common/document-title/use-document-title.ts
+++ b/ui/src/features/common/document-title/use-document-title.ts
@@ -13,9 +13,8 @@ import { DocumentTitleContext } from './document-title-context';
 export const useDocumentTitle = (parts: (string | undefined | null | false)[]) => {
   const { appName } = useContext(DocumentTitleContext);
 
-  const filtered = parts.filter(Boolean) as string[];
-
   useEffect(() => {
-    document.title = [...parts, appName].join(' - ');
-  }, [filtered, appName]);
+    const filtered = parts.filter(Boolean) as string[];
+    document.title = [...filtered, appName].join(' - ');
+  }, [parts, appName]);
 };


ui/src/features/common/document-title/document-title-context.ts
rpelczar, line 1: I don't think we need context for this. Left's have a simple hook with useEffect there.

ui/src/features/common/document-title/document-title-context.ts
```diff
@@ -0,0 +1,16 @@
...
+export type DocumentTitleContextType = {
+  /**
+   * Formats an array of title parts into the final document title string.
+   * Parts are ordered from most-specific to least-specific (e.g. ["Stage: prod", "my-project"]).
+   * The default implementation appends "Kargo" and joins with " | ".
+   */
+  formatTitle: (parts: string[]) => string;
+};
+
+const defaultFormatTitle = (parts: string[]) => [...parts, 'Kargo'].join(' | ');
```
rpelczar: Please use `-` instead of `|` as this is a more common separator in the document title.

ui/src/features/common/document-title/use-document-title.ts
```diff
@@ -0,0 +1,22 @@
...
+ * e.g. ["Stage: prod", "my-project"] → "Stage: prod | my-project | Kargo"
+ *
+ * Falsy parts are filtered out, so callers can safely pass undefined/null
+ * params without extra guards.
+ */
+export const useDocumentTitle = (parts: (string | undefined | null | false)[]) => {
+  const { appName } = useContext(DocumentTitleContext);
+
+  const filtered = parts.filter(Boolean) as string[];
+  const key = filtered.join('|');
```
rpelczar: Why is this here?

ui/src/features/common/document-title/use-document-title.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Parts should be ordered from most-specific to least-specific,
+ * e.g. ["Stage: prod", "my-project"] → "Stage: prod | my-project | Kargo"
+ *
+ * Falsy parts are filtered out, so callers can safely pass undefined/null
+ * params without extra guards.
+ */
+export const useDocumentTitle = (parts: (string | undefined | null | false)[]) => {
+  const { appName } = useContext(DocumentTitleContext);
+
+  const filtered = parts.filter(Boolean) as string[];
```
rpelczar: I'm still not sure why you're assigning this here, and don't use it later.

ui/src/features/common/document-title/use-document-title.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  useEffect(() => {
+    document.title = [...parts, appName].join(' - ');
+  }, [filtered, appName]);
```
rpelczar: `filtered` is always a new object instance, so useEffect is called on every re-render. Add console.log and make sure it's not setting in the loop.

rpelczar (approved):

PR #5814, opened 2026-02-27
fix: explicitly set init.defaultBranch in a cloned repo's home dir

krancour:
The root cause of the issue that was impetus for both #5767 and its reversal in #5795 has been identified.

The very abbreviated explanation is that some maintainers were using `git` from homebrew while others were using Apple's `git` binary, which has some small behavioral differences.

The well-intentioned change in #5767 did not account for the fact of all operations in our `git` package being carried out with `HOME` overridden to ensure git configuration that is completely isolated from a contributor's own git configuration.

The well-intentioned reversal restored things to good working order for those who were experiencing no issues to begin with, but re-broke others.

The right fix for the issue is some minor modifications in the `git` package to _explicitly_ set `init.defaultBranch` in the dummy home dir to compensate for behavioral differences in different implementations of `git`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,7 @@ test-unit: install-helm
 				-race \
 				-coverprofile=coverage.txt \
 				-covermode=atomic \
+				-count=1 \
 				./... $(GO_TEST_ARGS); \
 			cd - > /dev/null; \
 		done; \
--- a/pkg/controller/git/base_repo.go
+++ b/pkg/controller/git/base_repo.go
@@ -50,29 +50,50 @@ type ClientOptions struct {
 	InsecureSkipTLSVerify bool
 }
 
+func (b *baseRepo) setupDirs(baseDir string) error {
+	var err error
+	if b.homeDir, err = os.MkdirTemp(baseDir, "repo-"); err != nil {
+		return fmt.Errorf(
+			"error creating home directory for repo %q: %w",
+			b.originalURL, err,
+		)
+	}
+	if b.homeDir, err = filepath.EvalSymlinks(b.homeDir); err != nil {
+		return fmt.Errorf("error resolving symlinks in path %s: %w", b.homeDir, err)
+	}
+	b.dir = filepath.Join(b.homeDir, "repo")
+	cmd := b.buildGitCommand("config", "--global", "init.defaultBranch", "main")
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -140,6 +140,7 @@ test-unit: install-helm
 				-race \
 				-coverprofile=coverage.txt \
 				-covermode=atomic \
+				-count=1 \
```
EronWright, line 143: intentional?

pkg/controller/git/base_repo.go
```diff
@@ -50,29 +50,50 @@ type ClientOptions struct {
...
+		return fmt.Errorf(
+			"error creating home directory for repo %q: %w",
+			b.originalURL, err,
+		)
+	}
+	if b.homeDir, err = filepath.EvalSymlinks(b.homeDir); err != nil {
+		return fmt.Errorf("error resolving symlinks in path %s: %w", b.homeDir, err)
+	}
+	b.dir = filepath.Join(b.homeDir, "repo")
+	cmd := b.buildGitCommand("config", "--global", "init.defaultBranch", "main")
```
EronWright, line 65: This change also affects production behavior — Kargo will now always set `init.defaultBranch = main` when cloning repos. Assumedly none of the use cases (warehouse stuff, `git-clone` step, etc) involve empty git repositories.

Makefile
krancour: Yes. When switching back and forth between the two `git` implementations, tests were sometimes "passing" when they ought not have because the results were cached. Frankly, I'd always meant for it to be this way from the start and it was an oversight that it was not.

pkg/controller/git/base_repo.go
krancour: The possibility of cloning an empty repo as part of a promotion process is one that admittedly hadn't occurred to me, but this should be ok. In theory, you'd never clone the _same_ empty repo more than once. You'd populate it with something on the first promotion and on subsequent promotions, it would no longer be empty. So although this change means an empty repo cloned yesterday would have resulted in creation of a master branch and an empty repo cloned tomorrow would result in the creation of a main branch, that change in behavior would never impact an established promotion process.

EronWright (approved):
LGTM

Issue #5010, opened 2025-09-08
remove `freightMetadata` two-argument form


fykaa:
The `freightMetadata` function currently supports a deprecated two-argument form that needs to be removed in `v1.10` as per the deprecation notice.

**Current Usage**
```go
// Two-form argument (deprecated)
freightMetadata(freightName, 'key-name')
```

**New Usage**
```go
// Single-form argument with map access
freightMetadata(freightName)['key-name']
```

## Tasks
- [ ] remove the two-argument form support from `freightMetadata` function in functions.go
- [ ] update tests in functions_test.go to remove tests for two-argument form
- [ ] ensure documentation is updated to remove mentions of two-argument form
- [ ] optional: add migration notes in the v1.10 release notes

**Additional Context**
The two-argument form has been deprecated since v1.8.0. Users have been encouraged to use the single-argument form with map access syntax instead.

## Related Links
- Documentation: https://docs.kargo.io/user-guide/reference-docs/expressions#freightmetadatafreightname
- deprecated in https://github.com/akuity/kargo/pull/4861/files

PR #5828, opened 2026-03-02
chore: remove deprecated 2 arg form of freightMetadata expr func

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5010

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
+++ b/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
@@ -477,17 +477,6 @@ features.
 
 :::
 
-:::note
-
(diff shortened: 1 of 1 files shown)


pkg/expressions/function/functions.go
```diff
@@ -298,7 +297,7 @@ func freightMetadata(
 	project string,
 ) exprFn {
 	return func(a ...any) (any, error) {
-		if len(a) != 1 && len(a) != 2 {
+		if len(a) != 1 {
 			return nil, fmt.Errorf("expected 1 argument, got %d", len(a))
 		}
```
krancour, line 303: Ha. Well that error message was a bug.

krancour (approved):

Issue #5347, opened 2025-11-05
Freigh promotion stalls in Unknown health state for 3 minutes


henriropp:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [~] I've pasted the output of `kargo version`.
* [-] I've pasted logs, if applicable.

# Description

A freight is being auto promoted to a stage after a build. The health status indicates "Unknown" status for what seems 3 minutes until it goes healthy. There is Analysis Template configured for the stage. The Stage indicates the Verify is running, but it is not. It will start after health status changes to "Healthy".

However, going into a Stages menu and selecting "Refresh" (after 5-10 seconds) will get the health status changed to "Healthy" and the verification starts.

This happen on all (3) stages on all our projects/repos.

# Screenshots

<img width="654" height="336" alt="Image" src="https://github.com/user-attachments/assets/41ec521b-da83-41b3-84fc-fa234431003b" />

# Steps to Reproduce
1. (auto) promote a freight
2. see it happen
3. wait 5-10 seconds
4. stages menu and selecting "Refresh"

# Version

```
1.8.2
```

# Logs

```
```

krancour:
> The health status indicates "Unknown" status for what seems 3 minutes until it goes healthy.

App health (when there is an App involved, that is) is one factor in Stage health. Because of some peculiarities of Argo CD, we are unable to trust an App's self-reported health for some time after a sync operation has completed. We allot 10s as the buffer. Any time a Stage's health checks are executed and we have an unknown result (for any reason) and if we're also able to determine that health should have been know-_able_ (it isn't always), the Stage goes back on the queue with a progressive backoff so that the health check will be retried.

One of two things is happening:

1. You're getting really unlucky and backing off quickly/a lot before the 10s is up and then you end up waiting a while before the next check. And/or you have a lot of Stages and the work queue is pretty deep and that's slowing things down.

1. _Other operations_ are happening against the App (auto-sync? something else?) and Kargo is, as described, not trusting and App's self-reported heath status if its last operation completed fewer than 10s prior.

Or it could be a combination of these two things. What's happening to you isn't unheard of. I've seen it, but there's usually a reason it's happening.

> The Stage indicates the Verify is running, but it is not. It will start after health status changes to "Healthy".

[...]

PR #5832, opened 2026-03-02
feat: add argocd-wait promotion step

### LINKED ISSUE #5400: `argocd-wait` step
# Proposed Feature

`argocd-update` only waits until a sync operation completes. However, in applications that have no PostSync hooks or sync waves, this will return immediately, as the sync is essentially a `kubectl apply` of all manifests.

We need something equivalent to the CLI's:

```
argocd app wait my-app
```

NOTE: Kargo already has a wait for health functionality. This occurs as part of an "implicit verification" process to ensure an app is healthy before running analysis. However, this process may be happening too late for some use cases.

# Motivation

Users often prefer that the app sync and wait happen as part of a _promotion_, instead of verification. Some use cases:
1. In an app that contains multiple `argocd-updates` for different microservices (e.g. backend + frontend) as in the "[grouped services](https://docs.kargo.io/user-guide/patterns/#grouped-services)" pattern, the backend may need to reach health before the frontend is updated

2. Post update steps, like notifications, could be sent only after ensuring apps are healthy after the update.

EronWright:
## Summary

Implements a new \`argocd-wait\` promotion step that blocks until one or more Argo CD Applications reach desired statuses, modeled on \`argocd app wait\` CLI semantics.

- Health/sync/operation checks with CLI-matching defaults and OR'd health flags
- Degradation detection across re-invocations via step output state
- Reliable health status trust using \`reconciledAt\` (see below)
- Label selector support for matching multiple apps
- No authorization annotation required — \`argocd-wait\` is read-only; RBAC is the appropriate gate

### Health status trust via reconciledAt

Argo CD executes operations and assesses health in separate reconciliation loops. Immediately after an operation completes, the health status may reflect state from prior to the operation, creating the risk of an overly-optimistic assessment.

This PR uses \`app.Status.ReconciledAt\` to determine whether health can be trusted: if \`reconciledAt\` is at or after the operation's \`finishedAt\`, health was assessed after the operation completed and can be trusted immediately. Otherwise, a hard refresh is requested (via the \`argocd.argoproj.io/refresh\` annotation) so Argo CD will reconcile and update \`reconciledAt\`; the step returns not-ready.

This is the same approach introduced for the Stage health checker in #5995. The shared \`RequestAppRefresh\` helper is extracted into \`pkg/argocd\` for reuse.

### Reactive Promotion enqueuing

To avoid polling-based delays, the Promotions controller watches Argo CD Application events and reactively enqueues running Promotions when relevant state changes occur. This PR extends the existing predicate chain with two additions:

1. **Indexer fix** (`pkg/indexer/indexer.go`): The `RunningPromotionsByArgoCDApplicationsField` index now covers both `argocd-update` **and** `argocd-wait` steps. Previously, Promotions using only `argocd-wait` were never indexed and could not be reactively enqueued.

2. **`ArgoCDAppReconciledAfterOperation` predicate** (`pkg/controller/promotions/predicates.go`): Fires when `reconciledAt` transitions from stale (nil or before `finishedAt`) to trusted (at or after `finishedAt`). This closes the responsiveness gap identified in review: after `argocd-wait` requests a hard refresh, the Promotion is immediately re-enqueued when Argo CD writes the fresh `reconciledAt` — no polling required.

Design proposal: https://github.com/akuity/kargo/issues/5400#issuecomment-3975214210

Closes #5400

## Test plan

- [x] Unit tests for `argocd-wait` step runner (convert, run, and checkAppReadiness — 13+ cases covering all readiness checks, error conditions, reconciledAt trust, and multi-app scenarios)
- [x] Unit tests for `ArgoCDAppReconciledAfterOperation` predicate (8 cases)
- [x] Unit tests for indexer covering `argocd-wait` step kind
- [x] Schema validation tests
- [x] Existing `argocd-update` tests pass
- [x] Manual testing with Tilt local cluster:
  - Scenario A (`waitFor: [sync]`): app OutOfSync → Promotion Running → ArgoCD sync → `ArgoCDAppSyncChanged` fires → Promotion wakes reactively → Succeeded (~1s latency after sync)
  - Scenario B (default `waitFor`): app OutOfSync → sync triggered → operation completes → `reconciledAt` stale → hard refresh requested → `ArgoCDAppReconciledAfterOperation` fires → Promotion wakes reactively (no health transition) → Succeeded (~0s polling)


EronWright:
## Integration Test Results

Tested the `argocd-wait` step end-to-end against a local Tilt cluster with the `kargo-demo` project (ArgoCD Application backed by a Helm chart).

### Test Scenarios

| Scenario | Result | Notes |
|---|---|---|
| **Full pipeline: argocd-update → argocd-wait** | ✅ Succeeded | Wait step completed in ~10s |
| **Wait on already-healthy app** | ✅ Succeeded | Resolved immediately (no cooldown) |
| **Non-existent app** | ✅ Errored | `unable to find Argo CD Application` |
| **Unauthorized app** | ✅ Errored | `is not authorized` |
| **Unit tests** | ✅ 25 waiter + 14 predicate tests pass | `go test -race` clean |

### Latency Optimization

The initial full-pipeline test took **~29s** for the wait step due to two factors:
1. The 10s health cooldown (operation just finished, health not yet re-assessed)
2. The hardcoded 30s `RetryAfter`

After the optimizations in `651ad85d`:
- `checkAppReadiness` now returns the **remaining cooldown duration** as a `retryAfter` hint
- The `run` method uses `min(appHints, 30s)` instead of always 30s
- New `ArgoCDAppHealthChanged` and `ArgoCDAppSyncChanged` watch predicates enable near-instant requeue on health/sync transitions

Result: wait step now completes in **~10s** (the cooldown period itself) instead of ~29s.

pkg/promotion/runner/builtin/argocd_common.go
```diff
@@ -0,0 +1,108 @@
...
+			appMeta.Namespace,
+			kargoapi.AnnotationKeyAuthorizedStage,
+			allowedStage,
+		)
+	}
+	if projectName != stepCtx.Project || stageName != stepCtx.Stage {
+		return permErr
+	}
+	return nil
+}
```
krancour: This seems to be a refactoring of `authorizeArgoCDAppUpdate()`, which was originally used to assert authorization to mutate an App resource.

I don't think we really require any such check for the sake of querying App resource status.

pkg/promotion/runner/builtin/argocd_updater.go
```diff
@@ -890,90 +888,15 @@ func (a *argocdUpdater) authorizeArgoCDAppUpdate(
...
-		req, err := labels.NewRequirement(expr.Key, op, expr.Values)
-		if err != nil {
-			return nil, fmt.Errorf("invalid matchExpression: %w", err)
-		}
-		labelSelector = labelSelector.Add(*req)
-	}
-
-	return labelSelector, nil
+	return buildArgoCDAppLabelSelector(selector)
 }
```
krancour: Any reason that the caller wouldn't invoke `buildArgoCDAppLabelSelector()` directly?

pkg/health/checker/builtin/argocd.go
```diff
@@ -232,22 +232,24 @@ func (a *argocdChecker) getApplicationHealth(
...
-				appStatus,
-				fmt.Errorf(
-					"last operation of Argo CD Application %q in namespace %q completed "+
-						"less than %s ago; Application health status not trusted",
-					appKey.Name,
-					appKey.Namespace,
-					appHealthCooldownDuration,
-				)
+			if app.Status.Health.LastTransitionTime == nil ||
+				app.Status.Health.LastTransitionTime.Before(app.Status.OperationState.FinishedAt) {
```
krancour: This won't work a lot of the time (if health status doesn't actually change), but this is a great optimization for the cases where it has. 👍

krancour:
I think there might be an important element missing from here.

The handler that enqueues Promotions for reconciliation when a related Application changes in some way relies on a `indexer.RunningPromotionsByArgoCDApplicationsField` index.

Before you look at it, let me warn you that the code that builds that index is probably the most shameful bit of code in all of Kargo because, out of desperation, it does something we really, _really_ don't want to be doing. We've been trending toward all steps not being special in any way, but this index can only be built by inferring relationships between a Promotion and an App by... examining promotion steps, specifically on the lookout for argocd-update steps. 🤮 (It actually gets worse. We have to do a lot of awkward expression evaluation to account for the possibility of expressions having been used in defining _which_ App(s) the step updates.)

Anyway... where I am going with this is that this new step is new and unaccounted for way to establish an association between a Promotion and an App, which means the code that builds that index is going to need to be updated.

If it seems like things are working as desired without having done this and your examples happen to _also_ include an argocd-update step, its presence is the reason things are working. Presumably, many who use the argocd-wait step will _not_ also be using the argocd-update step, so for this feature to be responsive for them, it will need to not rely on an argocd-update step also being present.

krancour:
This PR also needs docs.

pkg/promotion/runner/builtin/argocd_common.go
EronWright: Thanks, that was the key open question - whether to check the authorization annotation.  I'll revert this refactoring.

krancour:
> This PR leverages status.health.lastTransitionTime (available in newer ArgoCD versions per https://github.com/argoproj/argo-cd/pull/21120) to skip the cooldown when ArgoCD has already re-assessed health post-operation. For older ArgoCD versions that don't report lastTransitionTime, the 10-second cooldown remains as a fallback. This optimization is applied to both the existing health checker and the new argocd-wait step.

Trusting the status if `lastTransitionTime` hints that you may is clever, but in the interim, I've found something similar, but even more reliable. It's in #5995, which is trying to solve the frequent complaint of getting stuck for long periods in an unknown health status (#5347).

#5995 removes the cooldown period entirely. Instead of relying on `lastTransitionTime`, which is a newer field, and which only changes when a health status transition has actually occurred, it relies on `reconciledAt`. If `reconciledAt` is after the last operation completed, we can trust that health has been assessed since then _even if there was no change in health and `lastTransitionTime` was never updated as a result._

Even this is imperfect, however, since Argo CD does not necessarily update `reconciledAt` after every reconciliation but, importantly, it _does_ reliably update it after a hard refresh operation.

So the improved App health evaluation in #5995 works like this: Health checker sees `reconciledAt` after the last operation completed? Great. Trust the health status. Done. Otherwise, request a hard refresh. Argo CD _will_ update `reconciledAt` after the hard refresh. An improved App event handler sees that and enqueues related Stages for reconciliation. On next Stage reconciliation, the App health checker sees `reconciledAt` that _is_ after the last operation completed and health status is now trusted.

The result is that App health and Stage health, in turn, are now re-assessed almost instantly after the first case of "unknown health," which also no longer needs to be treated as an error in order to get "free" progressive backoff.

I believe this same strategy would work here and the most complex chunk of it (`appHealthOrSyncStatusChanged()`) could pretty easily be relocated to someplace more reusable.

EronWright:
Thanks for the detailed review @krancour. Here's a summary of what was addressed:

- **Authorization check**: Removed from `argocd-wait` — the step no longer requires the `kargo.akuity.io/authorized-stage` annotation. Also reverted the refactoring of `authorizeArgoCDAppUpdate` so its logic is back inline.
- **`buildLabelSelector` wrapper**: Removed from `argocdUpdater`; `buildLabelSelectorFn` now wires directly to `buildArgoCDAppLabelSelector` in `argocd_common.go`.
- **`reconciledAt`-based health trust**: Incorporated the approach from #5995 — replaced the `lastTransitionTime`-based cooldown with a `reconciledAt` check plus hard refresh, aligning the step with the health checker behavior.
- **Indexer**: Extended `RunningPromotionsByArgoCDApplicationsField` to parse `argocd-wait` steps alongside `argocd-update`, so Promotions using only `argocd-wait` are reactively enqueued when an Application changes.
- **Documentation**: Added `argocd-wait.md` to the promotion steps reference.
- **Test structure**: Added `checkAppReadinessFn` (symmetric with `getApplicationsFn`) and split tests into `Test_argocdWaiter_run` (outer loop only) and `Test_argocdWaiter_checkAppReadiness` (readiness logic, parallel).

krancour:
> reconciledAt-based health trust

There is a gap here still. You are initiating a hard refresh when it's been detected that health status is potentially stale, however, the event handling doesn't appear to have been updated to respond to `reconciledAt` being updated with a value greater than the completion time of the last operation. So while the decision to trust or not is correct, and the hard refresh when health isn't trusted is also correct, nothing is responsive to the benefits of that hard refresh.

If that is addressed, the 30 second suggested requeue interval can be removed. Note that the same suggested requeue interval _also_ exists in the `argocd-update` step's promotion runner, _but should have been removed in #5995 and is my own oversight that it was not._ (I will open a follow-up to #5995 to correct that.)

I had previously proposed that the most expedient way of closing this gap in responsiveness would be to (relocate, export and) re-use the `appHealthOrSyncStatusChanged()` helper function that was introduced in #5995. After more careful analysis, the conditions is responds to and those that the Promotion reconciler need to respond to are slightly, but meaningfully, different.

| # | Signal | Stages needs it? | Stages watches it? | Promotions needs it? | Promotions watches it? |
|:-:|--------|:-:|:-:|:-:|:-:|
| 1 | Health status | Yes | Yes | Yes | Yes |
| 2 | Operation phase | Yes | Yes (any change) | Yes | Yes (only to-completed) |
| 3 | Sync revision | Yes | Yes | No | No |
| 4 | Sync status | No | No | Yes | Yes |
| 5 | reconciledAt/finishedAt crossing | Yes | Yes | Yes | **No** |

Looking at the above, anywhere the first three columns are all yes, there's an opportunity for predicate re-use.

I think the ideal path forward here is to break up `appHealthOrSyncStatusChanged()` and the event handler that depends upon it (which I am happy either to do myself or see done in this PR) into a simpler event handler and four separate predicates. Three of those predicates could easily be used for both use cases.

krancour:
Update: the 30s retry workaround in argocd-update step's runner that should have been removed in #5995 has now been removed in #6033.

pkg/controller/promotions/promotions.go
```diff
@@ -189,6 +189,9 @@ func SetupReconcilerWithManager(
 					ArgoCDAppSyncChanged[*argocd.Application]{
 						logger: logger,
 					},
+					ArgoCDAppReconciledAfterOperation[*argocd.Application]{
+						logger: logger,
+					},
```
krancour, line 196: This closes the responsiveness gap. Implementation of this predicate looks great. I think you can safely remove the aggressive 30s retry now.

Once that's set, this LGTM. I can, separately, do a follow-up to break up #5995's updated `appHealthOrSyncStatusChanged()` to use this predicate and some of your others as well.

krancour (approved):

EronWright:
Added `ArgoCDAppReconciledAfterOperation` to the Promotions Application watch — it fires when `reconciledAt` advances from a stale value (nil or before `finishedAt`) to a newer one, completing the reactive chain: operation completes → hard refresh requested → Argo CD reconciles → predicate fires → Promotion re-evaluated with trusted health.

#6033 prompted removing the 30s `RetryAfter` from `argocd-wait` as well, which is now done.

Filed https://github.com/akuity/kargo/issues/6035 for the follow-up.

pkg/controller/promotions/promotions.go
EronWright: Filed a follow-up: https://github.com/akuity/kargo/issues/6035

EronWright:
## Manual test results

Tested the three reactive scenarios on a local Tilt cluster (branch `EronWright/issue-3762`, post-merge).

---

### Scenario A — sync-transition path (`waitFor: [sync]`)

App was `OutOfSync+Healthy`. Promotion created → Running. ArgoCD sync triggered manually.

```
22:20:09  created → first run: "sync check not passed (OutOfSync)" → Running
22:20:19  enqueued [watches.go:101] ← ArgoCDAppSyncChanged fired
22:20:20  enqueued again → "application is ready" → Succeeded
```

**~1s latency** after ArgoCD updated sync status.

---

### Scenario B — full default waitFor (operation + reconciledAt + sync)

App was `OutOfSync+Healthy`. Promotion created → Running. Sync triggered.

```
22:23:35  created → first run: "sync check not passed (OutOfSync)" → Running
22:23:45  enqueued ← ArgoCDAppOperationCompleted fired
          "application not yet reconciled after last operation, health status not trusted"
          → hard refresh requested → Running
22:23:45  enqueued ← ArgoCDAppReconciledAfterOperation fired
          health=Healthy sync=Synced → Succeeded
```

Health never transitioned — reconciledAt trust was an additional blocker on top of sync.

---

### Scenario C — reconciledAt is the **sole** blocker

App was `Healthy+Synced` throughout (no-op sync, no health or sync transition). Sync triggered (creating a new `finishedAt`) immediately followed by Promotion creation.

```
22:33:38  created → first run: "operation is in progress" → Running
22:33:39  enqueued ← ArgoCDAppOperationCompleted fired
          "application not yet reconciled after last operation, health status not trusted"
          → hard refresh requested → Running
          [health=Healthy, sync=Synced — no change]
22:33:39  enqueued ← ArgoCDAppReconciledAfterOperation fired
          "application is ready" → Succeeded
```

`ArgoCDAppReconciledAfterOperation` was the **only** reason the Promotion advanced on the third run. Without it (and with the 30s `RetryAfter` also gone), the Promotion would have stalled indefinitely waiting for a polling cycle or unrelated event.

Issue #5226, opened 2025-10-16
Git-Open-PR fails when there is no change


kaminski-dev:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

# Description

Current behaviour:

- git-open-pr action fails when there are no changes between the stage branch and Kargo branch
- This causes the promotion pipeline to fail, even though there's nothing to promote 

Desired behaviour:

- When there are no changes, git-open-pr should skip gracefully (not fail)
- git-wait-for-pr should also be skipped in this case
- This mirrors how git-commit behaves—if there's nothing to commit, it just skips

How our stages are configured: 

- Multi-stage pipeline (DEV -> SIT -> UAT -> PROD)
If we have changes which impact the customer we have the following agreement. We just change the values for the specific stage (for example UAT). This will create a new freight which will work until UAT. For PROD the git-open-pr action will fail since it can't create a PR because there are no changes. After few days when rolling out to PROD we will have the same behaviour but in UAT and now we need to manually approve the freight because it fails in UAT. 

The above approach gives us the flexibility to rollout multiple different changes because if we would change all the values at once we still need to wait for a couple of days if there is no customer complaining. This would lead to the situation that co workers of mine would need wait in a queue for their changes to be rolled out. 

The step should do following: 

- Skip if nothing to do

# Screenshots

<img width="1185" height="542" alt="Image" src="https://github.com/user-attachments/assets/3ae50736-4fd5-4871-a966-c654b224c098" />

# Steps to Reproduce

- Create a pipeline with 3 stages (DEV, SIT, PROD)
- For SIT / PROD add the git-open-pr and git-wait-for-pr step
- Change only something in SIT
- New freight is created
- Promote the freight until PROD -- here you will face the issue since no PR can be created
- Change only something in PROD
- New freight is created
- Promote the freight until PROD -- it will fail in SIT since no PR can be created

# Version

```
v1.7.5
```

# Logs

```
[...]

PR #5836, opened 2026-03-03
feat: skip open PR if no changes between src + target

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5226
Closes: https://github.com/akuity/kargo/issues/5847

https://github.com/user-attachments/assets/2b3bec0a-1eb1-4f0e-b45b-3aa33b392898

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
@@ -32,15 +32,6 @@ system to access the git repos.
 | `description` | `string` | N | The description for the pull request. |
 | `labels` | `[]string` | N | Labels to add to the pull request. |
 
-### v1.10
-
-:::warning
-
-Starting with `v1.10`, `createTargetBranch` is a no-op and will be removed in 
-`v1.12`.
-
-:::
-
 ## Output
 
 | Name | Type | Description |


pkg/promotion/runner/builtin/git_pr_opener.go
```diff
@@ -304,28 +332,29 @@ func (g *gitPROpener) getPRID(
 
 // ensureRemoteTargetBranch ensures the existence of a remote branch. If the
 // branch does not exist, an empty orphaned branch is created and pushed to the
-// remote.
+// remote. If the branch already exists, no action is taken. The first boolean return value
+// indicates whether the branch already exists. The second indicates whether it was created.
 func (g *gitPROpener) ensureRemoteTargetBranch(
```
fuskovic: I believe this entire function is unnecessary.  The test for this promotion step originally was testing if a non-existent branch would be created _and_ open a PR against it. This passed locally because the mock git provider is setup to just return the pr details. However, upon end-to-end testing, I discovered that attempting to open a PR against an orphaned branch returns the following error from the API:

```
The <src> branch has no history in common with <target>
```

With that said, I think the right thing to do is:

1. Delete this function and just use `repo.RemoteBranchExists`
2. Update the step to fail if the remote branch doesn't already exist.
3. Remove the `createTargetBranch` field from the config ( since it's essentially useless )

Thoughts @krancour

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
```diff
@@ -97,3 +97,33 @@ steps:
...
+  config:
+    path: ./out
+    generateTargetBranch: true
+  - uses: git-open-pr
+    as: open-pr
+    config:
+      repoURL: https://github.com/example/repo.git
+      sourceBranch: ${{ outputs.push.branch }}
+      targetBranch: stage/${{ ctx.stage }}
+  - if: ${{ status('open-pr') == 'Skipped' }}
```
EronWright: Shouldn't this say "if NOT skipped then wait for PR"?

fuskovic:
@krancour Updated this PR to include the deprecation of `createTargetBranch` in the `git-open-pr` step.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
```diff
@@ -32,6 +32,27 @@ system to access the git repos.
...
+| `repoURL` | `string` | Y | The URL of a remote Git repository. |
+| `provider` | `string` | N | The name of the Git provider to use. Currently `azure`, `bitbucket`, `gitea`, `github`, and `gitlab` are supported. Kargo will try to infer the provider if it is not explicitly specified. |
+| `insecureSkipTLSVerify` | `boolean` | N | Indicates whether to bypass TLS certificate verification when interfacing with the Git provider. Setting this to `true` is highly discouraged in production. |
+| `sourceBranch` | `string` | Y | Specifies the source branch for the pull request. |
+| `targetBranch` | `string` | N | The branch to which the changes should be merged. |
+| `createTargetBranch` | `boolean` | N | **Deprecated**. Is a no-op if set. Will be removed in a future release. |
+| `title` | `string` | N | The title for the pull request. Kargo generates a title based on the commit messages if it is not explicitly specified. |
+| `description` | `string` | N | The description for the pull request. |
+| `labels` | `[]string` | N | Labels to add to the pull request. |
+
```
krancour: I'm guessing we'd not meant to duplicate the whole table?

```suggestion
| `createTargetBranch` | `boolean` | N | **Deprecated**. Is a no-op if set. Will be removed in a future release. |
| `title` | `string` | N | The title for the pull request. Kargo generates a title based on the commit messages if it is not explicitly specified. |
| `description` | `string` | N | The description for the pull request. |
| `labels` | `[]string` | N | Labels to add to the pull request. |

:::warning

Starting with `v1.10`, `createTargetBranch` is a no-op and will be removed in 
`v1.12`.

:::
```

pkg/controller/git/work_tree.go
```diff
@@ -327,6 +329,22 @@ func (w *workTree) DeleteBranch(branch string) error {
...
+func (w *workTree) Fetch() error {
+	// ensure the git config is configured to fetch all branches from the remote, not just the default branch.
+	if _, err := libExec.Exec(w.buildGitCommand(
+		"config",
+		"remote.origin.fetch",
+		"+refs/heads/*:refs/remotes/origin/*",
+	),
+	); err != nil {
+		return fmt.Errorf("error prepping git config to fetch all branches for repo %q: %w", w.originalURL, err)
+	}
```
krancour: I would probably try to do this on the tail end of the cloning process when other settings are also established.

krancour: Actually... disregard. I have a different idea...

krancour: Ok... I have not tested this, but something like this should be generally useful and also allow for some additional optimizations (specify branch and depth = 1 when using from the PR opener step).

```
// FetchOptions represents options for fetching from a remote git repository.
type FetchOptions struct {
	// Branch optionally specifies a single branch to fetch. If empty, all
	// branches are fetched.
	Branch string
	// Depth optionally limits fetching to the specified number of commits. A
	// value of 0 (the default) indicates no depth limit.
	Depth uint
}

func (w *workTree) Fetch(opts *FetchOptions) error {
	if opts == nil {
		opts = &FetchOptions{}
	}
	args := []string{"fetch", "origin"}
	if opts.Branch != "" {
		args = append(args, opts.Branch)
	} else {
		args = append(args, "+refs/heads/*:refs/remotes/origin/*")
	}
	if opts.Depth > 0 {
		args = append(args, "--depth", fmt.Sprint(opts.Depth))
	}
	if _, err := libExec.Exec(w.buildGitCommand(args...)); err != nil {
		return fmt.Errorf(
			"error fetching from repo %q: %w", w.originalURL, err,
		)
	}
	return nil
}
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
krancour, line 1: You may want to update the examples to show conditional execution of a `wait-for-pr` step based on whether this step was skipped (due to no diffs) or not:

https://docs.kargo.io/user-guide/reference-docs/promotion-templates#conditional-steps

pkg/promotion/runner/builtin/git_pr_opener.go
```diff
@@ -192,20 +190,41 @@ func (g *gitPROpener) run(
...
-			"error ensuring existence of remote branch %s: %w",
-			cfg.TargetBranch, err,
+			"target branch %q does not exist", cfg.TargetBranch,
 		)
 	}
 
+	// Ensure we have the latest commits from the remote before checking for diffs.
+	if err = repo.Fetch(); err != nil {
+		return promotion.StepResult{Status: kargoapi.PromotionStepStatusErrored}, err
+	}
```
krancour, line 215: I think if we do this: https://github.com/akuity/kargo/pull/5836/changes#r2913651991

Then we can fetch just the target branch here, with depth 1. That should be a substantial performance improvement for large repos.

krancour:
Apart from a few small optimizations, this looks great!

pkg/controller/git/work_tree.go
fuskovic: Doesn't work as is but if we first update the git config with:

```

	libExec.Exec(w.buildGitCommand(
		"config",
		"remote.origin.fetch",
		"+refs/heads/*:refs/remotes/origin/*",
	)
	
```


Then it works.


I did that within the `Fetch` method itself. I can try to update the git config towards the tail-end of `git-clone` step like you originally suggested and then we can nix the `else` clause in `Fetch`.

pkg/controller/git/base_repo.go
```diff
@@ -102,6 +102,19 @@ func (b *baseRepo) setupClient(opts *ClientOptions) error {
 	return nil
 }
 
+func (b *baseRepo) setupFetch() error {
+	// ensure the git config is configured to fetch all branches from the remote, not just the default branch.
+	if _, err := libExec.Exec(b.buildGitCommand(
+		"config",
+		"remote.origin.fetch",
+		"+refs/heads/*:refs/remotes/origin/*",
```
krancour: Can't this (or something similar) be specified as an argument to `git fetch` when no single branch was specified by the fetch options? I had thought that was possible, which would make this setup unnecessary.

pkg/controller/git/work_tree.go
fuskovic: Disregard

pkg/controller/git/base_repo.go
fuskovic: So I just tested and this. If a target branch is not specified it works without the config setup. However, if the target branch is specified, fetching it fails. Good news is I was able to fix that by just using the same pattern for the target branch e.g. when target branch is specified the command uses `+refs/heads/<target-branch>:refs/remotes/origin/<target-branch>"` and that works without the config setup.

krancour: Oh right... that makes sense. If we did a single branch and/or shallow clone, we'd end up needing the full refspec even for fetching a single branch.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
krancour: Didn't we mean to update this table row?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
```diff
@@ -54,22 +63,56 @@ This is a common pattern when implementing GitOps-based promotion workflows,
...
+    config:
+      repoURL: ${{ vars.gitRepo }}
+      sourceBranch: ${{ outputs.push.branch }}
+      targetBranch: ${{ vars.targetBranch }}
+  - if: ${{ status('open-pr') != 'Skipped' }}
+    uses: git-wait-for-pr
+    as: wait-for-pr
+    config:
+      repoURL: ${{ vars.gitRepo }}
+      prNumber: ${{ outputs['open-pr'].pr.id }}
```
krancour: This example is strange because what it does is deletes everything from the target branch.

I understand your motivation here was removing `createTargetBranch`, which never worked to begin with from the example, but then feeling the example was incomplete. You felt compelled to still show how target branch might originally have been created? Something like that?

imho, the original example (minus the `createTargetBranch` line) is fine as is.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
```diff
@@ -94,11 +137,47 @@ steps:
...
+    prNumber: ${{ outputs['open-pr'].pr.id }}
 ```
+
+### Skipped
+
+The following example conditionally runs the 
+[`git-wait-for-pr` step](git-wait-for-pr.md) based on whether or not the 
+`git-open-pr` step was skipped. If there are no changes between the 
+`sourceBranch` and `targetBranch`, and the `targetBranch` already exists, the 
+`git-open-pr` step will be skipped. The
```
krancour: If the target branch didn't exist on the remote, the step would actually fail, wouldn't it?

So maybe simplify this statement to only mentioning the skip if there's no difference.

fernandomendieta-facephi:
May I know if this will be released in the next version?
It will help us a lot!!

Thanks for the work

fuskovic:
> May I know if this will be released in the next version? It will help us a lot!!
> 
> Thanks for the work

Yes, this work is targeting `v1.10`.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-open-pr.md
```diff
@@ -27,11 +27,20 @@ system to access the git repos.
...
 | `sourceBranch` | `string` | Y | Specifies the source branch for the pull request. |
 | `targetBranch` | `string` | N | The branch to which the changes should be merged. |
-| `createTargetBranch` | `boolean` | N | Indicates whether a new, empty orphaned branch should be created and pushed to the remote if the target branch does not already exist there. Default is `false`. |
+| `createTargetBranch` | `boolean` | N | **Deprecated**. Is a no-op if set. Will be removed in a future release.|
 | `title` | `string` | N | The title for the pull request. Kargo generates a title based on the commit messages if it is not explicitly specified. |
 | `description` | `string` | N | The description for the pull request. |
 | `labels` | `[]string` | N | Labels to add to the pull request. |
 
+### v1.10
+
```
krancour: Not sure this warning deserves its own section.

Other than this, LGTM

Issue #5732, opened 2026-02-12
Support commit signing for GitHub Apps


eddiewebb:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

While #1843 allows individual users and machine accounts to sign commits, it does not support GitHub Apps.

GitHub Apps appear to either need an email address not supported by JSON Schema validation (use of []) or require an explicit dance via multiple API calls.

- https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#signature-verification-for-bots claims proper email and user info should work, but JSON Schema restricts the format of GH App Email needed (`app-user-id+your-app-name[bot]@users.noreply.github.com`)
- https://github.com/orgs/community/discussions/50055 and https://josh-ops.com/posts/github-apps-commit-email/ claim that an explicit API dance to create tree, commit, and reference is needed.

# Motivation

Organizations are increasingly adopting GitHub Apps, which is their stated go-forward recommendation for integrations.

Additionally many organizations mandate the use of commit signing.

Today these two are irreconcilable.

# Additional Info

Kargo v1.9.1

While use of the email with brackets is prevented, I did find that at least using the email without bot brackets  (`app-user-id+your-app-name@users.noreply.github.com`) allows GH to correspond commits to the GH App, but does not consider them verified.

The default Kargo user provided does not associate or show in user history.

<img width="989" height="393" alt="Image" src="https://github.com/user-attachments/assets/6c5ba5ec-9fd4-4f95-8032-379534350994" />
<img width="1096" height="201" alt="Image" src="https://github.com/user-attachments/assets/9b01d614-2193-4d44-8479-44880f669ced" />
<img width="371" height="206" alt="Image" src="https://github.com/user-attachments/assets/b222a4e3-9b0a-4269-9e0f-db0e61a99e1f" />

jessesuen:
Regarding the email that github uses to author commits:
* GitHub wants to author commits with an email that looks like: `1234567+xxxxxxxx[bot]@users.noreply.github.com`
* Due to the square brackets, the above email is only considered valid in RFC 5322, **_but not RFC 5321_**
* Kargo uses the current/latest [json schema specification](https://json-schema.org/draft/2020-12/json-schema-validation#name-email-addresses) to validate emails
[...]

PR #5837, opened 2026-03-03
feat: add github-sign promotion step

EronWright:
## Summary
- Adds a new `github-sign` builtin promotion step that signs commits on a GitHub branch by replaying them via the GitHub REST API
- Enables GitHub App commit verification (the "Verified" badge) for commits created by Kargo's git-commit/git-push steps
- Uses a **staging branch pattern**: unsigned commits are pushed to a temporary branch (via `generateTargetBranch`), then replayed as signed commits and fast-forward updated onto the target branch — compatible with branch protection rules that disallow force pushes

image

## How It Works

GitHub Apps cannot produce verified (signed) commits via `git commit` + `git push`. GitHub only signs commits created through its REST API, authenticated with an App installation token, with no custom author/committer overrides.

This step bridges the gap by replaying commits that were pushed (unsigned) to a staging branch:

```
Staging branch:  base → R1 → R2 → R3  (unsigned, pushed via git CLI)

Target branch:   base → R1' → R2' → R3'  (signed by GitHub App, fast-forward)
```

Each replayed commit preserves the original tree (snapshot) and commit message. Author/committer are set by GitHub to the App's bot user (e.g. `my-app[bot]`).

**API flow**: Compare (`base...head`) → Create signed commits (one per revision) → Fast-forward update target branch ref.

### Staging branch pattern

The recommended workflow uses `git-push` with `generateTargetBranch: true` to push unsigned commits to a temporary branch (e.g. `kargo/promotion/<name>`), then `github-sign` replays them as signed commits onto the real target branch (e.g. `stage/test` or `main`).

Since the target branch never moved (it still points to `base`), updating it to the final signed commit is a fast-forward — no force push required. This is compatible with branch protection rules that disallow force pushes and history rewriting.

### Design: concrete SHAs over ref names

The step accepts concrete commit SHAs for both `base` and `head` rather than branch names. This eliminates a race condition where GitHub's Compare API could return stale results when resolving a branch ref immediately after `git-push`. By passing SHAs directly (from `git-clone` and `git-push` outputs), the comparison is fully deterministic.

The `targetBranch` field is separate — it specifies which branch ref to update after signing, decoupled from the comparison range.

## Example

A typical promotion pipeline that clones a source branch and a stage branch, renders changes, pushes to a staging branch, and then signs the result onto the target branch:

```yaml
steps:
- as: clone
  uses: git-clone
  config:
    repoURL: https://github.com/example/gitops-repo
    checkout:
    - branch: main
      path: ./src
    - branch: stage/test
      create: true
      path: ./out

# ... render changes into ./out (kustomize, helm, etc.) ...

- as: commit
  uses: git-commit
  config:
    message: ${{ ctx.promotionInfo }}
    path: ./out

- as: push
  uses: git-push
  config:
    path: ./out
    generateTargetBranch: true  # push to kargo/promotion/<name>

- as: sign
  uses: github-sign
  config:
    repoURL: https://github.com/example/gitops-repo
    targetBranch: stage/test
    base: ${{ outputs.clone.commits["./out"] }}
    head: ${{ outputs.push.commit }}

- uses: argocd-update
  config:
    apps:
    - name: my-app
      sources:
      - repoURL: https://github.com/example/gitops-repo
        desiredRevision: ${{ outputs.sign.commit }}
```

The `git-push` step pushes unsigned commits to a generated staging branch. The `github-sign` step then replays the range `base..head` as signed commits and fast-forward updates the target branch. The `base` is the stage branch HEAD at clone time (from `git-clone` output, keyed by checkout path). The `head` is the commit SHA that was pushed (from `git-push` output). The `argocd-update` step uses the signed commit SHA from the `sign` step output.

**Prerequisites**: A GitHub App with **Contents: read & write** permission, installed on the target repository, with credentials stored as a Kargo Git credential secret (using `githubAppClientID`, `githubAppPrivateKey`, and `githubAppInstallationID` fields).

## Config

| Field | Required | Description |
|-------|----------|-------------|
| `repoURL` | Yes | GitHub repository URL |
| `targetBranch` | Yes | Branch whose ref will be updated to the final signed commit |
| `head` | Yes | Inclusive head of the revision range (commit SHA, typically from `git-push` output) |
| `base` | Yes | Exclusive base of the revision range (commit SHA, typically from `git-clone` output) |
| `insecureSkipTLSVerify` | No | Skip TLS verification for GitHub Enterprise |

The maximum number of revisions to sign is controlled by the `GITHUB_SIGN_MAX_REVISIONS` environment variable (default: 10).

### Output

| Field | Description |
|-------|-------------|
| `commit` | SHA of the final signed commit |
| `commitURL` | Web URL to the signed commit on GitHub |
| `branch` | The target branch that was updated |

## Test plan
- [x] Unit tests (24 cases): schema validation, revision range signing, idempotency, URL building, client creation
- [x] Manual test: promotions succeeded with signed commits in a local Kind cluster against a real GitHub repo using the staging branch pattern
- [x] Verified "Verified" badge on signed commits on GitHub

Refs: #5732

### DIFF
--- a/pkg/x/promotion/runner/builtin/zz_config_types.go
+++ b/pkg/x/promotion/runner/builtin/zz_config_types.go
@@ -308,7 +308,7 @@ type GitHubSignConfig struct {
 	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
 	// The URL of the GitHub repository.
 	RepoURL string `json:"repoURL"`
-	// The branch whose ref will be force-updated to point to the final signed commit.
+	// The branch whose ref will be updated to point to the final signed commit.
 	TargetBranch string `json:"targetBranch"`
 }
 
--- a/ui/src/gen/directives/github-sign-config.json
+++ b/ui/src/gen/directives/github-sign-config.json
@@ -11,7 +11,7 @@
   },
   "targetBranch": {
    "type": "string",
-   "description": "The branch whose ref will be force-updated to point to the final signed commit.",
+   "description": "The branch whose ref will be updated to point to the final signed commit.",
    "minLength": 1
   },
   "head": {


EronWright:
## Manual test results

Ran 5 consecutive promotions against a real GitHub repo (`EronWright/kargo-demo`) in a local Kind cluster. Each run: reset branch to initial commit → create promotion → verify all steps succeed.

**Result: 5/5 succeeded**, sign step passed every time with consistent ~1s execution time.

```
NAME              STAGE   FREIGHT                                    PHASE       AGE
test-sign-25jmq   test    5fea7f4302f02d87d304bdcd7f2494032e347a43   Succeeded   5s
test-sign-7bmd8   test    5fea7f4302f02d87d304bdcd7f2494032e347a43   Succeeded   5s
test-sign-rmsjr   test    5fea7f4302f02d87d304bdcd7f2494032e347a43   Succeeded   5s
test-sign-c7khf   test    5fea7f4302f02d87d304bdcd7f2494032e347a43   Succeeded   5s
test-sign-tmk8n   test    5fea7f4302f02d87d304bdcd7f2494032e347a43   Succeeded   5s
```

Example signed commit: [`cee7d646`](https://github.com/EronWright/kargo-demo/commit/cee7d64605de6cf38caac57da8ee77aa9429b908)
- Author: `kargo-demo-signer-eronwright[bot]`
- Committer: `GitHub`
- Verified: ✅

### Race condition fix

Earlier testing revealed a race condition when using a branch ref name in the Compare API call — GitHub's ref resolution could be briefly stale after `git-push`, causing the Compare to return "identical" when commits existed. This was fixed by accepting concrete commit SHAs (`base` and `head`) instead of resolving branch names. The comparison is now fully deterministic.

jessesuen:
Thanks. Now that I'm seeing how this works, I'm trying to understand how users will use this in practice. 

Assumptions:
1. Users require signed commits in `main` (the whole point of this PR)
2. Users do not allow history rewrite on `main` (I think this is standard practice, and actually enabled by default when adding a branch protection rule)

So the force-update behavior of `base...branch` doesn't seem practical unless it's in a non-main branch. But if it's in a non-main branch, a PR would anyway create a verified commit as part of the merge.

Basically, I'm questioning if replaying with force update is the right thing to do, as opposed to replaying from a temporary, staged branch that is used only as a reference of commit(s) that need to be replayed+verified to a target branch (i.e. `main`), and subsequently discarded.

The implementation in this PR would not support a repo configured with a branch protection rule like this:

image

EronWright:
Updated per feedback: the `github-sign` step now uses a **non-force (fast-forward) ref update** instead of force-updating the target branch. This makes it compatible with branch protection rules that disallow force pushes.

The recommended workflow uses a staging branch pattern:
1. `git-push` with `generateTargetBranch: true` pushes unsigned commits to a temp branch
2. `github-sign` replays them as signed commits and fast-forward updates the real target branch

Tested successfully in a local Kind cluster against `EronWright/kargo-demo` with the staging branch pattern.

EronWright:
Closing in favor of: https://github.com/akuity/kargo/pull/5840

PR #5839, opened 2026-03-03
fix: restore credential resolution for deprecated globalCredentials.namespaces

### LINKED ISSUE #5838: Warehouse git subscriptions fail with dial-out errors after upgrade from v1.8.9 to v1.9.x
We've attempted upgrading from v1.8.9 to v1.9.x three times now — each time warehouses with git subscriptions break. Rolling back to v1.8.9 immediately resolves the issue.

## Timeline

| Attempt | Version | Date | Result |
|---------|---------|------|--------|
| 1 | v1.9.0 | 2026-01-29 | Warehouses broken (subscription nodes hidden — #5663) |
| 2 | v1.9.2 | 2026-02-04 | Warehouses broken (workflows failed, likely git-clone --filter regression) |
| 3 | v1.9.5 | 2026-03-03 | Warehouses broken — GitHub sync / dial-out errors |

## Symptoms (v1.9.5, most recent)

- Warehouse settings page shows broken state
- GitHub sync / dial-out errors from warehouse controller
- All warehouses use **git-only subscriptions** — no image or chart subscriptions
- Reverting to v1.8.9 immediately fixes it every time

## Environment

- **Platform:** GKE Autopilot
- **Deployment:** ArgoCD Application (multi-source) with Helm chart from `ghcr.io/akuity/kargo-charts`
- **Upgrade path:** v1.8.9 → v1.9.x (no intermediate versions deployed)

## Warehouse Configuration

All warehouses follow this pattern (git-only, no images):

```yaml
kind: Warehouse
metadata:
  name: api-gateway-warehouse
  namespace: api-gateway
spec:
  interval: 5m0s
  subscriptions:
  - git:
      repoURL: https://github.com/translucent-ai/kubernetes-resources.git
      branch: main
      includePaths:
      - app/dev/us-central1/api-gateway
      - app/nonprod/us-central1/api-gateway
      - app/prod/us-central1/api-gateway
```

## Helm Values (relevant sections)

```yaml
controller:
  globalCredentials:
    namespaces:
      - kargo-cluster-secrets
  reconcilers:
    warehouses:
      maxConcurrentReconciles: 8
    stages:
      maxConcurrentReconciles: 8
    promotions:
      maxConcurrentReconciles: 8
```

## Expected Behavior

Warehouses should continue to sync git commits from GitHub after upgrading to v1.9.x, same as v1.8.9.

## Actual Behavior

Warehouses fail with dial-out / sync errors when connecting to GitHub. The Kargo UI shows the warehouse in a broken state. Rolling back to v1.8.9 restores normal operation immediately.

## Notes

- We suspect this may be related to the git-clone `--filter` change that was introduced and then reverted in v1.9.1 (#5679), but the issue persists through v1.9.5
- The `globalCredentials.namespaces` config uses the pre-v1.9.0 terminology — unclear if the "Secret Shuffle" migration is causing credential resolution failures for git operations
- We have ~20+ projects all using the same warehouse pattern, all affected simultaneously
- Controller logs from the v1.9.5 window may still be available — will attach if recoverable

markshields-tl:
Fixes #5838

## Problem

In v1.9.0 the "Secret Shuffle" renamed `controller.globalCredentials.namespaces` (`GLOBAL_CREDENTIALS_NAMESPACES`) to `global.sharedResources.namespace` (`SHARED_RESOURCES_NAMESPACE`). The management-controller ConfigMap template received a backward-compat shim for the old value, but the **controller ConfigMap template did not**.

Operators who had not yet migrated their Helm values silently got `SHARED_RESOURCES_NAMESPACE=kargo-shared-resources` (the new default) instead of their actual credentials namespace. This caused all warehouse git subscriptions to fail with dial-out errors because credentials could no longer be found.

We hit this three times in production upgrading from v1.8.9 to v1.9.0, v1.9.2, and v1.9.5 — each time requiring an emergency rollback.

## Fix

Two layers:

1. **Helm chart** (`charts/kargo/templates/controller/configmap.yaml`): When `controller.globalCredentials.namespaces` is set, use its first element as `SHARED_RESOURCES_NAMESPACE` — same shim already present in the management-controller template.

2. **Go code** (`pkg/credentials/kubernetes/database.go`): `DatabaseConfigFromEnv()` now falls back to the first element of the legacy `GLOBAL_CREDENTIALS_NAMESPACES` env var when `SHARED_RESOURCES_NAMESPACE` is empty, providing defense-in-depth for deployment paths that bypass the Helm chart.

## Tests

Added `TestDatabaseConfigFromEnv` covering:
- New env var takes precedence when set
- Falls back to first element of legacy env var
- Empty config when neither is set

### DIFF
--- a/charts/kargo/templates/controller/configmap.yaml
+++ b/charts/kargo/templates/controller/configmap.yaml
@@ -23,7 +23,11 @@ data:
   {{- if .Values.kubeconfigSecrets.kargo }}
   KUBECONFIG: /etc/kargo/kubeconfigs/kubeconfig.yaml
   {{- end }}
+  {{- if .Values.controller.globalCredentials.namespaces }}
+  SHARED_RESOURCES_NAMESPACE: {{ first .Values.controller.globalCredentials.namespaces | quote }}
+  {{- else }}
   SHARED_RESOURCES_NAMESPACE: {{ quote .Values.global.sharedResources.namespace }}
+  {{- end }}
   ALLOW_CREDENTIALS_OVER_HTTP: {{ quote .Values.controller.allowCredentialsOverHTTP }}
   GITCLIENT_NAME: {{ quote .Values.controller.gitClient.name }}
   GITCLIENT_EMAIL: {{ quote .Values.controller.gitClient.email }}
--- a/pkg/credentials/kubernetes/database.go
+++ b/pkg/credentials/kubernetes/database.go
@@ -19,6 +19,17 @@ import (
 	"github.com/akuity/kargo/pkg/urls"
 )
 
+// legacyConfig holds deprecated environment variables for backward compatibility.
+// These were renamed in v1.9.0 as part of the "Secret Shuffle" refactor.
+type legacyConfig struct {
+	// GlobalCredentialsNamespaces was renamed to SharedResourcesNamespace in
+	// v1.9.0. When SHARED_RESOURCES_NAMESPACE is unset, the first element of
+	// this slice is used as a fallback so that operators who have not yet
+	// migrated their Helm values (controller.globalCredentials.namespaces) do
+	// not silently lose credential resolution.
+	GlobalCredentialsNamespaces []string `envconfig:"GLOBAL_CREDENTIALS_NAMESPACES" default:""`
+}
+
 // database is an implementation of the credentials.Database interface that
 // utilizes a Kubernetes controller runtime client to retrieve credentials
 // stored in Kubernetes Secrets.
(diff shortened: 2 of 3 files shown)


krancour:
I think there may be some confusion over how this was intended to work. There's a reason the controller wasn't configured like the management controller.

What we deliberately aimed to do was transition all of Kargo to exclusively using the newer settings, except for the management controller, which would be the one component aware of both old and new settings and would perform secret migration. If the management controller is doing its job, all other components should be fine.

I'm not doubting that you're encountering a real issue, but I don't believe this is the correct resolution.

Is the management controller, perhaps, _not_ migrating your secrets as it was intended to?

markshields-tl:
> I think there may be some confusion over how this was intended to work. There's a reason the controller wasn't configured like the management controller.
> 
> What we deliberately aimed to do was transition all of Kargo to exclusively using the newer settings, except for the management controller, which would be the one component aware of both old and new settings and would perform secret migration. If the management controller is doing its job, all other components should be fine.
> 
> I'm not doubting that you're encountering a real issue, but I don't believe this is the correct resolution.
> 
> Is the management controller, perhaps, _not_ migrating your secrets as it was intended to?

Probably not.

PR #5840, opened 2026-03-04
feat: add github-verified-push promotion step

### LINKED ISSUE #5732: Support commit signing for GitHub Apps

EronWright:
## Summary

Adds `github-verified-push`, a new promotion step that is a drop-in replacement for `git-push` when commits need GitHub&#39;s &#34;Verified&#34; badge. This is essential for organizations that use **GitHub App installation tokens** for authentication and enforce **commit verification** via branch protection rules.

Closes #5732

## Motivation

Organizations increasingly adopt GitHub Apps as their go-forward recommendation for integrations. Many also mandate commit signing via branch protection rules. Today these two requirements are irreconcilable in Kargo — GitHub Apps cannot GPG-sign commits through `git push`, and the email format required for App attribution (`id+name[bot]@users.noreply.github.com`) is incompatible with standard commit signing.

The GitHub API offers a path: commits created via the REST API with an App installation token are automatically signed by GitHub and marked as &#34;Verified&#34; — no GPG key management on the GitHub side required.

## How It Works

The step replays local commits through the GitHub REST API instead of using `git push`:

1. **Push to staging ref** — Local commits are pushed via `git push` to `refs/kargo/staging/`, a non-branch ref invisible in GitHub&#39;s branch UI.
1. **Compare** — The GitHub Compare API enumerates commits between the target branch HEAD and the staging ref.
1. **Verify signatures** — Each local commit&#39;s GPG signature is checked against the controller&#39;s keyring to determine whether Kargo vouches for it.
1. **Replay** — Each commit is recreated via the GitHub CreateCommit API:
   - **System-key-signed commits** (signed by the controller&#39;s `GITCLIENT_SIGNING_KEY_PATH` key with Ultimate trust) → author/committer omitted → GitHub App signs → **Verified** badge. If the original author differs from the Kargo default, a `Co-authored-by` trailer is added.
   - **Unsigned commits, commits signed by untrusted keys, or commits signed by unrecognized keys** → original author/committer preserved → replayed as-is (not verified).
   - **Bad or revoked signatures** → terminal error (potential tampering).
1. **Fast-forward** — The target branch is updated to the final replayed commit via UpdateRef (or force-updated when `force: true`).
1. **Cleanup** — The staging ref is deleted.

The output contract matches `git-push` (`branch`, `commit`, `commitURL`), so downstream steps like `git-open-pr` work unchanged.

### Conditional signing via GPG trust

Rather than unconditionally re-signing all commits through the GitHub App, the step uses the controller&#39;s GPG signing key as a trust signal. Only commits signed by the **system-level signing key** (configured via `GITCLIENT_SIGNING_KEY_PATH`, which receives Ultimate trust) are eligible for GitHub App verification. This preserves the intent of unsigned or externally-signed commits.

Per-commit author GPG keys (provided via the `git-commit` step&#39;s `author.signingKey`) are imported into the repo-level keyring **without trust assignment**. Their commits report GPG status `U` (untrusted) and are propagated as-is without re-signing. A future whitelist/trusted-keys feature will allow administrators to designate additional keys as trusted.

When `GITCLIENT_SIGNING_KEY_PATH` is **not** configured, all commits preserve their original authorship and none receive the Verified badge.

### Force push support

The `force: true` option enables scenarios like pushing fully-rendered manifests that replace previous branch content entirely, matching the equivalent `git-push` capability.

## Example: Mixed Commit Authors

A single promotion with three commits exercising all signing paths:

```yaml
steps:
- uses: git-clone
  config:
    checkout:
    - branch: main
      path: ./src
    - branch: stage/test
      path: ./out
- uses: git-clear
  config:
    path: ./out
- uses: kustomize-set-image
  config:
    path: ./src/base
    images:
    - image: public.ecr.aws/nginx/nginx
      newName: public.ecr.aws/nginx/nginx
      tag: "1.29.5"
- uses: kustomize-build
  config:
    path: ./src/stages/test
    outPath: ./out

# Commit 1: Default Kargo author (GPG-signed by system key)
- uses: git-commit
  config:
    path: ./out
    message: "render manifests for 1.29.5"

- uses: copy
  config:
    inPath: ./src/base/service.yaml
    outPath: ./out/extra-service.yaml

# Commit 2: External contributor WITH signing key (untrusted)
- uses: git-commit
  config:
    path: ./out
    message: "add extra service file"
    author:
      name: External Contributor
      email: external@example.com
      signingKey: ${{ secret("external-contributor-signing-key").signingKey }}

- uses: copy
  config:
    inPath: ./src/base/kustomization.yaml
    outPath: ./out/stranger-file.yaml

# Commit 3: Stranger WITHOUT signing key
- uses: git-commit
  config:
    path: ./out
    message: "add stranger file"
    author:
      name: Stranger
      email: stranger@example.com

# Push — each commit handled according to its signature
- uses: github-verified-push
  config:
    path: ./out
    targetBranch: stage/test
```

**Result on GitHub:**

| Commit | Author on GitHub | Verified | Why |
|--------|-----------------|----------|-----|
| &#34;render manifests for 1.29.5&#34; | `my-app[bot]` | :white_check_mark: Yes | Signed by system key (Ultimate trust) → re-signed by App |
| &#34;add extra service file&#34; | External Contributor | :x: No | Signed by untrusted per-commit author key (status `U`) → original author preserved |
| &#34;add stranger file&#34; | Stranger | :x: No | Unsigned → original author preserved |

## Tested Use Cases

All cases tested end-to-end against [EronWright/kargo-demo](https://github.com/EronWright/kargo-demo) on a local Tilt dev cluster.

### 1. Basic verified push (GPG configured)

Single commit signed by Kargo&#39;s system key → re-signed by GitHub App → **Verified**.

```
imported GPG signing key  path="/etc/kargo/git/signingKey"  trustLevel="ultimate"
commit signature verified  commit=83154ad  keyID=9A72EEED9004F4D2  signer="Kargo"
created revision  signed=true
```

Result: [`case-b-gpg-signed`](https://github.com/EronWright/kargo-demo/commit/60e53cddddd000913958df9212ca8432c3a23635) — Verified :white_check_mark:, author = `kargo-demo-signer-eronwright[bot]`

### 2. No GPG configured

All commits preserve original authorship, none verified.

Result: [`case-a-no-gpg`](https://github.com/EronWright/kargo-demo/commit/997f3dcb2c13764e9ab19c5f008ff82175169c6b) — Not verified, author = `Kargo`

### 3. Unsigned commit (GPG configured)

Commit without signing key → original author preserved, not verified.

```
commit is unsigned, propagating  commit=ecabc5f
created revision  signed=false
```

Result: [`case-c-unsigned`](https://github.com/EronWright/kargo-demo/commit/d4cb620f559585c24615a457afed34dde67e0849) — Not verified, author = `External Contributor`

### 4. Mixed signed + unsigned commits

Two commits in one promotion: one signed by system key, one unsigned. Both replayed correctly with per-commit signing decisions.

| Commit | Author on GitHub | Verified |
|--------|-----------------|----------|
| `acab8835` — render manifests | `kargo-demo-signer-eronwright[bot]` | :white_check_mark: Yes |
| `f6dae900` — add extra service file | `External Contributor` | :x: No |

### 5. Commit signed by unknown GPG key (status `E`)

External contributor&#39;s GPG key not in controller&#39;s keyring → detected as unknown, original author preserved.

```
commit signed by unknown key, propagating  commit=5765b32  keyID=5633972AD665B75E
created revision  signed=false
```

Result: [`eb9d1d4`](https://github.com/EronWright/kargo-demo/commit/eb9d1d4dd3c96f8e2d38cc4a46895d88f5314609) — Not verified, author = `External Contributor`

### 6. Per-commit author with signing key (untrusted)

Per-commit author GPG key imported into repo keyring without trust. Commit reports status `U` and is propagated without re-signing.

```
imported GPG signing key  path=".../signing-key.asc"  trustLevel=""
commit not signed by trusted key, propagating  commit=8a4fc26  status="U"
created revision  signed=false
```

Result: Not verified, original author preserved.

### 7. Generated target branch

`generateTargetBranch: true` → new branch `kargo/promotion/verified-push-gen-branch-sr5dc` created with verified commit [`a5218f1a`](https://github.com/EronWright/kargo-demo/commit/a5218f1a2acd139f350eb510e36e44156d40a706).

### 8. Force push

`force: true` → diverged branch force-updated successfully.

## Files Changed

| File | Change |
|------|--------|
| `pkg/promotion/runner/builtin/github_verified_pusher.go` | New promotion step implementation |
| `pkg/promotion/runner/builtin/github_verified_pusher_test.go` | Unit tests (40+ test cases) |
| `pkg/promotion/runner/builtin/schemas/github-verified-push-config.json` | JSON schema for step config |
| `pkg/controller/git/work_tree.go` | Add `CommitSignatureStatuses` to `WorkTree` interface |
| `pkg/controller/git/base_repo.go` | GPG keyring helpers, `setupAuthor` returns `*User` with `gpgFingerprint` |
| `pkg/controller/git/commit_signing.go` | `CommitSignatureInfo` struct and GPG status/trust types |
| `pkg/controller/git/mock_repo.go` | Mock implementation |
| `pkg/promotion/runner/builtin/git_cloner.go` | Pass signing key path; per-commit author keys imported without trust |
| `docs/.../github-verified-push.md` | Full documentation page |

## Test Plan

- [x] 40+ unit tests covering schema validation, all GPG status codes, signing decisions, error handling, cleanup
- [x] `make lint-go` passes
- [x] Manual e2e: no GPG → unsigned, original author preserved
- [x] Manual e2e: GPG + system-key-signed → Verified via App
- [x] Manual e2e: GPG + unsigned commit → original author, not verified
- [x] Manual e2e: mixed signed/unsigned commits in single promotion
- [x] Manual e2e: commit signed by unknown key → propagated as-is
- [x] Manual e2e: per-commit author with signing key → untrusted, propagated as-is
- [x] Manual e2e: `generateTargetBranch: true` → branch created with verified commit
- [x] Manual e2e: `force: true` push

### DIFF
--- a/pkg/controller/git/base_repo.go
+++ b/pkg/controller/git/base_repo.go
@@ -122,15 +122,18 @@ type User struct {
 	// signing key. When set, setupAuthor marks the key with the
 	// corresponding ownertrust level after import.
 	SigningKeyTrustLevel SigningKeyTrustLevel
+	// gpgFingerprint is the fingerprint of the imported GPG signing key.
+	// It is set by setupAuthor after a successful key import.
+	gpgFingerprint string
 }
 
 // setupAuthor configures the git CLI with a default commit author.
 // Optionally, the author can have an associated signing key. When using GPG
 // signing, the name and email must match the GPG key identity. The directory
 // specified by homeDir is used as a virtual home directory for all commands
-// executed by this method. Returns the GPG key fingerprint if a signing key
-// was imported, or an empty string otherwise.
-func (b *baseRepo) setupAuthor(homeDir string, author *User) (string, error) {
+// executed by this method. Returns the fully resolved User (with
+// gpgFingerprint set when a signing key was imported).
+func (b *baseRepo) setupAuthor(homeDir string, author *User) (*User, error) {
 	if author == nil {
 		author = &User{}
 	}
@@ -149,7 +152,7 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) (string, error) {
 	// identity.
 	b.setCmdHome(cmd, homeDir)
 	if _, err := libExec.Exec(cmd); err != nil {
-		return "", fmt.Errorf("error configuring git user name: %w", err)
+		return nil, fmt.Errorf("error configuring git user name: %w", err)
 	}
 
 	if author.Email == "" {
@@ -161,7 +164,7 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) (string, error) {
 	cmd.Dir = homeDir
 	b.setCmdHome(cmd, homeDir)
 	if _, err := libExec.Exec(cmd); err != nil {
-		return "", fmt.Errorf("error configuring git user email: %w", err)
+		return nil, fmt.Errorf("error configuring git user email: %w", err)
 	}
 
 	// For now, since only GPG signing is supported, we will assume GPG if the
@@ -175,20 +178,21 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) (string, error) {
 			cmd.Dir = homeDir
 			b.setCmdHome(cmd, homeDir)
 			if _, err := libExec.Exec(cmd); err != nil {
-				return "", fmt.Errorf(
+				return nil, fmt.Errorf(
 					"error configuring commit gpg signing: %w", err,
 				)
 			}
 
 			fingerprint, err := b.importGPGSigningKey(homeDir, author)
 			if err != nil {
-				return "", err
+				return nil, err
 			}
-			return fingerprint, nil
+			author.gpgFingerprint = fingerprint
+			return author, nil
 		}
 	}
 
-	return "", nil
+	return author, nil
 }
 
 // importGPGSigningKey imports a GPG signing key into the keyring rooted at
--- a/pkg/controller/git/work_tree.go
+++ b/pkg/controller/git/work_tree.go
@@ -216,7 +216,7 @@ func (w *workTree) Commit(message string, opts *CommitOptions) error {
 	}
 
 	var homeDir string
-	var fingerprint string
+	var author *User
 	if opts.Author != nil {
 		// This author information is specific to this commit, so we will override
 		// repository-level author information by creating a temporary home
@@ -235,7 +235,7 @@ func (w *workTree) Commit(message string, opts *CommitOptions) error {
 					Error(cleanErr, "error removing virtual home directory", "path", homeDir)
 			}
 		}()
-		if fingerprint, err = w.setupAuthor(homeDir, opts.Author); err != nil {
+		if author, err = w.setupAuthor(homeDir, opts.Author); err != nil {
 			return fmt.Errorf(
 				"error setting up author information for commit command: %w", err,
 			)
@@ -259,10 +259,10 @@ func (w *workTree) Commit(message string, opts *CommitOptions) error {
 	// Import the per-commit author's public key into the repo-level GPG
 	// keyring so that downstream steps (e.g. github-verified-push) can
 	// verify the signature.
-	if fingerprint != "" {
+	if author != nil && author.gpgFingerprint != "" {
 		if err := w.importGPGPublicKey(
-			homeDir, w.homeDir, fingerprint,
-			SigningKeyTrustLevelUltimate,
+			homeDir, w.homeDir, author.gpgFingerprint,
+			SigningKeyTrustLevelUnknown,
 		); err != nil {
 			return fmt.Errorf(
 				"error importing author signing key into repo keyring: %w",
--- a/pkg/promotion/runner/builtin/git_cloner.go
+++ b/pkg/promotion/runner/builtin/git_cloner.go
@@ -152,9 +152,8 @@ func (g *gitCloner) run(
 			Email:      cfg.Author.Email,
 			SigningKey: cfg.Author.SigningKey, // Optional, may be empty
 		}
-		if repoUser.SigningKey != "" {
-			repoUser.SigningKeyTrustLevel = git.SigningKeyTrustLevelUltimate
-		}
+		// SigningKeyTrustLevel is intentionally left at the default
+		// (unknown). Only the system-level signing key is trusted.
 	} else {
 		repoUser = g.gitUser // Default to the system-level gitUser
 	}
--- a/pkg/promotion/runner/builtin/github_verified_pusher.go
+++ b/pkg/promotion/runner/builtin/github_verified_pusher.go
@@ -39,6 +39,8 @@ const (
 )
 
 func init() {
+	var once sync.Once
+	var pusher promotion.StepRunner
 	cfg := githubVerifiedPusherConfig{}
 	envconfig.MustProcess("", &cfg)
 	promotion.DefaultStepRunnerRegistry.MustRegister(
@@ -49,10 +51,17 @@ func init() {
 					promotion.StepCapabilityAccessCredentials,
 				},
 			},
+			// This factory function closes over a single instance of
+			// githubVerifiedPusher so that its mutexes are shared across
+			// all executions of this step runner, which is necessary to
+			// ensure proper locking behavior.
 			Value: func(
 				caps promotion.StepRunnerCapabilities,
 			) promotion.StepRunner {
-				return newGitHubVerifiedPusher(caps, cfg)
+				once.Do(func() {
+					pusher = newGitHubVerifiedPusher(caps, cfg)
+				})
+				return pusher
 			},
 		},
 	)
@@ -276,7 +285,7 @@ func (g *githubVerifiedPusher) run(
 			}, &promotion.TerminalError{
 				Err: fmt.Errorf(
 					"no credentials found for %s;"+
-					" a GitHub App installation token or personal access token is required",
+						" a GitHub App installation token or personal access token is required",
 					workTree.URL(),
 				),
 			}
@@ -464,8 +473,7 @@ func (g *githubVerifiedPusher) signAndUpdate(
 					Err: fmt.Errorf(
 						"cannot sign revision range %s..%s: "+
 							"comparison status is %q "+
-							"(target branch may have diverged; "+
-							"use force to overwrite)",
+							"(target branch may have diverged)",
 						targetHead, localHead, status,
 					),
 				}
--- a/pkg/controller/git/base_repo_test.go
+++ b/pkg/controller/git/base_repo_test.go
@@ -18,15 +18,16 @@ func TestSetupAuthor(t *testing.T) {
 		// the repo directory hasn't been created yet.
 		createRepoDir bool
 		author        *User
-		assert        func(*testing.T, string, string, string, error)
+		assert        func(*testing.T, string, string, *User, error)
 	}{
 		{
 			name:          "nil author uses defaults",
 			createRepoDir: true,
 			author:        nil,
-			assert: func(t *testing.T, homeDir, _ string, fingerprint string, err error) {
+			assert: func(t *testing.T, homeDir, _ string, author *User, err error) {
 				require.NoError(t, err)
-				require.Empty(t, fingerprint)
+				require.NotNil(t, author)
+				require.Empty(t, author.gpgFingerprint)
 				assertGitConfig(t, homeDir, "user.name", defaultUsername)
 				assertGitConfig(t, homeDir, "user.email", defaultEmail)
 			},
@@ -38,9 +39,10 @@ func TestSetupAuthor(t *testing.T) {
 				Name:  "Test User",
 				Email: "test@example.com",
 			},
-			assert: func(t *testing.T, homeDir, _ string, fingerprint string, err error) {
+			assert: func(t *testing.T, homeDir, _ string, author *User, err error) {
 				require.NoError(t, err)
-				require.Empty(t, fingerprint)
+				require.NotNil(t, author)
+				require.Empty(t, author.gpgFingerprint)
 				assertGitConfig(t, homeDir, "user.name", "Test User")
 				assertGitConfig(t, homeDir, "user.email", "test@example.com")
 			},
@@ -55,7 +57,7 @@ func TestSetupAuthor(t *testing.T) {
 				Name:  "Per-Commit Author",
 				Email: "per-commit@example.com",
 			},
-			assert: func(t *testing.T, homeDir, repoHomeDir string, _ string, err error) {
+			assert: func(t *testing.T, homeDir, repoHomeDir string, _ *User, err error) {
 				require.NoError(t, err)
 				assertGitConfig(t, homeDir, "user.name", "Per-Commit Author")
 				assertNoGitConfig(t, repoHomeDir)
@@ -67,7 +69,7 @@ func TestSetupAuthor(t *testing.T) {
 			name:          "succeeds when repo dir does not exist",
 			createRepoDir: false,
 			author:        nil,
-			assert: func(t *testing.T, homeDir, _ string, _ string, err error) {
+			assert: func(t *testing.T, homeDir, _ string, _ *User, err error) {
 				require.NoError(t, err)
 				assertGitConfig(t, homeDir, "user.name", defaultUsername)
 			},
@@ -80,7 +82,7 @@ func TestSetupAuthor(t *testing.T) {
 				Email:          "test@example.com",
 				SigningKeyPath: "/nonexistent/key.asc",
 			},
-			assert: func(t *testing.T, homeDir, _ string, _ string, err error) {
+			assert: func(t *testing.T, homeDir, _ string, _ *User, err error) {
 				// git config succeeds but gpg --import fails because the
 				// key file doesn't exist. This exercises the setCmdHome
 				// calls in the signing path.
@@ -96,7 +98,7 @@ func TestSetupAuthor(t *testing.T) {
 				Email:      "test@example.com",
 				SigningKey: "not-a-real-key",
 			},
-			assert: func(t *testing.T, homeDir, _ string, _ string, err error) {
+			assert: func(t *testing.T, homeDir, _ string, _ *User, err error) {
 				// The key file is written and git config succeeds but gpg
 				// --import fails because the key content is invalid.
 				require.ErrorContains(t, err, "error importing gpg key")
@@ -122,8 +124,8 @@ func TestSetupAuthor(t *testing.T) {
 				homeDir: repoHomeDir,
 			}
 
-			fingerprint, err := b.setupAuthor(homeDir, tc.author)
-			tc.assert(t, homeDir, repoHomeDir, fingerprint, err)
+			author, err := b.setupAuthor(homeDir, tc.author)
+			tc.assert(t, homeDir, repoHomeDir, author, err)
 		})
 	}
 }
--- a/pkg/promotion/runner/builtin/github_verified_pusher_test.go
+++ b/pkg/promotion/runner/builtin/github_verified_pusher_test.go
@@ -311,7 +311,7 @@ func Test_githubVerifiedPusher_signAndUpdate(t *testing.T) {
 				require.Error(t, err)
 				require.True(t, promotion.IsTerminal(err))
 				require.Contains(
-					t, err.Error(), "use force to overwrite",
+					t, err.Error(), "target branch may have diverged",
 				)
 				require.Equal(
 					t,
@@ -812,7 +812,7 @@ func Test_githubVerifiedPusher_signAndUpdate(t *testing.T) {
 				require.Error(t, err)
 				require.True(t, promotion.IsTerminal(err))
 				require.Contains(
-					t, err.Error(), "use force to overwrite",
+					t, err.Error(), "target branch may have diverged",
 				)
 				require.Equal(
 					t,


EronWright:
## Manual Test Results

Tested end-to-end against [EronWright/kargo-demo](https://github.com/EronWright/kargo-demo) using a local Tilt dev environment.

### Promotion

Created a promotion in the `kargo-demo` project with steps: `git-clone` → `git-clear` → `kustomize-set-image` → `kustomize-build` → `git-commit` → `github-verified-push`.

The promotion (`test-verified-push-l6lcs`) completed successfully — all 6 steps succeeded in ~4 seconds.

### Verification

**Commit is signed:**
```
$ gh api repos/EronWright/kargo-demo/commits/c8f93a4703ef363c7cd177a2ada8b2ff7d74a883 \
    --jq '{verified: .commit.verification.verified, reason: .commit.verification.reason, author: .commit.author.name, committer: .commit.committer.name}'

{
  "verified": true,
  "reason": "valid",
  "author": "kargo-demo-signer-eronwright[bot]",
  "committer": "GitHub"
}
```

**Staging ref cleaned up:**
```
$ gh api repos/EronWright/kargo-demo/git/matching-refs/kargo/staging --jq '.[].ref'
(empty — no leftover refs)
```

**Step output:**
```json
{
  "branch": "stage/test",
  "commit": "c8f93a4703ef363c7cd177a2ada8b2ff7d74a883",
  "commitURL": "https://github.com/eronwright/kargo-demo/commit/c8f93a4703ef363c7cd177a2ada8b2ff7d74a883"
}
```

**Branch HEAD correctly updated:** `stage/test` points to the signed commit, with the previous commit as its parent (clean fast-forward, no force push).

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,112 @@
+---
+sidebar_label: github-verified-push
+description: Pushes committed changes to a GitHub repository as verified (signed) commits using the GitHub REST API.
+---
```
jessesuen, line 4: This is new, so we should add a `beta` tag to this.

```

<span class="tag beta"></span>
---

**Comment by EronWright** [comment]
## Manual Test Results — Conditional Commit Signing

Tested in a local Tilt dev cluster with `EronWright/kargo-demo` fork. All three cases exercise the `github-verified-push` replay loop with different GPG configurations.

| Case | GPG Configured? | Commit Signed? | `signed` field | GitHub Verified? | Author |
|------|----------------|----------------|----------------|-----------------|--------|
| **A** | No | N/A | `false` | No (`unsigned`) | `Kargo <no-reply@kargo.io>` |
| **B** | Yes | Yes (G) | `true` | **Yes** (`valid`) | `kargo-demo-signer-eronwright[bot]` |
| **C** | Yes | No (N) | `false` | No (`unsigned`) | `External Contributor <external@example.com>` |

### Tagged commits on [EronWright/kargo-demo](https://github.com/EronWright/kargo-demo)

- [`case-a-no-gpg`](https://github.com/EronWright/kargo-demo/commit/997f3dcb2c13764e9ab19c5f008ff82175169c6b) — No GPG config, original authorship preserved, unsigned
- [`case-b-gpg-signed`](https://github.com/EronWright/kargo-demo/commit/60e53cddddd000913958df9212ca8432c3a23635) — GPG configured, Kargo-signed → GitHub App re-signed, **verified**
- [`case-c-unsigned`](https://github.com/EronWright/kargo-demo/commit/d4cb620f559585c24615a457afed34dde67e0849) — GPG configured but commit unsigned → original author preserved, not verified

### Controller log excerpts

**Case B** (GPG signed):
```
verifying commit signatures  signingKeyPath=/etc/kargo/git/signingKey  numCommits=1
commit signature verified    commit=83154ad  keyID=9A72EEED9004F4D2  signer="Kargo <no-reply@kargo.io>"
all commit signatures verified
created revision  signed=true
```

**Case C** (unsigned):
```
verifying commit signatures  signingKeyPath=/etc/kargo/git/signingKey  numCommits=1
commit is unsigned, propagating  commit=ecabc5f
all commit signatures verified
created revision  signed=false
```

File: docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+3. Fast-forwards the target branch to the final signed commit
+4. Deletes the staging ref
+
+Because commits are created through the GitHub API using a GitHub App
+installation token, they are automatically marked as verified by GitHub.
+
+:::info
+This step requires a **GitHub App installation token** stored as Git
+credentials for the repository. The GitHub App must have **Contents: read &
+write** permission on the target repository.
```
jessesuen: One caveat worth mentioning. Something along the lines of:

This step is intended only for pushing small, lightweight configuration changes. It uses the GitHub API, where payloads are limited to 40MiB (at the time of writing), is subject to GitHub API rate limiting, and is not optimized for large content pushes.

EronWright: That's actually not the case, because the content is pushed via the git client (not via the blob/tree API).

jessesuen: Ah, good to know.

EronWright:
## Manual test: mixed signed + unsigned commits

Tested conditional signing with **a promotion that produces two commits** — one signed by Kargo (default author) and one unsigned (custom author). GPG signing key is configured on the controller (via `GITCLIENT_SIGNING_KEY_PATH`).  We see both commits, one Verified as expected:

image


### Promotion steps

```yaml
steps:
- uses: git-clone
  config:
    repoURL: https://github.com/EronWright/kargo-demo.git
    checkout:
    - branch: main
      path: ./src
    - branch: stage/test
      path: ./out
- uses: git-clear
  config:
    path: ./out
- uses: kustomize-set-image
  config:
    path: ./src/base
    images:
    - image: public.ecr.aws/nginx/nginx
      newName: public.ecr.aws/nginx/nginx
      tag: "1.29.5"
- uses: kustomize-build
  config:
    path: ./src/stages/test
    outPath: ./out
- uses: git-commit
  config:
    path: ./out
    message: "render manifests for 1.29.5"
- uses: copy
  config:
    inPath: ./src/base/service.yaml
    outPath: ./out/extra-service.yaml
- uses: git-commit
  config:
    path: ./out
    message: "add extra service file"
    author:
      name: External Contributor
      email: external@example.com
- uses: github-verified-push
  config:
    path: ./out
    targetBranch: stage/test
```

### Result: Succeeded

Controller logs show correct per-commit behavior:

```
commit signature verified  commit=3bb77238  status=U  keyID=9A72EEED9004F4D2  signer="Kargo <no-reply@kargo.io>"
commit is unsigned, propagating  commit=8d0574e9
all commit signatures verified
created revision  index=1  total=2  original=3bb77238  new=acab8835  signed=true
created revision  index=2  total=2  original=8d0574e9  new=f6dae900  signed=false
```

### GitHub verification

| Commit | Message | Author | Verified |
|--------|---------|--------|----------|
| `acab8835` | render manifests for 1.29.5 | `kargo-demo-signer-eronwright[bot]` | ✅ Yes |
| `f6dae900` | add extra service file | `External Contributor` | ❌ No |

The Kargo-signed commit was re-signed by the GitHub App (verified), while the unsigned commit preserved its original author and was not marked verified. This matches the expected behavior for conditional signing.

EronWright:
## Manual Test: Commit signed by unknown GPG key (status `E`)

Tested the case where `git-commit` uses an explicit author with a GPG signing key that is **not** the Kargo controller's key. The controller should detect the unknown signature and propagate the original authorship without re-signing.

### Promotion steps

```yaml
steps:
- uses: git-clone
  config:
    repoURL: https://github.com/EronWright/kargo-demo.git
    checkout:
    - branch: main
      path: ./src
    - branch: stage/test
      path: ./out
- uses: git-clear
  config:
    path: ./out
- uses: kustomize-set-image
  config:
    path: ./src/base
    images:
    - image: public.ecr.aws/nginx/nginx
      newName: public.ecr.aws/nginx/nginx
      tag: "1.29.5"
- uses: kustomize-build
  config:
    path: ./src/stages/test
    outPath: ./out
- uses: git-commit
  config:
    path: ./out
    message: "render manifests (signed by external contributor)"
    author:
      name: External Contributor
      email: external@example.com
      signingKey: ${{ secret("external-contributor-signing-key").signingKey }}
- uses: github-verified-push
  config:
    path: ./out
    targetBranch: stage/test
```

### Controller logs

```
verifying commit signatures  signingKeyPath=/etc/kargo/git/signingKey  numCommits=1
commit signed by unknown key, propagating  commit=5765b32  keyID=5633972AD665B75E  signer=""
all commit signatures verified
created revision  index=1  total=1  original=5765b32  new=eb9d1d4  signed=false
```

### Result on GitHub

| Field | Value |
|-------|-------|
| Commit | [`eb9d1d4`](https://github.com/EronWright/kargo-demo/commit/eb9d1d4dd3c96f8e2d38cc4a46895d88f5314609) |
| Author | External Contributor / external@example.com |
| Committer | External Contributor / external@example.com |
| Verified | ❌ No (`unsigned` — original signature not preserved during replay) |
| GPG status | `E` (signed by unknown key) |

### Summary

The commit was signed by a GPG key unknown to the controller (a separate "External Contributor" key, not the Kargo controller's signing key). The controller correctly detected status `E`, set `signed=false`, and preserved the original author/committer identity. The commit appears on GitHub as unverified, attributed to the external contributor — as expected.

.claude/settings.json
krancour, line 1: I have some reservations about deciding for other users what permissions they'll give an agent. We should save this for another PR and debate it separately from this.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
+---
+sidebar_label: github-verified-push
+description: Pushes committed changes to a GitHub repository by replaying commits through the GitHub REST API, with optional commit signing.
```
krancour: Curious if there's ever a good reason to do this if _not_ for the benefit of the GitHub App signing the commit.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+
+<span class="tag beta"></span>
+
+`github-verified-push` pushes committed changes from a local working tree to a
+GitHub repository using the GitHub REST API. It is a drop-in replacement for
+[`git-push`](git-push.md) that replays commits through the API instead of using
+`git push`.
+
+This step is designed to work with repositories that enforce commit signing via
+branch protection rules. Under the hood it:
```
krancour: I think that option only requires _verified_ signatures, so the use case this is targeted at seems to go above and beyond even that branch protection. Is it worth being more specific?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+
+1. Pushes local commits to a temporary, non-branch staging ref on GitHub
+   (invisible in the branch list)
+2. Replays each commit via the GitHub REST API
+3. Fast-forwards the target branch to the final commit
+4. Deletes the staging ref
```
krancour: I'd usually prefer to let Docusaurus handle the numbering. It saves us from goofs later if we insert an item.

```suggestion
1. Pushes local commits to a temporary, non-branch staging ref on GitHub
   (invisible in the branch list)
1. Replays each commit via the GitHub REST API
1. Fast-forwards the target branch to the final commit
1. Deletes the staging ref
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+
+:::info
+This step requires a **GitHub App installation token** stored as Git
+credentials for the repository. The GitHub App must have **Contents: read &
+write** permission on the target repository.
+:::
```
krancour, line 54: Although not strictly required by Docusaurus, my understanding is it's a best practice to have a blank line after the start of an admonition and another before the end. Apparently it doesn't matter except in certain edge cases, so it's really a defensive practice. Applies in other spots as well.

```suggestion
:::info

This step requires a **GitHub App installation token** stored as Git
credentials for the repository. The GitHub App must have **Contents: read &
write** permission on the target repository.

:::
```

krancour: Formatting aside, we should link to relevant (Kargo) docs about authenticating using GitHub Apps.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Commit Signing Behavior
+
+Which commits receive GitHub's "Verified" badge depends on whether the Kargo
+controller is configured with a
```
krancour: Nit. And take it or leave it. I think this clarifies just a teensy bit.


```suggestion
controller is configured with its own
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+[GPG signing key](../../../40-operator-guide/20-advanced-installation/30-common-configurations.md#signing-commits).
```
krancour, line 60: The key can be specified on individual clone steps as well. idk if it's overkill to get into that here. I leave it up to you.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+
+**When a signing key is configured**, this step verifies the GPG signature on
+each local commit before replaying it:
+
+- **Kargo-signed commits** (including those with expired signatures or keys)
+  are re-signed by the GitHub App and appear as "Verified" on GitHub.
+- **Unsigned commits** (or commits signed by a different key) preserve their
```
krancour, line 71: Worth noting the loss of the original signature? Up to you.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+## Commit Signing Behavior
+
+Which commits receive GitHub's "Verified" badge depends on whether the Kargo
+controller is configured with a
+[GPG signing key](../../../40-operator-guide/20-advanced-installation/30-common-configurations.md#signing-commits).
+
+**When a signing key is configured**, this step verifies the GPG signature on
+each local commit before replaying it:
+
+- **Kargo-signed commits** (including those with expired signatures or keys)
```
krancour: Can signatures expire? Just the keys can, right?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+  are re-signed by the GitHub App and appear as "Verified" on GitHub.
+- **Unsigned commits** (or commits signed by a different key) preserve their
+  original author and committer and are _not_ marked as "Verified".
+- **Commits with bad or revoked signatures** cause the step to fail with a
```
krancour, line 73: Similar question to [here](https://github.com/akuity/kargo/pull/5840/changes#r2891612998). Can a signature actually be revoked?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+  terminal error, since they may indicate tampering.
+
+**When no signing key is configured**, all commits preserve their original
+author and committer. None are marked as "Verified" on GitHub.
```
krancour, line 77: I'm sure I'll get to the answer as I continue the review, but this immediately makes me wonder if this scenario actually falls back on the existing push step runner.

EronWright:
## Test Results: `generateTargetBranch` fix

Tested the branch creation feature end-to-end on a local Tilt cluster with the `kargo-demo` project.

**Promotion:** Created an ad-hoc promotion with `generateTargetBranch: true` on the `github-verified-push` step.

**Result:** All 6 steps succeeded. A new branch was created on GitHub with a verified commit:

- **Branch:** `kargo/promotion/verified-push-gen-branch-sr5dc` on `EronWright/kargo-demo`
- **Commit:** [`a5218f1a`](https://github.com/EronWright/kargo-demo/commit/a5218f1a2acd139f350eb510e36e44156d40a706)
- **Verified:** ✅ `true` (reason: `valid`)
- **Author:** `kargo-demo-signer-eronwright[bot]`

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+:::
+
+## Configuration
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `path` | `string` | Y | Path to a Git working tree containing committed changes. |
+| `targetBranch` | `string` | N | The branch to push to in the remote repository. Mutually exclusive with `generateTargetBranch=true`. If neither of these is provided, the target branch will be the same as the branch currently checked out in the working tree. |
+| `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. A value of `true` is mutually exclusive with `targetBranch`. This is useful when a subsequent step will open a pull request. |
+| `insecureSkipTLSVerify` | `boolean` | N | Whether to skip TLS verification when communicating with the GitHub API. Default is `false`. Intended for GitHub Enterprise instances with self-signed certificates. |
```
krancour, line 96: Since this aims to be a drop-in replacement for `git-push`, you may need to account for the fact of #5785 adding support for pushing a tag...

Having said that, I haven't given any serious thought yet to how any of this process does or doesn't differ for that scenario.

.claude/settings.json
EronWright: Sorry, not intentional to have in this PR (and I generally try to separate concerns).

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Output
+
+| Name | Type | Description |
+|------|------|-------------|
+| `branch` | `string` | The name of the remote branch pushed to by this step. |
+| `commit` | `string` | The ID (SHA) of the final commit created by this step. |
+| `commitURL` | `string` | The URL of the final commit on GitHub. |
+
+## Examples
```
krancour, line 106: I love how straightforward this is.

pkg/controller/git/base_repo.go
```diff
@@ -133,10 +133,13 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
...
-	// Override the cmd.Dir that's set by b.buildGitCommand(). It's normally the
-	// repository's path, but if this method was called as part of the cloning
-	// process, that path may not exist yet.
+	// Override the cmd.Dir and HOME that are set by b.buildGitCommand().
+	// cmd.Dir is normally the repository's path, but if this method was called
+	// as part of the cloning process, that path may not exist yet. HOME must
+	// match homeDir so that git config --global writes to the correct
+	// .gitconfig (especially when Commit() creates a temporary home).
 	cmd.Dir = homeDir
+	b.setCmdHome(cmd, homeDir)
```
krancour, line 153: Just fyi, #5857 is approved if you want to merge that and rebase.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Agreed, will make it more explicitly about verification and is _based_ on commit signing.

pkg/controller/git/work_tree.go
```diff
@@ -665,6 +669,52 @@ func (w *workTree) UpdateSubmodules() error {
...
+	args := append(
+		[]string{"log", "--no-walk", "--format=%H%x09%G?%x09%GK%x09%GS"},
+		ids...,
+	)
+	res, err := libExec.Exec(w.buildGitCommand(args...))
+	if err != nil {
+		return nil, fmt.Errorf("error getting signature statuses: %w", err)
+	}
+	statuses := make(map[string]CommitSignatureInfo, len(ids))
+	scanner := bufio.NewScanner(bytes.NewReader(res))
```
krancour, line 729: Nit: Probably already hasn't been handled well in a lot of existing spots, but a comment with a sample response will make it much easier to understand the logic that follows.

krancour:
Did the full codegen process run? I think step schemas are usually copied (with some small changes for compatibility) over to `ui/`, but I don't see those.

pkg/promotion/runner/builtin/schemas/github-verified-push-config.json
krancour, line 1: Just flagging another spot where https://github.com/akuity/kargo/pull/5785 adds some changes that might need to be replicated here if the goal remains to be a drop-in replacement.

(Also worth saying: If it's better at some point that we don't struggle to hard for this to remain a 100% drop-in replacement, I think that's ok, too, as long as we're _close_.)

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+	DeleteRef(
+		ctx context.Context,
+		owner, repo, ref string,
+	) (*github.Response, error)
+}
+
+// githubVerifiedPushClientWrapper wraps a *github.Client to implement
+// githubVerifiedPushClient.
+type githubVerifiedPushClientWrapper struct {
+	client *github.Client
```
krancour, line 107: Is there any reason not to do this? I think this would make `githubVerifiedPushClientWrapper` implement `githubVerifiedPushClient` instantly without any need to explicitly implement all the methods -- which are all doing nothing except calling the method of the same name on the wrapped client.

```suggestion
	*github.Client
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: I agree, the "above and beyond" is working with such repositories _using an app installation token_.

I sense your deeper question. If Kargo is configured to use a PAT, is there reason to use the `github-verified-push` step, and how does it work?  I will explore that question and update the PR.  First guess is the commits will appear as verified by the GH user account, even if the GPG isn't associated to them.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Good idea, and I'll update the project's memory file to watch out for that in future.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+func init() {
+	promotion.DefaultStepRunnerRegistry.MustRegister(
+		promotion.StepRunnerRegistration{
+			Name: stepKindGitHubVerifiedPush,
+			Metadata: promotion.StepRunnerMetadata{
+				RequiredCapabilities: []promotion.StepRunnerCapability{
+					promotion.StepCapabilityAccessCredentials,
+				},
+			},
+			Value: newGitHubVerifiedPusher,
```
krancour: I might do something like this because it leaves `newFileDeleter()` cleaner, with injectable config, instead of always having to get its config from env vars. But, arguably, that's overkill if you're not going to be calling `newFileDeleter()` in different context, e.g. for testing purposes, but you're not currently, so take it or leave it.

```suggestion
			Value: func() promotion.StepRunnerFactory {
				cfg := githubVerifiedPusherConfig{}
				envconfig.MustProcess("", &cfg)
				return func(capabilities promotion.StepRunnerCapabilities) promotion.StepRunner {
					return newFileDeleter(capabilities, cfg)
				}
			}(),
```

krancour: Actually... the thing that's a teensy bit better about this is that the potential panic would happen at startup this way.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: I'm seeking as much alignment as possible, and will consider it.  Also missing a "force" option, and there's "max tries" stuff to consider.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+
+func (g *githubVerifiedPusher) run(
+	ctx context.Context,
+	stepCtx *promotion.StepContext,
+	cfg builtin.GitHubVerifiedPushConfig,
+) (promotion.StepResult, error) {
+	logger := logging.LoggerFromContext(ctx)
+
+	// Short-circuit if shared state has output from a previous execution of
+	// this step that contains a commit SHA.
```
krancour: You can _probably_ get by without this. Most steps did this at one point in time because the earliest version of the step execution engine didn't keep track itself of where we were in the process. Every step re-executed every time and had to short-circuit if it believed it had already done its job.

Those days are long behind us. I think there is only one remaining step and I struggle to recall why it's still required in that one case.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+				"error getting credentials for %s: %w", workTree.URL(), err,
+			)
+	}
+	if creds != nil {
+		loadOpts.Credentials = &git.RepoCredentials{
+			Username:      creds.Username,
+			Password:      creds.Password,
+			SSHPrivateKey: creds.SSHPrivateKey,
+		}
+	}
```
krancour, line 269: Should we have some kind of sanity check here? If we know this step to be 100% dependent on a GitHub App, we could fail right here if we _didn't_ find creds or if we found creds, but they don't resemble an installation token. Installation tokens all have the prefix `ghs_`.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+}
+
+// releaseBranchLock releases the per-branch mutex.
+func (g *githubVerifiedPusher) releaseBranchLock(repoURL, branch string) {
+	key := fmt.Sprintf("%s:%s", repoURL, branch)
+	g.masterMu.Lock()
+	mu := g.branchMus[key]
+	g.masterMu.Unlock()
+	mu.Unlock()
+}
```
krancour, line 767: Sadly, this may not work because other controllers or human users might be writing to the same branch.

I think what you may need to do is try the whole process over from the start if the fast forward on the target branch fails. The existing push step does something along those lines, iirc.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: The normal case is definitely key expiration, but sig expiration is a possible result code from GPG. 
Git commits generally have no such expiration. Here's a table of signature status codes from `git log` for reference:

Code | Meaning | Git docs
-- | -- | --
G | Good (valid) signature with a trusted key |  
U | Good signature with an untrusted key (not in the trust web) |  
X | Good signature that has expired (the signature itself is past its expiration timestamp) |  
Y | Good signature made by an expired key (the key was valid when signing, but has since expired) |  
B | Bad signature (cryptographically invalid) |  
E | Cannot check (key not in keyring) |  
N | No signature |  
R | Good signature by a revoked key

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+			repoURL,
+		)
+	}
+	owner, repo := parts[0], parts[1]
+
+	httpClient := cleanhttp.DefaultClient()
+	if insecureSkipTLSVerify {
+		transport := cleanhttp.DefaultTransport()
+		transport.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: true, // nolint: gosec
```
krancour, line 810: Shouldn't this be conditional?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Yes, GnuPG can import or download a key's revocation status, yielding status `R` in subsequent checks, and treated as unverified.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Not at this time.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+	g.masterMu.Unlock()
+	mu.Unlock()
+}
+
+// newGitHubClient parses a repository URL and creates an authenticated GitHub
+// API client. Returns the owner, repo name, and client.
+func (g *githubVerifiedPusher) newGitHubClient(
+	repoURL, token string,
+	insecureSkipTLSVerify bool,
+) (string, string, githubVerifiedPushClient, error) {
```
krancour: Nit: It's a little strange for the resurn values to be more than what is implied by the method name `newGitHubClient()`. It's not a hill I'd die on, but I might consider breaking this method up intol one for finding owner + repo and another for returning the client, _given_ owner and repo as args.

pkg/promotion/runner/builtin/github_verified_pusher.go
EronWright: For the sake of generality, I would advocate for the step to be fully functional with a PAT, even if there's no real value proposition other than generality (e.g. development-time use of promotion templates that will go into production with an app token).

I will polish the code and docs to clarify all this and ensure that it works.

pkg/promotion/runner/builtin/github_verified_pusher.go
EronWright: Yeah, this simply replicates the approach in git-push, which seems to have the limited goal of preventing concurrent access across steps.   I suppose the two "push" variants should share a structure to keep that idea going.

This isn't related to recovery, which always starts from scratch.

pkg/promotion/runner/builtin/github_verified_pusher.go
krancour: Actually...

Now that I've gotten a little farther, I also see `buildCommitURL()`. That along with the logic for building the client look very similar to what's in the "GitHub Provider." The `Provider` interface (I don't love the name) is an abstraction over the common capabilities of the various git hosting providers.

In no way do I think any of what you're doing here needs to involve that interface. What we're doing involve a unique GitHub feature that doesn't necessarily have any direct analog anywhere else.

But what might be good would be to break a few pieces like extracting owner + repo and building a client out into a separate github package. Then the GitHub implementation of that Provider interface and this step, alike, can both use that. I think GitHub App credential provider could also, potentially, benefit from some of that.

The thing I just want to avoid is having very similar logic for these things spread out and replicated throughout the code base.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+	workTree git.WorkTree,
+	commits []*github.RepositoryCommit,
+) (map[string]git.CommitSignatureInfo, error) {
+	if g.gitUser.SigningKeyPath == "" {
+		return nil, nil
+	}
+
+	logger := logging.LoggerFromContext(ctx)
+
+	// Collect non-empty SHAs.
```
krancour: How does an empty sha happen?

EronWright: A non-commit object in the iteration? I'll investigate.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,848 @@
...
+	return fmt.Sprintf(
+		"https://%s%s/commit/%s",
+		parsedURL.Host, parsedURL.Path, sha,
+	)
+}
+
+// verifyCommitSignatures verifies GPG signatures on local commits before they
+// are replayed via the GitHub API. When the controller is configured with a
+// signing key (GITCLIENT_SIGNING_KEY_PATH), this method uses the WorkTree's
+// GPG keyring (populated by git-clone's setupAuthor) to check signatures.
```
krancour, line 848: 😬 Something just occurred to me. If a `git-commit` step had overridden the signing key, it wouldn't be recognized. (I don't think?)

EronWright: Correct, because the keyrings are kept separate, and so the push step doesn't recognize the key and downgrades to unsigned.  See [this report](https://github.com/akuity/kargo/pull/5840#issuecomment-4006529233).  In concept, Kargo could vouch for numerous keys beyond the system key, we just need to be able to explain it.

At a technical level, I'm considering a change such that the "[trusted](https://gpgtools.tenderapp.com/kb/faq/what-is-ownertrust-trust-levels-explained)" status of a given key in the keyring would determine whether we vouch for its commits.  We can then decouple the "which keys are in the ring" from "which keys produce verified commits" on the basis of trust level.  Note that the output of `git log` relays whether a given key is trusted.

With this approach, one could certify a given key using Kargo's GPG key (`gpg --sign-key user@example.com`), to bless it as one that produces Verified commits.

Update: implemented in  https://github.com/akuity/kargo/pull/5840/commits/8b90457eac88b26c5bc8b1f1d846a5bd9a2bfbd6 by importing the public key component into the main repo.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Update: the logic was adjusted to accept only code G as a Verified Commit.

pkg/promotion/runner/builtin/github_verified_pusher.go
EronWright: It isn't a perfect substitute, because the wrapper flattens the API somewhat.

pkg/promotion/runner/builtin/github_verified_pusher.go
EronWright: Thanks.  Made a change to have the envvar be parsed eagerly.

EronWright:
## Manual test: per-commit author GPG key import

Tested the expanded validation from `8b90457e` (import per-commit author GPG key into repo keyring). Two separate promotions, each with three commits covering all three signing scenarios.

### Test 1: git-commit author override

`git-clone` uses defaults (Kargo author). Three commits with different `git-commit` authors.

| Commit | Message | Author on GitHub | Verified | Co-authored-by |
|--------|---------|-----------------|----------|----------------|
| `f62a826a` | render manifests for 1.29.5 | kargo-demo-signer-eronwright[bot] | ✅ Yes | — |
| `dcf150d0` | add extra service file | kargo-demo-signer-eronwright[bot] | ✅ Yes | External Contributor \<external@example.com\> |
| `a0d3116d` | add stranger file | Stranger | ❌ No | — |

### Test 2: git-clone author override

`git-clone` sets author to External Contributor with signing key. Three commits — two inherit, one overrides.

| Commit | Message | Author on GitHub | Verified | Co-authored-by |
|--------|---------|-----------------|----------|----------------|
| `0080a185` | render manifests for 1.29.5 | kargo-demo-signer-eronwright[bot] | ✅ Yes | External Contributor \<external@example.com\> |
| `ccfb2e07` | add extra service file | Stranger | ❌ No | — |
| `1aa930b4` | add another file from clone author | kargo-demo-signer-eronwright[bot] | ✅ Yes | External Contributor \<external@example.com\> |

### Summary

All three code paths work correctly:
- **Kargo default author** → signed by Kargo's GPG key, re-signed by GitHub App → Verified
- **External contributor with signing key** (via `git-commit` or `git-clone` author) → key imported into repo keyring, signature verified, re-signed by GitHub App with `Co-authored-by` trailer → Verified
- **Stranger without signing key** → unsigned commit propagated as-is → Not Verified

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
krancour: TIL

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
krancour: I think how I posed the question initially wasn't clear. Sorry.

I was trying to highlight that branch protection rules can only require "Commits pushed to matching branches must have verified signatures." There's no option to require the verified signature was produced via a GitHub App. i.e. This step isn't required to clear the particular hurdle that's mentioned, which makes me question whether our description of when you'd use this is, perhaps, not sufficiently narrow.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,180 @@
...
+step completes. This is expected and does not affect subsequent promotions, since
+each promotion clones a fresh working tree.
+
+:::info
+
+This step requires a **GitHub App installation token** stored as Git
+credentials for the repository. The GitHub App must have **Contents: read &
+write** permission on the target repository. See
+[GitHub App Authentication](../../50-security/30-managing-secrets.md#github-app-authentication)
+for setup instructions.
```
krancour: Related to some other questions, like this [one](https://github.com/akuity/kargo/pull/5840#discussion_r2905996475).

Does it really require that specific sort of credential. Or _can_ this work with a PAT, for instance, but there just isn't any really good reason to do things that way?

pkg/promotion/runner/builtin/github_verified_pusher.go
krancour: Related: https://github.com/akuity/kargo/pull/5840/changes#r2906130401

EronWright:
## Manual Test: PAT Authentication Compatibility

Tested `github-verified-push` with a **classic Personal Access Token** (instead of a GitHub App installation token) to verify compatibility. Used the same mixed-author promotion (3 commits covering all signing paths).

### Setup

- Disabled the GitHub App credential by changing its `kargo.akuity.io/cred-type` label
- Used an existing PAT credential (`kargo-demo-repo` secret) with `username: EronWright` and a classic PAT (`ghp_*`)
- GPG signing key configured on the controller

### Promotion: `demo-commit-author-sck6b`

All 10 steps succeeded in ~12 seconds. Controller logs show identical GPG verification behavior to the App token case:

```
commit signature verified    commit=d4288d39  status=G  keyID=9A72EEED9004F4D2  signer="Kargo <no-reply@kargo.io>"
commit signature verified    commit=02092b58  status=G  keyID=5633972AD665B75E  signer="External Contributor <external@example.com>"
commit not signed by trusted key, propagating  commit=d44d4d06  status=N
created revision  index=1  total=3  original=d4288d39  new=0fa0fe35  signed=true
created revision  index=2  total=3  original=02092b58  new=a747c251  signed=true
created revision  index=3  total=3  original=d44d4d06  new=63987a18  signed=false
```

### Results on GitHub

| Commit | Message | Author | Verified | Reason |
|--------|---------|--------|----------|--------|
| [`0fa0fe35`](https://github.com/EronWright/kargo-demo/commit/0fa0fe3508d17b1f3a2bc6cb06c4539ad2601a48) | render manifests for 1.29.5 | Eron Wright | ❌ No | `unsigned` |
| [`a747c251`](https://github.com/EronWright/kargo-demo/commit/a747c2512fcb29b49f9a5e2d02baa0318dd1d8cc) | add extra service file | Eron Wright (+Co-authored-by) | ❌ No | `unsigned` |
| [`63987a18`](https://github.com/EronWright/kargo-demo/commit/63987a18431272f6ebe09ccfa7eeef0521414f61) | add stranger file | Stranger | ❌ No | `unsigned` |

### Key Finding

**The step works correctly with PAT auth** — commits are replayed, authorship is attributed to the PAT owner for Kargo-signed commits, `Co-authored-by` trailers are added, and unsigned commits preserve their original author. However, **classic PATs do not produce the "Verified" badge**.

This is expected GitHub behavior: GitHub only auto-signs commits created via the REST API when authenticated as a **bot** (GitHub App installation token or `GITHUB_TOKEN`). PATs authenticate as a user, and GitHub's position is that users must sign their own commits with GPG/SSH keys. This is consistent with how GitHub's own server-side rebase-and-merge also produces unverified commits for users.

**Docs updated** to clarify that PATs are supported but do not produce verified commits.

pkg/promotion/runner/builtin/github_verified_pusher.go
EronWright: @krancour do you agree that I should factor the promotion controller's "branch locker" such that it is shared between the two pushers?  Seems nonsensical to have a separate lock for git-push vs github-verified-push.  Or, rethink the lock idea altogether (in a follow-up).

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: True, it wasn't clear; revised.

I tested the use of a PAT, and it is able to push but commits do not get a Verified stamp.

https://github.com/akuity/kargo/pull/5840#issuecomment-4025862617

pkg/promotion/runner/builtin/git_cloner.go
```diff
@@ -148,6 +152,9 @@ func (g *gitCloner) run(
 			Email:      cfg.Author.Email,
 			SigningKey: cfg.Author.SigningKey, // Optional, may be empty
 		}
+		if repoUser.SigningKey != "" {
+			repoUser.SigningKeyTrustLevel = git.SigningKeyTrustLevelUltimate
+		}
```
krancour: I think this is the spot where we decided against trusting developer-provided keys, to any extent.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,957 @@
...
+				},
+			},
+			Value: func(
+				caps promotion.StepRunnerCapabilities,
+			) promotion.StepRunner {
+				return newGitHubVerifiedPusher(caps, cfg)
+			},
+		},
+	)
+}
```
krancour, line 68: See #5888

```suggestion
func init() {
	var once sync.Once
	var pusher promotion.StepRunner
	cfg := githubVerifiedPusherConfig{}
	envconfig.MustProcess("", &cfg)
	promotion.DefaultStepRunnerRegistry.MustRegister(
		promotion.StepRunnerRegistration{
			Name: stepKindGitHubVerifiedPush,
			Metadata: promotion.StepRunnerMetadata{
				RequiredCapabilities: []promotion.StepRunnerCapability{
					promotion.StepCapabilityAccessCredentials,
				},
			},
			// This factory function closes over a single instance of
			// githubVerifiedPusher so that that its mutexes are shared across all
			// executions of this step runner, which is necessary to ensure proper
			// locking behavior.
			Value: func(caps promotion.StepRunnerCapabilities) promotion.StepRunner {
				once.Do(func() {
					pusher = newGitHubVerifiedPusher(caps, cfg)
				})
				return pusher
			},
		},
	)
}```

File: pkg/promotion/runner/builtin/github_verified_pusher.go [hunk 29]
```diff
@@ -0,0 +1,957 @@
...
+		if !force {
+			return promotion.StepResult{
+					Status: kargoapi.PromotionStepStatusFailed,
+				}, &promotion.TerminalError{
+					Err: fmt.Errorf(
+						"cannot sign revision range %s..%s: "+
+							"comparison status is %q "+
+							"(target branch may have diverged; "+
+							"use force to overwrite)",
+						targetHead, localHead, status,
```
krancour, line 477: Since it's an option that should only be exercised with great caution (and imho usually points to a poorly engineered promotion process), I'm uneasy about channeling Obi-Wan Kenobi here and suggesting to "use the force." I'm too worried someone will blindly follow that advice.

```suggestion
						"cannot sign revision range %s..%s: "+
							"comparison status is %q "+
							"(target branch may have diverged)"
						targetHead, localHead, status,
```

pkg/promotion/runner/builtin/git_cloner.go
krancour: Or one of the sports, rather.

pkg/controller/git/base_repo.go
```diff
@@ -116,14 +118,19 @@ type User struct {
...
 
 // setupAuthor configures the git CLI with a default commit author.
 // Optionally, the author can have an associated signing key. When using GPG
 // signing, the name and email must match the GPG key identity. The directory
 // specified by homeDir is used as a virtual home directory for all commands
-// executed by this method.
-func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
+// executed by this method. Returns the GPG key fingerprint if a signing key
+// was imported, or an empty string otherwise.
+func (b *baseRepo) setupAuthor(homeDir string, author *User) (string, error) {
```
krancour: `setupAuthor()` returning a GPG key feels off. We should add a non-exported `gpgFingerprint` field to the `User` struct, set `author.gpgFingerprint` in here, and return `*author`.

```suggestion
func (b *baseRepo) setupAuthor(homeDir string, author *User) (user, error) {
```

pkg/controller/git/base_repo.go
```diff
@@ -116,14 +118,19 @@ type User struct {
 	// SigningKeyPath is an optional path referencing a signing key for
 	// signing git objects. Ignored if SigningKey is provided.
 	SigningKeyPath string
+	// SigningKeyTrustLevel is the GPG trust level to assign to the imported
+	// signing key. When set, setupAuthor marks the key with the
+	// corresponding ownertrust level after import.
+	SigningKeyTrustLevel SigningKeyTrustLevel
```
krancour, line 124: We should consider leaving this unexported (if we need to keep it at all), since the only key we plan to trust right now is the system-level one obtained from `gitUserFromEnv()`. We don't need a way for trust level to be dictated by something outside this package (at this time).

```suggestion
	signingKeyTrustLevel SigningKeyTrustLevel
```

krancour: Nevermind... just realized `gitUserFromEnv()` _is_ outside this package.

EronWright:
Addressed round 2 review feedback in 4245ff6b:

**Trust level changes:**
- Only the system-level signing key (from `GITCLIENT_SIGNING_KEY_PATH`) gets Ultimate trust
- Per-commit author keys are still imported into the repo keyring (for future whitelist support) but without any trust assignment — their commits report status `U` and are propagated without re-signing

**Other changes:**
- Refactored `setupAuthor` to return `*User` with an unexported `gpgFingerprint` field instead of a bare string
- Applied `sync.Once` pattern to `github-verified-push` init() per #5888
- Removed "use force to overwrite" hint from diverged-branch error message

**Tested in cluster:**

| Test Case | Author | Verified | Notes |
|-----------|--------|----------|-------|
| System key (no per-commit author) | Kargo | ✅ true | Key imported with `trustLevel="ultimate"`, signature status `G`, commit re-signed |
| Per-commit author with signing key | Test Author | ❌ false | Key imported with `trustLevel=""`, signature status `U`, commit propagated unsigned |

EronWright:
## Proposal: Simplify the `github-verified-push` trust model

After investigating the PullRebase signature laundering issue (#5732 context), I'd like to propose a simplification to how `github-verified-push` decides whether to App-sign commits.

### Problem with the current model

Today, `github-verified-push` uses GPG signature trust level to decide what to do with each commit:
- **"G" (trusted/system key)** → App signs the commit (Verified badge), credits original author via `Co-authored-by`
- **"U" (untrusted/author key)** → preserves original author+committer as-is (no Verified badge)

This is fragile. GPG trust level is a cryptographic artifact, not a semantic property of the promotion. Notably, `git pull --rebase` (used by `git-push` with `PullRebase: true`) silently re-signs commits with the system key, laundering "U" → "G" and replacing the committer identity — making the trust distinction unreliable.

### Proposed model

Separate the concerns:

| Concern | Signal | Decision |
|---------|--------|----------|
| **Signing authority** | Promotion provenance | Always App-sign — every commit replayed by `github-verified-push` is a promotion artifact |
| **Attribution** | Author identity field | Add `Co-authored-by` when author ≠ system identity ("Kargo \<no-reply@kargo.io\>") |
| **Integrity** | GPG verification | Detect tampering only ("B"/"R" → error); never gates signing or attribution |

In short: `github-verified-push` always produces Verified commits. Custom authors get credited via `Co-authored-by` based on the commit's author field (which survives rebase), not GPG status.

### Why this is stronger

- **Rebase-proof**: Author identity survives rebase; GPG status does not
- **Simpler**: One code path for all commits, no branching on signature status
- **Semantically grounded**: The question is "who authored this?" (answered by the author field) not "which key signed this?" (a git implementation detail)
- **Consistent**: If you're using `github-verified-push`, you want Verified commits — that should apply uniformly

EronWright:
Superseded by #5893, which squashes this branch into a single commit and simplifies the trust model as discussed above (all commits App-signed, Co-authored-by based on author identity, GPG for integrity only).

Issue #5700, opened 2026-02-03
ProjectConfig webhook receivers fail to sync: management-controller cache doesn't include project namespaces


IamStubborN:
## Description

ProjectConfig webhook receivers fail to sync with error:
```
error syncing WebhookReceiver "harbor-images": error getting Secret "harbor-webhook-secret" in namespace "my-project": unable to get: my-project/harbor-webhook-secret because of unknown namespace for the cache
```

This affects **all projects** with webhook receivers configured in ProjectConfig.

## Root Cause

In `cmd/controlplane/management_controller.go`, the management-controller's cache for `Secret` resources is limited to:
- `systemResourcesCfg.SourceNamespace` / `DestinationNamespace`
- `sharedResourcesCfg.SourceNamespace` / `DestinationNamespace`

```go
&corev1.Secret{}: {
    Namespaces: namespaceCacheConfigs,
},
```

**Project namespaces are NOT included in the cache.**

When `pkg/controller/management/projectconfigs/project_configs.go` calls `external.NewReceiver()`:
```go
receiver, err := external.NewReceiver(
    ctx,
    r.client,
    r.cfg.ExternalWebhookServerBaseURL,
    projectCfg.Name,
    projectCfg.Name, // Secret namespace is the same as the Project name/namespace
    receiverCfg,
)
```

The management-controller tries to read the secret from the project namespace, but the namespace is not in its cache, causing the "unknown namespace for the cache" error from controller-runtime.

## ClusterConfig Works

ClusterConfig webhook receivers work correctly because they use `r.cfg.SystemResourcesNamespace` which IS in the cache:
```go
receiver, err := external.NewReceiver(
    ctx,
    r.client,
    r.cfg.ExternalWebhookServerBaseURL,
    "",
    r.cfg.SystemResourcesNamespace, // This IS in the cache
    receiverCfg,
)
```

## Affected Versions

- Kargo v1.9.0
- Kargo v1.9.1
- Kargo v1.9.2

## Steps to Reproduce

1. Create a Kargo Project
2. Create a ProjectConfig with a webhook receiver (Harbor, GitHub, etc.)
3. Create the referenced secret in the project namespace
4. Check management-controller logs

## Expected Behavior

ProjectConfig webhook receivers should be able to read secrets from their project namespace.

## Possible Solutions

1. Add project namespaces to the management-controller's secret cache dynamically as Projects are created
2. Allow ProjectConfig webhook receivers to reference secrets from shared resources namespace
[...]

PR #5841, opened 2026-03-04
fix: project webhook receivers + support shared secrets

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5700


fuskovic:
Closing this to go with a simpler fix per: https://github.com/akuity/kargo/issues/5700#issuecomment-3998387638

PR #5843, opened 2026-03-04
fix: syncing of project webhook receivers

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5700

```
 faris@Fariss-MacBook-Pro  ~/go/src/github.com/fuskovic/wh-test-repo   main ±✚  \
> k describe projectconfig kargo-demo -n kargo-demo
Name:         kargo-demo
Namespace:    kargo-demo
Labels:       <none>
Annotations:  <none>
API Version:  kargo.akuity.io/v1alpha1
Kind:         ProjectConfig
Metadata:
  Creation Timestamp:  2026-03-04T16:24:46Z
  Generation:          1
  Resource Version:    2345
  UID:                 05570fe5-6668-44ec-8a3d-b5427ebcbf18
Spec:
  Webhook Receivers:
    Github:
      Secret Ref:
        Name:  gh-wh-secret
    Name:      gh-wh-receiver
Status:
  Conditions:
    Last Transition Time:  2026-03-04T16:24:46Z
    Message:               ProjectConfig is synced and ready for use
    Observed Generation:   1
    Reason:                Synced
    Status:                True
    Type:                  Ready
  Observed Generation:     1
  Webhook Receivers:
    Name:  gh-wh-receiver
    Path:  /github/f068a66328ead65a354b6219afd9dbcdc19b0d5f4714e770d2298bf94b7a8814
    URL:   http://localhost:30083/github/f068a66328ead65a354b6219afd9dbcdc19b0d5f4714e770d2298bf94b7a8814
```

### DIFF
--- a/pkg/webhook/external/server.go
+++ b/pkg/webhook/external/server.go
@@ -13,8 +13,9 @@ import (
 )
 
 type server struct {
-	cfg       ServerConfig
-	client    client.Client
+	cfg    ServerConfig
+	client client.Client
+	// project scoped secrets are not cached so we need to query the api-server directly
 	apiReader client.Reader
 }
 
--- a/pkg/webhook/external/server_test.go
+++ b/pkg/webhook/external/server_test.go
@@ -15,6 +15,8 @@ import (
 func TestNewServer(t *testing.T) {
 	testCfg := ServerConfig{}
 	testClient := fake.NewFakeClient()
+	// we pass the same client for both the client and api-reader since the
+	// fake client doesn't use the cache at all.
 	s, ok := NewServer(ServerConfig{}, testClient, testClient).(*server)
 	require.True(t, ok)
 	require.Equal(t, testCfg, s.cfg)


pkg/controller/management/projectconfigs/project_configs.go
```diff
@@ -240,6 +246,7 @@ func (r *reconciler) syncWebhookReceivers(
 		receiver, err := external.NewReceiver(
 			ctx,
 			r.client,
+			r.apiReader, // project scoped secrets are not cached so we need to query the api-server directly
```
krancour, line 250: Nit: This explanation would be more visible if you put this comment on the `apiReader` field of the `reconciler` type. It would save people from having to read this initialization code to understand why the reconciler hangs on to one of these.

pkg/controller/management/projectconfigs/project_configs_test.go
```diff
@@ -17,7 +17,8 @@ import (
 
 func TestNewReconciler(t *testing.T) {
 	testCfg := ReconcilerConfig{}
-	r := newReconciler(fake.NewClientBuilder().Build(), testCfg)
+	kc := fake.NewClientBuilder().Build()
+	r := newReconciler(kc, kc, testCfg)
```
krancour, line 22: Nit: A comment explaining why this expects two clients but is, for testing purposes, being passed the same one twice might be helpful.

pkg/webhook/external/receiver.go
```diff
@@ -96,10 +97,12 @@ func NewReceiver(
 		return nil, fmt.Errorf("error getting receiver factory: %w", err)
 	}
 	factory := reg.Value
+
+	// project secrets are not cached so we need to query the api-server directly
```
krancour: Whether it's true or not that we're querying the API server directly depends on what was passed to `NewReceiver()`, right? I think it would be helpful to clarify here that we _might_ be querying the API server directly (and why) depending on the implementation of `reader.Reader`.

pkg/webhook/external/receiver_test.go
```diff
@@ -109,6 +109,7 @@ func TestNewReceiver(t *testing.T) {
 			receiver, err := NewReceiver(
 				context.Background(),
 				testCase.client,
+				testCase.client,
```
krancour, line 113: Just another spot where I would clarify we we're passing the same client twice. If you don't already know why, this looks inexplicable.

pkg/webhook/external/server.go
```diff
@@ -13,18 +13,20 @@ import (
 )
 
 type server struct {
-	cfg    ServerConfig
-	client client.Client
+	cfg       ServerConfig
+	client    client.Client
+	apiReader client.Reader
```
krancour, line 19: Same as some other spot, I think this would be the most visible place to note why we hang on to one of these.

pkg/webhook/external/server_test.go
```diff
@@ -15,14 +15,14 @@ import (
 func TestNewServer(t *testing.T) {
 	testCfg := ServerConfig{}
 	testClient := fake.NewFakeClient()
-	s, ok := NewServer(ServerConfig{}, testClient).(*server)
+	s, ok := NewServer(ServerConfig{}, testClient, testClient).(*server)
```
krancour, line 20: Same as some other spots, this looks unusual if not explained.

Issue #5552, opened 2026-01-05
Better pull request URL for Azure DevOps


zeraye:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

After successful promotion step show more user friendly pull request URL for Azure DevOps. Currently pull request URL look like this (for legacy URL, for modern URL it will be analogous):
`https://<orgName>.visualstudio.com/<projectID>/_apis/git/repositories/<repositoryID>/pullRequests/<pullRequestID>`.

I want this URL to look like this:
`https://<orgName>.visualstudio.com/DefaultCollection/<orgName>/_git/<repoName>/pullrequest/<pullRequestID>`.

First link redirect user to some JSON, second one redirect user to Azure DevOps UI with pull request.

# Motivation

It's much better user experience to have one click option with redirect to Azure DevOps UI with pull request instead of raw JSON.

# Implementation

I can create PR if this feature will be accepted. Example implementation:

```diff
diff --git i/pkg/gitprovider/azure/azure.go w/pkg/gitprovider/azure/azure.go
index d5afaed3..5729f47f 100644
--- i/pkg/gitprovider/azure/azure.go
+++ w/pkg/gitprovider/azure/azure.go
@@ -292,9 +292,11 @@ func convertADOPullRequest(pr *adogit.GitPullRequest) (*gitprovider.PullRequest,
 		return nil, fmt.Errorf("no last merge source commit found for pull request %d", ptr.Deref(pr.PullRequestId, 0))
 	}
 	mergeCommit := ptr.Deref(pr.LastMergeCommit, adogit.GitCommitRef{})
+	prNumber := int64(ptr.Deref(pr.PullRequestId, 0))
+
 	return &gitprovider.PullRequest{
-		Number:         int64(ptr.Deref(pr.PullRequestId, 0)),
-		URL:            ptr.Deref(pr.Url, ""),
+		Number:         prNumber,
+		URL:            fmt.Sprintf("%s/pullrequest/%d", ptr.Deref(pr.Repository.WebUrl, ""), prNumber),
 		Open:           ptr.Deref(pr.Status, "notSet") == "active",
 		Merged:         ptr.Deref(pr.Status, "notSet") == "completed",
 		MergeCommitSHA: ptr.Deref(mergeCommit.CommitId, ""),
```

singhvibhanshu:
[...]

Issue #5450, opened 2025-12-05
Configurable wait time in git-wait-for-pr


AmyJoT:
__Note:__ If you're requesting a Helm chart option that may be very niche and not useful to the community at large, please consider using Kustomize to apply "last mile" tweaks to the output of `helm template` to suit your needs instead.

# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Configurable wait time for `git-wait-for-pr`

# Motivation

The 5 minute delay is slow for us, in our intended implementation.  We would like to be able to configure this for more flexibility.

This was mentioned here, and is something we would benefit from

https://github.com/akuity/kargo/issues/3303

aslafy-z:
We’re also running into this. The current reconcile duration feels quite long on our side as well, and it becomes noticeable in environments with multiple stages and frequent promotions.

AmyJoT:
Any thoughts on this? It would be a really impactful feature for us

morremeyer:
You can technically configure the wait time by setting the timeout, since there is logic that catches this case in https://github.com/akuity/kargo/blob/dd4c1b91c2008e37f516c0cc90b819f988f11cbe/pkg%2Fcontroller%2Fpromotions%2Fpromotions.go#L719

To do so, configure the step as follows:

```
retry:
  timeout: 3m
```

However, what I **think** is needed here is not a configurable timeout, but a configurable polling interval, so that kargo checks the PR every e.g. 1 minute until the (configured) timeout is reached.

There should be a workaround as documented in https://docs.kargo.io/user-guide/reference-docs/promotion-templates#step-retries, however this is not working right now due to https://github.com/akuity/kargo/issues/5784.

krancour:
@morremeyer, I'm afraid that guidance is incorrect, per my comment on that thread: https://github.com/akuity/kargo/issues/5784#issuecomment-3952948061

> However, what I think is needed here is not a configurable timeout

That is already what @AmyJoT was asking for.

Responding to the proposal, however, there are dozens of issues, either open or closed, about adjusting polling/retry/reconciliation intervals. In almost all cases, our preference is to pursue other optimizations _first_.

[...]

Issue #4969, opened 2025-08-31
Make Stage controller requeue delay configurable


omerzifman:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.
* [x] I've pasted logs, if applicable.

# Proposed Feature

Make the reconcile requeue delay configurable for these controllers:

* `cluster_config`
* `stage`
* `projects`
* `project_config`

Currently the controllers uses a hard-coded five-minute delay. for example in `internal/controller/stages/regular_stages.go` at lines 383–385:

```go
// Otherwise, requeue after a delay.
// TODO: Make the requeue delay configurable.
return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
```

Expose a single config knob per controller (env var and Helm value) with a default of 5m.

# Motivation

Operators cannot tune reconciliation frequency without rebuilding or patching.
Configurable intervals allow faster feedback for time-sensitive workloads and lower resource usage for quiet ones, without using custom patched versions.

mrfelton:
This - which seems to have some crossover with https://github.com/akuity/kargo/issues/5162 - would help to make Kargo promotions more responsive. An issue that we have curerntly is that any steps that do effective polling (e.g. wait on PR merge, or http steps that use polling) take too long to detect changes in the remote systems that they are awaiting on.

Ideally, promotion flows should be responsive, with near realtime detection to changes in these types of polling tasks.

As a basic workaround, one can hit the refresh button on a promotion which seems to force an immediate re-request. But that's not a workable solution in the wild.

As a better workaround, we're considering implementing a cron job that runs every few seconds that basically just calls `kargo refresh stage` on the stages that we need to be more responsive. Though I'm not sure what the side effects of that might be.

Ideally, kargo would make this configurable either per stage, per promotion, or per step so that we can tailor things a little more to our needs.

krancour:
We know this is something that people want -- so much so that I'm pinning it for visibility and a lesser chance of more duplicate issues being opened. This is probably going to happen eventually. To be transparent about why we're not prioritizing this more highly, we anticipate significant performance impacts when everyone inevitably goes and cranks the interval way down.

There are many other issues open having to do with performance improvements and, in general, our goal is improving on individual factors that are making people _want_ to crank the interval way down so that when this feature is eventually added, not everyone will feel compelled to immediately start re-reconciling Promotions every 10s (for instance) and bring their cluster to its knees.

[...]

Issue #3303, opened 2025-01-17
Delay in git-wait-for-pr Step Execution


evantissot:
# Checklist

* [X] I've searched the issue queue to verify this is not a duplicate feature request.
* [X] I've pasted the output of `kargo version`, if applicable.
* [X] I've pasted logs, if applicable.

# Motivation

I am experiencing an issue with the `git-wait-for-pr` step in my workflow, which is causing a delay of 5 minutes before it successfully completes. Here is a brief overview of my stage:
- I have a `git-open-pr` step. This PR is automerge with a GitHub automation in place.
- I have a `git-wait-for-pr` step

Problem:
The `git-wait-for-pr` step seems to take 5 minutes to succeed. Based on my debugging, I suspect that the issue arises because when the `git-wait-for-pr` step is first executed, the PR has not yet been merged. As a result, the step is requeued and executed again 5 minutes later, at which point it successfully completes.

# Proposed Feature
- Introduce a dedicated step to sleep or wait before executing the git-wait-for-pr step. This would ensure that the step succeeds on its first run by allowing enough time for the PR to be merged.
- Allow users to configure the 5-minute requeue interval that is currently [hard-coded](https://github.com/akuity/kargo/blob/03632de63754c16552f18038a729f831d4b39f15/internal/controller/promotions/promotions.go#L414-L421). This would enable more flexibility in adjusting the wait time according to specific workflow needs.

# Suggested Implementation

krancour:
I'm not opposed to making the retry interval configurable, but it would provide no guarantee of the step actually being retried at the specified interval. I think the consequences of configuring this poorly (increased back-pressure on the queue with "busy work," for instance) may make it better suited for adjustment at the system-level by an operator than at the Promotion or step level by a user.

i.e. I don't know how _well_ it solves your particular problem.

> Introduce a dedicated step to sleep or wait...

Controllers support a finite number of concurrent reconciliations, so a step of this nature can become a major bottleneck. This is the same reason that the git-wait-for-pr step (and the http step, too) doesn't use its own, internal retry loop. Imagine a whole slew of pending Promotions being fully blocked because a small handful of Promotions are just _waiting_ for a PR to be merged.

A more reliable solution would be something like #647 so that a Promotion can be "jostled" after a PR it's waiting on has been merged.

Short term workaround, in case it helps even a little:

Reconciling the Stage (by refreshing) will put its active Promotion back on the queue immediately.

hiddeco:
[...]

PR #5852, opened 2026-03-04
feat: wake up Promotions waiting on PRs via pull_request webhooks

### LINKED ISSUE #5834: wake up Promotions waiting on git-wait-for-pr via pull_request webhooks

# Checklist

# Proposed Feature

**Related:** #5450, #3303, #647, #4969

## Problem

When a Promotion is paused on a `git-wait-for-pr` step, it polls the GitHub
API on the default 5-minute requeue interval. For users with auto-merge or
fast review cycles, this delay dominates the promotion pipeline. As noted in
#5450, webhooks are the preferred path to reducing this latency.

## Proposal

Extend the existing GitHub webhook receiver to handle `pull_request` events
and immediately refresh matching Promotions, following the same patterns
already established in the codebase:

1. **New field indexer** (`RunningPromotionsByPullRequestField`) — indexes
   running Promotions by `{normalizedRepoURL}:{prNumber}`, following the
   `RunningPromotionsByArgoCDApplicationsField` pattern. Evaluates template
   expressions in step config using the Promotion's shared state, same as the
   ArgoCD indexer does for app names.

2. **`pull_request` event handling in `githubWebhookReceiver`** — on `closed`
   action (covers both merged and closed-without-merge), extract repo URL and
   PR number from the event payload, query the index, and call
   `api.RefreshObject()` on each matching Promotion. This is the same
   mechanism the receiver already uses for Warehouse refresh on `push` events.

3. **No changes to the Promotion controller or `git-wait-for-pr` step** — the
   existing `RefreshRequested` predicate triggers immediate reconciliation,
   and the step runner re-checks PR status as it already does today.

Polling remains unchanged as a reliability fallback. Users who don't add
`pull_request` to their GitHub webhook event types see no behavior change.

## Scope

- **GitHub only** for the initial contribution. The architecture (index +
  refresh) extends naturally to GitLab, Bitbucket, Azure DevOps, and Gitea
  in follow-up work.
- **`git-wait-for-pr` only.** `git-merge-pr` with `wait=true` has the same
  polling delay and could be added to the index in a follow-up.
- **No CRD changes.** The index is built from existing Promotion spec/status
  fields.

## User-facing change

Users add `pull_request` to the event types on their existing GitHub webhook
configuration (a one-time change in GitHub's UI). No changes to Kargo CRDs,
ProjectConfig, or ClusterConfig are needed.

## Offer to contribute

I have time this week and would be happy to implement this. I have a detailed
design based on the investigation of the indexer and webhook subsystems and
can have a PR up quickly once the approach is agreed on.

Before writing code, I wanted to confirm:
1. Does this approach align with what you had in mind for #5450?
2. Any preferences on where the Promotion refresh helper lives — alongside
   `refreshWarehouses` in `refresh.go`, or in a separate file?
3. Should the index registration go in both the webhook server
   (`cmd/controlplane/external_webhooks.go`) and the promotion controller
   (`pkg/controller/promotions/promotions.go`), or webhook server only?

# Motivation

This is a major bottleneck in my team's promotion pipeline.

DavidS-ovm:
## Summary

When a GitHub `pull_request` event with a `closed` action arrives, Kargo can now
immediately re-reconcile any running Promotion whose current `git-wait-for-pr`
step is waiting on that PR — reducing detection latency from ~5 minutes (polling)
to near-instant. Polling remains as a reliability fallback.

Closes #5834 and https://github.com/akuity/kargo/issues/5552. Related: #3303, #647, #4969, #5450

## Design

This follows the approach discussed in
https://github.com/akuity/kargo/issues/5450#issuecomment-3953002302 —
extending the existing webhook infrastructure rather than introducing a new
mechanism.

**Indexer** — A new `RunningPromotionsByPullRequestField` indexer maps
`{normalizedRepoURL}:{prNumber}` to running Promotions. It follows the
`RunningPromotionsByArgoCDApplications` pattern: filtering to running
Promotions, iterating steps up to `CurrentStep`, evaluating template
expressions for `repoURL` and `prNumber`, and normalizing Git URLs. Registered
in the external webhook server.

**Webhook handler** — The GitHub receiver now accepts `pull_request` events.
On `closed` action (merged or not), it queries the index with both the clone
URL and SSH URL from the event payload, de-dupes matching Promotions by
`ObjectKey`, and calls `api.RefreshObject` on each. Non-closed actions return
200 OK with no side effects. The new `refreshPromotions` helper in `refresh.go`
mirrors `refreshWarehouses` and is provider-agnostic so future GitLab/Bitbucket
handlers can reuse it.

**Why only `closed`?** A merged PR fires `closed` with `merged: true`; the
`git-wait-for-pr` step detects the merge on next reconcile and succeeds. A PR
closed without merge fires `closed` with `merged: false`; the step detects
closure and fails with a terminal error. All other actions (`opened`,
`synchronize`, etc.) are irrelevant to the step's outcome.

## User-facing changes

- Users must add `pull_request` to the event types on their GitHub webhook
  configuration (one-time change). Users who do not make this change experience
  no behavior change.
- No new CRD fields, no new CLI flags, no breaking changes.

## Documentation

- GitHub webhook receiver page: added `pull_request` to supported events, added
  Promotion "refreshing" explanation, updated event selection instructions for
  all three setup methods (single repo, org, GitHub App)
- `git-wait-for-pr` step reference: added a tip about webhook acceleration
- Webhook receivers index, Working with Warehouses guide, and Cluster
  Configuration guide: broadened webhook framing to mention Promotion refresh

## Test plan

- 15 table-driven unit tests for the indexer covering: expression evaluation,
  literal values, non-`git-wait-for-pr` steps ignored, non-running Promotions
  excluded, multiple PRs across steps, URL normalization (clone URL, SSH URL),
  template expressions with outputs and vars, steps beyond `CurrentStep`
- 5 unit tests for the webhook handler covering: closed+merged refreshes
  Promotion, closed+not-merged refreshes Promotion, non-closed action returns
  200, no matching Promotions returns 200 with 0 refreshed, SSH URL form matches
- All existing tests pass unchanged

Signed-off-by: David Schmitt <david@overmind.tech>


cmd/controlplane/external_webhooks.go
```diff
@@ -125,6 +125,20 @@ func (o *externalWebhooksServerOptions) run(ctx context.Context) error {
 		return fmt.Errorf("error registering project configs by webhook receiver path indexer: %w", err)
 	}
 
+	if _, err = cluster.GetCache().GetInformer(ctx, &kargoapi.Stage{}); err != nil {
+		return fmt.Errorf("error registering Stage informer: %w", err)
+	}
```
krancour: What was the reason for these lines?

pkg/controller/promotions/promotions.go
```diff
@@ -117,6 +117,16 @@ func SetupReconcilerWithManager(
...
+	// Index running Promotions by pull requests they are waiting on
+	if err := kargoMgr.GetFieldIndexer().IndexField(
+		ctx,
+		&kargoapi.Promotion{},
+		indexer.RunningPromotionsByPullRequestField,
+		indexer.RunningPromotionsByPullRequest(ctx, kargoMgr.GetClient()),
+	); err != nil {
+		return fmt.Errorf("index running Promotions by pull request: %w", err)
+	}
+
```
krancour: Why does the controller need this index? It makes sense that the external webhooks server would need it, but I can't spot the rationale for it here.

pkg/indexer/indexer.go
```diff
@@ -299,6 +300,185 @@ func RunningPromotionsByArgoCDApplications(
...
 }
 
+// RunningPromotionsByPullRequest returns a client.IndexerFunc that indexes
+// running Promotions by the pull requests they are waiting on via
+// git-wait-for-pr steps. Index keys have the form
+// "{normalizedRepoURL}:{prNumber}".
+func RunningPromotionsByPullRequest(
+	ctx context.Context,
+	cl client.Client,
+) client.IndexerFunc {
```
krancour: We might be doing a lot more than is necessary to build this index. I _think_ all the expression evaluation stuff probably isn't necessary. The PR URL is in a git-open-pr step's output. That value will always be static; not an expression. Isn't that sufficient information for inferring a relationship between a PR and a Promotion?

pkg/webhook/external/github.go
krancour, line 1: It would be good if we didn't do this only for GitHub, but for all git hosting providers that support a similar event.

pkg/webhook/external/refresh.go
```diff
@@ -186,6 +186,96 @@ func refreshWarehouses(
 	)
 }
 
+// refreshPromotions refreshes all running Promotions that are waiting on a
+// pull request matching any of the given repository URLs and PR number. If the
+// project is non-empty, only Promotions in that namespace are considered.
+// Note: Callers are responsible for normalizing the provided repository URLs.
+func refreshPromotions(
```
krancour: I would consider renaming this to `refreshPromotionsByPR()`.

I know for Warehouses, we just called it `refreshWarehouses()`, but that was a little different because the args it takes are all things instrinsic to a Warehouse -- what it's subscribed to. A Warehouse that needs to be refreshed _for any reason_ could be refreshed by supplying that information.

Here, the arguments include something very use-case-specific -- a PR number. I can think of a lot of _other_ reasons to refresh a Promotion, which we're likely to get to eventually, so the more specific function name here would pay off later.

pkg/webhook/external/github.go
```diff
@@ -209,6 +211,21 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
...
 
+		case *gh.PullRequestEvent:
+			if e.GetAction() != "closed" {
+				xhttp.WriteResponseJSON(w, http.StatusOK, nil)
+				return
+			}
+			prNumber := e.GetNumber()
+			repoURLs = []string{
+				urls.NormalizeGit(e.GetRepo().GetCloneURL()),
+				urls.NormalizeGit(e.GetRepo().GetSSHURL()),
```
krancour: I don't think we'd need the SSH URL.

krancour:
I have this an initial look, but wondering if @fuskovic would be so kind as to take over as the sherpa here.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-wait-for-pr.md
```diff
@@ -9,6 +9,17 @@ description: Waits for a specified open pull request to be merged or closed.
 closed. This step commonly follows a [`git-open-pr` step](git-open-pr.md)
 and is commonly followed by an `argocd-update` step.
 
+:::tip[Accelerate with webhooks]
+
+By default, Kargo polls the Git provider every few minutes to check whether the
+PR has been merged or closed. If you configure a
+[webhook receiver](../80-webhook-receivers/github/index.md) and enable
```
fuskovic:
```suggestion
[github webhook receiver](../80-webhook-receivers/github/index.md) and enable
```

cmd/controlplane/external_webhooks.go
DavidS-ovm: This was part of the attempt to evaluate the stage inputs. But see below for my fix.

pkg/controller/promotions/promotions.go
DavidS-ovm: Removed. The important index is the one in `external_webhooks.go`

pkg/indexer/indexer.go
DavidS-ovm: This was an interesting one. I was under the impression that running steps didn't have outputs yet (but then I only looked at our 1.8 install, and maybe just the wrong steps, or it doesn't show up in the app). Since then I've updated to 1.9 and looked at our `git-wait-for-pr` step, and indeed, the required values are right there! So I removed the whole index and replaced it with a small check of those outputs.

This might have a small window where github is fast enough to deliver an event but the step hasn't yet updated - maybe if people are doing weird stuff in the stage between opening the PR and waiting for the merge, but in that case either the task already sees the merged PR and proceeds, or - worst case - the polling loop will catch it later.

pkg/webhook/external/github.go
DavidS-ovm: I agree that that would be good.

I don't have infrastructure to test, nor budget to support anything other than what my team uses.

pkg/webhook/external/refresh.go
DavidS-ovm: done.

pkg/webhook/external/github.go
DavidS-ovm: I saw that those are getting deprecated anyways? I just removed it then.

DavidS-ovm:
@krancour @fuskovic finally managed to take the time to go through the feedback. Thanks for your kind words and expertise. The code change looks much better now thanks to that!

fuskovic (changes requested):
I did some superficial research and found Gitea and Gitlab also emit events that make supporting this possible. I'd love to get this into `1.10` but not without support for gitea and gitlab. If you don't have accounts for those providers it's ok because I do from when I implemented them. If you can add support for the other git providers I can help end-to-end test it.

fuskovic:
@DavidS-ovm I can alternatively add support for the others to your branch as well.

DavidS-ovm:
> @DavidS-ovm I can alternatively add support for the others to your branch as well.
 
This would be very much appreciated! I'm very much at the limits of what my company can donate to the cause.

fuskovic:
> > @DavidS-ovm I can alternatively add support for the others to your branch as well.
> 
> This would be very much appreciated! I'm very much at the limits of what my company can donate to the cause.

@DavidS-ovm Can you give me push access

DavidS-ovm:
@fuskovic invite is :airplane:

fuskovic:
> @fuskovic invite is ✈️

@DavidS-ovm  Looks like you sent access to `DavidS-ovm/kargo` but the branch is on `overmindtech/kargo`

DavidS-ovm:
@fuskovic how did I screw that up so badly :upside_down_face: I've invited you to the overmindtech/kargo repo now.

fuskovic:
> @fuskovic how did I screw that up so badly 🙃 I've invited you to the overmindtech/kargo repo now.

No worries. I have the other provider implementations pushed. Working on getting everything end-to-end tested soon.

fuskovic:
@krancour @DavidS-ovm 

There were a couple bugs.

1. The indexer didn't account for variable/expression expansion e.g. it was creating a cache key like: `${{ vars.gitRepo }}:7`. So I fixed that and had to account for both global promotion vars as well as `promotionTask` vars.

2. RBAC permissions for the external-webhooks-server cluster role for promotions and stages e.g.

```
2026-04-01T21:34:27Z	ERROR	external/refresh.go:253	failed to refresh Promotion: patch annotation: admission webhook "promotion.kargo.akuity.io" denied the request: Internal error occurred: Promotion.kargo.akuity.io "test.01kn5fey6mqq0q8dg7d6g1jwbr.e1ef7ae" is forbidden: subject "system:serviceaccount:kargo:kargo-external-webhooks-server" is not permitted to update Promotions for Stage "test"
```

Both of those are fixed now and the Github webhook receiver is processing promotion refreshes for pull request merges successfully.

### Github - promotion refresh success waiting for PR close event via merge

https://github.com/user-attachments/assets/3fa9c7db-2846-4646-b641-7c20e35c8b35

However, I noticed if while waiting, the PR is closed manually e.g. no merge, the promotions itself fails. This appears to be the expected behavior according to the `wait-for-pr` step.

I'll e2e test the gitlab and gitea implementations tomorrow.

charts/kargo/templates/external-webhooks-server/cluster-role.yaml
```diff
@@ -16,6 +16,21 @@ rules:
...
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
+  - kargo.akuity.io
+  resources:
+  - stages
+  verbs:
+  - promote
```
krancour: This doesn't look right. When does the webhook server need to initiate a promotion?

fuskovic: Kargo's admission webhook (`ValidateUpdate`) requires the `promote` verb on `stages`.

pkg/indexer/indexer.go
```diff
@@ -299,6 +300,89 @@ func RunningPromotionsByArgoCDApplications(
...
+		}
+		cfgMap := map[string]any{}
+		if err := json.Unmarshal(step.Config.Raw, &cfgMap); err != nil {
+			continue
+		}
+		repoURL, _ := cfgMap["repoURL"].(string)
+		if repoURL == "" {
+			continue
+		}
+		// repoURL may be an unevaluated template expression (e.g.
```
krancour: I cannot recall if I asked about this previously, but do we really need to do this? Is it possible to determine the repo URL (which can be normalized afterwards) from the `pr.url` output?

pkg/webhook/external/gitea.go
```diff
@@ -126,49 +127,76 @@ func (g *giteaWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc {
...
-		repoURLs := []string{urls.NormalizeGit(payload.Repo.URL)}
-
-		var changedFiles []string
-		if payload.TotalCommits > len(payload.Commits) {
-			logger.Info(
-				"push event commits were truncated by Gitea; "+
-					"skipping path filtering for this event",
-				"totalCommits", payload.TotalCommits,
-				"receivedCommits", len(payload.Commits),
+		switch eventType {
```
krancour, line 130: I would potentially look toward breaking this up into smaller methods and just routing to the appropriate one based on event type.

pkg/webhook/external/github.go
```diff
@@ -211,6 +213,20 @@ func (g *githubWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 			)
 			return
 
+		case *gh.PullRequestEvent:
```
krancour, line 156: Similar to here: https://github.com/akuity/kargo/pull/5852/changes#r3029158836

pkg/webhook/external/gitlab.go
```diff
@@ -118,6 +118,21 @@ func (g *gitlabWebhookReceiver) getHandler(requestBody []byte) http.HandlerFunc
 		}
 
 		switch e := event.(type) {
+		case *gl.MergeEvent:
```
krancour, line 121: Similar to here: https://github.com/akuity/kargo/pull/5852/changes#r3029158836

pkg/indexer/indexer.go
fuskovic: The `prURL` usually contains a suffix like `/pulls/` but this varies from provider to provider. I feel like implementing custom string parsing logic for each provider to resolve the `repoURL` is more brittle than the current solution.

charts/kargo/templates/external-webhooks-server/cluster-role.yaml
krancour: Oh. The patching of the Promotion triggers that. ☹️ 

The external webhooks server is an internet-facing component. In terms of threat-modeling, the ability to kick off promotions to any Stage throughout all Projects, which is what this grants, isn't a desirable capability for an internet-facing component to have.

We need to find a less risky way to get past this. This is a blocker, unfortunately.

pkg/indexer/indexer.go
krancour: I can see where that may be brittle. I think the existing way here isn't great either. I know it's modeled after how the "running promos by app" index was built, but I hate that equally.

It's still possible there's a third option here.

Do the pull request merged payloads for each provider happen to contain the PR's URL? See where I am going with this?

If the PR URL were the key in the index (instead of repo URL + pr number), this would all be very straightforward. But this definitely depends on what is available to us in the payloads.

fuskovic: | Platform | Event | PR URL Field | Example URL |
|---|---|---|---|
| **GitHub** | `pull_request` (closed + merged) | `pull_request.html_url` | `https://github.com/org/repo/pull/42` |
| **GitLab** | `merge_request` hook | `object_attributes.url` | `https://gitlab.com/org/repo/-/merge_requests/42` |
| **Bitbucket Cloud** | `pullrequest:fulfilled` | `pullrequest.links.html.href` | `https://bitbucket.org/org/repo/pull-requests/42` |
| **Bitbucket Server** | `pr:merged` | `pullRequest.links[0].href` | `https://bitbucket.example.com/projects/ORG/repos/repo/pull-requests/42` |
| **Azure DevOps** | `git.pullrequest.merged` | `resource._links.web.href` | `https://dev.azure.com/org/project/_git/repo/pullrequest/42` |
| **Gitea** | `pull_request` (closed + merged) | `pull_request.html_url` | `https://gitea.example.com/org/repo/pulls/42` |

That's do-able but I like the current approach better tbh. 

Why implement a custom resolution strategy for each of the above existing git providers ( and do the same for any future providers) when we can use a uniform solution that works the same way for every future and existing provider while having less code to maintain?

charts/kargo/templates/external-webhooks-server/cluster-role.yaml
fuskovic: Can hash this one out offline

pkg/indexer/indexer.go
krancour: This actually looks encouraging.

> Why implement a custom resolution strategy for each of the above existing git providers ( and do the same for any future providers) when we can use a uniform solution that works the same way for every future and existing provider while having less code to maintain?

There's potentially nothing custom about this. We need to look more closely at how the providers populate `pr.url` in the output of the `git-wait-for-pr` step's output. If it happens to come straight _from_ the provider, it should always match the URL in the PR merged event payload.

krancour: I just looked and good news: All providers populate `gitprovider.PullRequest`'s `URL` field directly with URLs from a get PR API request. You'd need to confirm, but I think those are going to be exact matches for what's in the PR merged event payloads received by the webhooks server.

fuskovic:
@krancour I think this is ready for another look

Pushed:
- support for Azure DevOps + BitBucket
- security updates we discussed offline
- indexer refactor to use `prUrl`
- use `webUrl` instead of `remoteUrl` for Azure DevOps

Demos for the following table are located [here](https://drive.google.com/drive/folders/1ymaEjN6qOwR-wXfpLzZAz0lfOAz6adlG) (maintainer access only)

| git-provider | test case | result |
|---|---|---|
| GitHub | merge | ✅  |
| GitHub | close pr with no merge |  ✅  |
| GitLab | merge |  ✅ |
| GitLab | close pr with no merge | ✅  |
| Gitea | merge |  ✅ |
| Gitea | close pr with no merge | ✅  |
| Azure DevOps | merge |  ✅ |
| Azure DevOps | close pr with no merge |  ✅ |
| Bitbucket | merge |  ✅ |
| Bitbucket | close pr no merge |  ✅ |

Note:

- The demo for Azure DevOps in the drive folder is from after making the `remoteUrl` -> `webUrl` change.
- Azure DevOps requires configuring two separate webhooks. One to handle PR updated -> Abandoned and the other for Merge attempted -> Succeeded

fuskovic (approved):
Since all my end-to-end tests passed going to approve but will hold off on merging until final approval from @krancour

fuskovic:
Updated the generic webhook receiver (and docs) to support the new index + refreshing promotions by prURL

simonwalton:
Looking forward to this :)

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/bitbucket/index.md
```diff
@@ -4,9 +4,22 @@ sidebar_label: Bitbucket
...
 # Bitbucket Webhook Receiver
 
-The Bitbucket webhook receiver responds to `repo:push` events originating from
-Bitbucket repositories by _refreshing_ all `Warehouse` resources subscribed to
-those repositories.
+The Bitbucket webhook receiver responds to `repo:push`,
+`pullrequest:fulfilled`, and `pullrequest:rejected` events originating from
+Bitbucket Cloud repositories, and `repo:refs_changed`, `pr:merged`, and
+`pr:declined` events originating from Bitbucket Server and Data Center
+repositories.
```
krancour, line 11: Since `git-open-pr` and `git-wait-for-pr` don't support BB Server/Data Center yet, would handling of `pr:merged` and `pr:declined` effectively be no-ops?

pkg/gitprovider/azure/azure.go
```diff
@@ -391,9 +391,17 @@ func convertADOPullRequest(pr *adogit.GitPullRequest) (*gitprovider.PullRequest,
 		return nil, fmt.Errorf("no last merge source commit found for pull request %d", ptr.Deref(pr.PullRequestId, 0))
 	}
 	mergeCommit := ptr.Deref(pr.LastMergeCommit, adogit.GitCommitRef{})
+	var webURL string
+	if pr.Repository != nil {
+		webURL = ptr.Deref(pr.Repository.WebUrl, "")
+	}
```
krancour, line 397: nil check is built into `ptr.Deref`


```suggestion
	webURL := ptr.Deref(pr.Repository.WebUrl, "")
```

pkg/webhook/kubernetes/promotion/webhook.go
```diff
@@ -344,6 +350,13 @@ func (w *webhook) authorize(
 		)
 	}
 
+	// The external webhooks server is trusted to create Promotions on behalf
+	// of webhook callers. Skip the "promote" verb check for this subject.
```
krancour, line 354: This doesn't seem accurate. Aren't we just trusting it to refresh promotions?

pkg/webhook/kubernetes/config.go
```diff
@@ -18,6 +18,13 @@ type Config struct {
 	// management controller (not the API server or other controlplane
 	// components) should be permitted to act.
 	ManagementControllerUsername string `envconfig:"MANAGEMENT_CONTROLLER_USERNAME"`
+	// ExternalWebhooksServerUsername is the exact username (typically a service
+	// account name) of the external webhooks server. When an admission request
+	// originates from this subject, the "promote" verb authorization check is
+	// bypassed, as the external webhooks server is permitted to create
+	// Promotions on behalf of webhook callers without holding that permission
+	// itself.
```
krancour, line 26: Same as [here](https://github.com/akuity/kargo/pull/5852/changes#r3114511896)

pkg/webhook/kubernetes/promotion/webhook.go
```diff
@@ -344,6 +350,13 @@ func (w *webhook) authorize(
...
+	if w.externalWebhooksServerUsername != "" &&
+		req.UserInfo.Username == w.externalWebhooksServerUsername {
```
krancour, line 356: Is the check that this isn't empty necessary?

When would `req.UserInfo.Username == w.externalWebhooksServerUsername` when `w.externalWebhooksServerUsername` _is_ empty?

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/bitbucket/index.md
fuskovic: Some options include:

1. Remove it as dead code
2. No-op
3. Return an error indicating that this feature isn't supported for BB Server/Data Center.

I don't think we'd ever actually get to number 3 since our BB gitprovider ( and transitively the `git-open-pr` step ) doesn't currently support Server/Data Center yet.

Thoughts?

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/generic.md
```diff
@@ -10,12 +10,13 @@ user-defined set of resources when it does.
...
-lack dedicated webhook receiver implementations. Since this component
-effectively enables imperatively refreshing a `Warehouse` from any external
-process, other uses are possible and practical.
+Currently, these actions are limited to "refreshing" `Warehouse` and
+`Promotion` resources. Refreshing a `Warehouse` triggers its artifact
+discovery process, making this useful for responding to "push" events from
+artifact repositories that lack dedicated webhook receiver implementations.
+Refreshing a `Promotion` wakes up a running `Promotion` that is waiting on a
+pull request, which is useful for integrating with external version control
+systems that lack a dedicated webhook receiver.
```
krancour: Non-blocking, but highly recommended: I'd rework this a bit to be more generic. Refreshing a Promotion enqueues it for reconciliation whether it's waiting on a PR or not. And since this is the generic webhook receiver, the potential actually exists here that you could wake a promotion based on something other than a PR.

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/generic.md
```diff
@@ -14,9 +14,10 @@ Currently, these actions are limited to "refreshing" `Warehouse` and
 `Promotion` resources. Refreshing a `Warehouse` triggers its artifact
 discovery process, making this useful for responding to "push" events from
 artifact repositories that lack dedicated webhook receiver implementations.
-Refreshing a `Promotion` wakes up a running `Promotion` that is waiting on a
-pull request, which is useful for integrating with external version control
-systems that lack a dedicated webhook receiver.
+Refreshing a `Promotion` enqueues a running `Promotion` for reconciliation,
+which is useful for waking up a `Promotion` that is waiting on an external
+signal — such as a pull request merge from a version control system that lacks
+a dedicated webhook receiver.
```
krancour, line 20: Love how this was reworded.

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/generic.md
```diff
@@ -282,10 +286,31 @@ actions:
...
+  - action: Refresh
+    whenExpression: "request.header('X-Event-Type') == 'pull_request'"
+    targetSelectionCriteria:
+      - kind: Promotion
+        indexSelector:
+          matchIndices:
+            - key: pullRequestURL
+              operator: Equals
+              value: "${{ request.body.pull_request.html_url }}"
+```
```
krancour: I think an example along these lines only works if the PR was opened against one of the supported git hosting providers (otherwise, idk how the Promo would exist in that index). All of the supported platforms have built-in receivers. So, more than likely, no one is ever going to do something like this, and if they try, I don't think they'll get it to work.

What would you think about reworking the example in a way that's a bit more generic? Pull the name of a Promo out of the body of a request and refresh that Promo.

This would actually be an extra-ordinarily useful pattern. If a promotion process had, for instance, used the http step to trigger some external process, it might tell that external process its name. The external process could resume the Promo via webhook once it completes. This could unblock Promos that are waiting on the external process, polling periodically to ask if it's complete.

krancour: LGTM after this.

pkg/gitprovider/azure/azure.go
fuskovic: `WebUrl` is the pointer being derefed so this can still panic if `pr.Repository` is nil

fuskovic: There are tests in the existing suite that demonstrate this point when this suggestion is applied.

fuskovic:
@krancour this is good for another look. There's a couple of unresolved comments up there I want to bring to your attention if you can have a look at those.

pkg/gitprovider/azure/azure.go
krancour: Sorry. I read it wrong. I somehow saw `if pr.Repository.WebUrl != nil {`

Carry on!

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/generic.md
```diff
@@ -210,6 +210,32 @@ spec:
                 name: "${{ normalizeGit(request.body.repository.name) }}"
 ```
 
+The following example depicts `targetSelectionCriteria` that selects a running
+`Promotion` by name sourced dynamically from the request body. This pattern is
+useful for resuming a `Promotion` from an external process: a promotion step
+(such as `http`) can trigger an external system and pass along its own name;
+the external system can then POST back to this receiver once complete, waking
+the `Promotion` for reconciliation without polling.
```
krancour: Polling would probably still have to be involved because I'm not sure how to make a Promo just stop and wait at a given step otherwise. A step for "stop until manually resumed" could be an interesting (but difficult) thing to pursue someday.

Apart from that, polling would also still be advisable anyway because of webhooks generally being considered unreliable.

So my only suggestion here is to say more with less. Just drop the "without polling part" and all of this pretty much works itself out.

LGTM in advance.

krancour (approved):

docs/docs/50-user-guide/60-reference-docs/80-webhook-receivers/bitbucket/index.md
krancour: How about we keep the functionality so it's already ready when https://github.com/akuity/kargo/issues/6020 is eventually dealt with.

We could just drop the docs claim of supporting it until #6020 is dealt with.

PR #5854, opened 2026-03-04
chore(ui): pin all npm dependency versions

rpelczar:
This change removes the ^ and ~ prefixes from dependencies in ui/package.json, locking each one to the exact version that’s currently installed.

Because Dependabot in the ui/ directory uses the versioning-strategy: increase setting, any future updates will still come through as dedicated pull requests with explicit version bumps. That makes flexible ranges unnecessary. By pinning versions, package.json now reflects the precise dependency set in use, so any change is deliberate and clearly visible in version control.

After making the updates, a clean pnpm install was run to confirm everything aligns with the lockfile. During that process, a few transitive dependencies were updated at the patch level as part of normal re-resolution (including esbuild, es-abstract, array-includes, jsonfile, and some @babel packages).


Marvin9 (approved):
Just confirming. There are some minor and patch version updates, I guess they were already in place but if not did you verify UI? Doesn't matter that much but just skimming through changelog and UI

rpelczar:
@Marvin9  I haven't bumped any of the packages. I just matched it to pnpm-lock file

PR #5856, opened 2026-03-04
feat(ui): git commit deep links for self-hosted providers

### LINKED ISSUE #5480: Deep links to self-hosted github/gitlab should work automatically through inference

rpelczar:
Replace manual regex-based URL normalization in `getGitCommitURL` with `git-url-parse` (already a project dependency, previously unused).

The old implementation used exact domain matching (`url.includes('github.com')`), which broke deep links for self-hosted instances. The new implementation uses the `resource` field from `git-url-parse`, which contains the full hostname (`gitlab.internal.net`, `github.internal.net`, etc.), so provider detection works regardless of the domain.

SSH URLs (`git@...`) are handled automatically by the library — no more manual regex substitutions.

Added unit tests covering:
- GitHub / GitLab / Bitbucket hosted (HTTPS + SSH)
- GitHub Enterprise and GitLab Self-Managed custom domains (HTTPS + SSH)
- Unknown provider fallback

Fixes #5480

Related to https://github.com/akuity/kargo/pull/5726

### DIFF
--- a/ui/src/features/freight-timeline/open-container-initiative-utils.test.ts
+++ b/ui/src/features/freight-timeline/open-container-initiative-utils.test.ts
@@ -1,0 +1,83 @@
+import { describe, expect, it } from 'vitest';
+
+import { getGitCommitURL } from './open-container-initiative-utils';
+
+describe('getGitCommitURL', () => {
+  const revision = 'abc1234';
+
+  const cases: { name: string; url: string; expected: string }[] = [
+    // GitHub — hosted
+    {
+      name: 'github HTTPS',
+      url: 'https://github.com/akuity/kargo.git',
+      expected: 'https://github.com/akuity/kargo/commit/abc1234'
+    },
+    {
+      name: 'github SSH',
+      url: 'git@github.com:akuity/kargo.git',
+      expected: 'https://github.com/akuity/kargo/commit/abc1234'
+    },
+    // GitHub Enterprise — self-hosted
+    {
+      name: 'github enterprise custom domain',
+      url: 'https://github.internal.net/akuity/kargo.git',
+      expected: 'https://github.internal.net/akuity/kargo/commit/abc1234'
+    },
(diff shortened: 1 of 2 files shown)


ui/src/features/freight-timeline/open-container-initiative-utils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    {
+      name: 'bitbucket HTTPS',
+      url: 'https://bitbucket.org/akuity/kargo.git',
+      expected: 'https://bitbucket.org/akuity/kargo/commits/abc1234'
+    },
+    {
+      name: 'bitbucket SSH',
+      url: 'git@bitbucket.org:akuity/kargo.git',
+      expected: 'https://bitbucket.org/akuity/kargo/commits/abc1234'
+    },
```
jessesuen, line 58: Not sure if it was intentional, but did not see the "self-hosted" test for bitbucket (Bitbucket Data Center is still a thing).

rpelczar: Good catch, added it.

Marvin9 (approved):

PR #5857, opened 2026-03-05
fix: git-commit with custom author fails with "Author identity unknown"

EronWright:
## Summary

- Fixes a regression from #5814 where `setCmdHome()` calls were removed from `setupAuthor()`, breaking the `git-commit` step's `author` config block
- When `Commit()` creates a temporary home directory for per-commit author overrides, `setupAuthor()` writes `user.name` and `user.email` to the wrong `.gitconfig` (the base repo's home instead of the temporary home), causing `git commit` to fail with "Author identity unknown"

### To reproduce

```yaml
- uses: git-commit
  config:
    path: ./out
    message: "test"
    author:
      name: External Contributor
      email: external@example.com
```

### Root cause

`setupAuthor()` runs `git config --global` commands that inherit `HOME` from `buildGitCommand()`, which always points to `b.homeDir`. PR #5814 removed `setCmdHome()` calls because `setupClient()` was refactored to always pass `b.homeDir` — making them appear redundant. But `setupAuthor()` is also called from `Commit()` with a temporary home directory, where the `HOME` override is essential.

### Fix

Restore `setCmdHome(cmd, homeDir)` calls in `setupAuthor()`. This is a no-op when `homeDir == b.homeDir` (the clone path) and fixes the mismatch when they differ (the `Commit()` path).

## Test plan

- [ ] Existing unit tests pass (`go test ./pkg/controller/git/...`)
- [ ] Manual test: create a Promotion using `git-commit` with a custom `author` block — commit succeeds

### DIFF
--- a/pkg/controller/git/base_repo_test.go
+++ b/pkg/controller/git/base_repo_test.go
@@ -1,0 +1,151 @@
+package git
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSetupAuthor(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name string
+		// createRepoDir controls whether b.dir exists before calling
+		// setupAuthor. When false, simulates a clone-in-progress where
+		// the repo directory hasn't been created yet.
+		createRepoDir bool
+		author        *User
+		assert        func(*testing.T, string, string, error)
+	}{
+		{
+			name:          "nil author uses defaults",
+			createRepoDir: true,
+			author:        nil,
+			assert: func(t *testing.T, homeDir, _ string, err error) {
+				require.NoError(t, err)
+				assertGitConfig(t, homeDir, "user.name", defaultUsername)
+				assertGitConfig(t, homeDir, "user.email", defaultEmail)
+			},
+		},
+		{
+			name:          "custom name and email",
+			createRepoDir: true,
+			author: &User{
+				Name:  "Test User",
+				Email: "test@example.com",
+			},
+			assert: func(t *testing.T, homeDir, _ string, err error) {
+				require.NoError(t, err)
+				assertGitConfig(t, homeDir, "user.name", "Test User")
+				assertGitConfig(t, homeDir, "user.email", "test@example.com")
+			},
+		},
+		{
+			// Regression: setupAuthor must use homeDir (not b.homeDir) so
+			// that Commit() can set up per-commit author config in a
+			// temporary home directory.
+			name:          "config written to homeDir not repo homeDir",
+			createRepoDir: true,
+			author: &User{
+				Name:  "Per-Commit Author",
+				Email: "per-commit@example.com",
+			},
+			assert: func(t *testing.T, homeDir, repoHomeDir string, err error) {
+				require.NoError(t, err)
+				assertGitConfig(t, homeDir, "user.name", "Per-Commit Author")
+				assertNoGitConfig(t, repoHomeDir)
+			},
+		},
+		{
+			// During clone, b.dir doesn't exist yet. cmd.Dir is overridden
+			// to homeDir so the command can still execute.
+			name:          "succeeds when repo dir does not exist",
+			createRepoDir: false,
+			author:        nil,
+			assert: func(t *testing.T, homeDir, _ string, err error) {
+				require.NoError(t, err)
+				assertGitConfig(t, homeDir, "user.name", defaultUsername)
+			},
+		},
+		{
+			name:          "with signing key path",
+			createRepoDir: true,
+			author: &User{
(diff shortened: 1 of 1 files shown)


krancour:
> When Commit() creates a temporary home directory for per-commit author overrides

🤦‍♂️ 

Nice catch.

pkg/controller/git/base_repo.go
```diff
@@ -133,10 +133,13 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
...
-	// Override the cmd.Dir that's set by b.buildGitCommand(). It's normally the
-	// repository's path, but if this method was called as part of the cloning
-	// process, that path may not exist yet.
+	// Override the cmd.Dir and HOME that are set by b.buildGitCommand().
+	// cmd.Dir is normally the repository's path, but if this method was called
+	// as part of the cloning process, that path may not exist yet. HOME must
+	// match homeDir so that git config --global writes to the correct
+	// .gitconfig (especially when Commit() creates a temporary home).
 	cmd.Dir = homeDir
+	b.setCmdHome(cmd, homeDir)
```
krancour: Agree with the code change -- again, good job finding this. I think I'd break up the comment to justify the two overrides individually since they're not both for the same reason.

Something like:

```suggestion
	// Override the cmd.Dir that's set by b.buildGitCommand(). It'll initially be
	// equal to the repo's own path, but if this method was invoked as part of a
	// clone operation, that path won't exist yet. Whether that is or is not the
	// case, it's safe to execute this command from homeDir instead since its aim
	// is only to establish "global" configuration for a synthetic user.
	cmd.Dir = homeDir
	
	// Override cmd's HOME env var as set by b.buildGitCommand(). It'll initially
	// be equal to the repo's dedicated home directory, which is nothing more
	// than a container of sorts for a synthetic user's "global" configuration,
	// but if this method may have been invoked with the caller having provided
	// some other value for the homeDir argument, it may be the caller's intent
	// to setup some ephemeral "global" config for a different synthetic user.
	// i.e. The caller may be doing this in preparation to make a commit
	// attributed to a specific author.
	b.setCmdHome(cmd, homeDir)
```

Please feel free to wordsmith further. The main thing I care about was breaking up the justification.

pkg/controller/git/base_repo.go
```diff
@@ -146,10 +149,8 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
 	}
 
 	cmd = b.buildGitCommand("config", "--global", "user.email", author.Email)
-	// Override the cmd.Dir that's set by b.buildGitCommand(). It's normally the
-	// repository's path, but if this method was called as part of the cloning
-	// process, that path may not exist yet.
 	cmd.Dir = homeDir
+	b.setCmdHome(cmd, homeDir)
```
krancour, line 155: I don't want to lose track of the justifications for these, but I also don't want to repeat the verbose comment from above.

Maybe we could just do something like this?

```suggestion
	// See justification for both of these overrides above.
	cmd.Dir = homeDir
	b.setCmdHome(cmd, homeDir)
```

pkg/controller/git/base_repo.go
```diff
@@ -180,19 +181,15 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
 
 		if author.SigningKeyPath != "" {
 			cmd = b.buildGitCommand("config", "--global", "commit.gpgsign", "true")
-			// Override the cmd.Dir that's set by b.buildGitCommand(). It's normally the
-			// repository's path, but if this method was called as part of the cloning
-			// process, that path may not exist yet.
 			cmd.Dir = homeDir
+			b.setCmdHome(cmd, homeDir)
```
krancour, line 188: Same as a previous comment:


```suggestion
			// See justification for both of these overrides above.
			cmd.Dir = homeDir
			b.setCmdHome(cmd, homeDir)
```

pkg/controller/git/base_repo.go
```diff
@@ -180,19 +181,15 @@ func (b *baseRepo) setupAuthor(homeDir string, author *User) error {
...
 			if _, err := libExec.Exec(cmd); err != nil {
 				return fmt.Errorf("error configuring commit gpg signing: %w", err)
 			}
 
 			cmd = b.buildCommand("gpg", "--import", author.SigningKeyPath)
-			// Override the cmd.Dir that's set by b.buildCommand(). It's normally the
-			// repository's path, but if this method was called as part of the cloning
-			// process, that path may not exist yet.
 			cmd.Dir = homeDir
+			b.setCmdHome(cmd, homeDir)
```
krancour, line 196: Same as a previous comment:

```suggestion
			// See justification for both of these overrides above.
			cmd.Dir = homeDir
			b.setCmdHome(cmd, homeDir)
```

PR #5866, opened 2026-03-05
feat: filter webhook-triggered warehouse refreshes by changed paths

jmmclean:
## Summary

When a Git push webhook is received, Kargo currently refreshes **all** warehouses
subscribed to the matching repo URL and branch — even if the warehouse's
`includePaths`/`excludePaths` filters would exclude the changed files. This causes
unnecessary git fetches, particularly impactful for monorepo setups where many
warehouses subscribe to the same repository but watch different paths.

This change extracts the per-commit changed file lists (`added`, `modified`, `removed`)
from Git push webhook payloads and evaluates them against each warehouse's path filters
**before** triggering a refresh. Warehouses whose path filters don't match any changed
files are skipped entirely.

- **GitHub, GitLab, Gitea**: Changed files are extracted from the push event payload
  and used for path filter evaluation
- **Bitbucket, Azure DevOps, and image/chart providers**: These payloads don't include
  per-commit file lists, so existing behavior is preserved (refresh on ref match alone)
- **Backwards-compatible**: Warehouses with no `includePaths`/`excludePaths` configured,
  or webhooks without changed file data, behave exactly as before

### Flow

```mermaid
flowchart TD
    A[Webhook Received] --> B[Parse push event payload]
    B --> C[Extract repo URLs, ref, and changed files]
    C --> D[Find warehouses subscribed to repo URL]
    D --> E{Ref matches subscription branch?}
    E -- No --> F[Skip warehouse]
    E -- Yes --> G{Subscription has includePaths or excludePaths?}
    G -- No --> I[Refresh warehouse]
    G -- Yes --> H{Changed files provided by webhook?}
    H -- No --> I
    H -- Yes --> J{Any changed file matches path filters?}
    J -- Yes --> I
    J -- No --> F
```

## Test plan

- [ ] New unit tests in `TestShouldRefresh` cover: matching/non-matching include paths,
  exclude paths, glob patterns, no path filters, and empty changed files (backwards compat)
- [ ] New `TestMatchesChangedFiles` covers: include-only, exclude-only, combined
  include+exclude, glob patterns, regex patterns, and edge cases
- [ ] All existing tests pass unchanged
- [ ] `make lint-go` passes with zero issues

### DIFF
--- a/docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
+++ b/docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
@@ -707,6 +707,21 @@ Paths may _also_ be specified using glob patterns (by prefixing the string with
 
 :::
 
+:::info
+
+When using
+[webhook receivers](../60-reference-docs/80-webhook-receivers/index.md) with
+GitHub, GitLab, or Gitea, `includePaths` and `excludePaths` are also evaluated
+at webhook time. Push event payloads from these providers include per-commit
+file change lists, which Kargo uses to determine whether a `Warehouse` needs to
+be refreshed. If none of the changed files match the subscription's path
+filters, the `Warehouse` is not refreshed, avoiding unnecessary artifact
+discovery. This is particularly beneficial for monorepo setups where many
+`Warehouse` resources subscribe to the same repository but watch different
+paths.
+
+:::
+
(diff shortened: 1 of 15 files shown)


fuskovic:
Duplicate of https://github.com/akuity/kargo/pull/5611

PR #5870, opened 2026-03-05
fix: return error on deleteResource errors

hairyhum:
Returning error here will allow UI to render error message instead of ignoring it silently.
Also making the behaviour more consistent with fixes in https://github.com/akuity/kargo/pull/5641

### DIFF
--- a/pkg/server/delete_resource_v1alpha1.go
+++ b/pkg/server/delete_resource_v1alpha1.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -27,7 +28,11 @@ func (s *server) DeleteResource(
 	res := make([]*svcv1alpha1.DeleteResourceResult, 0, len(resources))
 	for _, r := range resources {
 		resource := r // Avoid implicit memory aliasing
-		res = append(res, s.deleteResourceProto(ctx, &resource))
+		deleteResult := s.deleteResourceProto(ctx, &resource)
+		if deleteResult.GetError() != "" && len(resources) == 1 {
+			return nil, errors.New(deleteResult.GetError())
+		}
+		res = append(res, deleteResult)
 	}
 	return &connect.Response[svcv1alpha1.DeleteResourceResponse]{
 		Msg: &svcv1alpha1.DeleteResourceResponse{


krancour:
There is definitely an inconsistency here with how all the other bulk resource create/update/delete endpoints work, but I think the thing that would stop masking errors on deletes of a single resource (which is definitely a bug, and I think, the main case the UI cares about), and improve consistency with the other endpoints at the same time would be something like this:

```
func (s *server) DeleteResource(
	ctx context.Context,
	req *connect.Request[svcv1alpha1.DeleteResourceRequest],
) (*connect.Response[svcv1alpha1.DeleteResourceResponse], error) {
	projects, otherResources, err := splitYAML(req.Msg.GetManifest())
	if err != nil {
		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("parse manifest: %w", err))
	}
	resources := append(otherResources, projects...)
	results := make([]*svcv1alpha1.DeleteResourceResult, 0, len(resources))
	for _, r := range resources {
		resource := r // Avoid implicit memory aliasing
		res := s.deleteResourceProto(ctx, &resource)
		// This conditional is what the other endpoints have that this
		// one was mistakenly missing.
		if res.GetError() != "" && len(resources) == 1 {
			return nil, errors.New(res.GetError())
		}
		results = append(results, res)
	}
	return &connect.Response[svcv1alpha1.DeleteResourceResponse]{
		Msg: &svcv1alpha1.DeleteResourceResponse{
			Results: results,
		},
	}, nil
}
```

hairyhum:
@krancour I'm not sure why would we want to preserve `deleteResourceProto` function as it is (inconsistent with updateResources). I can't find any other place this function would be used. Is there something I miss?

Functionally the difference between the PR proposal and your comment proposal is that we fail early. I can change the PR to fail early if you want.

krancour:
I'm going from memory here. `deleteResourceProto()` remains temporarily (until we kill of the whole grpc API), because using its replacement, which is used by the REST API, would have required some extra looping afterwards to create the convert the results to the endpoint's expected return type.

So I agree what I proposed isn't 100% consistent with the implementation of create/update resource endpoints in either API and not even consistent with the delete resource endpoint in the rest API, although it's close. What I was aiming for was to make it _behaviorally_ consistent with all of those:

1. If dealing with only one resource and there's an error, just return the error.
2. If dealing with multiple resources handle mixed successes/failure by returning a list of results that individually indicate success or failure.

Isn't it only no.1 that was mistakenly not happening before and is the only thing that really needs fixing? Or have I misunderstood?

krancour (approved):

krancour:
@hairyhum did you want this backported to 1.9?

pkg/server/delete_resource_v1alpha1.go
```diff
@@ -27,6 +28,10 @@ func (s *server) DeleteResource(
 	res := make([]*svcv1alpha1.DeleteResourceResult, 0, len(resources))
 	for _, r := range resources {
 		resource := r // Avoid implicit memory aliasing
+		deleteResult := s.deleteResourceProto(ctx, &resource)
+		if deleteResult.GetError() != "" && len(resources) == 1 {
```
EronWright, line 32: Is it intentional that `deleteResourceProto` is called twice when `len(resources)` != one?

EronWright (changes requested):

PR #5874, opened 2026-03-06
feat: add "my projects" filter

### LINKED ISSUE #5707: Multi-tenancy/UX Projects dashboard improvement
# Checklist

# Proposed Feature
From the perspective of hosting Kargo for multiple tenants, it would be useful to have the ability to restrict access to projects that a user is not assigned to.

# Motivation
Currently, whenever any user opens the Kargo UI, the main projects dashboard is populated with a list of all projects Kargo hosts, even if namespace level RBAC restricts user access to those projects. As they are unusable for the user, this creates unnecessary noise. It can be an issue if administrators want to restrict visibility of projects that are of no concern to the users; furthermore, from a user experience point of view, it is overwhelming to handle loads of data presented to them without any reason.

# My Story
After setting up promotion access for specific OIDC roles, I noticed that my test user has no projects on the main dashboard. Then, after entering the rabbit hole of #4082 and #3977, I ended up with a few setups, each of which has different issues.

The first is a global kargo-user service account. That is the solution I decided to use for now for default cluster-wide permissions for all users. It allows the user to get, watch, and list all projects, namespaces, and other cluster-level Kargo resources. This solution renders the main projects dashboard properly, but shows all projects in the cluster, which, as described earlier, is not necessarily desired. The next issue is that when project-level access is managed based on more granular OIDC roles, whenever a user clicks on a project that appears on the dashboard (implying the user can interact with it), the UI shows an empty dashboard and returns permission errors.
https://github.com/user-attachments/assets/f9bec55b-93a5-4d7a-8f43-d248704f7e38

The second is adding separate service accounts for Kargo to handle cluster-level permissions. That possibility is supported even by the Kargo Helm chart via `globalServiceAccounts.namespaces`, where it is possible to include service accounts with their cluster roles and cluster role bindings so Kargo can interpret permissions of users based on their OIDC claims. This gave me hope, but the issue here has two separate sources. One is that Kubernetes RBAC allows the list verb only on all resources of a kind; restricting access to projects and namespaces in a `ClusterRole` based on the `resourceNames` field is only possible for the get and watch verbs. This leaves the same level of permissions to list all projects and namespaces. The other thing is that the Kargo projects dashboard uses listing projects to render the view; whenever this permission is missing, the dashboard fails to render and throws errors. Also, without get permission, it fails to populate status fields which are also visible.

Here project had to be entered by putting its path into URL because otherwise it's not visible in projects dashboards:
https://github.com/user-attachments/assets/7e5e8966-c944-4efa-912a-88c98a907f72

# Proposed Solution
To maintain a 1:1 permission mapping between Kargo and Kubernetes RBAC, the issue should be addressed by the Kargo backend, which would have to return list of projects that the user has `get` permission to, whenever the UI requests a list of projects. Instead of directly listing all projects in cluster.

I am open for better ideas to tackle this issue, even if my proposed solution won't be accepted, the issue itself remains.

krancour:
Fixes #5707

**_First and foremost, this is not a security feature._**

Kargo relies on pure Kubernetes RBAC. Kubernetes RBAC has no way to say anything as complex as, "You can list this type of resource, but only certain ones." List is an all or nothing proposition.

As ever, if you can list Projects, you can list them all.

This is a _usability_ feature. What it delivers is a "My Projects" filter which can be enabled or disabled at will. It is enabled by default.

What makes a Project qualify as "mine?" It's "mine" if I've been explicitly mapped (via OIDC claims) to one or more ServiceAccounts in the Project's underlying namespace. i.e. If I've _explicitly_ been given permissions within this _specific_ Project.

Here's what _does not_ qualify as "my projects": Projects that one can access due to _broader, system-level permissions_. If, for instance, I were mapped to the `kargo-admin` ServiceAccount in Kargo's own namespace, I would have very broad permissions across all Projects, but this alone would not qualify those projects as "mine."

Remember the filter is enabled by default. So if you're a `kargo-admin` with broad access, but weren't explicitly given access to specific Projects _individually_, you will log in and see _no projects_. Disable the filter. (One click in the upper right of the Project list view.)

https://github.com/user-attachments/assets/7dba75fa-e4e6-4fbf-9ab1-9bc81936fb9c

### DIFF
--- a/ui/src/features/project/list/projects-list.tsx
+++ b/ui/src/features/project/list/projects-list.tsx
@@ -1,7 +1,7 @@
 import { useQuery } from '@connectrpc/connect-query';
 import { faStar, faUser } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { Button, Empty, Flex, Pagination } from 'antd';
+import { Button, Empty, Flex, Pagination, Space } from 'antd';
 import { useEffect, useState } from 'react';
 
 import { LoadingState } from '@ui/features/common';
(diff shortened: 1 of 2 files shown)


krancour:
@rpelczar I won't merge this without your approval because I really don't know if the UI changes are sensible or not. Code wise, I think they're probably ok, but I'm less certain about visual style.

krancour:
@rpelczar I saw you improved this quite a bit. Thank you. Please feel free to merge yourself whenever you have it in a state you're happy with.

PR #5886, opened 2026-03-09
fix(chart): enforce the group on the certificate

danielloader:
Spent a morning debugging a squirrelly issue with Cert Manager refusing to renew self signed certificate for the internal webhook server.

Found https://github.com/aws/eks-charts/issues/1242 and manually created the resource outside of ArgoCD (which owns and syncs kargo) with sync disabled:

```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: kargo-webhooks-server
  namespace: kargo
spec:
  dnsNames:
    - kargo-webhooks-server.kargo.svc
  issuerRef:
    group: cert-manager.io
    kind: Issuer
    name: kargo-selfsigned-cert-issuer
  secretName: kargo-webhooks-server-cert
```

This then allowed the renewal of the expired certificate - I don't know if it's necessarily the right solution to be explicit with the group, and cert-manager seems to be permissive of installing and creating a certificate from nothing without one, but when it comes to renew time a year later it was not in a happy state.

Given it's the default value - just been explicit and added it to the template.

Hard to test unless you have an installation that's existed for nearly a year and due a certificate renew soon - but since it's not a change in behaviour except for adding the explicit field that cert-manager does default to if omitted, should be safe.

### DIFF
--- a/charts/kargo/templates/api/cert.yaml
+++ b/charts/kargo/templates/api/cert.yaml
@@ -11,6 +11,7 @@ spec:
   dnsNames:
   - {{ quote .Values.api.host }}
   issuerRef:
+    group: cert-manager.io
     kind: Issuer
     name: kargo-selfsigned-cert-issuer
   secretName: {{ .Values.api.tls.secretName }}
--- a/charts/kargo/templates/api/ingress-cert.yaml
+++ b/charts/kargo/templates/api/ingress-cert.yaml
@@ -11,6 +11,7 @@ spec:
   dnsNames:
   - {{ quote .Values.api.host }}
   issuerRef:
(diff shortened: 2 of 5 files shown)


krancour:
Thanks @danielloader. After some brief research of my own, I agree with your assessment of the situation.

The only thing I'd suggest is that this same fix is probably needed for the other five self-signed certs that the chart might produce. (To be fair, unlike the one you've already fixed, I wouldn't expect any of those to be used for a year or more, but stranger things have happened.)

Happy to approve this once that's set.

danielloader:
Done and thanks!

PR #5893, opened 2026-03-10
feat: add github-verified-push promotion step

### LINKED ISSUE #5732: Support commit signing for GitHub Apps

EronWright:
## Summary

Adds a new `github-verified-push` promotion step that pushes commits to GitHub via the REST API, producing "Verified" commits when authenticating with a GitHub App. This is a drop-in replacement for `git-push` when branch protection rules require verified signatures.

Closes #5732

Supersedes #5840 with a simplified trust model.

### Trust model

- **App-authored commits receive the Verified badge** — when a commit's author and committer match the configured app identity (and optionally pass GPG fingerprint verification), both Author and Committer are omitted from the API call, allowing the GitHub App to sign the commit
- **Non-app commits preserve original authorship** — the original author is preserved on the replayed commit, and the configured app identity is set as the committer (mirroring `git rebase` semantics). These commits do not receive the Verified badge
- **Optional `author` config** — configure the app identity (`name`, `email`, optional `signingKey`) on the push step. When `signingKey` is provided, commit GPG signatures are verified by fingerprint match. Bad or revoked signatures cause a terminal error

### Features

- Replays commits via the GitHub REST API with automatic "Verified" badge for app-authored commits
- `pullPolicy` field: `Merge` (default), `Rebase`, or `FFOnly` for reconciling with remote branch
- `generateTargetBranch` for PR workflows
- `force` push support for diverged branches
- Retry logic for concurrent promotions targeting the same branch — on 422 (ref update conflict), reconciles and retries
- Merge conflicts are detected as terminal errors (no retries)
- `maxAttempts` config to limit retry attempts (default: 10 for Merge/Rebase, 1 for FFOnly)
- GPG signature verification: bad ("B") and revoked ("R") signatures cause terminal errors regardless of whether a signing key is configured
- Comprehensive unit tests

### Bug fix: `IsRebasing()` in linked worktrees

Fixed a pre-existing bug in `IsRebasing()` where `git rev-parse --git-path rebase-merge` returns an absolute path in linked worktrees (as used by Kargo in containers), but the code unconditionally joined it with `w.dir`, producing an invalid path. This caused merge conflicts to never be detected as terminal errors in both `git-push` (on `main`) and `github-verified-push`.

## Test plan

- [x] Unit tests pass (`go test ./pkg/promotion/runner/builtin/ -run githubVerifiedPusher`)
- [x] Lint passes (`make lint-go`)
- [x] Manual test: unsigned app-authored commit → Verified, author = GitHub App ([454b3cc](https://github.com/EronWright/kargo-demo/commit/454b3cc70b9a2481abe3b6efb71fd1ce13d96446))
- [x] Manual test: unsigned external author → not Verified, author = "External Contributor", committer = "Kargo" ([606a66f](https://github.com/EronWright/kargo-demo/commit/606a66fcce77645a785e3d6695bfa3381de00ad5))
- [x] Manual test: GPG-signed app-authored commit (fingerprint match) → Verified ([9d33cb5](https://github.com/EronWright/kargo-demo/commit/9d33cb5e94864caf41d19f035594439ae2dd2cfb))
- [x] Manual test: GPG-signed external author (fingerprint mismatch) → not Verified, author preserved ([96bfe19](https://github.com/EronWright/kargo-demo/commit/96bfe195144ec942afcbafa95f5dcba80703671d))
- [x] Manual test: force push with real divergence — divergent content overwritten
- [x] Manual test: merge conflict during rebase — terminal error, no retries

### DIFF
--- a/pkg/promotion/runner/builtin/github_verified_pusher.go
+++ b/pkg/promotion/runner/builtin/github_verified_pusher.go
@@ -902,18 +902,22 @@ func (g *githubVerifiedPusher) replayCommits(
 			Message: &message,
 			Tree:    &github.Tree{SHA: rc.Commit.Tree.SHA},
 			Parents: parents,
+			Author:  rc.Commit.Author,
+			Committer: &github.CommitAuthor{
+				Name:  &appName,
+				Email: &appEmail,
+			},
 		}
 
-		// Determine whether this is an app-authored commit.
+		// When the commit is app-authored, clear Author and Committer
+		// so the authenticated identity (GitHub App) is used for both
+		// and the commit receives the "Verified" badge.
 		appAuthored := g.isAppAuthored(
 			rc, appName, appEmail, appFingerprint, commitSigs,
 		)
-		if !appAuthored {
-			// Preserve original author. The commit will not receive
-			// GitHub's "Verified" badge. Committer is always left nil
-			// so the authenticated identity (GitHub App) is used — it
-			// is the actual committer in all cases.
-			commit.Author = rc.Commit.Author
+		if appAuthored {
+			commit.Author = nil
+			commit.Committer = nil
 		}
 
 		newCommit, _, createErr := client.CreateCommit(
--- a/pkg/promotion/runner/builtin/github_verified_pusher_test.go
+++ b/pkg/promotion/runner/builtin/github_verified_pusher_test.go
@@ -653,7 +653,7 @@ func Test_githubVerifiedPusher_replayCommits(t *testing.T) {
 			},
 		},
 		{
-			name:   "app-authored commit omits author",
+			name:   "app-authored commit clears author and committer",
 			pusher: &githubVerifiedPusher{},
 			client: &mockGitHubVerifiedPushClient{
 				createCommitFn: func(
@@ -694,7 +694,7 @@ func Test_githubVerifiedPusher_replayCommits(t *testing.T) {
 			},
 		},
 		{
-			name:   "non-app commit preserves author",
+			name:   "non-app commit preserves author and sets app as committer",
 			pusher: &githubVerifiedPusher{},
 			client: &mockGitHubVerifiedPushClient{
 				createCommitFn: func(
@@ -703,14 +703,14 @@ func Test_githubVerifiedPusher_replayCommits(t *testing.T) {
 					commit github.Commit,
 					_ *github.CreateCommitOptions,
 				) (*github.Commit, *github.Response, error) {
-					// Non-app: Author/Committer preserved.
+					// Non-app: Author preserved, Committer set to app.
 					require.NotNil(t, commit.Author)
 					require.Equal(
 						t, "Custom", commit.Author.GetName(),
 					)
 					require.NotNil(t, commit.Committer)
 					require.Equal(
-						t, "Custom", commit.Committer.GetName(),
+						t, "Kargo", commit.Committer.GetName(),
 					)
 					return &github.Commit{
 						SHA: ptr.To("signed1"),


EronWright:
## Manual Test Results

Tested against a local Tilt cluster with kargo-demo (EronWright/kargo-demo), controller configured with a GPG signing key.

### Test 1: Default author (system identity)

Promotion: `git-commit` (no custom author) → `github-verified-push`

| Field | Value |
|-------|-------|
| Author | `kargo-demo-signer-eronwright[bot]` |
| Committer | `GitHub <noreply@github.com>` |
| Verified | ✅ `true` (reason: `valid`) |
| Co-authored-by | none (correct — system author) |

Commit: https://github.com/EronWright/kargo-demo/commit/e7984b7fced5b63fb6a4bb61d7b0feac5dc270b5

### Test 2: Custom author

Promotion: `git-commit` with `author: {name: "Custom Author", email: "author@example.com"}` → `github-verified-push`

| Field | Value |
|-------|-------|
| Author | `kargo-demo-signer-eronwright[bot]` |
| Committer | `GitHub <noreply@github.com>` |
| Verified | ✅ `true` (reason: `valid`) |
| Co-authored-by | `Custom Author <author@example.com>` |

Both cases confirm:
- All commits are App-signed and receive the "Verified" badge
- Co-authored-by trailer is added only when the original author differs from Kargo's system identity

pkg/promotion/runner/builtin/git_cloner.go
```diff
@@ -63,12 +63,21 @@ func gitUserFromEnv() git.User {
 		SigningKeyPath string `envconfig:"GITCLIENT_SIGNING_KEY_PATH"`
 	}{}
 	envconfig.MustProcess("", &cfg)
-	return git.User{
-		Name:           cfg.Name,
-		Email:          cfg.Email,
+	name := cfg.Name
+	if name == "" {
+		name = git.DefaultUsername
```
EronWright: Rationale: the system user account's name is defaulted to "Kargo" elsewhere, so do the defaulting eagerly for the benefit of all.

EronWright:
## Manual Test Results: Retry/Rebase Feature

Deployed to a local Tilt dev cluster and tested the key code paths using `EronWright/kargo-demo` with a GitHub App installation token.

### Test 1: Happy Path (Normal Push) ✅

Standard promotion with `github-verified-push` to `stage/test`. Commits replayed through the API, marked as "Verified." Output `commit` SHA matches GitHub HEAD (ForcePull verification). No retry log messages.

### Test 2: Force Push with Real Divergence ✅

Used an `http` gate step to block the promotion mid-flow, then pushed a divergent commit to `stage/test` while blocked. After releasing the gate, `github-verified-push` with `force: true` correctly detected the "diverged" status, used the merge base as parent SHA, and overwrote the divergent content. No PullRebase attempted.

### Test 3: Merge Conflict (Terminal Error) ✅

Used the same `http` gate technique to block mid-promotion, then pushed a conflicting change to `stage/test`. After releasing the gate, `github-verified-push` attempted `PullRebase`, detected the merge conflict, and returned a terminal error immediately with no retries:

```
an unrecoverable error occurred: error running step "step-7": merge conflict
```

### Bug Found & Fixed: `IsRebasing()` in Linked Worktrees

Test 3 initially failed — the merge conflict was classified as `Errored` (retryable) instead of terminal. Investigation revealed a **pre-existing bug on `main`** in `IsRebasing()` (`pkg/controller/git/work_tree.go`):

- `git rev-parse --git-path rebase-merge` returns an **absolute path** in linked worktrees (which is how Kargo uses them in containers)
- The code unconditionally did `filepath.Join(w.dir, path)`, producing an invalid path when the result was already absolute
- `IsRebasing()` always returned `false` in linked worktrees, so merge conflicts during `PullRebase` (and `Push` with `pullRebase: true` on `main`) were never detected as terminal

Fix: check `filepath.IsAbs()` before joining. Commit: a5bea280

pkg/promotion/runner/builtin/git_cloner.go
krancour: I agree with this improvement, but want to suggest there's a better place to apply it now, which was probably not the case at the time this `gitFromEvnv()` function was originally written.

We should probably kill this function off and modify `init()` above like this:

```
func init() {
	cfg := struct {
		Name           string `envconfig:"GITCLIENT_NAME" default:"Kargo"`
		Email          string `envconfig:"GITCLIENT_EMAIL" default:"no-reply@kargo.io"`
		SigningKeyType string `envconfig:"GITCLIENT_SIGNING_KEY_TYPE"`
		SigningKeyPath string `envconfig:"GITCLIENT_SIGNING_KEY_PATH"`
	}{}
	defaultGitUser := git.User{
		Name:           cfg.Name,
		Email:          cfg.Email,
		SigningKeyType: git.SigningKeyType(cfg.SigningKeyType),
		SigningKeyPath: cfg.SigningKeyPath,
	}
	promotion.DefaultStepRunnerRegistry.MustRegister(
		promotion.StepRunnerRegistration{
			Name: stepKindGitClone,
			Metadata: promotion.StepRunnerMetadata{
				RequiredCapabilities: []promotion.StepRunnerCapability{
					promotion.StepCapabilityAccessCredentials,
				},
			},
			Value: func(caps promotion.StepRunnerCapabilities) promotion.StepRunner {
				return newGitCloner(caps, defaultGitUser)
			},
		},
	)
}
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,195 @@
...
+[`git-push`](git-push.md) that replays commits through the API instead of using
+`git push`.
+
+This step exists because GitHub Apps have no associated GPG to sign their commits
+with in the standard way. If your Kargo installation authenticates to GitHub via a
+[GitHub App installation token](../../50-security/30-managing-secrets.md#github-app-authentication)
+and your branch protection rules require verified signatures, this step provides
+an alternative solution. Commits created through the API with an
+App token are automatically signed by GitHub and marked as "Verified" — without
+requiring GPG key management on the GitHub side.
```
krancour: I get what you're saying, but I'm not sure the dots are adequately connected here for someone without much prior knowledge of the topic.

Can we draw some sharper connections between some of these statements with something like this?

> Many users prefer that Kargo authenticates to GitHub using a [GitHub App](../../50-security/30-managing-secrets.md#github-app-authentication) because it avoids coupling authentication to any one GitHub account. GitHub Apps, however, cannot be associated with a GPG signing key, so when branch protection rules require verified commits, the key used for signing must still be associated with a user account, which undermines the benefits of having used an App for authentication. As a better alternative to signing commits with a key associated with a user account (in the [`git-commit`](git-commit.md) step), `git-verified-push`, a drop-in replacement for the [`git-push`](git=push.md) step, replays local commits on a remote branch using the GitHub API, resulting in commits signed directly by GitHub's own key, with nothing Kargo depends upon having been associated with a user account.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,195 @@
...
+Two authentication methods are supported:
+
+- **GitHub App installation token** (recommended) — Commits replayed through
+  the API are automatically signed by GitHub and marked as "Verified." The
+  GitHub App must have **Contents: read & write** permission. See
+  [GitHub App Authentication](../../50-security/30-managing-secrets.md#github-app-authentication)
+  for setup instructions.
+- **Personal access token (PAT)** — Commits are replayed successfully and
+  attributed to the PAT owner, but are **not** marked as "Verified." This
+  can be useful for testing.
```
krancour: I'm iffy on the value of doing this with a PAT, even for testing. What would you think about saying that the step doesn't explicitly require you to authenticate via an App, but only has substantial value when you are?

We could maybe even get away with saying nothing about this at all here if the introductory text is clear enough on when and why you'd use this.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,195 @@
...
+`Co-authored-by` trailer is added to the commit message to credit the original
+author. GitHub renders these trailers as linked avatars in the commit UI.
+
+:::note
+
+If you use the [`git-commit`](git-commit.md) step with a custom `author`
+configuration, that author will appear as a co-author on the resulting
+GitHub commit.
+
+:::
```
krancour, line 120: This doesn't strike me as the correct thing to do.

Why wouldn't authorship continue to be explicitly attributed to the original author?

Or is it that the API doesn't _let_ us do that?

pkg/controller/git/base_repo.go
```diff
@@ -15,8 +15,12 @@ import (
...
 
 const (
-	defaultUsername = "Kargo"
-	defaultEmail    = "no-reply@kargo.io"
+	// DefaultUsername is the default git user name used for commits when
+	// the GITCLIENT_NAME environment variable is not set.
+	DefaultUsername = "Kargo"
+	// DefaultEmail is the default git email used for commits when
+	// the GITCLIENT_EMAIL environment variable is not set.
+	DefaultEmail = "no-reply@kargo.io"
```
krancour: It was intentional that these weren't exported before. These were package-level / low-level defaults and I'd intended for higher-level components, like steps, to be able to have their own defaults that might be different from these.

See related comment: https://github.com/akuity/kargo/pull/5893/changes#r2947437955

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: It doesn't let us do that, meanwhile the co-authored-by trailer surfaces nicely in the Github UI:
https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,976 @@
...
+) (*github.Reference, *github.Response, error) {
+	return w.client.Git.UpdateRef(ctx, owner, repo, ref, updateRef)
+}
+
+func (w *githubVerifiedPushClientWrapper) DeleteRef(
+	ctx context.Context,
+	owner, repo, ref string,
+) (*github.Response, error) {
+	return w.client.Git.DeleteRef(ctx, owner, repo, ref)
+}
```
krancour: I think previously mentioned a way of simplifying this wrapper with something like this:

```
type githubVerifiedPushClientWrapper struct {
	*github.Client
}
```

You correctly pointed out that you were flattening two different sub-clients into one.

It just occurred to me that you can drop a lot of code and still accommodate that concern like this:

```suggestion
// githubVerifiedPushClientWrapper wraps a *github.GitService and
// *github.RepositoriesService to implement githubVerifiedPushClient.
type githubVerifiedPushClientWrapper struct {
	*github.GitService
	*github.RepositoriesService
}
```

krancour: Obviously requires a tiny change where you instantiate the `githubVerifiedPushClientWrapper`.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,976 @@
...
+
+// githubVerifiedPusherConfig holds controller-level configuration for the
+// github-verified-push step, populated from environment variables.
+type githubVerifiedPusherConfig struct {
+	MaxRevisions int `envconfig:"GITHUB_VERIFIED_PUSH_MAX_REVISIONS" default:"10"`
```
krancour: These struct tags are no longer doing anything in this location.

```suggestion
	MaxRevisions int
```

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,976 @@
...
+			return "", "", nil, fmt.Errorf(
+				"error configuring GitHub Enterprise URLs: %w", err,
+			)
+		}
+	}
+	client = client.WithAuthToken(token)
+
+	return owner, repo,
+		&githubVerifiedPushClientWrapper{client: client}, nil
+}
```
krancour, line 861: I may have mentioned this in the PR that this one superseded, but this same code appears elsewhere as well. I think it's time to create a small `github` utility package with this function in it.

As DRY violations go, this is two occurrences only and isn't totally heinous, but some small bit of history... there have been multiple GitHub Enterprise-related bug fixes to the _other_ spot where this appears. Fearing we're not done finding such issues, I'd rather have this logic centralized instead of duplicated.

pkg/promotion/runner/builtin/github_verified_pusher.go
```diff
@@ -0,0 +1,976 @@
...
+	// have stale refs. Non-fatal if it fails — the push already succeeded.
+	if !createBranch {
+		if pullErr := workTree.ForcePull(targetBranch); pullErr != nil {
+			logger.Error(
+				pullErr,
+				"error syncing local branch after push (non-fatal)",
+				"branch", targetBranch,
+			)
+		}
+	}
```
krancour: You may have a good reason for it, but I'm wondering why this is conditional.

pkg/promotion/runner/builtin/github_verified_pusher.go
krancour: Same applies to `parseGitHubRepoURL()` above and `buildCommitURL()` below.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,195 @@
...
+`github-verified-push`, a drop-in replacement for the
+[`git-push`](git-push.md) step, replays local commits on a remote branch using
+the GitHub API, resulting in commits signed directly by GitHub's own key, with
+nothing Kargo depends upon having been associated with a user account.
+
+Under the hood it:
+
+1. Compares the local branch to the remote target branch to identify new commits
+1. Pushes local commits to a temporary, non-branch staging ref on GitHub
+   (invisible in the branch list)
```
krancour: Are these in the wrong order here? Don't we stage first and then compare between the staging ref and the head of the remote target?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
```diff
@@ -0,0 +1,195 @@
...
+   GitHub REST API, creating new commits with new SHAs
+1. Updates the target branch to the final replayed commit (fast-forward by
+   default, or force-update when `force: true`)
+1. Deletes the staging ref
+
+Because commits are recreated through the API, the resulting remote commits have
+different SHAs than the original local commits — even when the content is
+identical. This is expected and does not affect subsequent promotions, since each
+promotion clones a fresh working tree. After a successful push, the step syncs
+the local branch to match the remote, so subsequent steps see the correct state.
```
krancour: There might be a lot of room for misinterpretation here, depending on how well one really understands the preceding explanation. I think a small bit of disambiguation here might go a long way:

```suggestion
the local branch to match the remote target branch, so subsequent steps
see the correct state.
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
krancour: > It doesn't let us do that

I dug into this. It lets us. It's just that we don't get the verified badge if we do.

My take on this isn't that we should explicitly omit formal (not as a trailer) author information just to ensure we get a green check. We can omit it if Kargo was the author, in which case authorship is automatically attributed to the authenticated user -- the App. If Kargo _wasn't_ the author, we should include that information  formally (not as a trailer) and let GitHub decide whether their own verification criteria are met.

jessesuen: A couple of points here:

* We are splitting hairs on a corner case. In the common, idiomatic usage of Kargo for GitOps promotion, Kargo creates net-new content to push, and it's appropriate for that content to be authored by the system identity. So, for 99% of use cases, Kargo does not rebase content authored & signed by other users, and people won't know or care about this behavior.

* However, for the remaining 1%:
  * Users **_should_** be aware of what they are doing and the consequences of Kargo behavior. Namely, that original commits that may have been properly signed & attributed to the user's GPG keys, will no longer be.
  * They should get some say in the matter. It's not for us to unilaterally decide what the appropriate behavior is for their situation (e.g., they may not care and just want the green check, or they do care and want to preserve proper attribution/provenance)

Therefore, the proper thing to do seems to be:
* Provide the **_option_** to replay it with Kargo's identity when authorships differ, rather than discarding the original altogether.
* Choose the secure option as the default. When they **_are_** doing something weird, we should error out by default so the users can: (1) adjust their practices to avoid the scenario or (2) accept the consequences by opting in using this option. Again, I consider this a corner case that most users will never encounter, so it's fine to fail by default.

pkg/promotion/runner/builtin/github_verified_pusher.go
EronWright: My understanding is that `createBranch` implies that `targetBranch` doesn't exist and that a pull would fail.

krancour: That's definitely true up-front, but by this point in the method, doesn't that branch now exist?

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Thanks for the thoughtful discussion. I've implemented a `pullPolicy` field that addresses Jesse's "secure by default" requirement while being immediately familiar to anyone who's configured `git pull`:

**`pullPolicy`** (enum: `Merge` | `Rebase` | `FFOnly`, default: `Merge`)

Controls how local commits are reconciled with the remote target branch when it has advanced. `force` remains orthogonal (controls whether to overwrite remote history).

| Policy | Behavior |
|--------|----------|
| **`Merge`** (default) | Creates a merge commit integrating remote changes. The replay is merge-aware: original remote commits remain untouched, and the merge commit is created via the API with its proper multi-parent structure preserved. Only the merge commit and local-only commits are replayed. |
| **`Rebase`** | Rebases local commits onto the remote branch, producing a linear history. All rebased commits are replayed through the API. |
| **`FFOnly`** | Fast-forward only. Errors immediately if the remote has advanced. No reconciliation, no retries. |

This addresses the concerns raised:

1. **Secure by default.** Merge-by-default preserves original remote commits — they keep their original authorship, signatures, and SHAs. Kargo only rewrites its own commits (the local-only ones) and the merge commit itself. This maximizes interoperability with `git-commit` + custom `author` and produces output that reliably passes branch protection rules.

2. **Users opt in to rewriting.** `pullPolicy: Rebase` is explicitly opt-in. When used, rebased commits from outside the promotion flow will be signed by Kargo with a `Co-authored-by` trailer to credit the original author.

3. **General utility.** The concept maps directly to `git pull --no-rebase` / `--rebase` / `--ff-only`, making it self-documenting.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Good catch — fixed. The list now reflects the actual execution order: stage first, then compare.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Fixed — now reads "the local branch to match the remote target branch".

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
krancour: These are meaningful improvements insofar as they make it possible to not rewrite history on the local branch as remote commits from the target branch are integrated into it, but my main concern remains unaddressed. Formal attribution to the original author isn't being included in the commits as they are replayed via GitHub API, which means GitHub is unconditionally attributing authorship to Kargo and counting commits as verified that would not have been had we not withheld that information.

pkg/promotion/runner/builtin/schemas/github-verified-push-config.json
```diff
@@ -0,0 +1,65 @@
...
+      "description": "Whether to force push to the target branch, overwriting any existing history. This is useful for scenarios where you want to completely replace the branch content (e.g., pushing rendered manifests that don't depend on previous state). Use with caution as this will overwrite any commits that exist on the remote branch but not in your local branch.",
+      "default": false
+    },
+    "pullPolicy": {
+      "type": "string",
+      "description": "Controls how local commits are reconciled with the remote target branch when it has advanced. 'Merge' (default) creates a merge commit preserving remote history. 'Rebase' rebases local commits onto the remote branch. 'FFOnly' permits only fast-forward updates and errors if the remote has advanced.",
+      "enum": ["FFOnly", "Rebase", "Merge"],
+      "default": "Merge"
+    }
+  },
```
krancour, line 64: This is probably better left to the operator as a matter of policy.

EronWright:
## Manual Test Pass: `pullPolicy` Feature

Ran a full end-to-end test pass against a local Tilt dev cluster using the `kargo-demo` project (`EronWright/kargo-demo`), with GitHub App installation credentials for verified commit signing.

**Test technique:** For divergence tests, an `http` step was inserted before `github-verified-push` pointing at a local gate server (Python HTTP server with a 60s sleep). During the gate window, external commits were pushed to `stage/test` to create real divergence between the local working tree and the remote.

### Results

| # | Scenario | pullPolicy | Result | Details |
|---|----------|------------|--------|---------|
| 1 | Happy path (no divergence) | Merge (default) | ✅ Pass | Linear history, verified commit |
| 2 | **Remote advanced** | **Merge** | ✅ Pass | **Merge commit with correct SHA mapping** — replayed merge has 2 parents: replayed promo commit (mapped via shaMap) + external commit (as-is) |
| 3 | Remote advanced | FFOnly | ✅ Pass | Fails with `comparison status is "diverged"` (terminal error, no retries) |
| 4 | No divergence | FFOnly | ✅ Pass | Fast-forward succeeds |
| 5 | **Remote advanced** | **Rebase** | ✅ Pass | **Linear history** — promo commit rebased on top of external commit, no merge commits |
| 6 | Force push | Merge + `force: true` | ✅ Pass | Succeeds (force only relevant during retry when remote advances after clone) |
| 7 | Custom author (`git-commit` with `author`) | Merge | ✅ Pass | `Co-authored-by: Test Author <test@example.com>` trailer present, commit verified |
| 8 | **Merge conflict** | Merge | ✅ Pass | Fails with `merge conflict` (terminal error) |

### Key Verification: Merge-Aware Replay (Test 2)

The critical test validated that `resolveParents()` correctly maps SHAs through the replay:

```
PullMerge → local merge commit (parents: promo-commit + external-commit)
Compare ca9c9b4...5ef432c → status "ahead", aheadBy 2, totalCommits 2
Replay 1/2: b68a6d8 → a32aa28 (promo commit, merge: false)
Replay 2/2: 5ef432c → 7c75349 (merge commit, merge: true)
  Parents: a32aa28 (replayed via shaMap) + ca9c9b4 (external, unchanged)
```

Resulting topology on GitHub:
```
*   7c75349 Merge branch 'stage/test' ...
|\  
| * ca9c9b4 external commit by another actor
* | a32aa28 promote nginx image
|/  
* aa21f54 Initial commit
```

All commits show "Verified" badge on GitHub.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Got it, the "next stop" on this train of thought is to check each commit's author (and GPG signature, if configured) against the system account (or an explicit `committer` argument to `github-verified-push`?).  We then suppress that as the `author` of the replayed commit (so that the app is treated as the author).

pkg/promotion/runner/builtin/schemas/github-verified-push-config.json
EronWright: I think you're saying that the pull strategy - merge, rebase, ff - would be set system-wide. Strikes me as overly-complex and I would advocate for reducing the scope by just doing "ff-only".

EronWright:
## Manual test results: author/committer attribution

Tested four scenarios covering the matrix of {unsigned, GPG-signed} × {app-authored, external author}. All tests used the `kargo-demo` repo with a GitHub App for authentication.

### Results

| # | Scenario | Verified? | Author on GitHub | Committer on GitHub | Commit |
|---|----------|-----------|------------------|---------------------|--------|
| 1 | Unsigned, app-authored (default identity) | ✅ Yes | kargo-demo-signer-eronwright[bot] | GitHub | [454b3cc](https://github.com/EronWright/kargo-demo/commit/454b3cc70b9a2481abe3b6efb71fd1ce13d96446) |
| 2 | Unsigned, external author | ❌ No | External Contributor | Kargo | [606a66f](https://github.com/EronWright/kargo-demo/commit/606a66fcce77645a785e3d6695bfa3381de00ad5) |
| 3 | GPG-signed, app-authored (fingerprint match) | ✅ Yes | kargo-demo-signer-eronwright[bot] | GitHub | [9d33cb5](https://github.com/EronWright/kargo-demo/commit/9d33cb5e94864caf41d19f035594439ae2dd2cfb) |
| 4 | GPG-signed, external author (fingerprint mismatch) | ❌ No | External Contributor | Kargo | [96bfe19](https://github.com/EronWright/kargo-demo/commit/96bfe195144ec942afcbafa95f5dcba80703671d) |

### How it works

- **App-authored** (scenarios 1 & 3): Both `Author` and `Committer` are set to nil on the GitHub API call → GitHub fills in the App identity and signs the commit → **Verified badge**.
- **Non-app** (scenarios 2 & 4): `Author` is set to the original commit author, `Committer` is set to the configured app identity ("Kargo"). This mirrors standard `git rebase` semantics where the rebaser becomes the committer. → **No Verified badge**, proper attribution preserved.

### GPG verification

- Scenario 3: Commit signed by the Kargo system key → signature status "U" (valid but untrusted in the ephemeral work tree keyring), fingerprint matches the app's key → `isAppAuthored=true`
- Scenario 4: Commit signed by an external contributor's key → signature status "U", fingerprint does NOT match → `isAppAuthored=false`, no terminal error (valid signature, just not the app's key)

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/github-verified-push.md
EronWright: Alright, the requested behavior is in place.  Only those commits that are signed by the system key are considered app commits and become signed and Verified.  All others become unsigned.

pkg/promotion/runner/builtin/schemas/github-verified-push-config.json
krancour: We can't realistically have this _not_ be a system-level policy. Put yourself in the shoes of the Kargo admin in an environment with rigorous supply chain requirements, like a financial institution. You wouldn't want developers deciding on their own accord to YOLO.

#5904 already made this a chart option. All that's needed is to make the step read that config and pass it on as a pull option.

krancour:
https://github.com/akuity/kargo/pull/5893#issuecomment-4114824195 is a big improvement.

My only concern is around how Kargo identifies itself as the author of a commit when there's no signature on the commit. Last I'd looked, it was determining this by matching on the name and email address, but that's trivial to forge and it fools this scheme.

The reliable thing to do would be to have Kargo believe itself to be the original author of a commit only when the commit is already signed by itself.

Before we get too deep into this, I'm going to acknowledge right up front that it starts to sounds pretty complicated -- that commit signing _must_ be configured for Kargo to trust itself. That's not at all lost on me, but I'm afraid I know of no reliable alternative.

In its defense, I will say that requiring Kargo to be configured with a signing key, at least, does not undermine the point of the feature. It can be a key generated entirely for the sake of making Kargo trust itself. It does not need to be associated with any GitHub user account.

EronWright:
Seeing that https://github.com/akuity/kargo/pull/5904 was merged, this PR clearly needs to be reworked to comport with it.

PR #5895, opened 2026-03-11
feat(ui): add tooltips to icon-only buttons on stage and warehouse nodes

stoposto:
The icon-only buttons in the stage and warehouse node headers (pipeline graph view) have no tooltips, which makes it unclear what they do - especially for new users. I found myself hesitant to click them when I first started using Kargo.

This wraps them with Ant Design's `<Tooltip>` to add labels:

- Truck icon on stage nodes: **Promote**
- Staggered-bars icon on stage nodes: **Stage Details**
- Staggered-bars icon on warehouse nodes: **Warehouse Details**

It's a small change so I skipped creating an issue first but I'm happy to open one retroactively if that's preferred.

### Screenshots

**"Promote" tooltip on the truck icon:**

promote-tooltip

**"Stage Details" tooltip on the bars icon:**

stage-details-tooltip

### DIFF
--- a/ui/src/features/project/pipelines/nodes/stage-node.tsx
+++ b/ui/src/features/project/pipelines/nodes/stage-node.tsx
@@ -8,7 +8,7 @@ import {
   faTruckArrowRight
 } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { Button, Card, Dropdown, Flex, message, Space, Typography } from 'antd';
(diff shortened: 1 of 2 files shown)


Marvin9:
I think new users still will be hesitant to click even with tooltip as it doesn't convey if it opens dropdown or just perform promotion instantly.

cc @rpelczar

rpelczar:
@Marvin9 Agree. I'm not a font of those tooltips. If we really want to add this, we can add 2s delay at least.

PR #5897, opened 2026-03-11
fix(api): ignore role bindings to cluster roles in roles database

### LINKED ISSUE #4046: Kargo does not support a RoleBinding with a reference to a ClusterRole
# Checklist

# Description

Kargo does not support a RoleBinding with a reference to a ClusterRole. With such a RoleBinding defined it is impossile to use Kargo RBAC. 

With such a role is defined in the project namespace you get the following error:
Role.rbac.authorization.k8s.io "<cluster-role>" not found

The issue appears to be that Kargo looks only for Role objects when a RoleBinding is used and not for a ClusterRole even if it is referenced in the RB.

The role binding in question in our cluster has nothing to do with Kargo, but is a default rolebinding getting created with the namespace. Because of this other rolebinding, RBAC definitons are not being applied in Kargo nor is it possible to view them via UI or CLI.

# Screenshots

Image

# Steps to Reproduce

Create a Rolebinding with a reference to a ClusterRole in a project namespace. 
Try to open the Roles tab in the kargo UI. 

# Version

```
1.4.4
```

# Logs

```
Role.rbac.authorization.k8s.io "<cluster-role>" not found
```

# Suggestions

- Kargo should check the roleRef.kind field and search for the appropriate object (role or clusterRole) 
- Optionaly kargo should only load serviceAccounts/role/rolebindigs that are marked with a custom annotation or label (to avoid loading irrelevant objects)

emmanuel099:
OpenShift creates role bindings from service accounts to cluster roles by default. This broke the assumption of the code, which assumed that role bindings only refer to (plain) roles.

Tested on OpenShift.

Fixes #4046

### DIFF
--- a/pkg/server/rbac/roles.go
+++ b/pkg/server/rbac/roles.go
@@ -369,6 +369,10 @@ func (c *rolesDatabase) GetAsResources(
 	rbs := make([]rbacv1.RoleBinding, 0, len(rbList.Items))
 	for i := range rbList.Items {
 		rb := &rbList.Items[i]
+		if rb.RoleRef.Kind != "Role" {
+			// Ignore RoleBindings that don't reference a Role.
+			continue
+		}
 		for _, subject := range rb.Subjects {
 			if subject.Kind == rbacv1.ServiceAccountKind &&
 				subject.Namespace == namespace &&
--- a/pkg/server/rbac/roles_test.go
+++ b/pkg/server/rbac/roles_test.go
@@ -754,6 +754,67 @@ func Test_rolesDatabase_List(t *testing.T) {
 			kargoRoles,
 		)
 	})
+
+	t.Run("with a non-kargo-managed cluster role", func(t *testing.T) {
+		c := fake.NewClientBuilder().WithScheme(scheme).WithObjects(
+			plainServiceAccount(map[string]string{
+				rbacapi.AnnotationKeyOIDCClaim("sub"):    "foo-sub,bar-sub",
+				rbacapi.AnnotationKeyOIDCClaim("email"):  "foo-email,bar-email",
+				rbacapi.AnnotationKeyOIDCClaim("groups"): "foo-group,bar-group",
+			}),
+			clusterRole([]rbacv1.PolicyRule{
+				{ // This rule has groups and types that we don't recognize. Let's
+					// make sure we don't choke on them. This could happen with cluster
+					// roles that aren't Kargo-managed.
+					APIGroups: []string{"fake-group-1", "fake-group-2"},
+					Resources: []string{"fake-type-1", "fake-type-2"},
+					Verbs:     []string{"get", "list"},
+				},
+				{
+					APIGroups: []string{kargoapi.GroupVersion.Group},
+					Resources: []string{"stages", "promotions"},
+					Verbs:     []string{"list", "get"},
+				},
+			}),
+			plainRoleBindingToClusterRole(),
+		).Build()
+		db := NewKubernetesRolesDatabase(c, RolesDatabaseConfig{})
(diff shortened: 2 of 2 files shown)


Planet-X:
Works well on our infrastructure, OpenShift `builder` and `deployer` service accounts don't pose a problem anymore :+1:

pkg/server/rbac/roles.go
```diff
@@ -369,6 +369,10 @@ func (c *rolesDatabase) GetAsResources(
 	rbs := make([]rbacv1.RoleBinding, 0, len(rbList.Items))
 	for i := range rbList.Items {
 		rb := &rbList.Items[i]
+		if rb.RoleRef.Kind != "Role" {
+			// Ignore RoleBindings that don't reference a Role.
+			continue
```
EronWright, line 374: I don't know much about why these roles are being collected, but is there a "kargo" annotation or label that could be used to identify the intended roles?

emmanuel099: Yes, there is a kargo annotation for kargo managed service accounts. By default we have automatically generated kargo-viewer, kargo-promoter and kargo-admin service accounts which are annotated with `rbac.kargo.akuity.io/managed: "true"`.

In theory, a user can also use non-kargo managed service accounts by adding role bindings to kargo roles, so I fear that filtering service accounts by this annotation may break existing deployments.

My change should be fully backward compatible because this just ignores role bindings to cluster roles, which never worked in the past because the code always assumed "plain" roles.

krancour:
I'm afraid this isn't quite the right fix.

Project admins (or others authorized to see relevant resources) rely  on this to show them _all_ the permissions of a "Kargo Role" virtual resource. (A "Kargo Role" is thin abstraction over SA + RoleBindings + Roles providing a comprehensive view of who can do what with what.)

Instead of showing permissions that one or more ClusterRoles may have contributed to a Kargo Role, this PR is sweeping them under the rug. Focused narrowly on OpenShift having bound the Project namespace's `default` SA to one or more ClusterRoles that have nothing to do with Kargo, this probably seems ok, but it's not really a safe thing to be doing in the general case.

If a SA has certain permissions in my Project namespace because of some ClusterRole(s) its bound to, I don't want that fact hidden from me.

The correct fix is:

1. Kargo Roles are considered "not manageable" if they either lack correct annotations OR are not _precisely_ a trio of 1 SA, 1 Role, and 1 RoleBinding, all with the same name. They're still viewable. They just cannot be mutated through the API.

    We'd need to ensure any Kargo Role with a SA referencing one or more ClusterRoles is counted as "not manageable."

1. Enumerate the permissions contributed to a Kargo Role by bindings to ClusterRoles along with those contributed by bindings to Roles.

    Most Kargo users, even Project admins, have their permissions assigned to them at the Project level and would therefore not ever have permissions to list cluster-scoped things like ClusterRoles. We would, however, want anyone authorized to see "Kargo Roles" to be able to see how any ClusterRoles contribute to its permissions. The workaround for this is to authorize on ServiceAcccount get, then use the API server's own permissions to get ClusterRoles.

    The API server does not currently have permission to get ClusterRoles, so that would have to be added.

    With relevant ClusterRoles in-hand, their rule sets would be fed into existing normalization logic along with the rule sets already fed in from relevant Roles.

I will open a draft of something like this.

krancour:
Closing in favor of #5916

emmanuel099:
> Closing in favor of #5916

Thank you very much, looks much better than my quick fix. :)
 I'll give it a try on Monday. Have a nice weekend.

krancour:
@emmanuel099 thank you for being the forcing function that got me to finally look deeper into #4046.

I test drove it, albeit not with OpenShift, which I don't have access to.

lmk how it goes.

Issue #5899, opened 2026-03-11
feat: shared git repo cache for warehouse discovery to reduce redundant clones


markshields-tl:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [x] I've pasted the output of `kargo version`, if applicable.

# Version

```
v1.9.5
```

# Proposed Feature

Add a shared, per-repo-URL bare git cache for warehouse discovery, so that multiple warehouses subscribing to the same git repository share a single local clone rather than each performing an independent `git clone` on every reconciliation.

# Motivation

We run 57 warehouses that all subscribe to the same monorepo (our `kubernetes-resources` repo — ArgoCD Application manifests + kustomize overlays), each with different `includePaths` to filter for their specific application.

When a webhook fires, all 57 warehouses are triggered to refresh. Each one independently runs `git clone --filter=blob:none --single-branch` into a temp directory, performs commit discovery with path filtering, then deletes the clone. This means:

- **57 independent clones** of the same repository per webhook push
- With `maxConcurrentReconciles: 16`, that's 4 batches of 16 clones
- Our repo has ~15k commits and is ~45MB — each blobless clone takes several seconds
- Total refresh latency is **~90 seconds** from webhook to discovery completion
- GitHub rate limits become a real concern at this volume

This was partially raised in #4465 (closed — the `git log` batching was improved by #4552, but the clone-per-refresh architecture was not addressed). #5591 addresses the complementary problem of webhook fan-out triggering all warehouses.

# Proposed Solution

Maintain a shared bare repo cache keyed by repository URL within the controller:

1. **First access**: `git clone --bare --filter=blob:none` into a persistent cache directory
2. **Subsequent accesses**: `git fetch` to update refs (incremental, typically milliseconds)
3. **Warehouse discovery**: Each warehouse reads from the shared cache (read-only ref lookups + path filtering)
4. **Eviction**: LRU or TTL-based cleanup of unused repos

This would reduce our 57 × full clone to 1 clone + 56 × local ref lookups — an order of magnitude improvement.

The cache directory could be configurable (env var or Helm value) to support ephemeral or persistent storage depending on the operator's preference.

# Related Issues

- #4465 — Original report of suboptimal git operations (closed, partially addressed)
- #5591 — Webhook receiver filter on `includePaths` (open — addresses fan-out, complementary to this)
- #5725 — Sparse checkout regression (different code path but similar monorepo pain)

markshields-tl:
[...]

Issue #4465, opened 2025-06-24
The Kargo Controller is performing suboptimal Git operations


boh-dan:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [ ] I've pasted logs, if applicable.

# Description

The Kargo Controller is performing suboptimal Git operations, which increases commit discovery and resource consumption.

 1. Kargo Controller is running `git clone` for each warehouse sync. It is extra wasteful when working with a monorepo where Kargo Projects are watching over the same repo but different `IncludePaths` (subscription IncludePaths). It would've been faster to clone the repo on first sync but git fetch/git pull on the following syncs.

2. Kargo Controller discovers new commits by iterating over the output of the `git log` command. It runs `git log --skip=$ --max-count=$` to receive a batch of commits. If the desired commit is far behind the `HEAD`, Kargo Controller needs to spawn many `git log ...` processes to discover it. The monorepo can get hundreds of commits a day, which means each warehouse will have to spawn a lot of processes to find the latest N commits.

# Steps to Reproduce

1. Create a Warehouse with a subscription to the git repository.
2. Log git commands executed by Kargo Controller

# Version

```
1.5.1
```

# Logs

```
Paste any relevant application logs here.
```

krancour:
I think this should have been closed by #4552

PR #5900, opened 2026-03-11
feat(controller): shared git repo cache for warehouse discovery

### LINKED ISSUE #5901: feat: shared git repo cache for warehouse discovery
# Checklist

# Version

\`\`\`
v1.9.5
\`\`\`

# Proposed Feature

Add a shared, per-repo-URL bare git cache for warehouse discovery, so that multiple warehouses subscribing to the same git repository share a single local clone rather than each performing an independent \`git clone\` on every reconciliation.

The cache would:
1. Maintain bare git repo clones keyed by repository URL + credential hash
2. On first access: \`git clone --bare\`
3. On subsequent access: \`git fetch\` to update refs (milliseconds for incremental updates)
4. Return independent worktrees per caller so existing selector code works unchanged
5. Evict least recently used entries when at capacity

# Motivation

We run 57 warehouses that all subscribe to the same monorepo (\`kubernetes-resources\` — ArgoCD Application manifests + kustomize overlays), each with different \`includePaths\`.

When a webhook fires, all 57 warehouses refresh. Each one independently runs \`git clone --filter=blob:none --single-branch\` into a temp directory, performs commit discovery, then deletes the clone. This means:

- **57 independent clones** of the same repository per webhook push
- With \`maxConcurrentReconciles: 16\`, that's 4 batches of 16 clones
- Our repo has ~15k commits — each blobless clone takes several seconds of network I/O
- Total refresh latency is **~90 seconds** from webhook to discovery completion
- GitHub rate limits become a concern at this volume

A shared cache would reduce this to 1 clone + 56 local worktree creates (no network I/O), bringing refresh latency down to seconds.

This was partially raised in #4465 (closed — the \`git log\` batching was improved by #4552, but the clone-per-refresh architecture was not addressed). #5591 addresses the complementary problem of webhook fan-out triggering all warehouses before path filtering.

# Related Issues

- #4465 — Suboptimal git operations (closed, partially addressed)
- #5591 — Webhook receiver filter on \`includePaths\` (open)
- #5725 — Sparse checkout regression (different code path, similar monorepo pain)

# Draft Implementation

We have a draft PR at #5900 implementing this as an opt-in \`RepoCache\`. Happy to iterate based on maintainer feedback on the approach.

markshields-tl:
## Context

When multiple warehouses subscribe to the same git repository, each warehouse independently clones the repo on every reconciliation cycle. For monorepo setups (e.g., 57 warehouses watching the same repo with different `includePaths`), this results in 57 redundant `git clone` operations per webhook push — each fetching ~15k commits of metadata from the remote.

With `maxConcurrentReconciles: 16`, this creates 4 batches of 16 clones, leading to ~90 second refresh latency. This also risks hitting GitHub API rate limits.

Closes #5901. Related: #5591, #4465 (supersedes #5899).

## Solution

Introduces an opt-in `RepoCache` that maintains shared bare git repo clones:

- **First access**: `git clone --bare` into a persistent cache directory
- **Subsequent access**: `git fetch` to update refs (milliseconds for incremental updates vs seconds for full clone)
- **Per-caller isolation**: Returns independent worktrees from the shared bare repo, so callers can use them concurrently without conflicts
- **Per-URL locking**: Different repo URLs are not serialized; same-URL access is serialized to protect the bare repo during fetch
- **Credential isolation**: Cache keys are `(repoURL, sha256(credentials))` — different credentials for the same repo get separate entries, preventing cross-project credential leakage
- **LRU eviction**: Configurable `MaxEntries` (default 100). When at capacity, the least recently used entry is evicted. Actively locked entries are skipped during eviction.
- **Lifecycle**: `RepoCache.Close()` cleans up all cached bare repos

The cache is threaded through: warehouse reconciler → subscriber registry → git subscriber → commit selector factory → base selector's `gitCloneFn` override.

## Design Decisions

- **Opt-in**: Passing `nil` for the cache preserves existing clone-per-call behavior. No behavior change by default.
- **Bare repo + worktrees**: The bare repo is the shared cache; worktrees are cheap to create from local data (no network I/O). Each worktree satisfies the existing `git.Repo` interface, so selector code is unchanged.
- **Credential-keyed entries**: Prevents cross-project data leakage by ensuring projects with different credentials get separate bare repo caches, even for the same repository URL.
- **LRU eviction**: Bounds memory/disk usage. Active entries (locked by an in-flight reconciliation) are never evicted.

## Changes

- `pkg/controller/git/repo_cache.go` — New `RepoCache` with per-URL bare clone caching, credential-keyed isolation, LRU eviction, and worktree creation
- `pkg/controller/git/repo_cache_test.go` — Unit tests
- `pkg/controller/git/commit/` — All selector factories accept optional `*RepoCache`
- `pkg/subscription/git.go` — Cache-aware git subscriber factory
- `pkg/subscription/registry.go` — `NewSubscriberRegistryWithGitCache` helper
- `pkg/controller/warehouses/warehouses.go` — Accepts optional `*RepoCache`
- `cmd/controlplane/controller.go` — Passes `nil` (opt-in default)

21 files changed, +473 / -15 lines.

## Testing

All existing tests pass. New tests cover cache creation, entry deduplication, and concurrent access patterns.

---

† — Mort, Mark's AI extension. Review directed by Mark.


markshields-tl:
Hi @krancour, thanks for adding the appropriate labels.

What needs to happen here to get this moving?

krancour:
> What needs to happen here to get this moving?

https://github.com/akuity/kargo/issues/5901#issuecomment-4043835630

> I wish you'd have waited for some discussion with us before charging ahead with an implementation.
>
> I sympathize with the circumstances that pushed you to this, and overall, I think this is a decent idea. That said, I think we're, rightly, going to be incredibly cautious about this. Not in terms of LOC, but in terms of behavior, this is a huge change.
>
> This will most definitely be reviewed, but please understand it may be some time before we can do our own diligence on this.

Patience

markshields-tl:
> > What needs to happen here to get this moving?
> 
> [#5901 (comment)](https://github.com/akuity/kargo/issues/5901#issuecomment-4043835630)
> 
> > I wish you'd have waited for some discussion with us before charging ahead with an implementation.
> > I sympathize with the circumstances that pushed you to this, and overall, I think this is a decent idea. That said, I think we're, rightly, going to be incredibly cautious about this. Not in terms of LOC, but in terms of behavior, this is a huge change.
> > This will most definitely be reviewed, but please understand it may be some time before we can do our own diligence on this.
> 
> Patience

@krancour do you have any updates?

markshields-tl:
nevermind.

PR #5903, opened 2026-03-11
refactor: Replace raw object responses in APIs with kargo and k8s api structures

hairyhum:
Some APIs are still returning `object` because they are not mapping to structures:
```
getProjectRole
getSystemRole
listProjectRoles
listSystemRoles
```
Are returning role OR underlying resources, which is not a single struct type.

`promoteDownstream` returns a map of lists of promotions

`queryFreight` returns a map of groups of freights

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ vendor/
 internal/server/ui/
 .DS_Store
 docs/static/swagger.json
-docs/static/swagger.yaml
--- a/docs/package.json
+++ b/docs/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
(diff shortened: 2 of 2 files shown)


krancour:
I amended this with some changes to how the swagger.json/yaml are generated and how the Go client is generated from those. The CLI was the only existing user of that generated client and the `pkg/cli/tests/e2e.sh` smoke tests pass with flying colors now.

@rpelczar would you mind looking at this branch (don't need to review, so much as test drive) to see if some of the problems you were having before are resolved now? Let us know if more is needed?

fuskovic:
Closes https://github.com/akuity/kargo/issues/5950

krancour:
Note to self to try and improve the names of generated client types.

krancour (approved):

PR #5904, opened 2026-03-12
feat: configurable policies for remote change integration into local branch before push

krancour:
## Summary

The `git-push` step now evaluates GPG signature trust before choosing how to integrate remote changes, preferring rebase for clean, linear history but falling back to merge when a rebase would alter commit signature semantics (e.g. stripping, fabricating, or re-signing with a different trust level). This PR also adds an optional `committer` field to `git-push` configuration, allowing users to specify a name, email, and GPG signing key for any merge or replacement commits created during remote integration.

## How it works

When the push step needs to integrate remote changes (i.e. the remote branch has new commits), it inspects each local commit that would be replayed by a rebase:

| Local commit status | Signing configured? | Strategy | Rationale |
|---|---|---|---|
| Signed, trusted key | Yes | Rebase | Committer can re-sign the replacement commits |
| Unsigned | No | Rebase | Replacement commits stay unsigned — no change in trust |
| Signed, trusted key | No | Merge | Would strip valid signatures |
| Unsigned | Yes | Merge | Would add signatures to commits the committer didn't author |
| Signed, untrusted key | Either | Merge | Can't vouch for these commits regardless |

In idiomatic Kargo usage, the commits being replayed are typically ones Kargo created, so rebase remains the common path. The merge fallback is a safeguard for edge cases.

__Edit:__ Just to catch the description up to the current state, this PR offers the operator a choice of four options on the permissive <--> strict spectrum, with a plan to keep the default as it currently is until v1.12 so the (very slight) change in behavior will have been well-advertised before it becomes the default.


docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
```diff
@@ -43,19 +64,23 @@ system to access the git repos.
...
 | `generateTargetBranch` | `boolean` | N | Whether to push to a remote branch named like `kargo/promotion/<promotionName>`. If such a branch does not already exist, it will be created. A value of `true` is mutually exclusive with `targetBranch` and `tag`. If none of these are provided, the target branch will be the currently checked out branch. This option is useful when a subsequent promotion step will open a pull request against a Stage-specific branch. In such a case, the generated target branch pushed to by the `git-push` step can later be utilized as the source branch of the pull request. |
+| `committer` | `object` | N | Optional committer information for merge commits or replacement commits created when integrating remote changes before pushing. If provided, this takes precedence over both system-level defaults and any default authorship information configured in the [`git-clone`](./git-clone.md) step. |
+| `committer.name` | `string` | Y | The committer's name. |
+| `committer.email` | `string` | Y | The committer's email address. |
+| `committer.signingKey` | `string` | N | A GPG signing key for the committer. This field is optional. |
```
fuskovic, line 104: Does it make sense to use `pusher` here? Out of scope for this PR but possibly even change `git-commit` step to use `committer` instead of `author`? 

Rationale being symmetry across steps e.g.

`git-push` -> pusher
`git-commit` -> committer
`git-tag` -> tagger

pkg/controller/git/base_repo.go
```diff
@@ -403,6 +410,49 @@ func (b *baseRepo) URL() string {
 	return b.originalURL
 }
 
+// setUltimateTrust sets ultimate trust (level 6) on a GPG key so that
+// signatures made by this key are considered fully trusted during verification.
+func (b *baseRepo) setUltimateTrust(homeDir, keyPath string) error {
+	// Get the fingerprint of the key using --with-colons output format.
+	// The fingerprint line starts with "fpr" and the fingerprint is in field 10.
```
fuskovic: I think this comment line should be closer down to where the string parsing acrobatics take place.

pkg/controller/git/base_repo.go
```diff
@@ -403,6 +410,49 @@ func (b *baseRepo) URL() string {
...
+		return fmt.Errorf("error getting fingerprint of gpg key: %w", err)
+	}
+	var fingerprint string
+	for _, line := range strings.Split(string(res), "\n") {
+		fields := strings.Split(line, ":")
+		if len(fields) >= 10 && fields[0] == "fpr" {
+			fingerprint = fields[9]
+			break
+		}
+	}
```
fuskovic, line 443: I think an input example in lieue of `The fingerprint line starts with "fpr" and the fingerprint is in field 10` provides more context. It's not as succinct but the visual makes the code easier to reason about.

pkg/controller/git/rebase_safety.go
```diff
@@ -0,0 +1,330 @@
...
+					"path", homeDir,
+				)
+		}
+	}
+	if err = w.setupAuthor(homeDir, committer); err != nil {
+		cleanup()
+		return "", noop, fmt.Errorf(
+			"error setting up committer for pull command: %w", err,
+		)
+	}
```
fuskovic: I think it's more idiomatic to just defer the cleanup logic here. Since `defer` captures variables by reference you can just add the err check inside since we reassign on `w.setupAuthor` e.g.

```suggestion
	defer func() {
	    if err != nil {
	        if cleanErr := os.RemoveAll(homeDir); cleanErr != nil {
  			    logging.LoggerFromContext(context.TODO()).
  				    Error(
  					    cleanErr,
  					    "error removing virtual home directory",
  					    "path", homeDir,
  				    )
  		    }
	    }
	}()
	if err = w.setupAuthor(homeDir, committer); err != nil {
		return "", noop, fmt.Errorf(
			"error setting up committer for pull command: %w", err,
		)
	}
```

pkg/controller/git/rebase_safety.go
```diff
@@ -0,0 +1,330 @@
...
+			"fetch", "origin",
+			fmt.Sprintf(
+				"+refs/heads/%s:refs/remotes/origin/%s",
+				targetBranch, targetBranch,
+			),
+		)); fetchErr != nil {
+			return nil, fmt.Errorf(
+				"error fetching remote branch %q: %w", targetBranch, fetchErr,
+			)
+		}
```
fuskovic: Once https://github.com/akuity/kargo/pull/5836 is merged I think we can use `w.Fetch` here e.g.


```suggestion
                opts := &FetchOptions{TargetBranch: targetBranch, Depth: 1})
		if _, fetchErr := w.Fetch(opts); fetchErr != nil {
			return nil, fmt.Errorf(
				"error fetching remote branch %q: %w", targetBranch, fetchErr,
			)
		}
```

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/git-push.md
krancour: I know that at a glance, it looks inconsistent the way it is now, so I think you're right to question it, but with respect to actual git terminology, we're accurately reflecting the options' (highest) uses.

* When making a brand new commit, the user is, formally, both the committer and, more significantly, the author.

  ```
  $ git cat-file -p 56da2bfca13fbdc0583e95524dac2b71432db324

  tree c6e74db119b386086f244243b8394eed1b866aa0
  parent a24a46a6d977005069a7eb9948bca757ab491379
  author Kent Rancourt <kent.rancourt@gmail.com> 1773287232 -0400
  committer Kent Rancourt <kent.rancourt@gmail.com> 1773287351 -0400

  # ...
  ```

* When making a new annotated tag, the user is, formally, the tagger.

  ```
  $ git cat-file -p v1.9.5

  object fd25620c2473ed19bec4be4d0f181287ef0f0391
  type commit
  tag v1.9.5
  tagger Kent Rancourt <kent.rancourt@gmail.com> 1772481716 -0500

  # ...
  ```

* User/key isn't directly involved in a push, but because we potentially rebase or merge before before push, the user is a committer, although not necessarily the author.

pkg/controller/git/rebase_safety.go
krancour: Agree. Whichever is merged second can account for that.

pkg/controller/git/rebase_safety.go
krancour: Deferring cleanup in this way would clean up the temporary home dir before it's ever used by the commands that called this method. 😉 

Having said that, that's only true because temporary home setup is being done in its own function here instead of being done more directly, as it is in other places. This is an artifact of an interim version of this code where this method was called from multiple places. It no longer is. So the more direct approach would be better here and then what you're proposing would work.

Fixed in 7a247d4121693fae15760f95e153c6b16305ed9b

pkg/controller/git/base_repo.go
krancour: Fixed in 004ba979bb24b8ddb70fbe099cc6bcb45a796e8c

pkg/controller/git/base_repo.go
krancour: Fixed in 004ba979bb24b8ddb70fbe099cc6bcb45a796e8c

pkg/controller/git/rebase_safety.go
krancour: #5836 won the race, so I'll refactor this once it clears the merge queue.

krancour: Fixed in 6d3284871fc1fa9b2d7bb476189f76b96bb40ae8 along with a related bug -- fetch should be unconditional because even if we already had a remote tracking ref, it might have been stale.

krancour:
@fuskovic thanks so much for the review. I will hold off on merging until @EronWright also gives his ok since he's done a lot of recent work in the signing/trust space.

thomastaylor312:
Sorry it took a bit for me to think through this and get back to it. Although I like the idea of eventually supporting user provided signing keys, I think if we don't do this fix we end up the same problem with rewriting signing details. This was the example I thought of: Let's say a company has strict security requirements on who signs what. Now lets say we have this commit graph:

- fix: hotfix for prod **<- Current main that was pushed after you started a promotion**
   signed by sally@acme.com <- key is not trusted directly by kargo
- feat: new version **<- main at the time the promotion started**
   signed by kargo-key@acme.com

With the current logic, when Kargo re-signs commits for something it is pushing, it now looks like this:

- feat: another kargo change from promotion
   signed by kargo-key@acme.com
- fix: hotfix for prod
   signed by kargo-key@acme.com <- was resigned because the key wasn't trusted
- feat: new version
   signed by kargo-key@acme.com

This is an issue because we have now _overwritten_ the signed commit. Did Sally sign off on this or did kargo? This can happen in several use cases (this is basically restating what @krancour was saying in his original table) based on my understanding of how the pull+rebase would work:

1. **Unsigned local commit + signing configured** Kargo didn't sign the original commit (maybe it was made by a different step, or signing wasn't configured at commit time), but now during rebase, commit.gpgSign=true would fabricate Kargo's signature on a commit Kargo didn't vouch for. This is the most unlikely but is still a possibility
2. **Signed with untrusted key + either config** A commit signed by a key Kargo doesn't trust gets replayed on the rebase. If signing is on, Kargo re-signs it (essentially laundering trust). If signing is off, the signature is stripped.
3. **Signed with trusted key + signing not configured** A validly signed commit would have its signature stripped by the rebase because the replayed commit won't be re-signed.

**The tl;dr for me in why I think I agree that this is necessary** There are several ways we might either _strip_ existing trust or _add_ trust that shouldn't necessarily exist whether by mistake or someone trying to exploit the process. That shouldn't happen with a default setting

Yes, I know that automatic signing is Good Enough™ for a large chunk of users (possibly above 90%), especially if they are following completely idiomatic use cases. But often times people are committing to a busy branch or committing a hotfix that could go out during a long-running promotion and that is what this would solve for. It is definitely an edge case I'll admit, but we have a growing user base of many "high security" environments like banks, telcoms, etc. where messing with signatures will likely be a problem. So I think I'm in favor of this with the caveat that we should allow an opt-out for users who don't care about yolo-ing the commit signing so they can preserve the original behavior

krancour:
> So I think I'm in favor of this with the caveat that we should allow an opt-out for users who don't care about yolo-ing the commit signing so they can preserve the original behavior

I agree with that, and that's not in here, so I would happily add that.

EronWright:
I think the concept of "author" and "committer" are being conflated here. In the example, sally@acme.com remains the author end-to-end.  To quote [GitLab](https://docs.gitlab.com/user/project/repository/signed_commits/):

> The committer and author fields are distinct in Git. The author writes the commit, and the committer applies it. Commit signing verifies only the committer’s identity.

A Kargo promotion is applying commits and signing them, and preserving authorship.  There's simply nothing untoward happening here, is there?

As an anecdote: when you edit a file using the GitHub web interface, the resultant commit is signed by GitHub's GPG key (not by your own key). Did GitHub "launder" your commit? No. The word launder has undue connotations in my opinion and is misleading us.

krancour:
> As an anecdote: when you edit a file using the GitHub web interface, the resultant commit is signed by GitHub's GPG key (not by your own key). Did GitHub "launder" your commit? No. The word launder has undue connotations in my opinion and is misleading us.

This is quite a different scenario. GitHub can sign that commit because GitHub knows who you are. You can't have edited a file using the web interface without being authenticated.

EronWright:
Now you're implicating the trust model of Kargo's promotion templates. There's authentication and authorization to allow promotions to be defined, and that confers a degree of trust (to use the system as intended).  That the system would sign the commits that it applies as a result, is wholly reasonable to me.

I think the real problem here is that all projects have access to a shared signing key.  It's possible already to use project-level signing keys, so that if a promotion does a Bad Thing, it it attributed to that project.

Another idea is to add trailers to outgoing commits like `Kargo-Promotion-ID` or what-not.

krancour:
> Now you're implicating the trust model of Kargo's promotion templates. There's authentication and authorization to allow promotions to be defined, and that confers a degree of trust (to use the system as intended). That the system would sign the commits that it applies as a result, is wholly reasonable to me.

There's a distinct difference between authorizing Kargo to sign its own commits vs authorizing it to sign arbitrary and untrusted commits by other authors.

> I think the real problem here is that all projects have access to a shared signing key. It's possible already to use project-level signing keys, so that if a promotion does a Bad Thing, it it attributed to that project.

Project-level signing keys don't solve the problem. It changes whose signature is applied to the commits in question, but doesn't change the fact of that signature attesting to the authenticity of commits that Kargo couldn't verify.

I'm certain there's a problem here that isn't abstract or theoretical. It's trivial to implement a promotion process that opens the door to Kargo signing arbitrary commits with unverified provenance. Although I would consider such a process to fall outside of idiomatic usage, users do much stranger things than this. Just because most people come through the front door doesn't mean we can leave the side door unlocked.

---

As to the solution:

> How is signing the merge commit any different than re-playing the original commits?

It's substantially different. A merge commit says, "I, Kargo, merged these other commits." Those other commits are not re-written or re-signed. Original attestations remain 100% intact for verification downstream from Kargo.

> I would like a solution that interoperates well with [Argo CD signature verification](https://oneuptime.com/blog/post/2026-02-26-argocd-require-signed-commits/view)

For what little it's worth, this solution does. I say "for what little it's worth" because Argo CD signature verification only checks the signature of the commit at a branch's head. In this case, that's the merge commit, which is signed by Kargo. So assuming Argo CD trusts Kargo, it works, although, arguably, Argo CD isn't verifying signatures rigorously enough. There can be many unsigned commits or commits with untrusted signatures _behind_ that head commit and Argo CD never even looks at them.

Contrast this with GitHub. When branch protections require verified signatures, GitHub verifies _all the commits_ in a push before permitting it to proceed. And that matters a lot. This is where the real distinction between the rebase and the merge commit is evident. _The rebase can lend credibility to commits not authored by Kargo that GitHub would have rejected if it ever got to see the originals._ The merge commit, however, would allow GitHub to see and reject those.

---

> I think it problematic that it would dynamically switch between rebase and merge. 

I see less of a problem here than you do. For a well-designed promotion process, you'd always get a rebase. There would be no whiff of the possibility of a merge commit.

The "dynamism" here only manifests in cases that I would consider non-idiomatic usage of Kargo. But it's important to note that it's _not_ non-deterministic. Behavior is consistently driven by well-defined rules that put security first.

Having said that, I'll readily concede that a user who's not prepared to dig into the _why_ of sometimes getting a rebase and sometimes getting a merge could be very confused by this. And it's a point well taken.

And it may have inspired @thomastaylor312's comment:

> So I think I'm in favor of this with the caveat that we should allow an opt-out for users who don't care about yolo-ing the commit signing so they can preserve the original behavior

I readily agreed with this, but now think we can do better than that even.

---

Where I'm at on this now:

There's a problem here worth fixing. Or if we don't want to look at it as a problem, there's an _opportunity_ to tighten things up a little bit more to make us more suitable for extremely rigorous environments.

Point taken that many users may not care. 🤷‍♂️ (imho, shame on them. Supply chain integrity isn't a joke.) We can't force people to care. We _can_ nudge them gently in the right direction.

It would be reasonable to offer a spectrum of choices to the operator:

1. AlwaysRebase: YOLO

    Keep this as the default for now. Advertise in the v1.10.0 and v1.11.0 release notes that this will change in v1.12.0.

1. RebaseOrMerge: Rebase if it won't degrade trust. Merge otherwise.

    A slightly better option that puts commit integrity ahead of linear history. This would be a sensible default in the future -- starting with v1.12.0. If anyone missed the memo, this won't break them. At worst it will make them scratch their head and wonder why they have merge commits now.

1. RebaseOrFail: Rebase if it won't degrade trust. Fail otherwise.

    This is a great option if you work with carefully designed promotion processes that are what we'd consider idiomatic, in which case the non-rebase scenario _shouldn't_ arise. If it does, _something is wrong and a human should look at it._

1. AlwaysMerge: Uglier history, but secure and predictable.

Issue #5672, opened 2026-01-30
generated docs for new restful api


krancour:
Supersedes #4956 

We need to automate generation of docs for the new RESTful API from swagger.json.

Ideally, it would be markdown that slots smoothly into the Docusaurus site. Ideally, would include curl-based examples as well. But it's fine also to get _something_ out there sooner rather than later and improve it incrementally.

PR #5908, opened 2026-03-12
feat: embed swagger docs into docusauras

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5672

### DIFF
--- a/docs/docs/90-api-documentation.md
+++ b/docs/docs/90-api-documentation.md
@@ -1,4 +1,4 @@
-# API Documentation
+# gRPC API Documentation (Deprecated)
 <a name="top"></a>
 
 <a name="api_service_v1alpha1_service-proto"></a>
--- a/docs/api/docs.templ
+++ b/docs/api/docs.templ
@@ -1,4 +1,4 @@
-# API Documentation
+# gRPC API Documentation (Deprecated)
 <a name="top"></a>
 
 {{- range .Files}}


docs/docusaurus.config.js
```diff
@@ -74,9 +74,22 @@ const config = {
...
+          petstore: {
+            specPath: "static/swagger.yaml",
+            outputDir: "docs/90-api-docs",
+          },
+        },
+      },
+    ],
   ],
-
+  themes: ["docusaurus-theme-openapi-docs"],
```
krancour: As far as I am able to tell, these aren't actually doing anything.

`docusaurus-plugin-openapi-docs`, by my understanding, renders static doc pages at build time, and that appears not to be happening -- in fact, this even appears to be misconfigured.

`swagger-ui-react` seems to be what's actually rendering the docs, dynamically, at runtime.

I can easily imagine things only worked with `docusaurus-plugin-openapi-docs` installed, but that seems to be because it was transitively importing polyfills that `swagger-ui-react` needs, which weren't explicitly listed as dependencies. With a little experimentation, I was able to make all of this work with this config and with the `docusaurus-plugin-openapi-docs` removed.

Now all that being said...

As I understand it, there are some other alternatives to `swagger-ui-react` like `@scalar/docusaurus`, for example, that don't rely on polyfills. They might be worth a look. We might want to get @rpelczar's opinion on whether we're best avoiding polyfills.

docs/docs/90-api-documentation.md
krancour, line 1: I think we need to keep this until the gRPC API is actually gone.

swagger.json
krancour, line 1: Given the importance of these this file and `swagger.yaml`, I'd find it preferable to keep these in the root of the repo, but I know that, there, they'd be inaccessible as static content that the API docs are dynamically generated from.

My best recommendation would be to keep them in the root of the repo and copy them to `docs/static/` during the site build or run.

Something like this in package.json:

```
"scripts": {
  ...
  "copy-swagger": "cp ../swagger.yaml ../swagger.json static/",
  "prestart": "pnpm copy-swagger",
  ...
  "prebuild": "pnpm copy-swagger",
  ...
},
```

You can add these to .gitignore:

```
docs/static/swagger.json
docs/static/swagger.yaml
```

docs/src/components/Swagger.js
```diff
@@ -0,0 +1,15 @@
...
+
+const store = configureStore({ reducer: {} });
+
+export default function Swagger() {
+    return (
+        <Provider store={store}>
+            <SwaggerUI url="/swagger.yaml" />    
+        </Provider>
+    );
+}
```
krancour: Combined with changes to the CSS (see below), this will hide interactive "try it out" options.

```suggestion
import BrowserOnly from '@docusaurus/core/lib/client/exports/BrowserOnly';
import React from 'react';

const ReadOnlyAuthPlugin = () => ({
    wrapComponents: {
        authorizeBtn: () => () => null,
    },
});

export default function Swagger() {
    return (
        <BrowserOnly>
            {() => {
                const SwaggerUI = require('swagger-ui-react').default;
                require('swagger-ui-react/swagger-ui.css');
                return (
                    <SwaggerUI
                        url="/swagger.yaml"
                        supportedSubmitMethods={[]}
                        plugins={[ReadOnlyAuthPlugin]}
                    />
                );
            }}
        </BrowserOnly>
    );
}

```

docs/src/css/custom.css
```diff
@@ -103,3 +103,7 @@ main.landing-page {
   font-weight: normal;
   font-size: x-small;
 }
+
+.swagger-ui .info {
+  display: none;
+}
```
krancour: See https://github.com/akuity/kargo/pull/5908/changes#r2949984867

```suggestion
.swagger-ui .authorization__btn {
  pointer-events: none;
  cursor: default;
  opacity: 0.6;
}```

File: pkg/server/rest_router.go [hunk 4]
```diff
@@ -15,6 +15,7 @@ import (
 // @title Kargo API
 // @version v1alpha1
 // @description REST API for Kargo
+// @host localhost:30081
```
krancour: We shouldn't do this... no one will have the API running at this address unless they're actively hacking on Kargo. The conventional way to handle docs like this for self-hosted APIs is to leave this off and make the docs non-interactive. See https://github.com/akuity/kargo/pull/5908/changes#r2949984867

docs/docusaurus.config.js
rpelczar: @krancour, you did a really good research here... I think it's a great idea to try `@scalar/docusaurus`. If we like it, we avoid having to explicitly manage polyfill dependencies.
In this code, it's not clear why it was added here.

krancour: Thanks @rpelczar!

@fuskovic this may change some of the other suggestions I made. The tweaks for disabling interactive requests would obviously be different for a different framework.

docs/docusaurus.config.js
```diff
@@ -74,9 +74,19 @@ const config = {
...
       },
     ],
-    'docusaurus-plugin-sass'
+    'docusaurus-plugin-sass',
+    [
+      '@scalar/docusaurus',
+      {
+        label: 'API Reference',
+        route: '/api-docs',
+        showNavLink: true, // optional, default is true
```
krancour:
```suggestion
        showNavLink: false, // Don't include in the top navbar
```

docs/docusaurus.config.js
```diff
@@ -74,9 +74,19 @@ const config = {
...
+        configuration: {
+          url: '/swagger.json',
```
krancour, line 88: I found the secret sauce:

```suggestion
          url: '/swagger.json',
          servers: [{
            url: '{baseUrl}',
            variables: {
              baseUrl: {default: 'https://kargo.example.com'}
            }
          }]
```

docs/src/css/custom.css
```diff
@@ -103,3 +103,9 @@ main.landing-page {
   font-weight: normal;
   font-size: x-small;
 }
+
+.swagger-ui .authorization__btn {
+  pointer-events: none;
+  cursor: default;
+  opacity: 0.6;
+}
```
krancour: This was applicable for `swagger-ui`, but not needed for `@docusaurus/scalar`.


```suggestion

```

docs/docusaurus.config.js
krancour: Could move it into the sidebar with something like this in `sidebar.js`, but you'd also have to also add some kind of filter to the sidebar builder so the existing gRPC docs don't show up twice.

```
    {
      type: 'category',
      label: 'API Reference',
      collapsible: true,
      collapsed: true,
      items: [
        {
          type: 'link',
          label: 'REST API',
          href: '/api-docs',
        },
        {
          type: 'doc',
          id: 'api-documentation',
          label: 'gRPC API (Deprecated)',
        },
      ],
    },
```

Makefile
```diff
@@ -273,6 +273,7 @@ codegen-ui:
 .PHONY: codegen-docs
 codegen-docs:
 	npm install -g @bitnami/readme-generator-for-helm
+	pnpm install --dir docs
```
krancour, line 276: Curious about this. Is something being generated during the codegen process? I thought scalar was rendering the docs dynamically from the swagger.json at runtime... although, tbh, I'm really not _positive_ that's what it's doing.

krancour:
This looks fantastic and really useful. I'm glad we were able to keep it interactive.

The only, very, very small problem I am noticing is the legacy docs are in the sidebar twice now:

Screenshot 2026-03-19 at 9 22 38 AM

I suppose we could go one of two ways:

1. Abandon the collapsible "API Reference" section in the doc tree and only add a link to the REST API reference docs at the top level. It's not quite as nice as having both links in the collapsible section, but especially since the gRPC AGPI is deprecated and being removed within a few releases, this is a short-term ugliness that I could live with.

1. Alternatively (according to Claude, at least; I haven't verified this), it's relatively simple to insert a filter into the sidebar builder that will exclude anything from the auto-generated portion of the sidebar for which we have explicitly added a link of our own to the sidebar.

I think no.2 sounds nicer, but I wouldn't spend a lot of time on it before falling back on no.1.

Either way though, I think we need to add a deprecation notice to the gRPC docs, or even include `(deprecated)` directly in its title. However the reader gets to that page, we don't want that detail to be overlooked.

Makefile
fuskovic: It makes sure that the dependencies for the docs are installed. Similarly to the `codegen-ui` target.

PR #5911, opened 2026-03-13
fix(ui): add smooth scrolling to freight list

henrycatalinismith:
We're finding scrolling the freight list somewhat disorienting in my team. As each item within is so visually similar to its neighbours, the instantaneous scroll means your eyes kind of lose context of the relationship between where you were before the scroll and where you've landed after it. It produces an optical illusion where the list appears to scroll in the opposite direction than it really is.

Adding a slight transition to smooth it seems to address the issue and improves the usability of this part of the UI quite nicely.


Marvin9:
~As of now, you can also scroll on wheel or using trackpad when hovering over freight timeline. Would that work?~

Oh anyways.. this is improvement on top of that

PR #5916, opened 2026-03-13
fix(api server): show what permissions clusterroles contribute to a kargo role

### LINKED ISSUE #4046: Kargo does not support a RoleBinding with a reference to a ClusterRole

krancour:
Supersedes #5897

Fixes #4046

See https://github.com/akuity/kargo/pull/5897#issuecomment-4057335473 for details.

### DIFF
--- a/pkg/server/rbac/roles.go
+++ b/pkg/server/rbac/roles.go
@@ -57,7 +57,7 @@ type RolesDatabase interface {
 		systemLevel bool,
 		project string,
 		name string,
-	) (*corev1.ServiceAccount, []rbacv1.Role, []rbacv1.ClusterRole, []rbacv1.RoleBinding, error)
+	) (Resources, error)
 	// GrantPermissionsToRole amends the Role underlying a Kargo Role with new
 	// rules. It will return an error if no underlying ServiceAccount exists or
 	// any underlying resources are not Kargo-manageable. It will create
@@ -145,6 +145,15 @@ type RolesDatabase interface {
 		project string,
 		roleName string,
 	) ([]corev1.Secret, error)
+}
+
+// Resources is a struct that encapsulates the Kubernetes resources underlying a
+// Kargo Role.
+type Resources struct {
+	ServiceAccount *corev1.ServiceAccount
+	Roles          []rbacv1.Role
+	ClusterRoles   []rbacv1.ClusterRole
+	RoleBindings   []rbacv1.RoleBinding
 }
 
 // rolesDatabase is an implementation of the RolesDatabase interface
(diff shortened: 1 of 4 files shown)


emmanuel099:
Our platform team tested this on OpenShift, works as expected.

The OpenShift cluster roles `deployer` and `builder` are now visible in Kargo. No more errors when accessing projects.

image

Definitely much better than https://github.com/akuity/kargo/pull/5897, thanks again!

krancour:
Thanks for verifying @emmanuel099!

pkg/server/rbac/roles.go
```diff
@@ -345,14 +348,14 @@ func (c *rolesDatabase) GetAsResources(
 
 	sa := &corev1.ServiceAccount{}
 	if err := c.client.Get(ctx, objKey, sa); err != nil {
-		return nil, nil, nil, fmt.Errorf(
+		return nil, nil, nil, nil, fmt.Errorf(
```
fuskovic: NIT: I've looked at this code before and thought the function signature had too many return values. 
I think consolidating the return values into a single struct type could clean this up a bit e.g.

```go
type RbacResources struct {
     serviceAccount
     roles
     clusterRoles
     roleBindings
}
```

krancour: I also wasn't crazy about it and almost had done exactly what you're suggesting. I should. Or I will, rather.

krancour: Refactored in 14814fbb866a5163f58ed323fed5cf8d2719aea7

Issue #5950, opened 2026-03-19
docs: REST API doc improvement suggestions


fuskovic:
# Description

Tag descriptions, response payloads, and request param/fields need to provide more information.

# Screenshots

Some things @kencochrane pointed out to me.

Native Kubernetes and Kargo types don't render the full response bodies, they just say object.

<img width="535" height="231" alt="Image" src="https://github.com/user-attachments/assets/b427b137-5e24-4da0-b8ed-483a6d894e93" />

Tag sections don't provide a description and thus lead to a large amount of white space. 

<img width="1069" height="386" alt="Image" src="https://github.com/user-attachments/assets/2d5b6da1-1c7d-43ce-9b00-951d0200cd75" />

Path params could use more descriptive descriptions:

<img width="524" height="187" alt="Image" src="https://github.com/user-attachments/assets/34b81872-cd63-4e0e-9d46-03fb0cef4360" />

krancour:
> Native Kubernetes and Kargo types don't render the full response bodies, they just say object.

That's a workaround for some incompatibilities between k8s resource types and swag. It's being addressed already in https://github.com/akuity/kargo/pull/5903, so there's no action required here to resolve that.

kencochrane:
Faris has made some progress on this recently; there is still some more to go.

krancour:
#5903, merged just now, will move the needle on this further.

krancour:
@fuskovic do you know if this can be closed at this point? You know the status of this better than I do.

PR #5966, opened 2026-03-23
feat: swagger tag description annotations

fuskovic:
Part of https://github.com/akuity/kargo/issues/5950

### Before

Screenshot 2026-03-23 at 6 17 48 PM

### After

Screenshot 2026-03-23 at 7 22 56 PM

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -5156,7 +5156,7 @@
             "name": "Shared"
         },
         {
-            "description": "Only one instance of this resource kind exists per scope. For example, each project has exactly one ProjectConfig and each cluster has exactly one ClusterConfig.",
+            "description": "At most, one instance of this resource kind exists per scope. For example, each project has, at most, one ProjectConfig and each cluster has, at most, one ClusterConfig.",
             "name": "Singleton"
         }
     ]
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -3578,6 +3578,7 @@ tags:
     scopes. For example, a shared ConfigMap or credential can be referenced by any
     project without being duplicated per namespace.
   name: Shared
-- description: Only one instance of this resource kind exists per scope. For example,
-    each project has exactly one ProjectConfig and each cluster has exactly one ClusterConfig.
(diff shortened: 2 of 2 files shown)


fuskovic:
This PR has a commit that solves the same issue as https://github.com/akuity/kargo/pull/5903.

However, I reverted the commit in favor of the aforementioned PR.

Although, I kept it in the commit history so reviewers can clone this branch, checkout the commit in detached HEAD mode and see how the change looks locally.

The commit is https://github.com/akuity/kargo/pull/5966/commits/929c1a4de46e3f51918ddcebe482de2e6f5f46ca.

pkg/server/rest_router.go
```diff
@@ -12,10 +12,43 @@ import (
...
+// @tag.name Project-Level
+// @tag.description The resource or operation is scoped to a single Kargo project (Kubernetes namespace). Access is governed by the requesting user's permissions within that project.
+// @tag.name System-Level
+// @tag.description The resource or operation lives in the Kargo system namespace and applies across all projects. Typically restricted to cluster administrators.
+// @tag.name Cluster-Scoped Resource
+// @tag.description The resource is a Kubernetes cluster-scoped object (not namespaced), such as a Project or ClusterAnalysisTemplate. These exist at the cluster level and are accessible to users with appropriate cluster-wide permissions.
+// @tag.name Shared
+// @tag.description The resource is defined once and shared across multiple projects or scopes. For example, a shared ConfigMap or credential can be referenced by any project without being duplicated per namespace.
+// @tag.name Singleton
+// @tag.description Only one instance of this resource kind exists per scope. For example, each project has exactly one ProjectConfig and each cluster has exactly one ClusterConfig.
```
krancour:
```suggestion
// @tag.description At most, one instance of this resource kind exists per scope. For example, each project has, at most, one ProjectConfig and each cluster has, at most, one ClusterConfig.
```

krancour: Other than this, LGTM. I might want to revisit what tags we've put on each endpoint, but I see that as separate from this. This is great!

PR #5977, opened 2026-03-25
docs: modest roadmap updates

krancour:


krancour:
@fuskovic great question. To answer, I try to keep the roadmap light on details and focused on release themes with only the very biggest highlights called out. Deprecations and breaking changes will be called out prominently in release notes.

PR #5978, opened 2026-03-25
chore: deprecate ssh urls and keys

### LINKED ISSUE #5858: deprecate ssh urls
This is mentioned in quite a number of places, but perhaps discussed most extensively in #5457.

@jessesuen and I agreed that we'll deprecate using ssh urls and keys for accessing git repositories, but it will be several releases (more than we'd usually wait) before it's removed. We are deliberately creating ample room for the community to convince us _not_ to remove it and to de-deprecate it, but the bar for convincing us may be high.

Before anyone starts dumping their use cases here, please be sure you understand the rationale laid out in #5457. The tl;dr is that ssh keys cannot be used to authenticate to the APIs of any of the major git hosting providers. If you use ssh for things like cloning repos and pushing commits, you will need a second, different kind of kind of credential (e.g. a PAT of some sort) for common API operations like opening PRs. It is both a better UX and more secure to use https URLs only and thereby get away with handling a single set of creds instead of two.

If we're overlooking something, we're willing to hear it.

krancour:
Closes #5858

### DIFF
--- a/docs/docs/40-operator-guide/40-security/40-managing-secrets.md
+++ b/docs/docs/40-operator-guide/40-security/40-managing-secrets.md
@@ -137,20 +137,20 @@ SSH-style URLs of the form `git@github.com:example/repo.git`), the key
 `sshPrivateKey` in the `Secret`'s `data` block may have as its value a
 PEM-encoded SSH private key.
 
-:::warning[Not Recommended]
+:::warning[Deprecated]
 
-While an SSH private key is an adequate credential for basic Git operations that
-are formally part of the Git specification (i.e. `clone`, `checkout`, etc.), the
-proprietary APIs offered by the major git hosting platforms (e.g. GitHub or
+Support for SSH URLs and SSH private keys is **deprecated as of v1.10.0** and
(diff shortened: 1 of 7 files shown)


docs/docs/40-operator-guide/40-security/40-managing-secrets.md
```diff
@@ -137,20 +137,20 @@ SSH-style URLs of the form `git@github.com:example/repo.git`), the key
...
-:::warning[Not Recommended]
+:::warning[Deprecated]
 
-While an SSH private key is an adequate credential for basic Git operations that
-are formally part of the Git specification (i.e. `clone`, `checkout`, etc.), the
-proprietary APIs offered by the major git hosting platforms (e.g. GitHub or
+Support for SSH URLs and SSH private keys is **deprecated as of v1.10.0** and
+**scheduled for removal in v1.13.0**.
+
+The proprietary APIs offered by the major git hosting platforms (e.g. GitHub or
```
krancour:
```suggestion
The proprietary APIs offered by the major Git hosting platforms (e.g. GitHub or
```

PR #5982, opened 2026-03-26
feat: replicated resources follow ups: immutable replicas, docs, ui changes

### LINKED ISSUE #5954: ensure immutability of replicated resources using webhook
Follow-up to #5789

Replicated Secrets and ConfigMaps (those with the `replicated-from` label) should be immutable to end users. Currently, nothing prevents a user from modifying or deleting them via kubectl, which could cause silent drift from the source resource. We don't detect or correct for drift, so this is a problem.

Add a ValidatingWebhookConfiguration that intercepts mutations to Secrets and ConfigMaps bearing the `replicated-from` label:

- **UPDATE** operations: `failurePolicy: Fail` — block modifications to replicated resources
- **DELETE** operations: `failurePolicy: Ignore` — block deletions under normal circumstances, but degrade gracefully if the webhook service is unavailable, so replicated resources don't become impossible to clean up

The UI should also detect replicated resources and not present them as editable.

Full context: https://github.com/akuity/kargo/pull/5789#issuecomment-4101087415

__Edit:__ Follow-up also needs to add docs for the feature.

krancour:
Closes #5954

Follows up on #5789

* Make replicas immutable
* Adds docs
* UI changes to prevent even attempting to mutate resource replicas

@jessesuen webhooks are implemented exactly as we'd discussed. Deletes don't fail if the webhook server is down.

@rpelczar the UI changes are all Claude and seem minimally viable. I'm sure you could improve on them and if you'd like to do so, please feel free to amend the PR directly.

### DIFF
--- a/pkg/webhook/kubernetes/replicatedresource/webhook_test.go
+++ b/pkg/webhook/kubernetes/replicatedresource/webhook_test.go
@@ -1,7 +1,6 @@
 package replicatedresource
 
 import (
-	"context"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -45,7 +44,6 @@ func TestHandle(t *testing.T) {
 				},
 			},
 			assert: func(t *testing.T, resp admission.Response) {
-				t.Helper()
 				require.True(t, resp.Allowed)
 			},
 		},
@@ -60,7 +58,6 @@ func TestHandle(t *testing.T) {
 				},
 			},
 			assert: func(t *testing.T, resp admission.Response) {
-				t.Helper()
 				require.True(t, resp.Allowed)
 			},
 		},
@@ -75,7 +72,6 @@ func TestHandle(t *testing.T) {
 				},
 			},
 			assert: func(t *testing.T, resp admission.Response) {
-				t.Helper()
 				require.True(t, resp.Allowed)
 			},
 		},
@@ -90,7 +86,6 @@ func TestHandle(t *testing.T) {
 				},
 			},
 			assert: func(t *testing.T, resp admission.Response) {
-				t.Helper()
 				require.False(t, resp.Allowed)
 			},
 		},
@@ -105,7 +100,6 @@ func TestHandle(t *testing.T) {
 				},
 			},
 			assert: func(t *testing.T, resp admission.Response) {
-				t.Helper()
 				require.False(t, resp.Allowed)
 			},
 		},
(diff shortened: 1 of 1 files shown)


pkg/webhook/kubernetes/replicatedresource/webhook_test.go
```diff
@@ -0,0 +1,151 @@
...
+				t.Helper()
+				require.False(t, resp.Allowed)
+			},
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+			w := newWebhook(cfg)
+			resp := w.Handle(context.Background(), testCase.req)
```
fuskovic: NIT:

```suggestion
			resp := w.Handle(t.Context(), testCase.req)
```

pkg/webhook/kubernetes/replicatedresource/webhook_test.go
```diff
@@ -0,0 +1,151 @@
...
+					Operation: admissionv1.Create,
+					UserInfo: authnv1.UserInfo{
+						Username: testManagementControllerUsername,
+					},
+				},
+			},
+			assert: func(t *testing.T, resp admission.Response) {
+				t.Helper()
+				require.True(t, resp.Allowed)
+			},
```
fuskovic, line 48: NIT: The use of `t.Helper()` here ( and other test cases )looks odd. I've historically found these used for `setupTestX` funcs or some test utilities I'm trying to keep DRY. I don't think it's meaningful in the context of one line assertions.

pkg/webhook/kubernetes/replicatedresource/webhook_test.go
krancour: Nice catch.

pkg/webhook/kubernetes/replicatedresource/webhook_test.go
krancour: Agree that was a misuse that doesn't belong.

krancour: Fixed in ~~c8f5449011cf30de1a2130dd78a9589ceab37430~~ dae2d3653ac288cf0a356125ef19561baa211894

pkg/webhook/kubernetes/replicatedresource/webhook_test.go
krancour: Fixed in ~~c8f5449011cf30de1a2130dd78a9589ceab37430~~ dae2d3653ac288cf0a356125ef19561baa211894

krancour:
@fuskovic @rpelczar, thanks, both!

Issue #5636, opened 2026-01-27
Reduce log level for "Stage health evaluated to Unknown" during active promotions


tal-hason:
## Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`.
* [x] I've pasted logs, if applicable.

## Description

During active promotions, the Stage controller logs ERROR level messages for "Stage health evaluated to Unknown" even though this is expected transient behavior during normal operations.

PR #4871 (v1.7.3) refined **when** Unknown health is treated as an error, but the log level remains ERROR for the "promotion in progress" case. When `!lastPromo.Status.Phase.IsTerminal()` evaluates true (i.e., promotion is Running), the controller returns an error at line 479 of `pkg/controller/stages/regular_stages.go`, which controller-runtime logs at ERROR level.

**Impact:**
- Log noise during normal promotion operations
- Difficult to distinguish actual errors from expected transitional states
- Alert fatigue if monitoring ERROR level logs

**Related:**
- Discussion #5024
- PR #4871 (fix that refined when to error, but not the log level)
- Issue #5045 (closed, but log level concern not fully addressed)

## Screenshots

N/A

## Steps to Reproduce

1. Create a Stage with health checks configured
2. Trigger a promotion to that Stage
3. Observe kargo-controller logs during the promotion's Running phase
4. ERROR logs appear repeatedly until the promotion completes and health becomes Healthy

## Version

```
version: v1.8.7
```

## Logs

```
2026-01-27T09:22:16Z ERROR controller/controller.go:421 Reconciler error {"controller": "stage", "controllerGroup": "kargo.akuity.io", "controllerKind": "Stage", "Stage": {"name":"candidate-errata","namespace":"v4-17"}, "namespace": "v4-17", "name": "candidate-errata", "reconcileID": "7364c7fa-7f02-4828-80e0-076f7ce57df4", "error": "Stage health evaluated to Unknown"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:421
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1
sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:296
```

## Suggested Fix

[...]

PR #5984, opened 2026-03-26
feat: make work queue backoff configurable via env vars

### LINKED ISSUE #5347: Freigh promotion stalls in Unknown health state for 3 minutes

alexmt:
The stage reconciler returns an error for "Stage health evaluated to Unknown" during active promotions. This is fine in theory, but the default controller-runtime backoff caps at 1000s — so a burst of ~20 quick
   errors can lock out reconciliation for 15+ minutes. In practice this means verifications never show up because the stage never transitions to Healthy within a reasonable window.

  This exposes the backoff base/max as env vars (RECONCILER_BACKOFF_BASE_DELAY, RECONCILER_BACKOFF_MAX_DELAY) so operators can tune it without a code change. Defaults are unchanged from controller-runtime's
  built-ins (5ms / 1000s).

Fixes #5347

Related to #5636

### DIFF
--- a/pkg/controller/options.go
+++ b/pkg/controller/options.go
@@ -1,13 +1,45 @@
 package controller
 
 import (
+	"time"
+
+	"github.com/kelseyhightower/envconfig"
+	"golang.org/x/time/rate"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+// rateLimiterConfig holds work queue rate limiter settings. These mirror the
+// controller-runtime defaults but can be overridden via environment variables
+// to tune reconciler backoff without a code change.
+type rateLimiterConfig struct {
+	// BackoffBaseDelay is the initial delay for the per-item exponential
+	// failure rate limiter.
+	BackoffBaseDelay time.Duration `envconfig:"RECONCILER_BACKOFF_BASE_DELAY" default:"5ms"`
+	// BackoffMaxDelay caps the per-item exponential failure rate limiter.
+	// The controller-runtime default is 1000s; a burst of reconciliation
+	// errors (e.g. "Stage health evaluated to Unknown" during an active
+	// promotion) can drive the backoff up to that cap and stall the stage
+	// for 15+ minutes. Lower this value to recover faster.
+	BackoffMaxDelay time.Duration `envconfig:"RECONCILER_BACKOFF_MAX_DELAY" default:"1000s"`
+}
+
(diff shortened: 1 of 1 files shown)


krancour:
@alexmt good understanding of what's happening here. If it backs off too far, too fast, things get bad. ☹️ (Reliance on an error to produce the backoff effect also clutters the logs, which I hate.)

Making backoff limits tunable would be one solution -- and it's a much less heavy-handed solution than the often-proposed one of making the requeue interval configurable. 👏 (Most people would get themselves into a lot of trouble by reconciling too frequently under normal conditions.)

My only concern is that what's being tuned here is being applied more widely than may be necessary. The underlying problem of an App's health status not being trustworthy for some period of time after the completion of the last operation is a fairly unique one. I feel as if there must be a more targeted solution that we've not identified yet.

This PR looks good to me, but if it's ok, I think we treat this as "the one to beat." We keep trying for a better solution and if it doesn't materialize on time for 1.10, then we go with this. Reasonable?

alexmt:
Sure, it makes sense to me. Thank you

pkg/controller/options.go
```diff
@@ -1,13 +1,45 @@
...
 	return controller.Options{
 		MaxConcurrentReconciles: maxConcurrentReconciles,
 		RecoverPanic:            ptr.To(true),
+		RateLimiter: workqueue.NewTypedMaxOfRateLimiter(
+			workqueue.NewTypedItemExponentialFailureRateLimiter[reconcile.Request](
+				cfg.BackoffBaseDelay,
+				cfg.BackoffMaxDelay,
+			),
+			&workqueue.TypedBucketRateLimiter[reconcile.Request]{
+				Limiter: rate.NewLimiter(rate.Limit(10), 100),
```
EronWright, line 41: Curious, what are these constants?

krancour:
Closing in favor of #5995, which testing shows almost instantly resolves the issue of untrusted App health status. If we discover further issues, we can re-consider this approach.

PR #5987, opened 2026-03-26
feat(warehouse): allow chart subscriptions to skip TLS verification

### LINKED ISSUE #5976: Supporting insecureSkipTLSVerify for chart subscriptions
# Checklist

# Proposed Feature

Hi, we're using Kargo 1.9.2 and it seems that it's not possible to skip the TLS checks when using a `chart` subscription based on OCI:

```yaml
kind: Warehouse
metadata:
  name: acip
  namespace: kargo-test
spec:
  subscriptions:
    - chart:
        repoURL: oci://docker.test.com/test/test
        semverConstraint: ">=0.1.0"
        discoveryLimit: 5
        insecureSkipTLSVerify: true # Not supported today
```

# Motivation

We're using an internal OCI registry with a custom CA. At least in the testing phase, we'd like not to have to setup the CA at Kargo level and we'd like to be able to carry on by just not checking the TLS.

EvannDev:
Fixes #5976

 Git and image subscriptions already support insecureSkipTLSVerify for connecting to repositories with self-signed or custom CA certificates. This PR brings the same capability to chart subscriptions, for both HTTP/S and OCI repositories.

 A new insecureSkipTLSVerify field is added to ChartSubscription. When set, TLS certificate verification errors are ignored. This mirrors the existing behavior for other subscription types and should only be used with care.

 What changed:
 - api/v1alpha1: new InsecureSkipTLSVerify bool field on ChartSubscription; protobuf bindings and TypeScript client regenerated
 - pkg/subscription/schemas/chart.json: schema updated accordingly
 - pkg/helm/chart: HTTP/S selector creates a scoped *http.Client with TLS verification disabled when the flag is set (avoiding mutation of http.DefaultClient); OCI selector propagates the flag to the authorizer's transport
 - pkg/helm/registry.go: NewEphemeralAuthorizer now accepts an insecure bool parameter

 Tests cover both the HTTP/S and OCI paths, including a httptest.NewTLSServer-based test that verifies the flag actually controls whether TLS errors are ignored.

### DIFF
--- a/pkg/helm/chart/http_selector_test.go
+++ b/pkg/helm/chart/http_selector_test.go
@@ -117,7 +117,7 @@ func Test_httpSelector_Select_insecureTLS(t *testing.T) {
 				indexURL:  fmt.Sprintf("%s/index.yaml", repoURL),
 				chartName: "fake-chart",
 			}
-			versions, err := s.Select(context.Background())
+			versions, err := s.Select(t.Context())
 			testCase.assertions(t, versions, err)
 		})
 	}
--- a/pkg/helm/registry_test.go
+++ b/pkg/helm/registry_test.go
@@ -49,7 +49,7 @@ func TestNewRegistryClient(t *testing.T) {
 }
 
 func TestNewEphemeralAuthorizer(t *testing.T) {
-	t.Run("secure", func(t *testing.T) {
+	t.Run("skipping tls cert verification", func(t *testing.T) {
 		authorizer := NewEphemeralAuthorizer(false)
 
 		assert.NotNil(t, authorizer)


pkg/helm/chart/http_selector_test.go
```diff
@@ -69,6 +69,60 @@ func TestNewHTTPSelector(t *testing.T) {
...
+			repoURL := fmt.Sprintf("%s/fake-repo", tlsServer.URL)
+			s := &httpSelector{
+				baseSelector: &baseSelector{
+					repoURL:               repoURL,
+					insecureSkipTLSVerify: testCase.insecureSkipTLSVerify,
+				},
+				indexURL:  fmt.Sprintf("%s/index.yaml", repoURL),
+				chartName: "fake-chart",
+			}
+			versions, err := s.Select(context.Background())
```
krancour:
```suggestion
			versions, err := s.Select(t.Context())
```

krancour: We're trying to shift to this way, over time. Any new usage of context in a test should look like this. Someday we'll fix all the existing occurrences of `context.Background()` in tests.

pkg/helm/registry_test.go
```diff
@@ -48,16 +49,42 @@ func TestNewRegistryClient(t *testing.T) {
...
+		assert.NotNil(t, authorizer.Client.Client)
+		assert.NotNil(t, authorizer.Cache)
+		assert.NotNil(t, authorizer.Credential)
+
+		// Verify the user agent is set correctly
+		expectedUserAgent := "Kargo/" + version.GetVersion().Version
+		assert.Contains(t, authorizer.Header.Get("User-Agent"), expectedUserAgent)
+	})
+
+	t.Run("insecure", func(t *testing.T) {
```
krancour: Just because "secure" and "insecure" can be so broadly interpreted, I'd rather we were more specific here:

```suggestion
	t.Run("skipping tls cert verification", func(t *testing.T) {
		authorizer := NewEphemeralAuthorizer(false)

		assert.NotNil(t, authorizer)
		assert.NotNil(t, authorizer.Client)
		assert.NotNil(t, authorizer.Store)
		assert.NotNil(t, authorizer.Client.Client)
		assert.NotNil(t, authorizer.Cache)
		assert.NotNil(t, authorizer.Credential)

		// Verify the user agent is set correctly
		expectedUserAgent := "Kargo/" + version.GetVersion().Version
		assert.Contains(t, authorizer.Header.Get("User-Agent"), expectedUserAgent)
	})

	t.Run("insecure", func(t *testing.T) {
```

krancour (approved):
🔥 Really nicely done @EvannDev!

Issue #5077, opened 2025-09-19
add "since" option (or similar tbd feature) to git subscriptions


krancour:
### Discussed in https://github.com/akuity/kargo/discussions/5074

<div type='discussions-op-text'>

<sup>Originally posted by **shamsalmon** September 18, 2025</sup>
Git is extremely slow pulling from monorepos with large commit histories. I have determined the primary reason for this not directly repo size, but rather how many git log operations are being run and the history of the repo itself.

The main issue is with the logic defined here:
https://github.com/akuity/kargo/blob/main/internal/controller/git/work_tree.go#L395

There is no way to set filters *within* this work tree function. My proposal is to add several options here:
1. Set the git log path to only look at folders within the IncludePath to reduce history
3. Ability to set a --since argument ("1 year ago" for ex) to reduce history

There are more potential improvements here, such as also making ExcludePaths transition into the pathspec format for git log. </div>

---

More context. We've nailed down the root cause and it's this:

https://github.com/akuity/kargo/discussions/5074#discussioncomment-14457660

This is the _tentative_ solution:

> Your --since idea might be the most viable thing here, as when you are creating a subscription involving new apps or environments in your monorepo, it should probably be relatively easily to say, "hey, this stuff didn't exist before this date, so if you go back that far and haven't hit the discovery limit, you're done."

I'd like @hiddeco and myself to discuss that further before we commit to that approach, but I'm tentatively including this in v1.9.0 because the ratio of pain relief to effort is very high, so I'd like to see us address it soon-ish.

shamsalmon:
I added this to a fork we run where I modified the `selectCommits` function to do this:

```
			if len(selectedCommits) == 0 && commit.CommitDate.Before(n.sinceDate) {
				return nil, fmt.Errorf("no commits found in branch before %s", n.sinceDate.Format("2006-01-02"))
			}

			if len(selectedCommits) >= 1 && commit.CommitDate.Before(n.sinceDate) {
				return trimSlice(selectedCommits, n.discoveryLimit), nil
			}
```

Would you like me to submit a PR with the crd / function changes or do you want to discuss internally more about the best way to approach this?

krancour:
@shamsalmon, if you want to give the cliff notes version of your changes by first showing tentative changes to API types, that will be a good springboard for further conversation. Just be aware that some patience will be in order. We're juggling a lot of high priority issues right now.

shamsalmon:
[...]

PR #5359, opened 2025-11-06
feat: add sincedate to skip long git lookbacks [DRAFT] 

shamsalmon:
This feature adds the ability to optionally limit how far the git newest from branch selector will go to find commits that match your specific discovery limit / included files.

Discussed in https://github.com/akuity/kargo/issues/5077

This PR is mainly to discuss API types / drafted implementation. Open to any requested changes.

PR #5992, opened 2026-03-27
feat: sinceDate discovery limiter

fuskovic:
Closes https://github.com/akuity/kargo/issues/5077

Builds largely on [the draft here](https://github.com/akuity/kargo/pull/5359) (Thanks @shamsalmon). However, since that PR is half a year old we decided it would be faster to just implement the same idea on an up-to-date clean branch than resolve merge conflicts. In addition to the approach outlined in the draft, this PR adds tests and docs.


pkg/subscription/schemas/git.json
```diff
@@ -86,6 +86,11 @@
             "maximum": 100,
             "default": 20,
             "description": "DiscoveryLimit is an optional limit on the number of commits that can be discovered for this subscription. The upper limit is 100."
+        },
+        "sinceDate": {
```
krancour: `sinceDate` seems somewhat redundant. There's an implied time element to the word "since." Can we just call it `since`?

api/v1alpha1/zz_subscription_types.go
krancour, line 1: @fuskovic weird thing going on here. There's two layers to the codegen on this file. quicktype re-creates it from scratch from the schema every time. All the protobuf struct tags get lost. When other parts of the codegen process run after, the protobuf struct tags get added back, but they're re-enumerated. Oridinarily, that's a no no, BUT...

They are dead weight here. Since 1.9, all subscriptions move over the wire as opaque chunks of JSON. So these types literally are _never_ serialized as protocol buffers.

To avoid future confusion, I altered the codegen process slightly so that types in this file don't get protobuf struct tags at all anymore.

Makefile
```diff
@@ -216,7 +216,7 @@ build-cli-with-ui: build-ui build-cli
 ################################################################################
 
 .PHONY: codegen
-codegen: codegen-openapi codegen-schema-to-go codegen-proto codegen-controller codegen-ui codegen-docs
+codegen: codegen-openapi codegen-proto codegen-controller codegen-schema-to-go codegen-ui codegen-docs
```
krancour, line 219: See https://github.com/akuity/kargo/pull/5992/changes#r3003755932 for explanation.

pkg/subscription/schemas/git.json
```diff
@@ -86,6 +86,11 @@
             "maximum": 100,
             "default": 20,
             "description": "DiscoveryLimit is an optional limit on the number of commits that can be discovered for this subscription. The upper limit is 100."
+        },
+        "since": {
+            "type": "string",
+            "format": "date-time",
+            "description": "An optional date (RFC 3339) that limits commit discovery to commits at or after this date. When specified, discovery stops upon reaching a commit older than this date. When left unspecified, there is no cutoff."
```
krancour, line 93: I made this modification to make it an RFC 3339 date -- which is how both Go time.Time and k8s metav1.Time are respresented as JSON.

Changing this means we get a real metav1.Time in the config struct.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -329,6 +329,19 @@ Git repository subscriptions can be defined using the following fields:
...
+  encounters a commit older than this date and returns any matching commits found
+  up to that point. When left unspecified, there is no date cutoff.
+
+  :::note
+
+  `sinceDate` only has effect when `commitSelectionStrategy` is
+  `NewestFromBranch` (or unspecified, since `NewestFromBranch` is the default).
+  It is particularly useful for large repositories with long histories where
+  `discoveryLimit` alone is not sufficient to prevent slow lookbacks.
+  :::
```
krancour: Needs to change to be RFC 3339.

hack/codegen/subscriptions.sh
```diff
@@ -31,7 +31,12 @@ sed -i.bak 's/Digest[[:space:]]*ImageSelectionStrategy/ImageSelectionStrategyDig
 sed -i.bak 's/NewestBuild[[:space:]]*ImageSelectionStrategy/ImageSelectionStrategyNewestBuild ImageSelectionStrategy/' ${out_file}
 sed -i.bak 's/CacheByTag[[:space:]]*\*bool/CacheByTag bool/' ${out_file}
 sed -i.bak 's/InsecureSkipTLSVerify[[:space:]]*\*bool/InsecureSkipTLSVerify bool/' ${out_file}
+sed -i.bak 's/time\.Time/metav1.Time/g' ${out_file}
```
krancour, line 34: See https://github.com/akuity/kargo/pull/5992/changes#r3003762518

Quicktype will use time.Time for RFC 3339, but we're changing it to be metav1.Time here. This config ultimately goes in a CRD, so this small change makes it look and feel like k8s.

hack/codegen/subscriptions.sh
```diff
@@ -31,7 +31,12 @@ sed -i.bak 's/Digest[[:space:]]*ImageSelectionStrategy/ImageSelectionStrategyDig
...
+# quicktype emits per-schema import statements that can end up in the middle of
+# the file when multiple schemas are combined. Remove them and let goimports
+# add a proper import block based on usage.
+sed -i.bak '/^import /d' ${out_file}
 
 rm ${out_file}.bak
 
-gofmt -w ${out_file}
+hack/bin/goimports -w ${out_file}
```
krancour, line 42: This is just a weird quicktype thing that's plain wrong. It seems to do a pretty dumb/straightforward concatenation of generated structs into one file. The new import for time ends up in the middle of the file. All tools, even gofmt and goimports choke on that. So we remove all imports and then add them back with goimports. nbd.

Makefile
```diff
@@ -260,7 +260,7 @@ codegen-controller: install-controller-gen
 		paths=./...
 
 .PHONY: codegen-schema-to-go
-codegen-schema-to-go:
+codegen-schema-to-go: install-goimports
```
krancour, line 263: Explanation for this here: https://github.com/akuity/kargo/pull/5992/changes#r3003773914

krancour:
@fuskovic @shamsalmon I amended the PR with a few tweaks to the schema and the codegen process. Explanations inline with the diffs.

I haven't modified any logic, docs, or tests. If I spot anything in those, I'll just comment like normal. i.e. I'm done putting my greasy fingers all over this and am turning it back over to you guys.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
krancour: Format looks like `2026-01-01T00:00:00Z`

pkg/controller/git/commit/newest_from_branch_selector.go
```diff
@@ -140,12 +145,24 @@ func (n *newestFromBranchSelector) selectCommits(
...
+			if n.sinceDate != nil && commit.CommitDate.Before(*n.sinceDate) {
+				if len(selectedCommits) == 0 {
+					return nil, fmt.Errorf(
+						"no commits found in branch since %s",
+						n.sinceDate.Format("2006-01-02"),
+					)
+				}
+				return trimSlice(selectedCommits, n.discoveryLimit), nil
+			}
+
```
krancour: I think it would be much more efficient to leave this loop untouched and handle this at a lower level. Could we make `ListCommits()` in the `WorkTree` interface start accepting a new struct `ListCommitsOptions`? A "since" date (`*time.Time`) can be one of the options. `git log` has a `--since` flag that accepts RFC 3339, which is exactly what the default string representation of a `time.Time` is.

To be clear, the loop you have here _works_, but the premise of the whole feature is "I have a large monorepo and I'm only interested in paths that were added recently and I don't want to search all the way back to the initial commit looking to exhaust the discovery limit when I know I won't." For this scenario, getting git to do most of the work for us will be much faster.

krancour: This selector won't need to do anything more than pass the option on.

pkg/controller/git/work_tree.go
```diff
@@ -514,14 +524,19 @@ func (w *workTree) ListCommits(limit, skip uint) ([]CommitMetadata, error) {
...
+		if opts.Limit > 0 {
+			args = append(args, fmt.Sprintf("--max-count=%d", opts.Limit))
+		}
+		if opts.Skip > 0 {
+			args = append(args, fmt.Sprintf("--skip=%d", opts.Skip))
+		}
+		if opts.Since != nil {
+			args = append(args, fmt.Sprintf("--since=%s", opts.Since.Format(time.RFC3339)))
+		}
 	}
```
krancour: This 100% works, but I want to suggest that a way of dealing with opts that is more consistent with other spots throughout the code base that have the same concern. Put this at the top of the method:

```
if opts == nil {
	opts = &ListCommitsOptions
}
```

This removes any further worry about nil checks in the body of the method and if this code ever evolves to pass these options along to lower-level methods or functions, those lower level ones can accept `ListCommitsOptions` (instead of `*ListCommitsOptions`) and not need to worry about nil checks either.

docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
```diff
@@ -329,6 +329,19 @@ Git repository subscriptions can be defined using the following fields:
...
+  commits at or after this date are considered. When left unspecified, there is
+  no date cutoff.
+
+  :::note
+
+  `since` only has effect when `commitSelectionStrategy` is
+  `NewestFromBranch` (or unspecified, since `NewestFromBranch` is the default).
+  It is particularly useful for large repositories with long histories where
+  `discoveryLimit` alone is not sufficient to prevent slow lookbacks.
+  :::
```
krancour, line 344: I can't remember where warnings about this surface, but Docusaurus prefers an extra line break here.

```suggestion

  :::
```

pkg/controller/git/work_tree.go
krancour: LGTM after this.

krancour:
This got bounced back from the merge queue because of a conflict with #5999. I merged main into this branch and fixed the conflict.

PR #5996, opened 2026-03-27
feat: add toml related promotion steps

### LINKED ISSUE #5993: `toml-update` promotion step
# Checklist

# Proposed Feature

I would love to see the `toml-update` and, optionally, `toml-parse` promotion steps.

# Motivation

We use KCL to generate kubernetes manifests, and it would be nice if kargo could update `[dependencies]` section of `kcl.mod` file, which uses toml syntax.

nikita-unity:
Closes #5993.

(!) This code was created by AI with human supervision.

Added TOML promotion step support to match the existing JSON and YAML patterns. This introduces new `toml-parse` and `toml-update` runners, plus a TOML helper that updates existing scalar values in place while preserving untouched bytes in the source file.

The change also adds schema and UI support for both directives, generated config type updates, focused unit tests for parsing and updating behavior, and user-facing documentation for the new steps.

### DIFF
--- a/pkg/promotion/runner/builtin/toml_parser.go
+++ b/pkg/promotion/runner/builtin/toml_parser.go
@@ -113,7 +113,7 @@ func (tp *tomlParser) readAndParseTOML(workDir string, path string) (any, error)
 // []builtin.TOMLParse.
 func (tp *tomlParser) extractValues(
 	data any,
-	outputs []builtin.TOMLParse,
+	outputs []builtin.TomlParse,
 ) (map[string]any, error) {
 	results := make(map[string]any, len(outputs))
 	for _, output := range outputs {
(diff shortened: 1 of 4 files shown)


krancour:
Hey @nikita-unity, it's been a minute. These would be welcome additions and they're pretty-well implemented. Thank you.

I did layer a few changes on top of what you've got here. I didn't change anything too significantly.

* Docs: Made them more consistent with their JSON and YAML counterparts. This actually involved more edits to the existing docs than to the new ones. Also designated the new steps as beta.

* Removed your `splitKeyPath()` function in favor of reusing the one that the YAML `pkg` was already using. It is a little more robust and adds perfect parity with the key path syntax native to the sjson lib used by the json-update step. i.e. All three update steps use identical syntax now. It's not a big change. It's just a bit more thorough with escaping of special charcter.

* Fixed some small bugs. Nothing major.

* Added more test cases. I felt this was necessary given the use of the unstable API. The more tests the better, to minimize the possibility of unanticipated breaking changes in the unstable API breaking us. That's also why I designated these as beta.

That's all just cleanup. Many thanks to you and your agent for doing most of the leg work.

LGTM.

krancour:
Whoops. Forgot to run codegen.

PR #6013, opened 2026-03-29
feat: add kcl-run promotion step

### LINKED ISSUE #6012: add `kcl-run` promotion step
# Checklist

# Proposed Feature

`kcl-run` should follow the `kustomize-build` and `helm-template` model, rendering yaml manifests from KCL source

# Motivation

It could be useful if you have KCL instead of helm or kustomize

nikita-unity:
Fixes #6012 

🤖 The code was generated by AI with human supervision

Adds a new builtin promotion step, `kcl-run`, for rendering manifests from KCL sources in promotion workflows.

The step follows the same general model as `kustomize-build` and `helm-template`.
 It supports rendering from a single KCL file, a package directory, or a `kcl.yaml` settings file, and can write output either to a single file or to a directory of split manifests.

It also accepts top-level KCL arguments via an arguments list, mirroring `kcl run -D name=value`


krancour:
Sorry @nikita-unity I may have set a bad precedent yesterday when I was quick to clean up and merge the toml steps your contributed (they were obvious low-hanging fruit and arguably stakes), but ordinarily, maintainers would like some opportunity to discuss anything that's been proposed before we become willing to review or accept contributions. @jessesuen and I are going to have to talk about whether KCL is something we want to support. I it may be a while before we get to this. Stay tuned.

nikita-unity:
I think KCL is not mature enough, so let's postpone this integration

PR #6034, opened 2026-04-02
feat(projects): support contributor registry for default project role rules

EronWright:
## Summary

- Adds `GetAll()` to the `PredicateBasedRegistry` interface (and `listBasedRegistry` implementation) in `pkg/component`, enabling "apply all matching" semantics in addition to the existing "find first" semantics.
- Introduces a `RoleRulesContributorRegistration` registry in `pkg/controller/management/projects` that allows callers to inject additional `PolicyRule`s into the default project roles (`kargo-admin`, `kargo-viewer`, `kargo-promoter`) at startup time.
- The `ensureDefaultUserRoles` function applies all registered contributors when creating default roles for a new project.

Companion EE PR: akuityio/kargo-enterprise#628

## Test plan

- [x] Unit tests for `GetAll()` in `pkg/component/list_based_registry_test.go`
- [x] Unit tests for contributor mechanism in `pkg/controller/management/projects/projects_test.go`

### DIFF
--- a/pkg/controller/management/projects/role_rules_contributor.go
+++ b/pkg/controller/management/projects/role_rules_contributor.go
@@ -1,0 +1,42 @@
+package projects
+
+import (
+	"context"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+
+	"github.com/akuity/kargo/pkg/component"
+)
+
+type (
+	// roleRulesContributorPredicate returns true if the contributor has
+	// PolicyRules to contribute for the given role name.
+	roleRulesContributorPredicate = func(context.Context, string) (bool, error)
+
+	// roleRulesContributorFunc returns additional PolicyRules for a given role
+	// name.
+	roleRulesContributorFunc = func(roleName string) []rbacv1.PolicyRule
+
+	// RoleRulesContributorRegistration associates a predicate with a
+	// contributor function.
+	RoleRulesContributorRegistration = component.PredicateBasedRegistration[
+		string,
+		roleRulesContributorPredicate,
+		roleRulesContributorFunc,
+		struct{},
+	]
+)
+
+var defaultRoleRulesContributorRegistry = component.MustNewPredicateBasedRegistry[
+	string,
+	roleRulesContributorPredicate,
+	roleRulesContributorFunc,
+	struct{},
+]()
+
+// RegisterRoleRulesContributor adds a contributor to the global registry used
+// by the project reconciler when creating default project roles. It should be
+// called before SetupReconcilerWithManager (e.g. at program startup).
+func RegisterRoleRulesContributor(reg RoleRulesContributorRegistration) {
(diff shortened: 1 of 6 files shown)


fuskovic (approved):
I think `GetMatches` is a more suitable name than `GetAll` but not a hill I wanna die on.

krancour:
@EronWright can this replace `ensureExtendedPermissions()`?

EronWright:
Yes — confirmed effective, and I'd like to handle it in a follow-up PR to keep the blast radius small here.

The follow-up would introduce two additional registries modeled on the same `PredicateBasedRegistry` + `GetAll()` pattern, with `*kargoapi.Project` as the predicate arg (so a contributor can opt out for certain projects if needed):

```go
ProjectSetupContributorRegistration  // called during reconciliation
ProjectCleanupContributorRegistration // called on deletion
```

`ensureExtendedPermissions` moves to the EE as a `ProjectSetupContributorRegistration`, and the Argo CD CRB deletion becomes a `ProjectCleanupContributorRegistration`. All the EE-specific fields currently in `ReconcilerConfig` go away from OSS.

One ordering note worth capturing: setup contributors run after the OSS base resources are created (OSS → EE), while cleanup contributors run *before* the OSS base resources are deleted (EE → OSS), so that EE resources referencing OSS ClusterRoles are removed first. That symmetry wasn't present in the original `cleanupProject` (which used a map, giving non-deterministic order).

pkg/controller/management/projects/role_rules_contributor.go
```diff
@@ -0,0 +1,42 @@
...
+	// contributor function.
+	RoleRulesContributorRegistration = component.PredicateBasedRegistration[
+		string,
+		roleRulesContributorPredicate,
+		roleRulesContributorFunc,
+		struct{},
+	]
+)
+
+var defaultRoleRulesContributorRegistry = component.MustNewPredicateBasedRegistry[
```
hairyhum, line 30: If that's a new registry, why use list with predicates and function with lookup by name instead of map registry? 
Am I missing something?

EronWright: The main thing is that there's numerous registered contributors for a given key, and you want all of the contributors.

hairyhum: > The main thing is that there's numerous registered contributors for a given key, and you want all of the contributors.

I don't see any examples of that. Is that an attempt to futureproof it? Maybe we can come up with a better API?

EronWright: I would say it is an extension point that conceptually should allow for numerous contributors to, in this case, a given project role's rules.

As for the API, let me mention that `mapBasedRegistry` can't store multiple independent contributors registering for the same key, whereas `listBasedRegistry` can and was the basis here.

PR #6049, opened 2026-04-08
fix(rbac): add missing release namespace to Role and RoleBinding resources

sjoukedv:
When rendering the chart with kustomize I noticed the following namespaced-scoped resources are missing the `namespace` on their spec.

### DIFF
--- a/charts/kargo/templates/api/role-bindings.yaml
+++ b/charts/kargo/templates/api/role-bindings.yaml
@@ -3,6 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: kargo-api
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "kargo.labels" . | nindent 4 }}
 roleRef:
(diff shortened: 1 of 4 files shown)


krancour:
Strange... It is not typical to explicitly specify the release namespace as a resource's namespace in a manifest template. It's usually a given. Oddly, we _have been_ explicit about it nearly everywhere, and you've correctly identified these as outliers. Before merging this, I want to do a little research and refresh my own memory as to why we were ever explicit about this in most places to begin with.

sjoukedv:
Argo CD was complaining about this when using kustomize with `--enable-helm`. With plain helm it worked without issues guess it might override the namespace if not specified with the installed namespace. The workaround we have for kustomize is to manually patch the resources that are included in this PR, but would be great to solve this upstream.

krancour:
I haven't had time to dig into why we were doing this to begin with, but regardless, I agree this brings these few templates up to date with the established pattern, so if for no other reason than consistency, this lgtm.

krancour:
This needs a rebase to pick up the changes from #6080 or it won't pass CI.

sjoukedv:
> This needs a rebase to pick up the changes from #6080 or it won't pass CI.

Done @krancour

krancour:
Thank you @sjoukedv!

krancour (approved):

PR #6057, opened 2026-04-09
feat: secret/configmap replication feature 

### LINKED ISSUE #6053: UI support for secret/configmap replication feature
# Proposed Feature

When managing a shared secret or configmap in UI:

Image

We should have a `Replicate` checkbox which when checked, will set the `kargo.akuity.io/replicate-to: "*"` annotation on the secret/configmap. 

A tooltip can be: "Replicate the resource to all projects to be use by AnalysisTemplates"

# Motivation

The feature can currently be enabled only via GitOps or CLI, but not the UI

rpelczar:
image

fixes #6053


Marvin9 (approved):

Marvin9 (commented):
Need backend review

pkg/server/update_config_map_v1alpha1.go
```diff
@@ -30,6 +30,7 @@ func (s *server) UpdateConfigMap(
 		name:        req.Msg.Name,
 		data:        req.Msg.Data,
 		description: req.Msg.Description,
+		replicate:   req.Msg.Replicate,
 	})
 
 	if err := s.client.Update(ctx, configMap); err != nil {
```
fuskovic, line 36: I know this isn't your doing but I think there's something important wrong here.

Instead of initializing a `configMap` object then calling `Update`, we need to be retrieving the existing object first. This is to:

1. Validate the resource we are trying to update exists.
2. Make sure the object we pass to `Update` has a valid `ResourceVersion`. This otherwise doesn't even work ( even though the tests pass because the fake-k8s-client doesn't check for this constraint )

You can look at the pattern that exists for `updateSystemConfigMap`, `updateSharedConfigMap`, and `updateProjectConfigMap`.

cc: @krancour

fuskovic (changes requested):

pkg/server/update_config_map_v1alpha1.go
Marvin9: Actually I realised we recently migrated to new endpoints.. and UI is using new instead of this one..

fuskovic (approved):
@Marvin9 ok cool. I was wondering what the purpose of this endpoint was when we have projectX, sharedX, systemX. Feels like it needs to be removed but can do in a separate PR if thats the case.

cc: @krancour

Marvin9:
> @Marvin9 ok cool. I was wondering what the purpose of this endpoint was when we have projectX, sharedX, systemX. Feels like it needs to be removed but can do in a separate PR if thats the case.
> 
> cc: @krancour

You are right. There are more unused endpoints like this now but that will part of complete removal of connectrpc

Marvin9:
Does it need backport to have in 1.10?

krancour:
@Marvin9 yep. I added it.

PR #6082, opened 2026-04-13
feat: pod disruption budget

### LINKED ISSUE #5396: PodDisruptionBudget and TopologySpreadContraints for better HA configuration

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5396

Cherry-picked to preserve original authorship of @bafulton  from https://github.com/akuity/kargo/pull/5638

Additionally, I added the change described here: https://github.com/akuity/kargo/pull/5638#discussion_r3016706307


charts/kargo/values.yaml
```diff
@@ -430,11 +430,21 @@ api:
   tolerations: []
   ## @param api.affinity Specifies pod affinity for api pods. Defaults to `global.affinity`.
   affinity: {}
+  ## @param api.topologySpreadConstraints Topology spread constraints for api pods.
+  topologySpreadConstraints: []
```
krancour, line 434: Non-blocking nit: I would include a commented example here (and in other similar spots). We don't need to annotate the example field by field or anything (our intent isn't too teach people about topology spread constraints), but just having the commented example will make it pretty clean that you just define constraints as usual. Without such a comment, it's not entirely clear what I'd be expected to put here. Again... non-blocking.

krancour (dismissed):

krancour:
Still LGTM

krancour (approved):
Nevermind... false alarm. Still LGTM.

PR #6083, opened 2026-04-13
fix(controller): Fix flapping owner reference write

thomastaylor312:
As part of the work done in #5763 (and as referenced in #4627), we said we weren't going to use owner references anymore. In fact, because we still have them, we have some thrashing going on because the project controller still writes a reference, and the new code in #5763 added code to the namespace controller that deletes it. This removes the code that adds an owner reference


EronWright (approved):
LGTM

fuskovic (approved):
Looks solid!

krancour (changes requested):
Refer to offline conversation. Just want to clear that up first.

krancour (approved):

PR #6099, opened 2026-04-16
fix: more thorough cleaning of .gpgnu state when loading an already-c…

krancour:
Follow-up to #6097.

## What #6097 left uncovered

#6097 cleaned committed dotlock files (`*.lock`) from a handful of enumerated subdirectories under `GNUPGHOME`. Testing on heavier workloads surfaced two additional failure modes that the original cleanup doesn't catch:

1. **Orphaned dotlock temp files (`.#lk*`).** GnuPG creates these via `O_CREAT|O_EXCL` before hard-linking them to the committed `.lock` file. Their names embed an in-process memory address + hostname + PID. Under weak ASLR and colliding container PIDs those names can repeat across processes, and a crashed writer's orphan causes the next `gpg` invocation to fail with `gpg: ... File exists` *before it ever reaches the `.lock` file*.

2. **Stale UNIX-domain sockets** left by dead `gpg-agent` / `dirmngr` / `scdaemon` processes. A new daemon binding on the same path can race with the stale socket.

Separately, the original cleanup enumerated specific subdirectories (`GNUPGHOME`, `public-keys.d/`, `private-keys-v1.d/`). That list is brittle — GnuPG's on-disk layout has changed before (the ongoing move toward `public-keys.d/pubring.db`, for instance) and may change again.

## Changes

- Extend `reconcileGPGState` to also remove `.#lk*` orphan temp files.
- Extend it to remove any file with `os.ModeSocket` set — covers `S.gpg-agent*`, `S.dirmngr`, `S.scdaemon`, and any future daemon without needing to know its name. The `gpgconf --kill gpg-agent` invocation still runs first, so anything remaining is by definition stale.
- Replace the enumerated-subdirectory glob loop with a single `filepath.WalkDir` over `GNUPGHOME`, matching files by predicate. Future layout changes are picked up automatically.

Single-file change to `pkg/controller/git/base_repo.go`. No call-site changes.

## Consequences

- Same as #6097: no-op in deployments that create fresh `homeDir`s per repo; fixes the newly-identified hazards in deployments that share a `GNUPGHOME` across process boundaries.
- Preserved files (`pubring.kbx`, `trustdb.gpg`, `random_seed`, `*.key`, `gpg.conf`, revocation certs, SQLite `-wal`/`-shm`/`-journal`, etc.) don't match any predicate and are never touched.
- Errors remain best-effort and logged.
- No public API or configuration changes.

### DIFF
--- a/pkg/controller/git/base_repo.go
+++ b/pkg/controller/git/base_repo.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):

release-1.10-examples-plan.md
krancour, line 1: did not mean to commit this

jessesuen (approved):

PR #6101, opened 2026-04-16
feat: Gitea pull request label support

### LINKED ISSUE #6021: feat: add gitea label support
The Gitea SDK's `AddIssueLabels `expects `labelIDs []int64`, but Kargo's `CreatePullRequestOpts.Labels` are `names []string`. A previous implementation attempted this but silently discarded the labels entirely.

See the TODO that was added in `pkg/gitprovider/gitea.go/CreatePullRequest` in https://github.com/akuity/kargo/pull/5581

# Proposed Feature

To support properly: list repo labels, match by name to get IDs, then call `AddIssueLabels`.

gnanirahulnutakki:
Fixes #6021

## Summary
- resolve configured Gitea label names to repository label IDs ( with pagination support ) before creating the PR
- add regression coverage for both the successful label pagination path and missing-label validation

## Testing
- `go test ./pkg/gitprovider/gitea`
- `./hack/bin/golangci-lint run --config .golangci.yaml ./pkg/gitprovider/gitea/...`

### DIFF
--- a/pkg/gitprovider/gitea/gitea.go
+++ b/pkg/gitprovider/gitea/gitea.go
@@ -169,7 +169,7 @@ func (p *provider) resolveLabelIDs(labelNames []string) ([]int64, error) {
 		return nil, nil
 	}
 
-	repoLabels := make([]*gitea.Label, 0, len(labelNames))
+	labelIDsByName := make(map[string]int64, len(labelNames))
 	for page := 1; ; {
 		pageLabels, resp, err := p.client.ListRepoLabels(
 			p.owner,
@@ -181,19 +181,16 @@ func (p *provider) resolveLabelIDs(labelNames []string) ([]int64, error) {
 		if err != nil {
 			return nil, fmt.Errorf("error listing repository labels: %w", err)
 		}
-		repoLabels = append(repoLabels, pageLabels...)
+		for _, repoLabel := range pageLabels {
+			if repoLabel == nil {
+				continue
+			}
+			labelIDsByName[repoLabel.Name] = repoLabel.ID
+		}
 		if resp == nil || resp.NextPage == 0 {
 			break
 		}
 		page = resp.NextPage
-	}
-
-	labelIDsByName := make(map[string]int64, len(repoLabels))
-	for _, repoLabel := range repoLabels {
-		if repoLabel == nil {
-			continue
-		}
-		labelIDsByName[repoLabel.Name] = repoLabel.ID
 	}
 
 	labelIDs := make([]int64, 0, len(labelNames))
--- a/pkg/gitprovider/gitea/gitea.go
+++ b/pkg/gitprovider/gitea/gitea.go
@@ -74,13 +74,6 @@ type giteaClient interface {
 		number int64,
 	) (*gitea.PullRequest, *gitea.Response, error)
 
-	AddIssueLabels(
-		owner string,
-		repo string,
-		number int64,
-		opts gitea.IssueLabelsOption,
-	) ([]*gitea.Label, *gitea.Response, error)
-
 	MergePullRequest(
 		owner string,
 		repo string,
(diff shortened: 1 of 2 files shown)


pkg/gitprovider/gitea/gitea.go
```diff
@@ -149,15 +166,73 @@ func (p *provider) CreatePullRequest(
...
+	}
+
+	repoLabels, _, err := p.client.ListRepoLabels(
+		p.owner,
+		p.repo,
+		gitea.ListLabelsOptions{},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error listing repository labels: %w", err)
+	}
```
fuskovic: `ListRepoLabels` is called with an empty `gitea.ListLabelsOptions{}`, which uses Gitea's default page size https://gitea.com/gitea/go-sdk/src/tag/gitea/v0.20.0/gitea/list_options.go#L19.

If a repository has more labels than the default, the function will not see them and will incorrectly report them as missing.

One solution for making sure we are listing all labels here could be incrementing `opts.Page` in a loop until the response is empty.

I'd like to see a test for this too.

pkg/gitprovider/gitea/gitea.go
```diff
@@ -149,15 +166,73 @@ func (p *provider) CreatePullRequest(
...
+			giteaPR.Index,
+			gitea.IssueLabelsOption{Labels: labelIDs},
+		); err != nil {
+			return nil, fmt.Errorf(
+				"error adding labels to pull request %d: %w",
+				giteaPR.Index,
+				err,
+			)
+		}
+	}
```
fuskovic: So if we get here it means we successfully created a PR but because we failed to add labels we report an error.

We could improve the error here e.g. `successfully created pr number %d but failed to add labels: %w`.

Although, I think a better solution would be to not have this be a separate call at all.

It looks like `CreatePullRequestOption` has a `Labels` field: https://gitea.com/gitea/go-sdk/src/tag/gitea/v0.20.0/gitea/pull.go#L159 and since you already resolve `labelIDs` before the `CreatePullRequest` call it makes more sense to just add your `labelIDs` there.

fuskovic (changes requested):

pkg/gitprovider/gitea/gitea.go
gnanirahulnutakki: Paged label lookup now walks `ListRepoLabels` using `Response.NextPage` until exhaustion instead of assuming the first page contains every repository label. I added regression coverage for a label that only appears on a later page so this does not silently regress.

pkg/gitprovider/gitea/gitea.go
gnanirahulnutakki: The separate `AddIssueLabels` call is gone. Resolved label IDs are now passed directly in `CreatePullRequestOption.Labels`, so label application stays part of PR creation instead of becoming a second failure point after the PR already exists. The updated test asserts that path, and I also validated it end-to-end against a live OrbStack-backed Gitea instance with a real paginated label set.

fuskovic (approved):

pkg/gitprovider/gitea/gitea.go
```diff
@@ -149,15 +160,69 @@ func (p *provider) CreatePullRequest(
...
+		page = resp.NextPage
+	}
+
+	labelIDsByName := make(map[string]int64, len(repoLabels))
+	for _, repoLabel := range repoLabels {
+		if repoLabel == nil {
+			continue
+		}
+		labelIDsByName[repoLabel.Name] = repoLabel.ID
+	}
```
fuskovic: This loop is unnecessary. You can build the map directly inside the pagination loop; eliminating the intermediate `repoLabels` slice entirely.

fuskovic (changes requested):

pkg/gitprovider/gitea/gitea.go
gnanirahulnutakki: Updated in `358cd408`.

The intermediate `repoLabels` slice is gone now. The label name -> ID map is built directly while paging `ListRepoLabels`, so the behavior stays the same for missing-label reporting and duplicate suppression without the extra accumulation pass.

Local recheck:

```bash
/opt/homebrew/bin/go test ./pkg/gitprovider/gitea
```

fuskovic (changes requested):
Tests are failing. 

Looks like you're using the wrong URL type.

PR #6102, opened 2026-04-16
feat(helm): add control options to non-main namespaces

snuggie12:
- Conditionally create RBAC (role/rolebinding) resources
- Ability to add annotations and labels to namespaces

Alternatively I could link the RBAC to the `Namespaces` being created, but felt this ultimately gave more flexibility.

Motivation
----------
The conditional RBAC is because we install the controller and the manager in separate argo applications so currently we need to create kustomize patches to delete these.

The annotations and labels is because we have kyverno policies that require certain ones to be there so this too will save on a lot of patches.


snuggie12:
Also, is there automatic doc generation or should I go update it?

krancour:
Thanks @snuggie12.

LGTM and I'm happy to merge this once conflicts are resolved and codegen is run (to update the chart's `README.md`).

krancour (approved):

charts/kargo/templates/shared-resources-namespace/namespace.yaml
```diff
@@ -2,8 +2,11 @@
...
 metadata:
-  name: {{ .Values.global.sharedResources.namespace }}
+  {{- include "kargo.annotations" (dict "root" . "annotations" .Values.global.sharedResources.extraNamespaceAnnotations) | nindent 2 }}
   labels:
     {{- include "kargo.labels" . | nindent 4 }}
-  {{- include "kargo.annotations" (dict "root" .) | nindent 2 }}
+  {{- with .Values.global.sharedResources.extraNamespaceLabels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  name: {{ .Values.global.sharedResources.namespace }}
```
krancour: This arrangement would be more consistent with what's done elsewhere, throughout the chart.

```suggestion
  name: {{ .Values.global.sharedResources.namespace }}
  labels:
    {{- include "kargo.labels" . | nindent 4 }}
  {{- with (mergeOverwrite (deepCopy .Values.global.labels) .Values.global.sharedResources.extraNamespaceLabels) }}
    {{- range $key, $value := . }}
    {{ $key }}: {{ $value | quote }}
    {{- end }}
  {{- end }}
  {{- include "kargo.annotations" (dict "root" . "annotations" .Values.global.sharedResources.extraNamespaceAnnotations) | nindent 2 }}
```

charts/kargo/templates/system-resources-namespace/namespace.yaml
```diff
@@ -2,8 +2,11 @@
...
 metadata:
-  name: {{ quote .Values.global.systemResources.namespace }}
+  {{- include "kargo.annotations" (dict "root" . "annotations" .Values.global.systemResources.extraNamespaceAnnotations) | nindent 2 }}
   labels:
     {{- include "kargo.labels" . | nindent 4 }}
-  {{- include "kargo.annotations" (dict "root" .) | nindent 2 }}
+  {{- with .Values.global.systemResources.extraNamespaceLabels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  name: {{ quote .Values.global.systemResources.namespace }}
```
krancour: Suggest doing it like this for consistency: https://github.com/akuity/kargo/pull/6102/changes#r3438568582

krancour:
It appears this still needs to happen:

> and codegen is run (to update the chart's README.md)

EronWright (changes requested):
A heads-up on process: the `kargo` chart recently gained a [`helm-unittest` suite](https://github.com/akuity/kargo/tree/main/charts/kargo/tests) (run via `make test-chart`), which now runs in CI. I took the liberty of pushing a commit to this branch that merges `main` in and **adds test coverage for the new behavior** — the `global.*.createRBAC` gating and `extraNamespaceLabels`/`extraNamespaceAnnotations` across the system, shared, and cluster-secrets namespaces. Hope that's helpful; feel free to adjust.

That coverage surfaced two things:

**1. Template typo (blocking — CI is red on this).**
In `charts/kargo/templates/system-resources-namespace/roles.yaml`, line 1 has a misplaced `}}`:

```gotemplate
{{- if and .Values.dataPlane.install }} .Values.global.systemResources.createRBAC }}
```

The other three RBAC files have the correct form — looks like just this one slipped through. The new `system-resources-namespace/roles.yaml` test suite fails on it, which is why CI is red.

**2. Possible gap — controller-reader RoleBindings ignore the new flag.**
The `*-controller-*-reader` RoleBinding in each `role-bindings.yaml` is gated only on `dataPlane.controller.install` + `controller.enabled`, so it still renders even when `createRBAC`/`createClusterSecretsRBAC` is `false`. If the intent is "this flag turns off RBAC creation in the namespace," you'd probably want that binding gated on the flag too. This is pre-existing structure, not something your PR introduced, so I've left it as-is and the tests document the current behavior — just flagging in case you'd like to fold it into the same flag here.

With the typo fixed, the suite goes green. Happy to help if anything in the test files is unclear.

krancour:
Thanks for being proactive about that @EronWright! Greatly appreciated!

EronWright:
Took the liberty of extending the PR a little — hope nobody minds!

- **Controller-reader RoleBindings now gated on `createRBAC`** (`system`/`shared`/`cluster-secrets`). Previously these rendered whenever the controller was installed, even with `createRBAC: false`, which produced a `RoleBinding` referencing a `Role` the chart never created. Now `createRBAC: false` consistently produces *no* RBAC in that namespace. Updated the test suite to match.
- **Regenerated `charts/kargo/README.md`** for the new `global.*.createRBAC` / `extraNamespaceLabels` / `extraNamespaceAnnotations` params (resolves `check-codegen`).

Happy to back any of this out if you'd rather keep the scope tighter.

fuskovic (approved):

charts/kargo/templates/shared-resources-namespace/namespace.yaml
EronWright: Done in 9d29d7d9 — switched to the standard `name` → `labels` → `mergeOverwrite(deepCopy global.labels, …)` → annotations idiom. Kept `name` quoted to match the sibling cluster-secrets namespace; happy to drop the quote if you prefer the unquoted form from your suggestion.

charts/kargo/templates/system-resources-namespace/namespace.yaml
EronWright: Done in 9d29d7d9 (same idiom as the shared-resources namespace).

EronWright (approved):

PR #6128, opened 2026-04-20
feat: BB  datacenter/server support

### LINKED ISSUE #6020: add support for opening, waiting on, and merging Bitbucket Data Center (nee Server) PRs
See original issue #2569 

#3029 closed it, but missed the mark, having added support only for Bitbucket's SaaS -- incredibly useful, but not what the issue was about.

See discussion [here](https://github.com/akuity/kargo/issues/2569#issuecomment-4170301594).

This is a second try at #2569, emphasizing the word "Server," although my current understanding is that Bitbucket Server was really just a narrower license, no longer offered, for the same software licensed more expansively as "Bitbucket Data Center."

Unfortunately, the API for Bitbucket Data Center is entirely different from the API for their SaaS, so this will not be a walk in the park.

Note that the external webhooks server is already supporting Bitbucket Data Center and this issue is focused entirely on support for opening, waiting on, and merging PRs on that platform.

Having recently discovered various "provider" (Git hosting provider) implementations, most of which were contributed by the community, to contain numerous as-yet-unreported bugs (which I am currently working to resolve, separately from this) _**we will not accept a community contribution for this issue. It needs to be addressed by a maintainer.**_

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6020


Makefile
```diff
@@ -216,7 +216,7 @@ build-cli-with-ui: build-ui build-cli
 ################################################################################
 
 .PHONY: codegen
-codegen: codegen-openapi codegen-proto codegen-controller codegen-schema-to-go codegen-ui codegen-docs
+codegen: codegen-openapi codegen-proto codegen-controller codegen-schema-to-go codegen-bitbucket-client codegen-ui codegen-docs
```
krancour, line 219: I wouldn't include this here since it's so rarely going to actually require re-generation.

Makefile
```diff
@@ -259,6 +259,10 @@ codegen-controller: install-controller-gen
 		object:headerFile=hack/boilerplate.go.txt \
 		paths=./...
 
+.PHONY: codegen-bitbucket-client
+codegen-bitbucket-client: install-oapi-codegen
```
krancour, line 263: We were already using go-swagger for generating the client for our own APIs. I'm open to using something else, such as this, but ideally we wouldn't use two different tools for the same purpose.

Did you happen to try either generating the BB client using go-swagger or, inversely, try generating our own client using oapi-codegen?

pkg/gitprovider/bitbucket/bitbucket_client.gen.go
krancour, line 1: Nit: If we have any control over the filename, prefixing this with `zz_` is a pretty widely recognized signal that something is generated. It's a nice-to-have and some of our existing tooling know to be on the lookout for that. The linter exempts such files and code coverage doesn't count lines from such files.

krancour:
I think there are some things to discuss here still, but I am really happy to see that generating a BB client and using that appears tenable.

Makefile
fuskovic: The two tools target different spec versions and are not interchangeable here.`go-swagger` only supports Swagger 2.0 (OAS2) specs. The Bitbucket Cloud API spec is OpenAPI 3.0 (OAS3),

fuskovic:
@krancour so far we are able to:

1. Remove `github.com/ktrysmt/go-bitbucket`
2. Support Datacenter/Server
3. Support alternative merge strategies for both

This obviously needs a lot of end-to-end testing.

Per our offline discussion I have sub-packages:
- `pkg/gitprovider/bitbucket/cloud`
- `pkg/gitprovider/bitbucket/datacenter`

Implemented each with their own individual registration predicate.

fuskovic:
I can take this out of draft and start end-to-end testing unless there's something else we still want to discuss?

fuskovic:
To unblock non-server/datacenter work, bb cloud client gen and alt merge strategy support have been moved into a separate PR [here](https://github.com/akuity/kargo/pull/6315).

Will leave this open for now in the event we are unblocked on testing the datacenter path in the future.

PR #6129, opened 2026-04-20
feat: retry mechanism for GAR WIF

### LINKED ISSUE #6127: GAR WIF provider silently disabled on init failure — no retry, no health signal
## Summary

When `NewWorkloadIdentityFederationProvider` fails during `init()` (e.g., GKE metadata server not ready at pod startup), the GAR credential provider is **silently and permanently disabled** for the lifetime of the process. There is no retry, no health check signal, and no warning when GAR/GCR URLs are later requested without a matching provider. All Warehouse image discovery against Artifact Registry fails with a cryptic "Unauthenticated request" error.

## How it happens

On GKE Autopilot, the Kargo controller pod can start before the metadata server is fully available. This is a known race condition — the GKE metadata server at `169.254.169.254` may not be ready immediately when a pod is scheduled, especially on newly provisioned nodes.

The `init()` function in `pkg/credentials/gar/workload_identity_federation.go` runs once during package import:

```go
func init() {
    if provider := NewWorkloadIdentityFederationProvider(context.Background()); provider \!= nil {
        credentials.DefaultProviderRegistry.MustRegister(...)
    }
}
```

`NewWorkloadIdentityFederationProvider` returns `nil` if any of these fail:
- `metadata.OnGCE()` returns false
- `metadata.ProjectIDWithContext(ctx)` errors
- `google.DefaultTokenSource(ctx, ...)` errors

When this happens:
1. The provider is never registered — no retry
2. `database.Get()` finds no matching provider for GAR URLs and returns `nil, nil` (not an error)
3. Kargo proceeds to call Artifact Registry without credentials
4. Every Warehouse polling GAR/GCR fails with: `"Unauthenticated request. Unauthenticated requests do not have permission..."`
5. This persists until the pod is manually restarted

## Impact

- **Cluster-wide** — all Warehouses watching GAR/GCR images are affected
- **Silent** — the only trace is a startup log line (`"GCP Workload Identity Federation disabled"`), which is easy to miss. The Kargo UI shows per-Warehouse "Unable to discover artifacts" errors with no indication that the root cause is a missing credential provider
- **Permanent** — requires a pod restart to recover. There is no self-healing

## Environment

- Kargo v1.9.6
- GKE Autopilot (australia-southeast1)
- Workload Identity Federation for GKE (non-legacy, `principal://` format)

## Workaround

1. Restart the controller pod: `kubectl rollout restart deployment/kargo-controller -n kargo-system`
2. Potentially an init container that checks and waits for the metadata server to respond?

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6127

On GKE Autopilot (and occasionally other GKE clusters), the instance metadata server may not be reachable when the Kargo controller starts. Previously, `init()` called the metadata server immediately — if that failed, the WIF credential provider was permanently unregistered for the lifetime of the process, causing all GAR/GCR requests to fail silently with no recovery path short of a restart.

NewWorkloadIdentityFederationProvider now accepts a `context.Context` and retries initialization up to 5 times with 1-second intervals. If all attempts fail the constructor returns `nil` and `init()` skips registration, so the provider is never in a partially-initialized state.


pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -21,90 +22,100 @@ import (
...
+		),
 	}
+	p.initFn = p.initialize
+	p.getAccessTokenFn = p.getAccessToken
+	credentials.DefaultProviderRegistry.MustRegister(
+		credentials.ProviderRegistration{
+			Predicate: p.Supports,
+			Value:     p,
+		},
+	)
```
krancour, line 50: I don't love this stuff having moved out of the constructor-like `NewWorkloadIdentityFederationProvider()` function. If a struct needs to be initialized in a particular non-trivial state in order to function properly, it's best to keep all of that in a constructor-like function so the burden of know exactly _how_ to initialize doesn't fall to the caller.

(I know `NewWorkloadIdentityFederationProvider()` wasn't actually called from anywhere else, but for all of these providers, the ability to instantiate them directly in a small, standalone test program has historically been useful in troubleshooting.)

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -21,90 +22,100 @@ import (
...
 }
 
 type WorkloadIdentityFederationProvider struct {
 	tokenCache       *cache.Cache // For short-lived Project-specific tokens
 	tokenSourceCache *cache.Cache // For long-lived token sources
 
-	projectID string
+	mu          sync.Mutex
```
krancour, line 57: Naming this something like `projectMu` would make its purpose more immediately understood. I wouldn't mind it being called just `mu` if it were super obvious what it was for, but with this code, I end up needing to read more to figure it out.

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -21,90 +22,100 @@ import (
...
+}
+
+// ensureInitialized calls initFn if projectID has not yet been populated.
+// It is safe to call from multiple goroutines.
+func (p *WorkloadIdentityFederationProvider) ensureInitialized(ctx context.Context) error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.projectID != "" {
+		return nil
 	}
```
krancour, line 97: It's admittedly a trivial difference, but this would perform better. If `p.projectID` is already observably not empty, there's no need to wait for a lock.

```suggestion
	if p.projectID != "" {
		return nil
	}
	p.mu.Lock()
	defer p.mu.Unlock()
```

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -114,6 +125,10 @@ func (p *WorkloadIdentityFederationProvider) GetCredentials(
 	ctx context.Context,
 	req credentials.Request,
 ) (*credentials.Credentials, error) {
+	if err := p.ensureInitialized(ctx); err != nil {
+		return nil, fmt.Errorf("GCP Workload Identity Federation not initialized: %w", err)
+	}
```
krancour, line 130: `Supports()` returns `false` and no error if initialization hasn't been ensured. Which means we should theoretically never get to here. I think it's important to call that out with a comment. Absent such a comment, I end up asking myself whether returning an error is the correct behavior here. Returning nil and no error means the caller moves on to the next provider. Returning an error aborts the credential lookup. I found myself a bit hung up on which of those two is the correct behavior here _until_ I realized this should never happen. So a comment would help avert similar confusion for others in the future.

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -21,90 +22,100 @@ import (
...
 	}
-	if !garURLRegex.MatchString(req.RepoURL) &&
-		!gcrURLRegex.MatchString(req.RepoURL) {
+	if err := p.ensureInitialized(ctx); err != nil {
+		logging.LoggerFromContext(ctx).Error(
+			err,
+			"GCP Workload Identity Federation not yet available; will retry on next request",
+		)
 		return false, nil
 	}
```
krancour, line 120: I'm iffy on logging an error here. On the one hand, there may be meaningful information. On the other, I'm thinking about how this plays out in a cluster where WIF simply isn't configured and no one has any expectation of this working. This would end up as a lot of noise in the logs for what would be an expected condition.

krancour: Perhaps there needs to be a cap on how many times this will be retried.

pkg/credentials/gar/workload_identity_federation.go
fuskovic: The suggested change is racy.

Test Claude wrote that flagged the race.

```go
func TestEnsureInitializedRace(t *testing.T) {
	ready := make(chan struct{})
	p := &WorkloadIdentityFederationProvider{
		tokenCache:       gocache.New(time.Minute, time.Minute),
		tokenSourceCache: gocache.New(time.Minute, time.Minute),
	}
	p.initFn = func(_ context.Context) error {
		<-ready               // block until we signal
		p.projectID = "test-project"
		return nil
	}

	var wg sync.WaitGroup

	// goroutine 1: acquires the lock, blocks inside initFn waiting on `ready`
	wg.Add(1)
	go func() {
		defer wg.Done()
		_ = p.ensureInitialized(t.Context())
	}()

	// goroutine 2: reads p.projectID in the outer check while
	// goroutine 1 is inside the lock about to write it
	wg.Add(1)
	go func() {
		defer wg.Done()
		time.Sleep(5 * time.Millisecond) // let goroutine 1 acquire the lock first
		close(ready)                     // unblock goroutine 1 — it will now write projectID
		_ = p.ensureInitialized(t.Context()) // outer read races the write
	}()

	wg.Wait()
}
```

Thoughts?

krancour: 😬 You're right and I apologize. I overlooked that the initialization was doing _more_ than just acquiring the project ID. It's also getting the token source. My suggestion was unsafe and should be disregarded.

That said, while re-examining this, I've come to question the whole approach.

According to the issue:

> On GKE Autopilot, the Kargo controller pod can start before the metadata server is fully available.

This is either a transient condition that will resolve quickly OR it will not resolve at all because WIF isn't configured. Overcoming it with a synchronous retry loop within the original `NewWorkloadIdentityFederationProvider()` function would be substantially simpler than lazy initialization and all the complications that come along with it:

```golang
var projectID string
var tokenSource oauth2.TokenSource

if err := retry.OnError(
	wait.Backoff{
		Steps:    5,
		Duration: 500 * time.Millisecond,
		Factor:   2.0,
		Jitter:   0.1,
		Cap:      4 * time.Second,
	},
	func(err error) bool {
		return true
	},
	func() error {
		var err error
		projectID, err = metadata.ProjectIDWithContext(ctx)
		if err != nil {
			return err
		}
		tokenSource, err = google.DefaultTokenSource(ctx, iamcredentials.CloudPlatformScope)
		return err
	},
); err != nil {
	logger.Error(err, "could not initialize Workload Identity Federation provider")
	return nil
}
logger.Debug("got GCP project ID", "project", projectID)
```

krancour: Even the progressive backoff is probably overkill. A simple for loop with a 1s wait within it would be fine.

__Edit:__ Actually, it's about the same number of lines to do it with a for loop. I would stand by the previous suggestion.

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -18,96 +18,127 @@ import (
...
-			credentials.ProviderRegistration{
-				Predicate: provider.Supports,
-				Value:     provider,
-			},
-		)
+	if !metadata.OnGCE() {
+		return
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
```
krancour: What's the rationale for this? `cancel()` has no potential to be called other than when `init()` returns and the 30s timeout is longer than the max time the function can run, so I'm not sure what value this adds.

krancour: Unless the thought is that `metadata.ProjectIDWithContext()` or `google.DefaultTokenSource()` could run long?

I could see that.

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -18,96 +18,127 @@ import (
...
-	if provider := NewWorkloadIdentityFederationProvider(context.Background()); provider != nil {
-		credentials.DefaultProviderRegistry.MustRegister(
-			credentials.ProviderRegistration{
-				Predicate: provider.Supports,
-				Value:     provider,
-			},
-		)
+	if !metadata.OnGCE() {
+		return
+	}
```
krancour: Nit: I would consider keeping this check where it was originally. Moving it here makes `NewWorkloadIdentityFederationProvider()` unsafe to call unless the caller knew to perform this check themselves.

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -18,96 +18,127 @@ import (
...
+	p := NewWorkloadIdentityFederationProvider(ctx)
+	if p == nil {
+		return
+	}
+	credentials.DefaultProviderRegistry.MustRegister(
+		credentials.ProviderRegistration{
+			Predicate: p.Supports,
+			Value:     p,
+		},
+	)
```
krancour:
```suggestion
	if p := NewWorkloadIdentityFederationProvider(ctx); p != nil {
		credentials.DefaultProviderRegistry.MustRegister(
			credentials.ProviderRegistration{
				Predicate: p.Supports,
				Value:     p,
			},
		)
	}
```

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -18,96 +18,127 @@ import (
...
+			time.Hour,      // Cleanup interval
+		),
+		tokenSourceCache: cache.New(
+			// Token sources are long-lived. We could hang on to them indefinitely,
+			// but we'll cap it at 12 hours to prevent memory leaks.
+			12*time.Hour, // Default ttl for each entry
+			time.Hour,    // Cleanup interval
+		),
+	}
+	p.getAccessTokenFn = p.getAccessToken
```
krancour, line 96: Everything from here on down is hard to follow. I understand the rationale. You want pluggable behaviors and retry intervals, but I think that can be done more neatly.

krancour: Thinking about this more. You really cannot test anything here that _actually_ integrates with Google, which is why you've made those bits pluggable. So in the end, all you're testing is your retry loop. You could just write a `retry()` function and test that in isolation and it would come out simpler than all this. Alternatively, re-use existing retry functionality that's already well-tested.

pkg/credentials/gar/workload_identity_federation.go
fuskovic: Rationale here is if `init()` completes in like 2-3 seconds the timer goroutine will still stay alive for the remaining ~27 or so seconds for no useful reason so why not clean it up?

pkg/credentials/gar/workload_identity_federation.go
fuskovic: `Error()` from  `k8s.io/client-go/util/retry` doesn't support context-cancellation. 

`ExponentialBackoffWithContext` does e.g.

```go
func retryWithContext(
    ctx context.Context,
    maxAttempts int,
    interval time.Duration,
    fn func(context.Context) error,
) error {
    var lastErr error
    _ = wait.ExponentialBackoffWithContext(
        ctx,
        wait.Backoff{Duration: interval, Steps: maxAttempts},
        func(ctx context.Context) (bool, error) {
            if err := fn(ctx); err != nil {
                lastErr = err
                return false, nil // return nil to keep retrying; error would abort immediately
            }
            return true, nil
        },
    )
    return lastErr
}
```

But I think I prefer:

```go

func retryWithContext(
	ctx context.Context,
	maxAttempts int,
	interval time.Duration,
	fn func(context.Context) error,
) error {
	var err error
	for i := range maxAttempts {
		if err = fn(ctx); err == nil {
			return nil
		}
		if i < maxAttempts-1 {
			select {
			case <-time.After(interval):
			case <-ctx.Done():
				return err
			}
		}
	}
	return err
}
```

Do you have a preference here?

pkg/credentials/gar/workload_identity_federation.go
krancour: You are obviously correct that if you create a new context, as you have here, you should cancel it when done with it. I think the meaning of my question wasn't clear.

Forget `cancel()` for now. It's a byproduct of the 30s deadline you're setting. I was trying to ask why we're setting that timer given that the retry loop puts its _own_ (lesser) boundaries on how long initialization is allowed to take.

pkg/credentials/gar/workload_identity_federation.go
krancour: If context cancellation is a real concern, `retry.OnError()` accepts what's usually a one-liner predicate function that judges what errors are retryable or not. You just make it return false on context cancellation.

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -13,69 +13,75 @@ import (
...
+const (
+	gcpResourceNameFormat = "projects/-/serviceAccounts/kargo-project-%s@%s.iam.gserviceaccount.com"
+	initMaxAttempts       = 5
+	initRetryInterval     = time.Second
+)
 
 func init() {
-	if provider := NewWorkloadIdentityFederationProvider(context.Background()); provider != nil {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
krancour: I see the 30s timer is gone now, but this child context of `context.Background()` is unnecessary. Just pass `context.Background()`.

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -13,69 +13,75 @@ import (
...
-		)
+	var projectID string
+	var tokenSource oauth2.TokenSource
+	if err := retry.OnError(
+		wait.Backoff{
+			Steps:    initMaxAttempts,
+			Duration: initRetryInterval,
+		},
+		func(_ error) bool {
+			return ctx.Err() == nil
```
krancour, line 58: Oh... I'd been thinking `errors.As()` Your approach is nicer. 👍

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -13,69 +13,75 @@ import (
...
-			"error getting GCP project ID; GCP Workload Identity Federation disabled",
-		)
+	var projectID string
+	var tokenSource oauth2.TokenSource
+	if err := retry.OnError(
+		wait.Backoff{
+			Steps:    initMaxAttempts,
+			Duration: initRetryInterval,
+		},
+		func(_ error) bool {
```
krancour:
```suggestion
		func(error) bool {
```

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -13,69 +13,75 @@ import (
...
+				return fmt.Errorf("error getting GCP project ID: %w", err)
+			}
+			tokenSource, err = google.DefaultTokenSource(ctx, iamcredentials.CloudPlatformScope)
+			if err != nil {
+				return fmt.Errorf("error getting GCP default token source: %w", err)
+			}
+			return nil
+		},
+	); err != nil {
+		logger.Info("GCP Workload Identity Federation provider could not be initialized", "err", err)
```
krancour, line 73: This was a nice choice to log the error but log it as info. It will be visible enough if WIF _should_ be working and isn't and won't look more severe than it really is for the case where this is running on GKE without WIF configured.

pkg/credentials/gar/workload_identity_federation.go
krancour: Not a problem introduced by this PR, but if we're already touching this code, I'd recommend fixing. This is the one and only place this constant is used. It'd turn it into a string literal here instead. Format strings as constants are hard to understand and should be avoided if possible.

Non-blocking.

pkg/credentials/gar/workload_identity_federation.go
```diff
@@ -180,8 +193,8 @@ func (p *WorkloadIdentityFederationProvider) GetCredentials(
 	}, nil
 }
 
-// getAccessToken returns a GCP access token retrieved using the provided base64
-// encoded service account key. The access token is valid for one hour.
+// getAccessToken returns a Project-specific GCP access token by impersonating
+// the kargo-project-<project> service account via the IAM Credentials API.
```
krancour: Project is an overloaded term in this component. There's the Kargo Project and the GCP project. I'm not sure which of the two you're referring to here. If the former, this comment is inaccurate because the code _tries_ to do that, but falls back on just using the controller's own identity. If it was the latter, the statement is accurate, but still confusing due to the overload.

Either way, if you want to change this comment, please disambiguate "project."

krancour (approved):
Though I called out a few nits, pre-emptively LGTM.

Issue #5855, opened 2026-03-04
Freight should manage its own `discoveryTimestamp`, not rely on `creationTimestamp`


jessesuen:
# Proposed Feature

Currently, the `creationTimestamp` of a Freight item is significant:
* it affects ordering in the UI
* it affects what is considered what is the "latest" Freight eligible for a promotion

The fact that we rely so much on `metadata.creationTimestamp` makes Freight unfriendly for backup/import/export tools, which have no control over the `creationTimestamp`.

This proposal is to switch to a model where `creationTimestamp` no longer matters in Kargo; instead, we would introduce a new timestamp field on the Freight object itself, e.g. `discoveryTimestamp`. This timestamp would be the one the UI and controller rely on for Kargo's application logic.

# Motivation

In an import/export scenario, Freight had to be created in the same order that they were originally discovered. This meant that Freight could be imported only at a rate of 1 second per Freight, so that the creationTimestamp would be different for each Freight.

Brightside56:
> This proposal is to switch to a model where creationTimestamp no longer matters in Kargo; instead, we would introduce a new timestamp field on the Freight object itself, e.g. discoveryTimestamp. This timestamp would be the one the UI and controller rely on for Kargo's application logic.

How this is supposed to work with `NewestBuild` strategy for images?

krancour:
> How this is supposed to work with NewestBuild strategy for images?

It's unrelated.

This is exclusively about adding a synthetic `creationTimestamp` to resources that are part of a time series so that their chronological order is maintained in a backup/restore scenario.

2003Aditya:
Hey! @krancour , 
i saw the issue and read the code related to this , and i think to add `DiscoveryTimestamp`  in `freight_types.go` , `warehouses.go`  , `query_freights_v1alpha1.go` and `regular_stages.go` .
lgtm?

krancour:
@2003Aditya you're on the right track. Promotions have the same problem, btw. They also should have DiscoveryTimestamp.

jessesuen:
> Promotions have the same problem, btw. They also should have DiscoveryTimestamp

Good call out. Though I would call it `status.startedAt` since that's the name we use for the individual step's start time.

BWagenerGenerali:
This is exactly what we wanted with #4494, so I'm happy to see this shaping up :)

The linked issue is currently "Not planned", but could be considered resolved when this is done.

PR #6133, opened 2026-04-21
fix: Freight should manage its own discoveryTimestamp, not rely on cr…

2003Aditya:
#5855


api/v1alpha1/freight_types.go
```diff
@@ -30,6 +30,10 @@ type Freight struct {
 	// present, the defaulting webhook will choose an available alias and assign
 	// it to both the field and label.
 	Alias string `json:"alias,omitempty" protobuf:"bytes,7,opt,name=alias"`
+	// DiscoveryTimestamp is the time at which the Freight was first discovered
+	// from its source. This is distinct from metadata.creationTimestamp, which is
+	// set by Kubernetes and cannot be controlled by users.
+	DiscoveryTimestamp *metav1.Time `json:"discoveryTimestamp,omitempty" protobuf:"bytes,11,opt,name=discoveryTimestamp"`
```
krancour, line 36: It appears codegen wasn't run after this change. Among many other things, codegen will also set the `protobuf` struct tag for you (making it harder to get wrong).

I would suggest this change:


```suggestion
	DiscoveryTimestamp *metav1.Time `json:"discoveryTimestamp,omitempty"`
```

When you run codegen (`make hack-codegen`), which you will need to do for reasons other than just this, it will _probably_ put it back to exactly as you had it here, but it may do something slightly different. Better safe than sorry.

api/v1alpha1/promotion_types.go
```diff
@@ -142,6 +142,10 @@ func (s PromotionStepStatus) Compare(other PromotionStepStatus) int {
 type Promotion struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+	// DiscoveryTimestamp is the time at which the Promotion was created.
+	// This is distinct from metadata.creationTimestamp, which is
+	// set by Kubernetes and cannot be controlled by users.
+	DiscoveryTimestamp *metav1.Time `json:"discoveryTimestamp,omitempty" protobuf:"bytes,4,opt,name=discoveryTimestamp"`
```
krancour: Same comment as here: https://github.com/akuity/kargo/pull/6133/changes#r3118215938

pkg/controller/stages/regular_stages.go
```diff
@@ -1721,10 +1721,15 @@ func (r *RegularStageReconciler) autoPromoteFreight(
...
+		// Find the latest Freight by sorting the available Freight by discovery time
 		// in descending order.
 		slices.SortFunc(freight, func(lhs, rhs kargoapi.Freight) int {
-			return rhs.CreationTimestamp.Compare(lhs.CreationTimestamp.Time)
+			lhsTime := lhs.DiscoveryTimestamp
+			rhsTime := rhs.DiscoveryTimestamp
+			if lhsTime == nil || rhsTime == nil {
+				return rhs.CreationTimestamp.Compare(lhs.CreationTimestamp.Time)
+			}
+			return rhsTime.Time.Compare(lhsTime.Time)
```
krancour: This was really smart to fall back on `CreationTimestamp` since existing Freight will obviously not already have a `DiscoveryTimestamp`.

I might suggest a small improvement though. Instead of falling back on CreationTimestamp for _both_ when _either_ doesn't have a `DiscoveryTimestamp`, I would let each of them fall back individually like this:

```suggestion
			lhsTime := lhs.DiscoveryTimestamp
			if lhsTime == nil {
				lhsTime = lhs.CreationTimestamp.Time
			}
			rhsTime := rhs.DiscoveryTimestamp
			if rhsTime == nil {
				rhsTime = rhs.CreationTimestamp.Time
			}
			return rhsTime.Time.Compare(lhsTime.Time)
```

pkg/controller/stages/regular_stages.go
```diff
@@ -1811,11 +1816,16 @@ func (r *RegularStageReconciler) autoPromoteFreight(
...
+if len(promotions.Items) > 0 {
+			// Sort the terminal Promotions by discovery time in descending order
+			slices.SortFunc(promotions.Items, func(lhs, rhs kargoapi.Promotion) int {
+				lhsTime := lhs.DiscoveryTimestamp
+				rhsTime := rhs.DiscoveryTimestamp
+				if lhsTime == nil || rhsTime == nil {
 					return rhs.CreationTimestamp.Compare(lhs.CreationTimestamp.Time)
-				})
+				}
+				return rhsTime.Time.Compare(lhsTime.Time)
```
krancour: Same comment as here: https://github.com/akuity/kargo/pull/6133/changes#r3118278060

I start to wonder whether we should include a generic `SortByDiscoveryDate()` helper somewhere.

pkg/server/query_freights_v1alpha1.go
```diff
@@ -292,7 +292,12 @@ func sortFreightSlice(orderBy string, reverse bool, freight []*kargoapi.Freight)
 				return strings.Compare(lhsTag, rhsTag)
 			}
 		}
-		return lhs.CreationTimestamp.Compare(rhs.CreationTimestamp.Time)
+		lhsTime := lhs.DiscoveryTimestamp
+		rhsTime := rhs.DiscoveryTimestamp
+		if lhsTime == nil || rhsTime == nil {
+			return lhs.CreationTimestamp.Time.Compare(rhs.CreationTimestamp.Time)
+		}
+		return rhsTime.Time.Compare(lhsTime.Time)
```
krancour: Same as here: https://github.com/akuity/kargo/pull/6133/changes#r3118302803

Another point in favor of a generic helper.

krancour:
This looks pretty good overall @2003Aditya. Thank you for doing this.

One spot(s) I think may have been overlooked for sorting would be any API endpoints that return a list of Promotions.

api/v1alpha1/promotion_types.go
jessesuen: Can we call this `startedAt`? That's the field name we use in `stepExecutionMetadata`, and "discovery" is more of a freight thing.

```yaml
kind: Promotion
metadata:
  creationTimestamp: 2026-04-08T15:50:21Z
  name: dev.01knpwk5z8ngczawgqctc79w39.efe1b47
status:
  startedAt: 2026-04-08T15:40:22Z   # <<<<< my preference
  finishedAt: 2026-04-08T15:51:10Z
  phase: Succeeded
  stepExecutionMetadata:
    - alias: task-1::step-1
      finishedAt: 2026-04-08T15:50:24Z
      startedAt: 2026-04-08T15:50:22Z
      status: Succeeded
```

jessesuen: Also, unlike Freight, for the promotion object, this needs to be under `status`. Not a top-level field.

jessesuen (changes requested):

2003Aditya:
@jessesuen @krancour Okay, I will change `discovery` to `startAt` and make the rest of the changes as well.

pkg/kargo/promotion_builder.go
```diff
@@ -82,6 +83,7 @@ func (b *PromotionBuilder) Build(
 			Namespace:   stage.Namespace,
 			Annotations: annotations,
 		},
+		DiscoveryTimestamp: &metav1.Time{Time: time.Now()},
```
jessesuen: Since I'm asking for StartedAt timestamp to be a `status` thing, this will no longer be appropriate to set here. It's also something that the controller would need to set when it transitions the Promotion from `Pending` to `Running`.

krancour:
@2003Aditya were you intending to do any further work on this or should I look for someone from my team to take over the PR?

EronWright:
Superseded by https://github.com/akuity/kargo/pull/6494

PR #6153, opened 2026-04-23
feat: API support for filtering Stages/Freight by Warehouse names

jessesuen:
## Summary

Adds warehouse-name filtering to the Kargo API for Stages and Freight, so callers can scope list/watch operations to resources associated with specific Warehouses.

I named the field `origins` for the Freight filter (to be consistent with existing QueryFreight field name), and `freightOrigins` for the Stage filter.

NOTES:
* QueryFreight (ConnectRPC) and queryFreight (REST) already have an `origins` filter, so it was unnecessary to add one there.
* There is no watch capability in queryFreight (REST) as of now, but when it is added, it will need the filtering support.

## Proto / generated code
  - ListStagesRequest gains a freightOrigins field (repeated string) — Warehouse names to filter by
  - WatchStagesRequest gains the same freightOrigins field
  - WatchFreightRequest gains an origins field (repeated string) — Warehouse names to filter Freight by

## Server-side filtering logic (pkg/server/)
  - New listStagesByWarehouses helper fetches all Stages in a project and filters client-side using stageMatchesAnyWarehouse, which checks each Stage's RequestedFreight[*].Origin fields. Filtering has OR semantics and includes both direct subscribers (Sources.Direct = true) and indirect subscribers (those that
  receive Freight via upstream Stages).
  - ListStages (ConnectRPC) and listStages (REST GET /v1beta1/projects/:project/stages) both pass the freightOrigins query parameter through to listStagesByWarehouses.
  - watchStages (REST ?watch=true) applies the same stageMatchesAnyWarehouse filter in its SSE event loop.
  - WatchStages (ConnectRPC) applies stageMatchesAnyWarehouse in its watch event loop.
  - WatchFreight (ConnectRPC) filters watch events with slices.Contains(warehouses, freight.Origin.Name).

## Tests
  - TestListStages (ConnectRPC) and Test_server_listStages / Test_server_listStages_watch (REST): cover the no-filter, single-warehouse, and multi-warehouse cases, including indirect subscribers and deduplication. Parallel subtests now deep-copy shared stage objects before passing to WithObjects to eliminate a data
  race.
  - TestWatchStages (ConnectRPC): verifies that the freightOrigins filter in the WatchStages event loop passes direct and indirect warehouse subscribers while dropping unrelated Stages.
  - TestWatchFreight (ConnectRPC): verifies that the origins filter in the WatchFreight event loop passes matching Freight while dropping Freight from other Warehouses.

The UI changes will come as a separate PR, because they were blindly vibecoded and need a UI pair of eyes on them, but it is staged here: https://github.com/jessesuen/kargo/pull/1/changes if curious.


jacobboykin (approved):
lgtm

PR #6156, opened 2026-04-23
feat: add resourceVersion to Freight, Stage, Warehouse list and watch endpoints

jessesuen:
## Summary
- New proto fields: Added `resource_version` to the List response messages for Stages, Warehouses, and Freight (`ListStagesResponse`, `ListWarehousesResponse`, `QueryFreightResponse`), and to the Watch request messages (`WatchStagesRequest`, `WatchWarehousesRequest`, `WatchFreightRequest`).
- Backend — List handlers: The ConnectRPC and REST List handlers now return a meaningful `resourceVersion` in responses. Because the controller-runtime cached client always returns "0" for the list-level ResourceVersion, a new effectiveResourceVersion helper falls back to max(item.ResourceVersion) across all returned objects when the list-level value is absent or "0". This yields a valid lower-bound resource version without bypassing the cache.
- Backend — Watch handlers: The ConnectRPC and REST Watch handlers for Stages, Warehouses, and Freight now accept an optional `resourceVersion` parameter and forward it to the underlying Kubernetes Watch call. When provided, Kubernetes sends only events newer than that version, eliminating the full ADDED-event replay that previously occurred on every Watch connection.
- REST API: The REST queryFreight response now includes a top-level resourceVersion field alongside groups, mirroring the ConnectRPC behavior.

## Motivation

This enables the standard Kubernetes list-then-watch pattern: a client can call a List endpoint, record the returned resourceVersion, then start a Watch from that exact point to receive all subsequent changes without gaps or redundant initial events.

## Tests
- Tests: New unit tests for effectiveResourceVersion, and additional integration-style tests for ListStages and QueryFreight covering the resource version passthrough and fallback behaviors.


pkg/server/resource_version.go
```diff
@@ -0,0 +1,35 @@
...
+// lower bound for a subsequent Watch: any event with a higher resource version
+// is guaranteed to be newer than the listed state.
+//
+// Background: the controller-runtime cached client returns "0" for the
+// list-level ResourceVersion, which causes a Kubernetes Watch to replay all
+// existing objects as ADDED events. Using the maximum object-level resource
+// version avoids this replay while still being correct: every modification to
+// a listed object, and every new object created after the list, will have a
+// resource version greater than this value.
+func effectiveResourceVersion(rv string, itemVersions []string) string {
```
jessesuen, line 17: This was an interesting discovery during end-to-end testing.

The Kargo API uses the controller-runtimes cached client for list requests. Before this helper, the List call had blindly carried over the `resourceVersion` from the cached client's List response. But that value was always `0,` and so the result had no benefit for UI performance since the subsequent Watch would replay all objects.

This helper allows us to continue using the cached client while benefiting from a smaller payload in the subsequent Watch call by using the max resourceVersion from the cached list. In my testing/research, this works as advertised, but need to proceed cautiously here.

PR #6157, opened 2026-04-23
fix: use server-side apply for RBAC resources in management controllers

### LINKED ISSUE #6155: Management controllers generate excessive 409s by using Create→AlreadyExists→Update instead of server-side apply
## Description

The project and serviceaccount management controllers create RBAC resources (Roles, RoleBindings, ClusterRoles, ClusterRoleBindings, ServiceAccounts) using a `Create` → `AlreadyExists` → `Update` fallback pattern. In clusters with many projects, this generates a high volume of unnecessary 409 errors on every reconcile cycle.

**Affected files:**
- `pkg/controller/management/serviceaccounts/serviceaccounts.go` — `ensureControllerPermissions`
- `pkg/controller/management/projects/projects.go` — `ensureSystemPermissions`, `ensureControllerPermissions`, `ensureDefaultUserRoles`, `ensureExtendedPermissions`

## Current pattern

```go
if err := r.client.Create(ctx, roleBinding); err != nil {
    if !apierrors.IsAlreadyExists(err) {
        return err
    }
    r.client.Update(ctx, roleBinding) // always fires if resource exists
}
```

## The problem

- `ensureControllerPermissions` in the serviceaccount reconciler iterates over **all project namespaces** and attempts to `Create` a RoleBinding in each. Since the RoleBindings already exist after first reconcile, every subsequent reconcile generates N × 409s (where N = number of projects).
- The project reconciler does the same across 5+ resource types on every reconciliation, which is triggered frequently by Warehouse/Stage health condition changes.

## Observed impact

On a cluster with active Kargo usage, the k8s API server shows ~800 RBAC-related 409s per minute accumulating continuously:

```
26,589  rolebindings POST → 409
15,951  roles POST → 409
15,951  serviceaccounts POST → 409
 5,318  clusterroles POST → 409
 5,318  clusterrolebindings POST → 409
```

Every 409 is a wasted API server round trip (etcd read with no write). The subsequent `Update` then performs an unconditional etcd write even when the resource hasn't changed.

## Suggested fix

Replace the `Create`/`Update` pattern with **server-side apply**:

```go
if err := r.client.Patch(ctx, roleBinding, client.Apply,
    client.ForceOwnership,
    client.FieldOwner("kargo"),
); err != nil {
    return fmt.Errorf("error applying RoleBinding %q in namespace %q: %w",
        roleBinding.Name, roleBinding.Namespace, err)
}
```

**Benefits:**
- Single API call instead of two
- No 409s — server handles create-or-update atomically
- No etcd write when the object hasn't changed (unlike unconditional `Update`)
- Already the idiomatic pattern for Kubernetes controllers

alexmt:
## Summary

Fixes #6155

Replaces the `Create` → `AlreadyExists` → `Update` pattern with **server-side apply** (SSA) in the project and serviceaccount management controllers.

- In `newReconciler`, all `createXFn` function pointers now use an `applySSA` closure that calls `client.Patch(..., client.Apply, client.ForceOwnership, client.FieldOwner("kargo"))`. The GVK is populated automatically from the client's scheme so no TypeMeta boilerplate is needed at each call site.
- All call sites are simplified: AlreadyExists checks and fallback `Update` calls are removed.
- Same fix applied to `serviceaccounts.go` which called `r.client.Create`/`r.client.Update` directly.

**Before:** every reconcile cycle generates N × 409s (one per project/resource), then a redundant `Update` even when nothing changed.

**After:** a single `PATCH` per resource — no 409s, and no etcd write when the desired state already matches.

### DIFF
--- a/pkg/controller/management/projects/projects.go
+++ b/pkg/controller/management/projects/projects.go
@@ -253,11 +253,19 @@ func newReconciler(kubeClient client.Client, cfg ReconcilerConfig) *reconciler {
 	r.ensureControllerPermissionsFn = r.ensureControllerPermissions
 	r.ensureDefaultUserRolesFn = r.ensureDefaultUserRoles
 	r.ensureExtendedPermissionsFn = r.ensureExtendedPermissions
-	r.createServiceAccountFn = r.client.Create
-	r.createRoleFn = r.client.Create
-	r.createRoleBindingFn = r.client.Create
-	r.createClusterRoleFn = r.client.Create
-	r.createClusterRoleBindingFn = r.client.Create
+	applySSA := func(ctx context.Context, obj client.Object, _ ...client.CreateOption) error {
+		gvks, _, err := r.client.Scheme().ObjectKinds(obj)
+		if err != nil {
+			return fmt.Errorf("could not determine GVK for object: %w", err)
+		}
+		obj.GetObjectKind().SetGroupVersionKind(gvks[0])
+		return r.client.Patch(ctx, obj, client.Apply, client.ForceOwnership, client.FieldOwner("kargo"))
+	}
+	r.createServiceAccountFn = applySSA
+	r.createRoleFn = applySSA
+	r.createRoleBindingFn = applySSA
+	r.createClusterRoleFn = applySSA
+	r.createClusterRoleBindingFn = applySSA
 	r.deleteClusterRoleFn = r.client.Delete
 	r.deleteClusterRoleBindingFn = r.client.Delete
 	return r
@@ -686,22 +694,12 @@ func (r *reconciler) ensureSystemPermissions(
 	for _, roleBinding := range roleBindings {
 		rbLogger := logger.WithValues("roleBinding", roleBinding.Name)
 		if err := r.createRoleBindingFn(ctx, &roleBinding); err != nil {
-			if !apierrors.IsAlreadyExists(err) {
-				return fmt.Errorf(
-					"error creating RoleBinding %q in Project namespace %q: %w",
-					roleBinding.Name, project.Name, err,
-				)
-			}
-			if err = r.client.Update(ctx, &roleBinding); err != nil {
-				return fmt.Errorf(
-					"error updating existing RoleBinding %q in Project namespace %q: %w",
-					roleBinding.Name, project.Name, err,
-				)
-			}
-			rbLogger.Debug("updated RoleBinding")
-			continue
+			return fmt.Errorf(
+				"error applying RoleBinding %q in Project namespace %q: %w",
+				roleBinding.Name, project.Name, err,
+			)
 		}
-		rbLogger.Debug("created RoleBinding in Project namespace")
+		rbLogger.Debug("applied RoleBinding in Project namespace")
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


jessesuen:
I have a question. If my understanding is correct, it seems we are continually attempting to update resources (perhaps on every reconciliation) even when it is not necessary (the object is the same). So even if we switch to SSA (which would be a good improvement and reduce API calls), nothing would stop us from _continually_ performing the apply. 

I wonder if there is a second improvement we can make that prevents the apply if we know it's not needed.

krancour:
@jessesuen I didn't look at this in detail yet, but have a recollection of there having been a guard to avoid unnecessary updates over and over, though it's possible I'm confusing it with another similar spot elsewhere in the code. If it's absent, I agree it should be there.

pkg/controller/management/projects/projects.go
```diff
@@ -253,11 +253,19 @@ func newReconciler(kubeClient client.Client, cfg ReconcilerConfig) *reconciler {
...
-	r.createRoleBindingFn = r.client.Create
-	r.createClusterRoleFn = r.client.Create
-	r.createClusterRoleBindingFn = r.client.Create
+	applySSA := func(ctx context.Context, obj client.Object, _ ...client.CreateOption) error {
+		gvks, _, err := r.client.Scheme().ObjectKinds(obj)
+		if err != nil {
+			return fmt.Errorf("could not determine GVK for object: %w", err)
+		}
+		obj.GetObjectKind().SetGroupVersionKind(gvks[0])
+		return r.client.Patch(ctx, obj, client.Apply, client.ForceOwnership, client.FieldOwner("kargo"))
```
EronWright, line 262: One implication of switching to `Apply` is that the controller will NOT overwrite third-party contributions (of annotations, the RBAC subjects of a given CRB, etc).  Looking at the current code, the `Update` call would've replaced everything.  I think the SSA behavior is a good thing, just wanted to call out that the remediation behavior is slightly different.

Also, is the field manager name the same before-after here?  It should be, to ensure that a newer version of Kargo replaces any earlier's contribution to a given object.

krancour:
> it's possible I'm confusing it with another similar spot elsewhere in the code

And I was.

pkg/controller/management/projects/projects.go
krancour: Good difference in behavior to call out.

The purpose of this code, as is reflected in many of the comments on methods in this file, is to ensure the existence of RBAC resources required for the system to operate correctly upon each Project, so...

_Overwriting_ "third-party contributions" was a feature; not a bug. If these are tampered with, things can break.

jessesuen:
I can confirm through code inspection that we are [ensuring RBAC](https://github.com/akuity/kargo/blob/596be3dee64b3e99654eb4f72ff5fa8833ea2ecc/pkg/controller/management/projects/projects.go#L505-L556) with every Project reconciliation. So while SSA **_would_** help prevent the object's `resourceVersion` from constantly being updated, it would still result in an API call. To eliminate unnecessary API calls entirely, we could check what we have in an informer cache is already what we want and skip the call(s) completely. Even something like a DeepEquals check would help prevent this problem.

jessesuen:
I'm targeting v1.11.0 for the potential switch to SSA, since there is debate if we are okay *preserving* users' annotations (which will be allowed with SSA), versus clobbering them completely (current behavior). Personally, I do see value in allowing users to **_annotate_** the OIDC claims of our generated ServiceAccounts, so they don't need to duplicate the entire set of Role/ServiceAccount/RoleBindings just to have an OIDC claim map to the `kargo-admin` role. But I do agree that the actual RBAC definitions should not drift from what we want (which might happen if we allow SSA).

To fix the continuous Update calls, I feel there is a safer approach to avoid unnecessary API calls using a DeepEquals check against the RBAC config. This avoids the change in behavior (risk) that comes with SSA, and is something I would like in v1.10.x. I have filed https://github.com/akuity/kargo/issues/6193 for that.

fuskovic:
> To fix the continuous Update calls, I feel there is a safer approach to avoid unnecessary API calls using a DeepEquals check against the RBAC config. This avoids the change in behavior (risk) that comes with SSA, and is something I would like in v1.10.x. I have filed #6193 for that.

I have the suggested approach implemented here: https://github.com/akuity/kargo/pull/6194

hairyhum:
> I do see value in allowing users to annotate the OIDC claims of our generated ServiceAccounts, so they don't need to duplicate the entire set of Role/ServiceAccount/RoleBindings just to have an OIDC claim map to the kargo-admin role.

As an off-topic, I wonder what the purpose of that would be?

jessesuen:
> As an off-topic, I wonder what the purpose of that would be?

After looking at the current implementation, my concern may be a non-issue. I am concerned about the following scenario:
1. Kargo vends `kargo-admin`, `kargo-promoter`, `kargo-viewer` SA/Role/RoleBinding
2. User adds annotations to one of the above ServiceAccounts, in order to map it to OIDC groups/users/etc... This can be done using the UI.
3. Controller **_undoes_** the user's change to the ServiceAccount annotations because it replaces/updates it to the hardwired version.

It turns out that the way `ensureDefaultUserRoles()` is currently implemented, (3) never happens because it only creates ServiceAccounts but never updates them. As long as we keep that behavior, I am good.

krancour:
@jessesuen is this something that can be closed because of alternative fixes @fuskovic implemented and merged for the one specific issue this PR was aimed at?

PR #6158, opened 2026-04-23
fix(conditions): truncate condition messages exceeding Kubernetes 32768-byte limit

### LINKED ISSUE #6147: Promotion stuck if ArgoCD error message is too big
# Description

When a freight is "argocd updated" to a stage, it helpfully includes the status of the ArgoCD sync operation.  Once, when someone had a syntax error in a large pod spec, the ArgoCD error message included the entire podspec json.  Kargo then failed to update its status, since the error message was too big for the condition message.

# Steps to Reproduce

* A simple Deployment, but with a large enough json serialization that it is > 32 kilobytes.  e.g. environment variables and scripts, managed fields, etc.
* A kargo project, warehouse, and promotion that includes an ArgoCD update step
* Make a deliberate error in the Deployment manifests (in my case forgetting to indent the Lifecycle correctly).
* Use Kargo to promote this to a stage
* ArgoCD will attempt to update the deployment and include the deployment current state, in its error message.  This message is > 32 kilobytes.
* Kargo will attempt to include this error message in the stage's status.

The stage status is not updated, but remains perpetually "pending".

Workaround: delete the stage; when it was recreated, it could deploy new freight as before.

# Version

(I don't use the kargo CLI and haven't set it up for auth.

```
v1.9.5
```

# Logs

```
2026-04-22T09:42:49Z	ERROR	stages/regular_stages.go:547	failed to update Stage status after syncing Promotions: Stage.kargo.akuity.io "REDACTED" is invalid: [status.conditions[0].message: Too long: may not be more than 32768 bytes, <nil>: Invalid value: null: some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]	{"namespace": "REDACTED-release-control", "stage": "REDACTED", "controlFlow": false}
github.com/akuity/kargo/pkg/controller/stages.(*RegularStageReconciler).reconcile
	github.com/akuity/kargo/pkg/controller/stages/regular_stages.go:547
github.com/akuity/kargo/pkg/controller/stages.(*RegularStageReconciler).Reconcile
	github.com/akuity/kargo/pkg/controller/stages/regular_stages.go:368
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile
	sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:216
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:461
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:421
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1
	sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:296
```

EronWright:
## Summary

Fixes #6147

When an upstream source such as ArgoCD embeds a large payload (e.g. an entire pod spec JSON) in a sync error, that text flows through the promotion engine into a Stage condition. Kubernetes enforces a hard 32768-byte limit on `status.conditions[*].message`, so every subsequent `PatchStatus` call fails, leaving the Stage permanently stuck in "pending".

- Truncate condition messages to the 32768-byte limit inside `conditions.Set`, the single choke point for all condition writes across all resource types
- Truncation is UTF-8-safe (walks back to a valid rune boundary) and appends `" ... (truncated)"` so operators know the message was cut

## Test plan

- [x] `go test ./pkg/conditions/...` passes — new `TestTruncateMessage` and `TestSet_TruncatesLongMessage` cases cover the boundary, the over-limit, and the multi-byte UTF-8 split case

### DIFF
--- a/pkg/conditions/conditions_test.go
+++ b/pkg/conditions/conditions_test.go
@@ -498,6 +498,23 @@ func TestSet(t *testing.T) {
 	}
 }
 
+func TestSet_TruncatesLongMessage(t *testing.T) {
+	t.Parallel()
+
+	longMsg := string(make([]byte, maxConditionMessageLength+100))
+	setter := &mockSetter{}
+	Set(setter, &metav1.Condition{
+		Type:    "Test",
+		Status:  metav1.ConditionTrue,
+		Reason:  "Reason",
+		Message: longMsg,
+	})
+
+	conds := setter.GetConditions()
+	require.Len(t, conds, 1)
+	require.LessOrEqual(t, len(conds[0].Message), maxConditionMessageLength)
+}
+
 func TestDelete(t *testing.T) {
 	const (
 		mockType1 = "MockType1"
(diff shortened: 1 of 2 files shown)


jacobboykin (approved):

pkg/conditions/conditions_test.go
```diff
@@ -733,6 +750,53 @@ func TestEqual(t *testing.T) {
...
+			name:  "over limit gets truncated",
+			input: string(make([]byte, maxConditionMessageLength+1)),
+			assert: func(t *testing.T, got string) {
+				require.LessOrEqual(t, len(got), maxConditionMessageLength)
+				require.Contains(t, got, "(truncated)")
+			},
+		},
+		{
+			name:  "multi-byte UTF-8 not split",
+			input: string(make([]byte, maxConditionMessageLength-1)) + "é", // 2-byte rune crosses boundary
```
mogsie: This doesn't really test the walking back, since the message is first truncated by 16 or so characters. Maybe test a message with 20 multibyte Runes at the end, and see that it remains a valid string?

EronWright: Good catch, thanks again for the review and the original report.

PR #6159, opened 2026-04-23
feat(api): add summary projection to ListStages/WatchStages + GetStageHealthOutputs RPC

jacobboykin:
## Summary

Adds a lightweight Stage projection for clients rendering list and graph views of many Stages at once. Instead of a new endpoint, a `summary` flag is threaded through the existing `ListStages` and `WatchStages` RPCs. When set, heavy fields are stripped from each returned Stage in place:

- `status.freightHistory` truncated to the current element (index 0)
- `spec.promotionTemplate.spec.steps[*]` — `config`, `if`, and `vars` cleared; `uses`, `task`, `as`, `continueOnError`, `retry` preserved so the UI can still derive step kind and count
- `status.health.output` cleared — fetch lazily via the new `GetStageHealthOutputs` RPC for Stages currently in viewport

The UI continues to derive `hasVerification` from `spec.verification != nil` and `promotionStepCount` from `len(spec.promotionTemplate.spec.steps)`.

## API

- `ListStages` / `WatchStages` gain `summary: bool`
- `ListStagesResponse` and `WatchStagesRequest` gain `resource_version` for list-then-watch
- New `GetStageHealthOutputs(project, stage_names[])` RPC + REST endpoint for lazy per-Stage health output fetch (batch cap 1000)

## Structure

Reusable Stage-domain helpers are exported in `pkg/api/stage.go` alongside the existing `GetStage`, `ListFreightAvailableToStage`, etc.:

- `api.StripStageForSummary(*Stage)` — in-place summary projection
- `api.ListStageHealthOutputs(ctx, c, project, names)` — parallel per-name health-output fetch (concurrency cap 16)

Server handlers are thin adapters.

## Measurements

Tested end-to-end against a live multi-project instance with around 1,900 Stages behind the public ingress, using the bundled UI on headless Chrome. With `summary=true`, gzipped wire size on `ListStages` drops by roughly 5–7× on large projects, JSON parse on the main thread (which blocks every cache write during a watch storm) is 2.4–3.9× faster, and heap retained per cached response drops about 4.8×. Browser-perceived fetch ends up 1.7–2.3× faster end-to-end. The wins scale with stage count, so they're most noticable where the pain is biggest.

What's left on the table: each Stage still carries its promotion-step skeletons (kept so the UI can derive step kind and count without falling back to `GetStage`), which puts a roughly 10 KB floor on stages with hundreds of steps. The response and watch request carry `resourceVersion` for proper list-then-watch, but the controller-runtime cache reader doesn't populate it today, so resume currently lands as "watch from now"; a separate decision tracked in #6156.


jacobboykin:
Some numbers from a local stress test (500 Stages in one project, 10 `FreightHistory` entries each, realistic freight / verification data populated):

| Endpoint | Wire size | Latency | Per item |
|---|---:|---:|---:|
| `GET /v1beta1/projects/:p/stages` | 22.32 MB | ~100 ms warm | 45.7 KB |
| `GET /v1beta1/projects/:p/stage-summaries` | 3.72 MB | ~55 ms | 7.8 KB |

**~6× payload reduction**, consistent with the fields the summary omits (`freightHistory[1..]`, full `PromotionTemplate.spec.steps`, `Verification`).

JS heap impact from the browser, measured by fetching + `JSON.parse`ing each response, holding the result, and diffing `performance.memory.usedJSHeapSize`:

| Response | Heap delta |
|---|---:|
| `ListStages` (22 MB JSON) | +48 MB |
| `ListStageSummaries` (3.7 MB JSON) | +7.8 MB |

**~6× reduction in JS heap held by the parsed response**, tracking the payload reduction linearly.

`ListStages` response is byte-identical between `main` and this branch (23,409,060 bytes both times) — no regression on the existing endpoint.

pkg/server/list_stage_summaries_v1alpha1.go
```diff
@@ -0,0 +1,227 @@
...
+// @Description not need full FreightHistory, PromotionTemplate steps, or
+// @Description Verification configuration. Use GetStage for detail fields.
+// @Tags Core, Project-Level
+// @Security BearerAuth
+// @Produce json
+// @Param project path string true "Project name"
+// @Param freightOrigins query []string false "Warehouse name(s) to filter by" collectionFormat(multi)
+// @Success 200 {object} svcv1alpha1.ListStageSummariesResponse
+// @Router /v1beta1/projects/{project}/stage-summaries [get]
+func (s *server) listStageSummaries(c *gin.Context) {
```
krancour: Is there a defensible reason for this being a dedicated endpoint instead of just adding an optional `summary=true` query param to the existing endpoint, in which case, you'll just drop the fields that "summary" excludes?

jacobboykin:
I'm going to close this one for now.

The underlying idea here is still valid: large Stage payloads can be expensive, and a summary projection plus lazy health-output fetches could be useful if `ListStages` payload size remains a dominant bottleneck. But this PR ended up carrying a lot of API, generated-client, REST/Connect, and UI cache-safety surface area.

Since opening this, we've validated a narrower backend fix for the more direct issue: list-then-watch flows should not replay the resources the UI just listed. That directly addresses the main performance problem we saw with large projects, including the Promotion watch case, with much less behavioral/API complexity.

So I think the right move is to close this rather than keep rebasing and expanding it. If profiling after the watch/resourceVersion work lands still shows `ListStages` payload size or JSON parse/heap cost as the next major bottleneck, we can revisit a smaller version of this idea with that sharper evidence.

PR #6164, opened 2026-04-24
feat(projects): support contributor registry for project setup and cleanup

EronWright:
## Summary

Follow-up to #6034 (specifically [Kent's ask](https://github.com/akuity/kargo/pull/6034#issuecomment-4284819279)).

Introduces `ProjectSetupContributorRegistration` and `ProjectCleanupContributorRegistration` registries, allowing downstream distributions to inject additional setup and cleanup logic into the project reconciler at startup time.

Both registries follow the same `pkg/component.PredicateBasedRegistry` + `GetAll()` pattern introduced in #6034, with `*kargoapi.Project` as the predicate arg — enabling contributors to opt out for specific projects if needed.

**Ordering** is intentionally symmetric:
- **Setup:** OSS base resources first, then contributors (OSS → EE)
- **Cleanup:** Contributors first, then OSS base resources (EE → OSS), so EE resources that reference OSS ClusterRoles are removed before those ClusterRoles are deleted. This also fixes a non-determinism bug in the previous cleanup, which used map iteration to decide deletion order.

As a result, `ensureExtendedPermissions` and all EE-specific fields in `ReconcilerConfig` are removed from OSS. The EE companion PR registers its logic as contributors instead.

## Test plan

- [x] Unit tests for setup contributor error/success paths in `TestReconciler_syncProject_setupContributors`
- [x] Unit tests for cleanup contributor ordering and error/success paths in `TestReconciler_cleanupProject_cleanupContributors`

### DIFF
--- a/pkg/controller/management/projects/project_setup_contributor.go
+++ b/pkg/controller/management/projects/project_setup_contributor.go
@@ -1,0 +1,74 @@
+package projects
+
+import (
+	"context"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/pkg/component"
+)
+
+type (
+	// projectSetupContributorPredicate returns true if the contributor has
+	// setup work to perform for the given Project.
+	projectSetupContributorPredicate = func(context.Context, *kargoapi.Project) (bool, error)
+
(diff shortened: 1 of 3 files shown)


fuskovic (approved):
LGTM

PR #6166, opened 2026-04-25
feat: add shard liveness API and UI indicator

krancour:
Surface per-shard liveness in the API and UI, derived from heartbeat ConfigMaps in the Kargo namespace. A shard is considered alive when its heartbeat ConfigMap carries an observedAt timestamp within AGENT_STATUS_DEADLINE (default 10m), and dead otherwise.

- Add REST endpoint GET /v1beta1/system/shards returning each known shard's alive/dead status and last-heartbeat timestamp.
- Add api.defaultShardName chart value plus DEFAULT_SHARD_NAME on the API server, used to resolve Stages with no explicit spec.shard.
- Render a small shard-status indicator (broadcast tower icon, green for alive, red for dead, hidden when no shard data is available) on each Stage card in the project DAG view.
- Default the Tilt dev environment to a single shard named 'tilt-shard' so the indicator has something to render.

cc @rpelczar, I let Claude try the UI parts. I don't feel git to evaluate front end stuff. I can at least tell you it doesn't _look_ spectacular. Would you mind having a look and amending the PR directly if you would like to change anything?

Screenshot 2026-04-25 at 12 48 19 AM


pkg/server/config/config.go
```diff
@@ -37,7 +37,14 @@ type ServerConfig struct {
...
 	SystemResourcesNamespace    string
 	KargoNamespace              string
-	RestConfig                  *rest.Config
+	// DefaultShardName is the name of the shard that Stages with no explicit
+	// spec.shard should be associated with for shard-liveness purposes. When
+	// empty, such Stages are not associated with any shard.
+	DefaultShardName string
+	// AgentStatusDeadline is the maximum age of a shard heartbeat before the
+	// shard is considered dead.
+	AgentStatusDeadline time.Duration
```
thomastaylor312, line 46: Nit: Suggest renaming to `ShardStatusDeadline`

thomastaylor312 (approved):
Good to go from the backend perspective. Someone with more frontend experiences needs to look at that side of things

krancour:
Closing in favor of #6184

PR #6178, opened 2026-04-27
feat(cli): add delete freight command

### LINKED ISSUE #6162: cli: add a delete freight command
The API for this already exists and this functionality exists in the UI, but is conspicuously absent from the CLI.

mail2sudheerobbu-oss:
## Summary

Adds a `kargo delete freight` subcommand to the CLI, resolving #6162.

Freight can be deleted by:
- **Name** (one or more positional arguments): `kargo delete freight --project=my-project abc123 def456`
- **Alias** (via `--alias` flag): `kargo delete freight --project=my-project --alias=my-alias`

Name and alias are mutually exclusive. The `--project` flag defaults to the configured default project from config.

## Changes

- `pkg/cli/cmd/delete/freight.go`: new command implementation following the existing stage/warehouse pattern
- `pkg/cli/cmd/delete/freight_test.go`: unit tests for `validate()`
- `pkg/cli/cmd/delete/delete.go`: register the new `freight` subcommand

## Testing

```
go test ./pkg/cli/cmd/delete/... -run Test_deleteFreightOptions_validate -v
```

Closes #6162

### DIFF
--- a/pkg/cli/cmd/delete/delete.go
+++ b/pkg/cli/cmd/delete/delete.go
@@ -76,6 +76,7 @@ kargo delete warehouse --project=my-project my-warehouse
 	// Register subcommands.
 	cmd.AddCommand(newClusterConfigCommand(cfg, streams))
 	cmd.AddCommand(newConfigMapCommand(cfg, streams))
+	cmd.AddCommand(newFreightCommand(cfg, streams))
 	cmd.AddCommand(newGenericCredentialsCommand(cfg, streams))
 	cmd.AddCommand(newRepoCredentialsCommand(cfg, streams))
 	cmd.AddCommand(newProjectCommand(cfg, streams))
--- a/pkg/cli/cmd/delete/freight.go
+++ b/pkg/cli/cmd/delete/freight.go
@@ -1,0 +1,136 @@
+package delete
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/pkg/cli/client"
+	"github.com/akuity/kargo/pkg/cli/config"
+	"github.com/akuity/kargo/pkg/cli/io"
+	"github.com/akuity/kargo/pkg/cli/kubernetes"
+	"github.com/akuity/kargo/pkg/cli/option"
+	"github.com/akuity/kargo/pkg/cli/templates"
+	"github.com/akuity/kargo/pkg/client/generated/core"
+)
+
+type deleteFreightOptions struct {
+	genericiooptions.IOStreams
+	*genericclioptions.PrintFlags
+
+	Config        config.CLIConfig
+	ClientOptions client.Options
+
+	Project string
+	Names   []string
+	Aliases []string
+}
+
+func newFreightCommand(cfg config.CLIConfig, streams genericiooptions.IOStreams) *cobra.Command {
+	cmdOpts := &deleteFreightOptions{
+		Config:     cfg,
+		IOStreams:  streams,
+		PrintFlags: genericclioptions.NewPrintFlags("deleted").WithTypeSetter(kubernetes.GetScheme()),
+	}
+
+	cmd := &cobra.Command{
+		Use:   "freight [--project=project] [--name=name] [--alias=alias]",
+		Short: "Delete freight by name or alias",
+		Args:  cobra.NoArgs,
+		Example: templates.Example(`
+# Delete a piece of freight by name
+kargo delete freight --project=my-project --name=abc123
+
+# Delete a piece of freight by alias
+kargo delete freight --project=my-project --alias=my-alias
+
+# Delete multiple pieces of freight by name
+kargo delete freight --project=my-project --name=abc123 --name=def456
+
+# Delete a piece of freight in the default project
+kargo config set-project my-project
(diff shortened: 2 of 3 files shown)


pkg/cli/cmd/delete/delete.go
```diff
@@ -81,6 +81,7 @@ kargo delete warehouse --project=my-project my-warehouse
 	cmd.AddCommand(newProjectCommand(cfg, streams))
 	cmd.AddCommand(newProjectConfigCommand(cfg, streams))
 	cmd.AddCommand(newRoleCommand(cfg, streams))
+	cmd.AddCommand(newFreightCommand(cfg, streams))
```
krancour: The subcommands were alphabetical before. Let's try to preserve that.

pkg/cli/cmd/delete/freight.go
```diff
@@ -0,0 +1,168 @@
...
+		IOStreams:  streams,
+		PrintFlags: genericclioptions.NewPrintFlags("deleted").WithTypeSetter(kubernetes.GetScheme()),
+	}
+
+	cmd := &cobra.Command{
+		Use:   "freight [--project=project] (NAME ... | --alias=alias)",
+		Short: "Delete freight by name or alias",
+		Args:  cobra.ArbitraryArgs,
+		Example: templates.Example(`
+# Delete a freight by name
```
krancour: "A freight" (even if you have seen it somewhere else within the project) is grammatically incorrect. This is a weird English thing. Freight is a https://en.wikipedia.org/wiki/Mass_noun. The word is essentially plural by default, so singularizing it requires extra words. We would typically say "a piece of freight" or "a freight resource." I prefer the latter, however, the former is what's currently used throughout the CLI, so for consistency, we should stick with that.

This needs to be fixed in a number of places throughout the PR.

pkg/cli/cmd/delete/freight.go
```diff
@@ -0,0 +1,168 @@
...
+kargo delete freight --project=my-project abc123
```
krancour: This isn't consistent with `kargo get freight`, which does not allow positional args, but requires using `--name` or `--alias` for each freight resource you are retrieving (or deleting in this case).

krancour: Applies in other spots as well.

mail2sudheerobbu-oss:
Thanks for the detailed review, @krancour! I've addressed all three points:

1. **Alphabetical order** — `newFreightCommand` is now registered between `newConfigMapCommand` and `newGenericCredentialsCommand` in `delete.go` (c → f → g).

2. **Positional args removed** — Replaced with a `--name` flag (repeatable). `Args` is now `cobra.NoArgs`, the `complete(args []string)` method is gone, the `slices` import is removed, and the `Use` string now reads `(--name=name ... | --alias=alias)`. Examples updated accordingly.

3. **Grammar fixed** — All example comments now read "Delete a piece of freight …" instead of "Delete a freight …".

pkg/cli/cmd/delete/freight.go
```diff
@@ -0,0 +1,163 @@
...
+type deleteFreightOptions struct {
+	genericiooptions.IOStreams
+	*genericclioptions.PrintFlags
+
+	Config        config.CLIConfig
+	ClientOptions client.Options
+
+	Project string
+	Names   []string
+	Alias   string
```
krancour: This doesn't appear to be consistent with get.

mail2sudheerobbu-oss:
Good catch, @krancour! Updated to match `get freight` exactly:

- `Alias string` → `Aliases []string` (plural, repeatable)
- `option.Name` → `option.Names`, `option.Alias` → `option.Aliases` (using the `StringArrayVar`-backed helpers)
- `run()` now iterates over `append(o.Names, o.Aliases...)`, same pattern as `get freight`
- Added `cmd.MarkFlagsMutuallyExclusive(option.NameFlag, option.AliasFlag)` so cobra enforces the constraint
- `validate()` simplified to mirror the `get freight` style (project check + name-or-alias check)
- `freight_test.go` updated to use `Aliases: []string{...}`

pkg/cli/cmd/delete/freight.go
```diff
@@ -0,0 +1,140 @@
...
+	o.AddFlags(cmd)
+
+	option.Project(cmd.Flags(), &o.Project, o.Config.Project,
+		"The Project for which to delete Freight. If not set, the default project will be used.")
+	option.Names(cmd.Flags(), &o.Names,
+		"Name of a piece of freight to delete. Mutually exclusive with --alias. May be specified more than once.")
+	option.Aliases(cmd.Flags(), &o.Aliases,
+		"Alias of a piece of freight to delete. Mutually exclusive with --name. May be specified more than once.")
+
+	cmd.MarkFlagsMutuallyExclusive(option.NameFlag, option.AliasFlag)
```
krancour: I don't think these need to be mutually exclusive.

`kargo get freight --name foo --alias bar` works, so it seems `kargo delete freight --name foo --alias bar` should as well.

"May be specified more than once." We should probably drop this. tbh, it's a nice touch, but no existing commands explicitly say this for flags that can be used more than once.

Even if not ideal, there _is_ a signal of plurality that automatically surfaces in the help, demonstrated using the existing `get` subcommand:

```
kargo get freight -h

# ...

Flags:
      --alias stringArray             The alias of a piece of freight to get.

# ...

      --name stringArray              The name of a piece of freight to get.

# ...
```

mail2sudheerobbu-oss:
Done, @krancour! Two changes in the latest push:

1. **Removed `cmd.MarkFlagsMutuallyExclusive`** — `--name` and `--alias` can now be used together, consistent with `kargo get freight`.
2. **Simplified flag descriptions** — dropped the "Mutually exclusive with..." and "May be specified more than once." verbiage. The `stringArray` type already signals plurality in the help output.

krancour (approved):
@mail2sudheerobbu-oss very nicely done. Thank you!

PR #6179, opened 2026-04-27
feat(chart): support additional API server hostnames via api.ingress.extraHosts

ajaytikoo:
### Summary

- Adds api.ingress.extraHosts — a list of additional hostnames that route to the same API server backend via the existing Ingress resource
- Extra hostnames are automatically registered as valid Dex redirect URIs, so OIDC login works from any configured hostname
- TLS is handled in one of two ways depending on api.ingress.tls.selfSignedCert:
  - true: the cert-manager-generated certificate's dnsNames is extended to cover all extra hosts automatically — no additional configuration needed
  - false: supply api.ingress.tls.extraHostsSecretName pointing to a pre-existing secret with a certificate covering the extra hosts (e.g. a wildcard cert)
- Webhook paths (/webhooks) are included on extra host rules when the external webhooks server is co-hosted on the API ingress, consistent with the primary host rule
- Refactors repeated port number if/else blocks in ingress.yaml into $apiPort/$webhookPort template variables

### Test plan

- [x] -  helm template with api.ingress.extraHosts empty — output is identical to pre-change
- [x] -  helm template with one or more extra hosts and selfSignedCert: true — extra hosts appear in Ingress rules, TLS hosts list, and cert dnsNames; Dex config includes extra redirect URIs
- [x] -  helm template with extra hosts and selfSignedCert: false + extraHostsSecretName set — extra hosts appear in a separate TLS stanza using the provided secret name
- [x] -  helm template with extra hosts and selfSignedCert: false + extraHostsSecretName empty — extra hosts get routing rules only (no TLS stanza), primary host TLS is unaffected
- [x] -  helm template with externalWebhooksServer.enabled: true and extra hosts — /webhooks path appears under each extra host rule


krancour:
Please open an issue to discuss things like this before opening a PR. I'm converting this to draft until such a conversation takes place.

krancour:
Apologies... I see there _is_ and open issue, but this wasn't linked to it.

Proposals still require discussion before implementing.

PR #6184, opened 2026-04-27
feat: make controllers produce heartbeats; surface liveness in ui

krancour:
Alternative to https://github.com/akuity/kargo/pull/6166

This PR causes controllers to produce a heartbeat — a `coordination.k8s.io/v1.Lease` in the Kargo namespace, renewed every 10s and deleted on graceful shutdown. The API server reads them at a new `GET /v1beta1/system/controller-heartbeats` endpoint, returning a map of every reported controller plus the resolved default-controller name. (`Lease`s are the idiomatic way of implementing heartbeats in k8s.)

On DAG load, the UI fetches the feed once and joins each Stage to its controller via `spec.shard` (or the default). When that controller's heartbeat is dead, missing, or has never been reported, the existing synthesized phase short-circuits to `Failed` with a tooltip explaining why.

cc @thomastaylor312 for back end. This is exactly as we've been discussing.

cc @rpelczar for front end. Consider any UI changes here a placeholder and you are welcome to completely change whatever you want and amend the PR directly.


pkg/heartbeat/renewer.go
```diff
@@ -0,0 +1,188 @@
...
+		holderIdentity = "kargo-controller"
+	}
+	return &renewer{
+		client:         c,
+		namespace:      namespace,
+		controllerName: controllerName,
+		leaseName:      leaseName,
+		holderIdentity: holderIdentity,
+		leaseDuration:  defaultLeaseDuration,
+		renewInterval:  defaultRenewInterval,
```
gdsoumya: can we make this configurable?

krancour: Is there an immediate need to do so?

krancour: Talked offline.

pkg/heartbeat/heartbeattest/heartbeattest.go
```diff
@@ -0,0 +1,67 @@
...
+		l.Spec.RenewTime = &metav1.MicroTime{Time: t}
+	}
+}
+
+// WithDuration sets the lease's leaseDurationSeconds, the validity window
+// applied to the renewTime by readers. Omit it to seed a record that
+// readers should treat as dead regardless of renewTime.
+func WithDuration(d time.Duration) Option {
+	return func(l *coordinationv1.Lease) {
+		l.Spec.LeaseDurationSeconds = ptr.To(int32(d.Seconds())) //nolint:gosec
```
thomastaylor312: Super duper nit: It should be self-evident, but might be worth adding "because this is a test" to this gosec comment for clarity

pkg/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Status is point-in-time liveness synthesized from a single controller's
+// heartbeat.
+type Status string // @name HeartbeatStatus
+
+const (
+	// StatusAlive represents a controller that has reported a heartbeat recently
+	// enough to be considered alive.
+	StatusAlive Status = "Alive"
+	// StatusDead represents a controller that has not reported a heartbeat
```
thomastaylor312, line 26: Do we need a third status called "unknown" or "unreported"? I know you mentioned it is treated as failure, which is ok. But it is more accurate to describe if something reported once and is now offline versus never having reported in before. This is useful information to anyone/anything consuming this data. Expired leases are kept around for an hour by default, so if it has been off for more than an hour, it would fall into this new state as well, so I think that would be fine

pkg/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, lease := range list.Items {
+		controllerName := lease.Labels[kargoapi.LabelKeyController]
+		out[controllerName] = leaseToHeartbeat(&lease, now)
+	}
+	return out, nil
+}
+
+// leaseToHeartbeat synthesizes a Heartbeat from a Lease and the current time.
+func leaseToHeartbeat(
+	lease *coordinationv1.Lease,
```
thomastaylor312: This is safe because it is internal, but we don't have a `nil` check on this in the function

pkg/heartbeat/renewer.go
```diff
@@ -0,0 +1,188 @@
...
+	leaseName      string
+	holderIdentity string
+	leaseDuration  time.Duration
+	renewInterval  time.Duration
+}
+
+// NewRenewer returns an implementation of controller-runtime's manager.Runnable
+// interface that creates, renews, and (on shutdown) deletes a heartbeat record
+// for a Kargo controller.
+func NewRenewer(
```
thomastaylor312, line 46: Nit: I don't love the "Renewer" name here. Maybe something like "Pinger," "Heartbeater", or "Pulsometer" (if you're feeling clever/funny)

pkg/heartbeat/renewer.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+}
+
+// Start implements controller-runtime's manager.Runnable interface. It produces
+// a heartbeat at a scheduled interval by creating or updating a Lease resource.
+func (r *renewer) Start(ctx context.Context) error {
+	logger := logging.LoggerFromContext(ctx)
+	logger.Info(
+		"Starting controller heartbeat",
+		"lease.name", r.leaseName,
```
thomastaylor312, line 85: Nit: I think at least this field (and _maybe_ duration) should be added using `WithFields` above so they're present on all errors below

pkg/heartbeat/renewer.go
```diff
@@ -0,0 +1,188 @@
...
+
+// NeedLeaderElection implements controller-runtime's manager.Runnable
+// interface. It explicitly reports false so the renewer runs on every replica
+// rather than just the leader. Kargo controllers don't currently leader-elect,
+// but we make this explicit for safety.
+func (r *renewer) NeedLeaderElection() bool { return false }
+
+func (r *renewer) renew(ctx context.Context) error {
+	now := metav1.MicroTime{Time: time.Now()}
+	durationSeconds := int32(r.leaseDuration.Seconds()) //nolint:gosec
```
thomastaylor312: Nit: Give an explanation for the gosec lint

pkg/server/kubernetes/client.go
```diff
@@ -197,6 +199,24 @@ func newDefaultInternalClient(
 					},
 				},
 			}
+			// The API server has only namespaced (Role-based) RBAC for
+			// Leases, so the default cluster-wide watch the cache would
+			// otherwise set up fails to start. Scope the lease informer to
+			// the Kargo namespace, where the API server's RBAC actually
+			// permits list/watch.
+			kargoNamespace := os.Getenv("KARGO_NAMESPACE")
```
thomastaylor312: It feels a little janky to be reading this env var deep in the client. I'd rather this be passed directly via configuration (unless this is a pattern we do elsewhere)

pkg/server/get_controller_heartbeats_v1alpha1.go
```diff
@@ -0,0 +1,64 @@
...
+// @Router /v1beta1/system/controller-heartbeats [get]
+func (s *server) getControllerHeartbeats(c *gin.Context) {
+	ctx := c.Request.Context()
+
+	// The heartbeat records live in the Kargo namespace and may not be
+	// readable by typical authenticated users. Use the internal (non-
+	// authorizing) client: controller liveness is operational information that
+	// should be visible to any authenticated user, and the data exposed
+	// here (controller name, alive/dead, last-seen timestamp) is non-sensitive.
+	heartbeats, err := heartbeat.GetAll(
```
thomastaylor312, line 52: Nit: Now that I've seen this in action, I think it would make more sense as a `HeartbeatClient` that has a `GetAll` method. That way we're not manually passing the client and namespace each time (plus more useful later if we want/need to use it anywhere else). Not big deal though since it is pretty simple

pkg/server/rest_router.go
```diff
@@ -97,6 +97,9 @@ func (s *server) setupRESTRouter(ctx context.Context) *gin.Engine {
 			system.POST("/cluster-config/refresh", s.refreshClusterConfig)
 			system.DELETE("/cluster-config", s.deleteClusterConfig)
 
+			// Controller heartbeats
+			system.GET("/controller-heartbeats", s.getControllerHeartbeats)
```
thomastaylor312, line 101: Should this be `/controller/heartbeats` in case we ever want to add any more controller-related endpoints?

thomastaylor312 (changes requested):
Most of these are nits, but there are a few discussion items

pkg/heartbeat/heartbeat.go
krancour: The thing is idk when we'd ever _use_ "unknown" or "unreported."

If the shape of any of this were a function answering the query "is x alive?" I could see that as making sense. You don't have a heartbeat for x -- not even a stale one? Hard to say it's dead because you can't say whether it was ever alive. "Unknown" could make sense.

But as it stands, the shape of this is just, "Give me all the heartbeats." A heartbeat that exists and is fresh reflects a controller that is alive. A heartbeat that is stale reflects a controller that was once alive and is now dead. There's an infinite number of possible controllers with an "unknown" state, but they don't appear in the list, so an unknown status doesn't buy us anything.

pkg/heartbeat/renewer.go
krancour: I didn't love it at first either. I still don't, tbh. But when you put it in the context of the package name...

`heartbeat.Pinger` sounds like it pings a heartbeat, which doesn't sound right.

`heartbeat.Heartbeater` stutters.

`hearbeat.Renewer` does make some kind of sense, as would something like `heartbeat.Producer`. If you have suggestions along these lines, I'd happily hear them.

pkg/heartbeat/renewer.go
krancour: I was going to initially, then I realized that every controller will only write leases having one single name that never changes. It didn't feel like valuable context to repeat with every log message. But it's not a hill I'd die on, so I'll happily change this.

pkg/server/kubernetes/client.go
krancour: I agree and had meant to fix this. Overlooked it somehow.

pkg/server/get_controller_heartbeats_v1alpha1.go
krancour: I'm ok with this either way. Choosing to leave this one be.

pkg/server/rest_router.go
krancour: Well... it would be `api/v1beta1/system/controller/...`. I can't really see controller stuff living under `system/`

krancour: After more thought, I'd definitely leave this as it is.

`/system/controller-heartbeats` is a system-wide aggregate, not a per-controller resource — it sits naturally alongside `/system/server-version`, `/system/cluster-config`, etc.

If we add per-controller endpoints later, the natural shape is `/controllers/:name[/...]`. Putting `heartbeats` under `/controllers/` (or `/system/controllers/`) would either collide with `:name` or reserve `heartbeats` as a name nobody can use for a controller.

So the current way feels most future-proof to me.

pkg/heartbeat/heartbeattest/heartbeattest.go
krancour: Fixed in badf00b8e832fc825723fa4545f59ef9d9c0b4d3

pkg/heartbeat/heartbeat.go
krancour: Fixed in badf00b8e832fc825723fa4545f59ef9d9c0b4d3

pkg/heartbeat/renewer.go
krancour: Fixed in badf00b8e832fc825723fa4545f59ef9d9c0b4d3

pkg/heartbeat/renewer.go
krancour: Fixed in badf00b8e832fc825723fa4545f59ef9d9c0b4d3

pkg/server/kubernetes/client.go
krancour: Fixed in 8e47f0fce48543dbdd85acc9239f24ea1d519b85

rpelczar (approved):

pkg/heartbeat/heartbeat.go
thomastaylor312: I think there is a useful difference between “the heartbeat was there but expired” (recent disconnect) and “heartbeat is gone” (never registered or died more than an hour ago). For the UI this matters less, but for an API consumer that could matter. 

Very much a nit though and not required because the counter argument is good as well

pkg/heartbeat/renewer.go
thomastaylor312: Producer isn’t bad and I’m kind of a fan of pulsometer but honestly not strongly opinionated here

pkg/heartbeat/renewer.go
thomastaylor312: Yeah I was just thinking more in the moment of debugging when you might be filtering multiple log lines in something like datadog or the like. Not a huge deal either way

thomastaylor312 (approved):

PR #6187, opened 2026-04-28
feat: promotion orchestrator access for cluster configs + shared secrets

fuskovic:
### Problem

The promotion orchestrator needs to read two resources it currently has no RBAC access to:

- `ClusterConfig`: Needs to be read at step execution time to resolve the system-level git user (name, email, signing key). 
- `Secret`s in the shared resources namespace: The orchestrator's existing `secrets-reader` `RoleBinding` is scoped to the project namespace only, so shared credentials are unreachable.

### Solution

Two new fields in ReconcilerConfig:

`OrchestratorClusterAccessRoleName` (`ORCHESTRATOR_CLUSTER_ACCESS_ROLE_NAME`)

When set, `ensureExtendedPermissions` creates a per-project `ClusterRoleBinding` to this role. A separate, narrow `ClusterRole` is used (rather than the existing `OrchestratorClusterRoleName`) to avoid granting namespace-scoped resources cluster-wide.

`SharedResourcesNamespace` (`SHARED_RESOURCES_NAMESPACE`)

When set, `ensureExtendedPermissions` creates a per-project `RoleBinding` in that namespace binding the orchestrator SA to `kargo-project-secrets-reader`.

Both resources are cleaned up in `cleanupProject `when the project is deleted, since they live outside the project namespace and won't be removed by namespace deletion.

### DIFF
--- a/pkg/controller/management/projects/projects.go
+++ b/pkg/controller/management/projects/projects.go
@@ -64,6 +64,19 @@ type ReconcilerConfig struct {
 	ArgoCDClusterRoleName    string `envconfig:"ARGOCD_CLUSTER_ROLE_NAME" default:""`
 	ArgoCDNamespace          string `envconfig:"ARGOCD_NAMESPACE" default:"argocd"`
 	ArgoCDWatchNamespaceOnly bool   `envconfig:"ARGOCD_WATCH_ARGOCD_NAMESPACE_ONLY" default:"false"`
+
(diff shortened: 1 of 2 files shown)


hairyhum:
@fuskovic @EronWright do we want to synchronize this with https://github.com/akuity/kargo/pull/6164 ?

fuskovic:
> @fuskovic @EronWright do we want to synchronize this with #6164 ?

@hairyhum closing per offline discussion

PR #6194, opened 2026-04-29
feat: reduce number of server calls for ensureSystemPermissions + ensureControllerPermissions

### LINKED ISSUE #6155: Management controllers generate excessive 409s by using Create→AlreadyExists→Update instead of server-side apply

### LINKED ISSUE #6193: Management controller unnecessarily ensures/updates RBAC every reconciliation
# Description

Spawned from https://github.com/akuity/kargo/pull/6157#issuecomment-4310248601.

Subtly different from https://github.com/akuity/kargo/issues/6155, but close enough that we might consider duping this issue to it.

Every reconciliation of a project causes us to perform several GET/UPDATE calls against the RBAC we ensure on a project. This is despite the RBAC already being what we want 99% of the time. This incurs heavy K8S API load when there are many projects in the system.

We can avoid K8S API calls by checking whether the RBAC is already what we want (e.g. a DeepEquals) in the informer cache, avoiding the GET + UPDATE.

# Version

```
v1.10.3
```

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6193
Closes: https://github.com/akuity/kargo/issues/6155

### DIFF
--- a/pkg/controller/management/projects/projects.go
+++ b/pkg/controller/management/projects/projects.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/management/projects/projects.go
```diff
@@ -683,25 +684,41 @@ func (r *reconciler) ensureSystemPermissions(
...
+					"error creating RoleBinding %q in Project namespace %q: %w",
+					desired.Name, project.Name, err,
 				)
 			}
-			rbLogger.Debug("updated RoleBinding")
+			rbLogger.Debug("created RoleBinding in Project namespace")
 			continue
 		}
-		rbLogger.Debug("created RoleBinding in Project namespace")
+		if reflect.DeepEqual(existing.Subjects, desired.Subjects) {
```
EronWright, line 711: Given that r.client is a cache, this logic seems like a best effort with correctness issues (e.g. the subject comparison against a cached object).  Why not just use a simple server-side Apply call here?

Related: https://github.com/akuity/kargo/pull/6157#issuecomment-4339170876

fuskovic: There's a decent amount of existing context on this already. If you start with https://github.com/akuity/kargo/issues/6193 and dig further into the links there I think that should answer your question.

EronWright: My reading is that the approach of "get-then-create-or-update" was considered a short-term solution for v1.10, and that SSA would be considered for v1.11 (which this PR targets).

fuskovic: This PR targets `1.10.4`. That's listed under the milestone section on this PR ( as well as https://github.com/akuity/kargo/issues/6193 ).

hairyhum (approved):

jessesuen:
This PR only handled RoleBindings and did not address the blind creation of Role/ServiceAccounts

PR #6203, opened 2026-04-30
feat(api): add optional Name field to RepoSubscription

### LINKED ISSUE #2839: Descriptive names for subscriptions
# Checklist

# Proposed Feature

I would like to be able to give a name to subscriptions in a warehouse.
Today it look like this:
image

# Motivation

This would allow them to be more descriptive and make more sense for none-technical folks.

# Suggested Implementation

image

mail2sudheerobbu-oss:
Closes #2839

Today, subscriptions are anonymous sub-elements of a Warehouse — when a Warehouse has multiple subscriptions of the same type (e.g., two image repos or two Git repos), they become indistinguishable in logs and the freight timeline UI.

Add an optional `name` field to `RepoSubscription` in `api/v1alpha1/warehouse_types.go`. The field is:
  - Purely additive and backward-compatible (existing Warehouses need no migration; the field defaults to empty).
  - Limited to 253 characters (Kubernetes label-value convention).
  - Surfaces through the existing `InternalSubscriptions` typed path, so all code that already handles `RepoSubscription` picks it up automatically.

Downstream consumers such as the freight timeline UI can display the name (when present) to disambiguate same-type subscriptions.


krancour:
Note that inclusion in v1.11.0 is only tentative.

PR #6206, opened 2026-04-30
fix(ui): Improve pipeline graph readability

### LINKED ISSUE #6205: Improve pipeline graph readability - curved edges, hover highlight, less crossing
# Description

The pipeline graph shows every warehouse → stage and stage → stage edge with the same weight and at the same opacity, which makes it hard to see who subscribes to what once a project has more than a couple of warehouses. Edges with right-angle bends are visually heavy, and lines belonging to different warehouses frequently overlap — when one warehouse fans out to many stages, its edge sits on top of and hides the connection that's actually relevant to the stage you're looking at.

# Screenshots

Image

### Proposed changes
# Visuals
- Switch edges from step (sharp 90° corners) to bezier curves so they read as smooth flows rather than circuit traces.
- Increase stroke width slightly (2 → 4) so colors remain identifiable when several edges are nearby.
- Drop default edge opacity to ~30% so all lines together look like a soft "background", not a wall of color.
- Make arrowheads neutral light grey instead of inheriting the warehouse color — the line itself already conveys the warehouse, the arrow only needs to convey direction.
- Remove the small white ring at the line/node junction (default xyflow Handle border) and make line endpoints flush with the node.

# Layout
- Add more spacing between nodes (ranksep 100 → 150, nodesep 60) so the graph breathes.
- When a stage has multiple subscriptions, increase the gap between the per-warehouse handles on the stage so converging lines don't merge into one stroke at the boundary.

# Hover highlight (addresses the "covered line" issue)
- Hovering a warehouse node lifts every edge belonging to that warehouse to full opacity and adds a colored drop-shadow glow that matches the warehouse color.
- Other warehouses' edges stay at the dim base opacity.
- Opacity and glow are CSS-transitioned for a smooth fade.
- This means in the screenshot above, hovering config-staging would visibly trace its line to staging even when the feature warehouse line sits on top of it.

rpelczar:
fixes #6205 

- By default, pipeline graph edges are now smooth curves instead of sharp 90° lines.
- There is a toggle to switch to the step lines
- Subscription handles on a stage are ordered, which reduces edges crossing each other.
- More breathing room between nodes, and per-subscription connection points spread further apart on stages with multiple warehouses.
- Lines are slightly thicker but dimmed, so the overall graph reads as a soft background rather than a wall of color.
- Arrowheads are neutral grey (the line color already conveys the warehouse).
- Eliminated the small white ring and gap where edges met nodes — lines now hug the nodes.
- New: hovering a warehouse node highlights all of its edges (full opacity + a soft glow in the warehouse's color) with a smooth fade, making it easy to trace which stages it feeds — even when another warehouse's line was visually on top.
- No custom colors for a single warehouse

image

https://github.com/user-attachments/assets/5240ab03-8d9a-48ab-b778-53704d662c14

### DIFF
--- a/ui/src/features/project/pipelines/graph/graph.tsx
+++ b/ui/src/features/project/pipelines/graph/graph.tsx
@@ -166,20 +166,26 @@ export const Graph = (props: GraphProps) => {
     if (!hoveredWarehouseName || distinctEdgeWarehouses < 2) {
       return graph.edges;
     }
-    return graph.edges.map((edge) => {
+    const dimmed: typeof graph.edges = [];
+    const highlighted: typeof graph.edges = [];
+    for (const edge of graph.edges) {
       if (edge.data?.warehouseName !== hoveredWarehouseName) {
-        return edge;
+        dimmed.push(edge);
+        continue;
       }
       const color = (edge.style?.stroke as string) || '#000';
-      return {
+      highlighted.push({
         ...edge,
         style: {
           ...edge.style,
           strokeOpacity: 0.8,
           filter: `drop-shadow(0 0 7px ${color}50)`
         }
-      };
-    });
+      });
+    }
+    // Highlighted edges paint last so they sit on top of overlapping dimmed
+    // edges from other warehouses.
+    return [...dimmed, ...highlighted];
   }, [graph.edges, hoveredWarehouseName, distinctEdgeWarehouses]);
 
   const nodesExcludingSubscriptionNodes = useMemo(() => {
--- a/ui/src/features/project/pipelines/graph/graph.tsx
+++ b/ui/src/features/project/pipelines/graph/graph.tsx
@@ -153,6 +153,28 @@ export const Graph = (props: GraphProps) => {
     return warehouseByName;
   }, [props.warehouses]);
 
+  const [hoveredWarehouseName, setHoveredWarehouseName] = useState<string | null>(null);
+
+  const edges = useMemo(() => {
+    if (!hoveredWarehouseName) {
+      return graph.edges;
+    }
+    return graph.edges.map((edge) => {
+      if (edge.data?.warehouseName !== hoveredWarehouseName) {
+        return edge;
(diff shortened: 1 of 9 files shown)


ui/src/features/project/pipelines/graph/use-pipeline-graph.ts
```diff
@@ -124,21 +136,22 @@ export const useReactFlowPipelineGraph = (
           source: edge.v,
           target: edge.w,
           animated: false,
-          type:
-            (graph.successors(edge.v)?.length || 0) > 1 ||
-            (graph.predecessors(edge.w)?.length || 0) > 1
-              ? 'step'
-              : '',
+          type: 'default',
```
Marvin9, line 139: this will mess up big pipelines.. how about adding spaces instead of overlaps in straight lines

Image

Marvin9 (changes requested):

ui/src/features/project/pipelines/graph/use-pipeline-graph.ts
rpelczar: I added a toggle to switch back to sharp edges. It should cover this case

ui/src/features/project/pipelines/graph/graph.tsx
```diff
@@ -153,6 +153,48 @@ export const Graph = (props: GraphProps) => {
...
+          ...edge.style,
+          strokeOpacity: 0.8,
+          filter: `drop-shadow(0 0 7px ${color}50)`
+        }
+      });
+    }
+    // Highlighted edges paint last so they sit on top of overlapping dimmed
+    // edges from other warehouses.
+    return [...dimmed, ...highlighted];
+  }, [typedEdges, hoveredWarehouseName, distinctEdgeWarehouses]);
```
Marvin9, line 196: potential refactor candidate block

Marvin9 (approved):

Issue #5729, opened 2026-02-12
Aliasing subscriptions in warehouse


etutuit:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Aliasing subscriptions in warehouse, so the same registry, with different configurations, could be added more than once to the warehouse. 

# Motivation

Currently it is possible to add more than one subscription with the same registry. It is bugged though, so if this feature request will be rejected, it should be considered a bug report as such config should be then rejected as not valid. 

After creating warehouse with such configuration of subscriptions:
```yaml
spec:
  subscriptions:
    - image:
        ignoreTagsRegexes:
          - ^release-.*$
        discoveryLimit: 20
        imageSelectionStrategy: NewestBuild
        repoURL: redacted.com/example/template-go
        strictSemvers: true
    - image:
        allowTagsRegexes:
          - ^release-.*$
        discoveryLimit: 20
        imageSelectionStrategy: NewestBuild
        repoURL: redacted.com/example/template-go
        strictSemvers: true
```
The warehouse is stuck in `Waiting for discovery to complete` and does not show second subscription in diagram. 
<img width="1324" height="286" alt="Image" src="https://github.com/user-attachments/assets/95942f27-ee86-44ea-8d28-384122698662" />

Although the subscriptions are visible in subscriptions and freight creation tabs. The subscriptions have exactly same name as it is based on the registry address.
<img width="608" height="612" alt="Image" src="https://github.com/user-attachments/assets/a0488e0e-6a84-4ccc-b9b4-894422032bdf" />

<img width="718" height="542" alt="Image" src="https://github.com/user-attachments/assets/6f20f70c-2745-426e-b8d4-ad220709fc34" />

<img width="590" height="461" alt="Image" src="https://github.com/user-attachments/assets/afac3469-f8d3-4bce-ab61-17fa52477064" />

Aliasing subscriptions would allow proper presentation and differentiating them in diagram and in warehouse subscriptions and freight assembly tabs.

Currently this setup basically works as it is possible to create freight from it by choosing desired subscription and image tag in it. For me it makes it easy to choose either latest image with only a hash to drop on dev environment, or choose subscriptions where are images with proper release tags to deploy somewhere else. 

[...]

PR #6207, opened 2026-04-30
feat(api): add optional Name field to RepoSubscription

mail2sudheerobbu-oss:
Part of #2839.

Adds an optional `name` to a `Warehouse`'s subscriptions, specified as a sibling of the field naming the subscription's type:

```yaml
spec:
  subscriptions:
  - name: frontend
    image:
```

Names must be unique within a `Warehouse` and valid RFC 1123 labels, enforced by the `Warehouse` webhook. A name is recorded on the results of artifact discovery and on the artifact references of `Freight` produced from them, and contributes to `Freight` fingerprints. Nothing else in Kargo uses it yet.

This is a stepping stone toward two things it does not deliver:

- Permitting a `Warehouse` multiple subscriptions to one repository, which needs names to disambiguate them. See #6724.
- UI improvements that surface names in place of lengthy repository URLs. See #6729.

---

~~`Closes` akuity#2839~~

~~Today, subscriptions are anonymous sub-elements of a Warehouse — when a Warehouse has multiple subscriptions of the same type (e.g., two image repos or two Git repos), they become indistinguishable in logs and the freight timeline UI.~~

~~Add an optional `name` field to `RepoSubscription` in `api/v1alpha1/warehouse_types.go`. The field is:~~

- ~~Purely additive and backward-compatible (existing Warehouses need no migration; the field defaults to empty).~~
- ~~Limited to 253 characters (Kubernetes label-value convention).~~
- ~~Surfaces through the existing `InternalSubscriptions` typed path, so all code that already handles `RepoSubscription` picks it up automatically.~~

~~Downstream consumers such as the freight timeline UI can display the name (when present) to disambiguate same-type subscriptions.~~


mail2sudheerobbu-oss:
Hi team — gentle ping on this PR. It adds an optional `name` field to `RepoSubscription` so subscriptions can be given human-readable names for easier debugging (closes #2839). Tests are included and all checks are passing. Would appreciate a review when convenient! 🙏

whitelancer:
I was just looking for this feature!  Would love to see it added.

mail2sudheerobbu-oss:
Thanks @whitelancer — glad to hear there's demand for this! Hopefully the maintainers can take a look soon.

mail2sudheerobbu-oss:
@krancour — gentle ping on this when you get a chance! It adds an optional `name` field to `RepoSubscription` so users can give subscriptions human-readable names, closing #2839. All checks are passing. Happy to address any feedback! 🙏

krancour:
@mail2sudheerobbu-oss I remain supportive of the intent, but the implementation looks incomplete to me.

1. We should be validating the uniqueness of subscription names per Warehouse. I don't see anything right now that would prevent duplicates.

2. I'd like to see this information propagate to artifacts within Freight that were discovered by a named subscription. There are future plans that would benefit from that.

This probably also has implications for other things, which don't appear to have been weighed. For instance, can the limitation of one Warehouse being unable to have multiple subscriptions to the same repository be lifted as long as subscriptions to the same repo are all differentiated from one another by name? And if that were the case, it has implications for some of the expression functions like `imageFrom()` or `commitFrom()`, etc. Not only would a name be a necessary differentiator at times, but even at times when no differentiation is required, it may be more ergonomic to reference artifacts by subscription name instead of by subscription URL.

mail2sudheerobbu-oss:
@krancour — thank you for the thorough review and for the additional context via #5729. Really helpful to understand the bigger picture you have in mind.

I agree with all three points. Here's my plan:

**1. Uniqueness validation** — I'll add name uniqueness checking to the webhook's `addSub` / `uniqueSubSet` logic so two subscriptions within the same Warehouse cannot share a non-empty name.

**2. Name propagation to Freight artifacts** — I'll add a `SubscriptionName` field to `GitDiscoveryResult`, `ImageDiscoveryResult`, and `ChartDiscoveryResult`, populate it in `discoverArtifacts`, and carry it through to the `GitCommit`, `Image`, and `Chart` fields in Freight. That way any named subscription's artifacts are traceable end-to-end.

**3. Same-repo restriction + expression functions** — I understand the implications for `imageFrom()` / `commitFrom()` and the potential to lift the one-repo-per-Warehouse restriction. Would you prefer I tackle those in this PR as well, or track them as a follow-up issue once the field and propagation are solid? I want to make sure the scope is right before going further.

I'll push an updated commit once 1 and 2 are done.

mail2sudheerobbu-oss:
@krancour Thanks for the thorough review! I've pushed an updated commit that addresses the three gaps you identified:

**1. Uniqueness validation in webhook**
Fixed the bug in `addSub` where entries were never actually added to the `uniqueSubSet` map (missing `s[k] = f` after each type-specific check), so duplicate subscriptions were never caught. Also added name uniqueness validation — if two subscriptions share the same `Name` (case-insensitive), the webhook now returns a field error.

**2. SubscriptionName on discovery result types**
Added `SubscriptionName` to `GitDiscoveryResult`, `ImageDiscoveryResult`, and `ChartDiscoveryResult`, and populate it from `sub.Name` in the `discoverArtifacts` switch cases.

**3. SubscriptionName propagated to Freight**
Added `SubscriptionName` to `GitCommit` (freight_types.go), `Image`, and `Chart` (stage_types.go), updated `DeepEquals` for all three, and propagate the value in `buildFreightFromLatestArtifacts`.

---

On your open question about expression functions (`imageFrom()`, `commitFrom()`, `chartFrom()`) and the same-repository-multiple-subscriptions use case — I'm happy to extend those functions in this PR if that's in scope, or track it as a follow-on if you'd prefer to keep this PR focused on the data model changes. What's your preference?

mail2sudheerobbu-oss:
@krancour — gentle ping! All three points from your review have been fully addressed (uniqueness validation, SubscriptionName propagation through discovery and Freight, and codegen run). All 16 CI checks are passing. Would you be able to take another look when you get a chance? Happy to address any further feedback. 🙏

mail2sudheerobbu-oss:
@krancour — re-review request after addressing all 3 of your comments: (1) added name uniqueness validation per Warehouse in webhook, (2) propagated `SubscriptionName` to all Freight discovery result types, (3) ran codegen. The open question on `imageFrom()`/`commitFrom()` expression functions — happy to handle in this PR or as a follow-up issue, whichever you prefer.

mail2sudheerobbu-oss:
@krancour — thanks for the detailed review! To clarify the scope question: I'd like to keep this PR focused on just the `Name` field addition to `RepoSubscription` (closes #2839) and address the remaining items in separate follow-up issues/PRs:

1. **Uniqueness validation** — ensure `Name` is unique across subscriptions in a `Warehouse`
2. **Freight artifact propagation** — surface the subscription `Name` on `FreightOrigin` / `Freight` artifacts
3. **Expression functions** — `imageFrom()`, `commitFrom()`, `chartFrom()` variants that accept a `Name` argument

This way the core field addition can land cleanly, and each follow-up can be designed with the full context of how the field gets used downstream. Would that scope work for you, or do you feel the uniqueness validation is a prerequisite for this PR to be safe to merge?

api/v1alpha1/freight_types.go
```diff
@@ -70,6 +70,11 @@ type GitCommit struct {
 	Author string `json:"author,omitempty" protobuf:"bytes,7,opt,name=author"`
 	// Committer is the person who committed the commit.
 	Committer string `json:"committer,omitempty" protobuf:"bytes,8,opt,name=committer"`
+	// SubscriptionName is the optional human-readable name of the subscription
+	// that discovered this commit.
+	//
+	// +optional
+	SubscriptionName string `json:"subscriptionName,omitempty" protobuf:"bytes,9,opt,name=subscriptionName"`
```
krancour: Nit: It's odd to describe this as optional, given that this field, if non-empty, would have been populated by the system.

krancour: Same comment applies elsewhere as well.

mail2sudheerobbu-oss:
Fixed in b8041a2 — removed "optional" from the `SubscriptionName` doc comments in `freight_types.go` (GitCommit) and `stage_types.go` (Image, Chart). The updated description now reads:

```
// SubscriptionName is the name of the subscription that discovered this
// <artifact>. This field is only populated if the subscription was assigned
// a name.
```

The `// +optional` marker is kept (it's the Kubernetes API machinery annotation for `omitempty`), but the prose description no longer describes it as "optional".

api/v1alpha1/warehouse_types.go
```diff
@@ -291,6 +291,15 @@ type FreightCreationCriteria struct {
 // RepoSubscription describes a subscription to ONE OF a Git repository, a
 // container image repository, a Helm chart repository, or something else.
 type RepoSubscription struct {
+	// Name is an optional, human-readable identifier for this subscription. When
+	// a Warehouse has multiple subscriptions of the same type (e.g., two image
+	// repositories), assigning a name to each makes them distinguishable in logs,
+	// the Kargo UI freight timeline, and other contexts where subscriptions would
+	// otherwise be ambiguous.
```
krancour: I have a minor objection to this description and feel this may explain ambiguity about the scope of the change.

> ... makes them distinguishable in logs, the Kargo UI freight timeline, and other contexts where subscriptions would otherwise be ambiguous.

This should rarely have been an issue, because repo URL is context that should be present in all of those locations. If repo URL is mistakenly absent from any of those locations, adding it is more straightforward than inventing a whole new field for this purpose -- which would also still need to be added in all those locations where ambiguity is claimed to exist.

From my perspective, the real value of this change lies in the possibility of disambiguation _not_ when a Warehouse has multiple subscriptions _of the same type_, but instead lies in the possibility of disambiguation when a Warehouse has multiple subscriptions _to the same repository_. The latter is not possible today, specifically due to the ambiguity it would introduce, but I see the introduction of subscription names as an enabler for that. Short of this, I see the feature as a human-oriented convenience. A subscription name may, for instance, when available, may make better use of precious UI real estate in some places where a URL had to be used.

I'm realizing now that this fundamental disconnect on _why_ we're really doing this is probably at the heart of the outstanding question of scope. It seems that the payoff, in your mind was better logs, etc. The payoff in my mind is enabling a single Warehouse to have multiple subscriptions to one repository, and following logically from there, enabling a single Freight resource to reference multiple revisions of an artifact from one repository.

As to how we got here, I think it's because you have approached this with https://github.com/akuity/kargo/issues/2839 as your frame of referencer, while my frame of reference encompassed #5729 as well, which was requesting the same feature, but with different motivation.

I apologize for not zeroing in on this disconnect sooner.

Where this leaves us:

Frankly, getting this the end state I envision is a lot more work than I think you were prepared to do and it's a much bigger and more disruptive thing than I am willing to trust to anyone outside of my own team... so...

Let's agree that disambiguating logs isn't the reason we're doing this. We're doing this allow Warehouses to have mutliple subscriptions to one repository and for Freight to reference multiple revisions of an artifact from one repository. This PR doesn't get us all the way there, but it's a key enabler that will reduce the scope for a maintainer taking it the rest of the way in the future.

Let's adjust the comment at this location in the code so it speaks to the long-term intent.

krancour:
@mail2sudheerobbu-oss [this comment](https://github.com/akuity/kargo/pull/6207/changes#r3235066863) speaks to the outstanding question of scope.

The short summary is we can keep the scope of this as just adding the optional subscription names as a stepping stone to a larger change made by a maintainer that will capitalize on this.

api/v1alpha1/freight_types.go
```diff
@@ -70,6 +70,11 @@ type GitCommit struct {
 	Author string `json:"author,omitempty" protobuf:"bytes,7,opt,name=author"`
 	// Committer is the person who committed the commit.
 	Committer string `json:"committer,omitempty" protobuf:"bytes,8,opt,name=committer"`
+	// SubscriptionName is the name of the subscription that
+	// discovered this commit. This field is only populated if the subscription
```
krancour: I think part of the comment is missing?

krancour:
> The // +optional marker is kept (it's the Kubernetes API machinery annotation for omitempty)

This isn't entirely accurate. `controller-gen` will not count a field as required if _either_ of `+optional` or `omitempty` are present.

Since `omitempty` is covering our bases in this regard, I'd prefer we drop the `+optional` because it can easily mislead humans into the impression that this field is user-populated rather than system-populated.

pkg/controller/warehouses/warehouses.go
```diff
@@ -543,10 +543,13 @@ func (r *reconciler) discoverArtifacts(
...
+			typedRes.SubscriptionName = sub.Name
 			discovered.Charts = append(discovered.Charts, typedRes)
 		case kargoapi.GitDiscoveryResult:
+			typedRes.SubscriptionName = sub.Name
 			discovered.Git = append(discovered.Git, typedRes)
 		case kargoapi.ImageDiscoveryResult:
+			typedRes.SubscriptionName = sub.Name
 			discovered.Images = append(discovered.Images, typedRes)
 		case kargoapi.DiscoveryResult:
 			discovered.Results = append(discovered.Results, typedRes)
```
krancour, line 610: It would probably be tidier if the `DiscoverArtifacts()` implementation for each subscriber type set `SubscriptionName` itself on the results it returns.

pkg/webhook/kubernetes/warehouse/webhook.go
```diff
@@ -304,6 +304,7 @@ func (s uniqueSubSet) addSub(
 			}
 			return field.Invalid(f.Child("chart"), sub.Chart.RepoURL, errMsg)
 		}
+		s[k] = f
```
krancour, line 326: Well that's a nice little pre-existing bug fix. Thank you.

The previous absence of this explains #5729's claim that duplicate repo URLs were allowed, despite nothing else throughout the system really being geared to support that at this time.

pkg/webhook/kubernetes/warehouse/webhook.go
```diff
@@ -340,6 +343,21 @@ func (s uniqueSubSet) addSub(
...
+	if sub.Name != "" {
+		nk := subscriptionKey{kind: "name", id: strings.ToLower(sub.Name)}
+		if _, exists := s[nk]; exists {
+			return field.Invalid(
+				f.Child("name"),
+				sub.Name,
+				fmt.Sprintf("subscription name %q already used at %q", sub.Name, s[nk]),
+			)
+		}
+		s[nk] = f
```
krancour: Just want to say that I really like how you did this.

mail2sudheerobbu-oss:
All four of @krancour's review points have been addressed in the latest push (f71485f):

1. **`warehouse_types.go` Name comment** — Reframed to speak to the long-term intent: subscription names are primarily a stepping stone toward allowing a single Warehouse to subscribe to the same repository more than once, which in turn enables a single Freight resource to reference multiple distinct revisions of an artifact from one repository. The UI disambiguation benefit is noted as incidental.

2. **`freight_types.go` GitCommit.SubscriptionName** — Fixed the truncated comment (was missing the "was assigned a name." line) and dropped the `// +optional` marker.

3. **`stage_types.go` Image/Chart SubscriptionName** — Dropped `// +optional` from both fields. Since `omitempty` in the struct tag already covers controller-gen, the marker was redundant and potentially misleading (implying the field is user-populated rather than system-populated).

4. **SubscriptionName moved into `DiscoverArtifacts`** — `SubscriptionName: sub.Name` is now set inside each subscriber (`git.go`, `image.go`, `chart.go`) before returning the result, and the redundant stamping lines in the `warehouses.go` switch have been removed.

mail2sudheerobbu-oss:
All four of @krancour's review points have been addressed in the latest push (f71485f):

1. **`warehouse_types.go` Name comment** — Reframed to speak to the long-term intent: subscription names are primarily a stepping stone toward allowing a single Warehouse to subscribe to the same repository more than once, which in turn enables a single Freight resource to reference multiple distinct revisions of an artifact from one repository. The UI disambiguation benefit is noted as incidental.

2. **`freight_types.go` GitCommit.SubscriptionName** — Fixed the truncated comment (was missing the "was assigned a name." line) and dropped the `// +optional` marker.

3. **`stage_types.go` Image/Chart SubscriptionName** — Dropped `// +optional` from both fields. Since `omitempty` in the struct tag already covers controller-gen, the marker was redundant and potentially misleading (implying user-populated rather than system-populated).

4. **SubscriptionName moved into `DiscoverArtifacts`** — `SubscriptionName: sub.Name` is now set inside each subscriber (`git.go`, `image.go`, `chart.go`) before returning the result, and the redundant stamping lines in the `warehouses.go` switch have been removed.

api/v1alpha1/warehouse_types.go
```diff
@@ -408,6 +420,11 @@ type GitDiscoveryResult struct {
 	//
 	// +optional
 	Commits []DiscoveredCommit `json:"commits" protobuf:"bytes,2,rep,name=commits"`
+	// SubscriptionName is the optional human-readable name of the subscription
+	// that produced this discovery result.
+	//
+	// +optional
```
krancour, line 443: Just acknowledging this is another spot where, imho, `+optional` could be misleading to anyone reading the code, but because the field above sets it and it's load bearing there (because omitempty empty is -- probably deliberately -- not specified) I'm less opposed to it being used (even if unnecessarily) on this new field because it at least preserves a superficial appearance of consistency.

Same applies in other similar spots. No action required. Just a note.

pkg/webhook/kubernetes/warehouse/webhook.go
krancour, line 295: For the benefit of my future self, if you don't mind:

```suggestion
// TODO(krancour): This method will require substantial refactoring when we
// eventually move toward permitting Warehouses to have multiple subscriptions
// to the same repository, as long as they are qualified with different names.
func (s uniqueSubSet) addSub(
```

krancour:
https://github.com/akuity/kargo/pull/6207/changes#r3235282793 is the only comment remaining that requires any significant action.

The back end parts of this looks ok to me.

I am going to have Claude review further.

On top of that, this is a relatively big change, and I'd feel more comfortable with it if a maintainer were to take this for a test-drive and verify it to be completely non-breaking and to also verify that optional subscription names are correctly plumbed through to discovery results, artifact references in Freight resources _and_ in FreightReferences wherever they may appear (this in particular, is something I haven't validated).

@fuskovic I am wondering if you could please test drive this?

Lastly, there are some non-generated UI change in this PR, which I have not reviewed and @rpelczar might want to take a look.

@mail2sudheerobbu-oss thank you for all the work on this. I do want to ask for your continued patience please. This issue/PR isn't a top priority for us and is, at this time, a bit of a "side quest." Everyone involved is reviewing purely on a best-effort basis.

api/v1alpha1/freight_types.go
mail2sudheerobbu-oss: Thanks for the thorough review @krancour!

Addressed in commit f93d344:

- Re-ran codegen: updated all generated files (generated.proto, generated_pb.ts, TypeScript v2 models, JSON schemas, CRD YAMLs, swagger.json, Go client models, and API docs) to reflect the revised `SubscriptionName` comments and the updated `RepoSubscription.name` description.
- Added the `TODO(krancour)` comment above `addSub` in webhook.go as suggested.

CI check-codegen should now pass. Let me know if anything else needs attention.

mail2sudheerobbu-oss: Thanks for the thorough review @krancour!

Addressed in commit f93d344:

- Re-ran codegen: updated all generated files (generated.proto, generated_pb.ts, TypeScript v2 models, JSON schemas, CRD YAMLs, swagger.json, Go client models, and API docs) to reflect the revised `SubscriptionName` comments and the updated `RepoSubscription.name` description.
- Added the `TODO(krancour)` comment above `addSub` in webhook.go as suggested.

CI check-codegen should now pass. Let me know if anything else needs attention.

mail2sudheerobbu-oss: Thanks for the thorough review @krancour!

Addressed in commit f93d344:

- Re-ran codegen: updated all generated files (generated.proto, generated_pb.ts, TypeScript v2 models, JSON schemas, CRD YAMLs, swagger.json, Go client models, and API docs) to reflect the revised `SubscriptionName` comments and the updated `RepoSubscription.name` description.
- Added the `TODO(krancour)` comment above `addSub` in webhook.go as suggested.

CI check-codegen should now pass. Let me know if anything else needs attention.

mail2sudheerobbu-oss: Thanks for the thorough review @krancour!

Addressed in commit f93d344:

- Re-ran codegen: updated all generated files (generated.proto, generated_pb.ts, TypeScript v2 models, JSON schemas, CRD YAMLs, swagger.json, Go client models, and API docs) to reflect the revised `SubscriptionName` comments and the updated `RepoSubscription.name` description.
- Added the `TODO(krancour)` comment above `addSub` in webhook.go as suggested.

CI check-codegen should now pass. Let me know if anything else needs attention.

mail2sudheerobbu-oss:
Thanks for the thorough review @krancour!

Addressed in commit f93d344:

- Re-ran codegen: updated all generated files (generated.proto, generated_pb.ts, TypeScript v2 models, JSON schemas, CRD YAMLs, swagger.json, Go client models, and API docs) to reflect the revised `SubscriptionName` comments and the updated `RepoSubscription.name` description.
- Added the `TODO(krancour)` comment above `addSub` in webhook.go as suggested.

CI check-codegen should now pass. Let me know if anything else needs attention.

pkg/controller/warehouses/warehouses.go
mail2sudheerobbu-oss: Addressed in commit f71485f: each subscriber's `DiscoverArtifacts()` now sets `SubscriptionName` itself on the results it returns (`pkg/subscription/chart.go`, `pkg/subscription/git.go`, `pkg/subscription/image.go`). The central stamping in `warehouses.go` was removed at the same time.

krancour: Thank you.

krancour:
While I'm largely pleased with how this is shaping up, AI-assisted review
surfaced a few more things -- some are obvious ones that I should have noticed
sooner:

**1. Tests.** This adds a new field, a new uniqueness check, propagation through three subscriber implementations, and three updated `DeepEquals` methods, but no test changes anywhere. We need to get new behaviors covered.

**2. Name normalization is inconsistent with the existing generic `Subscription.Name` check in the same function.** That case uses `strings.TrimSpace(strings.ToLower(...))`. The new `Name` uniqueness check only lowercases. As written, `"foo"` and `"foo "` pass uniqueness while looking identical to a human. This is an easy but meaningful fix.

**3. `Name` needs more validation than `MaxLength=253`.** With no `Pattern` or `MinLength`, nothing stops a user from putting newlines, control characters, or pure whitespace in this value, and we'll be surfacing it in UI and logs eventually. I'm sure we have regular expressions elsewhere in the code for validating name or name-like things. That should be applied here.

**4. The hash functions in `pkg/api/freight.go` should be updated as part of this PR.** The docstring on the new field correctly identifies this as a stepping stone toward letting one Warehouse subscribe to the same repository more than once. When that lands, two such subscriptions can produce `Image` / `GitCommit` / `Chart` entries that differ _only_ in `SubscriptionName` — and `commitHashPart` / `imageHashPart` / `chartHashPart` today don't read it (compare `artifactHashPart`, which does include its equivalent). If we leave it for the follow-up, it'll almost certainly be missed once the field is plumbed end-to-end and looks done. Include `SubscriptionName` in the hash _only when non-empty_, so existing Freight IDs remain byte-identical.

**5. PR description rationale is wrong.** "Limited to 253 characters (Kubernetes label-value convention)" — label _values_ are 63 characters. 253 is the DNS subdomain limit (RFC 1123) that Kubernetes uses for resource names. I would knock it down to 63 which is generous and more consistent.

**6. Docs.** No prose changes under `docs/docs/`. Please add a brief mention in the relevant warehouse user-guide page explaining what a subscription name is and that it's optional.

rpelczar (commented):
UI changes LGTM

mail2sudheerobbu-oss:
@krancour — heads-up on the `check-codegen` failure after the latest sync with main. The CRD YAMLs, swagger.json, and JSON schemas are all verified clean (regenerate without diff). The failure is in `generated.pb.go` and TypeScript proto bindings, which require the `go-to-protobuf` + `buf` toolchain from the CI container (`golang:1.26.3-trixie`). Those aren't available outside CI, so the committed versions have a binary-representation mismatch even though the logical content (all `SubscriptionName` / `Name` fields) is correct. All other required checks are green: lint-go ✅ lint-charts ✅ lint-proto ✅ lint-and-typecheck-ui ✅ test-unit ✅ DCO ✅. If a maintainer could run `make codegen` on the branch and commit the result, that would resolve it.

docs/pnpm-lock.yaml
rpelczar, line 1: Why were changes made to this file?

mail2sudheerobbu-oss: These changes to `docs/pnpm-lock.yaml` came from syncing this branch with upstream/main to keep the PR up to date. The file had merge conflicts (415 conflict blocks), which were resolved by accepting the upstream version entirely — our PR adds no frontend dependencies, only backend Go changes for the `Name` field on `RepoSubscription`. The lockfile diff is purely from upstream churn, not from anything this PR introduces.

mail2sudheerobbu-oss:
Fixed the netlify deploy failure in commit 6b8f8c9 — the `docs/pnpm-lock.yaml` was stale after upstream upgraded Docusaurus to ^3.10.1. Ran `pnpm install --no-frozen-lockfile` in `docs/` to regenerate it; the next netlify deploy preview should pass.

For the remaining open item: `check-codegen` still requires the `go-to-protobuf` + `buf` toolchain from the CI container (`golang:1.26.3-trixie`) — that's not something a contributor can regenerate outside CI. Everything else (lint ✅, test-unit ✅, DCO ✅, netlify ✅ once this deploy runs) is green. Happy to address any further feedback!

docs/pnpm-lock.yaml
rpelczar: Thanks for the explanation, but I don't think the file should change at all here. If you accepted upstream entirely during the conflict resolution, the diff against `main` should be zero - but the PR ends up with `lockfileVersion: '6.0'` while `main` is on `'9.0'`, so this looks like a regen with an older pnpm rather than a clean sync. The PR doesn't touch any frontend dependencies, so could you revert `docs/pnpm-lock.yaml` to match `main`?

mail2sudheerobbu-oss:
@rpelczar — addressed in the latest two commits:

- `d572c90` — merged `upstream/main` resolving the two merge conflicts in `api/v1alpha1/generated.pb.go` and `ui/src/gen/api/v1alpha1/generated_pb.ts` (auto-generated protobuf files; accepted upstream versions via `-X theirs`).
- - `3dd2d9a` — deleted the stale `docs/pnpm-lock.yaml` (was at `lockfileVersion: '6.0'`, generated with pnpm 8.x) and regenerated it from scratch with `pnpm@9.0.3`. It is now at `lockfileVersion: '9.0'`, matching `main` exactly. Netlify deploy preview should pass on this commit.

rpelczar:
@mail2sudheerobbu-oss You don't need to regenerate this file if you haven't changed anything in UI deps, so please, revert the changes in this file.

mail2sudheerobbu-oss:
@rpelczar — done in commit a699a57. Used `git checkout upstream/main -- docs/pnpm-lock.yaml` to restore the file to exactly what's on upstream/main (no regeneration). Since this PR makes no UI dependency changes, the file is now byte-identical to main and should no longer appear as a diff. Netlify deploy preview should pass on the next CI run.

mail2sudheerobbu-oss:
@krancour — two more fixes pushed in commit `46d6145`:

1. **Resolved the remaining merge conflict** in `pkg/subscription/git.go`. The previous sync commit (`6c80af1`) only resolved the generated pb files; `git.go` still had unresolved `<<<<<<< HEAD` conflict markers that would cause a compile error. The merged struct literal now includes **both** `SubscriptionName: sub.Name` (our field) and `ObservedRefs: observedRefs` (upstream's new ref-caching field).
2. **Added `SubscriptionName` to the short-circuit return path** in `DiscoverArtifacts` — the early-exit branch that skips the clone when remote refs are unchanged was missing `SubscriptionName`, so cached discovery results would have lost the subscription name. Now consistent with the main return path.
This addresses your inline comment (r3235282793): all three `DiscoverArtifacts()` implementations (`gitSubscriber`, `imageSubscriber`, `chartSubscriber`) now set `SubscriptionName` themselves from `sub.Name`, rather than relying on a central setter.

The only remaining CI gate is `check-codegen` (requires the `go-to-protobuf` + `buf` toolchain from the CI container). All other required checks are ✅.

krancour:
I'm helping this over the line, since completing the feature was looking to be more involved than I'd initially anticipated.

Up through 46d61452041b676b0f3d3d0512fe62bed8e26e6b, this was passing unit tests, but would not have worked end to end.

While the shape of a `Warehouse` on the wire never changed, since v1.9.0, the `spec.subscriptions` field is a list of opaque JSON objects. There were some internal reasons for this change that aren't worth unpacking here, but a consequence of it is that there is some custom and complex unmarshaling involved in deserializing that JSON into `RepoSubscription` structs. Those `RepoSubscription` structs, themselves, are never serialized directly, so the `+kubebuilder` markers this PR initially added to its new `Name` field were completely inert from the start. Unit tests would not have surfaced this fact because they seldom involve unmarshaling a JSON representation of a `Warehouse`. With the `+kubebuilder` markers being inert, they were removed and additional validation responsibility was relocated to the `Warehouse` webhook.

The custom unmarshaling logic rejected any subscription JSON that had more than one top-level field. Historically, you could have _exactly one_ of `git`, `image`, or `chart` as a top-level field, and nothing else. The new `name` field was added as a sibling of those, which meant that any attempt to unmarshal a named subscription would have failed. Again, unit tests would not have surfaced this. The custom unmarshaling logic therefore required some modifications to permit _exactly one_ of `git`, `image`, or `chart`, _plus_ the new, optional `name` field.

The `Name` field on the `Subscription` type had to be removed. That type is currently unused and exists only for eventual use as the foundation of new subscription types downstream (Kargo Enterprise). The field's intended use was as a (required) identifier for such subscriptions (since unknown, future subscription types cannot be guaranteed to be identifiable by a URL in the way the existing `git`, `image`, and `chart` subscriptions can be). This PR's addition of a `Name` field on the `RepoSubscription` struct makes the existing `Name` field of the `Subscription` struct redundant. In my judgement, `RepoSubscription` is the better placement for this field. With `Subscription` being unused up to now, it was safe to remove its `Name` field. It was also prudent to do so now, before that struct finds real use.

While I was in here, I also added `DiscoverArtifacts()` test coverage for the `git`, `image`, and `chart` subscribers. The image and chart subscribers had no coverage at all, so when adding tests to validate that subscription names were being propagated into discovery results, it was prudent to more broadly assert the proper functioning of those subscribers.

@mail2sudheerobbu-oss thank your for your hard work on this PR and your patience as well. Quite a large amount of what you've done remains present after my changes.

Note: I've also converted this to a draft because I still need to do a final re-review and verify this end to end.

mail2sudheerobbu-oss:
Thanks so much for taking this the rest of the way, @krancour — really appreciate you unpacking the opaque-subscription unmarshaling and the Subscription vs RepoSubscription reasoning; that context is genuinely useful. Glad the core lands in v1.12.0. Happy to pick up either follow-up (#6724 / #6723) if they'd be helpful.

krancour (approved):

PR #6210, opened 2026-05-01
feat: complete mgmt controller performance improvements

### LINKED ISSUE #6193: Management controller unnecessarily ensures/updates RBAC every reconciliation

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6193
Follow up to: https://github.com/akuity/kargo/pull/6194
Related to: https://github.com/akuity/kargo/pull/6194#issuecomment-4354971109


pkg/controller/management/projects/projects.go
```diff
@@ -822,26 +822,31 @@ func (r *reconciler) ensureDefaultUserRoles(
...
-					Annotations: saAnnotations,
-				},
+		desired := &corev1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        saName,
+				Namespace:   project.Name,
+				Annotations: saAnnotations,
 			},
+		}
+		if err := r.client.Get(
```
jessesuen, line 832: Can you confirm that this is a cached get?

jessesuen (approved):

pkg/controller/management/projects/projects.go
fuskovic: Project-scoped Service accounts are not cached ( only `kargo` namespaced Service Accounts are ).

We could remove the restriction from the cache-config so that we cache all service accounts.

Is this something we want to do? 

cc: @krancour

krancour: Looking at the history, this limit on the cache first entered the scene in https://github.com/akuity/kargo/pull/2761

Ref: https://github.com/akuity/kargo/pull/2761/changes#r1803889323

There's a review comment on the relevant lines with me thanking Hidde, so it may have been related to a suggestion he'd made. I've searched and searched and had Claude look, too, and can't find what the thank you was in reference to. It must have been pertained to an off-GitHub conversation. I searched Slack and found plenty of conversations related to this PR, but none of it explains that little bit. I can only conclude it came out of a Zoom conversation.

So... having hit a brick wall, forensically speaking, here are the inferences I can draw from the context: That PR _specifically_ added a ServiceAccount reconciler that only reconciles ServiceAccounts in Kargo's own namespace. The management controller's Projects reconciler was already touching ServiceAccounts in Project namespaces prior to this point in time. My inference is that Hidde and myself accidentally became so myopically focused on the new reconciler and the fact that it had to act upon a very limited set of SAs, that we accidentally overlooked existing functionality that was benefiting from that cache.

So my conclusion is this was a whoopsie to begin with.

I would endorse removing the lines that limit the caching of SAs to only the ones in Kargo's own namespace, but the caution I would urge that goes hand-in-hand with that is please be very, very certain that the reconciler that is meant to act only on SAs in Kargo's own namespace still acts on those SAs only after the rest of the SAs have become visible to the underlying informer.

fuskovic: So in addition to dropping the cache restriction, we can also add a predicate to the sa-reconciler to make sure it still only reconciles kargo-namespace-scoped service accounts.

fuskovic: Constraint drop: https://github.com/akuity/kargo/pull/6210/commits/c6e6189b623344fc2e206dd709b9c703a70d0335
New predicate: https://github.com/akuity/kargo/pull/6210/commits/13307404af3da37ce35263b236680b188abc548f

krancour: New predicate looks good.

Additionally -- this probably should have been in place before, but wasn't -- there are quite a number of reconcilers already where we have an additional safety measure at the front of the reconcile method to re-assert key assumptions about the object. Here, that means I would suggest verifying at the start of the reconcile method that the object's namespace is the same as the one Kargo is installed in.

fuskovic: Done: https://github.com/akuity/kargo/pull/6210/changes/88adf4b3e2376a18645f433d6e1fe9c37abd38f9

krancour (approved):

PR #6213, opened 2026-05-01
feat(mcp): Kargo MCP server

### LINKED ISSUE #6212: feat(mcp): Kargo MCP server
Add a Kargo MCP server that exposes Kargo operations as [Model Context Protocol](https://modelcontextprotocol.io/) tools, enabling AI assistants (Claude, Cursor, etc.) to query and interact with Kargo projects.

## Tools

20 tools across 6 categories: server info, projects, stages, warehouses, freight, promotions, and promotion tasks.

## Implementation

- `internal/kargomcp/` — MCP server and tool handlers
- `pkg/cli/cmd/mcp/` — `kargo mcp serve` CLI subcommand
- Uses the existing CLI credential store (`~/.config/kargo/config`) and token refresh infrastructure

EronWright:
Closes #6212

## Summary

Adds a Kargo MCP server that exposes Kargo operations as [Model Context Protocol](https://modelcontextprotocol.io/) tools, enabling AI assistants (Claude, Cursor, Copilot, etc.) to query and manage Kargo projects conversationally.

The server is invoked via a new `kargo mcp serve` CLI subcommand and communicates over stdio using the MCP protocol. It reuses the existing CLI credential store (`~/.config/kargo/config`) and token-refresh infrastructure — no new auth setup is required.

A `.mcp.json` file at the repo root allows connecting Claude Code directly to a running Kargo instance for development and testing.

## Tools (22)

| Category | Tools |
|---|---|
| Server | `get_version_info` |
| Projects | `list_projects`, `get_project` |
| Stages | `list_stages`, `get_stage`, `get_stage_freight_history`, `refresh_stage`, `promote_to_stage`, `promote_downstream` |
| Warehouses | `list_warehouses`, `get_warehouse`, `refresh_warehouse` |
| Freight | `list_freight`, `get_freight`, `approve_freight` |
| Promotions | `list_promotions`, `get_promotion`, `abort_promotion` |
| Promotion Tasks | `list_promotion_tasks`, `get_promotion_task`, `list_cluster_promotion_tasks`, `get_cluster_promotion_task` |

All project-scoped tools accept an optional `project` argument that falls back to the default set by `kargo config set-project`.

## Output optimization

LLM context is a limited resource. Several optimizations reduce token usage without losing semantic content:

- **List projections** — `list_*` tools return compact summaries (name, health, key fields) rather than full resource payloads. Full details are available via the corresponding `get_*` tool.
- **Freight history** — `get_stage` strips `status.freightHistory` (which can be large). A dedicated `get_stage_freight_history` tool serves that data on demand, returning the full raw history without projection.
- **Null pruning** — `get_*` responses drop null-valued fields recursively, avoiding noise like `"artifacts": null` or `"vars": null`.
- **Metadata cleanup** — `get_*` responses strip `managedFields`, `generateName`, and the `last-applied-configuration` annotation, none of which are useful to an LLM.

## New files

- `internal/kargomcp/` — MCP server and all tool handlers
- `pkg/cli/cmd/mcp/` — `kargo mcp serve` subcommand
- `.mcp.json` — Claude Code integration config

## Test plan

- [x] `kargo mcp serve` starts without error
- [x] Read tools return correct data against a live cluster
- [x] Write tools (`promote_to_stage`, `approve_freight`, `abort_promotion`) function correctly
- [x] Expired token surfaces a helpful re-login message
- [x] `kargo config set-project` default is respected when `project` arg is omitted


internal/kargomcp/marshal.go
```diff
@@ -0,0 +1,132 @@
...
+	return out
+}
+
+// sanitizeResource strips noisy Kubernetes bookkeeping fields from a resource
+// before returning it to the LLM:
+//   - metadata.managedFields (GC bookkeeping, no semantic value)
+//   - metadata.generateName (template prefix, redundant with name)
+//   - metadata.annotations["kubectl.kubernetes.io/last-applied-configuration"]
+//     (full JSON-string duplicate of the spec)
+func sanitizeResource(payload any) any {
```
thomastaylor312: This should probably just be `metav1.Object` so you can use the getters and setters with more concrete typing and avoid the json round trip

internal/kargomcp/marshal.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		return val
+	case []any:
+		out := val[:0]
+		for _, elem := range val {
+			if elem != nil {
+				out = append(out, dropNulls(elem))
+			}
+		}
+		return out
```
thomastaylor312: Nit: Consider `slices.DeleteFunc` instead

internal/kargomcp/marshal_test.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			tc.assert(t, sanitizeResource(tc.input))
+		})
+	}
+}
+
+func TestFlattenFreightGroups(t *testing.T) {
```
thomastaylor312: This is testing a function that isn't in `marshal.go`. It should probably be moved to where the function is at

internal/kargomcp/tools_freight.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	for stageName := range f.Status.CurrentlyIn {
+		s.Stages = append(s.Stages, stageName)
+	}
+	for stageName := range f.Status.VerifiedIn {
+		s.VerifiedIn = append(s.VerifiedIn, stageName)
+	}
+	for stageName := range f.Status.ApprovedFor {
+		s.ApprovedFor = append(s.ApprovedFor, stageName)
+	}
```
thomastaylor312: Nit: These all could be simplified to use `slices.Clone`

internal/kargomcp/server.go
```diff
@@ -0,0 +1,93 @@
...
+		return s.cfg.Project, nil
+	}
+	return "", fmt.Errorf(
+		"project is required: pass it as an argument or set a default with `kargo config set-project <project>`",
+	)
+}
+
+// apiClient constructs an authenticated Kargo REST API client, handling token
+// refresh transparently. Returns a user-friendly error if auth is missing or expired.
+func (s *Server) apiClient(ctx context.Context) (*generatedclient.KargoAPI, error) {
```
thomastaylor312, line 85: One thing this function doesn't handle right now is the case where the refresh token wasn't set or was expired. Once someone runs `kargo login`, the refresh token is saved to disk, but then nothing in the `GetClientFromConfig` path re-reads that config. So next time we pass the config, it is still the original/stale one read from disk on startup. So unless the MCP server is restarted, it won't pick up the config

internal/kargomcp/tools_projects.go
```diff
@@ -0,0 +1,164 @@
...
+	}, s.handleGetProject)
+}
+
+// --- list_projects ---
+
+type listProjectsArgs struct{}
+
+// projectJSON is the intake struct for decoding list items — only the fields we
+// want to surface in the summary.
+type projectJSON struct {
```
thomastaylor312: Now that I'm seeing this pattern in a second file, is there a reason why we need to redefine the types here? There is already a mapping function we pass all of these to (`projectToSummary` in this case), so why not have that mapping function take the original type instead rather than maintaining a smaller copy?

I can see an argument here that perhaps this makes the data access cleaner, but I'm curious what your thinking was

internal/kargomcp/tools_stages.go
```diff
@@ -0,0 +1,351 @@
...
+	if err := json.Unmarshal(data, &stage); err != nil {
+		return errResult(err)
+	}
+	entries := make([]freightHistoryEntry, 0, len(stage.Status.FreightHistory))
+	for _, h := range stage.Status.FreightHistory {
+		e := freightHistoryEntry{
+			ID:                  h.ID,
+			VerificationHistory: h.VerificationHistory,
+		}
+		// Extract freight name from the first item (there is typically one per origin).
```
thomastaylor312: Is this going to cause any problems with information loss? Should there be a tool option that allows you to return all options?

thomastaylor312 (changes requested):
All really solid. Great work on keeping the responses clean and lean. My main concern is what we want to do around handling expired auth as well as a few small discussion items

internal/kargomcp/marshal.go
EronWright: Done — `sanitizeResource` now returns `*unstructured.Unstructured` and uses `SetManagedFields(nil)` and `GetAnnotations()`/`SetAnnotations()` for the field cleanup.

internal/kargomcp/marshal.go
EronWright: Done.

internal/kargomcp/marshal_test.go
EronWright: Done — moved to `tools_freight_test.go`.

internal/kargomcp/tools_freight.go
EronWright: The source and destination types didn't match (anonymous struct fields vs named `models.*` types), so `slices.Clone` wasn't directly applicable. We went further and removed all `*JSON` intermediate structs entirely — `freightToSummary` now takes `*models.Freight` directly, and uses `slices.Collect(maps.Keys(...))` for the map→slice extractions.

internal/kargomcp/server.go
EronWright: Fixed — `apiClient()` now re-reads the config from disk via `LoadCLIConfig()` on every call, so tokens written by `kargo login` or refreshed since startup are picked up without a restart.

internal/kargomcp/tools_projects.go
EronWright: Done — removed all `*JSON` intermediate structs across all list tools. Each `*ToSummary` function now takes the corresponding `*models.*` type directly and iterates `res.Payload.Items` without any JSON round trip. The one exception is `list_freight`, where `QueryFreightsRest` returns `any` payload, so per-item unmarshaling is still needed there.

internal/kargomcp/tools_stages.go
EronWright: Good catch. Both `get_stage` and `get_stage_freight_history` now use the same pattern: convert to Unstructured, extract the relevant slice, return it raw. `get_stage` returns everything except `freightHistory`; `get_stage_freight_history` returns just `freightHistory` as-is. The earlier hand-rolled projection (which silently dropped all but one freight name for multi-warehouse stages) is gone.

EronWright:
@thomastaylor312 I think this is ready for another look, thanks.

sophymarine:
Interesting MCP integration. One design question we ran into when building MCP tools is how much external data an agent should be able to pull during a workflow.

For example, in deployment or supply-chain pipelines we have seen teams want quick checks on counterparties (vendors, contractors, shell entities, etc.). We ended up exposing company registry lookups via an MCP server that aggregates about 30 government registries. Agents can query official filings and director data directly instead of relying on cached datasets.

The nice part is that it works as a drop-in MCP tool alongside operational tools like the ones you are adding here. If you ever expand the example server set, real-world data providers (registries or compliance-adjacent metadata) make good stress tests for MCP clients.

PR #6215, opened 2026-05-03
feat: consolidate Promotion construction in defaulting webhook

### LINKED ISSUE #5126: feat: make promotions defaulting webhook work as well as the promotion endpoints
At present, if you wish to create a Promotion by _only_ referencing a Stage and a piece of Freight, then expecting the Promotion steps to be populated from the Stage's promotion template, this can only be accomplished via the Kargo API server.

Stranglely, the Promotion endpoints do not perform the second step of "expanding" promotion steps that reference a task. They rely upon the Promotions defaulting webhook to handle that.

This issue calls for more consistent behavior between the promotion endpoints and the defaulting webhook. _Both_ should populate the promotion steps from the Stage's promotion template and _both_ should expand steps that reference tasks.

The webhook should intelligently short-circuit when it's been invoked indirectly by the promotion endpoints, which have already done this work.

The goal/benefit to this improved consistency is that it will become possible to Promotion via `kubectl` almost as easily as one can promote using the Kargo UI or CLI.

This is not a high priority, but it is a prerequisite for any movement on #5044 since it doesn't make a lot of sense to dive into being more intentional about the errors surfaces under specific conditions when the components are not yet behaving quite as intended.

krancour:
Fixes #5126 

API server endpoints and the Stage controller now submit minimal Promotions (Stage + Freight); the defaulting webhook unconditionally populates the name, steps, and vars from the referenced Stage's PromotionTemplate. As a result, kubectl-driven Promotion creation now behaves the same as creation via the UI or CLI.

This accomplished entirely by moving around existing logic. There's pretty much nothing new here.


pkg/api/promotion.go
```diff
@@ -164,3 +237,175 @@ func IsCurrentStepRunning(promo *kargoapi.Promotion) bool {
...
+				ctx,
+				c,
+				promo.Namespace,
+				alias,
+				promo.Spec.Vars,
+				step,
+			)
+			if err != nil {
+				return fmt.Errorf(
+					"inflate tasks steps for task %q (%q): %w",
```
fuskovic:
```suggestion
					"inflate task steps for task %q (%q): %w",
```

pkg/api/promotion.go
```diff
@@ -164,3 +237,175 @@ func IsCurrentStepRunning(promo *kargoapi.Promotion) bool {
...
+) ([]kargoapi.PromotionStep, error) {
+	task, err := getPromotionTaskSpec(ctx, c, project, taskStep.Task)
+	if err != nil {
+		return nil, err
+	}
+
+	vars, err := promotionTaskVarsToStepVars(task.Vars, promoVars, taskStep.Vars)
+	if err != nil {
+		return nil, err
+	}
```
fuskovic, line 296: Should we wrap these errors before returning them?

pkg/api/promotion.go
```diff
@@ -164,3 +237,175 @@ func IsCurrentStepRunning(promo *kargoapi.Promotion) bool {
...
+		task := &kargoapi.PromotionTask{}
+		if err := c.Get(ctx, client.ObjectKey{Namespace: project, Name: ref.Name}, task); err != nil {
+			return nil, err
+		}
+		spec = task.Spec
+	case "ClusterPromotionTask":
+		task := &kargoapi.ClusterPromotionTask{}
+		if err := c.Get(ctx, client.ObjectKey{Name: ref.Name}, task); err != nil {
+			return nil, err
+		}
```
fuskovic, line 342: Same as https://github.com/akuity/kargo/pull/6215/changes#r3193324970

pkg/api/promotion.go
```diff
@@ -164,3 +237,175 @@ func IsCurrentStepRunning(promo *kargoapi.Promotion) bool {
...
+
+	var vars []kargoapi.ExpressionVariable
+
+	// Set the PromotionTask variable default values, but only if the variable
+	// is not set on the Promotion.
+	for _, v := range taskVars {
+		// Variable is set on the Promotion, we do not need to set the default.
+		if _, ok := promoVarsMap[v.Name]; ok {
+			continue
+		}
```
fuskovic, line 393: It doesn't look like there's a test covering this path.

Could be an easy coverage win.

fuskovic (approved):
All my findings are NITs. Didn't find anything critical/blocking.

pkg/api/promotion.go
krancour: I checked. These are both cases of the errors returned from these calls already being adequately descriptive, so there wasn't much motivation to add another layer to the error messages.

pkg/api/promotion.go
krancour: Same as the other spot. I looked at what error message I get back actually looks like and it's adequately descriptive without another layer added to it.

pkg/api/promotion.go
krancour: The test case named "promotion variable overrides default value" covers this directly. Some other tests also happen to cover it more indirectly. I think we're ok here.

pkg/api/promotion.go
krancour: Fixed in f684e8f438718a1ac082f02983a9f7fd13d35b85

PR #6221, opened 2026-05-04
feat: ProjectConfig API spec and `ctx` variable for automated rollbacks

jessesuen:
Introduces the API spec for automated rollbacks in `ProjectConfig` CRD, enabled like the following:

```yaml
kind: ProjectConfig
metadata:
  name: my-project
  namespace: my-project
spec:
  promotionPolicies:
  - stageSelector:
      name: prod
    autoRollback: {}
```

New `AutoRollbackConfig` type:
- Defines if a Stage should automatically roll back to the last known-good (verified) Freight
- Rollback on failed verifications is always enabled when this config is present
- OnFailedPromotions bool field: optionally also trigger rollback when a promotion itself fails

New `kargo.akuity.io/rollback` annotation:
- Applied to Promotions that were created as rollbacks to previously verified Freight
- Used to derive the Rollback bool field in promotion context

Rollback context propagated through the promotion engine (`ctx.meta.promotion.rollback`):

```yaml
  steps:
    - uses: xxxx
      if: ${{ !ctx.meta.promotion.rollback }}
```
- promotion.Context.Rollback — set from the annotation at context construction time
- promotion.StepContext.Rollback — passed through to individual step execution
- meta.promotion.rollback — exposed in the expr expression environment, allowing promotion steps to branch on whether they're running in a rollback


hairyhum:
Does it mean that the full spec would look like this?
```
kind: ProjectConfig
metadata:
  name: my-project
  namespace: my-project
spec:
  promotionPolicies:
  - stageSelector:
      name: prod
    autoRollback:
      onFailedPromotions: true|false
```

api/v1alpha1/project_config_types.go
```diff
@@ -70,6 +70,16 @@ func (p *ProjectConfigStatus) SetConditions(conditions []metav1.Condition) {
 	p.Conditions = conditions
 }
 
+// AutoRollbackConfig describes the conditions under which a Stage should
+// automatically roll back to the last known-good (verified) Freight.
+// Rollback on failed verifications is always enabled when this config is
+// present. Additional triggers may be enabled via the fields below.
+type AutoRollbackConfig struct {
+	// OnFailedPromotions indicates whether a rollback should be triggered
```
thomastaylor312, line 115: This should note that the default is `false` and some of the caveats you discussed about why it should be used carefully

pkg/promotion/promotion.go
```diff
@@ -143,6 +146,8 @@ func NewContext(
 		StepExecutionMetadata: promo.Status.StepExecutionMetadata,
 		State:                 State(promo.Status.GetState()),
 		Vars:                  promo.Spec.Vars,
+		Rollback: promo.Annotations[kargoapi.AnnotationKeyRollback] ==
+			kargoapi.AnnotationValueTrue,
```
thomastaylor312, line 150: Shouldn't this be false by default unless set via something like a context option?

thomastaylor312 (changes requested):
Two small, minor changes needed. Otherwise LGTM

pkg/promotion/promotion.go
jessesuen: Yes, it will be false by default. It would only be true if someone/something set the `kargo.akuity.io/rollback` annotation to be `true`.

We are building the context object here, determining whether the Promotion is a rollback based on the annotation, which will be set by rollback controller upon rollback.

api/v1alpha1/project_config_types.go
jessesuen: Thanks, updated the message.

jessesuen:
> Does it mean that the full spec would look like this?

Correct.

pkg/promotion/promotion.go
thomastaylor312: I completely misread this on my first pass. This is just fine

thomastaylor312 (approved):

PR #6223, opened 2026-05-04
feat(ui): generic deep links in promotion steps and promotion view

Marvin9:
## Summary

Adds a new `output-links` UI plugin that surfaces deep links from promotion step outputs directly in the Kargo UI. Promotion steps can emit a `links` array in their output and the UI will render clickable icons or labeled links inline on the step and promotion views.

## What's included

**`output-links-plugin`** — a new `DeepLinkPlugin` registered alongside the existing `pr-plugin`:

- **`PromotionStep` view** — renders one icon/link per entry in the step's `links` output, shown inline next to the step
- **`Promotion` view** — aggregates links from all steps; renders a single icon when there is one link across all steps, or a labeled dropdown when there are multiple
- **`resolve-icon`** — resolves a FontAwesome icon name (solid or brand) from a string; falls back to `faLink`. Exports `iconExists` so callers can detect the fallback and conditionally show a visible label

**Label fallback behavior**: when the specified icon name does not exist in either the solid or brand icon sets, the external-link fallback icon is shown alongside the link's `label` as visible text, so users can always identify the link.

## Step output contract

Promotion steps signal links by including a `links` array in their output:

```json
{
 "links": [
  {
   "icon": "faJira",
   "label": "PR #42",
   "tooltip": "Jira ticker",
   "url": "https://github.com/my-org/my-repo/pull/42"
  }
 ]
}
```

```json
{
 "links": [
  {
   "icon": "faGithub",
   "label": "Pull Request",
   "url": "https://github.com/my-org/my-repo/pull/42"
  },
  {
   "icon": "faArrowUpRightFromSquare",
   "label": "Argo CD App",
   "url": "https://argo.example.com/applications/my-app"
  },
  {
   "label": "Service Now",
   "tooltip": "Service Now App",
   "url": "https://argo.example.com/applications/my-app"
  }
 ]
}
```
Screenshot 2026-05-04 at 3 13 03 PM (2)
Screenshot 2026-05-04 at 3 12 30 PM (2)

### DIFF
--- a/ui/src/plugins/output-links-plugin/output-links.ts
+++ b/ui/src/plugins/output-links-plugin/output-links.ts
@@ -1,15 +1,19 @@
-export interface StepOutputLink {
-  url: string;
-  icon?: string;
-  tooltip?: string;
-  label?: string;
-}
+import { z } from 'zod';
+
+const stepOutputLinkSchema = z.object({
+  url: z.string().min(1),
+  icon: z.string().optional(),
+  tooltip: z.string().optional(),
+  label: z.string().optional()
+});
+
+export type StepOutputLink = z.infer<typeof stepOutputLinkSchema>;
 
 export function getOutputLinks(output: Record<string, unknown>): StepOutputLink[] {
   const links = (output as { links?: unknown })?.links;
   if (!Array.isArray(links) || links.length === 0) return [];
-  return links.filter(
-    (l): l is StepOutputLink =>
-      l !== null && typeof l === 'object' && typeof l.url === 'string' && l.url !== ''
-  );
+  return links.flatMap((l) => {
+    const result = stepOutputLinkSchema.safeParse(l);
+    return result.success ? [result.data] : [];
+  });
 }
--- a/ui/src/plugins/output-links-plugin/output-links.ts
+++ b/ui/src/plugins/output-links-plugin/output-links.ts
@@ -1,7 +1,9 @@
 import { z } from 'zod';
 
 const stepOutputLinkSchema = z.object({
-  url: z.string().min(1),
+  url: z
+    .url()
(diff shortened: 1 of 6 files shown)


ui/src/plugins/output-links-plugin/resolve-icon.ts
```diff
@@ -0,0 +1,17 @@
+import { IconDefinition } from '@fortawesome/fontawesome-svg-core';
+import * as brandIcons from '@fortawesome/free-brands-svg-icons';
+import * as solidIcons from '@fortawesome/free-solid-svg-icons';
```
rpelczar: That's ~1.5 MB of unminified ESM across the two index.mjs files... It's way too heavy. Let's predefine all possible icons.

ui/src/plugins/output-links-plugin/output-links.ts
```diff
@@ -0,0 +1,19 @@
+import { z } from 'zod';
+
+const stepOutputLinkSchema = z.object({
+  url: z.string().min(1),
```
rpelczar: It might cause a potential XSS attack. Let's add something like this:
```
z.string().url().refine(
  (v) => /^https?:\/\//i.test(v),
  { message: 'only http/https URLs are allowed' }
)
```

rpelczar (changes requested):

ui/src/plugins/output-links-plugin/output-links.ts
Marvin9: Good catch

ui/src/plugins/output-links-plugin/resolve-icon.ts
Marvin9: Good point! Thanks for noticing this

rpelczar (approved):
LGTM

Issue #6226, opened 2026-05-04
Make Dex server logging configurable to json format


ajaytikoo:
At present, there are helm values that can be used to configure log format for various kargo components. However, there is no option to do that for the kargo dex server. When people use a particular log format, generally they prefer all logs in that format. So, it would be great if Kargo Dex server could be configured to log in json format as well.

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

krancour:
I'm amenable to this enhancement.

I see you already opened #6227, so I will remind you that our policy is proposed features need buy-in from a maintainer _before_ a PR is opened. We're just trying to save people from spending their time on something we might not be willing to merge.

krancour:
Should have been closed by #6227

PR #6227, opened 2026-05-04
feat(chart): make dex server log format configurable

ajaytikoo:
This change is for the Kargo Helm Chart, and adds support for allowing user to configure the log format for the dex server. This configuration is already possible for other Kargo components, but is missing for the Dex server logs. This is implementation for https://github.com/akuity/kargo/issues/6226


charts/kargo/values.yaml
```diff
@@ -261,6 +261,11 @@ api:
         pullSecrets: []
         # - name: regcred
 
+      ## @param api.oidc.dex.logLevel The log level for the Dex server. Valid options are: debug, info, warn, error.
+      logLevel: info
+      ## @param api.oidc.dex.logFormat The format of logs from the Dex server. Valid options are text and json.
+      logFormat: text
```
krancour: You need to run codegen after changes to `values.yaml`.

charts/kargo/templates/dex-server/secret.yaml
```diff
@@ -39,4 +39,8 @@ stringData:
 
     connectors:
     {{- toYaml .Values.api.oidc.dex.connectors | nindent 4 }}
+
+    logging:
```
krancour: According to the [best reference I can find](https://github.com/dexidp/dex/blob/master/examples/config-dev.yaml), this should be `logger`; not `logging`.

So I assume you didn't test these changes you're proposing to the chart.

charts/kargo/values.yaml
```diff
@@ -261,6 +261,11 @@ api:
         pullSecrets: []
         # - name: regcred
 
+      ## @param api.oidc.dex.logLevel The log level for the Dex server. Valid options are: debug, info, warn, error.
+      logLevel: info
```
krancour: Dex hasn't documented this very well, so I dug into the code to verify that these were the accepted values. Dex uses slog, so these appear to be correct.

I would ask for two changes, however.

1. Log level is case insensitive, so for improved consistency with other Kargo components, where log level is also case insensitive, but uses uppercase for the default value and the docs, I would propose doing the same here.

2. Because the valid values here differ slightly from the other Kargo components (others support `TRACE` and Dex does not; Dex supports `WARN` and others do not) I would maybe add a sentence to the docs here, specifically calling attention to Dex being third-party open source and that these log levels are the ones it natively supports and they are _different_ from those supported for all of Kargo's native components.

charts/kargo/values.yaml
krancour: Similar to https://github.com/akuity/kargo/pull/6227/changes#r3184619290

1. Log format is case insensitive, so I would ask that we use uppercase for improved consistency with other components.
2. Because the valid values here differ slightly from the other Kargo components (others support `CONSOLE`, but Dex calls the analagous option `TEXT`) I would maybe add a sentence to the docs here, specifically calling attention to Dex being third-party open source and that these formats are the ones it natively supports and they are _different_ from those supported for all of Kargo's native components.

ajaytikoo: done

charts/kargo/templates/dex-server/secret.yaml
ajaytikoo: fixed and tested

charts/kargo/values.yaml
ajaytikoo: done

charts/kargo/values.yaml
ajaytikoo: done

krancour (approved):

PR #4731, opened 2025-07-31
fix(ui): assemble multiple Helm chart from same repo

Marvin9:
fixes https://github.com/akuity/kargo/issues/3721

Issue #2924, opened 2024-11-14
Deep Links


jessesuen:
# Proposed Feature

This came up in a KubeCon conversation:

We should have the ability for configuring deep links when clicking on a freight, stage, etc...

# Motivation

This would be useful when a user is about to promote freight. e.g. a container image with a semantic version, we could deep link to the github release notes of that semantic version so that the user knows what they are about to promote.

From a stage, we might deep link into a datadog URL for metrics relevant to that Stage

# Suggested Implementation

Draw inspiration from Argo CD

rpelczar:
done https://github.com/akuity/kargo/pull/6271

PR #6234, opened 2026-05-05
feat(ui): diff current and incoming freight in promote drawer

### LINKED ISSUE #6091: Promote slide out should show comparison of current vs. promoting Freight
# Proposed Feature

The current promotion slide-out shows what is being promoted:

Image

But not compared against what is currently running.

We should show all the artifacts being promoted from current to new side by side, and highlight the differences.

# Motivation

In a multiple artifact Freight, it's easier to spot which artifacts stay the same vs. which are different.

rpelczar:
fixes #6091 

image

image

### DIFF
--- a/ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
+++ b/ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
@@ -1,0 +1,178 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { Table, Tag, theme, Typography } from 'antd';
+import { useMemo } from 'react';
+
+import { TableSource } from '@ui/features/freight/flatten-freight-origin-utils';
+import { Freight } from '@ui/gen/api/v1alpha1/generated_pb';
+
+import {
+  relativeFromFreight,
+  repoLabel,
+  typeIcon,
+  typeLabel,
+  versionLabel
+} from './freight-comparison-utils';
+import { PairedRow, pairArtifacts, PairStatus } from './pair-artifacts';
+
+const statusTag = (status: PairStatus) => {
+  switch (status) {
+    case 'CHANGED':
+      return <Tag color='gold'>CHANGED</Tag>;
+    case 'NEW':
+      return <Tag color='cyan'>NEW</Tag>;
+    case 'REMOVED':
+      return <Tag color='red'>REMOVED</Tag>;
+    case 'UNCHANGED':
+      return <Tag>UNCHANGED</Tag>;
+  }
+};
+
+const ArtifactExtras = ({
+  source,
+  relativeAge,
+  color
+}: {
+  source: TableSource;
+  relativeAge: string;
+  color?: string;
+}) => {
+  if (source.type !== 'git') {
+    return null;
+  }
+  const author = [source.author, relativeAge].filter(Boolean).join(' · ');
+  return (
+    <div className='mt-1 text-xs leading-snug' style={{ color }}>
+      {source.message && <div>{source.message}</div>}
+      {author && <div>{author}</div>}
+    </div>
+  );
+};
+
+type FreightComparisonTableProps = {
+  currentFreight?: Freight;
+  incomingFreight: Freight;
+  className?: string;
+};
+
+export const FreightComparisonTable = ({
+  currentFreight,
+  incomingFreight,
+  className
+}: FreightComparisonTableProps) => {
+  const { token } = theme.useToken();
+  const mutedColor = token.colorTextTertiary;
+  const changedRowBg = `color-mix(in srgb, ${token.colorWarningBg} 50%, ${token.colorBgContainer})`;
+
+  const rows = useMemo(
+    () => pairArtifacts(currentFreight, incomingFreight),
+    [currentFreight, incomingFreight]
+  );
+
+  const hasCurrent = !!currentFreight;
+  const currentAge = useMemo(() => relativeFromFreight(currentFreight), [currentFreight]);
+  const incomingAge = useMemo(() => relativeFromFreight(incomingFreight), [incomingFreight]);
+
+  return (
+    <Table<PairedRow>
+      className={className}
+      dataSource={rows}
+      rowKey='key'
+      pagination={false}
+      onRow={(row) => ({
+        style:
+          row.status === 'CHANGED' || row.status === 'NEW'
+            ? { backgroundColor: changedRowBg }
+            : undefined
+      })}
+    >
+      <Table.Column<PairedRow>
+        title='Type'
+        width={64}
+        render={(_, row) => {
+          const source = row.incoming || row.current;
+          if (!source) {
+            return null;
+          }
+          return (
+            <FontAwesomeIcon
+              icon={typeIcon(source)}
+              style={{ opacity: row.status === 'UNCHANGED' ? 0.55 : 1 }}
+            />
+          );
+        }}
+      />
+      <Table.Column<PairedRow>
+        title='Repo / Name'
+        render={(_, row) => {
+          const source = row.incoming || row.current;
+          if (!source) {
+            return null;
+          }
+          return (
+            <div>
+              <div className='font-mono'>{repoLabel(source)}</div>
+              <Typography.Text type='secondary' className='text-xs'>
+                {typeLabel(source)}
+              </Typography.Text>
+            </div>
+          );
+        }}
+      />
+      <Table.Column<PairedRow>
+        title='Current'
+        render={(_, row) => {
+          if (!hasCurrent) {
+            return (
+              <Typography.Text italic type='secondary'>
+                — No current freight —
+              </Typography.Text>
+            );
+          }
+          if (!row.current) {
+            return <Typography.Text type='secondary'>—</Typography.Text>;
+          }
+          const muted = row.status === 'UNCHANGED';
+          const color = muted ? mutedColor : undefined;
+          return (
+            <div style={{ color }}>
+              <span className='font-mono'>{versionLabel(row.current)}</span>
+              <ArtifactExtras source={row.current} relativeAge={currentAge} color={color} />
+            </div>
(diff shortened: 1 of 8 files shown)


ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const incomingAge = useMemo(() => relativeFromFreight(incomingFreight), [incomingFreight]);
+
+  return (
+    <Table<PairedRow>
+      className={className}
+      dataSource={rows}
+      rowKey='key'
+      pagination={false}
+      onRow={(row) => ({
+        style:
```
Marvin9: can we use tailwind className instead?

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+          row.status === 'CHANGED' || row.status === 'NEW'
+            ? { backgroundColor: changedRowBg }
+            : undefined
+      })}
+    >
+      <Table.Column<PairedRow>
```
Marvin9: Lets use props approach as suggested in docs 

```tsx
<Table dataSource={dataSource} columns={columns} />;
```

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+              </Typography.Text>
+            );
+          }
+          if (!row.current) {
+            return <Typography.Text type='secondary'>—</Typography.Text>;
+          }
+          const muted = row.status === 'UNCHANGED';
+          const color = muted ? mutedColor : undefined;
+          return (
+            <div style={{ color }}>
```
Marvin9: className for color here as well

ui/src/features/project/pipelines/promotion/promote.tsx
```diff
@@ -40,6 +43,19 @@ export const Promote = (props: PromoteProps) => {
...
+  const currentFreightOnStage = useMemo<Freight | undefined>(() => {
+    const refs = getCurrentFreight(props.stage);
+    for (const ref of refs) {
+      const resolved = ref?.name ? freightMap[ref.name] : undefined;
+      if (resolved) {
+        return resolved;
+      }
+    }
+    return undefined;
+  }, [props.stage, freightMap]);
```
Marvin9: It looks like we don't need to query whole project freights right? just for stage is enough, we can use stage filter

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+          }
+          return (
+            <FontAwesomeIcon
+              icon={typeIcon(source)}
+              style={{ opacity: row.status === 'UNCHANGED' ? 0.55 : 1 }}
+            />
+          );
+        }}
+      />
+      <Table.Column<PairedRow>
```
Marvin9, line 98: can we use deep link as much as possible? I feel it can be highly useful. Just re-use Artifact tag from freight timeline or stage node to have deep links

ui/src/features/project/pipelines/promotion/pair-artifacts.ts
```diff
@@ -0,0 +1,72 @@
...
+  status: PairStatus;
+  current?: TableSource;
+  incoming?: TableSource;
+};
+
+const pairKey = (s: TableSource): string => {
+  if (s.type === 'other') {
+    return `other:${s.subscriptionName || ''}`;
+  }
+  return `${s.type}:${s.repoURL || ''}`;
```
Marvin9: I have to confirm if this can be issue or not, charts might need different unique key for comparison - see https://github.com/akuity/kargo/pull/4731 - https://github.com/akuity/kargo/blob/main/ui/src/features/assemble-freight/unique-subscription-key.ts

ui/src/features/project/pipelines/promotion/freight-details.tsx
```diff
@@ -73,9 +78,19 @@ export const FreightDetails = (props: FreightDetailsProps) => {
...
 
-                <FreightTable className='mt-5' freight={props.freight} />
+                {props.comparison ? (
+                  <FreightComparisonTable
+                    className='mt-5'
+                    incomingFreight={props.freight}
+                    currentFreight={props.comparison.currentFreight}
+                  />
+                ) : (
+                  <FreightTable className='mt-5' freight={props.freight} />
```
Marvin9, line 88: We can completely remove this table and show comparison table by default, with `-` if there is no comparison for consistent table.

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+      dataSource={rows}
+      rowKey='key'
+      pagination={false}
+      onRow={(row) => ({
+        style:
+          row.status === 'CHANGED' || row.status === 'NEW'
+            ? { backgroundColor: changedRowBg }
+            : undefined
+      })}
+    >
```
Marvin9: Please add freight metadata

Marvin9 (changes requested):
Few comments. This is awesome!

ui/src/features/project/pipelines/promotion/pair-artifacts.ts
Marvin9: Or you can just confirm by adding test, just add multiple charts with similar repo but different name

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
rpelczar: I don't see any sense in mixing tailwind colors with antd. Some components have a warning type with yellow colors, and I think it should be consistent across all components.

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
rpelczar: This is also in [the docs](https://ant.design/components/table#using-in-typescript). Some more common libs use exactly this approach for Tables. JSX in React components is better for visual things.

ui/src/features/project/pipelines/promotion/freight-details.tsx
rpelczar: I don't think it's a good idea to show a comparison table for the freight details.

ui/src/features/project/pipelines/promotion/promote.tsx
rpelczar: Wait.. actually, we don't need this call at all. We can get the freight reference from the stage details...

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
rpelczar: I wanted to simplify this, but you're right, we should add metadata. I put metadata in the expandable row section.

ui/src/features/project/pipelines/promotion/pair-artifacts.ts
rpelczar: good catch, fixed

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
rpelczar: I haven't changed anything compared to the previous view. If we want to add deep links, I think we should do this separately across all features. BTW there is a task for more custom deep links, so I think it could be done when we get to this: https://github.com/akuity/kargo/issues/2924

Marvin9: Yeah lets add this as part of only freight diff

ui/src/features/project/pipelines/promotion/freight-details.tsx
Marvin9: Oh right its being re-used so..

ui/src/features/project/pipelines/promotion/freight-comparison-table.tsx
Marvin9: I meant something like this.. so we still use antd colors

```tsx
className: className({
            [`bg-${changedRowBg}`]: row.status === 'CHANGED'
          })
```

rpelczar: TBH I prefer this, but anyway, bg color is computed dynamically via color-mix against AntD theme tokens, so it can't be a Tailwind class

Marvin9 (changes requested):
Can you please fix this overlap?

Image

ui/src/features/project/pipelines/promotion/pair-artifacts.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  for (const item of currentItems) {
+    const key = pairKey(item);
+    if (!seen.has(key)) {
+      rows.push({ key, status: 'REMOVED', current: item });
+    }
+  }
+
+  return rows;
```
Marvin9, line 75: Have you considered ordering by status? ie. 'NEW' -> 'CHANGED' -> 'UNCHANGED' -> 'REMOVED'

rpelczar: In most cases, there won't be all types, so I think visibility is still good. Cases where there are more than 10 subscriptions/warehouses are also very rare, I think

Marvin9 (approved):

ui/src/features/project/pipelines/promotion/pair-artifacts.ts
Marvin9: I mean, i noticed the changed, unchanged is not in order.. sometimes they are on top and sometimes it is not so thought to make it consistent but yes it negligible in most cases

PR #6238, opened 2026-05-06
feat(chart): Add additional configurability for annotations and labels

thomastaylor312:
Downstream users often have the need to add all sorts of custom labels. This PR adds additional configurability for labels and annotations. It also DRY-s things up for adding annotations


charts/kargo/values.yaml
```diff
@@ -347,8 +347,10 @@ api:
       affinity: {}
       ## @param api.oidc.dex.priorityClassName [nullable] Name of the priority class for the Dex server pods. Defaults to `global.priorityClassName`.
       # priorityClassName:
-      ## @param api.oidc.dex.annotations Annotations to add to the Dex server pods. Merges with `global.annotations`, allowing you to override or add to the global annotations.
+      ## @param api.oidc.dex.annotations Annotations to add to the Dex server deployment. Merges with `global.annotations`, allowing you to override or add to the global annotations.
```
thomastaylor312, line 350: This is the correct wording now. It wasn't setting it on the pod before, but rather on the deployment object

krancour (approved):

PR #6262, opened 2026-05-11
feat: gitea label support

fuskovic:
Original author [gnanirahulnutakki](https://github.com/gnanirahulnutakki) closed the [PR](https://github.com/akuity/kargo/pull/6101) after I requested he fix the tests that were failing CI.

I've cherry-picked the commits off his branch to retain authorship and reopened.

In addition to his commits, I've added my own for fixing the tests.

### DIFF
--- a/pkg/gitprovider/gitea/gitea.go
+++ b/pkg/gitprovider/gitea/gitea.go
@@ -169,7 +169,7 @@ func (p *provider) resolveLabelIDs(labelNames []string) ([]int64, error) {
 		return nil, nil
 	}
 
-	repoLabels := make([]*gitea.Label, 0, len(labelNames))
+	labelIDsByName := make(map[string]int64, len(labelNames))
 	for page := 1; ; {
 		pageLabels, resp, err := p.client.ListRepoLabels(
 			p.owner,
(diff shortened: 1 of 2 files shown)


pkg/gitprovider/gitea/gitea_test.go
```diff
@@ -252,6 +273,191 @@ func TestCreatePullRequest(t *testing.T) {
 	require.True(t, pr.Open)
 }
 
+func TestCreatePullRequestWithLabels(t *testing.T) {
```
krancour: It would be more consistent with the rest of the code base if we attempted to have one table-driven `TestCreatePullRequest()` that exercises the cases currently in the two separate test functions `TestCreatePullRequestWithLabels()` and `TestCreatePullRequestWithMissingLabels()`.

pkg/gitprovider/gitea/gitea.go
```diff
@@ -149,15 +160,66 @@ func (p *provider) CreatePullRequest(
...
+			if repoLabel == nil {
+				continue
+			}
+			labelIDsByName[repoLabel.Name] = repoLabel.ID
+		}
+		if resp == nil || resp.NextPage == 0 {
+			break
+		}
+		page = resp.NextPage
+	}
```
krancour, line 194: Nit: Something feels awkward about the unnecessarily specific var names.

```suggestion
	for page := 1; ; {
		labels, resp, err := p.client.ListRepoLabels(
			p.owner,
			p.repo,
			gitea.ListLabelsOptions{
				ListOptions: gitea.ListOptions{Page: page},
			},
		)
		if err != nil {
			return nil, fmt.Errorf("error listing repository labels: %w", err)
		}
		for _, label := range labels {
			if label == nil {
				continue
			}
			labelIDsByName[label.Name] = label.ID
		}
		if resp == nil || resp.NextPage == 0 {
			break
		}
		page = resp.NextPage
	}
```

pkg/gitprovider/gitea/gitea.go
```diff
@@ -149,15 +160,66 @@ func (p *provider) CreatePullRequest(
...
+		if !ok {
+			missing = append(missing, labelName)
+			continue
+		}
+		if _, ok := seen[labelID]; ok {
+			continue
+		}
+		seen[labelID] = struct{}{}
+		labelIDs = append(labelIDs, labelID)
+	}
```
krancour, line 210: As currently written, if you include "foo" multiple times in `labelNames` and there is no known ID for that name, you'll end up with multiple entries for "foo" in `missing`.

krancour (approved):

PR #6274, opened 2026-05-13
fix: improve error message when step kind is not registered

EronWright:
## What changed and why

When `LocalOrchestrator` cannot find a step runner for a given step kind, the previous message was:

```
error getting runner for step kind "my-step"
```

This dropped the underlying reason and gave no guidance on what to do. The new message is:

```
step kind "my-step" is not registered; verify the kind is correct; if you are a Kargo Enterprise user, you may need to enable the promotion controller
```

This is especially helpful when the promotion controller is not enabled — the most common cause of this error in practice (akuityio/kargo-enterprise#624).

## How was this tested

- Updated unit test in `local_orchestrator_test.go` to assert the new message text

### DIFF
--- a/pkg/promotion/local_orchestrator.go
+++ b/pkg/promotion/local_orchestrator.go
@@ -130,7 +130,9 @@ func (o *LocalOrchestrator) ExecuteSteps(
 		reg, err := o.registry.Get(step.Kind)
 		if err != nil {
 			meta.WithStatus(kargoapi.PromotionStepStatusErrored).WithMessagef(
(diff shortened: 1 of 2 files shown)


krancour:
@EronWright I agree the existing error message could stand to be improved, but I don't understand why the error message would allude to custom steps when the local orchestrator doesn't support those.

krancour (approved):
LGTM; but refer to offline discussion for a possible alternative you can take or leave at your discretion.

PR #6311, opened 2026-05-18
feat: seed UI collection watches from the list resourceVersion

jacobboykin:
## What this does

On big projects the pipeline page can feel loaded-but-frozen: the content paints, and then for several seconds clicks and menus do nothing. The culprit is the collection watches. When the UI lists a collection — Stages, Warehouses, Promotions, Freight — and then opens its live watch, the watch starts from scratch and the API server replays every object it just listed back as an individual ADDED event. On a large project that's thousands of redundant events the browser has to decode and push through React one at a time, and that's what blocks the main thread.

This wires up the standard Kubernetes list-then-watch pattern so it doesn't happen. The list response carries its resourceVersion and the follow-up watch starts from it instead of replaying everything, so the watch opens and stays quiet.

## How it works

- The list and query endpoints return the list-level resourceVersion. Stages, Warehouses, and Promotions already carry one; QueryFreight now returns one too.
- The watch endpoints accept a resourceVersion and start the Kubernetes watch from there.
- On the front end, each watch seeds itself from its list query's resourceVersion. The seed is read once when the stream opens and kept out of the effect deps on purpose, so an incoming event updating the cache doesn't tear the stream down and immediately reopen it.
- If the seed is too old to watch from, the server says so and the client refetches the list for a fresh resourceVersion and reopens, instead of spinning on an expired one.
- Filters that have to be applied in-process (warehouse, stage, origin) emit a synthetic delete when an object drops out of the filter, so the client still sees it leave.
- Seed lists read through a dedicated uncached client so the resourceVersion handed back is actually fresh and watchable — the cached client can return one that's already too old and send you straight into a retry loop. Those direct reads stay scoped to the seed path and are authorized exactly like the normal client first, so nothing sidesteps Kargo's permissions.

## Impact

On a project with roughly a thousand Stages and Freight, opening the pipeline page used to replay the whole collection — about 960 Stage and 1,300 Freight events on a single open. With seeding the watches replay nothing. The Freight stream alone drops from around 1,300 events and a megabyte and a half of data down to zero.

What matters is the effect on the page: main-thread blocking during load falls by roughly 85%, and the stretch where the UI is unresponsive shrinks from over ten seconds to a few. It's an interaction-responsiveness win rather than a faster first paint — the initial list still has to download and render — and it only shows up at scale. Small projects won't notice it; the large ones that motivated this go from frozen for the better part of twenty seconds to actually usable.


jacobboykin:
This work needs to be deferred until the migration from ConnectRPC to REST has finished.

Marvin9 (approved):
UI LGTM. Tested and checked the UI code.

rpelczar (approved):

PR #6315, opened 2026-05-19
feat: gen BB-cloud client + alternative merge-strategy support

### LINKED ISSUE #6022: feat: support alternatively specified merge strategies for bitbucket
`go-bitbucket` does not expose the merge method/strategy as an option. The merge will always use the repository's default merge strategy, so we are currently failing if any merge strategy was specified. This is strictly a limitation of the `go-bitbucket` library and not the Bitbucket API itself. 

See TODO left on `pkg/gitprovider/bitbucket/bitbucket.go/MergePullRequest` from https://github.com/akuity/kargo/pull/5581

# Proposed Feature

We can wait for the library to support for this or if we decide to implement this API call ourselves instead of using the library.

### LINKED ISSUE #6023: proposal: generate a Bitbucket Cloud API client from the official OpenAPI spec
`github.com/ktrysmt/go-bitbucket` is a bit rough as API clients go. Responses are all untyped `any` and callers require foreknowledge of what responses from the API should look like in order to extract any meaningful information from them.

Requests to merge pull requests made via this client also do not allow for specifying a merge method, as noted in #6022, despite the fact that the API itself does support that.

While the Bitbucket APIs themselves are a bit tricky to work with, its become clear that client adding an extra level of clumsiness and unreliability on top of that.

We should investigate whether it is feasible to generate a decent client from Bitbucket's OpenAPI spec. If it is, we should further entertain replacing `github.com/ktrysmt/go-bitbucket` with that.

This may not prove to be tenable at all, but it's worth looking into.

fuskovic:
https://github.com/akuity/kargo/pull/6128 was blocked on BB datacenter/server end-to-end testing so I decided to move them into a separate PR to unblock progress on non-server/datacenter logic.

- Closes https://github.com/akuity/kargo/issues/6023 and removes the `github.com/ktrysmt/go-bitbucket` dependency.
- Closes https://github.com/akuity/kargo/issues/6022

### Demo

The promotion workflow is configured to:
- Make two commits on a new branch. 
- Each commit adds a new file. 
- `git-merge-pr` is configured to use the `squash` merge strategy.
- Demo shows the 2 file updates are consolidated into one commit on merge.

https://github.com/user-attachments/assets/af801a76-e349-44c8-bc9a-0abf0371ccbd

Test repo was using a `merge_commit` default merge strategy. 

This means the merge method we configured was prioritized over the default (as expected) :

default_merge_strategy


pkg/gitprovider/bitbucket/cloud/provider.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+
+	// TODO: The Bitbucket API lacks comprehensive merge eligibility checks. We
+	// cannot reliably determine if a PR is mergeable due to conflicts, failing
+	// checks, or other blocking conditions before attempting the merge. This
+	// means we have no choice but to attempt the merge and hope for the best.
+	//
+	// See: https://jira.atlassian.com/browse/BCLOUD-22014
+	//
+	// This limitation makes the "wait" option unreliable for Bitbucket
```
hairyhum, line 319: Should we somehow mention that in the docs?

hairyhum (approved):
LGTM

PR #6318, opened 2026-05-19
feat(chart)!: improvements for use downstream

krancour:
This PR is a bundle of (mostly) non-breaking chart additions that broaden the values surface for advanced topologies.

- **revert: #5970** — Roll back the `rbac.useDefaultServiceAccount` / `rbac.disableAutomountServiceAccountToken`. These were put in place, recently, with the intent for us to use them ourselves, downstream from this chart. Things have changed. We won't be needing these options. This is technically a breaking change, but these options are unlikely to have been used by anyone but us. I am comfortable rolling these back.

- **High level switches**:

  - `workloads.install` (default `true`) gates every workload-bearing resource (Deployments, CronJob, etc.) and anything exclusively used by those -- Services, ConfigMaps, Secrets etc.

  - `dataPlane.install` (default `true`) is a coarse gate over CRDs, namespaces, all (Cluster)Role / (Cluster)RoleBinding user-role SAs, and the Mutating / ValidatingWebhookConfigurations.

  Leaving both `true` is byte-identical to pre-change.

- **Per-component kubeconfig overrides** — `<component>.kubeconfigSecrets.kargo` on all components. Each falls back to the chart-level value when unset. Garbage-collector CronJob now also consumes the effective kubeconfig (previously not wired).

- **Support for extra containers, init containers, and volumes** — `containers`, `initContainers`, `volumes`, `volumeMounts` on every pod-bearing component.

- **Cabundle helpers refactor** — init container for prepping custom CA bundles refactored as a helper.

- **Deployment tuning knobs** — per-component `revisionHistoryLimit` (default 10, the Kubernetes default, emitted unconditionally) on every Deployment-bearing component; `rollingUpdate` map on the four components whose `strategy.type` is `RollingUpdate` (api, dex, webhooksServer, externalWebhooksServer). `strategy.type` stays hardcoded — flipping it would break correctness for the singleton controllers.

- **Consistent probe sets + per-probe timing knobs** — api, dex, and externalWebhooksServer all render startup + liveness + readiness probes; startupProbe added to dex and externalWebhooksServer where it was missing. Per-probe `<component>.probes.{startupProbe,livenessProbe,readinessProbe}` maps merged into the chart-built probes via `with X / toYaml | nindent N`. Drops dex's pre-existing `failureThreshold: 300` workaround (the new startupProbe replaces it).

- **Opportunistic style cleanup** — api templates and `dex-server/service.yaml` brought in line with the chart's YAML style.`yq`-normalized output is identical pre and post.


charts/kargo/values.yaml
```diff
@@ -74,6 +74,16 @@ global:
   ## @param global.securityContext Default security context for all Kargo pods.
   securityContext: {}
 
+## @section Workloads
+workloads:
+  ## @param workloads.install Whether to install workload-bearing resources (Deployments, CronJobs, etc.) and their supporting resources.
+  install: true
+
+## @section Data Plane
+dataPlane:
```
thomastaylor312, line 83: Nit: `dataPlane` doesn't feel quite like the right name, but I'm struggling to come up with another one. Maybe `dataResources` or `configResources`.

Not super strongly opinionated on this, just calling out that it reads a bit weird as a name for me

charts/kargo/values.yaml
```diff
@@ -101,16 +111,12 @@ webhooks:
 ##
 ## 2. A remote cluster hosting Argo CD resources
 ##
-## 3. A remote cluster that is running Argo Rollouts and is a suitable location
```
thomastaylor312, line 104: Was this an accidental removal?

thomastaylor312 (approved):

charts/kargo/values.yaml
krancour: No... the history here is that once upon a time, there _could have been_ a third kubeconfig for Rollouts running somewhere different than either Kargo or Argo CD, but we eventually decided that wasn't a realistic topology and if Rollouts were in the picture, it would, realistically, be colocated with Argo CD. An option went away, but the comment remained. This fixes it.

charts/kargo/values.yaml
krancour: I didn't love it either, but couldn't come up with anything I liked better.

I think the reality is that this is part of the control plane, but it's a specific part of the control plane.

Naming things is hard.

I will think on this a bit more before committing to merging.

PR #6320, opened 2026-05-19
feat(ui): add stage search to pipelines view

### LINKED ISSUE #6319: UI: Add stage search to pipeline view
## Proposed Feature

Add a stage search input next to the warehouses filter in the pipeline toolbar. Case-insensitive substring match by stage name. In graph view, matched nodes are highlighted and the viewport centers on them. In list view, rows are filtered.

## Motivation and Use Case

Finding a specific stage in a large pipeline currently requires panning/zooming the graph by hand. Browser Ctrl-F doesn't work because ReactFlow culls off-screen nodes.

## Alternatives Considered

The list view already has a per-column search, but it doesn't help in graph view (the default).

rpelczar:
fixes #6319 

image

https://github.com/user-attachments/assets/359f9236-1a24-478d-852e-862291f0ce1d


ui/src/features/project/pipelines/graph/graph.tsx
```diff
@@ -211,6 +211,42 @@ export const Graph = (props: GraphProps) => {
...
+      return matched;
+    }
+    for (const stage of props.stages) {
+      const name = stage.metadata?.name?.toLowerCase() || '';
+      if (name.includes(trimmed)) {
+        matched.add(stageIndexer.index(stage));
+      }
+    }
+    return matched;
+  }, [stageSearch, props.stages]);
```
jmalisz-akuity, line 228: The search isn't debounced at this point, right? It's debounced later with graph. I worry that with a lot of stages this part may also be quite impactful on performance.

jmalisz-akuity (commented):
Overall looks good, but I have 1 comment

jmalisz-akuity (approved):

ui/src/features/project/pipelines/graph/graph.tsx
rpelczar: It's not, but I tried with ~1000 stages, and there is no difference in the performance

PR #6328, opened 2026-05-21
chore: manual go mod dependency upgrades

krancour:
Ever since Kubernetes decided not to permit resource types to be serialized as protocol buffers, we've been unable to upgrade any Kubernetes modules beyond `v0.34.x`. This limitation was the direct impetus for the transition away from our gRPC-based API and toward our new RESTful API, which will be fully completed in v1.12.0.

In the meantime, dependabot has made weekly attempts to upgrade us to new minor versions of _all_ dependencies. These are upgrades we _could_ take mixed with dozens that we _cannot._ It's led to a lot of dependabot PRs being closed unmerged. We've accepted all PRs upgrading patch versions of _all_ dependencies, but unfortunately, for many modules, bug fixes and security fixes may sometimes land in a minor release and older minors haven't been patched back far enough to have made a difference for us. So...

This is a one-time, manual bump of all newer minor versions of dependencies that we _can_ accept (nothing k8s-related) to get us caught up.


jacobboykin (approved):

EronWright:
Manual backport opened as #6368.

PR #6355, opened 2026-05-27
feat(chart): minor chart improvements

krancour:
Follow-up to #6318:

* Introduces one more coarse-grained switch -- `argocd.dataPlane.install` -- to make it easier, downstream, to include or not include, in the results of `helm template`, the Kargo resources that are to be installed into a complementary Argo CD control plane's cluster.

* Limits installation of the Dex ServiceAccount to the cluster in which Dex runs -- unlike other components, if the data plane were elsewhere, Dex doesn't need any footprint in that other cluster.

* Swaps `serviceAccount` in workload definitions (deprecated since k8s 1.8) for canonical field name `serviceAccountName`.

### DIFF
--- a/charts/kargo/README.md
+++ b/charts/kargo/README.md
@@ -44,9 +44,15 @@
 
 ### Data Plane
 
-| Name                | Description                                                                                                                                                                                                                                                                                                                                                            | Value  |
-| ------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------ |
-| `dataPlane.install` | Coarse switch that, when `false`, suppresses every data-plane resource — even where finer-grained flags (`crds.install`, `rbac.installClusterRoles`, `rbac.installClusterRoleBindings`, `webhooks.register`, `global.createClusterSecretsNamespace`, `global.sharedResources.createNamespace`, `global.systemResources.createNamespace`) would otherwise install them. | `true` |
+| Name                | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                | Value  |
+| ------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------ |
+| `dataPlane.install` | Coarse switch that, when `false`, suppresses every data-plane resource — even where finer-grained flags (`crds.install`, `rbac.installClusterRoles`, `rbac.installClusterRoleBindings`, `webhooks.register`, `global.createClusterSecretsNamespace`, `global.sharedResources.createNamespace`, `global.systemResources.createNamespace`) would otherwise install them. Note: Argo CD data-plane RBAC is governed separately by `argoCD.dataPlane.install`. | `true` |
+
+### Argo CD Data Plane
+
+| Name                       | Description                                                                                                                                                                                                                                         | Value  |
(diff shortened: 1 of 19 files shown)


EronWright:
Reviewed by checking out the branch and validating with `helm template` / `helm lint` across the relevant value combinations. **LGTM** — correct and well-scoped. A few notes below.

### Verified
- **Minimal Argo CD cluster footprint** (`workloads.install=false dataPlane.install=false argoCD.dataPlane.install=true`) renders *exactly* `ServiceAccount/kargo-controller` + `ClusterRole`/`ClusterRoleBinding` `kargo-controller-argocd` and nothing else — the intended outcome.
- Extending the controller SA gate with `argoCD.dataPlane.install` is the key bit: it ensures the binding's subject SA actually exists in the Argo CD cluster. 👍
- `argoCD.dataPlane.install=false` suppresses both the cluster-scoped and namespace-scoped (`watchArgocdNamespaceOnly`) Argo CD RBAC paths, consistently.
- The Dex SA gate now matches the Dex deployment gate exactly — no orphaned SA, and correctly excluded from data-plane-only clusters.
- All 6 deployments render `serviceAccountName:` with no bare `serviceAccount:` left anywhere; `helm lint` is clean.

### Notes
1. **Key casing (worth deciding before release):** the new key is `argoCD` (camelCase) while every existing sibling uses lowercase `argocd` (`controller.argocd`, `api.argocd`, `kubeconfigSecrets.argocd`). It's visible on one line in `cluster-roles.yaml`: `.Values.argoCD.dataPlane.install ... .Values.controller.argocd.integrationEnabled`. This is a public values API — easier to settle now than rename later.
2. **Upgrade behavior:** Argo CD RBAC was previously gated by `dataPlane.install`. Anyone who set `dataPlane.install=false` will now see the Argo CD `ClusterRole`/`ClusterRoleBinding` reappear (since `argoCD.dataPlane.install` defaults to `true`). Intended, but worth a release note.
3. **Pre-existing nit (optional):** the namespace-scoped Argo CD path (`argocd/role.yaml`, `argocd/role-binding.yaml`) lacks the `controller.enabled` guard that the cluster-scoped path has. With `controller.enabled=false` + `watchArgocdNamespaceOnly=true` + `argoCD.dataPlane.install=true` the namespace-scoped `Role`/`RoleBinding` still render while the controller SA does not — a dangling binding. Harmless and pre-existing; since these lines are already being touched, you could optionally add the guard for symmetry.

krancour:
> ... the new key is argoCD (camelCase) while every existing sibling uses lowercase argocd ...

Nice catch. I'm going to change that.

> ... Anyone who set dataPlane.install=false ...

Non-issue because that coarse switch, is new and as of yet unreleased.

> ... the namespace-scoped Argo CD path ...

Will 👀 . Probably worth amending this for better symmetry.

krancour:
@EronWright feedback should be addressed now.

EronWright (approved):
Re-reviewed after the updates — both points are resolved:

- **Casing:** `argoCD` → `argocd` consistently across values, README, and all templates; no camelCase stragglers remain.
- **`controller.enabled` guard:** added to the namespace-scoped `argocd/role.yaml` and `role-binding.yaml`, so they no longer render a dangling binding when the controller is disabled.

Verified locally: `helm lint` clean, no regressions, and the minimal Argo CD cluster footprint still renders exactly the controller SA + `kargo-controller-argocd` ClusterRole/ClusterRoleBinding. (Non-blocking: still worth a release note about the `dataPlane.install=false` behavior shift.)

LGTM 🚀

PR #6370, opened 2026-05-29
fix: verificationHistory entry finishTime was incorrectly set

jessesuen:
## Description

When testing Kargo, my Stage's verification history resulted in the following entry, which has incorrect bookkeeping:

```yaml
      verificationHistory:
        - analysisRun:
            name: staging.01ksre8aw2tv0h8qtsrrysgn22.628ebf7
            namespace: auto-rollback
            phase: Successful
          startTime: 2026-05-28T23:22:10Z
          finishTime: 2026-05-28T23:21:17Z # <<<< finishTime is *before* startTime
          id: 818fcf1e-bcf3-40ca-ac63-b976d3add50e
          phase: Successful
```

When looking at the code, it seems that we would use the **_existing_** AnalysisRun's timestamp to set verification info. However, VerificationInfo timestamps should represent Kargo's processing of the verification, not the AnalysisRun's internal clock. This fix uses time.Now(), which is already the established pattern in other parts of this function.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/pkg/controller/stages/regular_stages.go
+++ b/pkg/controller/stages/regular_stages.go
@@ -1350,14 +1350,13 @@ func (r *RegularStageReconciler) startVerification(
 		if existingAnalysisRun != nil {
 			logger.Debug("AnalysisRun already exists for FreightCollection")
 
-			newVI.FinishTime = existingAnalysisRun.Status.CompletedAt()
+			newVI.FinishTime = ptr.To(metav1.Now())
 			newVI.Phase = kargoapi.VerificationPhase(existingAnalysisRun.Status.Phase)
 			newVI.AnalysisRun = &kargoapi.AnalysisRunReference{
 				Name:      existingAnalysisRun.Name,
 				Namespace: existingAnalysisRun.Namespace,
 				Phase:     string(existingAnalysisRun.Status.Phase),
 			}
-			newVI.FinishTime = existingAnalysisRun.Status.CompletedAt()
 			return newVI, nil
 		}
 	}
--- a/pkg/controller/stages/regular_stages_test.go
+++ b/pkg/controller/stages/regular_stages_test.go
@@ -3692,6 +3692,13 @@ func TestRegularStageReconciler_startVerification(t *testing.T) {
 				assert.NotEmpty(t, vi.ID)
 				assert.Equal(t, kargoapi.VerificationPhaseSuccessful, vi.Phase)
 				assert.Equal(t, "existing-analysis", vi.AnalysisRun.Name)
+				// Both timestamps should be the reconciliation time so that
+				// startTime <= finishTime always holds regardless of when the
+				// AnalysisRun internally ran.
+				require.NotNil(t, vi.StartTime)
+				require.NotNil(t, vi.FinishTime)
+				assert.Equal(t, now.Unix(), vi.StartTime.Unix())
+				assert.Equal(t, now.Unix(), vi.FinishTime.Unix())
 			},
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/controller/stages/regular_stages.go
```diff
@@ -1350,14 +1350,13 @@ func (r *RegularStageReconciler) startVerification(
 		if existingAnalysisRun != nil {
 			logger.Debug("AnalysisRun already exists for FreightCollection")
 
-			newVI.FinishTime = existingAnalysisRun.Status.CompletedAt()
+			newVI.FinishTime = ptr.To(metav1.Now())
```
EronWright, line 1353: Would this increase the volatility of the status block, e.g. at each reconciliation?  Am asking because I don't know how often the status is updated in this way.

jessesuen: It shouldn't. This should will only update status when inserting a new VerificationInfo entry into the verificationHistory (basically only when we want record a completed verification).

thomastaylor312, line 1353: Nit: 
```suggestion
			newVI.FinishTime = new(metav1.Now())
```

thomastaylor312 (approved):

pkg/controller/stages/regular_stages.go
jessesuen, line 1341: @krancour I'm thinking that this `findExistingAnalysisRun()` is incorrect behavior. The current behavior is: if I re-promote a Freight into a Stage (and that Freight had previously failed verification), the controller will discover the existing failed AnalysisRun and ***immediately*** insert a VerificationInfo entry into the history, even though the AnalysisRun is associated with an older promotion. 

In other words, we will never create 2nd AnalysisRun for the 2nd promotion, and it is impossible to generate a new AnalysisRun if one already exists (unless I click "re-verify").

What is the purpose of skipping the creation of a new AnalysisRun if we found an existing one? Is it to avoid creating duplicates?

At the very least, I think I would like to change this so that there is always a ***new*** AnalysisRun for every **_new_** promotion. We do bookkeep the associated Promotion in an AnalysisRun:

```yaml
kind: AnalysisRun
metadata:
  annotations:
    kargo.akuity.io/promotion: staging.01ksre7yrgjkx4jgzb1re76zkt.0df205a
```

So it should be easy to filter older runs.

jessesuen: I searched the codebase for when we started looking for existing AnalysisRuns. This has been there since the very beginning:
 https://github.com/akuity/kargo/blame/b5cca09d1f05141dc109ace01cef31e1fdb4c848/internal/controller/stages/verification.go#L28-L41

I believe we did this check for existing runs in order to prevent the creation of duplicate AnalysisRuns for the same promotion. However, this doesn't account for the re-promotion case where I'm promoting something that previously failed and where the user wants a ***new*** AnalysisRun to be kicked off.

I think we should be more specific in checking for existing runs by looking at the `kargo.akuity.io/promotion` label instead of relying solely on Freight ID.

```yaml
kind: AnalysisRun
metadata:
  annotations:
    kargo.akuity.io/promotion: staging.01ksre7yrgjkx4jgzb1re76zkt.0df205a
```

pkg/controller/stages/regular_stages.go
jessesuen: I was confused about your suggestion until I learned that Go 1.26 `new()` builtin can now take expressions. TIL.

But there are 11 other places (in this file alone) that use the ptr.To() pattern, and I'll leave this for consistency until we ultimately decide to switch over.

pkg/controller/stages/regular_stages.go
jessesuen: I'm going to merge this PR for now to address the timestamp bookkeeping issue, and create a different issue to address what I feel is a bug.

PR #6373, opened 2026-05-29
fix: prevent AnalysisRun reuse across promotions of the same freight

### LINKED ISSUE #6372: Second Promotion of Freight never creates a second AnalysisRun
### Pre-submission checklist

### Description

Spawned from https://github.com/akuity/kargo/pull/6370#pullrequestreview-4385979396

The following is the current behavior of Kargo:

1. Freight `F1` is promoted into Stage, creating Promotion `P1`
2. AnalysisRun `AR1` is created and associated with Promotion `P1`
3. Freight `F1` is re-promoted into the Stage, creating Promotion `P2`
4. No AnalysisRun is created or associated with `P2` because the controller will discover `AR1` with the Stage/Freight combo, and skip the creation of an `AR2`

### Steps to Reproduce

1. Configure a stage with verification
2. Promote a Freight into the Stage for the very first time. IMPORTANT: It must be the very first time
3. An AnalysisRun will be created. Wait for verification to complete. It does not matter if the AR succeeds or fails.
4. Promote the same Freight again into the Stage
5. A second AnalysisRun will not be created. Instead, the second verification will be associated with the existing AnalysisRun created in step 3

### Expected Behavior

When I promote a second time, I expect a second verification to be created.

### Actual Behavior

After a promotion, when we start verification, we discover an existing AnalysisRun for the same Stage/Freight combination, and skip creating a new AnalysisRun.

This is surprising behavior, which I don't think was intended. Most users expect there to be a 1:1 relationship between the Verification we show in the UI, and its associated AnalysisRun. Instead, there is a Many:1 relationship.

I also searched our docs but couldn't find any explanation stating that existing AnalysisRuns for the same Stage/Freight combination will result in the AnalysisRun being skipped.

What makes things worse is that the behavior is confusing and indeterminate:
1. We'll only create a new AnalysisRuns the very first Promotion of the Freight into the Stage
2. AnalysisRuns get GCed. So if GC happened to delete the existing AR, a re-promotion of the Freight **_would_** actually generate a new AnalysisRun.
3. The existing AnalysisRun object is only associated with a single promotion (the first one):

```
kind: AnalysisRun
metadata:
  annotations:
    kargo.akuity.io/promotion: staging.01ksre7yrgjkx4jgzb1re76zkt.0df205a
```

### Kargo Version

v1.11.0 (pre-release)

### Relevant Logs

```shell

```

### Screenshots

_No response_

jessesuen:
## Description

Closes https://github.com/akuity/kargo/issues/6372

When the same freight was promoted a second time, `findExistingAnalysisRun` would find the AnalysisRun from the first promotion (matching on stage + freight collection labels) and return it as already-existing — so `startVerification` returned early and never created a new AnalysisRun for the second promotion.

The fix adds the `kargo.akuity.io/promotion` annotation as an additional filter in `findExistingAnalysisRun`. An AnalysisRun from a prior promotion of the same freight is therefore invisible to the check, and a new run is created as expected.

The promotion annotation is now always set when creating a new AnalysisRun, ensuring the filter is reliably populated on every run.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/pkg/controller/stages/regular_stages.go
+++ b/pkg/controller/stages/regular_stages.go
@@ -1292,8 +1292,9 @@ func (r *RegularStageReconciler) recordFreightVerificationEvent(
 }
 
 // startVerification starts a new verification for the Freight that is associated
-// with the Stage. If the Freight has already been verified, then no verification
-// is started unless a re-verification is requested.
+// with the Stage. If an AnalysisRun already exists for the current promotion,
+// its status is returned without creating a new run. A re-verification request
+// always creates a new AnalysisRun.
 //
 // If there is no verification configuration for the Stage, then the verification
 // is automatically considered successful and no verification is started.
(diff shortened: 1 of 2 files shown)


jessesuen:
Results of some testing I did

Screenshot 2026-05-29 at 1 58 13 AM 1

krancour:
Just calling attention to [this comment](https://github.com/akuity/kargo/issues/6372#issuecomment-4574688646) so that it's not overlooked.

hairyhum:
Pod promotions supposed to set reverify flag https://github.com/akuity/kargo/blob/main/pkg/controller/promotions/promotions.go#L626 
Something else must be going wrong here.

jessesuen:
Closing PR as this is actually a bug when using the pod-based promotion-controller and does not happen in OSS.

PR #6378, opened 2026-05-29
build: source Helm binary from official releases instead of Wolfi

EronWright:
## Problem

The published Kargo image's `helm` binary came from the Wolfi `helm~3` apk, pulled in via the apko base image (`kargo-base.apko.yaml`). [Wolfi has stopped patching Helm 3](https://github.com/orgs/wolfi-dev/discussions/77550) (they direct consumers to Helm 4), so the binary lagged at **v3.19.2** and became the single largest contributor to the Critical/High CVEs that customers (via Wiz/Grype) flag against the image.

The binary is used only by the `kustomize-build` promotion step's Helm plugin (`kustomize_builder.go` → `HelmConfig.Command = "helm"`). The other Helm steps use the `helm.sh/helm/v3` library, which is a separate (k8s-frozen) dependency.

## Change

- Remove `helm~3` from the apko base image.
- Download the Helm binary directly from Helm's official releases (`get.helm.sh`) in the Dockerfile `tools` stage, **verifying the published SHA-256**, pinned via `ARG HELM_VERSION` to **v3.21.0**.
- Copy the binary into both the published `final` image and the `back-end-dev` image used by Tilt — restoring `kustomize-build` Helm-plugin parity in local dev (it previously had no `helm`).
- Align `hack/tools` `helm.sh/helm/v3` 3.19.4 → 3.19.5 with the root `go.mod` pin.

### Why the binary is intentionally ahead of the library

The current constraint applies to the `helm.sh/helm/v3` **library** — bumping it cascades blocked Kubernetes minor upgrades until the gRPC API removal (v1.12.0). The standalone **binary** from `get.helm.sh` carries no such cascade: it changes zero Kargo compiled dependencies. Both remain on the **Helm 3** line, so chart/CLI compatibility is preserved.

## CVE impact (grype, linux/arm64)

Helm binary in isolation:

| Helm | Critical | High |
|---|---|---|
| 3.19.2 (current OSS) | 3 | 18 |
| 3.19.5 (latest 3.19 patch) | 3 | 17 |
| 3.20.2 | 1 | 7 |
| **3.21.0 (this PR)** | **0** | **1** |

Staying within minor 3.19 barely helps — ~16 of those findings are Go **stdlib** CVEs present only because Helm 3.19.x is built with an old Go toolchain. 3.21.0 (built with go1.25.10) clears them. The one remaining High is a vendored `containerd` advisory with no fixed Helm 3 release past it yet.

Full image, rebuilt on this branch: **155 → 2** total findings (15 Critical/68 High → 0 Critical/1 High). Note that a main-based rebuild also folds in current base-image packages and `main`'s dependency bumps; the largest single cluster this PR is directly responsible for is the **8 Helm-binary Criticals**.

## Testing

- `tools` stage builds; checksum verifies (`/tmp/helm.tar.gz: OK`); binary reports `v3.21.0` (go1.25.10, clean tree).
- apko base builds without `helm~3`.
- Full `final` image builds successfully.

## Follow-ups (not in this PR)

- Backport to `release-1.10` → **v1.10.5** for blocked customers.
- The residual kargo-binary Go-dep CVEs (`go-git`, `go-jose`, `grpc`, …) are the k8s-frozen set, tracked separately for v1.12.0.


hack/tools/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/oapi-codegen/oapi-codegen/v2 v2.6.0
 	github.com/swaggo/swag v1.16.6
 	golang.org/x/tools v0.42.0
-	helm.sh/helm/v3 v3.19.4
+	helm.sh/helm/v3 v3.19.5
```
krancour, line 12: Any reason not to go up to v3.21.0 here like the image does?

EronWright, line 12: Note that `v3.19.5` is already in use in the root `go.mod`.

krancour: Sorry... I see the note in the Dockerfile about the intentionality of this now, but I am still confused by that.

EronWright: It would bump the k8s libs to like 35 or 36, which I believe is blocked on grpc stuff.

Dockerfile
```diff
@@ -87,6 +100,10 @@ FROM alpine:latest AS back-end-dev
 
 RUN apk update && apk add ca-certificates git gpg gpg-agent openssh-client tini
 
+# Match the published image: source Helm (needed by the kustomize-build step's
+# Helm plugin) from the official-binary "tools" stage rather than a distro
+# package.
+COPY --from=tools /tools/helm /usr/local/bin/helm
```
krancour, line 106: This `back-end-dev` stage is the quick-n-dirty image build that Tilt is using. Shouldn't we also copy this in the `final` stage, which is what real releases are built from?

EronWright: The final stage does indeed copy it:
```dockerfile
```

The fix here is to close an apparent gap in the quick-n-dirty image (lack of a `helm` binary).

krancour: D'oh. `final` was already copying all of the tools.  🤦‍♂️ 

I'd been on the lookout for an added Helm-specific line and not seen one. I completely overlooked that catch-all.

Carry on.

hack/tools/go.mod
krancour: Right. 👍

krancour (approved):

PR #6381, opened 2026-05-29
feat(chart): make KARGO_NAMESPACE overridable for use downstream

krancour:
It's rare that anyone should ever need to use this, so omitting it from values.yaml and the docs is intentional. I anticipate ourselves being the only beneficiaries of this.


charts/kargo/templates/api/secret.yaml
```diff
@@ -9,16 +9,16 @@ metadata:
     {{- include "kargo.labels" . | nindent 4 }}
     {{- include "kargo.api.labels" . | nindent 4 }}
   {{- include "kargo.annotations" (dict "root" . "annotations" .Values.api.annotations) | nindent 2 }}
-{{- if or .Values.api.adminAccount.enabled (and .Values.api.rollouts.integrationEnabled .Values.api.rollouts.logs.enabled) }}
+{{- if or .Values.api.adminAccount.enabled (and .Values.api.rollouts.integrationEnabled .Values.api.rollouts.logs.enabled) .Values.api.extraStringData }}
```
EronWright: seems a bit awkward

krancour: Not worth discussing... I had this as a draft because I was so unsure of these being changes I actually wanted to pursue. I'm repurposing the PR with a very different shape.

PR #6396, opened 2026-06-02
fix: no freight created for moved files

### LINKED ISSUE #5751: Kargo can't discover commit with moved files
# Checklist

# Description

If the commit contains only files moved to a different directory, Kargo can't discover it.

# Steps to Reproduce

1. Create a Kargo Project for the application with the manifest living in `app/` directory.
2. Create a warehouse with the  include path `/app`
3. The repo should contain a file `app/pod.yaml`
4. Create a commit that moves the file from `app/pod.yaml` to `different-app/pod.yaml`.

Expected behavior:
Kargo discovers the commit with the removed file `app/pod.yaml` and creates Freight

Actual behavior:
Kargo doesn't discover the commit and no Freight is created

Why does it happen?
Kargo uses `git show --name-only <commit>` to inspect changed files which outputs the final destination of the moved file.

# Version

```
1.9.1
```

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5751

### Before

No freight created for commit with moved file on manual refresh

https://github.com/user-attachments/assets/8798306a-49aa-4a85-b547-5c51197df84a

### After

Freight created for commit with moved file on manual refresh

https://github.com/user-attachments/assets/6771abb5-ff90-45f9-9cdc-91dc6ecf9836

### Manifest used

```yaml
kind: ProjectConfig
metadata:
  name: kargo-demo
  namespace: kargo-demo
---
kind: Warehouse
metadata:
  name: my-warehouse
  namespace: kargo-demo
spec:
  interval: 30m
  subscriptions:
  - git:
      repoURL: git@github.com:fuskovic/testrepo.git
      branch: main
      strictSemvers: false
      includePaths:
      - app/
---
apiVersion: v1
kind: Secret
metadata:
  name: warehouse-secret
  labels:
stringData:
  repoURL: git@github.com:fuskovic/testrepo.git
  sshPrivateKey: redacted

### DIFF
--- a/pkg/controller/git/work_tree.go
+++ b/pkg/controller/git/work_tree.go
@@ -60,8 +60,10 @@ type WorkTree interface {
 	// who cloned the repo associated with this working tree.
 	HomeDir() string
 	// GetDiffPathsForCommitID returns a string slice indicating the paths,
-	// relative to the root of the repository, of any files that are new or
-	// modified in the commit with the given ID.
+	// relative to the root of the repository, of any files that are added,
+	// modified, deleted, or renamed/moved in the commit with the given ID. For
+	// renamed or moved files, both the old and new paths are included so that
+	// path-based filters can match on either side of the move.
 	GetDiffPathsForCommitID(commitID string) ([]string, error)
 	// IsAncestor returns true if parent branch is an ancestor of child
 	IsAncestor(parent string, child string) (bool, error)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/controller/git/work_tree.go [hunk 1]
```diff
@@ -399,18 +401,30 @@ func (w *workTree) Fetch(opts *FetchOptions) error {
...
+		line := scanner.Text()
+		if line == "" {
+			continue
+		}
+		// --name-status output is tab-separated: <status>\t<path> for most
+		// changes, or <status>\t<old-path>\t<new-path> for renames and copies.
+		parts := strings.SplitN(line, "\t", 3)
+		if len(parts) < 2 {
+			continue
+		}
```
krancour, line 425: Are there ever legitimate lines that wouldn't have at least two parts? I am guessing no -- in which case I would probably return an error here instead of silently ignoring the unmet expectation.

pkg/controller/git/work_tree.go
```diff
@@ -399,18 +401,30 @@ func (w *workTree) Fetch(opts *FetchOptions) error {
...
+		// For renames (R) and copies (C), include both old and new paths so
+		// that a path filter matching the source directory detects the removal.
+		if len(parts) == 3 && len(parts[0]) > 0 && (parts[0][0] == 'R' || parts[0][0] == 'C') {
```
krancour: Wouldn't this work just as well while being easier to understand?

```suggestion
		if len(parts) == 3 && (parts[0] == "R" || parts[0] == "C") {
```

Separate question from that:

I don't think I'd ever seen `C` in git diffs before. I'm in no way questioning that it is possible. It's just a TIL. But my follow up question: If `C` indicates a copy, wouldn't the diff only be the destination path? Semantically speaking, copying doesn't sound like an operation that would have modified the source.

krancour:
Couple things here that are more than nits, but very small. LGTM after they're addressed. Overall, looks great and is a meaningful improvement.

pkg/controller/git/work_tree.go
fuskovic: > If C indicates a copy, wouldn't the diff only be the destination path? Semantically speaking, copying doesn't sound like an operation that would have modified the source.

That's true and I've made that adjustment but the suggested code as is breaks the tests.

There's a similarity score that gets appended to "R" e.g.

with-similarity-score

Which is why the original code uses `parts[0][0] == 'R'` instead of `parts[0] == 'R'`.

I ended up just going with `strings.HasPrefix`.

EronWright (approved):
LGTM

PR #6398, opened 2026-06-02
feat: support multiple authorized Stages in Argo CD Application annotation

### LINKED ISSUE #6027: Support multiple stages in Application's `authorized-stage` annotation
# Proposed Feature

Currently, the `kargo.akuity.io/authorized-stage` annotation that we set on an Application can only map to a single stage. This prevents multiple Stages from performing an `argocd-update` against the same application.

A long, long time ago, we actually **_used to_** support things like wildcards in `kargo.akuity.io/authorized-stage`, but had to be removed [on performance & technical grounds](https://github.com/akuity/kargo/issues/2971#issuecomment-2489614247). 

One thing we need to check is whether we still rely on this annotation to optimize lookups for which stage/promotion to refresh when application health changes. If we do, then we may only be able to support explicit lists:

`kargo.akuity.io/authorized-stage: project1:stage1,project1:stage2`

but not wildcards:

`kargo.akuity.io/authorized-stage: project1:*`

NOTE: This issue has been [raised before](https://github.com/akuity/kargo/issues/2971), but I am filing a new one because some of the previous discussion points are two years old and may no longer be relevant. 

# Motivation

There are use cases where stages don't necessarily represent an environment, but rather a process of their promotion pipeline.

EronWright:
## Description

Closes #6027

The `kargo.akuity.io/authorized-stage` annotation on an Argo CD `Application` previously mapped to exactly one Stage, preventing multiple Stages from running `argocd-update` against the same `Application`.

This PR makes the annotation accept a comma-separated list of `<project>:<stage>` entries. A single `<project>:<stage>` value remains valid and authorizes exactly one Stage.

- Adds `kargoapi.AuthorizedStages`, which parses the annotation into a set of project/stage references, trims whitespace, ignores empty entries, and rejects malformed entries and deprecated glob expressions (`*`).
- `argocd-update` authorization (`authorizeArgoCDAppUpdate`) now authorizes the step if its Stage matches **any** entry in the list.
- The Stage enqueuer that reacts to Argo CD `Application` health/sync changes now enqueues **every** authorized Stage (skipping control-flow Stages and tolerating per-entry not-found).
- Documentation updated (annotations reference + Argo CD integration how-to), including a note about the conflicting-desired-state foot-gun when multiple Stages manage one `Application`.

Wildcards are intentionally **not** supported: the `Application` → `Stage` reverse lookup relies on the annotation being an explicit, enumerable set of references.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:


api/v1alpha1/authorized_stage.go
```diff
@@ -0,0 +1,50 @@
...
+// The value is a comma-separated list of "<project>:<stage>" entries,
+// permitting more than one Stage to be authorized. A single "<project>:<stage>"
+// value (the historical format) parses to a single-element list. The returned
+// NamespacedNames use the project as the Namespace and the Stage name as the
+// Name.
+//
+// Whitespace around entries and around the project and Stage of each entry is
+// ignored. Deprecated glob expressions (entries containing "*") are rejected
+// with an error, as are malformed or empty entries.
+func AuthorizedStages(value string) ([]types.NamespacedName, error) {
```
jessesuen: This would be the first and only stand-alone function we have in our types package. The rest are methods on our structs. I think there may be a better shared utility package we can move this to. Probably pkg/api/stage.go since that already has some helper functions on annotations.

jessesuen (commented):
Logic looks good. Just have a concern about location of the helper.

api/v1alpha1/authorized_stage.go
EronWright: Moved it to `pkg/api/stage.go` alongside the other Stage annotation helpers and updated the callers (`ce69a902`).

api/v1alpha1/annotations.go
```diff
@@ -34,8 +34,8 @@ const (
 	// resource to indicate that one or more Stages are authorized to manage it.
 	// The value of the annotation is a comma-separated list of
 	// "<project>:<stage>" entries. A single "<project>:<stage>" value is also
-	// valid and authorizes exactly one Stage. Use AuthorizedStages to parse the
-	// value.
+	// valid and authorizes exactly one Stage. The value can be parsed with the
+	// AuthorizedStages helper in pkg/api.
```
jessesuen: I don't think it's worth mentioning how to parse this in our Go packages.

```suggestion
	// valid and authorizes exactly one Stage.
```

jessesuen (approved):

PR #6404, opened 2026-06-04
feat: subpath deployment support, plus assorted chart improvements

### LINKED ISSUE #6090: feat: native subpath deployment support (api.basePath)
Proposed Feature

  Native subpath deployment support via a api.basePath (or equivalent) Helm value, allowing Kargo to be served at a URL prefix such as https://platform.example.com/kargo/
   instead of requiring a dedicated hostname.

  Motivation

  Many platform teams consolidate internal tools under a single hostname using path-based routing (e.g. via Kubernetes Gateway API or an ingress controller):

  https://platform.example.com/argocd/   → ArgoCD
  https://platform.example.com/kargo/    → Kargo
  https://platform.example.com/grafana/  → Grafana

  This is a common pattern when operating on a single ALB/ingress with TLS termination and a single wildcard or SAN certificate.

  Currently Kargo is not deployable at a subpath because:

  1. The UI (React SPA) is built without a configurable base path. All asset references (/assets/, /fonts/, logo images) and React Router navigation use absolute paths
  rooted at /, so the app breaks when served under a prefix.
  2. The OIDC redirect URI is constructed as https://{api.host}/login with no way to inject a path prefix, so the callback always lands at the root — conflicting with
  other services on the same hostname.
  3. The API server does not expose a base path option, so gRPC-web calls from the UI (/akuity.io.kargo.service.v1alpha1.KargoService/...) are absolute and fail when the
  proxy strips the prefix.

  We attempted to work around this with an nginx reverse proxy doing:
  - sub_filter rewrites for HTML/CSS/JS asset paths
  - JavaScript injection into <head> to intercept history.pushState / replaceState and window.addEventListener("popstate") so React Router sees clean paths while the
  browser URL retains the /kargo/ prefix
  - fetch / XMLHttpRequest interceptors for API calls

  This is extremely fragile — it breaks across Kargo upgrades (hashed bundle filenames change), depends on nginx sub_filter_types including text/plain because kargo-api
  serves CSS with the wrong Content-Type, and required patching Location.prototype.pathname which is non-configurable in Chrome.

  The correct fix is native support, similar to what ArgoCD provides via --basehref / server.rootpath.

  Suggested implementation:

  api:
    basePath: "/kargo"   # default: "/"

  This would configure:
  - Vite base at build time (or a runtime <base href> injection)
  - The OIDC redirect URI as https://{host}{basePath}/login
  - The API server to strip the prefix from incoming requests or mount handlers under it

  kargo version

  version: v1.9.6
  buildDate: 2026-04-14T01:34:27Z
  gitCommit: 62b544749e958be17cb109445348783c2073dd67
  goVersion: go1.25.5
  platform: linux/amd64

krancour:
All of these improvements are things we will use in one capacity or another downstream from here, but each is valuable in and of itself.

- Add `global.kargoNamespace` to override the install/control-plane namespace independently of the Helm release namespace.

- Add chart-side support for proxying to an externally-managed Dex server. Useful if you want to BYO dex, but still have the Kargo API server integrate with it as if it were an internal component.

- Expose CronJob/Job knobs (concurrency, history, restart policy) on the garbage collector.

- Add per-resource controller fullname helpers and a `fullnameSuffix` knob so multiple controllers can coexist in one cluster.

- Drop a stray trailing comma in `ARGOCD_URLS` when the configured list ends up with a single entry. The parsing was forgiving, so this is largely cosmetic.

- Serve Kargo behind a configurable URL path prefix: new `api.basePath` and `externalWebhooksServer.basePath` chart values, with corresponding wiring in the API server, external webhooks server, UI, and local dev setup. Closes #6090.

- Drop the unused `ENABLE_PROMOTION_RECONCILER` controller toggle (no consumer in the chart; defaulted to true).


charts/kargo/templates/_helpers.tpl
```diff
@@ -6,6 +6,40 @@ Expand the name of the chart.
...
+
+{{/*
+Per-resource fullname helpers for the controller subsystem. Each composes
+the resource's natural name with the chart-name prefix and the optional
+suffix. Resources reference these directly (rather than deriving sibling
+names by string concatenation) so the suffix lands at the end of every
+name uniformly.
+*/}}
+{{- define "kargo.controller.fullname" -}}
+{{- printf "%s-controller%s" (include "kargo.name" .) (include "kargo.controller.fullnameSuffix" .) | trunc 63 | trimSuffix "-" -}}
```
thomastaylor312: This should work just fine, but just calling out that we could run into problems if the fullnameSuffix is too long. For example if I had two names that were something like `a-super-duper-very-long-because-i-like-java-function-style-naming-impl-{cluster1,cluster2}`. That could get trimmed to `a-super-duper-very-long-because` and you'd have two controller deployments with the same name. They probably _won't_ be running in the same place, but could be worth noting in the comment here or on `fullnameSuffix` in the values that it should be kept relatively short, and, if longer, identifying information should be at the front of the suffix string

pkg/server/option/auth.go
```diff
@@ -157,9 +161,20 @@ func getKeySet(ctx context.Context, cfg config.ServerConfig) (oidc.KeySet, error
...
+			return nil, fmt.Errorf(
+				"error parsing OIDC issuer URL %q: %w",
+				cfg.OIDCConfig.IssuerURL,
+				err,
+			)
+		}
+		// Dex routes its endpoints based on the path of its configured issuer
+		// URL, which includes the API server's basePath when one is set. Use
+		// that same path with the in-cluster Dex address so paths line up.
+		dexBaseAddr = cfg.DexProxyConfig.ServerAddr + issuerURL.Path
```
thomastaylor312, line 175: Does this path need the `/dex` suffix? That used to be added below, but isn't any more so I wanted to double check

pkg/server/auth_middleware.go
```diff
@@ -479,9 +483,20 @@ func getKeySet(ctx context.Context, cfg config.ServerConfig) (oidc.KeySet, error
...
+			return nil, fmt.Errorf(
+				"error parsing OIDC issuer URL %q: %w",
+				cfg.OIDCConfig.IssuerURL,
+				err,
+			)
+		}
+		// Dex routes its endpoints based on the path of its configured issuer
+		// URL, which includes the API server's basePath when one is set. Use
+		// that same path with the in-cluster Dex address so paths line up.
+		dexBaseAddr = cfg.DexProxyConfig.ServerAddr + issuerURL.Path
```
thomastaylor312, line 497: Same note here as the other comment around the dex path

thomastaylor312 (approved):
All good to go from the backend side. Probably want a UI person to look at the TS changes

pkg/server/option/auth.go
krancour: When you set Dex's issuer URL, you're setting Dex's own base path.

So if the issuer URL is something like `https://kargo.example.com/basepath/dex`, the Dex server's own base path is `/basepath/dex`.

For this bit of API server code to derive an _internal_ URL for Dex's well-known config endpoint, all it needs to do is swap `https://kargo.example.com` for `https://kargo-dex-server.kargo.svc` (for instance), leave the issuer path unchanged, then tack on `/.well-known/openid-configuration`, yielding `https://kargo-dex-server.kargo.svc/basepath/dex/.well-known/openid-configuration` -- without `dex` having been hard coded anywhere.

charts/kargo/templates/_helpers.tpl
krancour: Moot because I reverted this piece of the PR and want to approach this differently in a follow-up.

ui/index.html
```diff
@@ -2,7 +2,22 @@
...
+      The window.__KARGO_BASE_PATH__ global (no trailing slash) is read
+      by JS code that constructs URLs explicitly — ConnectRPC/REST
+      baseUrl, React Router basename, window.location redirects.
+      When the server is at the root, both placeholders resolve to their
+      no-op forms ("/" and "" respectively).
+    -->
+    <base href="__BASE_HREF__" />
+    <script>
+      window.__KARGO_BASE_PATH__ = '__BASE_PATH__';
+    </script>
```
Marvin9, line 19: If I am understanding this correctly then the point of this snippet is to override the value on the build time so eventually UI can access via `window.__KARGO_BASE_PATH`.

I noticed the custom vite plugin that transforms this value.

This can be solved by vite global variables directly - https://vite.dev/guide/env-and-mode

pkg/server/server.go
```diff
@@ -342,3 +412,28 @@ func newDashboardRequestHandler(uiFS fs.FS) http.HandlerFunc {
...
+		return nil, time.Time{}
+	}
+	baseHref := "/"
+	basePathValue := ""
+	if basePath != "" {
+		baseHref = basePath + "/"
+		basePathValue = basePath
+	}
+	rendered := bytes.ReplaceAll(raw, []byte(indexHTMLBasePlaceholder), []byte(baseHref))
+	rendered = bytes.ReplaceAll(rendered, []byte(indexHTMLBasePathPlaceholder), []byte(basePathValue))
```
Marvin9, line 434: ~we can just use vite global variables to remove this approach.. as mentioned in previous comment.~

https://github.com/akuity/kargo/pull/6404/changes#r3362719054

ui/index.html
Marvin9: Oh now I know why what I suggested wouldn't be possible. The base path _can_ change at runtime.

What do you think if we expose this base path in public config API?

krancour: > override the value on the build time

Not this.

At build time, we don't know the server's address, and the one image that's produced needs to generically work for anyone who installs it, so no address is baked in at build time -- neither before nor after this change.

Historically, this wasn't a problem. The API server _also_ serves up the UI. So you point your browser at `https://kargo.example.com` and the UI, which needs to know the address of the API server, can infer it simply by looking at `window.location` -- done.

With support for a configurable base path added, Kargo might be hosted at an address like `https://example.com/kargo`. If you point your browser at that and the UI needs to infer the address of the API server, the `window.location` trick no longer works. It doesn't work because there' no way to know if the API is served from `https://example.com/kargo` OR if it's served from `https://example.com` and `/kargo` is just a path to a specific API endpoint or UI view. 🤷‍♂️

So there needs to be another way...

What this change does is injects what the API server knows about its own location into `index.html` _at startup_. Every request for `index.html` is served those modified bytes. All other UI assets continue to be served identically to how they were before.

> The base path can change at runtime.

Sort of. Base path is injected into the `index.html` bytes at runtime, yes, but as noted earlier, it happens only once, at startup. The base path cannot change in the lifetime of that container.

PR #6406, opened 2026-06-04
fix: empty actor for promos created with kargo api tokens

### LINKED ISSUE #6201: Promotions from kargo CLI using tokens should set ctx.meta.promotion.actor
# Checklist

* [x ] I've searched the issue queue to verify this is not a duplicate bug report.
* [ x] I've included steps to reproduce the bug.
* [ x] I've pasted the output of `kargo version`.
* [x ] I've pasted logs, if applicable.

# Description

When a promotion is created by kargo CLI using a token instead of a user, the field `ctx.meta.promotion.actor` in the promotion task is empty.

# Steps to Reproduce

My CICD workflow on Github does basically this

```
KARGO_API_ADDRESS=https://kargo.domain
KARGO_API_TOKEN=...token...
kargo promote --project=myproject --stage=dev --freight=$newFreightid --wait
```

The Promotion Task is running this step

```
    - uses: git-commit
      config:
        message: "${{ctx.project}}: ${{ctx.meta.promotion.actor}} promoted ${{ctx.targetFreight.origin.name }} to ${{ctx.stage}}"
```

While promotions done by real persons using the UI contain an actor, and  my previous CICD pipeline where kargo CLI used "kargo login" to authenticate also had the actor set, my new pipeline using the new API token feature from version 1.9 does not set this variable any more. It's just empty.

It would expect for the actor to contain the name of the token I used when running "kargo create token".

If somehow it's not possible to transport the name info from the token into the context, I would suggest to be able to set this name manually in the kargo CLI, for example like this
```
kargo promote --actor="github deployment" ...
```

# Version

```
$ kargo version
Client Version: v1.10.3
Server Version: v1.10.3
```

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6201

### Before

To reproduce the issue I:
1. created a new test token.
2. added it to my kargo config
3. created a promotion with an http step that sends a POST request to webhook site to check the actor.
4. verified it was empty.

https://github.com/user-attachments/assets/e1e8b157-bd0b-4fe6-8f2f-bcfa30e8eed3

### After

Verified the actor was populated after adding parsing for service account username from the token.

https://github.com/user-attachments/assets/5b530aff-aac2-4db0-9d8e-27537507e1fc

### DIFF
--- a/pkg/server/auth_middleware.go
+++ b/pkg/server/auth_middleware.go
@@ -244,6 +244,10 @@ func (a *authMiddleware) authenticate(
 // account name extracted from the standard "sub" claim.
 func serviceAccountUsernameFromToken(rawToken string) (username, usernameClaim string) {
 	mapClaims := jwt.MapClaims{}
+	// #nosec G115 -- The token has already been verified by Kubernetes before
+	// this function is called; Kubernetes validates the JWT signature, so any
+	// claims present are trustworthy. The extracted username is used only for
+	// the create-actor display annotation and plays no role in authorization.
 	if _, _, err := jwt.NewParser(jwt.WithoutClaimsValidation()).
 		ParseUnverified(rawToken, &mapClaims); err != nil {
 		return "", ""
--- a/pkg/server/auth_middleware.go
+++ b/pkg/server/auth_middleware.go
@@ -225,12 +225,43 @@ func (a *authMiddleware) authenticate(
 	}
 	logger.Debug("token recognized by Kubernetes")
 
+	username, usernameClaim := serviceAccountUsernameFromToken(rawToken)
 	return user.ContextWithInfo(
 		ctx,
 		user.Info{
-			BearerToken: rawToken,
+			BearerToken:   rawToken,
+			Username:      username,
+			UsernameClaim: usernameClaim,
 		},
 	), nil
+}
+
+// serviceAccountUsernameFromToken extracts a meaningful username from a
+// Kubernetes service account JWT. For Kargo API tokens (legacy
+// ServiceAccountToken-type Secrets), the JWT contains the secret name in the
+// "kubernetes.io/serviceaccount/secret.name" claim, which is the name the user
+// chose when running "kargo create token --name=...". Falls back to the service
+// account name extracted from the standard "sub" claim.
+func serviceAccountUsernameFromToken(rawToken string) (username, usernameClaim string) {
+	mapClaims := jwt.MapClaims{}
+	if _, _, err := jwt.NewParser(jwt.WithoutClaimsValidation()).
+		ParseUnverified(rawToken, &mapClaims); err != nil {
+		return "", ""
+	}
+	if secretName, ok := mapClaims["kubernetes.io/serviceaccount/secret.name"]; ok {
+		if name, ok := secretName.(string); ok && name != "" {
+			return name, "serviceaccount"
+		}
+	}
+	// Fall back to the service account name from the sub claim.
+	// Sub format: "system:serviceaccount:<namespace>:<name>"
+	sub, _ := mapClaims["sub"].(string)
+	if strings.HasPrefix(sub, "system:serviceaccount:") {
+		if parts := strings.SplitN(sub, ":", 4); len(parts) == 4 && parts[3] != "" {
+			return parts[3], "serviceaccount"
+		}
+	}
+	return "", ""
 }
 
 func (a *authMiddleware) listServiceAccounts(
--- a/pkg/server/option/auth.go
+++ b/pkg/server/option/auth.go
@@ -454,12 +454,43 @@ func (a *authInterceptor) authenticate(
 	}
 	logger.Debug("token recognized by Kubernetes")
 
+	username, usernameClaim := serviceAccountUsernameFromToken(rawToken)
 	return user.ContextWithInfo(
 		ctx,
 		user.Info{
-			BearerToken: rawToken,
+			BearerToken:   rawToken,
+			Username:      username,
+			UsernameClaim: usernameClaim,
 		},
 	), nil
+}
+
+// serviceAccountUsernameFromToken extracts a meaningful username from a
+// Kubernetes service account JWT. For Kargo API tokens (legacy
+// ServiceAccountToken-type Secrets), the JWT contains the secret name in the
+// "kubernetes.io/serviceaccount/secret.name" claim, which is the name the user
+// chose when running "kargo create token --name=...". Falls back to the service
+// account name extracted from the standard "sub" claim.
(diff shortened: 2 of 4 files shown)


pkg/server/auth_middleware.go
```diff
@@ -225,14 +225,45 @@
...
+// serviceAccountUsernameFromToken extracts a meaningful username from a
+// Kubernetes service account JWT. For Kargo API tokens (legacy
+// ServiceAccountToken-type Secrets), the JWT contains the secret name in the
+// "kubernetes.io/serviceaccount/secret.name" claim, which is the name the user
+// chose when running "kargo create token --name=...". Falls back to the service
+// account name extracted from the standard "sub" claim.
+func serviceAccountUsernameFromToken(rawToken string) (username, usernameClaim string) {
+	mapClaims := jwt.MapClaims{}
+	if _, _, err := jwt.NewParser(jwt.WithoutClaimsValidation()).
+		ParseUnverified(rawToken, &mapClaims); err != nil {
```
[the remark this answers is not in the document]
fuskovic: This is a false-positive `verifyKubernetesTokenFn` runs literally right before `serviceAccountUsernameFromToken` is called ( in both places ). Added a `nolint` directive for this.

pkg/server/option/auth.go
```diff
@@ -454,14 +454,49 @@ func (a *authInterceptor) authenticate(
...
 	), nil
 }
 
+// serviceAccountUsernameFromToken extracts a meaningful username from a
+// Kubernetes service account JWT. For Kargo API tokens (legacy
+// ServiceAccountToken-type Secrets), the JWT contains the secret name in the
+// "kubernetes.io/serviceaccount/secret.name" claim, which is the name the user
+// chose when running "kargo create token --name=...". Falls back to the service
+// account name extracted from the standard "sub" claim.
+func serviceAccountUsernameFromToken(rawToken string) (username, usernameClaim string) {
```
hairyhum, line 474: This is exactly the same as in `auth_middleware.go`, is duplication intentional here?

fuskovic: The duplication exists because `auth_middleware.go` (package server) and `option/auth.go` (package option) are separate packages that both happen to implement the same auth logic. So yes, it's intentional for now, but for the same reason the rest of the auth logic is duplicated between these two files. Since the ConnectRPC stack (`option/`) is deprecated and slated for removal in `v1.12.0`, extracting a shared helper would add complexity that gets deleted soon anyway.

pkg/server/option/auth.go
```diff
@@ -454,14 +454,49 @@ func (a *authInterceptor) authenticate(
...
+		}
+	}
+	// Fall back to the service account name from the sub claim.
+	// Sub format: "system:serviceaccount:<namespace>:<name>"
+	sub, _ := mapClaims["sub"].(string)
+	if strings.HasPrefix(sub, "system:serviceaccount:") {
+		if parts := strings.SplitN(sub, ":", 4); len(parts) == 4 && parts[3] != "" {
+			return parts[3], "serviceaccount"
+		}
+	}
```
EronWright, line 496: Can you explain why the `secret.name` would be used, and whether (in any case) the resultant username would include the namespace?  How are these values compatible with the normal usernames?  Can I spoof someone's username by creating a service account?

EronWright (changes requested):

pkg/server/option/auth.go
fuskovic: > Can you explain why the `secret.name` would be used, and whether (in any case) the resultant username would include the namespace? 

It's the name the user would/should expect to see. When a user runs `kargo create token --name=my-token` K8s embeds that secret name as the `kubernetes.io/serviceaccount/secret.name` claim in the JWT. 

Without this, we'd fall back to the underlying service account name (e.g. `kargo-api`) which is the same for all Kargo API tokens and gives no useful context.

>How are these values compatible with the normal usernames? 

No. The extracted username is just the bare secret name (e.g. `my-token`) or bare SA name (e.g. `kargo-api`). The resulting actor annotation is formatted as `serviceaccount:my-token` by `formatOIDCUsername` here: https://github.com/akuity/kargo/blob/main/pkg/api/event.go#L28.

Namepsaces are omitted, they're not part of the actor display value for OIDC users either (which also produce `<claim>:<value>` with no namespace).


>Can I spoof someone's username by creating a service account?

No, because:

1. The username is purely cosmetic. it appears only in the `kargo.akuity.io/create-actor` annotation used for display in promotion history. It plays no role in authorization anywhere. The authorizing client uses the raw `BearerToken` for SSAR, not the Username field.

2. To forge the actor, an attacker would need to create a Secret named identically to someone else's token (e.g. `my-token`) in the same Kargo project namespace. That requires create/write access to that namespace. At that point they already have equivalent or greater access than the token they're impersonating.

pkg/server/option/auth.go
```diff
@@ -454,14 +454,49 @@ func (a *authInterceptor) authenticate(
...
+			BearerToken:   rawToken,
+			Username:      username,
+			UsernameClaim: usernameClaim,
 		},
 	), nil
 }
 
+// serviceAccountUsernameFromToken extracts a meaningful username from a
+// Kubernetes service account JWT. For Kargo API tokens (legacy
+// ServiceAccountToken-type Secrets), the JWT contains the secret name in the
```
krancour, line 470: Confused by the "legacy" bit here. It's possibly my memory is failing me, but I don't remember us ever issuing API tokens differently than the way we currently do.

EronWright (changes requested):
Suggest closing this.

pkg/server/option/auth.go
krancour: I don't entirely understand the notion of going after the name of the Secret first and falling back on the SA name. Attributing creation of something to the _token_ instead of the _principal_ the token belongs to seems odd. If it were a human who'd created something, you'd want to see the human's name (or GitHub handle, or email address, or whatever); _not_ an identifier for the token they authenticated with. I'm not sure why that should be different when the user isn't human.

Just to prove that point...

> The resulting actor annotation is formatted as `serviceaccount:my-token`

Isn't it a bit odd for `serviceacccount:` to be followed by the name of something that is _not_ a SA, but a Secret?

imho, the whole PR needs a rethink. I think it's worth considering adding a separate field to `user.Info` that's a boolean SA/not SA or indicates a principal type (e.g. "human" or "bot"). I would make the authn middleware extract SA name (not token name) from the claims, treat that as the username, and set the value of the name field appropriately, then teach `FormatEventUserActor()` (not `formatOIDCUsername()` as there isn't really any OIDC involved) to format that nicely.

fuskovic:
Will close and take another stab at it when I get back next week.

PR #6410, opened 2026-06-04
ci: scan the kargo image for CVEs with Grype at build and release time

EronWright:
Adds `hack/cve-scan.sh` — the CVE scan recipe used across Akuity image pipelines — and wires it into the two places the kargo image is built:

| Workflow / job | Scan source |
|---|---|
| `ci.yaml` / `build-image` | job-local registry service (the multi-platform build previously stayed only in buildx cache with `push: false`; it now also pushes to the job's `registry:3.0.0` service — nothing leaves the runner) |
| `release.yaml` / `publish-image` | the just-pushed digest, right after signing |

Behavior:

- **Report-only**: findings table in the job step summary + a `::warning::` annotation with Critical/High counts. `FAIL_ON: ""` at each call site; setting it to `high` later turns the scan into a hard gate (one-line diff).
- Fetches and applies Akuity's published VEX statements (OpenVEX) from `https://vex.akuity.io/pkg/oci/ghcr.io/akuity/kargo/vex.json` via `grype --vex`, so findings assessed as not-affected are suppressed. A 404 is tolerated until VEX publishing goes live — scans tighten automatically once it does.
- `-unstable` builds use the canonical `ghcr.io/akuity/kargo` path for VEX lookups, sharing the release dispositions.
- `download-grype` is pinned to a full commit SHA (v7.4.0) with DB caching.

The same recipe is live in Akuity's other image repos; this PR's own `build-image` job exercises the CI scan path end-to-end.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -266,10 +266,15 @@ jobs:
       uses: anchore/scan-action/download-grype@e1165082ffb1fe366ebaf02d8526e7c4989ea9d2 # v7.4.0
       with:
         cache-db: true
+    # cve-scan.sh re-stamps the VEX product @id to the scanned digest, which it
+    # resolves from the job-local registry tag via crane.
+    - name: Install Crane
+      uses: imjasonh/setup-crane@31b88efe9de28ae0ffa220711af4b60be9435f6e # v0.4
     - name: Scan image for CVEs
       env:
         GRYPE_CMD: ${{ steps.grype.outputs.cmd }}
         GRYPE_REGISTRY_INSECURE_USE_HTTP: "true" # job-local registry service
+        CRANE_INSECURE: "true" # job-local registry is plain HTTP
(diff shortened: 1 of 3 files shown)


EronWright:
## CI verification

The `build-image` job on this PR ran the scan end-to-end via the job-local registry (job green, report-only):

> ⚠️ CVE scan of `registry:localhost:5000/kargo:scan`: **17 critical / 14 high** (full table in the job step summary)

Context for the counts: a large batch of `golang.org/x/crypto` (fixed in 0.52.0), `x/net` (fixed in 0.55.0), and Go stdlib (fixed in 1.26.4) advisories was published on 2026-06-03→04 and accounts for much of this. The VEX 404 notice is expected until `vex.akuity.io` goes live — at which point dispositioned findings are suppressed automatically with no workflow change.

krancour (approved):

krancour:
@EronWright this might be worth backporting. wdyt?

PR #6412, opened 2026-06-05
feat(ui): project section grpc to REST API migration

### LINKED ISSUE #6282: UI: API migration to REST - project

Marvin9:
## Description

Closes https://github.com/akuity/kargo/issues/6282

### Eligibility

### Quality

- [ ] Adds or updates corresponding tests.

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:


Marvin9:
Remaining:

- [x] Refactor watch endpoints for performance (right now they are same as they were pre-performance improvement)
- [x] Analysis run module
- [x] Testing each and every affected change and cases

ui/src/features/assemble-freight/artifact-type-guards.ts
rpelczar, line 1: I think we shouldn't lose the type information along the way, so we don't have to determine the type like this later. There is a risk that we add some fields to any of these types, and it won't work as expected.
WDYT?

ui/src/extend/types.ts
rpelczar, line 1: Shouldn't we try to add those types to the backend?

ui/src/features/assemble-freight/unique-subscription-key.ts
```diff
@@ -1,21 +1,26 @@
...
 
 export const getSubscriptionKey = (res: DiscoveryResult) => {
-  if (res.$typeName === 'github.com.akuity.kargo.api.v1alpha1.DiscoveryResult') {
-    return res.name;
+  if ('artifactReferences' in res) {
+    return res.name || '';
   }
 
-  if (res.$typeName === 'github.com.akuity.kargo.api.v1alpha1.ChartDiscoveryResult') {
+  if ('versions' in res) {
```
rpelczar: Isn't it safer to check if there is a repoURL and name?

ui/src/features/project/pipelines/graph/custom-node.tsx
```diff
@@ -1,184 +1,102 @@
...
+    value: WarehouseExpanded;
     subscriptionParent?: Warehouse;
     warehouseY?: Record<string, number>;
   };
   id?: string;
 }) => {
-  const ready = useGraphContext()?.ready ?? true;
+  const graphContext = useGraphContext();
+
+  const ready = graphContext?.ready || true;
```
rpelczar: This is always true, `??` you mean?

ui/src/features/project/pipelines/graph/use-events-watcher.ts
rpelczar, line 1: I tested it, and it seems the bug I fixed here: https://github.com/akuity/kargo/pull/6440 is gone. 
Just want to confirm you have tested this as well.

rpelczar (changes requested):
Unfortunately, I noticed a significant performance drop for bigger projects. I think we should review what we can improve to achieve the same results as before the migration.

Marvin9:
> Unfortunately, I noticed a significant performance drop for bigger projects. I think we should review what we can improve to achieve the same results as before the migration.

Yes, as discussed earlier, the reason I didn't implement the improvements in UI is because we were waiting for https://github.com/akuity/kargo/pull/6311.

ui/src/features/assemble-freight/artifact-type-guards.ts
Marvin9: Thats true. I will add orval transformer so backend doesn't have to do anything

Marvin9: Ok so I checked a bit. 

In short its more like implementing the layer that inject $type, similar to how `protobuf-es` used to do.

- Guide the schema to inject type on compile time
- Stamp those type as API response arrives and before being used

I have better idea to avoid the risk you mentioned. Write unit tests that check the key we are checking do not overlap with each other.

Let me know if want to sync up for this

ui/src/extend/types.ts
Marvin9: I am trying to figure out some ways

ui/src/features/project/pipelines/graph/custom-node.tsx
Marvin9: Good catch

ui/src/extend/types.ts
Marvin9: Can we loop in some backend folks for this?

Claude suggested to have one dummy route that reference this types in order to have in swagger and eventually in typescript. but I have a feeling there must be some simpler way cc @krancour

ui/src/features/assemble-freight/artifact-type-guards.ts
rpelczar: I think it's an unnecessary complication. We are receiving artifacts grouped by type, then we are merging them without type, so you needed to add this function to determine what type it is. Let's not lose the type we have from the backend.

ui/src/extend/types.ts
rpelczar: We can also improve it later...

ui/src/features/assemble-freight/artifact-type-guards.ts
Marvin9: Can we do outside of this PR? I think not mixing up in same array would be the cleanest solution, but its also lots of changes, so I want to do it outside of this PR.

rpelczar (changes requested):
https://github.com/akuity/kargo/pull/6440 the bug I recently fixed seems to be back with your changes

ui/src/features/project/pipelines/watch-utils.ts
```diff
@@ -0,0 +1,138 @@
+import { authTokenKey } from '@ui/config/auth';
+
+export const getBaseUrl = () => (import.meta.env.VITE_API_URL as string | undefined) || '';
```
youhonglian: Do we need to prefix the SSE URL with the runtime [base URL](https://github.com/akuity/kargo/blob/84212de4db32fc45a13a6134220273847045147b/ui/src/config/base-path.ts#L26)?  [custom-fetch.ts](https://github.com/akuity/kargo/blob/84212de4db32fc45a13a6134220273847045147b/ui/src/lib/api/custom-fetch.ts#L22) and the old [transport.ts](https://github.com/akuity/kargo/blob/84212de4db32fc45a13a6134220273847045147b/ui/src/config/transport.ts#L81) both use it.

Marvin9: Correct. We should re-use that function instead of re-defining it here

ui/src/features/project/pipelines/promotion/promotion.tsx
```diff
@@ -120,7 +126,7 @@ const Content = (props: { promotion: TPromotion; yaml: string }) => {
     )
   });
 
-  const promotionStartTime = timestampDate(promotion?.metadata?.creationTimestamp);
+  const promotionStartTime = new Date(promotion?.metadata?.creationTimestamp || '');
```
youhonglian: `''` returns Invalid Date, and I think it would throw an error right after.

Marvin9: Thats true. But I wouldn't imagine any promotion without creationTimestamp or terminal promotion without finishedAt status

Marvin9:
> #6440 the bug I recently fixed seems to be back with your changes

lets sync up for this. I have a solution that refetching list queries would solve this.. the more I look at the problem, the more fundamental components feel missing..

For example, we can obviously reconnect the watch endpoint, but watch doesn't really replay. So lets say you reconnect the watch but meanwhile there are events missed due to failure or connection drop, when we reconnect watch endpoint, we certainly miss the events b/w when watch was dead.. so essentially you need to make list call to get the latest state

pkg/server/list_stages_v1alpha1.go
```diff
@@ -124,19 +127,59 @@ func (s *server) watchStages(c *gin.Context, project string, warehouses []string
...
+) (*kargoapi.StageList, error) {
+	var list kargoapi.StageList
+	if err := s.listForWatchSeed(ctx, "stages", &list, client.InNamespace(project)); err != nil {
+		return nil, err
+	}
+	list.ResourceVersion = normalizeListResourceVersion(list.ResourceVersion)
+	if len(warehouses) == 0 {
+		return &list, nil
+	}
+	var stages []kargoapi.Stage
```
fuskovic: Looks like we already have the cap here so we should use `make`:

```suggestion
	stages := make([]kargoapi.Stage, 0, len(list.Items))
```

pkg/server/list_for_watch_seed.go
```diff
@@ -0,0 +1,57 @@
...
+	resource string,
+	list client.ObjectList,
+	opts ...client.ListOption,
+) error {
+	if s.client == nil {
+		return fmt.Errorf("kubernetes client is not configured")
+	}
+	if s.authorizeFn == nil {
+		return fmt.Errorf("authorize function is not configured")
+	}
```
fuskovic, line 38: we don't have a formatting directive here so we can just use `errors.New`:
```suggestion
	if s.client == nil {
		return errors.New("kubernetes client is not configured")
	}
	if s.authorizeFn == nil {
		return errors.New("authorize function is not configured")
	}
```

pkg/server/watch_helpers_test.go
```diff
@@ -0,0 +1,136 @@
...
+			name:      "non-matching modified becomes delete",
+			eventType: watch.Modified,
+			wantType:  watch.Deleted,
+			wantSend:  true,
+		},
+		{
+			name:      "non-matching added is skipped",
+			eventType: watch.Added,
+		},
+	}
```
fuskovic, line 143: The tests cover:
1. matching modified
2. non-matching modified
3. non-matching added

So it looks like we're missing coverage for:

1. matching added
2. matching deleted
3. non-matching deleted

fuskovic (changes requested):

ui/src/features/stage/verifications.tsx
```diff
@@ -81,7 +80,7 @@ export const Verifications = ({ verifications, images }: Props) => {
           title='Date'
           width={220}
           render={(_, verification) => {
-            const date = timestampDate(verification.startTime);
+            const date = new Date(verification.startTime || '');
             return date ? format(date, 'MMM do yyyy HH:mm:ss') : '';
```
rpelczar: Potentially crash. new Date('') returns `Invalid Date`

ui/src/features/stage/verifications.tsx
```diff
@@ -93,8 +92,8 @@ export const Verifications = ({ verifications, images }: Props) => {
             }
 
             try {
-              const startTime = moment(timestampDate(verification.startTime));
-              const finishTime = moment(timestampDate(verification.finishTime));
+              const startTime = moment(verification.startTime);
+              const finishTime = moment(verification.finishTime);
```
rpelczar: if undefined, moment returns now. You should do something like this: `moment(verification.finishTime || null)`

rpelczar (changes requested):

ui/src/features/stage/verifications.tsx
rpelczar: done

ui/src/features/stage/verifications.tsx
rpelczar: done

pkg/server/watch_helpers_test.go
Marvin9: Done

pkg/server/list_for_watch_seed.go
Marvin9: Done

pkg/server/list_stages_v1alpha1.go
Marvin9: Done

ui/src/features/stage/verifications.tsx
Marvin9: Thanks for the changes. I wouldn't worry though since its already there. We loose that context in type generation

rpelczar (approved):
UI LGTM

fuskovic (approved):
Backend LGTM

PR #6414, opened 2026-06-05
feat: errorExpression for http steps (#6356)

### LINKED ISSUE #6356: `errorExpression` in http steps for richer error messages
### Pre-submission checklist

### Proposed Feature

It would be great to surface richer error messages from http steps, right now they only get the status code.

```yaml
- uses: http
  as: cat-facts
  retry:
    timeout: 10m
  config:
    method: GET
    url: https://www.catfacts.net/api/
    successExpression: response.status == 200
    failureExpression: response.status == 400
    terminalFailureErrorExpression: response.body.errorText
```

Suppose our cat facts api return some `errorText`. If the failure is terminal (`failureExpression` evaled to true) then `terminalFailureErrorExpression` will extract something from the response that can be used to provide a richer error message in the UI.

`errorExpression` is shorter, but we wouldn't want to confuse it with errors from retries timing out, thus I have used `terminalFailureErrorExpression` in the example.

`timeoutMessage` would also be useful to surface docs on how to resolve a particular issue if a polling step fails.

Happy to work on this if its accepted :)

### Motivation and Use Case

Users currently see the error code for a `http` step failure, but have no infomation to act on this. An API may respond with an error message, but there is no way to surface this to the user.

### Alternatives Considered

Using the `fail` step, pretending like `http` steps always succeed, then extracting the error message from a steps outputs.

Kind of works, but you loose retries which are important for us.

zoeyfyi:
## Description

Closes #6356

Adds `errorExpression` to a `http` steps config. If a `http` step reaches a terminal failure (i.e. `failureExpression` evaluates to true) we attempt to extract an error message from the response by evaluating `errorExpression`. If evaluation fails, fall back to the original error message, otherwise we add the extracted error message to the message to the promotion step result.

Screenshot 2026-06-01 at 14 17 57

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:


krancour:
Thank you @zoeyfyi! I only looked at docs so far -- which look good overall. My only ask would be to see if this feature can be worked into an example. As you had called out in the original issue, expr lang features like nil coalescing play a particularly important role here. Or expressions that check response code first to determine if a body from which to extract information may even exist may also be realistic.

@fuskovic would you mind reviewing the implementation please?

krancour:
_Tentatively_ in v1.11.0. It's going to depend on our review bandwidth.

fuskovic (changes requested):
Can we get a test for the "did not evaluate to a string" case? Looks like an easy coverage win. Otherwise looks solid!

EronWright:
Reviewed this with an eye toward landing it in 1.11 — looks solid overall. I took the liberty of pushing a small follow-up commit (`10bdadd`) to distinguish the `errorExpression` failure modes; happy to drop or amend it if you'd prefer.

The motivation: re-reading the discussion in #6356, @krancour endorsed specifically "ignore this if it evaluates to `nil`" — the *nil-result* case. The original implementation funneled four distinct outcomes into the same silent trace-log-and-fallback path, which meant (a) the blessed `nil` case was actually flowing through the "did not evaluate to a string" *error* branch, and (b) a genuine misconfiguration (an expression that doesn't compile) was buried at trace level with no feedback. So we split them by intent:

- **nil / non-string / empty result** → fall back to the default message, no error, no log. This is the expected, endorsed behavior (e.g. nil coalescing that finds no matching field).
- **evaluation error** (response not shaped as the expression expects) → best-effort; log at `debug` and fall back. This is the "fails to evaluate, just ignore it" case, now at least discoverable.
- **compile error** (a malformed expression, e.g. `(1 + 2`) → surfaced as a `TerminalError`, consistent with how `successExpression`/`failureExpression` already treat compile failures. A non-compiling expression is a static config bug the user should fix, not something to silently swallow.

Also updated the reference-doc row to match (it previously said compile failures fall back to the default), added a non-string-result test, and fixed a missing trailing newline in `http.md`.

fuskovic (approved):

EronWright:
Thanks again @zoeyfyi for the contribution to Kargo!

PR #6440, opened 2026-06-09
fix(ui): reconnect pipeline watch on tab visibility instead of canceling

### LINKED ISSUE #6439: Pipeline watch stops collecting updates when the browser tab is inactive
### Description

The pipeline view watches Stages and Warehouses over a streaming connection.
When the browser tab is inactive, the tab is throttled/frozen by the browser
and the watch stops collecting data. The stream can also be silently dropped
while the tab is hidden, and there is no reconnect - so the data stays stale
even after returning to the tab.

### Current behavior

- The watch is torn down when the tab loses visibility and rebuilt when it
  regains it, causing connection churn on every tab switch.
- If the stream dies while the tab is hidden, it is not re-established.

### Expected behavior

- The watch keeps running while the tab is inactive.
- When the tab becomes visible again, the watch reconnects to guarantee a
  live stream.

rpelczar:
### Summary

Keeps the pipeline Stages/Warehouses watch alive while the tab is inactive,
and reconnects when the tab becomes visible again instead of tearing the
watch down on every tab switch.

### Changes

- Do not cancel the watch when the tab loses visibility
- Reconnect the watch when the tab becomes visible again (heals a stream that
  was throttled, frozen, or silently dropped while hidden)
- Watch is still cleaned up on unmount and when project/warehouses change

Closes #6439

### DIFF
--- a/ui/src/features/project/pipelines/graph/use-events-watcher.ts
+++ b/ui/src/features/project/pipelines/graph/use-events-watcher.ts
@@ -4,7 +4,6 @@ import { useEffect } from 'react';
 import { Watcher } from '@ui/features/project/pipelines/watcher';
 import { queryCache } from '@ui/features/utils/cache';
 import { Stage, Warehouse } from '@ui/gen/api/v1alpha1/generated_pb';
-import { useDocumentEvent } from '@ui/utils/document';
 
 export const useEventsWatcher = (
   project: string,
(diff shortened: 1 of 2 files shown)


ui/src/features/project/pipelines/graph/use-events-watcher.ts
```diff
@@ -15,26 +14,42 @@ export const useEventsWatcher = (
...
 
-    watcher.watchStages(act?.onStage, warehouses);
-    watcher.watchWarehouses({
-      onWarehouseEvent: act?.onWarehouse,
-      refreshHook: queryCache.freight.refetchQueryFreight
-    });
+    // (Re)establish the watch streams. Aborts any previous watcher first so we
+    // never leak a connection.
+    const connect = () => {
+      watcher?.cancelWatch();
```
Marvin9, line 28: wondering why we need to cancel and cannot just re-use?

Marvin9 (approved):
LGTM.

PR #6444, opened 2026-06-10
fix: allow Kubernetes namespace GC to delete replicated secrets

### LINKED ISSUE #6443: replicated secrets cannot be deleted by namespace/gc controller
### Pre-submission checklist

### Description

I tried to delete a Project that had a replicated secret. The Project resource was deleted, but the underlying namespace remained. In the conditions, it had the following ContentDeletionFailed error:

```
replicated resources are managed by Kargo and cannot be created, modified, or deleted by end users
```

```yaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: "2026-03-30T23:48:29Z"
  deletionTimestamp: "2026-06-10T22:08:18Z"
  labels:
    kargo.akuity.io/project: "true"
    kubernetes.io/metadata.name: auto-rollback
  name: auto-rollback
  resourceVersion: "126627054"
  uid: ffe966f7-8610-4e20-9c34-70b0830765ac
spec:
  finalizers:
  - kubernetes
status:
  conditions:
  - lastTransitionTime: "2026-06-10T22:08:25Z"
    message: All resources successfully discovered
    reason: ResourcesDiscovered
    status: "False"
    type: NamespaceDeletionDiscoveryFailure
  - lastTransitionTime: "2026-06-10T22:08:25Z"
    message: All legacy kube types successfully parsed
    reason: ParsedGroupVersions
    status: "False"
    type: NamespaceDeletionGroupVersionParsingFailure
  - lastTransitionTime: "2026-06-10T22:08:25Z"
    message: 'Failed to delete all resource types, 1 remaining: admission webhook
      "replicated-resource-delete.kargo.akuity.io" denied the request: replicated
      resources are managed by Kargo and cannot be created, modified, or deleted by
      end users'
    reason: ContentDeletionFailed
    status: "True"
    type: NamespaceDeletionContentFailure
  - lastTransitionTime: "2026-06-10T22:08:25Z"
    message: All content successfully removed
    reason: ContentRemoved
    status: "False"
    type: NamespaceContentRemaining
  - lastTransitionTime: "2026-06-10T22:08:25Z"
    message: All content-preserving finalizers finished
    reason: ContentHasNoFinalizers
    status: "False"
    type: NamespaceFinalizersRemaining
  phase: Terminating
```

I can confirm that the management controller was not the one attempting to delete it, because I can replicate and unreplicate the secret and the secret will be created/deleted from the destination namespaces. I also confirmed that the webhooks server does have the MANAGEMENT_CONTROLLER_USERNAME set:

```
$ kshell kargo-webhooks-server-7876669786-6bqk7 -- sh
/ $ env | grep MANAGEMENT
MANAGEMENT_CONTROLLER_USERNAME=system:serviceaccount:kargo:kargo-management-controller
```

I concluded that the Kubernetes GC/namespace controller itself was attempting to delete the Secret, not the kargo-management-controller.

### Steps to Reproduce

1. Replicate a Secret
2. Delete a project
3. See the namespace remain

### Expected Behavior

We should be able to delete projects with replicated secrets

### Actual Behavior

Project is deleted, but not the underlying namespace

### Kargo Version

v1.10

### Relevant Logs

```shell

```

### Screenshots

_No response_

jessesuen:
## Description

Fixes https://github.com/akuity/kargo/issues/6443

## Summary

- When a Project is deleted, Kubernetes garbage collection controllers (`namespace-controller`, `generic-garbage-collector`, `kube-controller-manager`) attempt to delete the replicated secrets inside the terminating namespace. The replicated-resource-delete admission webhook was denying these requests, leaving namespaces permanently stuck in Terminating.
- The webhook's Handle function now also allows requests from well-known Kubernetes system controller identities and from members of the `system:masters` group (for break-glass scenarios).

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/pkg/webhook/kubernetes/admission.go
+++ b/pkg/webhook/kubernetes/admission.go
@@ -2,9 +2,20 @@ package webhook
 
 import (
 	"regexp"
+	"slices"
 
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
+
+// kubernetesSystemControllerUsernames is the set of well-known Kubernetes
+// system identities that perform garbage collection and namespace teardown.
+// These must be allowed to delete replicated secrets when a namespace is
+// removed so that namespaces do not become stuck in Terminating.
+var kubernetesSystemControllerUsernames = map[string]struct{}{
+	"system:serviceaccount:kube-system:namespace-controller":      {},
+	"system:serviceaccount:kube-system:generic-garbage-collector": {},
+	"system:kube-controller-manager":                              {},
+}
 
 type IsRequestFromKargoControlplaneFn func(admission.Request) bool
 
@@ -16,4 +27,19 @@ func IsRequestFromKargoControlplane(regex *regexp.Regexp) IsRequestFromKargoCont
 		}
 		return regex.Match([]byte(req.UserInfo.Username))
 	}
+}
+
+// IsRequestFromKubernetesSystemController returns true when the admission
+// request originates from one of the well-known Kubernetes system controllers
+// responsible for garbage collection and namespace teardown.
+func IsRequestFromKubernetesSystemController(req admission.Request) bool {
+	_, ok := kubernetesSystemControllerUsernames[req.UserInfo.Username]
+	return ok
+}
+
+// IsRequestFromClusterAdmin returns true when the admission request originates
+// from a member of the system:masters group, which maps to cluster-admin
+// privileges and is used for break-glass operations.
+func IsRequestFromClusterAdmin(req admission.Request) bool {
+	return slices.Contains(req.UserInfo.Groups, "system:masters")
 }
--- a/pkg/webhook/kubernetes/admission.go
+++ b/pkg/webhook/kubernetes/admission.go
@@ -7,14 +7,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-// kubernetesSystemControllerUsernames is the set of well-known Kubernetes
+// kubernetesGarbageCollectorUsernames is the set of well-known Kubernetes
 // system identities that perform garbage collection and namespace teardown.
 // These must be allowed to delete replicated secrets when a namespace is
 // removed so that namespaces do not become stuck in Terminating.
-var kubernetesSystemControllerUsernames = map[string]struct{}{
+var kubernetesGarbageCollectorUsernames = map[string]struct{}{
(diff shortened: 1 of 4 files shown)


jessesuen:
NOTE: I considered doing this in the `MutatingWebhookConfiguration` using `matchConditions` and filtering by user identity (CEL expressions like request.userInfo.username != "..."), but it requires Kubernetes 1.28+ (beta) / 1.30+ (stable).  Also, the management controller username exemption is already handled in code. Splitting the logic, with some exemptions in the Helm chart, some in Go, would make the full picture harder to reason about.

jessesuen:
I tested this end-to-end test and verified that after updating the kargo-webhooks-server Deployment with the fixed image, the Terminating namespaces were properly deleted.

pkg/webhook/kubernetes/admission.go
```diff
@@ -17,3 +28,18 @@ func IsRequestFromKargoControlplane(regex *regexp.Regexp) IsRequestFromKargoCont
...
 	}
 }
+
+// IsRequestFromKubernetesSystemController returns true when the admission
+// request originates from one of the well-known Kubernetes system controllers
+// responsible for garbage collection and namespace teardown.
+func IsRequestFromKubernetesSystemController(req admission.Request) bool {
+	_, ok := kubernetesSystemControllerUsernames[req.UserInfo.Username]
+	return ok
+}
```
fuskovic: The issue, tests, and comments seem to intend to scope to a `DELETE` operation but the logic here does not. So it looks like non-DELETE requests would get through as well. In addition to making the sure the operation is a DELETE, we should probably add tests for making sure we're not unintentionally allowing unintended operations as well.

fuskovic (changes requested):

pkg/webhook/kubernetes/admission.go
jessesuen: Good callout. I still feel cluster admins should be able to do any operation, not just deletes, for break-glass situations. So I would break it down like this:
* ~system~ GC controllers would only be allowed DELETE
* cluster admins can do anything

I'll make that change.

jessesuen: Ok. I renamed it to be more accurate, `IsRequestFromKubernetesGarbageCollector`, and only check for DELETE requests for those. Cluster admins can do anything.

fuskovic (approved):

PR #6450, opened 2026-06-14
feat(ui): add freight provenance card view

leonardollobato:
Summary
- Add an opt-in provenance view for Freight cards in the pipeline timeline.
- Keep the existing compact card view as the default.

Changes
- Add a `Card` filter with `Compact` and `Provenance` modes.
- Render provenance rows for alias, Git source/revision, chart version, image tag, generic artifact, Freight ID, and age.
- Persist the selected card view in URL params/local filter state.
- Use a wider carousel card width only when provenance mode is selected.

Testing
- `pnpm test -- src/features/project/pipelines/freight/freight-provenance-utils.test.ts --run`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build`
- `git diff --cached --check`
- `gitleaks protect --staged --redact --no-banner`

Risks/Notes
- The existing compact Freight card rendering remains unchanged unless users select `Card: Provenance`.
- `pnpm build` still reports existing bundle-size/dynamic-import warnings.

### DIFF
--- a/ui/src/features/project/pipelines/context/freight-timeline-controller-context.ts
+++ b/ui/src/features/project/pipelines/context/freight-timeline-controller-context.ts
@@ -22,6 +22,7 @@ export type FreightTimelineControllerContextType = {
     view: 'graph' | 'list';
     showMinimap: boolean;
     stepEdges: boolean;
+    freightCardView: 'compact' | 'provenance';
   };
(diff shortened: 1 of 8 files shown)


leonardollobato:
Closed because the requested solution is being kept GitOps-only; no upstream Kargo UI change is needed for this request.

PR #6451, opened 2026-06-15
fix(management-controller): skip Role normalization when create-actor annotation is absent

### LINKED ISSUE #6409: management-controller should disable normalizing kargo-* Roles in the absence of the create-actor Project annotation
### Pre-submission checklist

### Description

If you choose to self-manage the various kargo-* Roles, (for example to prevent granting kargo-admin to tenants), you do not add the `akuity.kargo.io/create-actor` annotation on Projects. Unfortunately, even without that annotation, the management-controller will normalize the Project's kargo-* Roles, which wipes additional scope and causes thrashing with the GitOps tooling which provisioned them.

### Steps to Reproduce

1. Create a Project without the Annotation.
2. Create a kargo-* Role with different scope than the management-controller ex pects.
3. Observe the drift.

### Expected Behavior

The management-controller should not modify resources that it did not create.

### Actual Behavior

The management-controller modifies the resources.

### Kargo Version

1.10.4

### Relevant Logs

```shell

```

### Screenshots

_No response_

AruneshDwivedi:
## What this fixes

The management-controller unconditionally normalizes kargo-* Roles (kargo-admin, kargo-viewer, kargo-promoter) on every Project reconciliation. When a Project does not have the `kargo.akuity.io/create-actor` annotation, these Roles are self-managed (e.g. via GitOps), and the controller overwriting them causes:
- Loss of custom scope on the Roles
- Thrashing with GitOps tooling that provisioned the Roles

## What I changed

In `pkg/controller/management/projects/projects.go`, the Role normalization loop in `ensureDefaultUserRoles` now checks for the `create-actor` annotation before running. When the annotation is absent, Role creation/update is skipped entirely.

## How to test

1. Create a Project without the `kargo.akuity.io/create-actor` annotation
2. Manually create a kargo-admin Role with custom rules
3. Trigger a reconciliation (e.g. update the Project)
4. Verify the custom Role rules are preserved (not overwritten)

Fixes #6409

### DIFF
--- a/pkg/controller/management/projects/projects.go
+++ b/pkg/controller/management/projects/projects.go
@@ -1000,6 +1000,10 @@ func (r *reconciler) ensureDefaultUserRoles(
 			}
 		}
 	}
+	// Only normalize Roles and RoleBindings when the create-actor annotation
+	// is present. Without it, the Project's kargo-* Roles are self-managed
+	// (e.g. via GitOps) and the controller should not overwrite them.
+	if _, ok := project.Annotations[kargoapi.AnnotationKeyCreateActor]; ok {
 	for _, role := range roles {
 		roleLogger := logger.WithValues(
 			"name", role.Name,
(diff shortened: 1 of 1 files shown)


krancour:
@AruneshDwivedi did you read through the thread in #6409 ? This annotation was never intended to have any bearing on that behavior. _It is 100% unrelated_.

PR #6461, opened 2026-06-17
test(charts): add helm-unittest suite for the kargo chart

EronWright:
## Context

The `charts/kargo` chart has **no unit tests** today — only `make lint-charts` (`helm lint`), which
catches YAML/schema errors but not rendering logic. The chart carries a lot of conditional behavior
that can regress silently: per-component `enabled` flags, the `dataPlane.install` / `workloads.install`
master gates, RBAC scope toggles (`rbac.installClusterRoles/Bindings`), TLS-aware URLs / probes /
service ports, CRD keep-policy annotations, controller sharding, webhook CA-bundle injection, and
OIDC/Dex wiring.

This adds a [`helm-unittest`](https://github.com/helm-unittest/helm-unittest) suite covering that
logic, mirroring the style used in Akuity's internal `akuity-platform` chart tests.

## What's here

**Test suites** (`charts/kargo/tests/`) — **99 tests / 34 suites**, one suite per template file:

| File | Coverage |
|---|---|
| `_helpers_test.yaml` | `kargo.image` tag defaulting/override, `kargo.labels` + `global.labels` merge, `kargo.baseURL`/`useTLS` (http vs https across direct/ingress/upstream TLS) |
| `crds_test.yaml` | 9 CRDs, `helm.sh/resource-policy: keep` toggle, `crds.install` / `dataPlane.install` gating |
| `namespaces_test.yaml` | system / shared / cluster-secrets namespace creation + `createNamespace` toggles |
| `common_test.yaml` | `kargo-project-admin` / `-secrets-reader` cluster roles; secret-rule gating on `api.secretManagementEnabled` |
| `users_test.yaml` | the four system SAs/cluster roles/bindings; OIDC claim annotations |
| `argocd_test.yaml` | namespaced Argo CD role/binding gated on `watchArgocdNamespaceOnly` |
| `api_test.yaml` | Deployment/ConfigMap/Secret/Service/Ingress; TLS-aware issuer/probes/ports; OIDC; `/webhooks` ingress path |
| `controller_test.yaml` | Deployment, sharding (`IS_DEFAULT_CONTROLLER`/`SHARD_NAME`), API base URL, integration ClusterRoleBindings |
| `management-controller_test.yaml` | Deployment + ConfigMap (namespaces, reconciler concurrency, webhook base URL) |
| `garbage-collector_test.yaml` | CronJob (schedule, `Forbid`, args) + retention ConfigMap |
| `kubernetes-webhooks-server_test.yaml` | Mutating/Validating webhook configs, CA-bundle injection, control-plane user regex |
| `external-webhooks-server_test.yaml` | Deployment + ConfigMap (TLS env, base URL) + Service |
| `dex-server_test.yaml` | Deployment/Secret/Service gated on `api.oidc.dex.enabled` |

Plus `tests/README.md` (conventions) and `test/values-unittest.yaml` (baseline = only the API
admin-account `passwordHash`/`tokenSigningKey` the chart `fail`s without).

**Runner & packaging**
- `make test-chart` / `hack-test-chart` — installs helm-unittest v1.1.0, then `helm unittest`.
- `charts/kargo/.helmignore` — keeps `tests/` and `test/` out of the packaged chart.

**CI**
- New `test-chart` job in `.github/workflows/ci.yaml`, mirroring the existing `lint-charts` job.

## Notes for reviewers

- Workload Deployments (api, controller, management-controller, both webhook servers) and the GC CronJob
  `include` their sibling ConfigMap/Secret for a checksum annotation, so those suites list both
  templates and scope each test with leaf-level `template:` (documented in `tests/README.md`).
- The baseline reflects chart defaults — note TLS is **on** by default, so default assertions expect
  https/443/exec-probes and flip via `api.tls.enabled: false`.

## Verification

- `make test-chart` → 99 passed / 34 suites.
- `make lint-charts` passes.
- Confirmed non-vacuous: a deliberately-wrong assertion fails as expected.

### DIFF
--- a/charts/kargo/tests/api_test.yaml
+++ b/charts/kargo/tests/api_test.yaml
@@ -1,0 +1,228 @@
+suite: api/deployment.yaml
+values:
+  - ../test/values-unittest.yaml
+# configmap + secret are listed so the Deployment's checksum includes resolve;
+# each test scopes assertions to the Deployment with template:.
+templates:
+  - templates/api/deployment.yaml
+  - templates/api/configmap.yaml
+  - templates/api/secret.yaml
+tests:
+  - it: renders the API Deployment by default
+    template: templates/api/deployment.yaml
+    asserts:
+      - hasDocuments:
+          count: 1
+      - isKind:
+          of: Deployment
+      - equal:
+          path: metadata.name
+          value: kargo-api
+      - equal:
+          path: spec.template.spec.serviceAccountName
+          value: kargo-api
+      - matchRegex:
+          path: spec.template.spec.containers[0].image
+          pattern: "^.+/kargo:.+"
+      - equal:
+          path: spec.template.spec.containers[0].ports[0].containerPort
+          value: 8080
+      - equal:
+          path: spec.template.spec.containers[0].envFrom[0].configMapRef.name
+          value: kargo-api
+      - equal:
+          path: spec.template.spec.containers[0].envFrom[1].secretRef.name
+          value: kargo-api
+
+  - it: is not rendered when the API is disabled
+    template: templates/api/deployment.yaml
+    set:
+      api.enabled: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: is not rendered when workloads are not installed
+    template: templates/api/deployment.yaml
+    set:
+      workloads.install: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: uses an exec health probe when TLS is enabled (default)
+    template: templates/api/deployment.yaml
+    asserts:
+      - exists:
+          path: spec.template.spec.containers[0].livenessProbe.exec
+      - notExists:
+          path: spec.template.spec.containers[0].livenessProbe.grpc
+
+  - it: uses a gRPC health probe when TLS is disabled
+    template: templates/api/deployment.yaml
+    set:
+      api.tls.enabled: false
+    asserts:
+      - exists:
+          path: spec.template.spec.containers[0].livenessProbe.grpc
+      - notExists:
+          path: spec.template.spec.containers[0].livenessProbe.exec
+
+---
+suite: api/configmap.yaml
+values:
+  - ../test/values-unittest.yaml
+templates:
+  - templates/api/configmap.yaml
+tests:
+  - it: renders API configuration with defaults
+    asserts:
+      - isKind:
+          of: ConfigMap
+      - equal:
+          path: data.LOG_LEVEL
+          value: INFO
+      - equal:
+          path: data.SECRET_MANAGEMENT_ENABLED
+          value: "true"
+      - equal:
+          path: data.ROLLOUTS_INTEGRATION_ENABLED
+          value: "true"
+      - equal:
+          path: data.ADMIN_ACCOUNT_ENABLED
+          value: "true"
+
+  - it: derives an https token issuer and TLS env when TLS is enabled (default)
+    asserts:
+      - equal:
+          path: data.ADMIN_ACCOUNT_TOKEN_ISSUER
+          value: https://localhost
+      - equal:
+          path: data.TLS_ENABLED
+          value: "true"
+
+  - it: derives an http token issuer when TLS is disabled
+    set:
+      api.tls.enabled: false
+    asserts:
+      - equal:
+          path: data.ADMIN_ACCOUNT_TOKEN_ISSUER
+          value: http://localhost
+      - notExists:
+          path: data.TLS_ENABLED
+
+  - it: omits secret management when disabled
+    set:
+      api.secretManagementEnabled: false
+    asserts:
+      - notExists:
+          path: data.SECRET_MANAGEMENT_ENABLED
+
+  - it: enables OIDC config when OIDC is turned on
+    set:
+      api.oidc.enabled: true
+      api.oidc.issuerURL: https://idp.example.com
+      api.oidc.clientID: kargo
+    asserts:
+      - equal:
+          path: data.OIDC_ENABLED
+          value: "true"
+      - equal:
+          path: data.OIDC_ISSUER_URL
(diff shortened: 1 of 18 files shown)


charts/kargo/tests/api_test.yaml
```diff
@@ -0,0 +1,228 @@
...
+# each test scopes assertions to the Deployment with template:.
+templates:
+  - templates/api/deployment.yaml
+  - templates/api/configmap.yaml
+  - templates/api/secret.yaml
+tests:
+  - it: renders the API Deployment by default
+    template: templates/api/deployment.yaml
+    asserts:
+      - hasDocuments:
```
krancour, line 14: Repeats many places in the PR...

Usage of suite-level `templates` and job-level `template` seems inconsistent.

I believe that if a suite is named something like `api/deployment.yaml` and, indeed, scoped to testing _just_ the deployment template, you could list _just_ the deployment template under suite-level `templates` (the other two templates aren't actually involved in any of the test jobs in this suite) and then _skip_ re-specifying `template: templates/api/deployment.yaml` in each job.

Something like this:


```suggestion
suite: api/deployment.yaml
values:
  - ../test/values-unittest.yaml
# configmap + secret are listed so the Deployment's checksum includes resolve;
# each test scopes assertions to the Deployment with template:.
templates:
  - templates/api/deployment.yaml
tests:
  - it: renders the API Deployment by default
    asserts:
      - hasDocuments:
```

charts/kargo/tests/api_test.yaml
```diff
@@ -0,0 +1,228 @@
...
+
+  - it: derives an http token issuer when TLS is disabled
+    set:
+      api.tls.enabled: false
+    asserts:
+      - equal:
+          path: data.ADMIN_ACCOUNT_TOKEN_ISSUER
+          value: http://localhost
+      - notExists:
+          path: data.TLS_ENABLED
```
krancour: I don't think TLS enabled/disabled and admin account token issuer have anything to do with one another.

krancour: Oh... nevermind. I understand now... you're testing http vs https in the issuer URL.

I think the wording/naming of the test cases tripped me up a bit.

Possibly worth rephrasing?

Not a hill I would die on.

charts/kargo/tests/api_test.yaml
```diff
@@ -0,0 +1,228 @@
...
+    set:
+      api.tls.enabled: false
+    asserts:
+      - exists:
+          path: spec.template.spec.containers[0].livenessProbe.grpc
+      - notExists:
+          path: spec.template.spec.containers[0].livenessProbe.exec
+
+---
+suite: api/configmap.yaml
```
krancour, line 72: General comment/question... applies beyond this suite...

There are _so many_ conditionals in these templates and this test suite covers just a very small number of them. I'm just trying to wrap my head around whether it's deliberate that certain conditions are tested and others, not tested, and if so, what is the framework is for determining test-worthiness?

Of course... this brings us back to an offline discussion we'd had once about if you _really_ test all these conditionals thoroughly, in the end, what you've really achieved isn't much more than writing everything twice in different syntax. 🤷‍♂️

krancour: I'll hold off on reviewing exhaustively until an answer to the above gives me more clarity on the approach.

charts/kargo/tests/api_test.yaml
EronWright: The README treats your concern [here](https://github.com/akuity/kargo/pull/6461/changes#diff-a621ad82d0316e7dae8b26a8a26cfbad7c462c667033db1b30acbef733e3ad89R40-R48) - your suggestion is indeed the preferred approach and is used everywhere it can be.  The special case you see here is out of necessity, I recall related to a checksum calculation, and is compensated for by specifying the `template:` at the test level.

charts/kargo/tests/api_test.yaml
EronWright: _Here's the rubric under which it was developed:_

1. Gating / conditional rendering — a resource appears or disappears under the right flags (the `hasDocuments: 0` cases, master gates like `dataPlane.install/workloads.install`, RBAC scope toggles). This is the single most common silent breakage: someone edits an `if` and a resource vanishes or leaks.
2. Derived values — where output is a function of input, not a passthrough: `kargo.baseURL` http-vs-https from the TLS flags, `selectCpuResourceField` (limits→requests→fallback), `controlplane.defaultUserRegex` assembled from enabled components, controller sharding (`IS_DEFAULT_CONTROLLER`/`SHARD_NAME`), webhook operation sets, CA-bundle injection conditions.
3. Cross-resource wiring / invariants — SA names that bindings depend on, the checksum-include actually resolving, the `/webhooks` ingress path routing to the right backend.
4. Sharp edges with real cost — TLS port/probe selection, `helm.sh/resource-policy: keep`, the `fail`-required admin-account values.

What we deliberately don't do: assert every scalar passthrough (`logLevel` → `LOG_LEVEL`) beyond one representative case per template, test static boilerplate, or enumerate the cross-product of independent flags — each branch once, not the combinatorics, unless flags actually interact.

charts/kargo/tests/api_test.yaml
krancour: > related to a checksum calculation

Oh... that makes total sense. Thank you for the explanation!

charts/kargo/tests/api_test.yaml
krancour: Thanks for the explanation. I think all of that is quite sensible. I do wonder if "Sharp edges with real cost" might qualify many additional things to be tested that currently aren't, but I think this is a case where there's no reason to let perfect be the enemy of good. We had no tests for the chart before and this at least covers a fair number of the so-called sharp edges, so it's a win.

krancour (approved):

PR #6470, opened 2026-06-18
fix(ui): compare promotion freight against matching warehouse

### LINKED ISSUE #6371: New version comparison preview in promotion drawer inaccurate
### Pre-submission checklist

### Description

The new preview of what is changing during a promotion is not accurate

Image

### Steps to Reproduce

1. Promote a Freight in a multi-warehouse stage with UI

### Expected Behavior

1. It should show the current version we are moving from
2. For unrelated freight that is not promoting, it should not say "removed"

### Actual Behavior

1. It doesn't show what I'm moving from
2. In a multi-warehouse Stage, it's showing "Removed" if I'm promoting a Freight that is from a Warehouse that I'm not promoting.

### Kargo Version

v1.11

### Relevant Logs

```shell

```

### Screenshots

https://github.com/user-attachments/assets/b458d79c-1836-4204-9cde-e340fbe022ac

rpelczar:
fixes #6371 

https://github.com/user-attachments/assets/7b906186-dc2d-4589-b952-36f5e7668d52

### DIFF
--- a/ui/src/features/common/utils.test.ts
+++ b/ui/src/features/common/utils.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, test } from 'vitest';
+
+import { Freight, FreightReference, Stage } from '@ui/gen/api/v1alpha1/generated_pb';
+
+import { getCurrentFreightForComparison } from './utils';
+
+const ref = (origin: string, image: string): FreightReference =>
+  ({
+    origin: { kind: 'Warehouse', name: origin },
+    images: [{ repoURL: image }]
+  }) as unknown as FreightReference;
+
+// stage builds a Stage whose current freight collection holds one piece of
+// Freight per Warehouse, keyed by origin name (mirroring the API shape).
+const stage = (refs: FreightReference[]): Stage =>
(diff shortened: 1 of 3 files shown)


jessesuen:
So, is it the intention of this fix is that for a multi-origin Stage, we will only show the single warehouse that is currently being promoted and it's difference? As opposed to: list of all of the Stages origins and "unchanged" for the origins that are not changing?

I think that is fine, since only one origin can anyways change at a time. I just ask because the current version appears to want to show a table of all origins. Maybe it is still useful to expand the table to show the unchanged origins in case they want to review all the versions of the warehouse before promotion?

rpelczar:
@jessesuen That's my understanding too: a Promotion references a single Freight, and each Freight comes from exactly one Warehouse - so only one origin can change per promotion. The fix scopes the diff to that origin (incoming vs. the Stage's current Freight from the same Warehouse). We can change it if you want, but I thought it wasn't right.

jessesuen (approved):

Marvin9 (approved):

PR #6473, opened 2026-06-19
feat(ui): surface current freight, show aliases, drop UIDs

### LINKED ISSUE #6308: freight view does not show origin
Something I noticed -- the slide out when a Freight resource is selected from the timeline does not indicate the Freight's origin (Warehouse).

A lot of the time, this may not matter, but because an unfiltered timeline could show Freight from multiple origins, it would be nice to see this information once a Freight resource has been selected.

### LINKED ISSUE #6465: UI nits/improvements
### Pre-submission checklist

### Proposed Feature

Some UI nits that I'd like to improve:

## 2. Also show the alias in addition to the hash. The alias is often more useful to users

Image

Image

## 3. UID of the object is never useful to show. We can remove it across the board:

Freight drawer:
Image

Promote drawer:
Image

### Motivation and Use Case

Small usability improvements

### Alternatives Considered

_No response_

rpelczar:
fixes #6465, fixes #6308

- Stage drawer: show currently-running freight (alias + artifacts) in each
  requested-freight card
- Promotions and Verifications tabs: show alias alongside the hash
- Remove UID display from the freight and promotion drawers
- Extract shared FreightArtifactList ("show N + X more") used by the
  freightline card and the current-freight panel
  
  
image

### DIFF
--- a/ui/src/features/freight/freight-details.tsx
+++ b/ui/src/features/freight/freight-details.tsx
@@ -1,10 +1,9 @@
 import { toJson } from '@bufbuild/protobuf';
 import { faFile, faInfoCircle, faPencil } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { Button, Drawer, Space, Tabs, Typography } from 'antd';
-import classNames from 'classnames';
+import { Button, Descriptions, Drawer, Space, Tabs, Typography } from 'antd';
 import { useEffect, useState } from 'react';
-import { generatePath, useNavigate, useParams } from 'react-router-dom';
+import { Link, generatePath, useNavigate, useParams } from 'react-router-dom';
 
 import { paths } from '@ui/config/paths';
 import { useExtensionsContext } from '@ui/extensions/extensions-context';
@@ -22,12 +21,6 @@ import { FreightMetadata } from './freight-metadata';
 import { FreightStatusList } from './freight-status-list';
 import { UpdateFreightAliasModal } from './update-freight-alias-modal';
 
-const CopyValue = (props: { value: string; label: string; className?: string }) => (
-  <div className={classNames('flex items-center text-gray-500 font-mono', props.className)}>
-    <span className='text-gray-400 mr-2 text-xs'>{props.label}</span>
-    <Typography.Text copyable>{props.value}</Typography.Text>
-  </div>
-);
 export const FreightDetails = ({
   freight,
   refetchFreight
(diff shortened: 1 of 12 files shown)


ui/src/features/freight/freight-details.tsx
```diff
@@ -111,9 +111,6 @@ export const FreightDetails = ({
                         {alias && freight?.metadata?.name && (
                           <CopyValue label='NAME:' value={freight.metadata?.name} />
                         )}
```
Marvin9: can we do https://github.com/akuity/kargo/issues/6308 in this PR?

ui/src/features/project/pipelines/freight/freight-card.tsx
```diff
@@ -287,34 +287,7 @@ export const FreightCard = (props: FreightCardProps) => {
...
-                {noOfGitCommits +
-                  noOfHelmReleases +
-                  noOfContainerImages -
-                  (props.freight?.charts?.slice(0, 2)?.length +
-                    props.freight?.commits?.slice(0, 2)?.length +
-                    props.freight?.images?.slice(0, 2)?.length)}{' '}
-                more
-              </Typography.Text>
-            )}
+            <FreightArtifactList freight={props.freight} />
```
Marvin9, line 290: FYI this will change previous behaviour - for example if there are 3 commits and 1 image and if max if 3, then it will show 3 commits, instead of 2 commits and 1 image

ui/src/features/stage/promotions.tsx
```diff
@@ -44,15 +45,7 @@ export const Promotions = ({ argocdShard }: { argocdShard?: ArgoCDShard }) => {
...
-  const [curFreight, setCurFreight] = useState<string | undefined>();
-
-  const { data: freightData, isLoading: isLoadingFreight } = useQuery(
-    getFreight,
-    { project: projectName, name: curFreight },
-    {
-      enabled: !!curFreight
-    }
-  );
+  const freightMap = useGetFreightMap(projectName || '');
```
Marvin9, line 48: just heads up that we are now querying all freights. though its cached so I think it won't be any issue but worth checking anyways

Marvin9 (commented):
Can you also check if layout doesn't break or looks weird with more wareohouses?

Marvin9 (approved):
Comments above but not blockers. LGTM

ui/src/features/freight/freight-details.tsx
rpelczar: done

ui/src/features/stage/promotions.tsx
rpelczar: done

ui/src/features/stage/tabs/freight-history/use-get-freight-map.ts
```diff
@@ -1,25 +1,40 @@
...
+
+  // Stage details is rendered by the pipeline view, which already loads the
+  // project's Freight into DictionaryContext. Reuse that map instead of issuing
+  // a duplicate queryFreight request. This is only safe when no warehouse filter
+  // is active: a filtered context map omits Freight that Stage details still
+  // needs to resolve (e.g. aliases in the promotions/verifications tables), so
+  // in that case we fall back to our own unfiltered fetch.
+  const canReuseContext = !!dictionaryContext && preferredFilter.warehouses.length === 0;
+
+  const freightQuery = useQuery(queryFreight, { project }, { enabled: !canReuseContext });
```
Marvin9, line 21: Sorry I forgot to mention this earlier, just realised now. But we have query freight by stage. How about just using that? I should have used that when I wrote this hook initially, can't recall why I didn't already.

Marvin9: Lets do it separately as discussed

PR #6486, opened 2026-06-23
feat: expose metrics Services and optional ServiceMonitors in the Helm chart

### LINKED ISSUE #6405: feat: expose metrics Services and optional ServiceMonitors in the Helm chart
## Pre-submission checklist

## Proposed Feature

The Kargo controller, management-controller and webhooks-server can already expose Prometheus metrics by setting `METRICS_BIND_ADDRESS`, but the Helm chart provides no way to make those metrics scrapeable. None of these components declare a metrics container port, and the chart ships no metrics `Service` and no `ServiceMonitor`. Operators currently have to add all of this out of band.

I would like the chart to adopt the same metrics pattern already used by the sibling Argo projects (`argo-cd` and `argo-workflows` in `argoproj/argo-helm`), namely:

1. A per-component metrics `Service` gated by a values flag, for example:

```yaml
controller:
  metrics:
    enabled: false
    service:
      type: ClusterIP
      clusterIP: ""        # "None" for a headless service
      annotations: {}
      labels: {}
      servicePort: 9090
      portName: metrics
    serviceMonitor:
      enabled: false
      interval: 30s
      relabelings: []
      metricRelabelings: []
      additionalLabels: {}
      scheme: ""
      tlsConfig: {}
      namespace: ""
```

2. The matching metrics container port declared on the component's pod spec when `metrics.enabled` is true.

3. An optional `ServiceMonitor` gated by both `metrics.enabled` and `metrics.serviceMonitor.enabled`, guarded by a `Capabilities.APIVersions.Has` check so it is skipped on clusters without the Prometheus Operator CRDs.

The same structure would apply to `managementController` and `webhooksServer`. All flags default to `false`, so the change is fully backward compatible.

## Motivation and Use Case

We run Kargo on EKS with Grafana Alloy for metrics collection. Because the chart declares no metrics container port and no metrics Service, our annotation based pod discovery cannot scrape the controllers, and we had to add three headless Services and a separate Alloy scrape config ourselves to get any Kargo metrics into Grafana.

This is the standard pattern every other controller-runtime based tool in the ecosystem ships, including Argo CD and Argo Workflows, which live in the same Helm repository and are maintained alongside Kargo. Adopting it would let operators monitor Kargo with a couple of values instead of maintaining bespoke manifests, and would work out of the box for both Prometheus Operator users (via ServiceMonitor) and annotation or Service based scrapers.

## Alternatives Considered

- Adding the metrics Services and ServiceMonitors as raw manifests in a chart wrapper (what we do today). It works but every operator has to reinvent it, and it drifts from the upstream pod labels.
- Kustomize "last mile" patches over `helm template`. This still requires each operator to hand write the Service, the container port and the ServiceMonitor, and does not benefit the wider community.

We are happy to contribute the implementation once the chart restructuring referenced in #6250 settles and maintainers are open to chart PRs again.

EronWright:
## Summary

The controller, management-controller, and webhooks-server already serve Prometheus metrics via controller-runtime when `METRICS_BIND_ADDRESS` is set, but the Helm chart provided no way to make those metrics scrapeable — no metrics container port, no metrics `Service`, and no `ServiceMonitor`. Operators had to add all of this out of band.

This adopts the same metrics pattern used by the sibling `argo-cd` and `argo-workflows` charts, for the controller, `managementController`, and `webhooksServer` components.

Closes #6405

## What's added

Per component, all gated `false` by default (fully backward compatible):

- A `metrics.{service,serviceMonitor}` values block.
- A metrics `Service` gated by `metrics.enabled`.
- The matching metrics container port, plus wiring of `METRICS_BIND_ADDRESS` in the component's ConfigMap so the metrics server actually binds.
- An optional `ServiceMonitor`, gated by `metrics.serviceMonitor.enabled` **and** a `Capabilities.APIVersions.Has "monitoring.coreos.com/v1"` check, so it is silently skipped on clusters without the Prometheus Operator CRDs.

## Objects

The same shape is rendered for each of `controller`, `managementController`, and `webhooksServer`:

```mermaid
flowchart TD
  subgraph component["Per component: controller / managementController / webhooksServer"]
    cm["ConfigMap<br/>+ METRICS_BIND_ADDRESS = :9090"]:::changed
    pods["Deployment Pods<br/>+ container port 'http-metrics' (9090)"]:::changed
  end

  svc["Service: kargo-COMPONENT-metrics<br/>gate: metrics.enabled"]:::new
  sm["ServiceMonitor: kargo-COMPONENT-metrics<br/>gate: serviceMonitor.enabled<br/>+ monitoring.coreos.com/v1 CRDs"]:::new
  prom(["Prometheus / scraper"]):::ext

  cm -->|envFrom| pods
  svc -->|"selector -> pod endpoints"| pods
  sm -->|"selector -> Service"| svc
  prom -->|"discovers via ServiceMonitor"| sm
  prom -.->|"scrapes HTTP /metrics on each pod"| pods

  classDef new fill:#dcfce7,stroke:#16a34a,color:#000;
  classDef changed fill:#fef9c3,stroke:#ca8a04,color:#000;
  classDef ext fill:#dbeafe,stroke:#2563eb,color:#000;
```

🟩 new object &nbsp;·&nbsp; 🟨 modified existing object &nbsp;·&nbsp; 🟦 external

## Notes

- **No Go changes.** Metrics are served over plain HTTP (controller-runtime's metrics server is configured with only `BindAddress`, no secure serving), so the `ServiceMonitor` default `scheme` works out of the box.
- **Default port name is `http-metrics`** (port `9090`), so service meshes that infer the protocol from the port name treat it as HTTP. Both are configurable per component.
- **`garbage-collector` is intentionally excluded** — it's a `CronJob` with ephemeral pods, which the Service/ServiceMonitor pull model can't reliably scrape (that's a Pushgateway use case).
- The `ServiceMonitor` selects the Service's endpoints (not its cluster IP), so per-replica metrics are preserved when a component runs more than one replica.

## Testing

- `helm unittest` — 126 passed (40 suites), covering: default-off rendering, Service render/port/headless, the `METRICS_BIND_ADDRESS` wiring, the container port, and the ServiceMonitor's dual flag + CRD-capability gating (via helm-unittest `capabilities.apiVersions`).
- `helm lint` passes.
- Chart `README.md` regenerated from the new `@param` annotations.
- Added a **Monitoring** page to the operator guide.


docs/docs/40-operator-guide/45-monitoring.md
```diff
@@ -0,0 +1,146 @@
...
+
+:::
+
+:::note
+
+The default port name (`http-metrics`) carries an `http-` prefix so that
+service meshes that infer a port's protocol from its name treat it as HTTP. If
+you rename the port for a meshed cluster, keep the `http-` prefix.
+
+:::
```
thomastaylor312, line 70: Nit: Two admonitions in a row like this looks a little odd to me. Kinda feels like the first one should be part of the paragraph starting on line 51

thomastaylor312 (approved):

docs/docs/40-operator-guide/45-monitoring.md
EronWright: Good call — folded the plain-HTTP/TLS note into the paragraph above the example, so there's now a single admonition (the service-mesh one). Pushed in 30cf8453.

EronWright:
### Live validation

Verified end-to-end on an isolated kind cluster (k8s 1.35) with `metrics.enabled: true` for all three components.

**Metrics Services** — all created:

```
kargo-controller-metrics              ClusterIP   9090/TCP
kargo-management-controller-metrics   ClusterIP   9090/TCP
kargo-webhooks-server-metrics         ClusterIP   9090/TCP
```

**Metrics actually served** — port-forwarded the `kargo-management-controller-metrics` Service and scraped it:

```
$ curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:19090/metrics
200
```

677 series returned, including real `controller_runtime_*`, `workqueue_*`, and `go_*` metrics. Confirms the full chain: ConfigMap `METRICS_BIND_ADDRESS=:9090` → server binds → `http-metrics` container port → Service routes → plain HTTP (no TLS).

**ServiceMonitor** — correctly skipped by Tilt's offline `helm template` (no `monitoring.coreos.com/v1` capability), as designed. Validated separately against the real upstream prometheus-operator CRD: rendered with `--api-versions monitoring.coreos.com/v1` and applied — all three accepted, e.g. `kargo-controller-metrics` with `port: http-metrics`, `interval: 30s`, and the correct component selector.

(Independent note, not part of this PR: `tilt up` on current `main` fails to install the `projectconfigs` CRD on a k8s 1.35 apiserver — the `autoRollback` CEL `x-kubernetes-validations` exceed the cost budget. Surfaced during testing; will follow up separately.)

PR #6489, opened 2026-06-23
fix: validate AutoRollbackConfig elements with CEL instead of array-level enum

jessesuen:
## Description

The previous [fix](https://github.com/akuity/kargo/pull/6485), wasn't proper and would fail valid ProjectConfigs with:

```
$ k apply -f ./kargo/projectconfig.yaml                                                                                                                                                                                                                                                                                        The ProjectConfig "auto-rollback" is invalid:
* spec.promotionPolicies[0].autoRollback.onPromotion: Unsupported value: []interface {}{"Failed"}: supported values: "Failed", "Errored"
```

The problem was that +kubebuilder:validation:Enum markers on the OnPromotion and OnVerification slice fields placed the enum constraint at the array schema level, causing Kubernetes to reject any array value because the serialized array (e.g. ['Failed']) never equals one of the enum strings. This broke AutoRollbackConfig entirely.

Replace with two XValidation CEL rules per field. Each rule checks one element (self[0] or self[1]), keeping the string-comparison count at two per rule so the Kubernetes static CEL cost estimator stays below its 10M-unit per-rule budget.

## Checklist

### Eligibility

### Quality

### AI Use Disclosure
This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/api/v1alpha1/project_config_types.go
+++ b/api/v1alpha1/project_config_types.go
@@ -120,7 +120,8 @@ type AutoRollbackConfig struct {
 	// +optional
 	// +listType=set
 	// +kubebuilder:validation:MaxItems=2
-	// +kubebuilder:validation:Enum=Failed;Errored
+	// +kubebuilder:validation:XValidation:message="onPromotion[0] must be Failed or Errored",rule="self.size() == 0 || self[0] == 'Failed' || self[0] == 'Errored'"
+	// +kubebuilder:validation:XValidation:message="onPromotion[1] must be Failed or Errored",rule="self.size() <= 1 || self[1] == 'Failed' || self[1] == 'Errored'"
 	OnPromotion []PromotionPhase `json:"onPromotion,omitempty" protobuf:"bytes,1,rep,name=onPromotion"`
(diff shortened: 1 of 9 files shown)


thomastaylor312 (approved):

api/v1alpha1/project_config_types.go
```diff
@@ -120,7 +120,8 @@ type AutoRollbackConfig struct {
 	// +optional
 	// +listType=set
 	// +kubebuilder:validation:MaxItems=2
-	// +kubebuilder:validation:Enum=Failed;Errored
+	// +kubebuilder:validation:XValidation:message="onPromotion[0] must be Failed or Errored",rule="self.size() == 0 || self[0] == 'Failed' || self[0] == 'Errored'"
+	// +kubebuilder:validation:XValidation:message="onPromotion[1] must be Failed or Errored",rule="self.size() <= 1 || self[1] == 'Failed' || self[1] == 'Errored'"
```
jessesuen, line 124: It's a little ugly, but it's the only way for these checks to come under the CEL budget, despite playing around with maxLength and maxItems on both the phases and the array.

PR #6490, opened 2026-06-24
fix(git-merge-pr): resilience to concurrent-merge races via mergeable_state + 405 retry

### LINKED ISSUE #5761: git-merge-pr: 'Base branch was modified' 405 should be retryable when wait=true
# Checklist

# Bug Description

When multiple promotions target the same Git repository concurrently, the `git-merge-pr` step fails with an unrecoverable error when the base branch is modified between PR creation and the merge attempt:

```
an unrecoverable error occurred: error running step "task-1::merge-pr":
  error merging pull request 96223: error merging pull request 96223:
  PUT https://api.github.com/repos/<org>/<repo>/pulls/96223/merge:
  405 Base branch was modified. Review and try the merge again. []
```

This GitHub 405 response is transient — the merge would succeed on a simple retry after the base branch settles. However, Kargo treats it as an unrecoverable error, causing the entire promotion to fail even when `wait: true` is set.

# Expected Behavior

When `wait: true` is configured on `git-merge-pr`, the `405 Base branch was modified` response from GitHub should be treated as a retryable condition (similar to how pending checks or non-mergeable states are handled). The step should return a running status and retry on the next reconciliation.

# Steps to Reproduce

1. Configure two or more Kargo stages that promote to the same Git repository (e.g., a development-canary and staging-canary stage targeting the same infrastructure repo)
2. Set `autoPromotionEnabled: true` for both stages
3. Configure `git-merge-pr` with `wait: true` in the promotion task
4. When new freight arrives, both stages promote concurrently
5. The first PR merges successfully, but the second PR fails with `405 Base branch was modified` because `main` moved forward

# Promotion Task Configuration

```yaml
- as: merge-pr
  config:
    prNumber: ${{ task.outputs['open-pr'].pr.id }}
    repoURL: ${{ vars.gitRepo }}
    wait: true
  if: ${{ status('open-pr') == 'Succeeded' && task.outputs['open-pr'].pr.id != nil }}
  uses: git-merge-pr
```

# Environment

- Kargo version: 1.8.4
- Git provider: GitHub Enterprise

EronWright:
## What

Makes the `git-merge-pr` step's GitHub provider resilient to the concurrent-merge race in #5761, in two complementary places:

1. The **pre-merge gate** reads GitHub's richer `mergeable_state` field instead of relying solely on the `Mergeable` boolean.
2. The **merge call** treats a transient `405 "Base branch was modified"` as *not ready* (rather than a hard failure), so the step retries.

## Why

`MergePullRequest` previously gated the merge attempt on `ptr.Deref(ghPR.Mergeable, false)`. That boolean is tri-state (`true`/`false`/`nil`, computed asynchronously by GitHub) and collapses several genuinely different conditions into one "not ready" outcome:

- a **real merge conflict** (`dirty`) — will never resolve without a human
- the **base branch moved forward** (`behind`) — the time-of-check/time-of-use race described in #5761
- **blocked** by required reviews/checks (`blocked`) — clears when checks pass
- mergeability **still being computed** (`unknown`) — clears on its own

Today all of these look identical: a vague "not ready" that either retries forever (when `wait: true`) or fails with an unhelpful message (when `wait: false`).

A correct gate is necessary but not sufficient: it is a snapshot, so the base branch can still move between the mergeability `GET` and the `PUT …/merge`. GitHub then rejects the merge with `405 "Base branch was modified"`, which previously terminated the promotion. That residual race is the second half of this fix.

## How

**Gate** — switch on `mergeable_state`:

| state | behavior |
|---|---|
| `clean`, `unstable`, `has_hooks` | proceed to merge |
| `dirty` | **terminal error** (`pull request N has conflicts and cannot be merged`) |
| `behind`, `blocked`, `draft`, `unknown` | not ready → caller retries |
| unset / unrecognized | fall back to the prior `Mergeable`/`Draft` boolean logic |

**Merge-time 405** — detect a `405` from the merge call and return not-ready (`nil, false, nil`) so the step's `wait` loop returns `Running` and retries on the next reconciliation. Because the gate already rejects the permanent conditions *before* the merge call, a 405 here is the transient TOCTOU race; and the retry re-runs the gate, which surfaces any condition that has since become permanent (e.g. a real conflict) as such.

> Note: the retry rides the `wait: true → Running` path — the documented mechanism for not-ready retries. Set `wait: true` on the `git-merge-pr` step to benefit.

## Testing

- **Unit** (`TestMergePullRequest`): cases for `clean`, `unknown`, `behind`, `dirty`, and the merge-time `405`. Existing cases (which leave `mergeable_state` unset) exercise the boolean fallback path unchanged.
- **Integration** (`TestMergeGate`, build-tagged `integration && github`): drives a real GitHub repo into `clean` / `dirty` / `blocked` and asserts the gate against the live API. Verified in the dev container. Setup is documented in `pkg/gitprovider/testing/README.md`. The shared `RunPRTests` cleanup was also made hang-free (benefits all providers).

Fixes #5761

### DIFF
--- a/pkg/gitprovider/github/github.go
+++ b/pkg/gitprovider/github/github.go
@@ -52,6 +52,13 @@ const (
 	mergeableStateUnknown = "unknown"
 )
 
+// mergeBaseModifiedMsg is the substring GitHub includes in the message of a 405
+// merge error when the base branch moved between the mergeability check and the
+// merge call. This is the one transient 405 (it clears on retry); all other 405s
+// from the merge endpoint are permanent. Full message: "Base branch was
+// modified. Review and try the merge again."
+const mergeBaseModifiedMsg = "Base branch was modified"
+
 var registration = gitprovider.Registration{
 	Predicate: func(repoURL string) bool {
 		u, err := url.Parse(repoURL)
@@ -368,17 +375,25 @@ func (p *provider) MergePullRequest(
 		&github.PullRequestOptions{MergeMethod: opts.MergeMethod},
 	)
 	if err != nil {
-		// GitHub returns 405 ("Method Not Allowed") when the merge cannot be
-		// performed -- most notably "Base branch was modified" when the base tip
-		// moved between the mergeability check above and this merge call (a TOCTOU
-		// race that fires whenever concurrent merges land on the same base). The
-		// gate above already rejected the permanent conditions (conflict, blocked,
-		// etc.), so a 405 here is transient: report "not ready" so the caller
-		// retries. The next attempt re-runs the gate, which surfaces any condition
-		// that has since become permanent (e.g. a real conflict) as such.
+		// GitHub returns 405 ("Method Not Allowed") whenever the merge cannot be
+		// performed, which spans both transient and permanent conditions. The one
+		// transient case is "Base branch was modified": the base tip moved between
+		// the mergeability check above and this merge call (a TOCTOU race that
+		// fires when concurrent merges land on the same base). It is the only 405
+		// we retry -- report "not ready" so the caller tries again on the next
+		// reconciliation.
+		//
+		// Other 405s are permanent and must not be retried, or the step would loop
+		// forever under wait=true. The most common is a configured merge method
+		// that is disabled on the repo (e.g. mergeMethod: squash with squash
+		// merges turned off): the gate sees the PR as mergeable, so every retry
+		// would re-attempt and re-fail. We fall through and surface GitHub's
+		// message (carried by the wrapped error) so the misconfiguration is
+		// visible instead of silently looping.
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) && ghErr.Response != nil &&
-			ghErr.Response.StatusCode == http.StatusMethodNotAllowed {
+			ghErr.Response.StatusCode == http.StatusMethodNotAllowed &&
+			strings.Contains(ghErr.Message, mergeBaseModifiedMsg) {
 			return nil, false, nil
 		}
 		return nil, false, fmt.Errorf("error merging pull request %d: %w", id, err)
--- a/pkg/gitprovider/github/github.go
+++ b/pkg/gitprovider/github/github.go
@@ -2,7 +2,9 @@ package github
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


EronWright:
## Verified against live GitHub, with legacy-vs-new comparison

`TestMergeGate` (new, build-tagged `integration && github`) drives a real PR into each `mergeable_state` and logs both what the **legacy** `Draft || !Mergeable` gate would have done and what the **new** `mergeable_state` gate does. Run in the dev container against a real repo:

| PR state | `mergeable_state` | `mergeable` | legacy gate | new gate | |
|---|---|---|---|---|---|
| non-conflicting | `clean` | `true` | merge | merge | same |
| conflicting | `dirty` | `false` | **retry forever** | fail (terminal error) | **DIVERGES** |
| required check unmet | `blocked` | `true` | **attempt merge → GitHub rejects** | not-ready (retry) | **DIVERGES** |

The two divergences are exactly the bugs this fixes: the legacy gate retries a permanent conflict indefinitely, and it tries to merge a `blocked` PR that GitHub then rejects. (The `blocked` improvement is a bonus beyond the original ask — a PR merely waiting on a required check is now correctly treated as not-ready instead of attempt-and-fail.)

Setup/run instructions are in `pkg/gitprovider/testing/README.md`.

fuskovic:
@EronWright According to the [Github docs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2026-03-10#merge-a-pull-request--status-codes) a 405 can be produced if a merge cannot be performed which I think is broader than just merge state. More specifically, configuration level. e.g. If the step is configured for a squash merge but squash merges are not enabled on the repo that would ( in theory ) produce a 405. In that scenario if `wait: true` we would loop indefinitely. Can you please verify this? When doing so, you may want to debug log the `ghErr.Message` ( which we're not currently inspecting ). I suspect we could use that field to distinguish between 405 due to merge state or 405 due to configuration not allowing the merge method specified.

krancour (approved):
🔥🔥🔥

EronWright:
@fuskovic Good catch — verified, and you're right.

I reproduced it against live GitHub: a squash merge on a repo with squash merges disabled returns `405 "Squash merges are not allowed on this repository."` With the previous "any 405 → not ready" handling and `wait: true`, that would have looped forever (the gate sees the PR as mergeable, so every retry re-attempts and re-fails) — exactly your scenario.

Fixed in 17e96876 by narrowing the retry to the **one** transient merge 405, distinguished via `ghErr.Message` as you suggested:

- 405 whose message contains `"Base branch was modified"` → not-ready (retried). This is the only transient merge 405 ([community #24462](https://github.com/orgs/community/discussions/24462)).
- **Any other 405** (disabled merge method, etc.) → terminal error. GitHub's message is surfaced through the wrapped error (go-github's `(*ErrorResponse).Error()` includes `.Message`), so the misconfiguration is visible instead of silently looping.

On the debug-log suggestion: we now both classify on `Message` and surface it via the returned error. The git providers don't currently take a logger, so I surfaced it through the error rather than introducing logging — happy to add a debug line if you'd prefer.

Added a unit test using the real `"Squash merges are not allowed on this repository."` message asserting it's terminal; the existing base-branch-modified case still returns not-ready.

PR #6492, opened 2026-06-24
feat(ui): display and support Secret type for generic secrets

### LINKED ISSUE #6445: Secret management UI should display and support Secret `type`
### Pre-submission checklist

### Proposed Feature

I want to input a Secret in Kargo of `type: kubernetes.io/dockerconfigjson`

However, the UI won't let me control the type. 

Image

Nor will it show the types of my secrets:

Image

### Motivation and Use Case

`kubernetes.io/dockerconfigjson` is needed for `imagePullSecrets`, which are used by Job AnalysisRuns.

We already support different secret types. The problem is that the only way to get it into the project is by using `kargo apply`. So UI just needs to support this option.

NOTE: Secret type is immutable, so edit secret would not support changing the type. It would only be during secret creation. And read-only/displayed on secret edit.

### Alternatives Considered

_No response_

Marvin9:
## Description

Closes #6445

Adds support for the Kubernetes Secret `type` field to the generic secrets ("Add Secret") UI flow. Previously the UI always created `Opaque` Secrets and never surfaced a Secret's type, so types like `kubernetes.io/dockerconfigjson` (needed for `imagePullSecrets`) could only be created via `kargo apply`.

- **Backend:** the generic-credentials *create* REST API now accepts an optional `type`, which is applied to the created Secret. When empty, Kubernetes defaults it to `Opaque`. Secret type is immutable, so it is only settable on create; the read path already preserves it (no change needed there). Swagger spec and Go client regenerated.
- **UI:** a "Type" field (AutoComplete defaulting to `Opaque`, with common Kubernetes secret types as suggestions) on the create modal, shown read-only when editing since the type is immutable; a "Type" column in the generic secrets list.
- **Cleanup:** unified the repo and generic form schemas into a single zod schema and derived the form value type via `z.infer`, removing the `@ts-expect-error` directives and the schema-reassignment hack.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/pkg/client/generated/models/create_generic_credentials_request.go
+++ b/pkg/client/generated/models/create_generic_credentials_request.go
@@ -25,6 +25,11 @@ type CreateGenericCredentialsRequest struct {
 
 	// replicate
 	Replicate bool `json:"replicate,omitempty"`
+
+	// Type is the Kubernetes Secret type (e.g. "Opaque" or
+	// "kubernetes.io/dockerconfigjson"). It is immutable, so it may only be set
+	// at creation time. When empty, Kubernetes defaults it to "Opaque".
+	Type string `json:"type,omitempty"`
 }
 
 // Validate validates this create generic credentials request
--- a/pkg/server/create_generic_credentials_v1alpha1.go
+++ b/pkg/server/create_generic_credentials_v1alpha1.go
@@ -22,6 +22,7 @@ type genericCredentials struct {
 	name        string
 	description string
 	replicate   bool
+	secretType  string
 	data        map[string]string
 }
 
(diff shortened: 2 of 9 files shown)


Marvin9:
https://github.com/user-attachments/assets/27d42370-a221-4c49-9be4-28200a4c7cb0

Marvin9:
I think the label "Type" becomes explanatory when value is default to "Opaque". The reason why I just used "Type" as label in UI. Happy to change if you think otherwise

rpelczar (approved):
UI LGTM

jessesuen (approved):
Backend LGTM

Issue #5727, opened 2026-02-11
Configurable backoff for http steps


zoeyfyi:
__Note:__ If you're requesting a Helm chart option that may be very niche and not useful to the community at large, please consider using Kustomize to apply "last mile" tweaks to the output of `helm template` to suit your needs instead.

# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.
* [ ] I've pasted the output of `kargo version`, if applicable.
* [ ] I've pasted logs, if applicable.

# Proposed Feature

Http requests get retried with progressive backoff, but its not possible to change the parameters. This preposal would add backoff settings to the http step configuration.

# Motivation

In some scenarios, we want to poll more aggressively, in other scenarios we are ok with the step eventually completing. Tuning the backoff parameters would be incredibly useful here.

krancour:
It's not as simple as it sounds. The existing backoff behavior you're observing is just what happens in Kubernetes when a reconciliation fails. The get fails --> the step fails --> the attempt to reconcile the promotion fails --> the Promotion gets scheduled for another reconciliation attempt no sooner that some interval that's a function of the number of consecutive failures. The backoff/retry being handled at that high level unblocks the controller from working on _other_ Promotions, which is important because only a small number of them can be reconciled concurrently.

It sounds like you're asking for the `http` step to handle retry/backoff _internally_. This means that reconciliation attempt remains in-progress for an extended period of time, blocking other work from being done.

I get what you want and why you want it. I think it's not the first time this has been requested. We just haven't quite figured out yet how to do this without creating an extremely deleterious impact on performance.

krancour:
Closing in favor of https://github.com/akuity/kargo/issues/6310

PR #6493, opened 2026-06-24
feat(promotion): suggested poll interval for polling steps

### LINKED ISSUE #6310: support suggested requeue interval for select steps
We've turned down many requests to make the requeue interval tunable on various reconcilers. The short version: cranking it down -- which is almost certainly what everyone will do -- is more likely to degrade overall system performance than improve it. The preferred alternative, which we've been pursuing incrementally, is to improve responsiveness to specific events instead:

- v1.7.0 introduced webhook receivers that trigger reconciliation of specific Warehouses (and thus artifact discovery and Freight production) on push events to subscribed repositories. Every release since has built on this.

- v1.10.0 fixed issues that made the Stage reconciler less responsive to Argo CD Application status changes than it should have been.

Requests to tune the Warehouse and Stage reconcilers have dropped off accordingly. The Promotion reconciler now stands out, and with good reason.

The pain points cited most often are the `git-wait-for-pr` and `http` steps. Both poll, and -- deliberately -- neither loops internally. Controllers reconcile a small, finite number of resources concurrently; an in-step poll loop would block reconciliation of *other* Promotions until it resolved. So the model is: poll once per reconciliation. If neither success nor failure criteria are met, the Promotion goes back on the queue and, absent some other event, is reconciled again at the regular interval -- five minutes. A Promotion that could clear its step with a few retries over a few seconds ends up waiting at least five minutes to resume.

I stand by my position that cranking the interval down is a nuclear option, and prefer targeted improvements that obviate it. #5852, landing in v1.11.0, extends the v1.7.0 webhook receivers to trigger Promotion reconciliation on PR merge events -- addressing `git-wait-for-pr`.

That leaves the `http` step (and possibly others) as the remaining driver. My proposal: make a *suggested* polling frequency configurable on the affected step(s). The once-per-reconciliation model doesn't change. Instead, when a polling step reports back as "running" (yielding resources to other Promotions), it suggests the interval for next reconciliation based on its configured frequency.

The capability to do this already exists; no step currently uses it.

Caveats:

1. It really is just a *suggestion*. Events can pull reconciliation sooner; load can push it later. That's how Kubernetes works.

2. We need a lower bound. Reconciling a Promotion every second is unwise: with enough concurrent Promotions, CPU rises sharply and throughput drops, since more frequent reconciliation of *n* Promotions means each one waits longer for its turn. A floor of 10s would let `http` steps resolve much faster than today, with less risk.

3. Progressive backoff as requested in #5727 seems it would be a nice alternative to a lower bound, however, this is much harder to do. Controller runtime gives us progressive backoff for free when a reconcilation errors, but effecting backoff when things are working smoothly requires us to implement it ourselves. Although the algorithm is easy, this would require that we track polling attempts across reconciliations. Let us deem this out of scope and reserve the progressive backoff as a possible, future improvement.

To do:

1. Identify the steps that could benefit -- there shouldn't be many.

2. Implement a uniform set of optional configuration options across them. Absent explicit configuration, any affected step should not recommend an interval to the orchestrator, preserving the current behavior.

EronWright:
## Description

Closes #6310

Polling steps that report `Running` while waiting on an external condition previously had no say in when the Promotion is next reconciled. Absent some other event, they waited the full default requeue interval (5 minutes), so a step that could clear in seconds stalled for minutes.

This introduces an optional, per-step **`pollInterval`** config on the three steps that poll — `http`, `git-wait-for-pr`, and `git-merge-pr` — that each uses to *suggest* a requeue interval (via `StepResult.RetryAfter`) when it returns `Running`.

This implements the plan agreed on the issue, including the resolution that **polling is distinct from retry** (error handling): the field lives on step config rather than the step-level `retry` policy.

## How

- **`pollInterval` field** added to the three steps' JSON schemas, regenerated Go config types (`zz_config_types.go`), and the UI directive schemas.
- **Resolution**: explicit config → per-step default, via a shared `resolvePollInterval` helper. Defaults: `http` 30s, `git-wait-for-pr` 30s, `git-merge-pr` 10s. Shipping defaults means the median experience improves out of the box — not only for users who discover the knob. Opt out of the faster cadence by setting a larger value (e.g. `pollInterval: 5m`).
- **Lower bound** of 10s enforced centrally in `calculateRequeueInterval` (`minSuggestedRequeueInterval`), clamping any sub-floor suggestion (logged at debug). No step — now or future — can drive a pathologically short interval. The value remains only a suggestion: events can pull reconciliation sooner; load can push it later.
- **Retires `git-merge-pr`'s internal 3×5s sleep loop**, which blocked a reconciliation worker for up to ~15s, contrary to the poll-once-per-reconciliation model. It now attempts the merge once and, when `wait: true`, reports `Running` with the suggested interval.

Progressive backoff (#5727) remains deferred; a fixed `pollInterval` with a floor naturally becomes its base interval if it lands later.

## Examples

```yaml
# http: poll a rollout status endpoint every 15s instead of waiting up to 5m
- uses: http
  config:
    method: GET
    url: https://example.com/api/rollout/status
    successExpression: response.status == 200 && response.body.phase == "Healthy"
    failureExpression: response.body.phase == "Degraded"
    pollInterval: 15s
```

```yaml
# git-wait-for-pr: poll every minute. If a PR-closed webhook receiver is
# configured, reconciliation is still triggered near-instantly on merge;
# pollInterval governs the fallback polling cadence.
- uses: git-wait-for-pr
  config:
    prNumber: ${{ outputs['open-pr'].pr.id }}
    pollInterval: 1m
```

```yaml
# git-merge-pr: re-attempt the merge every 10s (the default) until the PR
# becomes mergeable, rather than blocking a worker in an internal sleep loop.
- uses: git-merge-pr
  config:
    prNumber: ${{ outputs['open-pr'].pr.id }}
    wait: true
    pollInterval: 10s
```

Omitting `pollInterval` keeps each step's default cadence. The interval is also expression-evaluable, like other step config, so it can be set dynamically.

## Testing

- **Unit** — added `pollInterval` schema-validation cases and `Running`-path `RetryAfter` assertions (default and explicit) for all three steps; a dedicated `resolvePollInterval` test; and floor clamp/honor cases in `Test_calculateRequeueInterval`.
- `make lint-go` clean; affected packages pass under `-race`.

### Eligibility

- [x] Linked to an existing issue with no blocking labels.

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:


fuskovic (approved):
LGTM

PR #6494, opened 2026-06-25
feat: add discoveredAt field to Freight and use it for time-based ordering

### LINKED ISSUE #5855: Freight should manage its own `discoveryTimestamp`, not rely on `creationTimestamp`

jessesuen:
## Description

Resolves https://github.com/akuity/kargo/issues/5855

- Adds a `discoveredAt` field to the Freight CRD representing when the artifact was first discovered. A mutating webhook initializes it on both creates (using `metav1.Now()`) and updates (backfilling from `creationTimestamp` for pre-existing Freight).
- Adds `EffectiveDiscoveredAt()` helper on Freight that returns `discoveredAt` when set, falling back to `creationTimestamp`.
- Replaces all uses of `Freight.CreationTimestamp` for ordering/age with `EffectiveDiscoveredAt()` — covering auto-promotion (stage reconciler), REST API freight sorting, CLI get freight age display, garbage collector sort and min-age guard, and Freight events.

## Demo

https://github.com/user-attachments/assets/224575c8-12f9-472e-89c3-78d988e18307

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:


pkg/webhook/kubernetes/freight/webhook.go
```diff
@@ -129,6 +130,18 @@ func (w *webhook) Default(ctx context.Context, obj runtime.Object) error {
...
 
+	// Initialize discoveredAt when unset, regardless of whether this is a create
+	// or update. For updates this backfills the field on pre-existing Freight.
+	if freight.DiscoveredAt == nil {
+		if !freight.CreationTimestamp.IsZero() {
+			t := freight.CreationTimestamp
+			freight.DiscoveredAt = &t
+		} else {
+			t := metav1.Now()
+			freight.DiscoveredAt = &t
```
thomastaylor312: Super duper nit: I think both of these don't need var assignments. The first one can be `&freight.CreationTimestamp` or `&(freight.CreationTimestamp)` (I can't remember which one Go allows) and the second can be `new(metav1.Now())`

thomastaylor312 (approved):

PR #6496, opened 2026-06-25
feat(promotion): expose ctx.targetFreight.alias in expression language

### LINKED ISSUE #4584: add the current Freight's alias to `ctx`
# Checklist

# Proposed Feature

Add a way to reference the alias of `ctx.targetFreight` like we can with the name.

# Motivation

The slack message I can generate only has the long name of the freight (the uuid). This is fine for linking, but I wanted to be able to use the alias as well for a shorter friendlier name.

https://docs.kargo.io/user-guide/reference-docs/expressions#context-ctx-object-structure

jessesuen:
## Description

Resolves https://github.com/akuity/kargo/issues/4584

**Feature:**
- Adds `ctx.targetFreight.alias` to the promotion expression language environment, giving step authors access to the human-friendly alias of the Freight being promoted.

**Refactor:**
- Adds `TargetFreightAlias` to StepContext so that external step executors receiving a serialized StepContext (e.g. pod-based runners) can access the alias without needing the full promotion Context.
- Exports `BuildCtxMap(StepContext)` `map[string]any` as a package-level function so that external executors can construct a consistent expression environment without reimplementing the field mapping.
- Promotes `BuildExprEnv` from a method on `StepEvaluator` to a package-level function — it never used receiver state, and the cleanup makes the public surface clearer.

### Demo

https://github.com/user-attachments/assets/1251a24b-3a28-41e8-9386-2d2e9346ada1

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
+++ b/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
@@ -197,13 +197,22 @@ ctx
 ├── stage: string             # The name of the Stage
 ├── promotion: string         # The name of the Promotion
 ├── targetFreight
-│   ├── name: string          # The name of the Freight that is initiated this Promotion
+│   ├── name: string          # The name of the Freight that initiated this Promotion
+│   ├── alias: string         # The human-friendly alias of the Freight
 │   └── origin
 │       └── name: string      # The name of the Warehouse that contains the Freight
 └── meta
     └── promotion
-        └── actor: string     # The creator of the Promotion
+        ├── actor: string     # The creator of the Promotion
+        └── rollback: bool    # Whether this Promotion is a rollback
 ```
+
+:::warning
+Freight aliases (`ctx.targetFreight.alias`) are mutable — they can be changed
+by users at any time after Freight is created. Avoid using an alias anywhere
+its value must remain stable, such as in git commit messages, artifact paths,
+or any other durable record.
+:::
 
 The following example promotion process clones a repository and checks out
 two branches to different directories, uses Kustomize with source from one
--- a/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
+++ b/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
@@ -193,6 +193,7 @@ The `ctx` object has the following structure:
 
 ```console
 ctx
+├── uiBaseUrl: string         # The base URL of the Kargo UI
 ├── project: string           # The name of the Project
 ├── stage: string             # The name of the Stage
 ├── promotion: string         # The name of the Promotion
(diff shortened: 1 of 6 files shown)


pkg/promotion/evaluator.go
```diff
@@ -89,35 +89,59 @@ func ExprEnvWithTaskOutputs(alias string, outputs State) ExprEnvOption {
...
+	env := BuildCtxMap(StepContext{
+		UIBaseURL:          promoCtx.UIBaseURL,
+		Project:            promoCtx.Project,
+		Promotion:          promoCtx.Promotion,
+		Stage:              promoCtx.Stage,
+		TargetFreightRef:   promoCtx.TargetFreightRef,
+		TargetFreightAlias: promoCtx.TargetFreightAlias,
+		PromotionActor:     promoCtx.Actor,
+		Rollback:           promoCtx.Rollback,
+	})
```
fuskovic, line 151: Should we be propogating the step alias here?

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -197,14 +197,23 @@ ctx
 ├── stage: string             # The name of the Stage
 ├── promotion: string         # The name of the Promotion
 ├── targetFreight
-│   ├── name: string          # The name of the Freight that is initiated this Promotion
+│   ├── name: string          # The name of the Freight that initiated this Promotion
+│   ├── alias: string         # The human-friendly alias of the Freight
 │   └── origin
 │       └── name: string      # The name of the Warehouse that contains the Freight
 └── meta
```
hairyhum, line 205: We can also mention `meta.step.alias`, since it's generally accessible now

fuskovic (changes requested):

pkg/promotion/evaluator.go
```diff
@@ -89,35 +89,59 @@ func ExprEnvWithTaskOutputs(alias string, outputs State) ExprEnvOption {
...
+
+// BuildExprEnv builds an environment map for evaluating expressions in
+// Promotion steps. The environment includes context information, such as the
+// Project, Promotion, Stage, and target Freight reference.
+//
+// The environment can be extended with additional options provided via the
+// ExprEnvOption functional options. These options can be used to add variables
+// or modify the expression language environment.
+func BuildExprEnv(promoCtx Context, opts ...ExprEnvOption) map[string]any {
+	env := BuildCtxMap(StepContext{
```
hairyhum, line 141: One artifact of this would be that `meta.step.alias` will be present in `ctx`, but it will always be empty. Even if evaluated in the context of a step.
While this doesn't impact existing logic or templates people might have, it might get confusing as of why it's there and why is it empty.
It shouldn't be too hard to add step here, since it's present in most callers of this function. Maybe we should do that?

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
jessesuen: Good suggestion. Also added `uiBaseUrl`

jessesuen (commented):
Thanks for the feedback. Added the suggestions.

fuskovic (approved):

PR #6516, opened 2026-06-29
fix(controller): deflake startVerification via injected clock

EronWright:
## Summary

`TestRegularStageReconciler_startVerification` is flaky on `main`. The "finds existing analysis run" cases assert the returned `VerificationInfo.FinishTime` against a fixed time, but production `startVerification` ignored the caller's injected reconciliation time and read the wall clock again via `metav1.Now()`. When the two reads straddled a one-second boundary, `FinishTime` landed a second after `StartTime` and the assertion failed by 1 — frequently on the slower CI runner.

## Changes

- **`regular_stages.go`**: thread the `endTime func() time.Time` clock already used by the sibling `verifyStageFreight` into `startVerification`, `getVerificationResult`, and `abortVerification`. Every synchronous-return path now stamps `FinishTime` from the injected clock instead of `metav1.Now()`, so `startTime <= finishTime` holds deterministically and all three functions become time-testable. Paths that use the AnalysisRun's own `CreationTimestamp`/`CompletedAt`, or carry forward an existing `FinishTime`, are unchanged.
- **`regular_stages_test.go`**: inject a fixed end clock and assert exact `FinishTime` equality across the three functions' tests, mirroring the existing `TestRegularStageReconciler_verifyStageFreight` pattern. In the `startVerification` test, the injected start value is renamed `now` → `startTime` so the assertions read against the value actually passed in.

## Test

```
go test ./pkg/controller/stages/ -run TestRegularStageReconciler -count=20
ok  github.com/akuity/kargo/pkg/controller/stages
```

20 iterations pass with no off-by-one failures.

### DIFF
--- a/pkg/controller/stages/regular_stages.go
+++ b/pkg/controller/stages/regular_stages.go
@@ -1071,7 +1071,7 @@ func (r *RegularStageReconciler) verifyStageFreight(
 				logger.Debug("aborting verification of Stage Freight")
 
 				// Abort the verification.
-				newVI, err = r.abortVerification(ctx, *curFreight, abortReq)
+				newVI, err = r.abortVerification(ctx, *curFreight, abortReq, endTime)
 				if newVI != nil {
 					newStatus.FreightHistory.Current().VerificationHistory.UpdateOrPush(*newVI)
 				}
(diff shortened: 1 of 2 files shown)


fuskovic (approved):
LGTM

PR #6518, opened 2026-06-30
fix(ui): Feature/clone freight from history

### LINKED ISSUE #6517: Historical freight entries in Stage history cannot be cloned after the source Freight is deleted
### Pre-submission checklist

### Description

When a freight has already been promoted to a Stage, it remains visible in the stage it's history. If that Freight resource is deleted by garbage collection or manually, clicking the historical entry no longer opens the assemble Freight UI. So historical entries become unusable once the freight is gone.

[PR 6246](https://github.com/akuity/kargo/pull/6246) addresses a related (but different) problem; it keeps clone freight usable when the source freight's artifact versions have fallen out of the Warehouse discovery by injecting those versions back into discoveredArtifacts.

I will create a PR with a fix for this.

### Steps to Reproduce

1. Promote a Freight to a stage.
2. Delete the original Freight resource
3. Open the Stage history.
4. Click the older Freight entry.

### Expected Behavior

The historical freight entry should open the assemble Freight UI, using the Freight data preserved in Stage history, so we can clone and or edit it.

### Actual Behavior

Nothing happens, because the UI tries to fetch the deleted Freight resource and cannot reconstruct the historical freight.

### Kargo Version

1.10.8

kevinetore:
Closes #6517

This PR allows users to reconstruct and clone a historical Freight from a stage's Freight history, even when the original Freight resource has been removed by garbage collection.

Instead of failing when the referenced Freight no longer exists, the UI reconstructs a Freight from the historical Stage data and opens the existing assemble freight workflow. This allows users to modify versions before assembling a new Freight.

### DIFF
--- a/ui/src/features/assemble-freight/clone-freight-note.tsx
+++ b/ui/src/features/assemble-freight/clone-freight-note.tsx
@@ -15,7 +15,9 @@ export const CloneFreightNote = (props: {
   missingArtifacts: (Image | GitCommit | Chart)[];
   className?: string;
 }) => {
-  if (!props.cloneFreight) {
+  const cloneFreight = props.cloneFreight;
+
+  if (!cloneFreight) {
     return null;
   }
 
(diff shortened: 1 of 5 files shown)


kevinetore:
Could someone from the Kargo team review this PR? This functionality is really needed by our teams. @krancour

PR #6519, opened 2026-06-30
feat(credentials/ecr): add cross-account ECR registry support

rv9812:
## Description

Closes #5587

When the ECR registry URL belongs to a different AWS account than the controller, Kargo now attempts to assume `kargo-project-<project>` in the ECR account before falling back to the controller's own IAM role.

Resolution order:
1. Assume `kargo-project-<project>` in the controller's account (existing)
2. If ECR account != controller account, assume `kargo-project-<project>` in the ECR account (new)
3. Fall back to the controller's IAM role directly (existing)

Same-account setups are unaffected — step 2 is skipped entirely.

Extracted account ID from ECR URL via `ecrURLAccountRegex` (`common.go`) and threaded it through `getAuthToken`. Refactored the fallback logic into `getAuthTokenWithRole` to avoid duplication.

Docs: updated `ambient-credentials.md` to document the 3-step flow and the IAM trust/permission requirements for cross-account registries.

### Eligibility

- [x] Linked to an existing issue with no blocking labels.
- [ ] Changes documentation only.
- [ ] Changes ten lines or fewer.

### Quality

- [x] Adds or updates corresponding tests.
- [x] Adds or updates corresponding documentation.

### AI Use Disclosure

- [ ] By a human _without_ AI assistance.
- [x] By a human _with_ AI assistance. A human has reviewed every line prior to opening the PR.
- [ ] By an AI _with human supervision._ A human has reviewed every line prior to opening the PR.
- [ ] Entirely by an AI. No human has reviewed this prior to opening the PR.

### Sign-Off

- [x] Are signed off by their author (`git commit -s`)
- [ ] Are cryptographically signed (`git commit -S`) **(encouraged)**

PR #6521, opened 2026-06-30
feat(credentials/ecr): add cross-account ECR registry support

### LINKED ISSUE #5587: Support cross-account IAM Role assumption for AWS ECR access in Ambient Credentials
# Checklist

# Proposed Feature

Enable Kargo's ambient credential provider to support cross-account ECR access by dynamically assuming IAM roles in the account where the ECR registry resides.

## Current Behavior

The ``ManagedIdentityProvider`` always uses the controller's AWS account ID (from ``sts:GetCallerIdentity``) to construct the role ARN. Accessing ECR in a different account requires granting cross-account permissions to the controller role, which is less secure and more complex to manage.

## Proposed Behavior

- Extract account ID from ECR URL — The URL already contains the target account ID (``${account_id}.dkr.ecr.${region}.amazonaws.com``)
- Auth token workflow:
  - Try assume ``kargo-project-<project>`` in controller account (A) as usual
  - If fails AND ECR account differs (current AWS account ≠ AWS account in repo URL) → try assume ``kargo-project-<project>`` in ECR account (B)
  - Fall back to controller role directly

# Motivation

- Multi-account AWS architectures are common. ECR registries can be spread across multiple accounts
- Keeping ECR permissions scoped to the ECR account follows least privilege

krancour:
Supersedes #6519

Fixes #5587

cc @rv9812

### DIFF
--- a/pkg/credentials/ecr/common.go
+++ b/pkg/credentials/ecr/common.go
@@ -9,8 +9,12 @@ import (
 
 const tokenCacheExpiryMargin = 5 * time.Minute
 
-// ecrURLRegex is a regex that matches ECR URLs.
+// ecrURLRegex is a regex that matches ECR URLs and captures the region.
 var ecrURLRegex = regexp.MustCompile(`^[0-9]{12}\.dkr\.ecr\.(.+)\.amazonaws\.com/`)
+
+// ecrURLAccountRegex is a regex that matches ECR URLs and captures the account
+// ID and region.
+var ecrURLAccountRegex = regexp.MustCompile(`^([0-9]{12})\.dkr\.ecr\.(.+)\.amazonaws\.com/`)
 
 // tokenCacheKey returns a cache key in the form of a hash for the given parts.
 // Using a hash ensures that any sensitive data is not stored in a decodable
--- a/pkg/credentials/ecr/common.go
+++ b/pkg/credentials/ecr/common.go
@@ -9,12 +9,9 @@ import (
 
 const tokenCacheExpiryMargin = 5 * time.Minute
 
-// ecrURLRegex is a regex that matches ECR URLs and captures the region.
-var ecrURLRegex = regexp.MustCompile(`^[0-9]{12}\.dkr\.ecr\.(.+)\.amazonaws\.com/`)
-
-// ecrURLAccountRegex is a regex that matches ECR URLs and captures the account
-// ID and region.
-var ecrURLAccountRegex = regexp.MustCompile(`^([0-9]{12})\.dkr\.ecr\.(.+)\.amazonaws\.com/`)
+// ecrURLRegex is a regex that matches ECR URLs and captures the account ID (group 1)
+// and region (group 2).
+var ecrURLRegex = regexp.MustCompile(`^([0-9]{12})\.dkr\.ecr\.(.+)\.amazonaws\.com/`)
 
 // tokenCacheKey returns a cache key in the form of a hash for the given parts.
 // Using a hash ensures that any sensitive data is not stored in a decodable
--- a/pkg/credentials/ecr/managed_identity.go
+++ b/pkg/credentials/ecr/managed_identity.go
@@ -212,7 +212,7 @@ func (p *ManagedIdentityProvider) getAuthToken(
 		return token, expiry, nil
 	}
 
-	// Step 3: fall back to the controller's IAM role directly.
+	// Fall back to the controller's IAM role directly.
 	logger.Debug(
 		"Falling back to using controller's IAM role directly.",
 	)
--- a/pkg/credentials/ecr/managed_identity.go
+++ b/pkg/credentials/ecr/managed_identity.go
@@ -204,7 +204,6 @@ func (p *ManagedIdentityProvider) getAuthToken(
 		fmt.Sprintf(roleARNFormat, p.accountID, project),
 		region,
 		stsSvc,
-		logger,
 	)
 	if err == nil && token != "" {
 		return token, expiry, nil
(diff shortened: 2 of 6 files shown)


pkg/credentials/ecr/common.go
```diff
@@ -9,9 +9,13 @@ import (
...
 const tokenCacheExpiryMargin = 5 * time.Minute
 
-// ecrURLRegex is a regex that matches ECR URLs.
+// ecrURLRegex is a regex that matches ECR URLs and captures the region.
 var ecrURLRegex = regexp.MustCompile(`^[0-9]{12}\.dkr\.ecr\.(.+)\.amazonaws\.com/`)
 
+// ecrURLAccountRegex is a regex that matches ECR URLs and captures the account
+// ID and region.
+var ecrURLAccountRegex = regexp.MustCompile(`^([0-9]{12})\.dkr\.ecr\.(.+)\.amazonaws\.com/`)
+
```
krancour: Why have two separate regular expressions? They're close to identical.

Couldn't `ecrURLRegex` be redefined to include both capture groups (like `ecrURLAccountRegex` currently does), then elsewhere in the code, use only that one regex and discard what the first group captures in cases where it is not needed?

rv9812: Spot on. We can consolidate `ecrURLRegex `to capture both account ID (group 1) and region (group 2). Then wherever only the region is needed, we just discard group 1 and use group 2. 
Eliminates the duplication and makes this way easier to maintain.

pkg/credentials/ecr/managed_identity.go
```diff
@@ -113,14 +114,15 @@ func (p *ManagedIdentityProvider) GetCredentials(
...
+	// Extract the account ID and region from the ECR URL
 	matches := ecrURLRegex.FindStringSubmatch(req.RepoURL)
-	if len(matches) != 2 { // This doesn't look like an ECR URL
+	if len(matches) != 3 { // This doesn't look like an ECR URL
 		return nil, nil
 	}
 
-	region := matches[1]
-	cacheKey := tokenCacheKey(region, req.Project)
+	ecrAccountID := matches[1]
```
krancour: Nit: idk that there's such a thing as an "ECR account." It's really an AWS account. And rather than call it `awsAccountID`, I'd suggest that the context is well understood and we could get away with _just_ `accountID`, which despite being less specific is still more accurate than the current name.

pkg/credentials/ecr/managed_identity.go
```diff
@@ -179,56 +189,109 @@ func (p *ManagedIdentityProvider) getAuthToken(
...
+			ctx,
+			fmt.Sprintf(roleARNFormat, ecrAccountID, project),
+			region,
+			stsSvc,
+			logger,
 		)
+		if err == nil && token != "" {
+			return token, expiry, nil
+		}
+	}
```
krancour: Doesn't all of this logically reduce to just this?

```
if ecrAccountID != p.accountID {
	logger.Debug(
		"ECR registry is in a different account than the controller; " +
			"attempting to assume project-specific role in ECR account",
	)
}
token, expiry, err = p.getAuthTokenWithRole(
	ctx,
	fmt.Sprintf(roleARNFormat, ecrAccountID, project),
	region,
	stsSvc,
	logger,
)
if err == nil && token != "" {
	return token, expiry, nil
}
```

This has the added benefit of never attempting what you currently call step 1 when the registry is known to be in a different account.

pkg/credentials/ecr/managed_identity.go
```diff
@@ -179,56 +189,109 @@ func (p *ManagedIdentityProvider) getAuthToken(
...
+
+// getAuthTokenWithRole attempts to obtain an ECR auth token by assuming the
+// given IAM role. Returns an empty token (and nil error) if the role cannot be
+// assumed or does not have ECR permissions, so the caller can try alternatives.
+func (p *ManagedIdentityProvider) getAuthTokenWithRole(
+	ctx context.Context,
+	roleARN string,
+	region string,
+	stsSvc *sts.Client,
+	logger interface{ Debug(string, ...any) },
```
krancour: As a rule, we never pass loggers around. Ubiquitous as they are, function signatures throughout the entire code base would be polluted with what is basically noise. The conventional thing is to retrieve an appropriate logger from `ctx` using: `logging.LoggerFromContext(ctx)`

pkg/credentials/ecr/managed_identity.go
rv9812: Done — renamed `ecrAccountID` to `accountID` throughout. Also updated the log key from `ecrAccountID` to `accountID` and `awsAccountID` to `controllerAccountID` to avoid ambiguity between the two account IDs in the same log line.

pkg/credentials/ecr/managed_identity.go
rv9812: Good catch — you are right, it reduces cleanly to a single `getAuthTokenWithRole` call always using the registry's `accountID`, with the conditional debug log only when accounts differ. Done.

pkg/credentials/ecr/managed_identity.go
rv9812: Done — removed the `logger` parameter from `getAuthTokenWithRole` and retrieve it inside the function via `logging.LoggerFromContext(ctx)` instead.

krancour:
@rv9812 thank you for your work on this... and your patience. I just wanted to let you know this hasn't been forgotten. It will get what I hope is a final review after v1.11.0 ships.

krancour:
Closed in favor of #6722

PR #6542, opened 2026-07-02
docs: document automated rollbacks

jessesuen:
## Description

Documents Automated Rollback Feature for Kargo EE v1.11.

### Eligibility
- [x] Changes documentation only.

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/docs/docs/50-user-guide/20-how-to-guides/20-working-with-projects.md
+++ b/docs/docs/50-user-guide/20-how-to-guides/20-working-with-projects.md
@@ -50,10 +50,12 @@ creation. This includes things such as project-level RBAC resources and
 ## Project Configuration
 
 A `ProjectConfig` resource defines project-level configuration for an associated
-`Project`. At present, this only includes
+`Project`. This includes
 [promotion policies](#promotion-policies)
 that describe which `Stage`s are eligible for automatic promotion of newly
-available `Freight`.
+available `Freight`, as well as
+[auto-rollback](#auto-rollback) configuration for automatically reverting a
+`Stage` to a previously verified `Freight` when verification fails.
 
 The `ProjectConfig` resource must have the same name as its associated `Project`
 and be created in the `Namespace` of the `Project`. This separation of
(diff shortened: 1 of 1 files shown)


hairyhum (commented):
Looks good overall. Is decision not to mention multi-origin stages deliberate? Maybe we should make a separate page explaining that?
Also amount of em dashes seems excessive, is that an LLM generation artifact?

fuskovic (dismissed):
For @hairyhum's comment:

> Is decision not to mention multi-origin stages deliberate? Maybe we should make a separate page explaining that?

I agree we should mention it but I'm not sure another whole page is necessary. 

Maybe we can add it to the "Considerations" section as another bullet point instead.

e.g.

_**Multi-origin Stages**: For a `Stage` that subscribes to more than one `Warehouse`, verification runs once against the whole combined set of `Freight`. It does not indicate which origin's `Freight` caused the failure. The rollback controller tracks stable `Freight` independently per `Warehouse` origin and rolls back each drifted origin with its own `Promotion`. Because only one `Promotion` can be in flight per `Stage` at a time, a `Stage` with more than one drifted origin is rolled back one origin per `Promotion`. That happens across successive reconciliations rather than all at once. The `Stage` can be briefly left in a partially-rolled-back state where some origins are already at their stable `Freight` and others are not yet._

On the topic of the "Considerations" list, I'm not sure the "MatchUpstream stages are silently skipped" claim is true.

I can't find any `MatchUpstream/AutoPromotionSelectionPolicy` reference anywhere in the rollback controller or (the rest of the EE repo for that matter). Stable-Freight tracking is keyed only by Warehouse origin, regardless of selection policy.

Should we nix that claim?

krancour:
> For a Stage that subscribes to more than one Warehouse, verification runs once against the whole combined set of Freight. It does not indicate which origin's Freight caused the failure. The rollback controller tracks stable Freight independently per Warehouse origin and rolls back each drifted origin with its own Promotion. Because only one Promotion can be in flight per Stage at a time, a Stage with more than one drifted origin is rolled back one origin per Promotion. That happens across successive reconciliations rather than all at once. The Stage can be briefly left in a partially-rolled-back state where some origins are already at their stable Freight and others are not yet.

Question more than anything: Since Promotions only upgrade the Freight from _one_ Freight request (origin) at a time (which is already acknowledged here), if verification fails, I would think the failure _can_ actually be attributed to one specific piece of Freight -- the last one promoted -- and wouldn't that mean that things _can_ be rolled back to a stable state in a single gesture?

krancour:
I'm going to remove this from the v1.11.0 milestone just so I'm no longer tracking it as something blocking that release. The live docs can be updated after the fact by the backport we were already going to do anyway.

jessesuen:
Thanks for the feedback. Just getting back to this. I'll be implementing the following changes:

1. Since writing this, the `MatchUpstream` restriction is no longer true. We will now place an auto-promotion hold (if enabled), which solves the original concern of auto-rollback fighting MatchUpstream. I'll be removing this from docs.
2. I will add a note on how auto-rollback interplays with auto-promotion. Namely, how auto-rollback will cause an auto-promotion hold to be placed (assuming auto-promotion is enabled)
3. I'll explain how multi-origin works. But Kent is correct in that we can identify which origin caused the problem. The more interesting corner case is when:
* you have multi-origin Stage
* there were back-to-back promotions, and 
* auto-rollback is delayed because it was waiting for all the Promotions to settle before issuing a rollback

In this scenario, it will rollback in a FIFO manner, "undoing" the Freight from different origins in the same order that it was promoted (going back to the stable for each origin). In other words, it will issue up to N rollback promotions, where N is the number of origins.

fuskovic (approved):

PR #6543, opened 2026-07-02
docs: document access-credentials capability for custom steps

fuskovic:


krancour:
Closing in favor of implementing https://github.com/akuity/kargo/issues/6549 instead.

It would fill the same gap as this documentation.

PR #6545, opened 2026-07-02
refactor(tools): use `go tool` for Go-based dev tools instead of hack/tools

thomastaylor312:
Move Go-based dev tools (buf, controller-gen, go-to-protobuf, protoc-gen-gogo, protoc-gen-doc, goimports, swag, go-swagger, oapi-codegen, ctlptl, kind) into `tool` directives in the root go.mod, invoked via `go tool <name>`, and drop the separate hack/tools module and its install make targets. Prebuilt-binary tools (golangci-lint, helm, protoc, quill, tilt, k3d, jq) still install into hack/bin via hack/tools.mk.

I also made sure code-generator and controller-tools are hard-pinned via replace directives to avoid a known breaking change in gRPC/protobuf and CRD codegen.

Also switch the golangci-lint installer to https://golangci-lint.run/install.sh to follow best guidance from the project on how to install


fuskovic (approved):
LGTM

krancour:
This makes me really happy!

krancour (approved):

PR #6558, opened 2026-07-06
feat(expressions): add repoCredentials() expression function

### LINKED ISSUE #6549: expression function for repo credential lookup by url
There are already functions available within expressions that allow directly accessing a Secret ("generic credential") _by name_, but _only_ by name.

Many built-in promotion steps that deal in some way with repositories seamlessly resolve credentials through a lookup by repo URL. Importantly, these lookups don't exclusively use the credentials they find _directly_. In many cases, they seamlessly use details from the Secret they find to obtain narrowly-scoped, short-term credentials, which are what it actually returns and are subsequently used for repository access.

A great example of the above involves authentication to GitHub repositories via a GitHub App. The Secret that is actually stored and indexed by URL (or a URL regex) contains app ID, installation ID, and a private key. None of these are usable for directly accessing a repository or GitHub APIs. Instead, the key is used to sign a request and obtain a narrowly-scoped, short-lived _installation token_, which is the useful credential.

Other examples include so-called "ambient credentials," as in the case of any cloud provider's Pod identity scheme. Keep in mind, all of this is 100% transparent for built-in promotion steps like `git-clone` or `gti-open-pr`.

Currently, should these sort of credentials be required within step configuration (imagine an `http` step attempting to interact with a GitHub API), there is no way at all to obtain them.

This issue calls for making repository credential lookups by URL accessible from within an expression anywhere the existing `secret()` function is available. I would propose the name `repoCredentials()` for the function, and that it take URL and credential type (e.g. `git`, `image`, etc.) as arguments. This would be a thin entry-point into the same "credentials database" component that's utilized by the built-in steps.

jsbroks:
## Description

Closes #6549

Adds a `repoCredentials()` expression function that resolves repository credentials **by repo URL and credential type** (`git`, `helm`, `image`), returning them as a `map[string]string`. It is a thin entry point into the same credentials database used by built-in promotion steps like `git-clone`.

Unlike `secret()`, which returns the raw `Data` of a `Secret` selected by _name_, `repoCredentials()` returns the **resolved** credentials — not the raw contents of the underlying `Secret`. This means it transparently handles credential schemes that yield narrowly-scoped, short-lived credentials (e.g. a GitHub App installation token minted from an app ID + private key, or a cloud provider's ambient / Pod-identity credentials), which is exactly what custom and generic steps (e.g. `http` calling a provider API) need but previously had no way to obtain.

### Behavior

- Signature: `repoCredentials(repoURL, type)` where `type` is one of `git`,
  `helm`, or `image`. The type is required because credentials are indexed by
  type and the same URL may resolve differently per type (URL normalization also
  differs per type).
- When credentials are found, it always returns the **fixed** set of keys
  `username`, `password`, `sshPrivateKey` (empty string when unset), so
  expression authors can rely on a predictable shape regardless of provider.
  API keys / personal access tokens / minted tokens are surfaced via `password`.
- When no credentials are found, it returns an empty map (mirroring `secret()`).

### Implementation

- New `RepoCredentials()` option and `getRepoCredentials()` implementation in
  `pkg/expressions/function`, wired into `DataOperations()` (which now takes a
  `credentials.Database`).
- `StepEvaluator` now carries the credentials database and threads it through
  `Vars` / `ShouldSkip` / `Config`; `LocalOrchestrator` passes the DB it already
  receives.
- Caching mirrors `secret()` (per-step cache, distinct `RepoCredentials`
  key prefix).

### Reviewer notes

- **Contexts without a credentials database** — Stage verification argument
  evaluation, the Argo CD selector, and the indexer — register the function but
  return a clear "not available in this context" error if it is invoked. Fully
  wiring a credentials DB into the Stage controller was intentionally left out to
  limit blast radius; happy to follow up if we want it there too.
- **Custom / pod-based steps** are the primary intended consumer (per the
  discussion on #6549). This PR wires the function into the OSS `StepEvaluator`
  config-evaluation path; the Enterprise pod-based executor's config-evaluation
  path should be confirmed to flow through the same wiring.
- **Security consideration for discussion:** because the function materializes a
  (possibly minted) token into the expression environment, that token can flow
  into a step's `outputs`, logs, or committed files. Worth deciding whether any
  guardrails are wanted here.

### Eligibility

- [ ] Linked to an existing issue with no blocking labels.

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
+++ b/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
@@ -466,8 +466,8 @@ config:
 ### `repoCredentials(repoURL, type)`
 
 The `repoCredentials()` function resolves repository credentials by repository
-URL and credential type, returning them as a `map[string]string`. It takes two
-required arguments:
+URL and credential type, returning them as an object whose fields are accessed
+by name. It takes two required arguments:
 
 - `repoURL` (Required): A string representing the URL of the repository whose
   credentials should be resolved.
@@ -487,17 +487,18 @@ tokens or a cloud provider's ambient (Pod identity) credentials — instead of
 returning the raw material (e.g. an app ID and private key) from which those
 credentials are derived.
 
-Because the result is normalized, the set of available keys is **fixed** and
-does not vary by credential type or provider. When credentials are found, the
-returned map always contains all of the following keys:
+Because the result is normalized, the set of available fields is **fixed** and
+does not vary by credential type or provider. The returned object always exposes
+all of the following fields:
 
-| Key | Description |
-|-----|-------------|
-| `username` | The username identifying the principal. For token-based credentials this is often an inconsequential placeholder. |
-| `password` | The password or token used to authenticate. **API keys and personal access tokens are surfaced here.** |
-| `sshPrivateKey` | The SSH private key, when applicable. **Deprecated as of v1.10.0** and slated for removal in v1.13.0. |
+| Field | Description |
+|-------|-------------|
+| `Username` | The username identifying the principal. For token-based credentials this is often an inconsequential placeholder. |
+| `Password` | The password or token used to authenticate. **API keys and personal access tokens are surfaced here.** |
+| `SSHPrivateKey` | The SSH private key, when applicable. **Deprecated as of v1.10.0** and slated for removal in v1.13.0. |
 
-If no matching credentials are found, an empty map is returned.
+If no matching credentials are found, an object with all fields empty is
+returned.
 
 For details on how each field is populated for a given credential type or
 provider (e.g. GitHub App, ECR, or basic username/password), see the
@@ -510,13 +511,14 @@ Examples:
 config:
   headers:
   - name: Authorization
-    value: Bearer ${{ repoCredentials('https://github.com/example/repo.git', 'git').password }}
+    value: Bearer ${{ repoCredentials('https://github.com/example/repo.git', 'git').Password }}
 ```
 
 :::note
 
-`repoCredentials()` is only available within `Promotion` step expressions, where
-Kargo's credentials database is available.
+`repoCredentials()` is available wherever Kargo's credentials database is wired:
+within `Promotion` step expressions and within a `Stage`'s verification argument
+expressions.
 
 :::
 
--- a/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
+++ b/docs/docs/50-user-guide/60-reference-docs/40-expressions.md
@@ -463,6 +463,63 @@ config:
 
 :::
 
+### `repoCredentials(repoURL, type)`
+
+The `repoCredentials()` function resolves repository credentials by repository
+URL and credential type, returning them as a `map[string]string`. It takes two
+required arguments:
+
+- `repoURL` (Required): A string representing the URL of the repository whose
+  credentials should be resolved.
+
+- `type` (Required): A string representing the credential type. Must be one of
+  `git`, `helm`, or `image`. The type is required because credentials are
+  indexed by type, and the same URL may resolve to different credentials
+  depending on it.
+
+Unlike `secret()`, which returns the raw `Data` of a `Secret` selected by
+_name_, `repoCredentials()` performs a lookup by repository URL through the same
+credentials database used by built-in promotion steps like `git-clone`. It
+returns the **resolved** credentials — not the raw contents of the underlying
+`Secret`. This means it transparently handles credential schemes that yield
+narrowly-scoped, short-lived credentials — such as GitHub App installation
+tokens or a cloud provider's ambient (Pod identity) credentials — instead of
+returning the raw material (e.g. an app ID and private key) from which those
+credentials are derived.
+
+Because the result is normalized, the set of available keys is **fixed** and
+does not vary by credential type or provider. When credentials are found, the
+returned map always contains all of the following keys:
+
+| Key | Description |
+|-----|-------------|
+| `username` | The username identifying the principal. For token-based credentials this is often an inconsequential placeholder. |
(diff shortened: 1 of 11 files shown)


pkg/expressions/function/functions.go
```diff
@@ -1060,6 +1090,104 @@ func isGenericSecretType(secret corev1.Secret) bool {
...
+		// When credentials are found, always return the full, fixed set of keys
+		// so that expression authors can rely on a predictable shape regardless
+		// of credential type or provider. When no credentials are found, an empty
+		// map is returned, mirroring the behavior of secret().
+		data := make(map[string]string)
+		if creds != nil {
+			data[credentials.FieldUsername] = creds.Username
+			data[credentials.FieldPassword] = creds.Password
+			data[credFieldSSHPrivateKey] = creds.SSHPrivateKey
+		}
```
krancour: I think we could consider returning the actual creds struct.

pkg/controller/stages/regular_stages.go
```diff
@@ -1386,6 +1386,10 @@ func (r *RegularStageReconciler) startVerification(
 				exprfn.DataOperations(
 					ctx,
 					r.client,
+					// The Stage reconciler does not wire a credentials database,
+					// so repoCredentials() is unavailable during verification
+					// argument evaluation.
+					nil,
```
krancour: I'm iffy on this.

If we've already managed to justify directly accessing Secrets by name in this context, my assumption is someone will inevitably want to access repo credentials by URL in this context as well.

If there were a safety reason not to do it, it would have already stopped us from allowing Secret access directly by name.

I would probably consider plumbing the creds DB through to here.

Iffy as I am though, I'm happy to hear different opinions about this.

cc @thomastaylor312 also.

jsbroks: I think you are right

docs/docs/50-user-guide/60-reference-docs/40-expressions.md
```diff
@@ -463,6 +463,65 @@ config:
...
+all of the following fields:
+
+| Field | Description |
+|-------|-------------|
+| `Username` | The username identifying the principal. For token-based credentials this is often an inconsequential placeholder. |
+| `Password` | The password or token used to authenticate. **API keys and personal access tokens are surfaced here.** |
+| `SSHPrivateKey` | The SSH private key, when applicable. **Deprecated as of v1.10.0** and slated for removal in v1.13.0. |
+
+If no matching credentials are found, an object with all fields empty is
+returned.
```
krancour: This might not be the optimal behavior. It's common to use optional chaining and nil-coalescing in expressions, so returning `nil` is probably a more useful signal when no credentials are found.

pkg/expressions/function/functions.go
```diff
@@ -1060,6 +1090,101 @@ func isGenericSecretType(secret corev1.Secret) bool {
...
+			return nil, fmt.Errorf(
+				"error getting %s credentials for %q: %w", credType, repoURL, err,
+			)
+		}
+
+		// Return the resolved credentials as a value struct so that expression
+		// authors can rely on a predictable shape regardless of credential type
+		// or provider. When no credentials are found, a zero-value struct (all
+		// fields empty) is returned.
+		var result credentials.Credentials
```
krancour: See https://github.com/akuity/kargo/pull/6558/changes#r3661696941

pkg/expressions/function/functions.go
```diff
@@ -525,6 +533,28 @@ func Secret(ctx context.Context, c client.Client, cache *gocache.Cache, project
...
+func RepoCredentials(
+	ctx context.Context,
+	credsDB credentials.Database,
+	cache *gocache.Cache,
+	project string,
+) expr.Option {
+	return expr.Function(
+		"repoCredentials",
+		getRepoCredentials(ctx, credsDB, cache, project),
+		new(func(repoURL, credType string) map[string]string),
```
thomastaylor312: Shouldn't this type be the Credentials type? I can't remember if expr-lang converts a struct to a map, but I thought it could handle concrete types

pkg/expressions/function/functions.go
```diff
@@ -1134,6 +1164,105 @@ func isGenericSecretType(secret corev1.Secret) bool {
...
+			cacheKeyPrefixRepoCredentials,
+			project,
+			fmt.Sprintf("%s/%s", credType, repoURL),
+		)
+		if cache != nil {
+			if cachedData, ok := cache.Get(cacheKey); ok {
+				// A cached nil records a previous lookup that found nothing.
+				if cachedData == nil {
+					return nil, nil
+				}
```
thomastaylor312: Nit: This comment is a bit misleading because I don't see us setting a nil item down below when `creds` is nil. Maybe other places where we use the creds DB we might. Either way, that should be called out or this should say this is just a sanity check before type assertion

pkg/expressions/function/functions_test.go
```diff
@@ -1457,6 +1460,261 @@ func Test_getSecret(t *testing.T) {
...
+			},
+			cache: cache.NewFrom(cache.NoExpiration, cache.NoExpiration, map[string]cache.Item{
+				cacheKey: {Object: nil},
+			}),
+			args: []any{testRepoURL, "git"},
+			assertions: func(t *testing.T, _ *cache.Cache, result any, err error) {
+				assert.NoError(t, err)
+				assert.Nil(t, result)
+			},
+		},
```
thomastaylor312: Nit: Not sure if this test name is right? Is this one supposed to be testing that the value gets pulled from the cache or that a cache miss pulls from the DB? The comment in the `GetFn` indicates the former and the name indicates the latter. I also see those cases tested below, so maybe this is just checking the fallback when the cache returns nil?

thomastaylor312 (approved):
Good to go with a few small nits

PR #6334, opened 2026-05-22
feat: pause auto-promotion after promoting older freight

jacobboykin:
Relates to #3016

## Summary

This adds auto-promotion holds for Stages.

When a Promotion other than the Stage controller's own auto-promotion succeeds for Freight that is not the current auto-promotion candidate for the same origin, Kargo records a hold that pauses auto-promotion for that origin on that Stage. That keeps later auto-promotion from moving the Stage back to the candidate. Other origins on the same Stage can continue to auto-promote.

Auto-promotion resumes after a Promotion other than the Stage controller's own auto-promotion succeeds for the current candidate for that origin.

## How It Works

For created Promotions on auto-promotion-enabled Stages, the Promotion webhook infers promotion intent unless the Promotion is marked as one created by the Stage controller's auto-promotion loop. That keeps Kargo's own auto-promotions from creating holds while allowing UI, CLI, `kubectl`, external automation, and Enterprise auto-rollback Promotions to participate without special handling.

- If the selected Freight is not the current auto-promotion candidate, the webhook stamps a hold intent annotation.
- If the selected Freight is the current candidate, the webhook stamps a release intent annotation.
- Stage-controller auto-promotions are not stamped.

The Stage controller is the only writer of committed hold state. It records active holds in `status.autoPromotionHolds`, keyed by origin, after a hold-intent Promotion succeeds. A succeeded release-intent Promotion clears the hold.

Hold processing uses a timestamp/name watermark so effects are applied once and remain stable even after Promotion GC.

`autoPromoteFreight` skips an origin when:

- an active hold exists for that origin, or
- a non-terminal hold-intent Promotion is still pending or running.

That second check closes the window where a user-selected hold Promotion exists but has not yet reached Stage status.

## Promote By Origin

This also adds a small general primitive: promote by origin.

REST, CLI, and raw `Promotion` specs can now ask Kargo to promote the current candidate for an origin in the direct Stage promotion path:

```sh
kargo promote \
  --project my-project \
  --stage test \
  --origin Warehouse/my-warehouse
```

For raw Kubernetes usage, create a `Promotion` with `spec.origin` instead of `spec.freight`. The mutating webhook resolves the origin to the current candidate and fills `spec.freight` before the Promotion is persisted.

The REST promote request now accepts exactly one of `freight`, `freightAlias`, or `origin`. The Promotion CRD accepts exactly one of `spec.freight` or `spec.origin`.

[...]

PR #6559, opened 2026-07-07
fix(controller): sub-reconcilers no longer act on stale status

krancour:
Several reconcilers (regular Stages, Projects, ProjectConfigs, ClusterConfigs) run a sequence of sub-reconcilers, patching status after each one to show progress. A failed patch was logged and tolerated, but the in-memory object was only ever refreshed from the server's response to a *successful* patch. A failed intermediate patch therefore had two quiet consequences:

1. Every later sub-reconciler in the pass computed from the pre-patch status, silently dropping the failed sub-reconciler's work for the remainder of the pass -- including from the final end-of-pass patch, which is built from the last sub-reconciler's output. The dropped work was only recovered by full recomputation on a later pass.

2. Decisions with side effects could be made from the stale view. In the Stage reconciler, auto-promotion could re-promote Freight the Stage had just received (under the matchUpstream selection policy) because the Freight history recorded moments earlier was not visible to it. In the Project reconciler, stats collection gates on the Ready condition and would act on the previous pass's value.

These windows only open on transient API errors (these are unguarded merge patches, so conflicts are not a factor), which is why the hazard has gone unnoticed. But work in progress (auto-promotion holds, #6334) is about to make the Stage reconciler's instance of it consequential: acting on stale hold state would auto-promote over a rollback the user just performed.

The fix separates the two roles the reconciled object was playing:

- Sub-reconcilers now operate on a working deep copy whose status is unconditionally brought up to date after every sub-reconciler, whether or not persistence succeeded. No sub-reconciler ever sees a stale view of the current pass.

- The original object is only ever touched by PatchStatus: refreshed from the server's response on success, untouched on failure. It always mirrors persisted state, and because it is also the base PatchStatus diffs against, a failed patch loses nothing -- the next patch (including the final one) diffs against true server state and re-carries the unpersisted changes.

Patching after each sub-reconciler thereby returns to being purely a progress-visibility optimization whose failure costs nothing.

The ProjectConfig and ClusterConfig reconcilers each have only one sub-reconciler today, so they had no observable bug; they receive the same treatment so the pattern is safe by construction whenever a second sub-reconciler is added.

New tests pin both behavioral fixes; both fail against the previous code.

### DIFF
--- a/pkg/controller/management/clusterconfigs/cluster_configs.go
+++ b/pkg/controller/management/clusterconfigs/cluster_configs.go
@@ -146,7 +146,21 @@ func (r *reconciler) reconcile(
 	clusterCfg *kargoapi.ClusterConfig,
 ) (kargoapi.ClusterConfigStatus, error) {
 	logger := logging.LoggerFromContext(ctx)
-	status := *clusterCfg.Status.DeepCopy()
+
+	// working is the ClusterConfig the sub-reconcilers operate on. Its status is
+	// unconditionally brought up to date after each sub-reconciler, whether or
+	// not persisting that status succeeded, so no sub-reconciler ever computes
+	// from a stale view of this pass.
+	//
+	// clusterCfg itself is only ever touched by PatchStatus, which refreshes it
+	// from the server's response on success and leaves it alone on failure. It
+	// therefore always reflects persisted state. Because it is also the base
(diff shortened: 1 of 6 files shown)


fuskovic (approved):
LGTM

PR #6560, opened 2026-07-07
fix(controller): prevent AnnotateStageWithArgoCDContext mutating caller state

krancour:
Small follow-up to #6559

On a successful patch, the controller-runtime client hydrates the patched object from the server's response. Any helper that patches a caller-owned object therefore overwrites the caller's in-memory copy -- in its entirety, status included -- as a side effect.

AnnotateStageWithArgoCDContext had exactly this shape: it accepted a caller-owned *Stage and patched it directly. Its one substantive caller is the Stage reconciler, which invokes it mid-pass on the working copy whose status accumulates the results of each sub-reconciler. A successful annotation silently reverted that working status to persisted state. Nothing currently reads the working status between that call and the point where the reconcile loop reassigns it, so this has never produced wrong behavior -- but that safety is circumstantial, unenforced, and invisible at the call site. Code added to syncPromotions downstream of the annotation call would inherit the hazard without any signal that it exists.

The helper only ever needed the Stage's identity. It now takes a types.NamespacedName (matching its neighbors RefreshStage and ReverifyStageFreight) and patches an internally constructed stub, making the side effect impossible rather than merely unobserved. A doc comment records the reasoning so the signature does not regress.

### DIFF
--- a/pkg/api/stage.go
+++ b/pkg/api/stage.go
@@ -228,11 +228,16 @@ func RefreshStage(
 // the latest Promotion.
 //
 // If no ArgoCD apps are found, the annotation is removed.
+//
+// This deliberately takes the Stage's identity rather than a caller's working
+// Stage object. This is to avoid the patchAnnotation() call in this method
+// overwriting pending status changes to the working Stage object with live
+// status, which may be stale in comparison.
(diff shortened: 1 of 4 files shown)


fuskovic (approved):
LGTM

PR #6562, opened 2026-07-07
fix: approval no longer makes Freight available across origin boundaries

krancour:
Corrects existing bug found while reviewing #6334

cc @fuskovic 

IsFreightAvailable treated manual approval as an unconditional grant: it returned true for approved Freight before ever consulting the Stage's Freight requests. Freight from an origin a Stage does not request could therefore be promoted to it, so long as someone approved it -- most plausibly by accident, since such a promotion accomplishes nothing the Stage's promotion process can act on.

The codebase already disagreed with itself about this. The other implementation of availability, ListFreightAvailableToStage, only ever queries the Warehouses of requested origins, so cross-origin approvals never surfaced in the UI's listings; they were honored only when named directly in a promote request. Nothing on the approval side guarded the door either: the approve endpoints validated only that the Freight and Stage exist and that the caller may promote to the Stage.

Approval's purpose is to bypass verification -- the hotfix path -- not to bypass a Stage's declared sources. This closes the gap at both layers:

- IsFreightAvailable now applies the approval check only within the origin-matched Freight request, so approval bypasses verification and soak requirements but not origin membership. This is the enforcement point: it guards every promotion path, including Promotions created with kubectl and approvedFor entries planted by direct status patches.

- The approve endpoints now reject approving Freight for a Stage that does not request Freight from its origin, so the mistake is caught when it is made rather than surfacing later as a baffling "Freight not available" on promote.

Existing cross-origin approvals become inert rather than invalid: the approvedFor entry remains on the Freight, but no longer makes the Freight available to the non-requesting Stage.

### DIFF
--- a/api/v1alpha1/stage_types.go
+++ b/api/v1alpha1/stage_types.go
@@ -156,20 +156,33 @@ func (s *Stage) IsControlFlow() bool {
 	}
 }
 
+// RequestsFreightFromOrigin returns whether any of the Stage's Freight requests
+// name the specified origin.
+func (s *Stage) RequestsFreightFromOrigin(origin FreightOrigin) bool {
+	if s == nil {
+		return false
+	}
+	for _, req := range s.Spec.RequestedFreight {
+		if req.Origin.Equals(&origin) {
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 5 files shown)


fuskovic (approved):
LGTM

PR #6570, opened 2026-07-08
feat(api): resolve promotion Targets for a Stage

jsbroks:
## Summary

Adds `api.GetTargetsForStage`, which resolves the `Target`(s) a `Stage` promotes to, based on `stage.spec.targetSelector` (added in the base PR).

> **Stacked on #6566** (`feat/target-per-stage`), which is itself stacked on #6565. Merge the stack bottom-up: #6565 → #6566 → this. The base is `feat/target-per-stage`; this PR will retarget automatically as the stack merges.

## Behavior

- **`spec.targetSelector == nil`** → returns a single **ephemeral Target synthesized from the Stage** (its name/namespace, a `kargo.akuity.io/stage` label, and the mirrored shard label). Nothing is written to the cluster, so the common selector-less case needs no managed `Target` resource.
- **`spec.targetSelector != nil`** → returns every `Target` in the Stage's namespace whose labels satisfy the selector (`List` + `MatchingLabelsSelector`).

This gives the promotion engine one uniform `[]Target` resolution path while keeping the no-selector case free of any persisted object.

## Why `pkg/api`

The helper lives in `pkg/api` alongside the other client-aware `kargoapi` helpers (`GetStage`, `RefreshStage`, …), so both the controllers and the promotion engine can call it without an import cycle.

## Changes

- `pkg/api/target.go` — `GetTargetsForStage` + the in-memory default synthesis.
- `pkg/api/target_test.go` — table-driven tests (ephemeral default, shard mirroring, selector match scoped to namespace, no-match empty list, list-error propagation).
- `charts/kargo/templates/controller/cluster-roles.yaml` — **read-only** (`get`/`list`/`watch`) RBAC on `targets` for the selector list.

## Verification

- `go build ./...` (both modules) and `make lint-go` — clean (0 issues).
- `go test ./pkg/api/...` — passing (incl. the 5 `GetTargetsForStage` subtests).

## Follow-ups (not in this PR)

- The promotion engine does not yet **call** `GetTargetsForStage`. Wiring it into the promotion process (`pkg/promotion`) is the next step.

### DIFF
--- a/charts/kargo/templates/controller/cluster-roles.yaml
+++ b/charts/kargo/templates/controller/cluster-roles.yaml
@@ -76,6 +76,14 @@ rules:
 - apiGroups:
   - kargo.akuity.io
   resources:
+  - targets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - kargo.akuity.io
+  resources:
   - clusterpromotiontasks
   - promotiontasks
   - warehouses
--- a/pkg/api/target.go
+++ b/pkg/api/target.go
@@ -1,0 +1,71 @@
+package api
+
+import (
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+)
+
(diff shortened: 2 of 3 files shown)


jsbroks:
Reverting the in-memory resolution approach. Going back to persisting a Target per Stage (in #6566) so downstream logic uniformly handles real Target CRs without special-casing an ephemeral object.

PR #6583, opened 2026-07-09
fix(argocd): Makes the updater more tolerant to discovery failures

thomastaylor312:
In new clusters or on a fresh startup, there can be hiccups when the initial discovery step happens for the k8s client. These were rare, but we did observe it happening several times (especially in tests). This adds retries to the discovery if and only if it was a partial discovery and not when something genuinely doesn't exist

### AI Use Disclosure

This PR was written:

### DIFF
--- a/pkg/promotion/runner/builtin/argocd_updater.go
+++ b/pkg/promotion/runner/builtin/argocd_updater.go
@@ -11,7 +11,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
 	"github.com/akuity/kargo/pkg/api"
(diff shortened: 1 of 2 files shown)


pkg/promotion/runner/builtin/argocd_updater.go
```diff
@@ -869,6 +878,61 @@ func (a *argocdUpdater) getAuthorizedApplications(
...
+
+// withDiscoveryRetry runs fn, retrying it with backoff for as long as it fails
+// because the discovery-backed RESTMapper could not map the Argo CD Application
+// kind due to incomplete or temporarily unavailable API discovery. It does NOT
+// retry when the Application kind is genuinely not served by the cluster (e.g.
+// the Argo CD CRDs are not installed), so a real misconfiguration still fails
+// fast instead of blocking for the full backoff.
+func (a *argocdUpdater) withDiscoveryRetry(ctx context.Context, fn func() error) error {
+	logger := logging.LoggerFromContext(ctx)
+	return retry.OnError(discoveryRetryBackoff, isIncompleteDiscoveryError, func() error {
```
jsbroks, line 905: nit: doesnt look like `retry.OnError` takes into account ctx cancel, so if the promotion step's deadline expires or it's cancelled mid-backoff, you keep sleeping for up to the remainder of the ~15s. I dont know if that matters

jsbroks (approved):

pkg/promotion/runner/builtin/argocd_updater.go
thomastaylor312: This was reusing existing things, but let me check on it and plumb through the cancellation if I need to

thomastaylor312: So technically it was plumbed through because the context was captured in the closures passed to `withDiscoveryRetry`. However, I think it is clearer if the passed function takes a `context.Context` to be clearer, even if it is more verbose. So I've made that change and will push it shortly

PR #6606, opened 2026-07-13
fix: retry if NoKindMatchError is from Argo CD

### LINKED ISSUE #6603: argocd-update: transient NoKindMatchError from discovery is treated as permanent — extend #6583's retry to cover it
## Summary

#6583 made the `argocd-update` step retry discovery failures, but deliberately only `*apiutil.ErrResourceDiscoveryFailed`. A bare `*meta.NoKindMatchError` is treated as "the kind is genuinely not served" and fails the step immediately:

> When the kind is genuinely absent, that targeted discovery returns NotFound and the mapper instead returns a bare *meta.NoKindMatchError, which is not matched here, so the caller fails fast rather than looping until the backoff is exhausted.

We have production-like evidence that `NoKindMatchError` can be **transient** when the Argo CD API server is reached through a proxy/LB layer, and the current fail-fast turns a sub-minute hiccup into a failed promotion.

## Ask

Include `meta.NoKindMatchError` (for the configured Argo CD Application GVK) in `withDiscoveryRetry`'s retryable set, using the same bounded backoff (~15s over 5 attempts). Cost when the CRD is genuinely absent: the step fails ~15s slower. Benefit: the step survives transient discovery inconsistency behind proxies/HA API servers, which is the normal topology for remote/managed Argo CD.

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6603

### DIFF
--- a/pkg/promotion/runner/builtin/argocd_updater.go
+++ b/pkg/promotion/runner/builtin/argocd_updater.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/xeipuuv/gojsonschema"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


jessesuen (approved):

yhxlele:
Merge to unblock akp e2e CI.

PR #6608, opened 2026-07-14
feat(api): add Target CRD type

jsbroks:
## Description

Introduces the `Target` CRD — the leaf kind of the fleet-management model. A `Target` is a single deployment destination (e.g. a cluster) that Stages promote Freight to. It is pure data: it holds target-specific `spec.params` consumed by a governing Stage's promotion steps (via `${{ target.params.* }}` in a later PR) and records which Stages own it per Freight origin in `status.ownedBy`. It deliberately defines **no** promotion steps and **no** Freight sources, so the group/leaf role collision is unrepresentable and a Target can never promote itself.

**Scope: types only.** No controller, webhook, RBAC, or CLI surface — those follow in later PRs, along with `spec.constraints` and `spec.deployOverride`. This PR only establishes the API surface that `Stage.spec.targetSelector` + the stage-self Target lifecycle (next PR) build on.

Hand-authored:
- `api/v1alpha1/target_types.go` — `Target`, `TargetSpec` (`params`), `TargetStatus` (`conditions` + `ownedBy`), `TargetOwnership`, `TargetList`, and the `GetStatus`/`GetConditions`/`SetConditions` helpers. Reuses the existing `FreightOrigin` and `HealthState` types.
- `api/v1alpha1/groupversion_info.go` — scheme registration.
- `api/v1alpha1/target_types_test.go` — unit tests for the hand-written helpers.

Everything else in the diff is regenerated (`make codegen` output): deepcopy, protobuf, the CRD manifest, UI bindings/schema, and the API reference doc.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:


api/v1alpha1/target_types.go
```diff
@@ -0,0 +1,93 @@
...
+	// Conditions contains the last observations of the Target's current state.
+	//
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchMergeKey:"type" patchStrategy:"merge" protobuf:"bytes,1,rep,name=conditions"`
+	// OwnedBy describes the Stages that currently govern this Target, with one
+	// entry per (Stage, Freight origin) pair.
+	OwnedBy []TargetOwnership `json:"ownedBy,omitempty" protobuf:"bytes,2,rep,name=ownedBy"`
```
jessesuen, line 57: ~What do you envision the ownedBy to be used for?~ The way this is structured, I think, will be expensive to maintain an accurate backreference at scale. Especially because it is in a list format.

Don't get me wrong, I had thought of the need for a back reference from leaf targets as well, but I thought it would be best to implement it as labels. For example, a similar bookkeeping could be achieved with:

```yaml
kind: Target
metadata:
  labels:
    kargo.akuity.io/warehouse.warehouse-a: stage-a
    kargo.akuity.io/warehouse.warehouse-b: stage-b
```

But the above has the advantage of:
1. being able to efficiently **_query_** Targets governed by a warehouse/stage tuple using the kube API
2. being able to efficiently set/remove using json PATCHes vs. status updates

jessesuen: You can disregard my above comment. I didn't take a close enough look at the details of `TargetOwnership`. I see that there's a lot more user informational stuff that you are planning to store. But the labeling idea may still be desirable if we want the efficient querying capability.

api/v1alpha1/target_types.go
```diff
@@ -0,0 +1,93 @@
...
+	Stage string `json:"stage,omitempty" protobuf:"bytes,1,opt,name=stage"`
+	// Origin is the origin of the Freight that the governing Stage promotes to
+	// this Target.
+	Origin FreightOrigin `json:"origin,omitempty" protobuf:"bytes,2,opt,name=origin"`
+	// CurrentFreight is the name of the Freight from Origin most recently
+	// promoted to this Target by the governing Stage.
+	CurrentFreight string `json:"currentFreight,omitempty" protobuf:"bytes,3,opt,name=currentFreight"`
+	// Health is the Target's last observed health with respect to the Freight
+	// indicated by CurrentFreight.
+	Health HealthState `json:"health,omitempty" protobuf:"bytes,4,opt,name=health"`
```
jessesuen: I didn't expect a Health status to be per origin/stage tuple. Some questions:

1. Why would this not be per origin (as opposed to per origin/stage combo). e.g. how could it be possible for origin-a/stage-a to be Healthy but origin-a/stage-b to be Degraded?
2. How would we even assess health per origin/stage tuple?

api/v1alpha1/target_types.go
jsbroks: I don't have a strong preference on the labels approach, happy to add them if the query efficiency is worth the extra bookkeeping. One thing I do think matters regardless is knowing which Freight (and version) is currently on a given Target

api/v1alpha1/target_types.go
jsbroks: Good questions, and thinking about it more, I don't think I have a solid answer to either one, so let's just drop Health from TargetOwnership for now. Additive changes are always easier than reworking something we've already shipped

thomastaylor312 (approved):
This looks like a clean, simple start for stuff

PR #6612, opened 2026-07-14
fix: emission of phantom objects for optional struct fields

### LINKED ISSUE #6590: generated go client emits phantom empty objects for optional struct fields
Optional struct fields in the go-swagger client (`pkg/client/generated/models`) are generated as non-pointer values with `,omitempty` — which does nothing for structs. So they always marshal as empty objects that aren't on the real resource.

Shows up as junk empty objects in `kargo get … -o yaml|json`, and anything that round-trips that output (re-applying an edited manifest, diffing, etc.) can trip over fields the resource never had.

~261 fields across the generated client. Root cause is `swagger.json` marking these optional but not nullable; go-swagger only emits a pointer when a field is `x-nullable`.

Fix: mark optional object-typed properties `x-nullable` in `hack/codegen/fix-swagger-spec.sh`, then regenerate — corrects all of them at once. Requests are unaffected (client sends scalars / raw manifests), so this is display/round-trip only. Land before the REST CLI is the only CLI.

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6590


Makefile
```diff
@@ -272,13 +272,15 @@ codegen-openapi: install-jq
...
+	hack/codegen/flatten-nullable-refs.sh swagger.json /tmp/swagger-go-client.json
 	mkdir -p pkg/client/generated
 	go tool swagger generate client \
-		-f swagger.json \
+		-f /tmp/swagger-go-client.json \
 		-t pkg \
 		--client-package client/generated \
 		--model-package client/generated/models \
 		--skip-validation
+	rm -f /tmp/swagger-go-client.json
```
krancour: Would it be better to integrate this into the existing `fix-swagger-spec.sh` script instead of adding net new machinery for mutating the same file that existing script does?

krancour:
Thanks for the deep investigation here — your findings about `allOf` and `x-nullable` were spot-on and are what prompted me to look at this from a different angle.

After some experimentation, I think there's a cleaner exit than maintaining a separate, Go-client-only derivative of `swagger.json` (or monkeying with the spec any further than we already do): retiring go-swagger and generating the Go client with openapi-generator instead. It consumes our existing spec as-is, emits pointers for optional fields as its default behavior — fixing the phantom objects by construction — and it turns out to also fix a few pre-existing serialization defects the current client has (e.g., `Quantity` values are currently *silently corrupted* on round-trip). I've spiked this locally: the generated client builds, and round-trip fidelity tests prove its output is byte-identical to what the server sends, verified against richly-populated Stage/Freight/Promotion/Warehouse fixtures. A ported slice of the CLI confirms the call-site migration is small and mechanical -- perfect busy work for Claude.

Given that, I'm inclined to hold off on merging this and put the effort toward the generator swap instead. Let's sync offline on next steps — and thanks again for the legwork on this one; it's what made the real root cause clear.

Issue #1099, opened 2023-11-13
Support OCI manifests


joebowbeer:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate feature request.

# Proposed Feature

Kargo should be able to use OCI registries for k8s manifests.

This includes the ability for a stage to subscribe to manifests from an OCI registry as well as to publish rendered manifests to an OCI registry.

# Motivation

In the rendered manifest pattern, the rendered manifests are artifacts and, as such, should be published to an OCI registry (e.g., ghcr).

In addition, OCI registries support signed manifests.

# Suggested Implementation

ArgoCD is adding first-class support for OCI

https://github.com/argoproj/argo-cd/blob/master/docs/proposals/native-oci-support.md

Flux already supports OCI. The flux cli can push/pull manifests to/from OCI registries.

https://fluxcd.io/flux/cheatsheets/oci-artifacts/

krancour:
Thanks for an interesting idea @joebowbeer. I think there's a lot to work through here, but I really like the idea.

krancour:
Not stale.

hoopty:
The Argo CD link is https://github.com/argoproj/argo-cd/blob/master/docs/proposals/native-oci-support.md

blakepettersson:
Something being worked on atm with argoproj/argo-cd#18646

hoopty:
Is this still in progress?  We have some air gap use cases that would be nice to not have local git server dependency.

joebowbeer:
Update: The ArgoCD support is now beta

https://github.com/argoproj/argo-cd/pull/18646

blakepettersson:
not stale

shabbir-saifee-cmm:
Is this something currently worked on? Can we expect this one in 1.8 release?

shabbir-saifee-cmm:
Any updates?

blakepettersson:
While there's now an `oci-push` step, it would still be nice to have an `oci-build` step

vcalmic:
Is this initiative actively being pursued at this time?

blakepettersson:
@vcalmic no

vcalmic:
> [@vcalmic](https://github.com/vcalmic) no

I'd like to pick this up and implement the `oci-build` step. We're currently transitioning to using OCI artifacts for our manifest distribution, so this is a piece we're missing as well.

[...]

PR #6626, opened 2026-07-15
feat(oci-push): support pushing a local archive

EronWright:
## Summary

Enhances the `oci-push` promotion step to push a **local archive** from the workspace (e.g. a tarball of rendered manifests) to a registry as a single-layer OCI artifact, alongside the existing registry-to-registry copy/retag mode. This is the `oci-push` half of the work discussed in #1099 (the complementary `tar` step is being contributed separately by @vcalmic).

Media types are settable, per @blakepettersson's request. Because the pinned go-containerregistry has no top-level `manifest.artifactType`, `artifactType` is carried via the manifest's `config.mediaType` — the OCI-1.0-compatible convention used by Helm/Flux/ORAS.

Refs #1099.

### Push a local archive (Flux/Argo-style OCI manifests)

Package a directory of rendered manifests and publish it as an OCI artifact for consumption by Argo CD, Flux, or other OCI-aware tooling. Provenance is recorded via standard OCI annotations.

```yaml
steps:
- uses: tar                       # complementary step (forthcoming, #1099)
  config:
    path: ./manifests
    outPath: ./manifests.tar.gz
- uses: oci-push
  config:
    srcPath: ./manifests.tar.gz
    destRef: oci://ghcr.io/example/config/app:${{ ctx.promotion }}
    mediaType: application/vnd.cncf.flux.content.v1.tar+gzip   # layer media type
    artifactType: application/vnd.cncf.flux.config.v1+json     # → config.mediaType
    annotations:
      org.opencontainers.image.source: ${{ commitFrom("https://github.com/example/app.git").repoURL }}
      org.opencontainers.image.revision: ${{ commitFrom("https://github.com/example/app.git").id }}
```

### Minimal (default media type)

`srcPath` is content-agnostic; when `mediaType` is omitted it defaults to `application/vnd.oci.image.layer.v1.tar+gzip`.

```yaml
steps:
- uses: oci-push
  config:
    srcPath: ./artifact.tar.gz
    destRef: registry.example.com/example/app:${{ ctx.promotion }}
```

### Existing copy/retag mode (unchanged)

```yaml
steps:
- uses: oci-push
  config:
    srcRef: registry.example.com/myapp@${{ imageFrom("registry.example.com/myapp").digest }}
    destRef: registry.example.com/myapp:v1.2.3
```

## Config

Exactly one of `srcRef` or `srcPath` is required (enforced via schema `oneOf`).

| Field | Description |
|-------|-------------|
| `srcPath` | Workspace-relative local file to push as a single-layer artifact. Mutually exclusive with `srcRef`. |
| `mediaType` | Layer media type. Default `application/vnd.oci.image.layer.v1.tar+gzip`. `srcPath` only. |
| `artifactType` | Applied as the manifest config media type. `srcPath` only. |

## Notes / scope

- `srcPath` is content-agnostic — not limited to tarballs (e.g. WASM modules, SBOMs work too).
- The artifact size limit (`MAX_OCI_PUSH_ARTIFACT_SIZE`) is always enforced for local-archive pushes, based on the file's size.
- Producing canonical **Helm charts** via `srcPath` is out of scope (their config blob must carry `Chart.yaml` metadata); use the copy/retag mode with the `oci://` prefix to promote existing charts.
- A literal top-level OCI 1.1 `artifactType` and settable config content are natural follow-ups (would need a ggcr bump or a custom manifest wrapper).

## Testing

- Regenerated config types via `make codegen-schema-to-go`.
- New unit tests: schema `oneOf` validation, local-archive happy path (read-back asserts OCI manifest media type, layer media type, `artifactType`-as-config-media-type, scoped annotations, byte round-trip, digest), default media type, and negatives (directory, missing file, contained path traversal, size limit).
- `make lint-go` — clean.

### DIFF
--- a/pkg/promotion/runner/builtin/oci_pusher.go
+++ b/pkg/promotion/runner/builtin/oci_pusher.go
@@ -3,11 +3,17 @@ package builtin
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 
+	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	ggcrmutate "github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/static"
+	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/xeipuuv/gojsonschema"
 
@@ -48,8 +54,9 @@ func init() {
 }
 
 // ociPusher is an implementation of the promotion.StepRunner interface that
-// copies/retags OCI artifacts (container images and Helm charts) between
-// registries.
+// pushes OCI artifacts to a registry. It can copy/retag artifacts (container
+// images and Helm charts) between registries, or push a local file from the
+// workspace as a single-layer OCI artifact.
 type ociPusher struct {
 	schemaLoader    gojsonschema.JSONLoader
 	credsDB         credentials.Database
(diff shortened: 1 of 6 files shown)


pkg/promotion/runner/builtin/oci_pusher.go
```diff
@@ -153,6 +147,139 @@ func (p *ociPusher) run(
...
+			fmt.Errorf("failed to build artifact: %w", err)
+	}
+	// empty.Image is a Docker manifest by default; emit an OCI manifest instead.
+	img = ggcrmutate.MediaType(img, types.OCIManifestSchema1)
+	if cfg.ArtifactType != "" {
+		// This ggcr version has no top-level manifest artifactType, so carry the
+		// artifact type via the config media type per the OCI/Helm/Flux/ORAS
+		// convention.
+		img = ggcrmutate.ConfigMediaType(img, types.MediaType(cfg.ArtifactType))
+	}
```
fuskovic, line 258: If we make `ConfigMediaType` default to `types.OCIConfigJSON` a `srcPath` push with no `artifactType` yields a fully OCI-consistent manifest, and still honors `cfg.ArtifactType` when the caller sets it. Then we can add an assertion for config media type defaults to the OCI config type in the tests.

```suggestion
// empty.Image is a Docker manifest (and Docker-config-typed) by default;
// emit an OCI manifest instead.
img = ggcrmutate.MediaType(img, types.OCIManifestSchema1)

// This ggcr version has no top-level manifest artifactType, so carry the
// artifact type via the config media type per the OCI/Helm/Flux/ORAS
// convention. When no artifactType is configured, still override the
// config media type otherwise it stays types.DockerConfigJSON, 
// pairing an OCI manifest with a Docker-typed
// config, which strict OCI/ORAS consumers may reject.
configMediaType := types.OCIConfigJSON
if cfg.ArtifactType != "" {
	configMediaType = types.MediaType(cfg.ArtifactType)
}
img = ggcrmutate.ConfigMediaType(img, configMediaType)
```

fuskovic (changes requested):

PR #6630, opened 2026-07-16
fix(controller): validate new GitHub App tokens before first use

thomastaylor312:
GitHub replicates newly minted installation access tokens to its edge infrastructure asynchronously, so a token used immediately after being minted is sometimes rejected. Because GitHub masks authorization failures on private repositories as 404s, this has long manifested as intermittent "repository not found" errors during git operations that immediately follow a token cache miss (see #5610). The problem became dramatically more frequent in v1.8.0, when tokens began being scoped to individual repositories, multiplying how often new tokens are minted.

Before caching a newly minted token and releasing it to the caller, the GitHub App credential provider now confirms the token actually works by requesting metadata about the repository the token is scoped to, retrying on the backoff schedule GitHub themselves recommend for exactly this situation (3s/10s/30s): https://github.com/aws-amplify/amplify-hosting/issues/4080 contains the issue and the response from GitHub support. In the typical case the first check succeeds and the added cost is a single lightweight API request per mint. A token that cannot be validated after exhausting the schedule is presumed usable, leaving the caller no worse off than before this change.

### DIFF
--- a/pkg/credentials/github/app.go
+++ b/pkg/credentials/github/app.go
@@ -40,11 +40,6 @@ const (
 	tokenCacheExpiryMargin = 5 * time.Minute
 
 	tokenValidationRequestTimeout = 10 * time.Second
-
-	// retryKeySuffix is appended to a token's singleflight key when retrying a
-	// mint whose winner's context was canceled, so that all callers recovering
-	// from that shared failure coalesce again instead of racing.
-	retryKeySuffix = ":retry"
 )
 
 var base64Regex = regexp.MustCompile(`^[a-zA-Z0-9+/]*={0,2}$`)
@@ -71,6 +66,12 @@ type AppCredentialProvider struct {
 	// minted installation access token. This is set as a field so that it can be overridden in
 	// tests to avoid long waits.
 	validationBackoff []time.Duration
+
+	// mintTimeoutBuffer is added to the sum of validationBackoff when computing the maximum time to
+	// wait for a token to be minted and validated. It accounts for network round-trips that are not
+	// captured by the backoff schedule. This is set as a field so that it can be overridden in tests
+	// to avoid long waits.
+	mintTimeoutBuffer time.Duration
 
 	getAccessTokenFn func(
 		appOrClientID string,
@@ -104,6 +105,10 @@ func NewAppCredentialProvider() credentials.Provider {
 			10 * time.Second,
 			30 * time.Second,
 		},
+		// This is set to 10 seconds to account for the sum of the durations + accounting for any
+		// slow network round-trips. This is moderately generous because it is used only as a
+		// fail-safe
+		mintTimeoutBuffer: 10 * time.Second,
 	}
 	p.getAccessTokenFn = p.getAccessToken
 	p.validateAccessTokenFn = p.validateAccessToken
@@ -209,7 +214,24 @@ func (p *AppCredentialProvider) getUsernameAndPassword(
 	// they miss the cache in unison when that entry expires and, without coordination, would each
 	// mint their own token. Coalescing concurrent mints for the same cache key avoids that
 	// thundering herd.
-	accessToken, err, _ := p.mintGroup.Do(cacheKey, func() (any, error) {
+	var accessToken any
+	var err error
+
+	// maxWait bounds how long the mint may run. Because the mint executes under an orphaned context
+	// (see mintAndCacheToken) rather than any caller's context, this is the only thing bounding its
+	// duration. It is the sum of the validation backoff schedule plus a buffer for network
+	// round-trips.
+	maxWait := p.mintTimeoutBuffer
+	for _, d := range p.validationBackoff {
+		maxWait += d
+	}
+
+	// NOTE(thomastaylor312): Right now if there is a single error that isn't a context error, it
+	// will fail all of the other requests as well. This _could_ end up biting us if we have a
+	// transient error. If that ever becomes the case, we can add an OR to this that uses the
+	// `shared` return value from the singleflight `DoChan` channel and then re-enqueue here as
+	// well, but for now we'll let real errors fall through to everyone.
+	ch := p.mintGroup.DoChan(cacheKey, func() (any, error) {
 		return p.mintAndCacheToken(
 			ctx,
 			appOrClientID,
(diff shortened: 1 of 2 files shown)


thomastaylor312:
As a note, if we don't like the `singleflight` thing, I can remove it. It does have trade offs, but is helpful for large monorepo-type situation to avoid thundering herds

thomastaylor312:
I split singleflight into its own commit so it can be popped off if we don't like it

pkg/credentials/github/app.go
```diff
@@ -170,18 +205,105 @@ func (p *AppCredentialProvider) getUsernameAndPassword(
...
+	// flight another waiter had already started.
+	//
+	// Right now if there is a single error that isn't a context error, it will fail all of the
+	// other requests as well. This _could_ end up biting us if we have a transient error. If that
+	// ever becomes the case, we can add an OR to this that uses the `shared` return value from the
+	// singleflight `Do` call and then re-enqueue here as well, but for now we'll let real errors
+	// fall through to everyone.
+	if err != nil &&
+		(errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)) &&
+		ctx.Err() == nil {
```
krancour: I've got to be interpreting something wrong here. It isn't immediately obvious to me how the error returned from the function could be the result of a canceled or timed out context with `ctx.Err()` _also_ returning `nil`.

pkg/credentials/github/app.go
```diff
@@ -170,18 +205,105 @@ func (p *AppCredentialProvider) getUsernameAndPassword(
...
+				ctx,
+				appOrClientID,
+				installationID,
+				encodedPrivateKey,
+				repoURL,
+				cacheKey,
+			)
+		})
+	}
+	// At this point if we're checking an error, we've now already retried
```
krancour: I don't see a bug here, but I wonder if this comment isn't accurate. Wouldn't you also get to here via an error that _wasn't_ the result of a canceled or timed out context, in which case there has been no retry (nor should there have been)?

pkg/credentials/github/app.go
```diff
@@ -170,18 +205,105 @@ func (p *AppCredentialProvider) getUsernameAndPassword(
...
+	cacheKey string,
+) (string, error) {
+	logger := logging.LoggerFromContext(ctx)
+
+	// A concurrent call may have minted and cached a token between this
+	// call's cache miss and its turn in the singleflight group.
+	if entry, exists := p.tokenCache.Get(cacheKey); exists {
+		logger.Debug("installation access token cached by concurrent call")
+		return entry.(string), nil // nolint: forcetypeassert
+	}
```
krancour: I'm not quite understanding this, and it's possibly me. My understanding of singleflight wasn't that it serializes function calls within the group, but that the function is called once and all callers get the same result. So shouldn't multiple invocations of this function never actually be run concurrently? If it's _possible_ I would think it might be on account of the key suffix used by retries? Which might cause me to question whether using a separate key for retries is really the right thing to do.

To be clear, this is a question more than anything.

pkg/credentials/github/app.go
krancour: Nevermind -- I read the previous conditional backwards.

pkg/credentials/github/app.go
krancour: Sorry. I read this backwards, but am still confused -- just in a different way now.

You're looking for errors that _aren't_ context errors, according to the comment and according to the `ctx.Err() == nil` criterion, but isn't `(errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)` looking for errors that _are_ context errors. I'm reading the entirety of the conditional as "error isn't nil and both is and isn't a context error."

krancour: Ooooh...

This is checking that _this_ call's context wasn't canceled or timed out. The cancelation or timeout had to do with another caller's context -- the goroutine that was actually running the work.

I get it.

pkg/credentials/github/app.go
krancour: Retracting the retraction -- you _can_ reach this without having retried.

pkg/credentials/github/app.go
krancour: Claude's backing me up on the notion that the suffix is unnecessary and is the exact thing creating the potential for concurrency that doesn't otherwise happen.

pkg/credentials/github/app.go
krancour: Possibly requires further clarification of this in the comment. Not a hill I would die on.

pkg/credentials/github/app.go
```diff
@@ -170,18 +205,105 @@ func (p *AppCredentialProvider) getUsernameAndPassword(
...
+	accessToken, err, _ := p.mintGroup.Do(cacheKey, func() (any, error) {
+		return p.mintAndCacheToken(
+			ctx,
+			appOrClientID,
+			installationID,
+			encodedPrivateKey,
+			repoURL,
+			cacheKey,
+		)
+	})
```
krancour: There's a small gotcha here in that for all but one caller that's blocked, all callers aren't honoring cancelation or timeout of _their own_ context. Mine can be canceled and I can continue waiting on the call that's using yours, which is not.

`DoChan` is apparently the usual way of resolving that tension:

```
ch := p.mintGroup.DoChan(cacheKey, func() (any, error) {
    return p.mintAndCacheToken(ctx, appOrClientID, installationID, encodedPrivateKey, repoURL, cacheKey)
})
var accessToken any
var err error
select {
case res := <-ch:
    accessToken, err = res.Val, res.Err
case <-ctx.Done():
    return nil, ctx.Err()
}

```

pkg/credentials/github/app.go
krancour: Oooh... this has another benefit... With one key/no suffix, the retry can go into a loop that breaks only on success, timeout of your own context, or non-context error. Then we don't have the limitation of supporting only one retry attempt.

pkg/credentials/github/app.go
thomastaylor312: I can add clarification to the comment for sure. I tried to be my normal verbose self, but when you're staring at it for too long you can miss/elide details unintentionally

pkg/credentials/github/app.go
thomastaylor312: Yeah I'll make the comment clearer. The point is that this handles _real_ errors (or at least errors after a single retry)

pkg/credentials/github/app.go
thomastaylor312: So I think it is a bit more nuanced here. The concurrency comes from multiple promotions using this same credentials provider. Singleflight is essentially sync point for those token minting calls. The reason I use a suffix is explicitly to avoid a race condition _in the case of a failure_. Only one function runs, but multiple concurrent things are _waiting_ on that one function run. So I was worried about one of the other concurrent runs accidentally starting a new call on the same key. 

However, I just realized that the single flight will not return until the cacheKey is unlocked, so my retry is useless and the loop will work. I'll refactor this to use a loop here instead

pkg/credentials/github/app.go
thomastaylor312: Oh good call, I thought I about the one variant (the _other_ context) being cancelled, but not my own

pkg/credentials/github/app.go
```diff
@@ -71,8 +97,21 @@ func NewAppCredentialProvider() credentials.Provider {
...
 		),
+		// GitHub's own recommendation for retrying use of a newly minted token that has not yet
+		// replicated to all of their edge caches. See the Github support response in
+		// https://github.com/aws-amplify/amplify-hosting/issues/4080
+		validationBackoff: []time.Duration{
+			3 * time.Second,
+			10 * time.Second,
+			30 * time.Second,
+		},
+		// This is set to 10 seconds to account for the sum of the durations + accounting for any
```
krancour, line 108: Should the comment say 2 or should the code say 10?

thomastaylor312: Should say 10. I'll address

krancour (approved):
This is positively beautiful.

PR #6637, opened 2026-07-17
fix(kubeclient): move defer after error check to prevent nil pointer panic

magic-peach:
## Summary

In `GetCredential` (`pkg/kubeclient/auth.go`), the `defer` statement capturing `res.Body.Close()` was registered **before** the error check on `rc.Do()`:

```go
res, err := rc.Do(req)
defer func() {
    _ = res.Body.Close()
}()
if err != nil {
    return "", err
}
```

When `http.Client.Do()` returns an error, the `res` value is `nil` per the Go standard library contract. Because the deferred function captures `res` by reference through the closure, when the deferred close runs at function exit, it dereferences `res.Body` on a nil `res`, causing a **panic**.

This function is called to retrieve Kubernetes credentials, so any transient network error or misconfigured REST config will crash the server.

**Fix**: Moved the `defer` to after the `if err != nil` check so `res.Body.Close()` only runs when `res` is non-nil.

### DIFF
--- a/pkg/kubeclient/auth.go
+++ b/pkg/kubeclient/auth.go
@@ -39,12 +39,12 @@ func GetCredential(ctx context.Context, cfg *rest.Config) (string, error) {
 	// #nosec G704 -- This code executes client-side and does not, therefore,
 	// represent a vector for SSRF attacks.
 	res, err := rc.Do(req)
-	defer func() {
-		_ = res.Body.Close()
-	}()
 	if err != nil {
 		return "", err
 	}
+	defer func() {
+		_ = res.Body.Close()
+	}()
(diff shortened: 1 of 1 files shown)


krancour (approved):
Nice catch. Thank you for the fix.

krancour:
Only one minor correction to the original report for posterity. This actually cannot crash the server because this bit of code is used only by the CLI. In the context of the CLI, this is a real bug and the fix is correct.

krancour:
Made a call to backport to v1.11.0 because this is a small, no-risk change and there's really no reason not to have the upcoming release go out the door with one less bug.

PR #6647, opened 2026-07-17
fix!: replace go-swagger client gen w/open api client gen

### LINKED ISSUE #6590: generated go client emits phantom empty objects for optional struct fields

fuskovic:
## Summary

Closes #6590

- **Spec fixes** (`.swaggo` + `hack/codegen/fix-swagger-spec.sh`): map `Quantity`/`MicroTime` to their real string wire format and `IntOrString` to a typeless schema, so every consumer of `swagger.json` (Go, TS, third parties) gets the corrected shape.
- **New Go client** (`pkg/x/client/generated`, `hack/codegen/generate-go-client.sh`): generated via the standalone `openapi-generator-cli` JAR (`hack/tools.mk`) — no Docker-in-Docker needed. Lives under `pkg/x/` (no stability promise), with a wire-fidelity test suite (`pkg/x/client/fidelity`) round-tripping real `kargoapi`/`k8s.io` types.
- **CLI migration**: all ~50 command files ported to the new client, one API-tag group per commit, each independently reviewed.
- **Old client removed**: `pkg/client/generated` (go-swagger) and its codegen are deleted entirely; no coexistence period.


.gitignore
```diff
@@ -12,3 +12,5 @@ vendor/
 internal/server/ui/
 .DS_Store
 docs/static/swagger.json
+/swagger-v2-client-input.json
+docs/superpowers
```
krancour: Both of these look like they don't belong.

pkg/x/client/generated/client.go
```diff
@@ -0,0 +1,680 @@
...
+}
+
+// callAPI do the request.
+func (c *APIClient) callAPI(request *http.Request) (*http.Response, error) {
+	if c.cfg.Debug {
+		dump, err := httputil.DumpRequestOut(request, true)
+		if err != nil {
+			return nil, err
+		}
+		log.Printf("\n%s\n", string(dump))
```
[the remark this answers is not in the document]
fuskovic: done

krancour:
It appears the generated TS client is stale wrt to `swagger.json`.

hack/codegen/generate-go-client.sh
```diff
@@ -0,0 +1,168 @@
...
+  { print }
+' "${OUT_DIR}/client.go" > "${OUT_DIR}/client.go.tmp"
+mv "${OUT_DIR}/client.go.tmp" "${OUT_DIR}/client.go"
+
+sed -i.bak \
+  -e 's/httputil\.DumpRequestOut(request, true)/httputil.DumpRequestOut(request, false)/' \
+  -e 's/httputil\.DumpResponse(resp, true)/httputil.DumpResponse(resp, false)/' \
+  -e 's/log\.Printf("\\n%s\\n", string(dump))/log.Printf("\\n%s\\n", sanitizeHTTPDump(string(dump)))/g' \
+  "${OUT_DIR}/client.go"
+rm -f "${OUT_DIR}/client.go.bak"
```
krancour: When this was only taking care of the `interface{}{}` quirk which was relatively tiny, I was fine with it. Expanded into dealing with the (legitimate) issue flagged by CodeQL, this is starting to look complicated and brittle. Dealing with the second part in a cleaner way alters my thinking on the first part also.

It would be less brittle to provide the code generation tool with custom templates that will resolve both problems. You can get the originals from the tool's own source code and tweak them a bit. (I actually have the templates in-hand if you want them.)

hack/codegen/generate-go-client.sh
```diff
@@ -0,0 +1,97 @@
...
+  --global-property apiTests=false,modelTests=false,apiDocs=false,modelDocs=false \
+  --skip-validate-spec
+
+# --- Step 3: write our own go.mod (withGoMod=false above) and tidy -----------
+cat > "${OUT_DIR}/go.mod" <<'EOF'
+module github.com/akuity/kargo/pkg/x/client/generated
+
+go 1.26.0
+EOF
+(cd "${OUT_DIR}" && go mod tidy && go build ./...)
```
krancour: Three concerns here:

1. Dependabot's not going to know to reach into this script to update line 93.

2. Dependabot _will_ try to update dependencies in the generated `go.mod` and I don't want to get in its way.

3. Running `go mod tidy` against an _empty_ `go.mod` means without any other change to the repo, the result of running codegen can change. This will lead to frequent failures in CI's `check-codegen` job.

The thing that would address all three would be if we copy the existing `go.mod` and `go.sum` from the destination directory to a temporary location before running the generator and then restoring those files from the copy after.

krancour: You can keep the `go mod tidy` call, though.

That will handle the possibility of the generated client genuinely introducing a new dependency or dropping an existing one, but it won't introduce volatility in any other case.

krancour: The `go build` command probably slows things down unnecessarily though. If it doesn't build, there are other steps in CI that will catch it.

.github/workflows/ci.yaml
```diff
@@ -222,7 +222,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
     - name: Install tools
-      run: apt update && apt install unzip
+      run: apt update && apt install -y unzip default-jre-headless
```
krancour: There's an official `setup-java` action that we should probably use.

pkg/cli/client/client.go
```diff
@@ -7,51 +7,27 @@ import (
...
-	"github.com/go-openapi/strfmt"
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/spf13/pflag"
 
 	"github.com/akuity/kargo/pkg/cli/config"
 	"github.com/akuity/kargo/pkg/cli/option"
-	client "github.com/akuity/kargo/pkg/client/generated"
 	"github.com/akuity/kargo/pkg/client/watch"
 	"github.com/akuity/kargo/pkg/server"
+	kargogen "github.com/akuity/kargo/pkg/x/client/generated"
```
krancour, line 19: `kargogen` as an alias is distractingly awkward.

I am aware that `client` and `apiClient` are two aliases that would have made sense, but those symbols are already spoken for in most of the spots the `github.com/akuity/kargo/pkg/x/client/generated` is imported.

imho, `rest` would work nicely, but it's not a hill I would die on.

fuskovic: I personally like `kargogen` more then `rest` but can change if you insist. 

How do you feel about `clientgen`?

krancour: `clientgen` sounds more like it generates a client than that it _is_ a generated client. That was part of my issue with `kargogen`, which _also_ has the awkward quality of not hinting that it has anything to do with a client or an API. So I would say `clientgen` is an improvement in that regard, at least.

My very first thought had been `genedClient`. That's awkward because, phonetically, it reads like "jeaned." I started to ask myself whether the fact of it being generated actually was even relevant. What we care about is that it's a _client_. But then we've come full circle. That symbol's already spoken for by something worthy of that package name.

I found `rest` less objectionable, but it isn't perfect either.

Naming things is genuinely hard sometimes.

Since there isn't anything here that is _obviously_ an improvement, I'm truly content to leave it as it is.

Dockerfile.dev
```diff
@@ -9,7 +9,7 @@ RUN apt update && apt install -y ca-certificates curl gnupg unzip \
     && mkdir -p /etc/apt/keyrings \
     && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
     && echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list \
-    && apt-get update && apt-get install nodejs -y \
+    && apt-get update && apt-get install nodejs default-jre-headless -y \
```
krancour: We should probably pin to the same major version of Java/OpenJDK as it used in `ci.yaml`:

```suggestion
ARG JAVA_MAJOR=21
ARG NODE_MAJOR=22
ARG PNPM_VERSION=11.13.0

RUN apt update && apt install -y ca-certificates curl gnupg unzip \
    && mkdir -p /etc/apt/keyrings \
    && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
    && echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list \
    && apt-get update && apt-get install nodejs openjdk-${JAVA_MAJOR}-jre-headless -y \
```

go.mod
```diff
@@ -44,7 +44,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.3.0
 	github.com/adrg/xdg v0.5.3
 	github.com/akuity/kargo/api v0.0.0
-	github.com/akuity/kargo/pkg/client/generated v0.0.0
+	github.com/akuity/kargo/pkg/x/client/generated v0.0.0-00010101000000-000000000000
```
krancour: This is cosmetic, but we should just use `v0.0.0` as on the line above. Both are modules that only exist locally and come in through a `replace` directive. It's odd to use different schemes for their pseudo-versions.

hack/codegen/generate-go-client.sh
```diff
@@ -0,0 +1,125 @@
...
+  --skip-validate-spec
+
+# --- Step 3: restore go.mod/go.sum (or bootstrap them on a first-ever run) --
+if [[ -f "${WORK_DIR}/go.mod" ]]; then
+  cp "${WORK_DIR}/go.mod" "${OUT_DIR}/go.mod"
+else
+  cat > "${OUT_DIR}/go.mod" <<'EOF'
+module github.com/akuity/kargo/pkg/x/client/generated
+
+go 1.26.0
```
krancour: This was the problematic thing we were working around. It's still problematic now, but only becomes so under specific circumstances.

I'm contemplating under what scenario we're concerned with `go.mod` not having previously existed in the output dir, which is when we'd run into this...

The only scenario I can think of is one where codegen has _previously_ failed after the `rm -rf` of the output dir.

Codegen will fail locally from time to time. It's the nature of the beast. As it stands now, when we recover, we mess up the `go.mod` and `go.sum` when we do.

I think I'd rather just do nothing in that case. The following `go mod tidy` will fail, but will fail loudly. The fix is easy enough. `git checkout` the missing files.

All this is to say, let's remove the `else`.

krancour:
@fuskovic the back end and automation all LGTM now.

Holding on actual approval until @rpelczar has had a chance to look at the small UI changes.

@rpelczar the TS client is generated from the same tool as before (orval). The difference is some mistakes in `swagger.json` were fixed.

As far as I can tell, this is an improvement and it eliminates some workarounds you previously required...

But I will let you be the judge.

rpelczar (approved):

krancour:
I'm investigating one more possible correctness improvement to the `swagger.json` spec that would be best made now while we are already touching it for exactly reasons or correctness.

krancour (dismissed):

krancour:
> I'm investigating one more possible correctness improvement to the swagger.json spec that would be best made now while we are already touching it for exactly reasons or correctness.

Two fields previously described as `*apiextensionv1.JSON` (opaque objects) were better described as `[]apiextensionv1.JSON` (arrays of opaque objects). This results in a more accurate `swagger.json`. This is piggybacking on the `swagger.json` changes already in this PR.

Importantly, wrt _all_ `swagger.json` changes introduced in this PR, clients generated from older `swagger.json` will still work, because the shape of data on the wire does not change. It is only described more accurately. However, note the older Go client, specifically -- and therefore older CLI versions -- remains susceptible to the phantom object problem that, more broadly, sparked this PR to begin with. The release notes will call this out prominently.

An honest accounting of what _does_ break: The corrected `swagger.json` means newly-generated clients might impose small changes on the code that uses them. Though technically breaking, this is a very small price to pay to finally have _correct_ `swagger.json`. Again, release notes will call this out.

PR #6654, opened 2026-07-18
fix(controller): keep dead network connections from hanging artifact discovery indefinitely

### LINKED ISSUE #6651: Warehouse discovery has no timeout and the git subprocess is uncancellable — a stalled network call permanently parks a Warehouse's reconcile until the controller is restarted
**Affected version:** v1.10.9 (mechanism unchanged on `main` at time of writing)

**Summary:** If a single Warehouse's artifact discovery makes an external call (git or registry) that stalls at the network layer, the reconcile goroutine blocks indefinitely. Because the controller has already persisted `Reconciling=True / Ready=False` (reason `DiscoveryInProgress`) before the call, and because the hung goroutine never returns `Done` to the per-key workqueue, that Warehouse is **permanently parked** — no interval requeue, `refresh` annotation, or generation change re-dispatches it — until the controller pod is killed. The rest of the controller (other Warehouse workers, the Promotion controller) keeps working, so this presents as *one* Warehouse silently stuck while everything else is healthy.

**Mechanism / root cause:**
- No `context.WithTimeout`/`WithDeadline` anywhere in the discovery path (`pkg/controller/warehouses`, `pkg/subscription`, `pkg/image`, `pkg/controller/git`). The reconcile receives the manager's process-lifetime context.
- Git discovery shells out with `exec.Command`, not `exec.CommandContext` (`pkg/controller/git/base_repo.go:426-427`), so the `git` subprocess is not bound to the reconcile context and cannot be cancelled; there is no `--timeout` / `GIT_HTTP_LOW_SPEED_LIMIT`/`TIME`. A `ls-remote`/clone/fetch against an endpoint that black-holes the connection blocks on `cmd.Wait()` forever.
- Image discovery builds its client on `cleanhttp.DefaultTransport()` (`pkg/image/repository_client.go:102`) with no `ResponseHeaderTimeout` / overall timeout; `remote.List`/`remote.Get` pass a context that never cancels.
- Discovery persists the in-progress conditions before the long-running call (`pkg/controller/warehouses/warehouses.go:216-252`); the hung goroutine never calls `Done`, so per the client-go workqueue contract the in-processing key is marked dirty but never re-dispatched.

**Impact:** A transient network condition (e.g. a node scale-down/drain that severs a mid-flight discovery connection) permanently disables discovery for the affected Warehouse with no self-recovery. Only `kubectl rollout restart` of the controller clears it.

**Suggested fix:** Bound discovery with a per-reconcile `context.WithTimeout`, and use `exec.CommandContext` (plus git network timeouts) so a stalled git/registry call is cancelled and the reconcile returns/errors and requeues instead of hanging.

**Repro (sketch):** point a Warehouse's git subscription at an endpoint that accepts the TCP connection but never responds; trigger discovery; observe the Warehouse stuck in `DiscoveryInProgress` indefinitely while other Warehouses/promotions proceed, recovering only on controller restart.

krancour:
Fixes #6651: a connection that went silent mid-discovery would park a Warehouse's reconcile goroutine until the controller was restarted.

- Git invocations now set `GIT_HTTP_LOW_SPEED_LIMIT/TIME`, aborting any transfer moving no data (test included: clone from a silent server).
- Image and OCI chart registry transports set a 30s `ResponseHeaderTimeout`.
- The classic chart repo selector now honors its context (previously discarded) and gets the same header bound instead of `http.DefaultClient`.

Stall detection was chosen over timeouts because it bounds lack of progress, not total duration: dead connections die in seconds while large, slow, healthy clones are never cut off. A comment in the reconciler marks where a deadline would go if this ever proves insufficient.

The context plumbing through `pkg/controller/git` is not strictly needed for this fix; it's kept as a general improvement (cancellable subprocesses, clean shutdown) that also positions us to add deadlines quickly if ever warranted.

Note to reviewers: Commit 4073dc1a6050c349668bc7072880d55f25ca79b9 is the largest, but also the most mechanical. It is nothing but plumbing of context throughout the various git interfaces and their consumers. The other commits are substantially smaller and more worthy of scrutiny.

Possibly makes it into v1.11.0, but will not block the release. Can be deferred to v1.11.1.

### DIFF
--- a/pkg/controller/git/bare_repo.go
+++ b/pkg/controller/git/bare_repo.go
@@ -3,6 +3,7 @@ package git
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -15,10 +16,14 @@ import (
 // BareRepo is an interface for interacting with a bare Git repository.
 type BareRepo interface {
 	// AddWorkTree adds a working tree to the repository.
-	AddWorkTree(path string, opts *AddWorkTreeOptions) (WorkTree, error)
+	AddWorkTree(
+		ctx context.Context,
+		path string,
+		opts *AddWorkTreeOptions,
+	) (WorkTree, error)
 	// Close cleans up file system resources used by this repository. This should
 	// always be called before a repository goes out of scope.
-	Close() error
+	Close(ctx context.Context) error
 	// Dir returns an absolute path to the repository.
 	Dir() string
 	// HomeDir returns an absolute path to the home directory of the system user
@@ -26,14 +31,14 @@ type BareRepo interface {
 	HomeDir() string
 	// RemoteBranchExists returns a bool indicating if the specified branch exists
 	// in the remote repository.
-	RemoteBranchExists(branch string) (bool, error)
+	RemoteBranchExists(ctx context.Context, branch string) (bool, error)
 	// RemoveWorkTree removes a working tree from the repository. The working tree
 	// will be removed from the file system.
-	RemoveWorkTree(path string) error
+	RemoveWorkTree(ctx context.Context, path string) error
 	// URL returns the remote URL of the repository.
 	URL() string
 	// WorkTrees returns a list of working trees associated with the repository.
-	WorkTrees() ([]WorkTree, error)
+	WorkTrees(ctx context.Context) ([]WorkTree, error)
 }
 
 // bareRepo is an implementation of the BareRepo interface for interacting with
@@ -83,6 +88,7 @@ type BareCloneOptions struct {
 // will also perform any setup that is required for successfully authenticating
 // to the remote repository.
 func CloneBare(
+	ctx context.Context,
 	repoURL string,
 	clientOpts *ClientOptions,
 	cloneOpts *BareCloneOptions,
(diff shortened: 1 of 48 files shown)


EronWright (commented):
Nicely reasoned fix — stall detection over a total-duration timeout is the right model for discovery, and the mechanical `ctx` plumbing is well-isolated to one commit as flagged. The `Cancel`/`WaitDelay` handling (SIGTERM-then-delay, with the comment explaining why a non-zero `WaitDelay` is required) is careful and correct. A few thoughts:

**`ResponseHeaderTimeout` only guards time-to-first-header — consider a total request timeout for mid-body stalls.** git's low-speed limit protects the transfer mid-stream, but on the registry/chart HTTP transports a connection black-holed *after* headers arrive (e.g. mid-blob-download) would still hang indefinitely. Discovery mostly reads small manifests/tags, so this is largely fine in practice, but to close the gap symmetrically you could set an `http.Client.Timeout` or derive a per-request context deadline on those clients.

A few smaller points:

- **SSH transfers are unprotected.** `GIT_HTTP_LOW_SPEED_*` is HTTP(S)-only, and with no deadline applied yet, context won't fire either — so #6651's failure mode remains reachable over SSH until its removal in v1.13.0. Just worth noting explicitly.
- **No tests for the registry/image `ResponseHeaderTimeout` changes**, whereas the git and chart-selector paths did get behavioral tests. Testing it directly is awkward, so this is minor — just an uneven-coverage note.
- **`responseHeaderTimeout` is duplicated** (same value + comment) across `pkg/image/repository_client.go`, `pkg/helm/registry.go`, and `pkg/helm/chart/http_selector.go`. Different packages so no conflict, but a candidate for a shared constant.
- **`http_selector` now builds a fresh transport per `Select` call** even on the secure path (previously the shared `http.DefaultClient`), losing cross-call connection pooling. Negligible at discovery frequency, just a behavior change to be aware of.

None of these block merging.

krancour:
> ResponseHeaderTimeout only guards time-to-first-header — consider a total request timeout for mid-body stalls.

I don't actually care how long the request takes. I'm just trying to avoid black holes. I realize that this is asymmetric with the approach on the git side where we're looking at rate of transfer instead. The strategies are indeed different, but what is common to both cases is that we're reaching for the simplest mechanism available to determine that there's something responding on the other end of the connection.

> SSH transfers are unprotected

I'm unconcerned with this since Kargo's support for git over SSH is deprecated.

> No tests for the registry/image ResponseHeaderTimeout changes

`cleanhttp.DefaultTransport()` returns a `http.Transport` from the stdlib and I think it's a safe assumption that `ResponseHeaderTimeout` has been well-tested in the stdlib's own tests. I feel fine taking it for granted that it works.

> responseHeaderTimeout is duplicated

It is, but I'm making a judgement call. 30s is a reasonable and generous timeout that all three uses here agree with, but I don't think that's a universally suitable value such that it deserves to be hoisted into some other package for the sake of reuse.

> http_selector now builds a fresh transport per Select call

This is true and I did give this a lot of consideration. I cannot see a clean way to hoist a pooled client/transport out of here. Some subscriptions use `InsecureSkipVerify`, but most do not. We'd have to maintain _two_ long-lived clients/transports. More damning, however, selectors are ephemeral by design. Long-lived selectors would be a sizeable architectural shift that I don't think the would-be connection pools justify. The only remaining possibility would be package-level clients/transports and package-level state always gives me the ick. It would work, but this would need to be a real problem before I'd want to fall back on that.

fuskovic (approved):
LGTM

PR #6660, opened 2026-07-19
fix: correct misleading doc comment

magic-peach:
## Summary
- Remove `fmt.Println` debug output that could leak sensitive OIDC response body content to stdout
- Fix contradictory doc comment in `Exec()` function

## Changes
- `pkg/server/option/auth.go`: Removed `fmt.Println(string(bodyBytes))` on JSON unmarshal error path. This debug output was leaking the full OIDC discovery response (which may contain sensitive endpoint URLs) to stdout in production.
- `pkg/exec/exec.go`: Fixed doc comment that incorrectly stated "When the command completes successfully, with a non-zero exit code, the error is nil" - this is self-contradictory. Corrected to "When the command completes successfully, the error is nil."

## Impact
The `fmt.Println` in auth.go could leak potentially sensitive OIDC provider configuration data to application logs/stdout. The doc comment fix improves code clarity for future contributors.

## Checklist
- [x] Code follows the project's coding style
- [x] Tests added if applicable

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -28,13 +28,12 @@ func (e *ExitError) Error() string {
 
 // Exec is a custom replacement for cmd.CombinedOutput(). It executes the
 // provided command and returns the command's combined output (stdout + stderr)
-// and an error. When the command completes successfully, with a non-zero exit
-// code, the error is nil. If the command's exit code is non-zero, the error is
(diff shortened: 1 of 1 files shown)


krancour:
After a rebase, the `fmt.Println()` is gone already.

krancour (approved):

PR #6662, opened 2026-07-20
fix: invalid httpsGet handler

### LINKED ISSUE #6653: chart: API probes render invalid `httpsGet` handler when api.tls.enabled=true (1.11 regression)
### Summary

Since #6634 ("replace grpc health probe with new REST health check"), the API server's three probes render `httpsGet:` when `api.tls.enabled=true`. `httpsGet` is not a valid Kubernetes probe handler — the `Probe` schema accepts only `exec`, `httpGet`, `tcpSocket`, and `grpc`. Because `api.tls.enabled` and `api.probes.enabled` both default to `true`, a minimal install renders a Deployment the API server rejects.

Affects `main`, `release-1.11`, and all `v1.11.0-rc.*`. Not present in `v1.10.9`.

### Reproduction

```sh
helm template kargo charts/kargo -n kargo \
  --set api.adminAccount.passwordHash='$2y$10$dummy' \
  --set api.adminAccount.tokenSigningKey=dummy \
  | grep -c httpsGet
```

| ref | `httpsGet` occurrences |
|---|---|
| `v1.10.9` | 0 |
| `v1.11.0-rc.1` / `rc.4` | 3 |
| `release-1.11` | 3 |
| `main` | 3 |

### Rendered output (invalid)

```yaml
livenessProbe:
  httpsGet:
    path: /healthz
    port: 8080
    scheme: HTTPS
```

Applying this fails schema validation, e.g.:

```
ValidationError(Deployment.spec.template.spec.containers[0].livenessProbe):
unknown field "httpsGet"
```

`kubectl apply --dry-run=server` uses `--validate=strict` by default, so this is rejected at apply time rather than surfacing as a readiness failure later. Installs that use `--atomic` will roll back.

### Root cause

`charts/kargo/templates/api/deployment.yaml` — the conditional introduced in #6634 emits a different *key* per branch rather than a different `scheme`:

```
-          grpc:
+          httpsGet:      # <- when api.tls.enabled
...
-          grpc:
+          httpGet:       # <- otherwise
```

### Suggested fix

The block already sets `scheme: HTTPS`, so the handler key just needs to be `httpGet` in both branches:

```yaml
httpGet:
  path: /healthz
  port: 8080
  scheme: HTTPS
```

That makes the two branches differ only by `scheme`, which is the intended distinction.

### Workaround

`--set api.probes.enabled=false` renders cleanly (verified), at the cost of losing the probes.

fuskovic:
closes: https://github.com/akuity/kargo/issues/6653

### DIFF
--- a/charts/kargo/templates/api/deployment.yaml
+++ b/charts/kargo/templates/api/deployment.yaml
@@ -104,44 +104,32 @@ spec:
           protocol: TCP
         {{- if .Values.api.probes.enabled }}
         startupProbe:
-          {{- if .Values.api.tls.enabled }}
           httpGet:
             path: /healthz
             port: 8080
+            {{- if .Values.api.tls.enabled }}
             scheme: HTTPS
-          {{- else }}
-          httpGet:
-            path: /healthz
-            port: 8080
-          {{- end }}
+            {{- end }}
           {{- with .Values.api.probes.startupProbe }}
           {{- toYaml . | nindent 10 }}
           {{- end }}
         livenessProbe:
-          {{- if .Values.api.tls.enabled }}
           httpGet:
             path: /healthz
             port: 8080
+            {{- if .Values.api.tls.enabled }}
             scheme: HTTPS
-          {{- else }}
-          httpGet:
-            path: /healthz
-            port: 8080
-          {{- end }}
+            {{- end }}
           {{- with .Values.api.probes.livenessProbe }}
           {{- toYaml . | nindent 10 }}
           {{- end }}
         readinessProbe:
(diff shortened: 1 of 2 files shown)


charts/kargo/templates/api/deployment.yaml
krancour: Since most of this whole block repeats for both cases, would it be better to have the whole block be unconditional, with just the scheme within that block being conditional? In fact, that is how it's been done already for the external webhooks server.

Applies to the other probes as well.

krancour (approved):

PR #6667, opened 2026-07-20
feat(stages): add pluggable PromotionDispatcher extension point

EronWright:
## What

Introduces a `PromotionDispatcher` extension point in the regular Stage
reconciler, controlling which of a Stage's eligible **pending** Promotions is
selected to run next.

## Why

Today `RegularStageReconciler.syncPromotions` sorts a Stage's Promotions and
records the highest-priority one as `Stage.Status.CurrentPromotion` (the
Promotion reconciler then flips it `Pending → Running`). The ordering is
hard-coded FIFO (oldest ULID) with a Running-first invariant. Making the
selection pluggable lets alternative strategies (e.g. priority or fairness
between pending Promotions) be introduced without changing the reconciler, and
makes the selection logic easier to test in isolation. The default keeps the
current behavior.

## What changed

- New `PromotionDispatcher` interface with
  `Select(ctx, stage, pending) (*Promotion, error)`. The full eligible-pending
  slice is passed so an implementation can consider the pending Promotions
  relative to one another.
- `defaultPromotionDispatcher` preserves the historical FIFO behavior, so this
  change is **behavior-preserving** by default.
- Supplied via a new variadic `WithPromotionDispatcher(...)` option on
  `NewRegularStageReconciler` (existing call sites are unaffected).
- The dispatcher is consulted only when the `CurrentPromotion` slot is open and
  no Promotion is Running — the reconciler keeps ownership of the Running-first
  invariant and the verification/health gates, so a dispatcher can never
  preempt a mid-flight Promotion.
- The "finished promoting" check and the current selection now key off the
  current Promotion's own terminal status rather than whether it remains the
  highest-priority Promotion. This keeps a non-oldest selection **sticky**
  instead of flip-flopping `CurrentPromotion` each pass.

## Tests

Extended `TestRegularStageReconciler_syncPromotions` with cases for: default ==
FIFO parity, a custom dispatcher selecting a non-oldest pending Promotion, the
dispatcher declining (slot stays open, Stage requeued), a Running Promotion
never being preempted, dispatcher error surfacing as `Promoting=Unknown`, a
non-oldest selection remaining sticky, and freeing the slot once the current
selection is terminal.

---

> Draft: opened for early feedback on the interface shape and the
> selection/guard refactor in `syncPromotions`.

### DIFF
--- a/pkg/controller/stages/regular_stages.go
+++ b/pkg/controller/stages/regular_stages.go
@@ -75,23 +75,70 @@ func ReconcilerConfigFromEnv() ReconcilerConfig {
 }
 
 type RegularStageReconciler struct {
-	cfg            ReconcilerConfig
-	client         client.Client
-	eventSender    kargoEvent.Sender
-	healthChecker  health.AggregatingChecker
-	shardPredicate controller.ResponsibleFor[kargoapi.Stage]
+	cfg                 ReconcilerConfig
+	client              client.Client
+	eventSender         kargoEvent.Sender
+	healthChecker       health.AggregatingChecker
+	shardPredicate      controller.ResponsibleFor[kargoapi.Stage]
+	promotionDispatcher PromotionDispatcher
 
 	backoffCfg wait.Backoff
+}
+
+// PromotionDispatcher selects which Promotion, from the set of eligible pending
+// Promotions for a Stage, should become the Stage's current (running)
+// Promotion. It is consulted only when no Promotion is currently Running for the
+// Stage and at least one Pending Promotion has passed the controller's
+// eligibility gates (verification and health). Implementations therefore never
+// need to preserve the invariant that a Running Promotion is not preempted, and
+// may reason about conflicts across the full set of pending Promotions.
+type PromotionDispatcher interface {
+	// Select returns the Promotion to start next, or nil to start none on this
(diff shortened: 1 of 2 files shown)


EronWright:
Withdrawing this.

Promotion Windows was the motivating consumer, and its design has moved to **admission-only** enforcement: an Enterprise validating webhook on Promotion CREATE, backed by an Enterprise-side sweep of still-`Pending` Promotions when a window closes. Neither needs a hook in the OSS Stage controller, so the seam no longer has a caller.

More to the point, windows only ever needed *"not yet"* — never reordering. A `Select`-shaped seam advertises that a dispatcher may pick a non-oldest pending Promotion, which is a queue-ordering contract, and one worth defining deliberately rather than as a side effect of a scheduling feature. It also carries real correctness weight: non-oldest selection lets Promotions complete out of ULID order, which breaks the completion-order assumption in `syncPromotions`' status replay (the fourth commit here). None of that should land without the ordering discussion it implies.

The API type that motivated this work continues in #6710, and the generic `Forbidden`-tolerance fix — a Stage should not fall into error-backoff because an admission policy denied an auto-Promotion create — is now #6716, standing on its own merits.

The branch is preserved; happy to revive it if a conflict-resolution/queue-ordering proposal comes forward.

PR #6686, opened 2026-07-23
fix(ui): open ArgoCD links in a new tab natively

### LINKED ISSUE #6688: ArgoCD links on Stage nodes can't be opened in a new tab
### Pre-submission checklist

### Description

In the project pipeline view, the ArgoCD link on a Stage node cannot be opened in a new browser tab. The link's real destination lives only in a JavaScript click handler; the anchor's `href` is `#`, so any browser-native navigation ignores it.

### Steps to Reproduce

1. Open a project's pipeline view with a Stage that has one or more associated ArgoCD apps.
2. Click the ArgoCD icon on the Stage node (for multiple apps, a dropdown lists them).
3. Try cmd/ctrl-click, or right-click → "Open in new tab", on an app link.

### Expected Behavior

The ArgoCD application opens in a new tab.

### Actual Behavior

The link opens in the same tab, or a new tab loads the current Kargo page (`#`) instead of the ArgoCD application.

### Kargo Version

Akuity-managed (hosted Kargo).

### Screenshots

The multi-app dropdown lists ArgoCD apps as `<app> - argocd` entries, each rendered as an anchor whose `href` is `#`, so browser-native "open in new tab" lands on the Kargo page rather than the app.

kpurdon:
Closes #6688

## Problem

In the pipeline view, the ArgoCD link on a Stage node can't be opened in a new tab. Cmd/ctrl-click and right-click → "Open in new tab" either open in the same tab or land on the current Kargo page instead of the ArgoCD app.

The dropdown renders each item as `<a href='#'>` with the real destination only in the `onClick` handler. Native new-tab actions follow the anchor's actual `href` — which is `#` — so the browser never sees where the link really points.

## Approach

Put the real destination in `href` and let the browser handle every click mode natively, rather than trying to intercept modifier keys in JS. That's what anchors are for, and it fixes plain click, cmd/ctrl-click, and right-click in one move.

## Implementation

- `argoCDHref()` computes the real destination per app:
  - external ArgoCD → the app URL, rendered with `target="_blank"`
  - in-app extension route → the SPA path, wrapped in `withBasePath` (a real `href` needs the full prefix that the router basename would otherwise add for `navigate`)
- `openArgoCD()` only intercepts the in-app extension route on a plain click, to keep SPA navigation. Modified clicks (cmd/ctrl/shift/middle) fall through to the browser's native new-tab handling.
- Applied the same treatment to the single-app case — antd's `Button` renders an `<a>` when given `href`, so it gets the same behavior.

## Validation

- `pnpm run typecheck` and `eslint` pass on the changed file.
- UI-only change; worth a manual check in a Tilt env that both the single-app link and the multi-app dropdown open ArgoCD in a new tab across click/cmd-click/right-click.

Ticket: None

### DIFF
--- a/ui/src/features/project/pipelines/nodes/argocd-link.tsx
+++ b/ui/src/features/project/pipelines/nodes/argocd-link.tsx
@@ -5,6 +5,7 @@ import React from 'react';
 import { generatePath, useNavigate, useParams } from 'react-router-dom';
 import z from 'zod';
 
+import { withBasePath } from '@ui/config/base-path';
 import { ARGOCD_CONTEXT_KEY, SHARD_LABEL_KEY } from '@ui/config/labels';
 import { paths } from '@ui/config/paths';
 import { useExtensionsContext } from '@ui/extensions/extensions-context';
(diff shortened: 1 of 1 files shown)


Marvin9 (approved):
Tested. LGTM! Thanks

PR #6694, opened 2026-07-26
fix: reconcile UI auto-promotion hold state with the reconciler

krancour:
Prior to this PR, there is a map in Stage status of "active auto-promotion holds." The contents of this map solidify as Promotions with hold or release intent complete (successfully).

In the Stage reconciliation process, quite a bit happens between that point and the point where auto-promotion might create a new Promotion. In that window, additional Promotions could have completed. There might _also_ be pending Promotions that express auto-promotion hold/release intent and should be factored into the decision. To account for both of these possibilities, there's an additional look at recently-completed and pending promotions just before making an auto-promotion decision.

_The problem is what we learn at this point isn't persisted anywhere. The UI's notion of what is or isn't on hold is derived from the persisted, but possibly stale map that the reconciliation process understood not to take at its word when making a final auto-promotion decision._

This PR establishes a second map of "effective auto-promotion holds" that is persisted late in the reconciliation process. While the original map is a durable record of holds that survives GC of the Promotions that established them, the second map is the first overlaid with holds and releases based on what's recently-completed or pending. It is discarded and re-computed in its entirety on every reconciliation. Auto-promotion decisions and UI state are mutually based on this one map. i.e. There is now proper agreement between the back end and the UI.

Why a second map? Why not have two-phase pending/active holds in the original map?

The semantics of the existing map are entry == hold. Absence of entry == no hold. Writing a pending hold to the map when no active hold existed would have been easy. The pending hold could easily be reversed if the Promotion that established it were, for instance, aborted. Pending releases would have been the problem. Without changing the semantics of the map, there is no way to _tentatively_ remove a hold in a way that is reversible if the Promotion that tentatively released the hold were, for instance, aborted.

One other tiny change in the PR: Hold/release isn't predicated at all on Promotion success. Inadvertent failures do not negate user intent. The only thing that _does_ negate hold/release intent is when a Promotion bearing that intent is aborted, since that _does_ signal a change in user intent.

Overall, none of this alters the feature in any appreciable way. It just makes things slightly more intuitive and closes a gap between the back end and the UI.


pkg/controller/stages/regular_stages.go
fuskovic: NIt: It looks like we can use `newAutoPromotionHold` here to keep this DRY

fuskovic (changes requested):

pkg/controller/stages/regular_stages.go
krancour: Nice catch. Fixed in e8d26d572c304ed88b03d6d77a147b7fc24e5afe

fuskovic (approved):
LGTM 🚢

PR #6705, opened 2026-07-27
fix: improvements to doc publishing script

krancour:
First attempt to promote the `release-1.11` docs to prod this morning failed. My Netlify PAC `NETLIFY_AUTH_TOKEN` was expired, but every API call using it failed silently, creating the impression that the script has succeeded.

This PR hardens the script to actually check response codes from curl-based Netlify API calls.


fuskovic (approved):
LGTM

PR #6617, opened 2026-07-14
chore: remove connect rpc methods

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6550

PR #6711, opened 2026-07-27
chore(deps): bump Kubernetes to v0.36.3 and controller-runtime to v0.24.1

krancour:
Brings Go dependencies forward across all eight modules, now that the gRPC removal blocking these upgrades has landed (#6617).

Three commits: the bump, the source changes it forces, and regenerated clients.

controller-runtime v0.24 is a breaking change. `NewWebhookManagedBy` takes the API type as an argument instead of via `For()`, and the typed `WithValidator`/`WithDefaulter` replace the deprecated `WithCustom*`, so the admission methods are retyped from `runtime.Object` to concrete types — which lets 31 forced type assertions go away. `SubResourceClient` also gained `Apply`.

Two fixes surfaced along the way: the deprecated `k8s.io/utils/strings/slices` gives way to the standard library, and `AnnotatedEventf` was being handed a runtime value as its format string, which would have mangled any event message containing a `%`.

Also repoints the Dependabot `gomod` entry for `/pkg`, which has matched no manifest since #5089 — hence the rotted `go.sum` in `pkg/x/client/fidelity`.

Client regeneration is additive only: three optional fields and five new models from the Kubernetes 1.36 `core/v1` surface.


fuskovic (approved):

hack/test/e2e/go.mod
```diff
@@ -14,23 +14,15 @@ replace (
 	github.com/akuity/kargo/hack/test/e2e/envfuncs => ./framework/envfuncs
 )
 
-replace (
-	helm.sh/helm/v3 => helm.sh/helm/v3 v3.19.5
```
EronWright, line 18: I suppose helm wasn't an actual dep?  not seeing it below.

EronWright (approved):
Great to see the typing on the webhooks.

PR #6716, opened 2026-07-27
fix(stages): tolerate admission denial of auto-promotion creates

EronWright:
When auto-promotion creates a Promotion, any error from `r.client.Create` is treated as fatal to the reconcile. An admission webhook denial is therefore indistinguishable from a genuine failure.

The consequence: a cluster running any admission policy that rejects Promotion creates — Kyverno, Gatekeeper, OPA, or any other validating webhook — drives every affected Stage into error-backoff for as long as the policy applies, with a `Reconciling` condition and a steady stream of reconcile errors in the logs. Nothing recovers until the policy is removed.

## The fix

A denial is a decision, not a failure. The API server persists nothing when a create is denied, so the reconcile has no partial state to unwind and nothing to retry imperatively.

`autoPromoteFreight` now treats `apierrors.IsForbidden` as non-fatal: it logs the denial and continues to the next requested Freight instead of returning an error. This is level-triggered — each subsequent reconcile re-derives the candidate and re-attempts the create, so promotion resumes on its own once the denying policy no longer applies. Any other create error remains fatal.

## Why no event

An earlier revision of this PR recorded an `AutoPromotionDenied` event here. That was wrong, and the type is gone.

A policy that denies keeps denying, so this branch is taken on every reconcile for as long as it holds — roughly every five minutes per affected Stage per blocked origin. An event per occurrence reports one unchanging fact indefinitely: a multi-hour freeze over a handful of Stages is thousands of Event writes, and `GET /v1beta1/projects/{project}/events` returns nothing but denials, which is worse than silence for the operator trying to see what actually happened.

Kargo's recorder is deliberately non-aggregating (`pkg/kubernetes/event`: "records all events without aggregation… without throttling / spam filtering"), because each event is meant to be one entry in an audit trail. That is the right default and this is simply not an event: it is a persistent condition, and a condition belongs in status. Kargo Enterprise, where promotion windows deny on purpose, reports it in `Stage.status.promotionSchedule` (#6737) — open or closed, why, and when it next changes.

## Testing

An `autoPromoteFreight` case returns a `Forbidden` `StatusError` from a create interceptor and asserts the reconcile succeeds, no Promotion is persisted, and nothing is recorded.

Verified end to end against a live cluster (Kargo Enterprise promotion windows, which deny Promotion creates through a validating webhook): with a Stage under an active Deny window and auto-promotion enabled, the Stage stays out of error-backoff, its conditions read `NoFreight` rather than `LastPromotionErrored`, and auto-promotion resumes unattended within the reconciler's 5-minute resync once the window reopens.

No API or CRD change; no codegen; no docs change.

### DIFF
--- a/pkg/controller/stages/regular_stages.go
+++ b/pkg/controller/stages/regular_stages.go
@@ -1994,6 +1994,25 @@ func (r *RegularStageReconciler) autoPromoteFreight(
 		// PromotionTemplate.
 		promotion := api.NewMinimalPromotion(stage, candidate.Name)
 		if err = r.client.Create(ctx, promotion); err != nil {
+			// An admission webhook may deny the create. Tolerate this as a
+			// non-error: nothing is persisted, so this reconcile moves on, and a
+			// later one re-derives and re-attempts the auto-promotion once the
+			// denying policy no longer applies. Any other error is still fatal to
+			// the reconcile.
+			//
+			// Deliberately not recorded as an event. A policy that holds keeps
+			// denying, so this branch is taken on every reconcile for as long as
+			// it does; an event per occurrence would report one unchanging
+			// condition thousands of times and bury the Project's event feed. A
+			// condition belongs in status, and Kargo Enterprise reports the one
+			// it knows about in Stage.status.promotionSchedule.
+			if apierrors.IsForbidden(err) {
+				freightLogger.Debug(
+					"auto-promotion was denied by an admission webhook",
+					"error", err.Error(),
+				)
+				continue
+			}
 			return newStatus, fmt.Errorf(
 				"error creating Promotion for Freight %q in namespace %q: %w",
 				candidate.Name, stage.Namespace, err,
--- a/pkg/controller/stages/regular_stages_test.go
+++ b/pkg/controller/stages/regular_stages_test.go
@@ -7546,6 +7546,93 @@ func TestRegularStageReconciler_autoPromoteFreight(t *testing.T) {
 			},
 		},
 		{
+			// A Forbidden error from an admission webhook must not fail the
+			// reconcile: nothing is persisted and the pass continues, so a later
+			// reconcile can re-attempt once the denying policy no longer applies.
+			name:                 "tolerates a forbidden Promotion create",
+			autoPromotionEnabled: true,
+			stage: &kargoapi.Stage{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "fake-project",
+					Name:      "test-stage",
+				},
+				Spec: kargoapi.StageSpec{
+					RequestedFreight: []kargoapi.FreightRequest{
+						{
+							Origin: kargoapi.FreightOrigin{
+								Kind: kargoapi.FreightOriginKindWarehouse,
+								Name: "test-warehouse",
+							},
+							Sources: kargoapi.FreightSources{
+								Direct: true,
+							},
+						},
+					},
+					PromotionTemplate: &kargoapi.PromotionTemplate{
+						Spec: kargoapi.PromotionTemplateSpec{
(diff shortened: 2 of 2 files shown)


pkg/controller/stages/regular_stages.go
```diff
@@ -1994,6 +1994,31 @@ func (r *RegularStageReconciler) autoPromoteFreight(
 		// PromotionTemplate.
 		promotion := api.NewMinimalPromotion(stage, candidate.Name)
 		if err = r.client.Create(ctx, promotion); err != nil {
+			// An admission webhook may deny the create. Tolerate this as a
+			// non-error: nothing is persisted, so this reconcile records one event
+			// and continues, and a later reconcile re-derives and re-attempts the
+			// auto-promotion once the denying policy no longer applies. Any other
+			// error is still fatal to the reconcile.
+			if apierrors.IsForbidden(err) {
```
krancour, line 2009: I know this is intended to be tolerant of admission that fails due to _policy enforcement_ and I can very clearly see where such a failure is arguably a 403, but it does defy what I think may be a colloquial understanding of 403 -- that the error is tied to insufficient permissions.

RFC 9110 has this delightfully ambiguous text in its desription of a 403 status code:

> If authentication credentials were provided in the request, the server considers them insufficient to grant access. The client SHOULD NOT automatically repeat the request with the same credentials. The client MAY repeat the request with new or different credentials. However, a request might be forbidden for reasons unrelated to the credentials.

https://www.rfc-editor.org/rfc/rfc9110.html#name-403-forbidden:

To me this simultaneously says 403 is and isn't specifically meant to reflect insufficient credentials. 🤷‍♂️ 

So I had been contemplating whether 409 might be more appropriate:

> The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request. The server SHOULD generate content that includes enough information for a user to recognize the source of the conflict.

https://www.rfc-editor.org/rfc/rfc9110.html#name-409-conflict

By descriptions alone, I think I would come down in favor of 409 by a very narrow margin. The thing that I think tips the scales further in its favor is that 409 feels like a more acceptable thing to treat as non-fatal than a 403.

Having had this debate within my own head, I thought this was perhaps worth sharing, but in the end, my opinion on this is rather weakly held.

EronWright: Claude had this rationale, wdyt?

**403 / `Forbidden` is the right default.** A freeze is exactly what 403 means — the server understood the request and refuses to authorize it as a matter of policy, and no amount of retrying by this client changes that. It's also what kubectl renders most legibly: `Error from server (Forbidden): ... denied the request: change freeze in effect until 2026-08-04, see CHG-1234`.

The one case for something else is when your main clients are controllers rather than humans. A GitOps reconciler hitting 403 will just keep requeueing on its own backoff schedule. If you'd rather tell it *when* to come back, return `TooManyRequests` (429) with `details.retryAfterSeconds` set to the time until thaw — `apierrors.SuggestsClientDelay` picks that up and workqueue rate limiters honor it. The tradeoff is that it reads as throttling in logs and metrics, and some client paths will transparently retry it and swallow your message.

Things to avoid:

- **409 `Conflict`** — anything using `RetryOnConflict` will hot-loop against your webhook for the duration of the freeze.
- **5xx** — indistinguishable from your admission plugin being broken, and it corrupts the signal you'd use to decide whether admission itself is healthy.
- **423 Locked** — semantically perfect, but there's no matching `StatusReason`, so client-go can't classify it and kubectl drops the parenthetical entirely.

krancour: I don't feel strongly at all. As long as it was contemplated and you can defend the choice, I'm content with this.

hairyhum (approved):

PR #6722, opened 2026-07-28
feat(credentials/ecr): add cross-account ECR registry support

### LINKED ISSUE #5587: Support cross-account IAM Role assumption for AWS ECR access in Ambient Credentials

krancour:
Supersedes #6521, which superseded #6519. Same work, opened from a branch I'm able to push to.

Fixes #5587

Credit to @rv9812, who wrote the original implementation in #6519 and iterated on it through #6521 — thank you.

### Identity order

Deriving the role ARN solely from the registry's account made that role the only one ever tried. That silently breaks setups where `kargo-project-<project name>` in the **controller's** account reaches the registry through a repository policy: role assumption fails, Kargo drops to the controller's own shared role, and Project-level isolation is lost.

Kargo now tries each of the following in order, stopping at the first that succeeds:

1. `kargo-project-<project name>` in the registry's AWS account
1. `kargo-project-<project name>` in the controller's AWS account, when the accounts differ
1. The controller's own IAM role

### Error handling

Only a denial (HTTP 403) advances to the next identity; any other error is returned. #6521 let every error fall through, so a transient STS or network failure could silently downgrade to the controller's shared identity. HTTP 400 is also no longer treated as a denial — ECR documents `AccessDeniedException` as 403, while 400 covers `ThrottlingException` and `ValidationError`.

Confirmed against STS that a nonexistent role, a nonexistent account, and a role name exceeding IAM's 64-character limit all surface as a 403 through the ECR client's credential resolution, so an absent role advances the chain rather than aborting it.

### Structure

The intermediate refactor is reverted. Generalizing the single fallback into an ordered list is the only departure from the original shape.

### Tests

Cover identity ordering, the same-account collapse, denial-advances-to-next, all-denied, and non-denial-aborts. `smithy-go` becomes a direct dependency because the tests construct a synthetic 403.

### Docs

Rewrote the ECR section of the ambient credentials guide. The identity order is now body text rather than an admonition inserted immediately ahead of a second admonition describing the same fallback.


pkg/credentials/ecr/managed_identity.go
```diff
@@ -179,56 +196,121 @@ func (p *ManagedIdentityProvider) getAuthToken(
...
+	identity authIdentity,
+) (string, time.Time, error) {
+	if identity.roleARN != "" {
+		// The STS client is deliberately built before cfg is retargeted at the
+		// registry's region, so that role assumption continues to use whichever
+		// region the controller's own configuration resolved to.
+		cfg.Credentials = stscreds.NewAssumeRoleProvider(
+			sts.NewFromConfig(cfg),
+			identity.roleARN,
+		)
```
EronWright: To better protect against [confused deputy attacks](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html#mitigate-confused-deputy), the request should carry an `ExternalID` equal to the project name, and project owners should be advised to check it in their trust policy.

```go
provider := stscreds.NewAssumeRoleProvider(
	sts.NewFromConfig(cfg),
	identity.roleARN,
)
provider.RoleSessionName = "kargo-controller"
provider.Options.ExternalID = aws.String(project)

cfg.Credentials = provider
```

The project-level IAM role's trust policy should resemble:
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::KARGO_CONTROLLER_ACCOUNT_ID:role/CONTROLLER_ROLE_NAME"
      },
      "Action": "sts:AssumeRole",
      "Condition": {
        "StringEquals": {
          "sts:ExternalId": "KARGO_PROJECT_NAME"
        }
      }
    }
  ]
}
```

pkg/credentials/ecr/managed_identity.go
```diff
@@ -179,56 +196,121 @@ func (p *ManagedIdentityProvider) getAuthToken(
...
+	// the last resort.
+	return append(identities, authIdentity{
+		description: "controller's own role",
+	})
+}
+
+// getAuthTokenAs attempts to obtain an ECR authorization token for the given
+// region as the given identity. Errors are returned unexamined; interpreting
+// them is the caller's responsibility.
+func getAuthTokenAs(
```
EronWright: This method returns a short-term token, a seems to create a fresh provider each time it's called.  I'm wondering how caching works, e.g. of the assumed-role intermediate token.

EronWright (changes requested):

pkg/credentials/ecr/managed_identity.go
krancour: There isn't a fresh provider each time. There's one long-lived `ManagedIdentityProvider`. See the registry entry in `init()`:

```
func init() {
	if provider := NewManagedIdentityProvider(context.Background()); provider != nil {
		credentials.DefaultProviderRegistry.MustRegister(
			credentials.ProviderRegistration{
				Predicate: provider.Supports,
				Value:     provider,
			},
		)
	}
}
```

krancour: Oh. "Provider" is overloaded here. You are probably talking about the `AssumeRoleProvider`.

I have a vague recollection of having gone down this rabbit hole once a long, long time ago. Let me refresh my memory...

pkg/credentials/ecr/managed_identity.go
krancour: I spent about an hour trying to construct a workable attack. I could not, nor could Claude.

Crucially, I don't think what Kargo does particularly resembles the AWS example. There is no way to fool Kargo into thinking it's acting on behalf of one Project when it's actually acting on behalf of another. That alone makes a confused deputy impossible.

Even if this were not the case, since the assumed role name is derived from the Project name, using the Project name as the external ID, as proposed, would do no real work. If Kargo were wrong about one, it would be wrong about the other in the same way, and the trust policy's condition would always be satisfied.

If you can disprove my first assertion and there _is_ a viable way to fool Kargo into thinking it's acting on behalf of one Project when it's actually acting on behalf of another, then I would happily further entertain the idea of using an external ID, but using Project name as that value would not be the solution.

pkg/credentials/ecr/managed_identity.go
krancour: Went spelunking on this again to remind myself of what I once knew. Since we're caching the token we actually use for ~12 hours, the intermediate token used to obtain it is long-expired by the time we'd ever possibly use it again.

pkg/credentials/ecr/managed_identity.go
EronWright: Adding protections against confused deputies is something we do for the customer's benefit and peace of mind.   It doesn't matter whether it is exploitable today.  I think it is a bad practice to assign security properties to the role name, e.g. that "it always contains the project name" which is an assumption that could break with a future feature enhancement like "make the role name be configurable".

Don't overthink this, it's easy and highly recommended that we use the `ExternalId` feature.  Its opt-in on the part of customers.

As for the external-id value, here's how to think about it.  The Kargo controller has an singular IAM identity that will be trusted by multiple tenants (projects).  Each tenant makes an IAM Role to be assumed by the controller in the context of their project.  Therefore the external ID identifies the project, either by UID or name.

pkg/credentials/ecr/managed_identity.go
EronWright: Huh?  The token returned from `getAuthTokenAs` probably has a one-hour expiration.  I guessing we will re-assume the role and then get a fresh token, on that schedule.  There's an intermediate credential in there that could be cached for more than an hour, I think.

pkg/credentials/ecr/managed_identity.go
krancour: I think the initial disconnect here is the result of different frames of reference. Your initial feedback registered to me as a claim that a confused deputy _was already possible today_. It's not.

Your second comment (thank you) brought things into focus:

> an assumption that could break with a future feature enhancement like "make the role name be configurable".

I'm certain now that you had this future what-if in mind from the start, but neither had you explicitly said so, initially, nor had I inferred it.

Using the Project name as the external ID does no real work _today_, but if I take your point correctly now, that it future-proofs against hypothetical (and not entirely unlikely) changes to Kargo that do indeed make it 
_exactly_ resemble the canonical example in the AWS docs, then I agree.

I'm knocking this down to a draft while I iterate a bit more.

pkg/credentials/ecr/managed_identity.go
krancour: Per the AWS docs (emphasis mine):

> Retrieves an authorization token. An authorization token represents your IAM authentication credentials and can be used to access any Amazon ECR registry that your IAM principal has access to. **The authorization token is valid for 12 hours.**

https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_GetAuthorizationToken.html

We set a TTL on cached authz token that is time-to-expiry minus 5m -- so 11:55. The intermediate credentials, meanwhile, are quite short-lived. `AssumeRoleProvider` defaults to requesting a 15 minute session. The thing we do cache outlives the thing we don't cache by a factor of ~48.

So for the happy path, at least, a throwaway `AssumeRoleProvider` per attempt is just fine. For a less-happy path, when no identity is authorized, we cache nothing at all, so the full chain gets re-walked on every lookup. Critically, holding onto an `AssumeRoleProvider` wouldn't change that. `CredentialsCache` doesn't cache errors, so a retained `AssumeRoleProvider` would call STS and collect the same denial every time. What _would_ change it is remembering the denials themselves, but it's always been deliberate that we don't do that. Doing so would mean an admin's fix on the AWS side doesn't take effect until we forget it. That's a cost, but not one introduced by this PR.

Exploring your question did shake a few other meaningful things loose along the way. We were calling `config.LoadDefaultConfig()` on every cache miss, and since it wraps credentials in a cache of its own, every miss got an empty one and re-resolved the controller's *own* ambient credentials. Since 4cb921d2d384c74b8972b3d4fc62160a9ad6c817, we now hold that config for the life of the provider. That, in turn, made connection pooling both safe and worthwhile. (A pooled transport per cache miss would previously have leaked file descriptors while also seeing little to no connection reuse.) 868f958cf89fd7587144baf422f1d0c80b0a7575 moved the client to using `cleanhttp.DefaultPooledClient()`.

Thanks for motivating some improvements here.

pkg/credentials/ecr/managed_identity.go
krancour: Resolved and thoroughly documented in 745a0b5633ce961809f5db340564635d3c551ac9

EronWright: Looks great, thanks!

docs/docs/40-operator-guide/40-security/50-ambient-credentials.md
```diff
@@ -34,51 +36,137 @@ will be) installed.
...
+IAM role to assume it -- regardless of which account the controller belongs to.
 
 :::
 
-Tokens Kargo obtains for accessing any specific ECR repository on behalf of any
+:::note
+
+The name of the IAM role associated with each Kargo Project is deliberately not
+configurable in order to avert the
+[confused deputy](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html)
```
EronWright, line 124: I would suggest we rephrase this, to not suggest "confused deputy" as the rationale for the lack of configurability (which was designed before the security issue was raised).  Assumedly the naming convention is for _automatic discoverability_.  In concept, a configured role name would bypass discovery (and still be secure). 

We mitigate against "confused deputy" using `sts:ExternalId`.

docs/docs/40-operator-guide/40-security/50-ambient-credentials.md
```diff
@@ -34,51 +36,137 @@ will be) installed.
...
+problem. Trust policies should not, however, be written with an assumption that
+this remains true forever.
+
+To safeguard against future changes, it is recommended that trust policies
+require an external ID equal to the Project's name.
+
+AWS requires an external ID to be at least two characters long, so Kargo
+presents none at all on behalf of a Project whose name is only one character.
```
EronWright: An alternative here would be to not support role assumption on single-character projects, rather than weakening their security properties (by omission of an `ExternalId`).

pkg/credentials/ecr/managed_identity.go
```diff
@@ -162,73 +187,171 @@ func (p *ManagedIdentityProvider) GetCredentials(
...
+	// The region override applies only to the ECR client. Role assumption uses
+	// whichever region the controller's own configuration resolved to.
+	stsSvc := sts.NewFromConfig(p.cfg)
+	ecrSvc := ecr.NewFromConfig(p.cfg, func(o *ecr.Options) {
+		o.Region = region
+		if identity.roleARN != "" {
+			o.Credentials = stscreds.NewAssumeRoleProvider(
+				stsSvc,
+				identity.roleARN,
+				func(aro *stscreds.AssumeRoleOptions) {
```
EronWright, line 335: Let's consider setting the `RoleSessionName` to `kargo-controller`.  Session names are super helpful in CloudTrail logging and are routinely set to a constant.  The default is like `aws-go-sdk`.

```go
aro.RoleSessionName = "kargo-controller"
```

https://github.com/aws/aws-sdk-go-v2/blob/3bf9ef9dce470565ce8e4e5c2dbd673e72607b58/credentials/stscreds/assume_role_provider.go#L278

EronWright (dismissed):
Thanks for integrating `ExternalId`.  Aside from a couple of polish items, seems ready to ship.

docs/docs/40-operator-guide/40-security/50-ambient-credentials.md
krancour: I'll be honest that without having explicitly named "confused deputy" as what we were avoiding, the rationale behind not being configurable, from the start, _was_ in fact preventing assumption of arbitrary roles, and existing docs say as much. "Automatic discoverability" was never of any primary concern, but was a consequence of a strategy for preventing the assumption of an arbitrary role. If I had known then what I know now about external ID, I probably would have made role name configurable from the beginning and automatic discoverability based on a naming convention would probably never have been a thing.

I think we've established that lack of configurability _is_ why we're not susceptible to the confused deputy problem _today_. `ExternalId` as we agreed is a safeguard against the existing mitigation potentially going away _in the future_, and I agree that's a worthwhile precaution, but in the present, the lack of configurability is doing more work to avert the confused deputy problem than the external ID is.

In short, although we may be retroactively putting a name to the problem, I don't think this line in the docs is technically incorrect or historically inaccurate.

docs/docs/40-operator-guide/40-security/50-ambient-credentials.md
krancour: I thought about several possibilities here and convinced myself this was "least bad," which is definitely _not_ the same as _good._

If you have a Project name that's one character long and you attempt to configure a trust policy that requires that one character Project name as the external ID, _I think_ (but should confirm) that AWS just won't allow that. (i.e. Disallowed not only when _sending_ external ID, but also disallowed when _defining_ the trust policy.) In my head, that leads any user who cares about this, very quickly, to the conclusion that they just shouldn't use one character Project names.

Of course, this doesn't guard against the possibility of someone realizing this limitation and saying, "oh well," and moving forward without requiring any external ID. And I fully agree with that being the critical weakness here.

> An alternative here would be to not support role assumption on single-character projects

I considered that _first_, but it's a breaking change.

Re-calibrating...

It's possible a better "least bad" is something else that was rejected. If we sent `kargo-project-<project name>` (same as the current role name, but crucially, derived from the Project name _independently_ from the the derivation of the role name), then the external ID would always be longer than one character. This doesn't place any _significant_, new knowledge burden on the AWS admin, because they already needed to be aware of `kargo-project-<project name>` as the format for the role _name_. The only new thing they're expected to know is that the format for the expected external ID is effectively the same.

This would be non-breaking and sidesteps the minimum length problem entirely.

pkg/credentials/ecr/managed_identity.go
krancour: Sorry... I noticed that in your original suggestion. It's not load-bearing in terms of mitigating (present or future) susceptibility to the confused deputy problem, so I'd been saving it for one last, tangential improvement to possibly tack on after the load bearing parts were settled. I forgot.

Let me make sure I understand the implications and then will almost certainly do this. `kargo-controller` might not be the right value though. There are often _multiple_ controllers in a single Kargo instance.

docs/docs/40-operator-guide/40-security/50-ambient-credentials.md
krancour: Fixed in 0dd26f97add07c9f4f75d4d02e5ef8a289a5efee

pkg/credentials/ecr/managed_identity.go
krancour: Fixed in 24aefd4fc31cf0b8d927db6b469e8398b4f03a94

fuskovic (approved):
LGTM

PR #6737, opened 2026-07-30
feat(api): report promotion window state on Stage status

EronWright:
Adds `PromotionScheduleStatus` to `StageStatus`, reporting whether promotion windows (#6710) currently permit promotion of a Stage.

Admission only answers this question once someone has already tried to promote and failed. This lets a client explain a freeze — or warn of an approaching one — beforehand.

Nothing in OSS evaluates windows, so nothing in OSS writes this field: OSS carries the API for compatibility, as it does for other Enterprise-only fields.

## Variations a client will observe

**No windows gate the Stage.** The common case — the field is absent entirely, not an empty object.

```yaml
status:
  autoPromotionEnabled: true
  # no promotionSchedule key
```

**Frozen, reopening known.**

```yaml
status:
  promotionSchedule:
    closed: true
    reason: 'Promotion freeze "change-freeze" is active'
    nextOpen: "2026-07-30T22:00:00Z"
```

**Frozen, no reopening known.** `nextOpen` is omitted. Here a one-shot `Allow` window has elapsed, so promotion is forbidden for good; an implementation may also report an indefinite freeze this way.

```yaml
status:
  promotionSchedule:
    closed: true
    reason: No promotion window is currently active
```

**Promotion allowed, freeze approaching.** `closed` is omitted (false). `reason` describes the freeze that is coming, so a warning can name it.

```yaml
status:
  promotionSchedule:
    nextClose: "2026-07-31T09:00:00Z"
    reason: 'Promotion freeze "change-freeze" begins'
```

**Promotion allowed, nothing scheduled ahead.** A one-shot `Deny` window that has elapsed leaves no boundary to report.

```yaml
status:
  promotionSchedule: {}
```

## Contract

- `reason` describes the freeze being reported, and `closed` says which one that is: in effect, or coming. It is set whenever `closed` is true **or** `nextClose` is present.
- `nextOpen` is optional **even when `closed` is true**. A schedule need not ever reopen, so requiring it would be a contract the first implementation breaks. No CEL validation links the two. Treat its absence as "frozen, with no known end" and render `reason`.
- `nextClose` is meaningful only while promotion is allowed, and is likewise optional.
- At most one of `nextOpen` / `nextClose` appears: only the boundary ahead is reported.
- **Advisory.** The value is only as current as the Stage's last reconciliation. Admission remains the sole arbiter of whether a Promotion is permitted, so this should not gate a promote action — a client that treats it as authorization will eventually act on a stale value.

`AutoPromotionEnabled` is the precedent: likewise a status field derived from `ProjectConfig` so clients need not recompute policy.

## A note on wording

The messages say "promotion freeze" rather than "Deny promotion window", and "promotion window" rather than "Allow promotion window". `kind: Allow|Deny` remains the configuration vocabulary, where an explicit pair is unambiguous; it just reads poorly in a sentence, and these strings are shown to people.

## Testing

API-only; no behavior change. Full codegen (CRD, deepcopy, OpenAPI, generated clients, UI schema and models). `api` and root modules build and their tests pass. The Stage CRD gains one optional object under `status`, so the change is additive for existing resources. Every shape above was produced by an Enterprise build against a live cluster rather than written by hand.


hairyhum:
Replaced by https://github.com/akuity/kargo/pull/6772 - use repo-local branch, minor changes in field names.

PR #6747, opened 2026-07-31
fix(credentials): coalesce concurrent credential cache misses

### LINKED ISSUE #6732: Coalesce concurrent credential cache misses
Most credential providers that cache short-lived tokens do a plain check-then-act against their cache. Nothing coordinates the gap between the cache miss and the cache write, so concurrent callers that miss the same key each perform the full token acquisition independently.

Note that every caller still ends up with a valid credential, so this issue is not dangerous -- it's merely wasteful, and worth fixing since #6630 already did this for `git.AppCredentialProvider` and established the pattern that the rest of the providers can follow.

krancour:
Providers that cache short-lived tokens did a plain check-then-act against their caches, so concurrent callers that missed the same key each performed a full token acquisition. A thundering herd is not possible because of the controller's bounded concurrency, however, on the occasion that two or more callers do concurrently experience a cache miss for the same key, they do race to perform duplicate work and cache their result. Acquiring a new token does not invalidate like tokens already issued, so the "last one wins" mechanics are not dangerous, but the duplicate work is genuinely wasteful.

The GitHub App credential provider was the only one with an existing answer to this problem. #6630 increased the relative cost of duplicate work performed by that particular credential provider, so that same PR introduced the pattern of concurrent callers that missed the same key coalescing on one execution of the process that acquires (and tests) a new token. This PR generalizes the solution introduced by #6630 and formalizes it into a reusable component shared by all the credential providers.

The coalescing load-through cache subtly raises the stakes on latent defects discovered during its introduction, so this commit corrects those as well: three token requests that nothing bounded, four responses dereferenced without a check (one at startup, where nothing recovers a panic), and a credential already past its expiry cached as though its lifetime were merely unknown — served, in ECR's case, for as long as ten hours.

Closes #6732

### DIFF
--- a/pkg/credentials/ecr/managed_identity.go
+++ b/pkg/credentials/ecr/managed_identity.go
@@ -15,8 +15,8 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/ecr"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/hashicorp/go-cleanhttp"
-	"github.com/patrickmn/go-cache"
 
+	"github.com/akuity/kargo/pkg/cache/coalescing"
 	"github.com/akuity/kargo/pkg/credentials"
 	"github.com/akuity/kargo/pkg/logging"
 )
@@ -34,8 +34,18 @@ func init() {
 	}
 }
 
+// managedIdentityInput identifies the token a load should obtain. The cache key
+// is a hash of these values, so they cannot be recovered from it.
+type managedIdentityInput struct {
+	region  string
+	project string
+}
+
 type ManagedIdentityProvider struct {
-	tokenCache *cache.Cache
+	// tokenCache holds authorization tokens keyed by a hash of the region and
+	// Kargo Project they were obtained for. It fills its own misses, coalescing
+	// concurrent loads for any given key.
+	tokenCache coalescing.Cache[managedIdentityInput, string]
 
 	accountID string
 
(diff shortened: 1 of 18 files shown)


pkg/credentials/ecr/managed_identity.go
```diff
@@ -126,40 +153,59 @@ func (p *ManagedIdentityProvider) GetCredentials(
...
-		// project-specific IAM role, or that the project-specific IAM role doesn't
-		// have the necessary permissions to get an ECR auth token. We're making
-		// a choice to consider this the will of the AWS admins and not a controller
-		// error. We'll just log it and move on as if we found no credentials.
-		return nil, fmt.Errorf("error getting ECR auth token: %w", err)
+		// An IAM role not authorized to assume the project-specific role, or a
+		// project-specific role not authorized to obtain an ECR auth token, is
+		// taken to be the will of the AWS admins rather than an error, and reaches
+		// here as an empty token rather than as this error. See getAuthToken.
+		return "", nil, fmt.Errorf("error getting ECR auth token: %w", err)
```
krancour, line 248: Note to self: This isn't a change in behavior. It's a change in the explanation of the behavior, because the previous explanation didn't match what was actually happening. I think the behavior that was described in the comment was correct and the code was wrong. i.e. I think this little piece needs a closer look.

krancour: I've now verified this to be correct, as-written. The updated comment is a more accurate description of behavior that was already correct.

fuskovic (approved):
Looks solid

PR #6753, opened 2026-08-02
feat(promotion): allow HTTP request bodies from workdir files

### LINKED ISSUE #6743: Support uploading tar archive from http step
### Pre-submission checklist

### Proposed Feature

As discussed in https://github.com/akuity/kargo/issues/1099#issuecomment-4745602703, it would be useful to be able to upload files from the workdir. Support for creating an archive from the workdir was added in #6666. The `http` step now needs to be extended to support uploading a file in addition to a string `body`.

The intent of opening this issue is to determine the ideal interface extension to the `http` step so that it can be implemented.

### Motivation and Use Case

Files can be downloaded to the workdir but not uploaded except via `git` or upcoming `oci-push` support for uploading archives (#6626). Uploading from the workdir is useful for running tasks outside of Kargo which need to act on the workdir (e.g. hydration).

### Alternatives Considered

`oci-push` could be used as an alternative but that is rather heavyweight when an archive just needs to be sent to a service.

tomatotomata:
## Summary

- allow the `http` promotion step to use a request body from a path inside the promotion work directory
- preserve the existing inline `body` behavior and reject setting both body sources
- add path traversal protection, focused runner tests, generated schema updates, and reference documentation

Closes #6743

## Validation

- `git diff --check` passes
- JSON schema files parse successfully
- Go tests were not run because the Go toolchain is not installed in this environment

The branch is a draft while the API shape and CI results are reviewed. I chose the generic `bodyFromFile` field so it can support the existing tar step and other files without coupling the HTTP step to archives.


pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -299,17 +310,33 @@ func (h *httpRequester) extractErrorMessageFromResponse(
...
 func (h *httpRequester) buildRequest(
 	ctx context.Context,
+	stepCtx *promotion.StepContext,
 	cfg builtin.HTTPConfig,
 ) (*http.Request, error) {
 	method := cfg.Method
 	if method == "" {
 		method = http.MethodGet
 	}
+	body := cfg.Body
```
mhoran:
```suggestion
	body := bytes.NewBufferString(cfg.Body)
```

pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -299,17 +310,33 @@ func (h *httpRequester) extractErrorMessageFromResponse(
...
+		}
+		bodyPath, err := securejoin.SecureJoin(stepCtx.WorkDir, cfg.BodyFromFile)
+		if err != nil {
+			return nil, fmt.Errorf("could not secure join bodyFromFile %q: %w", cfg.BodyFromFile, err)
+		}
+		bodyBytes, err := os.ReadFile(bodyPath)
+		if err != nil {
+			return nil, fmt.Errorf("could not read bodyFromFile %q: %w", cfg.BodyFromFile, err)
+		}
+		body = string(bodyBytes)
```
mhoran:
```suggestion
		body = bytes.NewBuffer(bodyBytes)
```

pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -299,17 +310,33 @@ func (h *httpRequester) extractErrorMessageFromResponse(
...
+	}
 	req, err := http.NewRequestWithContext(
 		ctx,
 		method,
 		cfg.URL,
-		bytes.NewBufferString(cfg.Body),
+		bytes.NewBufferString(body),
```
mhoran:
```suggestion
		body,
```

pkg/promotion/runner/builtin/http_requester_test.go
```diff
@@ -766,6 +778,42 @@ func Test_httpRequester_buildRequest(t *testing.T) {
...
+			BodyFromFile: "payload.json",
+		},
+	)
+	require.NoError(t, err)
+	body, err := io.ReadAll(req.Body)
+	require.NoError(t, err)
+	require.Equal(t, `{"from":"file"}`, string(body))
+}
+
+func Test_httpRequester_buildRequest_bodyFromFile_rejectsTraversal(t *testing.T) {
```
mhoran: This test is failing.

pkg/promotion/runner/builtin/http_requester.go
mhoran: Don't convert `bodyBytes` to a `string` (it may be binary).

pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -84,15 +86,24 @@ func (h *httpRequester) Run(
 // convert validates httpRequester configuration against a JSON schema and
 // converts it into a builtin.HTTPConfig struct.
 func (h *httpRequester) convert(cfg promotion.Config) (builtin.HTTPConfig, error) {
-	return validateAndConvert[builtin.HTTPConfig](h.schemaLoader, cfg, stepKindHTTP)
+	_, hasBody := cfg["body"]
+	_, hasBodyFromFile := cfg["bodyFromFile"]
+	if hasBody && hasBodyFromFile {
+		return builtin.HTTPConfig{}, fmt.Errorf("body and bodyFromFile cannot be set together")
+	}
```
krancour: This sort of mutual exclusivity should be enforced by the schema.

tomatotomata:
I went through the review points and pushed `e3ef24a`.

- The body/bodyFromFile exclusivity is now represented in the JSON schema as well as the existing runtime guard.
- File-backed request bodies stay as bytes, so binary payloads are not converted through a string.
- The focused body-from-file test now asserts the exact byte sequence, including non-UTF-8 bytes.
- `git diff --check` passes. I still cannot run the Go tests here because the Go toolchain is not installed.

The PR is still ready for CI validation. Let me know what you think.

krancour:
> The body/bodyFromFile exclusivity is now represented in the JSON schema as well as the existing runtime guard.

All other built-in steps rely exclusively on the schema for validation. They do _test_ the schema validates things as expected using Go unit tests. Their _might_ be a case or two where there is very complex validation written in Go because it was beyond what can be expressed with JSON schema. But this isn't that.

Long story short... let's avoid the unnecessary deviation from the established pattern.

krancour:
@fuskovic can I ask you to take over reviews for this one?

pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -6,19 +6,22 @@ import (
...
 
+	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/expr-lang/expr"
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/xeipuuv/gojsonschema"
 	"sigs.k8s.io/yaml"
 
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
-	"github.com/akuity/kargo/pkg/io"
+	kargoio "github.com/akuity/kargo/pkg/io"
```
fuskovic: The`stdio` alias looks unnecessary here since the `kargoio` alias is already in place

pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -84,15 +87,24 @@ func (h *httpRequester) Run(
...
+	_, hasBody := cfg["body"]
+	_, hasBodyFromFile := cfg["bodyFromFile"]
+	if hasBody && hasBodyFromFile {
+		return builtin.HTTPConfig{}, fmt.Errorf("body and bodyFromFile cannot be set together")
+	}
+	converted, err := validateAndConvert[builtin.HTTPConfig](h.schemaLoader, cfg, stepKindHTTP)
+	if err != nil {
+		return builtin.HTTPConfig{}, err
+	}
+	return converted, nil
```
fuskovic: NIT: the following is more succinct and is what we're already doing in other places:

```suggestion
    return validateAndConvert[builtin.HTTPConfig](h.schemaLoader, cfg, stepKindHTTP)
```

pkg/promotion/runner/builtin/schemas/http-config.json
```diff
@@ -62,6 +62,13 @@
   "type": "object",
   "additionalProperties": false,
   "required": ["url"],
+  "allOf": [
+    {
+      "not": {
+        "required": ["body", "bodyFromFile"]
+      }
+    }
+  ],
```
fuskovic: The established pattern for mutual exclusivity is using `oneOf`. You can check the git push config for prior art

pkg/promotion/runner/builtin/http_requester.go
fuskovic: +1 @tomatotomata you should be able to delete this after you address https://github.com/akuity/kargo/pull/6753/changes#r3728784094

pkg/promotion/runner/builtin/http_requester.go
```diff
@@ -299,17 +311,33 @@ func (h *httpRequester) extractErrorMessageFromResponse(
...
+	var bodyReader stdio.Reader = strings.NewReader(cfg.Body)
+	if cfg.BodyFromFile != "" {
+		if stepCtx == nil {
+			return nil, fmt.Errorf("cannot read bodyFromFile %q without a step context", cfg.BodyFromFile)
+		}
+		bodyPath, err := securejoin.SecureJoin(stepCtx.WorkDir, cfg.BodyFromFile)
+		if err != nil {
+			return nil, fmt.Errorf("could not secure join bodyFromFile %q: %w", cfg.BodyFromFile, err)
+		}
+		bodyBytes, err := os.ReadFile(bodyPath)
```
fuskovic: Limit reading here should be safer

```suggestion
		bodyBytes, err := kargoio.LimitRead(bodyPath, maxResponseBytes)
```

fuskovic (changes requested):

tomatotomata:
I followed up on the latest review in `03075c8`:

- moved the `body`/`bodyFromFile` exclusivity rule fully into the HTTP schema using the project’s `oneOf` pattern, then removed the duplicate runtime guard;
- removed the redundant `stdio` alias and used the standard `io.Reader` directly;
- bounded `bodyFromFile` reads with `kargoio.LimitRead` and added an oversize regression test;
- kept the existing binary-body coverage and updated the validation expectation for the schema-owned error;
- simplified `convert` to the repository’s usual `validateAndConvert` form.

`gofmt`, `git diff --check`, and JSON parsing pass. The focused Go test is currently blocked on this Windows checkout by the existing `gitkit` syscall incompatibility (`Setpgid` and `syscall.Kill`), not by the changed package. Let me know what you think.

fuskovic:
@tomatotomata you need to run `make hack-codegen` to fix the required `check-codegen` CI check.

You also need to fix this failing test:

```
=== RUN   Test_httpRequester_buildRequest_bodyFromFile_rejectsTraversal
    http_requester_test.go:836: 
        	Error Trace:	/__w/kargo/kargo/pkg/promotion/runner/builtin/http_requester_test.go:836
        	Error:      	Error "could not read bodyFromFile \"../payload.json\": open /tmp/Test_httpRequester_buildRequest_bodyFromFile_rejectsTraversal590914904/001/payload.json: no such file or directory" does not contain "could not secure join bodyFromFile"
        	Test:       	Test_httpRequester_buildRequest_bodyFromFile_rejectsTraversal
--- FAIL: Test_httpRequester_buildRequest_bodyFromFile_rejectsTraversal (0.00s)
```

tomatotomata:
I pushed `fe4e285` for these two failures.

The generated Go config and UI directive schema now come from the pinned project generators. For the traversal case, I changed the test to put a readable payload just outside the work directory, then verify `../payload.json` still cannot read it. That matches `securejoin`'s containment behavior instead of expecting it to reject the path itself.

The Linux test package cross-compiles successfully. I still cannot execute that package on this Windows checkout because the existing `gitkit` dependency uses Unix-only syscall fields, so the new CI run will provide the Linux execution result.

tomatotomata:
I followed up on the two CI items from the review against the current head `2815523`.

- The body-file traversal test now matches the repository's `securejoin` behavior: traversal stays under the work directory, so the expected failure is the missing in-workdir file (`could not read bodyFromFile`), not a secure-join rejection. The test is named `Test_httpRequester_buildRequest_bodyFromFile_keepsTraversalInsideWorkDir`.
- I ran the promotion-schema generator equivalent to `make hack-codegen` on Windows. It produced no tracked diff, and `git diff --exit-code -- .` plus `git diff --check` both pass.

The focused Go test is still blocked before test execution by the existing Windows incompatibility in `github.com/sosedoff/gitkit@v0.4.0` (`Setpgid` and `syscall.Kill`), so I am not calling it green from this checkout. The current head is pushed and ready for the Linux CI run. Let me know what you think.

pkg/promotion/runner/builtin/http_requester_test.go
```diff
@@ -766,6 +779,72 @@ func Test_httpRequester_buildRequest(t *testing.T) {
...
+	_, err := (&httpRequester{}).buildRequest(
+		t.Context(),
+		&promotion.StepContext{WorkDir: workDir},
+		builtin.HTTPConfig{
+			URL:          "http://example.com",
+			BodyFromFile: "../payload.json",
+		},
+	)
+	require.ErrorContains(t, err, "could not read bodyFromFile")
+}
```
fuskovic, line 863: NIT: Can you refactor these to use the table-test pattern please?

fuskovic (changes requested):
One small nit and then this should be good to go

pkg/promotion/runner/builtin/http_requester_test.go
fuskovic: All of these should be sub-tests of `Test_httpRequester_buildRequest` ( including the existing logic for `Test_httpRequester_buildRequest` )

tomatotomata:
I regrouped the request-building coverage into one table-driven `Test_httpRequester_buildRequest`, including the original request option assertions and the three body-file cases as sub-tests. The cases still cover binary bytes, the size limit, and work-directory traversal.

The focused Go test command is currently blocked by the Windows toolchain: the transitive `gitkit` dependency uses Unix-only `syscall.SysProcAttr.Setpgid` and `syscall.Kill`. `gofmt` and `git diff --check` pass, and the refactor is pushed in `dbdf0c7`. Let me know what you think.

tomatotomata:
I folded the two test-shape requests into `0b22183`: the existing request-options case now lives in the same table, and each body-file case is a sub-test under `Test_httpRequester_buildRequest`.

The new commit also carries the authorized `Signed-off-by: ahmad <ahmadalgaidy@hotmail.com>`. GitHub's DCO, title, security, and preview checks pass.

The focused Go command is still blocked on this Windows checkout before tests start because `github.com/sosedoff/gitkit` uses Unix-only `syscall.Setpgid` and `syscall.Kill`. I recorded that as an environment limitation, not a test pass. Do you think this matches the pattern you had in mind?

fuskovic:
@tomatotomata lint check is failing:

```
pkg/promotion/runner/builtin/http_requester_test.go:770:16: unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _ (revive)
			setup: func(t *testing.T) (*promotion.StepContext, builtin.HTTPConfig) {
			            ^
1 issues:
* revive: 1
make: *** [Makefile:80: lint-go] Error 1
```

You can run it locally using `make hack-lint-go`.

tomatotomata:
I fixed the lint blocker in the first request-options case by naming the unused setup parameter _, then pushed 5f82e79.\n\nThe change is intentionally one line and keeps the table-driven test structure unchanged. gofmt and git diff --check pass. I also tried the focused Go test, but this Windows environment still stops in the existing gitkit dependency (syscall.Setpgid and syscall.Kill are unavailable here), before compiling the Kargo package. The CI matrix should provide the runtime validation.

pkg/promotion/runner/builtin/http_requester_test.go
```diff
@@ -199,6 +203,16 @@ func Test_httpRequester_convert(t *testing.T) {
 				"proxy: Does not match pattern",
 			},
 		},
+		{
+			name: "body and bodyFromFile are mutually exclusive",
+			config: promotion.Config{
+				"body":         "",
```
fuskovic: Shouldn't `body` have a value here?

pkg/promotion/runner/builtin/schemas/http-config.json
```diff
@@ -62,6 +62,28 @@
...
+    },
+    {
+      "not": {
+        "anyOf": [
+          {"required": ["body"]},
+          {"required": ["bodyFromFile"]}
+        ]
+      }
+    }
+  ],
```
fuskovic, line 100: This is still not totally consistent with our established pattern.  Look at the git push config schema again for mutual exclusivity examples. We don't use `not` this way.

fuskovic (changes requested):
Warmer. 

At least 4 times when you've commented to let us know suggested changes have been pushed, you've mentioned:

> I also tried the focused Go test, but this Windows environment still stops in the existing gitkit dependency (syscall.Setpgid and syscall.Kill are unavailable here), before compiling the Kargo package.

You don't need to repeat that every time. It looks obvious the comments are AI generated and that's ok but at least check them before posting so you have an opportunity to clean them up you can remove things like literal new line characters and repetitive information.

tomatotomata:
I was thinking the empty string should behave like an omitted body, so I kept that case valid while removing the `not` branches. The schema now follows the git push pattern with property constraints for `body` and `bodyFromFile`, and the mutual-exclusion test uses a real body value so it exercises the conflict. I also fixed the unused test parameter in `84941ab`; the pushed commit has a green PR check set. Let me know what you think.

tomatotomata:
You are right about the repetition. I have removed the environment note from the follow-up and will only include it again if it changes the decision.

The current head remains `84941ab`; the lint fix and schema/test updates are pushed, and the repository checks shown on the PR are green. Let me know if the remaining test shape needs another adjustment.

fuskovic (approved):

PR #6754, opened 2026-08-03
fix: kargo cli tokens not setting ctx.meta.promotion.actor

### LINKED ISSUE #6201: Promotions from kargo CLI using tokens should set ctx.meta.promotion.actor

fuskovic:
Closes: #6201
Supersedes: #6406

#6406 attributed the actor to the token itself, parsing an unverified JWT claim (the Secret name given to `kargo create token --name=...`).This conflated the credential with the principal and required parsing JWT claims before verification (flagged by CodeQL).

This PR submits the token to K8s via a `TokenReview` instead. That verifies the token and returns its K8s-authenticated identity (`system:serviceaccount:<namespace>:<role>`) in one call, with no unverified claim parsing. The identity is formatted with the same helper already used for promotions created directly against the K8s API, so the actor is now consistent regardless of which path authenticated the request.

Built on top of #6757's error-handling fix: a failed `TokenReview` call now reports as a 500, while K8s rejecting the token reports as a 401, rather than conflating the two.


pkg/api/event.go
```diff
@@ -16,14 +16,20 @@ func FormatEventControllerActor(name string) string {
...
-// 2. If the email is available, it returns email in "email:<email>" format.
-// 3. If the subject is available, it returns subject in "subject:<subject>" format.
-// 4. Otherwise, it returns EventActorUnknown.
+//  1. If the user is admin, it returns EventActorAdmin since other information is not available.
+//  2. If the user was authenticated by Kubernetes directly (e.g. a Kargo API token),
+//     it returns that identity in "kubernetes:<username>" format (see
+//     FormatEventKubernetesUserActor).
+//  3. If the email is available, it returns email in "email:<email>" format.
+//  4. If the subject is available, it returns subject in "subject:<subject>" format.
+//  5. Otherwise, it returns EventActorUnknown.
```
krancour: Small, existing issue that we can fix easily right now and should. The original code had five cases, but the comment enumerated only four. It missed the one where username was known. The revised comment is still overlooking that case, acknowledging five when there are actually six.

pkg/server/auth_middleware.go
```diff
@@ -383,45 +397,28 @@ func (a *authMiddleware) verifyKargoIssuedToken(rawToken string) bool {
...
+			Token: rawToken,
+		},
 	}
-
-	apiURL := strings.TrimSuffix(a.cfg.RestConfig.Host, "/") + "/api"
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, apiURL, nil)
-	if err != nil {
-		return fmt.Errorf("create request: %w", err)
+	if err := a.internalClient.Create(ctx, review); err != nil {
+		return nil, fmt.Errorf("%w: %w", errKubernetesTokenReviewFailed, err)
```
krancour: I see what you're doing here. You're returning a sentinel the caller can look for so it can choose whether to log this error more loudly than others. That's not wrong, but it points to a bad smell in the existing code, which was probably my own fault to begin with, but I'm afraid this is a half-measure that ends up making the situation worse instead of better.

The existing, bad smell is that authentication middleware does a poor job of distinguishing between missing/invalid tokens and incidental errors that should be treated as 500s _instead_ of 401s. (Again, my fault.) If _that_ weren't already a problem, you wouldn't need this sentinel. Returning an error type that clearly reflects an issue with the token would result in a debug message only, while returning any other error would take the "I don't recognize this error" path in the error-handling middleware and log that as an error before returning a 500.

I started looking at what it would take to fix that and while doing so uncovered a few other bugs in our error-handling middleware. #6757 corrects those bugs and very slightly updates this authentication middleware to be clear about different types of errors in ways that the (fixed) error-handling middleware will respond to correctly.

If you'd be willing to review #6757 and if we could get that merged first, I believe that rebasing this PR and resolving a small number of merge conflicts will leave this in a better place.

pkg/api/event.go
krancour, line 44: Just something I noticed while reviewing this PR. This should be an edge case. It would mean a verified token didn't include the claim that Kargo was configured to treat as "username." It's possible, but unlikely. While we're in here, adding a comment to this effect might be worthwhile because I spend way too long scratching my head about whether this could actually happen.

krancour, line 53: Similar to above, but to the extreme. This would be an extraordinary edge case involving a verified token that didn't include the claim that Kargo was configured to treat as "username" _and_ also didn't include an "email" claim, despite the fact the Kargo unconditionally requests the email scope from the OIDC. (There's nothing that guarantees the IdP will actually send all the claims associated with requested scopes.) "sub" is the backstop for this ultra-edge case because the "sub" claim always is included.

pkg/server/auth_middleware.go
```diff
@@ -383,45 +397,28 @@ func (a *authMiddleware) verifyKargoIssuedToken(rawToken string) bool {
...
-// provided token by making a GET request to the /api endpoint. This is a
-// lightweight check that doesn't require any specific permissions.
+// verifyKubernetesToken submits the token to Kubernetes via a TokenReview
+// and returns the verified identity. Requires the kargo-api ClusterRole to
+// grant permission to create TokenReviews.
 func (a *authMiddleware) verifyKubernetesToken(
 	ctx context.Context,
 	rawToken string,
-) error {
-	if a.cfg.RestConfig == nil { // This shouldn't happen, but just in case...
```
krancour, line 425: With the `/api` request gone, nothing reads `ServerConfig.RestConfig` anymore -- `cmd/controlplane/api.go:86` is the only remaining reference and it only assigns it. Mind dropping the field and that assignment while you're in here?

krancour:
The `kubernetes:` actor format contains four colons, but [`CreateActorAnnotationValue`](https://github.com/akuity/kargo/blob/09226666d8f3a21a499356af81fa13dd017927ba/pkg/api/annotations.go#L152-L167) strips a prefix only when the value splits into exactly two parts, so the prefix survives into `ctx.meta.promotion.actor` this contradicts how the UI is parsing the annotation value.

Fix belongs here: https://github.com/akuity/kargo/blob/09226666d8f3a21a499356af81fa13dd017927ba/pkg/api/annotations.go#L152-L167

Switching to `strings.SplitN(v, ":", 2)` fixes it. A new test case to cover this would be good, too.

pkg/server/auth_middleware.go
```diff
@@ -383,45 +397,28 @@ func (a *authMiddleware) verifyKargoIssuedToken(rawToken string) bool {
...
-
-	transport, err := rest.TransportFor(a.cfg.RestConfig)
-	if err != nil {
-		return fmt.Errorf("create transport: %w", err)
+) (*authnv1.UserInfo, error) {
+	review := &authnv1.TokenReview{
+		Spec: authnv1.TokenReviewSpec{
+			Token: rawToken,
+		},
 	}
```
krancour, line 434: This approach is _way_ better than what was here before. 👏

ui/src/features/stage/promotions.tsx
```diff
@@ -111,9 +111,14 @@ export const Promotions = ({ argocdShard }: { argocdShard?: ArgoCDShard }) => {
...
-        const email = annotation ? annotation.split(':')[1] : 'N/A';
 
-        return email || annotation;
+        if (!annotation) {
+          return 'N/A';
+        }
+
+        const separatorIndex = annotation.indexOf(':');
+
+        return separatorIndex === -1 ? annotation : annotation.slice(separatorIndex + 1);
```
krancour: This duplicates the change in `get-promotion-actor.ts`.

The duplication of parsing logic is _not_ new in this PR, but idk if needing to make the same change twice here points to it being time to fix that.

My concerns are mainly over the back end, so this isn't a hill I'd die on, but it might be worth checking in with @rpelczar on this.

fuskovic: @rpelczar I can add a `resolveCreatorByAnnotation` helper here to keep things DRY if you'd like

pkg/server/auth_middleware.go
fuskovic: I've approved https://github.com/akuity/kargo/pull/6757. Will mark this resolved once that gets merged and I rebase here.

krancour: Thank you. Merged.

ui/src/features/stage/promotions.tsx
rpelczar: @fuskovic ok feel free to do that

rpelczar (approved):

krancour (approved):

PR #6755, opened 2026-08-03
fix: yaml-merge silently dropping YAML aliases and null-valued keys  

### LINKED ISSUE #6740: yaml-merge silently drops YAML aliases and null-valued keys
### Pre-submission checklist

### Description

yaml-merge round-trips the entire document through sigs.k8s.io/kustomize/kyaml/yaml/merge2 (pkg/yaml/merge.go:10-17). Alias nodes and explicitly-null keys don't survive the round-trip. There's no error and no warning — the step reports success and the data is gone.

This is surprising given the sibling step: yaml-update is lossless, because pkg/yaml/update.go does node-level surgery with gopkg.in/yaml.v3 and handles yaml.AliasNode explicitly (line 50). Merging two files is strictly more destructive than updating one.

### Steps to Reproduce

Repro

values.yaml:
```yaml
db:
  name: &db-name mydb
pgbouncer:
  readinessDbName: *db-name
  nodeSelector:
defaults: &defaults
  image: my-image
other:
  <<: *defaults
```

overlay.yaml:

```yaml
someFlag: false
```

kargo.yaml

```yaml
- uses: yaml-merge
  config:
    inFiles: [overlay.yaml, values.yaml]
    outFile: values.yaml
```

### Expected Behavior

Either the aliases are resolved to their values, or they're preserved as-is. Silently dropping the keys is the one outcome that can't be right. Failing the step on an unsupported node kind would also be acceptable — anything but silent data loss.

### Actual Behavior

Three distinct losses (each observed on real manifests; reduced to the above for clarity):

1. pgbouncer.readinessDbName: *db-name — removed entirely. Any key whose value is an alias disappears.
2. pgbouncer.nodeSelector: (explicit null) — removed, leaving pgbouncer: {}. Explicit nulls are meaningful in Helm; they clear a value inherited from an earlier values file.
3. other: {<<: *defaults} — merge key not expanded and not preserved, so the whole inherited block is lost.

### Kargo Version

v1.11.0

### Relevant Logs

```shell

```

### Screenshots

_No response_

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6740

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/yaml-merge.md
@@ -23,8 +23,12 @@ Merging is performed as follows:
   key, the sequence from the second document replaces the first (no merging).
 - **Keys present only in one document:** Keys that exist in only one document
   are included as-is in the result.
-- **Null values:** If a key is set to `null` in the second document, it removes
-  or overrides the value from the first document.
+- **Null values:** If a key is set to `null` in the second document, the
+  result has that key set to `null` as well -- it is not removed. `null` is
+  treated like any other scalar value.
+- **Anchors, aliases, and merge keys:** YAML anchors (`&name`), aliases
+  (`*name`), and the `<<` merge key are preserved as-is. They are never
+  resolved or expanded by this step.
 
 :::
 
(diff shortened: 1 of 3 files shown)


krancour (changes requested):
Let's put a pin in this and take the discussion back to #6740.

Two separate concerns.

The null half is a breaking change to how `null` is handled — today it removes
a key, and after this it wouldn't. Anyone relying on that in an existing
promotion would silently get a different file committed. We'd need either to
avoid that or have a very good justification plus a migration plan, and I'd
like to settle whether it's a bug at all first.

The alias handling also looks buggy — I was able to get it to produce YAML that
doesn't parse. Happy to get into specifics once we've settled the larger
question.

PR #6757, opened 2026-08-04
fix(server): correct REST API error handling

krancour:
Three distinct but related changes to the REST API's error handling, one per commit.

1. **`handleError` now writes exactly one response for every error reported to it.** It previously wrote none in one case and two in another.

    1. When an error matched none of its known types, it re-reported the error instead of answering it. Nothing ever reads that: `handleError` inspects `c.Errors` on its way back out of the middleware chain, so it is the last code to look, and appending to that slice afterward has no effect. Gin then wrote its default 200 and ended the request. A handler reporting an untyped error answered as though it had succeeded.

    1. An error carrying a 500 was answered twice, the second write disclosing the underlying message the first had deliberately obfuscated.

1. **The authentication middleware no longer reports server failures as rejected credentials.** It aborted with a 401 for every error, so an unreachable Kubernetes API server, an identity provider that could not be reached at startup, and a username claim of the wrong type all reached the client as a bad credential. The dashboard signs a user out on any 401, so a control plane problem presented as "your credentials are bad" and logged everyone out, with the real cause visible only at debug level.

    It now returns errors carrying a status code and lets `handleError` answer. A 401 is opt-in and reserved for failures attributable to the credential; anything else is reported as an internal error and logged at error level. Clients also get a body on these responses now, where a 401 previously had none.

1. **The error envelope type is named for what it is.** `handleError` returns it for every error from every endpoint, not only for the resource endpoints it was named after. Its published schema name is preserved, so `swagger.json` and every generated client are byte-for-byte unchanged.

### DIFF
--- a/pkg/server/auth_middleware.go
+++ b/pkg/server/auth_middleware.go
@@ -25,6 +25,7 @@ import (
 	libClient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	libhttp "github.com/akuity/kargo/pkg/http"
 	"github.com/akuity/kargo/pkg/indexer"
 	"github.com/akuity/kargo/pkg/logging"
 	"github.com/akuity/kargo/pkg/server/config"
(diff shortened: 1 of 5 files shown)


fuskovic (approved):
LGTM

PR #6758, opened 2026-08-04
fix(ui): Cannot view AnalysisRun result in UI when web provider returns text instead of JSON

### LINKED ISSUE #4630: Cannot view AnalysisRun result in UI - not a valid JSON
# Checklist

# Description
I have a very simple AnalysisTemplate that uses `web` provider. It pings an URL that returns plain/text. I know Rollouts docs say the webhook must return JSON, but it actually does work just fine with plan text.

However, when clicking on the AnalysisRun link under Stage's Verifications tab in Kargo UI, it fails with this error in the console log:
```
SyntaxError: Unexpected token 'P', "PASS" is not valid JSON
```

This is my AnalysisTemplate:
```yaml
kind: AnalysisTemplate
metadata:
  name: healthcheck
spec:
  args:
    - name: url
  metrics:
    - name: diags
      count: 4
      failureLimit: 2
      initialDelay: 30s
      interval: 30s
      successCondition: result == "PASS"
      provider:
        web:
          url: "{{ args.url }}"
```

The webhook is test simply returns `PASS` as text. Unfortunately, I cannot change the output.
It would be great if Kargo ignored the requirement for the output to be JSON and supported any string.

# Version

1.6.1

jmurret:
## Description

Closes #4630

This PR represents the UI extension of https://github.com/argoproj/argo-rollouts/pull/4770 that addressed:
>Fixes a bug where the web metric provider always returned Successful for empty or non-JSON HTTP response bodies, without evaluating successCondition / failureCondition.

text was always supported but the `successCondition` was ignored and always returned as successful.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

- [x] By a human _without_ AI assistance.

### Sign-Off

All commits:

### DIFF
--- a/ui/src/features/common/analysis-modal/transforms.ts
+++ b/ui/src/features/common/analysis-modal/transforms.ts
@@ -783,7 +783,22 @@ const transformMeasurementValue = (
     };
   }
 
-  const parsedValue = JSON.parse(value);
+  let parsedValue;
+  try {
+    parsedValue = JSON.parse(value);
+  } catch {
+    // this behavior matches rollouts with the following reasoning:
+    // - canChart == true - just one measurement with this set to false will prevent the chart from being displayed
+    // - chartValue == null - the chart will just show an absence of a point or a blip for this measurement
+    // - tableValue == null - while a healthcheck of "PASS" or "OK" might be worth seeing, allowing a string of
+    // onbounded length could make the table unusable (a stack trace, error message, a failed parsoning because 
+    // a long json string got truncated, etc.)
+    return {
+      canChart: true,
+      chartValue: null,
+      tableValue: null
+    };
+  }
 
   // single number measurement value
   if (isFiniteNumber(parsedValue)) {
--- a/ui/src/features/common/analysis-modal/transforms.ts
+++ b/ui/src/features/common/analysis-modal/transforms.ts
@@ -783,7 +783,15 @@ const transformMeasurementValue = (
     };
   }
 
-  const parsedValue = JSON.parse(value);
+  let parsedValue: unknown;
+  try {
+    parsedValue = JSON.parse(value);
+  } catch {
+    return {
+      canChart: false,
+      tableValue: value
+    };
+  }
 
   // single number measurement value
   if (isFiniteNumber(parsedValue)) {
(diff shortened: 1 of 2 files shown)


Marvin9:
On the surface from what you shared upstream, it looks like instead of failing, if it returns text, we would want to parse it as text and compare against?

This PR silently drops it. I didn't look in depth but I will

hdost:
Perhaps https://github.com/akuity/kargo/pull/6799 looks like it's trying to fix the same bug

ui/src/features/common/analysis-modal/transforms.ts
```diff
@@ -783,7 +783,15 @@ const transformMeasurementValue = (
...
   }
 
-  const parsedValue = JSON.parse(value);
+  let parsedValue: unknown;
+  try {
+    parsedValue = JSON.parse(value);
+  } catch {
+    return {
+      canChart: false,
+      tableValue: value
```
Marvin9: Ok I finally got a chance to take a look at this issue. 

https://github.com/argoproj/argo-rollouts/blob/master/ui/src/app/components/analysis-modal/transforms.ts#L628

upstream discards it, I don't know motivation behind it but does it make sense NOT TO discard it. Maybe its also better to show something then nothing?

jmurret: @Marvin9 thank you for pointing out the upstream.  I've added some reasoning in a comment of why this should match the behavior in rollouts.  let me know what you think, but it is basically:
- one failed measurement shouldn't make cause the whole chart to not be rendered
- allowing any string of any length to be displayed in a table is risky

Marvin9:
@jmurret could you please sign and amend all commits in this PR? And fix the lint issue? 

Ready to merge after that

jmurret:
Thanks @Marvin9 .  I consolidated to one signed commit.  also the lint error was fixed in the last revision and `make type-checkui` runs locally without error, so should be 👌 when you run CI.

REVISED: I saw the `lint-ui` after and added this commit to fix those 2 issues: 5cbc24797132a25da7cecfd558b15da9c77eca67

Marvin9:
Sorry @jmurret DCO is failing, you would need to amend signed commits

jmurret:
@Marvin9 So sorry about that.  This is resolved and DCO is passing.

Marvin9 (approved):

PR #6764, opened 2026-08-04
refactor(server): authorize by identity rather than by replaying the token

krancour:
Follow-up to #6754, which taught the authentication middleware to submit a `TokenReview` and keep the identity Kubernetes resolved a bearer token to. With that identity on hand, the token itself has no remaining job.

Authorization for these users was a `SelfSubjectAccessReview` submitted *as them*, which is the only reason the API server retained their bearer token at all. A `SubjectAccessReview` naming the identity answers the same question, so `user.Info` no longer carries a credential. The identity is described in full — groups, UID and extras — because a `SubjectAccessReview` specifying a user without groups asks whether that user is permitted on the strength of their username alone, and group bindings are the ordinary way a cluster's identity provider grants access.

Calling this a security improvement would be generous. A credential no longer sits in a context-bound struct for the life of a request, which is one fewer place it exists, but Go is memory-safe and the token was never transmitted anywhere except back to the API server that issued the review. Take it as tidiness with a mild security flavor.

The simplification is the more tangible part. Nothing needs a per-request Kubernetes client anymore, so `reviewSubjectAccess` submits the review with the API server's own client — which retires the `rest.Config` construction, the credential-blanking with its "are there more things to blank out here?" TODO, the `SelfSubjectAccessReview` branch, and the options plumbing that existed only to choose between a token and a subject. That also makes this authorization path unit-testable for the first time, since it no longer reads a kubeconfig from the environment.

Removing the field takes one feature with it: when no token `Secret` was configured, `${{ token }}` in an AnalysisRun log request header resolved to the bearer token the user presented to Kargo. That credential cannot be validated by an arbitrary HTTP log service, which is the only kind this feature retrieves from — logs are aggregated outside the cluster precisely because the control plane cannot serve them. The behavior was never documented and never useful.

Verified end to end against a live cluster with API tokens, including that permissions are still enforced exactly as before: granted verbs succeed, ungranted verbs and other projects are refused.

### DIFF
--- a/pkg/server/auth_middleware.go
+++ b/pkg/server/auth_middleware.go
@@ -210,10 +210,7 @@ func (a *authMiddleware) authenticate(
 			logger.Debug("admin token verified as issued by Kargo API server")
 			return user.ContextWithInfo(
 				ctx,
-				user.Info{
-					IsAdmin:     true,
-					BearerToken: rawToken,
-				},
+				user.Info{IsAdmin: true},
 			), nil
 		}
 		return ctx, errInvalidToken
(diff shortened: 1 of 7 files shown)


fuskovic (approved):
LGTM

PR #6766, opened 2026-08-05
feat: populate the Argo CD context annotation from promotion step output

### LINKED ISSUE #6762: argocd-wait: report resolved Applications in step output for the Argo CD context annotation
## Pre-submission checklist

## Proposed Feature

Have `argocd-wait` report the Argo CD `Application`s it resolved as step output:

```yaml
apps:
- name: guestbook
  namespace: argocd
```

and have `api.AnnotateStageWithArgoCDContext()` read that output in addition to
the health check criteria it already consults, so the
`kargo.akuity.io/argocd-context` annotation is populated for promotion processes
that use `argocd-wait`.

`argocd-wait` would still not emit a `health.Criteria` — the step already
assesses health while the `Promotion` runs, and Stage health semantics would be
unchanged.

## Motivation and Use Case

The annotation exists to give the UI deep links to Argo CD from the Stage view
(#3570). It is derived only from the health check criteria registered by
`argocd-update`.

`argocd-wait` (#5400) is for pipelines where the sync is triggered by something
else — Argo CD auto-sync, an external system, an operator. Those never register
an Argo CD health check, so their Stages get no deep links, even though the step
resolved the exact `Application`s moments earlier. And because the annotation is
removed when nothing contributes apps, a Stage that had working links loses them
after switching from `argocd-update` to `argocd-wait`.

The step already knows the answer; it just isn't reported anywhere the Stage
controller can see it.

Marvin9:
## Summary

The `kargo.akuity.io/argocd-context` annotation, which the UI uses to build deep links into Argo CD from the Stage view, was derived solely from the health check criteria registered by `argocd-update`. `argocd-wait` resolves Argo CD `Application`s but registers no health check, so Stages promoted with it got no deep links — and a Stage that switched from `argocd-update` to `argocd-wait` lost the links it already had, because the annotation is removed when nothing contributes apps.

Both steps now report the `Application`s they resolved as step output, and the annotation is built from that.

### How it works

`ArgoCDAppRef` is the single shape every contributor is reduced to. Two transformers feed it, and the results are merged and deduplicated:

- **Step output** — `argocd-update` and `argocd-wait` both report the `Application`s they resolved under an `apps` output key. Refs are built from the resolved `Application` objects rather than from step config, so the namespace is always concrete even when apps were selected by label selector.
- **Health check criteria** — the existing behavior, retained for backward compatibility (see below).

Deduplication is by `namespace/name`. It matters because in the normal case a single `Application` is reported by up to three contributors: an `argocd-update` step's output, the health check that same step registers, and a subsequent `argocd-wait` step targeting the same app.

`argocd-wait` still registers no `health.Criteria`. It assesses health only while the `Promotion` runs, and Stage health semantics are unchanged.

### Backward compatibility

Steps are not re-executed across reconciliations, so a `Promotion` whose `argocd-update` step completed before an upgrade to this version finishes after it with health checks but no `apps` output. Dropping the health-check transformer would silently lose that Stage's links until its next promotion, so it is kept. It can be removed a release or two later.

### Note on decoding

Everything read out of step output and health check criteria is untyped. Extraction is written so that a type assertion that does not hold contributes nothing rather than panicking, which would take down the Stage controller. This is covered by 38 test cases that break the decode chain at every link.

The reverse direction is equally hazardous and cost us a live crash during manual testing (see below): step output becomes shared state, which the engine deep-copies with `runtime.DeepCopyJSON` before every subsequent step. That panics on anything outside `bool/int64/float64/string/[]any/map[string]any/nil`. `ArgoCDAppRefsToOutput()` flattens refs to the maps they would have been decoded into, and both runners' `run` tests now assert that every case's output can serve as shared state.

### Automated

```bash
make lint-go        # 0 issues
make test-unit
```

New unit coverage:

- `pkg/api/argocd_context_test.go` — both transformers against malformed, absent, and wrongly-typed input; non-Argo-CD steps emitting their own `apps` key are excluded; dedupe ordering and nameless-ref dropping; output survives `DeepCopyJSON` and round-trips back through the extractor unchanged.
- `pkg/api/stage_test.go` — table-driven: health-checks-only (the backward-compat path), step-output-only, a `task-1::wait` alias, cross-source dedupe, unparsable `status.state`, nil `Promotion`.
- Both runners — output assertions, plus a `requireJSONNativeOutput` guard in each `run` test loop.

### Manual test matrix

Run against a Tilt dev environment with three Stages in one Project, each promoting independently: one using `argocd-update`, one using `argocd-wait` (against an auto-synced `Application`, so Kargo never triggers the sync), and one using both against the same `Application`.

| Stage | Before | After |
|---|---|---|
| `argocd-update` only | annotation present | identical |
| `argocd-wait` only | **no annotation** | annotation present, deep links work |
| both, same app | present (health check) | present, exactly one entry |

The matrix caught a crash the unit tests did not: emitting `[]ArgoCDAppRef` directly panicked the Promotion controller with `cannot deep copy []api.ArgoCDAppRef` as soon as a second step ran. Only the two-step Stage reproduced it — a step's own output is never deep-copied during its own iteration, and state is reseeded from JSON-decoded `status.state` at the start of every reconciliation, so the struct never survives a reconciliation boundary. Fixed, and guarded by the test described above.

The backward-compatibility leg was exercised separately by starting a promotion on `main`, letting the `argocd-update` step complete, then swapping in the branch binary: the annotation is still produced, from health checks alone.

## Follow-ups

Not addressed here, both pre-existing:

- The "step output must be JSON-native" invariant is unwritten and unenforced. Nothing in `StepResult` or the runner interface signals it, and no shared guard exists. Every other built-in step hand-rolls `map[string]any` and happens to be safe — but `DeepCopyJSON` rejects plain `int`, so a step emitting one would panic identically. The `requireJSONNativeOutput` helper added here is trivially promotable into a shared test util.
- `health.Input.DeepCopy()` and `promotion.Context.DeepCopy()` have no non-test callers today, but both would panic the same way if called mid-execution: `argocd-update` puts `[]checkers.ArgoCDAppHealthCheck` structs into `health.Input`, and `Context.DeepCopy()` calls `State.DeepCopy()`.

Closes #6762

### DIFF
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/argocd-update.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/argocd-update.md
@@ -149,6 +149,21 @@ refresh groups of `Application`s with heterogeneous configurations.
 | `apps[].sources[].helm.images[].key` | `string` | Y | The key to update within the target `ApplicationSource`'s `helm.parameters` map. See Helm documentation on the [format and limitations](https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set) of the notation used in this field. |
 | `apps[].sources[].helm.images[].value` | `string` | Y | Specifies the new value for the key. Typically, a value from [`chartFrom()`](../40-expressions.md#chartfrom) is used here. |
 
+## Output
+
+| Name | Type | Description |
+|------|------|-------------|
+| `apps` | `[]object` | The Argo CD `Application` resources the step resolved and updated. Kargo uses this to link the target `Stage` to those `Application`s in the UI. |
+| `apps[].name` | `string` | The name of the Argo CD `Application`. |
+| `apps[].namespace` | `string` | The namespace of the Argo CD `Application`. |
+
+:::note
+
+When `apps` are selected by label selector, this output is the concrete list of
+`Application`s the selector matched.
+
+:::
+
 ## Health Checks
 
 The `argocd-update` step is unique among all other built-in promotion steps in
--- a/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/argocd-wait.md
+++ b/docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/argocd-wait.md
@@ -81,6 +81,25 @@ during the wait.
 | `apps[].selector.matchExpressions[].values` | `[]string` | N | An array of string values. Required when `operator` is `In` or `NotIn`. Must be empty when `operator` is `Exists` or `DoesNotExist`. |
 | `apps[].waitFor` | `[]string` | N | Conditions to wait for. Valid values: `health`, `sync`, `operation`, `suspended`, `degraded`. Defaults to `[health, sync, operation]` when omitted. |
 
+## Output
+
+| Name | Type | Description |
+|------|------|-------------|
+| `apps` | `[]object` | The Argo CD `Application` resources the step resolved and waited for. Kargo uses this to link the target `Stage` to those `Application`s in the UI. |
+| `apps[].name` | `string` | The name of the Argo CD `Application`. |
+| `apps[].namespace` | `string` | The namespace of the Argo CD `Application`. |
+
+:::note
+
+Unlike [`argocd-update`](argocd-update.md), this step registers no health checks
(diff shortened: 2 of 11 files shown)


fuskovic (approved):
LGTM

PR #6767, opened 2026-08-05
refactor(server): log requests and recover from panics with our own middleware

krancour:
`gin.Default()` installs Gin's logger, which writes a line per request to stdout in its own format and ignores `LOG_LEVEL` entirely. Requests are now recorded through Kargo's logger, with `method`, `path`, `status` and `duration` as fields.

Levels are chosen so that a stock installation, which runs at `INFO`, sees what matters and nothing else. A refused request — a `401` or a `403` — is recorded at `info`, because a refusal is never routine and is the first thing anyone looks for when a user cannot do something. Server errors are recorded at `error`. Everything else, including a request for something absent, is `debug`: the dashboard asks for resources that may not exist yet and treats their absence as an answer.

Errors are logged without a stack trace, via a new `Logger.WithoutStackTraces`. A trace describes the path to the logging call, so it is worth having only where an error is logged near where it arose, which a middleware sitting above every handler never is.

Dropping Gin's logger means dropping `gin.Default()`, which takes Gin's panic recovery with it. Rather than reinstate `gin.Recovery()`, which answers a panic by aborting with an empty-bodied `500`, recovery now reports the panic and its stack as an error and lets the error-handling middleware answer it. A panic therefore gets the same response body as any other unanticipated error. That also leaves the error-handling middleware as the only thing writing an error response, so it no longer logs one itself.

Gin is finally put in release mode, which stops it writing a line per registered route, and a warning about the mode, at startup. Nothing else Gin gates on the mode is reachable from here.

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -207,6 +207,19 @@ func (l *Logger) WithValues(keysAndValues ...any) *Logger {
 	}
 }
 
+// WithoutStackTraces returns a copy of a logger that does not attach a stack
+// trace to the records it writes. A stack trace describes the path to the
+// logging call, not the path to whatever went wrong, so it is worth having only
+// where an error is logged near where it arose. Callers that log errors on
+// behalf of code far below them should use this.
+func (l *Logger) WithoutStackTraces() *Logger {
+	return &Logger{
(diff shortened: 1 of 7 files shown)


fuskovic (approved):
LGTM

PR #6771, opened 2026-08-06
fix(ui): unblock the token renewal page's own requests

### LINKED ISSUE #6746: UI fails to renew OIDC token
### Pre-submission checklist

### Description

We were hoping to reduce our OIDC ID token lifetime to 1h and rely on refresh tokens to keep the sessions going. While this works perfectly in the CLI tool since #6523 landed in v1.11.0, we started to notice the UI logged users out every hour, returning to the login screen.

### Steps to Reproduce

Steps to reproduce in the local development environment:

1. Shorten the ID token lifetime. On our side, we're using a custom Dex config, but the easiest is to add these lines to the `kargo-dex-server` template:
```diff
  stringData:
    config.yaml: |-
+     expiry:
+       idTokens: 2m
```
2. Start the Local Tilt dev env
3. Log in to the UI with SSO
4. Wait at least 2 minutes 
5. Click on any links to trigger a token refresh

### Expected Behavior

I'd expect the UI to transparently refresh the session token without interruption. 

### Actual Behavior

The UI navigates to `/token-renew`, reloads itself several times, then lands on
`/login` with both `auth_token` and `refresh_token` cleared. The user must sign
in again.

```
18:28:59.106  GET  ---  /token-renew?redirectTo=/
18:28:59.134  GET  ---  /token-renew?redirectTo=/
18:28:59.166  GET  200  /token-renew?redirectTo=/
18:28:59.561  GET  304  /token-renew?redirectTo=/token-renew
18:28:59.893  GET  200  /v1beta1/system/public-server-config
18:28:59.924  GET  401  /v1beta1/projects?pageSize=100&page=0&filter=
18:28:59.993  GET  304  /login?redirectTo=/
18:29:00.296  GET  200  /v1beta1/system/public-server-config
18:29:00.369  GET  200  /dex/.well-known/openid-configuration
        ... interactive sign-in ...
18:29:05.677  POST 200  /dex/token
```

### Kargo Version

v1.11.0

### Relevant Logs

```shell
# Kargo API server (LOG_LEVEL=DEBUG), from the reload above.
# The request arrives with no credential, because logout() has already cleared
# auth_token by the time it is sent.
2026-07-31T13:34:50Z	DEBUG	server/auth_middleware.go:123	authenticating request	{"path": "/v1beta1/projects"}
2026-07-31T13:34:50Z	DEBUG	server/auth_middleware.go:131	authentication failed	{"path": "/v1beta1/projects", "error": "no token provided"}
2026-07-31T13:34:50Z	DEBUG	server/auth_middleware.go:123	authenticating request	{"path": "/v1beta1/system/public-server-config"}
2026-07-31T13:34:50Z	DEBUG	server/auth_middleware.go:154	skipping authentication for exempt path	{"path": "/v1beta1/system/public-server-config"}
2026-07-31T13:34:50Z	DEBUG	server/auth_middleware.go:140	authentication successful	{"path": "/v1beta1/system/public-server-config"}
[GIN] 2026/07/31 - 13:34:50 | 200 |    145.25µs |      10.42.0.44 | GET      "/v1beta1/system/public-server-config"
2026-07-31T13:34:51Z	DEBUG	server/auth_middleware.go:123	authenticating request	{"path": "/v1beta1/system/public-server-config"}
2026-07-31T13:34:51Z	DEBUG	server/auth_middleware.go:154	skipping authentication for exempt path	{"path": "/v1beta1/system/public-server-config"}
2026-07-31T13:34:51Z	DEBUG	server/auth_middleware.go:140	authentication successful	{"path": "/v1beta1/system/public-server-config"}
[GIN] 2026/07/31 - 13:34:51 | 200 |   186.875µs |      10.42.0.44 | GET      "/v1beta1/system/public-server-config"

# Dex (logLevel: DEBUG) for the same session. The only entry is the original
# login. Nothing is logged at 13:34, because no refresh grant ever arrives:
time=2026-07-31T13:32:35.062Z level=INFO msg="login successful" connector_id=mock username="Kilgore Trout" preferred_username="" email=kilgore@kilgore.trout groups=[authors] request_id=e0055f50-c33a-4caa-babe-ffa432626a56
```

### Screenshots

_No response_

antoinedeschenes:
## Description

Closes #6746

With OIDC enabled, the UI could not renew a session: at ID token expiry it bounced through /token-renew and landed logged out on /login, without ever sending the refresh token grant.

The renewal page needs the public server config before it can run the grant. Until v1.10 it fetched that config over the unauthenticated ConnectRPC transport, so the fetch was never subject to any token checks. The v1.11 REST migration regenerated that call onto customFetch, whose pre-flight sees the expired token, starts another navigation to /token-renew, and throws before sending. The renewal page's first request is blocked by the very check that sent the user there, so renewal cannot complete no matter how often it is retried. The config query then errors, and TokenRenew responds by logging out, which also discards the refresh token. Direct visits to /login with an expired token stored trip the same check through that page's own config fetch.

customFetch now skips the token handling for endpoints that require no authentication, keeping a client-side copy of the entries the UI calls from the server's exempt list (exemptPaths in pkg/server/auth_middleware.go): the public server config and admin login. Exempt requests are sent without a token, and a 401 on them no longer ends the session.

Verified against a Dex install issuing 2m ID tokens: renewal completes at expiry (POST /token returns 200 and the session continues, matching v1.10.9 behavior on the same cluster), /login with an expired token stored no longer bounces, and admin login and its expiry logout are unaffected.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/ui/src/lib/api/custom-fetch.ts
+++ b/ui/src/lib/api/custom-fetch.ts
@@ -36,6 +36,13 @@ const renewToken = () => {
   );
 };
 
+// The endpoints the UI calls that require no authentication. Their requests
+// must not be blocked by the token expiry check, or the renewal and login
+// pages could never fetch the public config they depend on. Every entry here
+// must be on the server's exempt list (exemptPaths in
+// pkg/server/auth_middleware.go).
+const authExemptPaths = new Set(['/v1beta1/system/public-server-config', '/v1beta1/login']);
+
 /**
  * Custom fetch function used by all generated API hooks.
  *
@@ -50,7 +57,8 @@ export const customFetch = async <T>(url: string, options?: RequestInit): Promis
   const baseUrl = getBaseUrl();
   const fullUrl = `${baseUrl}${url}`;
 
-  const token = localStorage.getItem(authTokenKey);
+  const requiresAuth = !authExemptPaths.has(url.split('?')[0]);
+  const token = requiresAuth ? localStorage.getItem(authTokenKey) : null;
   const refreshToken = localStorage.getItem(refreshTokenKey);
 
   if (token) {
@@ -103,7 +111,7 @@ export const customFetch = async <T>(url: string, options?: RequestInit): Promis
     headers
   });
 
-  if (response.status === 401) {
+  if (requiresAuth && response.status === 401) {
     logout();
   }
 
--- a/ui/src/lib/api/custom-fetch.ts
+++ b/ui/src/lib/api/custom-fetch.ts
@@ -36,12 +36,31 @@ const renewToken = () => {
   );
 };
 
+// Reduces a request URL to just its lowercased path, so that neither a query
+// string, a fragment, a dot segment, nor casing can silently defeat an
+// exemption. Resolving against the current origin is what the browser does
+// with these paths anyway, so the result is the path the server receives.
+const normalizePath = (url: string): string => {
+  try {
+    return new URL(url, window.location.origin).pathname.toLowerCase();
+  } catch {
+    // Unparseable, so it cannot match a known path. Returning the input
+    // unchanged fails closed: the caller treats it as requiring auth.
+    return url;
+  }
+};
+
 // The endpoints the UI calls that require no authentication. Their requests
 // must not be blocked by the token expiry check, or the renewal and login
 // pages could never fetch the public config they depend on. Every entry here
 // must be on the server's exempt list (exemptPaths in
-// pkg/server/auth_middleware.go).
-const authExemptPaths = new Set(['/v1beta1/system/public-server-config', '/v1beta1/login']);
+// pkg/server/auth_middleware.go). Each is registered both bare and with a
+// trailing slash, the two forms a caller could reasonably write. Entries are
+// written pre-normalized because normalizePath reads window, which is not
+// available when this module is first evaluated.
+const authExemptPaths = new Set(
+  ['/v1beta1/system/public-server-config', '/v1beta1/login'].flatMap((path) => [path, `${path}/`])
+);
 
 /**
  * Custom fetch function used by all generated API hooks.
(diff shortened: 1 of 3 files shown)


ui/src/lib/api/custom-fetch.ts
rpelczar, line 1: Nice catch on the root cause. One suggestion on scoping: key the exemption off the **endpoint** rather than the page. `public-server-config` has to bypass the check because it's unauthenticated, not because the renewal page is the caller - and the server already keeps that list, in `exemptPaths` (`pkg/server/auth_middleware.go`)

Let's do something like this:

```ts
const authExemptPaths = new Set(['/v1beta1/system/public-server-config', '/v1beta1/login']);

const requiresAuth = !authExemptPaths.has(url.split('?')[0]);
const token = requiresAuth ? localStorage.getItem(authTokenKey) : null;
// ...
if (requiresAuth && response.status === 401) {
  logout();
}
```

This also covers `/login`, which trips the same check today - its `useGetPublicConfig()` fires before the `isLoggedIn` redirect - and it avoids suppressing the 401 logout for *every* request on the renewal page, which is worth avoiding on an `area/security` issue.

antoinedeschenes: Ah right that looks cleaner. I updated the PR to use the exemption list

ui/src/lib/api/custom-fetch.ts
```diff
@@ -50,7 +57,8 @@ export const customFetch = async <T>(url: string, options?: RequestInit): Promis
   const baseUrl = getBaseUrl();
   const fullUrl = `${baseUrl}${url}`;
 
-  const token = localStorage.getItem(authTokenKey);
+  const requiresAuth = !authExemptPaths.has(url.split('?')[0]);
```
Marvin9: We should do proper URL check here. if `url` is either `/v1beta1/system/public-server-config/` or `/v1beta1/login#foobar` for any reason, the bug remains.

I will push changes now in this PR

Marvin9 (commented):
Tested it with expiry configuration in dex and token is renewed now instead of log-in.

Marvin9 (approved):

PR #6772, opened 2026-08-06
feat(api): report promotion window state on Stage status

hairyhum:
## Description

Supersedes https://github.com/akuity/kargo/pull/6737

Some name replacements compared to #6737:
- `promotionSchedule` -> `promotionWindowStatus` - unify naming with `promotionWindows`
- `closed` is non-optional if `promotionWindowStatus` exists for clarity
- `reason` should only describe a reason for closed window, `nextCloseReason` is added to describe a reason for `nextClose` - this is to prevent people from thinking that `reason` describes a reason why it's open.

Adds PromotionWindowStatus to StageStatus, reporting whether promotion windows (https://github.com/akuity/kargo/pull/6710) currently permit promotion of a Stage.

## Variations a client will observe

**No windows gate the Stage.** The common case — the field is absent entirely, not an empty object.

```yaml
status:
  autoPromotionEnabled: true
  # no promotionSchedule key
```

**Frozen, reopening known.**

```yaml
status:
  promotionSchedule:
    closed: true
    reason: 'Promotion freeze "change-freeze" is active'
    nextOpen: "2026-07-30T22:00:00Z"
```

**Frozen, no reopening known.** `nextOpen` is omitted. Here a one-shot `Allow` window has elapsed, so promotion is forbidden for good; an implementation may also report an indefinite freeze this way.

```yaml
status:
  promotionSchedule:
    closed: true
    reason: No promotion window is currently active
```

**Promotion allowed, freeze approaching.** `closed` is `false`. `nextCloseReason` describes the freeze that is coming, so a warning can name it.

```yaml
status:
  promotionSchedule:
    closed: false
    nextClose: "2026-07-31T09:00:00Z"
    nextCloseReason: 'Promotion freeze "change-freeze" begins'
```

**Promotion allowed, nothing scheduled ahead.** A one-shot `Deny` window that has elapsed leaves no boundary to report. **This is identical to `promotionSchedule` field being absent**

```yaml
status:
  promotionSchedule:
    closed: false
```

## Contract

- `closed` indicates whether promotions are blocked
- `reason` describes why `closed` is `true`, should only be non-empty if `closed` is `true`
- `nextOpen` is optional **even when `closed` is true**. A schedule need not ever reopen, so requiring it would be a contract the first implementation breaks. No CEL validation links the two. Treat its absence as "frozen, with no known end" and render `reason`.
- `nextClose` is meaningful only while promotion is allowed, and is likewise optional.
- `nextCloseReason` describes a reason for expected next closure, should only be non-empty if `nextClose` is present.
- At most one of `nextOpen` / `nextClose` appears: only the boundary ahead is reported.
- **Advisory.** The value is only as current as the Stage's last reconciliation. Admission remains the sole arbiter of whether a Promotion is permitted, so this should not gate a promote action — a client that treats it as authorization will eventually act on a stale value.

`AutoPromotionEnabled` is the precedent: likewise a status field derived from `ProjectConfig` so clients need not recompute policy.

## A note on wording

The messages say "promotion freeze" rather than "Deny promotion window", and "promotion window" rather than "Allow promotion window". `kind: Allow|Deny` remains the configuration vocabulary, where an explicit pair is unambiguous; it just reads poorly in a sentence, and these strings are shown to people.

## Testing

API-only; no behavior change. Full codegen (CRD, deepcopy, OpenAPI, generated clients, UI schema and models). `api` and root modules build and their tests pass. The Stage CRD gains one optional object under `status`, so the change is additive for existing resources.


api/v1alpha1/promotion_window_types.go
```diff
@@ -80,3 +82,57 @@ type PromotionWindow struct {
...
+	// an approaching freeze, which NextCloseReason names. Like NextOpen it is optional, and
+	// absent when no such boundary is known -- a schedule that will never forbid
+	// promotion again has none.
+	//
+	// +optional
+	NextClose *metav1.Time `json:"nextClose,omitempty"`
+	// NextCloseReason explains in human-readable terms why promotion will be forbidden
+	// at NextClose, naming the freeze responsible where there is one.
+	// It is set whenever NextClose is set.
+	NextCloseReason string `json:"nextCloseReason,omitempty"`
```
fuskovic, line 139: Should we add an `+optional` marker here?

```suggestion
    //
    // +optional
	NextCloseReason string `json:"nextCloseReason,omitempty"`
```

fuskovic (approved):
One small nit other than that LGTM

hairyhum:
@rpelczar please take a look since you were going to work on the UI part of it.

rpelczar (approved):
LGTM

Issue #5179, opened 2025-10-09
Wrong image builds sorting


lucj:
# Checklist

* [x] I've searched the issue queue to verify this is not a duplicate bug report.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `kargo version`
* [ ] I've pasted logs, if applicable.

# Description

I've created a Warehouse for each microservice of my app. For example, for the `api` microservice.

```yaml
apiVersion: kargo.akuity.io/v1alpha1
kind: Warehouse
metadata:
  name: api
  namespace: kargo-mygroup
spec:
  subscriptions:
  - image:
      repoURL: registry.gitlab.com/mygroup/api
      imageSelectionStrategy: NewestBuild
      allowTags: '^([0-9a-f]{7,})$'
      discoveryLimit: 10
```

I noticed that for some microservices, the Freight is correctly sorted, but for others, it's not. The items appear with the same date, leading to a wrong tag being promoted in my values file.

# Version

I'm using Kargo 1.7.5

lucj:
I've noticed that:

- If the Source Repo is not known, then the image tags are in the correct order (screenshot from the Warehouse "Freight Assembly" tab).

<img width="1554" height="273" alt="Image" src="https://github.com/user-attachments/assets/e03beaf2-391a-442b-a8a5-a337458d277a" />

- If the Source Repo is known, then the date is always the same (probably the one of the Source Repo, which seems to be the base image used to build the microservice) => wrong tag selected.

<img width="1570" height="412" alt="Image" src="https://github.com/user-attachments/assets/19ee2492-e446-476d-b87e-a95a78716553" />

lucj:
Closing this one as using specific oci labels fixed the thing.
For example, these are the labels I'm using in my CI:

```
--label org.opencontainers.image.source="$CI_PROJECT_URL"
--label org.opencontainers.image.revision="$CI_COMMIT_SHA"
--label org.opencontainers.image.created="$(date -u +'%Y-%m-%dT%H:%M:%SZ')"
--label org.opencontainers.image.version="${CI_COMMIT_TAG:-$CI_COMMIT_SHORT_SHA}"
--label org.opencontainers.image.title="$CI_PROJECT_NAME"
--label org.opencontainers.image.description="Built by GitLab CI for $CI_PROJECT_PATH"
```

jlsan92:
[...]

PR #6777, opened 2026-08-07
fix(controller)!: use the latest available creation time for NewestBuild

krancour:
An image's creation time was taken from an `org.opencontainers.image.created` or `org.label-schema.build-date` label in preference to the creation time recorded in its config. Labels are copied from the base image unless a build overwrites them, so an image that sets no label of its own reports the build time of the image it was built from.

`ubuntu:26.04` -- and therefore `ubuntu:latest` -- carries such a label; `ubuntu:24.04` did not. Every image built from one of these between two republishes of the base reports an identical creation time. `NewestBuild` then sorts on a tie, falls back to reverse-lexical tag order, and truncates to `discoveryLimit`, so new images can go undiscovered rather than merely mis-ranked.

Separately, annotations on an index, and on an index's references to the images it contains, were read for display but never for creation time. Multi-arch builds annotated with `buildx --annotation index:...` got no benefit from them.

Creation time is now taken from the outermost metadata that claims one -- an index's annotations, then an index's reference to an image, then the image's own manifest -- and, failing all of those, from the later of the config's creation time and any claim in its labels. No object in a registry can be older than anything referencing it, and every way one of these timestamps goes wrong understates it: a label copied from a base image, a timestamp carried over with a reused layer, a fixed timestamp chosen to make digests reproducible. Taking the outermost claim, and otherwise the later of the two remaining, amounts to using the latest date available.

Verified against a local registry with images built to exhibit each shape: base image label inheritance, index-only annotations, a reproducible build with a fixed config timestamp, and all four metadata locations populated at once.

Note for upgrades: images that were previously mis-ranked, or truncated away by `discoveryLimit`, may become visible. On a `Warehouse` with auto-promotion enabled, that can trigger promotions for images that already existed. This is why this fix will not be back ported to v1.11.x and will, instead, ship with v1.12.0.

Refs: #4635, #4636, #5179

cc @34fathombelow

### DIFF
--- a/docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
+++ b/docs/docs/50-user-guide/20-how-to-guides/30-working-with-warehouses.md
@@ -284,10 +284,21 @@ strategies are:
 - `NewestBuild`: This strategy selects the image with the most recent build
   time.
 
-  The build time is evaluated using the labels
-  `org.opencontainers.image.created` or `org.label-schema.build-date`. If
-  neither label is set, Kargo will fall back to using the `config.Created` time
-  of the image.
+  Build time is the latest date Kargo can find for an image: any
+  `org.opencontainers.image.created` or `org.label-schema.build-date` annotation
+  or label the image carries, and the creation time recorded in the image's own
+  config.
+
+  :::note
+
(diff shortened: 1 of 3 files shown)


fuskovic (approved):
lgtm

PR #6787, opened 2026-08-10
fix(ui): handle non-JSON AnalysisRun measurement values

### LINKED ISSUE #6785: ui(verification): AnalysisRun fails to popup
### Pre-submission checklist

### Description

When you have a basic Analysis template with a call to a service which should return plain text it attempts to parse the output as JSON, which then causes a failure in the UI meaning the panel won't appear. 

### Steps to Reproduce

1. Create stage with an analysis run
```yaml
kind: AnalysisTemplate
metadata:
  name: my-healthy-app
  namespace: kargo-my-healthy-app
spec:
  metrics:
    - name: echo-test
      initialDelay: 30s
      interval: 30s
      count: 3
      failureLimit: 1
      successCondition: result.status == "ok"
      provider:
        web:
          url: http://mycool.thing.svc.cluster.local/v1/echo/kargo
```
2. The service returns plain text.
3. Try to view the AnalysisRun in the Kargo UI

### Expected Behavior

I expect it to pop up with the Analysis Run

### Actual Behavior

Instead a small "flash" happens and nothing else.

### Kargo Version

v1.11.0

### Relevant Logs

```shell
index-DDDIzxEX.js:421 SyntaxError: Unexpected token 'k', "kargo" is not valid JSON
    at JSON.parse (<anonymous>)
    at Ldr (index-DDDIzxEX.js:2709:19863)
    at t.length.t.reduce.chartable (index-DDDIzxEX.js:2709:18686)
    at Array.reduce (<anonymous>)
    at jdr (index-DDDIzxEX.js:2709:18665)
    at index-DDDIzxEX.js:2709:16008
    at Array.forEach (<anonymous>)
    at Cdr (index-DDDIzxEX.js:2709:15623)
    at index-DDDIzxEX.js:2709:23747
    at Object.mo [as useMemo] (index-DDDIzxEX.js:421:56520)
```

### Screenshots

_No response_

SuyogShinde942:
## Summary
- `transformMeasurementValue()` in the analysis modal called `JSON.parse()` on every metric measurement value unconditionally. Providers such as `web` can return a plain-text response (not JSON), which throws inside the `useMemo()` that builds the AnalysisRun chart/table data and leaves the panel unable to open (a brief "flash" with no further UI, per the console error in the issue).
- Catch the parse failure and fall back to displaying the raw string in the table, non-chartable — consistent with how other unsupported value shapes are already handled in this function.
- Added a unit test covering a plain, non-JSON string measurement value.

Fixes #6785

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm lint` (targeted eslint run on changed files)
- [x] `pnpm test` (`vitest run`) — 187/187 passing, including new test case

### DIFF
--- a/ui/src/features/common/analysis-modal/transforms.test.ts
+++ b/ui/src/features/common/analysis-modal/transforms.test.ts
@@ -24,7 +24,8 @@ import {
   metricStatusLabel,
   metricSubstatus,
   printableCloudWatchQuery,
-  printableDatadogQuery
+  printableDatadogQuery,
+  transformMeasurements
 } from './transforms';
 import { AnalysisStatus, FunctionalStatus } from './types';
 
@@ -557,6 +558,21 @@ describe('analysis modal transforms', () => {
       canChart: false,
       chartValue: { latency: null, cpuUsage: null },
       tableValue: { latency: null, cpuUsage: null }
+    });
+  });
+
+  test('transformMeasurements() with a plain, non-JSON string measurement value', () => {
+    expect(transformMeasurements([], [{ value: 'kargo' }])).toEqual({
+      chartable: false,
+      min: 0,
+      max: null,
+      measurements: [
+        {
+          value: 'kargo',
+          chartValue: undefined,
+          tableValue: 'kargo'
+        }
+      ]
     });
   });
 });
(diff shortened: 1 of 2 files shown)


SuyogShinde942:
With this fix applied, the modal renders normally and shows the raw value in the measurement table instead of crashing:
after-fix-6785-table-renders

Marvin9:
There is already a PR open similar to this https://github.com/akuity/kargo/pull/6758

PR #6789, opened 2026-08-11
feat(api): add PromotionRequest type and reconciler seam

jsbroks:
## Description

Introduces the `PromotionRequest` API. A PromotionRequest expresses standing intent to promote a piece of Freight to the Targets its Stage governs -- a living object whose target selectors the Stage keeps in sync, closer to a ReplicaSet than a Job:

- Defines immutable `stage` and `freight`, plus mutable `targetSelectors` matching Targets in the Project. Resolution lives in status: one entry per selected Target with its child Promotion and phase, alongside summary counts, conditions, and timestamps.
- Generates the corresponding CRD, deepcopy implementations, and UI schema so the resource is installable and discoverable.
- Adds a reconciler with an injectable reconciliation function, so the fan-out behavior can be supplied separately. The default implementation records a terminal `Errored` status explaining that nothing is currently reconciling PromotionRequests.
- Adds RBAC: the controller can create PromotionRequests and patch their status. Default project roles deliberately grant users no PromotionRequest access -- only the Stage controller creates them, so no user needs to write one, and withholding read access keeps the resource inert until something reconciles it. Any installation that does can grant those permissions through `RegisterRoleRulesContributor`.

This PR intentionally does not add the fan-out reconciler, REST endpoints, UI workflows, or request-level abort behavior.

## Validation

- `make lint`
- `make test-unit`
- `make test-chart`
- `go test ./...` from `api/`

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

- [ ] Are signed off by their author (`git commit -s`) **(required)**


api/v1alpha1/promotion_set_types.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// Targets specifies the Targets to which this PromotionSet promotes Freight.
+	// Each Target MUST be in the same namespace as the PromotionSet. The list
+	// may be empty, which records that the governing Stage's target selectors
+	// matched no Targets at the time the PromotionSet was created.
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf"
```
thomastaylor312: Looking at this with fresh eyes, I think we might not want to do this. If I remember correctly, the map list type will make the field manager record an entry for each item of the list. This essentially doubles the amount of space taken up by each target. Assuming we will have thousands of targets (and possibly 10s of thousands), we might get a bit close to the size limit for etcd and/or the API payload. Also, the CEL rule here may run into issues if the list is really long. Not super familiar with the "weight" calculation for those CEL requests, but we could run into it

My back of napkin math with how this currently work says we can have roughly 9k-17k targets captured here before we hit the limit (depending on how long the names are). We can probably get away with this for now, but we should keep it in mind.

thomastaylor312 (commented):
One small observation I found. I'll reach out to you directly with some more nuance on a few other things

thomastaylor312 (approved):

PR #6796, opened 2026-08-12
feat(cli): export flag for kargo get subcommands

### LINKED ISSUE #6738: Kargo Export API
### Pre-submission checklist

### Proposed Feature

Kargo API should have an option to export all the managed resources (Stages, Warehouses, ProjectConfig, AnalysisTemplates, ConfigMaps, etc...) so that they can commit the files to git. During export, the API would strip any fields that would make those manifests incompatible with gitops (e.g. `.status` and other non-applyable fields like `.metadata.creationTimestamp`, `metadata.resourceVersion`, etc...).

### Motivation and Use Case

Users might experiment/tweak Stage, ProjectConfig, etc... through Kargo UI, but later want to commit those changes to git later.

Additionally, features like https://github.com/akuity/kargo/issues/6735 will allow users to get started using a wizard, but then want to convert to GitOps after they are up and running

### Alternatives Considered

_No response_

fuskovic:
Closes: https://github.com/akuity/kargo/issues/6738

### DIFF
--- a/docs/docs/50-user-guide/05-cli/20-usage.md
+++ b/docs/docs/50-user-guide/05-cli/20-usage.md
@@ -61,9 +61,19 @@ kargo get freight --project my-project --warehouse my-warehouse
 
 # List Promotions for a Stage
 kargo get promotions --project my-project --stage test
+
+# Export Stages as git-friendly manifests, e.g. after tweaking them in the UI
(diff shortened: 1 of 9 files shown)


fuskovic:
Closing. We want to spec this out first.

PR #6802, opened 2026-08-12
fix(ui): tolerate non-JSON AnalysisRun measurement values

### LINKED ISSUE #4630: Cannot view AnalysisRun result in UI - not a valid JSON

utkarshcmu:
## Description

Closes #4630

Argo Rollouts' `web` provider passes a plain text response body through unchanged, so an AnalysisRun measurement value is not guaranteed to be JSON. `transformMeasurementValue` called `JSON.parse` without a guard, so clicking an AnalysisRun under a Stage's Verifications tab failed with `SyntaxError: Unexpected token 'P', "PASS" is not valid JSON` and the analysis modal did not render at all.

An unparseable value is now treated as a non-chartable table value, surfacing the raw string rather than discarding it. This is consistent with how the function already handles unsupported formats a few lines below (`// unsupported formats are stringified and put into table`).

Verified the added tests fail without the fix, reproducing the reporter's exact `SyntaxError`, and pass with it. Full UI suite passes (190 tests); `lint-ui` and `typecheck-ui` are clean.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/ui/src/features/common/analysis-modal/transforms.test.ts
+++ b/ui/src/features/common/analysis-modal/transforms.test.ts
@@ -24,7 +24,8 @@ import {
   metricStatusLabel,
   metricSubstatus,
   printableCloudWatchQuery,
-  printableDatadogQuery
+  printableDatadogQuery,
+  transformMeasurements
 } from './transforms';
 import { AnalysisStatus, FunctionalStatus } from './types';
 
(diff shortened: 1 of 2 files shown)


utkarshcmu:
Closing as a duplicate — #6758 addresses the same issue and was opened first. Apologies for the noise; I missed the existing PR when picking this up.

PR #6803, opened 2026-08-12
feat(cli): remember the SSO callback port across logins

### LINKED ISSUE #6800: kargo login should remember when `--port` is used for the SSO callback
### Pre-submission checklist

### Proposed Feature

When logging in with `kargo login --sso --port <port>`, persist the specified callback port in the CLI configuration file and reuse it on subsequent logins, the same way the server address and auth method are already remembered.

Proposed behavior:

- `kargo login <server> --sso --port 8085`: logs in and saves the port alongside the auth method and server address.
- A later bare `kargo login`: reuses the stored SSO method, server address, and callback port.
- `kargo login --sso --port <other-port>`: overrides and saves the new port.
- If no port was ever specified, nothing is stored and the CLI keeps selecting a random unprivileged port (current behavior, unchanged).
- The stored port is scoped to the server address it was saved for; logging in to a different server does not reuse it.

### Motivation and Use Case

Some OIDC identity providers (e.g. Okta) require redirect URIs to be registered exactly, including the port, and don't support wildcard loopback ports the way Dex does. In those setups the administrator registers a fixed redirect URI such as `http://localhost:8085/auth/callback`, and every CLI user must log in with `--port 8085`.

Today the CLI remembers the auth method and server address, so `kargo login` alone is enough to re-authenticate, except that the port resets to random every time, and the login fails with a redirect URI mismatch until the user remembers to re-add `--port`. This comes up on every interactive re-login once the refresh token has expired.

This is the same pain point described from the admin side in #614 ("that admin must instruct all of their kargo users to run the `kargo login` command with the `--port 1234` flag"). Persisting the port makes it a one-time setup per machine instead of a per-login requirement.

### Alternatives Considered

- Shell aliases or wrapper scripts around kargo login. Works but every user has to set it up.
- Registering multiple redirect URIs at the IdP: the random port range is too wide to do this.

antoinedeschenes:
## Description

Closes #6800

When `--port` is specified during SSO login, persist it in the CLI configuration (as an optional `callbackPort` field) and reuse it on subsequent logins to the same server, the same way the server address and auth method are already remembered.

- `kargo login <server> --sso --port 8085` saves the port alongside the auth method and server address; a later bare `kargo login` reuses all three.
- Specifying `--port` again overrides and saves the new value.
- When no port was ever specified, nothing is stored and the CLI keeps selecting a random unprivileged port (current behavior, unchanged).
- The stored port is scoped to the server address it was saved for; logging in to a different server does not reuse it.

This helps with OIDC identity providers that require redirect URIs to be registered exactly, including the port (e.g. Azure AD / Entra ID, Okta), where users previously had to re-specify `--port` on every interactive re-login.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/pkg/cli/cmd/login/login.go
+++ b/pkg/cli/cmd/login/login.go
@@ -83,6 +83,14 @@ kargo login https://kargo.example.com --kubeconfig
 # Log in using the local kubeconfig and ignore cert warnings
 kargo login https://kargo.example.com --kubeconfig --insecure-skip-tls-verify
 `),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// A login method must be specified unless we're falling back on a stored
+			// one, which only happens when no server address is specified.
+			if len(args) == 1 || cmdOpts.Config.AuthMethod == "" {
+				cmd.MarkFlagsOneRequired("admin", "kubeconfig", "sso")
+			}
+			return nil
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmdOpts.complete(args)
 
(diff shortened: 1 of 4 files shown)


pkg/cli/cmd/login/login.go
```diff
@@ -136,6 +136,12 @@ func (o *loginOptions) complete(args []string) {
 		o.ServerAddress = normalizeURL(args[0])
 	}
 
+	// If no port was specified, reuse any port stored in the configuration,
+	// provided we're logging in to the same server it was stored for.
+	if o.CallbackPort == 0 && o.ServerAddress == o.Config.APIAddress {
+		o.CallbackPort = o.Config.CallbackPort
+	}
```
krancour: The `o.ServerAddress == o.Config.APIAddress` requirement for inheriting port from the config made realize other aspects of what we were already doing here are odd. As currently written, you could log in to a new server, but inherit options such as authn method from the _previous_ server you'd been logged into. Super odd. These new lines of yours make more sense than what's already here... but we can do even better.

If you don't mind, I'd like to see the whole method re-written like this:

```
// complete sets the options from the command arguments.
func (o *loginOptions) complete(args []string) {
	if len(args) == 1 {
		o.ServerAddress = normalizeURL(args[0])
	}
	// If a server address was not specified, but one is stored in the
	// configuration, inherit it from the configuration. Only if doing so, also
	// inherit other login options from the configuration.
	if o.ServerAddress == "" && o.Config.APIAddress != "" {
		o.ServerAddress = o.Config.APIAddress
		if !o.UseAdmin && !o.UseKubeconfig && !o.UseSSO && o.Config.AuthMethod != "" {
			switch o.Config.AuthMethod {
			case authMethodAdmin:
				o.UseAdmin = true
			case authMethodKubeconfig:
				o.UseKubeconfig = true
			case authMethodSSO:
				o.UseSSO = true
			}
		}
		if o.CallbackPort == 0 {
			o.CallbackPort = o.Config.CallbackPort
		}
		// Note: We deliberately do NOT inherit o.InsecureTLS from the stored config
		// because we want to force the user to re-assess that choice each time they
		// log in.
	}
}
```

This has two small consequences.

The following needs to be removed from the `addFlags()` method. It is now insufficient because not requiring an authn method to be explicitly specified now _also_ requires that no URL was explicitly specified. Args cannot be inspected here because of the point in the command's lifecycle at which `addFlags()` is invoked.

```
	// If we do not have a saved configuration, we require the user to specify a
	// login method. If we do have a saved configuration, we allow the user to
	// omit the login method and use the last used method.
	if o.Config.AuthMethod == "" {
		cmd.MarkFlagsOneRequired("admin", "kubeconfig", "sso")
	}
```

Something slightly different from what was removed from `addFlags()` needs to be added as a `PreRunE` in `NewCommand()`. Args are available for inspection at that point without being too late to mark flags required.

```
		PreRunE: func(cmd *cobra.Command, args []string) error {
			// A login method must be specified unless we're falling back on a stored
			// one, which only happens when no server address is specified.
			if len(args) == 1 || cmdOpts.Config.AuthMethod == "" {
				cmd.MarkFlagsOneRequired("admin", "kubeconfig", "sso")
			}
			return nil
		},
```

Obviously some of the tests you just added might need a tweak after these changes.

🙏

antoinedeschenes: Ah these are good points, changes applied with new tests. I validated the CLI drops saved parameters when specifying the hostname, still accepts overrides and otherwise reuses all saved parameters.

krancour: Thank you so much for this!

krancour (approved):
🔥

PR #6804, opened 2026-08-12
test(e2e): E2E test improvements

hairyhum:
## More test suites

1. Added generated test suites from https://github.com/akuity/kargo-examples
- Using git repo and credentials configurable with env files
- Removed verification templates because they were just noop waiting (for faster tests)
- Using promotion names instead of stages to prevent idempotent steps like `git-commit` and `git-open-pr` from being skipped **Not a production-compatible examples, but is good for tests**

2. Added env functions to support creating `kind` cluster with kargo and argocd installed.


hairyhum:
Added remaining examples to suites. 

- HTTP example is skipped unless `http_endpoint` is set
- `02-git-driven/03-kustomize-driven/01-image-only/02-monorepo` example is deliberately not adapted.

hack/test/e2e/envs/home_config.yaml
```diff
@@ -8,4 +8,7 @@ argocd_cli:
   config_file: ~/.config/argocd/config
 
 context:
-  kargo_demo_gitops_repo: "https://github.com/hairyhum/kargo-demo-gitops.git"
\ No newline at end of file
+  # kargo_demo_gitops_repo: <specify a repo here>
+  # git_pat: <specify a PAT here>
+  # http_endpoint: <URL of an HTTP endpoint that returns 2xx to a POST>
```
thomastaylor312, line 13: Nit: Might be worth saying this can be specified by an env var (which I think it can if I'm reading things right later down)

hack/test/e2e/envs/kind_config.yaml
```diff
@@ -0,0 +1,80 @@
...
+  image: ""
+  set: []
+  values_files: [
+    ../../values.test.yaml
+  ]
+
+kargo_cli:
+  ## Indicate that we need to run `kargo login` to login into a kargo instance in `kind` cluster
+  kargo_login:
+    ## Uste tmp directory to store kargo config
```
thomastaylor312: Nit(spelling): s/Uste/Use

hack/test/e2e/framework/envfuncs/cluster.go
```diff
@@ -0,0 +1,372 @@
...
+	// This is a simplified approach when we just run a background shell.
+	// There is no error handling here, it might fail silently.
+	// FIXME: replace that with goroutine and error channels?
+	// FIXME: implement forwarding to an non-predefined port
+	cmd := fmt.Sprintf("sh -c \"kubectl port-forward --kubeconfig %s --namespace %s %s %d:%d > /dev/null 2>&1 &\"", 
+		kubeconfig, namespace, service, outport, inport)
+
+	fmt.Printf("Port forwarding %s to %d", service, outport)
+	
+	p := utils.RunCommand(cmd)
```
thomastaylor312: Heads up that `RunCommand` is deprecated in favor of `RunCommandContext`. Also, does this reap the process on shutdown or does it stay hanging?

hack/test/e2e/framework/envfuncs/kargo_cli.go
```diff
@@ -3,21 +3,114 @@ package envfuncs
...
+		}
+		return ctx, fmt.Errorf("kargo login failed: %w : %s", p.Err(), outBytes)
+	}
+
+	return ctx, nil
+}
+
+func processLoginConfig(ctx context.Context, loginConfigVal any) (context.Context, func(), error) {
+	loginConfig, ok := loginConfigVal.(map[string]any)
+	if ok {
```
thomastaylor312: Any reason we don't just make this extra config a concrete type? It would avoid a bunch of nesting and I don't think there are a ton of values. If this was being passed verbatim to something else to decode, I think it would be fine, but this seems fairly concrete

If we do keep it this way, one nit is that we can just invert the logic to avoid the huge nested block by doing:

```go
if !ok {
	return ctx, nil, nil
}
```

hack/test/e2e/framework/envfuncs/load_env.go
```diff
@@ -58,4 +63,34 @@ func GetEnvMap(ctx context.Context, path []string) (map[string]any, error) {
 		return nil, fmt.Errorf("cannot convert env to map %v", env)
 	} 
 	return nil, err
+}
+
+func GetValueOrEnv(ctx context.Context, valueKey ContextKey, path []string) (any, error) {
```
thomastaylor312, line 68: Nit: I think this would be cleaner if the path argument was variadic (`...string`) as it would be cleaner to call

hack/test/e2e/framework/funcsloader/funcsloader.go
```diff
@@ -2,22 +2,39 @@ package funcsloader
...
 	"context"
+	"slices"
 
+	"github.com/akuity/kargo/hack/test/e2e/envfuncs"
 	"sigs.k8s.io/e2e-framework/pkg/env"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
-	"github.com/akuity/kargo/hack/test/e2e/envfuncs"
 )
 
 func GetFuncs() ([]env.Func, []env.Func) {
```
thomastaylor312, line 17: I just noticed this now. Could you give a doc comment documenting which each slice of funcs is for?

hack/test/e2e/framework/utils/git_fixtures.go
```diff
@@ -0,0 +1,169 @@
...
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+
+	"github.com/akuity/kargo/hack/test/e2e/envfuncs"
+	"github.com/akuity/kargo/pkg/gitprovider"
+
+	_ "github.com/akuity/kargo/pkg/gitprovider/azure"           // Azure provider registration
+	_ "github.com/akuity/kargo/pkg/gitprovider/bitbucket/cloud" // Bitbucket Cloud provider registration
+	_ "github.com/akuity/kargo/pkg/gitprovider/gitea"           // Gitea provider registration
+	_ "github.com/akuity/kargo/pkg/gitprovider/github"          // GitHub provider registration
+	_ "github.com/akuity/kargo/pkg/gitprovider/gitlab"          // GitLab provider registration
```
thomastaylor312: Nit: Just put a single comment over this saying this are for the provider registrations

hack/test/e2e/envs/home_config.yaml
hairyhum: Not realy. `envfuncs.GetEnv` is reading from this config, not from env vars.

hack/test/e2e/framework/envfuncs/kargo_cli.go
hairyhum: We could, but there's not much in that config at the moment to require that.
I'm trying to avoid typing the test config for now, since it's parsed into a map on the top level.

hack/test/e2e/framework/envfuncs/load_env.go
hairyhum: I tried that, but it looks worse, since I can't read the "path" from there when I scan the code. Plus it makes `optionalString` signature look weird.

thomastaylor312 (approved):

PR #6805, opened 2026-08-13
feat: create PromotionRequests for target-aware Stages

jsbroks:
> [!NOTE]
> Stacked on #6829, which nests the Stage's Target selector under `spec.targets`. Review only the top three commits.

## Description

Connects `PromotionRequest` to every place that creates a Promotion. A Stage that selects Targets promotes Freight to them rather than to itself, so the Stage controller's auto-promotion loop, `promote-to-stage`, and `promote-downstream` now create a PromotionRequest instead of a Promotion when the Stage is target-aware.

- `api.NewPromotionRequest(ctx, c, stage, freight)` resolves the Stage's target selector to concrete Targets once and records them in `spec.targets`, sets the Stage as controlling owner, and stamps stage + shard labels. The list is a snapshot and is never recomputed, so a request's threshold is computed against a fixed set.
- The server endpoints create PromotionRequests through the **internal client**: users hold no PromotionRequest permissions by design, and the endpoints' existing `promote`-verb check on the Stage is the authorization decision. Both call sites document the justification per the internal-client policy and are covered by tests.
- Auto-promotion stands down whenever any PromotionRequest for the (Stage, Freight) exists, in any phase — Stage status doesn't track requests yet, so a stricter guard than the Promotion path's is the only one that can't loop. Backed by a new `PromotionRequestsByStageAndFreight` index.
- Promotion by origin is refused (400) for target-aware Stages: nothing resolves an origin for a PromotionRequest, and falling back to a Promotion would silently bypass the Stage's Targets.
- `GenerateChildPromotionName` establishes the `<stage>.<target>.<ulid>.<hash>` naming contract for the child Promotions the reconciler will create.
- `promote-downstream` returns created requests under a separate `promotionRequests` key; `promote-to-stage` returns the PromotionRequest as its 201 body.

Deliberately **not** in this PR:

- Keeping `spec.targetSelectors` in sync with the Stage's own as they change (follow-up; until then a request reflects the Stage's selectors as of creation).
- Reflecting PromotionRequest aggregation data in Stage status.
- The fan-out reconciler itself.

## Validation

- `make lint-go`
- `go test -race ./pkg/api/... ./pkg/indexer/... ./pkg/server/... ./pkg/controller/stages/...`
- `make test-chart`

### DIFF
--- a/pkg/controller/stages/regular_stages.go
+++ b/pkg/controller/stages/regular_stages.go
@@ -1966,7 +1966,7 @@ func (r *RegularStageReconciler) autoPromoteFreight(
 		}
 
 		if api.IsTargetAware(stage) {
-			if err = r.autoPromoteToTargets(ctx, stage, &candidate, origin); err != nil {
+			if err = r.createAutoPromotionRequest(ctx, stage, &candidate, origin); err != nil {
 				return newStatus, err
 			}
 			continue
@@ -2067,10 +2067,10 @@ func (r *RegularStageReconciler) autoPromoteFreight(
 	return newStatus, nil
 }
 
-// autoPromoteToTargets creates a PromotionRequest expressing the intent to
-// promote the candidate Freight to every Target that the target-aware Stage
-// governs. Resolving the Stage's target selectors to Targets is the
-// PromotionRequest reconciler's job, not this controller's.
+// createAutoPromotionRequest creates a PromotionRequest expressing the intent
+// to promote the candidate Freight to the Targets that the target-aware Stage
+// governs. Those Targets are resolved once, as the request is built, and
+// recorded on it; the request does not promote anything itself.
 //
 // The guard against duplicate work here is deliberately stricter than the one
 // autoPromoteFreight applies to Promotions: a PromotionRequest is created only when
@@ -2079,7 +2079,11 @@ func (r *RegularStageReconciler) autoPromoteFreight(
 // "succeeded, but the outcome is not yet recorded in status" to reason about --
 // and absent a guard that holds unconditionally, every reconcile would create
 // another PromotionRequest.
-func (r *RegularStageReconciler) autoPromoteToTargets(
+//
+// The guard is confined to auto-promotion. Promoting the same Freight to the
+// same Stage again deliberately -- rolling back to it, say -- goes through the
+// API server, which creates a PromotionRequest unconditionally.
+func (r *RegularStageReconciler) createAutoPromotionRequest(
 	ctx context.Context,
 	stage *kargoapi.Stage,
 	candidate *kargoapi.Freight,
--- a/pkg/controller/stages/regular_stages.go
+++ b/pkg/controller/stages/regular_stages.go
@@ -154,6 +154,19 @@ func (r *RegularStageReconciler) SetupWithManager(
 		)
 	}
 
+	// This index is used to determine if a PromotionRequest already exists for a
+	// Stage and Freight combination.
+	if err := sharedIndexer.IndexField(
+		ctx,
+		&kargoapi.PromotionRequest{},
+		indexer.PromotionRequestsByStageAndFreightField,
+		indexer.PromotionRequestsByStageAndFreight,
+	); err != nil {
+		return fmt.Errorf(
+			"error setting up index for PromotionRequests by Stage and Freight: %w", err,
+		)
+	}
+
 	// This index is used to find Freight that are directly available from a
 	// Warehouse and can be automatically promoted to a Stage.
 	if err := sharedIndexer.IndexField(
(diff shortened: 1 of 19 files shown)


pkg/controller/stages/regular_stages.go
```diff
@@ -2038,6 +2058,94 @@ func (r *RegularStageReconciler) autoPromoteFreight(
...
+// PromotionRequest reconciler's job, not this controller's.
+//
+// The guard against duplicate work here is deliberately stricter than the one
+// autoPromoteFreight applies to Promotions: a PromotionRequest is created only when
+// no PromotionRequest for this Stage and Freight exists at all, in any phase. Stage
+// status does not yet track PromotionRequests, so there is no equivalent of
+// "succeeded, but the outcome is not yet recorded in status" to reason about --
+// and absent a guard that holds unconditionally, every reconcile would create
+// another PromotionRequest.
+func (r *RegularStageReconciler) autoPromoteToTargets(
```
thomastaylor312: Nit: I think this function is misnamed. You're not auto promoting targets, right? Just promoting. Or am I misunderstanding how this is used?

pkg/controller/stages/regular_stages.go
```diff
@@ -2038,6 +2058,94 @@ func (r *RegularStageReconciler) autoPromoteFreight(
...
+			)
+			return nil
+		}
+		return fmt.Errorf(
+			"error creating PromotionRequest for Freight %q in namespace %q: %w",
+			candidate.Name, stage.Namespace, err,
+		)
+	}
+
+	// No event is recorded. Kargo's promotion events carry a Promotion, and a
```
thomastaylor312, line 2134: This is fine for now, but we probably should add in events later

pkg/controller/stages/regular_stages.go
```diff
@@ -2038,6 +2058,94 @@ func (r *RegularStageReconciler) autoPromoteFreight(
...
 }
 
+// autoPromoteToTargets creates a PromotionRequest expressing the intent to
+// promote the candidate Freight to every Target that the target-aware Stage
+// governs. Resolving the Stage's target selectors to Targets is the
+// PromotionRequest reconciler's job, not this controller's.
+//
+// The guard against duplicate work here is deliberately stricter than the one
+// autoPromoteFreight applies to Promotions: a PromotionRequest is created only when
+// no PromotionRequest for this Stage and Freight exists at all, in any phase. Stage
```
thomastaylor312, line 2077: What about when someone is repromoting (i.e. rolling back) to a previous piece of freight?

pkg/server/promote_to_stage.go
```diff
@@ -125,6 +127,21 @@ func (s *server) promoteToStage(c *gin.Context) {
 	}
 
 	if req.Origin != "" {
+		if api.IsTargetAware(stage) {
+			// Promotion by origin relies on the Promotion defaulting webhook to
+			// resolve the origin to a candidate Freight at admission time.
+			// PromotionRequests have no such webhook, so there is nothing to resolve
+			// the origin. Refuse rather than silently create a Promotion that
+			// bypasses the Stage's Targets.
+			_ = c.Error(libhttp.ErrorStr(
```
thomastaylor312, line 136: Can you make sure we capture this as a follow up? We should have a defaulting webhook for this. Fine for now though

thomastaylor312 (approved):
A few minor things. Nothing blocking

pkg/controller/stages/regular_stages.go
jsbroks: Good catch, renamed to `createAutoPromotionRequest` — it creates a PromotionRequest naming the Targets, the reconciler does the promoting. Also fixed the stale doc comment that still claimed selector resolution happened in the reconciler.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

pkg/controller/stages/regular_stages.go
jsbroks: Agreed — captured as a follow-up under P4 on the rollout plan, no code change here.

pkg/controller/stages/regular_stages.go
jsbroks: Re-promotion still works: this stand-down guard only exists in the auto-promotion loop, and the API server path creates unconditionally, so a manual rollback to earlier Freight gets a fresh PromotionRequest (names embed a ULID, so no collision). Added a test for exactly that case and noted the confinement in the doc comment, since the guard reads as though it would block a repeat.

pkg/server/promote_to_stage.go
jsbroks: Captured under P4 on the rollout plan — a defaulting webhook that resolves the Stage's selectors into `spec.targets` at admission, and resolves origin for target-aware Stages so this 400 can go away.

PR #6807, opened 2026-08-13
feat(ui): show how long each promotion step took

### LINKED ISSUE #6806: UI: show how long each promotion step took
### Pre-submission checklist

### Proposed Feature

Show how long each individual step of a promotion took, in the promotion view in the UI.

Today the UI shows a duration for the promotion **as a whole** — `promotion.tsx` computes it from `status.startedAt` and `status.finishedAt` and renders it as a `duration` field once the promotion is terminal. Individual steps, however, show only their status. When a promotion takes a long time there is no way to tell *which* step was responsible without inspecting the `Promotion` resource by hand.

The data required is already there. `PromotionStatus.stepExecutionMetadata[]` carries `startedAt` and `finishedAt` for every step (`api/v1alpha1/promotion_types.go`), and both fields are already present on the generated `StepExecutionMetadata` TypeScript model that the UI consumes (`ui/src/gen/api/v2/models/stepExecutionMetadata.ts`). So this needs no API, CRD or codegen change — it is purely a presentation gap.

Suggested behaviour:

- **Completed step** — show the elapsed time, i.e. `finishedAt - startedAt`.
- **Running step** — show a live elapsed time counting up from `startedAt`. This is arguably the most valuable case, since it tells you a step is stuck rather than merely slow.
- **Not yet started** — show nothing.
- Reuse the formatting already used for the promotion-level duration so the two read consistently.

### Motivation and Use Case

Promotions routinely contain steps whose durations differ by orders of magnitude. `git-clone` and `yaml-update` complete in milliseconds, while `argocd-update` waits on a sync and `git-wait-for-pr` can wait hours for a human to merge. When a promotion is slower than expected, the first question is always "which step is taking the time?" and the UI cannot currently answer it.

Concrete situations where this matters:

1. **A promotion that normally takes 30 seconds suddenly takes 10 minutes.** Right now you must run `kubectl get promotion <name> -o yaml` and subtract timestamps by hand across steps to find the culprit.
2. **A promotion appears hung.** A live elapsed time on the running step immediately distinguishes "waiting 20 seconds on a sync" from "waiting 3 hours on a PR nobody has merged".
3. **Tuning `retry.timeout`.** Choosing a sensible timeout for a polling step requires knowing how long that step actually takes in practice.
4. **Non-engineers watching a promotion.** Per the discussion in #5871, people who are not comfortable in `kubectl` are increasingly the ones watching promotions in the UI, and inspecting raw resources is not an option for them.

### Alternatives Considered

- **Reading the `Promotion` resource directly** (via `kubectl` or the UI's YAML view). This works and is what people do today, but it requires manually subtracting RFC 3339 timestamps for each step, is impractical while a promotion is still running, and is out of reach for users without cluster access.
- **Showing only the total promotion duration.** Already implemented, but it answers "was this slow?" and not "what was slow?", which is the actionable question.
- **Surfacing timings only in logs/metrics.** Useful for operators in aggregate, but it does not help someone looking at a specific promotion in the UI at the moment it matters.

Related presentation work on the same view, for context: #5871 (step naming and grouping), #5034 (step failure message placement), #5033 (collapsing long step lists).

utkarshcmu:
## Description

Closes #6806

The promotion view shows a duration for the promotion as a whole, but individual steps show only their status. When a promotion is slow there is no way to tell *which* step accounted for it without inspecting the `Promotion` resource by hand.

This renders an elapsed time against each step:

- **Finished step** — measured from `startedAt` to `finishedAt`.
- **Running step** — counts up from `startedAt` on a one second tick, which is what distinguishes a step that is stuck from one that is merely slow.
- **Not yet started** — renders nothing.

The data was already present in `status.stepExecutionMetadata` and already exposed on the generated `StepExecutionMetadata` model, so there is **no API, CRD or codegen change** — this is purely presentational.

### Completed promotion

`git-clone` took 2s; the two sub-second steps read `<1s`.

completed promotion with per-step durations

### Running promotion — before / after

Before, a running promotion showed no timing at all (note the promotion-level `duration` row is absent entirely until the promotion is terminal):

before

After, the running step reports how long it has been waiting:

after

### Live counter

The elapsed time on a running step ticks, then settles on its final value when the step reaches a terminal state:

live counter ticking then finishing

## Notes for reviewers

**`<1s` rather than `0ms`.** API timestamps are RFC 3339 with second precision, so every sub-second step computes to exactly `0`. Rendering that as `0ms` would imply a precision the data does not have, so a zero duration renders as `<1s`. Non-zero sub-second values are still formatted exactly, in case finer-grained timestamps arrive later.

**Clock skew.** If the browser clock is behind the control plane, a running step's elapsed time would go negative; `stepDurationMs` returns `null` in that case so nothing is rendered rather than showing `-3s`.

**Timer scope.** The one second interval is owned by a small `StepDuration` component and only runs while a step is actually running, so finished steps do not hold a timer.

**Out of scope.** While testing this I noticed the promotion drawer does not refetch while it is open, so a step reaching a terminal state is only reflected after the view is reopened (the elapsed time itself ticks client-side regardless). That is pre-existing behaviour and untouched here — happy to open a separate issue for it.

Tested against a real promotion in a local kind cluster, including a step left running for 20+ minutes.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/ui/src/features/stage/promotion-step-duration.tsx
+++ b/ui/src/features/stage/promotion-step-duration.tsx
@@ -1,0 +1,42 @@
+import { useEffect, useState } from 'react';
+
+import { StepExecutionMetadata } from '@ui/gen/api/v2/models';
+
+import { formatStepDuration, stepDurationMs } from './utils/step-duration';
+
+// A running step re-renders on this interval so its elapsed time counts up,
+// which is what distinguishes a step that is stuck from one that is merely slow.
+const TICK_INTERVAL_MS = 1000;
+
+export const StepDuration = ({ meta }: { meta?: StepExecutionMetadata }) => {
+  const running = !!meta?.startedAt && !meta?.finishedAt;
+
+  const [now, setNow] = useState(() => Date.now());
+
+  useEffect(() => {
+    if (!running) {
+      return;
+    }
+    const interval = setInterval(() => setNow(Date.now()), TICK_INTERVAL_MS);
+    return () => clearInterval(interval);
+  }, [running]);
+
+  const elapsed = stepDurationMs(meta, now);
+
+  if (elapsed === null) {
+    return null;
+  }
+
+  return (
+    <span
+      className='text-xs text-gray-500 ml-auto tabular-nums'
+      title={
+        meta?.finishedAt
+          ? `Started ${meta.startedAt}, finished ${meta.finishedAt}`
+          : `Started ${meta?.startedAt}`
+      }
+    >
+      {formatStepDuration(elapsed)}
+    </span>
+  );
+};
--- a/ui/src/features/stage/promotion-step-duration.tsx
+++ b/ui/src/features/stage/promotion-step-duration.tsx
@@ -1,27 +1,23 @@
 import { useEffect, useState } from 'react';
 
-import { StepExecutionMetadata } from '@ui/gen/api/v2/models';
+import {
+  getPromotionDirectiveStepStatus,
+  PromotionDirectiveStepStatus
+} from '@ui/features/common/promotion-directive-step-status/utils';
+import { Promotion, StepExecutionMetadata } from '@ui/gen/api/v2/models';
+import { parseDate } from '@ui/utils/dates';
 
-import { formatStepDuration, stepDurationMs } from './utils/step-duration';
+import { elapsedStepDurationMs, formatStepDuration } from './utils/step-duration';
 
-// A running step re-renders on this interval so its elapsed time counts up,
-// which is what distinguishes a step that is stuck from one that is merely slow.
 const TICK_INTERVAL_MS = 1000;
 
-export const StepDuration = ({ meta }: { meta?: StepExecutionMetadata }) => {
-  const running = !!meta?.startedAt && !meta?.finishedAt;
-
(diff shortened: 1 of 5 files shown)


Marvin9 (commented):
Pushing some refactors and bug that makes duration counter ticks continuously even though promotion is failed

ui/src/features/stage/promotion-step-duration.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { StepExecutionMetadata } from '@ui/gen/api/v2/models';
+
+import { formatStepDuration, stepDurationMs } from './utils/step-duration';
+
+// A running step re-renders on this interval so its elapsed time counts up,
+// which is what distinguishes a step that is stuck from one that is merely slow.
+const TICK_INTERVAL_MS = 1000;
+
+export const StepDuration = ({ meta }: { meta?: StepExecutionMetadata }) => {
+  const running = !!meta?.startedAt && !meta?.finishedAt;
```
Marvin9: this means that any terminated promotion will count as a running promotion

Marvin9 (approved):

PR #6814, opened 2026-08-13
feat(ui): surface PromotionRequests when promoting to target-aware Stages

jsbroks:
> [!NOTE]
> Top of the PromotionRequest stack: #6829 → #6805 → this. Only the last two commits are new here.

## Description

Makes drag-and-drop promotion onto a target-aware Stage work end to end in the UI, and shows what became of the resulting PromotionRequest.

Before this, dragging Freight onto a target-aware Stage left the user with nothing to look at:

- the promote drawer's `onSuccess` navigated to the **Promotion** detail route using the returned **PromotionRequest**'s name, landing on a page for an object that does not exist;
- the Stage's Promotions tab lists Promotions only, so the request was invisible;
- and there was no REST endpoint for PromotionRequests in the first place.

### Back end

- `GET /v1beta1/projects/{project}/promotion-requests` — list, optional `?stage=` filter, watch support. Modelled on `listPromotions`, including the in-process stage filtering that the uncached watch-seed reader requires.
- `GET /v1beta1/projects/{project}/promotion-requests/{promotion-request}` — get + watch.
- **Project users get read-only access** (`get`/`list`/`watch`) to PromotionRequests. Still no create/update/delete for anyone — the Stage controller and API server own writes.

### Front end

- The promote drawer routes to the Stage when the Stage is target-aware, instead of to a Promotion that will never exist.
- The Promotions tab lists that Stage's PromotionRequests above the Promotions table: phase, Freight, and either the resolved Targets or a note that the selectors are unresolved.
- A `Ready=False` condition is surfaced as a warning banner with its message, so "nothing happened" arrives with its reason.
- `isStageTargetAware` mirrors `api.IsTargetAware`, testing for a *present* selector list rather than a non-empty one.

## A note on the RBAC change

#6789 deliberately gave users **no** access to PromotionRequests, so the resource would be inert. This PR softens that to read-only, on the grounds that "inert" should mean a user cannot *drive* the feature, not that they cannot *see why their own promotion did nothing*. Writes remain closed, and `kubectl edit promotionrequest` is still denied.

## Validation

- `make lint-go`, `make test-chart`, `make typecheck-ui`, `pnpm --dir=ui lint`
- `go test ./pkg/server/... ./pkg/controller/management/projects/...`
- New endpoint tests cover empty, populated, name ordering, and stage filtering.

Manual: drag Freight onto a target-aware Stage in the pipeline view, confirm the drawer closes to the Stage and the Promotions tab shows the request with its status.


thomastaylor312 (approved):
The backend changes are good. Have a UI person take a look at this once you move out of draft and they can give final approval

jsbroks:
@

jsbroks:
Closing to many conflicts now. Will make a new one

Issue #6623, opened 2026-07-15
git-merge-pr: opt-in flag to attempt merge when mergeable_state is blocked, for bypass-authorized actors


TMigue:
### Pre-submission checklist

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I understand that submitting a feature request does not guarantee it will be implemented.

### Proposed Feature

An opt-in field on the `git-merge-pr` step config (naming open to suggestion, maybe `force`). When true, treat GitHub's blocked mergeable state as "attempt the merge" instead of "not ready, retry," and let the caller's actual permissions, including any branch-ruleset bypass, decide whether it goes through.

Scoped narrow on purpose:

- Only blocked changes. dirty (real conflicts) still hard-fails immediately, no change there.
- behind, draft, and unknown stay untouched. Those are correctly still "wait and retry" and have nothing to do with review bypass.
- Default is false, so nothing changes for anyone who doesn't set it.

### Motivation and Use Case

We run Kargo's promotion task against GitOps repos where `main` requires an approving review and CODEOWNERS review, enforced through a GitHub ruleset. Kargo's GitHub App identity is listed as a ruleset bypass actor, GitHub's own mechanism for letting specific apps or teams skip PR rules that apply to everyone else. Not every app in our org uses Kargo as its promotion pipeline, so we can't just turn the review requirement off for the whole repo. Bypass keeps it scoped to Kargo's identity only, while humans and other tooling stay fully gated.

Concretely: a promotion PR opens, no human has reviewed it yet, GitHub reports `mergeable_state: blocked`. Kargo's own permissions (via the bypass grant) would let the merge succeed if attempted, but since #6490, the step returns "not ready" before ever calling the merge API, so it waits indefinitely instead.

Before #6490, `MergePullRequest` gated only on the Mergeable boolean, which ignores required reviews and status checks entirely, so the merge call was always attempted and GitHub decided server-side whether it succeeded based on the caller's permissions. That's what let our bypass-authorized identity merge these PRs right away. #6490 introduced a better gate based on `mergeable_state`, which is the right call, but it also removed that path as a side effect while fixing the real race condition in #5761.

Anyone using Kargo as an automated GitOps pipeline behind a branch ruleset that names Kargo's identity as a bypass actor would hit this.

### Alternatives Considered

[...]

PR #6820, opened 2026-08-14
fix(github): Fixes regression where bypass rules are respected for GitHub

### LINKED ISSUE #6811: git-merge-pr: transient provider errors bypass configured step retries
### Pre-submission checklist

### Description

`git-merge-pr` treats every error returned by the git provider's `MergePullRequest` call as terminal. That includes clearly transient failures such as provider 5xx responses and network timeouts.

The step returns `Failed` with a `TerminalError`, so its configured `retry.errorThreshold` and `retry.timeout` do not apply. `wait: true` does not change this behavior.

https://github.com/akuity/kargo/blob/v1.11.0/pkg/promotion/runner/builtin/git_pr_merger.go#L128-L140

```go
mergedPR, merged, err := gitProv.MergePullRequest(...)
if err != nil {
    return promotion.StepResult{
        Status: kargoapi.PromotionStepStatusFailed,
    }, &promotion.TerminalError{...}
}
```

`wait: true` only affects the `err == nil && !merged` path, which returns `Running` with `RetryAfter`. For errors returned by `MergePullRequest`, there is no retryable path.

Downstream, `Failed` with a `TerminalError` short-circuits before the retry branch:

- https://github.com/akuity/kargo/blob/v1.11.0/pkg/promotion/local_orchestrator.go#L287-L315
- https://github.com/akuity/kargo/blob/v1.11.0/pkg/controller/promotions/promotions.go#L293-L297

The Promotion therefore fails on the first provider error, even with a retry configuration such as:

```yaml
retry:
  errorThreshold: 10
  timeout: 1h
```

This is the same error-classification boundary exposed by #6397 and #5761. #6490 addressed those GitHub-specific cases using two complementary mechanisms:

- a `mergeable_state` pre-gate for known transient PR states
- a narrow carve-out for the `405 Base branch was modified` race

State-based classification cannot cover provider unavailability, though. A 5xx or network timeout can occur during the initial PR read, before mergeability is known:

https://github.com/akuity/kargo/blob/v1.11.0/pkg/gitprovider/github/github.go#L328-L331

It can also occur during the merge request or the follow-up PR read after GitHub has already merged the PR:

https://github.com/akuity/kargo/blob/v1.11.0/pkg/gitprovider/github/github.go#L369-L418

The post-merge case is especially awkward because the Promotion can fail even though the PR was successfully merged. Retrying would be safe because the next initial read recognizes that the PR has already been merged.

I do not think every provider error should be retried. Authentication failures, invalid PRs, and PRs that are closed without being merged can remain terminal. This seems to need classification based on the provider error, so clearly transient failures can use the step's existing retry policy.

### Steps to Reproduce

1. Configure a `git-merge-pr` step with a retry policy:

```yaml
- uses: git-merge-pr
  as: merge-rendered-pr
  retry:
    errorThreshold: 10
    timeout: 1h
  config:
    repoURL: <repo>
    prNumber: ${{ task.outputs['open-rendered-pr'].pr.id }}
    mergeMethod: squash
    wait: true
```

2. Have the provider API return a 5xx or network error during the initial PR read, merge request, or post-merge PR read. In our case this happened during a GitHub degradation affecting one repository.
3. Run a Promotion.

### Expected Behavior

A clearly transient provider failure returns `Errored`, allowing the configured `retry.errorThreshold` and `retry.timeout` to apply.

`wait: true` can remain specific to PRs that are valid but not yet ready to merge. Provider-specific conditions that mean "not ready" may continue returning `Running` with `RetryAfter`.

A brief provider outage should not immediately fail a Promotion with an hour of retry time configured for the affected step.

### Actual Behavior

The step fails terminally on the first provider error.

In our incident, the Promotion failed 21 seconds after the PR was opened, during an approximately nine-minute window in which the same repository was also rejecting git pushes with `Internal Server Error`.

The PR itself was valid: `mergeable` was `MERGEABLE`, `mergeStateStatus` was `CLEAN`, and every required check passed a few minutes later. The Promotion was already terminal and was never retried. The PR and its `kargo/promotion/*` branch were left behind and had to be cleaned up manually.

### Kargo Version

v1.11.0

### Relevant Logs

Concurrent failures against the same repository during the same window, corroborating a GitHub-side degradation:

```text
ERROR promotions/promotions.go:431 error executing Promotion: error running step
"<git-push-step>": error pushing commits to remote: error pushing branch:
error executing cmd [/usr/bin/git push origin HEAD:refs/heads/kargo/promotion/<promotion>]:
remote: Internal Server Error
remote: Time 2026-08-13T14:41:57Z
To https://github.com/<org>/<rendered-repo>.git
 ! [remote rejected] HEAD -> kargo/promotion/<promotion> (Internal Server Error)
```

Those `git-push` steps retried and recovered because their operational errors are classified as `Errored`. The `git-merge-pr` step during the same window did not.

thomastaylor312:
## Description

The change in #6490 made it so that we didn't even attempt to merge when it showed that merging was blocked. Before, we would still attempt to merge and let GitHub block the merge if the user wasn't authorized. This restores that behavior for PRs in the blocked and behind states.

This is the default behavior requested in #6623, so this has the knock on effect of closing that issue as well

### Quality

I manually validated this against a live repo to ensure that merges would still be blocked appropriately

### AI Use Disclosure

This PR was written:

### DIFF
--- a/pkg/gitprovider/github/github.go
+++ b/pkg/gitprovider/github/github.go
@@ -52,12 +52,28 @@ const (
 	mergeableStateUnknown = "unknown"
 )
 
-// mergeBaseModifiedMsg is the substring GitHub includes in the message of a 405
-// merge error when the base branch moved between the mergeability check and the
-// merge call. This is the one transient 405 (it clears on retry); all other 405s
-// from the merge endpoint are permanent. Full message: "Base branch was
-// modified. Review and try the merge again."
-const mergeBaseModifiedMsg = "Base branch was modified"
+// Substrings GitHub includes in the message of a 405 from the merge endpoint.
+// GitHub uses that one status code for every reason a merge cannot be
+// performed, so the message is the only way to tell a condition that may clear
+// on its own from one that never will.
+const (
+	// mergeBaseModifiedMsg appears when the base branch moved between the
+	// mergeability check and the merge call. Full message: "Base branch was
+	// modified. Review and try the merge again."
+	mergeBaseModifiedMsg = "Base branch was modified"
+	// mergeNotMergeableMsg is GitHub's generic rejection when the PR's state does
+	// not permit a merge -- most often an unsatisfied branch protection rule.
+	// Full message: "Pull Request is not mergeable"
+	mergeNotMergeableMsg = "Pull Request is not mergeable"
+	// mergeNotAllowedMsg marks a merge method the repository will not accept.
+	// GitHub words this several ways -- "Squash merges are not allowed on this
+	// repository", "Merge commits are not allowed on this repository.", "Rebase
+	// merges are not allowed", and, where a ruleset restricts the methods
+	// permitted for a branch, wording that names the selected method -- so match
+	// only the phrase they share. No transient rejection contains it as far as we
+	// can tell.
+	mergeNotAllowedMsg = "not allowed"
+)
 
 var registration = gitprovider.Registration{
 	Predicate: func(repoURL string) bool {
@@ -219,7 +235,8 @@ func (p *provider) CreatePullRequest(
 	if opts == nil {
 		opts = &gitprovider.CreatePullRequestOpts{}
 	}
-	ghPR, _, err := p.client.CreatePullRequest(ctx,
+	ghPR, _, err := p.client.CreatePullRequest(
+		ctx,
 		p.owner,
 		p.repo,
 		&github.NewPullRequest{
@@ -238,7 +255,8 @@ func (p *provider) CreatePullRequest(
 	}
 	pr := convertGithubPR(*ghPR)
 	if len(opts.Labels) > 0 {
-		_, _, err = p.client.AddLabelsToIssue(ctx,
+		_, _, err = p.client.AddLabelsToIssue(
+			ctx,
 			p.owner,
 			p.repo,
 			int(pr.Number),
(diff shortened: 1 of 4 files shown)


pkg/gitprovider/github/github.go
```diff
@@ -419,6 +447,39 @@ func (p *provider) MergePullRequest(
...
+		return false
+	case strings.Contains(msg, mergeBaseModifiedMsg):
+		// The base tip moved between the mergeability check and the merge call: a
+		// TOCTOU race that fires when concurrent merges land on the same base.
+		return true
+	case strings.Contains(msg, mergeNotMergeableMsg):
+		// GitHub's generic rejection, reported when its own view of the PR's state
+		// does not permit a merge. That state is recomputed as checks report and
+		// reviews land, so it is worth another attempt.
+		return true
```
fuskovic, line 477: Shouldn't this be gated on `policyBlocked`? Otherwise a clean-state PR treats a new conflict the same as a pending CI check and we retry forever instead of surfacing the error.

fuskovic: Better yet, I think we could just delete this case entirely

fuskovic (changes requested):

pkg/gitprovider/github/github.go
thomastaylor312: I'm pretty sure I manually tested for this and it didn't, but let me double check

thomastaylor312: I do think it is still useful to have because it could be a transitive state that we can retry. I just added an additional test and some more documentation on this case to explain how it works. Let me know what you think

fuskovic (approved):
LGTM

PR #6826, opened 2026-08-17
feat(api)!: PromotionRequest names its Targets instead of selecting them

jsbroks:
## Description

Replaces `PromotionRequest`'s `spec.targetSelectors` with `spec.targets`, an atomic list of Target names.

The Stage's target selectors are now evaluated **once, at creation**, and the result recorded on the request. A PromotionRequest's membership is therefore a snapshot of what the Stage governed at that moment, rather than a query that could match differently later — so its threshold and terminal state are computed against a fixed, inspectable set.

```yaml
spec:
  stage: production        # immutable
  freight: 01jexample...   # immutable
  targets:                 # mutable
  - name: us-east
  - name: us-west
```

### Mutability

`spec.targets` is the **one mutable field** in the spec. The governing Stage owns it and may add Targets to a request that is still in flight, so a Target discovered after creation can still receive the Freight — without reintroducing a selector whose meaning drifts underneath the request.

`spec.stage` and `spec.freight` keep their `self == oldSelf` transition rules. A request represents exactly one freight transition for its whole lifetime, which is what gives it a terminal state to verify against. The generated CRD reflects this precisely: transition rules on `stage` and `freight` only, `targets` atomic with none.

### Why atomic rather than a list-map

A list-map keyed on `name` makes the API server record per-item ownership in `managedFields`, storing each Target name a second time as a JSON-escaped key — roughly doubling the storage cost of every entry, which matters at fleet scale. Uniqueness of Target names moves to the validating webhook, where it costs one pass instead of an `O(n²)` CEL rule over exactly the lists this change exists to keep affordable.

## Notes for reviewers

- **This is a breaking change to an unreleased type.** `PromotionRequest` landed in #6789 and nothing creates one yet, so there is no migration concern.
- **Conflicts with two open PRs.** #6805's `api.NewPromotionRequest` copies `stage.spec.targetSelectors` onto the request and will not compile against this; it needs to resolve selectors to Target names instead (either in the constructor, or in a defaulting webhook). #6814 is stacked on #6805 and inherits the same. Both want rebasing onto this.
- **`status.targets` is unchanged** and is now purely per-Target progress. Worth a separate look: per-Target progress in status means `O(N)` writes to an `O(N)`-sized object, and the alternative — aggregate counts in status plus a label on the child Promotions — may scale better.
- **`TestPromotionRequestSpec_Immutability`** scans the type's kubebuilder markers via the Go AST, in the same style as `TestKubernetesNamePattern` in `regexp_test.go`. Unit tests use a fake client, which does not evaluate CEL, so this guards the markers that generate the rules; the rules themselves are exercised against a real API server.

## Validation

- `make lint-go`, `make test-chart` (153 tests)
- `go test ./pkg/controller/promotionrequests/...`, `go test ./v1alpha1/...` from `api/`
- Regenerated CRD, deepcopy, and UI schema via `make codegen-controller` + `pnpm run generate:schema`
- Manually verified against a live cluster that `spec.targets` accepts additions while `spec.stage` and `spec.freight` are rejected with `failed rule: self == oldSelf`

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:


api/v1alpha1/promotion_request_types.go
```diff
@@ -97,21 +97,41 @@ type PromotionRequestSpec struct {
...
+	// evaluated once, at creation, and the result recorded here. The membership
+	// of the PromotionRequest is therefore a snapshot of what the Stage governed
+	// at that moment, and its threshold and terminal state are computed against
+	// it rather than against a selector that could match differently later.
+	//
+	// This is the only mutable field in the spec. The governing Stage owns it,
+	// and may add Targets to an in-flight PromotionRequest so that Targets
+	// discovered after creation can still receive the Freight. Target names MUST
+	// be unique; this is enforced by the validating webhook rather than by the
+	// schema, since a list-map's per-item ownership tracking would roughly
```
fuskovic, line 116: > enforced by the validating webhook rather than by the schema

I don't see the validating webhook in the diff

fuskovic (changes requested):

api/v1alpha1/promotion_request_types.go
jsbroks: Im doing stacked PRs, it will be in a follow up

thomastaylor312 (approved):
Approving this. I had the same question as @fuskovic, but figured it was in a follow up

fuskovic (approved):

PR #6828, opened 2026-08-17
feat(webhook): validate PromotionRequests at admission

jsbroks:
> [!NOTE]
> Stacked on #6829, which nests the Stage's Target selector under `spec.targets`. Review only the top two commits.

## Description

Adds a validating webhook for `PromotionRequest`, covering three things the schema cannot express.

### Target names must be unique

`spec.targets`' doc comment already promises this:

> Target names MUST be unique; this is enforced by the validating webhook rather than by the schema…

but nothing enforced it until now. The field is an atomic list rather than a list-map precisely to avoid the per-item ownership tracking a list-map records in `managedFields`, which roughly doubles the storage cost of every entry — a real cost on a fleet request naming thousands of Targets. A CEL rule is no better: expressing uniqueness costs `O(n²)` over exactly the lists the atomic list exists to keep affordable. One pass in the webhook costs `O(n)`.

### The Stage must exist and must select Targets

A PromotionRequest promotes Freight to a Stage's Targets. A classic Stage — one that promotes to itself — has nothing for a PromotionRequest to do, so a request naming one would be reconciled forever without effect.

### Every named Target must exist in the Project

Done with a single `List` rather than a `Get` per Target, since this runs on every admission and a fleet request may name thousands.

**Deliberately absent:** any check that a Target still *matches* the Stage's selectors. `spec.targets` is a snapshot of what the Stage governed when the request was created, so a Target relabelled afterwards must not retroactively invalidate a request already in flight. There's a test for that.

## Notes for reviewers

- **Validation runs on update as well as create.** `spec.targets` is mutable — the governing Stage may add Targets to an in-flight request — so an addition gets the same scrutiny as the original list. `spec.stage` and `spec.freight` can't change; the schema's transition rules reject that before the webhook runs.
- **The webhooks server needed a new grant.** It now reads Targets, which weren't in its ClusterRole.
- **`stage.Spec.TargetSelectors == nil` is inlined** rather than using `api.IsTargetAware`, which lives on #6805. Worth converging on the helper once both have landed.
- **Not in this PR:** the defaulting webhook. Resolving the Stage's selectors into `spec.targets` at admission, and resolving promote-by-origin for target-aware Stages (currently a 400), are tracked separately — @thomastaylor312 asked for the latter as a follow-up on #6805.

## Validation

- `make lint-go`, `make test-chart`, `go test -race ./pkg/webhook/...`
- Rendered the chart and confirmed the webhook registration and the `targets` grant both appear
- Tests cover: project missing, Stage missing, Stage not target-aware, duplicate names, Target missing, Target in another Project, empty Targets list, a relabelled Target still validating, and update-path additions

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

### DIFF
--- a/charts/kargo/templates/kubernetes-webhooks-server/cluster-role.yaml
+++ b/charts/kargo/templates/kubernetes-webhooks-server/cluster-role.yaml
@@ -50,6 +50,7 @@ rules:
   - projectconfigs
   - promotiontasks
   - stages
+  - targets
   - warehouses
   verbs:
   - get
--- a/charts/kargo/templates/kubernetes-webhooks-server/webhooks.yaml
+++ b/charts/kargo/templates/kubernetes-webhooks-server/webhooks.yaml
@@ -210,6 +210,24 @@ webhooks:
     resources: ["promotions"]
     operations: ["CREATE", "UPDATE", "DELETE"]
   failurePolicy: Fail
+- name: promotionrequest.kargo.akuity.io
+  admissionReviewVersions: [ "v1" ]
+  sideEffects: None
+  clientConfig:
+    service:
+      namespace: {{ .Release.Namespace }}
+      name: kargo-webhooks-server
+      path: /validate-kargo-akuity-io-v1alpha1-promotionrequest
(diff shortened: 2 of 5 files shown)


fuskovic (approved):
LGTM

PR #6615, opened 2026-07-14
feat(api): add TargetSelector to StageSpec for enhanced governance

jsbroks:
## Description

Adds `Stage.spec.targetSelector`, building on the `Target` CRD introduced in #6608.

- `api/v1alpha1/stage_types.go` — adds a `TargetSelector *metav1.LabelSelector` field to `StageSpec`. It selects the `Target`s that a Stage governs and promotes Freight to, matching by label within the Stage's own Project. When nil (the default), the Stage operates in classic mode, governing a single implicit "stage-self" `Target` maintained by the controller, preserving the behavior of Stages authored before `Target`s existed. A non-nil but empty selector is distinct from nil and selects all `Target`s in the Project.

Everything else in the diff is regenerated (`make codegen` output): deepcopy, protobuf (`generated.pb.go`/`generated.proto`), the CRD manifest, UI bindings/schema, and the API reference doc.

### Eligibility

- [ ] Linked to an existing issue with no blocking labels (`kind/proposal`, `needs discussion`, `needs research`, `maintainer only`, `area/security`, `size/large`, `size/x-large`, `size/xx-large`).
- [ ] Changes documentation only.
- [ ] Changes ten lines or fewer.

### Quality

- [ ] Adds or updates corresponding tests.
- [ ] Adds or updates corresponding documentation.

### AI Use Disclosure

This PR was written:

- [ ] By a human _without_ AI assistance.
- [ ] By a human _with_ AI assistance. A human has reviewed every line prior to opening the PR.
- [ ] By an AI _with human supervision._ A human has reviewed every line prior to opening the PR.
- [ ] Entirely by an AI. No human has reviewed this prior to opening the PR.

### Sign-Off

All commits:

- [ ] Are signed off by their author (`git commit -s`) **(required)**
- [ ] Are cryptographically signed (`git commit -S`) **(encouraged)**

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #6829, opened 2026-08-17
refactor(api)!: nest a Stage's Target selectors under spec.targets

jsbroks:
Replaces `Stage`'s `spec.targetSelectors` with a `spec.targets` block:

```yaml
# before
spec:
  targetSelectors:
  - matchLabels:
      environment: prod

# after
spec:
  targets:
    selectors:
    - matchLabels:
        environment: prod
```

### Why nest it

`targetSelectors` was the first of what will be several fleet-level knobs. Concurrency limits and ordering are the obvious next ones, and adding them at the top of `StageSpec` means `targetSelectors`, `targetConcurrency`, `targetStrategy` — a prefix doing the work a struct should do. A `targets` block gives them somewhere to go.

Nesting also improves the target-awareness test. It was `spec.targetSelectors != nil`, which forced the awkward rule that a present-but-empty list means something different from a nil one. Now the block's presence is the switch, and an empty selector list simply governs no Targets at the moment — nothing subtle left to explain.

### What is unchanged

The selectors themselves. Still a list, still OR'd together, and a Target matching more than one is still governed once. An empty selector still selects every Target in the Project.

`selectors` is required inside the block, so a `targets` block always says which Targets it governs.

### Compatibility

This changes the `Stage` schema, but the field it changes has never shipped. `spec.targetSelectors` merged in #6615 on 2026-07-15, after `release-1.11` was cut; it is absent from `release-1.11`, from `v1.11.1`, and from every tag. No released Kargo understands the field, so no user's `Stage` can have it set and no conversion is needed.

Nothing else in the tree read the field on `main`. The stacked PRs that do (#6828, #6805, #6814) are rebased onto this branch.


jessesuen (approved):
I like this better

PR #6830, opened 2026-08-17
refactor(controller): drop the PromotionRequest reconciler's injectable seam

jsbroks:
`SetupReconcilerWithManager` took a `ReconcileFn` so the reconciler's behavior could be swapped out later:

```go
type ReconcileFn func(ctx, kubeClient, promotionRequest) (ctrl.Result, error)

func SetupReconcilerWithManager(ctx, mgr, cfg, reconcileFn ReconcileFn) error
```

Only one function was ever passed. And working through what a reconciler that actually fans a `PromotionRequest` out into `Promotion`s would need, the seam does not fit it.

### Why a function is the wrong extension point here

A fan-out reconciler must watch the child `Promotion`s it creates. A `PromotionRequest`'s spec barely changes after creation — `spec.stage` and `spec.freight` are immutable and only `spec.targets` may grow — so nearly every reconcile after the first is triggered by a child changing phase. That watch is an `Owns(&kargoapi.Promotion{})` on the controller builder, and this package builds the controller with `For` alone. **A swappable `Reconcile` body cannot contribute a watch.**

Two smaller ones point the same way:

- **An uncached reader.** The seam hands over `mgr.GetClient()`. Listing owned `Promotion`s through the informer cache immediately after creating them reads as "this Target has no Promotion yet" and creates a duplicate, so fan-out needs `mgr.GetAPIReader()`.
- **Fan-out concurrency.** A request naming thousands of Targets should not hand the promotion engine the entire fleet in one pass. That is a knob about the fan-out, and it does not belong in this package.

Each would have to be smuggled in by closure, or added here on behalf of a caller this package does not have. At that point the caller is not supplying a function, it is supplying a controller through a keyhole.

### What changes

The reconciler becomes concrete: it records its status on the `PromotionRequest` and nothing else. Fan-out, when it is implemented, is its own controller with its own watches — a smaller commitment than an extension point that only appears to fit, and it leaves nothing here to be misread as an integration contract.

- `ReconcileFn` and the exported `DefaultReconcile` are gone; the body moves into `reconciler.Reconcile`.
- `SetupReconcilerWithManager` loses its fourth parameter.
- Net −57 lines.

No behavior change: `DefaultReconcile` was the only `ReconcileFn` ever passed, and its logic is unchanged — same phase, same condition, same idempotence (a second reconcile produces an empty merge patch and no write). The test asserting that is unchanged; only the subtest exercising the delegation itself is removed.

### DIFF
--- a/cmd/controlplane/controller.go
+++ b/cmd/controlplane/controller.go
@@ -477,7 +477,6 @@ func (o *controllerOptions) setupReconcilers(
 		ctx,
 		kargoMgr,
 		promotionrequests.ReconcilerConfigFromEnv(),
-		promotionrequests.DefaultReconcile,
 	); err != nil {
 		return fmt.Errorf("error setting up PromotionRequests reconciler: %w", err)
 	}
--- a/pkg/controller/promotionrequests/promotionrequests.go
+++ b/pkg/controller/promotionrequests/promotionrequests.go
@@ -60,13 +60,11 @@ func ReconcilerConfigFromEnv() ReconcilerConfig {
 }
 
 // SetupReconcilerWithManager initializes the PromotionRequest reconciler and
-// registers it with the provided Manager. The behavior that fans a
-// PromotionRequest out into Promotions is supplied by reconcileFn.
+// registers it with the provided Manager.
 func SetupReconcilerWithManager(
 	ctx context.Context,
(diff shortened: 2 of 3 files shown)


fuskovic (approved):
LGTM

PR #6831, opened 2026-08-17
fix(health): Fixes pointless status write for argocd health

thomastaylor312:
When in heavy load situations (with thousands of applications), we were causing status write updates on _any_ timestamp change, which happens quite often with Argo. This causes basically a consistent storm of updates to a stage's status. This drops those timestamps from the stage state. This doesn't affect the health checks because those are read from the event and not from stage status

### DIFF
--- a/pkg/health/checker/builtin/argocd.go
+++ b/pkg/health/checker/builtin/argocd.go
@@ -197,7 +197,15 @@ func (a *argocdChecker) getApplicationHealth(
 	}
 
 	// Reflect the health and sync status of the Argo CD Application.
-	appStatus.ApplicationStatus = app.Status
+	//
+	// The status is sanitized because it is persisted verbatim in Stage status,
+	// which is written with a diff-based patch: any field that Argo CD ticks on
+	// every refresh (most notably reconciledAt) would otherwise make every
+	// periodic reconciliation of every Stage produce a non-empty patch -- and a
+	// watch event -- even when nothing meaningful changed. The health
+	// assessment below reads such fields from the live App, not from this
+	// sanitized copy.
+	appStatus.ApplicationStatus = sanitizeAppStatus(app.Status)
 
 	// Argo CD has separate reconciliation loops for operations (like syncing) and
 	// assessing App health. This means that immediately following a completed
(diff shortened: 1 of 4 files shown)


jsbroks (approved):

pkg/health/checker/builtin/argocd.go
```diff
@@ -270,6 +278,23 @@ func (a *argocdChecker) getApplicationHealth(
...
+// Application remains safe to read (and to mutate, e.g. by a refresh
+// request) after the sanitized copy has been taken.
+func sanitizeAppStatus(status argocd.ApplicationStatus) argocd.ApplicationStatus {
+	sanitized := *status.DeepCopy()
+	sanitized.ReconciledAt = nil
+	for i := range sanitized.Conditions {
+		sanitized.Conditions[i].LastTransitionTime = nil
+	}
+	return sanitized
+}
```
jessesuen, line 296: I understand that the stage reconciler will no longer persist a non-meaningful change to our own Stage status. This alone will be enough to solve the issue.

However, I wonder if we could also add a second approach where non-meaningful updates to the Argo CD Application never enqueue the Stage to be reconciled in the first place. In other words, if when noticing a change to the Application, if the old and new version differ *only* by reconciledAt, but not Application health or other meaningful changes, then we never enqueue the Stage to be reconciled in the first place. This would act as the primary line of defense (not reconciling the Stage for unmeaningful updates), and your current change in this PR would act as a second line of defense (by preventing a mutation on the Stage itself by avoiding storing unmeaningful data)

jessesuen: Hmmm, I think maybe the improvement I am thinking of is already implemented:
https://github.com/akuity/kargo/blob/main/pkg/controller/stages/event_handlers.go#L522-L560

jessesuen: Your original statement made me believe that we would enqueue Stages for reconciliation even if nothing but the timestamp changed:

> When in heavy load situations (with thousands of applications), we were causing status write updates on any timestamp change, which happens quite often with Argo

But based on the implementation of `appHealthOrSyncStatusChanged` as well as the corresponding `reconciledAt changed but was already fresh` test case, it seems like something else must be causing the Stage to reconcile. So I wonder if there is still a bug somewhere. The bug might be that something else is causing stages to reconcile unnecessarily / very frequently. And the update to the `reconciledAt` in the Stage's status is just a side effect of that hot loop.

jessesuen: In any case, this change is needed and LGTM.

jessesuen (approved):

pkg/health/checker/builtin/argocd.go
thomastaylor312: I was seeing a bunch of things when I looked into this that was just showing this data getting written. I’ll do some more digging on this shortly

thomastaylor312: So I double checked myself here. I was just a bit imprecise. You are correct that it isn't triggering on reconcile. What this helped with was during _resync_. When resync occurred every 5 minutes, it was reading the status (which would still update with the new timestamp, even if it didn't trigger a reconcile). If the timestamp had updated, when resync happen, it would patch _every single stage_ that had a timestamp update, even if nothing had changed. So that is what is really fixed here

jessesuen: Got it. Makes sense. Thanks

PR #6835, opened 2026-08-18
fix(api): order Promotions by the ULID in their name, not the whole name

jsbroks:
`ComparePromotionByPhaseAndCreationTime` says it orders Promotions *"based on the ULID in the Promotion name"*, but compares the names whole:

```go
```

Those are equivalent only while everything to the left of the ULID is identical. That held until a Promotion could be named after the Target it promotes to (#6834), and it does not hold now.

## The failure is request interleaving, not just Target order

A Stage sorts **every** Promotion it owns — across every `PromotionRequest` — and takes `promotions.Items[0]` as its `currentPromotion`. Two children of one Stage now diverge at the Target segment, so the comparison resolves there and never reaches the ULID.

Request A is created first, request B second. All of A's ULIDs precede all of B's:

```
A: tier-2.us-east.<ulidA>.aaaa111     B: tier-2.us-east.<ulidB>.bbbb444
A: tier-2.us-west.<ulidA>.aaaa111     B: tier-2.us-west.<ulidB>.bbbb444
```

Sorted by whole name:

```
tier-2.us-east.<ulidA>   ← A
tier-2.us-east.<ulidB>   ← B
tier-2.us-west.<ulidA>   ← A
tier-2.us-west.<ulidB>   ← B
```

A's `us-east` runs and finishes; the next non-terminal name is **B's** `us-east`. The Stage begins the newer request while the older one still has `us-west` outstanding. Under the older `<stage>.<ulid>.<hash>` naming every one of A's names preceded every one of B's, so A drained completely before B started — a property that was load-bearing and silently lost.

## The fix

Compare the embedded ULID. Both generated formats place it in the second-to-last dot-separated segment:

```
<stage>.<ulid>.<short-hash>
<stage>.<target>.<ulid>.<short-hash>
```

Locating it **from the end** is what makes this reliable: `KubernetesName` permits dots, so a Stage or Target called `foo.bar` adds segments on the left, while the short hash never contains one. A name with no parseable ULID — hand-written, or predating generated names — falls back to comparing whole names, exactly as before, rather than comparing equal and leaving the sort unstable.

## Tests

`TestComparePromotionByPhaseAndCreationTime_doesNotInterleaveRequests` builds two requests' children, sorts them together, and asserts **every** child of the older request precedes **every** child of the newer. Checking only `promos[0]` would pass while interleaved, so it walks the whole slice.

Verified the test fails against the old comparator before it passed against the new one:

```
--- FAIL: TestComparePromotionByPhaseAndCreationTime_doesNotInterleaveRequests
    "tier-2.us-west.01m0b5sbr54v1dp08z0kz2hc14.aaaa111" belongs to the older
    request but sorts after a child of the newer one; the Stage would start the
    newer request before the older one finished
```

Plus table tests for `promotionNameULID` (both formats, dotted Stage/Target names, hand-written names, right-length non-ULIDs) and for the fallback path.

## Notes

Strictly this was always a latent mismatch between what the function said and what it did; nothing exercised it until names stopped sharing a prefix. Fixing the comparator rather than the naming keeps the Target visible in child names — which is what makes a fan-out greppable — and leaves the ULID authoritative for ordering, as the doc comment always claimed.

### DIFF
--- a/pkg/api/promotion.go
+++ b/pkg/api/promotion.go
@@ -224,14 +224,68 @@ func ComparePromotionByPhaseAndCreationTime(a, b kargoapi.Promotion) int {
 		// Non-terminal Promotions are ordered in ascending order based on the
 		// ULID in the Promotion name. This ensures that the Promotion which
 		// was (or will be) enqueued first is at the top.
-		return strings.Compare(a.Name, b.Name)
+		return comparePromotionNamesByULID(a.Name, b.Name)
 	default:
 		// Terminal Promotions are ordered in descending order based on the
 		// ULID in the Promotion name. This ensures that the most recent
 		// Promotion is at the top, limiting the number of Promotions which
 		// have to be further inspected to collect the "new" Promotions.
-		return strings.Compare(b.Name, a.Name)
+		return comparePromotionNamesByULID(b.Name, a.Name)
 	}
+}
+
+// comparePromotionNamesByULID orders two generated Promotion names by the ULID
+// they embed, rather than by comparing the names whole.
+//
+// Comparing whole names is only equivalent to comparing ULIDs while everything
+// to the left of the ULID is identical, which stopped being true once a
+// Promotion could be named after the Target it promotes to. Two Promotions of
+// the same Stage then differ at the Target segment, the comparison resolves
+// there, and it never reaches the ULID at all -- so an older Promotion to
+// "us-west" would sort after a newer one to "us-east", and a Stage would start
+// promoting to the second while the first was still outstanding.
+//
+// Both generated name formats put the ULID in the second-to-last
+// dot-separated segment:
+//
+//	<stage>.<ulid>.<short-hash>
+//	<stage>.<target>.<ulid>.<short-hash>
+//
+// Indexing from the end is what makes that reliable: a Kubernetes name may
+// contain dots, so a Stage or Target named "foo.bar" adds segments on the left,
+// while the short hash never contains one.
+//
+// Names that do not carry a parseable ULID -- a Promotion created before Kargo
+// generated names, or one named by hand -- fall back to comparing whole names,
+// preserving the previous behavior rather than declaring them equal and leaving
+// the sort unstable.
+func comparePromotionNamesByULID(a, b string) int {
+	aULID, aOK := promotionNameULID(a)
+	bULID, bOK := promotionNameULID(b)
+	if !aOK || !bOK {
+		return strings.Compare(a, b)
+	}
+	return strings.Compare(aULID, bULID)
+}
+
+// promotionNameULID returns the ULID embedded in a generated Promotion name,
+// and whether one was found. The returned value keeps the case it appears in:
+// generated names are lowercased, and lowercasing a ULID preserves ordering,
+// since Crockford base32 digits sort before its letters in ASCII either way.
+func promotionNameULID(name string) (string, bool) {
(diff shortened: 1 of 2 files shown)


pkg/api/promotion.go
```diff
@@ -224,16 +224,70 @@ func ComparePromotionByPhaseAndCreationTime(a, b kargoapi.Promotion) int {
...
+// preserving the previous behavior rather than declaring them equal and leaving
+// the sort unstable.
+func comparePromotionNamesByULID(a, b string) int {
+	aULID, aOK := promotionNameULID(a)
+	bULID, bOK := promotionNameULID(b)
+	if !aOK || !bOK {
+		return strings.Compare(a, b)
+	}
+	return strings.Compare(aULID, bULID)
+}
```
fuskovic, line 295: I think there is a bug here.

Mixing

`strings.Compare(a, b)` (used when either name lacks a ULID) 

and 

`strings.Compare(aULID, bULID)` (used when both have one)

across pairs breaks transitivity and `slices.SortFunc` (called from `pkg/controller/stages/regular_stages.go`) has undefined behavior with a non-transitive comparator, so a Promotions list containing even one legacy/hand-named Promotion alongside generated ones could sort inconsistently. 

Consider the following:

A = "b.7ZZZZZZZZZZZZZZZZZZZZZZZZZ.hash1234" (Stage b, max-value ULID)
B = "m" (no ULID segment — legacy/hand-created)
C = "z.00000000000000000000000000.hash1234" (Stage z, min-value ULID)
A vs B → falls back, 'b' < 'm' → A < B
B vs C → falls back, 'm' < 'z' → B < C
A vs C → both have ULIDs, compares ULIDs only → '7' > '0' → A > C

A < B < C but A > C (not valid). 

We should do some TDD here to validate this.

The fix would be:

```go
func comparePromotionNamesByULID(a, b string) int {
	aULID, aOK := promotionNameULID(a)
	bULID, bOK := promotionNameULID(b)
	switch {
	case aOK && bOK:
		return strings.Compare(aULID, bULID)
	case aOK != bOK:
		// Exactly one name carries a ULID. A ULID and an arbitrary string
		// aren't on the same comparison axis, so bucket by "has a ULID"
		// rather than falling through to a lexical compare. Mixing the
		// two axes is what breaks transitivity. Legacy/hand-named
		// Promotions sort before generated ones.
		if aOK {
			return 1
		}
		return -1
	default:
		return strings.Compare(a, b)
	}
}
```

So A < B < C can no longer imply A > C.

fuskovic: 2 tests actually. 1 for the mixed case e.g. assert a legacy name always sorts before a ULID name regardless of lexical order plus one for the example

fuskovic (changes requested):
In addition to my first comment there's a couple comments Claude is flagging as stale.

1. `pkg/webhook/kubernetes/promotion/webhook.go:221-227`
1. `pkg/api/promotion_request_test.go:366-371`


For the scenario where we have two promos from the same `Stage`, generated via `GenerateChildPromotionName`, differing only in `Target`, they will now always hit the `aOK && bOK` branch from the example code in my first comment which fixes the interleaving bug the comments there currently describe.

pkg/api/promotion.go
jsbroks: nice caught!

fuskovic (dismissed):

Issue #6028, opened 2026-04-01
Simpler `argocd-sync` step


jessesuen:
# Proposed Feature

Kargo's existing `argocd-update` step supports modifying the app spec of the live application, e.g. to update source revision, etc... This is in addition to it's ability to simply sync an app.

For those who are practicing a traditional GitOps process that is syncing from the tip of a branch, users may benefit from a simpler `argocd-sync` step, which simply syncs the app but does not allow updates to it.

# Motivation

`argocd-update` has been a source of confusion. You can say it does *too* much. Having a simpler primitive would make things easier for folks to understand. It would also model how the argocd CLI works in CI pipelines.

jessesuen:
@dhpup brought up a good use case for having a separate step like this. Having an `argocd-sync` step would allow us to mimic some of the features of the `argocd sync` CLI, namely, selectively syncing some (but not all) resources in the app. That reason alone makes a lot of sense to introduce this step.

KyriosGN0:
I am currently implementing Kargo in my org, having an argo-sync step is excatly a feature i was looking when desinging my PromotionTask, since i would simply like to sync the app (after the previous step is commiting the change to git)

cdemarco-drw:
This makes it harder to use the `argocd-wait` action, see https://github.com/akuity/kargo/issues/6357

jessesuen:
In addition to partial sync, this step could add support for sync options, such as the option to prune resources

PR #6837, opened 2026-08-19
feat(argocd): add prune option to argocd-update step

kamilkrampa:
## What

Adds an optional per-app boolean `prune` to the `argocd-update` promotion step. When `true`, the sync operation Kargo initiates for the Argo CD `Application` sets `operation.sync.prune`, so resources that are no longer defined in the Application's sources are deleted during the sync. Defaults to `false`, preserving existing behavior.

```yaml
- uses: argocd-update
  config:
    apps:
      - name: my-app
        prune: true
```

## Why

Argo CD's `Prune` flag is not a sync option: it exists only as a field on the sync operation itself (or as `syncPolicy.automated.prune` for auto-sync). The step already copies `syncPolicy.syncOptions` onto the operation it creates, but that can never enable pruning — so for users who keep auto-sync off so that Kargo drives all deploys, Kargo-initiated syncs can never prune, and resources removed from manifests linger indefinitely.


jessesuen:
The team has been feeling for some time that `argocd-update` is getting overloaded. There are other sync options (such as selective sync) that warrant [a separate `argocd-sync` step](https://github.com/akuity/kargo/issues/6028). My preference is not to add these types of features to argocd-update and put them in a new argocd-sync instead.

PR #6840, opened 2026-08-19
feat(webhook): reject manual Promotions on target-aware Stages

jsbroks:
A Stage promotes Freight either to itself or to the Targets it governs, never both. Every in-tree caller already honors that split — the promote-to-stage and promote-downstream endpoints and the auto-promotion loop all create a PromotionRequest for a target-aware Stage — but a client creating the Promotion itself could still promote such a Stage to a destination it does not have.

This closes that path at admission, and with it establishes the invariant the rest of the Target work can rely on: **a Promotion naming a Target is always a child of a PromotionRequest, and a target-aware Stage is only ever promoted to through one.**

## What the webhook now rejects

- A Promotion with `spec.target` set that is not controlled by a PromotionRequest (reported on `spec.target`).
- A Promotion against a target-aware Stage (`spec.targets != nil`) that is not controlled by a PromotionRequest (reported on `spec.stage`, worded to mirror the message the PromotionRequest webhook already gives for the opposite case).

## Notes on the design

**Ownership is the signal for "child."** It is the one a caller cannot usefully forge: Kubernetes garbage collection deletes an object whose controller owner does not exist, so a reference to a PromotionRequest that was never created takes the Promotion with it. Defaulting and validation now share `isPromotionRequestChild`, so the owner reference `Default` preserves is exactly the one validation accepts.

**The Target check also runs on update.** Spec immutability already pins `spec.target`, but the owner reference that justifies it is metadata and is not covered — an apply whose manifest simply omits `ownerReferences` could otherwise orphan a child and leave it promoting to a Target.

**The target-aware Stage check is create-only, deliberately.** A Stage can gain a `targets` block while Promotions created before it are still running, and those must remain updatable: aborting one is an update, and refusing it would leave a running Promotion with no way to stop.

**No selector re-check.** This does not verify that the Stage currently governs the named Target. That name came from the snapshot the PromotionRequest took at creation, and re-resolving selectors here would let a label edit invalidate a Promotion already in flight — the same reasoning that keeps the PromotionRequest webhook from re-checking its own `spec.targets`.

## Testing

Five new cases (three on create, two on update). `go test -race` passes for the webhook package and for `pkg/api`, `pkg/server`, `pkg/controller/stages`, and `pkg/controller/promotionrequests`; `make lint-go` is clean. I also confirmed the new tests are not vacuous by temporarily forcing `isPromotionRequestChild` to return true — all three rejection cases fail without the check.

Docs are unchanged: `docs/docs/` does not cover PromotionRequests or Targets yet.

### DIFF
--- a/pkg/webhook/kubernetes/promotion/webhook.go
+++ b/pkg/webhook/kubernetes/promotion/webhook.go
@@ -343,13 +343,27 @@ func (w *webhook) Default(ctx context.Context, promo *kargoapi.Promotion) error 
 	// Stage-based garbage collection entirely, whereas a PromotionRequest is
 	// itself owned by its Stage, so deleting the Stage still cascades to the
 	// Promotion through it.
-	if owner := metav1.GetControllerOf(promo); owner == nil ||
-		owner.APIVersion != kargoapi.GroupVersion.String() ||
-		owner.Kind != promotionRequestKind {
+	if !isPromotionRequestChild(promo) {
 		ownerRef := metav1.NewControllerRef(stage, kargoapi.GroupVersion.WithKind("Stage"))
 		promo.OwnerReferences = []metav1.OwnerReference{*ownerRef}
 	}
 	return nil
+}
+
+// isPromotionRequestChild returns true if the Promotion was created by a
+// PromotionRequest as one part of that request's fan-out to Targets, rather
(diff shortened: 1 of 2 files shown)


fuskovic (approved):
lgtm

PR #6842, opened 2026-08-19
feat(api): add `currentPromotionRequest `and `latestPromotionRequest`

jsbroks:
## Summary

Types + codegen only, no writers. Nothing sets these fields yet; an unwritten
field is harmless, which is what lets this land ahead of everything that uses
it.

- `PromotionRequestReference` (name, freight, phase, finishedAt), hung off
  `StageStatus` as `currentPromotionRequest` and `lastPromotionRequest` --
  mirroring what `currentPromotion` / `lastPromotion` already do for
  Promotions. A reader of a Stage can see which round of fan-out it is in, and
  how the last one ended, without listing PromotionRequests.

The writer of these fields is the Enterprise fan-out reconciler (E1 on the EE
side), which is why they carry the same
"Kargo Enterprise only: This field is ignored in Kargo OSS" note as
`promotionWindowStatus`.

## Note on freight

Unlike `PromotionReference.freight` (a full `FreightReference` snapshot --
commits, images, charts, artifacts), `freight` here is a
`PromotionRequestFreightReference` carrying only the Freight's name:

```yaml
status:
  currentPromotionRequest:
    name: test.01jxyz.abc1234
    phase: Promoting
    freight:
      name: abc1234def5678
```

Freight names are deterministic content hashes, so the name alone is enough to
look the rest up. Keeping the field an object means more fields can be added
compatibly as they are needed, whereas starting with the full snapshot and
narrowing later would be a breaking schema change. This also keeps ~350 lines
of duplicated FreightReference schema out of the Stages CRD.

## Testing

- `make codegen-controller`, `codegen-openapi`, `codegen-ui` -- all committed
  output is generated, not hand-edited
- `make lint-go` (0 issues), `make lint-ui`, `make typecheck-ui`
- `go build ./...`, `go test ./api/...`, and `go test -race` over
  `pkg/controller/stages`, `pkg/controller/promotionrequests`, `pkg/server/...`

No unit tests accompany the change: the new fields are plain struct members with
no behavior, and their deepcopy methods are generated.


fuskovic (approved):

api/v1alpha1/stage_types.go
```diff
@@ -835,6 +850,31 @@ func (r *PromotionReference) GetHealthChecks() []HealthCheckStep {
...
+	FinishedAt *metav1.Time `json:"finishedAt,omitempty"`
+}
+
+// PromotionRequestFreightReference identifies the Freight promoted by a
+// PromotionRequest. It intentionally carries only the Freight's name --
+// details about the Freight's contents can be looked up from the Freight
+// itself. Fields will be added here as they are needed.
+type PromotionRequestFreightReference struct {
+	// Name is the name of the Freight.
+	Name string `json:"name"`
```
thomastaylor312, line 875: One minor note: Why not have the standard k8s object reference here? Having the struct so we can add _other_ things in the future if needed makes sense, but object reference feels like the correct thing

jsbroks: Its so we can add more stuff later, we acutally have a `FreightReference` struct that we use for PromotionReferences. Originally I was going to use that but it adds a lot of fields to the struct

thomastaylor312: And, I just took a look one more time. Other places we use freight we only reference by name, so I withdraw my minor comment

PR #6850, opened 2026-08-20
chore(deps): Bump golang version to 1.26.7

IsMiseLeam:
## Description

Closes #

### Eligibility

### Quality

- [ ] Adds or updates corresponding tests.

### AI Use Disclosure

This PR was written:

- [x] By a human _without_ AI assistance.

### Sign-Off

All commits:

- [ ] Are signed off by their author (`git commit -s`) **(required)**

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,7 @@
 version: "2"
 run:
   # Go version to use for running golangci-lint
-  go: "1.26.7"
+  go: "1.26"
   # Build tags to be passed to all linters
   build-tags:
     - dockerhub
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,7 @@
 version: "2"
 run:
   # Go version to use for running golangci-lint
-  go: "1.26"
+  go: "1.26.7"
   # Build tags to be passed to all linters
   build-tags:
     - dockerhub
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/akuity/kargo
 
-go 1.26.7
+go 1.26.0
 
 replace (
 	github.com/akuity/kargo/api => ./api
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/akuity/kargo
 
-go 1.26.0
+go 1.26.7
 
 replace (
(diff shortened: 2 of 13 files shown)


thomastaylor312 (approved):

.golangci.yaml
```diff
@@ -1,7 +1,7 @@
 version: "2"
 run:
   # Go version to use for running golangci-lint
-  go: "1.26"
+  go: "1.26.7"
```
kencochrane: I think we can keep this as is. I'm pretty sure that if you have 1.26, it will always pull the latest one. But if we want consistency, I guess pinning to 1.26.7 is probably fine.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/akuity/kargo
 
-go 1.26.0
+go 1.26.7
```
kencochrane: We need to keep this at 1.26.0. This keeps the module broadly compatible with other developers who might have Go 1.26 installed but haven't updated to the absolute latest patch level yet.

IsMiseLeam: That's fair, I was thinking because this was for some CVEs, then we had to have it, I will remedy this now.

.golangci.yaml
Leam-Akuity: If we can, I would like this one to be 1.26.7 for the sake of maintenance later. If we can avoid having a bunch of different version strings lying around it would help us avoid missing one in an a future where we are cmd+f'ing around, looking for the right one to replace. I could see somebody coming in and looking for 1.26.7 and not finding this line. 

That said, while I have this opinion, it is loosely held, and not a hill I am willing to die on! I can "go" either way on it!

go.mod
IsMiseLeam: This has been changed back to 1.26.0

.golangci.yaml
Leam-Akuity: @kencochrane - Actually never mind - these changes are failing linting and unit tests - I think partially because of misaligned versions, I will look at this again first thing. No need to review just yet.

IsMiseLeam: I had initially commented that I disagreed with this, but on further reflection, I think you are right. I reverted this back to 1.26.

PR #6868, opened 2026-08-24
feat(ui): surface subscription names in the UI

### LINKED ISSUE #6729: Surface subscription names in the UI
## Background

As of #6207, a `Warehouse`'s subscriptions may each be given a name:

```yaml
spec:
  subscriptions:
  - name: frontend
    image:
```

The name is recorded on the results of artifact discovery and on the artifact references of the `Freight` produced from them, so it is available anywhere the UI already has a subscription or an artifact reference in hand. Nothing in the UI uses it today.

## Proposal

Wherever the UI currently identifies a subscription or an artifact by repository URL, prefer the subscription's name when one is available and fall back to the URL when it isn't. Names are short and chosen by the user, so they should read better than a URL in most places -- pipeline graph nodes, warehouse details, the freight timeline, and artifact listings are all candidates. Which of those are worth changing, and how, is a judgment call best left to @rpelczar or @Marvin9.

It is also worth considering whether the create-warehouse wizard should offer a field for the name, since a name can currently only be set by editing YAML.

## When this shifts from nice to necessary

#6724 will permit a `Warehouse` to subscribe to the same repository more than once, distinguished only by name. At that point a repository URL will no longer identify a subscription or its artifacts uniquely, and any UI element that labels or groups by URL alone will be ambiguous -- the freight assembly view being the clearest example. Using names where they are available is a prerequisite for that, not merely a readability improvement.

Marvin9:
Closes #6729
Screenshot 2026-08-25 at 3 45 24 PM
Screenshot 2026-08-25 at 3 45 03 PM
Screenshot 2026-08-25 at 3 44 55 PM
Screenshot 2026-08-25 at 3 44 29 PM
Screenshot 2026-08-25 at 3 43 55 PM

Shows a subscription's name wherever the UI identifies a subscription or the artifacts it produced.


rpelczar (commented):
I think using tags for names isn't the best idea. Especially when one is clickable and the next one is not.

I think we should replace the current name taken from the URL with the new name field. It should fall back to the previous value.

In the table, we can just add a new column.

Marvin9:
Done @rpelczar

rpelczar (approved):
LGTM

PR #6869, opened 2026-08-24
chore: go version bump for ci and releases

fuskovic:
follow up for https://github.com/akuity/kargo/pull/6860 per https://github.com/akuity/kargo/pull/6860#issuecomment-5397229479

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -143,6 +143,30 @@ linters:
           - goconst
           - gosec
         path: (.+)_test\.go
+      # goconst findings surfaced by the golangci-lint v2.13.1 upgrade
+      # (bumped for Go 1.27 support). Left as follow-up cleanup rather than
+      # bundling it with an unrelated Go version bump.
+      - linters:
+          - goconst
+        path: pkg/cli/cmd/get/.*\.go
+      - linters:
+          - goconst
+        path: pkg/controller/management/projects/projects\.go
+      - linters:
+          - goconst
+        path: pkg/controller/warehouses/warehouses\.go
+      - linters:
+          - goconst
+        path: pkg/promotion/runner/builtin/(git_pr_opener|git_pr_waiter|helm_chart_updater|json_updater|kustomize_image_setter|toml_updater|yaml_merger|yaml_updater)\.go
+      - linters:
+          - goconst
+        path: pkg/server/(get_cluster_config|get_project_config|get_promotion|get_stage|get_warehouse)\.go
+      - linters:
+          - goconst
+        path: pkg/server/rbac/(generic_escalation|roles)\.go
+      - linters:
+          - goconst
+        path: pkg/webhook/external/(azure|github|refresh)\.go
     paths:
       - .*/*_types.go
       - .*/groupversion_info.go
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -53,6 +53,16 @@ linters:
       min-len: 3
       # Minimal occurrences count to trigger, 3 by default
       min-occurrences: 3
+      # Ad hoc map keys (e.g. step result output fields) don't warrant a
+      # named constant just because the same key is reused elsewhere.
+      ignore-map-keys: true
+      # Call is excluded by default; CompositeLit is added here because
+      # struct/slice literal values (Kubernetes API kinds, verbs, resource
+      # names, table column headers, etc.) are frequently repeated
+      # vocabulary rather than magic strings that need centralizing.
+      exclude-types:
+        - Call
+        - CompositeLit
 
     govet:
       enable:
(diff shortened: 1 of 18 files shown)


krancour (approved):

.golangci.yaml
```diff
@@ -143,6 +143,30 @@ linters:
...
+        path: pkg/promotion/runner/builtin/(git_pr_opener|git_pr_waiter|helm_chart_updater|json_updater|kustomize_image_setter|toml_updater|yaml_merger|yaml_updater)\.go
+      - linters:
+          - goconst
+        path: pkg/server/(get_cluster_config|get_project_config|get_promotion|get_stage|get_warehouse)\.go
+      - linters:
+          - goconst
+        path: pkg/server/rbac/(generic_escalation|roles)\.go
+      - linters:
+          - goconst
+        path: pkg/webhook/external/(azure|github|refresh)\.go
```
krancour: There's just no way we're going to enjoy keeping this up to date.

krancour: Is there some other way to relax goconst a bit without naming files?

pkg/governance/comment_handler_test.go
```diff
@@ -247,8 +247,9 @@ func Test_commentHandler_handleCreated(t *testing.T) {
 				Action: github.Ptr("created"),
 				Issue:  issue,
 				Comment: &github.IssueComment{
-					Body:              github.Ptr(testCase.body),
-					User:              &github.User{Login: github.Ptr(authorLogin)},
+					Body: github.Ptr(testCase.body),
+					User: &github.User{Login: github.Ptr(authorLogin)},
+					// nolint:staticcheck // production code still reads this field
 					AuthorAssociation: github.Ptr(testCase.authorAssoc),
```
krancour, line 253: This comment didn't really make it clear to me what's going on. I had to run the linter myself to understand that `AuthorAssociation` was deprecated.

It's strange that `GetAuthorAssociation()` isn't also deprecated. (Probably, it should have been.)

It gets stranger. The claim in the deprecation notice is:

> GitHub will remove this field from Events API payloads on October 7, 2025. Use the Issue Comments REST API endpoint to retrieve this information.

Obviously, that date has come and gone.

So...

What I would say is, first, we should research this more and find out the reality of whether payloads will continue to include this or if it could disappear any time now without further notice.

Depending on what we find, we should do one of the following:

1. Improve this comment and others like it to explain why it's safe to continue using this field.
2. Add TODO comments on the corresponding uses of `GetAuthorAssociation()` to call for the strategic resolution,
3. Actually resolve the issue as prescribed.

No.3 is probably the best and chances are good that Claude would make short work of it.

pkg/promotion/runner/builtin/tar_creator.go
```diff
@@ -254,6 +254,8 @@ func (t *tarCreator) createTarball(
 		}
 
 		if info.Mode().IsRegular() {
+			// #nosec G122 -- path is produced by WalkDir over the promotion's
+			// own already-checked-out work tree, not attacker-supplied input.
```
krancour: This might need a closer look. Two things jump out at me:

1. The path _is_ supplied by a user, so appropriate precautions need to be taken if they haven't been already.

2. "the promotion's own already-checked-out work tree" is language that seems to be coupling what's going on here to git in some way and there's really no guarantee that is the case.

It may very well be that there isn't an issue here and that the linter's analysis is wrong, but the existing comment doesn't convince me.

pkg/server/server.go
```diff
@@ -311,6 +311,8 @@ func wrapWithBasePath(inner http.Handler, basePath string) http.Handler {
 		if r.URL.Path == basePath {
 			u := *r.URL
 			u.Path = basePath + "/"
+			// #nosec G710 -- u.Path is the fixed, server-configured basePath,
+			// not attacker-controlled; only the query string passes through.
```
krancour: I'm not clear on what query string has to do with this.

Total nit: I would speak in terms of "users" rather than "attackers" because the danger in a lot of these things isn't strictly confined to cases of malicious intent.

pkg/cli/cmd/config/view.go
```diff
@@ -73,6 +73,8 @@ func (o *viewOptions) run() error {
 		cfg = config.MaskedConfig(cfg)
 	}
 
+	// #nosec G117 -- This command's purpose is to display the user's own
+	// CLI config; secrets are masked unless --raw is explicitly passed.
```
krancour: Possibly worth saying that this is not persisted or transmitted.

pkg/cli/config/config.go
```diff
@@ -133,6 +133,8 @@ func SaveCLIConfig(config CLIConfig) error {
 }
 
 func saveCLIConfig(config CLIConfig, configPath string) error {
+	// #nosec G117 -- This is the user's own CLI config file, written to their
+	// own Kargo home directory so it can be read back on later invocations.
```
krancour: Nit: Possibly worth being more explicit that this is never transmitted.

.golangci.yaml
krancour: Looking at the sort of things it's been flagging... A good example is the string literal `"Name"` as a column heading. The headings in a table are easier to read and understand as literals and as we discussed earlier, this isn't the kind of scenario where fat-fingering something will break anything.

I'm really starting to feel like goconst has incrementally become more trouble than it's worth. It's not like we don't all flag things that that really _should_ be constants in our own reviews within the team. We do it all the time, and I think we're better judges of it than goconst is.

I'm coming around to the idea of disabling that particular linter.

fuskovic:
> I'm coming around to the idea of disabling that particular linter.

I've relaxed the rule and that's certainly helped.

Take a look and let me know if that looks ok or if your preference is still to disable it entirely

.golangci.yaml
```diff
@@ -53,6 +53,16 @@ linters:
       min-len: 3
       # Minimal occurrences count to trigger, 3 by default
       min-occurrences: 3
+      # Ad hoc map keys (e.g. step result output fields) don't warrant a
+      # named constant just because the same key is reused elsewhere.
+      ignore-map-keys: true
```
krancour, line 58: fwiw, I see the value in the rule, but I also think relaxing this one rule instead of disabling goconst entirely is a fair compromise. 👍

.golangci.yaml
```diff
@@ -53,6 +53,16 @@ linters:
...
+      # Call is excluded by default; CompositeLit is added here because
+      # struct/slice literal values (Kubernetes API kinds, verbs, resource
+      # names, table column headers, etc.) are frequently repeated
+      # vocabulary rather than magic strings that need centralizing.
+      exclude-types:
+        - Call
+        - CompositeLit
```
krancour, line 65: TIL

krancour (approved):

PR #6892, opened 2026-08-27
feat(controller): Add prometheus metrics

### LINKED ISSUE #6876: Richer prometheus metrics
### Pre-submission checklist

### Proposed Feature

Kargo should expose prometheus metrics in its various controllers that can be scraped to show useful information about system/user activity. This is different than controller/operational metrics (e.g. reconciliation performance) that only a kargo developer/admin would be interested in.

Examples of the metrics for this feature would include:

## Promotions:
* count
* current phase (`Running`, `Pending`, `Successful`, etc...)
* frequency
* duration

## Promotion Steps:
* which steps are used
* duration

## Stages:
* count
* current phase

## Freight:
* count

## Projects:
* count

It seems Warehouse and Project would be natural dimensions we would want to use for the above metrics.

### Motivation and Use Case

It's been a repeated request that users would like to have more insight on promotions. If these metrics are exposed, then grafana dashboards are able to be built.

### Alternatives Considered

_No response_

thomastaylor312:
This adds an initial set of metrics to promotions and stages to give more insight into current work. I kept these fairly constrained for now, but they should be easy to extend in the future

Closes #6876

### DIFF
--- a/docs/docs/40-operator-guide/45-monitoring.md
+++ b/docs/docs/40-operator-guide/45-monitoring.md
@@ -77,9 +77,7 @@ labeled by `project`:
 
 | Metric | Type | Description |
 |--------|------|-------------|
-| `kargo_promotions_created_total` | Counter | Number of `Promotion` resources the controller has observed being created. |
-| `kargo_promotions_errored_retryable_total` | Counter | Number of retryable errors encountered while executing Promotions. A single Promotion may contribute more than once, since each failed attempt is counted. |
-| `kargo_promotions_errored_terminal_total` | Counter | Number of Promotions that reached the terminal `Errored` phase. |
+| `kargo_promotions_completed_total` | Counter | Number of Promotions that reached a terminal phase, by the phase they finished in. |
 | `kargo_promotions_non_terminal` | Gauge | Number of Promotions currently in a non-terminal phase (`Pending` or `Running`), sampled at scrape time. |
 | `kargo_promotion_duration_seconds` | Histogram | Time each Promotion spent running, from the moment it started until it reached a terminal phase. |
 | `kargo_promotion_step_duration_seconds` | Histogram | Time each individual promotion step spent running, from the moment it started until it reached a terminal status. |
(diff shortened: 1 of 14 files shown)


docs/docs/40-operator-guide/45-monitoring.md
```diff
@@ -69,6 +69,141 @@ you rename the port for a meshed cluster, keep the `http-` prefix.
...
+
+Alongside the Go runtime and controller-runtime metrics that every component
+reports, the controller exports its own metrics, all prefixed with `kargo_` and
+labeled by `project`:
+
+| Metric | Type | Description |
+|--------|------|-------------|
+| `kargo_promotions_created_total` | Counter | Number of `Promotion` resources the controller has observed being created. |
+| `kargo_promotions_errored_retryable_total` | Counter | Number of retryable errors encountered while executing Promotions. A single Promotion may contribute more than once, since each failed attempt is counted. |
+| `kargo_promotions_errored_terminal_total` | Counter | Number of Promotions that reached the terminal `Errored` phase. |
```
jessesuen: 1. Instead of two counters, one on `created` promotions and another on `errrored_terminal`, I was thinking that there could be a _single_ counter about completed (terminal) promotions, that included the phase as a label. 

Because the more useful information for users, is tracking how promotions are finishing and their result, not that they are just being created. Completion count is anyways (eventually) equal to creation count, just slightly delayed. The metric would also have the benefit of always having the exact same totals as the `kargo_promotion_duration_seconds` histogram, and so the two metrics would go hand and hand in dashboards.

```
kargo_promotions_completed_total{project="guestbook",phase="Successful"}     10
kargo_promotions_completed_total{project="guestbook",phase="Errored"}        2
kargo_promotions_completed_total{project="guestbook",phase="Failed"}         3
kargo_promotions_completed_total{project="wordpress",phase="Successful"}     3
kargo_promotions_completed_total{project="wordpress",phase="Errored"}        0
kargo_promotions_completed_total{project="wordpress",phase="Failed"}         0
```

2. What is the use case you are thinking for having a `kargo_promotions_errored_retryable_total` metric? So users will know if their promotions are being retried? I would exclude this metric if there isn't a strong need.

thomastaylor312: 1. That is easy enough to change
2. I was thinking if a user wanted to alert if some jobs were retrying a lot. If that doesn't seem too useful, I'm fine to drop

thomastaylor312: Just pushed a commit to consolidate the two of them and remove the retryable total

pkg/controller/metrics/promotions.go
```diff
@@ -0,0 +1,176 @@
...
+				Help: "Time a Promotion spent running, from the moment it started " +
+					"until it reached a terminal phase.",
+				Buckets: durationBuckets,
+				// Because of the wide range of promotion durations, we also enable Native
+				// Histograms if the Prometheus server supports them. Notes on each of the settings
+				// is in the main metrics package
+
+				NativeHistogramBucketFactor:     metrics.NativeHistogramBucketFactor,
+				NativeHistogramMaxBucketNumber:  metrics.NativeHistogramMaxBucketNumber,
+				NativeHistogramMinResetDuration: metrics.NativeHistogramMinResetDuration,
```
jessesuen, line 91: Oh cool. I actually didn't know about this. This alleviates my concerns about picking the right buckets.

jessesuen (approved):
Instrumentation to reconcilers is very clean. Thank you!

pkg/controller/metrics/promotions.go
thomastaylor312: I didn't either until I started researching what the best way was to handle this. It is a nice feature

PR #6666, opened 2026-07-20
feat: add tar promotion step

vcalmic:
## Description

~~Closes~~ Related to #1099 

Adds a new `tar` promotion step that archives a file or directory into a tar file (by default gzipped, unless configured otherwise) at a specified location.

This promotion step expands on existing OCI support. Combined with the upcoming enhancements to the `oci-push` step, contributed by @EronWright, this will enable pipelines to package local manifests into a tarball and push them to a remote registry as an OCI artifact.

However, the step is generic enough to be reused for other workflows, such as http uploads.

### Eligibility

- [x] Linked to an existing issue with no blocking labels (`kind/proposal`, `needs discussion`, `needs research`, `maintainer only`, `area/security`, `size/large`, `size/x-large`, `size/xx-large`).
- [ ] Changes documentation only.
- [ ] Changes ten lines or fewer.

### Quality

- [x] Adds or updates corresponding tests.
- [x] Adds or updates corresponding documentation.

### AI Use Disclosure

This PR was written:

- [ ] By a human _without_ AI assistance.
- [x] By a human _with_ AI assistance. A human has reviewed every line prior to opening the PR.
- [ ] By an AI _with human supervision._ A human has reviewed every line prior to opening the PR.
- [ ] Entirely by an AI. No human has reviewed this prior to opening the PR.

### Sign-Off

All commits:

- [x] Are signed off by their author (`git commit -s`) **(required)**
- [x] Are cryptographically signed (`git commit -S`) **(encouraged)**

PR #6893, opened 2026-08-28
feat: oci push local archive

### LINKED ISSUE #1099: Support OCI manifests

blakepettersson:
## Description

Closes #1099. This is a continuation of Eron's work in #6626 and @vcalmic's work in #6666 (merged).

What has changed vs the original PR:

*  Uses `os.Root` instead of `filepath-securejoin`; paths escaping workspace are rejected rather than silently contained.
* Produces a proper OCI 1.1 artifact manifest: top-level `artifactType` (mirrored onto `config.mediaType` for Flux/Helm/ORAS compatibility) and the empty-descriptor config (`application/vnd.oci.empty.v1+json`, `{}`) rather than an image config. Defaults to `application/vnd.unknown.artifact.v1` when unset.
* The artifact size limit is always enforced for `srcPath`.
* Streams the file from disk instead of buffering it in memory.
* Minor test tweaks

### Push a local archive (Flux/Argo-style OCI manifests)

Package a directory of rendered manifests and publish it as an OCI artifact for consumption by Argo CD, Flux, or other OCI-aware tooling. Provenance is recorded via standard OCI annotations.

```yaml
steps:
  - uses: tar
    config:
      inPath: ./manifests
      outPath: ./manifests.tar.gz
      gzip: true
  - uses: oci-push
    as: push
    config:
      srcPath: ./manifests.tar.gz
      destRef: ghcr.io/example/config/app:${{ ctx.promotion }}
      annotations:
        org.opencontainers.image.source: ${{ commitFrom("https://github.com/example/app.git").repoURL }}
        org.opencontainers.image.revision: ${{ commitFrom("https://github.com/example/app.git").id }}
  - uses: argocd-update
    config:
      apps:
      - name: app
        sources:
        - repoURL: oci://ghcr.io/example/config/app
          desiredRevision: ${{ outputs.push.digest }}
          updateTargetRevision: true
```

### Minimal (default media type)

`srcPath` is content-agnostic; when `mediaType` is omitted it defaults to `application/vnd.oci.image.layer.v1.tar+gzip`.

```yaml
steps:
- uses: oci-push
  config:
    srcPath: ./artifact.tar.gz
    destRef: registry.example.com/example/app:${{ ctx.promotion }}
```

### Existing copy/retag mode (unchanged)

```yaml
steps:
- uses: oci-push
  config:
    srcRef: registry.example.com/myapp@${{ imageFrom("registry.example.com/myapp").digest }}
    destRef: registry.example.com/myapp:v1.2.3
```

Beyond unit tests, manually verified by running a promotion containing steps similar to above, and verified that it works/syncs as expected in Argo CD.

### Eligibility

### Quality

### AI Use Disclosure

This PR was written:

- [x] By a human _with_ Eron's prior assistance. A human has reviewed every line prior to opening the PR.

### Sign-Off

All commits:


pkg/promotion/runner/builtin/oci_pusher.go
```diff
@@ -153,6 +148,248 @@ func (p *ociPusher) run(
...
+	// unconditionally. A negative maxArtifactSize disables the check. This
+	// mirrors the cross-repository size check on the retag path.
+	if p.maxArtifactSize >= 0 && info.Size() > p.maxArtifactSize {
+		return v1.Hash{}, kargoapi.PromotionStepStatusErrored, &promotion.TerminalError{
+			Err: fmt.Errorf(
+				"artifact size %s exceeds maximum allowed size of %s",
+				libfmt.FormatByteCount(info.Size()), libfmt.FormatByteCount(p.maxArtifactSize),
+			),
+		}
+	}
```
fuskovic: This looks like it's technically vulnerable to TOCTOU. We should move the limit check closer to the actual reaad.

fuskovic: We could make this a helper and call it from both `push` and `pushLocalFile`. Something like:

```go
func (p *ociPusher) checkSize(size int64) error {
    if p.maxArtifactSize < 0 {
        return nil
    }
    if p.maxArtifactSize == 0 {
        return &promotion.TerminalError{Err: errors.New("push is disabled")}
    }
    if size > p.maxArtifactSize {
        return &promotion.TerminalError{Err: fmt.Errorf(
            "artifact size %s exceeds maximum allowed size of %s",
            libfmt.FormatByteCount(size), libfmt.FormatByteCount(p.maxArtifactSize),
        )}
    }
    return nil
}
```

christianh814 (approved):

pkg/promotion/runner/builtin/oci_pusher.go
```diff
@@ -153,6 +148,248 @@ func (p *ociPusher) run(
...
+	configDesc, err := layerDescriptor(configLayer)
+	if err != nil {
+		return v1.Hash{}, kargoapi.PromotionStepStatusErrored,
+			fmt.Errorf("failed to describe artifact config: %w", err)
+	}
+	artifactDesc, err := layerDescriptor(artifactLayer)
+	if err != nil {
+		return v1.Hash{}, kargoapi.PromotionStepStatusErrored,
+			fmt.Errorf("failed to describe artifact layer: %w", err)
+	}
```
fuskovic: Since we already use the `go-containerregistry` dep, we can refactor this to use `partial.Descriptor`, then you can just delete the `layerDescriptor` func.

pkg/promotion/runner/builtin/oci_pusher.go
```diff
@@ -112,29 +112,24 @@ func (p *ociPusher) run(
 			}
 	}
```
fuskovic, line 113: `srcRef` is now parsed after `dstOpts` is built. `dstOpts` involves a credential lookup/ network call. Previously, both refs were validated before any credential call. 

I think we should keep it that way.

Otherwise, a malformed `srcRef` can now be masked by a retryable credential error instead of failing fast and terminally. Which could potentially result in unnecessary network calls and taking longer than needed as a result.

pkg/promotion/runner/builtin/oci_pusher.go
fuskovic, line 283: Comment says "mirrors the cross-repository size check on the retag path" but the zero-disables branch isn't actually mirrored. The retag path treats `maxArtifactSize == 0` as a hard "disabled" case. Which is checked before size is even computed. Here it's just size `> 0`, so a 0-byte file still pushes.

Also, I noticed we don't have a test for this. Looks like a good spot to verify and fix with some TDD.

fuskovic (changes requested):

blakepettersson:
Thank you for the review @fuskovic! I believe I addressed your concerns, PTAL.

pkg/promotion/runner/builtin/oci_pusher.go
```diff
@@ -153,6 +168,231 @@ func (p *ociPusher) run(
...
+	if err != nil {
+		return nil, err
+	}
+	return &fileLayer{
+		open:      open,
+		mediaType: mediaType,
+		digest:    digest,
+		size:      size,
+	}, nil
+}
```
fuskovic: The gap between when the file size is checked to when we read from it introduces a TOCTOU bug.

To solve for it, we can refactor`newFileLayer` to:

1. Take a `limit` input param ( then we can pass `p.maxArtifactSize` in for it when we call `newFileLayer` )
2. Use the in-house `LimitCopy` helper

```go
// newFileLayer reads at most limit bytes to hash and measure the file,
// returning an *intio.BodyTooLargeError if the content exceeds limit.
func newFileLayer(
	root *os.Root,
	path string,
	mediaType types.MediaType,
	limit int64,
) (*fileLayer, error) {
	open := func() (io.ReadCloser, error) { return root.Open(path) }
	f, err := open()
	if err != nil {
		return nil, err
	}
	defer f.Close()

	hasher := sha256.New()
	size, err := intio.LimitCopy(hasher, f, limit)
	if err != nil {
		return nil, err
	}
	return &fileLayer{
		open:      open,
		mediaType: mediaType,
		digest:    v1.Hash{Algorithm: "sha256", Hex: hex.EncodeToString(hasher.Sum(nil))},
		size:      size,
	}, nil
}
```

fuskovic (changes requested):

fuskovic (approved):
:ship:

PR #6894, opened 2026-08-28
test(e2e): Runnable nightly e2e tests

### LINKED ISSUE #6144: Add nightly e2e tests
With Kargo growing in complexity, we are well past time where we should have e2e tests that we can run to ensure behavior hasn't changed and to help us catch bugs in an automated way. Up to this point maintainers have relied on a cross of manual test cases we've built up for a while and, increasingly, some AI automation to help us run those. The biggest chunk of work here will be coming up with the scaffolding that allows contributors to easily add new test cases. This will entail manifests (projects, warehouses, stages, etc.) as well as common setup tasks/functions that get a connection to a k8s cluster along with cleaning up. 

For this task, we will consider it complete when we have all of said scaffolding and at least two different tests types added (to make sure it has exercised the scaffolding setup). The new e2e test should run nightly and have the option to manually trigger it. For now this should be a separate workflow, but when we gain confidence in the tests, it might be worth gating the nightly release on these tests passing.

hairyhum:
Adds a nightly schedule for e2e test workflow and workflow running fixes. Some tests still fail, but the environment is runnable, tests would be addressed separately.

Closes https://github.com/akuity/kargo/issues/6144


hairyhum:
Example run https://github.com/akuity/kargo/actions/runs/33101101167/job/98618713592

.github/workflows/e2e-nightly.yaml
```diff
@@ -23,6 +23,30 @@ jobs:
...
+    # nightly_config.yaml ships with an empty `context` section. Populate it from
+    # repository secrets so the git- and http-dependent suites can run. Each key
+    # is only written when its secret is set; suites whose context is missing
+    # skip or fail on their own.
+    - name: Configure e2e context
+      env:
+        E2E_KARGO_DEMO_GITOPS_REPO: ${{ secrets.E2E_KARGO_DEMO_GITOPS_REPO }}
+        E2E_GIT_PAT: ${{ secrets.E2E_GIT_PAT }}
+        ## TODO: We need to set up the endpoint to ebable that
+        E2E_HTTP_ENDPOINT: ${{ secrets.E2E_HTTP_ENDPOINT }}
```
thomastaylor312, line 35: Do we need to capture this as an issue so we follow up on it?

.github/workflows/e2e-nightly.yaml
```diff
@@ -35,22 +59,29 @@ jobs:
...
     # hack/bin, which is added to PATH for the subsequent steps.
     - name: Add hack/bin to PATH
-      run: echo "${GITHUB_WORKSPACE}/hack/bin" >> "${GITHUB_PATH}"
+      run: |
+        mkdir "${GITHUB_WORKSPACE}/hack/bin"
+        echo "${GITHUB_WORKSPACE}/hack/bin" >> "${GITHUB_PATH}"
 
-    - name: Install helm
-      run: make install-helm
+    - parallel:
```
thomastaylor312, line 66: Is this valid syntax? I didn't think there was a `parallel` directive in a set of steps

.github/workflows/e2e-nightly.yaml
```diff
@@ -35,22 +59,29 @@ jobs:
...
+      - name: Install helm
+        run: make install-helm
```
thomastaylor312, line 68: One note with all of these install steps (minus the kargo one). None of the tools are cached so we'd have to download them and/or compile every time. Ideally we should cache them if possible

.github/workflows/e2e-nightly.yaml
```diff
@@ -23,6 +23,30 @@ jobs:
...
+    # is only written when its secret is set; suites whose context is missing
+    # skip or fail on their own.
+    - name: Configure e2e context
+      env:
+        E2E_KARGO_DEMO_GITOPS_REPO: ${{ secrets.E2E_KARGO_DEMO_GITOPS_REPO }}
+        E2E_GIT_PAT: ${{ secrets.E2E_GIT_PAT }}
+        ## TODO: We need to set up the endpoint to ebable that
+        E2E_HTTP_ENDPOINT: ${{ secrets.E2E_HTTP_ENDPOINT }}
+      run: |
+        set -x
```
thomastaylor312, line 37: This should be removed or it will print secrets

.github/workflows/e2e-nightly.yaml
```diff
@@ -35,22 +59,29 @@ jobs:
     # The test shells out to helm, kind, kargo and argocd; install them all into
     # hack/bin, which is added to PATH for the subsequent steps.
     - name: Add hack/bin to PATH
-      run: echo "${GITHUB_WORKSPACE}/hack/bin" >> "${GITHUB_PATH}"
+      run: |
+        mkdir "${GITHUB_WORKSPACE}/hack/bin"
```
thomastaylor312, line 63: Nit: Generally safer to use `mkdir -p` here to avoid any failures

.github/workflows/e2e-nightly.yaml
```diff
@@ -60,32 +91,12 @@ jobs:
...
+      working-directory: hack/test/e2e
       run: |
-        make hack-build
+        go test -c -tags=e2e,shared ./suites/shared
 
     - name: Run e2e tests
       working-directory: hack/test/e2e
       run: |
-        go test -tags=e2e,shared ./suites/shared -timeout 120m -args -env-file=nightly_config.yaml
+        go test -tags=e2e,shared ./suites/shared -v -timeout 120m -args -env-file=nightly_config_ready.yaml
```
thomastaylor312, line 102: You're compiling the test suite on the step before this. Should you be using the `./shared.test` binary instead of the bare `go test` here?

thomastaylor312 (changes requested):
The caching improvement is optional, but I'd recommend it if we can swing it

.github/workflows/e2e-nightly.yaml
hairyhum: Yes. You can see it in the [example run](https://github.com/akuity/kargo/actions/runs/33101101167/job/98618713592)

.github/workflows/e2e-nightly.yaml
hairyhum: https://github.com/akuity/kargo/issues/7202

.github/workflows/e2e-nightly.yaml
hairyhum: Done

.github/workflows/e2e-nightly.yaml
hairyhum: Done

hairyhum:
Applied feedback fixes. New run with cache and cleanup of commands: https://github.com/akuity/kargo/actions/runs/34881065124/job/104100207881

.github/workflows/e2e-nightly.yaml
thomastaylor312: Cool. TIL

.github/workflows/e2e-nightly.yaml
```diff
@@ -29,22 +29,78 @@ jobs:
...
-        chmod +x hack/bin/argocd
+    - parallel:
+      - name: Install helm
+        id: install-helm
+        run: make install-helm
+
+      - name: Build kind
+        id: install-kind
+        if: steps.cache-kind.outputs.cache-hit != 'true'
+        run: go build -o hack/bin/kind sigs.k8s.io/kind
```
thomastaylor312, line 66: Good enough for now. We might want to version pin and use that version in the cache action. Same goes for argocd

thomastaylor312 (approved):

PR #6916, opened 2026-08-31
chore: remove deprecated createTargetBranch option

### LINKED ISSUE #5848: Remove createTargetBranch option from git-open-pr step
See #5847

fuskovic:
Closes: https://github.com/akuity/kargo/issues/5848


docs/docs/100-roadmap.md
```diff
@@ -34,7 +34,6 @@ every submission.
 The following deprecated features are scheduled for **removal** in v1.12.0:
 
 * The deprecated ConnectRPC (gRPC) API, in favor of the REST API.
-* The `createTargetBranch` option in the `git-open-pr` promotion step.
```
krancour, line 37: This is still true. Roadmap needs work, but not sure this should be removed.

krancour: LGTM after this.

docs/docs/50-user-guide/60-reference-docs/30-promotion-steps/compose-output.md
```diff
@@ -43,7 +43,6 @@ steps:
   as: open-pr
   config:
     repoURL: ${{ vars.repoURL }}
-    createTargetBranch: true
```
krancour, line 46: Yikes. We probably should have removed this from examples when it was deprecated instead of waiting until it was removed. That was probably my oversight. 😬

krancour (approved):

PR #6926, opened 2026-08-31
feat(ui): surface freight release context

### LINKED ISSUE #6812: Surface release context in Freight details and promotion views
### Pre-submission checklist

### Proposed Feature

Add a **Release Context** tab to the existing Freight details and promotion views.

For each container image, show:

- Tag and digest when available
- Standard OCI source, revision, and image creation time
- Source and commit links when they can be derived safely
- Optional organization-scoped commit metadata using this convention:
    - `<reverse-domain>.image.commit.subject`
    - `<reverse-domain>.image.commit.author`
    - `<reverse-domain>.image.commit.committer`
    - `<reverse-domain>.image.commit.created`
- Remaining annotations in their existing raw form

The organization prefix should be discovered generically. Kargo should recognize fields from at most one coherent namespace, never combine values from different organization prefixes, and continue to show unrecognized annotations without assigning them new semantics.

In the promotion view, compare the Stage's current Freight resource with the incoming Freight resource. Corresponding images should be shown side by side and identified as changed, unchanged, added, or removed.

#### Freight details

Freight Release Context

#### Promotion comparison

Promotion Release Context

### Motivation and Use Case

Kargo already records OCI and custom image annotations in Freight, but users currently need to inspect raw annotation values and manually compare incoming freight with what is already promoted.

At promotion time, users commonly need to answer:

- What source revision produced this image?
- What changed from the version currently in the Stage?
- Who authored the change?
- When were the image and commit created?
- Can I inspect the source or commit before promoting?

Surfacing this information in the existing views keeps the promotion decision inside Kargo without adding content to space-constrained pipeline nodes.

This builds on the OCI metadata work in [#2147](https://github.com/akuity/kargo/issues/2147), complements the comparison use case in [#821](https://github.com/akuity/kargo/issues/821), and follows the side-panel direction discussed in [#5300](https://github.com/akuity/kargo/issues/5300).

I have explored this behavior in a local prototype and would be happy to take a stab at implementing it if the proposed direction looks reasonable.

### Alternatives Considered

- Continue using the raw annotation view. This preserves all metadata but provides no semantic grouping or current-versus-incoming comparison.
- Rely on source-control links and comparison pages. This requires leaving Kargo and does not cover multi-image Freight or organization-owned metadata.

Z3r0Sum:
## Description

Closes #6812. Depends on #6925.

This adds a Release Context tab to Freight details and promotion views so users can inspect image release metadata without digging through annotations. The [API PR description](https://github.com/akuity/kargo/pull/6925) covers the configuration schema, override semantics, validation, and authorization model.

- Freight details show image tags and digests, source revisions and links, build times, and configured commit metadata.
- Promotion views compare current and incoming images from the same Warehouse. The shared image comparison also detects digest-only changes, while subscription-name changes preserve pairing.
- Raw annotations remain accessible. Configuration failures show a retryable warning while preserving standard OCI metadata; unsupported URLs and invalid timestamps remain readable text.

Pipeline tiles, artifact discovery, and Freight storage are unchanged. The user guide covers the new views and their configuration.

**Review scope:** This branch is stacked on #6925, so its diff against `main` currently includes the API commit. [View the UI-only changes](https://github.com/akuity/kargo/compare/9a7134cc8e0ee05f48ead47da36eb7bb7ae8de31...3197b96adb985f807e2afe609a23dfe335653da0).

**Local validation**

- All 220 frontend tests passed, along with TypeScript checking, ESLint, and production builds of the UI and documentation.
- Manual kind validation exercised the combined API/UI prototype and confirmed that configuration changes reached the promotion comparison view. The precedence scenarios are described in #6925; no promotions were triggered.

These results apply to this source snapshot, not a merge with the latest upstream `main`.

### Eligibility

This remains a draft while #6812 carries the `kind/proposal` label.

### Quality

### AI Use Disclosure

This PR was written:

### Sign-Off

All commits:

- ClusterConfig defines custom image annotation mappings for commit
  details without changing Freight discovery or storage.
- ProjectConfig replaces cluster mappings as a whole. Omitted settings
  inherit defaults, while an explicit empty object disables mappings.
- Freight readers can retrieve effective mappings without access to
  full configuration resources. Freight access is checked first.
- Custom mapping keys reject empty values, whitespace, and the reserved
  org.opencontainers namespace to preserve standard OCI meanings.
- Generated clients and editor schemas expose the configuration contract
  while preserving omitted versus explicitly empty settings.

Refs: #6812
Signed-off-by: Stephen Schmidt <stephen.s@posthog.com>


krancour:
I'd not weighed in on #6812. I know @Marvin9 did.

I don't object to what #6812 is requesting and this PR is attempting to solve, but I have a couple concerns about _how_ it's approached.

First and biggest is that this introduces new terminology. "Release" wasn't a word Kargo used anywhere prior to this. I hate to be a pedant, but how things are named matters a lot. Out of the gate, "release" feels overloaded. To you it might be the bundle of artifacts the Freight represents. Others would consider the bundles of artifacts that actually reach production to be a release, and it would be hard to argue with that. This PR locks Kargo into your definition. Not necessarily wrong -- but certainly not something to be done without some serious deliberation.

My second concern regards a tension that I don't necessarily know how to resolve, but deserves to be named. We've been incrementally working toward subscriptions being more generic, with the thought that subscription types might eventually work more similarly to promotion steps. i.e. How they're configured and what they do is their own business, opaque to the rest of Kargo. We're not there yet, but we're already headed in that direction. We introduced a subscriber interface a while ago and already got most of the commit, image, and chart subscription logic into. If that was one step forward, this PR, useful as its feature actually is, represents two steps back in that it distributes more image-specific knowledge to other parts of the system outside of the subscriber implementation that handles image subscriptions.

Please don't interpret this as any hostility toward the proposed feature. This isn't a no. But there needs to be more discussion about how this fits into Kargo in a way we're comfortable with.

Z3r0Sum:
> I'd not weighed in on #6812. I know @Marvin9 did.

Thanks @krancour!  It might be best to have the conversation in the issue given the concerns.  I'll respond over there if you don't mind.

PR #7006, opened 2026-09-01
feat(controller): promote Targets in parallel, one PromotionRequest at a time

jsbroks:
Builds on **#6843** (merged), which recorded `currentPromotionRequest` on the Stage.

## Summary

Promotions to a target-aware Stage's Targets — the children a PromotionRequest
fans out — were serialized by the same single slot that serializes the Stage's
own Promotions: `stage.status.currentPromotion` admits exactly one Promotion at
a time, so N Targets promoted strictly one after another, and, because a child's
generated name carries a Target segment ahead of the ULID, in alphabetical order
of Target name rather than creation order.

The mutex for children is now the Stage's **current PromotionRequest**, which
the Stage already records. A request creates at most one child per Target, so
admitting all of the current request's children at once lets Promotions to
distinct Targets run in parallel with no same-Target contention, while the
children of a queued request wait for the current round of fan-out to end.
Rounds remain a barrier by construction: consecutive PromotionRequests stay
serialized by the same phase-and-ULID ordering as before. No new API is needed —
`currentPromotionRequest` simply stops being a passive mirror and becomes the
admission token, the same role `currentPromotion` plays for everything else.

## The pieces

- **The Promotion reconciler's gate becomes `api.StageAwaitsPromotion`**: a
  child runs while the PromotionRequest that owns it is the Stage's current
  one; everything else gates on the Stage's own `currentPromotion` exactly as
  before. Ownership (the controller OwnerReference) is the signal, as in the
  Promotion webhook, because it is the one a caller cannot usefully forge.
- **`syncPromotions` partitions children out of the Stage's own flow.** Without
  this, a child would occupy the Stage's single Promotion slot — the very
  serialization being removed — and its terminal phases would replay into the
  Stage's `lastPromotion`, Freight history, health, and verification state,
  none of which are the Stage's business for a Target: the Stage only
  arbitrates which request is current. The `spec.target` discriminator is
  admission-enforced: the webhook rejects any Promotion that names a Target but
  is not owned by a PromotionRequest.
- **`PromotionAcknowledgedByStageHandler` now also enqueues every non-terminal
  child of a newly current PromotionRequest**, so a queued round starts the
  moment the Stage acknowledges it rather than waiting for a resync. Children
  are found by the Stage label and filtered by owner, and the handler gains the
  client this requires.

## Deliberately not done

- **Naming is untouched.** ULID-sortability of child Promotion names stops
  mattering once all of a request's children run in parallel — there is no
  ordering among them to preserve. Ordering only matters between
  PromotionRequests, whose names carry no Target segment and sort by ULID as
  before.
- **The stricter alternative** — a per-Target current-Promotion slot in Stage
  status — would also let Targets pipeline across rounds. It remains open if
  ever wanted; this is deliberately the simplest mutex that parallelizes a
  round.

## Interim state: `freightHistory` on target-aware Stages stops advancing

`FreightHistory.Record` has exactly one call site — `syncPromotions`, when a
succeeded Promotion becomes `lastPromotion` — and this PR removes children from
that path. Since a target-aware Stage's only Promotions *are* children (manual
Promotions on such Stages are rejected at admission, and auto-promotion creates
PromotionRequests), its `freightHistory` no longer advances at all after this
PR. That is deliberate: today each succeeded child records essentially the same
collection in turn — N Targets means up to N records, N health wipes, and N
rounds of Stage-level verification gating — and per-Target outcomes are not the
Stage's business. This PR fires the wrong writer; a follow-up hires the right
one, recording `freightHistory` from succeeded PromotionRequests instead.

Until that follow-up lands, on target-aware Stages `FreightHistory.Current()`
is nil (or frozen), and its consumers feel it:

- Stage-level verification cannot run or be tracked; reverify/abort return
  "stage has no current freight". Verification for Targets is meant to surface
  on the PromotionRequest side instead.
- Freight is never marked verified in a target-aware Stage, so downstream
  Stages requesting Freight from one will not see it as available.
- Auto-promotion loses its "already current" dedup signal, leaving only the
  no-request-in-flight guard between a terminal round and re-promoting the
  same Freight.
- `freightSummary` and the UI's current-Freight display are empty for
  target-aware Stages.

Ordinary Stages are untouched: non-child Promotions hit the `Record` path
exactly as before.

## Follow-up 

Promotion reconciler carries a duplicate of the old gate and
constructs `PromotionAcknowledgedByStageHandler`, whose constructor now takes a
client — the dependency bump is a compile break that forces both adaptations.
Follow-up tickets for the interim `freightHistory` state above are tracked on
the E side as well.

### DIFF
--- a/pkg/controller/promotions/watches.go
+++ b/pkg/controller/promotions/watches.go
@@ -12,6 +12,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	kargoapi "github.com/akuity/kargo/api/v1alpha1"
+	"github.com/akuity/kargo/pkg/api"
 	"github.com/akuity/kargo/pkg/controller"
 	argocd "github.com/akuity/kargo/pkg/controller/argocd/api/v1alpha1"
 	"github.com/akuity/kargo/pkg/indexer"
@@ -201,19 +202,26 @@ func (u *UpdatedArgoCDAppHandler[T]) enqueueSelectorMatches(
 }
 
 // NewPromotionAcknowledgedByStageHandler creates a new
-// PromotionAcknowledgedByStageHandler with the given shard predicate.
+// PromotionAcknowledgedByStageHandler with the given client and shard
+// predicate.
 func NewPromotionAcknowledgedByStageHandler[T any](
+	kargoClient client.Client,
 	shardPredicate controller.ResponsibleFor[kargoapi.Stage],
 ) *PromotionAcknowledgedByStageHandler[T] {
 	return &PromotionAcknowledgedByStageHandler[T]{
+		kargoClient:    kargoClient,
 		shardPredicate: shardPredicate,
 	}
 }
 
 // PromotionAcknowledgedByStageHandler is an event handler that enqueues a
 // Promotion for reconciliation when it has been acknowledged by the Stage\
-// it is for.
+// it is for. That is the Stage's current Promotion when one changes, and
+// every child of the Stage's current PromotionRequest when that changes: the
+// children of a newly current request -- at most one per Target -- are all
+// admitted at once.
 type PromotionAcknowledgedByStageHandler[T any] struct {
+	kargoClient    client.Client
 	shardPredicate controller.ResponsibleFor[kargoapi.Stage]
 }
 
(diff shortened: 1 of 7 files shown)


pkg/controller/stages/regular_stages.go
```diff
@@ -682,6 +682,23 @@ func (r *RegularStageReconciler) syncPromotions(
...
+	// fanned out to run in parallel -- and its terminal phases would replay
+	// into the Stage's Freight history.
+	var stagePromos []kargoapi.Promotion
+	for _, promo := range promotions.Items {
+		if promo.Spec.Target == "" {
+			stagePromos = append(stagePromos, promo)
+		}
+	}
+	promotions.Items = stagePromos
+
```
fuskovic: Should we be doing the same thing in `computeEffectiveAutoPromotionHolds` and `unprocessedPromotionExistsForStageFreight` ? If so, we might want to refactor into a helper to keep it DRY. A good spot for `slices.DeleteFunc` as well:

```go
func filterOutTargetPromotions(items []kargoapi.Promotion) []kargoapi.Promotion {
    return slices.DeleteFunc(items, func(p kargoapi.Promotion) bool { 
        return p.Spec.Target != "" 
    })
}
```

pkg/controller/promotions/watches.go
```diff
@@ -272,23 +280,78 @@ func (p *PromotionAcknowledgedByStageHandler[T]) Update(
...
+	wq workqueue.TypedRateLimitingInterface[reconcile.Request],
+) {
+	logger := logging.LoggerFromContext(ctx)
+
+	promotions := &kargoapi.PromotionList{}
+	if err := p.kargoClient.List(
+		ctx,
+		promotions,
+		client.InNamespace(stage.Namespace),
+		client.MatchingLabels{kargoapi.LabelKeyStage: stage.Name},
```
fuskovic:
```suggestion
client.MatchingLabels{kargoapi.LabelKeyStage: kubernetes.ShortenLabelValue(stage.Name)},

```

fuskovic (approved):
couple small things, pre-emptively LGTM'ing

PR #5995, opened 2026-03-27
fix: improve responsiveness to changing app status

krancour:
Fixes #5347

Fixes #5636

Possible alternative to #5984

cc @alexmt 

## Problem

Argo CD executes operations and assesses health in separate reconciliation
loops. This can lead to stale health status and, in turn, an overly optimistic
assessment of Stage health.

We've known about this for a long time and have gone through _many_ iterations
of attempting to solve this elegantly. A purely event-driven approach has been
elusive because there are many cases where the App's reconciliation loop writes
no changes to the App, so a watcher on Apps that enqueues related Stages for
reconciliation couldn't quite be relied upon. We ended up going through several
different iterations of a 10-second cooldown period. Last op completed less than
10 seconds ago? Throw an error. The Stage reconciler goes into progressive
backoff. It shouldn't be long before another attempt finds trustworthy health
status.

This brought a new problem along with it -- if something else is causing
frequent operations on an App (maybe auto-sync is enabled), it's possible, with
some bad luck, to get stuck in a spiral of untrusted App health status for a
_very_ long time.

## Fix

No more cooldown period. We simply look at the App's `reconciledAt` field. If
its date/time is >= the time the last operation completed, then we know health
must have been assessed at some point post-op and we can immediately trust the
App's health status.

If these conditions are NOT met, we force a hard refresh of the App so that they
_will_ be met on the next reconciliation attempt. A new and improved App watcher
enqueues Stages for reconciliation promptly after the forced refresh completes.

Issue #5708, opened 2026-02-05
Verification skipped on high-velocity pipeline


Nulluminati:
# Checklist

* [X] I've searched the issue queue to verify this is not a duplicate bug report.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `kargo version`.
* [X] I've pasted logs, if applicable.

# Description
We're experiencing an issue where verification on our staging stage is being skipped or missed during high-velocity deployments. We have a standard staging → production pipeline with ArgoCD sync steps in our promotions.

When a new piece of freight is promoted to staging, the verification step frequently does not complete (or does not run) before the next freight is promoted to staging. Because verification never succeeds for that freight, it is never eligible for promotion to production. The result is that staging accumulates multiple freight ahead of production, and production falls behind significantly.

If velocity of new images slows down, as long as there is no newer freight queued up, then we see verification succeeding.

We have attempted the following mitigations with no improvement:

1. Disabled auto-sync on our ArgoCD Applications — hoping that removing ArgoCD's own reconciliation loop would reduce race conditions with Kargo's sync step.
2. Added a post-sync empty ConfigMap resource — the intent was to force the argocd-update promotion step to wait for the full rollout of all pods before marking the sync as complete, giving verification a cleaner window to run.

  Neither of these changes resolved the issue.

# Steps to Reproduce

  1. Configure a Kargo pipeline with two stages: staging → production
  2. Both stages use argocd-update promotion steps to sync ArgoCD Applications
  3. Push multiple image updates in rapid succession (high-velocity CI pipeline producing new freight frequently)
  4. Observe that staging promotions happen faster than verification can complete — verification is skipped or never finishes for some freight
  5. Observe that production falls behind because freight never reaches a "verified" state on staging

# Version

```
1.8.4
```

# Logs

```

```

EronWright:
@Nulluminati has the issue been resolved for you?  I have a mild suspicion this issue could be related to https://github.com/akuity/kargo/issues/5282 (fixed in Kargo v1.10).

JohnstonLiu:
@Nulluminati, in case this is still biting you: we hit the same symptom on v1.10.2, where #5282 is already fixed, so that wasn't the cause. Root cause is in the Stage reconciler, not the promotion controller's cache.

[...]

PR #7031, opened 2026-09-02
fix(controller): hold queued Promotions while health settles after a Promotion

### LINKED ISSUE #7032: Queued Promotions supersede unverified Freight while Stage health is still Unknown/Progressing after a successful Promotion
### Pre-submission checklist

### Description

When `Promotion`s to a `Stage` queue up back-to-back, every promoted `Freight` except the last is never verified in that `Stage`, so it never becomes available to downstream `Stage`s. Downstream `Stage`s silently skip it, and no `AnalysisRun` is ever created for it even when the `Stage` has `verification` configured.

Root cause (`pkg/controller/stages/regular_stages.go`, `syncPromotions`): a queued `Promotion` is only held while the `Stage` is `Healthy` (or health is not yet assessed) and the current `Freight` has no terminal verification. Any other health state lets the next `Promotion` start immediately, on the premise that the `Stage` is unhealthy and the next `Promotion` may fix it (#3903).

Immediately after a *successful* `Promotion`, however, health is almost never `Healthy` yet:

- The Argo CD health checker deliberately reports `Unknown` until the `Application` has been reconciled after the operation completed (`reconciledAt >= operationState.finishedAt`) and requests a refresh in the meantime.
- Once reconciled, the `Application` is typically `Progressing` while the rollout completes.

Verification cannot start in either state (`verifyStageFreight` requires `Healthy`). Because `reconcile()` requests an immediate requeue whenever `Promotion`s are pending, the very next pass sees `Unknown`/`Progressing`, bypasses the hold, and starts the next `Promotion`. By the time the `Stage` is `Healthy`, the current `FreightCollection` has changed and verification of the previous `Freight` never begins. The hold documented in the verification guide ("no other `Promotion`s to that `Stage` will run until verification has completed") never engages, because no `AnalysisRun` exists to wait on.

### Prior reports

- #5708 reported the same symptom on v1.8.4 ("verification is skipped or never finishes for some freight ... as long as there is no newer freight queued up, then we see verification succeeding") and was closed on the suspicion that it was #5282 (stale cache re-executing promotion steps, fixed in v1.10). It is not: we reproduce it on v1.10.2 with #5282 fixed, and the mechanism above does not involve the promotion controller's cache. The reporter's second mitigation there (a PostSync resource so the sync waits for the rollout) also would not help, since the `Unknown` state comes from Argo CD not having re-assessed health after the operation, regardless of what the operation waited for.
- #5347 / #5995 cover the related "verification does not start for minutes" symptom (stale health after an operation). #5995 made the health checker return `Unknown` until the `Application` is re-reconciled; that is the state that now trips the bypass when a `Promotion` is queued.
- #3053 / #3903 are the history of the gate itself: #3903 changed the hold condition from "not `Unhealthy`" to "`Healthy`" to fix a deadlock when health is `Unknown` after a *failed* `Promotion`. That case must keep working; the fix proposed below distinguishes it via `LastPromotion.Status.Phase`.
- #3108 and #5342 propose superseding/pausing queued `Promotion`s. Related, but different: this issue is about a `Freight` that *was* promoted never getting the verification window the docs promise.

### Steps to Reproduce

1. A `Warehouse` producing `Freight` frequently (e.g. one image per merge), a `Stage` `a` auto-promoting from it via `argocd-update` against an `Application` whose rollout takes longer than the interval between `Freight`, and a downstream `Stage` `b` with `requestedFreight.sources.stages: [a]`. Optionally give `a` a trivial `verification` (an `AnalysisTemplate` running a `Job` that exits 0).
2. Produce two pieces of `Freight` less than one `Promotion` duration apart, so the second `Promotion` to `a` is queued while the first runs.
3. Observe `Promotion` timestamps: the second `Promotion`'s `startedAt` equals the first's `finishedAt` (same second).
4. Observe the first `Freight`: it has no `status.verifiedIn[a]`, no `AnalysisRun` was created for its `FreightCollection`, and it never appears as available to `b`. The second `Freight` verifies normally.

### Expected Behavior

Every successfully promoted `Freight` gets its verification window before a queued `Promotion` supersedes it, as the verification guide describes, so that `b` sees every `Freight` that was promoted to `a`.

### Actual Behavior

Any `Promotion` that starts while the previous one's `Application` is still settling supersedes the previous `Freight` before verification can begin. In our pipeline over two weeks, 193 of 193 `Promotion`s that started within a second of the previous one finishing left the previous `Freight` unverified, versus 5 of 387 `Promotion`s with a gap between them.

### Kargo Version

v1.10.2 (the relevant logic is unchanged on `main` / v1.11.3)

### Relevant Logs

```shell
# Two consecutive Promotions to Stage "staging" in one Project (kubectl get promotion -o json, abridged):
#   name                                       created               startedAt             finishedAt            phase
#   staging.01m1hecnegnghej43v2jx9zcge.5f3c729 2026-09-02T16:14:17Z  2026-09-02T16:14:17Z  2026-09-02T16:18:10Z  Succeeded
#   staging.01m1heg53x6mgk2pcd0src2jpc.d255a40 2026-09-02T16:16:11Z  2026-09-02T16:18:10Z  2026-09-02T16:21:06Z  Succeeded
# AnalysisRuns in the Project namespace: one for FreightCollection c8c270c4... (the second Freight), none for f482ac39... (the first).
# Downstream Stage "review" (sources.stages: [staging]) received the second Freight only.
```

### Screenshots

_No response_

---

Workaround: an `argocd-wait` step after `argocd-update` makes the `Promotion` end only once the `Application` is reconciled and `Healthy`, so the first reconcile after it sees `Healthy` and verification begins before the next `Promotion` can start.

Proposed fix in #7031: hold queued `Promotion`s while health is `Unknown`/`Progressing` following a *successful* `Promotion`, bounded by a configurable grace period (default 5m) so a `Stage` whose health never settles cannot block the queue indefinitely, while preserving #3903's behavior for `Unknown` health after a failed `Promotion`.

JohnstonLiu:
Closes #7032

## Problem

When `Promotion`s to a `Stage` queue up back-to-back, the `Freight` promoted by each one, except the last, is never verified in that `Stage` and therefore never becomes available to downstream `Stage`s. Downstream `Stage`s silently skip it.

Observed on v1.10.2 with a `Stage` that auto-promotes from a `Warehouse`, uses `argocd-update`, and has a trivial `AnalysisTemplate` (a `Job` that runs `true`): every `Promotion` that started within a second of the previous one finishing (193 of 193 over two weeks) left the previous `Freight` unverified, while `Promotion`s with any gap between them verified normally.

#5708 reported the same symptom on v1.8.4 and was closed on the suspicion that it was #5282 (stale promotion cache, fixed in v1.10). It reproduces on v1.10.2 with #5282 fixed; the mechanism is in the `Stage` reconciler, described below.

## Root cause

`syncPromotions` only holds a queued `Promotion` while the `Stage` is `Healthy` (or health has not been assessed) and the current `Freight` has no terminal verification:

```go
if stage.Status.Health == nil || stage.Status.Health.Status == kargoapi.HealthStateHealthy {
    // hold until the current Freight has been verified
}
```

Anything else falls through to starting the next `Promotion`, on the premise that the `Stage` is unhealthy and the next `Promotion` may fix it.

Immediately after a successful `Promotion`, though, health is almost never `Healthy` yet:

- The Argo CD health checker deliberately reports `Unknown` until the `Application` has been reconciled after the operation completed (`reconciledAt >= operationState.finishedAt`), requesting a refresh in the meantime (#5995).
- Once reconciled, the `Application` is typically `Progressing` while the rollout completes.

Verification cannot start in either state (`verifyStageFreight` requires `Healthy`). And because `reconcile()` requests an immediate requeue whenever `Promotion`s are pending, the very next pass sees `Unknown`/`Progressing`, bypasses the hold, and starts the next `Promotion`. By the time the `Stage` is `Healthy`, the current `FreightCollection` has changed, so the previous `Freight`'s verification never begins. The AnalysisRun-based hold documented in the verification guide never engages, because no `AnalysisRun` is ever created.

The current condition comes from #3903, which fixed a real deadlock: after a *failed* `Promotion`, health is `Unknown` for lack of health checks to perform, and holding in that state blocked the `Stage` forever. That case is preserved here.

## Fix

Hold queued `Promotion`s while health is `Unknown` or `Progressing` *following a successful `Promotion`*, for a bounded period:

- `awaitingVerificationStart` returns true when the last `Promotion` succeeded, health is `Unknown`/`Progressing`, the current `Freight` has no terminal verification, and the configured grace period since the `Promotion` completed has not elapsed.
- The bound (`VERIFICATION_START_GRACE_PERIOD`, default 5m, matching `argocd-wait`'s default timeout) guards against a `Stage` whose health never settles (e.g. a `StatefulSet` rollout stuck `Progressing`) holding the queue indefinitely. Once it elapses, behavior reverts to today's: the next `Promotion` runs in the expectation that it can fix things. Setting it to `0` disables the hold entirely.
- `Unknown` after a failed `Promotion` never triggers the hold, so #3903's fix stands.
- Manually created `Promotion`s (those carrying a `create-actor` annotation, which the webhook sets for every request not originating from the Kargo control plane) are never held. A person promoting while a rollout settles is usually rolling back, and unlike a running verification this wait cannot be aborted, so holding them would leave an operator with no way out for the length of the grace period. Auto-promotions, where the high-velocity skips actually happen, are held.
- While holding, `reconcile()` no longer requests the 100ms immediate requeue (which would spin for the duration of the wait). The `Application` watcher re-enqueues the `Stage` as health changes, and the `Stage` is requeued for when the grace period expires. To express that, `reconcile()` now returns a requeue duration instead of a bool.

Exposed in the chart as `controller.reconcilers.stages.verificationStartGracePeriod`, and documented in the verification guide.

## Testing

- New unit tests for `awaitingVerificationStart`, for `syncPromotions` (hold on `Unknown`/`Progressing` after success; no hold after a failed `Promotion`, after the grace period, with a terminal verification, or when disabled), and for `reconcile()` (no immediate requeue while holding; timed requeue for the grace period expiry).
- Helm unit tests for the new ConfigMap key (default, custom, `0s`, unset).
- `make lint-go`, `go test ./pkg/controller/stages/...`, `make test-chart` all pass locally.


krancour:
Things like this need to be discussed prior to implementation. Let's move the conversation to #7032.

PR #7052, opened 2026-09-03
feat(controller): record a Stage's freight history from succeeded PromotionRequests

jsbroks:
Follows **#7006**, which partitioned a PromotionRequest's child Promotions out of the Stage's own flow and documented this as the interim state to fix.

## Summary

A Stage's `freightHistory` had exactly one writer: `syncPromotions`, recording
when a succeeded Promotion becomes `lastPromotion`. A Stage that promotes
through PromotionRequests never has such a Promotion — its Promotions are
children of a request — so its history never advanced, and everything derived
from `FreightHistory.Current()` went with it: `syncFreight` marking Freight
`currentlyIn` the Stage, `freightSummary`, auto-promotion's "already current"
check, and the UI's current-Freight display.

The round's outcome is now recorded from the unit that represents it. When the
newest terminal PromotionRequest is `Succeeded` and is first recorded as the
Stage's `lastPromotionRequest`, `syncPromotionRequests` fetches the Freight the
request names, builds the collection the Stage is now running, records it, and
resets health and verification exactly as `syncPromotions` does. A partial round
records nothing, as a failed Promotion does.

## The collection is built at record time, not stored on the request

An earlier draft added `status.freightCollection` to PromotionRequest, built
when the request is created. That is too early: requests queue, so a request
created while another is running would snapshot the Stage's other origins
before the running round finished changing them, and recording that snapshot
later would quietly roll those origins back. Promotions avoid this because
their collection is built only once they are admitted to run. Building at
record time gives a request the same guarantee, and means **no API changes**
and **no EE changes** — EE inherits this on its next CE bump. If a request ever
needs to carry its collection (for the UI, say), it should be computed at
admission, not creation.

## Once per request

`lastPromotionRequest` only moves forward by name, so a request is recorded the
moment it first becomes the Stage's last and never on the reconciles after —
which matters because `Record` is prepend-only and recording resets
health/verification. This is exact today, where a request completes once. A
request that could later re-open and complete again (single-Target retry) would
need history entries to name the request that produced them; that belongs with
the change that makes re-opening possible.

## The shared helper, and a multi-origin fix

`buildTargetFreightCollection` moves out of the promotion controller into
`api.NewFreightCollectionForStage`, so a request and its child Promotions are
built by one function. It carries the Stage's other requested origins over from
`lastPromotion` as before, and now **falls back to `freightHistory.Current()`**
when there is no `lastPromotion` — which, for a Stage that promotes through
requests, is always. Without the fallback a two-origin Stage's collection named
only the promoted origin, and recording it would have dropped the other.

## What this does not change

Health and verification remain gated on `lastPromotion`, so Freight is still
not marked verified-in a target-aware Stage, and downstream Stages still cannot
consume from one. Aggregating Target health and verification onto the Stage is
separate follow-up work. The one visible change: `assessHealth`'s `NoFreight`
reason ("Stage has no current Freight") is false for such a Stage once history
records Freight, so it now reports `TargetAwareStage` instead.

## Tests

- `pkg/api`: the helper — single origin, inheritance from `lastPromotion`,
  fallback to history, precedence, dropping no-longer-requested origins.
- `syncPromotionRequests`: records + resets; Failed records nothing; Freight
  gone records nothing; fetch error returned; already-recorded request not
  recorded again; newer request records on top; and a queued two-origin round
  inherits the round recorded before it.
- `assessHealth`: the `TargetAwareStage` reason.


pkg/controller/stages/regular_stages.go
```diff
@@ -1049,13 +1049,108 @@ func (r *RegularStageReconciler) syncPromotionRequests(
...
+				ctx,
+				stage,
+				&newStatus,
+				promotionRequest,
+			); err != nil {
+				return newStatus, err
+			}
 		}
 		break
 	}
```
fuskovic, line 1082: The break here means we're silently dropping older terminal promotion requests so their `Freight` never reaches `FreightHistory`.

I believe the correct order of operations is:

1. Collect every terminal request newer than `LastPromotionRequest`
2. Sort oldest-first
3. Record each in order instead of breaking after the first.

Based on a similar pattern in `syncPromotions`

fuskovic (changes requested):

fuskovic (approved):

PR #7004, opened 2026-09-01
feat(ui): visualize PromotionRequests on a Stage's Promotions tab

jsbroks:
## Description

Builds on the merged PromotionRequest series (most recently #6843). Two commits:

**`feat(api): read PromotionRequests through the REST API`** adds
`GET /v1beta1/projects/{project}/promotion-requests` (list, optionally filtered
by Stage, with watch support) and `GET .../promotion-requests/{name}`, modelled
directly on the Promotion equivalents, and grants project users read-only
access. When a target-aware Stage stops promoting, the PromotionRequest's
status is the only place the reason is recorded, so withholding reads would
make the failure invisible.

**`feat(ui): visualize PromotionRequests on a Stage's Promotions tab`** is the
first UI for PromotionRequests, scoped entirely to target-aware Stages —
classic Stages render exactly as before, and the gating predicate
(`isStageTargetAware`, i.e. `spec.targets != null`) mirrors `api.IsTargetAware`
on the back end. For a target-aware Stage, the Promotions tab now leads with a
Promotion Requests table:

- Phase, name, Freight (linked), a per-phase summary of the Targets the request
  names, and creation time, newest first (names embed a ULID, so name order is
  creation order).
- Each request expands to the Targets it names — merged from `spec.targets`
  (the resolved snapshot) and `status.targets` (what became of each) — with the
  child Promotion opening in the existing Promotion drawer. The request
  currently fanning out starts expanded.
- When the newest request is not going to progress, its `Ready` condition's
  message surfaces as an alert. In Kargo OSS that is how a user learns that
  fanning Freight out to Targets is an Enterprise-only feature.
- Live updates ride the same seeded SSE watch mechanism the Promotions table
  uses.

Promoting to a target-aware Stage now lands on the Stage's Promotions tab
instead of navigating to a Promotion drawer that could never load, since the
promote endpoint answers with a PromotionRequest for such a Stage.

### Reviewer notes

- The pure helpers (ordering, terminal-phase check, blocking message,
  spec/status target merge) live in `ui/src/features/stage/utils/promotion-request.ts`
  with unit tests, including the case where status names a Target the cached
  spec does not yet.
- Verified end to end on a kind cluster: promoted Freight to a target-aware
  Stage, watched the request appear live, flip to `Errored` with the
  Enterprise-only alert, and expand to its three pending Targets. Classic
  Stages show an unchanged Promotions tab.
- `swagger.json`, the generated Go client, and the generated TS client are
  regenerated; all other diffs are hand-written.

[...]

PR #7070, opened 2026-09-04
feat(ui): read a target-aware Stage's promotion state from its PromotionRequests

jsbroks:
## Problem

A target-aware Stage promotes Freight through a PromotionRequest, which fans out
one child Promotion per Target. Its `status.lastPromotion` is therefore never
populated — verified on a live Enterprise cluster, where a Stage mid-promotion
with three children in flight still reports an empty `currentPromotion` and
`lastPromotion`.

The UI read those fields as the Stage's promotion state, so for a target-aware
Stage it had nothing to show. Stage status already records the request in
`currentPromotionRequest` / `lastPromotionRequest` (added in #6843).

## Approach

Two helpers in `stage-meta-utils.ts`, `getCurrentPromotionRef` and
`getLastPromotionRef`, normalize the two shapes into one `StagePromotionRef`
carrying kind, name, phase, message, finish time, Freight name, and a link path.
Classic Stages read the Promotion fields exactly as before; target-aware Stages
read the PromotionRequest fields. Consumers branch once, in one place, on the
existing `isStageTargetAware`.

A Promotion has a page of its own, so its ref links there. A PromotionRequest
does not, so its ref links to the Stage page, whose Promotions tab already leads
with the Stage's requests and expands the one fanning out (#7004). This PR adds
no per-request route or drawer.

## What changes visibly

Verified against a live Enterprise cluster by running this branch and its parent
commit side by side:

- **Pipeline graph node** — target-aware Stages gain the "Last Promotion: N ago"
  footer, which previously did not render at all. It links to the Stage page.
- **List view "Last Promotion" column** — `-` becomes a real relative time for
  target-aware Stages, likewise linking to the Stage page.

Classic Stages are pixel-identical in both. Screenshots available on request.

## What is defensive rather than a live fix

Worth a reviewer's attention, since it may argue for trimming this PR. I
verified each of these on the cluster rather than assuming:

- **Phase-indicator links** (`stage-node-phase.tsx`, `phase-cell.tsx`) — a
  target-aware Stage never reports `Promoting`. Its conditions are
  `Ready=False (TargetAwareStage)`, `Healthy=Unknown`, `Verified=Unknown`, so the
  phase renders as `Verifying` and this branch never executes. Whether such a
  Stage *should* report `Promoting` looks like a separate question.
- **Step-label suppression** (`useCurrentPromotion`, consumed by
  `StepWaitingLabel` and `PullRequestLink`) — `currentPromotion` is already
  empty on these Stages, so the query was already disabled. This only makes the
  intent explicit.
- **Stage tag and popover** (`stage-tag.tsx`, `stage-popover.tsx`) — the only
  call site, `requested-freight.tsx`, passes a name-only stub
  (`{ metadata: { name: stage } } as Stage`) with no `spec` and no `status`, so
  both helpers return undefined and neither the icon nor the popover renders.
  Confirmed: the rendered chip contains zero SVG icons.

The one behavior these do change is for **classic** Stages: a `Promoting`
indicator with no Promotion to link to now renders as plain text instead of a
link to `/project/<p>/promotion/` with an empty name.

If reviewers prefer this tight, it can shrink to `stage-meta-utils.ts`,
`stage-node.tsx` and `last-promotion-column.tsx`.

## Also touched

- `promotion-status-icon.tsx` gains an optional `subject` prop so a tooltip can
  read "Promotion Request Errored". Additive; the default is unchanged.
- `stage-details.tsx` sources the Freight History `currentActiveFreight`
  highlight from the ref. Not visually verified.

## Testing

- `stage-meta-utils.test.ts`, 15 cases: classic and target-aware for both
  current and last, link paths, the absent-path cases, and that a target-aware
  Stage never yields a single current Promotion.
- `pnpm typecheck`, `pnpm lint` and the full `pnpm vitest run` (30 files, 340
  tests) all pass.
- End-to-end on a live Enterprise cluster: promoted a target-aware Stage, held a
  request `Running` by pausing the promotion controller to observe three parallel
  children, and confirmed the rendered links and timestamps.

## Note on the commit message

The commit body claims the phase indicator "links to one arbitrary Target's
Promotion (or a broken link)" and that the step labels "read one Promotion's
steps". Both predate the verification above and are inaccurate, since
`currentPromotion` is never set on these Stages. Happy to amend the commit if
preferred.

### DIFF
--- a/ui/src/features/common/promotion-status/promotion-status-icon.tsx
+++ b/ui/src/features/common/promotion-status/promotion-status-icon.tsx
@@ -3,19 +3,26 @@ import { PromotionStatus } from '@ui/gen/api/v2/models';
 
 import { getPromotionPhasePresentation } from './promotion-phase';
 
-const PhaseAndMessage = ({ status }: { status: PromotionStatus }) => (
+const PhaseAndMessage = ({ status, subject }: { status: PromotionStatus; subject: string }) => (
   <div>
-    <div className='font-semibold'>Promotion {status.phase}</div>
+    <div className='font-semibold'>
+      {subject} {status.phase}
+    </div>
     <div>{status.message}</div>
   </div>
 );
 
 export const PromotionStatusIcon = ({
   status,
+  subject = 'Promotion',
   color,
   ...props
 }: {
   status?: PromotionStatus;
+  // subject names what the phase belongs to in the tooltip. A
+  // PromotionRequest's phases share the Promotion vocabulary, so the same icon
+  // presents both.
+  subject?: string;
   placement?: 'right' | 'top';
   color?: string;
   size?: 'lg' | '1x';
@@ -27,7 +34,7 @@ export const PromotionStatusIcon = ({
 
   return (
     <MessageTooltip
-      message={<PhaseAndMessage status={status} />}
+      message={<PhaseAndMessage status={status} subject={subject} />}
       icon={icon}
       iconColor={color ? color : iconColor}
       spin={spin}
--- a/ui/src/features/common/stage-tag.tsx
+++ b/ui/src/features/common/stage-tag.tsx
@@ -1,5 +1,9 @@
 import { Tooltip } from 'antd';
 
+import {
+  getCurrentPromotionRef,
+  getLastPromotionRef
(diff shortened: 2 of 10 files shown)


rpelczar (approved):
LGTM

PR #7109, opened 2026-09-05
docs: fix info directive in managing secrets

yashrajshuklaaa:
Fixed the info block formatting in the managing secrets docs and verified the docs build passes.


yashrajshuklaaa:
@krancour ptal

PR #7139, opened 2026-09-08
chore: Refactor stages reconciler to make it easier to make changes

hairyhum:
Main changes:
- Break the `regular_stages.go` into multiple files for sub-reconciler functions
- Rework `syncPromotions` to separate branching from operations and simplify analysis of promotions backlog

`syncPromotions` rework maintains the logic but changes the way it analyzes promotions:
- Remove the `highestPrioPromo` and use of `ComparePromotionByPhaseAndCreationTime`
- Instead analyze promotions to get the `promotionsSummary` struct with:
  - Terminal promotions - used for autopromotion holds, last promotion and freight history recording
  - `hasNonTerminalPromotions` flag - used to decide if we need to reconcile immidiately
  - `currentPromotion` is a promotion object referenced by CurrentPromotion field in the stage status, nil if not found in promotions
    Used in determining if we need to finalize a promotion (if it terminates) and in tracking the CurrentPromotion
    Replaces `highestPrioPromo`, but references CurrentPromotion directly instead of first order of running->pending->terminal
    Used to determine if CurrentPromotion is terminated or no longer exists in the promotions.
  - `nextPromotion` is a first promotion in the running->pending order, can overlap with `currentPromotion`
    Replaces `highestPrioPromo` for the purpose of tracking CurrentPromotion status (if it did not terminate)
    and to unblock next pending promotion.
- Break down `syncPromotions` into mutiple functions:
  - `finalizePromotion` - to call when promotion finishes (or is gone)
  - `verificationBlocksPromotion` - check if verification blocks unblocking pending promotion
  - `trackPromotion` - track current promotion status OR unblock a new pending promotion


hairyhum:
**Note for reviewers:** I did at first messed up the order in `getPromotionsSummary` and no unit tests detected that. I had to double-check with my test and manually.

PR #6848, opened 2026-08-20
chore(api): empty out unused Target status

jsbroks:
Empties out the Target status rather than carrying fields nothing reads.
None of the status fields (`conditions`, `ownedBy`) were used anywhere --
no controller writes them and nothing in the server, UI, or docs reads
them. TargetStatus stays in place as an empty struct, along with the
status subresource and the `GetStatus()` accessor, so fields can be
added back one at a time as something actually needs them.

Also removes the Ready/Status printer columns, which pointed at the
now-removed conditions. CRD manifest, deepcopy methods, and the UI's
JSON schema are regenerated to match; the REST API and its generated
clients never exposed Target status, so they are untouched.

### Eligibility

- [ ] Linked to an existing issue with no blocking labels (`kind/proposal`, `needs discussion`, `needs research`, `maintainer only`, `area/security`, `size/large`, `size/x-large`, `size/xx-large`).
- [ ] Changes documentation only.
- [ ] Changes ten lines or fewer.

### Quality

- [x] Adds or updates corresponding tests.
- [ ] Adds or updates corresponding documentation.

### AI Use Disclosure

This PR was written:

- [ ] By a human _without_ AI assistance.
- [ ] By a human _with_ AI assistance. A human has reviewed every line prior to opening the PR.
- [x] By an AI _with human supervision._ A human has reviewed every line prior to opening the PR.
- [ ] Entirely by an AI. No human has reviewed this prior to opening the PR.

### Sign-Off

All commits:

- [x] Are signed off by their author (`git commit -s`) **(required)**
- [ ] Are cryptographically signed (`git commit -S`) **(encouraged)**

PR #7174, opened 2026-09-09
feat(api): record per-Stage current Freight, health checks, and health on Target status

jsbroks:
# feat(api): record per-Stage current Freight, health checks, and health on Target status

## Summary

A target-aware Stage promotes Freight to its Targets by fanning a PromotionRequest out into one child Promotion per Target. Since the children run in parallel (#7006) they no longer feed the Stage's own status, and #7052 records only the round's Freight on the Stage once every Target has succeeded. Nothing records what became of each individual Target: which Freight is on it, which Argo CD Applications that Promotion updated, and whether they are healthy. Target status has been an empty struct since #6848, deliberately, until there was a concrete design and a consumer.

This PR adds the types only. No controller reads or writes them yet. The writers -- recording each Target's Freight and health checks when a child Promotion succeeds, and assessing health from them -- and the Stage-level aggregate that folds Target state back into Stage status follow in separate PRs.

```yaml
status:
  stages:
    fleet:                          # keyed by governing Stage
      currentFreight:               # FreightCollection: one entry per origin
        id: 9d1c0e...
        items:
          Warehouse/nginx: {name: f-abc, images: [...]}
        verificationHistory: []     # per-Target verification results go here
      health: {status: Healthy}     # ongoing observation
      healthChecks:                 # what to observe
      - uses: argocd-update
        config: {apps: [{name: nginx-us-east-1, desiredRevisions: ["3f2a9c"]}]}
    fleet-multi:
      ...
```

## Each property and what it is for

**`status.stages` — map keyed by Stage name.**
A Target is selected by label selectors, so several Stages can govern the same Target at once, each promoting its own Freight with its own health checks. Everything observed about a Target is therefore scoped to the Stage that observed it. A map keyed by Stage name is how Freight status already records per-Stage state (`currentlyIn`, `verifiedIn`, `approvedFor`); it makes duplicate entries impossible, gives upsert and delete without a search, and reads naturally in kubectl JSONPath (`.status.stages.fleet.health.status`) and as a `Record` in the UI. An entry exists only once a Promotion from that Stage to the Target has succeeded.

**`stages[<stage>].currentFreight` — `FreightCollection`.**
The Freight that Stage's most recent successful Promotion rendered onto this Target. It is a collection rather than a single reference because a Stage may request Freight from several origins and one Promotion changes only one of them; the collection holds one entry per origin so the Target's full desired state is known, exactly as the Stage's own current collection does. It is copied from the child Promotion's status rather than derived from the Target's previous entry, so a Target that missed a round or was newly discovered converges to the Stage's desired state on its next successful Promotion. Reusing `FreightCollection` brings two things with it: an `id` that is deterministic in its contents, so a Target that received the same round as the Stage carries the same `id` as the Stage's current collection and the two can be compared to tell whether the Target is up to date; and a `verificationHistory` in which per-Target verification results will be recorded, so a result is always attached to the specific Freight it verified.

**`stages[<stage>].health` — `Health`.**
The Target's health with respect to that Stage's Freight, as last assessed by executing `healthChecks`. Absent until the first assessment. Health is an ongoing observation rather than a Promotion outcome: it is reassessed periodically and whenever the observed systems change, so it may become Unhealthy long after the Promotion succeeded. This is the same `Health` type the Stage records for itself, so the Stage-level aggregate can fold Target health together without translation.

**`stages[<stage>].healthChecks` — `[]HealthCheckStep`.**
The health check directives produced by the steps of that Stage's most recent successful Promotion to this Target — for instance, which Argo CD Applications it updated and to what revisions. They are the input from which `health` is assessed. They are recorded on the Target because health must be reassessed for as long as `currentFreight` remains there, while the Promotion that produced them is eventually garbage collected. A Stage keeps the same information for itself by embedding its last Promotion's status in `lastPromotion`; copying only the directives is the smaller form of the same move.

**Deliberately not added.** A `lastPromotion` or `promotionRequest` name: neither is needed as an idempotency key, since the writer can overwrite `currentFreight` and `healthChecks` on every pass (identical bytes are a no-op patch) and clear `verificationHistory` only when the collection `id` changes. Top-level `conditions`: Ready is not meaningful for a purely descriptive object. Printer columns: none has anything to show yet.

## Also in this PR

- `SetStageStatus` / `RemoveStageStatus`: nil-map-safe helpers for the writers, modeled on the Freight status helpers, with tests.
- Two stale doc comments corrected: `Target` no longer claims to record which Stages govern it (nothing does), and `StageSpec.targets` no longer describes an implicit "stage-self" Target that no controller creates.
- Generated: CRD manifests for Targets and Stages, deepcopy methods, and the UI's JSON schemas. Targets have no REST routes, so the OpenAPI spec and generated clients are untouched.

### Eligibility

- [x] Linked to an existing issue with no blocking labels.

### Quality

### AI Use Disclosure

### Sign-Off


api/v1alpha1/target_types.go
```diff
@@ -41,9 +41,85 @@ type TargetSpec struct {
...
+	// them is eventually garbage collected. A Stage keeps the same
+	// information for itself in its lastPromotion.
+	//
+	// +optional
+	HealthChecks []HealthCheckStep `json:"healthChecks,omitempty"`
+}
+
+// SetStageStatus records the Target's status with respect to the specified
+// Stage, replacing any existing record for that Stage.
+func (s *TargetStatus) SetStageStatus(stage string, status TargetStageStatus) {
```
thomastaylor312: Super nit: This naming is a little weird to me on first read, maybe `SetStatusForStage`

api/v1alpha1/target_types.go
```diff
@@ -41,9 +41,85 @@ type TargetSpec struct {
...
+	if s.Stages == nil {
+		s.Stages = make(map[string]TargetStageStatus)
+	}
+	s.Stages[stage] = status
+}
+
+// RemoveStageStatus removes the Target's status with respect to the specified
+// Stage, if any. It is a no-op when no such record exists.
+func (s *TargetStatus) RemoveStageStatus(stage string) {
```
thomastaylor312: Same nit around naming here

api/v1alpha1/target_types.go
```diff
@@ -41,9 +41,85 @@ type TargetSpec struct {
...
+	CurrentFreight *FreightCollection `json:"currentFreight,omitempty"`
+	// Health is the Target's health with respect to this Stage's Freight, as
+	// last assessed by executing HealthChecks. It is absent until the first
+	// assessment. Health is an ongoing observation rather than a Promotion
+	// outcome: it is reassessed periodically and whenever the systems the
+	// health checks observe change, so it may become Unhealthy long after the
+	// Promotion that produced CurrentFreight succeeded.
+	//
+	// +optional
+	Health *Health `json:"health,omitempty"`
```
thomastaylor312, line 100: Just for my edification: So is the idea that for target-aware stages the health of the stage itself is an aggregation of these health statuses?

thomastaylor312 (commented):
Nothing but nits here and a question. I think this is a solid start

api/v1alpha1/target_types.go
jsbroks: Excatly, each Target's health comes from its own healthChecks. A target-aware Stage health is a fold over its Targets' entries: Healthy when all are Healthy on the Stage's current collection, Unhealthy if any isn't, Unknown while any is unassessed. Verified works the same way, and only the aggregate marks Freight verifiedIn the Stage. That's why these fields reuse the Stage's own types.

thomastaylor312 (approved):

